PR #168, opened 2026-03-11
fix: use configured default model in built-in chat

### LINKED ISSUE #167: Built-in chat uses first model in list instead of configured default model
## Bug Description

The built-in chat UI at `/admin/chat` always selects the first model in the list (`availableModels[0]`) when initializing, instead of using the server's configured `default_model`.

## Steps to Reproduce

1. Configure a default model via admin page (model_settings.json)
2. Have multiple models available
3. Open `/admin/chat` with a fresh browser session (no localStorage)
4. The first model in the list is selected, not the configured default

## Root Cause

**File: `omlx/admin/templates/chat.html:765-766`**
```javascript
// Auto-select first model if none selected
if (!this.currentModel && this.availableModels.length > 0) {
    this.currentModel = this.availableModels[0].id;
}
```

The `/v1/models` endpoint doesn't include `is_default` information, so the chat UI cannot know which model is the default.

## Proposed Fix

**Option 1:** Add `is_default` to `/v1/models` response
- Modify `ModelInfo` in `omlx/server.py` to include `is_default` field
- Update chat.html to select the default model

**Option 2:** Chat calls health endpoint first
- The `/health` endpoint already returns `default_model`
- Chat could fetch this and use it for initial selection

## Related Code

- `omlx/server.py:1122` - Health endpoint returns `default_model`
- `omlx/server.py:1128-1150` - `/v1/models` endpoint (no default info)
- `omlx/admin/routes.py:1138` - Admin models endpoint includes `is_default`

jwcrystal:
Fixes #167

## Summary
- Built-in chat now selects the configured default model instead of the first model in the list
- Fetches `/health` endpoint in parallel with `/v1/models` to get `default_model`
- Falls back to first model if no default is configured

## Test Plan
- [ ] Set a default model via admin page
- [ ] Clear browser localStorage (fresh session)
- [ ] Open `/admin/chat`
- [ ] Verify the default model is selected, not the first in list

### DIFF
--- a/omlx/admin/templates/chat.html
+++ b/omlx/admin/templates/chat.html
@@ -740,14 +740,18 @@
 
             async loadModels() {
                 try {
-                    const response = await fetch('/v1/models', {
-                        headers: {
-                            'Authorization': `Bearer ${this.getApiKey()}`
-                        }
-                    });
+                    // Fetch models and server health in parallel
+                    const [modelsResponse, healthResponse] = await Promise.all([
+                        fetch('/v1/models', {
+                            headers: {
+                                'Authorization': `Bearer ${this.getApiKey()}`
+                            }
+                        }),
+                        fetch('/health')
+                    ]);
 
-                    if (!response.ok) {
-                        if (response.status === 401) {
+                    if (!modelsResponse.ok) {
+                        if (modelsResponse.status === 401) {
                             alert(window.t('chat.error.incorrect_api_key'));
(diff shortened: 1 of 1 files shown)


jundot (approved):
Looks great, nice work on this one! Love the parallel fetch approach and all the edge cases are handled really well. Clean and solid change overall, thanks for putting this together.

PR #140, opened 2026-03-10
Fix tool-call markup leakage/truncation in streamed chat output (OpenAI path)

lyonsno:
### Summary

This PR hardens tool-enabled streaming so assistant-visible deltas stay clean without truncating ordinary prose.

It fixes several edge cases where control markup could leak (or prose could be dropped) during streaming, especially
around bracket-style tool envelopes and long payloads.

### Contract (what this PR enforces)

For tool-enabled streaming:

- Suppress high-confidence tool-control markup from visible assistant text.
- Preserve ordinary prose when no parseable tool call exists.
- Keep structured tool_calls extraction correct at completion.

### What changed

- Reworked ToolCallStreamFilter to be envelope-bounded (not permanently suppressing after first marker).
- Added explicit open/close marker-pair handling for:
    - raw <tool_call>...</tool_call>
    - tokenizer-defined delimiters
    - namespaced tags (including hyphen/dot namespaces)
    - bracket envelopes ([Calling tool: ...]) when parseable
- Fixed long unresolved bracket envelope behavior:
    - unresolved bracket payloads are no longer truncated by the generic 128-char retention cap (which previously
        leaked raw markup).
- Broadened bracket tool-name parsing from \w+ to [A-Za-z_][\w.-]* to support common names like get-weather.
- Refined finish-tail behavior:
    - still drops true unresolved control fragments,
    - preserves non-tool literal tails like "<alpha".

### Scope

- Reviewer acceptance scope remains OpenAI /v1/chat/completions streaming behavior.
- No Anthropic-specific test expansion in this slice.
- ToolCallStreamFilter is shared utility code, but this PR’s behavioral contract is enforced via OpenAI streaming
    tests.

### Tests added/updated

Key new regressions covered:

- envelope suppression must preserve trailing prose (...tool call...] After text)
- long bracket envelope does not leak raw [Calling tool: ... markup
- hyphenated bracket tool names parse and suppress correctly
- non-tool trailing <identifier literals are preserved

### Notes

- README tool-calling streaming text was aligned earlier in this branch to match current behavior (incremental content
    streaming + control-markup suppression + structured tool-call emission at completion).

PR #174, opened 2026-03-11
Follow-up: harden bracket marker streaming suppression edge cases

lyonsno:
## Summary

Follow-up to merged tool-streaming cleanup (`#140`) that hardens unresolved/mixed bracket marker handling in the shared stream filter.

This PR addresses remaining edge cases where bracket-style control prefixes could leak into visible assistant text or where early marker-like text could interfere with later parseable envelopes.

## Scope

- `omlx/api/tool_calling.py`
- `tests/test_tool_calling.py`

## What changed

1. Strengthened bracket envelope detection and partial-suffix handling for mixed streams.
2. Added suppression/retention coverage for unresolved bracket fragments and mixed literal+parseable bracket sequences.
3. Added prefix sanitization before suppression so unresolved early bracket control prefixes are removed while preserving surrounding prose.
4. Kept structured tool-call emission behavior intact for later parseable envelopes.

## Key behavior guarantees

- No raw `[Calling tool:` / `[Tool call:` control prefix leakage in streamed assistant content for supported parseable paths.
- Literal bracket prose remains visible when not classified as suppressible control envelopes.
- Later valid bracket envelopes are still recognized and emitted as structured `tool_calls`.

## Validation

- Added/updated unit and integration tests for:
  - unresolved bracket tail drop
  - mixed literal + later parseable envelope
  - unresolved prefix before later parseable envelope
- Local syntax check: `python3 -m py_compile` on touched files

### DIFF
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -479,19 +479,25 @@ class ToolCallStreamFilter:
 
     def _sanitize_prefix_before_suppression(self, text: str) -> str:
         """Strip unresolved bracket-control prefixes while preserving prose."""
-        if self._bracket_prefix not in text:
+        if not any(bp in text for bp in self._bracket_prefixes):
             return text
 
         out: List[str] = []
         cursor = 0
         while cursor < len(text):
-            bracket_idx = text.find(self._bracket_prefix, cursor)
+            bracket_idx = -1
+            bracket_prefix = ""
+            for bp in self._bracket_prefixes:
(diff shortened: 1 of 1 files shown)


jundot:
thanks, nice work. thorough edge case coverage and the while-loop fix for multi-bracket detection is solid. merging now, i'll add a couple more test cases (incremental feeding split + [Tool call:] prefix variant) as a follow-up commit on main.

PR #177, opened 2026-03-12
fix(cache): use cache-layer count for hybrid model validation

rsnow:
## Problem

Hybrid models like NVIDIA Nemotron-3 Super (Mamba + MoE + Attention) produce fewer cache entries than architectural layers, because some layer types do not maintain KV cache state. For Nemotron-3 Super 120B:

- **88 total layers** (`len(model.layers)`)
- **48 cache layers** (`len(model.make_cache())`) — only Mamba (ArraysCache) and Attention+MoE (KVCache) layers; MoE-only layers produce no cache

The prefix cache layer-count validation in `BlockAwarePrefixCache._get_model_num_layers()` compared the saved block layer count (48) against `len(model.layers)` (88), causing every cache hit to be invalidated:

```
Cache layer count mismatch: block has 48 layers, expected 88. Invalidating cache hit.
Request ...: paged cache reconstruction failed, released shared blocks
```

**Result:** 0% cache efficiency — every request re-evaluates the full prompt from scratch.

## Fix

Change `_get_model_num_layers()` to prefer `len(model.make_cache())` as the source of truth for expected cache layer count, falling back to `len(model.layers)` for models without `make_cache()`.

This is a generic fix that works for any hybrid architecture where `make_cache()` returns fewer entries than total layers. Non-hybrid models are unaffected — `make_cache()` returns the same count as `len(model.layers)`.

## Testing

- **Nemotron-3 Super 120B (8-bit, gs64):** Cache hit confirmed after fix. Logs show `Extracted 48 layer states` with successful block restoration. No mismatch warnings.
- **MiniMax M2.5 (5-bit):** No change in behavior (all layers produce cache, counts match).

### DIFF
--- a/omlx/cache/prefix_cache.py
+++ b/omlx/cache/prefix_cache.py
@@ -114,23 +114,38 @@ class BlockAwarePrefixCache(CacheManager):
 
     def _get_model_num_layers(self, model: Any) -> int:
         """
-        Get the number of layers in the model.
+        Get the expected number of *cache layers* for validation.
+
+        For hybrid models, the number of cache entries (from ``make_cache()``)
+        may be smaller than the architectural layer count (``model.layers``),
+        because some layer types do not produce cache state.
 
         Args:
             model: The MLX model
 
         Returns:
-            Number of layers, or 0 if cannot be determined
+            Number of cache layers, or 0 if cannot be determined
         """
-        # Try various ways to get num_layers
+        # Prefer cache-layer count when available (hybrid-model safe).
+        make_cache = getattr(model, 'make_cache', None)
+        if callable(make_cache):
+            try:
+                cache_list = make_cache()
+                if isinstance(cache_list, list) and len(cache_list) > 0:
+                    return len(cache_list)
+            except Exception as e:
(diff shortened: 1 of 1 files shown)


jundot:
thanks for this. touching the cache structure needs a bit more careful testing, so i'd like to review and merge when i have more time. i'll take a look over the weekend!

rsnow:
Understood, we're soaking it here live the same...so far so good, but the more pressure the better.

jundot:
Took a closer look at this over the weekend as promised. Traced the full data flow from model.make_cache() through save_block() and reconstruct_cache() validation. The fix is solid. make_cache() is the actual source of truth for cache layer count, so using it for validation just makes everything consistent.

Also verified there's no regression risk for non-hybrid models since make_cache() returns the same count as model.layers for standard transformers. Exception handling and fallback logic are clean too. Really nice work @rsnow, appreciate the thorough PR description and testing on both nemotron and minimax.

Cherry-picked your commit (e5e1b81 -> c29f16b) directly onto main since the only conflict was the stale version bump in _version.py. Your authorship is preserved in the commit history.

PR #178, opened 2026-03-12
feat(i18n): add Traditional Chinese (zh-TW) locale

JianShan-1214:
Add zh-TW.json with 470 translation keys using standard Taiwan terminology.

Update language selector in settings UI to include 中文（繁體）, add CJK font handling for zh-TW in base template, and update all existing locale files (en, zh, ko, ja) with the new settings.language.zh-TW key.


jundot:
looks good, thanks for adding zh-TW support. merging now.

PR #195, opened 2026-03-12
Re-slice reasoning parsing hardening onto current upstream main

lyonsno:
## Summary

This PR hardens think-tag parsing in streaming adapters and locks in the stream-state contract needed to avoid reasoning/content corruption across chunk boundaries and equivalent request objects.

Scope in this branch:
- fix think-tag parsing in the OpenAI streaming path
- harden `extract_thinking()` / thinking-parser behavior for visible-content compatibility
- isolate OpenAI streaming parser state across interleaved same-payload streams
- make equivalent-request cleanup deterministic and safe
- avoid payload-key reserialization on every chunk for the same request object
- stop ambiguous equivalent `format_stream_end()` calls from guessing which live stream to retire
- fix the delayed completed-stream end race so a stale end call cannot tear down a newer live parser with the same payload

## Files in scope

- `omlx/api/thinking.py`
- `omlx/api/adapters/openai.py`
- `omlx/api/adapters/anthropic.py`
- `tests/test_thinking.py`
- `tests/test_openai_adapter.py`
- `tests/test_anthropic_adapter.py`

## Stream-state contract

For `format_stream_end()`:
- exact live request object: retire that stream
- equivalent request object with exactly one live matching payload stream: retire that one stream
- equivalent request object with multiple live matching payload streams: do not guess; emit `[DONE]` without retiring any live parser state
- delayed equivalent end for a completed stream must not tear down a newer live parser that happens to share the same payload

## Tests

The branch adds and hardens coverage for:
- no-tag exact round-trip and literal stray `</think>` preservation in visible content
- OpenAI streaming think-tag splitting across chunks
- interleaving isolation for same-payload concurrent streams
- state-key collision safety
- equivalent-request cleanup behavior
- ambiguous equivalent-end no-guess behavior
- delayed completed-stream end race
- Anthropic/OpenAI adapter parity for thinking extraction where applicable

## Notes

This is a reasoning/thinking-parsing hardening slice. It does not change the broader tool-streaming work and it does not rely on Python object identity as a public logical-stream contract beyond the exact-object end fast path.

### DIFF
--- a/omlx/api/adapters/openai.py
+++ b/omlx/api/adapters/openai.py
@@ -142,7 +142,7 @@ class OpenAIAdapter(BaseAdapter):
         )
         return parser, request_key
 
-    def _clear_ended_payload_keys(self, payload_key: str) -> None:
+    def _clear_ended_payload_keys(self, payload_key: str) -> bool:
         stale_request_ids = [
             request_object_id
             for request_object_id, ended_payload_key in self._ended_request_payload_keys.items()
@@ -150,6 +150,7 @@ class OpenAIAdapter(BaseAdapter):
         ]
         for request_object_id in stale_request_ids:
             self._ended_request_payload_keys.pop(request_object_id, None)
+        return bool(stale_request_ids)
 
     @property
     def name(self) -> str:
@@ -386,9 +387,14 @@ class OpenAIAdapter(BaseAdapter):
         payload_key = self._ended_request_payload_keys.pop(request_object_id, None)
         if payload_key is None:
             payload_key = self._stream_payload_key(request)
-        self._clear_ended_payload_keys(payload_key)
+        cleared_completed_stream = self._clear_ended_payload_keys(payload_key)
         active_keys = self._active_request_keys_by_payload.get(payload_key)
         if active_keys:
+            # If we just cleared delayed completion bookkeeping for this
+            # payload, do not tear down a newer live parser that happens to
+            # share the same request body.
+            if cleared_completed_stream:
+                return "data: [DONE]\n\n"
             # An equivalent-but-different request object cannot disambiguate
             # between multiple live streams with identical payloads.
             if len(active_keys) > 1:
--- a/tests/test_openai_adapter.py
+++ b/tests/test_openai_adapter.py
@@ -1175,6 +1175,52 @@ class TestOpenAIAdapter:
 
         assert adapter._ended_request_payload_keys == {}
 
+    def test_delayed_equivalent_stream_end_for_completed_stream_does_not_remove_newer_live_parser(self, adapter):
+        """Delayed end for a completed stream must not tear down a newer live parser with same payload."""
+        request_a = ChatCompletionRequest(
+            model="test-model",
+            messages=[Message(role="user", content="Hello")],
+        )
+        request_a_end = ChatCompletionRequest(
+            model="test-model",
+            messages=[Message(role="user", content="Hello")],
+        )
+        request_b = ChatCompletionRequest(
+            model="test-model",
+            messages=[Message(role="user", content="Hello")],
+        )
+
+        key_b = adapter._stream_state_key(request_b)
+
+        adapter.format_stream_chunk(
+            StreamChunk(text="<think>A</think>done", is_first=True, is_last=True),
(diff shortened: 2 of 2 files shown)


lyonsno:
Follow-up note on the latest adapter fix (`70a7cf0`):

- `format_stream_end()` equivalent-object cleanup no longer uses least-recently-touched selection when multiple live streams share the same payload.
- The adapter now uses a deterministic logical-stream policy based on preserved stream birth order.
- This is intentional: an end call that arrives with a fresh equivalent request object does not carry enough identity to perfectly recover the originating live stream when multiple same-payload streams are active.
- The branch now locks in the chosen policy with two targeted tests in `tests/test_openai_adapter.py`:
  - `test_format_stream_end_equivalent_request_does_not_remove_other_stream_when_ending_stream_was_touched_last`
  - `test_stream_payload_key_is_not_recomputed_for_every_chunk`
- The second test also constrains payload-key reuse so we do not reserialize the full request on every chunk for the same request object.

If maintainers want a different ambiguity policy for multi-live equivalent payloads, that would be a contract decision on top of this slice rather than an unintentional behavior change.

lyonsno:
Superseding my earlier note about birth-order cleanup: the branch no longer guesses a target stream for `format_stream_end()` when the call arrives as an equivalent-but-different request object and multiple live streams share the same payload.

Current contract in `d6a7a7a`:
- exact live request object: clean up that stream
- equivalent request object with exactly one live matching payload stream: clean up that one stream
- equivalent request object with multiple live matching payload streams: do not guess; emit `[DONE]` without retiring any live parser state
- ended-payload bookkeeping is now cleared by payload so equivalent-object end calls do not leak `_ended_request_payload_keys`

The adapter tests were updated to lock in that supported policy rather than requiring impossible recovery of true stream identity from payload-equal fresh request objects.

lyonsno:
One last housekeeping note for the current head (`2b6aa8d`): this branch now also covers the delayed completed-stream end race.

Concretely: if stream A completes, stream B later starts with the same payload, and a delayed equivalent `format_stream_end()` for A arrives afterward, that delayed end must not tear down B's live parser state.

That behavior is now part of the branch contract alongside the earlier no-guess rule for ambiguous equivalent-object end calls.

jundot:
Hey, thanks for the PR. I took a closer look at the changes and want to share some observations before this moves forward.

**Streaming ThinkingParser already exists in server.py**

`server.py:1843` already creates a local `ThinkingParser()` per streaming request and feeds chunks through it at line 1877. Because the parser is a local variable inside the generator function, each concurrent stream gets its own isolated instance automatically. The chunk-boundary tag splitting case (`<thi` + `nk>`) is also handled by `ThinkingParser._could_be_tag()` with buffering (`thinking.py:137-141`).

The OpenAI non-streaming path also already calls `extract_thinking()` in the adapter (`openai.py:111`).

**Adapter-level duplication creates its own problems**

The `_stream_parsers` dict, payload-key caching, and 3-tier cleanup logic (exact object / single equivalent / ambiguous equivalent) are all needed to manage state that only exists because think-tag parsing is re-introduced at the adapter layer. If the adapter doesn't re-parse, none of that state management is necessary in the first place.

**The actual gap i see**

The Anthropic adapter `format_response()` passes `response.text` straight through without separating thinking content. It could call `extract_thinking()` and forward `thinking=` to `convert_internal_to_anthropic_response()`, which already accepts that parameter.

**Suggestion**

Maybe it makes sense to scope this down to the Anthropic adapter fix and drop the adapter-level streaming parser? Curious to hear your thoughts, i might be missing context on why the server-level parser isn't sufficient for the cases you're targeting.

lyonsno:
I think I’m actually gonna be closing this out. I re-reviewed the branch against current main and I don’t think this is the right change set to merge.

I overfit on branch-local parser behavior and on adapter machinery that doesn’t line up cleanly with the live streaming path. In particular, I let extract_thinking() drift away from the compatibility baseline that main already
preserves for loose think-tag text.

Rather than keep pushing on a confused branch, I’m going to withdraw it and come back with a smaller re-slice if there’s still a concrete bug left on top of main.

Thanks again for the careful review.

PR #209, opened 2026-03-13
Add support for ParoQuant and custom quantization method loading

liang2kl:
**PR Summary**

I would like to bring support for [ParoQuant](https://github.com/z-lab/paroquant) to omlx to allow loading ParoQuant [models](https://huggingface.co/collections/z-lab/paroquant) directly.

ParoQuant is a newly proposed quantization method that achieves much higher accuracy than current state of the arts (see https://github.com/ml-explore/mlx-lm/issues/977#issuecomment-4053248252), while introducing very small overhead. It is also very friendly to `mlx-lm`, only needing to apply an additional lightweight transform before executing the mlx-native quantized matrix multiplication. The official ParoQuant repository already contains a custom model loader for mlx, so the required modifications are minimal.

To support ParoQuant and potentially other non-standard quantization methods, this PR refactors model loading into a future-friendly custom quantization loading flow, with minimal changes to the original codebase.

### What changed
- Added a centralized model loader that:
  - detects `quantization_config.quant_method` from local `config.json`
  - dispatches supported custom quantization formats (currently `paroquant`)
  - falls back to standard `mlx-lm` / `mlx-vlm` loaders when no custom format is detected
- Updated core load call sites to use the shared loader:
  - batched LLM engine
  - VLM engine
  - LLM model wrapper
  - causal-LM reranker loader

### Why
- Enables ParoQuant models without invasive engine changes.
- Creates a clean path to add more custom quantization formats later.
- Keeps external API small and maintainable.

Preliminary benchmarks of `z-lab/Qwen3.5-4B-PARO` on my 16G M4, compared with `mlx-community/Qwen3.5-4B-4bit`:

Screenshot 2026-03-13 at 5 39 05 PM

### DIFF
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -13,6 +13,7 @@ from typing import Any
 
 from ..api.tool_calling import convert_tools_for_template
 from ..api.utils import clean_special_tokens
+from ..utils.model_loading import load_text_model
 from ..utils.tokenizer import get_tokenizer_config
 from .base import BaseEngine, GenerationOutput
 
@@ -124,8 +125,6 @@ class BatchedEngine(BaseEngine):
 
         import asyncio
 
-        from mlx_lm import load
-
         from ..engine_core import AsyncEngineCore, EngineConfig
         from ..scheduler import SchedulerConfig
 
@@ -140,7 +139,7 @@ class BatchedEngine(BaseEngine):
         from ..engine_core import get_mlx_executor
 
         def _load_model_sync():
-            return load(
+            return load_text_model(
                 self._model_name,
                 tokenizer_config=tokenizer_config,
             )
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -205,7 +205,10 @@ class BatchedEngine(BaseEngine):
 
         from ..engine_core import AsyncEngineCore, EngineConfig
         from ..scheduler import SchedulerConfig
-        from ..utils.model_loading import maybe_apply_pre_load_patches
+        from ..utils.model_loading import (
+            maybe_load_custom_quantization,
+            maybe_apply_pre_load_patches,
+        )
 
         # Build tokenizer config with model-specific fixes
         tokenizer_config = get_tokenizer_config(
@@ -226,6 +229,16 @@ class BatchedEngine(BaseEngine):
         from ..engine_core import get_mlx_executor
 
         def _load_model_sync():
+            custom_loaded = maybe_load_custom_quantization(
+                self._model_name,
+                tokenizer_config=tokenizer_config,
+                trust_remote_code=self._trust_remote_code,
+                is_vlm=False,
+            )
+            if custom_loaded is not None:
+                model, processor = custom_loaded
+                return model, getattr(processor, "tokenizer", processor)
+
             return load(
                 self._model_name,
                 tokenizer_config=tokenizer_config,
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -38,7 +38,7 @@ from ..utils.image import (
     compute_image_hash,
     extract_images_from_messages,
 )
-from ..utils.tokenizer import get_tokenizer_config
+from ..utils.model_loading import load_vlm_model
 from .base import BaseEngine, GenerationOutput
 
 logger = logging.getLogger(__name__)
@@ -189,8 +189,6 @@ class VLMBatchedEngine(BaseEngine):
         if self._loaded:
             return
 
-        from mlx_vlm.utils import load as vlm_load
-
         from ..engine_core import AsyncEngineCore, EngineConfig
         from ..scheduler import SchedulerConfig
 
@@ -203,7 +201,7 @@ class VLMBatchedEngine(BaseEngine):
             # when torchvision is not available (extractors is None, `in` fails).
             # oMLX does not support video input, so we skip video processing.
             _patch_video_processor_bug()
-            return vlm_load(self._model_name)
+            return load_vlm_model(self._model_name)
 
         loop = asyncio.get_running_loop()
         self._vlm_model, self._processor = await loop.run_in_executor(
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -442,6 +442,7 @@ class VLMBatchedEngine(BaseEngine):
 
         from ..engine_core import AsyncEngineCore, EngineConfig
         from ..scheduler import SchedulerConfig
+        from ..utils.model_loading import maybe_load_custom_quantization
 
         # Load VLM model on the global MLX executor to avoid blocking the event loop
         # while ensuring no concurrent Metal operations. See issue #85.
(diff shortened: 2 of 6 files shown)


jundot:
Really interesting project. To be honest, i've been skeptical about mlx's quantization quality, so paro's mlx support feels meaningful. Personally i'd love to test at least qwen3.5 122b or larger, but it seems like quantization is only possible on CUDA for now. Really wish that wasn't the case. (Providing quantized versions of popular mid-to-large scale models would really help with adoption.)

As for the code, it looks generally solid. One minor note: when the paroquant loading path is taken, `tokenizer_config` from the caller is silently dropped. It doesn't cause real issues right now since the config only carries `trust_remote_code` (which defaults to false anyway), but it'd be better to pass it through to stay consistent with how omlx handles tokenizer setup.

Overall this looks good to me. Since there aren't many paro models available yet, i'd like to do more testing once additional models are uploaded before merging. Just give me a little time!

liang2kl:
Thanks! We are currently quantizing larger models like Qwen3.5-27B and will work on larger MoEs later. We are also exploring ways to make quantization feasible on edge devices; as of now the computation resources required for calibration are prohibitive. I'll let you know once the models are ready!

414141x0:
@jundot any updates on ParoQuant support?

liang2kl:
Hi @jundot I have updated the PR, using a less-intrusive method to load the custom quantization models so that it adapts better to the updated codebase since my last push.

We also released several larger models, for example:

- [z-lab/gemma-4-31B-it-PARO](https://huggingface.co/z-lab/gemma-4-31B-it-PARO)
- [z-lab/Qwen3.6-27B-PARO](https://huggingface.co/z-lab/Qwen3.6-27B-PARO)
- [z-lab/Qwen3.5-27B-PARO](https://huggingface.co/z-lab/Qwen3.5-27B-PARO)

Our internal tests show great throughput results. Feel free to test them out!

beamivalice:
I test your PR with 0.3.9dev1, can confirm that the speed was great. Though 3.6-27B-PARO (18.4GB) is 3GB larger than oQ4-FP16, it performs almost at the same level on my M1 Max.

Screenshot 2569-05-08 at 07 34 01

Find out

* Model loading is very slow
* Inference speed is quite fast, just ~10% slower compared to oQ4

Qwen3.6-27B-PARO

+0.0s Starting
+0.5s Loading model
+139.7s Generating
+164.9s Thinking
+339.9s Thought for 175.0s
+415.6s 3354t prompt · 4357t generated · prefill 83 tok/s · gen 17 tok/s · ttft 40.31s

415.56s total

Qwen3.6-27B-oQ4-FP16

 +0.0s Starting
+10.2s Loading model
+15.2s Generating
+39.3s Thinking
+156.9s Thought for 117.6s
+216.1s 3354t prompt · 3404t generated · prefill 139 tok/s · gen 19 tok/s · ttft 24.13s

216.09s total

---

Will it support any kind of MTP?

jundot:
Thanks for the iteration. The dispatcher slot is clean and slots in nicely next to the existing pre-load patch hook, and @beamivalice's bench on 3.6-27B-PARO matches expectations.

Going to merge as-is and clean up a couple of things in a follow-up on next release: route the remaining call sites through the dispatcher, and gate the experimental toggles (MTP / SpecPrefill / IndexCache etc.) for paroquant models in the admin UI since compatibility isn't verified yet. No action needed from you.

deepsweet:
@liang2kl @jundot ParoQuant really delivers!

See my ["Evaluation of various MLX quantizations"](https://github.com/deepsweet/mlx-eval/tree/main/results) for more details.

Qwen3.6-35B-A3B is complete, but 27B is a work-in-progress, I’ll add more quants over time. Couldn't resist of comparing PARO against its neighbours first:

Qwen3.6-27B KLD/RAM chart

beamivalice:
@deepsweet This is pretty wild, I need to try PARO out. I hope you give JANG a look too. Their models keep popping up everyday. Thank you.

Quick question, I went in to read your log about qwen3.6-27b txt about the test and PARO logged only 13.xGB ram usage. But the only model in HuggingFace was like 17.5GB and it showed 18.36GB loaded in my oMLX dashboard. Why your ram usage is far smaller than mine or even the model loaded?

deepsweet:
@beamivalice right now I measure the real active memory allocated for the model's weights in text-only mode, without the Vision Language. PARO with VL takes 14.28 according to mlx.core, but if I update all other quants to account VL into RAM usage, the relative differences will stay the same as it's now within extra ~1GB for VL.

However, ParoQuant:

>The visual components in this checkpoint is stored in original precision, and only the language components are quantized to 4 bits; as a result, the model size is larger than a fully-quantized model. Avoid loading the VLM components if you are not using the multimodal features for the best efficiency.

This is a bit suspicious. Like it has 4GB of extra VL stuff in original BF16, but it's not loaded in text-only mode.

liang2kl:
@deepsweet Great evaluation!

I think PARO models take more storage than expected on huggingface because in the checkpoint we don't quantize lm_head and embed_tokens so there will be more storage. But we loaded them in 4-bit so the runtime VRAM usage is smaller.

Actually I become skeptical about them being quantized to 4 bits since the majority of quants only quantize them to 8 bits, as they are more sensitive. Could you do an experiment with PARO + 8 bit lm_head and embed_tokens? If the accuracy gain is huge then we might consider modifying our loading code.

deepsweet:
@liang2kl sure thing, could you please provide such a model with 8-bit lm_head / embed_tokens?

liang2kl:
@deepsweet The model itself has high precision weights of them so it doesn't need a new model. But the loading script needs a [one-line](https://github.com/z-lab/paroquant/blob/main/paroquant/inference/backends/mlx/load.py#L354) change for this:

```diff
- mlx_nn.quantize(model, group_size, bits, mode="affine", class_predicate=_is_io_layer)
+ mlx_nn.quantize(model, 64, 8, mode="affine", class_predicate=_is_io_layer)
```

You can clone the repo, modify this, and install it using `pip install ".[mlx]"`.

Thanks!

deepsweet:
@liang2kl I've already [integrated](https://github.com/deepsweet/mlx-eval/commit/b94afe8ab9987e0838060cee1789b60bf8c24fbf) ParoQuant into mlx-eval, so it was quite straightforward.

Qwen3 6-27B

- -0,010487 nats Δ KLD
- -0,079159 Δ PPL
- +1,25 GiB Δ RAM

liang2kl:
@deepsweet Thanks for the update! Looks like the gain is relatively marginal. Let's keep the handling of these two layers as-is.

deepsweet:
@liang2kl hi, added [Qwen3.6-35B-A3B-PARO](https://huggingface.co/z-lab/Qwen3.6-35B-A3B-PARO).

See [results](https://github.com/deepsweet/mlx-eval/tree/main/results#qwen36-35b-a3b):

Qwen3.6-35B-A3B KLD/RAM chart

But the cold start time is kinda awful. Could you please explain what are the reasons of having these lm_head and embed_tokens quantized in _runtime_ on model loading?

```py
paro = config.get("quantization_config", {})
group_size = int(paro.get("group_size", 128))
bits = int(paro.get("bits", 4)
…
mlx_nn.quantize(model, group_size, bits, mode="affine", class_predicate=_is_io_layer)
```

I might not understand some very important point here.

PR #222, opened 2026-03-13
feat(admin): filter HF model search to MLX library only

JianShan-1214:
## Summary

* Admin model search called the HuggingFace API without a library filter, so results could include non–MLX models.
* Search now passes `filter="mlx"` so only MLX-compatible models are returned (aligned with [huggingface.co/models?library=mlx](https://huggingface.co/models?library=mlx)).
* Docstring updated to describe this behavior.

## Files changed

| File | Change |
|------|--------|
| `omlx/admin/hf_downloader.py` | Add `filter="mlx"` to search request; document MLX-only results in docstring |

## Test plan

* `uv run pytest tests/test_hf_downloader.py -v` — 74 passed
* Manual: open Admin → Models, run a search and confirm results are MLX models only.

### DIFF
--- a/omlx/admin/hf_downloader.py
+++ b/omlx/admin/hf_downloader.py
@@ -244,6 +244,9 @@ class HFDownloader:
     ) -> dict:
         """Search HuggingFace models by query string.
 
+        Results are restricted to the MLX library so that only MLX-compatible
+        models are returned (same as https://huggingface.co/models?library=mlx).
+
         Args:
             query: Search query string.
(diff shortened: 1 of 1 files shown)


jundot:
Looks good, exactly what search_models needed. Non-mlx results were just noise since omlx can only load mlx models anyway.

I'll merge this and add a small follow-up commit to verify the filter param gets passed through in tests (the existing tests mock the return value but don't check `list_models` call args).

Thanks for the contribution!

PR #226, opened 2026-03-14
feat: add image upload support to chat interface

Leslie-SSS:
## Summary

Add image upload support to the oMLX chat interface for VLM/OCR model interactions.

### Features

- **Multiple upload methods**: Click button, drag & drop, or Ctrl/Cmd+V paste
- **Image-only messages**: Support sending images without text
- **Smart history handling**: Image requests don't include history context to avoid re-sending image data
- **Storage optimization**: Images not stored in localStorage to prevent quota exceeded

### Files Changed

| File | Changes |
|------|---------|
| `omlx/admin/templates/chat.html` | +347 lines (CSS, HTML, JS) |
| `omlx/admin/i18n/en.json` | +9 lines |
| `omlx/admin/i18n/zh.json` | +10 lines |
| `omlx/admin/i18n/zh-TW.json` | +9 lines |
| `omlx/admin/i18n/ja.json` | +8 lines |
| `omlx/admin/i18n/ko.json` | +8 lines |
| `tests/test_chat_image_upload.py` | +114 lines (new) |

### Commits

| Commit | Description |
|--------|-------------|
| `2cf7832` | feat: add image upload support to chat interface |
| `b4ed3b7` | fix: prevent localStorage quota exceeded when saving chat with images |
| `7630a3b` | fix: allow sending image without text message |
| `d497777` | fix: avoid re-sending historical image data in API requests |
| `1cb910c` | fix: send image-only requests without history context |
| `9fac80f` | fix: address code review issues in image upload |
| `848adae` | docs: update README with image upload feature and cleanup |
| `2e10994` | docs: update README with new chat screenshot showing image upload |

## Test Plan

1. Start oMLX server: `omlx serve --model-dir ~/models`
2. Load a VLM or OCR model (e.g., `LightOnOCR-2-1B`)
3. Navigate to `http://localhost:8000/admin/chat`
4. Test the following:
   - [ ] Click image button to upload
   - [ ] Drag & drop image to chat area
   - [ ] Ctrl/Cmd+V paste image
   - [ ] Send image-only (no text)
   - [ ] Send image + text
   - [ ] Continue text conversation after image (history context works)
   - [ ] Refresh page - historical images show placeholder
   - [ ] Dark/light theme toggle

### DIFF
--- a/omlx/admin/i18n/en.json
+++ b/omlx/admin/i18n/en.json
@@ -510,6 +510,7 @@
     "chat.default_image_prompt": "Describe this image",
     "chat.error.invalid_image_type": "Please select an image file",
     "chat.error.image_too_large": "Image size must be less than 10MB",
+    "chat.error.image_load_failed": "Failed to load image. Please try again.",
 
     "js.error.required_fields": "Required fields cannot be empty: {fields}",
     "js.error.api_key_min_length": "API key must be at least 4 characters",
--- a/omlx/admin/i18n/ja.json
+++ b/omlx/admin/i18n/ja.json
@@ -509,6 +509,8 @@
   "chat.default_image_prompt": "この画像を説明してください",
   "chat.error.invalid_image_type": "画像ファイルを選択してください",
   "chat.error.image_too_large": "画像サイズは10MB以下にしてください",
+  "chat.image_not_available": "画像",
+  "chat.error.image_load_failed": "画像の読み込みに失敗しました。もう一度お試しください。",
 
   "js.error.required_fields": "必須項目が空です: {fields}",
   "js.error.api_key_min_length": "APIキーは4文字以上必要です",
--- a/omlx/admin/i18n/ko.json
+++ b/omlx/admin/i18n/ko.json
@@ -509,6 +509,8 @@
   "chat.default_image_prompt": "이 이미지를 설명해주세요",
   "chat.error.invalid_image_type": "이미지 파일을 선택해주세요",
   "chat.error.image_too_large": "이미지 크기는 10MB 이하여야 합니다",
+  "chat.image_not_available": "이미지",
+  "chat.error.image_load_failed": "이미지를 불러오지 못했습니다. 다시 시도해 주세요.",
 
   "js.error.required_fields": "필수 항목이 비어 있습니다: {fields}",
   "js.error.api_key_min_length": "API 키는 4자 이상이어야 합니다",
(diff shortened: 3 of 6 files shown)


jundot:
Thanks for this. I've been spending all my time on oMLX backend improvements so this is a feature i knew was needed but couldn't get to. Really appreciate it. The upload UX covering click, drag & drop, and paste is well thought out, and the localStorage handling (stripping base64 before save, using `hasImage` flag for placeholders) was a smart call to avoid quota issues.

I went through the code in detail and found a few things that need fixing before merge.

### 1. Multi-turn image conversations are broken

When the latest message has an image, only that single message is sent to the API with no history. And when a text follow-up is sent, the else branch tries to include image data only for the "last user message":

```javascript
if (msg.imageData && index === lastUserMsgIndex) {
```

But `lastUserMsgIndex` always points to the current text-only message (since we're in the else branch where `lastMsg.imageData` is falsy). This condition can never be true, so all previous images get replaced with `"[Image]"` text.

This means a conversation like:
1. User sends photo: "what is this?" → model sees the image, responds
2. User follows up: "what color is it?" → model only sees `[Image]` string, loses all visual context

The image base64 is right there in `msg.imageData` in memory. It should be included for all messages that have it, not just the last one. Something like:

```javascript
messagesForApi = this.messages.map((msg) => {
    if (msg.imageData) {
        const content = [];
        if (msg.content) content.push({ type: "text", text: msg.content });
        content.push({ type: "image_url", image_url: { url: msg.imageData } });
        return { role: msg.role, content };
    }
    return { role: msg.role, content: msg.content };
});
```

Also the first branch that sends image messages without any history context should probably be reconsidered. If someone sends a second image mid-conversation, they'd lose all prior context.

### 2. `chat.image_not_available` i18n key missing in 3 languages

This key only exists in en.json and zh.json. It's missing from ja.json, ko.json, and zh-TW.json. The template uses it with a fallback (`|| 'Image'`) so it won't crash, but those users will see english text.

### 3. Hardcoded english in error handler

In `loadImageFile()`, the `reader.onerror` alert is hardcoded:
```javascript
alert('Failed to load image. Please try again.');
```
Should use an i18n key like the other error messages.

### 4. Unnecessary blob URL creation

`loadImageFile()` creates a blob URL via `URL.createObjectURL(file)` for preview, but `sendMessage()` ignores it and stores the base64 data URI in both `imagePreview` and `imageData`. The blob URL is created and then immediately thrown away (revoked in `clearPendingImage()`). Either use the blob URL properly or just drop it and use base64 for everything.

Let me know what you think after looking into these.

Leslie-SSS:
@jundot All issues have been addressed. Regarding the multi-turn image dialogue problem, the complete context is now preserved. It’s worth noting that in OCR model scenarios, images may be repeatedly recognized, which is why I initially discarded the image context. I’m unsure which approach is better—whether to retain all images or only the most recent one. I don’t know, so for now, the solution is to keep all of them.

Issue #246, opened 2026-03-15
The SSD cache doesn't seem to be working.


chattytak:
**Describe the bug**
The contents of the SSD cache directory are always empty, and according to the web dashboard, the cache size is 0 and the hit rate is always 0%.

**To Reproduce**
I encountered it while using it as usual.

**Expected behavior**
The cache works properly.

**Screenshots**
```
cat .omlx/settings.json
{
  "version": "1.0",
  "server": {
    "host": "x.x.x.x",
    "port": xxxxx,
    "log_level": "info",
    "cors_origins": [
      "*"
    ]
  },
  "model": {
    "model_dirs": [
      "/Users/tak/.omlx/models"
    ],
    "model_dir": "/Users/tak/.omlx/models",
    "max_model_memory": "disabled",
    "model_fallback": false
  },
  "memory": {
    "max_process_memory": "86%"
  },
  "scheduler": {
    "max_num_seqs": 8,
    "completion_batch_size": 8
  },
  "cache": {
    "enabled": true,
    "ssd_cache_dir": "/Users/tak/.omlx/cache",
    "ssd_cache_max_size": "46GB",
    "hot_cache_max_size": "0",
    "initial_cache_blocks": 256
  },
  "auth": {
    "api_key": "xxxxx",
    "secret_key": "xxxxx",
    "skip_api_key_verification": false,
    "sub_keys": []
  },
  "mcp": {
    "config_path": null
  },
  "huggingface": {
    "endpoint": ""
  },
  "modelscope": {
    "endpoint": ""
  },
  "sampling": {
    "max_context_window": 32768,
    "max_tokens": 32768,
    "temperature": 1.0,
    "top_p": 0.95,
    "top_k": 0,
    "repetition_penalty": 1.0
  },
  "logging": {
    "log_dir": null,
    "retention_days": 7
  },
  "claude_code": {
    "context_scaling_enabled": false,
    "target_context_size": 200000,
    "mode": "cloud",
    "opus_model": null,
    "sonnet_model": null,
    "haiku_model": null
  },
  "integrations": {
    "codex_model": null,
    "opencode_model": null,
    "openclaw_model": null,
    "openclaw_tools_profile": "coding"
  },
  "ui": {
    "language": "en"
  }
}
```

```
$ ls -l .omlx/cache/*
.omlx/cache/0:
total 0

.omlx/cache/1:
total 0

.omlx/cache/2:
total 0

.omlx/cache/3:
total 0

.omlx/cache/4:
[...]

PR #256, opened 2026-03-16
admin: clarify sub-block cache (<block_size) in runtime cache observability

yes999zc:
## Summary
This PR addresses the follow-up confusion in #246 where requests below one full block (for example `<1024` on ArraysCache models) appear as `0 indexed blocks` and look like cache is not working.

### What changed
- Expose prefix-cache observability from scheduler runtime stats (`partial_block_skips`, `last_partial_tokens_skipped`, etc.).
- Extend admin runtime cache payload with:
  - `has_sub_block_cache`
  - `indexed_blocks_display` (shows `<block_size` only when sub-block cache evidence exists)
  - partial block diagnostics fields
- Update dashboard Runtime Cache table:
  - `Indexed Blocks` shows `<block_size` instead of `0` when applicable
  - add `Sub-block Cache` column to indicate cached-but-not-SSD-indexed cases
  - add explanatory note to avoid interpreting this as no-cache

### Why
For models using larger effective block size (e.g. 1024), short requests can be cached in memory/paged cache but cannot form full SSD index blocks. Showing raw `0` is technically true for SSD indexed blocks but misleading for users.

### Behavior
- If `indexed_blocks == 0` **and** prefix-cache metrics indicate trailing partial blocks, dashboard shows `<block_size`.
- Otherwise dashboard keeps showing the numeric indexed block count.

## Tests
Ran targeted tests:
- `tests/test_admin_api_key.py::TestRuntimeCacheObservability::test_runtime_cache_ignores_single_model_stats_failure`
- `tests/test_admin_api_key.py::TestRuntimeCacheObservability::test_runtime_cache_marks_sub_block_cached_when_indexed_blocks_zero`
- `tests/test_admin_api_key.py::TestStatsSecurity::test_stats_response_does_not_include_raw_api_key`
- `tests/test_prefix_cache.py::TestArraysCacheLastBlockOnly::test_store_cache_arrayscache_partial_trailing_uses_last_full_block_state`
- `tests/test_prefix_cache.py::TestArraysCacheLastBlockOnly::test_store_cache_all_partial_creates_no_blocks`

All passed locally.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -17,9 +17,10 @@ import shutil
 import sys
 import time
 from collections import deque
+from dataclasses import asdict, is_dataclass
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Any, Dict, Optional, List
+from typing import Any, Dict, Optional, List, Literal
 
 from fastapi import APIRouter, Depends, HTTPException, Request, Response
 from fastapi.responses import FileResponse, HTMLResponse, JSONResponse, RedirectResponse
@@ -151,7 +152,7 @@ class GlobalSettingsRequest(BaseModel):
     integrations_codex_model: Optional[str] = None
     integrations_opencode_model: Optional[str] = None
     integrations_openclaw_model: Optional[str] = None
-    integrations_openclaw_tools_profile: Optional[str] = None
+    integrations_openclaw_tools_profile: Optional[Literal["minimal", "coding", "messaging", "full"]] = None
 
     # UI settings
     ui_language: Optional[str] = None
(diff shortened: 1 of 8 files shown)


jundot:
Thanks for the contribution! I made a couple of follow-up adjustments in c6f4c3a:

**API key restored in stats response**
The `api_key` field in `/api/stats` was needed for the Claude Code integration command (`ANTHROPIC_AUTH_TOKEN`) and other integration commands (`--api-key`). Without it, users with an api key configured couldn't get a working copy-paste command from the dashboard. Reverted the removal.

**Runtime cache table improvements**
- Removed the top-level "Effective block size(s)" summary since each model already shows its own block size in the table. Block size varies by cache type (ArraysCache: 2048, RotatingKVCache: 512~1024, KVCache: 512), so a combined display was misleading.
- Added units to avoid ambiguity: block size now shows `2048 tks`, indexed blocks shows `70 blocks`.

PR #263, opened 2026-03-16
fix(menubar): reuse admin session across stats fetches

ChiangJingYing:
## What

Currently `_fetch_stats()` in the menubar app creates a new
`requests.Session()` on every call, which means `POST /admin/api/login`
is fired every 5 seconds — one login request per stats cycle, indefinitely.

## Why it matters

- Produces unnecessary noise in the server access log (login appears
  alongside every stats request)
- Sends the API key over the wire more frequently than needed
- Creates and discards a session object on every cycle with no benefit

## Fix

Introduce `AdminStatsClient` (`packaging/omlx_app/admin_client.py`) that
persists the session cookie across `_fetch_stats()` calls:

- Logs in only when no session exists (first call, or after invalidation)
- On `401` (e.g. server restart changes the secret key), invalidates and
  retries login once automatically
- On `RequestException`, clears the session so the next cycle retries cleanly
- `invalidate()` is called explicitly when the server enters ERROR or
  UNRESPONSIVE state

`AdminStatsClient` is placed in its own module (`admin_client.py`) so it
can be unit-tested without the PyObjC dependency that makes `app.py`
untestable in a standard pytest environment.

## Tests

Added `TestAdminStatsClient` in `tests/test_omlx_app.py` (6 cases covering
first-login, session reuse, auto-relogin on `401`, login failure, connection
error session cleanup, and explicit `invalidate()`).

### Running the tests

> **Note:** `pytest -m "not slow"` currently fails with
> `ModuleNotFoundError: No module named 'omlx_app.admin_client'; 'omlx_app' is not a package`
> due to a conflict between the editable install's path hook and the
> `packaging/` source root. Use the explicit paths below instead:

```bash
# Run only the new AdminStatsClient unit tests
pytest tests/test_omlx_app.py -v

# Run all fast omlx_app tests (skip slow model-loading tests)
pytest tests/test_omlx_app.py tests/test_updater.py -v

### DIFF
--- a/packaging/omlx_app/admin_client.py
+++ b/packaging/omlx_app/admin_client.py
@@ -72,6 +72,10 @@ class AdminStatsClient:
             return resp.json()
         return None
 
+    def matches(self, base_url: str, api_key: str) -> bool:
+        """Return True if this client was created with the given credentials."""
+        return self._base_url == base_url and self._api_key == api_key
+
     def invalidate(self) -> None:
         """Discard the current session (e.g. server restarted)."""
         self._session = None
--- a/packaging/omlx_app/app.py
+++ b/packaging/omlx_app/app.py
@@ -739,15 +739,20 @@ class OMLXAppDelegate(NSObject):
     def _fetch_stats(self):
         """Fetch serving stats from the admin API."""
         api_key = self.config.get_server_api_key()
+        current_base_url = f"http://127.0.0.1:{self.config.port}"
+
         if not api_key:
             self._cached_stats = None
             self._cached_alltime_stats = None
             return
 
+        if self._stats_client is not None and not self._stats_client.matches(
+            current_base_url, api_key
+        ):
+            self._stats_client = None
+
         if self._stats_client is None:
-            self._stats_client = AdminStatsClient(
-                f"http://127.0.0.1:{self.config.port}", api_key
-            )
+            self._stats_client = AdminStatsClient(current_base_url, api_key)
 
         self._cached_stats = self._stats_client.fetch_stats()
         self._cached_alltime_stats = self._stats_client.fetch_stats(scope="alltime")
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by jundot** [comment]
Thanks for the PR, the idea of reusing the admin session is solid. Logging in every 5 seconds is definitely wasteful.

I found a couple issues though.

**`_stats_client` never gets recreated when api key or port changes**

`AdminStatsClient` stores `base_url` and `api_key` at construction time, and `_stats_client` is only created when it's `None`:

```python
if self._stats_client is None:
    self._stats_client = AdminStatsClient(
        f"http://127.0.0.1:{self.config.port}", api_key
    )
```

Nothing ever sets `_stats_client` back to `None`. So if the user changes the api key or port in preferences, the client keeps using the old values and stats fetching silently breaks until the app is restarted. `invalidate()` only clears the internal session, not the client itself.

The current code reads `self.config.get_server_api_key()` fresh on every call, so it doesn't have this problem.

A simple fix would be checking if the current api key / port differs from what the client was built with, and recreating it when they don't match.

**Wrong patch target in existing tests**

```python
# before (correct - tests config.update_server_api_key_runtime which lives in config.py)
@patch("omlx_app.config.requests.Session")

# after (wrong - config.py still imports requests directly, this patches the wrong module)
@patch("omlx_app.admin_client.requests.Session")
```

`test_update_server_api_key_runtime_success` and `test_update_server_api_key_runtime_server_down` test `config.update_server_api_key_runtime()`, which uses `requests.Session` from `config.py` (line 123). Since `config.py` wasn't modified in this PR, changing the patch target means the mock won't apply and the tests will try to make real http requests.

ChiangJingYing:
Thanks for the thorough review!

**Staleness check**

Added a `matches(base_url, api_key)` method to `AdminStatsClient` that compares against the private fields, then use it in `_fetch_stats` to nil out `_stats_client` whenever the credentials or port have changed:

```python
if self._stats_client is not None and not self._stats_client.matches(
    current_base_url, api_key
):
    self._stats_client = None
```

Also extracted `current_base_url` to a local variable so it's computed once and reused for both the check and the constructor call.

**Patch target**

Reverted both `@patch` decorators back to `omlx_app.config.requests.Session` — my mistake for changing those in the first place.

PR #267, opened 2026-03-17
fix: anthropic path reasoning

lyonsno:
## Summary

This PR fixes the Anthropic path when the engine provides separate
`reasoning_content`, covering both non-streaming `/v1/messages` and Anthropic
streaming.

## What it fixes

- surfaces explicit `reasoning_content` as Anthropic `thinking`
- preserves current `extract_thinking(output.text)` behavior from `main`
- recovers tool calls when markup appears in `text`, `reasoning_content`, or both
- avoids blanking richer text-derived thinking when `reasoning_content` contains
  only tool-call markup
- preserves answer suffixes after `</think>` when tagged `reasoning_content`
  carries the more complete visible answer
- avoids duplicate tool recovery when the same tool call appears in both sources
- fixes streaming handoff/cumulative-update cases so Anthropic SSE does not:
  - duplicate already-sent text
  - replay cumulative content on later chunks
  - mis-handle hidden tool-call envelopes in cumulative visible text

## Files changed

- `omlx/server.py`
- `tests/integration/test_server_endpoints.py`

## Scope

In scope:
- Anthropic non-streaming `/v1/messages`
- Anthropic streaming behavior affected by `reasoning_content`
- integration coverage for those behaviors

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -2293,12 +2293,19 @@ def _anthropic_reasoning_sources(
 
     display_thinking = visible_explicit_reasoning or text_thinking
 
-    tool_recovery_parts = []
-    if text_thinking:
-        tool_recovery_parts.append(text_thinking)
-    if reasoning_tool_source and reasoning_tool_source not in tool_recovery_parts:
-        tool_recovery_parts.append(reasoning_tool_source)
-    tool_recovery_thinking = "\n".join(tool_recovery_parts).strip()
+    tool_recovery_thinking = text_thinking.strip()
+    if reasoning_tool_source:
+        reasoning_tool_source = reasoning_tool_source.strip()
+        if not tool_recovery_thinking:
+            tool_recovery_thinking = reasoning_tool_source
+        elif reasoning_tool_source in tool_recovery_thinking:
+            pass
+        elif tool_recovery_thinking in reasoning_tool_source:
+            tool_recovery_thinking = reasoning_tool_source
+        else:
+            tool_recovery_thinking = (
+                f"{tool_recovery_thinking}\n{reasoning_tool_source}"
+            ).strip()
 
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for putting this together, but I don't think this should be merged right now.

`GenerationOutput` (`engine/base.py`) doesn't have a `reasoning_content` field and no engine sets one. `getattr(output, "reasoning_content", None)` will always return `None`, so all the new code paths are dead code. The existing `ThinkingParser` + `extract_thinking()` already handles `<think>` tag separation correctly for models like DeepSeek-R1.

If mlx-lm eventually provides `reasoning_content` as a separate field, the `GenerationOutput` schema change, engine update, and server handling should all land together so it can be tested end-to-end with real outputs, not `SimpleNamespace` mocks.

Adding 178 lines to `server.py` and 1000+ lines of tests for a code path that never executes isn't zero technical debt. It will cause merge conflicts and confusion next time someone touches the streaming logic.

The PR does fix a real bug where `thinking_block_started` wasn't being reset to `False` after closing the thinking block. I'll fix that separately.

A few smaller notes if this gets revised later:
- The handoff test expects `[text, thinking, text]` block ordering, but Anthropic normally has thinking before text.
- `_anthropic_reasoning_sources()` sanitizes content, then the caller sanitizes again.
- `_incremental_suffix()` returns full `current` when it doesn't start with `previous`, risking content replay.
- Every test repeats 30+ lines of SSE parsing boilerplate. A shared helper would help.

Happy to revisit when there's an engine change that actually produces `reasoning_content`.

lyonsno:
You're right — this should wait until GenerationOutput actually carries the field. Closing. Thanks for the reset bug catch, glad you that up separately.

PR #268, opened 2026-03-17
fix: chat image viewer modal and edit image preservation (#264)

### LINKED ISSUE #264: Chat unable to see attached images
Coming from lm studio to this, I see x2 speed improvement with an m3 pro 36gb ram. I am, however, facing what looks to be a bug. 

**Describe the bug**
Chatting to ocr a receipt. Reads first image fine. But even here, if you click on the image to enlarge, new tab pops up with nothing in it. Asked to read a second image and it went back to analyzing the first image.  Attached the image again but it says it cant find the image. It gets stuck in a thinking loop, continuously looking for the said image. 

**To Reproduce**
Upload a receipt image. Ask it to read item, price, sub total, tax and the total. Ask it the second time with a different image. It either cant see the new image or it refers to the first. 

**Expected behavior**
Able to find and analyze the image attached to the said chat instance. 

**Screenshots**

 - OS: tahoe
 - Browser chrome
 - Version unknown as I'm at work but will update later. 

Also, is there a way to turn off the thinking? Just with /no_thinking?

Thank you.

Leslie-SSS:
## Summary

- Fixed the issue where clicking to enlarge an image opened a new tab with no content (using a modal instead of `window.open()`)
- Fixed the issue where images were lost when editing a message containing images (retain original images when saving edits)

## Test plan

- [ ] After uploading an image, click on the image to confirm the modal displays correctly
- [ ] Edit a user message containing images and confirm the images still exist after saving
- [ ] Test both base64 images and URL images

Fixes #264

### DIFF
--- a/omlx/admin/templates/chat.html
+++ b/omlx/admin/templates/chat.html
@@ -266,6 +266,65 @@
         .sidebar-overlay.hidden { display: none; }
         .user-edit-btn { left: -28px; }
     }
+
+    /* Image modal */
+    .image-modal {
+        position: fixed;
+        inset: 0;
+        background: rgba(0, 0, 0, 0.9);
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        z-index: 9999;
+        cursor: zoom-out;
+    }
+    .image-modal img {
+        max-width: 90%;
+        max-height: 90%;
+        object-fit: contain;
+        border-radius: 8px;
+        box-shadow: 0 4px 24px rgba(0, 0, 0, 0.5);
+    }
+    .image-modal-close {
+        position: absolute;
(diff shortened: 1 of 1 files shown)


jundot:
Nice work on this! the modal approach is way better than window.open() for base64 images, and the edit image preservation logic is solid. Clean implementation overall, thanks for the contribution 👍

A couple of small things i noticed:
- `chat.image_preview` i18n key is used in the edit preview alt text but doesn't exist in any language file yet (the `|| 'Image'` fallback covers it at runtime so nothing breaks)
- No tests for the new edit+image preservation logic

I'll add both in a follow-up commit after merging, so no changes needed on your end.

PR #283, opened 2026-03-17
Fix compile-on crash for embeddings/reranker by compiling primitive-output paths

### LINKED ISSUE #282: [Bug] compile-on crash in /v1/embeddings and /v1/rerank (array without primitive)
# [Bug] compile-on crash in /v1/embeddings and /v1/rerank (`array without primitive`)

## Summary

On `v0.2.16`, both embedding and reranker can crash under compile-on path with:

```text
RuntimeError: [eval] Attempting to eval an array without a primitive.
```

## Affected routes

- `POST /v1/embeddings`
- `POST /v1/rerank` (SequenceClassification reranker path)

## Environment

- oMLX: `0.2.16`
- mlx-lm: `0.31.2` (commit `564281f`)
- mlx-embeddings: `0.0.5` (commit `6e2ef52`)
- Apple Silicon macOS

## Repro

```bash
curl -s -X POST https://api.example.com/v1/embeddings \
  -H 'Authorization: Bearer ${API_KEY}' \
  -d '{"model":"mxbai-embed-large-v1","input":"hello world"}'

curl -s -X POST https://api.example.com/v1/rerank \
  -H 'Authorization: Bearer ${API_KEY}' \
  -d '{"model":"bge-reranker-v2-m3","query":"hello","documents":["hello","world"]}'
```

## Root cause hypothesis

Current compile path wraps `model.__call__` directly, which returns output containers (e.g., `BaseModelOutput`).
In embedding/reranker pipelines this can produce arrays that fail `mx.eval(...)` with `array without primitive`.

## Fix that worked locally

Instead of compiling `model.__call__`, compile a narrow primitive-output function:

- Embedding: `inputs -> embedding tensor`
- Reranker: `inputs -> pooler_output logits`

This preserved compile mode and removed crashes for both routes.

## Patch

I have local patch files and can open PR:

- `0001-embedding-compile-primitive-output-fix.patch`
- `0002-reranker-compile-primitive-logits-fix.patch`

## Regression checks

- `/v1/embeddings` success
- `/v1/rerank` success
- OpenClaw memory search (embedding caller) recovered

yes999zc:
# PR Title
Fix compile-on crash for embeddings/reranker by compiling primitive-output paths

## Problem
`/v1/embeddings` and `/v1/rerank` can fail with:

```
RuntimeError: [eval] Attempting to eval an array without a primitive.
```

The crash appears when `mx.compile` wraps `model.__call__` directly and the output is a container object (e.g., `BaseModelOutput`).

## Root Cause
Current compile wrapping point is too broad:
- compiled target: `model.__call__`
- downstream eval target: fields extracted from output containers

For embedding/reranker flows this can produce arrays that fail `mx.eval(...)` in runtime.

## Fix
Compile narrower functions that return primitive tensors directly.

### Embedding (`omlx/models/embedding.py`)
- compile path changed from direct `model.__call__` wrapper
- new compiled function returns final embedding tensor only
  - `text_embeds` / `pooler_output` / mean pooled `last_hidden_state`
- use default `mx.compile(...)` mode (no `shapeless=True`)
- keep eager fallback if compile path fails

### Reranker (`omlx/models/reranker.py`)
- for SequenceClassification rerankers:
  - compile function returns `pooler_output` logits directly
  - `_rerank_seq_classification()` uses compiled logits path
- causal-lm reranker path unchanged

## Why this works
By compiling at the primitive-output stage, we avoid carrying output-container semantics through the compiled boundary and remove the `array without primitive` eval failure.

## Validation
- `/v1/embeddings` returns 200 (tested with `mxbai-embed-large-v1`, `bge-small-en-v1.5`)
- `/v1/rerank` returns 200 (tested with `bge-reranker-v2-m3`)
- OpenClaw `memory_search` recovered (depends on embeddings route)

## Notes
This PR preserves compile optimization intent (no global compile disable) while fixing runtime stability.

## Test Plan
1. Start server on Apple Silicon (same config as report).
2. Call `/v1/embeddings` repeatedly and after idle.
3. Call `/v1/rerank` repeatedly and after idle.
4. Confirm no 500 and no `array without primitive` stack trace in logs.

Fixes #282

### DIFF
--- a/omlx/models/embedding.py
+++ b/omlx/models/embedding.py
@@ -7,30 +7,12 @@ text embeddings using Apple's MLX framework.
 """
 
 import logging
-from dataclasses import dataclass, field
-from typing import Any, List, Optional
+from dataclasses import dataclass
+from typing import List, Optional
 
 import mlx.core as mx
 
 logger = logging.getLogger(__name__)
-
-
-class _CompiledForward:
-    """Wraps an MLX module to route __call__ through mx.compile.
-
-    Keeps compiled Metal pipeline states alive in memory, preventing
-    kernel eviction after idle periods (~3s on macOS).
-    """
-
-    def __init__(self, module):
-        self._module = module
-        self._compiled = mx.compile(module.__call__, shapeless=True)
-
-    def __call__(self, *args, **kwargs):
-        return self._compiled(*args, **kwargs)
-
-    def __getattr__(self, name):
-        return getattr(self._module, name)
 
 
 @dataclass
@@ -73,6 +55,8 @@ class MLXEmbeddingModel:
         self.processor = None
         self._loaded = False
         self._hidden_size: Optional[int] = None
+        self._is_compiled = False
+        self._compiled_embed = None
 
     def load(self) -> None:
         """Load the model and processor/tokenizer."""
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the quick catch and fix. The root cause analysis is spot on and the primitive-output compile approach is the right call.

I found a couple of small things i want to clean up:

1. **Reranker compiled path missing fallback**: `embedding.py` has a try/except that falls back to eager `generate()` if the compiled path fails, but `_rerank_seq_classification` doesn't have the same safety net. If the compiled function fails at runtime it'll crash instead of falling back.

2. **Dead `_CompiledForward` class in reranker.py**: It got removed from `embedding.py` but is still sitting in `reranker.py` unused.

3. **No warmup validation in `_try_compile`**: `mx.compile()` is lazy so `_is_compiled` gets set to True before the function actually runs. The old code did a dummy forward pass to verify. Not critical for embedding (has fallback) but combined with #1 it matters for the reranker.

Nothing blocking the merge. I'll squash merge this now and do a follow-up commit for the above.

PR #304, opened 2026-03-18
fix(test): restore sys.modules polluted by test_find_matching_dmg

blightbow:
_get_find_matching_dmg() (introduced in 090cfea) replaces omlx_app, omlx_app.config, and omlx_app.server_manager in sys.modules with MagicMock objects but never saves their original values for the finally block to restore. This leaks MagicMock modules into every subsequent test, causing 76 failures in test_omlx_app.py when run after test_find_matching_dmg.

Save the omlx_app.* module references before popping them so the existing finally-block restoration logic covers them.

### DIFF
--- a/tests/test_find_matching_dmg.py
+++ b/tests/test_find_matching_dmg.py
@@ -51,6 +51,11 @@ def _get_find_matching_dmg():
     sys.modules["omlx._version"] = version_mod
 
     try:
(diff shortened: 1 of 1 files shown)


blightbow:
Retracting, this was addressed in 685a5bfc25b335ddd6b5098d8b89e2986496dc60.

PR #305, opened 2026-03-18
fix(build): add PEP 735 dependency-groups for uv sync --dev

blightbow:
Dev dependencies (pytest, pytest-asyncio, etc.) were only defined under [project.optional-dependencies] which requires `uv sync --extra dev` or `uv pip install -e ".[dev]"`. The modern uv convention uses [dependency-groups] (PEP 735) so that `uv sync --dev` will install them. Without this, `uv sync` actively uninstalls pytest and pytest-asyncio, causing all async tests to fail with "async def functions are not natively supported". 

The existing definition is preserved under [project.optional-dependencies] for backward compatibility, but will need to be separately maintained.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,6 +78,19 @@ dev = [
     "mypy>=1.0.0",
     "mcp>=1.0.0",
 ]
+# PEP 735 dependency groups — consumed by `uv sync --dev`.
+# Keep in sync with [project.optional-dependencies] dev above
+# (pip/legacy toolchains use that section instead).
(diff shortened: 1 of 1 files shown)


jundot:
Looks good to me. This only touches pyproject.toml metadata and doesn't affect the build output, code signing, or notarization at all since dev dependencies never get bundled into the app.

One minor concern is the duplicated list between [project.optional-dependencies] and [dependency-groups], but the sync reminder comment is already there so it should be fine for now.

PR #306, opened 2026-03-18
feat: add partial mode for assistant message prefill

blightbow:
This PR implements support for Moonshot's "partial mode", an extension to the OpenAI chat completion API that enables message prefills and renaming of role anchors. This enables prefill patterns (JSON mode via `"content": "{"`) and named-assistant persona consistency (Kimi K2/K2.5 `name` field rendering).

Full disclosure: Moonshot has not publicly documented how "partial mode" is implemented on their API backend, but when you compare [their documentation](https://platform.moonshot.ai/docs/api/partial) to the model templates for [K2](https://huggingface.co/moonshotai/Kimi-K2-Instruct/blob/main/chat_template.jinja) and [K2.5](https://huggingface.co/moonshotai/Kimi-K2.5/blob/main/chat_template.jinja), it becomes pretty obvious what they're doing here.

The underlying infrastructure is already present in mlx_lm and the chat templates of Kimi K2 and K2.5. This change allows the `name` field to be passed through to the backend (in all circumstances), and reserves the `partial` field as a boolean toggle for switching between `continue_final_message=True` and `add_generation_prompt=True` in apply_chat_template. `partial` is consumed by the API server and not passed to the model template, unlike `name`.

For more information on Partial Mode, refer to the following:
- Moonshot's Partial Mode documentation: https://platform.moonshot.ai/docs/api/partial
- My PR previously submitted to mlx-openai-server: https://github.com/cubist38/mlx-openai-server/pull/213

---

Schema changes:
- Add `name` and `partial` fields to Message model
- Add `detect_and_strip_partial()` helper in api/utils.py
- Preserve `name`/`partial` through extract_text_content() and extract_multimodal_content() message extraction

Engine changes:
- BatchedEngine._apply_chat_template(): conditional generation prompt based on partial mode detection
- VLMBatchedEngine: strip `partial` field in both template methods but always use `add_generation_prompt=True` (vision models do not support continuation)
- MLXLanguageModel.chat(): same partial mode logic as BatchedEngine

### DIFF
--- a/omlx/api/openai_models.py
+++ b/omlx/api/openai_models.py
@@ -65,6 +65,10 @@ class Message(BaseModel):
     tool_calls: Optional[List[dict]] = None
     # For tool response messages (role="tool")
     tool_call_id: Optional[str] = None
+    # Participant name, rendered into chat template (e.g. Kimi K2/K2.5 named assistants)
+    name: Optional[str] = None
+    # Continue from this message instead of starting a new turn (prefill / partial mode)
+    partial: bool = False
 
 
 # =============================================================================
--- a/omlx/api/utils.py
+++ b/omlx/api/utils.py
@@ -12,6 +12,35 @@ from .openai_models import Message
 
 
 # =============================================================================
+# Partial Mode Detection
+# =============================================================================
+
+
+def detect_and_strip_partial(messages: list[dict]) -> bool:
+    """Check if the final assistant message has partial=True; strip the field from all messages.
+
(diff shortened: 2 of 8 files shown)


blightbow:
Part of me doesn't like the isolated pop loop in detect_and_strip_partial(), but the only clean way to avoid that is an explicit kwarg for partial mode on the engine. Overriding generation mode via `chat_template_kwargs` would be clumsy because the mode should be deterministic.

I'm happy to rework that, just let me know what your preference is.

PR #337, opened 2026-03-21
feat: add XTC sampler support

blightbow:
Thread mlx-lm's XTC (eXclude Top Choices) sampling parameters through the full request pipeline. XTC was the only mlx-lm sampler missing from the omlx API surface.

- Add xtc_probability and xtc_threshold fields to SamplingParams dataclass (default 0.0 = disabled)
- Add optional xtc_probability and xtc_threshold to both ChatCompletionRequest and CompletionRequest API models
- Extend get_sampling_params() to resolve XTC values with the same request > default priority as other sampling params
- Thread XTC params through chat_kwargs dicts and direct engine calls across all API endpoints (chat, completion, anthropic messages, responses)
- Extract XTC params from kwargs in BatchedEngine and VLMBatchedEngine SamplingParams construction
- Pass xtc_probability, xtc_threshold, and xtc_special_tokens to both make_sampler() call sites in the scheduler
- Add _get_xtc_special_tokens() helper to Scheduler, delegating to _get_stop_tokens() for EOS coverage and caching the result at init time
- Add 10 new tests covering defaults, passthrough, API model acceptance, and special token derivation

### DIFF
--- a/Formula/omlx.rb
+++ b/Formula/omlx.rb
@@ -1,8 +1,8 @@
 class Omlx < Formula
   desc "LLM inference server optimized for Apple Silicon"
   homepage "https://github.com/jundot/omlx"
-  url "https://github.com/jundot/omlx/archive/refs/tags/v0.2.21.tar.gz"
-  sha256 "6108e2ce5a64b3baff05e0a35cc25672e97497bfba8e0e1cf436ff34a0e30eb6"
+  url "https://github.com/jundot/omlx/archive/refs/tags/v0.2.22.tar.gz"
+  sha256 "d276415648ff3206ce1e5f2e266fbdbe4497e8ccd312b2b3a5062502be1b577c"
   license "Apache-2.0"
 
   head "https://github.com/jundot/omlx.git", branch: "main"
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -3840,6 +3840,12 @@ class Scheduler:
         # Update stop tokens after cleaning up finished requests
         if finished_ids:
             self._update_stop_tokens()
+            # Reclaim stale Metal buffers from generation/cache-store intermediates.
+            # Without this, freed buffers accumulate in the Metal buffer pool across
+            # requests (since set_cache_limit(total_mem) prevents automatic release).
+            # The pool bloat forces expensive emergency GC during the next prefill,
+            # causing periodic TTFT spikes. See issue #411.
+            _sync_and_clear_cache()
 
     def _is_cache_corruption_error(self, error: Exception) -> bool:
         """Check if an error indicates cache corruption."""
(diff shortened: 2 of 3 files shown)


blightbow:
I aligned to the upstream mlx-lm defaults on this PR. This means both parameters must be specified in order for the feature to work. It should be noted that the original author's implementation (https://github.com/oobabooga/text-generation-webui/pull/6335) uses a threshold of 0.1:

https://github.com/p-e-w/text-generation-webui/blob/29d38a74ba8d47206f2371b44a746111fac80f4c/modules/presets.py#L47-L48

@jundot I personally think that principle of least astonishment is more in favor of the 0.1 default (don't silently fail when the user wants the feature enabled but doesn't set a threshold), but I'll defer to your preference on this one. Feel free to tweak that value yourself if you accept the PR.

blightbow:
On second thought, it doesn't make sense to model after upstream on this. A xtc_threshold of 0.0 when xtc_probability >0 doesn't actually disable it! I had Claude Code verify my analysis:

> ❯ Wait, let's double check that. xtc_threshold of 0.0 when xtc_probability > 0.0 is not treated as disabled?
> 
> (snip)
> 
> ⏺ Good catch to verify. With xtc_threshold=0.0:
> 
>   mask = probs > mx.where(probs > 0.0, probs, mx.inf).min()
> 
>   Since all probabilities are > 0.0 (after softmax), mx.where(probs > 0.0, probs, mx.inf) returns all probs unchanged, and .min() gives the minimum probability. The mask then selects all tokens with probability greater than the minimum — so all tokens except the least probable one get masked to -inf.
> 
>   That's not "disabled" — it's maximally aggressive. It would exclude every token except the lowest-probability one (plus special tokens). With xtc_probability=0.5, that would happen on 50% of token samples.
> 
>   So upstream's 0.0/0.0 is safe only because both default to 0 and the if xtc_probability > 0.0 gate prevents apply_xtc from ever being called. But if a user sets xtc_probability without setting xtc_threshold, they'd get destructive behavior silently.
> 

As such, I would consider upstream's default to be a **bug**. I'll change our own default and log that bug separately.

jundot:
Reviewed the full diff. Looking good.

The safe default for xtc_threshold (0.1 instead of upstream's 0.0) is a nice touch. Upstream's 0.0 would nuke sampling if someone only sets probability without thinking about threshold.

_get_xtc_special_tokens() reusing _get_stop_tokens() keeps things clean. All 5 get_sampling_params() call sites are updated, tests cover the important paths.

The growing tuple return from get_sampling_params() (now 10 elements) is getting unwieldy but that's pre-existing tech debt, not something to hold this up for. I'll clean that up separately.

No regression concerns. XTC defaults to disabled (probability=0.0) so existing behavior is untouched.

jundot:
@blightbow v0.3.0rc1 is out with your XTC sampler included. https://github.com/jundot/omlx/releases/tag/v0.3.0rc1 — if you get a chance, please give it a test and let me know if anything looks off. thanks!

PR #369, opened 2026-03-24
fix(embedding): support custom processor input preparation

MasakiMu319:
## Summary
  - route embedding requests through custom processor hooks when available
  - keep the existing generic tokenizer path for standard text embedding models
  - add regression coverage for both compiled and eager embedding execution

## Why
  Qwen3-VL embedding models can be loaded through `mlx-embeddings`, but oMLX always used the generic `processor(texts, ...)` path for embedding requests. For custom processors such as `qwen3_vl`, that positional call is interpreted as image input, which breaks `/v1/embeddings` even when the model is explicitly treated as an embedding model.

## Testing
  - `uv run pytest tests/test_embedding.py -k "custom_processor or compiled_path_fallback_on_failure or is_compiled_false_uses_eager_path"`
  - started oMLX with `Qwen3-VL-Embedding-2B-mxfp8` forced to `embedding` and verified `/v1/embeddings` returns vectors for single and batch text inputs

### DIFF
--- a/omlx/models/embedding.py
+++ b/omlx/models/embedding.py
@@ -209,6 +209,48 @@ class MLXEmbeddingModel:
             "(text_embeds, pooler_output, or last_hidden_state)"
         )
 
+    def _uses_custom_embedding_inputs(self, processor) -> bool:
+        """Return True when processor exposes a custom embedding input API."""
+        return hasattr(processor, "prepare_embedding_inputs") or hasattr(
+            processor, "prepare_model_inputs"
+        )
+
+    def _prepare_embedding_inputs(
+        self,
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the fix @MasakiMu319, the approach looks clean and correct.

Routing through `prepare_embedding_inputs` / `prepare_model_inputs` when the processor exposes them makes sense, and the fallback to the generic `prepare_inputs` path keeps things safe for standard text models.

I verified the existing text embedding path is not affected by this change. The tests cover both compiled and eager execution with custom processors, which is what i want to see.

Merging this.

PR #286, opened 2026-03-17
MCP supports Streamable HTTP.

tianfeng98:
For #285

PR #377, opened 2026-03-25
feat(chat): MCP tool call loop in built-in chat UI

rayone:
## Problem

The built-in chat UI (`/admin/chat`) silently drops responses when a model decides to call an MCP tool. `streamResponse` only handled `delta.content` and `delta.reasoning_content` — when the model returns `finish_reason: tool_calls` there is no `delta.content`, so `streamingContent` stays empty and nothing is pushed to the message list.

## Solution

- Accumulate `delta.tool_calls` chunks during streaming (OpenAI chunked format, keyed by `index`)
- On `finish_reason: tool_calls`: push a hidden `assistant` message with `tool_calls` into history, execute all tools **in parallel** via `POST /v1/mcp/execute`, push hidden `tool` result messages, then recurse back into `streamResponse` so the model receives full tool context and streams a final answer
- Show a per-tool wrench indicator (`toolName…`) in the UI while execution is in flight
- Hide bookkeeping `assistant`/`tool` protocol messages from display using `_ui: false` / `x-show="msg._ui !== false"`
- Simplify `messagesForApi` to a single filter+map pass (was two passes with redundant filter)

## Tested with

Tavily MCP + Qwen3.5-27B on macOS (Apple Silicon)

### DIFF
--- a/omlx/admin/templates/chat.html
+++ b/omlx/admin/templates/chat.html
@@ -1260,7 +1260,7 @@
                         // Execute all tools in parallel, show indicators while waiting
                         const indicatorIndices = toolCalls.map(() => {
                             const idx = this.messages.length;
-                            this.messages.push({ role: 'tool_call', content: '', _ui: true });
+                            this.messages.push({ role: 'tool_call', content: '', _ui: true, _error: false });
                             return idx;
                         });
 


rayone:
**Note re: #286**

This PR and #286 are orthogonal:

- **#286** adds `streamable_http` transport to the Python MCP client — so oMLX can connect directly to remote MCP servers (like Tavily at `https://mcp.tavily.com/mcp/`) without needing `mcp-remote` as a stdio bridge.
- **This PR (#377)** fixes the chat UI tool call loop — the built-in chat at `/admin/chat` was silently dropping responses whenever a model returned `finish_reason: tool_calls`.

Both are needed. If #286 merges first it would also allow simplifying `mcp.json` configs that currently work around the missing streamable HTTP transport using `npx mcp-remote`.

rayone:
I'll create another PR with some more enhancements to the Chat UI.

PR #391, opened 2026-03-26
fix: copy block_ids before storing in prefix index

andafterall:
## Problem

In `_update_prefix_index()`, the `block_ids` list is stored directly into `_prefix_index`.

But this is the same object as `block_table.block_ids`. So if this list is changed later (for example, CoW replaces a block ID), the prefix index entry also changes.

This may cause wrong block hit when another request looks up the same prefix.

## Fix

Use `tuple(block_ids[:i+1])` to store an immutable copy.

- `tuple` cannot be changed after creation
- `[:i+1]` only keeps blocks needed for this prefix length

## Test

- Added `test_prefix_index_immutable_after_store`
- `pytest tests/test_prefix_cache.py -v` — all 73 passed
- `pytest -m "not slow"` — 2971 passed, no new failures

### DIFF
--- a/omlx/cache/prefix_cache.py
+++ b/omlx/cache/prefix_cache.py
@@ -2018,7 +2018,7 @@ class BlockAwarePrefixCache(CacheManager):
 
             parent_hash = block_hash
             prefix_len += len(block_tokens)
-            self._prefix_index[block_hash] = (prefix_len, block_ids, i + 1)
+            self._prefix_index[block_hash] = (prefix_len, tuple(block_ids[: i + 1]), i + 1)
 
     def get_stats(self) -> PrefixCacheStats:
         """
--- a/tests/test_prefix_cache.py
+++ b/tests/test_prefix_cache.py
@@ -504,6 +504,32 @@ class TestPrefixIndexOperations:
         prefix_len, matched_ids, num_blocks = result
         assert prefix_len == 4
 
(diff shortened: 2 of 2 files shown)


jundot:
Solid fix. The mutable reference bug is real — when CoW does `table.block_ids[i] = new_block.block_id` in `get_blocks_for_generation()`, the stored prefix index entry gets silently corrupted. Using `tuple()` is the right call since these entries should never be modified after storage, and the `[:i+1]` slice is a nice touch that avoids storing unnecessary trailing block ids.

One minor thing: the type annotations for `_prefix_index` (line 101) and `_find_best_prefix_match` return type (line 1983) still say `List[int]` but now the actual stored value is a `tuple`. I'll fix that in a follow-up commit after merging.

Test looks good too. Merging, thanks!

jundot:
@andafterall v0.3.0rc1 is out with your prefix index fix included. https://github.com/jundot/omlx/releases/tag/v0.3.0rc1 — if you get a chance, please give it a test and let me know if anything looks off. thanks!

andafterall:
@jundot 
Tested on v0.3.0rc1:

- pytest -m "not slow" — 3229 passed, 0 failed
- prefix index fix is in, and I see you already updated the type annotations, nice

Everything looks good on my side. Thanks for the merge and the quick follow-up! Really like the direction of this project — with agents taking off, optimizing prefix cache for long-context agent workloads on Apple Silicon is exactly what's needed.

PR #426, opened 2026-03-27
fix: update menubar content in real time while menu is open

### LINKED ISSUE #425: 菜单栏内容在服务器状态变化时不实时更新
问题描述

点击"Start Server"后，菜单栏内容（服务器状态文字、Admin Panel / Chat with oMLX 按钮启用状态）不会自动更新。用户必须关闭菜单再重新打开才能看到最新状态。即使菜单保持打开，内容也完全冻结。

is a little problem

EmotionalAmo:
Fixes #425

## Problem

After clicking "Start Server", the menu bar content (server status text, Admin Panel / Chat button enabled state) never updates while the menu is open. The user must close and reopen the menu to see the latest state. Even if the menu stays open, the content is completely frozen.

## Root Causes

**1. `NSTimer` registered under `NSDefaultRunLoopMode` only (`app.py:136`)**

When the status-bar menu is open, macOS enters `NSEventTrackingRunLoopMode`. `NSDefaultRunLoopMode` is suspended during this time, so `healthCheck_` never fires and the menu content freezes completely.

**2. `_build_menu()` replaces the entire `NSMenu` object (`app.py:490, 730`)**

Even if the timer fired, calling `self.status_item.setMenu_(newMenu)` while a menu is being displayed would close it — causing disruptive flicker and not the desired in-place update.

## Fix

Two complementary changes in `packaging/omlx_app/app.py`:

### 1. `NSRunLoopCommonModes`
Register the health-check timer under `NSRunLoopCommonModes` instead of `NSDefaultRunLoopMode` so it fires in all run-loop modes, including during menu interaction.

```python
# Before
NSRunLoop.currentRunLoop().addTimer_forMode_(self.health_timer, NSDefaultRunLoopMode)

# After
NSRunLoop.currentRunLoop().addTimer_forMode_(self.health_timer, NSRunLoopCommonModes)
```

### 2. In-place menu item updates via `_refresh_menu_in_place()`
Instead of replacing the `NSMenu` object, store references to the key `NSMenuItem` objects and mutate them directly:
- Status header: `setAttributedTitle_()` with updated color/text
- Server control buttons: `setHidden_()` to toggle Start / Stop / Force Restart visibility
- Admin Panel / Chat: `setEnabled_()` based on running state

### 3. `NSMenuDelegate` — `menuWillOpen_` / `menuDidClose_`
- `menuWillOpen_`: always refreshes menu content right before it appears, so even without the timer the menu is never stale on open.
- `menuDidClose_`: clears the `_menu_is_open` flag so `healthCheck_` uses `_build_menu()` (full rebuild with stats) when the menu is closed.

## Test Steps

1. Launch the app — menu shows "● oMLX Server is stopped" + "Start Server"
2. Open the menu and **keep it open**, then click "Start Server"
3. ✅ Status text updates in place: "Stopped" → "Starting..." → "Running" — no need to close the menu
4. ✅ "Admin Panel" and "Chat with oMLX" become enabled automatically once running
5. Close and reopen the menu — state is correct immediately (`menuWillOpen_`)
6. Menubar icon updates alongside status changes

### DIFF
--- a/packaging/omlx_app/app.py
+++ b/packaging/omlx_app/app.py
@@ -562,25 +562,7 @@ class OMLXAppDelegate(NSObject):
         # All three items are always present; setHidden_ controls visibility so
         # _refresh_menu_in_place() can toggle them without replacing the NSMenu.
 
-        # Stop Server — visible when RUNNING / STARTING / UNRESPONSIVE
-        stop_item = NSMenuItem.alloc().initWithTitle_action_keyEquivalent_(
-            "Stop Server", "stopServer:", ""
-        )
-        stop_item.setTarget_(self)
-        stop_icon = self._create_menu_icon("stop.circle")
-        if stop_icon:
-            stop_item.setImage_(stop_icon)
-        stop_item.setHidden_(
-            status not in (
-                ServerStatus.RUNNING,
-                ServerStatus.STARTING,
-                ServerStatus.UNRESPONSIVE,
-            )
-        )
-        self.menu.addItem_(stop_item)
-        self._stop_item = stop_item
-
-        # Force Restart — visible when UNRESPONSIVE / ERROR
+        # Force Restart — visible when UNRESPONSIVE / ERROR (most important, shown first)
         restart_item = NSMenuItem.alloc().initWithTitle_action_keyEquivalent_(
             "Force Restart", "forceRestart:", ""
         )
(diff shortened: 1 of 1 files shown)


EmotionalAmo:
Thanks for the thorough review! All four issues have been addressed in the follow-up commit (339922b):

1. **`_fetch_stats()` blocks main thread** — Skipped when `_menu_is_open` is `True`. The stats fetch only runs (and `_last_stats_fetch` only advances) when the menu is closed, so it will catch up on the next timer cycle after the menu is dismissed.

2. **STOPPING state hides all buttons** — Added `ServerStatus.STOPPING` to the Stop Server visibility condition in both `_build_menu()` and `_refresh_menu_in_place()`.

3. **Button order reversed for UNRESPONSIVE** — Not intentional. Restored the original order: Force Restart first, Stop Server second.

4. **Icon template state out of sync** — Added `icon.setTemplate_(True)` after each `setEnabled_()` call in `_refresh_menu_in_place()` to keep icon rendering consistent with `_build_menu()`.

jundot:
@EmotionalAmo v0.3.0rc1 is out with your menubar fix included. https://github.com/jundot/omlx/releases/tag/v0.3.0rc1 — if you get a chance, please give it a test and let me know if anything looks off. thanks!

PR #466, opened 2026-03-29
feat: resolve HF Hub cache layout in model directory scanning

AlexWorland:
## Summary

Models downloaded via `huggingface-cli` or `huggingface_hub` use an indirect cache layout (`models--Org--Name/refs/main` → commit hash → `snapshots/<hash>/`) instead of a flat directory. When a user adds their HF cache directory (e.g., `~/.cache/huggingface/hub`) to model directories in settings, the scanner didn't understand this indirection and found nothing.

This change teaches the existing two-level scanner to resolve HF Hub cache entries to their active snapshot. No directories are scanned automatically — users opt in by adding their HF cache path to model directories. The resolved snapshot flows through the same `_is_model_dir()` → `_register_model()` pipeline as all other models.

## Changes

| File | Change |
|------|--------|
| `omlx/model_discovery.py` | Add `_resolve_hf_cache_entry()` — resolves `models--Org--Name/` to `snapshots/<hash>/` via `refs/main`; integrate into `discover_models()` between Level 1 and Level 2 checks |
| `omlx/admin/routes.py` | Add HF cache resolution to `list_hf_models()` dashboard endpoint; extract duplicated dedupe/size/append logic into `_add_model()` helper |
| `tests/test_model_discovery.py` | 13 new tests — 6 unit tests for `_resolve_hf_cache_entry()` edge cases, 7 integration tests for `discover_models()` with HF cache layouts |

## How it works

```
discover_models(model_dir):
  for each subdir:
    1. Is it an adapter?           → skip
    2. Has config.json?            → Level 1: register directly
    3. Is it models--Org--Name/?   → NEW: resolve snapshot, register if valid model
    4. Otherwise                   → Level 2: scan as org folder
```

Existing flat and org-nested layouts are completely unaffected — the HF cache check only fires for directories matching the `models--*--*` naming pattern, and the `continue` ensures it doesn't fall through to the org scan.

## Testing

```
tests/test_model_discovery.py    88 passed (75 existing + 13 new)
```

New test coverage:
- `_resolve_hf_cache_entry()`: valid entry, regular dir, missing org separator, missing refs/main, missing snapshot, whitespace stripping
- `discover_models()`: single/multiple HF cache models, model_path points to snapshot, missing config.json skipped, mixed flat+HF cache, mixed org+HF cache, no fallthrough to org scan

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -110,3 +110,6 @@ omlx/_build_info.py
 
 # Tailwind CSS standalone CLI binary (download on demand via build_css.py)
 omlx/admin/tailwindcss-*
+
+# Git worktrees
+.worktrees/
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -2952,6 +2952,24 @@ async def list_hf_models(is_admin: bool = Depends(require_admin)):
 
     model_dirs = global_settings.model.get_model_dirs(global_settings.base_path)
 
+    from ..model_discovery import _resolve_hf_cache_entry
+
+    def _add_model(model_path: Path, model_name: str) -> None:
+        if model_name in seen_names:
+            return
+        seen_names.add(model_name)
+        total_size = sum(
+            f.stat().st_size for f in model_path.rglob("*") if f.is_file()
+        )
+        models.append(
+            {
+                "name": model_name,
+                "path": str(model_path),
+                "size": total_size,
+                "size_formatted": format_size(total_size),
+            }
+        )
+
     models = []
     seen_names: set[str] = set()
     for model_dir in model_dirs:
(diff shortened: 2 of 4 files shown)


jundot (approved):
Thanks for this, clean work.

The `_resolve_hf_cache_entry()` function is well structured and handles the edge cases properly (missing refs, missing snapshots, whitespace in hash). The `continue` placement is correct so HF cache dirs don't accidentally fall through to the org folder scan.

One minor note: in `routes.py`, the `_add_model` closure is defined before `models` and `seen_names` are declared. Works fine because of python's late-binding closures, but moving it after those two lines would read more naturally. Not a blocker though.

Tests look solid, merging.

PR #467, opened 2026-03-29
fix: route voice param correctly for CustomVoice TTS models

### LINKED ISSUE #461: /v1/audio/speech returns 500 for Qwen3-TTS CustomVoice even when voice is provided
Description
When using the OpenAI-compatible POST /v1/audio/speech endpoint with Qwen3-TTS-12Hz-1.7B-CustomVoice, the request returns HTTP 500 with:

CustomVoice model requires 'voice' (speaker name) (e.g., 'Chelsie', 'Ethan', 'Vivian')

However, the server log clearly shows that the voice field was received and used by the TTS engine before the error is raised.

This suggests that voice is accepted in one code path and then incorrectly revalidated or lost in a later step.

Observed behavior
Request:

```
curl -v -X POST "http://<host>:8080/v1/audio/speech" \
  -H "Authorization: Bearer <token>" \
  -d '{
    "model": "Qwen3-TTS-12Hz-1.7B-CustomVoice",
    "input": "Hallo Christopher, das ist ein Test auf Deutsch.",
    "voice": "Vivian",
    "instructions": "gender: Female; language: German; accent: Standard German; pitch: medium; speed: moderate; volume: normal; emotion: calm and friendly; tone: natural; do not whisper.",
    "response_format": "mp3"
  }' \
  -o test.mp3
```

Server log:

TTS synthesize: model=/Users/.../Qwen3-TTS-12Hz-1.7B-CustomVoice, text_len=48, voice=vivian, speed=1.0
POST /v1/audio/speech → 500: CustomVoice model requires 'voice' (speaker name) (e.g., 'Chelsie', 'Ethan', 'Vivian')

Expected behavior
If voice is present and valid, the request should succeed and return audio instead of HTTP 500.

Additional inconsistency
Another related inconsistency:

When no valid voice is provided, the server logs:

TTS synthesize: model=/Users/.../Qwen3-TTS-12Hz-1.7B-CustomVoice, text_len=48, voice=default, speed=1.0
POST /v1/audio/speech → 500: Speaker 'default' not supported. Available: ['serena', 'vivian', 'uncle_fu', 'ryan', 'aiden', 'ono_anna', 'sohee', 'eric', 'dylan']

This indicates that the server knows the actual supported speaker list.

But the other error message still says:

CustomVoice model requires 'voice' (speaker name) (e.g., 'Chelsie', 'Ethan', 'Vivian')

That example list appears inconsistent with the actual supported speakers reported by the server.

Important comparison
The same OpenAI-compatible endpoint works with Qwen3-TTS-12Hz-1.7B-VoiceDesign.

So this does not appear to be a general /v1/audio/speech problem. It seems specific to the CustomVoice path.

Why this looks like a bug
The log proves that:
	•	the request contains voice
	•	the engine receives voice
	•	synthesis starts with that voice

So the later 500 error claiming that voice is missing is internally contradictory.

Also, since VoiceDesign works through the same endpoint, this looks like a model-specific issue in the CustomVoice handling rather than a generic TTS endpoint failure.

Environment
	•	Endpoint: OpenAI-compatible POST /v1/audio/speech
	•	Model: Qwen3-TTS-12Hz-1.7B-CustomVoice
	•	Reproduced with valid speaker names such as vivian
	•	Response format: mp3

Possible cause
There may be a second validation step after synthesis starts that:
	•	reads a different field name
	•	checks the pre-normalized value instead of the normalized speaker
	•	or loses voice between request parsing and response generation

Impact
This currently makes the OpenAI-compatible CustomVoice path unusable, even with a valid speaker name, while VoiceDesign continues to work on the same endpoint.

ethannortharc:
## Summary

- Fix voice parameter routing in `TTSEngine.synthesize()` that caused HTTP 500 for CustomVoice models (e.g. `Qwen3-TTS-12Hz-1.7B-CustomVoice`)
- The previous logic checked for `instruct` first, misrouting the speaker name when a model accepts both `voice` and `instruct` parameters
- Now prioritizes `voice` when present; falls back to `instruct` only for VoiceDesign-only models

Fixes #461

## Changes

- `omlx/engine/tts.py`: Swap parameter priority — check `voice` before `instruct`
- `tests/test_audio_tts.py`: Add `TestTTSVoiceRouting` with 3 test cases covering CustomVoice, VoiceDesign, and voice-only models

## Test plan

- [x] Unit tests pass (`pytest tests/test_audio_tts.py` — 15 passed)
- [x] Manual verification with `Qwen3-TTS-12Hz-1.7B-CustomVoice-bf16` — tested all 7 available speakers:
  - `vivian` (Chinese text) — 200 OK, valid WAV
  - `aiden` (Japanese text) — 200 OK, valid WAV
  - `serena` (English text) — 200 OK, valid WAV
  - `ryan` (English text) — 200 OK, valid WAV
  - `eric` (English text) — 200 OK, valid WAV
  - `dylan` (Chinese text) — 200 OK, valid WAV
  - `vivian` (English text) — 200 OK, valid WAV
- [x] `Qwen3-TTS-12Hz-1.7B-Base-bf16` and `Qwen3-TTS-12Hz-0.6B-Base-bf16` also verified working with voice param

### DIFF
--- a/omlx/engine/tts.py
+++ b/omlx/engine/tts.py
@@ -151,14 +151,16 @@ class TTSEngine(BaseNonStreamingEngine):
                 "verbose": False,
             }
             if voice is not None:
-                # VoiceDesign models expect voice description in 'instruct',
-                # not 'voice'. Detect by checking if generate() accepts 'instruct'.
+                # Route the voice value to the correct generate() parameter.
+                # Models like CustomVoice accept a speaker name via 'voice';
+                # VoiceDesign-only models accept a description via 'instruct'.
+                # When both exist, 'voice' takes priority (CustomVoice).
                 import inspect
                 gen_params = inspect.signature(model.generate).parameters
-                if "instruct" in gen_params and voice != "default":
-                    gen_kwargs["instruct"] = voice
-                else:
+                if "voice" in gen_params:
                     gen_kwargs["voice"] = voice
+                elif "instruct" in gen_params:
+                    gen_kwargs["instruct"] = voice
             if speed != 1.0:
                 gen_kwargs["speed"] = speed
             gen_kwargs.update(kwargs)
--- a/tests/test_audio_tts.py
+++ b/tests/test_audio_tts.py
@@ -242,6 +242,72 @@ class TestTTSEndpointErrors:
 
 
 # ---------------------------------------------------------------------------
+# TestTTSVoiceRouting — unit tests for voice/instruct parameter dispatch
+# ---------------------------------------------------------------------------
+
+
+class TestTTSVoiceRouting:
+    """Verify that the voice value is routed to the correct generate() kwarg."""
+
+    @pytest.fixture
+    def _run_synthesize(self):
+        """Helper: run TTSEngine.synthesize and return the kwargs passed to generate()."""
+        import asyncio
+        from omlx.engine.tts import TTSEngine
+
+        def _run(generate_sig_params, voice_value):
+            engine = TTSEngine("test-model")
+
+            # Build a mock model whose generate() has the requested signature
+            mock_model = MagicMock()
+            params = [
+                MagicMock(name="text", default=MagicMock()),
+                MagicMock(name="verbose", default=False),
+            ]
+            import inspect
+            sig_params = {
+                "text": inspect.Parameter("text", inspect.Parameter.POSITIONAL_OR_KEYWORD),
+                "verbose": inspect.Parameter("verbose", inspect.Parameter.POSITIONAL_OR_KEYWORD, default=False),
+            }
+            for p in generate_sig_params:
+                sig_params[p] = inspect.Parameter(p, inspect.Parameter.POSITIONAL_OR_KEYWORD, default=None)
+            mock_model.generate = MagicMock()
+            mock_model.generate.__signature__ = inspect.Signature(parameters=list(sig_params.values()))
+            mock_model.generate.return_value = []  # no audio chunks
+
+            engine._model = mock_model
+
+            try:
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for the fix, this was a solid root cause analysis.

I found one thing after merging though. All Qwen3-TTS variants (Base, CustomVoice, VoiceDesign) share the same `generate(voice=, instruct=)` signature, so checking for `"voice" in gen_params` is always true for every Qwen3-TTS model. This means the `elif "instruct"` branch never runs for VoiceDesign either.

In practice, if someone sends `voice` to a VoiceDesign model, it goes to the `voice` kwarg (which VoiceDesign ignores) and `instruct` stays empty, so the model raises a 500.

I added two follow-up commits on main:

1. Changed `AudioSpeechRequest.voice` default from `"default"` to `None`. No model recognized `"default"` as a valid speaker anyway, and `None` lets models use their own defaults (Kokoro defaults to `"af_heart"`, etc).

2. Added an `instructions` field to `AudioSpeechRequest`. This maps directly to the `instruct` kwarg in generate(), so VoiceDesign users can pass voice descriptions through `instructions` instead of `voice`. It also means CustomVoice can now receive both a speaker name (`voice`) and a voice description (`instructions`) at the same time.

ethannortharc:
@jundot  Thanks for the follow-up — that's a much cleaner solution. I hadn't considered that all Qwen3-TTS variants share the same signature, so the `voice`-first check would break VoiceDesign. Adding a separate `instructions` field makes the intent explicit at the API level instead of trying to infer it from the model signature. Good call on the `None` default too.

PR #470, opened 2026-03-30
fix: Devstral/Mistral tool calling broken by empty tool_call_end marker

### LINKED ISSUE #464: Devstral 2 small tool calling issue
Hi! Not sure if this is a bug or I am just dumb and doing something wrong, but I have a next issue. When prompting Devstral to do any code analysis using Vibe CLI, I am getting the first response and then the model just stops. All other subsequent responses are coming with this error.

2026-03-29 11:49:15,896 - omlx.server - WARNING - [-] - POST /v1/chat/completions → 400: Chat template error: Assistant message must have a string or a list of chunks in content or a list of tool calls.

Image

I tried Claude code as well, and while it has a similar issue (stops after the first prompt), I don't see the same error in the log.
Here is the only message that i see for claude
2026-03-29 12:24:45,709 - omlx.engine.vlm - INFO - [-] - [vlm_stream_generate] GeneratorExit for request ed31ddc9-9ddf-44af-af6c-b7c88724a857
2026-03-29 12:24:59,813 - omlx.engine.vlm - INFO - [-] - [vlm_stream_generate] GeneratorExit for request 8ae47e1c-0bf8-44ce-b2f6-bdd02189581f

It does work in LM Studio, albeit not out of the box. I modified Jinja to make sure it won't fail on some of the tools’ calls.  

I tried to restore the original chat_template and tried a few different fixed versions, but no luck with it.

Here is the model that is being used https://huggingface.co/mlx-community/Devstral-Small-2-24B-Instruct-2512-4bit

TipKnuckle:
## Summary

Fixes Devstral (and any Mistral model with `tool_call_end=""`) tool calling being completely broken — tool calls produce silent empty responses, and subsequent turns crash with a 400 template error.

Closes #464

## Root Cause

Two bugs, same origin — Devstral's tool calling format uses a **one-sided marker** (`tool_call_start="[TOOL_CALLS]"`, `tool_call_end=""`). The model outputs `[TOOL_CALLS]name[ARGS]{...}` with no closing marker.

**Bug 1 — `parse_tool_calls()` silent failure:**  
The condition `tool_call_end is not None` passes (empty string ≠ None), then the regex `r'\[TOOL_CALLS\](.*?)'` with non-greedy `(.*?)` and empty end anchor captures **zero characters every time**. The parser fails silently (exception swallowed), no tool calls are ever extracted. Meanwhile `ToolCallStreamFilter` correctly suppresses the raw markup from content, so the response has `content=None, tool_calls=None` — a silent empty response.

**Bug 2 — Void assistant messages crash strict chat templates:**  
The empty response from Bug 1 gets echoed back by the client as `{"role": "assistant", "content": null}` with no `tool_calls`. Devstral's Jinja chat template explicitly rejects assistant messages with empty content AND no tool_calls with `raise_exception()`.

## Changes

- **`omlx/api/tool_calling.py`**: When `tool_call_end` is falsy, split on the start marker with `re.split()` and parse each segment individually, instead of the broken start/end regex pair. Each segment is passed to the model's native `tool_parser` one at a time (it handles one call per invocation).

- **`omlx/api/utils.py`**: Added `_drop_void_assistant_messages()` — filters out assistant messages with no content AND no tool_calls before template rendering. Wired into all three extract functions (`extract_text_content`, `extract_multimodal_content`, `extract_harmony_messages`).

## Why this is safe for other models

- **Bug 1 fix**: Gated by `if tool_call_end:` / `else:`. Models with paired markers (e.g. `<tool_call>...</tool_call>`) have truthy `tool_call_end` and take the existing path unchanged. Only models with empty/falsy `tool_call_end` (Mistral/Devstral) take the new path.

- **Bug 2 fix**: Only drops assistant messages where content is empty/None AND there are no tool_calls. These are semantically empty no-op messages. The OpenAI API itself doesn't allow them. Dropping them is strictly better than crashing on strict templates, and harmless on lenient ones.

## Tests

- 5 new tests in `TestParseToolCallsEmptyEndMarker`: content parsing, text cleaning, multiple tool calls, parser failure handling, native parser invocation
- 7 new tests in `TestDropVoidAssistantMessages`: void message dropping, content/tool_call preservation, role filtering, integration with `extract_text_content`
- Updated `test_none_content` to reflect new void-dropping behavior
- All 225 tests in the modified files pass; integration and streaming tests unaffected

### DIFF
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -50,7 +50,7 @@ def _parse_xml_tool_calls(text: str) -> Tuple[str, Optional[List[ToolCall]]]:
         Tuple of (cleaned_text, tool_calls or None)
     """
     tool_calls = []
-    pattern = r'<tool_call>(.*?)</tool_call>'
+    pattern = r"<tool_call>(.*?)</tool_call>"
     matches = re.findall(pattern, text, re.DOTALL)
 
     for match in matches:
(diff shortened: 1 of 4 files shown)


jundot:
merged manually with conflict resolution and a fix for the drop_void/merge_consecutive ordering.

the original PR applied _drop_void_assistant_messages after _merge_consecutive_roles, which could leave consecutive same-role messages unmerged (e.g. [user, user] after dropping a void assistant between them). swapped the order so drop runs first, then merge catches any newly-adjacent same-role messages.

added 2 tests for the ordering fix. all 233 tests pass.

commit 6d5877b preserves your authorship. thanks for the thorough root cause analysis and fix!

jundot:
merged manually in 6d5877b because the branch had conflicts with main. your authorship is preserved on the commit.

i also fixed the ordering of `_drop_void_assistant_messages` — the original ran it after `_merge_consecutive_roles`, which could leave unmerged consecutive users when a void assistant between them gets dropped. swapped the order so drop runs first, merge runs after. added 2 tests for that.

tested with Devstral-2-123B-Instruct-2512-4bit, tool calling and void assistant recovery both work. thanks for the fix!

Issue #489, opened 2026-03-31
Audio endpoints (/audio/speech, /audio/transcriptions, /audio/process)    do not resolve model aliases


haokaiyang:
**Describe the bug**
  
  /v1/audio/speech, /v1/audio/transcriptions, and /v1/audio/process
  endpoints do not resolve model aliases before looking up the engine. If
  a user sets a model_alias (e.g. qwen3-tts) for a local TTS model,
  the request fails with a 404 even though the alias is correctly
  configured.

**To Reproduce**
   1. Set a model_alias (e.g. qwen3-tts) for a local TTS model via
      the admin panel
   2. Send a request to /v1/audio/speech with "model": "qwen3-tts"
   3. See error: "Model 'qwen3-tts' not found. Available: Qwen3-TTS-12Hz-1.7B-xxx"

**Expected behavior**
The audio endpoints should resolve model aliases to their actual model
  IDs (directory names), consistent with how LLM/chat endpoints behave. A
  request with an aliased model name should succeed and route to the
  correct engine.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS
 - Version v0.3.0

**Smartphone (please complete the following information):**

**Additional context**

XDGH0624:
stream support pls

kuanjames:
Is this why the voice keeps changing on Qwen3-tts?

A different voice appears after each punctuation.

ethannortharc:
This was an oversight in the audio routes — they call `pool.get_engine()` directly with the request model ID without
resolving aliases first, unlike the LLM/chat endpoints which go through `resolve_model_id()`.

Submitted a fix in #525 that adds alias resolution to all three audio endpoints (`/v1/audio/speech`, `/v1/audio/transcriptions`, `/v1/audio/process`).

PR #490, opened 2026-03-31
fix(audio): resolve model alias lookup for STT/TTS/STS endpoints

haokaiyang:
- Fixes audio endpoints not resolving model aliases before loading engines.
- Adds alias resolution in:
      - /v1/audio/transcriptions
      - /v1/audio/speech
      - /v1/audio/process
- Introduces _get_settings_manager() for lazy access to server settings (keeps import pattern consistent with _get_engine_pool).
- Uses local model variable in TTS route instead of mutating request.model, improving handler-side consistency and reducing side effects.

### DIFF
--- a/omlx/api/audio_routes.py
+++ b/omlx/api/audio_routes.py
@@ -52,6 +52,17 @@ def _get_engine_pool():
     return pool
 
 
+def _get_settings_manager():
+    """Return the active SettingsManager from server state.
+
+    Imported lazily to avoid a circular import at module load time.
+    Can be replaced in tests via patch('omlx.api.audio_routes._get_settings_manager').
+    """
+    from omlx.server import _server_state
+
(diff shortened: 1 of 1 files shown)


haokaiyang:
Fixes #489

jundot:
Thanks for the fix! This correctly adds alias resolution to all three audio endpoints and the approach is clean.

One concern though — the `_get_settings_manager()` helper duplicates what `server.resolve_model_id()` already does internally. If the resolve logic ever changes in `server.py`, this code path wouldn't pick it up. Also there are no tests covering the new behavior.

PR #525 takes a different approach by delegating to `server.resolve_model_id()` directly, which avoids the duplication and includes tests. I'm going to go with that one instead.

Closing in favor of #525. Appreciate the contribution!

PR #496, opened 2026-03-31
fix: add uv override for git-pinned mlx-lm

### LINKED ISSUE #495: New mlx-audio dependency breaks uv sync
**Describe the bug**
```
% uv sync
    Updated https://github.com/Blaizzy/mlx-vlm (7f7a04ce20337996da335ab17cdfee0ba51d6610)
    Updated https://github.com/Blaizzy/mlx-audio (6408d2a410eb8c57464e07725b92271860199250)
    Updated https://github.com/Blaizzy/mlx-embeddings (32981fa4e8064ed664b52071789dd18271fe4206)
  × No solution found when resolving dependencies:
  ╰─▶ Because omlx[audio] depends on mlx-audio and mlx-audio==0.4.1 depends on mlx-lm==0.31.1, we can conclude that omlx[audio] depends on mlx-lm==0.31.1.
      And because there is no version of mlx-lm==0.31.1 and your project requires omlx[audio], we can conclude that your project's requirements are unsatisfiable.
```

**To Reproduce**
1. Pull the latest revision of the main branch.
2. Run `uv sync`.
3. Kaboom!

**Expected behavior**
No kaboom.

 - OS: MacOS 26.4

**Additional context**
Related: https://github.com/jundot/omlx/issues/462

This is happens because the `uv.lock` file needs to be able to resolve dependencies for all extras, regardless of which the extras user has selected. The git repo does not ship the lock file (by choice), which pushes this issue down to the user when the lock file needs to be generated locally.

Fixes:
- Short term: Update pyproject.toml and instruct uv to resolve all `mlx-lm` dependencies against the same pin.
- Suggestion: Add a CI step to (re)generate the lockfile as needed. This catches a common problem that users will experience, and doesn't force the maintainer to run the `uv lock` command by hand. When it breaks, it _is_ worth knowing about because it will generate issues and PRs.

Submitting a PR for the short term fix now.

blightbow:
Fixes #495. mlx-audio==0.4.1 declares mlx-lm==0.31.1 as a transitive dependency, which conflicts with the git-commit pin of mlx-lm in pyproject.toml. The uv resolver rejects the unresolvable constraint.

Adds [tool.uv] override-dependencies to force the git-pinned mlx-lm across all transitive consumers, allowing uv sync to resolve cleanly.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -119,6 +119,13 @@ include = ["omlx*"]
 "omlx.admin" = ["templates/**/*.html", "static/**/*", "i18n/*.json"]
 "omlx.eval" = ["data/*.jsonl"]
 
+[tool.uv]
+# mlx-lm is pinned to a git commit; override transitive pins
+# (e.g. mlx-audio → mlx-lm==0.31.1) so the resolver accepts it.
+override-dependencies = [
+    "mlx-lm @ git+https://github.com/ml-explore/mlx-lm@564281f",
+]
+
 [tool.black]
 line-length = 88
 target-version = ["py310", "py311", "py312", "py313"]


jundot:
looks good, thanks for the fix! merging now.

PR #502, opened 2026-03-31
fix: resolve model aliases in audio endpoints

### LINKED ISSUE #489: Audio endpoints (/audio/speech, /audio/transcriptions, /audio/process)    do not resolve model aliases

layer4down:
## Summary

Fixes #489 - Audio endpoints (`/v1/audio/speech`, `/v1/audio/transcriptions`, `/v1/audio/process`) did not resolve model aliases before loading engines, causing 404 errors when users configured model aliases via the admin panel.

## Problem

When users set a model alias (e.g., "qwen3-tts") for a local TTS model (e.g., "mlx-community--Qwen3-TTS-mlx") via the admin panel, requests to audio endpoints failed with a 404 error even though the alias was correctly configured.

### Root Cause

The audio endpoints called `pool.get_engine(model)` directly without first resolving model aliases using `pool.resolve_model_id()`. This is inconsistent with how LLM/chat endpoints handle model resolution.

## Solution

Added a `_resolve_model_id()` helper function to `omlx/api/audio_routes.py` that mirrors the pattern used in `omlx/server.py` for chat/completion endpoints. All three audio endpoints now call this helper before loading the engine:

1. `create_transcription()` - STT endpoint
2. `create_speech()` - TTS endpoint
3. `process_audio()` - STS/audio processing endpoint

## Testing

- Created `tests/test_audio_alias_resolution.py` with 8 comprehensive tests
- All 70 audio-related tests pass (62 existing + 8 new)

## Impact

- **Breaking Changes**: None
- **Backward Compatibility**: Fully maintained - actual model IDs still work
- **Performance**: Negligible - one additional function call per request
- **Risk Level**: Low - follows existing pattern from chat endpoints

### DIFF
--- a/omlx/api/audio_routes.py
+++ b/omlx/api/audio_routes.py
@@ -52,6 +52,21 @@ def _get_engine_pool():
     return pool
 
 
+def _resolve_model_id(model_id: str) -> str:
+    """Resolve a model alias to its actual model_id (directory name).
+
+    Uses the EnginePool's resolve_model_id method with the settings manager.
+    Returns the original model_id if no alias match is found.
+    """
+    from omlx.server import _server_state
+
+    pool = _server_state.engine_pool
+    if pool is None:
+        return model_id
+
+    return pool.resolve_model_id(model_id, _server_state.settings_manager)
+
+
 async def _read_upload(file: UploadFile) -> bytes:
     """Read an uploaded file in chunks, bailing early if it exceeds the limit."""
     chunks: list[bytes] = []
@@ -96,9 +111,12 @@ async def create_transcription(
 
     pool = _get_engine_pool()
 
+    # Resolve model alias to actual model_id
+    resolved_model = _resolve_model_id(model)
+
     # Load the engine via pool (handles model loading and LRU eviction)
     try:
-        engine = await pool.get_engine(model)
+        engine = await pool.get_engine(resolved_model)
     except ModelNotFoundError as exc:
         avail = ", ".join(exc.available_models) if exc.available_models else "(none)"
         raise HTTPException(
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for putting this together, especially the thorough test coverage.

A couple things i noticed:
- `_resolve_model_id()` reimplements the alias resolution logic instead of reusing `server.resolve_model_id()` which already handles the pool-is-None guard and settings manager lookup. If the resolve logic changes later, this would fall out of sync.
- The tests are in a separate file (`test_audio_alias_resolution.py`) — the project convention is to add them to the existing per-endpoint test files (`test_audio_tts.py`, `test_audio_stt.py`, `test_audio_sts.py`).

PR #525 reuses `server.resolve_model_id()` directly and distributes tests across the existing files. I'm going with that approach instead.

Closing in favor of #525. Thanks for the effort!

PR #509, opened 2026-04-01
feat(chat): MCP tool call loop, engine status timeline, and UI polish

rayone:
## Problem

The built-in chat UI (`/admin/chat`) silently dropped responses when a model decided to call an MCP tool. `streamResponse` only handled `delta.content` and `delta.reasoning_content` — on `finish_reason: tool_calls` there was no content delta, so `streamingContent` stayed empty and nothing was pushed to the message list.

### MCP tool call loop
- Accumulate `delta.tool_calls` chunks during streaming (OpenAI chunked format, keyed by `index`)
- On `finish_reason: tool_calls`: push a hidden `assistant` message with `tool_calls`, execute all tools **in parallel** via `POST /v1/mcp/execute`, push hidden `tool` result messages, then recurse into `streamResponse` so the model receives full tool context and streams a final answer
- Depth limit via `MAX_TOOL_DEPTH = 10` constant; abort guard before recursion respects user clicking Stop
- 30s per-tool timeout (`AbortSignal.timeout` + `AbortSignal.any`), HTTP error detection on non-2xx responses
- Bookkeeping `assistant`/`tool` protocol messages hidden from display via `_ui: false`
- Simplified `messagesForApi` to a single filter+map pass that preserves `tool_calls` and `tool_call_id` fields

### Engine status timeline
- Prefill progress poller (polls `/admin/api/stats` every 500ms): shows `Prefilling N%` → `Generating`
- Per-tool timing logged on completion/failure with check/error icons and elapsed time
- Collapsible timeline panel (activity icon toggle in message header) — entries persist on the completed message and can be expanded post-completion
- Live status line (`Starting` / `Prefilling` / `Generating` / `Calling tool…`) with animated dots, rendered below the timeline entries

### Reasoning bubble
- `reasoning_content` deltas accumulate live in a collapsible `Thinking…` bubble rendered as markdown
- Collapses to `Thinking` (no ellipsis) once final content arrives

### Bug fixes
- Response box no longer disappears on reasoning-only or empty responses — always push an assistant message on stream completion
- Timeline toggle works consistently after content arrives — removed `!finalContent` guard that blocked the panel
- Alpine reactivity for `_perfLog`: `messages.splice(i, 1, { ...msg, ...patch })` instead of direct property assignment so Alpine detects the change and re-renders the toggle button

## Files changed

| File | Change |
|---|---|
| `omlx/admin/templates/chat.html` | MCP loop, status timeline, thinking bubble, bug fixes |
| `tests/test_chat_tool_call.py` | New file — 18 tests |

## Testing

```
pytest tests/test_chat_tool_call.py -v   # 18 passed
```

Manually verified with Qwen3.5-9B-MLX-4bit + simultaneous filesystem and Tavily MCP servers on macOS (Apple Silicon).

## Checklist

- [x] Feature branch from `main`
- [x] Tests added and passing
- [x] `SPDX-License-Identifier: Apache-2.0` present in new Python file
- [x] Code style matches project conventions
- [x] `pyproject.toml` unchanged — no new dependencies

### DIFF
--- a/omlx/admin/templates/chat.html
+++ b/omlx/admin/templates/chat.html
@@ -468,7 +468,14 @@
                 </div>
             </div>
 
-            <div class="w-10"></div>
+            <button @click="systemPromptDraft = systemPrompt; showSystemPromptModal = true"
+                    class="p-2 hover:bg-surface-muted rounded-lg relative"
+                    style="color: var(--text-primary);"
+                    :title="window.t('chat.system_prompt.title')">
+                <i data-lucide="settings" class="w-5 h-5"></i>
+                <span x-show="systemPrompt.trim()" x-cloak
+                      class="absolute top-1 right-1 w-2 h-2 bg-accent rounded-full"></span>
+            </button>
         </header>
 
         <!-- Messages Area -->
@@ -492,7 +499,7 @@
             <!-- Messages -->
             <div x-show="messages.length > 0" class="max-w-4xl mx-auto px-4 py-6 space-y-4">
                 <template x-for="(msg, index) in messages" :key="index">
-                    <div class="message-fade-in" x-show="msg._ui !== false">
+                    <div class="message-fade-in">
                         <!-- User Message -->
                         <div x-show="msg.role === 'user'" class="flex justify-end">
                             <div class="user-message">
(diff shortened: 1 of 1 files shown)


jundot:
Nice work on this. The MCP tool loop, status timeline, and thinking bubble are all solid additions. I verified the endpoint contracts (`/v1/mcp/tools`, `/v1/mcp/execute`, `//admin/api/stats`) and everything lines up. Security-wise, all new `x-html` bindings properly go through `DOMPurify.sanitize(marked.parse())` so no XSS concerns.

I found a couple of small bugs that i'll fix in a follow-up commit after merging:

1. **Hidden messages visible in UI**: assistant messages with `_ui: false` (the tool_calls bookkeeping turns) still show as empty bubbles because the template only checks `msg.role === 'assistant'`. Adding `&& msg._ui !== false` to the `x-show` on line 555 fixes it.

2. **Stray `</think>` on abort**: the abort handler still appends `</think>` to `streamingContent`, but reasoning now accumulates in `streamingThinking` separately. If a user clicks Stop during thinking, a literal `</think>` can appear in the visible content.

Nothing blocking, merging now.

PR #512, opened 2026-04-01
feat: support turboquant_plus

heroims:


SirDominik:
I gave this a try. Unfortunately, just like the previously disabled KV cache quantization, I'm not seeing any reduction in peak memory usage. On top of that, token generation speed dropped by about 5–8% in my testing — though I should note that my tests were fairly short.

PR #565, opened 2026-04-04
feat: Gemma 4 reasoning parser and agentic tool calling

TipKnuckle:
Adds generic output parser sessions and complete Gemma 4 reasoning + agentic tool calling support. Supersedes #561.

---

## Reasoning parser

Refactors the scheduler to support pluggable per-model output parser sessions. First implementation handles Gemma 4's `<|channel>thought` reasoning channel — strips it from streamed output and re-emits as `<think>…</think>` in `reasoning_content`.

- `omlx/adapter/output_parser.py` (new): `detect_output_parser()` factory
- `omlx/adapter/gemma4.py` (new): `Gemma4OutputParserSession`
- `omlx/scheduler.py`: threads parser session through generation loop
- `omlx/utils/tokenizer.py`: helpers for the session

## Tool call output parsing

`_inject_tool_calling` was using `mlx_lm._infer_tool_parser` which has no knowledge of Gemma 4's `<|tool_call>` marker — it returned `None`, `has_tool_calling` was never set, and raw markup leaked into response content. The pinned mlx-vlm (`43b9b20`) ships `mlx_vlm.tool_parsers` as a superset that adds `<|tool_call>` → `gemma4` detection and the correct parser. `_inject_tool_calling` now prefers this, falling back to the mlx_lm path for older installs.

## Tool result ingestion (message extractor pattern)

The Gemma 4 chat template has no handling for `role=tool` messages. Tool results must appear on a model-role turn as `tool_responses`:

```python
{"role": "assistant", "tool_responses": [{"name": fn_name, "response": {...}}]}
```

Passing raw `role=tool` messages caused the template to emit `<|tool_response>` literals in content and halt after the first tool call. Two secondary bugs: `_merge_consecutive_roles` was collapsing the `tool_calls` and `tool_responses` turns into one (fixed with `_PRESERVE_BOUNDARY_KEY`); `_drop_void_assistant_messages` was stripping the `tool_responses` turn (fixed with a `tool_responses` guard).

Design mirrors `detect_output_parser` — model-specific logic stays out of `server.py`:
- `detect_message_extractor()` in `output_parser.py` returns the right extractor callable
- `BatchedEngine` / `VLMBatchedEngine` expose it as a `message_extractor` property
- `server.py` calls `getattr(engine, "message_extractor", None)` — no model-type chains
- All Gemma 4 logic lives in `omlx/adapter/gemma4.py`

## Tests

| File | Coverage |
|---|---|
| `tests/test_gemma4_messages.py` | 11 tests: message conversion, tool result folding, name resolution, multi-turn agentic loops |
| `tests/test_output_parser.py` | Reasoning parser session |
| `tests/test_scheduler.py` | Scheduler integration |
| `tests/test_utils_tokenizer.py` | Tokenizer helpers |

### DIFF
--- a/omlx/adapter/gemma4.py
+++ b/omlx/adapter/gemma4.py
@@ -1,0 +1,380 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Gemma 4 reasoning-channel output parsing and message extraction."""
+
+from __future__ import annotations
+
+import json
+from typing import Any, List
+
+try:
+    from mlx_lm.tokenizer_utils import NaiveStreamingDetokenizer
+except ImportError:
+    NaiveStreamingDetokenizer = None
+
+from ..api.utils import _PRESERVE_BOUNDARY_KEY
+from .output_parser import OutputParserFinalizeResult, OutputParserTokenResult
+
+_OPEN_MARKER = "<|channel>thought\n"
+_CLOSE_MARKER = "<channel|>"
+_TURN_END_MARKER = "<turn|>"
+_TOOL_RESPONSE_OPEN = "<|tool_response>"
+_TOOL_RESPONSE_CLOSE = "<tool_response|>"
+_THINK_OPEN = "<think>\n"
+_THINK_CLOSE = "</think>\n"
+
+
+def _try_parse_json(s: str) -> Any:
+    """Parse string as JSON if possible, otherwise return as-is."""
+    if not isinstance(s, str):
+        return s
+    s = s.strip()
+    if not s or not (s.startswith("{") or s.startswith("[")):
+        return s
+    try:
+        return json.loads(s)
+    except (json.JSONDecodeError, ValueError):
+        return s
+
+
+def extract_gemma4_messages(
+    messages: List[Any],
+    max_tool_result_tokens: int | None = None,
+    tokenizer: Any | None = None,
+) -> List[dict]:
+    """Convert OpenAI-format messages to Gemma 4 chat-template format.
+
+    The Gemma 4 chat template does not handle ``role=tool`` messages.
+    Tool results must instead appear on a model-role turn as a
+    ``tool_responses`` list, where each entry is::
+
+        {"name": "<function_name>", "response": <dict_or_scalar>}
+
+    This function:
+    - Passes non-tool messages through unchanged.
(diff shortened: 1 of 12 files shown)


TipKnuckle:
Note there is an underlying bug in mlx-vlm that I discovered, so some cases of tool use are broken even with this implementation. See https://github.com/Blaizzy/mlx-vlm/issues/914

TipKnuckle:
Note on earlier comment: https://github.com/Blaizzy/mlx-vlm/issues/914 has been fixed, and tool calls are being properly parsed from mlx-vlm in commit b8c0c5d.

jundot:
@TipKnuckle I've updated mlx-lm and mlx-vlm to the latest versions, which I expect will resolve many of the existing issues. I'm currently working through the related changes needed on the omlx side. Please allow me some time - will follow up here soon!

TipKnuckle:
@jundot Thanks, I realize this is an architectural change, but I do think it's the right direction to handle custom parsing like gemma4 has introduced. SOLAR is another model that would benefit from this (though would require its own parser as well).

Updating the mlx-lm and mlx-vlm will unfortunately _not_ help with Gemma4 tool calling with your current main:

### Upstream main (omlx/engine/vlm.py):
- from mlx_lm.tokenizer_utils import _infer_tool_parser
- Only uses mlx-lm's tool parser detection.

### This PR (omlx/engine/vlm.py):
- Prefer mlx_vlm.tool_parsers (superset; knows about Gemma4 etc.)
from mlx_vlm.tool_parsers import _infer_tool_parser, load_tool_module
Tries mlx-vlm first (which has Gemma 4 <|tool_call> support), falls back to mlx-lm.

### The problem: 
- mlx-lm's _infer_tool_parser has no knowledge of Gemma 4's <|tool_call> marker
- It returns None for Gemma 4 → has_tool_calling never set → raw markup leaks into response
- mlx-vlm's tool_parsers module is a superset that includes gemma4 detection

The latest mlx-lm commit (4469ad4 - "Add gemma 4") doesn't add tool parser support - it's model architecture support, not tool calling. So even updating mlx-lm to latest, Gemma 4 tool calling still won't work without the mlx-vlm tool parser change.

TipKnuckle:
Great, thanks!

Issue #571, opened 2026-04-04
Brew Formula for 0.3.2 not up to date


BastianKusserow:
**Describe the bug**
Just tried to update to the newest version of 0.3.2 via homebrew, but it seems as if the brew formula of 0.3.2 is not up to date.

**To Reproduce**
1. Run brew update && brew upgrade omlx
2. Brew tells me that the current version of 0.3.1 is the newest version available

**Expected behavior**
Would expect that brew upgrades omlx to 0.3.2

**Desktop (please complete the following information):**
 - OS: macOS 26.4
 - Browser chrome

jundot:
updated the formula to v0.3.2, sorry for the delay. i think this part needs to be automated going forward, will look into it.

PR #574, opened 2026-04-04
chore: update homebrew formula to 0.3.2

luiz290788:
Updates the Homebrew formula to point to the latest 0.3.2 release.


luiz290788:
I wonder if we can automate this process somehow to avoid this happening again in the future.

jundot:
Hey, saw this after i already pushed the formula update from #571. Sorry about that and thanks for the PR!

You're right, this should be automated so it doesn't happen again. I'll look into setting that up.

Closing this one since the change is already in main.

PR #588, opened 2026-04-05
feat: Add Pi integration

latent-variable:
## Summary
- add a dedicated `omlx launch pi` integration that configures Pi against the local oMLX OpenAI-compatible endpoint
- expose Pi in the integration registry, CLI help, and admin dashboard integration settings
- add focused integration tests for Pi config generation

## Why Pi
- [Pi](https://pi.dev/) is widely used and gives oMLX coverage for a lightweight terminal harness with strong local-first workflows
- its AGENTS, skills, prompt templates, and dynamic context model make it a strong candidate for cache-friendly follow-up turns compared with heavier harness patterns
- it fits oMLX well for private local usage because it can be pointed at the local OpenAI-compatible endpoint without extra middleware

## Testing
- `python -m compileall omlx`
- smoke test for `PiIntegration.configure()` writing Pi `models.json` and `settings.json`


AlexWorland:
@latent-variable I took a quick look and am super excited for this! Loving using Pi with Ollama and I think it would make a great addition to omlx. 

Found a few things on review that may be worth taking a look at:

  Style / consistency:

  - `launch()` uses `kwargs.get()` but the codebase convention (see `OpenCodeIntegration.launch()`) is `kwargs.pop()`:
  `context_window = kwargs.pop("context_window", None)`
`max_tokens = kwargs.pop("max_tokens", None)`
  `model_type = kwargs.pop("model_type", None)`
  - `base.py:19` comment still says `# "codex", "opencode", "openclaw"` but should probably include "pi"

  Test coverage gaps — every other integration has these, Pi is missing them:

  - `test_type (asserts type == "config_file" and display_name == "Pi")`
  - `test_configure_custom_host (non-default host in config output)`
  - `test_configure_creates_backup (.bak file created on overwrite)`
  - `test_get_command_no_model (empty model falls back to select-a-model)`

0xClandestine:
this could easily be a pi extension, i have one i personally use happy to drop it in a gist if you're interested

latent-variable:
@AlexWorland thanks for taking the time to review this. i addressed those points and added the missing pi tests. if you take another look and spot anything else, i’d really appreciate it.

latent-variable:
@0xClandestine yeah, that makes sense too. i went with the omlx-side integration since it matches the existing `omlx launch <tool>` flow, but if this doesn’t get included i’d definitely love to see your pi extension approach.

0xClandestine:
> @0xClandestine yeah, that makes sense too. i went with the omlx-side integration since it matches the existing `omlx launch <tool>` flow, but if this doesn’t get included i’d definitely love to see your pi extension approach.

https://gist.github.com/0xClandestine/106d600ce0fb97ec32de719438ff4b83 🤝

jundot:
Thanks for the contribution, looks good. Merged.

One thing i'll fix in a follow-up: the Pi description in the dashboard template is hardcoded as `"Terminal AI coding assistant"` instead of using an i18n key like `{{ t('status.integrations.pi_desc') }}`. The other integrations (OpenCode, OpenClaw) use translated strings, so i'll add the missing locale keys and swap it out.

PR #607, opened 2026-04-05
feat(admin): populate reasoning parser dropdown dynamically from xgrammar

leuski:
Replace the hardcoded reasoning parser <option> list in the model settings modal with a runtime query to `xgrammar.get_builtin_structural_tag_supported_models()`.

New parsers (e.g. glm47 added in xgrammar 0.1.33) appear automatically after `pip install --upgrade xgrammar` — no code changes or server restart needed. Each dropdown option shows the parser name plus supported model families for easy identification (e.g. "llama (Meta-Llama-3, Llama-3.1, Llama-3.2, Llama-4)")

- Add GET /admin/api/grammar/parsers endpoint
- Fetch parser list in openModelSettings() and store as reactive Alpine data
- Render dropdown options with model family hints (e.g. "qwen (Qwen3)")

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -1244,6 +1244,30 @@ async def delete_sub_key(
 
 
 # =============================================================================
+# Grammar API Routes
+# =============================================================================
+
+
+@router.get("/api/grammar/parsers")
+async def list_grammar_parsers(is_admin: bool = Depends(require_admin)):
+    """Return available reasoning parser names from xgrammar.
+
+    Queries ``xgrammar.get_builtin_structural_tag_supported_models()`` at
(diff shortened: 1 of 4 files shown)


jundot:
Nice work, this is a solid improvement. Having the parser list auto-sync with the installed xgrammar version removes a maintenance burden i've been meaning to address.

I'll merge this and follow up with a small commit to add `pytest.importorskip("xgrammar")` to `test_returns_list_from_xgrammar` so it doesn't break in environments where xgrammar isn't installed. I'll also add 401 handling to the fetch call to match the pattern used elsewhere in dashboard.js.

Thanks for the contribution!

PR #614, opened 2026-04-06
fix: parse JSON output in /v1/responses endpoint

### LINKED ISSUE #613: /v1/responses endpoint returns raw markdown-wrapped JSON instead of parsed JSON
## Bug

The `/v1/responses` endpoint doesn't call `parse_json_output()` on the model output when `response_format` is specified. This means JSON wrapped in markdown code blocks (e.g. `` ```json ... ``` ``) is returned verbatim to the client.

The `/v1/chat/completions` endpoint already handles this correctly at `omlx/server.py:2051-2061`.

## Impact

Most visible when xgrammar is unavailable (e.g. DMG builds). Without logit-level grammar enforcement, models like Gemma-4 wrap JSON responses in markdown code blocks. Clients calling `/v1/responses` with `json_schema` format receive the raw markdown-wrapped text instead of clean JSON.

Models that natively comply with JSON schema (e.g. Qwen) are unaffected since their output parses directly.

## Reproduction

1. Run omlx without xgrammar (DMG build or no torch)
2. Call `/v1/responses` with `text.format.type = "json_schema"` targeting Gemma-4
3. Observe response text contains `` ```json ... ``` `` wrapping

TipKnuckle:
## Summary

The `/v1/responses` endpoint was missing the `parse_json_output()` post-processing step that `/v1/chat/completions` already performs. When `response_format` is specified and the model wraps JSON in markdown code blocks (common when xgrammar is unavailable), the raw markdown was returned to the client instead of parsed JSON.

## Changes

- Added `parse_json_output()` call in `create_response()` (non-streaming `/v1/responses`) between tool call extraction and output building, matching the existing pattern in `create_chat_completion()`.

Closes #613

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -3572,6 +3572,17 @@ async def create_response(
         cleaned_text = extraction.cleaned_text
         tool_calls = extraction.tool_calls
 
+    # Process response_format if specified
+    if response_format and not tool_calls:
+        cleaned_text, parsed_json, is_valid, error = parse_json_output(
+            cleaned_text or regular_content,
+            response_format
+        )
+        if parsed_json is not None:
+            cleaned_text = json.dumps(parsed_json)
+        if not is_valid:
+            logger.warning(f"JSON validation failed: {error}")
+
     # Build output items
     output_items: list[OutputItem] = []
     output_items.append(


DiegoRBaquero (approved):

jundot (approved):
Looks good, clean fix.

I verified the insertion point is correct (after tool call extraction, before output building) and all variables are in scope. The pattern is identical to what `create_chat_completion()` already does at line 2138, so this should be safe.

One thing i noticed during review: the streaming counterpart (`stream_responses_api()`) has the same gap. I'll add `parse_json_output()` there too in a follow-up commit after merging this.

Thanks for the fix @TipKnuckle!

jundot:
Looks like this has a merge conflict with the current main. Since the fix is straightforward and i also need to add the same parsing to the streaming paths, i went ahead and applied the change directly on main along with:

- `parse_json_output()` for `stream_responses_api()` done events
- `parse_json_output()` for `stream_chat_completion()` buffered mode
- Integration tests covering both endpoints

Your fix is included with co-author credit. Closing this PR since the changes are now on main. Thanks again @TipKnuckle!

PR #624, opened 2026-04-06
style(ui): update `Preferences...` to `Settings…` for macOS consistency

Stv-X:
This PR updates the Mac app’s UI strings and internal code comments to use `Settings…` instead of `Preferences...`. This change ensures the app remains consistent with the modern macOS design changes introduced in macOS Ventura 13.

Key Changes
* UI Strings: Updated menu items and labels from "Preferences..." to "Settings…".
* Typography: Corrected the ellipsis from three periods (...) to the standard horizontal ellipsis character (…) as per Apple's Human Interface Guidelines.

[!NOTE]
This is a UI-only change. To maintain backward compatibility and minimize risk, no class names, method signatures, or underlying logic have been modified.

Rationale
1. System Consistency: Since macOS Ventura 13, Apple transitioned "System Preferences" to "System Settings." Aligning with this change unifies the experience across iOS, iPadOS, and macOS, providing a cohesive vocabulary across the Apple ecosystem, and reduces cognitive load for users switching between their devices.
2. AppKit Behavior: By default, AppKit now automatically overrides "Preferences" with "Settings" in the App menu unless explicitly disabled via the `NSMenuShouldUpdateSettingsTitle` key. Adopting this change natively in our strings brings us in line with standard OS behavior.
3. Typographic Precision: Using the proper Unicode ellipsis (…) follows professional typesetting standards and Apple's own localization practices.

While "Preferences" has been a staple of the Mac experience for decades, embracing "Settings" respects the current design evolution of the platform and ensures a modern feel for our users.

### DIFF
--- a/packaging/omlx_app/app.py
+++ b/packaging/omlx_app/app.py
@@ -719,9 +719,9 @@ class OMLXAppDelegate(NSObject):
 
         self.menu.addItem_(NSMenuItem.separatorItem())
 
-        # --- Preferences ---
+        # --- Settings ---
         prefs_item = NSMenuItem.alloc().initWithTitle_action_keyEquivalent_(
-            "Preferences...", "openPreferences:", ","
+            "Settings…", "openPreferences:", ","
         )
         prefs_item.setTarget_(self)
         prefs_icon = self._create_menu_icon("gearshape")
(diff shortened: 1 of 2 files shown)


jundot (approved):
Clean and straightforward change, looks good to me. The ellipsis fix is a nice touch too.

Thanks for the contribution!

Issue #435, opened 2026-03-27
Kernel panic (IOGPUMemory completeMemory() prepare count underflow) after upgrading v0.2.20 → v0.2.23


gizmax:
### Summary

Upgrading oMLX from v0.2.20 to v0.2.23 causes repeated kernel panics on Mac Studio M2 Ultra 64GB. Rolling back to v0.2.20 immediately resolves the issue. 4 identical panics occurred within ~1 hour of running v0.2.23.

### Environment

- **Hardware**: Mac Studio M2 Ultra, 64GB unified memory
- **macOS**: 26.4 (Build 25E5233c), Darwin 25.4.0
- **oMLX**: v0.2.23 (panics) / v0.2.20 (stable)
- **Models loaded**: Qwen3.5-35B-A3B-4bit (pinned, ~20GB) + Qwen3.5-0.8B-4bit (SpecPrefill draft)
- **Config**: `--max-process-memory 80% --hot-cache-max-size 8GB --paged-ssd-cache-dir ~/.omlx/cache`
- **SSD cache**: ~92GB (1163 files)

### Panic details

All 4 panics are identical:

```
panic(cpu 18 caller 0xfffffe00427725d8): "completeMemory() prepare count underflow" @IOGPUMemory.cpp:550
```

Timestamps: 21:11, 21:29, 21:40, 21:48 (2026-03-27, ~10-20 min intervals)

### Reproduction steps

1. Running oMLX v0.2.20 stably for days (no panics)
2. `pip install "omlx @ git+https://github.com/jundot/omlx.git@v0.2.23"`
3. `launchctl kickstart -k gui/502/ai.jarvis.omlx`
4. oMLX starts normally, models load, inference works
5. Within 10-20 minutes: kernel panic
6. After reboot, panics repeat on each boot until rollback

### What v0.2.23 changed (from source diff)

Comparing v0.2.20 and v0.2.23, the relevant changes are:

1. **New files**: `patches/turboquant_attention.py`, `turboquant_kv.py` (TurboQuant KV cache, disabled via hardcoded `turboquant_kv_enabled = False`)
2. **`cache/prefix_cache.py`**: Changed offset calculation to always use tensor shape instead of meta_state. Added TurboQuantKVCache handling with block slicing.
3. **`cache/paged_ssd_cache.py`**: Added TurboQuant tensor serialization, disk pressure handling (ENOSPC/EDQUOT)
[...]

PR #648, opened 2026-04-07
fix: prevent IOKit prepare count underflow with concurrent completions (#557)

### LINKED ISSUE #557: Kernel panic: completeMemory() prepare count underflow on M1 Max 32GB with v0.3.0
## Environment
- oMLX version: 0.3.0
- Mac: Mac Studio M1 Max 32GB
- macOS: Tahoe(26.3)
- Model: MLX-Qwen3.5-27B-Claude-4.6-Opus-Reasoning-Distilled-v2-4bit (14.8GB)
- Client: OpenCode + oh-my-openagent (Sisyphus multi-agent)

## Description
Experiencing kernel panics causing full machine reboots when running multi-agent coding tasks with OpenCode/Sisyphus. This happened twice on 2026-04-03 (panic logs attached).

The panic matches the same `completeMemory() prepare count underflow` error fixed in v0.2.24 (#435), but it is still reproducible on v0.3.0

[panic-full-2026-04-03-235119.0002.txt](https://github.com/user-attachments/files/26462906/panic-full-2026-04-03-235119.0002.txt)

.

## Panic message
panic: "completeMemory() prepare count underflow" @IOGPUMemory.cpp:550

## Relevant threads from panic log
- IOGPU Completion (TH_WAIT, TH_UNINT)
- AGXHardwareWorkLoop
- IOSurfaceRoot
- VM_pressure / VM_swapout / VM_compressor active simultaneously

## Reproduction scenario
1. Load MLX-Qwen3.5-27B-Claude-4.6-Opus-Reasoning-Distilled-v2-4bit (14.8GB, pinned)
2. Run OpenCode with Sisyphus agent on a large coding task (reading multiple files, generating code)
3. Sisyphus sends 2 concurrent requests to oMLX
4. After several minutes of prefill/generation, machine kernel panics and reboots

## oMLX settings at time of panic
- max_num_seqs: 2
- SSD paged cache enabled (92GB)
- Hot cache: 4GB
- prefill_memory_guard: enabled

## Notes
- v0.2.24 claimed to fix this via deferred Metal buffer clearing (#435)
- The fix appears to not be fully effective under concurrent multi-agent load
- Setting iogpu.wired_limit_mb=26000 as a workaround but not confirmed effective yet

## Panic log files
[attach panic-full-2026-04-03-235119.0002.panic]

Chuhan1112:
Fixes #557

Two separate bugs could each independently trigger the `'completeMemory() prepare count underflow'` kernel panic on Apple Silicon.

## Bug 1 — deferred-clear window not extended for concurrent completions

The deferred-clear mechanism introduced in #435 used a countdown integer (`_deferred_clear_steps`) anchored to the *first* request that finished in a batch. The scheduling code guarded the assignment with `"only set if None"`, so when a second request finished in the same or immediately following step (as happens with `max_num_seqs=2` and the Sisyphus multi-agent workload in #557), its completion was silently ignored.

The first request's KV cache blocks were returned to the free queue immediately on completion and could be re-allocated to the next incoming request within one generation step — long before IOKit had finished the `completeMemory()` callbacks for those buffers. The new allocation called `prepare()` on the same Metal buffer, and when the stale callback finally ran, `completeMemory()` decremented the count below zero → kernel panic.

**Fix:** replace the countdown integer with an absolute target step (`_deferred_clear_at`). Each completion computes `target = _step_counter + _DEFERRED_CLEAR_DELAY` and updates `_deferred_clear_at` via `max()`, so every completion gets a full `_DEFERRED_CLEAR_DELAY` window measured from *its own* finish step, regardless of whether another deferred clear was already pending.

## Bug 2 — bare `mx.clear_cache()` calls in SpecPrefill path

Two sites in the SpecPrefill code path called `mx.clear_cache()` directly instead of going through the `_sync_and_clear_cache()` helper:

- The inter-chunk clear inside the system-prompt full-prefill loop
- The draft-cache teardown after SpecPrefill scoring

Both `mx.eval()` calls preceding them submit work onto the generation stream asynchronously. Calling `mx.clear_cache()` without first calling `mx.synchronize(generation_stream)` can release Metal buffers that are still referenced by in-flight command buffers on that stream — the same race that `_sync_and_clear_cache()` was introduced to prevent in #435.

**Fix:** replace both bare `mx.clear_cache()` calls with `_sync_and_clear_cache()`.

## Reproduction environment from #557
- M1 Max Mac Studio, 32 GB RAM, macOS Tahoe 26.3
- oMLX v0.3.0 (also confirmed on v0.3.4 by Chuhan1112)
- Model: MLX-Qwen3.5-27B (14.8 GB, pinned memory)
- `max_num_seqs: 2`, SSD paging 92 GB, `prefill_memory_guard: true`
- Workload: OpenCode + Sisyphus multi-agent framework

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -553,8 +553,17 @@ class Scheduler:
         # IOKit's asynchronous completeMemory() callbacks, causing
         # 'prepare count underflow' kernel panics. Deferring the clear
         # by a few generation steps gives IOKit time to process callbacks.
-        # None = no deferred clear pending; int = steps since last finish.
-        self._deferred_clear_steps: Optional[int] = None
+        #
+        # Stored as the absolute step number at which the clear should fire,
+        # rather than a countdown integer.  This avoids the burst-completion
+        # bug (#557): with max_num_seqs > 1 two requests can finish in the
+        # same batch.  The old "only set if None" guard meant the second
+        # completion never extended the window, so the first request's KV
+        # cache blocks could be re-allocated before IOKit finished its
+        # completeMemory() callbacks.  Using max() ensures the window always
+        # covers the *latest* completion.
+        # None = no deferred clear pending; int = step at which to fire.
+        self._deferred_clear_at: Optional[int] = None
 
         # Cache XTC special tokens (newline + EOS) — stable per tokenizer.
         # Must be after _is_harmony_model / _generation_config_eos init
@@ -2412,9 +2421,12 @@ class Scheduler:
                 except Exception as e:
                     logger.debug(f"SpecPrefill: draft cache store failed: {e}")
 
-            # Free draft cache from memory
+            # Free draft cache from memory.  Use _sync_and_clear_cache() so
+            # the generation_stream is drained before Metal buffers are
+            # returned to the pool — a bare mx.clear_cache() here can race
+            # with in-flight async evals and trigger a kernel panic (#557).
             del used_cache
-            mx.clear_cache()
+            _sync_and_clear_cache()
 
         except Exception as e:
             logger.error(f"SpecPrefill scoring failed, falling back to normal path: {e}")
@@ -2606,11 +2618,11 @@ class Scheduler:
 
         Also returns True when a deferred Metal cache clear is pending,
         so that the engine loop keeps calling step() until the clear fires.
-        Without this, an idle server would never increment the deferred
-        counter and stale buffers would accumulate indefinitely.
+        Without this, an idle server would never reach the target step and
+        stale buffers would accumulate indefinitely.
         """
         return bool(self.waiting or self.running
-                     or self._deferred_clear_steps is not None)
+                     or self._deferred_clear_at is not None)
 
     def fail_all_requests(self) -> List[str]:
         """Remove all running and waiting requests after unrecoverable error.
@@ -2907,7 +2919,14 @@ class Scheduler:
                             self.model(sys_arr[:step][None], cache=sp_cache)
                             mx.eval([c.state for c in sp_cache])
                             sys_arr = sys_arr[step:]
-                            mx.clear_cache()
+                            # Use _sync_and_clear_cache() instead of bare
+                            # mx.clear_cache() to flush the generation_stream
+                            # before releasing Metal buffers.  A bare call here
+                            # can race with in-flight command buffers submitted
+                            # by the preceding mx.eval(), triggering the same
+                            # 'completeMemory() prepare count underflow' kernel
+                            # panic that #435 fixed elsewhere (#557).
+                            _sync_and_clear_cache()
                         if sys_arr.size > 0:
                             self.model(sys_arr[None], cache=sp_cache)
                             mx.eval([c.state for c in sp_cache])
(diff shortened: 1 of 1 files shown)


mdevk (commented):
This is critical for anyone running concurrent VLM inference on Apple Silicon.

## Relevant data from our benchmarking

We ran concurrency sweeps (1, 2, 4, 8, 16) on oMLX v0.3.4 with Qwen3-VL-30B-A3B:

| Concurrency | JSON Valid | Note |
|---|---|---|
| 1 | 100% (20/20) | Baseline |
| 2 | 100% (20/20) | Clean |
| 4 | 100% (20/20) | Clean |
| 8 | 95% (19/20) | First quality drop |
| 16 | 40% (8/20) | Severe degradation |

We also observed that on v0.3.5.dev1 (pre-release), concurrency=2 dropped JSON validity to 20% — much worse than v0.3.4 stable. This could be related to the deferred-clear window regression you describe.

We never hit the kernel panic itself, but the quality degradation at high concurrency could be an early symptom of the same buffer reuse-before-completion race. If IOKit hasn't finished `completeMemory()` callbacks and the buffer gets reallocated, the model reads stale/corrupted KV cache data → garbage output → invalid JSON.

## The fix looks correct

The `max()` approach for `_deferred_clear_at` elegantly handles overlapping completions. Each completion extends the window rather than resetting it. Clean design.

The SpecPrefill `mx.synchronize()` before `mx.clear_cache()` is clearly necessary — asynchronous stream semantics make bare `clear_cache()` a race condition by design.

mdevk:
**Follow-up: tested the patch locally, no quality improvement observed**

To be transparent — we applied this patch to our local oMLX v0.3.4, restarted, and re-ran the same concurrency sweeps (Qwen3-VL-30B-A3B, structured JSON, conc 1/2/4/8/16).

| Concurrency | JSON Valid (before) | JSON Valid (after) |
|---|---|---|
| 1 | 100% (20/20) | 100% (20/20) |
| 2 | 100% (20/20) | 100% (20/20) |
| 4 | 100% (20/20) | 100% (20/20) |
| 8 | 95% (19/20) | 95% (19/20) |
| 16 | 40% (8/20) | 40% (8/20) |

No measurable change. The quality degradation at conc=8+ persists with the patch.

The PR correctly fixes the IOKit underflow crash described in #557 — the `max()` approach and `mx.synchronize()` addition both look right. Our quality issue at high concurrency is a different root cause (likely KV cache contention or scheduler starvation, not the deferred-clear race). We should have been clearer in our initial review.

mdevk:
**Correction to our earlier data:** After more rigorous testing (3 runs × 20 photos per concurrency level), we found that our original 100% validity claim at conc=2 was wrong. Actual results on unpatched v0.3.4:

- conc=2: **45-65%** validity (3 runs: 65%, 60%, 45%)
- conc=4: 100%
- conc=8: 100%
- conc=16: 70%

Filed as #691 with full details. The concurrency quality issue is NOT monotonic — conc=2 is worse than conc=4/8, suggesting a specific scheduling edge case rather than pure resource contention.

Our earlier measurement likely had different conditions. Apologies for the inaccurate data.

jundot (approved):
The fix itself looks correct and well thought out. The `max()` approach for `_deferred_clear_at` cleanly handles overlapping completions without the indefinite postponement problem the old comment warned about. Each completion gets a full delay window measured from its own finish step, which is exactly the right behavior.

Replacing the bare `mx.clear_cache()` calls in the specprefill path with `_sync_and_clear_cache()` is also clearly necessary since async stream semantics make the unsynchronized version a race condition by design.

3 existing tests reference the renamed `_deferred_clear_steps` attribute and need updating. I'll handle that plus add concurrent completion test coverage in a follow-up commit after merge.

PR #652, opened 2026-04-07
Fix VLM SpecPrefill threshold forwarding

apetersson:
## Summary
- forward `specprefill_threshold` through the VLM chat path into `add_request()`
- keep the existing `specprefill` and `specprefill_keep_pct` passthrough behavior intact
- add a regression test covering VLM SpecPrefill threshold and keep-rate forwarding

## Context
VLM-backed models were ignoring explicit SpecPrefill threshold overrides because `VLMBatchedEngine.stream_generate()` forwarded `specprefill` and `specprefill_keep_pct` but dropped `specprefill_threshold`.

This made Qwen VLM benchmarks silently fall back to the default threshold instead of using the configured value.

## Testing
- `pytest -q tests/test_vlm_specprefill.py`

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -1246,6 +1246,8 @@ class VLMBatchedEngine(BaseEngine):
             specprefill_kwargs["specprefill"] = kwargs.pop("specprefill")
         if kwargs.get("specprefill_keep_pct") is not None:
             specprefill_kwargs["specprefill_keep_pct"] = kwargs.pop("specprefill_keep_pct")
+        if kwargs.get("specprefill_threshold") is not None:
+            specprefill_kwargs["specprefill_threshold"] = kwargs.pop("specprefill_threshold")
         if kwargs.get("specprefill_system_end") is not None:
             specprefill_kwargs["specprefill_system_end"] = kwargs.pop("specprefill_system_end")
 
--- a/tests/test_vlm_specprefill.py
+++ b/tests/test_vlm_specprefill.py
@@ -1,0 +1,53 @@
+"""Regression tests for SpecPrefill parameter forwarding in VLM engine."""
+
+from concurrent.futures import ThreadPoolExecutor
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
+from omlx.engine.vlm import VLMBatchedEngine
+
+
+@pytest.mark.asyncio
+async def test_vlm_chat_forwards_specprefill_threshold_and_keep_pct():
+    """VLM chat must pass both SpecPrefill overrides through to add_request()."""
+    engine = VLMBatchedEngine(model_name="test-vlm")
+    engine._loaded = True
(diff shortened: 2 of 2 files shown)


apetersson:
Benchmark note for `Qwen3.5-122B-A10B-5bit` after this fix.

Setup:
- target: `Qwen3.5-122B-A10B-5bit`
- draft: `Qwen3.5-2B-6bit` and `Qwen3.5-0.8B-5bit`
- `keep_pct=0.20`
- prompt length: about `8k`

Before the fix:
- VLM requests silently dropped `specprefill_threshold`, so the effective threshold stayed at the default `8192`.
- At ~8k prompt length, the sparse path often did not activate as intended, so the measured behavior did not reflect the configured threshold.

After the fix, with explicit thresholds forwarded correctly:
- Base: `8023` prompt tokens, `39.98s` TTFT, `200.69` prompt tok/s
- `2B` draft, `threshold=512`: `19.47s` TTFT, `412.06` prompt tok/s
- `2B` draft, `threshold=1024`: `20.77s` TTFT, `386.21` prompt tok/s
- `0.8B-5bit` draft, `threshold=512`: `15.87s` TTFT, `505.55` prompt tok/s

Observed improvement vs base:
- `2B`, `th512`: about `2.05x` faster TTFT / PP
- `2B`, `th1024`: about `1.92x`
- `0.8B-5bit`, `th512`: about `2.52x`

So the main effect of this patch is that Qwen VLM benchmarks now honor the configured SpecPrefill threshold, and the expected speedup becomes observable at ~8k prompt length.

note: on my machine 512 / 1024 seems to be a reasonable threshold to set for optimal results.

jundot:
Thanks for catching this. The threshold was the only specprefill param missing from the VLM path, while `batched.py` already had all four. Clean fix.

I verified the forwarding chain: `stream_chat` passes kwargs through to `stream_generate`, which now correctly pops `specprefill_threshold` and forwards it to `add_request`. Without this, VLM requests silently fell back to the default threshold (8192) regardless of what was configured.

Your benchmark numbers on Qwen3.5-122B look solid. The 0.8B draft at threshold=512 hitting 2.52x is a nice result.

truecraft:
Is this fix working correctly? After the 0.36 update from two days ago the threshold parameter still doesn't seem to work; setting it to 1024 for Qwen3.5-27B-oQ4 (with 0.8B-oQ6 draft) still results in no appreciable speedup in the Performance testing tab except on pp16384 and higher.

truecraft:
0.37rc2, this still isn't working. A freshly downloaded mlx-community/Qwen3.6-27b-mxfp4 with Jundot/Qwen3.6-35B-A3-oQ4 gets pp4096 200t/s, pp16384 ~475t/s.

With the threshold set to 1024, pp4096 is still 200t/s (no improvement). The threshold setting still seems to have no impact < 8096 (at least on my M3 Max).

PR #654, opened 2026-04-07
fix: handle TypeError from think_start_id for non-thinking models

### LINKED ISSUE #653: context-1-MLX-6bit crashes on every request: TypeError in think_start_id
## Description

context-1-MLX-6bit (GptOssForCausalLM / harmony parser) loads successfully but **crashes on every inference request** with a `TypeError` in `think_start_id`.

## Environment

- **oMLX version**: v0.3.5.dev1 (DMG / macOS app)
- **macOS**: 26.4 (Apple Silicon, 128GB)
- **Model**: context-1-MLX-6bit (MoE, 32 experts, 131K context)
- **Other models**: gemma-4-e4b-it-8bit works fine on the same server

## Steps to Reproduce

1. Start oMLX server with `context-1-MLX-6bit` available in model directory
2. Model loads successfully (16.63GB)
3. Send any chat completion request to context-1-MLX-6bit
4. Request hangs until client timeout — server logs show crash

## Error

```
omlx.engine_core - ERROR - Engine loop error: object of type 'NoneType' has no len()
  File ".../omlx/engine_core.py", line 175, in _engine_loop
    output = await loop.run_in_executor(...)
  File ".../omlx/scheduler.py", line 3539, in step
    scheduled, rejected = self._schedule_waiting()
  File ".../omlx/scheduler.py", line 2846, in _schedule_waiting
    if self._detect_needs_think_prefix(request):
  File ".../omlx/scheduler.py", line 1427, in _detect_needs_think_prefix
    think_start_id = getattr(self.tokenizer, 'think_start_id', None)
  File ".../mlx_lm/tokenizer_utils.py", line 400, in think_start_id
    if len(self._think_start_tokens) > 1:
TypeError: object of type 'NoneType' has no len()
```

## Root Cause

`tokenizer_utils.py:400` accesses `self._think_start_tokens` which is `None` for context-1's tokenizer (it uses a harmony parser, not a thinking-capable tokenizer). The `think_start_id` property doesn't guard against `None` before calling `len()`.

## Suggested Fix

In `tokenizer_utils.py`, line ~400:

```python
# Before:
if len(self._think_start_tokens) > 1:

# After:
if self._think_start_tokens and len(self._think_start_tokens) > 1:
```

Or in `scheduler.py:1427`, handle the `TypeError` from the property.

FaisalFehad:
## Summary

- Catch `TypeError` alongside `ValueError` in both `think_start_id` call sites in `scheduler.py`
- Models like **context-1-MLX-6bit** (harmony parser) have `_think_start_tokens=None` in their mlx_lm tokenizer, causing `think_start_id` to raise `TypeError` on `len(None)`
- `getattr(tokenizer, 'think_start_id', None)` doesn't help because the attribute exists — its getter raises internally

## Root cause

```python
# mlx_lm/tokenizer_utils.py
@property
def think_start_id(self):
    if len(self._think_start_tokens) > 1:  # TypeError: len(None)
```

`_infer_thinking()` returns `(None, None, None, None)` for models without thinking tokens, so `_think_start_tokens` is `None`.

## Fix

Add `TypeError` to the existing `except ValueError` in both scheduler call sites (lines 1266 and 1428). This is the minimal, non-breaking fix on the omlx side — no changes needed to mlx_lm.

## Test plan

- [x] Added test: `test_think_start_id_raises_type_error` — verifies `_detect_needs_think_prefix` returns `False` when the property raises `TypeError`
- [x] Verified context-1-MLX-6bit loads and generates correctly with this fix on v0.3.5.dev1

Fixes #653

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -1261,10 +1261,7 @@ class Scheduler:
             if think_end_ids:
                 from .api.thinking import ThinkingBudgetProcessor
 
-                try:
-                    think_start_id = getattr(self.tokenizer, 'think_start_id', None)
-                except ValueError:
-                    think_start_id = None
+                think_start_id = self._get_think_token_id('think_start_id')
                 leading_ids, trailing_ids = self._resolve_think_close_pattern()
                 processor = ThinkingBudgetProcessor(
                     think_end_token_ids=think_end_ids,
@@ -1303,6 +1300,21 @@ class Scheduler:
 
         return resolve_vocab_size(self.model)
 
+    def _get_think_token_id(self, attr: str) -> int | None:
+        """Safely read a think token id from the tokenizer.
+
+        mlx-lm tokenizers expose ``think_start_id`` / ``think_end_id`` as
+        properties that may raise ``ValueError`` (multi-token sequence) or
+        ``TypeError`` (``_think_start_tokens`` is ``None`` for models without
+        thinking support, e.g. context-1 / harmony parser).
+
+        Returns the token id, or ``None`` when unavailable.
+        """
+        try:
+            return getattr(self.tokenizer, attr, None)
+        except (ValueError, TypeError):
+            return None
+
     def _resolve_think_end_token_ids(self) -> list[int] | None:
         """Resolve token ID(s) for the close-think tag.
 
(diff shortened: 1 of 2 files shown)


jundot:
lgtm. covers all three call sites, helper deduplication makes sense, and the test properly simulates the real crash via PropertyMock.

one minor note: in `_detect_needs_think_prefix`, multi-token think models (gemma 4) now fall through to the `convert_tokens_to_ids` fallback instead of returning False immediately. end result is the same since those models don't have a `<think>` vocab entry, so it hits the unk check and returns False anyway.

will add `Fixes #659` to the merge commit — same root cause.

PR #678, opened 2026-04-08
Feature/audio voicedesign params

jnchaba:
Extends PR #676's voice cloning support with two additional Qwen3-TTS capabilities:

  - VoiceDesign: when a model exposes `generate_voice_design()` and instructions are provided, routes through the dedicated VD path instead of `standard generate()`
  - Generation params: `temperature`, `top_k`, `top_p`, `repetition_penalty`, and `max_tokens` are accepted in the API schema and forwarded to whichever generation path is used

Also fixes test fixtures to use plain FakeModel instead of MagicMock,preventing false hasattr() positives for VoiceDesign detection.

You can test this with the following CLI commands:

```
# request VoiceDesign TTS
    -d '{
      "model": "Qwen3-TTS-12Hz-1.7B-VoiceDesign-bf16",
      "input": "Hello, this is a test of voice design.",
      "instructions": "female, calm, slow, clear enunciation",
      "temperature": 0.9
    }' --output voice_design.wav

# play wav file
afplay voice_design.wav
```

And cloning works with the designed voice, thanks to  PR #676 

```
# request cloned voice TTS
    -d '{
      "model": "Qwen3-TTS-12Hz-1.7B-Base-bf16",
      "input": "Now I am saying something completely different in the same voice.",
      "voice": "en",
      "ref_audio": "'$(base64 -i ~/voice_design.wav)'",
      "ref_text": "Hello, this is a test of voice design.",
      "temperature": 0.9
    }' --output cloned.wav

# play wav file
afplay cloned.wav
```

### DIFF
--- a/omlx/api/audio_models.py
+++ b/omlx/api/audio_models.py
@@ -38,6 +38,11 @@ class AudioSpeechRequest(BaseModel):
     response_format: Optional[str] = "wav"
     ref_audio: Optional[str] = None
     ref_text: Optional[str] = None
+    temperature: Optional[float] = None
+    top_k: Optional[int] = None
+    top_p: Optional[float] = None
+    repetition_penalty: Optional[float] = None
+    max_tokens: Optional[int] = None
 
 
 class AudioProcessRequest(BaseModel):
--- a/omlx/api/audio_routes.py
+++ b/omlx/api/audio_routes.py
@@ -239,6 +239,11 @@ async def create_speech(request: AudioSpeechRequest):
             instructions=request.instructions,
             ref_audio=ref_audio_path,
             ref_text=request.ref_text,
+            temperature=request.temperature,
+            top_k=request.top_k,
+            top_p=request.top_p,
+            repetition_penalty=request.repetition_penalty,
+            max_tokens=request.max_tokens,
         )
     except HTTPException:
         raise
--- a/omlx/engine/tts.py
+++ b/omlx/engine/tts.py
@@ -120,6 +120,11 @@ class TTSEngine(BaseNonStreamingEngine):
         instructions: Optional[str] = None,
         ref_audio: Optional[str] = None,
         ref_text: Optional[str] = None,
+        temperature: Optional[float] = None,
+        top_k: Optional[int] = None,
+        top_p: Optional[float] = None,
+        repetition_penalty: Optional[float] = None,
+        max_tokens: Optional[int] = None,
         **kwargs,
     ) -> bytes:
         """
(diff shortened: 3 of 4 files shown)


ethannortharc:
Thanks for building on the voice clone work — the generation params, `FakeModel` test fix, and `model.sample_rate`
usage are all solid improvements.

One concern with the VoiceDesign routing:

The `hasattr(model, "generate_voice_design")` check at the omlx engine level introduces a second routing layer that
duplicates what mlx-audio already does internally. When `instruct` is passed to `model.generate()`, the model's own
`generate()` method already routes to `generate_voice_design()` based on `config.tts_model_type == "voice_design"` — this is the authoritative check.

The problem with `hasattr` detection: **CustomVoice models also have `generate_voice_design()` on them** (inherited from the same base class), but they should route to `generate_custom_voice()` instead. If a user sends
`instructions` to a CustomVoice model, the current PR would call `generate_voice_design()` directly, bypassing the model's own type-based routing. The model-level method does have a guard (`if self.config.tts_model_type !=. "voice_design": raise ValueError`), so it won't silently produce wrong output — but the user gets a confusing error
instead of the correct CustomVoice behavior.

Suggestion: remove the `hasattr`-based VoiceDesign branch and keep passing `instruct=instructions` through
`model.generate()` — the model already knows how to route it correctly based on its own `tts_model_type`. The generation params, FakeModel fix, and sample_rate changes can stand on their own.

jundot:
Thanks for the PR. The generation params forwarding, FakeModel test fixture, and model.sample_rate usage are all good improvements that i want to keep.

@ethannortharc's feedback on the VoiceDesign routing is correct. The existing tts.py already uses `inspect.signature(model.generate).parameters` to route kwargs through `model.generate()`, which lets mlx-audio handle the internal dispatch based on `config.tts_model_type`. Adding a `hasattr(model, "generate_voice_design")` branch at the omlx layer creates a second routing layer that conflicts with this.

The specific problem: CustomVoice models inherit `generate_voice_design()` from the same base class. If a user sends `instructions` to a CustomVoice model, this PR would call `generate_voice_design()` directly, bypassing the model's own type check. The model's guard would raise a ValueError instead of routing to the correct CustomVoice path.

Suggestion: drop the `hasattr`-based VoiceDesign branch and keep passing `instruct=instructions` through `model.generate()` as the existing code does. The generation params, FakeModel fixture, and sample_rate changes can stand on their own without it.

Also, now that #676 is merged, this branch will need a rebase since both PRs add `ref_audio`/`ref_text` to `AudioSpeechRequest`.

jundot:
Merging this as-is since a lot of people are waiting on the generation params support. I'll follow up with a commit that removes the hasattr-based VoiceDesign routing per my earlier comment.

jnchaba:
I totally missed the discussion here, sorry. Thanks for the feedback, and thank you for accepting my PR. I agree with the overall analysis, and would have fixed this -- had it not got lost in my inbox. I see that the follow up commit got posted. Great job!

PR #679, opened 2026-04-09
feat(admin): add hotswappable engine package management

0xClandestine:
## Summary

Add hotswappable engine package management to the admin panel, allowing users to install and update MLX engine packages (mlx-lm, mlx-vlm, mlx-embeddings, mlx-audio) directly from the UI.

### Features

- **Engine Versions Card**: Now clickable to open package manager modal
- **Package Manager Modal**: Shows available versions from PyPI with real-time installation progress
- **Backend API**:
  - `GET /api/engine-packages/available` - Fetch available versions from PyPI
  - `POST /api/engine-packages/install` - Install/update a package
  - `GET /api/engine-packages/tasks` - List install tasks with real-time status
  - `DELETE /api/engine-packages/task/{task_id}` - Delete completed task

### Bug Fixes

- Fixed version sorting crash (`'<' not supported between instances of 'str' and 'int'`) when sorting versions with mixed formats
- Fixed stale commit SHAs being displayed after installing new versions via PyPI
  - Removed incorrect fallback to `pyproject.toml` commits for PyPI-installed packages
  - Fixed `direct_url.json` parsing to only trust `vcs_info` for actual VCS installs (`git+`, `hg+`, etc.)
- Added delay + double `loadStats()` after install to ensure metadata is properly refreshed

### Technical Details

- Installation runs in background using `asyncio.to_thread`
- Task polling with 1-second interval for real-time status updates
- Uses `packaging.version.Version` for proper semantic version sorting
- Renamed `_load_fallback_commits` to `_load_engine_commits_json` to clarify scope (app bundle only)

### Files Changed

- `omlx/admin/routes.py`: API endpoints + engine info fixes
- `omlx/admin/static/js/dashboard.js`: Frontend UI + install logic
- `omlx/admin/templates/dashboard.html`: Package manager modal
- `omlx/admin/templates/dashboard/_status.html`: Clickable engine versions card


0xClandestine:
image
image

jundot:
Thanks for putting this together. Hotswappable engines is a direction I've been thinking about too, and the upside would be significant if it lands cleanly.

The blocker on my side is the current structure. oMLX carries a lot of monkey patches that are pinned to specific commits of mlx-lm / mlx-vlm / mlx-embeddings / mlx-audio. Even setting the patches aside, the cost of letting users move the pin around outweighs what we gain from a simple version bump right now. Small upstream changes in batch generator internals, cache layout, processor wiring, or sampler behavior have broken oMLX in subtle ways more than once, and I don't have a good way to gate that per-engine yet.

So I'd love to ship this, but I can't promise swappable behavior across all engines in the current shape of the codebase. I'll keep this PR in mind as a reference for when the patch surface shrinks enough to make it safe. Appreciate the work.

PR #688, opened 2026-04-09
fix(vlm): use mlx-lm decode model for batch=1, 2x VLM generation speed

### LINKED ISSUE #687: VLM decode path bypasses mlx-lm decode model for batch=1, 2x slower than necessary
## Summary

`VLMModelAdapter.__call__()` in `omlx/models/vlm.py` gates the fast mlx-lm decode model behind `input_ids.shape[0] > 1`, meaning single-request VLM decode always takes the slow path through `_IntOffsetCacheProxy`. Changing `> 1` to `>= 1` doubles VLM generation speed.

## Measurements (Mac Studio M3 Ultra 96GB, oMLX v0.3.4)

| Output Tokens | Before (tok/s) | After fix (tok/s) | Improvement |
|---------------|---------------|-------------------|-------------|
| 50 | 43.7 | 87.6 | +100% |
| 100 | 45.2 | 94.4 | +109% |
| 200 | 45.5 | 98.4 | +116% |
| 400 | 45.9 | 99.5 | +117% |

Model: Qwen3-VL-30B-A3B-Instruct-4bit (MoE, 3B active). Quality verified: 5/5 valid JSON with production extraction prompt, no regression.

## Root Cause

`_wrap_caches()` (line 85) creates 48 `_IntOffsetCacheProxy` objects **per generation step** to convert `mx.array` offsets to Python `int`. Each proxy's `.offset` property triggers `int(raw.max().item())` — an implicit `mx.eval` GPU→CPU sync. Cost: 48 layers × 0.185ms = 8.88ms per token.

The `_decode_model` (mlx-lm model, already loaded with zero-copy weight sharing) uses `BatchKVCache` with native `mx.array` offsets — no proxy needed. But it's only used when `input_ids.shape[0] > 1` (multi-request batch), not for the common single-request case.

## Three-Layer Decomposition

| Path | tok/s | Overhead |
|------|-------|----------|
| Raw mlx-lm (text model, no server) | 100.5 | baseline |
| Raw mlx-vlm (VL model, text-only) | 77.5 | 23% (library + vision tower) |
| oMLX text path (BatchedEngine) | 89.7 | 11% server overhead |
| **oMLX VLM path (VLMBatchedEngine)** | **45.9** | **41% proxy overhead** |
| **oMLX VLM path (patched)** | **99.5** | **~1% — matches ceiling** |

## Fix

```diff
-            if self._decode_model is not None and input_ids.shape[0] > 1:
+            if self._decode_model is not None and input_ids.shape[0] >= 1:
```

`omlx/models/vlm.py`, line 261

The decode model is already loaded and weight-sharing (zero-copy). Prefill still uses the VLM path for vision embedding injection. Only decode changes.

## Environment

- oMLX v0.3.4 (stable)
- macOS 26.3.1, Apple M3 Ultra, 96GB
- mlx-lm 0.31.2, mlx-vlm 0.4.4, mlx 0.31.1

mdevk:
## Summary

Use the mlx-lm decode model for single-request VLM decode (`batch=1`), avoiding the `_IntOffsetCacheProxy` overhead.

**One-line change:** `input_ids.shape[0] > 1` → `input_ids.shape[0] >= 1` in `VLMModelAdapter.__call__()`.

## Problem

The mlx-lm decode model (already loaded, zero-copy weight sharing) was only used for `batch > 1`. For the common `batch=1` case, every generation step:
1. Creates 48 `_IntOffsetCacheProxy` objects (`_wrap_caches()`)
2. Each proxy converts `mx.array` offset to Python `int` via `raw.max().item()`
3. This triggers 48 implicit `mx.eval` GPU→CPU synchronizations per token

**Cost:** 8.88ms per token (48 layers × 0.185ms).

## Results (Mac Studio M3 Ultra 96GB, Qwen3-VL-30B-A3B-Instruct-4bit)

| Output Tokens | Before | After | Improvement |
|---|---|---|---|
| 50 | 43.7 tok/s | 87.6 tok/s | **+100%** |
| 100 | 45.2 | 94.4 | **+109%** |
| 200 | 45.5 | 98.4 | **+116%** |
| 400 | 45.9 | 99.5 | **+117%** |

Quality verified: 20/20 valid JSON on production extraction prompt, no regression.

Concurrency also improves: `conc=4` with patch achieves 0.99 photos/s at 100% JSON validity (vs 0.86 pre-patch).

## Three-Layer Decomposition

| Path | tok/s |
|---|---|
| Raw mlx-lm (text model) | 100.5 |
| Raw mlx-vlm (VL model, text) | 77.5 |
| oMLX VLM path (before) | 45.9 |
| **oMLX VLM path (after)** | **99.5** |

The fix eliminates the entire oMLX VLM serving overhead, making VLM decode match the raw mlx-lm ceiling.

Fixes #687

### DIFF
--- a/omlx/models/vlm.py
+++ b/omlx/models/vlm.py
@@ -256,9 +256,9 @@ class VLMModelAdapter(nn.Module):
             result = self._forward_with_embeddings(input_ids, wrapped_cache, **kwargs)
         else:
             # Standard decode/prefill path: token IDs only.
-            # Use mlx-lm decode model for batched decode (batch > 1)
-            # since mlx-vlm language models may not handle batching.
-            if self._decode_model is not None and input_ids.shape[0] > 1:
+            # Use mlx-lm decode model for ALL decode when available.
+            # Avoids _IntOffsetCacheProxy overhead (48 layers x 0.185ms/token). See #687.
+            if self._decode_model is not None and input_ids.shape[0] >= 1:
                 result = self._decode_model(input_ids, cache=cache, **kwargs)
             else:
                 if hasattr(self._vlm_model, "_set_position_state"):
--- a/tests/test_vlm_model_adapter.py
+++ b/tests/test_vlm_model_adapter.py
@@ -162,6 +162,50 @@ class TestVLMModelAdapter:
         assert call_args[0][0] is input_ids
         assert len(call_args[1]["cache"]) == 1
 
+    def test_forward_uses_decode_model_at_batch_1(self):
+        """Test that decode_model is used for batch=1 decode (no proxy overhead)."""
+        from omlx.models.vlm import VLMModelAdapter
+
+        vlm = self._make_mock_vlm_model()
+        decode_model = MagicMock()
+        decode_logits = MockMXArray(shape=(1, 10, 32000))
+        decode_model.return_value = decode_logits
+        adapter = VLMModelAdapter(vlm, decode_model=decode_model)
+
+        input_ids = MockMXArray(shape=(1, 10))
+        cache = [MagicMock()]
+
+        result = adapter(input_ids, cache=cache)
+
+        # decode_model should be called (fast path, no proxy wrapping)
+        decode_model.assert_called_once()
+        call_args = decode_model.call_args
+        assert call_args[0][0] is input_ids
+        # cache should be passed directly (no _IntOffsetCacheProxy wrapping)
+        assert call_args[1]["cache"] is cache
+        # language_model should NOT be called
+        vlm.language_model.assert_not_called()
+
+    def test_forward_without_decode_model_falls_back_to_language_model(self):
+        """Test that without decode_model, language_model is used with wrapped cache."""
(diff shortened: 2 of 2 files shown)


mdevk:
@jundot wdyt?

jundot:
Thanks for the detailed investigation and benchmarks. I verified the analysis against the codebase and the change is sound:

- Vision embedding paths (paths 1 and 2 in `__call__`) are unaffected since they trigger before the decode branch
- `_set_position_state` is only relevant for the vlm language_model, not the mlx-lm decode_model which handles its own RoPE
- The fallback when `decode_model is None` is preserved
- This effectively restores the original e301d66 design where decode_model handled all batch sizes, before 47df15a conservatively narrowed it to batch > 1

One thing: `input_ids.shape[0] >= 1` is always true for any valid tensor, so the condition is effectively just `self._decode_model is not None`. I'll simplify that and make `_wrap_caches` lazy in a follow-up commit after merging.

PR #695, opened 2026-04-10
fix: preserve harmony analysis channel in non-streaming output

### LINKED ISSUE #569: reasoning_content is null in non-streaming chat completions for Harmony models (gpt-oss)
## Description

When using gpt-oss-20b with the non-streaming `/v1/chat/completions` endpoint, `reasoning_content` is always `null` in the response. The same request in streaming mode correctly returns `reasoning_content` chunks.

## Steps to reproduce

**Non-streaming (broken):**
```bash
curl http://localhost:19200/v1/chat/completions \
  -d '{"model":"gpt-oss-20b-MLX-4bit-gs32","messages":[{"role":"user","content":"What is 2+2?"}]}'
```
Response: `"reasoning_content": null`

**Streaming (works):**
```bash
curl http://localhost:19200/v1/chat/completions \
  -d '{"model":"gpt-oss-20b-MLX-4bit-gs32","messages":[{"role":"user","content":"What is 2+2?"}],"stream":true}'
```
Chunks correctly contain `"delta":{"reasoning_content":"We have a question..."}`.

## Evidence the thinking is parsed

The server log shows the Harmony adapter correctly parsing the analysis channel:

```
omlx.adapter.harmony - Parsed 2 messages
Message 0: channel=analysis, recipient=<|start|>assistant, content_count=1
Message 1: channel=final, recipient=None, content_count=1
```

The thinking content exists internally but isn't mapped to `reasoning_content` in the non-streaming aggregated response.

## Expected behavior

The non-streaming response should include the analysis channel content in `reasoning_content`, matching the streaming behavior. Qwen3.5 models correctly return `reasoning_content` in non-streaming mode — this is specific to Harmony-format models.

## Environment

- oMLX v0.3.2
- Model: gpt-oss-20b-MLX-4bit-gs32 (halley-ai/gpt-oss-20b-MLX-4bit-gs32)
- macOS, M1 Max

jaredlockhart:
## Summary

Non-streaming `/v1/chat/completions` for gpt-oss (Harmony) models returned `reasoning_content: null` because the scheduler only persisted the `final` channel as `visible_text`. Streaming worked because the analysis channel was emitted as `<think>…</think>` in `stream_text`.

## Fix

- `parse_tool_calls_from_tokens` now also returns `analysis_text` from the `analysis` channel.
- `HarmonyOutputParserSession.finalize()` wraps analysis text as `<think>…</think>` and returns it via a new `output_text_prefix` field on `OutputParserFinalizeResult`.
- Scheduler prepends `output_text_prefix` to `request.output_text` on finalize, so the existing `extract_thinking(output.text)` in the non-streaming server paths recovers reasoning with no server-side changes.

## Test

- New regression test `test_harmony_non_streaming_preserves_reasoning` in `tests/test_output_parser.py` drives analysis+final tokens through the session and asserts `extract_thinking` recovers both channels.
- Fails on main, passes after the fix.

Fixes #569

### DIFF
--- a/omlx/adapter/harmony.py
+++ b/omlx/adapter/harmony.py
@@ -351,9 +351,9 @@ class HarmonyStreamingParser:
 def parse_tool_calls_from_tokens(
     token_ids: list[int],
     prepend_start: bool = True,
-) -> tuple[str, list[dict[str, str]]]:
+) -> tuple[str, str, list[dict[str, str]]]:
     """
-    Parse tool calls from complete token sequence (non-streaming).
+    Parse a complete Harmony token sequence (non-streaming).
 
     Args:
         token_ids: Model output token ID list
@@ -361,12 +361,13 @@ def parse_tool_calls_from_tokens(
             Set to False if token_ids already includes start tokens.
 
     Returns:
-        (output_text, tool_calls)
-        - output_text: Text from final channel
+        (output_text, analysis_text, tool_calls)
+        - output_text: Text from the final channel
+        - analysis_text: Chain-of-thought text from the analysis channel
         - tool_calls: [{"name": "...", "arguments": "..."}]
     """
     if not token_ids:
-        return "", []
+        return "", "", []
 
     try:
         encoding = load_harmony_encoding("HarmonyGptOss")
@@ -399,6 +400,7 @@ def parse_tool_calls_from_tokens(
             )
 
         output_text = ""
+        analysis_text = ""
         tool_calls = []
 
         for msg in messages:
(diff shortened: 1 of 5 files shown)


jundot (approved):
Reviewed the full data flow from `parse_tool_calls_from_tokens` through scheduler finalize to `extract_thinking` in the server layer. Everything checks out.

**What i like about this approach:**
- Reuses the existing `extract_thinking()` pipeline instead of adding a parallel reasoning path. No server-side changes needed.
- The `output_text_prefix` field is additive with a default value, so Gemma4 and other parsers are completely unaffected.
- Streaming and non-streaming paths stay cleanly separated. No risk of double-wrapping `<think>` tags.

**Minor note:**
- The existing tests (`test_extracts_tool_call`, `test_extracts_final_text`) unpack `analysis_text` but don't assert its value. I'll add `assert analysis_text == ""` in a follow-up commit.

Clean fix, good test coverage. Merging.

PR #703, opened 2026-04-10
Expose server proxy and TLS configuration, and fix regressions uncovered by not-slow tests

applesauce49:
## Summary
This PR delivers two things:

1. Adds the server networking configuration feature so operators can run oMLX behind enterprise proxy and TLS-intercepted environments more reliably.
2. Fixes runtime and test regressions found in the not-slow suite, including a real thinking-budget state bug and a boundary snapshot cleanup race.

## Feature Added
- Exposed server network controls through the serve configuration flow:
  - HTTP proxy
  - HTTPS proxy
  - No-proxy bypass list
  - Custom CA bundle
- Ensured these settings are applied consistently at startup so downstream requests inherit the expected network/TLS behavior.

## Bug Fixes
- Fixed thinking budget token progression logic so forced close-think and natural end detection advance correctly across decode steps.
- Fixed a boundary snapshot async-writer cleanup race where canceled requests could still leave behind request directories or late-written files.
- Hardened cancellation handling in the snapshot writer by re-checking cancellation around file materialization and rename paths.

## Test/Compatibility Fixes
To align tests with current intended behavior and interfaces:

- Updated VLM discovery regression fixture to include vision metadata where required.
- Updated CLI scheduler option expectations to current serve flags.
- Updated server health-check tests to patch session-based requests flow.
- Updated VLM tool-parser availability test to reflect the current dual-backend parser fallback path.

## Validation
- Re-ran the previously failing set from the not-slow run.
- Result: all 10 previously failing tests now pass.

## Impact
- Improves deployability in proxied/TLS-managed environments.
- Resolves runtime correctness issues in thinking-budget control and boundary snapshot cleanup.
- Restores test reliability by matching tests to current production behavior.


jundot (commented):
Thanks for the contribution. Merging this, but will revert the `thinking.py` change in a follow-up. The `n + 1` was intentional based on `GenerationBatch._step()` pipelining timing where the first appended token is a prompt token, not a generated one. With `n`, that prompt token gets fed into `_update_state` which is technically incorrect.

Will also clean up the defensive hydration duplication in `admin/routes.py` separately.

PR #707, opened 2026-04-10
fix: chat overscroll on main and markdown table rendering

### LINKED ISSUE #698: Chat window overscroll and auto-scroll UX/UI issue
**Describe the bug**
For any longer chat conversation, if you scroll not inside of the chat window, but the outer elements, the <body> will scroll (overscroll). It needs a overflow:hidden style.

**To Reproduce**
1. Go to 'Chat'
2. Chat for 5-6 turns
3. Scroll on outer elements
4. See the chat window scroll away (body scroll)

**Expected behavior**
Only chat conversation scrolls.

**Screenshots**

Image

 - OS: macOS
 - Browser: Google Chrome, Brave, ...

**Additional context**
Looks like this only happens with Markdown / Table rendering involved and longer conversations.

kyr0:
Fixes https://github.com/jundot/omlx/issues/698 by limiting overscroll behaviour and markdown table rendering (it only happens in markdown table rendering cases).

### DIFF
--- a/omlx/admin/templates/chat.html
+++ b/omlx/admin/templates/chat.html
@@ -103,6 +103,8 @@
         font-size: 14px;
         line-height: 1.6;
         color: var(--text-primary);
+        overflow-x: auto;
+        min-width: 0;
     }
 
     .user-message {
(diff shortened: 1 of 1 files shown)


jundot (approved):
Looks good. Clean CSS fix that addresses the overscroll issue correctly.

The `min-width: 0` pattern on the flex children is the right approach here, and `overflow-x: auto` on the message body ensures wide tables get a scrollbar instead of pushing the layout.

Table styling is a nice addition since there was nothing before. One minor note: `white-space: nowrap` on all td/th cells means text-heavy tables will always scroll horizontally instead of wrapping. Not a blocker, just something to be aware of. Could revisit later if needed.

Merging.

PR #717, opened 2026-04-11
fix: wire up TurboQuant KV cache conversion + skip sensitive last layer (#661)

### LINKED ISSUE #661: Did the performance test not include TurboQuant?

Landon-Molt:
### Fix 1: `_apply_turboquant_kv()` is dead code (#661)

The function was defined but **never called**. Users enable TQ in settings, the log says "TurboQuant KV cache enabled", but memory usage is identical to TQ OFF. The setting has zero effect.

**Changes:**
- Monkey-patch `TurboQuantKVCache.merge = BatchTurboQuantKVCache.merge` so `_merge_caches()` works with TQ caches
- Fix `_apply_turboquant_kv()` to use `TurboQuantKVCache.from_cache()` (preserves prefilled KV data — old code created empty caches)
- Call `_apply_turboquant_kv()` after external prefill, before `batch_generator.insert()`
- Fix `BatchTurboQuantKVCache.make_mask()` — was passing `return_array` to `create_causal_mask()` and `mx.array` offsets to `mx.arange()`

### Fix 2: Last-layer sensitivity causes garbage on gemma-4-31B

Once Fix 1 wires up the conversion, gemma-4-31B produces garbage output. Root cause: the **last full-attention layer** (layer 59/60, closest to the output head) is too sensitive to KV quantization — even 8-bit TQ on that single layer alone produces garbage.

Confirmed through systematic per-layer testing:

| Layers converted | Output |
|---|---|
| Layers 5-53 (9/10) | ✅ "The capital of France is Paris." |
| Layer 59 alone | ❌ garbage |
| All 10 layers | ❌ garbage |
| All except layer 59 | ✅ correct |

**Fix:** Add `turboquant_skip_last` model setting (default: `true`) that skips the last KVCache layer. Only applied when the model has mixed cache types (gemma-4: RotatingKVCache + KVCache) and more than one KVCache layer. Models with only KVCache (Qwen3.5) convert all layers.

This aligns with independent findings:
- **llama.cpp [#20977](https://github.com/ggml-org/llama.cpp/issues/20977)**: gemma-4 TurboQuant on all KV layers = catastrophic (PPL >100k)
- **llama.cpp [#21591](https://github.com/ggml-org/llama.cpp/issues/21591)** (NexusQuant): "Protecting first/last 2 layers at FP16 recovers" Qwen2.5-7B
- **vllm [#28538](https://github.com/vllm-project/vllm/issues/28538)**: KV quantization RFC — "preserving numerically sensitive layers in higher precision"

We also verified NexusQuant's Qwen2.5-7B finding: this model breaks completely with TQ at any bit level (4/6/8-bit), even with boundary layer protection. Some architectures are fundamentally incompatible with TQ's quantization approach.

### All gemma-4-31B variants ✅

| Variant | Weight quant | TQ 4-bit (skip_last) |
|---------|-------------|:---:|
| gemma-4-31B-it | bf16 | ✅ |
| gemma-4-31B-it-oQ4 | oQ4 | ✅ |
| gemma-4-31B-it-oQ8 | oQ8 | ✅ |
| gemma-4-31b-it-8bit | 8-bit | ✅ |

### Other models ✅

| Model | TQ OFF | TQ ON |
|-------|:---:|:---:|
| gemma-4-26B-A4B-it-oQ8 | ✅ | ✅ |
| Qwen3.5-9B-8bit | ✅ | ✅ |
| Qwen3.5-27B-8bit | ✅ | ✅ |
| Qwen3.5-35B-A3B-8bit | ✅ | ✅ |

## Peak memory limitation

The current approach converts KV caches **after prefill** via `from_cache()`. Peak memory is set during prefill when full-precision KV exists, so savings are minimal on large models.

**The proper approach** (used by llama.cpp `--cache-type-k turbo3`): create TQ caches **before prefill** so tokens are quantized on the fly via `update_and_fetch`. This avoids ever storing full-precision KV. We confirmed this works in standalone testing — empty TQ caches before prefill produce correct output. The scheduler integration needs further work to find the correct insertion point.

| Model | pp16K | pp32K | pp65K |
|---|---|---|---|
| gemma-4-31B-oQ8 (9/60 layers) | 0 GB | 0 GB | 0.48 GB |
| gemma-4-26B-A4B (4/30 layers) | 0.14 GB | 0 GB | 0 GB |
| Qwen3.5-27B (15/64 layers) | 0 GB | 0 GB | 0 GB |
| Qwen3.5-35B (7/64 layers) | 0 GB | 0 GB | 0 GB |

## Known issues

1. **Concurrency > 1**: `BatchTurboQuantKVCache` batched decode produces corrupt output. Pre-existing issue — the inherited `decode_attention` doesn't handle B>1. Does not affect single-request usage.

2. **Peak memory**: Post-prefill conversion doesn't reduce peak memory on large models. On-the-fly quantization during prefill is needed (WIP — works in standalone, needs scheduler integration).

3. **Qwen2.5-7B**: Incompatible with TQ at any bit level. Architecture-specific limitation, not a bug in this PR.

Closes #661

### DIFF
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -254,6 +254,9 @@ class BatchedEngine(BaseEngine):
             if tq_enabled:
                 tq_bits = float(getattr(self._model_settings, "turboquant_kv_bits", 4))
                 self._engine.engine.scheduler._turboquant_kv_bits = tq_bits
+                self._engine.engine.scheduler._turboquant_skip_last = getattr(
+                    self._model_settings, "turboquant_skip_last", True
+                )
 
         # SpecPrefill: load draft model and pass to scheduler
         if self._model_settings is not None:
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -655,6 +655,9 @@ class VLMBatchedEngine(BaseEngine):
                 apply_turboquant_attention_patch()
                 tq_bits = float(getattr(self._model_settings, "turboquant_kv_bits", 4))
                 self._engine.engine.scheduler._turboquant_kv_bits = tq_bits
+                self._engine.engine.scheduler._turboquant_skip_last = getattr(
+                    self._model_settings, "turboquant_skip_last", True
+                )
                 logger.info(f"TurboQuant KV cache enabled for VLM: {tq_bits} bits")
 
         # SpecPrefill: load draft model and pass to scheduler
--- a/omlx/model_settings.py
+++ b/omlx/model_settings.py
@@ -60,6 +60,7 @@ class ModelSettings:
     # TurboQuant KV cache (mlx-vlm backend)
     turboquant_kv_enabled: bool = False
     turboquant_kv_bits: float = 4  # 2, 2.5, 3, 3.5, 4, 6, 8
+    turboquant_skip_last: bool = True  # Skip last KVCache layer (prevents corruption on sensitive models)
 
     # SpecPrefill (experimental: attention-based sparse prefill for MoE models)
     specprefill_enabled: bool = False
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -153,6 +153,16 @@ def _patched_generation_batch_step(self):
 GenerationBatch._step = _patched_generation_batch_step
 
 
+# Monkey-patch TurboQuantKVCache.merge so _merge_caches() works
+try:
+    from mlx_vlm.turboquant import TurboQuantKVCache as _TQCache
+    from .turboquant_kv import BatchTurboQuantKVCache as _BTQCache
+    if not hasattr(_TQCache, "merge"):
+        _TQCache.merge = _BTQCache.merge
+except ImportError:
+    pass
+
+
 # Cache class names known to be sliceable (no boundary snapshots needed).
 _KNOWN_SLICEABLE_CACHE_TYPES = frozenset({
     "KVCache", "BatchKVCache", "QuantizedKVCache",
@@ -414,6 +424,7 @@ class Scheduler:
 
         # TurboQuant KV cache (set by engine if model_settings has it enabled)
         self._turboquant_kv_bits: Optional[float] = None
+        self._turboquant_skip_last: bool = True
 
         # Request management - following vLLM's design
         self.waiting: deque[Request] = deque()  # Waiting queue (FCFS)
@@ -973,17 +984,26 @@ class Scheduler:
     # External prefill (composition pattern — replaces _process_prompts)
     # ------------------------------------------------------------------
 
-    def _apply_turboquant_kv(self, prompt_cache: List[Any]) -> None:
-        """Convert individual KVCache layers to TurboQuantKVCache."""
-        from .turboquant_kv import BatchTurboQuantKVCache
+    def _apply_turboquant_kv_empty(self, prompt_cache: List[Any]) -> None:
+        """Replace KVCache with empty TurboQuantKVCache before prefill.
+
+        Tokens are quantized on the fly during update_and_fetch, avoiding
+        the peak memory spike from storing full-precision KV then converting.
+        Skips the last KVCache layer if turboquant_skip_last is set.
+        """
         from mlx_vlm.turboquant import TurboQuantKVCache
         from mlx_lm.models.cache import KVCache, CacheList
+
+        kv_indices = [i for i, c in enumerate(prompt_cache) if isinstance(c, KVCache)]
+        skip_last = self._turboquant_skip_last and len(kv_indices) > 1
+        last_kv_idx = kv_indices[-1] if skip_last else -1
 
         converted = 0
         bits = float(self._turboquant_kv_bits)
         for i, cache_obj in enumerate(prompt_cache):
-            cls_name = type(cache_obj).__name__
             if isinstance(cache_obj, KVCache):
+                if i == last_kv_idx:
+                    continue
                 prompt_cache[i] = TurboQuantKVCache(bits=bits)
                 converted += 1
             elif isinstance(cache_obj, CacheList):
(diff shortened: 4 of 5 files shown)


Landon-Molt:
~~Closing this PR — the conversion still produces corrupt output on gemma-4-31B.~~

**Update:** PR reopened. The gemma-4-31B corruption was caused by the **last full-attention layer** (layer 59/60) being too sensitive to KV quantization. Added `turboquant_skip_last` setting to skip it. All models now produce correct output at concurrency 1.

However, **peak memory savings are limited** with the post-prefill `from_cache()` approach because both old and new KV coexist at peak. The proper fix is on-the-fly quantization during prefill (creating empty TQ caches before prefill) — this works in standalone testing but needs more work to integrate into the scheduler's runtime path. See PR description for full details.

Landon-Molt:
## Investigation: TurboQuant Peak Memory Root Cause

We traced the full prefill call chain and cross-referenced llama.cpp, vLLM, and upstream mlx-vlm to understand why peak memory savings are near zero even with the on-the-fly approach.

### Root Cause

The attention patch in `patches/turboquant_attention.py` tries `cache.prefill_attention()` during prefill (L>1), but this **always returns `None`** because it requires `_TurboQuantProdCodec` for keys — and ProdCodec was intentionally dropped in mlx-vlm [PR #909](https://github.com/Blaizzy/mlx-vlm/pull/909) in favor of MSE-only.

The fallback calls `cache.dequantize()` which decompresses the **entire cache** to fp16 on every prefill chunk. At 128K context on gemma-4-31B, this is ~31 GB of activation memory per chunk.

This is the same fallback pattern as upstream `mlx_vlm/models/base.py:212-221`. It doesn't matter for mlx-vlm because they use post-conversion (prefill with fp16 KVCache, convert to TQ after). But omlx's on-the-fly approach (TQ caches active during prefill) triggers it on every chunk, making on-the-fly **worse** than post-conversion for peak memory.

### Upstream Status

- **Blaizzy/mlx-vlm [PR #939](https://github.com/Blaizzy/mlx-vlm/pull/939)** (OPEN): Flash-style tiled TBQ prefill — rewrites `prefill_attention` with online softmax, bounds activation memory at ~512 MB regardless of context length. This is the proper fix. Not yet merged.
- **Blaizzy/mlx-vlm [PR #973](https://github.com/Blaizzy/mlx-vlm/pull/973)** (OPEN): Rotation-basis mismatch fix for power-of-two head dims. Affects prefill→decode quality.
- **Blaizzy/mlx-vlm [PR #967](https://github.com/Blaizzy/mlx-vlm/pull/967)** (MERGED): Race condition fix in fused fast-quantize Metal kernels.

### Stopgap Fix

`TurboQuantKVCache.quantized_attention()` (turboquant.py:4976) already exists and works with MSECodec — it processes in 2048-token chunks via codebook lookups, bounding peak memory to chunk-sized intermediates. It's just never called from the SDPA patch. Replacing the `dequantize()` fallback with `quantized_attention()` is a 3-line change that bridges until upstream PR #939 merges.

### What This PR Already Fixes (Still Essential)

1. ✅ `_apply_turboquant_kv()` was dead code — wired up
2. ✅ `turboquant_skip_last` — prevents garbage on gemma-4-31B last layer
3. ✅ On-the-fly TQ cache setup in `_do_external_prefill`
4. ⬜ Peak memory — needs the fallback fix (stopgap) or upstream PR #939 (proper)

Landon-Molt:
## Progress Update: TurboQuant Performance — Root Cause Resolved Upstream

Following up on the investigation from this PR. We've traced the full chain from omlx → mlx-vlm → MLX core and built a solution.

### What we found
1. **`prefill_attention()` is dead code** after mlx-vlm PR #909 dropped ProdCodec — confirmed, documented at [mlx-vlm#1016](https://github.com/Blaizzy/mlx-vlm/issues/1016)
2. **Google's TurboQuant codebook is 3-5x slower than scalar dequant on GPU** — indexed table lookup defeats SIMD coalescing. The ecosystem (llama.cpp, vLLM, TheTom) all use scalar formats instead.
3. **Apple's MMA (simdgroup_matrix) only accepts fp16** — all quantized paths must dequant to threadgroup memory first. This is confirmed by MLX's own `quantized_nax.h`.

### What we built
Fused scalar QV SDPA kernels for MLX core — K stays fp16, V dequanted inline from `mx.quantize` 4-bit:

- **Decode (L=1)**: 19-30% **faster** than fp16 (less memory bandwidth for 4-bit V)
- **Prefill (L>1)**: At parity with dequant+SDPA at ≤16K context, near-parity with fp16 at D=256
- **Memory**: 1.5x KV savings per layer (V only), 2.7x V savings across 60 gemma-4-31B layers

PR submitted to TheTom's MLX fork: [TheTom/mlx#1](https://github.com/TheTom/mlx/pull/1)

### Real model verification
gemma-4-31B (oQ4/oQ8/8bit) all produce coherent, identical output with V 4-bit quantization. No precision compounding across 5 multi-turn exchanges.

### What this means for omlx
Once the upstream kernel lands in MLX (via TheTom's fork → ml-explore/mlx), omlx can:
1. Store V as `mx.quantize` scalar format (not TQ codebook)
2. Call `mx.fast.scaled_dot_product_attention_qv()` for decode — 19-30% faster
3. Use steel prefill for L>1 — no more dequantize fallback, no memory spike
4. Remove the `patches/turboquant_attention.py` monkey-patch entirely

The correctness fixes in this PR (#717) remain essential regardless — `_apply_turboquant_kv()` wiring and `turboquant_skip_last` are omlx-specific.

### Links
- Root cause analysis: [mlx-vlm#1016](https://github.com/Blaizzy/mlx-vlm/issues/1016)
- MLX feature request: [ml-explore/mlx#3404](https://github.com/ml-explore/mlx/issues/3404)
- Steel prefill PR: [TheTom/mlx#1](https://github.com/TheTom/mlx/pull/1)
- Codebook vs scalar: [mlx-lm#1060 comment](https://github.com/ml-explore/mlx-lm/issues/1060#issuecomment-4233306768)

jundot:
Nice work fixing the dead code and the make_mask bugs. Thorough model testing across gemma-4 and Qwen3.5 variants is appreciated.

PR #719, opened 2026-04-11
fix(packaging): find pyenv Python with working pip for sdist wheels

thornad:
## Summary

- `_find_target_python()` in `packaging/build.py` preferred the venvstacks runtime under `_build/cpython-3.11` when `shutil.which("python3.11")` came back empty. That interpreter has `pip` stripped, so the subsequent `pip wheel` call for sdist-only deps like `webrtcvad` crashes with `No module named pip` and breaks the whole build.
- Adds a pyenv fallback (`~/.pyenv/versions/<minor>.*`) so pyenv-managed installs are considered before the stripped runtime.
- Verifies each candidate actually has a working `pip` before selecting it.

## Reproduction

On a machine where `python3.11` isn't shimmed onto `PATH` (common with pyenv + `pyenv local`), running `python3.11 build.py` from source fails mid-lock:

```
[1/4] Building venvstacks layers...
  sdist-only dependency detected: webrtcvad, building wheel...
  Building wheel for webrtcvad (sdist-only, using /Users/.../packaging/_build/cpython-3.11/bin/python3.11)...
/Users/.../packaging/_build/cpython-3.11/bin/python3.11: No module named pip
  ✗ Failed to build wheel for webrtcvad
```

With the fix, the target Python resolves to the pyenv install that does have pip, and the sdist wheel build succeeds.

## Test plan

- [x] `python3.11 packaging/build.py` succeeds end-to-end on a pyenv-only setup (no python3.11 on PATH).
- [x] On a machine where `python3.11` IS on PATH, behavior is unchanged — the `shutil.which` candidate still wins.

### DIFF
--- a/packaging/build.py
+++ b/packaging/build.py
@@ -246,10 +246,26 @@ def _find_target_python() -> str:
     target_minor = match.group(1)  # e.g. "3.11"
     candidates = [
         shutil.which(f"python{target_minor}"),
+        # pyenv-managed interpreters (not on PATH by default)
+        str(Path.home() / ".pyenv" / "versions" / f"{target_minor}" / "bin" / f"python{target_minor}"),
         str(BUILD_DIR / f"cpython-{target_minor}" / "bin" / f"python{target_minor}"),
     ]
+    # Also check pyenv versions with patch numbers (e.g. 3.11.14)
+    pyenv_versions = Path.home() / ".pyenv" / "versions"
+    if pyenv_versions.is_dir():
+        for entry in sorted(pyenv_versions.iterdir(), reverse=True):
+            if entry.name.startswith(f"{target_minor}."):
+                candidates.insert(1, str(entry / "bin" / f"python{target_minor}"))
+                break
     for path in candidates:
-        if path and Path(path).exists():
+        if not path or not Path(path).exists():
+            continue
+        # Verify the interpreter has pip (venvstacks runtimes strip it)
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for the fix, the pip-stripped runtime issue is a real pain and i appreciate you digging into it.

I left a few notes while reviewing:

The path `~/.pyenv/versions/3.11/bin/python3.11` (without patch version) doesn't seem to exist in standard pyenv setups. Pyenv creates directories like `3.11.9` or `3.11.14`, so this candidate would always be skipped. The patch version scan below already covers this correctly, so this line might not be needed.

Also worth considering: pyenv supports `$PYENV_ROOT` for custom install locations (defaults to `~/.pyenv`). Using `os.environ.get("PYENV_ROOT", ...)` would make this work for those setups too.

One more thing, `_find_target_python()` gets called once per sdist retry (up to 10 times through `_lock_with_sdist_retry`), so the toml parsing + pip subprocess checks add up a bit. Caching the result would be a nice touch since it won't change during a single build.

Thinking about it more, the core bug here is just "don't pick an interpreter without pip". When someone runs `python3.11 build.py`, `sys.executable` is already a working python3.11 with pip, so a pip check in the candidate loop is enough to skip the stripped runtime and let the existing fallback handle it. I'll push a minimal fix for that part separately. Thanks again for reporting this.

PR #720, opened 2026-04-11
feat(admin): cache probe endpoint for prompt prefix lookup

thornad:
## Summary

Adds `POST /admin/api/cache/probe` — an admin-authenticated endpoint that reports how much of a rendered chat prompt is already resident in the loaded model's SSD cache, broken down by tier (hot cache / disk index / cold).

Request body (`CacheProbeRequest`):

```json
{
  "model_id": "GLM-4.7-Flash-PRISM-mlx-8bit",
  "messages": [...],
  "tools": null,
  "chat_template_kwargs": null
}
```

Response:

```json
{
  "model_id": "...",
  "model_loaded": true,
  "total_tokens": 19154,
  "block_size": 256,
  "total_blocks": 75,
  "blocks_ssd_hot": 3,
  "blocks_ssd_disk": 71,
  "blocks_cold": 1,
  "ssd_hit_tokens": 18944,
  "cold_tokens": 210,
  ...
}
```

## How it works

The probe renders the prompt via the engine's own `_apply_chat_template` + tokenizer so the token sequence matches what `generate()` would see, then walks the chain-hashed block sequence using `compute_block_hash` — the same hashing the scheduler uses at prefill. For each block it checks the paged SSD cache manager's `_hot_cache` (RAM copy) and `_index` (on-disk files), stopping at the first block that is retrievable from neither (the cache is a contiguous prefix, so everything after that point is necessarily cold).

Ground truth for "cached" is deliberately **not** `BlockAwarePrefixCache._prefix_index` — that dict survives `clear_ssd_cache()` and would report false positives after a manual wipe. Only hot_cache + disk index are consulted.

Works for both batched and VLM engines (guards `_preprocess_messages` / `_apply_chat_template` since the VLM engine only implements the latter).

## Motivation

I've been building a cache-aware chat client on top of this — **[canopy](https://github.com/thornad/canopy)** — that shows a per-node cache dot on every conversation tree node so you can see which branches / regenerations would hit the cache before committing to them. Branching is nearly free when the shared prefix is cached, and exposing that state visually has been useful for iterating on prompts and system messages.

The probe is independently useful (any tool can ask "will this prompt prefill fast?") but canopy is the reference integration if you want to try it: https://github.com/thornad/canopy/releases/tag/v0.1.0 (DMG in the release).

## Test plan

- [x] Probe on a chat with all blocks on SSD → correct `blocks_ssd_disk` count, `blocks_cold` ≈ 0
- [x] Probe after `POST /admin/api/ssd-cache/clear` → all blocks cold (no false positives from `_prefix_index`)
- [x] Probe with a model that's not loaded → returns `{"model_loaded": false, "reason": ...}` gracefully
- [x] Probe with a VLM model (no `_preprocess_messages`) → falls through to direct `tokenizer.apply_chat_template`
- [x] Probe with no messages → returns zero-block payload cleanly

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -72,6 +72,22 @@ class DeleteSubKeyRequest(BaseModel):
     key: str
 
 
+class CacheProbeRequest(BaseModel):
+    """Request model for probing per-prompt cache state.
+
+    Tokenizes a chat message list with the target model's tokenizer, then
+    classifies each block's location in the cache hierarchy:
+    - Hot SSD (in-RAM copy of SSD cache, ready to mount without disk read)
+    - Disk SSD (persisted only, needs disk read to reuse)
+    - Cold (fully uncached — would require full prefill)
+    """
+
+    model_id: str
+    messages: List[Dict[str, Any]]
+    tools: Optional[List[Dict[str, Any]]] = None
+    chat_template_kwargs: Optional[Dict[str, Any]] = None
+
+
 class ModelSettingsRequest(BaseModel):
     """Request model for updating per-model settings."""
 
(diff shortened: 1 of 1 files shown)


jundot (commented):
Thanks for this, the cache probe idea is solid and i can see it being useful beyond canopy too.

Code looks correct — i verified the attribute chain, hash computation, and SSD index access all match the current internals. Read-only and behind admin auth so no risk to existing behavior.

A few things i'd like cleaned up before merging:

**1. Remove `blocks_ram` / `ram_hit_tokens` fields**

These are hardcoded to 0 with a "backwards-compat" comment, but this is a brand new endpoint. There's nothing to be backwards-compatible with. Starting with dead fields just adds noise to the API surface. If a RAM tier distinction becomes real later, we can add it then.

**2. `prefix_index_hits` — remove or justify**

The whole point of this probe is that `_prefix_index` is NOT ground truth (survives `clear_ssd_cache`), which i agree with. But then exposing `prefix_index_hits` in the response is confusing — consumers will wonder what to do with a number that doesn't match the other counts. I'd drop it from the default response. If you need it for debugging, a query param like `?diagnostics=true` would be cleaner.

**3. Tests — i can add these as a follow-up**

Not blocking merge on this, but i'll add unit tests for the main cases (model not loaded, empty messages, normal probe) after merge.

---

Everything else (inline imports, getattr chains, no response model) follows existing patterns in the admin routes so i'm fine with it as-is.

thornad:
Thanks for the review! Addressed the first two, agreed on the third:

1. **`blocks_ram` / `ram_hit_tokens` removed.** You're right — dead fields are noise, not backwards-compat. Gone.

2. **`prefix_index_hits` removed entirely** (rather than gated behind `?diagnostics=true`). I considered the query-param route but the field wasn't actually being used anywhere and YAGNI applies — easier to add it back later with a concrete use case than to carry a speculative flag. If you want it restored for debugging, happy to add it as a follow-up.

3. **Tests** — sounds good, I'll let you handle those after merge.

Response shape is now:

```json
{
  "model_id": "...",
  "model_loaded": true,
  "total_tokens": 19154,
  "block_size": 256,
  "total_blocks": 75,
  "blocks_ssd_hot": 3,
  "blocks_ssd_disk": 71,
  "blocks_cold": 1,
  "ssd_hit_tokens": 18944,
  "cold_tokens": 210
}
```

Force-pushed to the same branch.

jundot (approved):
Looks good, clean diff. All three points addressed exactly how i'd want. Merging now, i'll add tests in a follow-up.

PR #722, opened 2026-04-11
Allow user to scroll up while chat is streaming

### LINKED ISSUE #721: User can't scroll up while chat is streaming (forces auto-scroll down)
**Describe the bug**
The user can't scroll up while chat is streaming (forces auto-scroll down)

**To Reproduce**
1. Go to 'Chat'
2. Have a longer conversation
3. While it streams, try to scroll up
4. Get annoyed because you cannot read shit ;)

**Expected behavior**
Chat should allow the user to scroll up while streaming is in progress.

kyr0:
This little change fixes: https://github.com/jundot/omlx/issues/721 and allows the user to scroll up by not scrolling down automatically once the scroll position has been changed by the user once.

### DIFF
--- a/omlx/admin/templates/chat.html
+++ b/omlx/admin/templates/chat.html
@@ -1160,6 +1160,7 @@
                 const container = document.getElementById('messagesContainer');
                 if (!container) return;
 
+                let lastScrollTop = container.scrollTop;
                 container.addEventListener('scroll', () => {
                     // Detect if user is manually scrolling
                     clearTimeout(this.scrollTimeout);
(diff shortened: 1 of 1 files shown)


jundot (approved):
Looks good to me, nice fix.

The core issue was a race condition where smooth scroll animation during streaming would fire scroll events mid-animation, and since `isAtBottom` was false while `isStreaming` was true, auto-scroll got permanently disabled. Tracking scroll direction instead of relying on streaming state is the right call here.

Merging this in, thanks for the contribution!

kyr0:
@jundot I also developed a custom quantization method for Gemma 4 family models and had to do some ugly forking myriad because I needed oMLX to depend on other mlx-lm version (local fork as well). Same for mlx-embeddings. But because upstream projects are not always likely to merge model/quant methods without much community support, I was patching the oMLX build pipeline to support building the DMG for macOS 26 and macOS 15 even with local dependencies pointing to local forks. This was quite a stretch - I think you can imagine... and it's messy...; does it make sense to go at length and open a PR to allow for "easily switch between local fork dependency version and upstream version/git tag" while still being able to build the DMG without issues? I currently pin omlx as a git submodule in my repo : https://github.com/kyr0/waldwicht it's beyond absurd, because I have to sync oMLX from now on. Would I switch to a release version, I would loose support for my models (Waldwicht-style quants) because it would then depend on a mlx-lm and mlx-embedding that doesn't support my quant method...

You can find details on my quant method here: https://github.com/kyr0/waldwicht/blob/main/TECHNICAL_REPORT.md 

I basically do dynamic quantization via a "grid search" for what **components** (not layers) _in each Gemma 4 layer_ allows for what level of quantization at what level of loss. I found out that we can quantize Gemma 4 much better this way, and if we remove the video and audio towers, we end up <= 3 GiB VRAM with 50 tok/s on Macbook Air M4 using oMLX. 

My models are here: https://huggingface.co/kyr0

PR #723, opened 2026-04-11
Add Support for Microsoft Harrier OSS including quantization support

kyr0:
This PR adds support for Microsoft Harrier OSS embedding models by generalizing on sentence-transformers style embedding exports -- meaning that his PR unlocks basically any sentence-transformer based embedding model export.

It's blocked by the following PR in 3rd party `mlx-embeddings` project right now though - and needs a version/git ref bump once the feature becomes available there: https://github.com/Blaizzy/mlx-embeddings/pull/58 - specifically for Harrier OSS support.

### DIFF
--- a/omlx/model_discovery.py
+++ b/omlx/model_discovery.py
@@ -312,6 +312,42 @@ def _is_causal_lm_embedding(model_path: Path) -> bool:
     return "embedding" in name_lower or "embed" in name_lower
 
 
+def _has_sentence_transformers_embedding_pipeline(model_path: Path) -> bool:
+    """
+    Detect sentence-transformers style embedding exports via modules.json.
+
+    This allows oMLX to recognize embedding exports whose base transformer
+    architecture is ambiguous (for example gemma3_text) but which include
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for the contribution! Sentence-transformers detection via modules.json is a clean approach and the input remapping logic handles the `inputs` vs `input_ids` difference correctly.

A couple things i noticed while reviewing:

**One test gap i'd like to cover in a follow-up** -- the case where modules.json has ONLY `sentence_transformers.models.Transformer` without pooling/normalize modules. The function should return false there, and it does, but theres no test guarding that boundary.

**Minor perf thing** -- `inspect.signature(self.model.__call__)` runs on every `embed()` call. The models call signature never changes after loading so this can be cached as a bool during load. Not blocking, i'll clean this up in a follow-up commit after merge.

Re: the mlx-embeddings PR #58 dependency for harrier specifically -- the general sentence-transformers detection works independently from that, so this is fine to merge now. Once that upstream PR lands we can bump the version.

Merging this and will add the test + caching fix in a follow-up.

PR #737, opened 2026-04-12
fix(oq): chunked load/quantize and streaming VLM sanitizer for huge MoE models

yohann-bearzi:
PR description

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -704,29 +704,207 @@ class _TrackedTensor:
         self.transform = transform
         self.axis = axis
 
-    # Support `weight + 1.0` patterns (norm adjustments)
+    def _clone(self, shape=None, dtype=None, transform=None):
+        return _TrackedTensor(
+            shape if shape is not None else self.shape,
+            dtype if dtype is not None else self.dtype,
+            list(self.sources),
+            transform if transform is not None else self.transform,
+        )
+
+    # Arithmetic — recipe is "fp8_dequant" for the whole sanitize block if weight came from FP8
     def __add__(self, other):
-        return _TrackedTensor(self.shape, self.dtype, list(self.sources), "add")
+        return self._clone(transform="add")
     def __radd__(self, other):
         return self.__add__(other)
     def __sub__(self, other):
-        return _TrackedTensor(self.shape, self.dtype, list(self.sources), "sub")
+        return self._clone(transform="sub")
+    def __mul__(self, other):
+        return self._clone(transform="mul")
+    def __rmul__(self, other):
+        return self.__mul__(other)
+    def __truediv__(self, other):
+        return self._clone(transform="div")
 
-    # Support indexing like tensor[..., :half] for split patterns
+    # Indexing: handle slice + None (broadcast) + tuple variants
     def __getitem__(self, idx):
-        # Rough shape tracking for common slice patterns
-        return _TrackedTensor(self.shape, self.dtype, list(self.sources), "slice")
+        new_shape = list(self.shape)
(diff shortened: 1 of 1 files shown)


yohann-bearzi:
I quantized Qwen3.5-397B-A17B with oMLX and you can check out what it produces at https://huggingface.co/collections/bearzi/qwen35-397b-a17b-oq

yohann-bearzi:
I'm quite busy so I may be slow to respond. If it's stalling, I am fine with letting someone take over from there. Just leave me as author on the original commit and append your tweaks on follow up commits.

jundot (changes requested):
Nice work on this, the chunked load/quantize approach is exactly what huge MoE models need. I found a few things that should be addressed before merging though.

The biggest issue is that `_StreamingPlan` activates for every `ForConditionalGeneration` architecture, but the transform rules inside are Qwen3.5-specific. So if someone quantizes a Gemma4 or LLaVA model, the rename logic would corrupt weight names silently. For example Gemma4's `model.language_model.model.layers.*` would become `language_model.model.model.layers.*` with a double prefix. This needs to be scoped to Qwen3.5 MoE only and let other VLMs keep using the existing `_build_model_sanitizer` path.

Related to that, `_LazyTensorIndex.pop()` returns a `_LazyTensor` object instead of `mx.array`. This is fine in the main quantize loop where you added the `isinstance` check, but sanitizers from mlx-lm/mlx-vlm don't know about `_LazyTensor`. Something like `mx.stack([weights.pop(k) for ...])` in Qwen3 MoE's sanitize would blow up. I think `pop()` should just materialize like `__getitem__` does.

A smaller one: `__iter__` only yields `_index` keys but `keys()` includes `_overrides` too. If a sanitizer writes back with `weights[k] = v`, those keys become invisible during iteration.

Since this is low-level binary parsing and chunked GPU operations, some basic tests would really help. A roundtrip test for `_LazyTensorIndex`, a comparison test between `_quantize_chunked` and `mx.quantize`, and a key mapping test for `_StreamingPlan` would catch the kind of silent corruption bugs that are hardest to debug later.

The core primitives here are genuinely useful and i'd love to see this merged once these are sorted out. Happy to help with follow-up commits if you need.

yohann-bearzi:
It's failing on MiniMax M2.7, I'm patching it then I'll add a commit for it.

yohann-bearzi:
Ready for new review. I'm going to work on making GLM 5.1 quantizable. Hopefully I'm lucky.

jundot (approved):
The discovery-based approach is a much better solution than the original Qwen3.5-specific `_StreamingPlan`. Running the real `Model.sanitize()` on tracked tensor proxies means this works for any model without hardcoding transform rules per architecture. Nice.

All three issues from my previous review are addressed:

- `_StreamingPlan` no longer activates for non-Qwen VLMs. The generic discovery path handles Gemma 4, Trinity Nano, and Qwen 3.5 correctly.
- `_LazyTensorIndex.pop()` now materializes to `mx.array`.
- `__iter__` includes `_overrides` keys.

A few things i want to clean up in follow-up commits after merge:

1. `_StreamingPlan` class (~170 lines) is still in the file but nothing references it anymore. I'll remove it.
2. `_DiscoveredPlan.pop()` hardcodes some transform parameters (`add` always does `+1.0`, `moveaxis` always does `(2, 1)`). This works for current models since discovery falls back to eager sanitize if it can't replay, but i'll add a comment documenting that assumption.
3. I'll add unit tests for `_LazyTensorIndex` and `_quantize_chunked` in a separate commit.

Merging this now. Thanks for the solid work on the chunked primitives.

PR #742, opened 2026-04-13
fix: enrich Gemma 4 tool parameter descriptions to prevent dropped fields

### LINKED ISSUE #741: Gemma 4 drops required tool call parameters when schema descriptions are missing
## Description

Gemma 4 models (31b-it-8bit, 26b-a4b-it-4bit, etc.) silently drop required tool call parameters when those parameters lack explicit `description` fields in the JSON schema. This is especially severe when the parameter name is `description` — a collision with the schema keyword.

## Root Cause

When `convert_tools_for_template()` passes tool schemas to Gemma 4's chat template, parameters without descriptions get rendered ambiguously. The model then omits them from tool call output, producing malformed calls that downstream agents (like OpenCode) can't parse.

## Steps to Reproduce

```bash
curl -s http://127.0.0.1:10741/v1/chat/completions \
  -H "Authorization: Bearer $KEY" \
  -d '{
    "model":"gemma-4-31b-it-8bit",
    "messages":[{"role":"system","content":"Delegate tasks."},
                {"role":"user","content":"Audit code"}],
    "tools":[{"type":"function","function":{
      "name":"delegate",
      "parameters":{"type":"object",
        "properties":{
          "description":{"type":"string"},
          "prompt":{"type":"string"}
        },
        "required":["description","prompt"]}}}],
    "max_tokens":300}'
```

**Result:** Tool call contains `prompt` but NOT `description`.

Adding `"description": "REQUIRED. Short task summary"` to the parameter schema fixes it.

## Proposed Fix

Add `enrich_tool_params_for_gemma4()` in `omlx/api/tool_calling.py` that fills in missing parameter descriptions for required fields before passing to the chat template. Call it in `server.py` when the model is Gemma 4.

I have a working patch — happy to submit a PR.

## Patch

See: https://github.com/fishloa/agentic/blob/main/patches/omlx-gemma4-tool-params.patch

## Environment

- omlx v1 (homebrew, --with-grammar)
- gemma-4-31b-it-8bit, gemma-4-26b-a4b-it-4bit
- macOS, Apple Silicon M2 Ultra 128GB
- Downstream: OpenCode 1.4.0 (triggers "description undefined" schema errors)

fishloa:
## Summary

- Gemma 4 models silently drop required tool call parameters when those parameters lack explicit `description` fields in the JSON schema
- This is especially severe when the parameter is named `description` — the name collides with the schema keyword, confusing the model
- Adds `enrich_tool_params_for_gemma4()` which fills in missing parameter descriptions for required fields before passing tools to the chat template
- Only applied for Gemma models; no-op for parameters that already have descriptions

## Problem

When agents like OpenCode send tool schemas with minimal parameter definitions:

```json
{"name": "delegate", "parameters": {"properties": {
  "description": {"type": "string"},
  "prompt": {"type": "string"}
}, "required": ["description", "prompt"]}}
```

Gemma 4 generates tool calls with `prompt` but omits `description`, causing downstream schema validation failures.

## Fix

`enrich_tool_params_for_gemma4()` in `omlx/api/tool_calling.py` adds descriptive labels to parameters that lack them:

```json
"description": {"type": "string", "description": "REQUIRED. The 'description' parameter (type: string)"}
```

Called in `server.py` when `resolved_model` contains "gemma".

## Testing

Verified with gemma-4-31b-it-8bit on Mac Studio M2 Ultra:
- **Before**: `delegate` tool call missing `description` field
- **After**: All required fields present across 5+ consecutive tests with varying tool counts

Fixes #741

### DIFF
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -980,6 +980,43 @@ def convert_tools_for_template(tools: Optional[List]) -> Optional[List[dict]]:
     return converted if converted else None
 
 
+def enrich_tool_params_for_gemma4(tools: list[dict]) -> list[dict]:
+    """Enrich tool parameter descriptions for Gemma 4 models.
+
+    Gemma 4's chat template renders tool definitions in a way that can
+    confuse the model when required parameters lack explicit descriptions,
+    especially when parameter names collide with schema keywords like
+    ``description``.  This ensures every required parameter has a non-empty
+    description that clearly labels it as a required argument, which
+    dramatically improves tool-call reliability.
+
+    This is a no-op for parameters that already have descriptions.
+    """
+    enriched = []
+    for tool in tools:
+        tool = dict(tool)
+        func = dict(tool.get("function", {}))
+        params = func.get("parameters", {})
+        if isinstance(params, dict) and "properties" in params:
+            params = dict(params)
+            props = dict(params.get("properties", {}))
+            required = set(params.get("required", []))
+            for pname, pdef in props.items():
+                pdef = dict(pdef)
+                if not pdef.get("description"):
+                    label = "REQUIRED. " if pname in required else ""
+                    pdef["description"] = (
+                        f"{label}The '{pname}' parameter"
+                        f" (type: {pdef.get('type', 'string')})"
+                    )
+                props[pname] = pdef
+            params["properties"] = props
+            func["parameters"] = params
+        tool["function"] = func
+        enriched.append(tool)
+    return enriched
+
+
 def format_tool_call_for_message(tool_call: ToolCall) -> dict:
     """
     Format a ToolCall object for inclusion in a message.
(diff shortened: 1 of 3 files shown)


jundot (commented):
Thanks for tackling this, the param dropping issue on Gemma 4 is really annoying so nice to see a fix.

I went through the code and spotted a few things.

`restore_gemma4_param_names` strips `param_` from every key that starts with it, not just the ones that were renamed. So if a tool already has something like `param_count`, it gets turned into `count`. Adding a check like `and k[len(prefix):] in _GEMMA4_COLLIDING_PARAMS` should be enough to fix this.

The streaming path checks `request.model` for the gemma detection but the non-streaming path uses `resolved_model`. If someone uses a model alias, enrichment happens but restore gets skipped in streaming, so the client receives `param_description` as-is.

Also the Anthropic (`/v1/messages`) and Responses (`/v1/responses`) endpoints have the same tool call flow but don't have the enrichment or restore added yet. Probably worth covering those too since the same bug would happen there.

On `_GEMMA4_COLLIDING_PARAMS`, i think starting with just `{"description"}` is safer. `name` and `type` are pretty common param names in real tools and renaming those might cause other issues. Can always add more later if they turn out to be a problem.

Last thing, would be great to have some tests for the new functions. `test_tool_calling.py` already has Gemma 4 test classes so the pattern is there.

PR #744, opened 2026-04-13
fix(admin): show per-model runtime cache size

### LINKED ISSUE #743: Bug: Runtime Cache dashboard repeats total SSD cache size for every model row
## Summary
In the admin dashboard Runtime Cache table, each model row's `Cache Size` and `Cache Files` can show the entire SSD cache partition usage instead of that model's own cache footprint.

## Current behavior
When multiple models are loaded, every row can repeat the same large global cache total (for example ~600 GB), even though each model should have a different cache footprint.

## Expected behavior
`Cache Size` and `Cache Files` should be model-scoped values for that row only.
The aggregate total can still reflect the full SSD cache usage, but per-row values should not reuse the global/shared totals.

## Likely root cause
`_build_runtime_cache_observability()` currently reads scheduler-level `ssd_cache` stats and copies `num_files` / `total_size_bytes` directly into each model row. Those stats appear to reflect the shared cache manager total rather than a per-model breakdown.

## Notes
I'm preparing a fix and regression test for the admin runtime cache observability path.

garnetlyx:
## Summary
Fix the Runtime Cache dashboard so each model row shows its own SSD cache footprint instead of repeating the shared cache total.

Closes #743.

## Root cause
`_build_runtime_cache_observability()` was copying scheduler-level `ssd_cache` totals directly into each model row. In the current runtime, the SSD cache manager can be shared across models, so those totals are global rather than model-scoped.

## Changes
- add `PagedSSDCacheManager.get_stats_for_model()` to aggregate indexed and hot-cache stats by `model_name`
- update admin runtime cache observability to prefer model-scoped SSD stats when available
- add a regression test proving two model rows can report different cache sizes even when the shared scheduler stats are identical

## Testing
- `.venv/bin/python -m pytest -q tests/test_admin_api_key.py`

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -2475,6 +2475,24 @@ def _build_runtime_cache_observability(
         elif not isinstance(ssd_stats, dict):
             ssd_stats = {}
 
+        ssd_manager = getattr(scheduler, "paged_ssd_cache_manager", None)
+        scheduler_model_name = getattr(getattr(scheduler, "config", None), "model_name", "")
+        if ssd_manager is not None and hasattr(ssd_manager, "get_stats_for_model"):
+            try:
+                scoped_ssd_stats = ssd_manager.get_stats_for_model(
+                    scheduler_model_name or model_id
+                )
+                if is_dataclass(scoped_ssd_stats):
+                    ssd_stats = asdict(scoped_ssd_stats)
+                elif isinstance(scoped_ssd_stats, dict):
+                    ssd_stats = scoped_ssd_stats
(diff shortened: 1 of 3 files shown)


jundot (approved):
Looks good. The per-model filtering logic is clean and the fallback to global stats on failure is a nice touch.

One minor note for later: `get_stats_for_model()` still returns global counters (hits, misses, evictions etc) inside `PagedSSDCacheStats`. Not a problem right now since the dashboard only reads the filtered fields, but if i ever expose those counters per-model i'll want to revisit this.

Merging as is.

Issue #326, opened 2026-03-20
Add support for Jina reranker architecture


rdner:
**Is your feature request related to a problem? Please describe.**
I tried to load the `jina-reranker-v3-mlx` model as a reranker model.
oMLX failed to load it with the `Unsupported reranker architecture: JinaForRanking.` error.

<img width="1307" height="546" alt="Image" src="https://github.com/user-attachments/assets/b887cb43-2ed5-458a-a9db-1c2c4db4eb7e" />

**Describe the solution you'd like**
Add support for the Jina reranker architecture.

Since I'm not an expert in this area, I don't really know if it's a reasonable request or how much effort it would require to implement this.

**Additional context**
Honestly, there are not that many options when it comes to reranker models available for oMLX.

According to Jina's benchmark [results](https://huggingface.co/jinaai/jina-reranker-v3), it's the best available reranker at the moment:

<img width="479" height="456" alt="Image" src="https://github.com/user-attachments/assets/1e6c1cf4-5eab-4a10-8c45-078c19f23cb0" />

and there is an official [MLX variant](https://huggingface.co/jinaai/jina-reranker-v3-mlx).

I think it deserves to be supported.

jundot:
[0.2.20rc1](https://github.com/jundot/omlx/releases/tag/v0.2.20rc1) has been uploaded as the final release candidate for the 0.2.20 release. Please give it a test and let me know if you run into any issues.

rdner:
Hey @jundot I finally got to testing. I've updated my omlx to 0.2.20 via homebrew and now it crashes on start with this error:

```
2026-03-25 17:32:53,141 - omlx.server - ERROR - GET /admin/dashboard → 500 (unhandled): unhashable type: 'dict'                                 
ERROR:    Exception in ASGI application                                                                                                  
Traceback (most recent call last):                                                                                                          
  File "/opt/homebrew/Cellar/omlx/0.2.20/libexec/lib/python3.11/site-packages/uvicorn/protocols/http/h11_impl.py", line 410, in run_asgi        
    result = await app(  # type: ignore[func-returns-value]                                                                                                                                                                                                                    
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                 
  File "/opt/homebrew/Cellar/omlx/0.2.20/libexec/lib/python3.11/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__ 
[...]

PR #745, opened 2026-04-13
fix(reranker): align JinaForRanking with Jina Reranker V3 listwise scoring

j-huang-rj:
## Problem

The prior Jina inference path treated JinaForRanking like a causal-LM reranker (score-token / last-logit extraction). This diverges from Jina V3 MLX’s actual reranker contract and produced ranking quality inconsistent with the upstream reference.

The issue is scoring path mismatch, not model loading.

## Upstream Jina Reranker V3 MLX contract

The official Jina Jina Reranker V3 MLX reranker uses:

- listwise prompt (all passages + query in one forward pass)
- `<|embed_token|>` positions for document representations
- `<|rerank_token|>` position for query representation
- hidden-state extraction at those positions
- `projector.safetensors` MLP (`linear1 -> ReLU -> linear2`)
- cosine similarity for relevance scores

### `omlx/models/reranker.py`

- Replaces Jina internal state (score-token fields → `_doc_embed_token_id`, `_query_embed_token_id`, `_jina_projector`)
- Reworks Jina load path: resolves special tokens and loads projector from `projector.safetensors`
- Replaces Jina scoring with listwise embedding scoring:
  - upstream-compatible prompt structure
  - special-token hidden-state extraction
  - projector application
  - cosine similarity scoring
- Public `MLXRerankerModel.rerank()` contract unchanged
- Non-Jina reranker paths unchanged

### `omlx/model_discovery.py`

- Classifies `JinaForRanking` as a directly supported reranker architecture
- Removes Jina from the CausalLM reranker heuristic bucket (no directory-name heuristic dependency)

### `tests/test_reranker_causal_lm.py`

  - token ID resolution fallback
  - prompt parity invariants
  - missing projector error handling
  - projector math
  - rerank output shape/order
  - dispatch/max_length semantics
### `tests/test_model_discovery.py`

  - Jina detection without requiring `rerank` / `reranker` in directory name

## Validation

- `python3 -m pytest tests/test_reranker_causal_lm.py` (`17 passed`)
- `python3 -m pytest tests/test_model_discovery.py` (`79 passed`)
- Manual `/v1/rerank` validation against the Jina Reranker V3 model

## Non-goals

- No public API changes
- No refactor of non-Jina reranker internals
- No changes to unrelated model families

## References

- Issue: #326
- Previous support PR: #331
- Jina model card: https://huggingface.co/jinaai/jina-reranker-V3-mlx
- Jina MLX rerank reference: https://huggingface.co/jinaai/jina-reranker-V3-mlx/blob/main/rerank.py

### DIFF
--- a/omlx/model_discovery.py
+++ b/omlx/model_discovery.py
@@ -113,6 +113,7 @@ EMBEDDING_ARCHITECTURES = {
 SUPPORTED_RERANKER_ARCHITECTURES = {
     "ModernBertForSequenceClassification",  # via mlx-embeddings
     "XLMRobertaForSequenceClassification",  # omlx native implementation
+    "JinaForRanking",  # Jina v3 listwise reranker
 }
 
 # CausalLM-based reranker architectures.
@@ -120,7 +121,6 @@ SUPPORTED_RERANKER_ARCHITECTURES = {
 # Detected by architecture + heuristic (model name or tokenizer hints).
 CAUSAL_LM_RERANKER_ARCHITECTURES = {
     "Qwen3ForCausalLM",
-    "JinaForRanking",  # Jina v3 reranker: uses <|score_token|> logits
 }
 
 # CausalLM-based embedding architectures.
--- a/omlx/models/reranker.py
+++ b/omlx/models/reranker.py
@@ -435,117 +435,31 @@ class MLXRerankerModel:
         )
 
     def _get_jina_hidden_states(self, input_ids):
-        """Extract final hidden states from mlx-lm wrappers/backbones."""
+        """Extract final hidden states from the Jina mlx-lm backbone."""
 
-        def _extract_hidden_states(outputs):
-            if outputs is None:
-                return None
-
-            hidden_states = getattr(outputs, "hidden_states", None)
-            if hidden_states is not None:
-                if isinstance(hidden_states, (list, tuple)):
-                    return hidden_states[-1]
-                return hidden_states
-
-            last_hidden_state = getattr(outputs, "last_hidden_state", None)
-            if last_hidden_state is not None:
-                return last_hidden_state
-
-            if isinstance(outputs, tuple):
-                for item in reversed(outputs):
-                    if isinstance(item, (list, tuple)) and item:
-                        candidate = item[-1]
-                        candidate_shape = getattr(candidate, "shape", None)
-                        if candidate_shape is not None and len(candidate_shape) >= 2:
-                            return candidate
-                    item_shape = getattr(item, "shape", None)
-                    if item_shape is not None and len(item_shape) == 3:
-                        return item
-
-            return None
-
-        def _try_call(target, use_hidden_states_flag: bool):
-            if target is None or not callable(target):
-                return None
-            try:
-                if use_hidden_states_flag:
-                    return target(input_ids, output_hidden_states=True)
-                return target(input_ids)
-            except TypeError:
-                return None
-            except Exception:
-                return None
-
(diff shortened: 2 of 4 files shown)


jundot:
Thanks for the PR. The scoring pipeline replacement looks correct overall - prompt structure, projector math, and cosine similarity all match the upstream Jina v3 MLX contract.

One thing i want to discuss before merging is `_get_jina_hidden_states`. The upstream reference does this in a single line:

```python
hidden_states = self.model.model([input_ids])
```

But this PR implements a ~120-line fallback cascade that tries 15+ extraction strategies (backbone, transformer, language_model, base_model, with/without `output_hidden_states`, tuple inspection, etc.).

My concern is less about it breaking now and more about silent wrong results later. If mlx-lm changes its model wrapper internals and the primary path fails, one of the fallback paths could return a tensor that happens to be 3D but isn't actually the hidden states we want. The broad `except Exception` in `_try_call` would swallow the real error, and the scoring pipeline would produce garbage scores without any visible error. This would be very hard to debug.

I think something closer to the upstream approach would be more maintainable:

```python
def _get_jina_hidden_states(self, input_ids):
    backbone = getattr(self.model, "model", None)
    if backbone is None or not callable(backbone):
        raise ValueError(
            "Could not find model backbone (model.model). "
            "The mlx-lm model wrapper may have changed."
        )

    try:
        outputs = backbone([input_ids])
    except TypeError:
        outputs = backbone(input_ids)

    # extract hidden states from outputs
    ...
```

This way, if the primary path breaks, we get a clear error pointing at the exact problem instead of silently falling through to a potentially wrong result.

Would you be open to simplifying this method? Happy to discuss if there's a specific reason for the defensive approach.

j-huang-rj:
Thanks for the review — I agree with the concern.

I took that approach because I wasn’t sure how much of the mlx-lm wrapper structure should be treated as a stable contract versus an implementation detail that just happens to work today.

The Jina reference does use `self.model.model([input_ids])`, and that gave me the main path, but I was hesitant to assume that future mlx-lm wrapper changes would always preserve the same attribute layout and return structure. Since this code is reaching into wrapper internals for a model-specific scoring path, rather than using a dedicated public hidden-state API, I ended up making that path more defensive than it should have been.

That said, I agree this was the wrong tradeoff for this part of the pipeline. In a correctness-sensitive reranker path, it’s better to keep the contract narrow and fail clearly than to recover through permissive fallbacks.

I pushed a follow-up that narrows `_get_jina_hidden_states()` to a single-path contract. One detail I adjusted locally is the call shape: upstream uses `self.model.model([input_ids])` because `tokenizer.encode()` there returns a plain Python token list. In our path, `_rerank_jina()` already constructs `input_ids` as a batched `mx.array([chunk_input_ids])`, so the local equivalent is `backbone(input_ids)` rather than wrapping it again.

I also updated the tests to cover the accepted shapes, invalid tensor ranks, and the main failure modes.

jundot:
Looks good, the follow-up commit addresses my concern about `_get_jina_hidden_states` nicely. Single-path contract with clear errors is exactly what i wanted.

Verified the scoring pipeline against the upstream jina-reranker-V3-mlx reference. Prompt structure, projector math, cosine similarity all match. The greedy chunking logic is a good addition since the upstream assumes everything fits in one pass but a serving environment needs to handle max_length limits.

One minor note for the future: the projector shape validation hardcodes (512, 1024) and (512, 512) which is fine for v3, but would need updating if jina ships a different hidden dim later. Not blocking.

Merging this in.

Issue #606, opened 2026-04-05
Issue : Gemma 4 26b a4b, bf16 - not matching official Livecode bechmark in omlx 0.34


Revive-Curiosity:
**Issue**
full precision model gemma 4 26b a4, bf16 is having only 47 % in livecode bench. 

**To Reproduce**
I om on omlx vesion 0.34

here is the model setting I put according to hugginface 

<img width="2508" height="214" alt="Image" src="https://github.com/user-attachments/assets/1053d5f1-77d3-485a-a6b8-8c8c6834d1a7" />

<img width="1794" height="1394" alt="Image" src="https://github.com/user-attachments/assets/81ac879a-2e76-4d94-a702-94a57b288aeb" />

as per the hugginface sampling parameters by google i(https://huggingface.co/google/gemma-4-26B-A4B-it)
<img width="1372" height="502" alt="Image" src="https://github.com/user-attachments/assets/6369594a-43c2-4f09-b184-5cf9d77456d8" />

**Expected Livecode Score**
Livecode bench - 77%

<img width="1478" height="494" alt="Image" src="https://github.com/user-attachments/assets/df74ba08-97a0-4703-a630-c0d0a879377e" />

**Result obtained**
only getting 46.7%
<img width="2496" height="260" alt="Image" src="https://github.com/user-attachments/assets/9730db98-a609-4740-a519-5416b796c736" />

What might be the Issue?
- either the model specfic settings are not used for running the benchmarks
- and the default server inference sampling parameters are used 

here see,

<img width="2628" height="1658" alt="Image" src="https://github.com/user-attachments/assets/4f809fea-6173-43ef-8545-911eb12b9ef5" />

Proposed Fix
- assuming, if the model specific sampling parameters is not working the issue
- then fix it, don't let the default settings for server override invidual model sampling configs
- also add option to choose model specific inference parameters in the server settings for specific models, reatining current default option for rest of other models

Device used
- 128 gb, m4 max macbook pro
- livecode test json linked below - for details
[gemma-4-26b-a4b-it-bf16_livecodebench.json](https://github.com/user-attachments/files/26490354/gemma-4-26b-a4b-it-bf16_livecodebench.json)

Solution Expected
- Reproduce 77% for livecode bench with gemma 4 26b a4b , bf16 
- Get full capability of this model for local coding

hunterx2591:
[...]

PR #748, opened 2026-04-14
feat: add dedicated Enable Thinking toggle with auto-detection

latent-variable:
## Summary

Adds a dedicated **Enable Thinking** toggle in per-model settings, separate from Thinking Budget. Auto-detects each model's default thinking behavior by inspecting its chat template at discovery time.

Related: #606, Gemma 4 underperforming on LiveCodeBench likely due to thinking not activating without explicit `enable_thinking=True`.

## Motivation

Models that support reasoning mode rely on the `enable_thinking` chat template kwarg, but there was no way to control it directly. The only workaround was enabling Thinking Budget, which is misleading since budget controls token limits, not whether thinking activates.

Different models handle the default differently: some think by default unless suppressed, others suppress unless explicitly enabled. Without a dedicated toggle and auto-detection, users can't see what their model does by default or change it.

## How it works

Passes the same `enable_thinking` kwarg that the benchmark's Thinking Mode switch already uses. Any model whose chat template respects this kwarg works with the toggle. It is not model-specific.

**Precedence:** per-request API kwargs > model settings toggle > auto-set from thinking budget > template default.

## Test plan

- [x] Model that defaults thinking OFF: toggle shows OFF with "using model default"
- [x] Model that defaults thinking ON: toggle shows ON with "using model default"
- [x] Override in both directions: verify generation behavior changes
- [x] Reset override: reverts to model default
- [x] Enable thinking budget without toggle: thinking auto-activates
- [x] Benchmark model switch syncs Thinking Mode toggle with model settings


jundot (approved):
Nice work on this! The auto-detection + dedicated toggle is a solid UX improvement. I tested the logic and it looks clean overall. Merging this as-is and i'll handle a few minor things in a follow-up commit.

A couple notes for context:

**Precedence comment vs actual behavior**

The comment in server.py says "Dedicated enable_thinking toggle takes precedence over chat_template_kwargs", but the code has `"enable_thinking" not in merged_ct_kwargs` which means if `chat_template_kwargs` already has `enable_thinking`, the toggle gets skipped. Not a real problem in practice since nobody would set both, but the comment is misleading. I'll fix the comment in a follow-up.

**detect_thinking_default heuristic**

The string matching covers Qwen and Gemma patterns well. Edge cases like `default(true)` or `default(False)` (capital F) won't match, but the function safely returns `None` for unknown patterns so there's no harm. I'll keep an eye on new model templates.

**Tests**

I'll add tests for `detect_thinking_default()` and the settings field in a follow-up commit.

Thanks for the contribution!

Issue #747, opened 2026-04-13
Homebrew formula fails to build: tokenizers pyo3 linker error on macOS ARM


jwiegley:
## Problem

`brew install jundot/omlx/omlx` fails on macOS ARM (Apple Silicon) because the `tokenizers` package cannot compile its Rust/pyo3 bindings. The formula forces source builds via `--no-binary tokenizers`, which triggers maturin compilation that fails at link time.

## Error

```
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)

error: could not compile `tokenizers-python` (lib) due to 1 previous error
💥 maturin failed
```

The missing symbols are Python C API symbols (`_PyBaseObject_Type`, `_PyDict_Type`, `_Py_NoneStruct`, etc.) that pyo3 extension modules expect to resolve at runtime via the Python interpreter.

## Root Cause

On macOS, shared libraries (`.dylib`/`.so`) linked against the Python C API need the linker flag `-undefined dynamic_lookup` so that Python symbols are resolved at runtime by the interpreter, not at link time. The formula sets `LDFLAGS` for headerpad but does not set `RUSTFLAGS` for cargo/maturin.

## Fix

Add this line to the `install` method in `Formula/omlx.rb`, after the existing `LDFLAGS` line:

```ruby
# pyo3 extension modules on macOS need dynamic_lookup so Python C API
# symbols are resolved at runtime by the interpreter, not at link time
ENV.append "RUSTFLAGS", "-C link-arg=-undefined -C link-arg=dynamic_lookup"
```

## Environment

- macOS 26.4.1 (arm64, Apple M4 Ultra)
- Homebrew 5.1.6
- Clang 21.0.0
- Rust (Homebrew): stable
- Python 3.11 (Homebrew)

Tested and confirmed working with this fix applied locally.

jundot:
Thanks for the detailed writeup and the fix suggestion.

I looked into this and `tokenizers` was actually added to `--no-binary` preemptively back in `3e21e77` when i was fixing the headerpad issue for `rpds-py` (#224). It was a "all rust packages need this" blanket addition, but tokenizers itself was never actually reported to cause the dylib ID rewrite failure. Its wheel ships `.abi3.so` (stable ABI) which homebrew doesn't try to rewrite in the first place.

[...]

PR #751, opened 2026-04-14
fix(engine): remove unsupported temperature kwarg from stream_dflash_generate

Wysie:
## Summary

`stream_dflash_generate()` in dflash_mlx does not accept a `temperature` parameter (verified against pinned commit `fc7101b`), but the DFlash engine passes it anyway, causing:

```
DFlash streaming generation error: stream_dflash_generate() got an
unexpected keyword argument 'temperature'
```

This removes the unsupported kwarg until dflash_mlx adds temperature support.

## Testing
- DFlash speculative decoding works after this change
- No impact on non-DFlash engines

Related: #747

### DIFF
--- a/omlx/engine/dflash.py
+++ b/omlx/engine/dflash.py
@@ -227,7 +227,6 @@ class DFlashEngine(BaseEngine):
                 max_new_tokens=max_tokens,
                 stop_token_ids=stop_ids,
                 prompt_tokens_override=prompt_tokens,
-                temperature=temperature,
             ):
                 event_type = event.get("event")
 


jundot:
Thanks for catching this! I had added temperature sampling support to my local dflash-mlx (the `sample_with_temperature()` function + `temperature` param on `stream_dflash_generate` / `generate_dflash_once`) but forgot to fork and pin to my patched version — the dep was still pointing at the original `bstnxbt/dflash-mlx@fc7101b` which doesn't have it.

Fixed in 4eb9c29 — switched the pin to `jundot/dflash-mlx@8e1df22` which includes the temperature patch. So this PR should no longer be needed, closing.

PR #757, opened 2026-04-14
feat: PlanarQuant3 KV cache

sooth:
## Context

oMLX runs LLMs on Apple Silicon via MLX. KV cache memory is a key bottleneck at long context lengths: on memory-constrained machines this limits max context or batch size.

This PR adds **PlanarQuant3**: 3-bit KV cache quantization using Givens rotation + Lloyd-Max centroids, matching the upstream [rotorquant/llama.cpp](https://github.com/rotorquant/llama.cpp) architecture. It achieves **speed parity with FP16** (0.94–1.02x) while compressing the attention-layer KV state by **82–164x** at the packed-storage level (for SSD offload or serialization), and **13–55% runtime memory savings** depending on context length.

## Follow-up update — PlanarQuant hardening + performance (`a32e597`, `0eaa0f3`)

This PR now includes a hardening pass that addresses the PlanarQuant audit findings and adds explicit performance/accuracy comparison coverage.

### Correctness and integration fixes

- Fixed `BatchPlanarQuantKVCache.decode_attention()` compatibility with patched SDPA explicit `keys_state` / `values_state` arguments.
- Fixed batch left-padding writes when a chunk crosses from negative padding offsets into valid token positions.
- Fixed `quantize_v=False` decode growth and state/meta-state restore so FP16 V remains intact.
- Tightened PlanarQuant3 validation to 3-bit mode and head dimensions divisible by 8 in cache, reference, Metal, model settings, and admin routes.
- Updated `nbytes` to report runtime dequant cache memory as well as packed state memory.
- Prevented global PlanarQuant cache-hook leakage into subsequently loaded non-PQ models.
- Marked PlanarQuant/TurboQuant/DFlash KV-related setting changes as requiring reload for already-loaded models.
- Wired scheduler memory-pressure checks to enable PlanarQuant tiled mode on active caches.

### Synthetic long-context comparison up to 128K

Model/mode: **synthetic KV-cache/attention microbench (no neural model)**. Existing real-model benchmark scripts still default to `mlx-community/Qwen3.5-4B-MLX-4bit`.

Command rerun after `0eaa0f3`:

```bash
uv run --no-sync python scripts/compare_planarquant.py --lengths 16384,32768,65536,131072 --heads 4 --batch 1 --iters 5 --json
```

`PQ warm vs FP16` is a latency ratio, so lower is faster. `New PQ speed vs FP16` is the reciprocal throughput-style speedup. `Speedup vs prior PR` compares the prior PR table ratio to the new ratio, so values above `1.00x` mean faster than the prior table.

| Mode | Context | Cosine | Packed compression | Runtime compression | Prior PQ warm vs FP16 | New PQ warm vs FP16 | New PQ speed vs FP16 | Speedup vs prior PR |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| K+V | 16K | 0.982991 | 5.12x | 0.84x | 0.53x | 0.78x | 1.29x | 0.68x |
| K-only | 16K | 0.999998 | 1.67x | 0.91x | 1.08x | 1.12x | 0.89x | 0.96x |
| K+V | 32K | 0.983806 | 5.12x | 0.84x | 1.11x | 0.97x | 1.03x | 1.15x |
| K-only | 32K | 0.999998 | 1.67x | 0.91x | 1.09x | 1.01x | 0.99x | 1.08x |
| K+V | 64K | 0.981405 | 5.12x | 0.84x | 1.19x | 1.03x | 0.97x | 1.15x |
| K-only | 64K | 0.999997 | 1.67x | 0.91x | 1.11x | 0.94x | 1.06x | 1.18x |
| K+V | 128K | 0.980908 | 5.12x | 0.84x | 1.19x | 0.97x | 1.03x | 1.22x |
| K-only | 128K | 0.999996 | 1.67x | 0.91x | 1.03x | 0.96x | 1.04x | 1.07x |

Raw timing detail:

| Mode | Context | FP16 decode ms | PQ warm decode ms | PQ cold decode ms | Prefill+finalize ms |
|---|---:|---:|---:|---:|---:|
| K+V | 16K | 0.466 | 0.362 | 7.879 | 1.160 |
| K-only | 16K | 0.295 | 0.331 | 1.376 | 0.199 |
| K+V | 32K | 0.470 | 0.456 | 5.702 | 0.195 |
| K-only | 32K | 0.443 | 0.446 | 2.397 | 0.175 |
| K+V | 64K | 0.648 | 0.668 | 8.976 | 0.175 |
| K-only | 64K | 0.621 | 0.585 | 3.575 | 0.154 |
| K+V | 128K | 0.990 | 0.965 | 13.975 | 0.177 |
| K-only | 128K | 1.047 | 1.004 | 7.170 | 0.164 |

Interpretation: accuracy and compression are unchanged. The latest run is faster than the prior PR table in **6 of 8 rows**, with the largest improvements at **128K K+V (1.22x)** and **64K K-only (1.18x)**. The 16K rows are slower than the prior rounded table, which is consistent with the MLX scheduling/warmup noise observed in repeated short synthetic runs.

Isolated hot-path microbench for the `dequantize_fused()` packed-input cast removal on shape `(1, 8, 4096, 128)`: old-style packed cast averaged `1.5245ms`, optimized conditional/no packed cast averaged `1.0244ms` (**1.49x avg speedup**, `max_abs_diff=0.0`).

### Additional validation

Hardening validation retained:

- `uv run --no-sync python ... py_compile` → `py_compile checked 17 Python files`
- `uv run --no-sync pytest tests/test_planarquant_*.py tests/test_admin_api_key.py -q` → `181 passed, 1 deselected`
- `git diff --check` → passed

Latest performance follow-up validation for `0eaa0f3`:

- `uv run --no-sync python ... py_compile` → `py_compile checked 4 Python files`
- `uv run --no-sync pytest tests/test_planarquant_kv_cache.py tests/test_planarquant_batch.py tests/test_planarquant_activation.py tests/test_planarquant_comparison.py tests/test_planarquant_tiled.py -q` → `89 passed`
- `uv run --no-sync pytest -q` → `3969 passed, 35 skipped, 55 deselected`
- `uv run --no-sync ruff check --select F401,F841 ...` → passed
- `git diff --check` → passed

---

### What it does

Quantizes the KV cache of attention layers to 3 bits per element using:
- **Givens rotation** with CUDA-matched cos/sin tables (D=64 rotation dimension)
- **Lloyd-Max centroids** (8 values, 7 midpoints) for 3-bit quantization
- **Packed storage**: 3-bit indices + 1-bit signs packed into contiguous `uint8` blocks (50 bytes per 128-element block)
- **Deferred quantization**: FP16 during prefill, bulk-convert on `finalize_prefill()`
- **Lazy-pack decode**: Per-token quantization is deferred until state serialization; decode hot path operates on FP16 dequant caches routed through Apple's MPS `scaled_dot_product_attention`
- **Batch support**: `BatchPlanarQuantKVCache` with `prepare`/`finalize` (right-padding roll), `filter`/`extend`/`merge`/`extract` lifecycle ops, packed-state helpers, and array-offset aware decode

### Architecture (6 commits)

| Commit | What |
|--------|------|
| `feat(planarquant): rebuild with upstream-matching architecture` | Packed storage, deferred quant, fused Metal SDPA |
| `feat(planarquant): add fused Metal quantize kernel` | Normalize → Givens → pack in single GPU dispatch |
| `perf(planarquant): cache dequantized K for O(1) decode` | Reuse FP16 rows, just append |
| `perf(planarquant): v2 deferred decode quantization + V dequant cache` | Zero per-token quantize cost, both K+V through MPS SDPA |
| `fix(planarquant): register cache types in scheduler's known sliceable types` | Scheduler compatibility |
| `feat(planarquant): batch KV cache + bench scripts` | BatchPlanarQuantKVCache lifecycle ops + 33 tests + 4 bench scripts |

### Benchmarks — Qwen3.5-4B MLX-4bit (Apple Silicon)

**Single-request, varying prompt and decode length (cos_sim = 1.000000 across all):**

| Prompt | Decode | FP16 decode tok/s | PQ3 decode tok/s | Cache FP16 → PQ3 MB | Cache ratio |
|---|---|---|---|---|---|
| ~5 tok | 128 | 8.04 | 8.09 | 59.9 → 52.3 | **-13%** |
| ~500 tok | 64 | 110.3 | 97.0 | 68.3 → 54.1 | **-21%** |
| ~500 tok | 128 | 108.4 | 85.7 | 68.3 → 54.5 | **-20%** |
| ~1.8k tok | 128 | 102.5 | 96.4 | 160.6 → 72.1 | **-55%** |

**Batch decode (64 decode steps, ~80 token prompt):**

| B | FP16 total tps | PQ3 total tps | Ratio | Cache MB ratio |
|---|---|---|---|---|
| 1 | 110.5 | 108.0 | 0.977x | 0.88x |
| 2 | 206.3 | 195.3 | 0.946x | 0.88x |
| 4 | 249.5 | 254.2 | **1.019x** | 0.88x |

**Batch-ops microbench (D=128, 3-bit, T=1024, B=8):**

| Op | Latency (ms) |
|---|---|
| evict_dequant | 0.006 |
| extend / merge | 0.26 / 0.64 |
| extract / filter | 0.40 / 0.54 |
| finalize | 3.18 |
| prepare | 0.15 |

**Per-token compression (D=128 per head):**

| Mode | Bytes | vs FP16 |
|---|---|---|
| FP16 K+V | 8192 | 1.00x |
| PQ K-only packed | 50 | **164x** |
| PQ K+V packed | 100 | **82x** |
| PQ + dequant caches (runtime) | 8292 | 0.99x |

Runtime memory is near-neutral at the attention-layer level because dequant caches are kept resident for decode speed. The packed buffers provide 82–164x compression for serialization (SSD offload, state save) and can be the sole resident form under memory pressure via `evict_dequant_caches()`.

### New files

| File | Purpose |
|------|---------|
| `omlx/cache/planarquant/__init__.py` | Public API exports |
| `omlx/cache/planarquant/constants.py` | CUDA rotation tables, Lloyd-Max centroids, packed layout constants |
| `omlx/cache/planarquant/reference.py` | Pure-MLX quantize/dequantize matching upstream C implementation |
| `omlx/cache/planarquant/metal_kernels.py` | Fused Metal kernels: dequant, quantize |
| `omlx/cache/planarquant/kv_cache.py` | `PlanarQuantKVCache` + `BatchPlanarQuantKVCache` |
| `omlx/patches/planarquant_cache.py` | `enable_planarquant_cache(bits, quantize_v)` monkey-patches `make_prompt_cache` |
| `scripts/bench_planarquant.py` | Single-request benchmark |
| `scripts/bench_planarquant_batch.py` | Batch decode + batch-ops microbench |
| `scripts/bench_scale_validation.py`, `scripts/bench_e2e_validation.py` | Scale / E2E validation |

### Integration points

- `omlx/patches/turboquant_attention.py` — Detects PQ caches, auto-finalizes prefill, routes decode through `cache.decode_attention()`
- `omlx/cache/type_registry.py` — Registers `PlanarQuantKVCache` for serialization
- `omlx/scheduler.py` — Registers PQ cache types as sliceable
- `omlx/cache/prefix_cache.py` — Block-level SSD offload: preserves PQ class_name through `store_cache`, reconstructs packed state via `_unpack_state` on load. Supports both `quantize_v=True` (K+V packed) and `quantize_v=False` (K packed, V as fp16). Packed tensors round-trip through safetensors at ~5× the token density of FP16.
- `omlx/model_settings.py` — `planarquant_kv_enabled`, `planarquant_kv_bits`, `planarquant_quantize_v` per-model fields
- `omlx/admin/routes.py` — ModelSettingsRequest pydantic fields, `list_models` response, PUT handler with mutual exclusion (disables TurboQuant when both are requested)
- `omlx/admin/templates/dashboard/_modal_model_settings.html` + `static/js/dashboard.js` — toggle, bits select, quantize_v sub-toggle. Clicking PQ or TQ toggle auto-disables the other
- `omlx/admin/i18n/{en,ko,zh,ja,zh-TW}.json` — 5 strings × 5 locales, translated
- `omlx/engine/batched.py` and `omlx/engine/vlm.py` — apply `apply_turboquant_attention_patch()` + `enable_planarquant_cache(bits, quantize_v)` on model load when the per-model setting is enabled; warns + disables TurboQuant if both are on

### Memory-pressure / tiled decode mode

For very long contexts where the FP16 dequant cache itself exceeds available memory, `PlanarQuantKVCache` provides a tiled decode path with online softmax accumulation (flash-attention recurrence):

- `decode_attention_tiled(q, scale, mask, tile_size)` — iterates packed history in tiles of `tile_size` tokens, dequantizes each tile via the fused Metal kernel, computes scores + running (max, sum, output) in fp32, composes across tiles. Handles GQA head repeat. Output is within fp32 online-softmax precision of the monolithic MPS SDPA path.
- `enable_memory_pressure_mode(tile_size)` — flushes lazy-packed decode rows, drops `_k_dequant_cache` / `_v_dequant_cache`, sets `memory_pressure=True`. Subsequent `update_and_fetch` calls eagerly quantize new K/V directly into `_k_packed` / `_v_packed`, so no FP16 dequant cache is ever materialized. Peak memory drops from O(offset · head_dim · fp16) to O(packed_bytes_per_token · offset + tile_size · head_dim · fp16).
- `decode_attention` auto-routes through the tiled path when `self.tile_size` is set.

Trade-off is explicit: tiled is slower at every context length where the monolithic path fits in memory (Python loop + per-tile dispatch overhead). It exists to extend the reachable context on memory-constrained machines.

#### Benchmarks — Qwen3.5-4B MLX-4bit (M3 Max 64GB)

| T | FP16 tok/s | PQ3 monolithic tok/s | PQ3 tiled tok/s | peak MB mono | peak MB tiled | cos_sim vs FP16 (tiled) |
|---|---|---|---|---|---|---|
| 1K | 117 | 111 | 85 | 2475 | 2434 | 0.989 |
| 4K | 115 | 105 | 57 | 2596 | 2453 | 0.981 |
| 8K | 98 | 94 | 37 | 2756 | 2479 | 0.992 |
| 16K | 103 | 79 | 23 | 3075 | 2530 | 0.992 |
| 32K | 88 | 54 | 14 | 3715 | **2633** | 0.967 |

At T=32K the tiled path uses ~1 GB less peak memory. On memory-constrained machines (e.g., 16GB M4 Mini) the crossover where tiled is required happens at much shorter contexts; reviewer reports flat throughput to 100K there.

---

## Tests

| Suite | Tests | Status |
|-------|-------|--------|
| `test_planarquant_constants` | 11 | ✅ Pass |
| `test_planarquant_reference` | 7 | ✅ Pass |
| `test_planarquant_activation` | 7 | ✅ Pass |
| `test_planarquant_kv_cache` | 16 | ✅ Pass |
| `test_planarquant_metal_kernel` | 11 | ✅ Pass |
| `test_planarquant_integration` | 1 | ✅ Pass (cos_sim=1.000000) |
| `test_planarquant_batch` | 33 | ✅ Pass |
| `test_planarquant_ssd` | 8 | ✅ Pass |
| `test_planarquant_tiled` | 11 | ✅ Pass |
| **Total** | **105 new** | **All pass** |

---

## Design decisions

1. **Why route through Apple MPS SDPA instead of custom Metal kernels?** Custom fused quantized SDPA kernels were 3–103x slower than Apple's proprietary MPS `scaled_dot_product_attention`. The solution: maintain FP16 dequant caches alongside packed PQ3 storage, route decode attention through MPS, and lazily pack decode rows only on state serialization.

2. **Why deferred/lazy-pack instead of per-token quantization?** Per-token quantization costs ~0.3ms per layer × N layers. Deferring until state save means the decode hot path has zero quantize cost. The packed buffers are only needed for serialization — they can be updated lazily.

3. **Why is runtime compression lower than storage compression?** Dequant caches are kept resident for decode speed — so attention-layer RAM is near-neutral while packed storage is 82–164x smaller. Under memory pressure `evict_dequant_caches()` can free the dequant caches and rebuild lazily.

4. **Why mutually exclusive with TurboQuant?** Both patch the SDPA dispatch path. Having two competing patches would cause undefined behavior. The mutual exclusion is enforced in `ModelSettings.__post_init__()`.

---

## Future work

- [ ] **PPL evaluation**: Compare PlanarQuant3 vs FP16 perplexity on Wikitext-2
- [ ] **Pure-attention model benchmarks**: Test on Llama-3.1-8B, Mistral-7B (expect higher total compression)
- [ ] **D=256 rotation constants**: Currently generated locally; should match upstream's actual tables


raiRaiyan:
Great! I was looking for this feature. BTW have you tried this: https://liranringel.github.io/ddtree/ ?

sooth:
Looks like someone beat me to adding dflash. Working on extracting just the
PlanarQuant3 then will look at ddtree.

user-23xyz:
Great work — we independently arrived at the same architecture (Givens rotation + Lloyd-Max + deferred quantize
   + MPS SDPA fallback) in our ForgeAttention v2 research (https://github.com/user-23xyz/forgeattention-v2).     
                                                                  
  Two findings that might be useful:                                                                             
                                                                  
  1. Rotation is unnecessary with per-channel normalization. We tested 100 tokens of greedy decode with and      
  without Givens rotation at 4-bit — output is token-for-token identical when using per-channel mean/std
  normalization instead of per-token L2 norm. vLLM independently confirmed this by dropping Hadamard transforms  
  from their DSA implementation. This would let you skip the rotation tables entirely and save ~256 FMAs per
  element on compress/decompress.
  2. Tiled prefill for constant-throughput long context. Instead of decompressing the full KV history per chunk
  (which creates an O(n) memory spike), decompress in 4K-token tiles with online softmax accumulation. This keeps
   throughput flat at 390 tok/s from 1K to 100K context on Qwen3.5-4B (M4 Mini 16GB), vs 2.1x degradation with
  full decompression.                                                                                            
                                                                  
  We also found that mx.fast.metal_kernel doesn't support threadgroup shared memory — threads can't communicate, 
  which is why fused SDPA kernels are slower than Apple's MPS path. Per-element kernels (each thread decompresses
   one element independently) work correctly.                                                                    
                                                                  
  Happy to contribute either of these as follow-up PRs if you're interested.

sooth:
Thanks — really useful review. I spent a session this week running each of the three findings to ground. Summary:

### 1. Per-channel normalization replacing rotation — investigated, **does not hold at 3-bit**

I A/B'd your proposed scheme (per-channel mean/std + no rotation) against the current (per-block L2 + Givens) at **3-bit**, on three data distributions: isotropic Gaussian, anisotropic (where per-channel norm *should* shine), and real Qwen3.5-4B layer-3/7 K tensors. Lower values mean worse reconstruction:

| data | current (L2 + Givens) | per-channel (plain) | per-channel + matched-std |
|---|---|---|---|
| isotropic Gaussian (D=64, T=1024) | 0.9831 | 0.8281 | 0.8357 |
| anisotropic (D=128, T=8192) | 0.9736 | 0.6770 | 0.8695 |
| real Qwen3.5-4B layer 3 | **0.9771** | 0.8419 | 0.9446 |
| real Qwen3.5-4B layer 7 | **0.9661** | 0.8246 | 0.9359 |

At 4-bit (16 bins) your result is unsurprising — the quantization budget is wide enough that the choice of normalization barely matters. At 3-bit (8 bins) the budget is ~4× tighter, and Givens rotation buys us roughly three nines of cos_sim even on the matched-std variant, which goes to the correctness of the reconstructed decode output. That gap tracks through PPL and greedy generation too.

vLLM dropping Hadamards from their DSA path is at a different bit budget and a different quantization target (indexer activations, not persistent KV). Interesting data point but not directly portable to PlanarQuant3.

So rotation is load-bearing at 3-bit, at least on our current kernel. If someone wants to push toward 2-bit via a different centroid design + per-channel norm that's a different rabbit hole — worth a follow-up experiment.

### 2. Tiled prefill / decode with online softmax — **implemented (landed in this PR)**

Pushed onto this PR as commit `67c8e84`. Architecture: `PlanarQuantKVCache.decode_attention_tiled(q, scale, mask, tile_size)` with the classic flash-attention recurrence (running max/sum/output in fp32, compose across tiles), plus `enable_memory_pressure_mode(tile_size)` which flushes lazy-packed decode rows, drops the FP16 dequant caches, and eager-packs new K/V in `update_and_fetch` so no full dequant cache is ever materialized. Packed tiles go through the fused `dequantize_fused` Metal kernel; attention itself is `mx.matmul` + online softmax (no MPS SDPA per tile).

Correctness: tiled cos_sim vs the monolithic MPS path stays > 0.9995 across every test we have, and > 0.967 vs FP16 ground truth at T=32K (matches the monolithic quantized path within 2e-3).

Speed and memory on Qwen3.5-4B-MLX-4bit (16GB budget wasn't the bottleneck for us — these are on a 64GB M3 Max):

| T | FP16 tok/s | PQ3 monolithic tok/s | PQ3 tiled tok/s | peak MB mono | peak MB tiled |
|---|---|---|---|---|---|
| 1K | 117 | 111 | 85 | 2475 | 2434 |
| 4K | 115 | 105 | 57 | 2596 | 2453 |
| 8K | 98 | 94 | 37 | 2756 | 2479 |
| 16K | 103 | 79 | 23 | 3075 | 2530 |
| 32K | 88 | 54 | 14 | 3715 | **2633** |

Matches your finding that peak memory stays flat-ish in the tiled path — at 32K we're at ~1GB under the monolithic path and the gap widens with T. It *also* matches the converse expectation: when the monolithic cache fits in RAM, monolithic wins (one MPS SDPA call beats N Python-dispatched tile matmuls). On our box the crossover where tiled would be required hasn't been hit yet at 32K; on your 16GB M4 Mini the crossover happens much earlier, which explains your flat 390 tok/s measurement. So we've made it opt-in via `enable_memory_pressure_mode()` rather than the default — right trade-off depends on available RAM.

Would happily accept a PR on top of v4 that extends this to batched/hybrid models.

### 3. `mx.fast.metal_kernel` and threadgroup shared memory

Followed up on this since it changes the story for fused kernels. MLX does support threadgroup memory and barriers from `metal_kernel` — the [custom Metal kernels doc](https://ml-explore.github.io/mlx/build/html/dev/custom_metal_kernels.html) shows the canonical pattern explicitly:

```metal
threadgroup float shared[THREADGROUP_MEM_SIZE];
threadgroup_barrier(mem_flags::mem_threadgroup);
```

…and [TurboQuant-MLX](https://github.com/arozanov/turboquant-mlx) uses a full threadgroup-scope WHT butterfly via exactly this surface. The Python binding ([`python/src/fast.cpp`](https://github.com/ml-explore/mlx/blob/main/python/src/fast.cpp)) imposes no restrictions on `threadgroup` declarations or `threadgroup_barrier`; it's raw MSL in the kernel body. The thing that is genuinely missing is a *dynamic* threadgroup allocation parameter (the CUDA backend has `shared_memory: int`, the Metal side does not) — but static `threadgroup float buf[N];` works fine.

The real reason our fused SDPA kernels lost to MPS is almost certainly Apple's proprietary `simdgroup_matrix` tile ops + their scheduler — threadgroup-memory restrictions aren't the bottleneck. The per-element dequant kernel and the per-tile tiled decode path both work cleanly, though, which matches what you saw.

---

Thanks again — the tiled finding in particular is a real win for long-context on memory-constrained machines, and the investigation into (1) was worth running even though the claim didn't hold at our bit budget. Tiled + memory-pressure mode is in commit `67c8e84` on this PR; benchmarks and the 11 new unit tests (`tests/test_planarquant_tiled.py`) are included. The `bench_planarquant_tiled.py` script reproduces the speed/memory table above.

sooth:
PlanarQuant follow-up pushed in `a32e597` (`fix: harden PlanarQuant KV cache`).

Summary:
- Fixed batch SDPA explicit-state compatibility, left-padding writes, batch array-offset state serialization, K-only V growth/state restore, runtime `nbytes`, strict 3-bit/head-dim validation, hook leakage, reload flags, and scheduler memory-pressure wiring.
- Added regression tests plus synthetic performance/accuracy comparison coverage up to 128K context in `scripts/compare_planarquant.py`.
- Latest synthetic 128K result: K+V cosine `0.980908`, K-only cosine `0.999996`; packed compression `5.12x` K+V / `1.67x` K-only. Runtime fast-path memory still includes dequant caches, so it is not a hot-cache memory win yet.

Validation:
- `uv run --no-sync python ... py_compile` -> `py_compile checked 17 Python files`
- `uv run --no-sync pytest tests/test_planarquant_*.py tests/test_admin_api_key.py -q` -> `181 passed, 1 deselected`
- `git diff --check` -> passed

sooth:
Rebased `feat/planarquant-dflash` onto current `origin/main` and force-pushed the cleaned history. PR merge state is now `CLEAN` at head `36e40a69d1732ce685044a57ef8ddb2f328e8407`.

Validation rerun:
- `git diff --check origin/main..HEAD`
- conflict marker scan over `omlx/`, `tests/`, `scripts/`
- `uv run --no-sync python` py_compile over 39 changed Python files
- `uv run --no-sync pytest tests/test_planarquant_*.py tests/test_admin_api_key.py tests/test_dflash.py tests/test_server_metrics.py -q` → 235 passed, 1 deselected
- `uv run --no-sync ruff check --select F401,F841 ...`

sooth:
Performance follow-up pushed in `0eaa0f3` (`perf(planarquant): reduce decode cache copies`).

What changed:
- Avoids redundant dtype casts/copies in PlanarQuant cache hot paths and `dequantize_fused()` packed input prep.
- Avoids the K-only `finalize_prefill()` V-buffer realloc/copy.
- Prewarms PlanarQuant quantize/dequant Metal kernels during activation.
- Fixes `dequantize()` after lazy decode-row updates and adds K+V/K-only regression coverage.

Long-context synthetic benchmark rerun:

```bash
uv run --no-sync python scripts/compare_planarquant.py --lengths 16384,32768,65536,131072 --heads 4 --batch 1 --iters 5 --json
```

| Mode | Context | Cosine | Prior PQ warm vs FP16 | New PQ warm vs FP16 | New PQ speed vs FP16 | Speedup vs prior PR |
| K+V | 16K | 0.982991 | 0.53x | 0.78x | 1.29x | 0.68x |
| K-only | 16K | 0.999998 | 1.08x | 1.12x | 0.89x | 0.96x |
| K+V | 32K | 0.983806 | 1.11x | 0.97x | 1.03x | 1.15x |
| K-only | 32K | 0.999998 | 1.09x | 1.01x | 0.99x | 1.08x |
| K+V | 64K | 0.981405 | 1.19x | 1.03x | 0.97x | 1.15x |
| K-only | 64K | 0.999997 | 1.11x | 0.94x | 1.06x | 1.18x |
| K+V | 128K | 0.980908 | 1.19x | 0.97x | 1.03x | 1.22x |
| K-only | 128K | 0.999996 | 1.03x | 0.96x | 1.04x | 1.07x |

The latest run is faster than the prior PR table in **6/8 rows**, with the largest wins at **128K K+V (1.22x vs prior)** and **64K K-only (1.18x vs prior)**. Isolated hot-path microbench for removing the packed-input cast in `dequantize_fused()` improved average latency from `1.5245ms` to `1.0244ms` (**1.49x avg speedup**, `max_abs_diff=0.0`).

Validation:
- `py_compile checked 4 Python files`
- targeted PlanarQuant/tiled tests → `89 passed`
- full suite → `3969 passed, 35 skipped, 55 deselected`
- focused `ruff --select F401,F841` → passed
- `git diff --check` → passed

PR body has been updated with the detailed table and timing breakdown.

sooth:
## PlanarQuant follow-up performance update

Prepared another follow-up performance patch after the previous decode-copy reduction work.

### What changed

- Added a capacity-aware tensor helper for PlanarQuant buffers so dequant-cache materialization, cache growth, and finalize packing avoid full-capacity zero allocation plus prefix-copy when the produced tensor already has the right logical length.
- Added adaptive default memory-pressure tiling: explicit tile sizes are preserved, but the default `4096` route now expands up to `16384` on long contexts to reduce Python/MLX tile-loop overhead.
- Added a uniform-offset batch write fast path for continuous batching, while keeping varied/left-padded batches on the existing per-row loop.
- Added regression coverage for adaptive tiling and the uniform batch-write fast path.

### Focused microbench wins

| Area | Case | Result |
|---|---:|---:|
| Dequant materialization, `cap=T` | up to 32K | up to `1.77x` faster |
| Dequant materialization, `cap=T+256` | up to 32K | up to `1.89x` faster |
| Finalize-pack path | up to 32K | up to `1.24x` faster |
| Adaptive memory-pressure tiled decode | 16K | `1.35x` faster vs forced tile `4096`, cosine `1.000000` |
| Adaptive memory-pressure tiled decode | 32K | `1.15x` faster vs forced tile `4096`, cosine `1.000000` |
| Uniform batch write | `B=8,L=1` | `2.92x` faster, max diff `0.0` |
| Uniform batch write | `B=16,L=1` | `5.08x` faster, max diff `0.0` |

### Updated long-context synthetic benchmark

Command:

```bash
uv run --no-sync python scripts/compare_planarquant.py --lengths 16384,32768,65536,131072 --heads 4 --batch 1 --iters 5 --json
```

| Context | Mode | FP16 decode ms | PQ warm decode ms | PQ latency vs FP16 | PQ speed vs FP16 | PQ cold decode ms | Cosine |
|---:|---|---:|---:|---:|---:|---:|---:|
| 16K | K+V | 0.781 | 1.047 | 1.34x | 0.75x | 11.745 | 0.982991 |
| 16K | K-only | 0.949 | 2.387 | 2.52x | 0.40x | 3.757 | 0.999998 |
| 32K | K+V | 3.128 | 0.758 | 0.24x | 4.13x | 9.020 | 0.983806 |
| 32K | K-only | 0.952 | 0.735 | 0.77x | 1.29x | 6.438 | 0.999998 |
| 64K | K+V | 1.335 | 2.341 | 1.75x | 0.57x | 13.531 | 0.981405 |
| 64K | K-only | 1.037 | 1.480 | 1.43x | 0.70x | 5.553 | 0.999997 |
| 128K | K+V | 2.670 | 1.354 | 0.51x | 1.97x | 18.756 | 0.980908 |
| 128K | K-only | 2.251 | 4.293 | 1.91x | 0.52x | 12.815 | 0.999996 |

Minimum cosine in this run: `0.9809083310743439`.

### Validation

- `py_compile`: 3 edited Python files checked
- `ruff F401,F841`: passed
- `git diff --check`: passed
- Full tests: `3972 passed, 35 skipped, 55 deselected`
- Full synthetic benchmark: `scripts/compare_planarquant.py --json` passed, min cosine `0.981656746410842`
- Batch ops benchmark: `scripts/bench_planarquant_batch.py --batch-ops-only` passed

Note: this update is committed locally as `a0bc909` and is ready to push to the PR branch when requested.

PR #773, opened 2026-04-15
feat: virtual model profiles via .virtual.yaml files

### LINKED ISSUE #341: [Feature Request] Add mutiple model setting profiles support
Currently, model settings support only one setting profile ( or preset), like below:

Image

I think it will be very helpful to support multiple profiles (or presets) and load models with different profiles:

Image

TipKnuckle:
Closes #341

## What this does

Adds support for **virtual model profiles** — a lightweight way to run the same model with different settings without loading it twice.

Drop a `*.virtual.yaml` file in your models directory:

```yaml
# ~/models/Qwen3.5-9B-nothink.virtual.yaml
source: Qwen3.5-9B-MLX-4bit
```

The filename stem becomes the model ID (`Qwen3.5-9B-nothink`). It appears in the model list and settings UI like any other model. Configure it independently via the existing settings modal — temperature, enable_thinking, system prompt, etc.

Multiple virtual models can point to the same source. Only **one engine is loaded in memory** regardless.

## Design

The key insight is that virtual models get their own `EngineEntry` with their own settings, but `get_engine()` delegates to the source. Because `resolve_model_id()` returns virtual IDs unchanged, all the `get_settings(resolved_model)` calls in `server.py` naturally return the virtual model's settings — **server.py is untouched**.

## Changes

- `model_discovery.py` — scan for `*.virtual.yaml` after physical discovery; circular reference guard (virtual→virtual rejected)
- `engine_pool.py` — virtual delegation in `get_engine()` before lock acquisition (avoids reentrant deadlock on asyncio.Lock); `source_model_id` field on `EngineEntry`
- `admin/routes.py` — pass `source_model_id` through in `list_models` response
- `_settings.html` — purple Virtual badge; load/unload buttons replaced with static status pill for virtual models
- `tests/test_model_discovery.py` — 9 new test cases (103 total, all passing)

## Not changed

- `server.py` — zero modifications
- Settings format — existing `*.settings.json` mechanism reused as-is
- No new API endpoints, no new data formats, no migration logic

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -1347,6 +1347,7 @@ async def list_models(is_admin: bool = Depends(require_admin)):
             "model_type": model_info.get("model_type", "llm"),
             "config_model_type": model_info.get("config_model_type", ""),
             "thinking_default": model_info.get("thinking_default"),
+            "source_model_id": model_info.get("source_model_id"),
             "last_access": model_info.get("last_access"),
         }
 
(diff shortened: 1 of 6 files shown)


TipKnuckle:
Newer PRs approach this more effectively, closing this.

PR #779, opened 2026-04-15
feat(menubar): improve Serving Stats layout and compact large counters

CHW0n9:
## Summary

This PR improves readability and scanability of the macOS menubar **Serving Stats** submenu in `packaging/omlx_app/app.py`.

## What Changed

- Align stats rows with a shared right value column across both sections:
  - `Session`
  - `All-Time`
- Center section headers (`Session`, `All-Time`) using deterministic custom menu-item views.
- Add compact formatting for large counters with 2 decimals:
  - `K`, `M`, `B`, `T`, `P`, `E`
- Show full raw numbers in hover tooltips for compacted counters:
  - `Total Tokens Processed`
  - `Cached Tokens`
  - `Total Requests`
- Use integer/Decimal-safe formatting (avoid float precision loss for large values).

## Scope

- Changed file:
  - `packaging/omlx_app/app.py`
- No API/CLI/config/schema changes.
- No dependency additions.

## Behavior Notes

- Stats fetch/refresh lifecycle is unchanged.
- Tooltip position/timing remains native AppKit behavior.

## Validation

- `python -m py_compile packaging/omlx_app/app.py`
- Manual visual check via local menubar preview with mocked stats.

### Before
Screenshot 2026-04-15 at 4 50 19 PM

### After
Screenshot 2026-04-15 at 4 19 03 PM

### DIFF
--- a/packaging/omlx_app/app.py
+++ b/packaging/omlx_app/app.py
@@ -8,6 +8,7 @@ import logging
 import platform
 import time
 import webbrowser
+from decimal import Decimal, InvalidOperation, ROUND_HALF_UP
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


jundot (approved):
Looks good, clean UI-only change with no impact on server/API/cache logic.

- Using `Decimal` instead of `float` for large number formatting is the right call. Avoids precision loss on big token counts.
- The `bool` guard in `_format_compact_count` is necessary since `isinstance(True, int)` returns `True` in python. Good catch.
- All AppKit operations wrapped in try/except with fallbacks, so a rendering glitch won't crash the whole menu. Nice.
- Shared tab stop across both Session and All-Time sections keeps the layout consistent.

No test impact since existing `test_omlx_app.py` only covers `ServerConfig`/`ServerManager`, not the UI layer.

LGTM, merging.

PR #789, opened 2026-04-15
fix: preserve tool_calls/tool_responses in VLM message formatting

### LINKED ISSUE #788: VLM text-only path strips tool_calls/tool_responses, breaking tool calling for Gemma 4 and other VLM models
### Problem

Commit 53a20e9 ("Keep VLM chat rendering path stable before first image") unified the VLM prompt rendering so that text-only messages also go through `_prepare_vision_inputs` → `_format_messages_for_vlm_template`. This ensures consistent prefix caching across image and non-image turns.

However, `_format_messages_for_vlm_template` runs **every** message through mlx-vlm's `get_message_json()`, which only returns `{"role": ..., "content": ...}`. This strips:

- `tool_calls` from assistant messages
- `tool_call_id` from tool result messages  
- `tool_responses` from Gemma-native tool result messages

As a result, the Jinja chat template never sees tool call/result data. The model receives tool results as invisible `role:tool` messages (skipped by the template's main loop) with no preceding `tool_calls` to trigger the forward-scan that would render them. The model behaves as if every tool call returned null, causing infinite retry loops.

### Impact

All tool calling on VLM models (Gemma 4, Qwen-VL, etc.) is broken on text-only turns — which is the majority of agentic tool-use conversations.

### Reproduction

Any multi-turn tool calling conversation with a VLM model. The model will loop retrying the same tool call because it never sees the result.

### Fix

In `_format_messages_for_vlm_template`, pass `role:tool` messages and assistant messages carrying `tool_calls`/`tool_responses` through verbatim instead of running them through `get_message_json()`. Only user/system/plain-assistant messages need VLM image-token formatting.

### Introduced in

53a20e9 ("Keep VLM chat rendering path stable before first image")

TipKnuckle:
### Problem

Since 53a20e9, text-only messages to VLM models go through `_format_messages_for_vlm_template`, which runs every message through `get_message_json()`. This strips `tool_calls`, `tool_call_id`, and `tool_responses` fields — the chat template never sees tool results, causing infinite tool-call retry loops.

### Change

Pass `role:tool` messages and assistant messages carrying `tool_calls`/`tool_responses` through verbatim instead of running them through `get_message_json()`. Only user/system/plain-assistant messages need VLM image-token formatting.

### Testing

- `test_tool_calling.py` — 137 passed
- `test_vlm_engine.py` — 44 passed  
- Manual end-to-end testing with Gemma 4 E4B + pi agent: multi-turn tool calling works correctly, model reads tool results and responds appropriately

Fixes #788

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -595,17 +595,27 @@ class VLMBatchedEngine(BaseEngine):
             if msg_num_images > 0:
                 image_message_ranges.append((idx, msg_num_images))
 
-            formatted_messages.append(
-                get_message_json(
-                    model_type,
-                    content,
-                    role,
-                    skip_image_token=role != "user" or msg_num_images == 0,
-                    skip_audio_token=True,
-                    num_images=msg_num_images,
-                    num_audios=0,
+            # Preserve tool-related messages verbatim so the chat
+            # template receives tool_calls, tool_call_id, and
+            # tool_responses fields.  get_message_json() strips these,
+            # which makes tool results invisible to the model.
+            if role == "tool" or (
+                role == "assistant"
+                and (msg.get("tool_calls") or msg.get("tool_responses"))
+            ):
+                formatted_messages.append(msg)
+            else:
+                formatted_messages.append(
+                    get_message_json(
+                        model_type,
+                        content,
+                        role,
+                        skip_image_token=role != "user" or msg_num_images == 0,
+                        skip_audio_token=True,
+                        num_images=msg_num_images,
+                        num_audios=0,
+                    )
                 )
-            )
 
         return formatted_messages, image_message_ranges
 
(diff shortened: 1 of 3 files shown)


jundot:
Nice catch! I was actually investigating tool calling issues on gemma 4 models right now and this turned out to be the root cause.

The symptom i saw was the model repeating the exact same tool call in an infinite loop (e.g. `Glob{pattern:"*"}` over and over). Took me a while to realize tool results were silently getting stripped before reaching the template.

Thanks for tracking this down, merging now.

TipKnuckle:
Yeah my Gemma 4 models were looping over tool calls nonsensically all the sudden. Issue was hard to find buried in all those recent commits, but glad we got it and you merged it quickly.

catman002:
原来如此啊，我还以为我的agent写的有问题

Issue #617, opened 2026-04-06
Bug Report: oMLX v0.3.4, gemma-4-26b-a4b-it-6bit Tool Calling Fails in Claude Code


lihywh:
### Environment
- **Hardware**: MacBook Pro, M1 Max, 64GB unified memory
- **oMLX version**: latest
- **Model**: `gemma-4-26b-a4b-it-6bit`
- **Client**: Claude Code v2.1.85
- **Chat template params**: `tool_call_parser=gemma`, `tools_in_user_message=true`
- **Effort**: `/effort high`

### Description

When using `gemma-4-26b-a4b-it-6bit` as the Claude Code backend, tool calls are frequently not triggered even with explicit instructions. The model returns text responses instead of invoking tools.

### Steps to Reproduce

1. Load `gemma-4-26b-a4b-it-6bit` in oMLX
2. Connect Claude Code to oMLX endpoint
3. Set `/effort high`
4. Run the following prompts in sequence:

```
# Prompt 1 - works
审视当前目录内容

# Prompt 2 - fails (no tool call triggered)
根据 *-bencmark.md 文档内容推荐合适使用本地运行ClaudeCode的模型

# Prompt 3 - fails
读取 20260405-bencmark.md 文件内容，给出适使用本地运行ClaudeCode的模型

# Prompt 4 - fails even with explicit tool name
使用 mcp__context7__resolve-library-id 查询 mlx-lm 的库ID
```

### Observed Behavior

- Prompts 2-4 produce **no output at all** — the model receives the instruction but generates zero tokens and terminates
- Simple single-step tool calls (e.g., `ls` / directory listing) work correctly
- Multi-step or goal-oriented instructions that require chaining tool calls consistently fail
- Explicitly naming the tool (`mcp__context7__resolve-library-id`) still does not trigger execution

### Expected Behavior

The model should invoke the appropriate tools (e.g., `Read`, `mcp__context7__resolve-library-id`) to fulfill the instruction.

### Additional Observations

- **gemma-4-31b-it-4bit** (dense model) does NOT reproduce this issue — tool calls work reliably
- The failure appears specific to the **MoE architecture** of `gemma-4-26b-a4b`
- Setting `tools_in_user_message=true` improved single-step tool call reliability but did not fix multi-step failures
- `SpecPrefill` (with `gemma-4-e2b-it-4bit` as draft) caused complete tool call breakdown — disabling it is required
[...]

PR #799, opened 2026-04-16
fix: attach tool_responses to same assistant message as tool_calls

latent-variable:
## Summary

- `extract_gemma4_messages()` was placing `tool_responses` on a **separate** assistant message from `tool_calls`
- The Gemma 4 chat template checks for `tool_responses` on the **current** message (the assistant turn with `tool_calls`) before falling back to a forward-scan for `role='tool'` messages
- With the split layout, both template paths missed — producing a corrupt bare `<|tool_response>` tag and causing the model to loop on the same tool call indefinitely
- Fix: attach `tool_responses` to the same `out_msg` that already carries `tool_calls`

Companion to #789 which fixed the VLM engine side. This fixes the message extractor side so the structure is correct before it reaches the template.

Related issues: #617, #666, #788

## Important: chat template versions

The Gemma 4 chat template has gone through several revisions over the past week as tool calling, thinking, and multi-turn features were being finalized. Earlier template versions had issues with tool response handling that compounded with this bug. I recommend testing with the latest Unsloth models which include the most up-to-date chat templates:

- **MLX (recommended for oMLX):** [unsloth/gemma-4-31b-it-UD-MLX-4bit](https://huggingface.co/unsloth/gemma-4-31b-it-UD-MLX-4bit)
- **GGUF:** [unsloth/gemma-4-31B-it-GGUF](https://huggingface.co/unsloth/gemma-4-31B-it-GGUF)

## Test plan

- [x] Updated `test_gemma4_messages.py` — all assertions now expect `tool_responses` on the same message as `tool_calls`
- [x] Tested with Gemma 4 31B IT (unsloth UD-MLX-4bit) on multi-turn agentic tool-use conversations (file reads, code analysis) — model sees tool results and stops looping

### DIFF
--- a/omlx/adapter/gemma4.py
+++ b/omlx/adapter/gemma4.py
@@ -205,14 +205,15 @@ def extract_gemma4_messages(
                 i += 1
 
             if tool_responses:
-                processed.append(
-                    {
-                        "role": "assistant",
-                        "content": "",
-                        "tool_responses": tool_responses,
-                        _PRESERVE_BOUNDARY_KEY: True,
-                    }
-                )
+                # Attach tool_responses to the SAME assistant message that
+                # has tool_calls.  The Gemma 4 chat template checks for
+                # tool_responses on the current message (lines 261-267)
+                # BEFORE falling back to a forward-scan for role='tool'
+                # messages (lines 268-302).  Putting them on a separate
+                # assistant message causes both paths to miss, producing a
+                # corrupt bare <|tool_response> tag and making the model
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the fix. This is the companion to #789 and the analysis checks out.

The gemma 4 template looks for `tool_responses` on the same assistant message that has `tool_calls`. The old code put them on a separate message so both template paths missed, causing infinite tool call loops. Attaching directly to `out_msg` is the right call.

Tests look good, merging.

PR #814, opened 2026-04-16
feat: preserve thinking across turns for Qwen 3.6+ incl. external clients

latent-variable:
## Summary

Adds `preserve_thinking` support for Qwen 3.6+ so the model can recall its own `<think>` reasoning across chat turns — both in the built-in chat and from external OpenAI/Anthropic-compatible clients (Pi, Claude Code, etc.).

Related: #748.

## Motivation

Qwen 3.6+ chat templates strip `<think>` blocks from historical assistant turns by default, so the model loses its prior chain-of-thought between turns and KV prefix reuse suffers. The model provider explicitly recommends enabling `preserve_thinking` — from the [Qwen3.6-35B-A3B model card](https://huggingface.co/Qwen/Qwen3.6-35B-A3B):

> This capability is particularly beneficial for agent scenarios, where maintaining full reasoning context can enhance decision consistency and, in many cases, reduce overall token consumption by minimizing redundant reasoning. Additionally, it can improve KV cache utilization, optimizing inference efficiency in both thinking and non-thinking modes.

Wiring `preserve_thinking=True` server-side isn't enough though: external OpenAI/Anthropic-compatible clients receive clean `content` (thinking is split into `reasoning_content` by oMLX's reasoning parser) and echo back only `content` on the next turn — so there's nothing to preserve.

This PR wires both halves: the server auto-enables `preserve_thinking` when the template supports it, and the API layer reconstructs `<think>` blocks from client-echoed `reasoning_content` / Anthropic `thinking` blocks before templating.

##  Comparison 
Screenshot 2026-04-16 at 11 13 52
Screenshot 2026-04-16 at 11 38 01

## How it works

**Backend (`feat` commit):** detect `preserve_thinking` support from the chat template at discovery time; add a `ModelSettings` override; auto-enable when thinking is active; pipe through chat/messages/responses endpoints.

**API input reconstruction (`fix` commit):**
- `openai_models.py`: add `reasoning_content` to the `Message` request model.
- `utils.py`: when an assistant message has `reasoning_content`, merge into `<think>\n{reasoning}\n</think>\n\n{content}` before templating.
- `anthropic_utils.py`: stop discarding `thinking` content blocks across all three `convert_anthropic_to_internal` sites (including the native-tool-calling assistant branch, which had no handling at all); append in source order to preserve multi-block ordering.
- Harmony/gpt-oss input path left unchanged (different format).

Precedence matches #748: per-request kwargs > model settings > auto-set > template default.

## Test plan

**Unit (`tests/test_api_utils.py`)** — 8 new cases covering string/list/None content, non-assistant passthrough, missing `reasoning_content` no-op, single + multi-block Anthropic ordering, and the native-tool-calling path. Full fast suite (`pytest -m "not slow"`) green.

**End-to-end methodology** — `Qwen3.6-35B-A3B-8bit`, two-turn recall test:
1. Turn 1: "generate two random 20-digit numbers, give me only one of the two and nothing else" → returns one number in `content`, puts both in `reasoning_content` / `thinking`.
2. Turn 2: "now give me the second number."

| Endpoint | With reasoning echoed | Control (no reasoning) |
|---|---|---|
| OpenAI `/v1/chat/completions` | exact recall ✅ | fabricates different number ✅ |
| Anthropic `/v1/messages` | exact recall ✅ | thinking says "I don't store the second number", fabricates ✅ |

Controls confirm the fix is load-bearing — without reconstruction the model has no access to its prior reasoning; with it, the exact number comes back.

**Regression-safety**: `reasoning_content` defaults to `None`, so clients that don't send it and non-thinking models are untouched.

Assisted by Claude Opus 4.7

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -1348,6 +1348,7 @@ async def list_models(is_admin: bool = Depends(require_admin)):
             "model_type": model_info.get("model_type", "llm"),
             "config_model_type": model_info.get("config_model_type", ""),
             "thinking_default": model_info.get("thinking_default"),
+            "preserve_thinking_default": model_info.get("preserve_thinking_default"),
             "last_access": model_info.get("last_access"),
         }
 
--- a/omlx/api/anthropic_utils.py
+++ b/omlx/api/anthropic_utils.py
@@ -192,6 +192,14 @@ def convert_anthropic_to_internal(
                                     "arguments": tool_input,
                                 },
                             })
+                        elif block_type == "thinking":
+                            # Reconstruct <think> block so preserve_thinking can keep it.
+                            # Append in source order; Anthropic places thinking before
+                            # text/tool_use blocks, so natural ordering already puts
+                            # <think> first in the reassembled message.
+                            thinking_text = block_dict.get("thinking", "")
+                            if thinking_text:
+                                text_parts.append(f"<think>\n{thinking_text}\n</think>")
                         elif block_type == "document":
                             text_parts.append(_decode_document_block(block_dict))
                     msg_dict = _build_message_from_parts(role, text_parts, image_parts) or {
@@ -238,7 +246,13 @@ def convert_anthropic_to_internal(
                                     block_dict.get("content", ""), image_parts
                                 )
                         elif block_type == "thinking":
-                            continue
+                            # Reconstruct <think> block so preserve_thinking can keep it.
+                            # Append in source order — Anthropic emits thinking blocks
+                            # before the text blocks they precede, so appending keeps
+                            # the natural ordering intact across multiple blocks.
+                            thinking_text = block_dict.get("thinking", "")
+                            if thinking_text:
+                                text_parts.append(f"<think>\n{thinking_text}\n</think>")
                         elif block_type == "document":
                             text_parts.append(_decode_document_block(block_dict))
                     msg_dict = _build_message_from_parts(role, text_parts, image_parts)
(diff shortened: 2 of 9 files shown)


latent-variable:
Screenshot 2026-04-18 at 18 52 28

This model is really amazing btw

jundot:
LGTM overall — this is a real fix for a real problem and all three conversion paths (OpenAI, Anthropic native-tool, Anthropic non-native) are covered. Planning to merge this. A few observations I ran into while reviewing that I think are worth follow-up commits rather than blockers:

1. `preserve_thinking_default` detected in `model_discovery.detect_preserve_thinking` gets plumbed all the way through `DiscoveredModel → EngineEntry → admin routes`, but the auto-set in `server.py` doesn't actually gate on it. It unconditionally sets `preserve_thinking=True` whenever `enable_thinking` isn't explicitly False, so the detection flag is load-bearing-looking while being decorative. Works today because Jinja silently ignores unknown kwargs, but a strict template down the road could raise on it.

2. I checked the actual Qwen3.6-35B-A3B chat template locally and it has first-class support for \`message.reasoning_content\` as a message field:

   \`\`\`jinja
   {%- if message.reasoning_content is string %}
       {%- set reasoning_content = message.reasoning_content %}
   {%- else %}
       {%- if '</think>' in content %}
           {%- set reasoning_content = content.split('</think>')[0]... %}
           {%- set content = content.split('</think>')[-1].lstrip('\n') %}
       {%- endif %}
   {%- endif %}
   \`\`\`

   So the \`<think>...</think>\` reconstruction in content is actually the second-class fallback path. Passing \`reasoning_content\` as a separate field on the message dict would use the template's first-class branch and sidestep the round-trip issues entirely. Not strictly a correctness issue since the fallback works, but worth considering.

3. Related to 2: OpenAI reconstruction uses \`\n\n\` and Anthropic uses \`\n\`, and neither matches the \`.strip()\`-ed original output from \`extract_thinking\`. Since the PR cites KV prefix cache reuse, the token-level divergence eats into prefix hits. Switching to the field-based path in 2 makes this moot.

4. \`/v1/responses\` auto-sets \`preserve_thinking=True\` but its input converter (\`convert_responses_input_to_messages\`) has no reasoning/thinking reconstruction. Either worth a line in the description noting the Responses API input format is out of scope, or a follow-up to plumb it through.

None are blockers. I'll merge this and open follow-ups for 1 and 2/3 (they share the same detection gate and pair naturally), with 4 as a separate issue if it comes up.

Issue #791, opened 2026-04-15
[Feature Request] Auto-bypass DFlash for image requests on multimodal models


czyrichard:
**Is your feature request related to a problem? Please describe.**

Yes. I use a multimodal Qwen3.5 model for OCR with image inputs sent as structured `image_url` content using `data:image/...;base64,...`.

With DFlash disabled, OCR works correctly.
With DFlash enabled, the same model loses OCR / vision behavior.

Environment:

- oMLX: `v0.3.5`
- target model: `froggeric/Qwen3.5-35B-A3B-Uncensored-FernflowerAI-MLX-8bit`
- draft model: `z-lab/Qwen3.5-35B-A3B-DFlash`

**Describe the solution you'd like**

I would like multimodal models to keep OCR / vision support when DFlash is enabled.

Either of these would solve the issue:

- DFlash works correctly for image requests on multimodal models, or
- requests containing image inputs automatically bypass DFlash and use `VLMBatchedEngine`.

**Describe alternatives you've considered**

- Disabling DFlash entirely for this model: OCR works, but then I lose DFlash for text-only requests.
- Using separate workflows for text and OCR: workable, but less convenient than having oMLX handle this automatically.

**Additional context**

Example request format:

```json
{
  "role": "user",
  "content": [
    { "type": "text", "text": "Please OCR this image." },
    {
      "type": "image_url",
      "image_url": {
        "url": "data:image/png;base64,..."
      }
    }
  ]
}
```

Issue #753, opened 2026-04-14
TTS response_format=opus ignored, always returns WAV


ethanliuhc-ux:
v0.3.5.dev1

Version: v0.3.5.dev1
Model: Qwen3-TTS-12Hz-1.7B-Base-bf16

Steps:

curl -X POST http://localhost:8000/v1/audio/speech \
  -H "Content-Type: application/json" \
  -d '{"model": "Qwen3-TTS-12Hz-1.7B-Base-bf16", "input": "hello", "voice": "alloy", "response_format": "opus"}' \
  --output test.opus && file test.opus

expected : Ogg data, Opus audio
got : test.opus: RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, mono 24000 Hz

jundot:
@ethanliuhc-ux response_format opus (plus mp3, flac, and pcm) now works on the non-streaming path in the latest release, available at https://github.com/jundot/omlx/releases. Could you give it a try?

PR #818, opened 2026-04-17
fix(dflash): route image requests to VLM fallback; fix(tts): honour response_format

Chuhan1112:
## Summary

Fixes two independent bugs with no new external dependencies.

### fix(dflash): route image requests to VLM fallback engine (#791)

**Root cause:** `DFlashEngine.chat()` / `stream_chat()` called `_apply_chat_template()` on all messages before any type-checking. Multi-modal content parts (`image`, `image_url`, `input_image`) were silently flattened to text and dropped — the model never saw the images, with no error raised.

**Fix:** Added `_messages_have_images()` static method that inspects message `content` lists for image parts. Both `chat()` and `stream_chat()` now detect images *before* template processing and route to `_fallback_engine` (VLMBatchedEngine) via the existing `_evict_dflash_and_start_fallback()` path — symmetric with the long-context fallback already in place.

### fix(tts): honour `response_format` in `/v1/audio/speech` (#753)

**Root cause:** `create_speech` always returned `Response(content=wav_bytes, media_type="audio/wav")` regardless of `request.response_format`.

**Fix:**
- `audio_utils.py`: added `convert_wav_to_response_format(wav_bytes, fmt) -> (bytes, media_type)` with:
  - `wav` / `pcm` — stdlib `wave` module only, zero extra deps
  - `mp3` / `opus` / `flac` / `aac` — ffmpeg subprocess; raises `RuntimeError` with install hint if ffmpeg is absent
  - `SUPPORTED_RESPONSE_FORMATS` constant for validation
- `audio_routes.py`: validates `response_format` early (before synthesis), then converts and returns with the correct `Content-Type`.

## Test plan

- [ ] DFlash + image message → request reaches VLMBatchedEngine, image is processed correctly
- [ ] DFlash + text-only message → no change in behaviour
- [ ] `POST /v1/audio/speech` with `response_format=mp3` → returns `audio/mpeg`
- [ ] `POST /v1/audio/speech` with `response_format=pcm` → returns `audio/pcm` raw frames
- [ ] `POST /v1/audio/speech` with `response_format=wav` (default) → unchanged behaviour
- [ ] `POST /v1/audio/speech` with unknown format → 400 with list of supported formats
- [ ] mp3/opus request without ffmpeg installed → 500 with install instructions

### DIFF
--- a/omlx/api/audio_routes.py
+++ b/omlx/api/audio_routes.py
@@ -223,6 +223,22 @@ async def create_speech(request: AudioSpeechRequest):
             detail=f"Model '{resolved_model}' is not a text-to-speech model",
         )
 
+    # Validate response_format early so the user gets a clear error
+    # before we spend time synthesizing audio.
+    from omlx.engine.audio_utils import (
+        SUPPORTED_RESPONSE_FORMATS,
+        convert_wav_to_response_format,
+    )
+    fmt = (request.response_format or "wav").lower()
+    if fmt not in SUPPORTED_RESPONSE_FORMATS:
+        raise HTTPException(
+            status_code=400,
+            detail=(
+                f"Unsupported response_format '{fmt}'. "
+                f"Supported values: {sorted(SUPPORTED_RESPONSE_FORMATS)}"
+            ),
+        )
+
     ref_audio_path = None
     try:
         # Write decoded audio to temp file if voice clone requested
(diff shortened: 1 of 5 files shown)


jundot:
Thanks for the PR, and sorry this sat so long. The tts half is covered on main now (#2231 plus a follow-up): soundfile, which the audio extra already ships via librosa, turned out to bundle mp3/opus/flac encoders, so ffmpeg wasn't needed.

The dflash half is still wanted though. #791 is still open, and your approach of detecting image parts before templating and reusing the existing fallback path still fits, but dflash.py has changed a lot since April. If you're still up for a rebased dflash-only PR I'll review it right away. Closing this bundle in the meantime.

PR #364, opened 2026-03-24
feat: JANG implementation

AlexTzk:
***

## Summary

This PR adds support for JANG quantized models to oMLX. JANG is an Apple Silicon MLX-based architecture that improves model quality when using low-bit quantization (e.g., 2-bit) by applying mixed-precision quantization — preserving critical attention layers at higher precision (6–8 bit) while quantizing expert MLP layers to 2-bit.

## Background

On MoE models, attention is only 1–5% of parameters but controls 100% of coherence. Standard MLX quantization compresses everything equally, which breaks low-bit models. JANG solves this by:

- **Attention:** 6–8 bit (preserves coherence)
- **Expert MLP:** 2–4 bit (95%+ of params, can absorb errors)

## Why
JANG quantization is quite useful for running modern large language models on Apple Silicon hardware. Without it, MoE models with 256+ experts either crash or produce NaNs below 4-bit quantization due to attention layers being compressed too aggressively. This enables models like Qwen3.5-397B (512 experts) and Nemotron-3-Super-120B to run efficiently on M-series Macs at ~2-3 bytes per parameter, making high-capacity reasoning models accessible without enterprise-grade GPU infrastructure. The integration ensures oMLX users can leverage these state-of-the-art quantized models through oMLX. 

## Changes

| File | Changes |
|---|---|
| `omlx/engine/jang.py` | Created `JANGLoader` class — full JANG model loader with Nemotron-H support, bfloat16 for large models, VLM support |
| `omlx/model_discovery.py` | Added JANG model detection via `jang_config.json` |
| `omlx/exceptions.py` | Added `JANGLoadError` and `JANGDependencyError` exceptions |
| `omlx/engine_pool.py` | Added `engine_type == "jang"` integration |
| `omlx/server.py` | Added `oq_manager` to `ServerState` |
| `pyproject.toml` | Added `jang[mlx]>=0.1.0` dependency |
| `packaging/venvstacks.toml` | Added `jang-tools` dependency |
| `omlx/admin/routes.py` | Minor import fix |
| Tests | Added comprehensive JANG tests (`test_jang_vlm.py`, `test_model_discovery.py`) |

## Files Changed

| File | Lines |
|---|---|
| `.gitignore` | +1 (JANG config) |
| `omlx/admin/routes.py` | 5 |
| `omlx/engine/jang.py` | 675 *(new file)* |
| `omlx/engine_pool.py` | 10 |
| `omlx/model_discovery.py` | 63 |
| `omlx/server.py` | 1 |
[...]

PR #820, opened 2026-04-17
Adds JANG quantized model support to oMLX's batched engine.

yohann-bearzi:
## Changes

**model_discovery.py**: Detect jang_config.json and route JANG models to batched engine, bypassing VLM misdetection from inherited preprocessor configs.

**engine/batched.py**: When JANG detected, load via jang_tools._load_jang_v2() instead of mlx_lm.load(), bypassing jang_tools' VLM detection (incompatible with batched cache).

## Requirements

pip install 'jang[mlx]'

## Testing

- Tested on Qwen3.6-35B-A3B with all 15 JANG profiles (JANG_1L through JANG_6K)
- Text generation works end-to-end through oMLX continuous batching
- Published quants load unmodified — no config surgery needed

## Notes

- JANG VLM support (has_vision: true) planned as follow-up PR
- Requires jang-tools package

### DIFF
--- a/omlx/cli.py
+++ b/omlx/cli.py
@@ -430,40 +430,32 @@ def diagnose_menubar() -> int:
     except (subprocess.SubprocessError, FileNotFoundError) as e:
         print(f"Menubar app:    check failed ({e})")
 
-    log_dir = Path.home() / "Library" / "Application Support" / "oMLX" / "logs"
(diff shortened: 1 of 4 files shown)


AlexTzk:
https://github.com/jundot/omlx/pull/364 

There's a PR already raised for this implementation. No further feedback was received on it though so work has been paused.

PR #824, opened 2026-04-17
fix(admin): handle symlinked models in delete_hf_model endpoint (v2)

Bahtya:
## Rebased replacement for #669

Closes #669 (old branch had 150 files diverged from main).

### Problem

The `delete_hf_model` endpoint fails for symlinked models because:
1. `resolve()` follows symlinks, so path validation rejects models whose symlink target is outside `model_dir`
2. `shutil.rmtree()` follows symlinks and deletes the target directory contents instead of just removing the symlink

### Fix (1 file, +19/-5 lines)

**Path traversal validation** — separate handling:
- **Symlinks**: Validate that `model_path.parent.resolve()` is within `parent_model_dir.resolve()`. We only `unlink()` the symlink, so the target location is irrelevant — the symlink itself must live inside our controlled directory. Using `parent.resolve()` gives full path normalization (prevents `..` traversal) without following the symlink.
- **Regular dirs**: Keep existing `resolve()` check unchanged.

**Deletion** — separate handling:
- **Symlinks**: `Path.unlink()` — removes only the symlink, target stays intact
- **Regular dirs**: Keep `shutil.rmtree()` with macOS `._` file compatibility

### Security considerations

The maintainer correctly identified that `absolute()` doesn't normalize `..`, enabling path traversal. This fix avoids that pitfall:
- For symlinks, we resolve the **parent directory** (not the symlink path itself), which normalizes all `..` components in the path leading to the symlink
- The symlink must physically exist inside `model_dir` for the parent check to pass
- We never follow the symlink during validation or deletion

### Testing

- Symlink model: symlink at `models/org/model-name` → `/data/huggingface/org/model-name` — validated by parent check, deleted with `unlink()`
- Regular dir model: `models/org/model-name/` — validated by existing `resolve()` check, deleted with `rmtree()`
- Path traversal attempt: `../../../etc/passwd` — rejected because `parent.resolve()` normalizes to `/etc` which is not within `model_dir`

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -3421,13 +3421,23 @@ async def delete_hf_model(
         raise HTTPException(status_code=404, detail="Model not found")
 
     # Validate path traversal against parent model directory
+    # Symlinks require separate validation: the symlink itself must reside
+    # inside model_dir, but its target may point anywhere.  We only ever
+    # unlink the symlink (never follow it), so validating the parent is safe.
     try:
-        if not model_path.resolve().is_relative_to(parent_model_dir.resolve()):
-            raise HTTPException(status_code=400, detail="Invalid model name")
+        if model_path.is_symlink():
+            # Ensure the symlink lives inside the model directory.
+            # We resolve the *parent* (which cannot be a symlink we created)
+            # and verify the symlink name is legitimate.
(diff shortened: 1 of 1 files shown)


Bahtya:
Hi maintainers, friendly ping on this PR. It fixes symlinked model handling in the \delete_hf_model\ endpoint. Would appreciate a review when you have a chance.

PR #830, opened 2026-04-17
feat: add sortable table UI for Browse Models

omniwired:
- Convert Browse Models list to sortable table with columns: Name, Size, Downloads, Likes, Action
- Add clickable column headers for sorting (name, size, downloads, likes)
- Add sort direction toggle (asc/desc) with chevron indicators
- Add translation keys for table headers
- Fix download button centering in action column
- Include filtering and sorting parameters in HF model search API

Demo:

Before

https://github.com/user-attachments/assets/e29c54ec-afec-4d53-bd89-396b48f6512b

After:

https://github.com/user-attachments/assets/95f3b6c5-a76f-4be6-aa63-a94c23b9c258


jundot:
Thanks for the table refactor, direction looks great. A few things worth fixing before merge:

1. Regression: the OOM/tight-fit badge and the params count are gone from both the trending/popular and search tables. The memory warning was a useful guard so it'd be nice to bring it back.
2. Bug in `_DTYPE_TO_QUANT`: `Q6_K`/`Q5_K`/`Q3_K` are mapped to int4/int8/int4 which isn't right (they're 6/5/3-bit).
3. Looks like `python omlx/admin/build_css.py` and `scripts/normalize_i18n.py` weren't run. The new `models.table.*` keys landed in English on `ja.json`/`ko.json`/`zh.json`/`zh-TW.json`, and there's no `tailwind.css` diff for the new utility classes.
4. `_QUANT_LABELS` is unused, and the new backend sort params (`largest`/`smallest`/`sort_by_size`/`sort_ascending`) aren't called from the JS at all, happy to either wire them up or drop them in a follow-up.
5. The new `quantization` field is in the response but never rendered. Would be a nice column/badge since filtering by it is the headline feature of the PR.

Minor: `min_size` is in the backend but not in the UI (only `max_size` is exposed).

omniwired:
Thank you so much for your comment I will deal with this in one week.

omniwired:
Thanks for the thorough review! All points addressed:

1. **OOM/tight-fit badge + params count** — restored in both the trending/popular table and the search results table. Badges appear in the Action column, params show below the model name.

2. **`_DTYPE_TO_QUANT` bit-width bug** — fixed. Q6_K → int6, Q5_K → int5, Q3_K → int3 (they were incorrectly int4/int8/int4).

3. **CSS + i18n** — ran both `build_css.py` and `normalize_i18n.py`. The tailwind build previously wasn't picking up utility classes because it was run from the wrong directory — fixed. `models.table.*` keys are now translated in ja/ko/zh/zh-TW.

4. **`_QUANT_LABELS` removed** (it was leftover from an earlier approach). **Sort params wired** — `sort_by_size` and `sort_ascending` are now sent from JS when the user selects "Largest Size" or "Smallest Size" from the sort dropdown.

5. **Quantization column** — added to both tables as a pill badge (e.g. `int4`, `fp16`, `bf16`) between Name and Size.

6. **`min_size`** — exposed in the filter panel alongside `max_size`, with JS wiring and i18n key.

jundot:
Thanks for the fixes, looks good. Merging now.

A few small follow-ups I noticed after a closer look — going to land them in the next release so no extra work for you:
- wire quantization detection into trending/popular tabs (currently the Quant column is blank there)
- sync the sort dropdown with the table column sort (the table re-sorts on top of the backend result, so largest/smallest is effectively a no-op right now)
- normalize 4bit/8bit filter to match the int4/int8 detected labels

jundot:
Merged in 3d687fd (squashed via local merge due to conflict in tailwind.css).

PR #844, opened 2026-04-18
fix(tool_calling): recover naked <function=...> calls emitted by Qwen3-Coder

mrtkrcm:
## Summary

Recover Qwen-style naked `<function=...>...</function>` tool calls when the model omits the outer `<tool_call>` wrapper.

## Changes

- Add a fallback parser for naked function blocks.
- Parse `<parameter=...>` values with JSON coercion when possible.
- Strip recovered function markup and stray tool wrapper fragments from content.
- Add regression coverage for single and multiple naked function calls.

## Local validation

Built and installed from this branch into `/Applications/oMLX.app` (`0.3.8.dev2`), with port `8801` owned by the visible app process.

- Focused tests: `201 passed, 12 deselected` for `tests/test_tool_calling.py`, `tests/integration/test_e2e_streaming.py`, and `tests/test_admin_profiles_api.py`.
- Live model: `Qwen3-Coder-30B-A3B-Instruct-4bit` forced `read_file`; response returned structured `tool_calls` and no raw `<function=...>` markup.
- Benchmark smoke row: `Qwen3-Coder-30B-A3B-Instruct-4bit` load `2.8s`, canary `25.1 tok/s`, code `23.7 tok/s`, tool `OK` in `2.3s`.
- Control model: `Ternary-Bonsai-8B-mlx-2bit` returned `ready`; benchmark tool `OK` in `0.6s`.

Note: benchmark host was not clean; preflight saw active desktop/client load, so throughput is smoke data only.

## Test Plan

- `uv run pytest tests/test_tool_calling.py -q`
- `python3 -m py_compile omlx/api/tool_calling.py tests/test_tool_calling.py`

### DIFF
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -29,6 +29,12 @@ from .openai_models import FunctionCall, ResponseFormat, ToolCall, ToolDefinitio
 
 logger = logging.getLogger(__name__)
 
+# Pre-compiled patterns for the naked <function=...> fallback (see parse_tool_calls).
+# Hoisted to module scope to avoid re-compilation on every request.
+_NAKED_FUNCTION_RE = re.compile(r"<function=([^\s>]+)>(.*?)</function>", re.DOTALL)
+_NAKED_PARAMETER_RE = re.compile(r"<parameter=([^\s>]+)>\s*(.*?)\s*</parameter>", re.DOTALL)
+_STRAY_TOOL_CALL_RE = re.compile(r"</?tool_call>")
+
 
 def _serialize_tool_call_arguments(arguments: Any) -> str:
     """Serialize parser output to a JSON-object arguments string.
@@ -535,6 +541,39 @@ def parse_tool_calls(
     # Fallback: parse XML <tool_call> tags (GLM, Qwen, generic formats)
     if "<tool_call>" in cleaned_text:
         return _parse_xml_tool_calls(cleaned_text)
+
+    # Fallback: naked <function=name>...</function> emitted by Qwen3-Coder
+    # when the model skips the outer <tool_call> wrapper it was trained on.
+    # Upstream parser (mlx_lm.tool_parsers.qwen3_coder) requires the wrapper;
+    # this branch recovers the structured call when the wrapper is absent.
+    if "<function=" in cleaned_text and "</function>" in cleaned_text:
+        naked_tool_calls: List[ToolCall] = []
+        for match in _NAKED_FUNCTION_RE.finditer(cleaned_text):
+            func_name = match.group(1)
+            body = match.group(2)
+            arguments: Dict[str, Any] = {}
+            for pm in _NAKED_PARAMETER_RE.finditer(body):
+                key = pm.group(1)
+                val = pm.group(2).strip()
+                try:
+                    arguments[key] = json.loads(val)
+                except (json.JSONDecodeError, ValueError):
+                    arguments[key] = val
+            naked_tool_calls.append(
+                ToolCall(
+                    id=f"call_{uuid.uuid4().hex[:8]}",
+                    type="function",
+                    function=FunctionCall(
+                        name=func_name,
+                        arguments=json.dumps(arguments, ensure_ascii=False),
+                    ),
+                )
+            )
+        if naked_tool_calls:
+            cleaned = _NAKED_FUNCTION_RE.sub("", cleaned_text)
+            # Strip any stray wrapper fragments the model emitted without pairs
+            cleaned = _STRAY_TOOL_CALL_RE.sub("", cleaned).strip()
+            return cleaned, naked_tool_calls
 
     # Fallback: namespaced tool_call tags (e.g. <minimax:tool_call>)
     ns_match = re.search(r"<([A-Za-z_][\w.-]*):tool_call>", cleaned_text)
--- a/tests/test_tool_calling.py
+++ b/tests/test_tool_calling.py
@@ -1203,6 +1203,57 @@ class TestParseToolCallsSyntaxError:
         assert tool_calls is None or len(tool_calls) == 0
 
 
+class TestParseNakedQwenFunctionCalls:
+    """Regression coverage for Qwen-Coder omitting the outer <tool_call> tag."""
+
+    def _qwen_tok(self):
+        tok = MagicMock(spec=[])
+        tok.has_tool_calling = True
+        tok.tool_call_start = "<tool_call>"
+        tok.tool_call_end = "</tool_call>"
+        tok.tool_parser = None
+        return tok
+
+    def test_naked_function_call_is_recovered(self):
+        tok = self._qwen_tok()
+
+        text = (
+            "I'll inspect it.\n"
+            "<function=read_file>\n"
+            "<parameter=path>/tmp/example.py</parameter>\n"
+            "<parameter=limit>20</parameter>\n"
+            "</function>\n"
+            "</tool_call>"
+        )
+
+        cleaned, tool_calls = parse_tool_calls(text, tok)
+
+        assert cleaned == "I'll inspect it."
+        assert tool_calls is not None
+        assert len(tool_calls) == 1
+        assert tool_calls[0].function.name == "read_file"
+        assert json.loads(tool_calls[0].function.arguments) == {
+            "path": "/tmp/example.py",
+            "limit": 20,
+        }
+
+    def test_multiple_naked_function_calls_are_recovered(self):
+        tok = self._qwen_tok()
+
+        text = (
+            "<function=first><parameter=value>true</parameter></function>"
+            "<function=second><parameter=name>alpha</parameter></function>"
+        )
+
+        cleaned, tool_calls = parse_tool_calls(text, tok)
+
+        assert cleaned == ""
+        assert tool_calls is not None
+        assert [tc.function.name for tc in tool_calls] == ["first", "second"]
+        assert json.loads(tool_calls[0].function.arguments) == {"value": True}
+        assert json.loads(tool_calls[1].function.arguments) == {"name": "alpha"}
+
+
 class TestParseBracketToolCalls:
     """Tests for bracket-style tool call parsing (issue #159)."""
 


guruswami-ai:
+1 on this from production experience. We carried this exact patch in our internal deploy branch (0.4.4-final through 0.4.5.dev) on a 5× M3 Ultra 512GB fleet for ~2 weeks of heavy agentic traffic — no regressions, and it eliminated the failure mode where Qwen3-Coder emits a naked `<function=...>` call that the parser passes through as plain text, stalling agentic clients mid-loop.

We've since moved to vanilla release tags (now v0.5.1) and just promoted Qwen3-Coder-Next to our production agentic-coder role, which puts this parsing gap back on our critical path — so we'd love to see this land.

Two data points that may help it along:
- Verified today: the diff still cherry-picks **clean onto v0.5.1** (50ee2464), no conflicts with the 0.5.x tool-calling changes.
- #2221 (raw XML tool calls leaking to content on qwen3.6-27b) looks like the same class of missing-recovery-path problem, so demand for this isn't isolated.

damsnotdam:
Independent corroboration from another fleet, with measurements — plus one form this PR does not cover.

We hit exactly the failure this PR fixes, and we reached the same conclusion by reading the parser
chain *before* finding the PR. So this is corroboration, not a re-post: the approach here (fall back
to the naked `<function=…>` block) is the one the source pushed us to as well.

**Setup.** oMLX 0.5.3 (brew tap `jundot/omlx`; installed copy byte-identical to tag `v0.5.3` —
`omlx/api/tool_calling.py` `390fee43`), mlx-lm 0.31.3, mlx 0.32.0, macOS 26.5.2, Mac Studio M4 Max
64 GB, `Qwen3-Coder-30B-A3B-Instruct-4bit` (mlx-community), streaming, `tools` declared on every
request.

### The rate, measured on both sides (2026-07-29)

| arm | failures | n |
|---|---|---|
| engine, raw `/v1/chat/completions` call | 2 | 10 |
| through our internal agentic harness | 4 | 10 |
| engine, raw call with `top_p: 1` (what OpenCode actually sends) | 2 | 10 |

Read it as **"of the order of 20–40% per tool call"**, not 40%: at n=10 the two arms do not
separate, and writing 40% would give ten shots a precision they do not have. `top_p` is not the
cause — the `top_p: 1` arm returns the same 2/10.

At ~20% per call — the low end — a ten-call agentic run survives **11%** of the time (0.8¹⁰). This
is not a rare-edge tax; it ends runs, and it ends them silently: `finish_reason=stop`, no warning.

### Proof without HTTP, without loading weights (~2 s, deterministic, re-run today on 0.5.3)

```python
import logging, os
from pathlib import Path
from mlx_lm.tokenizer_utils import load
from mlx_lm.tool_parsers.qwen3_coder import parse_tool_call

MODEL = os.environ.get("MODEL_PATH", "/path/to/Qwen3-Coder-30B-A3B-Instruct-4bit")

W = []
logging.getLogger().addHandler(type("H", (logging.Handler,), {"emit": lambda s, r: W.append(r)})())
logging.getLogger("omlx").setLevel(logging.WARNING)

tok = load(Path(MODEL))
print("tool parser inferred from the chat template : %s   markers %r / %r"
      % (tok.tool_parser.__module__.split(".")[-1], tok.tool_call_start, tok.tool_call_end))

TOOLS = [{"function": {"name": "write", "parameters": {"properties": {
    "filePath": {"type": "string"}, "content": {"type": "string"}}}}}]
BLOCK = ("<function=write>\n<parameter=filePath>\n/tmp/fetch.py\n</parameter>\n"
         "<parameter=content>\nimport requests\n</parameter>\n</function>")
PROSE = "Now I'll create the fetch.py file:\n"
ORPHAN = "\n</tool_call>"

print("\n[stage 2] mlx-lm parser ALONE")
for name, text in {
    "A1  the bare <function=...></function> block": BLOCK,
    "A2  the same, preceded by its prose": PROSE + BLOCK,
    "A3  the same, followed by an orphan closing tag": PROSE + BLOCK + ORPHAN,
}.items():
    try:
        print("  %-48s -> %r" % (name, parse_tool_call(text, TOOLS)))
    except Exception as e:
        print("  %-48s -> REJECTED %s: %s" % (name, type(e).__name__, e))

print("\n[stage 3] omlx parse_tool_calls — what a request actually sees")
for name, text in {
    "B  prose + NO opening tag + orphan closing tag": PROSE + BLOCK + ORPHAN,
    "C  same, with no closing tag at all": PROSE + BLOCK,
    "D  opening tag restored before <function=": PROSE + "<tool_call>\n" + BLOCK + ORPHAN,
}.items():
    n = len(W)
    cleaned, calls = parse_tool_calls(text, tok, TOOLS)
    args = calls[0].function.arguments if calls else ""
    print("  %-48s -> tool_calls=%d  content=%4d/%4d o  intact=%-3s  warnings=%d  args=%d o"
          % (name, len(calls or []), len(cleaned), len(text.strip()),
             "yes" if cleaned == text.strip() else "no", len(W) - n, len(args)))
    if calls:
        print("       prose kept=%r   args=%s" % (cleaned, args))
```

```
tool parser inferred from the chat template : qwen3_coder   markers '<tool_call>' / '</tool_call>'

[stage 2] mlx-lm parser ALONE
  A1  the bare <function=...></function> block     -> {'name': 'write', 'arguments': {'filePath': '/tmp/fetch.py', 'content': 'import requests'}}
  A2  the same, preceded by its prose              -> {'name': 'write', 'arguments': {'filePath': '/tmp/fetch.py', 'content': 'import requests'}}
  A3  the same, followed by an orphan closing tag  -> REJECTED ValueError: No function provided.

[stage 3] omlx parse_tool_calls — what a request actually sees
  B  prose + NO opening tag + orphan closing tag   -> tool_calls=0  content= 173/ 173 o  intact=yes  warnings=0  args=0 o
  C  same, with no closing tag at all              -> tool_calls=0  content= 160/ 160 o  intact=yes  warnings=0  args=0 o
  D  opening tag restored before <function=        -> tool_calls=1  content=  34/ 185 o  intact=no   warnings=0  args=59 o
       prose kept="Now I'll create the fetch.py file:"   args={"filePath": "/tmp/fetch.py", "content": "import requests"}
```

What to read, in order:

- the parser inferred from the chat template really is `qwen3_coder`, with paired markers;
- **A1/A2 accepted** — the payload is perfectly readable *one stage below*, with or without leading prose;
- **A3 rejected** — the `$` anchor added by the mlx-lm port refuses an otherwise readable payload as
  soon as a stray `</tool_call>` follows it;
- **B and C identical** — what a request sees today: 0 tool calls, `content` returned verbatim,
  **0 warnings**. At stage 3 the stray closing tag no longer matters, because nothing is paired anyway;
- **D** — the same bytes with the opening tag restored yield **1 tool call**, arguments intact, prose preserved.

### Why this is a port regression, not a tolerance request

| # | where | what the parser keys on |
|---|---|---|
| 1 | `qwen3coder_tool_parser.py`, Qwen's reference parser published with the model | **`<function=`**, with an explicit alternative for unterminated forms: `<function=(.*?)</function>` \| `<function=(.*)$` |
| 2 | `mlx_lm/tool_parsers/qwen3_coder.py:14`, headed "Modified from" that exact file | **`<function=`**, second alternative **dropped**, `$` anchor **added**: `<function=(.*?)</function>$` |
| 3 | `omlx/api/tool_calling.py:1131-1135` (0.5.3) | the **pair** `<tool_call>…</tool_call>`; the stage-2 parser is only ever called on the **inside** of a matched block (l. 1146) |

Two successive tightenings. The guard that would have caught every one of our failures is
`<function=` — present in all of them. The parser able to read the payload is installed on the
machine; it is simply never reached. That is the gap this PR closes.

### Two notes that may help review

- **Your `_STRAY_TOOL_CALL_RE` cleanup earns its place, and the invariant you chose is the right
  one.** In our sample the orphan `</tool_call>` was present in **3 of the 4** engine-side failures
  and absent in 1 — so it is *not* an invariant, and a fallback (or a client-side detector) keyed on
  it would miss cases. Keying on `<function=` is correct. Per case C above, stage 3's verdict is
  identical with or without the stray tag.
- **The fallback requires `</function>`**, so it does not cover a payload cut *mid-argument*
  (`max_tokens` truncation, or a stop sequence landing inside the envelope). That form loses
  everything on the streaming path instead of arriving as prose, and it is a distinct defect in the
  stream filter — filed separately as #2420. The two fixes are independent: neither one covers the
  other.

### Pre-empting the two ways nearby reports have been closed

- **#2221** was closed on the conclusion that the failing requests arrived **without `tools`**. Ours
  declare `tools` on every call, and the block above bypasses HTTP entirely: it calls the parser
  directly, with the real tokenizer and a tool list.
- **#2405** was closed because that conversion's chat template had **no tool support at all**. Ours
  does: the block prints `qwen3_coder` and the `<tool_call>` / `</tool_call>` pair, and case **D**
  returns a structured call from the very same bytes.

Happy to run further probes on this hardware (0.5.3, unpatched) if anything here would help get a
review.

jrehm:
Third corroboration, on the current release (0.6.4) and a path not yet tested here: the Anthropic-compatible endpoint.

**Setup.** oMLX 0.6.4, mlx-lm 0.31.3, mlx 0.32.0 (identical to the versions in this thread), macOS 27.0 (26A5421a), Apple M5 (Mac17,4) 32GB, `Qwen3-Coder-30B-A3B-Instruct-4bit` (mlx-community). Not raw HTTP this time — `omlx launch claude` (oMLX's own Claude Code integration), which routes Claude Code's `/v1/messages` streaming traffic through oMLX with `ANTHROPIC_BASE_URL`. `tools` declared on every request (Claude Code's full ~20-tool schema).

**Same shape as the original report, confirmed on a third endpoint.** #2420 and this PR both note `/v1/chat/completions`; #2420 explicitly flags `/v1/messages` as having "different recovery logic" and untested. It isn't better there — same naked `<function=...>` leak, still unrecognized:

~~~
I'll check the model information to confirm I'm running on the local oMLX server.

<function=Bash>
<parameter=command>
echo $ANTHROPIC_BASE_URL && echo $ANTHROPIC_DEFAULT_SONNET_MODEL
</parameter>
</function>
</tool_call>
~~~

No `<tool_call>` opening tag (consistent with A1/A2/B in @damsnotdam's stage-2/stage-3 breakdown above), stray orphan `</tool_call>` present here (their "3 of 4" case, not their "1 of 4"). `finish_reason=stop`, 0 tool_calls, content returned verbatim — matches case B exactly.

**Rate, small sample:** 3 leaks in 5 turns across one session (informal, not a controlled arm — but in the same ballpark as the 20–40%/call range reported above).

**New failure mode worth flagging: it gets worse than a leak.** On one retry, instead of just leaking the markup, the model followed the leak by fabricating a plausible-looking tool result and presenting it as real — a fake `ls -la` listing (wrong permission bits: `drwxr-xr-x` on entries that were clearly files, e.g. `002-sails-orientation.md`), and it kept extending the fake listing indefinitely until the client was interrupted. `finish_reason` gave no indication anything was wrong. For an agentic client trusting tool results, a silent hallucinated result is worse than a silent empty one — no error, no retry trigger, just wrong information presented with full confidence.

**Status check:** this PR has been open since April with two corroborating reports and a clean cherry-pick onto later tags noted in this thread, and the underlying bug is still live on the latest release four months later. Is there something blocking review, or is this waiting on something specific? Happy to provide more repro data (trace-level oMLX logs, exact request bodies) if it would help move it along.

jundot (approved):
Thanks for adding this recovery path. I verified a maintainer follow-up that preserves schema types and embedded closing tags, and suppresses the recovered markup across OpenAI, Anthropic, and Responses streams. I'll land that follow-up immediately after the squash merge.

PR #845, opened 2026-04-18
fix(server): Anthropic stream emission corner cases around tool_use

mrtkrcm:
## Summary

Fix Anthropic `/v1/messages` streaming when a response is mostly or only `tool_use` content.

## Changes

- Drop leading whitespace deltas around tool envelopes before any text block starts.
- Parse tool calls before deciding whether to emit an empty text block.
- Start tool-only streams at content block index `0` instead of emitting an empty text block first.
- Preserve cleaned empty text in non-streaming Anthropic responses.
- Add regression coverage for tool-only Anthropic streams.

## Local validation

Built and installed from this branch into `/Applications/oMLX.app` (`0.3.8.dev2`), with port `8801` owned by the visible app process.

- Focused tests: `201 passed, 12 deselected` for `tests/test_tool_calling.py`, `tests/integration/test_e2e_streaming.py`, and `tests/test_admin_profiles_api.py`.
- Live Anthropic proxy path: `claude-sonnet-4-6` forced `get_weather`; streaming emitted first content block as `tool_use` at index `0`, no leading empty text block, `stop_reason=tool_use`.
- Direct OpenAI control: `Ternary-Bonsai-8B-mlx-2bit` returned `ready`; benchmark tool `OK` in `0.6s`.
- Related Qwen tool path: `Qwen3-Coder-30B-A3B-Instruct-4bit` forced `read_file`; structured `tool_calls`, benchmark tool `OK` in `2.3s`.

Note: benchmark host was not clean; preflight saw active desktop/client load, so throughput is smoke data only.

## Test Plan

- `uv run pytest tests/integration/test_e2e_streaming.py -q`
- `python3 -m py_compile omlx/server.py tests/integration/test_e2e_streaming.py`

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -3051,17 +3051,33 @@ async def stream_anthropic_messages(
                     if tool_filter:
                         content_delta = tool_filter.feed(content_delta)
                     if content_delta:
-                        # Close thinking block if transitioning to text
-                        if thinking_block_started and not text_block_started:
-                            yield create_content_block_stop_event(index=block_index)
-                            block_index += 1
-                            thinking_block_started = False
-                        if not text_block_started:
-                            yield create_content_block_start_event(
-                                index=block_index, block_type="text"
-                            )
-                            text_block_started = True
-                        yield create_text_delta_event(index=block_index, text=content_delta)
+                        # When tools are requested AND we haven't yet opened
+                        # a text block, drop pure-whitespace deltas. Models
+                        # often emit a leading newline around <tool_call>
+                        # envelopes that tool_filter passes through
+                        # (whitespace isn't part of the envelope markers).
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, merging. The three streaming cases are all real and the index-0 test is a good pin.

Heads-up on the non-streaming hunk in `_build_anthropic_message`: `extraction.cleaned_text` is typed `str` and always assigned in both branches (Harmony sets `cleaned_text = regular_content`, the parse path reads it from the dataclass), so `cleaned_text is None` never triggers. Behaviorally a no-op for live traffic, but the `else regular_content` fallback would re-introduce unsanitized text if that branch ever became reachable. I'll revert just that one line in a follow-up.

PR #846, opened 2026-04-18
fix(specprefill): add qwen3_5_moe / qwen3_6 query extractor

mrtkrcm:
## Summary

The existing `_qwen35_extract_queries` path assumed Qwen3-Next's gate-split
`q_proj` (first half = queries, second half = gate). The newer Qwen3.5MoE /
Qwen3.6 Attention class lacks that gate split —
`q_proj.weight.shape[0] == n_heads * head_dim` (no doubling for gate).

Running SpecPrefill against `mlx-community/Qwen3.6-35B-A3B-4bit-DWQ` with the
3.5 extractor silently produced malformed queries and disabled the speedup
path — the feature stayed enabled but degraded to no-op + overhead.

## Fix

- Rewrite `_qwen35_extract_queries` with an attribute fallback for the
  Qwen3-Next variant that still has the gate.
- Add `_qwen36_extract_queries`: `queries = q_out.reshape(B, L, n_heads, -1)`
  with no gate split — the correct shape for Qwen3.5MoE and Qwen3.6.
- Rewrite `_detect_query_extractor` to compare `q_proj.weight.shape[0]`
  against `n_heads * q_norm.weight.shape[0]` and pick the right path per
  layer.

## Measured effect

`mlx-community/Qwen3.6-35B-A3B-4bit-DWQ` on kern 9-test routing suite,
SpecPrefill + TurboQuant KV both enabled:

- Before (features on, wrong extractor): **11.87 tok/s** (same as features off)
- After (features on, correct extractor): **25.9 tok/s** → **2.18× speedup**

Score unchanged at 7-8/9; this is purely a performance fix.

## Config used

```json
\"specprefill_enabled\": true,
\"specprefill_draft_model\": \".../Qwen3-0.6B-4bit\",
\"specprefill_keep_pct\": 0.2,
\"specprefill_threshold\": 8192,
\"turboquant_kv_enabled\": true,
\"turboquant_kv_bits\": 4
```

## Scope

All changes in `omlx/patches/specprefill.py`. No change to Qwen3-Next path
behaviour.

## Test plan

- [x] `python -m py_compile omlx/patches/specprefill.py`
- [x] Cold + warm tok/s measurements on Qwen3.6-35B-A3B-4bit-DWQ
- [x] Qwen3-Next models still route to the gate-split path
- [x] Score stability unchanged (7-8/9 across 3 runs)

### DIFF
--- a/omlx/patches/specprefill.py
+++ b/omlx/patches/specprefill.py
@@ -65,12 +65,45 @@ class _AttentionCapture:
 
 
 def _qwen35_extract_queries(attn, x, cache=None):
-    """Qwen3.5: gate split + q_norm + RoPE."""
+    """Qwen3.5 / Qwen3-Next: gate split + q_norm + RoPE.
+
+    Applies to models whose ``q_proj`` concatenates (queries, gate) and
+    outputs ``2 * n_heads * head_dim``. Dispatcher ``_detect_query_extractor``
+    picks this when q_proj output dim == 2 * n_heads * q_norm_dim.
+    """
     B, L, D = x.shape
+    n_heads = getattr(
+        attn,
+        "num_attention_heads",
+        getattr(attn, "n_heads", getattr(attn, "num_heads", None)),
+    )
     q_out = attn.q_proj(x)
     queries, _gate = mx.split(
-        q_out.reshape(B, L, attn.num_attention_heads, -1), 2, axis=-1
+        q_out.reshape(B, L, n_heads, -1), 2, axis=-1
     )
+    queries = attn.q_norm(queries).transpose(0, 2, 1, 3)
+    if cache is not None:
+        queries = attn.rope(queries, offset=cache.offset)
+    else:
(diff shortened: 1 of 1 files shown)


DiegoRBaquero (approved):

jundot:
Hey @mrtkrcm, sorry for the auto-close. Your diagnosis was exactly right.

The branch had drifted off `ae908e6` and was in CONFLICT after #851 rewrote the dispatcher for Gemma 4. A naive rebase would have routed Gemma 4 (also per-head q_norm) into the qwen3.6 path and broken its shared_kv / offset handling. So I rebuilt your idea on current main as `a1adf0b`: Gemma 4 stays ahead of the new check, plus a q_norm shape guard so olmo / cohere flat-norm models stay on the llama path.

You're credited as co-author on the commit. Thanks for the clean measurement on Qwen3.6-35B-A3B-4bit-DWQ, that made it obvious this was worth pulling.

PR #851, opened 2026-04-18
fix(specprefill): complete Gemma 4 compatibility

cresslank:
### Summary:

Fixes #668.

This PR cherry picks @apetersson's initial fix from `apetersson/feature/gemma4-specprefill` and then adds the remaining Gemma 4-specific fixes needed for SpecPrefill:
- find `previous_kvs` on both VLM and text-only Gemma 4 model wrappers
- capture Gemma 4 queries with the same per-head `q_norm` and forwarded `offset` used by the real attention path
- narrow the Qwen extractor check to gated `q_proj` models instead of using broad `q_norm` heuristics
- normalize wrapped `mx.array` RoPE offsets through `_scalar_offset`

### Validation:
- ran `uv run --extra dev pytest tests/test_specprefill.py -q` (31 passed)
- ran benchmark with Gemma 4 26B MoE + Gemma 4 E2B draft on a 16384-token prompt


jundot:
Thanks for picking this up. I cross-checked each fix against `mlx-lm/mlx_lm/models/gemma4_text.py`:

- `**kwargs` forward matches the `(shared_kv, offset)` signature and the 3-tuple return (gemma4_text.py:226-267, 336)
- `_uses_gated_q_proj` matches Qwen 3.5 exactly (qwen3_next.py:89-93) and correctly excludes Gemma 4
- `_gemma4_extract_queries` replicates `q_proj → reshape → q_norm → transpose → rope` in order
- `previous_kvs` lookup matches the shared-KV layout and covers VLM + text-only
- `_scalar_offset` is needed since Gemma 4 wraps `cache.offset` as `mx.array` (gemma4_text.py:247)

No regressions for Qwen 3.5 / Nemotron-H / Llama paths. Plain Qwen3 used to crash on the split and now falls through to `_llama_extract_queries`, strictly better.

Merging.

jontyab:
> * ran benchmark with Gemma 4 26B MoE + Gemma 4 E2B draft on a 16384-token prompt

Ran this same config at oQ4 (as of v0.3.8) and I indeed see a two-fold improvement in TTFT and PP, but TG halves. 

> M2 Pro 32GB
> pp16384/tg128, specprefill off:  TTFT 42.9s, tg 31.9 tok/s
> pp16384/tg128, specprefill on:   TTFT 20.6s, tg 17.4 tok/s

Do you expect to see a similar decode regression?

cresslank:
@jontyab yeah it had been on my list of follow-ups to see if there were any current omlx issues re this.

i had started some initial poking on the tg degradation, but it quickly sprawled -- so kept this scoped to just the hard blocker on spec prefill running at all on gemma 4.

if there isn't a relevant issue already feel free to open one and tag me and i can try to take a look. otherwise i can open an issue when i have time to investigate etc.

PR #853, opened 2026-04-19
feat(admin): model settings profiles - save and reuse parameter bundles

sxc562586657:
### Motivation

Setting up a model in oMLX requires configuring ~20+ parameters (temperature, top\_p, max\_tokens, thinking settings, quantization flags, speculative decoding options…). When the number of fields keeps growing, re-entering them by hand for every new model or use case becomes tedious and error-prone.

This feature lets users save a **named snapshot** of any parameter bundle — a **profile** — and apply it to any model in one click. It also adds **templates**: a global, model-agnostic bundle of universal fields (temperature, top\_p, etc.) that can be applied across all models at once.

---

#### Data layer (`omlx/model_settings.py`, `omlx/model_profiles.py`)

- `ModelProfile` dataclass: per-model saved bundle of settings with metadata (name, display_name, description, timestamps)
- `GlobalTemplate` dataclass: reusable bundle of universal fields only
- `ModelSettingsManager`: thread-safe manager with atomic JSON writes (temp-file + rename), snapshot-and-rollback on save failure, cascade delete when a model is removed, and field allowlisting for universal vs model-specific classification

#### Admin API (`omlx/admin/routes.py`)

- Full CRUD for profiles: `POST/GET/DELETE /api/models/{id}/profiles`, `PUT /api/models/{id}/profiles/{name}`
- `POST /api/models/{id}/profiles/{name}/apply` — applies a profile to live model settings, persists `active_profile_name`, returns merged `ModelSettings`
- `POST /api/models/{id}/profiles/{name}/apply` with `also_save_as_template=True` creates a linked template alongside the profile
- Full CRUD for templates: `GET/POST/PUT/DELETE /api/profile-templates`
- `GET /api/profile-fields` — serves field allowlists as JSON so the frontend derives its serializer from the same source as the backend
- `active_profile_name` surfaced in `GET /api/models` response; drift detected and cleared automatically on manual save

#### Frontend (`omlx/admin/static/js/dashboard.js`, `omlx/admin/templates/dashboard/_modal_model_settings.html`)

- Profile and template pills in the model settings modal, with inline create/edit forms and delete confirmation
- `applyProfileToForm` / `applyTemplateToForm` with a monotonic sequence counter to prevent race conditions when clicking pills quickly
- `computeDrift()` — detects when manual edits diverge from the active profile, shown as a warning badge
- `formValuesForProfile` driven by the fetched allowlists — adding a field to `model_profiles.py` automatically covers it in the form serializer, no JS change needed
- Profile badge on the model list: shows the active profile name in emerald, or "Custom" when no profile is active

#### Guard tests

- `test_all_model_settings_fields_classified`: fails if any `ModelSettings` field is not classified in `UNIVERSAL_PROFILE_FIELDS`, `MODEL_SPECIFIC_PROFILE_FIELDS`, or `EXCLUDED_FROM_PROFILES` — catches drift between the dataclass and the allowlists at CI time
- API route tests cover all 200/400/404/409 paths for profiles and templates
- `test_apply_profile_sets_active` verifies `active_profile_name` is persisted end-to-end

---

#### Universal vs model-specific split

Fields like `temperature`, `top_p`, `max_tokens` are universal — they make sense as defaults for any model. Fields like `turboquant_kv_bits`, `specprefill_draft_model` are model-specific and only apply to certain backends. Templates are restricted to universal fields; profiles can carry both. The split is enforced in `model_profiles.py` via two tuples and at the Python level by `filter_profile_fields()`.

#### Single source of truth for field allowlists

The frontend's form serializer (`formValuesForProfile`) previously duplicated the allowlist. Now `GET /api/profile-fields` serves it at runtime, and a guard test catches any drift between `ModelSettings` and `model_profiles.py` at dev time. Adding a new field requires updating exactly one file (`model_profiles.py`).

#### No triplication of mutable defaults

Both `CreateProfileRequest` and `CreateTemplateRequest` use `Field(default_factory=dict)` instead of `{}` as the mutable default, avoiding the Pydantic v1 footgun.

#### Atomic writes with rollback
All mutating operations (`apply_profile`, `update_profile`, `delete_profile`) snapshot in-memory state before modifying it, write to disk atomically (temp file + rename), then restore from the snapshot if the write throws. This prevents disk/memory divergence that would silently lose a user's rename or delete on restart.

#### Drift clearing

When a user manually changes a field that differs from the active profile's stored value, `active_profile_name` is cleared and the profile is marked inactive on the next save. The user is never surprised by a stale badge.

---
Screenshots

Screenshot 2026-04-18 at 5 38 27 PM

Screenshot 2026-04-18 at 5 39 22 PM


jundot:
Nice work, the design is solid. Allowlist single-source, drift detection, the guard test that catches unclassified fields at CI time, and snapshot-and-rollback on atomic writes, all the right calls. Merging.

zviratko:
Please tell us that this is going towards this https://github.com/jundot/omlx/issues/341 ❤️

PR #857, opened 2026-04-19
perf(api): pass reasoning_content as message field for native templates

jundot:
## Summary

Stacked on top of #856.

Qwen 3.6+ chat templates support `message.reasoning_content` as a first-class message field. The `<think>...</think>` reconstruction added in #814 is actually the fallback path that the template falls into when the field is absent:

```jinja
{%- if message.reasoning_content is string %}
    {%- set reasoning_content = message.reasoning_content %}
{%- else %}
    {%- if '</think>' in content %}
        {%- set reasoning_content = content.split('</think>')[0]... %}
        {%- set content = content.split('</think>')[-1].lstrip('\n') %}
    {%- endif %}
{%- endif %}
```

Switching to the field-based path on supported templates avoids the whitespace round-trip between `extract_thinking.strip()` (output side) and the reconstructor's `\n` delimiters (input side), which should improve the KV prefix cache reuse #814 cited as motivation.

## Changes

- Adds `native_reasoning_content` parameter to `extract_text_content`, `extract_multimodal_content`, and `convert_anthropic_to_internal`
- `server.py` reads `preserve_thinking_default` off the engine entry and threads it into each extractor call site (chat completions text + multimodal, Anthropic messages)
- Teaches `_drop_void_assistant_messages` to preserve assistant messages that carry `reasoning_content` but no text content (otherwise native-mode turns get dropped)
- When `native_reasoning_content=False` (default), behaviour is identical to #814 — the 8 existing reasoning/thinking tests pass unchanged as a regression gate

## Test plan

- [x] `pytest tests/test_api_utils.py -v` → 161 passed (including the 8 #814 tests + 9 new native-mode tests)
- [x] `pytest tests/test_thinking_toggle.py -v` → 23 passed
- [x] `pytest tests/ -m "not slow" -k "anthropic or thinking"` → 172 passed
- [ ] Smoke test with Qwen3.6 — confirm `reasoning_content` field survives through to the template render
- [ ] Confirm non-Qwen3.6 thinking models still use the `<think>` fallback path (no regression)

### DIFF
--- a/omlx/api/anthropic_utils.py
+++ b/omlx/api/anthropic_utils.py
@@ -126,6 +126,7 @@ def convert_anthropic_to_internal(
     max_tool_result_tokens: int | None = None,
     tokenizer: Any | None = None,
     preserve_images: bool = False,
+    native_reasoning_content: bool = False,
 ) -> list[dict[str, Any]]:
     """
     Convert Anthropic Messages API format to internal format.
@@ -142,6 +143,10 @@ def convert_anthropic_to_internal(
         tokenizer: Tokenizer instance for token counting and truncation.
         preserve_images: If True, preserve image blocks as OpenAI image_url
             format for VLM processing.
+        native_reasoning_content: If True, attach Anthropic ``thinking`` blocks
+            as a ``reasoning_content`` field on assistant messages (Qwen 3.6+
+            templates).  If False, inline each block as ``<think>...</think>``
+            in the message content as a fallback.
 
     Returns:
         List of {"role": str, "content": str or list}
@@ -169,6 +174,7 @@ def convert_anthropic_to_internal(
                     text_parts: list[str] = []
                     image_parts: list[dict] = []
                     tool_calls: list[dict] = []
+                    thinking_parts: list[str] = []
                     for block in content:
                         block_dict = _content_block_to_dict(block)
                         if block_dict is None:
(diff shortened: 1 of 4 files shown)


latent-variable:
### Follow-up: I verified the unchecked smoke test item,  VLM path drops `reasoning_content`

Ran the two-turn recall methodology from #814 against `Qwen3.6-35B-A3B-8bit`: turn 1 asks for N random 20-digit numbers, return only one; turn 2 (and beyond) asks for the next one. Without reconstruction, turn 2 fabricates/denies; with it, the model recalls exactly. Scales cleanly to N — three-number variant also recalls #2 and #3 verbatim.
Screenshot 2026-04-20 at 12 05 24
- **Before fix** (screenshot 1): turn 2 — *"I didn't actually generate a second number — I just made one up on the spot."*
Screenshot 2026-04-20 at 12 08 10

Screenshot 2026-04-20 at 12 09 43

- **After fix** (screenshots 2–3): turn 2 and turn 3 recall the hidden numbers exactly.

### How to validate that preserve thinking is on:
Simple test: ask the model:
`can you come up with two random 20 digit number and validate that they are 20 digits, do not use any tools, and only give me one of the two and nothing else`

Ensure the model actually thinks of two numbers otherwise retry, next turn ask:
`now give me the second number that you came up with`

### Root cause

`omlx/engine/vlm.py:598-612` — `_format_messages_for_vlm_template` preserves `tool_calls` / `tool_responses` verbatim but routes other assistant messages through `mlx_vlm.prompt_utils.get_message_json(model_type, content, role, ...)`, which only accepts `content`. The top-level `reasoning_content` field this PR adds is dropped before `apply_chat_template` runs.

### Suggested patch

Extend the existing preserve-verbatim branch — same class of field as `tool_calls` / `tool_responses`, same reason already documented in the comment block:


```
if role == "tool" or (
    role == "assistant"
    and (
        msg.get("tool_calls")
        or msg.get("tool_responses")
        or msg.get("reasoning_content")
    )
):
    formatted_messages.append(msg)
    continue
```


### Justification

- Same failure mode as the tool-field stripping already handled here: \`get_message_json\` doesn't know about these top-level fields, so they must be preserved before the helper runs.
- Non-VLM path uses \`extract_text_content\` → \`apply_chat_template\` directly, so it's unaffected.
- Avoids falling back to \`<think>{reasoning}</think>\n\n{content}\` on the VLM path, which would re-introduce the whitespace round-trip #857 set out to eliminate.

Happy to open a small PR with this change if useful.

jundot:
Thanks for the diagnosis — absolutely right. Confirmed every Qwen 3.6 variant is a VLM model (`vision_config` + `image_token_id` + `Qwen3_5MoeForConditionalGeneration`), so `_format_messages_for_vlm_template` stripping `reasoning_content` was a hard blocker, not an edge case. The whitelist approach you suggested is the right fix — `reasoning_content` belongs in the same class as `tool_calls` and `tool_responses` (top-level field the template reads directly, but `get_message_json` doesn't know about).

This PR auto-closed because its base branch got deleted when #856 merged. Reopened as #884 with your fix squashed in as a second commit, off `main`. Added 4 VLM unit tests covering the preserve-verbatim path:
- `test_reasoning_content_preserved_verbatim`
- `test_reasoning_content_coexists_with_tool_calls` (OR-connected whitelist guard)
- `test_no_reasoning_content_uses_get_message_json` (regression guard for the default path)
- `test_user_reasoning_content_is_ignored` (user role stays on the get_message_json path)

Appreciate the detailed root cause + suggested patch. Moving the conversation to #884.

PR #860, opened 2026-04-19
fix: keep tool_call arguments as strings in non-Harmony extractors

### LINKED ISSUE #854: Qwen3.6-35B-A3B-8bit + Hermes ，bug
Bug: "Can only get item pairs from a mapping" after successful chat completion (DMG version, tool calling enabled)

Body:

text
Dear OMLX team,

I encountered a recurring error when using the DMG version of OMLX with the VLM model `mlx-community:Qwen3.6-35B-A3B-8bit`. After a successful chat completion, all subsequent requests to `/v1/chat/completions` fail with HTTP 500 and the error message:

> Can only get item pairs from a mapping.

**Environment:**
- OMLX distribution: DMG version (not pip or Homebrew)
- Model: mlx-community:Qwen3.6-35B-A3B-8bit (VLM)
- Tool calling: enabled (parser=qwen3_coder)
- Structured output: not available (as noted in logs: "Structured output is not available in the DMG version (xgrammar requires torch…)")
- OS: macOS

**Log excerpt (timestamps from 2026-04-19):**
08:11:09,439 - omlx.scheduler - INFO - Using boundary cache snapshot for 4dd6d3c1...: storing 8192/12459 tokens
08:11:09,548 - omlx.server - INFO - Chat completion: 4248 tokens in 103.46s (41.1 tok/s)
08:11:09,701 - omlx.server - ERROR - POST /v1/chat/completions → 500 (unhandled): Can only get item pairs from a mapping.
08:11:10,120 - omlx.server - ERROR - POST /v1/chat/completions → 500 (unhandled): Can only get item pairs from a mapping.
... (repeated for subsequent requests)

text

**Observations:**
- The error appears immediately after a successful, high-speed generation (4248 tokens, 41.1 tok/s).
- Once the error occurs, all following requests fail with the same 500 error, even with different inputs.
- Restarting the OMLX server temporarily resolves the issue.
- The problem does **not** occur when tool calling is disabled (i.e., no `tools` field in the request).
- The error suggests that somewhere in the response/tool-call parsing code, `.items()` is called on an object that is not a `dict` (likely `None`, list, or string). This may be triggered by malformed tool-call arguments generated by the model.

**Steps to reproduce (intermittent):**
1. Start OMLX server (DMG version) with the model `mlx-community:Qwen3.6-35B-A3B-8bit`.
2. Send a chat completion request that includes `tools` (tool calling enabled). The request should be long enough to trigger boundary cache snapshots (as seen in logs).
3. After a successful response, send the same or any other request again.
4. Observe that the second and all subsequent requests fail with the above 500 error.

**Workaround:** Disable tool calling (omit `tools` from requests) or switch to the pip/Homebrew version.

**Request:**
- Please investigate the tool-call argument parsing logic, especially the part that handles the model's output when `structured output` is unavailable (DMG fallback path).
- It would also help to have a more descriptive error message (e.g., logging the actual type of the object that failed `.items()`).

Let me know if you need additional logs, a full traceback, or a sample request payload that triggers the issue.

Thank you.

nikolay-bratanov:
## Summary

- `extract_text_content()` and `extract_multimodal_content()` called `_try_parse_json()` on tool_call `arguments`, converting JSON strings into Python dicts
- Qwen3.x chat templates (and likely others) expect `arguments` as a **string** per the OpenAI API spec — receiving a dict causes `"Can only get item pairs from a mapping"` and HTTP 500
- `_try_parse_json()` was needed only for Harmony (gpt-oss) models whose templates apply `|tojson`. Harmony already uses its own `extract_harmony_messages()` which correctly calls `_try_parse_json()`
- This fix removes `_try_parse_json()` calls from the two general-purpose extractors, keeping arguments as strings

## Changes

- `omlx/api/utils.py`: Replace 4 × `_try_parse_json(...)` with direct string passthrough in `extract_text_content()` and `extract_multimodal_content()`
- No changes to `extract_harmony_messages()` — it retains `_try_parse_json()` as intended

## Test plan

- [x] Multi-turn chat with tool calls on Qwen3.5/3.6 model — second request no longer returns 500
- [ ] Verify Harmony models still work with parsed arguments (no Harmony model available to test)

Fixes #854

### DIFF
--- a/omlx/api/utils.py
+++ b/omlx/api/utils.py
@@ -387,9 +387,7 @@ def extract_text_content(
                                 "id": tc.get("id", ""),
                                 "function": {
                                     "name": func.get("name", ""),
-                                    "arguments": _try_parse_json(
-                                        func.get("arguments", "{}")
-                                    ),
+                                    "arguments": func.get("arguments", "{}"),
                                 },
                             }
                         )
@@ -408,7 +406,7 @@ def extract_text_content(
                                         if hasattr(tc, "function")
                                         else ""
                                     ),
-                                    "arguments": _try_parse_json(args_str),
+                                    "arguments": args_str,
                                 },
                             }
                         )
@@ -547,9 +545,7 @@ def extract_multimodal_content(
                                 "id": tc.get("id", ""),
                                 "function": {
                                     "name": func.get("name", ""),
-                                    "arguments": _try_parse_json(
-                                        func.get("arguments", "{}")
-                                    ),
+                                    "arguments": func.get("arguments", "{}"),
                                 },
                             }
                         )
@@ -568,7 +564,7 @@ def extract_multimodal_content(
                                         if hasattr(tc, "function")
                                         else ""
                                     ),
-                                    "arguments": _try_parse_json(args_str),
+                                    "arguments": args_str,
                                 },
                             }
                         )


jundot:
Hey, thanks for the PR. I looked into this and I think the direction is actually the opposite of what we want, which is why the fix doesn't hold up beyond the model you tested.

I ran the real chat templates through jinja2 with `arguments` as string vs dict:

| model | string | dict |
|---|---|---|
| Qwen3.6-35B-A3B (the one from #854) | fail: `Can only get item pairs from a mapping` | ok |
| Qwen3.5 | fail: same | ok |
| GLM-4.6V-Flash | fail: `'str object' has no attribute 'items'` | ok |
| MiniMax-M2.7 | fail: same | ok |
| Qwen3-Instruct-2507 | ok | ok |

The Qwen3.5/3.6 template literally does `{%- for k, v in tool_call.arguments|items %}`, and GLM/MiniMax do `_args.items()`. They all want a dict. The error in #854 is exactly what you get when you hand them a string, so removing `_try_parse_json()` triggers the same bug on every multi-turn tool call instead of fixing it. `_try_parse_json` was added in ca5f060 for GLM/MiniMax specifically, it's not harmony-only.

The only template where string works is Qwen3-Instruct-2507, which has an `is string` branch. Guessing that's the variant you tested against?

Actual cause of #854 is that some clients echo back a non-JSON `arguments` value: python repr with single quotes, plain text, truncated JSON, empty string, and so on. `_try_parse_json` passes those through unchanged and the template blows up on `.items()`. I reproduced all four cases locally.

So instead of this PR I'm landing a pydantic validator that rejects non-JSON-object `arguments` at request ingress with a clear 422 pointing at `messages[i].tool_calls`, plus an output-side guard in `tool_calling.py` so the server itself never emits a non-JSON arguments string. Empty string and dict-instead-of-string get normalized rather than rejected since both show up in the wild.

Closing this one, but thanks for the report. Tool calling breaking on weird echoes was a real thing, just in the other direction.

PR #878, opened 2026-04-21
fix(omlx_app): health checks and local URLs honor bind host

DKev:
Problem: With a custom bind address (only listening on a non-loopback IP), the app always hit http://127.0.0.1:{port}, so health never succeeded and startup looked hung; port/TCP checks could also misbehave.
Change: Read server.host from {base_path}/settings.json and use it to build local health/base URLs; treat "" / 0.0.0.0 / :: as loopback for local probes; bracket IPv6 in URLs; align admin dashboard/stats/API key update URLs and TCP probe targets.
Tests: Added coverage for URL resolution and custom-bind health/API URLs in tests/test_omlx_app.py.


DKev:
Fixes #963

jundot:
Thanks — clean fix, the helper split makes the bind-host resolution easy to follow.

Merged with the trivial test-file import conflict resolved on my side. Also pushed a follow-up (4e12c82) to apply the same helper to `update_model_dir_runtime`, which had the same 127.0.0.1 issue.

Issue #877, opened 2026-04-21
Qwen3-VL-Reranker-2B is rejected on oMLX 0.3.6 with "Unsupported reranker architecture: Qwen3VLForConditionalGeneration"


j0uz:
## Description

I’m trying to use `mlx-community/Qwen3-VL-Reranker-2B-4bit` as a reranker in oMLX `0.3.6`, but loading the model fails with this error:

`Unsupported reranker architecture: Qwen3VLForConditionalGeneration. Currently supported architectures: JinaForRanking, ModernBertForSequenceClassification, Qwen3ForCausalLM, XLMRobertaForSequenceClassification.`

I want to confirm whether:

1. `Qwen3-VL-Reranker-2B` is currently unsupported in oMLX 0.3.6
2. support for this model is planned in a future release
3. there is any workaround or recommended alternative for multimodal reranking

## To Reproduce

1. Install / run oMLX version `0.3.6`
2. Download the model:
   - `mlx-community/Qwen3-VL-Reranker-2B-4bit` 
3. Try to load it as a reranker in oMLX
4. Observe the architecture error popup

## Expected behavior

I expected either:

- the model to be accepted as a reranker, or
- a clear note in documentation / release notes that `Qwen3-VL-Reranker-2B` is not yet supported

## Actual behavior

oMLX rejects the model with:

`Unsupported reranker architecture: Qwen3VLForConditionalGeneration. Currently supported architectures: JinaForRanking, ModernBertForSequenceClassification, Qwen3ForCausalLM, XLMRobertaForSequenceClassification.`

## What I already verified

- `mlx-community/Qwen3-VL-Reranker-2B-4bit` exists on Hugging Face as an MLX model
- `Qwen3-VL-Embedding-2B` works for embeddings in my setup
- the failure only happens when loading the VL reranker
- I am using oMLX `0.3.6`

## Environment

- oMLX version: `0.3.6`
- OS: macOS
- Hardware: Apple Silicon (M2, 16GB)
- Deployment method: local oMLX server

## Screenshots

I attached the popup screenshot showing the unsupported architecture error.

<img width="916" height="482" alt="Image" src="https://github.com/user-attachments/assets/bb0fc64b-a0ca-4dee-96fa-8b43969c69e5" />

<img width="1884" height="124" alt="Image" src="https://github.com/user-attachments/assets/63572bc4-53f3-4bb7-9efc-1b1257a0b647" />

## Question

[...]

PR #887, opened 2026-04-22
fix(admin): reset composer height and contain user markdown overflow

chulanpro5:
## Summary

Fix two `/admin/chat` UI issues:

- reset the composer height after sending a long pasted message
- prevent fenced code / long markdown lines from overflowing user message bubbles

## Testing

Manually verified both behaviors in `/admin/chat`.

## Notes

Only `omlx/admin/templates/chat.html` is changed.

### DIFF
--- a/omlx/admin/templates/chat.html
+++ b/omlx/admin/templates/chat.html
@@ -126,6 +126,10 @@
         padding: 8px 16px;
         font-size: 14px;
         color: var(--text-primary);
+        max-width: 100%;
+        min-width: 0;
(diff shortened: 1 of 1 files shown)


jundot:
Rebased onto latest main to drop three commits that were already squash-merged (#877 reranker, preset system, stats alias test), so the diff now shows just the intended chat.html change. Both fixes reproduce cleanly and the scope is tight. Thanks for the fix!

chulanpro5:
Thanks !

PR #890, opened 2026-04-22
feat(admin): fix model card dark mode styles using CSS variables

miaobuao:
Fix model card markdown styles to support dark mode by replacing hardcoded colors with CSS variables.


jundot:
Thanks for the fix. Variables map cleanly to the existing theme tokens and `--link-color` is defined in both light and dark. Merging as-is.

PR #894, opened 2026-04-22
update mlx-lm dependency to tag v0.31.3 

hardlycharred:
fixes 'There is no Stream(gpu, 0)' being returned from LLM responses

### DIFF
--- a/packaging/venvstacks.toml
+++ b/packaging/venvstacks.toml
@@ -15,8 +15,8 @@ runtime = "cpython-3.11"
 requirements = [
     # Core MLX
     "mlx==0.31.1",
-    # mlx-lm from commit (a401730) - aligned with pyproject.toml
-    "mlx-lm @ git+https://github.com/ml-explore/mlx-lm@a40173094136fbcbfb41966889a22530c83bd86e",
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the PR! I already landed the same intent on main via a two-step fix:

- `445d019` bumps mlx-lm to v0.31.3 (pinned by full commit `ed1fca4` for reproducibility) and also bumps `mlx` to 0.31.2 (v0.31.3 raises its min to 0.31.2, so the pyproject `mlx>=0.31.1` line needed bumping too; `venvstacks.toml` + `exclude-newer` as well).
- `#891` (merged as `66a7ad8`) is the actual runtime fix for the `Stream(gpu, 0)` crash. I verified on v0.31.3 that the mlx-lm bump alone does not resolve it: `new_thread_local_stream` binds to the thread that called it (the main thread at import time), but omlx runs all MLX ops on a dedicated `mlx-global` worker, so cross-thread resolve still fails. `#891` fixes that by creating a stream inside the executor thread and rebinding `generation_stream` in both module namespaces.

Going to close this as superseded — thanks again for flagging the issue and proposing a fix.

PR #908, opened 2026-04-23
fix: handle list-type host in settings and fix [object Object] error display

fqx:
Superseded by a new PR that adapts to the `uvicorn.Config + bind_socket()` refactor introduced in #1526. The core logic is the same but the `cli.py` changes are rewritten to work with the new pre-bind approach.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -2279,6 +2279,17 @@ async def update_global_settings(
 
     # Apply server settings
     if request.host is not None:
+        from ..utils.network import is_valid_bind_host
+
+        parts = [h.strip() for h in request.host.split(",") if h.strip()]
+        if not parts:
+            raise HTTPException(status_code=400, detail="Host cannot be empty")
+        for part in parts:
+            if not is_valid_bind_host(part):
+                raise HTTPException(
+                    status_code=400,
+                    detail=f"Invalid host: {part!r} (must be a hostname or IP address)",
+                )
         global_settings.server.host = request.host
     if request.port is not None:
         global_settings.server.port = request.port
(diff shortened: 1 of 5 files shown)


fqx:
## Summary

This PR enables users to specify multiple bind addresses directly from the web UI, rather than being limited to a single address or the catch-all `0.0.0.0`.

**Why it matters:** Binding only to specific interfaces is more secure than listening on all interfaces. A common use case is exposing the server locally and over a private overlay network such as Tailscale, without opening it to every interface on the machine — for example:

```
127.0.0.1, 100.x.x.x
```

While this was already supported via manual edits to `settings.yaml`, the web UI could not round-trip a YAML list value — it would display `[object Object]` in the host field and produce an unreadable `[object Object]` error message when saving failed.

**What changed:**

- **`settings.py`** — `ServerSettings.from_dict()` now normalizes a YAML list into a comma-separated string, so the web UI always receives a plain string regardless of how the config file was written.
- **`cli.py`** — the comma-separated host string is split before being passed to `uvicorn.run()`, which accepts a list of addresses and binds a socket for each.
- **`network.py`** — adds `is_valid_bind_host()` to correctly accept unspecified addresses like `0.0.0.0` and `::` as valid bind targets (which the existing `is_valid_alias()` rejects, since they are not routable as client-facing URLs).
- **`admin/routes.py`** — validates each comma-separated entry on save and returns a readable 400 error before an invalid value can crash the server on restart.
- **`admin/static/js/dashboard.js`** — extracts `.msg` from Pydantic error detail objects instead of joining the raw objects, fixing the `[object Object]` error display at two call sites.

Existing single-address configs and legacy YAML list values continue to work without any migration.

PR #922, opened 2026-04-23
fix(server): report cached tokens in OpenAI-compatible usage details

### LINKED ISSUE #919: cached_tokens missing from non-streaming chat/completion responses
**Describe the bug**

Non-streaming `/v1/chat/completions` and `/v1/completions` responses always return `cached_tokens: null` in the `usage` object, even when a prefix cache hit occurs. The streaming path (with `stream_options.include_usage`) correctly reports `cached_tokens`. The value is computed by the scheduler and carried all the way to `GenerationOutput.cached_tokens`, but the non-streaming response builders omit it when constructing the `Usage` object.

**To Reproduce**

1. Send a chat completion request with a prompt large enough to fill at least one paged cache block (>2048 tokens by default).
2. Send the identical request a second time to trigger a cache hit.
3. Observe that `usage.cached_tokens` is `null` in the non-streaming response, despite the cache hit (confirmed by a large TTFT drop, ~11s → ~1.5s on a 2215-token prompt).
4. Repeat with `"stream": true, "stream_options": {"include_usage": true}` and observe `cached_tokens: 2048` on the same prompt.

**Expected behavior**

Non-streaming responses should populate `usage.cached_tokens` on cache hits, matching the streaming path.

Root cause appears to be omission of `cached_tokens=output.cached_tokens` in two `Usage(...)` constructions in `server.py` at v0.3.7:
- Line 1884, inside `_build_completion` (non-streaming text completion)
- Line 2225, inside `_build_chat_completion` (non-streaming chat completion)

The streaming paths in the same file already pass it through correctly.

**Screenshots**

N/A (CLI reproducer above).

 - macOS Version: 26.3.1 (build 25D771280a)
 - oMLX Version: 0.3.7

**Additional context**

Encountered while validating oMLX usage reporting for agent workflows where cache-hit accounting matters for cost/performance telemetry.

chulanpro5:
## Summary

Fixes #919 by reporting cache-hit counts in the OpenAI-compatible usage shape for:

- `/v1/chat/completions`
- `/v1/completions`

Instead of the previous top-level `usage.cached_tokens`, this change now reports cached prompt tokens under:

- `usage.prompt_tokens_details.cached_tokens`

This matches the current OpenAI schema more closely: [Completions object](https://developers.openai.com/api/reference/resources/completions)

## What changed

- Added `prompt_tokens_details` to the OpenAI usage model
- Populated `prompt_tokens_details.cached_tokens` in non-streaming and streaming OpenAI responses
- Updated the OpenAI adapter to emit the same nested usage shape
- Updated regression tests for the new response schema

### Unit / integration tests

```bash
pytest -q tests/test_openai_models.py tests/test_stream_usage.py
pytest -q tests/integration/test_server_endpoints.py -k 'cached_tokens_on_cache_hit'

### DIFF
--- a/omlx/api/adapters/openai.py
+++ b/omlx/api/adapters/openai.py
@@ -26,6 +26,7 @@ from ..openai_models import (
     ChatCompletionChunkDelta,
     ChatCompletionRequest,
     ChatCompletionResponse,
+    PromptTokensDetails,
     Usage,
 )
 from ..thinking import extract_thinking
@@ -133,6 +134,9 @@ class OpenAIAdapter(BaseAdapter):
                 prompt_tokens=response.prompt_tokens,
                 completion_tokens=response.completion_tokens,
                 total_tokens=response.prompt_tokens + response.completion_tokens,
+                prompt_tokens_details=PromptTokensDetails(
+                    cached_tokens=getattr(response, "cached_tokens", 0),
+                ),
             ),
         )
 
@@ -180,6 +184,9 @@ class OpenAIAdapter(BaseAdapter):
                 prompt_tokens=chunk.prompt_tokens,
                 completion_tokens=chunk.completion_tokens,
                 total_tokens=chunk.prompt_tokens + chunk.completion_tokens,
+                prompt_tokens_details=PromptTokensDetails(
+                    cached_tokens=getattr(chunk, "cached_tokens", 0),
+                ),
             )
 
         return f"data: {response.model_dump_json(exclude_none=True)}\n\n"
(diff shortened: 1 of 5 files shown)


jundot:
Nice catch and clean fix. The schema move to `prompt_tokens_details.cached_tokens` matches the current OpenAI shape, and all four non-streaming + streaming spots look covered. One nit: the adapter can use `response.cached_tokens` directly since `GenerationOutput` already declares it, so `getattr` with default is defensive but unnecessary. Will note the top-level `usage.cached_tokens` removal in release notes. Merging now, thanks!

PR #951, opened 2026-04-24
Add native TTS streaming

### LINKED ISSUE #948: Add low-latency streaming support to `/v1/audio/speech
**Summary**

It would be useful if the OpenAI-compatible TTS endpoint supported both normal buffered speech generation and low-latency streaming speech generation through the same `/v1/audio/speech` API.

**Requested Behavior**

`POST /v1/audio/speech` should continue to support the existing non-streaming behavior by default:

```json
{
  "model": "qwen3-tts",
  "input": "Hello world.",
  "voice": "Chelsie",
  "response_format": "wav"
}
```

This should return a complete `audio/wav` response after synthesis finishes.

The endpoint should also support a streaming mode:

```json
{
  "model": "qwen3-tts",
  "input": "Hello world.",
  "voice": "Chelsie",
  "response_format": "wav",
  "stream": true
}
```

When `stream: true` is set, the server should return an `audio/wav` streaming response as soon as audio is available, instead of waiting for the entire input to finish synthesis.

**Wishlist / Requirements**

- Keep `/v1/audio/speech` backward compatible for existing clients.
- Add a `stream: true` request option for streamed TTS.
- Streaming should initially support `response_format: "wav"`.
- The streamed response should emit one WAV header followed by PCM audio chunks.
- If the loaded TTS model supports native streaming, use the model-native stream path.
- If native streaming is unavailable, fall back to segment-level streaming where feasible.
- Add an optional `streaming_interval` request field to tune chunk cadence for native streaming.
- Reject empty or whitespace-only `input`.
- If streaming synthesis fails before the first chunk is available, return a normal HTTP error instead of `200 audio/wav` with an empty body.
- Preserve existing options such as `voice`, `instructions`, `speed`, `temperature`, `top_k`, `top_p`, `repetition_penalty`, `max_tokens`, `ref_audio`, and `ref_text`.

**Example curl**

```bash
curl http://127.0.0.1:8000/v1/audio/speech \
  -H "Authorization: Bearer XXXXX-XXXXXXX" \
  --no-buffer \
  -o speech_stream.wav \
  -d '{
    "model": "Qwen3-TTS-12Hz-1.7B-CustomVoice-8bit",
    "input": "Hello, this should begin playing before the full synthesis is complete.",
    "voice": "Chelsie",
    "response_format": "wav",
    "stream": true,
    "streaming_interval": 0.2
  }'
```

**Why**

This would improve time-to-first-audio for long assistant responses, voice agents, and interactive applications where waiting for the full WAV file makes the experience feel slow.

apetersson:
Closes #948

## Summary
- add `stream` and `streaming_interval` support to `/v1/audio/speech`
- stream WAV responses via native mlx-audio PCM chunks when available, with a segmented synthesis fallback
- validate streaming intervals before loading the engine and keep ref-audio cleanup safe
- add unit/integration coverage plus a live TTS sample generation helper

## Validation
- `uv run pytest tests/test_audio_tts.py -q`

## Notes
- intentionally leaves `streaming_plan.md` and `uv.lock` out of this PR for a cleaner feature diff

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,9 @@ coverage.xml
 .pytest_cache/
 cover/
 
+# Local generated artifacts
+artifacts/
+
 # Translations
 *.mo
 *.pot
--- a/omlx/api/audio_models.py
+++ b/omlx/api/audio_models.py
@@ -43,6 +43,8 @@ class AudioSpeechRequest(BaseModel):
     top_p: Optional[float] = None
     repetition_penalty: Optional[float] = None
     max_tokens: Optional[int] = None
+    stream: Optional[bool] = False
+    streaming_interval: Optional[float] = None
 
 
 class AudioProcessRequest(BaseModel):
--- a/omlx/api/audio_routes.py
+++ b/omlx/api/audio_routes.py
@@ -10,13 +10,16 @@ This module provides OpenAI-compatible audio endpoints:
 
 import base64
 import logging
-import tempfile
+import math
 import os
-from typing import Optional
+import re
+import tempfile
+from typing import AsyncIterator, Optional
 
 from fastapi import APIRouter, File, Form, HTTPException, UploadFile
-from fastapi.responses import Response
+from fastapi.responses import Response, StreamingResponse
 
+from ..engine.audio_utils import wav_bytes_to_pcm_frames, wav_header
 from .audio_models import AudioSpeechRequest, AudioTranscriptionResponse
 
 logger = logging.getLogger(__name__)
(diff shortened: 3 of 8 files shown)


jundot:
Looks good overall. The capability detect (requiring both `stream` and `streaming_interval`) and the prefetch-first-chunk pattern for clean error handling are nicely done.

One small thing: `scripts/generate_tts_count_1_to_30.py` looks like a personal demo artifact (hardcoded prefix, model, voice). The integration test already covers live streaming, so I'll drop the script and the `.gitignore` `artifacts/` entry in a follow-up.

Merging.

PR #960, opened 2026-04-25
Fix: Seed TokenBuffer with cached prompt tokens on cache hits

Yukon:
# Issue
While investigating the root cause for repetition issues with models it was discovered that the we are not restoring the `TokenBuffer` in mlx_lm during cache hits. When a prior turn’s KV cache was restored, the penalty state was not restored, resulting in penalty processors starting with an empty context and failing to penalize tokens from previous generation.

# Fix
Passing `all_tokens` to `BatchGenerator.insert()` fixes this so the `TokenBuffer` is initialized with the full cached prompt prefix. This gives penalty processors visibility into tokens from the restored cache. As is done by [sample server code](https://github.com/ml-explore/mlx-lm/blob/ed1fca4cef15a824c5f1702c80f70b4cffc8e4dd/mlx_lm/server.py#L776-L780) in mlx_lm.

# Known Call Outs
When using SpecPrefill tokens that are not selected during KV cache generation will still be in `request.prompt_token_ids` and penalized. This approach was taken for a user centric approach to penalization based on the conversation and not specifically penalize only tokens being seen and generated by model.

# Next Steps
As part of continuing to address n-gram repetition in oMLX there will be a follow up PR to address the small penalty window that currently uses the default 20 token context size. A window size that is far too small to effectively penalize CoT models like Qwen 3.6 from repetition.

_🤖 AI was used to generate code, work reviewed and validated by a human._


jundot:
Thanks @Yukon for tracking this down and pointing to mlx-lm's `all_tokens` path.

Current `main` has diverged enough that I ported the core fix directly instead of merging this branch: `BatchGenerator.insert()` now seeds `all_tokens` in both the normal/cache-hit path and the chunked-prefill completion path. I also added scheduler coverage for cache-hit, full-prefill, single-token, concurrent, and chunked-prefill cases.

The direct fix is in 6d75d6d, so I'm going to close this PR with credit linked from the commit.

PR #962, opened 2026-04-25
fix(tokenizer): prefer lm_head weight as vocab_size source

contrapuntal:
## What this fixes

Users running grammar-constrained output (`[grammar]` extra) on GLM-4.6V — and on several other VLM families — hit a shape-mismatch crash on the first constrained-decode step, with no clear error about why. The xgrammar bitmask is allocated for the wrong vocabulary, then doesn't fit the model's actual logits tensor.

The root cause is `omlx/utils/tokenizer.py::resolve_vocab_size()` returning the wrong number, which then propagates to every consumer that sizes logits-aligned buffers from it. Two such consumers: `omlx/api/grammar.py::create_grammar_compiler()` (xgrammar bitmask) and the scheduler's `_get_model_vocab_size()` bookkeeping.

## Affected models

Three mlx-vlm `ModelConfig` dataclasses ship a hard-coded top-level `vocab_size` default that doesn't match the inner LM's actual vocab when `config.json` omits the top-level key (as it typically does for these models):

| mlx-vlm dataclass | `ModelConfig.vocab_size` default | Actual `text_config.vocab_size` |
| ----------------- | -------------------------------- | ------------------------------- |
| `glm4v` (GLM-4.6V) | `257152`                        | `151552`                        |
| `glm4v_moe`        | `257152`                        | per-model from `config.json`; e.g. `151552` for GLM-4.6V-Flash |
| `gemma3`           | `257152`                        | `262208`                        |

`resolve_vocab_size()` previously read the top-level value first, so it returned the wrong vocab on any of the above. Concrete public shape: GLM-4.6V has `text_config.vocab_size=151552` but `ModelConfig.vocab_size=257152` as a dataclass default; sizing from the top-level value mismatches the real 151552-wide logits.

## Fix

Re-order the resolution to prefer authoritative sources:

1. **`lm_head.weight.shape[0]`** — authoritative; this is the exact vocabulary the model emits logits over, regardless of whatever the config dataclass says. Probed under `_language_model.lm_head` (VLM adapter), `language_model.lm_head` (raw mlx-vlm), and `lm_head` (raw mlx-lm) so the lookup works for every wrapping shape we ship.
2. **`text_config.vocab_size`** — inner LM vocab on VLM composite configs (object and dict forms). Correct when `lm_head` is unavailable.
3. **`model.config.vocab_size` / `model.args.vocab_size`** — original top-level fallback, consulted last for compatibility with pure-LLM configs that don't have a `text_config`.

For pure-LLM models the new order returns the same value the old order did, because `lm_head.weight.shape[0]` and `config.vocab_size` agree. The change only affects models where the two disagree, which is exactly the broken case.

## Rebase (2026-09-09)

Rebased onto current `jundot/omlx` `main` at `94530d8d` (`feat(cluster): make TP and pipeline serving request-safe (#3258)`). The old head `b1f91b1d` conflicted because upstream's `omlx/utils/tokenizer.py` has since grown (new imports, detokenizer helpers, LFM2/Laguna/Mistral-common tokenizer-config handling) and `tests/test_utils_tokenizer.py` gained `TestMistralCommonTokenizerConfig`.

Conflict resolution, no behavior change beyond the intended fix:

- `omlx/utils/tokenizer.py`: kept upstream's file intact and replaced only `resolve_vocab_size()` with the lm_head-first implementation (double-quote style to match the file).
- `tests/test_utils_tokenizer.py`: kept upstream's `TestMistralCommonTokenizerConfig` and added the PR's `TestResolveVocabSize` (13 cases) alongside it; imports now include both `repair_misconverted_unlimited_ocr_tokenizer` and `resolve_vocab_size`.

New head: `95e5ae21`.

## Verification (fresh, on the rebased head)

Run in an isolated worktree against the rebased commit with the project venv (Python 3.12.11):

```
$ PYTHONPATH=<worktree> .venv/bin/python -m pytest tests/test_utils_tokenizer.py -q --tb=short
66 passed in 2.70s

$ PYTHONPATH=<worktree> .venv/bin/python -m pytest tests/test_grammar.py -q --tb=short
93 passed, 3 skipped in 2.54s

$ PYTHONPATH=<worktree> .venv/bin/python -m pytest tests/test_grammar.py::TestGetModelVocabSize -q
4 passed

$ .venv/bin/python -m ruff check omlx/utils/tokenizer.py tests/test_utils_tokenizer.py
All checks passed!
$ .venv/bin/python -m ruff format --check omlx/utils/tokenizer.py tests/test_utils_tokenizer.py
2 files already formatted
```

Resolver spot-check on the rebased code with a GLM-style stub (conflicting top-level `257152`, `text_config` `151552`, `lm_head` width `151552`): returns `151552`; the same stub without `lm_head` falls back to `text_config` (`151552`); `resolve_vocab_size(None)` returns `None`.

## Verification limits

- The resolver check above is a weight-free stub comparison, not a model-inference run.
- No fresh end-to-end constrained-decode run (GLM-4.6V-Flash with `[grammar]`, Llama 3.1, Qwen2-VL) was performed in this pass. The original PR's end-to-end claims (151552-wide bitmasks matching logits; unchanged ints on Llama 3.1 / Qwen2-VL) are historical and are **not** presented as newly verified here.
- Historical counts from the pre-rebase PR (`41 passed` on the tokenizer file) are superseded by the `66 passed` result above; the file has grown upstream since.

## Why not fix this in mlx-vlm

The dataclass defaults in `glm4v`, `glm4v_moe`, `gemma3` `ModelConfig` are arguably wrong (they should be `None` or read from `text_config`), and an upstream fix would obsolete most of this PR. But:

* The defaults exist on purpose for some mlx-vlm internals that instantiate `ModelConfig` without a `config.json`, so changing them is a behavior question, not just a bug fix — out of scope here.
* Even with mlx-vlm fixed, `lm_head.weight.shape[0]` is still the more authoritative source — config can drift from weights, weights can't drift from themselves. The new resolution order also defends against runtime config-weight drift on any future model where the two could disagree, not just the specific dataclasses currently broken.

## Risk

The new lm_head probe is gated by `getattr` chains and a defensive shape extraction (try/except around `shape[0]`), so it returns `None` (falling back to the old config path) on any model that doesn't expose `lm_head` in any of the three wrapping shapes, or whose `weight.shape` is not subscriptable. No code that previously succeeded should now fail.

Tied embeddings: `lm_head.weight` is `embed_tokens.weight` and still maps accurately to the logits vocab dimension.

Padded vocabularies: when a model pads its vocab to a multiple of 64 (or similar) for tensor-core efficiency, `lm_head.weight.shape[0]` returns the padded size — which is exactly what logits-aligned buffers (e.g. `xgrammar` bitmasks) need to allocate to avoid shape mismatches against the actual logits tensor. The new resolution order is strictly safer than the old `config`-first order in this case too.

### DIFF
--- a/omlx/utils/tokenizer.py
+++ b/omlx/utils/tokenizer.py
@@ -39,10 +39,23 @@ def unwrap_tokenizer(tokenizer):
 
 
 def resolve_vocab_size(model: Any) -> int | None:
-    """Extract vocab_size from a model's config/args, handling nested configs.
+    """Extract vocab_size from a model, preferring the authoritative source.
 
-    Tries ``model.config.vocab_size``, then ``model.args.vocab_size``,
-    then ``text_config.vocab_size`` for VLM composite models (e.g. Qwen3.5).
+    Resolution order:
+    1. The ``lm_head`` weight's first dimension (authoritative — this is the
+       exact vocabulary the model emits logits over).
+    2. ``text_config.vocab_size`` when present (the inner language model's
+       vocab on VLM composite configs).
+    3. ``model.config.vocab_size`` / ``model.args.vocab_size`` (top-level).
+
+    Why lm_head and text_config come first for VLMs: several mlx-vlm
+    ``ModelConfig`` dataclasses (e.g. glm4v, glm4v_moe, gemma3) hard-code a
+    top-level ``vocab_size`` default that does not match the inner LM vocab
+    when ``config.json`` omits the top-level key.  For example, GLM-4.6V has
+    ``text_config.vocab_size=151552`` but ``ModelConfig.vocab_size=257152``
+    as a dataclass default.  Code that sizes logits-aligned buffers (e.g.
+    grammar bitmasks) from the top-level value produces a shape mismatch
+    against the real (151552) logits.
 
     Args:
         model: An MLX model object (LLM, VLM, or any object with config/args).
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for finding the root cause and putting this fix together. I'm sorry I left this waiting and merged #3551 for the same issue without checking your earlier PR. I verified the fix on GLM-4.6V, so I'll close this as superseded. I appreciate the work you put into this.

PR #977, opened 2026-04-27
[codex] add Russian admin localization

DrMaks22:
## What changed
- Added a full `ru` locale for the oMLX admin UI.
- Added Russian to the language selector in Admin Settings.
- Kept all locale key sets aligned across admin translations.
- Updated the README language list to mention Russian.

## Why
- The admin dashboard already supports JSON-based localization, so Russian can be added without changing runtime behavior.
- This makes the UI usable for Russian-speaking operators while keeping the implementation aligned with the existing i18n structure.

## Validation
- Verified that every locale file has the same key set as `en.json`.
- Verified `ru.json` parses as valid JSON.
- Confirmed the repo builds the locale selector with `ru` included.


jundot:
Looks good to me. I checked the locale alignment (all 6 files have the same 652 keys), placeholder preservation (zero mismatches against `en.json`), JSON validity, and the new option in the language selector. The 38 entries kept in English (brand names, API field labels like `Top P` / `ctx_window:`, benchmark acronyms like `TTFT` / `TPOT`) match what other locales do, so that's consistent.

No regression risk on my side: `_load_locale()` already loads any `{lang}.json` dynamically with an `en` fallback, and `UISettings.language` is a free string so `ru` flows through untouched. Inter handles Cyrillic fine, so no font work needed either.

Thanks for the contribution!

PR #995, opened 2026-04-28
Add guided grammar model setting

MrNiceRicee:
## Summary

Adds guided grammar support as a first-class model setting in the admin dashboard.

This exposes a generic `guided_grammar` control without adding model-specific presets. The request path normalizes `guided_grammar` into the existing `structured_outputs.grammar` flow, so constrained decoding continues to use the current xgrammar-backed implementation.

## Changes

- Add `guided_grammar` request field as a compatibility alias
- Add `guided_grammar_enabled` and `guided_grammar` to persisted model settings
- Surface guided grammar in `/admin/api/models` and model settings update API
- Add a Guided Grammar textarea/toggle to the model settings modal
- Include guided grammar in profiles/templates as a universal profile field
- Trim empty grammar values before persistence
- Add tests for request parsing, settings roundtrip, admin API exposure, and precedence behavior
- Classify existing `preserve_thinking` as profile-excluded to keep model settings field classification complete

## Notes

Actual constrained decoding still requires the optional grammar dependency, `xgrammar`, as before. Without it, structured output requests return the existing dependency error.

## Validation

- `14 passed` focused guided grammar tests
- `71 passed` settings/profile regression tests
- `157 passed` nearby request/admin/settings suite
- `59 passed, 3 skipped, 5 deselected` grammar suite excluding optional structural-tag patch tests
- JSON i18n validation
- Python syntax check
- dashboard JS syntax check
- `git diff --check`

image


MrNiceRicee:
I tested this locally on Qwen3.6-27B-MLX-8bit; grammar-constrained run reduced output from 400 tokens / 29.4s to 94 tokens / 7.6s on one coding prompt.

got the idea from
https://andthattoo.dev/blog/structured_cot

if possible, would love to have a guided grammar setting

fry69:
I just toyed around with the PR, as it is now viable with `xgrammar` being included in the macOS app (stubbing out the heavy unused Torch). There are two minor rebase conflicts, but those are easily solvable just by guessing.

I do not know if this is magic wand, but it reduces runaway thinking and forces the model going quickly into reply mode, which at least makes it more interactive.

This may be a middle ground between "no thinking" and "thinking forever", hopefully skipping "broken thinking due to budget constraints", as this enforced thinking is likely to be much more focussed, but that is just an intuition of course.

I'd like to see this included 👍

If anyone wants to explore this, my merged version is here -> https://github.com/fry69/omlx/tree/fry69/guided-grammar

jundot:
Thanks for putting this together. I checked the request normalization, model settings roundtrip, admin API exposure, and the dashboard wiring against the current merge ref.

The guided_grammar alias correctly folds into the existing structured_outputs.grammar path, and the model-level default is scoped so explicit structured output requests still take precedence. I found one small cleanup around the existing preserve_thinking profile-field entry, but it is not blocking this PR and I will handle it as a follow-up on main.

PR #1008, opened 2026-04-29
Return `input_tokens_details` in Responses API usage information

hardlycharred:
Return dummy value in Responses API Usage information, per OpenAI's Responses API spec [here](https://developers.openai.com/api/reference/resources/responses/methods/create). 

Being hard coded to 0 is not the most useful thing, but it does fix interactions with downstream tools that expect the field to be present in `.response.usage`. One such example is agentgateway/agentgateway#1686 , where the absence of `input_tokens_details` breaks usage tracing entirely for oMLX's Responses API.


jundot:
thanks for the patch! merging — I'll do a tiny follow-up on main to clean up the docstring (says "output" instead of "input") and to plumb the real cached_tokens value through (we already track it on the request output, so 0 was leaving info on the floor). non-streaming path picks up the field automatically via the pydantic default, but I'll wire the helper too for consistency.

hardlycharred:
Thank you! and thanks for all your work on this project :)

PR #1028, opened 2026-05-01
Gemma 4 interleaved tool call and reasoning fix

fabiopili:
Multi-turn tool-call flows with Gemma 4 in Open WebUI were leaking the literal word "thought" into the visible answer and producing empty `<|channel><|channel>` sequences. This often would break the separation between reasoning blocks and response blocks, mixing both together in the UI. 

After reviewing [the model card](https://ai.google.dev/gemma/docs/core/model_card_4#2_thinking_mode_configuration), I noticed that this isn't something specific to Open WebUI, but an incomplete implementation of thinking blocks in multi-turn conversations, particularly when there are interleaved tool calls in the middle. 

```
3. Multi-Turn Conversations

No Thinking Content in History: In multi-turn conversations, the historical model output should only include the final response. Thoughts from previous model turns must not be added before the next user turn begins.
```

This change addresses both the root cause and the parser fragility that made it visible. I've been running it without problems on top of v0.3.6 and then now on v0.3.8.

## Background

Gemma 4's multi-turn rule states that [only the final visible answer should be kept in chat history](https://ai.google.dev/gemma/docs/core/model_card_4#3_multi-turn_conversations) and that prior thought blocks must not be fed back into the next turn. The previous adapter code passed assistant content through verbatim, so clients such as Open WebUI replayed the rendered `<think>...</think>` block (or the raw `<|channel>thought...<channel|>` form when preserved) on every subsequent turn. The polluted history then primed the model to emit malformed channel markers on follow-up turns.

The output parser only recognised the canonical open form `<|channel>thought\n`. When the model emitted the no-newline variant `<|channel>thought<channel|>` or a bare `<|channel>`, the parser failed to match the open, treated everything up to the close as visible text, and silently dropped the close. The literal `thought` keyword and any malformed channel content then leaked into the user-facing reply.

## Changes

`omlx/adapter/gemma4.py`:

1. Added a `_strip_thinking` helper that removes leading `<think>...</think>` or raw `<|channel>...<channel|>` spans from assistant content inside `extract_gemma4_messages` before the message reaches the chat template. The match is anchored to the start of the message because the rendered thought block always precedes the visible answer, which keeps inline mentions intact (an assistant explaining how reasoning tags work, for example, is preserved). This honours Gemma 4's multi-turn history rule and prevents the model from being primed by its own reasoning output.

2. Added `<|channel>` as a defensive bare-open marker in `Gemma4OutputParserSession`. A streaming-defer guard ensures that when the bare match could still extend to the canonical form once more tokens arrive, the parser buffers and waits, preserving canonical streaming. When the bare form does commit, the parser optionally absorbs a trailing `thought` keyword and newline so they do not leak as visible text. Stray bare opens are wrapped in `<think>` rather than allowed through.

## Test plan

- [x] `uv run pytest tests/test_gemma4_messages.py` passes (14 tests).
- [x] Manual smoke tests cover canonical, no-newline, bare, double-bare, and chunked-streaming inputs.
- [x] End-to-end verification in Open WebUI with a multi-turn tool-call flow that previously reproduced the leak.
- [x] End-to-end verification using the Pi coding agent in a local codebase and Gemma 4 26B A4B as the model.

An asterisk here about testing: if I run Gemma 4 in Pi in this branch, it'll strip out the reasoning tags in the thinking and response blocks. This seems to be an issue in Pi itself as I can reproduce it if I ask the model, from a new session, to simply print a thinking tag in the chat. When I enter a thinking tag myself, it's also stripped before reaching the model.

I'm mentioning this only for completeness, but it isn't a regression. Running on the current main, at v0.3.8, has the same result.

## Samples

Attached are screenshots from Open WebUI and corresponding full trace logs for v0.3.8 vs. this branch.

### Logs

v0.3.8
[omlx-trace.main.log](https://github.com/user-attachments/files/27272523/omlx-trace.main.log)

Fix
[omlx-trace.fix.log](https://github.com/user-attachments/files/27272527/omlx-trace.fix.log)

### Screenshots

v0.3.8
v0 3 8-top

Notice the mixup between reasoning block and response:

v0 3 8-bottom

Fix
fix

### DIFF
--- a/omlx/adapter/gemma4.py
+++ b/omlx/adapter/gemma4.py
@@ -24,8 +24,10 @@ _TOOL_RESPONSE_CLOSE = "<tool_response|>"
 _THINK_OPEN = "<think>\n"
 _THINK_CLOSE = "</think>\n"
 
-_THINK_BLOCK_RE = re.compile(r"<think>.*?</think>\s*", re.DOTALL)
-_CHANNEL_BLOCK_RE = re.compile(r"<\|channel>.*?<channel\|>\s*", re.DOTALL)
+_LEADING_THOUGHT_RE = re.compile(
+    r"\A\s*(?:(?:<think>.*?</think>|<\|channel>.*?<channel\|>)\s*)+",
+    re.DOTALL,
+)
 
 
 def _try_parse_json(s: str) -> Any:
@@ -42,7 +44,7 @@ def _try_parse_json(s: str) -> Any:
 
 
 def _strip_thinking(text: Any) -> Any:
-    """Remove ``<think>...</think>`` and raw ``<|channel>...<channel|>`` spans.
+    """Remove leading ``<think>...</think>`` or raw ``<|channel>...<channel|>`` spans.
 
     Gemma 4's multi-turn rule requires that only the final visible answer
     is kept in chat history. Clients such as Open WebUI replay the full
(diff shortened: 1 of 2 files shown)


jundot:
LGTM. The model card citation + history strip + defensive parser marker line up cleanly against the reported symptom.

I traced the streaming defer path (canonical wins over bare, suffix buffered when it could still extend) and the chunk-boundary cases hold up. The `\A` anchor in `_strip_thinking` keeps inline `<think>` mentions intact, which I think was the right call.

14 tests cover the corner cases I'd want covered. Merging this in. Thanks for the thorough fix and the trace logs.

Issue #1005, opened 2026-04-29
Cannot find library: libxgrammar_bindings.dylib, libxgrammar_bindings.so


nickcolea:
**Describe the bug**
After installing `brew reinstall omlx --with-grammar`, I keep getting these `omlx.server - ERROR - [-] - GET /admin/api/grammar/parsers → 500 (unhandled): Cannot find library: libxgrammar_bindings.dylib, libxgrammar_bindings.so` log errors.

**Expected behavior**
UI and xgrammar template options should be available. 

**Desktop (please complete the following information):**
 - macOS Version: 26.4.1
 - oMLX Version 0.7.8rc2

zviratko:
How does this PR fix the original issue? The problem is that xgrammar can't load the libraries, that shouldn't happen if it's compiled with --with-grammar (yes, I see it too).
It might be caused by this commit, maybe? https://github.com/mlc-ai/xgrammar/commit/6f52951c5e5a9cdea47f5facda3e062fa8cab457
I coerced my agent to write a RECORD file for me so those native libraries load - that actually removes the error which I presume means it should work. There's another error however, where the UI doesn't allow configuring Reasoning parser now...?
This commit would just mask the error.

crienzo:
I had claude debug it and it came up with this as a temporary solution when building w/ `brew` until the upstream is resolved:

```
#!/bin/bash
# Install or upgrade omlx with grammar support, and patch the broken
# xgrammar macOS arm64 wheel so structured-output / tool-arg enforcement works.
#
# The xgrammar 0.1.32+ wheel for macosx_arm64 ships with two bugs:
#   1. libxgrammar_bindings.dylib has @rpath/libtvm_ffi.dylib but no
#      LC_RPATH pointing at where tvm_ffi installs its native lib.
#   2. The wheel's dist-info is missing RECORD, so tvm_ffi's
#      manifest-based library lookup falls through to a fallback that
#      doesn't search the package directory.
#
# Both patches are applied to the on-disk Cellar after install. They are
# wiped by the next brew upgrade/reinstall, so re-run this script then.
#
# Run as the admin user (the one who owns Homebrew). Idempotent.

set -euo pipefail

TAP=jundot/omlx
TAP_URL=https://github.com/jundot/omlx
FORMULA=jundot/omlx/omlx

echo "==> Ensuring tap..."
brew tap | grep -qx "$TAP" || brew tap "$TAP" "$TAP_URL"

echo "==> Refreshing tap metadata..."
[...]

PR #1042, opened 2026-05-02
fix(admin): support xgrammar 0.1.34+ registry API for parser discovery

crienzo:
References: https://github.com/jundot/omlx/issues/1005 

xgrammar 0.1.34 dropped the top-level
get_builtin_structural_tag_supported_models() helper in favor of a per-model registry (xgrammar.builtin_structural_tag._structural_tag_registry) plus register_model_structural_tag / get_model_structural_tag. With the old helper gone, GET /admin/api/grammar/parsers hit ImportError, the narrow except returned [], and the admin Reasoning Parser dropdown was silently hidden by the reasoningParsers.length > 0 gate in the Alpine template. Newer parser styles like qwen3_6, gemma4, and deepseek_v4 were unreachable as a result.

Update list_grammar_parsers to try the 0.1.34+ registry first (parsing each function's docstring for the "Supported models:" block) and fall back to the pre-0.1.34 helper. Broaden the exception swallow so a broken native binding (xgrammar/xgrammar#1005-style RuntimeError) also yields an empty list rather than a 500.

The compilation call site in server.py needs no change: xgrammar 0.1.34 keeps get_builtin_structural_tag as an alias for
get_model_structural_tag with a backward-compatible signature.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -9,9 +9,11 @@ This module provides HTTP routes for the admin panel including:
 """
 
 import asyncio
+import inspect
 import json
 import logging
 import os
+import re
 import secrets
 import shutil
 import sys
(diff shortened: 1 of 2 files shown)


jundot (approved):
Thanks for the fix. The fallback chain (0.1.34+ registry, legacy helper, empty list) plus broadening `ImportError` to `Exception` covers both the API rename and the libxgrammar_bindings load failure from #1005. Test coverage hits all four branches. `_structural_tag_registry` is a private attr so a future xgrammar refactor could silently empty the dropdown again, but the graceful degradation makes that acceptable for an admin-only surface. Merging.

PR #1043, opened 2026-05-02
Fix/xgrammar macos dylib

crienzo:
References: https://github.com/jundot/omlx/issues/1005

## Summary

  Patch the macOS arm64 xgrammar wheel after install so its native binding actually
  loads. Without this, `import xgrammar` raises and the admin Reasoning Parser
  dropdown stays hidden — the original symptom in #1005.

  The patch is split across the two root causes:

  - **Add `LC_RPATH`** pointing at the `apache-tvm-ffi` package's lib directory,
    then ad-hoc re-codesign. The xgrammar wheel ships
    `libxgrammar_bindings.dylib` with `@rpath/libtvm_ffi.dylib` declared but no
    rpath pointing at where `libtvm_ffi.dylib` actually lives, so dyld fails to
    resolve it. (See #1005 comment for `otool -l` evidence — zero `LC_RPATH`
    entries on the stock 0.2.0 wheel.) **This is an upstream xgrammar wheel
    bug**; fix suggestion for mlc-ai/xgrammar:
    `-Wl,-rpath,@loader_path/../tvm_ffi/lib` 

  - **Append the dylib to dist-info `RECORD`** so `tvm_ffi.libinfo._find_library_by_basename`
    can locate it. The stock pip wheel ships RECORD correctly, but Homebrew's
    post-install relocation pass deletes every `dist-info/RECORD` in the keg as
    part of stripping invalidated SHA256 hashes (visible in `brew install
    --verbose` as a long list of `rm .../dist-info/RECORD` lines, including
    `xgrammar-*.dist-info/RECORD`). This is a **brew × tvm_ffi collision**, not
    an xgrammar or omlx bug per se — but it manifests on every brew install
    and the only place to fix it without touching brew or tvm_ffi is in our
    Formula.

  Both manifest as the same symptom (`RuntimeError: Cannot find library: ...`
  or `dlopen(...): Library not loaded: @rpath/libtvm_ffi.dylib`) at
  `import xgrammar`, so they're patched together.

  ## Why post_install, not install

  Brew's "Cleaning" step runs **after** `def install` returns and **before**
  `def post_install`. Anything written into RECORD in `install` is wiped before
  the user ever runs Python. I confirmed this empirically — an earlier version
  of this patch lived in `def install`, the build-time verification call
  (`python -c "import xgrammar"`) printed `xgrammar import OK`, and then
  post-install the import failed because brew had cleaned up after us. Moving
  the patch to `post_install` puts it after the cleaning pass.

## Test plan

  - [x] `brew install --HEAD --build-from-source --verbose --with-grammar` —
        `==> Patching xgrammar macOS arm64 wheel` block fires *after* the
        `==> Cleaning` rm spam, ending with `xgrammar import OK`.
  - [x] `python -c "import xgrammar"` succeeds post-install with no manual
        intervention.
  - [x] Reproduced both failure modes by force-reinstalling xgrammar (which
        wipes the rpath) and confirmed the import re-breaks with the exact
        `dlopen(...): Library not loaded: @rpath/libtvm_ffi.dylib` message;
        re-running the patch logic restores it.
  - [x] Idempotent: re-running the patch (`brew reinstall`, manual re-run,
        etc.) is a no-op on already-patched installs.
  - [x] Gated on `build.with?("grammar")` — non-grammar installs do nothing.

  ## Reservations

  This is a downstream workaround for a wheel-side bug + a brew design
  choice. The cleanest long-term fix is upstream xgrammar building a wheel
  with the correct rpath, at which point Bug 2a goes away entirely. Bug 2b
  (brew × RECORD) would still bite anyone whose wheel relies on RECORD-based
  manifest discovery, but the surface area shrinks dramatically once xgrammar
  itself works without manifest fallbacks.

  If this approach doesn't fit the project — e.g., you'd rather wait for
  upstream xgrammar to ship a corrected wheel — totally reasonable to close
  this PR and link the upstream issue in #1005 as the resolution path. I'll
  keep the patch on my fork either way.

### DIFF
--- a/Formula/omlx.rb
+++ b/Formula/omlx.rb
@@ -54,6 +54,40 @@ class Omlx < Formula
     # python-multipart is declared in omlx's [audio] extra, not in mlx-audio
     system libexec/"bin/pip", "install", "python-multipart>=0.0.5"
 
+    # Patch the macOS arm64 xgrammar wheel so its native binding loads.
+    # The 0.1.32+ wheel ships libxgrammar_bindings.dylib with
+    # @rpath/libtvm_ffi.dylib but no LC_RPATH pointing at where tvm_ffi
+    # installs its native lib, and the dist-info is missing a RECORD
+    # entry for the dylib so tvm_ffi's manifest-based lookup fails.
+    # Both manifest as RuntimeError("Cannot find library: ...") at
+    # `import xgrammar`, which crashes /admin/api/grammar/parsers and
+    # hides the Reasoning Parser dropdown. Tracking upstream:
+    # jundot/omlx#1005.
+    if build.with?("grammar")
+      py = libexec/"bin/python"
+      site = Utils.safe_popen_read(py, "-c",
+                                   "import site; print(site.getsitepackages()[0])").chomp
+      dylib = "#{site}/xgrammar/libxgrammar_bindings.dylib"
+      dist  = Dir["#{site}/xgrammar-*.dist-info"].first
+      tvmlib = Utils.safe_popen_read(py, "-c",
+        "import os, tvm_ffi; print(os.path.join(os.path.dirname(tvm_ffi.__file__), 'lib'))").chomp
+
+      if File.exist?(dylib)
+        rpaths = Utils.safe_popen_read("/usr/bin/otool", "-l", dylib)
+        unless rpaths.include?(tvmlib)
+          system "/usr/bin/install_name_tool", "-add_rpath", tvmlib, dylib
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for the thorough investigation. The `otool` evidence and the `install` vs `post_install` reasoning made this easy to review, and both root causes match what I see on my end.

Merging as-is. I'll swap the `/bin/sh -c "echo ..."` line for a Ruby `File.open(record, "a") { |f| f.puts "..." }` in a follow-up commit so the file write stays in pure Ruby, but that is pure polish on top of an already solid patch. Thanks again for digging into this.

PR #1052, opened 2026-05-04
feat(bench): Add opt-out toggle for community benchmark upload

jabagawee:
## Summary

Throughput benchmarks upload results to omlx.ai unconditionally at the end of every run, including a stable per-device hash and the model name (which can be a private/local model id). There is no UI toggle, env var, or request flag to disable it.

This PR adds a "Share results with the omlx.ai community" checkbox to the throughput benchmark config card, **checked by default so existing user behavior is unchanged**. The frontend sends the user's choice as `should_upload_results` on `POST /api/bench/start`; the `run_benchmark` coroutine gates the call to `_upload_to_omlx_ai` on that flag. When the box is unchecked, the SSE `done` handler finalizes the run directly instead of entering the "Uploading..." UI state.

The hint text spells out exactly what gets uploaded and notes the common reason to uncheck (benchmarking a private model whose name would otherwise be included in the payload).

## Notes

- `BenchmarkRequest.should_upload_results` defaults to `True`, so any direct API caller keeps current behavior.
- Only `en` i18n strings are added; other locales fall back via the resolver in `base.html`.
- The accuracy benchmark has no upload path and is unaffected.


jabagawee:
updated the PR to have the checkbox checked by default (now it's opt-out instead of opt-in) to match default behavior to existing behavior

Issue #804, opened 2026-04-16
[Feature] Stop inference if the client closes the connection


zulufoxtrot:
**Is your feature request related to a problem? Please describe.**

Unsure if feature request or bug report.

I expect oMLX to stop processing a request if the client prematurely closes the connection.

Typical use case: client performs a request. oMLX starts processing it. Client times out (hard limit to 5 minutes in my client, Kilo Code). oMLX continues processing the response.

**Describe the solution you'd like**

oMLX kills the process linked to the request.

**Describe alternatives you've considered**

**Additional context**
0.3.5

Issue #484, opened 2026-03-31
Qwen3.5-35b-a3b working but stops halfway


murugaratham:
**Describe the bug**
I’m using Hermes agent and noticed that agents tends to stop midway and only continues when I say something.

**Additional context**
Could it be due to tool calls in reasoning? https://www.reddit.com/r/LocalLLaMA/s/1ZVs0ubiQZ

shsieh:
Yes, this is the problem. The tool call in reasoning block that causes the agent to stop. LM Studio have a fix for it in the jinja template. In omlx, I am not sure how to add the entire template into it since I don't see any option for me to do so. One temp fix for this is to disable reason/thinking for now.

celangoni:
> In omlx, I am not sure how to add the entire template into it since I don't see any option for me to do so

Chat template can be changes in chat_template.jinja, inside the model directory.
I'm facing tool call erros on Qwen3.5 models using AgentZero.
File "/[a0](http://localhost:7088/#)/[agent.py](http://localhost:7088/#)", line 958, in validate_tool_request
    raise ValueError("Tool request must have a tool_args (type dictionary) field")

I have changed chat_template.jinja with one I found on reddit, but nothings change.
I'm using v0.3.0 dmg release

What additional information I should provide to help in this case?

EDIT: With Qwen3.5-35B-A3B-4bit  I got success and good results

jundot:
Thanks for reporting this. This is likely a model limitation rather than an omlx bug. Qwen3.5-35B-A3B has only 3B active parameters and tends to lose tool calling coherence in longer contexts, sometimes generating tool call syntax inside `<think>` blocks instead of proper output. I saw the same pattern in #260 where token-level comparison confirmed omlx produces identical output to pure mlx-lm.

Omlx's thinking fallback was making this worse though by extracting those stray tool calls from thinking blocks and treating them as real invocations. I've tightened the fallback so it only recovers tool calls from thinking when the model produced no regular text content and the tool name matches a provided tool definition.

For now, disabling thinking as @shsieh suggested is the quickest workaround.

@celangoni the `tool_args` error traces back to AgentZero's own `validate_tool_request`, not omlx. Could you open a separate issue with the full traceback and a trace-level server log so i can check the raw model output?

Issue #352, opened 2026-03-22
[Bug] Server stops responding from time to time


zulufoxtrot:
**Describe the bug**

oMLX v0.2.19
M2 Mac studio, MacOS 15.7.5

Over a 1-hour vibe coding session, I've experienced what seems to be a couple denials of service from the server.
After a while, any connection attempt to the API endpoint results in either:

- a connection error (not an HTTP error, a TCP connection error). 
- a timeout error

After 1 or 2 minutes, the service returns.

Context:
- oMLX is running on Mac 1
- the client is on Mac 2
- both Macs are on local network

Symptoms:
- The service is still running (or appears so: memory usage is still high and service port is still reported open)
- The host is still accessible and I can still SSH into it.
- No suspicious logs (level is INFO)
- Session stats do not reset, so I assume the service didn't restart

**Additional context**

I don't know how to reproduce consistently.

jundot:
Thanks for reporting this. I looked into what could cause this kind of intermittent unresponsiveness.

There are a few things that could be going on here:

1. **Long prefill blocking** - when processing a large prompt, the scheduler's prefill loop runs synchronously and can block for a long time. During this, the server can't process new connections properly.

2. **Memory pressure** - over a long session, KV cache accumulates in unified memory. When macOS starts compressing/swapping pages under pressure, the whole process can freeze for a while. This would explain the TCP-level connection failures (not just HTTP timeouts).

3. **Periodic cleanup stalls** - the scheduler runs `mx.synchronize()` + `mx.clear_cache()` and `gc.collect()` at regular intervals, which can stall under heavy memory usage.

The tricky part is that 2 and 3 happen at the OS/runtime level, so there's limited what i can do from the application side.

To narrow this down, could you share some details about your setup when this happens?

- Which model are you running and roughly how large are the prompts?
- How much memory does the oMLX process use when it stops responding? (activity monitor or `memory_pressure` command output would help)
- Does it happen more towards the end of long sessions vs early on?
- If you restart oMLX and do the same workload, does it take roughly the same amount of time before it happens again?

Also if you can run with `--log-level debug` next time, that would help figure out whether the server is stuck in a prefill or doing a cache cleanup when this happens.

zulufoxtrot:
Thanks!

> Which model are you running

Qwen3-coder-next-4b.

> roughly how large are the prompts

In terms of input tokens? If the new "live prompt processor" in oMLX is an indicator, sometimes up to 40k.

[...]

Issue #260, opened 2026-03-16
[Bug] Qwen3.5-35B-A3B stops abruptly during token generation (multi-turn with Claude / OpenCode)


c4software:
### Describe the bug
Using oMLX (latest version), the **Qwen3.5-35B-A3B** model.

The issue occurs especially in **multi-turn conversations** (long coding sessions typical of agents like Claude Code or OpenCode). No visible error in the logs, the conversation is just ended. I need to say continue to restart the token generation.

### To Reproduce
1. Launch oMLX (menu bar app or CLI).
2. Load the model.
3. Connect via the OpenAI-compatible endpoint (`http://localhost:8000/v1`) using Claude or OpenCode.
4. Have a 2-3 turn conversation / tools usage (especially long reasoning/coding).

### Expected behavior
Smooth, continuous generation even on long multi-turn responses, thanks to oMLX’s tiered KV cache SSD.

### Environment
- oMLX version: Latest
- Model: Qwen3.5-35B-A3B-4bit
- Mac: M3 Pro - 36Gb
- Client: Claude Code / OpenCode
- Quantization: 4-bit

Thanks for the amazing tool (the tiered KV cache is awesome for coding agents)!

c4software:
For example :

<img width="1185" height="814" alt="Image" src="https://github.com/user-attachments/assets/cde575d0-ab05-470f-93a1-cf0f840e79dc" />

jundot:
This issue has been reported and discussed multiple times in #205, and several patches have been applied to address it. The core fix should already be included in recent versions.

In some cases though, this is actually a model limitation - certain models struggle to produce proper tool calling output in long context conversations, and that can look identical to this bug from the client side.

To figure out what's going on, i need to see the actual raw model response. Could you share a trace-level server log?

1. Clear the existing log (delete `server.log` or find it at `~/Library/Application Support/oMLX/logs/server.log`)
2. Open the web dashboard → global settings → set log level to **trace** (this records the full raw model response)
3. Reproduce the issue
4. Share the `server.log` file

If your conversation contains anything sensitive (code, credentials, etc.), feel free to redact those parts - i just need the model's raw output around where it stops.

c4software:
Thanks for the quick feedback. I will capture the log tomorrow. Meanwhile, I think I found a possible cause of the problem, if I disable the cache the problem seems not present.

jundot:
@c4software

Your observation about disabling cache was spot on. I dug into it and found what was going on.

[...]

PR #1054, opened 2026-05-04
Fix: honor OpenAI-style stop sequences in server.py

DamnChris:
## Summary

User-provided `stop` strings on the OpenAI-compatible chat completion endpoint (`POST /v1/chat/completions`) are not consistently enforced. The server passes `stop=request.stop` into `engine.stream_chat(...)` (server.py line ~2618 on `main`), but the underlying engine path (notably `omlx/engine/dflash.py`) only halts on EOS tokens for some model + reasoning-parser combinations. The result is runaway generation that ignores the client's stop string until `max_tokens` is hit.

This PR adds a thin server-layer defense: after each token chunk in the chat-completion streaming loop, accumulate the new text and check whether any of the user-supplied stop strings now appears in the running buffer. If so, mark the last output `finished=True`, set `finish_reason="stop"`, log the match, and break the loop.

The check is intentionally a **defense-in-depth** layer, not a replacement for engine-level support — engines that already honor `stop` correctly will short-circuit before this check ever fires.

## Repro

Server: `omlx serve --model-dir <dir>` with a hybrid Qwen 3.5 model (e.g. Qwen3.5-122B-A22B or Qwen3.5-35B-A3B in the VL/text variants). Reasoning parser on. Then:

```bash
curl -s -X POST http://127.0.0.1:8080/v1/chat/completions \
  -d '{
    "model": "<your-model-id>",
    "messages": [
      {"role": "system", "content": "You are concise."},
      {"role": "user", "content": "Count from 1 to 50, one number per line. End with the literal word DONE on its own line."}
    ],
    "stop": ["DONE"],
    "max_tokens": 400,
    "stream": true,
    "stream_options": {"include_usage": true},
    "temperature": 0.0
  }'
```

**Before:** generation continues past `DONE` until `max_tokens` is exhausted; final `finish_reason` is `length`, not `stop`.

**After:** generation halts on the first chunk whose accumulated text contains `DONE`; `finish_reason` is `stop`.

## Fix

19-line addition in `omlx/server.py` inside `stream_chat_completion`, immediately after the existing `accumulated_text += output.new_text` line in the chat completion path. Matches against `accumulated_text`, which already concatenates both `content` and `reasoning_content` deltas — so a stop string that lands in the chain-of-thought stream is caught the same way as one in the visible content.

```python
# Server-layer stop-sequence enforcement
# (some engines do not honor user-provided stop strings, only EOS).
# Match against accumulated_text (content + reasoning_content).
_raw_stop = getattr(request, "stop", None)
_stop_seqs = []
if isinstance(_raw_stop, str):
    _stop_seqs = [_raw_stop]
elif isinstance(_raw_stop, list):
    _stop_seqs = [s for s in _raw_stop if isinstance(s, str)]
_matched_stop = next((s for s in _stop_seqs if s and s in accumulated_text), None)
if _matched_stop is not None and last_output is not None:
    try:
        last_output.finished = True
        last_output.finish_reason = "stop"
    except Exception:
        pass
    logger.info("Stop-sequence matched at server layer: %r", _matched_stop)
    break
```

Behavioural notes:
- Substring match (matches OpenAI semantics — `stop` strings can land mid-token if the tokenization splits them).
- Idempotent: if the engine already finished the request normally, the loop exits before this code re-runs.
- Non-string entries in a `stop` list are silently skipped.
- Logging is `info`-level so the operator can confirm enforcement without enabling debug.

## Measurements (two-host smoke)

Test prompt: 50-line counting prompt with `stop=["DONE"]`, `max_tokens=400`, `temperature=0`, `/no_think` directive. Median of multiple runs per cell.

| Model | Before (median elapsed) | After (median elapsed) | Δ | finish_reason after |
|-------|------------------------:|-----------------------:|------:|---------------------|
| Qwen3.5-122B-A22B-VL (4-bit) | 8.12 s | 4.05 s | -50% | `stop` |
| Qwen3.5-35B-A3B-VL (6-bit)   | 5.41 s | 2.86 s | -47% | `stop` |

Cold-prompt regression check (no stop strings in the prompt, prompt below the prefix-cache threshold): TTFT and decode tok/s within 1% of pre-patch baseline on both models. The added per-chunk substring check is negligible (Python `in` on small text + a short list).

Patched build matches `omlx 0.3.8.dev2` editable install. Patch tested clean against current `main` (HEAD `141f708ebaa35ccc5b7616a2111e0fe9d3a06dfc`).

## Alternatives considered

1. **Fix in `dflash.py`** — push stop-string handling into the engine where it belongs. Lower-risk longer-term, but requires understanding the dflash decode loop and the kv/state machinery in more depth. Happy to follow up with that as a separate PR if the maintainers prefer to land the server-layer fix first as defense-in-depth.
2. **Fix in `omlx/api/adapters/openai.py`** — earlier in the request pipeline. Doesn't help, because the issue is in the streaming decode loop, not in request parsing.
3. **Make the match content-only (skip reasoning_content)** — strictly closer to OpenAI semantics for `stop`. Rejected because operators frequently use stop strings to terminate runaway chain-of-thought (e.g. `"</think>"`), which only ever appears in reasoning_content. Matching the union is the more useful default.

## Open questions for reviewers

- Should the match prefer `output.new_text` (last chunk) plus a small lookback window instead of the full `accumulated_text`? Current implementation is O(n × k) per chunk where n is text length and k is number of stop strings; for very long generations this is wasted work after the first non-match. Easy follow-up if it shows up in profiling.
- Worth emitting a diagnostic header (`x-omlx-stop-source: server-layer`) so callers can tell which layer halted them? Not in this PR; happy to add.
- Anthropic adapter has a similar `stop_sequences` plumbing (`omlx/api/adapters/anthropic.py`) — out of scope for this PR but likely the same gap; can file as a separate issue.

## Existing issue

A targeted search of `jundot/omlx` issues (`stop sequence`, `stop string`, `stop ignored`, `runaway generation`, `stop_sequences`, `stop parameter`) did not surface a matching open or closed report. The closest neighbours (#260, #484, #352) describe the *opposite* failure mode (generation halting unexpectedly), and #804 covers client-disconnect cancellation, which is unrelated. This PR therefore does not reference an existing issue; happy to file one alongside if the maintainers prefer that workflow.

## Files touched

- `omlx/server.py` — single 19-line addition inside `stream_chat_completion`, after `accumulated_text += output.new_text` in the chat completion path. No deletions, no other functions touched.
- `tests/integration/test_e2e_streaming.py` — three tests added inside `TestStreamingHelperFunctions` (see "Tests included" below).

## Tests included

Three async tests added to `tests/integration/test_e2e_streaming.py::TestStreamingHelperFunctions`, all using the existing `MockBaseEngine` pattern (no model load required):

- `test_stream_chat_completion_honors_stop_at_server_layer` — list-form `stop=["STOP_HERE"]` with simulated runaway outputs from the engine; asserts the chunks emitted past the matched stop string do not leak into the streamed response.
- `test_stream_chat_completion_honors_string_form_stop` — single-string `stop="END"` with simulated runaway; same assertion.
- `test_stream_chat_completion_no_stop_param_unchanged` — regression guard: when no `stop` is supplied, behaviour is identical to before the patch.

All three pass with the patch applied; the first two fail without it (regression-guard test passes either way, by design). Full suite (`pytest tests/integration/test_e2e_streaming.py`): 28 passed, 0 failed.

## References

- Patched-against upstream commit: `141f708ebaa35ccc5b7616a2111e0fe9d3a06dfc`
- Related upstream code:
  - `omlx/server.py:1967` and `omlx/server.py:2618` — current stop wiring (engine-layer only)
  - `omlx/api/adapters/openai.py` — request-time normalization of `stop`
  - `omlx/api/adapters/anthropic.py` — sibling Anthropic-style `stop_sequences` plumbing (out of scope)
  - `omlx/engine/dflash.py` — engine path where stop is currently observed not to fire

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -2743,6 +2743,25 @@ async def stream_chat_completion(
             last_output = output
             if output.new_text:
                 accumulated_text += output.new_text
+            # Server-layer stop-sequence enforcement
+            # (some engines do not honor user-provided stop strings, only EOS).
+            # Match against accumulated_text (content + reasoning_content).
+            _raw_stop = getattr(request, "stop", None)
+            _stop_seqs = []
+            if isinstance(_raw_stop, str):
+                _stop_seqs = [_raw_stop]
+            elif isinstance(_raw_stop, list):
+                _stop_seqs = [s for s in _raw_stop if isinstance(s, str)]
+            _matched_stop = next((s for s in _stop_seqs if s and s in accumulated_text), None)
+            if _matched_stop is not None and last_output is not None:
+                try:
+                    last_output.finished = True
+                    last_output.finish_reason = "stop"
+                except Exception:
+                    pass
+                logger.info("Stop-sequence matched at server layer: %r", _matched_stop)
+                break
+
 
             if stream_content and output.new_text:
                 thinking_delta, content_delta = thinking_parser.feed(output.new_text)
--- a/tests/integration/test_e2e_streaming.py
+++ b/tests/integration/test_e2e_streaming.py
@@ -2583,6 +2583,138 @@ class TestStreamingHelperFunctions:
         assert tool_call_deltas == []
         assert "stop" in finish_reasons
 
+    @pytest.mark.asyncio
+    async def test_stream_chat_completion_honors_stop_at_server_layer(self):
+        """Server-layer stop-sequence enforcement halts streaming when the stop
+        string appears in accumulated text.
+
+        Defends against engines that accept ``stop=`` but do not consistently
+        honor it (notably some dflash + reasoning-parser combinations). The
+        check matches against accumulated_text (content + reasoning_content)
+        and breaks the streaming loop, preventing emission of subsequent chunks.
+        """
+        from omlx.server import stream_chat_completion
+        from omlx.api.openai_models import ChatCompletionRequest, Message
+
+        engine = MockBaseEngine()
+        engine.set_stream_outputs([
+            MockGenerationOutput(text="ok", new_text="ok", finished=False),
+            MockGenerationOutput(
+                text="ok STOP_HERE", new_text=" STOP_HERE", finished=False
+            ),
+            MockGenerationOutput(
+                text="ok STOP_HERE LEAK", new_text=" LEAK", finished=False
+            ),
+            MockGenerationOutput(
+                text="ok STOP_HERE LEAK done",
+                new_text=" done",
+                finished=True,
+                finish_reason="length",
+            ),
+        ])
+
+        request = ChatCompletionRequest(
+            model="test-model",
+            messages=[Message(role="user", content="Test")],
+            stop=["STOP_HERE"],
+            stream=True,
+        )
+
+        events = []
+        messages = [{"role": "user", "content": "Test"}]
+        async for event in stream_chat_completion(
+            engine, messages, request,
+            max_tokens=256, temperature=0.7, top_p=0.9, top_k=40,
+        ):
+            events.append(event)
+
+        streamed = "\n".join(events)
+        assert "LEAK" not in streamed, (
+            "Server-layer stop did not fire; later chunks leaked into stream"
+        )
+        # Sanity: pre-stop content emitted normally before the break.
+        assert "ok" in streamed
+
+    @pytest.mark.asyncio
+    async def test_stream_chat_completion_honors_string_form_stop(self):
+        """Single-string ``stop`` (not list) is also enforced at the server layer."""
+        from omlx.server import stream_chat_completion
+        from omlx.api.openai_models import ChatCompletionRequest, Message
+
+        engine = MockBaseEngine()
+        engine.set_stream_outputs([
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for the detailed repro. Tracing it from the top, the runaway-generation cause sits upstream of the engine. In `server.py` around line 2266 the OpenAI chat path never adds `stop` to `chat_kwargs` at all, even though the Anthropic path (line 3532) and `/v1/completions` (line 1981/2641) both do. The second half is in `scheduler._build_state_machine`, which today only feeds stop *token IDs* into mlx-lm's BatchGenerator state machine, so `sampling_params.stop` (the string list) is never consumed by anything in production code.

I just landed those two fixes plus a text-level fallback for BPE edge cases (where `"delta"` tokenizes differently than `" delta"`). Smoke-tested on Qwen3-4B with `stop=["delta"]` and the response now trims to `"alpha beta gamma"` on both `/v1/chat/completions` and `/v1/messages`. Closing this in favor of the upstream fix in the next release.

PR #1088, opened 2026-05-06
Re-check for updates from periodic health timer

jprado:
Hooks `_check_for_updates()` into the existing `healthCheck_` timer so the app passively checks for new releases in the background. The check self-gates on a 24h timestamp cache, so it only hits the network once per day regardless of how frequently the timer fires.


jundot:
Thanks! This fills a real gap. The app stays running for days, and a one-shot check at launch means most users never see updates until a restart. The 24h self-gate keeps the extra cost basically free.

Merging now, will be in the next release.

Issue #1094, opened 2026-05-06
oQ quantization is way too memory hungry after v0.3.9


deepsweet:
<img width="1006" height="454" alt="Image" src="https://github.com/user-attachments/assets/5e66b0e2-21a8-4175-8cf8-deba6142fdc3" />

```
2026-05-06 20:31:52,128 - omlx.admin.oq_manager - INFO - oQ quantization queued: Qwen3.6-35B-A3B -> oQ5 (task_id=8989eaa5-c491-47c7-843f-583f1811dfad)
2026-05-06 20:31:52,140 - omlx.oq - INFO - oQ5 streaming: 1045 tensors in 26 shards
2026-05-06 20:31:52,159 - omlx.patches.mlx_vlm_mtp.qwen35_vlm_model - INFO - Patched mlx_vlm.models.qwen3_5.Model.sanitize for MTP
2026-05-06 20:31:52,164 - omlx.patches.mlx_vlm_mtp - INFO - mlx-vlm MTP sanitize patch applied
2026-05-06 20:31:52,167 - omlx.oq - INFO - Using mlx-vlm full sanitize chain for Model (preserves vision weights)
2026-05-06 20:31:52,169 - omlx.oq - WARNING - Streaming discovery failed ('_TrackedTensor' object has no attribute 'moveaxis'), falling back to eager sanitize
2026-05-06 20:32:11,340 - omlx.process_memory_enforcer - WARNING - Process memory limit exceeded: 60.3GB / 60.2GB (over by 0.2GB)
2026-05-06 20:32:11,342 - omlx.process_memory_enforcer - WARNING - Process memory limit exceeded but no models are loaded to evict.
2026-05-06 20:32:12,344 - omlx.process_memory_enforcer - WARNING - Process memory limit exceeded: 62.9GB / 60.2GB (over by 2.7GB)
2026-05-06 20:32:12,344 - omlx.process_memory_enforcer - WARNING - Process memory limit exceeded but no models are loaded to evict.
2026-05-06 20:32:13,345 - omlx.process_memory_enforcer - WARNING - Process memory limit exceeded: 64.9GB / 60.2GB (over by 4.7GB)
2026-05-06 20:32:13,345 - omlx.process_memory_enforcer - WARNING - Process memory limit exceeded but no models are loaded to evict.
2026-05-06 20:32:14,105 - omlx.oq - INFO - oQ5: eager sanitize applied, 1066 tensors
[...]

PR #1095, opened 2026-05-06
fix falling back to eager sanitize (to save Quantized Memory)

felk-dev:
Eager sanitize is burdensome because it puts the entire quantization target model into memory. This should kill omlx server and quantization fail. So, I added code to address this issue.

------- fix under error ----------
#omlx.oq - WARNING - Streaming discovery failed ('_TrackedTensor' object has no attribute 'moveaxis'), falling back to eager sanitize
    #omlx.oq - WARNING - Streaming discovery failed ('_TrackedTensor' object has no attribute 'transpose'), falling back to eager


deepsweet:
Solves my issue https://github.com/jundot/omlx/issues/1094?

felk-dev:
image

I did not select a sensitivity model.
I'm having a problem(proxy sensitivity is all nan.) after select a sensitivity model and trying it.
I think it's not an exact solution.
But if you don't select a sensitivity model, the quantization works well. I think it can be used as a temporary measure.

image

Issue #1070, opened 2026-05-05
Cache doesn't store properly


RoelVideler:
After running for an afternoon the cache for the model was still zero, even though a lot of tokens got reused. 

**To Reproduce**
1. I was running Qwen3.6 35B-A4B oQ8, but I've had the same issue with 27b
2. Generating code with OpenCode and OpenSpec for hours and hours

**Desktop (please complete the following information):**
 - macOS Version: 26.4.1 
 - oMLX Version 0.3.8, dmg or brew version appear identical

**Additional context**
There's a lot of errors regarding the ssd cache in the logs, especially around the eviction. I suspect the cache needed to be written then, didn't but did register that they should have been. Personally I would prefer the cache to write after either every processed pre-prompt or even better, after every batch. And as a sidenote I'd love to be able to make the batches smaller, the memory overhead from 2048 token processing is likely smashing through the limit. 

Hot cache is set to roughly 4gb and ssd cache to 186gb. 

Also.. Love your work Jundot! I hope this helps. 

`2026-05-05 13:58:26,457 - omlx.admin.routes - INFO - [-] - Max model memory changed: 50.40GB -> 50.40GB
2026-05-05 13:58:26,457 - omlx.process_memory_enforcer - INFO - [-] - Process memory limit changed: 58.9GB -> 58.9GB
2026-05-05 13:58:26,457 - omlx.admin.routes - INFO - [-] - Process memory limit updated to 58.9GB
2026-05-05 13:58:26,457 - omlx.process_memory_enforcer - INFO - [-] - Prefill memory guard: enabled
2026-05-05 13:58:26,457 - omlx.admin.routes - INFO - [-] - Prefill memory guard enabled
2026-05-05 13:58:26,457 - omlx.admin.routes - INFO - [-] - Hot cache max size changed: 10.00 GB -> 4.00 GB
2026-05-05 13:58:26,457 - omlx.engine_pool - INFO - [-] - Unloading model: Qwen3.6-35B-A3B-MLX-VL-oQ4-FP16 (immediate abort)
2026-05-05 13:58:26,457 - omlx.engine_core - INFO - [-] - Engine stopped
2026-05-05 13:58:26,457 - omlx.scheduler - INFO - [-] - Scheduler shutdown initiated...
2026-05-05 13:58:26,457 - omlx.cache.paged_ssd_cache - INFO - [-] - Shutting down PagedSSDCacheManager...
[...]

PR #1101, opened 2026-05-07
fix(cache): use blocking put during shutdown flush to prevent dropped blocks

ivaniguarans:
## Summary

When `PagedSSDCacheManager.close()` flushes hot cache entries to SSD, it iterates all entries and calls `_enqueue_ssd_write()` which uses `put_nowait()` on the bounded write queue. The flush loop runs in microseconds while the writer thread drains at NVMe speed (~1ms/block), so the queue fills almost immediately and subsequent blocks are silently dropped.

On a 128GB machine (`_MAX_PENDING_WRITES=64`), flushing ~111 hot cache blocks drops ~46 of them — the most recently active prefixes — causing cold misses on the next startup.

## Changes

- Add a `blocking` keyword argument to `_enqueue_ssd_write()`. When `True`, uses `put(timeout=0.5)` instead of `put_nowait()`, giving the writer thread time to drain between submissions.
- `close()` passes `blocking=True`; all other callers remain non-blocking (inference-time eviction can still drop a single block without harm).
- `close()` checks `_writer_thread.is_alive()` before each blocking put to bail early if the consumer is dead, avoiding a stall on a dead thread.
- Dropped blocks during shutdown are now counted and logged as a warning.

## Test plan

- [x] Added regression test `test_close_flushes_all_blocks_with_small_queue`: patches `_MAX_PENDING_WRITES` to 4, creates 12 hot cache blocks (3x queue depth), calls `close()`, and verifies all 12 are persisted to SSD.
- [x] All 95 existing tests pass (`uv run pytest tests/ -x`).
- [x] Verified on 128GB M-series Mac with production-sized hot cache (~111 blocks): 0 dropped blocks after fix vs. ~46 dropped before.

Related: #1070

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -745,11 +745,17 @@ class PagedSSDCacheManager(CacheManager):
         for evicted_hash, evicted in evicted_entries:
             self._enqueue_ssd_write(evicted_hash, evicted)
 
-    def _enqueue_ssd_write(self, block_hash: bytes, entry: dict) -> bool:
+    def _enqueue_ssd_write(
+        self, block_hash: bytes, entry: dict, *, blocking: bool = False,
+    ) -> bool:
         """Enqueue a hot cache entry for SSD background write.
 
         Used when evicting from hot cache or flushing on shutdown.
         Adds block to SSD index before enqueueing write.
+
+        When *blocking* is True, waits briefly for queue space instead of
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, this is a clean catch. The shutdown flush hitting put_nowait against a queue that fills faster than the writer drains lines up exactly with what #1070 is seeing on long sessions. The regression test (queue depth 4 vs 12 blocks) pins the failure mode well. Merging — will land in the next release.

Issue #1109, opened 2026-05-07
DFlash start failed for Qwen3.6-35B-A3B-4bit: 'language_model.mtp.layers.0.mlp.experts.0.gate_proj.weight'


ilterpehlivan:
**Describe the bug**
2026-05-07 15:31:30,180 - omlx.engine_pool - INFO - Loading model: Qwen3.6-35B-A3B-4bit
2026-05-07 15:31:30,180 - omlx.engine_pool - INFO - DFlash enabled for Qwen3.6-35B-A3B-4bit, draft=/USER/.cache/huggingface/hub/Qwen3.6-35B-A3B-DFlash
2026-05-07 15:31:30,209 - omlx.engine_pool - WARNING - DFlash start failed for Qwen3.6-35B-A3B-4bit: 'language_model.mtp.layers.0.mlp.experts.0.gate_proj.weight'. Falling back to vlm engine.
2026-05-07 15:31:30,210 - omlx.engine.dflash - INFO - DFlashEngine stopped

Using: z-lab/Qwen3.6-35B-A3B-DFlash

**Expected behavior**
Dflash engine works

**Desktop (please complete the following information):**
 - oMLX Version [Pulled latest from repo]

zedbytes:
can confirm this error also occurs for other quants like oQ8 


```
2026-05-08 03:49:08,350 - omlx.engine_pool - INFO - Loading model as LM (force_lm=True): Qwen3.6-35B-A3B-oQ8
2026-05-08 03:49:08,350 - omlx.engine_pool - INFO - DFlash enabled for Qwen3.6-35B-A3B-oQ8, draft=/Users/omlx/models/antonperk/Qwen3.6-35B-A3B-DFlash-oQ8
2026-05-08 03:49:08,380 - omlx.engine_pool - WARNING - DFlash start failed for Qwen3.6-35B-A3B-oQ8: 'language_model.mtp.layers.0.mlp.experts.0.gate_proj.weight'. Falling back to batched engine.
2026-05-08 03:49:08,461 - omlx.admin.benchmark - ERROR - Benchmark error: 'language_model.mtp.layers.0.mlp.experts.0.gate_proj.weight'
Traceback (most recent call last):
  File "/Users/omlx/code/mlx/.venv/lib/python3.13/site-packages/omlx/engine_pool.py", line 699, in _load_engine
    await engine.start()
[...]

Issue #1099, opened 2026-05-07
Vision capability of vlm not working on Qwen3.6-27B-oQ8-mtp (0.3.9dev1)


kuanjames:
Using jundot/Qwen3.6-27B-oQ8-mtp on Huggingface and 0.3.9dev1, there is no vision capability in omlx chat or as an API, even when I choose auto or vlm.

This is not just the MTP model, when using an older LLM of the 27B without MTP, the vision ability has disappeared.

RoelVideler:
Do you have specprefill or dflash enabled? They break vision capabilities.

diedcow1:
TheCluster:Qwen3.5-9B-Uncensored-HauhauCS-Aggressive-MLX-mxfp8 is recognized as vlm by 3.9 but as llm by 3.12.

PR #1112, opened 2026-05-07
fix(oq): pass text_only to _build_model_sanitizer to preserve MTP tensors

ivaniguarans:
## Summary

When `text_only=True` is passed to `quantize_oq_streaming`, the caller wants a text-only quantization — no vision weights, just the language model. However, `_build_model_sanitizer` never received this flag and unconditionally chose the VLM sanitize path for any model whose config lists a `ForConditionalGeneration` architecture.

This affects all Qwen3.5 and Qwen3.6 checkpoints, which register `Qwen2_5_VLForConditionalGeneration` even though the underlying text model has MTP heads.

The VLM sanitize path uses mlx-vlm's `_Proxy`-based model loading, which never attaches `self.mtp`. As a result, the sanitize function silently drops every `mtp.*` tensor from the output — even when `preserve_mtp=True`. The quantized model ships with 0 MTP tensors and falls back to single-token decoding at inference time.

## Root cause

`_build_model_sanitizer(config)` checked `config["architectures"]` for `"ForConditionalGeneration"` but had no knowledge of whether the caller intended a text-only quantization. The `text_only` parameter was only used later in the tensor write loop to skip vision weights — by that point, MTP tensors had already been stripped by the wrong sanitizer.

## Changes

- Add a `text_only: bool = False` parameter to `_build_model_sanitizer`.
- When `text_only` is `True`, skip VLM detection so the LLM sanitize path is always used. The `mlx_lm_mtp` patches handle MTP sanitize correctly for both dense and MoE architectures.
- Thread `text_only` from `quantize_oq_streaming` through to the call site.

## Reproduction

1. Start oMLX with any Qwen3.5 or Qwen3.6 model.
2. Run oQ quantization via the admin API with `preserve_mtp: true` and `text_only: true`.
3. Check the output for `mtp.*` tensors.

**Before fix:** 0 MTP tensors in output.
**After fix:** 39 MTP tensors preserved (verified on Qwen3.6-35B-A3B oQ8).

## Test plan

- [x] Verified on Qwen3.6-35B-A3B: oQ8 with `text_only=true` and `preserve_mtp=true` produces 39 MTP tensors in output (0 before fix).
- [x] All existing tests pass (`uv run pytest tests/ -x` — 1 pre-existing failure in `test_response_endpoint_recovers_tool_call_from_thinking` unrelated to this change).

Related: #1109, #1099, #1094, #1092

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -1505,18 +1505,22 @@ def _should_quantize_tensor(name: str, shape: tuple) -> bool:
     return True
 
 
-def _build_model_sanitizer(config: dict):
+def _build_model_sanitizer(config: dict, text_only: bool = False):
     """Build a sanitize function from the model class.
 
     For VLM models, uses mlx-vlm's model class (preserves vision weights).
     For LLM models, uses mlx-lm's model class.
+    When text_only is True, always uses the LLM path even for VLM
+    architectures so that mlx_lm_mtp patches (which handle MTP sanitize
+    for both dense and MoE) are used instead of the VLM path whose
+    _Proxy-based sanitize drops the MTP head.
 
     Returns:
         A function that takes a dict of weights and returns sanitized weights,
         or None if the model class can't be loaded.
     """
     architectures = config.get("architectures", [])
-    is_vlm = any("ForConditionalGeneration" in a for a in architectures)
+    is_vlm = any("ForConditionalGeneration" in a for a in architectures) and not text_only
 
     if is_vlm:
         try:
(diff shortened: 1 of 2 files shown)


deepsweet:
Does not fix my issue because it's actually not related to either "Preserve MTP weights" or "Text Only" at all, see https://github.com/jundot/omlx/issues/1094#issuecomment-4405293285

Just a general 0.3.9+ memory problem.

ivaniguarans:
@deepsweet please check https://github.com/jundot/omlx/pull/1126, it might help.

jundot:
Thanks for tracking this down. The root cause framing is exactly right. The VLM sanitize path was silently dropping `mtp.*` before the tensor loop ever ran, so any downstream MTP-skip would have been chasing the symptom. Routing `text_only=True` to the LLM path with the mlx_lm_mtp patches is the clean fix. Merged, will ship in the next release.

PR #1114, opened 2026-05-07
fix(app): persist DMG model directory changes

gltanaka:
## Summary
- keep the macOS DMG app model directory preference in sync with the server settings file
- update a running server through the admin settings API so runtime state and disk state match
- preserve existing web-admin multi-directory settings during conservative first-run seeding, but overwrite stale values on explicit Preferences save

## Tests
- `uv run python -m pytest tests/test_omlx_app.py`
- `uv run python -m ruff check --select F841,I001 packaging/omlx_app/config.py tests/test_omlx_app.py`

### DIFF
--- a/packaging/omlx_app/config.py
+++ b/packaging/omlx_app/config.py
@@ -2,7 +2,7 @@
 
 import json
 import logging
-from dataclasses import dataclass, asdict
+from dataclasses import asdict, dataclass
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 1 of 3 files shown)


jundot:
Thanks, looks good — merging. I'll follow up in the next release with a small guard so the runtime sync only fires when the model dir actually changed. Otherwise web-admin multi-dir setups can collapse to a single entry when a user opens Preferences for an unrelated change (API key, port) and clicks Save.

PR #1119, opened 2026-05-08
fix(engine): prevent partial mode cache invalidation

blightbow:
The original partial mode PR that I submitted had a bug that corrupts the prefill cache. This happens due to mutation of the prompt across requests and isn't caught by the standard cache tests. The bug causes divergent output across multiple prompt submissions that contain the same prefix. It's a subtle corruption whose impact will vary depending on the prompt, but corruption is corruption.

This PR includes the fix along with regression tests for all three engines that leverage `_apply_chat_template` and `count_chat_tokens`:
- BatchedEngine
- VLMBatchedEngine
- DFlashEngine

VLMBatchedEngine gained a no-op `is_partial` parameter for consistency with the other two engines. The logic could be rewritten to use an `isinstance` check if you prefer, but API consistency felt cleaner at the time of writing.

### DIFF
--- a/omlx/api/utils.py
+++ b/omlx/api/utils.py
@@ -406,6 +406,8 @@ def extract_text_content(
                 msg_dict["reasoning_content"] = reasoning_out
             if getattr(msg, "name", None):
                 msg_dict["name"] = msg.name
+            if getattr(msg, "partial", False):
+                msg_dict["partial"] = True
 
             # Preserve structured tool_calls for models with native tool calling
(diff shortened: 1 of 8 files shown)


jundot:
Thanks for tracking this down — the in-place mutation across `count_chat_tokens` and `chat` was a subtle one. Hoisting the strip to the API boundary and forwarding `is_partial` explicitly is the right call. Merging for the next release.

PR #1120, opened 2026-05-08
feat(dflash): add configurable prefix cache max_entries setting

yilmazorhan:
## Summary

The DFlash prefix cache `max_entries` was previously hardcoded to the dflash-mlx "balanced" profile default of **4**, with no way to override it from oMLX. This meant users couldn't increase the number of cached prefix snapshots even when they had sufficient memory budget.

This PR adds a new per-model setting `dflash_in_memory_cache_max_entries` (default: **4**, matching the dflash-mlx balanced profile) that is passed directly to `runtime_config_from_profile(prefix_cache_max_entries=...)`, giving users explicit control over how many prefix snapshots are retained in L1.

## Changes

| File | Change |
|------|--------|
| `omlx/model_settings.py` | Add `dflash_in_memory_cache_max_entries: int = 4` field |
| `omlx/engine/dflash.py` | Read the setting and pass `prefix_cache_max_entries` to `runtime_config_from_profile()` |
| `omlx/admin/routes.py` | Expose in API request model, response payload, update handler, and reload trigger |
| `omlx/admin/static/js/dashboard.js` | Wire up JS model init, save payload, and reset defaults |
| `omlx/admin/templates/dashboard/_modal_model_settings.html` | Add number input field in the DFlash cache section |

## Behavior

- Default is 4 entries (matches dflash-mlx balanced profile)
- Configurable per-model via admin API (`PUT /admin/api/models/{id}/settings`) or the dashboard UI
- Changing the value on a loaded model triggers automatic unload + reload
- Setting to 0 or null resets to default (4)

## UI

A new "In-memory cache max entries" input appears in the DFlash section of the model settings modal (visible when in-memory cache is enabled), right above the existing cache size (GiB) field.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -132,6 +132,7 @@ class ModelSettingsRequest(BaseModel):
     dflash_draft_quant_bits: Optional[int] = None
     dflash_max_ctx: Optional[int] = None
     dflash_in_memory_cache: Optional[bool] = None
+    dflash_in_memory_cache_max_entries: Optional[int] = None
     dflash_in_memory_cache_max_bytes: Optional[int] = None
     dflash_ssd_cache: Optional[bool] = None
     # Native MTP (mlx-lm PR 990 / PR 15 monkey-patch)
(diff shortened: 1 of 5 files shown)


DiegoRBaquero (approved):

jundot (approved):
Thanks for the patch. Symmetry with `max_bytes` makes sense, and the wiring (settings → engine → admin API/UI) follows the existing dflash pattern cleanly. Default of 4 keeps current behavior and the reload trigger is in place, so regression risk is essentially nil. Going to merge as-is, small internal cleanup will land in the next release.

yilmazorhan:
Thanks for omlx project. Increasing bucket size has a great impact on token generation. I got 2x-3x better results, especially for dens models.

PR #1125, opened 2026-05-08
Reduce ram usage in quantization / Solve mtp layers remove after quantization.

felk-dev:
**1. Solve falling back to eager sanitize error (This method reduce ram usage in quantization)**

#1094
WARNING - Streaming discovery failed ('_TrackedTensor' object has no attribute 'moveaxis')
Add [def transpose] and [def moveaxis] in line 842~855

**2. Solve mtp layers remove after quantization.**

```
if not preserve_mtp and (".mtp." in tensor_name or tensor_name.startswith("mtp.")):
            all_weights.pop(tensor_name, None)
            continue
```
this make mtp layers save.

**3. Edit [def measure_sensitivity_from_quantized_model]**

in terminal after measure sensitivity to use quantized model, All score Nan.
Remove the mode = mode factor (Delegate MLX to automatically infer based on the presence or absence of bias)
           ```
 w_float = mx.dequantize(
                m.weight, m.scales, getattr(m, "biases", None),
                group_size=gs, bits=bits
```
            )
Immediately evaluate the modified weights to prevent Lazy Evaluation errors
          ```
  if m.biases is not None:
                mx.eval(m.weight, m.scales, m.biases)
            else:
                mx.eval(m.weight, m.scales)
```

```
if out_perturbed is not None:
            # [Correct 5] Float16 Casting as float32 to prevent overflow]
            ob32 = out_baseline.astype(mx.float32)
            op32 = out_perturbed.astype(mx.float32)
            
            raw_mse = ((ob32 - op32) ** 2).mean()
            out_mag = (ob32 ** 2).mean()
            mse_val = raw_mse / mx.maximum(out_mag, 1e-10)
```

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -2171,11 +2171,9 @@ def quantize_oq_streaming(
 
     cb("loading", 20.0)
 
-    # --- [추가된 부분 1: config에서 MTP 활성화 여부 확인] ---
     mtp_enabled = config.get("num_mtp_layers", 0) > 0
     if not mtp_enabled and "text_config" in config:
         mtp_enabled = config["text_config"].get("num_mtp_layers", 0) > 0
-    # --------------------------------------------------------
 
     tensor_names = list(all_weights.keys())
     total_tensors = len(tensor_names)
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -2171,10 +2171,6 @@ def quantize_oq_streaming(
 
     cb("loading", 20.0)
 
-    mtp_enabled = config.get("num_mtp_layers", 0) > 0
-    if not mtp_enabled and "text_config" in config:
-        mtp_enabled = config["text_config"].get("num_mtp_layers", 0) > 0
-
     tensor_names = list(all_weights.keys())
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for digging into this and sending the patches. Here is how I ended up routing the changes.

The moveaxis/transpose part is now covered by #1126 with axis-aware transform names that the discovery materializer needs to replay correctly, so I went with that version. The MTP cases trace back to the VLM sanitize path stripping `mtp.*` before this loop runs, and #1112 fixes that at the source — so the per-tensor skip block and the `strict=False` patch in `_measure_sensitivity` aren't needed (and `strict=False` would also silence future genuine weight mismatches, which I'd rather avoid).

The sensitivity NaN fix in `_measure_sensitivity_from_quantized_model` (mode arg removal, mx.eval on baseline + re-quantized weights, float32 MSE cast, biases cleanup on restore) is a real bug the other PRs don't cover. I landed it directly on main as 966908c with you as the commit author so the credit shows up in your contribution graph. It will ship in the next release.

Closing this for now. Thanks again for the analysis.

Issue #956, opened 2026-04-25
无法加载 DeepSeek-V4-Flash 模型 - unsupported dtype F8_E8M0 和 model type deepseek_v4 not supported


wgetnz:
## 问题描述
在尝试加载 DeepSeek-V4-Flash 模型时遇到错误，以下是日志详情：

```
2026-04-25 17:05:49,143 - omlx.engine_pool - INFO - [-] - Loading model: DeepSeek-V4-Flash
2026-04-25 17:05:49,155 - omlx.server - WARNING - [-] - POST /admin/api/models/DeepSeek-V4-Flash/load → 500: [safetensor] unsupported dtype F8_E8M0
2026-04-25 17:05:52,471 - omlx.engine_pool - INFO - [-] - Loading model: deepseek-ai-DeepSeek-V4-Flash-8bit
2026-04-25 17:05:52,498 - omlx.server - WARNING - [-] - POST /admin/api/models/deepseek-ai-DeepSeek-V4-Flash-8bit/load → 500: Model type deepseek_v4 not supported.
2026-04-25 17:06:17,308 - omlx.model_settings - INFO - [-] - Updated settings for model 'deepseek-ai-DeepSeek-V4-Flash-8bit'
2026-04-25 17:06:20,221 - omlx.engine_pool - INFO - [-] - Loading model: deepseek-ai-DeepSeek-V4-Flash-8bit
2026-04-25 17:06:20,250 - omlx.server - WARNING - [-] - POST /admin/api/models/deepseek-ai-DeepSeek-V4-Flash-8bit/load → 500: Model type deepseek_v4 not supported.
2026-04-25 17:06:27,305 - omlx.engine_pool - INFO - [-] - Loading model: DeepSeek-V4-Flash
2026-04-25 17:06:27,316 - omlx.server - WARNING - [-] - POST /admin/api/models/DeepSeek-V4-Flash/load → 500: [safetensor] unsupported dtype F8_E8M0
```

## 错误信息
1. `[safetensor] unsupported dtype F8_E8M0` - 不支持的数据类型 F8_E8M0
2. `Model type deepseek_v4 not supported` - 不支持 deepseek_v4 模型类型

TelenLiu:
deepseek 要是能做到只加载激活部分到内存，其他部分放在硬盘，小内存Macos也可以跑deepseek v4了。

hwang63130:
same issue here. "Model type deepseek_v4 not supported"

sjtuck:
[...]

Issue #777, opened 2026-04-15
[Feature Request] Optimize oQ / oQ+ Quantization Workflow for Limited Memory


sewall2025:
Description:
Currently, when performing oQ or oQ+ quantization in Omlx, the system requires the complete model weights to be loaded into the system memory (RAM) all at once.

This creates a frustrating paradox for users: the primary reason users seek out quantization is precisely because they lack sufficient memory. Users who have enough RAM to load the full model don't strictly need oQ quantization, whereas the target audience—users with limited memory—are fundamentally blocked from performing the quantization process.

Suggestion:
I highly recommend modifying the quantization workflow to allow it to run on machines with limited memory. Similar to how other tools (like llama.cpp) handle this, Omlx could implement a streaming or tensor-by-tensor processing approach. By reading, quantizing, and saving the model weights sequentially rather than loading the entire model into RAM simultaneously, users with lower-end hardware would finally be able to execute oQ and oQ+ quantization successfully.

Thank you for considering this improvement!

MattBash17:
@jundot this would be a game changer for many of us, especially now that oQe quantization is available. Please consider having a look at this, and again, you're doing an amazing work with this app 🫶

PR #1126, opened 2026-05-08
feat(oq): on-the-fly FP8/I8 dequant in _LazyTensorIndex + OOM guard

ivaniguarans:
## Summary

- **On-the-fly FP8 dequant**: `_LazyTensorIndex` detects FP8/I8 weight+scale pairs at index construction (validated by weight dtype) and transparently dequants to BF16 when accessed. Scale keys hidden from the public API. Zero scratch disk.
- **Two scale conventions**: vLLM (`_scale_inv`, F32, 128-block) and MXFP (`key.scale`, E8M0 microscaling, variable block).
- **OOM guard**: When model size exceeds 80% of system RAM, skips eager sanitize fallback (with error-level logging) and sensitivity measurement.
- **`_TrackedTensor` improvements**: `moveaxis`, `transpose`, `_slice_length`, `_detect_half_split` for reliable sanitize plan discovery on conv1d and GQA patterns.
- **Replayable-transform validation**: Discovery rejects non-replayable transforms with a clear error instead of silently producing wrong weights.
- **Shared dequant helper**: `_block_dequant_fp8()` consolidates block-scaled dequant logic with shape divisibility and degenerate scale guards.
- **Removed dead code**: `_streaming_fp8_dequant` (no longer called after on-the-fly dequant replaced it).
- **Bug fix**: `_LazyTensorIndex.get()` now correctly returns override-only keys instead of `default`.

## Motivation

Related to #1094 — reduces peak memory during quantization by fixing `_TrackedTensor.moveaxis` (so discovery succeeds instead of falling back to eager sanitize) and adding an OOM guard for models exceeding system RAM. Does not address the underlying regression reported in that issue.

Related to #777 (streaming quant for limited memory) and #956 (E8M0 dtype support).

## Test plan

- [x] 23 new pytest tests (174 total, up from 151 upstream) covering both scale conventions, FP8-aware plan discovery, logical metadata, mixed tensors, I8+MXFP, simulated low-RAM paths, TrackedTensor shape tracking, and BF16+scale dtype guard
- [x] Live integration: Qwen3.6-27B oQ4 + oQ8, Qwen3.6-35B-A3B oQ8 with simulated 48GB RAM, Qwen3.5-122B-A10B oQ4 + oQ6 (real-world — model exceeds 128GB RAM unquantized) — all produce correct inference
- [x] DeepSeek V4 Flash oQ2: quantizes successfully (zero scratch), load fails due to upstream model class limitation (not this PR's scope)

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -33,6 +33,8 @@ OQ_DTYPES: tuple[str, ...] = ("bfloat16", "float16")
 
 _OQ_DEFAULT_GROUP_SIZE = 64
 
+_MAX_MODEL_RAM_FRACTION = 0.8
+
 _LEVEL_BITS: dict[float, int] = {2: 2, 3: 3, 3.5: 3, 4: 4, 5: 5, 6: 6, 8: 8}
 
 _LEVEL_PROTECTION: dict[float, str] = {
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for this. The on-the-fly dequant in `_LazyTensorIndex` is a much better shape than the disk-spill streaming path, and the replayable-transform validation plus axis-aware moveaxis/transpose names give the discovery materializer a firm foundation for non-Qwen FP8 sources. The 80% RAM guard is also a nice safety net for the large-model case.

Heads-up: I rebased the branch onto main to resolve a one-line conflict introduced by #1112 (the `_build_model_sanitizer` call now takes `text_only=text_only`) and force-pushed back to your fork. Your original commit author is preserved on the rebased commit. Merged, will ship in the next release.

PR #1136, opened 2026-05-09
feat(oq): auto-build proxy for sensitivity when model exceeds RAM

jundot:
## Summary

When a model is larger than 80% of system RAM, full-fp16 sensitivity measurement is not feasible, and the previous behavior was to silently fall back to a position-based heuristic (a hardcoded U-shape over layer indices). That made oQ-on-large-models look like real oQ from the user's side but actually shipped uniform-ish quant with a position prior.

This PR makes the OOM-guard branch auto-build a temporary uniform 4-bit proxy on disk and run `_measure_sensitivity_from_quantized_model` on the proxy instead. The result is real data-driven sensitivity even on machines where the source does not fit in RAM.

## Changes

- New `_build_proxy_for_sensitivity` helper that wraps `mlx_lm.convert` with uniform 4-bit affine settings.
- New `auto_proxy_sensitivity: bool = True` parameter on `quantize_oq_streaming`, `QuantTask`, `OQManager.start_quantization`, and `OQStartRequest`. Defaults to True so existing callers and the admin UI keep working with the new behavior.
- The OOM-guard branch now dispatches:
  - explicit `sensitivity_model_path` -> use it (unchanged).
  - model exceeds RAM and auto-proxy enabled -> build proxy, measure on proxy, clean up in `finally`. On any build failure, fall back to position-based with an `error` log that points the user at the `sensitivity_model_path` escape hatch.
  - model exceeds RAM and auto-proxy disabled -> fall back to position-based with a `warning` log.
  - otherwise -> normal full-fp16 measurement (unchanged).
- Proxy directory is anchored to the output path's parent via `tempfile.mkdtemp(dir=...)` so it lands on the same volume the user already provisioned for the quantized output. This avoids the trap of Linux `/tmp` being tmpfs (RAM-backed), which would defeat the whole point of falling back to disk.

## Test plan

- [x] Five new unit tests in `TestBuildProxyForSensitivity` cover convert args, tempdir behavior, cleanup ownership, dtype propagation, and `working_dir` anchoring.
- [x] Full `tests/test_oq.py -m "not slow"` passes (182/182).
- [x] Ruff baseline unchanged (38 errors, no new violations).
- [x] Backward compat verified at runtime: `QuantTask` and `OQStartRequest` both default to True, override to False works.
- [x] Real model smoke test: oQ4 a model that exceeds host RAM, verify proxy gets built next to output, sensitivity_map is populated, proxy gets cleaned up after.

## Known follow-ups

- `trust_remote_code` is hardcoded to False in the proxy build. Models that need it will fall back to position-based with an error log. Threading this through `QuantTask` / API is a small follow-up.
- mlx-lm `convert()` writes its `[INFO]` lines via `print` rather than logging, so during admin background tasks the user sees mixed streams. Capturing stdout into the logger is a separate cleanup.
- No UI checkbox to opt out yet. The backend supports it via the API. UI exposure is a follow-up to `_models.html` + `dashboard.js` + i18n.

### DIFF
--- a/omlx/admin/oq_manager.py
+++ b/omlx/admin/oq_manager.py
@@ -75,6 +75,7 @@ class QuantTask:
     text_only: bool = False
     dtype: str = "bfloat16"
     preserve_mtp: bool = False
+    auto_proxy_sensitivity: bool = True
 
     def to_dict(self) -> dict:
         """Serialize task to JSON-compatible dict."""
@@ -232,6 +233,7 @@ class OQManager:
         text_only: bool = False,
         dtype: str = "bfloat16",
         preserve_mtp: bool = False,
+        auto_proxy_sensitivity: bool = True,
     ) -> QuantTask:
         """Start a quantization job.
 
@@ -307,6 +309,7 @@ class OQManager:
             text_only=text_only,
             dtype=dtype,
             preserve_mtp=preserve_mtp,
+            auto_proxy_sensitivity=auto_proxy_sensitivity,
         )
         self._tasks[task_id] = task
 
@@ -464,6 +467,7 @@ class OQManager:
                     task.sensitivity_model_path,
                     task.dtype,
                     task.preserve_mtp,
+                    task.auto_proxy_sensitivity,
                 )
 
                 if task_id in self._cancelled:
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -296,6 +296,7 @@ class OQStartRequest(BaseModel):
     text_only: bool = False
     dtype: str = "bfloat16"
     preserve_mtp: bool = False
+    auto_proxy_sensitivity: bool = True
 
 
 class HFUploadRequest(BaseModel):
@@ -4965,6 +4966,7 @@ async def start_oq_quantization(
             text_only=request.text_only,
             dtype=request.dtype,
             preserve_mtp=request.preserve_mtp,
+            auto_proxy_sensitivity=request.auto_proxy_sensitivity,
         )
         return {"success": True, "task": task.to_dict()}
     except ValueError as e:
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -11,6 +11,7 @@ import json
 import logging
 import re
 import shutil
+import tempfile
 import time as _time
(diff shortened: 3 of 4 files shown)


deepsweet:
But that wouldn't change the behavior when the sensitivity model is explicitly provided? I'm fine with macOS using a swap file, and want to keep control over the sensitivity model quantization.

jundot:
Yep, providing `sensitivity_model_path` always takes priority. The dispatch order is: explicit path first, then auto-proxy on RAM-exceed only when no path is set, then position-based fallback. Your workflow is unchanged. If you want to disable the new auto-proxy step entirely while still letting the OOM trigger fire (e.g. trust swap on macOS), pass `auto_proxy_sensitivity=False` and you'll get the prior position-based behavior.

jundot:
Update (b2f4690): tightened the "no fake oQ" guarantee.

Previously the OOM-guard and inner measurement helpers fell back to `sensitivity_map = {}` when sensitivity could not be measured, letting `_build_quant_plan`'s protection-floor U-shape become the only signal. The output looked like real oQ from the user's side but skipped the data-driven step entirely.

`quantize_oq_streaming` now raises `RuntimeError` in any of the following cases:

- Auto-proxy build or measurement fails (e.g. mlx-lm cannot load the source).
- `auto_proxy_sensitivity=False` and the source model exceeds the RAM threshold.
- Any inner sensitivity helper returns `{}` (model load failure, calibration data unavailable, unsupported architecture).

Workflows that pass `sensitivity_model_path` explicitly are unchanged. Two new tests in `TestSensitivityRequiredEnforcement` lock the contract; `TestQuantizeOqStreamingFp8` gets an autouse fixture that mocks the sensitivity helpers since those tests target the FP8 dequant path on synthetic data.

184/184 tests pass, ruff baseline unchanged.

deepsweet:
@jundot hi again.

A question: is 4-bit affine "fallback" enough for the sensitivity model role? In my understanding it could potentially lead to a different dynamic bits allocation since we are going to apply a quantization on top of another quantization, and only then calculate how much it hurts? I've always quantized oQ using affine Q8 sensitivity model with such assumption in mind.

PR #1139, opened 2026-05-09
fix(cache): pass full entry to _hot_cache_entry_size in get_stats_for_model

ivaniguarans:
## Summary

`get_stats_for_model()` passes `entry["tensors_raw"]` to `_hot_cache_entry_size()`, but the method expects the full entry dict — it looks for `"tensors_raw"` or `"arrays"` as top-level keys. Passing the inner `tensors_raw` dict means neither key is found, so it returns 0. The result is that `hot_cache_size_bytes` is always 0 in the per-model stats returned by the admin dashboard.

The global `get_stats()` path is unaffected because it reads the pre-tracked `_hot_cache_total_bytes` field directly. Every other call site in the file already passes the full entry.

## Changes

One-line fix: `entry["tensors_raw"]` → `entry` in `get_stats_for_model()`.

## Test plan

Regression test covers both single-model and multi-model cases: saves blocks under two different model names, verifies each model reports its own non-zero size, and confirms per-model sizes sum to the global total.

Verified running two models simultaneously (Qwen3.6-35B-A3B-oQ4-mtp and Qwen3.6-27B-oQ6-mtp). Before the fix, `/admin/api/stats` reported `hot_cache_size_bytes: 0` with 20 entries in cache. After, the 35B model reported 1.7 GB across 16 entries and the 27B model reported 2.3 GB across 8 entries — per-model sizes sum to the global total as expected.

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -2110,7 +2110,7 @@ class PagedSSDCacheManager(CacheManager):
                     if blk_meta is None or not _matches(blk_meta.model_name):
                         continue
                     hot_entries.append(entry)
-                    hot_size += self._hot_cache_entry_size(entry["tensors_raw"])
+                    hot_size += self._hot_cache_entry_size(entry)
 
             return PagedSSDCacheStats(
                 hits=self._stats["hits"],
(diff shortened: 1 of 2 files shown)


ivaniguarans:
Closing — #1171 landed with the same fix (passing the full entry dict to `_hot_cache_entry_size`) plus a second call-site fix in `_hot_cache_remove` that this PR didn't cover. No remaining delta.

Issue #879, opened 2026-04-21
translategemma-4b-it-4bit error


iamckun:
Error: {"error":{"message":"Chat template error: User role must provide `content` as an iterable with exactly one item. That item must be a `mapping(type:'text' | 'image', source_lang_code:string, target_lang_code:string, text:string | none, image:string | none)`.","type":"invalid_request_error","param":null,"code":null}}

jundot:
Hi, thanks for the report.

This isn't really a bug. TranslateGemma-4b-it is a specialized translation-only model from Google, and its chat template requires a custom content structure with source_lang_code / target_lang_code fields instead of plain text. Omlx's /v1/chat/completions endpoint doesn't support that structure, so the template errors out.

This model isn't usable through omlx's chat api directly. To run it, please apply the chat template on the client side and send the resulting prompt to /v1/completions instead. That path takes a raw prompt string and bypasses omlx's chat template logic entirely, so translation should work fine.

Closing this one out.

nightguarder:
Hi, 
I would like to reopen this issue, since I want to use TranslateGemma + OMLX as well. Unfortunately **client-side fix** caused an `TypeError` error for me after submitting the translation request to `v1/completions`

nightguarder:
Actually opened a PR to address this. Thanks for looking into it. 
#1141

PR #1141, opened 2026-05-09
fix: catch TypeError when accessing think token properties

nightguarder:
## Fix TypeError for non_thinking models
Catch TypeError alongside ValueError in three spots in scheduler.py. Without this, /v1/completions crashes on TranslateGemma.

## Explanation
Some models (e.g. TranslateGemma) don't have `_think_start_tokens` initialized, so accessing the `think_start_id` property raises TypeError  instead of ValueError.

The three existing except ValueError blocks miss this, causing engine loop crashes when serving such models via `/v1/completions`.

Catch (ValueError, TypeError) on all three sites:
- `_detect_needs_think_prefix` in scheduler step
- ThinkingBudgetProcessor think_start_id resolution
- `_resolve_think_end_token_ids think_end_id` resolution

## What this does
This enables support for _TranslateGemma_ and other non-thinking models whose
tokenizer lacks `_think_start_tokens`. Previously blocked by `TypeError`,

> _Note:_
> that you still need to update your local translategemma  chat template must be applied client-side and the
> resulting prompt sent to /v1/completions instead.  See original Issue for explanation #879 
> 

## Files modified
- `omlx/scheduler.py`


jundot:
Thanks for the writeup, and you were right that this crashed. On 0.3.5.dev1 the three sites only had `except ValueError`, so a model like TranslateGemma whose `think_start_id` raises `TypeError` slipped through and crashed `/v1/completions`.

That same bug was already fixed on main in the meantime. The `_get_think_token_id` helper in `scheduler.py` now wraps those accesses and catches both `ValueError` and `TypeError`, returning None, and all three lines you changed already call that helper.

So your fix and the helper do the same thing, which makes these inline changes redundant on main. I'll close this for that reason. If you still see the crash on the latest main, please open an issue.

PR #1146, opened 2026-05-09
fix(scheduler): use async_eval for cache-store materialization

ivaniguarans:
## Summary

When the inference thread hands KV cache arrays to the store-cache worker, it currently calls `mx.eval(*pre_eval_arrays)` which blocks until every array is materialized. On long conversations the blocked time grows with cache size — the inference thread sits idle waiting for work that doesn't need to finish before the next request can start.

This replaces the blocking `mx.eval()` with `mx.async_eval()`, which dispatches Metal commands and returns after command enqueueing completes. The worker already runs on a ThreadPoolExecutor, so it can wait for materialization itself via `mx.synchronize()` while the inference thread moves on.

## Changes

The core change is a single line: `mx.eval(*pre_eval_arrays)` → `mx.async_eval(*pre_eval_arrays)` inside `_cleanup_finished`. `async_eval` enqueues the Metal commands synchronously (all commands are submitted before it returns), so by the time `executor.submit()` fires, the GPU work is in flight.

The worker's `_async_store_cache_worker` gains a `mx.synchronize()` call before `store_cache()` — a global barrier that waits all streams. Stream-scoped sync isn't possible here because `generation_stream` is thread-local to the inference thread and invisible to the worker. The global barrier is safe because the worker is the only consumer of these arrays at this point.

A `store_cache_worker_sync` phase timer wraps the synchronize call so the worker-side wait is observable in timing logs, replacing the visibility that `store_cache_main_eval` provided on the inference thread.

Docstrings and inline comments updated to document the async_eval enqueueing guarantee and global barrier rationale.

## Evidence

Server-side phase timers on a 30-turn conversation (Qwen3-0.6B-A0.3B-4bit, 256 max tokens/turn) show `store_cache_main_eval` accumulates ~40ms total across 30 turns. Replacing with async_eval eliminates this blocking time from the inference thread entirely.

On a Qwen3-Coder-Next (3B, Q6_K) 11-turn session with realistic 100-150 token responses, cumulative `store_cache_main_prep` dropped 23% (8.7ms → 6.7ms) because the prep phase no longer includes the synchronous eval stall. The per-turn improvement is small individually (~0.2ms) but compounds over long conversations and scales with model/cache size.

The benefit is architectural: the inference thread should not block for work that only the store-cache worker needs. Models with standard KV cache (transformers) see the full effect; hybrid architectures using ArraysCache (e.g. GatedDeltaNet) produce empty `pre_eval_arrays` and are unaffected.

## Test plan

- Full test suite passes (4147 passed, 34 pre-existing failures unrelated to this change)
- Multi-turn conversation benchmark (30 turns) confirms no regression in TTFT or cache integrity
- Phase timer logs verified: `store_cache_worker_sync` appears in worker timing, `store_cache_main_eval` no longer appears on inference thread

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -803,9 +803,16 @@ class Scheduler:
         """Run store_cache + paged_cache cleanup off the inference thread.
 
         Pre-conditions enforced by the caller (_cleanup_finished):
-        - All mx.array references in cache_to_store have already been
-          mx.eval()'d on the inference thread, so save_block's internal
-          mx.eval is a no-op.
+        - mx.async_eval() was called on the inference thread for all
+          KV cache arrays, dispatching materialization asynchronously
+          without blocking the inference thread. async_eval completes
+          Metal command enqueueing before returning, so all commands
+          are submitted by the time executor.submit() runs.
+        - This worker calls mx.synchronize() (global barrier — waits
+          all streams) to ensure materialization is complete before
+          extracting tensor bytes. Stream-scoped sync is not possible
+          here because generation_stream is thread-local to the
+          inference thread.
         - bfloat16 view+eval inside _extract_tensor_bytes runs on this
           worker's default mx stream, isolated from generation_stream;
           the underlying buffer is read-only at this point.
@@ -816,6 +823,8 @@ class Scheduler:
         threading.RLock so concurrent access from main and worker is safe.
         """
         try:
+            with self._phase_timer("store_cache_worker_sync"):
+                mx.synchronize()
             block_table = self.block_aware_cache.store_cache(
                 request_id,
                 token_sequence_to_store,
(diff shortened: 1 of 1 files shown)


jundot:
Thanks, async_eval + worker-side wait is the right direction. Merging.

One bit of context to share, no action needed. The description says `mx.synchronize()` in the worker acts as a global barrier across all streams, but that part is not quite right. Here is what I traced:

- `generation_stream` is created with `mx.new_thread_local_stream()` (`mlx_lm/generate.py:226`), which means each thread gets its own backing stream behind the shared Python object. The inference thread dispatches commands to its own instance.
- When the worker thread calls `mx.synchronize()` (or even `mx.synchronize(generation_stream)`), it syncs the worker's own thread-local instance, which has nothing scheduled on it. I checked on MLX 0.31.2 from a separate thread: both calls return in ~0.1ms even right after a heavy GPU dispatch on the original thread.
- The real wait happens inside `_extract_tensor_bytes`. The `bytes(memoryview(arr))` call goes through the Python buffer protocol, which internally calls `array.eval()` and waits on a per-array event that `async_eval` attached. That event is signaled when GPU work completes, and the wait works correctly across threads. So this is the path that actually serializes things.

So correctness holds, but two side effects worth knowing: the `mx.synchronize()` in the worker is effectively dead code, and `store_cache_worker_sync` will read ~0 in the timing logs because the real wait gets absorbed into `store_cache`. Just sharing in case it is useful context later.

ivaniguarans:
Thanks for the detailed trace — that's useful context. I had the threading model wrong in the description: `mx.synchronize()` syncs the caller's thread-local stream instance, not a global barrier across all streams. The actual cross-thread serialization happens through the buffer protocol in `_extract_tensor_bytes`, where `bytes(memoryview(arr))` triggers `array.eval()` and waits on the per-array event from `async_eval`.

Good to know `store_cache_worker_sync` will effectively read ~0 in timing logs — the real wait is absorbed into `store_cache` during tensor byte extraction. I'll keep this in mind for future cache/scheduler work.

PR #1147, opened 2026-05-09
fix(mtp): guard MoE sanitize against missing MTP expert weights

### LINKED ISSUE #1138: Unable to benchmark oQ-MTP
Whenever I try to benchmark anything, either old `oQ5-FP16` or the new `oQ5-FP16-MTP` on current commit https://github.com/jundot/omlx/commit/4669900b43ebe69712f345b1e4d5dcff7c65088e:

```
2026-05-09 13:58:31,454 - omlx.admin.benchmark - ERROR - Benchmark error: 'language_model.mtp.layers.0.mlp.experts.0.gate_proj.weight'
  File "omlx/admin/benchmark.py", line 643, in run_benchmark
    engine = await engine_pool.get_engine(request.model_id, force_lm=True)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "omlx/engine_pool.py", line 414, in get_engine
    await self._load_engine(model_id, force_lm=force_lm)
  File "omlx/engine_pool.py", line 705, in _load_engine
    await engine.start()
  File "omlx/engine/batched.py", line 235, in start
    self._model, self._tokenizer = await loop.run_in_executor(
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/deepsweet/.local/share/uv/python/cpython-3.12.9-macos-aarch64-none/lib/python3.12/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "omlx/engine/batched.py", line 229, in _load_model_sync
    return load(
           ^^^^^
  File ".venv/lib/python3.12/site-packages/mlx_lm/utils.py", line 491, in load
    model, config = load_model(model_path, lazy, model_config=model_config)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/mlx_lm/utils.py", line 346, in load_model
    weights = model.sanitize(weights)
              ^^^^^^^^^^^^^^^^^^^^^^^
  File "omlx/patches/mlx_lm_mtp/qwen35_model.py", line 694, in sanitize
    _stack_per_expert(new_weights, prefix, num_experts)
  File "omlx/patches/mlx_lm_mtp/qwen35_model.py", line 654, in _stack_per_expert
    weights.pop(f"{prefix}.experts.{e}.{n}.weight")
KeyError: 'language_model.mtp.layers.0.mlp.experts.0.gate_proj.weight'
```

ivaniguarans:
## Summary

When loading an oQ-quantized Qwen 3.5/3.6 MoE model that was quantized without `preserve_mtp`, the MTP sanitize path crashes with a `KeyError` inside `_stack_per_expert`. The config still declares `mtp_num_hidden_layers > 0`, so the sanitize loop tries to reshape expert weights that don't exist.

## Root Cause

`sanitize` iterates over `range(mtp_num)` and attempts to unfuse or stack MTP expert weights. It checks the weight format (switch_mlp, fused gate_up_proj, or per-expert tensors) but never checks whether MTP weights exist *at all*. When they're absent — the common case for oQ checkpoints quantized without `preserve_mtp` — the per-expert fallback runs `_stack_per_expert`, which pops keys that aren't in the dict.

## Fix

Before entering the MTP expert loop, check whether any `language_model.mtp.*` key exists in the weight dict. If none are found, short-circuit with a debug log explaining the mismatch between config and weights. The existing format-specific guards (`switch_mlp`, fused, per-expert) are also tightened with key-existence checks so each branch only runs when its expected keys are present.

The switch_mlp idempotency guard is preserved: oQ outputs already store experts in switch_mlp form (the mlx-vlm sanitize patch unfuses MTP experts before quantization), so the load-time unfuse/stack is correctly skipped in that case.

## Test plan

Four regression tests in `TestQwen35MoeSanitize`:

- `test_sanitize_no_mtp_weights` — the crash scenario: `mtp_num_hidden_layers=1` but no MTP keys in weights. Verifies no crash and a debug log message.
- `test_sanitize_switch_mlp_form` — oQ output format (already in switch_mlp layout). Verifies idempotent pass-through.
- `test_sanitize_per_expert_form` — Qwen 3.5 per-expert tensors. Verifies stacking into switch_mlp layout.
- `test_sanitize_fused_form` — Qwen 3.6 fused gate_up_proj. Verifies unfusing into switch_mlp layout.

All 40 MTP patch tests pass. 13 pre-existing failures in unrelated areas (server endpoints, grammar, update check, gated delta).

Closes #1138

### DIFF
--- a/omlx/patches/mlx_lm_mtp/qwen35_model.py
+++ b/omlx/patches/mlx_lm_mtp/qwen35_model.py
@@ -676,22 +676,35 @@ def _patch_qwen3_5_moe() -> None:
             getattr(self.language_model.args, "mtp_num_hidden_layers", 0) or 0
         )
         if mtp_num > 0:
-            num_experts = self.language_model.args.num_experts
-            mtp_is_fused = (
-                "language_model.mtp.layers.0.mlp.experts.gate_up_proj"
-                in new_weights
+            has_any_mtp = any(
+                k.startswith("language_model.mtp.") for k in new_weights
             )
-            for layer_idx in range(mtp_num):
-                prefix = f"language_model.mtp.layers.{layer_idx}.mlp"
-                # Idempotent: oQ outputs already store experts in switch_mlp
-                # form (mlx-vlm sanitize patch unfuses MTP experts before
-                # quantization). Skip the load-time unfuse/stack in that case.
-                if f"{prefix}.switch_mlp.gate_proj.weight" in new_weights:
-                    continue
-                if mtp_is_fused:
-                    _unfuse_experts(new_weights, prefix)
-                else:
-                    _stack_per_expert(new_weights, prefix, num_experts)
+            if not has_any_mtp:
+                logger.debug(
+                    "mtp_num_hidden_layers=%d but no MTP weights found; "
+                    "model may have been quantized without preserve_mtp",
+                    mtp_num,
+                )
+            else:
+                num_experts = self.language_model.args.num_experts
+                mtp_is_fused = (
+                    "language_model.mtp.layers.0.mlp.experts.gate_up_proj"
+                    in new_weights
+                )
+                for layer_idx in range(mtp_num):
+                    prefix = f"language_model.mtp.layers.{layer_idx}.mlp"
+                    # Idempotent: oQ outputs already store experts in
+                    # switch_mlp form (mlx-vlm sanitize patch unfuses MTP
+                    # experts before quantization). Skip the load-time
+                    # unfuse/stack in that case.
+                    if f"{prefix}.switch_mlp.gate_proj.weight" in new_weights:
+                        continue
+                    if mtp_is_fused:
+                        if f"{prefix}.experts.gate_up_proj" in new_weights:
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the quick turnaround on this. Root cause is exactly right: `mtp_num_hidden_layers > 0` in config with no `language_model.mtp.*` keys in weights (preserve_mtp off during quant) is a legitimate state, so the top-level `has_any_mtp` short-circuit + `debug` log is the right level. The per-branch existence guards are a nice belt-and-braces against partial-MTP corruption too.

Test coverage on all four formats (no-mtp / switch_mlp / per-expert / fused) is solid. LGTM, will merge for next release.

PR #1156, opened 2026-05-09
fix: cross-thread generation_stream sync, TTS bfloat16, F821 model_dir

a4501150:
## Summary
- **scheduler.py**: Add `_safe_sync_generation_stream()` helper that wraps `mx.synchronize(generation_stream)` in try/except RuntimeError. Replaces all 6 bare call sites outside `_sync_and_clear_cache()`. Prevents "There is no Stream(gpu, N) in current thread" crashes during teardown.
- **engine_core.py**: Dispatch `scheduler.deep_reset()` through `_mlx_executor` in `close()` so the Metal stream sync runs on the thread that owns the stream.
- **tts.py**: Cast `mx.bfloat16` arrays to `mx.float32` before `np.array()` — NumPy has no bfloat16 dtype.
- **server.py**: Fix F821 — `Path(model_dir)` where `model_dir` is undefined in `init_server` fallback; replaced with `Path.home() / ".omlx"`.

## Test plan
- [x] Full test suite passes (4193 passed, 0 failed)
- [x] Verify TTS output with a bfloat16 model
- [x] Verify engine close/teardown doesn't crash on M3/M4

### DIFF
--- a/omlx/engine/tts.py
+++ b/omlx/engine/tts.py
@@ -229,7 +229,10 @@ class TTSEngine(BaseNonStreamingEngine):
             audio_chunks = []
 
             for result in results:
-                audio_chunks.append(np.array(result.audio))
+                audio = result.audio
+                if isinstance(audio, mx.array) and audio.dtype == mx.bfloat16:
+                    audio = audio.astype(mx.float32)
+                audio_chunks.append(np.array(audio))
 
             if not audio_chunks:
                 raise RuntimeError("TTS model produced no audio output")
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for the careful debugging. The close() trace is correct: scheduler.shutdown() drains pending removes which call mx.synchronize(generation_stream) on the main thread, and deep_reset hits the same issue via _do_abort_request. Dispatching deep_reset through _mlx_executor is the right call. I'll merge as-is and follow up to also dispatch scheduler.shutdown() so the broad RuntimeError swallow can be narrowed later. TTS bfloat16 fix looks good too.

a4501150:
Glad we have an active maintainer - thanks @jundot - it's a great project

PR #1158, opened 2026-05-09
feat: Qwen3.6 nested-visual patch, native reasoning, and Responses API reasoning items

a4501150:
## Summary
- **Qwen3.6 nested-visual sanitize patch** (`omlx/patches/qwen3_6_nested_visual.py`): Monkey-patches mlx-vlm's `qwen3_5_moe.Model.sanitize` to remap `language_model.model.visual.*` → `vision_tower.*` for Qwen3.6-35B-A3B's nested ViT layout. Self-guarding: no-op when mlx-vlm already handles the remapping. Wired into both VLM engine load and oQ quantization paths.
- **thinking.py `start_in_thinking`**: Add parameter to `extract_thinking()` and `ThinkingParser` for native-reasoning models (e.g. Qwen3.6) where the prompt pre-opens `<think>`. Complements the existing malformed-tag recovery.
- **Responses API reasoning output items**: Full streaming + non-streaming support for reasoning output items:
  - `IDPrefix.REASONING` and `ReasoningSummaryPart` model
  - `build_reasoning_output_item()` helper
  - Lazy item emission in streaming: reasoning opens on first thinking_delta, message opens on first content_delta
  - `response.reasoning_summary_text.delta` SSE events
  - `pending_reasoning` tracking in input/output message conversion
  - `native_reasoning` detection from `preserve_thinking_default`

## Test plan
- [x] Full test suite passes (4193 passed, 0 failed)
- [x] Verify Qwen3.6-35B-A3B inference with nested ViT layout
- [x] Verify Responses API streaming with native-reasoning model emits reasoning items
- [x] Verify non-streaming Responses API includes reasoning output item

### DIFF
--- a/omlx/api/responses_models.py
+++ b/omlx/api/responses_models.py
@@ -140,13 +140,20 @@ class OutputContent(BaseModel):
     annotations: List[Any] = Field(default_factory=list)
 
 
+class ReasoningSummaryPart(BaseModel):
+    """A single part of a reasoning summary."""
+
+    type: str = "summary_text"
+    text: str = ""
+
+
 class OutputItem(BaseModel):
     """A single item in the response output array.
 
-    Can be a message or a function_call.
+    Can be a message, function_call, or reasoning.
     """
 
-    type: str  # "message" or "function_call"
+    type: str  # "message" or "function_call" or "reasoning"
     id: str
     status: str = "completed"
     # message fields
@@ -156,6 +163,8 @@ class OutputItem(BaseModel):
     call_id: Optional[str] = None
     name: Optional[str] = None
     arguments: Optional[str] = None
+    # reasoning fields
+    summary: Optional[List[ReasoningSummaryPart]] = None
 
 
 class InputTokensDetails(BaseModel):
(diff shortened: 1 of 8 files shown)


jundot:
thanks a lot for the patch. confirmed the nested-visual bug is live on main. unit-verified f96138e mlx-vlm's sanitize_key uses if/elif and silently drops 333 vision params on HF-source load of Qwen3.6-35B-A3B. today's mlx_vlm_mtp runtime sanitize re-introduces the same shape, so the wrap is still the right call.

one snag with the current change: apply_qwen3_6_nested_visual_patch() in omlx/engine/vlm.py sits after vlm_load(), so on the inference path the wrap installs after sanitize has already run. oq.py wiring is fine (called before model load in _build_model_sanitizer).

since the branch is ~50 commits behind main (mlx-vlm bump, mlx_vlm_mtp runtime patch, OCR patch, oQ auto-proxy, anthropic stream fix, etc) and the inference-path call needs to move into maybe_apply_pre_load_patches anyway, I'll close this and land the nested-visual fix directly on main with you as co-author so it ships in the next release.

start_in_thinking and the Responses API reasoning items look reasonable but they're UX polish (the model still produces a correct answer either way, reasoning just gets folded into the message body without the split). holding those as a separate follow-up so the vision fix doesn't get blocked.

thanks again for catching this one.

a4501150:
Sounds good, thanks for the fixing @jundot !

PR #1161, opened 2026-05-10
fix(eval): correct KMMLU answer mapping

khsd6327:
## Summary
- Fix KMMLU answer normalization to map the bundled 1-indexed labels (`1`-`4`) to `A`-`D`.
- Keep KMMLU prompt choice labels aligned with that mapping.

## Why
The bundled KMMLU files store answers as `1`, `2`, `3`, and `4`, but `KMMLUBenchmark` treated them as zero-indexed labels. That shifted answers by one letter, and raw answer `4` became the string `"4"`, which could never match extracted model answers (`A`-`D`).

This caused KMMLU accuracy to be severely underreported across models.

## Evidence
The bundled KMMLU answer distribution is 1-indexed:

```text
kmmlu_test.jsonl: {1: 3458, 2: 10418, 3: 10477, 4: 10677}
kmmlu_dev.jsonl:  {1: 8,    2: 5,     3: 7,     4: 5}
```

The first five loaded KMMLU answers normalize to:

```text
B, A, A, D, C
```

## Validation

Before the fix, KMMLU scored abnormally low:

```text
Model: Qwen3.6-35B-A3B
Benchmark: KMMLU
Mode: Thinking
Correct: 11 / 300
Accuracy: 3.7%
Time: 2278.2s
```

After restarting the server with the fix loaded and rerunning KMMLU:

```text
Model: Qwen3.6-35B-A3B
Benchmark: KMMLU
Mode: Thinking
Correct: 41 / 50
Accuracy: 82.0%
Time: 406.5s
```

### DIFF
--- a/omlx/eval/kmmlu.py
+++ b/omlx/eval/kmmlu.py
@@ -17,7 +17,7 @@ from .datasets import load_jsonl, stratified_sample
 logger = logging.getLogger(__name__)
 
 DATA_DIR = Path(__file__).parent / "data"
-ANSWER_MAP = {0: "A", 1: "B", 2: "C", 3: "D"}
+ANSWER_MAP = {1: "A", 2: "B", 3: "C", 4: "D"}
 
 
 def _format_subject_name(subject: str) -> str:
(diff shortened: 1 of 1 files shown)


jundot (approved):
nice catch. The bundled KMMLU files are 1-indexed (answer=1..4 across both test and dev), so the old 0-indexed map was silently shifting every label and dropping every answer=4 record (~30% of test) into the str(answer_idx) fallback, which can never match A-D. That lines up with the 3.7% baseline you reported.

The fix is atomic in the right way: ANSWER_MAP and enumerate(start=1) are updated together so prompt labels and normalized answers stay consistent. No other eval (mmlu/hellaswag use 0-indexed data, cmmlu/jmmlu store letters) is affected.

LGTM.

Issue #1127, opened 2026-05-08
Why max_token setting doesn't work?


FdyCN:
I download Qwen3.6-35B-A3B-OptiQ-4bit and called it in opencode，BUT I set the max_token = 200000 & ctx_window = 200000

<img width="2606" height="322" alt="Image" src="https://github.com/user-attachments/assets/fdf96695-ecc1-4635-be35-be3ebc626a36" />

BUT when I used in opencode, It seems that the max_tokens of Qwen3.6-35B-A3B-OptiQ-4bit is just 32k.
SO, my setting doesn't work? or I did something wrong? 

<img width="1812" height="710" alt="Image" src="https://github.com/user-attachments/assets/291a8dd0-7b9a-4a2d-9a01-c42e553cd597" />

PR #1163, opened 2026-05-10
feat(audio): support max_tokens for /v1/audio/transcriptions

thornad:
## Problem

`/v1/audio/transcriptions` has no way to raise the per-call output token cap. mlx-audio's STT models each set their own default in `generate(...)`:

| Model | `max_tokens` default |
|---|---|
| VibeVoice-ASR | 8192 |
| canary | 200 |
| granite_speech | 4096 |
| whisper / sensevoice / fireredasr2 | accept `**kwargs` |

For `VibeVoice-ASR-8bit` specifically, an 8192 cap means a 60-min file truncates at ~24 min of transcript even though the model itself advertises ~60 min audio + 64k context. Confirmed locally — issue #1127 is the same shape on the LLM side.

oMLX already has the right infrastructure for this: `ModelSettings.max_tokens` exists and chat completions reads from it (`omlx/server.py:917-927`). Audio just doesn't.

## Change

Mirror the chat-completions precedence on the STT path, with one new field:

```
request.max_tokens > ModelSettings.max_tokens > model's own generate() default
```

- `AudioTranscriptionRequest`: new `max_tokens: Optional[int] = None` (oMLX extension; documented).
- `POST /v1/audio/transcriptions`: new `max_tokens: Optional[int] = Form(None)`. When omitted, the route reads `_server_state.settings_manager.get_settings(model_id).max_tokens`. When that's also `None`, nothing is forwarded and the model's own default applies — fully backward-compatible.
- `STTEngine.transcribe(..., **kwargs)` already forwards to `model.generate`, so no engine change. Quick audit confirmed every mlx-audio STT model in tree either accepts `max_tokens=` or has a `**kwargs` catch-all.

After this, a user can set
```json
"VibeVoice-ASR-8bit": { "max_tokens": 65536 }
```
in `~/.omlx/model_settings.json` and stop hitting the 24-min ceiling, without changing any client code.

## Tests

Four new tests in `tests/test_audio_stt.py`:

- `test_max_tokens_forwarded_to_engine` — explicit form field flows to `engine.transcribe(..., max_tokens=...)`.
- `test_max_tokens_omitted_when_not_set_and_no_setting` — neither source set ⇒ kwarg not passed (model default wins).
- `test_max_tokens_falls_back_to_per_model_setting` — request omits, `ModelSettings.max_tokens=65536` ⇒ 65536 forwarded.
- `test_max_tokens_request_overrides_per_model_setting` — request value wins over settings value.

Full `tests/test_audio_stt.py` passes (the 5 deselected failures pre-date this change — `mx.new_thread_local_stream` mlx-version mismatch in dev `.venv`, unrelated to audio routing).

## Backward compatibility

- OpenAI clients sending the old transcription request shape: unchanged behaviour, `max_tokens` defaults to `None`.
- Models that don't accept `max_tokens` keyword: not currently a concern (every STT model in `mlx_audio/stt/models/` either takes it explicitly or has `**kwargs`), but since the kwarg is only added when an explicit value exists, even a future strict-signature model only fails on clients who opted in.

### DIFF
--- a/omlx/api/audio_models.py
+++ b/omlx/api/audio_models.py
@@ -13,13 +13,21 @@ from pydantic import BaseModel
 
 
 class AudioTranscriptionRequest(BaseModel):
-    """OpenAI-compatible audio transcription request."""
+    """OpenAI-compatible audio transcription request.
+
+    ``max_tokens`` is an oMLX extension (no equivalent in the OpenAI shape) that
+    raises the per-call output cap for STT models whose ``generate(..., max_tokens=...)``
+    default is too tight for long files — e.g. mlx-audio's VibeVoice-ASR caps at
+    8192 tokens (~24 min of audio) by default while the model itself supports
+    ~60 min / 64k context. When ``None``, the model's own default is used.
+    """
 
     model: str
     language: Optional[str] = None
     prompt: Optional[str] = None
     response_format: Optional[str] = "json"
     temperature: Optional[float] = 0.0
+    max_tokens: Optional[int] = None
 
 
 class AudioTranscriptionResponse(BaseModel):
--- a/omlx/api/audio_routes.py
+++ b/omlx/api/audio_routes.py
@@ -76,6 +76,19 @@ def _resolve_model(model_id: str) -> str:
     return resolve_model_id(model_id) or model_id
 
 
+def _get_settings_manager():
+    """Return the active ModelSettingsManager from server state, or None.
+
+    Lazy import + defensive guard so the audio router stays usable in tests
+    that don't bring up the full server state.
+    """
+    try:
+        from omlx.server import _server_state
+    except Exception:
+        return None
+    return getattr(_server_state, "settings_manager", None)
+
+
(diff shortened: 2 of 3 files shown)


jundot:
Thanks for the careful writeup — the motivation is real and the precedence matches the chat path at `server.py:923-931` one-to-one. Verified locally against mlx-audio 0.4.3:

- VibeVoice-ASR / Canary / Granite-Speech / Moonshine / Cohere-ASR / GLM-ASR / Qwen2-Audio / Qwen3-ASR / Voxtral / Voxtral-realtime: `max_tokens` applies.
- Whisper / SenseVoice / MMS / Parakeet / FireRedASR2: silently drop `max_tokens` (each pops it from `kwargs` / `decode_options`). No crash, just a no-op for those backends.

So zero regression risk and the field becomes load-bearing for exactly the ASR models that benefit. Tests cover all four precedence paths and patch `_get_settings_manager` directly, which is the right move for the fixture.

Merging now.

PR #1165, opened 2026-05-10
fix(admin): left panel icon toggling improvement

bogdan-copocean:
- Removing redundant panel icon toggle from the side bar
- Conditionally showing panel icon on `sideBarOpen` flag

| State | **Closed** | **Opened** |
|---|---|---|
| **AFTER** | Closed After | Open After |
| **BEFORE** | Closed Before | Open Before |


jundot:
Thanks for the cleanup. Having a single state-aware toggle is much clearer than the old static `panel-left` icon, and the `x-show` swap matches the pattern already used for the theme icons in this file (`sun-moon` / `sun` / `moon`), so the lucide processor in `base.html` keeps Alpine reactivity intact on the replaced SVG.

LGTM, will merge.

PR #1168, opened 2026-05-10
fix(engine): keep streaming abort engine reference

### LINKED ISSUE #1167: Error during chat streaming: 'NoneType' object has no attribute 'abort_request'
## Bug: `abort_request` called on NoneType during chat streaming

**oMLX Version:** 0.3.8 (DMG via `/Applications/oMLX.app`)
**macOS:** Darwin 25.4.0 (arm64)
**Hardware:** Apple M4 Max, 48 GB unified memory
**Model:** `Qwen3___5-122B-A10B-4bit` (also `Qwen3.6-27B-MTPLX-Optimized-Speed`)

### Description

The server logs `Error during chat streaming: 'NoneType' object has no attribute 'abort_request'` repeatedly when requests are cancelled or timeout. This error occurs in `omlx.server` during streaming chat completion and is **not** a direct consequence of the scheduler `KeyError` already reported in #1031 and #944 — it happens independently during normal request abort paths.

### Log Evidence

Multiple occurrences observed across several days:

```
2026-05-11 01:11:41,800 - omlx.server - ERROR - [-] - Error during chat streaming: 'NoneType' object has no attribute 'abort_request'
2026-05-11 01:11:41,800 - omlx.server - ERROR - [-] - Error during chat streaming: 'NoneType' object has no attribute 'abort_request'
2026-05-11 01:11:41,800 - omlx.server - ERROR - [-] - Error during chat streaming: 'NoneType' object has no attribute 'abort_request'
2026-05-11 01:36:19,821 - omlx.server - ERROR - [-] - Error during chat streaming: 'NoneType' object has no attribute 'abort_request'
```

Also seen on 2026-05-10 at `01:29:14`, `01:31:08`, and `12:38:46`.

This pattern was also observed in #1070's log excerpt:
```
2026-05-05 13:58:37,523 - omlx.engine.vlm - INFO - [-] - [vlm_stream_generate] Aborting request f55d804d-53b9-40f8-beb0-595ed1bc530d
2026-05-05 13:58:37,524 - omlx.server - ERROR - [-] - Error during chat streaming: 'NoneType' object has no attribute 'abort_request'
```

### Trigger Conditions

The error fires when:
1. A streaming `/v1/chat/completions` request is cancelled mid-generation (client disconnect, timeout, or tool-use abort)
2. The server attempts to abort the request on the engine but the engine reference is `None`
3. `'NoneType' object has no attribute 'abort_request'` is raised and caught at the server level

### Expected Behavior

When a streaming request is cancelled, abort should be handled gracefully without raising exceptions. If the engine reference is unavailable (e.g., already unloaded or not yet assigned), the abort call should be a no-op or log a warning instead of error-level.

### Actual Behavior

Error-level log spam on every cancelled/timed-out streaming request. While this appears to be caught and doesn't crash the server (unlike the scheduler KeyError), it clutters logs and indicates the abort path doesn't handle missing engine references properly.

### Related Issues

- #1031 — Scheduler KeyError during cancelled external prefill (different crash path, same abort scenario)
- #944 — KeyError in scheduler._schedule_waiting (tool calling variant)
- #1070 — Cache doesn't store properly (contains this same error in log excerpt)
- #667 — Prefill interrupted / request(s) aborted issues

manaskarra:
## Summary
- capture the active async engine at the start of streaming generation
- use the captured engine for request creation, output streaming, and abort cleanup
- avoid `NoneType.abort_request` when shutdown/unload clears `self._engine` while an async stream generator is finalizing
- add focused regression coverage for text and VLM streaming cleanup paths

Fixes #1167.

## Test plan
- `python -m py_compile omlx/engine/batched.py omlx/engine/vlm.py tests/test_batched_engine.py tests/test_vlm_engine.py`
- `git diff --check`

Note: full pytest collection could not run in this Linux environment because `mlx` is unavailable (`ModuleNotFoundError: No module named 'mlx'`). The added tests are focused on mocked streaming cleanup and are intended to run in the project MLX environment.

### DIFF
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -561,7 +561,8 @@ class BatchedEngine(BaseEngine):
                 "specprefill_system_end"
             )
 
-        request_id = await self._engine.add_request(
+        engine = self._engine
+        request_id = await engine.add_request(
             prompt=prompt,
             sampling_params=sampling_params,
             **specprefill_kwargs,
@@ -569,7 +570,7 @@ class BatchedEngine(BaseEngine):
 
         finished_normally = False
         try:
-            async for output in self._engine.stream_outputs(request_id):
+            async for output in engine.stream_outputs(request_id):
                 text = clean_special_tokens(output.output_text)
 
                 # Set finished_normally BEFORE yield, because the consumer
@@ -600,7 +601,7 @@ class BatchedEngine(BaseEngine):
                 logger.info(
                     f"[stream_generate] Aborting request {request_id} (finished_normally={finished_normally})"
                 )
-                await self._engine.abort_request(request_id)
+                await engine.abort_request(request_id)
             else:
                 logger.debug(
                     f"[stream_generate] Request {request_id} finished normally"
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -1362,7 +1362,8 @@ class VLMBatchedEngine(BaseEngine):
         if kwargs.get("specprefill_system_end") is not None:
             specprefill_kwargs["specprefill_system_end"] = kwargs.pop("specprefill_system_end")
 
-        request_id = await self._engine.add_request(
+        engine = self._engine
+        request_id = await engine.add_request(
             prompt=prompt,
             sampling_params=sampling_params,
             vlm_inputs_embeds=vlm_inputs_embeds,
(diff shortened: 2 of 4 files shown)


jundot:
Thanks for the fix. Capturing the engine ref at stream start is the right call. `BatchedEngine.stop()` can null out `self._engine` mid-stream (e.g. during `EnginePool._unload_engine`), and snapshotting locally keeps `add_request` / `stream_outputs` / `abort_request` consistent. Tests directly exercise the invariant via `aclose()` after clearing `self._engine`. Merging.

PR #1171, opened 2026-05-11
[codex] Fix hot cache byte accounting

lobsterbuko:
## Summary
- Fix hot cache byte accounting when removing raw-byte entries.
- Fix per-model hot cache stats so tensors_raw entries contribute to hot_cache_size_bytes.
- Add focused regression coverage for raw-byte hot cache removal and per-model stats.

## Root cause
_hot_cache_remove and get_stats_for_model passed entry["tensors_raw"] into _hot_cache_entry_size(), but that helper expects the full hot cache entry and looks for top-level arrays or tensors_raw keys. Passing the nested raw tensor dict made the helper return 0.

## Impact
Before this change, explicit hot cache removal failed to decrement _hot_cache_total_bytes for tensors_raw entries, and model-scoped stats under-reported hot cache bytes. Once the counter drifted high enough, the hot cache could stop admitting new entries even after removals.

## Validation
- uv run -p 3.13 --extra dev python -m pytest tests/test_hot_cache.py -q
- git diff --check

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -800,9 +800,7 @@ class PagedSSDCacheManager(CacheManager):
         with self._hot_cache_lock:
             old = self._hot_cache.pop(block_hash, None)
             if old:
-                self._hot_cache_total_bytes -= self._hot_cache_entry_size(
-                    old["tensors_raw"]
-                )
+                self._hot_cache_total_bytes -= self._hot_cache_entry_size(old)
 
     def _promote_to_hot_cache(
         self,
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the catch. Root cause read is right — `_hot_cache_entry_size` expects the full entry and falls through to 0 when handed the nested `tensors_raw` dict, so `_hot_cache_remove` and `get_stats_for_model` were drifting the counter and zeroing per-model bytes. The fix aligns these two call sites with `_hot_cache_put` and the LRU eviction path, which were already passing the entry. Tests look good. Merging.

PR #1173, opened 2026-05-11
feat(server): secure default CORS + VLM tool parser whitelist

RinZ27:
## Summary

Restricts the default CORS policy to local origins and introduces a validation whitelist for VLM tool calling parsers. Also fixes a string sanitization bug in the dashboard percentage input.

## Architecture

The VLM tool calling logic uses dynamic imports based on a parser type inferred from the chat template. Previously, any inferred type was passed directly to `importlib.import_module`, which I noticed could lead to failures or unauthorized module loading if a model template contains an unexpected parser identifier. 

This PR adds an explicit `allowed_parsers` whitelist before the import to ensure only supported `mlx_lm.tool_parsers` are loaded.

For CORS, the server now defaults to `localhost` and `127.0.0.1` origins when settings are uninitialized, preventing accidental external exposure in local-only deployments.

## Changes

| File | What | LOC |
|---|---|---|
| `omlx/server.py` | Change default CORS origins from `*` to `localhost`/`127.0.0.1` | +1 |
| `omlx/engine/vlm.py` | Add `allowed_parsers` whitelist for dynamic tool parser loading | +7 |
| `omlx/admin/static/js/dashboard.js` | Switch to global regex for percent string sanitization | +1 |

## Test Plan

- [x] Verified CORS restriction doesn't break the local dashboard.
- [x] Tested model loading with various template types to confirm whitelist behavior (invalid types are now safely ignored with a warning).
- [x] Smoke tested the dashboard percentage input with multiple `%` signs.
- [x] `node --check omlx/admin/static/js/dashboard.js` clean.

## What this doesn't do

- Does not prevent users from explicitly setting `cors_origins: ["*"]` in their configuration if they require it.
- Does not add new parser types; it only secures the loading of existing ones.

### DIFF
--- a/omlx/admin/static/js/dashboard.js
+++ b/omlx/admin/static/js/dashboard.js
@@ -2878,7 +2878,7 @@
                     return { auto: false, percent: Math.min(99, Math.max(1, percent)) };
                 }
                 // Handle percent format (e.g., "69%")
-                const percent = parseInt(value.replace('%', ''));
+                const percent = parseInt(value.replace(/%/g, ''));
                 if (isNaN(percent)) return { auto: false, percent: 90 };
                 return { auto: false, percent: Math.min(99, Math.max(0, percent)) };
             },
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -606,6 +606,13 @@ class VLMBatchedEngine(BaseEngine):
             tool_parser_type = _mlx_lm_infer(chat_template)
             if tool_parser_type is None:
                 return
+            
+            # Validation whitelist for dynamic loading
+            allowed_parsers = {"gemma", "llama", "mistral", "hermes", "phi", "qwen"}
+            if tool_parser_type not in allowed_parsers:
+                logger.warning(f"Unauthorized VLM tool parser type: {tool_parser_type}")
+                return
+
             try:
                 tool_module = importlib.import_module(
                     f"mlx_lm.tool_parsers.{tool_parser_type}"
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -1125,7 +1125,7 @@ def init_server(
         init_auth(global_settings.auth.secret_key, lambda: _server_state.global_settings)
 
     # Configure CORS middleware from settings
-    cors_origins = global_settings.server.cors_origins if global_settings else ["*"]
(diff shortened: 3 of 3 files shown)


jundot:
Thanks for the patch.

The dashboard.js fix is a real improvement. I'll commit that change separately with you as co-author so the credit lands on you.

For the server-side changes I want to flag a few things, and I'm going to close this PR after explaining them.

**CORS fallback**

The new fallback at `server.py:1125` only fires when `global_settings is None`. In practice every caller of `init_server` (cli, packaging app, tests) passes a populated `GlobalSettings`, so this branch never runs. The `cors_origins` default in `settings.py:116` is still `["*"]`, so this PR doesn't actually change behavior for real users.

If the fallback ever did run it wouldn't behave the way the description suggests. Starlette's `CORSMiddleware.allow_origins` does not support glob patterns. `"http://localhost:*"` is matched as a literal string, so no browser would ever match it and every CORS request would be blocked, including from the local dashboard. To get the intended behavior you'd need `allow_origin_regex` instead.

**VLM tool parser whitelist**

The whitelist `{"gemma", "llama", "mistral", "hermes", "phi", "qwen"}` doesn't line up with the actual parser names. `_infer_tool_parser` in `mlx_lm/tokenizer_utils.py` returns values like `gemma4`, `qwen3_coder`, `kimi_k2`, `glm47`, `minimax_m2`, `function_gemma`, `longcat`, `pythonic`, `mistral`, `json_tools`, and those are also the module names under `mlx_lm/tool_parsers/`. Only `mistral` overlaps with the whitelist, so this would log "Unauthorized VLM tool parser type" and silently disable tool calling for almost every VLM model.

Also, `_infer_tool_parser` already returns a hardcoded, internally-controlled set of strings, so there's no path for an untrusted value to reach `importlib.import_module` in the first place. The whitelist isn't necessary here.

Closing this for the reasons above. Thanks again for the dashboard fix, I'll get that landed in the next release.

RinZ27:
@jundot Thanks for the detailed feedback and for catching those nuances in the CORS middleware and parser naming.

You're absolutely right about `allow_origin_regex` — I overlooked Starlette's literal matching behavior there. Also, appreciate the clarification on the internal control of `_infer_tool_parser`; good to know that boundary is already well-defined.

Glad the dashboard fix was helpful. Looking forward to seeing it landed.

PR #1179, opened 2026-05-11
oq: preserve transpose/moveaxis permutation params in streaming sanitize

ijusss:
## Problem

Streaming sanitize discovery fails on any sanitizer that calls
`value.moveaxis(...)` or `v.transpose(perm)` as an **instance method**:
`_TrackedTensor` only exposed the wrappers at module level
(`mx.moveaxis`, `mx.transpose`), so the instance form raises
`AttributeError` and oQ drops to eager-sanitize fallback.

Eager loads the full model into Metal memory. For BF16 checkpoints
bigger than the unified-memory budget (~130 GB on M3 Ultra), this OOMs.
Qwen3.5-122B-A10B BF16 (233 GB) currently can't be quantized through
oQ for this reason.

Even when discovery did succeed, the writeback hardcoded the transform
parameters — `mx.transpose(arr)` was full-axis reverse,
`mx.moveaxis(arr, 2, 1)` was hardcoded for conv1d. Qwen3-VL
`VisionModel.sanitize` calls `v.transpose(0, 2, 3, 4, 1)` which is not
a reverse; the saved tensor came out with the wrong axis order.

## Fix

* Add `_TrackedTensor.moveaxis(src, dst)` and `.transpose(*axes)`
  instance methods.
* Record the actual `(src, dst)` / permutation tuple on
  `_TrackedTensor.axis`. Same for the existing `_fake_moveaxis` /
  `_fake_transpose` module-level shims.
* In `_DiscoveredPlan.pop`, read `info["axis"]` and replay the
  exact transform.

## Test results

| Model | Source | Output | Time | Engine | Inference |
|---|---:|---:|---:|---|---|
| Qwen3.5-35B-A3B-bf16 | 70 GB | 36 GB | ~40 s | VLMBatchedEngine | 75 tok/s decode, MTP active, vision OK |
| Qwen3.5-122B-A10B-bf16 | 233 GB | 124 GB | 147 s | VLMBatchedEngine | 40 tok/s decode, MTP active, vision OK |

Before the fix, 35B fell through eager (worked by luck — 70 GB fits in
Metal memory); 122B OOMed during eager fallback.

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -828,6 +828,50 @@ class _TrackedTensor:
     def astype(self, dtype):
         return _TrackedTensor(self.shape, dtype, list(self.sources), "astype")
 
+    def moveaxis(self, src, dst):
+        """Mirror ``mx.array.moveaxis`` for streaming discovery.
+
+        Stock mlx-vlm sanitizers call this as an instance method
+        (e.g. ``value.moveaxis(2, 1)`` to fix conv1d weight layout). Without
+        this method the streaming discovery falls back to eager sanitize,
+        which loads the entire model into Metal memory — fatal for >120 GB
+        checkpoints. The ``(src, dst)`` pair is stored in the ``axis`` slot
+        so the writeback phase can replay the exact permutation.
+        """
+        new_shape = list(self.shape)
+        moved = new_shape.pop(src)
+        new_shape.insert(dst, moved)
+        result = _TrackedTensor(
+            tuple(new_shape), self.dtype, list(self.sources), "moveaxis"
+        )
+        result.axis = (int(src), int(dst))
+        return result
+
+    def transpose(self, *axes):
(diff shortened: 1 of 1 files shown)


jundot:
Looked more carefully and this is already covered on current main by #1136 (6c282f4 `feat(oq): auto-build proxy for sensitivity when model exceeds RAM`) — the instance-method `_TrackedTensor.moveaxis` / `.transpose` are there, and the permutation is preserved by encoding it into the transform name (`transpose_0_2_3_4_1`, `moveaxis_2_1`) which `_DiscoveredPlan.pop` parses back. Different shape from your `axis` slot approach but functionally equivalent for the Qwen3-VL 5-axis and conv1d-(2,1) cases. Closing as obsolete. Thanks for the careful write-up either way.

PR #1180, opened 2026-05-11
mtp: runtime MTP head for mlx-vlm Qwen3.5-MoE (vision path)

ijusss:
## Problem

oMLX's MTP integration was mlx-lm only. Qwen3.5-MoE VLM checkpoints
with `mtp.*` weights either loaded as text-only via the LLM fallback
(losing vision) or silently dropped MTP. There was no way to get MTP
+ vision in the same model.

## Design

Zero modifications to the mlx-vlm decoder forward path. SSM rollback
on draft rejection uses the upstream
`LanguageModel.rollback_speculative_cache(...)`, which already exists
and consumes `gdn_states` — we just need to surface those alongside
`(logits, hidden)` on `return_hidden=True`.

## Files

* **New** `omlx/patches/mlx_vlm_mtp/qwen35_moe_vlm_runtime.py` —
  self-contained runtime patch (TextConfig.from_dict, LanguageModel
  __init__/__call__/sanitize, VLMModelAdapter pass-throughs, MTPModule
  built from mlx-vlm's own attention/MoE blocks).
* `omlx/patches/mlx_lm_mtp/batch_generator.py` — two new helpers:
  `_call_backbone` normalises 2-tuple (mlx-lm) vs 3-tuple (mlx-vlm)
  backbone return; `_rollback_after_reject` dispatches on model
  capability.
* `omlx/patches/mlx_lm_mtp/qwen35_model.py` — `_stack_per_expert`
  also stacks `.scales` / `.biases` so quantized per-expert MoE
  weights load on the MTP head.
* `omlx/utils/model_loading.py`, `omlx/engine/vlm.py`,
  `omlx/engine_pool.py` — small hook points (~20 lines) to register
  the runtime patch before model load and remove the legacy
  VLM→Batched shortcut.

## Test results

| Model | Engine | Decode tok/s (400 tok) | MTP accept | Vision |
|---|---|---:|---:|---|
| Qwen3.5-35B-A3B-oQ8-mtp | VLMBatchedEngine | **75.17** | 80–89 % | OK |
| Qwen3.5-35B-A3B-oQ8-mtp-text | BatchedEngine | 71.49 | 75–89 % | n/a |
| Qwen3.5-122B-A10B-oQ8-mtp | VLMBatchedEngine | **40.09** | 75–77 % | OK |

Greedy identity (sampler=None) preserved on the mlx-lm path — same
contract as PR 990's `test_mtp_generate_identity`.

## Notes

Depends on #1179 for streaming-sanitize support when converting
122B-class checkpoints. Smaller checkpoints (35B) work without it.

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -443,6 +443,17 @@ class VLMBatchedEngine(BaseEngine):
         from ..engine_core import AsyncEngineCore, EngineConfig
         from ..scheduler import SchedulerConfig
 
+        # Apply pre-load patches (MTP runtime patch, etc.) before the model
+        # is instantiated, so the patched ``__init__`` runs. ``maybe_apply``
+        # is a no-op when the model is incompatible.
+        try:
+            from ..utils.model_loading import maybe_apply_pre_load_patches
+            maybe_apply_pre_load_patches(
+                self._model_name, model_settings=self._model_settings
+            )
+        except Exception as e:
+            logger.debug(f"pre-load patches skipped: {e}")
+
         # Load VLM model on the global MLX executor to avoid blocking the event loop
         # while ensuring no concurrent Metal operations. See issue #85.
         from ..engine_core import get_mlx_executor
--- a/omlx/engine_pool.py
+++ b/omlx/engine_pool.py
@@ -607,17 +607,12 @@ class EnginePool:
             # encoder weights are ignored because the patched mtp_forward only
             # exists on the language model path. mtp_enabled was already
             # validated as mutually exclusive with dflash / turboquant in
-            # ModelSettings.__post_init__.
-            if (
-                model_settings is not None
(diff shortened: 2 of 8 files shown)


jundot:
Thanks for the careful split — capability-based dispatch on `rollback_speculative_cache` keeps the mlx-lm path untouched, and the `_call_backbone` / 3-tuple shape is the right seam. Bench numbers look consistent with the mlx-lm path's accept range. Your branch was on an older base so I cherry-picked d2eaf95 onto current main as f8fb5dd, your authorship preserved on the commit. The only conflict was `_stack_per_expert` in `omlx/patches/mlx_lm_mtp/qwen35_model.py` — main had a small early-return guard from #1147 and your `.scales`/`.biases` loop went on top of it. #1179 turned out to be already covered on main by #1136 so the 122B path works without it. Closing the PR — change lands in the next release.

PR #1184, opened 2026-05-11
Forward STT language to mlx-audio

### LINKED ISSUE #1117: STTEngine.transcribe drops the `language` parameter — Qwen3-ASR is forced into auto-detect, producing empty output for short audio
## Bug

`POST /v1/audio/transcriptions` accepts a `language` form field (per the OpenAI API spec) and even logs it, but the engine layer **never forwards it to the underlying mlx-audio model**. As a result, Qwen3-ASR (and any other mlx-audio STT model that takes a `language` arg) always runs in **auto-detect mode**, regardless of what the client sends.

This is mostly invisible for clean, longer audio — Qwen3-ASR auto-detect usually works. But for short / borderline audio (e.g. a real-time STT client sending fixed ~2s windows), auto-detect is **flaky** and the model often emits EOS as the first token → empty transcript. Resending the *exact same bytes* a moment later may transcribe correctly.

## Repro

Real-time STT client sending the same WAV twice (verified byte-identical, `Qwen3-ASR-1.7B-8bit`, language=zh):

```text
2026-05-08 11:52:06 - omlx.engine.stt - STT transcribe: file=tmpu6dbnrqp.wav (61896 bytes), language=zh
2026-05-08 11:52:06 - omlx.engine.stt - STT transcribe done: 0.11s, 0 chars output
2026-05-08 11:52:17 - omlx.engine.stt - STT transcribe: file=tmp2iv57o3i.wav (61896 bytes), language=zh
2026-05-08 11:52:17 - omlx.engine.stt - STT transcribe done: 0.12s, 8 chars output
```

The log line correctly reports `language=zh`, but mlx-audio receives `language=None` (its default) so it goes to auto-detect.

## Root Cause

[`omlx/engine/stt.py`](https://github.com/jundot/omlx/blob/main/omlx/engine/stt.py#L171-L231) — `STTEngine.transcribe`:

```python
async def transcribe(
    self,
    audio_path: str,
    language: Optional[str] = None,   # <-- received
    **kwargs,
) -> Dict[str, Any]:
    ...
    logger.info("STT transcribe: ..., language=%s", ..., language)   # <-- only logged

    def _transcribe_sync():
        result = model.generate(audio_path, **kwargs)                # <-- language NOT forwarded
```

`language` is consumed by the log line and the result-dict fallback (`raw_lang = language` at L240) but is **never passed to `model.generate()`**.

In mlx-audio's `Qwen3ASR._build_prompt` ([qwen3_asr.py L864-894](https://github.com/Blaizzy/mlx-audio/blob/main/mlx_audio/stt/models/qwen3_asr/qwen3_asr.py#L864-L894)) the two paths diverge significantly:

| input language | assistant prefix | first generated token must be |
|---|---|---|
| `\"Chinese\"` (or any supported lang) | `language Chinese<asr_text>` | a content character |
| `None` (auto-detect, current omlx behaviour) | empty | the literal token `language` — or EOS, on a coin-flip's worth of logit margin |

With temperature=0 greedy decoding, marginal logit differences (8-bit quantization noise + Metal SDPA reduction order across calls) can flip the first-token decision in auto-detect mode → empty transcript on one call, normal output on the next, for the same bytes. Forcing the prompt suffix via an explicit `language` removes the ambiguity.

## Secondary issue: ISO 639-1 → full-name mapping

Even if the parameter were forwarded, `Qwen3-ASR` expects English full names (`Chinese`, `English`, `German`, …) per its [README](https://github.com/Blaizzy/mlx-audio/blob/main/mlx_audio/stt/models/qwen3_asr/README.md), but the OpenAI `/v1/audio/transcriptions` spec uses **ISO 639-1** (`zh`, `en`, …). `_build_prompt`'s lookup falls back to the literal string when not found:

```python
supported_lower = {lang.lower(): lang for lang in supported}
lang_name = supported_lower.get(language.lower(), language)   # \"zh\" → \"zh\" (off-distribution)
```

So `language=zh` would still degrade quality even if forwarded. omlx is the right place to do this mapping (it is the OpenAI-compat adapter layer).

## Suggested Fix

Two small changes in `omlx/engine/stt.py`:

```python
# 1) ISO 639-1 → Qwen3-ASR full-name mapping (top of module)
_ISO_TO_QWEN3_ASR_LANG = {
    \"zh\": \"Chinese\", \"yue\": \"Cantonese\",
    \"en\": \"English\", \"de\": \"German\", \"es\": \"Spanish\",
    \"fr\": \"French\",  \"it\": \"Italian\", \"pt\": \"Portuguese\",
    \"ru\": \"Russian\", \"ko\": \"Korean\",  \"ja\": \"Japanese\",
}

# 2) Forward language conditionally inside _transcribe_sync
def _transcribe_sync():
    gen_kwargs = dict(kwargs)
    if language is not None:
        # Pass through if already a full name; otherwise map ISO 639-1.
        gen_kwargs[\"language\"] = _ISO_TO_QWEN3_ASR_LANG.get(
            language.lower(), language
        )
    result = model.generate(audio_path, **gen_kwargs)
    ...
```

Use the **conditional-kwarg pattern** (don't pass `language=None` explicitly) — same pattern mlx-audio adopted in their own server: [Blaizzy/mlx-audio#545](https://github.com/Blaizzy/mlx-audio/pull/545).

## Prior Art (mlx-audio)

mlx-audio's own server hit the exact same shape of bug and fixed it the same way:

- [Blaizzy/mlx-audio#544](https://github.com/Blaizzy/mlx-audio/issues/544) — Realtime WebSocket: `language=None` crashes Qwen3-ASR `_build_prompt()`
- [Blaizzy/mlx-audio#545](https://github.com/Blaizzy/mlx-audio/pull/545) (merged) — fix: avoid passing `language=None` to STT model `generate()`
- [Blaizzy/mlx-audio#546](https://github.com/Blaizzy/mlx-audio/issues/546) / [#541](https://github.com/Blaizzy/mlx-audio/pull/541) / [#547](https://github.com/Blaizzy/mlx-audio/pull/547) — auto-detect mode design (the path omlx is currently always taking)

## Environment

- omlx v0.3.8
- mlx-audio (current main)
- Model: `mlx-community/Qwen3-ASR-1.7B-8bit`
- Apple Silicon

## Workaround

Until fixed, clients can't influence Qwen3-ASR's language path through omlx — `language` is silently dropped. The only mitigation is client-side: VAD / audio length thresholds and retry on empty result.

---

Happy to send a PR with the two-line fix + ISO mapping table + a unit test that asserts `language` reaches `model.generate`.

Bortlesboat:
## Summary
- normalize OpenAI-style ISO language codes before passing them to mlx-audio STT models
- forward `language` to `model.generate()` only when a non-empty value is provided, preserving auto-detect for blank values
- add STTEngine regression coverage for ISO mapping and empty language handling

Fixes #1117

## Testing
- python -m ruff check omlx/engine/stt.py tests/test_audio_stt.py
- python -m compileall -q omlx/engine/stt.py tests/test_audio_stt.py
- focused local STTEngine harness with a stubbed MLX core/executor

Note: full pytest was not run locally because this Windows workspace does not have Apple MLX/MLX-LM installed.

### DIFF
--- a/omlx/engine/stt.py
+++ b/omlx/engine/stt.py
@@ -11,7 +11,7 @@ when mlx-audio is not installed.
 import asyncio
 import gc
 import logging
-from typing import Any, Dict, Optional
+from typing import Any
 
 import mlx.core as mx
 
@@ -19,6 +19,33 @@ from ..engine_core import get_mlx_executor
 from .base import BaseNonStreamingEngine
 
 logger = logging.getLogger(__name__)
+
+
+_ISO_TO_QWEN3_ASR_LANG: dict[str, str] = {
+    "zh": "Chinese",
+    "yue": "Cantonese",
+    "en": "English",
+    "de": "German",
+    "es": "Spanish",
+    "fr": "French",
+    "it": "Italian",
+    "pt": "Portuguese",
+    "ru": "Russian",
+    "ko": "Korean",
+    "ja": "Japanese",
+}
+
+
+def _normalize_stt_generate_language(language: str | None) -> str | None:
+    """Map OpenAI-style ISO codes to language names accepted by mlx-audio."""
+    if language is None:
+        return None
+
+    normalized = language.strip()
+    if not normalized:
+        return None
+
+    return _ISO_TO_QWEN3_ASR_LANG.get(normalized.lower(), normalized)
 
 
 # ---------------------------------------------------------------------------
@@ -72,7 +99,7 @@ def _validate_stt_processor(model_name: str, model: Any) -> None:
     # to None when WhisperProcessor.from_pretrained() failed on load.
     if not hasattr(model, "_processor"):
         return
-    if getattr(model, "_processor") is not None:
+    if model._processor is not None:
         return
     raise RuntimeError(_missing_processor_hint(model_name))
 
@@ -171,9 +198,9 @@ class STTEngine(BaseNonStreamingEngine):
     async def transcribe(
         self,
         audio_path: str,
-        language: Optional[str] = None,
+        language: str | None = None,
         **kwargs,
-    ) -> Dict[str, Any]:
+    ) -> dict[str, Any]:
         """
         Transcribe an audio file.
 
@@ -228,7 +255,12 @@ class STTEngine(BaseNonStreamingEngine):
         def _transcribe_sync():
             # Call model.generate() directly instead of
             # generate_transcription() which writes files to disk.
-            result = model.generate(audio_path, **kwargs)
+            gen_kwargs = dict(kwargs)
+            generate_language = _normalize_stt_generate_language(language)
+            if generate_language is not None:
+                gen_kwargs["language"] = generate_language
+
+            result = model.generate(audio_path, **gen_kwargs)
 
             # result is typically an STTOutput dataclass with:
             # text, segments, language, total_time, etc.
@@ -286,7 +318,7 @@ class STTEngine(BaseNonStreamingEngine):
                     lambda: (mx.synchronize(), mx.clear_cache()),
                 )
 
-    def get_stats(self) -> Dict[str, Any]:
+    def get_stats(self) -> dict[str, Any]:
         """Get engine statistics."""
         return {
             "model_name": self._model_name,
--- a/tests/test_audio_stt.py
+++ b/tests/test_audio_stt.py
@@ -10,11 +10,11 @@ required. Integration tests (marked @pytest.mark.slow) need a real model.
 
 import io
 import wave
+from types import SimpleNamespace
 from unittest.mock import AsyncMock, MagicMock, patch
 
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for picking this up, the core fix is exactly right and I'll merge as is.

One thing I missed in the original issue: this maps to Capitalized full-names ("Chinese"), which works for Qwen3-ASR but Whisper expects ISO codes. Whisper's `TO_LANGUAGE_CODE` normalizer only matches lowercase keys, so "Chinese" falls through and `<|Chinese|>` token lookup fails. I'll switch the table to lowercase ("chinese") in a follow-up on main: Whisper normalizes that to "zh", and Qwen3-ASR's `_build_prompt` already lowercases for lookup, so both paths work. Will land in the next release.

PR #1186, opened 2026-05-11
fix(mtp): skip MoE expert stacking when MTP layer is dense (MTPLX)

jasonpaulso:
## Summary

Fix `KeyError: 'language_model.mtp.layers.0.mlp.experts.0.gate_proj.weight'` when loading MTPLX-format MoE checkpoints (e.g. [`samuelfaj/Ornstein3.6-35B-A3B-SABER-6bit-MTPLX-Optimized-Speed`](https://huggingface.co/samuelfaj/Ornstein3.6-35B-A3B-SABER-6bit-MTPLX-Optimized-Speed)) under the batched LLM engine.

The `_patch_qwen3_5_moe` sanitize assumed every MTP layer carries MoE expert tensors in one of two layouts: fused (`experts.gate_up_proj`) or per-expert (`experts.{e}.{n}.weight`). MTPLX checkpoints (produced by `lightning-mlx convert-mtplx tensor-layout gate`) ship the MTP layer as a single dense MLP — the router/gate and `shared_expert` weights are present, but there are no per-expert tensors. The old code fell through to `_stack_per_expert`, which crashed.

Add a `has_experts` short-circuit: if the MTP layer has no `experts.*` keys, pass the dense weights through untouched. Backbone MoE unfusing is unaffected.

## Repro

Pre-fix:
```
omlx.engine_pool - INFO - Loading model: Ornstein3.6-35B-A3B-SABER-6bit-MTPLX-Optimized-Speed
omlx.admin.benchmark - ERROR - Benchmark error: 'language_model.mtp.layers.0.mlp.experts.0.gate_proj.weight'
...
File "omlx/patches/mlx_lm_mtp/qwen35_model.py", line 654, in <listcomp>
    weights.pop(f"{prefix}.experts.{e}.{n}.weight")
KeyError: 'language_model.mtp.layers.0.mlp.experts.0.gate_proj.weight'
```

Post-fix: model loads and benchmarks run cleanly.

## Why latent until now

Surfaced by [`6fe6d70`](../commit/6fe6d70) (`fix(mtp): MoE oQ-mtp variants for Qwen3.5 A3B and DeepSeek-V4`), which switched the patch's idempotency check from `hasattr(cls, "_omlx_mtp_patched")` to `"_omlx_mtp_patched" in cls.__dict__`. Before that change the MoE sub-patch was silently skipped on `qwen3_5_moe.Model` because its parent `qwen3_5.Model` already carried the marker (via inheritance) — so stock mlx_lm sanitize ran instead, which ignores MTP layers entirely. After `6fe6d70` the sub-patch installs as intended, exposing the dense-MTP gap.

## Test plan

- [x] `samuelfaj/Ornstein3.6-35B-A3B-SABER-6bit-MTPLX-Optimized-Speed` loads cleanly and runs benchmarks (`mtp_enabled=False`).
- [x] No regression for per-expert MoE MTP models (Qwen3.5 A3B / DeepSeek-V4) which exercise the same `_stack_per_expert` path when expert tensors are present.
- [x] Existing test suite passes (113 passed; the unrelated `test_all_valid_benchmarks` failure on `main` is independent).

### DIFF
--- a/omlx/patches/mlx_lm_mtp/qwen35_model.py
+++ b/omlx/patches/mlx_lm_mtp/qwen35_model.py
@@ -691,10 +691,14 @@ def _patch_qwen3_5_moe() -> None:
                 # MTPLX (lightning-mlx convert) layout: MTP layer is dense
                 # — no per-expert tensors to stack/unfuse. The router/gate
                 # and shared_expert weights are passed through unchanged.
-                has_experts = any(
-                    k.startswith(f"{prefix}.experts.") for k in new_weights
+                # Probe by layout sentinel (O(1)) rather than scanning all
+                # weight keys — matters on multi-thousand-tensor checkpoints.
+                sentinel = (
+                    f"{prefix}.experts.gate_up_proj"
+                    if mtp_is_fused
+                    else f"{prefix}.experts.0.gate_proj.weight"
                 )
-                if not has_experts:
+                if sentinel not in new_weights:
                     continue
                 if mtp_is_fused:
                     _unfuse_experts(new_weights, prefix)
--- a/omlx/patches/mlx_lm_mtp/qwen35_model.py
+++ b/omlx/patches/mlx_lm_mtp/qwen35_model.py
@@ -671,7 +671,7 @@ def _patch_qwen3_5_moe() -> None:
         for l in range(self.language_model.args.num_hidden_layers):
             _unfuse_experts(new_weights, f"language_model.model.layers.{l}.mlp")
 
-        # MTP layers: fused (Qwen3.6) or per-expert (Qwen3.5). Detect once.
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the careful write-up — the 6fe6d70 trace matches what I see (verified the idempotent-guard switch), and the sentinel-based short-circuit is the right shape. Cherry-picked both commits onto current main as c19a4f5 + 52c80a8, your authorship preserved on both. Only meaningful change during the rebase: kept main's existing `if not has_any_mtp` outer guard from #1147, and folded your dense test into the existing fixture-based `TestQwen35MoeSanitize` class so per-expert / fused / dense paths share the same setup. All 5 sanitize tests pass. Closing the PR — change lands in the next release.

PR #1191, opened 2026-05-11
fix(admin/benchmark): forward SpecPrefill model settings to engine

### LINKED ISSUE #1145: [Bug] v0.3.8 - SpecPrefill model settings are not loaded and always fallback to default values such as DEFAULT_THRESHOLD=8192
**Describe the bug**
When settings Threshold for SpecPrefill via model settings, it should load that model settings and use but it can't load and always use the default value e.g. DEFAULT_THRESHOLD=8192

**To Reproduce**
1. Go to any model and turn on SpecPrefill with correct draft model.
2. Set SpecPrefill threshold to 512 for the model.
3. Click save.
4. Unload and reload the model.
5. Run benchmark for the model.
6. You would see that SpecPrefill only start after Threshold 8192 instead of 512.

**Expected behavior**
SpecPrefill is expected to increase prompt processing tps when above threshold 512. but only start doing spec prefill when above default threshold 8192.

**Screenshots**

Image

Image

 - macOS Version: 26.4.1
 - oMLX Version 0.3.8

**Additional context**
It may already happen in previous version as well which I did not test. But definitely happen with stable release 0.3.8

SuperMarioYL:
## Summary

Fixes #1145.

The dashboard benchmark path read `ModelSettings.specprefill_enabled` only
to tag `run.experimental_features`, but never forwarded the matching
`specprefill_keep_pct` / `specprefill_threshold` overrides to
`engine.stream_generate` (in `_run_single_test`) or to
`engine_core.add_request` (in `_run_batch_test`).

The chat-completion path already loads these from `ModelSettingsManager`
(see [`server.py` around the per-request override block](https://github.com/jundot/omlx/blob/main/omlx/server.py)),
so end-users see SpecPrefill engage at their configured threshold during
normal chat — but a benchmark for the same model silently falls back to
`DEFAULT_THRESHOLD=8192` and reports numbers that don't reflect the
configured threshold (e.g. the issue's reproducer: threshold set to 512,
benchmark only engages SpecPrefill above 8192).

## What changed

`omlx/admin/benchmark.py`:

- Build a `specprefill_kwargs` dict once at run-start, alongside the
  existing experimental-features snapshot.
- Add an optional `specprefill_kwargs` param to `_run_single_test` and
  `_run_batch_test` and forward via `**kwargs` into
  `engine.stream_generate(...)` / `engine_core.add_request(...)`.
- Pass `specprefill_kwargs` through to the warmup call too, so the
  warmup path doesn't load a different code branch than the measured
  runs.

When the model has no SpecPrefill settings the dict stays empty and no
`specprefill_*` kwarg leaks into `stream_generate` — preserving the
default engine path for models the user has not opted in for.

`tests/test_benchmark.py`:

- New `TestSpecPrefillSettingsLoad` class with three regression tests
  that mock `stream_generate` / `add_request` and assert the threshold
  and keep_pct round-trip end-to-end through the benchmark helpers, plus
  a no-leakage case for models without SpecPrefill enabled.

## Test plan

- [x] `python -m py_compile omlx/admin/benchmark.py tests/test_benchmark.py`
- [x] Isolated test run of the four `TestSpecPrefillSettingsLoad` cases
      (stubbed `mlx.core` to import `omlx.admin.benchmark` standalone on
      a Linux dev machine without the MLX runtime) — all pass on the
      branch and the assertions fail on `main` as expected.
- [ ] Full `pytest tests/test_benchmark.py` in the project MLX
      environment (not runnable in this dev environment without the
      Apple Silicon MLX runtime).

## Reviewer kill-question (anticipated)

> "Why not push the settings-manager load into `engine.stream_generate`
> itself?"

The engine layer is intentionally settings-manager-agnostic — every
caller (chat-completion, benchmark, tests) forwards the per-request
override. Pushing settings-manager into the engine would duplicate the
existing forward-flow and add a new dependency cycle from
`omlx/engine/*` into `omlx.model_settings`. The benchmark route is the
only layer that was missing the forward step, so the fix lives there.

### DIFF
--- a/omlx/admin/benchmark.py
+++ b/omlx/admin/benchmark.py
@@ -304,8 +304,17 @@ async def _run_single_test(
     prompt: str,
     max_tokens: int,
     pp_len: int,
+    specprefill_kwargs: Optional[dict] = None,
 ) -> dict:
-    """Run a single request benchmark test and return metrics."""
+    """Run a single request benchmark test and return metrics.
+
+    Args:
+        specprefill_kwargs: Optional per-request SpecPrefill overrides
+            (e.g. ``specprefill``, ``specprefill_keep_pct``,
+            ``specprefill_threshold``) forwarded to ``engine.stream_generate``
+            so the model's configured threshold (not the engine default) takes
+            effect during benchmarking.
+    """
     # Reset peak memory tracking
     try:
         mx.reset_peak_memory()
@@ -323,6 +332,7 @@ async def _run_single_test(
         max_tokens=max_tokens,
         temperature=0.0,
         top_p=1.0,
+        **(specprefill_kwargs or {}),
     ):
         # Detect first generated token via completion_tokens count,
         # not new_text. Some models (e.g. Harmony/gpt-oss) produce
@@ -412,6 +422,7 @@ async def _run_batch_test(
     prompt_tokens: int,
     max_tokens: int,
     batch_size: int,
+    specprefill_kwargs: Optional[dict] = None,
 ) -> dict:
     """Run a continuous batching benchmark test.
 
@@ -423,6 +434,10 @@ async def _run_batch_test(
                  all entries are identical. For different-prompt tests, each
                  has a unique UUID prefix.
         prompt_tokens: Number of prompt tokens per request (for pp TPS calc).
+        specprefill_kwargs: Optional per-request SpecPrefill overrides
+            forwarded to ``engine_core.add_request`` so the model's configured
+            threshold (not the engine default) takes effect during batch
+            benchmarking.
     """
     from ..request import SamplingParams
 
@@ -436,6 +451,8 @@ async def _run_batch_test(
         top_p=1.0,
     )
 
+    add_request_kwargs = dict(specprefill_kwargs or {})
+
     async def _single_request(prompt: str) -> dict:
         """Run a single request within the batch."""
         start = time.perf_counter()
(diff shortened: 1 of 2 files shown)


SuperMarioYL:
Rebased onto latest `main` to clear the merge conflict. The conflict was pure base-drift in `omlx/admin/benchmark.py`: `main` refactored the experimental-flag snapshot into `_detect_experimental_features(model_settings)` and made the warmup call diffusion-aware (`warmup_max_tokens`). I kept both of those and re-applied the SpecPrefill forwarding on top of the new `model_settings` variable (dropping my now-redundant inline flag tagging, since `_detect_experimental_features` already covers it). `tests/test_benchmark.py` was an additive both-sides conflict — kept the new `TestSanitizeUploadError` and my `TestSpecPrefillSettingsLoad`. All 64 benchmark tests pass locally. Ready for review.

SuperMarioYL:
Rebased onto upstream/main to clear the merge conflict. The conflict was in `benchmark.py` (upstream moved a lot of the surrounding structure) and `tests/test_benchmark.py`; the SpecPrefill forwarding is re-applied on the new structure in 9b392a5d:

- `_run_single_test` and `_run_batch_test` both accept `specprefill_kwargs` and forward it to `engine.stream_generate` / `engine_core.add_request`.
- `run_benchmark` builds `specprefill_keep_pct` / `specprefill_threshold` from the model settings when SpecPrefill is enabled, and passes them through to the single/batch test calls.
- `warmup` forwards `specprefill_kwargs` too.

Verified issue #1145 is still open and the three call sites still lack the kwargs on current main, so this is still live. Happy to address any further feedback.

PR #1192, opened 2026-05-11
fix(bench): sanitize upload error so CF challenge HTML doesn't leak to UI

jasonpaulso:
## Summary

The Community Benchmark Upload panel on the Throughput Benchmark tab has been rendering the entire Cloudflare "Just a moment..." HTML challenge interstitial as the error message for every prompt-length row, for at least the past several weeks. Reproduced against `omlx.ai/api/benchmarks` from a clean network — every POST returns:

```
HTTP/2 403
cf-mitigated: challenge
content-type: text/html; charset=UTF-8
content-length: 5290
...
<!DOCTYPE html><html lang="en-US"><head><title>Just a moment...</title>...
```

The current error path in `_upload_to_omlx_ai` (`omlx/admin/benchmark.py:540-555`) falls back to `resp.text` when `resp.json()` raises, and sends the entire body to the dashboard via SSE. `_bench.html:321` renders that string via `x-text` — Alpine treats it as a text node, so the literal HTML markup ends up in the red-x cell next to each `pp{N}` row.

This PR does **not** fix the upload itself — that's an omlx.ai server-side configuration issue (filed separately as an issue). It fixes the UI so the panel surfaces a single actionable line instead of a 5KB markup blob.

## What changed

New helper `_sanitize_upload_error(resp)` in `omlx/admin/benchmark.py` with this resolution order:

1. **Cloudflare challenge** — header `cf-mitigated: challenge` is authoritative; a body sniff for `"just a moment"` / `"cf-chl"` covers edge transports / proxies that strip the header. Returns: `"Upload blocked by Cloudflare (HTTP {status}). This is a server-side issue with omlx.ai — retry later or report it to the maintainer."`
2. **JSON envelope** — the normal API error shape. Extracts `error` / `detail` / `message` if present, truncated to 300 chars.
3. **Plain-text body** — short responses pass through (capped at 300 chars). HTML-shaped bodies collapse to a one-line hint: `"HTTP {status} — unexpected non-JSON response (N bytes)"`.
4. **Empty body** — falls back to the bare HTTP status string (`"HTTP {status}"`).

Full response body (truncated to 500 chars) is logged at DEBUG level so it remains available for offline analysis, but never reaches the dashboard.

## Test plan

- [x] 7 new tests in `TestSanitizeUploadError` cover each branch: CF-via-header, CF-via-body-sniff, JSON error, JSON detail, generic HTML 5xx, short plain text, empty body.
- [x] `pytest tests/test_benchmark.py` — 47 passed (was 40, added 7).
- [x] Manually verified against the live endpoint: `curl -X POST https://omlx.ai/api/benchmarks -d '{}'` returns 403 + CF interstitial; sanitizer maps that to a one-line message in the dashboard.

## Why this is a partial fix

The underlying upload-failure is a server-side issue. Cloudflare is challenging all non-browser POSTs to `/api/benchmarks` — UA spoofing (Safari, Chrome, python-requests) all return identical 403 + challenge responses. The actual fix is on the omlx.ai side: allow API traffic, add API-key auth, or configure Bot Fight Mode exceptions for the benchmark endpoint. This PR just stops the failure from being unreadable in the meantime.

### DIFF
--- a/omlx/admin/benchmark.py
+++ b/omlx/admin/benchmark.py
@@ -378,6 +378,51 @@ def _clean_model_name(model_id: str, quantization: str) -> str:
     return name.strip("-_ ")
 
 
+def _sanitize_upload_error(resp: Any) -> str:
+    """Extract a user-presentable error string from a failed upload response.
+
+    Avoids dumping raw HTML bodies (e.g. Cloudflare's "Just a moment..."
+    challenge interstitial) into the dashboard's red-x error column.
+    Detects CF mitigation specifically so users get actionable context
+    instead of a 5KB markup blob.
+
+    Resolution order:
+    1. Cloudflare challenge — header ``cf-mitigated: challenge`` is
+       authoritative; a body sniff for "just a moment" / "cf-chl" covers
+       edge transports that strip the header.
+    2. JSON envelope — the omlx.ai API's normal error shape; extract
+       ``error`` / ``detail`` / ``message`` if present, truncated.
+    3. Plain-text body — short responses only; HTML-looking bodies are
+       collapsed to a one-line "non-JSON response (N bytes)" hint.
+    4. Fallback to the bare HTTP status code.
+    """
+    headers = getattr(resp, "headers", {}) or {}
+    cf_mitigated = str(headers.get("cf-mitigated", "")).lower()
+    body = getattr(resp, "text", "") or ""
+    status = getattr(resp, "status_code", "?")
+
+    body_head = body[:512].lower()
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the clean writeup — repro steps, partial-fix scoping, and the helper layout all read well.

The four-branch resolution (CF header → body sniff → JSON envelope → plain text → status fallback) covers the realistic cases, and dropping the raw body to DEBUG keeps it retrievable without polluting the panel. Tests cover each branch + the header-stripping proxy case, which is the one I'd have asked for.

Will merge once CI is green.

PR #1194, opened 2026-05-11
feat(admin): server restart endpoint + Settings > Server Restart button

jasonpaulso:
## Summary

Adds a web-UI **Restart** action under Settings > Server. Lets users managing remote oMLX instances bounce the server without reaching for the menubar app or SSH.

Stop and Start remain menubar-only — Restart is the 80% case worth wiring up, and the menubar app's `ServerManager` already handles the respawn half of the dance.

## Architecture

The menubar app's `ServerManager._health_check_loop` already auto-restarts the child process when it exits unexpectedly (backoff 5s, cap 3 within a 60s stable window). This PR piggy-backs on that: the admin endpoint schedules `SIGTERM` on itself and lets the supervisor's existing crash-recovery path do the respawn. No new IPC channel, no menubar daemon changes beyond a 2-line env-var marker.

```
1. Client POSTs /admin/api/server/restart
2. Endpoint returns 202 + schedules os.kill(getpid(), SIGTERM) via call_later(0.5)
3. FastAPI flushes the 202 to the client
4. SIGTERM fires → server exits
5. Menubar's _health_check_loop sees process.poll() != None
6. _try_auto_restart fires (5s backoff for first attempt)
7. Server respawns, /health returns 200
8. Client (which has been polling /health) observes the down→up transition
9. window.location.reload()
```

Total downtime ≈ 7s.

## Changes

| File | What | LOC |
|---|---|---|
| `packaging/omlx_app/server_manager.py` | Inject `OMLX_SUPERVISED=menubar` + `OMLX_SUPERVISOR_PID` into child env | +9 |
| `omlx/admin/routes.py` | `POST /admin/api/server/restart` endpoint (admin auth) | +55 |
| `omlx/admin/templates/dashboard/_settings.html` | New row in the Server section | +26 |
| `omlx/admin/static/js/dashboard.js` | State machine + `/health` polling | +107 |
| `tests/test_admin_restart.py` | 3 tests | +78 |

## Endpoint contract

**`POST /admin/api/server/restart`** (admin auth)

```json
202 Accepted
{
  "status": "restarting",
  "supervisor": "menubar",
  "expected_downtime_seconds": 7
}

503 Service Unavailable  // OMLX_SUPERVISED env var unset
{
  "detail": "Server is not running under a supervisor that can respawn it.
            Restart unavailable — use the menu bar app's Restart, or restart
            from your shell."
}
```

The 503 branch protects plain `omlx serve` / `pip install`-style deployments from being killed with no respawn path. The env-var label (rather than a bare `1`) leaves room for future supervisor types — a launchd / systemd follow-up could just set `OMLX_SUPERVISED=launchd` and reuse the endpoint as-is.

## Client state machine

`restartServer.status` transitions:

```
idle → restarting → waiting → idle (success → window.location.reload())
                         └─→ error  (60s timeout or non-200 health)
     └─→ unsupported  (503 from server)
     └─→ error  (4xx/5xx other than 503)
```

`_restartServerPoll` requires seeing the server actually go down (≥1 failed `/health` request) before it'll accept a 200 as success. Guards against the "supervisor hasn't killed the process yet" false-positive where polling lands on the dying old instance before SIGTERM completes.

## What this doesn't do

- **No Stop / Start.** Stop without supervisor-side IPC would re-trigger the auto-restart immediately; Start has a bootstrap problem (web UI doesn't exist when the server is down).
- **No backoff suppression.** The supervisor's 5s crash-recovery backoff is the same path our intentional restart uses, so user-initiated restarts also wait ~5-7s. Suppressing that would need an IPC channel telling the supervisor "this exit was intentional, respawn immediately."
- **No headless / launchd path.** Out of scope; could be a follow-up that adds another `OMLX_SUPERVISED=launchd` branch.

## Auto-restart safety net

`ServerManager` caps at `_max_auto_restarts=3` per `_stable_threshold=60s` window. Translation for this PR:
- Single Restart click: always succeeds (server runs >60s stable between clicks, counter resets)
- 3 rapid Restarts within 60s: the 3rd succeeds, the 4th would refuse and surface ERROR status in the menubar
- Server crashing during startup: counts against the cap, exponential backoff

This is fine for normal use — surfaced here for future readers.

## Test plan

- [x] `pytest tests/test_admin_restart.py` — 3 passed.
- [x] `pytest tests/test_admin_restart.py tests/test_benchmark.py tests/test_model_settings.py` — 104 passed (no regressions in adjacent admin / settings code).
- [x] `node --check omlx/admin/static/js/dashboard.js` clean.
- [x] Dashboard template renders end-to-end.
- [x] End-to-end DMG smoke — rebuilt, reinstalled, verified Restart cycles cleanly on a live install (post-build force-push includes a UI nit fix to keep the button height/width fixed across state transitions so the row layout no longer shifts during "Waiting for server…").


jundot:
Thanks Jason, this is a clean approach. Piggy-backing on the existing supervisor crash-recovery path keeps the surface area small, and the OMLX_SUPERVISED gate is a nice safety against plain `omlx serve` getting killed with no respawn. The down-then-up polling check is a particularly good touch.

Merging now. I'll push a small follow-up to main to:
- move the inline `import os`/`import signal` in `_schedule_self_terminate` to file-top (project convention is no inline imports)
- drop the unused `OMLX_SUPERVISOR_PID` env injection (no consumer yet)
- run `scripts/normalize_i18n.py` to sync the six non-English locale files for the new `settings.server.restart_*` keys

Ships in the next release.

PR #1196, opened 2026-05-12
Add inline chat rename action to sidebar history

beamivalice:
Add a simple rename button.

Screenshot 2569-05-12 at 11 24 20

Screenshot 2569-05-12 at 11 24 26

PR #1201, opened 2026-05-12
jinja2.exceptions.TemplateSyntaxError: unexpected '}' error fix (omlx/admin/templates/chat.html)

felk-dev:
jinja2.exceptions.TemplateSyntaxError: unexpected '}' error fix


jundot:
Thanks for catching this. Confirmed — #1196 (63dc911) ran an HTML formatter over chat.html that split the `{{ api_key | tojson }}` tag across lines, which Jinja rejects since `}}` can't have whitespace in between. Admin chat 500s on main right now. Merging.

PR #1203, opened 2026-05-12
fix(oq): patch mlx-lm during auto-building a sensitivity proxy

### LINKED ISSUE #1199: Auto-built sensitivity proxy still comes without MTP tensors?
Hi.

Following your PR #1136.

It still fails with the same error:

```
2026-05-12 12:20:21,427 - omlx.oq - ERROR - Sensitivity proxy load failed (Native MTP is enabled for this model but the converted weights are missing the mtp.* tensors. Default mlx-lm converters strip them; you need a converter that preserves MTP weights (or a Qwen3.6 / DeepSeek-V4 checkpoint that already preserves them). To recover without re-converting, open the model's settings in the oMLX admin UI and toggle 'Native MTP' off, then retry.)
2026-05-12 12:20:21,431 - omlx.admin.oq_manager - ERROR - oQ quantization failed: Qwen3.6-35B-A3B -> oQ5: sensitivity measurement produced no scores. Check the preceding log lines for the root cause (model load, calibration data, or layer discovery), and either fix it or pass an explicit sensitivity_model_path.
```

Even after this step succeeded:

```
2026-05-12 12:20:01,461 - omlx.oq - WARNING - oQ5: model size (71.9 GB) exceeds 80% of system RAM (68.7 GB). Auto-building a uniform 4-bit proxy on disk so sensitivity measurement stays data-driven
```

I've modified the source, saved the artifact, listed all the tensors and there are no `mtp.*` ones inside indeed:

```py
for filename in os.listdir(folder):
    if not filename.endswith(".safetensors"):
        continue

    filepath = os.path.join(folder, filename)
    
    with safetensors.safe_open(filepath, framework="pt", device="cpu") as f:
        for key in f.keys():
            if "mtp." in key:
                print(key)
```

Are we sure that this import:

https://github.com/jundot/omlx/blob/8d65e3d9ede2576ee073828c70196ff906a2eca5/omlx/oq.py#L3106

works through the monkey patching and doesn't strip out MTP tensors?

deepsweet:
Fixes #1199.

## Issue

The unpatched version of `mlx_lm.convert()` was used, leading to:

```
2026-05-12 12:20:21,427 - omlx.oq - ERROR - Sensitivity proxy load failed (Native MTP is enabled for this model but the converted weights are missing the mtp.* tensors. Default mlx-lm converters strip them; you need a converter that preserves MTP weights (or a Qwen3.6 / DeepSeek-V4 checkpoint that already preserves them). To recover without re-converting, open the model's settings in the oMLX admin UI and toggle 'Native MTP' off, then retry.)
2026-05-12 12:20:21,431 - omlx.admin.oq_manager - ERROR - oQ quantization failed: Qwen3.6-35B-A3B -> oQ5: sensitivity measurement produced no scores. Check the preceding log lines for the root cause (model load, calibration data, or layer discovery), and either fix it or pass an explicit sensitivity_model_path.
```

Because the auto-built affine 4-bit sensitivity model comes out with stripped out `mtp.*` tensors.

## Fix

Wrap `_build_proxy_for_sensitivity` with `apply_mlx_lm_mtp_patch` monkey-patching similarly to many other oQ helpers.

## Test

- `test_happy_path_with_active_patch` – Full happy path: patch applied, MTP toggled True→False, all kwargs forwarded, correct return value and working_dir.
 - `test_state_restored_on_error` – Parametrized (False/True) to verify that `finally` restores the exact prior state on exception.                   
 - `test_patch_returns_false_no_toggle` – Patch returns `False`, no toggle, convert() still runs.
 - `test_import_fails_graceful_degradation` – Patch import fails, proceeds without MTP gating, convert() still runs.

Probably way too verbose, the last 2 tests could be removed if you like.

## Also

My concerns from [here](https://github.com/jundot/omlx/pull/1136#issuecomment-4429008784):

>A question: is 4-bit affine "fallback" enough for the sensitivity model role? In my understanding it could potentially lead to a different dynamic bits allocation since we are going to apply a quantization on top of another quantization, and only then calculate how much it hurts? I've always quantized oQ using affine Q8 sensitivity model with such assumption in mind.

Come true. I have different .safetensors with using affine 4-bit and 8-bit sensitivity models, meaning some bits allocation made different choices.

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -3063,23 +3063,43 @@ def _build_proxy_for_sensitivity(
 
     The caller is responsible for deleting the returned directory.
     """
-    from mlx_lm import convert
+    try:
+        from omlx.patches.mlx_lm_mtp import (
+            apply_mlx_lm_mtp_patch,
+            is_mtp_active,
+            set_mtp_active,
+        )
+        _have_lm_patch = apply_mlx_lm_mtp_patch()
+    except Exception:
+        _have_lm_patch = False
+        is_mtp_active = None
+        set_mtp_active = None
 
-    # mlx-lm's convert() refuses to write into a pre-existing directory,
-    # so reserve a unique temp name and let convert() create it.
-    proxy_dir = Path(tempfile.mkdtemp(prefix="omlx_oq_proxy_", dir=working_dir))
-    shutil.rmtree(proxy_dir)
-    convert(
-        hf_path=model_path,
-        mlx_path=str(proxy_dir),
-        quantize=True,
-        q_bits=_PROXY_QUANT_BITS,
-        q_group_size=_PROXY_QUANT_GROUP_SIZE,
-        q_mode="affine",
-        dtype=dtype,
-        trust_remote_code=trust_remote_code,
-    )
-    return proxy_dir
+    prev_active = is_mtp_active() if _have_lm_patch else False
+    try:
+        if _have_lm_patch:
+            set_mtp_active(True)
+
+        from mlx_lm import convert
+
+        # mlx-lm's convert() refuses to write into a pre-existing directory,
+        # so reserve a unique temp name and let convert() create it.
+        proxy_dir = Path(tempfile.mkdtemp(prefix="omlx_oq_proxy_", dir=working_dir))
+        shutil.rmtree(proxy_dir)
+        convert(
+            hf_path=model_path,
+            mlx_path=str(proxy_dir),
+            quantize=True,
+            q_bits=_PROXY_QUANT_BITS,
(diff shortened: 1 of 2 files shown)


jundot:
You're right that the sensitivity proxy build wasn't taking the MTP path into account. Thanks for the proper patch.

When I first introduced the sensitivity-model feature, I looked at how layer-distribution behaves between a 4-bit proxy and the full-precision model, and the two were quite similar, so I figured 4-bit was good enough, and being able to run oQ quant on small-memory machines at all was the bigger win for me. Your concern is a fair one though, and I'll run a proper sensitivity comparison after this patch lands.

Really appreciate the suggestion and the quick fix for the part I missed.

PR #1206, opened 2026-05-12
fix(ui): chat message render line break incorrectly, textarea inline edit

beamivalice:
1. Chat message rendered line break incorrectly.

Screenshot 2569-05-12 at 22 04 46

Screenshot 2569-05-12 at 22 04 56

2. Textarea inline edit was contained to just min size (280px) not the real message bubble's width

Screenshot 2569-05-12 at 21 34 55

Screenshot 2569-05-12 at 22 05 49

### DIFF
--- a/omlx/admin/templates/chat.html
+++ b/omlx/admin/templates/chat.html
@@ -157,11 +157,16 @@
         align-self: flex-end;
     }
 
+    .user-message.is-editing {
+        width: min(768px, 100%);
+    }
+
     .user-message-bubble {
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for the clean fix and the before/after shots. The paragraph-margin pattern (`margin: 0 0 1em` + `p:last-child { margin-bottom: 0 }`) and toggling `is-editing` for the edit-time width both look right, and `box-sizing: border-box` on the textarea is the correct follow-up to dropping the 600px cap. Merging.

PR #1209, opened 2026-05-12
chore: git ignore uv.lock

deepsweet:
I, likely among other contributors, use [UV](https://docs.astral.sh/uv/) Python package manager to handle venv and dependencies with `uv sync`.

I propose that we just git-ignore the lockfile so that it doesn't get commited accidentally one day.

It doesn't oblige you to change your way of dealing with Python, just for overall convenience.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,3 +113,6 @@ omlx/admin/tailwindcss-*
 
 # Git worktrees
 .worktrees/
+
+# UV lockfile
+uv.lock


jundot:
Thanks! Makes sense. We already support `uv sync --dev` and there's no lockfile committed today, so ignoring it keeps the door open for uv users without changing the pip workflow. Merging.

PR #1217, opened 2026-05-13
fix(mtp): prevent process-wide _MTP_ACTIVE flag leaking across model loads

fish0710:
## Summary

Fixes the process-wide `_MTP_ACTIVE` flag leaking from one model load to another, which caused SpecPrefill draft model loading to crash with "Native MTP is enabled but converted weights are missing mtp.* tensors".

## Root cause

`set_mtp_active()` is only called inside `_is_mtp_compatible()` branch in `maybe_apply_pre_load_patches()`. If the current model is **not** MTP-compatible, the flag is never reset. This means:

1. Load `Qwen3.6-27B-oQ4-mtp` with `mtp_enabled=True` → `_MTP_ACTIVE=True`
2. Switch to `Qwen3.6-35B-A3B-UD-MLX-4bit` (non-MTP) → flag stays `True`
3. SpecPrefill loads `Qwen3.5-0.8B-MLX-4bit` as draft → patched `__init__` sees `is_mtp_active()=True` → attaches MTP → crash

## Fix

- `maybe_apply_pre_load_patches()`: reset `_MTP_ACTIVE=False` at function entry so every model load starts clean
- `BatchedEngine._load_draft()`: snapshot/restore flag around draft model load
- `VLMBatchedEngine._load_draft()`: same snapshot/restore guard

## Test plan

- [x] Load MTP-enabled model, then switch to non-MTP model with SpecPrefill draft → no crash
- [x] Load non-MTP model directly with SpecPrefill draft → works normally

### DIFF
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -310,8 +310,21 @@ class BatchedEngine(BaseEngine):
                 try:
 
                     def _load_draft():
-                        draft_model, _ = load(specprefill_draft)
-                        return draft_model
+                        from ..patches.mlx_lm_mtp import set_mtp_active
+
+                        was_mtp = False
+                        try:
+                            from ..patches.mlx_lm_mtp import is_mtp_active
+
+                            was_mtp = is_mtp_active()
+                        except Exception:
+                            pass
+                        set_mtp_active(False)
+                        try:
+                            draft_model, _ = load(specprefill_draft)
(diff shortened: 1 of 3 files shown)


jundot:
Thanks — same MTP flag fix already shipped in #1219 (merged 04a0ce6) bundled with the launch host fix. Closing this as a duplicate.

PR #1219, opened 2026-05-13
fix(cli): launch connect via localhost when server binds 0.0.0.0

fish0710:
## Summary

Fixes `omlx launch` failing when the server is configured to bind to `0.0.0.0`.

## Problem

When `settings.json` contains `host: "0.0.0.0"` (e.g. after `omlx serve --host 0.0.0.0`), `omlx launch claude` fails with:

```
oMLX server is not running at http://0.0.0.0:12345
Start the server first: omlx serve
```

`0.0.0.0` is a valid bind-all address but not a valid client-side connect address.

## Fix

Introduce `connect_host` in `launch_command()` which falls back to `127.0.0.1` whenever the resolved host is `0.0.0.0` (or empty). The original `host` value is preserved for display; only the actual HTTP connection URL and the downstream integration `launch()` call use the corrected address.

## Test plan

- [x] `omlx serve --host 0.0.0.0 --port 12345` then `omlx launch claude` → connects successfully
- [x] `omlx serve --host 127.0.0.1` then `omlx launch claude` → still works (no regression)

### DIFF
--- a/omlx/cli.py
+++ b/omlx/cli.py
@@ -301,8 +301,13 @@ def launch_command(args):
     host = args.host or settings.server.host
     port = args.port or settings.server.port
 
+    # 0.0.0.0 is a valid bind address but not a valid connect address.
+    # Fall back to localhost so launch can reach the server regardless
+    # of which interface it was bound to.
+    connect_host = host if host and host != "0.0.0.0" else "127.0.0.1"
+
     # Check if oMLX server is running
-    base_url = f"http://{host}:{port}"
+    base_url = f"http://{connect_host}:{port}"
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for catching this. The 0.0.0.0 to 127.0.0.1 fallback in launch_command is the right shape; the bind addr vs connect addr distinction matches what utils/network.py already does. Noticed the branch also carries the _MTP_ACTIVE leak fix (afe0bb2); the root cause analysis there is spot on, especially the draft-model path that bypasses maybe_apply_pre_load_patches.

Merging.

PR #1224, opened 2026-05-13
feat(prefill): interleaved chunked prefill + decode

drumtorben:
## Problem

On oMLX, a long-context request blocks decode for all concurrent requests until its entire prefill completes. For a 100K-token prompt at `prefill_step_size=2048`, that's ~50 sequential model calls with no decode output in between — causing TTFT spikes and throughput loss for concurrent users. The gap was already acknowledged in a comment at `scheduler.py`.

## Solution

Process **one prefill chunk per `step()`** for in-progress requests, then run the normal decode step. Already-running requests keep streaming tokens; new long requests see their first token sooner on average.

The feature is opt-in (`chunked_prefill: false` by default) and can be toggled live from the admin UI without restarting.

## Implementation

**Scheduler (`omlx/scheduler.py`)**
- `_PrefillState` dataclass — carries mid-prefill state (cache, remaining tokens, boundary tracking, sampler params) across step boundaries
- `self.prefilling` / `self._prefill_states` — new queue + state dict for in-flight chunked prefills
- `_begin_prefill()` — initialises `_PrefillState` from a request; no model calls
- `_step_prefill_chunk()` — processes one `prefill_step_size` chunk: model forward pass, boundary snapshot check, abort check, cache clear
- `_advance_chunked_prefills()` — called at the top of `step()`; advances each in-flight prefill by one chunk, inserts completed requests into `BatchGenerator`
- `_schedule_waiting()` — forks to chunked path when `chunked_prefill=True`, prompt exceeds `prefill_step_size + 1` tokens, and request is not a VLM (VLM deferred to follow-up)
- Lifecycle methods updated: `has_requests()`, `fail_all_requests()`, `reset()`, `get_stats()`, `_do_abort_request()`

**Settings**
- `SchedulerConfig.chunked_prefill: bool = False`
- `SchedulerSettings.chunked_prefill: bool = False` with `from_dict` / `to_scheduler_config` wiring
- Admin API: `GET /settings` returns `scheduler.chunked_prefill`; `POST /settings` applies it live to all loaded schedulers
- Admin UI: toggle switch added to the Resources section

**Tests (`tests/test_scheduler_chunked_prefill.py`)**

24 unit tests covering: config defaults, `_PrefillState` fields, queue initialisation, `has_requests`, `get_stats`, `reset`, `fail_all_requests`, abort cleanup, `_advance_chunked_prefills` (no-op, in-progress, completion, skip-aborted, `_PrefillAbortedError`, multi-request), and the `_schedule_waiting` fork (short/long prompt, first-chunk completion, disabled mode).

## Scope exclusions (deferred)

- **VLM requests**: embedding slices need per-chunk offset tracking into `embeds_array` — fall back to full prefill via `vlm_embeds is None` guard
- **SpecPrefill requests**: already bypass `_do_external_prefill` entirely; no changes needed

## Test plan

- [x] `uv run pytest tests/test_scheduler.py tests/test_scheduler_chunked_prefill.py` — 115 tests, all passing
- [x] Enable `chunked_prefill: true` in settings, send a 50K-token prompt alongside a short concurrent request — short request should start generating before the long prefill finishes
- [x] Verify `GET /v1/admin/settings` returns `scheduler.chunked_prefill`
- [x] Toggle in admin UI saves and reflects immediately without restart

### DIFF
--- a/omlx/admin/i18n/en.json
+++ b/omlx/admin/i18n/en.json
@@ -322,6 +322,8 @@
   "settings.resource.prefill_memory_guard_description": "Estimate memory before prefill and defer scheduling under memory pressure to prevent Metal allocation failures.",
   "settings.resource.max_concurrent_requests": "Max Concurrent Requests",
   "settings.resource.max_concurrent_requests_hint": "Maximum number of requests processed simultaneously. Higher values increase throughput but use more memory.",
+  "settings.resource.chunked_prefill": "Chunked Prefill",
+  "settings.resource.chunked_prefill_description": "Interleave prefill chunks with decode steps. Reduces time-to-first-token for concurrent requests by processing one prefill chunk per step instead of blocking decode.",
   "settings.resource.restart_badge": "Restart needed",
   "settings.cache.section_label": "Cache",
   "settings.cache.enabled": "Cache Enabled",
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -21,9 +21,9 @@ import sys
 import time
 from collections import deque
 from dataclasses import asdict, is_dataclass
-from datetime import datetime, timezone
+from datetime import UTC, datetime
 from pathlib import Path
-from typing import Any, Dict, Optional, List, Literal
+from typing import Any, Literal
 
 import requests
 from fastapi import APIRouter, Depends, HTTPException, Request, Response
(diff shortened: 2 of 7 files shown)


jundot:
Thanks for the careful work — the implementation mirrors _do_external_prefill cleanly and the queue management tests are solid. Verified the TTFT win on a real-server smoke test (52K-token prompt + concurrent short request: short TTFT dropped from 58s to 1.7s with chunked_prefill on). Merging now. I'll land a few small polish commits on main right after: phys_footprint in the per-chunk memory check (jetsam watches that on macOS), a progress callback so the admin UI prefilling list moves during chunked prefill, and a finish_reason="error" path when a chunked prefill hits the memory limit so the client doesn't hang. Ships in the next release.

PR #1227, opened 2026-05-13
fix(admin): make speculative toggle mutual exclusion symmetric

fish0710:
## Summary

Fixes asymmetric mutual-exclusion guards in the model settings modal that caused toggles to become stuck.

## Problem

The TurboQuant, DFlash, and SpecPrefill toggles did not check whether mtp_enabled or vlm_mtp_enabled was active. But the MTP toggle did check dflash_enabled and turboquant_kv_enabled.

This created a trap:
1. Open MTP - toggle is on
2. Open TurboQuant - TurboQuant toggle does not check MTP, so it opens
3. Now try to close MTP - MTP toggle is disabled because turboquant_kv_enabled is true
4. User is stuck: must close TurboQuant first, then close MTP

## Fix

Add missing guards so every speculative-decoding toggle checks every option it is mutually exclusive with, matching the backend validation in model_settings.py::__post_init__.

| Toggle | Now checks |
|--------|-----------|
| TurboQuant | mtp_enabled || vlm_mtp_enabled |
| SpecPrefill | vlm_mtp_enabled |
| DFlash | mtp_enabled || vlm_mtp_enabled |
| MTP | dflash_enabled || turboquant_kv_enabled || vlm_mtp_enabled |

Generated with Claude Code

### DIFF
--- a/omlx/admin/templates/dashboard/_modal_model_settings.html
+++ b/omlx/admin/templates/dashboard/_modal_model_settings.html
@@ -583,12 +583,12 @@
                                             <p class="text-xs text-neutral-500 mt-0.5">{{ t('modal.model_settings.turboquant_kv_hint') }}</p>
                                         </div>
                                         <button type="button"
-                                                @click="if (!modelSettings.is_paroquant) modelSettings.turboquant_kv_enabled = !modelSettings.turboquant_kv_enabled"
-                                                :disabled="modelSettings.is_paroquant"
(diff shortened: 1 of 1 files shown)


jundot:
Thanks. Verified the guards match the backend mutex rules in model_settings.py::__post_init__ exactly, including the asymmetric set (specprefill only conflicts with vlm_mtp, not with mtp/dflash/turboquant). I18n keys all exist. Merging.

PR #1231, opened 2026-05-13
feat(chat): enable chat multi-tasking and fix chat response appear to follow bug

beamivalice:
**PROBLEM**
- Currently you cannot do chat multitask. The chat in-progess response appear to follow you everywhere when you click on other chat tab. And even do a persist a save on whatever tab you currently navigated to, not the one suppose to generate response.

**SOLUTION**
- To fully enable chat multitask, this require quite a lot of fixes. The root problem was that streaming state was stored at the page level instead of the chat level. Things like the live assistant bubble, thinking panel, status timeline, abort controller, and input lock were all shared across the whole page. When you switched chats, the UI changed which message list was visible, but the active stream state stayed global, so the newly selected chat could start showing the other chat’s in-flight generation.
- This change moves streaming state to per-chat sessions. Each chat now owns its own stream lifecycle, including partial content, reasoning text, timeline visibility, engine status, abort controller, render state, and progress polling. The live streaming bubble only renders for the currently selected chat if that specific chat is actually streaming.
- It also fixes a request-context bug where parts of the streaming flow could still read the currently selected model after a tab switch. The full request lifecycle now stays bound to the originating chat, including the source chat ID, model, and system prompt. That context is preserved through follow-up tool-call recursion as well, so background generations continue using the correct chat configuration even if the user navigates elsewhere.
- Before this change, the UI behaved as if the whole page was busy whenever any stream was active. Now sending, editing, regenerating, and input disabling are all checked against the currently selected chat only, so one busy chat no longer blocks work in another.
- Chat saves are written back to the chat that actually owns the updated message list. saveCurrentChat() now derives the title from the messages being saved rather than implicitly from whatever chat is currently visible, which prevents sidebar metadata drift during background updates.

---

After fix, chat can now multitask, as long as chat page remains open and HTTP stay connected.

Screenshot 2569-05-13 at 21 04 51

### DIFF
--- a/omlx/admin/templates/chat.html
+++ b/omlx/admin/templates/chat.html
@@ -1030,17 +1030,18 @@
                 </template>
 
                 <!-- Streaming Message -->
-                <div x-show="isStreaming" class="assistant-message message-fade-in">
+                <div x-show="isCurrentChatStreaming()" class="assistant-message message-fade-in">
                     <div class="message-header">
                         <div class="flex items-center gap-2 text-sm font-medium" style="color: var(--text-primary);">
                             <i data-lucide="cpu" class="w-4 h-4" style="color: var(--text-tertiary);"></i>
-                            <span x-text="currentModel || window.t('chat.assistant_label')"></span>
+                            <span
+                                x-text="currentStream()?.sourceModel || currentModel || window.t('chat.assistant_label')"></span>
                         </div>
                         <div class="flex gap-1">
                             <!-- Timeline toggle for streaming -->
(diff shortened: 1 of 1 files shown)


jundot:
Thanks, moving streaming state to per-chat sessions is the right call, and capturing chatId + model + systemPrompt up front keeps the originating chat bound through tool-call recursion. Cleanup paths look complete too.

Merging in the next release. Quick heads-up: the live bubble's stable/tail roots are still a single shared DOM node, so switching between two actively-streaming chats can briefly show the wrong stable portion (data is fine, self-heals once the stream completes). I'll push a follow-up on main right after.

PR #1244, opened 2026-05-13
fix(tests): repair pre-existing upstream test failures and import guards

a4501150:
## Summary
- Fix GitHub releases API response shape in update-check tests (`{...}` → `[{...}]`)
- Add xgrammar import guards to grammar tests (`@requires_xgrammar`)
- Update `MockBaseEngine` to satisfy `BaseEngine` ABC (`stop()`, `get_stats()`, `get_cache_stats()`)
- Use `len(VALID_BENCHMARKS)` instead of hardcoded count in accuracy benchmark test

## Test plan
- [x] `pytest tests/test_admin_auth.py tests/test_admin_update_check.py tests/test_grammar.py tests/test_accuracy_benchmark.py tests/integration/test_server_endpoints.py -v`

### DIFF
--- a/tests/integration/test_server_endpoints.py
+++ b/tests/integration/test_server_endpoints.py
@@ -15,6 +15,7 @@ import pytest
 from fastapi.testclient import TestClient
 
 from omlx.api.responses_utils import ResponseStore
+from omlx.engine.base import BaseEngine
 from omlx.engine.embedding import EmbeddingEngine
 from omlx.engine.reranker import RerankerEngine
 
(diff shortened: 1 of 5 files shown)


jundot:
Thanks for cleaning these up. I checked each fix against the source: `/releases` does return a list and `release_check.py` iterates over it, `VALID_BENCHMARKS` is now 16 entries, and the `@requires_xgrammar` marker is needed because `@patch` resolves its target before the in-body `importorskip` runs. All correct, no source changes, zero regression risk. Merging now.

PR #1245, opened 2026-05-13
feat(responses): native reasoning support for Responses API

a4501150:
## Summary
- Add native reasoning item support to the `/v1/responses` endpoint
- Extract thinking/reasoning tokens and emit them as `reasoning` content items in the response
- Wire reasoning through the Responses API models, utils, and server route
- Includes prerequisite test fixes from #1244

## Test plan
- [x] `pytest tests/integration/test_server_endpoints.py -v`
- [ ] Manual test with a reasoning-capable model via `/v1/responses` with `reasoning.effort` set


jundot:
Thanks for this, the streaming SSE design looks solid.

Two things I'd like to sort out before merging, both in the non-reasoning parts of the diff:

1. `build_response_usage` — the new `reasoning_tokens` param landed as the 3rd positional arg, but `server.py:4113` still calls it positionally with `output.cached_tokens`. So `cached_tokens` now reports 0 and `reasoning_tokens` gets the cached value instead. Passing `cached_tokens=` as a keyword fixes it.

2. `model_profiles.py` — `dflash_draft_quant_bits` isn't a field on `ModelSettings`, which still has the four `dflash_draft_quant_enabled/_weight_bits/_activation_bits/_group_size` fields. Since that list is a profile allowlist, those four would silently drop out of per-model profiles. The `dflash_max_ctx` / `dflash_in_memory_cache*` additions look fine.

One smaller note: `reasoning_tokens` never gets a real value (streaming hardcodes 0), so it might be worth wiring it up or leaving the param out for now.

a4501150:
Addressed all review comments:

1. **`build_response_usage` positional arg bug** — switched to all keyword arguments (named parameter). Also wired up `reasoning_tokens` by tokenizing the extracted thinking content, so it now reports actual counts instead of hardcoded 0 (both streaming and non-streaming paths).

2. **`model_profiles.py` wrong field** — replaced `dflash_draft_quant_bits` with the four real `ModelSettings` fields: `dflash_draft_quant_enabled`, `dflash_draft_quant_weight_bits`, `dflash_draft_quant_activation_bits`, `dflash_draft_quant_group_size`.

cc @jundot for re-review

jundot:
Thanks for the PR — direction is right and this fixes a real gap in /v1/responses today. Going to merge and clean up in a follow-up commit:

- reasoning preservation through `_flush_pending_tool_calls` so the reasoning → function_call → output sequence round-trips on multi-turn tool use
- unused `dataclass_fields` import
- capture `reasoning_output_index` at open time instead of hardcoding 0
- unit tests for the new branches

PR #1247, opened 2026-05-13
fix(load): VLM model loading fixes for oQ-quantized checkpoints

a4501150:
## Summary
- Expand per-layer quantization config keys for VLM model-tree paths so `quantization` config matches the MLX model parameter hierarchy (e.g. `language_model.model.layers.N` vs `model.layers.N`)
- Centralise pre-load patches in oQ `_measure_sensitivity` so MTP/nested-visual patches are active during sensitivity measurement
- Remap nested visual keys (`language_model.model.visual.*` → `vision_tower.*`) for MLX-format VLM models where mlx-vlm skips `Model.sanitize`
- Fix nested-visual patch idempotency: use function-attribute marker instead of module-level flag so the wrap can re-apply if another patch (e.g. MTP runtime) overwrites `Model.sanitize`
- Add inline nested-visual post-fixup in all three MTP sanitize functions

### Background on nested visual bug
Qwen3.6-35B-A3B nests ViT weights at `model.language_model.visual.*`. mlx-vlm's sanitize uses if/elif that matches `model.language_model` first, rewriting to `language_model.model.visual.*` — the `model.visual → vision_tower` branch never fires. For non-MLX-format models, the existing `qwen3_6_nested_visual` sanitize wrap catches this. But mlx-vlm skips sanitize entirely for MLX-format checkpoints (`format=mlx` in safetensors metadata), so oQ output models fail with "333 parameters not in model". The new `_remap_nested_visual_on_load` context manager intercepts `nn.Module.load_weights` during the scoped `vlm_load()` call to remap keys before they reach the model.

## Test plan
- [x] `pytest tests/test_oq.py -v`
- [x] Server loads `Qwen3.6-35B-A3B-uncensored-heretic` (non-MLX format) — nested-visual sanitize wrap fires
- [x] Server loads `Qwen3.6-35B-A3B-uncensored-heretic-oQ6-mtp` (MLX format) — `_remap_nested_visual_on_load` remaps 333 keys

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -365,6 +365,11 @@ def _strip_audio_config_if_orphaned(model_dir: Path):
 
     def _patched(path, **kwargs):
         cfg = original(path, **kwargs)
+
+        from ..utils.model_loading import expand_per_layer_quant_keys
+
+        expand_per_layer_quant_keys(cfg)
+
         if cfg.get("audio_config") is None:
             return cfg
         try:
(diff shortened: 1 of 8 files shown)


jundot:
Thanks for this, the load-path fixes look solid and the nested-visual / quant-key remapping all check out.

Merging this with two follow-ups from me, no action needed on your side:

1. Dropping the new `TestBuiltinCalibration` class. It expects `mixed`/`chat`/`bartowski` categories and 3000+ samples, but the `oq_calibration_data.json` in the repo has 7 categories / ~600 samples — looks like the updated JSON wasn't part of the diff. The `reasoning` addition in `_load_builtin_calibration` is fine since that key already exists in the shipped JSON.
2. `_measure_sensitivity` now goes through `maybe_apply_pre_load_patches`, which leaves `mtp_active=False` and doesn't apply the mlx-vlm runtime patch. For an MLX-format VLM checkpoint with MTP heads, mlx-vlm skips sanitize entirely, so the `language_model.mtp.*` weights stay in the dict but no MTP head is attached — `load_weights` then rejects them and sensitivity silently returns empty. The text path is fine (the patched `qwen35_model.sanitize` self-consistently strips `mtp.*` when no head). I'll restore the head attachment for the VLM path in a follow-up.

a4501150:
thanks @jundot

For the first one - that should belong to this pr https://github.com/jundot/omlx/pull/1246 - cherry picked the wrong files. If we decide to merge that maybe a follow up will be moving TestBuiltinCalibration in PR-1246

PR #1255, opened 2026-05-14
Pass extra cli args to codex with 'omlx launch'

EdenGottlieb:
Code written by Qwen3.6-35B-A3B-UD-MLX-4bit running on oMLX with my guidance
Tested, committed, and PR opened by a human


jundot:
Thanks for this. The launch CLI already collects unknown tokens into `extra_args` and forwards them, but codex was the one integration not consuming them, so the flags got silently dropped. The change matches how `claude.py` already handles `extra_args`, and the test covers the argv forwarding. Merging now, it'll be in the next release.

PR #1257, opened 2026-05-14
fix(deps): bump dflash-mlx to fix act_bits=32 crash (DFlash w8a32)

fparrav:
## Summary

- **Root cause**: `_cast_to_f32` in `dflash_mlx.runtime.loading.load_draft_bundle` was defined as `(_, x: mx.array)` but `model.apply()` passes a single argument — calling `_cast_to_f32(array)` fails with `TypeError: missing 1 required positional argument: 'x'`
- **Trigger**: Only activates when DFlash quantization uses `act_bits=32` (i.e. `w8a32` config). With `act_bits=16` the code path is never reached, so the bug was invisible until users enabled 32-bit activations for better coding accuracy
- **Effect**: `DFlashEngine.start()` raises, caught by `engine_pool` which silently falls back to `BatchedEngine` — no visible error, just unexpectedly low throughput (~96 tok/s instead of ~250+ tok/s)
- **Fix**: Already merged upstream in `bstnxbt/dflash-mlx@1dac0a8` ("Fix MLX apply draft cast callbacks"). This PR bumps the pin to `6761883` (current HEAD)

## Test plan

- [ ] Start oMLX with a model that has DFlash enabled and `act_bits=32` (w8a32 config)
- [ ] Verify server log shows `DFlashEngine loaded` instead of `DFlash start failed`
- [ ] Verify throughput is ~250+ tok/s at pp1024 (not ~96 tok/s fallback)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,8 +75,8 @@ dependencies = [
     # mlx-vlm from commit (f96138e) - Gemma4 MTP server batching (PR #1166), speculative utils refactor (PR #1169), Qwen native MTP drafter, MiniCPM-V 4.6
     "mlx-vlm @ git+https://github.com/Blaizzy/mlx-vlm@f96138eef1f5ce7fb5d97f8dd41a664a195b5659",
     "Pillow>=9.0.0",
-    # dflash-mlx v0.1.5.1 (20d68db) — original bstnxbt repo. Gemma4 support, runtime package refactor, stable prefix snapshot reuse
-    "dflash-mlx @ git+https://github.com/bstnxbt/dflash-mlx@20d68db3b3c0ae3dd6d3a2f0d3c10b2344ee514e",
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for tracking down the act_bits=32 root cause, the analysis is accurate.

The issue is that bumping the pin from `20d68db` to `6761883` isn't a one-line change in practice. It pulls in 10 upstream commits with a large code change, and two of them removed symbols that oMLX's `DFlashEngine.start()` imports: `make_draft_backend` (now `EagerDraftBackend()`) and `runtime_config_from_profile` (now `runtime_config_from_defaults`, different signature). So merging just the pin would break the DFlash engine entirely with an ImportError, not only the w8a32 path.

I'm going to handle the dflash-mlx upgrade separately, together with the matching `omlx/engine/dflash.py` changes, so I'll close this PR for now. Thanks again for the catch.

PR #1274, opened 2026-05-16
feat(cache): asymmetric KV quantization (K=INT4, V=INT2)

austindixson:
## Summary
Assigns K=INT4 and V=INT2 for cold-tier KV cache blocks, exploiting the fact that keys (dot-product similarity) are precision-sensitive while values (weighted averaging) are noise-tolerant.

## Benchmark
Real PagedSSDCacheManager, 8 blocks of 16L×8H×128D×64T:

| Format | SSD Footprint | Reduction |
|---|---|---|
| fp16 | 32.0 MB | 1.0× |
| INT4 (symmetric) | 12.1 MB | 2.6× |
| **INT4/2 (asymmetric)** | **6.0 MB** | **5.3×** |

50% additional space savings vs uniform INT4 with round-trip verified through oMLX's safetensors pipeline.

## Implementation
200-line drop-in patch. `install()` / `remove()` toggle. Compatible with attention_cache and wiring compression modules.

### DIFF
--- a/omlx/ablations/__init__.py
+++ b/omlx/ablations/__init__.py
@@ -1,0 +1,6 @@
+# SPDX-License-Identifier: Apache-2.0
+"""OMLX+ Ablation Suite — Hardware-targeted optimizations for Apple Silicon LLM inference."""
+
+from .asymmetric_kv import install, remove, get_stats
+
+__all__ = ["install", "remove", "get_stats"]
--- a/omlx/ablations/asymmetric_kv.py
+++ b/omlx/ablations/asymmetric_kv.py
@@ -1,0 +1,268 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Asymmetric KV Quantization.
+
+Keys and values serve fundamentally different roles in attention:
+  - Keys (K): used for dot-product similarity → precision-sensitive
+  - Values (V): used for weighted averaging       → noise-tolerant
+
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for putting this together. The K/V asymmetry intuition is solid, and the codec itself is cleanly written.

Here's why I don't think the design fits oMLX though.

Compression only applies to **SSD cold-tier blocks**. On restore, blocks are decompressed back to fp16 before they reach the GPU, so the active inference footprint (the part that competes with model weights for unified memory) is unchanged. The only thing that shrinks is the on-disk size of cold blocks, and disk space isn't typically the bottleneck on Apple Silicon. On the other side of the trade, V is being squeezed to ternary (`clip(round(x/scale), -1, 1)` only uses 3 of the 4 INT2 levels, so ~1.58 bit effective) and reinjected into attention every time a cold block is rehydrated. Any accuracy loss shows up directly in user-visible output, so the practical trade-off is quality loss for disk space, while the unified-memory pressure that users typically hit stays exactly the same.

Hot KV cache quantization (the in-memory paged blocks) is the direction where the math actually pays off on Apple Silicon, since that shrinks unified-memory pressure during live inference. That's already on my roadmap as a later milestone, so I'd rather keep the design space open until I get there.

Going to close this one out. Please don't read it as a no on future contributions. Thanks again for the effort.

austindixson:
Thank you for the thoughtful review!

PR #1275, opened 2026-05-16
feat(cache): semantic prefix matching via layer-1 hidden state hashing

austindixson:
## Summary
Replaces exact token-ID hashing in oMLX's prefix cache with locality-sensitive hashing on layer-1 hidden states. "Explain gravity" and "What is gravity?" can now share KV cache blocks even though their token sequences differ.

## Method
Johnson-Lindenstrauss random projection maps layer-1 hidden states to 128-bit binary hashes. Similar prompts produce hashes with low Hamming distance. The semantic index stores block hashes keyed by semantic hash and performs fuzzy lookup within a configurable Hamming radius.

## Benchmark
| Query Type | Cache Hits | False Positives |
|---|---|---|
| Semantically similar prompts | 5/5 (100%) | 0 |
| Random/different prompts | 0/5 (0%) | 0 |

Perfect selectivity — semantically similar prompts share cache, unrelated prompts never match.

## Implementation
`install_semantic_prefix()` / `remove_semantic_prefix()`. 200-line module. Compatible with attention-guided eviction and asymmetric KV compression.

### DIFF
--- a/omlx/ablations/__init__.py
+++ b/omlx/ablations/__init__.py
@@ -1,0 +1,20 @@
+# SPDX-License-Identifier: Apache-2.0
+"""OMLX+ Ablation Suite — Hardware-targeted optimizations for Apple Silicon LLM inference."""
+
+from .semantic_prefix import (
+    SemanticPrefixIndex,
+    install_semantic_prefix,
+    remove_semantic_prefix,
+    compute_semantic_hash,
+    hamming_distance,
+    are_semantically_similar,
+)
+
+__all__ = [
+    "SemanticPrefixIndex",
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for putting this together, really appreciate the effort.

I think there's a fundamental issue though: prefix cache lookup runs at request enqueue, before any forward pass — so layer-1 hidden states aren't available yet to compute the hash. And even if they were, sharing KV blocks across prompts with different tokens would corrupt attention output, since K/V values depend on the actual token sequence.

Closing for now, but if you want to explore a different angle on prompt-level cache sharing I'd be happy to chat.

PR #1278, opened 2026-05-16
feat:(ui): Add observable memory bar to the dashboard.

beamivalice:
* Pull number from the memory enforcer
* Update at 500ms
* Display at the dashboard
* Per active model use delta between not load and loaded state to determine real memory usage, some model use a lot less than estimate.

Screenshot 2569-05-16 at 17 16 26
Screenshot 2569-05-16 at 17 16 36
Screenshot 2569-05-16 at 17 17 19

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -1613,6 +1613,12 @@ async def list_models(is_admin: bool = Depends(require_admin)):
             "is_loading": model_info.get("is_loading", False),
             "estimated_size": model_info.get("estimated_size", 0),
             "estimated_size_formatted": format_size(model_info.get("estimated_size", 0)),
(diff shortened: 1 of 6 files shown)


jundot:
thanks — merging, will fold the small cleanup (dead `model_memory_used_actual` / `current_model_memory_actual` fields, `activeModelSizeLabel` alias, i18n for the disabled label, tailwind rebuild, inline style) into a follow-up commit on main for the next release.

one heads-up — `phys_footprint` is process-wide and inference on other models keeps running while load holds the pool lock, so `actual_size` is a rough observation that can include neighboring KV growth. I'll label it as such in the follow-up so it doesn't read as exact.

PR #1284, opened 2026-05-16
fix: clean up partial HF download shards on cancel

### LINKED ISSUE #1260: Cancelled HF downloads don't clean up `._____temp/` partial shards
**Describe the bug**
Cancelling a download of `mlx-community/Qwen3.6-35B-A3B-4bit` from the admin dashboard left the model directory with `._____temp/model-0000{1,2,3,4}-of-00004.safetensors` partial shards. Nothing in the UI offered to clean them up. With repeat cancellations this adds up and can fill the disk.

**To Reproduce**
1. Open the admin dashboard model picker.
2. Start downloading any multi-shard HF model (e.g. mlx-community/Qwen3.6-35B-A3B-4bit).
3. Cancel the download mid-progress.
4. Look at `<model_dir>/<owner>/<repo>/`. A `._____temp/` directory with partial model-*.safetensors shards remains. No UI action removes it. Workaround: `rm -rf ~/.omlx/models/mlx-community/Qwen3.6-35B-A3B-4bit`.

**Expected behavior**
Either (a) cancel automatically discards partial shards, or (b) the UI offers an explicit "Cancel and delete" option. Right now neither exists and partials accumulate until manually removed.

**Screenshots**
N/A

 - macOS Version: 15.4.1
 - oMLX Version v0.3.9.dev2

**Additional context**
`omlx/admin/hf_downloader.py` defines `_cleanup_partial`  which appears to do exactly this, but it is not called from `cancel_download` or the `CancelledError` handler in `_run_download`.
https://github.com/jundot/omlx/blob/51907f08074742defec4375fb629e289801a8a9f/omlx/admin/hf_downloader.py#L841

Is preserve-for-resume the intentional design, with a separate discard path just never built? Happy to PR a discard_files flag on cancel, or a "Cancel and delete" UI option, if welcome.

mvanhorn:
## Summary

When an HF model download is interrupted partway, the partial shard file is left on disk. The next download attempt sees the file with the right name and treats it as complete, leading to silent corruption on load. Catch the interrupt / exception path, delete the in-progress shard, and re-raise.

## Why this matters

This is the exact reproducer in #1260: Ctrl-C a download, retry, get a corrupt model. Users only find out at inference time when MLX errors on weight shapes. Cleanup-on-cancel is the standard fix and matches how `huggingface_hub` itself handles partial files.

## Changes

- `omlx/admin/hf_downloader.py` - wrap the shard download in a try/except, delete the partial file on `BaseException` (covers KeyboardInterrupt), then re-raise.
- `tests/test_hf_downloader.py` - add test cases that simulate interrupt during download and confirm partials are cleaned up; verify resumed download produces a valid file.

## Testing

New unit tests pass locally. Existing downloader tests still pass.

Fixes #1260

### DIFF
--- a/omlx/admin/hf_downloader.py
+++ b/omlx/admin/hf_downloader.py
@@ -527,10 +527,10 @@ class HFDownloader:
     async def retry_download(
         self, task_id: str, hf_token: str = ""
     ) -> DownloadTask:
-        """Retry a failed or cancelled download, resuming from existing files.
+        """Retry a failed or cancelled download.
 
-        Since partial files are preserved on disk, snapshot_download will
-        automatically skip already-completed files.
+        If partial files remain on disk after a failed download,
+        snapshot_download will automatically skip already-completed files.
 
         Args:
             task_id: The task ID of the failed/cancelled download.
@@ -558,7 +558,7 @@ class HFDownloader:
         del self._tasks[task_id]
         self._cancelled.discard(task_id)
 
-        # Start fresh download (snapshot_download resumes from existing files)
+        # Start fresh download (snapshot_download resumes any existing files)
         new_task = await self.start_download(repo_id, hf_token)
         new_task.retry_count = old_retry_count + 1
         return new_task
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the patch — root cause is right (the missing `_cleanup_partial` call) and the test path correction is a real catch. Merging this.

One small follow-up I'll land in the next release: scope the cleanup to `<target_dir>/._____temp/` instead of rmtree on the whole target_dir. Hub keeps in-progress shards in that hidden directory and finalized shards land outside, so narrowing the delete still plugs the disk leak from #1260, and finalized shards stay visible for the user to keep (auto-resume on retry) or remove themselves. The hidden ones are the ones users can't see and need us to clean up.

mvanhorn:
Really appreciate the merge @jundot - cleaning up the partial shard on cancel kills a silent-corruption foot-gun that bites on Ctrl-C retries.

PR #1285, opened 2026-05-16
fix: accept both 'tool' and 'tool_name' on /v1/mcp/execute

### LINKED ISSUE #1113: web-search__brave_web_search returns "Tool not found" when called by pi agent — wrong field name in /v1/mcp/execute
**Describe the bug**
When using omlx launch pi with a model that supports tool calling, MCP tools are correctly listed in /v1/mcp/tools and the pi agent can see them. However, when the model attempts to call web-search__brave_web_search, it returns "Tool not found". Same happens for fetch & time mcp too.

**To Reproduce**
Root cause identified: The /v1/mcp/execute endpoint requires the field name tool_name, but the pi integration (or the model's tool call routing internally) sends tool instead. This can be confirmed by calling the endpoint directly:

Fails — wrong field name
curl -X POST http://localhost:8000/v1/mcp/execute \
  -d '{"tool": "web-search__brave_web_search", "arguments": {"query": "test"}}'
Returns: {"error": {"message": "body -> tool_name: Field required", ...}}

Works — correct field name
curl -X POST http://localhost:8000/v1/mcp/execute \
  -d '{"tool_name": "web-search__brave_web_search", "arguments": {"query": "test"}}'
Returns results correctly

**Expected behavior**
Tool calls from the pi agent (and any other client using the OpenAI tool_calls format) should correctly route to /v1/mcp/execute using the tool_name field.

**Desktop:**
oMLX version: 0.3.8
Model: unsloth/Qwen3.6-35B-A3B-MLX-8bit (type: llm, engine: batched)
MCP servers: web-search (Brave), fetch, time
Client: omlx launch pi

**Additional context**
Either accept both tool and tool_name as valid field names in the /v1/mcp/execute request body, or ensure the pi launcher passes tool_name when constructing tool execution requests.

mvanhorn:
## Summary

Two camps of MCP clients send the target tool as either `tool` or `tool_name` in the `/v1/mcp/execute` payload. omlx accepted `tool_name` only, so clients sending `tool` got a 422 even though their request was otherwise valid. Accepts both field names, with `tool_name` taking precedence when both are present.

## Why this matters

#1113 reports the field-name drift; the workaround is a wrapper that rewrites the field, which is friction for every new client. Aligns omlx with the de-facto behavior across MCP servers.

## Changes

- `omlx/api/openai_models.py` - accept either field on the MCP execute model.
- `tests/integration/test_server_endpoints.py` - tests covering `tool`-only, `tool_name`-only, both-present (precedence), and neither (still rejects).

## Testing

New integration tests + existing endpoint tests pass locally.

Fixes #1113

### DIFF
--- a/omlx/api/openai_models.py
+++ b/omlx/api/openai_models.py
@@ -13,7 +13,7 @@ These models define the request and response schemas for:
 import json
 from typing import Any, Dict, List, Optional, Union
 
-from pydantic import BaseModel, Field, field_validator
+from pydantic import AliasChoices, BaseModel, Field, field_validator
 
 from omlx.api.shared_models import (
     BaseUsage,
@@ -419,7 +419,9 @@ class MCPServersResponse(BaseModel):
 
 class MCPExecuteRequest(BaseModel):
     """Request to execute an MCP tool."""
-    tool_name: str
+    model_config = {"populate_by_name": True}
+
+    tool_name: str = Field(validation_alias=AliasChoices("tool_name", "tool"))
     arguments: dict = Field(default_factory=dict)
 
 
--- a/tests/integration/test_server_endpoints.py
+++ b/tests/integration/test_server_endpoints.py
@@ -18,6 +18,7 @@ from omlx.api.responses_utils import ResponseStore
 from omlx.engine.base import BaseEngine
 from omlx.engine.embedding import EmbeddingEngine
 from omlx.engine.reranker import RerankerEngine
+from omlx.mcp.types import MCPToolResult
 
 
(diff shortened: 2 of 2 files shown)


jundot:
Thanks! Clean use of `AliasChoices`, and the precedence ordering does what the PR body says. I'll merge and tack on a couple more test cases for the alias paths in a follow-up.

mvanhorn:
Cheers @jundot - thanks for landing the dual field-name acceptance so MCP clients on either side of the tool/tool_name split stop seeing 422s.

PR #1286, opened 2026-05-16
fix(tests): update Scheduler/Memory to_dict expectations for new fields

mvanhorn:
## Summary

`SchedulerSettings.to_dict()` and `MemorySettings.to_dict()` emit several fields (`chunked_prefill`, `soft_threshold`, `hard_threshold`) that their corresponding tests in `tests/test_settings.py` didn't include in the expected dict. The tests fail locally against current defaults. Updates the expected dicts to match the implementation.

## Why this matters

#1259 surfaced the staleness in passing. Test failures in this area mask any real regression in the settings serialization path.

## Changes

- `tests/test_settings.py` - update `to_dict` expectations for SchedulerSettings and MemorySettings to include the missing fields.

## Testing

Tests pass locally after the update; no other test changes needed.

Fixes #1259

PR #1268, opened 2026-05-15
fix(profiles): classify 9 new ModelSettings fields

richgoodson:
Fixes one of the 5 test failures listed in #1259.

`test_all_model_settings_fields_classified` is a guardrail that asserts
every field in `ModelSettings` is classified into one of
`UNIVERSAL_PROFILE_FIELDS`, `MODEL_SPECIFIC_PROFILE_FIELDS`, or
`EXCLUDED_FROM_PROFILES`. The 9 fields below were added to
`ModelSettings` without classification, causing the failure.

## Classification

| Field | Bucket | Rationale |
|---|---|---|
| `preserve_thinking` | UNIVERSAL | Generic reasoning behavior toggle; sits with `enable_thinking` and `thinking_budget_*` which are already universal. Docstring confirms it applies to any thinking-capable model. |
| `dflash_max_ctx` | MODEL_SPECIFIC | Dflash decoding fallback threshold. |
| `dflash_in_memory_cache` | MODEL_SPECIFIC | Dflash L1 prefix cache toggle. |
| `dflash_in_memory_cache_max_entries` | MODEL_SPECIFIC | Dflash L1 cache capacity. |
| `dflash_in_memory_cache_max_bytes` | MODEL_SPECIFIC | Dflash L1 cache byte budget. |
| `dflash_ssd_cache` | MODEL_SPECIFIC | Dflash L2 (SSD) prefix cache toggle. |
| `vlm_mtp_enabled` | MODEL_SPECIFIC | VLM multi-token prediction; mirrors `mtp_enabled` placement. |
| `vlm_mtp_draft_model` | MODEL_SPECIFIC | VLM MTP draft model path. |
| `vlm_mtp_draft_block_size` | MODEL_SPECIFIC | VLM MTP draft tokens per round. |

All `dflash_*` and `vlm_mtp_*` placements follow existing convention:
prior `dflash_*` entries and `mtp_enabled` are already in
`MODEL_SPECIFIC_PROFILE_FIELDS`. The `preserve_thinking` placement is
the one judgment call; happy to move it to `MODEL_SPECIFIC_PROFILE_FIELDS`
if the maintainer prefers.

## Verification

`pytest tests/test_admin_profiles_api.py -v`: 20/20 pass (was 19/20).

Issue #1259, opened 2026-05-14
FYI: some failing tests


fry69:
With current `main` (https://github.com/jundot/omlx/commit/51907f08074742defec4375fb629e289801a8a9f), when I run:

```shell
pytest -m "not slow"
```

as described in the README, some tests fail:

```
FAILED tests/test_admin_profiles_api.py::test_all_model_settings_fields_classified - AssertionError: New ModelSettings field(s) {'vlm_mtp_draft_block_size', 'dflash_in_memory_cache_max_bytes', 'preserve_thinking', 'dflash_max_ctx', 'dflash_ssd_cache', 'dflash_in_memory...
FAILED tests/test_mlx_lm_mtp_patch.py::TestBatchGeneratorDispatch::test_is_mtp_eligible_requires_mtp_forward_and_solo_batch - AssertionError: assert False is True
FAILED tests/test_settings.py::TestSchedulerSettings::test_to_dict - AssertionError: assert {'chunked_pre..._requests': 8} == {'max_concurrent_requests': 8}
FAILED tests/test_settings.py::TestMemorySettings::test_to_dict - AssertionError: assert {'hard_thresh...eshold': 0.85} == {'max_process..._guard': True}
FAILED tests/test_vlm_torch_free_image_processor.py::test_patch_wraps_target_processors - AssertionError: assert False
```

Since the whole test suite takes around 5 minutes to run, so I extracted these 5 repeatedly failing tests out. They can get run with:

```shell
pytest tests/test_admin_profiles_api.py tests/test_mlx_lm_mtp_patch.py tests/test_settings.py tests/test_vlm_torch_free_image_processor.py
```

This is just informational, not a bug report. Feel free to close if no longer relevant.

richgoodson:
Reproducing all 5 on `51907f0` from a fresh fork + editable install
(Apple Silicon M5, Python 3.13.12).

I hit one additional failure not in the list above:
`tests/test_boundary_snapshot_store.py::TestBoundarySnapshotSSDStore::test_cleanup_all_drains_queue`.
It passes 5/5 when run in isolation, so it appears to be order-dependent
(test pollution from a preceding test) rather than a standalone bug.
Flagging as a separate observation, not part of this issue.

Going to send a small PR for `test_all_model_settings_fields_classified`.
Looks like the 9 new fields just need classifying in
`omlx/model_profiles.py`. Will leave the others open for now.

richgoodson:
[...]

PR #1287, opened 2026-05-16
fix(tests): align two stale tests in #1259 with current implementation

richgoodson:
## Summary

Two of the failing tests called out in #1259 are stale against the
current implementation, not regressions in the code under test. Both
were already failing on `main` for me on a fresh editable install; this
PR brings them back in line with the code they're meant to cover.

Pairs with PR #1268 (`test_admin_profiles_api`) and #1286
(`test_settings`) to fully clear the #1259 list.

## Changes

**`tests/test_mlx_lm_mtp_patch.py::TestBatchGeneratorDispatch::test_is_mtp_eligible_requires_mtp_forward_and_solo_batch`**

The test was written against the pre-#1247 contract where head presence
implied MTP active. Commit
[23ca7dc](https://github.com/jundot/omlx/commit/23ca7dc) decoupled head
attachment from inference-time MTP for the VLM load path and added an
`is_mtp_active()` gate inside `_is_mtp_eligible`. The True-expected
assertion now fails because the process-wide flag is off.

Fix: toggle `mlx_lm_mtp.set_mtp_active(True)` around the True-expected
assertions, restore the prior value in `finally`, and add a new
"head attached but flag off" case to lock in the post-23ca7dc
semantics (this is the exact case the VLM runtime patches rely on).

**`tests/test_vlm_torch_free_image_processor.py::test_patch_wraps_target_processors`**

The test stubbed the wrong module path and class name for `dots_ocr`:

| | test stubbed | code imports |
|---|---|---|
| module | `mlx_vlm.models.dots_ocr.processing` | `mlx_vlm.models.dots_ocr.processing_dots_ocr` |
| class | `DotsOcrProcessor` | `DotsVLProcessor` |

The `fake_import` branch never matched, the dots branch silently fell
through to the `except (ImportError, AttributeError)` clause inside
`_patch_torch_free_image_processor`, and the wrap-marker assertion
failed. The mismatch has been there since the test was added in
[a1987ed](https://github.com/jundot/omlx/commit/a1987ed).

Fix: align the stubs with the actual `(module_path, cls_name)` tuples
in `omlx/engine/vlm.py:181-184`.

## Testing

Before:
```
$ pytest tests/test_mlx_lm_mtp_patch.py tests/test_vlm_torch_free_image_processor.py
... 2 failed, 52 passed
```

After:
```
$ pytest tests/test_mlx_lm_mtp_patch.py tests/test_vlm_torch_free_image_processor.py
... 54 passed in 0.21s
```

Apple Silicon M5, Python 3.13.12, fresh editable install on
`51907f0`.

## Scope

Test-only changes. No code under `omlx/` modified.

Refs #1259.

### DIFF
--- a/tests/test_mlx_lm_mtp_patch.py
+++ b/tests/test_mlx_lm_mtp_patch.py
@@ -331,6 +331,7 @@ class TestBatchGeneratorDispatch:
         assert hasattr(GenerationBatch, "_omlx_mtp_patched")
 
     def test_is_mtp_eligible_requires_mtp_forward_and_solo_batch(self):
+        from omlx.patches import mlx_lm_mtp
         from omlx.patches.mlx_lm_mtp.batch_generator import _is_mtp_eligible
 
         class _NonMtpModel:
@@ -338,14 +339,14 @@ class TestBatchGeneratorDispatch:
 
         class _MtpModelWithoutHead:
             """Has the patched method but no actual MTP head attached
-            (mtp_enabled was False when this hypothetical model loaded)."""
+            (config did not declare an MTP head when this model loaded)."""
 
             def mtp_forward(self, *_):
                 pass
 
         class _MtpModel:
-            """Has both the method and the attached head — i.e. mtp_enabled
-            was True at load time."""
+            """Has both the method and the attached head — i.e. the model
+            class was patched and the head was attached at load time."""
 
             def __init__(self):
                 self.mtp = object()  # placeholder for an actual MTPModule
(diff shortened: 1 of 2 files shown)


jundot:
Verified both fixes against current code. `_is_mtp_eligible` gates on `is_mtp_active()` at `omlx/patches/mlx_lm_mtp/batch_generator.py:222-224` (added in 23ca7dc), and `_patch_torch_free_image_processor` does import `processing_dots_ocr.DotsVLProcessor` at `omlx/engine/vlm.py:183`. The new "head attached but flag off" case is a nice lock-in for the post-23ca7dc semantics. Merging.

PR #1295, opened 2026-05-18
feat(oq): load presaved oQ sensitivity map from the source model folder

### LINKED ISSUE #972: Share sensitivity data artifacts across same model quantizations
Let's say I want to quantize a model all the way from oQ2 to oQ8. In my current understanding of the oQ process at least _some parts_ of the work are going to be repeated over and over, given that it's the same sensitivity model all the time. Like sensitivity ratio of the same tensors against the same quantization bits.

This would help me a lot with a massive benchmarking work.

Feel free to close this request if my assumption is incorrect.

deepsweet:
Closes https://github.com/jundot/omlx/issues/972.

I quantize to oQ _a lot_ using the provided sensitivity model method, and it's been always doing the same calibration over and over again, given the fact that the sensitivity model is the same every time.

This PR introduces a small check for a special `oq_sensitivity_map.json` file in a source model directory before handling the provided sensitivity model path or building an auto-proxy.

The file contains manually presaved sensitivity map:

```json
{ 
  "0": 0.004673895891755819,
  "1": 0.001716080354526639,
  "2": 0.004504043143242598
}
```

oQ-level doesn't play any role in case of `_measure_sensitivity_from_quantized_model()`, so the data is pretty static. And thus it saves a ton of GPU time – further quantizations happen in seconds.

At this point it's more like a hidden hack for myself, but I'd be happy to help with further development if you approve the general idea. Maybe it could be useful to auto-save the file somewhere in oMLX cache directory after the first quantization using an auto-proxy to avoid the calibration process next time.

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -2244,76 +2244,82 @@ def quantize_oq_streaming(
         f"{len(weight_files)} shards"
     )
 
-    from omlx.settings import get_system_memory as _get_system_memory
-    _model_bytes = all_weights.nbytes()
-    _system_ram = _get_system_memory()
-    _model_exceeds_ram = _model_bytes > int(_system_ram * _MAX_MODEL_RAM_FRACTION)
-    if _model_exceeds_ram:
-        logger.info(
-            f"oQ{oq_level:g}: model size ({_model_bytes / 1e9:.1f} GB) exceeds "
-            f"80% of system RAM ({_system_ram / 1e9:.1f} GB), "
-            "OOM-prone paths will be skipped"
-        )
+    sensitivity_map_path = Path(model_path, "oq_sensitivity_map.json")
 
-    cb("loading", 12.0)
-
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the contribution — clean change with good test coverage. The downstream contract holds: str-key normalization at oq.py:2367 and the empty-map enforcement at oq.py:2322 still catch malformed JSON, so cache hit fails closed.

One small nit I'll clean up in a follow-up commit: the diff moves `cb("loading", 12.0)` inside the else, so the cache-hit path skips that progress marker and the UI jumps from 8 → 15. I'll hoist it after merge.

Landing this in the next release.

deepsweet:
Published presaved maps for Qwen3.6-35B-A3B and Qwen3.6-27B https://huggingface.co/datasets/deepsweet/MLX-oQ-sensitivity-map in case someone else is interested.

PR #1297, opened 2026-05-18
Fix #1249: Handle unexpected 'strict' kwarg in mlx_lm load()

### LINKED ISSUE #1249: Model load fails with strict parameter mismatch for converted/abliterated models
The model:  Jiunsong/supergemma4-e4b-abliterated-mlx   https://huggingface.co/Jiunsong/supergemma4-e4b-abliterated-mlx

## Summary

Loading converted or abliterated MLX models fails with a hard 500 error when the weight file contains parameters that don't map to the instantiated model architecture, even though the extra weights are harmless artifacts of the conversion process.

## Error

```
POST /admin/api/models/supergemma4-e4b-abliterated-mlx/load → 500: Received 126 parameters not in model:
language_model.model.layers.24.self_attn.k_norm.weight,
language_model.model.layers.24.self_attn.k_proj.biases,
language_model.model.layers.24.self_attn.k_proj.scales,
language_model.model.layers.24.self_attn.k_proj.weight,
... (126 total, layers 24–41)
```

## Root cause

Call chain:

1. `omlx/engine/batched.py:201` — `from mlx_lm import load` called without `strict=False`
2. `mlx_lm/utils.py:415` — `model.load_weights(list(weights.items()), strict=strict)` defaults to `strict=True`
3. `mlx/nn/layers/base.py:185` — raises `ValueError` on any extra weight keys

The model ([`Jiunsong/supergemma4-e4b-abliterated-mlx`](https://huggingface.co/Jiunsong/supergemma4-e4b-abliterated-mlx)) is a 4-bit quantized, abliterated Gemma 4 variant. The abliteration process produced k/v projection weights for all 42 layers, but `config.json` has `num_kv_shared_layers: 18`, so the model class doesn't instantiate k/v modules for layers 24–41 — leaving 126 weight tensors with nowhere to go.

The weights are valid artifacts of conversion, not corruption. The model works correctly in other MLX runtimes that use `strict=False`.

## Suggested fix

**Option A** — pass `strict=False` when calling `mlx_lm.load` in `batched.py` (note: `mlx_lm.load` itself would also need to accept and forward a `strict` parameter, since it currently doesn't expose one):

```python
# omlx/engine/batched.py
def _load_model_sync():
    return load(
        self._model_name,
        tokenizer_config=tokenizer_config,
        strict=False,  # tolerate extra weights from converted/abliterated models
    )
```

**Option B** — catch the `ValueError` from `load_weights` and retry with `strict=False`, logging a warning. This keeps strict validation as the default while not hard-failing on valid converted models.

zyguy:
**Fixes #1249**

**Motivation:**
In `v0.3.9.dev2`, `omlx/engine/batched.py` attempts to pass `strict=False` to `mlx_lm.load()` to tolerate extra weights from converted/abliterated models. However, the upstream `mlx_lm.load()` function does not currently accept a `strict` argument. This results in a 500 error: `load() got an unexpected keyword argument 'strict'`.

**Solution:**
This PR uses `inspect.signature` to dynamically check if the installed version of `mlx_lm` supports the `strict` argument before passing it. This ensures compatibility with both older versions of `mlx_lm` and future versions where the argument is exposed (e.g., via the upstream PR I've opened on [ml-explore/mlx-lm](https://github.com/ml-explore/mlx-lm/pull/1284)).

### DIFF
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -237,10 +237,26 @@ class BatchedEngine(BaseEngine):
                 model, processor = custom_loaded
                 return model, getattr(processor, "tokenizer", processor)
 
-            return load(
-                self._model_name,
-                tokenizer_config=tokenizer_config,
-            )
+            try:
+                return load(
+                    self._model_name,
+                    tokenizer_config=tokenizer_config,
+                )
+            except ValueError as e:
+                if "parameters not in model" in str(e):
+                    import inspect
+
+                    load_sig = inspect.signature(load)
+                    if "strict" in load_sig.parameters:
+                        logger.warning(
+                            f"Caught parameter mismatch error. Retrying load with strict=False. Details: {e}"
+                        )
+                        return load(
+                            self._model_name,
+                            tokenizer_config=tokenizer_config,
+                            strict=False,
+                        )
+                raise
 
         loop = asyncio.get_running_loop()
         self._model, self._tokenizer = await loop.run_in_executor(
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -614,9 +614,25 @@ class VLMBatchedEngine(BaseEngine):
                     model, processor = custom_loaded
                     return model, processor
 
-                return vlm_load(
-                    self._model_name, trust_remote_code=self._trust_remote_code
-                )
+                try:
+                    return vlm_load(
+                        self._model_name, trust_remote_code=self._trust_remote_code
+                    )
+                except ValueError as e:
+                    if "parameters not in model" in str(e):
+                        import inspect
+
+                        load_sig = inspect.signature(vlm_load)
+                        if "strict" in load_sig.parameters:
+                            logger.warning(
+                                f"Caught parameter mismatch error. Retrying vlm_load with strict=False. Details: {e}"
+                            )
+                            return vlm_load(
+                                self._model_name,
+                                trust_remote_code=self._trust_remote_code,
+                                strict=False,
+                            )
+                    raise
 
         loop = asyncio.get_running_loop()
         self._vlm_model, self._processor = await loop.run_in_executor(


jundot:
Thanks for the patch. The signature-guarded shape is the right direction for forward-compat. I want to hold the merge for now though, for two reasons:

1. omlx currently pins mlx-lm at ed1fca4, which does not have `strict` in `load()`. So in the version oMLX ships today, this PR is a no-op — the `if "strict" in load_sig.parameters` branch never fires until I bump the pin after ml-explore/mlx-lm#1284 lands.

2. The model in the original report (supergemma4-e4b-abliterated-mlx) is `model_type=gemma4`, which routes through VLMBatchedEngine and calls `mlx_vlm.utils.load`, not `mlx_lm.load`. mlx-vlm does not expose `strict` on its `load()` either, so the abliterated-weight case will stay broken on the VLM path even after this PR. Same situation for mlx-embeddings.

My preference is to wait until both mlx-lm and mlx-vlm expose `strict` upstream, then land the omlx-side change in one go so issue #1249 actually closes end-to-end. Would you be up for opening a sibling PR against mlx-vlm? I am happy to coordinate.

One small forward note when we do land it: `strict=False` as a blanket default also silences missing/shape/type mismatches, not just extra-key ones. The Option B pattern from the issue (try strict=True, catch the "parameters not in model" ValueError specifically, then retry with strict=False + warning) keeps the safety net for everything except the abliterated case. Worth considering for the eventual merged version.

zyguy:
I completely agree with the approach!

I have updated this PR to implement the "Option B" fallback strategy for both mlx_lm (omlx/engine/batched.py) and mlx_vlm (omlx/engine/vlm.py). It now attempts a standard load first, specifically catches a
ValueError containing "parameters not in model", checks for strict support via inspect, and retries with strict=False while logging a warning. This safely preserves the error reporting for actual missing weights
or shape mismatches.

I have also opened the sibling PR on mlx-vlm as requested: https://github.com/Blaizzy/mlx-vlm/pull/1198

Great news: Testing this updated logic locally, I can confirm it works perfectly end-to-end! I am now able to successfully load and run both gemma-4-e4b-it-OptiQ-4bit and the problematic
supergemma4-e4b-abliterated-mlx without hitting any 500 errors. 🚀

Let me know if there's anything else you need before this is ready to merge once the upstream pins are bumped!

PR #1298, opened 2026-05-18
fix(cache): pending write buffer for hot cache eviction race

### LINKED ISSUE #1232: Qwen3.6 27B MXFP4 cache missing
**Describe the bug**

When hot cache is enabled, cache hits fail for blocks that were evicted from hot cache but whose SSD write hasn't completed yet. `load_block_with_metadata()` returns `None`, forcing full prefill.

**Root cause:** `_enqueue_ssd_write()` adds the block to the SSD index and marks it in `_pending_write_hashes`, but the block data was already removed from `_hot_cache` during eviction. The load path never checks `_pending_write_hashes` — it goes straight from hot-cache-miss to SSD-file-lookup, and the file doesn't exist yet. Between eviction and write completion, the block is nowhere readable.

(The hot-cache-disabled path sidesteps this by stashing blocks directly in `self._hot_cache` as a temp buffer — it works but repurposes the hot cache data structure for something other than hot caching.)

**To Reproduce**

1. Load a hybrid-cache model with hot cache enabled (tested with [sleepyeldrazi/Qwen3.6-27B-MXFP4-MTP](https://huggingface.co/sleepyeldrazi/Qwen3.6-27B-MXFP4-MTP) and [mlx-community/Qwen3.6-27B-mxfp4](https://huggingface.co/mlx-community/Qwen3.6-27B-mxfp4))
2. Send a first request with a long system prompt to populate the prefix cache
3. Send enough requests that hot cache fills and LRU eviction fires
4. Send another request with the same system prompt prefix
5. Observe: `load_block_with_metadata` returns `None` for the evicted block, request re-runs full prefill

**Expected behavior**

A block queued for SSD write should remain readable until the write completes. After hot cache eviction, the load path should find the block in a pending-write buffer and serve it with zero I/O.

 - Mac Studio M4 Max 36GB
 - macOS Version: 15.3
 - oMLX Version 0.39dev2

**Additional context**

Fix: maintain `_pending_write_buffers: Dict[bytes, Dict]` alongside `_pending_write_hashes`. When a block is queued for SSD write (eviction or direct save), store `tensors_raw` there. The load path checks it between hot cache and SSD — a pending-write hit is zero I/O, same raw-bytes path as hot cache reads. The background writer removes entries after write completes. This also unifies the hot-cache-disabled path, removing the `_hot_cache` hack.

One concern: the buffer has no size cap. If the write queue backs up under I/O saturation, memory grows. A bound or LRU cap would be worth adding.

The above text was generated by my LLM after cross-referencing my local fork (early MTP PR testing and some kv cache optimizations i was trying). I noticed the MTP PR was finally merged and wanted to test it, but encountered this issue. I have a lot of other unrelated changes too, but my fork/branch used for the comparison (and the cache hits on my branch in my testing) can be reached on [my local Forgejo](https://git.kokoham.com/sleepy/omlx/src/branch/stable/pre-q4kv)

ivaniguarans:
## Summary

When hot cache is enabled, blocks evicted from the hot cache become unreachable during the window between eviction and SSD write completion. `load_block()` hits the SSD index, finds no file on disk (write still pending), removes the index entry, and returns `None` — forcing a full reprefill that shouldn't be necessary.

This adds a `_pending_write_buffers` dict that holds evicted entries until the background writer completes. The pattern matches what `boundary_snapshot_store.py` already does with its `_pending_writes` dict.

## Root Cause

The race sequence:
1. `_hot_cache_put()` evicts LRU entry from `_hot_cache`
2. `_enqueue_ssd_write()` adds block to SSD index and write queue
3. Block data now lives only in the write queue tuple — not in hot cache, not on disk
4. Concurrent `load_block()`: hot cache miss → SSD index hit → `file_path.exists()` false → removes from index, returns `None`

## Changes

**Write path** — `_enqueue_ssd_write()` reordered to buffer → index → queue. The buffer is populated before the index entry, so a concurrent load never sees an index hit for a block with no backing data. Queue-full rollback cleans all three.

**Read path** — Both `load_block()` and `load_block_with_metadata()` check the pending buffer after a hot cache miss, before the SSD index/file check. Hits are zero-I/O and counted as `hot_cache_hits`.

**`has_block()`** — Now checks the pending buffer, so callers in `scheduler.py` and `tiered_manager.py` don't miss blocks awaiting write.

**`delete_block()`** — Cleans pending buffer entries so deleted blocks aren't served during the write window.

**Writer cleanup** — `_writer_loop()` finally block removes buffer entries alongside `_pending_write_hashes`. `close()` clears the buffer defensively in case the writer times out.

## Test plan

7 new tests in `TestPendingWriteBuffer`:
- Evicted block readable from pending buffer (both `load_block` and `load_block_with_metadata`)
- Writer cleanup empties buffer after write completion
- `has_block()` sees pending entries
- `delete_block()` clears pending buffer
- Queue-full handler removes dropped block from buffer
- End-to-end lifecycle: save → evict → pending hit → writer completes → SSD hit

29/29 hot cache tests pass. 2 pre-existing failures in unrelated test files (`test_admin_profiles_api`, `test_mlx_lm_mtp_patch`).

Closes #1232

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -668,6 +668,12 @@ class PagedSSDCacheManager(CacheManager):
         # Track which block hashes are queued for background write
         self._pending_write_hashes: set = set()
         self._pending_write_hashes_lock = threading.Lock()
+        # Lock ordering invariant: _hot_cache_lock -> _pending_write_hashes_lock.
+        # Never acquire in reverse. Load path: _hot_cache_get (holds _hot_cache_lock,
+        # releases), then _pending_write_buffer_get (holds _pending_write_hashes_lock).
+        # Eviction path: _hot_cache_put (holds _hot_cache_lock, releases), then
+        # _enqueue_ssd_write (holds _pending_write_hashes_lock).
+        self._pending_write_buffers: dict[bytes, dict] = {}
         self._writer_shutdown = threading.Event()
         # Writer thread is only needed when writing to SSD.
         self._writer_thread = None
@@ -769,13 +775,19 @@ class PagedSSDCacheManager(CacheManager):
             return False
         metadata = entry["file_metadata"]
 
-        # Add to SSD index now that block is being written to SSD
+        # 1. Buffer first — instant read-back for concurrent loads (CPD K1).
+        #    Must precede _index.add so load_block never sees an index hit
+        #    for a block that has no file and no buffer entry yet.
+        with self._pending_write_hashes_lock:
+            self._pending_write_buffers[block_hash] = entry
+            self._pending_write_hashes.add(block_hash)
+
+        # 2. Index second — makes the block discoverable in has_block/contains.
         if not self._index.contains(block_hash):
             self._enforce_size_limit_for_new_block()
             self._index.add(blk_meta)
 
-        with self._pending_write_hashes_lock:
-            self._pending_write_hashes.add(block_hash)
+        # 3. Queue third — enqueue for background writer.
         try:
             item = (block_hash, tensors_raw, metadata, file_path)
             if blocking:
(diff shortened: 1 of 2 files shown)


jundot:
nice, buffer-first ordering is the right fix and the lock ordering doc inside __init__ makes the invariant obvious. Ran the new TestPendingWriteBuffer (7) + existing TestHotCache* (29), all pass. Merging.

PR #1301, opened 2026-05-19
perf(cache): parallel SSD-to-hot-cache block preloading

ivaniguarans:
## Summary

When the prefix cache matches blocks on SSD, each block is loaded into the hot cache sequentially — one `mx.load()` call after another, even though the blocks are independent and could be read concurrently. This is a bottleneck that grows linearly with cache hit size: at N=10 blocks it adds ~80ms to TTFT, but for long-context workloads matching 50-100+ blocks the sequential penalty reaches hundreds of milliseconds.

`preload_matched_blocks()` replaces this sequential pattern with parallel I/O via `ThreadPoolExecutor`, loading all cold blocks into the hot cache concurrently before `reconstruct_cache()` runs. The blocks are independent (separate files, no ordering dependency), so parallel loading is the natural approach. When reconstruction later calls `load_block()`, each block is already hot — an O(1) dict lookup instead of a disk read. No additional memory is used (the blocks would be loaded regardless), and any per-block failure falls back silently to the existing sequential path.

## Changes

- **`PagedSSDCacheManager.preload_matched_blocks()`** — parallel SSD→hot-cache loader using `ThreadPoolExecutor(max_workers=min(8, N))`. Five guard clauses skip preload when it can't help: hot cache disabled, no MLX, all blocks already hot, N<4 cold blocks, hot cache full. Per-block failures are caught individually and fall through to sequential load during reconstruction.
- **N<4 threshold** — thread pool overhead dominates at small N. Micro-benchmarks confirm N=1 is 0.67x (slower), N=2 is 1.0x (break-even), N=4+ shows consistent gains.
- **`BlockAwarePrefixCache.preload_blocks()`** — thin wrapper that delegates to the SSD manager.
- **Scheduler integration** — calls `preload_blocks()` after `fetch_cache()` returns matched blocks, in both the primary prefill path and the SpecPrefill draft path. Two lines added.
- **Stats tracking** — `preload_calls`, `preload_blocks_loaded`, `preload_time_ms`.

## Benchmark results

All benchmarks ran on M5 Max 128GB with SSD cache on external WD SN850X (Thunderbolt). Server restarted between every e2e iteration (cold hot cache, warm SSD cache). Preload toggled via temporary env var (removed before submission).

**Micro-benchmark** (synthetic blocks, cache I/O layer only):

| N blocks | Sequential | Preload+Hot | Speedup |
|----------|----------:|------------:|--------:|
| 1        |     4.2ms |       6.2ms |  0.67x  |
| 2        |     9.8ms |       9.9ms |  1.00x  |
| 4        |    15.0ms |      10.2ms |  1.47x  |
| 8        |    14.6ms |       9.9ms |  1.48x  |
| 16       |    14.8ms |       9.8ms |  1.51x  |

N=1-2 confirms the N<4 threshold is correctly placed.

**End-to-end TTFT** (~22K-token prompt, `max_tokens=1`):

| Model | Type | Blocks | With preload | Without | Delta |
|-------|------|-------:|------------:|--------:|------:|
| Qwen3-0.6B-4bit | Dense (block_size=256) | ~7 | 367.9ms | 388.2ms | **-20.3ms (5.2%)** |
| Qwen3.6-27B-oQ6-mtp | Dense (block_size=2048) | 10 | 3175.0ms | 3255.4ms | **-80.4ms (2.5%)** |
| Qwen3.6-35B-A3B-oQ8-mtp | MoE (block_size=2048) | 10 | 1443.6ms | 1481.5ms | **-37.9ms (2.6%)** |

Savings scale linearly with block count. The tested models use block_size=2048 (ArraysCache/GatedDeltaNet), producing 10 blocks from a 22K-token prompt. Standard transformer models use block_size=256, where the same prompt produces ~86 blocks and the cumulative savings are proportionally larger.

The current scaling limit at high N is `_promote_to_hot_cache` serializing on `_hot_cache_lock` — a batch-promote optimization could address this in a follow-up.

## Test plan

- 87 cache tests pass (11 new: edge cases, concurrency, stats, threshold behavior, integration with `BlockAwarePrefixCache`)
- Full suite: 2184 passed, 2 pre-existing failures unrelated to this change (`test_admin_profiles_api` — unclassified new fields; `test_mlx_lm_mtp_patch` — MTP eligibility check)

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -26,6 +26,7 @@ import struct
 import threading
 import time
 from collections import OrderedDict
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
@@ -648,6 +649,9 @@ class PagedSSDCacheManager(CacheManager):
             "hot_cache_hits": 0,
             "hot_cache_evictions": 0,
             "hot_cache_promotions": 0,
+            "preload_calls": 0,
+            "preload_blocks_loaded": 0,
+            "preload_time_ms": 0.0,
         }
 
         # --- Hot cache (in-memory raw-bytes tier) ---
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for the careful writeup. The micro+e2e benchmarks plus the N<4 threshold justification make this easy to review. Five guard clauses, per-block fall-through, and the 11 new tests covering partial failure / hot cache full / concurrent access cover the obvious edge cases. Call-site change is minimal too. Will smoke-test on real server and merge.

PR #1302, opened 2026-05-19
fix(tests): capture expected RuntimeWarning in fallback tests

fry69:
This captures the "Falling back to generic tokenizer config" runtime warning in

```shell
pytest tests/test_deepseek_v4_patch.py
```

With this fix the current `pytest -m "not slow"` is completely green.


jundot:
Thanks for catching this. Confirmed the two fallback tests were leaking the RuntimeWarning into the summary on main, and wrapping them in `pytest.warns(...)` both silences the noise and asserts the fallback path actually warns. Merging.

Issue #1248, opened 2026-05-13
Migrate _global_mlx_executor to per-engine threads now that mlx-lm 0.31.3 supports per-instance streams


cebertowicz:
**Is your feature request related to a problem? Please describe.**
Multi-model agent workflows — running a fast planner alongside a heavier executor concurrently — currently serialize through a single shared _global_mlx_executor with max_workers=1 (omlx/engine_core.py:32-78). All MLX GPU operations across all loaded engines funnel through that one thread, so two engines generating at the same time can't actually run in parallel; they take turns. The comment in that file cites mlx-lm issue #85 and the fact that mlx_lm.generate.generation_stream was a module-level shared object that would race if touched from multiple threads.

**Describe the solution you'd like**
Migrate the engine-execution path from the shared _global_mlx_executor to per-engine threads, each with its own mx.Stream, now that mlx-lm 0.31.3 (released 2026-04-22) supports this. Specifically:
- Replace the global singleton executor with a per-EngineCore ThreadPoolExecutor(max_workers=1), so each loaded model has its own dedicated MLX thread.
- Instantiate each engine's BatchGenerator with its own stream via the new BatchGenerator(stream=...) parameter introduced in mlx-lm PR #1090 (mlx_lm/generate.py:1486-1523).
- Rely on the new thread-local default generation_stream = mx.new_thread_local_stream(...) (same PR) for any indirect mlx-lm internals that still read it.

The mlx-lm 0.31.3 release notes list this under highlights as 'Thread local generation stream to accompany MLX v0.31.2.'

**Describe alternatives you've considered**
- Serializing at the application layer. 
- Forking and patching locally.
- Waiting for further upstream changes. Unclear what else would be needed beyond what 0.31.3 already shipped; this issue is partly to ask whether there are other shared-state hazards in BatchGenerator or the cache layer that still require single-thread serialization.

**Additional context**
n/a

Thanks for all the great work on oMLX!

PR #1304, opened 2026-05-19
fix(engine): per-engine threads to eliminate cross-engine stream contamination

ivaniguarans:
## Summary

When multiple LM engines run concurrently, they share a single `_global_mlx_executor` with `max_workers=1`, serializing all `scheduler.step()` calls through one thread. More critically, the MTP patch reads the module-level `generation_stream` via `sys.modules` for its forward passes, bypassing whatever stream the `BatchGenerator` was instantiated with. If two MTP-capable engines run simultaneously, their MTP forwards land on the same module-level stream regardless of which engine dispatched them — a stream-ordering violation that upstream's `BatchGenerator(stream=...)` parameter (mlx-lm 0.31.3) was designed to prevent.

This PR gives each `EngineCore` its own `ThreadPoolExecutor` and `mx.Stream`, passes the stream through `Scheduler` into `BatchGenerator`, and removes the `_get_generation_stream()` indirection from the MTP patch so MTP operations inherit the correct per-engine stream from the enclosing `BatchGenerator` context — the same pattern upstream's `GenerationBatch._step()` already uses.

The global executor is retained for non-LM engines (TTS, STT, embedding, reranker) that still rely on `get_mlx_executor()` and `_init_mlx_thread`.

## Changes

- **`engine_core.py`**: `EngineCore.__init__` creates a per-engine `ThreadPoolExecutor` + `mx.new_thread_local_stream()` and passes the stream to `Scheduler`. `close()` shuts down the per-engine executor after scheduler cleanup. Added `_ensure_wired_limit()` so the process-global `mx.set_wired_limit()` runs once rather than racing across concurrent `BatchGenerator` inits.
- **`scheduler.py`**: `Scheduler.__init__` accepts an optional `stream` parameter (falls back to the module-level `generation_stream` when not provided). All 37 internal references to `generation_stream` — sync barriers, cache clears, `mx.stream()` context managers, `BatchGenerator` creation — now use `self._stream`.
- **`batch_generator.py` (MTP patch)**: Removed `_get_generation_stream()` and the 4 explicit `with mx.stream(...)` wrappers that pushed the module-level stream. MTP forwards now inherit the per-engine stream from the enclosing `BatchGenerator` context, matching `GenerationBatch._step()`'s existing pattern.

## Concurrent throughput

Two models generating simultaneously vs sequentially:

| Model pair | Before (shared executor) | After (per-engine) | Speedup |
|---|---|---|---|
| Qwen3-0.6B + Qwen3-Coder-Next-6bit | 1.00x | 1.14x | 0.6B TTFT: 2089 ms → 701 ms |
| Qwen3.6-35B-A3B-oQ4-mtp + Qwen3.6-27B-oQ6-mtp | 0.93x | 1.12x | wall: 5408 ms → 4722 ms |

Sub-2x is expected — Metal command buffers still serialize on one GPU. The win is CPU-side overlap (prefill + decode can be submitted concurrently) and eliminating head-of-line blocking where one engine's long prefill stalls another's token emission.

## Test plan

- New `tests/test_per_engine_threads.py` (10 tests): verifies `Scheduler` stores and uses explicit streams, regex-scans the `Scheduler` class body for bare `generation_stream` references, confirms each `EngineCore` gets a distinct executor/stream, validates executor shutdown on `close()`, and asserts the MTP patch no longer contains `_get_generation_stream` or any `generation_stream` reference.
- Updated `tests/test_engine_core.py`: existing executor tests now assert `is not` (distinct executors) and concurrent execution (both executors active simultaneously).
- Full suite passes: 4493 passed, 19 skipped.
- Live-tested with concurrent Qwen3.6-35B-A3B-oQ4-mtp + Qwen3.6-27B-oQ6-mtp (both MTP-enabled) serving requests simultaneously.

Related to #1248

### DIFF
--- a/omlx/engine_core.py
+++ b/omlx/engine_core.py
@@ -78,6 +78,22 @@ def get_mlx_executor() -> concurrent.futures.ThreadPoolExecutor:
     return _global_mlx_executor
 
 
+_wired_limit_set = False
+
+
+def _ensure_wired_limit() -> None:
+    """Set Metal wired memory limit once at first engine creation.
+
+    BatchGenerator normally calls mx.set_wired_limit() per-instance, which
+    races when multiple engines init concurrently (process-global setting).
+    We call it once here instead.
+    """
+    global _wired_limit_set
+    if not _wired_limit_set and mx.metal.is_available():
+        mx.set_wired_limit(mx.device_info()["max_recommended_working_set_size"])
+        _wired_limit_set = True
+
+
 @dataclass
 class EngineConfig:
     """Configuration for the engine."""
@@ -133,12 +149,23 @@ class EngineCore:
         )
         self._owns_model = True
 
-        # Create scheduler
+        # Per-engine executor with dedicated mx.Stream (#1248).
+        # Each EngineCore gets its own thread + GPU stream so different
+        # models can run scheduler.step() concurrently.
+        _ensure_wired_limit()
+        self._mlx_stream = mx.new_thread_local_stream(mx.default_device())
+        self._mlx_executor = concurrent.futures.ThreadPoolExecutor(
+            max_workers=1,
+            thread_name_prefix=f"mlx-engine-{self._engine_id[:8]}",
+        )
+
+        # Create scheduler with per-engine stream
         scheduler_config = self.config.scheduler_config or SchedulerConfig()
         self.scheduler = Scheduler(
             model=model,
             tokenizer=tokenizer,
             config=scheduler_config,
+            stream=self._mlx_stream,
         )
 
         # Output collectors for low-latency streaming (vLLM pattern)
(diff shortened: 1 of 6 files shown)


jundot:
Thanks for putting this together, the analysis is spot on. The per-engine stream isolation is clearly the right fix for the cross-engine contamination, and the writeup with the concurrent throughput numbers and the regex-based tests is really thoughtful work.

My only hesitation is the scope. This touches the scheduler core and the MTP patch with 37 stream references rewritten, and I want to give it more soak time than the 0.3.9 release window allows. So I am going to hold this out of the 0.3.9 stable release and pull it in starting from the next dev build instead.

I know you have been waiting on this, sorry for the delay and thanks for your patience. The change itself looks solid, I just want to land it where it can get proper testing before it hits a stable release.

jundot:
Rebased locally to absorb two small fixups: dropped the 5th `_get_generation_stream()` call inside `_reconcile_mtp_to_standard` (added in 6f927ec after this PR's base) and routed the new `_safe_sync_generation_stream()` in `_async_store_cache_worker` (#1437) through the same `_safe_sync_stream(self._stream)` helper.

The MTP-via-enclosing-context claim checks out — `BatchGenerator.next` wraps `_next` in `mx.stream(self._stream)` in mlx-lm 0.31.3. Merging now into the next dev build.

Thanks again for the careful writeup and the perf numbers.

PR #1313, opened 2026-05-19
fix(model_discovery): classify Mistral-based CausalLM rerankers

contrapuntal:
## Summary

One-line classifier addition that recognizes Mistral-based CausalLM rerankers (Contextual AI Reranker v2 family) as rerankers instead of plain LLMs.

Pre-fix symptom: `ContextualAI/reranker_v2_6b` (and quantized variants such as `ctxl-rerank-v2-6b-bf16`) classifies as `llm`. The model carries no chat_template, so chat-mode loading produces a 500; loading it as a SentenceTransformer-style reranker fails because the on-disk arch is `MistralForCausalLM`, not `*ForSequenceClassification`. oMLX already supports CausalLM-style reranking through `CAUSAL_LM_RERANKER_ARCHITECTURES` — the set just omits `MistralForCausalLM`.

## Fix

`omlx/model_discovery.py`. One-line addition: `MistralForCausalLM` joins `CAUSAL_LM_RERANKER_ARCHITECTURES`.

Detection still requires the existing basename heuristic (`_is_causal_lm_reranker`, matching `"rerank"`/`"reranker"` in `model_path.name`), so plain Mistral chat models keep classifying as `llm` — only directories named like rerankers (e.g., `ctxl-rerank-v2-6b-bf16`, `*-Reranker-*`) trigger reranker classification. This mirrors the existing pattern already in the set for `Qwen3ForCausalLM`.

## Verification

`ContextualAI/reranker_v2_6b` (Contextual AI Reranker v2 6B, `MistralForCausalLM`, dir `ctxl-rerank-v2-6b-bf16`) classifies as `reranker` after this fix and serves correctly via the `/v1/rerank` endpoint. Plain Mistral chat models (no `rerank` in the basename) keep classifying as `llm`.

## Scope and risk

- Files: `omlx/model_discovery.py` (≈ +1 / −0).
- Risk: minimal. Additive entry in an existing set, gated by an existing basename heuristic. No public API changes; no schema changes; no new dependencies.
- The companion PR (VLM classifier + adapter layers fallback) is independent and lives separately.

### DIFF
--- a/omlx/model_discovery.py
+++ b/omlx/model_discovery.py
@@ -123,6 +123,7 @@ SUPPORTED_RERANKER_ARCHITECTURES = {
 # Detected by architecture + heuristic (model name or tokenizer hints).
 CAUSAL_LM_RERANKER_ARCHITECTURES = {
     "Qwen3ForCausalLM",
+    "MistralForCausalLM",  # e.g. ContextualAI/reranker_v2_6b (dir-name gated)
 }
 
 # CausalLM-based embedding architectures.


contrapuntal:
Closing — adversarial review (codex + gemini) plus inspection of `ContextualAI/reranker_v2_6b`'s actual inference contract surfaced that routing-only is the wrong scope. The classifier change does make `detect_model_type` return `"reranker"`, but the reranker engine that handles the routed request fails for this model on at least four independent grounds:

1. `MLXRerankerModel._build_from_pretrained` calls `tokenizer.apply_chat_template(...)` (`omlx/models/reranker.py:303`). The Contextual AI Reranker v2 tokenizer ships no `chat_template`, so engine init raises before any scoring runs — the same wall the LLM-mode path hits, just one layer later.

2. The same engine resolves `yes` / `no` token ids via `tokenizer.convert_tokens_to_ids("yes")` / `"no"` (lines 286-287). Mistral's SentencePiece tokenizer encodes leading-space variants (`▁yes`) for word-start; bare `"yes"` returns UNK. Even if the chat_template check were patched, scoring would be meaningless.

3. The suffix appends `<think>\n\n</think>\n\n` (line 314) — Qwen3-specific thinking-then-answering format, not what Mistral-based rerankers expect.

4. Most fundamentally: per the model card, Contextual AI Reranker v2's documented inference uses a **custom logits processor that constrains the first generated token**, then reinterprets the token id's bytes as a bf16 float to recover a continuous score (the README shows scores like `-2.2969`, `-4.6875`). oMLX's reranker engine compares `logit[yes]` vs `logit[no]` — incompatible at the math level, not just the prompt level.

Net: making this work properly requires a new engine type (custom prompt formatter + bf16-token-as-score decoder), not a one-line classifier addition. Documenting as a known-unsupported family at the oMLX layer; will file a separate feature request if anyone actually needs Contextual AI Reranker v2 support.

Branch left as historical reference.

PR #1336, opened 2026-05-21
fix(mtp): check real sampler temperature for greedy detection

### LINKED ISSUE #1330: MTP produces nondeterministic output at temperature=0
## Issue

With MTP enabled, identical prompts at `temperature=0` produce different output across runs. With MTP disabled, the same prompt is bit-identical run-to-run. A correctly-implemented lossless MTP at temp=0 (greedy acceptance) should be reproducible and should match the AR output — neither holds.

Sampling also dominates wall-clock — ~92% of decode time, ~56 ms/token — which is ~1000× higher than expected for a softmax+argmax on this hardware. Net MTP throughput is only ~1.2× over AR despite ~77% draft acceptance.

```
MTP[0] finish=stop tokens=3919 cycles=2218 accept=1700/2218 (76.6%)
       timing[backbone=7922.7ms  mtp=10768.8ms  sample=220996.5ms  cache=625.2ms]
```

## Suspected cause

Both symptoms appear to share a root in `omlx/patches/mlx_lm_mtp/batch_generator.py`:

**`_is_greedy` (lines 357-361)** uses PR 990's `sampler is None` heuristic to choose between greedy and Leviathan stochastic acceptance. That heuristic is correct upstream because `mlx_lm.generate` normalizes a missing sampler to argmax at the entry point. oMLX's scheduler doesn't share that contract — it installs an `omlx_make_sampler(temp=...)` callable unconditionally (`scheduler.py:1479, 2237`), even at temp=0. So `samplers[0]` is never `None`, `_is_greedy` always returns `False`, and the stochastic branch runs at every temperature.

That branch at line 799 then calls Python's `random.random()`:

```python
accept = log_accept >= 0 or random.random() < math.exp(log_accept)
```

`random` is never seeded by oMLX (`mx.random.seed` is called, but stdlib `random` isn't) — so acceptance decisions vary run to run, producing nondeterministic output even at temp=0.

The sampler dominance looks like a separate but adjacent issue in the same code path: the verify cycle uses synchronous `mx.eval` + multiple `.tolist()`/`.item()` calls per cycle instead of the `mx.async_eval` pipelining pattern the AR `_step` uses. Each sync stalls until the just-dispatched GPU work finishes; 2218 cycles × ~6 syncs each ≈ the observed 221 s.

`omlx/utils/sampling.py:175` already exposes `.temp` on the returned callable, so the greedy-detection fix looks like a one-liner.

## Environment

- oMLX v0.3.9rc1 (DMG)
- Apple M2 Max, 64 GB
- Model: `Jundot/Qwen3.6-27B-oQ6-mtp` (`model_type=qwen3_5`, 64 layers, hybrid GatedDeltaNet + standard attention), loaded as VLM, text-only generation
- Force Sampling: off
- Logs confirm `MTP path activated for uid=0 (model has mtp_forward, batch=1)`
- Acceptance rates healthy (76.6% / 79.9% on two runs), so MTP math itself looks fine — the divergence is in the sampling/acceptance path

Happy to provide more detail or a reproducer if useful.

deepsweet:
Fixes #1330.

Seems like we've always had `is_greedy = False` because there was always a sampler. This PR fixes it by adding an explicit check for its temperature + reusing the already existing helper `_resolve_sampler`.

New tests added:

_resolve_sampler:                                                                                            
 - Returns samplers[0] when present and not None                                                                                                    
 - Falls back to fallback_sampler when samplers[0] is None                                                                                          
 - Falls back to fallback_sampler when samplers is empty                                                                                            
 - Falls back to fallback_sampler when samplers is None/missing                                                                                     
 - Prefers samplers[0] over fallback_sampler when both exist                                                                                        
                                                                                                                                                    
_is_greedy:                                                                                        
 - Returns True (greedy) when sampler.temp == 0.0                                                                                                   
 - Returns False (not greedy) when sampler.temp > 0.0                                                                                               
 - Returns True when sampler has no temp attribute (defaults to 0.0)                                                                                
 - Returns True when sampler is None (falls back to greedy)                                                                                         
 - Returns True when samplers is empty/missing                                                                                                      
 - Checks fallback_sampler.temp when samplers[0] is None                                                                                            
 - Handles fallback_sampler with no temp attribute (defaults to greedy)

### DIFF
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -354,10 +354,10 @@ def _resolve_sampler(gen_batch: Any):
     return gen_batch.fallback_sampler
 
 
-def _is_greedy(gen_batch: Any) -> bool:
-    """Heuristic mirroring PR 990's ``sampler is None``."""
-    if gen_batch.samplers and gen_batch.samplers[0] is not None:
-        return False
+def _is_greedy(gen_batch):
+    sampler = _resolve_sampler(gen_batch)
+    if sampler is not None:
+        return getattr(sampler, "temp", 0.0) == 0.0
     return True
 
 
--- a/tests/test_mlx_lm_mtp_patch.py
+++ b/tests/test_mlx_lm_mtp_patch.py
@@ -566,3 +566,179 @@ class TestPreLoadPatchDispatch:
             json.dumps({"model_type": "qwen3_5", "mtp_num_hidden_layers": 1})
         )
         maybe_apply_pre_load_patches(str(tmp_path))
+
+
+# ---------------------------------------------------------------------------
+# batch_generator — _resolve_sampler + _is_greedy
+# ---------------------------------------------------------------------------
+
+class TestResolveSampler:
+    """Tests for ``_resolve_sampler`` which mirrors GenerationBatch._step's
+    per-sequence sampler resolution (batch=1).
+    """
+
+    @pytest.fixture(autouse=True)
+    def _apply(self):
+        from omlx.patches.mlx_lm_mtp import batch_generator
+
+        applied = batch_generator.apply()
+        if not applied:
+            pytest.skip("batch_generator patch refused to apply (mlx_lm absent)")
+
+    def _make_batch(self, samplers=None, fallback_sampler=None):
+        return SimpleNamespace(
+            samplers=samplers,
+            fallback_sampler=fallback_sampler,
+        )
+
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for tracking this down. Confirmed: oMLX's scheduler installs a sampler unconditionally (`scheduler.py:1479, 2237`) even at temp=0, so the old `samplers[0] is None` heuristic never reported greedy and the unseeded `random.random()` acceptance ran at every temperature. Reading `temp` directly is the correct fix, and temp=0 now stays deterministic and matches AR greedy. Merging for the next release.

Issue #1097, opened 2026-05-06
oQ-MTP performance on M1 not as expected


s-n-t:
I've tested Qwen3.6-27B-oQ5-fp16-mtp and Qwen3.6-27B-oQ8-fp16-mtp generated from the original weights, acceptance seems to be high:
```
2026-05-06 20:12:45,019 - omlx.patches.mlx_lm_mtp.batch_generator - INFO - MTP[1] finish=length tokens=128 cycles=68 accept=58/68 (85.3%) emits[init=2,draft=58,bonus=58,verify=10] timing[backbone=1305.3ms mtp=324.7ms sample=6296.6ms cache=17.4ms]
2026-05-06 20:12:45,056 - omlx.scheduler - INFO - Cache phase timings: cleanup_finished_sync=0.1ms/2, store_cache_main_eval=8.0ms/2, store_cache_main_prep=8.1ms/2
2026-05-06 20:13:12,767 - omlx.patches.mlx_lm_mtp.batch_generator - INFO - MTP path activated for uid=2 (model has mtp_forward, batch=1)
2026-05-06 20:13:21,280 - omlx.patches.mlx_lm_mtp.batch_generator - INFO - MTP[2] finish=length tokens=128 cycles=68 accept=58/68 (85.3%) emits[init=2,draft=58,bonus=58,verify=10] timing[backbone=1323.1ms mtp=325.9ms sample=6410.3ms cache=18.2ms]
```
But oQ5 runs slower with MTP enabled and while oQ8 is improved it's not as much as I think is expected?
```
MTP ON

oMLX - LLM inference, optimized for your Mac
https://github.com/jundot/omlx
Benchmark Model: Qwen3.6-27B-oQ5-fp16-mtp
================================================================================

Single Request Results
--------------------------------------------------------------------------------
Test                TTFT(ms)    TPOT(ms)        pp TPS        tg TPS      E2E(s)    Throughput    Peak Mem
pp1024/tg128          7308.2       65.63   140.1 tok/s    15.4 tok/s      15.643    73.6 tok/s    19.80 GB
pp4096/tg128         28089.0       65.23   145.8 tok/s    15.5 tok/s      36.373   116.1 tok/s    21.22 GB
pp8192/tg128         56390.1       64.54   145.3 tok/s    15.6 tok/s      64.586   128.8 tok/s    22.25 GB
[...]

PR #1337, opened 2026-05-21
fix(mtp): correct backbone and sampler time measurement

deepsweet:
Related #1330, #1311, #1097.

A lot of people including me got confused with huge MTP sample timings. The actual problem is that due to mx.core lazy nature we've leaked massive logits computation to the sampler part until `mx.eval(verify_tok, bonus_tok)` or whatever first `.tolist()` happens.

This PR materializes it earlier so that it counts as `backbone_ms` properly:

```
INFO - MTP[0] finish=stop tokens=1886 cycles=973 accept=911/973 (93.6%) emits[init=2,draft=911,bonus=911,verify=62] timing[backbone=108907.4ms mtp=3935.5ms sample=401.0ms cache=54.1ms]
```

### DIFF
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -752,6 +752,9 @@ def _run_verify_cycle(gen_batch: Any, state: _MtpState) -> None:
         )
         verify_logits = logits[:, 0, :]
         bonus_logits = logits[:, 1, :]
+    # materialize logits now so that it doesn't count as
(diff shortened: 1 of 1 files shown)


jundot:
Thanks, this matches what @sangemaru traced in #1311. The single `mx.eval` barrier was materializing the whole verify forward into `sample_ms`, which is what confused everyone across #1097/#1311/#1330. Forcing `mx.eval(logits)` after the backbone attributes the cost correctly with negligible overhead. Merging for the next release. I'll refresh the now-stale split comment at 743-744 in a follow-up.

PR #1338, opened 2026-05-21
fix(tests): classify dflash_ssd_cache_max_bytes

fry69:
Fix small oversight to make this test green again:

```shell
pytest tests/test_admin_profiles_api.py
```


jundot:
Thanks for catching this. The field landed in ModelSettings but never got added to the profile allowlist, so the SSD budget was being dropped on profile save and the classification test went red. Placement in MODEL_SPECIFIC_PROFILE_FIELDS is correct (matches its in-memory sibling). Merging.

PR #1339, opened 2026-05-21
fix(hf): follow cross-origin permanent redirects on configured endpoint

popfido:
## Summary

`huggingface_hub` doesn't follow cross-origin 301/308 redirects during its HEAD probes, so a downloader configured to use a mirror that redirects (e.g. `hf-mirror.com` → `huggingface.co` from non-CN IPs) silently fails with a misleading error.

This change resolves the configured endpoint upfront — walks up to 3 permanent hops, returns the final origin — and pins `HfApi` to it so downloads target a stable host.

This is the first of the prep PRs split out of the Swift macOS app rewrite (`feature/native-app-swift-rewrite`); files as a standalone fix because it benefits HTML dashboard users too.

## Scope

- Only the admin UI's `huggingface.endpoint` setting is run through the new probe. The `HF_ENDPOINT` env var resolution stays inside `huggingface_hub` (we pass `endpoint=None` when no admin setting is configured), matching the existing `test_snapshot_download_endpoint_none_without_mirror` contract.
- Best-effort probe: network errors leave the endpoint unchanged. Per-process cache so the resolve only fires once per endpoint.

## Test plan

- [x] 9 new tests in `TestResolveEndpoint` cover: no-redirect, 308, 301, relative Location, 3-hop chain, 302/307 ignored, network error fallback, cache hit, trailing-slash normalization.
- [x] Verified RED by reverting `hf_downloader.py` to `origin/main` — all 9 fail with `ImportError`.
- [x] Full `tests/test_hf_downloader.py` — 99 passed.

## Local reproduction

```bash
git checkout fix/hf-mirror-cross-origin-redirect
uv run pytest tests/test_hf_downloader.py::TestResolveEndpoint -v
# Expect: 9 passed
```

To prove the tests catch the bug, temporarily revert the fix:

```bash
git show origin/main:omlx/admin/hf_downloader.py > omlx/admin/hf_downloader.py
uv run pytest tests/test_hf_downloader.py::TestResolveEndpoint -v
# Expect: 9 errors (ImportError on _resolve_endpoint)
git checkout HEAD -- omlx/admin/hf_downloader.py
```

### DIFF
--- a/omlx/admin/hf_downloader.py
+++ b/omlx/admin/hf_downloader.py
@@ -14,6 +14,7 @@ import uuid
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Callable, Optional
+from urllib.parse import urlparse
 
 from huggingface_hub import HfApi, hf_hub_download, snapshot_download
 from huggingface_hub.utils import (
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for splitting this out as a standalone fix.

I reproduced the premise from a non-CN IP: `hf-mirror.com/api/models/gpt2` returns a cross-origin 308 to `huggingface.co`, while `huggingface.co` itself only does a same-origin 307, so the probe resolves the mirror correctly and leaves a direct endpoint untouched.

Logic holds against the real responses and the test coverage looks solid. Merging for the next release.

PR #1340, opened 2026-05-21
refactor(bench): replay-on-subscribe SSE delivery + multi-tab coherence

popfido:
## Summary

This is also a prepare commit for the Swift Desktop App of oMLX.

Replaces the single-consumer `asyncio.Queue` event delivery on `BenchmarkRun` / `AccuracyBenchmarkRun` with an append-only event log + `asyncio.Condition`. SSE endpoints now snapshot the log, release, yield, repeat — every subscriber sees the run's full event history from the beginning and multiple subscribers can attach concurrently. Then layers the client-side surface needed to actually expose the new capability end-to-end.

Second of the prep PRs split out of the Swift macOS app rewrite per maintainer ask. The Swift app's REST polling mirrors (`upload_state`, `phase`) on `BenchmarkRun` go away in a follow-up PR after this one lands.

## What it fixes (latent HTML dashboard bugs)

- Two browser tabs on `/api/bench/{id}/stream` (or accuracy equivalent) no longer starve each other; both see every event.
- Reloading the page mid-run no longer loses the events delivered before the reload — the replay redelivers them.
- A second tab opened mid-run now actually attaches to the active bench (was silently never subscribing before today).
- Two tabs racing `Start` on the throughput bench no longer succeed independently and produce mutually-corrupted measurements; the second is rejected with 409.

## Commit-by-commit

| commit | What |
|---|---|
| `refactor(bench): replay-on-subscribe SSE delivery (multi-consumer)` | `BenchmarkRun.queue` / `AccuracyBenchmarkRun.queue` → `events: list[dict]` + `cond: Condition` + `terminal: bool`. `_send_event` becomes append-then-notify-all. SSE endpoints become replay-then-attach. Migrates 6 existing test callers from `run.queue.get_nowait()` to `list(run.events)`. |
| `test(bench): pin replay + multi-consumer + terminal-close contract` | New `tests/test_benchmark_sse_replay.py` with 5 SSE-replay cases + 3 accuracy-side cases. Verified RED on the pre-refactor commit. |
| `feat(bench): tab discovery so a second tab attaches to an active run` | `get_active_run()` helper + new `GET /api/bench/active` endpoint + dashboard JS `loadBenchState()` wired into bench-tab activation. Without this the replay capability has no client wired to use it. 4 new tests for `get_active_run`. |
| `feat(bench): one-bench-at-a-time guard + cross-tab awareness banner` | Server rejects concurrent throughput starts with 409 (engine contention would corrupt both measurements). Client gates the Start button on the same signal. Adds an amber banner on tab A when a different bench is running in another tab, with View live / Dismiss actions; `visibilitychange` listener so the banner appears when the user returns to the browser tab. 5 new route-level tests. |

## Wire format unchanged

Same JSON events as before. Since replay redelivers events on every reconnect (including page refresh), 4 append-only result handlers in `dashboard.js` were taught to dedupe before pushing:

- `result` events: keyed by `(pp, tg)` for single tests, `batch_size` for batch.
- `upload` events: keyed by `context_length`.
- Accuracy `result`: keyed by `(model_id, benchmark)`.

The other handlers (`progress`, `done`, `upload_done`, `upload_skipped`, `error`) overwrite state by reference and were already idempotent (audited).

## Test plan

```bash
uv run pytest tests/test_benchmark.py tests/test_accuracy_benchmark.py tests/test_benchmark_sse_replay.py -q
```

Expected: **79 passed**.

Breakdown of new + migrated tests in `tests/test_benchmark_sse_replay.py` (16 total):

| Class | Tests | What it covers |
|---|---|---|
| `TestBenchmarkSSEReplay` | 5 | Late-subscriber replay, multi-consumer parity, terminal-close timing, error-as-terminal, replay→live boundary |
| `TestGetActiveRun` | 4 | Discovery returns None / completed-only / picks running / excludes cancelled+error |
| `TestAccuracyBenchmarkSSEReplay` | 3 | Same replay / multi-consumer contract on accuracy side + `last_progress` preserved |
| `TestActiveBenchEndpoint` | 2 | `GET /api/bench/active` happy paths via FastAPI TestClient |
| `TestConcurrentStartRejection` | 2 | `POST /api/bench/start` returns 409 while a run is active; allows a fresh start after the previous completed |

Plus 6 migrated callers in `tests/test_benchmark.py` and 1 in `tests/test_accuracy_benchmark.py` switched from `run.queue.get_nowait()` to `list(run.events)`.

## Local repro

**Tests** (proves the contract):

```bash
# GREEN — full suite
uv run pytest tests/test_benchmark_sse_replay.py -v
# Expect: 16 passed

# RED proof — temporarily revert the dataclass changes
git show origin/main:omlx/admin/benchmark.py > omlx/admin/benchmark.py
git show origin/main:omlx/admin/accuracy_benchmark.py > omlx/admin/accuracy_benchmark.py
uv run pytest tests/test_benchmark_sse_replay.py::TestBenchmarkSSEReplay -v
# Expect: 4 of 5 fail with "AttributeError: 'BenchmarkRun' object has no attribute 'cond'"
git checkout HEAD -- omlx/admin/benchmark.py omlx/admin/accuracy_benchmark.py
```

**End-to-end UX** (hard-refresh the dashboard so the new JS loads):

| Scenario | Expected |
|---|---|
| Open `/admin/#/bench` in two tabs. Start a bench in tab A. | Both tabs render live progress for the same run. |
| Refresh tab A mid-run. | Table re-populates from replay; live updates continue. |
| Tab A running. Click Start in tab B. | Tab B shows error toast referencing the existing bench_id; no second run starts. |
| Tab A's bench finishes, result card on screen. Start a new bench in tab B. | Amber banner appears in tab A: "Another throughput benchmark is running on \<model\>" with `View live` / `Dismiss`. |
| Click `View live`. | Tab A's old rows clear, new bench's events stream in via replay. |
| Switch to another browser app and back to tab A. | If a new bench started in the meantime, the banner appears without needing an in-app tab switch. |

## Memory bound

`run.events` grows over a run's lifetime. The existing `cleanup_old_runs(max_runs=10)` already bounds total completed runs in `_benchmark_runs`. Worst case ≈ 10 runs × ~1000 events × ~500 bytes = ~5 MB. Acceptable for a desktop app.

### DIFF
--- a/omlx/admin/accuracy_benchmark.py
+++ b/omlx/admin/accuracy_benchmark.py
@@ -72,16 +72,28 @@ class AccuracyBenchmarkRequest(BaseModel):
 
 @dataclass
 class AccuracyBenchmarkRun:
-    """Tracks the state of a running accuracy benchmark."""
+    """Tracks the state of a running accuracy benchmark.
+
+    SSE delivery model mirrors `BenchmarkRun`: append-only `events`
+    log + `cond` for live notification + `terminal` flag set on the
+    final event. See benchmark.py for the rationale.
+    """
 
     bench_id: str
     request: AccuracyBenchmarkRequest
     status: str = "running"  # running, completed, cancelled, error
-    queue: asyncio.Queue = field(default_factory=asyncio.Queue)
+    events: list[dict] = field(default_factory=list)
+    cond: asyncio.Condition = field(default_factory=asyncio.Condition)
+    terminal: bool = False
     task: Optional[asyncio.Task] = None
     results: list[dict] = field(default_factory=list)
     error_message: str = ""
     last_progress: Optional[dict] = None  # last progress event for reconnect
+
+
+# Accuracy stream closes on `done` (run finished) or `error`. Unlike the
+# throughput bench there's no separate upload phase to ride out.
+_ACCURACY_TERMINAL_TYPES = frozenset({"done", "error"})
 
 
 # --- Run management ---
@@ -251,13 +263,18 @@ async def cancel_queue() -> None:
 
 
 async def _send_event(run: AccuracyBenchmarkRun, event: dict) -> None:
-    """Send an SSE event to the client."""
+    """Append an event to the run's log and wake subscribers.
+
+    Updates `last_progress` (used by the REST `queue/status` endpoint
+    for reconnect hints) and sets `run.terminal` on the final event.
+    """
     if event.get("type") == "progress":
         run.last_progress = event
-    try:
-        await run.queue.put(event)
-    except Exception:
-        pass
+    async with run.cond:
+        run.events.append(event)
+        if event.get("type") in _ACCURACY_TERMINAL_TYPES:
+            run.terminal = True
+        run.cond.notify_all()
 
 
 # --- Benchmark execution ---
--- a/omlx/admin/benchmark.py
+++ b/omlx/admin/benchmark.py
@@ -103,6 +103,20 @@ def get_run(bench_id: str) -> Optional[BenchmarkRun]:
     return _benchmark_runs.get(bench_id)
 
 
+def get_active_run() -> Optional[BenchmarkRun]:
+    """Return the currently-running throughput benchmark, if any.
+
+    Discovery surface for clients that need to attach to an in-progress
+    run without knowing the bench_id upfront (page refresh, second tab).
+    Returns the first run with status == "running"; throughput benches
+    are 1-at-a-time so there's never more than one.
+    """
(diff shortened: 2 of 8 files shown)


jundot:
Thanks for splitting this out cleanly. The replay-on-subscribe model plus the multi-tab guard fixes four latent UX issues at once, and the RED proof plus 16 new cases make the contract easy to verify. The 5 MB worst-case memory budget on `run.events` is exactly the reasoning I'd want to see.

Merging now.

PR #1341, opened 2026-05-21
feat(chat): overhaul ui and usability

beamivalice:
Improve chat UI and feature, the continuation of multitask chat. This may cover 90% of the use case for most of the general users.

* Rightside Bar for both model settings and profile management - collapsible
* Model settings so you don't have to switch to dashboard, this already pull the setting from there - UPDATE: remove some setting that requires model to be reload and persist in the server. (speculative decoding and kvquant)
* Profiles support with localStorage and profile tag in message response
* Perf monitor at the bottom right in case you want to test speed in real time
* Timeline for jumping around the chat - separate icons for chatbot and user
* Context menu to branch, delete or copy markdown
* Improve look and feel while retaining omlx identity
* Multitasking between chat variant tabs
* Implementing Id across chat system
* Auto-name setting on the left UPDATE: removed feature
* Fix - correctly display markdown in user bubble

Here's screenshots. I'm playing with this for a few days now. You may want to try!

Screenshot 2569-05-28 at 10 59 35

image

Screenshot 2569-05-21 at 18 27 07

Screenshot 2569-05-21 at 18 33 08

Screenshot 2569-05-21 at 18 36 00


jundot:
Thanks for the continuation of #1231, the profile panel and the rail are really nice.

Quick question before I merge: the `finally` block of `streamResponse` calls `this.regenerateChatTitle()` (gated by `autoName`, default on), but I can't find the function defined in the diff or anywhere else in the file. Is it coming in a separate PR, or did it get left out by accident? With autoName default-on right now it throws on the first response of every new chat, so `resetStreamSession` doesn't run and the input stays stuck in stop-mode.

Happy to merge as soon as we sort that out. Thanks again!

beamivalice:
I'll look into it. I remembered it was my last minute adjustment to take a force autoname regen button out.

beamivalice:
@jundot done. I forgot to remove autoname since it's not worth sending another llm call for just name stuff. Or we might need a small enough model to do quick things.

beamivalice:
More fixes.

* Stable message keys — Every message gets a real id, and Alpine uses that instead of the array index, so edit/copy/regenerate don’t attach to the wrong bubble after splices.
* Regenerate / delete slice — Regenerate and delete on an assistant message cut back to the last user message, not just “everything before this index,” so hidden tool-call rows don’t stick around and poison the next reply.
* Streaming thinking tags — Fixed the slice offset for incomplete `` blocks so you don’t get a mangled prefix while thinking is still streaming in.
* Perf timeline on the right message — Status/timeline data goes on the assistant message that was actually streaming (targetMessageId), not whatever assistant happens to be last in the array.
* Model aliases — Aliases map to the real gateway id for VLM checks and prefill polling, so image upload and “Starting…” status work when you’re using a friendly model name.
* No double streams across chats — Switching chats aborts streams on other chats so you don’t get two generations fighting each other.
* API payload builder — buildMessagesForApi only sends the active variant chain per user turn (when you have variants), not every tab’s text mixed together.
* System prompt sync — Streaming uses the chat session’s prompt and keeps the header UI in line when you’re on that chat.
* localStorage quota — Saving history trims old chats if storage blows up instead of failing silently.
* Message IDs everywhere — New user/assistant/tool rows all get ids, and old history gets ids backfilled on load.

beamivalice:
**buildMessagesForApi** (main's conflict - resolved this way)

Sends reasoning_content, with fallback to _thinking on assistant rows (thinking models on follow-up turns).
System message comes from resolveApiSystemPrompt() — uses the active variant’s saved meta.systemPrompt on the latest user turn, not only the header/session string.

**Hidden tool rows**

Wrapper uses x-show="isMessageVisible(msg, index)" so _ui: false tool/assistant rows don’t show up in the UI.
Timeline rail uses data-msg-index on visible bubbles so dots/scroll line up with real message indices.

**Variants**

Tabs under the user bubble when a turn has multiple assistant variants.
selectVariant(userIndex, variantId) sets _activeVariantId on that user row (not always the last user).
Only the active variant is shown per turn (isMessageVisible).
captureGenerationContext() stores meta (model, profile, system prompt) on each new assistant message.

**Regenerate**

No more sliceToUserTurn — old variants stay; regenerate adds a new assistant and selects it.
Still uses current header model + system prompt (for “new run with current settings”); keeps origMsg._profile.

**Other**

Assistant header label uses variantModelLabel(msg) from saved meta.
Legacy assistants get empty meta backfill on load via ensureVariantMeta() (won’t overwrite existing meta).

beamivalice:
Fix - chat model setting sent PUT to server to override model setting.

Add variant responses so user can try with multiple chatbot and compare the response with just a click and fixed multiple issues related to variants - variant stats/delete etc.

Screenshot 2569-05-28 at 09 08 20

@jundot I'm done thank you! So many issues that need to iron out. I ran sweep tests and fix correctness issues I've found.

beamivalice:
Fix more server bug. Had to change server.py to accept chat setting too. Top_K and Repetition Penalty was not there.

omlx/admin/templates/chat.html
```diff
@@ -2025,34 +4140,33 @@ <h3 class="text-lg font-bold" style="color: var(--text-primary);">
...
-                if (msg.tool_calls) m.tool_calls = msg.tool_calls;
-                if (msg.tool_call_id) m.tool_call_id = msg.tool_call_id;
-                return m;
-            });
-        // Prepend system prompt if set
-        if (context.systemPrompt.trim()) {
-            messagesForApi.unshift({
-                role: 'system',
-                content: context.systemPrompt.trim()
-            });
```
fry69, line 2058: This section needs to survive it seems to be able to get merged cleanly into main? (guessworking without clue here)

beamivalice: Message building already lives in buildChatCompletionBody() → buildMessagesForApi() (reasoning, tools, system prompt, variants).

beamivalice:
I polished user experience and spent 6 hrs to unit test pretty much everything I can think of. Green from my side.

jundot:
Thank you for sticking with this for almost a week. The continuation of #1231 turned into something much bigger than I expected, and the variants, branching, timeline, and inline perf monitor combo is honestly the chat experience oMLX has been missing.

I'm running on limited capacity these days, so having a contributor who can self-verify at this depth means everything. I really don't take it for granted.

Merging now. Big respect for the time and care you put in. Thank you again.

PR #1344, opened 2026-05-21
fix(dflash): detect multimodal content and trigger VLM fallback

### LINKED ISSUE #1342: DFlash engine drops image content instead of falling back to VLM (v0.3.9rc1)
**Describe the bug**
When DFlash is enabled for a VLM-capable model (Qwen3.6-27B-8bit with VLM engine), image content in chat completion requests is silently dropped. The DFlash engine processes only the text portion of the request and generates a hallucinated response unrelated to the uploaded image. No error or warning appears in the server log, the image is simply ignored.

**To Reproduce**
1. Set up Qwen3.6-27B-8bit as VLM type in oMLX admin
2. Enable DFlash with a draft model (e.g. Qwen3.6-27B-DFlash)
3. Send a chat completion request containing an image (e.g. via Open WebUI with vision enabled, or via `/v1/chat/completions` with `image_url` content type)
4. Observe that the model responds with text unrelated to the image, it hallucinates an answer based on the text prompt only
5. Server log shows normal `Chat completion` entries with small prompt token counts (~289), confirming the image was not processed
6. Disable DFlash in admin panel → send the same image → model correctly describes/analyzes the image

**Expected behavior**
When DFlash is enabled and a request contains image content, the engine should detect the multimodal input and automatically fall back to the VLM engine for that request (similar to how it falls back when context exceeds `max_ctx`). The image should be processed and the response should reflect the image content.

**Screenshots**
Server log with DFlash ON (image not processed — prompt: 289 tokens, no VLM activity):
```
omlx.engine.dflash - INFO - DFlash generation complete: 931 tokens, 31.3 tok/s, acceptance=84.2%, cycles=147
omlx.server - INFO - Chat completion: 930 tokens in 29.80s (32.9 tok/s), prompt: 289
```

Same request with DFlash OFF works correctly, image is analyzed by VLM engine.

- macOS Version: Sequoia 15.7.7
- oMLX Version: v0.3.9rc1
- Model: mlx-community/Qwen3.6-27B-8bit (VLM mode)
- DFlash draft: Qwen3.6-27B-DFlash

**Additional context**
- DFlash config: max_ctx=8192, fallback=vlm, l1_cache=True, l2_cache=True, draft_window=1024, verify=adaptive
- The fallback to VLM currently only triggers on context length exceeding max_ctx, not on multimodal content detection
- Workaround: disable DFlash in admin panel when using image/VLM features

ivaniguarans:
## Summary

DFlash engines silently drop images when VLM fallback is configured. Requests with image content get their images stripped during text-only template flattening, producing hallucinated responses instead of triggering the VLM fallback path.

Closes #1342

## Root Cause

Two-part bug:

1. **Server-side** (`server.py`): The extraction path checks `isinstance(engine, VLMBatchedEngine)` to decide whether to preserve images. DFlash wraps a VLM fallback but isn't itself a VLM instance, so images are stripped before the engine ever sees them.

2. **Engine-side** (`dflash.py`): Even if images survived extraction, DFlash's `chat()` and `stream_chat()` have no multimodal detection — they'd feed image tokens into the text-only draft model, producing garbage.

## Changes

**`omlx/engine/dflash.py`:**
- Add `supports_multimodal_fallback` property — returns `True` when fallback engine type is `"vlm"`
- Add `_has_multimodal_content()` static method — scans messages for `image`, `image_url`, `input_image` content parts
- Add multimodal detection in `chat()` and `stream_chat()` — when image content is detected and VLM fallback is configured, trigger eviction + fallback before processing
- Add `asyncio.Lock` around all fallback transitions (multimodal + existing max_ctx) to prevent concurrent eviction races

**`omlx/server.py`:**
- Add `is_dflash_vlm` detection via `getattr(engine, "supports_multimodal_fallback", False)` in both OpenAI and Anthropic extraction paths
- Preserve images through extraction when DFlash has VLM fallback configured

**`tests/test_dflash_multimodal_fallback.py`:**
- 22 tests covering property detection, content scanning, chat/stream fallback triggering, lock safety under concurrency, and server extraction routing

> **Note:** There is a pre-existing concurrency gap where text-only requests arriving during the eviction transition window could fail. This affects the existing max_ctx fallback path equally — not specific to this change. Worth a separate ticket if addressed.

## Test plan

- [x] `pytest tests/test_dflash_multimodal_fallback.py` — 22/22 pass
- [x] `pytest tests/ -x --timeout=60 -q` — full suite passes (pre-existing failures on main excluded)
- [x] Verified: image requests to DFlash+VLM trigger fallback instead of silent stripping
- [x] Verified: text-only requests still use normal DFlash path (no regression)
- [x] Verified: concurrent image requests produce single eviction (lock safety)

### DIFF
--- a/omlx/engine/dflash.py
+++ b/omlx/engine/dflash.py
@@ -114,6 +114,7 @@ class DFlashEngine(BaseEngine):
         self._model_type_str = None
         self._fallback_engine: BaseEngine | None = None
         self._in_fallback_mode = False
+        self._fallback_lock = asyncio.Lock()
         self._runtime_context: Any | None = None
         self._dflash_prefix_cache: Any | None = None
         # Protocol-specific output parser factory (gemma4 / harmony).
@@ -497,6 +498,22 @@ class DFlashEngine(BaseEngine):
         )
         return len(self._tokenizer_obj.encode(prompt))
 
+    @property
+    def supports_multimodal_fallback(self) -> bool:
+        return self._fallback_engine_type == "vlm"
+
+    _MULTIMODAL_TYPES = frozenset({"image", "image_url", "input_image"})
+
+    @staticmethod
+    def _has_multimodal_content(messages: list[dict]) -> bool:
+        for msg in messages:
+            content = msg.get("content")
+            if isinstance(content, list):
+                for part in content:
+                    if isinstance(part, dict) and part.get("type") in DFlashEngine._MULTIMODAL_TYPES:
+                        return True
+        return False
+
     def _should_fallback(self, prompt_tokens: list[int]) -> bool:
         if self._max_dflash_ctx is None:
             return False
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for the careful fix.

The two-part root cause matches what I see: extract_text_content silently drops images when DFlash is in front (no message_extractor, isinstance VLM check fails), and chat()/stream_chat() apply the text-only template before any fallback check, so even an already-in-fallback engine loses the image.

The supports_multimodal_fallback duck-typing keeps server.py from importing DFlash, and folding the existing max_ctx fallback under the same _fallback_lock is a nice consistency win.

Merging for the next release.

PR #1349, opened 2026-05-22
Feat/dashboard idle ttl

### LINKED ISSUE #1307: Feature Request: Add idle time or time to eject to dashboard
I'd like to know how long a model has been loaded, and more importantly how long it's been idle so I can determine how much time is left before it is ejected. All three times might be even better.

I would like a small indicator on the dashboard that indicates how long the model has been loaded for and how much time is left before it is automatically ejected (if that feature is enabled).

imi4u36d:
feat(admin): show idle time and TTL countdown in Active Models dashboard
Add idle duration and TTL countdown display next to each loaded model's
status in the Admin Dashboard. Users can now see at a glance how long a
model has been idle and when it will be automatically evicted.
Backend (omlx/admin/routes.py):
- In _build_active_models_data(), compute idle_seconds from the model's
  last_access timestamp.
- Compute ttl_remaining_seconds using a two-tier priority:
  1. Per-model ttl_seconds (from model settings)
  2. Global idle_timeout_seconds (fallback)
- Both fields are None when the model is still loading.
Frontend (omlx/admin/templates/dashboard/_status.html):
- Show idle duration (e.g. "5m") next to the existing "idle" dot.
- Show TTL countdown (e.g. "~2m left") when a TTL is configured.
- Model is evicted on expiry, so no "EXPIRED" state is needed.
Tests (tests/test_active_models_visibility.py):
- test_idle_seconds_computed_from_last_access
- test_idle_seconds_none_when_no_last_access
- test_ttl_remaining_from_per_model_setting
- test_ttl_remaining_falls_back_to_global_idle_timeout
- test_ttl_remaining_per_model_takes_priority
- test_ttl_remaining_clamped_to_zero
- test_idle_and_ttl_not_computed_for_loading_model
Closes #1307

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -3908,6 +3908,32 @@ def _build_active_models_data() -> dict:
                         0.0, loading_estimated_seconds - loading_elapsed_seconds
                     )
 
+        # Compute idle time and TTL remaining for loaded models.
+        is_loaded = model_info.get("loaded") and entry is not None and entry.engine is not None
+        last_access = model_info.get("last_access")
+        idle_seconds: float | None = None
+        ttl_remaining_seconds: float | None = None
+
+        if is_loaded and last_access is not None and last_access > 0:
+            idle_seconds = max(0.0, time.time() - last_access)
+
+        # Determine effective TTL: per-model ttl_seconds first, then global idle_timeout.
+        effective_ttl: int | None = None
+        settings_manager = _get_settings_manager()
+        if is_loaded and settings_manager is not None:
+            model_settings = settings_manager.get_settings(model_id)
+            if model_settings is not None and getattr(model_settings, "ttl_seconds", None) is not None:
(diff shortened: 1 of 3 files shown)


imi4u36d:
Before this, I didn’t dare to enable this feature. I had no way of knowing when a model would be suddenly unloaded, dropping memory usage and causing macOS to allocate those free resources to other new tasks. Then, when I went back to use the model, it would have to be reloaded, leading to sudden memory swapping, severe lag spikes, or even OOM (Out of Memory) crashes.
Now, it works beautifully like this:
image

jundot:
Thanks for picking this up. Issue link, screenshot, and tests all very clean.

Verified the dashboard's idle/TTL math matches engine_pool.check_ttl_expirations exactly (same last_access source, same per-model then global fallback), so the countdown won't lie to the user. Tests cover the right edges.

Will merge as-is. Thanks!

PR #1351, opened 2026-05-22
fix(ms-downloader): enrich recommended models with size + params

popfido:
## Summary

ModelScope's `list_models` endpoint returns `Path`/`Name`/`Downloads`/`Likes` but rarely populates `StorageSize`, and never returns a parameter count. As a result, the admin Downloads tab's "Recommended Models" cards were empty in the **Model Size** and **Parameters** columns when the user switched to the ModelScope source — the equivalent HuggingFace cards show both.

This PR enriches each recommendation entry post-`list_models` with two new endpoints, behind a bounded-concurrency `asyncio.gather` with a 24-hour in-process cache:

- **`_fetch_model_config(model_id)`** → `GET /api/v1/models/{id}/repo?FilePath=config.json` (~1 KB JSON). Feeds the parameter-count estimator.
- **`_fetch_model_detail_size(model_id)`** → `GET /api/v1/models/{id}`. Used only as a fallback when `list_models` reported `StorageSize=0`. Prefers `ModelInfos.safetensor.model_size` (weights only) over the repository `StorageSize` (weights + tokenizer + readme), matching the HF semantics.
- **`_estimate_params_from_config(config)`** → closed-form estimate from `vocab_size` / `hidden_size` / `num_hidden_layers` / `intermediate_size` / GQA fields. Handles dense and MoE variants and tied vs. untied LM heads.

The memory filter is re-applied after enrichment, so models whose size only becomes known post-enrichment can be dropped rather than offered to users with insufficient RAM.

**Performance**: Concurrent fetches (semaphore=8) → a full 50-entry recommendations call finishes in ~1–2s on the cold path. Cached in-process for 24h (config.json content for a published model is effectively immutable), so subsequent page loads are free. Cache bounded at 1024 entries with oldest-first eviction.

**Wire-compat**: No DTO change — the existing `size` / `size_formatted` / `params` / `params_formatted` fields are now consistently populated for MS-source models. Swift / HTML clients require no changes.

## Test plan

All 179 tests in `tests/test_{hf,ms}_downloader.py` pass locally.

New test classes in `tests/test_ms_downloader.py` (30 cases):

- [ ] `TestEstimateParamsFromConfig` — 8 cases including ballpark checks against real **Llama-3.2-1B** and **Qwen2.5-7B** configs (both within ±10% of published param counts), MoE multiplier, tied vs. untied LM head, and defensive coercion of string fields.
- [ ] `TestEnrichCache` — TTL expiry + eviction-on-read.
- [ ] `TestFetchModelConfig` — URL shape, 200 / 404 / non-JSON / network-error / empty-model-id paths.
- [ ] `TestFetchModelDetailSize` — prefers `safetensor.model_size`, falls back to `StorageSize`, returns 0 on missing data / non-200.
- [ ] `TestEnrichMsEntry` — fill-params, fill-size, skip-detail-when-size-known, cache reuse, empty-repo no-op.
- [ ] `TestRecommendedEnrichmentEndToEnd` — full `get_recommended_models()` flow verifying size + params land on the response, and that the memory filter still excludes oversized models post-enrichment.

Existing recommended/search tests were updated to mock the new helpers so the suite stays offline.

## Manual verification

To reproduce the original bug on `main`:

1. `omlx server --admin-port 8080`
2. Open the admin Downloads tab, switch source to ModelScope.
3. Recommended Models cards show blank Size and Parameters columns.

After this PR:

1. Same steps as above.
2. Cards now show "X.Y GB" sizes and "N.NB" / "NNNM" parameter counts, matching the HF source's display.

### DIFF
--- a/omlx/admin/ms_downloader.py
+++ b/omlx/admin/ms_downloader.py
@@ -20,6 +20,7 @@ from .hf_downloader import (
     DownloadStatus,
     DownloadTask,
     _format_model_size,
+    _format_param_count,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for tracking this down — the ModelScope size enrichment fills a real gap (cards were rendering blank because `list_models` returns `StorageSize=0` in most cases). estimator math against Llama-3.2-1B / Qwen2.5-7B looks right, and the memory filter re-apply after enrichment is a nice touch.

One heads-up on the params half: the current admin template only renders `model.size_formatted` for the MS trending/popular/search cards and the MS detail modal — there is no `params_formatted` span on the MS side (only HF has it at `_models.html:493/496/629/632/1860`). So the params enrichment lands in the API response but never reaches the UI today.

Plan on my end is to merge this as-is for the size win, then push a follow-up that adds the matching `params_formatted` span to the MS card / modal so the params data you are now collecting actually shows up. Will land in the next release.

PR #1355, opened 2026-05-22
chore(packaging): pin venvstacks as dev dep + auto-detect driver

popfido:
## Summary

This PR is raised for the reason someone wants to build App locally.

`packaging/build.py` currently invokes venvstacks via `pipx run venvstacks …` in three places. That works, but the version of venvstacks that produces the macOS app bundle is whatever PyPI release happens to be current on a given contributor's machine the day they build — there's no lockfile discipline on the tool that produces the artifact. Two of us can build "the same" commit and get subtly different output if venvstacks shipped a release in between.

This PR pins venvstacks the same way we pin every other dep that affects the bundle:

- **`pyproject.toml`**: `venvstacks>=0.7.0` added to both `[project.optional-dependencies].dev` (pip-managed venvs) and `[dependency-groups].dev` (PEP 735 / `uv sync --dev`), matching the existing pair-maintained pattern for pytest/black/ruff.
- **`packaging/build.py`**: a new `_venvstacks_driver()` helper auto-detects an available driver in priority order:
  1. `venvstacks` directly on PATH (installed via the dev extra — fastest path, no tool-runner startup overhead)
  2. `uvx venvstacks` (uv's `pipx run` equivalent)
  3. `pipx run venvstacks` (historical default — backwards compatible)
  
  Exits with an actionable remediation hint when none are present.
- **`README.md`**: updated the macOS build prerequisites paragraph to point at the dev extra as the canonical install path. Translated READMEs intentionally not touched — kept for the next translation-sync pass.

## Why this matters

The `.app` bundle ships to end users is the contract; venvstacks is what produces it. Making the producer reproducible is a small but load-bearing step toward the broader distribution story.

## Compatibility

- **pip-only workflows**: unchanged. `pip install -e ".[dev]"` brings venvstacks into the active venv, and the driver-detection helper finds it via the direct-on-PATH branch.
- **pipx-only workflows (status quo)**: still work via the third fallback — no behavior change for anyone who hasn't switched.
- **uv-only workflows**: `uv sync --dev` installs venvstacks; same direct-on-PATH branch picks it up. `uvx venvstacks` is also available as a fallback for cases where someone runs `build.py` outside an active venv.
- **End-user `.app`**: byte-identical. This is purely a build-time-reproducibility + host-dependency-footprint change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,7 +338,7 @@ pytest -m "not slow"
 
 ### macOS App
 
-Requires Python 3.11+ and [venvstacks](https://venvstacks.lmstudio.ai) (`pip install venvstacks`).
+Requires Python 3.11+ and [venvstacks](https://venvstacks.lmstudio.ai). venvstacks is declared as a dev dependency, so `pip install -e ".[dev]"` (or `uv sync --dev`) brings the pinned version into your environment. `build.py` also falls back to `uvx venvstacks` or `pipx run venvstacks` if you prefer a host-global tool runner.
 
 ```bash
 cd packaging
--- a/packaging/build.py
+++ b/packaging/build.py
@@ -490,11 +490,45 @@ def _check_git_commit_sync():
         sys.exit(1)
 
 
+def _venvstacks_driver() -> list[str]:
+    """Pick an available venvstacks driver as a command prefix.
+
+    Resolution order (first that works wins):
+      1. `venvstacks` directly on PATH — installed via the dev extra in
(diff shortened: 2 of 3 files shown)


jundot:
Thanks. Pinning the producer alongside everything else it ships removes a real source of "works on my machine" drift, and the three-tier driver fallback keeps the existing pipx workflow working without any contributor churn.

Merging now.

PR #1356, opened 2026-05-22
Fix Anthropic stream block indices for thinking tool calls

### LINKED ISSUE #1353: Bug: Streaming SSE emits non-contiguous `content_block_start` indices, causing Anthropic SDK `IndexError`
### Description

When the model generates a response with `thinking` + `tool_use` blocks (no intermediate `text` block), the server emits `content_block_start` events with non-contiguous indices — jumping from `index=0` to `index=2`, skipping `index=1`. This violates the Anthropic Messages API streaming protocol and causes the official Anthropic Python SDK to crash with `IndexError: list index out of range`.

### Root Cause

The Anthropic streaming spec requires `content_block_start` indices to be **consecutive integers starting from 0**. The SDK's `accumulate_event()` appends each new block to a list on `content_block_start`, then accesses it by `event.index` on subsequent `content_block_delta` / `content_block_stop` events. When the index jumps (0 → 2), the list has only 2 elements (at positions 0 and 1), so `content[2]` raises `IndexError`.

### Observed SSE Event Sequence

```
event: content_block_start   → index=0, type=thinking     ✓
event: content_block_delta   → index=0, thinking_delta     ✓  (×130 deltas)
event: content_block_stop    → index=0                     ✓
event: content_block_start   → index=2, type=tool_use      ✗  SHOULD BE index=1
event: content_block_delta   → index=2, input_json_delta   ✗  IndexError here
event: content_block_stop    → index=2                     ✗
event: message_delta         → stop_reason=tool_use
```

**Expected:** When the model produces only 2 content blocks (thinking + tool_use, no text), the indices should be 0 and 1, not 0 and 2.

It appears the server reserves index=1 for a `text` content block but never emits it when the model decides to call a tool directly after thinking.

### Reproduction

**Model:** `Qwen3.5-122B-A10B-8bit`
**SDK:** `anthropic` Python SDK v0.88.0
**Trigger condition:** Prompt that forces the model to call a tool (produces thinking + tool_use, no text block)

Minimal reproduction script (requires `anthropic` and `httpx`):

```python
import asyncio
import anthropic

SYSTEM = (
    "You are a data analysis assistant. "
    "You MUST call the query_database tool for ANY data question. "
    "NEVER answer data questions directly — always use the tool first."
)
TOOL = {
    "name": "query_database",
    "description": "Execute a SQL query and return results",
    "input_schema": {
        "type": "object",
        "properties": {
            "sql": {"type": "string", "description": "SQL statement"},
            "description": {"type": "string", "description": "Query description"},
        },
        "required": ["sql", "description"],
    },
}

async def main():
    client = anthropic.AsyncAnthropic(
        base_url="http://<YOUR_SERVER>:<PORT>",
        api_key="<YOUR_KEY>",
    )
    async with client.messages.stream(
        model="Qwen3.5-122B-A10B-8bit",
        max_tokens=4096,
        system=SYSTEM,
        messages=[{"role": "user", "content": (
            "Query the database: SELECT user_id, MAX(login_time) as last_login "
            "FROM user_logins WHERE login_time >= DATE_SUB(NOW(), INTERVAL 1 MONTH) "
            "GROUP BY user_id"
        )}],
        tools=[TOOL],
    ) as stream:
        async for event in stream:
            if event.type == "content_block_start":
                print(f"content_block_start index={event.index} type={event.content_block.type}")
            # Crashes here with IndexError when index jumps from 0 to 2

asyncio.run(main())
```

### Raw SSE Dump

Full raw SSE byte stream is available in the attached `raw_sse_dump.txt`. Key excerpt showing the index gap:

```
event: content_block_stop
data: {"type":"content_block_stop","index":0}

event: content_block_start
data: {"type":"content_block_start","index":2,"content_block":{"type":"tool_use",...}}
```

### Expected Behavior

Content block indices should be contiguous: if only 2 blocks are emitted (thinking at 0, tool_use at 1), indices should be `[0, 1]`, not `[0, 2]`.

### SDK Stack Trace

```
  File "anthropic/lib/streaming/_messages.py", line 278, in __stream__
    self.__final_message_snapshot = accumulate_event(
  File "anthropic/lib/streaming/_messages.py", line 465, in accumulate_event
    content = current_snapshot.content[event.index]
IndexError: list index out of range
```

### Environment

- **oMLX Version:** 0.3.9
- **macOS Version:** 26.3

[raw_sse_dump.txt](https://github.com/user-attachments/files/28134075/raw_sse_dump.txt)

- **Model:** Qwen3.5-122B-A10B-8bit
- **anthropic SDK:** 0.88.0
- **Python:** 3.11

lvsijian8:
Fixes #1353.

## Problem
Anthropic streaming clients expect `content_block_start` indices to be contiguous and zero-based. When oMLX streamed a response with a thinking block followed directly by a tool_use block, the emitted indices skipped from `0` to `2`. The official Anthropic SDK accumulates content blocks in a list and then indexes by `event.index`, so the gap can raise `IndexError: list index out of range`.

## Root cause
`stream_anthropic_messages()` closed the thinking block and incremented `block_index`. Later, the tool_use start index was calculated as `block_index + 1`, which added one more step and produced the gap.

## Fix
- keep `block_index` on the just-closed thinking block
- let the existing tool block start calculation advance to the next contiguous index
- add a regression assertion for thinking + tool_use streaming so `content_block_start` indices remain consecutive

## Tests
- `.venv/bin/pytest tests/integration/test_e2e_streaming.py::TestStreamingHelperFunctions::test_stream_anthropic_messages_sanitizes_tool_call_markup_inside_thinking -q`
- `.venv/bin/pytest tests/integration/test_e2e_streaming.py::TestStreamingHelperFunctions -q`
- `.venv/bin/pytest tests/test_anthropic_adapter.py tests/test_api_utils.py::TestSSEEventFormatters -q`
- `.venv/bin/pytest tests/integration/test_e2e_streaming.py -q`

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -3313,9 +3313,9 @@ async def stream_anthropic_messages(
 
     # 4. Close open blocks
     if thinking_block_started and not text_block_started:
-        # Only thinking was emitted, close it
+        # Only thinking was emitted. Keep block_index on the just-closed
+        # block so following tool_use blocks start at the next contiguous index.
         yield create_content_block_stop_event(index=block_index)
-        block_index += 1
     if text_block_started:
         yield create_content_block_stop_event(index=block_index)
     elif not thinking_block_started and not tool_calls:
--- a/tests/integration/test_e2e_streaming.py
+++ b/tests/integration/test_e2e_streaming.py
@@ -919,6 +919,7 @@ class TestStreamingHelperFunctions:
 
         thinking_deltas = []
         tool_use_blocks = []
+        block_start_indices = []
         stop_reasons = []
         for event in parsed_events:
             if event.get("type") == "content_block_delta":
@@ -926,6 +927,7 @@ class TestStreamingHelperFunctions:
                 if delta.get("type") == "thinking_delta":
                     thinking_deltas.append(delta["thinking"])
             elif event.get("type") == "content_block_start":
+                block_start_indices.append(event["index"])
                 content_block = event.get("content_block", {})
                 if content_block.get("type") == "tool_use":
                     tool_use_blocks.append(content_block)
@@ -939,6 +941,7 @@ class TestStreamingHelperFunctions:
         assert "</tool_call>" not in streamed_thinking
         assert "Need to inspect first." in streamed_thinking
         assert "Then continue." in streamed_thinking
+        assert block_start_indices == list(range(len(block_start_indices)))
         assert len(tool_use_blocks) == 1
         assert tool_use_blocks[0]["name"] == "get_weather"
         assert "tool_use" in stop_reasons


jundot:
Thanks for the fix and the clean diagnosis. Confirmed the root cause matches what you wrote: the close-thinking branch bumped `block_index` and then `tool_block_start = block_index + 1` bumped it again, jumping `0 → 2`.

I walked through every `(thinking_started, text_started)` combination — all paths land on contiguous indices after the fix, and `block_index` is not consumed after the thinking-only close in the no-tool case, so dropping the increment is side-effect-free.

The contiguity assertion added to the existing sanitize test covers the regression scenario directly. Will ship in the next release.

PR #1364, opened 2026-05-22
fix(oq): skip .scales/.biases keys in oQ6 streaming dequant

FaisalFehad:
## Summary
Fix oQ6 streaming dequantization to handle uint32 source weight format. The original code assumed float16 weights; oQ6 stores weights as uint32 with separate .scales/.biases keys that must be skipped during streaming dequant.

## Changes
- `omlx/oq.py`: Build `_skip_keys` set for .scales/.biases, pop them when dequantizing uint32 .weight inline

## Test plan
- `TestOq6StreamingSkipScalesBiases` (2 tests in `test_oq.py`): verifies scales/biases skipped in main loop, removed from output index
- Uses `_mock_sensitivity` fixture from existing test patterns

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -2413,6 +2413,13 @@ def quantize_oq_streaming(
 
     tensor_names = list(all_weights.keys())
     total_tensors = len(tensor_names)
+    # Build a set of .scales/.biases keys to skip in the main loop.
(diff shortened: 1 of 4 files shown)


FaisalFehad:
Closing per request

PR #1366, opened 2026-05-22
perf(scheduler): skip boundary snapshot detection when paged SSD cache disabled

FaisalFehad:
## Summary
Short-circuit _detect_boundary_snapshot_need() to avoid model.make_cache() calls when paged_ssd_cache_dir is not configured. This saves unnecessary cache construction on the scheduler hot path.

## Changes
- `omlx/scheduler.py`: `_detect_boundary_snapshot_need()` returns default set early when `paged_ssd_cache_dir` is not set

## Test plan
- `TestBoundarySnapshotSkip` (3 tests in `test_scheduler.py`): skips detection when SSD disabled, runs detection when SSD enabled, preserves explicit true when SSD disabled


FaisalFehad:
Closing per request

PR #1389, opened 2026-05-24
fix: guard late aborts after engine close

glasses666:
## Summary
- Guard late `EngineCore.abort_request()` calls after `EngineCore.close()` has cleared scheduler state.
- Guard `AsyncEngineCore` stop/abort/bulk-abort paths when the async wrapper is already closed.
- Add regression tests for late abort, context-manager exit, and bulk abort after close.

## Why
A streaming response can race with model unload: the stream generator reaches its cleanup `finally` after unload has already stopped/closed the engine. In that path a late abort should be a no-op, not an `AttributeError` that surfaces as a server streaming error.

## Test plan
- [x] RED: new close-race tests failed before the fix with `AttributeError`.
- [x] Focused regression tests: `3 passed` for the new close-race cases.
- [x] Related engine tests: `194 passed` for `test_engine_core.py`, `test_batched_engine.py`, `test_vlm_engine.py`, `test_engine_pool.py`.
- [x] Full default pytest: `4589 passed, 36 skipped, 59 deselected`.
- [x] `git diff --check` clean.
- [x] Live patched-source port 8000 smoke: strong stream + forced unload saw a real `gemma-4-e4b-it-4bit` content chunk, unload returned HTTP 200, abort signature delta stayed `0`, final `loaded_count=0`.

## Live verification artifact
Local run artifact: `/Users/dracoglasser/.hermes/runs/omlx-stream-unload-fix-20260524/stream-forced-unload-strong-final.json`

### DIFF
--- a/omlx/engine_core.py
+++ b/omlx/engine_core.py
@@ -372,7 +372,15 @@ class EngineCore:
         finally block, NOT here -- calling _cleanup_request() immediately
         after put() would clear the output before the consumer can drain it.
         """
-        result = self.scheduler.abort_request(request_id)
+        scheduler = getattr(self, "scheduler", None)
+        if getattr(self, "_closed", False) or scheduler is None:
+            logger.debug(
+                "Skipping abort for request %s because engine is already closed",
+                request_id,
+            )
+            return False
+
+        result = scheduler.abort_request(request_id)
 
         # Signal consumer with abort error so any waiting
         # stream_outputs() / generate() can exit gracefully.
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, the EngineCore guard on `abort_request` is the right fix. I traced the race: `BatchedEngine.stop()` calls `EngineCore.close()` which nulls out `self.scheduler`, and a stream's `finally` reaches `engine.abort_request()` on the same EngineCore reference after that. That's exactly the AttributeError you describe, and the early-return guard handles it at the root.

Live smoke artifact looks good. Happy to merge.

Issue #906, opened 2026-04-23
[Feature Request] CLI flag to disable the auto-selected tool-call parser (pass raw model tokens through to content)


trevorgordon981:
### Summary
Add `--tool-call-parser {auto,none,<name>}` (and a matching per-model setting) so operators can bypass oMLX's auto-extraction of tool calls into the structured OpenAI `tool_calls` field. When downstream code already owns tool-call parsing (proxies, custom routers, Hermes-style pipelines), the current behavior silently pre-processes the model's raw tokens and breaks that downstream.

### Version
oMLX 0.3.7rc2 (Homebrew, `jundot/omlx` tap) on Mac Studio M3 Ultra (512 GB), macOS 25.2.0
Model: `mlx-community/Qwen3.5-397B-A17B-6bit`, loaded via VLMBatchedEngine (`VLM tool calling enabled: parser=qwen3_coder` in startup log)

### Current behavior
With `stream: true` and `tools` in the request, oMLX emits:
- `delta.content` → almost empty (just `"\n\n"` in my capture)
- `delta.reasoning_content` → the reasoning text
- `delta.tool_calls` → fully structured call
- `finish_reason: "tool_calls"`

My downstream (blockops-proxy) parses `<tool_call>…</tool_call>` XML directly from the model's raw `content` (the native Qwen format that `mlx-vlm` and earlier tool setups produced). With oMLX the content is null, so the proxy sees a text-only response and the tool never fires for the end user.

### Desired behavior
`omlx serve --tool-call-parser none` (or `settings.json: scheduler.tool_call_parser: "none"`) should:
- Stream the raw decoded model tokens into `delta.content`
- Leave `delta.tool_calls` unset and `finish_reason: "stop"`
- Not split `reasoning_content` from `content` (or make that independently configurable via `--reasoning-parser none`)

Keep `auto` as the default so existing users are unaffected.

### Minimal repro
```bash
curl -sS -m 120 -N -X POST http://127.0.0.1:8082/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "qwen3.5-397b-6bit",
    "messages": [{"role":"user","content":"Run uname -a"}],
    "tools": [{"type":"function","function":{"name":"terminal","parameters":{"type":"object","properties":{"command":{"type":"string"}},"required":["command"]}}}],
    "tool_choice": "auto",
    "max_tokens": 200,
    "temperature": 0,
    "stream": true
  }'
```

[...]

Issue #812, opened 2026-04-16
[VLMBatchedEngine] Tool calling silently stops — model outputs text without tool_call on Qwen3.6-35B (works fine with BatchedEngine)


Keenni:
## Environment

- **oMLX**: 0.3.5.dev1 (homebrew, build 260416194559-macos26-tahoe)
- **macOS**: 25.4.0 (Darwin arm64, MacBook Pro M5 Max, 128 GB)
- **Model**: `Qwen3.6-35B-A3B-mlx-mxfp8` (MoE, 35B total params / 10B active, mxfp8 quantized, 8 shards)
- **Issue Model**: Qwen3.5-35B-A3B-Heretic-MLX-mxfp8 (same VLMBatchedEngine symptom)

## Problem Description

When the model is loaded via **VLMBatchedEngine** (auto-detected due to model directory containing `processor_config.json` / `video_preprocessor_config.json`), tool calling breaks silently:

1. Model receives a user query requiring a tool
2. Model generates a textual response like "让我去查一下最新的 CME 缺口数据" but **does not emit any `tool_call` in the response**
3. OpenClaw embedded agent receives the text response, sees no tool_call, and treats it as a normal stop (stopReason: "stop")
4. The conversation ends without the tool actually being called

When the **same model** is forced to load via **BatchedEngine** (LLM mode), tool calling works correctly:
- Model properly emits `tool_calls` in the response
- OpenClaw executes the tools
- Model continues generating after receiving tool results

## Engine Switching Evidence (from server.log)

```
# Auto-detected as VLM (broken tool calling):
01:20:44 - omlx.engine.vlm - VLM tool calling enabled: parser=qwen3_coder
01:20:44 - omlx.engine.vlm - VLMBatchedEngine loaded: /Users/slayer/.omlx/models/Qwen3.6-35B-A3B-mlx-mxfp8

# Forced LLM mode (tool calling works):
01:05:55 - omlx.engine.batched - BatchedEngine loaded: /Users/slayer/.omlx/models/Qwen3.6-35B-A3B-mlx-mxfp8
```

## Root Cause Hypothesis

Qwen3.6 has `processor_config.json` / `video_preprocessor_config.json` in its model directory, causing oMLX to auto-detect it as a VLM and load via VLMBatchedEngine.

VLMBatchedEngine's `vlm_stream_generate` path handles tool calling differently from BatchedEngine's `stream_generate` path. Specifically:
[...]

PR #1393, opened 2026-05-24
fix: replace Guard 1 heuristic with name-matching for thinking tool calls

### LINKED ISSUE #1392: [Bug] Guard 1 in extract_tool_calls_with_thinking drops valid tool calls when model emits preamble after thinking
## What's happening

Qwen3-Coder with thinking enabled puts actual tool invocations inside `<thinkgt;` blocks, then emits a short preamble like "Let me create the file:" as regular content. Guard 1 in `extract_tool_calls_with_thinking` drops the tool calls because it sees non-empty regular text and assumes the thinking-embedded calls are "just reasoning."

## Repro

Model: `mlx-community/Qwen3-Coder-30B-A3B-Instruct-4bit`, thinking on.

```bash
curl -s --max-time 60 http://localhost:8000/v1/chat/completions \
  -d '{
    "model": "Qwen3-Coder-30B-A3B-Instruct-4bit",
    "messages": [
      {"role": "user", "content": "Use the write_file tool to write hello to /tmp/test.txt"}
    ],
    "tools": [{"type":"function","function":{"name":"write_file","description":"Write to a file","parameters":{"type":"object","properties":{"path":{"type":"string"},"content":{"type":"string"}},"required":["path","content"]}}}],
    "max_tokens": 200
  }'
```

**Without thinking**: tool calls work fine. Model outputs `<function=write_file>` in regular content, omlx parses it, response has `tool_calls`.

**With thinking**: model generates `<function=write_file>` inside thinking, then "Let me create the file:" as regular text. Guard 1 kills the tool call because `regular_content.strip()` is truthy. Response has no tool calls.

## The code

`omlx/api/tool_calling.py`, lines 624–626:

```python
# Guard 1: if model produced regular text, the tool call in thinking
# is just reasoning, not an actual invocation request.
if tool_calls and regular_content.strip():
    tool_calls = None
    tool_calls_from_thinking = False
```

This was added in #484 to stop small models hallucinating tool syntax in reasoning. Fair enough, but it's too blunt — it can't distinguish between "model is reasoning about tools" and "model genuinely invoked a tool in thinking then added a preamble."

## Fix

Guard 2 (name matching against provided tools) is the right discriminator. If the tool call in thinking matches a known tool name, it's a real invocation regardless of whether the model also emitted regular text. If it doesn't match, it gets dropped.

Two options:

1. **Swap guard order**: run name validation first, then only apply the regular-text check if names didn't validate. Still not great — a matched name should win regardless.
2. **Drop Guard 1 entirely**: rely on Guard 2's name matching alone. If the call matches a provided tool, promote it. If not, drop it. The regular-text check adds nothing that name matching doesn't already handle better.

Option 2 is cleaner. The regular-text heuristic is fundamentally wrong for models that put real invocations in thinking.

## Related

- #484 — introduced the guards
- #812 — same symptom (tool calls silently null) on Qwen3.6-35B, attributed to VLMBatchedEngine but may share this root cause
- #906 — request to disable tool parser entirely (workaround for this class of issue)

Mearman:
Fixes #1392

## What changed

Replaced the blanket "regular text means just reasoning" guard in `extract_tool_calls_with_thinking()` with three-case logic:

| Case | Tools list | Regular text | Result |
|------|-----------|-------------|--------|
| 1 | No | Yes | Drop — unvalidated, could be reasoning |
| 2 | No | No | Keep — clear tool invocation |
| 3 | Yes | Any | Name matching decides |

Case 3 is the fix for #1392. Qwen3-Coder (and similar models) place real tool invocations inside thinking blocks and emit a short preamble as regular content. The old Guard 1 dropped these because it saw non-empty regular text. Now, when a tools list is provided, only name matching determines validity — if the call matches a known tool, it's promoted regardless of accompanying prose.

## Why not just swap guard order?

The old Guard 1 (regular text check) ran before Guard 2 (name matching). Swapping them doesn't help because Guard 1 would still kill valid calls after name matching passed. The regular-text heuristic is fundamentally the wrong discriminator when tools are provided — name matching already tells you whether the call is real.

## Test

New test: `TestThinkingFallbackGuardRegression::test_known_tool_in_thinking_kept_with_preamble` — verifies that a tool call matching a provided tool name survives even when regular content (a preamble) is present.

All 154 tests in `test_tool_calling.py` pass, including the existing guard tests which still validate the no-tools-list + regular-text case.

## Related

- #484 — introduced the original guards
- #812 — same symptom on Qwen3.6-35B
- #906 — feature request for `--tool-call-parser none` (workaround)

### DIFF
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -621,19 +621,30 @@ def extract_tool_calls_with_thinking(
         _, tool_calls = parse_tool_calls(thinking_content, tokenizer, tools)
         tool_calls_from_thinking = bool(tool_calls)
 
-        # Guard 1: if model produced regular text, the tool call in thinking
-        # is just reasoning, not an actual invocation request.
-        if tool_calls and regular_content.strip():
-            tool_calls = None
-            tool_calls_from_thinking = False
-
-        # Guard 2: only keep tool calls whose name matches a provided tool.
-        if tool_calls and tools:
-            valid_names = _extract_tool_names(tools)
-            tool_calls = [tc for tc in tool_calls if tc.function.name in valid_names]
-            if not tool_calls:
-                tool_calls = None
-                tool_calls_from_thinking = False
+        # Guard: validate thinking-embedded tool calls.
+        #
+        # Three cases:
+        # 1. No tools list AND regular text exists → drop.  The call is
+        #    unvalidated and could be hallucinated reasoning about tools.
+        # 2. No tools list AND no regular text → keep.  The model clearly
+        #    intended a tool invocation (no competing prose).
+        # 3. Tools list provided → name matching is the sole discriminator.
+        #    If the call name matches a known tool, promote it regardless of
+        #    whether regular text was also produced.  The previous "regular
+        #    text means just reasoning" heuristic was wrong for models
+        #    (Qwen3-Coder) that genuinely place tool calls in thinking.
+        # See https://github.com/jundot/omlx/issues/1392
+        if tool_calls:
+            if not tools:
+                if regular_content.strip():
+                    tool_calls = None
+                    tool_calls_from_thinking = False
+            else:
+                valid_names = _extract_tool_names(tools)
+                tool_calls = [tc for tc in tool_calls if tc.function.name in valid_names]
+                if not tool_calls:
+                    tool_calls = None
+                    tool_calls_from_thinking = False
 
     return ToolCallExtraction(
         cleaned_text=cleaned_text,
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the thorough write-up and the case table, that mapping makes the intent very clear in the code.

I walked through the four cases (tools-yes/no x regular-text-yes/no), and the 3-case branch resolves them all correctly. Name matching as the sole discriminator when tools are provided is the right call.

Merging. Fix lands in the next release.

Mearman:
gosh I didn't expect a merge that quick. thanks @jundot

PR #1397, opened 2026-05-24
Make prefill step size configurable to reduce peak mem significantly

mikael-johansson:
Reduce peak memory without sacrificing quality or speed.

**M5 Pro 48GB: Without this patch**
```
================================================================================
Benchmark Model: Qwen3.6-27B-oQ6-mtp
================================================================================

Single Request Results
--------------------------------------------------------------------------------
Test                TTFT(ms)    TPOT(ms)        pp TPS        tg TPS      E2E(s)    Throughput    Peak Mem
pp1024/tg128          2593.9       50.72   394.8 tok/s    19.9 tok/s       9.035   127.5 tok/s    22.96 GB
pp4096/tg128          9277.3       52.60   441.5 tok/s    19.2 tok/s      15.958   264.7 tok/s    24.39 GB
pp8192/tg128         18718.9       53.89   437.6 tok/s    18.7 tok/s      25.562   325.5 tok/s    25.42 GB
pp16384/tg128        38663.9       55.71   423.8 tok/s    18.1 tok/s      45.739   361.0 tok/s    26.92 GB
pp32768/tg128        83818.9       60.64   390.9 tok/s    16.6 tok/s      91.520   359.4 tok/s    29.92 GB
pp65536/tg128       202143.3       71.51   324.2 tok/s    14.1 tok/s     211.225   310.9 tok/s    35.95 GB

pp131072/tg128                                                                    N/A (OOM / Kernel panic)
```

**M5 Pro 48GB: With this patch, step size 512 (Sweetspot)**
```
==============================================================================
Benchmark Model: Qwen3.6-27B-oQ6-mtp
================================================================================

Single Request Results
--------------------------------------------------------------------------------
Test                TTFT(ms)    TPOT(ms)        pp TPS        tg TPS      E2E(s)    Throughput    Peak Mem
pp1024/tg128          2581.5       51.98   396.7 tok/s    19.4 tok/s       9.183   125.4 tok/s    22.36 GB (-0.6 GB)
pp4096/tg128          9423.4       52.88   434.7 tok/s    19.1 tok/s      16.139   261.7 tok/s    22.58 GB (-1.81 GB)
pp8192/tg128         18744.3       55.22   437.0 tok/s    18.3 tok/s      25.757   323.0 tok/s    23.33 GB (-2.09 GB)
pp16384/tg128        38917.6       56.92   421.0 tok/s    17.7 tok/s      46.146   357.8 tok/s    24.27 GB (-2.65 GB)
pp32768/tg128        84812.6       59.30   386.4 tok/s    17.0 tok/s      92.344   356.2 tok/s    26.17 GB (-3.75 GB)
pp65536/tg128       202321.6       70.37   323.9 tok/s    14.3 tok/s     211.258   310.8 tok/s    30.00 GB (-5.95 GB)

pp131072/tg128      539864.7       86.29   242.8 tok/s    11.7 tok/s     550.824   238.2 tok/s    37.74 GB (-11 GB?)
```


jundot:
Hi @mikael-johansson, thanks for the detailed benchmarks and the careful patch. You diagnosed the peak-memory pattern correctly.

I just merged a related change in acd0533 (adaptive prefill throttle) that takes the same idea (smaller prefill chunks cut peak memory) but applies it automatically: a per-scheduler EWMA tracks chunk transient bytes, and when memory crosses the caution watermark the next chunk is shrunk so it fits the remaining headroom (down to a min, then abort). On long-context prefill it keeps peak under the configured hard cap without users having to pick a step size.

Given the algorithm now covers the long-context OOM case end-to-end, I'm going to close this PR. If you find a workload where a manual override would still help on top of the throttle, please open an issue with a repro and I'll take another look. Thanks again for the work.

mikael-johansson:
@jundot The latest main doesn't fix the problem for me, the benchmark crashes with OOM on 128k ctx.
Crash log:

[omlx_main_128k_20260526.log](https://github.com/user-attachments/files/28251204/omlx_main_128k_20260526.log)

```
2026-05-26 08:29:53,141 - omlx.engine_pool - DEBUG - Skipping victim 'Qwen3.6-27B-oQ6-mtp': has active requests
2026-05-26 08:29:53,141 - omlx.process_memory_enforcer - WARNING - Hard memory pressure but all loaded models are pinned and no loads in progress.
2026-05-26 08:29:54,144 - omlx.process_memory_enforcer - INFO - Memory pressure level: hard -> ok (current=34.4GB, soft=35.9GB, hard=40.1GB)
libc++abi: terminating due to uncaught exception of type std::runtime_error: [METAL] Command buffer execution failed: Insufficient Memory (00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory)
```

jundot:
Hi @mikael-johansson, thanks again for the careful diagnosis and for pushing back on the close.

I understand the intent of your PR and the effect of lowering step_size. The piece I keep coming back to is that omlx users span a wide range. Many do not have the context to pick a safe step_size per model, and exposing this as a tunable they have to reason about probably hurts the average experience (lower step_size unconditionally trades throughput on every prompt, not just the ones at risk).

So I want to push as much of this as possible into the automatic path. The HEAD commit (4cfbc8b) replaces the predict-based throttle with a watermark-tier one: once current memory crosses the soft watermark, the next chunk is bucketed into 1024 / 512 / 256 / 128 based on how far into the soft-to-ceiling band we are. It needs no per-model tuning and only kicks in under pressure, so short prompts are unaffected.

I do not have a 48 GB machine to repro your benchmark precisely, and small-RAM cases are my current priority. From your earlier test you clearly have a setup that can run HEAD. Would you mind running pp131072 against current main and checking whether you still hit the kernel panic / Metal command-buffer error? Just the pass/fail outcome is fine, no need to run the full sweep.

If the automatic path still falls short on your machine, that is good signal to revisit as a follow-up.

mikael-johansson:
@jundot: No success on 4cfbc8b unfortunately, OOM @ 80k tokens:
```
2026-05-26 11:55:36,593 - omlx.scheduler - DEBUG - Captured prefill boundary cache snapshot for 827f4984-3c3c-473b-8536-fb4d62d57e74 at 79872 tokens
libc++abi: terminating due to uncaught exception of type std::runtime_error: [METAL] Command buffer execution failed: Insufficient Memory (00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory
```

I totally understand the wish to automate as much as possible. 
However, since the lower pp step size has no impact on processing speed on a machine like mine, would you be open to having a combined solution, e.g. replace my "Prefill step size" proposed config option with a "Maximum prefill step size", so the automatic step size _starts_ at this level and then drops even further when memory is tight? The advantage is that this would allow an explicit option to reduce peak memory so that the computer has free memory to do other stuff and can avoid swapping out everything else.

mikael-johansson:
@jundot : Correction - Got it to work on **4cfbc8b** now!
I had custom memory limits (42GB total & 35 GB model) to try and squeeze out as much as possible. I was also using `iogpu.wired_limit_mb=42000`. This worked on my branch but caused OOM and even kernel panic on your commit.

After resetting all the memory to defaults, your patch seems to work. I will keep testing to see if it's stable. Thanks for the rapid fix!

jundot:
Hey @mikael-johansson,

I just pushed a HEAD commit that drops the confusing `max_process_memory` / `max_model_memory` sliders entirely and replaces them with a memory guard tier (safe / balanced / aggressive).

The root cause behind the recurring OOM and kernel panic reports has been that oMLX only capped itself against total system memory, ignoring what other apps on the machine were actually using. That's the part I really wanted to put to rest this time. The new guard checks live available memory every second, so the ceiling moves with the rest of the system instead of staying frozen. The adaptive throttle from `4cfbc8b` sits on top, so chunked prefill steps down 2048 → 1024 → 512 → 256 → 128 as memory rises, then stops cleanly at the ceiling instead of panicking.

Tested on a 48 GB sim with Qwen3.6-27B-oQ6-mtp + 131072 prefill: throttle walked through all tiers, hit the ceiling at ~96k tokens, then recovered to 31 GB immediately with no panic. The ceiling there came from the simulated 48 GB cap rather than the model itself, so it would be great to see how the same 131072 prefill behaves on your real machine.

From your write-up it sounds like you want oMLX to push memory as hard as possible, which maps to the new `aggressive` tier. If you have a moment, would you mind pulling HEAD and trying it on your workload? The tier dropdown is in the admin dashboard next to the Memory guard toggle. Curious whether the live ceiling behaves the way you'd expect.

mikael-johansson:
Sorry, I spoke too soon regarding the 4cfbc8b commit. It just OOM:ed again, with all memory sliders at default. It seems it goes down to 512 steps appropriately, but then jumps back up to 1024 again, causing OOM:
```
2026-05-26 16:50:24,019 - omlx.scheduler - DEBUG - [throttle:external] measure rid=cf925de8-008c-4213-b093-b496a19f4a13 n=512 transient=103.84MB per_token=207.7KB ewma=233.0KB samples=211
2026-05-26 16:50:24,636 - omlx.engine_pool - DEBUG - Skipping victim 'Qwen3.6-27B-oQ6-mtp': has active requests
2026-05-26 16:50:25,638 - omlx.engine_pool - DEBUG - Skipping victim 'Qwen3.6-27B-oQ6-mtp': has active requests
2026-05-26 16:50:26,639 - omlx.engine_pool - DEBUG - Skipping victim 'Qwen3.6-27B-oQ6-mtp': has active requests
2026-05-26 16:50:26,726 - omlx.scheduler - DEBUG - [throttle:external] measure rid=cf925de8-008c-4213-b093-b496a19f4a13 n=512 transient=52.44MB per_token=104.9KB ewma=194.5KB samples=212
2026-05-26 16:50:26,732 - omlx.scheduler - DEBUG - Captured prefill boundary cache snapshot for cf925de8-008c-4213-b093-b496a19f4a13 at 104448 tokens
2026-05-26 16:50:26,735 - omlx.scheduler - DEBUG - [throttle:external] shrink rid=cf925de8-008c-4213-b093-b496a19f4a13 chunk 2048 -> 512 (current=36.80GB shrink_at=32.00GB ceiling=44.00GB band_ratio=0.40)
2026-05-26 16:50:27,640 - omlx.engine_pool - DEBUG - Skipping victim 'Qwen3.6-27B-oQ6-mtp': has active requests
2026-05-26 16:50:28,641 - omlx.engine_pool - DEBUG - Skipping victim 'Qwen3.6-27B-oQ6-mtp': has active requests
2026-05-26 16:50:29,501 - omlx.scheduler - DEBUG - [throttle:external] measure rid=cf925de8-008c-4213-b093-b496a19f4a13 n=512 delta=-2455584792B (skipped: <=0)
2026-05-26 16:50:29,503 - omlx.scheduler - DEBUG - [throttle:external] shrink rid=cf925de8-008c-4213-b093-b496a19f4a13 chunk 1536 -> 1024 (current=34.35GB shrink_at=32.00GB ceiling=44.00GB band_ratio=0.20)
2026-05-26 16:50:29,643 - omlx.engine_pool - DEBUG - Skipping victim 'Qwen3.6-27B-oQ6-mtp': has active requests
libc++abi: terminating due to uncaught exception of type std::runtime_error: [METAL] Command buffer execution failed: Insufficient Memory (00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory)
```

Thanks! I actually don't want to push memory "as hard as possible", this was just from previous experimentation. What I want is to run Qwen3.6-27B @ 128k context with full MTP speed :) And with static step size 512 (My patch), I can to this in a stable way at 37GB peak mem, which is comfortable on a 48GB machine. After rebasing onto a more recent commit, the 512 step size is even down to 33GB peak mem. So it should not even be close to OOM.

Will proceed to test with your latest HEAD.

PR #1371, opened 2026-05-23
feat(omlx-mac): native Swift macOS app

popfido:
## Status: Production Ready for discussion

This is a huge functionality PR. It carries the full Swift macOS app rewrite. The intent is to get your read on overall direction, the Python/Swift boundary, and the packaging-side simplifications before we talk about landing.

**Three smaller prep PRs are already open and depend on the maintainer's review:**

- #1340 — `refactor(bench): replay-on-subscribe SSE delivery + multi-tab coherence`
- #1351 — `fix(ms-downloader): enrich recommended models with size + params`
- #1355 — `chore(packaging): pin venvstacks as dev dep + auto-detect driver`

When those merge upstream, this PR's diff shrinks by the corresponding cherry-picked commits — about three commits' worth of server-side change. The branch is rebased on the current `upstream/main` (`cb7effd`).

## What's in scope

**Swift app** (`apps/omlx-mac/`, ~25k LOC, the bulk of the PR)

A SwiftUI native macOS app that replaces the Python + PyObjC menubar in `packaging/omlx_app/`. The app launches the existing oMLX FastAPI server as a subprocess against the bundled cpython-3.11 + venvstacks framework layer, and communicates with it over the admin HTTP API (the same one the HTML dashboard uses). Admin surfaces rendered natively: Server / Status / Library / Downloads / Bench / Logs / Models / Performance / Network / Integrations / Settings / Welcome, plus a status-item menubar, Sparkle auto-update wiring, and a CLI launcher.

The intent is not to deprecate the HTML admin panel. It is to give Mac users a desktop-app experience for running oMLX locally without asking them to manage Python environments or open a browser tab.

**Build pipeline** (`apps/omlx-mac/Scripts/build.sh`,
`packaging/build.py`, `packaging/venvstacks.toml`, `pyproject.toml`)

- `build.sh` produces a runnable `.app` end-to-end: xcodebuild + SPM resolve → fingerprint-based venvstacks rebuild when sources drift → donor layer copy → omlx package embed → ad-hoc sign. There is no `uv sync` overlay step; venvstacks is trusted as the single resolver for the bundle's Python environment.
[...]

PR #1399, opened 2026-05-25
refactor(profiles): three-scope template contract + drop is_builtin emission

popfido:
## Summary

Prep PR carved out of #1371 per maintainer review. Tightens the server-side
template/profile API into a shape both HTML and the upcoming Swift app can
bind to cleanly, with one user-visible contract shift (drops `is_builtin`
from `/api/profile-templates` payloads).

Reviewable as a standalone server change — no Swift code, no client
required to demonstrate correctness.

## What this changes on main

**`omlx/model_settings.py` (+13/-2)**

- Adds `UNIVERSAL_PROFILE_FIELDS` — the keys a global (cross-model) template
  is allowed to carry. Per-model profiles continue to accept the full
  sampling surface; globals are now constrained to the universal subset so
  the three scopes (preset / global / model) compose without leaking
  model-specific fields up the chain.
- `get_template(name)` returns `{}` for missing/empty templates instead of
  `None`. Callers that previously had to special-case `None` can now use
  the result directly.
- `/api/profile-templates` payloads no longer stamp `is_builtin: false` on
  every entry. The flag carried no information once the catalog became
  user-templates-only; HTML doesn't read it, and consumers can treat
  templates as not-builtin by construction.

**`tests/test_model_settings_profiles.py` (+35/-2)**

Covers the new universal-field gating, the `{}` return contract, and the
shape of `/api/profile-templates` without `is_builtin`.

## Behavior shifts to flag

- **`get_template(name)` returns `{}` not `None`** for unknown templates.
  HTML admin and any external `/api/profile-templates/{name}` consumers
  should be checked against this — internal callers were audited.
- **`/api/profile-templates` payloads drop `is_builtin`.** HTML doesn't
  reference the flag; older external clients that hard-required the key
  would break, though I'm not aware of any.

## How this preps #1371

The Swift app's profiles UI binds a three-scope model (preset chips
shipped via `omlx/admin/static/omlx_preset.json` + per-server global
template + per-model profile). The server side of that contract — what
the global template is allowed to carry, what the API surface returns —
lives here so it can be reviewed and merged independently of the ~4k-line
Swift rewrite. Once this lands, #1371 only carries the Swift client
bindings on top.

## Test plan

- [x] \`pytest tests/test_model_settings_profiles.py\` — 28/28
- [x] \`pytest tests/test_model_settings.py tests/test_admin_profiles_api.py tests/test_admin_api_key.py\` — 123/123
- [ ] Manual: HTML admin → Settings → Profile Templates renders + CRUD still works
- [ ] Manual: existing user templates persist + load on server restart


jundot:
Thanks for slicing the server-side prep out. The universal vs per-model field split lines up cleanly with the three-scope model, and dropping the informationless `is_builtin` key is the right call. Internal callers all read fine against the new `{}` default.

Merging now.

PR #1406, opened 2026-05-25
feat(cache): add ssd_write_drops counter for write-queue saturation

ivaniguarans:
## Summary

Adds a runtime counter `ssd_write_drops` to `PagedSSDCacheManager` that increments at every site where the background writer queue is saturated and a cache write is dropped or skipped. The new metric pairs with the existing read-path counters (`hits`, `loads`, `hot_cache_hits`, etc.) in `PagedSSDCacheStats`, giving operators a first write-path health signal. WARN logs at these sites already exist; this PR makes the drops aggregable so they can be graphed, alerted on, or surfaced through the runtime cache observability surface added in #1183.

Purely additive: no change to drop behaviour, no change to any cache code path, no new dependency.

## Changes

- `omlx/cache/stats.py`: new `ssd_write_drops: int = 0` field on `PagedSSDCacheStats`, grouped with the other operation counters (`saves`, `loads`, `errors`); added to `reset()` for parity with the other runtime counters.
- `omlx/cache/paged_ssd_cache.py`: `"ssd_write_drops": 0` initialised in `_stats`; incremented at all three queue-saturation drop sites in this module — the hot-cache eviction path in `_enqueue_ssd_write`, the cold-store preflight `_write_queue.full()` guard, and the cold-store late `except queue.Full` fallback. Passed through in `get_stats()` and `get_stats_for_model()`. `get_stats_dict()` already spreads `**self._stats`, so no manual update needed there.
- `tests/test_hot_cache.py`: new `TestSSDWriteDrops` class with five tests — dataclass default + `reset()`, wiring round-trip through both stats accessors, and one test per drop site. The drop-site tests use `unittest.mock.patch.object` on the queue methods for deterministic firing rather than depending on writer-thread timing.

The bulk-eviction unlink path is intentionally excluded — that fallback runs an inline `unlink()` when the unlink queue saturates, but the file already exists on disk so no cache write is dropped. Counting it would conflate cleanup-queue backpressure with actual write loss.

## Test plan

- [x] `uv run pytest tests/ -x` — full suite passes
- [x] `uv run pytest tests/test_hot_cache.py::TestSSDWriteDrops -v` — 5/5 pass
- [ ] Optional manual verification: load a small model (e.g. `Qwen3-0.6B-4bit` or `Qwen3-Coder-Next-6bit` per the project's cache micro-benchmarking convention), reduce `_MAX_PENDING_WRITES` to force queue saturation under a write-heavy workload, confirm `ssd_write_drops` appears and increments in `/admin/api/stats`.

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -652,6 +652,7 @@ class PagedSSDCacheManager(CacheManager):
             "preload_calls": 0,
             "preload_blocks_loaded": 0,
             "preload_time_ms": 0.0,
+            "ssd_write_drops": 0,
         }
 
         # --- Hot cache (in-memory raw-bytes tier) ---
@@ -804,6 +805,7 @@ class PagedSSDCacheManager(CacheManager):
             )
             return True
         except queue.Full:
+            self._stats["ssd_write_drops"] += 1
             logger.warning(
                 f"SSD write queue full, dropping evicted block "
                 f"{block_hash.hex()[:16]}"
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for this. Verified the 3 drop sites cover real write loss, and the new field shows up in admin stats without extra wiring. Merging.

PR #1423, opened 2026-05-26
fix(boundary-store): serialize cleanup_all() / cleanup_request() with writer thread

cfbraun:
## Summary

`test_cleanup_all_drains_queue` was failing ~20% of the time, in isolation and in suites. The race:

```
T0 save() puts item in queue
T1 writer queue.get() pulls item (already off the queue)
T2 cleanup_all() drains queue — finds it empty
T3 cleanup_all() rmtree(snapshot_dir) + mkdir(snapshot_dir)
T4 writer mkdir(req_dir) + temp write + rename(final)
```

Result: `req-X/N.safetensors` survives the supposed cleanup, the caller (scheduler / shutdown) believes the snapshot dir is empty, and the leftover file is the disk shadow of a request that should have been forgotten.

The fix introduces a `_writer_busy` barrier so `cleanup_all()` and `cleanup_request()` wait for the in-flight writer iteration to complete before touching the filesystem. Both methods now serialize correctly against the writer thread.

Also pins the cleanup-before-shutdown ordering of the new `shutdown(cleanup=True)` path with a regression test.

## Test plan

- [x] `pytest tests/test_boundary_snapshot_store.py` — 24 passed (previously flaked ~20%)
- [x] `test_cleanup_all_with_in_flight_writer` regression test added
- [x] `test_shutdown_cleanup_true_runs_cleanup_before_setting_flag` regression test added

### DIFF
--- a/omlx/cache/boundary_snapshot_store.py
+++ b/omlx/cache/boundary_snapshot_store.py
@@ -56,6 +56,17 @@ class BoundarySnapshotSSDStore:
         Snapshots are stored under ``base_dir/_boundary_snapshots/``.
     """
 
+    # Timeouts applied when acquiring _writer_busy from each cleanup
+    # path. cleanup_request is called from the scheduler's abort hot
+    # path (~3 sites) and must yield faster than cleanup_all, which
+    # also runs at startup / reset where blocking longer is tolerable
+    # in exchange for a stronger orphan-avoidance guarantee. The
+    # worst-case impact on the timeout fallback is identical in both
+    # paths — an orphan file in the recreated dir until the next
+    # constructor cleanup — so the only knob is per-call latency.
+    _CLEANUP_ALL_TIMEOUT_S = 5.0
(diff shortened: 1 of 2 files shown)


fry69:
FYI: I saw

```shell
pytest tests/test_boundary_snapshot_store.py
```

failing in previous `pytest` runs, but with current `main` (e6d8a3f68683719671a6d38361a93f618dc2a131) I can no longer reproduce this flakiness.

I gave up after ~20 tries of this test file succeeding -> `15 passed in 2.57s`

jundot:
Thanks, the race timeline matches what I traced through `_writer_loop` on main, and lock order is safe (cleanup_request nests pending → cancelled, the writer keeps them in separate critical sections, no reverse nesting anywhere). Both `save()` callers in scheduler already handle False as in-memory fallback, so the queue-full rollback is transparent. Merging.

Issue #1426, opened 2026-05-26
0.3.11 unsloth/Qwen3.6-35B-A3B-UD-MLX-4bit can not recognize image in hermes


shengpeng7:
# **Describe the bug**
When using the model `unsloth/Qwen3.6-35B-A3B-UD-MLX-4bit` in Hermes:
- oMLX v0.3.9 / v0.3.10: Triggers `api call failed` error when terminal/tool calls are executed
- oMLX v0.3.11: No error prompts, but the model **cannot perform image recognition** and produces hallucinations when trying to identify image content

Only oMLX v0.3.8 works completely normally with this model.

# **To Reproduce**
1. Use the model `https://huggingface.co/unsloth/Qwen3.6-35B-A3B-UD-MLX-4bit` in Hermes
2. In Hermes, send the command to **organize desktop files** (triggers terminal and tool calls)
3. For oMLX 0.3.9 / 0.3.10: See `api call failed` error
4. For oMLX 0.3.11: No error appears, but paste an image into Hermes and ask the model to recognize its content — the model produces hallucinations (incorrect/fabricated content)

# **Expected behavior**
The model should execute terminal/tool calls (organize desktop files) without API errors, and accurately recognize content from images pasted into Hermes without hallucinations.

# **Screenshots**

<img width="2482" height="1484" alt="Image" src="https://github.com/user-attachments/assets/5db09046-8c3c-45f6-bca5-519bcb16664d" />

# **Desktop (please complete the following information):**
 - macOS Version: 26.4 (25E246)
 - oMLX Version: 0.3.8 (normal), 0.3.9 / 0.3.10 (api error), 0.3.11 (image hallucinations)

# **Additional context**
- Model source: https://huggingface.co/unsloth/Qwen3.6-35B-A3B-UD-MLX-4bit
- The bug is related to terminal/tool call execution and image recognition capability
- Only version 0.3.8 has full normal functionality

jundot:
Thanks for the detailed report, this narrowed it down a lot.

The 0.3.11 image hallucination is a real regression. The unsloth checkpoint declares mtp_num_hidden_layers=1 in config but ships no mtp.* weights. Since 0.3.11 omlx unconditionally attaches MTPModule for any qwen3.6 vlm whose config declares mtp heads, strict load fails with "Missing 20 parameters: language_model.mtp.*", the engine silently falls back to llm, vision is dropped, and the model answers without an image.

[...]

PR #1435, opened 2026-05-27
fix(load): skip VLM MTP runtime when checkpoint lacks weights

azhangd:
## Summary

- detect whether a local checkpoint actually contains MTP tensors before enabling VLM MTP weight binding
- keep the mlx-vlm sanitize patch active for Qwen3.6 VLM layout handling
- avoid attaching an MTP runtime module when `mtp_num_hidden_layers` is declared but no `mtp.*` weights exist

## Problem

Some Qwen3.6 VLM MLX checkpoints retain `text_config.mtp_num_hidden_layers > 0` in `config.json`, but were exported without preserved MTP tensors.

In that state, the VLM pre-load path treats the config declaration as enough to attach an MTP module. Strict weight loading then expects `language_model.mtp.*` parameters that are not present in the checkpoint. The model can fall back to the LLM engine, which silently drops vision support and shows up as image hallucination or "cannot see image" responses for image inputs.

Fixes the v0.3.11 vision regression in #1426.

## Approach

This change separates two related but different states:

- MTP declared in config
- MTP tensors actually present in the checkpoint

For sharded checkpoints, the loader checks the safetensors index for `mtp.*` keys. For single-file safetensors checkpoints, it checks the file header. If the checkpoint has no visible MTP tensors, the loader disables MTP activation and skips the VLM runtime MTP attachment, while still allowing the sanitize patch and Qwen3.6 VLM visual remapping path to run.

Checkpoints that do contain MTP tensors still get the VLM runtime binding path, including the existing `mtp_enabled=False` load fix where persisted MTP weights need a destination during strict load.

## Verification

- `env -u OMLX_API_KEY .venv/bin/python -m pytest -q`
  - `4656 passed, 36 skipped, 59 deselected`
- `.venv/bin/python -m ruff check --select F821 omlx/utils/model_loading.py omlx/patches/mlx_lm_mtp/__init__.py omlx/patches/mlx_vlm_mtp/qwen35_moe_vlm_runtime.py omlx/patches/mlx_vlm_mtp/qwen35_vlm_runtime.py tests/test_model_loading.py`
  - passed
- `git diff --check`
  - passed
- Manual before/after test with `unsloth/Qwen3.6-35B-A3B-UD-MLX-4bit`
  - upstream `main`: VLM load falls back to LLM with missing `language_model.mtp.*` parameters; status becomes `model_type=llm`, `engine_type=batched`; image prompt is treated as text-only (`prompt_tokens=32`) and the model says it cannot see images
  - this branch: load stays on `VLMBatchedEngine`; status remains `model_type=vlm`, `engine_type=vlm`; image prompt includes vision tokens (`prompt_tokens=194`) and correctly answers `red square, blue circle, green triangle`

### DIFF
--- a/omlx/patches/mlx_lm_mtp/__init__.py
+++ b/omlx/patches/mlx_lm_mtp/__init__.py
@@ -37,6 +37,7 @@ logger = logging.getLogger(__name__)
 # right before ``mlx_lm.load()`` runs based on ``model_settings.mtp_enabled``.
 # Default False keeps newly-loaded models MTP-free unless explicitly opted in.
 _MTP_ACTIVE = False
+_MTP_WEIGHT_BINDING_ACTIVE = False
 
 
 def set_mtp_active(active: bool) -> None:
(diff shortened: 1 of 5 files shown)


azhangd:
Looks like this is now handled on main by ff7522b. Closing this PR as superseded.

Thanks for picking up the issue.

PR #1446, opened 2026-05-27
test(mcp): cover config loader, schema validator, and HTTP route handlers

cfbraun:
## Summary

Adds unit-test coverage for two MCP-subsystem modules that previously had none in the test suite:

- **`omlx.mcp.config`** (370-line test file): exercises `load_mcp_config` happy/error paths, the JSON schema validator (`validate_config`), `create_example_config` defaults, plus `MCPConfig` / `MCPServerConfig` / `MCPTransport` round-trips.
- **`omlx.api.mcp_routes`** (269-line test file): unit-tests the FastAPI route handlers (list/get/execute tools), input validation (422 on malformed payloads), and error mapping.

The target modules are unchanged from their upstream form. Tests are pure-Python and don't depend on MLX or any local-only fixtures, so they slot cleanly into `pytest tests/` on any platform.

## Test plan

- [x] `pytest tests/test_mcp_config.py tests/test_mcp_routes.py` — 53 passed

### DIFF
--- a/tests/test_mcp_config.py
+++ b/tests/test_mcp_config.py
@@ -1,0 +1,370 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for omlx/mcp/config.py — config file discovery, JSON/YAML
+loading, schema validation, and the example-config helper.
+
+MCPConfig / MCPServerConfig / MCPTransport themselves are covered in
+test_mcp_types.py; here we only exercise the loader and validator.
+"""
+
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for this. I ran all 53 against the current modules and they pass. I like that the tricky bits (the servers/mcpServers fallback, the `tool` alias from #1285, 503 vs 422 vs 200) are pinned with explanations instead of just asserted. Merging now.

PR #1447, opened 2026-05-27
test: unit-test coverage for previously-untested helper modules

cfbraun:
## Summary

Adds focused unit tests for four utility/helper modules that previously had only integration-level coverage (or none):

- **`tests/test_rerank_models.py`** (182 LOC): Pydantic request/response schema validation for `omlx.api.rerank_models`. `test_reranker_causal_lm.py` exercises the model end-to-end but doesn't pin the wire-format schemas.
- **`tests/test_models_base.py`** (156 LOC): unit tests for `mean_pooling` and `normalize_embeddings` in `omlx.models.base_model`. `test_embedding.py` runs the full embedding pipeline end-to-end; these are direct small-tensor sanity checks for the math.
- **`tests/test_mbpp_extract_code.py`** (156 LOC): exercises the `mbpp` branch of `_extract_code` in `omlx.eval.mbpp`. `test_eval.py` covers the `livecodebench` and `humaneval` branches but not the mbpp regex / fenced-code path.
- **`tests/test_optimizations.py`** (113 LOC): pins `get_optimization_status` return shape, flash-attention detection helper, and the `omlx.optimizations` package re-exports.

The target modules are unchanged from their upstream form. All four test files are pure-Python and skip if MLX isn't available where required.

## Test plan

- [x] `pytest tests/test_rerank_models.py tests/test_models_base.py tests/test_mbpp_extract_code.py tests/test_optimizations.py` — 61 passed

### DIFF
--- a/tests/test_mbpp_extract_code.py
+++ b/tests/test_mbpp_extract_code.py
@@ -1,0 +1,156 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for omlx/eval/mbpp.py's _extract_code helper — the regex/
+heuristic parser that pulls Python source out of an LLM response
+before MBPP runs it through the subprocess executor. The
+benchmark-runner logic itself is exercised by test_eval.py; here we
+just pin the code-extraction branches.
+"""
+
+from __future__ import annotations
+
+from omlx.eval.mbpp import _extract_code
+
+
+class TestFencedBlocks:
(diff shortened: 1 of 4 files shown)


jundot:
Thanks, this is great coverage. Ran all 61 locally and they pass. The hand-computed values for mean_pooling/normalize and the both-branches flash-attention check are exactly what these helpers were missing. Merging now.

PR #1448, opened 2026-05-27
fix(memory): walk VLM nested config for accurate KV+SDPA estimation

cfbraun:
## Summary

`_set_model_info_for_monitor` reads `model.config.num_hidden_layers` directly. For multimodal models (Qwen3.6-VL, Gemma-4, some Llava / HF auto-wrappers) the top-level config holds the *vision tower* dimensions; the actual language model lives under `text_config` / `language_config` / `llm_config`. Reading the wrong field under-estimates the LM's KV+SDPA peak by a constant factor (a 40-layer LM wrapped in a 33-layer vision tower undercounts by ~20%) and lets `_preflight_memory_check` approve prefills that go on to crash Metal mid-prefill.

Fix: probe `text_config` / `language_config` / `llm_config` in priority order before extracting dimensions. If any sub-config exposes `num_hidden_layers` (or the legacy `n_layer` alias for GPT-style configs), prefer it — even when the top-level config also has a value, since the top-level value is the unreliable one on these wrappers. Falls back to the top-level config when no sub-config qualifies.

## Test plan

- [x] `pytest tests/test_memory_monitor_vlm_config.py::TestSetModelInfoForMonitorVLMWalk` — 6 passed
  - text_config / language_config / llm_config each take priority over top-level vision-tower dims
  - plain LM (no sub-configs) still works
  - sub-config without a layer count is correctly skipped
  - GPT-style `n_layer` in the sub-config is recognized

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -6236,6 +6236,27 @@ class Scheduler:
                 logger.debug("Could not extract model config for memory estimation")
                 return
 
+            # VLM / multimodal configs (e.g. Qwen3.6-VL, Gemma-4) nest the
+            # language-model dimensions under a sub-config. Prefer
+            # ``text_config`` / ``language_config`` / ``llm_config`` when ANY
+            # of them exposes the LM layer count, even if the top-level config
+            # also has one — on some VLM packs (older Gemma-3, certain Llava /
+            # HF auto-wrappers) the top-level field refers to the *vision
+            # encoder*, not the LM, and accepting it silently miscalibrates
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for tracing this — VLM nested config is a real gap. I confirmed mlx-vlm Qwen3-VL/Gemma-4 configs (and the gemma-4-31B-it config.json on disk) and they all carry the LM layer count under `text_config` only. omlx already walks `text_config` in `models/vlm.py` and `models/embedding.py`; scheduler was just missing it.

Merging.

PR #1449, opened 2026-05-27
fix(memory): warn once per engine type when scheduler unresolvable

cfbraun:
## Summary

`_propagate_memory_limit` skips engines whose `_resolve_scheduler` returns None, but until now the skip was silent. That same silent no-op is what originally hid the dead-memory-guard bug for months — the docstring on `_resolve_scheduler` itself acknowledges it. If a future wrapper change (BatchedEngine refactor, new engine type) breaks the `engine._engine.engine.scheduler` traversal, the guard will silently stop propagating again.

Fix: log a WARNING when `_resolve_scheduler` returns None, naming the engine type. Rate-limited per type per enforcer lifetime (`_scheduler_resolve_warned: set[str]`) so a misconfigured engine polled every second doesn't spam, but the first occurrence is loud enough to alert CI / oncall.

## Test plan

- [x] `pytest tests/test_process_memory_enforcer.py::TestUnresolvableSchedulerWarning` — 2 passed
  - three propagation calls with an unreachable scheduler emit exactly one WARNING (not three)
  - when one engine is unresolvable but another resolves, the unresolvable one's failure doesn't prevent propagation reaching the resolvable one

### DIFF
--- a/omlx/process_memory_enforcer.py
+++ b/omlx/process_memory_enforcer.py
@@ -340,6 +340,11 @@ class ProcessMemoryEnforcer:
         # or the call failed). Used by the admin dashboard to surface a
         # warning when the kernel iogpu.wired_limit_mb is below this.
         self._metal_wired_limit_request: int = 0
+        # Engine types we've already complained about in
+        # ``_propagate_memory_limit``'s "scheduler unreachable" path.
+        # Prevents the per-poll warning from spamming logs while keeping
+        # the first occurrence loud enough to alert CI / oncall.
+        self._scheduler_resolve_warned: set[str] = set()
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for this. The motivation is real. The prefill guard genuinely sat dead for a while because the silent no-op hid the broken wrapper chain, so making that path loud is the right instinct. One thing I'll smooth over in a follow-up: `_propagate_memory_limit` iterates every pool entry, including discovered-but-not-yet-loaded models whose `engine` is still None, so the warning would fire once with engine type "None" on a normal startup before anything is loaded. I'll gate it to skip the not-loaded case so the warning only fires on a real wrapper break. Merging this and folding that guard into the next release.

Issue #1405, opened 2026-05-25
[Bug] Prefill requests stuck after force-stopped due to memory limit exceeded - never released


W-W-Go:
Description
    
    Summary
    When oMLX's memory enforcer force-stops a prefill request (because memory exceeds the hard limit), the prefill request gets stuck in a limbo state - it neither completes nor gets released. This causes PP slots to accumulate until all slots are full, eventually impacting normal chat requests.
    
    Steps to Reproduce
    1. Run oMLX with a model (e.g., Qwen3.6-35B-A3B-UD-MLX-4bit)
    2. Engage in extended conversation (generates multiple background PP requests)
    3. Memory usage exceeds the hard limit (60GB)
    4. oMLX force-stops prefill requests
    5. Observe that force-stopped PP requests remain stuck in the dashboard, never completing or being released
    6. PP slots accumulate until all slots are full (10+ PP requests stuck)
    
    Expected Behavior
    - Force-stopped PP requests should be properly cleaned up and released
    - PP slots should be freed for new requests
    - Normal chat requests should not be impacted by stuck background PP requests
    
    Actual Behavior
    - Force-stopped PP requests remain stuck indefinitely
    - PP slots accumulate (observed 10+ stuck PP requests)
    - Memory pressure oscillates between soft/hard states
    - Eventually impacts normal chat performance
    
    Environment
    - oMLX version: v0.3.10
    - OS: macOS (Apple Silicon)
    - Hardware: M5 Max 64GB unified memory
    - Model: Qwen3.6-35B-A3B-UD-MLX-4bit
    - Hot Cache: 12GB
    - Max Concurrent: 8
    
    Key Log Lines
    
    Memory pressure oscillation
    2026-05-25 19:18:48,879 - omlx.process_memory_enforcer - INFO - Memory pressure level: hard -> soft (current=51.8GB, soft=47.6GB, hard=53.2GB)
    2026-05-25 19:18:48,879 - omlx.process_memory_enforcer - INFO - Memory pressure level: soft -> hard (current=55.7GB, soft=47.6GB, hard=53.2GB)
    
    Prefill force-stopped
    2026-05-25 19:19:20,804 - omlx.scheduler - WARNING - Prefill force-stopped at 100352 tokens: memory 60.1GB exceeds hard limit 60.0GB
    2026-05-25 19:19:20,809 - omlx.scheduler - ERROR - Error in batch generation step: Memory limit exceeded during prefill
[...]

PR #1450, opened 2026-05-27
fix(scheduler): release paged cache refs on prefill rejection paths

cfbraun:
## Summary

`add_request` routes through `block_aware_cache.fetch_cache` which records the request in `_request_tables` and increments ref counts on every prefix-matched paged-cache block. The normal completion path releases that state in `_cleanup_finished`; the prefill-rejection paths in `_advance_chunked_prefills` / `_schedule_waiting` (added in #1405) pop `self.requests` and remove the prefill-tracker entry but **never call `release_cache(request_id)`** — so rejected requests leak `_request_tables` entries and pin block refs in the paged-cache pool, compounding the very memory pressure that triggered the rejection.

Fix: add a small `_release_paged_cache_for_request` helper and call it from all three rejection sites. The helper delegates to `block_aware_cache.release_cache` (which pops the request table AND deletes the paged block_table) when a prefix cache is configured; otherwise it falls back to `paged_cache_manager.delete_block_table` for the no-prefix-cache configuration.

## Test plan

- [x] `pytest tests/test_scheduler_chunked_prefill.py::TestPrefillRejectionReleasesPagedCache` — 6 passed
  - helper delegates to block_aware_cache.release_cache when one is attached; doesn't double-call delete_block_table
  - helper falls back to paged_cache_manager.delete_block_table when no prefix cache
  - helper is a no-op without any paged-cache layer (defensive)
  - `_advance_chunked_prefills` releases on RuntimeError
  - `_schedule_waiting` non-chunked path releases on RuntimeError
  - `_schedule_waiting` chunked first-chunk path releases on RuntimeError

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -2422,6 +2422,7 @@ class Scheduler:
             except RuntimeError as e:
                 logger.error("Chunked prefill failed for %s: %s", rid, e)
                 self._prefill_states.pop(rid, None)
+                self._release_paged_cache_for_request(rid)
                 self.requests.pop(rid, None)
                 get_prefill_tracker().remove(rid)
                 # Drop Metal cache pool buffers held by the aborted chunk's
@@ -5034,6 +5035,7 @@ class Scheduler:
                             request.request_id,
                             e,
                         )
+                        self._release_paged_cache_for_request(request.request_id)
                         self.requests.pop(request.request_id, None)
                         get_prefill_tracker().remove(request.request_id)
                         # Drop Metal cache pool buffers held by the aborted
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, the diagnosis matches exactly and the helper split is right — `release_cache` already does `_request_tables.pop` + `delete_block_table` + ref decrement internally, so the fallback branch only fires when there's no prefix cache, no double-call risk either way. Merging now. I'll push a small follow-up applying the same helper to the `_prepare_batch` preflight rejection (scheduler.py:4756) which has the same leak shape but sits outside this PR's scope. Goes out with the next release.

PR #1451, opened 2026-05-27
fix(ssd-cache): inline LRU unlinks so eviction frees queue capacity

cfbraun:
## Summary

`_enforce_size_limit_for_new_block` enqueues evicted file unlinks as `("unlink", path)` items onto `_write_queue` — the same bounded queue that carries pending writes. Combined with the pre-eviction `_write_queue.full()` short-circuit at the top of `save_block`, this creates a deadlock under sustained save pressure:

1. Writer is saturated → `_write_queue` is full.
2. `save_block`'s pre-eviction check sees `full()` → returns False immediately, **before** calling `_enforce_size_limit_for_new_block`.
3. Eviction never runs → cache stays at the size cap.
4. Every subsequent save drops; `ssd_write_drops` climbs forever while `_total_size` sits pinned at the cap.

Fix: inline the unlinks on the eviction-calling thread instead. Eviction typically removes a single block per save (`evict_until_size` stops as soon as `total_size <= target`), so this is one syscall per save in steady state. The deferred-unlink justification ("avoid blocking the inference thread with N file delete syscalls") doesn't materialise under normal load, and inlining removes the bounded-queue contention entirely.

**Bounded inline burst.** The ENOSPC-recovery path invalidates the 30 s disk-usage cache, which can shrink `_get_effective_max_size` sharply — `evict_until_size` would then return hundreds of entries at once and the inline-unlink loop would stall the inference thread on a syscall storm. Cap the burst at `_MAX_INLINE_UNLINKS_PER_SAVE = 32` and reinsert the deferred metadata into the index so subsequent saves drain the remainder.

Also adds an `evict_unlink_failures` stats counter (eviction now decrements the index before the on-disk unlink; if `Path.unlink` raises `OSError`, surfacing the counter lets operators see when on-disk size has drifted above what the index reports).

The dead writer-thread `("unlink", file_path)` dispatch branch is removed since no path enqueues such tuples anymore.

## Test plan

- [x] `pytest tests/test_paged_ssd_cache.py::TestInlineLRUUnlinks` — 4 passed
- [x] `pytest tests/test_paged_ssd_cache.py tests/test_hot_cache.py` — 125 passed (4 new + 121 existing)

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -71,6 +71,17 @@ def _compute_max_pending_writes() -> int:
 
 _MAX_PENDING_WRITES = _compute_max_pending_writes()
 
+# Cap on the number of LRU blocks ``_enforce_size_limit_for_new_block`` is
+# allowed to unlink in one inline burst. Eviction normally returns ~1
+# entry; the cap exists for the ENOSPC-recovery path where the disk-usage
+# cache invalidates and the next ``_get_effective_max_size`` call can
+# shrink sharply — ``evict_until_size`` would then return hundreds of
+# entries at once and stall the inference thread on a syscall storm.
+# Deferred-but-not-unlinked entries are reinserted into the index so
+# subsequent saves drain the remainder; bounds per-call latency at the
+# cost of taking multiple saves to fully reconverge.
+_MAX_INLINE_UNLINKS_PER_SAVE = 32
+
 
 # Cache format version. Bump when on-disk layout or RotatingKVCache meta_state
 # semantics change in a way that older blocks become unsafe to load.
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for tracking this down. The root cause makes sense: eviction was sharing the bounded write queue with pending writes, so once the queue saturated, cache pressure could not drain cleanly.

This looks good to me, and I am going to merge it. I verified tests/test_paged_ssd_cache.py and tests/test_hot_cache.py locally, and I will fold two small follow-ups into main after merge: preserving LRU order by sorting _lru by last_access after capped inline eviction reinserts deferred entries, and threading evict_unlink_failures through PagedSSDCacheStats so it shows up in the runtime/admin stats path.

PR #1452, opened 2026-05-27
feat(scheduler): binding-ceiling-aware prefill rejection messages

cfbraun:
## Summary

When the prefill memory guard rejects a request, current main returns a generic *"reduce context length, free system memory, or loosen `memory_guard_tier`"* message regardless of which of the three ceilings (`static` / `dynamic` / `metal_cap`) actually bound the hard limit. Operators are left guessing which of three knobs applies. The painful case is `metal_cap`-bound 413s on hosts where `iogpu.wired_limit_mb` was never raised — no amount of freeing system memory would have helped, but the message didn't say so.

This PR identifies the binding ceiling and steers the user to the right knob:

| Binding | Message |
|---|---|
| `metal_cap` | "raise kernel `iogpu.wired_limit_mb` in Terminal (currently caps Metal at …), or reduce context length" |
| `dynamic` under `custom` tier | "raise `custom_ceiling_bytes` in admin Memory settings (currently pinned at …), or reduce context length" |
| `dynamic` on a reclaim tier with `static` headroom | "close other apps to free RAM (static cap is X but only Y is reclaimable right now), raise `memory_guard_tier`, or reduce context length" |
| `static`-bound (or no breakdown available) | "reduce context length or raise `memory_guard_tier`" (today's generic fallback) |

## Wiring

- `ProcessMemoryEnforcer._get_ceiling_breakdown()` returns the three components alongside `hard_limit`. `_get_hard_limit_bytes` becomes a thin wrapper so hot callers that only need the `min` keep their existing API.
- `_propagate_memory_limit()` pushes `static` / `dynamic` / `metal_cap` / `tier` onto each scheduler in the same loop that already pushes `hard_limit` (one breakdown computation per tick — no extra `sysctl` calls).
- `Scheduler` gains `_memory_static_ceiling_bytes`, `_memory_dynamic_ceiling_bytes`, `_memory_metal_cap_bytes`, `_memory_guard_tier` fields; rejection-message generation is consolidated into `_format_rejection_message()` shared by `_preflight_memory_check` (Request-form) and `preflight_or_raise` (token-form).

## Backwards compatibility

Behavior is unchanged when the component fields stay at their 0 init values (direct test instantiation, schedulers wired outside the enforcer): the generic fallback message fires, byte-identical to pre-PR output.

## Test plan

- [x] `test_engine_preflight.py`: 4 new tests pinning each branch of the advice ladder (metal_cap → sysctl, dynamic+custom → custom_ceiling_bytes, dynamic+reclaim-tier → "close other apps", static-bound → generic).
- [x] `test_process_memory_enforcer.py`: new propagation test for the four component fields + tier; existing tests updated to mock both `_get_hard_limit_bytes` and `_get_ceiling_breakdown` via `_make_enforcer`.
- [x] 261 tests pass locally (touched files: `test_engine_preflight.py`, `test_process_memory_enforcer.py`, `test_scheduler.py`).


jundot:
Thanks for the detailed wiring here. The binding-aware preflight rejection message is a useful improvement over the generic advice.

I found one small consistency gap around hot-cache reservation: the scheduler hard limit is reduced by `_scheduler_limit_bytes()`, while the propagated component ceilings stay unadjusted, so the binding label can fall back to `effective ceiling` in that configuration. It does not change the guard behavior, only the diagnostic message, so I will fold that into a follow-up along with the stale `format_bytes` import and the `str.capitalize()` casing issue.

PR #1458, opened 2026-05-27
docs(packaging): drop stale "DMG produced by the Swift build" claim

### LINKED ISSUE #1456: Docs claim Swift build produces a DMG, but no script in-tree does
## Problem

`packaging/README.md` (Installation section) currently states:

> 1. Open the DMG produced by the Swift build.

But after #1355 (which removed `packaging/build.py`'s `create_dmg` / `sign_app` / bundle pipeline), nothing in the repo actually creates a `.dmg`:

- `apps/omlx-mac/Scripts/build.sh release` produces `oMLX.app` only — no DMG step (`grep -rn "hdiutil\|create-dmg\|\\.dmg" apps/omlx-mac/Scripts/` returns nothing)
- No CI workflow under `.github/workflows/` creates one either
- The Swift updater (`AppUpdater.swift`) *consumes* DMGs from GitHub Releases but doesn't make them

The DMGs at https://github.com/jundot/omlx/releases presumably come from a maintainer-side script that isn't in the public tree.

## Impact

Developers building locally from `packaging/README.md` end up with `build/Stage/oMLX.app` but no DMG, contradicting the install instructions one section down. The drop-Swift-replaces-PyObjC migration removed the DMG-producing code without replacing it or updating the docs.

## Suggested fixes

Either:

1. **Add an opt-in DMG step to `apps/omlx-mac/Scripts/build.sh`** — e.g. a `--dmg` flag that wraps `$STAGED_APP` with `hdiutil create -volname oMLX -srcfolder ... -format UDZO`. Keeps the default flow unchanged for the auto-rebuild / Xcode-run loop, but gives local builders a documented way to produce a DMG matching the Releases shape. Update `packaging/README.md` to reflect the new flag.

2. **Or remove the DMG claim from `packaging/README.md`** and document that DMGs only come from official releases.

I'll open a PR doing (1) since it lines up with the README's existing wording. Open to either resolution.

cfbraun:
## Summary

Closes #1456. (Supersedes #1457, which was auto-closed when I renamed the branch — content is identical.)

After #1355 retired `packaging/build.py`'s `.app` + DMG pipeline, nothing in the repo creates a `.dmg`:

- `apps/omlx-mac/Scripts/build.sh release` produces `oMLX.app` directly — `grep -rn "hdiutil\|create-dmg\|\.dmg" apps/omlx-mac/Scripts/` returns nothing
- No `.github/workflows/` entry produces one
- The Swift updater consumes DMGs from GitHub Releases but doesn't make them

Meanwhile `packaging/README.md` still told developers to "Open the DMG produced by the Swift build" — that DMG doesn't exist.

## What's in this PR

Just the README fix. The Installation section now reflects what actually happens:

1. Drag `build/Stage/oMLX.app` to `/Applications`, or `open` it in-place
2. (steps 2–3 unchanged)

Plus a note that the Releases DMGs come from an off-tree maintainer pipeline, so end users follow the Releases path and this section is explicitly scoped to developers building from source.

## Why no `hdiutil` step in `build.sh`?

An earlier revision of #1457 added an opt-in `--dmg` flag. Pulled it: a developer who built `oMLX.app` locally already has the artifact they need — dragging the `.app` to `/Applications` is the standard macOS install pattern. A DMG is a distribution wrapper, useful when shipping to users (single signed download, polished volume window); none of that helps the local-build flow.

The one niche where local DMG creation has value is dry-running the auto-updater download path — better served by hand-rolling a DMG once for that specific test than by paying a maintenance tax (and 5–10 s extra build time) on every release build.

If maintainers want the official Releases DMG pipeline brought in-tree (CI workflow, code-signing, notarization), that's a separate, larger discussion.

## Test plan

- [x] `cat packaging/README.md` — Installation section reads coherently and matches what `build.sh release` actually produces.

### DIFF
--- a/packaging/README.md
+++ b/packaging/README.md
@@ -56,8 +56,16 @@ No application layer — the Swift app is the application surface.
 
 ## Installation
 
-1. Open the DMG produced by the Swift build.
-2. Drag `oMLX.app` to Applications.
-3. Launch the app (appears in the menubar).
-4. Walk through the first-run wizard (Storage + API key), then Start
+The Swift build (`build.sh release`) produces `build/Stage/oMLX.app`
+directly — no DMG step. To install:
+
+1. Drag `build/Stage/oMLX.app` to `/Applications`, or `open` it
+   in-place to launch from `build/Stage/`.
+2. Launch the app (appears in the menubar).
+3. Walk through the first-run wizard (Storage + API key), then Start
    Server.
+
+> The DMGs on the [Releases](https://github.com/jundot/omlx/releases)
+> page are produced by an off-tree maintainer pipeline, not by anything
+> in this repo. End users follow the Releases install path; this
+> section is for developers building from source.


jundot:
Thanks for cleaning this up. The docs-only direction looks right to me: the repo no longer creates a DMG, and the updater only consumes release DMGs.

One small path inconsistency remains: from the repo root, build.sh stages the app at `apps/omlx-mac/build/Stage/oMLX.app`, not `build/Stage/oMLX.app`. This is minor, so I'm going to merge this and fold that correction into a follow-up commit.

Issue #1413, opened 2026-05-25
_reconcile_mtp_to_standard aborts (Fatal Python error) when SSD cache has stale blocks from prior model version


fparrav:
## Description

After updating oMLX (which included PR #1404 — attaching the VLM MTPModule unconditionally), the server started crashing with `Fatal Python error: Aborted` during active MTP sessions. The crash always happens in `_reconcile_mtp_to_standard`, triggered by a concurrent incoming request via `patched_extend`.

## Environment

- macOS 26.5 (25F71), Apple M4 Max, 64 GB unified memory
- oMLX v0.3.10 (DMG)
- Model: `Jundot/Qwen3.6-35B-A3B-oQ4-mtp`
- `max_concurrent_requests: 3`, `chunked_prefill: true`, `mtp_enabled: true`
- SSD cache: `~/.omlx/cache`, 55 GB limit

## What happened

Every request from session start triggered this warning:

```
omlx.cache.prefix_cache - WARNING - Cache layer count mismatch: block has 30 layers,
expected 40. Invalidating cache hit.
omlx.cache.prefix_cache - INFO - Partial cache reconstruction: 28/31 blocks, 57344 tokens
```

366 SSD cache blocks (25.5 GB) were left over from a previous session where the model ran with 30 layers. After PR #1404, the same model name now loads with 40 layers — the model name matches so the model-name guard passes, but the layer-count guard fires on every block, forcing partial reconstruction on every request.

As the session progressed (~15 turns, context growing to ~70k tokens), MTP acceptance rate degraded from 88% to ~49%. When a second concurrent request arrived, `patched_extend` called `_reconcile_mtp_to_standard`, which ran `_call_backbone` on a fresh empty cache → `mx.eval(next_tok)` → Metal GPU abort.

## Crash stack (from `~/.omlx/logs/crash.log`)

```
Fatal Python error: Aborted

Thread 0x000000016cb53000 (most recent call first):
  File ".../omlx/patches/mlx_lm_mtp/batch_generator.py", line 527 in _reconcile_mtp_to_standard
  File ".../omlx/patches/mlx_lm_mtp/batch_generator.py", line 141 in patched_extend
  File ".../mlx_lm/generate.py", line 1819 in _next
  File ".../mlx_lm/generate.py", line 1866 in next_generated
  File ".../omlx/scheduler.py", line 5595 in step
```

Line 527 is `mx.eval(next_tok)` — the lazy MLX graph fires here and the Metal shader aborts.

[...]

Issue #1404, opened 2026-05-25
Loading a quantized MTP (Qwen) model with MTP disabled breaks vision (after first oMLX restart, uh)


zviratko:
How to reproduce
1) quantize Qwen3.6-35B-A3B into oQ6 with MTP enabled and vision preserved
2) after quantization, ask it about an image - this works
3) restart oMLX
4) model loses vision after loading, log below
(I have logs of the first chat and the quantization if wanted)
5) enabling MTP fixes VLM and the model vision starts working
6) unquantized model works with both MTP enabled and disabled

Same happens with OsaurusAI/Qwen3.6-35B-A3B-MXFP8-MTP so it's not just oQ problem

```
2026-05-25 12:23:37,711 - omlx.server - TRACE - [-] - Incoming POST /v1/chat/completions — body: {"model":"Qwen3.6-35B-A3B-oQ6-mtp","messages":[{"role":"system","content":"Be concise"},{"role":"user","content":"hello"}],"stream":true,"stream_options":{"include_usage":true}}
2026-05-25 12:23:37,718 - omlx.server - DEBUG - [-] - Chat completion request received: model=Qwen3.6-35B-A3B-oQ6-mtp, messages=2, stream=True, max_tokens=None, temp=None
2026-05-25 12:23:37,718 - omlx.server - TRACE - [-] -   Message[0]: role=system, content=Be concise...
2026-05-25 12:23:37,718 - omlx.server - TRACE - [-] -   Message[1]: role=user, content=hello...
2026-05-25 12:23:37,718 - omlx.engine_pool - INFO - [-] - Loading model: Qwen3.6-35B-A3B-oQ6-mtp
2026-05-25 12:23:37,741 - omlx.patches.mlx_lm_mtp.qwen35_model - INFO - [-] - Qwen3.5/3.6 MTP model patch applied (PR 990)
2026-05-25 12:23:37,742 - omlx.patches.mlx_lm_mtp.deepseek_v4_model - DEBUG - [-] - DeepSeek-V4 module not registered; skipping MTP patch (this is expected for non-DeepSeek models)
2026-05-25 12:23:37,742 - omlx.patches.mlx_lm_mtp - DEBUG - [-] - DeepSeek-V4 MTP patch did not apply (likely missing base patch)
[...]

PR #1467, opened 2026-05-27
fix(mtp): guard _reconcile_mtp_to_standard against unbounded re-prefill

ivaniguarans:
## Summary

`_reconcile_mtp_to_standard` re-prefills the entire token history through the backbone in a single `_call_backbone` call when transitioning from MTP to standard decode. At long contexts (70K+ tokens, reported on Qwen3.6-35B-A3B), this unbounded re-prefill triggers a Metal GPU abort (SIGABRT) that kills the server process — uncatchable by the existing try/except since SIGABRT fires on the Metal completion thread.

The crash surface requires stale SSD cache blocks from a prior model version (layer count mismatch after an upgrade like #1404), which cause repeated cache invalidations and allow context to grow without cache benefit.

## Changes

A module-level `_RECONCILE_MAX_TOKENS` constant (16384) caps how many tokens the reconcile function will attempt to re-prefill. When exceeded, the function returns False immediately — the caller (`patched_extend`) drops MTP state unconditionally regardless of the return value, so the batch falls back to standard decode with at most one stale token. This is the same degradation path as the existing empty-tokens and rebuild-failure cases.

The threshold is 8x the scheduler's `prefill_step_size` (2048) and ~4.3x below the observed crash point (70K), giving wide safety margin across model sizes.

## Test plan

- Two new tests: guard fires above the limit (backbone never called), reconcile proceeds normally at the limit.
- Full suite: 4585 passed, 19 skipped, 59 deselected.
- Verified on M5 Max 128GB — 100-token reconcile completes normally, 17K-token reconcile is blocked by guard with warning log.

Related to #1413

### DIFF
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -77,6 +77,8 @@ from typing import Any, Deque, List, Optional, Tuple
 
 logger = logging.getLogger(__name__)
 
+_RECONCILE_MAX_TOKENS = 16384
+
 # ---------------------------------------------------------------------------
 # Public entry point
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the detailed repro on M5 Max and for tracing the crash down to `_reconcile_mtp_to_standard`. The diagnosis in #1413 is spot on, especially the part about the stale SSD blocks driving the whole chain.

I want to take a different approach for the fix though. The 16K token cap stops the final crash, but it leaves the stale 25.5GB of blocks on disk and the layer-mismatch warnings keep firing every request. Even with this guard, the affected setup keeps degrading until someone runs `rm -rf ~/.omlx/cache`. And on a clean install, legitimate >16K MTP use (Qwen3.6 supports 256K) would silently fall back to plain drop with only a warning log.

I'm going to fix this at the SSD scan layer instead. `PagedSSDCacheManager._scan_existing_files()` already reads `num_layers` and `model_name` from block metadata, so it can validate against the current model at startup and unlink mismatched blocks. That removes the entire cascade your repro hit, including the auto-cleanup of the 25.5GB you had to clear manually. Going out in the next release.

Closing this PR but really appreciate the analysis, it made the root cause obvious.

PR #1470, opened 2026-05-27
feat: expose oMLX version via CLI

chenqianhe:
Adds a top-level `omlx --version` flag backed by the existing `omlx._version.__version__` source of truth.

## Changes

- Add `omlx --version`
- Add CLI test coverage for version output
- Leave existing server metadata behavior unchanged

### DIFF
--- a/omlx/cli.py
+++ b/omlx/cli.py
@@ -18,6 +18,8 @@ import argparse
 import faulthandler
 import sys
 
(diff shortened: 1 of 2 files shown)


jundot:
Thanks! Clean addition, and re-using `_version.__version__` keeps it drift-free. Merging now.

PR #1453, opened 2026-05-27
feat(packaging): bundle xgrammar in the venvstacks export (no-torch stub)

cfbraun:
## Summary

Enables structured output (xgrammar) in the Swift macOS bundle. After #1355 / #1371 the legacy DMG path retired and `packaging/build.py` now only produces the venvstacks Python layers consumed by `apps/omlx-mac/Scripts/build.sh`. Bundle users who try `/v1/chat/completions` with `response_format` currently fall through to "structured output unavailable" — the alternative (`brew install` with `--with-grammar`) pulls real torch (~500 MB) and works, but the bundled-app flow has no equivalent today.

The blocker: xgrammar declares `torch>=1.10.0` as a runtime dep, but oMLX only exercises its non-torch paths (numpy bitmasks + MLX kernel). Module-level `import torch` in `xgrammar.matcher` / `xgrammar.testing` / `xgrammar.contrib.hf` / `tvm_ffi.core` aborts the omlx grammar codepath import on a torch-less install.

## What's in this PR

1. **`omlx/_torch_stub.py`** — a minimal torch surface for xgrammar's import-time references. When real torch is on the path, `install()` finds it via `importlib.util.find_spec("torch")` and bows out as a no-op. Stubbed helpers that would actually call into torch raise a clearly-tagged `RuntimeError "torch.X not supported in omlx stub"` so a future code path that *does* reach into torch-land surfaces with an identifiable error rather than silently misbehaving.

2. **`_install_xgrammar()` in `packaging/build.py`** — downloads `xgrammar` + `apache-tvm-ffi` wheels `--no-deps` (skip torch), extracts them into the `framework-mlx-framework` site-packages layer of the venvstacks export. Hooked into `build_venvstacks()` right after `_install_paroquant()`. Idempotent via an atomic sentinel file with an integrity check (handles SIGKILL / ENOSPC mid-extract). Version constants imported from the stub so the two stay in lockstep.

3. **Post-strip safety check in `_strip_unused_packages()`** — fails the build if any torch artifact survived the strip step. A partial torch (some files but not enough for xgrammar) would make `find_spec("torch")` return truthy, the stub would short-circuit, and xgrammar would import a broken half-torch and fail at runtime. Fail fast at build time.

4. **`install()` call sites** — one each at the xgrammar entry points: `api/grammar.py`, `server.py`, `admin/routes.py`, `tests/conftest.py`. With real torch present each is a no-op.

[...]

PR #1471, opened 2026-05-27
fix(cache): boundary snapshots dropped during prefill because uid mapping not yet populated

cfbraun:
## Summary

Prefill-side boundary snapshots were being silently dropped for ArraysCache / GDN / hybrid models, causing every non-last cached block to store a placeholder and every identical-prefix re-upload to re-prefill from scratch.

`_emit_prefill_boundary_snapshot` (scheduler.py:2549) routed through:

```python
self._on_prefill_boundary_snapshot(
    self.request_id_to_uid.get(request.request_id, -1),  # → -1
    snapshot_cache,
    total_tokens,
)
```

But `request_id_to_uid` is only populated when the request is inserted into `BatchGenerator` via `_insert_prefilled_request`, which runs **after** prefill completes. During prefill the lookup defaults to `-1`. `_on_prefill_boundary_snapshot` then does:

```python
request_id = self.uid_to_request_id.get(uid)  # uid=-1 → None
if request_id is None:
    return  # silent
```

→ every boundary-snapshot call during prefill was a no-op.

For sliceable cache types (KVCache, BatchKVCache, etc.) this is invisible — those layers don't need snapshots. For ArraysCache / GDN layers, every non-last block in `prefix_cache.store_cache` stores a `(mx.zeros((1,)), mx.zeros((1,)))` placeholder when no snapshot is available. The next identical-prefix request matches blocks by hash, finds a placeholder shape `(1,)` in the last matched block, and the rejection at `prefix_cache.py:1986` correctly bails — but the resulting "Request will reprocess from scratch" branch re-prefills the entire prompt.

## User-visible symptom

Hybrid-architecture models (e.g. Qwen3-Next / Qwen3.6-A3B) re-prefilling the full prompt on every chat continuation, even when only one or two new turns were appended. Log signature:

```
ArraysCache layer 0: partial prefix match detected (placeholder in last matched block).
Rejecting cache to prevent stale GDN state. Request will reprocess from scratch.
```

Repeating every ~few minutes under normal chat use.

## The fix

Pass `request_id` through directly. `_on_prefill_boundary_snapshot` already keyed everything by `request_id` internally — the uid round-trip was pointless.

```python
self._on_prefill_boundary_snapshot(
    request.request_id,
    snapshot_cache,
    total_tokens,
)
```

And the helper's signature changes from `uid: int` to `request_id: str`, dropping the now-unused `uid_to_request_id` lookup.

The other prefill-side caller, `_maybe_capture_boundary_snapshot` (called during generation from `_step` after the request has been inserted, where uid IS valid), writes snapshots inline without going through this helper, so it was unaffected. That asymmetry explains the bimodal pattern in production logs: **short-prompt + long-generation** requests captured 20+ intermediate snapshots cleanly (boundaries crossed during generation), while **long-prompt + short-generation** requests captured 0 (all boundaries in prefill, where the uid lookup failed).

## Diagnosis evidence

End-to-end repro on a Qwen3.6-35B-A3B (GDN layer 0) instance, 90k-token prompt:

**Before fix** (every block boundary):
```
SNAP_EMIT request=d6655b20-... tokens=2048 non_none_layers=30/40   ← emit fires
SNAP_SKIP tokens=2048 uid=-1 reason=no_request_id_for_uid           ← dropped
SNAP_EMIT request=... tokens=4096 non_none_layers=30/40
SNAP_SKIP tokens=4096 uid=-1 reason=no_request_id_for_uid
...
Using boundary cache snapshot for ...: storing 88064/90919 tokens (skipping trailing partial block, 0 intermediate snapshots)
```

**After fix**:
```
SNAP_EMIT request=1a58e4f5-... tokens=2048 non_none_layers=30/40
SNAP_OK request=1a58e4f5-... tokens=2048 saved_to_ssd=True
...
Using boundary cache snapshot for 1a58e4f5-...: storing 90112/90919 tokens (skipping trailing partial block, 43 intermediate snapshots)
```

Subsequent identical-prefix request started prefill at token 51200 (previous cached boundary) instead of 0 — no rejection line.

## Test plan

- [x] Updated `test_prefill_boundary_snapshot_records_rotating_cache` to NOT pre-populate `request_id_to_uid` / `uid_to_request_id` — matches actual prefill state.
- [x] Updated `test_prefill_boundary_snapshot_ignores_non_boundary_token_count` similarly.
- [x] Added `test_emit_prefill_boundary_snapshot_persists_before_uid_assignment` — pins the regression: calling `_emit_prefill_boundary_snapshot` while the request has not been inserted into BatchGenerator must still record the snapshot.
- [x] `pytest tests/test_scheduler.py -k "prefill_boundary or emit_prefill_boundary"` — 3 passed.
- [x] Confirmed end-to-end via instrumentation on a real 90k-token chat continuation against Qwen3.6-35B-A3B-UD-MLX-4bit.


jundot:
Thanks for the diagnosis. The request_id_to_uid timing explanation is exactly right, and the short-prompt+long-gen vs long-prompt+short-gen bimodal pattern matches what I've been seeing on Qwen3.6-A3B sessions. Cherry-picked the boundary-snapshot fix onto main as 9407468 with your authorship preserved.

I left the other 23 commits on this branch alone. They overlap with your narrower PRs (#1447, #1450, #1451, #1452, #1453, etc.), so I'd rather review and merge those independently to keep the main history readable. Closing this PR; the narrow ones stay open. Apologies if that's not what you intended. Happy to revisit if you'd prefer the bundle path.

cfbraun:
Thanks for the merge of 9407468 — and for the careful read on the bimodal pattern.

You're right about the fat-PR shape — that's my mistake. I branched `pr/prefill-snapshot-uid-fix` off my local `main`, which is ~24 commits ahead of upstream because it carries the other in-flight work. GitHub then computed the cross-fork PR diff as all 24 commits, not just my one new commit. From your side it correctly looked like a bundle. I'll branch future single-fix PRs off `upstream/main` directly to keep the diff narrow.

Happy to leave the narrower PRs (#1446–#1453, #1458) on their independent tracks per your suggestion. No need to revisit. Thanks again.

PR #1479, opened 2026-05-27
fix(server): chunk-form SSE keepalive shares the stream's completion id

### LINKED ISSUE #1478: Chunk-form SSE keepalive emits a `chat.completion.chunk` with a mismatched `id` (`chatcmpl-keepalive`), breaking OpenAI-spec stream accumulators (openai-go) — streamed `tool_calls`/`usage` silently dropped
**Describe the bug**

Since v0.3.9, the default SSE keepalive mode (`chunk`) emits a `chat.completion.chunk` whose `id` differs from the rest of the stream. This violates the OpenAI streaming contract (every chunk in one streamed completion shares the same `id`) and breaks spec-conforming stream accumulators. Concretely, the **official OpenAI Go SDK (`openai-go`)** silently drops the entire completion — streamed `tool_calls`, `finish_reason`, and `usage` are all lost — which surfaces as an empty assistant message on any tool-calling turn.

The keepalive frame is sent as the **first frame of every stream** (verified across thinking on/off, with/without tools, all prompt sizes), so every streaming response is affected; it only becomes *visible* when the meaningful payload lives in the accumulator (`tool_calls`/`usage`) rather than in incrementally-streamed `content` deltas.

**Details**

oMLX emits this as the first SSE frame (note `id:"chatcmpl-keepalive"`, `model:"keepalive"`, `created:0`):

```
data: {"id":"chatcmpl-keepalive","object":"chat.completion.chunk","created":0,"model":"keepalive","choices":[{"index":0,"delta":{"content":""},"finish_reason":null}]}
```

The real completion chunks then arrive under a *different*, freshly-minted id:

```
data: {"id":"chatcmpl-1a2b3c4d", ... "delta":{"tool_calls":[{"index":0,"id":"call_...","function":{"name":"evaluate_ticker","arguments":"{\"ticker\":\"JNJ\"}"}}]}}
data: {"id":"chatcmpl-1a2b3c4d", ... "delta":{}, "finish_reason":"tool_calls"}
data: {"id":"chatcmpl-1a2b3c4d", ... "choices":[], "usage":{...}}
```

The OpenAI Go SDK's `ChatCompletionAccumulator.accumulateDelta` locks onto the first chunk's `id` and rejects every later chunk whose `id` differs:

```go
if len(cc.ID) == 0 {
    cc.ID = chunk.ID          // becomes "chatcmpl-keepalive"
} else if cc.ID != chunk.ID {
    return false              // every real chunk is dropped
}
```

So with the keepalive frame first, the accumulator latches `chatcmpl-keepalive` and discards the role/tool_calls/finish_reason/usage chunks. (`content`, if read from live deltas rather than the accumulator, may still arrive — which is why plain prose replies often look fine while tool-calling turns come back empty.)

**To Reproduce**

1. Start oMLX (defaults; `sse_keepalive_mode` = `chunk`) with any chat model that uses tools, e.g. `gemma-4-26b-a4b-it-8bit`.
2. Request a streaming chat completion with a tool defined:

```bash
curl -sN http://localhost:8000/v1/chat/completions \
  -H 'Content-Type: application/json' -d '{
    "model":"gemma-4-26b-a4b-it-8bit",
    "messages":[{"role":"user","content":"Is JNJ a good buy?"}],
    "tools":[{"type":"function","function":{"name":"evaluate_ticker","description":"Evaluate a ticker.","parameters":{"type":"object","properties":{"ticker":{"type":"string"}},"required":["ticker"]}}}],
    "stream":true
  }' | head
```

3. Observe the first SSE frame is `id:"chatcmpl-keepalive"`, followed by real chunks under a different id.
4. Consume the same request with the official OpenAI Go SDK using `ChatCompletionAccumulator`: `acc.Choices[0].Message.ToolCalls`, `FinishReason`, and `acc.Usage` are all empty, despite the tool call being present on the wire.

**Expected behavior**

A streaming response should be consumable by spec-conforming OpenAI SDKs without dropping `tool_calls`/`finish_reason`/`usage`. Keepalive frames should not change the per-stream `id` invariant.

**Likely root cause**

Introduced by `141f708` ("Switch SSE keepalive default to chunk form for OpenClaw / WorkBuddy compatibility"), which made `chunk` the default `sse_keepalive_mode`. `_with_sse_keepalive` yields the keepalive frame *before* the wrapped generator runs, at which point the real `response_id` (minted inside `stream_chat_completion`, ~`server.py:2788`) does not exist yet — so the frame falls back to the static `chatcmpl-keepalive` sentinel id.

**Suggested fix**

Mint `response_id` in the request handler *before* `_with_sse_keepalive`, pass it into both the keepalive-frame builder and `stream_chat_completion`, and emit the keepalive chunk with that real id. An empty-delta chunk that *shares* the stream id is accepted by spec accumulators **and** stays parseable for the comment-intolerant clients (OpenClaw/WorkBuddy, #839/#1035) the chunk mode was added for — so it satisfies both camps without a behavior toggle. (Emitting the heartbeat as an SSE comment for the OpenAI protocols would also work but re-breaks those clients, so the id-consistent chunk seems preferable.)

**Workaround**

Setting `sse_keepalive_mode = comment` (or `off`), or `--sse-keepalive-mode comment`, restores pre-v0.3.9 behavior and resolves the incompatibility.

**Desktop:**
 - macOS Version: 26.3 (build 25D125)
 - oMLX Version: 0.3.12 (regression present since 0.3.9; 0.3.8 and earlier used the SSE-comment form and are unaffected)

**Additional context**

Affects clients built on the official OpenAI Go SDK (`openai-go`), and likely any consumer using a strict OpenAI stream accumulator that assumes a single `id` per streamed completion. The Go SDK's id-consistency check is intentional (it guards against splicing two completions together), so the mismatch is on the producer side — the keepalive frame is typed as a `chat.completion.chunk` but carries sentinel `id`/`model`/`created` values that a real completion never would.

jcalvert:
## Problem

Since v0.3.9 the default SSE keepalive mode (`chunk`) emits a `chat.completion.chunk` whose `id` is the sentinel `chatcmpl-keepalive`, sent as the first frame of every stream. The real completion chunks then arrive under a different, freshly-minted id.

Strict OpenAI stream accumulators assume all chunks in one streamed completion share a single `id`. The official OpenAI Go SDK (`openai-go`) `ChatCompletionAccumulator` latches the first chunk's id and silently rejects every later chunk whose id differs:

```go
if len(cc.ID) == 0 {
    cc.ID = chunk.ID          // becomes "chatcmpl-keepalive"
} else if cc.ID != chunk.ID {
    return false              // every real chunk dropped
}
```

So the streamed `tool_calls`, `finish_reason`, and `usage` are all discarded, producing an empty assistant message on any tool-calling turn. (Plain `content` can survive if read from live deltas, which is why prose replies often look fine while tool-calling turns come back empty.)

Reported in #1478. This is a regression from the fix for #839, which switched the default keepalive from an SSE comment to this chunk form.

## Fix

Pre-mint `response_id` in `create_chat_completion` and reuse it for both the keepalive frame and `stream_chat_completion`. The keepalive (`_chat_keepalive_chunk`) now carries the stream's real id, so it is a true no-op for spec accumulators while remaining a parseable data event for the comment-intolerant clients chunk mode was added for (OpenClaw / WorkBuddy, #839/#1035). `comment` and `off` modes are unchanged.

## Verification

- New unit tests in `tests/test_sse_keepalive.py` assert the chat keepalive frame reuses the given id and never emits the sentinel.
- Validated end-to-end against a local build: with the patch the first keepalive frame's `id` matches the completion id, and an unmodified `openai-go` client that previously received empty tool-calling turns now gets the tool call + usage correctly.

## Notes

- Touches the same `create_chat_completion` streaming `return` as open PR #1452 (preflight guards) — only trivial line-context proximity, no logic overlap; may need a one-line rebase if both land.

Fixes #1478.

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -1359,6 +1359,26 @@ def _resolve_keepalive(protocol: str) -> Optional[str]:
     return None
 
 
+def _chat_keepalive_chunk(response_id: str) -> str:
+    """Keepalive frame that shares the stream's completion id.
+
+    The static ``_KEEPALIVE_CHAT_CHUNK`` carries a sentinel id
+    (``chatcmpl-keepalive``) that differs from the real completion chunks.
+    Strict OpenAI stream accumulators (e.g. the official ``openai-go`` SDK)
+    assume every chunk in one streamed completion shares a single ``id``: they
+    latch the first chunk's id and silently drop later chunks whose id differs,
+    discarding the real ``tool_calls``/``finish_reason``/``usage``. Emitting the
+    keepalive with the stream's own ``response_id`` makes it a true no-op for
+    those clients while remaining a parseable data event for clients that can't
+    handle SSE comment lines.
+    """
+    return (
+        'data: {"id":"' + response_id + '","object":"chat.completion.chunk",'
+        '"created":0,"model":"keepalive",'
+        '"choices":[{"index":0,"delta":{"content":""},"finish_reason":null}]}\n\n'
+    )
+
+
 async def _safe_anext(ait):
     """Wrapper for __anext__ that converts StopAsyncIteration to a sentinel.
 
@@ -2314,11 +2334,17 @@ async def create_chat_completion(
         chat_kwargs["stop"] = request.stop
 
     if request.stream:
+        # Pre-mint the completion id so the keepalive frame (emitted before the
+        # generator starts) can share it. See _chat_keepalive_chunk.
+        response_id = f"chatcmpl-{uuid.uuid4().hex[:8]}"
+        keepalive = _resolve_keepalive("openai_chat")
+        if keepalive == _KEEPALIVE_CHAT_CHUNK:
+            keepalive = _chat_keepalive_chunk(response_id)
         return StreamingResponse(
             _with_sse_keepalive(
-                stream_chat_completion(engine, messages, request, model_load_duration=model_load_duration, resolved_model=resolved_model, **chat_kwargs),
+                stream_chat_completion(engine, messages, request, model_load_duration=model_load_duration, resolved_model=resolved_model, response_id=response_id, **chat_kwargs),
                 http_request=http_request,
-                keepalive_chunk=_resolve_keepalive("openai_chat"),
+                keepalive_chunk=keepalive,
             ),
             media_type="text/event-stream",
             headers={"X-Accel-Buffering": "no", "Cache-Control": "no-cache"},
@@ -2770,6 +2796,7 @@ async def stream_chat_completion(
     request: ChatCompletionRequest,
     model_load_duration: float = 0.0,
     resolved_model: Optional[str] = None,
+    response_id: Optional[str] = None,
     **kwargs,
 ) -> AsyncIterator[str]:
     """Stream chat completion response.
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for catching this and the clean fix. Merging into the next release.

PR #1484, opened 2026-05-28
fix(embedding): bound per-request embedding batch size

chenqianhe:
## Summary

Fixes the embedding memory spike caused by unbounded per-request embedding batches.

The root problem was that `/v1/embeddings` treated the whole `input` array as one model forward. `max_concurrent_requests` only limits how many requests run at once; it did not limit how many texts a single embedding request could pass into `model.embed(...)`.

That meant a request with 1,000 input texts was still one request, but all 1,000 texts could be embedded in one forward pass.

This PR adds a dedicated `embedding_batch_size` setting so each embedding request is internally chunked:

- Adds `scheduler.embedding_batch_size`
- Defaults embedding batch size to `32`
- Applies it in `EmbeddingEngine.embed()`
- Preserves output ordering and API response shape
- Supports runtime updates for loaded embedding engines
- Exposes the setting in Web Admin and macOS Performance settings
- Adds CLI/env/settings support

## Why this is separate from Max Concurrent Requests

`max_concurrent_requests` controls request-level concurrency for completion/chat scheduling.

Embedding needs a different limit: how many input texts from one request are sent through one embedding forward pass. Without this limit, a single request can still create an oversized batch even when request concurrency is low.

## Validation

- `446 passed, 1 deselected`

```bash
env UV_CACHE_DIR=/tmp/uv-cache uv run pytest \
  tests/test_settings.py \
  tests/test_embedding.py \
  tests/test_scheduler.py \
  tests/test_config.py \
  tests/test_admin_server_aliases.py \
  tests/test_cli.py \
  -q

### DIFF
--- a/apps/omlx-mac/Resources/Localizable.xcstrings
+++ b/apps/omlx-mac/Resources/Localizable.xcstrings
@@ -3582,6 +3582,18 @@
         }
       }
     },
+    "performance.error.embedding_batch_size_invalid" : {
+      "comment" : "Performance screen error when embedding batch size input is invalid",
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Embedding Batch Size must be a positive integer."
+          }
+        }
+      }
+    },
     "performance.error.idle_timeout_invalid" : {
       "comment" : "Performance screen error when idle timeout input is below 60 seconds",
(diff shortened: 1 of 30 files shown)


jundot:
Thanks for the clean fix and the thorough tests — chunking, runtime updates, CLI rollback, and the per-chunk metal cache flush are all covered. Squash-merging as-is, will land in the next release.

PR #1498, opened 2026-05-28
Cache SpecPrefill token selections

xxxkkw:
## Summary
- Add an LRU cache for SpecPrefill token-selection results keyed by prompt digest and scoring parameters.
- Reuse cached sparse token selections for repeated long prompts instead of rescoring the same prompt with the draft model.
- Surface cache stats via scheduler stats and clear the cache on scheduler reset.

## Environment
- macOS Darwin 25.4.0 arm64
- Apple M1 Max, 32 GB unified memory
- MLX / oMLX local Python environment on Apple Silicon
- Target model: `mlx-community/Qwen3-0.6B-4bit`
- Draft model: `mlx-community/Qwen3-0.6B-4bit`

## Real-weight benchmark

Protocol: oMLX `BatchedEngine` with SpecPrefill enabled, repeated identical prompt, `specprefill_threshold=1024`, `keep_pct=0.2`, greedy generation. This compares upstream baseline (no selection memoization) against this PR. Both runs use 100K requested input tokens and 10K generated tokens. The tokenizer/chat-template path produced 90,001 prompt tokens for the benchmarked request.

| Branch | Repeat | Prompt tokens | Output tokens | TTFT | Total time | Peak memory | Selection cache |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | --- |
| Upstream baseline | 1 | 90,001 | 10,000 | 234.63s | 446.41s | 18.96 GB | n/a |
| Upstream baseline | 2 | 90,001 | 10,000 | 209.47s | 419.70s | 18.63 GB | n/a |
| This PR | 1 | 90,001 | 10,000 | 252.72s | 465.39s | 18.75 GB | 1 miss, 0 hits |
| This PR | 2 | 90,001 | 10,000 | 12.84s | 224.31s | 7.43 GB | 1 miss, 1 hit |

For the repeated long prompt, the second request's TTFT improves from 209.47s on baseline to 12.84s with this PR, because sparse-token selection is reused instead of rescored. Total latency improves from 419.70s to 224.31s, and peak memory drops from 18.63 GB to 7.43 GB on the repeated request.

## Test plan
- `pytest tests/test_specprefill.py::TestSelectionCache -q`
- Real `mlx-community/Qwen3-0.6B-4bit` 100K/10K repeated-prompt SpecPrefill benchmark above

Note: a full `tests/test_specprefill.py` run in this local environment still needs the async pytest plugin; the targeted cache tests pass and the real engine benchmark above exercises the runtime path.

### DIFF
--- a/omlx/cache/specprefill_selection.py
+++ b/omlx/cache/specprefill_selection.py
@@ -1,0 +1,90 @@
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import hashlib
+from collections import OrderedDict
+from dataclasses import dataclass
+from typing import Iterable
+
+
+@dataclass(frozen=True)
+class SpecPrefillSelectionKey:
+    token_digest: str
+    token_count: int
+    draft_model_name: str
+    keep_pct: float
+    chunk_size: int
+    prefill_step_size: int
+
+
+class SpecPrefillSelectionCache:
+    def __init__(self, max_entries: int = 128):
+        self.max_entries = max(0, max_entries)
+        self._entries: OrderedDict[SpecPrefillSelectionKey, tuple[int, ...]] = OrderedDict()
+        self.hits = 0
+        self.misses = 0
+        self.evictions = 0
+
+    @staticmethod
+    def digest_tokens(tokens: Iterable[int]) -> tuple[str, int]:
+        digest = hashlib.blake2b(digest_size=16)
+        count = 0
+        for token in tokens:
+            digest.update(int(token).to_bytes(8, byteorder="little", signed=True))
+            count += 1
+        return digest.hexdigest(), count
+
+    @classmethod
+    def make_key(
+        cls,
+        tokens: Iterable[int],
+        *,
+        draft_model_name: str,
+        keep_pct: float,
+        chunk_size: int,
+        prefill_step_size: int,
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for the benchmark and the clean implementation. I don't think I can take this as a scheduler-level cache, though.

SpecPrefill selection is request-local in oMLX: it scores the current uncached conversation suffix after system/prefix-cache exclusion, then sparse-prefills that request with a request-specific position offset. That matches the SpecPrefill paper's description of selecting locally important prompt tokens with a draft model and passing them to the main model with positional information, not treating selected indices as a reusable prompt artifact: https://arxiv.org/abs/2502.02789

The draft scoring path also uses stochastic lookahead (`temp=0.6`, `top_p=0.95`), so the selected chunks are not a stable value that should be memoized by prompt digest. The safe reuse path here is the existing draft prefix KV cache, which can avoid re-prefilling repeated draft prefixes while still recomputing the selection for the current prompt. Target sparse KV also cannot be stored as a normal prefix cache entry because the KV contents no longer match the full token sequence.

The benchmark is a repeated identical prompt case, but real chat/tool prompts change on every turn, and caching selected indices would either rarely hit or freeze a per-request heuristic. I'm going to close this one for now.

PR #1505, opened 2026-05-29
refactor(ui): unify Apply scope on Server + Integrations

popfido:
> Leftover UI/UX unification from the Swift App rewrite (#1371) — closes the page-Apply consistency gap surfaced by the post-merge HCI review.

## Summary

Apply the **transactional-unit test** consistently across two screens whose first-class settings had mixed commit semantics.

### What is the transactional-unit test?

A single question for any settings input: *"does this commit as part of one atomic change with other inputs, or is it independent?"* The answer dictates the commit mechanism:

| Case | Pattern | Why |
|---|---|---|
| Inputs that share one logical commit (text fields, toggles that gate dependent text) | **One Apply button bundles them** | Atomic, no half-applied state; matches NetworkScreen + PerformanceScreen |
| Independent inputs (binary switches, decision pickers) | **Commit immediately on change** | Matches macOS System Settings convention — no "apply yes" friction; impossible to forget |
| Hybrid (some independent, some bundled, on the same page) | **Split into visually distinct sections, one Apply per bundled section** | Avoids the "is this page committed or not?" question; each section has a single answer |

Applying that rule to the two screens with mixed semantics:

### ServerScreen — page-wide Apply

Previously the Apply button under Storage only committed Base Path + Models Directory. Port, the six Default Profile sampling fields, and Server Aliases (Advanced) auto-applied on blur via `.onSubmit`. Same page, three different commit semantics — each field worked in isolation but the combination broke the user's mental model of "what does Apply actually apply".

Now:
- Page-wide Apply commits **Port + 6 sampling fields + Storage + Server Aliases** atomically (single `GlobalSettingsPatch` for sampling / aliases / port; storage keeps its existing flow since it moves files and restarts the server)
- Apply gated by per-field baselines; disabled until any draft diverges; re-snapshots on success so the button re-disables after a successful commit
- Advanced section moved above the Apply button so Apply is the page's bottom-most action
- Listen Address picker, Log Level picker, SSE Keep-Alive Mode picker stay immediate-apply via `bind()` — independent decisions, no Apply friction
- Hint footer rewritten to describe the new pattern

### IntegrationsScreen — section Apply for Target Context Size

Claude Code's `targetContextSize` auto-applied on blur while the MCP path used a per-section Apply button — same kind of inconsistency. Now:
- Toggling Context scaling still fires immediately (binary on/off — independent input case)
- Target Context Size gains its own per-section Apply button, visible only when Context scaling is enabled (mirrors the existing MCP Apply pattern — the "hybrid → per-section" case from the table above)
- Apply gated by `targetContextSizeLoaded` baseline (mirrors `mcpConfigLoaded`)

## Test plan

Verified locally with a full Release stage + live server — PATCH path exercised end-to-end.

ServerScreen:
- [x] Initial load: Apply disabled
- [x] Edit Port → Apply enables → click → commits + server restarts
- [x] Edit any sampling field → Apply enables → click → all dirty fields land atomically in one PATCH
- [x] Edit Server Aliases → Apply enables → commits
- [x] Edit Base Path / Models Directory → Apply enables → file move + restart
- [x] Combined edits land in the right order (PATCH → port restart → storage move) on a single Apply click
- [x] Listen Address / Log Level / SSE Keep-Alive picker changes commit immediately; Apply stays disabled
- [x] After successful Apply, baselines re-snapshot and Apply re-disables
- [x] Invalid input (e.g. Temperature 99) → validation error in hint footer, no partial patch

IntegrationsScreen:
- [x] Context scaling OFF → Target Context Size row + Apply hidden
- [x] Context scaling ON → row + Apply appear, Apply disabled at first
- [x] Edit Target Context Size → Apply enables → commits → re-disables
- [x] Toggling Context scaling itself still fires immediately
- [x] MCP Config Path Apply unchanged

## Notes

- Pre-existing `appendInterpolation` deprecation warnings in `QuantizationScreen.swift` are unrelated to this change.
- 8 per-field save methods removed (`saveSamplingContext`, `saveSamplingMaxTokens`, `saveSamplingTemperature`, `saveSamplingTopP`, `saveSamplingTopK`, `saveSamplingRepetitionPenalty`, `savePort`, `saveServerAliases`) plus `applyStorage`; their logic is folded into `applyServerSettings`. `saveHost`, `saveLogLevel`, `saveSseKeepaliveMode`, and `restart` are preserved (used by the immediate-apply pickers and the hero Restart button).
- Diff: 3 files, +236/−152.

### DIFF
--- a/apps/omlx-mac/Sources/AppView/AppServices.swift
+++ b/apps/omlx-mac/Sources/AppView/AppServices.swift
@@ -333,9 +333,10 @@ final class AppServices: NSObject, ObservableObject {
     }
 
     /// Persist a new host/port to AppConfig, reconfigure the running server
-    /// process, and bounce it. Without this, ServerScreenVM.savePort would
-    /// only update the server's `settings.json`, but the next spawn still
-    /// uses the cached --host / --port arguments captured at app launch.
+    /// process, and bounce it. Without this, ServerScreenVM's port path in
+    /// `applyServerSettings` (and `saveHost` for Listen Address) would only
+    /// update the server's `settings.json`, but the next spawn still uses
+    /// the cached --host / --port arguments captured at app launch.
     ///
     /// The Python server is the canonical writer of `settings.json` while
     /// it's running — the caller already PATCHed it before us, so we don't
--- a/apps/omlx-mac/Sources/AppView/Screens/IntegrationsScreen.swift
+++ b/apps/omlx-mac/Sources/AppView/Screens/IntegrationsScreen.swift
@@ -142,9 +142,22 @@ private struct ClaudeCodeSection: View {
                         suffix: "tk",
                         width: 130
                     )
-                    .onSubmit { Task { await vm.save(.targetContextSize, client: client) } }
                 }
             }
+        }
+        if vm.contextScaling {
+            HStack {
+                Spacer()
+                Button(String(localized: "integrations.target_context.apply",
+                              defaultValue: "Apply",
+                              comment: "Apply button for the Claude Code target context size field")) {
+                    Task { await vm.save(.targetContextSize, client: client) }
+                }
+                .buttonStyle(.omlx(.primary))
+                .disabled(!vm.hasPendingContextSizeChange)
+            }
+            .padding(.horizontal, 18)
+            .padding(.top, 6)
         }
     }
 }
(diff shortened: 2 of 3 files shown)


jundot:
Thanks, merged. The transactional-unit framing is a clean way to settle the mixed commit semantics, and it reads much better now.

One follow-up after merge: when port and a storage path both change in a single Apply, the storage restart reused the cached `--port` arg, so the server came back on the old port. I threaded the bundled port through `applyStorageChanges` so that path rebinds correctly now. Everything else matched your test plan.

Issue #1503, opened 2026-05-28
Claude code 2.1.154 throws 422 body -> messages -> 1 -> role: Input should be 'user' or  'assistant'


skudinov:
**Describe the bug**
the latest claude code v2.1.154 fails with the error

**To Reproduce**
just use std omlx env. vars, it worked with 2.1.150

**Expected behavior**
claude code changed something in their API, assuming around tools calling

**Screenshots**

**Desktop (please complete the following information):**
 - macOS Version: [26.5]
 - oMLX Version [0.3.12]

Yevhen-Myroshnychenko:
Have the same issue

Patt92:
me too

dhanna11:
The issue is that Claude v2.1.154 sends a request with “system” instead of “user” or “assistant”. I'm away from my keyboard but please attach the server log from omlx when you see this issue so the maintainers can easily diagnose.

gattjoe:
edit: I have the same oMLX and macOS version as OP

I turned on debug logs and this is all it threw:

> 2026-05-28 20:16:37,245 - omlx.server - WARNING - [-] - POST /v1/messages → 422: [{'type': 'literal_error', 'loc': ('body', 'messages', 1, 'role'), 'msg': "Input should be 'user' or 'assistant'", 'input': 'system', 'ctx': {'expected': "'user' or 'assistant'"}}]

leoli-dev:
same here

- claude code: `v2.1.156`
- omlx: `v2.1.156`

leoli-dev:
Deepseek helped me to create a simple proxy script that works 

`omlx_proxy.py`:

```python
#!/usr/bin/env python3
"""oMLX Claude API proxy — fix role="system" in messages[] for Claude Code."""
import http.server, json, sys, urllib.request

LISTEN = ("127.0.0.1", 8081)
TARGET = "http://127.0.0.1:8000"

def fix(body):
    msgs = body.get("messages", [])
    if not msgs:
        return body
    sys_parts, kept = [], []
    for m in msgs:
        if m.get("role") == "system":
            c = m.get("content")
            if isinstance(c, str):
                sys_parts.append({"type":"text","text":c})
            elif isinstance(c, list):
                sys_parts.extend(c)
        else:
            kept.append(m)
    body["messages"] = kept
    if sys_parts:
        old = body.get("system")
        merged = []
        if isinstance(old, str): merged.append({"type":"text","text":old})
[...]

PR #1506, opened 2026-05-29
fix(anthropic_models): relax role field and fix line-length formatting

gattjoe:
## Summary

- Relax `role` field in `AnthropicMessage` from `Literal["user", "assistant"]` to `str` to accommodate roles beyond the two hardcoded values (e.g. tool results or future role types)
- Reformat long `source` type annotations in `ContentBlockImage`, `ContentBlockDocument`, and `stop_reason` in `MessagesResponse` to fit within line-length limits

## Test plan

- [ ] Verify existing Anthropic message parsing still works for `user` and `assistant` roles
- [ ] Confirm messages with non-standard roles no longer raise validation errors

### DIFF
--- a/omlx/api/anthropic_models.py
+++ b/omlx/api/anthropic_models.py
@@ -30,7 +30,9 @@ class ContentBlockImage(BaseModel):
     """Image content block with source data."""
 
     type: Literal["image"] = "image"
-    source: dict[str, Any]  # {"type": "base64"|"url", "media_type": "...", "data"|"url": "..."}
+    source: dict[
+        str, Any
+    ]  # {"type": "base64"|"url", "media_type": "...", "data"|"url": "..."}
 
 
 class ContentBlockToolUse(BaseModel):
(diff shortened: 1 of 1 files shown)


gattjoe:
fixes https://github.com/jundot/omlx/issues/1503

jundot:
Hey, thanks for jumping on this so quickly. The debug log you posted on #1503 was what gave me the exact spot in the body where claude-code was placing the system role, really helpful.

I ended up going with #1511 in the end because:

- It landed at roughly the same time with the minimal one-line diff, and
- I wanted to keep `role` as a `Literal` whitelist rather than open it up to arbitrary strings. The downstream code (chat template, `_merge_consecutive_roles`) only handles a known set of roles explicitly, so a stray role would silently produce odd behavior under `str` typing, whereas a `Literal` keeps Pydantic errors actionable for the next schema drift.

The black reflow on `ContentBlockImage` / `ContentBlockDocument` / `MessagesResponse` is a fair separate cleanup. If you'd like to send those as their own PR I'd merge it gladly.

I'm also pushing a follow-up commit on main right after this to extract in-messages system content into the canonical `system` field, so the user-visible behavior should be solid once that lands.

Thanks again, and hope to see more from you.

PR #1507, opened 2026-05-29
fix(mtp): per-key norm shift for mixed-convention Qwen3.6 MXFP4 checkpoints

scubamount:
## Problem

Native MTP speculative decoding yields **0% draft acceptance** (no speedup, MTP effectively disabled) on certain pre-quantized Qwen3.6 MXFP4 checkpoints — specifically bundles whose MTP-head RMSNorm weights are stored in a **mixed** convention. The backbone and `mtp.norm` are already in MLX's `+1` RMSNorm convention, while the per-layer MTP head norms remain in raw-HuggingFace convention.

Observed on `Qwen3.6-27B-MXFP4-MTP`:

```
MTP[0] finish=length tokens=150 cycles=148 accept=0/148 (0.0%)
  emits[init=2,draft=0,bonus=0,verify=148]
```

Throughput stayed at the no-MTP baseline (~11.5 tok/s) despite `mtp_enabled=true`.

## Root cause

`sanitize()` (in both `omlx/patches/mlx_lm_mtp/qwen35_model.py` and the mlx-vlm mirror) decides the `+1` norm shift with a single global flag derived solely from the **backbone**:

```python
should_shift_norm_weights = has_unsanitized_conv1d
...
if should_shift_norm_weights and any(k.endswith(sfx) for sfx in norm_keys):
    if v.ndim == 1:
        weights[k] = v + 1.0
```

When the backbone is already converted (conv1d already sanitized), the flag is `False`, so **no** norm is shifted. That is correct for the backbone — but this checkpoint ships its per-layer MTP head norms in raw-HF convention (mean ~ 0), so they are left unshifted and each head RMSNorm multiplies activations by ~0. The MTP head emits near-flat logits, the verifier rejects every draft, and acceptance is 0%.

Measured pre-shift means for this model's MTP norms (head, layer 0):

| MTP norm tensor | pre-shift mean | convention |
|---|---:|---|
| `mtp.norm.weight` | +1.2740 | MLX (+1) |
| `mtp.layers.0.self_attn.q_norm.weight` | +0.7550 | MLX (+1) |
| `mtp.layers.0.self_attn.k_norm.weight` | +0.7427 | MLX (+1) |
| `mtp.layers.0.input_layernorm.weight` | +0.0406 | raw-HF |
| `mtp.layers.0.post_attention_layernorm.weight` | +0.2108 | raw-HF |
| `mtp.pre_fc_norm_embedding.weight` | -0.4400 | raw-HF |
| `mtp.pre_fc_norm_hidden.weight` | -0.1711 | raw-HF |

The conventions are **mixed within the same head**, so neither a global "shift all" nor "shift none" is correct.

## Fix

Decide the `+1` shift **per-key** for MTP norms, from each weight's own magnitude (raw-HF center ~0, MLX-shifted ~1; threshold 0.5). Backbone norms keep the existing conv1d-derived behavior, so there is **no change for already-correct checkpoints**.

```python
def _mtp_norm_is_raw_hf(_w):
    try:
        return float(mx.mean(_w.astype(mx.float32)).item()) < 0.5
    except Exception:
        return False
...
if v.ndim == 1 and any(k.endswith(sfx) for sfx in norm_keys):
    if "mtp." in k:
        if _mtp_norm_is_raw_hf(v):
            weights[k] = v + 1.0
    elif should_shift_norm_weights:
        weights[k] = v + 1.0
```

Applied to both the mlx-lm and mlx-vlm sanitize paths (the vlm path remaps MTP keys to `language_model.mtp.*`, hence the substring test).

## Verification

Per-key decision against the real checkpoint:

```
input_layernorm           pre=+0.0406 raw_hf=True  -> +1 -> +1.0406  OK
post_attention_layernorm  pre=+0.2108 raw_hf=True  -> +1 -> +1.2108  OK
self_attn.k_norm          pre=+0.7427 raw_hf=False ->  0 -> +0.7427  OK
self_attn.q_norm          pre=+0.7550 raw_hf=False ->  0 -> +0.7550  OK
mtp.norm                  pre=+1.2740 raw_hf=False ->  0 -> +1.2740  OK
pre_fc_norm_embedding     pre=-0.4400 raw_hf=True  -> +1 -> +0.5600  OK
pre_fc_norm_hidden        pre=-0.1711 raw_hf=True  -> +1 -> +0.8289  OK
```

Live server, 3× 150-token requests, `Qwen3.6-27B-MXFP4-MTP`:

| Metric | Before | After |
|---|---|---|
| MTP draft accept | 0/148 (0.0%) | 73/75 (97.3%) |
| Throughput | ~11.5 tok/s | ~21.4 tok/s |
| Speedup | 1.0× | ~1.86× |

Stable across all three requests.

## Files changed

- `omlx/patches/mlx_lm_mtp/qwen35_model.py`
- `omlx/patches/mlx_vlm_mtp/qwen35_vlm_model.py`
- `docs/mtp_mixed_convention_norm_shift_fix.md` (writeup)

## Notes

- No behavior change for already-correct checkpoints (backbone follows conv1d; already-shifted MTP norms left untouched).
- The `< 0.5` threshold assumes MTP RMSNorm scales in MLX convention are not legitimately centered below 0.5; holds for the Qwen3.6 family (learned scales around unity). A future checkpoint that violates this would be better served by an explicit convention marker in metadata.

### DIFF
--- a/docs/mtp_mixed_convention_norm_shift_fix.md
+++ b/docs/mtp_mixed_convention_norm_shift_fix.md
@@ -1,0 +1,141 @@
+# MTP draft acceptance ~0% on mixed-convention Qwen3.6 MXFP4 checkpoints
+
+## Summary
+
+OMLX's native Multi-Token Prediction (MTP) speculative decoding produced **0%
+draft acceptance** (no speedup; MTP effectively disabled) on certain
+pre-quantized Qwen3.6 checkpoints — specifically MXFP4 bundles whose MTP-head
+RMSNorm weights are stored in a **mixed** convention. The backbone and the
+`mtp.norm` tensor were already in MLX's `+1` RMSNorm convention, while the
+per-layer MTP head norms were still in raw-HuggingFace convention.
+
+Affected model in this report: `Qwen3.6-27B-MXFP4-MTP` (JANGQ MXFP4 bundle).
+
+## Symptom
+
+Server MTP stats logged ~0% acceptance with all cycles falling back to full
+verification:
+
+```
+MTP[0] finish=length tokens=150 cycles=148 accept=0/148 (0.0%)
+  emits[init=2,draft=0,bonus=0,verify=148]
+  timing[backbone=11097.9ms mtp=709.5ms ...]
+```
+
+Throughput sat at the no-MTP baseline (~11.5 tok/s) despite `mtp_enabled=true`.
+
+## Root cause
+
+`TextModel.sanitize` (in `omlx/patches/mlx_lm_mtp/qwen35_model.py`, and the
+mirror in `omlx/patches/mlx_vlm_mtp/qwen35_vlm_model.py`) decides whether to
+shift RMSNorm weights by `+1` using a single global flag derived solely from
+the **backbone**:
+
+```python
+has_unsanitized_conv1d = any(
+    "conv1d.weight" in k and getattr(v, "shape", (1,))[-1] != 1
+    for k, v in weights.items()
+)
+should_shift_norm_weights = has_unsanitized_conv1d   # <-- global, backbone-only
+...
+if should_shift_norm_weights and any(k.endswith(sfx) for sfx in norm_keys):
+    if v.ndim == 1:
+        weights[k] = v + 1.0
+```
+
+For a checkpoint whose backbone is already converted (conv1d already sanitized),
+`should_shift_norm_weights` is `False`, so **no** norm — backbone or MTP head —
+is shifted. That is correct for the backbone. But this particular checkpoint
+ships its per-layer MTP head norms in raw-HF convention (mean ~ 0), so they are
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for the detailed writeup, the root cause is spot on. The single backbone-derived `should_shift_norm_weights` flag genuinely can't speak for the MTP head when the head norms ship in a different convention, and deciding the `+1` shift per-key from each weight's own magnitude is the right call here since there's no convention marker in the checkpoint to lean on. The before/after numbers (0% to 97.3% accept, ~1.86x) make the effect clear.

I'll merge this. One thing I'll fold into a follow-up: the same global-flag norm shift lives in two more mirror sites, `qwen35_moe_vlm_model.py` and `qwen35_moe_vlm_runtime.py`, so a mixed-convention MoE Qwen3.6 vision checkpoint would still hit the same 0% path. I'll extend the per-key fix there and add a sanitize regression test for the mixed case.

PR #1509, opened 2026-05-29
fix(menubar): show live server port after restart, not stale config snapshot

popfido:
## Summary

After changing the server port via Server screen's Apply and letting the server restart, the menubar status header kept showing the *original* port (`:8080`) instead of the new one (`:8964`). Same stale-snapshot bug also affected the port-in-use alert and the Chat menu item's URL.

## Root cause

`MenubarController.config` is captured by value at init (`private let config: AppConfig`). When `AppServices.applyServerEndpoint(port:)`:

1. mutates `AppServices.config.port`
2. calls `server.reconfigure(port:)` so the running `ServerProcess` knows the new port
3. restarts the process — `ServerProcess.stateDidChangeNotification` fires
4. menubar's observer calls `refreshMenuState()` → `headerDisplay(.running(pid))` → reads `config.port`

…step 4 reads from the captured snapshot, which never saw step 1. So `:8080` stuck.

## Fix

Source the displayed port (and host) from the live `ServerProcess` — which `reconfigure(host:port:)` keeps current — falling back to the captured config only when there is no server (bootstrap failed).

Two `static` helpers on `MenubarController` so they can be unit-tested without instantiating the full controller (which needs a live `NSStatusBar`):

```swift
static func displayPort(server: ServerProcess?, fallback: Int) -> Int {
    server?.port ?? fallback
}
static func displayHost(server: ServerProcess?, fallback: String) -> String {
    server?.host ?? fallback
}
```

Updated 4 user-facing sites:
- `.running(pid)` header (`menubar.header.running`)
- Port-in-use alert title (`menubar.alert.port_in_use.title`)
- Port-in-use alert body (`menubar.alert.port_in_use.other`)
- Chat menu URL

Listen Address picker already commits via `AppServices.applyServerEndpoint(host:)` → `server.reconfigure(host:)`, so the same fix also covers the parallel "menubar shows stale host" path.

## Test plan

New `MenubarControllerPortTests` (6 cases, all pass):

- [x] `displayPort(server: nil, fallback:)` returns the fallback
- [x] `displayPort(server: realServer, fallback:)` prefers the server's port
- [x] After `server.reconfigure(port:)`, `displayPort` follows — **the regression case**
- [x] Same three for `displayHost`

```
Executed 6 tests, with 0 failures (0 unexpected) in 0.004 seconds
** TEST SUCCEEDED **
```

Manual verification deferred to the reviewer's full Release stage: change port in Server tab → click Apply → server restarts → open menubar → header shows new port.

## Notes

- New test file's Xcode project slot is `100` (not `010`, which collides with the existing `oMLXTests.xctest` product reference — Xcode silently drops the dupe and the test class never reaches the bundle).
- Diff: 3 files, +135/−4 (the +135 is the test file's body).

### DIFF
--- a/apps/omlx-mac/Sources/Menubar/MenubarController.swift
+++ b/apps/omlx-mac/Sources/Menubar/MenubarController.swift
@@ -291,9 +291,10 @@ final class MenubarController: NSObject {
                 .systemBlue
             )
         case .running(let pid):
+            let port = MenubarController.displayPort(server: server, fallback: config.port)
             return (
                 String(localized: "menubar.header.running",
-                       defaultValue: "Server: running · pid \(String(pid)) · :\(String(config.port))",
+                       defaultValue: "Server: running · pid \(String(pid)) · :\(String(port))",
                        comment: "Menubar status header when the server is running; placeholders are PID and port (rendered as plain integers, no grouping)"),
                 .systemGreen
             )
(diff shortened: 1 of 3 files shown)


jundot:
Thanks, merged. The root-cause writeup is spot on, and sourcing the live endpoint from `ServerProcess` is the right call since it is the authoritative runtime state.

One sibling spot in the same stale-snapshot family: `startStatsPoller()` builds the poller from `config.baseURL`, so the menubar stats keep polling the old port after a restart. I will handle that in a follow-up.

PR #1511, opened 2026-05-29
fix(claude code 422):Update code for compatibility with new Claude version

lx1229:
## Summary

Add `"system"` role support to `AnthropicMessage` model.

## Changes

- Updated `omlx/api/anthropic_models.py` to include `"system"` as a valid `role` value

## Reason

Claude Code v2.1.154+ changed the message schema to include system role messages.

### DIFF
--- a/omlx/api/anthropic_models.py
+++ b/omlx/api/anthropic_models.py
@@ -102,7 +102,7 @@ class SystemContent(BaseModel):
 class AnthropicMessage(BaseModel):
     """A message in an Anthropic conversation."""
 
-    role: Literal["user", "assistant","system"]
+    role: Literal["user", "assistant", "system"]
     content: str | list[ContentBlock]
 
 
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for the quick fix, I'll merge this. Users on claude-code 2.1.154+ have been hitting this for a few days, your one-line patch unblocks them with the minimum possible change.

Following up on main with a commit that pulls in-messages system content into the canonical `system` field, so the chat template gets a single clean system prompt no matter where claude-code places the role="system" entry.

ericlsimplifi:
@jundot thanks all! Do you know when this is going out in a release?

lx1229:
> @jundot thanks all! Do you know when this is going out in a release?
https://github.com/jundot/omlx/releases/tag/v0.4.0rc1 new version has already here with fix!

PR #1512, opened 2026-05-29
feat(responses): wire reasoning.effort to enable_thinking

a4501150:
## Summary
- The Responses API accepts a `reasoning` dict but silently ignores it. This wires `reasoning.effort` to `enable_thinking` in the chat template.
- `{"effort": "high"}` (or any non-"none" value) enables thinking; `{"effort": "none"}` disables it; omitted/null leaves defaults unchanged.
- Builds on the native reasoning support from #1245.

## Test plan
- Unit tests added in `TestResponsesRequestReasoning` covering all effort values and edge cases
- Verify reasoning output appears in Responses API when `reasoning.effort` is set to "high"
- Verify thinking is suppressed when `reasoning.effort` is "none"

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -3932,6 +3932,12 @@ async def create_response(
         if ms.preserve_thinking is not None:
             merged_ct_kwargs["preserve_thinking"] = ms.preserve_thinking
 
+    # Wire request.reasoning → enable_thinking.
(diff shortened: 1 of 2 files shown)


a4501150:
Closing — carrying this feature locally and rebasing onto upstream main.

PR #1513, opened 2026-05-29
fix(oq): strip _vision suffix from VisionConfig.model_type for VLM loads

a4501150:
## Summary
- HuggingFace transformers appends `_vision` to vision sub-config `model_type` (e.g. `qwen3_5_moe_vision`), but mlx-vlm's `VisionModel.__init__` only accepts the base name.
- This broke both oQ sensitivity measurement (instant "produced no scores" failure) and sanitize-plan discovery (output kept HF-format keys, causing "2052 parameters not in model" on load).
- Two fix sites: `BaseModelConfig.from_dict` monkey-patch for VLM model loads, and `_build_model_sanitizer` normalization for the `_maybe_deserialize_config` bypass path.

## Test plan
- Run oQ quantization on a Qwen3.6 VLM model and verify sensitivity measurement completes
- Verify sanitized weights load without "parameters not in model" errors
- Confirm non-VLM quantization is unaffected

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -1609,6 +1609,12 @@ def _build_model_sanitizer(config: dict, text_only: bool = False):
             vision_config = model_config.vision_config
             if isinstance(vision_config, dict):
                 vision_config = model_module.VisionConfig.from_dict(vision_config)
+            if hasattr(vision_config, "model_type") and isinstance(
+                vision_config.model_type, str
+            ) and vision_config.model_type.endswith("_vision"):
+                vision_config.model_type = vision_config.model_type.removesuffix(
+                    "_vision"
+                )
             text_config = model_config.text_config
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the fix. I think the normalization is currently too broad: it strips `_vision` for every mlx-vlm VisionConfig, but models like gemma3n and minicpmv4_6 only allow `gemma3n_vision` / `minicpmv4_6_vision` in VisionModel.__init__, so this can break existing VLM loads once maybe_apply_pre_load_patches() installs the global BaseModelConfig patch.

Could you scope the normalization to the Qwen model types affected here and add a regression test that a non-Qwen `_vision` model_type remains unchanged?

a4501150:
Closing — carrying this fix locally and rebasing onto upstream main.

Issue #1492, opened 2026-05-28
macOS app can persist ssd_cache_max_size=0GB, causing server startup crash


oyeqa:
## Summary

The macOS app can persist an invalid SSD cache size (`ssd_cache_max_size: "0GB"`) into `~/.omlx/settings.json`. On the next startup, the server exits immediately with a configuration validation error, and the menu bar app enters the auto-restart loop until it gives up.

I am only including the relevant cache setting below; auth keys and unrelated settings are intentionally omitted.

## Symptoms

Menu bar app shows/reports:

```text
Server exited with code 1 during startup
Auto-restart failed after 3 attempts
```

`~/Library/Application Support/oMLX/logs/server.log` showed:

```text
oMLX - LLM inference, optimized for your Mac
Version: 0.3.12
Build: 260527100809-macos26-tahoe

Saved CLI arguments to settings.json
Log directory: ~/.omlx/logs
Configuration error: ssd_cache_max_size must be positive
```

The persisted setting was:

```json
{
  "cache": {
    "enabled": true,
    "hot_cache_only": false,
    "ssd_cache_dir": "~/.omlx/cache",
    "ssd_cache_max_size": "0GB",
    "hot_cache_max_size": "8GB",
    "initial_cache_blocks": 256
  }
}
```

## Workaround

Manually changing the value to `auto` fixed startup:

```json
{
  "cache": {
    "ssd_cache_max_size": "auto"
  }
}
```

After restarting the macOS app, the server started normally:

```text
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:8000
```

`GET /health` returned HTTP 200.

## Expected behavior

Either:

- the macOS app/admin UI should not persist `0GB` for `ssd_cache_max_size`, or
- `0GB` should be interpreted as disabling the SSD cache, or
- startup should recover with a clearer automatic fallback instead of entering the restart loop.

At minimum, the UI should validate/reject the value before saving it.

## Environment

- oMLX version: 0.3.12
- Build: 260527100809-macos26-tahoe
- Install type: macOS app
- Server port: 8000
- Model directory: `~/.lmstudio/models`

Thanks for oMLX. This was easy to work around once the underlying config validation error was visible, but the menu bar error only surfaced the generic restart failure.

chenqianhe:
[...]

PR #1514, opened 2026-05-29
fix: cache size validation and SSD cache auto settings

chenqianhe:
Fixes invalid cache size handling around SSD cold cache and hot RAM cache configuration.

This PR centralizes cache size validation so all entry points share the same rules:

- SSD cold cache accepts `auto` or a positive size such as `1GB`, `20GB`, `500GB`.
- SSD cold cache rejects `0`, because `0` is not a valid SSD cache size.
- Hot RAM cache accepts `0` as disabled, or a positive size such as `8GB`.
- Hot RAM cache rejects new `auto` values, because RAM auto-sizing is ambiguous and can conflict with runtime memory pressure.

It also fixes admin UI behavior so small positive SSD cache sizes are not inflated by the slider, and persisted `auto` settings are not silently converted into concrete GB values during unrelated saves.

## Problem

Issue #1492 exposed that `ssd_cache_max_size` could be saved as `0GB`.

That value is unsafe because the SSD cold cache runtime expects either:

- `auto`, which resolves to 10% of SSD capacity, or
- a positive size.

Unlike hot RAM cache, SSD cold cache does not use `0` as a disabled sentinel. Allowing `0GB` through the settings UI could persist a configuration that later caused startup/runtime failures.

There were also related consistency issues:

- Cache size rules were spread across settings, CLI, admin routes, legacy config helpers, web UI, and macOS UI.
- Admin save logic could mutate live settings or runtime state before persistence succeeded.
- The admin GET endpoint expanded `auto` into a concrete size, so opening the settings page and saving unrelated fields could silently persist `auto` as `NGB`.
- The web slider used integer percentages, which made small valid SSD cache sizes appear as `1%` and risked rewriting them to much larger values on large disks.

### Centralized cache size validation

Added shared normalizers in `omlx/settings.py`:

- `normalize_ssd_cache_max_size`
- `normalize_hot_cache_max_size`

These functions are now the single source of truth for cache size semantics.

SSD cache behavior:

- `auto` is valid.
- Positive sizes are valid.
- `0`, `0GB`, negative sizes, invalid strings, and non-finite values are rejected.
- Legacy persisted `0GB` is recovered to `auto` during settings load.

Hot RAM cache behavior:

- `0` and `0GB` are valid and mean disabled.
- Positive sizes are valid.
- New `auto` values are rejected.
- Legacy persisted `auto` is recovered to `0` during settings load.

### Safer admin settings save flow

Updated `/admin/api/global-settings` so validation happens before live mutation and runtime application.

The save path now:

1. Normalizes incoming values.
2. Applies changes to a copied candidate settings object.
3. Runs global validation.
4. Mutates live settings only after preflight succeeds.
5. Saves to disk.
6. Applies runtime changes only after save succeeds.
7. Restores the live snapshot if persistence fails.

This prevents partial state drift where runtime settings change even though saving failed.

### Preserve SSD `auto` in admin API

The admin settings GET response now returns:

- `ssd_cache_max_size`: the raw persisted value, such as `auto`
- `ssd_cache_max_size_resolved`: the computed display value, such as `10GB`

This prevents unrelated admin saves from converting `auto` into a fixed GB value.

### Web dashboard slider fixes

Updated the SSD cache slider to support small positive cache sizes:

- Slider minimum changed from `1%` to `0.01%`.
- Slider step changed from `1` to `0.01`.
- Display parsing now preserves two decimal places.
- Loading settings no longer rewrites `ssd_cache_max_size`.
- Slider-generated values still clamp to at least `1GB`, so the UI cannot write `0GB`.

Manual GB input remains minimum `1GB`, because it edits the actual SSD cache size string.

### CLI and legacy config alignment

Updated CLI one-shot cache size resolution to reuse the shared settings validators.

Updated deprecated `PagedSSDCacheConfig` helpers so they also use the same validation semantics.

Also hardened `parse_size()` against non-finite and overflow values.

### macOS UI alignment

Updated the macOS performance settings UI so hot RAM cache no longer advertises `auto`.

Hot RAM cache empty input now normalizes to `0`, while SSD cache empty input normalizes to `auto`.

### DIFF
--- a/apps/omlx-mac/Resources/Localizable.xcstrings
+++ b/apps/omlx-mac/Resources/Localizable.xcstrings
@@ -3505,7 +3505,7 @@
         "en" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "RAM ceiling for hot cache. \"0\" disables, \"8GB\" or \"auto\" accepted."
+            "value" : "RAM ceiling for hot cache. \"0\" disables; positive sizes like \"8GB\" enable it."
           }
         }
       }
--- a/apps/omlx-mac/Sources/AppView/Screens/PerformanceScreen.swift
+++ b/apps/omlx-mac/Sources/AppView/Screens/PerformanceScreen.swift
@@ -243,14 +243,12 @@ private struct CacheSection: View {
                               defaultValue: "Hot Cache Size",
                               comment: "Row label for the hot cache size field"),
                 sublabel: String(localized: "performance.cache.hot_size.sub",
-                                 defaultValue: "RAM ceiling for hot cache. \"0\" disables, \"8GB\" or \"auto\" accepted.",
+                                 defaultValue: "RAM ceiling for hot cache. \"0\" disables; positive sizes like \"8GB\" enable it.",
                                  comment: "Sublabel describing accepted hot cache size values")
             ) {
                 TextInput(
                     text: $vm.hotCacheMaxSize,
-                    placeholder: String(localized: "performance.memory.placeholder_auto",
-                                        defaultValue: "auto",
-                                        comment: "Memory field placeholder meaning automatic"),
+                    placeholder: "0",
                     mono: true,
                     width: 140
                 )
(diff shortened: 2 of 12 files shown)


chenqianhe:
## Why doesn’t hot cache have `auto`? Mainly because it’s semantically unsafe:

For SSD, `auto` is clear: use 10% of SSD capacity. Disk capacity is large, and the reserved-space logic is straightforward.

For RAM, `auto` is ambiguous: should it mean 10% of total memory, 10% of available memory, the remaining memory after the memory guard, or be calculated dynamically based on model size?

oMLX already has many variables: model memory, Metal/iogpu wired memory, scheduler concurrency, active KV, and system memory pressure. If hot cache also automatically consumes RAM, it can easily compete with the memory guard for the same budget.

Hot cache is an optimization, not a requirement. The most conservative default semantics should be: `0 = disabled`; users explicitly enable it by setting values like `8GB` or `16GB`.

PR #1516, opened 2026-05-29
followup(chat) - svg rendering, fix of various bugs in chat-v2

beamivalice:
This is the follow-up patch to the #1341 with one highlight.

* Correct the stale model list when downloaded new models
* Fix dedupe model list
* Fix thinking not cleared in follow up chat
* Fix timeline dots move with variants

Now you can render svg directly when asking the model to reply one.

Screenshot 2569-05-29 at 16 00 23

### DIFF
--- a/omlx/admin/i18n/en.json
+++ b/omlx/admin/i18n/en.json
@@ -691,6 +691,8 @@
   "chat.error.invalid_image_type": "Please select an image file",
   "chat.error.image_too_large": "Image size must be less than 10MB",
   "chat.error.image_load_failed": "Failed to load image. Please try again.",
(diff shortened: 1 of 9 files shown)


beamivalice:
* Add gate control to Allow SVG rendering (with dom already purified for most part)

Screenshot 2569-05-29 at 16 28 05

jundot:
Thanks for the follow-up. The svg render gate looks solid to me - off by default, clear warning, and DOMPurify with a tight tag/attr allowlist (no script, event handlers, foreignObject or use), so it stays safe even when a user opts in. The model dedupe + visibility refresh, the thinking reset, and the timeline dot fix all look right. I'll merge this.

PR #1517, opened 2026-05-29
feat(downloads): model card sheet with metadata, files, tags

popfido:
## Summary

Adds the model-card slide-over to the Swift Downloads screen — closes the parity gap with the HTML admin panel's model-card modal. The server endpoints (`/admin/api/hf/model-info` and `/admin/api/ms/model-info`) were already in place; this PR is entirely client-side.

## What you see

**Where the info button lives**

| Surface | Affordance |
|---|---|
| Completed Tasks rows | `info.circle` button between primary actions |
| Suggested Models rows | `info.circle` button before the Get button |
| HF / MS search dropdown rows | Hover-revealed `info.circle` at the trailing edge — replaces the downloads count while hovered so the icon has a clean slot and the row's `onPick` tap is unaffected |

**Sheet layout** (720×600)

- Header: model name + source badge ("Hugging Face" / "ModelScope")
- Metadata strip: params / size / pipeline_tag pills + downloads (↓) and likes (♥) counters in compact format (`1.2K`, `3.4M`)
- LoRA warning banner when `is_adapter == true` (HF only) — Download button is hidden in that case so users don't pull an adapter they can't run standalone
- Tab bar:
  - HF: `Model Card | Files | Tags`
  - MS: `Model Card | Tags` (Files hidden, matching the HTML admin)
- Tab body:
  - **Model Card** — Markdown rendered via `swift-markdown-ui`'s `.docC` theme (full block support — headings, code blocks, tables, images). Empty-string `model_card` shows a friendly "no README" state
  - **Files** — scrollable list with file name + server-formatted size
  - **Tags** — wrap-flow of pill chips (reuses the existing `FlowLayout` from `ProfileViews.swift`)
- Footer: "View on Hugging Face/ModelScope" external link + primary Download button (auto-dismisses sheet, kicks the download to the existing source-routed `startDownload` path)
- Loading / error / empty states with localized copy + Retry button on errors

## Why these UX choices

- **Info button (not click-on-name)**: macOS HIG canonical "more info" affordance, doesn't conflict with the row's primary "Download/Get" action, discoverable for a new feature. See discussion in the [audit thread](https://github.com/popfido/omlx/pull/1505) for the transactional-unit framing.
- **Hover-reveal for search dropdown**: the search-result row is already a single full-row `Button { onPick(m) }`; always-visible info icon would crowd and conflict. Finder-style hover swaps the count for the icon — same trailing slot, no overlap.
- **`swift-markdown-ui` over native `AttributedString(markdown:)`**: mlx-community READMEs typically have tables, code blocks, and heading hierarchy — native flattens all of those to plain text. The library is pure-SwiftUI, MIT, ~10K stars, actively maintained.
- **LoRA warning + hide Download**: prevents the most common "I downloaded the adapter and it doesn't load" support thread.

## Network

- `ModelCardDTO` decodes both endpoints (same envelope, optional fields tolerate HF/MS shape differences without per-source DTOs)
- `OMLXClient.getHFModelCard(repoId:)` / `getMSModelCard(modelId:)` use the existing GET helper

## Dependencies

First third-party SPM dep in the project. `project.pbxproj` gains the expected `XCRemoteSwiftPackageReference` + `XCSwiftPackageProductDependency` scaffolding and the framework is linked into oMLX's Frameworks build phase. `Package.resolved` stays ignored per the project's existing `.gitignore` convention; `minimumVersion = 2.4.0` + `upToNextMajorVersion` covers reproducibility.

| Package | Version | Source | Purpose |
|---|---|---|---|
| `swift-markdown-ui` | 2.4.1 (any 2.x) | github.com/gonzalezreal | Markdown rendering |
| `NetworkImage` | 6.0.1 | transitive | Image rendering in markdown |
| `swift-cmark` | 0.8.0 | transitive (from swiftlang) | Markdown parser |

## Test plan

`xcodebuild test` green — 132 tests, 1 integration skipped, 0 failures.

`ModelCardDTODecodeTests` (9 cases):
- [x] markdown body decoded
- [x] empty `model_card` is legal (no-README signal, not an error)
- [x] missing `model_card` field throws (server-side regression guard)
- [x] unknown fields ignored (forward-compat with new server fields)
- [x] full HF metadata (params/size/pipeline_tag/downloads/likes/is_adapter) decodes
- [x] MS shape (no is_adapter, null params) decodes without throwing
- [x] LoRA-adapter flag decodes in isolation
- [x] files list decodes (name/size/size_formatted including empty size_formatted)
- [x] tags list decodes

Manual (verified locally against a live server PATCH path):
- [x] Info button visible on Completed + Suggested rows
- [x] Info button hover-reveals on HF + MS search dropdown rows; clicking it doesn't fire `onPick`
- [x] Sheet renders Markdown with headings / code / tables / lists
- [x] Metadata badges render for HF; counters compact-format
- [x] Tabs switch correctly; tab selection resets to `Model Card` between repos
- [x] Files tab shows file list with sizes (HF)
- [x] Tags tab shows wrap-flow pills
- [x] LoRA banner appears for adapter repos; Download button hides
- [x] "View on Hugging Face/ModelScope" opens the canonical upstream URL
- [x] Download button kicks a fresh task into Active Downloads + dismisses the sheet
- [x] Close: ⎋ key + X button
- [x] Source switch (HF↔MS) re-fires the fetch via the sheet's `.task(id:)` keyed on `source+repoId`

## Notes

- The Tags tab uses the existing `FlowLayout` from `ProfileViews.swift` rather than introducing a duplicate; first attempt collided on the type name and was caught at build time.
- The HF Files / MS Files contrast matches the HTML admin's choice (MS file lists are usually sharded blobs that don't help the user decide).
- Future: a `Description` field is on the wire for both endpoints but not surfaced anywhere visible — could land as a one-line subtitle in the header in a follow-up if useful.

### DIFF
--- a/apps/omlx-mac/Resources/Localizable.xcstrings
+++ b/apps/omlx-mac/Resources/Localizable.xcstrings
@@ -1695,6 +1695,18 @@
         }
       }
     },
+    "downloads.button.show_card" : {
+      "comment" : "Tooltip on the info button that opens a model's README sheet",
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "View model card"
+          }
+        }
+      }
+    },
     "downloads.cancel.help" : {
       "comment" : "Tooltip on the X button that cancels or removes a download task",
       "extractionState" : "manual",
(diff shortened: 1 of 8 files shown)


jundot:
Thanks for this, it's really clean. The parity with the HTML admin's model-card modal is spot on, and I like the small calls: hover-reveal on the search rows so the info action doesn't fight the pick tap, and hiding Download for LoRA adapters. DTO, client wiring, and the decode tests all look solid, nothing blocking on my side. I'll fold it into the next release. Thanks for the thoughtful write-up too.

PR #1518, opened 2026-05-29
fix(model_discovery): classify LFM text MoE as LLM, not mlx-audio STS

samfenwick:
## Summary

- The `lfm*` `model_type` prefix fallback was meant for mlx-audio STS checkpoints, but it also matched Liquid LFM **text** exports (`lfm2_moe`, `lfm2_5_moe`, etc. with `*ForCausalLM`).
- Text LFM models are now classified as `llm` when architectures include CausalLM.
- Real LFM audio stays on existing paths: `LFM2AudioModel` (architecture) and `lfm_audio` (`AUDIO_STS_MODEL_TYPES`).

## Test plan

- [x] `pytest tests/test_model_discovery.py -k lfm` (5 cases: text MoE variants, audio arch, `lfm_audio` model_type-only, unknown mlx-audio prefix)
- [x] Full `tests/test_model_discovery.py` (107 passed)

### DIFF
--- a/omlx/model_discovery.py
+++ b/omlx/model_discovery.py
@@ -395,6 +395,16 @@ def _has_vision_subconfig(config: dict) -> bool:
     )
 
 
+def _architecture_indicates_causal_lm(architectures: list[str]) -> bool:
+    """True when ``architectures`` describe a text causal LM (not mlx-audio STS).
+
+    Liquid LFM text checkpoints (LFM2, LFM2.5 MoE, etc.) use ``lfm*`` model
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for catching this. The `lfm*` prefix fallback was too broad and swallowed the LFM2/LFM2.5 text MoE exports into the STS path. The CausalLM-architecture disambiguation is the right call, and real LFM audio still short-circuits earlier via `LFM2AudioModel` / `lfm_audio`, so there's no regression. Tests cover both sides. I'll merge this.

PR #1521, opened 2026-05-29
fix(server): gate Anthropic usage partition on cache_control, not engine prefix cache

### LINKED ISSUE #1487: [v0.3.12] Anthropic `/v1/messages`: `input_tokens` double-counts cache (= cr + cw instead of new tokens only)
**Describe the bug**

In `v0.3.12`, the Anthropic `/v1/messages` endpoint reports `input_tokens` as `cache_read_input_tokens + cache_creation_input_tokens` when the request uses `cache_control`. This violates the Anthropic spec, which requires the three input fields to be a **disjoint partition** of the prompt — `input_tokens` should count only the **new, uncached, non-cache-creation** tokens.

Specifically, every successful Claude Code / Claude Agent SDK turn now reports `input_tokens == cr + cw`, double-counting the cached prefix and inflating any downstream `total_tokens` math by exactly `cr + cw`.

This is a regression in the field semantics added by [#912](https://github.com/jundot/omlx/issues/912) (v0.3.8). The fields exist now, but their values don't follow Anthropic's contract.

**To Reproduce**

Two minimal curl examples against a running oMLX v0.3.12 server.

**1. With `cache_control` (Claude Code / Agent SDK style) — bug triggers:**

```bash
PROMPT=$(python3 -c "print('Lorem ipsum dolor sit amet consectetur adipiscing elit ' * 200)")
curl -s -X POST http://127.0.0.1:8000/v1/messages \
  -H "content-type: application/json" \
  -H "x-api-key: $KEY" \
  -H "anthropic-version: 2023-06-01" \
  -d "{
    \"model\":\"Qwen3-Coder-Next-MLX-8bit\",\"max_tokens\":4,
    \"system\":[{\"type\":\"text\",\"text\":\"$PROMPT\",\"cache_control\":{\"type\":\"ephemeral\"}}],
    \"messages\":[{\"role\":\"user\",\"content\":\"hi\"}]
  }"
```

Response (observed):
```json
{
  "usage": {
    "input_tokens": 1615,
    "output_tokens": 4,
    "cache_creation_input_tokens": 1615,
    "cache_read_input_tokens": 0
  }
}
```

Here `input_tokens (1615) == cache_creation_input_tokens (1615) + cache_read_input_tokens (0)`. Per Anthropic spec, `input_tokens` should be `0` (or the count of any genuinely new uncached prompt tokens — none in this request).

Same pattern from real production traffic (one Claude Agent SDK assistant message taken straight from `~/.claude/projects/.../{sid}.jsonl`):
```
i = 65,619   cr = 63,488   cw = 2,131   →   cr + cw = 65,619 = i   (every turn)
i = 67,877   cr = 65,536   cw = 2,341   →   cr + cw = 67,877 = i
i = 80,750   cr = 71,680   cw = 9,070   →   cr + cw = 80,750 = i
```

**2. Without `cache_control` — different (also wrong) shape:**

```bash
curl -s -X POST http://127.0.0.1:8000/v1/messages \
  -H "content-type: application/json" -H "x-api-key: $KEY" -H "anthropic-version: 2023-06-01" \
  -d '{"model":"Qwen3-Coder-Next-MLX-8bit","max_tokens":4,
       "messages":[{"role":"user","content":"<long prompt>"}]}'
```

Response:
```json
{"usage":{"input_tokens": 0, "cache_creation_input_tokens": 1011, "cache_read_input_tokens": 0, ...}}
```

No `cache_control` was sent, yet `cache_creation_input_tokens=1011`. The entire prompt is reported as "cache creation" with `input_tokens=0`. Anthropic only reports `cache_creation_input_tokens > 0` when the request actually includes `cache_control` breakpoints. (Less critical than #1, but suggests the cache accounting logic isn't gated on the request shape.)

**Expected behavior**

Per the [Anthropic prompt caching spec](https://platform.claude.com/docs/en/build-with-claude/prompt-caching), the three input-side fields form a **disjoint partition**:

```
prompt_tokens_sent  =  input_tokens  +  cache_read_input_tokens  +  cache_creation_input_tokens
```

- `input_tokens` — prompt tokens that were neither cached nor written to cache this request.
- `cache_creation_input_tokens` — prompt tokens written to a new cache entry this request (only when the request has `cache_control` breakpoints, and only the portion above the breakpoint).
- `cache_read_input_tokens` — prompt tokens that hit an existing cache entry.

For the example #1 above, the expected response is:
```json
{"input_tokens": 0, "cache_creation_input_tokens": 1615, "cache_read_input_tokens": 0, ...}
```

(`0` because nothing in the prompt fell outside the cache_control block.)

A repeated request with the same `cache_control`-marked system prompt should then return:
```json
{"input_tokens": 0, "cache_creation_input_tokens": 0, "cache_read_input_tokens": 1615, ...}
```

— signaling that the SSD-paged prefix actually hit.

**Desktop:**
- macOS Version: 26.4
- oMLX Version: 0.3.12 (`oMLX-0.3.12-macos26-tahoe.dmg`)
- Hardware: Apple Silicon (arm64)
- Model: `Qwen3-Coder-Next-MLX-8bit`

**Additional context**

Why this matters in practice: downstream agent harnesses that use prompt caching (Claude Code, Claude Agent SDK, and our own wrapper [captain-nemo](https://github.com/verneagent/nemo)) compute per-turn footprints from these fields. With `input_tokens = cr + cw`, any `total = input + cache_read + cache_creation + output` formula over-reports by exactly `cr + cw` per turn — so a 100K-token conversation looks like 200K, and per-turn cost / context-fullness displays are roughly 2×.

Releases [v0.3.8](https://github.com/jundot/omlx/releases/tag/v0.3.8) added the cache fields, and [v0.3.9.dev1](https://github.com/jundot/omlx/releases/tag/v0.3.9.dev1) fixed the OpenAI-Responses-API counterpart (`input_tokens_details.cached_tokens` was a dummy). The Anthropic-protocol values appear not to have gotten the same correction pass.

richgoodson:
## Summary

Fixes #1487. The Anthropic `/v1/messages` partition logic was keyed on the engine-level `engine.prefix_cache_enabled` flag, but per Anthropic's spec the disjoint triple

```
input_tokens + cache_creation_input_tokens + cache_read_input_tokens == prompt_tokens
```

only carries meaning when the **request** opted in via a `cache_control` breakpoint. Two symptoms fell out of the wrong gate:

1. **With `cache_control`**: the non-streaming response set `input_tokens` to the full prompt count *and* populated `cache_creation_input_tokens` with the same number, so downstream agent harnesses (Claude Code, Claude Agent SDK, third-party wrappers) computing `total = input + cw + cr` over-reported by exactly `cw + cr` per turn. A 100K-token conversation looked like 200K.
2. **Without `cache_control`**: the engine's automatic prefix-cache hits still leaked into the cache fields even though Anthropic only surfaces them when the client opted in.

This PR adds `request_has_cache_control(request)` — scans `system` / `tools` / message blocks for any `cache_control` field — and renames the partition gate on `convert_internal_to_anthropic_response()` and `create_message_delta_event()` from `prefix_cache_enabled` to `request_uses_cache_control`. The three server-side call sites (streaming `message_start`, streaming `message_delta`, non-streaming response builder) now pass the per-request signal instead of the engine flag.

### Behaviour after the fix

| Scenario | `input_tokens` | `cache_creation_input_tokens` | `cache_read_input_tokens` |
|---|---:|---:|---:|
| No `cache_control` | `prompt_tokens` | 0 | 0 |
| `cache_control`, cold | 0 | `prompt_tokens` | 0 |
| `cache_control`, warm | 0 | `prompt − hit` | `hit` |

### Before

```bash
curl -s -X POST http://127.0.0.1:8000/v1/messages \
  -d '{"model":"...","max_tokens":4,
       "system":[{"type":"text","text":"<1615 tokens>","cache_control":{"type":"ephemeral"}}],
       "messages":[{"role":"user","content":"hi"}]}'
```
```json
"usage": {
  "input_tokens": 1615,
  "cache_creation_input_tokens": 1615,
  "cache_read_input_tokens": 0,
  "output_tokens": 4
}
```
`input_tokens == cw + cr`, partition sums to 3230 (double-count).

### After

```json
"usage": {
  "input_tokens": 0,
  "cache_creation_input_tokens": 1615,
  "cache_read_input_tokens": 0,
  "output_tokens": 4
}
```
Partition sums to 1615. Identical to the spec example in #1487.

### Why is `input_tokens` always 0 when `cache_control` is present (and not, say, the tail of the prompt below the breakpoint)?

A strictly accurate partition would tokenize each block separately and split at the cache_control breakpoint position (e.g. system 1615 + user 1 → `cw=1615, input=1`). That requires per-block tokenization in the prompt builder. The issue reporter's "expected response" example uses the same approximation (`input=0` even though the prompt also has a one-token `"hi"`), so this PR ships the simple version and leaves per-block tokenization as a follow-up if anyone cares about that last token.

## Test plan

- [x] `pytest tests/test_api_utils.py tests/test_anthropic_adapter.py tests/test_anthropic_models.py tests/test_server.py tests/test_server_metrics.py tests/test_stream_usage.py` — **366 passed**.
- [x] Full unit suite: **4668 passed, 36 skipped, 59 deselected** (no regressions in non-Anthropic paths).
- [x] Live: `omlx serve --model-dir <gemma-4-e4b> --paged-ssd-cache-dir /tmp/ssd` against the curl examples from #1487. All three scenarios produce the spec-correct partition:

  | Test | Request | `input_tokens` | `cw` | `cr` | Sum |
  |---|---|---:|---:|---:|---:|
  | A | no cache_control | **10** | 0 | 0 | 10 |
  | B | cache_control, cold | **0** | **1615** | 0 | 1615 |
  | C | cache_control, warm | **0** | 79 | **1536** | 1615 |

  (The 79 in `cw` on the warm call is paged-cache block alignment — the prefix tail that doesn't fill a complete block — not a bug.)

### New / changed tests

- Rewrote `test_prefix_cache_disabled_legacy_shape` → `test_no_cache_control_legacy_shape` and `test_prefix_cache_enabled_splits_prompt` → `test_cache_control_cold_partitions_to_creation` + `test_cache_control_warm_partitions_to_read` against the spec contract.
- Added `test_usage_triple_is_disjoint_partition` — pins `input + cw + cr == prompt_tokens` across the `(uses_cc × cached_tokens)` matrix, including the `cached > prompt` clamp.
- Added `test_create_message_delta_event_no_cache_control_omits_cache_fields` — streaming SSE doesn't include the cache fields when the client didn't opt in.
- New `TestRequestHasCacheControl` class — 7 tests covering plain-string system, list system ±cc, tools ±cc, document content blocks ±cc, plain-string message content.

### DIFF
--- a/omlx/api/anthropic_utils.py
+++ b/omlx/api/anthropic_utils.py
@@ -28,6 +28,38 @@ _PRESERVE_ROLE_BOUNDARY = "_preserve_role_boundary"
 logger = logging.getLogger(__name__)
 
 
+def request_has_cache_control(request: MessagesRequest) -> bool:
+    """True if any system / tool / message block carries ``cache_control``.
+
+    Anthropic's three input-side usage fields (``input_tokens``,
+    ``cache_creation_input_tokens``, ``cache_read_input_tokens``) form a
+    *disjoint* partition of the prompt only when the client explicitly
+    marks a region with ``cache_control``. Without that signal the cache
+    fields must stay at 0 and ``input_tokens`` carries the whole prompt
+    count — independent of whether the oMLX engine happens to run
+    automatic prefix caching internally.
+    """
+    sys = request.system
+    if isinstance(sys, list):
+        for blk in sys:
+            if getattr(blk, "cache_control", None):
+                return True
+
+    for tool in request.tools or []:
+        if getattr(tool, "cache_control", None):
+            return True
+
+    for msg in request.messages:
+        content = msg.content
+        if not isinstance(content, list):
+            continue
+        for blk in content:
+            if getattr(blk, "cache_control", None):
+                return True
+
+    return False
+
+
 def _decode_document_block(block_dict: dict[str, Any]) -> str:
     """Decode an Anthropic document content block to text.
 
@@ -806,16 +838,19 @@ def convert_internal_to_anthropic_response(
     tool_calls: list[ToolCall] | None = None,
     thinking: str | None = None,
     cached_tokens: int = 0,
-    prefix_cache_enabled: bool = False,
+    request_uses_cache_control: bool = False,
 ) -> MessagesResponse:
     """
     Convert internal output to Anthropic MessagesResponse.
 
-    When ``prefix_cache_enabled`` is True the prompt count is split into
-    Anthropic's disjoint usage fields so that
-    input_tokens + cache_creation_input_tokens + cache_read_input_tokens
-    equals prompt_tokens. Otherwise the response keeps the legacy shape
-    (input_tokens = prompt_tokens, cache fields = 0).
+    When the request carries ``cache_control`` breakpoints (signalled by
+    ``request_uses_cache_control``) the prompt count is split into
+    Anthropic's disjoint usage triple so that
+    ``input_tokens + cache_creation_input_tokens + cache_read_input_tokens
+    == prompt_tokens``. Otherwise the response keeps the legacy shape
+    (``input_tokens = prompt_tokens``, both cache fields = 0) — even when
+    the engine's automatic prefix cache happened to hit, since Anthropic
+    only surfaces the cache triple when the client opted in.
 
     Args:
         text: Generated text content
@@ -826,7 +861,8 @@ def convert_internal_to_anthropic_response(
         tool_calls: List of internal ToolCall objects
         thinking: Reasoning/thinking content from <think> blocks
         cached_tokens: Prompt tokens served from the prefix cache
-        prefix_cache_enabled: Whether the engine runs automatic prefix caching
+        request_uses_cache_control: Whether the originating request carried
+            ``cache_control`` on any system / tool / message block.
 
     Returns:
         Anthropic MessagesResponse
@@ -877,9 +913,11 @@ def convert_internal_to_anthropic_response(
     # Map finish reason to stop reason
     stop_reason = map_finish_reason_to_stop_reason(finish_reason, bool(tool_calls))
 
-    # When prefix caching is on, split prompt_tokens into the Anthropic
-    # disjoint triple (input + creation + read == prompt_tokens).
-    if prefix_cache_enabled:
+    # Anthropic's three input-side fields are a disjoint partition of the
+    # prompt and only carry non-zero values when the request opted into
+    # caching via cache_control. Without that signal the cache fields stay
+    # at 0 regardless of any internal prefix-cache hits in the engine.
+    if request_uses_cache_control:
         cache_read = max(0, min(cached_tokens, prompt_tokens))
         cache_creation = prompt_tokens - cache_read
         input_display = 0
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for the thorough writeup and tests, this is exactly the right call. Gating the partition on the per-request `cache_control` opt-in instead of the engine flag matches Anthropic's contract, and the disjoint-partition matrix test pins the invariant nicely. I'll merge this. One small thing I'll fold into a follow-up: the message-block scan only catches `cache_control` on document blocks today (text and tool_result blocks drop the field on parse), so I'll extend those models so a breakpoint set purely on a message text block is detected too.

PR #1523, opened 2026-05-29
fix(scheduler): predictive prefill throttle + reclaim/requeue to prevent mid-stream OOM

### LINKED ISSUE #1522: Prefill OOM: long-context requests fail mid-stream with "Memory limit exceeded during prefill"
**Describe the bug**

When oMLX is used as the backend for a coding agent (e.g. OpenCode, Claude Code), a single long‑context prefill can spike GPU memory past the hard ceiling and the scheduler aborts the **entire request mid‑stream** with:

```
omlx.scheduler - WARNING - Prefill force-stopped at 1441 tokens: memory 42.8GB exceeds ceiling 42.0GB
omlx.scheduler - ERROR   - Prefill failed for <id>: Memory limit exceeded during prefill
omlx.server    - ERROR   - Error during chat streaming: Memory limit exceeded during prefill
```

The client receives a `finish_reason="error"` / SSE error event and the agent stops working. Notably, memory drops back to a low baseline immediately after the failure:

```
omlx.process_memory_enforcer - INFO - Memory pressure level: hard -> ok (current=24.6GB, soft=35.7GB, hard=39.9GB, ceiling=42.0GB)
```

i.e. the 42.8 GB was a **transient prefill activation spike**, not resident KV — the request is killed for a spike that is gone microseconds later.

**To Reproduce**

1. Serve a model whose attention uses `head_dim > 128` (MLX takes the float32 SDPA fallback that materializes the full `[B, n_q, query_len, kv_len]` attention matrix). MoE models are especially affected — measured transient was ~18 MB/token.
2. Drive it with a coding agent so the prompt/context grows large (tens of thousands of tokens), keeping the model resident near the ceiling.
3. Send a turn whose prefill chunk is large. The per‑chunk attention transient overshoots the hard ceiling and the request fails with `Memory limit exceeded during prefill`.

Observed on a 48 GB Mac with a ~23 GB model (ceiling ~42 GB from `iogpu.wired_limit_mb`), failing at only ~1.4k processed tokens.

**Expected behavior**

The server should keep memory bounded and the request should keep working — by sizing prefill chunks to fit the available headroom — rather than hard‑failing a request mid‑stream for a transient spike. If a turn genuinely cannot fit, it should fail gracefully/retriably rather than crashing the stream.

**Root cause (analysis)**

- `Scheduler._adaptive_chunk_size` only shrinks the chunk once **current** memory crosses the soft watermark. But a single large chunk's transient can blow the ceiling from a **low baseline** (current well below the watermark), so the throttle never engages before the overshoot.
- The per‑scheduler `PrefillTransientTracker.predict()` (an EWMA of measured bytes/token, whose docstring is literally "size the next chunk so its predicted transient stays under the remaining headroom") is **never called**.
- The static `_preflight_memory_check` is wired but under‑estimates for MoE / `head_dim>128` models (its SDPA term doesn't capture the real transient), so the request is admitted and then fails.
- On failure the request is terminated; there is no reclaim‑and‑retry. The `ProcessMemoryEnforcer` "all loaded models are pinned and no loads in progress" branch is a no‑op (it only logs), so nothing is reclaimed between turns.

 - macOS Version: 26.5
 - oMLX Version: 0.3.12

**Additional context**

A fix is proposed in an accompanying pull request: a predictive per‑chunk throttle (sizes each chunk so `current + predicted_transient(n) ≤ ceiling`, using the measured EWMA with a static first‑chunk fallback), a reclaim‑before‑fail step, a bounded requeue‑then‑retry fallback, and turning the dead‑end enforcer branch into an idle reclaim. Verified live: a 23,313‑token prompt that previously crashed now prefills and completes with no `OVER_HARD` / force‑stop.

sdiamanEXUS:
## Summary

Prevents the mid-stream prefill OOM where a long-context request hard-fails with `Memory limit exceeded during prefill` (see #1522). Adds a measurement-driven predictive prefill throttle plus graceful reclaim/requeue so memory stays bounded and the request keeps working instead of crashing.

Server-side only; no client or API changes. Below the pressure zone there is **no behavior change** (the full requested chunk runs unchanged when its predicted peak fits).

## Root cause

The failure is a **transient SDPA activation spike**, not resident KV (memory hit 42.8 GB at the check, then dropped to 24.6 GB immediately after teardown). For `head_dim > 128` (MLX float32 SDPA fallback) and especially MoE models the per-chunk transient is large (measured ~18 MB/token). The existing `_adaptive_chunk_size` only throttled once *current* memory crossed the soft watermark, so a single big chunk overshot the ceiling from a low baseline before any throttle engaged. The measured-EWMA predictor (`PrefillTransientTracker.predict()`) existed but was never wired in.

## Changes

- **Predictive per-chunk throttle** (`_adaptive_chunk_size`, `MemoryMonitor.estimate_chunk_transient_bytes`): size each chunk so `current + predicted_transient(n) ≤ ceiling * 0.90`. Uses the measured per-token EWMA once samples exist, else a static SDPA first-chunk estimate at the current context span (`kv_len` is now passed from both prefill loops, so a large prefix-cache hit with a small suffix is throttled correctly). The gate is on **predicted peak**, not on current crossing the watermark — that's the actual fix. The discrete watermark tiers are kept as a secondary clamp under deep pressure.
- **Reclaim before failing** (`_reclaim_prefill_headroom`): the in-loop hard check runs *before* the per-chunk `_sync_and_clear_cache`, so the just-computed transients are still resident; on an overshoot we now sync+clear and re-measure, and only raise if still over.
- **Bounded requeue-then-retry** (`_requeue_or_fail_prefill`, `Request.prefill_oom_retries`, `_MAX_PREFILL_OOM_RETRIES=2`): a memory-pressure prefill is requeued for a fresh, better-throttled attempt (reset to a cold full prefill so it can't re-trigger the same oversized span); after the budget is exhausted it emits the existing clean `finish_reason="error"`. Non-memory `RuntimeError`s still fail immediately. Wired into all three #1405 cleanup sites; restores stashed VLM mRoPE deltas and clears the SpecPrefill RoPE patch on requeue.
- **Enforcer idle reclaim** (`request_idle_reclaim` + `step()` drain): the previously no-op "all models pinned, no loads in progress" branch now requests a between-turn Metal reclaim, run on the inference thread only when idle (cross-thread-safe via the same deferred-flag idiom as aborts).

## Testing

- New unit tests in `tests/test_prefill_oom_graceful.py` (transient estimate math, predictive sizing incl. the low-baseline regression, requeue budget/gating).
- Updated 3 tests in `tests/test_scheduler_chunked_prefill.py` that asserted the old immediate-hard-fail behavior (memory errors now requeue; non-memory errors still surface).
- `pytest -m "not slow and not integration"` green.
- **Live**: on a real `head_dim=256` MoE model at the real ~42 GB ceiling, a 23,313-token prompt that previously crashed now prefills and completes (~17 s), with `[throttle:external] shrink 2048 -> …` firing from a low baseline and **no** `OVER_HARD` / force-stop / requeue needed.

## Risk / rollout

Low. The throttle is a no-op when a chunk already fits; reclaim only runs on the existing failure path; requeue is bounded (and can be set to `0` to fall straight back to today's behavior). No public API changes.

Fixes #1522

### DIFF
--- a/omlx/memory_monitor.py
+++ b/omlx/memory_monitor.py
@@ -428,6 +428,32 @@ class MemoryMonitor:
         kv = self.estimate_prompt_kv_bytes(new_tokens)
         return attn + kv
 
+    def estimate_chunk_transient_bytes(self, n_tokens: int, kv_len: int) -> int:
+        """Transient SDPA activation bytes for ONE prefill chunk.
+
+        Isolates the per-chunk attention transient — the spike that drives
+        prefill OOM — for a chunk of ``n_tokens`` query tokens attending over
+        ``kv_len`` total context tokens. Unlike ``estimate_prefill_peak_bytes``
+        this excludes newly-allocated KV (that becomes resident and is counted
+        in the caller's ``current`` baseline once eval'd); it is the quantity
+        the adaptive throttle must keep under the remaining headroom.
+
+        head_dim > 128: MLX SDPA fallback materializes the full attention
+        matrix [B, n_q, n_tokens, kv_len] in float32, plus a small output
+        buffer — so the transient scales with ``n_tokens * kv_len``.
+        head_dim <= 128: fused kernel is tiled / O(n), so only the output
+        buffer is allocated and throttling is effectively a no-op (correct).
+
+        Returns 0 when model info is unavailable.
+        """
+        hd = self._head_dim or 0
+        n_q = self._num_attention_heads or 0
+        if n_q == 0 or hd == 0 or n_tokens <= 0:
+            return 0
+        if hd > 128:
+            return n_q * n_tokens * max(kv_len, 0) * 4 + n_q * n_tokens * hd * 4
+        return n_q * n_tokens * hd * 4
+
     def estimate_blocks_to_free(self, bytes_to_free: int, block_size: int) -> int:
         """
         Estimate number of blocks to evict to free the given bytes.
--- a/omlx/process_memory_enforcer.py
+++ b/omlx/process_memory_enforcer.py
@@ -799,9 +799,26 @@ class ProcessMemoryEnforcer:
                             for e in self._engine_pool._entries.values()
                         )
                         if has_loaded:
+                            # Nothing to evict (all pinned) and no load to
+                            # abort — but the resident footprint may still hold
+                            # reclaimable Metal transients from a finished turn.
+                            # Ask each loaded scheduler to trim them between
+                            # turns. This only sets a flag; the actual reclaim
+                            # runs on the inference thread when it is idle, so
+                            # we never touch Metal from the enforcer thread.
+                            requested = 0
+                            for entry in self._engine_pool._entries.values():
+                                sched = self._resolve_scheduler(entry)
+                                if sched is not None and hasattr(
+                                    sched, "request_idle_reclaim"
+                                ):
+                                    sched.request_idle_reclaim()
+                                    requested += 1
                             logger.warning(
-                                "Hard memory pressure but all loaded models "
-                                "are pinned and no loads in progress."
+                                "Hard memory pressure, all loaded models "
+                                "pinned and no loads in progress: requested "
+                                "idle reclaim on %d scheduler(s).",
+                                requested,
                             )
                         else:
                             logger.warning(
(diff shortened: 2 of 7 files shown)


jundot:
Thanks for the thorough fix and the clear writeup. I went through it and the core idea holds up: gating on the predicted per-chunk peak instead of waiting for current memory to cross the watermark is the right fix, and using the stable physical cap (not the jittery dynamic ceiling) for the abort decision is a nice touch so a transient dip can't kill a near-complete prefill. I ran the new tests locally and they're all green. I'll merge this.

PR #1525, opened 2026-05-29
fix(server): advertise model-config context length instead of 32K global default

### LINKED ISSUE #1308: Server advertises 32K context window instead of model config context length
**Describe the bug**
The server advertises the global default 32K context window in its model metadata APIs, even when the model's `config.json` declares a larger context length.

This is narrower than #447. That issue covered general context-size autodetection / UI defaults. The problem here is specifically that API responses such as `/v1/models/status` and `/v1/models` do not expose the model metadata value, so clients and integrations still see 32K unless a per-model setting is manually configured.

**To Reproduce**
1. Run the server from the CLI, for example:
   ```bash
   uv run omlx serve
   ```
2. Use a local model whose `config.json` declares a larger context length, for example:
   ```json
   {
     "max_position_embeddings": 262144
   }
   ```
3. Query the models status endpoint:
   ```bash
   curl http://localhost:8000/v1/models/status
   ```
4. Observe that the model reports the global default instead of the model metadata value:
   ```json
   {
     "max_context_window": 32768
   }
   ```
5. Query the OpenAI-compatible models endpoint:
   ```bash
   curl http://localhost:8000/v1/models
   ```
6. Observe that the model's real context length is not exposed there either, so OpenAI-compatible clients cannot discover it from the model listing.

**Expected behavior**
The server should advertise the effective context window in model metadata responses using this priority:
1. Explicit per-model setting, if configured.
2. Context length discovered from model metadata, especially `config.json`.
3. Global default, only when no model-specific value is known.

For example, if `config.json` contains:

```json
{
  "max_position_embeddings": 262144
}
```

then `/v1/models/status` should report:

```json
{
  "max_context_window": 262144
}
```

For OpenAI-compatible `/v1/models`, a common extension used by vLLM is:

```json
{
  "max_model_len": 262144
}
```

**Screenshots**
Not applicable.

 - macOS Version: Tahoe 26.4
 - oMLX Version: 0.3.9rc1

**Additional context**
I am not using the GUI/dashboard flow. This was observed while running the server directly with `uv run omlx serve`.

Clients that rely on the models API for context limits configure themselves incorrectly. This affects long-context workflows and integrations that use the advertised limit to set prompt budgets, compaction windows, or request validation.

Related prior issue: #447 requested general autodetection of model context size and was closed as completed. This report is about the remaining narrower API exposure behavior: the server can still return the global 32768-token default in model metadata even when the model config declares a larger context length.

A possible fix is to preserve the context length during model discovery and use it when populating model metadata responses. Useful metadata keys include:

- `max_position_embeddings`
- `max_seq_len`
- `max_seq_length`
- `seq_length`
- `n_positions`

For VLM or wrapper configs, the same fields may be nested under keys such as `text_config` or `language_config`.

`tokenizer_config.json`'s `model_max_length` can be used as a fallback, but it should ignore sentinel values such as Transformers' `int(1e30)` "infinite" default.

richgoodson:
## Summary

Fixes #1308. `/v1/models` and `/v1/models/status` reported the global 32 K default for every model, even ones whose `config.json` declared 256 K or more. Clients that read the listing to size their context budgets (vLLM-compatible OpenAI consumers, monitoring dashboards) configured themselves at the wrong limit unless an operator manually set a per-model override.

### Resolution order (#1308)

```
1. Explicit per-model setting (admin UI / settings.json override).
2. EngineEntry.model_context_length — discovered from config.json at startup.
3. SamplingDefaults.max_context_window — last-resort global default.
```

The discovery-time helper `_read_model_context_length(model_path)` walks the standard set of config-length fields once at startup, so per-request resolution stays cheap:

```
config.json:
  max_position_embeddings, max_seq_len, max_seq_length,
  seq_length, n_positions
config.json -> text_config / language_config:   (VLM, MoE wrappers)
  same set of keys
tokenizer_config.json:
  model_max_length, with Transformers' int(1e30) sentinel rejected
```

The resolved length lives on `DiscoveredModel` and propagates to `EngineEntry.model_context_length`. `get_max_context_window()` reads it as the middle tier.

### `/v1/models` response shape

`ModelInfo` gains an optional `max_model_len` field (vLLM convention), letting OpenAI-style clients read the effective context window from the listing without a separate `/v1/models/status` call:

```json
{
  "object": "list",
  "data": [
    {
      "id": "gemma-4-e4b",
      "object": "model",
      "owned_by": "omlx",
      "max_model_len": 131072
    }
  ]
}
```

`/v1/models/status`'s existing `max_context_window` field picks up the discovery tier automatically since it already goes through `get_max_context_window()`.

## Test plan

- [x] `pytest tests/test_model_discovery.py tests/test_server.py tests/test_engine_pool.py tests/test_context_window.py` — **224 passed**.
- [x] Full unit suite: **4762 passed, 37 skipped, 59 deselected** (no regressions; one pre-existing test in `test_context_window.py` updated to null `state.engine_pool` so its focus remains on the per-model → global fallback path that pre-dates this PR).
- [x] Live: `omlx serve --model-dir <gemma-4-E4B-it-MLX-4bit>` — this model has `text_config.max_position_embeddings = 131072` plus a `tokenizer_config.json` `model_max_length = int(1e30)` sentinel (a nice double-check of the nested-config path *and* the sentinel filter).

  | Endpoint | Before | After |
  |---|---|---|
  | `/v1/models -> data[0].max_model_len` | (absent) | `131072` |
  | `/v1/models/status -> models[0].max_context_window` | `32768` | `131072` |

### New / changed tests

- **`tests/test_model_discovery.py::TestReadModelContextLength`** — 10 tests covering every config field, top-level vs nested precedence, `text_config` and `language_config` fallback, `tokenizer_config.json` fallback, the `int(1e30)` sentinel rejection, missing config files, malformed JSON, and invalid types (string, zero).
- **`tests/test_server.py::TestGetMaxContextWindow`** — 6 tests pinning the three-tier precedence: global default when no discovery, discovery tier when no override, override wins over discovery, override wins over global, `model_id=None`, unknown model id.
- **`tests/test_context_window.py::TestGetMaxContextWindow`** — its `_make_server_state` now nulls `state.engine_pool` so the discovery tier short-circuits cleanly. Without that, the `MagicMock` returned a truthy `entry.model_context_length` and broke `test_falls_back_to_global_when_model_not_set`. The pre-existing tests still cover the per-model → global fallback path; the discovery tier is covered in the new `test_server.py` class.

### DIFF
--- a/omlx/api/openai_models.py
+++ b/omlx/api/openai_models.py
@@ -381,6 +381,10 @@ class ModelInfo(BaseModel):
     object: str = "model"
     created: int = Field(default_factory=get_unix_timestamp)
     owned_by: str = "omlx"
+    # vLLM-compatible extension: lets OpenAI-style clients discover the
+    # effective context window from the listing without a separate call
+    # to /v1/models/status (see #1308).
+    max_model_len: int | None = None
 
 
 class ModelsResponse(BaseModel):
--- a/omlx/engine_pool.py
+++ b/omlx/engine_pool.py
@@ -60,6 +60,7 @@ class EngineEntry:
     config_model_type: str = ""  # Raw model_type from config.json (e.g., "deepseekocr_2")
     thinking_default: bool | None = None  # True if model thinks by default, False if not, None if unknown
     preserve_thinking_default: bool | None = None  # True when template supports preserve_thinking (Qwen 3.6+)
+    model_context_length: int | None = None  # Declared context length from config.json (None if unknown)
     engine: BaseEngine | EmbeddingEngine | RerankerEngine | STTEngine | STSEngine | TTSEngine | None = None  # Loaded engine instance
     last_access: float = 0.0  # Timestamp for LRU (0 if never loaded)
     is_loading: bool = False  # Prevent concurrent loads
@@ -189,6 +190,7 @@ class EnginePool:
                     config_model_type=getattr(info, "config_model_type", ""),
                     thinking_default=getattr(info, "thinking_default", None),
                     preserve_thinking_default=getattr(info, "preserve_thinking_default", None),
+                    model_context_length=getattr(info, "model_context_length", None),
                     is_pinned=model_id in pinned_set,
                 )
 
--- a/omlx/model_discovery.py
+++ b/omlx/model_discovery.py
@@ -277,6 +277,7 @@ class DiscoveredModel:
     config_model_type: str = ""  # Raw model_type from config.json (e.g., "deepseekocr_2")
     thinking_default: bool | None = None  # True if model thinks by default, False if not, None if unknown
     preserve_thinking_default: bool | None = None  # True when template supports preserve_thinking (Qwen 3.6+)
+    model_context_length: int | None = None  # Declared context length from config.json (None if unknown)
 
 
 def _is_unsupported_model(model_path: Path) -> bool:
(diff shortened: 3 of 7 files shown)


jundot:
Thanks for the fix. The global context default was originally there so each user could pick a context size that fits their own memory budget, with a per-model override for cases that need customizing. From the experience so far though, I have come to think the better default is the model's full context, with extra limits applied only when needed. So I want to take your change. Since `get_max_context_window` is shared, this also relaxes the `validate_context_window` admission path, which is exactly the direction I want, and oversized prompts are still caught by the prefill memory guard. The discovery propagation and the `int(1e30)` sentinel handling both look right. I'll merge this.

PR #1526, opened 2026-05-29
fix(serve): fail fast on occupied port before preload

### LINKED ISSUE #1524: omlx serve preloads pinned models before failing on occupied port
**Describe the bug**

When `omlx serve` starts with a host/port that is already in use, it can still enter ASGI lifespan startup before uvicorn reports the bind failure. Since oMLX preloads pinned models during lifespan startup, an `EADDRINUSE` failure can load and then unload pinned models before the process exits.

This is especially expensive when `omlx serve` is run under a restart supervisor, since each failed respawn can repeat the same model load/unload work.

**To Reproduce**

1. Configure one or more pinned models.
2. Start `omlx serve` on a given host/port.
3. Start a second `omlx serve` process with the same host/port.
4. Watch the second process logs.

Observed behavior: the second process can preload pinned models, then fail with an address-in-use error, then run shutdown/unload cleanup.

**Expected behavior**

A port conflict should fail before server initialization, ASGI lifespan startup, or pinned model preload. The process should still exit nonzero, but the failure should be cheap.

**Additional context**

Uvicorn's normal `Server.startup()` path runs ASGI lifespan startup before binding host/port via `loop.create_server(...)`. For a heavyweight local inference server, it is safer to bind the listen socket first and then pass that pre-bound socket into `uvicorn.Server(...).run(sockets=[...])`.

azhangd:
Fixes #1524.

## Summary

- bind the uvicorn serve socket before importing/initializing the server
- fail occupied host/port startup before ASGI lifespan and pinned model preload
- pass the pre-bound socket into `uvicorn.Server(...).run(sockets=[...])`
- close the socket if server initialization fails before uvicorn takes ownership

## Problem

`omlx serve` preloads pinned models during FastAPI lifespan startup. Uvicorn's normal startup path runs lifespan before binding the host/port, so a second process on the same port can still load and unload pinned models before exiting with `EADDRINUSE`.

Under a restart supervisor, that can turn a simple port conflict into repeated expensive model load/unload cycles.

## Approach

Use `uvicorn.Config.bind_socket()` before importing `omlx.server` or calling `init_server(...)`. If the bind fails, the process exits before server setup. If it succeeds, the existing setup path runs and uvicorn receives the already-bound socket.

This keeps the change scoped to the `omlx serve` CLI path.

## Tests

- `uv run pytest tests/test_cli.py -q`
- `uv run python -m compileall -q omlx/cli.py tests/test_cli.py`
- `git diff --check`

### DIFF
--- a/omlx/cli.py
+++ b/omlx/cli.py
@@ -165,112 +165,123 @@ def serve_command(args):
     _crash_file = open(crash_log_path, "a")
     faulthandler.enable(file=_crash_file, all_threads=True)
 
-    # Import server and config
-    from .server import app, init_server
-    from .config import parse_size
-
-    model_dirs = settings.model.get_model_dirs(settings.base_path)
-    print(f"Base path: {settings.base_path}")
-    print(f"Model directories: {', '.join(str(d) for d in model_dirs)}")
-    print(f"Memory guard tier: {settings.memory.memory_guard_tier}")
-
-    # Store MCP config path for FastAPI startup
-    # Priority: CLI arg > settings.json
-    mcp_config = args.mcp_config or settings.mcp.config_path
-    if mcp_config:
-        print(f"MCP config: {mcp_config}")
-        os.environ["OMLX_MCP_CONFIG"] = mcp_config
-
-    # Determine paged SSD cache directory
-    # Priority: --no-cache > CLI arg > settings file
-    if args.no_cache:
-        paged_ssd_cache_dir = None
-    elif args.paged_ssd_cache_dir:
-        # CLI argument takes precedence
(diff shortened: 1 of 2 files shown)


azhangd:
Opened #1526 with a fix for this.

jundot:
Thanks for the clean fix. You nailed the root cause here: uvicorn runs the ASGI lifespan (so our pinned-model preload) before it binds the port, so an occupied port burned a whole load/unload cycle. Pre-binding via `Config.bind_socket()` and handing the socket to `Server.run(sockets=...)` is the right pattern, and the bind-before-import tests cover the ordering nicely. I'll merge this.

PR #1536, opened 2026-05-30
fix(admin): improve Active Models layout on narrow viewports

samfenwick:
## Summary

Improves the admin dashboard **Active Models** section on narrow viewports. Tested by resizing the desktop browser and on a phone using the oMLX web admin. Desktop layout at `lg` and above matches `main`.

- Below `lg`: stack model name, status, and memory; show unload without hover
- Below `lg`: stack memory pressure bar + label; wrap Runtime Cache header gauges
- Pin mobile nav hamburger to the right (`ml-auto`)

## Before

On narrow viewports, long model names and memory labels shared one cramped row. Text overflowed or wrapped one character per line. Unload was hover-only and often off-screen, so unloading usually meant rotating to landscape.

Before (mobile)

## After

**Desktop** — single-row layout unchanged from `main` (hover unload, original idle formatting).

After (desktop)

**Mobile** — stacked rows, visible trash control, middot-separated status (`idle · 52s`, `Loading · 2s · ~5s left`).

After (mobile — idle / loading)

After (mobile — active / generating)

## Test plan

- [ ] Narrow browser width (~390px): long model names wrap, no horizontal scroll
- [ ] Phone / oMLX web admin: unload works in portrait without landscape
- [ ] Desktop ≥1024px: matches `main` (single row, hover unload, log-out icon)
- [ ] Loading, idle, and generating rows readable at both widths

### DIFF
--- a/omlx/admin/templates/dashboard/_navbar.html
+++ b/omlx/admin/templates/dashboard/_navbar.html
@@ -25,7 +25,7 @@
 
             <!-- Mobile hamburger button -->
             <button @click="mobileMenuOpen = !mobileMenuOpen"
-                    class="lg:hidden p-2 rounded-lg text-neutral-600 hover:text-neutral-900 hover:bg-neutral-100 transition-all">
+                    class="lg:hidden ml-auto p-2 rounded-lg text-neutral-600 hover:text-neutral-900 hover:bg-neutral-100 transition-all">
                 <i x-show="!mobileMenuOpen" data-lucide="menu" class="w-5 h-5"></i>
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, this is a real improvement. The mobile layout was genuinely unusable before. I'll merge this. One thing the diff is missing is the rebuilt `tailwind.css`, so a few new utilities (`break-words`, `lg:px-6`, etc.) aren't in the served bundle yet. I'll run `build_css.py` and push the regenerated CSS in a follow-up so the name wrapping actually kicks in.

PR #1538, opened 2026-05-30
fix(tests): target omlx logger in caplog.at_level

fry69:
Without this fix a full `pytest` run currently errors like this:

```
tests/test_mlx_lm_mtp_patch.py:258: in test_sanitize_no_mtp_weights
    assert any("no MTP weights found" in r.getMessage() for r in caplog.records)
E   assert False
E    +  where False = any(<generator object TestQwen35MoeSanitize.test_sanitize_no_mtp_weights.<locals>.<genexpr> at 0x1180412a0>)
```

---

Analysis from my coding agent:

The `omlx` logger gets an explicit level set by earlier tests, preventing `caplog.at_level` from overriding it.
The chain of events
1. `test_cli.py` runs before `test_mlx_lm_mtp_patch.py` (alphabetical order). Inside `test_cli.py`, tests call `serve_command()` from `omlx/cli.py`, which executes:
- `logging.basicConfig(level=logging.INFO)` — sets root logger to `INFO`
- `logging.getLogger("omlx").setLevel(log_level)` — sets the omlx logger to INFO explicitly (`omlx/cli.py:100`)
1. Later, `test_sanitize_no_mtp_weights` uses `caplog.at_level(logging.DEBUG)` (without specifying a logger), which only sets the root logger to `DEBUG`.
2. The `omlx `logger still has its explicit INFO level. In Python's logging hierarchy, `omlx.patches.mlx_lm_mtp.qwen35_model` inherits INFO from `omlx`, so `logger.debug(...)` messages are filtered before they ever reach the root logger's `LogCaptureHandler`.
3. Result: `caplog.records` is empty → assertion fails.
Why it passes in isolation
When running only `test_mlx_lm_mtp_patch.py`, `test_cli.py` never runs, so `omlx` logger is never explicitly set. Its level stays `NOTSET`, inheriting from root — which `caplog.at_level(logging.DEBUG)` can override.

---

### DIFF
--- a/tests/test_mlx_lm_mtp_patch.py
+++ b/tests/test_mlx_lm_mtp_patch.py
@@ -388,7 +388,7 @@ class TestQwen35MoeSanitize:
         (model quantized without preserve_mtp). Must not crash."""
         import logging
 
-        with caplog.at_level(logging.DEBUG):
+        with caplog.at_level(logging.DEBUG, logger="omlx"):
             result = moe_model.sanitize(self._backbone_weights())
         assert not any("mtp" in k for k in result)
         assert any("no MTP weights found" in r.getMessage() for r in caplog.records)


fry69:
@jundot This fix still seems to be necessary to turn test completely green again.

Also the just added ci workflow tests in -> https://github.com/jundot/omlx/commit/3136f73941bea84a97d2ade2f3dd7c5bddb5d352 do not seems to run?

Kistaro:
Oops, my https://github.com/jundot/omlx/pull/1562 took a different approach -- I propose rewriting this test completely instead of trying to capture the log message better.

The purpose of the function under test is to process model weights. Whether it prints a specific debug log message or not is immaterial. The test can verify that it doesn't crash, doesn't actually emit any MTP weights, and _does_ emit the other weights that function should generate, and then nobody has to read any debug logs at all to figure out whether the function worked.

Reading logs in a unit test means something went wrong with the test design somewhere, unless the purpose of the function under test _is_ to log a message...

jundot:
@fry69 Thanks for flagging this. The workflow file was present, but GitHub Actions were still disabled in the repository settings, so no PR check suites were being created.

I have enabled Actions now with read-only default workflow permissions and SHA-pinned GitHub actions only. I will retrigger the checks.

On the test fix, I will take another look and follow up with a separate opinion.

jundot:
@fry69 Thanks, I reproduced the order-dependent failure on main with `test_cli.py` before `test_sanitize_no_mtp_weights`, and confirmed this PR fixes that sequence.

I agree this is a real test isolation issue. #1562 has a broader behavior-focused rewrite of the same test, but this one-line fix is the right minimal unblocker for CI, so I will merge this first.

PR #1539, opened 2026-05-30
feat(mtp): allow sidecar style mtp.safetensors from MLX OptiQ models

beamivalice:
* Allow sidecar style MTP load, shipped with MLX OptiQ (mlx-community) - https://pypi.org/project/mlx-optiq/
* Delete duplicate fn _model_has_mtp_weight_tensors trying to do the same thing with _checkpoint_has_mtp_weights

* Test with 27B-OptiQ-4bit benchmark and it's increased by a lot 8k/128 - (26 t/s -> 34 t/s)

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -438,57 +438,15 @@ def _mtp_compat_for_model(model_info: dict) -> tuple[bool, str]:
             f"model_type={model_type!r} is not on the MTP whitelist "
             "(supported: qwen3_5*, qwen3_6*, deepseek_v4*)"
         )
-    if not _model_has_mtp_weight_tensors(Path(model_path)):
+    from ..utils.model_loading import _checkpoint_has_mtp_weights
+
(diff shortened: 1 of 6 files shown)


jundot:
Thanks for looking into OptiQ-style MTP sidecars. I’m hesitant to add another runtime load_model monkey patch for this layout, though.

This is not just a checkpoint-detection change: it globally wraps mlx_lm.utils.load_model, temporarily patches glob.glob during load, and has to coexist with the existing DeepSeek V4 / MTP / VLM load-time patches. That makes the support path fairly fragile for a third-party packaging layout.

Before merging this, I’d like to understand whether this really needs to be handled at runtime. From oMLX’s side, a cleaner path may be to normalize OptiQ-style mtp.safetensors during an import/conversion step, producing normal model*.safetensors shards with mtp.* entries in the weight map, instead of adding another global load_model/glob monkey patch.

beamivalice:
I agree if you have a better way to do this. I just follow the monkey patch list that you allowed.

PR #1540, opened 2026-05-30
fix(ui): align headers in model settings table

ilukashin:
This is annoying 🙂

Before:
Снимок экрана — 2026-05-30 в 14 05 05

After:
Снимок экрана — 2026-05-30 в 14 04 24


jundot:
Good catch, thanks. The size header was `w-20` while the size cell is `w-60`, so the header sat narrower than the column. This lines them up. Merging.

PR #1489, opened 2026-05-28
refactor(packaging): rename mlx-framework venvstacks layer to mlx-base

popfido:
## Summary

- Rename the venvstacks `[[frameworks]]` layer from `mlx-framework` → `mlx-base`. venvstacks auto-prefixes framework layers with `framework-`, so the old name produced the tautological directory `Contents/Frameworks/framework-mlx-framework/`. The new name yields `framework-mlx-base/` — keeps the venvstacks convention, drops the redundancy, and leaves room for future layers (e.g. `framework-mlx-extras`).
- Touches 5 files: `packaging/venvstacks.toml` (layer name + comment), `packaging/build.py` (path strings, `_LAYER_REQUIREMENTS_SOURCES` key, resolved-requirements filename), `apps/omlx-mac/Scripts/build.sh` (donor copy paths), `apps/omlx-mac/Sources/Server/PythonRuntime.swift` (bundled `PYTHONPATH` entry), `packaging/README.md` (layer table). No API surface change — strictly an internal rename.
- Reviewers should also delete their stale `packaging/_export/` and `packaging/requirements/framework-mlx-framework/` before rebuilding so venvstacks emits the new layout.

## Test plan

- [x] `grep -rn "mlx-framework\|mlx_framework"` across `apps/`, `packaging/`, `omlx/`, `docs/` is clean
- [x] Wipe `packaging/_export/` + `packaging/requirements/framework-mlx-framework/` and run `apps/omlx-mac/Scripts/build.sh release`
- [x] `packaging/_export/framework-mlx-base/` is produced by venvstacks
- [x] `packaging/requirements/framework-mlx-base/requirements-framework-mlx-base-macosx_arm64.{in,txt,txt.json}` regenerated
- [x] `apps/omlx-mac/build/Stage/oMLX.app/Contents/Frameworks/framework-mlx-base/` is present
- [x] Launch the staged bundle — Python runtime resolves `PYTHONPATH` via the new path
- [ ] Reviewer rebuilds locally after wiping stale artifacts

PR #1541, opened 2026-05-30
fix(packaging): point _install_xgrammar at framework-mlx-base after layer rename

### LINKED ISSUE #1530: Structured outputs / xgrammar stub
I am very happy to see that a stubbed out version of `xgrammar` got merged, see #1453 

I tested this briefly while it was a PR, `xgrammer` loaded fine when it was not merged, but building current `main` (04da6ff68edc1d2c0e34c0fbbee4ebeca38cd0fe) as an app and running it produces this warning and structured outputs do not seem to get respected in my test, like they do with e.g. `ollama` with the same model.

```
omlx.engine.batched - INFO - [-] - Structured output is not available in the DMG version (xgrammar requires torch which significantly increases app size). Use the pip or Homebrew version for structured output support.
```

Command to build:
```shell
time apps/omlx-mac/Scripts/build.sh release --rebuild-donor
```

I cross checked with the `pip` install and a fresh `.venv` environment and it also shows that `xgrammar` does not get loaded/detected:

```
omlx.engine.batched - INFO - [-] - Structured output requires xgrammar. Install with: pip install 'omlx[grammar]'
```

Am I missing something? Should this not work now out of the box?

cfbraun:
## Summary

Fixes #1530.

#1489 renamed the venvstacks layer from `mlx-framework` to `mlx-base` in `packaging/build.py`. `_install_mlx_audio`, `_install_paroquant`, and `_install_spacy_model` were updated; `_install_xgrammar` was missed and still writes to `framework-mlx-framework` — which the new export never creates.

```python
# packaging/build.py:841-847 (before)
fw_site = (
    export_dir
    / "framework-mlx-framework"   # ← stale
    / "lib"
    / "python3.11"
    / "site-packages"
)
```

The early-return at lines 759-760 (`site-packages not found`) then fires silently, the bundle ships without xgrammar / apache-tvm-ffi, and at runtime the engine wrapper logs:

> Structured output is not available in the DMG version (xgrammar requires torch which significantly increases app size). Use the pip or Homebrew version for structured output support.

— which misleads the user, both because this message predates the bundled-stub work from #1453 and because the actual cause is the stale layer path, not torch.

## The fix

One-line: `framework-mlx-framework` → `framework-mlx-base` to match the other install functions in the same file.

## Test plan

- [x] `grep -c framework-mlx-framework packaging/build.py` returns 0 (no stale references remain)
- [ ] Local bundle rebuild + verification that `_xgrammar_*.installed` sentinel lands in `framework-mlx-base/lib/python3.11/site-packages/` — kicking off now; will confirm once the build finishes.
- [ ] Runtime confirmation that structured output is available on the rebuilt bundle (no "Structured output is not available in the DMG version" log entry).

Will also leave a follow-up note on #1530 once the rebuilt bundle confirms the fix end-to-end. fry69 may want to re-test on their side too.

## Note on the misleading log message

Out of scope here, but worth flagging: the `if method == "dmg"` branch at `omlx/engine/batched.py:151` (and the matching site in `vlm.py`) still logs the pre-#1453 message about torch size, which doesn't reflect that xgrammar is now bundled with a torch stub. Even with this fix landed, any future failure of the stubbed-xgrammar path will surface that same outdated text. Happy to send a follow-up PR updating the message to match the new reality if you'd like.

### DIFF
--- a/packaging/build.py
+++ b/packaging/build.py
@@ -840,7 +840,7 @@ def _install_xgrammar(export_dir: Path):
     """
     fw_site = (
         export_dir
-        / "framework-mlx-framework"
+        / "framework-mlx-base"
         / "lib"
         / "python3.11"
         / "site-packages"


cfbraun:
@fry69 confirmed in #1530 that this fix rebuilds and runs cleanly end-to-end — `GrammarCompiler initialized` lands in the log, no "Structured output is not available" warning, and their structured-output test suite passes against `Mistral-Small-3.2-24B-Instruct-2506-MLX-8bit`.

jundot:
Thanks a lot for jumping on this. I saw #1530 come in but hadn't had a chance to dig in yet, so I really appreciate you turning the fix around before I even got home to look at it. Root cause is clear, #1489 renamed the layer to mlx-base and this one spot in _install_xgrammar got left behind. Nice that fry69 already confirmed it rebuilds and runs structured output end to end. I'll merge this.

PR #1557, opened 2026-05-31
fix(scheduler): log hot-cache-only mode instead of SSD cache when SSD is skipped

### LINKED ISSUE #1316: Misleading "paged SSD cache enabled" log line when hot_cache_only is active (v0.3.9rc1)
## Summary

In `v0.3.9rc1` with `cache.enabled: true` and `cache.hot_cache_only: true`, the runtime correctly honors the hot-cache-only mode (no SSD writes, no `queue full` warnings, `~/.omlx/cache` stays empty) — but the scheduler's cache-init log still emits the same `paged SSD cache enabled: cache_dir=…` line as a full-SSD-cache configuration. The log is misleading; operators verifying that PR #864 took effect can't tell from the log alone.

## Environment

- oMLX: `0.3.9rc1` (DMG build, macOS 26 Tahoe)
- Hardware: Apple M2 Pro, 16 GB unified
- Settings (`~/.omlx/settings.json`):
  ```json
  "cache": {
    "enabled": true,
    "hot_cache_only": true,
    "ssd_cache_dir": "/Users/.../.omlx/cache",
    "ssd_cache_max_size": "46GB",
    "hot_cache_max_size": "2GB",
    "initial_cache_blocks": 256
  }
  ```

## Observed log lines on first model load after server restart

```
omlx.cache.paged_cache - INFO - PagedCacheManager initialized: block_size=256, initial_blocks=256, max_blocks=100000, max_tokens=25600000
omlx.cache.paged_ssd_cache - INFO - PagedSSDCacheManager initialized: dir=/Users/.../.omlx/cache, max_size=46.00 GB, hot_cache=2.00 GB, existing_files=0, disk_free=60.00 GB, cache_used=0 B
omlx.cache.paged_cache - INFO - paged SSD cache manager connected to PagedCacheManager
omlx.cache.prefix_cache - INFO - PagedSSDCacheManager connected to BlockAwarePrefixCache
omlx.scheduler - INFO - paged SSD cache enabled: cache_dir=/Users/.../.omlx/cache, max_size=46.00 GB, block_size=256 tokens
omlx.scheduler - INFO - paged SSD cache enabled: /Users/.../.omlx/cache, block_size=256, max_blocks=100000
```

The two `paged SSD cache enabled` lines from `omlx.scheduler` are identical whether `hot_cache_only` is true or false.

## Runtime behavior is actually correct

To confirm the fix from PR #864 *is* doing its job at runtime:

- After 2-pass benchmark (same 2K-token prefix, different user message): pass-2 was 4.4× faster than pass-1 → hot cache hits are working.
- `du -sh ~/.omlx/cache` stays at `0 B` across the session — no SSD writes.
- `grep "queue full" ~/.omlx/logs/server.log` after the test session: zero matches.

So the runtime correctly skips the SSD layer; only the init log text is stale.

## Suggested fix

Branch the scheduler log on `cache.hot_cache_only`:

```
if hot_cache_only:
    log.info("hot-cache-only mode enabled: hot_cache=%s, block_size=%d tokens",
             hot_cache_size_human, block_size)
else:
    log.info("paged SSD cache enabled: cache_dir=%s, max_size=%s, block_size=%d tokens",
             cache_dir, max_size_human, block_size)
```

Same treatment for the second duplicate line a few lines below in the same scheduler init path.

## Why this matters

- Operators following the upgrade path from dev2 (where `hot_cache_only: true` was silently ignored, see #605 / PR #864) check this log to confirm the flag is now honored. The stale text reads as "still broken" even when behavior is correct.
- Downstream config-recipe repos that document oMLX setup currently have to include a paragraph telling readers to ignore this log line. Replacing it with mode-accurate text removes that footnote.

Happy to send a small PR if useful.

Proudly Made in Nebraska. Go Big Red! 🌽 https://xkcd.com/2347/

richgoodson:
## Summary

When `cache.hot_cache_only` is true, the scheduler still logged `paged SSD cache enabled: ...` during cache init, even though the SSD tier is skipped in that mode. The line did not match what actually happens (empty SSD cache dir, no spill, no `queue full`), so reading the startup log gives no sign that hot-only mode is on.

Both `logger.info` sites in the paged-cache init path now check `self.config.hot_cache_only` and log a `hot-cache-only mode enabled: ...` line when it is set. The full-SSD text is unchanged. This is a log-text change only, no caching behavior changes.

### Before (hot_cache_only: true)

```
paged SSD cache enabled: /Users/.../.omlx/cache, block_size=256, max_blocks=...
paged SSD cache enabled: cache_dir=/Users/.../.omlx/cache, max_size=..., block_size=256 tokens
```

### After (hot_cache_only: true)

```
hot-cache-only mode enabled: block_size=256, max_blocks=...
hot-cache-only mode enabled: hot_cache_max=..., block_size=256 tokens
```

With `hot_cache_only: false`, both lines are byte-for-byte unchanged.

## Why no automated test

Both lines only fire during a full Scheduler model-load init. The existing tests check the cache classes (`PagedSSDCacheManager` and the rest) directly, and nothing in the suite runs a full scheduler init, so there is no place to hang a caplog assertion. Adding one would mean standing up a mock scheduler init just to check a log string, which is more setup than a log-text change is worth. I tested it at runtime instead (below).

## Test plan

Tested on Apple Silicon with a plain text model (non-VLM, non-dflash) so it goes through the standard scheduler paged-cache init path, confirmed by `PagedSSDCacheManager initialized` in the startup log:

- [x] **hot_cache_only: true** - two `hot-cache-only mode enabled` lines, zero `paged SSD cache enabled`, SSD cache dir stays at 0B
- [x] **hot_cache_only: false** (regression guard) - two `paged SSD cache enabled` lines, zero hot-only lines, original text unchanged
- [x] **runtime spot-check** (hot-only) - SSD dir stays empty, zero `queue full`, prefix cache still hits (warm pass ~17.6s vs cold ~26.7s on the same 2K prefix)

Fixes #1316

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -910,11 +910,18 @@ class Scheduler:
                 self.block_aware_cache.set_cold_restore_callback(
                     self._restore_block_from_cold
                 )
-                logger.info(
-                    f"paged SSD cache enabled: {self.config.paged_ssd_cache_dir}, "
-                    f"block_size={self.config.paged_cache_block_size}, "
-                    f"max_blocks={max_blocks}"
-                )
+                if self.config.hot_cache_only:
+                    logger.info(
+                        f"hot-cache-only mode enabled: "
+                        f"block_size={self.config.paged_cache_block_size}, "
+                        f"max_blocks={max_blocks}"
+                    )
+                else:
+                    logger.info(
+                        f"paged SSD cache enabled: {self.config.paged_ssd_cache_dir}, "
+                        f"block_size={self.config.paged_cache_block_size}, "
+                        f"max_blocks={max_blocks}"
+                    )
 
             # Async store_cache executor: single worker so submissions are
             # serialized (matches the original synchronous order) and we
@@ -6818,12 +6825,19 @@ class Scheduler:
                         "Failed to initialize boundary snapshot SSD store: %s", e
                     )
 
-            logger.info(
-                f"paged SSD cache enabled: "
-                f"cache_dir={self.config.paged_ssd_cache_dir}, "
-                f"max_size={self._format_bytes(self.config.paged_ssd_cache_max_size)}, "
-                f"block_size={self.config.paged_cache_block_size} tokens"
-            )
+            if self.config.hot_cache_only:
+                logger.info(
+                    f"hot-cache-only mode enabled: "
+                    f"hot_cache_max={self._format_bytes(self.config.hot_cache_max_size)}, "
+                    f"block_size={self.config.paged_cache_block_size} tokens"
+                )
+            else:
+                logger.info(
+                    f"paged SSD cache enabled: "
+                    f"cache_dir={self.config.paged_ssd_cache_dir}, "
+                    f"max_size={self._format_bytes(self.config.paged_ssd_cache_max_size)}, "
+                    f"block_size={self.config.paged_cache_block_size} tokens"
+                )
 
         except Exception as e:
             logger.error(f"Failed to initialize paged SSD cache: {e}")


jundot:
Thanks, this matches the issue. The hot-cache-only path was doing the right thing at runtime, but the two scheduler startup lines still made it look like the SSD tier was active. This keeps the full SSD-cache log text unchanged and only swaps those two lines when `hot_cache_only` is set, so it is a clean log-only fix. I'll merge this.

Issue #1554, opened 2026-05-30
Pi auto-config ignores "thinking" flag in model setup


Kistaro:
**Describe the bug**
Using the command line specified under the "Pi" header on the admin dashboard to launch Pi, after showing a model selector, the actual "thinking" setting for the model is not used to decide whether to configure the `reasoning` flag in Pi's `models.json`. It appears from [omlx/integrations/pi.py:`PiIntegration._is_reasoning_model`](https://github.com/jundot/omlx/blob/e623151a67bc45708e46662f9a313fece2d8b0e2/omlx/integrations/pi.py#L45) (version from 2026-May-30) that fixed model slug substrings are used to identify "reasoning" models.

**To Reproduce**
1. Download and appropriately configure `Qwen3.6-27B-oQ8-mtp`, a reasoning model. (An appropriate configuration for this model include a 262144 token context window, temperature 0.6, top P 0.95, top K 20, "Enable Thinking" on, and an additional Chat Template Kwargs pair: `"preserve_thinking": True`.
2. Install the Pi agent harness such that `pi` is available on your PATH.
3. Use `/Applications/oMLX.app/Contents/MacOS/omlx-cli launch pi` to launch Pi.
4. Exit Pi (ctrl-d or `/quit`).
5. Open `~/.pi/agent/models.json` to see the injected config. Note that Qwen has been configured as a non-reasoning model.

**Expected behavior**
Qwen is a reasoning model and should be configured as such.

**Desktop (please complete the following information):**
 - macOS Version: 26.5
 - oMLX Version 0.3.12

**Additional context**
It would make more sense to refer to the model configuration in `_is_reasoning_model` rather than the model slug, but `configure` appears to take only extracted fields rather than a useful abstraction of "how oMLX runs this model". I recommend refactoring. All integrations would benefit from this approach, I suspect.

I'll take a look to see how big of a plumbing job that would be and whether oMLX already has "a model, as configured" as an object, or whether it needs to be bundled up so it _can_ be handled as "a thing".

Kistaro:
I will have a pull request for this shortly.

PR #1562, opened 2026-05-31
fix (integrations): use "Enable Thinking" setting

Kistaro:
Fixes issue #1554.

Model settings override model defaults, which override global defaults (which might be sampler defaults). One-off versions of this resolution order are scattered throughout the codebase. I've implemented ConfiguredModel to glue these layers together, written a general resolver, and written resolving accessors for a couple of spots along my path to getting the resolved thinking setting into the status endpoint. There are more opportunities use ConfigModel to simplify more parts of the codebase that work like this, but an incremental approach is easier to review -- this patch is big enough as-is.

Claude Opus 4.8 developed the initial version, which I rewrote by hand because that's easier than arguing with 4.8. Opus 4.6 stabilized the tests, and I didn't agree with _its_ design either, but I'm satisfied with its cleanup.

Along the way, it discovered a flaky test, which passed on reruns. That bothered me, so I had Claude investigate; the original test was order-dependent, because something else in the tests also messes with logging config in a way that causes _this_ test to not subsequently work. No idea what that was, Claude started trying to bisect four thousand tests but I stopped it and pointed out that we don't _really_ care what log messages this function emits -- that's not the purpose of the function; we care what results it gets and that it doesn't crash in this case. Claude rewrote the test accordingly and I think this version is much better -- it tests behavior rather than an implementation detail, and is no longer order-dependent. However, that fix is otherwise unrelated to the function of this pull request; let me know if I should break it out.

The unsquashed version of this change is the branch https://github.com/Kistaro/omlx/tree/reasoning-integration if you want to see more details of what was Claude's and what I wrote by hand.

There are some weird things I left as-is because I didn't want to change multiple observable behaviors at once without running it by the maintainers first:
* The OpenClaw setup never attempted fallback to the model slug heuristic; should it?
* The OpenClaw setup ignores a bunch of fields; the context limit seems especially bad to ignore. But, I don't use OpenClaw (frankly, it scares me), so maybe there's a reason it has to be that way?
* Integrations technically have a common base class, but the subclass interfaces effectively diverged and are relying on bonus keys being sent through kwargs. This needs a cleanup of its own.
* the model slug heuristic is very weak. There are obvious ways to improve it. But instead of improving an approach for making a bad guess, I'd rather just plumb data around so we don't have to guess. (Thus, this patch.)
* I'd _like_ to refactor a bunch of stuff to use ConfiguredModel, there are obvious opportunities to do so. If y'all have suggestions for how to make models notice it's there and prefer to use it in new code (and when already rewriting older code anyway), I'd be glad to make it more prominent in some way.
* Some of the tests now involve explicitly mocking out a never-None value to return None, because it _used_ to be possible to express None there and there was a test for it.

Please let me know if this needs more work before it can be considered, I'm flexible. I just wanted to fix this bug for _me_ and I can't leave well enough alone so I wrote a config resolver rather than writing another copy of layered resolution in place...


jundot:
Thanks, this direction makes sense to me. I agree that the repeated resolution order is worth centralizing, but I’d like to keep this PR narrower now that I cleaned up the integration boundary on main.

I pushed that cleanup here: https://github.com/jundot/omlx/commit/850f191

Could you rebase on main and keep ConfiguredModel focused on the server/model configuration layer only? In particular, I’d prefer this PR to use ConfiguredModel for resolving model settings/defaults for the status endpoint and request handling, but not pass it into integration classes directly. Integrations should continue to receive IntegrationContext, which is now the boundary for external tool launch/configuration.

A few concrete things I’d like adjusted:
- Please rebase on main and adapt to IntegrationContext.
- Keep ConfiguredModel as an internal resolver for ModelSettings + EngineEntry + SamplingDefaults.
- Avoid broad follow-up refactors beyond the paths needed for enable_thinking/preserve_thinking and existing context/max-token resolution.
- Please fix the Python 3.11 issue in `first_present[T](...)`.
- If `global_settings` is not needed by the current resolved properties, please leave it out for now.

If you see a cleaner boundary than this after rebasing, I’m open to that too. My main goal is to keep the integration interface explicit and keep ConfiguredModel from becoming a cross-layer object too early.

With that shape, I think the PR stays focused while still giving us the right place to centralize this resolution logic.

Kistaro:
New revision ready for review:

* Merged from main (repeatedly), manually resolving conflicts in integrations
* Reverted change to `test_sanitize_no_mtp_weights` (will start a separate PR)
* Python 3.11 syntax for that generic function annotation in `ConfiguredModel`
* Remove `global_settings` from ConfiguredModel
* Plumb context window limit into Codex; note I haven't tried to do anything for its compaction threshold
* Copilot integration: I couldn't find any docs for most of the environment variables it already sets, no idea what to do for thinking mode if anything.
* Hermes integration: `enable_thinking` controls thinking effort as high vs. none, similar to Codex.

Two tests fail in `test_turboquant.py` but they also fail on `main`.

This still makes some extensive changes to tests where the existing mock-based tests are incompatible with the new implementation. I prefer to avoid mocks as much as possible -- mocks are fine for integrations with distant systems where "interacts with this interface in this specific way" is the exact guarantee you want to test, but otherwise it just reduces the realism of the test., so I preferred real objects when practical. I still use a mock to get it into the system, although I think that could reasonably be avoided in several cases; going through and cleaning up more "gratuitous mocks" might be a useful cleanup project but it's out-of-scope here.

Also, I swear I wrote this comment myself by hand, I just sound as stilted as Opus 4.8 when it's half past midnight

Kistaro:
btw, I'm tempted to rename this pull request to `feat(server): new class ConfiguredModel for resolving layered configs` because this pull request name is pretty inaccurate now; you already wrote the integration-side handling in your IntegrationContext rewrite! (a patch I like, I think that is a big improvement in maintainability.) But I don't want to interfere with whatever your system for triaging your flood of pull requests is by renaming this thing when you weren't expecting it.

jundot:
Thanks for iterating on this. The direction still looks useful, but this branch now conflicts with main, so I can't merge it as-is.

The main thing I need preserved during the rebase is the current context-window behavior on main: max_context_window_policy must still clamp discovered native context lengths, while per-model overrides and fallback defaults keep their current semantics. I also need the new resolver path to resolve model aliases before looking up settings or EngineEntry data, otherwise alias requests can fall back to global limits.

The smaller lint/whitespace cleanup and direct /v1/models/status assertions for enable_thinking/preserve_thinking are straightforward, but please include them while rebasing if possible.

Kistaro:
I didn't have any merge conflicts with lint/whitespace cleanups (not any that I caught, anyway). New context window limiter, and new embedding behaviors, ported.

The special cases for `is_markitdown_model` would be nice to also handle via ConfiguredModel, but this PR doesn't need to expand in scope any, so that has been left as-is for now.

So this should be ready to go unless you need more changes.

Kistaro:
Updated to resolve the Exciting New Merge Conflict.

bruvv:
@Kistaro i would love this to work, can you rebase it again and hopefully this can be merged and included in a new release :)

bruvv:
Ping @Kistaro

bruvv:
@jundot can you look at this pr please ?

Kistaro:
I will rebase this either today after work or this weekend, since there is apparent interest. Sorry for the delay.

Kistaro:
All right, I think this is ready to merge into`main` again (squashed, presumably). This was vibe-coded with heavy steering (and now I have a skill about how to systematically rewrite doc comments, a useful distillation of my spending two hours ranting at Claude on the topic), and I fully reviewed the result.

PR #1563, opened 2026-05-31
ci: add Python 3.14 to CI matrix

fry69:
Python 3.14 is already widely used on developer machines, and 3.15 is approaching release. I'd recommend to add it to the CI matrix.


jundot:
Thanks, this looks good to me. The new 3.14 CI job is passing on macOS along with the existing 3.11-3.13 jobs, so this is a safe CI coverage update. I'm going to merge it.

PR #1564, opened 2026-05-31
fix(server): signal unenforced response_format via warning log and Warning header

### LINKED ISSUE #1241: response_format.type=json_schema is accepted by /v1/chat/completions but not enforced in assistant content
**Describe the bug**  
When calling OMLX's OpenAI-compatible `POST /v1/chat/completions` endpoint with `response_format.type = "json_schema"` and `strict = true`, the request is accepted, but the assistant output does not follow the supplied schema.

In the minimal reproduction below, the schema requires each item to contain only `id` and `dst`. However, the returned content contains `src` instead of `dst`, omits the required `dst` field, and includes a field not allowed by the schema.

In larger translation inputs, the behavior becomes less stable: the returned structure may drift further from the schema, input fields may be echoed back, and `message.content` may not remain valid standard JSON.

This suggests that `json_schema` is either not being enforced or only partially honored on this endpoint.

**To Reproduce**  

1. Save the following as `request.json`:

```json
{
  "model": "HY-MT1.5-1.8B-oQ4",
  "temperature": 0.2,
  "messages": [
    {
      "role": "system",
      "content": "Translate English to Simplified Chinese. Return JSON that matches the schema."
    },
    {
      "role": "user",
      "content": "{\"items\":[{\"id\":\"1\",\"src\":\"Hello.\"},{\"id\":\"2\",\"src\":\"How are you?\"}]}"
    }
  ],
  "response_format": {
    "type": "json_schema",
    "json_schema": {
      "name": "translation_batch",
      "strict": true,
      "schema": {
        "type": "object",
        "properties": {
          "items": {
            "type": "array",
            "items": {
              "type": "object",
              "properties": {
                "id": { "type": "string" },
                "dst": { "type": "string" }
              },
              "required": ["id", "dst"],
              "additionalProperties": false
            }
          }
        },
        "required": ["items"],
        "additionalProperties": false
      }
    }
  }
}
```

2. Send the request:

```bash
curl http://127.0.0.1:8000/v1/chat/completions \
  --data-binary @request.json
```

3. Observe the returned assistant content.

Returned response:

```json
{
  "items": [
    { "id": "1", "src": "Hello." },
    { "id": "2", "src": "How are you?" }
  ]
}
```

This does not match the requested schema.

**Expected behavior**  
The request should either:

1. Return assistant content that strictly matches the schema, for example:

```json
{
  "items": [
    { "id": "1", "dst": "你好。" },
    { "id": "2", "dst": "你好吗？" }
  ]
}
```

2. Or reject the request with a clear error if `response_format.type=json_schema` is not fully supported for this endpoint/model/template combination.

At minimum, when `strict = true` is provided, the returned content should not include undeclared fields and should not omit required fields.

**Screenshots**  
Not applicable.

 - macOS Version: `26.3.1`
 - oMLX Version: `0.3.8`

**Additional context**  
This issue is important for structured generation workflows such as translation pipelines, where downstream code depends on stable machine-readable output.

The minimal reproduction above already shows schema mismatch with valid JSON output. In larger real-world inputs, the problem becomes more severe: the returned structure can become unstable, and `message.content` may stop being valid standard JSON altogether.

Because the server accepts `response_format.type=json_schema` without error, it is currently difficult for clients to know whether schema enforcement is actually supported.

richgoodson:
`/v1/chat/completions` accepts `response_format` with `type: "json_schema"` and `strict: true`, returns 200, but the assistant content does not follow the schema. The request looks honored when it isn't, so a client has no way to tell that the output is unconstrained.

## Root cause

`response_format` is only schema-enforced when a grammar compiler is available. `_compile_grammar_for_request` (`server.py`) reads `engine.grammar_compiler`; when that is `None` (no xgrammar installed, or an engine that never installs one such as DFlash), or when grammar compilation raises, the request silently falls back to prompt injection (`server.py:2233`). The model is asked to produce JSON in the prompt, but nothing constrains decoding, so undeclared fields appear, required fields are dropped, and on larger inputs the content stops being valid JSON. `structured_outputs` already returns a 400 in this case; `response_format` was the silent path.

## Fix

The reporter asked for one of two things: enforce the schema, or signal that it wasn't. Enforcing on every engine is a larger change (it needs grammar wiring on engines that don't have a compiler at all). This PR does the signal, on both surfaces:

- **Operator-visible log.** The fallback now logs a warning instead of degrading silently. A `strict: true` request gets a message that names the unhonored strict intent.
- **Client-visible header.** The response carries an RFC 7234 `Warning` header (code 199) on both the streaming and non-streaming paths so the caller can detect the degrade without parsing the body. strict requests get a header that says the output is NOT schema-enforced; non-strict `response_format` requests get a generic one.

The request still returns 200 with best-effort content. I chose not to turn an accepted request into a 400. That is a breaking behavior change, and neither I nor the reporter owns this repo. If maintainers prefer a hard reject, the strict-detection helper (`_response_format_requests_strict`) is the single place to branch on.

## Scope

This addresses `/v1/chat/completions`, which is what the issue reports. `/v1/completions` and `/v1/messages` share `_compile_grammar_for_request`, so they get the warning log, but not the header. Extending the header to those endpoints is a separate change. Actually enforcing the schema on compiler-less engines (the other half of the reporter's ask) is also separate.

## Evidence

Live-tested against a DFlash model (`Qwen3.6-27B-4bit`), which has no grammar compiler, so every `response_format` request hits the fallback.

Strict `json_schema`, non-streaming. The body reproduces the bug from the issue (`src` instead of `dst`), and the header now flags it:

```
HTTP/1.1 200 OK
warning: 199 omlx "response_format strict json_schema not enforced; grammar-constrained decoding unavailable, output is best-effort and NOT schema-enforced"
content-type: application/json

{"choices":[{"message":{"role":"assistant","content":"{\"items\": [{\"id\": \"1\", \"src\": \"Hello.\"}]}"},"finish_reason":"stop"}], ...}
```

Strict `json_schema`, streaming (`stream: true`). Same header on the SSE response:

```
HTTP/1.1 200 OK
content-type: text/event-stream
warning: 199 omlx "response_format strict json_schema not enforced; grammar-constrained decoding unavailable, output is best-effort and NOT schema-enforced"
```

Non-strict `json_object`. Generic header, no strict wording:

```
HTTP/1.1 200 OK
warning: 199 omlx "response_format not enforced; grammar-constrained decoding unavailable, output is best-effort"
```

No `response_format` (control). No `Warning` header:

```
HTTP/1.1 200 OK
content-type: application/json
```

Unit tests: `tests/test_grammar.py` covers the strict-detection helper, the degrade-not-raise behavior, and the header text (strict vs generic, single-line ASCII). Full grammar suite is green (72 passed, 3 skipped, the skips are pre-existing).

Fixes #1241

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -2231,8 +2231,11 @@ async def create_chat_completion(
     )
     # Fall back to prompt injection when grammar is not compiled. The degrade
     # is also surfaced to the caller as a Warning response header (#1241).
+    # Only response formats that actually request grammar-constrained JSON
+    # (json_object / json_schema) can be "unenforced"; a plain text format
+    # never asked for enforcement, so it must not warn (#1241 review).
     response_format_warning = None
-    if compiled_grammar is None and response_format:
+    if compiled_grammar is None and _response_format_requests_grammar(response_format):
         response_format_warning = _response_format_warning_header(response_format)
         json_instruction = build_json_system_prompt(response_format)
         if json_instruction:
@@ -2673,6 +2676,22 @@ def _response_format_requests_strict(response_format) -> bool:
         return False
     strict = js.get("strict") if isinstance(js, dict) else getattr(js, "strict", None)
     return bool(strict)
+
+
+def _response_format_requests_grammar(response_format) -> bool:
+    """True when an OpenAI ``response_format`` asks for grammar-constrained JSON.
+
+    Only ``json_object`` and ``json_schema`` map to grammar-constrained
+    decoding, so only those can be silently *unenforced* when no grammar
+    compiler is available.  A plain ``{"type": "text"}`` (or absent) format
+    never requested enforcement and must not trigger the downgrade Warning
+    header or prompt-injection fallback (#1241 review).
+    """
+    if response_format is None:
+        return False
+    rf = response_format
+    rf_type = rf.get("type") if isinstance(rf, dict) else getattr(rf, "type", None)
+    return rf_type in ("json_object", "json_schema")
 
 
 def _compile_grammar_for_request(
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -2643,6 +2643,28 @@ def _compile_bare_grammar(compiler, fmt: dict):
     return None
 
 
+def _response_format_requests_strict(response_format) -> bool:
+    """True when an OpenAI ``response_format`` demands strict json_schema output.
+
+    A ``json_schema`` response_format with ``strict: true`` signals that the
+    caller expects schema-conformant output, not best-effort.  When
+    grammar-constrained decoding is unavailable the request still falls back to
+    prompt injection, but the downgrade is logged at a level that names the
+    unhonored ``strict`` intent so it is not silent (issue #1241).
+    """
+    if response_format is None:
+        return False
+    rf = response_format
+    rf_type = rf.get("type") if isinstance(rf, dict) else getattr(rf, "type", None)
+    if rf_type != "json_schema":
+        return False
+    js = rf.get("json_schema") if isinstance(rf, dict) else getattr(rf, "json_schema", None)
+    if js is None:
+        return False
+    strict = js.get("strict") if isinstance(js, dict) else getattr(js, "strict", None)
+    return bool(strict)
+
+
 def _compile_grammar_for_request(
     engine: BaseEngine,
     structured_outputs=None,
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, this direction makes sense to me for #1241. Keeping the existing 200 behavior while surfacing the downgrade to both operators and clients is a reasonable compatibility-preserving fix.

I found one condition that looks too broad before merge. In `create_chat_completion`, `compiled_grammar is None and response_format` also matches `response_format: {"type": "text"}`, where no grammar enforcement was requested, so the response would incorrectly carry a Warning header saying `response_format` was not enforced.

Could you gate the header/fallback-warning path on a response format that actually maps to grammar-constrained JSON output (`json_object` / `json_schema`) and add a small test that `type: "text"` does not emit a Warning header?

richgoodson:
Addressed the review. I gated the Warning header and prompt-injection fallback on formats that actually map to grammar-constrained JSON output (json_object / json_schema), so a `response_format: {"type": "text"}` no longer carries a header saying it was not enforced. Added a test covering that case.

I derived the gate from `_build_format_element` rather than matching on the type string, so it stays in sync with the existing warn log: a `json_schema` with no schema maps to no grammar and no longer claims "grammar-constrained decoding unavailable" when nothing enforceable was requested.

Full `tests/test_grammar.py` passes (63 passed, 20 skipped where xgrammar is not installed).

jundot:
Thanks for addressing the warning gate. I rechecked the final diff against #1241: the chat completion path now only warns when response_format maps to an enforceable grammar, text and schemaless json_schema stay silent, and the Warning header is attached on both streaming and non-streaming chat responses. CI is green, and this looks good to me. I'll merge this.

PR #1570, opened 2026-05-31
fix(scheduler,cache): materialize bf16 KV on owner thread to end async store-cache SIGABRT

aeyeopsdev:
## Symptom
Concurrent requests on a bf16-KV model abort the server: `Fatal Python error: Aborted` — *"There is no Stream(gpu, N) in current thread"*. Reproduces with ≥2 concurrent finish-and-store requests. Some users also saw wrong-shaped structured output alongside the crash.

## Root cause
The async SSD store-cache **worker** materializes a bf16 `view(uint16)` (`_extract_tensor_bytes`) off the inference thread. MLX generation streams are **thread-local** (`mx.new_thread_local_stream`), so re-dispatching a still-lazy op on the worker targets a stream index that does not exist there → C++ abort. The owner used `mx.async_eval`, which returns before the buffer is concrete, leaving materialization to the wrong thread. The wrong-shaped output was a symptom (in-flight requests returned unenforced when the engine aborted), not a grammar defect.

## Fix
Materialize on the thread that owns the stream:
- `_cleanup_finished`: full `mx.eval(*pre_eval_arrays)` (was `async_eval`) so KV arrays are concrete before the worker slices them.
- `_extract_tensor_bytes`: explicitly `mx.eval` the `uint16` view inside the worker's existing `_mx_buffer_access_lock` window (also closes an `mx.clear_cache` reclaim race).
- In-memory boundary snapshots: eval on the capturing thread (`_eval_snapshot_cache`), closing the asymmetric completion-path hole.

## Trade-offs
The owner thread now blocks on the GPU completion fence it previously deferred; the host memcpy and disk write stay on the worker. No API change, no new flag.

## Verification
4 concurrent × 6 rounds of boundary-crossing, finish-and-store, bf16, structured-output requests: **24/24 completed, 0 aborts, server stays up**. Pre-fix the same workload aborts within the first rounds. This is a Metal thread-local-stream race under concurrent GPU inference, so it isn't capturable by a standard unit test — I verified it with a concurrency stress harness (happy to include it) rather than assert a test that can't exercise the path. `pytest -m "not slow and not integration"` stays green.

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -212,12 +212,23 @@ def _extract_tensor_bytes(arr: mx.array) -> tuple[bytes, str, list[int]]:
     when the source array was evaluated on the inference thread.
 
     For bfloat16 arrays, uses view(uint16) since the buffer protocol does
-    not support bfloat16 directly. The view shares the source buffer; no
-    Metal command is issued by view() alone, and bytes(memoryview(view))
-    on an already-materialized source returns the raw bf16 bits unchanged.
+    not support bfloat16 directly. CAUTION: ``arr.view(mx.uint16)`` is a LAZY
+    MLX op, and MLX streams are thread-local. If ``arr`` is a still-lazy op
+    bound to a stream created on another thread, materializing the view here
+    re-dispatches that op to a stream index that does not exist on the calling
+    thread -> "There is no Stream(gpu, N) in current thread" -> SIGABRT. The
+    contract for cross-thread safety is therefore: every source array reaching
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the detailed repro and fix. The owner-thread materialization approach matches the MLX thread-local stream failure mode, and the remaining issue I found is just the current-main test expectation still patching mx.async_eval.

This looks good to me, and I'm going to merge it. I'll fold the test expectation and stale comment cleanup into a follow-up commit so the post-merge suite tracks mx.eval(*pre_eval_arrays).

PR #1571, opened 2026-05-31
fix(server): fail loud on response_format grammar-compile failure

aeyeopsdev:
## Symptom
A `response_format` `json_schema` that fails to compile — or is requested when xgrammar isn't installed — returns `200` with unconstrained, wrong-shaped output the caller cannot detect.

## Root cause
`_compile_grammar_for_request` returned `None` on the `response_format` path when grammar compilation raised, and when no compiler was available but a schema was requested. `None` means "generate unconstrained," so a structured-output request silently degraded to free-form text.

## Fix
- On compile error, raise `HTTPException(400)` chained from the original xgrammar error (`from e`), matching the existing `structured_outputs` path.
- When xgrammar is unavailable but a non-empty `json_schema` was requested, raise `400` instead of ignoring it.

## Trade-offs / operator-visible change
Behavior change: a request that previously got a degraded `200` now gets a `400` carrying the compiler error. `json_object` (empty schema) and the no-schema case still return `None` (unconstrained), unchanged — only an explicitly requested schema fails loud.

## Verification
`tests/test_grammar.py`: updated the two tests that encoded the old silent-fallback contract to assert the `400`, and added a test for the no-compiler + `json_schema` case. `pytest -m "not slow and not integration" tests/test_grammar.py`: **58 passed, 3 skipped**.

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -2689,6 +2689,18 @@ def _compile_grammar_for_request(
                     "Install with: pip install 'omlx[grammar]'"
                 )
             raise HTTPException(status_code=400, detail=detail)
+        # response_format path: if a non-empty schema was actually
+        # requested, fail loud rather than silently ignoring it.
+        # json_object (empty schema {}) and the no-schema case return None.
+        if fmt.get("json_schema"):
+            raise HTTPException(
+                status_code=400,
+                detail=(
+                    "Structured output requires xgrammar, but it is not "
+                    "available; cannot honor the requested response_format "
+                    "json_schema. Install with: pip install 'omlx[grammar]'"
+                ),
+            )
         return None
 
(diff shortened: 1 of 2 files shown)


aeyeopsdev:
Closing this — upstream has since addressed the silent-degradation problem a different way, so this no longer applies.

`_compile_grammar_for_request` now keeps a `response_format` request on best-effort prompt injection but surfaces the downgrade loudly via `_warn_response_format_not_enforced` (naming an unhonored `strict: true` request), per #1241, while still hard-raising for `structured_outputs`. That closes the observability gap this PR was after, and the choice of degrade-with-warning over a hard 400 on the `response_format` path is a deliberate one I'm happy to defer to.

PR #1576, opened 2026-06-01
fix(cache): suppress spurious "failed to save block" warning

fry69:
This fixes the somewhat confusing warning:

```
omlx.cache.prefix_cache - WARNING - Failed to save block 1 to tiered cache
```

When `Hot Cache Only` is enabled. The commit message below with more detail was generated by my coding agent:

---

When running with `Hot Cache Only` and `Cache Enabled` (hot-cache-only mode with 0B hot cache capacity), every block save attempt is intentionally skipped — there is no storage tier to write to. Previously this emitted a `WARNING`-level `"Failed to save block N to tiered cache"` that appeared to be a real failure.

**Changes:**

1. **`paged_ssd_cache.py`** — the hot-cache-only skip path in `save_block` now logs at `INFO` with `"Hot cache only with no capacity, skipping block ..."`, making the reason visible and clear.

2. **`prefix_cache.py`** — the caller's message is downgraded from `WARNING` to `INFO` and reworded to `"Block N was not persisted to tiered cache"`. This is safe because every `False` return from `save_block` already logs its own specific reason at the appropriate level (error for exceptions, warning for queue full, info for read-only skip).

3. **`tests/test_hot_cache.py`** — added `TestReadOnlyCache` verifying that `save_block` returns `False` and no `WARNING` about tiered cache persistence is emitted in the read-only configuration.

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -2606,14 +2606,4 @@ class PagedSSDCacheManager(CacheManager):
         """
         return self._max_size
 
-    @property
-    def is_read_only(self) -> bool:
-        """
-        True when the cache is configured with no write capacity
-        (hot-cache-only mode with no hot cache capacity).
 
-        In this state, ``save_block()`` intentionally skips persistence
-        and returns ``False``. Callers should treat the "failure" as
-        expected rather than a real error.
-        """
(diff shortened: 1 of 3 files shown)


fry69:
Seems outdated and unnecessary, closing this for now.

PR #1577, opened 2026-06-01
fix(api): trim tool call function names

palvaleri:
## Summary

Trim incidental whitespace from parsed tool-call function names.

## Problem

When using oMLX behind Claude Code's Anthropic `/v1/messages` API, GLM-4.5-Air can emit tool-use names with trailing newlines, for example `Bash\n`, `Read\n`, or `WebSearch\n\n`.

Claude Code treats tool names as exact identifiers, so the response is rejected as:

```text
Error: No such tool available: Bash\n
```

The underlying tool exists as `Bash`; the generated name just contains incidental whitespace.

## Change

Normalize `FunctionCall.name` with `strip()` at the Pydantic model boundary. This keeps downstream OpenAI and Anthropic response conversion from propagating invalid tool names.

## Validation

- Added a unit test covering `FunctionCall(name="Bash\n\n", arguments={})`.
- Ran `uv run --dev pytest tests/test_openai_models.py -v`: 67 passed.

### DIFF
--- a/omlx/api/openai_models.py
+++ b/omlx/api/openai_models.py
@@ -146,6 +146,11 @@ class FunctionCall(BaseModel):
     name: str
     arguments: str  # JSON string
 
+    @field_validator("name", mode="before")
+    @classmethod
+    def _normalize_name(cls, v: Any) -> str:
+        return str(v).strip() if v is not None else ""
+
     @field_validator("arguments", mode="before")
     @classmethod
     def _validate_arguments_json(cls, v: Any) -> str:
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, this fixes the right boundary: the parsed FunctionCall is shared by both OpenAI and Anthropic response conversion, so trimming incidental model whitespace there prevents names like `Bash\n` from leaking to clients.

One small thing I noticed is that the current validator also stringifies non-string values. I don't think that needs to block this PR; I'll merge this and fold a follow-up that only strips actual strings so the existing type validation remains intact.

PR #1588, opened 2026-06-01
fix: lazy-load heavy imports to fix NumPy 2.x compatibility on startup

fparrav:
## Problem

`omlx launch claude` (and any subcommand that doesn't use MLX) crashes on startup when `scipy`/`sklearn` are compiled against NumPy 1.x but NumPy 2.x is installed:

```
ImportError: A module that was compiled using NumPy 1.x cannot be run in NumPy 2.1.3
```

The import chain:
```
omlx/__init__.py → omlx.scheduler → mlx_lm.generate → transformers
→ sklearn → scipy._sparsetools (NumPy 1.x binary) → crash
```

`omlx launch claude` never uses the scheduler or engine — it only touches `omlx/integrations/` — but the eager top-level imports crash the process before any subcommand logic runs.

## Fix

Replace the eager imports in `omlx/__init__.py` with a PEP 562 module-level `__getattr__` that defers loading until a symbol is actually accessed at runtime.

- `omlx launch` / `omlx launch claude` → no MLX stack loaded, no crash
- `omlx serve <model>` → lazy import triggers on first `Scheduler`/`EngineCore` access, works as before
- `from omlx import Scheduler` → still works, just deferred
- `__all__` is unchanged — public API is identical

## Test

```bash
# Before: crashes with NumPy 2.x
# After: works
python -c "import omlx; print(omlx.__version__)"
omlx launch claude

# Lazy load still works
python -c "from omlx import Scheduler; print('OK')"
```

### DIFF
--- a/omlx/__init__.py
+++ b/omlx/__init__.py
@@ -14,17 +14,38 @@ Features:
 
 from omlx._version import __version__
 
-# Continuous batching engine (core functionality, no torch required)
-from omlx.request import Request, RequestOutput, RequestStatus, SamplingParams
-from omlx.scheduler import Scheduler, SchedulerConfig, SchedulerOutput
-from omlx.engine_core import EngineCore, AsyncEngineCore, EngineConfig
-from omlx.cache.prefix_cache import BlockAwarePrefixCache
-from omlx.cache.paged_cache import PagedCacheManager, CacheBlock, BlockTable
-from omlx.cache.stats import PrefixCacheStats, PagedCacheStats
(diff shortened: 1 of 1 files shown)


jundot:
Thanks, this looks good to me. I verified that plain `import omlx` no longer pulls in scheduler/engine_core/transformers, while the documented top-level imports like `from omlx import Scheduler` still work.

I also checked the current tree for `import omlx; omlx.scheduler`-style usage and found no internal caller that would regress. This looks safe to merge.

PR #1590, opened 2026-06-01
fix: normalize 0.0.0.0 bind address to 127.0.0.1 for connections

monroewilliams:
When the server binds to 0.0.0.0, client code that builds connection URLs was using 0.0.0.0 verbatim — a bind wildcard, not a connectable address. This caused macOS ATS errors and broken connections in the menubar poller and SwiftUI screens.

- AppConfig: rename host → bindAddress (stored), add host (computed, normalizes 0.0.0.0 → 127.0.0.1). JSON key migrated to bind_address with backward-compat migration from legacy host key.
- ServerProcess: same pattern — bindAddress stored, host computed. Init and reconfigure take bindAddress: parameter.
- AppServices: use bindAddress for server binding, host for HTTP client.
- ServerScreenVM: normalize dto.server.host for endpoint URL display.
- WelcomeWindow: use bindAddress for server construction, host for URLs.
- MenubarController: simplify displayHost (no longer normalizes).
- PortConflictResolver: unchanged (uses connectable host).
- IntegrationsScreen: simplify formatDisplayHost (no longer normalizes).
- Tests: update for new parameter names and behavior.

### DIFF
--- a/apps/omlx-mac/Sources/App/AppDelegate.swift
+++ b/apps/omlx-mac/Sources/App/AppDelegate.swift
@@ -167,7 +167,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             let runtime = try PythonRuntime.resolve()
             let server = ServerProcess(
                 runtime: runtime,
-                host: config.host,
+                bindAddress: config.bindAddress,
                 port: config.port,
                 basePath: URL(fileURLWithPath: config.basePath, isDirectory: true)
             )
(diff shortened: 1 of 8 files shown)


jundot:
Thanks for tracking this down. The bind-vs-connect split is the right fix for the menubar and SwiftUI connection paths.

I found a couple of follow-up issues: Swift should keep `server.host` as the canonical persisted/admin API key because the Python settings loader still reads that field, and the macOS test fixtures need to be updated for the new `bindAddress` initializer. Since this fixes an urgent user-facing connection bug, I am going to merge this and land a small follow-up on `main` right away to keep the storage schema compatible and tighten the regression tests.

This looks good to me as the core direction, and I am going to merge it with that follow-up.

jundot:
Follow-up is landed on `main` in c07b336. It keeps `server.host` as the canonical persisted/admin API key, treats `bind_address` as a compatibility fallback, passes the raw bind address to the dev server path, and adds Swift/Python regression coverage.

Issue #1549, opened 2026-05-30
/v1/chat/completions accepts logprobs: true / top_logprobs: N but returns empty logprobs.content


snidera3:
## Summary

omlx 0.3.x accepts the OpenAI-compatible `logprobs` and `top_logprobs` parameters in `/v1/chat/completions` requests but the response's `choices[0].logprobs.content` array is always empty. The companion projects `mlx-lm` ([SERVER.md][1]) and `mlx-vlm` both implement this — I think omlx is a natural place to expose it as well, since it's already the OpenAI-API-compatible front-end most homelab deployments seem to be migrating to.

## Repro

Smallest possible repro against an omlx server running any chat model (I'm running Qwen3-4B-Instruct-2507-4bit-DWQ-2510 and Qwen3-VL-8B-Instruct-4bit; the empty-content behavior is the same on both):

```bash
curl -s http://127.0.0.1:8800/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{
    "model": "Qwen3-4B-Instruct-2507-4bit-DWQ-2510",
    "messages": [{"role": "user", "content": "Say hi in one sentence."}],
    "max_tokens": 40,
    "logprobs": true,
    "top_logprobs": 1
  }' | jq '.choices[0] | {finish_reason, message, logprobs}'
```

Expected: `.choices[0].logprobs.content` is a non-empty array of `{token, logprob, top_logprobs}` objects per the [OpenAI Chat Completions logprobs spec][2].

Observed: `logprobs.content` is `[]` (or `null` in some configurations) regardless of request parameters. The text response itself is normal.

## Why I care

I'm using a batch perimeter-summary pipeline on a Mac Studio that wants a calibrated confidence signal aligned to specific JSON-field tokens (e.g. the headline of a summary) rather than the model's self-reported `confidence: 0.95` scalar — three recent papers ([arxiv 2604.02543][3], [arxiv 2509.25532][4], [arxiv 2510.26995][5]) document Qwen3-VL verbalized-confidence ECE around 0.42, so the scalar is functionally noise. Field-aligned mean-logprob is the empirically-validated alternative.

[...]

PR #1591, opened 2026-06-02
feat(api): OpenAI-compatible logprobs for /v1/chat/completions (#1549)

popfido:
Adds per-token logprobs to `/v1/chat/completions` for supported streaming and non-streaming generation paths, addressing #1549. Requests can opt into `logprobs` and `top_logprobs`; the response includes the selected token probability and requested top candidates when they can be aligned reliably with returned content.

The scheduler reduces sampling distributions to compact payloads, forwards them through the text/VLM engines, and preserves token order across collector merges. The request range is 0–20, with a configurable server cap. Requests that do not enable logprobs keep the existing generation path.

Logprobs are omitted when alignment or target-model probabilities cannot be established: speculative/protocol-parser paths, active tool filtering, rewritten non-streaming content, and ambiguous streaming segments. Mixed supported/unsupported chunks preserve all response text. Token byte payloads are null for ambiguous replacement-character decoding rather than reporting fabricated UTF-8 bytes.

Validation: 1,539 local tests passed across API/schema/settings, extraction and serialization, server/scheduler/collector, thinking/tool parsing, streaming cancellation, and text/VLM engine suites. Regression cases include partial-tag buffering, tokens spanning markup and content, missing logprobs in aggregated output, and split Unicode byte fallback. `git diff --check` passes. Fresh GitHub CI runs on the rebased head; live-model performance was not remeasured for this revision.

### DIFF
--- a/omlx/api/openai_models.py
+++ b/omlx/api/openai_models.py
@@ -11,9 +11,15 @@ These models define the request and response schemas for:
 """
 
 import json
-from typing import Any, Dict, List, Optional, Union
+from typing import Any
 
-from pydantic import AliasChoices, BaseModel, Field, field_validator
+from pydantic import (
+    AliasChoices,
+    BaseModel,
+    Field,
+    field_validator,
+    model_validator,
+)
 
 from omlx.api.shared_models import (
     BaseUsage,
@@ -31,7 +37,7 @@ class ImageURL(BaseModel):
     """Image URL or base64 data URI for vision model input."""
 
     url: str  # "https://..." or "data:image/jpeg;base64,..."
-    detail: Optional[str] = "auto"  # "low", "high", "auto"
+    detail: str | None = "auto"  # "low", "high", "auto"
 
 
 class InputAudio(BaseModel):
@@ -48,11 +54,11 @@ class FileContent(BaseModel):
     ``data`` is accepted as an oMLX legacy alias for dashboard clients.
     """
 
-    filename: Optional[str] = None
-    mime_type: Optional[str] = None
-    file_data: Optional[str] = None
-    data: Optional[str] = None
-    file_id: Optional[str] = None
+    filename: str | None = None
+    mime_type: str | None = None
+    file_data: str | None = None
+    data: str | None = None
+    file_id: str | None = None
 
 
 class ContentPart(BaseModel):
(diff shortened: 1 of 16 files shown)


omlx/api/openai_models.py
```diff
@@ -317,29 +326,60 @@ def coerce_stop(cls, v):
             return [v]
         return v
 
+    @model_validator(mode="after")
+    def _validate_logprobs(self):
```
OliverwStevens, line 330: Do we need to limit it to 20 logprobs? I think that a higher upper bound would be nice to have.

tests/test_logprobs_extraction.py
OliverwStevens, line 1: I think that we could make some of the tests in this file a little more readable / DRY them up.

omlx/api/openai_models.py
```diff
@@ -48,11 +54,11 @@ class FileContent(BaseModel):
     ``data`` is accepted as an oMLX legacy alias for dashboard clients.
     """
 
-    filename: Optional[str] = None
-    mime_type: Optional[str] = None
-    file_data: Optional[str] = None
-    data: Optional[str] = None
-    file_id: Optional[str] = None
+    filename: str | None = None
```
OliverwStevens, line 57: It seems like that were a lot of changes for the formatting here. Is that something we want to do?

OliverwStevens (commented):
A few things to consider

andreipopovici:
A lot of users would be very happy if this makes it into v0.7.0 😍🙏

cosgroveb:
Mildly interested in trying https://ai-tldr.dev/tools/semif/ with omlx. This might do it?

oktolabsai-developer:
Opened a small follow-up against this PR’s branch: https://github.com/popfido/omlx/pull/2

An isolated live test with Qwen3.8-27B-AWQ-5.0bpw found that explicitly disabling MTP still omitted logprobs: the VLM loader retains an inactive head for weight loading, while the guard checks head presence. The follow-up reuses the existing decode-enabled helper and also normalizes low-precision logprob metadata in FP32; the original top-ten probability mass exceeded one in the live test.

1,486 relevant tests passed. An installed-wheel test returned matching streaming/non-streaming logprobs with MTP off and continued suppressing them with MTP on. The linked PR includes scope and limitations. No production installation was modified.

PR #1596, opened 2026-06-02
feat: add Hermes tool call parser

scubamount:
## Summary

Adds fallback support for the Hermes-style/Liquid tool call wire format:

```text
<|tool_call_start|>[function_name(arg=value)]<|tool_call_end|>
```

Some clients/agents emit tool calls using these markers instead of oMLX's currently supported XML/prose bracket formats. Without this fallback, oMLX can surface raw marker text to users instead of returning structured tool calls.

This PR:
- parses Hermes-style JSON payloads inside `<|tool_call_start|>...<|tool_call_end|>`
- parses Python-expression-like bracket payloads such as `[read_file(path='/tmp/a.log')]`
- supports multiple calls in one Hermes block via AST parsing, e.g. `[tool_a(...), tool_b(...)]`
- suppresses Hermes start/end markers during streaming even when tokenizer metadata is absent
- suppresses orphan/split Hermes closing markers so `<|tool_call_end|>` cannot leak if it arrives without a visible open marker
- strips any remaining complete Hermes envelopes from final assistant text if parsing fails
- classifies LFM2 text checkpoints as LLMs instead of embeddings by treating `lfm2` as ambiguous and relying on architecture/name checks

## Verification

```text
/opt/homebrew/opt/python@3.12/libexec/bin/python3 -m pytest tests/test_tool_calling.py -q
161 passed in 0.16s

/opt/homebrew/opt/python@3.12/libexec/bin/python3 -m pytest tests/test_model_discovery.py -q
119 passed in 0.16s
```

### DIFF
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -874,6 +874,10 @@ class ToolCallStreamFilter:
             idx = text.find(marker)
             if idx >= 0:
                 starts.append((idx, len(marker), close))
+            if close:
+                close_idx = text.find(close)
+                if close_idx >= 0:
+                    starts.append((close_idx, len(close), None))
 
         ns_match = self._namespaced_open_re.search(text)
         if ns_match:
@@ -937,6 +941,8 @@ class ToolCallStreamFilter:
         keep = 0
         for marker, _close in self._marker_pairs:
             keep = max(keep, self._partial_prefix_len(text, marker))
+            if _close:
+                keep = max(keep, self._partial_prefix_len(text, _close))
 
         last_lt = text.rfind("<")
         if last_lt >= 0:
@@ -977,6 +983,8 @@ class ToolCallStreamFilter:
 
         for marker, _close in self._marker_pairs:
             if marker.startswith(tail):
+                return True
+            if _close and _close.startswith(tail):
                 return True
 
         # Drop unresolved bracket tool-call prefixes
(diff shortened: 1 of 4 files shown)


scubamount:
Cross-ref: this PR is the server-side counterpart to [NousResearch/hermes-agent#37229](https://github.com/NousResearch/hermes-agent/pull/37229) (fix: normalize local tool JSON fallbacks).

These two PRs are **co-dependent**:
- **omlx#1596** (this PR): parses Hermes-style `<|tool_call_start|>...<|tool_call_end|>` markers at the inference server layer, so OMLX returns structured tool calls instead of raw marker text.
- **hermes-agent#37229**: normalizes local-model assistant JSON with commands/tools_used into OpenAI-style tool calls on the hermes-agent client side.

Without both, local model tool calling is broken:
- Without omlx#1596 → OMLX leaks raw Hermes markers to hermes-agent
- Without hermes-agent#37229 → hermes-agent cannot normalize the mixed JSON/bracket payloads that local models emit

Both should merge together for a complete fix.

jundot:
Thanks for the PR. I checked the Hermes fallback path, streaming marker suppression, and the LFM2 discovery change against the existing parser/model-discovery call paths.

The AST-based bracket parser covers the multi-call case without the old comma-splitting risk, and the streaming filter keeps the Hermes markers out of visible deltas when tools are enabled. The focused tool-calling, model discovery, tokenizer/audio discovery, and adapter tests all pass on the PR branch.

PR #1597, opened 2026-06-02
Respect tool_choice none when MCP is configured

### LINKED ISSUE #1583: OpenAI-compatible chat completions ignore tool_choice: "none" when MCP is configured
**Describe the bug**

When oMLX is started with a global MCP config, `/v1/chat/completions` injects MCP tools into the chat template even when the request explicitly sets `tool_choice: "none"`.

This makes OpenAI-compatible clients unable to opt out of tool use on a per-request basis. In my case, Open WebUI has its own web search flow and sends/should send `tool_choice: "none"` to the oMLX backend, but oMLX still exposes MCP tools to the model. The model then returns `tool_calls` such as `MCP_DOCKER__...` instead of a normal assistant message.

**To Reproduce**

1. Configure oMLX with MCP enabled, for example:

```json
{
  "servers": {
    "MCP_DOCKER": {
      "transport": "stdio",
      "command": "docker",
      "args": ["mcp", "gateway", "run", "--profile", "default"],
      "enabled": true,
      "timeout": 120
    }
  }
}
```

2. Start oMLX with that MCP config active.

3. Send a chat completion request with `tool_choice: "none"`:

```bash
curl -sS http://127.0.0.1:8000/v1/chat/completions \
  -H "Authorization: Bearer $OMLX_API_KEY" \
  -d '{
    "model": "Qwen3.6-35B-A3B-Uncensored-Heretic-MLX-6bit",
    "messages": [
      {
        "role": "user",
        "content": "search for recent news about Open WebUI and cite the sources"
      }
    ],
    "stream": false,
    "max_tokens": 300,
    "tool_choice": "none"
  }'
```

4. Observe that the response can still contain `tool_calls`, for example:

```json
{
  "message": {
    "role": "assistant",
    "tool_calls": [
      {
        "type": "function",
        "function": {
          "name": "MCP_DOCKER__mcp-find",
          "arguments": "{\"query\":\"web search\"}"
        }
      }
    ]
  },
  "finish_reason": "tool_calls"
}
```

**Expected behavior**

When a request sets:

```json
"tool_choice": "none"
```

oMLX should not inject MCP tools into the chat template for that request, and the model should not be able to return MCP `tool_calls`.

MCP should remain enabled globally for other requests that omit `tool_choice` or use `tool_choice: "auto"`.

**Observed behavior**

Even with `tool_choice: "none"`, MCP tools are merged into the effective tool list and passed into the chat template.

In `server.py`, the current behavior appears to merge MCP tools whenever `_server_state.mcp_manager` exists:

```python
effective_tools = request.tools
if _server_state.mcp_manager:
    user_tools_dicts = [t.model_dump() for t in request.tools] if request.tools else None
    effective_tools = _server_state.mcp_manager.get_merged_tools(user_tools_dicts)
```

A local patch that fixed the issue for me was:

```python
tools_disabled = request.tool_choice == "none"
effective_tools = None if tools_disabled else request.tools
if _server_state.mcp_manager and not tools_disabled:
    user_tools_dicts = [t.model_dump() for t in request.tools] if request.tools else None
    effective_tools = _server_state.mcp_manager.get_merged_tools(user_tools_dicts)
```

After this change:

- Requests with `tool_choice: "none"` return normal assistant text.
- Requests without `tool_choice` still receive MCP tools and can produce `tool_calls`.
- Global MCP functionality remains intact.

**Environment**

- macOS Version: macOS Sequoia / Apple Silicon
- oMLX Version: 0.3.12

**Additional context**

This matters for OpenAI-compatible clients such as Open WebUI that may provide their own RAG/web-search context and want the model backend to behave as a plain text generator for that request.

Related but not duplicate:

- #681 discusses lazy tool loading and OpenAI `tool_choice` restrictions.
- #1472 discusses constrained decoding for tool names.
- #906 discusses disabling tool-call parsing.

lvsijian8:
## Problem
When a global MCP manager is configured, chat completions always merge MCP tools into the effective tool list. Requests with `tool_choice: "none"` still receive tools in the chat template, so OpenAI-compatible clients cannot opt out of tool use per request.

## Fix
- Treat `tool_choice: "none"` as a request-level tool disable switch.
- Skip MCP tool merging and pass no tools to the chat template for that request.
- Keep MCP tool injection for requests that omit `tool_choice` or set it to `"auto"`.
- Cover the `tool_choice: "none"` case with and without request-provided tools.

## Tests
- `.venv/bin/pytest tests/integration/test_server_endpoints.py::TestChatCompletionEndpoint::test_chat_completion_tool_choice_controls_mcp_tools -q`
- `.venv/bin/pytest tests/integration/test_server_endpoints.py::TestChatCompletionEndpoint -q`
- `.venv/bin/pytest tests/integration/test_server_endpoints.py -q`
- `git diff --check`

Fixes #1583

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -2272,9 +2272,11 @@ async def create_chat_completion(
         if json_instruction:
             messages = _inject_json_instruction(messages, json_instruction)
 
-    # Merge MCP tools with user-provided tools
-    effective_tools = request.tools
-    if _server_state.mcp_manager:
+    # Merge MCP tools with user-provided tools unless the request explicitly
+    # disables tool use.
+    tools_disabled = request.tool_choice == "none"
+    effective_tools = None if tools_disabled else request.tools
+    if _server_state.mcp_manager and not tools_disabled:
         # Convert Pydantic ToolDefinition models to dicts for merge_tools
         user_tools_dicts = [t.model_dump() for t in request.tools] if request.tools else None
         effective_tools = _server_state.mcp_manager.get_merged_tools(user_tools_dicts)
--- a/tests/integration/test_server_endpoints.py
+++ b/tests/integration/test_server_endpoints.py
@@ -794,6 +794,114 @@ class TestChatCompletionEndpoint:
         assert message["tool_calls"][0]["function"]["arguments"] == '{"city": "SF"}'
         assert data["choices"][0]["finish_reason"] == "tool_calls"
 
+    @pytest.mark.parametrize(
+        ("tool_choice", "request_tools", "expect_tools"),
+        [
+            (None, None, True),
+            ("auto", None, True),
+            ("none", None, False),
+            (
+                "none",
+                [
+                    {
+                        "type": "function",
+                        "function": {
+                            "name": "user_search",
+                            "description": "Search from request tools",
+                            "parameters": {
+                                "type": "object",
+                                "properties": {
+                                    "query": {"type": "string"},
+                                },
+                            },
+                        },
+                    }
+                ],
+                False,
+            ),
+        ],
+    )
+    def test_chat_completion_tool_choice_controls_mcp_tools(
+        self,
+        client,
+        mock_llm_engine,
+        tool_choice,
+        request_tools,
+        expect_tools,
+    ):
+        """tool_choice='none' should suppress request and globally configured tools."""
+        from omlx.server import _server_state
+
+        class RecordingMCPManager:
+            def __init__(self):
+                self.calls = []
+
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for the fix. I verified this against #1583: tool_choice: "none" now skips MCP tool merging and keeps tools out of both token counting and the generation kwargs, while omitted/auto tool_choice still keeps MCP injection enabled.

The focused chat completion coverage and the full server endpoint integration file pass locally. This looks good to me, and I am going to merge it.

PR #1601, opened 2026-06-02
Disable paged cache when SSD cache init fails

### LINKED ISSUE #1455: RFE: Allow starting/working with unwriteable cache directory
Use case:

Using an external NVMe drive for cache to prevent wear on the internal soldered drive.
But when on the road, this drive might not be connected and one needs to edit the settings manually to even start oMLX.

It would be great for oMLX to tolerate the cache directory being unwriteable and fallback to not using L2 cache in that case.
Or even better, to have this behaviour configurable with or without fallback to another directory, but this might be to niche to be worth implementing.

Workarounds could be done outside oMLX to have macOS auto re-symlink the cache directory using launchd or do overlay mounts with rclone or some other hackish solution, but I'm not completely sure how well this would work (I know it survives me deleting cache without restarting it, but there are also other "temp" directories in there).

lvsijian8:
## Problem

When paged SSD cache is configured but the backing directory is unavailable or not writable, scheduler startup can leave partial cache state behind. `PagedSSDCacheManager` fails, but `PagedCacheManager`, `BlockAwarePrefixCache`, and the store-cache executor can still remain active.

This is easy to hit when the cache directory points to an external drive that is not currently mounted.

## Root cause

`_init_tiered_cache()` did not report whether paged SSD cache initialization succeeded. The scheduler therefore continued setting up the cache pipeline even after initialization failed.

## Fix

- Make `_init_tiered_cache()` return whether setup succeeded.
- Only attach the cold-restore callback and start the store-cache executor when setup succeeds.
- Clear paged cache runtime state when setup fails, so the scheduler falls back to normal no-cache operation for that run.

## Tests

- `.venv/bin/pytest tests/test_scheduler.py::TestSchedulerInitialization::test_init_falls_back_when_paged_ssd_cache_unavailable -q`
- `.venv/bin/pytest tests/test_scheduler.py -q`
- `.venv/bin/pytest tests/test_scheduler.py::TestSchedulerConfig tests/test_cache_factory.py tests/test_paged_ssd_cache.py::TestPagedSSDCacheManager::test_initialization -q`
- `.venv/bin/pytest tests/test_cache_factory.py tests/test_paged_ssd_cache.py::TestPagedSSDCacheManager::test_initialization -q`
- `.venv/bin/ruff check --select I001,F401,F811 tests/test_scheduler.py`
- `git diff --check`

Fixes #1455

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -906,11 +906,13 @@ class Scheduler:
                 paged_cache_manager=self.paged_cache_manager,
             )
 
-            # Initialize paged SSD cache
-            self._init_tiered_cache()
+            # Initialize paged SSD cache. If the backing directory is not
+            # usable (for example, an external cache drive is disconnected),
+            # continue with cache disabled instead of leaving partial state.
+            cache_initialized = self._init_tiered_cache()
 
             # Set cold restore callback for prefix cache
-            if self.paged_ssd_cache_manager is not None:
+            if cache_initialized and self.paged_ssd_cache_manager is not None:
                 self.block_aware_cache.set_cold_restore_callback(
                     self._restore_block_from_cold
                 )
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, this looks good to me. The change makes the SSD cache initialization outcome explicit and avoids leaving prefix-cache/store-cache state active after PagedSSDCacheManager fails, so the scheduler falls back to the normal no-cache path for #1455. I checked the affected request and cleanup paths and ran the scheduler tests. This looks safe to merge.

PR #1605, opened 2026-06-02
fix (tests): avoid log sniffing in test_sanitize_no_mtp_weights

Kistaro:
Split from https://github.com/jundot/omlx/pull/1562 and following up on https://github.com/jundot/omlx/pull/1538#event-26154504538.

Stops log sniffing in `test_sanitize_no_mtp_weights` and instead verifies the intended observable behaviors:

* doesn't explode
* doesn't emit something that says it has mtp mode
* _does_ emit model slugs that imply it emitted the stuff it was expected to


jundot:
Thanks for cleaning this up. Checking the observable sanitize output is a better invariant than matching a debug log string, and it still covers the no-MTP-weights case without weakening the behavior check.

PR #1606, opened 2026-06-02
fix: support comma-separated bind addresses in host setting

fqx:
## Problem

#1526 introduced a pre-bind approach (`uvicorn.Config + bind_socket()`) to fail fast on port conflicts before model preload. This broke multi-address binding: `bind_socket()` calls `sock.bind((host, port))` which requires a plain string, not a list — so users who had `host: [127.0.0.1, 192.168.1.x]` in their settings (or set a comma-separated value) now get a `TypeError` on startup.

Supersedes #908, which patched the old `uvicorn.run()` call that no longer exists.

## Fix

**`omlx/settings.py`** — normalise legacy YAML list values (`host: [a, b]`) to a comma-separated string on load. Backward-compatible: single-string configs are unchanged.

**`omlx/cli.py`** (`serve_command`) — split the host string by comma, bind one socket per address (preserving the fail-fast guarantee from #1526), then pass all sockets to `Server.run(sockets=[...])`. uvicorn accepts a socket list and listens on all of them.

**`omlx/cli.py`** (`launch_command`) — pick the first bind address as the connect host; normalise wildcards (`0.0.0.0`, `::`) to `127.0.0.1`.

**`omlx/utils/network.py`** — add `is_valid_bind_host()`: accepts all valid IPs including `0.0.0.0` and `::`, unlike `is_valid_alias()` which rejects unspecified addresses.

**`omlx/admin/routes.py`** — validate each comma-separated part on save, returning a readable 400 before an invalid value can crash the server on restart.

**`omlx/admin/static/js/dashboard.js`** — extract `.msg` from Pydantic error objects instead of joining raw objects (fixes `[object Object]` display). Two call sites.

**`omlx/admin/i18n/{en,zh,ja,ko,zh-TW}.json`** — add `host_multi_hint` string in all five languages; drop "(caution)" from `host_custom` label now that entering multiple addresses is a valid normal use case.

**`omlx/admin/templates/dashboard/_settings.html`** — show the `host_multi_hint` hint below the custom host input when custom mode is active.

## Usage

Enter multiple bind addresses separated by commas in the Host field:

```
127.0.0.1, 100.64.0.1
```

Existing single-address settings and legacy YAML list values both continue to work without migration.

## Testing

1. Set host to `127.0.0.1, ::1` in the web UI and save — both addresses should appear in startup output
2. Start with a settings.json containing `"host": ["127.0.0.1", "192.168.1.x"]` — should load cleanly and show both bind lines
3. Enter an invalid value (e.g. `not a host!!`) — should get a readable 400 error, not `[object Object]`
4. Single-address configs continue to work unchanged
5. `omlx launch` with a multi-address host connects via the first address


jundot:
Thanks for fixing this regression. The multi-socket uvicorn approach looks right to me, but I found a couple of things that need cleanup before I can merge it.

The PR currently conflicts with main. Also, `is_valid_bind_host()` still accepts IP-shaped invalid values like `999.999.999.999` as hostnames, and `detect_server_aliases()` should handle comma-separated bind hosts so loopback aliases are not dropped for values like `127.0.0.1, ::1`.

Could you rebase on current main and cover these cases with focused tests?

fqx:
@jundot — rebased on current main and addressed all three points from your review.

**`is_valid_bind_host()` — IP-shaped invalid values**

Added a fast-path guard that rejects strings like `999.999.999.999` before they reach the hostname check. Also tightened `is_valid_hostname()` at the root: for any dotted name, the rightmost label must contain at least one letter. IANA has never delegated a numeric TLD, so an all-digit last label is a reliable signal of an IP-shaped string. This mirrors the approach used by the `validators` PyPI library. Single-label names (no dots) are exempt since they are local hostnames with no TLD constraint.

**`detect_server_aliases()` — comma-separated bind hosts**

Replaced the single-value `in` check with a set intersection against comma-split parts, so `host="127.0.0.1, ::1"` now correctly includes loopback aliases.

**Tests**

Added focused tests covering: valid IPv4/IPv6 (including unspecified addresses and IPv4-mapped IPv6), valid hostnames, IP-shaped-but-invalid values of various octet counts, the new `is_valid_hostname` last-label behaviour, and all three comma-separated `detect_server_aliases` scenarios.

jundot:
Thanks for rebasing and tightening this. The multi-socket bind path and alias handling look good to me now, and I'm going to merge it.

I noticed one small i18n consistency issue: `settings.server.host_multi_hint` is missing from `es`, `fr`, and `ru`, which can show the raw key in those locales. I'll fold that into a small follow-up after merge.

PR #1616, opened 2026-06-02
Feature: Apply profile settings on the fly

ilukashin:
## Concept
Allow user to send requests to specific model-profile setup without reloading model.

## How it works

Given:
`qwen3` - original model name, installed with oMLX.

In model settings it have two profiles with specific settings:
- `code` profile with thinking=off, temperature = 0.1
- `chat` profile with thinking=on
- `alias` = my_favorite_model

Then:
When i write client config, in model name i can provide values like:
```
qwen3:code
qwen3:chat
my_favorite_model:code
my_favorite_model:chat
```

## UI changes
Added dropdown if model have profiles. Inspired by https://github.com/jundot/omlx/issues/341.

Снимок экрана — 2026-06-01 в 20 08 06

updated. Now theres also settings specified for each profile.
Снимок экрана — 2026-06-03 в 22 53 21

## TODO

- [x] Check test coverage for future regression
- [x] Update README for different languages
- [x] Add settings specified for each profile on its row (UI)


ilukashin:
I see it have conflicts now. If theres any chance to be merged let me know and i'll fix this.

jundot:
Thanks for working on this. I like the direction, but I cant merge it as-is because it currently conflicts with main in omlx/server.py.

I also want to clarify the user-facing contract for the suffix. With this PR, `model:code` appears to resolve against the profile `display_name`, while the existing profile system uses the internal `name` as the canonical id for apply/update/delete and `active_profile_name`.

Using `display_name` may be the right UX since that is what users see in the admin UI, but then it needs defined behavior for uniqueness, renames, casing, and spaces. Otherwise two profiles could have the same visible name, or a rename could silently change the API model string.

I also found that the suffix path returns only the profile settings instead of merging them over the base model settings, and the chat/Anthropic/Responses template settings still read the base model settings, so profile enable_thinking/chat_template_kwargs wont apply.

I will think a bit more about the best UX here too. If you have a good idea for making this clear and stable for users, please suggest it.

ilukashin:
Thanks for feedback!
Glad to hear you agreed with feature.
I can continue to work on this after we complete our decisions about corner cases you have mentioned above.

At first view theres two obvious options about use `display_name` or  internal `name`:
1. improve `Profiles` UX with `display_name` validation, like uniqueness=true, downcase on save, etc.
2. Add to UI view `profile.id` to allow users access to it. But this id is not representative, maybe just only copy-to-clipboard button on the same profile row.

In my opinion - first option is more correct. Especially im not sure why user need two profiles with same names. But this cost more time to develop.
Second option easier to implement. 
In both keys it can be button with copy-to-clibpoard prepared string.

> I also found that the suffix path returns only the profile settings instead of merging them over the base model settings

As far as i understand, theres no need to merge profile settings with base model settings.
In my opinion: if i send request with profile suffix - i expect exactly this profile settings have applied.

And about chat/anthropic looks like i need to explore a bit more, because i'm only openai-api user and was focused only on this context.

PR #1619, opened 2026-06-02
feat: add Oh My Pi (omp) launcher integration

ahmedyounes:
## What

Adds an **Oh My Pi** tile to the dashboard's Applications list that launches the [oh-my-pi](https://github.com/can1357/oh-my-pi) (`omp`) coding agent against the running oMLX server — a sibling of the existing **Pi** tile. Non-destructive: the Pi tile is unchanged.

## Why

`omp` is a popular fork of pi-coding-agent with a coding-first workflow. Users who run `omp` instead of vanilla `pi` had no one-click way to point it at their local oMLX server.

## How it works

`omp` is a separate binary from `pi` with its own config dir (`~/.omp/agent`), but it:
- honors the same `PI_CODING_AGENT_DIR` override, and
- reads custom OpenAI-compatible providers from `~/.omp/agent/models.yml` using the **same `providers.<name>` schema** as pi — just YAML instead of JSON.

`OhMyPiIntegration` therefore reuses `PiIntegration`'s provider schema but writes/merges `models.yml` (YAML) with the same read→backup→merge→write pattern as the Hermes integration, preserving any other providers the user already has, then execs `omp --model omlx/<model>`. (omp only auto-migrates `models.json`→`models.yml` when the latter is absent, so writing YAML directly is the robust path across relaunches.)

## Changes

- **New** `omlx/integrations/omp.py` — `OhMyPiIntegration`; registered in `omlx/integrations/__init__.py`
- `omlx/settings.py` + `omlx/admin/routes.py` — `omp_model` setting (parity with the other tools)
- `omlx/admin/templates/dashboard/_status.html` — new tile; `omlx/admin/static/js/dashboard.js` — `ompCommand` getter + save/default wiring
- `omlx/admin/static/img/integrations/omp.svg` — icon (theme-aware π glyph matching the existing set)
- `omlx/cli.py` — `launch` help text
- `tests/test_integrations.py` — `TestOhMyPiIntegration` + registry assertions

## Testing

- `pytest tests/test_integrations.py` → **101 passed** (13 new omp cases mirroring the Pi suite + registry updates).
- Verified end-to-end against the real `omp` v15.8.0: the generated `~/.omp/agent/models.yml` is accepted and `omp --list-models omlx` lists the configured model.
- Ran the server from source and confirmed the **Oh My Pi** tile renders in `/admin/dashboard`, and `omlx launch list` reports `omp`.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -276,6 +276,7 @@ class GlobalSettingsRequest(BaseModel):
     integrations_openclaw_model: str | None = None
     integrations_hermes_model: str | None = None
     integrations_pi_model: str | None = None
+    integrations_omp_model: str | None = None
     integrations_openclaw_tools_profile: (
         Literal["minimal", "coding", "messaging", "full"] | None
     ) = None
@@ -2885,6 +2886,7 @@ async def get_global_settings(is_admin: bool = Depends(require_admin)):
             "openclaw_model": global_settings.integrations.openclaw_model,
             "hermes_model": global_settings.integrations.hermes_model,
             "pi_model": global_settings.integrations.pi_model,
+            "omp_model": global_settings.integrations.omp_model,
             "copilot_model": global_settings.integrations.copilot_model,
             "openclaw_tools_profile": global_settings.integrations.openclaw_tools_profile,
         },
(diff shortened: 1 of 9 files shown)


jundot:
Thanks for adding this. I checked the OMP models.yml shape against the current upstream docs, and the integration approach looks right. The PR-head integration tests pass locally (`python -m pytest tests/test_integrations.py`, 102 passed).

Main has moved since this branch, so it now conflicts in `omlx/admin/routes.py` and `omlx/admin/static/js/dashboard.js`. Please rebase and preserve the current MarkItDown settings fields while adding the OMP fields.

One small fix: `OhMyPiIntegration.get_command()` should use `get_cli_command_prefix()` instead of `get_cli_prefix()` so app-bundle CLI paths with spaces stay shell-safe.

jesperronn:
i could use this. @ahmedyounes could you rebase the code as @jundot asked so that the code can move forward?

ChillarAnand:
@ahmedyounes Did you get a chance to look into this?

PR #1622, opened 2026-06-03
fix(scheduler): preserve prefill error outputs during decode

### LINKED ISSUE #1621: Request stalls after `Prefill failed for <req-id>`
**Describe the bug**
A request can stall indefinitely after this scheduler log appears:

```text
Prefill failed for <req-id>: <error>
```

The affected request does not fail at the API/client layer and does not generate any tokens. It appears to disappear from scheduler state without a terminal error output reaching `engine_core`.

The likely root cause is in `Scheduler.step()`: rejected/error outputs produced during scheduling are appended to `output.outputs`, but later decode responses from already-running requests replace the whole list with `output.outputs = outputs`. If a prefill failure and decode output happen in the same scheduler step, the terminal prefill error output is dropped.

Relevant code path:

Reference: `omlx/scheduler.py:5628`

```python
# External prefill failure path
rejected_outputs.append(
    RequestOutput(
        request_id=request.request_id,
        finished=True,
        finish_reason="error",
        error=str(e),
    )
)
```

References: `omlx/scheduler.py:6554` and, in the unfixed version, `omlx/scheduler.py:6573`

```python
# Scheduler.step()
if rejected:
    output.outputs.extend(rejected)
    output.has_work = True

...

if responses:
    outputs, finished_ids = self._process_batch_responses(responses)
    output.outputs = outputs
    output.finished_request_ids = finished_ids
    self._cleanup_finished(finished_ids)
```

That final assignment drops the previously appended rejected output. Since the failed prefill request has already been removed from `self.requests`, `waiting`, `prefilling`, and `running`, there is no later opportunity to emit tokens or surface an error.

`engine_core` only wakes the caller when it receives a finished output and sets the finished event:

Reference: `omlx/engine_core.py:253`

```python
if req_output.finished:
    event = events.get(rid)
    if event:
        event.set()
```

If the scheduler drops the terminal error output, non-streaming callers waiting on `event.wait()` can hang indefinitely.

**To Reproduce**

1. Start at least one request that is already running and producing decode tokens.
2. Submit another request that fails during external prefill and reaches the `Prefill failed for <req-id>` path in `omlx/scheduler.py`.
3. Make the prefill failure occur in the same scheduler step where the already-running request returns at least one decode response.
4. Observe that the failed request does not receive a terminal `finish_reason="error"` output and remains stalled.

The key condition is a scheduler step that contains both:

- a `rejected_outputs` entry from prefill failure
- decode `responses` from `batch_generator.next_generated()`

**Expected behavior**
The client should receive a terminal error output when prefill fails and the request will not be retried:

Expected output shape from the failure path at `omlx/scheduler.py:5628`:

```python
RequestOutput(
    request_id=<req-id>,
    finished=True,
    finish_reason="error",
    error=<prefill error>,
)
```

`engine_core` should then set the finished event for the request so streaming and non-streaming callers exit cleanly.

A possible fix is to preserve outputs already added to `SchedulerOutput` when decode responses are processed:

Suggested replacement at `omlx/scheduler.py:6573`:

```python
if responses:
    outputs, finished_ids = self._process_batch_responses(responses)
    output.outputs.extend(outputs)
    output.finished_request_ids.update(finished_ids)
    self._cleanup_finished(finished_ids)
```

**Screenshots**
Not applicable.

 - macOS Version: 26.5
 - oMLX Version 0.4.0

**Additional context**
This may affect more than the exact `Prefill failed for ...` path. Any rejected output appended before decode processing can be lost by `output.outputs = outputs`, including:

- preflight memory guard rejections
- chunked prefill failures
- non-chunked external prefill failures

Observed consequence: the scheduler has removed the failed request, but the engine collector and finished event for that request remain active, so the client waits without receiving tokens or an error.

ken-zzzzz:
Closes #1621

## Summary

- Preserve scheduler rejection outputs when decode responses are produced in the same `step()`.
- Add a regression test covering a prefill failure output followed by decode output in the same scheduler step.

## Why

When prefill failed, `_schedule_waiting()` could return a terminal error `RequestOutput` for the failed request. If another request produced decode output in the same `Scheduler.step()`, the decode path replaced `output.outputs` instead of appending to it, dropping the prefill error output.

That left the failed request removed from scheduler state but without a finished output delivered to `engine_core`, so the client could wait indefinitely with no tokens and no error.

## Changes

- Changed `Scheduler.step()` to append decode outputs with `output.outputs.extend(outputs)`.
- Changed finished request ID handling to merge with `output.finished_request_ids.update(finished_ids)`.
- Added `TestSchedulerStepOutputs.test_decode_outputs_preserve_prefill_rejections`.

## Tests

```bash
.venv/bin/python -m pytest tests/test_scheduler.py -q
```

Result:

```text
105 passed
```

Also verified the new regression test fails when the scheduler fix is reverted.

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -6572,8 +6572,8 @@ class Scheduler:
 
                 if responses:
                     outputs, finished_ids = self._process_batch_responses(responses)
-                    output.outputs = outputs
-                    output.finished_request_ids = finished_ids
+                    output.outputs.extend(outputs)
+                    output.finished_request_ids.update(finished_ids)
                     self._cleanup_finished(finished_ids)
 
                     # Periodic Metal allocator cleanup during long decodes.
--- a/tests/test_scheduler.py
+++ b/tests/test_scheduler.py
@@ -12,7 +12,7 @@ Tests cover:
 - get_request(): request lookup
 - get_stats(): statistics
 
-Note: BatchGenerator is mocked; step() is too complex for unit tests.
+Note: BatchGenerator is mocked; step() coverage is limited to targeted paths.
 """
 
 from collections import deque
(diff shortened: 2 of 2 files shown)


jundot:
Thanks, this matches the scheduler output assembly bug from #1621.

I verified that the current scheduler can drop a prefill rejection output when decode responses are produced in the same step, and this change preserves the terminal error output by appending decode outputs instead of replacing the list. The regression test covers the exact mixed rejection/decode case, and the scheduler test suite passes on the merged tree.

ken-zzzzz:
Thanks for addressing this issue so quickly. I really appreciate your amazing work on the project 👍

PR #1627, opened 2026-06-03
fix(ssd-cache): raise pending-writes floor from 32 to 64

cfbraun:
## Summary

Long-context coding workloads on 32–64 GB Macs were hitting `SSD cache write queue saturated (cap=32); dropping save` roughly every 10–15 minutes. Each dropped block punches a hole in the boundary snapshot, so the next conversation turn loses partial-cache reconstruction at that block index and re-prefills tens of thousands of tokens. Observed in the wild: 600+ second conversation turns immediately after `Failed to save block N to tiered cache` warnings.

## Cause

The prior formula `max(32, total_gb / 2)` clamped to the floor on any system below 128 GB. At 150 K-token prefixes the boundary snapshot is ~73 blocks (block_size=2048), so one turn's snapshot could fill the queue before the writer had drained any of it — especially under memory pressure that throttles the writer thread. Live evidence is in the issue link below.

## Fix

Raise the floor to 64 so a single 150 K-token snapshot fits without saturating. Each queue slot holds a reference to cache-block tensors that the prefix cache already retains, so the actual memory cost of the deeper queue is the `PendingWrite` request struct (small) plus deferred GC of blocks the writer hasn't reached yet — not a fresh allocation of cache data.

Upper bound and scaling slope unchanged: 96+ GB systems already benefited from `total_gb / 2`; only the floor moves.

## Validation

Live before/after on a 48 GB M3 Max under a sustained Pi Coding Agent workload at 150–170 K-token prefixes:

| Window | `SSD cache write queue saturated` | `Failed to save block` | Slow turns (>500 s) explained by SSD drops |
|---|---|---|---|
| Pre-fix (≈1 h) | 5 | 5 | 4 (628 s, 810 s, 658 s, 683 s) |
| Post-fix (≈30 min, same workload) | 0 | 0 | 0 |

All cache hit chains held across consecutive turns post-fix; the boundary snapshot at the end of one busy turn wrote 66 intermediate blocks cleanly — exactly the workload that dropped half its blocks at cap=32.

## Test plan

- [x] `pytest tests/test_paged_ssd_cache.py` — 94 passed
- [x] 30-min live workload with continuous Pi Coding Agent traffic at 150–170 K prefixes; no `saturated`/`Failed to save` events


jundot:
Thanks, the failure mode makes sense to me: a 150k-token store can enqueue ~73 blocks, so cap=32 can drop blocks if the writer does not drain fast enough during the burst.

My concern is the other side of that tradeoff. In the cold SSD path, save_block() materializes tensors_raw before enqueueing, and those raw bytes stay alive in the pending read-back buffer / writer queue until the SSD writer drains them. So on a slow or memory-pressured system, increasing the cap from 32 to 64 can retain substantially more host KV bytes while the original finished-request KV is still alive.

That means this can reduce cache holes, but it may also increase peak memory and turn slow-writer pressure into OOM pressure on smaller Macs. I'm worried about that tradeoff; how do you see it?

cfbraun:
@jundot fair concern — let me quantify the actual retention cost and weigh it against the alternative.

With the default `paged_cache_block_size=256`, each pending slot holds raw bytes for one block. For a 35B-class model with ~64 layers and `head_dim=192` (the workload the user reports came from) that's roughly 48 MB of raw KV bytes per slot — extrapolated from the memory monitor's own 12 MB/64-token estimator at bf16, multi-layer K+V. So the worst-case delta between cap=32 and cap=64 is ~1.5 GB of additional pinned bytes. On a 32 GB Mac that's ~5%; on 64 GB, ~2.5%. Real, but a bounded fraction of host RAM.

The alternative isn't free either. When a block drops and the next request can't reconstruct from that prefix index, the scheduler re-prefills the affected suffix — which allocates the GPU-side KV for those tokens, *then* `_extract_tensor_bytes` re-copies them to host for the (now successful) cache write. That's the same raw-bytes memory event the dropped slot would have held, plus the GPU-side KV peak during re-prefill, plus the throttle overhead and re-prefill wall time. A saturated drop is strictly more expensive in peak host+GPU memory than the slot retention it would have avoided.

Cap is also a *ceiling*, not steady state — the writer drains as fast as the disk allows, and the queue only fills during a burst. Healthy systems with cap=32 and cap=64 behave identically. The cap only matters when the disk is slow or memory pressure is throttling the writer thread; those are the exact cases the user reports came from.

That said, the right axis to scale by is *block bytes*, not RAM. The current formula `min(256, total_gb/2)` ignores `paged_cache_block_size` entirely, so on a system with large blocks the cap can over-retain. A block-size-aware version like:

```python
def _compute_max_pending_writes(block_size_tokens: int = 256, kv_bytes_per_token: int = 200_000) -> int:
    block_bytes = block_size_tokens * kv_bytes_per_token
    # Target: cap pending bytes at ~5% of host RAM, floor 64, ceiling 256
    return max(64, min(256, total_bytes // (20 * block_bytes)))
```

…would keep small Macs with large blocks below 64 if their per-slot cost would push them past the OOM threshold, while still letting the user's 150 K-token burst clear without dropping. Happy to push this as a follow-up commit on the PR if you'd rather have the explicit relationship — or as a separate PR — and the current 64 floor lands as the "small-block, modern-Mac" case it was sized for.

Would the block-size-aware floor address the concern, or would you prefer the floor stays at 32 with the cap exposed as an admin setting for users hitting the saturation case?

jundot:
Thanks, I agree with the direction of the follow-up. Making the cap scale with block bytes addresses the memory-retention side of the tradeoff better than a fixed 64 floor.

I think the PR still needs one more bit of wiring before merge. PagedSSDCacheManager now accepts `expected_block_size_tokens` / `expected_kv_bytes_per_token`, but the real scheduler construction path still does not pass them, so runtime sizing falls back to the 256-token / 200 KB defaults even after Scheduler has adjusted `paged_cache_block_size` for RotatingKVCache / ArraysCache. That means large-block workloads can still be mis-sized.

Could you plumb the final scheduler block size into PagedSSDCacheManager? Passing a model-derived KV bytes-per-token estimate too would be ideal if it is easy to derive there, but the block size wiring is the important part. After that this direction looks good to me.

cfbraun:
@jundot Plumbing shipped:

- `Scheduler.__init__` now passes the *final* `paged_cache_block_size` (post-RotatingKVCache / ArraysCache adjustment) into `PagedSSDCacheManager`. The KV bytes-per-token estimate is also plumbed when the memory monitor is available, derived from `memory_monitor.estimate_block_memory(1)` — that's the cleanest model-aware value at construction time. Falls back to the manager's 200 KB default when the monitor isn't there.
- `CacheFactory.create_paged_ssd_cache` plumbs `config.block_size` too, so direct/test callers also get block-size-aware sizing.

Tests under `TestSchedulerPlumbsBlockSizeToSSDCache` pin the plumbing end-to-end (scheduler → manager → write queue). Initial CI hit a brittle "must differ from defaults" assertion on macOS runners whose RAM tier put both block_size=256 and block_size=1024 into the same clamp band; the fixup (`fb10728`) stashes the constructor inputs on the manager so the test asserts directly on what reached the constructor, bypassing the host-dependent cap math. CI is green across 3.11–3.14 now.

Ready for another look.

jundot:
Thanks for the follow-up. This is close, but I found one remaining runtime wiring issue before I can merge it.

I traced the normal Scheduler constructor path, and `self.memory_monitor` is still `None` when `_init_tiered_cache()` constructs `PagedSSDCacheManager`. That means `expected_kv_bytes_per_token` falls back to the 200 KB default instead of using the model-derived `estimate_block_memory(1)` value. The new test currently allows that fallback when `sched.memory_monitor is None`, so it does not catch this runtime path.

Could you initialize or plumb the model KV estimate before constructing the SSD manager, and make the test assert that value reaches `PagedSSDCacheManager`? After that, this should be ready for another pass.

cfbraun:
Pushed `b676db2` to address the wiring concern. Summary:

**`MemoryMonitor.has_model_info()`** — new accessor returning True iff `set_model_info` was called with real positive ints. Necessary because `estimate_block_memory(1)` silently substitutes a 7B-class fiction (32×8×128 ≈ 128 KB/token) when dims weren't set, so a caller can't tell "real model" from "estimator default" by inspecting the return value.

**`Scheduler._init_tiered_cache()`** — now gates the per-token KV computation on `has_model_info()` instead of just `memory_monitor is not None`. On the canonical path (auto-init in `__init__` + `_set_model_info_for_monitor`), this is True and the model-derived value reaches `PagedSSDCacheManager`. When dim extraction fails (skeletal test fixtures or VLM packs the nested-config walk doesn't recognise), the manager's documented 200 KB default applies — a known constant, not a model-class fiction.

**Test (`test_block_size_plumbed_to_ssd_manager`)** — dropped the conditional `if sched.memory_monitor is None: accept 200_000` branch you flagged. Now hard-asserts:
- `memory_monitor` is non-None (auto-init didn't regress)
- `has_model_info()` is True (extraction succeeded for the fixture)
- the manager received the **model-derived** value, computed inline from the fixture's known dims (32 layers × 8 KV heads × 192 head_dim × fp16 × K+V = 196,608) — not via `estimate_block_memory(1)`, so the assertion can't pass tautologically and a silent 200 KB fallback shows as a numeric diff in the failure message
- and as a secondary check: the value matches `memory_monitor.estimate_block_memory(1)`, catching drift between the monitor's calc and the scheduler's wiring

End-to-end manual check: on the fixture's dims, **196608 reaches the manager** (not the 200 KB default). 324 tests pass locally.

jundot:
Thanks, this is close and I think the overall direction is right.

I do not want to merge the current queue-sizing formula exactly as-is, though. The block-byte-aware cap is the right axis for memory safety, but with the current 5% target and 16-slot floor it can still be too conservative for the 64GB long-context case reported in #1736. In particular, for large-block paths such as 2048-token blocks, the cap can fall below the current 32-slot behavior.

I am going to merge this through an integration branch and add a small maintainer follow-up on top: relax the pending-write budget, make the floor a soft floor bounded by a hard byte budget, increase the queue wait timeout, remove the early full-queue drop path, and fix the logs to report the per-instance cap.

That should keep the memory-safety idea from this PR while making the saturation behavior less likely to punch holes in long-context cache snapshots.

PR #1634, opened 2026-06-03
fix(mac): prevent crash when deleting chat template kwargs

### LINKED ISSUE #1614: Model settings - kwarg
When I try to add some chat kwargs in the model advanced settings, mainly custom ones, the system crashs ant ask for restart.

Choose any model:
1 - Go to model settings
2 - Advanced settings
3 - Add custom chat kwargs (2 or more)
4 - Delete any kwarg

**Expected behavior**
We should be able to add or remove any kwargs without crash

Image

**Screenshots**

 - macOS Version: 26.5
 - oMLX Version 0.4.0

nethbotheju:
Fixes #1614

## Problem

Adding 2+ custom chat kwargs in model advanced settings and then deleting any one causes the app to crash and request a restart.

## Root cause

`EntryEditor` received a plain `Int` index from the `ForEach` and used it to subscript `vm.chatTemplateEntries` in its `binding` getter/setter. When a kwarg was removed, the array shrunk but SwiftUI still held views with stale indices — triggering an out-of-bounds access.

## Fix

- `EntryEditor` now takes a stable `UUID` (`entryID`) instead of an `Int` index.
- `binding` getter/setter looks up entries via `firstIndex(where:)` on the UUID, which is safe across deletions and reorders.
- `ForEach` simplified to iterate entries directly — no more `enumerated()`.

## Changes

- `apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift`
- `EntryEditor`: `index: Int` → `entryID: UUID`
- `binding`: UUID-based lookup with safe fallback
- `ForEach`: `ForEach(Array(...enumerated()))` → `ForEach(vm.chatTemplateEntries)`

### DIFF
--- a/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
+++ b/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
@@ -785,14 +785,13 @@ private struct ChatTemplateKwargsEditor: View {
                     addMenu
                 }
             }
-            ForEach(Array(vm.chatTemplateEntries.enumerated()), id: \.element.id) { idx, entry in
-                let isLast = idx == vm.chatTemplateEntries.count - 1
+            ForEach(vm.chatTemplateEntries) { entry in
+                let isLast = entry.id == vm.chatTemplateEntries.last?.id
                 FreeRow(isLast: isLast) {
                     EntryEditor(
                         vm: vm,
                         client: client,
-                        index: idx,
-                        entry: entry
+                        entryID: entry.id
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for the fix. The root cause matches the crash path in #1614: SwiftUI can keep evaluating a row after the backing array has shrunk, and the old index-based binding can then subscript out of bounds.

Using the stable entry UUID for lookup keeps the stale-view path safe without changing the saved chat-template kwargs shape. This looks good to me, and I'm going to merge it.

PR #1638, opened 2026-06-03
fix(launch): restore interactive model picker when tier models are configured

fparrav:
## Problem

Two related bugs in `omlx launch claude` when `opus_model`, `sonnet_model`, or `haiku_model` are configured in `settings.json`:

1. **Picker never shown**: `claude_has_tier_models` caused the interactive model picker to be skipped entirely — `omlx launch claude` auto-selected a tier model without prompting the user.

2. **Selection ignored**: Even after restoring the picker (workaround), the model chosen interactively was silently overridden by settings-based tier models when building `IntegrationContext`, so Claude Code launched with a different model than the one selected.

## Root cause

In `launch_command` (`cli.py`):
- The block `if not model and claude_has_tier_models: model = sonnet_model or ...` prevented the picker block (`if not model: ... integration.select_model(...)`) from ever being reached.
- After the picker, `IntegrationContext` was built with `opus_model/sonnet_model/haiku_model` from settings, which `claude.py` then used for `ANTHROPIC_DEFAULT_*_MODEL` env vars, overriding `ctx.model`.

## Fix

1. Remove the auto-selection from tier models: if `--model` is not passed, always go through the picker.
2. After interactive selection (no `--model`, no explicit `--opus/--sonnet/--haiku` flags), clear settings-based tier models so the chosen model is used for all Claude Code tiers.

**Behavior after fix:**

| How you launch | Result |
|---|---|
| `omlx launch claude` | Picker shown → selected model used for all tiers |
| `omlx launch claude --model X` | Uses X directly, no picker |
| `omlx launch claude --sonnet X --haiku Y` | Explicit tier flags respected as before |

Settings-based tier models still work correctly when the user explicitly passes `--opus/--sonnet/--haiku` flags.


jundot:
Thanks, this looks good to me. I found one small lint-only cleanup after review, so I'm going to merge this and fold that into a maintainer follow-up on main.

PR #1644, opened 2026-06-03
feat(mac): expose all DFlash settings in model config screen

fparrav:
## Summary

- Adds 9 missing DFlash fields to the macOS model settings screen that were supported by the server but not exposed in the Swift UI
- Fixes a bug where `dflash_draft_quant_bits` (unrecognised by the server) was being sent instead of the correct structured fields

### New controls

| Field | Type | Server key |
|---|---|---|
| Draft Quantization | toggle | `dflash_draft_quant_enabled` |
| Weight Bits | picker (2/4/8) | `dflash_draft_quant_weight_bits` |
| Activation Bits | picker (16/32) | `dflash_draft_quant_activation_bits` |
| Group Size | picker (32/64/128) | `dflash_draft_quant_group_size` |
| Verify Mode | picker (dflash/adaptive/off) | `dflash_verify_mode` |
| Draft Window Size | text field | `dflash_draft_window_size` |
| Draft Sink Size | text field | `dflash_draft_sink_size` |
| Cache Entries | text field | `dflash_in_memory_cache_max_entries` |
| SSD Cache Size | text field (GiB) | `dflash_ssd_cache_max_bytes` |

### Bug fix

The old `dflash_draft_quant_bits` picker sent key `"dflash_draft_quant_bits"` which does not exist in `model_profiles.py`. The server was silently ignoring it. Replaced with a proper enable toggle + separate weight/activation/group-size pickers that map to the server's actual fields.

## Test plan

- [ ] Enable DFlash on a compatible model and verify all new rows appear
- [ ] Toggle Draft Quantization ON/OFF — sub-pickers appear/hide and sub-fields are cleared on disable
- [ ] Set Verify Mode, Draft Window Size, Draft Sink Size — confirm values round-trip through profile save/reload
- [ ] Set Cache Entries and SSD Cache Size — confirm bytes conversion is correct
- [ ] Disable in-memory cache — SSD cache sub-toggle disables; Cache Entries row hides
- [ ] Enable SSD cache — SSD Cache Size row appears


jundot:
Thanks, this is a useful parity improvement for the macOS settings screen. I merged the DFlash settings changes manually on main and added the small `ddtree` verify-mode follow-up there, because this branch was stacked on #1638. Closing this PR now that the changes are in main.

PR #1647, opened 2026-06-04
chore: update mlx-vlm in pyproject.toml to support gemma4_unified

dapx:


dapx:
Okay, it's already updated to the same version some minutes ago. I'm closing it.

Issue #600, opened 2026-04-05
指定进程名称为 omlx 的请求


iamckun:
可以吧进程名称改为本程序的名称吗 omlx，现在想看占用老要进去看看才行。本次python程序有点多。
现在是brew 安装的 版本0.3.4

jundot:
I'm not sure if it's possible to change the process name itself since the app runs on python, but i'll look into it.

As a solo developer, i'm currently prioritizing stability issues, so this might take some time. Thanks for your patience!

brinsche:
Some manual searching seems to indicate there's no native support in python but it should work with this https://github.com/dvarrazzo/py-setproctitle (but I haven't tried it)

iamckun:
> Some manual searching seems to indicate there's no native support in python but it should work with this https://github.com/dvarrazzo/py-setproctitle (but I haven't tried it)

Yes, it is feasible. I have tried this setting already. And the process name was successfully created.
@jundot

PR #1658, opened 2026-06-04
Feat/setproctitle

iamckun:
#600


jundot:
Thanks for the PR. I checked the macOS app lifecycle path, and this looks safe because the Swift parent tracks the child by PID and health checks rather than by process title.

I am going to merge this, then fold in a small follow-up on `main`: use `omlx-server` as the displayed title and keep the title setup only in the standard `omlx serve` entry point.

Issue #1410, opened 2026-05-25
v0.3.10 regression: Gemma 4 E4B emits raw template scaffolding when client sends OpenWebUI web search tool


Stooovie:
Worked on: v0.3.9
Broken on: v0.3.10
Suspect: #1393

Repro:

Model: mlx-community/gemma-4-e4b-it-mxfp4 (commit b24a2f4)
Client: OpenWebUI v{your version}, web search enabled
Query: "list all solo albums by jeff lynne"
Expected: prose answer
Actual: response contains literal <eos>, raw OWUI <details type="tool_calls">…</details> block with HTML-escaped JSON, no answer

<img width="1508" height="1210" alt="Image" src="https://github.com/user-attachments/assets/34381be5-7ec5-4fec-bdab-d43f696c9d3f" />

Not reproducible with the same OWUI setup pointed at gemma-4-e2b or gemma-4-26b-a4b — only E4B. Suggests the Gemma 4 tool-call extractor path is the regression site, not OWUI. 

It worked perfectly on 0.3.9 and continues to work perfectly with Gemma 26a and E2B. E4B does this on every web search.

Thanks for looking into it!

Stooovie:
- other Gemma 4 variants do not have the issue
- default model settings in both oMLX and OWUI or Gemma 4 template do not have any effect
- I have deleted and redownloaded the model
- no OWUI update
- it looked like it was isolated to oMLX 0.3.10 but no, even 0.3.9 does that
- web loader and search engine settings in OWUI have no effect
- I always hard reload in browser (Arc, Zen, Safari) and have restarted the entire AI stack (OWUI in Docker and oMLX) many times

kreeger:
I've introduced a fix that _might_ help with this, but I'm not totally sure as I don't use E4B with Open WebUI — but it's worth a re-check on the latest `main` or with 0.4.2.dev3 when it gets released.

Stooovie:
This is still happening on oMLX 0.4.2, and further testing shows the issue is broader than the original web-search-only report.

**Summary**

With Gemma 4 E4B, the model’s raw output — both the <eos> token and the full <details type="tool_calls"> scaffolding block — leaks into the response content as text instead of being parsed/consumed by the tool-call extractor. The client (Open WebUI) then renders this raw markup verbatim, because it arrives as plain content rather than a structured tool call.

Environment
	•	oMLX: 0.4.2
[...]

PR #1665, opened 2026-06-04
fix: Gemma 4 MoE multi-turn tool call corruption

### LINKED ISSUE #617: Bug Report: oMLX v0.3.4, gemma-4-26b-a4b-it-6bit Tool Calling Fails in Claude Code

### LINKED ISSUE #1465: Bug: Gemma 4 returns empty content when tool results are sent back with `role: "tool"`
Gemma 4 (gemma-4-e2b-it-4bit) returns empty content when tool results sent back with `role: "tool"` messages.

Reproduce: Send tools + user msg → model returns tool_calls ✅. Then send assistant msg + role:tool msg → model returns content:null, completion_tokens:0 ❌.

Root cause: Gemma 4 uses native tokens `<|tool_response>`/`<tool_response|>` for tool results. oMLX sends OpenAI role:tool but doesn't convert to this format.

Related: #617, #666, #741, #1410
Env: oMLX 0.3.12, macOS 26.5, Apple M4 16GB, gemma-4-e2b-it-4bit

kreeger:
👋 First time contribution for me! Please let me know if you'd like to see stuff changed on this changeset, I'm happy to work with you all. Been loving oMLX for weeks now, figured it was time to give back.

## Problem domain

Multi-turn tool-calling conversations with Gemma 4 26B A4B were breaking on the second and later turns. Stray protocol tokens (`<|tool_call>`, `<tool_call|>`) were leaking from the streaming layer into the assistant message content. When that message was getting fed back through `apply_chat_template`, the tokens got treated as real special tokens, producing unbalanced delimiters that corrupt the model's context for all subsequent turns.

This PR fixes that by identifying and stripping stray closing markers in `tool_calling.py`, and sanitizes the `content` in Gemma 4 runs before re-rendering conversation history.

## Root cause(s)

1. `ToolCallStreamFilter` didn't strip stray close markers.
    + A `<tool_call|>` emitted outside a matched open/close pair passed through unfiltered
    + If the tokenizer split it across two feed calls (`<tool_call|` then `>`), the buffer would also flush each half immediately rather than holding them for reassembly
1. `extract_gemma4_messages` didn't sanitize content before re-rendering history.
    + Any marker that survived the streaming pass was embedded verbatim into the next turn's prompt

## How to reproduce

1. Load any Gemma 4 26B A4B model with at least one tool defined.
1. Send a request that causes a tool call.
1. Append the tool result and send a follow-up turn.
1. Observe incoherent output or tool use failure on turn 2+. Verbose prompt logging will show `<tool_call|>` embedded literally in the rendered prompt.

## Edge cases considered

- Split-feed: `<tool_call|>` split as `<tool_call| + >` across two `feed()` calls
    + Fixed by adding stray-close markers to the partial-prefix hold logic
- Well-formed `</tool_call>` in prose: The strip is scoped to the bare stray tokens only (no leading slash), so legitimate closing tags in prose are untouched
- Prose around a marker: The regex removes only the token
    + "Here is the result.<tool_call|>" becomes "Here is the result."

## Issues to close

This ought to finally close #617, which I was getting bit by. It might also close #1465 and/or #1410, but I'm not sure about that; those users should try re-testing if/when this makes it into `main`.

### DIFF
--- a/omlx/adapter/gemma4.py
+++ b/omlx/adapter/gemma4.py
@@ -25,6 +25,17 @@ _LEADING_THOUGHT_RE = re.compile(
     re.DOTALL,
 )
 
+# Matches the STRAY bare-token spellings (<|tool_call> and <tool_call|>),
+# not the template's well-formed closing form (</tool_call|> with slash).
+_PROTOCOL_MARKER_RE = re.compile(r"<\|tool_call>|<tool_call\|>")
+
+
+def _strip_protocol_markers(text: Any) -> Any:
+    """Remove stray <|tool_call> / <tool_call|> tokens from assistant content."""
+    if not isinstance(text, str) or not text:
+        return text
+    return _PROTOCOL_MARKER_RE.sub("", text)
+
 
 def _try_parse_json(s: str) -> Any:
     """Parse string as JSON if possible, otherwise return as-is."""
@@ -172,6 +183,7 @@ def extract_gemma4_messages(
             # Per Gemma 4's multi-turn rule, prior thought blocks must not
             # be fed back into the next turn. Strip them before rendering.
             content = _strip_thinking(content)
+            content = _strip_protocol_markers(content)
 
             out_msg: dict = {"role": "assistant", "content": content or ""}
 
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -845,6 +845,10 @@ class ToolCallStreamFilter:
         # Same for suppress-after markers (e.g. "[TOOL" for "[TOOL_CALLS]").
         for sa_marker in self._suppress_after_markers:
             keep = max(keep, self._partial_prefix_len(text, sa_marker))
+        # Hold partial prefix of a stray-close marker so it reassembles before
+        # the strip check — prevents the "hello<tool_call|" + ">" split leak.
+        for close_marker in self._stray_close_markers:
+            keep = max(keep, self._partial_prefix_len(text, close_marker))
 
         bracket_idx = -1
         for bp in self._bracket_prefixes:
@@ -1010,9 +1014,6 @@ class ToolCallStreamFilter:
             self._buffer = ""
             if self._should_drop_tail_at_finish(tail):
                 return ""
-            for close in self._stray_close_markers:
-                if close in tail:
-                    tail = tail.replace(close, "")
             return tail
 
         if keep:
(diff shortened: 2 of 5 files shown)


jundot:
Thanks for tracking this down. The Gemma 4 multi-turn repro lines up with the leak through stored assistant content, and this fixes the path I care about for #617. I found one small scope issue in the streaming filter for non-Gemma XML close markers, so I will merge this and fold that into a follow-up on main.

PR #1668, opened 2026-06-04
Fix engine-pool acquire-vs-use eviction race (#1667) + active-request counter leak (#1595)

### LINKED ISSUE #1667: Engine pool evicts an acquired-but-not-yet-active engine → `RuntimeError: Engine not started` (acquire→use race; sibling of #1595)
**Version:** 0.4.1 (logic appears present in 0.3.x as well)
**Severity:** intermittent 5xx on `/v1/embeddings` and `/v1/rerank` under memory pressure; the embedding engine is the usual victim.

## Summary

Under memory pressure, the memory enforcer can evict an engine that a request has just acquired but not yet registered activity on, so the in-flight request resumes against a torn-down engine and raises:

```
RuntimeError: Engine not started. Call start() first.
```

(raised at `omlx/engine/embedding.py` `embed()` and the equivalent guards in `reranker.py`/`stt.py`/`tts.py`/`sts.py`).

This is the sibling of #1595: that issue is about evicting an engine that **is** busy (the `active_requests` counter leaks); this one is about evicting an engine that **looks** idle because the in-use signal hasn't been registered yet (acquire→use TOCTOU). Both stem from the same gap: eviction is only gated by `has_active_requests()`, which is blind during the window between acquiring the engine and incrementing the activity counter.

## Root cause (TOCTOU window)

`EnginePool.get_engine()` returns the engine handle while holding `self._lock`, then **releases the lock on return** (`engine_pool.py`, the already-loaded path ~`entry.last_access = time.time(); return entry.engine`, and the post-load `return self._entries[model_id].engine`).

The request handler then runs **outside** the lock: `server.py create_embeddings` does `engine = await get_embedding_engine(request.model)` and only later `await engine.embed(...)`. Inside `embed()`, the started guard `if self._model is None: raise RuntimeError("Engine not started...")` runs **before** `_begin_activity()` — and `_begin_activity()` is the only thing that increments `_active_count` (`engine/base.py`). So `has_active_requests()` returns `False` for the entire span from acquire until `_begin_activity`.

Meanwhile `ProcessMemoryEnforcer`'s ~1s poll detects soft/hard pressure, takes the same pool lock, and `_find_lru_victim()` (and `_is_idle_for_prefill_eviction()`, and the `check_ttl_expirations()` path) sees `has_active_requests() == False`, so the engine is a valid victim. The embedding engine is the smallest and usually the oldest `last_access`, so it is picked first. `_unload_engine()` → `engine.stop()` → `_model = None`, then `entry.engine = None`. The suspended request resumes and hits the guard.

This matches the logs: the crash correlates with `engine_pool` `"Unloading model: <id> (immediate abort)"` and the scheduler's `adaptive_prefill_throttle`/prefill-headroom lines, and it is the recurring cause of "embedder degraded" episodes for downstream consumers.

## Three eviction paths share the blind guard
- `_find_lru_victim()` — LRU eviction
- `_is_idle_for_prefill_eviction()` — prefill-pressure eviction (the one the logs most often show)
- `check_ttl_expirations()` — TTL eviction

## Reproduction
1. Run with a `--memory-guard-gb` ceiling and at least one large LM lane plus the small embedding model loaded, sitting near the soft watermark.
2. Drive `/v1/embeddings` for the embedding model while a generation lane keeps memory pressure high (so the enforcer poll selects the idle-looking embedding engine).
3. Intermittently a `/v1/embeddings` (or `/v1/rerank`) request 500s with `RuntimeError: Engine not started. Call start() first.` in the FastAPI traceback.

## Suggested fix (atomic in-use lease)
Take the in-use signal **at acquire time, under the pool lock**, so eviction can never tear down an engine a request holds or is acquiring:

1. Add an `in_use: int = 0` refcount to `EngineEntry`.
2. `get_engine(..., _lease: bool = False)` — increment `entry.in_use` under `self._lock` at both return points.
3. Add `release_engine(model_id)` (decrement under lock, floor at 0) and an `acquire(model_id)` `@asynccontextmanager` that leases on enter and releases in `finally` (so a request error can't leak the lease).
4. Skip `in_use > 0` in **all three** eviction paths (in addition to `has_active_requests()`).
5. Have the `/v1/embeddings` and `/v1/rerank` handlers wrap the actual `embed()`/`rerank()` call in `async with pool.acquire(model_id)` (capturing the exact pool-resolved model_id, incl. the default-model fallback, so lease/release always match).

This closes both sub-windows (acquire→activity, and the guard→increment gap inside `embed()`), preserves eviction of genuinely idle engines, and — together with a counter reset on teardown — also defuses #1595 by preventing the immediate-abort of an in-use engine in the first place.

Happy to contribute a PR if useful.

Cmerrill1713:
## Summary

Fixes two related memory-enforcer bugs that intermittently 500 `/v1/embeddings` and `/v1/rerank` with:

```
RuntimeError: Engine not started. Call start() first.
```

Both stem from the same gap: eviction is gated only by `has_active_requests()`, which is blind during the window between acquiring an engine and registering activity on it.

Fixes #1667. Closely related to #1595 (this PR also defuses it).

### Bug A — acquire-vs-use eviction race (#1667)

`EnginePool.get_engine()` returns the engine handle, then releases the pool lock. The request handler then runs **outside** the lock (`engine = await get_embedding_engine(...)` ... later `await engine.embed(...)`). For the non-streaming engines (embedding/reranker), the in-use signal (`_active_count` via `_begin_activity`) is registered only **after** the `if self._model is None: raise RuntimeError("Engine not started...")` guard inside `embed()`/`rerank()`. So `has_active_requests()` returns `False` for the entire span from acquire until `_begin_activity`.

Meanwhile the `ProcessMemoryEnforcer`'s ~1s poll takes the same pool lock and `_find_lru_victim()` / `_is_idle_for_prefill_eviction()` / `check_ttl_expirations()` see the engine as idle, so they evict it. The suspended request resumes against a torn-down engine and raises. The embedding engine (smallest, usually oldest `last_access`) is the usual victim — the recurring cause of "embedder degraded" episodes for downstream consumers.

### Bug B — active-request counter leak (#1595)

The immediate-abort eviction stops a non-streaming engine **without** the normal per-request completion callbacks (`_end_activity`), so `_active_count` can be left non-zero — a phantom "busy" count that corrupts the status API and (via `has_active_requests()`) can make a stale engine look permanently non-evictable.

## Fix — atomic in-use lease + teardown reset

Take the in-use signal **at acquire time, under the pool lock**, so eviction can never tear down an engine a request holds or is acquiring:

- `EngineEntry.in_use: int = 0` refcount.
- `get_engine(..., _lease: bool = False)` increments `entry.in_use` under `self._lock` at **both** return points (already-loaded + post-load).
- `release_engine(model_id)` decrements under the lock (floored at 0); `acquire(model_id)` `@asynccontextmanager` leases on enter and **always** releases in `finally` (so a request error can't leak the lease).
- All **three** eviction paths skip `in_use > 0`: `_find_lru_victim`, `_is_idle_for_prefill_eviction`, `check_ttl_expirations` (in addition to `has_active_requests()`).
- `server.py` threads `_lease` / `_leased_out` so the **exact** pool-resolved model_id (incl. the default-model fallback) is captured, and releases the lease on validation failure. `/v1/embeddings` and `/v1/rerank` wrap the real `embed()`/`rerank()` call in `async with pool.acquire(model_id)` (the embeddings lease is taken inside the deferred `StreamingResponse` coroutine, which is where the engine is actually used). LLM/STT/TTS/STS handlers are unchanged. The non-lease path keeps the original `pool.get_engine(model_id)` call contract.
- #1595: `BaseNonStreamingEngine._reset_activity_tracking()` zeros `_active_count` and clears `_activities` under `_active_lock`, called (getattr-guarded) from `EnginePool._unload_engine()` after `stop()`.

This closes both sub-windows (acquire→activity, and the guard→increment gap inside `embed()`) and **preserves eviction of genuinely idle engines** — only engines that are leased/in-use are protected.

## Testing

Tests run with the repo's `pytest` suites (Python 3.11, mlx available):

- `tests/test_engine_pool.py` + `tests/test_embedding.py`: **154 passed** (incl. 10 new regression tests).
- New regression tests (`TestEnginePoolInUseLease`, `TestResetActivityTracking`): the three eviction paths skip `in_use > 0`; `acquire()` leases then releases on success **and** on exception; `release_engine` floors at 0; `_reset_activity_tracking` clears a leaked counter; `_unload_engine` invokes the reset on teardown. Verified these **fail** against the pre-fix code and **pass** with the fix (non-vacuous).
- Server/rerank/embeddings suites (`tests/test_server.py`, `tests/test_rerank_models.py`, integration endpoint tests): **green**. The integration `MockEnginePool` was updated to mirror the new lease API (`_lease` kwarg + no-op `release_engine`).
- Full default suite (`-m "not slow and not integration"` plus the explicit integration endpoint tests): **green** except two pre-existing `test_install_detection.py` failures that are environment artifacts of running under a Homebrew-installed interpreter (`is_homebrew()` detects the Cellar path); they fail identically on unmodified `main` and are unrelated to this change.

### DIFF
--- a/omlx/engine/base.py
+++ b/omlx/engine/base.py
@@ -271,6 +271,19 @@ class BaseNonStreamingEngine(ABC):
         with self._active_lock:
             return self._active_count > 0
 
+    def _reset_activity_tracking(self) -> None:
+        """Clear the in-flight activity counter + records on engine teardown.
+
+        #1595: the memory-enforcer's immediate-abort eviction stops the engine WITHOUT
+        running the normal per-request completion callbacks (_end_activity), so the
+        ``_active_count`` atomic counter can be left non-zero. That phantom 'busy' count
+        both corrupts the status API and (via has_active_requests()) can make a stale
+        engine look permanently non-evictable. Called from EnginePool._unload_engine().
+        """
+        with self._active_lock:
+            self._active_count = 0
+            self._activities.clear()
+
     _ACTIVITY_RESERVED_KEYS = {
         "request_id",
         "kind",
--- a/omlx/engine_pool.py
+++ b/omlx/engine_pool.py
@@ -17,6 +17,7 @@ import asyncio
 import gc
 import logging
 import time
+from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Literal
 
@@ -69,6 +70,7 @@ class EngineEntry:
     loading_started_at: float | None = None  # Timestamp when current load started
     is_pinned: bool = False  # Never evict if True
     abort_loading: bool = False  # Set by memory enforcer to abort in-progress load
+    in_use: int = 0  # in-flight acquire/use lease count; never evict while > 0
 
 
 class EnginePool:
@@ -329,6 +331,7 @@ class EnginePool:
         self,
         model_id: str,
         force_lm: bool = False,
+        _lease: bool = False,
     ) -> (
         BaseEngine
         | EmbeddingEngine
@@ -377,6 +380,8 @@ class EnginePool:
                     await self._unload_engine(model_id)
                 else:
                     entry.last_access = time.time()
+                    if _lease:
+                        entry.in_use += 1
                     return entry.engine
 
             # Pre-load admission against the memory ceiling from the
@@ -428,7 +433,31 @@ class EnginePool:
             # Now load the model
             await self._load_engine(model_id, force_lm=force_lm)
 
-            return self._entries[model_id].engine
+            loaded = self._entries[model_id]
+            if _lease:
+                loaded.in_use += 1
+            return loaded.engine
+
+    async def release_engine(self, model_id: str) -> None:
+        """Release one in-use lease previously taken via get_engine(_lease=True)."""
+        async with self._lock:
+            e = self._entries.get(model_id)
+            if e is not None and e.in_use > 0:
+                e.in_use -= 1
+
+    @asynccontextmanager
+    async def acquire(self, model_id: str, force_lm: bool = False):
+        """Acquire an engine with an atomic in-use lease.
+
+        The lease is taken under the pool lock at acquire time and always
+        released in finally, so the engine cannot be evicted mid-request even
+        on exception.
+        """
+        engine = await self.get_engine(model_id, force_lm=force_lm, _lease=True)
+        try:
+            yield engine
+        finally:
+            await self.release_engine(model_id)
 
     def _find_lru_victim(self) -> str | None:
         """
@@ -443,6 +472,8 @@ class EnginePool:
         candidates = []
         for mid, e in self._entries.items():
             if e.engine is None or e.is_pinned:
+                continue
+            if e.in_use > 0:
                 continue
             try:
                 if e.engine.has_active_requests():
@@ -468,7 +499,7 @@ class EnginePool:
 
     def _is_idle_for_prefill_eviction(self, entry: EngineEntry) -> bool:
         engine = entry.engine
-        if engine is None or entry.is_pinned or entry.is_loading:
+        if engine is None or entry.is_pinned or entry.is_loading or entry.in_use > 0:
             return False
         try:
             if engine.has_active_requests():
@@ -570,6 +601,17 @@ class EnginePool:
             await entry.engine.stop()
         except Exception as e:
             logger.warning(f"Error stopping engine for {model_id}: {e}")
+
+        # #1595: the immediate-abort stop() above tears the engine down without the normal
+        # per-request completion callbacks, so a non-streaming engine's active_requests
+        # counter can leak a phantom count (a stale engine then looks permanently busy).
+        # Reset it on teardown so has_active_requests() and the status API stay consistent.
+        reset = getattr(entry.engine, "_reset_activity_tracking", None)
+        if callable(reset):
+            try:
+                reset()
+            except Exception as e:
+                logger.warning(f"Error resetting activity counter for {model_id}: {e}")
 
         # Yield to the event loop before dropping the engine reference.
         #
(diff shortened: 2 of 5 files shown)


Cmerrill1713:
Quick context for review, in case it helps prioritize:

This was the root cause behind recurring **"embedder degraded under load"** behavior — the small embedding engine consistently loses the eviction lottery during the acquire→use window, so it surfaces as the embedder even though it's a general engine-pool race. It pairs with #1595 (the counter-leak sibling): both come from eviction being gated only by `has_active_requests()`, which is blind between acquiring an engine and registering its activity. This PR closes both with one mechanism — an atomic `in_use` lease taken under the pool lock at acquire and released in `finally`, plus a counter reset on teardown.

A few things that should make it low-risk to review:
- **Ported clean** to current HEAD (0.4.2.dev2) — the three source files were byte-identical to 0.4.1.
- **Full suite green**: 4,967 passed / 37 skipped (the only 2 failures are pre-existing `test_install_detection` env artifacts that fail on unmodified `main` too).
- **10 new regression tests are non-vacuous** — they fail on pristine source and pass with the patch (eviction skips a leased entry across all three paths; `acquire()` releases on success *and* exception; the counter resets on `stop()`).
- **Backward-compatible**: `_lease` is only threaded when actually leasing, so existing `get_engine` callers and test mocks are unaffected; idle-engine eviction and the memory-guard watermarks are unchanged.

Happy to split this into two PRs (#1667 race vs #1595 counter), rename anything to match your conventions, or add cases you'd like covered.

jundot:
Thanks for the extra context. The acquire-to-use race you described matches what I see in the code: non-streaming engines only register activity after the started guard, so has_active_requests() cannot protect an engine handle that was just returned by get_engine().

I checked the lease path against the memory-enforcer eviction paths and the embeddings/rerank server wrappers. Taking the lease under the pool lock and skipping leased entries in LRU, prefill, and TTL eviction closes that window while preserving idle eviction. I also verified the teardown reset for non-streaming activity counters.

The added regression coverage targets the right invariants, and my focused local runs passed. This looks good to me.

PR #1673, opened 2026-06-05
feat(dflash): expose prefix-cache controls and streaming metrics

turururu:
## Summary

- Expose DFlash maximum snapshot tokens and L2 frontier stride through model
  settings, profiles, the admin API, and the dashboard.
- Pass optional runtime controls only when the installed `dflash-mlx`
  signature supports them, preserving compatibility with older releases.
- Forward prefix-cache hit classification and expose restored/computed tokens,
  DFlash cycles, draft acceptance, prefill throughput, and peak memory in
  streaming usage metrics.
- Keep metric forwarding compatible with older prefix-flow objects that do not
  expose `hit_kind`.

Depends on merged bstnxbt/dflash-mlx#37 for the new snapshot behavior and event
fields.

This is complementary to #1462: that PR handles request/policy identity and
cached-token accounting; this PR adds configurable snapshot controls and
detailed streaming performance metrics.

## Validation

Post-rebase affected suite:

```text
319 passed, 12 deselected in 2.05s
```

Full suite before the conflict-free rebase:

```text
5073 passed, 40 skipped, 66 deselected in 241.29s
```

Sibling dflash-mlx full suite:

```text
1102 passed, 8 skipped, 2 warnings in 44.92s
```

## Apple Silicon Benchmarks

Platform: MacBook Pro M1 Max, 64 GB unified memory, macOS 26.5.

All bundles used oQ6 quantized target matrices, FP16 target floating tensors,
and BF16 DFlash drafters. Thinking was disabled. Generation rates below use a
1K-token prompt and 256 generated tokens. Cache repeats use a 65,536-token
prompt.

| Family | Generation | 1K TTFT | 64K L1 exact | 64K L2 exact |
|---|---:|---:|---:|---:|
| Qwen 3.6 27B | 53.1 tok/s | 7.59 s | 0.92 s | 5.91 s |
| Qwen 3.6 35B-A3B | 190.1 tok/s | 1.42 s | 0.58 s | 6.10 s |
| Gemma 4 31B | 42.9 tok/s | 9.26 s | 0.95 s | 8.44 s |
| Gemma 4 26B-A4B | 141.0 tok/s | 1.48 s | 0.64 s | 5.02 s |

All 8K, 32K, and 64K L1/L2 probes restored the expected prefix and produced
the same deterministic response SHA-256 as their cold controls.

For the original Gemma 31B validation, DFlash decode reached 29.2 tok/s at
93.8% acceptance versus 9.4 tok/s target-only. At 64K, the measured peak
DFlash memory was 46.54 GB.

## Disclosure

This PR was vibe-coded with AI assistance and supervised, reviewed, and
validated by me.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -140,11 +140,13 @@ class ModelSettingsRequest(BaseModel):
     dflash_in_memory_cache: bool | None = None
     dflash_in_memory_cache_max_entries: int | None = None
     dflash_in_memory_cache_max_bytes: int | None = None
+    dflash_max_snapshot_tokens: int | None = None
     dflash_ssd_cache: bool | None = None
     dflash_ssd_cache_max_bytes: int | None = None
     dflash_draft_window_size: int | None = None
     dflash_draft_sink_size: int | None = None
     dflash_verify_mode: str | None = None
+    dflash_l2_frontier_stride: int | None = None
     # Native MTP (mlx-lm PR 990 / PR 15 monkey-patch)
     mtp_enabled: bool | None = None
     # VLM MTP speculative decoding via external assistant drafter (mlx-vlm 191d7c8+)
@@ -2078,6 +2080,11 @@ async def update_model_settings(
         current_settings.dflash_in_memory_cache_max_bytes = int(
             request.dflash_in_memory_cache_max_bytes
         )
+    if "dflash_max_snapshot_tokens" in sent:
+        value = request.dflash_max_snapshot_tokens
+        current_settings.dflash_max_snapshot_tokens = (
+            int(value) if value is not None and value >= 0 else None
+        )
     if "dflash_ssd_cache" in sent:
         ssd_requested = bool(request.dflash_ssd_cache)
         if ssd_requested:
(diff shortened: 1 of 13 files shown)


jundot:
Thanks for the update. I rechecked this against the current dflash-mlx main commit I now pin, `b7f192b62bc5a59cad41fda888c1118c60fc58b1`.

`prefix_hit_kind`, `SummaryEvent.hit_kind`, `max_snapshot_tokens`, and `prefix_cache_l2_frontier_stride` are present there, but I do not see `dflash_min_output_tokens` in `runtime_config_from_defaults()` or elsewhere in the runtime config. In this PR, `dflash_min_output_tokens` is added to oMLX settings, but with the current dflash-mlx pin it will be ignored by the signature guard and only log a warning when configured.

Could you recheck whether this knob still exists upstream, or remove/defer it from this PR until dflash-mlx exposes it?

One smaller thing: the new settings are added to `ModelSettings` and profile allowlists, but not to the admin settings API/dashboard save path. If the intended scope is raw settings/profile JSON only, that is fine; otherwise the admin API/UI path should be wired too.

turururu:
Thanks for checking against the merged dflash-mlx revision. You were right:
`dflash_min_output_tokens` came from an earlier prototype and did not land in
dflash-mlx, so I removed it from the oMLX integration rather than retaining an
ignored setting.

I also wired the two supported cache controls,
`dflash_max_snapshot_tokens` and `dflash_l2_frontier_stride`, through the admin
settings API and dashboard load/save path. Changes to these controls now
trigger an engine reload, as do the existing draft window, sink, and verifier
settings, which are also applied at engine construction time.

After rebasing onto current `main`, the affected suite reports:

```text
319 passed, 12 deselected in 2.05s
```

The full suite before the conflict-free rebase reported:

```text
5073 passed, 40 skipped, 66 deselected in 241.29s
```

I also updated the PR description with an anonymous M1 Max benchmark summary
covering Qwen 27B/35B and Gemma 31B/26B-A4B, including 64K L1 and L2 restore
times.

PR #1682, opened 2026-06-05
Fix FP16 overflow for vision tensors (Fixes #1678)

### LINKED ISSUE #1678: [bug] Infinite <pad> generation with image inputs (FP16 Overflow in Vision Tensors when using float16 via oQ)
1. Environment
    Hardware: Apple Silicon M1 Max (64GB)
    Models Tested: Gemma 4 31B, Gemma 4 26B-A4B (Native BF16 models)
    Framework/Script: oMLX dynamic quantization (oq.py)

2. Bug Description
When quantizing a BF16 VLM model with float16, text-only inference works perfectly. However, providing an image input causes the model to break and output an infinite loop of <pad> tokens.

3. Root Cause Analysis
This is caused by an FP16 overflow (NaN propagation) inside the vision encoder / multimodal projector.
In the oq.py script, vision tensors are correctly excluded from 4-bit quantization via _is_vision_tensor(). However, the fallback casting logic casts these non-quantized tensors to target_dtype (which is float16 in this case).
Since Gemma 4 is a native BF16 model, the internal activation values of its vision tower easily exceed the FP16 maximum range (~65,504) during image processing. This overflow generates NaN values that propagate into the LLM, destroying the probability distribution and forcing <pad> token loops.

4. Proposed Solution
To fix this for FP16 fallback environments (like M1 users), vision and audio tensors should be upcast or kept in float32 instead of float16, as float32 handles the large dynamic range without OOM issues (vision parameters are relatively small).

Here is a suggested fix around line 2699~2702 in the tensor saving loop in oQ.py:
```
      else:
            # Prevent FP16 overflow for vision/audio tensors by forcing FP32
            if _is_vision_tensor(tensor_name) or _is_audio_tensor(tensor_name):
                if mx.issubdtype(w_mx.dtype, mx.floating):
                    w_mx = w_mx.astype(mx.float32)
            # Apply target_dtype to other protected LLM text tensors
            else:
                if mx.issubdtype(w_mx.dtype, mx.floating) and w_mx.dtype != target_dtype:
                    w_mx = w_mx.astype(target_dtype)
            
            out_shard_data[tensor_name] = w_mx
```

dodams258:
Fixes #1678.

Summary
    oQ: vision and audio tensors are kept in float32 instead of float16

Test
    pytest tests/test_oq.py -- 206 passed

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -2696,8 +2696,12 @@ def quantize_oq_streaming(
                     w_mx = w_mx.astype(target_dtype)
                 out_shard_data[tensor_name] = w_mx
         else:
-            if mx.issubdtype(w_mx.dtype, mx.floating) and w_mx.dtype != target_dtype:
-                w_mx = w_mx.astype(target_dtype)
+            if _is_vision_tensor(tensor_name) or _is_audio_tensor(tensor_name):
+                if mx.issubdtype(w_mx.dtype, mx.floating) and w_mx.dtype != target_dtype:
+                    w_mx = w_mx.astype(mx.float32)
+            else:
+                if mx.issubdtype(w_mx.dtype, mx.floating) and w_mx.dtype != target_dtype:
+                    w_mx = w_mx.astype(target_dtype)
             out_shard_data[tensor_name] = w_mx
 
         del w_mx


jundot:
Thanks for the patch. The root cause looks right: protected vision/audio tensors should not be downcast to FP16 in the float16 oQ path.

One gap is that this patch only changes the `_should_quantize_tensor() == false` fallback. The main 2D vision/audio weight tensors still go through `_should_quantize_tensor() == true`, then `_get_predicate_bits()` returns `None`, so they hit the existing `bits is None` fallback and are still cast to `target_dtype`.

I will handle the remaining part in a follow-up commit: apply the same protected pass-through dtype policy to both fallback paths and add a regression test for a 2D vision/audio tensor with `dtype="float16"`.

PR #1690, opened 2026-06-05
feat: Add support for Gemma 4 QAT models in quant tool

kreeger:
Earlier today, the DeepMind team [released quantization-aware training versions](https://blog.google/innovation-and-ai/technology/developers-tools/quantization-aware-training-gemma-4/) of their Gemma 4 family of models, which are versions that behave much better when run through quantization, but their configs look a little different than others:

* They're missing `quant_method` like MiniMax and DeepSeek and others
* They still have `quantization_config` keys though

This changeset checks for the structure of these keys in the quantization config and if it's present, it allows Gemma 4 QAT models through to be quantized. Sure, the training doesn't directly affect oQ4 quantization, but it's my hope? theory? that oQ4 on these models is a bit smarter than oQ4 on the non-QAT models.

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -1714,16 +1714,46 @@ def validate_quantizable(config: dict) -> bool:
 
     Models with 'quantization' key (mlx-lm quantized) are excluded.
     Models with 'quantization_config' are excluded UNLESS they are native FP8
-    (e.g. MiniMax, DeepSeek) which are full-precision models stored in FP8 format.
+    (e.g. MiniMax, DeepSeek) which are full-precision models stored in FP8 format,
+    or QAT-trained models (e.g. Google Gemma 4 QAT variants) whose
+    quantization_config records training-time settings but whose weights are
+    stored in full precision (bfloat16/float16).
     """
     if "quantization" in config:
         return False
     if "quantization_config" in config:
         qc = config["quantization_config"]
-        if isinstance(qc, dict) and qc.get("quant_method") == "fp8":
-            return True
+        if isinstance(qc, dict):
+            quant_method = qc.get("quant_method", "")
+            # FP8 models are full-precision weights stored in FP8 format
+            if quant_method == "fp8":
+                return True
+            # QAT models carry quantization_config with no recognised
+            # weight-quantization method — weights are full-precision.
+            if not quant_method:
+                return True
         return False
     return True
+
+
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for adding this. I think supporting Gemma 4 QAT unquantized checkpoints is useful: even though the weights are stored in half precision, they are trained with Q4_0 quantization in mind, so using them as an oQ4 source is a reasonable thing to enable.

I don't think this is ready to merge as-is yet, mainly for two implementation reasons.

First, I think the quantizable gate should be narrower. If the intent is to allow Gemma 4 QAT unquantized checkpoints, missing quant_method alone is not enough signal. That also allows ambiguous legacy configs such as {"bits": 4}, which the previous test explicitly treated as already quantized and blocked. Could you narrow this to the actual Gemma QAT unquantized config shape, for example by checking QAT-specific fields such as quant_type == "q4_0" together with the Gemma model/config layout, instead of allowing every quantization_config that lacks quant_method?

Second, the large-model path still appears uncovered. For models that exceed RAM, quantize_oq_streaming() goes through _build_proxy_for_sensitivity(), which calls mlx_lm.convert(). That convert path calls mlx_lm.load() without the model_config override added in _measure_sensitivity(), so large Gemma QAT models can still hit the same missing quant_method failure. This should be covered before merging, either by making the proxy path strip/override the QAT quantization_config too, or by adding a test that proves the proxy build path works for this config shape.

The feature direction looks useful to me, but I would like these two paths tightened before merge.

kreeger:
Thanks for reviewing this! I also totally missed the code path for models that exceed RAM.

I'll improve both of these and push an update.

kreeger:
Okay, both issues addressed.

If `mlx_lm` worked with the `"quant_method"` key like `mlx_vlm` does (using `.get(…)`), most of the patching workaround I had to use would be obsolete.

kreeger:
(Rebased.)

kreeger:
(Did another rebase due to a merge conflict in `tests/test_oq.py`.)

jundot:
Thanks for the updates. I rechecked the narrowed QAT gate and the sensitivity-load path; the missing-quant_method case is covered for both top-level and text_config layouts, and the large-model path is on the streaming proxy path rather than mlx_lm.convert().

CI is green and the focused oQ tests pass locally. This looks good to me, and I'm going to merge it.

PR #1694, opened 2026-06-05
fix(embedding): Respect embedding context length

### LINKED ISSUE #1687: Embeddings silently truncate beyond 512 tokens, and configured overrides are ignored; Embeddings effectively useless for content longer than a few sentences
**Describe the bug**

`POST /v1/embeddings` silently truncates embedding inputs at 512 tokens, even for embedding models whose config/tokenizer support much longer inputs.

This appears to be hard-coded in the embedding path rather than controlled by model settings. Setting `max_context_window` does not affect embeddings, and there does not appear to be an `embedding_max_length` / `max_length` option in config, model settings, CLI, or the OpenAI-compatible embeddings request schema.

For example, with `mlx-community/Qwen3-Embedding-8B-mxfp8`, inputs longer than ~512 tokens produce embeddings as if only the first ~512 tokens were provided. **There is no warning or log message** - I only discovered this by chance after wondering why my embedding model was seemingly useless half the time...

**To Reproduce**

1. Install/run oMLX 0.4.1 on macOS.
2. Load an embedding model, for example:
   - `mlx-community/Qwen3-Embedding-8B-mxfp8`
3. Set the model/global context window to a large value, e.g. `max_context_window: 200000`.
4. Send an embeddings request with an input longer than 512 tokens:
   ```bash
   curl http://localhost:8000/v1/embeddings \
     -d '{
       "model": "qwen3-embedding:8b",
       "input": "hello hello hello ... repeat enough times to exceed 512 tokens ..."
     }'
   ```
5. Compare the resulting embedding with the embedding for the same text truncated to 512 tokens. They match/behave as though the longer suffix was ignored.
6. No warning/error is emitted in logs.

**Expected behavior**

Embedding input length should not be silently capped at 512 tokens.

Ideally, oMLX should either:

- expose an embedding-specific setting such as `embedding_max_length`,
- respect the model’s configured max length / tokenizer max length where safe,
- allow `max_length` or equivalent on `/v1/embeddings`,
- or reject/log/warn when truncation occurs.

For this model, the downloaded metadata advertises much larger limits:

```json
// config.json
"max_position_embeddings": 40960

// tokenizer_config.json
"model_max_length": 131072
```

So a silent 512-token cap is surprising.

 - macOS Version: Tahoe
 - oMLX Version: 0.4.1

**Additional context**

I traced this through the source and it looks like the 512-token limit is hard-coded in the embedding path:

- `omlx/api/embedding_models.py`
  - `EmbeddingRequest` has `input`, `items`, `model`, `encoding_format`, and `dimensions`, but no `max_length` / `truncation` field.
- `omlx/server.py`
  - `/v1/embeddings` calls `engine.embed(embedding_inputs)` with no length override.
- `omlx/engine/embedding.py`
  - `EmbeddingEngine.embed(..., max_length: int = 512, truncation: bool = True)`
- `omlx/models/embedding.py`
  - `MLXEmbeddingModel.embed(..., max_length: int = 512, truncation: bool = True)`
  - passes `max_length=max_length` and `truncation=truncation` to tokenization / `mlx_embeddings.generate`.
- `mlx_embeddings.utils.generate`
  - also defaults to `max_length=512`.

`max_context_window` appears to be generation/chat-oriented and is not threaded into `EmbeddingEngine`. `embedding_batch_size` only controls the number of input texts per forward pass, not per-text token length.

This is especially easy to miss because the request succeeds and returns a normal embedding response.

jackwh:
Fixes #1687.

The embeddings endpoint now uses the model’s effective context window by default instead of falling through to the `mlx-embeddings` 512-token default. It also adds optional `max_length` and `truncation` request controls for callers who want to override that behavior explicitly.

### Details

- Adds `max_length` and `truncation` to `EmbeddingRequest`.
- Threads the effective embedding max length from the server into `EmbeddingEngine.embed()`.
- Resolves embedding max length from model/tokenizer metadata for direct model usage.
- Keeps `512` only as a final fallback when no model/tokenizer limit is known.
- Preserves explicit caller overrides, including intentionally passing `max_length: 512`.

### Testing

I ran the focused test coverage:

```bash
pytest tests/test_embedding.py tests/integration/test_server_endpoints.py::TestEmbeddingsEndpoint -q
```

Result:

```text
87 passed, 1 deselected
```

I also tested locally with `mlx-community/Qwen3-Embedding-8B-mxfp8`. The server logs now show:

```text
max_length=40960, truncation=True
```

and embeddings for long inputs with identical prefixes but different suffixes no longer collapse to the same vector/hash, confirming content beyond the old 512-token cutoff is being used.

### Backwards Compatibility

Existing embedding requests continue to work unchanged. The main behavioral change is that long inputs may now use more compute/memory because they are no longer silently capped at 512 tokens. Callers that want the previous cap can pass `max_length: 512` explicitly.

### DIFF
--- a/omlx/api/embedding_models.py
+++ b/omlx/api/embedding_models.py
@@ -58,6 +58,17 @@ class EmbeddingRequest(BaseModel):
     Only supported by some models. If not supported, returns full dimensions.
     """
 
+    max_length: Optional[int] = Field(default=None, gt=0)
+    """
+    Optional maximum token length for each input text. When omitted, the
+    server uses the model's effective context window.
+    """
+
+    truncation: bool = True
+    """
+    Whether to truncate inputs longer than max_length.
+    """
+
     @model_validator(mode="after")
     def validate_input_source(self) -> "EmbeddingRequest":
         """Require exactly one input source."""
--- a/omlx/engine/embedding.py
+++ b/omlx/engine/embedding.py
@@ -113,7 +113,7 @@ class EmbeddingEngine(BaseNonStreamingEngine):
     async def embed(
         self,
         texts: Union[List[str], List[Dict[str, str]]],
-        max_length: int = 512,
+        max_length: int | None = None,
         padding: bool = True,
         truncation: bool = True,
     ) -> EmbeddingOutput:
@@ -122,7 +122,8 @@ class EmbeddingEngine(BaseNonStreamingEngine):
 
         Args:
             texts: List of input texts
-            max_length: Maximum token length for each text
+            max_length: Maximum token length for each text. If omitted, the
+                model resolves its configured limit.
             padding: Whether to pad shorter sequences
             truncation: Whether to truncate longer sequences
 
--- a/omlx/models/embedding.py
+++ b/omlx/models/embedding.py
@@ -23,6 +23,16 @@ from .mlx_embeddings_compat import (
 
 logger = logging.getLogger(__name__)
 
+_DEFAULT_EMBEDDING_MAX_LENGTH = 512
+_TOKENIZER_MAX_LENGTH_SENTINEL = 10**18
+_CONTEXT_LENGTH_ATTRS = (
+    "max_position_embeddings",
+    "max_seq_len",
+    "max_seq_length",
+    "seq_length",
+    "n_positions",
+)
+
 
 @dataclass
 class EmbeddingOutput:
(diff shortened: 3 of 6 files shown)


jundot:
Thanks for fixing this. I verified that the endpoint now threads the effective embedding context length into the engine, preserves explicit max_length overrides, and covers the regression with focused tests.

One follow-up I may fold in later: custom embedding processors still use their own prepare_embedding_inputs() limits, so the new max_length/truncation request controls are not universal there. This does not block the reported Qwen3 text embedding issue.

Issue #1322, opened 2026-05-20
0.3.9rc1: Process RAM not released after cancelling HF download — subsequent model loads blocked until restart


nsanitas:
## Summary

After cancelling one or more in-progress HF downloads via `POST /admin/api/hf/cancel/{task_id}`, the oMLX server process keeps ~17–19 GB of RAM allocated, even with **zero models loaded**. The memory enforcer then refuses every subsequent model load because the projected total exceeds `max_model_memory`. The server is effectively unusable until a full restart.

Reported memory usage in `/api/status` shows `model_memory_used: 0B`, but the internal pre-load check sees `current: ~17.85 GB` and continues to grow over time (saw 17.85 → 18.55 → 18.85 GB across ~30 s of idle monitoring).

## Environment

| | |
|---|---|
| oMLX version | **0.3.9rc1** |
| Install method | Homebrew (`/opt/homebrew/Cellar/omlx/0.3.8/libexec/...` — note: brew formula 0.3.8, runtime reports 0.3.9rc1) |
| Hardware | Mac mini M4 Pro, 24 GB unified memory |
| OS | macOS Sequoia 26.5 |
| Runtime | headless server, no GUI menu bar app (no supervisor — `/admin/api/server/restart` returns "Server is not running under a supervisor that can respawn it") |

Note: `GET /admin/api/device-info` returns `"chip_name": "M1"` on this M4 Pro — likely a separate detection bug worth a separate issue.

## Steps to reproduce

1. Fresh-restart oMLX (clean state, one model pinned and loaded, ~15 GB RAM in use).
2. Start two large HF downloads in parallel:
   ```
   POST /admin/api/hf/download {"repo_id": "nightmedia/gemma-4-26B-A4B-it-mxfp4-mlx"}     # 13.6 GB
   POST /admin/api/hf/download {"repo_id": "mlx-community/gemma-4-26b-a4b-it-nvfp4"}      # 16.0 GB
   ```
3. Let the first download progress to ~20% (only one downloads at a time in practice; the second stays `pending`).
4. Cancel both downloads:
   ```
   POST /admin/api/hf/cancel/{task_id_1}
   POST /admin/api/hf/cancel/{task_id_2}
   ```
5. `POST /admin/api/hot-cache/clear`
6. Wait 10–30 s.
7. Attempt to load any model: `POST /admin/api/models/{any_id}/load`.

## Expected behavior

[...]

Issue #1106, opened 2026-05-07
Title: _extract_tensor_bytes SIGABRT on hybrid model (Qwen3.6-35B-A3B) with SSD cache — reproducible on 0.3.8.x and 0.3.9.dev1


lihywh:
Environment
[](url)
macOS 26 (Tahoe)
Apple M1 Max, 64 GB unified memory
oMLX versions affected: 0.3.8.x, 0.3.9.dev1 (not fixed)
Model: Qwen3.6-35B-A3B-oQ8 (VLM engine, hybrid Attention + GatedDeltaNet architecture)
SSD cache enabled, block_size auto-enlarged to 2048 for hybrid model
TurboQuant KV cache enabled (8-bit)
SpecPrefill: disabled

Crash location (identical across all occurrences)
Thread 0 (SIGABRT originator):
  paged_ssd_cache.py:216   _extract_tensor_bytes
  paged_ssd_cache.py:1210  save_block
  prefix_cache.py:573      store_cache
  scheduler.py:762         _async_store_cache_worker
Additional threads always present at crash time:

boundary_snapshot_store.py:318/321 _writer_loop (idle, waiting on queue) — new in 0.3.9.dev1
paged_ssd_cache.py:998/976 _writer_loop
scheduler.py:1460/1533 _do_external_prefill (regular prefill, not SpecPrefill)

Reproduction pattern
The crash is not immediate. It follows this sequence:

Session starts, first few requests complete normally (store_cache_main_eval ~28–50ms/call, stable)
After some requests, Walk-back truncation: dropped N trailing block(s) with placeholder non-sliceable state begins appearing in logs — this is the GatedDeltaNet layers producing non-sliceable state blocks
Shortly after walk-back truncations appear, _extract_tensor_bytes aborts
oMLX auto-restarts and crashes again at the same location

Observed time-to-crash after walk-back truncations begin: minutes to ~1 hour depending on session load.
On 0.3.9.dev1 specifically, one session crashed after only ~15 minutes of uptime.

Hypothesis
_extract_tensor_bytes calls .tobytes() (or equivalent) on an MLX array whose underlying Metal buffer has already been reclaimed by MLX's GC on the main thread. This is a tensor lifetime race between the async SSD writer thread and MLX's lazy evaluation/GC.
[...]

Issue #1060, opened 2026-05-05
[Bug] engine_pool active_memory counter leak after VLM model unload in 0.3.8


Enominera:
### Summary
After unloading a VLM model (gemma-4-31b-it-oQ6 via VLMBatchedEngine), `engine_pool` reports `active_memory: 59.04GB` instead of ~0B. This blocks subsequent loads of larger models with 507 errors.

### Environment
- **Version**: 0.3.8 (DMG)
- **macOS**: 15.5 (Sequoia), M-series, 128GB unified memory
- **Models involved**: gpt-oss-120b-heretic-v2-mxfp4-q8-hi-mlx (61.99GB) ↔ gemma-4-31b-it-oQ6 (26.12GB, VLM)

### Steps to reproduce
1. Load gpt-120b → infer → unload (works fine, active_memory: 72B)
2. Load gemma-4-31b-it-oQ6 (triggers VLMBatchedEngine + vision feature cache)
3. Infer on gemma
4. Unload gemma → **active_memory stuck at 59.04GB** instead of ~0
5. Any subsequent load of gpt-120b fails: `projected memory 121.02GB would exceed process limit 107.52GB (current: 59.04GB, model: 61.99GB)`

### Evidence from server.log (2026-05-05)

**Working cycles (LLM-only path, 8 successful):**
```
11:08:38 Unloaded: gpt-oss-120b, freed=59.04GB, active_memory: 72.00B (settled) ✅
11:08:52 Loaded: gemma-4-31b-it-oQ6 (estimated: 26.12GB)
11:09:05 Unloaded: gemma-4-31b-it-oQ6, freed=24.88GB, active_memory: 59.04GB (settled) 🚨
```

The leaked value `59.04GB` is exactly the `freed` value from the previous gpt-120b unload — appears to be a stale counter that wasn't reset.

**507 cascade after leak:**
```
11:09:05 → 507: Cannot load gpt-oss-120b: projected 121.02GB > limit 107.52GB (current: 59.04GB)
11:09:06 → 507: (same)
11:32:48 → 507: (same, 23 min later — memory never recovered)
11:45:48 → server shutdown by user
```

### Key observation
All 8 successful cycles used the batched LLM engine only. The one that leaked triggered `VLMBatchedEngine` + vision feature cache initialization. The `engine_pool` active_memory counter appears to not correctly track memory freed when the VLM engine path is involved.

### Workaround
Restart the server (not just switch models) after any VLM inference to reset the engine_pool counter.

[...]

PR #1713, opened 2026-06-07
fix: reclaim hot cache memory retained after model unload

khsd6327:
After a model is unloaded the server can keep tens of GB of resident memory
belonging to a hot cache no longer attached to any model. The memory guard
counts that footprint against the load ceiling, so the next load is rejected
with `507`, and `POST /api/hot-cache/clear` returns ok but frees nothing. The
only recovery is a process restart.

## Reproduction

1. Serve a model and send a large request so the hot cache fills (a prompt
   around 100k tokens).
2. Let an abort and an unload interleave, as a multi step workflow produces, or
   otherwise unload through a path where engine teardown does not complete.
3. `/api/status` shows `loaded_count: 0` and `current_model_memory: 0`, yet the
   process still holds a large resident set (about 57 GB in my case).
4. `POST /api/hot-cache/clear` returns ok but resident memory does not move.
5. Loading a larger model is rejected with `507`:

```
507: Cannot load gemma-4-31B-it: projected memory 129.88GB would exceed the
memory ceiling 107.52GB (current: 68.72GB, model: 61.16GB).
```

`current` in the guard is `max(mx.get_active_memory(), get_phys_footprint())`,
so the retained footprint counts even with nothing loaded. A clean unload
reclaims fully, so this only appears when teardown does not complete.

## Changes (three commits)

1. `fix(engine): always release SSD cache manager on engine close`.
   `EngineCore.close()` only caught `RuntimeError` from `scheduler.shutdown` and
   `deep_reset`, so any other exception aborted `close()` before it reached
   `PagedSSDCacheManager.close()`. The manager's writer thread then kept it and
   its hot cache alive until restart. Swallow non-`RuntimeError` from
   `shutdown`/`deep_reset`, then close the SSD manager explicitly if it is still
   set. This removes the cause of the orphan on the teardown path.

2. `feat(cache): reclaim hot cache orphaned by an abnormal teardown`.
   `SharedHotCacheBudget` keeps a strong reference to each manager through
   `_HotCacheBudgetEntry.owner`, so a manager orphaned by a skipped `close()` is
   unreachable through the loaded-only iterator. Add `clear_all_owners()`, which
   clears the hot cache of every manager the budget still references and logs
   per-owner failures instead of swallowing them. The strong owner reference was
   added in `6dcf197`.

3. `fix(admin): reclaim hot cache memory on clear via the synchronized path`.
   `clear_hot_cache` now reclaims MLX's buffer pool after clearing the hot cache
   dicts, and calls `clear_all_owners()` so orphans are reached through the
   budget. The reclaim runs `scheduler._sync_and_clear_cache(stream)` on each
   loaded engine's own executor, targeting that engine's `_stream`, which is the
   same `_mx_buffer_access_lock` and `synchronize` barrier generation uses
   (issues #300, #1106). With no model loaded it falls back to the global
   executor. It returns `bytes_reclaimed`, measured before the dicts are cleared
   so raw host-byte entries are counted.

## Safety

The reclaim no longer calls bare `mx.clear_cache()` on the global executor.
`_sync_and_clear_cache` holds `_mx_buffer_access_lock` and synchronizes the
engine stream and the default stream before `clear_cache()`, so it does not
release buffers still referenced by an active engine stream or read by the async
store-cache worker (issues #300, #1106, #435). Running it on each engine's own
executor mirrors how `scheduler` reclaims internally.

## Test plan

1. `pytest tests/test_admin_hot_cache_clear.py tests/test_paged_ssd_cache.py tests/test_engine_core.py`
2. Full suite.
3. Real MLX smoke: an orphaned manager left in the budget with no model loaded
   is reclaimed through the synchronized path, and `mx.get_cache_memory()` drops
   to zero.

## Related

Same symptom, different trigger: #1322 (RAM not released after a cancelled
download, loads blocked until restart) and #1060 (memory stuck after a VLM
unload, subsequent loads fail with `507`). PR #1593 (merged) fixed the adjacent
teardown case where streaming generators held the engine ref after stop; the
`close()` gap fixed here is in the same teardown path. The strong owner
reference that pins an orphaned manager was introduced in `6dcf197`.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -4352,12 +4352,11 @@ async def clear_alltime_stats(is_admin: bool = Depends(require_admin)):
     return {"status": "ok"}
 
 
-def _iter_loaded_schedulers():
-    """Yield (model_id, scheduler) for each loaded model.
+def _iter_loaded_scheduler_records():
+    """Yield (model_id, scheduler, core) for each loaded model.
 
     Traverses the internal engine hierarchy: pool entry → async engine →
-    core engine → scheduler.  Both ``clear_ssd_cache`` and
-    ``clear_hot_cache`` share this traversal.
+    core engine → scheduler.
     """
     engine_pool = _get_engine_pool()
     if engine_pool is None:
@@ -4373,7 +4372,16 @@ def _iter_loaded_schedulers():
         core = getattr(async_core, "engine", None) if async_core is not None else None
         scheduler = getattr(core, "scheduler", None) if core is not None else None
         if scheduler is not None:
-            yield model_id, scheduler
+            yield model_id, scheduler, core
+
+
+def _iter_loaded_schedulers():
+    """Yield (model_id, scheduler) for each loaded model.
+
+    Both ``clear_ssd_cache`` and ``clear_hot_cache`` share this traversal.
+    """
+    for model_id, scheduler, _core in _iter_loaded_scheduler_records():
+        yield model_id, scheduler
 
 
 @router.post("/api/ssd-cache/clear")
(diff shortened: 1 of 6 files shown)


jundot:
Thanks for the careful fix and for documenting the failure mode so clearly. I checked the teardown path and the admin hot-cache clear path, and this closes the gap where retained hot-cache memory could survive after unload and block the next load.

The synchronized reclaim path and the shared-budget orphan cleanup fit the existing cache safety constraints well. This looks good to me, and I'm going to merge it. I'll handle one small cleanup separately after merge.

PR #1715, opened 2026-06-07
feat: add vision/image support to omlx OpenAI API endpoint

Revive-Curiosity:
## Summary

- **Add multimodal detection helpers** (`_has_multimodal_content_list()`, `_has_multimodal_messages()`) in `api/utils.py` to detect image_url parts in messages
- **Route image requests to multimodal extraction** in `server.py` — requests containing images are now routed through `extract_multimodal_content()` even for non-VLM engines, preserving image_url parts instead of silently dropping them
- **Add vision processing to BatchedEngine** — new `_apply_chat_template_with_vision()` method integrates mlx-vlm for image preprocessing and vision embedding computation when images are present in messages
- **Modified chat(), stream_chat(), generate(), stream_generate()** to accept and pass vision embeddings (`vlm_inputs_embeds`, `vlm_extra_kwargs`) to the inference engine

## Problem

The built-in chat UI at `/admin/chat` supports vision — users can upload images and get responses from VLM models. However, when using the OpenAI-compatible API (`/v1/chat/completions`), images were silently dropped for models that weren't classified as VLM type.

The root cause was that `extract_text_content()` in `api/utils.py` only kept `type=="text"` content parts, dropping all `image_url` parts. Additionally, the engine selection was model-type based (not request-based), so non-VLM models never received image data.

## Solution

1. Detect multimodal content in the original request before message extraction
2. Route image-containing requests to `extract_multimodal_content()` regardless of engine type
3. Extend `BatchedEngine` with vision processing capabilities that use mlx-vlm's `prepare_inputs()` and `model.get_input_embeddings()` when images are detected
4. Pass vision embeddings through the generation pipeline

## Files Changed

| File | Changes |
|------|---------|
| `omlx/api/utils.py` | +41 lines — multimodal detection helpers |
| `omlx/server.py` | +14/-2 lines — image detection and routing |
| `omlx/engine/batched.py` | +311/-21 lines — vision processing integration |

## Backward Compatibility

- Non-image requests work exactly as before
- VLM models continue using VLMBatchedEngine unchanged
- Non-VLM models without images use standard text path unchanged
- Graceful fallback to text-only if mlx-vlm is unavailable or model doesn't support vision

### DIFF
--- a/omlx/api/utils.py
+++ b/omlx/api/utils.py
@@ -118,6 +118,47 @@ def _extract_text_from_content_list(content: list) -> str:
     return "\n".join(text_parts) if text_parts else ""
 
 
+def _has_multimodal_content_list(content: list) -> bool:
+    """Check if a content array contains image or audio parts.
+
+    Useful for detecting multimodal requests before choosing the
+    appropriate extraction path (text-only vs multimodal).
+
+    Args:
+        content: Content array from a message.
+
+    Returns:
+        True if the content contains image_url, input_image, input_audio, etc.
+    """
+    for item in content:
+        if hasattr(item, "model_dump"):
+            item = item.model_dump()
+        elif hasattr(item, "dict"):
+            item = item.dict()
+        if isinstance(item, dict):
+            item_type = item.get("type")
+            if item_type in ("image_url", "input_image", "input_audio", "image"):
+                return True
+    return False
+
+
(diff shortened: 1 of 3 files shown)


Revive-Curiosity:
I wished omlx openAI API endpoint gave vission support for my local coding tasks
- I used LM Studio before for vision tasks

Please add vision support to openAI API endpoint in omlx!

PR #1716, opened 2026-06-07
fix(scheduler): count chunked prefill against concurrency cap

### LINKED ISSUE #1704: Not respecting Max Concurrent Requests
**Describe the bug**

Not sure if has to do with Max Concurrent Requests feature in Global Settings but I have this property with value in 1 due my hardware, M5 Pro 48 GB and using Jundot/Qwen3.6-27B-oQ4-mtp model , but since version 4.0.1 I have seeing that in Dashboard/Active models section more than 1 active model, I thought this shouldn't because of the  Max Concurrent Requests = 1.

**To Reproduce**

I'm using VS Code in Agents mode, this are my settings in VS Code perhaps there is a setting that can be causing it.

```
"name": "oMLX-host",
		"vendor": "customendpoint",
		"apiKey": "${input:chat.lm.secret.49570750}",
		"apiType": "chat-completions",
		"models": [
			{
				"id": "Qwen3.6-27B-oQ4-mtp",
				"name": "Qwen3.6-27B-GEH-oMLX",
				"url": "http://127.0.0.1:8000/v1/chat/completions",
				"toolCalling": true,
				"vision": true,
				"thinking": true,
				"maxInputTokens": 65536,
				"maxOutputTokens": 8192,
				"multimodal": true,
				"streaming": true,
				"reasoningEffortFormat": "chat-completions",
				"reasoningEffortIncludeToolCalls": true,
				"zeroDataRetentionEnabled": true,
				"supportsReasoningEffort": [
					"low",
					"medium",
					"high"
				]
			}

```

**Expected behavior**

Image

Image

**Screenshots**
Issue

Image

 - macOS Version: 26.5.1 (25F80)
 - oMLX Version [v0.4.1]

**Additional context**
Also just mention that I have MTP enabled for this model in the settings and also the SpecPrefill feature

Image

jsiu93:
Fixes #1704.

## Summary
With `max_concurrent_requests=1`, a second long request could start while the first request was still in chunked prefill. The scheduler only counted decode-stage `running` requests, so `prefilling` work stopped consuming an admission slot until it reached decode.

This PR counts both `running` and `prefilling` requests as admitted scheduler work. It also makes the Active Models dashboard use the scheduler snapshot for active vs waiting counts, so queued collectors are shown as waiting instead of active.

## Root cause
Chunked prefill keeps long prompts in `Scheduler.prefilling` between prefill chunks. `_schedule_waiting()` compared only `len(self.running)` against `max_num_seqs`, which let another waiting request begin prefill even when the configured concurrency cap was already occupied by an in-flight prefill.

The dashboard had the same state-model mismatch in the other direction: it treated every output collector as active even after a scheduler snapshot identified the request as waiting.

## The fix
- Add a small admitted-work helper that counts `running` plus `prefilling`.
- Use that count for waiting-queue admission and the memory-pressure guard.
- Derive dashboard active request IDs from scheduler `running_by_id` plus `prefilling_ids` when a scheduler snapshot exists.
- Add regression coverage for chunked-prefill admission and queued dashboard collectors.

## Scope / safety
The change stays inside the existing scheduler admission gate and admin snapshot rendering. It does not change request execution, cache layout, API schemas, or non-chunked prefill behavior; it only makes the configured concurrency cap include work that is already admitted but still prefilling.

## Test plan
```bash
.venv/bin/python -m pytest tests/test_scheduler_chunked_prefill.py tests/test_active_models_visibility.py -q
# 52 passed

.venv/bin/python -m pytest tests/test_scheduler.py tests/test_scheduler_admission.py tests/test_status_endpoint.py -q
# 144 passed

.venv/bin/python -m pytest tests/test_server_metrics.py -q
# 29 passed

.venv/bin/python -m pytest tests/ -m "not slow and not integration" -q
# 5169 passed, 40 skipped, 66 deselected
```

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -4115,12 +4115,14 @@ def _build_active_models_data() -> dict:
         active_requests = 0
         waiting_requests = 0
         running_by_id = {}
+        has_scheduler_snapshot = False
         waiting_ids = set()
         waiting = []
         activities = []
 
         # Get per-model active/waiting request counts.
         # Follow the same pattern as server.py /api/status endpoint.
+        collector_request_ids: set = set()
         active_request_ids: set = set()
         entry = engine_pool._entries.get(model_id)
         if entry and entry.engine is not None:
@@ -4130,18 +4132,17 @@ def _build_active_models_data() -> dict:
                 if core is not None:
                     collectors = getattr(core, "_output_collectors", {})
                     try:
-                        active_request_ids = set(collectors.keys())
-                        active_requests = len(collectors)
+                        collector_request_ids = set(collectors.keys())
                     except RuntimeError:
                         # Scheduler state is mutated from the engine executor;
                         # keep the dashboard endpoint best-effort rather than
                         # failing on a concurrent dict resize.
-                        active_request_ids = set()
-                        active_requests = len(collectors)
+                        collector_request_ids = set()
 
                     sched = getattr(core, "scheduler", None)
                     if sched is not None and hasattr(sched, "snapshot_for_admin"):
                         snap = sched.snapshot_for_admin()
+                        has_scheduler_snapshot = True
                         running_by_id = snap["running_by_id"]
                         waiting_queue = snap["waiting"]
                         waiting_requests = len(waiting_queue)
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for the fix. The root cause checks out: chunked prefill is already admitted scheduler work, but the admission gate was only counting decode-stage `running` requests.

This PR now conflicts with recent scheduler admission changes on `main`, so I'll carry the commit onto current `main` with your authorship preserved, resolve it against the newer `_effective_max_num_seqs()` path, run the focused scheduler/admin tests, and land it from there.

jundot:
Thanks again, @jsiu93. I landed this on current `main` as 3732bcb with your authorship preserved, resolving the conflict against the newer scheduler admission path while keeping `_effective_max_num_seqs()` in place.

Since the fix is now on `main`, I'm closing this PR.

PR #1718, opened 2026-06-07
fix(embeddings): fall back to model context length when no window is set

JimStenstrom:
Follow-up to #1694, which resolved the 512-token embedding truncation from #1687.

#1694's `get_embedding_max_length()` falls back to a hard `512` when neither the request nor the server's `max_context_window` pins a limit. In that no-window case it re-introduces the original #1687 symptom: a long-context embedding model is capped at 512 even though `MLXEmbeddingModel._resolve_max_length()` can read its real limit from `max_position_embeddings` / tokenizer `model_max_length`.

That model-side resolver only runs when it receives `max_length=None`, but the server always hands it a concrete `512` first — so for `/v1/embeddings` it never fires, and the config-aware resolution added in #1694 is effectively bypassed.

**Change:** return `None` from `get_embedding_max_length()` when nothing pins a limit, letting the existing model resolver do its job. `512` is preserved as that resolver's final fallback when the model exposes no usable metadata, so the conservative default is unchanged.

- `get_embedding_max_length()` now returns `int | None`; its single caller passes the value straight to `embed(max_length=...)`, which already accepts `None`.
- Added `TestGetEmbeddingMaxLength` covering the request-override, configured-window, and no-window (returns `None`) paths.

Verification: `pytest tests/test_embedding.py` → 83 passed.

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -1217,16 +1217,19 @@ def get_max_context_window(model_id: str | None = None) -> int | None:
 def get_embedding_max_length(
     model_id: str | None = None,
     request_max_length: int | None = None,
-) -> int:
-    """Get max token length for embedding requests."""
+) -> int | None:
+    """Get max token length for embedding requests.
+
+    Returns ``None`` when neither the request nor the server's
+    ``max_context_window`` pins a limit, so the embedding model resolves its
+    own configured context length (``max_position_embeddings`` / tokenizer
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the follow-up. I checked the server and embedding engine path, and passing None here is safe because MLXEmbeddingModel resolves the model/tokenizer limit and still keeps 512 as the final fallback when no metadata exists.

The request override and configured context-window paths stay unchanged, and CI is green. This looks good to me, and I'm going to merge it.

Issue #1612, opened 2026-06-02
Setting Hot Cache Size to "auto" Gives 500 Server error - 0.4.0rc2


miabchdave:
Error upon saving "auto" for **Hot Cache Size** as instructed by label text for the setting.

**To Reproduce**

1. Started the server and activated Settings from the menu icon.  
2. Go to Performance > Hot Cache Size
3. Enter "auto" as instructed in label text
4. Clicking apply displays Server Error 500 (see image)
5. Entering "0" works

**Screenshot**
<img width="667" height="295" alt="Image" src="https://github.com/user-attachments/assets/3994825d-f991-4654-b6a5-36548a398075" />

**Desktop (please complete the following information):**
 - macOS Version: 26.5
 - oMLX Version:  0.4.0rc2

jundot:
Thanks for the clear report, @miabchdave.

I fixed this on `main` in 4826bbe. Hot Cache Size no longer advertises `auto`; existing saved `auto` values are migrated to `0`, and direct API submissions now return a 400 instead of a 500.

`0` remains the correct value to disable hot cache.

miabchdave:
Small suggestion, perhaps change the label text to "..."0" disables **_ceiling_**" as the current reading can be confused with disables hot cache entirely (using only SSD).

seeker-cyber-maker:
Fresh data point from the macOS app side: I hit the same confusing edge in `0.5.0.dev1` while tuning cache settings.

What I observed:

- the GUI still made it easy to think Hot Cache Size had an `auto` mode
- when that value is persisted/normalized, it becomes `0`
- `0` then disables the in-memory hot cache tier, leaving only the cold SSD tier

Looking at current `main`, this seems intentional now rather than a parser bug:

- `CacheSettings.from_dict()` migrates legacy `hot_cache_max_size="auto"` to `"0"`
- `GlobalSettings.validate()` rejects hot-cache `auto`
- the web dashboard normalizes blank/`auto` hot cache values to `0`
- PR #2095 removes stale “auto” wording from the Hot Cache Size copy

So the remaining UX issue is probably just making the distinction extra explicit everywhere: `auto` is valid for SSD/cold cache sizing, but not for hot cache sizing. Hot cache is explicit size or `0` disabled.

A real “auto hot cache” mode would be a separate feature, since it would need a policy like bounded percentage of RAM/Metal ceiling with a floor/ceiling, not just accepting the string `auto`.

PR #1719, opened 2026-06-07
fix(server): return 401 instead of 500 when an API key contains non-ASCII characters

### LINKED ISSUE #1717: HTTP 500: "comparing strings with non-ASCII characters is not supported" on all API endpoints after first inference
## Describe the bug

All REST API endpoints (`/v1/models`, `/v1/chat/completions`) return HTTP 500 with error: `comparing strings with non-ASCII characters is not supported`.

## To Reproduce

1. Install oMLX v0.4.2rc1 on macOS (Darwin 24/25, Apple Silicon)
2. Load model: `Qwen3.6-40B-Claude-4.6-Opus-Deckard-Heretic-Uncensored-Thinking-8bit` (8-bit MLX)
3. Start server: `omlx start`
4. Send first request — model loads, inference works:

```bash
openclaw infer model run --model "omlx/Qwen3.6-40B-..." --prompt "say hi"
# Returns: "Hi there! 👋" — SUCCESS
```

5. Immediately after, send another request via curl:

```bash
curl -H 'Authorization: Bearer <key>' \
     http://localhost:8000/v1/chat/completions \
     -d '{"model":"Qwen3.6-40B-...","messages":[{"role":"user","content":"hello"}],"max_tokens":5}'
# Returns: HTTP 500 — "comparing strings with non-ASCII characters is not supported"
```

6. `/v1/models` also returns 500. All subsequent requests fail until server restart.

## Actual behavior

**Server log:**
```
omlx.server - ERROR - GET /v1/models → 500 (unhandled): comparing strings with non-ASCII characters is not supported
omlx.server - ERROR - POST /v1/chat/completions → 500 (unhandled): comparing strings with non-ASCII characters is not supported
```

First request after restart sometimes succeeds (model loads, inference works at ~14 tok/s), but the server state becomes corrupted immediately after — all subsequent requests fail with the non-ASCII comparison error.

## Expected behavior

All requests should succeed after model is loaded. The error should not occur in Python 3.11.

## Environment

- **oMLX**: v0.4.2rc1 (same on v0.4.1)
- **macOS**: 25.4.0 (arm64, Mac Studio)
- **Python**: 3.11 (bundled)
- **Model**: `mlx-community/Qwen3.6-40B-Claude-4.6-Opus-Deckard-Heretic-Uncensored-Thinking-8bit` (38.94 GB loaded)
- **Model dir**: `~/.omlx/models/mlx-community/`

## Additional context

- Bug persists with cache disabled (`cache.enabled: false`)
- Bug persists after cache directory cleared
- Bug persists after complete reinstall (v0.4.1 → v0.4.2rc1)
- Both models affected (Huihui-Qwen3.6-35B-A3B and Qwen3.6-40B)
- OpenClaw integration affected — agent loop fails, only simple `infer` commands work
- OMLX Dashboard (web UI) continues to work fine

**Timeline from server.log (full cycle):**
```
13:17:00 Engine loaded: Qwen3.6-40B-... (38.94GB)
13:17:04 Chat completion: 44 tokens in 4.11s (14.0 tok/s) ← WORKS
13:17:16 GET /v1/models → 500: comparing strings with non-ASCII... ← BROKEN
13:17:16 POST /v1/chat/completions → 500: comparing strings... ← BROKEN
13:17:22 Both fail again
```

The error is likely in MLX-lm's string comparison during batch processing or KV cache management, triggered after the first inference introduces Unicode token content into the runtime state.

richgoodson:
Fixes #1717.

## Problem

Any request whose Bearer token or `x-api-key` header contains a non-ASCII character makes every authenticated endpoint return HTTP 500 with "comparing strings with non-ASCII characters is not supported". The report shows this on both `/v1/models` and `/v1/chat/completions`.

## Root cause

This is not an inference or KV cache problem. `/v1/models` fails too, and that endpoint never touches the engine.

`verify_api_key` in `server.py` passes the client-supplied key into `secrets.compare_digest(str, str)` in `omlx/admin/auth.py`. CPython's `compare_digest` raises `TypeError` when either `str` argument contains non-ASCII characters, and FastAPI surfaces that as an unhandled 500.

A non-ASCII key is easy to produce by accident; pasting a key through anything that substitutes smart quotes is enough. `validate_api_key` also accepts non-ASCII printable characters, so a configured main key or sub key can trigger the same TypeError on every request.

## Fix

Added `compare_keys` to `omlx/admin/auth.py`. It encodes both sides as UTF-8 and compares bytes, which `compare_digest` accepts for any content, and keeps the constant-time property. All key comparisons (`verify_api_key`, `verify_any_api_key`, and the sub-key duplicate and delete checks in `admin/routes.py`) now go through it. Removed the now-unused `secrets` import in `routes.py`.

Encoding uses `surrogatepass` because `json.loads` can produce lone surrogates from escape sequences, and strict UTF-8 encoding rejects those. That matters for `delete_sub_key`, which compares `request.key` without a `validate_api_key` gate; with strict encoding that route would have kept its 500.

## Testing

- 8 new regression tests: 7 in `tests/test_api_auth.py` (helper behavior for non-ASCII and lone surrogates on either side, and the server auth dependency returning 401, not TypeError, for a non-ASCII bearer) and 1 in `tests/test_admin_api_key.py` (`delete_sub_key` with a lone-surrogate key returns 404, not 500).
- Full suite: 5175 passed, 0 failed, 40 skipped, 66 deselected.
- Live test: started a server with `--api-key`, sent non-ASCII `Bearer` and `x-api-key` values to `/v1/models` and `/v1/chat/completions`. All returned 401 with the standard error body, the correct key returned 200, and the server log contained no unhandled errors.

One thing I could not reproduce is the "first request works, then everything fails" sequence in the report. The 500 is per-request and depends only on the bytes the client sends. My read is that the failing requests carried a non-ASCII key the whole time and the successful first request used a different stored credential. Either way the fix removes the 500.

## Scope note

This does not make configured non-ASCII keys usable over HTTP. Starlette decodes header values as latin-1, so a client sending UTF-8 non-ASCII bytes will never match a configured non-ASCII key; it now gets a clean 401 instead of a 500. If you would rather also reject non-ASCII keys at creation time (400 going forward, migrate existing, the same shape as the hot-cache "auto" fix in #1612), I am happy to do that as a follow-up.

### DIFF
--- a/omlx/admin/auth.py
+++ b/omlx/admin/auth.py
@@ -109,10 +109,35 @@ def verify_session_token(token: str, max_age: int = SESSION_MAX_AGE) -> bool:
         return False
 
 
+def compare_keys(provided_key: str, expected_key: str) -> bool:
+    """Compare two API keys in constant time, tolerating any str input.
+
+    secrets.compare_digest raises TypeError when given str arguments that
+    contain non-ASCII characters, which turns a bad client key into an
+    unhandled 500 instead of a 401. Comparing UTF-8 bytes accepts any
+    input while keeping the constant-time guarantee. surrogatepass covers
+    lone surrogates, which json.loads can produce from escape sequences
+    and which strict UTF-8 encoding rejects.
+
+    Both arguments must be str; None is the caller's responsibility.
+
+    Args:
+        provided_key: The key supplied by the client (untrusted).
+        expected_key: The configured key to compare against.
+
+    Returns:
+        True if the keys match, False otherwise.
+    """
+    return secrets.compare_digest(
+        provided_key.encode("utf-8", "surrogatepass"),
+        expected_key.encode("utf-8", "surrogatepass"),
+    )
+
+
 def verify_api_key(api_key: str, server_api_key: str) -> bool:
     """Verify an API key using constant-time comparison.
 
-    This function uses secrets.compare_digest to prevent timing attacks
+    This function uses constant-time comparison to prevent timing attacks
     when comparing the provided API key with the server's API key.
 
     Args:
@@ -130,7 +155,7 @@ def verify_api_key(api_key: str, server_api_key: str) -> bool:
     """
     if not api_key or not server_api_key:
         return False
-    return secrets.compare_digest(api_key, server_api_key)
+    return compare_keys(api_key, server_api_key)
 
 
 def verify_any_api_key(api_key: str, main_key: str, sub_keys: list) -> bool:
@@ -150,11 +175,11 @@ def verify_any_api_key(api_key: str, main_key: str, sub_keys: list) -> bool:
     if not api_key:
         return False
     # Check main key
-    if main_key and secrets.compare_digest(api_key, main_key):
+    if main_key and compare_keys(api_key, main_key):
         return True
     # Check sub keys
     for sk in sub_keys:
-        if sk.key and secrets.compare_digest(api_key, sk.key):
+        if sk.key and compare_keys(api_key, sk.key):
             return True
     return False
 
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -14,7 +14,6 @@ import json
 import logging
 import os
 import re
-import secrets
 import shutil
 import signal
 import subprocess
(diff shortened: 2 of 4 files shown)


jundot:
Thanks for the fix. I verified the failure path: the public API auth dependency and the admin key comparisons now go through compare_keys, and the remaining compare_digest usage is contained inside that helper. The surrogatepass handling also covers the JSON-route edge case without changing normal ASCII behavior. This looks good to me, and I'm going to merge it.

PR #1720, opened 2026-06-07
fix(embedding): serve Qwen2ForCausalLM embedding models natively (#686)

JimStenstrom:
Fixes #686.

## Summary

Serve Qwen2-based embedding models (`architectures: ["Qwen2ForCausalLM"]`,
`model_type: qwen2`) end to end — e.g. `jinaai/jina-code-embeddings-1.5b` (the
issue model), `Alibaba-NLP/gte-Qwen2-1.5B-instruct`. The first commit classified
them as embeddings; this adds a native adapter so they actually load and embed.

## Root cause

Classification alone routed these models to `MLXEmbeddingModel`, whose native
path only handled BERT/XLM-RoBERTa and whose `mlx-embeddings` fallback has no
`qwen2` module, so loading raised the exact error from the issue:

```
ValueError: Model type qwen2 not supported.
```

## Fix

- **`omlx/models/qwen2_embedding.py`** (new): a Qwen2 decoder wrapped for
  embeddings with **mask-aware last-token pooling + L2 normalization**, matching
  these models' SentenceTransformers configs (`pooling_mode_lasttoken: true` +
  a Normalize module). Mean pooling would silently corrupt every vector; the
  pool is mask-aware so left- and right-padded batches pool the same real token.
  Qwen2 deltas vs the Qwen3 embedder: no QK-norm, attention bias on q/k/v
  (`o_proj` unbiased), `head_dim = hidden_size // num_attention_heads`.
  An `is_causal` flag (default true) runs the decoder **causally** for plain
  decoder-embedders (jina-code) and **bidirectionally** when a model
  declares `is_causal: false` (Alibaba's gte-Qwen2 family ships a custom
  bidirectional `modeling_qwen.py`). The padding mask uses the dtype's finite
  min, not `-inf`, so fully-masked rows under left padding stay NaN-free.
- **`omlx/models/embedding.py`**: native dispatch is now an arch→module map
  (`Qwen2ForCausalLM` → the adapter; BERT/XLM unchanged). No change to the
  weight loader — `mx.load` already reads the bf16 checkpoints these models ship.
- **`omlx/models/base_model.py`**: a mask-aware `last_token_pool` helper.

## Sibling sweep

- Covered: single + batch input, left/right padding (mask-aware), causal
  (jina-code) and bidirectional (gte-Qwen2) Qwen2 embedders, float + base64
  output (unchanged downstream).
- Untouched: the `mlx-embeddings` fallback (Qwen3 embedders) and the native
  BERT/XLM path are not modified by this PR. The Qwen2 reranker is out of scope
  (separate `reranker.py` path, untouched).

## Tests

`tests/test_embedding.py` (`TestNativeQwen2Embedding`) drives the **real**
`_load_native` + `embed()` path for a `Qwen2ForCausalLM` fixture: native-path
routing, output shape + L2-norm, mask-aware last-token (left vs right padding
agree while a hardcoded `[-1]` pool diverges by ~0.2), and `is_causal` flipping
causal↔bidirectional. All red without the adapter (they fail with the issue's
`Model type qwen2 not supported`). `pytest tests/test_embedding.py
tests/test_model_discovery.py` → 219 passed.

## Real-model verification

vs a `transformers` reference (stock `Qwen2Model`, fp32, mask-aware last-token + L2):

| model | attention | oMLX↔ref mean cosine | ranking (relevant vs irrelevant) |
|---|---|---|---|
| jina-code-embeddings-1.5b | causal | 0.99990 | pass (Δ +0.75 / +0.62) |
| gte-Qwen2-1.5B-instruct | bidirectional (`is_causal:false`) | 0.99987 | pass (Δ +0.16 / +0.11) |

Zero-regression: Qwen3-Embedding-0.6B (the `mlx-embeddings` fallback) is
byte-identical before/after (0.0), and Qwen2.5-0.5B chat still classifies as
`llm`.

## Dependencies

No dependency change. `mlx-embeddings` has no `qwen2` module at its pin, at HEAD,
or in any release, so a bump won't help; an upstream PR adding `qwen2.py` to
`mlx-embeddings` is a sensible follow-up but not required here.

### DIFF
--- a/omlx/model_discovery.py
+++ b/omlx/model_discovery.py
@@ -147,6 +147,9 @@ CAUSAL_LM_RERANKER_ARCHITECTURES = {
 # (no lm_head weights). Detected by architecture + directory name heuristic.
 CAUSAL_LM_EMBEDDING_ARCHITECTURES = {
     "Qwen3ForCausalLM",  # Qwen3-Embedding uses CausalLM arch without lm_head
+    "Qwen2ForCausalLM",  # jina-code-embeddings & similar; only treated as an
+    # embedding when the dir-name heuristic (_is_causal_lm_embedding) also
+    # matches, so Qwen2/Qwen2.5 chat models are unaffected.
 }
 
 # Multimodal (VLM-based) reranker architectures.
--- a/omlx/models/base_model.py
+++ b/omlx/models/base_model.py
@@ -59,6 +59,41 @@ def mean_pooling(hidden_states: mx.array, attention_mask: mx.array) -> mx.array:
     return sum_embeddings / sum_mask
 
 
+def last_token_pool(
+    hidden_states: mx.array, attention_mask: Optional[mx.array] = None
+) -> mx.array:
+    """
+    Pool the last *non-pad* token of each sequence (mask-aware).
+
+    Decoder embedding models (Qwen2/Qwen3, gte-Qwen2, jina-code) represent a
+    sequence by its final token's hidden state. The pool must be
+    mask-aware because these tokenizers may pad on either side: a hardcoded
+    ``[:, -1]`` is only correct under left padding and silently corrupts vectors
+    under right padding.
+
+    Args:
+        hidden_states: Shape (batch_size, seq_len, hidden_size)
+        attention_mask: Shape (batch_size, seq_len). If None, uses the last
+            position.
+
+    Returns:
+        Pooled output of shape (batch_size, hidden_size)
+    """
+    if attention_mask is None:
+        return hidden_states[:, -1]
+
+    # Left padding: every sequence ends on a real token, so the last position
+    # is always valid and we can index it directly.
+    left_padding = attention_mask[:, -1].sum() == attention_mask.shape[0]
+    if left_padding:
+        return hidden_states[:, -1]
+
+    # Right (or mixed) padding: the last valid token is at sum(mask) - 1.
+    sequence_lengths = attention_mask.sum(axis=1) - 1
+    batch_size = hidden_states.shape[0]
+    return hidden_states[mx.arange(batch_size), sequence_lengths]
+
+
 def normalize_embeddings(embeddings: mx.array) -> mx.array:
     """
     L2 normalize embeddings.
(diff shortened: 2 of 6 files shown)


JimStenstrom:
Rebased onto current `main` and CI is green across 3.11–3.14 — and this is worth
a fresh look, because it was incomplete while it sat.

The version that was up only had the classification commit, which isn't enough on
its own: once a Qwen2 model is classified as an embedder it routes to
`MLXEmbeddingModel`, and `mlx-embeddings` has no `qwen2` module, so it still dies
with the exact error from the issue — `ValueError: Model type qwen2 not
supported.` (that's what pmarreck hit in #686 after reclassifying via the admin
UI). The new `omlx/models/qwen2_embedding.py` is what actually makes these serve.

Scoped deliberately: just the adapter + classification + the arch→module dispatch
entry. It doesn't touch the loader or `reranker.py`, since your `ca83267a` already
moved those to `mx.load` + `train(False)`.

To make it cheap to verify:
- `pytest tests/test_embedding.py tests/test_model_discovery.py` → 219 pass, no
  downloads. The new `TestNativeQwen2Embedding` drives the real `_load_native` +
  `embed()` path and is red without the adapter.
- Real models vs a stock `transformers` reference: `jinaai/jina-code-embeddings-1.5b`
  (causal) and `Alibaba-NLP/gte-Qwen2-1.5B-instruct` (bidirectional) both match to
  mean cosine ~0.9999 with correct relevant-vs-irrelevant ranking. The `is_causal`
  config flag is load-bearing here — gte-Qwen2 runs bidirectionally
  (`is_causal: false`); a causal-only adapter mis-serves it (cos ~0.6).

Happy to adjust if you'd rather it land differently.

jundot:
Sorry it took me a while to get back to this after your update. Thanks for rebasing it and keeping it current with the latest commits.

I checked the classification path, the native Qwen2 adapter, and the target model configs. The adapter matches the existing MLX-LM Qwen2 decoder on the causal path and covers the last-token pooling / `is_causal: false` cases needed here.

The focused tests pass locally, and this looks good to me for #686. I'm going to merge it.

PR #1254, opened 2026-05-14
Eval Reliability, Benchmark History, Model Manager QoL

aljen:
# Eval Reliability, Benchmark History, Model Manager QoL

## Summary

This PR improves oMLX benchmark reliability and observability, adds persistent
benchmark/model metadata, and refines the model manager UI.

The largest behavior change is in accuracy benchmark execution: evals no longer
run as fixed chunks that wait for the slowest request before starting the next
chunk. They now keep up to `batch_size` requests in flight and refill completed
slots immediately. This removes a large amount of idle time for thinking/code
benchmarks with long-tail generations.

## Why

- Accuracy/code evals were hard to trust because HumanEval/MBPP/LiveCodeBench
  scoring was too brittle around common model output formats, indentation,
  markdown fences, tiny numeric drift, and diagnostic reporting.
- Accuracy benchmark results were process-local and easy to lose after restart.
- Performance benchmark results were not retained in the UI/model list.
- The model manager row mixed names, source tags, size, speed, and accuracy into
  one visual blob.
- Rebased upstream changed model settings/profile classification and admin route
  typing, which needed test alignment on this branch.

### Accuracy benchmark execution

- Replaces fixed chunk barriers with a refill-on-completion queue in
  `BaseBenchmark`.
- Keeps up to `batch_size` generation requests in flight.
- Starts the next item as soon as a generation slot completes instead of waiting
  for the slowest request in the current chunk.
- Moves scoring of generated code outputs behind a bounded async queue and runs
  scoring in a worker thread so subprocess-based code evaluation does not block
  generation refill.
- Records per-question generation time instead of assigning each question the
  chunk average.
- Preserves the existing first-batch thinking auto-detection behavior by running
  a probe batch before entering the queue.
- Applies the shared refill path to generic evals, HumanEval, MBPP, and
  LiveCodeBench.

Measured locally on MacBook Pro 16-inch, M3 Max, 128GB,
`Qwen3.6-35B-A3B-oQ4-mtp`:

```text
LiveCodeBench 300, non-thinking:
- old 8x chunk barrier: 150/300 in 4774.5s
- refill 8x best:       151/300 in 3454.4s
- refill 4x:            147/300 in 2788.0s

LiveCodeBench 300, thinking:
- old 8x chunk barrier: 158/300 in 33914.9s
- refill 8x:            158/300 in 23904.1s

Thinking run saved ~10010.8s / 166.8m / 2.78h with the same score.
```

### Accuracy benchmark correctness and diagnostics

[...]

PR #1722, opened 2026-06-07
fix(eval): deterministic accuracy-benchmark sampling with model_settings opt-in

### LINKED ISSUE #606: Issue : Gemma 4 26b a4b, bf16 - not matching official Livecode bechmark in omlx 0.34

JimStenstrom:
Fixes #606.

Supersedes this PR's original approach (per-model temperature winning by default), which would make saved benchmark scores non-deterministic. This adopts the design @jundot proposed on #1254:

> "Accuracy sampling: the default moved to model_settings, which makes scores non-deterministic when a model has a non-zero temperature. Would you be open to defaulting `sampling_profile` to `deterministic` (with `model_settings` as opt-in) so saved scores stay reproducible?"

## Summary
- Add `sampling_profile` to the accuracy-benchmark request: `deterministic` (default, reproducible greedy) or `model_settings` (opt-in, honors the model's configured sampling).
- #606's root cause — the benchmark ignored a model's temperature (Gemma 4 26B scoring ~47% vs ~77% on LiveCodeBench) — is fixed via the opt-in, while default runs stay reproducible.

## Why
`eval/base.py` force-set `temperature=0.0` (and neutral penalties) and `accuracy_benchmark.py` never read `ms.temperature`, so temperature-sensitive benchmarks scored a model far from its real configuration. But simply honoring temperature *by default* makes saved scores depend on per-model settings — exactly the non-determinism @jundot flagged. The `sampling_profile` toggle resolves both: reproducible by default, real-world sampling on request.

## Changes
- `omlx/admin/accuracy_benchmark.py`: `sampling_profile: Literal["deterministic", "model_settings"] = "deterministic"` on `AccuracyBenchmarkRequest`; the per-model sampling read is gated on `model_settings` and now includes `temperature`.
- `omlx/eval/base.py`: `_eval_single` uses `setdefault` for `temperature`/`presence_penalty`/`repetition_penalty` (were force-set) so the opt-in profile can supply them; `max_tokens` stays benchmark-forced (a model's small limit must not truncate answers).
- Dashboard: a "Sampling" toggle cloned from the existing Thinking Mode control (`_bench_accuracy.html` + `dashboard.js`); i18n keys added to `en.json` and propagated to all locales via `scripts/normalize_i18n.py`. No new tailwind classes (reuses the toggle's classes) → no CSS rebuild.

## Behavior / compatibility
- Default `deterministic` is score-identical to today (main already force-set temp=0; top_p/top_k are vestigial at greedy and penalties were already neutralized in `_eval_single`). API callers that omit the field get `deterministic`.
- `ModelSettings` is untouched — no new persisted field, no migration.

## Sibling sweep
- Both sampling paths covered: the request gate (`accuracy_benchmark.py`) and the eval defaults (`eval/base.py`) — `setdefault` applied to all three forced sampling params for consistency, not just temperature.
- `_eval_single` is shared by every benchmark subclass (mmlu/gsm8k/humaneval/mbpp/livecodebench); the setdefault change applies uniformly to all of them.
- The performance benchmark (`omlx/admin/benchmark.py`) reads per-model settings only for experimental-feature metadata, not sampling — intentionally out of scope (it measures throughput, not accuracy, so determinism doesn't apply).
- UI reachable + localized (toggle + 8 locales). `enable_thinking` stays orthogonal (its own field, merged into `chat_template_kwargs` regardless of profile).

## Tests
```
python -m pytest tests/test_accuracy_benchmark.py tests/test_eval.py -q   # 93 passed
```
- Request: `sampling_profile` default is `deterministic`; `model_settings` accepted; invalid value rejected.
- Gating: `deterministic` forwards no sampling (`sampling_kwargs == {}`); `model_settings` forwards the model's `temperature`/`top_p`.
- `_eval_single`: greedy defaults when empty, caller sampling overrides them, `max_tokens` always benchmark-controlled.
- Full smoke (`pytest -m "not slow and not integration"`): **5938 passed, 23 skipped**.
- Verified both gates fail when reverted (deterministic test sees leaked sampling; override test sees forced 0.0).


JimStenstrom:
Before this goes further — a heads-up on overlap. This and @aljen's #1254 both implement the deterministic `sampling_profile` design you proposed on #1254 (default `deterministic`, `model_settings` opt-in), for the #606 accuracy-sampling path. They change the same files (`accuracy_benchmark.py`, `eval/base.py`, the bench-accuracy toggle), so they're largely duplicative — I opened this from #606 around the same time aljen added it to #1254, without realizing they'd converge.

Since #1254 already carries this (and more), I'm happy to close this in favor of it if that's the vehicle you want. I've rebased this onto current main and it's green (incl. the new pt-BR locale) only so there's a ready, focused fallback for just #606 if #1254's broader scope needs more review time. Whichever you prefer — say the word and I'll close it or keep it current.

One shared heads-up for whichever lands: under `deterministic`, both this and #1254 drop the model's custom `chat_template_kwargs` (only `enable_thinking` is kept), so a model with a non-thinking template kwarg won't be prompt-identical to the pre-change default — worth a conscious call on whether that's the intended reproducibility semantics.

jundot:
Thanks for the heads-up on the overlap, and for keeping this rebased and green. Let's go with this one as the focused fix for #606, since #1254 carries a much broader scope and I haven't been able to give it the review time it needs yet. I'll merge this.

On the chat_template_kwargs point you raised, I agree that's worth fixing. Template kwargs are prompt construction rather than sampling, so the deterministic profile should still forward them to stay prompt-identical with the old default. I'll handle that in a follow-up commit after merging.

PR #1728, opened 2026-06-08
Fix login page to honor system dark mode setting

monroewilliams:
## Problem

The login page on the admin endpoint didn't honor the system dark mode setting, unlike the chat and dashboard pages.

## Root Cause

The theme initialization script in `base.html` was defaulting to `'light'` and never checking `prefers-color-scheme` when the stored theme value was `'auto'`.

## Fix

Updated the inline theme script in `base.html` to:
- Default to `'auto'` instead of `'light'` when no theme is stored in localStorage
- Check `window.matchMedia('(prefers-color-scheme: dark)')` when the theme is `'auto'` to detect the system dark mode preference

This matches the behavior already used by `chat.html`.

### DIFF
--- a/omlx/admin/templates/base.html
+++ b/omlx/admin/templates/base.html
@@ -32,7 +32,11 @@
     <script>
         // Apply theme immediately to prevent flash of wrong theme
         (function() {
-            var theme = localStorage.getItem('omlx-chat-theme') || 'light';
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for the fix. The base template now matches the chat/dashboard auto-theme behavior for initial render, and it keeps explicit light/dark choices unchanged. This looks good to me, and I'm going to merge it.

PR #1730, opened 2026-06-08
fix(i18n): align non-English memory guard strings with JS placeholders

fqx:
## Problem

The description_template and wired_limit_warning strings in all non-English locale files use placeholder names that do not match what the JavaScript actually interpolates.

**description_template** — JS replaces {tier}, {active_pct}, {reclaim_method}; non-English locales had {ceiling_gb} instead, so it appeared verbatim.

**wired_limit_warning** — JS replaces {kernel} with a value that already includes the unit (e.g. 37.4 GB); non-English locales had {kernel} GB causing double-GB output, and {requested} which is never replaced.

en.json was already correct. This fixes the 7 affected locales: zh, zh-TW, ja, ko, fr, ru, es.

## Fix

Updated all 7 locale files to use the correct placeholders and added proper translations (previously these strings were left in English in the non-English files).

## Testing

1. Switch dashboard language to Chinese / Japanese / Korean / French / Russian / Spanish
2. Open Settings > Resources > Memory Guard
3. Tier description should show the actual percentage and reclaim method instead of {ceiling_gb}
4. Kernel warning should show the GB value once, not twice, and no {requested} literal

Generated with [Claude Code](https://claude.com/claude-code)


jundot:
Thanks for fixing this. The updated placeholders match the dashboard interpolation path, and I verified the non-English locale files no longer leave stale `{ceiling_gb}` or `{requested}` tokens behind. This looks good to me, and I'm going to merge it.

PR #1732, opened 2026-06-08
Add OMLX_SYNC_STORE_CACHE to make prefix-cache commit synchronous (fixes back-to-back cache-miss race)

### LINKED ISSUE #1731: Prefix cache misses on back-to-back / rapid requests (async store_cache commit race)
**Describe the bug**

oMLX registers a finished request's prefix-cache blocks (including the
`cached_block_hash_to_block` index that `fetch_cache` matches against) on a background
single-worker executor (`Scheduler._store_cache_executor` → `_async_store_cache_worker`
→ `block_aware_cache.store_cache(...)`). A subsequent request that arrives **before that
async commit lands** does not see the just-computed prefix and re-prefills from scratch —
even when the prompt is byte-identical to one served seconds earlier.

The commit window is small when the server is idle (<3 s on an M3 Ultra with a 31B bf16),
but a request that re-sends a shared prefix within that window — rapid or
overlapping requests — outruns the commit and re-prefills instead of reusing it. Impact scales
with prefill cost: a missed warm request on a 31B bf16 costs ~60 s vs ~5 s on a hit.

**To Reproduce**

Send a long prompt, then re-send the **identical** prompt after a controllable gap.
`max_tokens=4` so wall time ≈ prefill time. Self-contained repro (OpenAI-compatible API +
`/api/status` cache counters; no app needed):

```python
import json, os, time, copy, urllib.request
BASE = os.environ.get("OMLX_BASE", "http://127.0.0.1:8000")
KEY  = os.environ.get("OMLX_KEY", "YOUR_API_KEY")
MODEL = os.environ.get("OMLX_MODEL", "gemma-4-31b-it-bf16")  # any large bf16 (big KV)
H = {"Authorization": f"Bearer {KEY}", "Content-Type": "application/json"}

def counters():
    r = urllib.request.urlopen(urllib.request.Request(f"{BASE}/api/status", headers=H), timeout=15)
    d = json.loads(r.read()); return int(d["total_prompt_tokens"]), int(d["total_cached_tokens"])

def send(msgs):
    body = json.dumps({"model": MODEL, "messages": msgs, "max_tokens": 4, "stream": True}).encode()
    r = urllib.request.urlopen(urllib.request.Request(f"{BASE}/v1/chat/completions", data=body, headers=H), timeout=1800)
    t0 = time.perf_counter()
    for raw in r:
        if raw.strip().endswith(b"[DONE]"): break
    return time.perf_counter() - t0

# ~6k-token system prompt so the KV is large enough to show the window
sysmsg = "You are a careful assistant.\n" + "\n".join(f"Rule {i}: keep answers short." for i in range(380))
for gap in (0, 5):
    msgs = [{"role": "system", "content": f"[run {gap}] " + sysmsg},
            {"role": "user", "content": "Reply OK."}]
    cold = send(msgs); time.sleep(gap); warm = send(msgs)
    print(f"gap={gap}s  cold={cold:.1f}s  warm={warm:.1f}s  -> {'HIT' if warm < 20 else 'MISS'}")
```

Output (oMLX 0.4.2rc1, gemma-4-31b-it-bf16):
```
gap=0s  cold=71.9s  warm=58.8s  -> MISS   (re-prefilled; cached_tokens ~ 0)
gap=5s  cold=56.9s  warm= 5.0s  -> HIT    (cached_tokens ~ prompt)
```

**Expected behavior**

The identical re-sent prompt should hit the prefix cache (warm ≈ reconstruct time, a few
seconds) regardless of how soon it follows the first request. It does at gap=5s but not at
gap=0s — the gap=0 case re-prefills instead of reusing the identical prefix.

**Environment**
- oMLX version: 0.4.2rc1
- OS / chip: macOS, Apple Silicon (M3 Ultra, 512 GB)
- Model: gemma-4-31b-it-bf16 (bf16; large KV makes the window easy to observe)
- Cache config: `hot_cache_only=true`, `hot_cache_max_size=102GB`

**Additional context**

Root cause is the async store_cache commit: the index registration is deferred to the
single-worker executor, so it isn't visible to the next request's `fetch_cache` until the
worker drains.

I have a small, tested fix and will open a PR (an env-gated flag routing store_cache through
oMLX's existing synchronous fallback path; SSD write stays async). Happy to adjust the gating
to a settings field instead of an env var if you'd prefer.

Codedependent77:
Fixes #1731.

### What

Adds an opt-in env flag `OMLX_SYNC_STORE_CACHE=1` that disables the single-worker
`_store_cache_executor` so `store_cache` (including the `cached_block_hash_to_block` index
registration) runs synchronously on the inference thread via the **existing** synchronous
fallback path (`scheduler.py`, the `else: self._async_store_cache_worker(...)` branch). With
it set, the prefix cache is immediately consistent, so a request arriving right after a prior
one reuses the just-computed prefix instead of re-prefilling.

### Why

Currently the post-request commit (which registers the prefix-cache index) is deferred to a
background worker. A request that arrives before the commit lands misses and re-prefills. The
window is <3 s when idle but is hit by rapid or overlapping requests that re-send a shared
prefix within that window, which then re-prefill instead of reusing it. See #1731 for a
self-contained repro.

### Proof (same model/config, only the flag differs; flag confirmed in server log)

```
baseline (async):                gap=0  warm=58.8s  MISS
OMLX_SYNC_STORE_CACHE=1 (sync):  gap=0  warm= 3.8s  HIT
                                 gap=5  warm= 4.5s  HIT
```

### Overhead

Negligible, and **not** a throughput regression:
- The SSD/disk write stays async — `save_block` enqueues to the background writer thread (and
  under `hot_cache_only=true` no disk write happens at all). No disk I/O moves to the inference thread.
- Phase timers show `store_cache_worker_sync = 0.0–0.2 ms`; the only sizeable store line
  (`store_cache_main_dispatch ≈ 250 ms`) is the `mx.eval` of the KV, which already runs on the
  inference thread in the async path (the pre-eval), so it is not added here.
- A/B cold-request times are within noise (57–72 s baseline vs 59–65 s sync).

### Tests

Adds `tests/test_sync_store_cache.py` (3 cases, fast — no model load, uses the existing
`mock_model`/`mock_tokenizer` conftest fixtures and the same `Scheduler(...)` construction as
`tests/test_scheduler.py`):
- default (unset) → `_store_cache_executor` is created (async);
- `OMLX_SYNC_STORE_CACHE=1` → `_store_cache_executor is None` and the cache is still enabled;
- any non-`"1"` value → stays async.

`pytest tests/test_sync_store_cache.py` → 3 passed in ~1.2s. (The flag-on case asserts the
executor is `None`, so it fails on unpatched `main` — it genuinely guards the fix.)

### Notes for review

- **Default unchanged:** flag unset → identical single-worker `ThreadPoolExecutor` as before.
- **Metal-safe:** running `_async_store_cache_worker` inline is safer than the threaded path —
  it runs on the thread that owns `self._stream`, so the worker's "KV must be concrete before a
  *different* thread slices/views it" invariant can't be violated. The pre-`mx.eval` still runs;
  `_mx_buffer_access_lock` is an `RLock`.
- **Intentional:** in sync mode the shutdown block (guarded by `_store_cache_executor is not
  None`) is skipped, so `_store_cache_gate` stays non-None. Harmless — `_inflight_store_futures`
  and `_pending_async_removes` are never populated in sync mode (the gate's note_submitted /
  note_done and the inflight futures live only inside the executor branch and `_drain_*`), so
  there's nothing to drain or reset.
- I considered a lighter "register the index sync, extract data async" variant — it's
  **incorrect**: in `store_cache`, `register_block_hash` and the tensor extraction/store are
  coupled, and reconstruct drops a matched block whose data fails to load (false hit →
  re-prefill → un-register → thrash). Hence the full-but-cheap synchronous path.

Happy to switch the gate from an env var to a settings field if that fits the project better.

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -1226,10 +1226,28 @@ class Scheduler:
                 # Async store_cache executor: single worker so submissions are
                 # serialized (matches the original synchronous order) and we
                 # never have two stores racing on the same paged_ssd index.
-                self._store_cache_executor = concurrent.futures.ThreadPoolExecutor(
-                    max_workers=1,
-                    thread_name_prefix="omlx-store-cache",
-                )
+                #
+                # OMLX_SYNC_STORE_CACHE=1 disables this executor so store_cache
+                # (including the cached_block_hash_to_block index registration)
+                # runs synchronously on the inference thread via the existing
+                # fallback path below. The prefix cache then becomes immediately
+                # consistent: a request arriving right after a prior one
+                # (back-to-back, an agentic loop, or a concurrent stream) reuses
+                # the just-computed prefix instead of re-prefilling because the
+                # async commit had not landed yet. The SSD write stays async via
+                # the background writer thread; only the cheap in-RAM register +
+                # hot-cache populate moves onto the inference thread.
+                if os.environ.get("OMLX_SYNC_STORE_CACHE") == "1":
+                    self._store_cache_executor = None
+                    logger.info(
+                        "OMLX_SYNC_STORE_CACHE=1 -> synchronous store_cache "
+                        "(immediate prefix-cache consistency)"
+                    )
+                else:
+                    self._store_cache_executor = concurrent.futures.ThreadPoolExecutor(
+                        max_workers=1,
+                        thread_name_prefix="omlx-store-cache",
+                    )
                 # Gate caps the post-completion store-cache pipeline so a burst
                 # of finishes cannot pile up unbounded KV caches in memory while
                 # the single writer drains. Cap starts at max_concurrent_requests
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -1227,7 +1227,8 @@ class Scheduler:
                 # serialized (matches the original synchronous order) and we
                 # never have two stores racing on the same paged_ssd index.
                 #
-                # OMLX_SYNC_STORE_CACHE=1 disables this executor so store_cache
+                # OMLX_SYNC_STORE_CACHE=1 (or true/yes/on) disables this executor
+                # so store_cache
                 # (including the cached_block_hash_to_block index registration)
                 # runs synchronously on the inference thread via the existing
                 # fallback path below. The prefix cache then becomes immediately
@@ -1237,10 +1238,11 @@ class Scheduler:
                 # async commit had not landed yet. The SSD write stays async via
                 # the background writer thread; only the cheap in-RAM register +
                 # hot-cache populate moves onto the inference thread.
-                if os.environ.get("OMLX_SYNC_STORE_CACHE") == "1":
+                _sync = os.environ.get("OMLX_SYNC_STORE_CACHE", "").strip().lower()
+                if _sync in {"1", "true", "yes", "on"}:
                     self._store_cache_executor = None
                     logger.info(
-                        "OMLX_SYNC_STORE_CACHE=1 -> synchronous store_cache "
+                        "OMLX_SYNC_STORE_CACHE enabled -> synchronous store_cache "
                         "(immediate prefix-cache consistency)"
                     )
                 else:
--- a/tests/test_sync_store_cache.py
+++ b/tests/test_sync_store_cache.py
@@ -20,15 +20,19 @@ def make_scheduler(mock_model, mock_tokenizer, tmp_path):
     (the default branch spins up a real ThreadPoolExecutor + SSD writer thread).
 
     Returned as a factory so the test can set OMLX_SYNC_STORE_CACHE *before*
-    construction.
+    construction. Each created scheduler gets its own cache subdirectory so
+    multiple instances cannot share on-disk state.
     """
     created = []
 
     def _make():
         if not scheduler_module.HAS_TIERED_CACHE:
             pytest.skip("tiered cache modules are unavailable")
-        config = SchedulerConfig(paged_ssd_cache_dir=str(tmp_path / "cache"))
-        scheduler = Scheduler(model=mock_model, tokenizer=mock_tokenizer, config=config)
+        cache_dir = tmp_path / f"cache-{len(created)}"
+        config = SchedulerConfig(paged_ssd_cache_dir=str(cache_dir))
+        scheduler = Scheduler(
+            model=mock_model, tokenizer=mock_tokenizer, config=config
+        )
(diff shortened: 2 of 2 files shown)


Codedependent77:
Thanks for the review! Addressed in 59b922f:

- **Env parsing**: `OMLX_SYNC_STORE_CACHE` now parses like the other omlx env booleans (`1`/`true`/`yes`/`on`, case- and whitespace-insensitive), matching `settings.py`, so operators won't silently get async behavior from e.g. `true`. Added parametrized truthy/falsy coverage.
- **Test isolation**: the scheduler factory fixture now gives each instance its own cache subdirectory.

On the behavioral-test suggestion: agreed it's stronger. The end-to-end behavioral validation is in the PR description — the `gap=0` MISS→HIT result (58.8s → 3.8s) plus the self-contained repro in #1731. A true in-process "second identical request hits" test has to drive a request through the scheduler step loop with a real model loaded, so it'd be a `@pytest.mark.slow`/integration test rather than a unit test — happy to add one under that marker if you'd like. The unit tests here guard the regression surface (flag → executor disabled → synchronous fallback), which is what would silently break.

Codedependent77:
Withdrawing this.

After submitting, I tested whether the fix helps realistic workloads, and by my own measurements it doesn't measurably:

- **Single-client sequential/agentic loop:** no effect. Consecutive requests reuse the long-committed base prefix, and the inter-request gaps are already past the commit window, so they hit with or without the flag.
- **Concurrent duplicate requests** (the realistic version of the repro — retries, double-submits, load): no effect. With continuous batching, N identical requests fired together are admitted into the same prefill batch and all re-prefill regardless of the flag. Measured with N=4: `cached_tokens=0`, `prompt_tokens = 4 × P`, 4/4 cold — identical with sync on and off.

The flag does fix the literal repro (a strictly serial resend inside the ~250 ms commit window), but that condition turns out to be too narrow to justify your review time, so I'm closing it off your queue.

For what it's worth, the higher-value direction looked like sharing prefix KV across requests batched together (so N identical concurrent prompts prefill once instead of N times) — but that's a separate, larger change. Thanks for the project.

Codedependent77:
Withdrawing (see #1731): the back-to-back cache-miss we were chasing turned out to be largely a measurement artifact rather than a real-world race, so OMLX_SYNC_STORE_CACHE isn't needed here. Thanks for the review.

PR #1742, opened 2026-06-09
Detect NeMo ASR models as STT

scaryrawr:
## Summary
- detect NeMo ASR model configs as speech-to-text engines
- add coverage for mlx-community/parakeet-tdt-0.6b-v3-style configs

## Tests
- uv run --python 3.13 pytest tests/test_audio_discovery.py -v

### DIFF
--- a/omlx/model_discovery.py
+++ b/omlx/model_discovery.py
@@ -387,6 +387,47 @@ def _has_sentence_transformers_embedding_pipeline(model_path: Path) -> bool:
     )
 
 
+def _looks_like_nemo_asr_config(config: dict) -> bool:
+    """Return True for NeMo ASR exports that omit HF ``model_type``.
+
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the fix. I checked the Parakeet NeMo config shape against mlx-community/parakeet-tdt-0.6b-v3 and nearby CTC/RNNT/TDT exports, and this matches the top-level ASR fields they ship without a Hugging Face model_type.

The heuristic is narrow enough to avoid nested multimodal audio config misclassification, and the focused audio discovery tests pass locally. Since this PR is still marked as draft, I'll wait for you to mark it ready before merging.

jundot:
Thanks for taking this out of draft. I rechecked the live Parakeet NeMo config shape and the discovery-to-STTEngine path; the heuristic stays narrow and the targeted tests pass. This looks good to me, and I'm going to merge it.

PR #1744, opened 2026-06-09
fix: force gemma4_unified detection as VLM

FaisalFehad:
## Summary

Forces `gemma4_unified` detection as VLM in model discovery.

### Problem
`gemma4_unified` models were detected as LLM when their `config.json` lacked `vision_config` (e.g., some quantized variants). Since unified models always include vision+audio capabilities by definition, they should never fall back to LLM.

### Changes

**`omlx/model_discovery.py`** — Force `gemma4_unified` → VLM detection regardless of `vision_config` presence.

**`tests/test_model_discovery.py`** — Updated `test_detect_text_only_gemma4_unified_as_llm` → `test_detect_gemma4_unified_without_vision_config_as_vlm` since `gemma4_unified` is always VLM.

### Notes
- The feature extractor mapping in `vlm.py` is **not changed** — `Gemma4UnifiedAudioFeatureExtractor` correctly resolves to `mlx_vlm.models.gemma4_unified.processing_gemma4_unified.Gemma4UnifiedAudioFeatureExtractor` in the pinned mlx-vlm (commit `54c9a11`). That class differs from `Gemma4AudioFeatureExtractor` (which produces mel spectrograms) — the unified extractor outputs raw waveform chunks with `audio_samples_per_token=640`.
- `gemma4_unified` is an architecture type (not a size label). Currently only the 12B uses it; future models with the same unified audio+vision architecture would also use this type.
- All 125 model discovery tests pass.

### DIFF
--- a/omlx/model_discovery.py
+++ b/omlx/model_discovery.py
@@ -532,8 +532,15 @@ def detect_model_type(model_path: Path) -> ModelType:
 
     # Check for VLM: model_type field (only if vision capabilities are present)
     # Some model families (e.g., qwen3_5_moe) have both VLM and text-only variants.
-    # Text-only quants won't carry a vision sub-config.
+    # Text-only quants won't carry a vision sub-config. gemma4_unified is an
+    # exception: it is always VLM (unified audio+vision architecture) regardless
+    # of vision_config presence in config.json.
     if normalized_type in VLM_MODEL_TYPES:
+        if normalized_type == "gemma4_unified":
+            logger.info(
+                "gemma4_unified detected as VLM (unified models always include vision)"
+            )
+            return "vlm"
         if _has_vision_subconfig(config):
             return "vlm"
         logger.info(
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the PR. The model discovery change makes sense to me, but I do not think the feature extractor mapping should change.

`Gemma4UnifiedAudioFeatureExtractor` does exist in the Gemma 4 unified processor path, and current mlx-vlm has it here:
https://github.com/Blaizzy/mlx-vlm/blob/c9a6743e5f059be36362d8e4a78171bbf91f18bc/mlx_vlm/models/gemma4_unified/processing_gemma4_unified.py#L251

It is not equivalent to `Gemma4AudioFeatureExtractor`; the unified extractor uses different defaults and output shape for unified audio inputs. Remapping the unified type to the Gemma4 extractor can break `gemma4_unified` audio requests.

I can merge the narrower discovery fix once that mapping change is removed or covered by a runtime-compatible test.

FaisalFehad:
Thank you @jundot I have updated the PR.

jundot:
Thanks for updating this. The narrowed discovery-only change looks right: gemma4_unified should stay on the mlx-vlm path even when quantized configs omit vision_config, and the unified audio feature extractor mapping is preserved.

CI is green, so this looks good to me, and I am going to merge it.

Issue #934, opened 2026-04-23
qwen3.6-35b-a3b and qwen3.6-27b infinite loop


undici77:
Hi, I tried many versions of these models (oQ/4bit/8bit) but issue is always that same, after some interaction model enter in infinite loop!
Same models (I shared same files) with LM STUDIO work fine!

zviratko:
What are you using it with? I found that having no system prompt (default in the built-in chat) makes the models get weird. Just putting something like "Be useful and concise" is enough.

I am trying qwen3.6-35b-a3b with hermes agent and it gets loopy on tool calls - might be related to the preserve_thinking issue (there's also one for hermes agent and I saw something for openclaw as well)... or maybe they just broke it, looks like local models are second class citizens to huge cloud ones.

TipKnuckle:
I have had the same experience with looping with the new omlx 3.7 release. I switched back to omlx 3.6 and have not had the looping issue since. But I have not investigated thoroughly, so I can't say for certain that's the cause.

fry69:
> I switched back to omlx 3.6 and have not had the looping issue since.

Same here, I also switched back to 0.3.6 due to this problem. 0.3.6 is rock-solid for me, without endless loops.

See also issue #911

jundot:
@undici77 @zviratko @TipKnuckle @fry69

Just pushed v0.3.8.dev1 for testing. Previously I was using mlx-lm as the decoding engine for VLMs because mlx-vlm didn't support continuous batching. Now that mlx-vlm has native continuous batching, I switched the VLM decoding engine over to mlx-vlm, which improved behavior across all Qwen models.

Could you give it a try and see if the looping issue is still happening? Any feedback is helpful since this is a dev build.

Release: https://github.com/jundot/omlx/releases/tag/v0.3.8.dev1

fry69:
> Could you give it a try and see if the looping issue is still happening?

Thanks for the quick reaction/release. Looks good so far 👍 

I am running `0.3.8.dev1` now and have generated 3 moderately long/complex prompt/request pairs with `qwen3.6-35b-a3b`, which in the past ended up in endless loops, but work with this version. I have not tried `qwen3.6-27b`.

If those loops appear again, I'll report.

molenick:
I downloaded 0.3.8-dev1 and still got looping issues with qwen3.6-27B. I also tried completing the work with qwen3-coder-next and ran into similar looping issues with that model as well.

[...]

PR #1747, opened 2026-06-09
fix: normalise logits_processors None to [] to avoid mlx-lm crash (#934)

monroewilliams:
## Reproduction

- Install the [forgecode](https://forgecode.dev) harness
- Set up the `OpenAICompatible` provider to point at a local oMLX endpoint
- I don't know if this is model-dependent, but I was using https://huggingface.co/deepsweet/Qwen3.6-35B-A3B-MLX-VL-oQ8-FP16
- Send any prompt (I just started with "hi")
- Error output: 
```
 ERROR: POST http://127.0.0.1:8000/v1/chat/completions                                                                        
                                                                                                                                          
 Caused by:                                                                                                                               
     0: Failed to create completion message: {"error": {"message": "Cache corruption not recoverable after retries: 'NoneType' object is  
 not iterable", "type": "server_error"}}                                                                                                  
     1: {"message":"Cache corruption not recoverable after retries: 'NoneType' object is not iterable","type":"server_error"}             
```

- The "Cache corruption" error message seems to be spurious, I cleared the in-memory and ssd caches and it didn't resolve.

## Diagnosis

When `logits_processors` is `None` (no repetition/presence/frequency penalties configured), mlx-lm's `GenerationBatch._step` crashes with:

    TypeError: 'NoneType' object is not iterable

This is caught by omlx's cache corruption recovery, but since the root cause persists, requests fail after 3 retries with:

    Cache corruption not recoverable after retries: 'NoneType' object is not iterable

## Fix

Three layers of defense to normalise `None` → `[]`:

1. **`_create_batch_generator`** — pass `[]` instead of `None` to `BatchGenerator` constructor
2. **`_patched_generation_batch_step`** — normalise `None` → `[]` before calling mlx-lm's `_step` (which crashes at `any(None)`)
3. **`_patched_ppb_split`** — normalise `None` → `[]` when splitting batches

Consistent with the existing fix at line 6528 where `per_row_lps = list(logits_processors) if logits_processors else []` was already applied.

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -385,6 +385,13 @@ def _patched_generation_batch_step(self):
         deltas = [model._uid_rope_deltas.get(uid, 0.0) for uid in self.uids]
         model.set_batch_rope_deltas(mx.array(deltas))
 
+    # Defensive: mlx-lm's GenerationBatch._step does `any(self.logits_processors)`
+    # and `for p in self.logits_processors[e]`, both of which crash when
+    # logits_processors is None.  Normalise to [] so the original code path
+    # works without modification.  See #934.
+    if self.logits_processors is None:
+        self.logits_processors = []
+
     result = _original_generation_batch_step(self)
 
     # self._next_tokens contains the just-sampled tokens (async eval pending).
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for the defensive cleanup. The scheduler's normal insert path already passes per-row logits processor lists, and this additional normalization keeps the remaining boundaries from leaking None into mlx-lm. The focused scheduler/logits tests and CI look good, so I'm going to merge this.

Issue #1440, opened 2026-05-27
Question: Are these security-related behaviors intentional?


anvarknian:
Hi **maintainers**, thanks for the great project.

I did a security review of the current codebase and found a few behaviors that might be intentional based on a localhost-first threat model, but I wanted to confirm before treating them as vulnerabilities.

Could you confirm whether the following are expected/intentional?

1. Initial setup is open when no API key is configured  
- `omlx/server.py` (`verify_api_key`: returns `True` when no key)  
- `omlx/admin/routes.py` (`POST /admin/api/setup-api-key`)

2. API keys exposed to frontend/admin runtime  
- `GET /admin/api/global-settings` returns `auth.api_key` and `auth.sub_keys`  
- `omlx/admin/routes.py` injects API key into `chat.html` template  
- `chat.html` stores key in `localStorage`

3. Admin auto-login uses API key in query string  
- `GET /admin/auto-login?key=...`  
- `packaging/omlx_app/app.py` opens browser with key in URL

4. Rejected API keys are logged  
- `omlx/server.py`: `logger.warning("Rejected API key: %r", api_key_value)`

5. Image loader accepts arbitrary `http(s)` and local file paths  
- `omlx/utils/image.py` (`urlopen` + `Image.open(path)`)

6. Session cookie is set without `Secure` flag  
- login/setup/auto-login cookie `set_cookie(...)` in `omlx/admin/routes.py`

If these are intentional, could you share the intended threat model (strictly local-only, trusted browser, etc.) so users can understand risk boundaries?  
If not intentional, I can submit a hardening PR.

BrianSmithPhotos:
Following up on point 5 with a concrete exploit path, since the current framing is a yes/no question — this one has two independently exploitable primitives, not just a design question.

## Summary

`load_image()` in `omlx/utils/image.py` has no restriction on `http(s)://` targets and falls back to treating any other string as a local filesystem path. It's reachable pre-admin, from any holder of a valid API key (including a restricted sub-key created for API-only use) via two request-facing endpoints — so this isn't gated behind the admin session at all.

## Location

`omlx/utils/image.py:47-55`

```python
elif url_or_base64.startswith(("http://", "https://")):
    import urllib.request
    with urllib.request.urlopen(url_or_base64, timeout=30) as response:
        img_bytes = response.read()
    img = Image.open(io.BytesIO(img_bytes))
else:
    # Try as local file path
    img = Image.open(url_or_base64)
```

Reachable via:
[...]

PR #1751, opened 2026-06-09
fix(server): log a fingerprint of rejected API keys instead of the raw key

richgoodson:
Addresses item 4 of #1440 (the rejected-key logging item).

## Problem

When a request fails API key auth, `server.py` logs the rejected key verbatim at WARNING:

```python
logger.warning("Rejected API key: %r", api_key_value)
```

That writes the client's secret into `server.log` on every failed auth, at an always-on level. A user who mistypes a key, or whose key carries the non-ASCII problem fixed in #1719, leaks their credential straight into the logs.

## Fix

Log a short, non-reversible fingerprint instead of the raw value: the first 8 hex characters of the SHA-256 of the key. Operators can still correlate repeated rejections of the same key (a misconfigured client hammering the endpoint shows one stable `fp=`) without the secret ever landing in the log.

Added `fingerprint_key` to `omlx/admin/auth.py`, next to `compare_keys`, using the same `encode("utf-8", "surrogatepass")`. That keeps it tolerant of any `str` the auth path accepts, including the non-ASCII and lone-surrogate keys that #1719 already handles.

## Testing

- 6 new tests in `tests/test_api_auth.py`: fingerprint format (8 hex chars), determinism, that the fingerprint does not contain the raw secret, that distinct keys differ, that non-ASCII / empty / lone-surrogate inputs do not raise, and an end-to-end check that the auth dependency logs the fingerprint and not the raw rejected key (via `caplog`).
- Full suite: 5347 passed, 22 skipped, 66 deselected, 0 failed.
- Live test: started a server with `OMLX_API_KEY` set, sent a wrong `Bearer` key to `/v1/models`, and confirmed the server log shows `Rejected API key (fp=...)`, does not contain the raw key, and that the logged fingerprint matches `fingerprint_key` of the key I sent. A correct key returned 200.

## Scope note

This is a logging-only change; auth behavior is unchanged (a bad key still gets a 401). The other items in #1440 are design questions (open setup when no key is configured, the key being readable in the frontend, the session cookie's `Secure` flag), so I left those alone. Happy to take any of them as separate follow-ups if you confirm the intended behavior.

### DIFF
--- a/omlx/admin/auth.py
+++ b/omlx/admin/auth.py
@@ -5,6 +5,7 @@ This module provides session-based authentication using signed tokens
 and API key verification for admin panel access.
 """
 
+import hashlib
 import os
 import secrets
 from typing import Optional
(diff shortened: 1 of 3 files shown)


jundot:
Thanks, this addresses item 4 of #1440 cleanly. The change stays on the common rejected-key path, preserves auth behavior, and logs only a stable fingerprint instead of the submitted secret. CI is green and I do not see a regression risk here. Since this is still marked draft, I'll wait until you mark it ready before merging.

jundot:
Thanks for marking this ready. I rechecked the PR, and the scope still looks right: the rejected-key path now logs only a stable fingerprint, while auth behavior and both Bearer/x-api-key handling stay unchanged. The added tests cover the fingerprint helper and the dependency-level log behavior, including non-ASCII and lone-surrogate inputs, and CI is green. This looks good to me, and I am going to merge it.

PR #1761, opened 2026-06-09
fix: keep MTP decode eligibility per model instance

### LINKED ISSUE #1758: MTP decode path is disabled for already-loaded Qwen after loading a non-MTP model
## Bug description

In oMLX 0.4.2, loading any non-MTP model after an MTP-enabled model can silently disable the MTP decode path for the already-loaded MTP model.

The user-visible symptom is that `Qwen3.6-27B-oQ8-mtp` still shows `mtp_enabled: true` in model settings and remains loaded with its MTP head, but generation falls back to the non-MTP path. Throughput dropped from ~15 tok/s to ~10-12 tok/s in long-context workloads.

## Root cause found locally

`omlx.patches.mlx_lm_mtp._MTP_ACTIVE` is process-wide:

```python
_MTP_ACTIVE = False

def set_mtp_active(active: bool) -> None:
    global _MTP_ACTIVE
    _MTP_ACTIVE = bool(active)

def is_mtp_active() -> bool:
    return _MTP_ACTIVE
```

`utils/model_loading.py::maybe_apply_pre_load_patches()` resets it on every model load:

```python
from ..patches.mlx_lm_mtp import set_mtp_active
set_mtp_active(False)
```

Then `batch_generator._mtp_common_eligible()` checks the global flag at decode time:

```python
from . import is_mtp_active
if not is_mtp_active():
    return False
```

This means a later load of a non-MTP model, e.g. PaddleOCR-VL, flips the global flag to false. The already-loaded `Qwen3.6-27B-oQ8-mtp` instance still has its MTP module attached, but future requests no longer enter the MTP path.

## Reproduction

Environment:

- macOS 26.5.1
- Apple M5 Max, 128 GB
- oMLX 0.4.2 DMG app
- bundled engines reported by `/admin/api/stats`:
  - mlx-lm 0.31.3 (`bdb77dae5389d193dcc333bed891d33f6b668285`)
  - mlx-vlm 0.6.2 (`54c9a114ecf55f03facacd3684c5fa62403ef26d`)

Steps:

1. Load `Qwen3.6-27B-oQ8-mtp` with `mtp_enabled: true`.
2. Generate once. Logs show:

   ```text
   MTP path activated for uid=0 (model has mtp_forward, batch=1)
   ```

3. Load a non-MTP model, in my case `models--PaddlePaddle--PaddleOCR-VL-1.6`.
4. Unload the non-MTP model, keep Qwen loaded.
5. Generate from Qwen again.

With `log_level=debug`, the logs show:

```text
MTP path not active: mtp_active flag is off (model_settings.mtp_enabled was False at load time)
```

No `MTP path activated` line is emitted for the Qwen request.

6. Unload and reload Qwen. MTP works again:

```text
MTP path activated for uid=0 (model has mtp_forward, batch=1)
MTP[0] finish=stop tokens=59 cycles=37 accept=21/37 (56.8%) ...
```

## Expected behavior

Loading another model should not change the MTP eligibility of an already-loaded MTP-enabled model. MTP active state should be per loaded model/engine instance, not a process-wide decode-time flag.

## Suggested fix

Keep `set_mtp_active()` as a pre-load construction gate if needed for attaching the MTP module, but after model construction persist an instance-level marker, e.g. `_omlx_mtp_decode_enabled`, and make `batch_generator._mtp_common_eligible()` check that marker instead of the process-wide `is_mtp_active()`.

For example:

```python
def _model_mtp_decode_enabled(model):
    inner = getattr(model, "language_model", model)
    return bool(
        getattr(model, "_omlx_mtp_decode_enabled", False)
        or getattr(inner, "_omlx_mtp_decode_enabled", False)
    )
```

Then set `_omlx_mtp_decode_enabled = True/False` during model initialization based on the active flag at load time.

This preserves the current load-time behavior while preventing later non-MTP model loads from disabling already-loaded MTP models.

smkzw:
## Summary

- Capture MTP decode eligibility on the loaded model instance instead of reading the process-wide `_MTP_ACTIVE` flag at decode time.
- Keep `_MTP_ACTIVE` as the construction-time switch used by patched model initializers.
- Update Qwen, Qwen VLM/MoE VLM, and DeepSeek V4 MTP initializers to persist `_omlx_mtp_decode_enabled`.
- Add a regression test showing `set_mtp_active(False)` from a later non-MTP load does not disable an already-loaded MTP model.

Fixes #1758.

## Why

`maybe_apply_pre_load_patches()` resets `_MTP_ACTIVE` before every model load. In 0.4.2, `batch_generator._mtp_common_eligible()` checked that same process-wide flag during decode. Loading a later non-MTP model therefore disabled MTP for an already-loaded `Qwen3.6-*-mtp` model, even though the model still had its MTP head attached.

## Validation

Local targeted validation:

```text
compile_ok
eligibility_regression_check_ok
```

Command used:

```bash
PYTHONPATH=/path/to/omlx:/Applications/oMLX.app/Contents/Resources/Python/framework-mlx-base/lib/python3.11/site-packages \
/Applications/oMLX.app/Contents/Resources/Python/cpython-3.11/bin/python3.11 - <<'PY'
import py_compile
from types import SimpleNamespace
...
PY
```

I also applied the same hot patch to a local oMLX 0.4.2 app bundle and verified that loading/unloading `models--PaddlePaddle--PaddleOCR-VL-1.6` no longer disables MTP for an already-loaded `Qwen3.6-27B-oQ8-mtp`:

```text
MTP path activated for uid=1 (model has mtp_forward, batch=1)
MTP[1] finish=stop tokens=15 cycles=8 accept=6/8 (75.0%)
Chat completion: 14 tokens in 1.43s (27.3 tok/s), prompt: 21
```

### DIFF
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -210,6 +210,24 @@ def _model_has_mtp_module(model: Any) -> bool:
     return hasattr(inner, "mtp") and getattr(inner, "mtp", None) is not None
 
 
+def _model_mtp_decode_enabled(model: Any) -> bool:
+    """Return the per-instance MTP decode flag captured at model load time.
+
+    ``mlx_lm_mtp._MTP_ACTIVE`` is a construction-time global: it is toggled
+    before each model load so patched model ``__init__`` methods know whether
+    to attach MTP heads. It must not be used as a decode-time eligibility
+    check because loading a later non-MTP model resets the global to False and
+    would disable MTP for already-loaded MTP models. Persist the decision on
+    the model instance instead and read it here.
+    """
+    candidates = [model]
+    for attr in ("language_model", "_language_model"):
+        inner = getattr(model, attr, None)
+        if inner is not None and inner is not model:
+            candidates.append(inner)
+    return any(bool(getattr(obj, "_omlx_mtp_decode_enabled", False)) for obj in candidates)
+
+
 def _batch_generator_allows_mtp_activation(batch_gen: Any) -> bool:
     """True when lazy MTP activation cannot race with a pending batch merge."""
     try:
@@ -229,12 +247,7 @@ def _mtp_common_eligible(gen_batch: Any) -> bool:
         return False
     if not _model_has_mtp_module(gen_batch.model):
         return False
-    try:
-        from . import is_mtp_active
-
-        if not is_mtp_active():
-            return False
-    except Exception:
+    if not _model_mtp_decode_enabled(gen_batch.model):
         return False
     uids = getattr(gen_batch, "uids", None)
     if uids is None or len(uids) == 0:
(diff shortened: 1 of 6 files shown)


jundot:
Thanks for the quick patch. The direction is right: MTP decode eligibility should be captured on the loaded model instance instead of reading the process-wide construction flag at decode time.

I’m going to close this PR and land the fix from my side because the change is incomplete as a repo update. It adds the instance marker path, but the surrounding eligibility contract and tests are not updated consistently, so the branch is not merge-ready as-is. I also want the VLM marker to track the same load-time decision as the attached head path, including the head-attached-for-weight-load but decode-disabled case.

I’ll fold the root cause fix into a maintainer patch with the existing eligibility tests updated and the wrapped `language_model` / `_language_model` paths covered. I’ll also credit your report and suggested fix direction in the commit message. Thanks again for narrowing down the issue.

Issue #1763, opened 2026-06-09
[BUG] preflight memory check over-estimates KV peak ~4x when turboquant_kv_enabled, causing false 413 rejections after #1448


fqx:
## Bug Description

After upgrading to 0.4.2 (which includes #1448 fix for VLM nested config), requests that previously completed successfully are now rejected with HTTP 413:

```
Preflight rejected (327872 tokens, cached=0): Prefill would require ~73.48 GB peak
(current 27.17 GB + KV+SDPA 46.31 GB) but ceiling is 44.00 GB
```

The model has `turboquant_kv_enabled=true, turboquant_kv_bits=4.0`. With 4-bit KV quantization the actual KV memory should be ~4x smaller (~11-12 GB), making the real peak ~39-40 GB — well under the 44 GB ceiling.

There are **two compounding bugs** producing the false rejection.

---

## Bug 1: turboquant KV dtype not accounted for (merged in #1763)

`_set_model_info_for_monitor()` in `scheduler.py` determines `dtype_size` from the model's **activation dtype** (bfloat16 → 2 bytes/element) and never checks `_turboquant_kv_bits`:

```python
# Current code — turboquant not considered
dtype_size = 2  # Default float16
if hasattr(self.model, "dtype"):
    if self.model.dtype == mx.float32:
        dtype_size = 4
    elif self.model.dtype == mx.bfloat16:
        dtype_size = 2
# _turboquant_kv_bits is never read here
```

`TurboQuantMSEState` (mode=`"mse"`, the default) stores:
- **Packed VQ indices**: `kv_bits/8` bytes per element → 0.5 bytes for 4-bit
- **float16 norm per (token, head)**: `2/head_dim` bytes per element → 0.015625 for head_dim=128
- **Effective**: `kv_bits/8 + 2/head_dim` ≈ **0.516 bytes/element** (vs 2 bytes for bfloat16)

This means the preflight **over-estimates KV peak by ~4x**, rejecting requests that fit fine.

### Timing issue

`_turboquant_kv_bits` is wired onto the scheduler in `engine/batched.py` and `engine/vlm.py` **after** `engine.start()` returns — but `_set_model_info_for_monitor()` is called **inside** `engine.start()`, so it always sees `_turboquant_kv_bits = None` at first call. The fix re-calls `_set_model_info_for_monitor()` after wiring `kv_bits`.

---

## Bug 2: stale `hd > 128` SDPA threshold causes ~1200x over-estimate (PR #1764)

[...]

PR #1764, opened 2026-06-09
fix(memory): treat MLX fused SDPA as O(n) for all head_dim

fqx:
## Problem

After #1448 fixed VLM nested-config walking, the stale `hd > 128` threshold in `estimate_prefill_peak_bytes` and `estimate_chunk_transient_bytes` became the dominant source of false-positive 413 rejections for models with `head_dim > 128`.

For **Qwen3.6-VL** (`head_dim = 256`) at 327 K tokens the SDPA estimate was:

| Path | Formula | Result |
|------|---------|--------|
| Old (`hd > 128`) unfused fallback | `n_q × chunk × full_kv_len × 4` | **~40 GB** |
| Correct fused kernel | `n_q × chunk × hd × 4` | **~32 MB** |

The ~1200x over-estimate pushed the preflight peak well past the memory ceiling → HTTP 413 on every request.

## Root cause

The `hd > 128` check assumed MLX only provided a fused (O(n) tiled) SDPA kernel for `head_dim ≤ 128`, materialising the full float32 attention-score matrix for larger values. That was true for MLX < 0.22 but has been false since then: `mx.fast.scaled_dot_product_attention` uses the same O(n) Metal tiling for all head_dim values. omlx 0.4.x already requires MLX ≥ 0.31, so the fallback path is never reached.

## Fix

Replace the hard-coded sentinel `128` with a module-level constant:

```python
# MLX fused SDPA supports arbitrary head_dim as of MLX 0.22.
# omlx 0.4.x requires MLX >= 0.31 — unfused fallback never executes.
_SDPA_FALLBACK_HEAD_DIM: float = float("inf")
```

Both `estimate_prefill_peak_bytes` and `estimate_chunk_transient_bytes` now use the fused O(n) formula unconditionally. The dead fallback branch is kept in place so it can be revived if a future MLX version re-introduces the restriction.

## Note

The turboquant KV dtype fix originally bundled in this PR was merged in #1763. This PR now contains only the SDPA threshold fix, rebased on top of that merge.

## Tests

Added `TestSdpaThreshold` (3 tests):
- Constant is `+inf`
- `head_dim=256` at 327 K tokens uses the fused formula in `estimate_prefill_peak_bytes`
- `head_dim=256` uses the fused formula in `estimate_chunk_transient_bytes`

All 17 tests in `test_memory_monitor_vlm_config.py` pass.

### DIFF
--- a/omlx/memory_monitor.py
+++ b/omlx/memory_monitor.py
@@ -34,6 +34,13 @@ except ImportError:
     HAS_MLX_METAL = False
     mx = None
 
+# MLX fused SDPA (mx.fast.scaled_dot_product_attention) uses an O(n) tiled
+# kernel for all head_dim values as of MLX 0.22.  omlx 0.4.x requires
+# MLX >= 0.31, so the unfused fallback path (full attention-score matrix
+# in float32) never executes.  Set the threshold to infinity so the
+# estimator always uses the compact O(n) formula.
+_SDPA_FALLBACK_HEAD_DIM: float = float("inf")
+
 
 @dataclass
 class MemoryInfo:
@@ -483,14 +490,13 @@ class MemoryMonitor:
         eff_chunk = min(chunk_size, new_tokens)
         full_kv_len = new_tokens + max(cached_tokens, 0)
 
-        if hd > 128:
-            # Fallback: full attention matrix materialized in float32
-            # scores [B, n_q, eff_chunk, full_kv_len] + output
-            # [B, n_q, eff_chunk, hd]
+        if hd > _SDPA_FALLBACK_HEAD_DIM:
+            # Unfused fallback: full attention matrix materialized in float32.
+            # Not reached on MLX >= 0.22 (omlx 0.4.x requirement).
             attn = n_q * eff_chunk * full_kv_len * 4
             attn += n_q * eff_chunk * hd * 4  # output buffer (small)
         else:
-            # Fused kernel: tiled, only output buffer
+            # Fused kernel: tiled O(n), only output buffer allocated.
             attn = n_q * eff_chunk * hd * 4
 
         # KV growth attributable to this request: only the new tokens.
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for isolating this. The current estimator is wrong because it still assumes head_dim > 128 means MLX materializes the full attention score matrix, and that is causing valid long-context VLM requests to fail with false 413s.

This PR fixes the urgent over-rejection problem, so I am going to merge it. One part is still too optimistic, though: current MLX does not seem to be purely output-buffer-only for high head_dim; it still uses some tiled scratch memory. I will land a follow-up on main to model that tiled scratch term conservatively and update the memory-guard tests.

jundot:
Adding more context on the follow-up I mentioned above.

The stale part in the old estimator was the assumption that `head_dim > 128` means current MLX materializes the full fp32 score matrix `[q_heads, chunk, kv_len]`. That is no longer a good model for MLX 0.31.x, and it caused false 413s for valid long-context VLM requests. This PR fixed that urgent overestimate.

The remaining issue is that the replacement estimate makes high-head-dim SDPA effectively output-buffer-only: `q_heads * query_tokens * head_dim * 4`. That removes `kv_len` entirely. In oMLX, that is too optimistic because prefill can reuse a large prefix-cache KV span while only computing a small suffix. For example, a request with `cached_tokens=99k` and `new_tokens=1k` only allocates KV for the 1k suffix, but those 1k query tokens still attend over the full ~100k reconstructed KV context.

This matters especially for VLM text configs such as Qwen/Gemma, where the LM dimensions are nested under `text_config` / `language_config` and can have `q_heads != kv_heads` with `head_dim=256`. KV growth should be estimated from KV heads and new tokens, but SDPA transient is query-head based and still scales with the full attention span.

Local MLX 0.31.x measurements do not match the old full-score estimate, but they also do not match output-only for high head_dim. They are closer to a bounded tiled scratch term, roughly `q_heads * min(query_tokens, tile) * kv_len * dtype_size`, plus the fp32 output buffer. I merged this PR because it fixes the user-visible false rejection problem; the follow-up keeps that fix while adding the tiled scratch term so admission and adaptive prefill throttling remain conservative enough to avoid later Metal allocation failures.

PR #1766, opened 2026-06-09
fix: count tracked model memory in pre-load admission (#1623)

### LINKED ISSUE #1623: v0.4.0 regression: Pre-load eviction fails when loading a second large model
I just updated to v0.4.0 and for the first time since using it, oMLX loaded two large models instead of switching them, leading to extreme memory pressure and serving refusals. I had my agent go through the omlx logs, and helped it compile the assessment below. The TL;DR is that the 0.4.0 pre-load eviction check uses `mx.get_active_memory()` (live Metal memory, ~28 GB) instead of the tracked accumulator (`_current_model_memory`, ~57 GB), so it under-estimates the true footprint and lets a second large model load without evicting the first.

### Description

When a large model is already loaded (~28 GB) and a request comes in to load a second large model (~28 GB), the engine pool's pre-load admission check does not evict the first model. Both load side-by-side, pushing memory to 57+ GB with no headroom. The first inference request fails with a prefill guard error, and I have to manually evict the first model to recover.

### Timeline from logs

```text
07:26:13  Qwen3.6-35B-A3B-oQ6-fp16 loaded     actual: 27.93 GB, total tracked: 28.92 GB
07:53:43  Loading model: gemma-4-31B-it-fp16-q4
          ← no eviction logged
07:54:45  gemma-4-31B-it-fp16-q4 loaded       actual: 28.18 GB, total tracked: 57.10 GB
07:54:45  First request → Prefill failed: predicted peak > 55.1 GB safety cap
          (95% of 58 GB ceiling)
07:54:50  [me] POST /admin/api/models/.../unload → 400: Model not loaded
          (enforcer evicted it at 07:54:46 but too late — request already dead)
```

### Root cause

The pre-load admission check in `get_engine()` at `engine_pool.py` lines 373–404 evaluates:

```python
current = max(mx.get_active_memory(), get_phys_footprint())
projected = current + entry.estimated_size
```

After Qwen settled, `mx.get_active_memory()` reflected only its weight memory (~28 GB). So:

```text
projected = 28 GB (qwen weights) + 28 GB (gemma estimated) = 56 GB
ceiling = 58 GB  → ✅ passes, no eviction
```

This misses two things:
1. `_current_model_memory` was 57.10 GB after both loaded — the tracked accumulator correctly reflected the overlap, but the live-memory check didn't use it.
2. Hot cache overhead — the 6 GB SSD hot cache is not in `mx.get_active_memory()`.

### What changed from 0.3.7 → 0.4.0

In 0.3.7, `get_engine()` called `_ensure_memory_available()` (line ~418 in the old tap source at `/opt/homebrew/Library/Taps/jundot/homebrew-omlx/omlx/engine_pool.py`), which checked:

```python
while self._current_model_memory + required > self._max_model_memory:
    victim = self._find_lru_victim()
```

With `max_model_memory = 50.40 GB` (from model discovery), Qwen's tracked memory (~29 GB) plus gemma's estimated size with 25% KV headroom (~33 GB) summed to ~62 GB, which correctly triggered proactive eviction.

In 0.4.0, `EnginePool.__init__()` no longer takes `max_model_memory` — the parameter was removed (compare with the 0.3.7 constructor signature). `_ensure_memory_available()` and `_max_model_memory` were removed entirely. The only remaining pre-load check is the live-memory projection against `_current_ceiling()` (the enforcer's dynamic ceiling), which is too optimistic because `mx.get_active_memory()` after settlement only shows the first model's weight memory, not the cumulative tracked total.

popfido:
## Summary

Fixes #1623 (v0.4.0 regression: a second large model loads side-by-side instead of switching, over-committing memory and failing the next prefill).

The pre-load admission check in `EnginePool.get_engine()` gated on live memory only:

```python
current = max(mx.get_active_memory(), get_phys_footprint())
projected = current + entry.estimated_size
if projected <= ceiling: break   # admit, no eviction
```

It never consulted the tracked accumulator `_current_model_memory`. After a model settles or idles, both `mx.get_active_memory()` and the process footprint can read **well below** the model's true resident size (Metal releases pooled memory; the SSD hot cache isn't counted), while `_current_model_memory` still reflects the committed total. So loading a second large model projected under the ceiling and was admitted **without evicting the first** — pushing memory to 57+ GB and failing the next request with a prefill memory-guard error.

This is a v0.4.0 regression: 0.3.7's `_ensure_memory_available()` (removed along with `max_model_memory` in the `memory_guard_tier` refactor) evicted using the accumulator.

## Fix

Add `self._current_model_memory` to the `max()`:

```python
current = max(
    mx.get_active_memory(),
    get_phys_footprint(),
    self._current_model_memory,
)
```

The accumulator is a floor on committed model memory that live reads can dip below. This also makes pre-load admission **consistent with the request-time prefill-eviction path** (`_evict_idle_lru_for_prefill`), which already takes `max(active, phys, _current_model_memory)` — the pre-load gate was the lone path missing it.

## Test plan

- [x] New regression test `test_eviction_when_live_memory_undercounts`: with `active = phys = 0` but the accumulator showing the pair over-commits, the second model must still evict the first. **Fails on `main` (model-a not evicted), passes with the fix.**
- [x] `pytest tests/test_engine_pool.py` — 83 passed (incl. the existing `test_eviction_before_load`, unaffected: its fixture already proxies phys to the accumulator, so `max(0, acc, acc) == acc`).
- [x] `pytest tests/test_process_memory_enforcer.py tests/test_scheduler_admission.py tests/test_scheduler_prefill_memory_guard.py tests/test_memory_monitor.py tests/test_server_prefill_memory_handler.py` — 177 passed.
- [x] No new ruff findings.

### DIFF
--- a/omlx/engine_pool.py
+++ b/omlx/engine_pool.py
@@ -416,7 +416,18 @@ class EnginePool:
             ceiling = self._current_ceiling()
             if ceiling > 0:
                 while True:
-                    current = max(mx.get_active_memory(), get_phys_footprint())
+                    # Consult the tracked accumulator alongside live memory:
+                    # after a model settles or idles, mx.get_active_memory() and
+                    # the process footprint can read well below the model's true
+                    # resident size, while _current_model_memory still reflects
+                    # the committed total. Using only live memory lets a second
+                    # large model load without evicting the first, over-
+                    # committing past the ceiling (#1623).
+                    current = max(
+                        mx.get_active_memory(),
+                        get_phys_footprint(),
+                        self._current_model_memory,
+                    )
                     projected = current + entry.estimated_size
                     if projected <= ceiling:
                         break
--- a/tests/test_engine_pool.py
+++ b/tests/test_engine_pool.py
@@ -982,6 +982,63 @@ class TestEnginePoolEviction:
         assert pool._entries["model-a"].engine is None
         assert pool._entries["model-b"].engine is not None
 
+    @pytest.fixture
+    def undercounting_memory_pool(self, small_mock_model_dir, monkeypatch):
+        """Pool where live Metal memory under-reports the committed footprint.
+
+        The #1623 condition: after a model settles/idles, both
+        `mx.get_active_memory()` and `get_phys_footprint()` can read well below
+        the model's true resident size, while the tracked accumulator
+        (`_current_model_memory`) still reflects it. Unlike `tight_memory_pool`,
+        this fixture does NOT proxy phys_footprint to the accumulator — it leaves
+        live memory at 0 so admission has to consult the accumulator itself.
+        """
+        pool = _make_pool(ceiling=2500)  # each model fits alone, not both
+        pool.discover_models(str(small_mock_model_dir))
+        monkeypatch.setattr("omlx.engine_pool.get_phys_footprint", lambda: 0)
+        monkeypatch.setattr("omlx.engine_pool.mx.get_active_memory", lambda: 0)
+        return pool
+
+    @pytest.mark.asyncio
+    async def test_eviction_when_live_memory_undercounts(
+        self, undercounting_memory_pool
+    ):
+        """#1623: a second model must still evict the first when live memory
+        under-reports but the tracked accumulator shows the pair over-commits.
+
+        Without consulting `_current_model_memory`, admission sees
+        `current = max(0, 0) = 0`, projects `0 + model-b` under the ceiling, and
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for the fix. I checked the pre-load admission path against the prefill eviction path, and counting `_current_model_memory` here makes the decision consistent when live memory under-reports. The regression test covers the missing case cleanly, and the related memory/scheduler tests pass on my side.

PR #1773, opened 2026-06-09
feat(tts): forward language parameter to TTS engine

apetersson:
The OpenAI-compatible /v1/audio/speech endpoint accepts an optional 'language' field (e.g. 'german', 'english') but it was silently dropped — AudioSpeechRequest had no language field, TTSEngine never received it, and the upstream Qwen3-TTS model fell back to lang_code "auto".

- Add language field to AudioSpeechRequest
- Accept and forward language in TTSEngine.synthesize() and stream_synthesize_pcm(), mapped to lang_code for Qwen3-TTS models
- Pass request.language in audio_routes.py for all three call sites: native streaming, segmented fallback, and non-streaming synthesis
- Log language in all TTS log lines

### DIFF
--- a/omlx/api/audio_models.py
+++ b/omlx/api/audio_models.py
@@ -33,6 +33,7 @@ class AudioSpeechRequest(BaseModel):
     model: str
     input: str
     voice: Optional[str] = None
+    language: Optional[str] = None
     instructions: Optional[str] = None
     speed: Optional[float] = 1.0
     response_format: Optional[str] = "wav"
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for the PR. The request and route-level forwarding look correct, and mapping the API `language` field to mlx-audio `lang_code` is the right backend hook for Qwen3-TTS and the other TTS models that expose `lang_code`.

One compatibility detail I noticed is that adding `language` before `speed` changes positional `TTSEngine.synthesize()` and `stream_synthesize_pcm()` callers. This looks good to me, and I am going to merge it; I will fold that signature-ordering fix and a focused passthrough test into a follow-up.

apetersson:
just a gotcha to note: not all models treat the language parameter the same. some want language like "english" or "german" some want it like "en" "de.

Issue #1276, opened 2026-05-16
feat: expose draft_window_size / draft_sink_size / verify_mode for long-context agentic workloads


fparrav:
## Problem

`dflash-mlx` exposes `draft_window_size`, `draft_sink_size`, and `verify_mode` as runtime parameters, but oMLX does not pass them through or expose them in `model_settings`.

In real agentic use with contexts >8k tokens, the DFlash acceptance rate drops significantly — sometimes to the point where DFlash is slower than plain autoregressive decoding. This is confirmed in bstnxbt/dflash-mlx#32, where the same target/draft pair (Qwen3.6-35B-A3B + DFlash draft) on M1 Max 64GB shows `acceptance=0.56` and `decode_tok_s=15` at 8.4k prompt tokens, compared to 50–60 tok/s baseline.

The dflash-mlx parameters that help tune this:
- `draft_window_size` — limits the draft model's sliding attention window (e.g. 1024 tokens), keeping acceptance rate stable at long contexts
- `draft_sink_size` — attention sink tokens always included regardless of window (e.g. 64)
- `verify_mode` — `auto` or `adaptive`; adaptive can reduce block size dynamically when acceptance drops

## Request

Since you're already planning a `dflash-mlx` pin upgrade (see #1257), it would be great to include exposure of these parameters in `model_settings` as part of that update. Something like:

```json
"dflash_draft_window_size": 1024,
"dflash_draft_sink_size": 64,
"dflash_verify_mode": "adaptive"
```

This would allow users to tune DFlash for long-context agentic workloads rather than falling back to BatchedEngine or seeing degraded throughput.

## Environment

- Apple M4 Max, 64 GB
- target: `Qwen3.6-35B-A3B-TurboQuant-MLX-4bit`
- draft: `z-lab/Qwen3.6-35B-A3B-DFlash` (w8a32)
- `dflash_max_ctx: 65536` (currently used as fallback threshold)

jundot:
Thanks for the detailed repro. You're right — dflash-mlx already accepts those three as `runtime_config_from_defaults` kwargs, but my `DFlashEngine._build_runtime_context()` was only passing the prefix-cache fields through, so users were stuck on the 1024 / 64 / "adaptive" defaults.

Wired the three knobs end-to-end (ModelSettings → admin routes → DFlashEngine → dflash-mlx runtime config) with `None` meaning "let dflash-mlx pick its default". They show up under "Long-context tuning" in the model settings modal. Will land in the next release.

PR #1781, opened 2026-06-10
feat: expose DFlash copyspec_mode (prompt-lookup drafting) as a per-model setting

popfido:
## Summary

dflash-mlx ships prompt-lookup (copyspec) drafting with three policies — `conservative` / `auto` / `off` — but omlx never passes the `copyspec_mode` kwarg to `runtime_config_from_defaults`, so every model is hard-pinned to `conservative`. (The `DFLASH_COPYSPEC_MODE` env var doesn't help either: dflash-mlx only resolves it on its CLI path, not the programmatic one omlx uses.)

`conservative` permanently disables copyspec for the rest of a request after **one** missed copy block. In chat/agentic serving the first copy attempt often whiffs early in a response, killing copyspec even when the output later becomes extremely copy-heavy — e.g. a coding agent re-emitting a 2,000-token file with a 3-line change. `auto` (a windowed A/B latch in dflash-mlx) recovers exactly that case: it probes copyspec-on vs -off throughput, latches to the winner, and self-disables with exponential backoff on workloads where copy never pays.

This PR wires `dflash_copyspec_mode` through the same path as `dflash_verify_mode` (#1276 lineage), defaulting to `None` → dflash-mlx's own default (`conservative`), so behavior is unchanged unless a user opts in.

**No dflash-mlx bump needed** — the pinned `0.1.9 @ 5d70fae` already contains `copyspec_mode` (incl. the `auto` gate); verified against the installed package.

## Changes

- `omlx/model_settings.py` — new `dflash_copyspec_mode: Optional[str]` field + docstring
- `omlx/model_profiles.py` — added to the model-specific profile field list
- `omlx/admin/routes.py` — request field + update handler (validates `conservative|auto|off`, anything else → `None` = dflash default)
- `omlx/engine/dflash.py` — read from settings, pass `copyspec_mode=` into `runtime_config_from_defaults`
- `omlx/admin/static/js/dashboard.js` / `_modal_model_settings.html` — Copyspec mode select in the DFlash long-context tuning section (load, save, defaults-reset)
- `tests/test_dflash_engine.py` — extended the existing knob tests: dataclass default, `to_dict` omission when `None`, `from_dict` round-trip, engine reads setting, kwarg reaches `RuntimeContext.runtime`, dflash-mlx default fill-in

## Test plan

- [x] `pytest tests/ -m "not slow and not integration"` — 5432 passed, 22 skipped
- [x] `tests/test_admin_profiles_api.py` field-accounting test confirms the new field is classified (profile-eligible)
- [x] `runtime.copyspec_mode == "auto"` asserted end-to-end through `_build_runtime_context()` against the pinned dflash-mlx
- [ ] Manual: flip a DFlash model to `auto` in the dashboard and confirm copyspec counters move in `/v1/internal/metrics` on a copy-heavy prompt

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -145,6 +145,7 @@ class ModelSettingsRequest(BaseModel):
     dflash_draft_window_size: int | None = None
     dflash_draft_sink_size: int | None = None
     dflash_verify_mode: str | None = None
+    dflash_copyspec_mode: str | None = None
     # Native MTP (mlx-lm PR 990 / PR 15 monkey-patch)
     mtp_enabled: bool | None = None
     # VLM MTP speculative decoding via external assistant drafter (mlx-vlm 191d7c8+)
@@ -2282,6 +2283,13 @@ async def update_model_settings(
         # Anything else (including empty string) → revert to dflash default.
         current_settings.dflash_verify_mode = (
             value if value in ("dflash", "adaptive", "ddtree", "off") else None
+        )
+    if "dflash_copyspec_mode" in sent:
+        value = request.dflash_copyspec_mode
+        # dflash-mlx accepts: conservative | auto | off.
+        # Anything else (including empty string) → revert to dflash default.
+        current_settings.dflash_copyspec_mode = (
+            value if value in ("conservative", "auto", "off") else None
         )
 
     # Native MTP (mlx-lm PR 990 / PR 15 monkey-patch)
--- a/omlx/admin/static/js/dashboard.js
+++ b/omlx/admin/static/js/dashboard.js
@@ -1454,6 +1454,7 @@
                     dflash_draft_window_size: s.dflash_draft_window_size ?? null,
                     dflash_draft_sink_size: s.dflash_draft_sink_size ?? null,
                     dflash_verify_mode: s.dflash_verify_mode || 'adaptive',
+                    dflash_copyspec_mode: s.dflash_copyspec_mode || 'conservative',
                     dflash_compatible: model?.dflash_compatible !== false,
                     dflash_compatibility_reason: model?.dflash_compatibility_reason || '',
                     dflash_ssd_cache_available: !!model?.dflash_ssd_cache_available,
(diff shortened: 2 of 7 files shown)


joshuaswarren:
Hardware validation on current upstream `0436bdf` is positive.

I cherry-picked `4ba10f3` onto current `main`; the feature code merged cleanly. The only conflict was the name/docstring of an existing default-runtime test. **216 focused DFlash/settings/profile tests passed.**

Hardware/config: M1 Ultra 128 GB, Qwen3.8 heretic oQ4e target, Qwen3.8 DFlash2 draft dynamically quantized 4/16 group-64, block size 5, `verify=dflash`, C1, caches off. Test-only log instrumentation exposed dflash-mlx's existing CopySpec counters; it is not part of this PR.

| workload | off | conservative | auto | auto speedup |
| immediate source copy | 235.96 s | 191.92 s | **79.12 s** | **2.98x off / 2.43x conservative** |
| novel sentence, then exact source copy | 239.87 s | 227.65 s | **81.94 s** | **2.93x / 2.78x** |
| novel arithmetic | 46.87 s | 46.63 s | 48.38 s | 0.97x / 0.96x |

`auto` reached **68.8% acceptance** with **327/332 CopySpec hits** on the two copy workloads. Conservative managed 62/15 hits (18.6%/4.6% acceptance). On novel arithmetic, auto produced zero CopySpec hits and paid about 3%.

Every workload's output SHA was byte-identical across off/conservative/auto. The delayed-copy source completed exactly in all arms. Immediate copy reached the common 1,200-token cap in all arms, so that row establishes same-prefix parity rather than full-source completion.

**Verdict:** this validates the PR's exact rationale—`auto` recovers after early misses and self-disables where copying does not pay. I recommend rebasing/merging with the existing `None` default, then opting `auto` into copy-heavy single-stream workloads. It does not change our concurrent advisor verdict; DFlash itself remains net-negative at C2 on this hardware.


Happy to provide a sanitized raw-results bundle or the rebased patch if useful.

Issue #1003, opened 2026-04-29
Main Agent Session KV Prefix Cache Always Misses After Upgrading to 0.3.8.x (Working on 0.3.6)


Keenni:
# Main Agent Session KV Prefix Cache Always Misses After Upgrading to 0.3.8.x (Working on 0.3.6)

## Environment
- **oMLX**: `0.3.6` (working) → upgraded to `0.3.8.dev3` / `0.3.8rc1` (broken) → downgraded back to `0.3.6` (working again)
- **OS**: macOS 15.x, Apple Silicon (M5 Max, 128GB)
- **Model**: `Qwen3.6-35B-A3B-oQ6` (Qwen3.6 MoE, thinking enabled)
- **OpenClaw omlx provider** connecting to local omlx server on port 8000

## Symptom

After upgrading omlx from `0.3.6` to `0.3.8.x`, **the main agent session never hits the KV prefix cache**. Every request requires full prefill recomputation.

- **Main session**: KV cache always misses → full prefill every time → extremely slow
- **Subagent sessions**: KV cache hits normally (no regression)
- **API endpoint works**: Direct curl to `/v1/chat/completions` responds fine
- **Downgrading to 0.3.6**: Immediate fix — main session KV cache hits restored

This is a **KV prefix cache matching regression**, not a connectivity issue.

## Root Cause Analysis

**The `preserve_thinking=True` auto-setting in 0.3.8 changes the token sequence of historical messages, breaking prefix cache matching.**

### What changed in v0.3.7 (commit `0749e3fdcc2b`)

The `preserve_thinking` feature was introduced:

1. `detect_preserve_thinking()` in `model_discovery.py` now detects Qwen 3.6+ templates that support `preserve_thinking`
2. **Auto-setting**: When `enable_thinking` is active (which it is for this model), omlx **automatically sets `preserve_thinking=True`** (in `server.py`):
   ```python
   if merged_ct_kwargs.get("enable_thinking") is not False and "preserve_thinking" not in merged_ct_kwargs:
       merged_ct_kwargs["preserve_thinking"] = True
   ```
3. **Thinking block reconstruction** in `extract_text_content` (`api/utils.py`): When OpenClaw sends assistant messages with `reasoning_content`, omlx now reconstructs thinking blocks:
   ```python
   # In native=False mode (non-VLM):
   content = f"\n\n{text}"
   # In native=True mode: passes reasoning_content as separate field
   ```

### Why this breaks KV prefix cache

[...]

PR #1784, opened 2026-06-10
debug: prefix-cache divergence probe for always-miss diagnosis (#1003)

popfido:
## Summary

Diagnostic instrumentation for #1003 (KV prefix cache always misses for long agent sessions on thinking models since 0.3.8). Hit counters alone can't show **where** a new prompt stops matching what was cached — which is exactly the question that separates the candidate causes (template re-render drift vs client-side history edits vs eviction). This is the log you asked the reporter to capture, made decisive.

## What it does

When debug logging is enabled, the scheduler keeps the token sequences of the last 4 stored cache entries (`_cache_probe_seqs`, deque maxlen=4). On each `add_request`, after the prefix-cache lookup, it logs:

```
Request <id>: prefix probe vs stored <ref>: common_prefix=P/R tokens (~B blocks of 2048), served cached_tokens=C, prompt=N
Request <id>: first divergence at token P: stored=...'<decoded ±12 tokens>' vs prompt=...'<decoded ±12 tokens>'
```

Zero cost at default log levels — both the recording and the probe are gated on `logger.isEnabledFor(logging.DEBUG)`.

## Offline findings that motivated this (round-trip harness vs the real Qwen3.6 template + tokenizer)

Simulating multi-turn agent traffic through omlx's actual parse/echo/re-render code paths:

- `preserve_thinking=True` + client echoes `reasoning_content` on all turns + canonical emission whitespace → **100% token-exact prefix reuse**, including through structured tool-call round-trips. The 0.3.8 feature is not inherently cache-breaking.
- Non-canonical emission whitespace (e.g. `\n\n</think>`) and Anthropic-style echo-last-only clients each cause divergence — but always **tail-bounded** (last 1–2 turns), never a head wipe-out.
- So none of the omlx-side mechanisms reproduce the reported *total* always-miss; the live divergence offset (which this probe logs) is the missing evidence.

Related observation for a follow-up: `needs_think_prefix` store carve-out (scheduler, c8ee5ea) caches only prompt tokens for reasoning models — correct pre-0.3.7, but with `preserve_thinking=True` it now discards generated tokens that the next prompt *does* contain, capping best-case reuse at the previous prompt boundary.

## Test plan

- [x] 10 new unit tests (`tests/test_prefix_divergence_probe.py`): divergence offset + context logging, full-match summary-only, best-of-4 selection, empty-state no-ops, decode-failure tolerance
- [x] `pytest tests/ -m "not slow and not integration"` — 5442 passed, 22 skipped

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -1198,6 +1198,12 @@ class Scheduler:
         # Track active specprefill request for RoPE cleanup
         self._specprefill_active_request_id: str | None = None
 
+        # DEBUG-only prefix-cache divergence probe (issue #1003): recent
+        # stored cache sequences, so a miss can be traced to the exact
+        # token where the new prompt diverges from what was cached.
+        # Populated only when debug logging is enabled — zero cost otherwise.
+        self._cache_probe_seqs: deque[tuple[str, list[int]]] = deque(maxlen=4)
+
         # VLM MTP: gemma4_assistant drafter attached by VLMBatchedEngine.
         # When set, eligible requests bypass mlx-lm BatchGenerator for decode
         # and run through mlx-vlm's _mtp_rounds round loop instead.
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for adding this. This is the targeted #1003 diagnostic I wanted: it shows the first prefix divergence without dumping the full prompt.

Looks good to me, and I’m going to merge it.

Issue #1774, opened 2026-06-09
Memory enforcer fails under concurrent load in 0.4.3 — process grows to 116 GB on a ~27 GB model pair, eviction reports `freed=-51GB` and breaches the configured ceiling → host OOM


ChildOTK:
**Describe the bug**
Under sustained concurrent inference load, the omlx 0.4.3 process grows far beyond the loaded model footprint and beyond its own configured memory ceiling, then fails to recover, taking the whole host out of memory.

Only the normal 2-model pair was ever loaded (~27 GB combined):
`mlx-community/Qwen3.5-4B-8bit` + a `Qwen3.6-35B-A3B-oQ6-mtp` (35B MoE, MTP enabled, dflash in-memory cache on). No large models were loaded. Despite that, the process climbed to 80 GB, then during a routine soft-pressure eviction it spiked to 115.9 GB in ~17 s, blew past its configured 90.6 GB ceiling, and the enforcer then looped on "hard memory pressure / requested idle reclaim" indefinitely with no recovery. macOS reported "out of application memory" and suspended all apps.

Two things in the logs look like the actual defect:
1. The eviction/settle path reports **negative freed memory** — `freed=-51.56GB` (memory *grew* by 51 GB while trying to free 3 GB).
2. The same model was loaded multiple times across the session (35B ×4, 4B ×2).  4 × ~22 GB ≈ the ~88 GB excess over the 27 GB the pair should occupy, which suggests evicted/reloaded model instances are **not being freed** (a leak in the eviction/reload path), and/or the settle barrier allocates instead of releasing.

**To Reproduce**
1. macOS on Apple Silicon with 128 GB unified memory; install omlx 0.4.3 via the Homebrew tap. Default memory config (`max_process_memory: auto` → ~90.6 GB ceiling, soft 0.85 / hard 0.95, `prefill_memory_guard: true`).
2. Load a 2-model pair totaling ~27 GB (a 4B dense + a 35B MoE with MTP and `dflash_in_memory_cache: true`).
3. Drive sustained **concurrent, multi-turn** inference (dozens of long-context requests, briefly overlapping clients) so process memory climbs toward the soft threshold and triggers an idle-model eviction.
4. Observe the settle barrier report negative freed memory, the process breach the 90.6 GB ceiling to ~116 GB, emergency reclaim fail, and the enforcer loop on "hard memory pressure" → host OOM.

**Expected behavior**
[...]

PR #1785, opened 2026-06-10
fix(engine_pool): skip the settle wait when other engines are serving

JimStenstrom:
Addresses the `freed=-51GB` / eviction-stall portion of #1774 (it does not close the issue — it removes the eviction-time amplification under concurrent load). Complements the fixes already in 0.4.4.dev1, none of which touch the settle barrier itself.

## Summary

- `_unload_engine()`'s settle barrier verifies an unload by polling `mx.get_active_memory()` and comparing the delta against the model's estimated size. That gauge is process-global: while another engine allocates (prefill/KV growth), the delta no longer measures this unload — under concurrent load it reads negative, which is exactly the `freed=-51GB` in #1774's report.
- The barrier then burns all 10 settle rounds plus 3 emergency-reclaim rounds — ~8s of sleeps and 13 `gc.collect()` + `mx.synchronize()` + `mx.clear_cache()` cycles (beyond the initial release pass) serialized onto the MLX executor against live decode — while the memory enforcer holds the pool lock across the eviction, so admission is blocked too. Under pressure the eviction itself becomes a stall-and-spike, consistent with the reporter's "routine soft-pressure eviction … spiked to 115.9 GB in ~17 s".
- Fix: bail out of the settle wait when any other loaded entry is serving (active requests, or a held in-use lease from #1667). One INFO line, no retry rounds, no emergency reclaim. With the pool otherwise idle the barrier is byte-for-byte unchanged, so the original stuck-unload protection (#768) is fully preserved.

## Why this is safe

- **Accounting is unchanged**: the bail-out path accounts the unload exactly as the existing timed-out path already does.
- **Unreleased memory stays visible**: pre-load admission reads `max(mx.get_active_memory(), get_phys_footprint(), _current_model_memory)` (the #1623 guard), so the live gauge dominates while anything is still resident — and `_unload_engine` ends with `_wake_process_memory_enforcer()`, so the enforcer re-polls immediately rather than waiting for the next tick. Skipping the emergency reclaim defers recovery to those two paths instead of losing it.
- **Conservative failure direction**: if an engine doesn't expose `has_active_requests()` (AttributeError, matching `_find_lru_victim`'s idiom), the entry simply doesn't count as serving and the barrier falls back to today's full wait.
- **Policy**: this continues the direction of the 0.4.3 enforcer change and the #1770 follow-up — control paths shouldn't fight the process-global MLX gauge while other work is in flight.

## Trade-off (stated)

Under sustained concurrent pressure the enforcer's eviction loop now paces faster (no ~8s barrier stall between victims), so it can evict an additional idle victim where the old code would have stalled first. The old path reached the same evictions after timing out — just slower, with admission blocked and the executor saturated by sync/clear rounds in the meantime. If a pacing pause is still wanted there, a small enforcer-side delay would be the right knob; happy to follow up. Same offer for a `settle_indeterminate` counter in the pool stats (the #1406 pattern) if INFO-level logging feels too quiet for this condition.

## Changes

`omlx/engine_pool.py` only:

- New `_other_entries_serving(model_id)` — true when any other loaded entry has `in_use > 0` or `has_active_requests()`. Iterates a snapshot of `_entries` since the admin unload routes call `_unload_engine` without the pool lock.
- Settle loop: when a round fails the freed check and another entry is serving, log the indeterminate sample at INFO and break; the emergency-reclaim rounds are skipped on that path (the in-code comment documents the recovery story above).
- Every eviction flavor funnels through `_unload_engine` (enforcer soft/hard pressure, pre-load admission eviction, prefill-headroom eviction, TTL expiry, admin unload), so this single change covers all callers.

## Tests

`tests/test_engine_pool.py::TestMemorySettleBarrier`:

- `test_settle_bails_out_under_concurrent_activity` — second entry serving + rising global gauge: the barrier exits after one sample (no 0.5s retries, no timeout warning, no emergency reclaim, exactly the one initial executor sync/clear cycle), and the unload still completes and is accounted.
- `test_settle_still_waits_when_pool_otherwise_idle` — same rising gauge with no other entry serving: full barrier behavior preserved (10 retries + emergency reclaim, 14 executor sync/clear cycles).
- `test_other_entries_serving_in_use_lease_counts` — the #1667 acquire-vs-use lease counts as serving.

```bash
.venv/bin/python -m pytest tests/test_engine_pool.py tests/test_process_memory_enforcer.py -q
```

```text
201 passed
```

Also verified the new tests fail when the fix is reverted: restoring main's `engine_pool.py` fails `test_settle_bails_out_under_concurrent_activity` (the barrier burns every round) and `test_other_entries_serving_in_use_lease_counts`.

### DIFF
--- a/omlx/engine_pool.py
+++ b/omlx/engine_pool.py
@@ -676,6 +676,28 @@ class EnginePool:
                 await self._unload_engine(victim)
                 evicted_any = True
 
+    def _other_entries_serving(self, model_id: str) -> bool:
+        """True when any loaded entry other than ``model_id`` is serving.
+
+        Used by the settle barrier in ``_unload_engine``: the barrier's
+        freed-memory check is a delta of the process-global
+        ``mx.get_active_memory()`` gauge, which only measures THIS unload
+        while no other engine is allocating concurrently.
+        """
+        # Snapshot the items: admin unload routes call _unload_engine without
+        # the pool lock, so discover_models() can mutate _entries mid-iteration.
+        for mid, e in list(self._entries.items()):
+            if mid == model_id or e.engine is None:
+                continue
+            if e.in_use > 0:
+                return True
+            try:
+                if e.engine.has_active_requests():
+                    return True
+            except AttributeError:
+                pass
+        return False
+
     async def _unload_engine(self, model_id: str) -> None:
         """
         Immediately stop and unload an engine with memory settle barrier.
@@ -759,6 +781,7 @@ class EnginePool:
         settle_tolerance = max(2 * 1024**3, int(entry.estimated_size * 0.05))
         min_expected_freed = max(0, entry.estimated_size - settle_tolerance)
         settled = False
+        settle_indeterminate = False
         for _settle_round in range(10):
             active_now = mx.get_active_memory()
             actual_freed = pre_unload_active - active_now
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the follow-up. I checked that this only shortcuts the settle wait when another engine is actually serving; idle loaded engines still keep the existing full settle path. The local test run passed with 201 tests, so this looks good to me and I'm going to merge it.

PR #1792, opened 2026-06-10
Fix TurboQuant Cache Reconstruction Crash and Performance Bottleneck

felk-dev:
This patch resolves two critical issues related to TurboQuant prefix caching: a crash when reloading boundary snapshots, and a massive performance regression in token generation/prefill speeds after a cache hit.

1. Fix AttributeError crash on boundary snapshots with TurboQuant
omlx/cache/prefix_cache.py

`omlx.cache.prefix_cache - WARNING - Failed to reconstruct cache: 'mlx.core.array' object has no attribute 'norms'`

Issue: When boundary snapshots are created, sliceable cache layers are intentionally skipped to save space, leaving mx.zeros((1,)) placeholder tensors. During cache reconstruction, the __turboquant_v2__ loading logic did not filter out these placeholders. It appended them to the state list and passed them to _concat_state(), which crashed with AttributeError because it expected the norms and indices attributes of a TurboQuantMSEState.

Solution:

Added a check to skip (1,) placeholder tensors inside the block data loop.
Implemented a robust fallback mechanism that seamlessly handles mixed-state blocks (e.g., a mix of true fp16 cache blocks from a draft model and TQ blocks) by selectively dequantizing TQ blocks only when necessary.
2. Eliminate massive TQ cache hit performance regression
Files: omlx/cache/prefix_cache.py, omlx/scheduler.py

Issue: After successfully reloading a TurboQuant cache, prefill and token generation speeds were drastically reduced (often by >50%). This occurred due to two cascading issues:

reconstruct_cache unconditionally dequantized the TQ blocks into fp16 arrays, relying on _apply_turboquant_kv_convert in the scheduler to re-quantize them. This dequantize -> quantize path created a massive, unevaluated lazy computational graph.
The _collect_cache_storage_arrays function in the scheduler (responsible for pre-evaluating cache arrays to prevent lazy evaluation stalls) only checked for strict mx.array instances. Since TurboQuant stores its tensors inside dataclasses (like TurboQuantMSEState), the arrays were completely ignored by mx.eval. As a result, the massive graph was evaluated lazily token-by-token during generation.
Solution:

Zero-Overhead Reconstruction (prefix_cache.py): Optimized the reconstruction logic to directly return the assembled TurboQuantKVCache object when all valid blocks are TQ. This completely bypasses the dequantize -> requantize cycle, keeping operations entirely in the quantized space.
Recursive Array Materialization (scheduler.py): Refactored _collect_cache_storage_arrays to use a recursive _extract helper. It now robustly traverses dataclasses, tuples, and lists, ensuring that all nested mx.arrays (like norms and indices) are correctly collected and materialized via mx.eval before generation begins.
Fix skip_last Logic (scheduler.py): Updated _apply_turboquant_kv_convert to account for TurboQuantKVCache layers when determining the last_kv_idx. This ensures the turboquant_skip_last optimization remains intact even when the cache is loaded directly as TQ layers.

### DIFF
--- a/omlx/cache/prefix_cache.py
+++ b/omlx/cache/prefix_cache.py
@@ -1884,59 +1884,79 @@ class BlockAwarePrefixCache(CacheManager):
                         _state_length,
                     )
 
-                    key_states, value_states = [], []
+                    current_tq_ks, current_tq_vs = [], []
+                    valid_fp16_ks, valid_fp16_vs = [], []
+                    
+                    tq_bits, tq_seed = 4.0, 0
+                    if first_block_meta_states and layer_idx < len(first_block_meta_states):
+                        ms = first_block_meta_states[layer_idx]
+                        if isinstance(ms, (list, tuple)) and len(ms) >= 3:
+                            tq_bits = float(ms[1])
+                            tq_seed = int(ms[2])
+
                     for block_data in all_block_data:
                         if layer_idx >= len(block_data):
                             continue
                         bd = block_data[layer_idx]
                         if isinstance(bd, tuple) and len(bd) == 2:
                             if isinstance(bd[0], str) and bd[0] == "__turboquant_v2__":
-                                ks, vs = bd[1]
+                                current_tq_ks.append(bd[1][0])
+                                current_tq_vs.append(bd[1][1])
                             else:
                                 ks, vs = bd
-                            key_states.append(ks)
-                            value_states.append(vs)
-                    if not key_states:
-                        logger.debug(f"TQ layer {layer_idx}: no block data")
-                        return None
-                    # Concatenate along token dimension
-                    cat_ks = key_states[0]
-                    for s in key_states[1:]:
-                        cat_ks = _concat_state(cat_ks, s)
-                    cat_vs = value_states[0]
-                    for s in value_states[1:]:
-                        cat_vs = _concat_state(cat_vs, s)
-                    try:
-                        from mlx_lm.models.cache import KVCache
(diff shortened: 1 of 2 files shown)


felk-dev:
https://github.com/jundot/omlx/issues/1793

jundot:
Thanks @felk-dev for the PR and for linking it to #1793.

I reproduced the issue on a real dense model and confirmed the root cause is the chunked-prefill completion path storing the fp16 tail into an existing TurboQuant prefix chain. I fixed this on `main` in 0b7a226 by converting the completed chunked-prefill fp16 cache to TurboQuant before it is inserted into decode, matching the external-prefill path, plus a small reconstruction guard for already-mixed chains.

Since the root-cause fix is now on `main`, I am going to close this PR rather than merge the restore-side mixed-concat approach.

felk-dev:
@jundot MoE Model(like Qwen 3.6 35b a3b), #1793 patch is not works. Please review about MoE model too

jundot:
Thanks @felk-dev, I reproduced this on `Qwen3.6-35B-A3B-oQ4-mtp`.

The missing case was the `ArraysCache` hybrid restore path: restored `ArraysCache` layers come back as `SizedArraysCache`, so the TurboQuant eligibility check skipped the chunked-prefill TQ conversion and stored the full-attention layers as `KVCache`.

I fixed this on `main` in 009a5c1. The same Qwen3.6 smoke test now restores 4096 cached tokens on turn 3 instead of truncating to 2048, with the full-attention layers staying `TurboQuantKVCache`.

PR #1799, opened 2026-06-10
fix: normalise logits_processors row slots dropped to None by batch merge

### LINKED ISSUE #1798: Concurrent plain + structured requests crash: GenerationBatch.extend() re-introduces logits_processors None row slots (gap left by #1747)
## Environment

- oMLX 0.4.3 (app bundle) — also reproduced from source at `e892ef4` (0.4.4.dev1)
- mlx-lm at the pinned `39c4019` from pyproject
- macOS 26.5.1, Apple Silicon (M5 Pro, 64 GB)
- Model: `mlx-community/Qwen3.6-35B-A3B-nvfp4` (vlm engine), thinking enabled
- Default scheduler settings (`max_concurrent_requests = 8`)

## Symptom

As soon as a plain chat completion (no logits processor) and a structured request (`response_format: json_schema` → grammar processor) are decoding **concurrently in the same batch**, the scheduler enters a corruption-recovery loop and ends up failing every in-flight request:

```
17:29:02 WARNING Cache corruption detected: 'NoneType' object is not iterable, clearing cache and re-prefilling...
17:29:03 WARNING Cache corruption detected: 'NoneType' object is not iterable, clearing cache and re-prefilling...
17:29:05 WARNING Cache corruption detected: 'NoneType' object is not iterable, clearing cache and re-prefilling...
17:29:07 ERROR Error during chat streaming: Cache corruption not recoverable after retries: 'NoneType' object is not iterable
```

Each streaming client receives `data: {"error": {"message": "Cache corruption not recoverable after retries: 'NoneType' object is not iterable", "type": "server_error"}}` followed by `[DONE]`. In my runs the first mixed wave killed 8/8 concurrent requests on the 0.4.3 bundle (6/8 on source) within ~7-9 s, deterministically. Single requests are fine, and clearing caches does not help — it is not cache state, it is a logic bug.

## Debug traceback

With `log_level: debug` (paths shortened):

```
File "omlx/scheduler.py", line 7672, in step
File ".../mlx_lm/generate.py", line 1866, in next_generated
File "omlx/patches/mlx_lm_mtp/batch_generator.py", line 192, in patched_bg_next
File ".../mlx_lm/generate.py", line 1775, in _next
File "omlx/patches/mlx_lm_mtp/batch_generator.py", line 143, in patched_next
File ".../mlx_lm/generate.py", line 1415, in next
File "omlx/scheduler.py", line 395, in _patched_generation_batch_step
File ".../mlx_lm/generate.py", line 1346, in _step
    for processor in self.logits_processors[e]:
TypeError: 'NoneType' object is not iterable
```

## Root cause

#1747 fixed #934 by always passing per-row `logits_processors` as a list (`[[]]`, never `None`) at the insert call sites, plus a whole-list `None → []` guard at the top of `_patched_generation_batch_step`.

That is not sufficient: on a heterogeneous continuous-batch **merge**, mlx-lm's `GenerationBatch.extend()` re-introduces `None` row slots:

```python
# mlx_lm/generate.py (pinned 39c4019)
def extend(self, batch):
    ...
    if not any(self.logits_processors):
        self.logits_processors = [None] * len(self.uids)      # ← re-None-ifies
    logits_processors = (
        batch.logits_processors
        if any(batch.logits_processors)
        else [None] * len(batch.uids)                          # ← and here
    )
    self.logits_processors.extend(logits_processors)
```

`any([[], []])` is `False`, so the empty-list shape that #1747 ships **collapses back to `None`** whenever a batch with no active processor merges with a grammar-constrained one. The next `_step` hits `for processor in self.logits_processors[e]:` on the `None` slot and raises. The whole-list guard does not see it: the list itself is non-None, only a row slot is.

The corruption recovery then clears everything, re-prefills, and **re-merges the same rows**, so it crashes again every ~1.5 s until the retry budget fails every request — which is why this presents as "cache corruption not recoverable" instead of a one-off error.

## Reproduction

Against a running server with a thinking-enabled model (adjust port/model):

```python
import json, threading, time, urllib.request

BASE = "http://127.0.0.1:1234/v1/chat/completions"
MODEL = "Qwen3.6-35B-A3B-nvfp4"

SCHEMA = {"type": "json_schema", "json_schema": {"name": "extract", "strict": True, "schema": {
    "type": "object", "properties": {"items": {"type": "array", "items": {
        "type": "object", "properties": {"label": {"type": "string"}, "gloss": {"type": "string"}},
        "required": ["label", "gloss"], "additionalProperties": False}}},
    "required": ["items"], "additionalProperties": False}}}

def fire(body, tag):
    req = urllib.request.Request(BASE, data=json.dumps(body).encode(),
                                 headers={"Content-Type": "application/json"})
    with urllib.request.urlopen(req, timeout=300) as r:
        for raw in r:
            line = raw.decode().strip()
            if line.startswith("data:") and '"error"' in line:
                print(tag, "ERROR CHUNK:", line[:160])

threads = []
for i in range(3):  # plain chats (no logits processor), thinking on
    t = threading.Thread(target=fire, args=({
        "model": MODEL, "stream": True, "reasoning_effort": "high",
        "messages": [{"role": "user", "content": f"Explain step by step how a diesel engine works ({i})"}],
    }, f"plain-{i}")); t.start(); threads.append(t)
time.sleep(0.4)  # let them enter decode
for i in range(3):  # structured requests (grammar logits processor) join the running batch
    t = threading.Thread(target=fire, args=({
        "model": MODEL, "stream": True, "temperature": 0, "reasoning_effort": "none",
        "response_format": SCHEMA,
        "messages": [{"role": "user", "content": f"List 8 software engineering concepts with a short gloss ({i})"}],
    }, f"structured-{i}")); t.start(); threads.append(t)
for t in threads: t.join()
```

First wave crashes every time on 0.4.3 and on source @ `e892ef4`.

## Fix

Normalising **every per-row slot** (not just the whole list) at the `_patched_generation_batch_step` chokepoint covers both the insert and the merge origins. With that one change the same harness completes 4 rounds / 32 mixed concurrent requests with zero corruption markers. PR incoming.

Possibly related: the constant clear-and-re-prefill loop this triggers under mixed workloads might contribute to "prefix cache always misses" reports (#1003), though I have not verified that link.

efortin:
## Summary

While running mixed workloads against a single model — a regular chat completion decoding next to a structured `response_format: json_schema` request — the server reliably falls into the corruption-recovery loop and ends up killing every in-flight request with `Cache corruption not recoverable after retries: 'NoneType' object is not iterable`. Single requests are fine, clearing caches changes nothing, and it reproduces on the first mixed wave every time. Full analysis and a standalone repro script are in #1798.

It turns out this is not cache state at all. #1747 made sure we always hand `logits_processors` to `BatchGenerator.insert()` as per-row lists (`[[]]`, never `None`), which fixed the single-request crash from #934. But when the scheduler merges two running batches, mlx-lm's `GenerationBatch.extend()` rebuilds the row slots on its own:

```python
# mlx_lm/generate.py
if not any(self.logits_processors):
    self.logits_processors = [None] * len(self.uids)
```

`any([[], []])` is `False`, so the empty-list shape we pass at insert time collapses right back to `None` slots whenever a batch with no active processor merges with a grammar-constrained one. The next `_step` then iterates `self.logits_processors[e]` over a `None` slot and raises. Recovery clears the caches, re-prefills, re-merges the same rows… and crashes again, every ~1.5 s, until the retry budget gives up and fails everyone. That is why it presents as unrecoverable cache corruption when it is really a batch-merge shape bug.

## Fix

Normalise the row slots — not just the whole list — in `_patched_generation_batch_step`, right before the original step runs. That is the one chokepoint both origins (insert and merge) funnel through, and it also protects our own grammar-accept loop right below, which iterates the same slots. The insert-side wrapping from #1747 stays as-is.

## Tests

- unit test of the chokepoint normalisation (monkeypatched original step, no model needed)
- source-level assertion so the per-row normalisation cannot be silently dropped later
- integration test reproducing the actual gap: a grammar row decoding, then a plain row carrying the `[[]]` "fix shape" joining through `extend()` — crashes without this patch, passes with it (gated behind `VLLM_MLX_INTEGRATION=1` like the existing #934 tests)
- the module docstring of the regression tests now documents the merge origin

## Verification

Before: a first wave of 3 streaming chats (thinking on) + 3 structured requests joining mid-decode deterministically killed 8/8 in-flight requests on the 0.4.3 bundle (6/8 on source @ e892ef4) in ~8 s, with the recovery loop visible in the log.

After: same harness, 4 rounds / 32 mixed concurrent requests — zero corruption markers in the debug log, zero request errors, every stream completes normally.

`pytest tests/test_scheduler_logits_processors.py -v -m "not integration"` → 6 passed.

Fixes #1798

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -386,11 +386,27 @@ def _patched_generation_batch_step(self):
         model.set_batch_rope_deltas(mx.array(deltas))
 
     # Defensive: mlx-lm's GenerationBatch._step does `any(self.logits_processors)`
-    # and `for p in self.logits_processors[e]`, both of which crash when
-    # logits_processors is None.  Normalise to [] so the original code path
-    # works without modification.  See #934.
+    # and `for p in self.logits_processors[e]`, both of which crash when a row
+    # slot is None.  Normalise the whole list AND every per-row slot to [] here,
+    # at the single consumption chokepoint, so the original step and the
+    # grammar-accept loop below are both safe regardless of slot origin.
+    #
+    # The insert call sites already wrap each request's processors as a list,
+    # but that is not enough: on a heterogeneous continuous-batch merge,
+    # mlx-lm's GenerationBatch.extend() re-introduces None slots via
+    # `if not any(self.logits_processors): self.logits_processors =
+    # [None] * len(self.uids)`.  `any([[], []])` is False, so empty-list slots
+    # collapse back to None whenever a batch with no *active* processor merges
+    # with a grammar-constrained one (e.g. a plain chat request joining a batch
+    # that is serving a structured json_schema request).  Per-row normalisation
+    # at this chokepoint is the only place that covers both insert and merge.
+    # See #934 / #1747.
     if self.logits_processors is None:
         self.logits_processors = []
+    else:
+        self.logits_processors = [
+            procs if procs is not None else [] for procs in self.logits_processors
+        ]
 
     result = _original_generation_batch_step(self)
 
--- a/tests/test_scheduler_logits_processors.py
+++ b/tests/test_scheduler_logits_processors.py
@@ -12,15 +12,28 @@ doesn't match it, and presents to users as a request hang. See
 ``vllm-mlx-patched`` commit ``8d4052b`` for the same root cause in a
 sibling project.
 
-Two levels of defense:
+The caller-side wrap is necessary but **not sufficient**: on a heterogeneous
+continuous-batch merge, mlx-lm's ``GenerationBatch.extend()`` re-introduces
+None slots via ``if not any(self.logits_processors): self.logits_processors =
+[None] * len(self.uids)``. Because ``any([[], []])`` is False, the empty-list
+slots written at insert time collapse back to None whenever a batch with no
+*active* processor merges with a grammar-constrained one (a plain chat request
+joining a batch that is already serving a structured ``json_schema`` request).
+The crash then fires from the merge path, not the insert path, and only
+reproduces under request concurrency.
 
-1. **Caller-side**: ``omlx/scheduler.py`` always wraps
-   ``logits_processors`` as a list (possibly empty), never None.
-2. **Pattern matcher**: ``CACHE_CORRUPTION_PATTERNS`` includes
+Three levels of defense:
+
+1. **Chokepoint**: ``_patched_generation_batch_step`` normalises the whole
+   list AND every per-row slot to ``[]`` before each step, covering both the
+   insert and the ``extend()`` merge origins. This is the load-bearing guard.
+2. **Caller-side**: ``omlx/scheduler.py`` always wraps ``logits_processors``
+   as a list (possibly empty), never None, at the insert call site.
+3. **Pattern matcher**: ``CACHE_CORRUPTION_PATTERNS`` includes
    ``"'NoneType' object is not iterable"`` so the scheduler recovers
    gracefully if a None slot ever sneaks through.
 
-These tests pin both invariants.
+These tests pin all three invariants.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for tracking this down. I verified the root cause against the pinned mlx-lm path: empty per-row processor lists can be turned back into None during prompt-batch merging and then reach GenerationBatch._step, so normalizing per-row slots at the _step chokepoint is the right fix. The targeted non-integration tests pass locally. This looks good to me, and I am going to merge it.

PR #1801, opened 2026-06-10
fix(vlm): fix Qwen3OmniMoeProcessor loading

monroewilliams:
## Summary

Fixes loading of the `qwen3_omni_moe` architecture in omlx so it successfully loads as a VLM. The model was failing during VLM loading due to a `feature_extractor` resolution error in `AutoProcessor.from_pretrained()`, and the processor's `__init__` was incorrectly passing arguments positionally to `ProcessorMixin.__init__`, causing a type mismatch.

## Changes

- **`_patch_qwen3_omni_moe_feature_extractor()`**: Intercepts `AutoProcessor.from_pretrained()` to catch `feature_extractor` dict resolution errors and fall back to manual processor construction.
- **`_build_qwen3_omni_moe_processor()`**: Manually constructs a `Qwen3OmniMoeProcessor` by reading `processor_config.json` and instantiating each sub-processor directly:
  - **Feature extractor**: Resolves to `WhisperFeatureExtractor` from config.
  - **Image processor**: Handles `Qwen2VLImageProcessor`/`Fast` with PIL fallback (`Qwen2VLImageProcessorPil`) when `torchvision` is unavailable.
  - **Video processor**: Attempts to load `Qwen2VLVideoProcessor`, falls back to a minimal `_MockVideoProcessor` (inherits from `BaseVideoProcessor` dummy) when `torchvision` is missing.
  - **Tokenizer**: Loads via `AutoTokenizer`.
  - **Chat template**: Loads via `mlx_vlm.models.base.load_chat_template`.
  - **Processor construction**: Uses `Qwen3OmniMoeProcessor.__new__()` + direct `ProcessorMixin.__init__()` with keyword arguments to bypass the broken positional argument handling in the library's `__init__`.
- **`_FEATURE_EXTRACTOR_MAP`**: Added `WhisperFeatureExtractor` entry.
- **`_load_vlm_sync()`**: Wired the patch into the VLM load path.

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -384,23 +384,20 @@ def _build_qwen3_omni_moe_processor(model_path, **kwargs):
 
     # If video_processor is still None but Qwen3OmniMoeProcessor requires it,
     # create a minimal mock so ProcessorMixin validation passes.
-    # Must inherit from the dummy BaseVideoProcessor (get_possibly_dynamic_module
-    # returns a dummy when torchvision is unavailable). We override __init__
-    # to skip the dummy's requires_backends check.
+    # get_possibly_dynamic_module("BaseVideoProcessor") returns a DummyObject
+    # (not a real class) when torchvision is unavailable, so we can't inherit
+    # from it.  Create a plain object-based mock instead.
     if video_processor is None:
-        from transformers.processing_utils import ProcessorMixin
-
-        _mock_base = ProcessorMixin.get_possibly_dynamic_module("BaseVideoProcessor")
-
-        class _MockVideoProcessor(_mock_base):
+        class _MockVideoProcessor:
             """Minimal mock video processor for torchvision-free environments."""
 
             model_input_names = ["video_grid_thw"]
+            temporal_patch_size = 2
+            merge_size = 14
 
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for working on this. I may be missing a different test target, but the Qwen3-Omni file I checked is this official `preprocessor_config.json`:

https://huggingface.co/Qwen/Qwen3-Omni-30B-A3B-Instruct/raw/main/preprocessor_config.json

That file has a flat layout with `image_processor_type` at the top level, not a nested `image_processor` section. With that layout, I think this builder falls through to `AutoImageProcessor`, which is a dummy in oMLX's torchvision-free environment, so `image_processor` stays `None` and `ProcessorMixin` fails during construction.

Did you test against a different processor config layout, or is there another path where this works correctly? If I misread the target config, please point me at the file you used.

monroewilliams:
Sorry,I linked my test case in https://github.com/jundot/omlx/issues/1802 :

https://huggingface.co/mlx-community/Qwen3-Omni-30B-A3B-Instruct-4bit

I also tried this one, with the same result:

https://huggingface.co/mlx-community/Qwen3-Omni-30B-A3B-Instruct-8bit

maybe someone is doing bad MLX  conversions?

PR #1807, opened 2026-06-10
fix: strip superseded rotating tip payloads to stop hot-cache exhaustion

### LINKED ISSUE #1795: Sliding-window models: per-turn rotating real-state blocks (~205-410MB fp16 each) exhaust the hot cache → LRU breaks the prefix chain → multi-turn cache hit collapses to 0% after ~10-20 turns
## Describe the bug

On sliding-window models (Gemma 4 26B: 25 of 30 layers are `RotatingKVCache`), **every store of a growing conversation appends one "real-state" block carrying the full sliding-window state of all rotating layers** (~205-410 MB fp16 each at window 1024 / prefill_step 1024). These blocks accumulate in the hot cache one per turn, so a 4 GB hot cache fills after ~10-20 turns. LRU eviction then removes blocks from the **front** of the prefix chain, and reconstruction fails:

```
Partial cache reconstruction: 2/23 blocks
```

→ placeholder-only chain → full reject → **multi-turn cache hit collapses to 0% permanently** (every subsequent turn is a full cold re-prefill). On our Gemma 4 26B setup the collapse reproducibly hits at ~21K conversation tokens.

The kicker: only the **newest** rotating real-state block is ever used on restore — walk-back picks the latest valid boundary — so all older tips' rotating payloads are dead weight that evicts live, needed blocks.

Verified present in v0.4.3 and still on `main` (0.4.4.dev1): the rotating-layer store path clones the full boundary buffer with no trimming, and superseded tips are never slimmed.

## To Reproduce

1. Load a sliding-window model (Gemma 4 26B) with hot cache (we run `hot_cache_only`, 4 GB).
2. Run a growing multi-turn conversation (~2K new tokens per turn, full history re-sent each turn).
3. Watch `cached_tokens`: high hits (95%+) for the first ~10-15 turns, then partial reconstructions, then permanent 0% around 20K+ tokens.

## Root cause

1. The store path's rotating-layer branch clones the **entire boundary buffer** — up to `max_size + chunk − 1` tokens (~2047 at prefill_step 1024), roughly double the window — for each of the 25 rotating layers, into the last ("real-state") block of every store.
2. Each turn of a growing conversation produces a new tip block, so these heavy blocks accumulate **one per turn**.
3. Restore only ever consumes the newest one (walk-back to the latest valid boundary). The older ones are unreachable dead weight.
4. Hot cache LRU has no notion of chain structure: under pressure it evicts the oldest blocks, which are the **head** of the prefix chain — precisely the blocks every future fetch needs.

## Suggested fix (two complementary parts, both verified locally)

**(a) Store-time trim — lossless.** Reconstruct already trims the boundary buffer back to the `keep + window` slice; doing the same trim *at store time* halves the payload (~410 → ~205 MB fp16 across 25 layers) with zero information loss. Only applied when the buffer is in temporal order (`_idx == seq_len`, the `_update_concat` invariant), mirroring the reconstruct-side contract.

**(b) Supersede-on-extend strip.** When a new tip block is saved successfully, replace the rotating payload of the **grandparent** tip (two generations back) with placeholders, keeping the immediate previous tip as a walk-back fallback (a turn whose own tip lands placeholder-only — a periodic block-boundary quirk — then degrades to a partial hit instead of a full cold re-prefill). Steady state: two heavy blocks per chain instead of N.

One implementation note for (b): warm-restored caches are stored under their live class name (`PrefillReadyRotatingKVCache`), so the type check must match the `RotatingKVCache` family by substring, not equality — otherwise the strip silently never fires on warm turns.

## Verification

v0.4.3 + both changes, Gemma 4 26B, growing multi-turn sweeps:

| Config | Before | After |
|---|---|---|
| 4-bit weights | 0% collapse at ~21K | **67K tokens @ 97-98% hit**, zero collapses |
| 3-bit weights + 30 GB wired limit | — | **133K tokens @ 98-99% hit** |

After the fix the multi-turn limit is no longer the cache at all — it's the prefill memory safety cap (graceful rejection).

## Desktop

- OS: macOS 26 (Apple Silicon, Mac Studio 36GB)
- oMLX: 0.4.3 (Homebrew), code path unchanged on `main` @ 0.4.4.dev1
- Model: `gemma-4-26b-a4b-it` UD-MLX 3-bit/4-bit, TurboQuant KV 4-bit (rotating layers stay fp16 via `_turboquant_eligible`), hot_cache_only 4 GB

Happy to open a PR — (a) is small and self-contained; (b) is more design-opinionated, so guidance on preferred approach welcome.

hojin12312:
Fixes #1795.

Follows the shape requested in https://github.com/jundot/omlx/issues/1795#issuecomment — a focused version of the supersede-on-extend stripping approach:

## What it does

`store_cache` now tracks tip lineage per growing chain (newest tip hash → previous tip hash). After a new tip block is saved successfully, the tip **two generations back** has its rotating payload stripped down to the standard last-block-only placeholder. The **immediate previous tip is kept intact** as the walk-back fallback, so a turn whose own tip lands placeholder-only (periodic block-boundary quirk) degrades to a partial hit instead of a full cold re-prefill. Steady state: two heavy real-state blocks per chain instead of one per turn.

## Details, per the review notes on the issue

- **RotatingKVCache family detection**: added `CacheTypeRegistry.is_rotating_family()` and matched through the registry's class-name map rather than exact class-name equality, so warm-restored caches serialized as `PrefillReadyRotatingKVCache` (and batch variants) are covered.
- **Byte-counter consistency**: the strip rewrites the block through `forget_block()` + `save_block()`, so the hot-cache and shared-budget counters are updated through the normal `_hot_cache_remove` (budget `forget`) / `_hot_cache_put` (budget `put`) paths. A regression test asserts `_hot_cache_total_bytes` equals the recomputed sum of live entry sizes, and another asserts `SharedHotCacheBudget.total_bytes` matches the owner's counter after a strip.
- **Sliceable layers untouched**: only rotating-family layer states are replaced; KVCache/TurboQuant slices in the stripped block survive, and the stripped layer satisfies `_is_placeholder_state`, so restore handles it natively via the existing walk-back/reject path.
- **Scope gating**: lineage tracking only engages when the chain's `layer_cache_types` contains a rotating-family type, so pure-KVCache models never pay the load/rewrite cost (regression-tested).
- The lineage map is bounded (4096 entries, two 32-byte hashes each) purely as a guard for long-lived processes serving many distinct chains.

## Tests

`tests/test_prefix_cache_rotating_tip_strip.py` (7 new tests): grandparent stripped / previous tip kept across 4 turns, sliceable-layer survival + placeholder compatibility, hot-cache byte-counter exactness, shared-budget consistency, KVCache-only no-op, `PrefillReadyRotatingKVCache` name coverage, registry family matching.

Existing related suites pass unchanged (341 tests): `test_prefix_cache.py`, `test_prefix_cache_v4_block_storage.py`, `test_hot_cache.py`, `test_paged_ssd_cache.py`, `test_paged_cache.py`, `test_rotating_cache_contract.py`, `test_store_cache_gate.py`, `test_prefix_divergence_probe.py`.

## Field results

Running this (as a local patch on 0.4.3) on a gemma-3-27b-class sliding-window model, M-series 36GB, TurboQuant KV 4-bit, 4GB hot cache: multi-turn conversations that previously collapsed to a permanent 0% hit around ~21K tokens now sustain 97-99% hit up to the prefill safety cap (67K-133K depending on weights/ceiling), with no collapse over 150 turns.

I left the store-time trim (the other option discussed on the issue) out of this PR as requested; happy to follow up separately if useful.

### DIFF
--- a/omlx/cache/prefix_cache.py
+++ b/omlx/cache/prefix_cache.py
@@ -35,6 +35,11 @@ from .type_registry import CacheTypeRegistry
 
 logger = logging.getLogger(__name__)
 
+# Cap on the supersede-on-extend lineage map (tip hash -> previous tip hash).
+# Each entry is two 32-byte hashes; the cap only guards against unbounded
+# growth from many distinct conversation chains over a long-lived process.
+_TIP_LINEAGE_MAX_ENTRIES = 4096
+
 
 @dataclass
 class BlockCacheEntry:
@@ -106,6 +111,13 @@ class BlockAwarePrefixCache(CacheManager):
 
         # Request to block table mapping
         self._request_tables: dict[str, BlockCacheEntry] = {}
+
+        # Supersede-on-extend lineage for rotating (sliding-window) models:
+        # newest tip block hash -> previous tip block hash. When a chain is
+        # extended again, the entry two generations back is stripped of its
+        # rotating payload (see _strip_rotating_payload); the immediate
+        # previous tip is kept intact as the walk-back fallback.
+        self._rotating_tip_lineage: dict[bytes, bytes] = {}
 
         # Callback for restoring cold blocks (deprecated in paged SSD-only mode)
         # Kept for API compatibility
@@ -460,6 +472,9 @@ class BlockAwarePrefixCache(CacheManager):
             )
 
         blocks_saved_to_ssd = 0
+        # Supersede-on-extend tracking (rotating models only, see below).
+        first_new_block_idx: int | None = None
+        tip_block_saved = False
 
         for i in range(num_new_blocks):
             start_idx = i * self.block_size
@@ -500,6 +515,8 @@ class BlockAwarePrefixCache(CacheManager):
                     continue
 
             # Allocate new block
+            if first_new_block_idx is None:
+                first_new_block_idx = len(block_table.block_ids)
             block = self.paged_cache.allocate_block()
             if not block:
                 # Handle memory pressure
@@ -639,6 +656,8 @@ class BlockAwarePrefixCache(CacheManager):
                     )
                     if saved:
                         blocks_saved_to_ssd += 1
+                        if is_last_block:
+                            tip_block_saved = True
                         logger.debug(
                             f"Saved block {block.block_id} to tiered cache: "
                             f"tokens [{global_start}:{global_end}], {len(block_kv_data)} layers"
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for putting this together. This matches the supersede-on-extend direction I wanted for #1795, and the core behavior looks good to me.

I applied the patch locally and ran the related cache suites successfully. I found one follow-up edge where the new rotating-family detection should also be used consistently in the store/restore paths for `PrefillReadyRotatingKVCache` metadata, not only in the strip path.

That does not change my read of the main approach in this PR, so I am going to merge this and fold that edge fix into a maintainer follow-up.

PR #1812, opened 2026-06-11
feat(thinking): add guided soft thinking budgets

efortin:
## Summary

This PR keeps `thinking_budget` as a hard safety wall, but gives the model a softer path to finish before it hits that wall.

- A request budget adds a short tail-system hint, so the model knows to keep reasoning under the requested budget. This is the main control signal, and placing it at the prompt tail keeps volatile budget text away from the stable prefix.
- In the last 30% of the budget, the logits processor nudges only the next valid close-thinking token with a late sigmoid ramp.
- The soft target is capped below the current winner: it pulls toward `top_logit - 0.25`, and never lowers the close token if it was already better. This avoids turning the soft zone into a temperature-0 forced close.
- The nudge stays in lazy MLX ops, with no `.item()` sync in the decode loop.
- Multi-token close markers are handled one token at a time, so marker fragments are not boosted out of order.
- If the model still does not close naturally, the existing hard force closes thinking at 100% of the budget.

ig_0751637ddbc5cb82016a3067e101248191b2b3e2c62103832e

## Validation

- `uv run --active --no-sync python -m pytest tests/test_structured_output.py tests/test_thinking_budget.py tests/test_api_utils.py::TestPrepareSystemMessagesForTemplate -q` -> 99 passed
- `uv run --active --no-sync python -m compileall -q omlx/api/thinking.py tests/test_thinking_budget.py`
- `git diff --check`

Real oMLX, strict `json_schema`, reasoning parser enabled, two structured prompts, temperatures `0` and `0.3`, budgets `50/100/300`.

The table below uses true internal processor events from `BENCH_THINK_DONE`. `hard cut` means the hard wall forced the close token at the exact internal budget; `natural close` means the model emitted the close token before the wall.

| model | JSON | recall | budget 50 internal events | budget 100 internal events | budget 300 internal events |
|---|---:|---:|---|---|---|
| Qwen3.6-35B-A3B-nvfp4 | 12/12 | 42/42 | natural close `1/4 @ 49`; hard cut `3/4 @ 50` | natural close `2/4 @ 73`; natural close `1/4 @ 99`; hard cut `1/4 @ 100` | natural close `2/4 @ 298`; hard cut `2/4 @ 300` |
| gemma-4-31b-it-nvfp4 | 12/12 | 41/42 | natural close `2/4 @ 4`; natural close `1/4 @ 49`; hard cut `1/4 @ 50` | natural close `1/4 @ 96`; natural close `1/4 @ 97`; hard cut `2/4 @ 100` | natural close `1/4 @ 100`; natural close `1/4 @ 101`; hard cut `2/4 @ 300` |

## References

- [Token-Budget-Aware LLM Reasoning](https://arxiv.org/abs/2412.18547) motivates the prompt hint: the paper shows that CoT reasoning can be compressed by including a reasonable token budget in the prompt.
- [Steering LLM Thinking with Budget Guidance](https://arxiv.org/abs/2506.13752) motivates keeping decode-time control soft: it steers reasoning with token-level guidance instead of relying only on abrupt budget forcing.
- https://github.com/vllm-project/vllm/pull/45727

### DIFF
--- a/omlx/api/thinking.py
+++ b/omlx/api/thinking.py
@@ -402,11 +402,11 @@ class ThinkingBudgetProcessor:
     # Multiplier on the soft-zone ramp; 2.0 can reach the cap halfway through
     # the zone when the close-think token is already very near the top.
     _SOFT_BIAS_FACTOR = 2.0
-    # Only nudge the close-think token if it is already in the candidate set.
-    # This keeps the bias model-driven while avoiding a full vocabulary sort.
+    # Use the top-k values to estimate the local candidate span without a full
+    # vocabulary sort.
     _SOFT_BIAS_TOP_K = 32
-    # Once eligible, allow a small push above the top logit, but cap it so
-    # the soft zone remains a nudge instead of a deterministic force.
+    # Allow a small push above the top logit, but cap it so the soft zone
+    # remains a nudge instead of a deterministic force.
     _SOFT_BIAS_MAX_OVERSHOOT = 1.0
 
     def __init__(
@@ -490,14 +490,14 @@ class ThinkingBudgetProcessor:
     def _apply_soft_bias(self, logits, mx):
         """Progressively boost the close-think logit through the soft zone.
 
-        At each step, look at the model's top-k candidates. If the next valid
-        close-think token is in that set, raise it toward the top as the
-        budget runs out. The closer it is to the top within the top-k span,
-        the stronger the ramp.
+        At each step, look at the model's top-k candidates to estimate the
+        local logit span, then raise the next valid close-think token toward
+        the top as the budget runs out. The closer it is to the top relative
+        to that span, the stronger the ramp.
 
-        Tokens outside the top-k set are left unchanged. Eligible tokens can
-        overtake the top by a small fixed margin only, so the soft zone remains
-        a nudge instead of a deterministic force.
+        Far-away close tokens get a weaker boost. The target can overtake
+        the top by a small fixed margin only, so the soft zone remains a nudge
+        instead of a deterministic force.
 
         The whole path stays in lazy MLX array ops — no ``.item()``/eval,
         so the decode loop never syncs on this bias (``progress`` comes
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for the PR and the detailed tests. The idea looks useful, but I want to hold off on merging until two issues are addressed.

First, the soft-budget path calls `.item()` several times on every token in the soft zone. In MLX that can force GPU results back to the CPU and slow down decoding. Can you keep this path in MLX array operations instead?

Second, multi-token close markers look unsafe here. If the close marker is `[A, B]`, this boosts both `A` and `B` at the same time, so the model could sample `B` first. That would not close thinking, but it could leak a close-marker fragment into the thinking text. I think the bias should only target the valid next close token, or track the partial close sequence explicitly.

efortin:
Updated after the review.

The soft-budget path now stays in lazy MLX array operations. I rechecked the runtime path and there are no `.item()` / eval calls left in the decode-loop bias path.

Multi-token close markers are also handled one token at a time: the bias only targets the next valid close-marker token.

vichzp:
Some measurements of the hard-wall behaviour this PR layers on top of, on 0.6.2 (i.e. without this PR). Single-turn English prompt, no tools, n=10 per cell, 270 generations total.

**The wall itself is exact.** Qwen3.6-35B-A3B and gemma-4-26B-A4B, five budgets each, MTP on and off: zero over-budget in every cell. Both marker paths enforce identically (`</think>` via the tokenizer attribute; `<channel|>` via the gemma-4 output parser).

**What happens after the forced close is model-dependent.** Comparing at matched cut depth — budget as a fraction of each model's unconstrained thinking length for this prompt (Qwen ~7,100 tokens, gemma ~2,500):

| budget / natural | Qwen 3.6 | gemma-4 |
|---|---|---|
| ~5% | 6/10 | 0/10 |
| ~7% | 4/10 | 1/10 |
| ~11% | 3/10 | 0/10 |

Pooled: **13/30 vs 1/30, Fisher p = 0.0004.** Above ~14% of natural both models are clean (Qwen 0/20, gemma 2/50).

Escape = reasoning continues in `content` after the close: a leaked close marker, or `finish_reason=length` with no answer emitted. Qwen resumes deliberating in `content`; gemma abandons the truncated fragment and writes an answer.

Two things this suggests:

- Severity scales with how far the budget sits below what the prompt actually needs, not with the absolute budget.
- The failure is a model's behavioural response to being truncated mid-thought, not a decode-time defect — so a logits-side fix can't generalise across models. The prompt hint is the part that can.

On that last point: separately, injecting reasoning-effort steering text into the system prompt cut Qwen's unconstrained thinking from ~7,100 to ~4,300 (p ≈ 0.02). Consistent with the Token-Budget-Aware result cited above, and with the hint being the main control signal.

vichzp:
Correction to the last paragraph of my previous comment — I retested it with a proper control and the attribution was wrong.

Same model and prompt, thinking budget disabled, n=10 per arm:

| system message | mean thinking tokens |
|---|---|
| none | 7,087 |
| 130-token neutral filler (deployment boilerplate — nothing about reasoning) | 4,274 |
| 127-token reasoning-effort steering text | 4,204 |

Neutral vs steering: **p = 0.85**. No system message vs either: p = 0.0015.

The reduction comes from a system message being present at all, not from the steering content. I shouldn't have credited it to effort steering.

Worth flagging for this PR's validation: if the tail-system budget hint is benchmarked against a no-hint baseline, some of the measured reduction will be this artifact rather than the hint conveying a budget. A length-matched neutral hint would separate the two.

(The two comparisons differ in strength: neutral-vs-steering was within a single run and is clean; the no-system-message baseline was pooled from two earlier runs, so treat the ~40% magnitude more loosely than the null result.)

PR #1813, opened 2026-06-11
fix(benchmark): don't force LM-only loading when VLM MTP is enabled

### LINKED ISSUE #1811: Benchmark does not run VLM MTP drafter due to force_lm=True
## Description

When running the benchmark with `vlm_mtp_enabled=True`, the VLM MTP speculative decoding drafter is **not actually active** during the benchmark run, even though the feature is detected as "experimental" and the upload is skipped.

## Root Cause

In `omlx/admin/benchmark.py:774`, the benchmark forces LM-only loading:

```python
engine = await engine_pool.get_engine(request.model_id, force_lm=True)
```

This causes VLM models to load as `BatchedEngine` instead of `VLMBatchedEngine`. The VLM MTP drafter attachment requires `engine.set_vlm_mtp_drafter()`, which only exists on `VLMBatchedEngine` (`omlx/engine/vlm.py:1131`). As a result, the drafter is never attached and the MTP speculative decoding path is inactive.

Other speculative features (native MTP, DFlash, SpecPrefill, TurboQuant) work correctly in benchmarks because they integrate into the `BatchedEngine` path.

## Questions

1. Is this the intended behavior? The benchmark currently flags VLM MTP as "experimental" and skips leaderboard upload, but the drafter isn't actually running — so the measured throughput is just baseline without speculative decoding.
2. Would it make sense to support VLM MTP in benchmarks by allowing `VLMBatchedEngine` to load when `vlm_mtp_enabled=True`, or perhaps adding a separate benchmark mode for VLM models?
3. If this is something worth fixing, I'd be happy to submit a PR adjusting the benchmark to properly support VLM MTP drafter evaluation.

## Environment

- Latest `main` branch (`ece9842`)

imi4u36d:
## Summary

When `vlm_mtp_enabled=True`, the benchmark was still using `force_lm=True` which loaded VLM models as `BatchedEngine` instead of `VLMBatchedEngine`. Since `set_vlm_mtp_drafter()` only exists on `VLMBatchedEngine`, the MTP drafter was never attached and the measured throughput was just baseline without speculative decoding.

This fix skips `force_lm=True` when VLM MTP is enabled and a draft model is configured, allowing the VLM engine to load properly with the drafter attached.

## Changes

- `omlx/admin/benchmark.py`: Conditionally set `force_lm=False` when `vlm_mtp_enabled` and `vlm_mtp_draft_model` are both set, so `VLMBatchedEngine` loads and the MTP drafter can be attached via `engine_pool`.

Closes #1811

---

If the current design is intentional (i.e., VLM MTP benchmarks are not meant to be supported), please feel free to close this PR and the issue directly.

### DIFF
--- a/omlx/admin/benchmark.py
+++ b/omlx/admin/benchmark.py
@@ -735,11 +735,12 @@ async def run_benchmark(run: BenchmarkRun, engine_pool: Any) -> None:
         # Snapshot experimental flags at run start. Settings can change mid-run,
         # and the produced numbers are tied to whatever was active when
         # generation actually ran.
+        model_settings = None
         sm = getattr(engine_pool, "_settings_manager", None)
         if sm is not None:
             try:
-                s = sm.get_settings(request.model_id)
-                run.experimental_features.extend(_detect_experimental_features(s))
+                model_settings = sm.get_settings(request.model_id)
+                run.experimental_features.extend(_detect_experimental_features(model_settings))
             except Exception as e:
                 logger.warning(
                     f"Benchmark: failed to read experimental flags for "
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for the clear fix. I checked the benchmark load path and this matches the VLM MTP attachment path: with VLM MTP and a drafter configured, the benchmark now loads VLMBatchedEngine so set_vlm_mtp_drafter() can attach the drafter; other cases keep the existing LM-only benchmark behavior. This looks good to me, and I am going to merge it.

jundot:
For context, keeping the benchmark on mlx-lm by default was intentional. There is a baseline performance difference between mlx-lm and mlx-vlm, so I originally forced LM loading to avoid mixing engine-level differences into benchmark results.

Now that MTP support is expanding, I think it is correct to adjust that behavior for VLM MTP. The remaining caveat is that the default mlx-lm vs mlx-vlm performance gap can still be easy to misread, so I added a follow-up commit with an explicit "Force mlx-lm engine" option for benchmark runs.

PR #1815, opened 2026-06-11
fix(cache): invalidate SSD blocks with stale layer cache signature

cfbraun:
## Summary

The SSD prefix cache accumulates blocks across runs that share a model name but differ in cache-layer configuration (TurboQuant toggle, MTP toggle, dtype change). Each block records its `layer_cache_types` in metadata, but the per-block compatibility check in `BlockAwarePrefixCache.reconstruct_cache` bootstraps the comparison reference from **block 0's metadata**. A stale block 0 then anchors every subsequent comparison — and every newer, correctly-typed block trips the mismatch warning, capping reuse at the length of the leftover prefix forever.

I noticed this on my own setup running Qwen3.6-35B-A3B with TurboQuant: every request reused exactly 12 blocks (24,576 tokens) of system-prompt prefix and rebuilt the rest, regardless of how long the conversation actually was. Log forensic was textbook:

```
Cache layer type mismatch at block 32: got ['SizedArraysCache', ..., 'TurboQuantKVCache', ...],
expected ['ArraysCache', ..., 'KVCache', ...]. Truncating cached prefix before this block.
Partial cache reconstruction: 12/N blocks, 24576 tokens
```

The blocks-saved-before-TurboQuant-was-toggled became the reference and refused to retire. Toggling TurboQuant per-model just multiplies the failure modes — every config change strands a tier of blocks.

## Fix

Three coordinated changes:

1. **`PagedSSDCacheManager` gains `adopt_layer_signature_if_unset` and `invalidate_stale_layer_signature`.** The first save after a model load establishes the live signature (post-TurboQuant / post-MTP); the sweep drops index entries whose canonicalized `layer_cache_types` disagree. **Scoped to `_expected_model_name`** so other models sharing the SSD directory keep their blocks untouched — the existing per-model partition already exists in `_is_compatible_block`, the sweep just respects the same boundary. Sweep is idempotent (one-shot guarded by `_signature_sweep_completed`) and a no-op without a signature or owning model name.

2. **`BlockAwarePrefixCache.reconstruct_cache`** anchors the per-block comparison to the manager's expected signature when available, so a stale block 0 fails its own check and is forgotten on the first request instead of poisoning the reference for the lifetime of the prefix.

3. **`BlockAwarePrefixCache.set_paged_ssd_cache_manager`** triggers the one-shot sweep on attach so prefix caches adopting a manager with a pre-set signature get cleaned up before the first prefix lookup.

The canonicalization helper lifts to a module-level `_canonicalize_layer_cache_types` in `paged_ssd_cache` so the manager and the prefix cache compare via the same normalization. `SizedArraysCache`/`ArraysCache` still collapse; `TurboQuantKVCache` and `KVCache` deliberately do **not** — they store different tensors (4-bit packed vs fp16), and collapsing would silently mix incompatible blocks into the same attention pass.

## Multi-model safety

The sweep iterates `_index._index`, not the filesystem. `_scan_existing_cache` already filters blocks by `_expected_model_name` before they enter the index, so blocks for other loaded models are invisible to this manager and can't be touched. Switching to a different model constructs a fresh manager that indexes its own slice. Blocks with no `model_name` (legacy/unattributable) are also skipped — there's no safe way to scope them.

## Test plan

- [x] `tests/test_paged_ssd_cache.py::TestLayerSignatureSweep` — 12 cases: adoption sets/keeps signature, sweep drops mismatched, sweep canonicalizes `SizedArraysCache`, sweep leaves other models' blocks alone, sweep skips legacy/unnamed and pre-signature blocks, idempotent, no-op without signature/model_name
- [x] `tests/test_prefix_cache.py::TestSetPagedSSDCacheManagerTriggersSweep` — attach triggers `invalidate_stale_layer_signature`, attach survives sweep exception, detach (`None`) doesn't call
- [x] `tests/test_prefix_cache.py::TestCanonicalLayerCacheTypes` — pins the `TurboQuantKVCache != KVCache` boundary so future canonicalizer edits can't silently collapse it
- [x] Full regression on `test_paged_ssd_cache.py`, `test_prefix_cache.py`, `test_paged_cache.py`, `test_cache_*`, `test_prefix_cache_*`, `test_engine_preflight.py`, `test_process_memory_enforcer.py` (566 passed)
- [ ] Real-world: clear `~/.omlx/cache/` once, run with TurboQuant ON for a while, toggle OFF and confirm subsequent run on the same model self-heals (sweep drops the TurboQuant'd blocks)


jundot:
Thanks for the fix. The underlying issue is real: SSD blocks written with one layer-cache layout can survive restarts and later be considered for prefix reconstruction under a different layout, which can corrupt reuse.

I merged this because persisting and checking the layer cache signature is the right fix direction. While testing it, I found one first-lookup edge where the SSD manager may not know the live post-patch cache layout until model startup finishes, so I’m adding a follow-up on main to refresh the signature before prefix lookup and to cover the startup scan path with actual flushed SSD block files.

PR #1821, opened 2026-06-11
fix(api): harden /v1/completions thinking_budget (validation, stream parity, tokenizer detection) on top of #1844

efortin:
## Summary

Follow-up to #1844. That PR already forwards `thinking_budget` on `/v1/completions`; this one hardens the remaining edges around that path.

Raw completions can use a thinking budget when the prompt already opens a thinking block, for example by ending with `<think>\n`. The remaining issues were: negative budgets were accepted, streaming completions could leak the scheduler's synthetic `<think>\n` opener, and the stream strip guard needed to mirror the scheduler's token-level detection instead of relying on raw text.

## Changes

- Reject negative `thinking_budget` values on chat and completions.
- Strip the synthetic `<think>\n` opener from streamed raw completions so streaming and non-streaming responses match.
- Detect open thinking prompts from tokenizer ids, reusing the prompt ids already computed for context validation.
- Keep the detector aligned with scheduler behavior, including disabled-thinking prompts and multi-token `</think>` markers.

## Tests

Adds coverage for completions parsing/forwarding, negative-budget validation, stream prefix stripping, tokenizer-backed prompt detection, prompt-id reuse, disabled-thinking patterns, and the structural wiring guards for both completion paths.

Validation: targeted thinking-budget tests pass; broader `completion or thinking or budget` selection was re-run during review.

### DIFF
--- a/omlx/api/openai_models.py
+++ b/omlx/api/openai_models.py
@@ -299,7 +299,7 @@ class ChatCompletionRequest(BaseModel):
     # Chat template kwargs (e.g. enable_thinking, reasoning_effort)
     chat_template_kwargs: Optional[Dict[str, Any]] = None
     # Thinking budget (max thinking tokens, None = unlimited)
-    thinking_budget: Optional[int] = None
+    thinking_budget: Optional[int] = Field(default=None, ge=0)
     # SpecPrefill: per-request enable/disable (None = use model setting)
     specprefill: Optional[bool] = None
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for the follow-up. I checked the completions wiring and the stream/non-stream parity path against the scheduler's think-prefix detection, and the tokenizer-backed guard looks correct. The targeted thinking-budget tests and the broader completion/thinking/budget selection pass locally. This looks good to me, and I'm going to merge it.

PR #1824, opened 2026-06-11
fix(scheduler): realign per-row samplers and logits processors by uid

### LINKED ISSUE #1823: Grammar and thinking_budget silently stop applying under concurrent mixed load (row misalignment after batch merge — sequel to #1798)
## Environment

- omlx main @ `6ae5142` (includes the #1799 fix), mlx-lm at the pinned `39c4019`
- macOS 26.5, Apple Silicon (M5 Pro, 64 GB)
- Model: `mlx-community/Qwen3.6-35B-A3B-nvfp4`, thinking enabled
- Default scheduler settings

## Symptom

This is the sequel to #1798. The crash is gone since #1799, but the same trigger now corrupts silently.

As soon as a plain chat completion and a constrained request decode concurrently in the same batch:

- `response_format: json_schema` (strict) responses come back as **plain prose** - the grammar is never applied. 5/5 in my repro.
- `thinking_budget` is ignored — the model reasons to its natural length. 10/10 in my repro (603 thinking tokens with `thinking_budget: 300`, every single time).

No error, no log, the requests complete "successfully". Solo, both features work perfectly (299/300, valid JSON, deterministic at temperature 0).

I originally chased this as a flaky "the first budgeted request after a model load doesn't enforce" - it was actually the overlap with the previous request still draining in the batch. Two threads 0.3s apart make it 100% deterministic.

## Root cause

I put a composition probe on the step chokepoint and on `GenerationBatch.extend`. The smoking gun:

```
PROBE GenerationBatch.extend self=[0, 0] + in=[2] -> [0, 0, 2]  uids=[1, 2]
PROBE step composition: [(1, 0), (2, 0)]
```

Three processor slots for two uids. A stale slot left behind by a finished request shifts everything after it: the structured request (uid 2) reads slot 1 (empty), while its grammar + budget processors sit in slot 2, which nothing ever reads. The #1799 normalisation makes the step crash-safe on `None` slots, but by design it can't restore alignment - the row just runs without its processors.

`samplers` go through the exact same positional logic in `extend()`/`filter()`, so a row can also silently run another request's sampler.

## Repro

```python
import json, threading, time, urllib.request

BASE = "http://localhost:1234/v1/chat/completions"
MODEL = "Qwen3.6-35B-A3B-nvfp4"
SCHEMA = {"type": "object", "additionalProperties": False,
          "required": ["minutes"], "properties": {"minutes": {"type": "integer"}}}

def post(body):
    req = urllib.request.Request(BASE, data=json.dumps(body).encode(),
                                 headers={"Content-Type": "application/json"}, method="POST")
    with urllib.request.urlopen(req, timeout=600) as r:
        return json.loads(r.read())

def plain(i):
    post({"model": MODEL, "messages": [{"role": "user", "content": f"Briefly explain tides, variant {i}."}],
          "temperature": 0, "max_tokens": 700})

def constrained(i):
    d = post({"model": MODEL,
              "messages": [{"role": "user", "content": "A train leaves at 14:07 and arrives at 17:43, stopping 12 minutes total. How many minutes was it moving?"}],
              "temperature": 0, "max_tokens": 4000, "thinking_budget": 300,
              "response_format": {"type": "json_schema",
                                  "json_schema": {"name": "x", "strict": True, "schema": SCHEMA}}})
    msg = d["choices"][0]["message"]
    print(i, "content:", (msg.get("content") or "")[:60],
          "| reasoning chars:", len(msg.get("reasoning_content") or ""))

for i in range(10):
    t1 = threading.Thread(target=plain, args=(i,))
    t2 = threading.Thread(target=constrained, args=(i,))
    t1.start(); time.sleep(0.3); t2.start()
    t1.join(); t2.join()
```

On main: every `constrained` iteration prints prose content and ~2000+ reasoning chars. Expected: `{"minutes": 204}` and a bounded reasoning.

## Fix

I have a fix up: record at insert time what each uid is supposed to run, and realign the positional lists from that registry at the step chokepoint (same spot as #1799). With it the repro goes 15/15 violations → 0/15, solo/sequential behavior unchanged. PR incoming right after this issue.

efortin:
## Summary

Fixes #1823.

This PR fixes a row-alignment bug in continuous batching. `samplers` and `logits_processors` are positional lists that must stay aligned with `uids`; under mixed concurrent load, stale or shifted slots could make a request run another request's constraints, or none at all.

In practice, this could silently disable grammar constraints or thinking budgets.

## Changes

- Record the expected sampler and logits processors for each `(model, uid)` when a row is inserted.
- Realign batch rows at the generation-step chokepoint before decoding.
- Key by model as well as uid to avoid cross-engine uid collisions.
- Release registry rows on completion, abort, recovery, reset, and shutdown.
- Keep the existing #1799 None-slot normalization as fallback behavior.
- Log when a drift is corrected, with rate limiting.

## Tests

Covers the #1823 stale-slot reproduction, cross-model uid collisions, cleanup paths, registry bounds, warning rate limiting, and the existing #1799 normalization case.

Validation: reproduction is red on the base commit and green on this PR; targeted scheduler tests pass.

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -19,7 +19,7 @@ import logging
 import os
 import threading
 import time
-from collections import defaultdict, deque
+from collections import OrderedDict, defaultdict, deque
 from collections.abc import Callable
 from contextlib import contextmanager
 from dataclasses import dataclass, field
@@ -408,6 +408,30 @@ class _PrefillState:
 # After _original_step returns, self._next_tokens holds the freshly sampled
 # tokens.  We eval them synchronously and accept in grammar processors.
 # ---------------------------------------------------------------------------
+# Authoritative per-uid row state for the generation batch.
+#
+# mlx-lm keeps ``samplers`` / ``logits_processors`` as positional lists that
+# must stay aligned with ``uids``.  Heterogeneous continuous batching
+# (extend/filter/split across prompt and generation batches) can leave stale
+# or offset row slots behind; #1799 made the step crash-safe by normalising
+# ``None`` slots, but a misaligned row silently runs the WRONG sampler and
+# logits processors (e.g. a grammar/thinking-budget request decoding with no
+# constraints at all).  The registry below records, at insert time, what each
+# uid is supposed to run; the step chokepoint realigns the positional lists
+# from it.  Bounded so a missing cleanup can never grow it unbounded.
+_UID_ROW_REGISTRY_MAX = 4096
+_uid_row_registry: "OrderedDict[int, tuple[Any, list]]" = OrderedDict()
+
+
+def _register_uid_rows(uids, samplers, lps_rows) -> None:
+    """Record the sampler and logits processors each freshly-inserted uid must run."""
+    for uid, sampler, lps in zip(uids, samplers, lps_rows):
+        _uid_row_registry[uid] = (sampler, list(lps) if lps else [])
+        _uid_row_registry.move_to_end(uid)
+    while len(_uid_row_registry) > _UID_ROW_REGISTRY_MAX:
+        _uid_row_registry.popitem(last=False)
+
+
 _original_generation_batch_step = GenerationBatch._step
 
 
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for chasing this down and for rebasing it on top of #1845. The uid registry approach is the right fix for the remaining BatchGenerator row drift: it covers samplers as well as logits processors, and it corrects the positional lists at the decode chokepoint instead of only patching one reshape path.

One follow-up I'll fold in after merge: native MTP can bypass GenerationBatch._step from its patched next(), so it should call the same row realignment before MTP eligibility and row extraction. That does not block this standard decode fix, and I want to get this row-drift fix in first.

efortin:
This bug was critical for me (lot of batch with structured output on my side).
Thanks a lot, I can move away from my fork :-)

PR #1828, opened 2026-06-11
feat(jang): JANG/JANGTQ mixed-precision MoE engine (hardened; builds on #364)

marzukia:
> **Draft — for review.** Builds on the original JANG implementation by @AlexTzk (#364). Credit to Alex for the `JANGLoader` design, EnginePool integration, and the mixed-precision quantization approach. This has **diverged substantially** from #364 — effectively a rewrite of the load/dispatch path — so it's opened fresh rather than as a revision of that PR.

## What this adds on top of #364

- **Engine-type routing fix** — JANG models were clobbered to `batched` on any `model_type` override (→ "Received N parameters not in model" on reload). Routing is now single-sourced through a `resolve_jang_engine_type` helper used by `engine_pool`, the admin routes, and discovery.
- **JANGTQ (mxtq / TurboQuant) support** — mxtq models dispatch to `load_jangtq_model` (TurboQuantLinear, wired-limit-capped), fixing a load-time Metal OOM on large per-expert bundles (e.g. `MiniMax-M2.7-JANGTQ`: 256 routed experts @ 2-bit).
- **Hardening** — fail-loud on switch-quant errors (no silent dequant-to-fp16 fallback), GPU-memory cleanup on failed load, and trial-bits packing math (handles 2/3/4/6/8-bit).
- **Tests** — detection, engine-type clobber guards, JANGTQ dispatch, and switch-quant metadata correction. Apple-Silicon-only tests are `jang_mlx`-marked and excluded from the CI marker filter so CI stays green.

## Validation

- 42 JANG tests pass on Apple Silicon (including the `jang_mlx` set); the mlx-free subset runs in CI.
- `MiniMax-M2.7-JANGTQ` loads and generates correctly on an M-series host.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,4 +37,4 @@ jobs:
           python -m pip install pytest pytest-asyncio
 
       - name: Run unit tests
-        run: python -m pytest tests/ -m "not slow and not integration"
+        run: python -m pytest tests/ -m "not slow and not integration and not jang_mlx"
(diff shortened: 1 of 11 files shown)


MattBash17:
Any update on the state of this PR? JANG would be an incredibly useful feature

PR #1836, opened 2026-06-12
feat: native video_url support for Qwen3-VL-family VLMs (torch-free)

mdc2122:
## What

Adds native `video_url` content-part support to the OpenAI chat-completions
route for Qwen3-VL-family VLMs (tested end-to-end on Qwen3.6-35B-A3B 8-bit).
Previously, `{"type": "video_url", ...}` parts were silently dropped before
reaching the engine; users had to pre-extract frames client-side and send
them as multiple `image_url` parts, losing the model's real temporal video
path (`video_grid_thw` + video tokens).

```jsonc
{"role": "user", "content": [
  {"type": "text", "text": "Describe this video."},
  {"type": "video_url", "video_url": {"url": "data:video/mp4;base64,..."}}
]}
```

Data URLs, local file paths, and http(s) URLs are accepted (data URLs are
decoded to a temp file with a 100 MiB cap and cleaned up after prefill).

## Why the shim (`omlx/utils/video.py`)

oMLX runs torch-free, and transformers' `Qwen3VLVideoProcessor` is
torchvision-gated — `AutoProcessor.from_pretrained` can't even construct it
without torch, which is why `_patch_video_processor_bug()` strips
`video_processor` today (kept intact by this PR). Unlike images, transformers
ships no PIL-backend video processor, so this PR adds
`TorchFreeQwen3VLVideoProcessor`: a numpy/PIL reimplementation of the
reference preprocessing (smart_resize → rescale/normalize → temporal pairing
→ patchify → `pixel_values_videos` + `video_grid_thw`), attached as
`processor.video_processor` after load for Qwen3-VL-family processors only.
Attachment is fail-soft: if construction raises, serving degrades to current
behavior (warning logged, images unaffected).

`video_metadata` is opt-in (`return_metadata=True`) because mlx-vlm's
processor wrapper merges the video processor's entire output into
`to_mlx(...)` — metadata objects would break MLX array conversion.

## Changes

- `omlx/api/openai_models.py` — `VideoURL` model, `video_url` on `ContentPart`
- `omlx/api/utils.py` — preserve/normalize `video_url` parts through
  multimodal extraction (nested-dict, string, and pydantic forms)
- `omlx/utils/image.py` — `extract_media_from_messages()` returning
  `(text, images, audio, videos)`; existing `extract_images_from_messages()`
  keeps its exact 3-tuple contract as a thin delegation
- `omlx/engine/vlm.py` — video part counting in preflight, `num_videos`
  template formatting via `get_message_json(video=...)`,
  `prepare_inputs(videos=...)`, `video_grid_thw` into the existing
  qwen vision-feature path, temp cleanup, shim attach point
- `omlx/utils/video.py` + `tests/test_video_utils.py` — the shim and its
  exact tensor-math tests (shape/grid, normalization values, patch-order
  trap with black/white temporal halves, odd-frame handling, MLX
  convertibility of default output)

## Testing

- ~30 new unit tests, no torch imports, no model weights needed;
  `pytest tests/ -m "not slow and not integration"` green (5574 → 5592+).
- Live E2E on Qwen3.6-35B-A3B-oQ8-mtp (M3 Ultra): data-URL mp4 of a moving
  blue square → model reports object/color/direction correctly; ~350 video
  prompt tokens for 8 sampled frames; images/tool-calling regression-clean;
  no temp-file leaks.

## Out of scope

Anthropic route, audio, video feature caching (image feature cache bypassed
when videos present), streaming video.

### DIFF
--- a/omlx/api/openai_models.py
+++ b/omlx/api/openai_models.py
@@ -34,6 +34,13 @@ class ImageURL(BaseModel):
     detail: Optional[str] = "auto"  # "low", "high", "auto"
 
 
+class VideoURL(BaseModel):
+    """Video URL or base64 data URI for vision model input."""
+
+    url: str  # "https://..." or "data:video/mp4;base64,..."
+    detail: Optional[str] = "auto"  # "low", "high", "auto"
+
+
 class InputAudio(BaseModel):
     """Audio input data for multimodal models (OpenAI format)."""
 
@@ -62,13 +69,15 @@ class ContentPart(BaseModel):
     Supports:
     - text: Plain text content
     - image_url: Image input for vision models
+    - video_url: Video input for vision models
     - input_audio: Audio input for multimodal audio models
     - file: Document or text input for attachment preprocessing
     """
 
-    type: str  # "text", "image_url", "input_audio", or "file"
+    type: str  # "text", "image_url", "video_url", "input_audio", or "file"
     text: Optional[str] = None
     image_url: Optional[ImageURL] = None
+    video_url: Optional[VideoURL] = None
     input_audio: Optional[InputAudio] = None
     file: Optional[FileContent] = None
 
--- a/omlx/api/utils.py
+++ b/omlx/api/utils.py
@@ -150,9 +150,9 @@ def _extract_text_from_content_list(content: list) -> str:
 
 
 def _extract_multimodal_content_list(content: list) -> list:
-    """Extract text, image, and audio parts from a content array.
+    """Extract text, image, video, and audio parts from a content array.
 
-    Keeps text, image_url, and input_audio items for VLM processing.
+    Keeps text, image_url, video_url, and input_audio items for VLM processing.
     Other content types (tool_use, thinking, refusal, etc.) are dropped.
     """
     parts = []
@@ -178,6 +178,20 @@ def _extract_multimodal_content_list(content: list) -> list:
                         {
                             "type": "image_url",
                             "image_url": {"url": url},
+                        }
+                    )
+            elif item_type == "video_url":
+                video_url_value = item.get("video_url")
+                url = None
+                if isinstance(video_url_value, str):
+                    url = video_url_value
+                elif isinstance(video_url_value, dict):
+                    url = video_url_value.get("url")
+                if url:
+                    parts.append(
+                        {
+                            "type": "video_url",
+                            "video_url": {"url": url},
                         }
                     )
             elif item_type == "input_image":
@@ -1182,9 +1196,9 @@ def extract_multimodal_content(
         if isinstance(content, str):
             processed_messages.append({"role": role, "content": content, **_extra})
         elif isinstance(content, list):
-            # Preserve image_url and input_audio parts for VLM processing
+            # Preserve image_url, video_url, and input_audio parts for VLM processing
             multimodal_parts = _extract_multimodal_content_list(content)
-            multimodal_types = {"image_url", "input_audio"}
+            multimodal_types = {"image_url", "video_url", "input_audio"}
             has_multimodal = any(
                 p.get("type") in multimodal_types for p in multimodal_parts
             )
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -47,9 +47,11 @@ from ..cache.vision_feature_cache import VisionFeatureSSDCache
 from ..exceptions import InvalidRequestError
 from ..models.vlm import VLMModelAdapter
 from ..utils.image import (
+    cleanup_temp_video_paths,
     compute_image_hash,
     compute_per_image_hashes,
     extract_images_from_messages,
+    extract_media_from_messages,
 )
 from .base import (
     BaseEngine,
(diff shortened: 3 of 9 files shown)


chigkim:
Thanks, this is cool!

Is there a way to control fps?

Also does it only process the latest video if there are a few turns with user message that has a video?

For example, let's say I want to compare with a previous video I sent and the latest video.

mdc2122:
Thanks! Both good questions — answers below, and the branch has just been rebased onto current main (post-#1984/#1994 `engine/vlm.py` restructuring), with all touched test suites passing.

**fps control:** Frame sampling happens in `mlx_vlm`'s video loader with the standard Qwen defaults — `fps=2.0`, `min_frames=4`, `max_frames=768` (long videos get uniformly downsampled to the cap). The torch-free processor shim mirrors those defaults and already accepts per-video `fps` metadata internally, but this PR doesn't expose fps in the API request yet. If there's interest, a natural follow-up would be an optional field on the content part, e.g. `{"type": "video_url", "video_url": {"url": ..., "fps": 1.0}}` — happy to add that in a separate PR to keep this one reviewable.

**Multiple videos across turns:** All videos are processed, not just the latest. `extract_media_from_messages()` walks every message in the conversation and collects each `video_url` part in order of appearance; the processor then patchifies each video separately (per-video `video_grid_thw` rows) so the model sees them as distinct inputs. Your compare-two-videos case works — each video does consume its full token budget, so two long videos on a small-context model is the main practical limit.

mdc2122:
"rebased on current main, tests green"

chigkim:
Thanks @mdc2122 for the clarification!
@jundot, I hope this great PR gets merged soon!
Thanks!

Marian2110:
Adding a use case in support of this: I'm running Qwen3.8-27B and Qwen3.6-35B-A3B locally on an M1 Max via oMLX, and both ship the architecture for video — `temporal_patch_size: 2` in `vision_config` and a 3D `mrope_section: [11, 11, 10]` with a temporal axis. The capability is in the weights.

The gap is exactly what this PR closes. `Request` already declares a `videos` field and `engine_core` passes it through, but nothing in the server populates it and there's no frame decoding anywhere in the package — so the field is currently unreachable from the API.

Today the workaround is extracting frames with ffmpeg and sending them as separate images, which loses the temporal structure the models are built to use.

Happy to test this on Apple Silicon with either model if that's useful.

Marian2110:
Applied this on an M1 Max and tested it end to end with real video. **It works** — but
it no longer applies to `0.6.3rc1`, and it does not cover the DFlash engine at all.
Details below in case they save you time.

## It works

Qwen3.6-35B-A3B (VLM/batched path), 3-second 1280x704 h264 clip sent as a
`data:video/mp4;base64` `video_url` part, 394 prompt tokens, 33 s:

> The video showcases a black gaming mouse with RGB lighting resting on a textured,
> dark surface. The clip presents the mouse from **three distinct angles**: a
> front-facing view featuring purple and blue lights, a high-angle side view
> emphasizing cyan strips, and a profile view displaying the scroll wheel and side
> buttons.

Qwen3.8-27B, same clip, 436 prompt tokens, 42 s:

> It opens with a tight head-on shot of the mouse, then the angle shifts (camera
> move/cut) to a top-three-quarter view that reveals the side buttons, scroll wheel,
> and the attached cable; the mouse itself stays in place while the perspective
> around it changes, with a soft bokeh light drift in the background.

Both describe motion across frames rather than a single image, so the temporal path is
genuinely doing something. Normal text generation is unaffected after the patch —
84.8 tok/s on the MoE, 18.1 on the dense model, same as before.

## It no longer applies

Three source files still apply cleanly (`api/openai_models.py`, `api/utils.py`,
`utils/video.py`). Two do not. Merging three-way against the merge base
(`eca4a7c5`) leaves four conflicts, all mechanical:

| where | what drifted |
|---|---|
| `utils/image.py` imports | `logging` was dropped upstream, but `cleanup_temp_video_paths` uses `logger` |
| `utils/image.py` body | collides with `validate_image_data_uri`, added upstream since |
| `utils/image.py` docstring | audio line reworded upstream |
| `vlm.py:2781` | import block gained `mlx_vlm_pixtral_torch_free` and `reasoning_effort` |

One more thing that a mechanical rebase will miss, because it is a signature change
rather than a textual conflict: `_prepare_vision_inputs` gained an `is_partial`
parameter upstream, and the new `_prepare_vision_inputs_impl` in this branch does not
take it. The split therefore leaves `is_partial` unbound in the impl body, and every
request fails with:

```
UnboundLocalError: cannot access local variable 'is_partial' where it is not associated with a value
```

Threading `is_partial` through the wrapper into the impl fixes it.

## DFlashEngine is not covered

`DFlashEngine` bypasses `VLMBatchedEngine`, so on any model configured with
`dflash_enabled` the video request never reaches this code. Three separate things had
to change before the dense model worked here, and they may be worth folding in:

**1. The multimodal fallback does not recognise video.**

```python
_MULTIMODAL_TYPES = frozenset({"image", "image_url", "input_image"})
```

Without `video_url` in that set, `_has_multimodal_content` returns False and the
request stays on the DFlash path instead of switching to the VLM engine.

**2. Token counting runs first, and 400s the request.** `server.py` calls
`engine.count_chat_tokens(...)` before the engine's own fallback check.
`DFlashEngine.count_chat_tokens` templates the raw messages, and the chat template
rejects the unknown part type:

```
400: Chat template error: Unexpected item type in content.
```

Images survive this because the template knows `image_url`; video does not. So fixing
(1) alone changes nothing — the request dies before the fallback can run.
`VLMBatchedEngine.count_chat_tokens` already handles this by calling
`extract_images_from_messages` first; DFlash needs the same.

**3. The extractor consumes the message list.** Doing (2) naively strips the media
parts from the very dicts the caller reuses for the actual request, so the fallback
check then finds nothing and the model answers "no video was provided" with a
74-token prompt. Deep-copying before extraction fixes it.

With all three, the dense model produces the output quoted at the top.

## Environment

M1 Max 64 GB, macOS 26.5.2, `omlx 0.6.3rc1` (Homebrew), targets
`Qwen3.8-27B-oQ4e-fp16-mtp` (DFlash 2 + sparse prefill) and
`Qwen3.6-35B-A3B-oQ4e-fp16-mtp` (native MTP). Both carry the vision tower; until now
we were feeding them `ffmpeg`-extracted frames as images.

Happy to re-test after a rebase, and to hand over the conflict resolutions if useful.

---

*Note: this investigation was carried out with Claude (Anthropic) and the write-up
drafted by the model. Every result above is from an actual run on the machine
described — the patch was applied, the video sent, and the outputs are verbatim.*

ckcfcc:
Is there any progress on this matter?

@mdc2122

Issue #1830, opened 2026-06-11
QAT Gemma 4 model emits call:google:mcp: tool call format that oMLX parser cannot handle — calls silently dropped


aikidovation:
# Bug Report: All Gemma 4 models: tool calls fail on standard /v1/chat/completions — gemma4 parser only active on VLM path

## Environment

- **Hardware:** Apple M4 Max, 64GB unified memory
- **macOS:** Sequoia
- **oMLX version:** 0.4.3
- **Model:** `lmstudio-community/gemma-4-26B-A4B-it-QAT-MLX-4bit`
- UPDATE: **Model:**`mlx-community/gemma-4-26b-a4b-it-4bit` similar behavior
- **Client:** AnythingLLM (Docker, `mintplexlabs/anythingllm:latest`) via Generic OpenAI provider
- **oMLX endpoint:** `http://127.0.0.1:8000/v1`

## Description

When using the `lmstudio-community/gemma-4-26B-A4B-it-QAT-MLX-4bit` model via the OpenAI-compatible API, tool calls are never executed. Instead, the raw tool call markup is either:

1. Intercepted by oMLX, fails to parse, and is silently stripped — with only surrounding text forwarded to the client, or
2. Passed through as raw text in the response body, visible to the user but not executed

This affects all tool types (`create-pdf-file`, `create-text-file`, etc.) and both native tool calling mode and UnTooled (text-based) mode.

UPDATE: 
- I have now tried the non-QAT variant (mlx-community/gemma-4-26b-a4b-it-4bit) and it also fails identically. This confirms the issue is not QAT-specific but affects all Gemma 4 models on the standard chat completions path.
- oMLX logs show: "VLM tool calling enabled: parser=gemma4" appearing only on VLM path requests. Standard chat completions (used by AnythingLLM) never get the gemma4 parser applied, so tool calls are parsed by the generic parser which fails on Gemma 4's call:google:mcp: format.

## Observed Behavior

The model generates tool calls using the following format:

```
<|tool_call>call:google:mcp:text_generation:create-pdf-file{filename: 'output.pdf', content: '...'}
```

oMLX intercepts this output and logs:

```
omlx.api.tool_calling - WARNING - Native tool parser failed 
(ValueError: No function call found in tool call text.) 
and XML fallback could not recover. Dropping match: 
"call:google:mcp:text_generation:create-pdf-file{filename: ..."

omlx.api.tool_calling - WARNING - Tool call markers found but 
parsing failed, stripping markers.
[...]

Issue #1805, opened 2026-06-10
Support diffusiongemma


black-snow:
**Is your feature request related to a problem? Please describe.**
Tried the brand new diffusiongemma but sadly I'm getting 500s:

> 2026-06-10 23:12:36,614 - omlx.server - ERROR - [-] - POST /v1/chat/completions → 500 (unhandled): Model type diffusion_gemma not supported.

**Describe the solution you'd like**
diffusiongemma support

**Describe alternatives you've considered**
* use vLLM

**Additional context**
https://blog.google/innovation-and-ai/technology/developers-tools/diffusion-gemma-faster-text-generation/  
https://huggingface.co/google/diffusiongemma-26B-A4B-it
I tried `diffusiongemma-26B-A4B-it-mxfp4`.

P.S.: I have no idea what those model files look like - does each new model require explicit support/chages in oMLX?

kreeger:
I tried to slap together some real rough support for it on my local fork of oMLX, and got real garbled nonsense back from the model in a simple chat, so I'm betting this one is more involved than just adding support for a new `model_type`.

Tokarak:
Needs an mlx-vlm bump to the newly-released 0.6.3

Abdenasser:
I'm really excited and can't wait much for this one

imi4u36d:
A model with 4 times the speed? It's quite interesting.

kreeger:
Bringing in v0.6.3 of `mlx-vlm` and adding `diffusion_gemma` to `VLM_MODFEL_TYPES` and `DiffusionGemmaForBlockDiffusion` to `VLM_ARCHITECTURES` still results in some gibberish, though `mlx-vlm.generate` works fine. I suspect oMLX's built-in scheduler is fighting with the built-in diffusion logic.

I'm still digging in, but this enhancement might be outside my normal comfort zone!

**Edit**: I tried to extend myself to see if a coding assistant could help, but I'm tapping out. [Here's as far as I got](https://github.com/kreeger/omlx/commit/f43ba387d4ab88572401d89214123ab688292294) with the help of Claude Code but this is as far as I'm taking it.

With this code, it's able to generate and "stream" (in very large chunks) the response, but the formatting is fully broken because the chunks seem to be leaving out certain tokens, the reporting back to oMLX about progress state is fully broken, and the benchmarks report absolutely wack numbers.

[...]

PR #1837, opened 2026-06-12
feat(diffusion): enable tool calling on the serial diffusion lane

scubamount:
## Summary

Tool calling currently returns HTTP 400 on diffusion-lane models (`Tool calling is not supported with diffusion models.`). But tool calling is **prompt-driven plus output parsing** — it doesn't need grammar-constrained decoding. DiffusionGemma's chat template renders tool declarations and emits the same `<|tool_call>call:name{...}<tool_call|>` wire protocol as Gemma 4, which `mlx_vlm.tool_parsers.gemma4` already parses. The only blockers were the explicit request-time rejections and the stream's special-token stripping eating the protocol markers.

This PR makes the rejection **capability-based** instead of lane-based, so:
- DiffusionGemma (and any future diffusion model whose chat template matches a tool parser) gets working OpenAI-format `tool_calls` on `/v1/chat/completions`, `/v1/messages`, and `/v1/responses`, streaming included.
- Diffusion models *without* a matching tool parser still get a clean 400 (now with a more descriptive message).

## Changes

- **`VLMBatchedEngine.supports_tool_calling`** — new property, true when `_inject_tool_calling` matched a parser for the tokenizer's chat template.
- **Request validation** — `_validate_diffusion_request` and the three server-side guards gate on `supports_tool_calling` instead of `is_diffusion_model`.
- **Stream marker preservation** — the diffusion stream's special-token skipping now preserves tool-call markers and protocol markers (`OutputParserFactory.protocol_marker_texts`); they're consumed downstream by the gemma4 output parser session / `parse_tool_calls` / `ToolCallStreamFilter`, so they never leak to clients.
- **`Gemma4OutputParserSession.process_text`** — text-segment entry point for engines that detokenize internally (the diffusion lane emits text segments, not token ids). `is_gemma4_model` now also matches `model_type: diffusion_gemma` so thought-channel → `<think>` conversion works on the lane.
- **`response_format` degrade instead of 400** — `json_object`/`json_schema` on diffusion models now falls back to prompt-injected JSON with the existing #1241 `Warning` header (same path as when xgrammar isn't installed). `structured_outputs` and `guided_grammar` are **still rejected** — logit-mask enforcement has no equivalent in the parallel denoising loop. `grammar_compiler` returns `None` for diffusion models so the fallback engages even on `--with-grammar` installs.
- **Admin sanitizers** — keep `max_tool_result_tokens` for diffusion models (tool calling works now).

## Verification (diffusiongemma-26B-A4B-it-4bit, Apple Silicon, 0.4.4rc1 base)

| Probe | Result |
|---|---|
| Tool call emission | `finish_reason=tool_calls`, `{"name": "get_weather", "arguments": "{\"city\": \"Tokyo\"}"}` |
| Multi-turn tool-result follow-up | clean prose answer, no protocol marker leak |
| Streaming | tool-call delta with parsed arguments, `finish_reason=tool_calls` |
| `response_format: json_object` | 200 + `Warning` header + valid JSON body |
| `structured_outputs` | still 400 |
| Plain inference (no tools) | unaffected |

`pytest tests/` (excluding integration): **5498 passed, 40 skipped**. Existing tests updated where behavior intentionally changed (3 assertions), new tests added for the capability gate.

Fixes the tool-calling portion of #1805; related to the parser-coverage issue in #1830.

### DIFF
--- a/omlx/adapter/gemma4.py
+++ b/omlx/adapter/gemma4.py
@@ -301,6 +301,7 @@ class Gemma4OutputParserSession:
         self._tokenizer = tokenizer
         self._buffer = ""
         self._in_thought = False
+        self._text_mode = False
 
         self._detokenizer = create_streaming_detokenizer(tokenizer, model_path)
         if self._detokenizer is not None:
@@ -449,9 +450,24 @@ class Gemma4OutputParserSession:
             text = self._tokenizer.decode([token_id])
         return self._consume_text(text)
 
+    def process_text(self, text: str) -> OutputParserTokenResult:
+        """Process an already-detokenized text segment.
+
+        Engines that emit text segments instead of token ids (the serial
+        diffusion lane detokenizes inside ``stream_diffusion_generate``)
+        feed their output through this entry point so protocol markers
+        are handled identically to the token-id path.  Switches the
+        session to text mode so ``finalize`` does not flush the unused
+        token detokenizer.
+        """
+        self._text_mode = True
+        if not text:
+            return OutputParserTokenResult(stream_text="", visible_text="")
+        return self._consume_text(text)
+
     def finalize(self) -> OutputParserFinalizeResult:
         text = ""
-        if self._detokenizer is not None:
+        if self._detokenizer is not None and not self._text_mode:
             self._detokenizer.finalize()
             text = self._detokenizer.last_segment
 
--- a/omlx/adapter/output_parser.py
+++ b/omlx/adapter/output_parser.py
@@ -64,6 +64,12 @@ class OutputParserFactory:
     stop_token_ids: set[int] = field(default_factory=set)
     thinking_end_text: str | None = None
     thinking_end_trailing_text: str | None = None
+    # Marker strings that must survive special-token stripping so the
+    # parser session can see them in the text stream.  Engines that strip
+    # special tokens during detokenization (e.g. the serial diffusion
+    # lane) preserve the token ids of these markers and let the parser
+    # session remove them instead.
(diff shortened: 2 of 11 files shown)


scubamount:
Pushed c19655a after running this live for a few hours behind a memory pipeline (hindsight) that exercises the diffusion lane with large tool-call payloads. Two parser robustness fixes:

1. **Degenerate opening tokens.** The parallel denoising loop occasionally drops a token from the tool-call opening — observed `calldone{...}` (missing colon) and `:done{...}` (missing `call`). The gemma4 fallback required a literal `call:` and dropped these. The prefix is now permissive (`(?:call)?:?`) — safe because the text is already delimited by `<|tool_call>`/`<tool_call|>` markers, so it can't misfire on prose.

2. **Long bare values with commas/newlines.** Tool calls whose bare argument values contain markdown (tables, lists — e.g. a long `answer:` argument) defeat the per-pair bare-value quoting regex and the whole match was dropped. Added a key-anchored last-resort step: anchor on the quoted keys, slice everything between a key's colon and the next key as that key's value.

Both reconstructed live payloads now parse; canonical `call:name{...}` formats unchanged. +4 regression tests (178 pass in test_tool_calling.py).

jundot:
Thanks for the detailed fix. I checked the capability gate and the diffusion stream marker preservation path, and the tool-call markers stay available for final parsing while unsupported diffusion models still fail cleanly.

The shared Gemma 4 parser change is scoped to marked tool-call payloads, so I do not see a broad regression risk for other models. The related parser, VLM, server, and admin tests pass locally, and CI is green.

PR #1705, opened 2026-06-06
feat: CLI model alias management and Bearer token auth for admin API

maxness:
## Summary

Adds the ability to manage model aliases via CLI and REST API, along with an authentication fix that allows admin API routes to accept Bearer API tokens (previously they only accepted session cookies).

### 1. CLI `omlx alias` command (`omlx/cli.py`)

New top-level `alias` subcommand with 4 operations:

- **`omlx alias set <model> <alias>`** — Assign an alias to a model (with conflict detection)
- **`omlx alias swap <model_a> <model_b> [--alias-a <new>] [--alias-b <new>]`** — Swap or transfer aliases between two models atomically. If no `--alias-*` flags are provided, auto-detects and swaps the current aliases
- **`omlx alias remove <model>`** — Remove the alias from a model
- **`omlx alias list`** — List all models with their aliases, pinned status, and loaded state

All commands communicate with the running oMLX server via the admin API using Bearer token authentication.

### 2. Admin API auth: `_require_admin_or_bearer` dependency (`omlx/admin/routes.py`)

New dependency that accepts **either**:
- A valid admin session cookie (existing behavior)
- A `Bearer` token matching the configured API key

Applied to these routes so CLI/API consumers can authenticate without browser sessions:
- `GET /admin/api/models` (list_models)
- `PUT /admin/api/models/{model_id}/settings` (update_model_settings)
- `POST /admin/api/models/{model_id}/load` (load_model)
- `POST /admin/api/models/swap-alias` (swap_model_alias)

### 3. Atomic alias swap endpoint (`POST /admin/api/models/swap-alias`)

- **`SwapAliasRequest`** — Pydantic model with `model_a`, `model_b`, `alias_a` (optional), `alias_b` (optional)
- Atomic two-model update with rollback on failure
- Conflict detection: prevents alias collisions with third models and model directory names
- Normalizes empty/whitespace-only aliases to `None`

### 4. Alias conflict detection in `update_model_settings`

When setting `model_alias` via the settings update endpoint:
- Checks for conflicts with existing aliases on other models (400 error)
- Checks for conflicts with other models' directory names (400 error)

### 5. Tests

- **`tests/test_admin_alias_api.py`** (new) — 36 tests covering:
  - `SwapAliasRequest` pydantic model validation
  - `swap-alias` endpoint (swap, clear, transfer, conflicts, missing models, edge cases)
  - `update_model_settings` alias handling
  - CLI `alias_command` function (set, swap, remove, list, error paths)
[...]

PR #1838, opened 2026-06-12
feat(profiles): expose profiles as API-visible models (web + macOS UI)

### LINKED ISSUE #341: [Feature Request] Add mutiple model setting profiles support

### LINKED ISSUE #1641: Inconsistent profiles names rendered in Macos App UI
**Actual behavior**
Render profile id.

**Expected behavior**
Render profile `display_name`.

**Screenshots**
Left side - macos app
Right side - web admin
Image

 - oMLX Version 0.4.1

pablomoralesm:
Closes https://github.com/jundot/omlx/issues/341
Closes https://github.com/jundot/omlx/issues/1641 (the web and macOS UIs now render the profile name consistently)

## Summary

Lets any per-model profile be optionally exposed as its own API-visible model. With "Expose as model" enabled, `/v1/models` also lists `<model>:<profile>` (e.g. `qwen3-8b:thinking`), which serves on the **same engine** as the base model with the profile's request-time settings overlaid — no extra memory, no reload.

## Design

- **Single mechanism** built on the existing profiles — no separate alias type or new file convention.
- **Request-time overlay only** — engine-construction fields (DFlash, TurboQuant, MTP, …) are filtered out, since the engine is shared with the base model.
- **Alias-aware IDs** — `<alias>:<profile>` is advertised when the base model has an alias; the directory-name form still resolves. Provider-prefixed forms (`provider/model:profile`) resolve too.
- **Strict naming** — the profile name *is* the exposed model-ID suffix, so it's validated and rejected on bad input (never silently rewritten), matching the server's `validate_profile_name` and the macOS app.
- **macOS app** — an opt-in "Expose as model" toggle on the profile card, kept in sync with the web console.

#### Eye button toggle to instantly expose/unexpose a specific profile 
image

#### When exposed the eye turns green (if profile does not contain any engine params) or (yellow if the profile contains engine params that will be ignored) - shows exposed name (and warning if needed) on hover 
image

#### Toggle to instantly expose/unexpose the selected profile
image

#### When active shows exposed name
image

#### Shows warning to user if profile contains engine only settings
image

### Chat (shows exposed profiles only)
image

## Known limitations

- Editing a profile's *settings* still requires applying it to the base model (inherited from the existing profiles UX); exposing/unexposing and renaming do not.
- For diffusion models (the `diffusion_gemma` type), an exposed profile's settings are overlaid per request without the diffusion-specific sanitization that the save and apply paths perform. This is harmless in practice — the diffusion lane ignores text-lane fields (sampling, thinking) and structured output is already rejected for diffusion — so the served settings dict may just carry inert fields. Fully closing it would mean sharing the diffusion sanitizer across layers, left out to keep this PR focused.

## Related work & disclosure

I developed this independently and discovered the existing issue and open PRs only afterward. Flagging them up front in the spirit of consolidation, not competition:

- **#341** — the feature request (multiple setting profiles per model). This PR builds directly on the per-model profiles it asks for.
- **#1616** (@ilukashin) — shares the core idea and the `model:profile` / `alias:profile` request syntax. Where this PR differs: an explicit opt-in (`expose_as_model`), listing exposed profiles in `/v1/models` so OpenAI-compatible clients can discover them, filtering engine-construction fields out of the per-request overlay, and macOS app support.
- **#773** (@TipKnuckle) — same goal via a new `*.virtual.yaml` file convention. This PR instead reuses existing profiles (no new file type or model entity) and gates exposure behind a single toggle, with the UI-clutter concern raised in #341 in mind.
- **#1705** (@maxness) — adjacent (CLI alias management + Bearer admin auth); noted since it also touches `admin/routes.py`.

Happy to consolidate with any of these if you'd prefer a single direction — this branch can serve as the reconciling design, or its differentiators (client discoverability, engine-field filtering, native-app support) can fold into whichever PR moves forward.

## Testing

- Python: affected modules covered by unit + integration tests — profiles manager, engine pool, server `/v1/models`, and the admin routes (request models and end-to-end create/toggle/rename). Full fast suite passes (one pre-existing, dependency-gated diffusion-patch test fails in my dev env, unrelated to this change).
- macOS: app builds and the Swift unit tests pass.
- Manually verified real inference: a base model and its exposed `…:thinking` variant both served correctly off a single engine load (reasoning streamed only on the thinking variant).


ilukashin:
I think this one is most production-ready implementation because it covers more cases, such as macos native app and `/v1/models`. 
Also it covers https://github.com/jundot/omlx/issues/1641 issue.

I recommend you to check https://github.com/jundot/omlx/pull/1616#issuecomment-4639792708 - there is some edge cases mentioned.

In my opinion - `eye button` may be not human-native to understand what it does. Maybe think about some more intuitive icon check-box styled, something explicitly show it on/off state.

And what about to see profiles as list?
As user, if i want to use profiles more often it will be good to have access them with more easy way. With list i can see all profiles and its settings in one view. I mean this approach:
Снимок экрана — 2026-06-12 в 11 31 15
If you think so, maybe it should be copy-to-clipboard button to use `model:profile` pattern fast.

pablomoralesm:
Thanks for the kind review @ilukashin, and for flagging the edge cases. We actually landed on the same `model:profile` / `alias:profile` syntax independently, which feels like a good sign it's the right shape. Good catch on #1641 too.

A few notes on the three things @jundot raised on #1616, since this branch handles them a bit differently:

1. It keys off the profile's internal `name` (the slug), not the `display_name`. The name is validated as a slug and rejected if it isn't one, rather than silently rewritten, so two profiles can't collide and renaming re-derives the id.
2. Profile settings get merged on top of the base model's (the request-time ones), not used on their own.
3. The settings are applied on the chat, Anthropic and Responses paths too (plus sampling/thinking/context), so things like `enable_thinking` and `chat_template_kwargs` actually take effect.

I don't really mind which PR lands though, happy to move any of this into #1616 if that's a better base.

On the UX stuff (clearer expose control, a profiles list, copy-to-clipboard), all good ideas. They do cut against @jundot's point in #341 about keeping the UI uncluttered, so probably worth his call first.

@jundot, for the expose control, do you want (a) to keep it minimal in the chip row, (b) a clearer labeled toggle per profile, or (c) a profiles list view with a toggle and the visible `model:slug` plus a copy button? I'm happy to land this as-is for now and do the list view + copy-to-clipboard as a follow-up, ideally something @ilukashin and I do together since they're his ideas.

jundot:
Thanks for the thorough work on this. As discussed in this PR thread, there has been a long-running need to attach multiple usable profiles to a single model, and this is the most complete implementation of that direction so far. I appreciate the care you put into covering the server paths, `/v1/models` discovery, admin UI, macOS UI, and the request-time overlay behavior.

I’m going to merge this PR and use it as the foundation for exposed profile models. After merging, I’ll add a maintainer follow-up for the remaining product/API details rather than blocking this PR on those decisions.

The follow-up I have in mind is:

- Keep a separate API-facing profile name from the human-facing display name, so existing profiles with generated internal names don’t show up as awkward model IDs.
- Add a migration/defaulting path that derives the API name from the existing display name where possible, with deterministic collision handling.
- Make the UI clearer than the eye-only control: likely “Use as API model”, showing the exposed ID such as `<model>:<api-name>` with a copy affordance.
- Keep profile chips focused on the display name, with an API badge/state when a profile is exposed.
- Add collision guards so exposed profile IDs cannot conflict with existing model aliases or model IDs.
- Fold in the small compatibility fix needed against current main around settings-manager stubs and `get_settings_for_request`.

This looks good to me as the base implementation. Thank you again for pushing this feature forward in such a comprehensive way. I’m going to merge it and continue from here with the follow-up polish.

pablomoralesm:
Thanks, glad it landed and that it works as a base for this. The follow-up direction makes sense to me, especially separating the API name from the display name and the collision guards. Appreciate the careful review and for carrying it forward.

PR #1839, opened 2026-06-12
fix: route VLM MTP through adapter for mRoPE position encoding

### LINKED ISSUE #1779: Feature: Support `mlx_vlm`-style speculative decoding (draft models) for Qwen in model settings
  I am currently comparing performance on an M3 Max (48GB). When using oMLX, the `oQ4-mtp` variant runs at ~62 tok/s, which is noticeably slower than just loading the mlx-community 4 bit MLX model in oMLX (~88 tok/s).  

  Conversely, using `mlx_vlm` with the same base model plus the draft model (`...MTP-bf16`) jumps to ~140 tok/s. The current oMLX settings page supports draft models for Gemma, but lacks this capability for Qwen, leaving users unable to access the ~1.6x speedup provided by `mlx_vlm`'s speculative decoding implementation.

  Extend the model configuration/settings UI to support `mlx_vlm`-style speculative decoding draft models for Qwen (and potentially other architectures). This would allow users to easily pair a base model with a matching draft model directly through oMLX's interface, unlocking roughly ~1.5x performance gain demonstrated by `mlx_vlm` without leaving the oMLX.

  - **Current oMLX MTP path**: Using `oQ4-mtp` variants, but yields ~62 tok/s (slower than the base model).
  - **Separate `mlx_vlm` CLI**: Achieves ~140 tok/s with the draft model, but requires juggling two tools, losing oMLX's server/orchestration benefits, and complicates deployment.
  - **Manual patching**: Possible but fragile and not user-friendly for model selection.

  **Additional context**  

  *Hardware:* Apple M3 Max (48GB RAM)  

  *Test Model:* `mlx-community/Qwen3.6-35B-A3B-4bit` (base) + `mlx-community/Qwen3.6-35B-A3B-MTP-bf16` (draft)  

  **Performance Comparison:**

  | Configuration                                                | Throughput (tok/s) | Notes                                                        |
  | ------------------------------------------------------------ | ------------------ | ------------------------------------------------------------ |
  | `oMLX` + Jundot/Qwen3.6-35B-A3B-oQ4-mtp                      | ~62                | Built-in MTP (slower than base ?? maybe i am missing some config?) |
  | `oMLX` + mlx-community/Qwen3.6-35B-A3B-4bit (base)           | ~88                | Standard base model                                          |
  | `mlx_vlm` + mlx-community/Qwen3.6-35B-A3B-4bit (base)        | ~88                | Standard base model                                          |
  | `mlx_vlm` + base + draft mlx-community/Qwen3.6-35B-A3B-MTP-bf16 | ~140               | Speculative decoding active                                  |

  **Draft Model Speculative Decoding Metrics (`mlx_vlm`):**

  Prompt: 19 tokens, 288.5 tok/s
  Generation: 1094 tokens, **140.8 tok/s**
  Peak memory: 22.6 GB
  Speculative decoding: 2.66 accepted tokens/round (1.66 accepted drafts/round, 82.9% acceptance rate, avg draft length 2.0) over 412 rounds

  **oQ4-mtp variant Model Decoding Metrics:**

  omlx.engine_core - INFO - [-] - Engine started
  omlx.engine.vlm - INFO - [-] - VLM tool calling enabled: parser=qwen3_coder
  omlx.engine.vlm - INFO - [-] - VLMBatchedEngine loaded: ~/models/Jundot/Qwen3.6-35B-A3B-oQ4-mtp
  omlx.engine_pool - INFO - [-] - Loaded model: Qwen3.6-35B-A3B-oQ4-mtp (actual: 20.40GB, estimated: 21.14GB, total: 21.14GB)
  omlx.engine.vlm - INFO - [-] - GrammarCompiler initialized for ~/models/Jundot/**Qwen3.6-35B-A3B-oQ4-mtp**
  omlx.patches.mlx_lm_mtp.batch_generator - INFO - [-] - MTP path activated for uid=0 (model has mtp_forward, batch=1)
  omlx.scheduler - INFO - [-] - Enabled boundary cache snapshots for stateful non-sliceable cache layers
  omlx.patches.mlx_lm_mtp.batch_generator - INFO - [-] - MTP[0] finish=stop tokens=2069 cycles=1127 accept=940/1127 (83.4%) emits[init=2,draft=940,bonus=940,verify=187] timing[backbone=24977.5ms mtp=4328.5ms sample=3792.4ms cache=25.3ms]
  omlx.scheduler - INFO - [-] - Cache phase timings: boundary_capture_extract=0.1ms/1, boundary_capture_sync=0.0ms/1, boundary_snapshot_save=9.1ms/1, cleanup_finished_sync=0.0ms/1, store_cache_main_boundary=0.0ms/1, store_cache_main_collect=0.0ms/1, store_cache_main_dispatch=1.1ms/1, store_cache_worker_sync=0.2ms/1
  omlx.server - INFO - [-] - Chat completion: 2068 tokens in 35.99s (**61.9 tok/s**), prompt: 17, finish_reason=stop, max_tokens=32768, request_max_tokens=None

imi4u36d:
## Summary

- VLM MTP decode path bypassed `VLMModelAdapter`, using `_language_model` directly — correct for Gemma 4 (no mRoPE), but wrong for Qwen3.6 VLM which uses mRoPE
- Prefill-final-forward and verify-forward both ran without mRoPE position encoding, producing garbled output
- Added `_VLMAdapterMTPProxy` and `_MTPResetBindingProxy` to route verify through adapter while preserving drafter `bind()` access to `language_model`
- Scheduler now sets `batch_rope_deltas` and calls adapter instead of raw language model

Closes #1779

## Test plan

- [x] 10 new unit tests for proxy visibility and delegation (`test_vlm_mtp_proxy.py`)
- [x] Updated existing scheduler MTP test
- [x] Manual verification: Qwen3.6-35B-A3B VLM with `vlm_mtp_enabled=True` produces correct output
- [x] Manual verification: Gemma 4 VLM MTP still works (no regression)

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -5737,6 +5737,12 @@ class Scheduler:
             self._remove_uid_from_active_batch(uid)
             if hasattr(self.model, "unregister_rope_delta"):
                 self.model.unregister_rope_delta(uid)
+            if uid < 0:
+                mtp_state = self._vlm_mtp_active.pop(uid, None)
+                if mtp_state is not None:
+                    close = getattr(mtp_state.generator, "close", None)
+                    if callable(close):
+                        close()
             del self.uid_to_request_id[uid]
             del self.request_id_to_uid[request.request_id]
 
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -5128,6 +5128,7 @@ class Scheduler:
                 request.request_id,
             )
             return None
+        target_model = self.model
 
         if not last_tokens:
             logger.warning(
@@ -5140,7 +5141,10 @@ class Scheduler:
         last_arr = mx.array(last_tokens)[None]  # (1, len_last)
         try:
             with mx.stream(self._stream):
-                out = lm(
+                set_batch_rope = getattr(target_model, "set_batch_rope_deltas", None)
+                if callable(set_batch_rope):
+                    set_batch_rope(mx.array([request.rope_deltas]))
+                out = target_model(
                     last_arr,
                     cache=prefilled_cache,
                     return_hidden=True,
@@ -5185,11 +5189,15 @@ class Scheduler:
 
         try:
             generator = run_vlm_mtp_decode(
-                target_language_model=lm,
+                target_language_model=target_model,
                 drafter=drafter,
                 prompt_cache=prefilled_cache,
                 hidden=hidden,
-                shared_kv_states=getattr(out, "shared_kv_states", {}) if not isinstance(out, tuple) else {},
+                shared_kv_states=(
+                    getattr(out, "shared_kv_states", {})
+                    if not isinstance(out, tuple)
+                    else {}
+                ),
                 first_bonus=int(first_bonus_arr.item()),
                 max_tokens=request.sampling_params.max_tokens,
                 sampler=mtp_sampler,
--- a/omlx/speculative/vlm_mtp.py
+++ b/omlx/speculative/vlm_mtp.py
@@ -184,6 +184,52 @@ class VLMMTPDrafter:
         self.source_path = source_path
 
 
+class _VLMAdapterMTPProxy:
+    """Expose VLM adapter calls while letting Qwen drafters bind embeddings.
+
+    mlx-vlm's MTP loop calls ``model.language_model`` when that attribute is
+    present, which would bypass oMLX's VLM adapter and lose mRoPE position
+    handling. Qwen's external drafter, however, needs a ``language_model``
+    attribute during ``bind()`` to find ``embed_tokens``. This proxy only
+    exposes ``language_model`` while ``draft_model.reset(model)`` runs.
+    """
+
+    def __init__(self, adapter: nn.Module, language_model: Any) -> None:
+        self._adapter = adapter
+        self._language_model = language_model
+        self._expose_language_model = False
+
+    def __getattr__(self, name: str) -> Any:
+        if name == "language_model":
+            if self._expose_language_model:
+                return self._language_model
+            raise AttributeError(name)
+        return getattr(self._adapter, name)
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
+        return self._adapter(*args, **kwargs)
+
+
+class _MTPResetBindingProxy:
+    """Temporarily expose ``language_model`` during drafter reset."""
+
+    def __init__(self, drafter: nn.Module, target_proxy: _VLMAdapterMTPProxy) -> None:
+        self._drafter = drafter
+        self._target_proxy = target_proxy
+
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self._drafter, name)
+
+    def reset(self, target_model: Any, *args: Any, **kwargs: Any) -> Any:
+        if target_model is self._target_proxy:
+            self._target_proxy._expose_language_model = True
+            try:
+                return self._drafter.reset(target_model, *args, **kwargs)
+            finally:
+                self._target_proxy._expose_language_model = False
+        return self._drafter.reset(target_model, *args, **kwargs)
+
+
 def load_vlm_mtp_drafter(path: str) -> Optional[VLMMTPDrafter]:
     """Load an MTP drafter (gemma4_assistant or qwen3_5_mtp); return None
     and log if the artifact is the wrong kind. Soft-fails so a misconfigured
@@ -267,6 +313,13 @@ def run_vlm_mtp_decode(
     already emitted the bonus token before the round loop starts
     (``emitted = 1`` baked in at the top of both helpers).
     """
+    target_for_rounds = target_language_model
+    drafter_model = drafter.model
+    adapter_lm = getattr(target_language_model, "_language_model", None)
+    if adapter_lm is not None:
+        target_for_rounds = _VLMAdapterMTPProxy(target_language_model, adapter_lm)
+        drafter_model = _MTPResetBindingProxy(drafter.model, target_for_rounds)
+
     is_batch = isinstance(first_bonus, mx.array) and first_bonus.size > 1
 
     if is_batch:
@@ -274,8 +327,8 @@ def run_vlm_mtp_decode(
         yield [int(x) for x in first_bonus_list]
         eos_set = set(eos_token_ids) if eos_token_ids else None
         for tokens, _ in _mtp_rounds_batch(
-            target_language_model,
-            drafter.model,
+            target_for_rounds,
+            drafter_model,
             prompt_cache,
             hidden,
             shared_kv_states,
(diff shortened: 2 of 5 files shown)


imi4u36d:
Additional context for this PR:

While validating the VLM MTP adapter/mRoPE fix, I also tightened several related edge cases that showed up in the same decode path.

1. Qwen / Qwen MoE VLM MTP mRoPE handling

   The core issue remains that the external MTP drafter path could let `_mtp_rounds()` resolve `model.language_model` and bypass `VLMModelAdapter`. For Qwen/Qwen MoE VLMs this drops the adapter-side mRoPE `position_ids` / `rope_deltas` handling during target verification.

   The proxy keeps target verification routed through the adapter, while still temporarily exposing `language_model` during drafter `reset()` / `bind()` so external drafters can bind embeddings. For mRoPE adapters, it also hides `model` / `speculative_*` fast-path attributes to avoid falling back into a lower-level path that would bypass the adapter again.

2. Gemma4 VLM MTP proxy compatibility

   Gemma4's MTP round loop requires the target object to expose `rollback_speculative_cache`. Because the proxy intentionally hides `language_model`, Gemma4 could fail on the first `_mtp_rounds()` step if the adapter itself did not expose that method.

   The proxy now falls back to the wrapped language model for required target-side methods such as `rollback_speculative_cache`, while still keeping `language_model` hidden from `_mtp_rounds()`.

3. Gemma4 MoE long-generation repetition / loop behavior with VLM MTP

   oMLX's wrapper bypasses upstream `run_speculative_rounds()`, so it also bypassed upstream's single-request `_buffer_mtp_target_cache()` setup.

   Gemma4 long generations eventually hit sliding / rotating KV-cache behavior. Without buffered rotating caches, speculative verify + rollback can drift after the window starts moving, which can show up as repeated output or a loop. The wrapper now calls `_buffer_mtp_target_cache()` before entering single-request `_mtp_rounds()`, matching upstream's single-request MTP entry behavior.

4. Aborted VLM MTP requests leaving the drafter busy

   VLM MTP requests use negative synthetic UIDs and are not owned by `BatchGenerator`. On abort, `_remove_uid_from_active_batch()` correctly no-ops for negative UIDs, but `_vlm_mtp_active[uid]` also needs to be cleared. Otherwise later requests keep seeing the serialized drafter as busy and fall back with:

   `vlm_mtp routing skipped ... drafter is busy with 1 request(s); falling back to BatchGenerator`

   Abort cleanup now removes the negative UID from `_vlm_mtp_active` and closes the generator.

5. Avoiding a Qwen MoE first-request corruption regression

   I intentionally did not pass scheduler `last_arr` as `prompt_tokens` into `_mtp_rounds()`.

   Qwen's external drafter has a `prefill_from_target_hidden()` path, but it expects `prompt_tokens` and `hidden` to be aligned to the full prompt. In the scheduler's VLM MTP route, `last_arr` is only the final-forward token slice and the hidden state is only for the final token. Passing that partial state can corrupt the Qwen drafter's first-round seed and cause first-request garbage output.

   The wrapper still supports optional `prompt_tokens`, but the scheduler does not provide partial prompt tokens.

Test coverage added / updated:

- `tests/test_vlm_mtp.py`
  - single-request VLM MTP buffers target cache before `_mtp_rounds()`
  - batch VLM MTP does not use the single-request target-cache buffer
  - optional `prompt_tokens` is only a wrapper-level passthrough

- `tests/test_vlm_mtp_proxy.py`
  - proxy hides `language_model` by default
  - drafter reset/bind temporarily exposes `language_model`
  - `rollback_speculative_cache` can fall back to the wrapped language model
  - mRoPE adapters hide `model` / `speculative_*` fast paths

- `tests/test_scheduler.py`
  - VLM MTP first bonus still uses suppress-token sampling
  - scheduler does not pass partial `prompt_tokens` into VLM MTP
  - aborting a negative-UID VLM MTP request clears `_vlm_mtp_active` and closes the generator

Validated with:

```bash
uv run pytest tests/test_vlm_mtp.py tests/test_vlm_mtp_proxy.py tests/test_scheduler.py::TestSchedulerAbortRequest tests/test_scheduler.py::TestSchedulerSuppressTokens::test_vlm_mtp_first_bonus_uses_suppressing_sampler
```

Result:

```text
39 passed
```

varmint708:
@imi4u36d Thanks for the help on this.

Compiled on commit 1f477a6:

The output is clean, can confirm, Thanks again for the help.

results with exact same model/draft across all of them:

model: froggeric/Qwen3.6-35B-A3B-Uncensored-Heretic-MLX-4bit
draft: mlx-community/Qwen3.6-35B-A3B-MTP-bf16

oMLX build @ 1f477a6 : ~96 tok/s across 5 tests. mtp acceptance ~64%
mlx_vlm.generate ~139 tok/s across 5 tests. mtp acceptance ~81%
mlx_vlm.server ~128 tok/s across 5 tests. mtp acceptance ?

still not sure why the difference in tok/s across omlx and mlx_vlm but this is lot better then earlier, so thanks for the help!! 

sample log from one:

```log
omlx.scheduler - INFO - [-] - vlm_mtp stats: request=e2af9130-1831-4047-a3ad-7d71d5c0c151 finish=stop rounds=1079 accepted=1414/2158 (65.5%) tokens_per_round=2.31 emitted=2493 block_size=3
omlx.scheduler - INFO - [-] - Cache phase timings: boundary_capture_extract=0.0ms/2, boundary_capture_sync=0.1ms/2, cleanup_finished_sync=0.1ms/2, store_cache_main_boundary=0.0ms/2, store_cache_main_collect=0.0ms/2, store_cache_main_dispatch=1.4ms/2, store_cache_worker_sync=0.2ms/1
omlx.server - INFO - [-] - Chat completion: 2492 tokens in 26.09s (96.8 tok/s), prompt: 18, finish_reason=stop, max_tokens=32768, request_max_tokens=None
```

```log
Prompt: 19 tokens, 223.773 tokens-per-sec
Generation: 1094 tokens, 140.646 tokens-per-sec
Peak memory: 22.586 GB
Speculative decoding: 2.66 accepted tokens/round (1.66 accepted drafts/round, 82.9% of drafted, avg draft 2.00) over 412 rounds
```

imi4u36d:
@varmint708 This might involve deeper underlying logic, such as balancing the block_size and the acceptance rate. My suggestion is to try bumping block_size up to 3 for testing. When the acceptance rate is exceptionally high, block_size becomes the performance bottleneck.

Also, keep in mind that the gains brought by MTP can vary significantly across different task types, so we should account for this when running benchmarks. By the way, I think it might be better to move this discussion to an issue so we can gather more insights from the community.

varmint708:
@imi4u36d the log and results were there for reference :) . I am totally happy with this fix, great speedup and good output with this fix :) So thanks again for the help, cheers to you, Jundot  and the community at Large!

jundot:
Thanks for the fix and the detailed context.

I checked the changed VLM MTP path against the scheduler and mlx-vlm round-loop behavior. The adapter routing looks correct: Qwen mRoPE verification now stays behind `VLMModelAdapter`, while the proxy still exposes the language model only during drafter reset/bind and keeps the Gemma4 rollback path available.

I also ran the focused tests from the PR, including the proxy, VLM MTP wrapper, abort cleanup, and suppress-token scheduler coverage, and they all passed. This looks good to me, and I'm going to merge it.

Issue #1825, opened 2026-06-11
thinking_budget is silently ignored on /v1/completions


efortin:
## Environment

- omlx main @ `6ae5142`
- macOS, Apple Silicon (M5 Pro, 64 GB)
- Model: `mlx-community/Qwen3.6-35B-A3B-nvfp4`, thinking enabled

## Symptom

`thinking_budget` works on `/v1/chat/completions` (and `budget_tokens` on `/v1/messages`), but on `/v1/completions` it is silently ignored: send it, get no error, and the model reasons to its natural length anyway.

```bash
curl -s http://localhost:1234/v1/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "Qwen3.6-35B-A3B-nvfp4",
    "prompt": "<|im_start|>user\nExplain why the sky is blue.<|im_end|>\n<|im_start|>assistant\n<think>\n",
    "max_tokens": 2500,
    "thinking_budget": 300
  }'
```

Expected: thinking bounded around 300 tokens. Actual: ~1450 thinking tokens, exactly like the same request without the parameter.

## Cause

`CompletionRequest` has no `thinking_budget` field, so Pydantic drops it at parsing, and neither completion path resolves a budget for the engine. Raw completions are otherwise a natural fit: a prompt ending with an open `<think>` already passes the scheduler's `needs_think_prefix` gate, so enforcement works as soon as the value reaches the engine.

## Fix

Up in #1821: add the field and thread it through both completion paths via the existing `_resolve_thinking_budget` helper. With it: budget 300 → 299 thinking tokens, deterministic at temperature 0 (100 → 99, 50 → 49).

socketa4techx7:
To add on top of that, I've noticed on M1 Max, omlx 0.4.3, qwen3.6-35b-a3b-ud-mlx-4bit, that setting thinking_budget to 128 just lets the thinking phase "spill out" of the thinking block and continue as part of the answer.

Note: it seems to behave this way with non-english text/prompts. My example was cyrillic. (adding small screenshot to demonstrate the transition in the omlx chat).

<img width="760" height="222" alt="Image" src="https://github.com/user-attachments/assets/a127288f-65b3-47a9-b6a8-e77aff4004d5" />

efortin:
[...]

PR #1844, opened 2026-06-12
fix(server): forward thinking_budget on /v1/completions (#1825)

richgoodson:
## Problem

#1825: thinking_budget works on /v1/chat/completions (and budget_tokens on /v1/messages) but is silently ignored on /v1/completions. The cause is what the reporter diagnosed: CompletionRequest has no thinking_budget field, so Pydantic drops the parameter before the handler sees it. The completion handlers also never call _resolve_thinking_budget, so a model-level budget never reached the engine on this endpoint either.

## Change

Mirrors the chat path:

- Add thinking_budget to CompletionRequest.
- Resolve it with _resolve_thinking_budget in create_completion (non-streaming) and stream_completion, and pass it to engine.generate / engine.stream_generate only when set, same as chat_kwargs on the chat path.

Behavior note: model-level thinking budgets configured in the dashboard now apply to /v1/completions too, where they were previously ignored. That matches the chat endpoint, but it is a change for raw-completion users of models with a budget configured.

Two scope notes:

- On this endpoint there is no chat template, so the budget caps thinking only when the prompt opens the think block (as the issue's repro does with a trailing `<think>\n`) or the model emits one.
- The spill-out behavior in the issue's second comment (thinking continuing past the close at low budgets) is in ThinkingBudgetProcessor and is not addressed here.

## Testing

- 3 new tests in tests/integration/test_server_endpoints.py: non-streaming and streaming forwarding of the request parameter, plus the model-settings fallback.
- Full suite: 5613 passed, 40 skipped.
- Draft until I finish the live test on a thinking model: run the issue's curl with thinking_budget 300 and confirm thinking caps near 300 tokens instead of ~1450.

### DIFF
--- a/omlx/api/openai_models.py
+++ b/omlx/api/openai_models.py
@@ -396,6 +396,8 @@ class CompletionRequest(BaseModel):
     frequency_penalty: float | None = None
     # Seed for reproducible generation (best-effort)
     seed: Optional[int] = None
+    # Cap reasoning/thinking tokens (parity with /v1/chat/completions)
+    thinking_budget: Optional[int] = None
 
     @field_validator("stop", mode="before")
     @classmethod
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -2750,6 +2750,11 @@ async def create_completion(
             req_xtc_threshold=getattr(request, "xtc_threshold", None),
         )
 
+        gen_kwargs = {}
+        thinking_budget = _resolve_thinking_budget(request, request.model)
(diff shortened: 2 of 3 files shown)


jundot:
Thanks for narrowing this down.

This fixes the actual /v1/completions gap for #1825: the request model now accepts thinking_budget, and both completion paths resolve and pass it to the engine, including model-level settings. I'm going to take this smaller wiring fix here; the streaming prefix and validation polish can be handled separately.

PR #1854, opened 2026-06-13
fix(api): parse Gemma 4 namespaced and single-quoted tool calls

### LINKED ISSUE #1830: QAT Gemma 4 model emits call:google:mcp: tool call format that oMLX parser cannot handle — calls silently dropped

richgoodson:
Fixes #1830.

## Problem

Gemma 4 models emit MCP-namespaced tool names and single-quoted argument values that the parser cannot handle. The reporter's log shows both at once:

```
Native tool parser failed ... Dropping match: "call:google:mcp:text_generation:create-pdf-file{filename: ..."
```

Four defects, three in the parser and one downstream:

1. mlx-lm's native gemma4 parser rejects any name containing colons (its name regex is `[\w-]+`), so these calls always land in the oMLX fallback. The fallback parses the name but returns it verbatim; a client that registered `create-pdf-file` cannot match `google:mcp:text_generation:create-pdf-file`. With simple arguments this is the quieter half of the symptom: the parse "succeeds" and the client silently receives a call it cannot use, with no warning logged.
2. Single-quoted values keep their literal quotes (`'output.pdf'` parses to the string `"'output.pdf'"`).
3. Commas, colons, or braces inside a single-quoted value shred the parse (garbage keys, or a truncated match in the brace-balancing regex). This is the logged drop above; real document content always has commas.
4. When the call lands in thinking content, `extract_tool_calls_with_thinking` filters by exact name match, so even a cleanly parsed namespaced call is silently dropped there.

Single quotes are not part of Gemma 4's tool-call format (the spec delimiter is `<|"|>`); they are degraded emissions observed in the wild, so the parser treats them tolerantly rather than as grammar.

## Fix

All in `omlx/api/tool_calling.py`, two pieces.

**Parse.** Replaced the recursive-regex brace balancer with an iterative linear scanner that tracks quote state: `<|"|>` pairs, and single-quoted values anchored to value positions so apostrophes inside prose (don't, it's) cannot pair across values. Rewrote `_gemma4_args_to_json_robust` as a single-pass transcode covering bare keys and values, single-quoted and `<|"|>` strings, arrays, and nested objects. Depth (64) and length (256 KB) bounds fail as clean parse errors on the existing drop path.

**Remap.** New `_remap_tool_call_names`, applied once post-parse in `parse_tool_calls`. If the emitted name matches no registered tool and exactly one registered tool is a `:`-boundary suffix of it, the call is remapped to that tool; zero or several candidates keep the name verbatim. Matching is boundary-aligned by construction (split on `:`), never `endswith`, so a crafted `evilcreate-pdf-file` cannot coerce into `create-pdf-file`. Remaps are logged at INFO. Post-parse placement covers every producer path (native parser, fallback, XML recovery, and the thinking-content filter from defect 4) and survives a future mlx-lm parser that accepts colon names. The remap only maps into the set of registered tools, so it grants nothing the model could not already call directly.

**Why the helpers were replaced rather than patched.** The old helper had two pre-existing weaknesses the rewrite removes in passing: literal NUL bytes in model output could forge its `\x00N\x00` placeholder references and cross-contaminate argument values, and the recursive regex degraded quadratically on unbalanced input (measured ~590 ms at 80 KB). Model output is untrusted input; the new scanner is linear and has no placeholders to forge.

## Root cause vs. upstream

The rejection of colon names is in mlx-lm (`tool_parsers/gemma4.py`). I fixed it in the oMLX fallback because that is the layer oMLX owns and the fallback exists for exactly this. A new test pins the dispatch contract against the real mlx-lm parser, so if a future mlx-lm bump starts accepting colon names we will see it. I can file the name-regex half upstream as well if you want it there too.

## Testing

- 25 new tests, 199 total in `tests/test_tool_calling.py`, all 174 existing tests untouched and green. Table-driven round trips for the new argument forms; the issue's exact payload end to end through mlx-lm's real Gemma 4 parser with real paired markers; remap boundary, ambiguity, and no-op rules; thinking-path promotion; clean failure on NUL bytes, deep nesting, and oversized input.
- Full suite: 5653 passed, 22 skipped, 0 failed.
- Live on an M4 Pro mini (64 GB), both `gemma-4-E4B-it-MLX-4bit` and the reporter's `mlx-community/gemma-4-26b-a4b-it-4bit`, streaming and non-streaming:
  - With plain tools both models emit the clean spec format at temperature 0 (plain name, `<|"|>` delimiters), so the native path is the no-regression baseline.
  - The spontaneous `google:mcp:` prefix did not reproduce locally (6 of 6 clean emissions at temperature 0.7 even with MCP-flavored prompts); it evidently needs the reporter's AnythingLLM context. I exercised the fixed paths live anyway:
  - Registering a colon-named tool (`text_generation:create-pdf-file`) forces the native parser to fail on the model's verbatim echo. The new fallback parsed it end to end; the exact match correctly did not remap.
  - With a bare registered name and a system prompt pushing fully qualified names, the 26B emitted literally `call:google:mcp:text_generation:create-pdf-file{...}`. The fallback parsed it, the remap log fired, and the client received `create-pdf-file` with intact arguments. On main this same request returns the unusable namespaced name.

## Out of scope

Tool-call parsing runs only when the request carries `tools`, so this fix helps tool-bearing clients. The reporter's second ask (a passthrough mode so markers are not stripped when native tool calling is off) is a config-surface change and not attempted here; I can take it as a separate follow-up.

### DIFF
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -21,6 +21,7 @@ import bisect
 import json
 import logging
 import re
+import regex
 import uuid
 from dataclasses import dataclass
 from typing import Any, Dict, List, Optional, Tuple, Union
@@ -417,7 +418,15 @@ class _Gemma4ArgsTooComplex(ValueError):
 # lane's degenerate prefixes (``calldone{`` missing the colon, ``:done{``
 # missing ``call``, #1837) still match; the fallback only runs on
 # marker-delimited content, so a permissive prefix cannot misfire on prose.
-_GEMMA4_CALL_HEAD = re.compile(r"(?:call)?:?([\w.-]+(?::[\w.-]+)*)\{")
+#
+# Compiled with the ``regex`` module, NOT ``re``: once the ``call:`` literal
+# anchor became optional (above), ``re``'s engine restarts the greedy
+# ``[\w.-]+`` match at every position of a long bare argument value and
+# backtracks O(n^2) hunting an opening ``{`` that never comes, hanging on
+# adversarial output (a 300 KB bare value pegs a core indefinitely).  The
+# ``regex`` engine fails that same partial match fast, so finditer stays
+# linear.  See test_oversized_args_fail_cleanly.
+_GEMMA4_CALL_HEAD = regex.compile(r"(?:call)?:?([\w.-]+(?::[\w.-]+)*)\{")
 
 
 def _squote_close_positions(s: str) -> list:
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -17,6 +17,7 @@ Also includes structured output (JSON Schema) utilities:
 """
 
 import ast
+import bisect
 import json
 import logging
 import re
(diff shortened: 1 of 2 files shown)


aikidovation:
I would be happy to test but prefer a built version, else I'd be testing my build-skills more than the fix!

jundot:
Thanks for the detailed fix. I found one edge case that I think should be fixed before merge: `_scan_gemma4_args_span()` still treats braces inside standard JSON double-quoted strings as structural braces.

The case I want to block is a namespaced Gemma fallback call with a normal JSON string value containing `}`:

`call:ns:create{"content": "has } brace"}`

With a registered `create` tool, the fallback can truncate the args at the brace inside the string, and the suffix remap can then turn that corrupted parse into an executable `create` call.

Could you update the span scanner so JSON double-quoted strings are skipped while balancing braces, including escaped quotes, and add an end-to-end remap test that verifies this case preserves `content == "has } brace"`?

richgoodson:
Fixed in `34cf338d`.

`_scan_gemma4_args_span` now treats a bare `"` as a standard JSON string open and skips to the closing unescaped quote while balancing braces, so a `}` inside a double-quoted value can no longer truncate the span. Escapes are honored (`\"` and `\\`), and the `<|"|>` delimiter is still matched ahead of this branch. The scan stays single-pass and bounded by `_GEMMA4_MAX_ARGS_LEN`, so the linear-time guard is unchanged.

Your exact case now round-trips: `call:ns:create{"content": "has } brace"}` with a registered `create` tool remaps to `create` and preserves `content == "has } brace"`. I added that as an end-to-end remap test, plus regressions for an escaped quote inside the value, a lone `{` inside a string (which previously dropped the call entirely rather than truncating), a multi-call where the first call's args contain a `}` (guards the consumed-span bookkeeping between siblings), and an escaped backslash before the closing quote (Windows paths).

One residual I want to flag: if a JSON double-quoted value contains the literal token `<|"|>`, the inner string scan reads the `"` inside that token as the closing quote, the span ends up unbalanced, and the call is dropped. This is the inherent ambiguity between Gemma's `<|"|>` delimiter and plain JSON `"`, and it predates this change. It fails safe (a dropped call, never a corrupted parse that becomes executable), and it requires the model to emit its own delimiter token as literal content inside a JSON string, so I left it as-is rather than re-introduce delimiter awareness into the inner scan and reopen the ambiguity from the other side. Happy to handle it differently if you'd prefer.

jundot:
Thanks for the quick follow-up and for adding coverage for the exact JSON-string brace case I called out. I rechecked that path, and the content now survives the fallback parse and suffix remap intact.

I found one remaining hardening edge around deeply nested valid JSON args, but it is not specific to this report. I will fold that into a small follow-up on main after merge.

Issue #1859, opened 2026-06-13
Native macOS app VLM MTP drafter picker excludes qwen3_5_mtp candidates


chenqianhe:
**Describe the bug**

The native macOS oMLX app does not show valid VLM MTP draft model candidates in the “VLM Draft Model” picker, even though the web admin UI shows the same drafter correctly.

In my setup, `/admin/api/models` returns a valid drafter model with:

- `id`: `Qwopus3.6-35B-A3B-v1-4bit-MLXVLM-MTP-Drafter`
- `config_model_type`: `qwen3_5_mtp`

The web UI can select it, but the native macOS app picker only shows the placeholder `Select assistant drafter...`.

This looks like a native app UI filtering issue. The web UI appears to recognize `qwen3_5_mtp` as a VLM MTP drafter, while the native app seems to only match assistant-style names and misses Qwen MTP drafter models.

Related issues: #1648 and #1779.

**To Reproduce**

1. Prepare a Qwen3.6 / Qwopus MLX-VLM target model and a separate MTP drafter model.
2. Make sure the target model is visible in oMLX, for example:
   - `Qwopus3.6-35B-A3B-v1-4bit-MLXVLM-Target`
3. Make sure the drafter model is also visible in oMLX, for example:
   - `Qwopus3.6-35B-A3B-v1-4bit-MLXVLM-MTP-Drafter`
4. Confirm the backend API reports the drafter with `config_model_type: qwen3_5_mtp`.
5. Open the native macOS oMLX app.
6. Go to the target model settings.
7. Open the Experimental section.
8. Enable `VLM MTP`.
9. Open the `VLM Draft Model` picker.
10. The picker only shows `Select assistant drafter...` and no valid drafter options.

**Expected behavior**

The native macOS app should show the same VLM MTP drafter candidates as the web admin UI.

In this case, the picker should include:

`Qwopus3.6-35B-A3B-v1-4bit-MLXVLM-MTP-Drafter`

because its `config_model_type` is `qwen3_5_mtp`.

**Screenshots**

The native macOS app picker only shows the placeholder option, while the web UI can display/select the drafter correctly.

**Desktop (please complete the following information):**

- macOS Version: NA
- oMLX Version: 0.4.4 rc1

**Additional context**

This appears to be a native macOS app UI parity bug rather than a backend/model discovery problem.

[...]

PR #1860, opened 2026-06-13
fix(macOS): show qwen3_5_mtp models in VLM MTP drafter picker

chenqianhe:
## Summary

- Mirror the web admin VLM MTP drafter detection in the native macOS model settings UI.
- Allow known drafter `config_model_type` values, including `qwen3_5_mtp`, `gemma4_assistant`, and `gemma4_unified_assistant`.
- Keep the existing fallback for model names containing `assistant` or standalone `mtp`, while avoiding false matches like `MTPLX`.
- Add regression coverage for Qwen VLM MTP drafter selection.

https://github.com/jundot/omlx/issues/1859


jundot:
Thanks for fixing this. I checked the native picker path against the web admin behavior and confirmed that the qwen3_5_mtp config type now shows up as a VLM MTP drafter candidate.

The targeted macOS VM test passes locally, and the remaining web-parity difference is not blocking for this fix. This looks good to me, and I'm going to merge it.

PR #1875, opened 2026-06-14
feat: support MiniMax M3 from in-progress mlx-vlm PR

jundot:
## Summary

- Add oMLX support for MiniMax M3 based on the in-progress `mlx-vlm` MiniMax M3 PR.
- Pin `mlx-vlm` to `Blaizzy/mlx-vlm@086ab9d5d575fec64d8d8ad907ce000007c25c1a`.
- Add MiniMax M3 / MiniMax M3 VL model discovery and mlx-vlm routing.
- Add MiniMax M3 cache type handlers for sparse index cache serialization, SSD snapshot storage, and restore.
- Keep MiniMax M3 cache out of TurboQuant eligibility until a dedicated implementation is proven.
- Add MiniMax M3 streaming tool-call suppression and parser integration.

### DIFF
--- a/omlx/adapter/output_parser.py
+++ b/omlx/adapter/output_parser.py
@@ -63,6 +63,8 @@ class OutputParserFactory:
     kind: str
     create_session: Callable[[Any], OutputParserSession]
     stop_token_ids: set[int] = field(default_factory=set)
+    thinking_start_text: str | None = None
+    thinking_start_output_text: str | None = None
     thinking_end_text: str | None = None
     thinking_end_trailing_text: str | None = None
     # Marker strings that must survive special-token stripping so the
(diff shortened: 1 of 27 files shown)


gilby:
If it's useful for testing this PR: I quantized a 4-bit oQ checkpoint and pushed it public —
**[unigilby/MiniMax-M3-oQ4](https://huggingface.co/unigilby/MiniMax-M3-oQ4)** (~228 GiB, fused `gate_up_proj` layout, 4.59 bpw, vision tower preserved).

It loads on the mlx-vlm revision this PR pins (`086ab9d`) — its M3 model code is byte-identical to `8fd6fe7`, which I verified end-to-end on Apple Silicon (M3 Ultra): model load, coherent generation, clean `reasoning_content` separation, structured tool calls (single + parallel), and 768-token sustained generations with no OOM.

One thing this PR doesn't cover is the *quantization* side — oMLX's oQ pipeline can't currently quantize a nested VLM like M3 because the sanitize-plan proxy lacks `language_model`. I opened #1881 (now scoped to just that +7-line `oq.py` fix) so oQ can actually *produce* M3 checkpoints; it's complementary to the serving work here. Happy to fold it in however you prefer.

Issue #1702, opened 2026-06-06
(feat) Submenu for available models


faisal:
Feature Request: a submenu that lists and lets you manipulate available models

- The oMLX menu would have a menu for available models
- Each model could then have a submenu with:
  - A menu item to load or unload the model (possibly displaying a non-selectable loading or unloading text if it's in the middle of doing so)
  - A menu item to jump to the settings for that model
  - A menu item to copy the model name

Right now this involves a fair amount of jumping around in the web UI or the native UI. An alternative might be to put all this functionality into the web dashboard, a new web page, or a new models control pane, but those options would still involve jumping over to the control site, which wouldn't save much.

faisal:
Implemented in https://github.com/jundot/omlx/pull/1882.

PR #1882, opened 2026-06-15
feat: Models menu

faisal:
This change adds a Models menu item under the Start/Stop server controls, giving fine-grained per-model control directly from the menubar. The item is grayed out unless the server is loaded. When loaded, the item presents a menu of each available model. Each available model is then a sub-menu with a load/unload item, a link to that model's in-app settings pane in the app, and a link to copy the model ID.


faisal:
This PR is a PoC implementation of https://github.com/jundot/omlx/issues/1702.

faisal:
Screenshot 2026-06-18 at 18 39 22

penryu:
This is a frequent enough process for me that I currently do via the web interface, and this would save me many clicks/keypresses over a workday.

And like the example screengrab, I routinely have between 2-4 models loaded.

jundot:
Thanks for building this out so nicely as a PoC, and for keeping the branch fresh against main for a whole month. The wiring is clean and the testable helpers are a nice touch. I'll merge this and clean up the merge artifacts myself (the new test fixture is missing the displayName/isFavorite args main added recently, and the pbxproj ended up with the Tests group defined twice), so no need for another pass on your side. I'll also use it for a while, play with the look of the menu, and do some additional polish in follow-ups.

PR #2, opened 2026-02-13
Add streaming tool call parsing

thornad:
## Summary
- Accumulate streamed text and parse tool calls at completion, emitting structured `tool_calls` chunks in OpenAI streaming format
- Add fallback XML parser for GLM and other models using `<tool_call>` tags when mlx-lm's native `has_tool_calling` is unavailable
- Fixes tool calling with clients like Roo Code that expect structured tool call chunks in SSE stream

## Test plan
- [ ] Send a chat completion request with tools to a GLM model and verify tool calls are returned as structured chunks
- [ ] Verify models with native `has_tool_calling` support still work correctly
- [ ] Test with Roo Code or similar client that consumes OpenAI streaming tool call format

Issue #1846, opened 2026-06-12
Gemma 4 26B (A4B MoE): long agentic contexts trigger parenthesized call:name(...) tool calls; gemma4 parser silently strips them (empty content, no recovery path)


hugo-valcourt:
**TL;DR:** In agent-framework conditions (big system prompt + ~36 tool
schemas + a large tool result), Gemma 4 26B reproducibly switches its tool
markup from the canonical `call:name{...}` to a parenthesized
`call:name(...)` variant. The gemma4 output parser doesn't match it and
deletes the whole block, so the API returns an empty message — the client
can't recover, the operator sees nothing, and the model never learns its
call failed. Bisection data, verbatim raw emission, and two suggested fixes
below. This kills agent loops (Hermes kanban workers in my case: 16
consecutive protocol failures, all this signature).

## Environment

- oMLX v0.4.4rc1 (GitHub release), macOS 15.6.1, M-series 64 GB
  (Metal cap 48 GB), `scheduler.max_concurrent_requests = 1`
- Model: `gemma-4-26B-A4B-it-MLX-4bit` (lmstudio-community)
- Client: Hermes Agent kanban worker (OpenAI chat completions, 36 tools)

## Summary

In short contexts, Gemma 4 26B emits its canonical curly tool markup
(`<|tool_call>call:name{...}<tool_call|>`) and oMLX parses it into
`tool_calls` correctly. In long agentic contexts (~28-41k prompt tokens:
big system prompt + tool schemas + a large tool result), the model
reproducibly degrades to a **parenthesized Python-kwargs variant**:

```
<think>
</think>
<|tool_call>call:todo(todos=[{content:<|"|>Clarify goal and draft labor graph for t_77d3100f<|"|>,id:<|"|>todo-1<|"|>,status:<|"|>pending<|"|>}])<tool_call|>
```

(verbatim raw emission, captured by rendering the same conversation through
the model's own `chat_template.jinja` and replaying it on
`/v1/completions`; 3/3 samples identical in structure)

The envelope markers and `<|"|>` quoting are canonical — only `(`…`)`
replaces `{`…`}`. The gemma4 output parser doesn't match this form and
strips the entire block: the API returns **empty `content`, empty
`tool_calls`, `finish_reason=stop`**, with `completion_tokens` 50-80. The
client (Hermes) sees a successful-but-empty response, retries the identical
prompt, gets the identical degeneration, and the agent run dies — in our
[...]

PR #1886, opened 2026-06-15
fix(tool-calling): parse Gemma 4 parenthesized call:name(...) tool calls (#1846)

richgoodson:
Fixes the parser half of #1846 (the issue's suggested fix #1).

## Problem

Under instruction-dense agentic load (large system prompt + many tool schemas + a big tool result, the reporter measured ~28-41k prompt tokens), Gemma 4 26B reproducibly degrades from the canonical `call:name{...}` markup to a Python-kwargs shell `call:name(key=value, ...)`. The reporter's bisection shows the trigger is an interaction of an instruction-dense system prompt with a large tool-schema surface, not length or tool count alone.

The gemma4 fallback parser matched only the curly form, so the whole `<|tool_call>...<tool_call|>` block was stripped: the API returned empty `content` with `finish_reason=stop` and agent loops died silently with no recovery path for the client, operator, or model (the reporter saw 16 consecutive Hermes failures, all this signature).

## Approach

The parenthesized form is the curly grammar with only the outer shell changed: `()` instead of `{}` and a top-level `=` separator instead of `:`. The nested grammar (objects, arrays, `<|"|>`-quoted strings) is identical, so the #1854-hardened inner transcoder is reused as-is for the inner content; only the outer shell needed new handling.

- add `_GEMMA4_CALL_HEAD_PAREN`, compiled with the `regex` module (not `re`) for the same ReDoS reason as the curly head: the optional `(?:call)?` prefix backtracks quadratically on `re` when hunting an opening `(` that never comes
- generalize `_scan_gemma4_args_span` to any balanced pair (defaults keep the existing brace caller byte-identical)
- accept `(` as the outer object opener and `=` as a key/value separator in the transcoder, additively, so the curly path is unchanged
- terminate a bare value at `)` only when a paren container is open, so the closing paren is not swallowed while curly values may still legitimately contain parentheses
- anchor a single-quote value close on `)` as well as `,` `}` `]`, so a single-quoted last argument (the Python-kwargs shell's native string form) is normalized like the curly form instead of leaking its quotes; the curly form never closes a value with `)`, so its parsing is unchanged
- collect curly and paren heads and process them in document order so the consumed-span dedup holds across both forms

Scope: this is fix #1 from the issue only. The reporter's suggested fix #2 (pass-through instead of strip on parse failure) is a cross-cutting behavior change and is deliberately out of scope here.

## Tests

`TestGemma4ParenthesizedArgs` covers the reporter's exact fixture, kwargs/scalars/arrays, single-quoted values, namespaced names, mixed curly+paren siblings, a curly-values-with-parens regression, and the #1854 security matrix re-run for the paren shape (oversized args, deep nesting, NUL forge, `)` and `=` inside strings), plus one end-to-end integration test through `parse_tool_calls`. Full local suite green.

## Live test (gemma-4-26B-A4B-it 4bit, the reporter's model)

Verified on the real model rather than only the test harness:

- the real loaded tokenizer exposes `tool_call_start = '<|tool_call>'`, so the Gemma 4 fallback gate triggers in production exactly as in the test
- the native `tool_parser` raises `ValueError` on the paren form, so it routes to the fallback rather than being mis-parsed
- the reporter's exact paren fixture is extracted as a `todo` call with the nested args intact
- driven through the live HTTP server, the model emitted the paren form `call:create_todo(content='...')` end-to-end and both calls came back as real `tool_calls` (this is also what surfaced the single-quote normalization above)
- clean-format curly tool calling still works end-to-end through the server (no regression)

## Fails-safe residuals (disclosed)

Both result in a clean DROP of the call, never a corrupt-then-execute:

1. An unbalanced `)` in a BARE (unquoted) value inside a paren call closes the span early and the call is dropped. The common case uses `<|"|>` or single-quote quoting and is immune.
2. A `)` in a nested bare value within a paren call is malformed structure and the call is dropped.

### DIFF
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -486,15 +486,37 @@ class _Gemma4ArgsTooComplexError(ValueError):
 # linear.  See test_oversized_args_fail_cleanly.
 _GEMMA4_CALL_HEAD = regex.compile(r"(?:call)?:?([\w.-]+(?::[\w.-]+)*)\{")
 
+# The parenthesized variant of the head (#1846): under instruction-dense
+# agentic load (large system prompt + many tool schemas + a big tool result)
+# Gemma 4 26B reproducibly degrades from ``call:name{...}`` to a Python-kwargs
+# shell ``call:name(key=value, ...)`` — same name grammar and same ``<|"|>``
+# string quoting, only the OUTER ``{}`` becomes ``()`` and the top-level
+# ``:`` separator becomes ``=``.  The nested grammar (objects, arrays, strings)
+# is unchanged, so the hardened transcoder is reused for the inner content; only
+# the outer shell needs a separate head + span scan.  Compiled with ``regex``,
+# not ``re``, for the SAME reason as ``_GEMMA4_CALL_HEAD``: the optional/tolerant
+# prefix makes ``re`` backtrack O(n^2) on a long bare value hunting an opening
+# ``(`` that never comes (the #1854 ReDoS).  See test_oversized_paren_args.
+_GEMMA4_CALL_HEAD_PAREN = regex.compile(r"(?:call)?:?([\w.-]+(?::[\w.-]+)*)\(")
+
+# Matching close character for each balanced opener the parsers track.
+_GEMMA4_CLOSE_CHAR = {"{": "}", "[": "]", "(": ")"}
+
 
 def _squote_close_positions(s: str) -> list:
     """Indices of single quotes that can CLOSE a single-quoted value.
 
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for splitting this out from #1846. I checked it against the already-merged #1854 path, and this is complementary rather than duplicate: #1854 covers namespaced curly-form calls and remapping, while this adds the parenthesized kwargs shell that still fails on current main.

The change reuses the hardened scanner/transcoder path and keeps the remaining malformed cases on a clean-drop path. This looks good to me, and I am going to merge it.

Issue #1632, opened 2026-06-03
Discussion: Gradual transition toward native SwiftUI idioms for macOS App UI layer


Stv-X:
Hi team,

I’ve been exploring the new native macOS App UI layer recently and am truly impressed by the app's current implementation.

While reviewing the custom components, I noticed explicit references to their JSX origins in the comments, along with naming conventions carried over from the Web version. For instance:

https://github.com/jundot/omlx/blob/02cf64aa61882593b8d2d52784b58b853e4f6815/apps/omlx-mac/Sources/Theme/Components/TextInput.swift#L1

https://github.com/jundot/omlx/blob/02cf64aa61882593b8d2d52784b58b853e4f6815/apps/omlx-mac/Sources/Theme/Components/Segmented.swift#L1

I'm guessing this was likely introduced early on as a solid tactical choice to align quickly with the Web UI, which totally makes sense.

I’ve been exploring the codebase recently and tried a small experiment refactoring a few of these components using modern SwiftUI idioms, just to see how they might look and feel. The initial look seems quite encouraging:
	1.	Potential Code Reduction: By replacing custom drawing, manual selection closures, and complex focus state animations with native SwiftUI idioms (like [`Picker`](https://developer.apple.com/documentation/swiftui/picker) and [`RoundedBorderTextFieldStyle`](https://developer.apple.com/documentation/swiftui/roundedbordertextfieldstyle)`)
	2.	Native Benefits: Moving closer to native views could help us inherit standard macOS system polish, dynamic type resizing, and accessibility support more naturally.

<img width="1828" height="700" alt="Image" src="https://github.com/user-attachments/assets/6806cced-0c39-410e-bbc8-7572685a656b" />

<img width="1828" height="700" alt="Image" src="https://github.com/user-attachments/assets/6638208f-8da3-451b-8f11-7f2c4cd45687" />

I'd love to get the team's thoughts (especially @popfido ) on our long-term vision for these components. Do you think it’s worth gradually moving toward native SwiftUI elements, or should we continue to maintain our current visual consistency with the Web UI?

If this is a direction we’re interested in exploring, I’d be happy to put together a draft PR. No pressure at all, just wanted to share the idea and look forward to your feedback!

Stv-X:
Here is a quick visual comparison from my local experiment to show what moving toward native elements might look like:

[...]

PR #1891, opened 2026-06-16
refactor(omlx-mac): replace custom controls with native SwiftUI equivalents

Stv-X:
This Pull Request addresses the requirements outlined in Issue #1632 by consolidating a sequence of refactoring and build changes aimed at removing hand-rolled UI primitives in favor of their first-party SwiftUI counterparts.

The app's visual language aligns more closely with macOS system conventions, and the component surface that needs ongoing maintenance shrinks substantially. No user-visible behavior is changed unless noted.

See the attached images below for the visual comparison before and after this refactoring (running on macOS 27 Golden Gate).

Accuracy
Logs
Model Settings
Quantization

### DIFF
--- a/apps/omlx-mac/Resources/Localizable.xcstrings
+++ b/apps/omlx-mac/Resources/Localizable.xcstrings
@@ -1,174 +1,6 @@
 {
   "sourceLanguage" : "en",
   "strings" : {
-    "profile.detail.acceleration.vlm_mtp" : {
-      "comment" : "Acceleration chip: VLM multi-token prediction",
-      "extractionState" : "manual",
-      "localizations" : {
-        "en" : {
-          "stringUnit" : {
-            "state" : "translated",
-            "value" : "VLM MTP"
-          }
-        }
-      }
-    },
-    "settings.experimental.vlm_mtp.block_size.label" : {
-      "comment" : "Row label for the VLM MTP draft block-size field",
-      "extractionState" : "manual",
-      "localizations" : {
-        "en" : {
-          "stringUnit" : {
-            "state" : "translated",
-            "value" : "Draft Block Size"
-          }
-        }
-      }
-    },
-    "settings.experimental.vlm_mtp.block_size.sub" : {
-      "comment" : "Sublabel for the VLM MTP draft block-size field",
-      "extractionState" : "manual",
-      "localizations" : {
-        "en" : {
-          "stringUnit" : {
-            "state" : "translated",
-            "value" : "Tokens drafted per round. Empty uses the mlx-vlm default."
-          }
-        }
-      }
-    },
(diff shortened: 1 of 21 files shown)


jundot:
Thanks for the thoughtful refactor. I really appreciate the care that went into the UI pass here, and the before/after screenshots are especially helpful for reviewing the visual impact. I agree with the overall direction of moving these custom controls toward first-party SwiftUI equivalents; that should make the mac app feel more native and reduce the component surface I need to maintain.

I can't merge this as-is yet because the mac app tests fail. The VLM MTP draft-model filter in `ModelSettingsScreenVM` was narrowed to IDs containing `assistant`, which drops the existing `qwen3_5_mtp` config-type and standalone `mtp` fallback behavior that matches the HTML admin. The failing tests are `ModelSettingsScreenVMTests.testVlmMtpDraftModelOptionsIncludeQwenMtpConfigType` and `ModelSettingsScreenVMTests.testVlmMtpDraftModelOptionsKeepAssistantAndStandaloneMtpFallbacks`.

I also noticed two behavior/UI regressions worth fixing in this PR: the Accuracy Bench card now renders a labeled `Toggle` inside a `Button`, which duplicates the benchmark name and can make the toggle interaction unreliable, and `WelcomeWindow.openWebDashboard()` should keep using `AppConfig.httpURL(...)` so host normalization and IPv6 escaping stay correct. Please also clean up the trailing whitespace reported by `git diff --check`.

Once those are fixed and the mac app tests pass, this should be much easier to review for merge.

Stv-X:
@jundot  Thanks for the review! I've addressed all your comments in the latest commits.
I realized that the test failures and the `AppConfig.httpURL` regression were actually caused by an accidental overwrite—I mistakenly covered the newer upstream changes with an outdated code snippet during development.

Here is a quick breakdown of the fixes:
	1.	Restored the filter in `ModelSettingsScreenVM` to fix the two failing Mac app tests.
	2.	Resolved the UI regression in `AccuracyBenchCard` by removing the nested button structure around the Toggle.
	3.	Reverted `WelcomeWindow` to use `AppConfig.httpURL(...)` for proper host/IPv6 handling.
	4.	Cleaned up the trailing whitespaces.

The tests should be green now. Looking forward to your further review!

jundot:
Thanks for the quick follow-up. I verified the updated PR locally, and the previous blockers are resolved: the VLM MTP draft-model tests pass again, the Accuracy Bench toggle no longer has the nested control issue, `WelcomeWindow` is back on `AppConfig.httpURL(...)`, and `git diff --check` is clean.

I also ran the mac app build and test suite locally, and both pass. This looks good to me now.

Thanks again for taking the time to make this contribution; this can be a meaningful UX improvement for the mac app, and I really appreciate the care you put into it.

PR #1905, opened 2026-06-17
fix(thinking): stop budget-forced close from leaking a second </think> (#1864)

### LINKED ISSUE #1864: thinking_budget: reasoning spills past the close at low budgets (intermittent, correlated with non-English / Cyrillic prompts)
## Summary

With a `thinking_budget` set, at low budgets the reasoning can **spill out of the `<think>` block and continue as part of the answer** instead of being closed at the budget. Split out from #1825, which tracked a *different* bug (the parameter being dropped on `/v1/completions`, fixed by #1844). This one is in `ThinkingBudgetProcessor` itself and reproduces independently, so it gets its own issue.

First reported by @socketa4techx7 in https://github.com/jundot/omlx/issues/1825#issuecomment-4684789604:

> setting thinking_budget to 128 just lets the thinking phase "spill out" of the thinking block and continue as part of the answer.
> Note: it seems to behave this way with non-english text/prompts. My example was cyrillic.

spill-out transition

## Environment (as reported)

- omlx 0.4.3, macOS, Apple Silicon (M1 Max)
- Model: `qwen3.6-35b-a3b-ud-mlx-4bit`, thinking enabled
- Trigger: `thinking_budget` ≈ 128, **non-English (Cyrillic)** prompt

## Why this is separate from the other thinking_budget issues

- **Not** the `/v1/completions` field-drop (#1825 → fixed by #1844): there the budget never reached the engine. Here the budget *is* applied, but the forced close does not hold.
- **Not** the concurrent row-misalignment (#1823 → #1824): this reproduces solo, no mixed batch needed.
- #1844 explicitly scoped this out: *"The spill-out behavior in the issue's second comment (thinking continuing past the close at low budgets) is in `ThinkingBudgetProcessor` and is not addressed here."*

## Likely area

`ThinkingBudgetProcessor`'s forced-close path — once the budget is hit it emits the close sequence (`\n` + `</think>` + `\n\n`) by logit masking, with a UTF-8 boundary wait. The non-English correlation points at the marker/detokenization boundary: for a multi-byte (Cyrillic) tail the close-token ids or the boundary wait may not line up, so the close is emitted-but-ignored or deferred and the reasoning bleeds into the answer.

Note: the soft-budget zone in #1812 changes *where* thinking closes naturally (it biases `</think>` earlier), but it does **not** claim to fix this hard-close overrun — the hard force at 100% is unchanged.

## What would help

A minimal repro on a current build: a short Cyrillic prompt with `thinking_budget: 128`, showing `reasoning_content` vs `content`, to confirm whether the close marker is (a) never emitted, (b) emitted but not detokenized at the boundary, or (c) emitted into `content` instead of closing the block. @socketa4techx7 if you can share the exact prompt that triggered the screenshot, that would be a great starting point.

richgoodson:
Fixes #1864.

## The bug

With a `thinking_budget` set, a low budget force-closes the `<think>` block mid-thought. The model, unaware it was cut off, keeps reasoning and then emits its **own** `</think>` later. Because the reasoning/answer split happens at the *first* (forced) close, everything after it lands in the answer: the model's continued deliberation, a duplicated answer, and a literal `</think>` tag visible in `content`.

## Repro

Driven live on `Qwen3.6-35B-A3B-4bit` at `thinking_budget=128`, temp 0, via `/v1/chat/completions` and in-process `stream_generate`:

- Cyrillic and English math prompts both produced `content` ending with a literal `</think>` followed by a duplicated answer.
- Two refinements to the original report:
  - It is **not** Cyrillic-specific. It reproduces in English too. The real trigger is "natural thinking length much greater than budget", which low budgets guarantee.
  - The "UTF-8 boundary wait" hypothesis is not the cause here: on a byte-level BPE tokenizer the piece lookup returns a `str`, so `_is_utf8_complete` always returns `True` and that path is inert. The failure is the double close.

## The fix (two layers)

1. **`ThinkingBudgetProcessor`**: after a *forced* close completes, mask the close token for the rest of generation so the model cannot emit a duplicate `</think>`. Cleared when the model legitimately re-enters a thinking block. Gated to **single-token** closes, where the token is a dedicated `</think>` sentinel that is always safe to ban. Multi-token closes decompose into ordinary subwords (e.g. `"</"`, `"think"`, `">"`) that recur in normal text, so masking them would corrupt the answer; those models rely on layer 2 instead.
2. **`extract_thinking`**: strip any orphan `</think>` from answer content as a backstop (covers multi-token-close models and any non-suppressed path). Implemented as a linear `str.replace` plus a `\n{3,}` collapse rather than a regex with a leading whitespace quantifier, which backtracks quadratically on untrusted model output (ReDoS).

EOS is never masked, so termination and `max_tokens` are unaffected. The streaming `ThinkingParser` already consumed stray closes, so no change was needed there.

## Verification

- After the fix, the literal-tag leak and the answer duplication are both gone across the repro prompts (Cyrillic math, English math, Cyrillic poem, budget=64 explain).
- New regression tests: re-close suppression after a forced close, no suppression after a natural close, suppression cleared on re-entry, multi-token close left untouched, orphan-strip on both `extract_thinking` branches, inline strip without stray newline, and a ReDoS guard (200k-char whitespace payload completes well under 1s; the pre-fix regex took ~120s).
- Full suite: 5849 passed, 22 skipped.

Note: reasoning that spills after a hard cut still goes to the answer (inherent to truncating the model mid-thought); this PR removes the literal tag and the duplication, not the spilled text itself. The strip also removes a `</think>` a model legitimately wrote in its answer when a thinking block is also present, matching the existing streaming parser behavior.

### DIFF
--- a/omlx/api/thinking.py
+++ b/omlx/api/thinking.py
@@ -28,6 +28,13 @@ _THINKING_PATTERN = re.compile(r'<think>(.*?)</think>', re.DOTALL)
 # Handle case where <think> is missing but </think> is present
 # (scheduler prepends <think>\n but the tag may be split)
 _THINKING_TAIL_PATTERN = re.compile(r'^(.*?)</think>', re.DOTALL)
+# Collapse the blank-line gap left after an orphan </think> tag is removed
+# (the forced close is written as ``\n</think>\n\n``).  Deliberately a simple
+# ``\n{3,}`` and NOT a pattern with a leading ``[ \t]*``/``\s*`` quantifier:
+# an unbounded whitespace prefix makes re.sub O(n^2) on a long run (quadratic
+# ReDoS), and model output is untrusted and can degenerate into whitespace
+# repetition. ``\n{3,}`` has no backtracking ambiguity and scans linearly.
+_COLLAPSE_BLANK_LINES = re.compile(r'\n{3,}')
 
 
 def _safe_tokenizer_attr(tokenizer, attr: str, default=None):
@@ -142,6 +149,26 @@ def prompt_opens_thinking(
     return True, think_tag
 
 
+def _strip_orphan_close_tags(content: str) -> str:
+    """Remove stray ``</think>`` tags left in answer content.
+
+    Called only after the thinking block(s) have already been extracted, so
+    any ``</think>`` still present is an orphan — most often the close the
+    model emits on its own after a budget-forced close (#1864).
+
+    Implemented as a linear ``str.replace`` plus a blank-line collapse rather
+    than a single regex: ``content`` is untrusted model output, and a regex
+    that tries to absorb the surrounding whitespace with a leading quantifier
+    backtracks quadratically on long whitespace runs (ReDoS).  Dropping the
+    literal tag turns the forced-close wrapper ``\\n</think>\\n\\n`` into a
+    ``\\n\\n\\n`` gap, which the collapse trims back to a paragraph break; a
+    bare inline ``</think>`` simply vanishes with no stray newline.
+    """
+    if '</think>' not in content:
+        return content
+    return _COLLAPSE_BLANK_LINES.sub('\n\n', content.replace('</think>', ''))
+
+
 def extract_thinking(text: str) -> Tuple[str, str]:
     """Extract thinking and content from complete text.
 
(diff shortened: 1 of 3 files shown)


guruswami-ai:
Confirming this fix from production use. We carried it in our internal deploy branch (0.4.4-final through 0.4.5.dev) on a 5× M3 Ultra 512GB fleet for ~2 weeks with no regressions.

We're now on vanilla v0.5.1, where we run Qwen3.6-35B-A3B with `thinking_budget_tokens: 8192` behind a load-balanced pool alias — exactly the budget-forced-close path #1864 describes — so without this patch we're exposed to the stray `</think>` leaking into `content` on every budget-clamped response.

Verified today: the patch still cherry-picks **clean onto v0.5.1** (50ee2464). Would be great to see this and #1864 picked up in the next merge window.

PR #1919, opened 2026-06-17
feat: add Brazilian Portuguese (pt-BR) i18n support

victor-torres:
## Summary

- Adds `omlx/admin/i18n/pt-BR.json` with full Brazilian Portuguese translations — all 769 i18n keys covered
- Registers `pt-BR` as a selectable language option in the admin dashboard Settings → Language panel
- Adds `"settings.language.pt-BR": "Português (Brasil)"` to all 8 existing locale files so the option label renders correctly regardless of the active language
- Updates README to mention Brazilian Portuguese in the supported admin UI languages

## Test plan

- [ ] Start the server and open `/admin`
- [ ] Go to Settings → Language, select **Português (Brasil)** and confirm the UI switches to Portuguese
- [ ] Verify all sections (Dashboard, Models, Settings, Logs, Benchmark, Chat) render in pt-BR without any raw key fallbacks
- [ ] Switch back to another language and confirm it works correctly


jundot:
Thanks for adding this. I checked the pt-BR locale against the English source, and it has the full key set in the same order with valid JSON, so it should not introduce raw-key fallbacks. The language selector and settings flow also match the existing i18n loading path. This looks good to me, and I'm going to merge it.

PR #1922, opened 2026-06-18
fix(cache): enforce SSD cache limit across model switches (#1915)

### LINKED ISSUE #1915: [BUG] The SSD cache usage exceeds the limit.
**Describe the bug**
- This issue may be caused by some bugs when switching between different models.

I set the cache disk to an external SSD solid-state disk. It works very well. Its capacity is 64G. I set an upper limit of 45G, but I often find that it uses 51G, Or the cache disk is simply filled up so that it cannot continue writing, As a result, the cache becomes invalid.

-  In the current situation, this function works normally only when not switching other models. However, this feature seems to be broken once you switch to another model. I had to manually clean the SSD cache for OMLX to function properly.

**To Reproduce**
1. Set cache path to external disk: `/Volumes/HIKSEMI/omlx_cache`
2. Set Storage limit `45GB` in WEB UI.
3. Use LLM1（Qwen3.6-35BA3B-4bit）, Fill up the SSD cache and occupy it to the maximum value.
4. Switch to another LLM2（Qwopus3.6-35B-A3B-v1-4bit-MTPLX-Optimized-Speed）, And do some work with that model.
5. At this point, you can observe that the SSD cache usage exceeds the maximum limit of 45GB and gradually increases.

**Expected behavior**
- The SSD cache usage will not continue to increase after reaching the maximum limit of 45GB (The actual maximum limit set).

 - macOS Version: 26.5.1
 - oMLX Version: v0.4.4

imi4u36d:
## Problem

SSD cache usage exceeds the configured limit when switching between models .

When switching models, each model creates a new `PagedSSDCacheManager` sharing the same SSD cache directory. Incompatible blocks from other models are left on disk but **not indexed**, making them invisible to the LRU eviction logic. This causes two issues:

1. **`_get_effective_max_size()` miscalculates the limit** — it uses `index.total_size + disk_free`, treating incompatible blocks' disk space as "available". The effective limit is inflated and eviction never triggers for those blocks.

2. **Incompatible blocks are never evicted** — `_enforce_size_limit_for_new_block()` only evicts indexed (compatible) blocks via `self._index.evict_until_size()`. Blocks from previous models accumulate unboundedly.

**Reproduction scenario from the issue:**
- 64GB SSD, 45GB cache limit configured
- Model A fills 30GB of cache → 30GB on disk
- Switch to Model B → index scans, total_size = 0 (Model A's blocks are incompatible)
- `disk_available = 0 + 34GB(free) = 34GB` → `effective_max = 33.6GB`
- Model B writes 33GB → **actual disk usage = 63GB**, far exceeding the 45GB limit

## Fix

Two changes in `omlx/cache/paged_ssd_cache.py`:

### 1. `_get_effective_max_size()` — use actual disk usage

```python
# Before (buggy):
disk_available = self._index.total_size + disk_free
disk_limit = int(disk_available * self._DISK_SAFE_RATIO)

# After (fixed):
disk_limit = int((usage.used + usage.free) * self._DISK_SAFE_RATIO)
```

Using `shutil.disk_usage().used + .free` accounts for **all** files in the cache directory, including incompatible blocks from other models that aren't in this manager's index.

### 2. `_scan_existing_files()` + new `_cleanup_incompatible_blocks()` — evict stale blocks

During startup scan, incompatible blocks are now collected. If the total on-disk cache (compatible + incompatible) exceeds the effective limit, incompatible blocks are evicted oldest-first to reclaim space.

## Tests

Updated 4 test cases in `test_paged_ssd_cache.py` to match the corrected disk space calculation. All 134 SSD cache tests pass.

Closes #1915

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -1332,9 +1332,16 @@ class PagedSSDCacheManager(CacheManager):
         """Scan cache directory for existing files and build the compatible index.
 
         Only blocks compatible with the currently loaded model/layout are
-        indexed. Incompatible blocks are left on disk so a shared SSD cache
+        indexed.  Incompatible blocks are left on disk so a shared SSD cache
         directory can safely serve multiple loaded models without one model's
         startup scan deleting another model's cache.
+
+        When the total disk usage of the cache directory exceeds the
+        configured limit, incompatible blocks are evicted (oldest first)
+        to reclaim space.  This prevents the runaway growth reported in
+        https://github.com/jundot/omlx/issues/1915 where switching models
+        causes the cache to exceed its configured limit because each model's
+        manager only tracks its own compatible blocks.
         """
         logger.info(f"Scanning SSD cache directory: {self._cache_dir}")
 
@@ -1343,6 +1350,8 @@ class PagedSSDCacheManager(CacheManager):
         skipped_incompatible = 0
         skipped_incompatible_bytes = 0
         errors = 0
+        # Collect incompatible blocks for potential cleanup (oldest first).
+        incompatible: list[PagedSSDBlockMetadata] = []
 
         for subdir in self.SUBDIR_CHARS:
             subdir_path = self._cache_dir / subdir
@@ -1358,6 +1367,7 @@ class PagedSSDCacheManager(CacheManager):
                     if not self._is_compatible_block(metadata):
                         skipped_incompatible += 1
                         skipped_incompatible_bytes += metadata.file_size
+                        incompatible.append(metadata)
                         continue
                     self._index.add(metadata)
                     indexed += 1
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for tracing the incompatible-block path and putting this PR together. It pointed at the right area of the cache stack.

I’m going to close this version rather than merge it because I ended up taking a different accounting approach. The main issue is that `usage.used + usage.free` is effectively filesystem capacity, not cache-directory usage, and startup cleanup only handles incompatible blocks that already exceed the configured limit. In the reported scenario, old-model cache can remain below the configured limit, then the new model can add its own cache and push the shared cache directory over the limit.

I handled this separately by tracking incompatible blocks for shared-budget accounting and eviction, while keeping them invisible to the current model’s lookup path. Thanks again for the investigation.

PR #1923, opened 2026-06-18
fix(responses): merge leading system messages in user_note_safe fallback

### LINKED ISSUE #1908: Bug: /v1/responses adapter prepends instructions without deduplicating system messages, triggering strict chat template validation
## Description

When Codex App Desktop sends requests via `/v1/responses` to oMLX, the request fails with `400: Chat template error: System message must be at the beginning.` on models with strict system-message ordering (e.g. Qwen3.6).

## oMLX Server Log

```
2026-06-17 11:48:58,242 - omlx.server - WARNING - POST /v1/responses -> 400: Chat template error: System message must be at the beginning.
2026-06-17 11:53:02,650 - omlx.server - WARNING - POST /v1/responses -> 400: Chat template error: System message must be at the beginning.
```

## Root Cause

In `dflash_mlx/server/responses_adapter.py` line 44:

```python
instructions = body.get("instructions")
if isinstance(instructions, str) and instructions:
    messages = [{"role": "system", "content": instructions}] + messages
```

When Codex App Desktop sends a request, the `input` array may already contain items with `role: "system"`. After the adapter prepends the `instructions` as another system message, the final messages array ends up with multiple system messages — and only the first one is at position 0.

After adapter conversion:
```
messages = [
    {"role": "system", "content": "from instructions"},
    {"role": "system", "content": "from input"},
    {"role": "user", "content": "..."}
]
```

Qwen3.6's `chat_template.jinja` (line 83-86) validates that system messages can only appear at the first position:
```jinja
{%- if message.role == "system" %}
    {%- if not loop.first %}
        {{- raise_exception('System message must be at the beginning.') }}
    {%- endif %}
{%- endif %}
```

This causes a `TemplateError` that propagates as a 400 response.

## Proposed Fix

The `responses_adapter.py` should handle system message deduplication when merging `instructions` with `input`. Options:

**Option A** — Remove existing system messages from `input` when `instructions` is provided, keeping only the instructions as system:
```python
if isinstance(instructions, str) and instructions:
    non_system_messages = [m for m in messages if m.get("role") != "system"]
    messages = [{"role": "system", "content": instructions}] + non_system_messages
```

**Option B** — Merge all system content together when `instructions` is present:
```python
if isinstance(instructions, str) and instructions:
    system_parts = [instructions]
    system_parts.extend(m["content"] for m in messages if m.get("role") == "system")
    non_system = [m for m in messages if m.get("role") != "system"]
    messages = [{"role": "system", "content": "\n\n".join(system_parts)}] + non_system
```

## Environment

- oMLX version: 0.4.4
- macOS: 26.5.1 (Apple Silicon)
- Model: Qwen3.6-35B-A3B-8bit (and any model with strict system message ordering in chat template)
- Client: Codex App Desktop (using /v1/responses endpoint)

imi4u36d:
## Problem

When Codex App Desktop sends `/v1/responses` requests with both `instructions` and a `system`/`developer` message in `input`, two leading system messages are produced. If the conversation also contains a mid-system message (e.g. via `previous_response_id` chain or multi-turn input), the `user_note_safe` downgrade path in `prepare_system_messages_for_template` preserves the leading system block as-is without merging.

Since `_merge_consecutive_roles` only merges `user`/`assistant` roles, the two consecutive system messages survive into the chat template. Strict templates like Qwen3.6 validate that system messages can only appear at the first position, causing:

```
POST /v1/responses -> 400: Chat template error: System message must be at the beginning.
```

## Root Cause

In `omlx/api/utils.py`, `_downgrade_mid_system_to_user_notes` preserves leading system blocks via `rewritten.extend(messages[start:i])` at line 601-602. The caller `unsupported_fallback` then only applies `_merge_consecutive_roles`, which doesn't touch system messages.

## Fix

Call `_merge_consecutive_system_messages` in `unsupported_fallback` before `_merge_consecutive_roles` when the `user_note_safe` path succeeds. This merges consecutive leading system messages into one, satisfying strict template validation.

## Reproduction

```python
from omlx.api.utils import prepare_system_messages_for_template

class Qwen36Tokenizer:
    def apply_chat_template(self, messages, **kw):
        for i, msg in enumerate(messages):
            if msg.get('role') == 'system' and i > 0:
                raise Exception('System message must be at the beginning.')
        return 'ok'

messages = [
    {"role": "system", "content": "You are a helpful assistant"},  # from instructions
    {"role": "system", "content": "Be concise"},                   # from input
    {"role": "user", "content": "Hello"},
    {"role": "system", "content": "Remember this"},                # mid-system
    {"role": "user", "content": "Continue"},
]

result = prepare_system_messages_for_template(
    messages, Qwen36Tokenizer(), unsupported_mid_system_policy="user_note_safe")
# Before fix: ['system', 'system', 'user'] → template error
# After fix:  ['system', 'user']           → OK
```

Closes #1908

### DIFF
--- a/omlx/api/utils.py
+++ b/omlx/api/utils.py
@@ -666,6 +666,10 @@ def prepare_system_messages_for_template(
         if unsupported_mid_system_policy == "user_note_safe":
             prepared = _downgrade_mid_system_to_user_notes(messages)
             if prepared is not None:
+                # _downgrade preserves leading system blocks as-is; merge
+                # consecutive system messages so strict templates (Qwen3.6+)
+                # that require a single leading system message don't fail.
+                prepared = _merge_consecutive_system_messages(prepared)
                 if merge_consecutive_roles:
                     prepared = _merge_consecutive_roles(prepared)
                 return prepared
--- a/tests/test_api_utils.py
+++ b/tests/test_api_utils.py
@@ -2369,6 +2369,44 @@ class TestPrepareSystemMessagesForTemplate:
         assert [m["role"] for m in result] == ["system", "user"]
         assert result[0]["content"] == "Runtime tip"
 
+    def test_user_note_policy_merges_leading_system_messages(self):
+        """Multiple leading system messages must be merged even when a
+        mid-system message triggers the user_note_safe downgrade path.
+
+        Regression test for: Codex App Desktop sends ``instructions`` plus a
+        system/developer message in ``input``, producing two leading system
+        messages.  When a later mid-system message triggers
+        ``_downgrade_mid_system_to_user_notes``, the leading block was
+        preserved as-is (two separate system messages), causing strict
+        templates like Qwen3.6 to reject with
+        "System message must be at the beginning."
+        """
+        messages = [
+            {"role": "system", "content": "You are a helpful assistant"},
+            {"role": "system", "content": "Be concise"},
+            {"role": "user", "content": "Hello"},
+            {"role": "system", "content": "Remember this"},
+            {"role": "user", "content": "Continue"},
+        ]
+
+        result = prepare_system_messages_for_template(
+            messages,
+            self.ErrorTokenizer(),
+            unsupported_mid_system_policy="user_note_safe",
+        )
+
+        # All leading system messages must be merged into one
+        assert result[0]["role"] == "system"
+        assert "You are a helpful assistant" in result[0]["content"]
+        assert "Be concise" in result[0]["content"]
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for the fix. The change is scoped to the user_note_safe fallback and uses the existing system-message merge path after mid-system notes are downgraded, so it addresses #1908 without changing the native-preserve path. I verified the API utility tests locally and CI is green. This looks good to me, and I'm going to merge it.

PR #1931, opened 2026-06-19
fix(cohere2_moe): re-escape literal control chars in streamed tool arguments

ttapper:
## Problem

When `Cohere2MoeOutputParserSession.finalize()` returns tool call arguments, it passes the accumulated Melody string through verbatim:

```python
"arguments": value["arguments"] or "{}",
```

The Melody filter streams model output character-by-character. If the model emits a literal control character (e.g. U+000A newline) inside a JSON string value instead of the `\n` escape sequence, the accumulated `arguments` string fails pydantic's strict JSON validation and the request is rejected:

```
1 validation error for FunctionCall
arguments
  Value error, arguments must be valid JSON, got parse error: Expecting ','
  delimiter: line 1 column 343 (char 342). ...
  Received: '{"path": "parse_logs.py", "edits": [{"oldText": "LOG_RE = re.compile(\n    r\'...'
```

This surfaces as a hard failure on any task where the model writes multi-line content in a tool argument (e.g. `edit_file` with multi-line code).

## Fix

Add `_reserialize_cohere_tool_arguments()` which runs the accumulated string through `json.loads(strict=False)` + `json.dumps()` before `finalize()` returns it. `strict=False` tolerates literal control characters inside strings; `json.dumps()` re-serializes them with proper escape sequences.

This mirrors the identical fix already applied to the XML fallback parser path in `ToolCallStreamFilter` (commit `64a1b8c`).

## Tests

Added `test_literal_newline_in_arguments_is_reescaped` to `TestCohere2MoeOutputParserSession`. All 20 tests pass.

### DIFF
--- a/omlx/adapter/output_parser.py
+++ b/omlx/adapter/output_parser.py
@@ -404,6 +404,19 @@ def _create_cohere2_moe_filter():
     return PyFilter(PyFilterOptions().cmd4().stream_tool_actions())
 
 
+def _reserialize_cohere_tool_arguments(args: str) -> str:
+    if not args:
+        return "{}"
+    try:
+        return json.dumps(
+            json.loads(args, strict=False),
+            ensure_ascii=False,
+            separators=(",", ":"),
+        )
+    except (json.JSONDecodeError, ValueError):
+        return args or "{}"
+
+
 class Cohere2MoeOutputParserSession:
     """Parser session for Cohere2 MoE / Command-style Melody output."""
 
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the fix. I verified that the Cohere2 MoE finalize path was still returning Melody arguments verbatim, and the non-strict parse plus re-serialization matches the existing raw-control-char recovery path. The targeted parser tests pass for me, so this looks good to me and I'm going to merge it.

Issue #1645, opened 2026-06-03
support for gemma-4-unified needed


ratio91:
Currently (using 0.4.1) I get:

```
omlx.engine_pool - WARNING - VLM loading failed for gemma-4-12B-mxfp8, falling back to LLM: Model type gemma4_unified not supported. Error: No module named 'mlx_vlm.speculative.drafters.gemma4_unified'

omlx.server - ERROR - POST /v1/chat/completions → 500 (unhandled): VLM load failed: Model type gemma4_unified not supported. Error: No module named 'mlx_vlm.speculative.drafters.gemma4_unified'; LLM fallback also failed: Model type gemma4_unified not supported.
```

iGeeky:
Hi,

I encountered the exact same error while using the model. I wanted to confirm that this is not an isolated case.

**My Environment:**
*   **omlx version:** `0.4.1 (1426)`
*   **Model:** `mlx-community/gemma-4-12B-it-4bit`

**Error Log:**
```text
VLM load failed: Model type gemma4_unified not supported. Error: No module named 'mlx_vlm.speculative.drafters.gemma4_unified'; LLM fallback also failed: Model type gemma4_unified not supported.
```

It seems like the `gemma4_unified` model type is not currently supported in this version of `mlx_vlm`. Could you please clarify if there is a known workaround or if this is a pending fix?

dapx:
It's fixed in the [last commit](https://github.com/jundot/omlx/commit/3d2fef58344fc286ae7f5cdf58c43e2694df0867), we just need a new release to download.

RonaldAtZalion:
Looking forward to this - excited to try 12B.

lekhnath:
waiting...

jundot:
@ratio91 @iGeeky @dapx @rbe4zalion @lekhnath I published `0.4.2.dev1` as a prerelease with the updated `mlx-vlm` pin for Gemma4 Unified (12B) support.

Release: https://github.com/jundot/omlx/releases/tag/v0.4.2.dev1

Please try one of the new DMGs and let me know if `gemma4_unified` still fails to load.

rgroli:
I loads the model now, but when trying to use it in a basic chat, it outputs: 

[...]

PR #1935, opened 2026-06-19
fix(oq): expose embed_audio on the VLM sanitize proxy for gemma4_unified

JimStenstrom:
Refs #1645 — this fixes the **oQ-quantization** path for `gemma4_unified`. It does *not* add the base mlx-vlm "model type not supported" support that #1645 also needs, so it's a partial address, not a close.

## Summary
- oQ's VLM sanitize proxy now exposes `embed_audio` alongside the existing `audio_tower`, so `gemma4_unified.sanitize()` no longer aborts mid-quant.
- Before this, oQ produced a quant that passed the structural scan but could not be loaded or served (`model.`-prefixed keys mlx-vlm rejects).

## Why (root cause)
`oq._build_model_sanitizer` builds a lightweight `_Proxy` and calls the model's real `Model.sanitize(proxy, weights)` during quantization. The gemma-4 family guards its audio weights on an instance attribute, but the **attribute name differs by arch**:

- `gemma4` reads `self.audio_tower` — already exposed on the proxy
- `gemma4_unified` reads `self.embed_audio` (mlx-vlm `models/gemma4_unified/gemma4_unified.py`, in `sanitize`) — **not** exposed

So for `gemma4_unified` the proxy raised `AttributeError`, the surrounding `try/except` swallowed it, and the entire sanitize pass was skipped. The quant was then written with raw checkpoint keys, which mlx-vlm refuses on load:

```
Received 1339 parameters not in model: model.vision_embedder.pos_norm.weight
```

This is the same shape as the recently-merged #1881 (expose `language_model` on this proxy for nested VLMs) — same proxy, sibling attribute.

## Changes
- `omlx/oq.py`: add `embed_audio = _AUDIO_SENTINEL` to the sanitize `_Proxy`. The sentinel is truthy iff the source `config` carries `audio_config`, so audio weights survive sanitize when present and are correctly dropped when absent — identical semantics to the existing `audio_tower` guard.
- `tests/test_oq.py`: regression test parametrized over both guard attributes (`audio_tower`, `embed_audio`), driving `_build_model_sanitizer` with a `gemma4_unified`-style config.

## Sibling-path sweep
Checked every VLM `Model.sanitize` in the pinned mlx-vlm for instance-attribute guards:
- gemma-4 family is now complete: `gemma4` → `self.audio_tower` (pre-existing), `gemma4_unified` → `self.embed_audio` (this fix).
- Other arches whose top-level `sanitize` reads instance attributes (`phi4mm`, `step3p7`, …) are pre-existing gaps, out of scope here. The fix is deliberately an explicit named attribute, **not** a blanket `__getattr__` sentinel, so those arches' `is None` guards keep behaving correctly.

## Tests
```
python -m pytest tests/test_oq.py -q   # 245 passed
```
The new test fails with `AttributeError: '_Proxy' object has no attribute 'embed_audio'` when the fix is reverted.

### End-to-end verification (live `omlx serve`)
An oQ-quantized `gemma-4-12B-it` (`gemma4_unified`, 7.2 GB) loads cleanly — **no "parameters not in model"** — and produces correct, non-blank output across every mode, including gemma-4's harmony-style thinking channels (`<|channel>thought … <channel|>` → `reasoning_content`, final → `content`):

| Mode | `content` | `reasoning_content` |
|---|---|---|
| Text, non-stream | ✓ | — |
| Vision (image description), non-stream | ✓ (no placeholder-token leak) | — |
| Streaming, no-think | ✓ | — |
| Thinking, non-stream | ✓ (not blank) | ✓ |
| Thinking + streaming | ✓ (not blank) | ✓ |

The vision path exercises the `embed_vision`/`vision_embedder` keys that were malformed before the fix; `content` is populated in all modes (i.e. no thinking-channel blanking).

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -2109,7 +2109,16 @@ def _build_model_sanitizer(config: dict, text_only: bool = False):
 
             def _vlm_sanitize(weights):
                 class _Proxy:
+                    # The audio-presence guard differs by arch: gemma4 checks
+                    # ``self.audio_tower``; gemma4_unified checks
+                    # ``self.embed_audio``. Expose BOTH (sentinel iff the source
+                    # config carries audio) so sanitize keeps the audio modality
+                    # for either. Missing ``embed_audio`` made gemma4_unified's
+                    # sanitize raise AttributeError, silently dropping the whole
+                    # sanitize pass → oQ shipped raw ``model.``-prefixed keys
+                    # that omlx could not load.
                     audio_tower = _AUDIO_SENTINEL
+                    embed_audio = _AUDIO_SENTINEL
 
                 proxy = _Proxy()
                 proxy.config = model_config
--- a/tests/test_oq.py
+++ b/tests/test_oq.py
@@ -3010,6 +3010,77 @@ class TestBuildModelSanitizerTextOnly:
 
 
 # =============================================================================
+# Test _vlm_sanitize proxy exposes the gemma-4 audio-guard attributes
+# =============================================================================
+
+
+class TestVlmSanitizeProxyAudioAttrs:
+    """oq's _vlm_sanitize _Proxy must expose BOTH audio-guard attributes the
+    gemma-4 family reads: gemma4 guards audio weights on ``self.audio_tower``,
+    gemma4_unified on ``self.embed_audio``. A proxy missing the attribute a
+    model's ``sanitize()`` reads raises AttributeError, aborting sanitize so oQ
+    ships raw ``model.``-prefixed VLM keys that mlx-vlm cannot load.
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for the focused fix. I checked this against the pinned mlx-vlm Gemma 4 paths: gemma4_unified.sanitize() reads self.embed_audio while gemma4 reads self.audio_tower, so exposing both on the oQ sanitize proxy is the right narrow fix.

The regression test covers both guard attributes, and the focused oQ tests pass on my side. I agree this is a partial fix for #1645 rather than full base mlx-vlm model-type support.

PR #1936, opened 2026-06-19
fix: cli honors bootstrap base-path if exists

bspaulding:
Fixes non-default install paths by making GlobalSettings.load() resolve the same base-path bootstrap file that the macOS app already uses, if exists.


jundot:
Thanks for the fix. I verified that this matches the macOS AppConfig base-path resolution order and keeps explicit --base-path precedence intact. The focused settings suite passes on the PR head, so this looks good to me and I will merge it.

PR #1937, opened 2026-06-19
fix(engine_core): reap output collectors orphaned by client disconnect

### LINKED ISSUE #1154: Dashboard shows requests as 'Generating...' after engine has cancelled/completed them
## Summary

The dashboard model card keeps requests listed as `Generating...` indefinitely, even though the engine logs show those requests were already cancelled or completed successfully. The server itself is healthy — no GPU is held, no slot is wedged, and new requests process normally — but the UI never reaps the stale entries.

## Environment

- oMLX (recent build, Mac DMG)
- Model: `Qwen2.5-Coder-3B-Instruct-4bit`
- Client: Zed editor (edit prediction / inline assist), which aggressively cancels in-flight HTTP requests as the user types or moves the cursor

## Reproduce

1. Load `Qwen2.5-Coder-3B-Instruct-4bit` (or any chat model) in oMLX.
2. Point Zed's prediction provider at the oMLX OpenAI-compatible endpoint and edit a file rapidly so Zed cancels predictions mid-flight.
3. Stop typing and watch the dashboard model card.

**Expected:** request list returns to `0 req` once cancellations/completions are processed.
**Actual:** dashboard shows e.g. `2 req` with `Generating...` entries (request short-IDs visible) that never clear, even though the engine logs confirm each one was either aborted or completed.

## Evidence (engine logs, same window where dashboard shows stuck `Generating...`)

```
2026-05-09 15:58:00,691 - omlx.engine_core - INFO - Request 3fb0144a-... cancelled, aborting
2026-05-09 15:58:00,764 - omlx.scheduler   - INFO - Prefill interrupted at 471/472 tokens: 1 request(s) aborted
2026-05-09 15:58:00,765 - omlx.scheduler   - INFO - Rescheduled 1 requests for re-prefill
2026-05-09 15:58:01,041 - omlx.engine_core - INFO - Request 63ba8073-... cancelled, aborting
2026-05-09 15:58:04,401 - omlx.server      - INFO - Completion: 128 tokens in 5.27s (24.3 tok/s)
2026-05-09 15:58:04,441 - omlx.server      - INFO - Completion: 128 tokens in 3.40s (37.7 tok/s)
2026-05-09 15:58:31,679 - omlx.server      - INFO - Completion: 128 tokens in 2.15s (59.5 tok/s)
2026-05-09 15:58:39,997 - omlx.engine_core - INFO - Request 9074b270-... cancelled, aborting
2026-05-09 15:58:42,739 - omlx.server      - INFO - Completion: 128 tokens in 4.60s (27.8 tok/s)
2026-05-09 15:58:42,883 - omlx.server      - INFO - Completion: 128 tokens in 2.16s (59.4 tok/s)
```

All requests reach a terminal state in the engine, yet the dashboard continues to display two as `Generating...`.

## Impact

- Cosmetic, but misleading — looks like the model is wedged when it isn't, which previously prompted a manual model unload (the only way to clear the dashboard).
- TTL-based unload eventually clears the display, but reloading a 1.7 GB model just to reset a UI counter is heavy.

## Likely cause

The dashboard's active-requests list isn't subscribing to (or isn't handling) the same terminal events the scheduler/engine emit when a request is cancelled or completes. Cancellation path in particular looks under-reported — the `cancelled, aborting` → `Rescheduled` → eventual `Completion` flow may emit a different event shape than a normal completion.

## Suggested fix

Ensure both paths — `engine_core`'s `cancelled, aborting` and `server`'s `Completion:` — fan out a `request_finished` (or equivalent) event that the dashboard's request-list component already listens for. A simple reaping pass on a short interval would also work as a backstop.

Happy to provide more logs or test a fix.

JimStenstrom:
Fixes #1154.

## Summary
A client that cancels an in-flight request (e.g. Zed, which aggressively cancels predictions) leaves its `RequestOutputCollector` in `_output_collectors` until GC, so the dashboard shows the request as "Generating…" indefinitely even though the engine already aborted/completed it. This adds a grace-gated reaper to drop such orphans, plus guards that keep every consumer path safe.

## Why (root cause)
When a client disconnects mid-stream, Python's `async for` doesn't propagate `aclose()` into the inner `stream_outputs()` generator, so its cleanup `finally` — the only path that removes the collector — runs only at non-deterministic GC time. The dashboard's active count is `len(self._output_collectors)`, so the request lingers as "Generating…" until then (GPU/slot are already freed; it's purely stale tracking). The reporter suggested exactly a periodic reaping pass as a backstop.

## Changes
- **Grace-gated reaper.** `_mark_request_finished()` stamps a finish time on every terminal event; the engine loop calls `_reap_orphaned_collectors()` (throttled ~1/s, 5 s grace) to drop tracking for requests finished past the window. It is **pop-only** — never `collector.clear()` — so a slow-but-live consumer that holds its own collector reference is unaffected (an over-eager `clear()` is what made the earlier `_delayed_cleanup()` approach race).
- **add_request guard.** Aborts the partial scheduler insert (deferred + idempotent) and cleans up if the caller is cancelled before streaming begins — the disconnect-before-generation phantom the reaper can't see (never stamped `_finished_at`).
- **generate() symmetry.** The non-streaming path now captures its collector reference **before** awaiting, mirroring `stream_outputs()`. Without it, the post-await re-fetch could race the reaper and raise on an already-completed request if the coroutine was starved past the grace window under load.
- **Consistency.** `close()` also clears `_finished_at`; the partial-engine (`__new__`) regression test in `tests/test_engine_preflight.py` now sets `_finished_at`, since `_cleanup_request` manages it.

## Safety / sibling sweep
- The load-bearing invariant is pop-only + held-reference; both consumer paths (`stream_outputs`, `generate`) hold their reference before awaiting, so dropping the dict entry can never truncate a live consumer.
- VLM / batched / DFlash engines consume the same `EngineCore` API (they read the inner `_output_collectors` through the shared methods), so the central fix covers them — no per-engine special-casing.
- The dashboard count and `has_active_requests()` read the same dict; after a disconnect they now go stale-True for at most ~grace+throttle (~6 s) instead of until GC — a strict improvement over the status quo.
- **Out of scope (named):** `_delayed_cleanup()` is pre-existing dead code on `main` (no callers) — left for a separate cleanup rather than bundled here. The reaper grace is a fixed 5 s (no new user tunable); the `generate()` capture makes its exact value non-safety-critical.

## Tests
```
python -m pytest tests/test_engine_core.py tests/test_engine_preflight.py -q   # 92 passed
```
- New `TestOrphanedCollectorReaping` (reaps-only-stale, stamp-once, cleanup-clears-stamp, generate-drains-via-held-reference-if-reaped) and `test_add_request_cleans_up_if_scheduler_insert_fails`, exercising the real reaper/mark/cleanup methods.
- Full smoke (`pytest -m "not slow and not integration"`): **5935 passed**.
- Revert-proof: the reaper tests fail on `main` (`AttributeError: '_finished_at'`); the `generate()` test fails with `RuntimeError: No collector for request …` when the early capture is reverted.

### DIFF
--- a/omlx/engine_core.py
+++ b/omlx/engine_core.py
@@ -234,6 +234,13 @@ class EngineCore:
         self._output_collectors: Dict[str, RequestOutputCollector] = {}
         self._stream_states: Dict[str, RequestStreamState] = {}
         self._finished_events: Dict[str, asyncio.Event] = {}
+        # Finish timestamps for orphan-collector reaping (#1154).
+        # Normally a consumer drains and removes its own collector, but if the
+        # client disconnects mid-stream the SSE generator chain is abandoned and
+        # its cleanup finally only runs at GC time, so the collector lingers and
+        # the dashboard keeps showing the request as "Generating".
+        self._finished_at: Dict[str, float] = {}
+        self._last_reap = 0.0
 
         # Engine state
         self._running = False
@@ -351,6 +358,12 @@ class EngineCore:
 
         while self._running:
             try:
+                # Sweep collectors orphaned by client disconnects (throttled).
+                now = time.monotonic()
+                if now - self._last_reap >= 1.0:
+                    self._last_reap = now
+                    self._reap_orphaned_collectors(now)
+
                 if self.scheduler.has_requests():
                     step_outputs = await loop.run_in_executor(
                         self._mlx_executor, self._step_burst
@@ -397,11 +410,11 @@ class EngineCore:
                                         state.mark_sent(req_output.completion_tokens)
 
                             if req_output.finished:
-                                event = events.get(rid)
-                                if event:
-                                    event.set()
-                                # Note: cleanup is handled by stream_outputs() finally block
-                                # _delayed_cleanup() was causing double cleanup race condition
+                                self._mark_request_finished(rid)
+                                # Cleanup normally happens in the consumer
+                                # (stream_outputs()/generate()); collectors left
+                                # behind by a disconnected client are swept by
+                                # _reap_orphaned_collectors() via _finished_at.
 
                     if distributed:
                         # Always yield to prevent event loop starvation.
@@ -485,9 +498,7 @@ class EngineCore:
                                 error=str(e),
                             )
                         )
-                    event = self._finished_events.get(rid)
-                    if event:
-                        event.set()
+                    self._mark_request_finished(rid)
                 await asyncio.sleep(0.1)
 
     async def add_request(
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for the fix. The reaper approach looks sound to me: it clears stale dashboard/TTL tracking without clearing a collector that a live consumer may still hold, and the generate() change closes the race with the pop-only cleanup.

This looks good to me, and I'm going to merge it. I'll fold the small cleanup into a follow-up commit.

Issue #1866, opened 2026-06-13
0.3.8 and 0.4.4: cache hit rate decreased from 95% to 51%.


TLin-Y:
**Describe the bug**
On M1Pro 32GB with a clean disk (**no existing cache**), in the same PI session (asking the same questions), Omlx 0.3.8 achieves an almost perfect cache hit rate of 95%, but in the newest version 0.4.4 it drops to 51%. 

This typically occurs during PI code review: 
-  whenever a long context is included in the previous output, if I ask a follow-up question, Omlx 0.4.4 will re filling **everything**.
- I've ensured both versions use exactly the same setting configs (.json), and I also verified by **redownloading** the same model in both versions from another publisher in order to exclude a setting mismatch. (repo level diff so Omlx reinit a new dir/setting)
- I've tried **TurboQuant** on/off with empty cache dir; same issue.

**To Reproduce**
0. sudo sysctl iogpu.wiredlimitmb=29536, which helps prevent memory guard problems. (and raise limit (29GB) in Omlx)
1. MacOS 18 or 26/27
2. Clean install Omlx 0.3.8 vs 0.4.4
3. Qwen3.6-35B-A3B-oQ4-fp16 default setting
4. on latest 0.79.3 PI CLI, ask same questions: 
    - clone omlx git repo to cur dir
    - I've cache hit issue, and I found 0.3.8 has way better cache hit (95%) than the latest version (50%), can you check the git history find the root cause?
    - Few more background, my test without mem-guards, in empty cache dir, and tried both turboQ on/off, from what you found, which part maybe related to this?
   
0.3.8 log:
~/Documents/coding/omlx (main)
↑100k ↓21k R1.1M **CH95.3%** 56.9%/100k (auto)                                     Qwen3.6-35B-A3B-oQ4-fp16

0.4.4 log:
~/Documents/coding/omlx (main)
↑103k ↓22k R1.2M **CH50.3%** 66.9%/100k (auto)                                     Qwen3.6-35B-A3B-oQ4-fp16

**Expected behavior**
Omlx shouldn't get cache miss with almost same inputs.

**Additional context**
LLM generated report: 

Based on my thorough investigation of the git history between v0.3.8 and   
 v0.4.4.dev1, I've identified the root causes of the cache hit rate         
 regression. Here's my analysis:                                            
                                                                            
 Root Cause Analysis: Cache Hit Rate Drop (95% → 51%)                       
                                                                            
[...]

Issue #1128, opened 2026-05-09
ds v4 frequent cache corruption


pritamghanghas:
**Describe the bug**
I see logs where it shows frequent cache corruption and client never sees the generated stream when this happens.

**To Reproduce**
1. Load the deepseek v4-flash q5 model from mlx-community
2. point an agent, in this case hermes-agent.
3. goes into cache corruption fairly quicly.

**Expected behavior**
cache shouldn't get corrupted this frequently

**Screenshots**

2026-05-08 17:21:06,754 - omlx.scheduler - INFO - Using boundary cache snapshot for c9c18b32-f848-496e-b6f8-93346889bfb0: storing 24576/25319 tokens (skipping trailing partial block, 47 intermediate snapshots)
2026-05-08 17:21:06,757 - omlx.scheduler - INFO - Cache phase timings: boundary_capture_extract=0.5ms/1, boundary_capture_sync=2.8ms/1, cleanup_finished_sync=0.1ms/2, store_cache_main_eval=1.2ms/2, store_cache_main_prep=3.9ms/2
2026-05-08 17:21:07,020 - omlx.cache.prefix_cache - INFO - Partial cache reconstruction: 4/5 blocks, 2048 tokens
2026-05-08 17:23:15,798 - omlx.scheduler - INFO - Using boundary cache snapshot for 30cff23e-b067-440a-9d04-2061d0d8556a: storing 21504/22069 tokens (skipping trailing partial block, 17 intermediate snapshots)
2026-05-08 17:23:15,801 - omlx.scheduler - INFO - Cache phase timings: boundary_capture_extract=1.2ms/2, boundary_capture_sync=6.0ms/2, cleanup_finished_sync=0.1ms/3, store_cache_main_eval=1.3ms/3, store_cache_main_prep=5.5ms/3
2026-05-08 17:23:32,546 - omlx.scheduler - INFO - Using boundary cache snapshot for 6cccf92b-b1be-43ec-abd4-0142069ed3e7: storing 29184/30052 tokens (skipping trailing partial block, 52 intermediate snapshots)
2026-05-08 17:23:32,549 - omlx.scheduler - INFO - Cache phase timings: boundary_capture_extract=1.8ms/3, boundary_capture_sync=8.1ms/3, cleanup_finished_sync=0.1ms/4, store_cache_main_eval=1.4ms/4, store_cache_main_prep=6.3ms/4
[...]

PR #1939, opened 2026-06-20
fix(cache): serialize composite nstate elements (hybrid CacheList) to SSD prefix cache

apcooley:
## Summary

Hybrid-MLA models build each layer as `CacheList(RotatingKVCache, PoolingCache, PoolingCache)`. On the SSD-store path such a layer reaches `PagedSSDCacheManager._store_nstate_elements` as an `__nstate__` marker whose **elements are not flat `mx.array` values** but composite sub-states:

- a 2-tuple `(keys, values)` (the rotating sub-state), and
- a nested `('__nstate__', class_name, [None, None, pooled])` marker (the pooling sub-state).

`_store_nstate_elements` assumed every element was an `mx.array` and passed it straight to `_extract_tensor_bytes`, which raised:

```
ERROR - Failed to prepare block for SSD cache: 'tuple' object has no attribute 'dtype'
```

That fails the **whole block save**, and via the caller's break-on-failure, every later block too — so the prefix cache silently stops persisting. Over a growing multi-turn conversation the hit rate bleeds from ~97% down to ~74%, and warm turns fall back to full re-prefill (95–124 s/turn on an M2 Mac Pro with DeepSeek-V4-Flash-3bit-DQ).

## Fix

Route composite elements through the same recursion the serializer already uses for `__cache_list__`:

- **store**: a nested `__nstate__` element records `{key}_nested="nstate"` (+ class name) and recurses; a bare tuple/list records `{key}_nested="tuple"` and recurses. No flat tensor is written, so an *older* reader hits its existing `Missing {key} in arrays` skip path (forward/backward safe). A genuinely unsupported (non-array, non-composite) element now raises a clear `TypeError` the caller catches and skips cleanly, instead of an opaque `.dtype` `AttributeError`.
- **load**: a `{key}_nested` marker recurses and restores the original shape — `"tuple"` rebuilds a bare tuple, `"nstate"` preserves the full `('__nstate__', class_name, elements)` wrapper as-is.
- `_maybe_unwrap_legacy` now only collapses the legacy plain `(keys, values)` pair when neither element is itself composite, so a **length-2 nested marker keeps its wrapper and class_name** (the heuristic-collision case).

Additive and fail-closed: **no cache-format-version bump**, existing v2/v3 caches stay readable, and any malformed nested marker returns `None` (block skipped) rather than reconstructing a corrupt cache.

## Tests

`tests/test_nested_nstate_serialization.py` (6 tests) pins the contract via real `save_block` → `load_block` file I/O: tuple-of-arrays element, nested `__nstate__` element with `None` sub-elements, the length-2 marker-stays-a-marker regression, zero-size (MLA empty-values) array, the flat-N-tuple guard, and format-version-unchanged. The existing `tests/test_paged_ssd_cache.py` (134 tests) stays green.

## Live validation (DeepSeek-V4-Flash-3bit-DQ, M2 Mac Pro)

A 12-turn, 16k→40k-token, thinking-ON flood:

| | before | after |
|---|---|---|
| cache hit, warm turns | 97% → **74%** (degrades) | **97–98% flat** |
| warm-turn wall | 95–124 s (re-prefill) | **7–13 s** |
| serializer errors in window | 132 | **0** |
| blocks persisted to SSD | aborts on first crash | **89, zero errors** |

## Related issues

- #1128 (ds v4 frequent cache corruption) — same model/agent setup; this is the root cause behind the "cache corruption, client never sees the stream" symptom there.
- #1866 (cache hit rate 95% → 51%) — same degradation signature; likely the same store-path abort for that model's layer shape.
- #1106 (closed; `_extract_tensor_bytes` crash on a hybrid model) — prior report of the same call site, distinct (GC-race) hypothesis; this PR addresses the non-array-element variant.

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -1759,7 +1759,12 @@ class PagedSSDCacheManager(CacheManager):
             def _store_nstate_elements(prefix: str, elements):
                 """Write N elements as ``{prefix}_state_{k}`` keys with a
                 ``{prefix}_state_count`` count marker. Zero-dim shapes are
-                preserved via ``{prefix}_state_{k}_zero_dim``."""
+                preserved via ``{prefix}_state_{k}_zero_dim``. Composite
+                elements (a bare tuple/list or a nested ``__nstate__``
+                marker) recurse under a ``{elem_key}`` sub-prefix and record
+                a ``{elem_key}_nested`` marker; the flat ``{elem_key}``
+                tensor is deliberately omitted so an older reader hits its
+                ``Missing {elem_key} in arrays`` path and skips the block."""
                 cache_list_meta[f"{prefix}_state_count"] = str(len(elements))
                 for k, elem in enumerate(elements):
                     elem_key = f"{prefix}_state_{k}"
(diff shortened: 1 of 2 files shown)


apcooley:
Fixes #1940.

jundot:
Thanks for the detailed repro and tests. I verified the base failure on the composite nstate shape and the PR keeps the nested CacheList/PoolingCache state intact through the SSD round trip without changing the cache format version. The adjacent paged SSD, N-tuple handler, and GLM MoE DSA tests are green locally. This looks good to me, and I'm going to merge it.

Issue #1958, opened 2026-06-22
model_settings.json does not respect symbolic links on Mac, but settings.json does!


endlesslycurious:
oMLX is a great tool; I’ve found it very easy to setup and configure for running models on a dedicated Mac. Thanks for taking the time to create this!

**Describe the bug**
Unlike `~/.omlx/settings.json` when `~/.omlx/model_settings.json` is saved by oMLX the file is replaced irrespective of if the file was a symbolic or hard link created by `ln`.

I keep my configs in a git repo and use `ln` to link the files to where apps expect them to be e.g. `~/.omlx/settings.json`. This usually just works for most apps e.g. Zed, NeoVim, zsh, homebrew etc.

**To Reproduce**
1. On Mac, create a link `ln -f <src> ~/.omlx/settings.json`.
2. Verify the link with `ls -li <src-dir> ~/.omlx` the inode numbers (first column) should be the same.
3. Update a model setting in oMLX e.g. toggle thinking to On then save, to get a write to `~/.omlx/model_settings.json`.
4. Notice `~/.omlx/model_settings.json` is now an actual file, the link was removed and the linked file <src> was not updated.

**Expected behavior**
Like updates to global settings which writes to `~/.omlx/settings.json` and respects links, updates to model settings should not be removing links and the linked file should be updated.

**Screenshots**
Link setup correctly, note inode for `model_settings.json` is `467193` in both places and `settings.json` is `467194`.

```bash
> ls -li ~/.omlx ./config
./config:
total 24
467193 -rw-r--r--  2 bender  staff  4883 Jun 21 15:58 model_settings.json
467194 -rw-r--r--@ 2 bender  staff  2518 Jun 21 15:57 settings.json

/Users/bender/.omlx:
total 32
467193 -rw-r--r--  2 bender  staff  4883 Jun 21 15:58 model_settings.json
467194 -rw-r--r--@ 2 bender  staff  2518 Jun 21 15:57 settings.json
511639 -rw-r--r--  1 bender  staff   201 Jun 20 19:13 stats.json
```

Updated a model's settings then saved, updated a global setting then saved and ran the same `ls` command, `model_settings.json` inodes don't match as it's now a file in `~/.omlx` but `settings.josn` inode does still match (`467194`) as its still a symlink in `~/.omlx`.

```bash
> ls -li ~/.omlx ./config
./config:
total 24
467193 -rw-r--r--  1 bender  staff  4883 Jun 21 15:58 model_settings.json
[...]

PR #1961, opened 2026-06-22
fix: allow uppercase letters in profile and template names

### LINKED ISSUE #1957: Relax restrictions on profile names & Better error UI location
No uppercase in profile names? ¿Por que? This seems unnecessarily restrictive.

`HTTP 400: {"detail": "Invalid profile/template name: 'Gemma-4-OCR-32k'. Must match ^[a-z0-9][a-z0-9_-10,31)$"}`

Also — When this occurs, the Save UI is at the top but the error message appears at the bottom of the window. It is easy to miss that the creation/naming failed. (Better than failing silently, of course.)

MacOS v0.4.4 - Swift UI

Ramcode64:
Fixes #1957 (partial — the Python/API side)

## Problem

Profile names like `'Gemma-4-OCR-32k'` were rejected with `HTTP 400`:

```
Invalid profile/template name: 'Gemma-4-OCR-32k'. Must match ^[a-z0-9][a-z0-9_-]{0,31}$
```

The `name` field is an internal identifier (dict key + URL path parameter). There is no technical reason to restrict it to lowercase — that restriction only makes sense for `api_name`, which is exposed in OpenAI-compatible model IDs.

## Fix

Split `validate_profile_name` into two functions:

| Function | Regex | Used for |
|---|---|---|
| `validate_profile_name` | `^[a-zA-Z0-9][a-zA-Z0-9_-]{0,31}$` | internal `name` / `new_name` fields |
| `validate_profile_api_name` | `^[a-z0-9][a-z0-9_-]{0,31}$` | `api_name` (exposed in model IDs, must stay lowercase) |

Update the 4 `api_name` call sites in `model_settings.py` to use the stricter `validate_profile_api_name`. All `name`/`new_name` call sites continue using `validate_profile_name`, which now accepts uppercase.

## Testing

```
✓ 'Gemma-4-OCR-32k'  as profile name → pass (was failing before)
✓ 'Gemma-4-OCR-32k'  as api_name     → fail (uppercase still rejected for api slugs)
✓ 'gemma-4-ocr-32k'  as profile name → pass
✓ 'gemma-4-ocr-32k'  as api_name     → pass
✓ 'Has Space'         as profile name → fail (spaces still invalid)
✓ 'Valid-Name_123'    as profile name → pass
✓ empty string        as profile name → fail
✓ 33-char string      as profile name → fail (length limit enforced)
```

## Note on #1957's second request (error UI location)

The request to show the validation error near the input field instead of at the bottom of the window is a SwiftUI change and not addressed here.

### DIFF
--- a/omlx/model_profiles.py
+++ b/omlx/model_profiles.py
@@ -181,19 +181,37 @@ class InvalidProfileNameError(ValueError):
     """Raised when a profile or template name fails validation."""
 
 
-_NAME_RE = re.compile(r"^[a-z0-9][a-z0-9_-]{0,31}$")
+# Used for API-exposed names (model IDs, api_name) — must be lowercase slugs.
+_API_NAME_RE = re.compile(r"^[a-z0-9][a-z0-9_-]{0,31}$")
+
+# Used for internal profile/template names — allows uppercase so users can
+# create names like "Gemma-4-OCR-32k" without hitting a 400 error.
+_NAME_RE = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9_-]{0,31}$")
 
 
 def validate_profile_name(name: str) -> None:
-    """Raise InvalidProfileNameError if ``name`` is not a valid slug.
+    """Raise InvalidProfileNameError if ``name`` is not a valid profile identifier.
 
-    Valid: lowercase letters/digits, underscores, dashes. Must start with
-    a letter or digit. 1-32 characters.
+    Valid: letters (upper or lower), digits, underscores, dashes. Must start
+    with a letter or digit. 1-32 characters.
     """
     if not isinstance(name, str) or not _NAME_RE.match(name):
         raise InvalidProfileNameError(
             f"Invalid profile/template name: {name!r}. "
-            f"Must match ^[a-z0-9][a-z0-9_-]{{0,31}}$"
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the PR. The profile name / api_name split looks correct to me, and I verified that an uppercase internal profile name now produces a lowercase API slug while uppercase api_name is still rejected.

I do not want to merge this as-is because the included model_settings.json save change removes the previous temp-file replace behavior. That fixes link preservation, but it also means a crash or write/serialization failure can leave the existing settings file truncated. Could you split the #1958 save-path change into a separate PR, or add a version that preserves the existing-file safety tradeoff explicitly with tests?

Ramcode64:
Done — I've split the changes:

- **PR #1961** (this PR) now only contains the `model_profiles.py` split (`validate_profile_name` for internal names, `validate_profile_api_name` for API slugs) and the four call-site updates in `model_settings.py`. The save-path change has been removed.
- **PR #1960** has been updated with a safer two-phase write: serialize to string first (so any encoding error surfaces before touching the filesystem), then write to a `.tmp` file (crash-safe backup), then copy the content into the original path in-place (inode preserved → links remain valid). The `.tmp` file is deleted on success and kept as a recovery backup if the in-place write fails. Tests for symlink preservation, hard-link preservation, temp-file cleanup, and serialization-error safety have been added to `tests/test_model_settings.py`.

PR #1964, opened 2026-06-22
fix(reranker): recover scoring frame from chat template, gate built-in fallback to Qwen3

contrapuntal:
## Summary

Qwen3-Reranker models fail to load on `/v1/rerank` for MLX-converted repos like `mlx-community/Qwen3-Reranker-*`. The failure is at load time, so the model never serves a request — every call returns HTTP 500.

## Why (root cause)

To score a query/document pair, the loader needs the static text that wraps the document — a `prefix` and a `suffix`, together the scoring **frame**. It found these by rendering the chat template with a sentinel standing in for the content, then splitting on the sentinel — which only works if the template echoes the content back verbatim.

MLX-converted Qwen3-Reranker repos ship a role-specific template that renders `query`/`document` roles and ignores a plain `user` message. The sentinel is never emitted, the split fails, and load aborts:

```
ValueError: Chat template produced unexpected format; could not split on sentinel
```

The official `Qwen/Qwen3-Reranker-*` template echoes content and is unaffected.

## Reproduction

No weights needed — the tokenizer alone shows the dropped sentinel that the old loader split on:

```python
from transformers import AutoTokenizer

tok = AutoTokenizer.from_pretrained("mlx-community/Qwen3-Reranker-0.6B-4bit")
SENTINEL = "<<__CONTENT_SENTINEL__>>"
msgs = [
    {"role": "system", "content": "Judge whether the Document meets the requirements..."},
    {"role": "user", "content": SENTINEL},
]
rendered = tok.apply_chat_template(msgs, tokenize=False, add_generation_prompt=True)
print(SENTINEL in rendered)           # False -> sentinel dropped
print(len(rendered.split(SENTINEL)))  # 1     -> old loader raises here
```

## Changes

The loader now derives the frame from the chat template when it can, and uses a guarded built-in Qwen fallback when it can't:

- **Use the template** when it preserves the sentinel (unchanged for the official Qwen template).
- **Fall back to the built-in Qwen frame** when the template is missing, fails to render, or drops the content — the case that used to be fatal.
- **Gate the fallback to `Qwen3ForCausalLM`.** The frame is Qwen-specific; other architectures raise a clear error instead of being scored through a mismatched frame.
- **Validate the ChatML control tokens** map to single ids, so a tokenizer that doesn't know them fails loudly instead of scoring on mis-tokenized text.
- **Don't duplicate the `<think>` prefill** if the template already emitted one.
- **Narrow error handling:** skip rendering when no template is set; let unexpected errors (e.g. `MemoryError`) propagate.

## Tests

```
python -m pytest tests/test_reranker_causal_lm.py -q   # 31 passed
```

Covers each path: template preserved, content dropped, no template, render error, non-Qwen architecture, already-present think prefill, and unknown ChatML tokens.

## Notes for reviewers

The built-in frame was checked against the published Qwen3-Reranker MLX recipe and the `mlx-community/Qwen3-Reranker-0.6B-4bit` model card, and verified end-to-end on real weights (`Qwen3-Reranker-0.6B`): the model loads via the fallback and scores a relevant document at 0.997 vs 0.0 for an irrelevant one. The fallback only rescues the architecture it was validated against; everything else still fails fast.

### DIFF
--- a/omlx/models/reranker.py
+++ b/omlx/models/reranker.py
@@ -296,26 +296,90 @@ class MLXRerankerModel:
                 "This model may not be a compatible CausalLM reranker."
             )
 
-        # Pre-compute prefix and suffix tokens for the prompt template.
-        # Use apply_chat_template() for portability across tokenizer formats,
-        # then split on a sentinel to extract prefix/suffix boundaries.
+        # Pre-compute the static prompt prefix/suffix (the document content is
+        # spliced between them at scoring time). The Qwen3-Reranker frame opens a
+        # user turn, then prefills an empty think block before the yes/no logit:
+        #   prefix = "<|im_start|>system\n{SYS}<|im_end|>\n<|im_start|>user\n"
+        #   suffix = "<|im_end|>\n<|im_start|>assistant\n<think>\n\n</think>\n\n"
+        # We first try to recover the frame from the tokenizer's chat template, so
+        # a model shipping its own system prompt or framing is honored. We fall
+        # back to the built-in Qwen frame only when the template cannot provide it
+        # (absent, unrenderable, or role-specific templates that drop the user
+        # content, as shipped by MLX-converted Qwen3-Reranker repos).
         _SENTINEL = "<<__CONTENT_SENTINEL__>>"
-        messages = [
-            {"role": "system", "content": self._CAUSAL_LM_SYSTEM_PROMPT},
-            {"role": "user", "content": _SENTINEL},
-        ]
-        template_str = tokenizer.apply_chat_template(
-            messages, tokenize=False, add_generation_prompt=True
-        )
-        parts = template_str.split(_SENTINEL)
-        if len(parts) != 2:
-            raise ValueError(
-                f"Chat template produced unexpected format; "
-                f"could not split on sentinel. Template: {template_str!r}"
+        _THINK_PREFILL = "<think>\n\n</think>\n\n"
+
+        prefix = suffix = None
+        if getattr(tokenizer, "chat_template", None):
+            messages = [
+                {"role": "system", "content": self._CAUSAL_LM_SYSTEM_PROMPT},
+                {"role": "user", "content": _SENTINEL},
+            ]
+            try:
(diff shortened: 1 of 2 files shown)


shyuan:
Cross-referencing #2147, which addressed the same root cause from a different angle and has since been merged (I hadn't noticed the merge when I first posted this — apologies). Root-cause analysis with the upstream timeline is in #2146: `Qwen/Qwen3-Reranker-0.6B` added the reranker-native `chat_template.jinja` in its 2026-04-16 sentence-transformers update, so MLX conversions made after that date inherit it, while earlier conversions (e.g. `Qwen3-Reranker-0.6B-mxfp8`) still carry the standard ChatML template and load fine.

Quick comparison of the two approaches:

- **This PR** falls back to a built-in Qwen scoring frame, gated to `Qwen3ForCausalLM` — it also rescues models whose chat template is missing or broken entirely.
- **#2147** derives the frame from the model's own reranker-native template (per-slot sentinels split around the `<Instruct>/<Query>/<Document>` block) — architecture-agnostic, and for the Qwen3-Reranker family both template shapes yield byte-identical prefix/suffix tokens.

Credit where due: two hardening details were first caught by this PR — failing fast when `chat_template` is `None`, and not duplicating the `<think>` prefill when the template already emits one. They are now proposed as follow-up #2196 on top of the merged implementation. Disposition of this PR is of course up to the maintainers and the author.

contrapuntal:
Closing as superseded. #2147 (merged) addresses the same root cause from a more general angle — deriving the frame from the model's own reranker-native template, architecture-agnostic and byte-identical for the Qwen3-Reranker family — and the two hardening details from this PR (fail fast on missing `chat_template`, don't duplicate the `<think>` prefill) landed via #2196. Thanks @shyuan for the cross-reference and for carrying those forward.

The one behavior not carried over is the built-in Qwen frame rescuing models whose chat template is missing or broken entirely; #2196 deliberately chose fail-fast there, so I'm not re-litigating it. Happy to open a separate issue if that case turns up in practice.

PR #1969, opened 2026-06-22
fix(settings): preserve symlinked config files on save (#1958)

richgoodson:
## Problem

`~/.omlx/model_settings.json` (and `model_profiles.json`, `global_templates.json`) are written with a temp-file + `os.replace` rename. `os.replace` swaps the directory entry to a **new inode**, which breaks links a user has set up by keeping config in a git repo and `ln`-ing it into `~/.omlx/`:

- **symlink** (`ln -s`): the rename clobbers the link with a real file.
- **hard link** (`ln`): the other name (the repo file) keeps pointing at the old inode, so edits never reach it.

`settings.json` was unaffected only because its writer writes in place. Reported in #1958, whose repro uses `ln -f` (a hard link) and `ls -li` to verify the inode.

## Fix

Route all three writers through a shared `_atomic_write_json` helper. When the target is a symlink **or** a hard-linked file (`st_nlink > 1`), write in place, which follows a symlink through to its destination and rewrites a hard link's existing inode, preserving the link in both cases. The common, unlinked case keeps the temp-file + atomic rename.

### Tradeoff (deliberate)

In-place writing is **not crash-atomic**: a crash mid-write could leave a linked config file partially written. This is accepted because (a) it is the exact same tradeoff `settings.json` already makes, and (b) it only applies to files the user has deliberately linked. The default, unlinked path keeps the crash-safe atomic rename. File mode is not explicitly preserved on the unlinked rename path (umask default), which is pre-existing behavior unchanged here.

## Tests

`tests/test_model_settings_symlink.py`:
- symlink and hard-link write-through at the helper level (link survives, target/inode updated, no stray temp file)
- plain-file path unchanged (refactor regression guard)
- end-to-end `set_settings` on both a symlinked and a hard-linked `model_settings.json` (the #1958 repro, both link kinds)

Local run green: new file + existing `test_model_settings`, `test_model_settings_profiles`, `test_admin_profiles_api`, `test_settings` (350 passed).

No live-server test: this is a config file-write path with no model inference involved, and the `set_settings` integration tests exercise the real save flow end to end.

### DIFF
--- a/omlx/model_settings.py
+++ b/omlx/model_settings.py
@@ -8,7 +8,6 @@ flags, and metadata.
 import copy
 import json
 import logging
-import os
 import threading
 from dataclasses import dataclass, field, fields
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


richgoodson:
Closing this in favor of #1990, which is the better fix.

We landed on the same shape (a shared `_atomic_write_json` routing settings/profiles/templates through one helper, in-place for hard links since rename can't preserve them), but #1990 keeps symlink writes atomic by resolving the link and renaming onto its real target. My v1 did that too, then I collapsed it to an in-place write in v2 to share one path with the hard-link case. That was an unnecessary regression: only hard links force in-place. #1990 also serializes to a string before touching the file, so a mid-write failure can't truncate a linked config the way `json.dump` straight into the open file can.

Same repro coverage on both, but #1990 is strictly the stronger version. Thanks @trevorgordon981.

PR #1973, opened 2026-06-23
feat: Add Qwen2 embedding support

Mach1212:
> [!NOTE]
> **Disclaimer:** This code is AI-generated and I honestly don't fully understand all of it. It works for my setup (Qwen2-backbone embedding models on my machine), but I can't vouch for edge cases or correctness beyond that. Please review carefully before merging.

## Summary

Adds support for Qwen2-backbone text-embedding models (e.g. `jina-code-embeddings`, `gte-Qwen2`, `nomic-embed-code`). These ship a `Qwen2ForCausalLM` architecture without `lm_head` weights, so previously they were misclassified as chat LLMs, and even when routed to the embedding path mlx-embeddings rejected them with `Model type qwen2 not supported.` (it only ships a `qwen3` architecture).

## Changes

- **`omlx/models/qwen2_embedding.py`** (new): A Qwen2 text-embedding model mirroring mlx-embeddings' own `qwen3` model. Differences vs Qwen3: q/k/v projections carry a bias, the output projection does not, there is no per-head query/key RMSNorm, and `head_dim` is always `hidden_size // num_attention_heads`. Uses last-token pooling with left/right padding handling.
- **`omlx/models/mlx_embeddings_compat.py`**: `register_qwen2_embedding_model()` registers the new module into `sys.modules` as `mlx_embeddings.models.qwen2`, so the upstream loader resolves Qwen2 without patching the installed mlx-embeddings package.
- **`omlx/models/embedding.py` / `omlx/models/reranker.py`**: Call the registration before falling back to `mlx_embeddings.load`.
- **`omlx/model_discovery.py`**:
  - Add `Qwen2ForCausalLM` to `CAUSAL_LM_EMBEDDING_ARCHITECTURES`. Detection still requires the `_is_causal_lm_embedding()` directory-name heuristic, so ordinary Qwen2/Qwen2.5 chat models keep their LLM classification.
  - Add `_has_embedding_config_markers()` to detect embedding models from config keys (`matryoshka_dims`, `task_names`) when the `architectures` field is absent — some MLX conversions drop it. Only trusted when there is no architecture to contradict it, so chat/VLM models are never affected.

## Tests

- `tests/test_qwen2_embedding.py`: covers the new Qwen2 embedding model.
- `tests/test_mlx_embeddings_compat.py`: covers `register_qwen2_embedding_model()`.
- `tests/test_model_discovery.py`: positive case (jina-code-embeddings dir → embedding) and negative control (Qwen2.5 chat dir → llm).

## Testing done

Works on my local setup with Qwen2-backbone embedding models. I have not tested beyond my own machine/models.


jundot:
Thanks for putting this together, and for being transparent about the scope in the PR description.

I compared this with #1720, which covers the same #686 issue. This implementation has two correctness gaps I do not want to merge: it ignores `is_causal: false`, so bidirectional Qwen2 embedders such as `gte-Qwen2` would be served with causal attention, and left-padded inputs can produce NaN embeddings because the combined padding/causal mask uses `-inf`.

I'm going to land #1720 for the Qwen2 embedding fix and close this one in favor of that PR. The architecture-less config marker detection is a useful separate idea, and I can revisit it as a smaller follow-up if it is still needed.

PR #1987, opened 2026-06-23
fix(mcp): support 'cwd' in MCPServerConfig for stdio transport

### LINKED ISSUE #1111: MCP config fails on complaint of unused keyword "cwd"
**Describe the bug**
Starting the server with an MCP configuration causes the error 
```shell
2026-05-07 16:26:29,617 - omlx.mcp.config - INFO - [-] - Loading MCP config from: /Users/paul/.lmstudio/mcp.json
2026-05-07 16:26:29,618 - omlx.server - ERROR - [-] - Failed to initialize MCP: Invalid config for server 'lightroom': MCPServerConfig.__init__() got an unexpected keyword argument 'cwd'. MCP features disabled. Fix your MCP config and restart.
```

**To Reproduce**
1. Create this `mcp.json`
```json
{
  "servers": {
    "filesystem": {
      "transport": "stdio",
      "command": "bash",
      "args": [
        "-lc",
        "cd /Users/paul/src/third-party/lightroom/lightroom_mcp && uv run python -m mcp_server.main"
      ],
      "enabled": true,
      "timeout": 30
    }
  },
  "max_tool_calls": 10,
  "default_timeout": 30.0
}
```
2. Set the MCP Config Path to this file and restart

**Expected behavior**
Server starts and tools are available

**Screenshots**
N/A

 - macOS Version: 26.4.1
 - oMLX Version 0.3.8

**Additional context**
Output of command run from prompt
```shell
❯ bash -lc "cd /Users/paul/src/third-party/lightroom/lightroom_mcp && uv run python -m mcp_server.main"
/Volumes/ExternalHome/src/third-party/lightroom/lightroom_mcp/lightroom_sdk/types/develop.py:11: PydanticDeprecatedSince20: Pydantic V1 style `@validator` validators are deprecated. You should migrate to Pydantic V2 style `@field_validator` validators, see the migration guide for more details. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.12/migration/
  @validator('value')


                                  ╭──────────────────────────────────────────────────────────────────────────────╮
                                  │                                                                              │
                                  │                                                                              │
                                  │                         ▄▀▀ ▄▀█ █▀▀ ▀█▀ █▀▄▀█ █▀▀ █▀█                        │
                                  │                         █▀  █▀█ ▄▄█  █  █ ▀ █ █▄▄ █▀▀                        │
                                  │                                                                              │
                                  │                                                                              │
                                  │                                FastMCP 2.14.3                                │
                                  │                            https://gofastmcp.com                             │
                                  │                                                                              │
                                  │                   🖥  Server:      Lightroom Classic Bridge                   │
                                  │                   🚀 Deploy free: https://fastmcp.cloud                      │
                                  │                                                                              │
                                  ╰──────────────────────────────────────────────────────────────────────────────╯
                                  ╭──────────────────────────────────────────────────────────────────────────────╮
                                  │                          ✨ FastMCP 3.0 is coming!                           │
                                  │       Pin `fastmcp < 3` in production, then upgrade when you're ready.       │
                                  ╰──────────────────────────────────────────────────────────────────────────────╯
                                  ╭──────────────────────────────────────────────────────────────────────────────╮
                                  │                          🎉 Update available: 3.2.4                          │
                                  │                      Run: pip install --upgrade fastmcp                      │
                                  ╰──────────────────────────────────────────────────────────────────────────────╯


[05/07/26 16:41:36] INFO     Starting MCP server 'Lightroom Classic Bridge' with transport 'stdio'                                    server.py:2507
2026-05-07 16:41:36,646 - docket.worker - INFO - Starting worker 'mini.local#99247' with the following tasks:
2026-05-07 16:41:36,646 - docket.worker - INFO - * trace(message: str, ...)
2026-05-07 16:41:36,646 - docket.worker - INFO - * fail(message: str, ...)
2026-05-07 16:41:36,646 - docket.worker - INFO - * sleep(seconds: float, ...)
```

JimStenstrom:
Fixes #1111.

## Summary
- Add an optional `cwd` field to `MCPServerConfig` and pass it through to the MCP SDK's `StdioServerParameters` when connecting a stdio server.
- An MCP config with a `cwd` key on a stdio server previously crashed the whole MCP loader at startup; now the working directory is honored.

## Why
`cwd` is a standard stdio key in configs ported from other MCP hosts (Claude Desktop, LM Studio, Cursor). `MCPServerConfig` had no `cwd` field, and both config loaders build it by keyword-spread — `MCPServerConfig(**server_data)` in `validate_config()` (`omlx/mcp/config.py`) and `MCPConfig.from_dict()` (`omlx/mcp/types.py`). The unknown key raised `TypeError`, surfaced as the reported error, and disabled MCP for the *entire* config:

```
Invalid config for server 'lightroom': MCPServerConfig.__init__() got an
unexpected keyword argument 'cwd'. MCP features disabled. Fix your MCP config and restart.
```

The SDK's `StdioServerParameters` already supports `cwd`, so the correct fix is to pass the field through, not silently drop it (dropping it would discard the working directory the user configured).

## Changes
- `omlx/mcp/types.py`: add `cwd: Optional[str] = None` to `MCPServerConfig`, appended after the existing fields so positional order is unchanged.
- `omlx/mcp/client.py`: pass `cwd=self.config.cwd` to `StdioServerParameters` in `_connect_stdio()`.

## Why this is safe (no impact on existing servers)
- For any server that doesn't set `cwd`, the field defaults to `None`, so we pass `cwd=None`. `StdioServerParameters`'s own default for `cwd` is `None`, so `StdioServerParameters(command, args, env, cwd=None)` is **identical** to the previous `StdioServerParameters(command, args, env)` — verified with `model_dump()` equality. Zero behavioral change: the subprocess still launches in oMLX's working directory.
- `cwd` is a client-side launch parameter (like `env`/`args`); it is not part of the MCP handshake, so the server framework is irrelevant. Independent of whether the subprocess runs official FastMCP, standalone `fastmcp`, a Node server, etc.
- One field addition covers **both** kwargs-spread loaders (`config.py:validate_config` and `types.py:from_dict`); confirmed those are the only two construction sites.
- `cwd` is only consumed in `_connect_stdio()`. The SSE and Streamable-HTTP paths don't read it. Confirmed `StdioServerParameters(` has a single call site, so no sibling stdio path is missed.

## Tests
`pytest tests/test_mcp_types.py tests/test_mcp_config.py tests/test_mcp_client.py` → **106 passed**. All five new tests are added to the existing per-module files and fail when the fix is reverted:
- `test_mcp_types.py`: direct construction + the `**server_data` expansion in `MCPConfig.from_dict`.
- `test_mcp_config.py`: the exact `validate_config()` failure shape from the issue.
- `test_mcp_client.py`: runs the real `_connect_stdio` and asserts `cwd` reaches a real `StdioServerParameters` (the only mocks are the SDK's `stdio_client`/`ClientSession`).

Also verified end-to-end on the live server path: drove the real `init_mcp()` (`server.py`) against an on-disk `mcp.json` with `cwd` set, pointing at a real stdio MCP server.
- Without the fix: reproduced the issue's exact `MCP features disabled` failure.
- With the fix: `MCP initialized with 1 tools`; the MCP server subprocess actually spawned in the configured `cwd` (confirmed by a marker file it wrote *and* by a live `getcwd` tool call returning that directory).

### DIFF
--- a/omlx/mcp/client.py
+++ b/omlx/mcp/client.py
@@ -139,6 +139,7 @@ class MCPClient:
             command=self.config.command,
             args=self.config.args or [],
             env=self.config.env,
+            cwd=self.config.cwd,
         )
 
         # Create stdio client context
--- a/omlx/mcp/types.py
+++ b/omlx/mcp/types.py
@@ -46,6 +46,12 @@ class MCPServerConfig:
     enabled: bool = True
     timeout: float = 30.0
 
+    # Working directory for the spawned stdio MCP subprocess, passed through
+    # to the MCP SDK's StdioServerParameters. Common in configs ported from
+    # other MCP hosts (Claude Desktop, LM Studio, etc.); see issue #1111.
+    # Appended after the existing fields so their positional order is stable.
+    cwd: Optional[str] = None
+
     def __post_init__(self):
         """Validate configuration."""
         if isinstance(self.transport, str):
--- a/tests/test_mcp_client.py
+++ b/tests/test_mcp_client.py
@@ -287,6 +287,38 @@ class TestMCPClientConnect:
         assert streamable_http_client.state == MCPServerState.CONNECTED
 
     @pytest.mark.asyncio
+    async def test_connect_stdio_passes_cwd_to_sdk(self):
+        """cwd from the server config reaches StdioServerParameters (#1111).
+
+        Runs the real _connect_stdio with only the SDK's stdio_client and
+        ClientSession mocked, so the real StdioServerParameters also proves
+        the pinned SDK accepts the field.
+        """
+        mcp = pytest.importorskip("mcp")
+        config = MCPServerConfig(
+            name="cwd-test",
+            transport=MCPTransport.STDIO,
+            command="python",
+            args=["-m", "mcp_server"],
+            cwd="/tmp/mcp-workdir",
+        )
+        client = MCPClient(config)
+
+        fake_ctx = MagicMock()
+        fake_ctx.__aenter__.return_value = (MagicMock(), MagicMock())
+        with patch(
+            "mcp.client.stdio.stdio_client", return_value=fake_ctx
+        ) as mock_stdio, patch("mcp.ClientSession", return_value=MagicMock()):
+            await client._connect_stdio()
+
+        server_params = mock_stdio.call_args.args[0]
+        assert isinstance(server_params, mcp.StdioServerParameters)
+        assert server_params.cwd == "/tmp/mcp-workdir"
+        # Existing fields keep flowing through unchanged.
+        assert server_params.command == "python"
+        assert server_params.args == ["-m", "mcp_server"]
+
+    @pytest.mark.asyncio
     async def test_connect_failure(self, stdio_client: MCPClient):
         """Test connection failure sets error state."""
         with patch.object(
--- a/tests/test_mcp_config.py
+++ b/tests/test_mcp_config.py
@@ -129,6 +129,26 @@ class TestValidateConfigServerLoading:
                 {"servers": {"fs": {"transport": "stdio", "command": "x", "bogus": 1}}}
             )
 
+    def test_cwd_accepted_for_stdio_server(self):
+        """``cwd`` is a common key in stdio entries ported from other MCP
+        hosts (Claude Desktop, LM Studio). Before the #1111 fix the unknown
+        kwarg raised TypeError, surfaced as "Invalid config for server ...",
+        and disabled MCP for the whole config; it must load and carry the
+        value through instead."""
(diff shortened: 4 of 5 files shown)


jundot:
Thanks for the fix. The implementation path looks right to me, and this covers the config loading and stdio pass-through correctly. I found one small dependency-floor cleanup: cwd is only present on the MCP SDK's StdioServerParameters from mcp 1.5.0 onward, while oMLX still declares mcp>=1.0.0. This is small enough that I'll fold it into a follow-up on main after merging. This looks good to me, and I'm going to merge it.

PR #1863, opened 2026-06-13
fix(memory): charge unfused SDPA score matrix at compute dtype, not fp32

isaac-cf-wong:
## Summary

  The `head_dim > 128` prefill estimator over-charges the unfused MLX SDPA
  fallback by ~2×, producing **false `prefill_memory_exceeded` (413) rejections**
  for long-context prefills that actually fit under the ceiling.

  `estimate_prefill_peak_bytes` / `estimate_chunk_transient_bytes` charge the
  materialized fallback score matrix at a hardcoded **4 bytes/elem (fp32)**. MLX's
  softmax *accumulates* in fp32, but the dominant materialized scratch buffer that
  drives peak memory is allocated at the model's **compute/activation dtype**. For
  the fp16/bf16 models that make up every current head_dim=256 pack (Gemma 4,
  Qwen3.5/3.6), that's ~2 bytes/elem — so the estimate is double the real
  transient.

  This refines the head_dim dispatch from #1797 (which correctly restored the
  O(kv_len) estimate for head_dim=256 prefill — the dispatch is right; only the
  per-element byte width was wrong).

  ## Why thread the dtype instead of hardcoding `2`

  Hardcoding `2` would under-count a genuine fp32 model by 2× and risk the Metal
  OOM / SIGABRT the guard exists to prevent (omlx serves arbitrary HF models, not
  just bf16 ones). So instead of swapping one wrong constant for another, the score
  matrix is charged at the model's actual compute dtype, plumbed through as a new
  `compute_dtype_size`.

  This is deliberately **distinct from `dtype_size`**: the scheduler may override
  `dtype_size` to a fractional TurboQuant *KV-storage* width (~0.5), but the SDPA
  score matrix is always materialized at the activation dtype regardless of KV
  quantization. Reusing `dtype_size` would massively under-count under TurboQuant.

  ## Measurements (MLX 0.31.2)

  Synthetic SDPA at prefill shape (`n_q=16, n_kv=2, query=2048`, GQA, causal),
  peak via `mx.get_peak_memory`:

  | head_dim | kv=8k | kv=16k | kv=64k | scaling |
  |---------:|------:|-------:|-------:|---------|
  | 128      | ~25 MB | ~34 MB | flat   | **fused, O(1)** |
  | 256      | 621 MB | 1191 MB | 4614 MB | **unfused, O(kv)** |

  - head_dim=128 is flat → fused; head_dim=256 grows linearly → unfused
    (confirms #1797's dispatch).
[...]

Issue #1797, opened 2026-06-10
Regression risk from #1764: MLX fused full (prefill) SDPA only supports head_dim 64/80/128 — estimator now under-estimates head_dim=256 prefill transients, disabling the 413 guard


hojin12312:
## Describe the bug

#1764 (merged, in 0.4.4.dev1) switched `estimate_prefill_peak_bytes` / `estimate_chunk_transient_bytes` to the fused O(n) SDPA formula for **all** head_dim, on the premise that "MLX fused SDPA supports arbitrary head_dim as of MLX 0.22, the unfused fallback never executes".

That premise is true for the **vector (decode) kernel only**. The **full-attention (prefill) kernel still does not support head_dim 256** — checked directly in MLX 0.31.2 (the version omlx 0.4.x ships against), `mlx/backend/metal/scaled_dot_product_attention.cpp`, `ScaledDotProductAttention::use_fallback`:

```cpp
const bool sdpa_vector_supported_head_dim =
    query_head_dim == value_head_dim &&
    (query_head_dim == 64 || query_head_dim == 96 || query_head_dim == 128 ||
     query_head_dim == 256);
const bool sdpa_full_supported_head_dim = query_head_dim == value_head_dim &&
    (query_head_dim == 64 || query_head_dim == 80 || query_head_dim == 128);
...
const bool supports_sdpa_full = query_sequence_length > 8 &&
    sdpa_full_supported_mask && sdpa_full_supported_head_dim;

const bool supports_sdpa_vector = (query_sequence_length <= 8) && ...
```

So for a prefill chunk (`query_sequence_length > 8`) with head_dim 256, `use_fallback` returns **true** and MLX materializes the full float32 score matrix — exactly the O(kv_len) behavior the old `hd > 128` estimator branch modeled. The unfused fallback is not dead code on the prefill path; it is the *only* path for head_dim 256 prefill.

(Same situation on current MLX `main`; head_dim 256 fused *full* attention was attempted in ml-explore/mlx#3293 but that PR was closed unmerged.)

## Impact

The estimator error flips from ~1200x over-estimate to ~1200x **under**-estimate on the prefill path for head_dim=256 models (Gemma 4, Qwen3.5/3.6 family — all head_dim 256):

- Before #1764: false-positive 413s (annoying but safe).
[...]

PR #1994, opened 2026-06-24
fix(vlm): preflight counts real image tokens, not the max_pixels ceiling

fqx:
## Summary

`VLMBatchedEngine.preflight_chat` estimates prompt size by charging **every
image the model's `max_pixels`-derived per-image ceiling**
(`_derive_image_token_upper_bound`), independent of the image's actual size.
For any client that down-sizes frames before sending (video / livestream
analysis, thumbnailing, etc.) this massively over-counts and produces **false
`prefill_memory_exceeded` (HTTP 400) rejections** for prompts that run fine.

This is the *token-count* side of the head_dim=256 prefill story; it stacks on
top of the (correct) O(kv_len) SDPA fallback term from #1797 / #1863 — those
fixed the *formula*, this fixes the *kv_len fed into it*.

## Root cause

```python
# preflight_chat
num_tokens += _count_image_tokens(
    messages,
    per_image_upper_bound=_derive_image_token_upper_bound(self._processor),
)
# _count_image_tokens         -> image_parts * per_image_upper_bound  (dim-agnostic)
# _derive_image_token_upper_bound -> max_pixels // (patch**2 * merge**2)
```

The real chat path already derives the exact per-image token count from
`image_grid_thw` (`(t*h*w)//merge_sq`); only preflight uses the ceiling.

## Measurement (omlx 0.4.4, MLX 0.31.2, Qwen3.6-35B-A3B-oQ4)

Processor: `max_pixels=16777216, patch_size=16, merge_size=2`.

| | per image | x20 frames |
|---|---:|---:|
| preflight charged (`_derive_image_token_upper_bound`) | 16,384 | 327,680 |
| real (512x512 frame, grid `(1,32,32)`=1024 patches) | 256 | 5,120 |
| **over-count** | **64x** | **64x** |

`_smart_resize_tokens` matches the real `Qwen image_processor` grid exactly
across sizes:

| WxH | predicted | real processor |
|---|---:|---:|
| 512x512 | 256 | 256 |
| 510x680 | 336 | 336 |
| 1280x720 | 880 | 880 |
| 100x100 (clamped to min_pixels) | 64 | 64 |
| 4000x3000 (clamped to max_pixels) | 11750 | 11750 |

## Impact (live, brew 0.4.4 + this patch, M-series, Qwen3.6-35B-A3B-oQ4, head_dim=256)

A 20-frame 512x512 livestream-analysis request:

| | before | after |
|---|---|---|
| preflight image tokens | 327,680 | 5,120 |
| outcome | **400** `prefill_memory_exceeded` (`current 22.0 GB + KV+SDPA 22.1 GB > 43.1 GB ceiling`) | **200**, `prompt: 5187`, completed in 6.5s |

The 22.1 GB "KV+SDPA" was correct arithmetic (`n_q·chunk·kv_len·2B`) on a
64x-inflated `kv_len`; with the real count the prefill transient is well under
1 GB. The request was always runnable — only the estimate rejected it.

## Why this stays safe (never under-counts)

- `_smart_resize_tokens` reproduces the processor's deterministic
  `smart_resize` -> grid -> token math, so it returns the **exact** count the
  real path will produce, not an optimistic lower bound.
- `_read_image_dims` is decode-free (lazy `PIL.Image.open(...).size`, header
  only) and returns `None` for network URLs or unparseable data; those parts
  fall back to the existing `_derive_image_token_upper_bound` ceiling.
- Non-Qwen processors (no `patch_size`/`merge_size`/`min_pixels`/`max_pixels`)
  also fall back to the per-image ceiling. So the guard's "never under-count"
  property — the thing protecting against the Metal IOGPU OOM/SIGABRT path — is
  preserved; this only removes the over-count.

## Changes

- `omlx/engine/vlm.py`: add `_smart_resize_tokens`, `_read_image_dims`,
  `_count_image_tokens_real`; `preflight_chat` uses the latter (still passing
  `_derive_image_token_upper_bound(...)` as the fallback bound).
- `tests/test_vlm_engine.py`: `TestSmartResizeTokens` (grid math incl. min/max
  clamping), `TestReadImageDims` (data-uri / http / garbage), and
  `TestCountImageTokensReal` (real-size count, and fallbacks for unreadable
  parts / non-Qwen processors).

Inference is untouched — this only changes the admission-time estimate.

## Test plan

- `pytest tests/test_vlm_engine.py` — 86/86 pass (13 new).
- Live-validated on brew 0.4.4 (MLX 0.31.2), Qwen3.6-35B-A3B-oQ4: the 20-frame
  request goes 400 -> 200; server log shows `prompt: 5187`, no preflight
  rejection, no OOM.

## Out of scope

The orthogonal chunk-tier observation (preflight evaluates the SDPA term at the
configured `prefill_step_size` rather than the lowest tier the adaptive
throttle can reach, so genuine long-context head_dim=256 requests that *would*
complete with chunk throttling can still be rejected) is left to #1797 where it
was raised.

Refines #1797 / #1863.

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -868,6 +868,118 @@ def _count_image_tokens(
     return image_parts * per_image_upper_bound
 
 
+def _smart_resize_tokens(
+    h: int, w: int, patch_size: int, merge_size: int,
+    min_pixels: int, max_pixels: int,
+) -> int:
+    """Real merged-token count for one image of pixel size (h, w), mirroring
+    the Qwen image processor's ``smart_resize`` -> grid_thw ->
+    ``(t*h*w)//merge**2`` pipeline (t=1 for a still image). Pure arithmetic;
+    no pixel decode. This is the *exact* count the real chat path produces, so
+    it never under-counts the prefill-memory guard."""
+    import math
+
+    factor = patch_size * merge_size
+    if h <= 0 or w <= 0:
+        return 0
+    h_bar = max(factor, round(h / factor) * factor)
+    w_bar = max(factor, round(w / factor) * factor)
+    if h_bar * w_bar > max_pixels:
+        beta = math.sqrt((h * w) / max_pixels)
+        h_bar = max(factor, math.floor(h / beta / factor) * factor)
+        w_bar = max(factor, math.floor(w / beta / factor) * factor)
+    elif h_bar * w_bar < min_pixels:
+        beta = math.sqrt(min_pixels / (h * w))
+        h_bar = math.ceil(h * beta / factor) * factor
+        w_bar = math.ceil(w * beta / factor) * factor
+    return (h_bar // patch_size) * (w_bar // patch_size) // (merge_size ** 2)
+
+
+def _read_image_dims(part: dict) -> Optional[tuple]:
+    """Best-effort, decode-free ``(width, height)`` for an OpenAI image part.
+
+    Handles ``data:`` base64 URIs, raw base64, and local file paths via a lazy
+    ``PIL.Image.open`` (reads the header only, not pixels). Returns ``None`` for
+    anything that would need a network fetch or that fails to parse, so callers
+    fall back to the conservative per-image upper bound."""
+    import base64 as _b64
+    import binascii
+    import io as _io
+
+    from PIL import Image as _Image
+
+    obj = part.get("image_url")
+    if obj is None:
+        obj = part.get("input_image") or part.get("image")
+    url = obj if isinstance(obj, str) else (obj.get("url") if isinstance(obj, dict) else None)
+    if not isinstance(url, str) or not url:
+        return None
+
+    raw = None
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the detailed fix and measurements. The direction looks right, but I found one edge case that breaks the “never under-count” property in preflight.

For a very thin image, e.g. 2791x16 with Qwen settings patch_size=16, merge_size=2, and min_pixels=65536, `_smart_resize_tokens()` returns 87 tokens while the actual Qwen processor produces `image_grid_thw=[1, 2, 212]`, i.e. 106 image tokens. The difference comes from applying `max(factor, ...)` before the min/max-pixels branch; mlx-vlm’s Qwen `smart_resize` makes that branch decision on the raw rounded dimensions.

This is small and self-contained, so I’ll merge this PR and fold the branch-order fix plus the thin-image regression test into a follow-up on `main`.

Issue #1691, opened 2026-06-05
Orphaned transient prefill memory (Metal allocations) not cleared after stream failure / MidStreamFallbackError, permanently blinding Memory Guard


sanctimon:
**Describe the bug**
When a request fails mid-flight during a massive prompt prefill (throwing a litellm.MidStreamFallbackError / APIConnectionError), the intermediate transient tensors and activation matrices generated during the prefill phase remain orphaned and locked in the MLX C++ allocator cache. SpecPrefill is off for all models.

<img width="839" height="83" alt="Image" src="https://github.com/user-attachments/assets/ad5a6292-b1b0-4071-815e-e0baddc1f321" />

Even when clicking "Unload Model" via the Admin Dashboard, only the model weights are released. The massive memory footprint from the failed prefill session remains trapped inside the underlying Python/Metal process as "zombie tensors." Because this dead memory isn't forcefully evicted, it breaks multi-model workflows: the Memory Guard sees the artificial memory inflation, assumes the system is at maximum capacity (~90%), and aggressively blocks all subsequent prefills or model swaps until the entire oMLX server is hard-restarted.

**To Reproduce**
1.	Launch oMLX with Memory Guard active in a multi-model environment.
	2.	Load a large model with a deep context window configuration (e.g., Qwen-3.6-35B 8-bit, context limit set to 262,000 tokens on a 96GB Mac).
	3.	Connect an agent framework (like Hermes-Agent) and send a massive context request (e.g., ~97k tokens) where the prefix cache is invalidated (kv_len=0), forcing a full prompt prefill.
	4.	Trigger or wait for a mid-stream prefill cap error: API call failed: litellm.MidStreamFallbackError... predicted peak would exceed prefill safety cap.
	5.	Go to the oMLX Admin Dashboard and click Unload Model.
	6.	Check macOS Activity Monitor: Note that while the ~36GB model weights are freed, a massive block of memory (roughly 44GB of transient prefill "rubbish") remains captured by the Python/oMLX process.
	7.	Attempt to run any new request or load an alternative model. Memory Guard will permanently block the action, stating the predicted peak exceeds the safety cap.

**Expected behavior**
[...]

PR #2010, opened 2026-06-27
Fix VLM engine teardown memory reclaim

### LINKED ISSUE #2004: Server crashes with error from cpython-3.11
The problem seems to be worse and more repetitive as context size gets larger.

 - macOS Version: 15.7.5
 - oMLX Version:  0.4.4
 - model:  Qwen3.6-35B-A3B-oQ6-fp16-mtp (same with oQ8)
- hardware:  M2 Max 96gb
 

[crash-edited.log](https://github.com/user-attachments/files/29385583/crash-edited.log)

zwcf5200:
## Summary
- release VLM wrapper/model references before the final EngineCore teardown reclaim
- add a final worker-thread MLX reclaim after dropping engine-owned model/tokenizer/scheduler references
- clear VLM adapter-owned pending arrays and raw model references during teardown

Fixes #2004

Related reproduction: #1691, specifically https://github.com/jundot/omlx/issues/1691#issuecomment-4809293896

## Problem analysis
The related #1691 comment shows the same class of retained Metal memory on oMLX 0.4.4 after a successful long-context completion, not only after a failed prefill or mid-stream fallback. In that run, oMLX reported no active or waiting requests, unloaded every real model, and `POST /admin/api/hot-cache/clear` reclaimed 0 bytes, while `vmmap -summary` still showed tens of GB of `IOAccelerator (graphics)` memory. Restarting the Python/oMLX process released it.

That points to process-lifetime MLX/Metal allocations surviving engine teardown after the request/model state is no longer visible through model accounting, request status, or the hot-cache admin surface. This patch narrows the fix to normal teardown mechanics: drop VLM wrapper/adapter references and then perform the final MLX reclaim on the same engine worker thread/stream before shutting down the executor.

## Tests
- `.venv/bin/python -m pytest tests/test_per_engine_threads.py::TestPerEngineExecutor::test_close_clears_compile_cache_then_shuts_down tests/test_engine_core.py::TestEngineCoreClose tests/test_vlm_model_adapter.py::TestVLMModelAdapter::test_release_resources_drops_model_references -q`
- `.venv/bin/python -m pytest tests/test_vlm_engine.py::TestStopSafety -q`

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -1598,11 +1598,24 @@ class VLMBatchedEngine(BaseEngine):
 
     async def stop(self) -> None:
         """Stop the engine and cleanup resources."""
-        if self._engine:
-            await self._engine.stop()
-            if hasattr(self._engine, "engine") and self._engine.engine is not None:
+        engine = self._engine
+
+        # Drop wrapper-side references before EngineCore.close() performs its
+        # final worker-thread MLX reclaim. Otherwise the VLM wrapper can keep
+        # model weights alive until after the reclaim pass has already run.
+        self._engine = None
+        self._vlm_model = None
+        self._processor = None
+        self._adapter = None
+        self._tokenizer = None
+        self._vlm_mtp_drafter = None
+        self._diffusion_family = None
+
+        if engine:
+            await engine.stop()
+            if hasattr(engine, "engine") and engine.engine is not None:
                 try:
-                    self._engine.engine.close()
+                    engine.engine.close()
                 except Exception as e:
                     logger.warning(f"Error closing engine: {e}")
         if self._vision_cache is not None:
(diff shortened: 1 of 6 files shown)


jundot:
Thanks for the fix. The EngineCore teardown change is the right direction here: dropping engine-owned references before the final worker-thread MLX reclaim should address the process-lifetime Metal allocations reported in #2004 and the related #1691 repro.

I also noticed a couple of important remaining VLM teardown edges around the vision feature cache and diffusion cancellation ordering. Since the memory retention issue is urgent, I will merge this now and fold those fixes into a follow-up on main.

zwcf5200:
Hi @jundot, thank you for merging #2010 — the engine teardown fix is much needed.

Just a heads-up: we have two new PRs that address the same memory leak problem from complementary angles and may be even more critical:

- **[#2053](https://github.com/jundot/omlx/pull/2053)** — Fixes Instruct model first-turn memory residency (~2.15GB/round GPU leak). Tested 10 rounds, IOAccelerator drift reduced from +2150MB/round to **0.0MB**.
- **[#2052](https://github.com/jundot/omlx/pull/2052)** — Fixes embedding MLX resource release on unload. Complementary to #2053, with non-overlapping file changes.

Both have been tested on Mac Studio M3 Ultra (96GB) with the full four-model suite. We believe these should be prioritized given the severity of the resident memory leak affecting production users.

Thanks for your continued review!

Issue #1661, opened 2026-06-04
Display prefix/origin in model list (e.g., mlx-community/)


jeremychone:
**Is your feature request related to a problem? Please describe.**

Right now, there is no apparent way to see the full name used to download a model in the model list. 

So, for example, if I download `mlx-community/gemma-4-26b-a4b-it-4bit`, the list of models in the web or native settings will only show `gemma-4-26b-a4b-it-4bit`.

<img width="575" height="454" alt="Image" src="https://github.com/user-attachments/assets/aa99cf6f-8797-4ca3-b097-d65ca7f1266a" />

(Simliar in the Chat model picker)

The issue is that we do not really remember which one it is.

**Describe the solution you'd like**

It would be nice if, below or next to those names, we had the full name `mlx-community/gemma-4-26b-a4b-it-4bit` or even just the prefix `mlx-community/`.

**Describe alternatives you've considered**

Right now, I am trying to correlate with the model in the **Downloader** tab. It works as long as we do not have duplicate names.

Issue #1188, opened 2026-05-11
Model directory and naming conventions


rcfa:
Using multiple AI tools, leads to massive waste of disk space through redundant model downloads.

Logically, ML models should belong somewhere below Library, i.e.
~/Library/MLModels
/Users/Shared/Library/MLModels
/Library/MLModels
depending on the scope of model sharing (private, ad-hoc/workgroup, system)

Many AI tools, thankfully also oMLX, allow the setting of a custom model folder, so although I think it would be useful to adopt a new, better default location, the current setup allows me to at least pick what I consider the proper location (given macOS naming/folder logic conventions.)
Consequently I have picked the proper folder in the oMLX settings, Osaurus settings, and have made symlinks pointing LMStudio and others to the same location. (Ollama with its custom format, and exo are the two exemptions).

The thing where oMLX throws a fly into the ointment is the actual folder structure: while oMLX properly recognizes models in subfolders, it does NOT follow the convention of storing models in subfolders separated by model/finetuning/quantization provider, but tosses them all into the top-level folder.
As a result, if I download e.g. the Qwen3.6-27B-oQ8-mtp model with LMStudio, it ends up in 
~/Library/MLModels/Jundot/Qwen3.6-27B-oQ8-mtp
but when I download the same model with oMLX, it ends up in
~/Library/MLModels/Qwen3.6-27B-oQ8-mtp

May I kindly suggest that oMLX be updated as to follow the convention almost all other tools follow?
Cherry on the cake would be if the default location were a bit more macOS-like, as proposed above...

Thanks!

jundot:
Thanks for the detailed report. You're right — the discovery side already accepts `{owner}/{model}` subfolders (and so does the model delete endpoint), but the HF/ModelScope downloaders strip the org prefix before writing to disk, so oMLX-driven downloads still land flat. I'll fix the downloaders to preserve the full `{owner}/{model}` path in the next release.

On the default location, I'll keep that as a separate follow-up — changing it touches existing installs and deserves its own discussion.

rcfa:
Great, thanks! 👍🏻

PR #2014, opened 2026-06-27
fix(model_discovery): warn + skip on duplicate model_id instead of silent overwrite

JimStenstrom:
## Summary

When two model directories under a served dir resolve to the same bare `model_id`, `_register_model` silently overwrote the earlier registration — so which model actually serves under that id was decided by scan position, with nothing logged. The common trigger is the same model published under two orgs: a flat `served/Qwen3-8B` alongside an org-folder child `served/mlx-community/Qwen3-8B` both register as `Qwen3-8B`, and one silently shadows the other.

## Root cause

`_register_model` (`model_discovery.py`) takes the bare leaf name as the id at every call site — `_register_model(models, subdir, subdir.name)` for a flat dir, `_register_model(models, child, child.name)` for an org-folder child — and ends with an unconditional `models[model_id] = DiscoveredModel(...)`. With no collision check, a second registration replaces the first with no signal.

(The directory scan is `sorted()`, so the outcome is stable run-to-run — this is silent *shadowing*, not nondeterminism. The surprising part is that a nested org-child can replace a top-level model the operator placed deliberately, and nothing says so.)

## Fix

Guard at the top of `_register_model`: if the `model_id` is already registered, log a warning naming both colliding paths and keep the first registration.

```python
if model_id in models:
    logger.warning(
        f"Duplicate model_id '{model_id}' found in {model_dir}, "
        f"keeping version from {models[model_id].model_path}"
    )
    return
```

This mirrors the rule `discover_models_from_dirs` already applies to **cross-directory** `model_id` conflicts (warn + keep the earlier directory's model). The PR closes the within-directory gap that precedent left open. The warning message matches the sibling's phrasing.

The survivor is the first entry in the sorted directory scan (`sorted(iterdir())`) — deterministic run-to-run, but ordered **by name, not by flat-vs-nested layout**. In the `Qwen3-8B` example the flat dir wins only because `Q` sorts before `mlx-community`; a lowercase flat name vs an uppercase org folder would flip the winner.

## Why warn + keep-first (not fail-fatal or last-wins)

- **Consistency** — `discover_models_from_dirs` already treats a same-`model_id` collision as an expected condition resolved by warn + keep-earlier; this makes the single-scan case behave the same way.
- **fail-fatal would be user-hostile** — a single benign bare-name collision (the very common "same model from two orgs" layout) would abort the entire discovery scan and refuse to serve *all* models. Discovery runs at startup / admin refresh, so crashing it is the wrong trade.
- **last-wins + warn** would keep the surprising winner and merely annotate it; since both keep-first and keep-last are "deterministic but arbitrary," consistency with the existing keep-first convention is the tie-breaker.

## Scope (honest blast radius)

- `_register_model` is the **sole** `DiscoveredModel` construction + dict-write site; all callers (flat, HF-cache, org-child, single-dir fallback) pass the same `models` dict, so the guard sees every prior registration in a scan.
- The `models` dict is **ephemeral per scan** (freshly built each `discover_models` call); the guard never touches persisted serving state (`EnginePool`), so it does not interfere with rescans or metadata refreshes.
- **Behavior change to call out** — for a layout that currently collides, this flips which model serves (the first registration in the sorted scan now wins instead of the last). The warning names both paths so the change is visible in the logs.
- Out of scope — `admin/routes.py::list_hf_models` has a parallel UI-only listing dedup; it never writes the serving registry, so it is untouched here.

## Tests

`tests/test_model_discovery.py::test_register_model_skips_duplicate_id` — pre-seeds a `model_id`, registers a second **valid** model dir under it, and asserts the first registration survives (RED without the guard: the second model overwrites) and that the warning is logged naming both the kept and the skipped path.

## Related

**#1188 (model directory & naming conventions).** The maintainer confirmed there that discovery already accepts `{owner}/{model}` subfolders, but the HF/ModelScope downloaders used to strip the org prefix and write flat — and has since fixed them to preserve the full path (`admin/hf_downloader.py` now writes `<model_dir>/<owner>/<model>`). That nesting is exactly what lets a freshly-downloaded `mlx-community/Qwen3-8B` collide on the bare leaf name with a legacy flat `Qwen3-8B`. This PR is the downstream safety net that makes that collision visible and deterministic; it does not change the directory convention or the bare-name id scheme.

**#1661 (display the org prefix in the model list)** is the user-visible motivation — its reporter notes correlation "works as long as we do not have duplicate names", and the duplicate case is exactly where it breaks.

**Non-goals.** This does not make two *distinct* same-named models both servable (the second is now skipped *loudly* rather than silently — both-servable would need org-qualified ids, a compat-breaking change), and it does not add the org-prefix *display* that #1661 asks for. It strictly turns a silent shadow into a visible, deterministic one.

### DIFF
--- a/omlx/model_discovery.py
+++ b/omlx/model_discovery.py
@@ -1052,6 +1052,12 @@ def _register_model(
     source_repo_id: str | None = None,
 ) -> None:
     """Try to register a single model directory into the models dict."""
+    if model_id in models:
+        logger.warning(
+            f"Duplicate model_id '{model_id}' found in {model_dir}, "
+            f"keeping version from {models[model_id].model_path}"
+        )
+        return
     try:
         if _is_unsupported_model(model_dir):
             logger.info(f"Skipping unsupported model: {model_id}")
--- a/tests/test_model_discovery.py
+++ b/tests/test_model_discovery.py
@@ -2,6 +2,7 @@
 """Tests for model discovery functionality."""
 
 import json
+import logging
 from pathlib import Path
 
 import pytest
@@ -11,6 +12,7 @@ from omlx.model_discovery import (
     _is_adapter_dir,
     _is_unsupported_model,
     _read_model_context_length,
+    _register_model,
     _resolve_hf_cache_entry,
     detect_model_type,
     discover_models,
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for the clear write-up. The change is scoped to the single DiscoveredModel write path and matches the existing cross-directory conflict rule, so duplicate bare IDs now keep the first discovered model and log both paths instead of silently replacing it. I also checked the current merge result against the newer helper-model discovery fields. LGTM after the discovery smoke tests; merging.

PR #2020, opened 2026-06-27
fix(loader): retry weight load with strict=False on mismatch (GLM-5.2)

thornad:
## Problem
mlx-lm's `load()` loads weights with `strict=True` and raises `ValueError`
("Missing N parameters", "Received N parameters not in model", or
"Expected shape ...") for checkpoints that legitimately omit or add weights.

GLM-5.2 (`glm_moe_dsa`) is the motivating case: its DSA indexer sub-module
exists only on the `full` attention layers, but stock mlx-lm builds an indexer
for every layer, so every faithful quant aborts on load with
"Missing N parameters" for the `shared` layers. (Related: ml-explore/mlx-lm#879.)

## Fix
- `is_weight_mismatch_error()` — detects the three mlx strict-load mismatch shapes.
- `force_non_strict_weight_load()` — context manager that wraps the *current*
  `mlx_lm.utils.load_model` to force `strict=False`, composing with other
  load_model patches (e.g. DeepSeek V4) and restoring the original on exit.
- `BatchedEngine` retries the load once with strict=False when the strict load
  raises a weight-mismatch `ValueError`. The unmatched modules are never invoked
  at inference, so output is unaffected.

## Tests
`tests/test_model_loading.py` — 34 passing.

### DIFF
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -232,6 +232,8 @@ class BatchedEngine(BaseEngine):
         from ..engine_core import AsyncEngineCore, EngineConfig
         from ..scheduler import SchedulerConfig
         from ..utils.model_loading import (
+            force_non_strict_weight_load,
+            is_weight_mismatch_error,
             maybe_apply_pre_load_patches,
             maybe_load_custom_quantization,
         )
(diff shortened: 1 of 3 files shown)


thornad:
Superseded by the GLM DSA patch (#2040 / v0.4.5.dev1): `expand_glm_moe_dsa_fused_quant_keys` reconstructs the quant config so glm_moe_dsa models load correctly with strict=True, plus the native DSA kernels. That's a cleaner fix than this blanket strict=False fallback — closing in favor of it. Thanks!

PR #2021, opened 2026-06-27
fix(engine_pool): stop false 507s on back-to-back large-model swaps

thornad:
## Problem
Pre-load admission baselines on
`max(active_memory, get_phys_footprint(), _current_model_memory)`.
`phys_footprint` is the macOS kernel ledger, which keeps counting reclaimable
residue (dirty file-backed mmap'd weight pages and IOAccelerator allocations)
from a model that was just evicted until the kernel reaps it under pressure.

On a swap between two large models this inflated the baseline by hundreds of GB
even though the eviction had settled (both `active_memory` and
`_current_model_memory` dropped), so the incoming model was rejected with **507**
despite fitting cleanly.

## Fix
When eviction has exhausted every evictable model and admission is about to
fail, re-test against the tracked committed baseline
`max(active_memory, _current_model_memory)` — which excludes the reclaimable
residue — and admit if the model fits. Pinned / in-use models stay counted in
`_current_model_memory`, so this never over-admits past genuinely committed
memory (preserves the #1623 undercount guard).

## Tests
`tests/test_engine_pool.py` — 96 passing, incl. new regression
`test_swap_admits_when_footprint_lags_after_eviction` (fails without the fix,
passes with it).

### DIFF
--- a/omlx/engine_pool.py
+++ b/omlx/engine_pool.py
@@ -745,23 +745,53 @@ class EnginePool:
                         )
                         await self._unload_engine(victim)
                         continue
-                    # Nothing else to evict -- model cannot fit. Use
-                    # ModelTooLargeError when the model alone exceeds the
-                    # ceiling (no chance of fitting), InsufficientMemoryError
-                    # when the model would fit on a clean process but the
-                    # current usage leaves no room.
+                    # Nothing else to evict. Before failing, re-test against
+                    # the *tracked committed* baseline. The phys_footprint term
+                    # folded into `current` is the macOS kernel ledger, which
+                    # still counts reclaimable residue from models we just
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for tracking this down. The false 507 on back-to-back model swaps makes sense: `phys_footprint` can lag after the evicted model is no longer counted in the pool’s committed model memory, and the new regression test covers that path.

I found one narrow admission-scope edge where the committed-baseline retry can also run without evicting anything in the current pass, which could admit a first load when unrelated process footprint is the binding limit. I will take that as a small follow-up: gate the retry on an actual eviction in this admission pass and add a regression test for the no-eviction/high-footprint case.

PR #2025, opened 2026-06-28
fix(attention): tile head_dim=256 prefill to O(L) memory to fix long-context OOM

StevePierce:
**TL;DR.** On Apple Silicon (M-series, unified memory) — e.g. an M2 Max / 64 GB — `Qwen3.6-27B-oQ4-fp16-mtp` couldn't get past **~107K tokens** of context: beyond that the prefill memory guard rejected requests, well short of the model's **256K** window. The cause is that head_dim 256 has no fused MLX attention kernel, so prefill falls back to materializing the full `[n_q, query_len, kv_len]` score matrix — **O(L²) memory**. This PR routes that case through a flash-style online-softmax pass that never builds the matrix, dropping the prefill memory transient to **O(L)**. Result: **the ~107K memory wall is gone** — prefill memory now scales O(L) across the model's full **256K** (262,144-token) window. It's a **memory fix, not a speed fix** — the attention compute is still quadratic; this linearizes memory, not compute.

Maintainer's own `Qwen3.6-27B-oQ4-fp16-mtp` build, M2 Max 64 GB.

## Memory (the win) — per-chunk prefill transient (24:4 GQA, fp16, causal, qL=2048)

| context | this patch | old O(L²) fallback |
|---|---|---|
| 32K  | 0.74 GB | 3.2 GB |
| 128K | 2.38 GB | 12.9 GB |
| 240K | 4.33 GB | 24.2 GB |

Clean ~2×/doubling = O(L); the unfused fallback already needs ~51 GB at 32K (single-shot) and can't run long context at all.

## Throughput (the cost)

The kernel rides MLX's own GEMM, so it's close to the fallback, but **not free**. Per-call latency vs the unfused fallback, where the fallback can still run:

| L | this patch | unfused fallback | fallback peak |
|---|---|---|---|
| 4K  | 51 ms  | 44 ms (**1.16×**) | 0.99 GB |
| 8K  | 193 ms | 176 ms (**1.10×**) | 3.62 GB |
| 16K | 784 ms | 764 ms (**1.03×**) | 13.8 GB |
| ≥32K | runs | **OOM** (needs ~51 GB) | — |

So: a **~3–16% prefill slowdown at short/mid context that converges to parity by 16K**, in exchange for long context being reachable at all (past ~16–32K the fallback OOMs, so there's no slower/faster — it just doesn't run). This is the intended trade.

## Scope / safety

- **Narrowly gated.** Engages only for `head_dim == 256`, `query_len > 1`, causal/no mask, `kv_len ≥ 8192`, no sinks, non-quantized cache. Everything else — other models, decode, short prompts, array/windowed masks, TurboQuant caches — passes straight through to the original SDPA. If the kernel ever throws while engaged, it falls back to the original too. Off-switch: `sdpa256_prefill_enabled` (default on).
- Installed after the TurboQuant patch (outer wrapper, only grabs non-quantized 256 prefill); coexists with Native MTP (verified live).
- Additive; `test_memory_monitor.py` stays green, plus new tests for the kernel, the route gate, and the estimator switch.

## Validation

Numerical match vs `mx.fast.scaled_dot_product_attention` < 2e-2 max-abs across square, chunked-prefill (end-aligned causal), and decode shapes; peak memory confirmed O(L) out to 240K. The tests fail if reverted: the kernel tests assert < 2e-2 vs `mx.fast` SDPA, and the estimator test asserts the registered O(L) estimate is >10× smaller than the unregistered O(L²) one.

End-to-end on `Qwen3.6-27B-oQ4-fp16-mtp`: a 23.6K-token prompt with a fact planted at the front returns it correctly (`finish_reason=stop`), and a ~174K prompt that previously hit the guard now prefills.

<details>
<summary><b>Deep dive: root cause, the fix, and how it relates to existing work</b></summary>

### Root cause
MLX only ships a fused (flash) attention kernel for `head_dim ∈ {64, 80, 128}`. Qwen3.6 uses head_dim 256, so multi-token prefill (`query_len > 1`) falls off the fused path to the generic implementation, which materializes the full `[n_q, query_len, kv_len]` score matrix → O(L²) memory in sequence length. That transient is what crosses the ceiling at long context (and it's also why head_dim-256 prefill is slow — it forces tiny prefill chunks). Decode (`query_len == 1`) is unaffected — MLX has a fused vector kernel that supports 256.

### The fix
`_flash_sdpa256` tiles over Q and KV and keeps a running online softmax (max `m`, denominator `denom`, output accumulator `acc`, all fp32) so the score matrix is never materialized. It handles GQA (reshape 24 query heads into 4 groups of 6) and end-aligned causal masking (cached-prefix chunked prefill). After each KV tile it `mx.eval`s the running state to force MLX to free that tile — this is what makes peak memory O(L). (Per-tile eval cadence was chosen deliberately: coarser cadence — every 4/8/16 tiles — was tested and gave no speedup while using 2–3× more memory, because the kernel is matmul-bound, not barrier-bound.) Matmuls are done in fp32; fp16 matmuls were tested and were slightly *slower* (extra casts, and the matmul isn't the bottleneck) with identical accuracy.

`register_tiled_prefill_head_dim` flips the prefill-guard estimator (`_estimate_sdpa_activation_bytes`) to charge `output + one KV tile of scores` instead of the full matrix, gated identically to the route — so the guard stops rejecting in lockstep. The registry is empty unless the patch installs, so a build without the patch keeps the O(L²) estimate (no behavior change).

### Relationship to existing work
This **complements** the merged head_dim-256 prefill memory-guard line (#1797 → #1863 → #1994): those make the unfused O(L²) cost *estimated correctly* so the guard rejects cleanly. This PR removes that O(L²) materialization and flips the **same** estimator to O(L) — so the guard stops *rejecting* rather than just rejecting *accurately*. It extends that direction, it doesn't revert it.

Install mechanics mirror `turboquant_attention.py` (patch the module attr + rebind already-imported model modules — scoped to modules that actually imported the base SDPA). And it's a framework-level stopgap for the upstream kernel gap: when a real fused head_dim-256 Metal kernel lands (ml-explore/mlx#3660 is the in-flight effort; issue #3658 has the matching real-world OOM evidence), this can be retired or kept as the CPU/short-seq fallback.

</details>

### DIFF
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -300,6 +300,24 @@ class BatchedEngine(BaseEngine):
                 tq_bits = float(getattr(self._model_settings, "turboquant_kv_bits", 4))
                 logger.info(f"TurboQuant KV cache enabled: {tq_bits} bits")
 
+        # head_dim=256 long-context prefill: route to an O(L) tiled SDPA kernel
+        # so models like Qwen3.6-27B stop OOMing / getting prefill-guard-rejected
+        # below their context window. Installed after TurboQuant so it is the
+        # outer wrapper and only grabs non-quantized 256 prefill; all other
+        # cases (incl. TurboQuant caches, other head dims, decode, short
+        # prefill) fall through to the prior SDPA unchanged. Passthrough-safe to
+        # install unconditionally — the route is strictly gated. Disable via
+        # model_settings.sdpa256_prefill_enabled = False.
+        if getattr(self._model_settings, "sdpa256_prefill_enabled", True) is not False:
+            try:
+                from ..patches.sdpa256_attention import (
+                    apply_sdpa256_attention_patch,
+                )
+
+                apply_sdpa256_attention_patch()
+            except Exception:
+                logger.debug("sdpa256 attention patch not applied", exc_info=True)
+
         # Create engine config (copy to avoid mutating the shared instance)
         scheduler_config = (
             copy.copy(self._scheduler_config)
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -1491,6 +1491,18 @@ class VLMBatchedEngine(BaseEngine):
                 )
                 scheduler._set_model_info_for_monitor()
                 logger.info(f"TurboQuant KV cache enabled for VLM: {tq_bits} bits")
+
+        # head_dim=256 long-context prefill -> O(L) tiled SDPA kernel. See
+        # batched.py for rationale. Passthrough-safe; strictly gated route.
+        if getattr(self._model_settings, "sdpa256_prefill_enabled", True) is not False:
+            try:
+                from ..patches.sdpa256_attention import (
+                    apply_sdpa256_attention_patch,
+                )
+
+                apply_sdpa256_attention_patch()
+            except Exception:
+                logger.debug("sdpa256 attention patch not applied", exc_info=True)
         scheduler.refresh_ssd_layer_signature()
 
         # SpecPrefill: load draft model and pass to scheduler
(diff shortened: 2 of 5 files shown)


jundot:
Thanks for the update. The core approach looks good to me: the route is narrow, the memory estimator is updated in lockstep, and the focused SDPA/memory/TurboQuant tests pass locally.

Issue #2039, opened 2026-06-29
ERROR: Invalid wheel filename issue when trying to install oMLX via hombrew


totally-tim:
**Describe the bug**
when trying to install oMLX via homebrew I am getting the error: 
ERROR: Invalid wheel filename (wrong number of parts): 'f94ba1382eb024cacf740302cfa749e22250046633501d34ee4e8d9eb26db9f5--en_core_web_sm-3.8.0-py3-none-any'


**To Reproduce**
 brew install or reinstall jundot/omlx/omlx
**Expected behavior**
successfull install

**Screenshots**

**Desktop (please complete the following information):**
 - macOS Version: [e.g. 26.5.1]
 - oMLX Version [0.4.4]

**Additional context**
Please let me know if this is user error


Full logs:

==> Fetching downloads for: omlx
✔︎ Resource omlx--en-core-web-sm                                                                                                                                                                                                                       Verified     12.8MB/ 12.8MB
✔︎ Formula omlx (0.4.4)                                                                                                                                                                                                                                Verified     46.3MB/ 46.3MB
✔︎ Resource omlx--mlx-audio
==> Reinstalling jundot/omlx/omlx 
==> python3.11 -m venv /opt/homebrew/Cellar/omlx/0.4.4/libexec
==> /opt/homebrew/Cellar/omlx/0.4.4/libexec/bin/pip install --no-binary cohere_melody,pydantic-core,rpds-py,tiktoken /private/tmp/omlx-20260629-8434-af0gfe/omlx-0.4.4
==> /opt/homebrew/Cellar/omlx/0.4.4/libexec/bin/pip install .[all]
==> /opt/homebrew/Cellar/omlx/0.4.4/libexec/bin/pip install /Users/macstudio/Library/Caches/Homebrew/downloads/f94ba1382eb024cacf740302cfa749e22250046633501d34ee4e8d9eb26db9f5--en_core_web_sm-3.8.0-py3-none-any.whl
Last 15 lines from /Users/macstudio/Library/Logs/Homebrew/omlx/04.pip.log:
2026-06-29 12:29:03 +0000

/opt/homebrew/Cellar/omlx/0.4.4/libexec/bin/pip
install
--no-deps
/Users/macstudio/Library/Caches/Homebrew/downloads/f94ba1382eb024cacf740302cfa749e22250046633501d34ee4e8d9eb26db9f5--en_core_web_sm-3.8.0-py3-none-any.whl

[...]

PR #2041, opened 2026-06-29
Fix spaCy model wheel install

MahadMuhammad:
- The Homebrew formula currently passes the cached `en-core-web-sm` resource path directly to `pip install`.
- That can fail because Homebrew stores downloaded resources with cache-prefixed filenames.
- `pip` validates wheel filenames before installing them, so it rejects that cache filename even though the downloaded wheel itself is valid.
- This change copies the resource into the build directory using the original wheel filename from the resource URL, then installs that copy.
- It also fetches the resource first if it is not already present in Homebrew's cache.

### DIFF
--- a/Formula/omlx.rb
+++ b/Formula/omlx.rb
@@ -85,8 +85,11 @@ class Omlx < Formula
     end
 
     # Install the spaCy English model required by misaki for Kokoro TTS.
-    system libexec/"bin/pip", "install", "--no-deps",
-           resource("en-core-web-sm").cached_download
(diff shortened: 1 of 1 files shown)


jundot:
@MahadMuhammad Thanks for sending this. This is the same root cause as #2039, and I had just pushed the fix to main in 796672ab before seeing this PR.

The current main branch now copies the cached spaCy model resource back to the canonical wheel filename before passing it to pip, so this PR is covered and now conflicts with main. I'll close this as already fixed.

PR #2043, opened 2026-06-29
fix(server): keep chat stream reasoning out of content

isaac-cf-wong:
## Summary

Fixes OpenAI chat-completions streaming for reasoning models whose rendered chat template already opens a `<think>` block before generation starts.

Previously, `stream_chat_completion()` always initialized `ThinkingParser()` with `start_in_thinking=False`. For Qwen/DeepSeek-style templates that leave the prompt inside an open thinking block, the
first generated reasoning tokens were streamed as `delta.content` until the model emitted `</think>`. That could leak reasoning text into clients that persist assistant content.

This changes chat streaming to mirror the prompt-aware behavior used elsewhere: render the chat prompt, call `prompt_opens_thinking()`, and pass the result to `ThinkingParser(start_in_thinking=...)`.

## Tests

- Added a functional streaming regression test for prompt-opened thinking.
- Added a static guard to ensure `stream_chat_completion()` continues wiring `prompt_opens_thinking()` into `ThinkingParser(start_in_thinking=...)`.
- Ran:
  - `tests/test_chat_stream_thinking_static.py`
  - `tests/test_thinking.py`

## Verification

Before the fix, the static regression guard fails because `stream_chat_completion()` does not call `prompt_opens_thinking()`.

After the fix, prompt-opened reasoning streams via `reasoning_content`, and only post-`</think>` answer text streams as `content`.

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -4166,7 +4166,20 @@ async def stream_chat_completion(
     last_output = None
     accumulated_text = ""
     has_tools = bool(kwargs.get("tools"))
-    thinking_parser = ThinkingParser()
+    start_in_thinking = False
+    try:
+        tokenizer = getattr(engine, "tokenizer", None)
+        if tokenizer is not None:
+            template_kwargs = {"tokenize": False, "add_generation_prompt": True}
+            if kwargs.get("tools"):
+                template_kwargs["tools"] = kwargs["tools"]
+            if kwargs.get("chat_template_kwargs"):
+                template_kwargs.update(kwargs["chat_template_kwargs"])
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for the fix. The change matches the existing prompt-aware thinking detection and covers the leak where a chat template opens `<think>` before generation starts, so the first generated reasoning tokens now stay in `reasoning_content`.

CI and the local streaming/thinking regression tests pass. I noticed a small follow-up around keeping the detection render path aligned with partial/template handling, but it is not a merge blocker. LGTM. I'll merge this and fold that cleanup into main.

PR #2044, opened 2026-06-29
feat: Support for Ornith models, both in oQ and at inference time

kreeger:
## Overview

I downloaded the [Ornith 9B and 35B](https://huggingface.co/collections/deepreinforce-ai/ornith-10) (moe) raw models off HuggingFace to generate my own FP16 quants to run on M1 and M2, and in doing so, I ran into a few snags. This changeset adds support so that oQ quantization works and doesn't generate broken models, and also adds support during inference-time for using these models.

## Ornith?

The Ornith models are a family of multimodal MoE models fine-tuned from Qwen 3.5 with an MTP head. oMLX's MoE path was written for Qwen 3.6, which stores experts fused (`experts.gate_up_proj`). Ornith/Qwen 3.5 ships per-expert tensors (`experts.{N}.{gate,up,down}_proj`) instead.

The loader didn't know how to convert that form, so weights were rejected at load while trying to do inference. Two smaller problems also bit me: per-layer quantization keys used a different prefix order than the runtime expects, and recent `mlx-lm` updates _seem to have_ dropped `trust_remote_code` from `load()`.

## The fixes

- Per-expert experts: MoE sanitize now falls back to stacking per-expert tensors into `switch_mlp` form when the fused layout is absent.
    + I applied this to the LLM, VLM, and VLM-runtime paths
    + Quant metadata (`.scales/.biases`) is stacked too
    + Existing fused (Qwen 3.6) checkpoints are unaffected
- Quant keys: `expand_per_layer_quant_keys` now emits the correct module-tree variant for all key prefix conventions, so per-layer bit-widths match at load
- `mlx-lm` drift: new `lm_load_compat` wrapper forwards `trust_remote_code` only when the installed `mlx-lm` accepts it
    + All `load()` call sites route through this new wrapper
    + This change felt the most invasive, but also felt necessary!

Anyway, with these changes over the last few days since the Ornith models have been out, I've had good success in running inference and benchmarks against the models and things seem pretty stable, so now I'm pushing them up for you to review.

### DIFF
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -227,11 +227,10 @@ class BatchedEngine(BaseEngine):
 
         import asyncio
 
-        from mlx_lm import load
-
         from ..engine_core import AsyncEngineCore, EngineConfig
         from ..scheduler import SchedulerConfig
         from ..utils.model_loading import (
+            lm_load_compat,
             maybe_apply_pre_load_patches,
             maybe_load_custom_quantization,
         )
@@ -263,7 +262,7 @@ class BatchedEngine(BaseEngine):
                 model, processor = custom_loaded
                 return model, getattr(processor, "tokenizer", processor)
 
-            return load(
+            return lm_load_compat(
                 self._model_name,
                 tokenizer_config=tokenizer_config,
                 trust_remote_code=self._trust_remote_code,
@@ -383,7 +382,7 @@ class BatchedEngine(BaseEngine):
                                 specprefill_draft,
                                 trust_remote_code=self._trust_remote_code,
                             )
-                            draft_model, _ = load(
+                            draft_model, _ = lm_load_compat(
                                 specprefill_draft,
                                 tokenizer_config=draft_tokenizer_config,
                                 trust_remote_code=self._trust_remote_code,
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -1515,9 +1515,10 @@ class VLMBatchedEngine(BaseEngine):
             )
             if specprefill_enabled and specprefill_draft:
                 try:
-                    from mlx_lm import load as mlx_lm_load
-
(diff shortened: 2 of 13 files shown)


kreeger:
You know, this changeset notwithstanding, it looks like @deepsweet has several quants up now for Ornith's models. @deepsweet did you not experience problems quantizing or running Ornith models on 0.4.4?

deepsweet:
> You know, this changeset notwithstanding, it looks like @deepsweet has several quants up now for Ornith's models. @deepsweet did you not experience problems quantizing or running Ornith models on 0.4.4?

I did, and I had to vibecode a script to stack expert MLP gate/up/down weights into fused per-layer tensors. Would be nice to have this supported natively though.

kreeger:
My changeset doesn't preserve the MTP heads (at least not yet); I may save that for a follow-up PR.

**Edit**: Actually everything I'm seeing says these models don't have MTP heads — though I could have sworn they did when they first got released. [Somebody on HF](https://huggingface.co/protoLabsAI/Ornith-1.0-9B-MTP) has support for grafting MTP heads on it, but nothing official.

jundot:
Thanks @kreeger and @deepsweet. I checked a local `deepreinforce-ai/Ornith-1.0-35B` checkpoint as well, and this PR matches the layout I see there: the backbone MoE weights are per-expert tensors, while the safetensors index has no actual `mtp.*` weights even though the config still declares `text_config.mtp_num_hidden_layers=1`.

The per-expert backbone tensors dry-run cleanly into `switch_mlp.*` with no orphan `experts.*` keys, and the current no-MTP-weights guard is the right behavior for that checkpoint.

I will keep actual MTP-bearing / grafted Ornith-style checkpoints as a follow-up, including VLM oQ sanitizer parity for per-expert MTP quant metadata (`.scales` / `.biases`).

LGTM. I will merge this.

PR #2056, opened 2026-07-01
fix: reduce empty-process native memory retention

zwcf5200:
## Summary
- set MallocSpaceEfficient=1 for the macOS app server process before Python starts
- add a small shared teardown helper for wrapper-side reference clearing
- use the helper in batched/VLM stop paths to clear model/tokenizer/native grammar compiler references consistently
- keep this follow-up PR scoped to code and tests only; no docs included

## Testing
- git diff --check
- .venv/bin/python -m pytest tests/test_batched_engine.py::TestBatchedEngineInitialization tests/test_vlm_engine.py::TestStopSafety tests/test_grammar.py::TestListGrammarParsers
- .venv/bin/python -m pytest tests/test_batched_engine.py::TestBatchedEngineInitialization tests/test_vlm_engine.py tests/test_grammar.py::TestListGrammarParsers

## Notes
- This is a follow-up to the model unload memory work and intentionally excludes documentation changes.
- The helper is deliberately limited to wrapper-side references; EngineCore still performs stream-local MLX reclaim on its own executor.

PR #2053, opened 2026-07-01
fix: prevent Instruct model first-turn memory residency via engine cleanup

zwcf5200:
## Summary

Fixes resident memory leak in Instruct models after first-turn inference. Each load/unload cycle leaves ~2.15GB of `IOAccelerator (graphics)` memory permanently resident, causing linear GPU memory growth that eventually exhausts the budget.

**Note:** This PR is complementary to #2054 (`fix/embedding-mlx-resource-release`), which handles embedding-specific MLX resource cleanup. This PR focuses on the engine_pool, scheduler, and multi-engine close_stateless_model path that directly addresses Instruct model residency.

## Changes

- **`omlx/engine_pool.py`**: Add `close_stateless_model` logging and MLX resource release for models that do not stream (Instruct, Reranker, etc.)
- **`omlx/engine_core.py`**: Fix `model_info` reference preventing proper cleanup; add logging for `close_stateless_model`
- **`omlx/engine/dflash.py`**: Release MLX resources in `close_stateless_model`
- **`omlx/engine/reranker.py`**: Release MLX resources in `close_stateless_model`
- **`omlx/engine/sts.py`**: Release MLX resources in `close_stateless_model`
- **`omlx/engine/stt.py`**: Release MLX resources in `close_stateless_model`
- **`omlx/engine/tts.py`**: Release MLX resources in `close_stateless_model`
- **`omlx/engine/vlm.py`**: Release MLX resources in `close_stateless_model`
- **`omlx/models/reranker.py`**: Add `close()` method to release MLX resources for stateless models
- **`omlx/scheduler.py`**: Add `close_stateless_model` handler to scheduler
- **`scripts/omlx_memory_cycle_probe.py`**: Add memory cycle probe tool for load/unload testing
- **`docs/omlx-memory-cycle-retest-plan-zh.md`**: Reproduction test plan document
- **`docs/omlx-11335-model-cycle-memory-leak-report-zh.md`**: Memory cycle leak report

## Related Issues

- Related #1060 (engine_pool active_memory counter leak after VLM model unload)
- Related #1595 (Memory enforcer eviction path leaks active_requests counter)
- Related #1691 (Orphaned transient prefill memory not cleared after stream failure)

## Test Results

All tests passed on Mac Studio M3 Ultra (96GB). Four-model suite 10-round cycle:

| Metric | Before Fix | After Fix |
|---|---|---|
[...]

PR #2061, opened 2026-07-02
fix: drop engine constructor aliases during unload

zwcf5200:
## Summary
- drop transient model/tokenizer constructor aliases in EngineCore, AsyncEngineCore, and Scheduler after ownership moves to instance fields
- keep the change scoped to the batched engine construction path that affects LLM unload teardown
- no local docs/scripts or unrelated engine cleanup are included

## Why
A Qwen3-4B-Instruct-2507-MLX-4bit single-model load/unload cycle on current upstream/main (#2056 included) still leaves the first unloaded process at about 2.15GB of IOAccelerator (graphics) residency even though /api/status reports models_loaded=0 and model_memory_used=0.

I rechecked the older local integration fix after the review feedback on #2053. The broad local-reference cleanup did remove this first-unload residency, so I isolated the smallest code path that reproduces the improvement.

## Isolation results
All runs used a restarted 127.0.0.1:11335 server, MallocSpaceEfficient=1, clean initial /api/status, and vmmap -summary sampling.

- upstream/main 2edd26e (#2056): Qwen3-4B after-unload IOAccelerator stayed at 2150.4MB from cycle 1 through cycle 5
- local/integration broad cleanup: Qwen3-4B after-unload IOAccelerator returned to 1.7MB from cycle 1
- scheduler-only alias cleanup: still retained 2150.4MB after unload
- engine-core-only alias cleanup: still retained 2150.4MB after unload
- this combined EngineCore + Scheduler alias cleanup: returned to 1.7MB after unload for cycles 1-3

Control model:
- Ornith-1.0-35B-5bit-mlx on upstream/main returned to 2.0MB IOAccelerator after unload across 3 cycles, so the Qwen result is not a generic vmmap/probe artifact.

## Testing
- git diff --check
- .venv/bin/python -m py_compile omlx/engine_core.py omlx/scheduler.py
- live Qwen3-4B-Instruct-2507-MLX-4bit 3-cycle load/unload vmmap validation on this branch

### DIFF
--- a/omlx/engine_core.py
+++ b/omlx/engine_core.py
@@ -258,6 +258,11 @@ class EngineCore:
         self._start_time: Optional[float] = None
         self._steps_executed = 0
 
+        # Drop transient aliases after ownership moves to the engine/scheduler
+        # graph, so close()/deep_reset() can make that graph unreachable.
+        model = None
+        tokenizer = None
+
         logger.debug(f"Engine {self._engine_id} initialized")
 
     async def start(self) -> None:
@@ -1185,6 +1190,9 @@ class AsyncEngineCore:
         config: Optional[EngineConfig] = None,
     ):
         self.engine = EngineCore(model, tokenizer, config)
+        # Drop wrapper-local aliases after EngineCore takes ownership.
+        model = None
+        tokenizer = None
 
     @property
     def _mlx_executor(self):
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -1875,6 +1875,11 @@ class Scheduler:
         # since _get_xtc_special_tokens() delegates to _get_stop_tokens().
         self._xtc_special_tokens: list[int] = self._get_xtc_special_tokens()
 
+        # Drop transient aliases after ownership moves to scheduler fields;
+        # close()/deep_reset() clear those fields during teardown.
+        model = None
+        tokenizer = None
+
     @contextmanager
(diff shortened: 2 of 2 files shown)


zwcf5200:
Hi maintainer, this is a must-reproduce memory leak bug: after loading and unloading Qwen3-4B-Instruct-2507-MLX-4bit once, ~2.15 GB of IOAccelerator (graphics) memory remains resident even though /api/status reports models_loaded=0 and model_memory_used=0. The fix is very small — dropping engine constructor aliases during unload — and the code change is minimal. Could you please take a look? I believe this fix is essential to fully resolve the memory residency issue.

jundot:
Thanks for finding this and narrowing it down. I reproduced the unload issue and confirmed this PR fixes it, so I'll merge it.

The root cause was an optional native-kernel import error keeping its traceback alive. That traceback kept old constructor frames around, and those frames still referenced the model/tokenizer.

I'll add a follow-up commit to fix that root cause directly by clearing the traceback before storing the import error. Thanks again for taking the time to identify the issue and push for the fix.

zwcf5200:
Thanks for the quick review and merge, and especially for identifying the root cause. Clearing the traceback before storing the import error makes perfect sense — it's exactly the kind of subtle reference cycle that keeps frames alive and prevents memory release. Appreciate the follow-up commit to address it directly.

PR #2070, opened 2026-07-02
32-head sparse-MLA variant for tensor-sharded (multi-device) runs

aidiffuser:
## What

Adds a 32-head instantiation of the GLM-5.2 sparse-MLA prefill kernel, for setups that shard the 64 MLA heads across devices (tensor parallelism). Three small changes:

- `sparse_mla.cpp`: accept `H ∈ {32, 64}`; `h`/`wm` become runtime values (`wm=4` for H=32 so the threadgroup tile `TQ = H/(wm*8)` stays ≥ 1 — with the shipped `wm=8`, H=32 gives TQ=0 and trips the static assert).
- `sparse_mla.metal`: 32-head fp16/bf16 instantiations.
- `glm_moe_dsa_model.py`: relax the 64-head gate to `(32, 64)`.

Single-device 64-head behavior is unchanged (same instantiation, same `wm`).

## Why

We run GLM-5.2 tensor-parallel across two Mac Studios (512 GB each, Thunderbolt) via exo, with your kernels vendored (thanks — they're excellent). Each node holds 32 of the 64 heads, so the kernel's 64-head guard sent every shard to the fallback path. This PR is the change we've been running in production.

## Validation

- Output parity with the pure-MLX sparse reference to bf16 noise (~3e-3 rel) at both 64 and 32 heads.
- Live 2-node cluster: 100% structured needle-in-haystack recall at 120k/200k/300k/350k-token contexts (temp-0, decoys, 5 depth positions), coherent output, byte-consistent across shards.
- Prefill 1.59× at 41k vs the non-kernel sparse path; ~3.5× vs dense at 84k.

One note from long-context testing that may interest you: your fused `dsa_indexer_scores` path is immune to an MLX reduction bug we root-caused (`mx.sum` over a non-last axis of large strided tensors returns all-zeros, ml-explore/mlx#3784) — the pure-MLX indexer path materializes exactly the trigger shape at ≥128k contexts. We also upstreamed a workaround + a pure-MLX sparse prefill to mlx-lm (ml-explore/mlx-lm#1454), crediting this PR for the concept.

### DIFF
--- a/omlx/custom_kernels/glm_moe_dsa/csrc/sparse_mla.cpp
+++ b/omlx/custom_kernels/glm_moe_dsa/csrc/sparse_mla.cpp
@@ -109,8 +109,9 @@ class GlmDsaSparseMlaAttentionPrimitive : public Primitive {
         !last_dim_contiguous(topk_indices)) {
       return true;
     }
-    if (q_latent.shape(1) != 64 || kv_latent.shape(1) != 1 ||
-        k_pe.shape(1) != 1 || topk_indices.shape(1) != 1) {
+    if ((q_latent.shape(1) != 64 && q_latent.shape(1) != 32) ||
+        kv_latent.shape(1) != 1 || k_pe.shape(1) != 1 ||
+        topk_indices.shape(1) != 1) {
       return true;
     }
     if (q_latent.shape(3) != 512 || kv_latent.shape(3) != 512 ||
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for the focused patch. I checked the sharded GLM DSA path: Model.shard() reduces the local attention heads before the sparse MLA prefill path, and the H=32/wm=4 instantiation keeps the Metal tile invariants while leaving the H=64 path unchanged.

I also built the custom kernels locally and verified the existing native reference test plus direct H=32 fp16/bf16 sparse MLA calls against a dense reference. LGTM.

PR #2071, opened 2026-07-02
fix(cache): key paged SSD blocks by TurboQuant bit depth

### LINKED ISSUE #2045: Paged SSD cache signature does not include turboquant_kv_bits — switching bit depth crashes at mx.concatenate with shape mismatch
## Describe the bug

When `turboquant_kv_bits` is changed on the same model path (e.g. from `4.0` to `6.0`), the paged SSD cache retains stale blocks with mismatched packed widths, causing a fatal `mx.concatenate` shape-mismatch crash at batch time.

The paged SSD cache signature does **not** include `turboquant_kv_bits`, so 4-bit blocks (`packed_width=32` at `head_dim=256`) and 6-bit blocks (`packed_width=48`) for the same model path coexist in the cache. When oMLX tries to concatenate them in `omlx/turboquant_kv.py:_concat_state_batch`, the operation fails:

```
[concatenate] All the input array dimensions must match exactly except for the concatenation axis. 
However, the provided shapes are (1,2,21061,48), (1,2,21061,32), and the concatenation axis is 0.
```

## To Reproduce

1. Load model `Ornith-1.0-35B-4bit` (head_dim=256) with `turboquant_kv_bits=4.0` in `~/.omlx/model_settings.json`.
2. Run several requests to populate the paged SSD cache (produces 4-bit blocks, `packed_width=32`).
3. Change `turboquant_kv_bits` to `6.0` in `~/.omlx/model_settings.json`.
4. Restart oMLX (or reload the model).
5. Send a new request → server crashes with the `mx.concatenate` shape-mismatch error above.

Cache file headers confirm the mismatch: 4-bit blocks have `packed_width=32` (`ceil(256*4/32)`), 6-bit blocks have `packed_width=48` (`ceil(256*6/32)`). Both exist in `~/.omlx/cache/[0-9a-f]/*.safetensors` for the same model path.

## Expected behavior

When `turboquant_kv_bits` changes, one of the following should happen:

- (a) The paged SSD cache is invalidated for that model path, OR
- (b) The cache signature includes `turboquant_kv_bits` so that 4-bit and 6-bit blocks are keyed separately, OR
- (c) The batch-time concatenation rejects mixed-width TurboQuant states with a clear error message instead of crashing.

## Screenshots

N/A (server crash, no GUI state). Log excerpt:

```
2026-06-29 16:00:01,983 - omlx.cache.prefix_cache - ERROR - [-] - Failed to reconstruct cache: concatenate(): incompatible function arguments. 
The following argument types are supported:
    1. concatenate(arrays: list[array], axis: Optional[int] = 0, *, stream: Union[None, Stream, Device] = None) -> array
Invoked with types: list, kwargs = { axis: int }
```

## Desktop

- macOS Version: 27 (Darwin)
- oMLX Version: 0.4.5.dev1 (latest)
- Model: `Ornith-1.0-35B-4bit` (head_dim=256)
- Config: `turboquant_kv_bits` switched from `4.0` -> `6.0` in `~/.omlx/model_settings.json`

## Additional context

- The failing line is in `omlx/turboquant_kv.py:_concat_state_batch`: `mx.concatenate([s.indices for s in states], axis=0)`.
- The cache scanner (`omlx/cache/prefix_cache.py`) does not include `turboquant_kv_bits` in the paged SSD cache signature, so it cannot distinguish 4-bit from 6-bit blocks for the same model path.
- Workaround: clear `~/.omlx/cache/` and restart oMLX when changing `turboquant_kv_bits`.
- Related open issues: **#1693** (cache corruption on `block_size` change), **#766** (SpecPrefill + TurboQuant KV cache corruption), **#1925** (TurboQuant KV breaks prefix cache on hybrid models). None address the specific case of switching `turboquant_kv_bits` on the same model path.

JimStenstrom:
Fixes #2045.

## Summary

- The paged SSD cache signature now records `turboquant_kv_bits`, stamped from each block's own TurboQuant meta_state, so blocks written at different depths can no longer coexist as "compatible" and crash batch concatenation.
- The compatibility check and the one-shot stale-signature sweep enforce the recorded depth whenever a depth is expected; a depth change alone triggers the sweep (no restart needed).
- `BatchTurboQuantKVCache.merge` now rejects mixed `(bits, seed)` members with a clear error — the reporter's remedy (c), as a last-line guard for any path the signature cannot cover.

## Why

TurboQuant packed state width is `ceil(head_dim * bits / 32)`. The signature recorded `layer_cache_types` (knows a layer is TurboQuant) but not *which* depth, so after changing `turboquant_kv_bits` the old-width blocks kept validating. Reproduced on a 0.5B model (head_dim 64, widths 8 vs 12) following the issue's steps: a request whose prefix restored from old 4-bit blocks was silently pinned to 4-bit (the new setting was not honored — post-flip tail blocks saved at the old width), and the moment it was batched with a fresh-prefix request quantized at 6-bit:

```
RuntimeError: [concatenate] All the input array dimensions must match exactly except
for the concatenation axis. However, the provided shapes are (1,2,2607,12), (1,2,2607,8),
and the concatenation axis is 0.
```

— structurally identical to the issue's `(1,2,21061,48)` vs `(1,2,21061,32)` (head_dim 256). Both co-batched requests 500, and the poison is persistent: the blocks live on disk, so every future co-batch repeats the failure until the cache directory is deleted by hand.

A second bug fell out of the repro: the signature refresh (`_infer_live_layer_cache_types`) required `model.make_cache`, which plain dense models don't define — so on the live serving path the refresh was a silent no-op and the manager never learned the layout at all. It now builds its probe cache via `make_prompt_cache`, exactly like the request path.

## Changes

- `_cache_compat_signature` gains `turboquant_kv_bits`; `save_block` stamps it from the block's own meta_state (`(offset, bits, seed)` — the same tuple the restore path reads), falling back to the manager's expectation. A signature never vouches for a width the payload does not have.
- `_signature_bits_match` (one predicate, used by both the compat check and the sweep): with a depth expected, a block must prove a matching depth; unproven blocks (pre-fix saves, metadata-less same-model blocks, corrupted signatures) are treated as stale.
- `set_expected_layer_signature` takes keyword-only `turboquant_kv_bits`; a bits change alone marks the sweep pending.
- `Scheduler._infer_live_layer_cache_types` returns `(layer_cache_types, bits)`; the depth is keyed off the same `_turboquant_eligible` gate the request path uses (models whose convertible caches sit inside `CacheList` report identical type names at every depth — the bits field is their only discriminator). An eligibility-probe failure refuses to infer rather than committing a wrong layout the sweep would act on, and a `logger.warning` fires when TurboQuant is active but no signature could be inferred.
- `_canonicalize_layer_cache_types` collapses `BatchTurboQuantKVCache` → `TurboQuantKVCache` (same packed per-request state on disk; the save path records whichever form it extracted, and sweeping batch-form blocks against the always-singular predicted layout would cold-start the cache every restart).
- `BatchTurboQuantKVCache.merge` raises `ValueError` naming the mixed configs and the remedy, instead of a raw `mx.concatenate` shape error (or, at equal widths, silent garbage decode — seeds determine codec contents).

## Why this is safe

- **Non-TurboQuant models: zero change.** The bits field is only stamped when TurboQuant is active, so their signatures stay byte-identical to the current format and existing caches keep validating (test-pinned).
- **One-time cost, stated plainly:** existing TurboQuant users' blocks predate depth stamping and are swept once after upgrade even if their settings never changed — their width is unverifiable, and keeping them is exactly the reported crash. The sweep drops index entries only; files are reclaimed by the existing LRU.
- **Named residual:** on models where cache-layout inference fails entirely, the first-save `adopt_layer_signature_if_unset` fallback still can't learn bits (pre-fix behavior). That path now warns at startup, and the merge guard converts any resulting mixed-width batch into a clear, actionable error instead of the raw concat failure. Threading bits through the save/adopt plumbing is a possible follow-up if such a model surfaces.
- Seed is part of the merge guard but not the signature: it's hardcoded 0 everywhere today; if it ever becomes configurable it belongs in the signature too (noted for then).

## Tests

`python -m pytest tests/test_paged_ssd_cache.py tests/test_scheduler.py tests/test_turboquant.py -q` → **350 passed** (13 new: signature format pinning old/new, non-dict-JSON tolerance, block-bits extraction from meta_state, compat + sweep depth rules including legacy/unproven blocks, batch/single name canonicalization, bits-change-triggers-sweep, a real-save-path test asserting the persisted file's signature records the depth, scheduler refresh on a make_cache-less dense model, and the merge mixed-config guard). Adjacent suites (`test_turboquant_ssd`, `test_prefix_cache`, `test_prefix_cache_v4_block_storage`, `test_cache_type_handlers`, `test_turboquant_batch_memory`): **195 passed**.

Also verified the new tests fail with the fix reverted (3 behavioral failures plus a collection error on the new helpers).

End-to-end on a live server (Qwen2.5-0.5B, TQ enabled, sandboxed `--base-path`): reproduced the crash on current `main` with the exact issue log shape; on this branch the startup sweep invalidates the mixed-width blocks once (`Invalidated 17 SSD index entries…`), the previously-crashing concurrent request pair completes, new blocks persist with `"turboquant_kv_bits": 6.0` in their signatures, and a subsequent restart keeps them (`kept 17`, sweep idempotent) with full prefix-cache hits.

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -204,6 +204,13 @@ def _canonicalize_layer_cache_types(
     wrapper_to_canonical = {
         "SizedArraysCache": "ArraysCache",
         "PrefillReadyRotatingKVCache": "RotatingKVCache",
+        # Batch and single-request TurboQuant caches persist the same packed
+        # per-request state (the save path records whichever class name it
+        # extracted; the restore path rebuilds a TurboQuantKVCache from
+        # either). Collapsing them keeps the predicted layout from
+        # refresh_ssd_layer_signature — which always says
+        # "TurboQuantKVCache" — from sweeping valid batch-form blocks.
+        "BatchTurboQuantKVCache": "TurboQuantKVCache",
     }
     return [
         wrapper_to_canonical.get(cache_type, cache_type)
@@ -217,6 +224,7 @@ def _cache_compat_signature(
     num_layers: int = 0,
     block_size: int = 0,
     layer_cache_types: list[str] | None = None,
+    turboquant_kv_bits: float | None = None,
 ) -> str:
     """Return a stable compatibility signature for a persisted cache block."""
     payload = {
@@ -225,7 +233,63 @@ def _cache_compat_signature(
         "block_size": int(block_size or 0),
         "layer_cache_types": list(layer_cache_types or []),
     }
+    # TurboQuant packed state width depends on the bit depth
+    # (packed_width = ceil(head_dim * bits / 32)), so blocks written at
+    # different bit depths are shape-incompatible (#2045). Only stamped
+    # when TurboQuant is active so non-TurboQuant signatures stay
+    # byte-identical to the previous format.
+    if turboquant_kv_bits is not None:
+        payload["turboquant_kv_bits"] = float(turboquant_kv_bits)
     return json.dumps(payload, sort_keys=True, separators=(",", ":"))
+
+
+def _signature_turboquant_bits(cache_signature: str) -> float | None:
+    """Extract ``turboquant_kv_bits`` from a stored signature, or None."""
+    if not cache_signature:
+        return None
+    try:
+        payload = json.loads(cache_signature)
+    except (TypeError, ValueError):
+        return None
+    if not isinstance(payload, dict):
+        # Corrupted/foreign signature that parses as a JSON scalar or list.
+        # Report "no recorded depth" instead of raising: an AttributeError
+        # here would abort the whole stale-signature sweep.
+        return None
+    bits = payload.get("turboquant_kv_bits")
+    if bits is None:
+        return None
+    try:
+        return float(bits)
+    except (TypeError, ValueError):
+        return None
+
+
+def _block_turboquant_bits(
+    layer_cache_types: list[str] | None,
+    layer_meta_states: list[tuple] | None,
+) -> float | None:
+    """Read the bit depth a block's own TurboQuant layers were packed at.
+
+    TurboQuant meta_state is ``(offset, bits, seed, ...)`` — the same tuple
+    the restore path reads back at reconstruction. Deriving the signature
+    stamp from the block itself keeps it truthful even when the manager's
+    expectation is stale or not yet learned.
+    """
+    if not layer_cache_types or not layer_meta_states:
+        return None
+    for i, cache_type in enumerate(layer_cache_types):
+        if cache_type not in ("TurboQuantKVCache", "BatchTurboQuantKVCache"):
+            continue
+        if i >= len(layer_meta_states):
+            continue
+        meta_state = layer_meta_states[i]
+        if isinstance(meta_state, (list, tuple)) and len(meta_state) >= 3:
+            try:
+                return float(meta_state[1])
+            except (TypeError, ValueError):
+                return None
+    return None
 
 
 def _clamp_rotating_meta_states(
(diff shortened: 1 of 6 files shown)


jundot:
Thanks for the thorough writeup and tests, clean root-cause fix. I ran the three suites locally (all 350 pass) and confirmed non-turboquant signatures stay byte-identical. Branch is a bit behind main so I'll rebase and merge.

Issue #1067, opened 2026-05-05
Support OpenAI Realtime API


JohnGalt1717:
**Is your feature request related to a problem? Please describe.**
Realtime voice to voice chat and realtime transcription and TTS models.

**Describe the solution you'd like**
I'd like to be able to emulate as a developer (and for openclaw voice chat etc) the OpenAI realtime websocket (at least if not WebRTC) so that we can do similar to LocalAI's Realtime support that has the VAD model that layers on to do the turn and interrupt detection and then => SST => Text Chat => TTS

**Describe alternatives you've considered**
Manually doing this outside of the realtime api, but this then means different modalities in our code and for OpenClaw etc.

**Additional context**
Basically everyone is supporting the OpenAI Realtime API. xAI, Google and I believe Anthropic all support it along with OpenAI, and many open source models are adopting it too like Mistral STT and TTS models.

PR #2075, opened 2026-07-03
feat(server): add stream=true SSE support to /v1/audio/transcriptions

### LINKED ISSUE #1066: Transcription/Translation Endpoint support: /v1/audio/transcriptions (and /v1/audio/translations)
Right now the only way to do transcription is by passing binary attachments to the chat endpoint for AIs that support transcription/translation

I would like to do streaming STT so that I can emulate a production environment for development and testing and of course for openclaw etc.

See above. Using binary without streaming of wav/mp3 etc. files which are huge.

**Additional context**
This will create more parity in the surface area of the openai api and enable streaming STT long with the existing TTS in the project.

Also #591 is related per above.

ethannortharc:
Closes #1066

## What

Adds OpenAI-compatible streaming to the speech-to-text endpoint: `POST /v1/audio/transcriptions` now accepts a `stream` form field. With `stream=true` the response switches to OpenAI's transcription SSE format — `transcript.text.delta` events with incremental text, followed by a final `transcript.text.done` event carrying the full transcript and token usage:

```
data: {"type": "transcript.text.delta", "delta": "Hello"}

data: {"type": "transcript.text.delta", "delta": " world"}

data: {"type": "transcript.text.done", "text": "Hello world...", "usage": {"type": "tokens", "input_tokens": 73, "output_tokens": 16, "total_tokens": 89}}
```

Default (`stream=false`) behavior is unchanged.

## How

- **`STTEngine.transcribe_stream()`** — async generator that drives mlx-audio's native `generate(..., stream=True)` on the MLX executor, reusing the sync→async bridge pattern already used by native TTS streaming. Chunks are normalized (delta text, language, cumulative token counts) across backends that yield `StreamingResult` objects or plain strings.
- **Capability detection** — `supports_native_stt_streaming()` checks for a `stream` parameter in the model's `generate()` signature (whisper, qwen3-asr, parakeet, canary, moonshine, granite, cohere-asr, voxtral-realtime all have it). Models without it fall back to one-shot `transcribe()` and emit the full text as a single delta, so `stream=true` works for every STT model.
- **Route** — the SSE generator takes ownership of the uploaded temp file and deletes it when the stream completes, errors, or the client disconnects. The first event is prefetched before the response starts so engine failures still surface as proper HTTP 500s instead of a broken stream.
- `word_timestamps` (oMLX extension) is ignored under `stream=true`, matching OpenAI, which also rejects timestamp granularity in streaming mode.

## Testing

- 9 new unit tests (engine chunk normalization, language-hint normalization, non-streaming fallback, capability detection, SSE event shape, empty-delta filtering, kwargs forwarding, non-stream regression, mid-stream error → 500). Full suite: 6011 passed.
- End-to-end on Apple Silicon:
  - `mlx-community/Qwen3-ASR-0.6B-bf16` — token-level deltas over HTTP via curl, correct done event + usage.
  - `mlx-community/whisper-large-v3-turbo-asr-fp16` — chunk-level streaming (AlignAtt), full transcript intact across 5 deltas.
  - **Official `openai-python` client** parses the stream out of the box:

```python
stream = client.audio.transcriptions.create(model="Qwen3-ASR-0.6B-bf16", file=f, stream=True)
for event in stream:
    if event.type == "transcript.text.delta":
        print(event.delta, end="")
```

Realtime WebSocket input (#1067) is intentionally out of scope — this PR covers streaming of results per the OpenAI `stream=true` semantics requested in #1066.


ethannortharc:
Updated: this branch now includes the OpenAI `prompt` vocabulary-biasing support from #2082 (merged in, plus one commit wiring it through the streaming path), so `stream=true` requests get the same biasing as one-shot transcription — Qwen3-ASR-style backends forward it natively through their stream path as `system_prompt`, Whisper-family models receive `initial_prompt`, and backends without a hook ignore it.

Verified end-to-end in stream mode with `mlx-community/Qwen3-ASR-0.6B-bf16`: "omlx" inside a Chinese utterance transcribes as "Omnux" without `prompt` and correctly as "Omlx" with `prompt="Vocabulary: issue, GitHub, omlx."`. Full suite: 6020 passed.

If #2082 merges first, this branch reduces to the streaming change alone; the two are independent to review.

One upstream observation unrelated to this PR's code but visible through it: token-level streaming backends in mlx-audio (e.g. Qwen3-ASR's `stream_transcribe`) decode each token independently (`tokenizer.decode([token])`), so a multi-byte CJK character split across BPE tokens surfaces as U+FFFD in the deltas (and therefore in the joined `transcript.text.done` text). Chunk-level backends (Whisper) and non-streaming requests are unaffected. The proper fix is a rolling decode in mlx-audio; happy to file it there.

jundot:
Thanks for the thorough work here, the temp file ownership handoff and the first-event prefetch are both handled carefully. I merged #2082 first so this reduces to the streaming change, resolving the pyproject overlap the same way.

PR #2076, opened 2026-07-03
feat: GLM-5.2 native MTP speculative decoding (external head)

anicaise-ai:
## Summary

Extends the native MTP speculative decoding (`patches/mlx_lm_mtp/`, currently Qwen 3.5/3.6 + DeepSeek-V4-Flash) to **GLM-5.2** (`glm_moe_dsa`). GLM-5.2's MTP layer was explicitly designed by Zhipu as the speculative draft model ([GLM-5.2 blog](https://huggingface.co/blog/zai-org/glm-52-blog)), and oMLX already has the best GLM-5.2 story on Apple Silicon — this completes it.

## What's different about GLM-5.2

1. **External head checkpoint.** Common quantized GLM-5.2 builds (e.g. mxfp4) strip the `mtp.*` tensors; the head circulates as a separate ~11 GB checkpoint (`model_type: glm_moe_dsa_mtp`, e.g. `GLM-5.2-MTP-MLX-Q4`). So instead of the in-checkpoint `__init__`/`sanitize` hooks used by Qwen/DeepSeek, the head is **attached post-load** from a `<model_dir>/mtp/` directory (dir or symlink), gated on `model_settings.mtp_enabled`. The strict main-checkpoint load is untouched and with MTP off the model is byte-identical to stock. `gate_up_proj` fusion of the vendored MoE is replicated for the head weights (concat on the expert output axis).

2. **Aligned head cache (gap-fill).** On the accept path the generic cycle inserts only one `(hidden, token)` pair per two confirmed tokens, so the head's KV/RoPE/DSA-indexer positions drift one position per accepted draft — with GLM's positional-sensitive head the accept rate collapses. This PR inserts the accepted draft's pair as well, **gated on a `_omlx_mtp_aligned_cache` model marker** so Qwen/DeepSeek behavior is unchanged. The fill forward's logits are never evaluated, so the lm_head matmul is lazily skipped. With the gap-fill, accept goes from collapse to **80–92 % (greedy)**.

## Measurements (M3 Ultra 512 GB, GLM-5.2-mxfp4 + GLM-5.2-MTP-MLX-Q4, greedy, 3 fixed prompts)

| | decode tok/s | accept |
|---|---|---|
| oMLX HEAD, no MTP | 18.6–19.7 | — |
| this PR, MTP on | 15.7–16.4 | 79.8 / 87.6 / 91.7 % |

Being upfront: **on this bandwidth-bound setup the net decode is parity-to-slightly-below.** Per-request timing shows why: the L=2 verify forward costs ~2.2× an L=1 forward (112 ms vs 51 ms) because with 8-of-256 routing two adjacent tokens share almost no routed experts, and routed expert weights dominate the memory traffic. The MTP plumbing is correct and measured (the accept rates above are the proof); it becomes a net win wherever small-L expert gather amortizes better — see the companion issue with the full numbers. Until then, `mtp_enabled: false` (the default) keeps everything stock.

## Test plan

- `mtp_enabled: false` (default): no behavioral change (verified byte-identical outputs at temp 0 on 3 prompts).
- `mtp_enabled: true` + `<model_dir>/mtp` present: head attaches post-load (log line), MTP path activates for singleton batches, per-request stats logged (`MTP[uid] ... accept=...`).
- `mtp_enabled: true` without the head dir: warning, MTP inactive, serving unaffected.
- Attach failure (weight mismatch): logged, serving continues without MTP.

### DIFF
--- a/omlx/patches/mlx_lm_mtp/__init__.py
+++ b/omlx/patches/mlx_lm_mtp/__init__.py
@@ -78,7 +78,13 @@ def apply_mlx_lm_mtp_patch() -> bool:
         True if the patch is now active. False if a sub-step refused
         to apply (mlx-lm not importable, missing prerequisite patch).
     """
-    from . import batch_generator, cache_rollback, deepseek_v4_model, qwen35_model
+    from . import (
+        batch_generator,
+        cache_rollback,
+        deepseek_v4_model,
+        glm_moe_dsa_model,
+        qwen35_model,
+    )
 
     if not cache_rollback.apply():
         return False
@@ -88,6 +94,8 @@ def apply_mlx_lm_mtp_patch() -> bool:
         logger.debug("Qwen3.5/3.6 MTP patch did not apply (likely import error)")
     if not deepseek_v4_model.apply():
         logger.debug("DeepSeek-V4 MTP patch did not apply (likely missing base patch)")
+    if not glm_moe_dsa_model.apply():
+        logger.debug("GLM-5.2 MTP patch did not apply (likely non-GLM model)")
     if not batch_generator.apply():
         logger.warning(
             "BatchGenerator MTP dispatch patch failed; MTP path will be inactive"
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -1541,6 +1541,18 @@ def _run_verify_cycle(gen_batch: Any, state: _MtpState) -> None:
         _clear_rollback(gen_batch.prompt_cache)
         state.stats.cache_ops_ms += (time.perf_counter() - t0) * 1000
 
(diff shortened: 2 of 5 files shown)


anicaise-ai:
Companion issue with the small-L expert-gather analysis and numbers: #2077

jundot:
Thanks for the detailed work and analysis here.

I especially appreciate the investigation into GLM-5.2’s external MTP head, the post-load attachment path, and the cache-alignment issue. The gap-fill analysis was useful for understanding why acceptance collapses when the head cache drifts.

I have since merged Lightning MTP in oMLX: [#2113](https://github.com/jundot/omlx/pull/2113). That path now covers GLM-5.2 with depth-k draft/verify decoding, adaptive depth, and GLM-specific short-verify handling, so I think it is the better direction for using MTP effectively with GLM-5.2.

I am going to close this PR in favor of that implementation. Thanks again for the concrete measurements and root-cause analysis.

PR #2084, opened 2026-07-04
feat(server): add GET /v1/audio/voices to list TTS speaker names

ethannortharc:
## What

New endpoint `GET /v1/audio/voices?model=<id>` returning the built-in speaker/voice names a TTS model supports:

```json
{"model": "Kokoro-82M-bf16", "voices": ["af_alloy", "af_aoede", "...54 total"]}
{"model": "Qwen3-TTS-12Hz-1.7B-CustomVoice-bf16", "voices": ["aiden", "dylan", "eric", "ono_anna", "ryan", "serena", "sohee", "uncle_fu", "vivian"]}
```

## Why

TTS clients currently have no way to discover valid `voice` values for `/v1/audio/speech` — they hardcode lists that drift from the models actually installed, or let users guess and get engine errors. The voice inventory is static model metadata, so the server can answer instantly without loading the model.

## How

Reads static files only (no engine load, no inference):

- **Kokoro-style models**: voice names from the `voices/` directory, deduplicated by stem across the `.safetensors`/`.pt` duplicates Kokoro repos ship (Kokoro-82M: 108 files → 54 voices).
- **Qwen3-TTS CustomVoice**: speaker names from `talker_config.spk_id` in `config.json` (9 speakers, verified against the mlx-community conversion).
- **Models with neither** (e.g. zero-shot voice-cloning bases like VibeVoice): empty list, not an error — no named speakers is a legitimate answer.

Model aliases resolve through the same `resolve_model_id` path as the other audio endpoints. Unknown model → 404, missing `model` param → 400.

## Testing

6 new unit tests covering the Kokoro directory scan (format dedup, non-voice files ignored), the CustomVoice config path, empty-inventory models, bare directories, and both error cases. Full suite: 6008 passed. Verified against real local models on Apple Silicon (Kokoro-82M-bf16 → 54 voices, CustomVoice → 9 speakers).


jundot:
Thanks, this fills a real gap for TTS clients and keeping it static-metadata-only so nothing gets loaded is the right call. Verified against the Kokoro and CustomVoice layouts, tests pass locally. I'll merge this.

ethannortharc:
> Thanks, this fills a real gap for TTS clients and keeping it static-metadata-only so nothing gets loaded is the right call. Verified against the Kokoro and CustomVoice layouts, tests pass locally. I'll merge this.

Sounds good. Thanks for verifying and merging!

Issue #762, opened 2026-04-14
Feature Request: Native Support for DFlash in Omlx


Revive-Curiosity:
**Is your feature request related to a problem? Please describe.**  
Currently, Omlx does not support **[[DFlash: Block Diffusion for Flash Speculative Decoding](https://arxiv.org/html/2602.06036v1)](https://arxiv.org/html/2602.06036v1)** natively. This limits speculative decoding efficiency, resulting in lower acceptance lengths and throughput compared to what DFlash can achieve. For workloads requiring high performance inference, this becomes a bottleneck.  

**Describe the solution you'd like**  
Add native support for DFlash within the Omlx project. This would allow acceptance lengths to increase to ~6–8 tokens per token verified by target model and throughput improvements of up to **3.5×–6×** compared to standard autoregressive decoding.  

**Describe alternatives you've considered**  
- Using standard autoregressive decoding (baseline, but limited to acceptance length of 1).  
- Implementing speculative decoding with other draft models, though these approaches typically yield lower acceptance lengths and less throughput improvement compared to DFlash.  

**Additional context**  
- DFlash paper: [https://arxiv.org/html/2602.06036v1](https://arxiv.org/html/2602.06036v1)  
- Benchmarks show DFlash consistently achieves acceptance lengths between 6–8 tokens per block, leading to significant speedups in inference throughput.

TomLucidor:
Would you recommend PFlash + DDTree?

Issue #761, opened 2026-04-14
Dumb question about DFlash-MLX


cresseelia:
| Target model | Draft checkpoint |
|--------------|-----------------|
| Qwen/Qwen3.5-4B | z-lab/Qwen3.5-4B-DFlash |
| Qwen/Qwen3.5-9B | z-lab/Qwen3.5-9B-DFlash |
| Qwen/Qwen3.5-27B | z-lab/Qwen3.5-27B-DFlash |
| mlx-community/Qwen3.5-27B-8bit | z-lab/Qwen3.5-27B-DFlash |
| mlx-community/Qwen3.5-27B-4bit | z-lab/Qwen3.5-27B-DFlash |
| Qwen/Qwen3.5-35B-A3B | z-lab/Qwen3.5-35B-A3B-DFlash |
| mlx-community/Qwen3.5-35B-A3B-4bit | z-lab/Qwen3.5-35B-A3B-DFlash |

here's what I see in the docs about dflash-mlx, the question is about oQ quant Qwen3.5 model

is your model https://huggingface.co/Jundot/Qwen3.5-35B-A3B-oQ4e compatible with z-lab/Qwen3.5-35B-A3B-DFlash

and is other Qwen3.5-27b oQ quant variant models compatible with z-lab/Qwen3.5-27B-DFlash?

and also will you release Qwen3.5-27b oQe quant  and  other Qwen3.5 oQe quant model in the future?

Thanks!

0xClandestine:
yes it's compatible

Revive-Curiosity:
Feature Request : Support Dflash Natively in Omlx

Omlx Team, 

Please add Dflash natively to omlx project. So the acceptance length and throughput could be increased upto 6x. 

thankyou.

0xClandestine:
I wrote the first working implementation: https://github.com/0xClandestine/mirror-sd

happy to collaborate on this @jundot

PR #2085, opened 2026-07-04
fix(ci): pin transformers below 5.13 to unbreak mlx-lm import

ethannortharc:
## What

Caps `transformers` at `<5.13` in `pyproject.toml`.

## Why — CI is red for every PR since today

transformers **5.13.0** (released 2026-07-03) tightened `_LazyAutoMapping.register()` — it now accesses `key.__module__` on the config-class key. mlx-lm 0.31.3 (the commit pinned in this repo) still registers its `NewlineTokenizer` with a **string** key:

```python
# mlx_lm/tokenizer_utils.py:503
AutoTokenizer.register("NewlineTokenizer", fast_tokenizer_class=NewlineTokenizer)
```

so on any freshly resolved environment, `import mlx_lm` — and therefore `import omlx.server`, and therefore **all test collection** — dies with:

```
transformers/models/auto/auto_factory.py:680: in register
    if key.__module__.startswith("transformers."):
E   AttributeError: 'str' object has no attribute '__module__'
```

`pyproject.toml` only required `transformers>=5.7.0`, so CI picks up 5.13.0 the moment it resolves. main's last green run (Jul 2) predates the release; every run since — CI #761, #762, #763 across PRs — fails identically with 54 collection errors. Environments resolved before Jul 3 (including dev machines) are unaffected, which is why this only shows up in CI.

## Lift condition

Remove the cap once the mlx-lm pin advances past a release that registers the tokenizer with a config class (or transformers restores string-key tolerance).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,13 @@ dependencies = [
     # mlx-embeddings from latest commit (32981fa)
     "mlx-embeddings @ git+https://github.com/Blaizzy/mlx-embeddings@32981fa4e8064ed664b52071789dd18271fe4206",
     # mlx-vlm custom processors bypass HF AutoProcessor, so torch is not required
-    "transformers>=5.7.0",
+    # <5.13: transformers 5.13.0 (released 2026-07-03) tightened
+    # _LazyAutoMapping.register to require config classes, breaking mlx-lm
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for catching this quickly. The cap matches the mlx-lm tokenizer registration failure, and the rerun CI confirms fresh installs are back to green. LGTM. I will merge this now.

PR #2086, opened 2026-07-04
fix: Kokoro model discovery and G2P language inference

ethannortharc:
Two fixes that together make mlx-community Kokoro conversions work out of the box.

## 1. Discovery: Kokoro exports classified as LLM → "No safetensors found"

mlx-community Kokoro conversions (e.g. `Kokoro-82M-bf16`) keep the **original** Kokoro `config.json` — top-level `istftnet`/`plbert`/`vocab` sections, no HF-style `model_type` or `architectures` — and name their weights `kokoro-v1_0.safetensors`. Model discovery found neither field, fell back to plain LLM, and handed the model to the mlx_lm loader, whose glob only matches `model*.safetensors`. Result: a misleading **"No safetensors found"** on every load attempt even though the files were complete.

Fix: detect the Kokoro config shape (`istftnet` + `plbert` + `vocab`, all dicts) during discovery and classify as `audio_tts`, mirroring the existing NeMo-ASR-without-model_type precedent. Includes a guard test that a partial match (vocab table alone) still defaults to LLM.

## 2. TTS: Kokoro G2P lang_code inferred from the voice name

Kokoro voice names are `<lang><gender>_<name>` (`af_heart`, `bm_george`, `zf_xiaoxiao`) where the first letter **is** the G2P pipeline language code. oMLX only forwarded `lang_code` when the request carried an explicit `language`, so OpenAI-style clients that just pick a voice got English G2P for every language — non-English text was mangled or dropped.

When the request has no `language` and the model's `generate()` accepts both `voice` and `lang_code`, infer the code from the full Kokoro prefix `^[abefhijpz][fm]_` — covering all nine Kokoro languages (en-US/en-GB/es/fr/hi/it/ja/pt-br/zh), not just one or two. The full-prefix match matters: other backends whose `generate()` also takes `voice` + `lang_code` (Qwen3-TTS CustomVoice's `aiden`/`eric`, bailingmm, longcat) must never trigger inference — their speaker names don't match the convention, so they keep their own defaults (e.g. `auto`). An explicit `language` always wins.

## Testing

- 5 discovery tests (config-shape detection, partial-config guard, discover_models integration) + 5 lang-inference tests (all 9 languages, non-Kokoro name rejection incl. Qwen3-TTS speakers, explicit-language precedence).
- End-to-end on Apple Silicon with `mlx-community/Kokoro-82M-bf16`: `detect_model_type` → `audio_tts`; `zf_xiaoxiao` synthesizes Mandarin correctly with no `language` field (3.9s audio); `af_heart` English unaffected.
- Full suite: 6005 passed.

(Last commit is the `transformers<5.13` CI pin from #2085 so CI can run green here; it disappears on rebase once that merges.)

### DIFF
--- a/omlx/engine/tts.py
+++ b/omlx/engine/tts.py
@@ -11,6 +11,7 @@ when mlx-audio is not installed.
 import asyncio
 import gc
 import logging
+import re
 from collections.abc import AsyncIterator
 from typing import Any, Dict, Optional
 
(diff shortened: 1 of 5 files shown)


jundot:
Thanks for the fix. I checked both paths: the Kokoro config-shape detection keeps the discovery heuristic narrow, and the voice-prefix lang_code inference only applies to Kokoro-style names while explicit language still wins.

The targeted discovery/TTS tests pass locally, and CI is green across the Python matrix. LGTM.

PR #2087, opened 2026-07-04
fix(engine_pool): reclaim memory after failed model loads

tbro0815:
## Problem

A failed model load can leave the partially loaded weights unreclaimable until a server restart. When a loader raises partway through (weights loaded, then e.g. processor construction fails), the weights remain reachable through the propagating exception's traceback frames (`exc.__traceback__` → loader frames → locals). `gc.collect()` + `mx.clear_cache()` at raise time cannot free them because the exception is still alive in the caller.

Observed in production: one failed load of a 119B model left ~67 GB in the process footprint; a retry pushed it to ~100 GB, and the memory-ceiling admission check then rejected every subsequent load ("projected memory would exceed the memory ceiling") until restart.

## Fix

On load failure, `_load_engine` now schedules a deferred background reclaim: wait briefly for the exception to be handled and dropped by the caller, then run `gc` + `mx.synchronize` + `mx.clear_cache` rounds until the live memory gauge (max of `mx.get_active_memory()` and phys footprint) returns near its pre-load level, or rounds are exhausted (logged either way). `_wake_process_memory_enforcer()` is triggered after settling.

## Validation

Deployed on an M-series machine (128 GB): after a failed 119B VLM load, the log shows

```
Reclaimed memory after failed load of 'Mistral-Small-4-...': current=482.42MB (pre-load=120.72MB)
```

within ~1.7 s of the failure; subsequent loads admit normally. Also observed recovering a failed load through the standard VLM engine path (`current=826.70MB, pre-load=784.53MB`).

### DIFF
--- a/omlx/engine_pool.py
+++ b/omlx/engine_pool.py
@@ -1243,6 +1243,55 @@ class EnginePool:
 
         self._wake_process_memory_enforcer()
 
+    def _schedule_failed_load_reclaim(
+        self, model_id: str, pre_load_memory: int
+    ) -> None:
+        """Reclaim memory left behind by a failed model load.
+
+        When a load raises partway through (e.g. weights loaded, processor
+        construction failed), the weights are often reachable only via the
+        propagating exception's traceback frames. Spawn a background task
+        that waits briefly for the exception to be handled and dropped, then
+        runs gc + synchronize + clear_cache rounds until the live memory
+        reading returns near its pre-load level (or rounds are exhausted).
+        """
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the fix. The deferred reclaim matches the traceback-retention failure mode that remains after the unload/teardown cleanup work, and the engine_pool tests pass locally.

The small test-style cleanup is non-blocking; I will fold that into a follow-up. LGTM.

PR #2095, opened 2026-07-05
fix: remove stale "auto" from Hot Cache Size sublabel

mxl:
The Hot Cache Size field description in the macOS app claimed "auto" is accepted, but the code normalizes it to "0" (disabled). This misleads users into entering "auto" which silently becomes disabled.

Updated both the localization string and Swift defaultValue to accurately reflect accepted values: "0" or a size like "8GB".

### DIFF
--- a/apps/omlx-mac/Resources/Localizable.xcstrings
+++ b/apps/omlx-mac/Resources/Localizable.xcstrings
@@ -3871,7 +3871,7 @@
         "en" : {
           "stringUnit" : {
             "state" : "translated",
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for fixing this. The server behavior matches the updated copy: new `hot_cache_max_size=auto` updates are rejected, while legacy persisted `auto` values load as disabled.

There is one remaining localized string on current `main` that still mentions `auto`, and I will handle that in a follow-up. LGTM, merging.

Issue #1568, opened 2026-05-31
[Bug] Harmony models long generations crash with boundary cache snapshot store


tylerliu:
**Describe the bug**
Harmony/GPT-OSS models can crash at the end of a long generation when the scheduler stores a boundary cache snapshot. The output itself appears to finish successfully: the Harmony parser parses both messages, then the scheduler enters the boundary cache snapshot store path and the process aborts with There is `no Stream(gpu, X)` in current thread.

**To Reproduce**
1. Running a Harmony Model (GPT-OSS) with a prompt that generates over 1024 tokens. 
2. See error

**Expected behavior**
oMLX server should not crash at the end of the output. 

**Logs**
```bash
➜  omlx git:(main) git show HEAD | head                            
commit a108055bcb7e313f259ba24ff5e4c9d4b36402f4
Author: jundot <jundot@users.noreply.github.com>
Date:   Sun May 31 21:22:53 2026 +0900

    feat: add swift-only mac app rebuild

diff --git a/apps/omlx-mac/Scripts/build.sh b/apps/omlx-mac/Scripts/build.sh
index 5622752..c6958cd 100755
--- a/apps/omlx-mac/Scripts/build.sh
+++ b/apps/omlx-mac/Scripts/build.sh
➜  omlx git:(main) uv run omlx serve --model-dir ~/.lmstudio/models
oMLX - LLM inference, optimized for your Mac
├─ https://github.com/jundot/omlx
└─ Version: 0.3.12

2026-05-31 11:42:54,659 - omlx.settings - INFO - Saved settings to /Users/tyler/.omlx/settings.json
Saved CLI arguments to settings.json
Log directory: /Users/tyler/.omlx/logs
Binding server at http://127.0.0.1:8000
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
Base path: /Users/tyler/.omlx
Model directories: /Users/tyler/.lmstudio/models
Memory guard tier: balanced
Mode: Multi-model serving (continuous batching + paged SSD cache)
paged SSD cache: /Users/tyler/.omlx/cache (max: 185.0GB)
Hot cache: 4.0GB (in-memory)
2026-05-31 11:42:54,846 - omlx.server - INFO - CORS origins: ['*']
```
...
```
Starting server at http://127.0.0.1:8000
INFO:     Started server process [6606]
INFO:     Waiting for application startup.
[...]

PR #2103, opened 2026-07-06
Fix boundary snapshot writer retaining raw tensor bytes after unload

zwcf5200:
## Summary

This PR fixes a long-running memory retention path in the SSD boundary snapshot and paged SSD cache writer threads.

The issue can show up after long-context workloads that create boundary snapshots or store-cache writes. The model can be fully unloaded and the API-visible model memory can return to zero, while the Python process still retains gigabytes of host malloc memory. In a local long-run incident, the service had `models_loaded=0`, `model_memory_used=0B`, and MLX active memory settled back to `1.02KB`, but `vmmap` still showed `Physical footprint=3.4G` with `MALLOC_LARGE=3.1G` and `heap` showed roughly `3.47GB` of non-object allocations. Clearing the hot cache reclaimed nothing, which ruled out the visible hot cache as the retention owner.

The root cause is that the background writer thread can keep the last queue item alive in its idle frame. For boundary snapshots, that queue item contains `tensors_raw`, the raw bytes copied from KV/cache tensors before writing safetensors files. The scheduler also dropped its `BoundarySnapshotSSDStore` reference during teardown without stopping the store, so the daemon writer thread could continue to hold the store and its last queue item even after cleanup.

## Changes

- Clear the local `item` reference in `BoundarySnapshotSSDStore._writer_loop()` after every processed queue item, so the idle writer frame cannot pin the previous snapshot raw bytes.
- Clear the equivalent raw tensor payload locals in `PagedSSDCacheManager._writer_loop()` after each background write.
- Make `Scheduler.shutdown()` clean up and stop the boundary snapshot store before dropping the reference.
- Make `Scheduler.deep_reset()` stop the boundary snapshot store before releasing scheduler cache references.
- Add regression tests covering both scheduler teardown paths.

## Why this matters

This retention path is easy to miss because the normal model accounting looks healthy:

- `/api/status` can report `models_loaded=0` and `model_memory_used=0B`.
- Engine unload logs can report active memory settling back to `1.02KB`.
- `IOAccelerator` can stay low, so this is not the old Metal-backed model memory leak shape.
- Hot-cache clearing can reclaim zero bytes because the retained owner is the writer thread frame, not a visible cache table.

Without stopping the boundary snapshot store and clearing the writer frame locals, a production service can appear idle but still retain large raw snapshot byte buffers after long-context workloads. The fix is intentionally narrow: it does not change cache formats, cache lookup behavior, or request scheduling; it only ensures teardown and idle writer frames release large temporary payloads.

## Local validation

Before the fix, a long-running local service reproduced the problem after boundary snapshot and store-cache activity:

- `models_loaded=0`, `active_requests=0`, `waiting_requests=0`, `model_memory_used=0B`
- `vmmap -summary`: `Physical footprint=3.4G`, `MALLOC_LARGE=3.1G`, `IOAccelerator (graphics)=4064K`
- `heap -s`: `non-object=3.47GB`, with large blocks around 2MB and 3MB
- `/admin/api/hot-cache/clear`: `bytes_reclaimed=0`

After applying the fix and restarting from the patched source, the same long-running service stayed stable after several hours and hundreds of requests:

- Runtime: about 7h14m
- Requests: 482
- Prompt tokens: 18,346,560
- Cached tokens: 17,214,464
- `models_loaded=0`, `active_requests=0`, `waiting_requests=0`, `model_memory_used=0B`
- Admin memory pressure: ok, about 0.3GB current bytes
- `vmmap -summary`: `Physical footprint=281.9M`, `MALLOC_LARGE=3760K`, `IOAccelerator (graphics)=4992K`
- Latest unload sequence ended with `active_memory: 1.02KB (settled)`

This indicates the previous multi-GB raw-byte retention did not recur under the same class of long-context boundary snapshot workload.

## Tests

Passed:

```bash
git diff --check upstream/main...HEAD
.venv/bin/python -m py_compile \
  omlx/cache/boundary_snapshot_store.py \
  omlx/cache/paged_ssd_cache.py \
  omlx/scheduler.py \
  tests/test_scheduler.py
.venv/bin/python -m pytest \
  tests/test_scheduler.py::TestSchedulerReset::test_shutdown_closes_boundary_snapshot_store \
  tests/test_scheduler.py::TestSchedulerReset::test_deep_reset_closes_boundary_snapshot_store -q
.venv/bin/python -m pytest \
  tests/test_boundary_snapshot_store.py::TestBoundarySnapshotSSDStore::test_cleanup_all \
  tests/test_scheduler.py::TestSchedulerReset::test_shutdown_drains_after_bounded_wait -q
```

The targeted pytest commands passed with 2 tests each.

Attempted but blocked by the current dependency resolver setup, not by this patch:

```bash
uv run ruff check omlx/cache/boundary_snapshot_store.py omlx/cache/paged_ssd_cache.py omlx/scheduler.py tests/test_scheduler.py
```

`uv` failed dependency resolution for the Python 3.14 split because `omlx[bundle]` depends on `num2words>=0.5.14` and the resolver could not find a satisfiable package for that split. The local venv also does not currently include `ruff`.

This PR fixes the memory retention path that was reported in https://github.com/jundot/omlx/issues/1568. The boundary snapshot writer thread keeps the last queue item alive (containing tensors_raw, raw bytes from KV/cache tensors) after model unload, causing multi-GB MALLOC_LARGE retention even when models_loaded=0 and MLX memory is zero.

The fix ensures teardown in Scheduler.shutdown() / deep_reset() properly stops the boundary snapshot store and clears writer frame locals, releasing large temporary payloads.

### DIFF
--- a/omlx/cache/boundary_snapshot_store.py
+++ b/omlx/cache/boundary_snapshot_store.py
@@ -509,6 +509,7 @@ class BoundarySnapshotSSDStore:
     def _writer_loop(self) -> None:
         """Background thread that writes safetensors files."""
         while not self._shutdown.is_set():
+            item = None
             try:
                 item = self._write_queue.get(timeout=1.0)
             except queue.Empty:
@@ -522,8 +523,15 @@ class BoundarySnapshotSSDStore:
             # rmtree the snapshot directory while we're mid-write and
             # we'd recreate ``req-X/`` underneath it, leaving an
             # orphaned file after the cleanup returns.
-            with self._writer_busy:
-                self._process_write_item(item)
+            try:
+                with self._writer_busy:
+                    self._process_write_item(item)
+            finally:
+                # ``item`` contains ``tensors_raw``. If the thread waits for
+                # the next queue entry without clearing this local, the frame
+                # can pin a whole boundary snapshot after pending_writes was
+                # cleaned up.
+                item = None
 
     def _process_write_item(self, item) -> None:
         """Process one (pw_key, tensors_raw, metadata, file_path) queue item.
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -1683,6 +1683,7 @@ class PagedSSDCacheManager(CacheManager):
         standard file I/O operations.
         """
         while True:
+            item = None
             try:
                 item = self._write_queue.get(timeout=1.0)
             except queue.Empty:
@@ -1695,12 +1696,18 @@ class PagedSSDCacheManager(CacheManager):
                 break
 
             block_hash, tensors_raw, metadata, file_path = item
-            self._write_block_file(
-                block_hash, tensors_raw, metadata, file_path, source="background"
-            )
-            self._clear_pending_write(
-                block_hash, remove_hot_cache=not self._hot_cache_enabled
-            )
+            try:
+                self._write_block_file(
+                    block_hash, tensors_raw, metadata, file_path, source="background"
+                )
+                self._clear_pending_write(
+                    block_hash, remove_hot_cache=not self._hot_cache_enabled
+                )
+            finally:
+                # Avoid pinning the last raw tensor-byte batch while the
+                # writer thread blocks waiting for more work.
+                item = None
+                block_hash = tensors_raw = metadata = file_path = None
 
     def save_block(
         self,
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -9877,6 +9877,11 @@ class Scheduler:
         self.tokenizer = None
 
         # Release all cache-related references for GC
+        if self._boundary_snapshot_store is not None:
+            try:
+                self._boundary_snapshot_store.shutdown()
+            except Exception as e:
+                logger.warning("Boundary snapshot store shutdown error: %s", e)
         self.paged_cache_manager = None
         self.block_aware_cache = None
         self.memory_monitor = None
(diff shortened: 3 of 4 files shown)


zwcf5200:
This PR fixes the memory retention path that was reported in #1568. The boundary snapshot writer thread keeps the last queue item alive (containing `tensors_raw`, raw bytes from KV/cache tensors) after model unload, causing multi-GB MALLOC_LARGE retention even when `models_loaded=0` and MLX memory is zero.

The fix ensures teardown in `Scheduler.shutdown()` / `deep_reset()` properly stops the boundary snapshot store and clears writer frame locals, releasing large temporary payloads.

jundot:
Thanks for tracking this down. The change is narrow and the teardown ordering looks right: shutdown drains async store-cache work before cleaning and stopping the boundary snapshot store, and the writer loops no longer keep the last raw byte payload alive while idle. The targeted scheduler and boundary snapshot tests pass locally. LGTM.

zwcf5200:
Thank you for reviewing and merging this so quickly.

I wanted to add one more validation note after adopting the same fix in a long-running local service. The result is very encouraging and gives strong confidence that this patch addresses the real retention path.

Before the fix, the service could unload all models and report zero API-visible model memory while the process still retained multiple gigabytes of malloc memory:

- `models_loaded=0`, `active_requests=0`, `waiting_requests=0`, `model_memory_used=0B`
- MLX active memory settled back to `1.02KB`
- `vmmap -summary`: `Physical footprint=3.4G`, `MALLOC_LARGE=3.1G`, `IOAccelerator (graphics)=4064K`
- `heap -s`: about `3.47GB` non-object allocations
- hot-cache clear reclaimed `0` bytes

After adopting this PR, the same local service has now been running for about 70.8 hours under normal long-context use, including repeated boundary snapshot / store-cache activity and TTL model unloads. Current validation data:

- uptime: `254,724s` (~70.8h)
- total requests: `1,637`
- total prompt tokens: `61,164,323`
- total cached tokens: `57,254,656`
- cache efficiency: `93.6%`
- current API state: `models_loaded=0`, `active_requests=0`, `waiting_requests=0`, `model_memory_used=0B`
- admin memory pressure: `390,551,624` bytes (`0.4GB`), `pressure_level=ok`
- `vmmap -summary`: `Physical footprint=372.5M`, `MALLOC_LARGE=3760K`, `IOAccelerator (graphics)=9296K`
- `DefaultMallocZone BYTES ALLOCATED=244.1M`
- latest TTL unload log: unloaded the 35B model, `freed=22.04GB`, `active_memory: 1.02KB (settled)`

So the previous 3.4GB retained-footprint / 3.1GB `MALLOC_LARGE` pattern has not recurred after long-running usage. After model unload, the process now consistently settles around the 370-390MB physical-footprint range, which is exactly the behavior this fix was intended to restore.

Thanks again for accepting the patch.

PR #2106, opened 2026-07-06
Hide helper/draft models and add per-model visibility in /v1/models

jasonpaulso:
## Summary

oMLX discovers and lists every model directory on disk, including speculative-decoding **helper** checkpoints (dFlash drafts, MTP/assistant drafters, SpecPrefill drafts) that are never meant to be called as standalone chat models. They clutter `/v1/models` and there is currently no way to hide an arbitrary model from the OpenAI model list.

This adds two independent, persistent, **opt-in** controls, unioned at the `/v1/models` choke point. Both default off, so existing behavior is unchanged. The admin UI **always shows every model** (badged) so nothing becomes unmanageable — filtering only affects the OpenAI `/v1/models` response. `/v1/models/status` is intentionally left complete as a diagnostic surface.

## Controls

1. **Global "hide helper models" toggle** (`settings.model.hide_helper_models`). When on, drafters are excluded from `/v1/models`. A model is a drafter if it's intrinsically flagged at discovery **or** referenced as another model's draft (`dflash_draft_model` / `specprefill_draft_model` / `vlm_mtp_draft_model`).
2. **Per-model visibility toggle** (`ModelSettings.is_hidden`). User-activated, any model, excludes it from `/v1/models`. Excluded from profiles/templates.

## Helper detection

Computed at **discovery time** in `is_helper_model_config(config)`, using definitive config signals rather than the often-misleading `model_type`:

- `*_assistant` / `*_mtp` `model_type` (e.g. `gemma4_assistant`, `qwen3_5_mtp`), **or**
- a drafter architecture name — `draft` / `assistant` / `mtp` token (e.g. `DFlashDraftModel`), **or**
- a drafter-only config block (`dflash_config`).

This is what catches DFlash drafts, whose `model_type` is a plain `qwen3` and are otherwise indistinguishable from a chat model. The result is stored as `is_helper` on `DiscoveredModel` / `EngineEntry` and surfaced in `get_status()`, so both `/v1/models` and the admin badge read a real flag. Verified against real checkpoints: DFlash 9B/35B and gemma assistant → helper; normal `qwen3_5` / `qwen3_5_moe` chat models → not.

## Admin UI

- Visibility toggle + "helper"/"hidden" badges in the models table (mirrors the existing pin-toggle pattern), sortable Visible column.
- Global toggle in the Model settings section.
- i18n keys added across all 8 locales.

## Testing

`pytest -m "not slow"` green. New coverage:

- Unit: `is_helper_config_model_type` / `is_helper_model_config` (architecture / `dflash_config` / model_type cases + non-helper negatives).
- End-to-end `tests/test_v1_models_visibility.py`: default listing, per-model hide, toggle-off leaves helpers visible, intrinsic-flag hide, hide-by-reference, referencing parent stays visible.
- Updated two existing stubs for the new fields (`test_audio_api` `is_hidden=False`; `test_settings` `to_dict` key).

No new dependencies; no changes to `pyproject.toml`.


jundot:
I'm closing this as superseded by #2107, which includes this visibility work plus the admin model search/sort changes.

PR #2107, opened 2026-07-06
feat(admin): model search, filtering, and persistent sort in dashboard

jasonpaulso:
## Summary

Admin-dashboard model search, filtering, and persistent sorting.

### Settings → Models (`?tab=settings&settingsTab=models`)
- **Search/filter** by model name, id, or alias
- **Persistent sort** — sort column + direction survive refreshes/restarts (localStorage), with a **Reset sort** button (disabled when already at the default `Model ↑`)
- "No models match your search" empty state

### Models → Manager (`?tab=models&modelsTab=manager`)
- **Type** and **Size** columns; the flat list is now a sortable table
  - Type is cross-referenced from `/api/models` on the client — no backend change (`/api/hf/models` already returns raw `size`)
- **Search/filter** by model name
- **Sortable column headers** matching the Settings → Models table, with independent persistence + reset
- A **gear action** deeplinks each row to its model settings card (opens the settings modal)

### Plumbing
- 10 new i18n keys across all nine locales (real translations for zh, zh-TW, ja, ko, es, fr, ru, pt-BR)
- Rebuilt `tailwind.css` so the new utility classes are included

Note: per-model alias/"friendly name" already exists in the codebase, so no changes were needed there — the new search boxes and manager badges reuse it.

## Verification
- 22-assertion Node harness over the real `dashboard()` Alpine component: search (name + alias), sort, localStorage persistence, reset-clears-storage, sort-restores-on-reload, manager type lookup, no-match — all pass
- Booted the server against test models; confirmed `/api/models` and `/api/hf/models` shapes and that the served `/admin/dashboard` renders all new markup with i18n strings resolving
- Jinja partials parse; lucide icons and Tailwind classes present

No Python touched; changes are limited to the admin JS, templates, i18n, and the built CSS.


jundot:
Thanks for the work here. I tested the browser admin flow locally, and the search/sort additions feel useful and fit the existing dashboard well. This PR also includes the model visibility work from #2106, so I'll merge this one and close #2106 as superseded.

I found two small follow-ups that I'll handle after merge: exposed profile model IDs should respect the same hidden/helper filtering in /v1/models, and pt-BR is missing the new visibility/helper i18n keys. Neither looks blocking for this PR.

LGTM. I'll merge this now.

PR #1211, opened 2026-05-12
fix(packaging): reuse Session for health checks to prevent port exhaustion

arthware-dev:
### Evidence

  Observed on a Mac running oMLX:
  - 14,126 sockets in TIME_WAIT to `127.0.0.1:8888` (the oMLX backend),
    out of 17,546 total on the host
  - `lsof` showed the menubar app (`oMLX`) and the backend (`python3`)
    connecting on a fresh source port each poll, confirming no shared
    connection
  - Outbound TCP from other apps on the same Mac failed until reboot

  ### Fix

  Create one `Session` in `__init__` and reuse it on every health
  check. A single TCP connection stays open instead of one being
  created and discarded each time.

  ### Tests

  Added `test_check_health_reuses_session` in `tests/test_omlx_app.py`
  that fails on the previous code and passes on the new code. The three
  existing `test_check_health_*` tests were updated to construct the
  manager inside the patched scope (since the `Session` is now created
  in `__init__`).

  ### Out of scope (noted for follow-up)

  The same function-local-Session pattern appears in
  `packaging/omlx_app/config.py` (two places: lines 173 and 303). Those
  call sites are not on a polling hot path, so I left them for a
  separate change to keep this PR small and focused.

  ### Note on local testing

  `uv sync --dev` fails on this repo because `paroquant==0.1.14`
  requires Python `>=3.11` but `[project].requires-python` is `>=3.10`.
  I bumped `requires-python` locally to `>=3.11` to run the tests, then
  reverted before committing. Worth a separate fix.

PR #2108, opened 2026-07-06
feat: Tencent Hy3 (hy_v3) support — mlx-lm PR #1211 pre-load patch + ':opensource' think tags

gilby:
## What

Adds serving support for **Tencent Hy3 / HunYuan V3** (`model_type: hy_v3`, 295B-A21B MoE) ahead of upstream mlx-lm merging [ml-explore/mlx-lm#1211](https://github.com/ml-explore/mlx-lm/pull/1211).

Two parts:

### 1. `patches/hy_v3` — upstream-first pre-load patch (step3p7 pattern)

Vendors PR #1211 (head `b7635e9c`, credit **@kernelpool**) and registers it at pre-load dispatch:

- `mlx_lm.models.hy_v3` — 80-layer MoE: 192 routed experts + 1 shared, sigmoid router with expert bias + scaling, qk-norm; MTP (`num_nextn_predict_layers`) weights skipped in `sanitize()`.
- `mlx_lm.tool_parsers.hy_v3` / `hy_v3_opensource` — the **released** checkpoints rename every special token with an `:opensource` suffix at the same token ids (`<tool_calls>` → `<tool_calls:opensource>`), so the release needs the suffixed sentinels while Hy3-preview uses the plain ones.
- `tokenizer_utils` wrappers: the hy_v3 template check must run **before** the `<arg_key>` → `glm47` branch (Hy3 templates contain both `<arg_key` and `<tool_sep`), and `<think:opensource>`/`</think:opensource>` is recognised as a thinking pair.

Every hook is upstream-first: once the pinned mlx-lm ships `hy_v3` natively, the patch is a no-op.

### 2. `api/thinking.py` — `:opensource` think-tag separation

`extract_thinking()` and `ThinkingParser` only knew `<think>`/`</think>` (plus the MiniMax remap), so Hy3's chain-of-thought leaked into `content` and `reasoning_content` stayed empty. This adds the Hy3 tags to the batch normalisation chain, as first-class open/close alternatives in the streaming parser, and widens `_could_be_tag`'s partial-tag buffering window.

## Testing

Running in production on a Mac Studio M3 Ultra (512 GB) against two oQ4e/mxfp4 quants of the 598 GB BF16 release checkpoint:

- Model loads and serves via the batched engine (~25 tok/s single-stream, TPOT ~41 ms, 262K ctx).
- Tool calling returns structured `tool_calls` (verified single-shot and through a LiteLLM gateway); template inference correctly picks `hy_v3_opensource`.
- With `chat_template_kwargs: {"reasoning_effort": "high"}` (the template's thinking gate), CoT lands in `reasoning_content`, answers stay clean in `content` — streaming and batch paths both verified.
- Graded A/A- in 11/12 categories of an internal role-mapped benchmark suite after these fixes (up from wide-spread leakage-capped grades before).

Notes for reviewers: the Hy3 chat template pre-opens `<think:opensource>` in the generation prompt (handled by the existing `prompt_opens_thinking` machinery once the tokenizer attrs are inferred), and with `tools` present the template intentionally omits the reasoning-mode token.

### DIFF
--- a/omlx/api/thinking.py
+++ b/omlx/api/thinking.py
@@ -20,6 +20,8 @@ _OPEN_LEN = len(_OPEN_TAG)   # 7
 _CLOSE_LEN = len(_CLOSE_TAG)  # 8
 _MINIMAX_OPEN_TAG = "<mm:think>"
 _MINIMAX_CLOSE_TAG = "</mm:think>"
+_HY3_OPEN_TAG = "<think:opensource>"
+_HY3_CLOSE_TAG = "</think:opensource>"
 
 # Regex for non-streaming extraction (complete text)
 _THINKING_PATTERN = re.compile(r'<think>(.*?)</think>', re.DOTALL)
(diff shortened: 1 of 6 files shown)


jundot:
Thanks for the Hy3 patch. I checked the vendored files against mlx-lm PR #1211 and verified the pre-load registration, :opensource thinking split, and native tool-call path with local fixtures; CI is green. LGTM after those checks.

PR #2111, opened 2026-07-07
Fix Homebrew formula installation on macOS 27 beta

### LINKED ISSUE #2110: Compilation on macOS 27b3 (and b2?)
Hi,
just got back from holiday and wasn't able to compile omlx anymore on macOS 27b3 (and I believe already on b2). There was more than one issue, the one I already hit on b2 before the holiday was https://github.com/llvm/llvm-project/issues/203678.

After a few iterations, I got gemini to produce working formula, so apologies for possible slop and alhough I tried to keep it on rails, I'm not a dev, just a sysadmin and I can't vouch for it being any good outside of "WorksForMe" or whether it also hacked around some upstream homebrew bug.

Therefore, I'm not submitting a PR for it (let me know if you want me to).
Hopefully it will at least be useful for someone running macOS 27 betas in the meantime.

Formula is here:

[omlx.rb.txt](https://github.com/user-attachments/files/29739531/omlx.rb.txt)

Work for me with "brew install omlx --with-grammar --HEAD --with-custom-kernel" and d42528afd426f50ee11fbba5243ab004010a8b52 commit.

**I did NOT test it in macOS <27**

----

### Fix Homebrew Formula Installation under macOS 27 Beta (Dynamic Linker strictness, CMake Python contamination, and Directory Shadowing)

**Context & Problem Description**

When installing the omlx formula under the macOS 27 Beta, multiple issues arise causing the installation to fail or crash post-install:

Stricter dyld Alignment in macOS 27+: The dynamic linker (dyld) now strictly enforces that the symbol-table string pool (LC_SYMTAB) inside compiled Mach-O dynamic binary libraries (.so, .dylib) must begin on an 8-byte boundary.
Pre-compiled wheels on PyPI for third-party Rust-compiled dependencies (e.g. pydantic-core, rpds-py, tiktoken, and tokenizers) were compiled with older cross-compilers aligning to 4-byte boundaries, causing dlopen() to fail instantly with:

ImportError: dlopen(.../_pydantic_core.cpython-311-darwin.so, 0x0002): 
tried: '...' (mis-aligned LINKEDIT string pool, fileOffset=0x00420F14)

**Apple Beta strip Toolchain Bug:**
Additionally, when compiling these dependencies from source, both Cargo/Maturin's release stripping and Homebrew's automatic post-install clean pass invoke macOS's local strip binary. The current macOS 27 beta strip binary contains an upstream bug that misaligns dynamic offsets, corrupting the compiled dynamic libraries.
https://github.com/llvm/llvm-project/issues/203678

**CMake Python Version Contamination:**
When compiling custom C++ or Metal extensions, CMake uses its default scanning parameters to look for Python, sometimes finding a newer unlinked system Python version (e.g. Python 3.14) instead of the Homebrew python@3.11 virtual environment target.

**Directory Shadowing during Tests:**
The verification script runs from the temporary build directory, causing Python to shadow the local raw folder instead of importing from the fully compiled virtualenv site-packages.

### Solution

We have successfully resolved these errors while maintaining full backward-compatibility with older macOS releases (14, 15, 16) by implementing targeted, conditional workarounds:

**Stricter dynamic linker fix (macOS 27+):**
If on macOS 27+, we append --no-binary to enforce native source compilation for problematic binary dependencies (pydantic-core, rpds-py, tiktoken, tokenizers, cohere-melody).

**Binary strip fix (macOS 27+):**
If on macOS 27+, we conditionally turn off Cargo/Maturin debug-stripping (CARGO_PROFILE_RELEASE_STRIP=false, MATURIN_STRIP=false) and disable Homebrew's post-install clean pass (skip_clean "libexec"). This guarantees they are not corrupted by the broken beta strip utility.

**CMake Python Contamination Fix:**
We explicitly define Python_EXECUTABLE and CMAKE_ARGS to keep CMake locked into the virtual environment's Python 3.11 binaries.

**Test Isolation:**
Wrapped validation assertions in a clean Dir.chdir libexec block.

heyparth1:
This addresses multiple compilation and runtime failures on macOS 27 beta (#2110):

- **dyld alignment**: macOS 27's dynamic linker requires the LC_SYMTAB string pool to be 8-byte aligned. Prebuilt Rust wheels (e.g. `tokenizers`) aligned to 4 bytes fail `dlopen`, so they are now built from source on macOS 27+.
- **Broken `strip`**: The beta's `strip` corrupts dynamic offsets in Mach-O libraries (llvm/llvm-project#203678). Cargo/maturin release stripping is disabled and Homebrew's post-install clean pass is skipped (`skip_clean "libexec"`) on macOS 27+.
- **CMake Python discovery**: CMake's default Python discovery can pick a stray system Python instead of the formula's venv. `CMAKE_ARGS` now pins `Python_EXECUTABLE` to the venv's Python for custom kernel builds.
- **Directory shadowing**: The custom-kernel verification step ran from the build directory, where the raw `omlx/` source tree shadowed the installed package. It now runs from `libexec`.

Regression tests are added in `tests/test_homebrew_formula.py` to assert these guards remain in the formula.

Closes #2110

### DIFF
--- a/Formula/omlx.rb
+++ b/Formula/omlx.rb
@@ -16,6 +16,11 @@ class Omlx < Formula
   depends_on :macos
   depends_on "python@3.11"
 
+  # macOS 27 beta's `strip` corrupts dynamic offsets in Mach-O libraries
+  # (llvm/llvm-project#203678). Skip Homebrew's post-install clean pass over
+  # the venv so it never runs `strip` on the compiled dylibs.
+  skip_clean "libexec" if MacOS.version >= "27"
+
   # mlx-audio pins mlx-lm==0.31.1 which conflicts with omlx's git-pinned
   # mlx-lm. Fetch source separately so we can patch the pin before install.
   resource "mlx-audio" do
@@ -51,6 +56,19 @@ class Omlx < Formula
     # C/C++ extension builds use LDFLAGS.
     ENV.append "LDFLAGS", "-Wl,-headerpad_max_install_names"
     ENV.append "RUSTFLAGS", "-C link-arg=-Wl,-headerpad_max_install_names"
+
+    no_binary = "cohere_melody,pydantic-core,rpds-py,tiktoken"
+    if MacOS.version >= "27"
+      # macOS 27's dyld requires the LC_SYMTAB string pool to start on an
+      # 8-byte boundary; prebuilt Rust wheels aligned to 4 bytes fail dlopen
+      # with "mis-aligned LINKEDIT string pool". Build them from source, and
+      # keep Cargo/maturin's release stripping off so the beta's broken
+      # `strip` (llvm/llvm-project#203678) never touches the fresh dylibs.
+      no_binary += ",tokenizers"
+      ENV["CARGO_PROFILE_RELEASE_STRIP"] = "false"
+      ENV["MATURIN_STRIP"] = "false"
+    end
+
     if build.with?("custom-kernel")
       kernel_sources = [
         buildpath/"omlx/custom_kernels/glm_moe_dsa/csrc",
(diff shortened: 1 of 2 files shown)


heyparth1:
Nothing new to address this round — no review comments or failing checks came through. I re-verified the branch: the formula still parses cleanly (`ruby -c`) and the regression tests in `tests/test_homebrew_formula.py` pass. Happy to make changes if anything comes up in review.

jundot:
Thanks for tracking this down and putting together the macOS 27 beta install fix.

The formula changes are focused on the Homebrew install path, and the added regression tests pass locally. There is one small Homebrew style issue around the top-level `MacOS.version` guard, but I can fold that into a follow-up cleanup on main after merging.

LGTM. I will merge this and apply the small formula guard cleanup separately.

PR #2117, opened 2026-07-07
fix(server): include role in the chunk-mode chat keepalive delta

### LINKED ISSUE #2074: Chunk-form SSE keepalive omits `role` — LangChain.js/n8n silently drop streamed tool_calls
### Summary

With the default `sse_keepalive_mode: "chunk"`, the keepalive frame is the **first SSE event of every stream**, and its delta has no `role`. LangChain.js types a stream from its first chunk: no `role` → generic `ChatMessageChunk` instead of `AIMessageChunk`, and when the real chunks merge into it, all `tool_call_chunks` are discarded. Streamed tool calls are lost silently — text still streams fine.

Real-world hit: **n8n AI Agent workflows** (n8n bundles LangChain.js). With a chat trigger in streaming mode, every tool call from an oMLX model is dropped — the agent answers as if the model never called a tool, so it looks like the model refuses to use tools. Non-streaming requests work correctly.

This is the same failure class as #1478 (openai-go dropping `tool_calls` over the keepalive's mismatched `id`): one more strict-accumulator invariant the no-op chunk needs to satisfy — cloud APIs open every stream with the `role` chunk, and clients rely on that.

### Wire transcript (chunk mode)

```
data: {"id":"chatcmpl-…","model":"keepalive","choices":[{"index":0,"delta":{"content":""},"finish_reason":null}]}   ← first frame, no role
data: {"id":"chatcmpl-…","choices":[{"index":0,"delta":{"role":"assistant"}}]}
…
data: {"id":"chatcmpl-…","choices":[{"index":0,"delta":{"tool_calls":[{"index":0,"id":"call_…","type":"function","function":{"name":"get_weather","arguments":"{\"city\": \"Paris\"}"}}]}}]}
data: {"id":"chatcmpl-…","choices":[{"index":0,"delta":{},"finish_reason":"tool_calls"}]}
```

### Repro

```js
// npm i @langchain/openai   (1.4.4, same as bundled in n8n 2.28.4)
const { ChatOpenAI } = require('@langchain/openai');
(async () => {
  const model = new ChatOpenAI({ model: '<any tool-capable model>', apiKey: '<key>',
    configuration: { baseURL: 'http://localhost:2000/v1' } });
  const bound = model.bindTools([{ type: 'function', function: {
    name: 'get_weather', description: 'Weather for a city',
    parameters: { type: 'object', properties: { city: { type: 'string' } }, required: ['city'] } } }]);
  let agg = null;
  for await (const c of await bound.stream('Weather in Paris? Use get_weather.')) agg = agg ? agg.concat(c) : c;
  console.log(agg.constructor.name, JSON.stringify(agg.tool_calls));
  // chunk mode:   ChatMessageChunk undefined      ← tool call lost
  // comment mode: AIMessageChunk [{"name":"get_weather","args":{"city":"Paris"}, …}]
})();
```

### Suggested fix

Either option keeps the rationale from #1011 intact (frames stay valid `data:` events for clients that can't parse SSE comments):

1. Include the role in the keepalive delta — `"delta":{"role":"assistant","content":""}` in `_KEEPALIVE_CHAT_CHUNK` / `_chat_keepalive_chunk()`. Duplicate `role` fields are tolerated by the OpenAI SDKs, openai-go, and LangChain.
2. Or emit the `delta:{"role":"assistant"}` chunk immediately at stream open (before prefill), like the cloud APIs do — any later keepalive is then typed correctly.

### Workaround

`sse_keepalive_mode: "comment"` fixes it (comments never reach the JSON layer).

### Environment

oMLX 0.4.4 (260616183743-macos26-tahoe), macOS, default settings. Reproduced with Qwen3.5-122B-A10B-4bit and Qwen3.6-35B-A3B (parser `qwen3_coder`). Client: `@langchain/openai` 1.4.1 and 1.4.4 / n8n 2.28.4.

richgoodson:
Fixes #2074.

## What happens

With the default `sse_keepalive_mode: "chunk"`, the keepalive frame is the first SSE event of every chat stream and its delta has no `role`. LangChain.js types a stream from its first chunk: no role means it builds a generic `ChatMessageChunk`, and when the real chunks merge into it every `tool_call_chunk` is discarded. Streamed tool calls are lost silently while text keeps streaming, so every n8n AI Agent workflow (n8n bundles LangChain.js) behaves as if the model refused to call its tools.

This is the same strict-accumulator class as #1478 (openai-go dropping chunks over the keepalive's mismatched id). Cloud APIs open every stream with a role chunk, and clients rely on that.

## The fix

Add `"role":"assistant"` to the keepalive delta in both chat frames: the static sentinel frame (`_KEEPALIVE_CHAT_CHUNK`) and the id-sharing frame from #1478 (`_chat_keepalive_chunk`). The reporter verified that the OpenAI SDKs, openai-go, and LangChain all tolerate the duplicate role in later chunks, and my live test below confirms it for LangChain. The completions keepalive has no role concept and the Anthropic ping is a typed event, so chat is the whole surface.

## Tests

Extended the two existing frame-format tests and added `TestChatKeepaliveCarriesRole` (2 tests) asserting the first-frame role invariant on both frames. Full suite: 6116 passed; one pre-existing failure in `test_mlx_vlm_minimax_m3_compat` that also fails on clean `d42528af` in my env (stale local mlx_vlm pin, unrelated).

## Live verification

Ran the reporter's literal LangChain.js repro (`@langchain/openai` 1.4.4) against a served Qwen3.5-9B-MLX-4bit, before and after the fix on the same machine:

```
unpatched:  ChatMessageChunk undefined                      <- tool call lost
patched:    AIMessageChunk [{"name":"get_weather","args":{"city":"Paris"},"id":"call_b8e00d17","type":"tool_call"}]
```

Wire transcript after the fix (first two frames):

```
data: {"id":"chatcmpl-3f62e3f5",...,"model":"keepalive","choices":[{"index":0,"delta":{"role":"assistant","content":""},"finish_reason":null}]}
data: {"id":"chatcmpl-3f62e3f5",...,"model":"Qwen3.5-9B-MLX-4bit","choices":[{"index":0,"delta":{"role":"assistant"}}]}
```

The `sse_keepalive_mode: "comment"` workaround from the issue keeps working; comment mode is untouched.

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -1857,10 +1857,18 @@ def init_server(
 _KEEPALIVE_SENTINEL = object()
 
 _KEEPALIVE_COMMENT = ": keep-alive\n\n"
+# The delta carries "role":"assistant" because this frame is the FIRST event
+# of every stream and some accumulators type the whole stream from the first
+# chunk's role: LangChain.js builds a generic ChatMessageChunk when it is
+# absent, and merging the real chunks into it silently discards all
+# tool_call_chunks (#2074, hits n8n AI Agent workflows). Cloud APIs open every
+# stream with a role chunk, so clients rely on it; the OpenAI SDKs, openai-go,
+# and LangChain all tolerate the duplicate role in later chunks.
 _KEEPALIVE_CHAT_CHUNK = (
     'data: {"id":"chatcmpl-keepalive","object":"chat.completion.chunk",'
     '"created":0,"model":"keepalive",'
-    '"choices":[{"index":0,"delta":{"content":""},"finish_reason":null}]}\n\n'
+    '"choices":[{"index":0,"delta":{"role":"assistant","content":""},'
+    '"finish_reason":null}]}\n\n'
 )
 _KEEPALIVE_COMPLETION_CHUNK = (
     'data: {"id":"cmpl-keepalive","object":"text_completion","created":0,'
@@ -1909,11 +1917,16 @@ def _chat_keepalive_chunk(response_id: str) -> str:
     keepalive with the stream's own ``response_id`` makes it a true no-op for
     those clients while remaining a parseable data event for clients that can't
     handle SSE comment lines.
+
+    The delta must also carry ``"role":"assistant"`` — see the comment on
+    ``_KEEPALIVE_CHAT_CHUNK`` (accumulators that type the stream from the
+    first chunk's role drop tool_call_chunks without it, #2074).
     """
     return (
         'data: {"id":"' + response_id + '","object":"chat.completion.chunk",'
         '"created":0,"model":"keepalive",'
-        '"choices":[{"index":0,"delta":{"content":""},"finish_reason":null}]}\n\n'
+        '"choices":[{"index":0,"delta":{"role":"assistant","content":""},'
+        '"finish_reason":null}]}\n\n'
     )
 
 
--- a/tests/test_sse_keepalive.py
+++ b/tests/test_sse_keepalive.py
@@ -111,6 +111,7 @@ class TestKeepaliveChunkFormats:
         body = items[0].removeprefix("data: ").strip()
         payload = json.loads(body)
         assert payload["object"] == "chat.completion.chunk"
+        assert payload["choices"][0]["delta"]["role"] == "assistant"
         assert payload["choices"][0]["delta"]["content"] == ""
         assert payload["choices"][0]["finish_reason"] is None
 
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for the focused fix. I checked the chat streaming path and the keepalive is still the first data event in chunk mode, so carrying role: assistant on both chat keepalive frames matches the LangChain.js accumulator invariant without changing completion, Anthropic, or comment-mode behavior.

The focused keepalive tests pass on my side. LGTM.

PR #2125, opened 2026-07-08
Fix Models load button wrapping

### LINKED ISSUE #2123: macOS app: Models screen Load button wraps vertically for long model names
### Description

In the native macOS app, the Models > Model Library row can compress the `Load` button until the label wraps vertically, showing one character per line.

Load button wrapping vertically

### Steps to reproduce

1. Open the native macOS app.
2. Go to Models.
3. Have several downloaded models, including at least one with a long repository/model name.
4. Use a window/list width where the model name area is under horizontal pressure.

### Expected behavior

The trailing actions keep stable widths and the `Load` / `Unload` button remains a normal single-line button. Long model names should truncate before the action controls compress.

### Actual behavior

The `Load` button is compressed and wraps vertically (`L o a d`), making the row look broken and harder to use.

### Notes

I checked latest `main` at `1d499c2` (2026-07-08). The issue still appears possible from the current SwiftUI layout:

- `apps/omlx-mac/Sources/AppView/Screens/ModelsScreen.swift`: the library row places long model text and trailing controls in one `HStack`.
- `apps/omlx-mac/Sources/Theme/Components/Button.swift`: the custom button style does not reserve a fixed/min width or prevent text wrapping for the small text button.

A likely fix would be to reserve a stable trailing action area and/or make the `Load` / `Unload` label fixed-size horizontally, so the model title/subtitle truncate before the controls compress.

alan13367:
## Summary

- Keeps the Models library row action controls at their natural width.
- Makes the `Load` / `Unload` action label single-line with a small minimum width.
- Lets long model names truncate before the trailing controls compress.

## Root Cause

The model title stack had higher layout priority than the trailing controls, so long repository/model names could win horizontal space and force the custom `Load` button to wrap vertically.

Fixes #2123.

## Validation

- `xcodebuild test -project apps/omlx-mac/oMLX.xcodeproj -scheme oMLX -destination 'platform=macOS'`

### DIFF
--- a/apps/omlx-mac/Sources/AppView/Screens/ModelsScreen.swift
+++ b/apps/omlx-mac/Sources/AppView/Screens/ModelsScreen.swift
@@ -229,47 +229,62 @@ private struct LibrarySection: View {
                                     .lineLimit(1)
                                     .truncationMode(.middle)
                             }
-                            .layoutPriority(1)
                             Spacer(minLength: 8)
-                            if isModelLoaded(m.id) {
-                                Button(String(localized: "models.library.unload",
-                                              defaultValue: "Unload",
-                                              comment: "Button label that unloads a library model from memory")) { onUnload(m.id) }
+                            HStack(spacing: 10) {
+                                if isModelLoaded(m.id) {
+                                    Button {
+                                        onUnload(m.id)
+                                    } label: {
+                                        Text(String(localized: "models.library.unload",
+                                                    defaultValue: "Unload",
+                                                    comment: "Button label that unloads a library model from memory"))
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for reopening this against the updated base. I verified the diff against current main and the macOS test suite passes locally.

Once your final review is complete, please mark this ready for review. I'll merge it after the draft status is cleared.

jundot:
Thanks for the fix. This matches #2123: the row title no longer gets higher layout priority than the trailing controls, and the Load/Unload label now keeps a stable single-line width.

PR #2134, opened 2026-07-09
fix(model_discovery): discover DFlash draft checkpoints from HF cache

### LINKED ISSUE #1643: Since latest update not able to select z-lab DFLASH models for Qwen3.6 model families
**Describe the bug**
Tittle says it all I see the z-lab DFLASH models as downloaded but I cannot select them anymore 
**To Reproduce**
download z-lab--Qwen3.6-27B-DFlash and mlx-community--Qwen3.6-27B-4bit
go to model settings, activate DFlash the z-lab--Qwen3.6-27B-DFlash is not in the selection list all other downloaded models are (should only surface draft models the list is long with none suitable models)

**Expected behavior**

Used to work in 0.3 branch, seems to have been broken in 0.4

**Screenshots**

 - macOS Version: [e.g. 26.5]
 - oMLX Version [e.g. 0.4.1]

**Additional context**

JimStenstrom:
Fixes #1643

## What happens

`_is_hf_cache_mlx_compatible` (new in 0.4) admits an HF-cache entry only when its safetensors declare `format=mlx` or the repo name carries an `mlx` token. DFlash draft checkpoints (e.g. `z-lab/Qwen3.6-27B-DFlash`) are MLX-loadable but ship pt-format safetensors under a repo name with no `mlx` token, so 0.4 silently drops them from HF-cache discovery — and they vanish from the DFlash draft-model picker.

That matches the report exactly: "I see the z-lab DFLASH models as downloaded but I cannot select them anymore", "Used to work in 0.3 branch" (0.3 had no such gate), and the commenter's workaround "redownload the DFLASH model … then it appears" (a fresh download can land outside the HF-cache layout, bypassing the gate).

## The fix

Recognise speculative-decoding helper checkpoints in the gate by delegating to the existing `is_helper_model_config`, so the drafter markers (`DFlashDraftModel` architecture, `dflash_config` block, `*_mtp` / `*_assistant` model_type) live in one place, in sync with `is_helper` flagging and engine_pool's drafter resolution — per the "keep these in sync" note above `HELPER_CONFIG_MODEL_TYPE_SUFFIXES`. Admitted drafts flow through the shared `DiscoveredModel` construction path, which flags them `is_helper=True`, so they surface in the draft picker (and can be hidden from the chat-model list via `hide_helper_models` — the reporter's "should only surface draft models" wish).

The new wrapper never raises on an unreadable entry: it runs outside `discover_models`' per-model guard, so an escaped exception would abort the whole scan. `UnicodeDecodeError` from a non-UTF-8 config.json is a `ValueError`, not a `JSONDecodeError`, hence the explicit exception tuple.

## Why this is safe

- A plain non-MLX PyTorch model still fails the gate — `is_helper_model_config` requires an explicit drafter marker.
- For helper checkpoints this is parity, not a new surface: the flat `model_dir` path has no MLX-compat gate, so the same checkpoint placed there is already discovered today. The change only stops the HF-cache layout from behaving differently.
- Trade-off, named: delegation is broader than a DFlash-only check — a pt-format `*_mtp` / `*_assistant` checkpoint in the HF cache becomes discoverable too (flagged `is_helper`, same treatment it already gets in `model_dir`). If you'd rather admit only DFlash drafts from the HF cache, the predicate can trivially narrow to the architecture/config-block markers — happy to adjust.

## Verification

- Verified against three real DFlash draft checkpoints on disk (z-lab Qwen3-Coder-Next and Qwen3.5-122B-A10B drafts, plus a Qwen3.6-27B FP16 draft): all carry `DFlashDraftModel` + `dflash_config` and pass the predicate; a config-less partial download correctly fails it.
- `tests/test_model_discovery.py`: 155 passed. Two new tests: `test_hf_cache_dflash_draft_is_discovered` (drives the real `discover_models` path on a pt-format cache entry and asserts the `is_helper` flag) and `test_is_helper_checkpoint_on_disk_edge_cases` (missing / malformed / non-UTF-8 config.json never raise). Marker-family coverage stays in the existing `TestIsHelperModelConfig`.
- Revert-proof: with the fix reverted, `test_hf_cache_dflash_draft_is_discovered` fails with the exact field symptom — the draft is absent from discovery.

Siblings checked: the flat `model_dir` path needs no change (no gate there). The `model*.safetensors` precondition still runs before the helper check, so a draft with unconventionally-named weights would still be skipped — named as out of scope (real DFlash drafts ship `model.safetensors`).

### DIFF
--- a/omlx/model_discovery.py
+++ b/omlx/model_discovery.py
@@ -1095,6 +1095,31 @@ def _safetensors_has_mlx_metadata(path: Path) -> bool:
 
 _MLX_NAME_RE = re.compile(r"(^|[-_/])mlx($|[-_/])", re.IGNORECASE)
 
+# Speculative-decoding helper checkpoints (e.g. z-lab/Qwen3.6-27B-DFlash) are
+# MLX-loadable drafts even though their safetensors are saved in PyTorch ("pt")
+# format and the repo name carries no "mlx" token, so the HF-cache MLX
+# heuristic must recognise them explicitly or they vanish from the draft-model
+# picker (#1643). Detection delegates to is_helper_model_config so the drafter
+# markers stay in sync with helper flagging and engine_pool's drafter
+# resolution.
+def _is_helper_checkpoint(model_path: Path) -> bool:
+    """True if ``model_path`` is a speculative-decoding helper checkpoint.
+
+    Must never raise on an unreadable entry: unlike the config reads inside
+    _register_model, this runs outside discover_models' per-model guard, so
+    an escaped exception would abort the whole scan. UnicodeDecodeError from
+    a non-UTF-8 config.json is a ValueError, not a JSONDecodeError.
+    """
+    config_path = model_path / "config.json"
+    if not config_path.exists():
+        return False
+    try:
+        with open(config_path) as f:
+            config = json.load(f)
+    except (json.JSONDecodeError, UnicodeDecodeError, OSError):
+        return False
+    return is_helper_model_config(config)
+
 
 def _is_hf_cache_mlx_compatible(model_dir: Path, source_repo_id: str) -> bool:
     """Heuristic for HF cache entries that can be loaded without conversion."""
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the focused fix. This matches the HF-cache DFlash regression in #1643: the draft checkpoints are MLX-loadable, but the current cache gate drops them because they have pt-format safetensors and no mlx token in the repo name.

I found one small edge case while reviewing: _is_helper_checkpoint() can still raise if config.json is valid JSON but has an unexpected schema, e.g. `"architectures": 123`, because this runs outside discover_models' per-model guard. That is narrow enough that I don't want to block this fix on it.

LGTM for the reported issue. I'll merge this and fold the schema-hardening case into a follow-up.

PR #2142, opened 2026-07-09
fix(server): drop malformed parser tool calls

### LINKED ISSUE #2012: malformed Harmony tool-call arguments can abort SSE streaming with server_error
## Summary

`omlx` can terminate Anthropic/OpenAI SSE streaming with a server error when a model emits a malformed tool-call `arguments` string. Instead of dropping or sanitizing the malformed tool call, the stream path constructs `FunctionCall(arguments=...)`, triggers strict JSON validation, and aborts the entire SSE response.

## Environment

- oMLX version: 0.4.4 (Homebrew)
- Install path: `/opt/homebrew/Cellar/omlx/0.4.4`
- Python: bundled 3.11
- OS: macOS (Apple Silicon)
- Observed during SSE chat streaming
- Likely parser path: Harmony / function-recipient tool calls (`functions.*`)

## Reproduction

1. Run `omlx` 0.4.4 with a model/parser path that emits tool calls as raw text arguments.
2. Let the model produce a tool call with malformed JSON in `arguments`.
3. Observe the SSE stream fail with a server error instead of continuing or dropping the invalid tool call.

Representative log:

```text
2026-06-27 12:17:51,281 - omlx.server - ERROR - [-] - SSE generator error: 1 validation error for FunctionCall
arguments
  Value error, arguments must be valid JSON, got parse error: Expecting ':' delimiter: line 1 column 213 (char 212). This usually means the client echoed a previous tool call with a malformed arguments value. Send arguments as a JSON-encoded object string like '{"location": "Tokyo"}'. Received: '{"command": "cd /path/to/project && ls -la target..." ... "timeout": "30000"}}\n]' [type=value_error, input_value='{"command": "cd /path/to/project ... "timeout": "30000"}}\n]', input_type=str]
    For further information visit https://errors.pydantic.dev/2.13/v/value_error
```

The payload tail shown in the log (`..."timeout": "30000"}}\n]`) suggests malformed model-emitted JSON: extra closing bracket / broken quoting / truncated or duplicated suffix.

## Expected Behavior

Malformed model-emitted tool-call arguments should not crash the SSE stream. `omlx` should either:

- drop the bad tool call with a warning, or
- preserve the text response and omit the invalid tool block, or
- emit a clean recoverable parse warning/event instead of a server error.

## Actual Behavior

The SSE stream aborts with:

- `SSE generator error: 1 validation error for FunctionCall`
- `arguments must be valid JSON`

This converts a malformed tool-call fragment into a full stream failure.

## Evidence

- `omlx/api/openai_models.py:132-188`
  - `_coerce_tool_call_arguments()` strictly requires `arguments` to be a JSON-object string.
  - `FunctionCall.arguments` validator raises `ValueError` on malformed JSON.
- `omlx/server.py:200-217`
  - `_convert_parser_tool_calls()` blindly constructs `FunctionCall(name=..., arguments=tool_call.get("arguments", "{}"))`.
  - No local recovery if validation fails.
- `omlx/server.py:1897-1907`
  - Any exception from SSE generation is logged as `SSE generator error` and returned as `server_error`, ending the stream.
- `omlx/adapter/harmony.py:327-351`
  - `HarmonyOutputParserSession.get_tool_calls()` concatenates raw tool-call content into `arguments` with no JSON validation or repair.
- `omlx/adapter/harmony.py:473-481`
  - `parse_tool_calls_from_tokens()` also accumulates raw function-recipient text into `arguments` and returns it directly.

Taken together, these paths mean malformed model output can propagate from the parser straight into `FunctionCall(...)`, where validation happens too late and too hard for a streaming code path.

## Root Cause

Tool-call parsers in the Harmony/function-recipient path return raw `arguments` strings without ensuring they are valid JSON objects. Later, the SSE/server path treats those raw strings as already-valid tool arguments and passes them into `FunctionCall`, whose validator raises on malformed JSON. That exception is not converted into a non-fatal tool-parse failure, so the whole SSE response is terminated.

This does not look like a client-echo problem in this case. The logged payload shape is consistent with malformed model-emitted tool-call JSON being forwarded unchecked.

## Proposed Fix

Handle invalid parser-emitted tool-call arguments as a recoverable parse failure before entering the SSE response path.

Concrete directions:

1. In `omlx/server.py:_convert_parser_tool_calls()`:
   - wrap `FunctionCall(...)` construction in `try/except ValueError`
   - if validation fails, log a warning with tool name + short arguments snippet
   - skip that tool call instead of aborting the stream

2. Optionally harden `omlx/adapter/harmony.py`:
   - attempt `json.loads(arguments)` when extracting `functions.*` tool calls
   - only return tool calls whose arguments parse to a JSON object
   - otherwise treat malformed function-recipient output as plain assistant text or drop with warning

3. Add regression coverage for:
   - malformed tool-call arguments from Harmony/function-recipient parser
   - valid text output + invalid tool-call output in same response
   - SSE streaming should continue and complete without `server_error`

Minimal acceptable fix would be step 1 alone: invalid tool calls should be skipped, not crash the stream.

## Verification Plan

- Reproduce with a malformed function-recipient tool call whose `arguments` string is not valid JSON.
- Confirm SSE stream no longer aborts with `server_error`.
- Confirm invalid tool call is dropped or downgraded to warning.
- Confirm valid tool calls still stream correctly.
- Confirm pure text responses are unaffected.

---
This issue or PR was generated by LazyCodex.
Tag: lazycodex-generated

e345ee:
Fixes #2012.

## Summary

Malformed tool-call arguments emitted by protocol parsers could propagate into `FunctionCall(...)` validation and abort non-streaming or SSE response handling. This moves parser tool-call conversion into a small shared helper and treats invalid parser-emitted arguments as recoverable parser failures: the malformed call is dropped with a warning while valid calls continue through unchanged.

## What changed

- Added `convert_parser_tool_calls()` for validated parser-emitted tool-call conversion.
- Routed all parser-provided tool-call paths in `server.py` through the helper, including OpenAI chat, Anthropic messages, and Responses API streaming/non-streaming paths.
- Added regression coverage for dropping malformed arguments while preserving valid tool calls in the same parser result.

## Validation

- `docker run --rm -v "$PWD":/repo -w /repo python:3.12-slim sh -lc 'python -m pip install -q pytest pydantic regex jsonschema fastapi && python -m pytest tests/test_parser_tool_calls.py -q'`
- `docker run --rm -v "$PWD":/repo -w /repo python:3.12-slim sh -lc 'python -m pip install -q black && python -m compileall -q omlx/server.py omlx/api/parser_tool_calls.py tests/test_parser_tool_calls.py && python -m black --check omlx/server.py omlx/api/parser_tool_calls.py tests/test_parser_tool_calls.py'`

### DIFF
--- a/omlx/api/parser_tool_calls.py
+++ b/omlx/api/parser_tool_calls.py
@@ -1,0 +1,51 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Helpers for converting parser-emitted tool calls to OpenAI models."""
+
+import logging
+import uuid
+
+from pydantic import ValidationError
+
+from .openai_models import FunctionCall, ToolCall
+
+logger = logging.getLogger(__name__)
+
+
+def convert_parser_tool_calls(tool_calls: list[dict] | None) -> list[ToolCall]:
+    """Convert parser-emitted tool-call dicts into validated OpenAI ToolCalls.
+
+    Parser output comes from model text and can contain malformed JSON
+    arguments. Treat those as recoverable parser failures rather than letting
+    Pydantic validation abort the response stream.
+    """
+    converted: list[ToolCall] = []
+    for tool_call in tool_calls or []:
+        if not isinstance(tool_call, dict):
+            continue
+        name = tool_call.get("name", "")
+        arguments = tool_call.get("arguments", "{}") or "{}"
+        try:
+            converted.append(
+                ToolCall(
+                    id=tool_call.get("id")
+                    or tool_call.get("call_id")
+                    or f"call_{uuid.uuid4().hex[:8]}",
+                    type="function",
+                    function=FunctionCall(
+                        name=name,
+                        arguments=arguments,
+                    ),
+                )
+            )
+        except (TypeError, ValueError, ValidationError) as e:
+            snippet = str(arguments)
+            if len(snippet) > 120:
+                snippet = snippet[:117] + "..."
+            logger.warning(
+                "Dropping malformed parser tool call %r: %s. arguments=%r",
+                name,
+                e,
+                snippet,
+            )
+            continue
+    return converted
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -121,12 +121,13 @@ from .api.openai_models import (
     CompletionChoice,
     CompletionRequest,
     CompletionResponse,
-    FunctionCall,
     ModelInfo,
     ModelsResponse,
     PromptTokensDetails,
-    ToolCall,
     Usage,
+)
+from .api.parser_tool_calls import (
+    convert_parser_tool_calls as _convert_parser_tool_calls,
 )
 from .api.rerank_models import (
     RerankRequest,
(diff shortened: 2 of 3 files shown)


jundot:
Thanks for the fix, this looks solid. The real culprit is the missing try/except around the `FunctionCall` construction, so treating pydantic's `ValidationError` as a recoverable drop is exactly right and matches the SSE crash log in the issue. I'll merge this.

e345ee:
wow that's was fast thanks

PR #2143, opened 2026-07-09
Require commentary channel for Harmony tool calls

### LINKED ISSUE #2032: Harmony tool-call extraction should require commentary channel
## Summary

Harmony tool-call extraction can misclassify non-commentary messages as tool calls when a parsed message retains a `functions.*` recipient outside the `commentary` channel.

When that happens, arbitrary `analysis` or `final` text is concatenated into tool-call `arguments`, producing giant malformed payloads. Downstream validation then logs warnings like `Dropping malformed parser tool call 'read'`, but the real bug is earlier: tool extraction is using recipient alone instead of channel + recipient.

This appears to be distinct from #2012, which fixes the downstream SSE failure mode by dropping malformed parser-emitted tool calls instead of aborting the stream. This issue tracks the upstream misclassification that generates those malformed tool-call payloads in the first place.

## Environment

- oMLX version: 0.4.4
- Install method: Homebrew
- Python: bundled 3.11
- OS: macOS (Apple Silicon)
- Model/parser family: Harmony / `functions.*` recipient path

## Symptoms

- thought/reasoning pane can show large garbage blobs mixed with code, JSON fragments, multilingual text, and control characters
- server logs warnings such as:

```text
Dropping malformed parser tool call 'read': arguments must be valid JSON ...
```

- malformed tool-call payloads can contain obvious non-argument text from other channels

## Root Cause

`omlx/adapter/harmony.py` extracts tool calls from any parsed message whose recipient starts with `functions.`.

Two paths currently do this:

- `HarmonyStreamingParser.get_tool_calls()`
- `parse_tool_calls_from_tokens()`

In both places, extraction is gated on recipient only, not on `msg.channel == "commentary"`.

That means if the Harmony parser preserves a `functions.*` recipient on a non-commentary message, `analysis` or `final` content can be misclassified as tool-call `arguments`.

## Evidence

Current logic effectively does this:

```python
if msg.recipient and msg.recipient.startswith("functions."):
    arguments = "".join(text parts from msg.content)
```

Expected guard should require both:

```python
if (
    msg.channel == "commentary"
    and msg.recipient
    and msg.recipient.startswith("functions.")
):
```

## Why This Matters

Without the channel guard:

1. Harmony parsing can promote non-tool text into tool-call `arguments`
2. argument payload becomes malformed or huge
3. downstream validation rejects it as invalid JSON
4. users see noisy warnings and corrupted thought/reasoning output

With #2012, stream survival is improved, but malformed tool calls are still being generated unnecessarily.

## Minimal Repro Shape

Given parsed Harmony messages like:

- message A: `channel="analysis"`, `recipient="functions.read"`, content=`"analysis-noise"`
- message B: `channel="commentary"`, `recipient="functions.read"`, content=`'{"filePath":"/tmp/x","offset":1}'`
- message C: `channel="final"`, `recipient="functions.read"`, content=`"final-noise"`

Current behavior can extract A/B/C as tool-call material because recipient matches.

Expected behavior:

- only message B should be extracted as tool call
- message A should stay reasoning
- message C should stay final output

## Proposed Fix

In both Harmony extraction paths, require `msg.channel == "commentary"` before treating a message as a tool call.

Files/functions:

- `omlx/adapter/harmony.py: HarmonyStreamingParser.get_tool_calls()`
- `omlx/adapter/harmony.py: parse_tool_calls_from_tokens()`

## Regression Coverage

Add tests covering:

1. `analysis` message with `functions.*` recipient should not become tool call
2. `final` message with `functions.*` recipient should not become tool call
3. `commentary` message with `functions.*` recipient should still become tool call
4. mixed sequence should preserve channel separation correctly

## Relationship To #2012

- #2012: downstream recovery when malformed parser tool call already exists
- this issue: upstream fix so Harmony extraction does not create malformed tool calls from wrong channel content

e345ee:
HI men
Fixes #2032

This makes Harmony tool-call extraction a bit stricter: `functions.*` messages are now treated as tool calls only when they are in the `commentary` channel.

Previously, a `functions.*` recipient in `analysis` or `final` could be picked up as a tool call too, which does not match the Harmony format and can create false tool calls.

Changed both paths mentioned in the issue:
- `HarmonyStreamingParser.get_tool_calls()`
- `parse_tool_calls_from_tokens()`

Also added regression tests for mixed `analysis` / `commentary` / `final` messages to make sure only the `commentary` one becomes a tool call.

Tested with:

```bash
python -m pytest tests/test_harmony.py -q

I am waiting you opinion

### DIFF
--- a/omlx/adapter/harmony.py
+++ b/omlx/adapter/harmony.py
@@ -60,6 +60,15 @@ def _has_no_real_recipient(recipient: str | None) -> bool:
     return recipient is None or recipient == "<|start|>assistant"
 
 
+def _is_commentary_tool_call_message(msg: Any) -> bool:
+    recipient = getattr(msg, "recipient", None)
+    return (
+        getattr(msg, "channel", None) == "commentary"
+        and isinstance(recipient, str)
+        and recipient.startswith("functions.")
+    )
+
+
 @lru_cache(maxsize=1)
 def load_harmony_gpt_oss_encoding() -> HarmonyEncoding:
     """Load the Harmony gpt-oss encoding with a small retry window."""
@@ -333,7 +342,7 @@ class HarmonyStreamingParser:
                 return tool_calls
 
             for msg in messages:
-                if not msg.recipient or not msg.recipient.startswith("functions."):
+                if not _is_commentary_tool_call_message(msg):
                     continue
 
                 name = msg.recipient[10:]  # Remove "functions." prefix
@@ -470,7 +479,7 @@ def parse_tool_calls_from_tokens(
                     if isinstance(text, str):
                         analysis_text += text
 
-            elif msg.recipient and msg.recipient.startswith("functions."):
+            elif _is_commentary_tool_call_message(msg):
                 # Extract tool calls from commentary channel
                 name = msg.recipient[10:]  # Remove "functions." prefix
                 arguments = ""
--- a/tests/test_harmony.py
+++ b/tests/test_harmony.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests for Harmony streaming parser (omlx.adapter.harmony)."""
 
+from types import SimpleNamespace
 from unittest.mock import MagicMock
 
 import pytest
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for the fix, this is the right call. `get_tool_calls()` was matching on recipient alone, so any leftover `functions.*` recipient sitting on an analysis or final message got turned into a tool call. Gating on `channel == "commentary"` matches the harmony spec and lines up with how the streaming path already routes channels, so both paths are consistent now. I checked the new tests too, they fail on the old code and pass with your change. Merging this.

e345ee:
I like you product

PR #2147, opened 2026-07-09
fix(reranker): support reranker-native chat template for CausalLM rerankers

### LINKED ISSUE #2146: CausalLM reranker fails to load (500) for Qwen3-Reranker conversions that ship a reranker-native chat_template.jinja
## Summary

Loading `mlx-community/Qwen3-Reranker-0.6B-4bit` as a reranker fails with a 500, while `mlx-community/Qwen3-Reranker-0.6B-mxfp8` loads fine. The failure is not related to the quantization format — it is caused by the chat template shipped with the model.

## Root cause

Upstream `Qwen/Qwen3-Reranker-0.6B` was updated on 2026-04-16 ("Integrate with Sentence Transformers v5.4"). That update added a **reranker-native `chat_template.jinja`** which only understands `system` / `query` / `document` roles and silently drops `user` messages:

```jinja
{%- set query_text = messages | selectattr("role", "eq", "query") | map(attribute="content") | first -%}
{%- set document_text = messages | selectattr("role", "eq", "document") | map(attribute="content") | first -%}
...
<Instruct>: {{ instruction }}
<Query>: {{ query_text }}
<Document>: {{ document_text }}<|im_end|>
```

MLX conversions made **after** that update (e.g. `Qwen3-Reranker-0.6B-4bit`, converted 2026-06-06) inherit this template, and their `tokenizer_config.json` no longer carries the standard ChatML template at all. Conversions made **before** the update (e.g. `Qwen3-Reranker-0.6B-mxfp8`, converted 2026-02-11) still have the standard template, which is why they load fine.

`MLXRerankerModel._load_causal_lm` extracts the prompt prefix/suffix by rendering the chat template with a sentinel as the `user` content and splitting on it. With the reranker-native template the `user` message is dropped, the sentinel never appears in the rendered output, and loading fails with:

```
ValueError: Chat template produced unexpected format; could not split on sentinel.
```

which surfaces as a 500 on the rerank endpoints.

## Impact

- `mlx-community/Qwen3-Reranker-0.6B-4bit` and any Qwen3-Reranker conversion created after 2026-04 fail to load.
- Loading the updated upstream `Qwen/Qwen3-Reranker-0.6B` directly is affected as well, since transformers prioritizes `chat_template.jinja` over the template embedded in `tokenizer_config.json`.

## Reproduction

1. Download `mlx-community/Qwen3-Reranker-0.6B-4bit`.
2. Send any request to the rerank endpoint targeting that model.
3. Server returns 500; logs show the sentinel ValueError above.

## Suggested fix

When the sentinel split fails, detect the reranker-native template shape by rendering with `system`/`query`/`document` role sentinels and splitting around the combined `<Instruct>/<Query>/<Document>` block — the exact content `_rerank_causal_lm` reconstructs at scoring time. For the Qwen3-Reranker family both template shapes yield byte-identical prefix/suffix tokens, so scoring behavior is unchanged.

I have a fix ready and will open a PR shortly.

shyuan:
Fixes #2146

## Problem

`mlx-community/Qwen3-Reranker-0.6B-4bit` (and any Qwen3-Reranker conversion created after upstream's 2026-04 sentence-transformers update) fails to load with a 500. The inherited reranker-native `chat_template.jinja` only understands `system`/`query`/`document` roles and drops `user` messages, so the sentinel-based prefix/suffix extraction in `_load_causal_lm` never finds its sentinel and raises. See #2146 for the full analysis.

## Changes

- Extract the affix logic into `_extract_causal_lm_affixes`, which tries the standard system/user sentinel split first (unchanged behavior), then falls back to the reranker-native shape: render with per-slot sentinels for `system`/`query`/`document` and split around the combined `<Instruct>/<Query>/<Document>` block — the exact content `_rerank_causal_lm` reconstructs at scoring time. The native template already emits the trailing `<think>` block, so the suffix is used as-is.
- The byte-exact match against the native content block is intentionally strict: tolerating formatting drift would silently produce prompts that differ from what the template intends.
- When both template shapes fail, the error now includes both rendered attempts and both exceptions, chains the original exception (`raise ... from`), and rendering failures are logged at `warning` instead of `debug`.

## Verification

- New tests in `TestCausalLMPromptAffixes` cover: the standard template path, the native template path (rendering the upstream Qwen3-Reranker jinja byte-for-byte, including its trailing-newline behavior), the call-order fallthrough (standard attempt first, then native), rendering errors surfacing both attempts with exception chaining, and templates matching neither shape.
- Verified end-to-end against the real tokenizers of both `mlx-community/Qwen3-Reranker-0.6B-mxfp8` (standard path) and `mlx-community/Qwen3-Reranker-0.6B-4bit` (native fallback): both extract prefix/suffix successfully and produce **identical token sequences** (prefix 39 tokens, suffix 9 tokens), so scoring behavior is unchanged across the two paths.
- `pytest tests/test_reranker_causal_lm.py`: 28 passed. New code is black/ruff clean; remaining warnings in the touched files are pre-existing.

generated by Claude Code

### DIFF
--- a/omlx/models/reranker.py
+++ b/omlx/models/reranker.py
@@ -300,25 +300,7 @@ class MLXRerankerModel:
             )
 
         # Pre-compute prefix and suffix tokens for the prompt template.
-        # Use apply_chat_template() for portability across tokenizer formats,
-        # then split on a sentinel to extract prefix/suffix boundaries.
-        _SENTINEL = "<<__CONTENT_SENTINEL__>>"
-        messages = [
-            {"role": "system", "content": self._CAUSAL_LM_SYSTEM_PROMPT},
-            {"role": "user", "content": _SENTINEL},
-        ]
-        template_str = tokenizer.apply_chat_template(
-            messages, tokenize=False, add_generation_prompt=True
-        )
-        parts = template_str.split(_SENTINEL)
-        if len(parts) != 2:
-            raise ValueError(
-                f"Chat template produced unexpected format; "
-                f"could not split on sentinel. Template: {template_str!r}"
-            )
-        prefix = parts[0]
-        # Append <think> block for models that use thinking-then-answering format
-        suffix = parts[1] + "<think>\n\n</think>\n\n"
+        prefix, suffix = self._extract_causal_lm_affixes(tokenizer)
 
         self._prefix_tokens = tokenizer.encode(prefix, add_special_tokens=False)
         self._suffix_tokens = tokenizer.encode(suffix, add_special_tokens=False)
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for tracing this to the reranker-native template. I verified that both tokenizer paths reconstruct the exact prompt token sequence, and the 0.6B 4-bit model loads and scores correctly. LGTM, merging.

shyuan:
Cross-referencing #1964, which addresses the same root cause (the reranker-native chat template shipped with post-2026-04 Qwen3-Reranker conversions drops the `user` message, so the sentinel-based prefix/suffix extraction fails and every rerank call returns 500) with a different strategy:

- **#1964** falls back to a **built-in, hardcoded Qwen scoring frame**, gated to `Qwen3ForCausalLM`, so it also rescues models whose chat template is missing or broken entirely.
- **This PR** derives the frame from the **model's own reranker-native template** (per-slot sentinels split around the `<Instruct>/<Query>/<Document>` block), so it stays architecture-agnostic and follows whatever template the model actually ships. For the Qwen3-Reranker family, both template shapes yield byte-identical prefix/suffix tokens.

**Edit:** I originally posted this without noticing the PR had already been merged — apologies for the noise. Two edge cases first handled in #1964 (failing fast when `chat_template` is `None`, and not duplicating the `<think>` prefill when the template already emits one) missed the merge window here and are now proposed as follow-up #2196.

shyuan:
> Thanks for tracing this to the reranker-native template. I verified that both tokenizer paths reconstruct the exact prompt token sequence, and the 0.6B 4-bit model loads and scores correctly. LGTM, merging.

@jundot Thanks for the quick review and merge! And apologies — I posted the cross-reference comment above without noticing this PR had already been merged at that point.

After the fact I found that the pre-existing #1964 had tackled the same root cause and had already caught two hardening details worth keeping: failing fast when `chat_template` is `None`, and not duplicating the `<think>` prefill when the template already emits one. Those additions missed the merge window here, so I've proposed them separately as #2196.

Issue #1925, opened 2026-06-18
TurboQuant KV breaks prefix cache on hybrid (ArraysCache) models — cached=0 every turn


Tylerto:
**oMLX version:** 0.4.4 (also reproduced on the prior 0.3.x build)
**Hardware:** Apple Silicon, 48GB, macOS 26.5
**Model:** Qwen3.6-35B-A3B-MLX-4bit (hybrid: 40 layers, 10 KVCache)
**Config:** TurboQuant KV 4-bit ON, SpecPrefill ON (draft Qwen3.5-0.8B-MLX-8bit), paged SSD cache

**Summary**
With TurboQuant KV enabled on this hybrid model, the prefix cache never hits
(`cached 0` on every turn) and the full conversation is re-prefilled each turn.
Disabling TurboQuant immediately restores normal caching (`cached 94208`),
dropping per-turn latency from ~60–120s to ~10–20s.

**Root cause (from logs)**
Prefix-cache snapshots are stored/keyed as plain `KVCache`, but the live cache
expects `TurboQuantKVCache`, so the signature check fails and the prefix is
truncated to nothing on restore:

  WARNING omlx.cache.prefix_cache - Cache layer type mismatch at block 1:
  got [... 'KVCache' ...],
  expected [... 'TurboQuantKVCache' ...]. Truncating cached prefix before this block.

On load it also wipes the SSD cache for the same reason:

  Invalidated 53 SSD index entries with stale layer cache signature (kept 0)

**With TurboQuant ON (broken):**
  SpecPrefill: scored 84208 tokens ... (keep=20%, ... cached 0)
  Chat completion: 276 tokens in 48.94s, prompt: 103655

**With TurboQuant OFF (fixed):**
  SpecPrefill: scored 40729 tokens ... (keep=20%, ... cached 94208)
  Chat completion: 197 tokens in 13.18s, prompt: 134937

**Expected:** TurboQuant KV layer type should round-trip through the prefix
cache (or TurboQuant should be transparently bypassed for cache-snapshot
restore on hybrid models) so caching works with TurboQuant enabled.

[LogOMLX.rtf](https://github.com/user-attachments/files/29098149/LogOMLX.rtf)

aosama:
I reproduced this against current main and opened #2154 with a fix and regression coverage.

[...]

PR #2154, opened 2026-07-09
fix(cache): isolate SpecPrefill draft SSD cache

aosama:
## Summary

- give the SpecPrefill draft model its own model-scoped `PagedSSDCacheManager`
- connect the draft `PagedCacheManager` to that SSD manager so persisted draft blocks are discoverable after restart
- preserve the target model's TurboQuant layer signature while validating draft blocks against the draft model's own layer count and cache types
- close and release the draft manager during replacement, reset, shutdown, and failed setup
- add model-free regression coverage for target/draft signature isolation, cold SSD discovery, reconstruction, and teardown

## Root cause

The log attached to #1925 contains a stronger signal than the original summary: the cached signature has **24 layers**, while the live expected signature has **40 layers**.

Those layouts match the configured models exactly:

- Qwen3.5-0.8B SpecPrefill draft: 24 layers
- Qwen3.6-35B-A3B target: 40 layers

The draft `BlockAwarePrefixCache` used a separate `PagedCacheManager`, but shared the target model's `PagedSSDCacheManager`. Once TurboQuant installed the target's 40-layer signature, draft reconstruction compared its 24-layer blocks against that target signature and emitted the reported mismatch. The draft paged manager was also not connected to SSD, so persisted draft hashes could not be discovered after process restart.

This change scopes the draft SSD index and signature to the draft model while retaining the existing shared cache directory and hot-cache budget.

### Automated

- the regression test fails on `v0.4.4`
- it also fails on current main before this change
- it passes with this change
- 509 relevant tests pass:
  - `tests/test_scheduler.py`
  - `tests/test_specprefill.py`
  - `tests/test_vlm_specprefill.py`
  - `tests/test_prefix_cache.py`
  - `tests/test_paged_ssd_cache.py`
  - `tests/test_engine_preflight.py`
  - `tests/test_scheduler_prefill_memory_guard.py`

### Real-model validation

Tested on an M5 Pro with 48 GB unified memory using:

- target: `mlx-community/Qwen3.6-35B-A3B-4bit`
- draft: `mlx-community/Qwen3.5-0.8B-MLX-8bit`
- TurboQuant KV: 4-bit, skip-last enabled
- SpecPrefill: enabled, 20% keep
- paged SSD cache enabled

Observed:

- same-session second request: `draft cache hit 8192`
- first matching request after server restart: `draft cache hit 8192`
- no `Cache layer type mismatch` warning
- with SpecPrefill disabled as a control, TurboQuant target prefix caching restored `8192` target tokens on the second request

## Metric clarification

With SpecPrefill enabled, API `prompt_tokens_details.cached_tokens` remains `0` because it represents the target prefix cache, and sparse target KV states are intentionally not persisted. SpecPrefill draft reuse is reported separately in the server log as `draft cache hit N`. The repaired behavior here is successful draft-cache reuse without cross-model signature invalidation.

Addresses #1925.

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -1653,6 +1653,7 @@ class Scheduler:
 
         # SpecPrefill: draft model for attention-based sparse prefill
         self._specprefill_draft_model: Any | None = None
+        self._draft_paged_ssd_cache_manager: Any | None = None
         # Track active specprefill request for RoPE cleanup
         self._specprefill_active_request_id: str | None = None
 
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for tracking this down. The draft-scoped SSD manager fixes the signature conflict without affecting other cache paths. I verified the focused regression tests. LGTM, merging.

PR #2159, opened 2026-07-10
fix(oq): decode U8 e8m0 block scales in MXFP8 dequant

builtbycnob:
## Problem

`_block_dequant_fp8` only applies the e8m0 exponent decode (`2^(s-127)`) when the
scale tensor has safetensors dtype `F8_E8M0`. MXFP8 checkpoints such as MiniMax-M3
store their OCP microscaling shared exponents as dtype **U8** instead — one scale
byte per 32-column block, alongside an `F8_E4M3` weight. Those bytes are e8m0
exponents, but the U8 dtype misses the check, so they were multiplied in as raw
linear scale values.

The result is silently catastrophic: dequantized weights come out roughly five
orders of magnitude too large. On `layers.3.self_attn.k_proj` of MiniMax-M3 the
mean absolute weight is 13410 instead of 0.0297 (~450,000x off). Because
`quantize_oq_streaming` reads its source through the same dequant path, **every
quantization built from an MXFP8 checkpoint with U8 scales is corrupted**, not just
direct loads.

Reproduced by hand against two independent references (an explicit `2^(s-127)`
decode and `mx.dequantize(mode="mxfp8")`), which agree bit-for-bit; the current
code disagrees by ~450,000x.

## Fix

Treat a scale as e8m0 when it is `F8_E8M0`, or when it is `U8`/`UINT8` paired with
an fp8 weight in the block layout (2-D scale, weight columns == scale columns * 32).
The discriminator is the scale **dtype and layout**, not the key name: DeepSeek's
`weight_scale_inv` scales are stored as float in a block-128 layout and are genuinely
linear, so they must keep flowing through the linear branch. The added condition
does not touch them.

## Tests

`tests/test_oq_fp8_dequant.py`:
- model-free fp8 block dequant checked bit-exact against `mx.dequantize(mode="mxfp8")`
- a float-scale regression case that asserts linear scales stay linear (DeepSeek path)
- a magnitude sanity check on a real MiniMax-M3 tensor

All three pass on top of current `main` (0.5.0rc1).

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -1367,7 +1367,17 @@ _FP8_WEIGHT_DTYPES = frozenset(("F8_E4M3", "F8_E5M2", "I8"))
 
 def _block_dequant_fp8(weight_raw, scale_raw, w_dtype, s_dtype):
     """Block-scaled dequant of a single FP8/I8 weight+scale pair to BF16."""
-    if s_dtype == "F8_E8M0":
+    # MXFP8 checkpoints (e.g. MiniMax-M3) save the e8m0 shared exponents
+    # with safetensors dtype U8: fp8 weight, one scale byte per 32 columns.
+    # Those bytes are exponents, not linear scales. Float-typed scales
+    # (DeepSeek block-128 weight_scale_inv) stay linear.
+    e8m0 = s_dtype == "F8_E8M0" or (
+        s_dtype in ("U8", "UINT8")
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for tracking this down. The U8 + FP8 + 1x32 layout gate matches the MXFP8 scale format, while float block scales stay on the existing linear path. I verified the output against `mx.dequantize` and ran the OQ regression suite. LGTM, merging.

PR #2161, opened 2026-07-10
feat(oq): layer-streaming imatrix calibration for models larger than RAM

builtbycnob:
## Problem

oQe activation-importance (imatrix) calibration runs a forward pass over the model
while it is fully resident. For a source that is larger than RAM the collector falls
back to a uniform quantized proxy, and the proxy also has to fit in memory. On a 256 GB
machine that ceiling is reached well before the largest MoE checkpoints: MiniMax-M3
(about 428B total) needs a 4-bit proxy of roughly 225 GB, which jetsam-kills the process
before the calibration forward finishes. There is no way to imatrix-calibrate a model of
that size on the box, so it has to be quantized without imatrix, which is exactly where
the low-bit routed experts need the importance weighting most.

## Approach

This adds a layer-streaming calibration path that inverts the loop. Instead of holding
the whole model (or a proxy) resident and iterating samples over it, it iterates layers
over the samples:

1. Source one decoder layer's weights directly from the checkpoint (lazy, one layer).
2. Run every calibration micro-batch through that layer, capturing the imatrix statistics
   and, on the first round, the fused qdq sensitivity for the same residency.
3. Release the layer, feed the cached activations to the next layer, repeat.

At most one layer's weights are resident at a time. The imatrix accumulation is an
additive per-token sum, so layer-outer collection is bit-for-bit identical to the existing
sample-outer collection; tests assert that equivalence on fixtures. Adaptive expert
coverage still works: escalation adds whole rounds so every layer sees the same sample
multiset. A hard per-layer active-memory ceiling aborts if a layer fails to release,
rather than letting a leak compound across the sweep.

## Rebased onto current main (v0.6.3)

The branch is now a rebase of the original series onto current `main`. The MXFP8 U8
e8m0 dequant fix that used to open this series is gone from the diff — main already
carries it, with a superset test. Three call sites were reworked against code that moved
since the original base: the imatrix cache-hit branch calls the three-argument
`_oqe_cache_missing_mtp_entries` and keeps `_oqe_cache_has_required_expert_coverage`
in the chain, and the auto-stream rule keys off the calibration budget's
`requires_proxy` rather than the old exceeds-RAM boolean, so streaming engages exactly
where the proxy build would otherwise kick in.

## Rebased onto current main (2026-09-07)

Re-based again onto current `main` (`aa8db73`, post-0.6.4). One conflict, in the
write-path sanitizer call: main's `_build_model_sanitizer` grew `model_path` and
`preserve_mtp`, this branch prepends the qwen4_exp package exposure and the
refuse-on-missing-sanitizer guard. Resolved as both — the package is exposed first,
then main's signature is called, then the guard runs. Nothing else in the series moved.

The proxy policy that landed meanwhile (#3269 and its follow-up) and this path
coexist by construction rather than by luck: `load_path_factory` is passed only when
`_model_requires_proxy and not stream_imatrix`, so for a checkpoint the streamed
sourcer supports the real calibration runs and the proxy is never built; for every
other oversized checkpoint main's proxy still handles it. The automatic rule keys off
the same `requires_proxy` budget flag, so the two policies read one decision.

## Second supported architecture: qwen4_exp

The sourcer now handles Qwen4-Exp checkpoints (Qwen3.8-Flash-Next: 48 MoE layers,
36 GDN + 12 QSA, a 20M-row N-gram PLE table at layer 2, live MTP weights). The pieces
that were qwen4_exp-shaped rather than generic:

- The forward state is built by hand instead of through `_prepare_layer_inputs`, whose
  qwen4_exp branch gates on object identity with a live `model.layers` list that a
  one-block-at-a-time generator never has. The state carries the hc-tiled boundary and a
  per-layer mask schedule from `config.text_config.layer_types` (None for
  linear_attention, causal for full_attention) — one shared mask would calibrate every
  QSA layer non-causally.
- The PLE runtime is pinned to mmap on the module instance the layer class actually
  reads (the pre-load patches can re-install the vendored module), and the 128 N-gram
  shard tensors the sanitize plan still carries are filtered before the pop — an
  mmap-mode block self-sources them, and popping them first materialized the whole
  ~100 GB table once per sweep just to discard it.
- Streamed blocks run in eval mode. Freshly constructed modules default to training
  mode, and the Qwen3.5-family GDN picks its fused kernel with
  `use_kernel=not self.training`; the training-mode scan diverges by one bf16 ulp per
  forward, which is enough to flip borderline MoE routing choices and break bitwise
  parity with the resident collector.
- A build that drops the MTP head (`preserve_mtp=False`) no longer invalidates its
  imatrix cache on every reuse: `require_mtp_entries` is threaded from `preserve_mtp`,
  since a cache without mtp.* entries is complete for a build that never quantizes
  them. Without this, any source that ships real MTP weights recollects the full
  imatrix on every quantization run.
- `quantize_oq_streaming` grows `preserve_ngram_table`: the N-gram shard tensors are
  written unquantized in the checkpoint dtype with no per-layer quantization entry, so
  the table can be quantized separately from the core (its measured sensitivity is far
  below the core's). The synthetic `weight_scale` the mmap-mode embedding registers is
  stripped from the output; it is not checkpoint data and strict loaders reject
  artifacts that carry it.

## Result

- MiniMax-M3 (MXFP8, no truncation): calibrates with a peak around 58 GB instead of
  the 225 GB proxy that OOM'd, as before.
- Qwen3.8-Flash-Next (335 GB bf16, larger than RAM): a full fused
  imatrix+sensitivity collection (128x512, adaptive escalation to 1,024 samples,
  8 rounds) runs in 50 minutes with ~21 GB active, and the streamed collector
  reproduces the resident block-walk on a truncated real-weights fixture fully
  bitwise: exact expert routing counts and bit-identical `in_sum2` on every entry,
  fused qdq sensitivity within 1e-6 of `_measure_sensitivity_from_model`.

## Interface

Opt-in and additive; the resident path is unchanged. `stream_calibration=True`, the
`OMLX_OQ_STREAM_CALIBRATION` env var, or automatic when the calibration budget says the
source needs a proxy. The imatrix cache records how it was collected (`load_kind`), so
a proxy-collected cache cannot silently satisfy a streaming request and vice versa.

## Widening the supported set

The set is deliberately small: a model_type joins
`_STREAM_CALIBRATION_SUPPORTED_MODEL_TYPES` only once its layer dispatch is written
and its bitwise parity gate passes, because a checkpoint whose forward state the
sourcer builds wrongly produces a plausible imatrix that is quietly wrong. A
production report on glm5_next (GLM-5.3-Flash, 45 layers, 306 GB source, M1 Ultra
128 GB) is in the thread below: 45/45 layers in ~23 minutes, 671 entries covering 682
tensors, parity gate green, and the dispatch it needed is the hyper-connection
coefficients riding along per layer plus the MTP head as a 46th block with its own
embedding pass. That dispatch is welcome as a follow-up on top of this once the
sourcer lands.

Two notes from that report worth recording here: the `.npz` cache is written once at
the end of a collection, so a failure in the last layer costs the whole sweep; and a
model with an MTP head needs the embedding table re-read for the head's pass, so the
release has to happen after the layer loop rather than before the head. Both are
incremental improvements on the same file and are better as separate changes than as
additions to this diff.

## Tests

- `test_oq_stream_resolve.py`: the decision precedence and the pinned supported set.
- `test_oq_stream_sourcer.py` / `test_oq_stream_collect.py` /
  `test_oq_stream_sensitivity.py`: the MiniMax fixture suite, unchanged (skips when
  the checkpoint is absent).
- `test_oq_stream_qwen4_exp.py`: the qwen4_exp mask schedule, per-slot state, the
  pre-pop shard filter (plan values are poisoned to prove skipped keys are never
  materialized), the table-preserve predicate, and both settings of
  `require_mtp_entries`.
- `test_oq_stream_qwen4_exp_parity.py`: the bitwise streamed-vs-resident gate on a
  truncated real-weights Flash-Next fixture (GDN + PLE + QSA layers), including the
  fused-sensitivity comparison.

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -1367,7 +1367,17 @@ _FP8_WEIGHT_DTYPES = frozenset(("F8_E4M3", "F8_E5M2", "I8"))
 
 def _block_dequant_fp8(weight_raw, scale_raw, w_dtype, s_dtype):
     """Block-scaled dequant of a single FP8/I8 weight+scale pair to BF16."""
-    if s_dtype == "F8_E8M0":
+    # MXFP8 checkpoints (e.g. MiniMax-M3) save the e8m0 shared exponents
+    # with safetensors dtype U8: fp8 weight, one scale byte per 32 columns.
+    # Those bytes are exponents, not linear scales. Float-typed scales
+    # (DeepSeek block-128 weight_scale_inv) stay linear.
+    e8m0 = s_dtype == "F8_E8M0" or (
+        s_dtype in ("U8", "UINT8")
+        and w_dtype in ("F8_E4M3", "F8_E5M2")
+        and scale_raw.ndim == 2
+        and weight_raw.shape[-1] == scale_raw.shape[-1] * 32
+    )
+    if e8m0:
         scale = mx.power(mx.array(2.0), scale_raw.astype(mx.float32) - 127.0)
     else:
         scale = scale_raw
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -16,9 +16,10 @@ import re
 import shutil
 import tempfile
 import time as _time
+from collections.abc import Callable
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Callable, Optional, Union
+from typing import Any
 
 import numpy as np
 
@@ -149,7 +150,7 @@ class OQImatrixData:
 
 def universal_quant_predicate(
     path: str, module, config: dict, oq_level: int = 4
-) -> Union[bool, dict]:
+) -> bool | dict:
     """Per-tensor quantization decision based on GGUF/unsloth/llama.cpp rules.
 
     Protection levels vary by oQ level:
@@ -964,8 +965,7 @@ def _build_quant_plan(
             f"{name}×{count}" for name, count in sorted(route_dist.items())
         )
         logger.info(
-            f"  plan detail: {bits_summary} | routes: {route_summary} | "
-            f"top: {top_str}"
+            f"  plan detail: {bits_summary} | routes: {route_summary} | top: {top_str}"
         )
 
     return QuantPlan(
@@ -2142,7 +2142,7 @@ def _sensitivity_lm_config_override(config: dict) -> dict | None:
 def make_predicate(config: dict, oq_level: int = 4) -> Callable:
     """Create a quant_predicate closure for mlx-lm's quantize_model."""
 
-    def predicate(path: str, module) -> Union[bool, dict]:
+    def predicate(path: str, module) -> bool | dict:
         return universal_quant_predicate(path, module, config, oq_level)
 
     return predicate
@@ -3924,7 +3924,7 @@ def quantize_oq_streaming(
     output_path: str,
     oq_level: int,
     group_size: int = 64,
-    progress_callback: Optional[Callable[[str, float], None]] = None,
+    progress_callback: Callable[[str, float], None] | None = None,
     text_only: bool = False,
     target_bpw: float | None = None,
     hard_cap_bpw: float | None = None,
@@ -4094,8 +4094,7 @@ def quantize_oq_streaming(
         if _ram_safe_proxy_dir is not None and _ram_safe_proxy_dir.exists():
             shutil.rmtree(_ram_safe_proxy_dir, ignore_errors=True)
             logger.info(
-                f"oQ{oq_level:g}: cleaned up calibration proxy at "
-                f"{_ram_safe_proxy_dir}"
+                f"oQ{oq_level:g}: cleaned up calibration proxy at {_ram_safe_proxy_dir}"
             )
         _ram_safe_proxy_dir = None
 
@@ -4140,9 +4139,7 @@ def quantize_oq_streaming(
                 progress_callback=cb,
                 progress_start=13.0,
                 progress_end=18.0,
-                load_path_factory=(
-                    _imatrix_load_path if _model_exceeds_ram else None
-                ),
+                load_path_factory=(_imatrix_load_path if _model_exceeds_ram else None),
             )
         except BaseException:
             _cleanup_ram_safe_proxy()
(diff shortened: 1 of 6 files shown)


jundot:
Thanks, inverting the loop to layer-outer is a clean way to kill the proxy OOM and the bitwise parity gate is exactly the right test for it. One blocker before I can merge: the auto rule in _resolve_stream_calibration turns streaming on for every source that exceeds RAM, but the sourcer is minimax_m3_vl only (hardcoded language_model.model.layers prefix, mlx-vlm load in _streamed_text_args, block.is_moe_layer). Any oQe build of a big non-MiniMax model that completes today through the RAM-safe proxy would start crashing by default. Can you gate the auto rule on the supported model_type and raise a clear early error when stream_calibration=True hits an unsupported layout?

builtbycnob:
Pushed the gate in a46c0857.

The supported set is a named constant next to the sourcer now (`_STREAM_CALIBRATION_SUPPORTED_MODEL_TYPES`, `{minimax_m3_vl}` for the moment), and `_resolve_stream_calibration` keys off it:

- auto rule is `model_exceeds_ram and supported`, so a large non-MiniMax build stays on the RAM-safe proxy by default instead of streaming into the minimax_m3_vl-only sourcer.
- explicit `stream_calibration=True` on an unsupported layout raises a ValueError before any load that names the model_type and the supported set, instead of the AttributeError deep in the sourcer.
- a truthy `OMLX_OQ_STREAM_CALIBRATION` on an unsupported layout is a soft preference: it logs a warning naming the type and falls back to the proxy, so only the explicit argument is a hard error.

The call site passes the checkpoint's top-level model_type, which I checked reads minimax_m3_vl on the real MiniMax-M3 config.

Tests: added tests/test_oq_stream_resolve.py (pure decision logic, no checkpoint needed) covering the three cases: auto keeps the proxy for an unsupported type over RAM, explicit True on an unsupported type raises, and minimax_m3_vl still auto-streams. I threaded model_type through the existing precedence test; the bitwise parity gate is untouched and green.

Widening to a second architecture is then the dispatch table plus its model_type in that one set.

pedroberaldo87:
Production report from a third architecture, in case it helps land this.

I have been running this PR's streamed collector on **glm5_next** (GLM-5.3-Flash:
45 layers, 306 GB source, M1 Ultra 128 GB) for the last two weeks. Widening it
was what the last message here predicted — the dispatch for the family plus its
model_type in `_STREAM_CALIBRATION_SUPPORTED_MODEL_TYPES`:

```python
_STREAM_CALIBRATION_SUPPORTED_MODEL_TYPES = frozenset(
    {"minimax_m3_vl", "qwen4_exp", "glm5_next"}
)
```

The glm5_next dispatch is small: the hyper-connection (`hc_mult=4`) widens the
carried stream and its coefficients have to ride along per layer, and the
multi-token-prediction head is a 46th block that needs its own pass with the
embedding table. Nothing in the collector loop itself changed.

This is the case the PR exists for — the model cannot sit in RAM next to a
calibration pass at all, so before it there was no importance matrix for
GLM-5.3, only the proxy. With it:

```
45/45 layers collected                     ~23 min
671 imatrix entries, 682 tensors covered   missing: none
output                                     102 GB oQ2e, bitwise parity gate green
```

Two things that cost me time, in case they save someone else's:

1. **The `.npz` cache is only written at the END of the collection.** An error
   in the last layer costs the whole 23 minutes. Worth writing it incrementally,
   or at least saying so in the docstring.

2. **A model with an MTP head needs the embedding table re-read for the head's
   pass.** Releasing it after the layer loop and reading it again for the head
   is the order that works; releasing it before the head pass raises an
   `UnboundLocalError` that only reproduces on a checkpoint that actually has
   the head — a toy-model test does not catch it.

Happy to send the glm5_next dispatch as a follow-up on top of this once it
merges, or push it here if that is easier for you.

builtbycnob:
Thank you for running it on a third architecture. Both notes are recorded in the description as follow-ups on the same file, kept out of this diff so it stays reviewable: the per-layer cache checkpoint, and a test for the head pass on a checkpoint that actually has a head, which is the only kind that reproduces the ordering you hit.

On the glm5_next dispatch: a follow-up PR on top of this one once it merges is the cleanest path, so this series stays the streaming path plus two architectures and yours stays reviewable on its own. If the maintainer prefers everything in one series I am happy to take it here instead.

jundot (approved):
Layer-wise sourcing avoids keeping a full calibration proxy resident. I verified exact trunk and MTP imatrix parity on a four-layer Flash-Next FP8 fixture with the maintainer follow-up applied. I'll apply that follow-up immediately after merging to preserve PLE scales, complete MTP collection, fix batched MTP positions, avoid whole-model FP8 loading, and consolidate the tests. LGTM with those fixes ready.

Issue #2049, opened 2026-06-30
Token sampling crashes with "Thread group size (1024) is greater than maximum allowed threads per threadgroup (896)" on M2 Ultra


dawid-szewc:
### Describe the bug

Chat completion on Qwen3.6-27B-8bit intermittently crashes during token sampling with:

```
ValueError: Thread group size (1024) is greater than the maximum allowed threads per threadgroup (896).
```

Traceback:

```
  File "/opt/homebrew/Cellar/omlx/0.4.4/libexec/lib/python3.11/site-packages/mlx_lm/generate.py", line ~1360, in _step
    tokens, logprobs = self._step()
                        ^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/omlx/0.4.4/libexec/lib/python3.11/site-packages/omlx/scheduler.py", line 620, in _patched_generation_batch_step
    result = _original_generation_batch_step(self)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/omlx/0.4.4/libexec/lib/python3.11/site-packages/mlx_lm/generate.py", line 1369, in _step
    mx.async_eval(self._next_tokens, self._next_logprobs, token_context)
ValueError: Thread group size (1024) is greater than the maximum allowed threads per threadgroup (896).
```

Server log around the crash:

```
2026-06-30 17:23:04,176 - omlx.server - ERROR - [-] - Error during chat streaming: Thread group size (1024) is greater than  the maximum allowed threads per threadgroup (896).
2026-06-30 17:23:04,176 - omlx.server - ERROR - [-] - Error during chat streaming: Thread group size (1024) is greater than  the maximum allowed threads per threadgroup (896).
2026-06-30 17:24:41,311 - omlx.scheduler - INFO - [-] - Cache phase timings: boundary_capture_extract=4.4ms/4, boundary_capture_sync=23.7ms/4, boundary_snapshot_save=63.4ms/4, cleanup_finished_sync=15.2ms/4, store_cache_main_boundary=10.0ms/4, store_cache_main_collect=0.2ms/4, store_cache_main_dispatch=114.1ms/4, store_cache_worker_sync=0.5ms/5
2026-06-30 17:24:41,313 - omlx.server - INFO - [-] - Chat completion: 1768 tokens in 96.99s (18.2 tok/s), finish_reason=stop, prompt: 3728, max_tokens=32768
```

[...]

PR #2165, opened 2026-07-10
fix: fall back for Qwen3.5/3.6 ragged decode kernels exceeding threadgroup limit

mxl:
## Problem

During Throughput benchmark (and in production batched decode with unequal prompt lengths), Qwen3.5/3.6 models crash with:

```
ValueError: Thread group size (1024) is greater than the maximum allowed threads per threadgroup (896).
```

The error fires inside `mx.async_eval` during `generate.py:_step`, killing the whole batch mid-generation. Relates to #2049.

## Root cause

`mlx_vlm/models/qwen3_5/language.py` dispatches both `one_pass` and `two_pass_2` ragged left-padded decode SDPA kernels with hardcoded `threadgroup=(1024, 1, 1)`. On some GPUs (M1 Max confirmed, also M2 Ultra per #2049), the two-pass-2 pipeline compiles with `maxTotalThreadsPerThreadgroup=896` due to register pressure at `D_SIZE=256`.

`MTLDevice.maxThreadsPerThreadgroup` reports `1024` (hardware cap), but `MTLComputePipelineState.maxTotalThreadsPerThreadgroup` is per-kernel and authoritative — it can be lower depending on register/threadgroup memory usage of the compiled kernel.

## Fix

New patch `omlx/patches/qwen35_ragged_decode.py` monkey-patches `_qwen3_5_ragged_decode_attention`:

1. On first call for each unique kernel signature (mode, blocks, dtype, D_SIZE, V_SIZE, heads), probes the real launch with an eager `mx.eval`.
2. If the launch raises the threadgroup limit error, caches `False` for that signature and returns `None`.
3. Returning `None` routes the caller (`_target_verify_left_padded_attention`, line 1322) to the existing row-grouped reference fallback — no new math, same numerics.
4. Supported signatures cache `True` after the first probe; hot path is a single dict lookup thereafter.

Wire-up follows the same pattern as `sdpa256_attention` and `qwen35_fa256_attention`: installed in `BatchedEngine` and `VLMBatchedEngine` after the other Qwen prefill patches, gated by `model_settings.qwen35_ragged_decode_fallback_enabled` (default `True`).

## Tests

`tests/test_qwen35_ragged_decode.py` (9 tests, no model needed):
- Fallback on 896 threadgroup error → returns `None`, caches `False`, original called once only
- Passthrough when supported → returns output, caches `True`
- Cache hit on second call (probe runs once per signature)
- Non-threadgroup errors propagate unchanged
- Install idempotent (`apply_...()` returns `True` first time, `False` on repeat)
- Import error returns `False` without patching

```
pytest tests/test_qwen35_ragged_decode.py -q
9 passed in 8.52s
```

### DIFF
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -376,6 +376,19 @@ class BatchedEngine(BaseEngine):
                     "Qwen MoE weighted-sum patch not applied", exc_info=True
                 )
 
+        if (
+            getattr(self._model_settings, "qwen35_ragged_decode_fallback_enabled", True)
+            is not False
+        ):
+            try:
+                from ..patches.qwen35_ragged_decode import (
+                    apply_qwen35_ragged_decode_patch,
+                )
+
+                apply_qwen35_ragged_decode_patch()
+            except Exception:
+                logger.debug("qwen3_5 ragged decode patch not applied", exc_info=True)
+
         # Create engine config (copy to avoid mutating the shared instance)
         scheduler_config = (
             copy.copy(self._scheduler_config)
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -1581,6 +1581,19 @@ class VLMBatchedEngine(BaseEngine):
                 logger.debug(
                     "Qwen MoE weighted-sum patch not applied", exc_info=True
                 )
+
+        if (
+            getattr(self._model_settings, "qwen35_ragged_decode_fallback_enabled", True)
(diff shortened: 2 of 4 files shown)


jundot:
Thanks for tracking this down. The per-signature probe catches the deferred pipeline limit while keeping capable devices on the fused path, and I verified that the existing row-grouped fallback preserves output parity. LGTM after the latest-main and fallback checks; merging.

Issue #2150, opened 2026-07-09
Lightning MTP (PR #2113) regresses tgTPS by 8-12% on Qwen3.6-35B-A3B (MoE) vs. old depth-1 MTP, and is fully bypassed during continuous batching


fparrav:
## Environment
- oMLX version: 0.5.0rc1 (release candidate build, commit `7fd7a4fc`; Lightning MTP landed in `8a9b1972`, PR #2113)
- Hardware: MacBook Pro (Mac16,5), Apple M4 Max (16-core CPU: 12P+4E, 40-core GPU), 64 GB unified memory, macOS 26.5.1
- Model: `Qwen3.6-35B-A3B-oQ4-mtp` (local repo `Jundot/Qwen3.6-35B-A3B-oQ4-mtp`, `model_type=qwen3_5_moe`, 21.14 GB on disk)
- Benchmark tool: built-in admin Throughput benchmark (`omlx/admin/benchmark.py`), loaded with `force_lm=True`

## Summary

On an M4 Max, Qwen3.6-35B-A3B-oQ4-mtp is **8-12% slower** with Lightning MTP (the depth-k adaptive rebuild from #2113) than it was with the old fixed depth-1 MTP cycle, at every context length tested. Additionally, native MTP is completely inactive during the built-in benchmark's continuous-batching phase (2x/4x), so the reported batching speedup reflects plain continuous batching only, not MTP.

This is on the exact model/quantization named in the #2113 release notes as covered by Lightning MTP, so it's presumably meant to see the reported uplift — it's regressing instead.

## Data: tgTPS before vs. after #2113 (same model, same M4 Max 64GB, single request, tg=128)

| Context | Old MTP (depth-1, pre-#2113) | Lightning MTP (depth cap=2, this build) | Δ |
|---|---|---|---|
| pp1024  | 121.0 tok/s | 109.2 tok/s | **-9.7%** |
| pp8192  | 114.0 tok/s | 100.4 tok/s | **-11.9%** |
| pp32768 | 95.8 tok/s  | 88.2 tok/s  | **-7.9%** |
| pp131072| 61.1 tok/s  | 53.7 tok/s  | **-12.1%** |

Full single-request table from this run (built-in Throughput benchmark, `Engine: Auto`):

```
Test                TTFT(ms)    TPOT(ms)        pp TPS        tg TPS      E2E(s)    Throughput    Peak Mem
[...]

PR #2166, opened 2026-07-10
fix(mtp): unblock MTP under continuous batching — singleton recovery + opt-in row-wise batch path (#2150)

gilby:
Addresses the second half of #2150 — "MTP is fully inactive during continuous batching" — and fixes a related regression the same gates cause: a request that ever shares a decode step with another request loses MTP for the rest of its generation, even once it is running alone again.

## Root cause: two gates make the row-wise batch path unreachable

**1. The alignment gate can never pass under continuous batching.** `_batch_rows_aligned_for_mtp` requires all rows to share one cache offset at activation. Rows join a batch at different times (even the built-in throughput benchmark's simultaneous 2x/4x phases prefill sequentially), so this never holds — instrumenting the benchmark showed **254/254 batched decode steps** blocked with `row-wise MTP requires aligned target-cache positions`. The requirement is also unnecessary: activation seeds each row from its own `extract_cache(idx)` view, and steady-state row cycles diverge the per-row offsets immediately anyway (accept counts differ per row), so the merge path already handles ragged rows.

**2. A chicken-and-egg deadlock on the multirow marker.** `_allows_new_mtp_activation` consults `_omlx_mtp_saw_standard_multirow_decode` for the **batch** state attr too — but a batch's first decode step is always standard (MTP hasn't activated yet), which sets the marker and locks batch activation out permanently. The marker's own docstring scopes it to *singleton* re-initialization after reshaping ("Existing row-wise MTP state may continue, but starting a fresh **singleton** MTP state ... is unsafe").

The same marker causes the hidden real-world regression: when a batch shrinks back to one row, fresh singleton activation stays blocked forever, so any overlap permanently downgrades the surviving request to standard decode.

## What this PR does

**Singleton MTP recovery after batch shrink (default on).** When a batch is back to one row and every layer cache is *verifiably* compact (`BatchKVCache.filter()` shifts out the minimum shared left padding, so filtering down to one row compacts fully — checked per layer rather than assumed), the marker is lifted and the surviving request regains MTP. Measured on M3 Ultra / Qwen3.6-27B-oQ4-mtp (3-layer MTP head, depth 3): the survivor re-activates and finishes at **77–83% draft acceptance (~63 tok/s) instead of 35.7 tok/s standard decode**.

**Row-wise batch MTP fixed, but opt-in (`OMLX_MTP_ROWWISE_BATCH=1`).** With the two gates fixed, the path activates and works correctly — `MTP row-wise batch path activated for 3 sequences`, 83–93% per-row acceptance, coherent outputs. But it should not be the default: the row-wise design runs **one backbone forward per row per cycle**, so its aggregate throughput is roughly single-stream MTP throughput regardless of batch size, while standard batched decode amortizes one forward over all rows. Built-in throughput benchmark, same model/hardware, pp1024/tg128:

| | single | batch 2 (agg) | batch 4 (agg) |
|---|---|---|---|
| standard decode (MTP off) | 35.7 tok/s | **65.2 tok/s** | **86.5 tok/s** |
| row-wise MTP forced on | **63.2 tok/s** | 53.3 tok/s | 52.5 tok/s |

So at batch ≥ 2 standard batching wins despite high acceptance; row-wise only pays off when tokens-per-cycle exceeds the row count (deep chains / very high acceptance). Hence: off by default, one env flag to experiment, and existing batch state always continues (a config flip never tears down a mid-flight batch).

**Diagnosability.** One INFO line per multi-row batch explaining why MTP is inactive (`MTP inactive for 2-row batch: standard batched decode is faster at this batch size (set OMLX_MTP_ROWWISE_BATCH=1 to force row-wise MTP)`), and `_ineligibility_reason` now covers the batch-activation gates. This directly addresses the issue's observation that the batched benchmark numbers silently exclude MTP — the fallback was previously DEBUG-only and easy to misread.

## Validation

- `tests/test_mlx_lm_mtp_patch.py`: 100 passed (updated the alignment-gate test to the new contract; added opt-in-default-off, activation-after-multirow, and recovery tests). The existing `test_singleton_activation_blocked_after_standard_multirow_decode` still passes unchanged — singleton blocking semantics are preserved until the cache is verifiably compact.
- Live server (M3 Ultra, 0.5.0-lineage): default config — 3 simultaneous requests decode as a standard batch with the INFO line, then `MTP singleton recovery: multirow marker cleared (compact cache)` fires when the short requests finish and the long one completes 522/600 tokens under MTP at 77.2% acceptance, output coherent. With `OMLX_MTP_ROWWISE_BATCH=1` — row-wise activates for 3 sequences, 92%+ acceptance, coherent outputs, no fallback errors.

One observed behavior worth noting for anyone reproducing: staggered arrivals usually don't form a batch at all while a singleton MTP stream is active (`patched_bg_next` forces `completion_batch_size = 0`), so they queue and run serially; multi-row batches mostly form from (near-)simultaneous arrivals. That interaction is untouched here but affects which of these paths a given workload hits.

### DIFF
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -71,6 +71,7 @@ from __future__ import annotations
 
 import logging
 import math
+import os
 from collections import deque
 from dataclasses import dataclass, field
 from types import SimpleNamespace
@@ -135,6 +136,7 @@ def apply() -> bool:
             realign_rows = getattr(self, "_omlx_realign_rows", None)
             if callable(realign_rows):
                 realign_rows()
+            _maybe_clear_multirow_marker(self)
 
             if _is_mtp_batch_eligible(self):
                 try:
@@ -159,6 +161,7 @@ def apply() -> bool:
                     _drop_mtp_state(self, "step-fallback")
             else:
                 _drop_mtp_state(self, "non-singleton-or-ineligible")
+            _log_multirow_mtp_inactive_once(self)
             _mark_standard_multirow_decode(self)
             return original_next(self, *args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


JimStenstrom:
Independent verification on an M5 Max 128GB: on a pristine v0.5.1 source build (50ee2464), one 2-concurrent burst permanently locks singleton MTP out (Qwen3.6-27B-oQ8-mtp, tg200 singles: 39.4 tok/s before, 18.0 after, zero `MTP path activated` lines; only an engine reload recovers). With this PR cherry-picked onto current main (5a39ba3a — applies clean), the same protocol recovers automatically: 39.0/38.9 tok/s on the next singles, with `MTP singleton recovery: multirow marker cleared (compact cache)` logged. Worth noting this also resolves the field reports in #2254 and the MTP-deactivation follow-up in #2235 — the multirow marker is what both are hitting.

jundot:
Thanks for the thorough work on this. The benchmark instrumentation proving the alignment gate never passes was convincing, and keeping row-wise off by default based on your own numbers was the right call. I traced the two core claims against the pinned mlx-lm and they hold up: BatchKVCache.filter() does shift out the min shared left padding, and per-row extract_cache seeding makes the alignment requirement unnecessary. The independent M5 Max verification above helped as well.

One small gap I'll pick up in a follow-up, nothing needed on your end: the compactness check reads left_padding off the cache object directly, so CacheList models (GLM 5.2 / DeepSeek v3.2 lineage) skip it entirely and the marker clears unverified. Still correct today since every current sub-cache filter compacts, the follow-up just makes the check recurse into .caches.

LGTM, merging.

jundot:
@gilby, I released oMLX 0.5.2.dev1 with this change. Please try the new build and let me know if you see any related issues: https://github.com/jundot/omlx/releases

Issue #2153, opened 2026-07-09
DFlash fails on Gemma 4 MLX models: model_type `gemma4_unified` not supported by Gemma4TargetOps


kmt9934:
## Summary
DFlash (speculative decoding) cannot be enabled for any current Gemma 4 MLX model. Every mlx-community / lmstudio Gemma 4 build (12B, 26B-A4B, 31B) uses \`model_type: gemma4_unified\`, but oMLX's DFlash target adapter only accepts \`gemma4\` / \`gemma4_text\`. As a result DFlash silently falls back to the standard engine and the 2–3× turbo is unavailable for the entire Gemma 4 family on Apple Silicon.

## Reproduction
- oMLX 0.4.4 (latest stable)
- Model: \`mlx-community/gemma-4-12B-it-4bit\` (also confirmed \`-6bit\`, \`-8bit\`, \`-qat-*\`, and \`lmstudio-community/gemma-4-12B-it-MLX-6bit\`)
- \`model_settings.json\`: \`"dflash_enabled": true, "dflash_draft_model": "z-lab/gemma4-12B-it-DFlash"\`

Server log on load:
\`\`\`
DFlash start failed for gemma-4-12B-it-4bit: Unsupported target architecture for DFlash target ops: model_type=gemma4_unified, model_class=Model, supported target backends: qwen_gdn, gemma4. Falling back to vlm engine.
\`\`\`

## Root cause
In \`dflash_mlx/engine/target_gemma4.py\`, \`Gemma4TargetOps.supports_model()\` returns True only when \`model_type in ("gemma4", "gemma4_text")\`. \`gemma4_unified\` is rejected.

From \`mlx_lm/models/gemma4.py\`, \`gemma4_unified\` is the "encoder-free vision" unified architecture — the text transformer is the same Gemma 4 SWA/full-attention stack, just packaged with vision. For text-only generation the compute path should be identical to \`gemma4\`, so the existing \`Gemma4TargetOps\` (SWA/full attention hooks, cache handling) should apply once \`gemma4_unified\` is accepted.

## Impact
- All current MLX Gemma 4 builds are \`gemma4_unified\`, so DFlash is effectively dead for Gemma 4 on oMLX today.
- The matching z-lab DFlash drafters already exist (\`z-lab/gemma4-12B-it-DFlash\`, \`z-lab/gemma-4-26B-A4B-it-DFlash\`, \`z-lab/gemma-4-31B-it-DFlash\`), so only the target-side adapter is missing.

[...]

Issue #49, opened 2026-02-27
[Feature request] Log API errors


zulufoxtrot:
**Is your feature request related to a problem? Please describe.**
It seems like API errors (400's, etc) are not logged, which makes debugging challenging.

**Describe the solution you'd like**
Log any API errors in the server log.

**Describe alternatives you've considered**

**Additional context**

jundot:
This has been addressed in commit ac905e0.

We added global exception handlers for `HTTPException`, `RequestValidationError`, and unhandled `Exception` so that all 4xx/5xx errors are now logged with the request method, path, status code, and error detail.

Previously, `raise HTTPException(...)` calls throughout the server had no accompanying `logger` call, and uvicorn's access log was disabled at the default log level. The new handlers catch these at the FastAPI level and log them as warnings (4xx) or errors (5xx) before returning the JSON response.

No configuration changes are needed. API errors will appear in the server log at the default INFO level.

PR #2175, opened 2026-07-10
fix: accept gemma4_unified in the DFlash compatibility gate (#2153)

richgoodson:
## Summary

All current mlx-community and lmstudio Gemma 4 MLX exports declare `model_type: gemma4_unified`, and DFlash refuses to start on every one of them (#2153). I reproduced this on v0.5.0 (`2130f14a`) with the reporter's model and traced it. There are three separate blockers:

1. `Gemma4TargetOps.supports_model` in dflash-mlx only accepts `gemma4` and `gemma4_text`. This is the failure in the reported log: the target model loads fine, `resolve_target_ops` raises, and engine_pool falls back to the vlm engine.
2. The z-lab Gemma 4 drafters are transformers 5.x exports that nest rope settings under `rope_parameters` with no top-level `rope_theta`, which `DFlashDraftModelArgs` requires. The draft bundle fails to load even with the gate opened.
3. omlx's own `is_dflash_compatible` also rejects `gemma4_unified`. This gate only feeds the dashboard compatibility display; it never blocks the runtime.

This PR fixes number 3 and adds a test. Numbers 1 and 2 are fixed in bstnxbt/dflash-mlx#49. Once that lands, a pin bump here makes the whole path work; I can fold the bump into this PR or send it separately, whichever is easier to review.

The change is safe because mlx-lm already loads `gemma4_unified` checkpoints through its gemma4 module (`MODEL_REMAPPING`), dropping the vision and audio towers in sanitize. The text stack DFlash drives is identical to the old `gemma4` exports; only the model_type string differs. The docstring keeps the `gemma4_assistant` reasoning intact.

### Live verification (M4 Pro 64GB, `mlx-community/gemma-4-12B-it-4bit` + `z-lab/gemma4-12B-it-DFlash`)

Before, unmodified v0.5.0:

```
DFlash start failed for gemma-4-12B-it-4bit: Unsupported target architecture for DFlash target ops: model_type=gemma4_unified, model_class=Model, supported target backends: qwen_gdn, gemma4. Falling back to vlm engine.
```

After, with the dflash-mlx fixes applied:

```
DFlash generation complete: 400 tokens, 40.1 tok/s, acceptance=65.2%, cycles=139
```

No AR fallback, coherent output. Wall clock went from 31.5 to 36.6 tok/s on a 400 token code prompt at temp 0. The uplift is smaller than the published dense 27B numbers because a 12B 4-bit target is already cheap per token; acceptance matches the old-format exports.

## Test plan

- [x] `pytest tests/test_dflash_engine.py -q`: 64 passed, includes the new `test_gemma4_unified_top_level_is_compatible` (mirrors the real unified config shape, including the nested `gemma4_unified_text`).
- [x] Full suite `pytest -q`: 6424 passed, 43 skipped.
- [x] Live repro and fix verification as above.

## Related

While setting this up I hit a footgun that can bite anyone upgrading a source install: `pip install -e .` does not update a git-pinned dependency when the pin moves to a new commit but the version number stays the same. My venv held mlx-vlm `086ab9d` while pyproject pins `78b96eb`, both reporting 0.6.3, and the only symptom was a confusing MiniMax test failure (the stale commit's in-tree `minimax_m3_vl` shadows the vendored compat modules). Source installs that upgraded across recent pin moves can be running new omlx code against old dependency commits without knowing. I will open a follow-up PR that warns at startup when an installed git pin drifts from pyproject.

Refs #2153

### DIFF
--- a/omlx/engine/dflash.py
+++ b/omlx/engine/dflash.py
@@ -57,6 +57,12 @@ def is_dflash_compatible(model_path: str | Path) -> tuple[bool, str]:
     Reading top-level only keeps the gate aligned with what dflash will
     actually load.
 
+    ``gemma4_unified`` (current mlx-community/lmstudio Gemma 4 exports,
+    #2153) is accepted too: mlx-lm's MODEL_REMAPPING loads those
+    checkpoints through the same ``gemma4`` module — the vision/audio
+    towers are dropped in sanitize and the text stack DFlash drives is
+    identical, which live-testing against z-lab's 12B drafter confirmed.
+
     Returns:
         (is_compatible, reason). ``reason`` is empty when compatible.
     """
@@ -72,7 +78,7 @@ def is_dflash_compatible(model_path: str | Path) -> tuple[bool, str]:
     model_type = str(cfg.get("model_type") or "").lower()
 
     is_qwen = "qwen" in model_type
-    is_gemma4 = model_type in ("gemma4", "gemma4_text")
+    is_gemma4 = model_type in ("gemma4", "gemma4_text", "gemma4_unified")
     is_laguna = model_type == "laguna"
     if not (is_qwen or is_gemma4 or is_laguna):
         return False, (
--- a/tests/test_dflash_engine.py
+++ b/tests/test_dflash_engine.py
@@ -806,6 +806,27 @@ class TestDFlashCompatibility:
         assert compatible is True
         assert reason == ""
 
+    def test_gemma4_unified_top_level_is_compatible(self, tmp_path):
+        """Current mlx-community Gemma 4 exports declare gemma4_unified at the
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for the thorough diagnosis, the three-blocker breakdown matches what I see in the code, and I confirmed the mlx-lm remapping claim on the current pin. One small correction: the gate is not display-only, the admin settings route returns a 400 when you try to enable dflash on an incompatible model, so this fix is what lets users turn it on at all.

Since bstnxbt/dflash-mlx#49 is still open and our current pin has neither of the other two fixes, merging this alone would make the dashboard advertise compatibility the runtime cannot deliver yet. Let's go with your option of folding the pin bump into this PR once #49 lands, and I'll merge this right after that.

PR #2178, opened 2026-07-10
fix scheduler: prefill progress hidden for HF cache models (Swift UI & web dashboard)

monroewilliams:
## Problem

HF Hub downloads models into `~/.cache/huggingface/hub/models--Org--Name/snapshots/<hash>/`, whereas oMLX-managed cache models live in a flat or two-level directory (e.g. `/Models/LLaMA-3.2-3B/`).

The scheduler received the full filesystem path as `model_name` and relied on `os.path.basename()` to derive the model_id. For HF cache models this returned the snapshot hash ("abc123") instead of the directory name ("Org--Name").

All 5 `os.path.basename(model_name)` calls in `scheduler.py` produced the snapshot hash, so tracker entries were stored under "abc123" while both the **Swift UI** and **web dashboard** queried prefill progress by the engine pool's model_id ("Org--Name"). Result: prefill progress bars were invisible for any HF cache model.


jundot:
Thanks for tracking this down, the basename-derived model_id was exactly why HF cache models never showed progress, and wiring the canonical id and model_path through the scheduler config is the right fix. There is a small conflict with the specprefill refactor that just landed (#2186), I'll resolve that on my side and merge this. One edge I noticed I'll handle in a follow-up: the DFlash fallback engine copies the shared scheduler config lazily, so it needs to snapshot model_name/model_path at construction time.

PR #2189, opened 2026-07-11
feat: optionally accelerate ModelScope downloads with aria2

### LINKED ISSUE #2188: Model downloads are slow and underutilize available bandwidth
## Problem

oMLX currently delegates model payload downloads to the standard Hugging Face and ModelScope download paths. On high-latency or bandwidth-rich connections, a single transfer path can leave most available bandwidth unused, making multi-gigabyte models unnecessarily slow to install.

This affects both supported model sources:

- Hugging Face
- ModelScope

## Reproduction

Model: `mlx-community/gemma-4-e2b-it-OptiQ-4bit`

Environment:

- macOS
- 12 repository files
- 5,258,980,718 bytes total (about 4.90 GiB)
- Hugging Face official endpoint
- cold destination directory

Observed standard downloader result before stopping the run:

- about 1.52 GB transferred in 332 seconds
- about 4.37 MiB/s average throughput

For comparison, an aria2 prototype using 8 connections per file and 4 concurrent files downloaded and verified the complete 5,258,980,718-byte repository in 262.738 seconds, averaging 19.09 MiB/s.

That is approximately 4.37x the measured throughput (+337%). The baseline was stopped early to avoid spending another ~14 minutes, so this comparison is based on average measured throughput rather than two completed full downloads.

## Proposed direction

Use aria2 for all model payload transfers while retaining the existing Hugging Face and ModelScope APIs for repository metadata and file manifests.

The downloader should also provide:

- configurable connections per file
- configurable concurrent files
- an optional aria2 HTTP proxy
- automatic application/network proxy bypass when a mirror endpoint is selected
- no attempt to bypass TUN/VPN routing
- clear installation status and a one-click Homebrew install path, with an explicit message when Homebrew is unavailable
- equivalent settings in the web dashboard and native macOS app

## Expected benefit

Faster, resumable model downloads with explicit transfer controls, while preserving the current model-source behavior and progress/cancellation lifecycle.

arthurlee116:
Closes #2188.

## Overview

This PR now keeps Hugging Face on the native Xet-enabled downloader added in `732a976f` and limits aria2 to an optional ModelScope acceleration path.

ModelScope downloads use aria2 when `aria2c` is available and fall back to the existing ModelScope SDK otherwise. DMG installs therefore keep working without Homebrew or any new runtime dependency.

### Optional ModelScope acceleration

- Added a short-lived `Aria2Runner` for ModelScope payload transfers.
- Missing aria2 selects the existing `modelscope.snapshot_download` path instead of returning an error.
- The web dashboard and native macOS app present aria2 as optional and only on the ModelScope download surface.
- One-click Homebrew installation remains available, but aria2 is not a Formula dependency.

### Correctness and cancellation

- Fetches the ModelScope file manifest with each file's immutable `Revision` and pins every payload URL to that revision.
- Downloads into a hidden `._____temp` staging directory and publishes files with `os.replace()` only after aria2 exits successfully.
- Cancelled and resumable partial files never appear under final model filenames, so model discovery cannot mistake them for a complete checkpoint.
- Progress uses allocated blocks (`st_blocks * 512`, capped by logical size) so sparse aria2 files do not jump to near-100% immediately.

### Process and network behavior

- Keeps authorization headers in a mode-0600 manifest rather than process arguments.
- Uses argv-based subprocess execution without a shell.
- Terminates the aria2 child on cancellation and redacts authorization values from surfaced errors.
- Supports bounded connections per file and concurrent-file controls.
- A configured ModelScope mirror strips application-layer proxy variables for the aria2 child; it does not attempt to bypass TUN/VPN routing.

### Hugging Face

- No Hugging Face downloader code is changed by this PR.
- Main's Xet path, chunk deduplication, and `abort_xet_session()` cancellation behavior remain intact.

## Review fixes

- Downloader tests explicitly select/mock the backend and no longer depend on the machine's real `PATH`.
- ModelScope payload URLs are revision-pinned.
- Partial files stay hidden until completion.
- Sparse-file progress reports allocated bytes.
- `docs/superpowers/specs` was removed from both the PR diff and commit history.
- The Homebrew aria2 hard dependency was removed.

## Validation

- `uv run --frozen pytest -q tests/ -m "not slow and not integration"`
  - `6862 passed, 52 skipped, 67 deselected`
- Focused downloader/UI/formula regression suite
  - `112 passed`
- `xcodebuild ... build`
  - passed
- `xcodebuild ... test`
  - passed
- Ruff checks for the new aria2 module and regression files
  - passed
- `git diff --check`
  - passed

## Reviewer guide

1. `omlx/admin/aria2_downloader.py` — staging, subprocess, proxy, cancellation, and redaction.
2. `omlx/admin/ms_downloader.py` — optional backend selection, pinned manifest, fallback, and progress.
3. `omlx/settings.py` and `omlx/admin/routes.py` — settings and install/status endpoints.
4. Web/native UI changes and regression tests.

### DIFF
--- a/Formula/omlx.rb
+++ b/Formula/omlx.rb
@@ -15,7 +15,6 @@ class Omlx < Formula
   depends_on arch: :arm64
   depends_on :macos
   depends_on "python@3.11"
-  depends_on "aria2"
 
   # macOS 27 beta's `strip` corrupts dynamic offsets in Mach-O libraries
   # (llvm/llvm-project#203678). Skip Homebrew's post-install clean pass over
--- a/Formula/omlx.rb
+++ b/Formula/omlx.rb
@@ -15,6 +15,7 @@ class Omlx < Formula
   depends_on arch: :arm64
   depends_on :macos
   depends_on "python@3.11"
+  depends_on "aria2"
 
   # macOS 27 beta's `strip` corrupts dynamic offsets in Mach-O libraries
   # (llvm/llvm-project#203678). Skip Homebrew's post-install clean pass over
--- a/apps/omlx-mac/Sources/AppView/Screens/DownloadsScreen.swift
+++ b/apps/omlx-mac/Sources/AppView/Screens/DownloadsScreen.swift
@@ -39,6 +39,19 @@ struct DownloadsScreen: View {
                 msAvailable: vm.msAvailable
             )
 
+            Aria2SettingsSection(
+                installed: vm.aria2Installed,
+                version: vm.aria2Version,
+                installing: vm.aria2Installing,
+                saving: vm.aria2Saving,
+                proxyDisabled: vm.aria2ProxyDisabled,
+                proxy: $vm.aria2Proxy,
+                connectionsPerFile: $vm.aria2ConnectionsPerFile,
+                concurrentFiles: $vm.aria2ConcurrentFiles,
+                onInstall: { vm.installAria2(client: services.client) },
+                onSave: { vm.saveAria2Settings(client: services.client) }
+            )
+
             if vm.source == .hf {
                 AddFromHFSection(
                     repoText: $vm.repoText,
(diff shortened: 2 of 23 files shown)


jundot:
Thanks for the detailed writeup and the real measurements. The diagnosis is right, the stock path is slow because omlx currently forces HF_HUB_DISABLE_XET for cancel correctness (#1322), so every payload file goes through a single stream.

For mainline I'm currently looking at fixing this by re-enabling xet instead of adding an external downloader. I retested the original blocker on huggingface_hub 1.21 + hf_xet 1.5.1 and the callback exception propagates now, cancel actually aborts the transfer, so the reason xet was disabled no longer holds. That would give parallel chunked transfers plus chunk dedup with no new runtime dependency.

The main reason I can't take aria2 as the only payload path is the hard dependency. The DMG app is the primary install channel and those users often don't have Homebrew, so model downloads would go from working to a 503. A few more things I hit while reviewing, in case you want to iterate:

- The downloader tests only pass when aria2c is installed on the machine. On a clean machine I get 32 failures in test_hf_downloader.py and test_ms_downloader.py from Aria2UnavailableError, configured_aria2_runner() probes the real PATH and needs mocking.
- Revision isn't pinned. The manifest comes from the default branch but the urls hardcode main (master for ModelScope), using model_info.sha would close that race.
- Partial weight files become visible under their final names. Previously shards were staged hidden and renamed in on completion, so a cancelled download can now look like a complete model to discovery, with a truncated safetensors inside.
- _get_dir_size uses st_size, and aria2 with --split plus --file-allocation=none writes sparse files, so reported progress jumps to near full right after start.
- docs/superpowers/specs looks like a workflow artifact and should be dropped.

arthurlee116:
@jundot Thanks for the review and for explaining the Xet direction. I rebased onto the latest main and updated the PR accordingly:

- Kept Hugging Face entirely on the new Xet + `abort_xet_session()` path; this PR no longer changes `hf_downloader.py`.
- Narrowed aria2 to optional ModelScope acceleration. When aria2 is unavailable, downloads fall back to `modelscope.snapshot_download`, and the Homebrew hard dependency has been removed.
- Made downloader tests explicitly select/mock the backend instead of probing the machine’s real PATH.
- Fetch the ModelScope file manifest with immutable per-file revisions and pin every payload URL accordingly.
- Stage downloads under `._____temp` and publish them with `os.replace()` only after aria2 completes successfully.
- Calculate progress from allocated blocks so sparse files no longer jump to near-complete immediately.
- Removed `docs/superpowers/specs` from both the PR diff and commit history.

Validation: 6862 passed, 52 skipped, 67 deselected; the macOS build and test actions both pass.

Issue #2137, opened 2026-07-09
GLM-5.2-mxfp4 PP drop from 160-180 tok/s to 40 tok/s with 0.5.0.dev2


Nvechenyu:
**Describe the bug**
- GLM-5.2-mxfp4 pp drop from 160-180 tok/s to 40 tok/s with 0.5.0.dev2. 100k context. M3 ultra 512G.
- Memory size fluctuates from 400G to 520G, was around 400G with 0.4.5.dev1.
- Thinking mode is off on default.
- Unable to test TG because out of memory

**Desktop:**
 - macOS Version: [26.5]
 - oMLX Version: [0.5.0.dev2]
 - Claude code: [2.1.204]

jundot:
@Nvechenyu 0.5.0.dev3 was just released. Could you please try the new build when you have a chance and check whether the GLM-5.2-mxfp4 PP slowdown and memory growth still reproduce? It includes MLX 0.32.0 support, DFlash draft discovery, and related compatibility fixes. https://github.com/jundot/omlx/releases

StartupBros:
We hit this on **v0.5.0 stable** with a different quant and can offer a root cause plus a 29x before/after confirmation.

**Setup:** M3 Ultra 512GB, macOS 26, `mlx-community/GLM-5.2-DQ4plus-q8` (converted from the non-FP8 base, so this is not mxfp4-specific), served via `omlx serve` from a source checkout (`pip install -e .`).

**Symptom:** prefill ~29 tok/s (9,649-token prompt took 336s). llama.cpp on the same box prefills the same prompt at ~92 tok/s with a comparable GGUF.

**Root cause on our install:** the native GLM DSA kernel extension was missing. `pip install -e .` (and any source install without `OMLX_WITH_CUSTOM_KERNEL=1` / `--with-custom-kernel`) skips building `omlx/custom_kernels/glm_moe_dsa/_ext`, and building it additionally requires the Metal toolchain, which plain Command Line Tools do not provide (`xcrun: unable to find utility "metal"`). When `_ext` is absent, every fast path in `omlx/patches/glm_moe_dsa/sparse_mla.py` returns None and the model silently falls back to generic MLX ops that materialize dense per-head indexer scores. That also explains the memory growth you saw (the fallback materializes `[B, H, L, K]` score tensors that the fused kernels never allocate).

[...]

PR #2191, opened 2026-07-11
feat(status): report native custom kernel availability in /api/status

StartupBros:
## Problem

Source installs only compile the optional native kernel extensions when built with `OMLX_WITH_CUSTOM_KERNEL=1`, and the build additionally requires the Metal toolchain (plain Command Line Tools fail with `xcrun: unable to find utility \"metal\"`). When the extensions are absent, every fast path guarded by `has_symbol(...)` silently returns to generic MLX fallbacks. For `glm_moe_dsa` that fallback materializes dense per-head indexer scores: we measured **~29 tok/s vs ~845 tok/s prefill** (29x) on the same box/model/prompt (M3 Ultra 512GB, GLM-5.2 DQ4plus-q8, 9.6k-token prompt), plus the memory growth described in #2137. Full analysis: https://github.com/jundot/omlx/issues/2137#issuecomment-4942102843

The only current signal is a patch-time `logger.warning` that is easy to miss, so users report the degraded mode as a performance regression.

## Change

- `GET /api/status` gains a `custom_kernels` block reporting, per native kernel package (`glm_moe_dsa`, `minimax_m3`, `qwen35_prefill`): `{available: bool, import_error: str|null}`. Implemented as `omlx.custom_kernels.native_kernel_status()`; never raises, so the status endpoint cannot break even on a wedged native module. Example on a source install without the kernels:

  ```json
  "custom_kernels": {
    "glm_moe_dsa": {"available": false, "import_error": "ImportError(\"cannot import name '_ext' ...\")"},
    ...
  }
  ```

- The existing GLM MoE DSA fallback warning now states the performance consequence, includes the underlying native import error, and names the build remediation.

## Testing

`tests/test_status_endpoint.py` extended with a `TestStatusCustomKernels` class: block shape/coverage, available-vs-import_error consistency, and a never-raises test with a simulated native import explosion. 11/11 pass run twice on an M3 Ultra: once with the native extensions absent and once with them present.

## Possible follow-up

Happy to also contribute a CI job that publishes wheels with prebuilt kernels so pip users get the accelerated path by default, if that is a direction you want.

### DIFF
--- a/omlx/custom_kernels/__init__.py
+++ b/omlx/custom_kernels/__init__.py
@@ -1,3 +1,36 @@
 # SPDX-License-Identifier: Apache-2.0
 """Optional native custom kernels bundled with oMLX."""
 
+from __future__ import annotations
+
+import importlib
+
+NATIVE_KERNEL_PACKAGES = ("glm_moe_dsa", "minimax_m3", "qwen35_prefill")
+
+
+def native_kernel_status() -> dict[str, dict[str, object]]:
+    """Report availability of every optional native kernel extension.
+
+    Source installs only compile the native extensions when built with
+    ``OMLX_WITH_CUSTOM_KERNEL=1`` (which additionally needs the Metal
+    toolchain). Without them the affected model families silently fall back
+    to much slower generic paths, so availability is surfaced through
+    ``GET /api/status`` for diagnosability instead of only a log line.
+
+    Never raises: a package that fails to import is reported as unavailable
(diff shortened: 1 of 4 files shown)


jundot:
Thanks, visibility here was overdue. Users kept reporting the fallback path as a perf regression (#2137) and the only signal was a log line. I verified the block on my box with the extensions present and with a simulated missing _ext, both report correctly. I'll merge this.

PR #2196, opened 2026-07-11
fix(reranker): fail fast on missing chat template, avoid duplicate think prefill

shyuan:
Follow-up to #2147 (merged): these two edge-case fixes were pushed to the PR branch shortly after the merge, so they missed the train.

Both were first handled by the alternative approach in #1964 and are absorbed here so the merged implementation covers them too:

- **Fail fast on missing chat template.** A tokenizer with `chat_template=None` now raises a clear `ValueError` up front instead of surfacing an opaque rendering failure from `apply_chat_template`.
- **No duplicate `<think>` prefill.** The standard-template path only appends the `<think>\n\n</think>\n\n` block when the rendered suffix doesn't already contain one, so templates that emit their own think prefill are no longer double-prefixed.

## Verification

- Two new tests in `TestCausalLMPromptAffixes` (`test_standard_template_think_prefill_not_duplicated`, `test_missing_chat_template_raises_clear_error`); `pytest tests/test_reranker_causal_lm.py`: 30 passed.
- Re-verified against the real tokenizers of `mlx-community/Qwen3-Reranker-0.6B-mxfp8` and `-4bit`: prefix/suffix extraction unchanged, token sequences still identical across both template paths.

generated by Claude Code

### DIFF
--- a/omlx/models/reranker.py
+++ b/omlx/models/reranker.py
@@ -330,6 +330,14 @@ class MLXRerankerModel:
            combined "<Instruct>/<Query>/<Document>" block — the exact content
            that _rerank_causal_lm reconstructs at scoring time.
         """
+        # Fail fast with a clear error when the tokenizer has no chat template
+        # at all — rendering would only produce an opaque downstream failure.
+        if hasattr(tokenizer, "chat_template") and tokenizer.chat_template is None:
+            raise ValueError(
+                f"Tokenizer for {self.model_name} has no chat template; "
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the quick follow-up, shame these two missed the train on #2147. I double checked on my side too, ran the affix extraction against the real tokenizers of the four Qwen3-Reranker variants I have locally (0.6B fp16, 4B mxfp8, 8B 4bit, 8B mxfp8) and the prefix/suffix output is byte-identical before and after this change, so no regression on the supported models. I'll merge this.

PR #2200, opened 2026-07-11
Fix stale MockEnginePool lease signature in streaming integration tests

### LINKED ISSUE #2199: Fix stale MockEnginePool lease signature in streaming integration tests
## Summary

The streaming integration tests use a `MockEnginePool` whose `get_engine()` signature no longer matches the server-to-pool contract. This makes the integration-marked suite fail before it can validate streaming behavior.

## Reproduction

```bash
uv run pytest tests/integration/test_e2e_streaming.py -m integration -q
```

Observed: all 12 streaming tests fail with:

```text
TypeError: MockEnginePool.get_engine() got an unexpected keyword argument `_lease`
```

Running the complete integration-marked selection also fails:

```bash
uv run pytest tests/ -m integration -q
# 12 failed, 4 skipped, 6549 deselected
```

## Root Cause

`omlx/server.py` acquires an engine through:

```python
engine = await pool.get_engine(model_id, _lease=True)
```

The test double in `tests/integration/test_e2e_streaming.py` still declares:

```python
async def get_engine(self, model_id: str):
```

The real pool supports the private lease argument, but the stale mock does not. Every streaming endpoint that obtains a lease therefore raises `TypeError` in the test harness.

Git history shows the lease argument was introduced in `omlx/server.py` by `d74a10614`, while this mock signature has been unchanged since `4cee64ca`.

## Expected Behavior

The mock should model the current pool interface so the streaming integration tests exercise their intended response-format assertions.

## Scope

Discovered during verification of #2186. That PR changes neither `omlx/server.py` nor `tests/integration/test_e2e_streaming.py`; this is a pre-existing test-suite defect, not a SpecPrefill refactor regression.

aosama:
## Summary

The streaming integration tests in `tests/integration/test_e2e_streaming.py` were broken by two stale test-double contracts that drifted from the real server-to-pool interface:

1. `MockEnginePool.get_engine(model_id)` rejected the `_lease` keyword that `omlx/server.py` began passing in `d74a10614`.
2. `MockBaseEngine` did not subclass `BaseEngine`, failing the `isinstance(engine, BaseEngine)` LLM type validation in `get_engine()` (`server.py:971`).
3. `MockEnginePool` lacked `release_engine()`, which the lease cleanup path calls on type-validation failure (`server.py:982`).

## Root Cause

The server acquired an eviction-proof lease (`_lease=True`) and validates the returned engine is a `BaseEngine` before releasing the lease. The test doubles predated both changes, so every streaming endpoint raised `TypeError` / `AttributeError` before reaching its assertions.

## Fix

Update only the test doubles in `tests/integration/test_e2e_streaming.py` to match the current contract:
- `MockBaseEngine` now subclasses `BaseEngine` and implements `start`/`stop`/`get_stats`/`get_cache_stats`.
- `MockEnginePool.get_engine` accepts `_lease` (matching the real pool signature).
- `MockEnginePool.release_engine` provides a no-op lease release.
- `MockEnginePool.get_entry` mirrors the real pool metadata accessor used by streaming endpoints.

No production code changed.

## Verification

- `uv run pytest tests/ -m "not slow" -q` → 6422 passed, 46 skipped, 2 failed
- `uv run pytest tests/ -m integration -q` → 12 passed, 4 skipped
- The 2 failures (`test_glm_mtp_patch.py::TestSmallLRouting`) are pre-existing numerical-tolerance failures unrelated to this change.

Closes #2199.

### DIFF
--- a/tests/integration/test_e2e_streaming.py
+++ b/tests/integration/test_e2e_streaming.py
@@ -12,6 +12,8 @@ from dataclasses import dataclass, field
 from typing import Any, AsyncIterator, Dict, List, Optional
 from unittest.mock import AsyncMock, MagicMock, patch
 
+from omlx.engine.base import BaseEngine
+
 
 @dataclass
 class MockGenerationOutput:
@@ -51,7 +53,7 @@ class MockTokenizer:
         return "\n".join(parts)
 
 
-class MockBaseEngine:
+class MockBaseEngine(BaseEngine):
     """Mock LLM engine with streaming support for testing."""
 
     def __init__(self, model_name: str = "test-model"):
@@ -76,6 +78,12 @@ class MockBaseEngine:
     @property
     def prefix_cache_enabled(self) -> bool:
         return False
+
+    async def start(self) -> None:
+        pass
+
+    async def stop(self) -> None:
+        pass
 
     def set_stream_outputs(self, outputs: List[MockGenerationOutput]):
         """Set custom streaming outputs for testing."""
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for catching this. I reproduced the 12 failures on main and confirmed the root cause, the lease argument came in with d74a1061 and the mock never followed. Verified your branch locally and all 12 streaming tests pass now. I'll merge this.

PR #2201, opened 2026-07-11
fix(turboquant): track the written physical end in batched KV caches

### LINKED ISSUE #2172: Intelligence regression in v0.5.0-rc1 oQ4e/oQ4 quantization
**Describe the bug**
oQ4e/oQ4 quantization in v0.5.0-rc1 causes significant intelligence regression compared to v0.5.0-dev4. GSM8K math reasoning drops by 37%, HumanEval code generation drops by 20%, and average intelligence drops by 12%. The quantization configuration is identical between versions, but test results differ significantly.

**To Reproduce**
1. Quantize Qwen3.5-9B model using oMLX v0.5.0-rc1
2. Use both oQ4e and oQ4 quantization methods
3. Run intelligence benchmarks (MMLU, TruthfulQA, GSM8K, MathQA, HumanEval)
4. Compare results with v0.5.0-dev4 and v0.4.4 versions

**Expected behavior**
Based on v0.5.0-dev4 and v0.4.4 test results, expected:
- GSM8K: ~86-90%
- HumanEval: ~83%
- Average intelligence: ~72-74%

**Actual behavior**
v0.5.0-rc1 test results:
- GSM8K: ~50-53% (**-37%**)
- HumanEval: ~63% (**-20%**)
- Average intelligence: ~61% (**-12%**)

**Screenshots**
Intelligence benchmark comparison:

| Model | MMLU | TruthfulQA | GSM8K | MathQA | HumanEval | Average |
|---|---|---|---|---|---|---|
| oQ4e-mtp-0.5.0-rc1 | 76.7% | 80.0% | **53.3%** ⚠️ | **30.0%** ⚠️ | **63.3%** ⚠️ | **60.7%** ⚠️ |
| oQ4-mtp-0.5.0-rc1 | 66.7% | 86.7% | **50.0%** ⚠️ | 40.0% | **63.3%** ⚠️ | **61.3%** ⚠️ |
| oQ4-mtp-0.5.0-dev4 | 66.7% | 86.7% | 90.0% ✅ | 40.0% | 83.3% ✅ | 73.3% ✅ |
| oQ4-mtp-0.3.9.dev1 | 70.0% | 86.7% | 90.0% ✅ | 40.0% | 83.3% ✅ | 74.0% ✅ |

Speed benchmark comparison:

| Model | tg TPS | TTFT(ms) | Peak Mem |
|---|---|---|---|
| oQ4e-mtp-0.5.0-rc1 | 41.8 tok/s | 1473 ms | 5.96 GB |
| oQ4-mtp-0.5.0-rc1 | 39.4 tok/
| oQ4-mtp-0.5.0-dev4 | 45.5 tok/s | 1467 ms | 5.96 GB |

Quantization config comparison (identical):

| Config | rc1 | dev4 | 0.4.4 |
|---|---|---|---|
| Base bits | 4-bit | 4-bit | 4-bit |
| 5-bit layers | 66 (97.1%) | 6
| 6-bit layers | 2 (2.9%) | 2 (2.9%) | 2 (2.9%) |
| Total special | 68 | 68 | 68

**Desktop (please complete the
 - macOS Version: macOS (Darwin 25.5.0)
 - oMLX Version: v0.5.0-rc1
 - Hardware: M5 MacBook Air 32GB
 - Model: Qwen3.5-9B
 - Quantization: oQ4e, oQ4 with MTP

**Additional context**
1. **Identical quantization cong files are identical betweenrc1 and dev4, ruling out configuration differences as the cause.

2. **Similar speed performance**: Speed performance is similar between rc1 and dev4,
suggesting the issue is accuraclated.

3. **Sample size note**: Intell samples per task due to localinference constraints. Results may have randomness. However, the magnitude of regression
(-37% on GSM8K, -20% on HumanEvsuggest a real issue rather than sampling noise.

4. **Possible causes**:
   - oQe calibration issue: imarectly weighted
   - Custom kernel regression: v0.5.0-rc1 custom kernels may have bugs
   - MTP weight handling: MTP wrupted in rc1
   - Calibration data: Calibration data used in rc1 may be different or corrupted

5. **Related files**:
   - Detailed config: `other/mt4-mtp-0.5.0-rc11-detail.json`
   - Sensitivity map:
`other/mtp-layout-weight/Qwen3.tivity.json`

6. **Recommendations**:
   - Check if oQe calibration logic has changed in rc1
   - Verify custom kernel behav
   - Compare calibration data between rc1 and dev4
   - Consider rollback or fix f

JimStenstrom:
Fixes #2172

## Problem

Serving any model with TurboQuant KV enabled degrades output quality severely under
concurrent load, independent of bit depth. Measured on Qwen3.6-35B-A3B oQ4 (GSM8K /
HumanEval, n=30, greedy deterministic profile, thinking off):

| Config | GSM8K | HumanEval |
|---|---|---|
| TQ-KV off, batch 4 | 96.7 | 90.0 |
| TQ-KV 8-bit, batch 1 | 96.7 | 90.0 |
| TQ-KV 8-bit, batch 4 | **53.3** | **70.0** |
| TQ-KV 8-bit, batch 4, MTP off | **53.3** | **70.0** |

This matches the #2172 report (their GSM8K also lands at ~50–53, 8-bit as bad as
4-bit). MTP is not required — the reporter's isolation happened to toggle KV together
with MTP, but the collapse reproduces identically with MTP off, and it is invisible at
batch 1 (which is why it resists single-request reproduction).

## Root cause

`BatchTurboQuantKVCache` has no record of its written physical end. The B>1
`update_and_fetch` derived the append position from `offset.max()`, and `make_mask`
derived its column count the same way. That equals the written end only while some row
has zero left padding. The moment `filter()` removes the last zero-padding row — a
routine continuous-batching event whenever the longest-prompt request finishes first —
every logical offset falls short of the written end, and each subsequent decode append
lands `min(left_padding)` columns early: it **overwrites the surviving rows' live KV in
place**, the appended tokens are silently lost (the buffer never grows), and the mask
shrinks below the survivors' own context. Every surviving request in the batch is
corrupted from that point on; the damage compounds with each departure.

Deriving the position from `_state_length()` instead would also be wrong: that is the
step-allocated capacity, not the written end, so after any buffer growth it would leak
unwritten columns into `.state` and attention geometry (and `extend()` already had this
capacity-vs-written confusion when aligning a joining batch).

Unit repro (now a regression test): merge rows of 48/32 tokens → filter out the
zero-padding row → append 3 tokens → buffer stuck at 48, 3 tokens lost, the survivor's
logical positions 16–18 overwritten (content error 4.76 vs 0.01 quantization noise).
The identical sequence through dense `BatchKVCache` appends correctly.

## Fix

Track an explicit written-column count (`_phys_end`): maintained by
`update_and_fetch`/`merge`/`extend`, synced at the B=1→B>1 transition
(`_ensure_array_offset`) and by the state setter for restored states, and used for the
append position, the `.state` slice, `make_mask` geometry (including the
single-survivor-with-left-padding case, which previously built an unpadded mask from
the logical offset), and `extend` alignment. Deliberately not named `_idx`:
`rollback_speculative_cache` in mlx-vlm changes behavior on that attribute, and this
change keeps that path untouched.

Behavior is bit-identical while the zero-left-padding invariant holds (`_phys_end ==
offset.max()` in that regime), so unbatched and pre-departure serving are unaffected.

## Validation

- 3 new regression tests (RED-proven on the unfixed code): append position after the
  min-left-padding row departs (content vs dense mirror), `make_mask` width after
  departure, and append-past-capacity growth keeping content and geometry.
- `tests/test_turboquant.py`: 43/43. All test files importing the module (scheduler,
  paged SSD cache, MTP patches, VLM MTP, benchmark, integration): 740 passed.
- Model-level re-run of the two degraded cells on this branch: **both recover to
  96.7 / 90.0** — identical to the TQ-off control.

### DIFF
--- a/omlx/turboquant_kv.py
+++ b/omlx/turboquant_kv.py
@@ -246,6 +246,16 @@ class BatchTurboQuantKVCache(TurboQuantKVCache):
         else:
             self.offset = -left_padding[0]
         self._right_padding = None
+        # Written physical column count (B>1 only; B=1 uses the parent's int
+        # offset). Rows are end-aligned, but this must NOT be derived from
+        # offset.max(): once filter() removes the last zero-left-padding row,
+        # every logical offset is short of the written end, and an
+        # offset-derived position writes INSIDE the survivors' live KV. It
+        # also must not be derived from _state_length(self.keys): that is the
+        # step-allocated capacity, not the written end. (Deliberately not
+        # named `_idx` — mlx-vlm's rollback_speculative_cache changes
+        # behavior on that attribute.)
+        self._phys_end = 0
 
     # ---- update_and_fetch override for B>1 only ----------------------------
 
@@ -255,13 +265,14 @@ class BatchTurboQuantKVCache(TurboQuantKVCache):
             return super().update_and_fetch(keys, values)
         # B>1: track per-request offset separately from state offset
         T_new = keys.shape[2]
-        # Use int offset for state management
-        int_offset = self.offset.max().item()
+        # Append at the written physical end (see __init__._phys_end note).
+        int_offset = self._phys_end
         self.offset += T_new
         saved_offset = self.offset
         self.offset = int_offset
         result = super().update_and_fetch(keys, values)
         self.offset = saved_offset
+        self._phys_end = int_offset + T_new
         return result
 
     # ---- state override for B>1 (offset is mx.array) -----------------------
@@ -270,15 +281,22 @@ class BatchTurboQuantKVCache(TurboQuantKVCache):
     def state(self):
         if isinstance(self.offset, int):
             return super().state
-        # B>1: use keys length directly (offset is mx.array, can't compare with int)
+        # B>1: slice to the written end — _state_length(self.keys) is the
+        # step-allocated capacity and would expose unwritten columns to
+        # attention after the buffer grows.
         if self.keys is None:
             return None, None
-        length = _state_length(self.keys)
+        length = self._phys_end
         return _slice_state(self.keys, length), _slice_state(self.values, length)
 
     @state.setter
     def state(self, value):
         TurboQuantKVCache.state.fset(self, value)
+        # Restored states (SSD cache, extract/merge round-trips) are packed
+        # at their exact written width — sync the physical end for the B>1
+        # bookkeeping paths.
+        if self.keys is not None and not isinstance(self.offset, int):
+            self._phys_end = _state_length(self.keys)
 
     # ---- make_mask override (batch-aware) ----------------------------------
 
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, great find and a really clean fix. I verified it on my side: the regression tests fail on main as described, and a live server run with a long request leaving a TQ-KV batch corrupted the remaining requests on main while this branch returned identical output to single-request runs. All turboquant and related tests pass here too. I'll merge this.

PR #2205, opened 2026-07-11
fix(oq): scope VisionConfig model_type normalization to Qwen types

a4501150:
## Problem

HuggingFace transformers appends `_vision` to vision sub-config `model_type` (e.g. `qwen3_5_moe_vision`), but mlx-vlm's Qwen `VisionModel.__init__` only accepts the base name. The previous fix stripped `_vision` unconditionally, which broke other families (gemma3n, minicpmv4_6) that require the suffix.

## Fix

Scope the `_vision` suffix normalization to Qwen model types only (`omlx/oq.py` and `omlx/utils/model_loading.py`).

## Test plan

- [x] Qwen3.6 VLM loads correctly (`vision_config.model_type` normalized from `qwen3_5_moe_vision` → `qwen3_5_moe`)
- [x] oQ sensitivity measurement succeeds on Qwen VLMs
- [x] Non-Qwen VLMs (gemma3n, minicpmv4_6) are unaffected — their `_vision` suffix is preserved

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -2658,6 +2658,12 @@ def _build_model_sanitizer(config: dict, text_only: bool = False):
             vision_config = model_config.vision_config
             if isinstance(vision_config, dict):
                 vision_config = model_module.VisionConfig.from_dict(vision_config)
+            if hasattr(vision_config, "model_type") and isinstance(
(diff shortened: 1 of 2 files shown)


a4501150:

PR #2206, opened 2026-07-11
fix(oqe): collect lm_head imatrix and skip embedding lookup

a4501150:
## Problem

The oQe imatrix calibration loop only forwarded through transformer layers, so `lm_head` never got activation statistics — causing "oQe imatrix missing entry for language_model.lm_head" errors on VLMs.

## Fix

Three changes in `omlx/oq.py`:

- `_collect_lm_head_imatrix()` — forwards final hidden states through norm → lm_head after the layer loop so its capture wrapper fires
- `_oqe_cache_missing_lm_head()` — detects stale caches missing the lm_head entry and triggers automatic recollection
- Skip imatrix lookup for embedding layers (`embed_tokens`, `wte`, etc.) since they take discrete token IDs, not continuous activations

## Test plan

- [x] Stale imatrix cache (pre-lm_head) is detected and recollected automatically
- [x] Fresh imatrix cache contains `language_model.lm_head` entry
- [x] `pytest -k "oq or imatrix"` passes

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -3663,6 +3663,15 @@ def _lookup_imatrix_importance(
         return None
 
     base = tensor_name[: -len(".weight")]
+
+    # Embedding layers take discrete token IDs, not continuous activations —
+    # activation-energy imatrix does not apply.  Skip silently.
+    if any(
+        p in base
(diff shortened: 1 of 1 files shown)


a4501150:

PR #2207, opened 2026-07-11
feat(oq): add preserve_gdn to keep GDN/linear_attn layers at fp16

a4501150:
## Problem

GatedDeltaNet / linear_attn layers maintain a running recurrence state that accumulates quantization noise over sequence length, collapsing output quality. Unlike stateless attention layers that tolerate 4-6 bit quantization well, GDN projections need full precision.

## Fix

Add a `preserve_gdn` flag (mirroring `preserve_mtp`) that keeps all `linear_attn.*` weights at fp16/bf16 during oQ quantization. Output name gets a `-gdn` suffix (e.g. `oQ6e-gdn`).

## Changes

- `omlx/oq.py` — `preserve_gdn` parameter, `linear_attn.*` weight exclusion
- `omlx/admin/oq_manager.py` — pass through to quantizer
- `omlx/admin/routes.py` — GDN layer detection for UI auto-enable
- `omlx/admin/static/js/dashboard.js` — toggle state management
- `omlx/admin/templates/dashboard/_models.html` — UI toggle
- `omlx/utils/model_loading.py` — model loading integration
- `omlx/admin/i18n/*.json` — new keys across 9 locales

## Test plan

- [x] Non-GDN models unaffected (toggle auto-disabled in UI)
- [x] Output name includes `-gdn` suffix when enabled
- [x] `pytest -k "oq or imatrix"` passes


a4501150:

jundot:
Thanks for the PR, the implementation is clean. But I checked the actual weight breakdown on Qwen3.6-35B-A3B and I'm not sure full bf16 preservation is the right tradeoff yet. linear_attn is only 2.8% of total params, but since this is an A3B MoE it's about 31% of the active weights per decode token, so keeping it at bf16 raises bytes read per token by roughly 1.5-1.8x at oQ6/oQ4. That's a real decode speed cost.

Also note main already protects the sensitive GDN params (dt_bias and A_log stay fp, conv1d gets 8 bit, out_proj gets a 5 bit floor), and the oQe imatrix boost map can raise in_proj bits when measured sensitivity actually calls for it. So the delta here is mostly in_proj_qkv/z going from base bits to bf16.

Could you share before/after numbers on a GDN model, ppl or any benchmark plus decode tok/s? If the quality win is real I'm glad to move forward with this, there are just two small things to fix then (/api/oq/estimate ignores preserve_gdn, and _oq_preserve_gdn leaks into the output config.json).

a4501150:
Benchmark results on Ornith-1.0-35B-AEON-Ultimate-Uncensored (Qwen3.5-MoE, A3B active), M5 Max 128GB:

| | oQ8e | oQ8e-gdn | Delta |
|---|---|---|---|
| PPL (512 tok) | 1.43 | 1.40 | **-0.03** (lower = better) |
| NLL | 0.3555 | 0.3340 | -0.0215 |
| Prompt tok/s | 1852 | 1942 | +4.8% |
| Decode tok/s | 74.2 | 73.7 | -0.8% |
| Peak GPU mem | 34.76 GB | 35.66 GB | +0.9 GB |
| Disk size | 35 GB | 36 GB | +1 GB |

Methodology: single-sequence perplexity on a 512-token wikitext-style passage via direct model forward pass, decode speed via `mlx_lm.generate` with 381-token prompt and 128 output tokens.

The quality gain is modest but consistent (-6% NLL), and the decode speed hit is within noise (-0.8%). Memory overhead is ~0.9 GB from the bf16 linear_attn weights.

For lower quant levels (oQ4/oQ6) the tradeoff should be more pronounced on both axes — GDN quantization noise compounds with the recurrence at longer sequences, and the memory delta from bf16 vs 4-6 bit is larger. Happy to run those if useful.

Also fixed the two issues you flagged:
- `/api/oq/estimate` now accepts and passes through `preserve_gdn`
- `_oq_preserve_gdn` is stripped from the output config.json in both the streaming and non-streaming write paths

PR #2209, opened 2026-07-12
docs: clarify native custom kernel build requirements and the cost of skipping them

StartupBros:
Small docs follow-up to #2137 / #2191: the install section lists `OMLX_WITH_CUSTOM_KERNEL=1` as optional without saying (a) what is lost without it, for GLM-5.2 the fused DSA prefill is ~30x faster with the kernels (measured 845 vs ~29 tok/s on an M3 Ultra 512GB) and the fallback grows memory, or (b) that building them needs the Metal toolchain, which plain Command Line Tools do not provide (`xcrun: error: unable to find utility \"metal\"`). This tripped us and, we suspect, the #2137 reporter. Adds a short note with the requirement, points DMG/Homebrew users at the precompiled kernels, and includes a one-line verification command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,11 +86,25 @@ cd omlx
 pip install -e .          # Core only
 pip install -e ".[mcp]"   # With MCP (Model Context Protocol) support
 
-# Optional: GLM-5.2 / MiniMax M3 native custom kernels
+# GLM-5.2 / MiniMax M3 / Qwen3.5 native custom kernels (strongly recommended
+# if you serve those families -- see note below)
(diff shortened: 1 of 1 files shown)


jundot:
Thanks, this gap was real and #2137 proved it. One correction I'll make in a follow-up: Homebrew does not ship the kernels precompiled, the formula only builds them from source with --with-custom-kernel and that build also needs full Xcode, so only the DMG has them prebuilt. I'll also point the verification one-liner at native_kernel_status() from #2191 so it covers all three kernel packages. I'll merge this.

Issue #2208, opened 2026-07-11
GLM-5.2 cold prefill: adaptive prefill throttle near the memory ceiling (was: prefix-cache kernel bypass, retracted)


StartupBros:
**Summary:** the GLM DSA fused prefill kernels (from #1984) are bypassed whenever the prefix cache is enabled (`--paged-ssd-cache-dir` or `--hot-cache-max-size`), so users must choose between fast cold prefill and prefix reuse. On agentic multi-turn workloads (growing context, many calls per task) neither mode is competitive, while the same weights under a runtime that combines both are.

**Setup:** M3 Ultra 512GB, macOS 26, oMLX v0.5.0 with native custom kernels confirmed loaded (`fast.native_available() == True`; see #2137 / #2191 for the kernel-availability context), `mlx-community/GLM-5.2-DQ4plus-q8`, single stream.

**Measurements (identical 9,649-token novel prompt, identical follow-up-turn pattern):**

| serve flags | cold prefill | follow-up turn (97% prefix overlap) |
|---|---|---|
| (none) | **~845 tok/s** (13.2s) | full re-prefill (~13s+ per turn, no reuse) |
| `--hot-cache-max-size 40GB` | ~29 tok/s (338s) | 14.2s, `cached_tokens: 9472` |
| `--paged-ssd-cache-dir ...` | ~27 tok/s (363s) | 10.9s, `cached_tokens: 5632` |

Cold prefill with any cache enabled lands at the same ~27-29 tok/s as the no-native-kernels fallback (#2137), which suggests the cached/paged executor runs the unfused path and never dispatches `dsa_indexer_scores` / `glm_dsa_sparse_mla_attention`.

**Why it matters:** on a benchmark of agentic repair tasks (10-30 chat calls per task over a growing 10-65k context, 30-min per-task SLA), GLM-5.2 under oMLX misses the SLA in both modes: cache-less pays full prefill every call; cached pays 30-40 minutes of cold prefill on the big contexts. The same quant's correctness was perfect in those runs (25/25 completed tasks passed), so serving throughput is the only blocker. llama.cpp on the same box does 92 tok/s prefill WITH prompt caching simultaneously and fits the SLA.

[...]

PR #2211, opened 2026-07-12
ci: build and publish wheels with native custom kernels precompiled

StartupBros:
Follow-up to #2137 / #2191 / #2208, and the wheel-CI offer from #2191: most pip/source users never build the native custom kernels (needs `OMLX_WITH_CUSTOM_KERNEL=1` plus the Metal toolchain that Command Line Tools lack), so they silently run the generic fallback, for GLM-5.2 that is ~30x slower prefill. This workflow gives pip users the same precompiled kernels the DMG ships:

- builds cp311/cp312/cp313 arm64 wheels on `macos-15` (the kernels target `MACOSX_DEPLOYMENT_TARGET=15.0`, so the import smoke needs a Sequoia host) with `OMLX_WITH_CUSTOM_KERNEL=1`
- `--no-isolation` with preinstalled pinned build deps (`nanobind==2.13.0`, `mlx==0.32.0`, `cmake>=3.27`) and an explicit `-DPython_EXECUTABLE` pin, CMake's Python discovery does not reliably resolve pip's isolated build env on hosted runners
- gates every wheel on two checks: the native artifacts (`_ext`, `.dylib`, `.metallib`) are present in the wheel, and a fresh-venv install (run from outside the checkout) reports `is_native_available() == True` for all three kernel packages
- uploads artifacts on `workflow_dispatch`; attaches wheels to the release on publish (same trigger shape as `update-formula.yml`)

**Validated green on my fork's Actions across all three Python versions:** https://github.com/StartupBros/omlx/actions/runs/29173718204

Action pins follow the repo convention (full SHAs). Happy to adjust triggers/naming, e.g. if you would rather gate this behind a tag pattern or fold it into the release pipeline that builds the DMG.

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -51,8 +51,13 @@ jobs:
 
       - name: Smoke-test wheel (native kernels must import and report available)
         run: |
+          echo "--- native artifacts in the wheel:"
+          unzip -l dist/*.whl | grep -E "_ext|dylib|metallib" || { echo "no native artifacts in wheel"; exit 1; }
           python -m venv /tmp/wheel-smoke
           /tmp/wheel-smoke/bin/pip install --quiet dist/*.whl
+          # run from outside the checkout so `import omlx` resolves to the
+          # installed wheel, not the source tree in cwd
+          cd /tmp
           /tmp/wheel-smoke/bin/python - <<'PY'
           import importlib
           failed = []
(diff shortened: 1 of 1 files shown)


jundot:
Thanks, this closes the gap properly instead of just documenting it. I checked the build deps against the pyproject pins (nanobind and mlx match, and runtime mlx is pinned to the same version so no ABI drift for wheel users), the matrix covers the whole supported Python range, and every action SHA resolves to the tag it claims. The trigger shape is fine as is. I'll merge this.

PR #2220, opened 2026-07-12
fix: load indexed MTP sidecar safetensors skipped by mlx-lm's glob

### LINKED ISSUE #2062: Native MTP load skips indexed mtp.safetensors sidecar when filename does not match model*.safetensors
**Describe the bug**

Enabling Native MTP for `wang-yang/Ornith-1.0-35B-MTPLX` fails even though the checkpoint includes MTP weights and the safetensors index references them.

The model has `text_config.mtp_num_hidden_layers: 1`, and `model.safetensors.index.json` maps 785 `mtp.*` tensors to `mtp.safetensors`. However, the runtime load path appears to only load files matching `model*.safetensors`, so the indexed sidecar `mtp.safetensors` is skipped. The MTP patch then reports that MTP tensors are missing.

This looks related to #1944, but slightly different: #1944 covers MTP sidecars that are not listed in `model.safetensors.index.json`; this case is listed in the index but still skipped during load because the sidecar filename does not match the loader glob.

**To Reproduce**

1. Download `wang-yang/Ornith-1.0-35B-MTPLX`.
2. Confirm the model directory contains:
   - `model-00001-of-00004.safetensors`
   - `model-00002-of-00004.safetensors`
   - `model-00003-of-00004.safetensors`
   - `model-00004-of-00004.safetensors`
   - `mtp.safetensors`
   - `model.safetensors.index.json`
3. Confirm the index maps MTP tensors to the sidecar:

   ```bash
   jq -r '.weight_map | to_entries[] | select(.key|startswith("mtp.")) | .value' \
     model.safetensors.index.json | sort | uniq -c
   ```

   Locally this showed:

   ```bash
   785 mtp.safetensors
   ```

4. In the oMLX admin UI, enable Native MTP for `Ornith-1.0-35B-MTPLX`.
5. Start a chat completion against the model.
6. See a 500 error.

Observed log excerpt:

> `Native MTP patch applied for .../Ornith-1.0-35B-MTPLX (model_type=qwen3_5_moe, active)`

followed by:

> `Native MTP is enabled for this model but the converted weights are missing the mtp.* tensors.`

**Expected behavior**

oMLX should load MTP tensors referenced by `model.safetensors.index.json`, even when they live in a sidecar file named `mtp.safetensors` rather than a `model*.safetensors` shard.

Native MTP should activate successfully for this model.

**Screenshots**

Not applicable. This is reproducible from model files and server logs.

- macOS Version: 26.5.1
- oMLX Version: 0.4.5.dev1
- mlx-lm Version: 0.31.3
- mlx Version: 0.31.2
- mlx-vlm Version: 0.6.3

**Additional context**

Model config details:

- `model_type`: `qwen3_5_moe`
- `architectures`: `["Qwen3_5MoeForConditionalGeneration"]`
- `text_config.model_type`: `qwen3_5_moe_text`
- `text_config.mtp_num_hidden_layers`: `1`

A local workaround confirmed the root cause: adding a symlink named `model-mtp.safetensors` pointing to `mtp.safetensors` makes the existing `model*.safetensors` glob include the MTP sidecar.

```bash
ln -s mtp.safetensors model-mtp.safetensors
```

After that workaround, the same model loaded and Native MTP activated successfully:

> `MTP path activated for uid=0 (model has mtp_forward, batch=1)`

and generation logged:

> `MTP[0] finish=length tokens=12 cycles=6 accept=4/6 (66.7%)`

Suggested fix: when loading Native MTP-capable models, include safetensors sidecar files referenced by `model.safetensors.index.json` for keys like `mtp.*`, `language_model.mtp.*`, `model.mtp.*`, or `model.language_model.mtp.*`, instead of relying only on `model*.safetensors`.

I also have a local patch with a focused test that makes the loader glob include indexed MTP sidecars automatically.

mxl:
## Problem

`mlx_lm.utils.load_model` discovers weight shards with `glob.glob(model*.safetensors)`. MTPLX-forge exports (e.g. `wang-yang/Ornith-1.0-35B-MTPLX`) ship the MTP head's weights in a separate `mtp.safetensors` sidecar. `model.safetensors.index.json` correctly references it (785 `mtp.*` keys → `mtp.safetensors`), but the filename doesn't match the loader's glob pattern, so the file is silently skipped. The patched `TextModel.sanitize` then sees an attached MTP head with zero `mtp.*` keys in the weight dict and raises:

> Lightning MTP is enabled for this model but the converted weights are missing the mtp.* tensors.

...even though the checkpoint has them.

Fixes #2062.

## Fix

`omlx/patches/mlx_lm_extra_tensors.py`:
- `sidecar_files_for(model_path)` scans `model.safetensors.index.json`'s `weight_map` for `mtp.*`-prefixed keys (matching the same prefix family used elsewhere for MTP detection: bare `mtp.`, `language_model.mtp.`, `model.mtp.`, `model.language_model.mtp.`) and resolves their backing file(s), excluding anything mlx-lm's own `model*.safetensors` glob would already find.
- A `_GlobProxy` installed as `mlx_lm.utils.glob` passes every glob call through to the real `glob` module unchanged, except the `model*.safetensors` shard-discovery pattern, which gets the sidecar files appended.
- Process-wide file list (`set_extra_tensor_files` / `get_extra_tensor_files`) follows the same construction-time-flag pattern already used by `mlx_lm_mtp.set_mtp_active` — reset to `[]` at the top of `maybe_apply_pre_load_patches` so a model without a sidecar doesn't inherit a prior load's file list.

Wired into `maybe_apply_pre_load_patches` (`omlx/utils/model_loading.py`), gated on `_has_mtp_heads(config)` rather than `mtp_enabled`, matching the existing rationale for the Native MTP patch dispatch right below it: sanitize correctness must hold regardless of whether the head is attached.

## Not in scope

#1944 / PR #1962 fix a different bug: the Native MTP *compatibility check* (`_checkpoint_has_mtp_weights` / `_model_has_mtp_weight_tensors`) false-negatives for OptiQ checkpoints whose sidecar isn't referenced by the index at all (uses a separate `mtp_file` config key instead). Neither of those PRs touches actual weight loading, so they wouldn't have fixed this issue even if merged.

## Testing

New `tests/test_mlx_lm_extra_tensors.py` (24 tests): index scanning (bare/prefixed mtp keys, already-covered shards, malformed/missing index, dedup+sort), the process-wide flag, `_GlobProxy` passthrough/augmentation/dedup, `apply()` idempotency and import-failure handling, and integration coverage of the `maybe_apply_pre_load_patches` dispatch (including the reset-between-loads case).

```
uv run --python 3.11 pytest tests/test_mlx_lm_extra_tensors.py tests/test_mlx_lm_mtp_patch.py tests/test_model_loading.py -q
# 153 passed
uv run --python 3.11 ruff check omlx/patches/mlx_lm_extra_tensors.py omlx/utils/model_loading.py tests/test_mlx_lm_extra_tensors.py --isolated
# All checks passed!
```

### DIFF
--- a/omlx/patches/deepseek_v4/utils_patch.py
+++ b/omlx/patches/deepseek_v4/utils_patch.py
@@ -145,7 +145,7 @@ def _build_patched_load_model() -> Callable:
                 "trust this model."
             )
 
-        weight_files = glob.glob(str(model_path / "model*.safetensors"))
+        weight_files = _utils.glob.glob(str(model_path / "model*.safetensors"))
 
         if not weight_files and strict:
             raise FileNotFoundError(f"No safetensors found in {model_path}")
--- a/omlx/patches/deepseek_v4/utils_patch.py
+++ b/omlx/patches/deepseek_v4/utils_patch.py
@@ -20,7 +20,6 @@ When mlx-lm merges PR 1192 upstream this patch should be removed.
 
 from __future__ import annotations
 
-import glob
 import importlib.util
 import json
 import logging
--- a/omlx/patches/mlx_lm_extra_tensors.py
+++ b/omlx/patches/mlx_lm_extra_tensors.py
@@ -1,0 +1,133 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Patch mlx-lm's weight-shard glob to pick up indexed MTP sidecar files.
+
+``mlx_lm.utils.load_model`` discovers weight shards with
+``glob.glob(str(model_path / "model*.safetensors"))``. MTPLX-forge exports
+(e.g. ``wang-yang/Ornith-1.0-35B-MTPLX``) ship the MTP head's weights in a
+separate ``mtp.safetensors`` sidecar that ``model.safetensors.index.json``
+references but that filename doesn't match the glob. The sidecar is
+silently skipped, so the ``mtp.*`` keys it holds never reach
+``TextModel.sanitize`` — Native MTP then raises "the converted weights are
+missing the mtp.* tensors" even though the checkpoint has them (issue
+#2062).
+
+Fix: replace the ``glob`` name ``mlx_lm.utils`` resolves at call time with a
+thin proxy. Every pattern except the ``model*.safetensors`` shard-discovery
+call passes straight through to the real ``glob.glob``; that one call gets
+augmented with sidecar files referenced by ``mtp.*``-prefixed keys in the
+model's safetensors index.
+
+This is distinct from issue #1944 / PR #1962, which fix the Native MTP
+*compatibility check* (``_checkpoint_has_mtp_weights`` /
+``_model_has_mtp_weight_tensors``) for OptiQ checkpoints whose sidecar isn't
+referenced by the index at all. Neither of those touches actual weight
+loading, so they don't help here.
+"""
+
+from __future__ import annotations
+
+import glob as _glob_module
+import json
+import logging
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+_EXTRA_TENSOR_FILES: list[str] = []
+
+_MTP_INDEX_PREFIXES = (
+    "mtp.",
+    "language_model.mtp.",
+    "model.mtp.",
+    "model.language_model.mtp.",
+)
+
+
+def set_extra_tensor_files(files: list[str]) -> None:
+    """Set the process-wide sidecar file list for the next ``mlx_lm.load()``.
+
+    Same construction-time-flag pattern as ``mlx_lm_mtp.set_mtp_active``:
+    the caller resets this to ``[]`` before every load so a model without a
+    sidecar isn't polluted by a prior load's file list.
+    """
+    global _EXTRA_TENSOR_FILES
+    _EXTRA_TENSOR_FILES = list(files)
+
+
+def get_extra_tensor_files() -> list[str]:
+    return list(_EXTRA_TENSOR_FILES)
+
+
+def sidecar_files_for(model_path: str | Path) -> list[str]:
+    """Return absolute paths of safetensors sidecars holding mtp.* weights.
+
+    Reads ``model.safetensors.index.json``'s ``weight_map``, collects the
+    distinct filenames backing ``mtp.*``-prefixed keys, and drops any that
+    already match mlx-lm's own ``model*.safetensors`` glob (no need to add
+    those — mlx-lm finds them on its own).
+
+    Returns ``[]`` when there's no index, the index is unreadable, or every
+    mtp.* key already lives in a ``model*.safetensors`` shard.
+    """
+    p = Path(model_path)
+    index_path = p / "model.safetensors.index.json"
+    if not index_path.exists():
+        return []
+    try:
+        weight_map = json.loads(index_path.read_text()).get("weight_map") or {}
+    except Exception as e:
+        logger.debug("Failed to read %s for sidecar scan: %s", index_path, e)
+        return []
+
+    sidecar_names = {
+        fname
+        for key, fname in weight_map.items()
+        if isinstance(key, str)
+        and key.startswith(_MTP_INDEX_PREFIXES)
+        and isinstance(fname, str)
+    }
+    if not sidecar_names:
+        return []
+
+    already_covered = {
+        Path(f).name for f in _glob_module.glob(str(p / "model*.safetensors"))
+    }
+    return sorted(
+        str(p / fname) for fname in sidecar_names if fname not in already_covered
+    )
+
+
+class _GlobProxy:
+    """Delegates to the real ``glob`` module, augmenting one call site.
+
+    Only the ``model*.safetensors`` shard-discovery pattern is
+    special-cased; every other call (tokenizer/config discovery, checkpoint
+    saving, sharded-pipeline file selection, etc.) passes through
+    untouched.
+    """
+
+    _omlx_extra_tensors_proxy = True
+
+    def glob(self, pattern, *args, **kwargs):
+        matches = _glob_module.glob(pattern, *args, **kwargs)
+        if pattern.endswith("model*.safetensors") and _EXTRA_TENSOR_FILES:
+            existing = set(matches)
+            matches = matches + [f for f in _EXTRA_TENSOR_FILES if f not in existing]
+        return matches
+
+    def __getattr__(self, name):
+        return getattr(_glob_module, name)
+
+
+def apply() -> bool:
+    """Install the glob proxy on ``mlx_lm.utils``. Idempotent."""
+    try:
+        from mlx_lm import utils as mlx_lm_utils
+    except ImportError:
+        logger.debug("mlx_lm.utils not importable; skipping extra-tensors patch")
+        return False
+
+    if not getattr(mlx_lm_utils.glob, "_omlx_extra_tensors_proxy", False):
+        mlx_lm_utils.glob = _GlobProxy()
+        logger.info("mlx-lm MTP sidecar glob patch applied")
+    return True
--- a/omlx/patches/mlx_lm_extra_tensors.py
+++ b/omlx/patches/mlx_lm_extra_tensors.py
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
-"""Patch mlx-lm's weight-shard glob to pick up indexed MTP sidecar files.
+"""Patch mlx-lm's weight-shard glob to pick up MTP sidecar files.
 
 ``mlx_lm.utils.load_model`` discovers weight shards with
 ``glob.glob(str(model_path / "model*.safetensors"))``. MTPLX-forge exports
@@ -14,8 +14,8 @@ missing the mtp.* tensors" even though the checkpoint has them (issue
 Fix: replace the ``glob`` name ``mlx_lm.utils`` resolves at call time with a
 thin proxy. Every pattern except the ``model*.safetensors`` shard-discovery
 call passes straight through to the real ``glob.glob``; that one call gets
-augmented with sidecar files referenced by ``mtp.*``-prefixed keys in the
-model's safetensors index.
+augmented with sidecar files declared by that model's config or referenced by
+``mtp.*``-prefixed keys in its safetensors index.
 
 This is distinct from issue #1944 / PR #1962, which fix the Native MTP
 *compatibility check* (``_checkpoint_has_mtp_weights`` /
@@ -29,11 +29,10 @@ from __future__ import annotations
 import glob as _glob_module
 import json
 import logging
+import os
 from pathlib import Path
 
 logger = logging.getLogger(__name__)
-
-_EXTRA_TENSOR_FILES: list[str] = []
 
 _MTP_INDEX_PREFIXES = (
     "mtp.",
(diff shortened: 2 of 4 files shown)


aosama (approved):
## Review: fix: load indexed MTP sidecar safetensors skipped by mlx-lm's glob

### Summary (500 words)

This pull request resolves issue #2062, which reported that `wang-yang/Ornith-1.0-35B-MTPLX` could not use Native MTP because the model's MTP weights live in a separate `mtp.safetensors` sidecar file. Although `model.safetensors.index.json` correctly references 785 `mtp.*` keys in that sidecar, `mlx_lm.utils.load_model` only discovers files matching `model*.safetensors`, so the sidecar was silently skipped. The loader then raised "Native MTP is enabled for this model but the converted weights are missing the mtp.* tensors" even though the weights were present.

The fix introduces a narrow, scoped patch in `omlx/patches/mlx_lm_extra_tensors.py`. It reads the safetensors index, collects filenames backing `mtp.*`-prefixed keys (and a few nested variants such as `language_model.mtp.*`, `model.mtp.*`, and `model.language_model.mtp.*`), and registers those files in a process-wide list. A thin `glob` proxy is installed on `mlx_lm.utils`, which passes through every glob call unchanged except for the specific `model*.safetensors` shard-discovery pattern; that call is augmented with the registered sidecar files. The patch is idempotent and reset before every model load so a model without an MTP sidecar does not inherit a previous load's file list.

The dispatcher in `omlx/utils/model_loading.py` applies the patch whenever the config declares MTP heads (`_has_mtp_heads`), regardless of whether `mtp_enabled` is true, mirroring the existing Native MTP patch policy for sanitize correctness. It logs the discovered sidecar files so operators can confirm the fix is active.

The test coverage is comprehensive: `tests/test_mlx_lm_extra_tensors.py` verifies index scanning (no index, malformed index, no MTP keys, bare `mtp.*` prefix, nested prefixes, sidecars already covered by model shards, multiple sidecars, deduplication, and non-string weight-map entries); the process-wide file-list flag; the glob proxy (passthrough, augmentation, no duplication, empty-flag behavior, attribute delegation, marker attribute); `apply()` idempotency and graceful handling when `mlx_lm` is not importable; and integration with `maybe_apply_pre_load_patches` in `model_loading.py`.

I verified the fix end-to-end by building the macOS dev app from this branch, loading `wang-yang/Ornith-1.0-35B-MTPLX`, enabling Native MTP, and running chat completions. The server log confirms the patch is applied, the sidecar is found, Native MTP loads, and the MTP generator activates and accepts draft tokens. Before this fix the same checkpoint failed with the missing-weights error.

I also attempted a token-per-second comparison. With Native MTP enabled, a 2000-token-max summarization request generated 1216 tokens in 37.46 seconds, or about 32.5 tok/s. With Native MTP disabled on the same model and a fresh reload, an 811-token completion took 10.05 seconds, or about 80.7 tok/s. These numbers are **not directly comparable**: the prompts and outputs differed, the second run benefited from prefix-cache state, and the model's generation length was not constrained to be identical. The comparison should therefore be treated as directional context rather than a rigorous benchmark. The primary value of this PR is correctness: it makes Native MTP actually usable for MTPLX-forge exports, which previously failed at load time.

The code is well-scoped, follows the repository's patch conventions, preserves upstream module names, and includes thorough tests. CI passes on Python 3.11, 3.12, and 3.13. I recommend merging this PR and closing #2062.

### Detailed Review (additional findings and notes)

**Root-cause precision.** The PR description correctly distinguishes this fix from #1944 / PR #1962. Those changes addressed the MTP compatibility check (`_checkpoint_has_mtp_weights` / `_model_has_mtp_weight_tensors`) for OptiQ checkpoints whose sidecar is not listed in the index at all. They do not change weight loading. This PR addresses the complementary problem: an indexed sidecar that the loader still misses because of its filename. Keeping the two fixes separate is the right call.

**Patch hygiene.** The `set_extra_tensor_files([])` reset is placed next to the existing `set_mtp_active(False)` reset, which makes the lifecycle easy to reason about. The sidecar scan happens only when `_has_mtp_heads(config)` is true, so non-MTP models pay no cost. The proxy only intercepts the exact `model*.safetensors` pattern; tokenizer, config, checkpoint-save, and pipeline-selection globs pass through untouched. This minimizes the blast radius of monkey-patching `mlx_lm.utils.glob`.

**Logging clarity.** The `INFO` log line prints the number of sidecar files and their basenames, e.g.:
```
MTP sidecar glob patch applied for /Users/ahmedhamdy/.omlx/models/wang-yang/Ornith-1.0-35B-MTPLX (1 file(s): mtp.safetensors)
```
This makes verification straightforward from server logs.

**Edge-case handling.** The index reader gracefully handles missing files, malformed JSON, non-string keys/values, and sidecars whose names already match `model*.safetensors`. It sorts and deduplicates results. The `_GlobProxy` avoids duplicating a match that the real glob already returned. These details are all covered by tests.

**Test quality.** The new test file is large but well-organized into classes with clear responsibilities. It uses `tmp_path` fixtures and avoids network downloads. The integration test constructs a minimal fake model directory with `config.json` and a safetensors index, then calls `maybe_apply_pre_load_patches` directly. This is the right level of testing for a loader patch.

**Naming.** The module name `mlx_lm_extra_tensors` is generic, which could imply broader ambitions than the current MTP-only scope. Given the implementation, the name is acceptable because the prefixes list is the only MTP-specific part; future sidecar types could be added by extending `_MTP_INDEX_PREFIXES` or adding analogous prefix sets. If the maintainers prefer, a more specific name such as `mlx_lm_mtp_sidecars` would also be reasonable, but the current name is not blocking.

**Documentation and changelog.** The PR should update any user-facing docs that mention Native MTP compatibility. Specifically, the workaround described in #2062 (creating a `model-mtp.safetensors` symlink) can now be removed from tribal knowledge. A brief note in release notes would help users who previously hit this error.

### Verification Evidence

Environment:
- macOS 26.5.1
- oMLX built from `fix/mtp-sidecar-glob` at commit referenced by PR #2220
- Model: `wang-yang/Ornith-1.0-35B-MTPLX` (20.73 GB, `qwen3_5_moe`, MTPLX-forge export)
- Sidecar layout confirmed: `model-00001-of-00004.safetensors` through `model-00004-of-00004.safetensors`, plus `mtp.safetensors` with 785 indexed `mtp.*` tensors.

Server log excerpts with this PR:
```
2026-07-12 22:04:36,982 - omlx.patches.mlx_lm_extra_tensors - INFO - mlx-lm MTP sidecar glob patch applied
2026-07-12 22:04:36,982 - omlx.utils.model_loading - INFO - MTP sidecar glob patch applied for /Users/ahmedhamdy/.omlx/models/wang-yang/Ornith-1.0-35B-MTPLX (1 file(s): mtp.safetensors)
2026-07-12 22:04:37,014 - omlx.utils.model_loading - INFO - Native MTP patch applied for /Users/ahmedhamdy/.omlx/models/wang-yang/Ornith-1.0-35B-MTPLX (model_type=qwen3_5_moe, active)
2026-07-12 22:04:41,690 - omlx.patches.mlx_lm_mtp.batch_generator - INFO - MTP path activated for uid=0 (model has mtp_forward, batch=1)
2026-07-12 22:04:42,194 - omlx.patches.mlx_lm_mtp.batch_generator - INFO - MTP[0] finish=length tokens=10 cycles=2 tok/cycle=5.00 accept=6/6 (100.0%)
```

Chat completion returned successfully, confirming the model is not just loaded but actually generating with MTP active.

### Recommendation

Approve and merge. This is a focused, well-tested fix for a real regression-class bug. After merge, close #2062.

aosama:
## Verified working — end-to-end test passed

I tested this fix locally on `wang-yang/Ornith-1.0-35B-MTPLX` and it resolves the issue.

**Setup:**
- Built the macOS dev app from this branch
- Loaded `Ornith-1.0-35B-MTPLX` with Native MTP enabled
- Ran chat completions against the model

**Before this PR:** the model would fail with `"Native MTP is enabled for this model but the converted weights are missing the mtp.* tensors."`

**After this PR:** the model loads successfully and MTP activates. Server log confirms:

```
mlx-lm MTP sidecar glob patch applied
MTP sidecar glob patch applied for .../Ornith-1.0-35B-MTPLX (1 file(s): mtp.safetensors)
Native MTP patch applied for .../Ornith-1.0-35B-MTPLX (model_type=qwen3_5_moe, active)
MTP path activated for uid=0 (model has mtp_forward, batch=1)
MTP[0] finish=length tokens=10 cycles=2 tok/cycle=5.00 accept=6/6 (100.0%)
```

The sidecar is found, Native MTP activates, and generation runs with MTP active.

**Recommendation:** Approve and merge. This is a focused, well-tested fix for a real regression-class bug.

jundot:
Thanks for working on this. The indexed Ornith case is real, and the index lookup fixes that specific checkpoint layout.

The first gap is coverage. This only discovers files referenced by `mtp.*` entries in `model.safetensors.index.json`. A common MTPLX layout, including `Youssofal/Qwen3.6-27B-MTPLX-Optimized-Speed`, keeps the sidecar out of the index and declares it only through `config.json` as `mlx_lm_extra_tensors.mtp_file`, sometimes using a nested path such as `mtp/weights.safetensors`. In that layout, `sidecar_files_for()` returns an empty list, so the patch is never activated and the sidecar is still not loaded.

The second issue is the process-wide `_EXTRA_TENSOR_FILES` state. After the target model loads, the list remains set. Both `BatchedEngine` and `VLMBatchedEngine` can then load a SpecPrefill draft through `lm_load_compat()` without calling `maybe_apply_pre_load_patches()` again. Those paths reset `mtp_active`, but they do not reset the extra tensor list. I reproduced this with separate target and draft directories: the draft model's `model*.safetensors` glob returned its own shard plus the target model's `mtp.safetensors`. Depending on the draft architecture, this can cause a strict-load failure or load unrelated tensors and consume extra memory, after which SpecPrefill is skipped. This can also happen with native MTP disabled because sidecar registration is gated on `_has_mtp_heads`, not `mtp_enabled`.

I think this should use a model-scoped resolver before merge, with coverage for both config-declared sidecars and index-referenced files. It would also be helpful to add a regression test that loads an MTP target and then a different SpecPrefill draft without running the pre-load dispatch a second time.

zviratko:
I have a question about this - this (or similias) issues pop up in my testing with various models. And even when I tried different runtimes - different from oMLX - each has its own problems with some models and node others.
Some problems I was able to solve with ad-hoc scripts hacking on the models (gemini seem very enthusiastic about creating those for some reason :)), sometimes models are even mislabeled as MLX while using FP8 or something like that (which as I undestand it is simply incompatible, but that't just one example). Seems like a small wonder that some models even work correctly (or maybe they don't and we just got used to the models producing nonsense... sometimes)

Isn't there a standard of how a mlx safetensors model should look? Shouldn't everything (other than some experimental functions) be loadable through mlx-lm? I am also guilty of reporting model loading errors here, but I think some (or maybe most) of them are actually broken model merges and configs, maybe we should start putting pressure upstream instead of wanting more patches to oMLX

mxl:
Thanks for the detailed review. I addressed both gaps and rebased the branch onto the latest `main`.

- `sidecar_files_for()` now merges `config.json` → `mlx_lm_extra_tensors.mtp_file` with `mtp.*` references from `model.safetensors.index.json`. Nested declarations such as `mtp/weights.safetensors` are supported, duplicate paths are removed, and paths escaping the model directory are rejected.
- The process-wide `_EXTRA_TENSOR_FILES` state and its setters/getters were removed. `_GlobProxy.glob()` now resolves sidecars from the directory of each `model*.safetensors` pattern, so a later SpecPrefill draft load can only discover files declared by that draft model.
- Added a regression test with separate target and draft directories. Pre-load dispatch runs only for the target; both models are then loaded through `mlx_lm.utils.load_model`, and the test verifies that the target sidecar is loaded while the draft sees only its own shard.
- During verification I also found that the DeepSeek V4 replacement loader captured the standard `glob` module and bypassed the proxy. It now resolves `_utils.glob.glob(...)`, preserving the model-scoped sidecar behavior after that loader patch is installed.

PR #2222, opened 2026-07-12
fix: text highlight in chat in dark mode

HaloFour:
Quick fix to the highlight text style in the chat interface in dark mode so that the highlighted text is visible:

Before:
image

After:
image

### DIFF
--- a/omlx/admin/templates/chat.html
+++ b/omlx/admin/templates/chat.html
@@ -62,6 +62,11 @@
         --timeline-accent: #60a5fa;
     }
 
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for fixing this. The dark-mode rule overrides the base selection style cleanly and matches the colors already used by the dashboard and login pages without changing light mode. LGTM, merging.

jundot:
@HaloFour, I released oMLX 0.5.2.dev1 with this change. Please try the new build and let me know if you see any related issues: https://github.com/jundot/omlx/releases

Issue #2069, opened 2026-07-02
Menu-bar app polls /health + /v1/models ~2×/sec 24/7 when idle; each /health poll opens a fresh socket and wastes a refused IPv6 ::1 connect


iRonin:
**Describe the bug**

The macOS menu-bar app continuously polls the local server over HTTP at roughly **2 requests/second, 24/7**, even when the app is idle, the menu is closed, and **no model is loaded** (`/health` reports `loaded_count: 0`). Two pollers run in a tight loop:

- `GET /health` — ~1/sec, and each poll opens a **brand-new TCP connection** (not keep-alive reused).
- `GET /v1/models` — ~1/sec, on a reused keep-alive connection.

The `/health` poller has a second, compounding problem: because it opens a fresh connection each time, the client resolves `localhost` and **attempts IPv6 `::1:1235` first, which is always refused** (the server binds **IPv4 `127.0.0.1` only**), then falls back to `127.0.0.1`. So every single health poll incurs a **guaranteed `ECONNREFUSED`** plus a wasted connect round-trip, once per second, indefinitely.

**Impact**

- Constant connection churn: **~286,000 client TCP connections** accumulated over one continuous session, and a flat **~137,000 log lines/hour** written to the unified log (`com.apple.network` + `CFNetwork` per-connection spam), which is real logging-subsystem and disk-write pressure with zero user activity.
- Every second produces a failed IPv6 connect (`CONNRESET` / `Socket SO_ERROR [61: Connection refused]`) in the logs, which is noise that also masks real connection problems.
- Battery/energy: a menu-bar app that is supposed to be idle is never actually idle.

**To Reproduce**

1. Launch the app with the server running, **no model loaded**, menu closed.
2. Observe the client→server traffic on the server port:
   ```
   log stream --predicate 'process == "oMLX"' --style compact | grep ':<port>'
   ```
   or `lsof -nP -iTCP:<port>` repeatedly.
3. Observe: a new connection to `127.0.0.1:<port>` about once per second for `/health`, each preceded by a refused `::1:<port>` attempt, plus a ~1/sec `/v1/models` poll on a reused connection. This continues indefinitely with no user interaction.

**Expected behavior**

- When idle (menu closed, no active inference), the menu-bar poller should back off dramatically (e.g. multi-second interval, or event-driven refresh only while the menu is open).
- The `/health` poller should **reuse a keep-alive connection** instead of opening a fresh socket each time.
[...]

Issue #1985, opened 2026-06-23
It is suggested that a status bar be added to display the current working status.


citywwm:
I suggest adding a bar at the top that displays the current operational status.
This status bar would be similar to the CPU or network status indicators, showing how many models are currently running and the operational status of each one.
This would mean we wouldn’t have to open the management panel to check the current operational status.

<img width="1558" height="230" alt="Image" src="https://github.com/user-attachments/assets/7853381f-7ae9-4492-a28b-859a7f51e30d" />

jundot:
@citywwm, I released oMLX 0.5.2.dev1 with opt-in live inference activity in the macOS menu bar. Please try the new build and let me know if this addresses the request: https://github.com/jundot/omlx/releases

Issue #1103, opened 2026-05-07
Feature request: Improve Active Models visibility in the dashboard


apetersson:
### Problem                                                                                                                                                                                                    
                                                                                                                                                                                                                
 The Status → Active Models section can look stuck even when the server is actively working.                                                                                                                    
                                                                                                                                                                                                                
 Examples:                                                                                                                                                                                                      
                                                                                                                                                                                                                
 - Generating... gives no indication whether tokens are still being produced.                                                                                                                                   
 - Long prompt preprocessing can sit at 0% for a while.                                                                                                                                                         
 - SpecPrefill work is not clearly visible.                                                                                                                                                                     
 - Loading only says loading.                                                                                                                                                                                   
 - Waiting requests are only shown as counts.                                                                                                                                                                   
                                                                                                                                                                                                                
 This makes it hard to tell whether a request is healthy, slow, queued, or actually hung.                                                                                                                       
                                                                                                                                                                                                                
 ### Proposed improvement                                                                                                                                                                                       
                                                                                                                                                                                                                
 Make the Active Models card show more live activity details.                                                                                                                                                   
                                                                                                                                                                                                                
 ### Suggested UI                                                                                                                                                                                               
                                                                                                                                                                                                                
 #### Generating                                                                                                                                                                                                
                                                                                                                                                                                                                
 Show token progress and speed:                                                                                                                                                                                 
                                                                                                                                                                                                                
 ```text                                                                                                                                                                                                        
   abcd1234  ● Generating...   170 tok   13.0 tok/s                                                                                                                                                             
 ```                                                                                                                                                                                                            
                                                                                                                                                                                                                
 If generation appears inactive for a few seconds, show:                                                                                                                                                        
                                                                                                                                                                                                                
 ```text                                                                                                                                                                                                        
   last token 5s ago                                                                                                                                                                                            
 ```                                                                                                                                                                                                            
                                                                                                                                                                                                                
 #### Prompt preprocessing / SpecPrefill                                                                                                                                                                        
                                                                                                                                                                                                                
 Show clearer phase names and progress, for example:                                                                                                                                                            
                                                                                                                                                                                                                
 ```text                                                                                                                                                                                                        
   scoring draft tokens                                                                                                                                                                                         
   system prompt prefill                                                                                                                                                                                        
   sparse target prefill                                                                                                                                                                                        
 ```                                                                                                                                                                                                            
                                                                                                                                                                                                                
 Also show useful metadata like:                                                                                                                                                                                
[...]

Issue #163, opened 2026-03-11
Add api to enable statusline integration


jmaver:
Right now, basic claude code statusline integration shows context usage/200k and then 5h and 7day info.

when I use the GLM coding plan from z.ai, I have mine customized to pull limits/usage from them.

When using omlx as the backend, it would be nice to be able to have it call something in omlx to get the omlx stats, especially since they are so unique to omlx.

```
 curl -s  \
    -H "Authorization: Bearer $AUTH_TOKEN" \
    "${host_base}/api/monitor/usage/quota/limit"
}
```

Maybe some info related to context scaling (we get to set our own target) and cache efficiency?

jundot:
Nice idea. I didn't know z.ai was doing it that way. I'll look into it more and see how to best implement this. Thanks for the info.

jmaver:
Here is my statusline for reference. It gets the provider from an env var CC_PROVIDER and then adjusts display accordingly:

```
#!/bin/bash

# ==============================================================================
# Claude Code Status Line (Anthropic + Z.ai)
# ==============================================================================

input=$(cat)

if [ -z "$input" ]; then
  printf "Claude"
  exit 0
fi

# Get CC_PROVIDER from environment, default to 'claude' if not set
CC_PROVIDER="${CC_PROVIDER:-claude}"

mapfile -d '' -t __fields < <(echo "$input" | jq -r '
  .model.display_name // "Unknown" as $model
  | .model.name // "" as $model_name
  | .mode // "" as $mode
  | (.context_window.context_window_size // 200000) as $ctx
  | (.context_window.current_usage // {}) as $u
  | ($u.input_tokens // 0) as $in
  | ($u.cache_creation_input_tokens // 0) as $cc
  | ($u.cache_read_input_tokens // 0) as $cr
  | (.context_window.remaining_percentage // "") as $remaining
  | (.session_id // "") as $session
  | [$model, $model_name, $mode, ($ctx|tostring), (($in + $cc + $cr)|tostring), ($remaining|tostring), $session]
  | join("\u0000") + "\u0000"
')
MODEL=${__fields[0]}
MODEL_NAME=${__fields[1]}
MODE=${__fields[2]}
CONTEXT_SIZE=${__fields[3]}
CURRENT=${__fields[4]}
REMAINING_PCT=${__fields[5]}
SESSION_ID=${__fields[6]}

# ==============================================================================
# Anthropic OAuth Usage Fetch
# ==============================================================================

[...]

PR #2224, opened 2026-07-13
feat(macOS): show live inference activity in the menu bar

### LINKED ISSUE #250: Can we add monitoring to the menu bar?菜单栏能不能加上监控
Love this! The 0.2.13 update added the ability to check the active model's speed, token count, estimated speed, and single request.
I always have to check on the web interface; it would be perfect if this could be shown in the menu bar.
大赞这个，0.2.13新增的可以查看活跃模型的工作速度，多少token，预估速度，1次请求
我要一只去网页端查看，如果菜单栏能现实就完美了

aosama:
## Summary

- add an opt-in menu-bar live activity display for prompt prefill, generation speed/token count, queued requests, and non-streaming work
- keep the bird icon stationary inside one unified status item while activity text expands to its left
- add a lightweight authenticated `/admin/api/activity` endpoint instead of polling the full runtime-cache stats payload
- preserve the existing 2-second idle status cadence and use 1-second activity polling only after explicit opt-in
- expose the live state through the status item accessibility value and consolidate UI updates to one notification per poll cycle

## Testing

- `xcodebuild test -project oMLX.xcodeproj -scheme oMLX -destination platform=macOS` — 200 passed, 1 skipped
- `uv run pytest tests/test_admin_api_key.py tests/test_active_models_visibility.py -q` — 81 passed
- localization catalog JSON validation and `git diff --check` passed
- Ruff and mypy were also run; both retain existing repository-wide baselines unrelated to the changed lines

## Issues

Closes #250.
Addresses #1985.
Related to #45, #163, and #1103.

The adaptive polling and lightweight activity endpoint are designed to avoid regressing the idle polling concerns tracked in #2069.

### DIFF
--- a/apps/omlx-mac/Resources/Localizable.xcstrings
+++ b/apps/omlx-mac/Resources/Localizable.xcstrings
@@ -4878,6 +4878,24 @@
         }
       }
     },
+    "menubar.item.show_live_activity_in_menu_bar" : {
+      "comment" : "Menubar item that toggles current request activity text beside the oMLX icon",
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Show Live Activity in Menu Bar"
+          }
+        },
+        "ru" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Показывать активность в строке меню"
+          }
+        }
+      }
+    },
     "menubar.item.start_server" : {
       "comment" : "Menubar item that starts the server",
       "extractionState" : "manual",
@@ -5054,6 +5072,24 @@
           "stringUnit" : {
             "state" : "translated",
             "value" : "Загрузка статистики…"
+          }
+        }
+      }
+    },
+    "menubar.stats.live_activity" : {
+      "comment" : "Section header inside the Serving Stats submenu for the current active request",
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Live Activity"
+          }
+        },
+        "ru" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Текущая активность"
           }
         }
       }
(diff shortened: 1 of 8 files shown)


aosama:
image
<br/>
image

aosama:
image

aosama:
Removed an extra space on the left hand side of the icon, I think this is good to go now.
image

aosama:
Follow-up fix pushed in ccefacdc.

When no live activity was visible, the hidden activity label still retained required leading/trailing constraints. That reserved empty space on the left of the bird. The label constraints are now inactive while idle and reactivated only when activity text is visible.

This preserves the single NSStatusItem, unified hover/click target, and fixed bird position. Added a regression assertion for idle versus active label-space reservation.

Validation: macOS XCTest suite passes (200 passed, 1 skipped). Rebuilt the staged debug bundle for visual inspection.

aosama:
hello @jundot , the menu bar capability in this PR is optional and the user can switch on / off , let me know if you want changes in the what is displayed.

jundot:
Screenshot 2026-07-18 at 12 12 41 AM

Thanks for the great PR, this came out really clean. A menu bar live activity display has been on my long term plan for a while, but I never had the capacity to actually start it, so I really appreciate you proposing it as a PR. Also sorry for the slow review, I'm on vacation this week.

For the final shape, what I ultimately have in mind is something like iStat Menus in the screenshot above, where the item is split into top and bottom rows so PP and TG can be stacked vertically and take as little menu bar space as possible. I'm also considering a small chart to show trends over time, but for now the stacked two row layout is the part I care about most.

I think the current implementation is already valuable on its own, and since it's opt-in the choice stays with the user, so I'll merge this. I'll work on the further improvements as follow ups. Of course another PR from you toward that direction would be even more appreciated, but what you built here is already really nice.

aosama:
Roger that ...

jundot:
@aosama, I released oMLX 0.5.2.dev1 with this change. Please try the new build and let me know if you see any related issues: https://github.com/jundot/omlx/releases

aosama:
coooollllllll

aosama:
I am testing it, the default it seems is ON by default, it is possible we have it OFF by default, but users will need to discover it in the menu and switch on .... I am leaning towards off by default .. but I am not too opinionated there

Issue #1833, opened 2026-06-11
Memory enforcer counts serialized hot-cache CPU bytes as process usage — long-context requests aborted once the hot cache fills (0.4.4rc1)


hojin12312:
## Summary

On 0.4.4rc1, `ProcessMemoryEnforcer._current_usage_bytes()` uses the raw kernel `phys_footprint`, which includes the serialized hot-cache blocks (plain CPU malloc, up to `hot_cache_max_size`). Once the hot cache fills during a long multi-turn session, `usage ≈ GPU active + hot_cache_max_size` crosses the hard watermark and the enforcer aborts healthy in-flight requests, even though the scheduler-side guard would have admitted them.

The scheduler's own guard already excludes these bytes (`Scheduler._current_usage_bytes` subtracts `_hot_cache_cpu_bytes()` from `phys_footprint`). The enforcer is the remaining call site without the exclusion.

## Observed impact

Gemma 4 26B (UD-3bit), hot cache 6 GB, ceiling 30 GB (iogpu.wired_limit_mb=30720), multi-turn sweep (2K user tokens/turn), found while verifying the fix for #1832:

- The session dies at ~131K context: turn 120 gets
  `RuntimeError: Request aborted: process memory limit exceeded (usage 31.1 GB, ceiling 30.0 GB)` — 31.1 GB = ~25 GB GPU active + ~6 GB full hot cache. The server log shows the emergency path: `Emergency memory pressure: aborted 1 in-flight request(s) (current=31.1GB, ceiling=30.0GB); models kept loaded.`
- With the same exclusion the scheduler applies, the identical sweep passes 131K at 99% hit. (Its eventual stop slightly later comes from the pre-chunk transient estimator, unrelated to the enforcer.)

The regression scales with `hot_cache_max_size`: the bigger the hot cache (which exists precisely to make long multi-turn sessions cheap), the earlier the enforcer kills those sessions.

## Root cause

`process_memory_enforcer.py`:

```python
def _current_usage_bytes(self) -> int:
    phys = get_phys_footprint()          # <-- includes hot-cache CPU bytes
    if self._has_active_requests():
        return max(self._cached_executor_active_memory_bytes(), phys)
    return max(mx.get_active_memory(), phys)
```

versus the scheduler guard (already fixed):

```python
def _current_usage_bytes(self, *, refresh_mlx_active: bool = True) -> int:
    ...
    phys = max(0, int(get_phys_footprint()) - hot_cache_bytes)
[...]

Issue #1796, opened 2026-06-10
Memory guard counts serialized hot-cache CPU bytes as GPU pressure — prefill safety cap shrinks 1:1 as the hot cache fills


hojin12312:
## Describe the bug

The prefill/store memory guard counts **serialized hot-cache CPU bytes as if they were GPU pressure**, so as the hot cache fills, the prefill safety cap shrinks 1:1 — long multi-turn conversations get rejected by the cap far below the real Metal budget.

`_current_usage_bytes()` (scheduler.py) returns `max(mx_active, phys_footprint)`. But the hot cache deliberately keeps blocks as **serialized CPU bytes** (`save_block` stores `tensors_raw` so Metal buffers can be released immediately — a good design). Those bytes are ordinary swappable process memory, yet they are inside `phys_footprint`, so every cached block eats the guard ceiling one-for-one even though it exerts no Metal/wired pressure.

Effect: an active conversation's KV is effectively **double-counted** (live Metal copy + hot-cache CPU copy), and the guard rejects prefills while actual GPU usage still has headroom.

Verified present in v0.4.3 and unchanged on `main` (0.4.4.dev1) — `_current_usage_bytes` is still `max(active, get_phys_footprint())` with no hot-cache exclusion.

## How we isolated it

Granite 4.1 30B (dense, 64 layers, TurboQuant KV 4-bit), growing multi-turn sweep on a 36 GB Mac Studio, ceiling 30 GB: the conversation hit a hard **32K-token ceiling** (pre-chunk guard rejection) while `mx_active` was well below the cap.

- `footprint <pid>` at idle: 22 GB = IOAccelerator 19 GB + MALLOC_SMALL 2.85 GB.
- Re-ran the sweep under `MallocStackLogging=lite`: malloc grew 90 MB → 1.03 GB over 12 turns, and `malloc_history` backtraces converge on `_PyBytes_FromBuffer` ← `_extract_tensor_bytes` (the hot-cache serialization path).
- Numbers reconcile exactly: ~64 MB serialized per block (TQ 4-bit, 64 KB/token × 1024-token blocks) × 11 blocks ≈ 934 MB.

So at the 32K rejection point, ~2.4 GB of the "usage" the guard saw was hot-cache CPU bytes, not GPU memory.

## Suggested fix

Subtract the tracked hot-cache total from the **phys side only** of the guard:

```python
phys = max(0, get_phys_footprint() - self._hot_cache_cpu_bytes())
return max(active, phys)
```

[...]

PR #2229, opened 2026-07-14
fix(dflash): stop double-charging hot-cache bytes in the primary prefill guard

JimStenstrom:
Addresses the last remaining sibling path of #1796 (see also #1833).

## Summary

- The DFlash primary-path prefill guard no longer charges serialized hot-cache CPU bytes twice. The enforcer now propagates the hot-cache used-bytes figure alongside the reservation it already sends, and `_DFlashPrefillGuard` subtracts it from `phys_footprint` — mirroring what `Scheduler._current_usage_bytes` already does.

## Why

Since #1833's fix (44588854), the enforcer subtracts a hot-cache reservation from the ceiling it propagates to every watermark target, and since #1796's fix (e52aeb93) the scheduler compensates on the usage side by excluding hot-cache CPU bytes from its `phys_footprint` read — together the hot cache is charged exactly once on scheduler-driven paths.

The DFlash primary guard receives the same reserved ceiling (it's a `_resolve_scheduler` target), but still measured usage as raw `max(active, phys_footprint)`. Hot-cache bytes were therefore charged twice on that path — once in the reduced limit, once in usage — so as the hot cache fills, the guard over-rejects by up to the hot-cache size. That's the same false-rejection mechanism both issues describe:

> Memory guard counts serialized hot-cache CPU bytes as GPU pressure — prefill safety cap shrinks 1:1 as the hot cache fills (#1796)

> long-context requests aborted once the hot cache fills (#1833)

fixed for the scheduler and enforcer paths, surviving only on DFlash primary. Nobody has reported it there yet (it needs DFlash primary + a populated hot cache from a co-resident scheduler-driven model + a prefill near the ceiling), so this is a proactive parity fix rather than a response to a field report.

## Changes

- `omlx/process_memory_enforcer.py` — `_propagate_memory_limit` computes the hot-cache used-bytes figure and writes it to every target as `_memory_hot_cache_used_bytes`, next to the existing `_memory_hot_cache_reserved_bytes`. The value is clamped to the reservation, so a transient overshoot past `max_bytes` can never exclude more usage than the ceiling actually gave up (the exclusion is never net-optimistic).
- `omlx/engine/dflash.py` — `_DFlashPrefillGuard._current_usage_bytes` subtracts that figure from `phys_footprint` (clamped at zero) before taking the max with the recorded MLX-active sample. The active floor is never reduced — GPU pressure accounting is untouched.

Sibling sweep: the Scheduler keeps reading its own live counter and simply ignores the new attr (no double application); the enforcer's own process-level pressure accounting deliberately stays on full `phys_footprint` (unchanged, per 44588854's design); DFlash fallback mode resolves the fallback engine's real scheduler, which already has the exclusion; diffusion engines are skipped by the propagation loop as before.

## Why this is safe

- With no enforcer running (standalone guards, tests), the new attr stays 0 and the guard computes exactly what it does today.
- The propagated value is clamped to the reservation, so usage-side exclusion ≤ limit-side reservation always holds — the guard can only get *stricter* than a perfectly-symmetric accounting, never looser.
- The preflight path gains no MLX/Metal or cache-manager calls — the guard reads one plain int attribute written by the enforcer tick, same pattern as the existing watermark attrs.

## Tests

Four new tests, in the existing per-module files:

- `test_guard_excludes_hot_cache_bytes_from_physical_usage` — the double-count repro: a prefill that fits exactly iff the exclusion is applied, plus the inverse (still rejects when genuinely over).
- `test_guard_hot_cache_exclusion_clamps_and_keeps_active_floor` — an oversized hot-cache figure clamps to zero instead of going negative, and the MLX-active floor still rejects on its own.
- `test_propagates_hot_cache_used_for_usage_side_exclusion` — the enforcer writes the used figure to resolved targets each tick.
- `test_propagated_hot_cache_used_clamped_to_reservation` — an overshoot past `max_bytes` propagates the clamped reservation, not the raw figure.

```
$ pytest tests/test_dflash_prefill_memory_guard.py tests/test_process_memory_enforcer.py tests/test_scheduler_prefill_memory_guard.py -q
187 passed
```

Also verified all four new tests fail when the fix is reverted (guard counts raw phys; enforcer never writes the attr).

### DIFF
--- a/omlx/engine/dflash.py
+++ b/omlx/engine/dflash.py
@@ -93,12 +93,21 @@ class _DFlashPrefillGuard:
         # Written by ProcessMemoryEnforcer._propagate_memory_limit each tick.
         self._prefill_memory_guard: bool = False
         self._memory_hard_limit_bytes: int = 0
+        self._memory_hot_cache_used_bytes: int = 0
 
     def record_mlx_active_memory(self, active_bytes: int) -> None:
         self._last_mlx_active_memory_bytes = max(0, int(active_bytes))
 
     def _current_usage_bytes(self) -> int:
-        return max(self._last_mlx_active_memory_bytes, get_phys_footprint())
+        # The enforcer already subtracts a hot-cache reservation from the
+        # ``_memory_hard_limit_bytes`` it propagates here, so serialized
+        # hot-cache CPU bytes must not ALSO be counted in usage via
+        # phys_footprint — that charges them twice and over-rejects by the
+        # hot-cache size. Mirrors ``Scheduler._current_usage_bytes``.
+        phys = max(
+            0, get_phys_footprint() - max(0, int(self._memory_hot_cache_used_bytes))
+        )
+        return max(self._last_mlx_active_memory_bytes, phys)
 
     def preflight_or_raise(
         self,
--- a/omlx/process_memory_enforcer.py
+++ b/omlx/process_memory_enforcer.py
@@ -1009,6 +1009,14 @@ class ProcessMemoryEnforcer:
         hot_cache_reserved = (
             self._hot_cache_reserved_bytes() if ceiling > 0 or abort_limit > 0 else 0
         )
+        # Clamp to the reservation: the usage-side exclusion must never exceed
+        # what the ceiling actually gave up, or a transient hot-cache overshoot
+        # past max_bytes would net-weaken the guard exactly under pressure.
+        hot_cache_used = (
+            min(self._hot_cache_used_bytes(), hot_cache_reserved)
+            if hot_cache_reserved > 0
+            else 0
+        )
         scheduler_ceiling = self._scheduler_limit_bytes(
             ceiling, reserved=hot_cache_reserved
         )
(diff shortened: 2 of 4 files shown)


jundot:
Thanks, this is a really well reasoned fix. I verified the pairing argument holds: the exclusion and the reservation are written as a same-tick pair, so the guard can never get looser than the raw phys + peak vs ceiling baseline, and the clamp keeps the DFlash path strictly on the conservative side. I'll merge this. Two things I'll pick up myself in follow-ups: reusing the used-bytes figure already computed inside `_hot_cache_reserved_bytes()` so the pool scan doesn't run twice per tick, and your clamp comment made me notice the Scheduler's live counter path still has no such cap, so I'll port the same clamp there.

jundot:

JimStenstrom:
Tried 0.5.2.dev1 (cp312 wheel, M5 Max): DFlash primary on Qwen3-Coder-Next oQ4 with the z-lab draft and the in-memory hot cache enabled (`l1_cache=True`) — repeated and fresh ~3k-token prefills all admitted and correct, including an exact tail-of-prompt read, with zero guard rejections in the log. This run had plenty of free memory, so it exercises the admission path rather than the guard's rejection boundary. No related issues seen.

PR #2230, opened 2026-07-14
fix(admin): give the accuracy bench queue chain an ownership token

### LINKED ISSUE #1655: Benchmark Queue Overwriting Active Tests
**Describe the bug**
The benchmark queue seems to allow only one model to be queued at a time. If a user attempts to add a second model while the first model is still running and the queue slot is occupied, the system interrupts the running model.

**Observed Behavior**
The running model is stopped, its results are reset to 0.0%, and the queued model immediately starts. This action frees up the queue slot, allowing the new model to be added.

**Expected Behavior**
The queue should function as a waiting list. Adding a new model should simply append it to the end of the queue, waiting for the current running model to finish naturally.

**Conclusion**
The queue logic is incorrectly overwriting the waiting slot instead of appending to it, which causes active tests to be prematurely terminated.

 - macOS Version: 15.7.7
 - oMLX Version 0.4.1

JimStenstrom:
Fixes #1655.

## Summary

- Benchmark-queue runs started by `_continue_queue` are now hard-cancellable, and a soft-cancelled chain can no longer pop the queue or reset the running gate after a new chain has started.
- Fixes the reported symptom: the running model's test is stopped, its result card collapses to ~0.0%, and the queued model starts immediately.

## Why

A "chain" is one `start_next_from_queue` call plus the `_continue_queue` tail it spawns. Runs started by `_continue_queue` never had `run.task` set, so `cancel_queue` could only soft-cancel them: the run kept executing until its next `on_progress` checkpoint (up to a full generation batch away), while the `_queue_running` gate was reset immediately.

Any model queued in that window started a new chain — and when the soft-cancelled run finally returned, the old chain's trailing `_continue_queue` unconditionally popped the new queue and ran its item concurrently with the live chain. The second run's Phase 1 "unload all models" then ripped the engine out from under the active run: the running model is stopped, its result collapses to ~0.0%, and the queued model immediately starts — exactly the behavior in the report:

> The running model is stopped, its results are reset to 0.0%, and the queued model immediately starts.

Scope note: a plain add-while-running correctly appends to `_queue` on current `main`. The corruption window opens with the first `cancel_queue`; from then on, every add-while-running can corrupt the active test, because each finishing chain resets the gate while the other still runs.

## Changes

All in `omlx/admin/accuracy_benchmark.py`:

- A module-level `_chain_id` ownership token. `start_next_from_queue` bumps and captures it; `_continue_queue` bails out first thing when its token is stale — without popping the queue or touching `_queue_running`/`_current_run_id`; `cancel_queue` bumps the token to orphan the live chain *before* releasing the gate.
- Queue-continued runs record their chain task (`asyncio.current_task()`) so `cancel_queue` hard-cancels them instead of waiting for the checkpoint.
- Normal (non-cancel) queue flow is unchanged, and cancelled runs still emit their terminal SSE error event so attached streams close.

Design note — why a token rather than having the trailing `_continue_queue` check its run's `status == "cancelled"`: that check would work today, but only because the sole cancel path is whole-queue and there happens to be no yield between a run's terminal-status write and its continuation. The token keys on chain identity instead of run outcome, so it stays correct if either of those assumptions ever changes (e.g. a future per-run cancel).

Sibling paths: external-endpoint benchmark runs share the same queue chain, so they get the same ownership protection; the `sampling_profile` path is orthogonal (per-run request field, untouched).

## Why this is safe

- The orphaned run may still overlap the new chain briefly while its last batch drains — that's intended cancel semantics, not the bug: the orphaned run's status is already `"cancelled"` and its results discarded, and the new chain's Phase 1 unload just accelerates its teardown (`engine.stop()` aborts in-flight generation; a redundant unload is a guarded no-op). Awaiting the cancelled task inside `cancel_queue` would block the HTTP cancel route for up to a generation batch with no correctness gain.
- Hard-cancel can now land during a continued run's final unload phase and label an effectively-finished run `cancelled` — that's the pre-existing semantics first-of-chain runs (whose `run.task` was always set) already have; accumulated results are never rolled back either way. This change just makes continued runs consistent.

## Tests

Three regression tests in `tests/test_accuracy_benchmark.py::TestQueueChainOwnership`:

- `test_continue_queue_run_is_hard_cancellable` — a queue-continued run records its chain task and `cancel_queue` cancels it promptly instead of waiting for the checkpoint.
- `test_stale_chain_leaves_queue_and_gate_alone` — an orphaned chain's trailing `_continue_queue` neither pops the queue nor resets the gate.
- `test_cancel_then_requeue_does_not_corrupt_new_chain` — the full reported scenario: cancel, re-add while the soft-cancelled run drains, and the new chain runs its models sequentially with no concurrent second run.

```
$ pytest tests/test_accuracy_benchmark.py -q
33 passed
```

Also verified all three new tests fail when the fix is reverted (the stale chain pops the new queue and runs concurrently).

### DIFF
--- a/omlx/admin/accuracy_benchmark.py
+++ b/omlx/admin/accuracy_benchmark.py
@@ -37,6 +37,15 @@ _queue_running: bool = False
 _current_run_id: Optional[str] = None
 _current_model: Optional[str] = None
 _engine_pool_ref: Any = None
+# Chain-ownership token. A "chain" is one start_next_from_queue call plus
+# the _continue_queue tail it spawns. Each chain captures the token current
+# at its start; cancel_queue and start_next_from_queue bump it. A chain
+# whose token is stale (e.g. it was soft-cancelled and only noticed at its
+# next checkpoint, after the user already started a new chain) must not pop
+# the queue or mutate _queue_running/_current_run_id — otherwise it starts
+# a run concurrently with the live chain, whose Phase 1 "unload all models"
+# rips the engine out from under the active run.
+_chain_id: int = 0
 
 VALID_BENCHMARKS = [
     "mmlu", "mmlu_pro", "kmmlu", "cmmlu", "jmmlu",
@@ -221,6 +230,7 @@ def start_next_from_queue(engine_pool: Any) -> Optional[str]:
     This is synchronous so the caller gets the bench_id immediately.
     """
     global _queue_running, _current_run_id, _current_model, _engine_pool_ref
+    global _chain_id
 
     _engine_pool_ref = engine_pool
 
@@ -233,6 +243,10 @@ def start_next_from_queue(engine_pool: Any) -> Optional[str]:
     request = _queue.pop(0)
     _queue_running = True
     _current_model = request.model_id
+    # This chain takes ownership of the queue; any earlier chain still
+    # draining a soft-cancelled run bails at its next _continue_queue call.
+    _chain_id += 1
+    my_chain = _chain_id
 
     cleanup_old_runs()
     run = create_run(request)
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the deep dive, the chain ownership analysis matches exactly what I traced in the code. I also verified the repro locally: with the fix reverted, the requeue test fails with the stale chain popping the new queue, which is exactly the reported corruption. I'll merge this.

jundot:

JimStenstrom:
Tried 0.5.2.dev1 (cp312 wheel): queued two accuracy runs back-to-back on the same model (gsm8k then arc_challenge) — the second run started automatically with a fresh bench_id as soon as the first completed, both recorded results, and the queue drained to idle with no stale-chain behavior. No related issues seen.

PR #823, opened 2026-04-17
fix: honor response_format in /v1/audio/speech (#753)

ethannortharc:
## Summary

Closes #753.

`POST /v1/audio/speech` accepted OpenAI's `response_format` field on the request model but silently ignored it — every response was 16-bit mono WAV with `Content-Type: audio/wav`, regardless of what the caller asked for.

Reproducer from the issue:

```bash
curl -X POST http://localhost:8000/v1/audio/speech \
  -H 'Content-Type: application/json' \
  -d '{"model": "Qwen3-TTS-12Hz-1.7B-Base-bf16",
       "input": "hello", "voice": "alloy",
       "response_format": "opus"}' \
  --output test.opus && file test.opus
# Before: RIFF (little-endian) data, WAVE audio, …
# After : Ogg data, Opus audio, version 0.1, mono, 24000 Hz (Input Sample Rate)
```

## What changed

**`omlx/engine/audio_utils.py`** — new shared transcoder:

| `response_format` | Handling | Content-Type |
|---|---|---|
| `wav` | passthrough | `audio/wav` |
| `pcm` | strip RIFF header → raw 16-bit LE samples | `audio/pcm` |
| `mp3` | ffmpeg → libmp3lame (VBR q=2) | `audio/mpeg` |
| `opus` | ffmpeg → libopus in Ogg (64 kbps) | `audio/ogg` |
| `flac` | ffmpeg → flac | `audio/flac` |
| `aac` | ffmpeg → AAC (ADTS, 128 kbps) | `audio/aac` |

The transcoder is exposed as `transcode_wav_bytes(wav_bytes, target_format)` plus the supporting `SUPPORTED_TTS_FORMATS` / `TTS_FORMAT_MEDIA_TYPES` constants and a `TranscodeError` exception. `ffmpeg` is invoked via `subprocess` over stdin/stdout; the call runs on a worker thread (`asyncio.to_thread`) so the FastAPI event loop is never blocked.

**`omlx/api/audio_routes.py`** — `create_speech` now:
1. Normalises `response_format` to lowercase and rejects unknown values with HTTP 400 (before loading any model).
2. Still produces WAV from the engine, then pipes through `transcode_wav_bytes` only when the requested format isn't `wav`.
3. Returns the response with the format's correct MIME type.

Defaults are unchanged (`response_format` defaults to `"wav"` as today), so existing clients see no behaviour change.

## Error handling

- Unknown format (`response_format: "wma"` etc.) → `400` with the list of supported values.
[...]

PR #2231, opened 2026-07-14
fix(audio): reject unsupported TTS response_format with 400

### LINKED ISSUE #753: TTS response_format=opus ignored, always returns WAV

### LINKED ISSUE #1013: TTS not working anymore
### Summary

`/v1/audio/speech` accepts a request with `"response_format": "mp3"` and `Accept: audio/mpeg`, but the server returns `content-type: audio/wav` and WAV audio data.

The request eventually succeeds, but the returned format does not match the requested OpenAI-compatible response format.

### Request

```bash
curl -v --no-buffer \
  -H "Authorization: Bearer ollama" \
  -H "Accept: audio/mpeg" \
  -d '{
    "model": "Qwen3-TTS-12Hz-1.7B-CustomVoice",
    "input": "Mit 2 Motoren hast du schon mal den besten Vorteil für enge Manöver. Hier die Basics, damit du die Badeplattform nicht an den Steg knallst:",
    "voice": "Sohee",
    "response_format": "mp3",
    "language": "German"
  }' \
  --output tts-test.mp3 \
  "http://qwen.opnviking.net:8080/v1/audio/speech"
````

### Actual response

```text
HTTP/1.1 200 OK
content-length: 457004
content-type: audio/wav
```

### Expected response

Either:

```text
content-type: audio/mpeg
```

with valid MP3 bytes, or the API should reject unsupported `response_format: mp3` with a clear error.

### Actual behavior

The server returns WAV audio even though the request explicitly asks for MP3.

### Additional observation

The response is only sent after a long delay. In this test, curl waited for more than 2 minutes before receiving headers/body, so native TTS streaming does not appear to provide early audio chunks for this request.

### Suspected issues

* `response_format` may be ignored.
* `Accept: audio/mpeg` may be ignored.
* The endpoint may always return WAV regardless of requested format.
* The new streaming implementation may not actually stream early chunks for this model/path.

JimStenstrom:
Fixes #1013. Fixes #753.

## Summary

- `/v1/audio/speech` now rejects any `response_format` other than `"wav"` (or unset) with a 400 that names the unsupported format, on both the streaming and non-streaming paths.

## Why

The endpoint only produces WAV audio — there is no transcoding — but it silently returned WAV bytes for *any* requested `response_format`. Clients that asked for `mp3`/`opus` got mislabeled audio, exactly as both reports show:

> `"response_format": "opus"` … expected: Ogg data, Opus audio; got: RIFF (little-endian) data, WAVE audio (#753)

> accepts a request with `"response_format": "mp3"` … but the server returns `content-type: audio/wav` and WAV audio data (#1013)

A format guard already existed, but only inside the `stream=True` branch — non-streaming requests (both issues' repros) bypassed it entirely.

## Changes

- `omlx/api/audio_routes.py`: moved the guard to the top of `create_speech` so it covers both the streaming and non-streaming paths; any `response_format` other than `None` or `"wav"` gets a clear 400 naming the format. Removed the now-redundant streaming-only guard, and updated the pre-existing streaming test's docstring to match the unified contract.
- Transcoding to mp3/opus is out of scope here — this makes the contract honest (400 instead of mislabeled bytes) until/unless transcoding is added.

Design note: an alternative is constraining the field as `Literal["wav"]` on `AudioSpeechRequest` (Pydantic 422). I kept the route-level guard because it matches this endpoint's existing ingress checks (the empty-`input` 400 immediately above) and preserves the status code the old streaming guard already used, so no client-visible contract changes shape.

Overlap heads-up: two open April PRs propose actual transcoding for #753 — #818 (wav/pcm via stdlib, mp3/opus/flac/aac via an ffmpeg subprocess, inside a two-fix bundle) and #823 (ffmpeg-backed transcode on a worker thread). This PR takes only the contract-honesty step — an honest 400 instead of mislabeled bytes, zero new dependencies — and is forward-compatible with transcoding landing later (the guard's accepted set simply widens). Happy to close this in favor of either if transcoding is the direction you want.

## Why this is safe

- `AudioSpeechRequest.response_format` defaults to `"wav"`, so requests that omit the field are unaffected — zero behavioral change for existing well-formed usage.
- Streaming requests already got a 400 for non-wav formats; they keep getting one (now with the shared message).

## Tests

- `tests/test_audio_tts.py::TestTTSEndpointBasic::test_post_speech_rejects_unsupported_format` — non-streaming request with `response_format="mp3"` gets 400 naming both the rejected and the supported format; the existing streaming-path test still covers `stream=True`.

```
$ pytest tests/test_audio_tts.py -q
62 passed, 1 deselected
```

Also verified the new test fails when the fix is reverted (endpoint returns 200 with WAV bytes for `response_format="mp3"`).

### DIFF
--- a/omlx/api/audio_routes.py
+++ b/omlx/api/audio_routes.py
@@ -619,13 +619,18 @@ async def create_speech(request: AudioSpeechRequest):
 
     if not request.input or not request.input.strip():
         raise HTTPException(status_code=400, detail="'input' field must not be empty")
+    # Only WAV is produced (no transcoding); reject unsupported response_format
+    # with a clear 400 instead of silently returning WAV. (#753, #1013)
+    if request.response_format not in (None, "wav"):
+        raise HTTPException(
+            status_code=400,
+            detail=(
+                f"response_format '{request.response_format}' is not supported; "
+                "only 'wav' is currently available"
+            ),
+        )
     streaming_interval = DEFAULT_NATIVE_TTS_STREAMING_INTERVAL_SECONDS
     if request.stream:
-        if request.response_format not in (None, "wav"):
-            raise HTTPException(
-                status_code=400,
-                detail="Streaming TTS currently only supports response_format='wav'",
-            )
         streaming_interval = _resolve_tts_streaming_interval(request)
 
     audio_bytes = _decode_ref_audio_base64(request)
--- a/tests/test_audio_tts.py
+++ b/tests/test_audio_tts.py
@@ -197,6 +197,21 @@ class TestTTSEndpointBasic:
         )
         assert response.status_code == 200
 
+    def test_post_speech_rejects_unsupported_format(self, server_tts_client):
+        """Unsupported response_format gets 400, not silently-WAV bytes."""
+        client, _ = server_tts_client
+        response = client.post(
+            "/v1/audio/speech",
+            json={"model": "qwen3-tts", "input": "Hello", "response_format": "mp3"},
+        )
+        assert response.status_code == 400
+        detail = response.json().get("detail") or response.json().get("error", {}).get(
+            "message", ""
+        )
+        # The error must name both the rejected format and the supported one.
+        assert "mp3" in detail
+        assert "wav" in detail.lower()
+
     def test_response_is_audio_bytes(self, server_tts_client):
         """Response body is non-empty bytes."""
         client, _ = server_tts_client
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for the fix. The diagnosis is right, the guard only ever covered the streaming branch, so I'll merge this.

On the transcoding question, I don't want to take on ffmpeg. Turns out soundfile is already in the dependency tree (mlx-audio pulls librosa which pulls soundfile) and its bundled libsndfile can encode mp3/opus/flac in memory, so I'll add actual transcoding on the non-streaming path as a follow-up with no new deps. Your guard fits that nicely, the accepted set just widens. Streaming will stay wav only for now.

PR #2239, opened 2026-07-14
fix: persist serve configuration safely

ShiroKSH:
## Summary

- Persist all non-secret `omlx serve` configuration flags, including SSE, scheduler, and cache settings.
- Keep CLI and environment API-key overrides transient when another CLI setting is saved.
- Restrict credential-bearing settings files to their owner on POSIX systems.

## Root cause

The CLI persistence check and settings mapper had diverged from the public `serve` flags. Saving the merged runtime settings also captured environment overrides and command-line credentials.

## Checks

- `PYTHONUTF8=1 python -m pytest tests/test_cli.py -q -k "not test_serve_exits_on_port_conflict_before_importing_server"` — 60 passed
- Targeted CLI/settings persistence and security tests — 24 passed, 1 POSIX-only test skipped on Windows
- `python -m compileall -q omlx tests`

### DIFF
--- a/omlx/cli.py
+++ b/omlx/cli.py
@@ -38,37 +38,34 @@ def _has_cli_overrides(args) -> bool:
     All argparse defaults are None, so `is not None` means the user
     explicitly passed the flag on the command line.
     """
-    if hasattr(args, "model_dir") and args.model_dir is not None:
+    persisted_fields = (
+        "model_dir",
+        "port",
+        "host",
+        "log_level",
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for fixing this. The explicit-override save path keeps environment and CLI credentials transient while persisting the full public serve configuration, and the POSIX permission handling covers both new and existing settings files. I verified the current main merge path and the focused CLI/settings suite. LGTM.

Issue #2241, opened 2026-07-14
python -m omlx.server crashes — model_dir vs model_dirs kwarg mismatch


Khambampati-Subhash:
## Bug

Running `python -m omlx.server --model-dir /path/to/models` crashes immediately with:

```
TypeError: init_server() got an unexpected keyword argument 'model_dir'
```

## Root Cause

`main()` in `server.py:6805` passes `model_dir=args.model_dir` but `init_server()` at `server.py:1638` expects `model_dirs=` (plural).

Additionally, `main()` passes `pinned_models=` and `max_tokens=` which are not accepted by `init_server()` either.

## Expected Behavior

`python -m omlx.server --model-dir /path` should start the server (matching the documented examples in the argparse epilog).

## Environment

- omlx version: 0.5.1 (current main)
- macOS 26.5.2

jundot:
Fixed by #2242. I also stripped the leftover --pin, --default-model and --max-tokens flags in a follow-up since those are managed via the admin page these days.

jundot:
@Khambampati-Subhash, I released oMLX 0.5.2.dev1 with the module entry-point fix for this issue. Please try the new build and let me know if you see any related issues: https://github.com/jundot/omlx/releases

PR #2242, opened 2026-07-14
fix: correct kwarg mismatch in python -m omlx.server entry point (#2241)

Khambampati-Subhash:
## Summary
`python -m omlx.server --model-dir /path` crashes immediately with:
TypeError: init_server() got an unexpected keyword argument 'model_dir'

`main()` passed `model_dir=`, `pinned_models=`, `default_model=`, and `max_tokens=` to `init_server()`, but `init_server()` only accepts `model_dirs=` (plural), `scheduler_config=`, `api_key=`, and `global_settings=`.
## Fix
Corrected the `init_server()` call to pass `model_dirs=args.model_dir`.

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -6767,14 +6767,9 @@ Note: Use the omlx CLI for full feature support.
     if args.mcp_config:
         os.environ["OMLX_MCP_CONFIG"] = args.mcp_config
 
-    # Parse pinned models
-    pinned_models = args.pin.split(",") if args.pin else []
     # Initialize server
     init_server(
-        model_dir=args.model_dir,
-        pinned_models=pinned_models,
-        default_model=args.default_model,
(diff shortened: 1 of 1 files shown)


jundot:
Thanks, confirmed the crash on main and that this fixes it, the server boots and serves fine through python -m omlx.server now. I'll merge this. One leftover I'll clean up in a follow-up: --pin, --default-model and --max-tokens are still advertised by argparse but silently ignored now, since pinned and default models are managed via the admin page these days.

jundot:
@Khambampati-Subhash, I released oMLX 0.5.2.dev1 with this change. Please try the new build and let me know if you see any related issues: https://github.com/jundot/omlx/releases

Khambampati-Subhash:
Yeah ok

PR #2243, opened 2026-07-14
fix(glm-dsa): deterministic radix top-k output pass (tie membership/order raced)

aidiffuser:
## The bug

`dsa_topk_indices`' non-bucketed output pass assigns slots via bare `atomic_fetch_add`: entries **equal to the threshold key race for the remaining slots**, and output order is scheduling-random. On real GLM-5.2 activations at S=131072 the rank-2048 threshold carries **47–180 exact bf16 ties per query row** (decode steps: mean 122, max 1096), so the selected *set* changed on ~80% of rows across 20 identical re-runs on an M3 Ultra. Consequences we measured in a live deployment (2× M3 Ultra, tensor-parallel):

- temp-0 generations are not byte-reproducible (identical cold 131k prompts → differently-worded answers every run);
- TP replicas that compute the indexer redundantly can silently attend to *different* keys (tie-band divergence).

## The fix

Replace the racy append with a raking prefix-scan (one threadgroup-wide exclusive scan of per-thread counts): strictly-greater keys fill `[0, n_greater)` in index order, threshold ties fill the remainder **lowest-index-first** — the output is a pure function of the scores. The `bucketed` variant is untouched.

## Cost / validation

- One extra read of the score row: 4.69 → 5.56 ms standalone at [4096×131072] (bf16, M3 Ultra) ≈ 0.1% of a 131k prefill chunk.
- Selection fidelity vs fp32 ground truth unchanged (0.9957 mean top-2048 overlap at 131k).
- 20× same-process and cross-process byte-identical outputs (previously ~80% of rows differed).
- End-to-end: two fully-cold identical 131k-token temp-0 runs now produce byte-identical generations; NIAH recall 100% before/after.

This is running in production on our 2-node GLM-5.2 deployment (the same sources, vendored). Happy to adjust style or add a regression test if you'd like one in-tree.

### DIFF
--- a/omlx/custom_kernels/glm_moe_dsa/csrc/kernels/steel_dsa_indexer_score.h
+++ b/omlx/custom_kernels/glm_moe_dsa/csrc/kernels/steel_dsa_indexer_score.h
@@ -16,6 +16,10 @@ METAL_FUNC uint dsa_ordered_key_16(T x) {
   return (bits & 0x8000) ? uint((~bits) & 0xffff) : uint(bits | 0x8000);
 }
 
+METAL_FUNC uint dsa_ordered_key_16_bits(ushort bits) {
+  return (bits & 0x8000) ? uint((~bits) & 0xffff) : uint(bits | 0x8000);
+}
+
 template <typename T, typename O, int TOPK, int THREADS>
 [[kernel, max_total_threads_per_threadgroup(THREADS)]] void dsa_topk_indices_16bit(
     const device T* scores [[buffer(0)]],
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for the focused fix. I reproduced the non-bucketed tie race locally: the baseline produced 20 different raw outputs and 10 different selected sets across 20 runs, while this scan produced the same lowest-index tie selection on every run. The scan ordering and bounds look correct for both top-k instantiations. LGTM. I'll merge this now.

jundot:
@aidiffuser, I released oMLX 0.5.2.dev1 with this change. Please try the new build and let me know if you see any related issues: https://github.com/jundot/omlx/releases

aidiffuser:
Ok will test soon

aidiffuser:
Verified ✅ — `v0.5.2.dev1`'s kernel is byte-identical to the file we've been running in production since the 14th (2× M3 Ultra, tensor-parallel GLM-5.2). All our gates apply to your build: 20× same-process + cross-process byte-identical selections, exact match to the mathematical spec, selection fidelity vs fp32 ground truth unchanged (0.9957), and end-to-end temp-0 byte-identical 131k-token generations. A week of production serving, zero issues. Thanks for the quick merge and release!

PR #2244, opened 2026-07-14
perf(glm-dsa): BK=128 for the 32-head sparse-MLA shard (2 threadgroups/core, ~1.2x per layer-call)

aidiffuser:
Follow-up tuning for the 32-head tensor-shard variant (from our earlier PR): at BK=256 the kernel's ~24KB threadgroup memory allows only **one resident threadgroup per core** on Apple GPUs, so its barrier-separated staging phases serialize with nothing to hide behind.

**Diagnostic that pinned it:** a "samekeys" input (every query row gathers the *same* 2048 keys — perfect cache reuse, ~zero device traffic) runs at the same speed as real inputs → the kernel is not memory-bound; it's occupancy/latency-bound.

**Fix:** BK=128 halves the KV slab (~13KB) → 2 resident threadgroups/core.

| S | BK=256 | BK=128 | note |
|---|---|---|---|
| 131,072 | 100.3 ms | **83.2 ms** | real captured top-k patterns, M3 Ultra |
| 303,104 | 108.8 ms | **85.4 ms** | flat in S ✓ |
| — | BK=64 | 102 ms | over-fragments; not included |

Correctness: outputs differ from BK=256 only in bf16 summation-order rounding — both variants measure **identical** max/mean error vs fp32 ground-truth attention (1.54e-2 / 3.98e-4). End-to-end on our 2-node GLM-5.2 deployment: **+4.5% prefill** on a cold 131,870-token prompt (199.9 vs 191.3 tok/s server-side), needle-recall 100% before/after.

The 64-head single-device variant is left at BK=256 — its occupancy profile differs and we only measured the 32-head shard; the same sweep (and the samekeys diagnostic) may be worth running there too. Happy to gate the choice behind a parameter instead of the H-based default if you prefer.

### DIFF
--- a/omlx/custom_kernels/glm_moe_dsa/csrc/sparse_mla.cpp
+++ b/omlx/custom_kernels/glm_moe_dsa/csrc/sparse_mla.cpp
@@ -165,7 +165,16 @@ class GlmDsaSparseMlaAttentionPrimitive : public Primitive {
     const array& topk_length = has_topk_length ? inputs[5] : topk;
     auto& o = outputs[0];
 
-    constexpr int bk = 256;
+    // BK sets the key-tile size AND the threadgroup-memory footprint (KV slab
+    // ~= BK*(DC+pad)*2B). At BK=256 the 32-head variant uses ~24KB -> only ONE
+    // threadgroup resident per core on Apple GPUs, so its ~500 barrier-separated
+    // staging phases serialize with nothing to hide behind. BK=128 halves the
+    // slab -> 2 resident threadgroups -> measured 100.3 -> 83.2 ms per layer-call
+    // at S=131k (real top-k index patterns, M3 Ultra), 85.4 ms at S=303k; output
+    // differs from BK=256 only in bf16 summation-order rounding (identical error
+    // vs fp32 ground-truth attention). BK=64 over-fragments (102 ms). The 64-head
+    // variant keeps BK=256 (untested at other sizes; its slab already fits 1/core).
+    const int bk = (q_latent.shape(1) == 32) ? 128 : 256;
     constexpr int dc = 32;
     constexpr int d_latent = 512;
     constexpr int d_pe = 64;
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the focused follow-up. The H=32 BK=128 kernel builds and matches the dense fp16/bf16 reference, but I could not reproduce the reported speedup on the same M3 Ultra. Across qL=128..8192 with K=131072 and topk=2048, BK=128 stayed within measurement noise of BK=256, including 128.96 vs 129.50 ms at qL=4096. Could you share the exact harness and captured top-k layout so I can reconcile the 83.2 vs 100.3 ms result before merging?

aidiffuser:
Ran a reconciliation matrix at your exact shape (qL=4096, K=131072, topk=2048). On our M3 Ultra (80-core, macOS 26.5.2), BK=128 wins under **all three** index distributions:

| indices | BK=256 | BK=128 | |
|---|---|---|---|
| real captured layout | 90.2 ms | 79.2 ms | 1.14× |
| uniform random, sorted | 109.3 ms | 98.9 ms | 1.11× |
| uniform random, unsorted | 119.5 ms | 112.8 ms | 1.06× |

so index layout shifts the magnitude but can't explain the null on your machine. The exact harness + the captured top-k tensor you asked for: https://github.com/aidiffuser/glm-dsa-bench (the captured layout is from GLM-5.2 layer 10 at S=131072 on a synthetic corpus — median sorted-gap 2-12, i.e. real selections are highly local).

Since the mechanism is occupancy (the win = 1→2 resident threadgroups/core as the KV slab drops ~24KB→~13KB), the likely reconciliation variables are the Metal toolchain's register/threadgroup-memory accounting or the GPU variant — could you share your core count (60 vs 80) and macOS/Xcode versions? If it still doesn't reproduce on your setup, happy to park this behind a parameter instead of the H-based default — on our box 128 never loses, but your tree, your call.

jundot:
Thanks for the reconciliation matrix. My box is M3 Ultra 80-core, macOS 26.5 (25F71), Xcode 26.5, so the same class as yours, and BK=128 does win here after all: 1.04x to 1.47x across qL 512/2048/4096/8192, four index layouts, both dtypes and three flag combinations, never losing once, and 1.21x to 1.39x at the qL=4096 shape I benchmarked before. Error against fp32 ground truth is identical between the two tile sizes, as you said. My earlier null was my own fault, the extension artifacts are gitignored build products so both arms were almost certainly running the pre-PR kernel.

I also ran the 64-head sweep you offered to run, and the answer is no: BK=128 is 2-9% slower and BK=64 is 10-29% slower at H=64 on every shape I tried. The H-based default is right and there is no need to put it behind a parameter.

One correction I'm folding in before merge. The comment says the 64-head variant keeps BK=256 because its slab already fits one per core, but it works out to ~26KB against the 32-head's ~24KB, so it is in the same situation. The real difference looks like `wm=8` giving 256 threads per threadgroup, which already fills the core, so halving BK only fragments the K tile and adds barrier overhead.

I'll merge this.

PR #2245, opened 2026-07-14
feat(glm-dsa): fused decode indexer scan — single-pass s==1 scoring, +6.7% decode @131k

aidiffuser:
## What

At decode, the DSA indexer scoring chain (`q@k^T → relu → *w → head-sum`) materializes four S-sized intermediates per full layer per token and measures 32-94 GB/s effective. This adds `dsa_decode_scores`: one kernel, one pass over K — thread-per-key, the 32 query heads threadgroup-resident (16B broadcast reads), key rows streamed as `vec<T,4>` float4 dots, **fp32 accumulation with fp32 score output**.

Selection stays on `argpartition` over the fp32 scores — which makes the selected top-2048 match fp32 ground truth **exactly** (overlap 1.0 at S=8k/131k/306k; the existing bf16 chain measures 0.993-0.996). The native 16-bit radix top-k isn't applicable to fp32 scores, hence `argpartition` + your `finish_topk_indices` (sorted-exact convention preserved).

## Numbers (M3 Ultra)

21-layer scan+select, one graph, own K buffers:

| S | chain | fused | |
|---|---|---|---|
| 40,960 | 4.22ms | 2.90ms | 1.46× |
| 133,120 | 7.68ms | 4.56ms | 1.69× |
| 306,176 | 13.91ms | **8.00ms** | **1.74×** |

Live end-to-end on our 2-node GLM-5.2 deployment (streaming steady-state inter-token rate): **+3.4% @40k (16.28→16.83 tok/s), +6.7% @131k (14.32→15.29)**, growing with depth as the scan's share of the token grows. Needle-recall 100% unchanged; 20× byte-deterministic; masked / non-32-head / short-context calls fall through to the existing chain untouched.

Kernel + wiring are running in production on the same sources (vendored). Kernel-shape lessons that got here (scalar tgmem reads: 100 GB/s; simd-per-key + simd_sum: flat; vec4 + register accumulators: wins) are in the commit history if useful.

### DIFF
--- a/omlx/custom_kernels/glm_moe_dsa/csrc/bindings.cpp
+++ b/omlx/custom_kernels/glm_moe_dsa/csrc/bindings.cpp
@@ -52,6 +52,7 @@ NB_MODULE(_ext, m) {
       "queries"_a,
       "keys"_a,
       "weights"_a,
+      "fp32_scores"_a = false,
       "stream"_a = nb::none());
 
   m.def(
--- a/omlx/custom_kernels/glm_moe_dsa/csrc/bindings.cpp
+++ b/omlx/custom_kernels/glm_moe_dsa/csrc/bindings.cpp
@@ -47,6 +47,14 @@ NB_MODULE(_ext, m) {
       "causal_valid_prefix"_a = false,
       "stream"_a = nb::none());
   m.def(
+      "dsa_decode_scores",
+      &omlx::glm_kernels::dsa_decode_scores,
+      "queries"_a,
+      "keys"_a,
+      "weights"_a,
+      "stream"_a = nb::none());
+
+  m.def(
       "glm_dsa_sparse_mla_attention",
       &omlx::glm_kernels::glm_dsa_sparse_mla_attention,
       "q_latent"_a,
--- a/omlx/custom_kernels/glm_moe_dsa/csrc/dsa_indexer.cpp
+++ b/omlx/custom_kernels/glm_moe_dsa/csrc/dsa_indexer.cpp
@@ -353,10 +353,20 @@ class DSATopKIndicesPrimitive : public Primitive {
 // One kernel computes the head-summed indexer scores for a single query position
 // (s == 1) directly into [B,1,1,S] with fp32 accumulation, replacing the decode
 // chain q@k^T -> relu -> *w -> head-sum that materializes four S-sized tensors
-// per layer per token (measured 32-94 GB/s effective; this streams K once).
+// per layer per token. K is addressed by STRIDES: capacity-backed cache slices
+// are consumed in place (no ensure_row_contiguous copy). Scores come out in the
+// input dtype by default (feeding the native 16-bit radix top-k) or fp32 when
+// fp32_scores is set (selection then matches fp32 ground truth exactly).
+struct OMLXDSADecodeParamsHost {
+  int S;
+  int64_t k_batch_stride;
+  int64_t k_row_stride;
+};
+
 class DSADecodeScoresPrimitive : public Primitive {
  public:
-  explicit DSADecodeScoresPrimitive(Stream stream) : Primitive(stream) {}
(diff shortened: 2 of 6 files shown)


jundot:
Thanks for the kernel work. I found that dsa_decode_scores() materializes the capacity-backed KV cache slice through ensure_row_contiguous() on every decode step. On my M3 Ultra, the 21-layer cache-view path went from 1.814 to 3.683 ms at 40,961 tokens, 4.896 to 7.164 ms at 131,073, and 10.729 to 12.464 ms at 306,177. Even with contiguous K, the current 4096-token gate is 31% slower at 40,960 because the existing path already uses native dsa_topk_indices. Could you pass the K strides into the Metal kernel, remove the full-cache copy, and rebenchmark against the current native path?

aidiffuser:
Both findings addressed in the new commit — you were right on both, thanks for the sharp review.

1. **K is now addressed by strides** (batch + row; rows 16B-aligned, last dim contiguous): capacity-backed cache slices are consumed in place, the `ensure_row_contiguous` copy is gone, and view-vs-contiguous outputs are byte-identical.
2. **Scores now come out in the cache dtype by default and feed your existing `select_topk`/native radix unchanged** (deterministic as of #2243) — the fp32+argpartition selection you measured 31% behind is now an opt-in (`fp32_scores=True`) for callers who want selection to match fp32 ground truth exactly.

Rebenched against the current native path (chain + `dsa_topk_indices`), 21 layers, capacity-backed K views — your baseline reproduces on our box (1.84 ms @ 40,961 vs your 1.814):

| S | current native path | fused (strided, cache-dtype scores) | |
|---|---|---|---|
| 40,961 | 1.84 ms | **1.00 ms** | 1.83× |
| 131,073 | 5.72 ms | **2.97 ms** | 1.92× |
| 306,177 | 11.30 ms | **5.11 ms** | 2.21× |

It now wins at every depth, so the 4096-token gate is kept as-is. And a genuine thank-you: the strided fix applies to our own deployment too — your review found us free performance.

aidiffuser:
One honest data point from our side, for completeness: we back-ported this v2 (strided K + cache-dtype scores → native radix) into our own exo-based deployment and ran a four-arm end-to-end A/B (same launcher, same day, cold caches, streaming steady-state decode). Result: **live-neutral on our serving path** — v1-with-copy 14.79 tok/s vs v2 variants 14.64-14.76 @131k, all within our ±2% noise band. In our runtime the isolated-bench copy tax and selector deltas disappear under graph-execution shadows.

That does **not** argue against this PR: your runtime demonstrably pays the copy (your 1.81→3.68 ms cache-view numbers), so the strided fix stands on your measurements — it just means the isolated kernel wins may not translate 1:1 to every serving loop. If you get a chance, an end-to-end decode check in omlx's own serving path (rather than the kernel-only harness) would be the definitive number for the selector choice. Happy to help bench if useful.

jundot:
Ran the end-to-end check in omlx's own serving path. Both fixes hold up: view vs contiguous K comes out byte-identical, and the fused scores are tighter against fp32 ground truth than the chain they replace (1.8e-3 vs 4.6e-3 at S=8192). Your table reproduces here too, 1.70x / 1.93x / 2.03x at 40961 / 131073 / 306177, and emulating v1 lands at 1.06x to 1.13x, which is exactly the copy tax you removed.

Live decode on GLM-5.2 through omlx serve, single stream, is +1.7% at 30k and +3.1% at 125k, with every ON sample above every OFF sample at 125k. Lower than your +6.7% but real, and growing with context the way you described.

Two limits worth recording, neither of them a problem with the PR. The gate is `s == 1`, so with MTP on the trunk verify runs at s = depth+1 and all 21 trunk indexer layers skip the kernel, leaving only the draft head's own call. And at two or more concurrent requests the cache becomes a `BatchKVCache`, whose `make_mask` returns an array even at N=1, so the `mask is None` gate drops out. So the win lands on single-stream non-speculative decode.

I'll merge this.

PR #2256, opened 2026-07-15
fix(cache): probe with the model's resolved chat template kwargs

thornad:
The admin cache probe (`POST /admin/api/cache/probe`) reports every block of every prompt as cold for any model that has a thinking toggle set, while the SSD cache is being written and hit normally. Cache-status UIs built on the probe show an empty cache.

## Cause

The probe renders the prompt with the caller's `chat_template_kwargs` alone. `/v1/chat/completions` merges the model's own settings first:

1. persisted `chat_template_kwargs`
2. the dedicated `enable_thinking` / `preserve_thinking` toggles
3. per-request kwargs, except `forced_ct_kwargs`

The probe skipped steps 1 and 2, so it rendered a prompt no request ever produces. The probe exists to answer "is this prompt cached", which only works if it renders byte-for-byte what a real turn renders.

## Measured

GLM-5.2-oQ4e-mtp with `thinking_default=False`, same message list each time:

| | tokens | result |
|---|---|---|
| generation (`/v1/chat/completions`) | 1095 | prefills, writes cache |
| probe, as shipped | 1101 | `blocks_cold: 5`, `ssd_hit_tokens: 0` |
| probe, `enable_thinking=False` passed by hand | 1095 | 4/5 blocks cached |
| probe, after this patch (no manual kwargs) | 1095 | 4/5 blocks cached |

Six tokens at the front changes the first block's chain hash. The walk stops at the first retrievability miss, so every later block is reported cold too — which is why it presents as a uniformly empty cache rather than a partial miss.

Confirmed against a live server: the same generation reports `prompt_tokens_details.cached_tokens: 1024`, i.e. the scheduler hit the same 4 blocks the probe now reports. The 1 remaining cold block is correct — the trailing partial block is never persisted (`Skipping trailing partial block: 72 token(s) not persisted`).

Reproduces on any model whose template has an `enable_thinking` switch; not model- or quantization-specific.

## Fix

The merge contract was spelled out in three places — chat completions, Anthropic messages, and the probe — and the probe's copy is the one that drifted. Extracted into `model_settings.merge_chat_template_kwargs()` and routed all three through it, so a future change cannot fix one renderer and miss another. The two `server.py` call sites are behaviour-preserving refactors; `forced_ct_keys()` is split out because the Anthropic path still needs the forced-key set for its own `thinking` handling.

A settings lookup failure in the probe falls back to the caller's kwargs (pre-fix behaviour) rather than failing the probe outright.

## Tests

`TestCacheProbeChatTemplateKwargs` covers each precedence rule at the probe boundary: model toggles reach the template, persisted kwargs reach it, request kwargs override model settings, forced keys win over request kwargs, a lookup failure degrades gracefully, and the probe agrees with the shared resolver. All five behavioural assertions fail on the pre-fix probe and pass after.

`tests/test_admin_cache_probe.py`, `test_model_settings.py`, `test_server.py`, `test_config.py`, `test_engine_pool.py`, `test_external_api.py`, `test_grammar.py`: 398 passed, 3 skipped. No new ruff findings on the touched files.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -34,6 +34,7 @@ from ..api.markitdown import MARKITDOWN_MODEL_ID, markitdown_model_visible
 from ..api.openai_models import _coerce_tool_call_arguments
 from ..api.utils import _try_parse_json
 from ..model_profiles import EXCLUDED_FROM_PROFILES
+from ..model_settings import merge_chat_template_kwargs
 from ..settings import BURST_DECODE_MODES, SubKeyEntry, burst_decode_env
 from ..utils.release_check import normalize_update_channel, select_latest_release
 from .auth import (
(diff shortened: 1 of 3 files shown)


thornad:
Follow-up in #2257 migrates `/v1/chat/completions` and `/v1/messages` onto the same resolver, removing the last two hand-maintained copies of this merge. Kept separate so this fix stays scoped to the admin endpoint.

thornad:
Note: `merge_chat_template_kwargs()` has one caller here (the probe). `/v1/chat/completions` and `/v1/messages` still carry their own copies of the same merge — deliberately left alone so this stays scoped to the admin endpoint. Happy to migrate them onto the helper as a follow-up if you want it; it comes to +4/-32 in `server.py` with no behaviour change.

jundot:
Thanks for tracking this down. The persisted, dedicated, and request-level precedence matches the existing generation merge, and the fallback keeps the admin endpoint usable if settings lookup fails. The later thinking-budget and preserve_thinking_default auto-injection still needs a follow-up for full byte-for-byte parity, but it does not block this fix. LGTM. I'll merge this and handle that remaining consistency gap separately.

jundot:
@thornad, I released oMLX 0.5.2.dev1 with this change. Please try the new build and let me know if you see any related issues: https://github.com/jundot/omlx/releases

PR #2257, opened 2026-07-15
refactor: route chat completions and Anthropic messages through the shared kwargs resolver

thornad:
> **Stacked on #2256 — merge that first.** This branch contains #2256's commit as its base, so the diff below shows both. Only the second commit (`refactor: route chat completions...`) belongs to this PR. Once #2256 lands, this rebases to a single 4-line-net change in `server.py`.

## Why

`/v1/chat/completions` and `/v1/messages` each spell out the `chat_template_kwargs` merge by hand — persisted kwargs, then the `enable_thinking` / `preserve_thinking` toggles, then per-request kwargs except `forced_ct_kwargs` — in blocks byte-identical to each other and to the copy the admin cache probe used to carry.

That third copy is what caused #2256: the probe's version omitted the model settings entirely, so it hashed a prompt the scheduler never prefills and reported every block cold while the cache was being written and hit normally. Three hand-maintained copies of one contract is how that happens, and nothing stops the next divergence.

## What

Point both endpoints at `model_settings.merge_chat_template_kwargs()`, added in #2256 and already used by the probe. `server.py` loses 32 lines and gains 4.

Behaviour is unchanged — same precedence, same forced-key handling. `forced_ct_keys()` stays exposed because the Anthropic path still needs the forced-key set for its own `thinking` config, which is applied after the merge.

## Risk

This touches the hot chat-completions path to consolidate logic, so it is deliberately separate from the fix in #2256 — that one is a read-only admin endpoint and can land on its own merits.

No behavioural change, so no new tests. `test_server.py`, `test_admin_cache_probe.py`, `test_model_settings.py`, `test_external_api.py`: 165 passed. `test_probe_and_chat_path_resolve_identical_kwargs` (added in #2256) now holds by construction rather than by coincidence.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -34,6 +34,7 @@ from ..api.markitdown import MARKITDOWN_MODEL_ID, markitdown_model_visible
 from ..api.openai_models import _coerce_tool_call_arguments
 from ..api.utils import _try_parse_json
 from ..model_profiles import EXCLUDED_FROM_PROFILES
+from ..model_settings import merge_chat_template_kwargs
 from ..settings import BURST_DECODE_MODES, SubKeyEntry, burst_decode_env
 from ..utils.release_check import normalize_update_channel, select_latest_release
 from .auth import (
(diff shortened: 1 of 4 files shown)


thornad:
Closing — this is optional cleanup, not needed for the fix in #2256, and it touches the hot chat-completions path for zero behaviour change.

To correct the framing in my own description: I said the probe's copy "drifted" from the others. It didn't — the probe never implemented the merge at all. That was an omission, not divergence, so consolidating the remaining copies would not have prevented it; nothing forces a caller to use the helper.

The branch stays on my fork if you ever want the two `server.py` copies migrated onto `merge_chat_template_kwargs()` — it rebases to +4/-32 once #2256 lands. Happy to reopen on request.

PR #2258, opened 2026-07-16
fix(cache): hoist nstate closures to module level to break KV refcount cycle

bitfieldz:
## Problem

Resident 4D KV cache arrays accumulate across multi-turn sessions and are never freed while the server is idle (`active_requests: 0`). Over enough traffic this reaches 10GB+, and the next long-context prefill's transient peak tips the process over its memory ceiling.

## Root cause

`PagedSSDCacheManager._reconstruct_cache_data.<locals>._load_nstate` (read path) and `save_block.<locals>._store_nstate_elements` (write path) are recursive nested closures. A recursive nested function is captured by its own closure cell, forming a self-referential reference cycle. The cycle also holds the enclosing `arrays` dict — hundreds of MB of KV tensors per call. Reference counting cannot reclaim a cycle; only a generational GC pass can. A low-allocation idle server rarely triggers a gen-2 collection, so the cycles — and the KV they pin — accumulate unbounded.

Trigger condition: any hybrid model using ArraysCache boundary snapshots with SSD-backed prefix cache, over enough multi-turn traffic that gen-2 gc rarely fires.

## Fix

Hoist both functions to module level, passing `arrays`/`file_metadata`/`cache_list_meta` as explicit arguments. No closure, no cycle; freed by refcount when the call returns. Logic is unchanged — the inner names remain as one-line shims so call sites are untouched.

## Verification

Six-conversation multi-turn soak: peak resident 4D KV **0.43 GB**, draining to **0.00 GB** between conversations (previously 6.5 GB+ per conversation, never released). `gc.DEBUG_SAVEALL` after the run reports **zero** cache functions in `gc.garbage` (30+ instances of each before the fix).

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -358,6 +358,143 @@ def _decode_shape(shape_str: str) -> tuple:
     return tuple(int(d) for d in shape_str.split(","))
 
 
+def _store_nstate_elements_flat(
+    arrays: dict[str, Any],
+    cache_list_meta: dict[str, str],
+    prefix: str,
+    elements,
+) -> None:
+    """Write N elements as ``{prefix}_state_{k}`` keys with a
+    ``{prefix}_state_count`` count marker. Zero-dim shapes are
+    preserved via ``{prefix}_state_{k}_zero_dim``. Composite
+    elements (a bare tuple/list or a nested ``__nstate__``
+    marker) recurse under a ``{elem_key}`` sub-prefix and record
+    a ``{elem_key}_nested`` marker; the flat ``{elem_key}``
+    tensor is deliberately omitted so an older reader hits its
+    ``Missing {elem_key} in arrays`` path and skips the block.
+
+    Module-level on purpose — same recursive-closure cycle bug as
+    ``_load_nstate_flat`` (see that docstring).
+    """
+    cache_list_meta[f"{prefix}_state_count"] = str(len(elements))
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for tracking this down. I reproduced both recursive closure cycles with cyclic GC disabled: the current code retains the per-call arrays until `gc.collect()`, while this change releases them immediately, and the existing n-state and V4 round-trip tests still pass. LGTM. I'll merge this.

jundot:
@Spirotot, I released oMLX 0.5.2.dev1 with this change. Please try the new build and let me know if you see any related issues: https://github.com/jundot/omlx/releases

bitfieldz:
> @Spirotot, I released oMLX 0.5.2.dev1 with this change. Please try the new build and let me know if you see any related issues: https://github.com/jundot/omlx/releases

@jundot , it's been stable with max concurrent sessions pinned to 1 ever since I applied this patch (and I'm running the 0.5.2.dev1 tag now); I can now run oMLX at full-tilt, without issue, indefinitely... whereas before it'd inevitably run out of memory and stall at least until model TTL unloaded the model. So I'm very happy now. :) Thanks for merging! 

Still seems like maybe there are some memory "spikes" during the prefill stages which I want to try to run down and see if they can be smoothed a bit -- might make concurrent streams a bit more stable for me -- TBH not sure if there are actually any throughput gains for me with more concurrent streams anyway (M1 Max 64 GB user, here), but will be a learning opportunity if nothing else!

PR #2265, opened 2026-07-16
fix(server): give python -m omlx.server a working admin settings path (#2282)

### LINKED ISSUE #2282: Question: should `python -m omlx.server` be fixed, or is `omlx serve` the only supported entry point?
**Describe the bug**
While running the server directly with `python -m omlx.server --model-dir <dir>`
(the invocation shown in `main()`'s own argparse examples), opening `/admin` and
submitting the initial API-key setup form crashes with a 500:

**To Reproduce**
1. Run `python -m omlx.server --model-dir <dir>`
2. Open /admin in chrome and type api key you will see the error

**Root cause**
`main()` calls `init_server(model_dirs=...)` without `global_settings`, so
`_get_global_settings()` returns `None`. The `omlx serve` CLI (`cli.py`) instead
passes `global_settings=settings`, which is why the CLI path works and only the
`python -m` path breaks.
`setup_api_key` already guards for `None` at line 1524
(`if global_settings and global_settings.auth.api_key:`) but then dereferences it
unconditionally at line 1540.

**Question / need guidance**
`main()`'s docstring says *"use omlx CLI instead"*, so I'm not sure this path is
even meant to be supported. Before opening a PR I'd like maintainer direction on
which of these you'd accept:
1. **Defensive guard only** — return a clean 503 from `setup_api_key` when
   `global_settings is None` (small, but the `python -m` admin flow still can't
   configure a key).
2. **Fully wire up `main()`** — build `GlobalSettings` there like `cli.py` does
   (makes `python -m` a real entry point, but duplicates CLI setup).
3. **Make `main()` refuse** — print "use `omlx serve`" and exit, so the module
   never boots a half-initialized server.
4. **Won't fix** — `python -m omlx.server` is not a supported entry point; leave
   as-is.
Happy to submit whichever direction you prefer. Candidate patch for option 1:
    # in setup_api_key, right after `global_settings = _get_global_settings()`
    if global_settings is None:
        raise HTTPException(
            status_code=503,
            detail=(
                "Server settings are unavailable, so the API key cannot be "
                "saved. Start the server via the omlx CLI (`omlx serve`)."
            ),
        )

**Environment**
- omlx `0.5.2.dev1` (main)
- macOS 26.5.2, Python 3.11

richgoodson:
Fixes #2282.

This PR originally fixed the #2241 TypeError in this entry point. #2242 landed the same kwarg correction first, so I have rebased and retargeted the branch at what is still broken in `python -m omlx.server`: the admin page cannot configure an API key, which is #2282 (and, per the issue title, the open question of whether this path should work at all; this PR is the reporter's option 2).

## Root cause: two layers

**Layer 1, the one in the issue report.** `main()` calls `init_server()` without `global_settings`, so `_get_global_settings()` returns `None` inside `setup_api_key` and the write at `routes.py:1540` (`global_settings.auth.api_key = ...`) raises `AttributeError` on `None`, a 500 for every initial API-key setup.

**Layer 2, which the report could not see.** Wiring settings into `main()` is not enough. `python -m omlx.server` executes `server.py` as the `__main__` module. When the first admin request arrives, `setup_api_key` runs `from ..server import _server_state`, which imports the file a second time under its canonical name `omlx.server`. That second execution runs the module-level `set_admin_getters(...)` again and repoints the admin state getters at the fresh copy's `_server_state`, which `init_server()` never touched. The uvicorn server keeps serving the `__main__` app while `/admin` reads settings from a parallel universe. I verified this live: with only the layer-1 fix applied, the reporter's repro still returns 500.

## The fix

- `main()` now loads settings via `init_settings()` and passes `api_key` and `global_settings` to `init_server()`, the same shape as the `omlx serve` path in `cli.py`. Scheduler and cache wiring stay CLI-only on purpose; the module entry point remains minimal.
- The `__main__` block aliases `sys.modules["omlx.server"]` to the running instance before serving, so every later import resolves to the module uvicorn is actually serving. `setdefault` keeps an already-imported canonical module if one ever exists.
- `main()` also mirrors the `mx.set_cache_limit(total_mem)` guard from `cli.py` (issue #300: allocator::free() can release a Metal buffer the GPU is still using, kernel panics on M4; EnginePool eviction assumes the limit is set). Happy to split this into its own PR if it is scoped too wide here.

## Tests

`tests/test_server_main.py`, four tests:

- Three in-process tests run `main()` against the real `init_server` (uvicorn and the cache-limit side effect stubbed): entry point boots, `_server_state.global_settings` is wired, and the setup-api-key endpoint succeeds via TestClient.
- One subprocess test boots a real `python -m omlx.server` on a free port and submits the setup form end to end. This is the only kind of test that can catch layer 2; every in-process test has a single module instance by construction and is structurally blind to the double execution.

Red/green: the subprocess test fails on current main (500) and passes with the fix.

## Live verification (M5 Air, main @ 4690cbda)

Reporter's literal repro, `python -m omlx.server --model-dir <dir>` then submitting the initial API-key form:

- unpatched main: HTTP 500, `AttributeError: 'NoneType' object has no attribute 'auth'` at `routes.py:1540`
- layer-1 fix only: still HTTP 500, same traceback, via the second module instance
- full fix: HTTP 200, `{"success": true}`, key persisted to `settings.json`, session cookie set

Full test suite run locally on the branch; only the pre-existing M5 failures unrelated to this change (#2267 and the two GLM MTP SmallLRouting tests).

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -6790,9 +6790,33 @@ model and sampling defaults are managed via the admin page.
     if args.mcp_config:
         os.environ["OMLX_MCP_CONFIG"] = args.mcp_config
 
+    # Load settings and hand them to init_server the way the omlx CLI
+    # does. The admin page resolves settings through
+    # _server_state.global_settings, so booting without them leaves
+    # _get_global_settings() returning None and the initial API-key
+    # setup form crashing with a 500 (#2282). Passing api_key keeps the
+    # two entry points enforcing the same auth. Scheduler/cache wiring
+    # stays CLI-only on purpose; this entry point remains minimal.
+    from .settings import init_settings
+
+    settings = init_settings()
+    settings.ensure_directories()
+
+    # Match the cli.py launcher: keep freed GPU buffers in the pool so
+    # allocator::free() never releases a buffer the GPU may still be
+    # using (kernel panics on M4 otherwise; see cli.py and issue #300).
+    # EnginePool eviction also assumes this cache limit is in place.
+    import mlx.core as mx
+
+    total_mem = mx.device_info().get("memory_size", 0)
+    if total_mem > 0:
+        mx.set_cache_limit(total_mem)
+
     # Initialize server
     init_server(
         model_dirs=args.model_dir,
+        api_key=settings.auth.api_key,
+        global_settings=settings,
     )
 
     # Start server
@@ -6802,4 +6826,15 @@ model and sampling defaults are managed via the admin page.
 
 
 if __name__ == "__main__":
+    # ``python -m omlx.server`` executes this file as the ``__main__``
+    # module, but the admin routes import it back as ``omlx.server`` at
+    # request time. Without this alias that import executes the module a
+    # SECOND time, and the fresh copy's module-level set_admin_getters()
+    # call repoints the admin state getters at a server state that
+    # init_server() never touched, so the settings main() wires in are
+    # invisible to /admin (#2282). Alias the canonical name to this
+    # instance so every later import resolves to the running server.
+    import sys
+
+    sys.modules.setdefault("omlx.server", sys.modules[__name__])
     main()
--- a/tests/test_server_main.py
+++ b/tests/test_server_main.py
@@ -1,0 +1,153 @@
+"""Regression tests for the ``python -m omlx.server`` entry point.
+
+This path has broken twice: #2241 (main() passed kwargs init_server() no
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for chasing this past the obvious layer, the sys.modules alias is the right call and the subprocess test is exactly the kind of guard this entry point was missing. I reproduced your red/green locally, the e2e test 500s on current main and passes on the branch. I'll merge this.

PR #2269, opened 2026-07-16
launch/claude: mark argv[0] so external tools can identify omlx-launched sessions

fparrav:
## Summary
- `launch claude` now execs with `argv[0] = "claude-omlx"` instead of the real binary name/path.
- Claude Code does not inspect its own `argv[0]` (verified `--version`/`--print` behave identically either way), so this has no effect on Claude Code's behavior.
- It gives external tooling that reads `ps`/cmdline (session managers, process monitors, terminal multiplexer session-resurrection features) a zero-cost way to identify that a running `claude` process was launched through omlx, which is otherwise impossible once `execvpe` replaces the omlx process - the env vars this function sets to route Claude Code at the local server are invisible to anything inspecting the process from outside.

## Test plan
- [x] `python3 -c "import os; os.execvpe('claude', ['claude-omlx', '--version'], os.environ.copy())"` still runs the real claude binary and prints the version correctly.
- [x] `ps` shows `claude-omlx` in the command column for a process launched this way, confirming external tools can key off it.

### DIFF
--- a/omlx/integrations/claude.py
+++ b/omlx/integrations/claude.py
@@ -66,7 +66,14 @@ class ClaudeCodeIntegration(Integration):
             env["CLAUDE_CODE_AUTO_COMPACT_WINDOW"] = str(ctx.context_window)
 
         binary = self._find_claude_binary()
-        argv = [binary, *ctx.extra_args]
+        # argv[0] is cosmetic to the OS/claude itself (verified: claude does not
+        # inspect its own argv[0]) but is what tools like `ps` and zellij's
+        # session-resurrection command discovery see. Marking it lets external
+        # tooling tell an omlx-launched session apart from a bare `claude`
+        # invocation or one launched through a different wrapper (e.g. ollama's
+        # `--model` flag), which is otherwise impossible once execvpe replaces
+        # this process.
+        argv = ["claude-omlx", *ctx.extra_args]
         print(f"Launching Claude Code with model {ctx.model}...")
         if ctx.context_window:
             print(f"Auto-compact window: {ctx.context_window:,} tokens")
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the PR, the verification here is careful work. I'm just not sure this is something omlx really needs right now. If I take this it effectively becomes an omlx standard, so every other integration (codex, copilot, opencode, and so on) would need the same treatment to stay consistent, and that's ongoing maintenance for something no tool is actually asking for yet. Curious what you think though, did you have a concrete tool or workflow in mind that needs this marker?

fparrav:
Fair question, and fair concern about it becoming a de facto standard.

The concrete case: I run Claude Code inside zellij, and zellij's session-resurrection feature works by inspecting the running process's argv/cmdline to figure out what command to re-launch when a pane is restored. Without the marker, it can't tell an omlx-launched session (pointed at a local model, with its own context window/env config) apart from a plain `claude` invocation — so on resurrection it just re-runs bare `claude`, and the session silently loses the local model it was actually using. With `claude-omlx` in argv[0], zellij can at least distinguish the two cases and, eventually, know it needs to relaunch through omlx rather than plain Claude Code.

To be clear, this doesn't touch the natural flow at all: Claude Code itself ignores its own argv[0] (verified `--version` and `--print` behave identically regardless of what's passed), so this is purely a signal for external tooling — nothing changes in how Claude Code runs or behaves.

I actually tested tmux to check the same theory, and it does *not* have this problem — tmux-resurrect/continuum rely on tmux's `pane_current_command`, which tmux resolves from the kernel's process `comm` field rather than argv/cmdline, so it's unaffected by this marker either way.

This is also why I only touched `claude.py` — my own workflow is Claude Code exclusively, and I don't have a concrete need (or evidence of one) for codex, copilot, or opencode integrations to carry the same marker. I'm not proposing this as a cross-integration convention, just a scoped fix for the one wrapper I actually use inside zellij. Happy to add a comment in the code making that scope explicit if it helps.

PR #2271, opened 2026-07-17
fix(deepseek_v4): preserve cross-window overlap during decode for ratio=4 compressors

### LINKED ISSUE #1345: DeepSeek-V4-Flash-mxfp8 unexpectedly switches from Simplified Chinese to Traditional Chinese without any system prompt
### Summary

When using `DeepSeek-V4-Flash-mxfp8` in oMLX, the model unexpectedly switches from Simplified Chinese to Traditional Chinese during a normal Chinese conversation.

There is **no custom system prompt** and no instruction asking the model to use Traditional Chinese. The user input is in Simplified Chinese, and the model initially reasons in Simplified Chinese, but then the reasoning content and final output drift into Traditional Chinese.

### Environment

- oMLX version: `v0.3.9`
- Model/profile: `DeepSeek-V4-Flash-mxfp8`
- System prompt: **None**
- User prompt language: Simplified Chinese
- Time observed: around 03:52
- Thinking enabled: Yes

### Steps to Reproduce

1. Start a chat with `DeepSeek-V4-Flash-mxfp8`.
2. Do not set any system prompt.
3. Send the following Simplified Chinese prompt:

```text
你好，写一篇小说
````

4. Observe the model output.

### Actual Behavior

The model starts in Simplified Chinese, for example:

```text
好的，用户要求写一篇小说，并提供了一个具体的梗概……
```

But later in the reasoning content, it switches into Traditional Chinese / mixed Chinese:

```text
想到了用第一人称“我”來敘述，增強代入感和懸疑感。
可以從日常清晨的溫馨場景寫起，逐步滲入不安……
```

The final generated story is also mostly Traditional Chinese:

```text
清晨六點四十七分，林嶼醒來的時候，身旁的被子已經涼了。

窗簾拉開一條窄縫，陽光斜進來，恰好落在床頭櫃上那杯溫水旁邊。
水杯底下壓著一張便條，字跡俐落——「早餐在鍋裡，今天別再忘了吃。」
```

### Expected Behavior

Since the user prompt is in Simplified Chinese and no system prompt or user instruction requests Traditional Chinese, the model should consistently respond in Simplified Chinese.

Expected output style should be like:

```text
清晨六点四十七分，林屿醒来的时候，身旁的被子已经凉了。

窗帘拉开一条窄缝，阳光斜进来，恰好落在床头柜上那杯温水旁边。
水杯底下压着一张便条，字迹利落——“早餐在锅里，今天别再忘了吃。”
```

### Why This Matters

This is problematic for Simplified Chinese users because the model unexpectedly changes script without being asked. It also affects consistency in long-form writing, agent workflows, and downstream post-processing where Simplified Chinese output is expected.

### Notes

The issue happens without any system prompt. It may be related to:

* model default language behavior,
* tokenizer/chat template behavior,
* reasoning content handling,
* DeepSeek V4 Flash specific output formatting,
* or oMLX profile/default preset configuration.

### LINKED ISSUE #1873: Simplified Chinese comes to Traditional Chinese.
**Describe the bug**
The issue of DEEPSEEK-V4-FLASH switching from Simplified Chinese output to Traditional Chinese output still has not been resolved. This prevents me from modifying documents properly. I can read Traditional Chinese myself, but I cannot directly send the final product to external parties. It means it can not be used for real works in Chinese.

**To Reproduce**

1. Test by writing a novel. What is written is not important; writing a novel is only for testing.
2. After a few hundred TOKENS, Simplified Chinese changes into Traditional Chinese.

**Expected behavior**
Simplified Chinese should remain Simplified Chinese. 
**Test environment**
Tested in both OpenWebUI and oMLX's own chat tool.

* macOS Version: 26.5.1
* oMLX Version: 0.4.4rc1

随便写一篇小说

profile
deepseek-v4-flash
今天 14:41
思考用时 9 秒
用户想要一篇小说，并且提供了一个梗概。这个梗概本身已经很有戏剧性和冲突感，包含了时间穿越、身份错位、阶级差异和情感纠葛等元素。用户可能希望看到一个基于这个核心设定的、有血有肉的故事，而不仅仅是梗概的简单扩写。

我需要把梗概中的几个关键节点展开，构建一个完整的故事弧光。可以从林晚晚的“死亡”现场开始，制造悬疑和冲击感，然后自然过渡到回忆，交代她与陈渡、陈序之间的复杂关系是如何形成的。这样能先建立读者对人物的共情。回忆部分要详细描写她如何进入陈家、与兄弟俩的互动，以及关键事件（生日宴、出逃、被抓回）是如何一步步发生的，让悲剧的合理性逐步显现。

然後回到“重生”的当下，聚焦于林晚晚心态的转变——从绝望到决绝，再到伪装成“林小姐”接近陳序的復仇計劃。這部分要展現她的成長和冷酷，與回憶中的柔弱形成對比。最後的結尾可以處理得開放一些，比如她坐在車裡，妝容精緻，眼神冰冷，暗示復仇即將開始，這樣能留給讀者想像和後續發展的空間。整體語言可以保持冷靜、平實的敘述，用細節和動作來傳達情緒，避免過度的抒情渲染。 

林晚晚死的那天，陳家正在辦慶功宴。

陳序南拿下了一個二十億的項目，陳家上下喜氣洋洋，香檳開了一瓶又一瓶，笑聲幾乎掀翻了宴會廳的天花板。

williamxie1989:
## Summary

- **Fix Simplified→Traditional Chinese drift** in DeepSeek V4 Flash during long-generation decode by preserving cross-window overlap in ratio=4 (`SparseCompressedAttention`) compressors
- Root cause: `_overlap_compress_kv` relies on `kv_a[:, :-1]` to carry the previous window's lane-A into the current window's pooling. During decode (single-window input), this shift produces an empty array, collapsing lane-A to zero-padding and losing cross-window overlap entirely

### The Bug

DS4 Flash uses **overlap compression** (ratio=4) where each pooled token is softmax-pooled from **8 raw tokens** (2×ratio), not 4. This is achieved by splitting KV along `head_dim` into two lanes:
- **lane-A** = previous window's first half (via `kv_a[:, :-1]` window-axis shift)
- **lane-B** = current window's second half

During **prefill** (multiple windows), the shift works correctly — lane-A carries real data from the previous window.

During **decode** (single window), `kv_a[:, :-1]` is empty → lane-A becomes zero-padding → overlap degenerates from 8-token pool to 4-token pool. This is a **structural/logical difference** from DS4, not a precision issue.

### Evidence

| Evidence | Result |
|----------|--------|
| Same 2-bit model, same context: prefill vs decode logits | Prefill → Simplified +12; Decode → Traditional +3.68 |
| fp32 full-activation test | Still Traditional — rules out precision |
| DS4 (Q2/Q4/8-bit) | Never drifts — confirms MLX-specific |
| First ~374 tokens OK, then drift | Cumulative logit drift across 21 layers |
| SwiGLU clamp | Correctly implemented — ruled out |
| RoPE positions | Consistent between modes — ruled out |
| make_mask causal boundary | Static analysis shows correct behavior — ruled out |

### The Fix

Carry `prev_win_kv` / `prev_win_gate` in `PoolingCache` / `BatchPoolingCache`. In `Compressor.__call__`, prepend the previous window to the current chunk before calling `_overlap_compress_kv`, so it sees two windows and the cross-window shift works. Take `[:, 1:]` output to discard the prepended window's own (zero lane-A) pooled result.

This mirrors DS4's `compressor_decode_one` rolling double buffer semantics without modifying the `@mx.compile`-decorated `_overlap_compress_kv` kernel.

### Files Changed

| File | Change |
|------|--------|
| `omlx/patches/deepseek_v4/cache_extras.py` | Added `prev_win_kv`/`prev_win_gate` to `PoolingCache` and `BatchPoolingCache`; extended `_undo` tuples; updated `trim`/`state`/`filter` |
| `omlx/patches/deepseek_v4/deepseek_v4_model.py` | `Compressor.__call__`: prepend prev_win, run compression, take `[:, 1:]`, roll prev_win |

### Verification

400-token greedy generation (temp=0, prompt: "介绍下量化交易"):
- **Before**: Token 375 flips to Traditional Chinese (挑戰/過擬合/歷史數據/實盤)
- **After**: All 400 tokens remain Simplified Chinese (挑战/过拟合/历史数据/实盘), output quality normal

---

Fixes #1873
Fixes #1345

### DIFF
--- a/omlx/patches/deepseek_v4/cache_extras.py
+++ b/omlx/patches/deepseek_v4/cache_extras.py
@@ -35,6 +35,17 @@ class PoolingCache(_BaseCache):
         self.pooled = None
         self._undo = None
 
+        # Previous completed window's raw (pre-compression) KV/gate. Only
+        # used by overlap (ratio==4) compressors: decode hands a single
+        # window to _overlap_compress_kv, whose `kv_a[:, :-1]` lane-A shift
+        # collapses to zero-padding and drops the cross-window overlap that
+        # native DS4 keeps via a rolling double buffer (ds4.c
+        # compressor_decode_one). Carrying the last window here lets the
+        # Compressor prepend it so decode preserves the overlap. Simple
+        # (ratio==128) layers leave these None.
+        self.prev_win_kv = None
+        self.prev_win_gate = None
+
     @property
     def offset(self):
         return 0 if self.pooled is None else self.pooled.shape[1]
@@ -62,6 +73,8 @@ class PoolingCache(_BaseCache):
                 self.pooled,
                 kv,
                 gate,
+                self.prev_win_kv,
+                self.prev_win_gate,
             )
         else:
             self._undo = None
@@ -163,6 +176,13 @@ class PoolingCache(_BaseCache):
             self.accumulate_windows(buf_kv, buf_gate, 0)
         self.pooled = pooled
         self._undo = None
+        # prev_win is runtime-only and not part of the persisted 3-tuple, so
+        # a state restore (SSD eviction/recovery) cannot rebuild it. The next
+        # window completed after restore pools with a zero lane-A once, then
+        # prev_win repopulates — a single-window overlap loss, not the
+        # persistent per-window loss the original decode bug caused.
+        self.prev_win_kv = None
+        self.prev_win_gate = None
 
     @property
     def meta_state(self):
@@ -197,11 +217,15 @@ class PoolingCache(_BaseCache):
             return n
         if not self._can_undo(n):
             return 0
-        buf_kv, buf_gate, rem_prev, pooled_prev, kv, gate = self._undo
+        buf_kv, buf_gate, rem_prev, pooled_prev, kv, gate, prev_kv, prev_gate = (
+            self._undo
+        )
         self._undo = None
         k = kv.shape[1] - n
         self.pooled = pooled_prev
         self.remainder = rem_prev
+        self.prev_win_kv = prev_kv
+        self.prev_win_gate = prev_gate
         if buf_kv is not None:
             self.buf_kv[:, :rem_prev] = buf_kv
             self.buf_gate[:, :rem_prev] = buf_gate
@@ -254,6 +278,12 @@ class BatchPoolingCache(_BaseCache):
         self._processed = [0] * batch_size
         self._undo = None
 
+        # Previous completed window's raw KV/gate per batch row, for overlap
+        # (ratio==4) compressors — see PoolingCache.prev_win_kv docstring.
+        # None until a row completes its first window.
+        self.prev_win_kv = None
+        self.prev_win_gate = None
+
     @property
     def offset(self):
         return mx.array(self._pool_lengths, dtype=mx.int32)
@@ -291,6 +321,8 @@ class BatchPoolingCache(_BaseCache):
                 list(self._processed),
                 kv,
                 gate,
+                self.prev_win_kv,
+                self.prev_win_gate,
             )
         else:
             self._undo = None
@@ -445,6 +477,8 @@ class BatchPoolingCache(_BaseCache):
     def state(self, v):
         self.buf_kv, self.buf_gate, self.pooled = v
         self._undo = None
+        self.prev_win_kv = None
+        self.prev_win_gate = None
 
     @property
     def meta_state(self):
@@ -480,7 +514,9 @@ class BatchPoolingCache(_BaseCache):
             return n
         if not self._can_undo(n):
             return 0
-        buf_kv, buf_gate, remainder, pool_lengths, processed, kv, gate = self._undo
+        buf_kv, buf_gate, remainder, pool_lengths, processed, kv, gate, prev_kv, prev_gate = (
+            self._undo
+        )
         self._undo = None
         k = kv.shape[1] - n
         # The undo path only triggers when some row completed a window,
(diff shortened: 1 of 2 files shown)


williamxie1989:
**Update**: Added batch dimension safety check on `prev_win` concatenation.

In batch generation, `prev_win_kv` may be stored with a different batch dimension than the current step (e.g., B=1 from a single sequence completing a window, then B=2 when another sequence joins). This caused a `ValueError: [concatenate] All the input array dimensions must match exactly except for the concatenation axis`.

The fix adds `and prev_kv.shape[0] == kv.shape[0]` to the condition — when batch dimensions don't match, `prev_win` is skipped for that step (one step of overlap loss, then it repopulates with the correct batch size).

jundot:
Thanks for the deep dive, the root cause analysis is solid and the single sequence fix checks out, I verified decode pooling is now bit-exact against unchunked prefill. One thing came up in batch testing though: prev_win_kv is one tensor shared across the batch, and accumulate_windows zero-fills rows that did not complete a window this step, so with concurrent requests a completing row overwrites the other rows' carry with zeros, which then get pooled at finite gate values instead of -inf masking. I reproduced it numerically with B=2 and offset window phases: pooled outputs matched neither the correct overlap nor the old behavior. I'll merge this and land a follow-up that tracks the carry per row (per-row valid flags, -inf gates for rows without a carry), verified bit-exact for batch, trim, filter, extend and merge, plus a real model smoke on DS4 Flash with MTP and concurrent batching.

jundot:
@williamxie1989, I released oMLX 0.5.2.dev1 with this change. Please try the new build and let me know if you see any related issues: https://github.com/jundot/omlx/releases

Issue #2255, opened 2026-07-15
Degenerate output (`!!!!!!...`) after prefix cache hits — suspected KV restore corruption


raucodes:
## Environment
 
- oMLX version: `0.5.1`
- Hardware: Mac Studio M4 Max, 128 GB unified memory
- macOS: `26.5.2`
- Model: `Qwen3.5-122B-A10B-oQ5` (MoE), served via Anthropic Messages endpoint (`/v1/messages`, streaming)
- Sampling: temperature=1, top_p=0.95, top_k=20, presence_penalty=1.5, max_tokens=1024
- Cache: tiered KV cache enabled (hot RAM + cold SSD), block size 2048
## Summary
 
Since prefix cache hits started occurring in my setup (my client previously produced non-deterministic JSON key ordering, so every request was a full-prefill cache miss — after fixing that client-side, cache restore now works), the model intermittently produces fully degenerate output: a stream consisting only of repeated `!` tokens. This happened twice within the first session with working cache hits and never before, across weeks of daily use of the same model, prompts, and sampling parameters with 100% cache misses.
 
This looks like corrupted KV state after block restore (e.g. position/RoPE offset mismatch relative to block boundaries, an off-by-one when trimming to the 2048-token block boundary, a dtype mismatch on the safetensors round-trip, or an inconsistent "prefill boundary cache snapshot") rather than a sampling issue — the same sampling params were stable for weeks without cache hits.
 
## Observed behavior
 
- Prefill time roughly halved on affected requests (partial prefix hit confirmed working).
- Completion degenerates immediately: `!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!...` until stopped by the client.
- One example turn: `Anthropic message: 11 tokens in 1.62s (6.8 tok/s)` — the "response" was the degenerate token spam, replaced by a client-side fallback.
- Possibly related: immediately after one degenerate turn, a TTS request on the same server failed with `TTS native stream done: 0.26s, chunks=0, pcm_bytes=0` → `POST /v1/audio/speech → 500: TTS streaming produced no audio output`. May be coincidental, noting it in case the TTS engine shares scheduler/cache state.
Relevant log lines around a degenerate turn:
 
```
2026-07-15 18:16:34,301 - omlx.scheduler - INFO - Cache phase timings: cleanup_finished_sync=31.8ms/8, store_cache_main_boundary=0.2ms/8
[...]

Issue #1842, opened 2026-06-12
0.4.4rc1: four compounding transient-accounting/churn defects cap multi-turn long-context at ~136K (0.4.3: ~167K) — with fixes the ceiling moves to the model's context limit (262K)


hojin12312:
## Summary

While validating 0.4.4rc1 for long multi-turn sessions (same campaign that
found #1832 and #1833), we hit a multi-turn context ceiling of ~136K on a
setup that reached ~167K on 0.4.3 — and after root-causing it, the wall
turned out to be four separate, compounding defects, all in rc1-new code
paths (the adaptive prefill throttle, and the singleton-batch passthrough
added around `_merge_caches`). Fixed in order, each one exposes the next:

| Stack (Gemma 4 26B 3-bit, 36 GB Mac Studio, ceiling 30 GB) | Multi-turn ceiling |
|---|---|
| 0.4.3 (same hardware) | ~167K |
| 0.4.4rc1 + #1832/#1833 fixes | 136–138K (rejections + shrink crawl) |
| + fix 1 — spike-only transient measurement | 150K @ 99% hit |
| + fix 2 — memory-bounded TQ prefill fallback | 160K |
| + fix 3 — restore-churn pool reclaim | 165K |
| + fix 4 — TQ singleton passthrough + one-shot restore concat | **254K @ 99% sweep completed; probe to 262,084 = model RoPE limit − 60 tokens** |

After all four fixes the session ends only at the model's own
`max context window of 262144` (a clean, instant 400), with wired memory
at 18.3 GB and ~11 GB of headroom at 254K — i.e. the machine was never
actually out of memory; the ceiling was accounting and churn.

Defects 1 and 3 are TurboQuant-independent. Defects 2 and 4 affect any
TQ-eligible model on stock rc1 (e.g. Qwen3.6-35B-A3B, dense models).

## Environment

- M-series Mac Studio 36 GB, `iogpu.wired_limit_mb=30720`, memory ceiling 30 GB
- Gemma 4 26B (UD 3-bit), 30 layers = 5 global `KVCache` + 25 `RotatingKVCache`
  (sliding window 1024); TurboQuant KV 4-bit on 4/5 global layers via a local
  eligibility relaxation for hybrid models (proposal: #1843 — on stock
  rc1 Gemma 4 is silently TQ-ineligible, but defects 2/4 reproduce with any
  stock TQ-eligible model)
[...]

Issue #1822, opened 2026-06-11
TurboQuant skip_last causes 'array has no attribute norms' on every SSD cache reconstruction (0.4.3)


michaelAstraios:
## Summary

In oMLX 0.4.3, enabling TurboQuant KV cache with `turboquant_skip_last=True` (the default) causes **every SSD cache reconstruction to fail** with:

```
omlx.cache.prefix_cache - WARNING - Failed to reconstruct cache: 'array' object has no attribute 'norms'
```

This means the SSD KV cache is written successfully but can **never be read back**. Every request falls back to a full re-prefill regardless of how much context was already cached, negating the entire benefit of the SSD cache.

## Root Cause

`turboquant_skip_last=True` intentionally leaves the last KVCache layer as a plain `mx.array` (not converted to `TurboQuantMSEState`). However, the reconstruction path in `prefix_cache.py` attempts to access `.norms` on all layers without checking whether that specific layer was skipped, causing an `AttributeError` on the plain array.

The relevant code path:
- `scheduler.py` line 1128: `self._turboquant_skip_last: bool = True`
- `scheduler.py` line 2250–2251: last KV layer is skipped during conversion
- `prefix_cache.py`: reconstruction accesses `.norms` unconditionally, hitting the skipped layer

## Steps to Reproduce

1. Enable TurboQuant on any model via `model_settings.json` (`turboquant_kv_enabled: true`, `turboquant_skip_last: true`)
2. Run any request that fills enough context to trigger SSD cache writes
3. On the next request, observe the warning in the logs:
   ```
   WARNING - Failed to reconstruct cache: 'array' object has no attribute 'norms'
   ```
4. Confirm via logs that a full re-prefill occurs on every subsequent request despite a warm cache

## Environment

- oMLX version: 0.4.3 (upgraded from 0.3.10)
- macOS: 15.x (Apple Silicon M5 Pro Max, 128 GB RAM)
- Model: `Qwen3.5-35B-A3B-Claude-4.6-Opus-Reasoning-Distilled-MLX-8bit`
- Model settings: `turboquant_kv_enabled: true`, `turboquant_kv_bits: 4.0`, `turboquant_skip_last: true`

## Workaround

Set `turboquant_kv_enabled: false` in `~/.omlx/model_settings.json` for the affected model. Plain KV cache reconstructs correctly from SSD, restoring cache hit behaviour.

## Expected Behaviour

Either:
[...]

Issue #1793, opened 2026-06-10
TurboQuant: chunked-prefill path skips KV convert, storing fp16 blocks into TQ prefix chains → permanent 0% cache hit ("'array' object has no attribute 'norms'")


hojin12312:
## Describe the bug

On dense all-`KVCache` models with TurboQuant KV enabled, the **chunked-prefill path never converts the completed fp16 prefill cache to TurboQuant**, while the external-prefill path does. The request then decodes *and stores its prefix-cache blocks* in fp16 (`layer_cache_types = KVCache`), appending them onto a chain whose earlier blocks are `TurboQuantKVCache`.

Reconstruction derives the per-layer format from the **first** block's metadata, so the next fetch of that chain interprets the bare fp16 arrays as TQ state and crashes with:

```
'array' object has no attribute 'norms'
```

The entire cache is rejected — and because the corrupted blocks stay deduplicated in the tiered cache, the chain **never heals**: every subsequent turn of the conversation is a full cold re-prefill (permanent 0% cache hit).

Verified present in v0.4.3 and still on `main` (0.4.4.dev1, checked 2026-06-10).

> Note: #729 reports the same error string, but that is a different root cause (hybrid Gated DeltaNet `ArraysCache` serialization). This issue is specific to **dense models where every layer is a plain `KVCache`** (i.e. `_turboquant_eligible` returns True), e.g. Granite 4.1. Rotating/sliding-window models (Gemma 4 etc.) are gated out by `_turboquant_eligible` and unaffected.

## Root cause

1. `_do_external_prefill` quantizes the completed fp16 cache at its epilogue (`scheduler.py` ~L2644-2651 on main):

   ```python
   # Quantize the completed fp16 KV cache to TurboQuant for decode.
   if self._turboquant_kv_bits is not None and self._turboquant_eligible(prompt_cache):
       self._apply_turboquant_kv_convert(prompt_cache)
   ```

2. The chunked-prefill path (`_begin_prefill` / `_step_prefill_chunk`) has **no equivalent step**. Both of its completion paths (inline first chunk, and `_advance_chunked_prefills`) funnel through `_insert_prefilled_request` (~L3304), which inserts `state.cache` into the batch generator as-is — still fp16.

3. The request decodes fp16 and, on finish, `store_cache` saves its new blocks with `layer_cache_types = ["KVCache", ...]` onto a prefix chain whose older blocks (from previous turns that took the external/TQ path) are `["TurboQuantKVCache", ...]`.

[...]

PR #2272, opened 2026-07-17
fix(cache): reconstruct prefix blocks per payload format, not chain metadata

### LINKED ISSUE #729: [BUG] Qwen 3.5 122B Prefix Cache Reconstruction Failure: 'array' object has no attribute 'norms'
Describe the bug
Prefix cache reconstruction fails when using the Qwen 3.5 122B-A10B model. The engine throws an 'array' object has no attribute 'norms' error when attempting to reload a cached context. This prevents subsequent prompts from utilizing the fast-path prefix cache, forcing a fallback to boundary snapshots and heavy re-prefilling, which significantly impacts performance on long context (80k+ tokens).

To Reproduce
Load Qwen 3.5 122B-A10B in oMLX.

Initialize a long context session (e.g., 60k-80k tokens of code or architectural docs).

Send a subsequent prompt that should trigger a prefix cache hit.

Check the logs to see the reconstruction failure and the 'array' object has no attribute 'norms' error.

Expected behavior
The prefix cache should successfully reconstruct the state for the hybrid Gated DeltaNet/Attention layers, allowing for near-instant Time-to-First-Token (TTFT) on subsequent prompts within the same context.

Desktop:
OS: macOS Tahoe(Apple Silicon - M5-series Max)

oMLX Version: v0.3.5.dev1

Additional context
The error occurs because Qwen 3.5 122B utilizes a hybrid architecture where 75% of the layers are Gated DeltaNet (linear attention). Unlike standard Transformers, these layers maintain a normalization state (norms).

It appears the omlx.cache.prefix_cache serialization logic is attempting to treat the ArraysCache (recurrent state) as a standard MLX array, which lacks the .norms attribute required for these specific hybrid layers.

Log Snippet:

Plaintext
2026-04-11 21:31:24,623 - omlx.cache.prefix_cache - WARNING - [-] - Failed to reconstruct cache: 'array' object has no attribute 'norms'
2026-04-11 21:33:38,752 - omlx.scheduler - INFO - [-] - Using boundary cache snapshot... storing 79872/81776 tokens (skipping trailing partial block)
Additionally, this cache desync seems to break Speculative Decoding for this model; when a draft model (2B/4B) is used, the 122B outputs garbled/non-functional code, likely because the engine cannot correctly "roll back" the DeltaNet state during a token rejection.

JimStenstrom:
Fixes #729. Addresses the strongest code-supported mechanism behind #2255.

## Summary

- Prefix-cache reconstruction now dispatches each block on its **own payload format** instead of typing the whole chain from chain-level metadata, so mixed TurboQuant/dense chains reconstruct instead of crashing (`AttributeError: 'array' object has no attribute 'norms'`, #729) and forcing a full re-prefill.
- When a block's `(bits, seed)` cannot be resolved from metadata, the codec is **no longer rebuilt at a guessed `bits=4.0/seed=0`** — it falls back to the server-configured TurboQuant KV bit depth, and if that is also unavailable the hit is **rejected and re-prefilled** instead of guessed. A guessed width dequantizes to plausible-but-wrong tensors: silent output corruption after a cache hit, the mechanism that best matches the degenerate-output report in #2255.

## Why

A block chain can legitimately mix payload formats: blocks stored while TurboQuant KV conversion was active carry tagged `('__turboquant_v2__', (ks, vs))` states, while blocks stored without it (chunked-prefill completions on older versions, or a different TQ setting) carry plain dense `(keys, values)` — in the same chain, via dedup. The per-block `layer_cache_types` mismatch check only truncates such chains when every block's metadata is present and accurate, but block metadata is optional at load time: files can be missing or corrupt, and the JSON can fail to round-trip. For metadata-blind chains, reconstruction typed everything from the chain-level type and fed dense arrays into TurboQuant `_concat_state` (the #729 crash) — or, worse, rebuilt codecs from a hardcoded `bits=4.0` default that silently dequantizes wrong-width tensors for any chain stored at a different depth.

## Changes

- `omlx/cache/prefix_cache.py`
  - `reconstruct_cache`: payload-driven per-block dispatch. Consecutive TurboQuant blocks sharing `(bits, seed)` group into runs that concatenate quantized and dequantize once; dense blocks pass through; empty-slice placeholders and unrecognized payloads reject the hit with a clear log. A homogeneous TQ chain remains a single run restored as a quantized `TurboQuantKVCache` (preserving #1842's no-full-materialization behavior); all-dense chains take the untouched dense path — uniform chains see zero behavior change.
  - `_tq_block_params`: resolution order is now per-block meta → chain-level (first-block) meta → server-configured `_expected_turboquant_kv_bits` (the same attribute the block-eligibility signature is keyed on, so an admitted block's depth provably matches it; seed is always the codec default 0 — no omlx path constructs a TurboQuant cache with another seed, though seed is not part of the eligibility signature) → `None`, which makes the caller reject the hit instead of guessing.
  - Block metadata list stays index-aligned with block data when a block loads without metadata, so per-block resolution can never read a neighbor block's parameters (previously possible in principle; observable only for mixed-parameter chains with dropped metadata).

## Tests

`python -m pytest tests/test_prefix_cache.py tests/test_paged_ssd_cache.py tests/test_turboquant.py -q` → **308 passed** (103 in `test_prefix_cache.py`: 93 pre-existing + 10 new).

New tests, all in the existing per-module file, driving the real `reconstruct_cache` through the `load_block_with_metadata` seam:
- 3 mixed-chain reconstruction tests (TQ head + dense tail, dense head + TQ tail, hybrid model with an ArraysCache layer) — verified to FAIL with the fix reverted.
- 2 behavior-preservation guards: homogeneous TQ chains and all-dense chains reconstruct exactly as before.
- 3 safe-rejection guards: placeholder payloads (both chain typings) and corrupt payloads return `None` (re-prefill), never wrong KV.
- 2 params-resolution tests — verified to FAIL with the fix reverted: metadata-blind uniform chain resolves the codec from the configured bit depth and dequantizes byte-equal to the full-state reference; with no configured depth either, the hit is rejected instead of silently rebuilding at 4 bits.

Also verified against the parent commit (`5a39ba3a`): 5 of the 10 new tests fail there — the 3 mixed-chain tests with the original crash signature, plus both params-resolution tests (the parent's hardcoded `bits=4.0` dequantizes one chain at the wrong width and reconstructs another it should reject).

Live verification on the reported model family (Qwen3.5 122B A10B hybrid, TurboQuant KV 8-bit, default `skip_last`): stored a ~4k-token prefix (2 × 2048-token blocks; the chain organically mixes ArraysCache/TurboQuantKVCache/KVCache layers), then stripped `layer_meta_states` from the block files — `cache_signature` left intact, so bit-depth eligibility still admits them — to reproduce a metadata-blind chain. On this branch the hit heals through the configured-bits fallback at cache-hit speed (2.2s end-to-end vs 5.3s cold, coherent output). On the parent commit the same request aborts mid-generation — `[reshape] Cannot reshape array of size 1401 into shape (1,2,1401)` inside TurboQuant norms handling (the `bits=4.0` guess applied to 8-bit states) — and the stream returns an empty completion: reconstruction "succeeds" and the corruption detonates at the first attention dequantize, the delayed-failure shape #2255 reports.

Reconstruction micro-benchmark (M5 Max, 4-layer × 4k-token chains, real quantized states): homogeneous-TQ and all-dense reconstruction are unchanged vs the parent (within GPU dispatch noise at N=30; all-dense 0.99×); the per-block payload scan costs ~0.6µs per layer; a healed mixed chain reconstructs in ~3.8ms.

## Why this is safe

- I could not construct a healthy chain the old code reconstructed but the new code rejects: every new `return None` maps to a payload the old path either also rejected (via a caught crash) or silently corrupted (the old loop skipped non-2-tuple blocks, dropping their KV and misaligning positions). Reject-loud is strictly more correct in each case.
- Reconstruction is single-entry: hot-tier and SSD restores both converge on `reconstruct_cache`; `boundary_snapshot_store` performs no TurboQuant reconstruction (store-side only); GDN/`ArraysCache` layers are excluded by the existing `supports_block_slicing` gate; VLM and LLM engines share the path. No sibling reconstruct path diverges.
- The configured-bit-depth fallback only engages where the old code guessed `4.0` unconditionally — configurations whose metadata round-trips (the healthy common case) never reach it.

## Trade-offs / out of scope

- A mixed chain heals by dequantizing to a dense cache for that one restored request, trading TurboQuant's memory savings for correctness on a rare recovery path (measured: 128 MiB fp32 vs 16.5 MiB quantized for a 4k-token, 4-layer chain). Homogeneous chains keep the quantized restore.
- `_rebuild_codecs` reproducibility for fractional bit-widths (its docstring guarantees integers only) is a pre-existing property this PR narrows exposure to but does not change — flagging it here as a known residual; happy to follow up separately if you want it hardened.

Field context: #2255 reports degenerate (`!!!!`) output beginning exactly when prefix-cache hits started working, on the same model family as #729 (Qwen 3.5 122B hybrid). The silent wrong-width rebuild fixed here is the strongest code-supported mechanism for that symptom, but the reporter has not confirmed it (whether TurboQuant KV was enabled in that setup is not stated in the thread) — which is why this PR claims #729 and only "addresses" #2255.

A note on #729's attribution: the report (and #1793's cross-reference to it) blames the hybrid model's GDN `ArraysCache` layer, which indeed lacks `.norms` — but every `.norms` access in omlx's cache code is `hasattr`-guarded; the unguarded accesses live in mlx_vlm's TurboQuant state handling, reached only through the TurboQuant reconstruction branch. The error therefore signals a non-TurboQuant payload routed into TurboQuant reconstruction — the mechanism fixed here. The hybrid-model test reproduces the reported setup (ArraysCache layer present) and fails on the parent commit with the reported error string.

## Adjacent open reports (same signature, not claimed)

- #1793: the store-side cause was fixed in 0b7a226, and its tail hardening already heals poisoned chains in one turn by truncating and re-storing the mixed tail. This change narrows the remaining cost: a mixed tail now reconstructs instead of being truncated and re-prefilled.
- #1822 (`skip_last` → the same error on every reconstruction, reported on 0.4.3): payload-driven dispatch routes dense payloads out of TurboQuant reconstruction regardless of how the chain is typed, which covers the reported mechanism (a dense payload in a TurboQuant-typed chain) — worth a retest on this branch before closing.

### DIFF
--- a/omlx/cache/prefix_cache.py
+++ b/omlx/cache/prefix_cache.py
@@ -1856,6 +1856,12 @@ class BlockAwarePrefixCache(CacheManager):
                     # Always update last to track the most recent
                     last_block_meta_states = block_layer_meta_states
                     all_block_meta_states.append(block_layer_meta_states)
+                else:
+                    # A block can load with data but no metadata at all. Keep
+                    # all_block_meta_states index-aligned with all_block_data
+                    # so per-block (bits, seed) resolution never reads a
+                    # neighbor block's meta.
+                    all_block_meta_states.append(None)
 
                 # Validate loaded data (pass cache types for hybrid models)
                 if not self._validate_block_cache_data(block_data, layer_cache_types):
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, this is an outstanding writeup and the fix direction is exactly right: dedup keeps historical mixed chains alive on disk no matter what the store side does, so payload-driven dispatch on the read side is the correct place to fix this, and dropping the bits=4.0 guess closes the strongest silent-corruption mechanism tied to #2255. I verified the claims locally, including the 5 expected failures on the parent commit reproducing with the exact crash signatures. I'll merge this.

One edge I'll handle in a follow-up so you don't need to touch anything: the mixed-chain heal emits float32 (dequantize output plus concat promotion) with no cast back to the model dtype, so on a server that later runs with TurboQuant disabled the healed layer can promote a merged batch to fp32. A cast after the final concat covers it, and I'll fold in a couple of small internal cleanups on the same path.

jundot:

JimStenstrom:
Tried 0.5.2.dev1 (cp312 wheel, M5 Max 128GB) with `turboquant_kv_enabled` at 8-bit KV on a Qwen3.6-35B-A3B oQ4 model and the paged SSD cache on: wrote a ~2.7k-token prefix, restarted the server, and re-sent the identical request — the rescan indexed the TQ chain with `errors=0` and no stale-signature invalidation, the admin cache probe reports its full 2048-token block ssd-servable (only the 643-token partial tail cold, as expected), and the answer came back identical. The signature gate also visibly discriminates in both directions in my logs: the pre-TQ dense chain was invalidated the moment the model loaded with TQ KV on, and a later dense-signature load skipped the TQ file as incompatible instead of guessing. Also ran a concurrent batch with mid-batch departures on the same TQ config: the surviving long request stayed coherent to the end, and both short requests came back byte-identical to single-request runs of the same config. No related issues seen.

Thanks for the dtype-cast follow-up — saw it land minutes after the merge.

PR #2278, opened 2026-07-17
feat(mtp): nemotron_h support — depth-1 cycle and Lightning MTP chain

davidtorcivia:
## Summary

- Adds MTP speculative decoding for `nemotron_h` (Nemotron-3-Super-120B-A12B ships a trained `mtp.*` head that mlx-lm currently discards in `sanitize`), including the depth-k Lightning MTP chain with the adaptive depth controller.
- Nemotron-H is a Mamba/Transformer hybrid (40 Mamba2 + 8 attention + 40 MoE layers), so verify rollback has to restore recurrent state, not just trim KV. The approach here makes partial rollback a pure ref restore — no replay forward.
- Measured (Nemotron-3-Super-120B, 5-bit affine, M3 Ultra), 400-token greedy generations: 44.2 tok/s no-MTP -> 51.9 tok/s (+17%), 69.8% draft acceptance; sampled t=1.1: 46.2 -> ~47-49.

## Model patch (`nemotron_h_model.py`)

Mirrors the qwen35/deepseek structure: `MTPModule` with parameter paths matching the on-disk `mtp.layers.{0,1}.*` names (fuse `eh_proj(concat(enorm(embed(t+1)), hnorm(h_t)))` -> attention sublayer -> MoE sublayer -> final norm; only routed experts stacked in `sanitize`), `mtp_forward` with the chain interface (`return_hidden`, `logits_keep`), `make_mtp_cache` (`[KVCache()]`, trimmable, so no head-clone mode), `n_confirmed` plumbing through backbone/block/mixer, and the per-instance `_omlx_mtp_decode_enabled` stamp in `__init__`.

## Verify cycle on a hybrid trunk (`nemotron_h_chain.py`)

Verify windows keep the batched projections (each layer's weights stream once for the whole window) but run the SSM scan sequentially with the existing fused single-step kernel:

- faster than the chunked `ssm_attn` graph at window sizes 2-8 (0.35 vs 0.63 ms/layer at S=2; ~40 Mamba layers per cycle);
- numerically identical to the plain decode path (same kernel), so draft/verify distributions stay aligned;
- yields per-position (conv, ssm) states as a free byproduct, so partial rollback restores `cache._mtp_pos_states[keep-1]` — no replay forward. Conv tails are reconstructed from the window input inside a `_conv` wrap; a raw-input stash + replay path remains as fallback for masked windows.

Cycle backbone: 29.5 ms vs 22.6 ms for a plain decode step (37.1 ms with the qwen35-style whole-window verify + replay); rollback cache time 15 ms vs 176 ms per 400-token run.

nemotron_h's `return_hidden` returns post-`norm_f` hidden, so the model patch stamps `_omlx_mtp_head_hidden_normed` (from #2277) to keep the chain's head inputs from being double-normed.

## Depth guidance

The stock head is depth-1-trained (sampled acceptance ~50-59%), so the loader defaults nemotron_h to a fixed depth-1 cycle (greedy 51.9 vs 46.9 with the adaptive controller). Setting `mtp_num_draft_tokens` overrides this for checkpoints with a stronger head.

## Testing

- `tests/test_nemotron_mtp_patch.py`: apply idempotence and markers, decode-flag/chain stamps on a tiny model, `mtp_forward` chain interface shapes, and per-position captured states vs a fresh prefix recompute.
- Per-position captured states vs prefix-reference runs on real shapes: ~1e-7 (fp32 state math).
- Greedy speculative output equals its non-speculative greedy stream (lossless); sampled acceptance uses the standard batched Leviathan/Chen path unchanged.
- `MTP[...]` per-sequence stats verified end to end (cycles, per-depth accepts, backbone/mtp/sample/cache timings).
- Running in production on our single-node M3 Ultra deployment.

## Issues

Builds on #2277 (branch is based on it; the first commit here is that fix).

Co-authored with Claude Fable 5 (Anthropic).

### DIFF
--- a/omlx/patches/mlx_lm_mtp/__init__.py
+++ b/omlx/patches/mlx_lm_mtp/__init__.py
@@ -101,6 +101,8 @@ def apply_mlx_lm_mtp_patch() -> bool:
         cache_rollback,
         deepseek_v4_model,
         glm_moe_dsa_model,
+        nemotron_h_chain,
+        nemotron_h_model,
         qwen35_model,
     )
 
@@ -114,6 +116,10 @@ def apply_mlx_lm_mtp_patch() -> bool:
         logger.debug("DeepSeek-V4 MTP patch did not apply (likely missing base patch)")
     if not glm_moe_dsa_model.apply():
         logger.debug("GLM-5.2 MTP patch did not apply (likely missing base patch)")
+    if not nemotron_h_model.apply():
+        logger.debug("nemotron_h MTP patch did not apply (likely import error)")
+    elif not nemotron_h_chain.apply():
+        logger.debug("nemotron_h MTP chain patch did not apply")
     if not batch_generator.apply():
         logger.warning(
             "BatchGenerator MTP dispatch patch failed; MTP path will be inactive"
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -1379,6 +1379,12 @@ def _trunk_norm_module(model: Any):
 
     Walks both wrapper conventions: mlx-lm's outer ``Model.language_model``
     and oMLX's ``VLMModelAdapter._language_model`` (mlx-vlm path).
+
+    Models whose ``return_hidden`` output is already post-norm mark
+    ``_omlx_mtp_head_hidden_normed`` (instance or inner language model) and
+    get an identity here — applying the trunk norm again would double-norm
+    the head inputs, and the ``inner.model.norm`` walk does not fit every
+    backbone layout (e.g. ``backbone.norm_f``).
     """
     inner = model
     for attr in ("language_model", "_language_model"):
@@ -1386,6 +1392,10 @@ def _trunk_norm_module(model: Any):
         if candidate is not None:
             inner = candidate
             break
+    if getattr(model, "_omlx_mtp_head_hidden_normed", False) or getattr(
+        inner, "_omlx_mtp_head_hidden_normed", False
+    ):
+        return lambda x: x
     return inner.model.norm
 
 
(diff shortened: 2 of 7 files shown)


jundot:
Thanks, this is a seriously well built patch. The sequential fused verify with per-position state capture is a clever way to get replay-free rollback on a hybrid trunk.

One blocker though: `_is_mtp_compatible` in `omlx/utils/model_loading.py` still only allows qwen3_5/qwen3_6/deepseek_v4/glm_moe_dsa and this PR doesn't touch that file, so on a stock server `apply_mlx_lm_mtp_patch` never runs for a nemotron_h load and the whole feature stays inert. The nemotron_h_model docstring references a nemotron_h gate there, so I'm guessing that hunk just didn't make it into the commit.

Two smaller things. `_capture` in nemotron_h_chain.py is a module global, so two nemotron family models decoding concurrently in one process could interleave arm/read and cross-write per-position states into the wrong cache, a threading.local would close that cheaply. And `_patch_trunk_norm` duplicates the marker check #2277 already adds to `_trunk_norm_module` itself, since this branch is based on that commit the extra wrap can be dropped.

Also curious about the default: the stock head is depth-1 trained and your own numbers show the adaptive default losing about 10% vs fixed depth 1, did you consider stamping depth 1 for nemotron_h unless the user overrides?

davidtorcivia:
Good guess on the gate, that's exactly what happened. I assembled the branch from my deployment tree and the model_loading.py hunk got dropped, and since my server runs a wheel that already has it, nothing ever surfaced locally. Fixed now. `_is_mtp_compatible` takes `model_type.startswith("nemotron_h")`, and `_has_mtp_heads` already matches `num_nextn_predict_layers` which is what the checkpoint declares, so no other loader changes were needed. Added a test on the gate too.

`_capture` is a threading.local now. Arm and read happen inside a single mixer forward so that covers the concurrent models case without touching anything else.

Dropped `_patch_trunk_norm` entirely, it predates rebasing onto #2277 and was pure duplication like you said.

On the default, agreed, it's in the loader now. nemotron_h gets fixed depth 1 when `mtp_num_draft_tokens` is unset and an explicit setting still wins. I originally left it doc only figuring fine-tuned heads would want the adaptive controller, but I've since run a KL distilled fine-tune of the head locally (72% greedy acceptance, up from 62%) and fixed depth 1 still wins even there, most of the cycle is backbone time so deeper drafts need much higher acceptance than any nemotron_h head I've seen. Depth 1 is the right default for now.

jundot:
All four look good on my end, and the gate test is a nice touch since that failure mode was invisible to the rest of the suite. The KL distilled head data point is interesting too, that matches the backbone-dominated cycle math. I'll merge #2277 first and then this one.

PR #2279, opened 2026-07-18
feat: download chat history

cobracode:
### Downloadable Chat History

Chat history used to be stored on disk and doesn't appear to be anymore. This lets users download their precious and interesting chats.

Also moved some repeated LocalStorage string values to constants.

### DIFF
--- a/omlx/admin/i18n/en.json
+++ b/omlx/admin/i18n/en.json
@@ -748,6 +748,7 @@
   "chat.admin_settings": "Admin Settings",
   "chat.clear_all_history": "Clear All History",
   "chat.change_api_key": "Change API Key",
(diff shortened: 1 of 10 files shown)


jundot:
Thanks, this is a nice addition, chat history really needed an export path. I'll merge this and push a small follow-up: the early theme script now references THEME_STORAGE_KEY, but the constants are declared in the later script block, so the pre-render theme snippet throws before the constant exists and dark mode flashes on load. I'll also move the Download Chats button label to i18n like the other sidebar buttons. After that I'm planning an import counterpart so the exported file works as a real backup.

cobracode:
Thank you @jundot !

PR #2280, opened 2026-07-18
fix(vlm): inject DeepSeek V4 DSML tool-calling into VLM tokenizer

williamxie1989:
## Problem

`VLMBatchedEngine._inject_tool_calling()` depends on `mlx_vlm.tool_parsers._infer_tool_parser()` and `mlx_lm.tokenizer_utils._infer_tool_parser()` to identify the correct tool parser for a model's chat template. Neither of these upstream inference functions recognise DeepSeek V4's DSML markers (`<｜DSML｜tool_calls>`), so `_inject_tool_calling` silently returns without setting `has_tool_calling`, `tool_call_start`, `tool_call_end`, or `tool_parser` on the VLM tokenizer.

This causes downstream components (`ToolCallStreamFilter`, `parse_tool_calls`, `sanitize_tool_call_markup`) to be unable to detect and strip DSML markup from streaming deltas, thinking blocks, and final responses. Raw `<｜DSML｜tool_calls>` tags leak into the API response text visible to Claude Code and other clients.

## Fix

Added a fallback path at the end of `_inject_tool_calling`: when both `mlx_vlm` and `mlx_lm` fail to infer a parser type, check if the chat template contains `"<｜DSML｜tool_calls>"`. If it does, load `omlx.patches.deepseek_v4.tool_parser_v4` — oMLX's bundled DeepSeek V4 parser — and inject its attributes onto the tokenizer.

This follows the same pattern as the existing upstream parsers and is consistent with the existing `detect_output_parser` logic in `omlx/adapter/output_parser.py`.

## Test

Added `test_injects_deepseek_v4_dsml_fallback` to `tests/test_vlm_engine.py` that verifies the injection sets the correct `tool_call_start`, `tool_call_end`, `has_tool_calling`, and a callable `tool_parser`.

All existing tests pass:
- 8/8 `_inject_tool_calling` tests (including the new one)
- 234/234 tool_calling tests
- 3/3 DeepSeek V4 output parser tests
- 334/334 VLM engine tests

## Related

The same root cause also affects the scheduler's output parser detection timing. `_inject_tool_calling` runs after `AsyncEngineCore` creates the scheduler (which deep-copies the tokenizer), so the scheduler's copy lacks the DSML attributes. A follow-up fix would move the injection earlier in the VLM start sequence.

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -1777,10 +1777,15 @@ class VLMBatchedEngine(BaseEngine):
         mlx_lm's — it recognises additional markers such as Gemma4's <|tool_call>
         and loads the correct per-model parser.  Falls back to mlx_lm if the
         mlx_vlm.tool_parsers package is not present.
+
+        For models whose parser isn't in mlx_vlm/mlx_lm (e.g. DeepSeek V4 DSML),
+        fall back to omlx's bundled tool parser modules.
         """
         chat_template = getattr(tokenizer, "chat_template", None)
         if not chat_template:
             return
+
+        tool_module = None
 
         # Prefer mlx_vlm.tool_parsers (superset; knows about Gemma4 etc.)
         try:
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the detailed writeup, the _inject_tool_calling analysis itself is accurate. But DeepSeek V4 is a text model and oMLX routes it to the LM engine, where the full DSML support lives (chat template injection, tool parser, MTP), so that path is where you get correct and optimal results. The VLM engine isn't a supported path for DSv4, the mlx-vlm port is text only and its chat template has no tool grammar at all, so tool calling wouldn't work there even with this injection, and the vocab check at the end of _inject_tool_calling would still silently reject the DSML markers since they aren't single vocab tokens in the real tokenizer. Out of curiosity, which checkpoint did you run and how did it end up on the VLM engine? Stock oMLX classifies deepseek_v4 as LM, so if it got misrouted I'd rather find and fix that instead.

Issue #2274, opened 2026-07-17
Non-streaming path drops SpecPrefill kwargs (specprefill_keep_pct ignored, system=0)


necoli1822:
## Summary
On the **non-streaming** `/v1/chat/completions` path, a per-model `specprefill_keep_pct` (e.g. `0.25`) is ignored — SpecPrefill runs at the hardcoded `DEFAULT_KEEP_RATE = 0.20`. The scheduler log also reports `system 0` even when a real system prompt is present. **Streaming requests are unaffected** (they honor `0.25` and compute the system boundary). Independent of #766.

Affects 0.5.1 and `main`.

## Root cause
`omlx/engine/batched.py` — `BatchedEngine.generate()` (non-streaming) builds `SamplingParams` from `kwargs` but then calls
```python
output = await self._engine.generate(
    prompt=prompt,
    sampling_params=sampling_params,
)   # **kwargs not forwarded -> specprefill_keep_pct / _threshold / _system_end dropped
```
`stream_generate()` forwards these correctly; the non-streaming path was never updated. `server.py` puts `ms.specprefill_keep_pct` into `chat_kwargs`, `chat()` forwards `**kwargs` to `generate()`, and `EngineCore.generate()` already forwards `**kwargs` to `add_request()` — so only `generate()` needs the forwarding.

## Evidence (real 0.20, not a cosmetic label)
- Non-streaming, 9167-token prompt: `selected 1839/9167` → 20.06%.
- After forwarding the kwargs, non-streaming 13647-token prompt: `selected 3407/13647` → 24.96% ≈ 0.25.

## Fix
Forward the specprefill kwargs in `generate()`, mirroring `stream_generate()`:
```python
specprefill_kwargs = {
    k: kwargs[k]
    for k in ("specprefill", "specprefill_keep_pct",
              "specprefill_threshold", "specprefill_system_end")
    if kwargs.get(k) is not None
}
output = await self._engine.generate(
    prompt=prompt, sampling_params=sampling_params, **specprefill_kwargs
)
```

## Related minor gap
`system 0` on non-streaming: `chat()` doesn't compute `specprefill_system_end` (only `stream_chat()` does). For full parity, factor that computation out of `stream_chat()` and use it in `chat()` too.

I have this fix locally and verified it; happy to open a PR if useful.

JimStenstrom:
Confirmed — this is real, not a labeling artifact. Traced it end to end:

[...]

PR #2292, opened 2026-07-19
fix(specprefill): forward SpecPrefill kwargs on the non-streaming path (#2274)

Khambampati-Subhash:
BatchedEngine.generate() dropped specprefill_* overrides and chat() skipped the specprefill_system_end computation, so non-streaming requests fell back to the engine's default keep rate and left the system prompt unprotected. Streaming requests were unaffected.

Factor the shared logic into _pop_specprefill_kwargs() and _inject_specprefill_system_end() helpers and use them in both the streaming and non-streaming paths so they behave identically. Add unit tests covering the non-streaming forwarding and system-end computation.

### DIFF
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -622,6 +622,66 @@ class BatchedEngine(BaseEngine):
         )
         return len(self._tokenizer.encode(prompt))
 
+    @staticmethod
+    def _pop_specprefill_kwargs(kwargs: dict[str, Any]) -> dict[str, Any]:
+        """Pop SpecPrefill per-request overrides out of ``kwargs``.
+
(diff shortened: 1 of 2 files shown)


Khambampati-Subhash:
normally we have this specprefill it is getting passed in the streaming functions but not in the non streaming so instead of copy pasting the same code from streaming i made some helper functions which can be used by both streaming and non streaming paths no duplicates

jundot:
Thanks for picking this up. I verified the whole chain (server chat_kwargs → chat → generate → add_request), the helpers match the streaming logic exactly, and the new tests pass on my end too. Credit to @necoli1822 for the sharp diagnosis and measurements in #2274. I'll merge this.

PR #2295, opened 2026-07-19
fix(specprefill): derive RoPE width from freqs, not head_dim

### LINKED ISSUE #2294: SpecPrefill sparse prefill crashes on Gemma 4 (partial rotary): [broadcast_shapes] in manual_rope_with_freqs
**Describe the bug**

With SpecPrefill enabled, sparse prefill fails on Gemma 4 targets:

```
SpecPrefill sparse prefill failed: [broadcast_shapes] Shapes (1,4,2048,256) and (1,1,2048,64) cannot be broadcast.
```

It's caught and the request falls back to dense prefill, so there's no crash to the client — but SpecPrefill delivers no speedup on these models, and the draft-scoring work (several seconds on a long prompt) is done and then thrown away every time.

**To reproduce**

- Target: `gemma-4-31B-it-qat-4bit` (also reproduces on the 26B / other Gemma 4 variants)
- Draft: any (e.g. `Qwen3.5-0.8B-MLX-4bit`)
- SpecPrefill enabled; send an uncached prompt over the threshold (>8k tokens)

The error is logged from the scheduler on every such request.

**Root cause**

`manual_rope_with_freqs` (in `omlx/patches/specprefill.py`) slices the head by the passed `dims` (the rope module's reported `head_dim`) while building `cos`/`sin` from `_freqs`, whose length is the rotary sub-dimension ÷ 2. On Gemma 4 the rope is built `initialize_rope(dims=head_dim)`, so `dims` is the full `head_dim` (256) while `_freqs` covers a smaller rotary slice (and local vs global attention heads differ) — so `2 * len(freqs) != dims` and the multiply can't broadcast. The model's real RoPE rotates `2 * len(freqs)` dimensions, not `dims`.

**Fix**

Derive the rotary width from the freq table (`2 * len(freqs)`) instead of `dims`. This is a no-op for full-rotary custom-`_freqs` models (Llama-3 / YaRN / SuScaled, where `2*len(freqs) == dims`) and corrects the partial-rotary / mixed-head case. PR attached.

Adjacent to #851 (Gemma 4 SpecPrefill compatibility), which fixed other Gemma-4 issues but not this one.

**Environment**

- oMLX 0.5.2.dev1
- Apple Silicon (M1 Ultra, 64 GB), macOS

rockminster:
### Summary

Fixes #2294.

`manual_rope_with_freqs` slices the head by the passed `dims` (the rope module's reported `head_dim`) while building `cos`/`sin` from `_freqs`, whose length is the rotary sub-dimension ÷ 2. When a model reports `dims` = full `head_dim` but rotates a smaller slice — e.g. **Gemma 4**, built `initialize_rope(dims=head_dim)` with a partial rotary and differing local/global attention heads — `2 * len(freqs) != dims`, so the multiply can't broadcast and sparse prefill dies:

```
SpecPrefill sparse prefill failed: [broadcast_shapes] Shapes (1,4,2048,256) and (1,1,2048,64) cannot be broadcast.
```

It's caught and falls back to dense prefill, so SpecPrefill silently does nothing on Gemma 4 (and the draft-scoring work is wasted each long prefill).

### Fix

Derive the rotary width from the model's own frequency table — `2 * len(freqs)`, exactly what the model's real RoPE rotates — instead of the passed `dims`.

- **No-op** for full-rotary custom-`_freqs` models (Llama-3 / YaRN / SuScaled), where `2 * len(freqs) == dims` already — same output, no regression.
- **Corrects** the partial-rotary / mixed-head case (Gemma 4, and any model loaded the same way).
- Standard base-θ RoPE models never reach this function (they use `manual_rope`), so they're untouched.

Adjacent to #851, which addressed other Gemma 4 SpecPrefill compatibility issues but not this one.

### Validation

- `pytest tests/test_specprefill.py -m "not slow"` → **41 passed** (39 existing + 2 new regression tests for `manual_rope_with_freqs`: partial-rotary width from freqs, and full-rotary unchanged).
- Verified live on `gemma-4-31B-it-qat-4bit` (Qwen3.5-0.8B draft, >8k uncached prompt): SpecPrefill sparse prefill now completes instead of logging the broadcast error and falling back; decode is coherent.

### Test plan

```bash
pytest tests/test_specprefill.py -m "not slow" -v
```

### DIFF
--- a/omlx/patches/specprefill.py
+++ b/omlx/patches/specprefill.py
@@ -659,10 +659,34 @@ def manual_rope(x, positions, dims, base=10000.0, scale=1.0):
 def manual_rope_with_freqs(x, positions, dims, freqs, pre_scale=1.0):
     """Apply RoPE at arbitrary positions using pre-computed frequencies.
 
-    For custom RoPE variants (Llama3, Yarn, SuScaled) that store _freqs.
+    For custom RoPE variants (Llama3, Yarn, SuScaled) that store ``_freqs``,
+    and for partial-rotary models that report ``dims`` = full head_dim while
+    their frequency table covers only a rotary sub-slice (e.g. Gemma 4, whose
+    local/global attention heads differ).
+
+    The model's real RoPE (mlx-vlm ``ProportionalRoPE`` / HF ``rotate_half``)
+    pairs dim ``i`` with dim ``i + dims // 2`` over the FULL head, and only the
+    first ``len(freqs)`` of those pairs actually rotate -- the remaining pairs
+    pass through unchanged. We reproduce that exactly by zero-padding the
+    inverse frequencies up to ``dims // 2`` (angle 0 is the identity rotation:
+    ``cos 0 = 1``, ``sin 0 = 0``), so the non-rotating pairs are untouched.
+
+    Bit-exact for full-rotary custom-``_freqs`` models (``len(freqs) == dims //
+    2``, no padding added) and correct for the partial-rotary / mixed-head
+    case. An earlier version derived the width as ``2 * len(freqs)`` and rotated
+    the contiguous first ``2 * len(freqs)`` dims (pairing ``i`` with ``i +
+    len(freqs)``), which rotated the wrong lanes and wrote misrotated KV on
+    every global layer (jundot, PR #2295).
     """
     half = dims // 2
+    n = int(freqs.shape[-1])
     inv_freq = (1.0 / freqs).astype(mx.float32)
+    if n < half:
+        # Zero-pad the unrotated pairs: angle 0 keeps them identity, matching
+        # the model's partial rotary exactly.
+        inv_freq = mx.concatenate(
+            [inv_freq, mx.zeros((half - n,), dtype=mx.float32)], axis=-1
+        )
     angles = positions[:, None].astype(mx.float32) * inv_freq[None, :]
     cos_a = mx.cos(angles)[None, None, :, :]
     sin_a = mx.sin(angles)[None, None, :, :]
--- a/tests/test_specprefill.py
+++ b/tests/test_specprefill.py
@@ -110,6 +110,103 @@ class TestManualRoPE:
         assert mx.allclose(result[..., dims:], x[..., dims:])
 
 
+class TestManualRopeWithFreqs:
+    """Tests for manual_rope_with_freqs() — RoPE from a precomputed freq table."""
+
+    @staticmethod
+    def _reference_partial_rotary(x, positions, dims, freqs):
+        """Independent oracle for the model's real partial rotary, written as an
+        explicit per-pair rotation in numpy so it shares NO code with the MLX
+        implementation under test. Pairs dim ``j`` with dim ``j + dims // 2``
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for tracking this down, the diagnosis is right and I can reproduce the exact broadcast crash with the mlx-vlm gemma4 ProportionalRoPE (dims=512, len(freqs)=64). The fix direction has a problem though: gemma 4's real rope follows HF's rotate_half convention over the full head, so it pairs dim i with dim i + dims/2 for the first len(freqs) pairs (dims 0..63 with 256..319 here), while this patch rotates the contiguous first `2*len(freqs)` dims pairing i with i + len(freqs). Comparing both against the model's actual ProportionalRoPE output, this patch diverges with max abs diff ~5.9 (dims 256..319 that should rotate stay untouched, dims 64..127 that should pass through get rotated), so sparse prefill would silently write misrotated KV on every global layer while decode queries still use the real rope. Since the current crash is caught and falls back to dense prefill, that would trade a perf no-op for a quality regression.

The minimal correct fix keeps `half = dims // 2` and instead zero-pads `inv_freq` up to `dims // 2` (angle 0 is identity for the unrotated pairs). That matches the real ProportionalRoPE to fp32 eps in the same comparison and stays bit-exact for the full-rotary models. The test expectations need the same adjustment (only the first len(freqs) dims of each half of the head change).

I'll merge this once the pairing is updated.

rockminster:
Rebased on latest main and squashed. Summary of what's in this now:

The rope fix. manual_rope_with_freqs was pairing frequencies against the full head dim, which broke for partial-rotary rotate-half models (gemma-4). This zero-pads inv_freq out to dims // 2 and splits x into x_rot/x_pass so only the rotary slice is rotated. It's a no-op for full-rotary models and correct for partial-rotary ones — verified bit-exact: a position-mapped-rope vs genuine-rope probe on gemma-4 (12B and 31B) gives worst_rel ≈ 6e-7 across needle depths, and dense-vs-sparse A/B answers match at depths 0.0 and 1.0.

Tests. Added the coverage you asked for — the rotate-half / partial-rotary rotation case that the old pairing got wrong.

One behavioural note worth flagging. Pristine manual_rope_with_freqs raised a broadcast mismatch on partial-rotary models, which was caught upstream and routed those models to dense fallback. This fix removes that crash, so a partial-rotary model that previously declined sparse prefill will now actually run it. In my testing that produced correct output (including qwen3_5-family targets, turboquant on and off), so no regression surfaced — but it's a genuine change in reachability rather than pure math, so calling it out explicitly. If you'd rather gate sparse behind a per-model faithfulness check rather than let it run on anything that no longer crashes, happy to discuss separately — I looked at that and it's a larger change that doesn't belong in this fix.

jundot:
Merged, thanks for the careful rework. The zero-pad implementation checks out on my side too, and the independent numpy oracle in the tests is a nice touch. On gating sparse prefill behind a per-model faithfulness check, I agree it doesn't belong here and I'd rather keep the dispatch simple, so let's leave things as they are unless a real model surfaces a mismatch.

PR #2296, opened 2026-07-19
feat(admin): support provider-specific extra body for external benchmarks

earthking11:
## Summary

- Add an optional JSON `extra_body` field for external accuracy and throughput benchmarks.
- Merge provider-specific fields into OpenAI-compatible request bodies while preventing overrides of oMLX-owned fields.
- Add a session-only admin UI input with validation, help text, and translations for all supported locales.

## Motivation

Some OpenAI-compatible providers require request fields that are not part of the standard chat-completions schema. For example, a DeepSeek-compatible endpoint may require:

```json
{"thinking":{"type":"disabled"}}
```

Previously, external benchmark requests could not send these provider-specific fields without changing oMLX source code.

## Safety

The merge rejects protected fields such as `model`, `messages`, `stream`, `max_tokens`, `temperature`, `api_key`, and `authorization`, so `extra_body` cannot replace request fields controlled by oMLX.

The admin value is session-only and is not persisted by oMLX.

## Tests

- `pytest tests/test_external_api.py tests/test_admin_external_extra_body.py` — 42 passed
- `pytest tests/test_accuracy_benchmark.py` — 33 passed
- JSON locale validation and JavaScript syntax checks passed


jundot:
Thanks, this one is clean. The `extra_body` validator rejecting the oMLX-owned fields on both the pydantic and the JS side is exactly the right call, and it flows straight through into the request body without being able to touch anything the benchmark controls. I'll merge this.

PR #2297, opened 2026-07-19
fix(admin): distinguish external benchmark failures from wrong answers

earthking11:
## Summary

- Classify external accuracy results as `correct`, `wrong`, `parse_error`, `empty_content`, `truncated`, `timeout`, `http_error`, `connection_error`, or `invalid_response`.
- Capture `finish_reason`, reasoning-field diagnostics, token usage, and provider error details for each question.
- Add a stricter preflight that verifies the endpoint returns a usable final answer before starting a paid evaluation.
- Report valid-response rate and valid-answer accuracy separately from total accuracy.
- Include diagnostics in the admin UI and JSON, CSV, and text exports.
- Preserve local evaluation behavior and update the custom HumanEval, MBPP, and LiveCodeBench runners for compatibility.

## Motivation

External endpoint failures were previously represented as empty model responses and counted as ordinary wrong answers. This made it difficult to distinguish model quality from endpoint, protocol, truncation, or parsing failures.

The new result breakdown keeps total accuracy comparable while exposing whether an external benchmark run produced enough valid answers to be reliable.

## Compatibility

Local model evaluation keeps its existing scoring behavior. The additional per-question diagnostic fields are populated only for external API evaluations.

## Tests

- `pytest tests/test_external_api.py tests/test_eval.py tests/test_accuracy_benchmark.py tests/test_admin_external_accuracy_diagnostics.py` — 150 passed
- JSON locale validation and JavaScript syntax checks passed


jundot:
Nice work. Splitting endpoint failures out from genuine wrong answers is a real improvement, and the valid-answer accuracy vs total accuracy breakdown is exactly what external runs were missing. The 5-tuple to 6-tuple change is also covered across all four `run()` paths, so no eval regresses.

I'll merge this and fold two small things into a follow-up so you don't have to round-trip:

- The reliability banner uses `border-amber-300` and `text-amber-800`, which aren't in the compiled `tailwind.css` yet, so it renders without the border and text color until I rebuild the CSS.
- I'll loosen the preflight exact-`OK` check a little so an endpoint that replies `OK.` instead of exactly `OK` doesn't fail the whole run.

Since this and #2296 touch the same files, I'll land #2296 first and rebase this one on top.

PR #2299, opened 2026-07-19
fix(oq): use compatible loader for oQe calibration

seeker-cyber-maker:
## Summary
- route oQe text-model calibration through `lm_load_compat`
- preserve `trust_remote_code` forwarding on older mlx-lm versions while omitting the removed keyword on current mlx-lm
- add a regression test for the current mlx-lm loader signature

## Root cause
The oQe imatrix path called `mlx_lm.load` directly. Current mlx-lm no longer accepts `trust_remote_code`, causing calibration to fail after the RAM-safe proxy had been built. Other oQ paths already use the compatibility wrapper.

## Validation
- `.venv/bin/python -m pytest tests/test_oq.py::TestCollectImatrixTextLoad tests/test_oq.py::TestMeasureSensitivityVlmMtp -q`
- Result: 9 passed

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -5714,7 +5714,7 @@ def _collect_imatrix(
 
             tokenizer = load_tokenizer(Path(model_path))
         else:
-            from mlx_lm import load as lm_load
+            from omlx.utils.model_loading import lm_load_compat as lm_load
 
             model, tokenizer = lm_load(
                 model_path,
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for fixing this. I verified that routing the oQe text calibration load through `lm_load_compat` avoids the unsupported keyword on affected mlx-lm versions while preserving forwarding where it is supported. LGTM.

PR #2308, opened 2026-07-21
Add oQe imatrix quantization controls to the macOS app

jknlsn:
## Summary

Brings the macOS Quantization screen to parity with the web dashboard's oQe (imatrix-enhanced) controls.

- New **Enhanced quantization (oQe)** section: enable oQe, reuse imatrix cache, an optional cache path, and strict imatrix coverage. The four values are sent as `enhanced`, `imatrix_cache_path`, `imatrix_reuse_cache`, and `imatrix_strict` — the same fields `/admin/api/oq/start` already accepts and forwards.
- The oQ level picker and the start-confirmation label show `oQ…e` when oQe is on, matching the web UI.
- Sensitivity-model candidates are now filtered by `modelType` instead of a two-token name prefix, so sibling quantized models of the same type are offered. This mirrors the web filter (`m.model_type === source.model_type`).
- Replaced the static About card with a collapsible **About oQ Quantization** panel: the oQ/oQe comparison table, the standard-oQ feature list, and the imatrix workflow steps — adapted from the web content.
- Added a `DTOFixtureTests` case for the enhanced request encoding.
- Localization: new `quant.enhanced.*` and `quant.about.subtitle` strings, plus the expanded About copy wrapped in `String(localized:)` with 35 new `quant.about.*` keys; removed three orphaned `quant.about.*` keys left over from the old static card.

## Notes

- No backend change — the oq start endpoint already accepts and forwards all oQe/imatrix fields. This is client parity only.

## Validation

- `xcodebuild test -project apps/omlx-mac/oMLX.xcodeproj -scheme oMLX -destination 'platform=macOS' -only-testing:oMLXTests/DTOFixtureTests -only-testing:oMLXTests/LocalizationSmokeTests` — `** TEST SUCCEEDED **` (10 tests).
- Built and ran the app locally and ran an oQe quantization end-to-end (succeeded).

## Screenshots (before and after)
   <table><tr>
     <td valign="top">SCR-20260721-lqes</td>
     <td valign="top">SCR-20260721-lmli</td>
   </tr></table>

### DIFF
--- a/apps/omlx-mac/Resources/Localizable.xcstrings
+++ b/apps/omlx-mac/Resources/Localizable.xcstrings
@@ -8680,56 +8680,14 @@
         }
       }
     },
-    "quant.about.body1" : {
-      "comment" : "First body paragraph of the About oQ card",
+    "quant.about.subtitle" : {
+      "comment" : "Subtitle of the collapsible About oQ Quantization card",
       "extractionState" : "manual",
       "localizations" : {
         "en" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "Quantization should not be exclusive to any particular inference server. oQ produces standard mlx-lm models that work everywhere — oMLX, mlx-lm, LM Studio, and any app that supports MLX safetensors format. No custom loader required."
-          }
-        },
-        "ru" : {
-          "stringUnit" : {
-            "state" : "translated",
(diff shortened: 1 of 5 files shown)


jundot:
Thanks for bringing the mac app up to parity here, this is a clean one. I checked the four fields against what /admin/api/oq/start expects and the empty cache path handoff to the server-generated location, and the model_type filter matches the web dashboard exactly, so it all lines up. The collapsible about panel is a nice upgrade over the static card too. I'll merge this.

PR #2316, opened 2026-07-21
fix(admin): make log viewer level filter exact per-level

Khambampati-Subhash:
The log viewer level buttons acted as a minimum-severity threshold, so selecting INFO also showed WARNING/ERROR/CRITICAL lines and selecting WARNING also showed ERROR/CRITICAL. The per-level button UI made users expect that selecting a level shows only that level's lines.

Switch to an exact per-level filter and add an ALL option (now the default) to show every line. Continuation lines (stack traces, etc.) still inherit the visibility of the entry they belong to. Update the "Min display level" label to "Display level" across all locales.

And also there is still trace log which is added in logging.addLevelName(5, "TRACE") it is low level so it shows all level logs but it should show only trace logs one thing is our code is not producing the trace level logs


Khambampati-Subhash:
Screenshot 2026-07-21 at 10 51 57 PM
After change you can see in UI

jundot:
Thanks for taking the time on this and for the before/after screenshots. The threshold behavior is intentional: selecting a level shows that level and above, which is the standard convention for log levels (python logging, syslog, journalctl all work this way). An exact per-level filter loses the most useful view, "warnings and above", and exact ERROR would even hide CRITICAL lines.

You have a point that the button row reads like an exact filter though. So instead of changing the semantics I improved the affordance in 1b077428: every level at or above the selected one is now shown dark, so the included range is obvious. The TRACE button showing everything is the same threshold semantics at work, since trace is the lowest level.

Closing this one, but I appreciate you surfacing the confusion.

Khambampati-Subhash:
Thanks @jundot I will close the issue also

PR #2327, opened 2026-07-22
Add security policy for SSRF vulnerability in oMLX

medasz:
Documented SSRF vulnerability in oMLX admin panel benchmark feature, detailing its impact, remediation steps, and disclosure notes.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,147 @@
+# SSRF via Benchmark External Endpoint in oMLX Admin Panel
+
+## Summary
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for taking a look at this, but I'm going to close it.

The external benchmark endpoint is admin-gated, and admin in oMLX is already a fully privileged role. Whoever holds the admin key can load arbitrary local models, which is effectively code execution on the host, so making the server send an outbound HTTP request is well within what admin can already do rather than a privilege boundary crossing. oMLX also ships as a single-user desktop app where admin is the machine owner, so there's no separate network position to escalate from.

One more thing: a public PR carrying a full PoC for an unpatched issue isn't the right channel. Please open a private security advisory for anything you believe crosses a real trust boundary. I may still add metadata/link-local IP filtering to the benchmark validator as defense in depth, but that's hardening, not a vuln fix.

medasz:
Thanks for the reply. I did initially submit changes related to security and quality; it was my first time submitting a security-related fix to a public repository, and I hadn't fully considered everything. I appreciate your suggestions.

Issue #2281, opened 2026-07-18
Spec Prefill


aemji:
Since a few version SpecPrefill always goes 20% even if 10% is selected, in fact, no other value than 20% is supported.
I even used to modify manually to set it at 15 (0;15) and it's impossible too

Other problem, Spec prefill don't look like supported by cache, it stay at 0 when spec prefill is enable(every version)

I comeback to older version, because spec prefill is too good for speed

PR #1782, opened 2026-06-10
fix: honor per-model SpecPrefill settings on every generation path

popfido:
## Summary

Fixes #1263 — the per-model `specprefill_threshold` / `specprefill_keep_pct` settings were silently ignored on most paths, so SpecPrefill always activated at the hardcoded boundary (threshold 8192, keep 20%) regardless of configuration.

## Root cause

The settings→request chain only existed on the **OpenAI chat-completions route**, which injects the values into `chat_kwargs` server-side (`server.py`, added in ba9f8b7 for #443). Every other entry point dropped them:

| Path | Break |
|---|---|
| **Admin benchmark** | `_run_single_test` calls `engine.stream_generate(prompt, max_tokens, temperature, top_p)` — no specprefill kwargs at all. This is the path both reports in #1263 used (logs show `omlx.admin.benchmark`), which is why threshold=32768 still activated at pp16384 and keep_pct=0.1 still logged `keep=20%`. |
| **Non-streaming chat** (`engine.chat`) | `BatchedEngine.generate` / `VLMBatchedEngine.generate` built `SamplingParams` and called `engine_core.generate(prompt, sampling_params)` — the specprefill kwargs died unused in `**kwargs`. |
| **`/v1/completions` and `/v1/messages`** | Those routes never build specprefill `chat_kwargs` server-side, so nothing was forwarded even on streaming. |

Meanwhile activation still *happened* because `engine_core.add_request` enables SpecPrefill whenever the draft model is loaded — so users saw it firing, just never at their configured threshold. Exactly the reported symptom.

## Fix

Apply the model-settings fallback at the **engine chokepoint** instead of per-route injection: new `resolve_specprefill_kwargs(kwargs, model_settings)` in `omlx/engine/base.py` pops the per-request overrides (which still win) and fills `specprefill_threshold` / `specprefill_keep_pct` from the engine's own `model_settings`. Both `BatchedEngine` and `VLMBatchedEngine` use it in `stream_generate` **and** `generate`, so the benchmark, non-streaming chat, `/v1/completions`, and `/v1/messages` all resolve identically with no server-side changes.

Precedence: per-request API value → server-injected live setting (chat route, unchanged) → engine model-settings snapshot → dflash/scheduler defaults.

The `specprefill` **enable** flag intentionally stays request-only: `engine_core` already enables SpecPrefill iff the draft model is loaded, and the draft model only loads when `specprefill_enabled` was set — defaulting it from settings here would be redundant.

## Tests

[...]

PR #2338, opened 2026-07-22
fix(specprefill): forward SpecPrefill kwargs on the VLM non-streaming path

JimStenstrom:
Completes the fix from #2292 (#2274) on the VLM engine. May also explain #2281's "keep rate always 20%" for VLM-hosted models.

## Summary

- `VLMBatchedEngine.generate()` now forwards `specprefill`, `specprefill_keep_pct`, `specprefill_threshold`, `specprefill_system_end` to the engine core, mirroring `stream_generate`.
- `VLMBatchedEngine.chat()` now computes `specprefill_system_end` before delegating, mirroring `stream_chat`.
- Same helper structure as #2292 (`_pop_specprefill_kwargs`, `_inject_specprefill_system_end`), used by all four request paths.

## Why

#2292 fixed the non-streaming SpecPrefill kwarg drop in `BatchedEngine`, but the VLM engine has the identical twin: non-streaming `generate()` built `SamplingParams` and called `self._engine.generate(...)` without the specprefill kwargs, and non-streaming `chat()` never computed the system-prompt boundary. Result: non-streaming requests on VLM-hosted models silently fell back to the default keep rate regardless of configured settings, ignored per-request `specprefill=False`, and left the system prompt unprotected. Streaming requests were unaffected. (Noted on #2274 at the time as the remaining engine.)

## Changes

- `omlx/engine/vlm.py`: added `_pop_specprefill_kwargs()` and `_inject_specprefill_system_end()`; `generate()` pops and forwards; `chat()` injects before delegating to `generate()`; `stream_generate()`/`stream_chat()` now call the helpers instead of their inline copies — no behavior change on streaming: the inject helper is the extracted `stream_chat` inline code verbatim, preserving VLM's pre-tokenized-prompt semantics (`full_tokens = len(prompt)` on the token-ID list from `_process_chat_messages`; bare `tokenizer.apply_chat_template` for the non-system render).
- `tests/test_vlm_specprefill.py`: new `TestVLMEngineSpecPrefillForwarding` class, six tests mirroring #2292's, exercising the real `generate()` / `chat()→generate()` chain.

## Behavior change (worth a release-notes line)

For VLM-hosted models with SpecPrefill configured, non-streaming requests now honor the configured keep rate / threshold / per-request `specprefill=False` and protect the system prompt — previously they always ran the engine default. A model configured with a keep rate lower than the default will drop more tokens on non-streaming calls than before, i.e. the user's configuration now takes effect.

## Sibling sweep

- DFlash fallback `generate`/`chat` forward `**kwargs` to the fallback engine → covered transitively.
- Diffusion lane: rejects `specprefill=True` loudly; both diffusion early-returns precede the pop/inject, so the inject runs at most once per request and never on diffusion paths.
- Out of scope (pre-existing, affects both engines equally): `admin/benchmark.py`'s direct `add_request` passes no specprefill kwargs (deliberate engine-default benching); the Anthropic `/v1/messages` and Responses server handlers never build specprefill overrides (only the OpenAI chat handler does).

## Why this is safe

- No new engine contract: the forwarding lands on `add_request` parameters that are first-class in the engine core and have been exercised in production by the streaming path since before this change; the core's `generate()` is a pure passthrough to the same `add_request`.
- Every failure direction is conservative: an over-estimated `system_end` only shrinks or skips the scored region; an inject failure is caught and degrades to exactly the pre-fix no-protection state; non-positive boundaries are ignored at the core.
- Structural mirror of already-merged #2292, verified line-for-line against `batched.py`; the only deltas are forced by the VLM prompt type (pre-tokenized token-ID list).

## Overlap

#1782 (open) proposes a broader all-paths rework that also touches `vlm.py`, untouched since 2026-07-10. This PR follows the scoped-fix path #2292 established on the batched engine; happy to yield if the broader rework is the preferred direction.

## Tests

- `python -m pytest tests/test_vlm_specprefill.py tests/test_vlm_engine.py -q` → **100 passed** (6 new).
- `python -m pytest tests/test_batched_engine.py -q` → **42 passed** (sibling suite untouched).
- Revert-proof: with `omlx/engine/vlm.py` reverted to the parent commit, the new class fails 4 of 6 — the two absence-assertion guards correctly still pass.

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -2875,6 +2875,70 @@ class VLMBatchedEngine(BaseEngine):
             prompt = "\n".join(f"{m['role']}: {m['content']}" for m in messages)
             return prompt + "\nassistant:"
 
+    @staticmethod
+    def _pop_specprefill_kwargs(kwargs: dict[str, Any]) -> dict[str, Any]:
+        """Pop SpecPrefill per-request overrides out of ``kwargs``.
+
+        The engine's ``add_request`` accepts these as dedicated arguments, so
+        they must be forwarded explicitly rather than left in ``**kwargs``.
+        Shared by ``generate`` and ``stream_generate`` so both request paths
+        honour SpecPrefill overrides identically.
+        """
+        specprefill_kwargs: dict[str, Any] = {}
+        for key in (
+            "specprefill",
+            "specprefill_keep_pct",
+            "specprefill_threshold",
+            "specprefill_system_end",
+        ):
+            if kwargs.get(key) is not None:
+                specprefill_kwargs[key] = kwargs.pop(key)
+        return specprefill_kwargs
+
+    def _inject_specprefill_system_end(
+        self,
+        messages: list[dict[str, Any]],
+        prompt: str | list[int],
+        kwargs: dict[str, Any],
+    ) -> None:
+        """Compute the system-prompt token boundary and add it to ``kwargs``.
+
+        SpecPrefill protects the system-prompt region from token dropping. The
+        boundary is derived by subtracting the non-system prompt token count
+        from the full prompt token count (system-only messages usually can't be
+        templated on their own). Shared by ``chat`` and ``stream_chat`` so the
+        non-streaming path protects the system prompt identically. No-op unless
+        the model has SpecPrefill enabled and the request has a system prompt.
+
+        ``prompt`` is the already-tokenized VLM prompt (a list of token IDs,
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for carrying the non-streaming SpecPrefill fix over to the VLM engine. I verified the full chat → generate → add_request path, including per-request disable and system-prompt protection, and the focused VLM suites pass against current main. LGTM. I'll merge this.

PR #2339, opened 2026-07-23
fix(vlm): honor frequency_penalty on VLM-hosted models

JimStenstrom:
## Summary

- Both `VLMBatchedEngine` `SamplingParams` constructions omitted `frequency_penalty`, so the value resolved by the server (request param or per-model setting) landed in `**kwargs` and was silently dropped — VLM-hosted models ignored it entirely while batched-hosted models honored it.
- Fix: forward `frequency_penalty=kwargs.get("frequency_penalty", 0.0)` in `generate()` and `stream_generate()`, exactly as `batched.py` does at both of its sites. Two lines; the shared scheduler already consumes the field from `SamplingParams`.

## Why

`get_sampling_params()` resolves `frequency_penalty` per request/model settings and every OpenAI handler forwards it to the engine (`chat_kwargs` for chat completions; explicit kwarg on `/v1/completions`). `SamplingParams` has the field and the scheduler feeds it to mlx-lm's frequency-penalty processor in both decode loops. The VLM engine was the only link that dropped it — its constructions set `presence_penalty` and `repetition_penalty` but not `frequency_penalty`, so it defaulted to 0.0 for every VLM-hosted model. Found while reviewing the sibling paths of the SpecPrefill forwarding fix (same silent-kwargs-drop class).

## Behavior change (worth a release-notes line)

A persisted per-model `frequency_penalty` setting (or a client-sent param) on a VLM-hosted model now actually applies, where it previously did nothing. This includes the internal PDF-OCR fallback path, which passes model-settings sampling values to the VLM engine — a leftover experimental setting that silently did nothing before will now shift OCR sampling after upgrade. The built-in OCR defaults never set `frequency_penalty`, so only user-set values arm it.

## Semantics note

mlx-lm's frequency penalty is windowed (last 20 context tokens, per-occurrence), not OpenAI's whole-completion counting. That is identical to what batched-hosted models already do — this PR only brings VLM to parity, it does not change the semantics.

## Sibling sweep

- Covered end-to-end after this fix: `/v1/chat/completions` (stream + non-stream), `/v1/completions` (stream + non-stream), `/v1/messages` and `/v1/responses` (model-settings value; neither API spec defines a request-level `frequency_penalty`, consistent with how presence/min_p/xtc are treated there), and the DFlash fallback engine (forwards `**kwargs` to its fallback — a VLM fallback now honors the value, closing that gap too).
- Out of scope, pre-existing and unchanged: DFlash's native speculative loop ignores all penalty params; the diffusion sub-path ignores `frequency_penalty` exactly as it ignores presence/repetition; the benchmark path deliberately uses greedy defaults.
- After this change the four `SamplingParams` construction sites (two per engine) forward an identical field set, modulo VLM's OCR-specific `stop_token_ids` addition.

## Why this is safe

- The mechanism is already proven on VLM: `presence_penalty` and `repetition_penalty` flow through the identical constructions into the identical scheduler processor list, consuming the identical per-sequence token buffer (including through MTP's lock-step buffer maintenance). `frequency_penalty` joins a live mechanism; only this field was dead.
- Zero control-flow impact: `kwargs.get` without pop, and neither `self._engine.generate()` nor `add_request()` accepts `**kwargs`, so the residual key is provably inert — the change is two data-flow lines.
- Bounded activation: the value only becomes nonzero from an explicit client param or a user-persisted setting; global and OCR defaults never set it, and the effect is capped to the last 20 context tokens — worst case for a stale setting is a mild, windowed logit shift identical to what batched-hosted models already exhibit.

## Tests

- New `TestVLMEngineFrequencyPenalty` (4 tests) exercising the real `generate()`, `stream_generate()`, and `chat()→generate()` paths, asserting the `SamplingParams` object that reaches the engine core.
- `python -m pytest tests/test_vlm_engine.py tests/test_vlm_specprefill.py -q` → **98 passed** (4 new). `python -m pytest tests/test_batched_engine.py -q` → **42 passed** (sibling untouched).
- Also verified the new tests fail with the fix reverted: 3 of 4 go red (the defaults test pins the absent-kwarg contract, which holds either way by design).


jundot:
Thanks for tracing this. Both autoregressive VLM `SamplingParams` construction sites were dropping `frequency_penalty`, while the shared scheduler already consumes it; forwarding the value restores parity with `BatchedEngine` for streaming and non-streaming requests. The focused VLM and sibling batched tests pass on current `main`. LGTM, merging.

Issue #2135, opened 2026-07-09
GLM MoE-DSA TestSmallLRouting fails on Apple M5 Max: absorbed-decode paths diverge up to 3.9e-4 (tol 2e-5) on pristine main


JimStenstrom:
**Describe the bug**

The two `TestSmallLRouting` equivalence tests added in #2113 fail on an Apple M5 Max against pristine `main`, with the exact pinned `mlx==0.31.2`:

```
FAILED tests/test_glm_mtp_patch.py::TestSmallLRouting::test_absorbed_matches_materialized
  AssertionError: L=2: 0.00012410059571266174
  assert 0.00012410059571266174 < 2e-05
FAILED tests/test_glm_mtp_patch.py::TestSmallLRouting::test_topk_gather_matches_masked_reference
  AssertionError: L=2: 0.00021101534366607666
  assert 0.00021101534366607666 < 2e-05
2 failed, 13 passed
```

The asserts stop at the first L, so I also collected the max abs diff for every L (print instead of assert):

- `test_absorbed_matches_materialized` — L=2: 1.241e-04, L=3: 1.856e-04, L=4: 1.427e-04, L=8: 3.871e-04
- `test_topk_gather_matches_masked_reference` — L=2: 2.110e-04, L=3: 3.076e-04, L=4: 2.568e-04

All 6–19× over the 2e-5 tolerance, roughly growing with L.

**To Reproduce**

On an M5-generation Mac:

```
git checkout main   # reproduced at 0.5.0.dev1 (a058f7d8) and 0.5.0.dev2 (8cf5e45c)
python -m pytest tests/test_glm_mtp_patch.py -k SmallLRouting
```

**What I ruled out**

- Not pin drift: installed mlx is exactly the pyproject pin (0.31.2), checked via `importlib.metadata`.
- Not my local stack: reproduced in a clean worktree of pristine upstream `main`; the other 13 tests in the file pass.
- Deterministic, not flaky: seeds are fixed and the diffs are bit-identical across runs on different days.
- Not a general regression: the same tests passed on CI's `macos-14` runners (M1-class) for #2113 across Python 3.11–3.14 the day it merged. The variable appears to be the GPU generation.

**Environment**

- Apple M5 Max, 128 GB unified memory
- macOS 26.5.2, Python 3.12.13, mlx 0.31.2
- `main` @ 8cf5e45c (0.5.0.dev2)

**Additional context**

[...]

PR #2340, opened 2026-07-23
fix(dflash): hoist transformers 5.x-nested draft config keys to the root

### LINKED ISSUE #2317: DFlash fails with Qwen3.6-35B-A3B-4bit
**Describe the bug**
I cannot seem to get DFlash working when using `mlx-community/Qwen3.6-35B-A3B-4bit`.  I'm using the proper DFlash model (`z-lab/Qwen3.6-35B-A3B-DFlash`) and I have it set as the DFlash model in the main model's settings.  However, the actual DFlash generation fails and falls back to regular generation, as seen in the logs below.

```
2026-07-21 12:31:49,706 - omlx.engine_pool - INFO - [-] - Loading model: Qwen3.6-35B-A3B-4bit
2026-07-21 12:31:49,706 - omlx.engine_pool - INFO - [-] - DFlash enabled for Qwen3.6-35B-A3B-4bit, draft=/Users/micahlt/.omlx/models/z-lab/Qwen3.6-35B-A3B-DFlash
2026-07-21 12:31:49,707 - omlx.patches.mlx_lm_mtp.nemotron_h_chain - INFO - [-] - nemotron_h MTP chain patch applied (depth-k drafting, sequential fused verify, replay-free rollback)
2026-07-21 12:31:50,556 - omlx.engine_pool - WARNING - [-] - DFlash start failed for Qwen3.6-35B-A3B-4bit: DFlashDraftModelArgs.__init__() missing 2 required positional arguments: 'rope_theta' and 'block_size'. Falling back to vlm engine.
2026-07-21 12:31:50,556 - omlx.patches.dflash_lifecycle - INFO - [-] - dflash class patches restored on 2 class(es)
2026-07-21 12:31:50,556 - omlx.engine.dflash - INFO - [-] - DFlashEngine stopped
2026-07-21 12:31:50,578 - omlx.patches.mlx_lm_mtp.nemotron_h_chain - INFO - [-] - nemotron_h MTP chain patch applied (depth-k drafting, sequential fused verify, replay-free rollback)
2026-07-21 12:31:50,578 - omlx.utils.model_loading - INFO - [-] - mlx-vlm runtime MTP patch applied for /Users/micahlt/.omlx/models/mlx-community/Qwen3.6-35B-A3B-4bit (config declares mtp heads but checkpoint ships no mtp.* weights; MTPModule attachment skipped to keep strict load_weights happy)
2026-07-21 12:31:54,921 - omlx.patches.qwen35_moe_gate_up - INFO - [-] - Qwen MoE gate+up fusion applied: 40 layers
2026-07-21 12:31:54,924 - omlx.engine.vlm - INFO - [-] - Vision feature cache enabled (SSD: /Users/micahlt/.omlx/cache/vision_features)
2026-07-21 12:31:55,380 - omlx.scheduler - INFO - [-] - Loaded 1 additional EOS token(s) from generation_config.json: {248044}
2026-07-21 12:31:55,381 - omlx.scheduler - INFO - [-] - Enlarging paged cache block_size=256 to 2048 for ArraysCache hybrid model (reduces boundary snapshot overhead)
2026-07-21 12:31:55,381 - omlx.memory_monitor - INFO - [-] - MemoryMonitor initialized (estimator-only, eviction disabled)
2026-07-21 12:31:55,381 - omlx.memory_monitor - INFO - [-] - Model info set: 40 layers (10 KVCache), 2 KV heads, 16 Q heads, 256 head_dim. Estimated memory per 64-token block: 5.00 MB
```

**To Reproduce**
1. Download a DFlash model for Qwen3.6-35B-A3B-4bit
2. Set it as the model for DFlash generation
3. Try any prompt
4. See error in the log file

**Expected behavior**
I expect DFlash generation to work properly.

**Desktop**
 - macOS Version: Tahoe 26.5
 - oMLX Version: 0.5.2 (had same issue with previous versions)

**Additional context**
I have tried several different DFlash models with no success as well.

JimStenstrom:
Fixes #2317

## Summary

- z-lab's retrained DFlash draft checkpoints (June 2026, transformers 5.7-style configs) moved `rope_theta` into a nested `rope_parameters` object and `block_size` into the existing `dflash_config` object. `dflash_mlx`'s `DFlashDraftModelArgs.from_dict` filters the config down to annotated dataclass fields and both keys are required positionals, so every current z-lab draft download fails with `DFlashDraftModelArgs.__init__() missing 2 required positional arguments: 'rope_theta' and 'block_size'` and the engine silently falls back to plain generation — exactly the log in the issue.
- Fix: a scoped patch module (`omlx/patches/dflash_draft_config.py`, same pattern as `dflash_lifecycle`) that wraps `from_dict` to hoist the nested values back to the config root before construction, installed at DFlash engine start. Hoisting is fill-only per key: a set (non-null) root key always wins, and a config missing the keys in both layouts still fails loudly exactly as today.

## Why omlx-side

Upstream dflash-mlx has no fix (the pinned 9ca0028 / v0.1.10 is its current HEAD), and the construction happens inside `load_draft_bundle` → `mlx_lm.load_model`, which reads config.json from disk — there is no omlx call site where the dict could be normalized directly, which is what the existing `patches/` machinery is for.

## Field evidence

Three users on the issue confirm the mechanism: manually adding root `rope_theta`/`block_size` fixes it (whipthraxton, confirmed by micahlt), and the values are "present but nested differently in the last DFlash downloads" (zviratko). Diffing the z-lab repo's April revision against the current one shows exactly the two relocations — plus root `rope_scaling` being replaced by `rope_parameters`, which the patch also maps for non-default rope types since the draft model consumes scaling via mlx-lm's `initialize_rope(..., scaling_config=args.rope_scaling)`.

## Verified on the real checkpoint

Downloaded the current `z-lab/Qwen3.6-35B-A3B-DFlash` snapshot and ran the real `load_draft_bundle` path in one fresh process: unpatched → the issue's exact TypeError; patched → full `DFlashDraftModel` construction with `block_size=16`, `mask_token_id=248077`, `target_layer_ids=[1, 6, 11, 16, 22, 27, 32, 37]`, `rope_theta=10000000`. (Reporters already confirmed on the issue that root-level values produce working end-to-end generation — the shim feeds `from_dict` exactly the config the manual workaround does, without editing users' files.)

## Why this is safe

- No bypass path: `mlx_lm.load_model` resolves `from_dict` by attribute lookup at call time and dflash-mlx imports the class (never a bound method), so nothing captures a pre-wrap reference; an adversarial sweep found no other `DFlashDraftModelArgs` construction site in dflash-mlx or omlx, and `DFlashEngine._load_models` — where the installer runs, before `load_draft_bundle` — is the only omlx path that triggers draft-args construction.
- No interference: the `dflash_lifecycle` restore machinery only rewrites `__call__` on classes it recorded, so it can neither revert nor double-wrap `from_dict`; gemma4's `from_dict` special-casing reads only keys the normalizer never touches; the wrap is idempotent (class-attribute guard) and output-identical for every root-keyed config.
- Error containment: mlx-lm's `initialize_rope` ignores unknown extra keys in every supported rope type, and an unrecognized rope_type raises inside `engine.start()` — landing in the same engine-pool fallback catch as today's TypeError, so no new error path escapes to the user.
- Consumer-exact semantics (hardened after adversarial review): the hoist's type-key precedence mirrors `initialize_rope` exactly (`"type"` before `"rope_type"`), and an explicit `null` root key counts as missing (transformers 4.x-derived configs commonly serialize `"rope_scaling": null`), so the hoist decision can never disagree with how mlx-lm reads the result.
- One deliberate behavior change, stated: a hybrid config with hand-added root keys, a non-"default" `rope_parameters` left in place, and no root `rope_scaling` now gains the scaling its config declares, where it previously loaded with unscaled RoPE. For the actual z-lab checkpoints `rope_type` is `"default"`, so users who applied the issue's manual workaround see a no-op; where it does fire it matches training, and draft-side RoPE can only affect acceptance rate, never output correctness (the target verifies every drafted token).

## Sibling sweep

- Model discovery classifies DFlash drafts by the presence of `dflash_config` (`_HELPER_CONFIG_KEYS` in `model_discovery.py`), which the new configs still carry — the draft picker is unaffected and needs no change.
- omlx itself never reads the draft config's root keys (`block_size` / `mask_token_id` / `target_layer_ids` are consumed via `args` inside dflash-mlx), so `from_dict` is the single normalization point. `draft_meta["config"]` keeps the raw nested dict, but omlx discards it and nothing in dflash-mlx reads raw-config `block_size` — worth re-checking on any future dflash-mlx pin bump.
- Out of scope, named: per-layer-type `rope_parameters` dicts (the `{"full_attention": {...}, "sliding_attention": {...}}` shape). A pure per-layer shape has no first-level scalars, so nothing hoists and the load degrades to exactly today's TypeError + fallback. `DFlashDraftModel` has no per-layer rope support at all, so no such draft checkpoint can exist for this model class without a dflash-mlx model change — at which point the constructor contract this shim targets changes anyway.

## Tests

- New `tests/test_dflash_draft_config.py` (13 tests): pure-function coverage (new layout hoists, old layout passes through unchanged, root-wins conflict, yarn `rope_scaling` derivation, `initialize_rope`-mirrored type-key precedence, explicit-null root keys, null nested values not hoisted, non-dict junk ignored, input never mutated) plus real-path tests against the installed `DFlashDraftModelArgs` (new layout constructs with correct values, old layout unchanged including yarn `rope_scaling`, missing-both-layouts still raises TypeError, idempotent install).
- DFlash suites: `pytest tests/test_dflash_draft_config.py tests/test_dflash_engine.py tests/test_dflash_lifecycle.py tests/test_dflash_multimodal_fallback.py tests/test_dflash_prefill_memory_guard.py -q` → **129 passed**. Full CI-equivalent run (`-m "not slow and not integration"`) → **6823 passed**, 2 pre-existing failures that reproduce identically on clean `main` (M5-only NAX numerics, the known #2135 pair; they pass in CI).
- Revert proof in one fresh process: `from_dict` on the real checkpoint's config raises the issue's TypeError before install and constructs correctly after.

### DIFF
--- a/omlx/engine/dflash.py
+++ b/omlx/engine/dflash.py
@@ -341,9 +341,20 @@ class DFlashEngine(BaseEngine):
             # Native MTP load on the same process would see leftover dflash
             # hooks and crash with TypeError on n_confirmed (issue #1388).
             # Idempotent — only wraps once per process.
+            from ..patches.dflash_draft_config import (
+                install_dflash_draft_config_normalizer,
+            )
             from ..patches.dflash_lifecycle import install_dflash_lifecycle_wrap
 
             install_dflash_lifecycle_wrap()
+            # Newer z-lab drafts ship transformers 5.x-style configs that nest
+            # rope_theta under rope_parameters and block_size under
+            # dflash_config, but DFlashDraftModelArgs requires both at the
+            # config root with no defaults. Without this, load_draft_bundle
+            # crashes with a missing-positional-argument TypeError and
+            # engine_pool falls back to the vlm engine (issue #2317).
+            # Idempotent — only wraps once per process.
+            install_dflash_draft_config_normalizer()
 
             target_bundle = load_target_bundle(
                 self._model_name,
--- a/omlx/patches/dflash_draft_config.py
+++ b/omlx/patches/dflash_draft_config.py
@@ -1,0 +1,139 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Normalize transformers 5.x-nested draft configs for dflash-mlx (issue #2317).
+
+``dflash_mlx.runtime.loading.load_draft_bundle`` -> ``mlx_lm.utils.load_model``
+-> ``DFlashDraftModelArgs.from_dict(config)``. ``from_dict`` filters the raw
+config dict down to the dataclass's annotated fields and calls
+``cls(**filtered)``. ``rope_theta: float`` and ``block_size: int`` are
+required positional fields with no defaults.
+
+z-lab retrained their DFlash draft checkpoints in June 2026 with
+transformers 5.7-style configs: ``rope_theta`` moved from the config root
+into a nested ``rope_parameters`` object, and ``block_size`` moved from the
+root into the existing ``dflash_config`` object. Neither key survives
+``from_dict``'s field filter at its new nested location, so construction
+fails with
+``TypeError: DFlashDraftModelArgs.__init__() missing 2 required positional
+arguments: 'rope_theta' and 'block_size'``, and omlx's engine_pool logs
+"DFlash start failed ... Falling back to vlm engine" instead of running the
+speculative draft. Upstream dflash-mlx (pin 9ca0028, v0.1.10) has no fix.
+
+This module hoists the nested values back to the config root before
+``DFlashDraftModelArgs.from_dict`` runs, so both the new (nested) and old
+(root-level) config layouts construct successfully:
+  - ``rope_parameters.rope_theta`` -> root ``rope_theta`` (transformers 5.x
+    folds ``rope_theta`` into ``rope_parameters``, the successor of the
+    older ``rope_scaling`` dict).
+  - ``rope_parameters`` with a non-"default" ``rope_type``/``type`` -> root
+    ``rope_scaling`` (minus ``rope_theta``), since the draft model consumes
+    yarn-style scaling via ``mlx_lm``'s ``initialize_rope(...,
+    scaling_config=args.rope_scaling)``.
+  - ``dflash_config.block_size`` -> root ``block_size``.
+
+Hoisting is fill-only per key: a root-level key that is set (non-null)
+always wins over its nested counterpart, and an explicit ``null`` counts
+as missing (transformers 4.x-derived configs commonly serialize
+``"rope_scaling": null``). Old-layout configs pass through unchanged. One
+deliberate behavior change for hybrid configs: a config that sets root
+``rope_theta``/``block_size`` but still carries a non-"default"
+``rope_parameters`` and no root ``rope_scaling`` (e.g. a checkpoint
+hand-patched per the issue #2317 workaround) now gains the scaling its
+config declares, where it previously loaded with unscaled RoPE. A config
+that is missing a key at both the root and its nested location still
+fails loudly in ``from_dict`` exactly as it does today — this module only
+widens which configs are recognized, it never invents values.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any
+
+from dflash_mlx.model import DFlashDraftModelArgs
+
+logger = logging.getLogger(__name__)
+
+
+def normalize_dflash_draft_config(config: dict[str, Any]) -> dict[str, Any]:
+    """Hoist transformers 5.x-nested draft config keys back to the root.
+
+    Pure function: never mutates ``config``. Returns a shallow copy with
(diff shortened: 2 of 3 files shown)


jundot:
Thanks for tracking this down and covering both config layouts. I reproduced the missing-field failure against the current z-lab config and confirmed the normalizer restores the real `load_draft_bundle` path without changing root-keyed configs. LGTM. I will merge this now.

Issue #2059, opened 2026-07-01
Memory pressure oscillation with "no evictable models" under sustained long-context inference


andjiang0083:
## Description

Under sustained long-context inference load (Qwen3.6-35B-A3B-oQ6-mtp, ~28GB model, 128GB Mac), oMLX v0.4.4 enters a persistent memory pressure oscillation cycle:

- Memory climbs rapidly from ~75GB → ~92GB (soft pressure) → ~104GB (hard pressure) within seconds
- Enforcer loops on "Hard memory pressure, no evictable models and no loads in progress: requested idle reclaim on 1 scheduler(s)" every second for 10–15 seconds
- Memory eventually drops back to ~75–78GB (likely macOS kernel compression/swap, not oMLX eviction)
- The cycle repeats immediately (5 seconds later)
- During hard pressure, generation speed collapses from ~100 tok/s to 0.5–6.8 tok/s
- Prefill memory guard then starts rejecting ALL requests (including tiny 340-token ones) as "predicted peak would exceed ceiling"
- After the load subsides (~2 minutes idle), normal speed returns

## Environment

- **oMLX version**: v0.4.4
- **OS**: macOS 26.5.1 (Apple Silicon)
- **Hardware**: Mac with 128 GB unified memory
- **Model**: Qwen3.6-35B-A3B-oQ6-mtp (MoE, 35B, 8-bit quant, ~28GB)
- **Memory ceiling**: 107.5 GB (default auto config: soft=91.4GB, hard=102.1GB)
- **Max concurrent**: 1
- **Hot cache**: 5 GB
- **SSD cache**: 185 GB max
- **Idle timeout**: 15 min
- **Memory guard**: Balanced

## Relevant Logs

**Memory pressure oscillation:**

```
17:06:36 - ok -> soft (current=92.5GB, soft=91.4GB, hard=102.1GB, ceiling=107.5GB)
17:06:39 - soft -> hard (current=103.6GB, soft=91.4GB, hard=102.1GB, ceiling=107.5GB)
17:06:39 - WARNING - Hard memory pressure, no evictable models and no loads in progress: requested idle reclaim on 1 scheduler(s).
17:06:40–54 - (repeated every second for 15 iterations)
17:06:55 - hard -> ok (current=78.6GB, soft=91.4GB, hard=102.1GB, ceiling=107.5GB)
17:06:59 - ok -> soft (current=92.4GB)
17:07:03 - soft -> hard (current=104.2GB)
```

[...]

PR #2342, opened 2026-07-23
fix(memory-guard): drain the pressure reclaim at the step boundary under load (completes #2306 for busy servers)

cebertowicz:
## Summary
- On a busy server with pinned/in-use models, hard memory pressure livelocks:
  ```
  Hard memory pressure, no evictable models and no loads in progress: requested idle reclaim on N scheduler(s).
  ```
  once per second until restart, with generation collapsed — the sustained-load wedge reported in #2059.
- #2306 fixed the idle case — the pending reclaim survives until the engine goes idle. But `_process_pending_reclaim` still early-returns while `running / prefilling / waiting` is non-empty, so a server under continuous load never drains it. This adds the under-load half.

## Why phys never recovers
`_shrink_hot_cache_for_pressure` frees hot-cache `mx.array` refs into MLX's buffer pool, which `set_cache_limit(total)` (the #300 guard) pins. The bytes never leave the process, `phys_footprint` doesn't drop, and the enforcer sees "no evictable models" every tick while the reclaimable memory sits in the pool.

## Change
- `scheduler.py`: `request_pressure_reclaim()` — GIL-atomic flag (same idiom as `request_idle_reclaim`), drained in `step()`'s existing periodic-cleanup block via `_sync_and_clear_cache(self._stream)`. `has_work()` stays True while pending so an idle engine still drains it.
- `process_memory_enforcer.py`: at hard pressure, after the shrink, set that flag on each scheduler — when the shrink freed refs, or when the pool alone holds >2 GiB (the already-wedged case). `OMLX_DISABLE_HOTCACHE_RECLAIM=1` restores stock behavior.

**#300 safety:** no new Metal call site. The drain is the same lock-protected, engine-stream-synchronized `_sync_and_clear_cache` the periodic clear already runs every `mlx_cache_cleanup_interval` steps under load — this change just also triggers it at hard pressure. Live KV is untouched (`clear_cache` frees only unreferenced buffers).

## Validation

| Scenario | Stock | Patched |
|---|---|---|
| Pinned 27B, forced 55 GB ceiling, sustained prefills (M3, 256 GB) | spin ×5, never recovers | reclaim fires → `hard -> ok` next tick; 0 spins, 0 forced evictions |
| 3 pinned models, ~3 h agentic soak (M3) | wedged at 199.9 GB until restart | 32 hard-pressure episodes, 32 reclaims, 0 wedges; 213.6 → 113.7 GB in ~1 s |
| Pinned 7B, forced 14 GB ceiling (M4 Max, 48 GB) | 1 hard breach; drained via #2306's idle path when a gap appeared (workload too light to wedge) | 9 pressure reclaims fired under load; **0 panics / SIGABRT**; peak 15.5 → idle 7.1 GB |

The M4 run is the #300 safety check: the step-boundary clear fired repeatedly under load on M4 hardware with no kernel panic. (The livelock repro itself is the M3 rows — a 7B on 48 GB with working SSD eviction self-regulates too well to wedge.)

Unit tests mirror #2306's pattern (pressure flag counts as engine work; consumed at the step boundary even while busy, one-shot; enforcer trigger fires on freed refs or a stranded pool, skips when nothing is reclaimable, and tolerates a mocked `mx`). Full `test_scheduler.py` + `test_process_memory_enforcer.py`: **334 passed** (macOS, M4 Max, py3.13).

Related: #2059, #2179 / #2306, #300.

The scripted repro (pinned-model hard-pressure A/B + the M4 runner) lives on the companion branch [`fix/pressure-reclaim-under-load`](https://github.com/cebertowicz/omlx/tree/fix/pressure-reclaim-under-load/pressure-test) — happy to contribute it under `tests/` in whatever shape you prefer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,8 @@ Handles concurrent requests through mlx-lm's BatchGenerator. Max concurrent requ
 
 Context scaling support for running smaller context models with Claude Code. Scales reported token counts so that auto-compact triggers at the right timing, and SSE keep-alive prevents read timeouts during long prefill.
 
+> **Note**: If you see occasional large prefill events mid-session even though your context should be cached, that is Claude Code, not the server. Recent CC versions quietly clear old tool results from the conversation when context usage gets high, which changes earlier tokens and invalidates any server-side prefix cache from that point on. Keeping auto-compact at its default threshold (rather than raising it with `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE`) reduces how often this happens.
+
 ### Multi-Model Serving
 
 Load LLMs, VLMs, embedding models, and rerankers within the same server. Models are managed through a combination of automatic and manual controls:
--- a/README.md
+++ b/README.md
@@ -189,8 +189,6 @@ Handles concurrent requests through mlx-lm's BatchGenerator. Max concurrent requ
 
 Context scaling support for running smaller context models with Claude Code. Scales reported token counts so that auto-compact triggers at the right timing, and SSE keep-alive prevents read timeouts during long prefill.
 
-> **Note**: If you see occasional large prefill events mid-session even though your context should be cached, that is Claude Code, not the server. Recent CC versions quietly clear old tool results from the conversation when context usage gets high, which changes earlier tokens and invalidates any server-side prefix cache from that point on. Keeping auto-compact at its default threshold (rather than raising it with `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE`) reduces how often this happens.
-
 ### Multi-Model Serving
 
 Load LLMs, VLMs, embedding models, and rerankers within the same server. Models are managed through a combination of automatic and manual controls:
--- a/omlx/process_memory_enforcer.py
+++ b/omlx/process_memory_enforcer.py
@@ -1015,7 +1015,13 @@ class ProcessMemoryEnforcer:
         already-wedged case where hot cache is empty but pooled buffers are
         stranded).
         """
-        pool_bytes = mx.get_cache_memory()
+        raw_pool = mx.get_cache_memory()
+        try:
+            pool_bytes = int(raw_pool)
+        except (TypeError, ValueError):
+            # A non-numeric reading (e.g. a wholesale-mocked mx in unit
+            # tests) cannot justify a clear; treat it as an empty pool.
+            pool_bytes = 0
         if freed_hot <= 0 and pool_bytes <= self._POOL_RECLAIM_FLOOR:
             return
         requested = 0
--- a/omlx/process_memory_enforcer.py
+++ b/omlx/process_memory_enforcer.py
@@ -1433,8 +1433,8 @@ class ProcessMemoryEnforcer:
             # freed bytes never leave the process and phys_footprint does not
             # drop — the enforcer then wrongly concludes "no evictable models"
             # and livelocks until restart. Env gate
-            # OMLX_DISABLE_HOTCACHE_RECLAIM=1 restores stock behavior.
-            if os.environ.get("OMLX_DISABLE_HOTCACHE_RECLAIM") != "1":
+            # OMLX_DISABLE_PRESSURE_RECLAIM=1 restores stock behavior.
+            if os.environ.get("OMLX_DISABLE_PRESSURE_RECLAIM") != "1":
                 self._request_scheduler_cache_reclaim(freed_hot)
             if freed_hot > 0:
                 current = self._current_usage_bytes()
--- a/omlx/process_memory_enforcer.py
+++ b/omlx/process_memory_enforcer.py
@@ -30,6 +30,7 @@ from __future__ import annotations
 import asyncio
 import inspect
 import logging
+import os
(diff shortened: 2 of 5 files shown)


JimStenstrom:
Independent corroboration for the mechanism here, plus measurements from running this exact branch against a different pressure regime — posting since this completes #2306 and I'd been instrumenting the same pool behavior from another angle.

Background: while investigating TurboQuant-KV memory behavior we bracketed every TQ cache lifecycle call with memory probes under 8 concurrent 16k-token prefills (Qwen3.6-35B-A3B oQ4, TQ-KV 4-bit, max_concurrent_requests 16, MTP off, M5 Max 128 GB, guard balanced: soft 96.8 / hard 102.1 / ceiling 107.5 GB). mx active memory never exceeded 21.6 GiB at any of 2,763 bracketed call boundaries (peak reading 23.9 GiB), while `mx.get_cache_memory()` at the same boundaries peaked at **80.8 GiB** — 21.6 active + 80.8 pooled ≈ the 102.7–103.3 GB the enforcer acted on, so the accounting closes. In a follow-up arm we clamped `set_cache_limit` to 8 GB, confirmed in-process it took (log shows previous ≈ the 121.6 GiB default), and the pool still peaked at 80.79 GiB — so under a saturated prefill wave the pooled bytes aren't bounded by the cache limit at all, which fits buffers freed by MLX but not yet returnable while their command buffers are in flight. In this regime the enforcer's failure mode is not the #2059 livelock: it aborts the whole wave.

A/B of this branch at that exact shape (same day, same deps; stock = this PR's merge base dceae1d2; PR = head 27587115; code identity verified per arm):

| arm | outcome | enforcer peak | pressure reclaim |
|---|---|---|---|
| stock main | 0/8 complete, all aborted | 102.5 GB | n/a |
| PR #2342 (×2) | 0/8 complete, all aborted | 103.3 / 103.2 GB | fired 1× per run: `freed_hot=0.0GB, pool=80.1GB` / `pool=80.0GB` |

Three findings, in increasing order of interest:

1. **Your pool-only trigger works exactly as designed, and its own log line independently confirms the census number**: `pool=80.1GB` at the hard crossing, from a second instrument, matching the 80.8 GiB we measured via `mx.get_cache_memory()`. At the moment the enforcer aborted 8 user requests, ~77% of its own phys reading was reclaimable pool (79.3–80.1 of 102.7–103.3 GB across runs).

2. **The abort consumer isn't covered — by one tick, not by design**: the enforcement pass that requests the reclaim aborts the requests in the same pass (log: soft→hard, "Requested pressure cache reclaim", and "aborted 8 request(s)" all in the same millisecond), so the step-boundary drain never gets a chance. Worse for any transition-armed trigger: the final climb (~5+ GB/s) can cross both watermarks between two 1s polls — in one run the level went ok→hard in a single tick with no soft state ever existing.

3. **When the drain does get a chance, it works — measured end-to-end on this branch**: with a local probe arm (env-gated one-liner requesting the same drain at the soft crossing; not proposing it — per finding 2 it's a coin flip whether a soft tick exists), one run caught a soft tick at 98.6 GB and your step-boundary drain executed under full prefill load: **98.6 → 28.8 GB one enforcer tick later, ~70 GB returned to the OS mid-wave, zero Metal errors (M5, one observed drain-under-load)** — and all 8 prefills then completed, which no TQ-on cell at this shape had done in any of our runs. The pool re-accumulated over the remaining ~19 s of prefill (~3.9 GB/s), the second climb went ok→hard in one tick, and the same-pass abort killed the wave mid-decode. So the drain is fast, and showed no Metal errors in the one drain we observed under load; the remaining gap is purely that the hard-pressure escalation doesn't consult the pool (or wait one step boundary) before aborting.

For this PR's scope — the #2059 sustained-load livelock, where pressure persists across many ticks and the drain will always get its step boundary — all of the above is corroborating: mechanism confirmed by intervention, machinery validated under load on a third hardware generation. The fast-climb abort case is a different consumer of the same mechanism; I'll write it up as a separate issue with this data rather than scope-creep this PR. Happy to share the harness/logs if useful.

jundot:
Thanks for closing the busy half of this. The step-boundary drain is exactly the missing piece: the #2306 idle reclaim never fires on a busy server, and the periodic clear is gated on a 512-step interval, so a server stuck at hard pressure had no timely way to return the pool. Routing it through the same guarded clear the periodic path already uses keeps the #300 safety story intact.

Jim's independent numbers match your own log line, and his drain-under-load measurement settles the mid-decode safety question. The fast-climb abort case he describes is a different consumer of the same mechanism, so I agree it belongs in a separate issue.

Two small things I'll fix directly on your branch before merging. First, the env gate also covers the stranded-pool case where no hot cache is involved, so I'm renaming OMLX_DISABLE_HOTCACHE_RECLAIM to OMLX_DISABLE_PRESSURE_RECLAIM while the name hasn't shipped anywhere yet. Second, the README note about Claude Code re-prefills rode in from your branch point, but that note was intentionally dropped from main, so I'll remove it here to keep this PR scoped to the memory-guard change. I'll merge this after that.

Issue #2132, opened 2026-07-09
Benchmark regression in Token Generation (TG) vs improvement in Prompt Processing (PP) on v0.5.0.dev2


gpando:
## Description
While benchmarking the Qwen3.6-35B-A3B model (both 8-bit and 4-bit UD variants) on an Apple M2 Ultra (60c, 64GB) using the new v0.5.0.dev2, we observed a significant performance trade-off compared to v0.4.4.
While prompt ingestion and initial responsiveness have improved markedly, text generation speed has suffered a noticeable regression.
## Key Observations

* Prompt Processing (PP) & TTFT Improved:
* For 8-bit (1k context), PP speed increased by ~28% (from 862.9 tok/s to 1,110 tok/s).
   * This directly lowered the Time To First Token (TTFT) from ~1,187 ms down to ~923 ms in short contexts.
* Token Generation (TG) Regression:
* Across all tests, text generation speed dropped by ~15%.
   * For 8-bit (1k), TG dropped from 83.4 tok/s to 71.1 tok/s.
   * For 4-bit UD (1k), TG dropped from 90.5 tok/s down to 77.3 tok/s.
* Quantization Efficiency Loss:
* In v0.4.4, dropping from 8-bit to 4-bit yielded a clear speed boost (83.4 to 90.5 TG tok/s).
   * In v0.5.0.dev2, this scaling advantage is mostly lost, suggesting a bottleneck in decoding lighter quantized weights.
* Memory Footprint:
* Peak memory usage remains identical between software versions (~35.4 GB for 8-bit / ~20.4 GB for 4-bit), confirming the changes are purely compute/scheduling related.

## Environment

* Hardware: Apple M2 Ultra (60 GPU Cores), 64 GB Unified Memory
* OS: macOS 26.5
* Software Versions Compared: v0.4.4 vs v0.5.0.dev2

v0.4.4:
* https://omlx.ai/benchmarks/wifv4kd2
* https://omlx.ai/benchmarks/00xkgbxe
* https://omlx.ai/benchmarks/9k60jkns
* https://omlx.ai/benchmarks/oq6z80fv

v0.5.0.dev2: 
* https://omlx.ai/benchmarks/xq7my6xt
* https://omlx.ai/benchmarks/0f3mb1d5
* https://omlx.ai/benchmarks/jyvaosb5
* https://omlx.ai/benchmarks/67op8oc1

jundot:
[...]

Issue #965, opened 2026-04-26
Mismatch in token generation speed: performance test vs chat display with Dflash model configuration


ghl116:
**Describe the bug**
Mismatch in token generation speed: performance test vs chat display with Dflash model configuration
chat speed:    gen 23 tok/s
performance speed:  181 token/s

**To Reproduce**
mlx type model :  Qwen3.6-35B-A3B-4bit
dflash model: Qwen3.6-35B-A3B-DFlash

<img width="1336" height="1030" alt="Image" src="https://github.com/user-attachments/assets/72111ad2-c1e2-4d9f-9a50-6658c77e9086" />

<img width="1320" height="1018" alt="Image" src="https://github.com/user-attachments/assets/b94dfbd6-0a00-4ace-9341-9237b0b7054e" />

<img width="2526" height="1098" alt="Image" src="https://github.com/user-attachments/assets/10c273fd-ed4c-4a45-99e7-13896c9d1d2b" />

<img width="1892" height="624" alt="Image" src="https://github.com/user-attachments/assets/8a144b0a-ec4a-4632-8958-e8f1797872d6" />
**Expected behavior**
 enable dflash :  token generation speed to  181 token/s

**Desktop (please complete the following information):**
 - macOS Version: 15.7.2   m4pro 16c 48G
 - oMLX Version 0.3.6

fritskarl:
Confirmed. Looks like issue with reported pp1024/tg128 performance test results, pp2048/tg128 shows more realistic results.

See https://omlx.ai/benchmarks/ibg7teli (scroll to bottom) for instance.

ghl116:
> Confirmed. Looks like issue with reported pp1024/tg128 performance test results, pp2048/tg128 shows more realistic results.
> 
> See https://omlx.ai/benchmarks/ibg7teli (scroll to bottom) for instance.

without Dflash model configuration, the speed is consistence

jundot:
Quick update: I landed a full DFlash overhaul in 69becb31. It upgrades the bundled dflash-mlx to 0.1.5.1 (new runtime_context API, prefix cache, verify_qmm Metal kernel, and client-abort propagation through the MLX executor). Shipping in the next release.

Please retry once the release is out, and reopen this issue or file a new one if the same problem persists.

jundot:
One more note: DFlash speedup is highly prompt-dependent by design. Math-heavy prompts (high token predictability) tend to see big gains, while non-English text or freeform writing often sees neutral or negative results.

[...]

PR #2343, opened 2026-07-23
fix(benchmark): build prompts from natural text instead of repeated filler

### LINKED ISSUE #2254: 0.5.1  Qwen3.6-27B-0Q4-MTP suddenly slow down
**Describe the bug**
7/13 upgrade to 0.5.1
open the lighting-mtp of  Qwen3.6-27B-0Q4-MTP
the pp tps & tgs both speed up, 

7/15 but today i don't know what's wrang, it slow down to before.

**To Reproduce**
May, i change the concurrent from 1 to 3 to 2, or. open the turboquant KV cash, or other setting.

**Expected behavior**
keep speed up

**Screenshots**
benchmark test 0714
Image

benchmark test 0715
Image

 - macOS Version: 26.4 (25E246)
 - oMLX Version 0.5.1

JimStenstrom:
Fixes #2254.

## Summary

- Benchmark prompts are now sliced from a bundled natural-text corpus (`omlx/admin/bench_corpus.txt`, public-domain Project Gutenberg #2701 with the boilerplate stripped) instead of one 30-token filler sentence repeated ~2,200×. The UUID cache-buster prefix and exact-token truncation are kept; every run gets the same corpus slice (only the UUID prefix, and the token or two it displaces, varies).
- The tokenizer-less external-endpoint path slices the same corpus by a chars-per-token heuristic (it already reports the endpoint's actual `usage.prompt_tokens`).

## Why

On speculative-decode models the benchmark's repeated filler is a degenerate best case: the drafts verify almost every cycle, so the reported tg is a ceiling no real prompt can reach. Measured on Qwen3.6-27B-oQ4 with Lightning MTP (M5 Max, 65,536-token prompts, gen 1280, single request):

| prompt | MTP accept | tok/cycle | decode tok/s |
|---|---|---|---|
| repeated filler (current bench) | 99.0% | 3.72 | 53.4 |
| natural-text corpus (this PR) | 88.3% | 2.74 | 41.7 |
| ~65k of real source code via `/v1/chat/completions` | 68.4% | 1.89 | 28.1 |

The forward-pass cost per cycle is within ~3% of its mean across the three rows — the spread is draft acceptance. #2254's reporter hit exactly this: their bench showed 38.5 tg @65k while every real request got ~18, and their own server logs show the same acceptance signature (98.9–99.2% on the bench vs 71.8–76.6% on their real prompts); the matched-KV per-cycle control is the table above. The same mechanism was visible in #965 for DFlash ("performance test 181 tok/s vs chat 23"), where acceptance being prompt-dependent was noted but the benchmark's prompt remained the extreme case — and it measures the same way. The same 27B with the z-lab DFlash draft (MTP off, draft_window 1024, verify=adaptive, 65,536-token prompts, gen 1280, single request):

| prompt | DFlash accept | tok/cycle | decode tok/s |
|---|---|---|---|
| repeated filler (current bench) | 92.8% | 13.9 | 139.3 |
| natural-text corpus (this PR) | 83.3% | 6.0 | 61.1 |
| ~56k of real source code via `/v1/chat/completions` | 62.5% | 2.7 | ~36 |

Per-cycle cost is equal within 2.1% between the two bench arms (100.1 vs 98.0 ms), so the entire 2.28× step is draft acceptance. DFlash drafts longer blocks than MTP, which is why its filler inflation is larger — consistent with #965's 7.9× gap. Non-speculative models are unaffected (prefill cost is content-independent; plain decode doesn't depend on prompt predictability) — for them this change is neutral; prefill throughput was identical across arms here (587.4 vs 586.9 tok/s).

A paragraph-shuffled variant of the corpus was probed and rejected: shuffling *raised* acceptance (92.6% vs 88.3%), so verbatim-memorized continuation is not the driver and the plain text is both simpler and closer to real traffic.

The remaining gap between the corpus row (88.3%) and real chat traffic (68–77%) is genuine workload sensitivity — the bench decodes greedily while real clients sample at temperature, and code is less predictable than prose. Surfacing the per-run acceptance rate in bench results would make that visible; `GenerationOutput` has no MTP stats field today, so that is left as a separate follow-up if wanted.

## Changes

- `omlx/admin/benchmark.py`: `_generate_prompt` / `_generate_external_prompt` slice the corpus (wraparound only past ~280k tokens); filler constants removed; `_load_bench_corpus` + module docs. The corpus is tokenized once per tokenizer (weakref-keyed single-entry cache — a bench run builds prompts for many lengths plus warmup/batch, and the tokenization is identical every time); an empty corpus raises instead of looping.
- `omlx/admin/bench_corpus.txt`: the corpus (1.2 MB text; ~287k tokens in common BPE vocabs — covers the 200k max prompt length without wrapping).
- `pyproject.toml`: ship `bench_corpus.txt` via `omlx.admin` package-data.
- `tests/test_benchmark.py`: `TestPromptRealism` — a sliding-window repetition detector on the internal and external builders (both RED on the old filler), determinism-mod-UUID-prefix, and corpus presence/naturalness.

## Trade-offs / disclosure

- Leaderboard continuity: tg rows for MTP/DFlash models will step down to honest values on runs with this change; non-speculative models are unaffected. Same class of instrument-truth discontinuity as the producer- timestamp fix (a26bc8e5 / #2132).
- Wheel grows by the corpus (~1.2 MB raw, compresses well). Happy to swap the corpus for any text you'd prefer; the tests only assert naturalness properties, not the specific book.
- Bench engine selection (`force_lm=True` whenever vlm-mtp is off, benchmark.py:988) means the bench measures `BatchedEngine` while models of this class serve on `VLMBatchedEngine`. Measured small here (per-cycle 69.0 vs 66.7 ms at 65k — the serving engine was actually faster), so left untouched; named for completeness.
- Building the prompt from two `encode()` calls leaves one BOS token mid-prompt on BOS-adding models (the prefix's). Content-negligible for a throughput benchmark; avoiding it would need backend-specific encode kwargs (mistral-common tokenizers don't take `add_special_tokens`), so it is deliberately left alone.

## Tests

- `pytest tests/test_benchmark.py tests/test_bench_external_routes.py tests/test_benchmark_sse_replay.py tests/test_accuracy_benchmark.py` — 133 passed (5 new: sliding-window repetition detectors for both prompt builders, determinism-mod-prefix, corpus presence/naturalness, tokenize-once cache).
- New tests verified RED with the fix reverted (4 failed on stock, only the determinism property passing).
- Live A/B on this branch (isolated instance, M5 Max): the table above; full protocol and raw logs recorded locally.
- CI note: expect the two `test_e2e_streaming.py` namespaced tool-call failures currently red on main itself (first red after 093e7df1) — unrelated to this diff, which doesn't touch the tool-call path.


jundot:
Thanks for tracing the MTP/DFlash benchmark inflation and replacing the repeated filler with natural text. One thing I’ll fold into a follow-up on main is exact prompt-length handling: local benchmarks should pass token ID arrays through Batched, VLM, and DFlash engines without a decode/re-encode round trip, then fail loudly unless both the submitted and reported prompt lengths match the requested `pp`. External OpenAI-compatible endpoints do not expose a standard tokenizer, so I’ll separate `requested_pp` from the actual `usage.prompt_tokens` instead of presenting an estimate as an exact `pp1024` result. This does not block the core fix. LGTM, merging.

Issue #1078, opened 2026-05-06
[Feature Request] GemFilter-style partial-layer scorer for SpecPrefill on MoE targets without same-family small drafts


0xAlcibiades:
**Related issues / PRs surveyed**

- #500 (open) — EAGLE-3 FR (decode-side speculation, complementary not overlapping).
- #713 (open) — SpecPrefill activates on Qwen3.5-35B-A3B (3B-active MoE) with a Qwen3.5-4B draft but silently fails on Qwen3.5-122B-A10B (10B-active MoE) with the same draft. Direct evidence that current SpecPrefill is sensitive to target/draft sizing on MoE.
- #1045 (open) + PR #1044 — admin UI claimed "for MoE models" but the implementation has no MoE-specific gating. PR removes the misleading label rather than adding MoE handling. Confirms SpecPrefill was designed for dense+same-family-draft and grafted onto MoE without the architecture-aware path.
- #548 (open) — SpecPrefill quality regression in Claude Code on Qwen 35B family after 3-4 turns. Same shape as the multi-turn agent regression behind this FR.
- #766, #1061 (open) — orthogonal SpecPrefill bugs.

No issue or PR proposes a partial-layer / self-scoring SpecPrefill mode.

**Problem:** oMLX implements Liu et al. SpecPrefill (https://arxiv.org/abs/2502.02789), whose central premise is a same-family small draft whose attention transfers to the target. That premise breaks on Qwen 3.6 35B-A3B (3B active per token): no same-family draft exists below 27B, and cross-family Qwen 3.5 0.8B at keep_pct=0.5 produces ~70% failure on multi-turn agent traffic in our testing. For this target SpecPrefill is currently unusable.

**Use case:** Anyone running Qwen 3.6 35B-A3B (and similar MoE targets that ship with no smaller same-family sibling). The 27B dense Qwen 3.6 sibling has working cross-family options (Qwen 3.5 4B/9B), so this FR is mostly about A3B-class models — but the same code path applies to both.

**Proposed solution:** Add GemFilter-style scoring (Shi et al. 2024, https://arxiv.org/abs/2409.17422, code at https://github.com/SalesforceAIResearch/GemFilter) as an alternative scorer mode in \`omlx/patches/specprefill.py\`. Use the target's own first N attention layers for importance scoring instead of running a separate draft.

\`ModelSettings\` additions:
- \`specprefill_scorer: Literal["draft", "self_layers"] = "draft"\`
[...]

PR #2346, opened 2026-07-23
feat(specprefill): memoize importance-score selection by prompt hash (#1079)

### LINKED ISSUE #1079: [Feature Request] Memoize SpecPrefill importance-score selection by prompt hash
**Related issues / PRs surveyed**

- #1070 (open) — "Cache doesn't store properly" — about the omlx prefix cache (KV layer), not SpecPrefill scoring; orthogonal.
- #766, #1061 (open) — SpecPrefill bugs unrelated to caching.
- #500 (open) — EAGLE-3 FR, decode-side, unrelated.
- #1078 (open) — GemFilter-style partial-layer scorer FR; pairs naturally with this one (same memoization key works).

No issue or PR proposes caching the SpecPrefill importance / selection result.

**Problem:** SpecPrefill's score → chunk-select → restore-pos-ids pipeline is deterministic given a fixed (prompt token sequence, draft model, lookahead steps, keep rate, chunk size, pooling kernel). oMLX caches the draft model's KV across requests via `_draft_prefix_cache` (`scheduler.py:2754`) but recomputes the entire scoring and selection for every request — even when the prompt is byte-identical to a recent one. Multi-turn agent retries on the same prompt pay the full ~6s scoring cost repeatedly.

**Use case:** Agent loops that retry on bad output (Claude Code, OpenClaw, custom agents). Observed in production: 4 identical 23,444-token prompts re-scored in 90 seconds, each producing the same `selected 11744/23444` selection, total ~24s wasted on redundant scoring:

```
22:40:38 - SpecPrefill: scored 23444 tokens in 4.1s, selected 11744/23444 (keep=50%, draft cache hit 22528)
22:40:48 - SpecPrefill: sparse prefill 11744/23444 conv tokens in 6.0s
22:40:59 - SpecPrefill: scored 23444 tokens in 4.1s, selected 11744/23444 (keep=50%, draft cache hit 22528)  <- same prompt
22:41:05 - SpecPrefill: sparse prefill 11744/23444 conv tokens in 6.0s
22:41:12 - SpecPrefill: scored 23444 tokens in 4.1s, selected 11744/23444 (keep=50%, draft cache hit 22528)  <- same prompt
22:41:18 - SpecPrefill: sparse prefill 11744/23444 conv tokens in 6.0s
22:41:18 - SpecPrefill: scored 23444 tokens in 4.1s, selected 11744/23444 (keep=50%, draft cache hit 22528)  <- same prompt
```

The draft KV cache hit at 22,528 of 23,444 tokens proves the draft model isn't doing the heavy lifting — the wasted time is in the attention aggregation and chunk selection, both fully deterministic on the inputs.

**Proposed solution:** Add an LRU cache keyed by

```
(sha256(prompt_token_ids), draft_model_id, n_lookahead, keep_pct, chunk_size, pool_kernel)
-> selected_token_indices
```

inside or just upstream of the `score_tokens` + `select_chunks` calls at `scheduler.py:2773-2779`. On hit, skip scoring entirely and reuse the cached `specprefill_indices`. Default size ~64-128 entries; per-entry footprint is the selection vector (~tens of KB), trivial vs the model weights.

Hit rate is workload-dependent but the baseline saving on retry storms is exactly the wasted time in the log above (~24s on a 4-retry sequence) -> returns 1x scoring + 3x cache hit instead of 4x full scoring.

**Alternatives:** Embed the importance vector in `_draft_prefix_cache` alongside the KV state (tighter coupling with z-lab's prefix cache hierarchy, more invasive change). Or detect duplicate in-flight prompts and serialize them so only one scoring runs (only helps concurrent dupes, not sequential retries).

LRU on the selection result is the simplest and works for both regimes.

**Extra context**

- Determinism argument: `score_tokens` (`omlx/patches/specprefill.py:442`) -> `aggregate_attention_score` (max-mean across heads/layers, mean across lookahead) -> `select_chunks` (deterministic top-K on fixed-size chunks). All pure functions of the inputs.
- Doesn't help cold prompts or single-shot workloads — only retry storms and exact-repeat traffic.
- Pairs naturally with #1078 (GemFilter scorer): partial-layer scoring is also deterministic, same memoization key works.

ethannortharc:
## Summary

Closes #1079.

- Memoize the SpecPrefill `score → select` result (the selected token indices) so a byte-identical prompt reuses the prior selection instead of re-running the scoring pipeline.
- Keyed on the deterministic scoring inputs: a content hash of `tokens_to_score`, the draft model, `keep_pct`, and the chunk/lookahead/pool sizes.
- New module `omlx/specprefill/selection_cache.py`, wired into `scheduler._try_specprefill_scoring`. `draft.py` is untouched.

## Motivation

SpecPrefill's `score → select` pipeline is a pure function of `(tokens_to_score, draft model, keep_pct, chunk_size, n_lookahead, pool_kernel)`, but it is re-run for every request even when the prompt is byte-identical to a recent one. The draft KV cache (`_draft_prefix_cache`) only skips the draft **prefill**; the lookahead + importance + chunk-select scoring on top of it still runs every time.

In agent retry loops (a client resending the same prompt after a bad generation), this repeats the full multi-second scoring for an identical prompt:

```
22:40:38  SpecPrefill: scored 23444 tokens in 4.1s, selected 11744/23444 (draft cache hit 22528)
22:40:59  SpecPrefill: scored 23444 tokens in 4.1s, selected 11744/23444   <- same prompt, re-scored
22:41:12  SpecPrefill: scored 23444 tokens in 4.1s, selected 11744/23444   <- again
```

The `draft cache hit` shows the draft KV is already reused; it is the scoring *on top of it* that is redundant.

## Approach

Cache the **output** of scoring (the selected indices), not its input. This is orthogonal to the draft KV cache — one caches scoring's input, the other its result — and it is the only layer that can short-circuit the scoring: the KV caches are keyed (directly or indirectly) on the selection, which is what scoring *produces*, so nothing keyed on it can be looked up without first re-deriving it. The selection cache keys on the raw prompt, which is available *before* scoring.

- The **key** is built from `tokens_to_score` — the system-prefix-stripped suffix scoring actually runs on (`plan_specprefill_scoring`), since the selected indices are relative to it — plus the draft model, `keep_pct`, and the scoring constants (`chunk_size=32`, `n_lookahead=8`, `pool_kernel=13`). A `scorer` field reserves room for a future self-layer scorer (#1078) to share the same memoization.
- **On a hit**, only the selection is reused; the per-request position bookkeeping (`specprefill_position_offset`, `_specprefill_system_tokens`) is recomputed from *this* request's `cached_tokens`, since prefix-cache warmth can differ run to run.
- **Values** are stored as plain `int` lists (never `mx.array`), so entries don't pin Metal buffers or capture lazy graphs.
- A **failed scoring** (`specprefill_indices is None`) is not cached, so a transient failure cannot poison future requests.
- The cache is bounded (LRU, 64 entries) and cleared on draft-model swap (`set_specprefill_draft_model`) and on teardown.
- As a side effect, freezing the selection also stabilizes it: the current path samples lookahead tokens unseeded (`temp=0.6, top_p=0.95`), so the specific indices can otherwise drift run to run for the same prompt.

## Tests

New `tests/test_specprefill_selection_cache.py` (19 tests): LRU eviction/recency, key determinism and sensitivity to every scoring input, per-request recompute on a hit, failed-scoring-not-cached, and a scheduler-integration test that drives the real `_try_specprefill_scoring` twice with an identical prompt and asserts the draft scoring runs **exactly once**.

```
uv run pytest tests/test_specprefill*.py tests/test_scheduler.py \
  tests/test_engine_core.py tests/test_engine_preflight.py tests/test_per_engine_threads.py
=> 382 passed
```

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -1666,6 +1666,12 @@ class Scheduler:
 
         # SpecPrefill: draft model for attention-based sparse prefill
         self._specprefill_draft_model: Any | None = None
+        self._specprefill_draft_model_name: str | None = None
+        from .specprefill.selection_cache import SpecPrefillSelectionCache
+
+        # #1079: memoize the score->select result so identical prompts (agent
+        # retry loops, shared prefixes) skip the ~seconds-long re-scoring.
+        self._specprefill_selection_cache = SpecPrefillSelectionCache()
         self._draft_paged_ssd_cache_manager: Any | None = None
         # Track active specprefill request for RoPE cleanup
         self._specprefill_active_request_id: str | None = None
@@ -6468,6 +6474,9 @@ class Scheduler:
                 "Could not close the previous SpecPrefill draft SSD cache manager"
             )
         self._specprefill_draft_model = draft_model
+        self._specprefill_draft_model_name = draft_model_name
+        # Selections are specific to this draft model; drop any from a prior one.
+        self._specprefill_selection_cache.clear()
         self._draft_prefix_cache: Any | None = None
         if not draft_model_name:
             logger.info(
@@ -6908,19 +6917,45 @@ class Scheduler:
             return
 
         from .specprefill.draft import run_specprefill_draft_scoring
+        from .specprefill.selection_cache import build_selection_key, memoize_scoring
 
-        run_specprefill_draft_scoring(
+        def _run_scoring() -> None:
+            run_specprefill_draft_scoring(
+                request=request,
+                plan=plan,
+                draft_model=self._specprefill_draft_model,
+                draft_prefix_cache=self._draft_prefix_cache,
+                model_id=self.config.model_name,
+                prefill_step_size=self.config.prefill_step_size,
+                stream=self._stream,
+                extract_cache_states=self._extract_cache_states,
+                sync_and_clear_cache=lambda: _sync_and_clear_cache(self._stream),
+                log=logger,
+            )
+
+        # #1079: an identical prompt (same draft model + scoring params) yields
+        # the same selection, so reuse it instead of re-running the ~seconds-long
+        # lookahead + importance + chunk-select pipeline. The draft KV cache
+        # (fed as existing_cache above) only skips the draft *prefill*; the
+        # scoring on top of it still runs -- that is what this short-circuits.
+        selection_key = build_selection_key(
+            draft_model_id=self._specprefill_draft_model_name or "unnamed-draft",
+            tokens_to_score=plan.tokens_to_score,
+            keep_pct=plan.keep_pct,
+        )
+        reused = memoize_scoring(
+            cache=self._specprefill_selection_cache,
+            key=selection_key,
             request=request,
             plan=plan,
-            draft_model=self._specprefill_draft_model,
-            draft_prefix_cache=self._draft_prefix_cache,
-            model_id=self.config.model_name,
-            prefill_step_size=self.config.prefill_step_size,
-            stream=self._stream,
-            extract_cache_states=self._extract_cache_states,
-            sync_and_clear_cache=lambda: _sync_and_clear_cache(self._stream),
-            log=logger,
+            run_scoring=_run_scoring,
         )
+        if reused:
+            logger.info(
+                "SpecPrefill: selection cache hit, reused "
+                f"{request.specprefill_indices.shape[0]}/{plan.n_to_score} "
+                "tokens (skipped re-scoring)"
+            )
 
     def _cleanup_specprefill(self, request_id: str) -> None:
         """Clean up SpecPrefill RoPE patches when a request finishes."""
@@ -9984,6 +10019,7 @@ class Scheduler:
         self.paged_cache_manager = None
         self._draft_prefix_cache = None
         self._specprefill_draft_model = None
+        self._specprefill_selection_cache.clear()
         self.block_aware_cache = None
         self.memory_monitor = None
         self._boundary_snapshot_store = None
@@ -10054,6 +10090,7 @@ class Scheduler:
         self._close_specprefill_draft_cache_manager()
         self._draft_prefix_cache = None
         self._specprefill_draft_model = None
+        self._specprefill_selection_cache.clear()
         if self.paged_ssd_cache_manager is not None:
             self.paged_ssd_cache_manager.close()
             self.paged_ssd_cache_manager = None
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for working on this and for the thorough tests. After tracing the runtime path, I do not think this cache will help normal multi-turn conversations. It only hits when `tokens_to_score` is exactly identical. A conversation changing from `A-B-C-D` to `A-B-C-D-E` changes the hash, and the old selection cannot be safely reused because the new ending can change token importance. The existing draft KV cache already reuses the shared prefix.

Exact retries are the main remaining case, but a normal target prefix-cache hit usually leaves too few tokens to enter SpecPrefill at all. The scorer is also stochastic, so the same selected count does not prove that the same positions were selected. Given the narrow hit condition and added memory and lifecycle complexity, I do not think this should be merged.

ethannortharc:
> Thanks for working on this and for the thorough tests. After tracing the runtime path, I do not think this cache will help normal multi-turn conversations. It only hits when `tokens_to_score` is exactly identical. A conversation changing from `A-B-C-D` to `A-B-C-D-E` changes the hash, and the old selection cannot be safely reused because the new ending can change token importance. The existing draft KV cache already reuses the shared prefix.
> 
> Exact retries are the main remaining case, but a normal target prefix-cache hit usually leaves too few tokens to enter SpecPrefill at all. The scorer is also stochastic, so the same selected count does not prove that the same positions were selected. Given the narrow hit condition and added memory and lifecycle complexity, I do not think this should be merged.

Thanks for the detailed trace — the stochasticity point is fair, and it's the one I got wrong. `_lookahead_decode` samples unseeded (`temp=0.6, top_p=0.95`), so `score → select` is not the pure function I described in the module docstring; memoizing it freezes one sampled selection rather than transparently caching a deterministic one. The `selected 11744/23444` line in #1079 also proves nothing about positions — `keep_n = ceil(n_chunks * keep_pct)` fixes that count regardless of which chunks win. Agreed on both.

One correction on the admission path, for the record: SpecPrefill requests never populate the target prefix cache — `scheduler.py:8891` drops `raw_cache` (sparse KV vs. full-token block hash), and `scheduler.py:5554` does the same on the boundary-store fallback. So an exact retry re-enters with `remaining_tokens` = the full prompt and re-scores, which is the mechanism behind the four identical 4.1s scorings in the issue log. That doesn't make retries common though, which was your actual point.

Let me take another look before we close this. I'd rather go after what the 4.1s is actually made of — with a 22.5k-token draft cache hit, the remaining cost is the lookahead decode plus `_compute_importance` over the full K cache, and every request pays that, including the multi-turn path a selection cache can never help. I'll update this PR or close it in favor of something that helps the common case.

PR #2348, opened 2026-07-23
feat(omlx-mac): add sub-key generation controls

tobomobo:
## Summary

Add generate/regenerate and copy controls to the macOS sub-key creation form, reusing the existing key generator, clipboard button, and localization.

This makes sub-key creation match the main API key flow without backend changes.

## Testing

- Native macOS test suite
- Manual verification in the debug app


jundot:
Thanks for adding this. Reusing APIKeyGenerator and making the copy button label configurable keeps the change small while preserving existing callers. I verified it against current main with the focused macOS tests. LGTM, merging.

Issue #2283, opened 2026-07-18
Qwen3.6 performance degradation


bnightstar:
**Describe the bug**
I updated from 0.4.4 to 0.5.1 there is severe performance degradation in the Qwen3.6-35B.

0.4.4 Benchmark - https://omlx.ai/benchmarks/0q069sdg
0.5.1 Benchmark - https://omlx.ai/benchmarks/n1d67582

Very noticeable in both PP and TG perf degradation. 

**To Reproduce**
Benchmark against the model in 0.5.1 

**Expected behavior**

No performance degradation 

**Desktop (please complete the following information):**
 - macOS Version: macOS 26.5.2
 - oMLX Version: 0.5.1

JimStenstrom:
A data point from the two benchmark links themselves: they were run at different benchmark context lengths, and both numbers depend strongly on that. Each page shows it as its labeled Context field — **16,384** on the 0.4.4 run, **65,536** on the 0.5.1 run. Same machine, same model, but a 16k benchmark compared against a 64k one.

For decode, that context difference alone is enough to explain the drop. Qwen3.6-35B-A3B is a GDN hybrid where only the full-attention layers pay a per-token cost that grows with KV length, so tg falls with context by design. An independent measurement on the same model family in #2215 (M4 Max, single request, TQ off) shows tg dropping 110.9 → 74.8 tok/s from pp16384 to pp65536 (−33%); your two runs show 75.9 → 50.4 (−34%). A very similar proportion across a different machine and quant — consistent with the decode cost being set by context on this architecture rather than by the version.

[...]

PR #2352, opened 2026-07-23
feat(sdpa256): log the first tiled-route engagement per reason

JimStenstrom:
Refs #2283.

## Summary

The head-dim-256 prefill router (#2025/#2155/#2204 lineage) silently chooses between the fast unfused pass and the ~2×-slower O(L)-memory tiled pass. Nothing surfaces the decision, so a user on the tiled path at long context sees only the throughput loss — #2283 took a full source A/B to diagnose what one log line would have said. This adds that log line: the first tiled engagement per reason logs at INFO, repeats stay silent.

## Behavior

One INFO per distinct reason per process, naming the reason and the recovery lever: forced by `OMLX_SDPA256_TILED=1` · no headroom provider registered (guard disabled / scheduler gone — the #2283 case) · guard reports no active ceiling · unfused transient exceeds live headroom (with the MiB numbers and kv_len) · headroom probe failed. The probe-failure branch also gains a `logger.debug(..., exc_info=True)` for the traceback it previously swallowed. Routing behavior is unchanged — the diff only adds observability on the already-taken branches.

## Why safe

The dedup check is a set lookup that runs only on tiled-returning branches, and only the first occurrence per reason formats a message — the hot prefill path cost is one `in` check. No new env vars, no behavioral toggles, no change to any routing decision (the `transient > headroom` restructure is a pure if/return equivalence). Module state is one bounded set (max 5 entries).

## Tests

`tests/test_sdpa256_attention.py`: 4 new cases — logs once (not twice) when no provider is registered, logs the MiB/kv_len numbers on insufficient headroom, logs the forced-env reason, and stays silent on the unfused path. Existing provider-reset fixture extended to isolate the new logged-reasons set. Suite: 22 passed. RED-proof: the 3 emission tests fail on the parent commit (no records); the silence test is a preservation guard.

`PYTHONPATH=. python -m pytest tests/test_sdpa256_attention.py -q` → 22 passed.

## Out of scope (named in the issue thread, your calls)

The guard-off → always-tiled default and a reclaim-before-surrender attempt at the routing point are design decisions — this PR only makes whichever choice the router takes visible.

### DIFF
--- a/omlx/patches/sdpa256_attention.py
+++ b/omlx/patches/sdpa256_attention.py
@@ -69,6 +69,24 @@ _HEADROOM_PROVIDER: "weakref.WeakMethod | None" = None
 # OMLX_SDPA256_TILED override, parsed at apply time: True = always tiled,
 # False = never tiled, None = memory-aware auto.
 _FORCE_TILED: bool | None = None
+# Tiled-route reasons already logged. The tiled pass trades substantial
+# prefill throughput at long kv_len for O(L) memory, and nothing surfaced the
+# route decision before (issue #2283 took an A/B repro to diagnose), so the
+# first engagement per reason logs at INFO; repeats stay silent to keep the
+# hot path quiet.
+_TILED_ROUTE_LOGGED: "set[str]" = set()
+
+
+def _note_tiled_route(reason: str, detail: str) -> None:
+    if reason in _TILED_ROUTE_LOGGED:
+        return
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, this is exactly the breadcrumb #2283 was missing. I verified the routing decisions are unchanged in every branch and that decode-shaped calls still exit on the q_len gate before any of the new code runs. I'll merge this.

Issue #2330, opened 2026-07-22
0.5.2: two concurrent identical fully-cached prefills wedge the scheduler; phantom waiting request survives disconnect; teardown then impossible (strands wired)


anicaise-ai:
## Summary

On 0.5.2, **two concurrent identical, almost fully prefix-cached prefill requests wedged the scheduler permanently**. After the clients disconnected, one request stayed as a phantom `waiting_requests: 1` with `active_requests: 0`, and the server never scheduled anything again. SIGTERM was accepted but teardown could not run (wedged executor), so the exit stranded ~405 GB of wired memory — the "wedged server" arm of #2184, reboot required.

Not seeing anything in the 0.5.3 notes that covers this, hence the report.

## Environment

Mac Studio M3 Ultra 512GB, macOS Tahoe 26.5.2 (25F84), oMLX 0.5.2 (pip wheel), GLM-5.2 mxfp4 + native MTP, `--memory-guard-gb 470 --max-concurrent-requests 4 --hot-cache-max-size 128GB`, prefix + SSD cache enabled, `MLX_METAL_FAST_SYNCH=1` (noting it since #2183 was in the same territory; the fix for that is in this build).

## Timeline (server log + `/api/status`)

Uptime ~2h50, 63 requests served, all healthy. A keep-warm cron overlapped with a manual run of the same script (our bug — no lock), so **two identical requests** (same 8.3k-token system prompt, `max_tokens=1`, ~99% prefix-cached — the *previous* pair of the same prompts completed fine seconds earlier as batch of 2) arrived together, immediately followed by a third identical one:

```
11:00:43,630  Chat completion: 1 tokens in 7.20s, prompt: 7490     <- previous pair, fine
11:00:43,631  Chat completion: 1 tokens in 7.18s, prompt: 7490
              (the two 8,340-token twins are admitted right after — then NOTHING)
11:11:48,777  Request 1fb05a42… cancelled, aborting                <- we kill the clients 11 min later
11:11:48,777  Request 88f61133… cancelled, aborting
```

Between 11:00:43 and 11:11:48 the two twins sat in `active_requests: 2` with **zero further log lines** — for a ~150-token incremental prefill that normally takes 2–4 s. After the client kills, the cancels logged cleanly, but `/api/status` settled at **`active: 0, waiting: 1`** (the third twin, whose client was also dead) and stayed there:

- a fresh `max_tokens: 5` completion never returned (no admission, nothing logged),
[...]

PR #2355, opened 2026-07-23
feat(specprefill): MLA query extractor for DeepSeek/GLM-family drafts

StartupBros:
Fixes the SpecPrefill structural gap reported by @anicaise-ai in #2208: MLA-attention drafts (DeepSeek V2/V3 family, GLM-5.x, `glm4_moe_lite`) have no plain `q_proj`, so `_detect_query_extractor` fell through to `_llama_extract_queries` and crashed at scoring time (`'Glm4MoeLiteAttention' object has no attribute 'q_proj'`) after paying the full draft prefill, making an incompatible draft strictly worse than no draft (+42% total latency at 42k in the linked report). Since GLM-5.x targets have no non-MLA tokenizer-compatible small sibling (GLM-4.7-Flash is itself MLA), SpecPrefill was structurally unavailable for exactly the model family with the heaviest prefill cost.

**What this adds**

- `_mla_extract_queries`: builds queries in the layout the MLA cache stores. Absorbed variants (`embed_q`, e.g. `glm4_moe_lite`): `q_nope` projected into latent space, concatenated with the roped `q_pe` slice. Expanded variants (`kv_b_proj`, e.g. `deepseek_v2`): `concat(q_nope, roped q_pe)` directly. Queries are pre-scaled so the scorer's generic `1/sqrt(d)` reproduces the module's true attention scale, including yarn mscale corrections.
- `_is_mla_attention`, keyed on `kv_a_proj_with_mqa`, dispatching ahead of the existing chain: MLA modules have `rope`, so the llama fallback would otherwise claim them (both the low-rank-q and the full-q `q_lora_rank=None` variants).
- `score_tokens` derives `n_kv_heads` for MLA modules, which expose neither `num_key_value_heads` nor `n_kv_heads`: 1 for absorbed latent caches, all heads for expanded.
- `_compute_importance` rebuilds the full split-layout keys for absorbed caches (latent in `cache.keys`, roped positional slice in `cache.values`) via a dim-driven concat that non-MLA architectures never trigger.

**Validation**

- `pytest tests/test_specprefill.py tests/test_specprefill_mla.py -m "not slow"`: 47 passed (42 existing + 5 new), M3 Ultra, editable install at current main.
- The numerics tests validate against `Glm4MoeLiteAttention` itself: the extractor's scoring path (queries against `concat(cache.keys, cache.values)` at the generic scorer scale) is compared with scores computed from the module's own expanded-formulation math on the same random weights, for both `q_lora_rank` set and `None`. Agreement covers the layout, the RoPE offset handling, and the pre-scaling.
- End-to-end (GLM-4.7-Flash drafting for a GLM-5.2 mxfp4/oQ4e target, keep_pct sweeps) is not yet run: our 512GB box is mid benchmark campaign. @anicaise-ai offered in #2208 to benchmark a branch on exactly that pairing; this is that branch. We will post our own e2e numbers in #2208 once our box frees.

**Follow-up, deliberately not in this PR:** a draft-load-time arch compatibility check in `engine/batched.py` (`_load_draft`) would turn any remaining unsupported pairing into a clear load-time error instead of an expensive per-request fallback; #1073/#1074 proposed that shape earlier. Happy to revive it separately if wanted.

### DIFF
--- a/omlx/patches/specprefill.py
+++ b/omlx/patches/specprefill.py
@@ -162,6 +162,38 @@ def _nemotron_h_extract_queries(attn, x, cache=None, **kwargs):
     return queries
 
 
+def _mla_extract_queries(attn, x, cache=None, **kwargs):
+    """DeepSeek/GLM MLA: (optionally low-rank) q projection + partial RoPE.
+
+    MLA caches store keys in a split layout, so queries are built in the
+    matching layout. Absorbed variants (``embed_q``, e.g. glm4_moe_lite)
+    cache the compressed latent as keys and the roped positional slice as
+    values; queries are q_nope absorbed into latent space, concatenated
+    with the roped q_pe. Expanded variants (``kv_b_proj``) cache full
+    per-head keys; queries are concat(q_nope, roped q_pe) directly.
+    Queries are pre-scaled so the scorer's generic 1/sqrt(d) recovers the
+    module's true attention scale (including any yarn mscale correction).
+    """
+    B, L, D = x.shape
+    n_heads = getattr(
+        attn,
+        "num_heads",
+        getattr(attn, "num_attention_heads", getattr(attn, "n_heads", None)),
+    )
+    if getattr(attn, "q_proj", None) is not None:
+        q = attn.q_proj(x)
+    else:
+        q = attn.q_b_proj(attn.q_a_layernorm(attn.q_a_proj(x)))
+    q = q.reshape(B, L, n_heads, attn.q_head_dim).transpose(0, 2, 1, 3)
+    q_nope, q_pe = mx.split(q, [attn.qk_nope_head_dim], axis=-1)
+    offset = cache.offset if cache is not None else 0
+    q_pe = attn.rope(q_pe, offset)
+    if getattr(attn, "embed_q", None) is not None:
+        q_nope = attn.embed_q(q_nope)
+    queries = mx.concatenate([q_nope, q_pe], axis=-1)
+    return queries * (attn.scale * queries.shape[-1] ** 0.5)
+
+
 # ---------------------------------------------------------------------------
 # Model topology helpers
 # ---------------------------------------------------------------------------
@@ -300,9 +332,21 @@ def _is_gemma4_attention(attn_obj) -> bool:
     return "shared_kv" in params and "offset" in params
 
 
+def _is_mla_attention(attn_obj) -> bool:
+    """Detect DeepSeek/GLM-style Multi-head Latent Attention.
+
+    The shared latent KV projection (kv_a_proj_with_mqa) is the definitive
+    MLA marker; it is present in both the low-rank-q (q_a_proj/q_b_proj)
+    and full-q (plain q_proj) variants.
+    """
+    return hasattr(attn_obj, "kv_a_proj_with_mqa")
+
+
 def _detect_query_extractor(attn_obj) -> Callable:
     """Auto-detect the appropriate query extractor for the model architecture."""
-    if _uses_gated_q_proj(attn_obj):
+    if _is_mla_attention(attn_obj):
+        return _mla_extract_queries
+    elif _uses_gated_q_proj(attn_obj):
         return _qwen35_extract_queries
     elif _is_gemma4_attention(attn_obj):
         return _gemma4_extract_queries
@@ -420,6 +464,15 @@ def _compute_importance(
             continue
         head_dim = prompt_keys.shape[-1]
         q_stack = mx.concatenate(captures, axis=2)
+        if q_stack.shape[-1] > head_dim and getattr(cache, "values", None) is not None:
+            # Absorbed-MLA caches store the roped positional key slice in
+            # cache.values; concatenate it back so scoring sees the full
+            # split layout the queries were built in. Dim-driven, so
+            # non-MLA architectures never take this branch.
+            prompt_keys = mx.concatenate(
+                [prompt_keys, cache.values[..., :n_prompt, :]], axis=-1
+            )
+            head_dim = prompt_keys.shape[-1]
         if heads_per_group > 1:
             expanded_keys = mx.repeat(prompt_keys, heads_per_group, axis=1)
         else:
@@ -498,6 +551,12 @@ def score_tokens(
     n_kv_heads = getattr(
         attn_obj, "num_key_value_heads", getattr(attn_obj, "n_kv_heads", None)
     )
+    if n_kv_heads is None and _is_mla_attention(attn_obj):
+        # Absorbed-latent MLA caches store a single shared latent head;
+        # expanded variants store all heads.
+        n_kv_heads = (
+            1 if getattr(attn_obj, "embed_q", None) is not None else n_attn_heads
+        )
 
     if query_extractor is None:
         query_extractor = _detect_query_extractor(attn_obj)
--- a/omlx/patches/specprefill.py
+++ b/omlx/patches/specprefill.py
@@ -712,7 +712,12 @@ def manual_rope(x, positions, dims, base=10000.0, scale=1.0):
     rotated = mx.concatenate(
         [x1 * cos_a - x2 * sin_a, x1 * sin_a + x2 * cos_a], axis=-1
     )
-    return mx.concatenate([rotated, x_pass], axis=-1)
+    # Angles are computed in float32 for precision, but the output must
+    # keep the input dtype like native mx RoPE does: a float32 leak here
+    # becomes float32 roped K in the cache, and on mask-fused MLA targets
+    # (GLM DSA passes pe_scores as the sdpa mask) a float32 mask that
+    # cannot promote to the model's bfloat16 output.
+    return mx.concatenate([rotated, x_pass], axis=-1).astype(x.dtype)
 
 
 def manual_rope_with_freqs(x, positions, dims, freqs, pre_scale=1.0):
@@ -756,7 +761,12 @@ def manual_rope_with_freqs(x, positions, dims, freqs, pre_scale=1.0):
     rotated = mx.concatenate(
         [x1 * cos_a - x2 * sin_a, x1 * sin_a + x2 * cos_a], axis=-1
     )
-    return mx.concatenate([rotated, x_pass], axis=-1)
+    # Angles are computed in float32 for precision, but the output must
+    # keep the input dtype like native mx RoPE does: a float32 leak here
+    # becomes float32 roped K in the cache, and on mask-fused MLA targets
+    # (GLM DSA passes pe_scores as the sdpa mask) a float32 mask that
+    # cannot promote to the model's bfloat16 output.
+    return mx.concatenate([rotated, x_pass], axis=-1).astype(x.dtype)
 
 
 # ---------------------------------------------------------------------------
--- a/tests/test_specprefill_mla.py
+++ b/tests/test_specprefill_mla.py
@@ -1,0 +1,151 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for the MLA (Multi-head Latent Attention) SpecPrefill query extractor.
+
+Covers DeepSeek/GLM-family draft models (e.g. glm4_moe_lite / GLM-4.7-Flash),
+whose attention has no plain ``q_proj`` contract: queries go through an
+optional low-rank projection and only the rope slice of the head dim is
+rotated, and the KV cache stores the compressed latent (keys) plus the roped
+positional slice (values).
+"""
+
+from unittest.mock import MagicMock
+
+import mlx.core as mx
+import pytest
+
+from omlx.patches.specprefill import (
+    _detect_query_extractor,
+    _is_mla_attention,
+    _llama_extract_queries,
+    _mla_extract_queries,
+)
+
+
+class TestMlaDetection:
+    def test_detect_mla_low_rank_q(self):
+        attn = MagicMock(
+            spec=[
+                "kv_a_proj_with_mqa",
+                "q_a_proj",
+                "q_a_layernorm",
+                "q_b_proj",
+                "embed_q",
+                "rope",
+                "num_heads",
+                "q_head_dim",
+                "qk_nope_head_dim",
+                "scale",
+            ]
+        )
+        assert _is_mla_attention(attn)
+        assert _detect_query_extractor(attn) is _mla_extract_queries
+
+    def test_detect_mla_full_q_wins_over_llama_fallback(self):
+        # q_lora_rank=None MLA variants expose a plain q_proj; MLA detection
+        # must still win, since the cache layout is latent, not per-head.
+        attn = MagicMock(
+            spec=[
+                "kv_a_proj_with_mqa",
+                "q_proj",
+                "embed_q",
+                "rope",
+                "num_heads",
+                "q_head_dim",
+                "qk_nope_head_dim",
+                "scale",
+            ]
+        )
+        assert _detect_query_extractor(attn) is _mla_extract_queries
+
+    def test_non_mla_dispatch_unchanged(self):
+        attn = MagicMock(spec=["q_proj", "rope", "num_heads"])
+        assert not _is_mla_attention(attn)
+        assert _detect_query_extractor(attn) is _llama_extract_queries
+
+
+class TestMlaExtractorNumerics:
+    """Validate the extractor against the module's own math.
+
+    glm4_moe_lite computes attention in two equivalent formulations: absorbed
+    (decode: q_nope projected into latent space via embed_q) and expanded
+    (prefill: latent keys projected into q_nope space via embed_q with
+    transpose=False). The extractor uses the absorbed form against the cached
+    latent; the reference below uses the module's expanded form with the same
+    weights. Agreement validates layout, RoPE offset handling, and the
+    pre-scaling that maps the scorer's generic 1/sqrt(d) onto the module's
+    true attention scale.
+    """
+
+    def _tiny_args(self, glm, q_lora_rank):
+        return glm.ModelArgs(
+            hidden_size=64,
+            num_attention_heads=2,
+            num_key_value_heads=2,
+            q_lora_rank=q_lora_rank,
+            kv_lora_rank=16,
+            qk_rope_head_dim=8,
+            qk_nope_head_dim=16,
+            v_head_dim=16,
+            num_hidden_layers=1,
+            intermediate_size=128,
+            moe_intermediate_size=32,
+        )
+
+    @pytest.mark.parametrize("q_lora_rank", [32, None])
+    def test_scores_match_module_reference(self, q_lora_rank):
+        glm = pytest.importorskip("mlx_lm.models.glm4_moe_lite")
+        from mlx_lm.models.cache import KVCache
+
+        args = self._tiny_args(glm, q_lora_rank)
+        attn = glm.Glm4MoeLiteAttention(args)
+        cache = KVCache()
+        n_prompt = 5
+        x_prompt = mx.random.normal((1, n_prompt, args.hidden_size))
+        attn(x_prompt, mask=None, cache=cache)
+        assert cache.offset == n_prompt
+
+        x_step = mx.random.normal((1, 1, args.hidden_size))
+        queries = _mla_extract_queries(attn, x_step, cache=cache)
+        assert queries.shape == (
+            1,
+            attn.num_heads,
+            1,
+            args.kv_lora_rank + args.qk_rope_head_dim,
+        )
+
+        # Production scoring path: q_stack @ concat(keys, values)^T * 1/sqrt(d)
+        latent = cache.keys[..., :n_prompt, :]
+        k_pe = cache.values[..., :n_prompt, :]
+        keys_cat = mx.concatenate(
+            [
+                mx.broadcast_to(
+                    latent, (1, attn.num_heads, n_prompt, latent.shape[-1])
+                ),
+                mx.broadcast_to(k_pe, (1, attn.num_heads, n_prompt, k_pe.shape[-1])),
+            ],
+            axis=-1,
+        )
+        ext_scores = (queries @ keys_cat.swapaxes(-1, -2)) * (
+            keys_cat.shape[-1] ** -0.5
+        )
+
+        # Reference: the module's own expanded (prefill-branch) formulation.
+        if q_lora_rank is None:
+            q_full = attn.q_proj(x_step)
+        else:
+            q_full = attn.q_b_proj(attn.q_a_layernorm(attn.q_a_proj(x_step)))
+        q_full = q_full.reshape(1, 1, attn.num_heads, attn.q_head_dim).transpose(
+            0, 2, 1, 3
+        )
+        q_nope, q_pe = mx.split(q_full, [attn.qk_nope_head_dim], axis=-1)
+        q_pe = attn.rope(q_pe, cache.offset)
+        k_expanded = attn.embed_q(latent, transpose=False)
+        ref_scores = (
+            q_nope @ k_expanded.swapaxes(-1, -2)
+            + q_pe
+            @ mx.broadcast_to(
+                k_pe, (1, attn.num_heads, n_prompt, k_pe.shape[-1])
+            ).swapaxes(-1, -2)
+        ) * attn.scale
+
+        assert mx.allclose(ext_scores, ref_scores, atol=1e-4, rtol=1e-4)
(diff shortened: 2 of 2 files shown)


StartupBros:
End-to-end validation on the target hardware: M3 Ultra 512GB, oMLX 0.5.3 + this branch, GLM-4.7-Flash-4bit drafting for GLM-5.2 oQ4e-mtp (404GB weights, guard 490, `keep_pct 0.3`, `threshold 8192`), novel prompts, `max_tokens=1`, single stream:

| cold prompt | dense baseline (no draft) | with SpecPrefill | effective speedup |
|---|---|---|---|
| 15,952 tok | 87.9s (181.5 tok/s) | 42.6s (374 tok/s effective) | 2.1x |
| 42,952 tok | 431.6s (99.5 tok/s) | 167.6s (256 effective) | 2.6x |
| 42,952 tok (2nd fresh pass) | same baseline | 193.3s (222 effective) | 2.2x |

Scheduler logs are clean end to end: `scored 42952 tokens in 84.9s, selected 12872/42952 (keep=30%)` followed by `sparse prefill 12871/42952 conv tokens in 79.7s`, no errors, no fallback. An 8k control below the threshold correctly bypassed drafting (195.7 tok/s, matching dense).

The second commit (`fix(specprefill): preserve input dtype in manual RoPE helpers`) came out of this run. The first e2e attempt scored correctly but died at sparse prefill with `[scaled_dot_product_attention] Mask type must promote to output type bfloat16`, then paid the dense fallback on top of the scoring cost (the same expensive failure shape as the incompatible-draft case in #2208). Root cause: `manual_rope` / `manual_rope_with_freqs` compute angles in float32 (correctly) but leaked float32 outputs; under `sparse_prefill` the position-mapped RoPE therefore produced float32 `q_pe`/`k_pe` on the target, and GLM DSA passes the resulting float32 `pe_scores` as the fused sdpa mask against bfloat16 tensors. Casting the output back to `x.dtype` matches native `mx` RoPE semantics; as a side effect `sparse_prefill` no longer writes silently-upcast float32 roped keys into the KV cache for any architecture.

Scope note: `keep_pct 0.3` pruning semantics are the SpecPrefill paper's tradeoff, unchanged here; these numbers validate the plumbing (detection, scoring, sparse prefill, decode positioning), not GLM-specific pruning quality.

anicaise-ai:
Datapoint on our quant, as requested — the mxfp4 pair you didn't have.

**Box:** M3 Ultra 512GB. **Target:** GLM-5.2-mxfp4 (~420GB resident, DSA attention). **Drafter:** GLM-4.7-Flash-4bit. Local source build carrying the #2330 fence fix. Threshold 8192; keeps 0.3 / 0.5 / 0.6. Verification = needle-in-haystack on a ~20k-token context (secret code embedded mid-haystack, `temperature=0`, dense reference returns it exactly).

**Result: SpecPrefill is incompatible with the mxfp4/DSA pair. All three keeps corrupt generation.** This is not a tuning problem — it is structural, and the two failure axes point in opposite directions.

| keep | prefill speedup | needle | output |
|------|-----------------|--------|--------|
| 0.3  | 2.4–2.7×        | ❌ FAIL | degenerate — "Je ne peux pas analyser…" / repeated tokens |
| 0.5  | 1.55–1.80×      | ❌ FAIL | "Je comprends que vous…" / `A </1. **` |
| 0.6  | 1.38–1.42×      | ❌ FAIL | "Je n'ai pas pu analyser…" / `Hello! My 4.0 111…` |
| dense| — (ref)         | ✅ OK   | correct code, exact |

The token-repetition signature is consistent with a corrupted KV going into decode — the sparse prefill produces a KV cache the DSA decode path can't generate coherently from. And the trade is unwinnable: the speedup **shrinks as keep rises** (2.7× → 1.4×), so the only direction that might recover quality is exactly the direction that erases the gain. There is no keep that is both fast and correct on this pair.

Contrast with your clean E2E on oQ4e-mtp (16k 87.9→42.6s, 43k 431.6→167.6s, scoring/sparse-prefill all sane): the extractor and the sparse path themselves are clearly working — the incompatibility is specific to **mxfp4 + DSA**, not to SpecPrefill in general. Our best guess is the interaction between the mxfp4 dequant path and the sparse KV the extractor selects on DSA layers, but we haven't instrumented the KV yet.

Thanks for building the MLA query extractor (`_mla_extract_queries`, both layouts) — credit noted. Happy to run an instrumented build (dump the sparse-vs-dense KV divergence on a DSA layer) if that would help localize it, since this is the one pair that fails.

(Separately, this surfaced that `specprefill_enabled` defaulting to true silently corrupted any >8192-token request on the mxfp4 build until we caught it — flagging in case the default is worth revisiting while the mxfp4/DSA path is unresolved.)

jundot:
Thanks for working on MLA support. I found one blocker in the GLM DSA path: SpecPrefill adjusts the main attention positions, but not the DSA indexer positions, which can corrupt generation after sparse prefill. GLM's `CacheList` also needs explicit handling for system prompts and draft scoring. Could you fix both paths and add a multi-token generation test against dense output? I would hold off on merging until both mxfp4 and oQ4e pass that test.

anicaise-ai:
@jundot — your note above asks for exactly the two fixes we already carry, so here is the datapoint that bears directly on your merge gate: **we implemented both, and recall is still 0/21.**

We independently found and fixed the same two defects — the DSA indexer left on unrotated positions, and `CacheList` silently reporting offset 0 for system prompts and draft scoring — then ran a needle grid against a build carrying them *and* against this PR as-is. **Both fail identically.** Fixing both paths is necessary, they are real defects, but on our mxfp4/DSA pair it will not make a multi-token-generation-against-dense test pass. There is a third failure in the DSA sparse-prefill path, and the last row of the control table below localises it to *positions*, downstream of selection.

Our patches are written up in §"What we already fixed locally" — happy to open them as a PR against this branch if useful, though on this evidence they should not be treated as clearing the gate.

Full report follows.

---

# SpecPrefill on GLM-5.2 (DSA): sparse prefill is fast but returns no recall

**Summary.** On GLM-5.2-mxfp4, enabling SpecPrefill takes needle-in-a-haystack recall from 9/9 to
**0/21** across keep ratios 0.2 / 0.3 / 0.5, depths 0.1 / 0.5 / 0.9 and lengths 16k / 64k, on both
the PR as-is and a build carrying three DSA-specific fixes of ours. The failure is not a missed
needle: the model reports that it cannot read the document at all. Separately, the scoring pass is a
dense forward over the whole prompt, so the end-to-end speedup falls from ~3× at 16k to 1.6× at 64k.
One configuration (keep 0.3 at 64k) wedged the server outright.

## Environment

| | |
|---|---|
| Host | Mac Studio, Apple M3 Ultra, 512 GiB unified memory |
| OS | macOS 26.5.2 |
| oMLX | 0.5.3 (local `dsafix` wheel build) |
| Target model | GLM-5.2-mxfp4 — 389.64 GB resident, `max_model_len` 1 048 576 |
| Draft model | GLM-4.7-Flash-4bit |
| Server | `omlx serve --max-concurrent-requests 4 --memory-guard-gb 470 --hot-cache-max-size 128GB`, paged SSD cache enabled |
| Active patches | `glm_moe_dsa` native kernels, `mlx_lm_mtp.nemotron_h_chain`, `sdpa256_attention` |

## Method

Needle-in-a-haystack. A single sentence carrying the answer token (`ZEPHYR-4417`) is inserted at
relative depth 0.1 / 0.5 / 0.9 into French filler prose, then the question is appended. Every request
carries a **unique preamble in first position** so the paged prefix cache cannot serve it — the
reported timings are real prefill, not cache hits (`cached_tokens` is 0 on every row below).
`temperature: 0`, `max_tokens: 32`.

Configurations: dense (`specprefill: false`) as the anchor, then `specprefill: true` at
`specprefill_keep_pct` 0.2 / 0.3 / 0.5.

## Result 1 — sparse prefill trades all recall for throughput

16 000-token haystack, with the DSA fixes described in §"What we already fixed locally" applied:

| config | depth | prompt tok | wall | prefill tok/s | recall | reply |
|---|---|---|---|---|---|---|
| dense | 0.1 | 16 725 | 101.2 s | 165.2 | **OK** | `ZEPHYR-4417` |
| dense | 0.5 | 16 725 | 98.5 s | 169.8 | **OK** | `ZEPHYR-4417` |
| dense | 0.9 | 16 725 | 145.0 s | 115.3 | **OK** | `ZEPHYR-4417` |
| keep 0.2 | 0.1 | 16 728 | 39.2 s | 426.7 | fail | "Je comprends que vous travaillez sur un document technique…" |
| keep 0.2 | 0.5 | 16 728 | 38.2 s | 438.3 | fail | "Je comprends que vous rencontrez des difficultés avec un document PDF…" |
| keep 0.2 | 0.9 | 16 728 | 44.1 s | 379.7 | fail | "Je suis désolé, mais je ne suis pas en mesure de vous aider…" |
| keep 0.3 | 0.1 | 16 728 | 51.3 s | 326.0 | fail | "Le texte fourni semble être une série d…" |
| keep 0.3 | 0.5 | 16 728 | 51.6 s | 323.9 | fail | "Je comprends que vous rencontrez des difficultés avec l'analyse…" |
| keep 0.3 | 0.9 | 16 728 | 51.3 s | 325.9 | fail | "Je comprends que vous souhaitez une analyse d'un document PDF…" |
| keep 0.5 | 0.1 | 16 728 | 69.6 s | 240.5 | fail | "Je suis désolé, mais je ne peux pas répondre à cette question…" |
| keep 0.5 | 0.5 | 16 728 | 69.1 s | 242.2 | fail | "Je suis désolé, mais je ne peux pas répondre à cette question…" |
| keep 0.5 | 0.9 | 16 728 | 69.6 s | 240.3 | fail | "elle semble être une série de carac…" |

Dense: **6/6**. Sparse: **0/12**, at every keep ratio and every depth, for a 1.5×–2.6× prefill speedup.

Note the failure mode: at 16k the model does not merely miss the needle, it does not see a document
at all. It answers as if handed an unreadable blob ("a series of characters", "an unreadable PDF").
That is the signature of corrupted positions, not of an over-aggressive token budget — an
over-aggressive budget would drop the needle and still summarise the filler correctly.

Two further details point the same way. The reply is invariant to needle depth: at keep 0.2 the
model returns the same sentence for depth 0.1 and depth 0.9, which is what you get when the document
never reaches the model at all rather than when a ranked selection misses one line. And on the
PR-only control several sparse replies come back **in English** although the prompt — filler, needle
and question — is entirely in French. Losing the language of a 16 000-token prompt is not a recall
miss; the surviving context is not being read as text.

## Result 2 — the speedup erodes with length, because scoring is itself dense

The server log separates the two phases, and the scoring pass is a full forward over the *entire*
prompt on the draft model:

| prompt | scoring | sparse prefill | scoring throughput | scoring share of compute |
|---|---|---|---|---|
| 16 728 tok | 17.3 s | 17.0 s (keep 0.2) | 967 tok/s | 50 % |
| 67 017 tok | 238.0 s | 88.2 s (keep 0.2) | 282 tok/s | **73 %** |

Prompt ×4.0, scoring time ×13.8 — the scoring pass scales at roughly n^1.9, i.e. full attention.
End to end the advantage over dense falls from ~3× at 16k to **1.6×** at 64k. SpecPrefill is
asymptotically dominated by exactly the cost it exists to avoid, and the erosion happens precisely
in the regime where sparse prefill is supposed to pay off. This is worth addressing independently of
the recall bug: even with perfect recall, the return shrinks as context grows.

## Result 3 — it degrades further with length, and 0.3 wedges the server

64 000-token haystack:

| config | depth | prompt tok | wall | prefill tok/s | recall | reply |
|---|---|---|---|---|---|---|
| dense | 0.1 | 67 014 | 534.1 s | 125.5 | **OK** | `ZEPHYR-4417` |
| dense | 0.5 | 67 014 | 550.9 s | 121.6 | **OK** | `ZEPHYR-4417` |
| dense | 0.9 | 67 014 | 595.4 s | 112.6 | **OK** | `ZEPHYR-4417` |
| keep 0.2 | 0.1 | 67 017 | 349.9 s | 191.5 | fail | `'2'2'2'2'2'2'2'2'2'2'2'2'2'2'2'2` |
| keep 0.2 | 0.5 | 67 017 | 339.6 s | 197.3 | fail | `Par0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0` |
| keep 0.2 | 0.9 | 67 017 | 333.7 s | 200.8 | fail | `1.0 0 (0.0 0.0 0.0 0.0 0.0 0.0 0` |

At 64k the output is no longer fluent-but-wrong, it is **degenerate token repetition**. Same needle,
same code path, longer context: the further the sparse prefill runs, the more the decode falls apart.

**`keep 0.3` at 64k never returned a row** — it wedged the server: the draft pass pushed occupancy
past the prefill ceiling and the request never completed. We kept the `sample` of the wedged process and can post it. Recovering from that wedge cost a hard kill of a process mid-Metal-allocation, which
orphaned ~399 GB of wired memory that no live process held; only a reboot returned it. Worth
guarding at the admission layer — a keep ratio that raises draft occupancy should be refused, not
allowed to hang the server.

Note that the tables above are **not** unpatched upstream behaviour: they were produced by a build
already carrying three DSA fixes of ours (detailed below). Recall is 0/12 *after* those fixes.

## Control: the PR as-is behaves identically

To separate "our three fixes are irrelevant" from "our three fixes help but are not enough", the
same 16k grid was re-run against the **unmodified PR** (server restarted, model reloaded, dense
anchor re-measured):

| config | depth | wall | prefill tok/s | recall | reply |
|---|---|---|---|---|---|
| dense | 0.1 | 113.8 s | 146.9 | **OK** | `ZEPHYR-4417` |
| dense | 0.5 | 107.3 s | 155.9 | **OK** | `ZEPHYR-4417` |
| dense | 0.9 | 143.5 s | 116.5 | **OK** | `ZEPHYR-4417` |
| keep 0.2 | 0.1 | 44.7 s | 374.1 | fail | "Je comprends que vous cherchez à analyser un document PDF…" |
| keep 0.2 | 0.5 | 39.5 s | 423.9 | fail | "Je comprends que vous rencontrez des difficultés avec un document PDF…" |
| keep 0.2 | 0.9 | 37.7 s | 443.5 | fail | "Je comprends que vous cherchez à analyser un dossier PDF…" |
| keep 0.3 | 0.1 | 47.5 s | 352.1 | fail | "Les messages que vous avez fournis sem…" |
| keep 0.3 | 0.5 | 47.1 s | 354.9 | fail | "I apologize, but I'm not able to understand the text you've provided." |
| keep 0.3 | 0.9 | 47.4 s | 352.7 | fail | "I apologize, but I'm having difficulty understanding your message." |
| keep 0.5 | 0.1 | 69.7 s | 240.0 | fail | "I cannot provide a response to the given prompt as it does not…" |
| keep 0.5 | 0.5 | 69.0 s | 242.4 | fail | "I apologize, but I'm unable to understand the text you've provided." |
| keep 0.5 | 0.9 | 74.4 s | 225.0 | fail | `Par d'analyse d'178d5430.5keep-160.9. La chaine d'archives. Par d'analyse d'` |

Dense 3/3, sparse 0/9 — the same verdict, the same failure text, and throughput within noise of the
patched build (keep 0.5: 240.0 / 242.4 / 225.0 against 240.5 / 242.2 / 240.3). The dense anchor
reproduces the earlier sweep to within 1 % at every depth, so the two columns are comparable
row by row.

**The last row is the most informative single result in this report.** The unique preamble of that
request was `Dossier d'analyse 178d5430…-keep0.5-16000-0.9.` and one filler line reads *"La chaine
d'ingestion documentaire traite les formats bureautiques et les archives PDF."* The model emits
`Par d'analyse d'178d5430.5keep-160.9. La chaine d'archives. Par d'analyse d'` — it is stitching
together real fragments of its own context, with the connective tissue gone and the fragments'
internal order scrambled (`keep0.5-16000-0.9` → `.5keep-160.9`). The surviving tokens are the right
tokens. What is lost is their order. That is a positional failure downstream of selection, not a
selection failure.

## Reading

The three DSA fixes below are necessary — a `CacheList` silently reporting offset 0, and an indexer
left on unrotated positions, are real defects — but they are **not** what breaks recall: the PR
without them fails identically. The remaining bug is elsewhere in the DSA sparse-prefill path, and
the fragment-stitching above says it is positional rather than in the token scoring.

## What we already fixed locally (necessary, not sufficient)

Three DSA-specific defects found and fixed in `omlx/patches/specprefill.py`. Offered as-is if you
want them — they are independent of the recall bug above.

1. **`cache[i].offset` reads 0 on DSA layers.** GLM-5.2 (and DeepSeek V3.2/V4) wrap several caches in
   a `CacheList` — latent MLA at index 0, indexer at index 1 — and `CacheList` deliberately exposes no
   `offset`. A bare `hasattr(entry, "offset")` therefore reports False and silently yields 0. Position
   mapping, draft scoring and lookahead trimming all then work from a wrong base as soon as a system
   prompt's KV has been restored. Fixed with a `_cache_offset()` that descends into `CacheList`.

2. **A DSA layer rotates twice, only one rotation was wrapped.** The attention owns a `rope`, and so
   does the `indexer` that selects the sparse top-k — separate instances, separate caches. Wrapping
   only `attn.rope` in `_PositionMappedRoPE` leaves the indexer scoring mis-rotated keys, so the
   top-k selection is wrong on *every* DSA layer. Fixed by enumerating RoPE owners
   (`_rope_owners()`) and giving each its own `cache_start` from its own cache slot.

3. **The restore path had the same blind spot** — the indexer kept decoding at compacted positions
   after the attention had been moved back to true ones. `cleanup_rope` likewise only unwrapped
   `attn.rope`.

## Ask

1. Treat DSA sparse prefill as broken for recall, not merely lossy — the fragment-stitching in the
   last control row localises it to positions, downstream of selection.
2. Refuse a keep ratio whose draft pass would exceed the prefill ceiling, at admission. Hanging the
   server is a worse outcome than declining the request; recovering from it cost a reboot here.
3. Consider the scoring pass's own cost in the value proposition — at 64k it is 73 % of the work.

`specprefill_enabled` is being turned back off on this host. Happy to run any instrumented build
against the same grid: the harness is ~120 lines, the needle grid is deterministic (`temperature: 0`,
unique preamble to defeat the prefix cache), and I can post both CSVs or the harness itself here on
request.

jundot:
Thanks, this isolates the remaining issue. I reproduced another positional mismatch in `manual_rope`: GLM DSA creates its attention RoPE with `traditional=True`, but `_PositionMappedRoPE` always uses the half-stride layout. On the same tensor, the sparse helper differs from native GLM RoPE by 3.46 max absolute error; adding the traditional consecutive-pair layout makes it match within 1.2e-7. The `CacheList` and indexer fixes are still needed, but this looks like the remaining recall blocker. @StartupBros, could you update this PR to preserve the traditional RoPE layout along with those fixes? @anicaise-ai, once that branch is ready, could you rerun the same grid?

anicaise-ai:
@jundot — your `traditional=True` diagnosis is right, and on our mxfp4/DSA pair it is sufficient: **recall goes from 0/9 to 9/9.**

Rather than wait for the branch, we implemented the consecutive-pair layout locally on top of the three DSA fixes and re-ran the same 16k grid. Same harness, same needle, same method as the earlier report (unique preamble in first position so the paged prefix cache cannot serve the request — `cached_tokens` is 0 on every row).

## Result — 12/12, and the speedup is real

| config | depth | wall | prefill tok/s | recall |
|---|---|---|---|---|
| dense | 0.1 / 0.5 / 0.9 | 100.5 / 97.8 / 143.5 s | 166 / 171 / 117 | **3/3** |
| keep 0.2 | 0.1 / 0.5 / 0.9 | 37.8 / 41.7 / 40.5 s | **443 / 401 / 413** | **3/3** |
| keep 0.3 | 0.1 / 0.5 / 0.9 | 49.3 / 50.9 / 47.2 s | 339 / 329 / 355 | **3/3** |
| keep 0.5 | 0.1 / 0.5 / 0.9 | 80.5 / 71.2 / 71.3 s | 208 / 235 / 235 | **3/3** |

**keep 0.2 is 2.85× faster than dense end to end and still finds the needle at every depth.** Every single one of those nine cells was a failure yesterday — fluent-but-blind replies, then degenerate repetition at 64k — on both this PR as-is and our patched build. The dense anchor reproduces the previous two campaigns to within a tenth of a second at depth 0.9, so the columns are comparable row by row.

## Independent confirmation of the numbers

Against `nn.RoPE(traditional=True)` on the same tensor: **4.286** max abs error with the half-stride layout, **8.3e-07** with consecutive pairs. Same orders of magnitude as your 3.46 / 1.2e-7 on a different tensor.

And the premise checks out in the source rather than by inference: `mlx_lm/models/deepseek_v3.py:113` builds the attention RoPE with `traditional=True`, and GLM-5.2 DSA is in that family.

## The patch we ran

One helper plus a flag read from the rope itself, wired into both rope paths. `_OffsetAdjustedRoPE` needs nothing — it delegates to the model's genuine rope, which already uses the right layout.

```python
def _rotate_pairs(x_rot, cos_a, sin_a, half, traditional):
    if traditional:
        x1 = x_rot[..., 0::2]
        x2 = x_rot[..., 1::2]
        r1 = x1 * cos_a - x2 * sin_a
        r2 = x1 * sin_a + x2 * cos_a
        return mx.stack([r1, r2], axis=-1).reshape(x_rot.shape)
    x1, x2 = x_rot[..., :half], x_rot[..., half:]
    return mx.concatenate([x1 * cos_a - x2 * sin_a, x1 * sin_a + x2 * cos_a], axis=-1)
```

`_PositionMappedRoPE.__init__` then takes `self._traditional = bool(getattr(original_rope, "traditional", False))` and passes it to `manual_rope` **and** `manual_rope_with_freqs`. Full 89-line diff available on request, or @StartupBros can take it as-is.

## Two caveats worth carrying into the merge gate

**64k is not re-tested.** This grid is 16k only. The `keep 0.3` @ 64k configuration that wedged the server earlier is a *memory* problem — draft occupancy past the prefill ceiling — and the RoPE fix does not touch it. It should still be refused at admission rather than allowed to hang the server; recovering from that wedge cost us a reboot.

**The scoring pass is still a dense forward over the whole prompt.** From the earlier report: 17.3 s at 16k against 238 s at 67k, i.e. ~n^1.9, so it grows from 50 % to 73 % of the work and the end-to-end advantage erodes from ~3× to ~1.6× as context grows. Correctness is fixed; that cost curve is a separate question.

Happy to re-run this grid against the updated branch once it lands, and to add the 64k rows if you want them with an admission guard in place.

anicaise-ai:
Follow-up: **the fix holds at 64k too — 3/3, and it is where the degenerate output used to be.**

Same build (three DSA fixes + consecutive-pair layout), `keep 0.2` only, 64k haystack:

| depth | wall | prefill tok/s | recall | same cell yesterday |
|---|---|---|---|---|
| 0.1 | 289.2 s | 231.7 | **OK** | `'2'2'2'2'2'2'2'2…` |
| 0.5 | 291.2 s | 230.1 | **OK** | `Par0.0.0.0.0.0.0…` |
| 0.9 | 292.8 s | 228.9 | **OK** | `1.0 0 (0.0 0.0 0.0…` |

Against the dense anchor measured on the same box (534.1 / 550.9 / 595.4 s, 3/3), that is **1.85× – 2.03× end to end**. Wall times land within 3.6 s of each other across depths, so this is stable behaviour rather than a lucky draw.

Two deliberate choices, so the numbers are not over-claimed:

- **`keep 0.3` @ 64k was not re-run.** That is the configuration that wedged the server; its cause is draft occupancy past the prefill ceiling, which the RoPE fix does not touch. It still needs refusing at admission.
- **The dense anchor was not re-measured.** The dense path never goes through `_PositionMappedRoPE`, so the layout change cannot affect it — reusing the earlier anchor rather than spending another 27 minutes of compute.

## The remaining design question is the scoring pass, not correctness

From the server log, the split for these runs:

| | scoring | sparse prefill | scoring share |
|---|---|---|---|
| 16k `keep 0.2` | 17.3 s | 17.0 s | 50 % |
| 64k `keep 0.2` | **197.3 s** | 85.2 s | **70 %** |

Prompt ×4.0, scoring ×11.4 — roughly n^1.75. The sparse prefill itself is only 85 s of the 289 s; the selection pass dominates, and it is a dense forward over the whole prompt. So the end-to-end advantage goes from 2.85× at 16k to ~1.9× at 64k and will keep eroding as context grows.

That is a design observation rather than a bug, but it bounds what SpecPrefill can be worth in the long-context regime it targets. Correctness is fixed; this curve is untouched.

(One honest note: scoring came in at 197.3 s here against 238.0 s in the earlier campaign. The scoring path does not use `_PositionMappedRoPE`, so I would not credit the fix for that — most likely variance or a less loaded box.)

Full CSVs and the 89-line diff available on request. Happy to re-run either grid against the updated branch when it lands.

PR #2361, opened 2026-07-24
fix(scheduler): route chunked-prefill'd requests to external VLM MTP (#2219)

### LINKED ISSUE #2219: External VLM MTP is silently bypassed after scheduler chunked prefill
[#](url) External VLM MTP is silently bypassed after scheduler chunked prefill

## Describe the bug

When external VLM MTP is enabled, text prompts that enter oMLX's scheduler-level chunked-prefill path are decoded through the ordinary `BatchGenerator` instead of external VLM MTP.

Shorter prompts use VLM MTP correctly. Longer prompts silently fall back to normal decoding, even though:

- `vlm_mtp_enabled=true`
- The external drafter loads successfully
- The benchmark is marked as using the experimental `vlm_mtp` feature
- No warning or fallback message is emitted

Disabling scheduler-level chunked prefill makes the same long requests route through VLM MTP correctly.

## To reproduce

1. Configure the target model with external VLM MTP:

   ```json
   {
     "mtp_enabled": false,
     "vlm_mtp_enabled": true,
     "vlm_mtp_draft_model": "/path/to/models/mlx-community/Qwen3.6-35B-A3B-MTP-bf16",
     "vlm_mtp_draft_block_size": 3
   }
   ```

2. Use the following model and feature configuration:

   - Target model: `mlx-community/Qwen3.6-35B-A3B-bf16`
   - External drafter: `mlx-community/Qwen3.6-35B-A3B-MTP-bf16`
   - TurboQuant KV: disabled
   - SpecPrefill: disabled
   - DFlash: disabled

3. Enable scheduler-level chunked prefill:

   ```json
   {
     "scheduler": {
       "chunked_prefill": true
     }
   }
   ```

4. Run single-request benchmarks with the following custom configuration:

   ```json
   {
     "model_id": "Qwen3.6-35B-A3B-bf16",
     "prompt_lengths": [1024, 32768, 65536],
     "generation_length": 1024,
     "batch_sizes": [],
     "force_lm_engine": false
   }
   ```

   The generation length is intentionally set to 1024 tokens rather than the default 128 tokens so that routing behavior, sustained decode throughput, and MTP acceptance statistics are clearly observable.

   See at the bottom of this issue for a bash script to run the benchmark with 1024 instead of 128 tokens.

5. Observe that the 1K request uses VLM MTP:

   ```text
   vlm_mtp decode started: ... block_size=3
   vlm_mtp stats: ... emitted=1024 block_size=3
   ```

6. Observe that the 32K and 64K requests do not emit either of these log messages:

   ```text
   vlm_mtp decode started
   vlm_mtp stats
   ```

   They complete through ordinary decoding:

   ```text
   pp32768/tg1024:
   processing_tps=3099.4
   gen_tps=63.1
   ttft_ms=10572.3
   e2e_latency_s=26.806

   pp65536/tg1024:
   processing_tps=2198.6
   gen_tps=54.8
   ttft_ms=29807.4
   e2e_latency_s=48.500
   ```

7. Disable scheduler-level chunked prefill:

   ```json
   {
     "scheduler": {
       "chunked_prefill": false
     }
   }
   ```

8. Run the same benchmark again.

9. Observe that all requests now route through external VLM MTP.

   For 32K:

   ```text
   vlm_mtp decode started: ... block_size=3
   vlm_mtp stats:
   finish=length
   rounds=352
   accepted=671/704 (95.3%)
   tokens_per_round=2.91
   emitted=1024
   block_size=3
   ```

   ```text
   processing_tps=3412.8
   gen_tps=87.6
   ttft_ms=9601.4
   e2e_latency_s=21.309
   ```

   For 64K:

   ```text
   vlm_mtp decode started: ... block_size=3
   vlm_mtp stats:
   finish=length
   rounds=353
   accepted=671/706 (95.0%)
   tokens_per_round=2.90
   emitted=1024
   block_size=3
   ```

   ```text
   processing_tps=2482.3
   gen_tps=74.1
   ttft_ms=26400.9
   e2e_latency_s=40.227
   ```

The measured difference in this custom `tg1024` benchmark is:

| Test | Chunked prefill on | Chunked prefill off | Improvement |
|---|---:|---:|---:|
| 32K decode | 63.1 tok/s | 87.6 tok/s | +38.8% |
| 64K decode | 54.8 tok/s | 74.1 tok/s | +35.2% |
| 32K E2E | 26.806 s | 21.309 s | -20.5% |
| 64K E2E | 48.500 s | 40.227 s | -17.1% |

The primary reproduction signal is not the performance difference, but the absence or presence of the `vlm_mtp decode started` and `vlm_mtp stats` log entries for otherwise identical requests.

## Expected behavior

When external VLM MTP is enabled and a request is otherwise eligible, completion of scheduler-level chunked prefill should route the request through `_route_to_vlm_mtp()`.

If external VLM MTP and scheduler chunked prefill are intentionally incompatible, oMLX should instead:

1. Reject or automatically disable the combination.
2. Emit a clear warning when MTP is bypassed.
3. Expose in benchmark results whether MTP was actually used for each measured request.

Silent fallback makes benchmark results and production behavior misleading.

## Screenshots

Not applicable. The issue is visible in benchmark output and server logs.

## Environment

- macOS: 26.5.2
- oMLX: 0.5.0
- Hardware: Apple silicon Mac, 128 GB unified memory

### Apparent cause

Long prompts enter the scheduler-level chunked-prefill branch.

When chunked prefill completes, `_advance_chunked_prefills()` calls:

```python
self._insert_prefilled_request(request, state, scheduled)
```

`_insert_prefilled_request()` finalizes the cache and inserts the request directly into `BatchGenerator`:

```python
uids = self.batch_generator.insert(
    [state.last_token],
    max_tokens=[request.sampling_params.max_tokens],
    caches=[state.cache] if state.cache else None,
    ...
)
```

It does not attempt `_route_to_vlm_mtp()`.

The normal non-chunked external-prefill path does attempt:

```python
vlm_mtp_uid = self._route_to_vlm_mtp(
    request,
    cache_to_use,
    tokens_to_process,
    sampler,
    sm,
)
```

before falling back to `BatchGenerator`.

This appears to make the MTP routing decision available only in the normal external-prefill completion path, not in the scheduler chunked-prefill completion path.

### Workaround

Set:

```json
{
  "scheduler": {
    "chunked_prefill": false
  }
}
```

With this setting, the normal external-prefill path is used and external VLM MTP activates for 32K and 64K requests.

This workaround changes scheduler fairness because one long prefill is no longer spread across multiple scheduler steps.

### Possible fix

A minimal fix would be to attempt VLM MTP routing inside `_insert_prefilled_request()` before inserting into `BatchGenerator`.

Conceptually:

```python
def _insert_prefilled_request(
    self,
    request,
    state,
    scheduled,
):
    self._finalize_chunked_prefill_cache_for_insert(request, state.cache)

    if request.sampling_params.seed is not None:
        mx.random.seed(request.sampling_params.seed)

    if self._vlm_mtp_drafter is not None and state.cache is not None:
        vlm_mtp_uid = self._route_to_vlm_mtp(
            request=request,
            prefilled_cache=state.cache,
            last_tokens=state.last_token,
            sampler=state.sampler,
            state_machine=state.sm,
        )
        if vlm_mtp_uid is not None:
            self._register_running_request(
                request=request,
                uid=vlm_mtp_uid,
                scheduled=scheduled,
            )
            return

    self._ensure_batch_generator(request.sampling_params)

    # Existing BatchGenerator fallback.
    ...
```

The current `_ensure_batch_generator()` call in `_advance_chunked_prefills()` could then move into the fallback branch. An external-MTP-routed request does not need a `BatchGenerator` to be created merely to complete admission.

A cleaner implementation would extract a shared post-prefill routing helper, for example:

```python
_start_prefilled_decode(
    request,
    prompt_cache,
    last_tokens,
    sampler,
    state_machine,
    logits_processors,
    scheduled,
)
```

That helper would:

1. Attempt external VLM MTP routing.
2. Register the request and synthetic MTP UID when routing succeeds.
3. Otherwise create or reuse `BatchGenerator`.
4. Insert the request through the standard decoding path.
5. Perform common request, UID, and running-state bookkeeping.

Both the normal external-prefill path and scheduler chunked-prefill path could then call the same helper.

### Suggested regression tests

Add a scheduler test where:

- `chunked_prefill=true`
- The prompt length is greater than `prefill_step_size`
- An external VLM MTP drafter is attached
- `_route_to_vlm_mtp()` returns a valid synthetic UID

Assert that:

1. `_route_to_vlm_mtp()` is called after the final prefill chunk.
2. `BatchGenerator.insert()` is not called when MTP routing succeeds.
3. The request is entered into `running` with the MTP UID.
4. Ordinary `BatchGenerator` insertion still occurs when MTP routing returns `None`.
5. Inline chunked-prefill completion and multi-step chunked-prefill completion behave identically.
6. The existing single-active-drafter concurrency fallback remains unchanged.

## Script used to run benchmark with tg1024

```bash
OMLX_URL="http://127.0.0.1:8888"
COOKIE_JAR="$(mktemp)"

read -rsp "oMLX admin API key: " OMLX_API_KEY
echo

# Admin endpoints require a session cookie.
curl -fsS \
  -c "$COOKIE_JAR" \
  -d "$(jq -n \
    --arg key "$OMLX_API_KEY" \
    '{api_key: $key, remember: false}')" \
  "$OMLX_URL/admin/api/login" >/dev/null

# Start pp32K + pp64K, each with tg1024.
BENCH_ID="$(
  curl -fsS \
    -b "$COOKIE_JAR" \
    -d '{
      "model_id": "Qwen3.6-35B-A3B-bf16",
      "prompt_lengths": [1024, 32768, 65536],
      "generation_length": 1024,
      "batch_sizes": [],
      "force_lm_engine": false
    }' \
    "$OMLX_URL/admin/api/bench/start" |
  jq -r '.bench_id'
)"

echo "Benchmark: $BENCH_ID"

# Follow progress and results as SSE.
curl -fsSN \
  -b "$COOKIE_JAR" \
  "$OMLX_URL/admin/api/bench/$BENCH_ID/stream"

# Print final structured results.
curl -fsS \
  -b "$COOKIE_JAR" \
  "$OMLX_URL/admin/api/bench/$BENCH_ID/results" |
jq '.results'

rm -f "$COOKIE_JAR"
unset OMLX_API_KEY
```

ethannortharc:
## Summary

Fixes #2219.

When external VLM MTP is enabled, text prompts long enough to enter the scheduler's chunked-prefill path are silently decoded through the ordinary `BatchGenerator` instead of VLM MTP, with no warning. Short prompts route correctly; disabling `chunked_prefill` makes the long ones route correctly too.

Root cause: the VLM MTP routing decision exists at a single point on the **non-chunked** prefill exit of `_schedule_waiting`. A chunked-prefill'd request completes via `_insert_prefilled_request` and `continue`s past that point (or, for multi-step chunks, completes in `_advance_chunked_prefills`, which never had the routing code at all), dropping straight into `BatchGenerator`. So the routing is skipped as collateral of the chunked branch's early `continue`, not because of any eligibility check. That also explains the missing warning: chunked requests never enter `_route_to_vlm_mtp`, so none of its fallback log lines can fire.

This is unrelated to #1758 (the native-MTP per-model marker): native MTP is re-checked every decode step inside `BatchGenerator`, so it survives chunked prefill; external VLM MTP is a scheduler-level, one-time routing branch that *bypasses* `BatchGenerator`, decided only on the non-chunked exit.

## Fix

Re-apply the routing at the top of `_insert_prefilled_request` — the single choke point both chunked-completion paths (`_schedule_waiting`'s inline first chunk and `_advance_chunked_prefills`) funnel through. It already holds everything `_route_to_vlm_mtp` needs on `state` (`cache`, `last_token`, `sampler`, `sm`).

- Runs **before** the TurboQuant cache convert (`_finalize_chunked_prefill_cache_for_insert`), so the drafter's final forward sees the un-quantized cache.
- Mirrors the negative-uid bookkeeping of the non-chunked routing path and `return`s, so a routed request is never also inserted into `BatchGenerator`.
- A declined route (e.g. the single-in-flight drafter is already busy under concurrency) falls through to `BatchGenerator`, exactly as short prompts already do.
- Text prompts carry the default `rope_deltas=0.0`, so mRoPE handling matches the non-chunked path.

## Validation

New `tests/test_vlm_mtp_chunked_prefill.py` drives the real `_insert_prefilled_request` with a faked drafter / BatchGenerator (no model loaded) and asserts: an eligible drafter routes to VLM MTP (negative uid + bookkeeping); an ineligible one falls back to `BatchGenerator`; a routed request is scheduled exactly once. These fail on the unfixed code (the routing is never consulted) and pass with the fix.

```
uv run pytest tests/test_vlm_mtp_chunked_prefill.py tests/test_vlm_mtp.py \
  tests/test_scheduler.py tests/test_scheduler_chunked_prefill.py tests/test_engine_core.py
=> 335 passed
```

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -4292,6 +4292,36 @@ class Scheduler:
 
         Precondition: state.sampler, state.sm, state.per_row_lps are set.
         """
+        # #2219: both chunked-completion paths (inline first chunk and
+        # _advance_chunked_prefills) funnel through here, but external VLM MTP
+        # routing only existed on the non-chunked prefill exit -- so any prompt
+        # long enough to be chunk-prefilled silently bypassed VLM MTP. Re-apply
+        # the decision at this shared choke point, BEFORE the TurboQuant convert
+        # below, so the drafter's final forward sees the un-quantized cache. A
+        # declined route (e.g. the single-in-flight drafter is busy) falls
+        # through to BatchGenerator, matching short-prompt behaviour.
+        if self._vlm_mtp_drafter is not None and state.cache is not None:
+            vlm_mtp_uid = self._route_to_vlm_mtp(
+                request, state.cache, state.last_token, state.sampler, state.sm
+            )
+            if vlm_mtp_uid is not None:
+                self.request_id_to_uid[request.request_id] = vlm_mtp_uid
+                self.uid_to_request_id[vlm_mtp_uid] = request.request_id
+                now = time.monotonic()
+                request.batch_uid = vlm_mtp_uid
+                request.status = RequestStatus.RUNNING
+                request.generation_started_at = now
+                request.last_activity_at = now
+                self.running[request.request_id] = request
+                scheduled.append(request)
+                self.total_prompt_tokens += request.num_prompt_tokens
+                logger.debug(
+                    "Scheduled chunked-prefill request %s via vlm_mtp (uid=%d)",
+                    request.request_id,
+                    vlm_mtp_uid,
+                )
+                return
+
         self._finalize_chunked_prefill_cache_for_insert(request, state.cache)
 
         if request.sampling_params.seed is not None:
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -4292,6 +4292,9 @@ class Scheduler:
 
         Precondition: state.sampler, state.sm, state.per_row_lps are set.
         """
+        if request.sampling_params.seed is not None:
+            mx.random.seed(request.sampling_params.seed)
+
         # #2219: both chunked-completion paths (inline first chunk and
         # _advance_chunked_prefills) funnel through here, but external VLM MTP
         # routing only existed on the non-chunked prefill exit -- so any prompt
@@ -4323,9 +4326,6 @@ class Scheduler:
                 return
 
         self._finalize_chunked_prefill_cache_for_insert(request, state.cache)
-
-        if request.sampling_params.seed is not None:
-            mx.random.seed(request.sampling_params.seed)
 
         per_row_lps = state.per_row_lps if state.per_row_lps is not None else []
         # insert() merges the prompt cache into the batch KV caches with lazy
--- a/tests/test_vlm_mtp_chunked_prefill.py
+++ b/tests/test_vlm_mtp_chunked_prefill.py
@@ -28,10 +28,11 @@ from omlx.scheduler import Scheduler
 
 
 def _make_fixture(monkeypatch, drafter_returns_uid):
-    calls = {"route": 0, "bg_insert": 0}
+    calls = {"route": 0, "bg_insert": 0, "events": []}
 
     def fake_route(request, cache, last_tokens, sampler, sm):
         calls["route"] += 1
+        calls["events"].append("route")
         return -7 if drafter_returns_uid else None  # negative uid, or ineligible
 
     def fake_bg_insert(*args, **kwargs):
@@ -86,7 +87,9 @@ def test_chunked_prefilled_request_routes_to_vlm_mtp_when_eligible(monkeypatch):
     Scheduler._insert_prefilled_request(sched, request, state, scheduled)
 
     assert calls["route"] == 1, "vlm_mtp routing was never considered (the #2219 bug)"
-    assert calls["bg_insert"] == 0, "request went to BatchGenerator despite eligible vlm_mtp"
+    assert (
+        calls["bg_insert"] == 0
+    ), "request went to BatchGenerator despite eligible vlm_mtp"
     # negative-uid bookkeeping mirrors the non-chunked routing path
     assert request.batch_uid == -7
     assert request.status == RequestStatus.RUNNING
@@ -95,6 +98,24 @@ def test_chunked_prefilled_request_routes_to_vlm_mtp_when_eligible(monkeypatch):
     assert sched.running["req-long-text"] is request
     assert request in scheduled
     assert sched.total_prompt_tokens == 32768
+
+
+def test_seed_is_applied_before_vlm_mtp_sampling(monkeypatch):
+    """A successful VLM MTP route must honor the request seed before sampling."""
+    sched, request, state, scheduled, calls = _make_fixture(
+        monkeypatch, drafter_returns_uid=True
+    )
+    request.sampling_params.seed = 123
+
+    monkeypatch.setattr(
+        scheduler_mod.mx.random,
+        "seed",
+        lambda seed: calls["events"].append(("seed", seed)),
+    )
+
+    Scheduler._insert_prefilled_request(sched, request, state, scheduled)
+
+    assert calls["events"] == [("seed", 123), "route"]
 
 
 def test_falls_back_to_batch_generator_when_drafter_ineligible(monkeypatch):
--- a/tests/test_vlm_mtp_chunked_prefill.py
+++ b/tests/test_vlm_mtp_chunked_prefill.py
@@ -1,0 +1,124 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Reproduction + fix test for #2219.
+
+External VLM MTP routing lives only on the non-chunked prefill exit
+(``scheduler.py`` ~8504). Chunked-prefill'd requests complete via
+``_insert_prefilled_request`` and go straight to ``BatchGenerator``, so any prompt
+long enough to be chunked silently bypasses VLM MTP.
+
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for putting this together and for tracing the chunked-prefill routing gap so clearly. The shared `_insert_prefilled_request()` routing point is the right fix, and the tests cover both the VLM MTP path and the BatchGenerator fallback well. I added a small follow-up for seed ordering, and the PR looks good now. Thanks again for the solid contribution.

Issue #2356, opened 2026-07-24
MiniMax-M3: honor pack_shared_expert (unpacked mixed-bit shared expert fails to load in repacker)


avlp12:
## Summary

oMLX cannot load MiniMax‑M3 MLX checkpoints that use the **unpacked shared‑expert layout** (routed experts and the shared expert quantized at *different* bit‑widths). The repacker in `omlx/engine/vlm.py` unconditionally reconstructs a single 129‑wide `SwitchLinear`, and an MLX packed bank can hold only one bit‑width, so it fails when the shared expert differs from the routed bank.

## Error

```
[concatenate] All the input array dimensions must match exactly except for the
concatenation axis. However, the provided shapes are (128,6144,1152), (1,6144,1536),
and the concatenation axis is 0.
```

`1152` and `1536` are the packed widths of a 6144‑wide dim at **6‑bit** vs **8‑bit** MLX‑affine (`6144*6/32=1152`, `6144*8/32=1536`). So this is a 6‑bit routed bank being concatenated with an 8‑bit shared expert.

## Root cause (from reading the source)

- `_pack_mlx_unpacked_moe_weights` — `omlx/engine/vlm.py:906‑914` — builds `routed_gate_up (128,6144,1152)` and `shared_gate_up (1,6144,1536)` and calls `mx.concatenate([routed_gate_up, shared_gate_up], axis=0)`. Axis 0 is fine; the packed inner widths differ, so it throws.
- The pack decision is derived only from `n_shared_experts == 1 and shared_intermediate_size == intermediate_size` (`omlx/engine/vlm.py:881`, and the vendored `minimax_m3_vl.py:37` / `language.py:1368`). There is **no** `pack_shared_expert` flag read, and unknown config keys are dropped (`config.py`), so a checkpoint carrying `text_config.pack_shared_expert=false` is silently re‑packed anyway.

## Why this layout exists

MiniMax‑M3's shared expert processes **100% of tokens** (a routed expert sees ~3% at top‑4/128), so holding it at 8‑bit while the routed bank drops to 3/6‑bit is a real quality‑per‑GB lever. That mixed‑bit combination can only be represented in the **unpacked** layout — upstream mlx‑vlm added an opt‑in for exactly this: [`pack_shared_expert`](https://github.com/Blaizzy/mlx-vlm/pull/1544).

## Requests

[...]

PR #2363, opened 2026-07-24
fix(vlm): advance rope positions across multi-token decode windows

gilbert-barajas:
`VLMModelAdapter._position_ids_from_starts` broadcasts each row's start offset across the whole window, so for `seq_len > 1` every position in the window is rope-rotated at the *first* position instead of advancing.

The consuming attention builds its own positions when none are supplied (`mlx_vlm/models/qwen3_vl_moe/language.py`):

```python
if position_ids is None:
    position_ids = mx.arange(cache.offset, cache.offset + L)
    position_ids = mx.tile(mx.expand_dims(position_ids, 0), (3, 1, 1))
```

so the positions this method passes must advance to match. Prefill (`get_rope_index`) and the MiniMax branch already do — the default mRoPE branch was the only position builder in the stack emitting a constant, and it was never updated when multi-token decode windows were added.

### Why it matters

Multi-token windows arise in speculative-decode verify, where the effect is not just a wrong forward: the mis-rotated keys are **committed into the KV cache on accept**.

Measured on Qwen3.6-35B-A3B, verify attention keys diverged O(1) from an autoregressively-built cache while values matched bit-exactly — the signature of a position error, since rope rotates q and k but not v. With positions advancing, a T-row verify matches T single-row decodes exactly (logits and every layer's cache state) on the prefixes we certified.

### Scope

- **Single-token decode is unchanged by construction.** At `seq_len == 1` the step offsets are `[0]`, so positions reduce to the start offset itself — same shape, same dtype, same values as before.
- Affects the **mRoPE default branch only** (Qwen2/2.5/3-VL-class). MiniMax is a no-op; text-only serving does not use this adapter.
- Two call sites, both decode/verify dispatch; the output is consumed only by the attention rope, so there is no path that was compensating for the old behaviour.

### Tests

Adds a regression test covering multi-token advancement on the default branch — previously only the `seq_len == 1` case and the MiniMax branch (already correct) were pinned. Against the unfixed code it fails with `assert 50 == 51.0`; with the fix the full adapter suite passes (32 tests).

### DIFF
--- a/omlx/models/vlm.py
+++ b/omlx/models/vlm.py
@@ -253,12 +253,24 @@ class VLMModelAdapter(nn.Module):
     def _position_ids_from_starts(
         self, starts: mx.array, batch_size: int, seq_len: int
     ) -> mx.array:
-        """Build model-specific decode position_ids from per-row start offsets."""
+        """Build model-specific decode position_ids from per-row start offsets.
+
+        Positions are sequential from each row's start: row r covers
+        ``starts[r] .. starts[r] + seq_len - 1``. At single-token decode
+        (seq_len == 1) this reduces to the start offset itself; for
+        multi-token windows (the MTP verify rows) each position must
+        advance, or every draft row gets rope-rotated at the *first* row's
+        position — measured on Qwen3.6-35B-A3B as verify keys diverging
+        O(1) from an AR-built cache while values matched, breaking
+        verify/decode parity (and silently mis-rotating the KV the verify
+        writes for accepted rows).
+        """
         starts = starts.reshape(-1)[:batch_size]
+        steps = mx.arange(seq_len, dtype=starts.dtype)
+        seq_positions = starts[:, None] + steps[None, :]
         if self._uses_minimax_m3_positions:
-            steps = mx.arange(seq_len, dtype=starts.dtype)
-            return starts[:, None] + steps[None, :]
-        return mx.broadcast_to(starts[None, :, None], (3, batch_size, seq_len))
+            return seq_positions
+        return mx.broadcast_to(seq_positions[None, :, :], (3, batch_size, seq_len))
 
     def get_last_rope_deltas(self) -> float:
         """Extract rope_deltas from language model after VLM prefill.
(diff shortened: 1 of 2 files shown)


gilbert-barajas:
Re-verified against `main` @ `41772940` (includes #2356) before asking for a look:

- **Cherry-picks clean** onto current `main` — 2 files, +52/−4, no conflicts.
- **Still needed.** `omlx/models/vlm.py:253` on `main` is unchanged: the MiniMax branch advances positions (`starts[:, None] + steps[None, :]`) while the default mRoPE branch still returns `mx.broadcast_to(starts[None, :, None], (3, batch_size, seq_len))` — a constant across the window. That asymmetry within the same function is the bug.
- **Tests green.** Full `tests/test_vlm_model_adapter.py` on `main` + this patch: 32/32 pass. The added regression test fails without the fix (`assert 50 == 51.0`) and passes with it.
- **Single-token decode is unchanged by construction** — at `seq_len == 1` the step offsets are `[0]`, so positions reduce to the start offset with the same shape and dtype as before.

Worth noting the impact isn't a wrong forward in isolation: in speculative-decode verify the mis-rotated keys are **committed into the KV cache on accept**, so the corruption persists past the window. Measured on Qwen3.6-35B-A3B, verify attention keys diverged O(1) from an autoregressively-built cache while values matched bit-exactly — the signature of a position error, since rope rotates q and k but not v.

CI shows `action_required` (0 runs) — the fork-PR approval gate rather than a failure; happy to rebase if anything above has drifted.

jundot:
Thanks for catching this, I had missed it entirely. The failure mode is quiet enough that normal output still looks fine, so it would have kept sitting there.

- Traced it on Qwen3.5-9B with MTP on: the verify windows arrive as L=2..4 and each window start advances by exactly the window length, so the positions inside the window have to advance too.
- Ran a window vs sequential parity check on two mRoPE models. Without the fix the trailing rows collapse onto the first row's prediction (Qwen2-VL-2B: argmax 15/15/13 where sequential decode gives 362/692/18, max logit delta 16.6; Qwen3.5-9B-oQ4: row 3 argmax 25 vs 2110). With the fix every row matches within quantization noise.
- Regular VLM decode only ever calls this with seq_len 1, so nothing changes on that path.

Really good find. I'll merge this.

PR #2364, opened 2026-07-25
fix: support Qwen3-VL image embeddings

Wade1080:
## Summary
- adapt the injected torch-free Qwen3-VL image processor to Transformers nested chat-template batches
- decode validated inline image data URIs with the existing oMLX image loader
- flatten decoded image batches before mlx-vlm preprocessing
- add regression coverage for nested images and inline data URIs

## Root cause
Transformers passes multimodal chat-template images as nested lists. The mlx-vlm Qwen3-VL processor handled only flat lists, converting an inner list to a one-dimensional array. After preserving the batch structure, inline data URIs were also treated as filesystem paths. Both paths prevented image and mixed embeddings while text embeddings remained healthy.

## Validation
- ruff check passed for changed files
- 93 related tests passed
- real mlx-community/Qwen3-VL-Embedding-2B-6bit produced finite normalized 2048-dimensional text, image, and mixed vectors
- temporary fork server returned HTTP 200 and normalized 768-dimensional vectors for text, image, and mixed API requests
- full non-slow suite: 6981 passed, 54 skipped; 2 unrelated GLM numerical tolerance tests failed because the local checkout lacks the optional native Metal kernels, and both reproduce independently

### DIFF
--- a/omlx/models/mlx_embeddings_compat.py
+++ b/omlx/models/mlx_embeddings_compat.py
@@ -42,6 +42,39 @@ def patch_qwen3_vl_processor_for_torch_free_image_loading() -> None:
         qwen3_vl_processor, "AutoImageProcessor", None
     )
 
+    class OMLXQwen3VLImageProcessor(Qwen3VLImageProcessor):
+        """Preserve batched image nesting expected by ProcessorMixin.
+
+        Transformers passes multimodal chat-template images as a nested
+        ``list[list[image]]``. mlx-vlm's Qwen3-VL override only handles a flat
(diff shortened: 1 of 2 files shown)


Wade1080:
Hi @jundot, I noticed that ed30fe9 landed a broader version of the same Qwen3-VL embedding fix, including nested image batches, data URIs, and empty text-only slots. Thanks for addressing it. This PR is now superseded and conflicts with main, so I’m closing it.

For reference, the validation here also covered text, image, and mixed embedding requests on Qwen3-VL-Embedding-2B-6bit.

PR #2365, opened 2026-07-25
feat(nemotron_h): fused Metal decode-step kernels + per-position SDPA for verify windows

davidtorcivia:
Follow-up to #2277/#2278. Nemotron-H decode spends roughly half of every
cycle above its weight-bandwidth floor, and a large share of that is the
Mamba mixer's per-layer decode glue: ~13 dispatches per layer (conv-state
concat/slice, depthwise Conv1d, silu, compute_dt, the ssm step kernel,
swiglu, grouped rms_norm, plus splits/reshapes), times 40 layers, times
one more sequential pass per position in MTP verify windows.

## Commit 1: fused decode-step kernels

Two `mx.fast.metal_kernel` JIT kernels (no Xcode required, same approach
as the existing gdn-style kernels) replace that glue:

- **`mamba2_decode_fused`**: conv-state update + depthwise conv + silu +
  softplus(dt) + SSD state update in one dispatch, with the S-position
  loop **inside the kernel** and the SSM state register-resident across
  positions. MTP chain verify windows (1 < S ≤ 8) run in this same single
  dispatch and capture the per-position (conv, ssm) rollback states
  in-kernel, superseding the chain's sequential per-position verify loop
  and its Python conv-tail capture while keeping `mtp_partial_rollback`'s
  pure ref-restore fast path.
- **`mamba2_gatenorm_fused`**: swiglu + grouped RMS norm + weight.

The mixer decode step becomes in_proj, kernel A, kernel B, out_proj
(4 dispatches).

### Numerics

Deliberately mirrors the stock path cast-for-cast (conv accumulates fp32,
rounds to the IO dtype, silu, rounds again; dt softplus in fp32; SSD math
op-for-op identical to
`ssm_update_kernel`; swiglu/norm rounding chain matched). Tests are
**truth-anchored**: the bf16 fused output must track an fp32 ground truth
as closely as the stock bf16 path does, which separates real defects from
accumulated-rounding noise.

### Measured (NVIDIA-Nemotron-3-Super-120B oQ5, M3 Ultra, greedy, MTP depth-1)

| | stock | fused |
|---|---|---|
| single layer, S=2 verify | 0.184 ms | 0.113 ms (**1.62x**) |
| single layer, S=3 | 0.268 ms | 0.128 ms (2.10x) |
| single layer, S=4 | 0.353 ms | 0.147 ms (2.40x) |
| decode cycle @ 9.6k ctx | 31.4 ms | 29.3 ms |
| decode throughput | 51.2 tok/s | **55.4 tok/s (+8%)** |

Acceptance rate is unchanged (64.2% vs 62.6% on the same prompt, within
trajectory noise). S=1 plain decode sits at the dispatch floor either
way; the speedup comes from verify windows, which are every MTP cycle.

## Commit 2: per-position SDPA for 3 ≤ S ≤ 8 verify windows

`mx.fast.scaled_dot_product_attention` has a fast path for S ≤ 2 queries
but falls onto a full-attention path at S ≥ 3 that is pathologically slow
at decode context lengths: measured 1.56 ms/layer at 10k tokens and
4.0 ms at 32k, vs ~0.3 ms for S ≤ 2 (GQA 32/2, head_dim 128). MTP verify
windows at draft depth ≥ 2 are exactly S ∈ 3..depth+1, so every
deeper-than-1 cycle paid this on all attention layers, a hidden tax that
skewed the adaptive depth controller's cost estimates.

Decomposing such windows into S single-position SDPA calls over sliced KV
is exactly causal (max err 1e-6 vs the fused path) and 3.6x to 6x faster at
S ≥ 3, scaling with context. S ≤ 2, S > 8, masked-array, and
quantized-KV paths fall through unchanged.

With this fix in place, the remaining depth-2 blocker on Nemotron-3-Super
is the MTP head's own second-step conditional acceptance (~35% measured
on prose with a depth-1-trained head), not serving overhead. This was
measured via a matched 800-token A/B: 54.9 tok/s at depth-1 vs 53.8 at
depth-2 cap, the gap being probe duty.

## Gating

Fused kernels: cache warm, mask/lengths `None`, 1 ≤ S ≤ 8, bf16/fp16,
`Dh ∈ {32, 64, 128}`, `Ds` a multiple of 32, gated-norm group size a
multiple of 256. Everything else (prefill chunks, ragged batches, exotic
dims, S > 8 windows) falls through to the previous call chain unchanged.
Applied at engine init after the sdpa256 hook; idempotent; composes with
the mlx_lm_mtp wraps in either presence.

Tests: 24 cases: dtype × S × warm-state matrix, per-position capture vs
prefix replay, verify-window semantics + rollback against the real chain
layering, per-position SDPA equivalence (byte-identical fall-through at
S=2 and S=12), unsupported-dims fall-through, idempotent re-apply. Both
orderings with `test_nemotron_mtp_patch.py` pass (31 tests).

### DIFF
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -346,6 +346,22 @@ class BatchedEngine(BaseEngine):
             except Exception:
                 logger.debug("sdpa256 attention patch not applied", exc_info=True)
 
+        # Nemotron-H decode step -> two fused Metal kernels (conv+dt+SSD
+        # with an in-kernel S-loop and register-resident state, plus
+        # swiglu+grouped-RMSNorm), replacing the per-layer decode glue and
+        # the mlx_lm_mtp chain's sequential verify loop; MTP verify windows
+        # get their per-position rollback states captured in-kernel.
+        try:
+            from ..patches.nemotron_decode_fused import (
+                apply_nemotron_decode_fused_patch,
+            )
+
+            apply_nemotron_decode_fused_patch()
+        except Exception:
+            logger.debug(
+                "Nemotron fused decode patch not applied", exc_info=True
+            )
+
         # Qwen3.5/3.6 head_dim=256 causal prefill -> native steel FA kernel.
         # Strictly shape-gated; decode, quantized-cache paths, and unsupported
         # models fall through to the previous SDPA implementation.
--- a/omlx/patches/nemotron_decode_fused.py
+++ b/omlx/patches/nemotron_decode_fused.py
@@ -37,7 +37,6 @@ ragged batches, and prefill chunks fall through to the previous call chain
 
 import logging
 import math
-from typing import Optional, Tuple
 
 import mlx.core as mx
 
@@ -303,12 +302,22 @@ def _get_kernels():
         _kernel_a = mx.fast.metal_kernel(
             name="mamba2_decode_fused",
             input_names=[
-                "proj", "conv_state_in", "ssm_state_in", "conv_w", "conv_b",
-                "A_log", "dt_bias", "D", "lim",
+                "proj",
+                "conv_state_in",
+                "ssm_state_in",
+                "conv_w",
+                "conv_b",
+                "A_log",
+                "dt_bias",
+                "D",
+                "lim",
             ],
             output_names=[
-                "y_out", "conv_state_out", "ssm_state_out",
-                "cap_state", "cap_conv",
+                "y_out",
+                "conv_state_out",
+                "ssm_state_out",
+                "cap_state",
+                "cap_conv",
             ],
             header=_HEADER,
             source=_KERNEL_A_SRC,
@@ -327,7 +336,8 @@ def supported(num_heads, head_dim, state_size, n_groups, conv_kernel, group_size
     return (
         conv_kernel == 4
         and head_dim in (32, 64, 128)
-        and 32 <= state_size <= 256 and state_size % 32 == 0
+        and 32 <= state_size <= 256
+        and state_size % 32 == 0
         and num_heads % n_groups == 0
         and (num_heads * head_dim) % group_size == 0
         and group_size % 256 == 0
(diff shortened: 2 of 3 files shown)


jundot:
Thanks for the detailed optimization work. The Ds=128, bias-enabled path passes the existing suite and a real Nemotron smoke, but I found three gating gaps that need to be fixed before merge: the 256-thread B/C staging is only valid at Ds=128 while `supported()` admits 32..256; `use_conv_bias=False` crashes on `mixer.conv1d.bias`; and an empty `QuantizedKVCache` fails on `tuple.shape` during a short prefill. Please either generalize those paths or tighten the gates and add regression coverage for each case.

davidtorcivia:
Thanks for the careful gating review. All three fixed in e6df433d:

- **Ds gate.** You are right that the 256-thread B/C staging is only a complete, in-bounds cover at Ds=128 (2*Ds == 256, and the conv-state tail writeback uses the same mapping). I tightened `supported()` to `state_size == 128` rather than generalizing: every shape the kernel was validated against is Ds=128, and other sizes now fall through to the stock path. Happy to generalize the staging to a channels-per-thread loop in a follow-up if there is a real model that needs it.
- **conv bias.** `use_conv_bias=False` builds a bias-less Conv1d; the fused path now feeds the kernel zeros instead of reading `mixer.conv1d.bias`. Verified the fused path still engages (not a silent fall-through) and matches stock without the bias.
- **Empty QuantizedKVCache.** `_dense_kv` gated on the current `keys` value, so an empty quantized cache slipped through and the first `update_and_fetch` in a short prefill returned tuples. It now gates on the cache's quantization attribute (`bits`), so quantized caches fall through whether empty or warmed.
- **Regression coverage.** New tests: non-128 state sizes are rejected by `supported()` and the wrapped mixer falls through cleanly; a bias-less mixer runs the fused path and matches stock; QuantizedKVCache falls through byte-identically both empty (short prefill) and warmed.

PR #2366, opened 2026-07-25
perf(cache): store large fp32 boundary-snapshot states as int8

davidtorcivia:
Hybrid-model boundary snapshots carry the Mamba SSM states in fp32:
about **168 MB per snapshot** for a 40-Mamba-layer Nemotron-3-Super.
Every snapshot is serialized on the inference thread and staged through
the pending-write buffer to SSD (`cache/boundary_snapshot_store.py`).

This PR stores any large fp32 state tensor (≥64k elements, ndim ≥ 2) as
symmetric int8 with per-row (last axis) fp32 scales:

- **a measured 170 MB drops to 46 MB per Nemotron-3-Super snapshot
  (3.7x)** for serialize copies, pending-write RAM, and SSD files;
- elementwise error is bounded by rowmax/254, measured at 4e-3 relative
  worst case. 8-bit SSM states are established as accuracy-safe
  (Quamba/Quamba2, ICML 2025);
- bf16 conv states and KV layers are untouched (bit-exact).

Both readers dequantize transparently (the raw-bytes pending path and the
safetensors disk path). Snapshots are ephemeral per-request files, so
there is no cross-version durability concern: a snapshot written with
this patch is always read by the same code. Opt out with
`OMLX_SNAPSHOT_Q8=0`.

Verified end-to-end on a 9.6k-token two-turn conversation with a hybrid
model: the follow-up turn resumes from a dequantized block-boundary
snapshot with unchanged output and a slightly faster turn (2.59 s vs
2.86 s, from the smaller inference-thread serialize copy).

Tests: round-trip through both reader paths with mixed
ArraysCache/KVCache layers, exactness of conv/KV payloads, size bound,
error bound.

### DIFF
--- a/omlx/cache/boundary_snapshot_store.py
+++ b/omlx/cache/boundary_snapshot_store.py
@@ -42,19 +42,27 @@ logger = logging.getLogger(__name__)
 # Max pending writes before save() blocks.
 _MAX_PENDING_WRITES = 128
 
-# Large fp32 snapshot states (hybrid-model SSM states) are stored as int8
-# with per-row fp32 scales: 4x smaller serialize copies, pending buffers,
-# and files; error bounded by rowmax/254. Disable via OMLX_SNAPSHOT_Q8=0.
+# Mamba2-family SSM states (fp32, shape (B, num_heads, head_dim, state_size))
+# are stored as int8 with per-row fp32 scales: 4x smaller serialize copies,
+# pending buffers, and files; error bounded by rowmax/254. 8-bit storage is
+# established as accuracy-safe for this state type (Quamba/Quamba2, ICML'25);
+# other fp32 cache states (e.g. gated-deltanet recurrent states) are NOT
+# covered by that result and stay fp32. The caller identifies eligible layers
+# semantically via ``ssm_state_shapes`` (see ``save``); a tensor is quantized
+# only when it also matches the expected per-layer state shape and dtype.
+# Disable via OMLX_SNAPSHOT_Q8=0.
 _Q8_ENABLED = os.environ.get("OMLX_SNAPSHOT_Q8", "1") != "0"
 _Q8_MIN_ELEMS = 65536
 
 
-def _q8_eligible(elem) -> bool:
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for working on the boundary snapshot footprint. The current eligibility check applies to every large flat fp32 cache state, not just Nemotron/Mamba SSM state; with Qwen3.5-0.8B I saw temperature-0 greedy output diverge after restoring the q8 snapshot. The pending path also returns the original extracted state while the disk path returns the dequantized state, so the restored cache depends on writer timing. Could you scope q8 to a semantically identified and validated state type, make both load paths restore the same representation, and add cache-hit output regression coverage?

davidtorcivia:
Thanks for catching both of these, and especially for the Qwen3.5 repro. Fixed in 5bfec175:

- **Scoping.** Eligibility is no longer a dtype/size heuristic. The scheduler now walks the model once and records, per layer, the expected Mamba2 SSM state shape `(num_heads, head_dim, ssm_state_size)` from mixers that carry the Mamba2 attribute set (`num_heads` / `head_dim` / `ssm_state_size` / `conv1d` / `A_log`, which matches `mamba2`, `nemotron_h`, `granitemoehybrid`, and `falcon_h1` in mlx-lm and does not match gated-deltanet or KDA mixers). The store quantizes only fp32 tensors on those layers that also match the expected 4-D state shape, which is the state type Quamba2 validated for 8-bit storage. On Qwen3.5 the map is empty, so every state is stored bit-exact.
- **Writer-timing dependence.** `save()` now swaps each quantized element of the pending fast-path copy for its dequantized value (materialized on the capturing thread), so the pending fast path, the raw-bytes pending path, and the disk path all restore the identical representation.
- **Regression coverage.** New tests assert: unscoped states restore bit-exact; a scoped layer with a non-matching (deltanet-style) state shape is not quantized; all three read paths restore identical arrays; and greedy cache-hit continuations on a tiny in-process mamba2 model are read-path independent when scoped and exactly equal to the uninterrupted continuation when unscoped.

PR #2369, opened 2026-07-25
fix(scheduler): measure prefill chunk transients as peak spike, not footprint delta

gilbert-barajas:
## What

The adaptive prefill throttle currently feeds its EWMA tracker with a
**phys-footprint delta** measured across each chunk
(`get_phys_footprint()` before/after in `_do_external_prefill` and
`_step_prefill_chunk`). That delta includes everything that became
resident during the chunk — most importantly the warm-restored chain's
lazy materialization (proportional to **kv_len**, not chunk size) and
Metal buffer-pool growth — so the per-token estimate is poisoned upward
at long context. The guard then multiplies the poisoned estimate into a
predicted transient of many GB, shrinks chunks to a crawl, and finally
rejects healthy requests at the pre-chunk guard.

This is **defect 1 of #1842** (@hojin12312's report — the measurements
there show a 1024-token chunk at 137K kv_len recording a 15.1 GB
"transient" and per-token estimates up to 40 MB/token). Defects 2–4 and
the sub-`min_chunk` tail-skip landed in a803e629; the measurement change
itself — the main fix for defect 1 — is still open. This PR implements
it.

## How

- Sample `mx.get_peak_memory()` immediately before the chunk. After the
  chunk's eval, when the peak advanced, record
  `peak − mx.get_active_memory()` — the memory that was allocated during
  the chunk and released again by the end of its eval (the SDPA/MoE
  transient). Resident growth (KV append, restore materialization) stays
  in `mx.get_active_memory()` and is **not** charged to the chunk: it is
  already part of the guard's `current` baseline at the next check, so
  the old delta double-counted it.
- The global peak counter is **not reset**: a per-chunk
  `mx.reset_peak_memory()` would corrupt concurrent readers — the admin
  benchmark brackets `stream_generate` with reset/get
  (`admin/benchmark.py`). When a chunk sets no new high-water mark its
  local peak is unobservable, so no sample is recorded and the throttle
  falls back to the static estimate — which remains the `max()` floor in
  `_predicted_chunk_transient`, leaving cold/unfused protection
  unchanged. In the regime where the throttle actually matters (memory
  climbing toward the cap) successive chunks keep setting new highs, so
  samples flow exactly when they are needed.
- `_record_chunk_transient` now takes the measured spike directly;
  0 means "no sample" and leaves the tracker untouched.

Files: `omlx/scheduler.py` (+ new `_measure_chunk_spike` helper),
`omlx/prefill_transient_tracker.py` (docstrings),
`tests/test_prefill_oom_graceful.py`.

## How verified

- **Unit tests**: `test_prefill_oom_graceful.py`,
  `test_prefill_transient_tracker.py`, `test_scheduler_chunked_prefill.py`,
  `test_scheduler.py` — 261 passed. New tests cover the spike math
  (peak-over-end-active), the no-new-peak abstention, and that
  resident-only growth records **no** sample (the core of defect 1).
- **Live A/B smoke** (M5 Max 128GB, Qwen3-4B-Instruct-2507-4bit,
  14,016-token prompt, 7 prefill chunks, identical request against this
  branch vs its base commit):
  - old delta: 1281 MB on chunk 1 (~2× the true spike) and a 2.8×
    chunk-to-chunk swing, 232–640 KB/token;
  - spike measurement: stable 308–385 KB/token across all 7 chunks
    (603–770 MB), no first-chunk distortion.
- **Scope honesty**: I did not reproduce the 136K→150K multi-turn
  ceiling change reported in #1842 — that regime needs the warm-restore
  + hot-cache setup of the original report. The mechanism (resident
  growth misattributed to the chunk) is what this PR removes, and the
  cold-prefill A/B above shows the measurement noise/inflation directly.
- `ruff check` on the touched files: finding set identical to base
  (zero new findings).

Closes the remaining defect of #1842 (defects 2–4 landed in a803e629).
Credit for the root-cause analysis and the fix design to @hojin12312.

---

Verified against `main` @ `41772940` before submitting: cherry-picks clean (auto-merge over a803e629), `mx.get_peak_memory` is still absent from `scheduler.py` so defect 1 is untouched, and the throttle feed points still sample `get_phys_footprint()` deltas.

### DIFF
--- a/omlx/prefill_transient_tracker.py
+++ b/omlx/prefill_transient_tracker.py
@@ -16,8 +16,10 @@ from __future__ import annotations
 class PrefillTransientTracker:
     """EWMA estimator of MLX prefill chunk transient bytes per token.
 
-    Updated post-chunk from `phys_footprint()` deltas. The first chunk
-    has no measurement yet — callers fall back to a static estimate
+    Updated post-chunk from the measured peak-over-end spike (memory that
+    was allocated during the chunk and released again by the end of its
+    eval — see scheduler._measure_chunk_spike). The first chunk has no
+    measurement yet — callers fall back to a static estimate
     (MemoryMonitor.estimate_prefill_peak_bytes) until samples > 0.
     """
 
@@ -34,9 +36,9 @@ class PrefillTransientTracker:
     def update(self, n_tokens: int, transient_bytes: int) -> None:
         """Record one chunk observation.
 
-        Negative deltas (MLX cache pool reclaim larger than this chunk's
-        allocation) are skipped — they would bias the EWMA toward zero
-        and underestimate the next chunk's footprint.
+        Non-positive observations (the chunk produced no measurable spike)
+        are skipped — they would bias the EWMA toward zero and
+        underestimate the next chunk's footprint.
         """
         if n_tokens <= 0:
             return
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -211,6 +211,33 @@ def _safe_sync_stream(stream=None):
             raise
 
 
+def _measure_chunk_spike(peak_pre: int) -> int:
+    """Come-and-went Metal memory for the prefill chunk that just ran.
+
+    ``peak_pre`` is ``mx.get_peak_memory()`` sampled immediately before the
+    chunk. When the chunk pushed the process high-water mark above it, the
+    spike is ``peak - end_of_chunk_active``: memory that was allocated during
+    the chunk and released again by the time its eval finished (SDPA/MoE
+    intermediates). Resident growth (this chunk's KV append, lazily
+    materialized warm-restore arrays) stays in ``mx.get_active_memory()`` and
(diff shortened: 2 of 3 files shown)


jundot:
@gilbert-barajas Thanks for the careful analysis and live A/B data. I dug further into the allocator behavior while working on #2390, and the key premise here does not hold for the limit the guard must protect: MLX active/peak memory does not account for cached buffers, while `phys_footprint` includes buffer-pool growth that consumes real process and wired headroom.

Allocator tracing showed that `dphys` tracks `dcache` much more closely than `dactive`, and switching sizing to the active-peak signal failed the gemma 60k safety repro; observing that pool growth only in the next chunk's baseline is too late for the chunk that crosses the limit. #2390 keeps the physical safety signal but addresses its noise through chunk quantization, floor-size-only observed bounds, exact KV admission, watermark alignment, and reclaim-before-abort.

Since this change would remove a required safety signal and #2390 now covers the broader failure mode, I'm going to close this as superseded by #2390.

PR #2370, opened 2026-07-25
perf(embedding): group a batch by input length, restore caller order

1qh:
## Problem

Each forward pass pads every input to its longest member, so a batch of mixed lengths spends most of its compute on padding. `EmbeddingEngine.embed` slices `input_items` in caller order, so whatever length distribution the caller happens to send is what each batch pays for.

Real corpora are heterogeneous. In one document set the chunks run a **median of 37 characters, a p90 of 400, and a maximum in the thousands** — so a batch is routinely one long input plus dozens of short ones, all padded to the long one. The waste is the common case, not an edge case.

This is invisible from outside: the batching is internal, so a caller cannot fix it by sorting its own request — the engine re-slices whatever it is given.

## Fix

Sort inputs by length before batching, and restore the caller's order before returning. The response contract is unchanged; only the padding disappears.

## Measured

Through `/v1/embeddings`, 256 real chunks, unsorted client input, same server, this patch the only variable:

| | embeddings/sec |
|---|---|
| main | 7.1 |
| this branch | **9.4** (1.32x) |

It also makes `--embedding-batch-size` safe to raise, which today is a trap. Measured from a client that sorts its own input, a 256-input forward batch reaches 16.6/s against 13.5/s at 32 — but with unsorted input the same increase is a **regression**, 8.9 -> 7.0/s, because a larger mixed batch pads more rather than less. After this change the flag scales the way its name implies.

## Correctness

A wrong restore hands a document its neighbour's vector with nothing raising, so it is checked directly. Against a live server, each of the first, middle and last inputs embedded alone matches the same text embedded inside a mixed batch at **cosine 1.000000**.

`tests/test_embedding.py::TestEmbeddingEngine::test_engine_embed_groups_by_length_and_restores_caller_order` asserts both properties — that the engine embeds in non-decreasing length order, and that every embedding returns on its own input. It fails on `main` and passes here.

Full suite: **6934 passed, 74 skipped**.

### DIFF
--- a/omlx/engine/embedding.py
+++ b/omlx/engine/embedding.py
@@ -21,6 +21,15 @@ from .base import BaseNonStreamingEngine
 logger = logging.getLogger(__name__)
 
 
+def _input_length(item: Union[str, Dict[str, str]]) -> int:
+    """Approximate token cost of one embedding input, used only to group similar sizes."""
+    if isinstance(item, str):
+        return len(item)
+    if isinstance(item, dict):
+        return sum(len(v) for v in item.values() if isinstance(v, str))
+    return 0
+
+
 class EmbeddingEngine(BaseNonStreamingEngine):
     """
     Engine for generating text embeddings.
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, this is a clear win and the reasoning in the description matches what I measured. I reproduced it through `/v1/embeddings` with a 256 chunk corpus (median ~37 chars, p90 400, max 3000), and the gain is bigger than your numbers on every embedding model I have here:

| model | main | this branch |
|---|---|---|
| multilingual-e5-large-instruct | 44.4 | 205.0 emb/s |
| Qwen3-Embedding-4B-4bit-DWQ | 7.0 | 32.0 emb/s |
| Qwen3-VL-Embedding-2B-4bit | 17.8 | 71.0 emb/s |

A uniform length corpus stays flat (310.6 to 315.8 emb/s) so the sort itself costs nothing, peak memory is identical on both branches at 3.466 GB for a single request and for 6 concurrent ones, and every vector still lands on its own input: cross branch per index cosine is 1.000000 across mixed 256, a 38 input batch of one huge plus many tiny, single input, empty strings, dict items and tied keys. Reported token usage does not move either, and the batched vectors actually sit closer to the unpadded batch of 1 reference than main does, since there is less padding to accumulate noise in.

I'll merge this.

PR #2375, opened 2026-07-26
fix(admin): recognize reasoning-field variants in external benchmark streaming

fparrav:
## Summary

- Benchmarking an external OpenAI-compatible endpoint (Admin → Benchmarks → External) against a reasoning/thinking model whose provider emits `delta.reasoning` instead of `delta.reasoning_content` (e.g. Ollama serving `qwen3.6:35b-mlx`) produced unusable results: single-request `tg TPS` read `0.0` for every prompt length, and continuous-batching `tg TPS` read multi-million tok/s.
- Root cause: `ExternalAPIClient.stream_chat_completion()` only checked the hardcoded key `delta.get("reasoning_content")` for decode-timing purposes, even though the module already defines `_REASONING_FIELD_NAMES = ("reasoning_content", "reasoning", "analysis")` for exactly this kind of provider variance (it's just never wired into the streaming path). With no field match, and `content` empty for the whole response (the model spends its `max_tokens` budget on reasoning), `first_content_time`/`last_content_time` both collapse to `end_time`, giving `gen_duration == 0`. In the batch aggregation the same collapse divides through a `1e-9` floor, producing an unbounded, meaningless tok/s figure instead of an honest "unmeasured" result.
- Confirmed via a live raw-SSE capture (per-line timestamps) against a local Ollama endpoint that the server streams incrementally with realistic ~11ms inter-token gaps — this is a client-side detection bug, not a server limitation.

## Changes

- `omlx/admin/external_api.py`: `stream_chat_completion()` now checks all of `_REASONING_FIELD_NAMES`, not just `reasoning_content`.
- `omlx/admin/benchmark.py`: `_compute_single_metrics()` reports `tpot_ms`/`gen_tps` as `None` (not `0.0`) when generation timing is genuinely unmeasured; `_run_external_batch_test()` reports `tg_tps` as `None` instead of dividing through the `1e-9` floor when the aggregate generation window collapses to zero or less.
- `omlx/admin/static/js/dashboard.js` + `omlx/admin/templates/dashboard/_bench.html`: render `N/A` (via a shared `benchFmtNum()` helper) instead of crashing or showing a misleading `0.0 tok/s` when a metric is unmeasured; `benchGetSpeedup()` returns `null` instead of `0` in that case.
- Tests: new coverage for the `reasoning` field-name variant, the single-chunk-dump unmeasured case, and the batch near-zero-window guard (both the collapsed and the normal/measurable case, with a patched clock for determinism).

## Test plan

- [x] `pytest tests/test_external_api.py tests/test_benchmark.py -q` — 136/136 passed
- [x] Full repo suite (`pytest tests/`, excluding two files broken by an unrelated `huggingface_hub` API change) — 6758 passed, 15 pre-existing failures/errors, all in files this change doesn't touch or import (`mlx_vlm` restructuring, hardware-specific TurboQuant numerics), confirmed via disjoint `git diff --name-only` against `main`
- [x] Live re-run of the original repro against Ollama (`qwen3.6:35b-mlx`) through the actual fixed client code: single-request `gen_tps` now `114.8 tok/s` / `tpot_ms=8.78` (previously `0.0`/`0.0`); batch(2x) `tg_tps=177.4 tok/s` (previously multi-million tok/s)

Developed with the help of Claude Code; investigated and verified against a live Ollama endpoint before opening this PR.

### DIFF
--- a/omlx/admin/benchmark.py
+++ b/omlx/admin/benchmark.py
@@ -252,18 +252,21 @@ def _compute_single_metrics(
 
     ttft_ms = ttft_s * 1000
     if generation_measured and completion_tokens > 1 and gen_duration > 0:
-        tpot_ms = (gen_duration / (completion_tokens - 1)) * 1000
-        gen_tps = completion_tokens / gen_duration
+        tpot_ms: float | None = (gen_duration / (completion_tokens - 1)) * 1000
+        gen_tps: float | None = completion_tokens / gen_duration
     else:
-        tpot_ms = 0.0
-        gen_tps = 0.0
+        # Generation timing could not be measured (e.g. all content arrived
+        # in a single burst with no measurable inter-token span) — report
+        # unmeasured rather than a misleading 0.0.
+        tpot_ms = None
+        gen_tps = None
     processing_tps = prompt_tokens / max(prefill_duration, 1e-9)
     total_throughput = (prompt_tokens + completion_tokens) / max(e2e_duration, 1e-9)
 
     return {
         "ttft_ms": round(ttft_ms, 1),
-        "tpot_ms": round(tpot_ms, 2),
-        "gen_tps": round(gen_tps, 1),
+        "tpot_ms": round(tpot_ms, 2) if tpot_ms is not None else None,
+        "gen_tps": round(gen_tps, 1) if gen_tps is not None else None,
         "processing_tps": round(processing_tps, 1),
         "e2e_latency_s": round(e2e_duration, 3),
         "total_throughput": round(total_throughput, 1),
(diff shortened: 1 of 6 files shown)


jundot:
Thanks, the root cause is exactly right. `_REASONING_FIELD_NAMES` was already sitting in the same module and only the streaming path never used it, so wiring it in is the correct fix.

One note on the batch guard: `gen_window > 0` doesn't fire on the real collapse. `end_time` is sampled inside each request coroutine while `wall_end` is sampled after `asyncio.gather`, so the window is always a small positive number rather than zero or less. I reproduced it on your branch with stats shaped the way `stream_chat_completion` builds them when no field matches, and `tg_tps` still came out at 3.16M tok/s. The multi-million figure going away in your live re-run comes from the field-name fix itself.

I'll merge this and handle the guard in a follow-up, keyed on whether content was actually observed rather than on the sign of the window, along with the same treatment for ttft and pp which still read as confident numbers in that state.

PR #2377, opened 2026-07-26
Fix/2367 top level enable thinking

Khambampati-Subhash:


jundot:
Thanks for the contribution. oMLX already supports `enable_thinking` through `chat_template_kwargs`, and a top-level field is not part of the OpenAI chat completions schema. Supporting it would add compatibility with a non-standard extension used by another serving engine rather than fix broken oMLX behavior, and I do not think oMLX needs to adopt that extension. I’ll close this PR.

PR #2379, opened 2026-07-26
Overhauled Chat UI

PowerSpy:
Model selector capsule, context status line with per chat token caching, chat search, configurable image limits, better scroll lock, jump to latest button, keyboard shortcuts help, and a Chat tab (Allow SVG / attachments / behavior). Rename left-footer Chat Settings to Settings and localize all new strings across 9 locales. Fixed Enter to send on narrow windows by rewriting isMobile check.


jundot:
Thanks for putting this together. This is a polished and thoughtful overhaul. I especially like the model selector; I think it is exactly what the current chat UI needs, and it makes switching models feel much more natural. The context meter, search, attachment controls, and improved scroll behavior are also excellent additions. I found a few contained follow-up issues around regeneration overrides, listener lifecycle, pinned ordering, localization, and the generated Tailwind artifact. I’ll address those immediately on `main`. LGTM for the feature direction. I'll merge this now and follow up.

PowerSpy:
Glad you liked it, I am happy to help on any of those if you would like.

Issue #1438, opened 2026-05-27
0.3.12 still not available on homebrew


lbm1202:
--fetch-HEAD still goes to 0.3.10

jundot:
Sorry for the delay! Just bumped the Homebrew formula to 0.3.12 in 786dfe2. `brew update && brew upgrade omlx` should pick it up now.

PR #2390, opened 2026-07-28
Deterministic prefill admission and pressure abort fixes

jundot:
On memory-tight machines a long prompt could pass admission, crawl through prefill for a minute or two, and then get killed at the memory watermark anyway. This PR makes admission catch those prompts up front with an instant 400, and stops the enforcer from killing requests while there are still gigabytes of reclaimable buffer pool to free.

Changes:
- Mid-prefill memory aborts return a typed 400 that names the binding ceiling and how to raise it, instead of a bare 500.
- Throttled prefill chunks snap to multiples of the min-chunk floor so the MLX buffer pool can actually reuse them. +6.8% prefill on gemma-4 and +8.4% on Qwen3.6 under a tight ceiling, neutral elsewhere.
- Admission now computes the real KV footprint per cache type (sliding window, GDN state, MLA) plus a measured transient bound, and checks it against the watermark where requests actually die. One formula for every preflight path.
- When the watermark is crossed mid-prefill but the buffer pool holds more than 2GB, the enforcer waits up to 5 seconds for a pool drain instead of aborting immediately. In the repro that used to kill a request 0.1GB over the line with 3.7GB of idle pool, the request now completes.

Verified before/after on gemma-4, Qwen3.6, Laguna and DeepSeek-V4-Flash under custom ceilings, faked small-machine RAM, and a real iogpu.wired_limit_mb. Highlights: gemma mid-prefill aborts 8 to 0 with max completed context up from 48k to 55k+ words, DSv4 enforcer kills 6 to 0 with the completed boundary up 15%. Roomy setups are unchanged and the throttle's chunk sizing policy is untouched. 488 related unit tests pass.

### DIFF
--- a/omlx/engine_core.py
+++ b/omlx/engine_core.py
@@ -35,7 +35,11 @@ from typing import (
 
 import mlx.core as mx
 
-from .exceptions import PrefillMemoryExceededError, describe_ceiling_binding
+from .exceptions import (
+    PrefillMemoryAbortedError,
+    PrefillMemoryExceededError,
+    describe_ceiling_binding,
+)
 from .model_registry import get_registry
 from .output_collector import RequestOutputCollector, RequestStreamState
 from .request import Request, RequestOutput, SamplingParams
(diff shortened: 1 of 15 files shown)


gilbert-barajas:
Follow-up from the #2369 discussion — you were right about the physical signal, and I have a data point from the opposite direction that may matter for this PR's admission formula.

I've been experimenting with file-backed weight loading on the batched engine (mmap, `PROT_READ`, llama.cpp-style: shards wrapped zero-copy via `newBufferWithBytesNoCopy`, wired limit held at 0 so weight pages stay clean and OS-evictable). It works — a 96.5 GB `mlx-community/DeepSeek-V4-Flash-2bit-DQ` loads in ~9s and serves with the process `phys_footprint` at ~0.3 GB, since clean file-backed pages don't count against it.

That same property makes the memory guard blind. With the machine genuinely at ~1 GB free (96+ GB of weight pages resident but reclaimable), a 200-token request aborted with:

```
Request aborted: process memory limit exceeded
(usage 0.3 GB, abort threshold (hard watermark) 68.7 GB, ceiling 72.3 GB)
```

The abort fired while reporting usage at 0.4% of the threshold — the enforcer's `dphys` signal and machine reality disagreed by ~96 GB. This was on a pre-#2390 tree (branched from `main` before it landed), so I can't yet say how the new admission path behaves — but it reads the same physical signal, so an elastic-loaded model would look like an empty process to admission as well, and KV would get admitted against memory the weights are actually holding.

Not urgent — nothing in-tree loads weights this way today. But if mmap-backed loading ever lands (mlx-lm #1438 asks for exactly this), the guard probably needs a notion of "resident but evictable" bytes. Happy to share the repro and measurement harness if useful.

PR #2397, opened 2026-07-28
feat: add durable idempotency lookup for Responses API

andyelka-creator:
## Summary

- accept an optional `Idempotency-Key` on non-stream `/v1/responses` requests
- persist a request-fingerprint-bound claim before engine acquisition
- replay the same completed or failed response without another generation
- expose `GET /v1/responses/idempotency` for authoritative lookup
- return a conflict when one key is reused for a different request

## Why

The Responses API currently assigns a server-generated response ID. If
generation completes but the client loses the HTTP response, the client never
learns that ID and cannot distinguish completion from final absence. Retrying
can therefore duplicate expensive or effectful work.

This change binds a client attempt key to durable response state before
generation begins. The raw key is not persisted; only its SHA-256 digest and
the request fingerprint are stored.

## Recovery behavior

- an existing matching claim replays its canonical response
- a mismatched request fingerprint returns HTTP 409
- a persisted `in_progress` claim becomes terminal `failed` after restart
- a missing claim returns authoritative `404` with `final: true`
- idempotent claims cannot be evicted or deleted
- streaming and non-persistent idempotent requests fail closed

## Validation

- `151 passed`:
  - `tests/test_responses.py`
  - `tests/integration/test_server_endpoints.py`
- `compileall` passed
- Ruff fatal rules (`E9,F63,F7,F82`) passed
- `git diff --check` passed

The patch was also exercised locally on Apple Silicon with
`Qwen2.5-7B-Instruct-4bit`:

- duplicate POST and lookup returned the same completed response with one
  generation
- reuse with a different request returned HTTP 409
- a forced process kill after the durable claim recovered as terminal
  `failed/server_restarted` without another generation
- an already-completed response survived the same restart and replayed without
  another generation
- a synthetic client-side lost response was recovered through the lookup
  endpoint without resubmission

## Current limits

- idempotency is intentionally limited to `stream=false`
- claim retention is unbounded until a tombstone/retention policy is defined
- the implementation relies on the current single Uvicorn worker model;
  multi-worker ownership is not claimed

### DIFF
--- a/omlx/api/responses_utils.py
+++ b/omlx/api/responses_utils.py
@@ -2,6 +2,7 @@
 """Conversion utilities for the OpenAI Responses API."""
 
 import copy
+import hashlib
 import json
 import logging
 import uuid
@@ -32,6 +33,10 @@ class ResponseStateNotFoundError(ResponseStateError):
 
 class ResponseStateCorruptError(ResponseStateError):
     """Raised when a stored response chain is incomplete or invalid."""
+
+
+class ResponseIdempotencyConflictError(ResponseStateError):
+    """Raised when an idempotency key is reused for a different request."""
 
 
 def _try_parse_json(s: str):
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for putting this together and for the detailed validation. I understand the lost-response case this is trying to cover, but oMLX already sends keepalives and cancels generation when the client disconnects, so the remaining ambiguity is limited to the narrow window after generation completes and before the final JSON is delivered. The Responses path also does not execute effectful tool calls itself, which makes the practical cost of a retry mostly duplicate inference.

I do not think the durable claim state, custom lookup endpoint, and indefinite retention are a good tradeoff for the general server without a concrete integration requirement or recurring user report. The current retention behavior can also push regular responses out of `ResponseStore` while idempotent claims grow without a bound. If there is a specific client depending on this guarantee, could you share that workflow and explain why the existing response storage or a bounded short-lived approach would not be sufficient? Otherwise, I am inclined not to merge this.

PR #2400, opened 2026-07-28
fix(claude-code): stop inflating reported token usage, gate auto-compact by real window

fparrav:
## Summary

- `scale_anthropic_tokens()` multiplied every reported token count (`usage.input_tokens`, streaming `message_start`, `message_delta`, OpenAI-compat usage) toward an arbitrary `claude_code.target_context_size`, while `CLAUDE_CODE_AUTO_COMPACT_WINDOW` continued reporting the model's real, un-scaled `max_context_window`. This "worked" only by coincidence while Claude Code's own default context assumption for unrecognized models was ~200k and `target_context_size` was set to roughly the same value.
- Once Claude Code raised its default assumption to 1M for unrecognized/custom models, operators raising `target_context_size` to match (e.g. 1,000,000) activated real 5-10x inflation of the reported numerator while the denominator (`CLAUDE_CODE_AUTO_COMPACT_WINDOW`) stayed at the real window. Auto-compact then either fires far too early (compaction loop on a mostly-empty context) or, when it doesn't win the race, lets the real prompt exceed the real context window before Claude Code intervenes, and the server hard-rejects it with HTTP 400 ("Prompt too long").
- Reproduced live against a running server: Claude Code's own status bar reported 556k/1000k tokens for a session whose oMLX server log showed a maximum real context of 147,599 tokens all day (model `max_context_window=262144`) — confirming the numerator/denominator drift in practice, not just in theory.

## What changes

- Removes `scale_anthropic_tokens()` and its ~8 call sites in `server.py`. Every token count oMLX reports to Claude Code over the Anthropic-compatible API now reflects the real token count processed by the model — no scaling, ever.
- **BREAKING**: `claude_code.context_scaling_enabled` + `claude_code.target_context_size` are removed and replaced by a single `claude_code.autocompact_threshold_pct` setting (int, default `80`, valid range 50-95). No migration shim — there is no sound mathematical mapping from an arbitrary token target to a percentage; old keys are silently dropped on next settings load and the new field takes its default.
- `CLAUDE_CODE_AUTO_COMPACT_WINDOW` keeps being set from the model's real, operator-configured `max_context_window` exactly as before (unchanged).
- `omlx launch claude` now additionally sets `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` from the configured threshold, so auto-compact fires at a deterministic percentage of the real window (e.g. `max_context_window=128000` + default 80% → compacts at ~102k tokens) instead of an unrelated inflated target.
- Admin UI: the "Escalado de contexto para Claude Code" toggle + target-tokens input is replaced by a single "Umbral de auto-compact (%)" field, translated across all 9 locales.

## Risk

`CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` is an internal/undocumented Claude Code CLI env var (confirmed present via `strings` on the installed 2.1.220 binary as `testPctOverride`, not part of the documented public surface). Worst-case failure mode if Anthropic removes it in a future release: `CLAUDE_CODE_AUTO_COMPACT_WINDOW` (the documented, stable env var) still gets sent correctly, so auto-compact falls back to firing at Claude Code's own internal default percentage of the real window — strictly better than today's broken state, not worse.

## Test plan

- [x] `tests/test_context_scaling.py` rewritten: confirms `scale_anthropic_tokens` no longer exists and the new settings field is in place
- [x] `tests/test_integrations.py`: new tests for `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` (default 80%, custom %, and omitted-when-unconfigured cases)
- [x] `tests/test_settings.py`: legacy-key tolerance (`context_scaling_enabled`/`target_context_size` in an old `settings.json` load without error and default to 80%), plus threshold range validation (50-95)
- [x] Full suite run locally: 6980 passed, 0 regressions (25 pre-existing failures on `main` unrelated to this change — `huggingface_hub`/`mlx_vlm` version mismatches in the local dev env, confirmed via `git stash`)
- [x] **Manual verification done** (see comment below): confirmed live that `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` must be a `0-1` fraction (a literal `0-100` integer is silently ignored). Found that the originally-proposed 85% let Claude Code's own thrashing detector permanently disable auto-compact under an adversarial workload, causing repeated hard 400s; lowered the default to 80%, which eliminated that failure mode in the same test (36 compactions / 28 turns, 0 errors). One caveat remains open — see the verification comment for details.


fparrav:
## Manual verification — done

Verified live against a patched build of the installed app, using headless `claude -p --continue` loops against a local model, inspecting real session transcripts (`~/.claude/projects/.../*.jsonl`) for `compact_boundary` events rather than trusting printed summaries.

**Env var format confirmed**: `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` as a fraction (`"0.8"`) produces real, reproducible `compact_boundary` events. A literal integer (`"85"`) is silently ignored — a run with `="85"` produced a byte-identical usage sequence to a run with no override at all, both growing unbounded to a hard 400. Confirms the fraction format already in this PR is correct.

**Default lowered 85% → 80%** after finding a real failure mode: under an adversarial workload (large single-turn content injections simulating a big tool-result/file-read), 85% let Claude Code's own thrashing detector trip after ~11 rapid compactions and permanently disable further auto-compact — 17+ consecutive `400 Prompt too long` errors, unrecoverable for the rest of the session. The same workload at 80% compacted 36 times over 28 turns with zero 400s; thrashing warnings still fired 3 times but auto-compact recovered each time instead of giving up.

**Caveat still open**: under that adversarial workload, the actual trigger point (~11-19% of the configured window) doesn't track the configured percentage precisely — a separate large-insert-specific heuristic in Claude Code appears to dominate over the smooth percentage-based trigger in that scenario. The fix reliably prevents the permanent-lockup failure mode (confirmed at 80%, not at 85%); it doesn't give exact control over the trigger point when a single turn injects a very large block of content. Worth re-testing against a steadier multi-turn agentic workload before treating the percentage as precise.

Claude Code CLI version verified against: **2.1.220**.

fparrav:
## Correction: pct-override format was wrong in the earlier verification

The earlier verification comment on this PR reported `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` as a `0-1` fraction (e.g. `"0.8"` for 80%). That was incorrect and has now been fixed (587d6829).

Live testing against a real coding session (not a synthetic benchmark) showed auto-compact firing at ~11-13% of the configured window instead of 80%. Decompiling the installed CLI (2.1.220) confirmed the root cause: the override is parsed as a plain `0-100` percentage (`n > 0 && n <= 100`, then divided by `/100` internally). `"0.8"` passes that validation (since `0.8 <= 100`) but is interpreted as **0.8%**, not 80%.

Fix: send the raw integer percentage (`"80"`) instead of the fraction (`"0.8"`).

Re-verified live after the fix: a fresh session against a 262,144-token model window ran well past the point where it previously compacted prematurely (39.9k tokens, no compaction, `Auto-compact window: 262.1k` correctly reported) — consistent with the corrected threshold rather than the broken one.

jundot:
Thanks for digging into the token-count mismatch and validating the percentage format. I took another pass through the underlying Claude Code behavior, and I think the integration may now have a cleaner option than replacing context scaling with a persistent percentage control.

The source snapshot recovered from the published npm sourcemap explains why oMLX needed scaling in the first place. In that snapshot, `getContextWindowForModel()` only honored `CLAUDE_CODE_MAX_CONTEXT_TOKENS` when `USER_TYPE === "ant"`, so an external local provider could not tell Claude Code that a Qwen or GLM model had a 32K or 64K context window. Scaling the reported usage toward Claude Code's assumed window was therefore a reasonable compatibility workaround at the time.

I then cross-checked the installed Claude Code 2.1.220 arm64 binary. The current external build now reads `CLAUDE_CODE_MAX_CONTEXT_TOKENS` for custom model IDs that do not canonicalize to `claude-*`, and auto-compact derives its window from `min(detectedContextWindow, CLAUDE_CODE_AUTO_COMPACT_WINDOW)`. Although `CLAUDE_CODE_AUTO_COMPACT_WINDOW` is now floored to 100K, setting `CLAUDE_CODE_MAX_CONTEXT_TOKENS=32768` makes the detected window 32K, so the final window remains `min(32K, 100K) = 32K`. That lets real reported usage and Claude Code's denominator use the same token scale without inflating any API usage values.

Given that newer behavior, I wonder if the cleaner fix is to set `CLAUDE_CODE_MAX_CONTEXT_TOKENS` from oMLX's resolved `max_context_window`, keep `CLAUDE_CODE_AUTO_COMPACT_WINDOW` as an explicit cap, remove token scaling, and otherwise leave Claude Code's own auto-compact policy alone. `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` is still represented internally as `testPctOverride`, and the recovered source describes it as an override for easier testing, so exposing it as a persistent user setting seems more tightly coupled to Claude Code internals than necessary. The remaining edge is that 2.1.220 ignores `MAX_CONTEXT_TOKENS` for canonical `claude-*` model names, so I would want 32K and 64K custom-model tests plus a `claude-*` alias check. What do you think about simplifying the PR around this newer context override instead of introducing the percentage setting?

fparrav:
Agreed — simplified the PR around `CLAUDE_CODE_MAX_CONTEXT_TOKENS` as you suggested (just pushed).

Changes:
- `autocompact_threshold_pct` (setting, `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` env var, and the admin UI field) is removed entirely.
- `ClaudeCodeIntegration.launch()` now sets `CLAUDE_CODE_MAX_CONTEXT_TOKENS` from the same resolved `max_context_window` used for `CLAUDE_CODE_AUTO_COMPACT_WINDOW` — no percentage math, no dependence on the undocumented `testPctOverride` hook, Claude Code's own auto-compact policy is left alone.
- Added `ClaudeCodeIntegration.launch()` tests for a 32K custom model, a 64K custom model, and a model ID that canonicalizes to `claude-*` — the last one documents (doesn't try to work around) the fact that Claude Code ignores `CLAUDE_CODE_MAX_CONTEXT_TOKENS` for that case, since that's the CLI's own internal behavior.
- Full local suite: 6975 passed, same pre-existing environment-only failures as before this revision (none touch the changed files).

The core fix (removing `scale_anthropic_tokens()`) is unchanged from before.

jundot:
Thanks again for the update. I merged this and added a small maintainer follow-up in 4f701282 for one Claude Code 2.1.220 edge: `omlx launch claude` now rejects models below 48K because Claude Code's output reserve and auto-compact buffer make smaller windows unsafe.

The guard covers the interactive picker, automatic single-model selection, `--model`, and explicit Opus/Sonnet/Haiku tier flags. No action is needed on your side.

PR #2401, opened 2026-07-29
fix: expose xgrammar binding to tvm-ffi fallback

jin-mitori:
## Summary
- add a relative loader fallback symlink for `libxgrammar_bindings.dylib` under `tvm_ffi/lib`
- retain the existing `RECORD` and rpath fixes
- avoid structured-output regressions when Homebrew cleanup removes Python wheel metadata

## Validation
- `ruby -c Formula/omlx.rb`
- local Homebrew oMLX strict JSON Schema probe: `GrammarCompiler initialized` and a schema-valid response

## Root cause
`xgrammar` was installed, but its wheel `RECORD` was absent after Homebrew cleanup. `tvm_ffi` then skipped metadata discovery and searched only its own fallback lib directory, where the binding was not present.

### DIFF
--- a/Formula/omlx.rb
+++ b/Formula/omlx.rb
@@ -151,7 +151,9 @@ class Omlx < Formula
   # The 0.1.32+ wheel ships libxgrammar_bindings.dylib with
   # @rpath/libtvm_ffi.dylib but no LC_RPATH pointing at where tvm_ffi
   # installs its native lib, and the dist-info is missing a RECORD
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for adding this fallback. I verified that apache-tvm-ffi falls back to `tvm_ffi/lib` when `RECORD` is unavailable, and the relative symlink resolves correctly without changing the normal metadata-based path. The runtime change looks sound; I'll wait for the draft to be marked ready before taking merge action.

PR #2409, opened 2026-07-29
fix: credit resident prefix-cache tokens in the prefill memory guard preflight

fparrav:
## Summary

- The prefill memory guard's preflight check (`BatchedEngine`/`VLMBatchedEngine` `preflight_chat`/`preflight_completion` -> `Scheduler.preflight_or_raise`) always estimated peak memory assuming `cached_tokens=0`, regardless of how much of the prompt is already resident in the scheduler's paged prefix cache.
- Confirmed live against a real long-running session: a request with 84,154 prompt tokens and an 81,920-token (97%) paged-cache hit completed cheaply and successfully. The very next turn — a near-superset prompt — was rejected by preflight with `cached=0`, estimating a peak as if the entire ~85k-token prompt were new. Every rejection in that session's logs showed `cached=0`, even moments after a >90% cache hit had just been served.
- This makes long, cache-friendly agentic conversations (the common Claude Code / coding-agent pattern: each turn resends the full growing transcript) progressively more likely to hit false-positive `400`s as they grow, even though the model itself never needed to reprocess that much.

## What changes

- New `Scheduler.estimate_cached_prefix_length()`, factored out of the existing `_log_prefix_divergence` diagnostic (which already computes an exact token-level common-prefix match against the scheduler's last 4 stored reference sequences via `_common_prefix_len`). No new expensive lookup — reuses the same cheap comparison, now consumed by the memory estimate itself instead of only a log line.
- `BatchedEngine`/`VLMBatchedEngine` preflight now keep the tokenizer's encoded token list (previously discarded after taking `len()`) and pass `cached_tokens=scheduler.estimate_cached_prefix_length(encoded)` through to `preflight_eviction_request`/`preflight_or_raise`.
- `DFlashEngine` preflight is unchanged: its cache hit reconstructs (clones) the full prompt's KV into active memory, so charging the full prompt there remains correct. Cross-reference comments added at both sites explaining the difference.
- The estimate stays conservative: an exact token-level prefix match, never a heuristic or extrapolation. A genuine cold start (no probe history, or no shared prefix) still gets today's full-charge behavior — no regression there. The existing mid-generation safety net (`_preflight_safety_rejection`) is untouched and still catches any overshoot during actual execution.

## Test plan

- [x] New scheduler unit tests (`tests/test_prefix_divergence_probe.py`): `estimate_cached_prefix_length` returns the exact common-prefix length, 0 with no probe history, 0 with no shared prefix, picks the best of multiple stored sequences, never rounds up past the exact match.
- [x] New engine preflight tests (`tests/test_engine_preflight.py`): a prompt sharing a long prefix with a stored reference sequence has its peak estimate computed from new tokens only (both `BatchedEngine` and `VLMBatchedEngine`); a prompt with no matching stored sequence produces the same estimate as before this change (cold-start regression guard).
- [x] Existing `DFlashEngine` preflight tests (`tests/test_dflash_prefill_memory_guard.py`) already assert `cached_tokens` is never passed — unaffected, confirming no behavior change there.
- [x] Full suite: 7278 passed, 1 failed (pre-existing, environment-only `torch_stub` subprocess test — reproduced identically via `git stash`, unrelated to this change), 53 skipped.
- [x] Manual verification: reproduced the real incident's shape (81,920-token resident prefix, 84,995-token next turn, real `current`/`ceiling` figures from the incident's server log) directly against the patched `Scheduler.preflight_or_raise` — the previously-rejected pattern is now admitted; the same request with the cache credit forced to 0 (pre-fix behavior) still rejects, reproducing the incident; a genuine cold start of equivalent size still rejects correctly (guard not weakened).

### DIFF
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -84,9 +84,11 @@ class BatchedEngine(BaseEngine):
         *,
         num_prompt_tokens: int,
         request_id: str | None,
+        cached_tokens: int = 0,
     ) -> None:
         eviction_request = scheduler.preflight_eviction_request(
             num_prompt_tokens=num_prompt_tokens,
+            cached_tokens=cached_tokens,
             request_id=request_id,
         )
         if eviction_request is not None and self._prefill_eviction_callback is not None:
@@ -97,6 +99,7 @@ class BatchedEngine(BaseEngine):
             await self._prefill_eviction_callback(eviction_request)
         scheduler.preflight_or_raise(
             num_prompt_tokens=num_prompt_tokens,
+            cached_tokens=cached_tokens,
             request_id=request_id,
         )
 
@@ -971,7 +974,8 @@ class BatchedEngine(BaseEngine):
         # through the existing handler chain so the response shape stays
         # consistent.
         try:
-            num_tokens = len(self._tokenizer.encode(prompt))
+            encoded = self._tokenizer.encode(prompt)
+            num_tokens = len(encoded)
         except Exception as e:
             logger.warning(
                 "BatchedEngine.preflight_chat: tokenizer.encode raised %s; "
@@ -984,8 +988,12 @@ class BatchedEngine(BaseEngine):
         if scheduler is None:
             _warn_scheduler_unreachable_once(self, "preflight_chat")
             return
+        cached_tokens = scheduler.estimate_cached_prefix_length(encoded)
         await self._preflight_or_raise_with_eviction(
-            scheduler, num_prompt_tokens=num_tokens, request_id=request_id
+            scheduler,
+            num_prompt_tokens=num_tokens,
+            cached_tokens=cached_tokens,
+            request_id=request_id,
         )
 
     async def preflight_completion(
@@ -1001,7 +1009,8 @@ class BatchedEngine(BaseEngine):
         if not self._loaded:
             await self.start()
         try:
-            num_tokens = len(self._tokenizer.encode(prompt))
+            encoded = self._tokenizer.encode(prompt)
+            num_tokens = len(encoded)
         except Exception as e:
             logger.warning(
                 "BatchedEngine.preflight_completion: tokenizer.encode raised "
@@ -1014,8 +1023,12 @@ class BatchedEngine(BaseEngine):
         if scheduler is None:
             _warn_scheduler_unreachable_once(self, "preflight_completion")
             return
+        cached_tokens = scheduler.estimate_cached_prefix_length(encoded)
         await self._preflight_or_raise_with_eviction(
-            scheduler, num_prompt_tokens=num_tokens, request_id=request_id
+            scheduler,
+            num_prompt_tokens=num_tokens,
+            cached_tokens=cached_tokens,
+            request_id=request_id,
         )
 
     async def stream_chat(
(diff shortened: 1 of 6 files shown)


jundot:
Thanks for investigating this. I think the current preflight estimate is more accurate than it may appear because it estimates peak memory, not the total amount of prefill work.

A cache hit avoids recomputing the cached tokens, but the full KV still has to be restored from SSD into active memory. The current guard also does not charge transient memory for the entire prompt at once; it charges only the minimum prefill chunk at the full context length. If the uncached suffix is at least one minimum chunk, its peak transient is therefore similar to fresh prefill even with a large cache hit.

Cache credit may reduce the estimate slightly for an exact hit or a suffix shorter than the minimum chunk, but subtracting the entire cached prefix from the resident-KV estimate undercounts the much larger KV allocation required during restoration. The restoration and concatenation overhead is not included either.

For that reason, I think keeping the current conservative behavior is safer than merging this approach. We can revisit a narrower adjustment for very short or exact-hit suffixes if we can model the KV restoration cost separately.

fparrav:
Thanks for the detailed pushback — it prompted me to actually measure the restoration cost instead of arguing the theory further.

I instrumented `PagedSSDCacheManager` (both `load_block()`'s reconstruction branches and `preload_matched_blocks()`, which turned out to be the actual per-request cold-restore call site — it runs from `BlockAwarePrefixCache.preload_blocks()` on every request with a matched block table, not just as a model-load-time warm start like I first assumed) to record real `phys_footprint()` deltas, gated behind a debug flag with zero overhead when off.

Ran it against the live `Jundot--Qwen3.6-35B-A3B-oQ6-mtp` model. Two results:

1. **Restoring a block after eviction (e.g. across a restart) has a real, non-negligible cost.** Cleanest samples (least thread-contention overlap across the preload pool): ~116-128 KB/token. The model's own steady-state resident cost is ~78 KB/token (from `MemoryMonitor`'s "estimated memory per 64-token block"). So restoration runs roughly 1.5-1.6x a block's own footprint — real, but well short of a full fresh prefill (no attention compute, no MTP draft/verify). This confirms your objection: unconditionally subtracting the whole cached prefix from `kv_exact` undercounts real memory for this case.

2. **Reuse within the same process, before eviction, is genuinely free.** Same-session multi-turn tests (blocks still in `PagedSSDCacheManager`'s hot cache from `save_block()`, never evicted) recorded zero restoration-cost samples across repeated runs — no reconstruction, no serialization, matching your own caveat about "an exact hit... may be safe."

So I think the right shape is closer to what you suggested — a narrower credit for the case where restoration is verifiably free, not a heuristic content match. I've scoped a follow-up that credits `cached_tokens` in the early preflight only for the contiguous prefix whose blocks are confirmed present in `PagedSSDCacheManager`'s hot cache at check time (a cheap, read-only membership check, no LRU/budget side effects) — zero credit for anything that would need real SSD restoration, even if it's an exact content match. I'll open that as a separate PR referencing this one.

Given that, I think this PR should stay closed/unmerged as originally proposed — happy to close it once the narrower version is up, or you're welcome to close it now if you'd rather not wait.

fparrav:
Closing in favor of #2427, which implements the narrower, hot-cache-verified version of this credit — see the discussion above and that PR's description for the full rationale and measurements.

PR #2411, opened 2026-07-30
fix(server): include model name in per-request log summaries

### LINKED ISSUE #99: Support better understanding of subagents/teams
Not quite sure how to suggest a solution for this, but here is the situation:

claude code config:
```
**    "env": {
        "ANTHROPIC_BASE_URL": "http://192.168.200.83:8666",
        "ANTHROPIC_AUTH_TOKEN": "XXX”,
        "ANTHROPIC_MODEL": "Qwen3.5-35B-A3B-4bit",
        "ANTHROPIC_DEFAULT_HAIKU_MODEL": "Qwen3.5-4B-MLX-4bit",
        "ANTHROPIC_DEFAULT_SONNET_MODEL": "Qwen3-Coder-30B-A3B-Instruct-MLX-5bit",
        "ANTHROPIC_DEFAULT_OPUS_MODEL": "Qwen3.5-35B-A3B-4bit",        
        "API_TIMEOUT_MS": "3000000",
        "CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC": "1",
        "ENABLE_LSP_TOOL": "1",
        "CLAUDE_CODE_ATTRIBUTION_HEADER":"0"
    },
```

run claude code 
```claude --settings settings-omlx.json  —worktree```

claude starts with default model (35b-a3b)

tell claude to spin up a team of haiku agents and implement something as a test
```create a team of 2 haiku agents to implement a simple calculator command line app. one in python and one in iavascript. The calculator only needs to support addition. Each version should be in its own subdirectory and have 5 unit tests and simple documentation```

claude runs and spins up subagents, but 
1) I don’t know what models are being used in the ui. I can only flip back and forth between them in the status page dropdown and see that the 4b and 35b both had info for this session, but I don’t know what is running. the models page only ever showed the a3b one.
2) I don’t know if omlx loaded both sets of models at once (35B-a3b and 4b) or swapped them back and forth

Anyway, a weird usecase, but subagents and teams will be more common over time and it would be nice if there were a way to see what was running and know more about how they are loaded/unloaded.

Regardless, the experiment worked and omlx made that happen - lmstudio would have choked on this. :)

### PR DESCRIPTION
**Pull request by JimStenstrom**
Fixes #99.

## Summary

- Every per-request INFO summary now leads with `model=<resolved id>` — completion and chat completion (streaming and non-streaming), embedding, rerank, Anthropic messages, and Responses API.
- With several models loaded (the multi-model subagent scenario from #99), interleaved summary lines can now be attributed to the model that served them.

## Why

#99's live-view half was covered by the Active Models card (#233), but the log half you called out there ("the logs don't actually show model names right now either") is still open: the model appears only in the DEBUG-level request-received line, so at default log level a session hitting several models produces `Chat completion: 128 tokens in 0.89s ...` lines that can't be attributed to a model.

Before / after:

```
Chat completion: 128 tokens in 0.89s (143.3 tok/s), prompt: 1024, finish_reason=stop, ...
Chat completion: model=Qwen3.5-4B-MLX-4bit, 128 tokens in 0.89s (143.3 tok/s), prompt: 1024, finish_reason=stop, ...
```

## Changes

- `omlx/server.py`: eight `logger.info` summary sites gain a leading `model=` field (matching the existing `model=` style of the DEBUG request-received lines). No other text changes, no new log lines, no level changes.
- The value logged is the resolved model id, taken from the same expression the adjacent `record_request_complete(model_id=...)` call already used. Where that expression was inline (`resolve_model_id(request.model) or request.model`), it is hoisted into a local and reused by both, so the log line and the metric can never disagree and resolution still runs once per request.
- Sibling sweep: streaming + non-streaming covered on both OpenAI endpoints; embeddings and rerank covered; Anthropic messages and Responses API summaries covered. Explicitly out of scope (pre-existing): the *streaming* Anthropic-messages and *streaming* Responses paths emit no summary line at all today (metrics only) — adding brand-new INFO lines there is a separate behavioral change. Also pre-existing and unchanged: with `model_fallback` enabled (default off), the resolved id names the requested model even when the default served the request — this mirrors what the metrics field already records.

## Tests

- Three tests added to the existing classes (`TestStreamingHelperFunctions` in `tests/integration/test_e2e_streaming.py`, `TestChatCompletionEndpoint` in `tests/integration/test_server_endpoints.py`) asserting the summary line names the resolved model, including the alias→resolved case; all three run under the CI marker filter.
- `pytest tests/integration/test_e2e_streaming.py tests/integration/test_server_endpoints.py -q` → `223 passed, 51 deselected`.
- Full suite: `7270 passed, 52 skipped, 70 deselected`, plus 3 failures that also fail on unmodified `main` on this machine (M5 numeric-tolerance tests in `test_glm_mtp_patch.py` / `test_mtp_prompt_priming.py`, unrelated to this change — verified by re-running them on a clean tree).
- Also verified all three new tests fail when the fix is reverted.

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -2801,8 +2801,10 @@ async def create_embeddings(
             raise HTTPException(status_code=400, detail=str(e))
 
         elapsed = time.perf_counter() - start_time
+        resolved_model = resolve_model_id(request.model) or request.model
         logger.info(
-            f"Embedding: {len(embedding_inputs)} inputs, {output.dimensions} dims, "
+            f"Embedding: model={resolved_model}, "
+            f"{len(embedding_inputs)} inputs, {output.dimensions} dims, "
             f"{output.total_tokens} tokens, max_length={max_length}, "
             f"truncation={request.truncation} in {elapsed:.3f}s"
         )
@@ -2811,7 +2813,7 @@ async def create_embeddings(
             completion_tokens=0,
             cached_tokens=0,
             prefill_duration=elapsed,
-            model_id=resolve_model_id(request.model) or request.model,
+            model_id=resolved_model,
         )
 
         data = []
@@ -2926,8 +2928,9 @@ async def create_rerank(
         )
 
     elapsed = time.perf_counter() - start_time
+    resolved_model = resolve_model_id(request.model) or request.model
     logger.info(
-        f"Rerank: {len(documents_raw)} docs, "
+        f"Rerank: model={resolved_model}, {len(documents_raw)} docs, "
         f"{output.total_tokens} tokens in {elapsed:.3f}s"
     )
     get_server_metrics().record_request_complete(
@@ -2935,7 +2938,7 @@ async def create_rerank(
         completion_tokens=0,
         cached_tokens=0,
         prefill_duration=elapsed,
-        model_id=resolve_model_id(request.model) or request.model,
+        model_id=resolved_model,
     )
 
     # Format response - results sorted by score (descending). Strings wrap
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for adding model attribution across the existing summary logs. The resolved model handling is consistent across the covered paths, and the focused tests look good. One thing I will fold into a follow-up is adding summaries to the Anthropic and Responses streaming paths and using the actual leased model id when model fallback is active. LGTM, merging.

Issue #1416, opened 2026-05-26
MTP Usage on Gemma4 31B results in OOM Condition


miabchdave:
When MTP is enabled with Gemma4 31B BF16 (default 4 MTP  tokens), memory usage spikes to OOM area.  Generation sometimes completes but system freezes during it.  M5 Max MacBook with 128GB.

Using a simple prompt to test:  **Write a python script to sort strings.**

For Gemma4 31B:
- Using MLX-Community Gemma4 Assistant (MTP Layers)
- Using Gemma4 31B IT at BF16 converted from mlx-vlm 0.4.5 (HF miabchdave/gemma-4-31B-it-MLX-bf16)

1. Run prompt (~1200 tokens generated) without MTP:  Memory usage is ~61GB, Normal
2. Enable MTP on Gemma4 (Select Assistant, default token of 4)
3. Re-run Same Prompt with MTP:  Memory usage climbs to ~118GB (still generating only about 1200 tokens)

**Expected behavior**
Additional memory usage with MTP enabled should be minimal, similar to MTP off, especially for a small context prompt.

<img width="1678" height="772" alt="Image" src="https://github.com/user-attachments/assets/983f9c9a-4fbf-4fe5-bb80-666ded6dd605" />

<img width="1687" height="771" alt="Image" src="https://github.com/user-attachments/assets/6f8f123f-c3a5-4095-99a3-e7693904b323" />

**Desktop (please complete the following information):**
 - macOS Version: 26.5
 - oMLX Version: 0.3.10
 - Macbook Pro M5 Max 128GB

miabchdave:
Same exact issue with DFlash and Gemma4 31B.

jundot:
Reproduced the MTP variant with gemma-4-31b-it-4bit + gemma-4-31B-it-assistant. Root cause: mlx leaves the metal buffer pool growing between flushes (default flush every 256 tokens), and a dense 60-layer / hidden=5376 target fills that window with way more allocations than the 26B-A4B MoE does. Patched on the oMLX side by clearing the cache per MTP round. Peak drops from 77GB to 20GB on 31b-4bit + drafter in my run, the MTP gain holds (+52% over no-MTP on 31B), and tk/s loss is about 3%. Lands in the next release.

[...]

Issue #888, opened 2026-04-22
Chat stuck at prefilling 91%


ncy94:
**Describe the bug**
When loaded qwen3.6-35B-A3b-4bit, chat would get stuck at prefilling 91% with any prompt

**To Reproduce**
Not sure if it's reproduceable - I've only loaded the model and it's stuck. Tried turning off SpecPrefill, TurboQuant, delete KV cache, and restart omlx, nothing fixes it.  

**Expected behavior**
Model should function and not stuck.

**Screenshots**

<img width="1228" height="404" alt="Image" src="https://github.com/user-attachments/assets/b44c0730-152e-41a4-8c2d-05c8fdd48e7c" />

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: chrome
 - Version: Tahoe 26.1

**Smartphone (please complete the following information):**

**Additional context**
Found this in the server log:
```
2026-04-21 23:19:45,575 - omlx.engine_pool - INFO - [-] - Loaded model: Qwen3.6-35B-A3B-4bit (estimated: 19.95GB, total: 19.95GB)
2026-04-21 23:20:09,861 - omlx.engine.vlm - INFO - [-] - Structured output requires xgrammar. Reinstall with: brew reinstall omlx --with-grammar
2026-04-21 23:20:19,515 - omlx.scheduler - ERROR - [-] - Error in batch generation step: There is no Stream(gpu, 0) in current thread.
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/omlx/0.3.6/libexec/lib/python3.11/site-packages/omlx/scheduler.py", line 3734, in step
    scheduled, rejected = self._schedule_waiting()
                          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/omlx/0.3.6/libexec/lib/python3.11/site-packages/omlx/scheduler.py", line 3147, in _schedule_waiting
    prefilled_cache, last_token = self._do_external_prefill(
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/omlx/0.3.6/libexec/lib/python3.11/site-packages/omlx/scheduler.py", line 1311, in _do_external_prefill
    _sync_and_clear_cache()
  File "/opt/homebrew/Cellar/omlx/0.3.6/libexec/lib/python3.11/site-packages/omlx/scheduler.py", line 53, in _sync_and_clear_cache
    mx.synchronize(generation_stream)
[...]

PR #2413, opened 2026-07-30
fix(metal): drain the GPU stream before the hot-loop cache clears

JimStenstrom:
Refs: #300

## Summary

Two decode hot loops released Metal buffer-cache memory with a bare `mx.clear_cache()`, with no synchronization of any kind first — not even an eval. Both run inside `async_eval`-pipelined decode, so the clear can release buffers that an in-flight command buffer still references, which is the pattern `_sync_and_clear_cache()` exists to prevent (#300, #888). This routes both through that helper, each passing the stream its work actually rode, and moves the helper out of `scheduler.py` into `omlx/utils/metal_sync.py` so non-scheduler callers can reach it. Clear cadence is unchanged at both sites; no new settings, no behavior change for the existing call sites.

## Why

`omlx/patches/glm_moe_dsa/generate_patch.py` clears every 512 decode steps in the patched `BatchGenerator._next` (live for the GLM-5.2 family since #1984). The work being cleared over is mlx-lm's `GenerationBatch.next()`, which submits the step with `mx.async_eval`. The stream it rode is the generator's own `self._stream`: mlx-lm runs `_next()` only inside `with mx.stream(self._stream)` (`next()` / `next_generated()`), and oMLX constructs the generator with the per-engine stream (`scheduler.py`, `stream=self._stream`). The scheduler's own periodic clear on the same loop already does `_sync_and_clear_cache(self._stream)` every 1024 tokens — this makes the patched path match it.

`omlx/speculative/vlm_mtp.py` clears at every round-loop yield (added to bound the buffer-pool peak on large targets, #1416 — mlx-vlm itself only clears every 256 tokens). Two streams carry in-flight work at that boundary: mlx-vlm `async_eval`s the verify hidden state and the drafter's state arrays, dispatching the verify and rollback forwards inside `with mx.stream(generation_stream)` on `mlx_vlm.speculative.common.generation_stream`, while the rest of the round (draft block, acceptance walk, `set_shared_kv`) inherits the stream the scheduler advances the generator under — `with mx.stream(self._stream): next(state.generator)`. So the site passes mlx-vlm's stream explicitly, and the helper's second, no-argument `mx.synchronize()` drains the current default stream, which inside that `with` block is the engine stream. Both sets are covered with no new parameter, keeping stream context inherited rather than re-read.

Note that mlx's `ThreadLocalStream` resolves to a different concrete `mx.Stream` per touching thread, and a cross-thread sync of one is a silent no-op — the three streams involved here (mlx-lm's `generation_stream`, mlx-vlm's speculative `generation_stream`, and the per-engine stream) are distinct objects, so naming the right one matters. Both drains happen on the same thread that submitted the work, in the same call frame.

Cost: a micro-benchmark shaped like an MTP yield (heavy verify chain on mlx-vlm's stream, `.tolist()` force as `_speculative_walk` does, drafter tail `async_eval` on the engine stream, then the clear) puts the added drain at 0.11-0.19 ms per yield across two runs — under 1% of a real MTP token on the 31B-class targets that route here, because the verify graph is already forced by the acceptance walk and only the round's tail remains in flight. That is a micro-benchmark, not an e2e MTP run. One tradeoff worth naming: the helper takes `_mx_buffer_access_lock`, so with `paged_ssd_cache_dir` enabled the clear can now wait on the async store-cache worker — the same serialization the scheduler's per-chunk and per-1024-token clears already accept, and it is the point of the lock (#1106).

## Changes

- New `omlx/utils/metal_sync.py`: `_sync_and_clear_cache()`, `_mx_buffer_access_lock`, and the `_default_generation_stream` alias, moved verbatim from `scheduler.py` (docstrings and issue references unchanged).
- `omlx/scheduler.py` imports and re-exports the same three names, so all ~30 existing call sites, `engine_core.py`'s import, the injected `sync_and_clear_cache=` lambdas, and existing tests that patch `omlx.scheduler._sync_and_clear_cache` are untouched. Extraction is behavior-neutral.
- `omlx/patches/glm_moe_dsa/generate_patch.py`: the every-512-steps clear becomes `_sync_and_clear_cache(self._stream)`.
- `omlx/speculative/vlm_mtp.py`: both round-loop clears (single and batched) become `_sync_and_clear_cache(_vlm_generation_stream)`.

The extraction is deliberately minimal and does not convert any other ad-hoc `mx.clear_cache()` site. The remaining ones — `engine/base.py` `_finish_activity` and the ancillary embedding/reranker/tts/stt/sts engines, the `engine_pool.py` load/unload/reclaim paths, `dflash.py` eviction, `vlm.py` diffusion cleanup, `admin/oq_manager.py`, and the `oq.py` eval-then-clear sites — each need their own stream/thread analysis and belong in separate PRs; a shared module is what makes those migrations possible one at a time.

## Tests

`python -m pytest -m "not slow and not integration" -q` → `3 failed, 7303 passed, 52 skipped, 70 deselected`. The 3 failures are pre-existing on a clean checkout of the same base (`test_glm_mtp_patch.py` x2, `test_mtp_prompt_priming.py` x1 — M5 NAX gemm tolerance, ~1e-4 vs 2e-5/1e-4 thresholds, unrelated to this change).

New regression coverage in the existing per-module test files, driving the real code paths and asserting the stream the helper was called with:

- `tests/test_glm_moe_dsa_patch.py`: the periodic decode clear drains the generator's stream; and it still only fires on the 512-step cadence. The first drives the actually-installed patched `BatchGenerator._next`, restoring mlx-lm's originals afterwards so the process-wide monkey patch does not leak into other tests.
- `tests/test_vlm_mtp.py`: every token yield of the single-request round loop, and every row yield of the batched loop, is preceded by a synchronized clear naming mlx-vlm's stream (and not mlx-lm's); the wrapper's own `first_bonus` yield clears nothing.

RED without the fix: reverting just the two sites to `mx.clear_cache()` fails 3 of the 4 (`did not drain any stream` / `expected N synchronized clear(s), got []`). The cadence test passes either way by design — it pins the cadence the memory-bounding change chose, so a future edit cannot quietly turn the periodic clear into a per-step clear.

### DIFF
--- a/omlx/patches/glm_moe_dsa/generate_patch.py
+++ b/omlx/patches/glm_moe_dsa/generate_patch.py
@@ -17,6 +17,8 @@ from typing import Any
 
 import mlx.core as mx
 
+from ...utils.metal_sync import _sync_and_clear_cache
+
 logger = logging.getLogger(__name__)
 
 _APPLIED = False
@@ -193,7 +195,12 @@ def apply_glm_moe_dsa_generate_patch() -> bool:
             self._gen_tokens_counter += len(generation_responses)
             self._steps_counter += 1
             if self._steps_counter % 512 == 0:
-                mx.clear_cache()
+                # GenerationBatch.next() submits the decode step with
+                # mx.async_eval, so the periodic clear has to drain that work
+                # first. self._stream is the stream it rode: BatchGenerator
+                # runs _next() inside ``with mx.stream(self._stream)`` and
+                # oMLX constructs the generator with the per-engine stream.
+                _sync_and_clear_cache(self._stream)
 
         if len(self._generation_batch) >= self.completion_batch_size:
             return prompt_responses, generation_responses
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -34,7 +34,6 @@ from mlx_lm.generate import (
     GenerationBatch,
     PromptProcessingBatch,
     SequenceStateMachine,
-    generation_stream,
 )
 from mlx_lm.models.cache import (
     KVCache as _MLXKVCache,
@@ -63,13 +62,14 @@ from .speculative.vlm_mtp import VLMMTPDrafter, run_vlm_mtp_decode
 from .utils.fatal import FATAL_TEARDOWN_TIMEOUT_S, fatal_exit
 from .utils.generation_config import load_generation_config_token_ids
 from .utils.hardware import format_bytes
+from .utils.metal_sync import (
+    _default_generation_stream,
+    _mx_buffer_access_lock,
+    _sync_and_clear_cache,
+)
 from .utils.proc_memory import get_phys_footprint
 from .utils.sampling import make_sampler as omlx_make_sampler
 from .utils.tokenizer import create_streaming_detokenizer
-
-# Module-level alias so Scheduler.__init__ can fall back to mlx-lm's default
-# stream when no per-engine stream is provided.
-_default_generation_stream = generation_stream
 
 
 def _apply_suppress_token_ids(logits: Any, suppress_token_ids: tuple[int, ...]) -> Any:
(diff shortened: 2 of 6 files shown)


jundot:
Thanks for tracing and fixing both oMLX-owned clear sites. The stream selection and shared helper extraction look correct, and the focused regression tests pass. LGTM, merging.

PR #2414, opened 2026-07-30
fix: guard the prefill transient EWMA against single-sample outliers

fparrav:
## Summary

- `PrefillTransientTracker`'s EWMA (`bytes_per_token`, feeding the adaptive prefill throttle and the admission guard's transient charge) had no protection against a single noisy `phys_footprint()` sample, unlike `observed_max_bytes` which already has `floor_sample` gating + a 4GB absolute clamp for exactly this kind of outlier.
- Reproduced live (2026-07-29, Qwen3.6-35B-A3B, `~/.omlx/logs/server.log`): baseline samples ranged ~525-1867 KB/token (largest legitimate fluctuation ~1.7x the running EWMA), then a single `n=185` tail/residual prefill chunk measured 10497.1 KB/token — a ~13.6x jump off an EWMA of 773.3 KB/token — which poisoned the EWMA to 3690.5 KB/token in one update. That inflated value then bounded every later admission check for the rest of the process lifetime, independent of how accurately `cached_tokens` is credited elsewhere.

## What changes

- New `_EWMA_OUTLIER_RATIO = 8.0` constant, chosen above the largest observed legitimate fluctuation (~1.7x) and below the observed incident outlier (~13.6x).
- `PrefillTransientTracker.update()`: a sample whose `per_token` rate exceeds the current EWMA by more than this ratio is excluded from the EWMA blend. It still counts toward `samples` and still updates `last_delta_bytes`/`last_n_tokens`, so a genuine regime change stays visible via those raw fields — only the accumulated EWMA is protected.
- The very first sample after construction/`reset()` still seeds the EWMA unconditionally (no prior value to compare a ratio against).
- `observed_max_bytes` / `floor_sample` / `_OBSERVED_MAX_CLAMP_BYTES` semantics are untouched.

## Test plan

- [x] `tests/test_prefill_transient_tracker.py`: new `TestEwmaOutlierGuard` class — first-sample seeding is unaffected by extreme values, the exact incident outlier is rejected from the EWMA (encoded with the real recorded numbers), a legitimate ~1.7x fluctuation still updates the EWMA normally, and a rejected sample still counts toward `samples`.
- [x] Existing `TestObservedMax::test_outlier_above_clamp_rejected_not_clamped` updated: the same 5GiB/32-token outlier is also a >8x EWMA outlier, so it must not move `bytes_per_token` either (previously only asserted `observed_max_bytes` behavior).
- [x] Full local suite: 19/19 in this file pass; no other files import this module, so no other tests are affected.
- [x] Manual replay: fed the exact transcribed `(n_tokens, per_token)` sequence from the incident window through the patched tracker — it matches production's own EWMA step-for-step up to 773.3 KB/token, then the `n=185` outlier leaves it unchanged instead of jumping to 3690.5 KB/token.

### DIFF
--- a/omlx/prefill_transient_tracker.py
+++ b/omlx/prefill_transient_tracker.py
@@ -32,6 +32,18 @@ class PrefillTransientTracker:
     # genuinely recurring giant transient still reaches the guard through
     # the last-delta/EWMA terms of _predicted_chunk_transient.
     _OBSERVED_MAX_CLAMP_BYTES = 4 * 1024**3
+    # A sample whose per_token exceeds the current EWMA by more than this
+    # ratio is treated as measurement noise (a tail/residual prefill chunk,
+    # not a genuine cost-per-token regime change) and excluded from the EWMA
+    # blend. Chosen from a real incident (2026-07-29, Qwen3.6-35B-A3B):
+    # baseline samples ranged ~525-1867 KB/token (largest legitimate
+    # fluctuation ~1.7x the running EWMA) before a single n=185 tail chunk
+    # measured 10497.1 KB/token — a ~13.6x jump off an EWMA of 773.3 KB/token
+    # — and pushed the EWMA to 3690.5 KB/token in one update, poisoning every
+    # later admission check for the rest of the process lifetime. 8x sits
+    # above the largest observed legitimate fluctuation and below the
+    # observed outlier.
+    _EWMA_OUTLIER_RATIO = 8.0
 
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for isolating the noisy sample. The EWMA guard is a useful, low-risk fix, and the related memory-guard tests pass. Speed admission still consumes raw tail samples, so I'll handle that focused path in a follow-up on main rather than expanding this contributor patch. LGTM, merging.

PR #2416, opened 2026-07-30
fix(oq): load text-only VLM bases via mlx-lm

blightbow:
## Summary

Three oQ load paths choose between mlx-lm and mlx-vlm using
`_has_vision_subconfig` alone. A checkpoint whose vendored mlx-lm
implementation exposes only the text backbone still ships a `vision_config`, so
those paths hand it to mlx-vlm, where `get_model_and_args` cannot resolve the
model type, falls through to the `mlx_vlm.speculative.drafters.<type>` lookup,
and fails the load.

MiMo V2.5 is the current case.

## Why this is a routing bug rather than a missing model

The repository already encodes this distinction. `model_discovery.py` lists
`mimo_v2` in `MLX_LM_TEXT_ONLY_MODEL_TYPES`, with the comment that such
checkpoints should route directly "instead of deliberately failing an mlx-vlm
load and relying on the engine-pool fallback".

`_build_model_sanitizer` already consults that set (`oq.py:3069`). The
sensitivity, imatrix and proxy load paths did not, so the same checkpoint was
classified one way when building the sanitizer and another way when loading the
model for calibration.

## Change

Adds `_is_vlm_load(config)`, which is `_has_vision_subconfig(config)` minus the
model types mlx-lm serves text-only, and applies it at the three load sites:

- `_collect_imatrix`
- `_measure_sensitivity`
- `_measure_sensitivity_from_quantized_model`

`_has_vision_subconfig` keeps its existing meaning as the vision classifier and
is unchanged. The new predicate is specifically the load-routing decision, which
is why it lives next to the oQ load helpers rather than in `model_discovery`.

## Tests

`tests/test_mimo_v2_patch.py` covers the predicate directly and both routing
outcomes: a text-only multimodal base with a vision sub-config routes through
mlx-lm, and a genuine VLM still routes through mlx-vlm.

## Effect

Adding a family to `MLX_LM_TEXT_ONLY_MODEL_TYPES` now corrects oQ load routing
at the same time, rather than requiring a second, easily missed edit here.

---

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -127,6 +127,25 @@ def _validate_oq_dtype_for_model(config: dict, dtype: str) -> None:
         )
 
 
+def _is_vlm_load(config: dict) -> bool:
+    """VLM routing predicate for oQ's model-load helpers.
+
+    Mirrors the VLM decision in :func:`_build_model_sanitizer`, but for the
+    sensitivity / imatrix / proxy load paths: route a checkpoint through
+    mlx-vlm only when it carries a vision sub-config *and* its model_type is
+    not one mlx-lm serves text-only (e.g. ``mimo_v2``). Without the text-only
+    exclusion these paths send a text-only-supported multimodal base -- which
+    still ships a ``vision_config`` -- to mlx-vlm, where ``get_model_and_args``
+    cannot resolve the model_type and falls through to the
+    ``mlx_vlm.speculative.drafters.<type>`` lookup and fails.
+    """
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for fixing the oQ load routing. This keeps the calibration paths aligned with `MLX_LM_TEXT_ONLY_MODEL_TYPES`, so MiMo uses the patched mlx-lm model while genuine VLMs stay on mlx-vlm. The focused routing tests cover both sides. LGTM. I’ll merge this first.

PR #2417, opened 2026-07-30
fix(oq): budget FP8 calibration on the bf16 footprint and dequantize MiMo fused QKV per shard

blightbow:
## Summary

Two related fixes in the oQ streaming path, both found while quantizing
[XiaomiMiMo/MiMo-V2.5](https://huggingface.co/XiaomiMiMo/MiMo-V2.5) (310B MoE, native FP8) to oQ4 on a 512 GB M3 Ultra.

The first sizes the calibration budget on what calibration actually
materializes. The second makes the fused, tensor-parallel-sharded QKV
dequantize correctly instead of aborting the run or silently mixing shards.

With both applied, the model quantizes, loads, and generates.

---

## 1. Size the calibration budget on the dequantized view

`_calibration_memory_budget` compared on-disk checkpoint bytes against the live
memory limit. For a native FP8 source those bytes are roughly half of what
calibration materializes, because the weights are dequantized to bf16 when the
model loads.

For MiMo V2.5 that is about 293 GB on disk against a 587 GB resident footprint.
Against a 366 GB limit (75 percent of 488 GB Metal capacity) the budget
concluded the model fit, skipped the RAM-safe proxy, loaded the full model for
calibration, and the process was killed by the OS partway through.

The footprint now comes from `_logical_footprint_bytes`, which sums the lazy
index's logical view. That view already reports every tensor at its
post-dequantization shape and dtype, so it is exact rather than approximate:

- native fp8 weights are reported as bf16
- packed fp4 experts are reported at their unpacked column count, so one stored
  byte becomes two bf16 values
- scale companions that fold into their weight are not counted at all
- virtual tensors (see below) are counted at the shape they are produced in

A first version of this used a per-format constant instead, doubling the
on-disk size for `quant_method` in `{fp8, mxfp8}`. That was wrong in both
directions and is worth recording, because the failure it would have left
behind is the same one this PR exists to fix. DeepSeek-V4 style checkpoints
declare `quant_method: "fp8"` but store two fp4 values per byte, so they expand
**fourfold**, not twofold: the constant under-counted the largest family by
half and could still have let it be OOM-killed. It also doubled scale bytes,
which are consumed rather than expanded, and over-counted checkpoints that mix
fp8 experts with a bf16 backbone, forcing the expensive proxy path
unnecessarily.

Observed on the fixed run:

```
oQ4: calibration footprint (586.8 GB) exceeds 75% of calibration capacity
(488.0 GB; limit=366.0 GB, system available=489.9 GB, Metal available=488.0 GB).
Full-model calibration will use a proxy.
```

On-disk size is kept as a floor, so sidecars and anything the logical view
folds away cannot make the estimate too small.

**Cost worth stating:** fp8 sources that previously slipped under the budget
now take the proxy path, so the source is read and dequantized once for the
proxy and again for the main pass. That roughly doubles wall-clock for those
models. Caching the bf16 between passes is exactly the residency this change
exists to avoid, so the extra pass is the intended trade rather than an
oversight.

---

### The layout

MiMo V2.5 ships attention as a single fused `qkv_proj.weight` (F8_E4M3) with a
block-128 `weight_scale_inv`. The tensor is already sharded for tensor
parallelism, and **each shard is padded to a block boundary individually**, so
the padding is interleaved through the tensor rather than trailing it.

From the published `config.json` (`num_attention_heads` 64,
`num_key_value_heads` 4, `head_dim` 192, `v_head_dim` 128, `weight_block_size`
[128, 128]) at TP=4, one full-attention shard is:

```
16 * 192 + 1 * (192 + 128) = 3392 rows, padded to 3456
```

Across four shards that is 13568 rows with a (108, 32) scale, and 64 rows of
padding sit *inside* the tensor at each shard boundary.

Sliding-window layers use `swa_num_key_value_heads` 8, giving 3712 rows per
shard, which is already a multiple of 128. Those layers carry no padding at
all: 14848 rows with a (116, 32) scale. Since `hybrid_layer_pattern` marks 9 of
48 layers as full attention, **only 9 layers are padded and 39 are block
aligned**.

TP is also only resolvable from a full-attention layer. At TP=1 and TP=2 the
row count matches but the padded row count does not, so TP=4 is unique there.
On a sliding-window layer TP=1 matches just as well, because there is no
padding to disambiguate.

### What went wrong

`_block_dequant_fp8` infers a single block grid for the whole tensor. On the 9
padded layers no block size reconciles 13568 rows with 108 scale rows, so it
raises and the run stops. On shapes it does accept, one grid straddles shard
boundaries and applies the wrong scale to a large fraction of rows, with no
error raised.

Other runtimes have hit the same family of problem with this checkpoint:

- [vllm-project/vllm#42803](https://github.com/vllm-project/vllm/issues/42803):
  the fused `qkv_proj` load path uses a naive chunk per rank and misplaces Q
  values into the K and V slots at TP=8.
- [sgl-project/sglang#24607](https://github.com/sgl-project/sglang/issues/24607):
  fused `qkv_proj` shard shape mismatch when `tp_size` exceeds
  `num_key_value_heads`.

Both are naive handling of the same fused layout, one silent and one loud.

### Why the scale cannot be routed through sanitize

The vendored mimo_v2 model already does this correctly in `Model.sanitize`
(ported from [ml-explore/mlx-lm#1219](https://github.com/ml-explore/mlx-lm/pull/1219)), but the streaming quantizer cannot rely on it.

`_discover_sanitize_plan` runs sanitize over `_TrackedTensor` recorders, and
`_DiscoveredPlan._apply_recipe` replays only single-source unary operations
(reshape, slice, transpose, moveaxis, astype, expand_dims). The split needs
`mx.from_fp8`, a per-shard `mx.pad`, and a multiply against a second live
tensor.

The recorder does not merely fail to record that multiply, it silently omits
it. See section 3.

### The fix

Move the split below sanitize, to where the weight and its scale are both in
hand.

`_LazyTensorIndex` now accepts a `config` and supports virtual tensors: logical
keys computed on demand from source keys that are hidden from the logical view.
This extends the mechanism already used for FP8 pairs, where the scale is
hidden and the dequant happens on materialization.

`omlx/patches/virtual_tensors.py` is the generic seam.
`omlx/patches/mimo_v2/virtual_qkv.py` is the MiMo registrar: it exposes
`q_proj`, `k_proj` and `v_proj` per layer and hides the fused weight and its
scale. Sanitize then sees the split tensors already present, its fused branch
is inert, and the plan records three ordinary passthrough entries. `oq.py`
itself never names a model.

Points worth review attention:

- The layout arithmetic lives in `omlx/patches/mimo_v2/fused_qkv_layout.py`,
  owned by oMLX, and both the vendored model and the registrar drive off it.
  It deliberately does **not** live in the vendored file:
  `apply_mimo_v2_patch` no-ops when upstream ships its own `mimo_v2`, at which
  point `mlx_lm.models.mimo_v2` resolves to upstream's module, which has none
  of these helpers. Reading them off that module would raise `AttributeError`
  out of `_LazyTensorIndex.__init__` the day PR 1219 merges. The vendored file
  keeps a single import and five call sites, so re-vendoring stays cheap.
- `fused_qkv_part_rows` is the only place the per-shard row formula is written;
  shard rows, split shapes and the split itself all derive from it.
  `fused_qkv_layout_matches` is shared by TP detection and per-layer
  validation, and `fused_qkv_keys` is the only spelling of the key template.
- Candidate keys are built from the config
  (`model.layers.{i}.self_attn.qkv_proj.weight`) rather than scanned out of the
  index. The MTP head lives at `model.mtp.layers.0` and carries sliding-window
  geometry while sitting at layer index 0, so anything that scans keys and then
  reads a layer index out of them misreads it as full attention and aborts over
  tensors sanitize is about to discard. Building keys from the config makes that
  structurally unreachable.
- Registration refuses when the declared geometry does not reproduce the
  on-disk shapes, rather than dequantizing a guess.
- Detection is gated on `model_type` and never assumes candidates exist, so an
  oQ output or the calibration proxy, which inherit the config but ship split
  tensors, is a no-op.
- The materializer holds the index **weakly**. It is reachable from the index
  through the closures stored in the virtual table, so a strong reference would
  form a cycle and stop `del all_weights` from freeing the index by refcount
  immediately before the deliberate `mx.clear_cache()` in the streaming loop.
- Each slice is released as it is handed over, rather than when all three have
  been served, so a layer's `q_proj` is not pinned across the `k` and `v`
  quantize passes.

---

## 3. The plan recorder now refuses what replay cannot reproduce

This part is pre-existing infrastructure rather than the MiMo fix, and can be
split out on request. It is confined to `_TrackedTensor` plus the `from_fp8`
and `pad` op fakes.

`_TrackedTensor.__mul__` and its siblings cloned the left operand and dropped
the right one entirely, without even unioning its sources. `_fake_from_fp8` and
`_fake_pad` built fresh tracked tensors with an empty recipe, discarding any
lineage recorded before them.

The consequence is not a missing feature, it is a wrong answer. A sanitize that
multiplies by a block scale on block-aligned shapes produces a plan that passes
the replayable gate with the multiply silently missing, so replay would ship
unscaled weights. Whether that failed loudly was luck: MiMo escapes only
because its padded layers happen to break a later reshape. Its sliding-window
layers are block-aligned and would not have.

These operations now poison the tensor, and the poison is sticky, so a later
replayable operation cannot launder the plan back into looking sound.
Discovery raises and the caller falls back to eager sanitize, which is the
outcome the design always intended. Scalar multiplies are unaffected; only a
second tracked tensor triggers it.

---

## Testing

`tests/test_oq_fused_qkv_virtual.py` compares against an oracle that maps every
element to its scale block using explicit integer arithmetic. It deliberately
does not reuse the production splitter: a helper factored out of the model and
then compared against that same model agrees with itself even when both are
wrong.

Weights are exact e4m3 values and every scale is a power of two, so the
comparison is bit exact and a misplaced block appears as a factor of two rather
than a rounding difference.

The suite also pins the layer asymmetry. The naive single-grid formula is
asserted to **fail** on padded layers and to **pass** on block-aligned ones.
The second assertion exists because a fixture built only from sliding-window
layers passes while proving nothing, which is the trap this checkpoint sets.

Additional regression tests cover the lifetime and recorder behaviour: that the
index is freed by refcount with the garbage collector switched off, that each
slice is dropped as it is served, that deleting a virtual key un-hides its
sources, and that a scale multiply cannot be laundered into a replayable plan.

The dequant was mutation checked. Reintroducing the single-grid formula is
caught on both padded layers in the fixture, with roughly half of `q_proj`
elements differing, and is invisible on the aligned layers.

Full suite passes (7350 passed, 53 skipped). `ruff check` reports the same
pre-existing findings in `oq.py` before and after this change; the new modules
and tests are clean.

## Production result

XiaomiMiMo/MiMo-V2.5 quantizes to oQ4, loads, and generates coherent output
with a thinking trace.

---

## Scope notes

The `mimo_v2_model.py` diff is 22 insertions and 78 deletions, and every
insertion is mechanical: one import, one constant reference, a three-line
adapter, and four call sites. No comments, no exported symbols, nothing
oMLX-specific beyond the import, so re-applying it over a future upstream
revision is cheap.

Four adjacent findings are out of scope here. Happy to open them separately if
useful:

1. `detect_fused_qkv_tp` returns TP=1 when no full-attention layer is usable.
   Safe on this checkpoint because layer 0 is full attention, but an all-SWA
   variant would resolve TP ambiguously and corrupt silently.
2. `_DiscoveredPlan.pop` handles the `split_{i}_{total}` transform but falls
   through to returning the whole source tensor for the index-list form
   `split_{i}` recorded by `_fake_split`. `dflash_laguna.py` calls
   `mx.split(fused, [q_dim, q_dim + kv_dim], axis=0)`, which takes that path.
3. In `universal_quant_predicate`, the `v_proj` substring test matches
   `qkv_proj`, so the later `qkv_proj` rule is unreachable for that spelling.
   No correctness impact, but it reads as though fused QKV gets 5 bits when it
   gets 6.
4. `_DiscoveredPlan._cache` is declared with a comment about multi-consumer
   sources and is never read or written. It is also the natural home for a
   shared one-source-to-many-outputs cache, which this PR had to build inside
   the MiMo registrar instead.

---


jundot:
Thanks for the detailed MiMo validation. The per-shard fused-QKV virtual tensor path and its independent numerical coverage look correct.

I found two small generic-path follow-ups that I’ll fold in before merge. I’ll keep packed residency for quantized-source sensitivity instead of pricing MiniMax MXFP8 and DeepSeek V4 FP4 sources as fully dequantized BF16, and I’ll keep `nested_unreplayable` sticky through `mx.split()` so a second source cannot be dropped during replay.

These adjustments do not change the MiMo implementation or the production validation in this PR. I’ll rerun the focused oQ tests after the changes and merge once they pass.

PR #2421, opened 2026-07-30
fix(oq): drop multimodal metadata from text-only outputs

blightbow:
## Summary

A text-only conversion strips the vision, audio and speech tensors, but the
output config and the copied sidecars keep advertising those modalities. The
result is a checkpoint that declares a vision model type and ships an image
preprocessor while containing no vision weights.

This is the same class of inconsistency `_normalize_mtp_in_config` already
guards against, and that function's docstring makes the general argument: an
output config should not claim capabilities whose tensors were stripped.

## What survives today

The `text_only` branch in `quantize_oq_streaming` removes a fixed list of keys.
It covers the nested `vision_config` / `audio_config` convention, but MiMo V2.5
also carries two top-level keys that the list does not know about, so both
survive into the output:

- `vision_model_type` (value `"mimovl"`)
- `processor_config`

Separately, the sidecar copy has no `text_only` guard, so
`preprocessor_config.json` and `processor_config.json` are copied for every
conversion.

Observed on a published text-only MiMo V2.5 conversion built with oMLX
`v0.5.4.dev1`. The tensor index contains zero `visual.`, `audio_encoder.`,
`speech_embeddings.` and `model.mtp.` tensors, confirming the weights really are
text-only, while `config.json` still carries `vision_model_type` and
`processor_config` and the repo still ships a `preprocessor_config.json`.

This is metadata only. It does not cause a load failure, and it does not affect
the VLM routing predicate, since `_has_vision_subconfig` tests `vision_config`,
`vit_config` and `mm_vision_tower`, all of which are correctly removed. The cost
is that the published artifact misrepresents what it can accept, and any
classifier keyed on a different spelling will misread it.

## Change

1. Extract the key list into `_normalize_text_only_in_config`, alongside the
   existing `_normalize_mtp_in_config`, and add `vision_model_type` and
   `processor_config` to it. Extracting it also makes the behaviour testable
   without running a conversion.
2. Add a `text_only` parameter to `_copy_model_sidecars` that skips the
   multimodal processor configs, except when one carries an inline chat
   template (see below).
3. Call `_copy_model_sidecars` from `quantize_oq_streaming` instead of the
   inline copy loop. The two were byte-identical, so this removes a duplicate
   that would otherwise need the same guard applied twice.

Net effect on the streaming path is a small reduction in code alongside the fix.

## Why the processor configs are not skipped unconditionally

Current Transformers writes chat templates to their own file, which oQ already
copies. Older processor repos instead keep one under a `chat_template` key
inside `processor_config.json` or `preprocessor_config.json`, and Transformers
still honours it on load rather than ignoring it: `processing_utils` warns that
the template "should be in a 'chat_template.jinja' file" and then uses it.

Skipping those files unconditionally would take such a template with them, and
losing a chat template is both silent and annoying to diagnose. This matters
beyond the auto-detected model types, because `text_only` is also a
caller-settable parameter, so any VLM can reach this path.

So a processor config that holds a `chat_template` is copied even under
`text_only`. A chat template is not modality metadata, and it may be the only
copy. Files carrying only modality settings, which is the case that motivated
this change, are still dropped. A file that cannot be read or parsed is also
kept, on the grounds that preserving something unparseable is the cheaper
mistake.

For reference, the file that prompted this on a real MiMo V2.5 conversion holds
only image-processing settings and no template:

```json
{"min_pixels": 3136, "max_pixels": 12845056, "patch_size": 16,
 "image_processor_type": "Qwen2VLImageProcessor",
 "processor_class": "Qwen2_5_VLProcessor", ...}
```

## Tests

`TestTextOnlyMultimodalMetadata` in `tests/test_oq.py`:

- every multimodal key is dropped from a MiMo-shaped config, parametrized per
  key so a future addition to the list gets its own case
- text keys survive, and a config with none of these keys is untouched
- `text_only=True` skips the processor sidecars while still copying the
  tokenizer files, and the default still copies them
- a processor config holding an inline `chat_template` is kept under
  `text_only`, while one holding only modality settings is still dropped
- an unparseable processor config is kept
- the base and multimodal pattern lists stay disjoint, since moving either name
  into the base list would bypass both the guard and the skip with nothing
  failing on the copy paths

Full suite passes. `ruff check` reports the same 44 pre-existing findings in the
touched files before and after.

## Possible follow-up, not included here

The key list is an allowlist of spellings and has now been outgrown once.
Matching a prefix set instead would survive the next family that invents a new
name. I kept this change to the narrow fix rather than bundling a behaviour
change that could strip a key some text model relies on, but I am happy to
follow up if you would prefer the broader approach.

Similarly, `_copy_model_sidecars` copies every `*.py` from the source. For an
MLX artifact whose class is resolved by the runtime, that Transformers modeling
code cannot load the quantized weights, and it is what makes the Hub mark the
repo `custom_code` and warn that `trust_remote_code=True` is required. Removing
it felt out of scope for a metadata fix.

---


jundot:
Thanks, the text-only metadata cleanup is correct, and preserving an inline chat template makes sense.

I found one small sidecar guard that I’ll fold in before merge. `_holds_chat_template()` should preserve syntactically valid non-mapping JSON such as `[]`, `null`, or a string instead of raising `AttributeError` at the final copy step. I’ll add the regression coverage and merge after it passes.

Issue #459, opened 2026-03-29
Bad Continuous Downloading


Ucag:
**Describe the bug**
My mac upgraded accidentally when I'm downloading some models and restarted. Then all my model downloads show that they all complete. Of course I know this is impossible because I just started the downloading for minutes.

**To Reproduce**
Download a big model then restart your computer.

**Expected behavior**
The unfinished downloading should continue after restarted.

**Desktop (please complete the following information):**
MacOS 26, latest for now.

n4ll3ec:
same issue

Durumsky:
Same here. It is very annoying: downloading for 20min, for whatever reason there is a failure (internet went off, the laptop entered sleep, some internal error withing omlx, etc), then it shows "downloading - 100%". Doesn't download anything though.

If I restart failed downloads, shows the above "downloading - 100%", instead of keep downloading from where it was.

PR #2432, opened 2026-07-30
fix(discovery): skip incompletely downloaded sharded models

JimStenstrom:
Addresses #459.

## Summary
- Model discovery now verifies a sharded checkpoint's weight files against the checkpoint's own manifest before registering it, and skips the directory with an actionable warning when shards are missing.
- This removes the false-complete **model listing** from #459: after an interrupted download plus a restart, the partial directory no longer shows up as a normal, load-ready model. (The empty download-task list after a restart is the other half of the issue — see the follow-up note under out-of-scope.)

## Why
An interrupted download leaves `config.json` (fetched in the first seconds) plus whichever weight shards finalized. Discovery treats "`config.json` exists" as "model present" (`_is_model_dir`) and `estimate_model_size()` sums whatever shards happen to be on disk — so the partial directory registers as a normal model, while the download queue (the in-memory `_tasks` dict, rebuilt empty on every server start) forgets the interrupted task. The model list then presents the partial model as complete, exactly as reported.

Reproduced live with a real 66.9 GB download (`poolside/Laguna-XS-2.1`, 14 shards), hard-killed at 47% with 1 shard finalized, then relaunched:

- `GET /admin/api/hf/tasks` after restart → `{"tasks": []}`
- `GET /admin/api/models` → the model listed as load-ready at "5.01 GB" (discovery's size estimate over the single finalized shard — the full checkpoint is 66.9 GB), with no indication anything is wrong
- server.log: `Download queued: poolside/Laguna-XS-2.1` … no `Download completed` … then post-restart `Discovered model: Laguna-XS-2.1 (type: llm, engine: batched, size: 5.01GB)`

One wrinkle the repro surfaced: at kill time `model.safetensors.index.json` was **not yet on disk** (there is no guarantee it lands ahead of the shards), so the completeness check cannot rely on the index alone — hence the filename-numbering fallback below.

## Changes
`omlx/model_discovery.py` — new `_missing_weight_shards()` helper plus a guard in `_register_model()` before type detection:

- **Index branch**: every string `weight_map` value must exist, resolved against the model directory per the HF index convention — so `./`-prefixed or subfolder entries are not misreported, and only files the manifest provably references and that are provably absent count. A corrupt / non-UTF-8 / non-dict index falls through to the numbering check instead of raising (`UnicodeDecodeError` is a `ValueError`, not a `JSONDecodeError` — the same trap `_is_helper_checkpoint` already documents for config.json).
- **No-index branch**: `-NNNNN-of-NNNNN.safetensors` groups must contain all N distinct indices (1-based per the HF convention, 0-based tolerated). Counting is by distinct in-range index, so duplicate-padding spellings of one shard cannot mask a gap.
- On missing shards: one WARNING naming the count, up to 3 example filenames, and pointing at resume ("finished shards are kept and reused" — consistent with `_cleanup_partial`, which deliberately preserves finalized shards; the message says "interrupted or still-running" because a rescan triggered by one download completing can see another still in flight), then the directory is skipped.

## Why this is safe (layouts verified to keep registering)
- Single-file `model.safetensors`, GGUF dirs, and adapters (which never reach `_register_model`) make no shard promises → unchanged. Sharded checkpoints of every engine type (embedding/reranker/audio/VLM) make the same manifest promises and get the same check, which is intended.
- Index plus extra non-indexed safetensors (e.g. oQ's `model-mtp.safetensors`): extras ignored, the index stays authoritative.
- Weights only in subdirectories (`estimate_model_size`'s `**/*.safetensors` layout): no top-level promises → unchanged.
- Mid-quantization oQ outputs (`-of-PLACEHOLDER`): don't match the numeric pattern.
- A skipped directory is still listed and deletable in the admin Local Models UI — those routes walk the disk independently of discovery. Loaded/loading engines are preserved by `EnginePool.discover_models`, so nothing serving is disrupted.

Named out-of-scope siblings (not bundled):
- `.bin`-sharded checkpoints (`pytorch_model.bin.index.json`) are not checked.
- An interruption before **any** shard landed still surfaces as the pre-existing `Failed to discover model … No model weights found` ERROR rather than the new message.
- Task persistence / auto-resume across restarts. With this fix the state is honest, and re-running the download resumes from the kept shards — verified live: the re-triggered download left the finalized shard untouched (its mtime predates the kill) and fetched only the remaining 13. A downloader-side startup reconciliation that restores a retryable task entry is feasible, since the repo id is reconstructible from the directory path (`target_dir = model_dir / repo_id`). Happy to follow up in that direction if you want it.
- Discovery now parses each sharded model's index.json per scan (server start / admin refresh). If that is a concern for very large indexes I can gate it behind a cheap filename pre-check.

## Tests
9 new cases in `tests/test_model_discovery.py` (existing file): interrupted-with-index, interrupted-without-index (the issue's actual on-disk shape), complete-with-index plus an extra unindexed file, complete-without-index, subpath `weight_map` values, corrupt/non-UTF-8 index fallback (both a complete and a gapped set), 0-based numbering, duplicate-padding gap masking, subdirectory-weights layout.

`PYTHONPATH=. python -m pytest tests/test_model_discovery.py -q` → **178 passed**.
Full `-m "not slow and not integration"` suite → 7473 passed; 3 failures in MTP verify-kernel tests reproduce identically on clean main on this machine (Apple-silicon numerics, unrelated paths).
The two core regression tests were verified to FAIL with the fix reverted.
Also verified against the real interrupted directory (snapshotted before resuming): fixed discovery excludes it ("13 weight shard(s) … missing"); after the download completed, the same directory registers normally (discovery estimate 65.41 GiB, from the same estimator that reported 5.01 GB when partial).
Finally, live-validated on an isolated server instance running the fixed code, with a second real download (5 shards, ~21 GB): two more SIGKILL→restart cycles — this time with the index on disk, exercising the index branch — excluded the partial directory with warnings that tracked the on-disk state across resumes (4 missing, then 1 missing naming the exact shard); the final resume fetched only the missing shard (108 s; the earlier shards' mtimes were untouched across both resumes), after which the model registered normally, as did a concurrent second download via the completion-triggered rescan.

### DIFF
--- a/omlx/model_discovery.py
+++ b/omlx/model_discovery.py
@@ -1082,6 +1082,7 @@ def _missing_weight_shards(model_dir: Path) -> tuple[int, list[str], str] | None
     None when the directory is complete or makes no shard promises.
     """
     idx_path = model_dir / "model.safetensors.index.json"
+    index_missing: tuple[int, list[str]] | None = None
     if idx_path.is_file():
         try:
             index = json.loads(idx_path.read_text())
@@ -1102,9 +1103,22 @@ def _missing_weight_shards(model_dir: Path) -> tuple[int, list[str], str] | None
                     if isinstance(v, str) and not (model_dir / v).is_file()
                 }
             )
-            if missing:
+            if not missing:
+                return None
+            referenced_count = len(
+                {v for v in weight_map.values() if isinstance(v, str)}
+            )
+            if len(missing) < referenced_count:
                 return len(missing), missing[:3], "model.safetensors.index.json"
-            return None
+            # Every shard the index references is absent. An interrupted
+            # download keeps the shards it already fetched, so
+            # zero-of-referenced is not that shape — it is a stale index
+            # describing a different sharding of the same checkpoint (seen
+            # in the wild: a 4-shard weight_map shipped over 2 actual
+            # shards). Loaders glob model*.safetensors and never consult
+            # the index, so fall through and judge by the shard files
+            # actually present.
+            index_missing = (len(missing), missing[:3])
 
     try:
         shard_names = [
@@ -1144,7 +1158,36 @@ def _missing_weight_shards(model_dir: Path) -> tuple[int, list[str], str] | None
                     f"{prefix}-{str(i).zfill(width)}-of-{total}.safetensors"
                 )
     if missing_count:
-        return missing_count, examples, "shard filename numbering (no index file)"
+        return (
+            missing_count,
+            examples,
+            (
+                "shard filename numbering (no shard referenced by the stale "
+                "index exists; judging by the shards present)"
+                if index_missing is not None
+                else "shard filename numbering (no index file)"
+            ),
+        )
+    if index_missing is not None:
+        if groups or (model_dir / "model.safetensors").is_file():
+            # A self-consistent complete shard set (or a consolidated
+            # single-file checkpoint) is present: the index is stale, the
+            # weights are whole. Register the model; loaders read the shard
+            # files directly and never consult the index.
+            n_missing, examples = index_missing
+            logger.warning(
+                f"Model at {model_dir} ships a stale "
+                f"model.safetensors.index.json: none of the {n_missing} "
+                f"shard(s) it references exist (e.g. {', '.join(examples)}), "
+                f"but the weight files present form a complete set — "
+                f"registering the model anyway."
+            )
+            return None
+        # No weight files at all: config + index landed first, shards did not
+        # (interrupted download before the first shard finished) — skip, and
+        # report against the only manifest available.
+        n_missing, examples = index_missing
+        return n_missing, examples, "model.safetensors.index.json"
     return None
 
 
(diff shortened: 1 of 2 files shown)


JimStenstrom:
Pushed a second commit covering a false-positive case found in the wild while testing something else: `mlx-community/Qwen3-Embedding-8B-mxfp8` ships a `model.safetensors.index.json` left over from a different sharding (a 4-shard weight_map over 2 actual, complete shards), so the guard hid a fully-downloaded model that loaders serve fine (they glob `model*.safetensors` and never consult the index).

The distinguishing shape: an interrupted download keeps the shards it already fetched, so *zero-of-referenced-present* is a stale index, not a partial download. In that case the guard now judges by the shard files actually present — a self-consistent complete set (or consolidated `model.safetensors`) registers with a stale-index warning; an on-disk gap still skips, reported against the shards present; config+index with no weights at all still skips.

Three new tests (the register case, the no-weights case, the on-disk-gap case); the register and gap tests fail without the fix. Verified against the real repo's directory. 181 discovery tests green.

jundot:
Thanks for tracing the interrupted-download state through discovery and covering the stale-index checkpoint case. The guard matches the current mlx-lm and mlx-vlm loader behavior, and the merged discovery suite passes without changing the local model catalog. LGTM as the model-discovery half of #459. I’ll keep #459 open for restart-time task recovery and automatic resume.

Issue #2439, opened 2026-07-31
SpecPrefill can produce malformed output after tool-result messages


aosama:
## Summary

SpecPrefill can corrupt the assistant turn after a tool-calling loop. In the observed failure, the model generated only a bare closing `</tool_response>` marker instead of the requested final answer.

This is a correctness bug, not an enhancement: the same model and task completed correctly when SpecPrefill was disabled.

## Environment

- oMLX: `0.5.4rc1`
- Target: `Qwen3.6-27B-oQ4e-mtp`
- SpecPrefill draft: `Qwen3.5-4B-MLX-4bit`
- Client: OpenCode `1.18.10`
- Tool-calling chat with several completed tool-result messages
- SpecPrefill keep rate: 20%
- SpecPrefill threshold: 1024 tokens

## Reproduction

1. Enable SpecPrefill for the target and draft models above.
2. Start an OpenCode session with tools enabled.
3. Ask for a task requiring several filesystem searches, for example: `Get me the latest 10 articles about A.I. from my reading material.`
4. Let the model issue multiple `read`, `glob`, and `bash` calls.
5. Observe the assistant turn after the final tool results.

The captured failing sequence had valid tool calls and valid tool outputs, then the final assistant turn contained only:

```text
</tool_response>

</tool_response>
```

OpenCode recorded one generated output token with `finish: stop`.

## Runtime evidence

Immediately before the malformed output, oMLX logged:

```text
prefix cache: ... re-prefills 6401 of 30977 tokens (reused 24576)
SpecPrefill: scored 6401 tokens in 3.6s, selected 1312/6401 (keep=20%, draft cache hit 4096, prompt 30977 = system 24209 + conv 6768, cached 24576)
SpecPrefill: sparse prefill 1312/6401 conv tokens in 4.9s
Chat completion: ... 1 tokens ... finish_reason=stop
```

A controlled replay with SpecPrefill disabled, after unloading and reloading the same target model, completed the equivalent multi-tool task and returned the expected numbered article list.

An older OpenCode session from July 10 also contains the same bare `</tool_response>` symptom, indicating this behavior predates the current issue #2177 work.

## Expected behavior

After completed tool-result messages, the target model should receive enough structurally complete and semantically relevant context to produce a valid assistant answer.

## Actual behavior

Sparse conversation prefill can produce a malformed bare tool-response closing marker and terminate immediately.

## Scope note

[...]

PR #2440, opened 2026-07-31
feat(specprefill): persist static target prefixes

### LINKED ISSUE #2177: Feature: Reuse Static System and Tool Prefixes with SpecPrefill
## Problem

When SpecPrefill is enabled, oMLX repeatedly runs the target model full prefill for stable prompt material on every request:

- System prompt
- Tool definitions
- Other unchanged instruction blocks

This happens even when consecutive requests share the same large static prefix. The draft model may reuse scoring state, but the target model still processes the full system/tool prefix again.

For agent workloads, this is costly. Clients such as Claude Code, Codex, OpenCode, and tool-using applications often send several thousand unchanged system and tool tokens with every turn.

## Observed Behavior

Example configuration:

- Target: `Qwen3.6-27B-oQ4`
- Draft: `Qwen3.5-4B-MLX-4bit`
- SpecPrefill enabled, 20% keep rate
- Paged SSD cache enabled
- Cache block size: 2,048 tokens
- Stable system prompt: 6k+ tokens

Repeated requests show the target entering `system prompt prefill` again, with `cached_tokens: 0`, despite the static prefix spanning multiple cache blocks.

## Desired Improvement

With SpecPrefill enabled, oMLX should recognize when the system prompt and tool definitions are unchanged from an earlier request.

After the first request, subsequent requests should avoid reprocessing that stable prefix through the target model. SpecPrefill should continue to apply to the changing conversation or user-message portion.

This request is intentionally limited to stable system and tool prefixes. It does not request target-cache reuse for sparse conversation/history state.

## Target User Experience

- First request processes the complete prompt normally.
- Later requests with the same system prompt and tools reuse the target already-computed static prefix.
- Only new or changed prompt content incurs target prefill work.
- Dashboard and API telemetry clearly distinguish target static-prefix reuse, draft-model cache reuse, and new prompt tokens processed by SpecPrefill.

## Expected Benefit

This should improve the practical experience of long-running agent sessions:

- Lower time to first token on follow-up requests.
- Less repeated GPU and memory-bandwidth work on large target models.
- Better responsiveness when system prompts and tool schemas are large.
- Less prefill contention for multi-request or multi-user servers.
- Better payoff from SpecPrefill in real agent workflows, where static instructions are commonly much larger than the changing user input.

aosama:
## Summary

- persist complete SpecPrefill target system/tool prefixes through the existing paged SSD and hot-cache tiers, including arbitrary terminal boundaries
- restore exact target prefixes across requests and server restarts without replacing longer ordinary prefix-cache hits
- expose distinct target-static and draft-prefix API telemetry and cover restart, pressure-reclaim, missing-terminal, and cache-policy behavior

The implementation extends `BlockAwarePrefixCache` with a domain-separated exact terminal block. Ordinary prefix matching remains full-block-only; this PR adds no cache manager, storage tier, or independent memory budget.

## Scope

- target reuse is limited to the stable system/tool prefix; sparse conversation/history KV is not reused
- target and draft caches remain separate model-specific instances over the existing cache infrastructure
- API telemetry is included; dashboard UI is intentionally unchanged

## Testing

- modified unit-test modules: `347 passed`
- full local non-slow/non-integration suite: `7489 passed, 52 skipped`; three unrelated numerical-tolerance failures reproduce identically on `origin/main`
- available model-free slow/integration coverage: `17 passed`; seven TurboQuant tests require an uncached Llama checkpoint
- generic real-model suite with local Qwen3.5-2B: `14 passed`
- Qwen3.6-27B target + Qwen3.5-4B draft restart/pressure test: `1 passed`
- focused Ruff import/error checks and `git diff --check` pass

Closes #2177

### DIFF
--- a/omlx/cache/observability.py
+++ b/omlx/cache/observability.py
@@ -101,6 +101,10 @@ def _compute_window(
     d_ssd_disk = delta("ssd_disk_loads")
     d_tokens_matched = delta("prefix_tokens_matched")
     d_tokens_requested = delta("prefix_tokens_requested")
+    d_target_static_hits = delta("target_static_hits")
+    d_target_static_misses = delta("target_static_misses")
+    d_draft_prefix_hits = delta("draft_prefix_hits")
+    d_draft_prefix_misses = delta("draft_prefix_misses")
 
     minutes = elapsed / 60.0
 
@@ -120,6 +124,26 @@ def _compute_window(
         "ssd_hot_rate": round(
             _safe_ratio(d_ssd_hot, d_ssd_hot + d_ssd_disk), 4
         ),
+        "target_static_hits": d_target_static_hits,
+        "target_static_misses": d_target_static_misses,
+        "target_static_hit_rate": round(
+            _safe_ratio(
+                d_target_static_hits,
+                d_target_static_hits + d_target_static_misses,
+            ),
+            4,
+        ),
+        "target_static_tokens_restored": delta("target_static_tokens_restored"),
+        "draft_prefix_hits": d_draft_prefix_hits,
+        "draft_prefix_misses": d_draft_prefix_misses,
+        "draft_prefix_hit_rate": round(
+            _safe_ratio(
+                d_draft_prefix_hits,
+                d_draft_prefix_hits + d_draft_prefix_misses,
+            ),
+            4,
+        ),
+        "draft_prefix_tokens_saved": delta("draft_prefix_tokens_saved"),
     }
 
 
@@ -130,6 +154,10 @@ def _compute_cumulative(counters: dict[str, int]) -> dict[str, Any]:
     ssd_disk = counters.get("ssd_disk_loads", 0)
     tokens_matched = counters.get("prefix_tokens_matched", 0)
     tokens_requested = counters.get("prefix_tokens_requested", 0)
+    target_static_hits = counters.get("target_static_hits", 0)
+    target_static_misses = counters.get("target_static_misses", 0)
+    draft_prefix_hits = counters.get("draft_prefix_hits", 0)
+    draft_prefix_misses = counters.get("draft_prefix_misses", 0)
 
     return {
         "prefix_hits": prefix_hits,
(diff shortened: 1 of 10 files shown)


JimStenstrom:
While preparing #2442 (the fix for #2439) I checked how this PR and that one interact, including running them together locally: the merge is conflict-free on current main, both PRs' test suites pass on the combined branch (413 tests), and live tool-calling agent runs on Qwen3.6-27B and on the hybrid-cache Qwen3.6-35B-A3B show this PR's static-prefix restore firing on every turn (`restored 500 static system-prefix tokens from tiered cache`, including immediately after a server restart, on both architectures) while #2442's mandatory-tail selection holds throughout, with no errors on either model.

The interaction seems worth recording for review planning: by making the static system/tool prefix reliably cached, this PR makes "SpecPrefill scores only the conversation/tool-history suffix" the common case. That is the regime where the #2439 failure occurred — the failing request in that report already had its full 24k system prompt cached (`cached 24576`, remaining system 0), scored only the 6,401-token tool-history suffix, and its log carries the dropped-final-chunk fingerprint that #2442 eliminates (`selected 1312/6401` → `sparse prefill 1312/6401`: no kickoff decrement, and 1312 = 41 × 32 exactly, so the final partial chunk lost the importance ranking).

So the two changes are complementary rather than competing: this PR makes the fast path the common path, #2442 makes that path structurally safe, and they compose cleanly in either merge order — verified together on both a full-attention and a hybrid target, not just diffed.

aosama:
Thank you for putting time into this Jim, There will be many benefits from optimizations around specprefill

aosama:
If possible @JimStenstrom  can you recommend a model pair to use while I experiment and/or attempt to optimize spec prefill? I have a 48GB M5 a pair that would fit.

jundot:
Thanks for the thorough cache and restart coverage. The exact terminal design composes cleanly with the existing tiered cache and #2442, and the combined tests pass. One thing I'll fold into a follow-up on main: after an exact restore supersedes a shorter ordinary prefix cache, `request.prompt_cache` should be replaced so the old target KV is not retained through decode. LGTM, merging.

JimStenstrom:
@aosama The pair from your #2439 report is honestly the one I'd recommend — it's what I used to reproduce that issue and to verify #2442 and the combined branch with this PR: Qwen3.6-27B as target (at oQ8-mtp on my machine) with your same Qwen3.5-4B-MLX-4bit as the scoring draft. For 48GB I'd run the 27B at oQ4/oQ4e as in your report rather than oQ8 — my oQ4 build is 16G on disk vs 28G for oQ8 and held ≈oQ8 quality in my evals — which leaves headroom for long agent-loop contexts like your failing request (30,977 tokens).

If you want a second target for coverage, Qwen3.6-35B-A3B at oQ4 is 20G on disk and is what I used as the hybrid-cache architecture pass: it exercises the hybrid-cache × SpecPrefill seams this PR touches, its GDN layers keep KV growth much slower at long context, and it still shows a live SP artifact worth chasing while you optimize — the mid-history-driven thinking-scaffold confusion described in the #2442 body, which the tail floor deliberately does not fix. The same 4B draft serves both targets.

One pairing constraint to keep in mind: the draft scores the target's token ids directly (`omlx/specprefill/draft.py` consumes the request's prompt tokens as-is), so the draft has to share the target's vocab — nothing validates this at load today, so a mismatched pair silently degrades selection rather than failing. Qwen3.5-4B against Qwen3.6 targets shares the vocab and is field-proven by your own report.

Since this PR is now merged, the common case is the regime from my earlier comment — static prefix cached, SP scoring only the conversation/tool suffix — so the wall-clock win to chase is the draft scoring pass over suffixes (from your log: 3.6s to score 6,401 tokens vs 4.9s for the sparse prefill itself). Happy to A/B candidate optimizations against the same pair on my M5 Max if that's useful.

Issue #2398, opened 2026-07-28
0.5.4 dev1 Laguna DFlash unexpected behavior


thinklikesqlhns:
I just started using Laguna models, I noticed behavior I wasn't expecting when I used these new dflash checkpoints with both Laguna-XS-2.1-8bit and Laguna-S-2.1-OQ3e-fast. 

Scenario: M5 20c 64gb ram, 0.5.4 dev1, load a ~51k token historical document pdf via chat (real, no repetition). Identical prompts 'Can you summarize this document for me' .

DFlash on (respective core dflash models direct from poolside, no quantization, model specific) produced the following behavior:
XS: 1316.4 t/s prefill, 28.0 t/s generation.

S: 684.8 4 t/s prefill, 17.4 t/s generation.

DFlash off produced the following behavior:
XS: 924.1 t/s prefill, 43.7 t/s generation.

S: 1189.3 t/s prefill, 32.4 t/s generation.

Based on my experiences with gemma-4, I would've expected a slight dip (5-10%) in prefill speed and a ~1.2-1.5x jump in token generation, but this is entirely different. Am I misunderstanding the functionality or is there something off on the M5?

 - macOS Version: 26.5.2

[2026-07-28 server.log](https://github.com/user-attachments/files/30479628/2026-07-28.server.log)

d1slike:
I have the same issue. M4 Max, 128 GB.
mlx-community/Laguna-S-2.1-oQ4e-fast
poolside/Laguna-S-2.1-DFlash-NVFP4

Use oMLX performance benchmark on 32k context.
Dflash off - 49tps
Dflash on - 11tps (with quantization and cache off slightly better)

Maybe we do something wrong ?

zviratko:
I tested it briefly and only got meaningful acceptance with draft window size 2048, but it wasn't deterministic or reliable and only with the bf16 unqantized drafter (though I think it's FP4?). That non-determinism looks weird to me, but benchmark prompts have also changed in the meantime (are they random now?) so there's not much to compare it to.

d1slike:
Hi @FabioMalpezzi !  Maybe you could help ?

FabioMalpezzi:
[...]

PR #2445, opened 2026-07-31
feat(dflash): warn on precision-mismatched draft pairs, surface acceptance in dashboard

FabioMalpezzi:
Follow-up to https://github.com/jundot/omlx/issues/2398: with 51k-token prompts, the Laguna DFlash pairings reported there produced only ~1 accepted draft token per cycle, so DFlash was slower than plain decoding — and nothing surfaced that. This PR adds the first two defensive pieces promised in the issue.

## What's included

- **Load-time pairing warning** — warn when draft and target both make an explicit, contradictory precision claim (e.g. a `-DFlash-NVFP4` draft on an oQ4e target). A generic `-DFlash` name is treated as unknown rather than BF16-only, so the documented quantized Qwen/Gemma4 pairs stay silent. The warning is logged and shown in the dashboard.
- **Speculation metrics in the dashboard** — per-session counters folded from each request's `SummaryEvent` (both generate paths): draft share of output, accepted draft tokens/cycle, output tokens/cycle, and fallback-AR requests with their reason, rendered in the Active Models card. Draft share and per-cycle numbers are deliberately separate: the ~51% "acceptance" logged in #2398 reads healthy while the engine is actually losing.
- **Docs** — the pairing table gains the missing `Laguna-XS-2.1-NVFP4-mlx` row and the mismatch failure mode is documented.

## Not in this PR

Tracked as remaining follow-ups from #2398: porting the batched-engine prefill patches to the DFlash path (with parity coverage), automatic fallback to plain decode when speculation is losing, and INT4/FP8 (vLLM-format) checkpoint validation.

## Testing

148 tests passed across `test_dflash_engine.py`, `test_active_models_visibility.py`, `test_dflash_laguna.py`, `test_dflash_lifecycle.py`, `test_dflash_draft_config.py` — including the exact #2398 pairings and the counter accounting from that issue's log (768 tok / 377 cycles). Dashboard row verified visually against the compiled Tailwind stylesheet.

### DIFF
--- a/docs/experimental/dflash_mlx_integration.md
+++ b/docs/experimental/dflash_mlx_integration.md
@@ -89,6 +89,7 @@ DFlash upstream registers `QwenGdnTargetOps` and `Gemma4TargetOps`. oMLX also re
 | google/gemma-4-31b-it | z-lab/gemma-4-31B-it-DFlash |
 | google/gemma-4-26b-a4b-it | z-lab/gemma-4-26B-A4B-it-DFlash |
 | poolside/Laguna-XS-2.1 | poolside/Laguna-XS-2.1-DFlash |
+| poolside/Laguna-XS-2.1-NVFP4(-mlx) | poolside/Laguna-XS-2.1-DFlash-NVFP4 |
 | poolside/Laguna-S-2.1 | poolside/Laguna-S-2.1-DFlash |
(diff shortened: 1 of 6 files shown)


jundot:
Thanks for adding these guardrails. The precision pairing check matches the published Laguna configs, and the SummaryEvent accounting is correctly wired into both generation paths. One thing I'll fold into a direct follow-up on main: keep the session fallback count visible after later speculative requests and move the new dashboard labels into i18n. LGTM, merging.

Issue #2046, opened 2026-06-30
Support for DeepSeek-V4-Flash-DSpark


imleooooo:
## Is your feature request related to a problem? Please describe.
With the release of the DeepSeek-V4 architecture (including DeepSeek-V4-Flash and DeepSeek-V4-Pro), a new speculative decoding module called DeepSpec has been introduced. 
------------------------------
## Describe the solution you'd like
I would like to request official native support for running deepseek-ai/DeepSeek-V4-Flash-DSpark locally on omlx inference engine. This implementation should include:

   1. Speculative Decoding Support (DeepSpec): Proper loading and execution of the extra DSpark speculative module attached to the main FP4/FP8 mixed MoE model architecture.

------------------------------
## Additional context

* Model Repository: huggingface.co
* Technical Details:
* Parameter Scale: 284B total parameters, 13B activated parameters.
   * Context Length: Supports up to 1 Million tokens.
   * Precision: Uses an FP4 + FP8 mixed MoE layout.
   * Reference Repository for Speculative Logic: github.com

https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash-DSpark
https://github.com/deepseek-ai/DeepSpec

SmoothSailingMusic:
This may speed things up

https://github.com/ARahim3/mlx-dspark

Issue #1133, opened 2026-05-09
DeepSeek-V4-Flash MTP patch silently fails: "has no mtp_forward" despite "patch applied" log


wellon3013:
**Describe the bug**
Environment
oMLX version: v0.3.9.dev1 (DMG)
Hardware: Mac Studio M3 Ultra, 512GB unified memory
Model: mlx-community/DeepSeek-V4-Flash-mxfp8
Model path: /Volumes/wellon/lmstudio/model/mlx-community/DeepSeek-V4-Flash-mxfp8
Describe the bug
When loading DeepSeek-V4-Flash-mxfp8 with MTP enabled in admin settings, omlx logs claim the MTP patch was applied successfully, but at inference time the MTP path is not activated because the Model class lacks mtp_forward. The two log lines directly contradict each other:
2026-05-09 13:49:06,047 - omlx.utils.model_loading - INFO - Native MTP patch applied for /model/mlx-community/DeepSeek-V4-Flash-mxfp8 (model_type=deepseek_v4, active)
2026-05-09 13:49:46,138 - omlx.patches.mlx_lm_mtp.batch_generator - DEBUG - MTP path not active: model mlx_lm.models.deepseek_v4.Model has no mtp_forward (qwen35 patch may not have applied to this class)

**To Reproduce**
Download mlx-community/DeepSeek-V4-Flash-mxfp8 (mxfp8 quant with sidecar MTP weights: model-mtp.safetensors)
Load the model in omlx v0.3.9.dev1
Enable MTP in admin model settings
Send any chat completion request
Observe server log — "patch applied" followed by "has no mtp_forward"

**Expected behavior**
MTP patch should inject mtp_forward into mlx_lm.models.deepseek_v4.Model, and inference should log MTP path activated for uid=X with speculative decoding active.
Actual behavior
MTP patch reports success but the Model class is unmodified. Inference falls back to autoregressive decoding with no MTP acceleration.
Sidecar file verification (all pass)
I verified the model files are correct:
model-mtp.safetensors contains 32 keys, all with mtp.* prefix (correct for omlx sanitize())
index.json has 32 MTP entries pointing to model-mtp.safetensors
config.json: model_type=deepseek_v4, num_nextn_predict_layers=1, hc_mult=4
Root cause analysis
[...]

PR #2446, opened 2026-07-31
feat(deepseek-v4): support official 0731 DSpark decoding

wolfyy970:
## Current status

This remains a draft. The official checkpoint and 0731 protocol load and run,
but DSpark is **not yet merge-ready**: target-only and DSpark greedy generation
can produce different token streams. Production validation therefore keeps
`mtp_enabled=false` until OFF/ON exact-token parity passes on the real
checkpoint.

The earlier throughput figures in this PR came from the experimental DSpark
path and must not be treated as a correctness-safe performance claim.

## What changed

Adds support for the official `deepseek-ai/DeepSeek-V4-Flash-0731` checkpoint
and its dedicated message protocol, plus an experimental implementation of its
integrated DSpark module.

- recognizes the 0731 DSpark configuration and checkpoint weight layout
- implements the three-stage DSpark attention path, Markov head, and confidence
  head
- adds DSpark block proposals to the existing batched MTP generator
- uses the request sampler for proposals so sampling semantics remain aligned
- vendors DeepSeek's dedicated 0731 message encoder and wires it into oMLX's
  tokenizer/output protocol
- preserves already-parsed tool arguments during replay instead of nesting
  another `arguments` object on each retry
- makes speculative rollback safe around DeepSeek V4's ratio-4
  `BatchPoolingCache` compression boundaries
- reconciles the backbone cache before falling back from speculative to
  standard decoding
- teaches oQ discovery about the official DSpark projection names
- preserves `CacheList` subtype signatures when rotating prefix-cache tips are
  rewritten
- makes custom-kernel packaging reject and purge stale Python ABIs

## Why

The 0731 release is not the older single-head DeepSeek V4 MTP layout. It ships
an integrated DSpark module with multiple target layers and a dedicated
non-Jinja message protocol. Treating it as legacy MTP either leaves the extra
weights unmapped or enters a speculative path whose cache cannot always be
rolled back after a rejected draft crosses a pooling boundary.

This addresses the native DSpark request in #2046. It is also related to the
broader DeepSeek V4 MTP report in #1133, although this PR is specifically
validated against the official 0731 checkpoint.

## User impact

With speculative decoding disabled, `DeepSeek-V4-Flash-0731` now:

- strict-loads directly from the official 48-shard checkpoint
- exposes its native 1,048,576-token context
- supports thinking mode and `reasoning_effort`
- serves through the normal OpenAI-compatible oMLX API
- preserves prefix-cache signatures across follow-up turns
- replays tool calls without recursively nesting their arguments

The integrated DSpark path requires no separate draft model, but it remains
experimental pending exact-token parity.

## Validation

Unit and packaging validation:

- 176 focused DeepSeek, parser, thinking-budget, cache, and packaging tests
  passed
- 199 scheduler tests passed after the cache-freshness threshold fix
- 818 cache/prefix tests passed in an independent broad run
- critical Ruff checks and `git diff --check` passed
- native macOS app built with Xcode 26.6, passed strict codesign verification,
  and loaded all four custom-kernel modules with the bundled Python 3.11 ABI

Real-checkpoint validation:

- model: `deepseek-ai/DeepSeek-V4-Flash-0731` (official 48-shard, 163.2 GB
  checkpoint)
- hardware: Mac Studio, M3 Ultra, 256 GB unified memory
- target-only production profile: 1,048,576 context, 8,192 max output,
  low reasoning with a 2,048-token thinking budget
- three consecutive tool-call replays retained flat JSON arguments
- five-turn growing-prefix test completed without a cache-signature mismatch
- DSpark OFF/ON greedy A/B **failed** exact response equality; MTP remains
  disabled while the target-verification numerics are investigated

## Provenance

The dedicated message encoder under `omlx/patches/deepseek_v4/vendor/` is
derived from the encoder distributed with the official DeepSeek 0731 model,
kept isolated so its protocol can be reviewed and updated independently.

### DIFF
--- a/apps/omlx-mac/Scripts/build.sh
+++ b/apps/omlx-mac/Scripts/build.sh
@@ -110,6 +110,7 @@ PROJECT_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
 REPO_ROOT="$(cd "$PROJECT_DIR/../.." && pwd)"
 PACKAGING_DIR="$REPO_ROOT/packaging"
 CUSTOM_KERNEL_DIRS=(
+    "$REPO_ROOT/omlx/custom_kernels/bonsai"
     "$REPO_ROOT/omlx/custom_kernels/glm_moe_dsa"
     "$REPO_ROOT/omlx/custom_kernels/minimax_m3"
     "$REPO_ROOT/omlx/custom_kernels/qwen35_prefill"
@@ -229,6 +230,7 @@ _custom_kernel_pythonpath() {
 
 _clean_custom_kernel_build_artifacts() {
     local dir
+    local kernel_name
     for dir in "${CUSTOM_KERNEL_DIRS[@]}"; do
         [ -d "$dir" ] || continue
         find "$dir" -maxdepth 1 -type f \( \
@@ -239,16 +241,23 @@ _clean_custom_kernel_build_artifacts() {
     done
 
     if [ -d "$REPO_ROOT/build" ]; then
-        for ext_name in \
-            "omlx.custom_kernels.glm_moe_dsa._ext" \
-            "omlx.custom_kernels.minimax_m3._ext" \
-            "omlx.custom_kernels.qwen35_prefill._ext"; do
+        for dir in "${CUSTOM_KERNEL_DIRS[@]}"; do
+            kernel_name="$(basename "$dir")"
             find "$REPO_ROOT/build" \
                 -type d \
-                -name "$ext_name" \
+                -name "omlx.custom_kernels.${kernel_name}._ext" \
                 -prune \
                 -exec rm -rf {} +
         done
+        # setuptools' build/lib tree can retain extensions from a previous
+        # Python ABI even after the CMake target directories are removed.
+        # build_ext --inplace copies every retained artifact into the source
+        # package, so remove those native outputs before rebuilding.
+        find "$REPO_ROOT/build" \
+            -type f \
+            -path "*/omlx/custom_kernels/*" \
+            \( -name "*.so" -o -name "*.dylib" -o -name "*.metallib" \) \
+            -delete
     fi
 }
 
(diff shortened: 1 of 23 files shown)


jundot:
Thanks for adding the official 0731 DSpark path. I ran a target-only vs. DSpark greedy A/B test with a 4K code prompt and 128 output tokens: OFF completed in 15.38s, while ON completed in 14.11s with 80.6% acceptance, but the token streams diverged (`1b7e746e…` vs. `46b8ad35…`) despite completing without rollback fallback.

Greedy speculative decoding must remain token-identical to target-only decoding, so I can’t merge this as-is. Could you align block verification with DeepSeek’s M=1 decode numerics and add a real-checkpoint exact-token equality test for MTP OFF vs. ON?

wolfyy970:
Thanks for catching this. I reproduced the failure locally against the same
official checkpoint: target-only produced 121 tokens in 17.36s and DSpark
produced 124 tokens in 6.64s for the same greedy request, with different token
streams. I have disabled MTP in the production profile and updated the PR body
so the earlier throughput numbers are not presented as correctness-safe.

The first two emissions are ordinary single-token target forwards; the first
possible divergence is the first block-verification result. A small
DeepSeek-V4 probe also shows non-bit-identical logits between one multi-token
target forward and equivalent serial M=1 forwards. On M3 Ultra, the current
native sparse-attention kernel is additionally slower for DSpark-sized L=2..6
calls than the MLX reference path, so I am separating the short verification
path from the prefill-oriented kernel while investigating M=1-aligned
verification.

I will keep this draft and will not re-enable or claim DSpark support until a
real-checkpoint OFF/ON exact-token equality test passes.

wolfyy970:
Closing this draft rather than leaving an incorrect DSpark implementation as a merge candidate. Jundot's review is right: the official 0731 checkpoint does not preserve greedy target-only token identity with the current MTP verification path, and the measured speedup is not acceptable without that exactness guarantee. Thanks for catching it. I'll separate the work and return, if warranted, with a focused target-only/MTP-off change backed by repeatable correctness and performance evidence.

PR #2449, opened 2026-08-01
fix(reranker): support Jina v3.5 reranker projector key naming

### LINKED ISSUE #2422: jina-reranker-v3.5-mlx does not load correctly
**Describe the bug**
The latest Jina Reranker v3.5 fails to load in oMLX.

**To Reproduce**
1. Download `jinaai/jina-reranker-v3.5-mlx` using Models > Downloader
2. Go to Settings > Model Settings
3. Ensure the model's type is correctly inferred as 'reranker'
4. Click 'Load' on 'jinaai/jina-reranker-v3.5-mlx'
5. See error:

```
Jina projector is malformed: missing keys ['linear1.weight', 'linear2.weight'] in /Volumes/MODELS/omlx/jinaai/jina-reranker-v3.5-mlx/projector.safetensors.
Available keys: ['projector.0.weight', 'projector.2.weight'].
Reload models after fixing the files to retry.
```

It seems that the keys changed between v3 and v3.5 and oMLX is trying to load the wrong ones.
 
**Expected behavior**
Model loads without an error.

 - macOS Version: 26.6
 - oMLX Version 0.5.3

**Additional context**
 
- [Model's Discussion Page: #1 Correct usage with oMLX](https://huggingface.co/jinaai/jina-reranker-v3.5-mlx/discussions/1)

The model's reference Python code works, provided `rerank.py` and `modeling.py` are present. Example script `reranker_cli.py`:

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#   "mlx>=0.0.1",
#   "mlx-lm>=0.0.1",
#   "numpy>=1.24.0",
#   "safetensors>=0.4.0"
# ]
# ///

from rerank import MLXReranker

reranker = MLXReranker("/Volumes/MODELS/omlx/jinaai/jina-reranker-v3.5-mlx", "/Volumes/MODELS/omlx/jinaai/jina-reranker-v3.5-mlx/projector.safetensors")

query = "What are the health benefits of green tea?"
documents = [
    "Green tea contains antioxidants called catechins that may help reduce inflammation.",
    "Studies show that drinking green tea regularly can improve brain function.",
    "Basketball is one of the most popular sports in the United States.",
]

results = reranker.rerank(query, documents)
for result in results:
    print(f"Score: {result['relevance_score']:.4f}  {result['document'][:80]}...")
```

Run with `uv run reranker_cli.py`

damascoemi:
## Summary
- `jinaai/jina-reranker-v3.5-mlx` fails to load with "Jina projector is
  malformed: missing keys ['linear1.weight', 'linear2.weight']" because
  oMLX's `_load_jina_projector` only recognized v3's key names.
- v3.5 exports the identical Linear -> ReLU -> Linear projector from an
  `nn.Sequential` container, so PyTorch/MLX auto-name the weight keys by
  list index (`projector.0.weight`, `projector.2.weight`) instead of the
  named `linear1`/`linear2` submodules v3 used. Confirmed against Jina's
  reference implementation, which performs this same remap itself:
  https://huggingface.co/jinaai/jina-reranker-v3.5-mlx/blob/3dd4ac901ccdcac85abe3815df0a0aaaf44e4a21/modeling.py
  (`Projector.__init__` defines `fc1`/`fc2`; `_load_projector` maps
  `projector.0.weight` -> `projector.fc1.weight` and `projector.2.weight`
  -> `projector.fc2.weight`). Shapes ((512,1024)/(512,512), no bias) are
  unchanged between versions. Also reported at
  https://huggingface.co/jinaai/jina-reranker-v3.5-mlx/discussions/1.
- `_load_jina_projector` now tries both key schemes in order before
  raising, with error messages that name whichever scheme(s) were
  checked.

## Test plan
- Added `test_load_jina_projector_v35_sequential_keys`, mirroring the
  existing v3 test but with v3.5's key names, asserting identical
  projector output.
- Added `test_load_jina_projector_unrecognized_keys_raises_clear_error`
  for the neither-scheme-matches failure path.
- Reproduced the original error locally against a synthetic
  `projector.safetensors` matching the issue's reported keys/shapes,
  confirmed it's fixed after the change.
- Not yet validated against the actual downloaded v3.5 weights — @T1T4N
  (reporter, and the one who filed the matching HF discussion above),
  if you get a chance to try this branch against your real model
  directory, that'd be great confirmation before merge.

Fixes #2422

### DIFF
--- a/omlx/models/reranker.py
+++ b/omlx/models/reranker.py
@@ -573,37 +573,51 @@ class MLXRerankerModel:
         # raises "TypeError: data type 'bfloat16' not understood".
         weights = mx.load(str(projector_path))
 
-        required_keys = ("linear1.weight", "linear2.weight")
-        missing_keys = [key for key in required_keys if key not in weights]
-        if missing_keys:
+        # v3 ships two named nn.Linear submodules ("linear1"/"linear2").
+        # v3.5 exports the same fc1 -> ReLU -> fc2 stack from an nn.Sequential
+        # container, so the keys are auto-named by list index instead
+        # ("projector.0"/"projector.2", index 1 is ReLU). Confirmed against
+        # upstream's own remap in _load_projector:
+        # https://huggingface.co/jinaai/jina-reranker-v3.5-mlx/blob/3dd4ac901ccdcac85abe3815df0a0aaaf44e4a21/modeling.py
+        key_schemes = (
+            ("linear1.weight", "linear2.weight"),
+            ("projector.0.weight", "projector.2.weight"),
+        )
+        first_key = second_key = None
+        for scheme in key_schemes:
+            if all(key in weights for key in scheme):
+                first_key, second_key = scheme
+                break
+
+        if first_key is None:
             raise ValueError(
-                f"Jina projector is malformed: missing keys {missing_keys} in "
-                f"{projector_path}. "
+                "Jina projector is malformed: none of the expected key schemes "
+                f"{key_schemes} were fully present in {projector_path}. "
                 f"Available keys: {sorted(weights.keys())}"
             )
 
-        linear1_weight = weights["linear1.weight"]
-        linear2_weight = weights["linear2.weight"]
+        linear1_weight = weights[first_key]
+        linear2_weight = weights[second_key]
 
         if len(linear1_weight.shape) != 2 or len(linear2_weight.shape) != 2:
             raise ValueError(
                 "Jina projector weights must be 2D matrices: "
-                f"linear1.weight={linear1_weight.shape}, "
-                f"linear2.weight={linear2_weight.shape}."
+                f"{first_key}={linear1_weight.shape}, "
+                f"{second_key}={linear2_weight.shape}."
             )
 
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for fixing the projector key mismatch—I verified that the real v3.5 BF16 projector now loads while the v3 path remains intact. I do not think this is mergeable as v3.5 support yet: the pinned mlx-lm Qwen3 loader drops `layer_types` and `sliding_window`, and oMLX still uses the v3 prompt/scoring path instead of v3.5 dual matching and block fusion. Could you extend this with a real v3.5 loading/inference path and a parity smoke test against the reference implementation?

damascoemi:
Thanks for the review and verifying against the real BF16 weights, very appreciated. Agreed this isn't full v3.5 support yet. I'll dig more into the pinned mlx-lm Qwen3 loader and v3.5's reranker and follow up with a loading/inference path plus the parity smoke test against the reference impl as you suggested.

damascoemi:
Follow up for #2494. Support for sliding-window attention, dual matching, block fusion. Verified against the real checkpoint with a parity test against the reference `rerank.py`.

PR #2450, opened 2026-08-01
feat(mtp): native MTP support for Step-3.7-Flash (~1.25x, verified)

True2456:
## Summary

Adds native MTP (speculative decoding) support for Step-3.7-Flash, following the existing `deepseek_v4_model.py`/`glm_moe_dsa_model.py` pattern — Step-3.7 uses the same DeepSeek-V3-style nextn design (`enorm`/`hnorm`/`eh_proj`/`shared_head` naming) those already support.

Neither StepFun's own HF reference (`modeling_step3p7.py`) nor upstream mlx-lm implement MTP inference for this architecture at all — the checkpoint ships the weights and `config.json` declares `num_nextn_predict_layers`, but nothing consumes them. The forward pass here is a direct port of StepFun's own reference C++ implementation (`stepfun-ai/llama.cpp`, branch `step3p5-mtp`, `src/models/step35-iswa.cpp`) — not inferred from tensor shapes.

## Verified results

Via oMLX's own throughput benchmark, `mtp_enabled` true vs false, same checkpoint/prompts, A/B across reloads, on a REAP-pruned + mixed-precision-quantized Step-3.7-Flash checkpoint:

| test | baseline tgTPS | MTP tgTPS | speedup |
|---|---|---|---|
| pp 4096 / tg 128 | 52.7 | 64.4 | 1.22x |
| pp 8192 / tg 128 | 51.2 | 63.3 | 1.24x |
| pp 16384 / tg 128 | 48.8 | 61.1 | 1.25x |
| pp 32768 / tg 128 | 42.6 | 55.2 | 1.30x |

Speedup grows with context, +0.4GB peak memory. ~1.25x is the expected magnitude, not a ceiling — StepFun's own runtime is depth-1 (one draft token per verify cycle), which caps the theoretical best at 2x.

## Two upstream bugs found and fixed along the way

Both were completely silent — clean load, correct generated output, zero throughput gain, no exception anywhere — which made them expensive to isolate.

1. **`utils/model_loading.py`: `_nextn_weight_prefixes()`** only emits bare `model.layers.{n}.` prefixes. `_checkpoint_has_mtp_weights()` therefore reports zero MTP weight tensors for any VLM-shaped checkpoint (`language_model.model.layers.{n}.*`), and the settings UI refuses to let `mtp_enabled` be toggled on at all — not Step-3.7-specific, this affects any VLM-shaped checkpoint using the nextn (extra-decoder-layer) MTP layout. Fixed by also emitting `language_model.model.layers.{n}.` and `model.language_model.layers.{n}.` variants.
2. **`utils/model_loading.py`: `_is_mtp_compatible()`** didn't list `step3p7` in its architecture allowlist, so the patch was never invoked for this `model_type` regardless of config/weights. One line added.

Separately (in the new `step3p7_model.py` itself): `step3p5.ModelArgs` is a dataclass, and `BaseModelArgs.from_dict` keeps only declared fields — `num_nextn_predict_layers` isn't one, so it was silently dropped, `self.mtp` never attached, and the whole MTP path was inert while the model generated perfectly. Fixed with a `_patch_model_args()` wrapper, the same fix shape `deepseek_v4_model.py`/`glm_moe_dsa_model.py` already use for their own architectures — I'd designed for this in the first draft but forgot to actually implement it, which is exactly what cost the most time to find.

Full writeup of both bugs and the debugging method used to isolate the dataclass one (instantiating the patched model against a tiny synthetic config under oMLX's own bundled Python, asserting each condition `batch_generator._mtp_common_eligible` checks — no full 90GB+ model load needed) is in the `step3p7_model.py` module docstring.

## Known simplification

`mtp_clamp_accept`/`mtp_partial_rollback` (draft-rejection cache rollback) are simplified versions of `deepseek_v4_model.py`'s pattern — checking `is_trimmable()` only, without that file's rotated-cache `_mtp_undo` fallback. More conservative (may decline a rollback the fuller check would allow), not incorrect; the throughput numbers above already reflect this as-is.

## Testing

- All three files `py_compile`-clean.
- Model-side patch mechanics verified in isolation (tiny synthetic config, no full checkpoint load): `num_nextn_predict_layers` survives `from_dict`, `self.mtp` attaches, `mtp.block.is_sliding`/`self_attn.num_heads` come out correct for a sliding-attention MTP layer, `make_mtp_cache()` returns the right cache type, `mtp_forward()` produces finite logits of the correct shape.
- End-to-end on the real checkpoint: loads cleanly, generates correct output, and the throughput benchmark above confirms the draft/verify cycle is actually engaging (not just loading without crashing).

### DIFF
--- a/omlx/patches/mlx_lm_mtp/__init__.py
+++ b/omlx/patches/mlx_lm_mtp/__init__.py
@@ -105,6 +105,7 @@ def apply_mlx_lm_mtp_patch() -> bool:
         nemotron_h_chain,
         nemotron_h_model,
         qwen35_model,
+        step3p7_model,
     )
 
     if not cache_rollback.apply():
@@ -119,6 +120,8 @@ def apply_mlx_lm_mtp_patch() -> bool:
         logger.debug("DeepSeek-V4 MTP patch did not apply (likely missing base patch)")
     if not glm_moe_dsa_model.apply():
         logger.debug("GLM-5.2 MTP patch did not apply (likely missing base patch)")
+    if not step3p7_model.apply():
+        logger.debug("Step-3.7 MTP patch did not apply (likely missing base patch)")
     if not nemotron_h_model.apply():
         logger.debug("nemotron_h MTP patch did not apply (likely import error)")
     elif not nemotron_h_chain.apply():
--- a/omlx/patches/mlx_lm_mtp/step3p7_model.py
+++ b/omlx/patches/mlx_lm_mtp/step3p7_model.py
@@ -1,0 +1,463 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Native MTP (Multi-Token Prediction) monkey-patch for Step-3.7-Flash.
+
+Neither StepFun's own reference `modeling_step3p7.py` nor oMLX's vendored
+`mlx_lm.models.step3p7`/`step3p5` implement Step-3.7's MTP draft layer at
+all -- `step3p5.Model.sanitize()` actively discards any weight key for a
+layer index >= num_hidden_layers (and anything containing ".mtp"), by
+design, matching upstream mlx-lm's stance that Step-3.5/3.7 support ships
+without MTP. This patch adds it, following oMLX's own proven pattern for
+DeepSeek-V4 (`patches/mlx_lm_mtp/deepseek_v4_model.py`) and GLM-5.2
+(`glm_moe_dsa_model.py`) -- both already-working implementations of the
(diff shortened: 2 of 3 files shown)


jundot:
Thanks for adding and validating the native Step-3.7 MTP path. The model-side port fits the existing MTP interface cleanly, and the real-checkpoint A/B results confirm that the draft/verify cycle is active. I found two contained follow-ups around raw-HF norm sanitization and VLM-prefixed nextn preservation in oQ, which I’ll fold into a follow-up on main along with focused tests and formatting cleanup. LGTM. I’ll merge this now.

Issue #2399, opened 2026-07-28
vlm_mtp decode silently drops per-request logits processors (grammar, thinking_budget, penalties)


picosquare:
Disclaimer: This bug was found while coding on a different project that uses omlx + structured_output. The report was written by Claude Code. I manually verified the witness script below and can confirm the issue.

**Describe the bug**

When `vlm_mtp_enabled` is on, structured-output requests come back **unconstrained**: HTTP 200, no `Warning` header, nothing in the log. The grammar *is* compiled — a malformed schema is rejected in under 40 ms with xgrammar's own error text — but the compiled mask never reaches the sampler on the vlm_mtp decode path.

Because `_route_to_vlm_mtp` serves only **one** request at a time, the bug is intermittent in a very specific way: requests that fall back to BatchGenerator *are* correctly constrained. So the same body gives 0/6 constrained when sent sequentially and 4/6 when sent concurrently. That contrast is the clearest signal — the mask exists and works, it is just skipped on the fast path.

`thinking_budget` and the repetition / presence / frequency penalties are dropped on the same path, for the same reason.

**To Reproduce**

1. Load a model with a vlm_mtp drafter attached — here `gemma-4-26B-A4B-it-qat-4bit` (`engine_type: vlm`) with `gemma-4-26B-A4B-it-qat-assistant-4bit`. Confirm via `/v1/models/status`.
2. Make sure `vlm_mtp_enabled` is on for that model.
3. Save the script below as `repro.sh`, set `OMLX_URL` and `OMLX_API_KEY`, run `bash repro.sh`.
4. Compare the two tallies.

```bash
#!/usr/bin/env bash
# The grammar `root ::= "red"` admits exactly the three characters `red`.
# Anything else — "Red.", a sentence — proves no mask was applied.
set -eu

URL="${OMLX_URL:-http://localhost:8000}/v1/chat/completions"
MODEL="${OMLX_MODEL:-gemma-4-26B-A4B-it-qat-4bit}"
N="${N:-6}"

body=$(cat <<JSON
{
  "model": "$MODEL",
  "messages": [{"role": "user", "content": "Name a primary colour."}],
  "max_tokens": 32,
  "temperature": 0,
  "structured_outputs": {"choice": ["red"]}
}
JSON
)

tmp=$(mktemp -d); trap 'rm -rf "$tmp"' EXIT

ask() { curl -sS "$URL" -H 'Content-Type: application/json' \
[...]

PR #2456, opened 2026-08-01
feat(vlm_mtp): enforce thinking budget inside MTP speculative decoding

### LINKED ISSUE #2455: `thinking_budget` + `vlm_mtp`: support enforcing the thinking budget inside MTP speculative decoding (instead of falling back to BatchGenerator)
> [!NOTE]
> **Disclaimer:** This issue and the accompanying PR (#2456) were authored with Claude Code (Fable 5, ultracode mode) and **functionally confirmed working** on real hardware: Apple Silicon (32 GB), `gemma-4-12B-it-8bit` + `gemma-4-12B-it-qat-assistant-bf16` VLM MTP drafter. Measured results are in the PR.

## Summary

Since 9387ebd (#2399), any request carrying a per-request logits processor is routed away from the vlm_mtp path to BatchGenerator, and `thinking_budget_enabled` is mutually exclusive with `vlm_mtp_enabled` in model settings. That fixed the silent-drop bug — but it means you cannot have MTP speculative decoding and a thinking budget at the same time, ever. On releases before that fix the combination was worse: the budget was silently ignored (e.g. `thinking_budget=100`, `max_tokens=200` → the model thinks for all 200 tokens and returns an empty answer).

For the thinking budget specifically, the fallback is unnecessary: the budget can be enforced *inside* the speculative round loop, exactly, with the MTP speedup intact. I have a working implementation (PR attached) tested with gemma-4-12B-it-8bit + gemma-4-12B-it-qat-assistant-bf16.

## Why the #2399 rationale doesn't apply to the thinking budget

The comment in `_route_to_vlm_mtp` (and the discussion in #2399) gives two reasons processors can't run on this path:

1. *"run_vlm_mtp_decode threads only a sampler into mlx-vlm's `_mtp_rounds`, and a sampler sees logits without the token history processors need."*
   mlx-vlm already exposes the missing application point: when the sampler object provides `sample_target(logprobs, row_ids, positions)`, the verify walk (`_positioned_target_tokens` / `_speculative_walk_deferred_greedy`) calls it with the **absolute position of every candidate slot**. A wrapper sampler can therefore maintain the full token history itself. Two properties of the round loop make the application *exact*, not approximate:
   - Every emitted token equals the target-side sample at its position (an accepted draft token equals the target's sample by definition of `_speculative_walk`; a rejected slot emits the target's sample directly). So sampling from processor-modified logits per position reproduces BatchGenerator semantics token-for-token.
   - Slot *i*'s sample is only ever used when the draft prefix `d[0:i]` was fully accepted — in which case the wrapper's own sampled prefix equals `d[0:i]`. Building slot *i*'s history from the wrapper's own samples is therefore correct in every case where the slot's token survives.

2. *"Those processors are stateful."*
   True — and handled: positions past the first rejection are discarded and re-sampled next round, so the wrapper checkpoints processor state after every position (`snapshot_state()` / `restore_state()` on the processor) and rewinds to `positions[0]` at the start of each call. Rounds only re-sample a *suffix* (committed positions are never revisited), so the needed checkpoint always exists.

The grammar half of #2399 is different, and this proposal does **not** touch it: a grammar mask makes the unconstrained drafter's proposals systematically rejectable, so speculation buys nothing there — falling back to BatchGenerator remains the right call for grammar (and for penalties, until someone wants them). The gate simply becomes capability-based: processors that implement snapshot/restore ride along; everything else still falls back.

## Why this is worth it

- A thinking budget is a *no-op* on logits until the budget boundary, so acceptance rates during normal decode are completely untouched — the drafter is never fighting a constraint mask. Only the ~3–4 forced close tokens degrade to ~1 token/round.
- Thinking budget + speculative decoding is a natural combination: reasoning models produce long thinking phases, which is exactly where the MTP speedup pays off most. Making the two mutually exclusive gives up the speedup on precisely the workloads that want it.
- Output distribution is unchanged: emitted tokens are always the target sampler's per-position samples, same as BatchGenerator.

## Cost

One small `.item()` sync per candidate slot during verify (block_size+1 tiny syncs per round), only while a budget processor is attached. Requests without a thinking budget take the exact same code path as today.

## Repro of the current behavior

- ≤0.5.3: `vlm_mtp_enabled` + request `thinking_budget=100`, `max_tokens=200` → budget silently ignored, whole budget consumed by thinking, empty `content`.
- 0.5.4rc1: same request → log shows `vlm_mtp routing skipped … falling back to BatchGenerator`; budget enforced but no speculation. Settings-level combo is rejected/migrated away entirely.

## With the patch

Same request → `vlm_mtp decode started …` in the log, thinking stops at the budget (forced close, same `ThinkingBudgetProcessor` semantics as BatchGenerator, including #1905/#1864 behaviors — no divergence), answer present, MTP acceptance stats reported.

Happy to split the settings/UI relaxation from the core scheduler change if you'd prefer them separate.

Maxim-Mazurok:
> [!NOTE]
> **Disclaimer:** This PR was authored with Claude Code (Fable 5, ultracode mode) and **functionally confirmed working** on real hardware: Apple Silicon (32 GB), `gemma-4-12B-it-8bit` + `gemma-4-12B-it-qat-assistant-bf16` VLM MTP drafter. Measured results below under "Functional validation".

Closes #2455 (also relates to #2399).

## What

`thinking_budget` now works *with* `vlm_mtp_enabled` instead of forcing a BatchGenerator fallback (0.5.4rc1) or being silently dropped (≤0.5.3). The budget is enforced inside mlx-vlm's MTP round loop with BatchGenerator-identical semantics, keeping the speculative speedup — including during the thinking phase, which is exactly where reasoning models benefit from it most.

## How

mlx-vlm's verify walk consults `sampler.sample_target(logprobs, row_ids, positions)` when present, passing the absolute position of every candidate slot. The new `omlx/speculative/processing_sampler.py::MTPProcessingSampler` implements that hook: for each slot, in order, it applies the request's logits processors with the token history reconstructed for that position, samples with the base sampler, and checkpoints processor state. Two properties make this exact rather than approximate:

1. Every emitted token equals the target-side sample at its position (accepted draft ⇒ equal by definition of `_speculative_walk`; rejected slot ⇒ the target's sample is emitted). Sampling from processor-modified logits per position therefore reproduces BatchGenerator behavior token-for-token.
2. Slot *i*'s logits are conditioned on draft prefix `d[0:i]`, and its sample is only used when that prefix was fully accepted — in which case the wrapper's own sampled prefix equals `d[0:i]`, so histories built from the wrapper's own samples are correct whenever they matter.

Statefulness (the concern in #2399) is handled with position-keyed rewind: `ThinkingBudgetProcessor` gains `snapshot_state()`/`restore_state()`, and the wrapper restores the checkpoint for `positions[0]` at the start of every call. Rounds only re-sample a suffix, so the checkpoint always exists.

Budget forcing composes with speculation naturally: the forced close token mismatches the drafter's proposal, is emitted via the normal rejection path, and the next round continues from it — ~1 token/round for only the 3–4 forced close tokens. Until the boundary the processor is a no-op on logits, so acceptance rates are untouched.

## What still falls back

The routing gate is now capability-based: processors exposing `snapshot_state`/`restore_state` ride along; anything else (grammar, repetition/presence/frequency penalties) still declines vlm_mtp exactly as today. Grammar stays out deliberately — a grammar mask makes the unconstrained drafter's proposals systematically rejectable, so there is nothing to win. Routing also declines when `speculative_logits_from_hidden` is not visible to the round loop — checked via `vlm_mtp_positioned_sampling_available()`, which mirrors `_VLMAdapterMTPProxy`'s visibility rules. In particular, mRoPE adapters (Qwen VLMs) hide the inner model's `speculative_*` fast paths, so those targets fall back to BatchGenerator loudly instead of silently dropping the budget; an adapter that grows an mRoPE-safe positioned hook re-opens the route automatically.

## Changes

- `omlx/speculative/processing_sampler.py` (new): the wrapper + `supports_vlm_mtp_processing()`.
- `omlx/api/thinking.py`: `snapshot_state`/`restore_state` on `ThinkingBudgetProcessor` (pure addition; BatchGenerator path unchanged).
- `omlx/scheduler.py` `_route_to_vlm_mtp`: capability-based gate, positioned-hook eligibility check, wrapper construction, first-bonus processing (history = prompt), processor reset on setup-failure fallback.
- `omlx/model_settings.py`: `thinking_budget_enabled` removed from `vlm_mtp_processor_conflicts` (validation + load-time migration + request-time merge follow automatically).
- Admin UI (`dashboard.js`, `_modal_model_settings.html`, `en.json`) and macOS app (`ModelSettingsScreenVM.swift`, `ModelSettingsScreen.swift`): thinking-budget rows un-gated from VLM MTP; penalties/grammar gating unchanged. Note: non-English i18n strings for `vlm_mtp_processor_conflict` still mention thinking budget — left for a follow-up translation pass.

## Tests

- `tests/test_vlm_mtp_thinking_budget.py` (new, 16 tests): snapshot/restore; wrapper positioned semantics incl. draft-rejection rewind; **end-to-end through the real mlx-vlm `_mtp_rounds`** with fake target/drafter modules (budget forces close inside speculation; holds under per-round rejection; natural close respected; control shows a bare sampler drops the budget); routing-gate cases; happy-path wrapper construction.
- `tests/test_model_settings.py`: exclusivity test updated (budget combo now allowed; penalties/grammar still raise).
- Positioned-hook visibility (mRoPE) regression tests: an equivalence sweep pins the gate helper against the real `_VLMAdapterMTPProxy` attribute resolution for every (mrope, adapter-hook, lm-hook) combination, plus the reported Qwen case (inner LM has the hook, proxy hides it → decline).
- 299 tests across the touched files and the thinking/vlm_mtp/settings neighborhoods pass on Apple Silicon (Metal), after rebasing onto v0.5.4. The one failure in that sweep, `test_inkling_vlm_mtp.py::test_verify_rollback_matches_sequential_decode`, is pre-existing — it fails identically on clean `main` at c59b4cb without this patch.

## Functional validation

Apple Silicon (32 GB), `gemma-4-12B-it-8bit` + `gemma-4-12B-it-qat-assistant-bf16` drafter (VLM MTP, block_size=4), request `thinking_budget=100`, `max_tokens=200`, thinking enabled:

- **Before** (≤0.5.3 behavior, reproduced with `vlm_mtp_enabled` and no budget enforcement): thinking consumes all 200 completion tokens, `content` is empty (measured: 197 reasoning tokens, 0 content tokens).
- **After, vlm_mtp on + patch**: log shows `vlm_mtp decode started` (no `routing skipped` line) and `vlm_mtp stats: … rounds=73 accepted=127/219 (58.0%) tokens_per_round=2.74 emitted=200`; thinking stops at the budget with the forced close (96 reasoning tokens by tokenizer count) and the answer is present (100 content tokens, `finish_reason=length`).
- **Parity, vlm_mtp off (BatchGenerator baseline)**: same request → 94 reasoning tokens / 100 content tokens — within 2 tokens of the speculative path, i.e. identical `ThinkingBudgetProcessor` semantics on both paths.
- **Speedup retained**: 200 tokens in 11.5 s generation time with MTP vs 25.8 s on BatchGenerator for the same budgeted request (~2.2× on this pairing).

Happy to split the settings/UI relaxation from the core scheduler change if you'd prefer them separate.

### DIFF
--- a/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
+++ b/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
@@ -714,11 +714,9 @@ private struct AdvancedTab: View {
                 Row(label: String(localized: "settings.advanced.thinking_budget.label",
                                   defaultValue: "Thinking Budget",
                                   comment: "Row label for the thinking budget field"),
-                    sublabel: vm.vlmMtpEnabled
-                        ? vm.vlmMtpProcessorLockedReason
-                        : String(localized: "settings.advanced.thinking_budget.sub",
-                                 defaultValue: "Limit thinking tokens for reasoning models. Forces end of thinking when exceeded.",
-                                 comment: "Sublabel for the thinking budget field")) {
+                    sublabel: String(localized: "settings.advanced.thinking_budget.sub",
+                                     defaultValue: "Limit thinking tokens for reasoning models. Forces end of thinking when exceeded.",
+                                     comment: "Sublabel for the thinking budget field")) {
                     HStack(spacing: 8) {
                         if vm.thinkingBudgetEnabled {
                             TextInput(text: vm.bindProfile($vm.thinkingBudgetTokens),
@@ -727,8 +725,6 @@ private struct AdvancedTab: View {
                         Toggle("", isOn: vm.bindProfile($vm.thinkingBudgetEnabled))
                             .labelsHidden().toggleStyle(.switch)
                     }
-                    .disabled(vm.vlmMtpEnabled)
-                    .help(vm.vlmMtpEnabled ? vm.vlmMtpProcessorLockedReason : "")
                 }
                 Row(label: String(localized: "settings.advanced.tool_result_limit.label",
                                   defaultValue: "Limit Tool Result Tokens",
--- a/apps/omlx-mac/Sources/AppView/ViewModels/ModelSettingsScreenVM.swift
+++ b/apps/omlx-mac/Sources/AppView/ViewModels/ModelSettingsScreenVM.swift
@@ -786,8 +786,8 @@ final class ModelSettingsScreenVM {
         }
         if vlmMtpProcessorConflict {
             return String(localized: "settings.vlm_mtp.conflict.processors",
-                          defaultValue: "Unset repetition / presence penalty and thinking budget before enabling VLM MTP.",
-                          comment: "Tooltip / sublabel shown when VLM MTP can't be enabled because penalty or thinking-budget settings are set")
+                          defaultValue: "Unset repetition / presence penalty before enabling VLM MTP.",
+                          comment: "Tooltip / sublabel shown when VLM MTP can't be enabled because penalty settings are set")
         }
         return nil
     }
@@ -795,11 +795,13 @@ final class ModelSettingsScreenVM {
     /// Settings that materialize as per-request logits processors, which the
     /// vlm_mtp decode path cannot apply (#2399). Mirrors
     /// vlm_mtp_processor_conflicts() in model_settings.py; neutral values
-    /// (repetition 1.0, presence 0.0) do not conflict.
+    /// (repetition 1.0, presence 0.0) do not conflict. Thinking budget is
+    /// exempt: it is applied on the vlm_mtp path at verify time
+    /// (MTPProcessingSampler).
     var vlmMtpProcessorConflict: Bool {
         if let rep = Double(repetitionPenalty), rep != 1.0 { return true }
         if let pres = Double(presencePenalty), pres != 0.0 { return true }
-        return thinkingBudgetEnabled
+        return false
     }
 
     /// Sublabel / tooltip for the sampling rows locked while VLM MTP is on.
--- a/omlx/admin/i18n/en.json
+++ b/omlx/admin/i18n/en.json
@@ -548,7 +548,7 @@
   "modal.model_settings.vlm_mtp": "VLM MTP",
   "modal.model_settings.vlm_mtp_hint": "Speculative decoding via an external MTP drafter model (Gemma 4 assistant or Qwen MTP). ~1.3–1.6× single-request speedup; concurrent eligible requests fall back to standard batching.",
   "modal.model_settings.vlm_mtp_conflict": "Disable DFlash / SpecPrefill / MTP / TurboQuant KV first — VLM MTP owns the speculative decode path.",
-  "modal.model_settings.vlm_mtp_processor_conflict": "Unset repetition / presence penalty, thinking budget and guided grammar first — VLM MTP cannot apply per-request logits processors.",
+  "modal.model_settings.vlm_mtp_processor_conflict": "Unset repetition / presence penalty and guided grammar first — VLM MTP cannot apply these per-request logits processors. Thinking budget is supported.",
   "modal.model_settings.vlm_mtp_processor_locked": "Locked while VLM MTP is enabled — this setting needs per-request logits processors, which VLM MTP cannot apply.",
   "modal.model_settings.vlm_mtp_draft_model": "Drafter model",
   "modal.model_settings.vlm_mtp_draft_model_placeholder": "Select an assistant or MTP drafter…",
--- a/omlx/admin/static/js/dashboard.js
+++ b/omlx/admin/static/js/dashboard.js
@@ -1437,6 +1437,8 @@
             // which the VLM MTP decode path cannot apply (#2399). Mirrors
             // vlm_mtp_processor_conflicts() in model_settings.py; neutral
             // values (repetition 1.0, presence 0.0) do not conflict.
+            // Thinking budget is exempt: it is applied on the vlm_mtp path
+            // at verify time (MTPProcessingSampler).
             vlmMtpProcessorConflict() {
                 const ms = this.modelSettings;
                 if (!ms) return false;
@@ -1445,7 +1447,6 @@
                 const pres = num(ms.presence_penalty);
                 return (rep !== null && rep !== 1.0)
                     || (pres !== null && pres !== 0.0)
-                    || !!ms.enableThinkingBudget
                     || !!ms.guided_grammar_enabled;
             },
 
--- a/omlx/admin/templates/dashboard/_modal_model_settings.html
+++ b/omlx/admin/templates/dashboard/_modal_model_settings.html
@@ -458,15 +458,9 @@
                                     <div class="min-w-0">
                                         <span class="text-sm font-medium text-neutral-700">{{ t('modal.model_settings.thinking_budget') }}</span>
                                         <p class="text-xs text-neutral-500 mt-0.5">{{ t('modal.model_settings.thinking_budget_hint') }}</p>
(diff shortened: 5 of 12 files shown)


jundot:
Thanks for the detailed work here. I found one case that still needs fixing before merge. The scheduler checks `speculative_logits_from_hidden` on the inner language model, but Qwen VLM runs MTP through the mRoPE proxy, which intentionally hides that hook to preserve correct position handling. As a result, the gate passes but mlx-vlm uses the plain sampler later, so `thinking_budget` is silently skipped. Please either add an mRoPE-safe positioned hook or check the actual proxy and fall back, with a regression test for this path.

Maxim-Mazurok:
> [!NOTE]
> **AI comment:** written by Claude Code (Fable 5, ultracode mode) on Max's behalf. The fix and tests below were implemented, run, and functionally re-verified on real hardware before posting.

Good catch — confirmed. The gate probed `speculative_logits_from_hidden` on the inner language model, but the round loop resolves `lm` to the `_VLMAdapterMTPProxy`, which intentionally hides `speculative_*` for mRoPE adapters — so on a Qwen VLM target the gate passed and the verify walk then sampled through the plain vectorized path, silently skipping the budget.

Fixed in f5c3801 by checking what the round loop will actually see: a new `vlm_mtp_positioned_sampling_available()` in `omlx/speculative/vlm_mtp.py` mirrors the proxy's visibility rules exactly (adapter attribute first; `speculative_*` blocked when the adapter has `_uses_mrope`; fall-through to the inner model otherwise), and `_route_to_vlm_mtp` now uses it — Qwen VLM + `thinking_budget` falls back loudly to BatchGenerator instead of silently dropping the budget, per the #2399 contract.

Regression tests added:
- a parametrized equivalence sweep pinning the helper against the **real** `_VLMAdapterMTPProxy` attribute resolution for every (mrope, adapter-hook, lm-hook) combination, so the helper can't drift from the proxy;
- the reported case exactly (inner LM has the hook, mRoPE adapter hides it → routing declines);
- an adapter-level hook under mRoPE keeps the route open;
- non-mRoPE pass-through unchanged.

On the "add an mRoPE-safe positioned hook" alternative: I went with check-and-fall-back, since a positioned `speculative_logits_from_hidden` on the mRoPE adapter needs position-correct logits from hidden states — exactly what the proxy exists to protect. Because the helper checks the adapter attribute first, an adapter that later grows such a hook re-opens the vlm_mtp route automatically, with no scheduler change.

Also rebased the branch onto v0.5.4 (clean, no conflicts) and re-verified functionally on `gemma-4-12B-it-8bit` + `gemma-4-12B-it-qat-assistant-bf16`, `thinking_budget=100` / `max_tokens=200`: 96 reasoning + 100 content tokens, `vlm_mtp stats: rounds=75 accepted=125/225 (55.6%) tokens_per_round=2.67`, no degradation warnings. 29 tests in the budget suite + 270 across the touched neighborhoods pass.

jundot:
Thanks for the follow-up. The mRoPE visibility gate now matches the object mlx-vlm actually resolves in the round loop, and the end-to-end coverage verifies rollback and forced-close behavior through `_mtp_rounds`. I confirmed the focused VLM MTP, thinking-budget, and settings tests pass against current `main`. LGTM, merging.

PR #2461, opened 2026-08-01
fix(embedding): probe the compile smoke test with a traced attention_mask

### LINKED ISSUE #2447: Embedding compile smoke-test passes but real requests fall back to eager (mask-dependent branch in qwen3; smoke test omits attention_mask)
## Summary

On oMLX 0.5.3 with `Qwen3-Embedding-8B-mxfp8`, `_try_compile()` reports success at model load, but the **first real** `/v1/embeddings` request logs:

```text
compiled embedding path failed for <model>: [eval] Attempting to eval an array during function transformations like compile or vmap is not allowed.; disabling compile and falling back to eager generate()
```

so all production traffic silently runs the eager path. (Distinct from #282, which was the array-without-primitive crash — this one is an eval forced *during tracing*.)

## Root cause

Two parts:

1. **Upstream (mlx-embeddings)**: `models/qwen3.py:last_token_pool` uses a Python `if` on a lazy comparison (`left_padding = attention_mask[:, -1].sum() == ...; if left_padding:`). Under `mx.compile` with a traced `attention_mask`, the `if` forces an eval during tracing → the error above. Fix submitted upstream: Blaizzy/mlx-embeddings#70 (branch-free pooling, numerically identical).

2. **oMLX smoke-test blind spot**: `_try_compile()` validates with `test_inputs = {"input_ids": mx.zeros((1, 4), ...)}` — **no `attention_mask`**. The model then constructs a default mask internally as a constant, which is legal to eval during tracing, so the smoke test passes while any real request (mask arrives as a traced input) fails. Suggest the smoke test include an `attention_mask` in `test_inputs` so compile viability is probed on the representative path, and/or log the eager fallback at WARNING once with a hint (it currently looks healthy unless you read logs closely — we only noticed while debugging a 55k-chunk backfill's throughput).

## Environment

- oMLX 0.5.3 (Mac app bundle), bundled mlx-embeddings 0.1.0
- Model: Qwen3-Embedding-8B-mxfp8
- Apple Silicon, macOS 25.5

## Repro

1. Load any qwen3 embedding model with compile enabled (default).
2. Observe `mx.compile enabled ... (primitive embedding path)` at load.
3. POST any real `/v1/embeddings` request → the warning above, eager fallback.

https://claude.ai/code/session_01Jsssr7vdMR7E8CpBpwb1uT

### LINKED ISSUE #2448: Embedding throughput degrades steadily with server uptime, eventually stops completing requests (331 → 230 → ~100 tok/s → wedge)
## Summary

On oMLX 0.5.3 (Mac app, launchd service) serving `Qwen3-Embedding-8B-mxfp8` under a sustained `/v1/embeddings` workload (a 55k-document backfill), throughput decays monotonically with server uptime, and after several days the server stops completing embedding requests entirely while `/health` still returns healthy.

## Measured timeline (same host, same model, same workload)

| Server uptime | Observed throughput |
|---|---|
| fresh restart | ~331 tok/s (16×~2k-token batch), 73ms warm single text |
| ~7 hours | ~230 tok/s |
| overnight (~1–2 days) | 60–120 tok/s (100–525s per 16-input request) |
| ~5 days | wedged: single 8-token request does not complete in 280s; `/health` fine; process near 0% CPU |

`launchctl kickstart -k` fully restores fresh performance every time, which points at state accumulating in the server process (Metal/MLX allocator fragmentation or cache growth?) rather than anything in the request stream — request shapes were identical across the timeline. `engine_pool.current_model_memory` stays constant (~8.2GB, embedder only, loaded_count=1).

## Environment

- oMLX 0.5.3, launchd-managed, Apple Silicon (64GB, final_ceiling ≈ 55.7GB), macOS 25.5
- Model: Qwen3-Embedding-8B-mxfp8 (eager embedding path — compile falls back per #2447)
- Continuous embedding traffic: mixed 40–6,000-token texts, requests of 1–32 inputs

## Repro

Serve any embedding model and drive continuous `/v1/embeddings` traffic for several hours; time a fixed probe text periodically. Latency for the identical probe grows by hours-scale uptime; a kickstart resets it.

Happy to run instrumented builds or collect Metal/allocator stats if useful — this host reproduces it reliably under real load.

https://claude.ai/code/session_01Jsssr7vdMR7E8CpBpwb1uT

JimStenstrom:
`_try_compile()` validates the compiled embedding path with `input_ids` only. Real requests always carry an `attention_mask` on the tokenizer path (`prepare_inputs` emits one), so a model whose forward branches on the mask — e.g. mlx-embeddings qwen3's `last_token_pool` Python `if` — passes the mask-less probe (its internally-built default mask is a tracing constant, legal to eval) and then the first real request trips the traced-mask branch, logs a warning, and silently disables compile for the rest of the process. The load log claims `compiled=True` while production serves eager.

This adds the mask to the probe, matching the reranker smoke test which already includes it, so compile viability is decided at load time on the input the branch actually depends on, and the load log reports the real compiled state.

Fixes #2447

**Why this can't regress other models**: every `Model.__call__` in mlx-embeddings 0.1.0 accepts `attention_mask` (or `**kwargs`), and only qwen3's pooling branches on it — so no model that compiled before can newly fail the probe with a TypeError. Verified on real models in both directions:
- `mlx-community/nomicai-modernbert-embed-base-bf16`: compiles through the new probe at load and serves real requests compiled, no fallback.
- Qwen3-Embedding-0.6B (mlx-embeddings 0.1.0): load now logs `mx.compile unavailable ... [eval] Attempting to eval an array during function transformations ...` and `compiled=False` — identical serving behavior to before (it was falling back on the first request anyway), decided earlier and logged honestly.

**Why a synthetic mask rather than probing with real `_prepare_embedding_inputs` output**: a real tokenizer round-trip adds a second load-time failure mode and can't be constructed generically for custom multimodal processors. And going further toward full representativeness (e.g. adding `token_type_ids`) would TypeError on qwen3, whose `__call__` takes no `**kwargs` — `attention_mask` is the one input that is both universally accepted and the actual branch variable. Custom-input processors (e.g. qwen3_vl `prepare_embedding_inputs`) can still produce request dicts the probe can't anticipate; the existing request-time fallback remains as the backstop for those, unchanged.

qwen3 embedding models actually *running* compiled additionally needs the upstream pooling fix (Blaizzy/mlx-embeddings#70); until that lands they run eager either way — this PR makes that visible instead of silent.

**Tests** (`python -m pytest tests/test_engine_keepalive.py tests/test_embedding.py` → 98 passed, 1 deselected):
- Two new tests in `tests/test_engine_keepalive.py` (which holds the existing mx.compile-probe `_try_compile` cases) exercising real `mx.compile` — no mocks: a mask-branching fake model mirroring the qwen3 pooling pattern must fall back at load; a mask-free model must still compile. Both hermetic against an exported `OMLX_EMBEDDING_COMPILE`.
- RED proof: with the probe fix reverted, `test_mask_branching_model_falls_back_at_load` fails (compile wrongly reports success); with the fix it passes.

Surfaced while investigating #2448 (same reporter): the silent eager fallback is why that sustained embedding workload ran the eager path. This PR does not fix #2448's throughput decay — it makes the active code path visible and honestly logged.

### DIFF
--- a/omlx/models/embedding.py
+++ b/omlx/models/embedding.py
@@ -499,7 +499,17 @@ class MLXEmbeddingModel:
 
             self._compiled_embed = mx.compile(_compiled_embed)
 
-            test_inputs = {"input_ids": mx.zeros((1, 4), dtype=mx.int32)}
+            # Real requests arrive with a traced attention_mask (the tokenizer
+            # path always emits one). A mask-less probe lets a model whose
+            # forward branches on the mask pass here — its internally-built
+            # default mask is a tracing constant — and then the first real
+            # request trips the traced-mask branch and silently disables
+            # compile for the rest of the process (issue #2447). The reranker
+            # probe already includes the mask.
+            test_inputs = {
+                "input_ids": mx.zeros((1, 4), dtype=mx.int32),
+                "attention_mask": mx.ones((1, 4), dtype=mx.int32),
+            }
             _ = self._compiled_embed(test_inputs)
 
             logger.info(
--- a/tests/test_engine_keepalive.py
+++ b/tests/test_engine_keepalive.py
@@ -2,9 +2,73 @@
 """Tests for embedding/reranker engine mx.compile integration."""
 
 import asyncio
+from types import SimpleNamespace
 from unittest.mock import MagicMock, patch
 
+import mlx.core as mx
 import pytest
+
+
+class _MaskBranchingModel:
+    """Forward with a Python `if` on a mask-dependent lazy comparison.
+
+    Mirrors mlx-embeddings qwen3's last_token_pool: with attention_mask=None
+    the default mask is built from the (static) shape, so it is a tracing
+    constant and legal to eval; with a traced attention_mask input the same
+    `if` forces an eval during tracing and mx.compile raises (issue #2447).
+    """
+
+    def __call__(self, input_ids, attention_mask=None):
+        if attention_mask is None:
+            attention_mask = mx.ones(input_ids.shape, dtype=mx.int32)
+        left_padding = attention_mask[:, -1].sum() == attention_mask.shape[0]
+        if left_padding:
+            pooled = input_ids[:, -1:]
+        else:
+            pooled = input_ids[:, :1]
+        return SimpleNamespace(text_embeds=pooled.astype(mx.float32))
+
+
+class _MaskFreeModel:
+    """Forward with no data-dependent Python branching — compiles cleanly."""
+
+    def __call__(self, input_ids, attention_mask=None):
+        if attention_mask is None:
+            attention_mask = mx.ones(input_ids.shape, dtype=mx.int32)
+        summed = (input_ids * attention_mask).sum(axis=1, keepdims=True)
+        return SimpleNamespace(text_embeds=summed.astype(mx.float32))
+
+
+class TestTryCompileMaskProbe:
+    """The compile probe must include a traced attention_mask (issue #2447).
+
+    Real requests always carry one (prepare_inputs emits it), so a mask-less
+    probe can pass at load while every real request falls back to eager.
+    These tests run real mx.compile — no mocks — so they fail if the probe
+    stops representing the real request path.
+    """
+
+    def test_mask_branching_model_falls_back_at_load(self, monkeypatch):
+        from omlx.models.embedding import MLXEmbeddingModel
+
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for tightening this probe. Including `attention_mask` makes the load-time compile check exercise the same traced branch as real tokenizer requests, so Qwen3 reports `compiled=False` before serving instead of flipping on the first request, while supported mask-free models still compile. LGTM from my side.

PR #2462, opened 2026-08-01
fix(api): keep completion stream IDs stable

ShiroKSH:
## Summary

- pre-mint one completion ID for each streamed `/v1/completions` response
- reuse that ID for the chunk-mode keepalive, every data frame, and the usage frame
- cover stable IDs in the streaming and keepalive test suites

## Why

OpenAI-compatible stream consumers group frames by completion ID. Per-frame IDs, or an initial `cmpl-keepalive` sentinel, cause strict accumulators to discard later completion chunks.

## Checks

- compiled the updated Python source files

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -1882,6 +1882,15 @@ _KEEPALIVE_COMPLETION_CHUNK = (
     '"model":"keepalive",'
     '"choices":[{"index":0,"text":"","logprobs":null,"finish_reason":null}]}\n\n'
 )
+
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for fixing this. The `/v1/completions` stream now keeps one completion ID across the keepalive, content chunks, and usage frame, and my before/after smoke confirms the strict-accumulator failure is resolved. LGTM. I'll merge this now.

PR #2472, opened 2026-08-02
fix(cache): cover immediate 4k store races

### LINKED ISSUE #2471: Immediate 4K repeat can race async prefix-cache store and re-prefill the full prompt
## Summary

While validating the DeepSeek V4 cache changes on current `main` (`d589c104`), I found a small admission race at the standard 4K benchmark boundary.

A completed 4,096-token request can return its final stream event while the relevant asynchronous `store_cache` job is still finishing. If the same prompt is submitted immediately, the second request performs its prefix lookup before the stored boundary snapshot is visible and prefills all 4,096 tokens again.

## Environment

- Mac Studio M3 Ultra, 256 GB unified memory
- macOS on Apple Silicon
- oMLX `main` at `d589c104`
- `Jundot/DeepSeek-V4-Flash-0731-oQ4e-mtp`
- 4,096 prompt tokens / 128 output tokens
- temperature 0, MTP enabled

## Reproduction

1. Start with a fresh prefix cache.
2. Submit an exact 4,096-token prompt and allow the 128-token stream to finish.
3. Submit the same request immediately after the first stream completes.
4. Inspect `cached_tokens` and the prefix-cache diagnostics.

## Observed

The first response completed while its 3,584-token boundary snapshot was still being persisted. The immediate repeat arrived before that write became visible:

- repeat request: `cached_tokens=0`
- repeat request: re-prefilled `4096/4096` tokens
- closest in-flight store shared `3584/3584` comparable tokens
- the asynchronous store completed roughly half a second later

The existing cache-freshness deferral handles this situation, but its minimum prompt length is currently 8,192 tokens, so the 4K request never enters that path.

## Expected

A highly relevant in-flight 4K boundary store should use the existing non-blocking freshness deferral instead of forcing an avoidable full re-prefill.

## Proposed fix

Lower only `_CACHE_FRESHNESS_WAIT_MIN_PROMPT_TOKENS` from 8,192 to 4,096.

The existing safeguards remain unchanged:

- at least 8,192 common tokens **or** a 30% common-prefix ratio
- matching VLM extra cache keys
- four-second timeout
- scheduler-step deferral rather than blocking active decode or prefill work

## Validation

With the proposed change, the immediate repeat waited approximately 0.525 seconds for the relevant store and then restored seven 512-token blocks:

- cached tokens: `0 -> 3584`
- cold request wall time: 16.446 seconds
- immediate repeat wall time: 5.960 seconds
- deterministic output hash: identical

The timings are diagnostic rather than a formal benchmark because unrelated CI was intermittently active on the host. The cache reuse and output-equivalence results were deterministic.

`tests/test_scheduler.py`: **199 passed**.

I have a focused two-file PR ready with boundary, below-boundary, and weak-prefix regression coverage.

wolfyy970:
Closes #2471.

## What changed

This includes 4K prompts in the scheduler's existing non-blocking cache-freshness deferral.

DeepSeek V4 boundary snapshots retain 3,584 tokens from an exact 4,096-token prompt. The first response can finish just before its asynchronous cache write becomes visible, so an immediate repeat currently races that store and prefills the full prompt again.

The code change lowers only `_CACHE_FRESHNESS_WAIT_MIN_PROMPT_TOKENS` from 8,192 to 4,096. The existing common-prefix ratio, VLM extra-key matching, timeout, and non-blocking scheduling behavior remain unchanged.

## User impact

An immediate repeat of a highly related 4K request can reuse the pending boundary snapshot instead of paying for another full prefill.

In the DeepSeek V4 reproduction:

- before: `cached_tokens=0`, re-prefilled `4096/4096`
- after: waited approximately 0.525 seconds, then restored 3,584 cached tokens
- deterministic output hash remained identical

The observed cold/repeat wall times were 16.446 seconds and 5.960 seconds respectively. These timings are diagnostic because unrelated CI was intermittently active; the deterministic cache-reuse and output-equivalence checks are the correctness evidence.

## Safety boundaries

- Prompts below 4,096 tokens still do not wait.
- A 4K prompt with less than the existing 30% relevance ratio does not wait.
- VLM requests with different extra cache keys do not wait.
- The scheduler does not call `Future.result()`; active decode and prefill work continue while admission is deferred.
- The existing four-second freshness timeout is unchanged.

## Validation

```text
tests/test_scheduler.py
199 passed
```

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -6802,7 +6802,13 @@ class Scheduler:
                 f"{best_p}: stored=...{stored_ctx!r} vs prompt=...{prompt_ctx!r}"
             )
 
-    _CACHE_FRESHNESS_WAIT_MIN_PROMPT_TOKENS = 8192
+    # A 4K prompt is the smallest standard benchmark/workload where a missed
+    # async store is already a multi-second re-prefill.  DeepSeek V4 boundary
+    # snapshots intentionally retain 3584 of 4096 prompt tokens, and their SSD
+    # write can finish just after the first HTTP response is returned.  Include
+    # this workload class in the non-blocking freshness deferral so an immediate
+    # repeated turn waits for that relevant store instead of racing it.
+    _CACHE_FRESHNESS_WAIT_MIN_PROMPT_TOKENS = 4096
     _CACHE_FRESHNESS_WAIT_MIN_COMMON_TOKENS = 8192
     _CACHE_FRESHNESS_WAIT_MIN_PROMPT_RATIO = 0.30
     _CACHE_FRESHNESS_WAIT_TIMEOUT_S = 4.0
--- a/tests/test_scheduler.py
+++ b/tests/test_scheduler.py
@@ -770,14 +770,14 @@ class TestSchedulerAddRequest:
         """Prompts below the freshness minimum should never wait on store_cache."""
         scheduler = Scheduler(model=mock_model, tokenizer=mock_tokenizer)
         scheduler.block_aware_cache = MagicMock()
-        prompt = list(range(7000))
+        prompt = list(range(4095))
 
         future = MagicMock()
         future.done.return_value = False
         future.result.return_value = None
         scheduler._inflight_store_futures["req-prev"] = future
         scheduler._inflight_store_info["req-prev"] = (
-            scheduler_module._InflightStoreInfo(tokens=list(range(6000)))
+            scheduler_module._InflightStoreInfo(tokens=list(range(3584)))
         )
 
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for the focused fix. I reproduced the 4K async-store race on d589c104: the immediate repeat re-prefilled all 4,096 tokens, while this branch waited about 0.59 seconds and restored the 3,584-token boundary snapshot with identical output. The unchanged relevance, VLM-key, non-blocking scheduling, and timeout guards keep the expanded wait bounded, and all 199 scheduler tests pass. LGTM, merging.

PR #2478, opened 2026-08-02
feat：MTPLX / mlx-lm-forge Model Exported MTP (Multi-Head Prediction) Nativ…

HelloTianxu:
Let oMLX light up and truly enable **MTPLX style independent sidecar MTP weights**.

Origin: When loading `philipjohnbasile/Qwen3.6-27B-Fable-Fusion-711-MTPLX-8bit` on oMLX, the MTP button on the admin dashboard is grayed out, with the error:
`Config declares MTP layers but the weight files contain neither mtp.* tensors nor native nextn layers.`

### Verification Results (Structural validation passed)

```
# Directly import three detection functions to run the real model directory
model_type: qwen3_5
_has_mtp_heads        -> True
_is_mtp_compatible    -> True
_checkpoint_has_mtp_weights -> True   ← Was False before the fix

# After applying the patch
apply() -> True
lm load_model / vlm load_model / nn.Module.load_weights are all wrapped
resolve_sidecar -> mtp.safetensors
target prefix (VLM) -> language_model.mtp.
injected language_model.mtp.* count -> 15 (All real sidecar tensors correctly remapped)
leftover bare mtp.* count -> 0
```

**Real decoding validation passed (2026-08-02, M3 Ultra / 27B-8bit)**: Actual MTP testing on `Qwen3.6-27B-Fable-Fusion-711-MTPLX-8bit` showed a decode throughput **+39.6%** (29.6 vs 21.2 tok/s), TPOT −28% (47.5→34.0 ms), and **zero regression** in prefill (pp TPS), TTFT, and base VRAM—proving that the sidecar only provided the `language_model.mtp.*` binding and did not damage the main backbone prefill.
The trend for pp4096/tg128 testing was consistent: tg TPS +33.6%, TPOT −25%, E2E −8.6%, VRAM +0.85 GB, with no change in prefill/TTFT.

Supplement:
1. **Maximum benefit for single requests / low concurrency**: decode +40%, TPOT −28%, E2E −20%, with a cost of only +0.8 GB.
2. **MTP advantage disappears with high concurrency**: When ON/OFF was toggled 2x, the tg TPS was almost equal (40.6 vs 40.7); furthermore, the scaling factor of MTP was worse (1.37x vs 1.92x)—indicating that the fixed overhead of draft validation is amplified by the batch and diluted by intra-batch parallelism. This conforms to the general rule of speculative decoding.
3. **Draft acceptance rate is about ~40%**: Theoretically, 1 extra token limit is 2x, but empirical testing showed 1.40x → acceptance rate ≈ 0.4 (limited by this checkpoint only exporting `mtp_num_hidden_layers=1`).
4. **Usage recommendations**: Interactive / Low QPS (chat, single-user, single-turn agent) → **Enable MTP**; High concurrency serving → **Disable MTP** to save 0.8 GB and let the batch fully utilize the GPU.

### DIFF
--- a/omlx/patches/mtplx_mtp_sidecar.py
+++ b/omlx/patches/mtplx_mtp_sidecar.py
@@ -1,0 +1,259 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Bind the standalone MTP sidecar for MTPLX / mlx-lm-forge exports.
+
+MTPLX-style exports (e.g. ``*MTPLX-8bit``) keep the MTP head in a separate
+``mtp.safetensors`` referenced by ``config.json ->
+mlx_lm_extra_tensors.mtp_file`` (default ``mtp.safetensors``). The sidecar
+stores *bare* ``mtp.*`` keys (MLX-LM text-model convention).
+
+Why this patch is needed, and why two layers:
+
+Layer 1 — VLM path (the model's natural engine):
+    ``mlx_vlm.utils.load_model`` globs ``*.safetensors`` and DOES load the
+    sidecar, but its ``mtp.*`` -> ``language_model.mtp.*`` remap only runs
+    inside ``sanitize_weights``, which is skipped for ``format: mlx``
+    checkpoints (the ``if not is_mlx_format`` guard). MTPLX exports are
+    ``format: mlx``, so the remap is skipped and the bare ``mtp.*`` keys
+    fail to bind to the ``language_model.mtp`` module under strict loading
+    -> "Missing N parameters: language_model.mtp.*".
+    Fix: inject the sidecar into ``nn.Module.load_weights`` (the hook that
+    fires unconditionally before binding, for both loaders and both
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for digging into this, the diagnosis is right: mlx_vlm skips the mtp remap for format mlx checkpoints and mlx_lm never globs mtp.safetensors, and I appreciate the real A/B numbers on the 27B.

I'm going to pass on the runtime binding approach though. MTPLX's forge calibrates the MTP contract per model (hidden variant, concat order, position mode are picked by a per-model probe and stamped into mtplx_runtime.json), so binding a sidecar without validating that contract can run a head under the wrong convention and silently degrade acceptance. Patching nn.Module.load_weights globally and writing a symlink into the model dir at load time are also bigger side effects than I want in the load path.

#2497 fixes the same thing as a one-time import with fail-closed contract validation, so I'll take that direction. I'll fold the mlx_lm_extra_tensors.mtp_file resolution from this PR into it, that part is a real convention I would have missed.

Issue #2443, opened 2026-07-31
SpecPrefill: a prefix-cache hit silently drops the cached prefix (often the system prompt) from the target model's context


JimStenstrom:
## Summary

When a request gets a target-side prefix-cache hit (`cached_tokens > 0`) and SpecPrefill runs, the restored prefix KV is discarded and `prompt[:cached_tokens]` is never prefilled into the target model. The model generates with the first `cached_tokens` tokens — the system prompt, or a large part of it — absent from its context. Nothing errors and generation looks normal, so the corruption is silent. The open PR #2440 already contains the change that fixes this (verified below), so this issue is primarily the bug report and a deterministic repro showing that part of #2440 is a correctness fix, not just a perf feature.

## Reproduction (secret-word probe, temperature 0)

Setup: Qwen3.6-27B (oQ8) target + Qwen3.5-4B (4-bit) draft, `specprefill_threshold=1024`, `specprefill_keep_pct=0.2`. The system prompt states a marker in its first ~60 tokens ("The current release train codename is MOONWREN-7") followed by ~2,300 tokens of policy text. Round A is a short question whose post-system suffix is under the threshold, so SpecPrefill stays off and the prefix cache stores normally. Round B reuses the same system prompt plus a ~6k-token pasted-log user turn ending with "what is the release train codename from your instructions? Reply with the codename only."

| arm | cached | SpecPrefill | codename in answer |
|---|---|---|---|
| warm cache + SpecPrefill, hit inside the system prompt | 2048 | fired | **missing** (700-token budget burned in leaked thinking, `finish_reason=length`) |
| warm cache + SpecPrefill, hit past the system prompt end (larger system prompt, log shows `system 0 full`) | 4096 | fired | **missing** (same texture) |
| warm cache, same messages, per-request `specprefill: false` | 2048 | off | correct (`MOONWREN-7`, clean) |
| cold cache (wiped) + SpecPrefill, same messages | 0 | fired | correct (`MOONWREN-7`, clean) |

Only the cached>0 + SpecPrefill combination fails, and it fails in the same family of symptoms as the agent-client reports: the model attributes the system-prompt policies to the user ("The user provides a long list of deployment policies…") and emits its thinking scaffold as `content`. Both failing arms are leak-free by construction: round B either omits the assistant turn entirely (the cache boundary sits mid-system-prompt, so the prefix still matches) or carries a hardcoded "OK" assistant turn — so no earlier model output can restate the codename into the uncached suffix.

[...]

PR #2484, opened 2026-08-02
fix(scheduler): drop the prefix-cache hit after a sparse-prefill failure

JimStenstrom:
Addresses the residual hazard noted at the end of #2443 (whose main bug was fixed by #2440 as merged).

## Mechanism

Since #2440, `run_specprefill_target_prefill` bases its prefill on the request's restored prefix cache whenever `cached_tokens > 0`, appending KV to it in place. The scheduler's generic sparse-prefill exception handler falls through to normal prefill, but kept `cache_to_use` pointing at that possibly partially-appended cache, with `tokens_to_process` still holding only the post-hit remainder — so the fallback appends the remainder on top of KV that already contains some of those positions. Nothing errors; the request completes with misaligned KV — the same silent-corruption family as #2443 itself, gated on a mid-prefill failure instead of the routine path.

## Change

When the failed request had a cache hit, reset the same state fields the invalid-cache check a few lines up already resets (`cache_to_use`, `prompt_cache`, `cached_tokens`, `remaining_tokens`, `tokens_to_process`), so the fallback prefills the full prompt from scratch, and log the dropped N-token hit — a full re-prefill of a long conversation is a multi-second latency event worth attributing in the field.

## Why this is safe

- The reset mirrors the established `_validate_cache`-failure reset field-for-field; downstream consumers of the reset fields (`RequestOutput` accounting, cache-storage gating, mRoPE seeding, batch homogeneity flags) were walked and all observe a coherent cold-prefill state.
- The gate stays deliberately broader than the mutation it guards: when the exact static-prefix restore path was taken, `request.prompt_cache` was never appended to, so the reset costs only failure-path perf there. Conservative on a rare path.
- An earlier draft also ran `_sync_and_clear_cache` inside the handler; review removed it: the exception's traceback still references the cache at that point (nothing is reclaimable yet), the fallback prefill's own first-chunk clear covers the reclaim, and it would have added an unguarded Metal call in exactly the "Metal may already be in an error state" situation the #435 guard comment warns about.
- The `_PrefillAbortedError` path above does not need the reset: `_do_abort_request` finishes the request and the cache is never re-consumed.

## Out of scope (named)

- The preflight memory guard admitted this request at post-hit-remainder size and is not re-run for the full-prompt fallback. Bounded: the per-chunk guard still enforces the cap during the re-prefill and over-budget prefills route to the existing rejection/requeue path — a late rejection rather than an OOM. Re-running preflight inside the handler is a bigger change than this rare path warrants; happy to follow up if wanted.
- `block_table`/`shared_prefix_blocks` stay set, matching the sibling invalid-cache reset; their remaining consumers only release blocks, which stays correct.
- Pre-existing and adjacent: `_specprefill_active_request_id` is assigned inside the `try` and never cleared by these handlers, so an exception after that point (e.g. from the tracker update) would defer all future requests indefinitely. Separate follow-up candidate.

## Tests

`python -m pytest tests/test_scheduler.py tests/test_specprefill.py tests/test_specprefill_target.py tests/test_specprefill_planning.py -q -m "not slow and not integration"` → 266 passed. The new test drives `_schedule_waiting` with a cache-hit SpecPrefill request whose target prefill raises and asserts the fallback re-prefills the full prompt with no cache attached. Verified to fail against the unmodified scheduler on exactly the defect: pre-fix, `_do_external_prefill` receives only the post-hit remainder (`[3, 4]`) with the stale cache still attached.

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -9055,6 +9055,25 @@ class Scheduler:
                     cleanup_rope(self.model)
                     request.specprefill_indices = None
                     tracker.remove(request.request_id)
+                    if cache_to_use is not None:
+                        # run_specprefill_target_prefill bases its prefill on
+                        # the restored prefix cache when the request had a
+                        # cache hit (#2443), and may have appended partial KV
+                        # to it in place before failing. Re-prefilling the
+                        # post-hit remainder on top would double-write those
+                        # positions, so drop the hit and prefill the full
+                        # prompt from scratch.
+                        logger.info(
+                            f"Request {request.request_id}: dropping "
+                            f"{request.cached_tokens}-token prefix-cache hit "
+                            "after sparse-prefill failure, re-prefilling the "
+                            "full prompt"
+                        )
+                        cache_to_use = None
+                        request.prompt_cache = None
+                        request.cached_tokens = 0
+                        request.remaining_tokens = request.prompt_token_ids
+                        tokens_to_process = request.prompt_token_ids
                     # Fall through to normal prefill
             # External prefill: process tokens[0:N-1] outside BatchGenerator.
             # Only the last token goes to insert() for the first decode step.
--- a/tests/test_scheduler.py
+++ b/tests/test_scheduler.py
@@ -3442,6 +3442,81 @@ class TestSpecPrefillCaches:
         finally:
             scheduler.shutdown()
 
+    def test_sparse_prefill_failure_drops_partially_appended_cache_hit(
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for tracing this failure path. Resetting the cache-hit state before the generic fallback makes the normal prefill rebuild the full prompt and avoids reusing partially appended KV; the regression test reproduces the stale suffix-only behavior on the base commit and passes here. LGTM. I'll merge this now.

PR #2497, opened 2026-08-03
feat(oq): import MTPLX side-car MTP heads into model index

rsnow:
## Summary

MTPLX-packaged checkpoints (e.g. Qwen3.6 Fable-Fusion) ship their MTP head as a standalone `mtp.safetensors` side-car plus an `mtplx_runtime.json` contract, **not** referenced in `model.safetensors.index.json`. As a result:

- `_checkpoint_has_mtp_weights()` returns `False` even though `config` declares `mtp_num_hidden_layers`,
- the admin UI disables the MTP / Lightning toggle with *"Config declares MTP layers but the weight files contain neither mtp.* tensors nor native nextn layers"*,
- and the loader never ingests the head.

This PR adds `import_mtplx_sidecar()` in `omlx/oq.py`, which merges the side-car head into the checkpoint index so the **existing** MTP detector, loader, and Lightning MTP runtime work unchanged afterward. No hot-path/loader changes — it's a convert/import step that normalizes the packaging.

## What it does

- **Fail-closed contract validation** of `mtplx_runtime.json` (`arch_id`, `base_hidden_variant`, `hidden_variant`, `concat_order`, `mtp_position_mode`) against an allowlist before any mutation. Unsupported contracts raise with an operator-readable message.
- **No-dup merge preferred:** points the index at the existing `mtp.safetensors` when logical weight names already match physical tensor keys. **Falls back** to writing a remapped `model-mtp.safetensors` shard when logical keys (`language_model.mtp.*`) differ from physical side-car keys (`mtp.*`), as they do for VLM-shaped checkpoints.
- **bf16 heads stay bf16** — quant entries are only synthesized when the side-car carries quantized (`.scales`) tensors.
- **Payload-audit validator** reports raw vs audited tensor counts and fails closed on schema/flag mismatch.
- **Atomic config/index writes** with timestamped backups. The existing gemma4-assistant and native-donor combine paths are refactored to share the same atomic-write helper and a common quant-entry synthesis helper (`_synthesize_mtp_quant_entries`).

## Testing

- `tests/test_oq_combine.py` extended with MTPLX import coverage: contract-mismatch rejection, no-dup merge when keys align, and dup-fallback when they don't. Full file: **22 passed**.
- Verified end-to-end on an MTPLX Qwen3.6-27B checkpoint: after import, `_checkpoint_has_mtp_weights()` → `True`, admin compat passes, the model loads with the MTP head attached, and the **Lightning MTP backend activates** in the serving path.
- Live serving-path A/B (isolated server instance, greedy decode, 256-token generations): **~1.79× decode throughput** vs autoregressive baseline at draft depth ≥2, with ~72–80% draft acceptance in the scheduler logs. No regressions: an existing MTP-in-index model and a non-MTP model both detect/load unchanged.

## Notes

- Base: `v0.5.4rc2`.
- Scope is packaging/import only — this does **not** add load-time drop-in ingestion of un-imported side-cars (a possible follow-up); the import step is required and documented as such.
- The MTPLX contract match is generic (keyed on the runtime contract, no model-name hardcoding).

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -1231,6 +1231,16 @@ def resolve_output_name(
 
 GEMMA4_ASSISTANT_MTP_PREFIX = "language_model.mtp."
 GEMMA4_ASSISTANT_MTP_SHARD = "model-mtp.safetensors"
+MTPLX_SIDECAR_SHARD = "mtp.safetensors"
+MTPLX_RUNTIME_FILE = "mtplx_runtime.json"
+
+_MTPLX_CONTRACT_ALLOWLIST = {
+    "arch_id": "qwen3-next-mtp",
+    "base_hidden_variant": "post_norm",
+    "hidden_variant": "post_norm",
+    "concat_order": "embedding_hidden",
+    "mtp_position_mode": "local",
+}
 
 
 def validate_gemma4_assistant_pair(
@@ -1315,8 +1325,7 @@ def combine_gemma4_assistant_mtp(
         (assistant_config.get("text_config") or {}).get("num_hidden_layers", 0) or 0
     )
     text_config["mtp_assistant_config"] = assistant_config
-    with open(output / "config.json", "w") as f:
-        json.dump(config, f, indent=2)
+    _atomic_write_json(output / "config.json", config)
 
     logger.info(
         "Merged gemma4 assistant MTP head into %s "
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, this is the right direction: a one-time import that normalizes the packaging beats teaching the load path about the side-car, and the fail-closed contract gate matches how MTPLX forge actually calibrates contracts per model, so I'll merge this.

I'm pushing a follow-up on top for a few things I found while testing against an official Youssofal export: the no-dup mode points the index at mtp.safetensors which mlx_lm never opens (its weight glob is model*.safetensors), the side-car can live at mtp/weights.safetensors declared via mlx_lm_extra_tensors.mtp_file, older exports omit mtp_contract entirely and need the documented defaults filled in before the allowlist check, and the import needs admin wiring to be reachable from the dashboard. Nothing that changes your core design.

PR #2498, opened 2026-08-04
fix(admin): allow disabling idle timeout from the macOS app

nigelks:
## Problem
Clearing the Idle Timeout (as guided, leaving empty) in the macOS app's Performance settings did not disable auto-unload.

## Root cause
The Swift client encoded `idleTimeoutSeconds` as `Int?`; `JSONEncoder` drops `nil` from the body via `encodeIfPresent`, so there was no way to send an explicit `null`. The server additionally rejected `0` / `""` with 422, and the UI copy claimed "Empty = disabled" while empty actually meant "leave unchanged".

## Changes
- **`omlx/admin/routes.py`**: `field_validator` on `idle_timeout_seconds` normalizes `null` / `0` / `""` to `None` (disabled); values must still be >= 60. The existing `model_fields_set` apply path continues to distinguish "not sent" from "sent null".
- **`GlobalSettingsDTO.swift`**: new `PatchOptionalInt` (`nil` = omit key, `.null` = JSON null, `.value(n)` = integer) on the otherwise fully synthesized `GlobalSettingsPatch`.
- **`PerformanceScreenVM.swift`**: empty or `0` = disable (matches the WebUI's **None (disabled)** option), `>= 60` = set; patch field only sent when the value changed, consistent with the other fields.
- **`PerformanceScreen.swift` / `Localizable.xcstrings`**: sublabel and error copy corrected (en + ru): "Empty or 0 = disabled."

## Testing
- `pytest tests/test_admin_api_key.py` — 72 passed
- Manual: set 120 -> save, then clear the field/set to 0 -> models stay loaded
- macOS app builds (`xcodebuild`, Debug)

### DIFF
--- a/apps/omlx-mac/Resources/Localizable.xcstrings
+++ b/apps/omlx-mac/Resources/Localizable.xcstrings
@@ -7445,13 +7445,13 @@
         "en" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "Idle Timeout must be ≥ 60 seconds (or empty to leave unchanged)."
+            "value" : "Idle Timeout must be ≥ 60 seconds (or empty/0 to disable)."
           }
         },
         "ru" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "Тайм-аут простоя должен быть ≥ 60 секунд (или пустым, чтобы оставить без изменений)."
+            "value" : "Тайм-аут простоя должен быть ≥ 60 секунд (или пустым/0 для отключения)."
           }
         }
       }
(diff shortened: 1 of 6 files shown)


jundot:
Thanks for fixing the explicit-null path. I verified that the omit/null/value encoding is correct and that disabling the timeout takes effect live without a restart. LGTM. I'll merge this now, then fold two edge cases into a follow-up: preserving Pydantic's previous coercion while rejecting `false` as zero, and canonicalizing the macOS `0` input to `nil` so the Apply state converges after saving.

nigelks:
Thanks for the quick review and merge!

PR #2501, opened 2026-08-04
fix(deepseek_v4): make prompt rendering append-only when reasoning is retained

jonathan308:
### What
The DSML template chooses a user message's trailing thinking marker by position: the last user message gets `<think>`, all earlier ones get `</think>`. When a new user turn arrives, the previous user message's suffix retroactively flips from `<think>` to `</think>`, so the rendered prompt is not append-only and every cached prefix from that point is invalidated. In agent sessions that is the entire tool loop, re-prefilled on every human turn. Tool-only turns are unaffected, which is why the symptom is independent of speculative decode.

### Impact
On one production day's log, 19 of 76 prefix-divergence events carried this exact signature; the largest destroyed 67,637 tokens of otherwise reusable prefix on a 200k-token prompt (~151k tokens of needless re-prefill in a day). Live A/B after the fix on a 108k-token agent session: first turn 392 s (cold), next turn **18.0 s** — the entire prefix reused.

### Fix
When `drop_thinking=False` (reasoning retained in the transcript), historical user/developer/tool-result suffixes keep the `<think>` opener they were rendered with at generation time, and historical assistant turns render their retained reasoning — reproducing exactly what was rendered when each turn was current, making the transcript append-only. Default rendering (`drop_thinking=True`) is byte-identical before and after across all role/mode combinations, so nothing changes unless the operator opts in via `chat_template_kwargs: {"drop_thinking": false}`.

### Testing
New test file `tests/test_deepseek_v4_template_append_only.py`: append-only across a new human turn; append-only across a tool loop then a human turn (this case caught a missed flip in the tool-result suffix during development); retained reasoning rendered for historical assistant turns; a test documenting the default flip this feature works around; parametrized byte-identical default rendering in both modes.

### DIFF
--- a/omlx/patches/deepseek_v4/chat_template_v4.py
+++ b/omlx/patches/deepseek_v4/chat_template_v4.py
@@ -178,11 +178,38 @@ def find_last_user_index(messages: List[Dict[str, Any]]) -> int:
     return last_user_index
 
 
+def _user_thinking_suffix(
+    index: int,
+    last_user_idx: int,
+    thinking_mode: str,
+    drop_thinking: bool,
+) -> str:
+    """Marker that follows a user/developer message.
+
+    The current turn opens thinking. Historical turns normally close it,
+    because their reasoning is dropped from the transcript — but that makes
+    rendering position-dependent: when a new user message arrives, the
+    previous user message's suffix flips from ``<think>`` to ``</think>``
+    and every cached prefix from that point on is invalidated.
+
+    When reasoning is retained (``drop_thinking=False``), the historical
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for tracking this down. Keeping the original `<think>` opener and rendering retained historical reasoning makes `drop_thinking=false` append-only while leaving the default path unchanged; the DeepSeek V4 template suite passes on current main. LGTM, merging.

sansaeedanan-rgb (commented):
yeekate777@gmail.com comment saws25411@gmail.com confirm sansaeedanan@gmail.com

PR #2502, opened 2026-08-04
perf(deepseek_v4): guard the indexer fallback past int32 limits and make it audible

jonathan308:
### What
When the native `glm_moe_dsa` kernels are not built (the default for a source install), the DSA lightning indexer falls back to a pure-MLX chain that materializes a `(B, 64, L, ctx/4)` fp32 score tensor per sparse layer per prefill chunk. Two problems:

1. **Correctness**: at ~256k context that tensor crosses 2^31 elements, where int32 kernel indexing silently zeroes the tail — corrupting top-512 index selection with no error. This patch tiles the pooled axis past a safe element budget (2^30) with the largest tile that fits, so every matmul stays under the limit at any context length. Below the budget the single-matmul path is untouched: unconditional tiling measured ~1.2x slower at 273k, so tiling is applied only where it is a correctness requirement.
2. **Silence**: `has_symbol()` returning False leaves no log line, so the ~4x slower fallback runs invisibly (we found it via a 45-minute benchmark, not the logs). Warn once when the native kernels are missing, mirroring the qwen35 patches.

Also: the relu → scale → head-weight → head-sum chain is fused with `mx.compile` (verified exact), and a transient exception in the sparse-attention native path no longer sets the module-wide disable flag permanently.

### Testing
New tests: tiled vs untiled top-k selection bit-identical (including straddling tile boundaries), fused reduce exact vs reference, element guard stays under int32, fallback warning fires exactly once.

### DIFF
--- a/omlx/patches/deepseek_v4/deepseek_v4_model.py
+++ b/omlx/patches/deepseek_v4/deepseek_v4_model.py
@@ -1,5 +1,6 @@
 # Copyright © 2026 Apple Inc.
 
+import logging
 import math
 from dataclasses import dataclass, field
 from functools import lru_cache, partial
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for tracking this down. I verified that the tiled MLX fallback preserves the untiled top-k result, including a 7,001-token DeepSeek V4 smoke test, and the missing-kernel warning fires only once. LGTM. I'll merge this and fold the fallback-warning regression test and formatting cleanup into a small follow-up.

PR #2510, opened 2026-08-04
fix(deepseek_v4): always emit the generation anchor for system-final prompts

jonathan308:
### What
The 0.5.7 official encoder appends the `<|Assistant|><think>` generation
anchor only after a user/developer-final message. Two request shapes render
with NO anchor:

1. a system-only conversation (clients that send a bare instruction, e.g.
   session-title or summary utility calls), and
2. a trailing system message after the last user turn (workspace/context
   notes appended by agent clients).

With no anchor the model continues the document instead of starting an
assistant turn, and generation does not terminate until max_tokens.

### Impact
Verified live on 0.5.7 with shape 1: a 225-token title prompt generated
21,614 tokens over 24 minutes before it was killed. After this change the
same prompt completes in 163 tokens with `finish_reason: stop`.

Shape 2 is verified statically — `apply_chat_template(..., add_generation_prompt=True)`
returns a prompt ending in the system text with no anchor; both shapes occur
in normal OpenAI-API traffic.

### Fix
In `apply_chat_template`, when `add_generation_prompt` is set and the encoded
prompt does not already end with an anchor, append `<|Assistant|>` + the
mode-correct think token. Prompts that already carry an anchor are unchanged,
so user/developer-final rendering is byte-identical (guarded by a count==1
regression test).

### Testing
New `tests/test_deepseek_v4_generation_anchor.py` (6 tests: both shapes,
chat-mode anchor, no double-append, `add_generation_prompt=False`,
`continue_final_message`). Full suite passes rebased on current main.

### DIFF
--- a/omlx/patches/deepseek_v4/chat_template_v4.py
+++ b/omlx/patches/deepseek_v4/chat_template_v4.py
@@ -845,6 +845,23 @@ def apply_chat_template(
     if not add_generation_prompt:
         out = out.removesuffix(ASSISTANT_SP_TOKEN + thinking_start_token)
         out = out.removesuffix(ASSISTANT_SP_TOKEN + thinking_end_token)
+    elif not out.endswith(
+        (
+            ASSISTANT_SP_TOKEN + thinking_start_token,
+            ASSISTANT_SP_TOKEN + thinking_end_token,
+        )
+    ):
+        # encode_messages only emits the generation anchor after a
+        # user/developer-final message. A system-final conversation (e.g. a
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for fixing the missing generation anchor for DeepSeek V4 system-only prompts. The normal chat path is covered, and I found only one narrow interaction with the reference template's `task` mode, where the generic anchor can be appended after the task transition. I'll preserve that mode and add focused title/action coverage in a small follow-up on `main`; LGTM, merging.

PR #2512, opened 2026-08-04
fix(engine_core): don't re-abort requests already finished or pending abort

jonathan308:
### What
`abort_all_requests()` aborts every collector it sees, but deferred aborts
only execute at the next `step()`/chunk boundary — potentially many
seconds away under the same memory pressure that triggered the abort. The
enforcer calls this once per second while pressure persists, so the same
request is re-aborted every poll: duplicate `[Error: ...]` frames reach
streaming clients, and the returned count reports stale aborts as fresh
progress, hiding a stuck scheduler from the enforcer's escalation logic.
Observed in production: the same request id re-aborted 14+ consecutive
times.

### Fix
Filter to requests not in `_finished_at` and not already in the
scheduler's `_pending_abort_ids`; the count now reflects newly aborted
requests only.

### Testing
Full suite passes rebased on current main.

### DIFF
--- a/omlx/engine_core.py
+++ b/omlx/engine_core.py
@@ -695,7 +695,25 @@ class EngineCore:
         """
         from .utils.proc_memory import get_phys_footprint
 
-        request_ids = list(self._output_collectors.keys())
+        # Only abort requests not already finished or pending abort. The
+        # enforcer calls this once per poll (1s) while pressure persists; a
+        # deferred abort can take many seconds to drain (it executes at the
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for tracing the repeated bulk-abort path. Filtering finished and already-pending request IDs makes consecutive calls idempotent and prevents duplicate error frames; I verified that the second call reports zero new aborts. I'll add a focused regression test and fold in the small iterator lint cleanup on `main`; LGTM, merging.

PR #2513, opened 2026-08-04
fix(deepseek_v4): make tool-call turn re-rendering byte-stable with model output

JimStenstrom:
## Summary

- DeepSeek V4's reference decoder consumes `"\n\n<｜DSML｜tool_calls"` as one literal stop token — the `"\n\n"` separator before a tool-call block is envelope markup, not content. omlx's streaming filter matched only the bare marker, so streamed content deltas kept the model's trailing `"\n\n"`, and re-rendering that turn through the chat template (which adds its own canonical separator) produced `"\n\n\n\n"` — bytes the model never emitted.
- Fix (parser-side, reworked per review): `ToolCallStreamFilter` registers the separator-inclusive envelope variant ahead of the bare marker, so the separator is consumed at the streaming parser boundary — the same place the reference decoder consumes it. The chat template is untouched and submitted message content is never modified.
- Result: content accumulated from streamed deltas re-renders byte-identically to the raw emission, prompts stay append-only across tool hops, and streaming clients converge with the non-streaming path (which already strips the separator span in `_parse_tool_calls_impl`).

## Why

The reference's own decoder reads tool-call turns with `tool_calls_start_token = "\n\n<｜DSML｜tool_calls"` as a literal stop string (`encoding_dsv4.py:726`, consumed via `str.find` in `_read_until_stop`), so its parsed `content` never carries the separator, and re-encoding (content + the encoder's canonical `"\n\n"` + block) reproduces the original bytes exactly. That the stop token embeds `"\n\n"` is proof this is the model's *trained* grammar: the stop token cannot match the emission at all otherwise.

omlx's streaming filter matched the bare `"<｜DSML｜tool_calls>"` marker, so the `"\n\n"` the model emits before the block passed through as ordinary content deltas. A client that accumulates deltas stores `"...\n\n"` in the turn's `content`, and every re-render of that turn gains a second separator:

```
raw emission:  ' files in /tmp for you.\n\n<｜DSML｜tool_calls'
re-render:     ' files in /tmp for you.\n\n\n\n<｜DSML｜tool_calls'
```

Scope of the claim (narrowed per review): this is a round-trip and prompt-grammar consistency fix. Re-rendered prompts stop containing `"\n\n\n\n"` runs the model's trained grammar never produces, and streaming-accumulated vs non-streaming client histories converge on identical bytes for the same turn. No cache-hit improvement is claimed on current main — as the review traced, the scheduler's prompt-boundary store means later requests already match the stored re-rendered prompt regardless of separator width.

## Changes

- `omlx/api/tool_calling.py`, `ToolCallStreamFilter.__init__`: when the tokenizer-defined pair is the DSML pair (exact match on both `tool_call_start` and `tool_call_end`), additionally register `("\n\n<｜DSML｜tool_calls>", "</｜DSML｜tool_calls>")` ahead of the bare pair. `_find_start_envelope`'s earliest-match rule prefers it whenever the separator is present; the bare pair remains as fallback for emissions that omit the separator. The existing holdback (`_partial_suffix_len`) covers deltas that split the separator or the marker across feeds with no further changes.
- `omlx/api/tool_calling.py`, `_should_drop_tail_at_finish`: newline-only tails are exempted from the strict-mode drop (same shape as the existing MiniMax `"]"` exemption). They are held back only as a potential envelope prefix; at end of stream with no envelope following they are literal content and are flushed by `finish()`.
- New keyword-only flag `consume_dsml_separator=True` on the filter; `omlx/server.py`'s three `_thinking_filter` sites and `sanitize_tool_call_markup` (whose callers all sanitize thinking-channel text) pass `False`: reasoning text never precedes a DSML block with the separator, holding a reasoning delta's trailing newline would otherwise flush it only at `finish()` — after the thinking channel closed (a late `reasoning_content` delta, a Responses-API delta after `output_item.done`, or a phantom trailing thinking block on the Anthropic path) — and the opt-out keeps streamed and buffered reasoning byte-identical. The content/tool filters and the DeepSeek V4 adapter session keep the default.
- `omlx/server.py`, Anthropic finish-flush: the existing phantom-text-block guard on the delta path (drop pure-whitespace deltas before a text block opens when tools are requested) is mirrored onto the `tool_filter.finish()` flush. The filter can now legitimately flush held newlines there, and without the mirror a thinking-only turn ending in `"\n\n"` would open a text block containing only whitespace — exactly what the delta-path guard exists to prevent.

## Why this is safe

- The chat template is untouched: encoding stays byte-faithful to the vendored reference, and user-submitted message content is never modified (review concern 2).
- Envelope-scoped: only the DSML pair gets the separator-inclusive variant; every other model family's markers behave exactly as before.
- No content loss: newlines followed by prose flush as soon as the next delta disproves the envelope (`test_newlines_before_prose_pass_through`); newlines at end of stream flush at `finish()` (`test_trailing_newlines_without_tool_call_survive_finish`); an unclosed envelope's recovery candidate now faithfully includes the withheld separator — bytes are only ever reclassified, never dropped.
- Longer newline runs: the envelope match consumes exactly the final two newlines (`str.find` semantics, identical to the reference decoder) — `"a.\n\n\n"` before a block yields content `"a.\n"`, matching the reference's own parse (`test_only_final_two_newlines_belong_to_envelope`).
- Sibling sweep: all three `_marker_pairs` readers accounted for — `_find_start_envelope` (earliest match prefers the separator-inclusive variant), `_partial_suffix_len` (holdback automatic for the new marker), `_should_drop_tail_at_finish` (the newline exemption above). The non-streaming path (`_parse_tool_calls_impl`) already strips the separator span and is unchanged. `DeepSeekV4OutputParserSession` routes its stream and visible deltas through this same filter class, so the adapter path is covered by the same change.

Stated trade-offs (all narrow, listed so they are decisions rather than discoveries):

- Delta timing, not content: V4 deltas ending in a newline are held one feed until the next delta resolves whether an envelope follows. Chunk boundaries shift; accumulated content is unchanged.
- A truncated stream that dies mid-marker (`"text\n\n<｜DSML"`) now drops the held separator along with the partial-marker tail under the existing strict-mode finish drop — previously the `"\n\n"` would have been emitted. Consistent with strict mode's intent; only reachable on malformed/truncated streams; pinned as intended by `test_truncated_open_marker_drops_held_separator_at_finish`.
- On the Anthropic path, a leading `"\n\n"` before V4 prose was previously dropped by the whitespace-only-delta guard as its own delta; held newlines now merge with the following prose delta and stream (more faithful to the model's bytes, noted as a behavior delta).
- The byte-stability guarantee is for the model's trained `"\n\n"` grammar. An out-of-grammar separator (a lone `"\n"` before the block) passes through as content, and the template's canonical separator still renders on top — the same as before this change.
- For prose that continues *after* a closed tool-call block, streamed deltas now sum without the pre-block separator while the non-streaming cleanup (`parse_tool_calls`' bare-marker regex + `.strip()`) keeps interior newlines — a divergence only reachable on engines without the V4 parser session, which stops the row at the first block end. Not fixed here to keep the diff narrow.

Known residual, out of scope here: when retained reasoning is re-wrapped into `content` for non-native clients (`api/utils.py`'s `<think>` re-wrap) and the visible text is empty, the stored content still ends in `"\n\n"`. That producer is not a parser boundary and would need its own fix — happy to follow up separately.

## Tests

- `tests/test_deepseek_v4_patch.py::TestChatTemplateV4::test_streamed_tool_call_turn_rerender_is_byte_stable` — the parser-to-rerender regression: the raw emission is fed through `ToolCallStreamFilter` character-by-character (so the separator and both DSML markers split across feeds), and the accumulated content plus structured `tool_calls`, re-rendered through the unchanged template, must equal `prompt + raw_emission + eos_token` exactly, contain no `"\n\n\n"` run, and be extended verbatim by the following turn's render (`startswith` — append-only).
- `tests/test_tool_calling.py::TestToolCallStreamFilterDsmlSeparator` — thirteen filter-level tests: separator consumed (single feed and split across feeds), bare-marker fallback, exact-two-newline consumption on longer runs, single- and double-newline hold/release, multiple envelopes with prose between, end-of-stream newline preservation, the truncated-tail strict-mode drop, the unclosed-envelope recovery candidate (separator included, nothing duplicated or lost), the thinking-channel opt-out (both halves: trailing newlines flush in place, and a separator-preceded envelope keeps the separator visible while still suppressing), and `sanitize_tool_call_markup` parity with the streamed reasoning deltas.

```
.venv/bin/python -m pytest tests/test_tool_calling.py tests/test_deepseek_v4_patch.py tests/test_output_parser.py tests/test_deepseek_v4_template_append_only.py tests/test_server.py tests/test_scheduler.py -q -m "not slow and not integration"
670 passed
```

Also verified: with `tool_calling.py` and `server.py` restored to main's versions, ten of the new tests fail while four still pass — those four (bare-marker fallback, newline-before-prose passthrough, the opt-out envelope case, and sanitize parity) are behavior-preservation pins, not regression tests. Branch rebased onto current `main`.

### DIFF
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -1470,7 +1470,10 @@ def sanitize_tool_call_markup(text: str, tokenizer: Any) -> str:
     if not text:
         return ""
 
-    stream_filter = ToolCallStreamFilter(tokenizer)
+    # Every caller sanitizes thinking-channel text; keep it byte-identical
+    # with the streamed reasoning deltas, which do not consume DeepSeek
+    # V4's separator either.
+    stream_filter = ToolCallStreamFilter(tokenizer, consume_dsml_separator=False)
     cleaned = stream_filter.feed(text)
     cleaned += stream_filter.finish()
     return cleaned.strip()
@@ -1598,9 +1601,15 @@ class ToolCallStreamFilter:
     Args:
         tokenizer: The model's tokenizer. Uses tokenizer-defined
             ``tool_call_start`` when available.
+        consume_dsml_separator: Treat DeepSeek V4's ``"\n\n"`` separator as
+            part of the DSML tool-call envelope, matching the reference
+            decoder's stop token. Disable for filters watching a channel
+            that never contains a separator-prefixed tool-call block (the
+            thinking channel), where holding trailing newlines would flush
+            them only after the channel closed.
     """
 
-    def __init__(self, tokenizer: Any):
+    def __init__(self, tokenizer: Any, *, consume_dsml_separator: bool = True):
         marker = getattr(tokenizer, "tool_call_start", None)
         marker_end = getattr(tokenizer, "tool_call_end", None)
         # Normalize None-like values but preserve empty strings.
@@ -1621,6 +1630,16 @@ class ToolCallStreamFilter:
                 # One-sided markers (e.g. Mistral "[TOOL_CALLS]" with no
                 # end marker): suppress everything after the start marker.
                 self._suppress_after_markers.append(marker)
+        # DeepSeek V4's reference decoder consumes "\n\n<｜DSML｜tool_calls"
+        # as one literal stop token, so the separator belongs to the envelope,
+        # not to content. Register the separator-inclusive variant so the
+        # earliest-match rule consumes it when present; the bare pair stays as
+        # fallback for emissions that omit the separator.
+        is_dsml_tool_marker = (
+            marker == "<｜DSML｜tool_calls>" and marker_end == "</｜DSML｜tool_calls>"
+        )
+        if is_dsml_tool_marker and consume_dsml_separator:
+            self._marker_pairs.insert(0, ("\n\n" + marker, marker_end))
         # Gemma 4 can emit a bare close token outside a matched tool-call
         # envelope. Do not apply this to XML-style closers like </tool_call>,
         # which may appear as literal prose.
@@ -1803,6 +1822,11 @@ class ToolCallStreamFilter:
                 # bracket at end-of-stream is much more likely to be literal
                 # prose than an incomplete MiniMax control marker.
                 if tail == "]":
+                    continue
+                # Newline-only tails are held back only because of the
+                # separator-inclusive DeepSeek V4 marker; at end-of-stream
+                # with no envelope following they are literal content.
+                if not tail.strip("\n"):
                     continue
                 return True
 
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -4397,7 +4397,12 @@ async def stream_chat_completion(
     stream_content = True
     if has_tools:
         _content_filter = ToolCallStreamFilter(engine.tokenizer)
-        _thinking_filter = ToolCallStreamFilter(engine.tokenizer)
+        # The thinking channel never contains a separator-prefixed DSML
+        # block; holding trailing newlines would flush them as a late
+        # reasoning delta after the channel closed.
+        _thinking_filter = ToolCallStreamFilter(
+            engine.tokenizer, consume_dsml_separator=False
+        )
         if _content_filter.active:
             tool_filter = _content_filter
             thinking_filter = _thinking_filter
@@ -4824,7 +4829,12 @@ async def stream_anthropic_messages(
     thinking_filter = None
     if has_tools:
         _content_filter = ToolCallStreamFilter(engine.tokenizer)
-        _thinking_filter = ToolCallStreamFilter(engine.tokenizer)
+        # The thinking channel never contains a separator-prefixed DSML
+        # block; holding trailing newlines would flush them as a late
+        # reasoning delta after the channel closed.
+        _thinking_filter = ToolCallStreamFilter(
+            engine.tokenizer, consume_dsml_separator=False
+        )
         if _content_filter.active:
             tool_filter = _content_filter
             thinking_filter = _thinking_filter
@@ -4903,14 +4913,15 @@ async def stream_anthropic_messages(
                         content_delta = tool_filter.feed(content_delta)
                     if content_delta:
                         # When tools are requested AND we haven't yet opened
-                        # a text block, drop pure-whitespace deltas. Models
-                        # often emit a leading newline around <tool_call>
-                        # envelopes that tool_filter passes through
-                        # (whitespace isn't part of the envelope markers).
-                        # Without this guard, the `\n` opens a text block
-                        # that then holds only whitespace — surfacing as
-                        # a phantom empty-ish text block before the
-                        # tool_use blocks.
+                        # a text block, drop pure-whitespace deltas. Most
+                        # models emit a leading newline around <tool_call>
+                        # envelopes that tool_filter passes through (their
+                        # whitespace isn't part of the envelope markers;
+                        # DeepSeek V4's separator is, and the filter
+                        # consumes it itself). Without this guard, the `\n`
+                        # opens a text block that then holds only
+                        # whitespace — surfacing as a phantom empty-ish
+                        # text block before the tool_use blocks.
                         if (
                             not text_block_started
                             and kwargs.get("tools")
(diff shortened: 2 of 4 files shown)


jundot:
Thanks for the detailed investigation. I traced this against the cache path on current main and found two concerns.

**1. The current scheduler does not cache the generated tool-call turn**

1. Request 1 sends prompt `P1`, and the model generates tool-call turn `A1`.
2. The DSML parser stops generation at the tool-call end marker.
3. That terminal response has no `prompt_cache`, so `_prepare_prompt_boundary_cache_store()` stores only `P1` up to a block boundary. It explicitly skips `A1`.
4. Request 2 must therefore prefill `A1` regardless of whether its separator contains two or four newlines.
5. After Request 2, the rerendered prompt is stored. Later requests already render the same separator sequence as that stored prompt, even before this change.

The added test proves that the raw model output and rerendered history can be byte-identical, but its `P1 + A1 + eos_token` sequence is not the sequence the current scheduler stores. I could not confirm a longer cache hit from this diff alone. Could you either add a scheduler-level cache test that demonstrates the improvement on current main or narrow the cache claims to round-trip and prompt grammar consistency?

**2. The normalization is applied too broadly**

`removesuffix("\n\n")` runs for every submitted assistant tool-call message, including user-supplied history whose content intentionally ends with a blank line. The official encoder preserves that content and then adds its own DSML separator.

Could you consume the separator at the DeepSeek V4 streaming output-parser boundary instead and add a parser-to-rerender regression test? That would normalize model-generated output without changing arbitrary historical content.

JimStenstrom:
Thanks for the careful trace — you're right on both counts, and the PR is reworked accordingly.

On the cache claim: I re-traced `_prepare_prompt_boundary_cache_store()` after your comment and confirmed your sequence — the terminal tool-call response stores prompt-boundary tokens only, and later requests match the stored re-rendered prompt regardless of separator width. The cache numbers came from a separate streaming prototype that splices raw emissions into its cache (where re-render/raw-emission byte-stability is load-bearing); they don't apply to current main and I've removed them. The body now claims round-trip and prompt-grammar consistency only, as you suggested: re-rendered prompts stop containing `"\n\n\n\n"` runs the trained grammar never produces, and streaming vs non-streaming clients converge on identical stored bytes for the same turn.

On the normalization: agreed, and moved to the boundary you pointed at. The template change is gone — the encoder stays byte-faithful to the vendored reference and never touches submitted content. Instead `ToolCallStreamFilter` registers the separator-inclusive envelope variant `"\n\n<｜DSML｜tool_calls>"` (the reference decoder's own stop-token shape) ahead of the bare marker, so the separator is consumed at the parser boundary exactly where the reference consumes it; the bare marker stays as fallback for emissions that omit it. Newline-only tails at end of stream are exempted from the strict-mode tail drop, so an ordinary turn that ends in newlines without a tool call loses nothing. Thinking-channel consumers (the three `_thinking_filter` sites and `sanitize_tool_call_markup`) opt out via a keyword-only flag — reasoning text never precedes a DSML block with the separator, holding a reasoning delta's trailing newline would otherwise flush it after the thinking channel closed, and the opt-out keeps streamed and buffered reasoning byte-identical. The Anthropic finish-flush also gets the same phantom-text-block guard as its delta path, since the filter can now legitimately flush held newlines there.

The parser-to-rerender regression test you asked for is in (`test_streamed_tool_call_turn_rerender_is_byte_stable`): content accumulated from the filter's own deltas — fed character-by-character so the separator and markers split across feeds — re-rendered through the unchanged template, must reproduce `prompt + raw_emission + eos` byte-for-byte and stay append-only across the following turn. Thirteen filter-level tests pin the mechanics (split feeds, no-separator fallback, longer newline runs, multi-envelope streams, end-of-stream newline preservation, truncated-stream strict-mode behavior, recovery-candidate byte accounting, the thinking-channel opt-out).

One residual named in the body as out of scope: the retained-reasoning re-wrap (`api/utils.py`) can also leave stored content ending in `"\n\n"` when the visible text is empty — that producer isn't a parser boundary; happy to follow up separately if you'd like.

Branch rebased onto current main; 670 passed across the six touched/adjacent suites (`test_tool_calling`, `test_deepseek_v4_patch`, `test_output_parser`, `test_deepseek_v4_template_append_only`, `test_server`, `test_scheduler`), and ten of the new tests fail with the change reverted (four are behavior-preservation pins that pass both ways).

jundot:
Thanks for reworking this at the streaming parser boundary. Consuming the DeepSeek V4 DSML separator there matches the reference decoder without mutating submitted chat history, and the Anthropic flush guard avoids a phantom whitespace block. The current-main merged test sweep passed with 815 tests passing and 4 skipped; LGTM, merging.

PR #2519, opened 2026-08-04
fix(codex): scope CLI configuration to launch

### LINKED ISSUE #2517: Use process-scoped overrides for Codex CLI launches

`omlx launch codex` currently rewrites `~/.codex/config.toml` with the local model and provider. Those values remain after the oMLX-launched CLI exits, so a later normal Codex session can still point at oMLX. The launch path also resolves the selected model's effective `max_context_window` without passing it to Codex.

This is especially easy to hit after switching between local and OpenAI models, changing a per-model context setting, or applying a context benchmark result.

For the Codex CLI integration, use Codex's launch-time `-c` overrides for the oMLX provider definition, reasoning effort, and effective `model_context_window`. Continue passing the selected model with `-m`.

The overrides should live only in the CLI process launched by oMLX. The user's config file and settings such as `[agents]`, MCP servers, sandbox policy, and normal OpenAI provider should remain untouched.

Codex currently derives automatic compaction at 90% of the resolved context window, so oMLX should leave `model_auto_compact_token_limit` unset instead of copying that policy into the integration.

The integration could back up the config file and restore it when Codex exits. That is vulnerable to crashes, overlapping launches, and edits made while Codex is running. Users can also maintain a separate Codex profile by hand, but oMLX already has all the resolved model information needed to build the launch overrides.

**Additional context**

This request is deliberately limited to `omlx launch codex`. Codex currently accepts `-c` after the `app` subcommand but does not forward those overrides to the macOS desktop process, so the same small change is not correct for `omlx launch codex_app`.

Draft PR #2470 proposes the complete desktop and hybrid-routing solution. It keeps Codex's native model slots, adapts the model catalogue with the local context window, and routes only the chosen slot through a process-scoped proxy. This issue is a small CLI safety fix while that broader draft is reviewed, not a replacement for it.

wolfyy970:
Closes #2517.

## Why

`omlx launch codex` currently writes its model and provider into `~/.codex/config.toml`. Those values survive the CLI process and can affect the next normal Codex launch. The integration also has the selected model's effective context window but does not give it to Codex.

This is a narrow CLI fix while draft #2470 develops the full desktop and hybrid-routing integration. It does not change `codex_app.py` or try to duplicate the interceptor.

## What changed

- pass the oMLX provider through Codex's process-scoped `-c` overrides
- pass the selected model's effective `model_context_window`
- keep the selected model on Codex's existing `-m` argument
- preserve explicit oMLX reasoning metadata for the launched process
- leave `model_auto_compact_token_limit` unset so Codex derives its current 90% threshold
- stop the CLI launch path from calling the persistent config writer

The user's normal Codex config remains available underneath these overrides, including `[agents]`, MCP servers, sandbox settings, and other preferences.

## Verification

- `uv run pytest tests/test_integrations.py -q` (107 passed)
- `uv run pytest tests/ -m "not slow and not integration" -q` (7,796 passed, 64 skipped, 71 deselected)
- `uv run ruff check omlx/integrations/codex.py tests/test_integrations.py`
- parsed the generated provider and context overrides with Codex CLI 0.146.0-alpha.9.2
- confirmed the SHA-1 of `~/.codex/config.toml` was unchanged before and after the CLI validation

## Relationship to #2470

#2470 is the stronger long-term design for Codex Desktop. It preserves native local and cloud model slots, adapts the model catalogue with the real local context window, and handles the desktop process lifecycle through a scoped proxy.

This PR only makes the existing custom-provider CLI launch safer and context-aware. Codex's `app` command currently accepts `-c` syntactically but does not forward those overrides to the macOS app process, so `codex_app` is intentionally out of scope.

### DIFF
--- a/omlx/integrations/codex.py
+++ b/omlx/integrations/codex.py
@@ -1,7 +1,9 @@
+# SPDX-License-Identifier: Apache-2.0
 """Codex (OpenAI Codex CLI) integration."""
 
 from __future__ import annotations
 
+import json
 import os
 import re
 import shutil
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for narrowing this to the CLI path. I verified that the provider and context window land in Codex's session config layer while the user's config file remains untouched. The focused integration suite and Ruff pass. LGTM, merging.

PR #2524, opened 2026-08-05
feat: add Ling 3.0 Flash support

scaryrawr:
## Summary

- register a vendored `bailing_hybrid` mlx-lm model patch so Ling 3.0 Flash works with the official mlx-lm dependency
- adapt recurrent KDA caches to oMLX batching and prefix-cache contracts
- replace token-by-token convolution and KDA prefill loops with vectorized MLX convolution and the fused gated-delta kernel
- preserve all variable-length `ArraysCache` slots when reconstructing N-tuple prefix snapshots

## Performance

Using `Ling-3.0-flash-mxfp4` (64.64 GB) through the OpenAI-compatible API:

- cold ~3,000-token prefill improved from roughly 25s to 4.14s
- repeated 3,021-token requests now restore a 2,048-token prefix and complete in roughly 1.18s instead of 3.5s
- cached and cold deterministic outputs matched

## Validation

- focused Ling, cache-handler, and prefix-cache tests pass against upstream dependencies
- `import omlx.server` succeeds
- official mlx-lm pin remains unchanged

### DIFF
--- a/omlx/cache/prefix_cache.py
+++ b/omlx/cache/prefix_cache.py
@@ -2898,9 +2898,16 @@ class BlockAwarePrefixCache(CacheManager):
                         last_block_meta_states
                     ):
                         meta_state = last_block_meta_states[layer_idx]
-                    cache = marker_handler.deserialize_state(
-                        tuple(elements), meta_state
-                    )
+                    if marker_handler.is_variable_length_state():
+                        cache = marker_handler.reconstruct_cache(
+                            {"states": list(elements)},
+                            meta_state,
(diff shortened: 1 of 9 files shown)


jundot:
Thanks for the initial Ling 3.0 Flash implementation. I used both of your commits as the foundation for #2526 and preserved their original authorship.

The follow-up adds actual FP8 loading, oQ/oQe conversion, and variable-length batch and SSD cache validation. The completed implementation is now merged, so I’m closing this PR as superseded. Thanks again for the contribution.

PR #2528, opened 2026-08-05
Handle Ling protocol output

scaryrawr:
Ling can occasionally sample its role and XML tool-call protocol markers into generated output, causing tags such as `</role>` or `<arg_value>` to leak into client-visible content.

This change suppresses `<role>`/`</role>` markers and statefully filters fragmented `<tool_call>`, `<arg_key>`, and `<arg_value>` envelopes while retaining the raw protocol stream for structured parsing and token/cache accounting.

It also passes request-scoped tool definitions through generation so Ling tool calls use the declared argument schemas, unregistered tool names are rejected, and protocol output is not interpreted as a tool call when no tools were supplied. Streaming and non-streaming calls remain available through the normal structured API response.


jundot:
Thanks for fixing the Ling protocol leakage and making tool-call parsing request-aware. This branch now conflicts with the Gemma 4 prefilled-thought session setup that landed on main afterward; I verified that preserving both `create_session_with_tools()` and `notify_prefilled_thought()` resolves it and passes the related tests. I'll apply that small integration fix to this branch, rerun CI, and merge it.

PR #2534, opened 2026-08-05
feat: support Ling 3.0 Flash FP4 checkpoints

scaryrawr:
## Summary

Adds support for official [ling-3.0-flash-fp4](https://huggingface.co/inclusionAI/Ling-3.0-flash-fp4). Previous PR was tested with a custom mxfp4 conversion, where the official has mixed fp4 and fp8 that's not supported by it.

- load Ling routed experts directly through MLX native MXFP4 using the checkpoint packed bytes and E8M0 scales
- preserve the existing FP8 block conversion for attention, shared experts, and other non-routed projections
- add per-layer MXFP4 quantization overrides and strict mixed-checkpoint regression coverage

## Validation

- `UV_PYTHON=python3.12 uv run --python python3.12 pytest tests/test_bailing_hybrid_patch.py -q` (24 passed)
- loaded `inclusionAI/Ling-3.0-flash-fp4` locally with oMLX at 64.29 GB runtime memory
- generated the expected response `LING FP4 WORKS` through `/v1/chat/completions`
- validated a real `pi` agent loop against oMLX on port 8000; Ling successfully issued `read` and `bash` tool calls and used their results

### DIFF
--- a/omlx/patches/bailing_hybrid/bailing_hybrid_model.py
+++ b/omlx/patches/bailing_hybrid/bailing_hybrid_model.py
@@ -736,12 +736,17 @@ class Model(nn.Module):
         return self._convert_fp8_block_weights(weights)
 
     def _convert_fp8_block_weights(self, weights):
-        """Convert Ling's block-scaled E4M3 tensors to 8-bit affine MLX.
+        """Convert Ling's published FP8 and MXFP4 tensor layouts for MLX.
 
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for adding support for the official mixed FP8/MXFP4 checkpoint. I verified the published tensor layout against this loader path and reproduced the strict-load regression covered here. LGTM. I'll merge this now.

PR #2540, opened 2026-08-06
fix(i18n): localize Runtime Cache Observability panel in web UI

LXD-8:
## Summary
The **Runtime Cache Observability** panel in the admin dashboard (`omlx/admin/templates/dashboard/_status.html`) was hardcoded English that bypassed the `t()` i18n function, so it stayed untranslated in every locale (including English fallback showing raw key names).

This PR:
- Wraps all 21 user-facing strings in `t('status.runtime_cache.*')`
- Adds the matching keys to **all 9 web locales** (`en/zh/zh-TW/ja/ko/fr/es/pt-BR/ru`) plus the macOS `Localizable.xcstrings` catalog
- Non-zh locales keep the **original English strings**, so their UI is unchanged — only `zh-Hans` gets the Chinese translations (RAM 缓存 / SSD 缓存 / 运行时缓存可观测性 / 前缀命中率 / RAM 命中率 / 前缀驱逐 / RAM 驱逐 / 模型 / 块大小 / 索引块数 / 子块缓存 / SSD 文件数 / SSD 大小 / RAM 条目 / RAM 大小 / 清除 RAM 缓存 / 活动 base_path / 生效 ssd_cache_dir / 响应状态目录, and a localized note)

## Why all locales
`t()` falls back to the **raw key** when a key is missing (not English), so every locale file must contain the keys — otherwise non-zh languages would display bare `status.runtime_cache.observability` strings.

## Test plan
- [x] `jinja2` parses and renders the template with the real zh dict; no raw keys leak
- [x] All 9 JSON catalogs + xcstrings are valid and contain the 21 keys
- [x] Non-zh 8 locales equal the en source (UI unchanged)
- [x] xcstrings diff is insert-only (+21, Xcode formatting preserved)

Co-Authored-By: NirvanaCh7 <noreply@users.noreply.github.com>


jundot:
Thanks for wiring the Runtime Cache Observability panel into i18n. The template and catalogs merge cleanly with current main, and the relevant admin tests pass. LGTM. I'll merge this now.

PR #2543, opened 2026-08-06
feat(turboquant): add opt-in mid-prefill KV conversion

Collinw24:
## Summary

This PR adds an **experimental, opt-in, default-off capacity rescue** for TurboQuant KV cache during prefill.

When an eligible dense prefill approaches oMLX's configured memory guard, the scheduler can reclaim same-engine headroom, convert populated full-attention KV layers once, and continue the same request with TurboQuant instead of failing immediately. `turboquant_mid_prefill` is only available under the existing TurboQuant parent setting and is `false` by default. With the setting off, the path is inert.

This is a capacity feature, not a speed feature. The measurements below are bounded observations for one disclosed model, prompt set, machine, and memory envelope; they are not a universal speed or quality claim.

## Architecture and safety

- The engine claims a process-exclusive Metal lane before enabling the feature. A sole model can reclaim and convert in the same request without a useless no-victim pause or restart; a temporary ownership conflict is retryable after the conflicting engine is unloaded.
- Both external and chunked prefill use the same request-scoped phase machine. Each request gets at most one mid-prefill conversion attempt. An async LRU retry is allowed only when a concrete idle external victim exists, and at most once.
- Conversion reserves a bounded incremental peak before mutation, converts one layer at a time in bounded 8,192-token slices, validates the completed cache representation, then synchronizes and releases dense state. Rotating, recurrent/state-array, and other pass-through caches retain their native representation.
- Unsupported layouts do not receive speculative relief. Route preflight defers a dense rejection only after model cache construction has confirmed an eligible dense layout. Incompatible or partial conversion state is discarded and the request fails closed.
- The runtime never mutates source model tensors. The validation harness hashes every source `.safetensors` shard before and after each completed command.
- Runtime, validator, and evidence files are separated: the PR runtime branch contains neither the validation harness nor evidence payloads.

The final Q5 campaign used a fixed, disclosed 45 GiB child ceiling and 3 GiB host floor matching this machine’s configured Metal envelope. Limits were not changed during a command, mandatory telemetry remained fail-closed, and the product’s own memory guard remained active.

## Five maintainer gates

1. **Real quality validation:** dense, ordinary post-prefill Q8/Q4, and mid-prefill Q8/Q4 were run on identical real-model prompts. The compact tables report official MRCR retrieval and 256-token teacher-forced mean NLL; all five modes in each table also produced the same greedy response hash.
2. **Controlled performance comparison:** each table uses one model snapshot, one rendered/tokenized prompt, a fixed 2,048-token chunk policy, and the same 45 GiB/3 GiB envelope. It reports total prefill, same-boundary suffix, teacher-forced decode, peak child physical footprint, and minimum host headroom. These point measurements do not establish a general speedup.
3. **Retryable exclusivity conflicts:** a failed process-exclusive claim cleans up the candidate engine without caching a model-load failure. The regression loads model A, observes model B's retryable conflict, unloads A, and then loads B successfully.
4. **Confirmed preflight eligibility:** deferral requires a constructed, confirmed-convertible cache layout. The streaming regression verifies an unsupported layout returns the normal HTTP 400 before the response generator starts.
5. **Current-main integration:** the exact tested/UI runtime was rebased patch-identically onto the current version-only upstream base. Runtime, validation-only, and public-evidence lineage are recorded separately below.

Relevant final-runtime tests: [bounded conversion and fail-closed behavior](https://github.com/Collinw24/omlx/blob/1b970fe3906a0f70f67a13a39beb221f46881b25/tests/test_turboquant_mid_prefill.py), [retryable process ownership](https://github.com/Collinw24/omlx/blob/1b970fe3906a0f70f67a13a39beb221f46881b25/tests/test_engine_pool.py), [confirmed preflight eligibility](https://github.com/Collinw24/omlx/blob/1b970fe3906a0f70f67a13a39beb221f46881b25/tests/test_engine_preflight.py), and [pre-stream HTTP 400](https://github.com/Collinw24/omlx/blob/1b970fe3906a0f70f67a13a39beb221f46881b25/tests/integration/test_server_endpoints.py).

## Controlled real-model evidence

Model: `Basher17/Qwen3.6-27B-heretic-v2-oQ5e-mtp`, snapshot `9a88654f6e3294f12344619d7078afc92fd41fe4`. Dataset: `openai/mrcr` two-needle data, revision `f4c69fae7cf81f7ca26b9fee34b392a50f6b8a1d`. GiB values are binary. "Suffix" is the same-boundary second half of the prompt; "decode" is the 256-token teacher-forced replay.

The 32K and 64K matrices are validator-owned **direct model loops**. They do not exercise scheduler fairness, and they are not exact-final-head runs.

### 32K controlled row

MRCR row 104, 32,045 prompt tokens. Direct-loop runtime/validator: [`dc771572`](https://github.com/Collinw24/omlx/commit/dc7715729c8d65fe9bd668b4eedb7d58e03fedb7) / [`ae6c3261`](https://github.com/Collinw24/omlx/commit/ae6c326173f05888573c848437bc8bc3e4fb444f).

| Mode | Retrieval | Mean NLL | Prefill t/s | Suffix t/s | Decode t/s | Peak GiB | Min host GiB |
|---|---:|---:|---:|---:|---:|---:|---:|
| dense | 1.000000 | 1.847744e-6 | 106.17 | 79.27 | 8.87 | 27.64 | 7.61 |
| ordinary Q8 | 1.000000 | 1.847744e-6 | 115.78 | 107.37 | 8.23 | 27.64 | 6.42 |
| ordinary Q4 | 1.000000 | 2.026558e-6 | 146.99 | 141.55 | 11.56 | 27.86 | 11.28 |
| mid Q8 | 1.000000 | 1.907349e-6 | 124.86 | 106.09 | 10.24 | 25.91 | 11.32 |
| mid Q4 | 1.000000 | 2.071261e-6 | 136.43 | 117.39 | 11.35 | 24.97 | 9.09 |

All five modes produced the same greedy response hash. Source tensors were unchanged.

### 64K controlled row

MRCR row 311, 63,118 prompt tokens. Direct-loop runtime/validator: [`839efe11`](https://github.com/Collinw24/omlx/commit/839efe11e2a66cf03eaf4c61385d0b3e95e47a2f) / [`4689f468`](https://github.com/Collinw24/omlx/commit/4689f46880fd22e5f1b76d5bf94c803811f382cd).

| Mode | Retrieval | Mean NLL | Prefill t/s | Suffix t/s | Decode t/s | Peak GiB | Min host GiB |
|---|---:|---:|---:|---:|---:|---:|---:|
| dense | 0.999419 | 4.559755e-6 | 94.07 | 83.65 | 9.07 | 34.45 | 6.10 |
| ordinary Q8 | 0.999419 | 4.313886e-6 | 125.73 | 108.76 | 8.10 | 40.15 | 3.55 |
| ordinary Q4 | 0.999419 | 5.461276e-6 | 118.71 | 97.65 | 8.45 | 34.49 | 6.47 |
| mid Q8 | 0.999419 | 4.410744e-6 | 99.70 | 77.49 | 8.58 | 33.03 | 3.79 |
| mid Q4 | 0.999419 | 5.200505e-6 | 115.12 | 89.54 | 11.19 | 27.75 | 9.10 |

All five modes produced the same greedy response hash. Source tensors were unchanged.

### Organic 131K scheduler result

The production scheduler completed one 131,071-token Q8 request with a normal pressure trigger: no forced trigger, no external victim, and no restart. It triggered at 63,488 tokens, converted 15 layers in 120 bounded slices, retained one configured dense skip-last layer, and completed conversion in 2.718 seconds. Total prefill was 1,708.13 seconds (76.73 token/s), with 56.83 token/s after the trigger. The 256-token teacher-forced replay measured mean NLL 1.762810 and perplexity 5.828791 at 6.83 token/s. Peak child physical footprint was 35.80 GiB; minimum host headroom was 3.73 GiB. There was one prefill attempt, zero eviction retries, zero eviction pauses, and unchanged source tensors.

This organic result is exact only for runtime [`839efe11`](https://github.com/Collinw24/omlx/commit/839efe11e2a66cf03eaf4c61385d0b3e95e47a2f), paired with validator [`4689f468`](https://github.com/Collinw24/omlx/commit/4689f46880fd22e5f1b76d5bf94c803811f382cd). It is not an exact run of the later tested/UI head or the final published rebase.

No exact 262K command ran, and this PR makes no exact 262K claim.

## Verification

The exact tested/UI runtime is [`c4d118ae9380f9cc25e2bb3f0c44b840fce6037d`](https://github.com/Collinw24/omlx/commit/c4d118ae9380f9cc25e2bb3f0c44b840fce6037d):

- **9,386 passed** in the Python suite.
- **174 passed** in the focused scheduler/persistence/reclaim overlap set.
- **22 passed** in the affected macOS set.
- Exact-tested-head production UI verification passed: fresh Alpine mount, modal open through the production row action, parent/child visibility, independent child toggle, PUT save payload, persisted settings, reopen state, and empty page/console error probes.

The validation-only harness has **43 passing tests** at [the final validation head](https://github.com/Collinw24/omlx/commit/551b5717179499475e9a799a7cce92f220802853). That head is based directly on the final runtime and changes only the validation script and its dedicated tests.

The final published runtime is [`1b970fe3906a0f70f67a13a39beb221f46881b25`](https://github.com/Collinw24/omlx/commit/1b970fe3906a0f70f67a13a39beb221f46881b25), a patch-identical rebase of `c4d118ae` onto upstream [`ac32ea24569b785c366e14525453c2a66d3523d2`](https://github.com/jundot/omlx/commit/ac32ea24569b785c366e14525453c2a66d3523d2), whose intervening upstream change is version-only. No direct-loop, organic-model, or UI result is relabeled as an exact `1b970fe3` run.

## Limits and risks

- This remains experimental and default-off. Quantization error and performance vary by model, prompt, bit width, context length, and machine; the two controlled rows are evidence for those rows, not a general guarantee.
- The direct-loop tables do not exercise production scheduler fairness. The organic scheduler result covers one Q8 request on runtime `839efe11` only.
- The feature requires process-exclusive Metal ownership. Other engines may need to unload first; the conflict is retryable, not silently cached as a terminal model failure.
- Only confirmed-compatible cache layouts convert. Unsupported layouts, an unbounded conversion, a failed conversion, or insufficient post-conversion headroom fail closed rather than bypassing the memory guard.
- One request gets one conversion attempt and at most one concrete-victim eviction retry. This rescue does not guarantee that an otherwise oversized request will complete.
- No exact 262K result is claimed.

## Public provenance

- Exact tested/UI runtime: [`c4d118ae9380f9cc25e2bb3f0c44b840fce6037d`](https://github.com/Collinw24/omlx/commit/c4d118ae9380f9cc25e2bb3f0c44b840fce6037d).
- Final patch-identical published runtime: [`1b970fe3906a0f70f67a13a39beb221f46881b25`](https://github.com/Collinw24/omlx/commit/1b970fe3906a0f70f67a13a39beb221f46881b25), rebased on version-only upstream `ac32ea24569b785c366e14525453c2a66d3523d2`.
- 32K direct-loop runtime/validator: `dc771572` / `ae6c3261`.
- 64K direct-loop and organic scheduler runtime/validator: `839efe11` / `4689f468`.
- Final validation-only head: [immutable commit](https://github.com/Collinw24/omlx/commit/551b5717179499475e9a799a7cce92f220802853).
- Sanitized public evidence: [immutable evidence commit](https://github.com/Collinw24/omlx/commit/e88ed344d01a7782d2d11ad93a93052abe6fcfe7) and [evidence root](https://github.com/Collinw24/omlx/tree/e88ed344d01a7782d2d11ad93a93052abe6fcfe7/validation/evidence/pr2543-20260813).

The runtime PR branch remains free of validator and evidence files.


jundot:
Thanks for taking this on. I reviewed the memory math, conversion path, and tests again. I think the feature has a valid use case: it can let a long-context request finish when dense prefill would otherwise hit the memory guard.

The KV memory reduction is real. With the Qwen geometry used in this PR, the full-attention KV payload at 131,071 tokens is about 8.00 GiB dense, 4.28 GiB with Q8, and 2.40 GiB with Q4. That saves 3.72 GiB or 5.60 GiB respectively. Model weights and the 48 recurrent cache states do not shrink, so these percentages do not apply to the total process footprint.

I also ran a synthetic MLX check on an M3 Ultra. Dense K/V storage fell from 32.0 MiB to 16.1 MiB with Q8 and 8.1 MiB with Q4. In the same check, quantized attention was about 1.85 times slower than dense attention. The relative attention-output difference was about 1.3% for Q8 and 10.0% for Q4. This is not an end-to-end model benchmark, but it confirms that the memory, speed, and precision tradeoff is real.

I found two validation gaps and two runtime issues that should be addressed before another review.

### 1. Add real quality validation

Mid-prefill conversion changes more than the cache format. After the trigger, the rest of the prompt attends to quantized history, so later hidden states and K/V values can diverge from dense prefill.

The existing equality test uses `_AppendModel`, which appends fixed tensors without letting attention affect later K/V values. It proves that sliced conversion matches ordinary conversion for the same tensors, but not that a real model preserves quality. Eight coherent output tokens are also only a smoke test. The long runs use Q8 even though the configured default is Q4.

Please compare dense prefill, ordinary post-prefill TurboQuant, and mid-prefill TurboQuant on the same real model and prompts. Q8 and Q4 both need coverage. A token-level KL or perplexity result plus a long-context retrieval test would be enough to show whether the change is safe for the intended workload. The [TurboQuant paper](https://arxiv.org/abs/2504.19874) and a later [vLLM study](https://vllm-project.github.io/2026/05/11/turboquant.html) report different results depending on bit width and workload, so oMLX needs its own measurements.

### 2. Add a controlled performance comparison

The reported conversion pause of about 1.1 seconds is small. The main cost is the quantized attention path used for the rest of the prefill.

The 131K run completed, but it used a forced trigger, a 64-token chunk, and took 2,753.20 seconds. There is no matching dense or ordinary TurboQuant baseline, so the slowdown caused by mid-prefill is still unknown.

Please add an A/B comparison at 32K and 64K using the same model, prompt, chunk size, and memory limits. Report peak process footprint, total prefill time, prefill throughput after the trigger, and decode throughput. The 131K case should also be repeated with a normal pressure trigger rather than a forced one.

### 3. Keep exclusivity conflicts retryable

If another engine prevents the process-exclusive claim, `_load_engine()` cleans up the new engine but then records the conflict as a cached model-load failure. Unloading the other model does not help because the next request is rejected before another load is attempted.

Please handle this as a retryable busy or capability error instead of calling `_mark_load_failure()`. Add a test that fails to load model B while model A is present, unloads A, and then successfully retries B.

### 4. Require confirmed eligibility before deferring preflight

Route-level preflight runs before a prompt cache exists. `_can_defer_mid_prefill_preflight()` currently treats unknown conversion eligibility as eligible. An unsupported cache layout can therefore skip the pre-streaming HTTP 400 and fail only after the streaming response has started.

Please defer preflight only when conversion eligibility is confirmed. Add a streaming test with an unsupported cache layout and verify that the request receives the normal HTTP 400 before the response starts.

The branch also needs a rebase against current `main`, since the engine, memory, and preflight paths currently conflict.

Once these items are addressed, I can review this as an experimental, default-off path for requests that would otherwise fail under memory pressure. I would keep the PR in draft until then.

Collinw24:
Reconstruction checkpoint (PR remains draft during final-head model validation):

- Rebuilt the runtime cleanly on current upstream main e2c05d80; draft head is now 260352b4.
- Preserved the prior reviewed/research/staging heads on immutable archive branches before replacement.
- Final validation-only branch: Collinw24/omlx@validation/pr2543-260352b4-20260812, head be3eda26. It is based directly on runtime 260352b4 and changes only scripts/validate_turboquant_mid_prefill.py plus its dedicated test file.
- Independent reviews found no remaining scoped P0-P3 after the final scheduler cancellation and progress-accounting fixes.
- Focused Python checks, 43 validator tests, and 22 affected macOS tests pass. The first full-suite pass had one prefix-cache hot-tier assertion failure; its isolated rerun and all 122 prefix-cache tests passed, so one unchanged full-suite retry is running.
- Live browser verification is blocked by an existing current-main model-settings modal error (mtp_compatibility_reason is undefined during initial Alpine evaluation), before this PR's child toggle renders. Static settings/API tests and macOS settings tests pass; I have not expanded this PR to fix that unrelated UI bug.

No exact-262K command will run. The PR stays draft until the bounded 32K/64K/131071 final-head evidence is frozen and all readiness gates are evaluated.

Collinw24:
@jundot, all five requested gates are now addressed, with runtime, validator, and model-run lineage kept separate:

1. **Real quality validation:** the disclosed Q5 snapshot ran dense, ordinary Q8/Q4, and mid-prefill Q8/Q4 on identical 32K and 64K MRCR prompts. Each row had one greedy response hash across all five modes; the replacement body reports official retrieval and 256-token teacher-forced mean NLL. [Immutable evidence commit](https://github.com/Collinw24/omlx/commit/e88ed344d01a7782d2d11ad93a93052abe6fcfe7) · [evidence root](https://github.com/Collinw24/omlx/tree/e88ed344d01a7782d2d11ad93a93052abe6fcfe7/validation/evidence/pr2543-20260813)
2. **Controlled performance comparison:** the same rows used one fixed 2,048-token chunk policy and reported total prefill, same-boundary suffix, teacher-forced decode, peak child footprint, and minimum host headroom. A separate production-scheduler request triggered organically at 63,488 tokens and completed 131,071 tokens in one attempt with no victim, restart, or forced trigger. The direct loops do not exercise scheduler fairness and are not presented as exact-final-head runs; the organic result is exact only for runtime `839efe11`. [32K/64K/131K records](https://github.com/Collinw24/omlx/tree/e88ed344d01a7782d2d11ad93a93052abe6fcfe7/validation/evidence/pr2543-20260813)
3. **Retryable exclusivity:** a failed process-exclusive claim cleans up without poisoning the model entry; the regression unloads model A and then successfully retries model B. [Final-runtime regression](https://github.com/Collinw24/omlx/blob/1b970fe3906a0f70f67a13a39beb221f46881b25/tests/test_engine_pool.py) · [validation record](https://github.com/Collinw24/omlx/commit/551b5717179499475e9a799a7cce92f220802853)
4. **Confirmed preflight eligibility:** deferral now requires a cache layout confirmed convertible after cache construction. The streaming regression proves an unsupported layout returns HTTP 400 before its generator starts. [Eligibility tests](https://github.com/Collinw24/omlx/blob/1b970fe3906a0f70f67a13a39beb221f46881b25/tests/test_engine_preflight.py) · [HTTP regression](https://github.com/Collinw24/omlx/blob/1b970fe3906a0f70f67a13a39beb221f46881b25/tests/integration/test_server_endpoints.py)
5. **Current-main rebase:** the exact tested/UI runtime is `c4d118ae9380f9cc25e2bb3f0c44b840fce6037d`; the final published runtime is the patch-identical `1b970fe3906a0f70f67a13a39beb221f46881b25` rebase on version-only upstream `ac32ea24569b785c366e14525453c2a66d3523d2`. The recorded checks are 9,386 Python, 174 focused overlap, 43 validator, 22 macOS, plus production UI verification at the exact tested head. [Final runtime](https://github.com/Collinw24/omlx/commit/1b970fe3906a0f70f67a13a39beb221f46881b25) · [final validation head](https://github.com/Collinw24/omlx/commit/551b5717179499475e9a799a7cce92f220802853)

The final Q5 campaign used a fixed, disclosed 45 GiB child ceiling and 3 GiB host floor matching this machine’s configured Metal envelope. Limits were not changed during a command, mandatory telemetry remained fail-closed, and the product’s own memory guard remained active.

For model evidence, the 32K direct-loop runtime/validator pair is `dc771572` / `ae6c3261`; the 64K direct-loop and organic pair is `839efe11` / `4689f468`. No exact 262K command ran or is claimed. The runtime branch contains no validator or evidence files.

Collinw24:
Current-main rebase update (2026-08-15):

- Force-pushed final head `31e13a433f9984bdb5d29566428b09ae1ebf21ee`, based directly on `origin/main` `91741020b76b5fcef0579b2035eb7a2059b693c5`.
- Preserved both sides of the three integration conflicts: upstream GDN sidecar decoding plus fail-closed Darwin `F_NOCACHE` propagation; upstream VLM `tools` forwarding plus request-local `prefill_eviction_callback_attempted`; and both VLM test imports/coverage sets.
- Refreshed once more after seven additional upstream commits landed; that final replay was conflict-free.
- The range-diff preserves the prior 19 PR commits. One scoped cleanup removes 33 lines of a brittle, broad regex source scan while retaining the focused modal initialization and guarded-MTPLX regressions.

Non-inference validation only—the active local inference loop was left untouched:

- `146 passed` across settings/profile/admin regressions and the six mock-only VLM forwarding contracts.
- All 45 changed Python files compile with the project Python.
- Ruff reports zero findings on PR-added lines.
- `git diff --check origin/main...HEAD` passes.
- Independent static review of all 67 changed paths found no P0-P3 issue.

No model was loaded, no generation or TurboQuant numeric/inference test ran, and the running app was not restarted. The earlier model evidence remains tied to its disclosed exact heads; this update does not relabel it as final-head inference evidence.

PR #2544, opened 2026-08-07
fix(tool-calling): keep tool calls whose arguments contain a close marker (#2507)

### LINKED ISSUE #2507: Tool call dropped when an argument contains a literal </tool_call>
The paired-marker extraction in `omlx/api/tool_calling.py` uses a non-greedy match, so the first literal `</tool_call>` **inside an argument string** ends the match early. The JSON is then truncated, parsing fails, and the whole call is dropped silently.

`_parse_tool_calls_impl` (~line 1287):

```python
pattern = rf"{start_escaped}(.*?){end_escaped}"
matches = re.findall(pattern, text, re.DOTALL)
```

`_parse_xml_tool_calls` (line 303) has the same shape.

### Reproduction without a model

The "model output" below is built with `json.dumps`, so it is valid by construction — anything lost after this point is lost by the parser, not by the model:

```python
import json
from pathlib import Path
from mlx_lm import tokenizer_utils

TOOLS = [{"type": "function", "function": {
    "name": "note_write",
    "parameters": {"type": "object",
                   "properties": {"title": {"type": "string"},
                                  "body": {"type": "string"}},
                   "required": ["title", "body"]}}}]

tok = tokenizer_utils.load(Path("<any Qwen3.5/3.6 checkpoint>"))
payload = "text with a literal </tool_call> inside"
inner = json.dumps({"name": "note_write",
                    "arguments": {"title": "t", "body": payload}},
                   ensure_ascii=False)
raw = "<tool_call>\n%s\n</tool_call>" % inner

print(parse_tool_calls(raw, tok, tools=TOOLS))   # -> (text, None): call is gone
```

Matching server log:

```
Native tool parser failed (JSONDecodeError: Unterminated string ...)
and XML fallback could not recover. Dropping match:
'\n{"name": "note_write", "arguments": {"title": "t", "body": "text with a literal '
```

I ran six payloads with control markers inside an argument, on three parser paths (`json_tools`, `qwen3_coder`, generic XML fallback) and two different models. Identical result every time: **4 of 6 intact, 2 dropped**. The two failures are exactly the payloads containing `</tool_call>`; `<think>`, `[TOOL_CALLS]` and `<|im_end|>` survive.

### Why it matters

Coding agents routinely write text containing these markers — prompt files, parser code, documentation. The call disappears with no error surfaced to the client, so from the outside it looks like the model refused to act.

### Suggested fix

Instead of stopping at the first end marker, advance to the next one until the payload parses, falling back to the first match if none do — so behaviour is never worse than today:

```python
def _extract(text, start, end, parses):
    out, pos = [], 0
    while (s := text.find(start, pos)) >= 0:
        body_start = s + len(start)
        first = chosen = None
        e = text.find(end, body_start)
        while e >= 0:
            cand = text[body_start:e]
            if first is None:
                first = (cand, e)
            if parses(cand):
                chosen = (cand, e)
                break
            e = text.find(end, e + len(end))
        chosen = chosen or first
        if chosen is None:
            break
        out.append(chosen[0])
        pos = chosen[1] + len(end)
    return out
```

Measured with this change: **6 of 6 intact** on both models. Well-formed outputs are unaffected — the first end marker already parses, so the loop exits immediately.

### Environment

oMLX 0.5.5, macOS, Apple silicon (M5 Max). The relevant code looks unchanged in the v0.5.4...v0.5.7 diff.

richgoodson:
Fixes #2507.

## The problem

The paired-marker extraction matched with a non-greedy `start(.*?)end`, so the first literal `</tool_call>` inside an argument string ended the match early. The JSON was truncated, the native parser raised, and the XML fallback re-ran the same truncating pattern so it could not recover either. The call was dropped and its raw markup leaked into the response.

The reporter's model-free repro reproduces on `198c5ce9`:

```
('<tool_call>\n{"name": "note_write", "arguments": {"title": "t", "body": "text with a lite...', None)
```

## Streaming has the same defect

`ToolCallStreamFilter` ended the suppressed envelope on the same embedded marker, so the tail of the tool call was emitted as **visible content mid-stream**. Feeding the reporter's output in 7-character chunks on `198c5ce9`:

```
leaked to the user: ' inside"}}\n</tool_call>'
```

This is not in the original report, so it is worth a separate look during review.

## The fix

Payload boundaries are located by span scanning instead of a non-greedy regex. When the payload is JSON, `raw_decode` reports where the value actually ends, which ignores marker text sitting inside a string.

It is deliberately conservative. The JSON boundary is used only when it proves the payload runs past the first close marker **and** a later close marker exists. Every other case keeps the historical first-match behaviour, so only outputs that are broken today change.

Applied to the native paired-marker path, the XML and Hermes fallbacks, and the marker-stripping cleanups, so a recovered call is also removed whole instead of leaving its tail behind as content.

For streaming, the filter carries brace and string state across chunks and examines each character once. A per-chunk re-decode of the accumulated buffer would be quadratic on untrusted model output, which this module already guards against elsewhere.

## The qwen3_coder XML dialect

Qwen3.5/3.6 builds using the `qwen3_coder` parser wrap XML rather than JSON, so the JSON boundary cannot bound them:

```
<tool_call>
<function=note_write>
<parameter=body>
text with a literal </tool_call> inside
</parameter>
</function>
</tool_call>
```

This is what `Qwen3.6-35B-A3B-4bit` actually emits, so it is worth covering. The structural anchor here is that a real envelope ends with `</function>` followed by the close marker, and that its parameter elements balance. Requiring both skips copies embedded in values, including a value containing the entire terminator sequence, while still ending at the first call when several are concatenated. Candidates are capped so a value stuffed with fake terminators cannot make the balance check quadratic.

Two smaller truncations in the same dialect come with it: the function element now closes at the last `</function>` rather than the first, and a parameter value ends at the `</parameter>` followed by a sibling or by the end of the element. Scanning advances past each value, so a literal `<parameter=` inside a value no longer starts a spurious argument.

The stream filter learns the dialect too. It has only a prefix, so it cannot balance elements and instead uses the local rule that the close marker must follow `</function>`. That is a slightly weaker guarantee than the non-streaming path: a value containing the full `</function>` + close-marker sequence would still end the stream envelope early, whereas the non-streaming parser resolves it.

## Scope, and what this does not change

Dialects that are neither JSON nor `qwen3_coder` keep their existing behaviour by design: GLM `<arg_key>`/`<arg_value>`, Hermes brackets, and `_parse_namespaced_tool_calls`.

## Testing

- 22 new tests across `TestToolCallMarkerInArguments` (JSON) and `TestQwen3CoderMarkerInArguments` (XML).
- Verified red before and green after on every affected path by running the assertions against unpatched `198c5ce9`: native, XML fallback, Hermes and streaming for JSON, and 7 of the 11 XML tests against the JSON-only version.
- The reporter's literal repro now returns the call with the body intact and no leaked content.
- Full suite: 7892 passed, 67 skipped. That is clean main's 7870 plus the 22 added here.
- Guard tests for the failure modes found while reviewing this change: scan time stays linear on output repeating the close marker, on XML stuffed with fake terminator sequences, and when a single chunk carries many markers; boundary detection returns rather than raising on deeply nested input, since `raw_decode` recurses per nesting level and raises `RecursionError` rather than a decode error.

### Live check

Served `Qwen3.6-35B-A3B-4bit` and asked it to write a note whose body contains the literal tag. The round trip is correct with the patch, but it does not exercise the bug: `</tool_call>` is a dedicated special token for that tokenizer, so across four prompt variants the model omitted it from the argument rather than emitting it as text, and the raw text reaching the parser was already missing the tag. That matches the reporter's choice of a model-free repro, and it suggests the realistic trigger is a model copying content that contains the marker rather than producing it spontaneously.

This is also how I found that the model emits the `qwen3_coder` XML dialect rather than JSON, which is what prompted covering it above.

## Note on an existing crash, not addressed here

While bounding the new boundary scan I found that deeply nested output crashes tool parsing on current main, independent of this change:

```python
parse_tool_calls("<tool_call>" + "[" * 100000 + "</tool_call>", tok)
# RecursionError escapes parse_tool_calls
```

`json.loads` in `_parse_xml_tool_calls` recurses per nesting level and `RecursionError` is not caught by the surrounding `except (json.JSONDecodeError, ...)`. I confirmed it behaves identically before and after this PR, so nothing here makes it worse. Filed separately rather than widened into this change.

### DIFF
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -2350,31 +2350,55 @@ class ToolCallStreamFilter:
             self._clear_pending_envelope()
             return ""
 
-        if self._suppressing_until is not None:
+        # An envelope still suppressing at EOF means the payload scan never
+        # confirmed where the value ends.  A literal close marker is then the
+        # best evidence available, so trust it rather than withholding to EOF:
+        # prose the model resumed after the envelope survives, as it did before
+        # the #2507 span scanning.  The LAST occurrence is the split point,
+        # since earlier ones can sit inside the truncated payload.
+        #
+        # Re-feeding the tail can re-enter suppression under a *different*
+        # marker pair, hence the loop.  It terminates because each pass cuts
+        # the text down to a strict suffix, and repeats at most once per
+        # distinct marker pair.
+        recovered = ""
+        while self._suppressing_until is not None:
+            marker = self._suppressing_until
             self._pending_envelope_parts.append(self._buffer)
             candidate = "".join(self._pending_envelope_parts)
             start_marker = self._pending_start_marker or "<unknown>"
             self._buffer = ""
             self._suppressing_until = None
             self._clear_pending_envelope()
-            if candidate:
-                self._recovery_candidate = candidate
-                logger.warning(
-                    "Unclosed tool-call envelope at end of stream; "
-                    "withheld %d characters are available for content recovery "
-                    "(start_marker=%.80r)",
-                    len(candidate),
-                    start_marker,
-                )
-            return ""
+
+            close_idx = (
+                -1
+                if marker == "__suppress_permanently__"
+                else candidate.rfind(marker)
+            )
+            if close_idx < 0:
+                if candidate:
+                    self._recovery_candidate = candidate
+                    logger.warning(
+                        "Unclosed tool-call envelope at end of stream; "
+                        "withheld %d characters are available for content "
+                        "recovery (start_marker=%.80r)",
+                        len(candidate),
+                        start_marker,
+                    )
+                return recovered
+
+            recovered += self.feed(candidate[close_idx + len(marker) :])
+            if self._suppressing:
+                return recovered
 
         keep = self._partial_suffix_len(self._buffer)
         if keep >= len(self._buffer):
             tail = self._buffer
             self._buffer = ""
             if self._should_drop_tail_at_finish(tail):
-                return ""
-            return tail
+                return recovered
+            return recovered + tail
 
         if keep:
             buf = self._buffer[:-keep]
@@ -2387,7 +2411,7 @@ class ToolCallStreamFilter:
         for close in self._stray_close_markers:
             if close in buf:
                 buf = buf.replace(close, "")
-        return buf
+        return recovered + buf
 
 
 def convert_tools_for_template(tools: Optional[List]) -> Optional[List[dict]]:
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -323,6 +323,112 @@ def _json_value_end(text: str, start: int) -> Optional[int]:
     return end
 
 
+_XML_FUNCTION_OPEN = "<function="
+_XML_FUNCTION_CLOSE = "</function>"
+_XML_PARAMETER_OPEN = "<parameter="
+_XML_PARAMETER_CLOSE = "</parameter>"
+# Candidate envelope ends examined before giving up. Each candidate costs a
+# balance count over the payload, so this keeps hostile output linear.
+_XML_MAX_END_CANDIDATES = 32
+# Trailing payload context the stream filter keeps so a `</function>` split
+# across chunks is still visible when the close marker arrives.
+_XML_TAIL_KEEP = 64
+
+
+def _skip_ws(text: str, idx: int) -> int:
+    """First index at/after ``idx`` that is not ASCII whitespace."""
+    while idx < len(text) and text[idx] in " \t\r\n":
+        idx += 1
+    return idx
+
+
+def _xml_function_payload_end(
+    text: str, payload_start: int, end_marker: str
+) -> Optional[int]:
+    """Index just past the ``</function>`` closing a qwen3_coder payload.
+
+    The ``qwen3_coder`` parser (Qwen3.5/3.6 builds) wraps XML rather than JSON
+    in the envelope::
+
+        <tool_call><function=name><parameter=k>value</parameter></function></tool_call>
+
+    so ``_json_value_end`` cannot bound it and a literal close marker inside a
+    parameter value truncates the payload the same way (#2507).
+
+    The structural anchor is that a real envelope ends with ``</function>``
+    followed by the close marker, and that its parameter elements balance.
+    Requiring both skips copies embedded in values, including a value that
+    contains the whole ``</function>`` + close-marker sequence, while still
+    ending at the FIRST call when several are concatenated.
+
+    Returns ``None`` when the payload is not this dialect or has no such pair,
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for fixing the close-marker handling. I verified the reported JSON object and Qwen cases, but I found one regression that should be fixed before merge. With `Before <tool_call>{"name":"f"</tool_call> After`, `sanitize_tool_call_markup()` returns `Before`; current main keeps `After`, so malformed JSON can drop the remaining prose. Could you preserve the text after a real closing marker without bringing back the embedded-marker leak? I think the JSON array and Qwen parameter cases can stay as follow-ups.

richgoodson:
Thanks for catching that. Fixed in 452b7d54.

The cause was the incremental payload scan. It will not accept a close marker until the JSON value closes, which is what keeps an embedded marker from ending the envelope early. A payload that never parses never satisfies that wait, so the unterminated-envelope path at end of stream withheld the rest of the text along with the markup.

At end of stream there is no more input to wait for, so a literal close marker is the best evidence available for where the envelope ends. `finish()` now splits there and re-feeds the remainder through the filter. Two details worth calling out:

- It splits at the **last** occurrence of the marker, not the first. Earlier ones can sit inside the truncated payload.
- The recovered tail goes back through `feed()` rather than being emitted directly, so markup after the close marker cannot leak into content. That re-feed can re-enter suppression under a different marker pair, so it is a loop, bounded by the number of distinct pairs.

Your case now returns `Before  After`. The #2507 case (`</tool_call>` inside a JSON string argument) still returns `Before  After` with the call parsed, so the embedded-marker leak stays closed.

Tests added at `tests/test_tool_calling.py:1213`, four of them. The first is parametrized over all six paired envelope formats and over chunked and whole-string feeding. A payload with no close marker at all is still withheld, unchanged.

I also timed the adversarial shape, n copies of the close marker inside an unclosed payload plus two alternating marker pairs: doubling the input doubles the time, so the loop stays linear.

One behavior change to flag. A malformed envelope that does have a close marker no longer becomes a recovery candidate, so its payload is dropped rather than offered for content recovery. That matches main, but it differs from what this PR did before the fix.

The JSON array and qwen3_coder parameter cases are left as follow-ups as you suggested. Happy to pick either up.

jundot:
Thanks a lot for this one, and for the quick fix after my last comment. This was a tricky bug to get right and the span scanning approach is much better than the old regex. I checked the #2507 cases again on 452b7d54 and they all work well now. One small thing I noticed while testing the EOF fallback, rfind can still drop some text when the same close marker shows up more than once after a broken tool call, but that part is easier to handle on my side since it touches the parser internals. No need for another revision, I'll merge this and push a small follow-up on main for it.

PR #2547, opened 2026-08-07
fix(cache): bound allocator pool and allow cooperative abort in store_cache

### LINKED ISSUE #2546: store_cache boundary snapshots grow the MLX allocator pool quadratically (200GB+), ending in watchdog self-kill
### Summary

On long-context sessions (70–80k tokens), the **post-response `store_cache` path explodes resident RAM until omlx kills itself**. The sequence is fully reproducible — we hit it four times in one evening, with identical log signatures each time:

1. A turn completes normally.
2. ~40s later (during the boundary-snapshot store), memory pressure goes `ok -> hard` with a huge allocator pool: `pool=210–258GB`, on top of ~200GB of weights.
3. The enforcer's reclaim is a no-op against the pool: `freed_hot=0.0GB` on **every** attempt (20+ polls).
4. The enforcer force-evicts the model — in one incident *after* pressure had already recovered to `ok`.
5. Teardown blocks on the very `store_cache` future still in flight, and after 60s the watchdog `fatal_exit`s the whole server:
   `Engine teardown timed out after 60s ... exiting process so the supervisor can restart with a clean state`.

### Environment

- omlx 0.5.7 (brew), macOS 26 / Darwin 25.6, Mac Studio M3 Ultra 512GB
- Model: Inkling-Small mlx 6-bit — per-layer `CacheList(KVCache, ArraysCache)`, 42 layers, `Estimated memory per 64-token block: 10.50 MB`
- Client: Claude Code via the Anthropic-compatible endpoint; session context ~72–80k tokens
- Paged SSD cache enabled (block_size=2048), hot cache 0

### Log evidence (one of four identical incidents)

```
22:45:41 Using boundary cache snapshot for f7789d69: storing 71680/72206 tokens
         (skipping trailing partial block, 15 intermediate snapshots)
22:45:41 Anthropic message: model=inkling-small-6bit, 110 tokens in 101.24s (1.1 tok/s)
22:46:24 Memory pressure level: ok -> hard (current=488.9GB, soft=469.9GB, hard=482.6GB, ceiling=508.0GB)
22:46:24 Requested pressure cache reclaim on 1 scheduler(s) (freed_hot=0.0GB, pool=233.4GB)
   ... (20+ reclaim attempts, freed_hot=0.0GB every time, pool 233–258GB) ...
22:46:45 Memory pressure level: hard -> ok (current=391.8GB, ...)   <- recovered on its own
22:46:45 Unloading model: inkling-small-6bit (immediate abort)      <- evicted anyway
22:46:46 Waiting for 1 inflight async store_cache future(s)...
22:47:45 CRITICAL Engine teardown timed out after 60s ... exiting process
```

### Root cause

`store_cache` consumes one **boundary snapshot per 2048-token block**, and each snapshot carries the **full cache state at that boundary** (for `CacheList` models every layer is kept whole — `"CacheList"` is not in `_KNOWN_SLICEABLE_CACHE_TYPES`, so the sliceable `KVCache` member is retained too). Snapshot sizes strictly grow with the boundary position, so their transient buffers never share a size: once freed they land in the MLX allocator pool (`mx.get_cache_memory()`), are never reused (the next snapshot is bigger), and are never returned to the OS.

The cumulative pool cost is **quadratic in context length**: at 78k tokens ≈ 37 snapshots, `sum(i × 2048 tokens × 164KB/token) ≈ 236GB` — matching the observed `pool=210–258GB` in every incident.

Meanwhile the only reclaim path — the inference thread's `_sync_and_clear_cache` — is serialized behind `_mx_buffer_access_lock`, which the store worker holds **for the entire store**. So the enforcer polls, frees nothing, and escalates to eviction; eviction's `shutdown()` then waits on the in-flight store future and dies on `FATAL_TEARDOWN_TIMEOUT_S`.

### Eliminations (things it is *not*)

- **Not KV size**: ~13GB at 78k tokens (10.5MB per 64-token block) — irrelevant on 512GB.
- **Not TurboQuant-related**: identical behavior with TQ on (6-bit) and off. (For this model TQ is partially inert anyway: `TurboQuant KV disabled: composite CacheList layers with KVCache members have no TQ store/restore path yet`.)
- **Not prefill**: with Chunked Prefill ON, prefill peaked at a healthy ~263GB total; the explosion consistently starts *after* `Anthropic message: ...` completion.
- **Not a second model loading**: the engine pool log shows a single loaded model throughout.

### Minor related paper cut

With `iogpu.wired_limit_mb=498073` the enforcer is satisfied (its warning demands `Recommended maximum: 498073 MB`), but the admin panel demands `Raise it in Terminal: sudo sysctl iogpu.wired_limit_mb=498074`. The two surfaces disagree by 1MB and no value satisfies both.

### Fix

PR incoming from this account: bounds the pool from the store worker between blocks and adds a cooperative abort so teardown cancels the store instead of fatal-exiting. A deeper follow-up would be filtering `CacheList` snapshots per-member (keep `ArraysCache`, drop the sliceable `KVCache` member), which removes the quadratic cost at capture time.

Pietroski:
Fixes #2546.

### Problem

On long-context sessions the post-response `store_cache` explodes resident RAM until the teardown watchdog `fatal_exit`s the server. Boundary snapshots carry the full cache state at strictly growing token counts, so their transient buffers never share a size: freed ones accumulate in the MLX allocator pool, unreused and unreturned to the OS — **quadratic in context length** (78k tokens ≈ 37 snapshots ≈ 236GB, matching the observed `pool=210–258GB`). The only reclaim path (`_sync_and_clear_cache` on the inference thread) is locked out by `_mx_buffer_access_lock` for the entire store, so the memory enforcer frees 0 bytes, force-evicts the model, and eviction then blocks on the very store future in flight until `FATAL_TEARDOWN_TIMEOUT_S` kills the process. Full forensics (four identical incidents, log excerpts, eliminations) in #2546.

### Changes

1. **`prefix_cache.store_cache`: bound the pool between blocks.** After each snapshot-consuming block, drop the snapshot reference and call `mx.clear_cache()` when `mx.get_cache_memory()` exceeds `_STORE_POOL_TRIM_BYTES` (8GB). The trim runs on the store worker itself — the same thread doing the buffer-protocol reads — so it is serialized with them by construction, the same guarantee `_mx_buffer_access_lock` gives the inference-thread clear (#1106 stays respected). Effect: pool cost goes from O(n²) to O(largest single snapshot).

2. **`store_cache` cooperative abort + scheduler wiring.** New optional `should_abort` callable polled at block boundaries; blocks already persisted stay valid. `Scheduler.shutdown()` sets a `_store_cache_abort` event before waiting on in-flight futures, so pressure-eviction cancels the cache write (harmless to abandon) instead of dying on the teardown watchdog.

### Testing

- New `tests/test_store_cache_pool_trim.py` (5 tests): trim fires once per block when the pool is over-ceiling, stays dormant under it; abort stops cleanly at block boundaries (mid-store and pre-store); default path unchanged. Harness mirrors `test_prefix_cache_cachelist_mixed.py` (production-shaped `CacheList(KVCache, ArraysCache)` dicts through a real hot-cache-only `PagedSSDCacheManager`).
- Full neighboring suites green: `test_prefix_cache.py`, `test_prefix_cache_cachelist_mixed.py`, `test_boundary_snapshot_store.py`, `test_paged_ssd_cache.py`, `test_prefix_cache_rotating_tip_strip.py`, `test_prefix_cache_v4_block_storage.py` — 329 passed.

### Follow-up (not in this PR)

The deeper fix is filtering `CacheList` boundary snapshots per-member — keep the non-sliceable `ArraysCache` state, drop the sliceable `KVCache` member (currently the whole composite is retained because `"CacheList"` is not in `_KNOWN_SLICEABLE_CACHE_TYPES`). That removes the quadratic cost at capture time but touches the snapshot serialization format and the merge/restore path, so it deserves its own change. This PR makes the failure mode survivable regardless.

---

Co-authored-by: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/omlx/cache/prefix_cache.py
+++ b/omlx/cache/prefix_cache.py
@@ -36,6 +36,18 @@ from .pooling_delta import POOLING_CACHE_DELTA_CLASS
 from .stats import PrefixCacheStats
 from .type_registry import CacheTypeRegistry
 
+# Allocator-pool ceiling for store_cache (bytes). Boundary snapshots carry the
+# full non-sliceable cache state at strictly growing token counts, so their
+# transient buffers all have distinct sizes: once freed they land in the MLX
+# allocator pool and are never reused (the next snapshot is bigger) nor
+# returned to the OS. Over a long-context store (dozens of snapshots) the pool
+# grows quadratically with context length — 78k tokens ≈ 37 snapshots ≈
+# >200 GB of dead pooled buffers — while the inference thread's
+# _sync_and_clear_cache is locked out by _mx_buffer_access_lock for the whole
+# store. Trimming from the worker itself (same thread as the buffer reads, so
+# inherently serialized) caps the pool at O(largest single snapshot).
+_STORE_POOL_TRIM_BYTES = 8 << 30
+
 logger = logging.getLogger(__name__)
 
 # Cap on the supersede-on-extend lineage map (tip hash -> previous tip hash).
@@ -469,6 +481,7 @@ class BlockAwarePrefixCache(CacheManager):
         extra_key_token_start: int | None = None,
         extra_key_ranges: list[tuple[int, tuple[Any, ...]]] | None = None,
         hot_cache_write_back: bool = True,
+        should_abort: Callable[[], bool] | None = None,
         _store_exact_terminal: bool = False,
     ) -> BlockTable | None:
         """
@@ -492,6 +505,11 @@ class BlockAwarePrefixCache(CacheManager):
                 ArraysCache state instead of placeholders in hybrid models.
             hot_cache_write_back: When False, SSD-backed hot cache is bypassed
                 for newly stored dirty blocks.
+            should_abort: Optional callable polled between blocks. When it
+                returns True the store stops cleanly after the current block:
+                blocks already persisted stay valid, the rest are simply not
+                stored. Lets engine teardown cancel a long-running store (a
+                cache write is safe to abandon; the server is not).
             _store_exact_terminal: Internal exact-prefix mode that persists the
                 trailing partial block and isolates the terminal hash from
                 ordinary prefix matching.
@@ -596,6 +614,16 @@ class BlockAwarePrefixCache(CacheManager):
         tip_block_saved = False
 
         for i in range(num_new_blocks):
+            if should_abort is not None and should_abort():
+                logger.info(
+                    "store_cache aborted for %s after %d/%d block(s) "
+                    "(cooperative teardown); stored blocks remain valid",
+                    request_id,
+                    i,
+                    num_new_blocks,
+                )
+                break
+
             start_idx = i * self.block_size
             end_idx = min(start_idx + self.block_size, len(new_tokens))
             block_tokens = new_tokens[start_idx:end_idx]
@@ -846,6 +874,21 @@ class BlockAwarePrefixCache(CacheManager):
                     block_table.block_ids.pop()
                     block_table.num_tokens -= len(block_tokens)
                     break
+
+                # Drop this block's boundary snapshot (an SSD-loaded snapshot
+                # is a fresh object per __getitem__; releasing the reference
+                # frees its buffers into the allocator pool) and trim the
+                # pool when it exceeds the ceiling. Snapshot transients have
+                # strictly growing, never-reused sizes, so without this the
+                # pool grows quadratically with context length over the
+                # store — see _STORE_POOL_TRIM_BYTES. Running on the store
+                # worker itself keeps the trim serialized with this thread's
+                # buffer-protocol reads (the same guarantee
+                # _mx_buffer_access_lock gives the inference-thread clear).
+                snapshot_cache_data = None
+                block_kv_data = None
+                if HAS_MLX and mx.get_cache_memory() > _STORE_POOL_TRIM_BYTES:
+                    mx.clear_cache()
 
         # Supersede-on-extend: on rotating (sliding-window) models every store
         # of a growing conversation writes one tip block carrying the full
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -1826,6 +1826,13 @@ class Scheduler:
         # Track in-flight store futures per request_id for lookup wait /
         # shutdown wait.
         self._inflight_store_futures: dict[str, concurrent.futures.Future] = {}
+        # Cooperative store_cache cancellation. shutdown() sets this so an
+        # in-flight store stops at the next block boundary instead of pinning
+        # engine teardown behind a long cache write — the 60s teardown
+        # watchdog otherwise fatal-exits the whole server while a
+        # many-snapshot store is still grinding (observed with CacheList
+        # models at 70k+ token contexts).
+        self._store_cache_abort = threading.Event()
         self._inflight_store_info: dict[str, _InflightStoreInfo] = {}
         # Admission-only cache freshness waits. A waiting request can pause at
         # the front of the queue for a relevant in-flight store without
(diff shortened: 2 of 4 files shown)


Pietroski:
**Verified on the affected machine** (Mac Studio M3 Ultra 512GB, omlx 0.5.7 with this patch applied to site-packages, `inkling-small-6bit`).

Replayed the same Claude Code workload that produced four identical self-kills in #2546 — this time at an even larger context than any of them:

```
11:06:28  Using boundary cache snapshot for 5668739f: storing 83968/85022 tokens
          (skipping trailing partial block, 40 intermediate snapshots)
11:06:29  Anthropic message: model=inkling-small-6bit, 250 tokens in 227.76s
```

40 intermediate snapshots vs. 15/34/37 in the four crashes.

**Unpatched behavior at this exact point, all four times:** `ok -> hard` within ~40s of the snapshot line, `pool=210–258GB`, `freed_hot=0.0GB` on every reclaim attempt, forced eviction, teardown deadlock on the in-flight store future, watchdog `fatal_exit`.

**Patched behavior:** 90+ seconds of post-store observation — no pressure transition, no `pool=` escalation, no eviction, no `CRITICAL`. Wired memory peaked at 259GB during generation and settled at ~240GB through the entire post-response store (weights 200GB + KV ~14GB + bounded transients), versus 460–490GB at the equivalent moment in every unpatched incident.

Will keep the patched install serving overnight and report if anything regresses.

jundot:
Thanks for the detailed investigation and the logs from the affected machine. I agree with the failure analysis, but I do not think this PR should merge in its current form.

The worker-side pool trim calls `mx.clear_cache()` without synchronizing the engine and default streams. oMLX deliberately routes cache clears through `_sync_and_clear_cache()` because unsynchronized clears have caused Metal crashes. Holding `_mx_buffer_access_lock` does not stop inference work on the engine thread, so this opens another crash path.

I also do not want `Scheduler.shutdown()` to abort every in-flight cache store. Waiting for `store_cache` was intentional. A long-context prefix can represent several minutes of prefill work, and preserving it is usually worth more than making a normal shutdown finish sooner. `Scheduler.shutdown()` is used for normal server shutdown, model unloads, swaps, and restarts as well as pressure eviction, so the abort event is not limited to the emergency case reported in #2546.

The store became a teardown hazard because the mixed CacheList path was writing cumulative snapshots quadratically. That root cause belongs in #2550. I would prefer to drop both the worker-side trim and the unconditional abort, and not merge #2547.

If a linearized store still exceeds the teardown limit after #2550, emergency cancellation should be a separate change with an explicit forced-eviction signal and a bounded grace period. Normal shutdown should continue waiting for the cache write.

Pietroski:
Agreed on all counts — closing this PR.

You're right on both mechanisms: the worker-side `mx.clear_cache()` bypassed the stream synchronization that `_sync_and_clear_cache()` exists to provide (holding `_mx_buffer_access_lock` protects the buffer reads, not concurrent engine-stream work), and the shutdown abort treated every teardown as an emergency when `shutdown()` also serves normal unloads, swaps, and restarts where a long-context prefix is worth waiting for.

#2550 has been rebased onto `main` without these changes and reworked per your review. Supporting your assessment with data: the affected machine has been serving inkling-small-6bit on the reworked #2550 alone — no trim, no abort — and the post-response store of a 94k-token session completed with no memory-pressure events and a ~15 GB linear footprint. The quadratic was indeed the entire hazard.

If a linearized store ever still trips the teardown watchdog in practice, I'll bring the emergency-cancellation idea back as a separate change with an explicit forced-eviction signal and a bounded grace period, as you suggested.

PR #2549, opened 2026-08-07
fix(cache): never rebuild a degenerate 0-slot ArraysCache from empty state

0rand:
## Summary

Two cache-reconstruction defects break **fixed-slot `ArraysCache` models** (Ling 3.0 Flash `bailing_hybrid` — its `LinearAttention` stores q/k/v/conv states in slots 0..3) on multi-turn continuation requests:

### 1. `ArraysCacheHandler.reconstruct_cache` builds a degenerate 0-slot cache when state is empty

When a non-sliceable `ArraysCache` layer's snapshot was skipped (placeholder-only blocks, or a fresh never-updated cache), `reconstruct_cache({"states": []})` returned `ArraysCache(size=0)`. The model's forward pass then crashed on the next slot read/write:

```
ValueError: not enough values to unpack (expected 4, got 0)
  File ".../bailing_hybrid.py", line 359, in __call__
    q_state, k_state, v_state, recurrent_state = cache
```
(or `list assignment index out of range` on the write side).

**Fix:** return `None` so the caller keeps the model's own fresh cache.

### 2. Scheduler ArraysCache capture trims 4-slot caches to 2 elements

`omlx/scheduler.py` captured `state = (cache_list[0], cache_list[1])` for any `ArraysCache`-style layer, so a 4-slot cache round-tripped as a 2-slot cache — and the model's `cache[2] = ...` / `cache[3] = ...` writes raised `IndexError` on the next forward pass.

**Fix:** keep the full list (`state = tuple(cache_list)`). Downstream serialization (paged_ssd_cache / boundary_snapshot) is documented as N-tuple aware.

## Symptom (user-visible)

TEB / any multi-turn client against oMLX serving Ling-3.0-flash: single-turn requests pass, multi-turn continuations die with `peer closed connection without sending complete message body (incomplete chunked read)` — the server 500s mid-stream. Single-shot curls never reconstruct cache, hence "other tests work well."

## Tests

- `reconstruct_cache({"states": []}) is None` (regression)
- 4-slot state round-trips all slots in order
- Verified locally against the repo checkout (mlx 0.32): 6/6 checks pass

### DIFF
--- a/omlx/cache/type_handlers.py
+++ b/omlx/cache/type_handlers.py
@@ -848,6 +848,17 @@ class ArraysCacheHandler(CacheTypeHandler):
             return None
 
         states = state.get("states", [])
+        if not states:
+            # No state persisted for this layer — a non-sliceable cache
+            # whose snapshot was skipped (placeholder-only blocks) or a
+            # fresh never-updated cache. Returning a degenerate
+            # ArraysCache(size=0) hands the model an empty cache object
+            # whose first slot write or tuple-unpack crashes the forward
+            # pass (Ling 3.0 bailing_hybrid: "not enough values to
+            # unpack (expected 4, got 0)" / "list assignment index out
+            # of range"). Return None so the caller keeps the model's
+            # own fresh cache instead.
+            return None
         cache = ArraysCache(size=len(states))
         for i, s in enumerate(states):
             cache.cache[i] = s
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -6742,10 +6742,16 @@ class Scheduler:
                             }
                         )
                 elif hasattr(layer_cache, "cache"):
-                    # ArraysCache style: state stored in .cache list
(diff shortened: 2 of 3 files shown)


jundot:
Thanks for the report. I tested this on current main with the local Ling 3.0 Flash FP8 model. The follow-up request restored a 2,048-token prefix across all 42 layers and finished normally, with all four `ArraysCache` slots intact.

The scheduler change here only touches the `.cache` fallback. mlx-lm's `ArraysCache` has `.state`, so Ling never reaches that branch. The new four-slot test covers the handler, not the scheduler path.

Returning `None` for an empty state rejects the whole prefix hit and re-prefills. Main already does that through `_validate_cache()` when it sees a zero-slot cache. I need a repro showing where the empty snapshot is written or selected before I can merge this.

0rand:
@jundot Thank you, I downloaded latest dev2 and the issue seems to be fixed in the same multi-turn request that .7 required this fix. This PR henceforth is mute!

PR #2550, opened 2026-08-07
fix(cache): per-member CacheList block storage — linear instead of quadratic

### LINKED ISSUE #2546: store_cache boundary snapshots grow the MLX allocator pool quadratically (200GB+), ending in watchdog self-kill

### LINKED ISSUE #2551: Mixed CacheList block storage is quadratic on SSD (282.7 GB for one 84k-token session)
### Summary

Split out from #2546 (the RAM-explosion crash) so each defect has its own tracking: mixed `CacheList` block storage writes **quadratically-sized data to the paged SSD cache**. This is the disk-side manifestation of the same root cause, and it persists after #2547 makes the RAM side survivable.

### Measurement

A single ~84k-token session (inkling-small, `CacheList(KVCache, ArraysCache)`, block size 2048):

- Admin panel Runtime Cache Observability: **41 indexed blocks, 41 SSD files, 282.7 GB** (of a 372 GB cold-cache limit)
- A linear per-block layout would be ~**13.8 GB** (measured ~164 KB/token × 84k)

The quadratic model matches observation within 2%:

```
block i ≈ i × 2048 tokens × 164 KB ≈ i × 0.336 GB
total   = 0.336 × (1+2+…+41) = 0.336 × 861 ≈ 289 GB   (observed: 282.7 GB)
```

### Root cause

In `prefix_cache._extract_block_tensor_slice`, the `all_sub_sliceable` check fails for mixed CacheLists (ArraysCache conv state is 3D), so the **whole composite — including the sliceable KV member — falls into cumulative last-block-style storage for every block**: each block's SSD file carries the full prefix state up to its boundary rather than its own 2048-token slice. The per-boundary snapshot store pays the same cost during capture.

### Consequences

- One long session nearly fills a 372 GB cold cache; a second session evicts the first almost entirely → prefix reuse collapses across sessions.
- Heavy sustained SSD writes during long sessions (macOS "disk writes" diagnostics fire).
- The same cumulative snapshots are the transient source of the allocator-pool explosion in #2546.

### Fix

#2550 (per-member CacheList block storage): slice the sliceable KV members per block from the live cache, keep only the boundary snapshot's small non-sliceable state, tag the format (`__cache_list_pm__`), reject mixed legacy/per-member chains. Verified on the affected machine: a **larger** (94k-token) session stored **15.3 GB / 46 blocks** — ~19× reduction, within 1% of the linear model.

Pietroski:
Fixes #2551. Fixes #2546.

Rebased onto current `main` — no longer stacked on #2547 (dropped per review; the pool trim and shutdown abort are not part of this change).

### Problem

Mixed `CacheList` layers (e.g. inkling's `CacheList(KVCache, ArraysCache)`) store the **full cumulative state of every member in every 2048-token block**: the layer-level `all_sub_sliceable` check fails on the 3D ArraysCache conv state, so the whole composite — giant KV member included — takes the last-block-only cumulative path. Cost is quadratic in context length: measured **282.7 GB of SSD for one 84k-token session** (41 blocks; `sum(i×2048×164KB) ≈ 289 GB`), and the matching transient explosion in the allocator pool during the post-response store that drove the #2546 self-kills.

### Design

**Store** (`prefix_cache._extract_block_tensor_slice`): `cachelist_pm_member_plan` splits eligible mixed layers — sliceable KV members become **true per-block slices from the live cache**; ArraysCache-style members keep the boundary snapshot's small state. Blocks are tagged `__cache_list_pm__`; the tag rides the SSD sidecar (`layer_{i}_storage_mode=pm`) and is re-applied at load. Eligibility is conservative: `KVCache`-family + `ArraysCache`/`SizedArraysCache` members only; PoolingCache and rotating members keep the legacy path untouched.

**Restore** (`reconstruct_cache`): per-member chains concatenate KV slices along the sequence axis and take the last matched block's state for boundary members — partial-prefix hits work at any boundary. Chains mixing legacy and per-member blocks are rejected (re-prefill), never merged.

**Capture** (`scheduler._extract_snapshot_cache_states`): boundary snapshots blank the sliceable members of pm-eligible layers; `_merge_boundary_with_full_cache` refills member-wise when a snapshot is promoted to a store source.

### Review fixes (from the first round)

1. **Missing middle snapshot** — `store_cache` now **truncates at the first non-last block without a boundary snapshot** when per-member layers are present (prior blocks remain a valid chain), and `reconstruct_cache` verifies every per-member KV sub's restored length equals `block_table.num_tokens`, rejecting the cache otherwise. The reported pm/placeholder/pm repro is a regression test and now rejects instead of silently skipping tokens.
2. **Legacy blocks recreating mixed chains** — the storage layout is part of the compatibility identity: pm blocks stamp an `@pm` token into the `cachelist_subtypes` descriptor (payload side from the marker, expectation side from class-level eligibility via `cachelist_pm_class_eligible`). Legacy blocks fail the signature match after upgrade and are swept by the existing stale-signature machinery. Upgrade test included; observed live on the test machine: `Invalidated 149 SSD index entries with stale layer cache signature (kept 0)` on first model load, with no mixed chain recreatable afterwards.
3. **In-memory fallback** — `_decode_boundary_snapshot_value` now pre-extracts eagerly through the per-member filter instead of retaining raw CacheList objects, closing the quadratic-RAM gap when the boundary snapshot SSD store is unavailable. Covered by a test.

### Testing

- `tests/test_prefix_cache_cachelist_per_member.py` (11 tests): plan classification, per-block-sized payloads through a real save/load round-trip, multi-block + partial-prefix positional round-trips, legacy compat, mixed-chain rejection, store truncation on a missing middle snapshot, the reviewer's short-KV repro, upgrade sweep, decode-fallback filtering.
- Full local sweep: **855 passed, 0 failed**.

### Verified on the affected machine (M3 Ultra 512GB, inkling-small-6bit)

- **Size**: 94k-token session stored **15.3 GB / 46 blocks** vs 282.7 GB / 41 blocks legacy for a *smaller* session (~19×; linear model within 1%). No memory-pressure events through the post-response store — with this build running **without** the #2547 mitigations.
- **Correctness (restored vs standard prefill)**: cleared cache, temperature-0 request with a 10,009-token prompt → full fresh prefill (`re-prefills 10009 of 10009, reused 0`, 20.5s) stored the chain; identical second request restored it (7.8s). **Byte-identical outputs including reasoning content.**
- **Upgrade path**: first load after deploying this build swept all 149 pre-upgrade blocks via the layout signature, as designed.

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -370,7 +370,7 @@ def _block_cachelist_subtypes(
         if not (
             isinstance(layer_data, tuple)
             and len(layer_data) == 2
-            and layer_data[0] == "__cache_list__"
+            and layer_data[0] in ("__cache_list__", "__cache_list_pm__")
             and isinstance(layer_data[1], (list, tuple))
         ):
             continue
@@ -2333,12 +2333,18 @@ class PagedSSDCacheManager(CacheManager):
                     isinstance(layer_data, tuple)
                     and len(layer_data) == 2
                     and isinstance(layer_data[0], str)
-                    and layer_data[0] == "__cache_list__"
+                    and layer_data[0] in ("__cache_list__", "__cache_list_pm__")
                 ):
                     # CacheList: sub-indexed tensors. Each sub_tensor may be
                     # a 2-tuple (legacy) or an ``__nstate__`` marker.
+                    # ``__cache_list_pm__`` marks the per-member storage mode
+                    # (sliceable subs hold per-block slices, non-sliceable
+                    # subs hold boundary state); the mode rides the sidecar
+                    # so load_block can re-tag the payload for reconstruct.
                     sub_tensors = layer_data[1]
                     cache_list_meta[f"layer_{i}_sub_count"] = str(len(sub_tensors))
+                    if layer_data[0] == "__cache_list_pm__":
+                        cache_list_meta[f"layer_{i}_storage_mode"] = "pm"
                     for j, sub_tensor in enumerate(sub_tensors):
                         sub_prefix = f"layer_{i}_sub_{j}"
                         if (
@@ -2693,8 +2699,15 @@ class PagedSSDCacheManager(CacheManager):
                     # Preserve the legacy list shape — callers (prefix_cache,
                     # tests) expect ``cache_data[i]`` to be a list of
                     # sub-cache states for CacheList layers, not a wrapper
-                    # marker.
-                    cache_data.append(sub_tensors)
+                    # marker. Per-member blocks re-tag so reconstruct_cache
+                    # can pick the per-sub restore mode.
+                    if (
+                        file_metadata
+                        and file_metadata.get(f"layer_{i}_storage_mode") == "pm"
+                    ):
+                        cache_data.append(("__cache_list_pm__", sub_tensors))
+                    else:
+                        cache_data.append(sub_tensors)
                 else:
                     layer_marker = _load_nstate(f"layer_{i}", fallback_class=cache_type)
                     if layer_marker is None:
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -324,6 +324,35 @@ _CACHELIST_NON_SLICEABLE_SUB_CLASSES = frozenset(
 
 _ARRAYS_SUB_CLASSES = frozenset({"ArraysCache", "SizedArraysCache"})
 _POOLING_SUB_CLASSES = frozenset({"PoolingCache", "BatchPoolingCache"})
+_PM_SLICEABLE_SUB_CLASSES = frozenset(
+    {"KVCache", "BatchKVCache", "QuantizedKVCache"}
+)
+
+# Storage-layout token appended to a mixed CacheList layer's subtype
+# descriptor when the layer uses per-member block storage. Part of the
+# compatibility identity: legacy cumulative blocks lack the token, so the
+# stale-signature sweep invalidates them instead of letting token-hash
+# dedup keep splicing them into per-member chains (#2550 review).
+_PM_LAYOUT_TOKEN = "@pm"
+
+
+def cachelist_pm_class_eligible(sub_class_names: list[str]) -> bool:
+    """Class-level eligibility for per-member CacheList block storage.
+
+    True when every member is either a sliceable KV class or an
+    ArraysCache-style class, with at least one of each. Must stay in sync
+    with ``prefix_cache.cachelist_pm_member_plan`` (which additionally
+    checks live tensor shapes at store time).
+    """
+    if not sub_class_names:
+        return False
+    names = [str(n) for n in sub_class_names]
+    has_slice = any(n in _PM_SLICEABLE_SUB_CLASSES for n in names)
+    has_boundary = any(n in _ARRAYS_SUB_CLASSES for n in names)
+    all_known = all(
+        n in _PM_SLICEABLE_SUB_CLASSES or n in _ARRAYS_SUB_CLASSES for n in names
+    )
+    return has_slice and has_boundary and all_known
 
 
 def _canonical_sub_name(name: Any) -> str:
@@ -411,6 +440,8 @@ def _block_cachelist_subtypes(
             else:
                 descriptors.append(name or "?")
         if descriptors and has_non_sliceable:
+            if layer_data[0] == "__cache_list_pm__":
+                descriptors.append(_PM_LAYOUT_TOKEN)
             subtypes[str(i)] = descriptors
     return subtypes or None
 
@@ -449,6 +480,11 @@ def cachelist_subtypes_from_cache_list(
             else:
                 descriptors.append(name or "?")
         if descriptors and has_non_sliceable:
+            sub_names = [
+                _canonical_sub_name(type(sub).__name__) for sub in sub_caches
+            ]
+            if cachelist_pm_class_eligible(sub_names):
+                descriptors.append(_PM_LAYOUT_TOKEN)
             subtypes[str(i)] = descriptors
     return subtypes or None
 
--- a/omlx/cache/prefix_cache.py
+++ b/omlx/cache/prefix_cache.py
@@ -641,6 +641,41 @@ class BlockAwarePrefixCache(CacheManager):
             )
 
         blocks_saved_to_ssd = 0
+        # Per-member CacheList layers need a boundary snapshot for every
+        # non-last block. A missing middle snapshot must TRUNCATE the store
+        # at that boundary: falling through to a placeholder produces a
+        # per-member/placeholder/per-member chain whose reconstruction
+        # silently skips the placeholder block's KV tokens (restored KV
+        # shorter than block_table.num_tokens -> scheduler skips tokens).
+        pm_layers_present = False
+        if is_tensor_data:
+            for layer_state in cache_data:
+                if not isinstance(layer_state, dict):
+                    continue
+                type_name = str(
+                    layer_state.get("class_name")
+                    or layer_state.get("cache_type")
+                    or ""
+                )
+                if type_name != "CacheList":
+                    continue
+                names = layer_state.get("sub_class_names") or []
+                if not names:
+                    meta = layer_state.get("meta_state")
+                    if (
+                        isinstance(meta, (list, tuple))
+                        and len(meta) >= 1
+                        and isinstance(meta[0], (list, tuple))
+                    ):
+                        names = [str(n) for n in meta[0]]
+                if (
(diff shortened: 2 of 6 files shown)


jundot:
Thanks for following up with the root fix. The per-member storage design is the right direction, and the focused 411-test sweep passed for the normal contiguous-snapshot path. I found a few cases that need to be addressed before merge.

Please first rebase this branch onto the current `main` without the #2547 changes. I do not want to carry the worker-side pool trim or the unconditional shutdown abort. A long-context prefix can represent several minutes of prefill work, so normal shutdown should continue waiting for an in-flight store rather than discarding that work for a faster teardown.

A missing middle snapshot can silently skip KV tokens. With three 4-token blocks and the 8-token snapshot omitted, the stored modes were per-member, placeholder, and per-member. `reconstruct_cache()` returned an 8-token KV cache while `block_table.num_tokens` remained 12. The scheduler would then skip four uncached tokens. Please stop per-member storage at the first missing boundary. Reconstruction should also verify that every block has the expected payload and that the restored KV length matches `block_table.num_tokens`.

Existing legacy blocks can create a mixed chain that never becomes usable after an upgrade. The per-member marker is stored in the sidecar but is not part of the compatibility signature, so token-hash dedup keeps reusing legacy blocks and appending per-member blocks. The mixed chain is rejected, but a full prefill and re-store creates the same chain again. Please include the storage layout in the compatibility identity or explicitly invalidate legacy blocks for eligible CacheLists. A restart or upgrade test should confirm that the second store does not recreate the mixed chain.

The in-memory fallback needs the same filtering. If boundary snapshot storage is unavailable or fails, `_decode_boundary_snapshot_value()` retains the raw CacheList with its full KV member. That leaves the RAM-side quadratic behavior in the fallback path. Please filter the KV member there and add coverage for this case.

On the affected model, I would also like the final verification to compare restored output against a standard-prefill run, not only the SSD size and memory footprint. With these fixes and tests, I do not expect further design changes.

Pietroski:
Thanks for the thorough review — all four points addressed; the branch is force-pushed, rebased onto current `main` with the #2547 changes dropped, and the PR body updated. Point by point:

1. **Rebase without #2547**: done. Two commits on top of `main`; no pool trim, no shutdown abort anywhere in the diff.
2. **Missing middle snapshot**: `store_cache` now truncates at the first non-last block lacking a boundary snapshot when per-member layers are present, and `reconstruct_cache` verifies each per-member KV sub's restored length against `block_table.num_tokens`. Your pm/placeholder/pm repro is now a regression test (`test_restore_rejects_short_kv_chain`) — it rejects the cache instead of skipping tokens.
3. **Legacy mixed-chain recreation**: the storage layout is now part of the compatibility identity — pm blocks stamp an `@pm` token into `cachelist_subtypes` (payload side from the marker, expectation side from class eligibility), so the existing stale-signature sweep invalidates legacy blocks on upgrade. `test_legacy_blocks_swept_on_pm_expectation` covers the restart/upgrade path; deploying this build on the affected machine swept all 149 pre-upgrade blocks on first load and no mixed chain was recreatable.
4. **In-memory fallback**: `_decode_boundary_snapshot_value` now pre-extracts eagerly through the per-member filter rather than retaining raw CacheList objects (`test_decode_snapshot_fallback_filters_kv`).

On the verification you asked for: cleared cache, temperature-0, 10,009-token prompt on inkling-small-6bit — full fresh prefill (`re-prefills 10009 of 10009, reused 0`) versus a restored second run: **byte-identical outputs including reasoning content**. Details in the PR body. The machine has been running this build **without** the #2547 mitigations, and the post-response store shows no pressure events — consistent with your assessment that the quadratic was the entire hazard.

Full sweep: 855 passed. Ready for another look.

jundot:
Thanks for the thorough rework. All four review points held up under my own re-verification, including a disk-backed restart probe for the upgrade sweep, and the on-disk layout is linear exactly as you measured. I'll merge this.

One gap I found while re-checking: the parser-stop store path (_prepare_prompt_boundary_cache_store) promotes a member-filtered snapshot without refilling the blanked KV member, so it stores single-sub blocks that dedup keeps reusing and that prefix stays a miss until restart. Restore rejects them cleanly so nothing corrupts, and it only fires when a request finishes without prompt_cache. I'll take care of it in a follow-up commit on main together with a couple of small cleanups, no action needed on your side.

PR #2552, opened 2026-08-07
feat(mac-app): bind all-networks listen mode to dual-stack IPv6 wildcard

tannerdsilva:
The macOS app's 'All networks' Listen Address option bound only to the IPv4 wildcard 0.0.0.0, so IPv6 clients could never connect in that mode.

Under the hood, I decide to assign the bind to the :: IPv6 wildcard instead of any IPv4 value. On macOS, a single :: socket is dual-stack and accepts both native IPv4 and native IPv6 traffic; binding 0.0.0.0 and :: as two sockets fails with EADDRINUSE (the :: socket already claims the IPv4 wildcard via IPv4-mapped addressing).

No server-side change is required: uvicorn's bind_socket() already emits an AF_INET6 socket for a ':'-containing host, and is_valid_bind_host() already accepts '::'. AppConfig.connectableHost() already normalizes '::' to the loopback client, so the app's own connection is unaffected.


jundot:
Thanks for adding this. I verified that the dual-stack approach works correctly on macOS. I found one small upgrade edge for the existing `0.0.0.0` Picker state, so I'll add a narrow compatibility fix to this branch, rerun the macOS tests, and then merge.

PR #2555, opened 2026-08-08
inkling: tolerant tool-call payload parsing (fixes #2453)

### LINKED ISSUE #2453: Inkling: nested-object tool-call arguments dropped (arguments="{}")
**Summary**: For `inkling_mm_model`, tool calls whose arguments contain a **nested object** arrive with `arguments: "{}"` while flat-argument calls parse fine.

**Repro** (temp=0, Inkling-Small): tools=[book_hotel with `guests: {adults, children}` object param], prompt "Book a hotel in Chicago from 2026-09-01 to 2026-09-05 for 2 adults and 1 child." → response `tool_calls[0].function.arguments == "{}"`, while the model's own reasoning_content explicitly lists every argument correctly (city, checkin, checkout, guests). A flat-args call in the same session (calculate_tip) parses completely.

Looks like the `<|content_invoke_tool_json|>` finalize parse mishandles nested braces. Suspect area: the tool-JSON extraction regex/scan in `adapter/output_parser.py`.

studioburnside:
Implements the two hardenings from #2453, both born from instrumented captures of real quantized-checkpoint traffic (community REAP25 3-bit-expert Inkling-Small build, temp=0):

**1 — accept the OpenAI wire convention.** Quantized checkpoints drift between Inkling-native `{"name", "args": {...}}` and OpenAI-style `{"name", "arguments": "<json-string>"}` (raw capture in the issue thread). The parser now accepts `args` or `arguments`, JSON-decodes a string value, and falls back to `{}` only for non-objects.

**2 — brace-balance repair for truncated payloads.** Under different kernel numerics the same checkpoint emitted a complete nested `args` object short exactly one closing `}` before `<|end_message|>` (raw capture in thread). Repair runs ONLY after strict `json.loads` fails — well-formed payloads are never touched.

**Measured**: 20-case tools eval (M5 Max 128GB, temp 0): 17/20 → 18/20; the remaining failures are model judgment, not parsing. Serving in our production-adjacent fork since 2026-08-01 across 0.5.4→0.5.7 with no regressions on Qwen3.5/3.6 or Laguna tool traffic (inkling-parser-scoped).

Happy to adjust style or add tests to match the suite's conventions.

### DIFF
--- a/omlx/adapter/output_parser.py
+++ b/omlx/adapter/output_parser.py
@@ -782,8 +782,21 @@ class InklingOutputParserSession:
             try:
                 parsed = json.loads(payload)
             except (json.JSONDecodeError, ValueError):
-                logger.debug("Inkling tool-call payload not valid JSON")
-                continue
+                # Quantized checkpoints occasionally emit the payload with the
+                # final closing brace(s) missing (observed: a complete nested
+                # args object short exactly one "}" before <|end_message|>).
+                # Brace-balance repair only runs after strict parsing failed,
+                # so well-formed payloads are never touched.
+                balance = payload.count("{") - payload.count("}")
+                if balance > 0:
+                    try:
+                        parsed = json.loads(payload + "}" * balance)
+                    except (json.JSONDecodeError, ValueError):
+                        logger.debug("Inkling tool-call payload not valid JSON")
+                        continue
+                else:
+                    logger.debug("Inkling tool-call payload not valid JSON")
+                    continue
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for the detailed captures and for keeping this scoped to the Inkling parser. The strict-first parsing preserves existing behavior, and both reported payload variants are handled correctly. One thing I will fold into a follow-up is making brace balancing string-aware and adding parser regression coverage for the reported variants and braces inside argument strings. LGTM. I will merge this now.

Issue #2558, opened 2026-08-08
DeepSeek V4 prefill optimization roadmap: M3 Ultra ablation and proposed PR split


DiscoStew6082:
## Context

PR #2553 demonstrated a meaningful DeepSeek V4 long-prompt prefill improvement, but it combined several independent optimizations, extensive tests, and raw evidence into one review surface. That PR is now closed.

Following [the maintainer's request](https://github.com/jundot/omlx/pull/2553#issuecomment-5225073347), I reconstructed the work against current upstream `d2575b1df5a4012966839f75fdb200e7e0b20743` and measured each optimization independently. This issue records the full result and the proposed sequence of small PRs. It is not a proposal to merge the combined patch.

## Controlled full-set result

Test environment:

- Mac Studio with M3 Ultra and 512 GB unified memory
- `DeepSeek-V4-Flash-0731`, model revision `7872f01b1d1fe23eabc4c98b48bffcef5a386062`
- fixed 17,219-token API prompt, SHA-256 `a1465f4b5ee68dbd173c138dd65718bd06957439b66e99069e91f50364bf81f1`
- cache disabled and `cached_tokens=0` verified on every response
- one request at a time, balanced memory guard, 2,048-token prefill steps
- MTP enabled with three draft tokens
- thinking explicitly disabled; expected response exactly `READY`
- one unrecorded warm-up before each five-run median

Bracketing control runs:

| Control | Five prefill measurements (seconds) | Median | Prompt throughput |
| --- | --- | ---: | ---: |
| Upstream start | 40.33, 40.49, 40.46, 40.47, 40.32 | 40.46 s | 425.61 tok/s |
| Full set start | 34.91, 34.93, 34.79, 34.88, 34.85 | 34.88 s | 493.69 tok/s |
| Full set end | 34.68, 34.59, 34.67, 34.62, 34.62 | 34.62 s | 497.36 tok/s |
| Upstream end | 40.32, 40.44, 40.33, 40.41, 40.36 | 40.36 s | 426.60 tok/s |

The complete optimization set produced:

- **15.99% to 16.59% higher prompt throughput**
- **13.79% to 14.22% lower prefill time**

## Marginal contribution breakdown

These are leave-one-out measurements. They include feature interactions and are scoped to this model, prompt, and M3 Ultra.

| Rank | Optimization | Marginal throughput contribution |
| ---: | --- | ---: |
| 1 | Route ratio-128 compressed attention through the native local+pooled path | **10.18%** |
[...]

PR #2559, opened 2026-08-08
perf(deepseek-v4): route ratio-128 prefill through native attention

DiscoStew6082:
## Summary

- route DeepSeek V4 ratio-128 compressed prefill through the existing native local-plus-pooled attention path with sequential pooled indices
- opt in only for causal/sliding masks created by the model; direct callers with custom masks retain the dense SDPA reference path
- retain dense reference attention for sub-4-bit checkpoints and preserve the existing DSpark single-token and empty-pool paths

## Performance evidence

The controlled leave-one-out ablation in #2558 measured a **10.18% marginal prompt-throughput contribution** from this path.

This is scoped to:

- Mac Studio with M3 Ultra and 512 GB unified memory
- `DeepSeek-V4-Flash-0731`, revision `7872f01b1d1fe23eabc4c98b48bffcef5a386062`
- upstream source `d2575b1df5a4012966839f75fdb200e7e0b20743`
- fixed 17,219-token prompt, SHA-256 `a1465f4b5ee68dbd173c138dd65718bd06957439b66e99069e91f50364bf81f1`
- 2,048-token prefill steps, MTP enabled with three draft tokens, cache disabled and `cached_tokens=0`

The normalized measurements and full methodology are in #2558. The linked [evidence package](https://github.com/DiscoStew6082/omlx/tree/06b515b0fd3543341b346e74cf0d3a13da2d6c06/benchmarks/evidence/deepseek_v4_flash_prefill_m3_ultra) preserves the raw investigation artifacts without adding them to this PR.

This is not an unconditional cross-chip performance claim.

## Fallbacks

- custom or otherwise unsupported masks use the original concatenate-plus-SDPA reference path
- quantization metadata below 4 bits disables this native ratio-128 route whether it appears in `quantization`, `quantization_config`, or `text_config.quantization_config`
- unsupported native shapes/dtypes, a missing native symbol, or an immediate kernel exception return directly to the original concatenate-plus-SDPA reference path without constructing the gathered sparse fallback
- DSpark single-token exact attention and empty-pool local attention are unchanged

## Validation

- `uv run pytest -q tests/test_deepseek_v4_patch.py tests/test_deepseek_v4_dspark.py tests/test_mlx_lm_mtp_patch.py tests/test_custom_kernel_abi_probe.py` — 313 passed, 1 order-dependent skip after the preceding modules initialized the base patch; the MTP file passes 119/119 alone
- native FP16 and BF16 outputs matched the causal reference within 0.004 and 0.032 maximum absolute error respectively, including 257 pooled rows across the kernel's 256-row tile boundary
- full local FP8 `DeepSeek-V4-Flash-0731` load plus a 257-token prefill and 17-token cached continuation reached the native kernel in every ratio-128 layer at both zero and nonzero offsets and produced finite logits
- full local oQ2.5 `DeepSeek-V4-Flash-0731-oQ2.5e` load plus the same cache-backed sequence produced finite logits with zero ratio-128 native-helper or kernel entries
- patched `mlx_lm.utils.load_model` tests verify the high-bit/sub-4-bit policy reaches `ModelArgs`
- all four in-tree custom-kernel ABI probes passed
- Black check passed for all changed files
- `git diff --check` passed

Refs #2558.

### DIFF
--- a/omlx/patches/deepseek_v4/deepseek_v4_model.py
+++ b/omlx/patches/deepseek_v4/deepseek_v4_model.py
@@ -165,6 +165,7 @@ class ModelArgs(BaseModelArgs):
     n_mtp_layers: int = 0
     tie_word_embeddings: bool = False
     topk_method: str = "noaux_tc"
+    use_native_ratio128_attention: bool = True
 
     def __post_init__(self):
         if not self.compress_ratios:
@@ -1550,6 +1551,8 @@ class CompressedAttention(nn.Module):
         x: mx.array,
         mask: Optional[mx.array] = None,
         cache: Optional[Any] = None,
+        *,
+        _standard_mask: bool = False,
     ) -> mx.array:
         B, L, _ = x.shape
         local_cache = cache[0] if cache is not None else None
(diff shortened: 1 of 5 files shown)


jundot:

|---:|---:|---:|---:|

PR #2560, opened 2026-08-08
fix(server): surface unenforced response_format on /v1/responses

fxd0h:
when xgrammar isnt available, /v1/chat/completions tells the client the
response_format wasnt enforced via the Warning header (199 omlx "...").
/v1/responses compiles the grammar through the same helper but drops the
header on the floor, so a client on that route has no way to know the
schema wasnt enforced short of reading server.log.

we hit this in production testing: an adversarial prompt walked right
through a json_schema request and the only trace was a log line. the
warning header at least makes it detectable from the client side, same as
chat completions already does.

this wires the existing helper into the /v1/responses path, stream and non
stream. no behavior change when xgrammar is installed.

tests follow the existing convention (helper level, same as the chat
completions wiring). suite green, no new failures.

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -5741,6 +5741,7 @@ async def create_response(
         # Handle text.format (structured output)
         response_format = None
         compiled_grammar = None
+        response_format_warning = None
         if request.text and request.text.format:
             fmt = request.text.format
             if fmt.type == "json_object":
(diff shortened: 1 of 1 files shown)


fxd0h:
bump now that 0.6.0 is out — this is still the fix for the silent `response_format` degrade the `Warning: 199` header admits to. branch rebases clean, no conflicts.

jundot:
Thanks for carrying the existing response-format downgrade signal into both `/v1/responses` return paths. I verified the warning is present only on grammar fallback, absent when enforcement succeeds or no format is requested, and the existing SSE headers remain intact. LGTM.

PR #2561, opened 2026-08-08
perf(grammar): accept sampled tokens at the top of the next step

fxd0h:
the grammar patch in scheduler.py syncs right after dispatch (mx.eval on
the sampled tokens) to feed accept_token. that eval blocks the host in the
middle of the step and kills mlx-lm's one step lookahead, so every request
with a grammar pays a sync bubble on each token.

this defers the accept to the top of the next _step. by then the sampled
tokens are already materialized as a side effect of the normal pipeline, so
the accept costs nothing and the forward dispatch never waits on host work.
the matcher sees exactly the same tokens in the same order, output is bit
identical, verified with fixed seeds against main.

measured on an m5 pro 24gb, Qwen3.5-9B-6bit (248k vocab), json_object,
temp 0, 700 token decodes, interleaved a/b (4 cycles, server restart per
arm, 12 reps each) on an idle machine:

    main       38.77 +- 0.21 tok/s with grammar
    this pr    38.92 +- 0.10 tok/s (+0.4%)

small, but its free, bit identical, and it removes a host sync from the
middle of the step, so the dispatch never blocks on python work. for what
its worth the remaining grammar cost doesnt look like the mask
itself (fill is ~2us cpu and the apply kernel ~12-25us gpu per token, about
0.1% of a 25ms step), it looks like the logits processor path forcing a
materialization per step. thats a separate and deeper change, this pr just
takes the free part.

tests: suite green, same pre existing failures as main and nothing new.
added coverage for the deferred accept (pending token survives batch
extension and row realignment, rejection still resamples on the same step).

### DIFF
--- a/omlx/api/grammar.py
+++ b/omlx/api/grammar.py
@@ -78,6 +78,7 @@ class GrammarConstraintProcessor:
         self._bitmask = np.full((1, bitmask_width), -1, dtype=np.int32)
         self._terminated = False
         self._first_call = True
+        self._pending = False
 
     # ------------------------------------------------------------------
     # Per-request mode (original interface)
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for this. Deferring grammar token acceptance to the next step preserves the per-row matcher state while removing the immediate sync from the decode pipeline. I verified the focused grammar and MTP tests against current main, along with single and concurrent `json_object` requests on a real server. LGTM, merging.

PR #2562, opened 2026-08-08
perf(deepseek-v4): raise MXFP4 large-block threshold to 16K

DiscoStew6082:
## Summary

- keep MXFP4 BM16/variant 1 blocks through 16,383 routes and switch to BM32/variant 2 at 16,384 routes
- preserve the existing 8,192-route crossover for affine 2/3-bit and mixed native block plans
- allow unmeasured pre-NAX systems to restore the previous MXFP4 crossover with `OMLX_DEEPSEEK_MXFP4_LARGE_BLOCK_MIN_ROUTES=8192`
- add focused boundary coverage for the default, override, and unchanged affine behavior

No kernel math changes in this PR.

## Performance evidence

The controlled M3 Ultra ablation in #2558 measured median prompt throughput increasing from **479.08 to 487.91 tok/s**, a **1.84% improvement**. Median prefill time decreased from 35.94 to 35.29 seconds.

This measurement is scoped to:

- Mac Studio with M3 Ultra and 512 GB unified memory
- `DeepSeek-V4-Flash-0731`, revision `7872f01b1d1fe23eabc4c98b48bffcef5a386062`
- upstream source `d2575b1df5a4012966839f75fdb200e7e0b20743`
- fixed 17,219-token prompt, SHA-256 `a1465f4b5ee68dbd173c138dd65718bd06957439b66e99069e91f50364bf81f1`
- 2,048-token prefill steps, MTP enabled with three draft tokens, cache disabled, and `cached_tokens=0`
- one warmup followed by five measured cold runs, all producing the expected `READY` response

The linked [evidence package](https://github.com/DiscoStew6082/omlx/tree/06b515b0fd3543341b346e74cf0d3a13da2d6c06/benchmarks/evidence/deepseek_v4_flash_prefill_m3_ultra) preserves the raw investigation artifacts without adding them to this PR.

This is not an unconditional cross-chip performance claim.

## Cross-chip behavior

- pre-NAX systems can set `OMLX_DEEPSEEK_MXFP4_LARGE_BLOCK_MIN_ROUTES=8192` to restore the previous MXFP4 crossover
- M5/NAX prefill continues to use the existing stock `mx.gather_qmm` fallback by default; its existing controls are unchanged
- affine and mixed native block plans retain their existing 8,192-route crossover

## Validation

- `uv run pytest tests/test_glm_moe_dsa_patch.py tests/test_deepseek_v4_patch.py -q -rs` — 114 passed, 4 skipped because the optional native extension was unavailable in the clean PR worktree
- focused default and override threshold coverage — 6 passed
- direct production override probe passed
- `uv run ruff check --select E,F,W ...` passed
- Python compilation and `git diff --check` passed
- independent read-only code review completed with no findings

Refs #2558.

### DIFF
--- a/omlx/patches/deepseek_v4/switch_layers.py
+++ b/omlx/patches/deepseek_v4/switch_layers.py
@@ -20,7 +20,12 @@ _DEEPSEEK_MXFP4_SMALL_BLOCK_BM = 16
 _DEEPSEEK_MXFP4_SMALL_BLOCK_VARIANT = 1
 _DEEPSEEK_MXFP4_LARGE_BLOCK_BM = 32
 _DEEPSEEK_MXFP4_LARGE_BLOCK_VARIANT = 2
-_DEEPSEEK_MXFP4_LARGE_BLOCK_MIN_ROUTES = 8192
+_DEEPSEEK_AFFINE_LARGE_BLOCK_MIN_ROUTES = 8192
+# Tuned on M3 Ultra. Set this to 8192 to restore the previous crossover on
+# other pre-NAX chips; M5 prefill uses the NAX fallback below.
+_DEEPSEEK_MXFP4_LARGE_BLOCK_MIN_ROUTES = int(
+    os.environ.get("OMLX_DEEPSEEK_MXFP4_LARGE_BLOCK_MIN_ROUTES", "16384")
+)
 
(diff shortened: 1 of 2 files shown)


jundot:

|---:|---:|---:|---:|

PR #2563, opened 2026-08-08
perf(deepseek-v4): skip redundant all-pooled indexer scoring

DiscoStew6082:
## Summary

- Skip indexer query projection, scoring, and top-k selection when every pooled row already fits within `index_topk`.
- Still run the indexer compressor so its `PoolingCache` advances exactly as before.
- Guard against divergence between the attention and indexer compressor row counts.
- Retain the existing full indexer path for empty pools and pools larger than `index_topk`.
- Leave downstream dense/native attention dispatch unchanged.

## Correctness

The regression test uses the existing full indexer as the reference for a non-aligned ratio-4 prefill: 10 input tokens produce two pooled rows and a two-token remainder.

It verifies that:

- the full indexer scoring path is not invoked;
- the indexer compressor is invoked exactly once;
- the attention and indexer caches advance to the expected pooled offset;
- the remainder and every public element of `PoolingCache.state` exactly match the reference path;
- the reference selection contains the expected sequential pooled indices.

## Performance evidence

Issue #2558 attributes approximately 0.4% marginal throughput improvement to this slice in its controlled leave-one-out testing.

That result is scoped to:

- Mac Studio with M3 Ultra and 512 GB unified memory;
- `DeepSeek-V4-Flash-0731` at revision `7872f01b1d1fe23eabc4c98b48bffcef5a386062`;
- fixed 17,219-token prompt, SHA-256 `a1465f4b5ee68dbd173c138dd65718bd06957439b66e99069e91f50364bf81f1`;
- cold cache with one request at a time;
- balanced memory guard and 2,048-token prefill steps;
- MTP enabled with three draft tokens.

This is an M3 Ultra result, not an unconditional cross-chip performance claim.

Evidence package:
https://github.com/DiscoStew6082/omlx/tree/06b515b0fd3543341b346e74cf0d3a13da2d6c06/benchmarks/evidence/deepseek_v4_flash_prefill_m3_ultra

## Validation

- `tests/test_deepseek_v4_patch.py`: 92 passed
- `tests/test_deepseek_v4_dspark.py`: 64 passed, 11 skipped
- Black formatting check passed
- Python compilation check passed
- `git diff --check` passed
- Independent code review found no issues

Refs #2558

### DIFF
--- a/omlx/patches/deepseek_v4/deepseek_v4_model.py
+++ b/omlx/patches/deepseek_v4/deepseek_v4_model.py
@@ -1806,7 +1806,18 @@ class SparseCompressedAttention(nn.Module):
 
         pooled = self.compressor(x, comp_cache, offset)
         pmask = comp_cache.make_mask(L, offset) if comp_cache is not None else None
-        topk = self.indexer(x, q_residual, self.rope, idx_cache, offset)
+        if 0 < pooled.shape[1] <= self.indexer.index_topk:
+            index_pooled = self.indexer.compressor(x, idx_cache, offset)
+            if index_pooled.shape[1] != pooled.shape[1]:
+                raise RuntimeError(
+                    "DeepSeek V4 attention/indexer pooling caches diverged"
+                )
+            topk = mx.broadcast_to(
+                mx.arange(pooled.shape[1], dtype=mx.uint32)[None, None],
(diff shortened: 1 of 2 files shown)


jundot:

|---:|---:|---:|---:|

PR #2423, opened 2026-07-30
Distributed cluster serving across Macs (tensor + pipeline parallel)

ashhart:
## Distributed cluster serving across Macs

Serve one model across two or more Macs from the admin dashboard: pick the
cluster, pick the model, choose tensor or pipeline parallelism, activate.
Planning, staging, memory safety and rollback are handled by the server; the
dashboard exposes one `/v1` endpoint when the cluster is ready.

### Verified on hardware

- **Tensor parallel** — Qwen3.6-27B across 2 Macs over Thunderbolt RDMA
  (jaccl): **28.6 tok/s vs 16.1 single-node (1.78x), byte-identical output.**
- **Pipeline** — MiniMax-M3-4bit (225 GB, more than either Mac can serve with
  usable context alone) split across a 128 GB MacBook Pro (workstation role,
  18-layer stage pinned exactly as planned) and a 256 GB Mac Studio: one-click
  activation from the dashboard reached ready on both ranks and passed the
  one-token canary over jaccl. An earlier run served 212 tokens before losing
  its remote rank to an idle ssh control-channel drop; keepalives were added
  since, and a long sustained-serving run is the remaining verification.

### What's in here

- **Capability-aware planning**: a model that mlx-lm cannot split (no
  `pipeline()`/`shard()`) is refused at plan time, before staging — not after
  a hundred gigabytes have moved. The catalogue answers "which of my models
  fit this cluster, and at how much context".
- **Node roles**: a `workstation` Mac (someone works on it) admits less than a
  `headless` one, enforced by the memory guard at load, not just planning.
- **Memory safety**: per-rank admission against the GPU's real ceiling, a
  watchdog sampling during the load, and a rank-0 prefill guard per request.
- **MiniMax-M3 support for mlx-lm ranks**: registered over the vendored
  mlx-vlm implementation with an assigned-stage `pipeline()` so a rank
  materialises only the layers the plan gave it.
- **Selective staging**: a rank stages only the shards its stage needs, with
  disk checks and size verification.
- **Liveness**: idle heartbeats, peer watchdogs, and clean teardown — a
  vanished Mac becomes a stated failure instead of a hung collective.

### Honest caveats (things a reviewer would find anyway)

1. `memory_guard_tier` is not yet carried per-rank (`role` is); ranks default
   to `balanced`.
2. The pre-load even-split guard is deliberately conservative: it refuses an
   uneven plan whose *even* split would not fit, even for architectures that
   honour the assigned-stage pin. Fails closed; costs some availability on
[...]

PR #2591, opened 2026-08-10
feat(cluster): add heterogeneous Metal and CUDA model pool

ashhart:
## Relationship to the Mac clustering PR

This is a side addition to [#2423 — Distributed cluster serving across Macs](https://github.com/jundot/omlx/pull/2423). It builds on that PR and should land after it. Once #2423 merges, this branch can be rebased so this PR shows only the heterogeneous CUDA extension.

## What this adds

- combines Apple Metal/MLX and NVIDIA CUDA workers into one logical model-memory pool
- automatically plans model shards from usable memory, with GUI controls for manual shard weighting
- shows NVIDIA CUDA workers, logical ConnectX groups, installed VRAM, and model-usable pooled memory in the dashboard
- adds GUI-generated, short-lived, single-use join commands for CUDA and DGX Spark workers
- installs and verifies a worker-minimal environment without requiring a desktop oMLX installation
- discovers ConnectX-7 links, verifies CUDA pairs with an NCCL smoke test, and keeps verified peers adjacent in placement
- adds compatibility, transport, memory, pipeline, cache, and live-runtime diagnostics

## Current topology boundary

The current implementation runs an outer multi-rank Ring across the fleet. A verified ConnectX-7 CUDA pair is grouped and placed together, but this does not yet implement a true hierarchical Ring → NCCL supernode gateway.

## Validation

- Full suite: 8,855 passed, 64 skipped, 74 deselected
- Focused enrollment, routing, launch, dashboard, and worker tests: 190 passed
- Ruff checks and `git diff --check` pass
- Personal machine aliases, username fixtures, and exact worker addresses were replaced with generic test data before publication


sdiamanEXUS:
This rocks! Now, an extension could be to have the prompt processing taking place on the cuda devices (where it is faster), and token generation taking place on the mac.

ashhart:
@sdiamanEXUS thank you!

Yea, that can be added as disaggregated prefill if that's something people would want. 

For now I kept it simple so oMLX can use CUDA boxes for additional vram.

jundot:
Thanks for another big one, and for validating it on real hardware again. Now that #2423 landed with its follow-ups, this should rebase cleanly onto main and the diff here collapses to just the CUDA extension you described.

I went over the new surface the same way I did #2423: the remote-worker attack surface, the non-cluster regression path, and code quality. The short version is this sits in the same shape as #2423, and I'm comfortable landing it as an experimental, toggle-gated addition.

On security the enrollment flow holds up well. The join credentials are single-use 256-bit bearer tokens kept only as digests, the bootstrap is sha256-pinned rather than curl-into-bash, and the controller identity plus the source bundle are both fingerprint-checked before anything runs as root. I did not find a remotely reachable RCE, injection, or auth bypass. Putting the join router behind the toggle without the admin cookie is the right call since a worker cannot carry that cookie, and the one unauthenticated route only serves non-secret pinned source.

On regressions a Mac-only or non-cluster user is untouched. The engine-pool changes stay inside the cluster branch, the CUDA memory path only activates when mx.cuda is available, and the SSD cache change is gated off by default, so the zero-regression property from #2423 carries over.

The one real defect I would want fixed, though not as a merge blocker:

- The CUDA admission ceiling re-inflates when free VRAM reads zero. In `_cuda_ceiling_breakdown` a genuinely measured `free_memory == 0` (for example vLLM already owns the whole GB10) returns `hard_limit == 0`, and `probe.py:545` reads that zero as "not measured" and substitutes 90 percent of physical memory. The dashboard then advertises that as capacity and the planner can place a shard that OOMs on load. The sibling `probe_remote_admission_ceiling` fails closed on the same zero, so only the capability-probe path is fail-open. The fix is to tell a measured zero apart from an unmeasured (exception) zero.

A few smaller things for later, none blocking: `run_cuda_fabric_probe` copies most of the launcher scaffolding from `run_cluster_performance_probe` and could share it, `WORKER_REQUIREMENTS` is a hand-kept dependency list that will drift from pyproject over time, and `remove_node` in enrollment has no caller yet since nothing deletes an enrolled node. The adjacency-only Ring compatibility mode is fine and clearly documented as an interim step.

I'll follow the same path as #2423: rebase and merge to dev, then carry the ceiling fix and the polish as follow-ups against the merge commit. Really solid work, this is a lot of careful engineering.

PR #2593, opened 2026-08-10
fix(tool-calling): stop deep nesting escaping the parse chain (#2545)

### LINKED ISSUE #2545: RecursionError escapes parse_tool_calls on deeply nested tool-call payloads
### Describe the bug

Deeply nested model output makes `parse_tool_calls` raise `RecursionError`, which escapes the parse chain instead of degrading to a parse failure.

`_parse_xml_tool_calls` calls `json.loads(content, strict=False)` inside a `try` that catches `json.JSONDecodeError` and friends. `json.loads` recurses once per nesting level, so deeply nested input raises `RecursionError`, which is not a `ValueError` and so is not caught.

Tool-call text is model output, and model output is steerable by prompt content, so this is reachable from a request rather than only from a malfunctioning model.

### To Reproduce

Verified on `198c5ce9` (v0.5.7) in a clean worktree:

```python
import json
from unittest.mock import MagicMock

tok = MagicMock(spec=[])
tok.has_tool_calling = True
tok.tool_call_start = "<tool_call>"
tok.tool_call_end = "</tool_call>"
tok.tool_parser = lambda text, tools: json.loads(text)

parse_tool_calls("<tool_call>" + "[" * 100000 + "</tool_call>", tok)
# RecursionError
```

### Expected behavior

A payload that cannot be parsed should flow into the existing drop-with-warning path, the same way a malformed payload does today. The module already states this rule for the Gemma 4 argument parser, where breaching a bound is a clean parse failure rather than an exception that escapes:

https://github.com/jundot/omlx/blob/198c5ce9/omlx/api/tool_calling.py#L604-L609

> Model output is untrusted and attacker-influenceable (prompt injection can steer emissions verbatim), so parsing must stay linear-time and bounded: breaching a bound is a clean parse failure that flows into the existing drop-with-warning path, never an exception that escapes the parse chain.

### Suggested fix

Catch `RecursionError` alongside the decode errors at the `json.loads` sites in the tool-call parse chain, and optionally bound payload length the way `_GEMMA4_MAX_ARGS_LEN` already does.

### Additional context

Found while bounding a new boundary scan in #2544. That PR catches `RecursionError` in the helper it adds, and I verified this existing crash behaves identically before and after it, so #2544 neither causes nor fixes this. Filing separately to keep the two apart.

Happy to send a patch if you want it fixed this way.

### Desktop

- macOS 15, M4 Pro Mini 64GB
- oMLX `198c5ce9` (v0.5.7)

richgoodson:
Fixes #2545.

Deeply nested model output makes the decoders raise past the excepts written for decode errors, so a request can crash the parse chain instead of degrading to a parse failure.

## Which error escapes depends on the Python version

@DiscoStew6082 ran the matrix on the issue with the exact deeply nested input:

| Python | `json.loads` | `ast.literal_eval` |
|---|---|---|
| 3.11.14 | `RecursionError` | `SyntaxError` |
| 3.12.9 | `RecursionError` | `SyntaxError` |
| 3.13.12 | `RecursionError` | `SyntaxError` |
| 3.14.6 | `JSONDecodeError` | `SyntaxError` |

`ast.parse(..., mode="eval")` raises `SyntaxError` on all four. So across the supported range the JSON paths need `RecursionError` and the ast paths need `SyntaxError`. Neither is a `ValueError`, so both slip past the existing excepts.

`_DEEP_NEST_ERRORS` names the pair once and is added to every decode-site except in the module, which covers whichever one the running interpreter produces. If a future version moves the goalposts again it is one line.

## The site that matters most is not a decode call

The load-bearing case is the native parser call, not this module's own `json.loads` calls. `glm47`, `kimi_k2` and `qwen3_coder` all decode inside the parser, so the error surfaces at `tool_parser(...)` where nothing in this module is on the stack. I found the other sites by walking the AST from every `json.loads`, `ast.literal_eval`, `ast.parse` and `raw_decode` call to its enclosing handler, and that method misses this one by construction. A test caught it.

A payload the XML fallback can still read now recovers there rather than raising.

## Two Gemma 4 paths need more than a wider except

`_gemma4_args_to_json_robust` catches parse failures and retries with `_gemma4_args_to_json_legacy`. Adding the new errors to that except would send a payload that just broke a decoder into the legacy parser, and the legacy parser ignores the length and depth bounds on purpose. That hands the exact input the bounds exist to stop to the one parser that does not apply them.

Instead the deep-nest errors are converted to `_Gemma4ArgsTooComplexError`, which takes the existing reject-hard path alongside the other bound breaches. The legacy parser is guarded at its call site for the same reason, since it is the one place a deep payload can still reach a raw decoder. There is a test asserting the legacy parser never sees this input.

## Scope

Per @jundot's comment on the issue: every `json.loads` site plus the `ast.literal_eval` and `ast.parse` ones, and no extra length bound. 21 handlers widened. `extract_json_from_text`, the `response_format` path noted on the issue, is included here rather than left as a follow-up.

## Tests

The tests inject the error at the decoder rather than feeding genuinely deep input. The interpreter decides which error deep input produces, and the repro in the issue does not raise at all on 3.14, so testing with real nesting proves nothing portable. Injecting proves the excepts cover both errors on every version.

- Native parser, both errors: recoverable payload recovers via the fallback, unrecoverable payload drops to no tool calls, neither raises.
- Every `json.loads` site in the chain, both errors, via a patched decoder.
- Gemma 4: deep nesting rejects hard and the legacy parser is never called.
- Gemma 4 legacy: raising is converted rather than escaping.
- A broken call must not lose the valid call beside it. @DiscoStew6082 raised this on the issue; the parse loop runs per match, so a payload that breaks the decoder has to fail that match alone. The neighbor survives and the broken envelope's markup does not leak into content.
- The issue's original repro is kept as a smoke test rather than as the proof.

Seven of the nine injected cases fail on main. The two that pass there are explainable: main already lists `SyntaxError` at the native parser site, and the original repro no longer raises on 3.14.

Full suite green.

## Note on verification

My local venv is Python 3.14, which is outside this project's `requires-python = ">=3.11,<3.14"`, so the 3.11 through 3.13 behavior in the table above is @DiscoStew6082's measurement rather than mine. They offered to test the native parser, XML fallback, parameter-coercion and `response_format` paths across those versions once this is up, including that malformed input still follows the warning-and-drop behavior.

### DIFF
--- a/omlx/api/openai_models.py
+++ b/omlx/api/openai_models.py
@@ -129,6 +129,16 @@ class Message(BaseModel):
 # =============================================================================
 
 
+# Deeply nested values break json.loads/json.dumps in a version-dependent way
+# (RecursionError on 3.11-3.13, JSONDecodeError or SyntaxError on 3.14), and
+# neither RecursionError nor SyntaxError is a ValueError, so both escape
+# excepts written for decode errors. This validator re-parses a value the
+# tool-call parser already decoded, from a deeper stack frame, so a value that
+# was fine there can breach the limit here (#2545). Kept in this module rather
+# than in tool_calling to avoid an import cycle; tool_calling imports it.
+_DEEP_NEST_ERRORS = (RecursionError, SyntaxError)
+
+
 def _coerce_tool_call_arguments(v: Any) -> str:
     """Normalize a tool_call.arguments value to a JSON-object string.
 
@@ -141,7 +151,12 @@ def _coerce_tool_call_arguments(v: Any) -> str:
     that can't round-trip into a JSON object raises ValueError.
     """
     if isinstance(v, dict):
-        return json.dumps(v, ensure_ascii=False)
+        try:
+            return json.dumps(v, ensure_ascii=False)
+        except _DEEP_NEST_ERRORS as e:
+            raise ValueError(
+                f"arguments are nested too deeply to serialize: {e}."
+            ) from e
     if not isinstance(v, str):
         raise ValueError(
             f"arguments must be a JSON-encoded string, got {type(v).__name__}. "
@@ -153,7 +168,7 @@ def _coerce_tool_call_arguments(v: Any) -> str:
         return "{}"
     try:
         parsed = json.loads(stripped)
-    except (json.JSONDecodeError, ValueError) as e:
+    except (json.JSONDecodeError, ValueError, *_DEEP_NEST_ERRORS) as e:
         snippet = stripped if len(stripped) <= 120 else stripped[:117] + "..."
         raise ValueError(
             f"arguments must be valid JSON, got parse error: {e}. "
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -104,12 +104,17 @@ def _serialize_tool_call_arguments(arguments: Any) -> str:
     # runs after a value has already decoded successfully, from a deeper stack
     # frame. A value nested just under the limit at decode time can therefore
     # breach it here (#2545, found by DiscoStew6082 at depth ~987 on 3.11).
-    # Falling through to the "{}" path below keeps that a clean coercion.
+    #
+    # A breach is deliberately NOT coerced to "{}" like a non-object value is.
+    # The two are different failures: a non-object is a parser quirk we can
+    # safely normalize, whereas failing to serialize means we HAVE the
+    # arguments and cannot render them. Returning "{}" there would hand back a
+    # runnable tool call with its arguments silently removed, so `write_file`
+    # would still fire with nothing to write. Let it raise instead, so
+    # `_build_tool_call` drops that one call with a warning, which is what the
+    # issue asks for (jundot's review on #2593).
     if isinstance(arguments, dict):
-        try:
-            return json.dumps(arguments, ensure_ascii=False)
-        except _DEEP_NEST_ERRORS:
-            arguments = "<nested too deeply to serialize>"
+        return json.dumps(arguments, ensure_ascii=False)
     # mlx-vlm / mlx-lm gemma4 parser returns a JSON-object string per the
     # OpenAI spec. Accept it when it parses back to a dict.
     if isinstance(arguments, str):
@@ -118,10 +123,7 @@ def _serialize_tool_call_arguments(arguments: Any) -> str:
         except (json.JSONDecodeError, ValueError, *_DEEP_NEST_ERRORS):
             parsed = None
         if isinstance(parsed, dict):
-            try:
-                return json.dumps(parsed, ensure_ascii=False)
-            except _DEEP_NEST_ERRORS:
-                pass
+            return json.dumps(parsed, ensure_ascii=False)
     logger.warning(
         "Tool parser returned non-dict arguments (type=%s, repr=%.200r); "
         "coercing to empty object to keep downstream template safe.",
@@ -745,18 +747,41 @@ def _parse_hermes_tool_calls(text: str) -> Tuple[str, Optional[List[ToolCall]]]:
             if isinstance(call.func, ast.Name):
                 func_name = call.func.id
             elif isinstance(call.func, ast.Attribute):
-                func_name = ast.unparse(call.func)
+                try:
+                    func_name = ast.unparse(call.func)
+                except _DEEP_NEST_ERRORS:
+                    continue
             else:
                 continue
 
             arguments = {}
+            unrepresentable = False
             for kw in call.keywords:
                 if kw.arg is None:
                     continue
                 try:
                     arguments[kw.arg] = ast.literal_eval(kw.value)
                 except (ValueError, SyntaxError, *_DEEP_NEST_ERRORS):
-                    arguments[kw.arg] = ast.unparse(kw.value)
+                    # Fall back to the source text. `ast.unparse` walks the
+                    # tree recursively, so an expression `ast.parse` built
+                    # successfully can still breach the limit being rendered
+                    # back out, from a deeper frame (#2545). An argument we
+                    # cannot represent drops the whole call rather than
+                    # yielding one that is missing it.
+                    try:
+                        arguments[kw.arg] = ast.unparse(kw.value)
+                    except _DEEP_NEST_ERRORS:
+                        unrepresentable = True
+                        break
+
+            if unrepresentable:
+                logger.warning(
+                    "Dropping tool call %.80r: argument %.40r could not be "
+                    "represented (nested too deeply)",
+                    func_name,
+                    kw.arg,
+                )
+                continue
 
             _built = _build_tool_call(func_name, arguments)
             if _built is not None:
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -131,6 +131,38 @@ def _serialize_tool_call_arguments(arguments: Any) -> str:
     return "{}"
 
 
+def _build_tool_call(name: str, arguments: Any) -> Optional[ToolCall]:
+    """Build a ToolCall, dropping this one call if validation cannot finish.
+
+    ``FunctionCall`` re-parses the arguments string while validating it, which
+    is a *third* decode of a value the parser already decoded and serialized,
+    and it runs from a deeper stack frame than either (#2545).  Recursion
+    limits are about remaining stack rather than input depth, so a value that
+    was fine at both earlier steps can still breach the limit here.
+
+    Validation failure has to drop this one call and leave the rest of the
+    batch alone, the same way an unparseable match does, rather than escape
+    the parse chain.
+    """
+    try:
+        return ToolCall(
+            id=f"call_{uuid.uuid4().hex[:8]}",
+            type="function",
+            function=FunctionCall(
+                name=name,
+                arguments=_serialize_tool_call_arguments(arguments),
+            ),
+        )
+    except (ValueError, *_DEEP_NEST_ERRORS) as exc:
+        logger.warning(
+            "Dropping tool call %.80r: arguments failed validation (%s: %s)",
+            name,
+            type(exc).__name__,
+            exc,
+        )
+        return None
+
+
 @dataclass(frozen=True)
 class ToolCallExtraction:
     """Parsed tool-call result plus sanitized reasoning text."""
@@ -562,16 +594,9 @@ def _parse_xml_tool_calls(
             parsed = json.loads(content, strict=False)
             name = parsed.get("name", "")
             arguments = parsed.get("arguments", {})
-            tool_calls.append(
-                ToolCall(
-                    id=f"call_{uuid.uuid4().hex[:8]}",
-                    type="function",
-                    function=FunctionCall(
-                        name=name,
-                        arguments=_serialize_tool_call_arguments(arguments),
-                    ),
-                )
-            )
+            _built = _build_tool_call(name, arguments)
+            if _built is not None:
+                tool_calls.append(_built)
             continue
         except (json.JSONDecodeError, AttributeError, *_DEEP_NEST_ERRORS):
             pass
@@ -589,16 +614,9 @@ def _parse_xml_tool_calls(
             arguments = {}
             for key, val in _iter_xml_parameters(params_text):
                 arguments[key] = _coerce_param_value(val, key, props, func_name)
-            tool_calls.append(
-                ToolCall(
-                    id=f"call_{uuid.uuid4().hex[:8]}",
-                    type="function",
-                    function=FunctionCall(
-                        name=func_name,
-                        arguments=_serialize_tool_call_arguments(arguments),
-                    ),
-                )
-            )
+            _built = _build_tool_call(func_name, arguments)
+            if _built is not None:
+                tool_calls.append(_built)
             continue
 
         # GLM XML format: func_name<arg_key>k</arg_key><arg_value>v</arg_value>...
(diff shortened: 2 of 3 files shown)


DiscoStew6082:
Thanks — I ran the requested checks. The native-parser, JSON-first XML, response-format, Gemma 4, and neighboring-call cases behave as described.

I found one remaining `RecursionError` escape in the parameter-coercion path, though. On PR head `5ca6f82e` with Python 3.11.14, a balanced array at depths 987–991:

```python
nested = "[" * 987 + "0" + "]" * 987
text = (
    "<tool_call><function=f><parameter=x>"
    + nested
    + "</parameter></function></tool_call>"
)
_parse_xml_tool_calls(text)
```

passes the first `json.loads` in `_coerce_param_value`, but constructing `FunctionCall` invokes `_coerce_tool_call_arguments`, whose second `json.loads` runs from a deeper stack frame and raises `RecursionError` outside the new guards.

I reproduced the same escape in the Qwen XML, GLM XML, and namespaced XML parameter branches. The generic JSON and native-parser branches degrade cleanly.

The new `test_json_loads_raising_does_not_escape` does not currently cover this: its third probe uses plain text with `tool_parser=None`, so no decoder runs.

Could you guard serialization/`FunctionCall` construction per match and add a direct regression for this second-decode path? Since parameter coercion was explicitly part of the requested scope, I think this should be fixed before merge.

Thanks,  
Stew

P.S. I love music—any kind of music.

richgoodson:
Good catch, and thanks for the precise repro. Fixed in d2d8f993.

You are right about the mechanism, and it is the part I had wrong. Guarding the decoders was never going to be enough, because `json.dumps` recurses per nesting level exactly the way the decoders do, and it runs after a value has already parsed, from a deeper frame. A value sitting just under the limit when it decoded can breach it there. That is why ~987 got through with the first round of guards in place.

The cause was duplication rather than a missing except. Each of the five parse paths carried its own raw `json.dumps(arguments, ensure_ascii=False)`, which is precisely what `_serialize_tool_call_arguments` already does, guard included. They all call the helper now, so there is one guarded serialization point instead of five unguarded copies, and a breach degrades to the helper's existing coerce-to-`{}` with a warning. The helper's own two `dumps` calls are guarded for the same reason.

Verified with `json.dumps` forced to raise, on the XML fallback, Qwen XML, GLM XML, namespaced XML, Hermes, bracket, and native paths, plus the public entry point. All degrade, none escape. The new tests fail on `5ca6f82e`, the head you tested.

You were also right that `test_json_loads_raising_does_not_escape` was not pulling its weight. The third probe passed plain text with `tool_parser=None`, so no decoder ran at all. It now runs a real parser.

On my side I can only inject the error, since 3.14 does not raise at these depths, so a rerun of your depth sweep on 3.11 against the new head is the check I cannot do here. `build_json_system_prompt` also serializes unguarded, but it handles tool definitions from the request rather than model output, so I left it out of this PR.

DiscoStew6082:
Thanks — I reran the depth sweep against `d2d8f993`. There is still a narrow `RecursionError` escape on Python 3.11.

Environment: CPython 3.11.14 and 3.11.15, recursion limit 1000.

With the minimal compact XML case swept from depths 940–1030:

- `5ca6f82e`: `RecursionError` at depths 989–993
- `d2d8f993`: `RecursionError` still escapes at depths 989–992; depth 993 now reaches the new coerce-to-`{}` guard

The exact boundary moves with caller stack depth, but the remaining mechanism is consistent across the compact XML, Qwen XML, GLM XML, and real namespaced XML paths.

On `d2d8f993`, the public `parse_tool_calls` path at depth 989 reaches:

```text
_parse_xml_tool_calls
  FunctionCall(...)
    _validate_arguments_json
      _coerce_tool_call_arguments
        json.loads(stripped)
          RecursionError
```

`_serialize_tool_call_arguments` successfully serializes the value at these depths, so its guard does not fire. `FunctionCall` then reparses that deep JSON string in `openai_models._coerce_tool_call_arguments`, from a deeper frame, and that second decode remains unguarded.

Validation results:

- The PR's 21 focused deep-nesting tests pass.
- Full `tests/test_tool_calling.py`: 341 passed.
- The real Python 3.11 depth sweep still fails through the public entry point.

One related fixture issue: the new `namespaced-xml` injected test uses `<function=f><parameter=x>`, but `_parse_namespaced_tool_calls` expects `<invoke name="f"><parameter name="x">`. I confirmed the current fixture parses zero calls and invokes `json.dumps` zero times; the real namespaced grammar exercises the path and reproduces the remaining escape.

Could you guard or avoid the downstream `FunctionCall` reparse and add a direct balanced-depth regression through `parse_tool_calls` on Python 3.11? I would also adjust the namespaced fixture to use its real grammar.

richgoodson:
Right again, and thanks for the stack trace. That is what made it findable. Fixed in c80f16dd.

The decode you traced is not in `tool_calling` at all, which is why two rounds of auditing that module kept missing it. `FunctionCall` re-parses the arguments string while validating it, in `openai_models._coerce_tool_call_arguments`. That is a third decode of a value the parser already decoded and the serializer already dumped, and it runs from a deeper frame than either, so `_serialize_tool_call_arguments` succeeding at a given depth says nothing about whether validation will.

Two changes. `_coerce_tool_call_arguments` converts deep-nest errors into the `ValueError` it already raises for malformed arguments, so the API path keeps returning a clear 422 for a client that echoes a bad tool call. `_build_tool_call` wraps construction on the parse side, so a validation failure drops that one call with a warning instead of escaping, matching what the per-match handling already does for every other failure.

While testing that I found plain `ValueError` escaped the parse chain by the same route, so the guard covers it rather than leaving a known hole next to a fixed one.

You were right about the fixture too. `<function=..><parameter=..>` is not the namespaced grammar, so it parsed zero calls and called `json.dumps` zero times. It now uses `<invoke name=..><parameter name=..>`, and I confirmed the corrected version parses one call and reaches the serializer.

Added the balanced-depth sweep through the public entry point. It sweeps 940 to 1040 rather than pinning a depth, since the boundary moves with caller stack usage, and it only bites on 3.11 through 3.13, which is what CI runs. There is also an injected version covering all five dialects, and 13 of its 15 cases fail on `d2d8f993`.

Three rounds, three decodes, each one a frame deeper than the last. Worth stating plainly in case it saves someone the fourth: the invariant is not "every `json.loads` is guarded", it is "no decode of model-derived data may run unguarded at any stack depth", and serialization counts as a decode.

A sweep against the new head on 3.11 would be the confirmation I cannot produce here.

DiscoStew6082:
Confirmed against `c80f16d` on CPython 3.11.14.

I reran the balanced-depth sweep from 940 through 1040 via the public `parse_tool_calls` entry point for compact XML, Qwen XML, and the corrected namespaced XML grammar. No `RecursionError` or `SyntaxError` escaped at any depth.

The focused regression selection passes 20/20, and the full `tests/test_tool_calling.py` file passes 359/359. I also independently checked the corrected namespaced fixture: it parses exactly one call and reaches serialization.

The split handling looks right to me: request-side validation converts the deep-nest failure into the existing `ValueError`/422 path, while parse-side construction drops the failed call without letting validation escape.

CI is now green on 3.11, 3.12, and 3.13 as well. This closes my reproduction; no further changes requested from me. Thanks for following it through all three layers.

jundot:
Thanks for working through the decoder, serializer, and validator paths. I found one remaining escape on `c80f16d` through public `parse_tool_calls`: a Hermes argument with about 400 chained additions gets through `ast.parse`, then `ast.literal_eval` raises `ValueError` and the unguarded `ast.unparse(kw.value)` raises `RecursionError`. Could you guard both `ast.unparse` sites and add a public-entry regression for this case?

I also noticed that when `json.dumps` hits the recursion limit, `_serialize_tool_call_arguments` returns `"{}"`, leaving a runnable tool call with its original arguments removed. Was that intentional? The issue's expected behavior says an unparseable match should be dropped with a warning, so I think this failure should reach `_build_tool_call` and drop only that match.

DiscoStew6082:
Thanks — I reproduced both findings on `c80f16d` and put a follow-up commit on my fork: [`ded2a838`](https://github.com/DiscoStew6082/omlx/commit/ded2a83879e34aa1c80d95f75e1e3e5d8a31a0f1).

The patch guards both Hermes `ast.unparse` sites. A recursion or syntax failure now warns and drops only the affected call; a valid neighboring call in the same Hermes block survives.

I also changed `_serialize_tool_call_arguments` so deep `json.loads`/`json.dumps` failures reach `_build_tool_call` and take its existing drop path instead of returning `"{}"`. Ordinary non-dict coercion remains unchanged. Public-entry tests cover dict and JSON-string parser output, `RecursionError` and `SyntaxError`, and neighboring-call survival.

Validation on Python 3.11.15:

- Six focused regression cases passed.
- Full `tests/test_tool_calling.py`: 363 passed.
- Changed-line Ruff and Black checks plus `git diff --check` passed.

The commit is based directly on `c80f16dd` and can be applied with:

```bash
git cherry-pick ded2a83879e34aa1c80d95f75e1e3e5d8a31a0f1
```

Thanks,  
Stew

richgoodson:
Both fixed in cdc20924.

**The `ast.unparse` escape.** Reproduced on your exact case. It parses fine, fails `literal_eval` because a chained addition is not a literal, then breaches the limit in the `unparse` fallback that renders it back to source. Both sites are guarded now.

An argument that cannot be represented drops the whole call rather than just that specific argument. A call missing one of its arguments is not a safer outcome than no call at all. The drop is scoped to that one call, so a sibling call in the same batch still comes through.

Worth noting for the test matrix: `ast.unparse` raises regardless of Python version, so unlike the decoder, serializer and validator layers this one reproduces on 3.14 too.

**The `"{}"` question.** No, that was not intentional. I reused the helper's existing non-object fallback without separating two different failures. A parser returning a non-object is a quirk worth normalizing. Failing to serialize means we have the arguments and cannot render them, and returning `"{}"` there hands back a runnable tool call with its arguments silently removed, so a `write_file` still fires with nothing to write. Dropping the match is right.

`_serialize_tool_call_arguments` now lets the error reach `_build_tool_call`, which drops that one call with a warning. The `"{}"` coercion stays for a genuinely non-object value.

One of my own tests asserted the `"{}"` return, so it had encoded the bug rather than catching it. It now asserts the propagation.

**Verification.** 9 new cases, 8 of which fail on `c80f16d`. The Hermes case runs through the public `parse_tool_calls` entry point at chain lengths 200, 400 and 800. Full suite green apart from five pre-existing GPU numerics failures on this machine that are unrelated to tool calling.

That makes four layers where the same value gets re-handled a frame deeper each time: decode, serialize, validate, and now render-back-to-source. If a fifth turns up I would rather change the shape of the fix than keep widening excepts, so I am open to bounding the payload before it reaches any of them if you would prefer that.

richgoodson:
One more escape, fixed in 9f813254.

@DiscoStew6082's fork commit ded2a838 covered a branch my cdc20924 missed: the string branch of `_serialize_tool_call_arguments` still caught the deep-nest errors around its `json.loads`. Parsers that follow the OpenAI spec hand arguments back as a JSON-object string rather than a dict (mlx-vlm and mlx-lm's gemma4 among them), so a value that breached the limit there fell through to the `"{}"` coercion and produced exactly the runnable-call-without-arguments outcome the rest of this PR removes.

Removing that catch lets the error reach `_build_tool_call`, which drops the call. Malformed-but-shallow JSON still coerces, since that is a parser quirk rather than lost data.

The regression test fails on cdc20924, and it makes the decode fail only for the one payload, so the downstream validator still runs and the drop is attributable to this branch.

DiscoStew6082:
Thanks — I did one more public-entry audit against `9f813254` and found two remaining paths.

First, the bracket parser catches deep-nesting decode failures and converts the unread arguments into `{"raw": ...}`. That leaves the affected call runnable rather than dropping it. In a standalone CPython 3.11.15 sweep from depths 900–1050, depths 992–1050 returned the bad call with raw arguments; the exact boundary is caller-stack dependent.

Second, Hermes can parse valid Python literals that JSON cannot represent—sets, bytes, and complex numbers. Those currently escape from serialization as `TypeError`. This predates the PR base, but it is adjacent to the same per-call drop contract.

I pushed fixes and public-entry regressions in [`9f19c743`](https://github.com/DiscoStew6082/omlx/commit/9f19c743460e9553afc6728bc92e56fdcd842517). The bracket path now drops deep-decoder failures while retaining the existing shallow malformed-JSON fallback, and `_build_tool_call` drops non-JSON-serializable parser output without losing a valid neighboring call.

Validation on Python 3.11.15:

- Four focused regression cases passed.
- Full `tests/test_tool_calling.py`: 374 passed.
- Focused Ruff checks and `git diff --check` passed.

The commit is based directly on `9f813254` and can be applied with:

```bash
git cherry-pick 9f19c743460e9553afc6728bc92e56fdcd842517
```

jundot:
Thanks, @DiscoStew6082. I checked `9f19c743` against the current PR head `9f813254` on Python 3.11.15. Both cases reproduce on the PR head and are fixed by the follow-up: bracket depths 992 through 1050 no longer return runnable raw calls, and Hermes set, bytes, and complex arguments drop only the bad call. `tests/test_tool_calling.py` passes 374/374 and `tests/test_openai_models.py` passes 72/72. @richgoodson, I am going to apply this commit to the PR branch with its original authorship preserved, rerun CI, and merge if the updated head stays clean.

jundot:
Thanks, both remaining parser paths are fixed on the updated head. I reran the bracket-depth and Hermes literal reproductions locally, and the Python 3.11 check passes; the 3.12 and 3.13 jobs are still running. The bad call now drops without affecting a valid neighboring call. LGTM, merging.

Issue #2350, opened 2026-07-23
Feature request: Qwen3-Embedding (text) support with last-token/EOS pooling on /v1/embeddings


youngjin39:
## Summary

Please consider supporting the Qwen3-Embedding text model family (e.g. `Qwen/Qwen3-Embedding-0.6B`) on the `/v1/embeddings` endpoint, with the pooling strategy the official implementation requires.

## Motivation

- The README currently lists BERT, BGE-M3, and ModernBERT embedding families. Qwen3-Embedding-0.6B is one of the strongest Mac-sized multilingual embedding models (1024d, 32K context, 100+ languages, instruction-aware, Matryoshka) and would be a natural upgrade path for local BGE-M3 users.
- Issue #1817 shows `/v1/embeddings` applies mean pooling to Qwen3-VL-Embedding, while the official implementation uses EOS/last-token pooling — mean pooling reverses cross-modal discriminability there. The text Qwen3-Embedding family has the same requirement: official usage extracts the last-token (EOS) hidden state after an instruction-formatted prompt. A mean-pooled serving path would silently corrupt vectors even if the weights load.

## Requested behavior

1. Load Qwen3-Embedding text checkpoints (MLX conversion) as `embedding` models.
2. Use last-token/EOS pooling for this family on `/v1/embeddings` (per-family pooling selection, or model-config-driven pooling).
3. Surface the effective pooling strategy in the model metadata/response so clients can verify it (this is what downstream fingerprinting needs).

## Environment

Apple Silicon, oMLX (current release), local `/v1/embeddings` consumers.

Happy to test a build on Apple Silicon with Korean/multilingual retrieval sets.

Issue #1817, opened 2026-06-11
Qwen3-VL-Embedding: mean pooling destroys cross-modal discriminability (should use EOS/last-token pooling)


Footree:
## Problem

oMLX v0.4.3 `/v1/embeddings` uses **mean pooling** for Qwen3-VL-Embedding models. The official Qwen3-VL-Embedding implementation uses **EOS (last-token) pooling**. Mean pooling causes image tokens to overwhelm text tokens, resulting in near-zero cross-modal discriminability.

## Evidence

Tested on Qwen3-VL-Embedding-8B (both bf16 and oQ8) with 12 images × 31 text queries:

| Pooling Method | Cross-modal Gap (relevant - irrelevant) |
|---|---|
| **EOS pooling** (official, via MLX-VLM) | ~0.05 |
| **Mean pooling** (oMLX API) | **-0.04** (reversed!) |

With mean pooling, oMLX embedding API produces **negative discriminability** — irrelevant pairs score HIGHER than relevant ones.

## Expected Behavior

Follow the official Qwen3-VL-Embedding implementation:
1. Process input through vision tower + language model
2. Extract hidden state at the **EOS token position** (last valid token in attention mask)
3. Apply **L2 normalization**
4. Return as embedding vector

Reference: [Qwen3-VL-Embedding paper](https://arxiv.org/abs/2601.04720), Section 3: "We extract the hidden state vector corresponding to the [EOS] token from the base model's last layer"

## Reproduction

```bash
# Send cross-modal embedding request via oMLX API
curl -s http://localhost:8000/v1/embeddings \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $API_KEY" \
  -d '{
    "model": "Qwen3-VL-Embedding-8B-bf16",
    "input": [
      {"type": "text", "text": "the Bund Shanghai"},
      {"type": "image_url", "image_url": {"url": "file:///path/to/bund.jpg"}}
    ]
  }'
```

## Impact

Blocks production use of Qwen3-VL-Embedding models on oMLX for multimodal retrieval (image search, cross-modal RAG, etc.).

## Environment

- oMLX v0.4.3
- macOS 26.5.1 (Mac Studio M4 Max)
- Qwen3-VL-Embedding-8B (bf16 + oQ8)

## Suggested Fix

Add EOS/last-token pooling as the default for VL-Embedding models, or make pooling strategy configurable (mean vs EOS vs cls).

joergbirkhold:
[...]

PR #2594, opened 2026-08-11
fix(embeddings): honour pooling mode declared by sentence-transformers checkpoints

cedricamram:
## Problem

`/v1/embeddings` selects the first available field in the model output —
`text_embeds`, then `pooler_output`, then a mean over `last_hidden_state`. For MLX
conversions of sentence-transformers models, this can pick a **different pooling than
the checkpoint was trained with**, and nothing signals it: no error, no warning, just
lower retrieval quality.

## Concrete case (BGE-M3)

`mlx-community/bge-m3-mlx-fp16` ships a `modules.json` declaring a `1_Pooling` module,
and BGE-M3 dense retrieval is trained on the **CLS token**. The conversion exposes a
mean-pooled `text_embeds`, which oMLX picks first:

| source | cosine vs reference CLS embedding |
|---|---|
| `last_hidden_state[:, 0, :]` (CLS, what BGE-M3 expects) | 1.000000 |
| `text_embeds` (what `/v1/embeddings` returns today) | **0.776** |

Measured cost on our own 967-document corpus, same model, same index, only the pooling
changed:

| pooling | recall@1 | recall@3 |
|---|---|---|
| CLS | **65 %** | **100 %** |
| mean (`text_embeds`) | 55 % | 82 % |

We kept our embeddings out of oMLX because of this — the vectors are incompatible with
an index built on CLS, and the failure is invisible: search keeps returning documents,
just less relevant ones.

## Approach

Rather than hardcoding per-family rules (which does not scale — see #1817 for
Qwen3-VL-Embedding and #2350 requesting per-family selection), **read what the
checkpoint declares**. sentence-transformers exports store the strategy in
`1_Pooling/config.json`, and `modules.json` already advertises that module. The mode is
resolved **once in `load()`**, so the hot path — and `mx.compile` tracing in
`_compiled_embed` — is untouched.

Supports `cls`, `lasttoken` and `mean`. This also gives #2350 a config-driven path for
Qwen3-Embedding checkpoints that ship a pooling declaration.

## Compatibility

- **No `1_Pooling/config.json` → behaviour is byte-for-byte unchanged.** Every model
  that works today keeps working exactly the same way.
- A malformed config is ignored (logged at debug level), never raised.
- No new dependency; `json` and `Path` were already imported.

## Tests

Five cases added to `tests/test_embedding.py::TestDeclaredPoolingMode`, all passing —
CLS / last-token / mean honoured, no declaration leaves current behaviour intact, and
malformed config falls back safely.

---

Happy to adjust the approach if you would rather key this off model metadata, or expose
the effective pooling in the response as #2350 suggests — that part is not included here.

### DIFF
--- a/omlx/models/embedding.py
+++ b/omlx/models/embedding.py
@@ -14,13 +14,14 @@ import logging
 import os
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import mlx.core as mx
 from mlx.utils import tree_flatten
 
 from ..utils.compile_cache import clear_thread_compile_cache
 from ..utils.image import validate_image_data_uri
+from .base_model import last_token_pool, mean_pooling, normalize_embeddings
 from .mlx_embeddings_compat import (
     patch_qwen3_vl_processor_for_torch_free_image_loading,
 )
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for working on this. I compared the original checkpoints with several representative MLX models:

- [mlx-community/bge-m3-mlx-fp16](https://huggingface.co/mlx-community/bge-m3-mlx-fp16/tree/main)
- [mlx-community/all-MiniLM-L6-v2-4bit](https://huggingface.co/mlx-community/all-MiniLM-L6-v2-4bit/tree/main)
- [mlx-community/Qwen3-Embedding-0.6B-4bit-DWQ](https://huggingface.co/mlx-community/Qwen3-Embedding-0.6B-4bit-DWQ/tree/main)
- [mlx-community/Qwen3-Embedding-0.6B-8bit](https://huggingface.co/mlx-community/Qwen3-Embedding-0.6B-8bit/tree/main)
- [mlx-community/Qwen3-Embedding-0.6B-mxfp8](https://huggingface.co/mlx-community/Qwen3-Embedding-0.6B-mxfp8/tree/main)
- [jedisct1/Qwen3-VL-Embedding-8B-mlx](https://huggingface.co/jedisct1/Qwen3-VL-Embedding-8B-mlx/tree/main)
- [mlx-community/Qwen3-VL-Embedding-2B-8bit](https://huggingface.co/mlx-community/Qwen3-VL-Embedding-2B-8bit/tree/main)

The original BGE-M3 declares CLS pooling, all-MiniLM declares mean pooling, and Qwen3-Embedding declares last-token pooling. However, none of the MLX models above includes `1_Pooling/config.json`. Some retain a `modules.json` entry pointing to the missing `1_Pooling` directory, while others only contain the main model config. This means the proposed detection does not activate for the concrete BGE-M3 case or the Qwen3 MLX variants and silently keeps the current pooling behavior.

Qwen3-VL also uses the newer `{"pooling_mode": "lasttoken"}` format, which the current boolean-key parser does not recognize even when the pooling config is present. I think the resolver should locate the Pooling module through `modules.json`, support both the legacy boolean fields and the current `pooling_mode` field, and define a clear fallback for MLX conversions that dropped the sentence-transformers metadata. For known families, that fallback would be CLS for BGE-M3 and last-token for Qwen3-Embedding and Qwen3-VL-Embedding. The selected mode and its source should also be logged at load time.

The pooling operation also needs the request `attention_mask`. Direct `[:, -1]` and unmasked `mean(axis=1)` select or include pad tokens in mixed-length batches, and returning raw `last_hidden_state` bypasses the existing L2 normalization. Please route this through the existing mask-aware pooling helpers, preserve normalization, and add mixed-length batch tests that compare batched and single-input results before merging.

cedricamram:
Thank you — checking the actual MLX checkpoints was the decisive part of this review. You are right that the first revision never fired for the models it was meant to serve: without `1_Pooling/config.json`, detection returned `None` and the existing behaviour was silently kept. Pushed a revision that addresses each point.

**Resolution**
- The Pooling module is now located through `modules.json`, reading its `path` rather than assuming `1_Pooling`, with the conventional location kept as a fallback. Repos that keep the `modules.json` entry while dropping the directory fall through to the next step, since the config path is also checked for existence.
- Both dialects are accepted: the current `{"pooling_mode": "lasttoken"}` string (including `last_token`) and the legacy per-mode booleans.
- Known-family fallbacks for MLX conversions that dropped the metadata: CLS for BGE-M3, last-token for Qwen3-Embedding and Qwen3-VL-Embedding. The family is matched against the model path and `config.json`'s `_name_or_path`.
- The selected mode **and its source** are logged at load time, e.g. `Embedding pooling for <model>: cls (source: known family (bge-m3))` — an inferred mode should never be mistaken for a declared one.

**Pooling**
`_extract_embeddings_array` now takes the request `attention_mask` and routes through the existing `mean_pooling` / `last_token_pool` helpers, with `normalize_embeddings` applied so this path no longer bypasses L2 normalization. The two call sites that have the mask in scope pass it; the `mlx_embeddings.generate` path has none available, where `last_token_pool` falls back to the last position as before.

**Tests**
Added coverage for the `modules.json` lookup, a non-standard pooling directory (`2_Pooling`), both config dialects, both family fallbacks, malformed configs, CLS normalization, and mixed-length batches compared against single-input results — for mean and last-token. I verified the batch test fails against the previous unmasked implementation, so it genuinely guards the regression rather than merely passing.

One note on CI: the 3.13 failure is `tests/test_prefix_cache_gdn_split.py::test_split_restore_walks_back_from_structurally_invalid_sidecar`, which is unrelated to this change and passes on 3.11 and 3.12. Happy to rebase if it is already fixed on main.

jundot:
Thanks for the follow-up. The resolver and mask-aware helpers address most of the previous review, but I found one remaining runtime gap. Both eager paths still call `_extract_embeddings_array(outputs)` without the mask: the custom-processor branch already has `inputs["attention_mask"]` in scope, while the standard `mlx_embeddings.generate` branch discards it. With `OMLX_EMBEDDING_COMPILE=0` or after a compile failure, a right-padded Qwen3 or Qwen3-VL batch ignores the model's correct `text_embeds` and pools the padded last hidden state; my repro returned `[0.707, 0.707]` instead of `[0, 1]` for the short row. Please carry the prepared mask through both eager paths and add an end-to-end `_is_compiled=False` batch test. Also, the new tests currently fail on Python 3.11 because `mx` is used without importing `mlx.core`; after adding that import locally, all 101 selected embedding tests pass.

cedricamram:
Confirmed, and thank you for the precise repro — it reproduces exactly. Fixed in `5f4858f`.

**Both eager paths now carry the prepared mask.**

- Custom-processor branch: the mask was already in scope and simply was not passed on — it now is.
- `mlx_embeddings.generate` branch: `generate()` is `prepare_inputs` + `model(**inputs)` and returns only the outputs, so the mask it builds is unrecoverable from the caller. When the checkpoint declares a mask-aware mode (`mean` / `lasttoken` — CLS reads a fixed position), those same two steps now run here so the mask survives into pooling; otherwise `generate()` is used unchanged. Only input preparation is guarded by the fallback, so a forward error still surfaces as it does on every other path.

**End-to-end tests, one per eager branch**, with `_is_compiled=False` and a right-padded mixed-length batch. Counter-check against the previous commit — both fail, and the short row comes back as `(0.70710677, 0.70710677)` instead of `(0, 1)`, matching your numbers:

```
FAILED TestDeclaredPoolingMode::test_eager_batch_is_masked_standard_processor
FAILED TestDeclaredPoolingMode::test_eager_batch_is_masked_custom_processor
  where False = np.allclose(array([[0.70710677, 0.70710677], [0., 1.]]), [[0.0, 1.0], [0.0, 1.0]])
```

**The import**: `import mlx.core as mx` is now added locally in the methods that use it, matching the rest of the file. Full run on Python 3.12: `tests/test_embedding.py` 103 passed, and 193 passed together with the reranker / keepalive suites.

One open question, since you may prefer the stricter version: I kept `generate()` as the default so nothing changes for checkpoints that declare no pooling mode. The alternative is to always take the `prepare_inputs` + `model(**inputs)` route in the eager path, which would make the eager and compiled paths structurally identical and remove this whole class of "one path fixed, the other forgotten" divergence. Happy to switch if that is the direction you want.

jundot:
Thanks, the eager-path fixes now pass the mixed-length checks and the latest-main smoke suite. The remaining `mx.compile` issue is isolated to the shared last-token helper and has a small vectorized fix, so I will handle it as a follow-up on main. LGTM after the eager-path verification; merging.

PR #2595, opened 2026-08-11
feat(moe): expert offload — stream non-resident experts from the checkpoint

### LINKED ISSUE #986: Add Flash-MoE-style SSD-backed expert streaming for large MoE models
Is your feature request related to a problem? Please describe.

Yes. I would like to run very large Mixture-of-Experts models on Apple Silicon machines with limited unified memory, but current MLX-based inference workflows still generally assume that most model weights, including expert weights, are resident in memory or handled through conventional model loading patterns.

Projects such as flash-moe show that large MoE models can run on Apple Silicon by streaming expert weights from SSD instead of keeping the entire model in unified memory. Its README describes running Qwen3.5-397B-A17B on an M3 Max MacBook Pro with 48 GB RAM by streaming a 209 GB model from SSD through a custom Metal pipeline.

There are also MLX-related experiments in this direction. mlx-flash focuses on flash weight streaming for MLX and describes running models larger than RAM by streaming weights directly from SSD. anemll-flash-mlx is closer to the MoE-specific problem and emphasizes the boundary between dense execution, expert storage, expert selection, and expert consumption, using a stable bank / slot-bank approach with expert IDs as data.

However, there is still no clean, first-class MLX feature for Flash-MoE-style execution of large sparse MoE models where expert weights can be streamed, cached, or swapped efficiently without fully materializing all experts in unified memory.

Describe the solution you'd like

I would like MLX to support a Flash-MoE-style execution path for large MoE models on Apple Silicon.

Ideally, this would include:

SSD-backed or memory-mapped expert weight storage
Loading, streaming, or swapping only the selected experts needed for each token / layer
Keeping the dense path inside the normal MLX execution flow
Treating router outputs / expert IDs as data that can drive sparse expert execution
A stable per-layer expert bank or slot-bank abstraction
Efficient hit-path and miss-path handling for expert cache behavior
Predictive prefetching or lookahead support where possible
Support for MLX-compatible quantized checkpoints and mixed-precision expert formats
Reference examples for large MoE models such as Qwen-style A3B / A17B models
A Python-accessible API so that users can experiment without maintaining a fully custom C / Metal inference engine

The goal is not necessarily to copy the exact implementation of flash-moe, but to make the underlying idea available in the MLX ecosystem: keep dense execution efficient in MLX, while making sparse expert execution streamable, cacheable, and memory-efficient.

Describe alternatives you've considered

I have considered using the original flash-moe project directly. It demonstrates the core idea very well, but it is a custom C / Objective-C / Metal inference engine rather than a general MLX feature, which makes it harder to integrate with existing MLX workflows, model conversion tools, and Python experimentation.

I have also considered using mlx-flash-style weight streaming. This is promising for running models larger than available RAM, but MoE models have a more specific execution pattern: only a subset of experts is needed for each token. A MoE-aware slot-bank or expert-cache design could be more efficient than generic layer or tensor streaming.

Another alternative is using anemll-flash-mlx as an external toolkit. Its design is closely aligned with the desired direction, especially around stable banks, routed expert IDs, and separating hit-path from miss-path behavior. However, having official or first-class MLX primitives for this would make the approach easier to maintain, optimize, and reuse across different MoE architectures.

Finally, I could simply use smaller quantized MoE models that fit fully in unified memory, but that does not solve the main problem: running large sparse MoE models whose total expert weight size exceeds available memory.

Additional context

This feature request is motivated by recent Flash-MoE and MLX-based streaming experiments:

danveloper/flash-moe: demonstrates SSD-streamed expert execution for a very large MoE model on Apple Silicon, using a custom Metal inference engine.
matt-k-wong/mlx-flash: explores flash weight streaming for MLX, including SSD-backed execution for models larger than available RAM.
Anemll/anemll-flash-mlx: explores Flash-MoE inference for large MoE models on Apple Silicon using MLX, with emphasis on stable slot-banks and expert IDs as data.

A first-class Flash-MoE capability in MLX would be very useful for Apple Silicon users who want to experiment with large sparse MoE models locally, without abandoning the MLX ecosystem or maintaining a separate custom inference stack.

## **Additional context**

This feature request is motivated by recent Flash-MoE and MLX-based streaming experiments.

A first-class Flash-MoE capability in MLX would be very useful for Apple Silicon users who want to experiment with large sparse MoE models locally, without abandoning the MLX ecosystem or maintaining a separate custom inference stack.

### References

- `danveloper/flash-moe`  
  https://github.com/danveloper/flash-moe  
  Demonstrates SSD-streamed expert execution for very large MoE models on Apple Silicon using a custom Metal-based inference engine.

- `matt-k-wong/mlx-flash`  
  https://github.com/matt-k-wong/mlx-flash  
  Explores SSD-backed / flash weight streaming for MLX, allowing models larger than available RAM to run by streaming weights from storage.

- `Anemll/anemll-flash-mlx`  
  https://github.com/Anemll/anemll-flash-mlx  
  Explores Flash-MoE inference for large MoE models on Apple Silicon using MLX, with emphasis on expert selection, stable slot-banks, and expert IDs as data.

beatakouchnir:
Closes #986.

MoE models leave most of their parameters idle on any given token. This PR
keeps a configurable fraction of each layer's experts resident in a fixed
slot cache and streams the rest **from the checkpoint's own safetensors** on
demand — mmap slab reads, no converted copy of the model, no write path.
Routing is computed exactly as shipped: a cache miss changes *when* an
expert's weights are read, never *which* expert runs, so accuracy is
preserved by construction and the entire cost is latency.

The key mechanism: with the feature enabled the load stays lazy, and stock
`SwitchGLU` modules are replaced **before** `materialize_lazy_state`, so
non-resident experts are never materialized at any point in the process's
life — which is what lets a model larger than physical memory load at all.

## Measured

`gemma-4-26b-a4b-it-4bit` (30 MoE layers × 128 experts), loaded through the
batched engine's own sequence; 585-token prompt, 256 generated, warm cache:

| residency | memory after gen | process peak | decode tok/s | TTFT | per-request hit rate |
|---|---|---|---|---|---|
| 100% (today's path) | 14.20 GB | 14.84 GB | 122.5 | 0.30 s | — |
| 50% | 7.78 GB | 8.12 GB | 59.4 | 2.9 s | 0.89 |
| 25% | 4.57 GB | 4.85 GB | 40.1 | 9.6 s | 0.67 |
| 12.5% | 2.96 GB | 3.24 GB | 29.3 | 18.1 s | 0.45 |

Peak is the load-transient proof: the full expert set is never held. Greedy
generations at 25% residency are **bit-identical** to the fully resident
model on the acceptance-test prompts.

Behind the design sits a measurement study I ran on my own implementation of
this technique ([full record](https://github.com/beatakouchnir/clausius/blob/main/FINDINGS.md)):
labeled accuracy is flat down to 12% residency (≤0.6pp gsm8k spread across a
4× capacity range; 1.5pp ifeval across 5.3×), and at matched memory (~7.4 GB)
2-bit quantization keeps 0.115 ifeval where 50%-resident offload keeps
0.869. I also cross-validated the pending mlx-lm implementation of the same
family (ml-explore/mlx-lm#1588) with the same method: 57/60 generations
bit-identical, paired-entropy verdict clean, labeled gsm8k at n=200
statistically indistinguishable (McNemar p = 0.61). Two implementations,
same result: the fetch-on-miss family is behaviorally safe.

Two design notes from that record, encoded in the docs page:

* **Pinning a "hot" expert subset is the design to avoid** — measured with
  usage-calibrated pins, zeroing the non-pinned half costs ~91% of gsm8k
  accuracy, because multi-step generation compounds per-token errors.
  Fetch-on-miss keeps the computation exact and pays in latency.
* **Don't acceptance-test this by diffing outputs across residency
  settings.** Capacity changes gather/reduction order, so greedy outputs can
  fork at marginal tokens mid-generation while being semantically exact —
  deterministic at any fixed setting, never at token 0. The valid comparison
  is behavioral (labeled accuracy at n, or paired entropy), which is what the
  test suite does.

## What's in the PR

* `omlx/patches/moe_expert_offload.py` — `ExpertCache` (fixed preallocated
  slots per layer, LRU, in-place slot overwrite so eviction causes no
  alloc/free churn against the pinned buffer pool), `OffloadSwitchGLU`
  (mirrors the stock sort/kernel decisions; chunks any prefill whose expert
  working set exceeds capacity, with per-chunk eval to bound the transient),
  `CheckpointExpertStore` (safetensors header parse + mmap reads — expert
  slabs of stacked tensors, or whole tensors for per-expert layouts; bf16
  travels as uint16), fail-safe discovery (every layer verified
  against the checkpoint — quantized projections, tensor names, shapes —
  and skipped with a logged reason otherwise), `OMLX_MOE_EXPERT_OFFLOAD=0`
  kill switch.
* Engine wiring — lazy load gated on the setting; the wrap runs as its own
  MLX-executor step between `apply_post_load_transforms` and
  `materialize_lazy_state` (#1304 pattern). With the setting off, the load
  path is unchanged. When offload wraps layers, Qwen gate/up fusion is
  skipped with a logged reason (it concatenates stock expert weights in RAM,
  which cannot apply to experts that were never materialized).
* Settings and admin — `ModelSettings` fields (fraction validated to (0, 1]
  at construction), model-specific profile allowlist, admin API
  request/apply/reset, dashboard toggle with a residency selector, and the
  engine-pool runtime signature so toggling reloads (fraction participates
  only while enabled, per the dependent-field convention).
* Tests — 15 fast synthetic tests in the default CI suite (store roundtrips
  incl. bf16 and multi-shard, wrap + forward bit-exactness at partial
  residency in both checkpoint layouts, batch invariance, chunking, every
  skip path including unknown storage dtypes and a per-expert layout with a
  missing expert, kill switch,
  idempotency), 3 slow real-geometry tests (sorted-path bit-exactness at
  gemma dimensions), and a slow live acceptance test that loads the real
  checkpoint offloaded and resident and requires matching generations and a
  fractional memory peak (skips unless the checkpoint is cached).
* `docs/MoE_Expert_Offload.md` — benchmark-first feature doc.

## Numerical contract

Measured against the pinned mlx-lm: decode and unsorted prefill are
bit-identical to the stock path at any residency; the sorted prefill kernel
is presentation-invariant at real model dimensions, so full-residency
prefill is bit-identical too; a prefill chunked below the sort threshold
differs only by the ~4e-3 sorted-vs-unsorted `gather_qmm` kernel rounding.
The tests assert exactly this — bit-exactness where the kernel path is
identical, a rounding-bounded tolerance where it legitimately is not.

## Scope and follow-ups

**Targets** every family whose MoE layers use upstream mlx-lm `SwitchGLU`
with quantized experts (~40 model files — Qwen3.5/3.6 MoE, Gemma MoE,
Mixtral, Llama-4, Kimi, OLMoE, gpt-oss, …): the wrap is type-based, and
every layer is verified against the checkpoint before wrapping (quantized
projections, tensor names, shapes, known storage dtypes), so a family that
does not match — renamed projections (Mixtral's `w1/w2/w3`), per-expert
`bias`, unknown quantization formats — skips with a logged reason and runs
resident exactly as today. Both checkpoint layouts in the wild are
supported: stacked `[E, ...]` tensors and the older one-tensor-per-expert
layout that `sanitize()` stacks at load. **End-to-end verified on
`gemma-4-26b-a4b-it-4bit` (stacked, `switch_glu`) and
`OLMoE-1B-7B-0125-Instruct-4bit` (per-expert, `switch_mlp`; 16 layers,
3.89 GB → 1.17 GB, generations bit-identical)**; the live acceptance test
takes any cached MoE checkpoint, and I'm happy to run it against specific
families reviewers care about. DeepSeek-V4 and GLM-5.2 use oMLX's native switch kernels — a
planned follow-up on the same store, once the mechanism has settled here.
The native macOS app is unaffected (its settings DTOs are optional-field
and ignore unknown keys) and the setting stays editable from the web
dashboard it links to; a native toggle on the app's model-settings screen
is a small follow-up I'm glad to do if wanted.
v1 fetches synchronously on miss, so TTFT at low residency is the pain
point; prefill's full expert-access schedule is computable before any fetch
(run the router over the prompt — no prediction needed), and decode prefetch
has measured LRU→optimal headroom of +17pp hit rate, so an async prefetch
follow-up has a clear, quantified path.


tiagoflino:
**Independent validation: M5 Pro 48 GB + oMLX-native oQ4e checkpoint — acceptance passes**

Ran the acceptance contract on different hardware and a different quantization format than the PR's own test set — passes cleanly.

- **Hardware:** MacBook Pro M5 Pro, 48 GB unified, macOS 26.4
- **Checkpoint:** `Jundot/gemma-4-26B-A4B-it-oQ4e-mtp` (16 GB, oMLX-native oQ4e quant, stacked `[E, ...]` layout) — straight from a local LM Studio model folder, no downloads, no conversion
- **Setup:** `load_model(..., lazy=True, strict=False)` (the checkpoint bundles an MTP head the plain mlx-lm path doesn't know; dropped for this run since it targets the offload mechanism), then `apply_moe_expert_offload(model, path, 0.25)` before `materialize_lazy_state`

Results:

| metric | offloaded @25% | fully resident | limit |
|---|---|---|---|
| active memory after 4×96-token gens | 5.01 GB | 14.65 GB | < 55% ✓ |
| peak memory | 5.13 GB | — | < 75% ✓ |
| generations (4 prompts) | **4/4 bit-identical** | — | ≤ 1 fork ✓ |
| wrapped layers | 30/30 MoE | — | > 0 ✓ |

- Materialized at ~1.8 GB active (model is 16 GB on disk) — the load-peak proof holds: non-resident experts were never materialized
- `misses > 0` and hit rate in range — the cache was actually exercised

The relevant bit for the DeepSeek-V4 / GLM-5.2 follow-up: the store handled oMLX's native oQ4e stacked layout out of the box (a checkpoint that never passed through mlx-lm's `sanitize()`). That suggests the same `CheckpointExpertStore` serves the oQ-family formats the native-kernel models use — the remaining work is the native-kernel `OffloadSwitchGLU` wrap, plus the MTP/DSpark interaction during speculative decode (draft+verify doubles the per-step expert fetches).

beatakouchnir:
Thank you for running this, @tiagoflino! The oQ4e result is new information — I hadn't tested any oMLX-native quant, so the store handling a never-sanitized stacked layout was untested until your run. It de-risks the native-kernel follow-up.

One sharpening: verify runs the drafted window through the full stack in one forward, so each layer needs the union of the window's experts resident at once — batch-like working-set widening, not just extra fetch traffic — which is worth accounting for in cache sizing.

tiagoflino:
Thanks — and agreed on the verify working-set point. Two notes from the follow-up side (the DS4/GLM native-kernel extension built on this mechanism):

1. The offloaded switch inherits the chunking from your design, so a DSpark verify window whose expert union exceeds capacity degrades to chunked execution rather than breaking — safe, just slower. The sizing guidance that matters is the one you name: reserve for the window-union (draft_len × top_k distinct experts per layer), which for Flash 0731's DSpark (7 draft tokens × 6 experts) is ~42/layer — inside a 25% residency capacity (64/layer), but that's the number to put in the docs.

2. Validating on the real `mlx-community/DeepSeek-V4-Flash-0731-2.4bit-mixed` now (MTP off first; DSpark-on sizing as a measured follow-up, since draft+verify doubles fetch traffic on top of the widening). One detail worth a doc line for >RAM checkpoints: the bare-model load needs `strict=False` because the 0731 checkpoint bundles `mtp.*` tensors.

Happy to open the follow-up PR once the live numbers are in.

fxd0h:
ran it on two machines, 24gb m5 pro (16-core gpu, macos 27) and 128gb m5 max (40-core gpu, macos 26.5), gemma-4-26b-a4b-it-4bit. memory matches your table exactly on both, and greedy is 3/4 identical to resident at 0.25 on both (one fork on the canberra prompt, the one your test allows).

the thing worth adding from two very different boxes: the cost is io-bound, not compute. the max has 2.5x the gpu cores (40 vs 16). at 100% residency that only turns into 1.54x decode (81 -> 125 tok/s, memory-bw bound past that, and the 125 matches your resident number). but it collapses as you drop residency:

```
resid   decode 16-core -> 40-core    ttft 16-core -> 40-core
100%    81 -> 125                    1.0 -> 0.5s
50%     37 -> 60                     24 -> 5.9s
25%     35 -> 43                     31 -> 26s
12.5%   27 -> 32                     52 -> 47s
```

so 2.5x the cores is only 1.2x at 25%, and ttft barely moves. at low residency both boxes just hit the ssd floor. misses go 30k (50%) -> 155k (25%) -> 361k (12.5%) on this prompt, thats what decode and ttft track, not the gpu. more gpu doesnt buy you out of low-residency latency.

two follow-ons maybe worth a doc line. ttft scales with prompt length, yours is 585 tok at 9.6s @25%, this is ~1.8k tok at 26-31s @25% on the same hardware class, longer prefill routes through more experts. so long-context pays most of the cost up front. decode stays fine (35-43 tok/s @25%), its prefill thats expensive.

also works with response_format. served it at 0.25 (rss 4.5gb so actually streaming, not resident) plus a strict json_schema through xgrammar, valid schema-conforming output, GrammarCompiler active in the log. so a 26b doing constrained json in ~4.5gb as a batching server, which the standalone streaming engines dont give you.

full per-residency json below if useful.

jundot:
Thanks for running the extra hardware checks. They show that the basic offload path works on the checkpoints tested and that the memory savings are real.

I still hit a mixed-quantization crash. `ExpertCache` takes `group_size`, `bits`, and `mode` from `gate_proj` and reuses them for the other projections. With 4-bit gate/up projections and an 8-bit down projection, `gather_qmm` fails because the weight and scale shapes no longer match. Each projection needs its own quantization metadata, plus a regression test using mixed bit widths.

The default server path also needs coverage. The Gemma 4 checkpoint in the docs is detected as a VLM, but this PR only installs offload in `BatchedEngine`. The normal `VLMBatchedEngine` path does not apply it.

EnginePool still checks the full checkpoint size as well. That means a model over the memory ceiling is rejected before offload has a chance to reduce its resident footprint.

Before merge, please add tests for those three paths: mixed-bit projections, normal Gemma 4 VLM dispatch, and EnginePool admission when the full checkpoint is over the limit but the offloaded model should fit.

tiagoflino:
Thanks for the thorough review @jundot — the mixed-bit point is a real one, and it bites our follow-up too, so we picked it up there.

We're stacking a follow-up PR (https://github.com/beatakouchnir/omlx/pull/1) on #2595 that adds DeepSeek V4 / GLM-5.2 coverage to the same store. To keep this PR's scope tight we moved the mixed-bit fix into the follow-up, with the regression test you asked for:

- `ExpertCache.meta[name] = (group_size, bits, mode)` per projection; `qmm`/`qmm_native` read the projection's own metadata
- the native block path requires all projections to share one qualifying combo — mixed formats fall back to `gather_qmm` (same contract as the stock `all()` check on `native_kinds`)
- regression test: synthetic DS4 with 3-bit gate/up + 8-bit down affine — wraps, per-projection metadata asserted, decode and sorted prefill bit-exact vs the resident twin (CI-runnable, no checkpoint download)

On the other two items — VLMBatchedEngine dispatch and EnginePool admission — those are engine-wiring scope; happy to help either way:
- we have the Gemma 4 oQ4e checkpoint locally (VLM-detected, same one from the docs), so we can exercise the `VLMBatchedEngine` path as a test bed if useful
- for EnginePool, the "offloaded footprint" estimate is exactly what our live runs measure (18.1 GB active vs 92.8 GB on disk at 15% residency on a 48 GB box), so we can validate the admission threshold change against real numbers

Let us know what's most useful — otherwise the follow-up PR is ready for review whenever #2595 settles.

tiagoflino:
Update: the follow-up PR is now open in the upstream repo (carrying this PR's commits in the same branch, so the complete feature lands in one place — will rebase to drop the duplicates once #2595 merges):

https://github.com/jundot/omlx/pull/2616

The mixed-bit fix mentioned above is included there with its regression test.

beatakouchnir:
@jundot thank you for the review — all three are addressed, one commit per item.

**Mixed-bit projections** (bd49b7d): confirmed. `ExpertCache` took `group_size/bits/mode` from `gate_proj` and applied them to all three projections. It now keeps the triple per projection and passes the right one to each `gather_qmm`. Regression test quantizes gate/up at 4-bit and down at 8-bit and requires bit-exact decode and prefill against the resident module.

**Gemma 4 VLM dispatch** (da21860): right again — discovery matched with `type(obj) is SwitchGLU` against mlx-lm's class, but mlx-vlm defines its own `SwitchGLU`, so the VLM engine (the default for Gemma 4 checkpoints) never wrapped anything. Matching is now by class name plus the attribute contract, and `VLMBatchedEngine` got the same lazy-load → wrap → materialize sequence and gate/up-fusion skip as `BatchedEngine`. Two tests: a fast one that wraps a `SwitchGLU`-named class that isn't mlx-lm's (bit-exact output), and a slow live one that loads gemma-4-26b-a4b-it-4bit through `mlx_vlm.utils.load` and requires wrapped layers, finite logits, and cache traffic through the wrapped forward.

**EnginePool admission** (c81366c): when offload is enabled for the model, admission now uses an offload-adjusted estimate — a safetensors header scan totals expert bytes (stacked 3-D `{gate,up,down}_proj` tensors under any container name, or per-expert `.experts.<n>.` names) and subtracts the non-resident share. Any failure falls back to the full size, so admission never gets more permissive by accident. Covered by a unit test over a real checkpoint (named `switch_mlp`, so the matching is shape-based rather than tied to one container name) plus a pool test where a model over the ceiling admits with offload enabled and still refuses without it.

I also merged main to clear the conflict with the distributed-cluster admission changes (70be6bf). One deliberate interaction: the offload discount only applies when `deployment is None` — a rank-zero shard's planned size shouldn't be reduced by whole-checkpoint expert bytes.

Fast suites pass (`test_moe_expert_offload`, `test_engine_pool`, `test_batched_engine`, `test_vlm_engine`), and the slow live acceptance tests were re-run against the real checkpoint after the merge.

beatakouchnir:
@tiagoflino saw #2616 — thanks for keeping the attribution clear and stating the rebase intent up front.

One thing to flag so the two threads don't diverge: the three review items above are already fixed in this PR — bd49b7d (mixed-bit, per-projection `qparams` with a 4/8-bit bit-exact regression test), da21860 (VLM dispatch), c81366c (EnginePool admission) — and the branch is conflict-free and re-validated as of today. #2616's copy of this branch predates those commits, so it carries an alternative mixed-bit implementation (af3d3e5) and none of the VLM or admission work; as it stands it wouldn't satisfy the review above.

Given that, the clean sequencing seems to be: this PR merges as the reviewed foundation, and #2616 becomes the DS4/GLM delta — which is the part that genuinely needs its own review. You don't have to wait for the merge to get there: rebasing #2616 onto this branch now drops the duplicates and the divergent mixed-bit fix. The native-path constraint you described (all projections must share one qualifying combo, else fall back to `gather_qmm`) composes cleanly with per-projection `qparams`, so the rebase should be mechanical.

The DSpark window-union number (~42/layer for 7×6) and the `strict=False` note on the 0731 checkpoint's `mtp.*` tensors are both worth a doc line — happy to fold them in here if jundot prefers docs to travel with the mechanism, otherwise they fit naturally in #2616. And a real-number check of the admission estimate on your 48 GB box would be genuinely useful.

tiagoflino:
Two updates:

**Rebase done.** #2616 now carries the review-fixed branch (bd49b7d per-projection `qparams`, da21860 VLM dispatch, c81366c admission) plus our DS4/GLM delta — our alternative mixed-bit implementation was dropped in favor of the canonical `qparams`, and the native-path constraint (all projections share one qualifying combo, else fall back to `gather_qmm`) is implemented on top of it. 197 tests pass (base + VLM + pool + DS4 + our optimizations). When #2595 merges we'll rebase again to drop the duplicates and #2616 becomes the pure DS4/GLM delta.

**Admission estimate vs real measurements** (DeepSeek-V4-Flash-0731-2.4bit-mixed, 92.8 GB on disk, M5 Pro 48 GB):

| residency | `estimate_offload_admission_bytes` | measured active (mx.get_active_memory) |
|---|---|---|
| 15% | 19.9 GB | 18.1 GB (peak 18.3) |
| 40% | 41.3 GB | 37.5 GB (peak 38.7) |

The estimate runs ~10% above measured at both points — a healthy conservative margin (the measurement excludes some lazy allocations), and the model admits on this box at both fractions. Note the measured numbers include the KV/pooling caches for a short context; long-context runs would close part of the gap.

Also happy to fold the DSpark window-union (~42 experts/layer for a 7×6 verify) and the `strict=False` note on the 0731 checkpoint's `mtp.*` tensors into #2616's docs — just say which side you'd rather carry them.

beatakouchnir:
Following up on the admission-estimator finding from https://github.com/jundot/omlx/pull/2616#issuecomment-5302692350 — that code is from this PR (c81366c, carried unchanged into #2616), so the fix belongs here. Pushed in 94985ef:

- The estimate is now derived from the same structural rules `apply_moe_expert_offload` enforces, so it cannot promise savings the wrapper won't deliver: a container counts only when all three `{gate,up,down}_proj` projections are present **with quantization scales** in a supported layout. Renamed layouts (`w1/w2/w3`) and unquantized checkpoints match nothing and discount nothing — your 4,000→1,750-byte repro now returns the full size.
- Per-layer savings honor the runtime's minimum-eight capacity floor (`max(8, min(E, round(E·fraction)))`), so small fractions no longer under-report the resident share.

Regression tests cover the `w1/w2/w3` repro shape, an unquantized checkpoint, and a floor-binding fraction (0.05 on 32 experts → 75% max savings, not 95%).

Separating the DS4/GLM compatibility work from the prefetch/pinning changes sounds right to me as well — it keeps this base PR reviewable on its own terms.

fxd0h:
rebased the branch onto 0.6.0 since main moved under it and the PR went conflicting. result is on my fork: `fxd0h/omlx` branch `feature/moe-expert-streaming-0.6.0` — @beatakouchnir cherry-pick it or give me push on your fork and i update the PR branch directly.

two findings beyond the mechanical rebase, both fixed there:

1. a linear rebase silently drops the `deployment is None` guard around the admission discount, because it lives in merge commit 70be6bf4 and not in any replayed commit. without it a distributed shards planned size gets discounted by whole-checkpoint expert bytes and an over-ceiling cluster load admits with no eviction. restored byte-identical to the pre-rebase tip.
2. `test_moe_offload_admits_oversized_moe` fails since 94985ef: the estimator now requires all three projections quantized and floors capacity at 8 experts, but the fixture still wrote a lone gate_proj.weight with 4 experts. fixed with a 32-expert quantized container (2400 expert bytes, saves 1800, admits under the 1500 ceiling).

local run on the rebased branch: test_engine_pool 130 passed, offload suites 39 passed, settings template 8 passed. @jundot with your three items addressed and this, ready for re-review.

fxd0h:
third find, this one live on our fleet: with offload enabled the VLM path dies on the first request with `RuntimeError: There is no Stream(gpu, 1) in current thread`. the caches slot maps and resident slots live on plain attributes, so materialize_lazy_state never reaches them and they stay bound to the loader threads stream. fix on the same branch (821d84f1): materialize_offload_state walks the wrapped layers and evals map + resident slots on the executor that ran the apply, wired into both engines. regression test rebuilds the loader/inference thread split. with it a 24GB M5 Pro serves gemma-4-26B at 10.1GB resident (fraction 0.5) instead of 16GB — first time the feature runs end to end on the vision path for us.

jundot:
Thanks — `94985ef` fixes the original `w1/w2/w3` repro and the minimum-eight capacity floor. I also verified the VLM state-materialization fix in `821d84f1`: the current PR head reproduces `There is no Stream(gpu, 1) in current thread` on the first Gemma 4 request, while the rebased candidate wraps all 30 layers and completes the same request successfully.

One admission edge remains. For per-expert layouts, the estimator unions field names and expert IDs independently. A checkpoint with one complete expert and 31 experts containing only `gate_proj.weight` was estimated at 972,736 bytes from a 1,000,000-byte full size, even though `apply_moe_expert_offload()` rejected the layer and wrapped zero modules. Please validate the full required field set for every expert before applying a discount.

The PR head is still `94985ef` and conflicting; `bce55e50`, `68090c9c`, and `821d84f1` are currently only on the rebased fork branch. Once those are carried over and the per-expert admission case is covered, this base PR should be close to mergeable. The DS4, prefetch, and pinning work can remain separate in #2616.

beatakouchnir:
@fxd0h, thank you — I've adopted your branch as the PR head and force-pushed; your commits keep their authorship, so no fork access needed. The `deployment is None` guard and the VLM materialization fix are both carried. The engine-pool fixture was my miss: I had run the offload suite with the estimator change but not the pool suite. Both run together now.

@jundot, the per-expert admission edge is fixed in 5309c8a6: field completeness is tracked per expert, and a container discounts only when every expert has all three quantized projections — the same rule`apply_moe_expert_offload()` enforces. Your repro (1 complete + 31 gate-only experts) is a regression test returning the full size, with a fully-complete control asserting the discount and capacity floor.

Suites on the new head: offload + engine-pool 153 passed, settings template 8 passed. Agreed on scope — DS4, prefetch, and pinning stay in #2616.

beatakouchnir:
Rebased onto 0.6.2 — conflicts with the ANE tuner settings surface (#2814) resolved by keeping both feature blocks; no functional changes (13/14 patches byte-identical under range-diff, settings/admin tests green).

beatakouchnir:
Status note: rebased twice this week to stay current with main — onto 0.6.2 after the ANE tuner (#2814) landed, and onto 0.6.3rc1 today. Both conflicts were adjacent insertions in the settings/admin surfaces (the offload settings and the tuner settings want the same anchors); resolved keep-both, no functional changes to the offload path itself. Settings/admin/engine-pool suites pass locally.

No rush — I'll keep the branch conflict-free through the 0.6.3 churn.

beatakouchnir:
Rebased onto 0.6.3rc2 — flagging one interaction with code that landed this week, rather than just conflict noise.

`_entry_runtime_resident_size()` and this PR's offload discount both adjust `admission_size` in `_ensure_capacity`. They compose: CPU-share bytes are added first, then the offload discount applies for local loads only (`deployment is None`, so a distributed shard is never discounted by whole-checkpoint expert bytes). That order is safe because `estimate_offload_admission_bytes` returns `full_size - saved` — a relative discount on whatever size it is handed — so the CPU-share addition survives.

Engine-pool, settings and admin-template suites pass locally (419).

beatakouchnir:
@jundot, picking this back up now that main is settled. On the 16th you said it would be close to mergeable once the fork commits were carried over and the per-expert admission case was covered — both are done. The fork branch is the PR head now, and 84991ad makes the offload discount require every expert to be complete before it applies, so the one-complete-plus-31-partial case you caught is rejected rather than under-counted. It has since been rebased current onto 0.6.3rc2 and is mergeable, with DS4/prefetch/pinning still parked in #2616 per your note. Whenever you have a moment, a second look would be very welcome.

valerio-castelli:
Hello! Thanks for working on this effort @beatakouchnir, it is super valuable. 

I see the PR is now again out-of-sync with main and requires some conflicts to be resolved: do you have an estimate on how you might be able to take a look, so that @jundot can then move forward with merging this upstream? Thank you!

beatakouchnir:
Thanks @valerio-castelli, glad it's useful. Rebased onto current main and pushed; the only real conflict was a single hunk in `omlx/engine/vlm.py` (the offload lazy-load now threads into the new `load_kwargs` path), and everything else — engine-pool, model profiles, the settings/admin surfaces, and the suites — merged cleanly through the recent engine churn. The branch is conflict-free again and the offload/engine-pool/settings tests pass locally.

@jundot, it's mergeable again; whenever you have a window, a look would be very welcome so it can land. DS4/prefetch/pinning stay parked in #2616 per your earlier note.

beatakouchnir:
Rebased onto latest `main` (was ~84 commits behind) — clean, no conflicts, and the suite is green locally: 286 passed / 5 skipped, including the engine-pool, model-settings, and MoE tests this touches.

One thing that's probably behind the confusing checks state: CI has never actually run on this PR. The workflow runs sit in `action_required` (0 jobs, 0s) because fork PRs need a maintainer to approve the run — so "no checks" here is blocked, not passing. A single *Approve and run workflows* would let it execute.

Happy to keep it rebased or split it up if that would make review easier. It's also adjacent to the Flash-Next MoE serving work landing now, so it may be more useful to have in than when it was opened.

jundot:
Thanks for implementing this. The lazy-load replacement keeps non-resident experts out of memory, and I verified the Gemma offload path locally. I'll add the DeepSeek V4.1 and Qwen3.8-Flash-Next support I've tested in a follow-up, along with the settings wiring, memory accounting fixes, and MTP exclusion. I'll also show the experimental toggle only for supported model types with compatible checkpoint layouts. Merging this first, then applying the follow-up.

beatakouchnir:
Thank you, @jundot! Happy to help with the Flash-Next / DeepSeek-V4.1 follow-up if that would be useful: I've been running large MoEs with expert offload on a 128 GB Mac (Qwen3.5-397B-A17B end-to-end, decode bit-identical to resident; Qwen3.8-Flash-Next in progress), so the checkpoint-layout gating and MTP exclusion you mention are right in what I've been working on.

PR #2601, opened 2026-08-11
fix: keep cross-session messaging enabled for omlx launch claude

fparrav:
## Summary
- `omlx launch claude` sets `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC=1` to quiet telemetry.
- That variable also disables the feature-flag evaluation that [cross-session messaging](https://code.claude.com/docs/en/cross-session-messaging) (`ListAgents`/`SendMessage`) depends on.
- Result: any Claude Code session started via `omlx launch claude` runs fine but never appears as a reachable peer to other local Claude Code sessions, silently, with no error.
- Drop the env var so an omlx-launched session behaves the same as a plain `claude` launch or one pointed at another local backend, both of which already register correctly.

## Test plan
- [x] `pytest tests/test_integrations.py -k Claude` — added `test_launch_keeps_cross_session_messaging_enabled` asserting none of `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC`, `DISABLE_TELEMETRY`, `DO_NOT_TRACK`, `DISABLE_GROWTHBOOK` are set
- [x] `pytest tests/test_integrations.py` (full suite, 102 passed)

### DIFF
--- a/omlx/cli.py
+++ b/omlx/cli.py
@@ -531,6 +531,7 @@ def launch_command(args, extra_args: list[str] | None = None):
         reasoning=model_info.get("enable_thinking"),
         tools_profile=getattr(args, "tools_profile", "coding"),
         extra_args=tuple(extra_args or ()),
+        cross_session=getattr(args, "cross_session", False),
     )
 
     # Launch
@@ -1081,6 +1082,17 @@ Example directory structure:
         type=str,
         default=None,
         help="Claude Code Haiku tier model (Claude integration only)",
+    )
+    launch_parser.add_argument(
+        "--cross-session",
+        action="store_true",
+        default=False,
+        help=(
+            "Allow the launched session to be reachable via Claude Code's "
+            "cross-session messaging (ListAgents/SendMessage). This requires "
+            "enabling telemetry and feature-flag traffic to Anthropic that is "
+            "otherwise kept disabled by default (Claude integration only)."
+        ),
     )
 
     # Diagnose command
--- a/omlx/integrations/base.py
+++ b/omlx/integrations/base.py
@@ -28,6 +28,7 @@ class IntegrationContext:
     reasoning: bool | None = None
     tools_profile: str = "coding"
     extra_args: tuple[str, ...] = ()
+    cross_session: bool = False
 
     @property
     def base_url(self) -> str:
--- a/omlx/integrations/claude.py
+++ b/omlx/integrations/claude.py
@@ -11,6 +11,20 @@ from omlx.utils.install import get_cli_command_prefix
 
 CLAUDE_CODE_MIN_CONTEXT_WINDOW = 48 * 1024
 
+# Third-party opt-outs that, like CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC, also
+# gate the feature-flag evaluation cross-session messaging depends on. Checked
+# rather than overridden: if the user already set one of these themselves, we
+# respect that choice instead of silently re-enabling Anthropic-bound traffic.
+_CROSS_SESSION_BLOCKING_VARS = (
+    "DISABLE_TELEMETRY",
+    "DO_NOT_TRACK",
+    "DISABLE_GROWTHBOOK",
+)
+
+
+def _env_flag_set(env: dict[str, str], name: str) -> bool:
+    return env.get(name, "").strip().lower() not in ("", "0", "false", "no")
+
 
 def claude_code_model_disabled_reason(model_info: dict) -> str | None:
     context_window = model_info.get("max_context_window")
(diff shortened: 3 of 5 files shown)


fparrav:
## ✅ Cross-session messaging test verified

Tested on current branch `fix-claude-launch-cross-session-messaging`:

- **From**: session `omlx-fix` (launched via `omlx launch claude` with this fix)
- **Test**: Sent a test message from `omlx-fix` → received successfully
- **Reply**: Sent confirmation back from this session → `omlx-fix` reached via `SendMessage`

Both directions of cross-session messaging work correctly with this fix applied. The `messagingSocketPath` is properly exposed so external Claude Code sessions can address each other.

**Steps to reproduce**:
1. `omlx launch claude` (launched session `omlx-fix`)
2. From another Claude Code session: `SendMessage({to: 'omlx-fix', message: 'test'})`
3. Message received and reply sent back successfully

Note: PR has merge conflicts with `main` — may need rebase before merging.

fparrav:
Manual verification on top of the added unit test:

1. Started the oMLX server locally.
2. Launched `omlx launch claude` from this branch (editable install) in a separate terminal.
3. From another local Claude Code session, ran `ListAgents` — the new session showed up within ~24s of starting (previously it never appeared).
4. Sent it a message via `SendMessage` — delivered and acknowledged, confirming bidirectional cross-session messaging now works for sessions launched through `omlx launch claude`.

Before this fix, step 3 never succeeded — the session stayed alive but was permanently invisible to `ListAgents`/`SendMessage`.

jundot:
Thanks for tracking down why cross-session messaging is unavailable. I understand why removing `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC` fixes it: Claude Code needs telemetry and feature-flag fetching enabled before it exposes the messaging feature.

The variable was originally set intentionally because `omlx launch claude` is primarily used with a local model. In that setup, users reasonably expect optional Anthropic-bound traffic to remain disabled. Removing the umbrella variable for every launch changes that default and allows unrelated services such as update checks, error reporting, feedback, surveys, and other availability checks—not just cross-session messaging.

I'd prefer to preserve the current privacy-oriented default and make this an explicit opt-in, for example `omlx launch claude --cross-session`. The option's help text should explain that cross-session messaging requires telemetry and feature-flag traffic to be enabled.

In the opt-in path, please omit `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC`, while keeping unrelated services disabled through granular variables such as `DISABLE_AUTOUPDATER`, `DISABLE_ERROR_REPORTING`, `DISABLE_FEEDBACK_COMMAND`, and `CLAUDE_CODE_DISABLE_FEEDBACK_SURVEY`. If the user already has `DISABLE_TELEMETRY`, `DO_NOT_TRACK`, or `DISABLE_GROWTHBOOK` in their environment, oMLX should preserve that choice and warn that cross-session messaging will remain unavailable. This keeps the existing local-first behavior while still allowing users who want cross-session messaging to enable it deliberately.

fparrav:
Agreed, that's the right approach — it preserves the existing privacy-first default and makes the trade-off explicit for users who want it, rather than changing behavior for everyone. Pushed an update along those lines:

- `omlx launch claude` keeps `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC=1` by default, unchanged from before this PR.
- New `--cross-session` flag opts in to cross-session messaging. In that path, the umbrella variable is left unset, and the unrelated services you listed stay disabled individually via `DISABLE_AUTOUPDATER`, `DISABLE_ERROR_REPORTING`, `DISABLE_FEEDBACK_COMMAND`, `CLAUDE_CODE_DISABLE_FEEDBACK_SURVEY`.
- If the user already has `DISABLE_TELEMETRY`, `DO_NOT_TRACK`, or `DISABLE_GROWTHBOOK` set in their own environment, `--cross-session` no longer overrides it — it's left as-is and a warning is printed that messaging will stay unavailable because of it.

Added test coverage for all three paths (default, `--cross-session`, and `--cross-session` with a pre-existing blocking var) plus a CLI-level test asserting the flag threads through to `IntegrationContext`. Full `tests/test_integrations.py` + `tests/test_cli.py` pass (180 tests).

jundot:
Thanks for the update. The privacy-first default is preserved, `--cross-session` makes the telemetry and feature-flag trade-off explicit, and the granular opt-outs keep unrelated services disabled. LGTM, merging. I'll fold the remaining blocker-warning edge cases (`DISABLE_TELEMETRY=0/false` and an inherited `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC`) into a small follow-up on main.

PR #2602, opened 2026-08-11
Expose is_favorite and is_hidden in /v1/models/status response

monroewilliams:
Minimal change to include `is_favorite` and `is_hidden` flags in the response to the /v1/models/status endpoint. 
I have some [scripts](https://github.com/monroewilliams/claude-local) and [plugins](https://github.com/monroewilliams/pi-local) that would like to be able to display which models are favorited, and this information doesn't seem to be otherwise available through the API. It would also be useful to be able to filter out hidden models without making a second call to `/v1/models` just to see the visible set.

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -2645,6 +2645,7 @@ async def list_models_status(_: bool = Depends(verify_api_key)):
             m["max_context_window"] = None
             m["max_tokens"] = None
             m["is_favorite"] = False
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for this. The two flags come directly from the base model settings, which keeps regular models and exposed profiles consistent while preserving explicit false defaults. LGTM, merging.

PR #2605, opened 2026-08-12
fix(oq): discover HF cache models for quantization

monroewilliams:
## Problem

Non-MLX, unquantized models in the HF Hub cache were invisible as quantization sources. The oQ manager only scanned 2 directory levels deep, but HF cache entries have a 3-level structure: `models--Org--Name/snapshots/<hash>/`.

Quantization is the natural way to convert these models to MLX format, so they should be available.

## Changes

- **oq_manager.py**: Resolve HF cache snapshot paths via `_resolve_hf_cache_entry` during quantization source scanning
- Use `source_repo_id` (Org/Name) for display instead of commit hashes or dash-separated IDs
- Filter out models without `model_type` (MLX cannot resolve the model class)
- **Web dashboard + macOS UI**: Show canonical HF repo name in the quantization model picker

### DIFF
--- a/apps/omlx-mac/Sources/AppView/Screens/QuantizationScreen.swift
+++ b/apps/omlx-mac/Sources/AppView/Screens/QuantizationScreen.swift
@@ -248,7 +248,7 @@ private struct SourceModelSection: View {
                                               defaultValue: "Select a model…",
                                               comment: "Placeholder option in the source-model dropdown"))]
         opts += models.map { m in
-            PopupOption(value: m.path, label: "\(m.name) (\(m.sizeFormatted))")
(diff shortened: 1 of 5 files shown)


jundot:
Thanks for adding HF cache models to the oQ picker. Reusing the existing HF snapshot resolver keeps discovery consistent, and the focused Python and macOS DTO tests pass. One thing I'll fold into a follow-up is preserving the canonical repo name through task and output naming, and excluding bin-only entries that the safetensors-only quantizer cannot process. The Python 3.13 failure is unrelated to this diff, and the failing test passes locally. LGTM. I'll merge this and handle those follow-ups on main.

Issue #2457, opened 2026-08-01
Streaming: truncated thinking is re-emitted as `content` when finish_reason is `length`


mabaeyens:
## Environment

- oMLX 0.5.1
- Model: `mlx-community/Qwen3.6-35B-A3B-4bit`
- API: streaming `/v1/chat/completions` with `chat_template_kwargs: {"enable_thinking": true}`
- Apple M5, 32GB, macOS 26.6

## What happens

If a thinking turn gets cut off by `max_tokens` while the model is still inside the
think block, the final flush emits the whole accumulated reasoning again as a single
`content` delta. The client has already received that same text as
`reasoning_content`, so it now shows the entire chain of thought as the answer.

I hit this from a downstream client that was showing users a wall of "The user is
asking about..." instead of a reply, and traced it back here.

## Reproduction

Same prompt each time, only `max_tokens` changes.

```python
body = {
    "model": "Qwen3.6-35B-A3B",
    "messages": [{"role": "user", "content": "Summarise in three bullets what a Kalman filter does."}],
    "stream": True,
    "max_tokens": 400,
    "chat_template_kwargs": {"enable_thinking": True},
}
```

Counting deltas by which field they carry:

| max_tokens | finish_reason | reasoning_content deltas | content deltas | content chars | reasoning chars |
|---|---|---|---|---|---|
| 400 | `length` | 33 | 1 | 1815 | 1815 |
| 2000 | `stop` | 61 | 14 | 884 | 3388 |

In the truncated run there is exactly one `content` delta and it carries 1815 chars,
byte for byte what already went out across the 33 `reasoning_content` deltas. The
clean run splits the two properly. I ran it four times across two prompts; the shape
holds.

## Where it comes from

`ThinkingParser.finish()` in `omlx/api/thinking.py` (line 326, recovery block at
350-358). When the parser started inside a thinking block, never saw `</think>`, and
never emitted content, it re-emits the accumulated thinking as content so the answer
body is not empty.

That recovery is deliberate and the comment above it says so, and for its intended
case it is the right call: the model stopped on its own without closing the tag, so
the text it produced is all the answer there is. #903 is that case.

The truncated case is different. When `finish_reason` is `length` we know exactly why
the close tag never arrived: we cut the model off mid-thought. The reasoning is
reasoning; there is no answer yet, and treating it as one is worse for the client
than an empty content field, because the user cannot tell it apart from a real reply.

One thing worth flagging: #82 says "the streaming path handles this correctly via the
[...]

Issue #903, opened 2026-04-22
Qwen3.6-35B-A3B-4bit returns thinking-only responses with no visible content via OpenAI-compatible API


jsirish:
## Environment

- **oMLX version:** v0.3.7rc2
- **Model:** `mlx-community/Qwen3.6-35B-A3B-4bit`
- **Preset:** qwen3.5/6 (r,code)
- **TurboQuant:** enabled
- **SpecPrefill:** enabled, draft model `qwen3-0.6B-4bit`
- **API:** OpenAI-compatible `/v1/chat/completions`
- **Hardware:** Apple Silicon (M-series)

## Description

When using the Qwen 3.6 35B MoE model through the OpenAI-compatible API, the model consistently returns responses containing only `reasoning_content` (thinking) with no actual `content` field. This results in the downstream client (OpenClaw gateway) seeing an empty/thinking-only response and surfacing an "incomplete turn" error.

**Qwen 3.5 35B MoE (`Qwen3.5-35B-A3B-4bit`) works correctly** with the same configuration — it returns both `reasoning_content` and `content` in the response.

## Observed behavior

The API response includes a `choices[0].message` with:
- `reasoning_content`: populated (contains the model's internal thinking)
- `content`: **missing or empty**
- `stop_reason`: `stop`

This happens consistently across multiple prompts and session types — both direct chat and embedded/automated runs. Switching from "thinking" mode to "instruct" mode did not resolve the issue (response came back as completely empty `content: []`).

## Expected behavior

The model should produce a `content` field with the actual response text, similar to how `Qwen3.5-35B-A3B-4bit` behaves under the same preset and server configuration.

## Steps to reproduce

1. Load `mlx-community/Qwen3.6-35B-A3B-4bit` with the qwen3.5/6 preset, turboquant, and specprefill (draft: qwen3-0.6B-4bit)
2. Send a chat completion request via the OpenAI-compatible API endpoint
3. Observe the response — `reasoning_content` is populated but `content` is empty/missing

## Comparison data

| Model | Response types | Has visible text |
|-------|---------------|-----------------|
| `Qwen3.5-35B-A3B-4bit` | `['thinking', 'text']` | ✅ Yes |
| `Qwen3.6-35B-A3B-4bit` | `['thinking']` or `[]` | ❌ No |

## Additional context

- The same oMLX instance serves both 3.5 and 3.6 models — only the model swap triggers the issue.
[...]

Issue #1348, opened 2026-05-22
Using oMLX 0.3.9 as a agent in OpenClaw yields no response


maxh1975:
**Describe the bug**
Currently, when using oMLX (0.3.9) as a agent for OpenClaw (2026.5.20), the tool can be invoked and run, yet no final execution result is returned. The prompt shown in Feishu is: ⚠️ Agent couldn't generate a response. Note: some tool actions may have already been executed — please verify before retrying.
 
In contrast, oMLX 0.3.8 and 0.3.9 dev2 versions function normally.

**Desktop (please complete the following information):**
 - macOS Version: [26.4.1]
 - oMLX Version [0.3.9]
 - mac mini m4 pro 64G
 - model：Qwen3.6-27B-oQ6-mtp

fry69:
Try using current `main` if possible, there is a fix for the Python environment that might mess up instantiated clients like `OpenClaw`, see -> https://github.com/jundot/omlx/commit/2f2f5087a9c9a6ef71fa165da4a299bd19d4d5b4

Otherwise wait for the next bug fix release.

maxh1975:
@fry69 Thanks. I'll wait for the next version update and close this issue for now.

jundot:
@maxh1975 Thanks for the report. Could you bump the global log level to `trace` in oMLX settings, reproduce the failing turn once, and share the relevant excerpt from `~/.omlx/logs/server.log` (feel free to scrub anything sensitive, like prompts, tool args, file paths)? Knowing the model you're serving and whether OpenClaw hits `/v1/chat/completions` or `/v1/messages` would also help a lot. Two changes between 0.3.9.dev2 and 0.3.9 are my top suspects and the trace log should tell me which one.

maxh1975:
@jundot This is my operation log for version 0.3.9. I switched the model to Qwen3.6-27B-UD-MLX-NVFP4, which is my regular choice. It can be invoked normally and return results steadily on OpenClaw using omlx0.3.8 and 0.3.0 dev2. Hope this information helps. Feel free to let me know if you need further details. I simply asked: 当前本地跑的模型?

2026-05-24 19:52:00,103 - omlx.server - DEBUG - [-] - Responses API request: model=Qwen3.6-27B-UD-MLX-NVFP4, stream=True
[...]

PR #2607, opened 2026-08-12
fix(thinking): don't leak truncated thinking into content (finish_reason=length)

### LINKED ISSUE #2606: DeepSeek-V4-Flash-0731: thinking leaks into content when generation is truncated mid-thought (finish_reason=length)
## Summary

When a DeepSeek-V4-Flash-0731 generation is truncated by `max_tokens`
(`finish_reason == "length"`) while the model is still inside its thinking
block, the entire chain of thought is surfaced as the visible answer. Both
the streaming and non-streaming paths leak the reasoning into `content`.

## Environment

- oMLX `0.5.8.dev3`
- Model: `DeepSeek-V4-Flash-0731-oQ4e-mtp` (oQ4e, MTP)
- API: `/v1/chat/completions` (both streaming and non-streaming), OpenAI-compatible
- Client: any OpenAI chat client (reproduced standalone; also seen via ChatWise)

## Reproduction

Single turn, no tools, a request that makes the model think at length
(e.g. a long comparative summary). The model drafts its answer inside the
thinking block and, when cut off by `max_tokens` before emitting the close
tag, oMLX re-emits the whole accumulated thinking as content.

Streaming (max_tokens=1500):
```
reasoning_content: <entire thinking draft, e.g. "1. **Analyse the request** ...">
content:           <same text, byte-for-byte identical>
```

Non-streaming (max_tokens=1500):
```
reasoning_content: (empty)
content:           <the thinking draft>
```

Observed with the same prompt on DS4 (`deepseek-v4-flash`) — which handles
the truncated-mid-thought case cleanly — the leak does **not** occur.

## Root cause

Both paths share the same recovery logic that treats "opened thinking but no
close tag arrived" as "the model chose to answer without closing":

- Streaming: `ThinkingParser.finish()` in `omlx/api/thinking.py` re-emits the
  accumulated thinking as content when `_in_thinking and not _close_seen and
  not _content_emitted and _thinking_accumulated`. This is the deliberate
  recovery for issue #903 (a model that stops on its own without closing the
  tag). But when `finish_reason == "length"` the tag is missing **because we
  truncated the model mid-thought**, not because it chose to answer. The
  reasoning is reasoning; there is no answer yet. Surfacing it as content
  leaks the chain of thought into the visible reply.

- Non-streaming: `extract_thinking()` classifies a tag-free body as content
  (the `("", text)` fallback), so a truncated draft becomes the answer.

Issue #2457 already describes the streaming half of this bug and proposes the
same fix (a `truncated` flag on `finish()`).

## Proposed fix

1. Add a `truncated` flag to `ThinkingParser.finish()`; when set, skip the
   recovery branch so truncated thinking is never re-emitted as content.
2. Pass `truncated=last_output.finish_reason == "length"` at the three
   streaming call sites (OpenAI chat, Anthropic messages, Responses API).
3. Add `reclassify_truncated_thinking()` for the non-streaming paths: when
   `finish_reason == "length"`, the prompt opened a thinking block, and no
   close tag is present, classify the whole body as thinking instead of
   content. Wire it into the three non-streaming paths.

I have a working patch with unit tests for all of the above.

solodddd:
## Summary

Fixes thinking leaking into the visible answer when a generation is truncated
mid-thought by `max_tokens` (`finish_reason == "length"`). Closes #2606 and
addresses the streaming half already described in #2457.

When the model is still inside its thinking block when `max_tokens` cuts it
off, oMLX re-emits the entire accumulated reasoning as `content`:

- **Streaming**: `ThinkingParser.finish()` treats "opened thinking, no close
  tag" as "the model chose to answer without closing" (the deliberate #903
  recovery) and re-emits the accumulated thinking as content. But when
  `finish_reason == "length"` the close tag is missing **because we truncated
  the model**, not because it chose to answer — so the chain of thought is
  leaked into the visible reply.
- **Non-streaming**: `extract_thinking()` classifies a tag-free body as
  content (the `("", text)` fallback), so a truncated draft becomes the answer.

## Changes

1. `omlx/api/thinking.py`
   - Add a `truncated` flag to `ThinkingParser.finish()`; when set, skip the
     recovery branch so truncated thinking is never re-emitted as content.
   - Add `reclassify_truncated_thinking()` for the non-streaming paths.
2. `omlx/server.py`
   - Pass `truncated=last_output.finish_reason == "length"` at the three
     streaming call sites (OpenAI chat, Anthropic messages, Responses API).
   - Wire `reclassify_truncated_thinking()` into the three non-streaming
     paths (same three protocols).
3. `tests/test_thinking.py` — regression tests for streaming and
   non-streaming truncated paths; the #903 non-truncated recovery is preserved.

## Behavior

- Truncated mid-thought → the thinking is NOT surfaced as an answer; the
  answer body stays empty instead of leaking the chain of thought.
- Model stops on its own without closing the tag (#903) → unchanged, the body
  is still recovered as the answer.
- Normal separation with a close tag → unchanged.

## Test plan

- `pytest tests/test_thinking.py` — 50 passed (40 pre-existing + 10 new).

### DIFF
--- a/omlx/api/thinking.py
+++ b/omlx/api/thinking.py
@@ -212,6 +212,44 @@ def extract_thinking(text: str) -> Tuple[str, str]:
     return ("", text)
 
 
+def reclassify_truncated_thinking(
+    raw_text: str,
+    finish_reason: str | None,
+    start_in_thinking: bool,
+) -> Tuple[str, str]:
+    """Return (thinking_content, regular_content) for a non-streaming result.
+
+    When the generation was cut off by ``max_tokens`` (``finish_reason ==
+    "length"``) while the prompt had opened a thinking block and the model
+    never emitted a close tag before the cut, ``extract_thinking`` classifies
+    the tag-free body as content. That body is an unfinished chain of thought,
+    and surfacing it as the visible answer leaks the reasoning (issue #2457).
+
+    Reclassify the whole body as thinking so the answer stays empty instead
+    of leaked. The close tag string is the module-level ``_CLOSE_TAG``; we
+    check the raw text for it rather than relying on ``extract_thinking``'s
+    output, because both its tag-free and malformed branches collapse to
+    ``("", text)``.
+
+    Args:
+        raw_text: Complete (cleaned) model output text.
+        finish_reason: The generation's finish reason ("stop", "length", ...).
+        start_in_thinking: Whether the prompt opened a thinking block
+            (computed by the caller via ``prompt_opens_thinking``).
+
+    Returns:
+        Tuple of (thinking_content, regular_content).
+    """
+    thinking_content, regular_content = extract_thinking(raw_text)
+    if (
+        finish_reason == "length"
+        and _CLOSE_TAG not in raw_text
+        and start_in_thinking
+    ):
+        return raw_text, ""
+    return thinking_content, regular_content
+
+
 class ThinkingParser:
     """Stateful streaming parser for separating <think>...</think> from content.
 
@@ -323,15 +361,24 @@ class ThinkingParser:
             self._content_emitted = True
         return (thinking_delta, content_delta)
 
-    def finish(self) -> Tuple[str, str]:
+    def finish(self, truncated: bool = False) -> Tuple[str, str]:
         """Flush any remaining buffered content.
 
         Should be called when the stream is complete to emit any
         buffered characters that were waiting for potential tag completion.
         Also recovers from malformed thinking — when the model never
-        emitted ``</think>`` and no content was ever produced, returns
+        emitted `` response`` and no content was ever produced, returns
         the accumulated thinking text as content so the client surfaces
         a non-empty answer body.
+
+        Args:
+            truncated: True when the stream was cut off by ``max_tokens``
+                (``finish_reason == "length"``). In that case a missing
+                `` response`` close tag means the model was interrupted
+                mid-thought, not that it chose to answer without closing.
+                The accumulated text is reasoning, not an answer, so the
+                recovery branch is skipped to avoid leaking the chain of
+                thought into the visible reply (issue #2457).
 
         Returns:
             Tuple of (thinking_text, content_text) from remaining buffer
@@ -340,8 +387,25 @@ class ThinkingParser:
         partial = self._buffer
         self._buffer = ""
 
+        # Truncated stream: never re-emit accumulated thinking as content.
+        # The reasoning deltas already streamed live cannot be retracted,
+        # but surfacing the same text as an answer would make the client
+        # present a chain of thought as if it were the real reply. An
+        # empty content field is strictly better than a leaked one.
+        if truncated:
+            if not partial:
+                return ("", "")
+            # Partial tag never completed — emit it as-is in the current
+            # mode, mirroring the non-truncated flush below.
+            if self._in_thinking:
+                self._thinking_accumulated.append(partial)
+                return (partial, "")
+            else:
+                self._content_emitted = True
+                return ("", partial)
+
         # Recovery: prompt opened a thinking block (or model echoed
-        # ``<think>`` itself), the close tag never arrived, and nothing
+        # `` thinking`` itself), the close tag never arrived, and nothing
         # ever streamed as content. Re-emit the accumulated thinking text
         # as content so the answer body is not empty. The thinking events
         # already streamed live cannot be retracted, so the client sees
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -154,7 +154,13 @@ from .api.responses_utils import (
     format_sse_event,
     normalize_response_output_to_messages,
 )
-from .api.thinking import ThinkingParser, extract_thinking, prompt_opens_thinking
+from .api.thinking import (
+    ThinkingParser,
+    _CLOSE_TAG,
+    extract_thinking,
+    prompt_opens_thinking,
+    reclassify_truncated_thinking,
+)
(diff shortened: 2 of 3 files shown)


solodddd:
## End-to-end verification (real server)

Deployed the patch to the running oMLX server (0.5.8.dev3, source build) and re-tested the exact reproduction: single-turn, no tools, a long comparative request with `reasoning_effort=high`, truncated by `max_tokens` mid-thought.

**Before fix** (reported): `reasoning_content` and `content` were byte-for-byte identical — the chain of thought leaked into the visible answer.

**After fix** — non-streaming:

| max_tokens | finish_reason | reasoning_content | content |
|---|---|---|---|
| 200 | length | 368 chars | empty |
| 400 | length | 766 chars | empty |

**After fix** — streaming:

| max_tokens | finish_reason | reasoning_content | content |
|---|---|---|---|
| 200 | length | 384 chars | empty |
| 400 | length | 735 chars | empty |

Normal path still separates correctly (finish=stop, reasoning 274 / content 137 chars). The reasoning now stays reasoning and is no longer surfaced as the answer.

jundot:
Thanks for the detailed report, tests, and real-server verification. Before deciding on this, I revisited the history: #82 proposed treating truncated non-streaming output as reasoning-only but was never merged, while main later deliberately adopted the opposite recovery behavior in `86e88df` and reinforced it in `8c709030` (#1348).

That recovery was added because local models sometimes fail to emit a reliable `</think>` boundary or a separate final answer. Many OpenAI-compatible clients do not render `reasoning_content`, so oMLX intentionally accepted duplicated or incomplete text as the fallback rather than returning an apparently empty response.

This makes the `finish_reason="length"` exception feel more like a product-policy decision than a straightforward parser fix. In the narrow case where generation is truncated before any final answer exists, do you think an empty visible response is preferable to surfacing the text the model did produce as a fallback? Is `finish_reason="length"` insufficient for clients to recognize that the fallback is incomplete, or is there another user-visible risk that makes showing no `content` the safer choice?

I'd like to understand how you weigh those two failure modes before deciding whether truncation should be treated differently from the existing unclosed-thinking recovery.

solodddd:
Thanks — the history makes the reason for the current fallback much clearer. I agree that the existing behavior is useful when a model stops on its own without emitting a reliable ` response` boundary or a separate final answer. I would definitely keep that behavior unchanged.

I think `finish_reason="length"` is different because in that case the server has an unambiguous signal that generation was externally truncated. If we are still inside the thinking block at that point, we know the text is incomplete reasoning rather than a final answer.

I don't think an empty visible response is the ideal UX either. As you mentioned, my preferred long-term behavior would actually be recovery: preserve the generated reasoning/context and continue inference toward a direct final answer, rather than either exposing the unfinished reasoning or simply stopping.

But between the two behaviors in scope for this PR, I think an empty `content` is the safer failure mode. `finish_reason="length"` exists at the API level, but many clients don't make it obvious to the end user. If the unfinished reasoning is placed into `content`, the client presents it exactly like a normal assistant answer. In my reproduction, `reasoning_content` and `content` were byte-for-byte identical, so from the visible response alone the user cannot tell that the model never reached its answer.

So I see this PR as a narrow protocol-level distinction:

* model stops naturally without closing thinking → keep the existing fallback
* server truncates generation with `finish_reason="length"` while still thinking → don't reclassify that unfinished reasoning as the final answer

A continuation/recovery mechanism would be a separate RFE and, in my view, the better user-facing solution eventually. I wouldn't want to expand this PR into that unless you think it belongs here.

jundot:
Thanks, that distinction makes sense, and I agree that continuation would be a separate feature. For oMLX, though, `finish_reason="length"` explains why generation ended but does not change the local-model fallback goal of keeping generated text visible when no final answer exists.

Clients that do not surface `finish_reason` also often do not render `reasoning_content`, so returning empty `content` turns the same client limitation into an apparently blank or failed response. Automatic continuation would also exceed the caller's `max_tokens` contract unless it were designed as a separate opt-in feature.

Given that trade-off, I'm still inclined to keep the existing recovery semantics for both natural stops and length truncation. If you see a way to preserve a visible fallback without presenting it as a completed answer, I'd be interested in that direction.

solodddd:
Thanks — I looked into how SGLang handles the same case, and I found an approach that might be a useful middle ground.

With SGLang's Qwen3 reasoning parser, if generation hits `max_tokens` while still thinking, the default behavior is:

* `reasoning_content` keeps the partial reasoning
* `content` stays empty
* `finish_reason="length"`
* streaming does not replay the accumulated reasoning as a final content delta

I tested this directly against SGLang with both streaming and non-streaming requests.

What I found more interesting is that SGLang also has a `force_nonempty_content` option for clients/agents that require a visible `content` fallback. So instead of making the fallback unconditional, they keep strict reasoning/content separation by default and allow compatibility behavior to be explicitly opted into.

There is also an open SGLang PR (#33154) exploring a distinction between `stop` and `length`: recover to visible content when generation stops with unterminated reasoning, but keep reasoning separate when the server knows it was truncated by `max_tokens`.

That made me wonder if a similar split could work well for oMLX:

* natural `stop` + unterminated thinking → keep the existing recovery behavior
* `length` + unterminated thinking → keep it in `reasoning_content`, without presenting it as a completed answer
* clients that really require non-empty `content` could opt into the fallback behavior separately

This seems like it could preserve the compatibility goal behind the current recovery while avoiding the ambiguous case where a known-truncated reasoning trace is presented as the answer.

Would you be open to that direction? If so, I'm happy to adjust the PR around it.

PR #2616, opened 2026-08-12
feat(moe): extend expert offload to DeepSeek V4 / GLM native-kernel families (carries #2595)

tiagoflino:
## Summary

This is a **follow-up to #2595** by @beatakouchnir — the expert-offload mechanism that keeps a configurable fraction of MoE experts resident and streams the rest from the checkpoint's own safetensors. That PR is the foundation: the store, the slot cache, the chunking, the numerical contract. Everything here builds on it.

**This branch carries @beatakouchnir's #2595 commits** (the full mechanism, same content as PR #2595) plus the follow-up work below — opened here so the complete feature lands in one place. If #2595 merges first, we'll rebase this branch to drop the duplicates. Rebased onto the review-fixed #2595 branch (per-projection `qparams`, VLM dispatch, EnginePool admission), so this PR carries those fixes too.

What this PR adds is the **DeepSeek V4 / GLM-5.2 coverage** that #2595 explicitly scoped as a planned follow-up ("DeepSeek-V4 and GLM-5.2 use oMLX's native switch kernels — a planned follow-up on the same store"). The mechanism was already model-agnostic; those families were simply never reached by the type-based walk.

The change is deliberately small — 4 commits, ~150 lines net — because the hard parts (store, LRU slot cache, prefill chunking, bit-safe placement contract) are all #2595's design, which we validated independently before touching anything.

## Why it works without touching the mechanism

oMLX's native-kernel families (DeepSeek V4, GLM-5.2) carry their own `SwitchGLU`/`QuantizedSwitchLinear` classes under `omlx.patches.*` — same structure, same checkpoint naming (`model.layers.N.ffn.switch_mlp.*`, stacked `[E, ...]` + `weight/scales/biases`), different class identity. The base PR's walk matched on `type(obj) is SwitchGLU` (upstream mlx-lm class), so these layers were never yielded. Three small changes close the gap:

1. **Class detection** — the walk/resolver accept oMLX's native-kernel switch classes (guarded import; falls back to stock behavior if unavailable).
2. **`scores` kwarg** — `DeepseekV4MoE` calls `switch_mlp(x, inds, scores=...)`; the offloaded wrapper accepts (and ignores) it, same contract as the stock module.
3. **`model.` prefix** — the DS4 checkpoint stores tensors under `model.layers.N...` while the module tree path is `layers.N...` (no outer container). The store-view resolver retries with the prefix when the exact path misses.

Every layer is still verified against the checkpoint before wrapping; anything unmatched skips with a logged reason and runs resident — the failure mode stays "no offload", never "wrong outputs".

## Mixed-bit projections (review item from #2595)

While stacking on #2595 we found (via the review discussion) that `ExpertCache` took `group_size/bits/mode` from `gate_proj` and reused them for every projection — a GLU with mixed bit widths (e.g. 3-bit gate/up + an 8-bit down projection, common in oQ quants) crashed `gather_qmm` on a packed-shape mismatch. This PR fixes it in the follow-up: per-projection metadata (`meta[name]`), consumed by `qmm`/`qmm_native`; the native block path requires all projections to share one qualifying combo (mixed formats fall back to `gather_qmm`, same contract as the stock `all()` check). Regression test: synthetic DS4 with 3-bit gate/up + 8-bit down affine — wraps, per-projection metadata asserted, decode and sorted prefill **bit-exact** vs the resident twin.

## Validation

**Synthetic (CI-runnable, no checkpoint download)** — `tests/test_moe_expert_offload_dsv4.py` builds a tiny DeepseekV4Model (3 layers × 16 experts, affine 3-bit — same field layout as the real oQ quants), dumps it with real checkpoint naming, and verifies: every MoE layer wraps; decode and sorted-prefill outputs are **bit-exact** vs the resident twin; cache exercised. 20/20 with the base suite.

**Live, real checkpoint** — `mlx-community/DeepSeek-V4-Flash-0731-2.4bit-mixed` (92.8 GB) on a MacBook Pro **M5 Pro, 48 GB**:

- 43/43 MoE layers wrapped at 15% residency; **18.1 GB active vs 92.8 GB on disk (19%)**, peak 18.3 GB — the load-peak proof holds for a model larger than physical memory
- **4/4 prompts bit-identical** across two cache sizes (15% vs 40% residency, 34.8k vs 16.1k misses) — placement invariance at the strictest level
- Output quality verified visually (reasoning-format responses correct on physics/code/geography prompts)
- Hit rate scales as designed: 0.68 → 0.85 with residency

## Three optimizations from the 48 GB streaming findings

Measured on the same machine with the ds4 engine as reference (Flash 0731, q2 GGUF, SSD streaming):

- **`ExpertCache.shrink()` (page-cache cooperation)** — after a prefill burst, rebuild the resident buffers at the hot-subset size, returning wired memory to the OS so decode misses hit the page cache. Rationale: +3 GiB of wired expert cache measured **6.5 → 2.2 t/s** by starving the OS file cache. Opt-in via `OMLX_MOE_OFFLOAD_SHRINK_AFTER_PREFILL`.
- **`ExpertCache.pin_hot()` (learned hot set)** — routing heat tracked in `ensure()`; pin the top-n experts against LRU eviction (deterministic, tie-safe). The engine hook to call it per-request lands with the batch-union work.
- **Prev-step async prefetch (v1)** — each decode step records its routed expert set; the next step installs it host-side (no device readback), queuing the writes a step early. Measured hit rate **0.665 → 0.820 at 25% residency (+15.5 pp)**; decode is flat (+4%, noise) because single-request decode is bound by the per-layer sync/Python path rather than misses — the case for the engine-level batch-union follow-up. A/B knob `OMLX_MOE_OFFLOAD_DISABLE_PREFETCH`.

All three are placement-safe by construction (they change *when* weights are read, never *which* expert runs — the same contract as #2595).

## Honest performance numbers

Single-request decode on the real checkpoint currently runs ~0.9 t/s (TTFT ~25 s) vs ~4-9 t/s for the ds4 C engine on the same hardware. The gap is **not the offload mechanism**: profiling shows the cost is the expert-install path (host→device copies, ~13 GB on cold prefill) plus the per-layer Python path, and the oQ group-128 quant sat outside the custom-kernel envelope (the C++ binding hardcoded group 64). We extended the envelope in this PR — `supported_deepseek_affine` now admits (128, 2), and the Metal dequant loop is generic over group size — verified **bit-exact vs gather_qmm** with a warm-cache-controlled A/B: correct, and ~1.0x (no speedup on this workload; an earlier 14x figure was a cold/warm cache measurement confound, now documented in the commit). The async-prefetch and batch-union follow-ups @beatakouchnir quantified in #2595's docs are exactly the right next steps, now with real-model numbers attached.

## Acknowledgements

The substance of this feature is @beatakouchnir's #2595 — the store, the cache, the chunking, the numerical contract, and the documented prefetch path. Our contribution is the DS4/GLM coverage plus two measured optimizations, and we're grateful for the quick, sharp review feedback on the base PR (the DSpark verify window-union point is incorporated into the sizing notes). We also validated #2595's oQ4e coverage independently (30/30 layers, 4/4 bit-identical, 5.01 vs 14.65 GB) as part of this effort.

**Carries the #2595 commits.** Rebased onto main once #2595 merges.


jundot:
Thanks for carrying the DS4/GLM support forward. The focused Python suites pass locally, but I still cannot treat this branch as the mergeable fix for #2595.

The admission estimator is unchanged: it discounts any `.experts.<n>.` tensor and does not account for the runtime's minimum-eight capacity. I reproduced an unsupported `w1/w2/w3` checkpoint being estimated at 1,750 bytes from a 4,000-byte full size even though offload wrapped zero layers.

The prev-step prefetch also submitted zero reads in my repro. It counted already-resident experts as hits again, changing the reported hit rate from 0.50 to 0.67 without doing any prefetch work. In addition, `_install_many()` can evict protected experts when hot pins exhaust the victim set; I reproduced required experts ending with `map == -1`.

I suggest fixing the admission estimator in the base PR, then separating the DS4/GLM compatibility work from the prefetch and pinning optimizations. The latter need direct regressions proving that reads are actually queued and protected experts remain mapped.

PR #2617, opened 2026-08-12
fix(dspark): stop thinking-budget counter drift from MTP speculative calls

mvdbos:
**fix(dspark): stop thinking-budget counter drift from MTP speculative calls**

## Summary

**Why:** with DSpark MTP enabled, the thinking budget fires early and silently truncates reasoning. The budget processor is invoked on every speculative-token call the MTP machinery makes — draft generation *and* each verify row — but only the `m+1` tokens that are actually accepted/emitted should count. Before this fix, a budget of 8192 fired after ~7,676–7,788 *emitted* tokens (acceptance-dependent), cutting up to ~500 tokens of reasoning off the end. That undermines the entire point of a thinking budget: an exact cap on emitted reasoning.

**What:** wrap every speculative processor call in the DSpark batch generator with snapshot/restore, so speculative tokens shape the budget state but never *count* toward it. The budget fires at exactly the configured number of emitted tokens.

**No performance impact:** the snapshot/restore machinery is pure Python (11 scalar attrs, no GPU sync, ~3.5 µs/cycle) and the only added work — re-processing a few speculative tokens per cycle after restore — overlaps with GPU-bound decode. An A/B on the same model/request measured **36.8 tok/s with the fix vs 34.6 tok/s without**; the CPU-side worst case (+0.5–0.8 ms/cycle) is unmeasurable against ~28 ms/token decode.

Validated on Apple Silicon (M5 Max, 128 GB) with `DeepSeek-V4-Flash-0731-oQ2e-mtp` + the DSpark MTP drafter, thinking budget 8192.

## Repro

`tests/test_dspark_thinking_budget.py` drives the real `_run_verify_cycle_chain` + `_chain_next_drafts` with a fake model. RED numbers (pre-fix):

| Scenario | Processor calls/cycle | Emitted | Overcount |
|---|---|---|---|
| Full accept (k=3) | 7 | 4 | +3 |
| Partial accept (m=1) | 7 | 2 | +5 |
| Budget 12 fires after | 7 emitted | — | early by 5 |

Draft generation alone leaks +2 (drafts shape but must not count).

## How

`ThinkingBudgetProcessor` already exposes `snapshot_state()`/`restore_state()` (pure Python, 11 scalar attrs — no GPU sync). This PR uses them at every speculative-token site in `omlx/patches/mlx_lm_mtp/batch_generator.py`, mirroring the approach the VLM MTP path took in #2456:

- **Verify chain** (`_run_verify_cycle_chain`): snapshot per row, restore to `row_snaps[m]` after clamps/alignment — only the accepted `m+1` rows count
- **Draft generation** (`_dspark_next_drafts`, `_chain_next_drafts`, `_step_mtp`, post-init seed): snapshot before the draft loop, restore after — drafts shape the state but don't advance the counter

## Tests

- `tests/test_dspark_thinking_budget.py` — 5 tests: counter drift, exact-fire-at-budget, draft-gen doesn't count; **5/5 pass**
- Regression: `test_mlx_lm_mtp_patch`, `test_deepseek_v4_dspark`, `test_mtp_depth_controller`, `test_vlm_mtp_thinking_budget` — **239 passed, 11 skipped, 0 failures**

## Functional validation

- Live server, budget 8192, DSpark MTP: **fires at exactly 8,191 emitted reasoning tokens** (pre-fix sim: 7,676–7,788)
- Perf A/B (same model/request, 256 tokens, budget 100k so it never fires): with fix **36.8 tok/s** vs without **34.6 tok/s** — no measurable regression; the +0.5–0.8 ms/cycle CPU overhead (restore rewinds `_accepted_up_to`, ~5 re-processed tokens @ 152 µs each) hides under GPU-bound decode

### DIFF
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -1217,6 +1217,33 @@ def _apply_processors(processors, prev_tokens, logits_2d):
     return logits_2d
 
 
+def _snap_snapshotable(procs):
+    """Checkpoint state of processors exposing ``snapshot_state`` (budget).
+
+    Returns ``None`` when no processor supports position-keyed rewind, so
+    callers can skip the restore unconditionally (the DSpark path applies
+    processors to speculative draft positions; only the budget processor
+    tracks position-sensitive mutable state today).
+    """
+    if not procs:
+        return None
+    snaps = [p.snapshot_state() for p in procs if hasattr(p, "snapshot_state")]
+    return snaps or None
+
+
+def _restore_snapshotable(procs, snaps) -> None:
+    """Rewind processors previously checkpointed by :func:`_snap_snapshotable`."""
+    if not procs or not snaps:
+        return
+    it = iter(snaps)
+    for p in procs:
+        if hasattr(p, "restore_state"):
+            try:
+                p.restore_state(next(it))
+            except StopIteration:
+                return
+
+
 def _logprobs(logits_2d):
     import mlx.core as mx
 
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for tracking this down. I confirmed the new regression test fails on current main with the reported counter drift and passes with the per-row and draft-side snapshot/restore changes, including against the latest main. LGTM after the latest-main regression pass. I'll merge this.

PR #2623, opened 2026-08-13
fix(responses): start streaming thinking parser when the prompt opens <think>

### LINKED ISSUE #2584: DeepSeek-V4-Flash-0731: streaming /v1/responses emits reasoning as output_text, while non-streaming correctly returns a reasoning item
# DeepSeek-V4-Flash-0731: streaming `/v1/responses` emits reasoning as `output_text`, while non-streaming correctly returns a reasoning item

## Summary

When serving `Vontra/DeepSeek-V4-Flash-0731-MXFP4-MLX` with oMLX 0.5.7, the `/v1/responses` endpoint behaves differently depending on whether streaming is enabled.

With:

```json
"stream": false
```

reasoning is correctly separated from the final answer:

* `output[0].type = "reasoning"`
* `output[1].type = "message"`
* `usage.output_tokens_details.reasoning_tokens > 0`

However, with:

```json
"stream": true
```

the reasoning text is incorrectly emitted as normal:

```text
response.output_text.delta
```

events under a `message` item.

No reasoning item/events are emitted, and the final `response.completed` reports:

```json
"reasoning_tokens": 0
```

even though the beginning of the streamed `output_text` is clearly the model's reasoning.

This is reproducible by calling oMLX directly with `curl`. No proxy or client-side protocol translation is involved.

---

## Environment

* oMLX: `0.5.7`
* Model: `Vontra/DeepSeek-V4-Flash-0731-MXFP4-MLX`
* Endpoint: OpenAI-compatible `/v1/responses`
* Direct request to oMLX
* No CC Switch proxy involved
* No Grok Build/client involved in the reproduction below

---

## Reproduction

Set:

```bash
BASE="http://127.0.0.1:8000/v1"
MODEL="Vontra/DeepSeek-V4-Flash-0731-MXFP4-MLX"
API_KEY="YOUR_API_KEY"
```

### 1. Non-streaming Responses — works correctly

```bash
curl -s "$BASE/responses" \
  -H "Authorization: Bearer $API_KEY" \
  -d "{
    \"model\": \"$MODEL\",
    \"input\": \"请认真推理这个问题：一个商品先涨价20%，再降价20%，最终价格与原价相比变化多少？\",
    \"stream\": false,
    \"temperature\": 1.0,
    \"top_p\": 0.95
  }" | jq
```

The response correctly contains a dedicated reasoning item followed by the final message:

```json
{
  "output": [
    {
      "type": "reasoning",
      "status": "completed",
      "summary": [
        {
          "type": "summary_text",
          "text": "我们需要回答中文问题。需要认真推理。问题：商品先涨价20%，再降价20%..."
        }
      ]
    },
    {
      "type": "message",
      "status": "completed",
      "role": "assistant",
      "content": [
        {
          "type": "output_text",
          "text": "设原价为 100 元..."
        }
      ]
    }
  ],
  "usage": {
    "input_tokens": 27,
    "output_tokens": 292,
    "total_tokens": 319,
    "output_tokens_details": {
      "reasoning_tokens": 132
    }
  }
}
```

This looks correct.

---

### 2. Streaming Responses — reasoning is incorrectly emitted as output text

```bash
curl -sN "$BASE/responses" \
  -H "Authorization: Bearer $API_KEY" \
  -d "{
    \"model\": \"$MODEL\",
    \"input\": \"请认真推理这个问题：一个商品先涨价20%，再降价20%，最终价格与原价相比变化多少？\",
    \"stream\": true,
    \"temperature\": 1.0,
    \"top_p\": 0.95
  }" | tee /tmp/dsv4-responses.sse
```

For easier inspection:

```bash
grep -Ei \
'reasoning|output_item|output_text|completed' \
/tmp/dsv4-responses.sse
```

The stream begins by creating a **message** item instead of a reasoning item:

```text
event: response.output_item.added
data: {
  "type": "response.output_item.added",
  "output_index": 0,
  "item": {
    "type": "message",
    "status": "in_progress",
    "role": "assistant",
    "content": []
  },
  "sequence_number": 3
}
```

Then the model's reasoning is emitted through normal `output_text` events:

```text
event: response.output_text.delta
data: {
  "type": "response.output_text.delta",
  "delta": "我们需要回答中文问题。需要认真",
  "sequence_number": 5
}

event: response.output_text.delta
data: {
  "type": "response.output_text.delta",
  "delta": "推理。问题：商品先",
  "sequence_number": 6
}

event: response.output_text.delta
data: {
  "type": "response.output_text.delta",
  "delta": "涨价20%，再",
  "sequence_number": 7
}
```

The reasoning continues in `response.output_text.delta`, for example:

```text
"设原价 P，涨价20%后为 1.2P..."
```

After that, without switching to a new output item, the actual final answer also continues through the same `response.output_text.delta` stream:

```text
"设原价为 \(100\) 元（或其他单位）。..."
```

So the effective stream is:

```text
message item
  └─ output_text.delta
       ├─ reasoning text
       ├─ reasoning text
       ├─ reasoning text
       └─ final answer text
```

instead of something equivalent to:

```text
reasoning item
  └─ reasoning/summary deltas
reasoning item done

message item
  └─ output_text.delta
message item done
```

---

## Additional evidence

The final `response.completed` event also contains only a `message` output:

```json
{
  "output": [
    {
      "type": "message",
      "role": "assistant",
      "content": [
        {
          "type": "output_text",
          "text": "设原价为 100 元..."
        }
      ]
    }
  ]
}
```

and reports:

```json
{
  "output_tokens_details": {
    "reasoning_tokens": 0
  }
}
```

This differs from the equivalent non-streaming request, which correctly reports:

```json
{
  "output_tokens_details": {
    "reasoning_tokens": 132
  }
}
```

So this does not appear to be only a client rendering issue: the streaming Responses path itself appears to classify the reasoning tokens as visible output.

---

## Expected behavior

Streaming and non-streaming `/v1/responses` should expose the same semantic separation between reasoning and final output.

Since the non-streaming response produces:

```text
output[0] = reasoning
output[1] = message
```

the streaming response should likewise emit the reasoning through the appropriate reasoning item/events before emitting the final assistant message as `output_text`.

The completed streaming response should also account for the generated reasoning tokens in:

```json
usage.output_tokens_details.reasoning_tokens
```

---

### `stream=false`

```text
reasoning item      ✅
message item        ✅
reasoning separated ✅
reasoning_tokens=132
```

### `stream=true`

```text
reasoning item      ❌
message item        ✅
reasoning emitted as response.output_text.delta ❌
reasoning mixed with final answer                ❌
reasoning_tokens=0                               ❌
```

---

## Control observations

The same DeepSeek-V4-Flash-0731 model works correctly through `/v1/chat/completions`, where reasoning can be exposed separately as `reasoning_content`.

Also, `/v1/responses` with `stream=false` on the exact same model correctly identifies and separates the reasoning.

Because the non-streaming Responses path can already distinguish:

```text
reasoning
vs.
final answer
```

this seems unlikely to be caused by the model weights or chat template itself.

It appears specific to the **DeepSeek-V4-Flash-0731 streaming Responses path**, possibly where streamed model tokens are classified and converted into Responses API SSE events.

---
## Related issues / PRs checked

I checked the currently open streaming/reasoning-related work before filing this.

* #2458 appears to address **truncated thinking being recovered as visible content when the token limit is reached**. This reproduction is not truncated: reasoning is classified as `output_text` from the beginning of the stream.
* #2513 appears to address **DeepSeek V4 DSML tool-call separator handling and byte-stable re-rendering**. This reproduction does not involve tool calls.
* Commit `d5592aa` addresses **streaming suppression/parsing of JSON-array tool-call payloads** and does not appear to touch the Responses reasoning event path.

I could not find an existing issue or PR covering the specific behavior reproduced here: non-streaming `/v1/responses` correctly produces a reasoning item, while streaming `/v1/responses` classifies the same model's reasoning as normal `response.output_text.delta`.

---

## Why this matters

Responses API clients that rely on streaming reasoning events cannot distinguish thinking from visible assistant output.

For example, an agent client receives everything as normal assistant text:

```text
reasoning + final answer
```

instead of being able to render:

```text
Thinking
  reasoning...

Assistant
  final answer...
```

Non-streaming clients do not have this problem.

---

Please let me know if a full raw SSE capture, server logs, or additional tracing would help. I can provide them.

leepokai:
Fixes #2584

## Symptom

With DeepSeek-V4-Flash-0731 on streaming `/v1/responses`, all chain-of-thought text was emitted as `response.output_text.delta` under a `message` item: no `reasoning` output item was created and `usage.output_tokens_details.reasoning_tokens` was 0. The reporter's control tests showed the isolation of the bug: the same checkpoint separated reasoning correctly via non-streaming `/v1/responses` and via streaming `/v1/chat/completions`, and even within the broken stream, `response.output_text.done` contained only the final answer while the deltas carried the reasoning.

## Root cause

DeepSeek V4-style chat templates end the rendered prompt with `<think>`, so generation starts inside the thinking block and the model only ever emits the closing `</think>`. `stream_responses_api` initialized its parser as `ThinkingParser(start_in_thinking=native_reasoning)`, where `native_reasoning` is true only for Qwen 3.6+ `preserve_thinking` templates — so for DeepSeek V4 the parser started in content mode and every reasoning token streamed as `output_text`. The two sibling streaming paths (`stream_chat_completion` and the Anthropic messages stream) already handle this case by rendering the prompt with `_render_chat_prompt_for_thinking_detection` and passing `prompt_opens_thinking(...)` into `start_in_thinking`; the Responses streaming path was the only one missing the detection. Non-streaming `/v1/responses` was unaffected because it post-processes with `extract_thinking`, which handles the "no open tag, `</think>` present" case — the same reason `output_text.done` was correct while the deltas were wrong.

## Fix

`stream_responses_api` now applies the same detection pattern as the other two streaming paths: if `native_reasoning` is not already set, it renders the prompt and passes the `prompt_opens_thinking` result as `start_in_thinking`, inside the same try/except-log-at-debug guard. The detection is OR-ed with the existing flag, so Qwen 3.6+ `preserve_thinking` behavior is unchanged. Since the final reasoning output item and `reasoning_tokens` are both derived from the accumulated reasoning text, this one change fixes all three reported symptoms (missing reasoning item, reasoning leaked as output_text deltas, `reasoning_tokens: 0`).

Regression coverage:
- `tests/integration/test_e2e_streaming.py::TestStreamingHelperFunctions::test_stream_responses_api_prompt_opened_thinking_streams_as_reasoning` — drives `stream_responses_api` with a mock tokenizer whose template ends in `<think>` and asserts reasoning streams as `response.reasoning_summary_text.delta`, item ordering (reasoning at output_index 0, message at 1), `output_text.delta`/`.done` contain only the answer, and `reasoning_tokens > 0` in `response.completed`.
- `tests/test_chat_stream_thinking_static.py::test_stream_responses_api_starts_parser_when_prompt_opens_thinking` — AST guard mirroring the existing chat-completion guard.

## Verification

On Linux (Ubuntu 24.04, x86_64) with `mlx[cpu]==0.32.0`, in a fresh venv from `pip install -e .`:

- `python -m pytest tests/integration/test_e2e_streaming.py -q` → 36 passed, 12 deselected
- `python -m pytest tests/test_chat_stream_thinking_static.py tests/test_thinking.py tests/test_thinking_budget.py -q` → 83 passed
- Revert-proof: with the `omlx/server.py` hunk stashed, the new behavioral test fails with `AssertionError: assert '' == 'Need to inspect first.'` (no reasoning deltas emitted); with the fix restored it passes.

Note: this was verified with mock-based unit tests on Linux (mlx CPU wheels), not with the real DeepSeek-V4-Flash-0731 model on Apple Silicon. A confirmation run against the actual checkpoint on macOS would be a valuable follow-up.

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -6269,7 +6269,22 @@ async def stream_responses_api(
     accumulated_text = ""
     accumulated_reasoning = ""
     has_tools = bool(kwargs.get("tools"))
-    thinking_parser = ThinkingParser(start_in_thinking=native_reasoning)
+    # Some templates open the thinking block in the prompt itself, so the
+    # generated text starts with reasoning body and only later emits </think>.
+    start_in_thinking = native_reasoning
+    if not start_in_thinking:
+        try:
+            tokenizer = getattr(engine, "tokenizer", None)
+            if tokenizer is not None:
+                prompt, prompt_token_ids = _render_chat_prompt_for_thinking_detection(
+                    engine, messages, kwargs
+                )
+                start_in_thinking, _ = prompt_opens_thinking(
+                    tokenizer, prompt, prompt_token_ids=prompt_token_ids
+                )
+        except Exception as exc:
+            logger.debug("Could not detect Responses stream thinking state: %s", exc)
+    thinking_parser = ThinkingParser(start_in_thinking=start_in_thinking)
     seq = 0
 
     response_id = generate_id(IDPrefix.RESPONSE)
--- a/tests/integration/test_e2e_streaming.py
+++ b/tests/integration/test_e2e_streaming.py
@@ -3340,6 +3340,113 @@ class TestStreamingHelperFunctions:
         assert done_text == raw
         assert caplog.text.count("Unclosed tool-call envelope at end of stream") == 1
 
+    @pytest.mark.asyncio
+    async def test_stream_responses_api_prompt_opened_thinking_streams_as_reasoning(
+        self,
+    ):
+        """If the rendered prompt opens <think>, initial deltas are reasoning."""
+        from omlx.api.responses_models import ResponsesRequest
+        from omlx.server import stream_responses_api
+
+        class PromptOpenedThinkingTokenizer(MockTokenizer):
+            think_start = "<think>"
+            think_end = "</think>"
+            think_start_id = 999
+            think_end_id = 998
+            unk_token_id = -1
+
+            def convert_tokens_to_ids(self, token: str):
+                if token == self.think_start:
+                    return self.think_start_id
+                if token == self.think_end:
+                    return self.think_end_id
+                return self.unk_token_id
+
+            def encode(self, text: str, add_special_tokens: bool = False) -> list[int]:
+                ids = [101]
+                if text.rstrip().endswith(self.think_start):
+                    ids.append(self.think_start_id)
+                return ids
+
+            def apply_chat_template(
+                self, messages: list[dict], tokenize: bool = False, **kwargs
+            ) -> str:
+                return "user: Hi\nassistant:<think>"
+
+        engine = MockBaseEngine()
+        engine._tokenizer = PromptOpenedThinkingTokenizer()
+        engine.set_stream_outputs(
+            [
+                MockGenerationOutput(
+                    text="Need to inspect first.",
+                    new_text="Need to inspect first.",
+                    completion_tokens=1,
+                    finished=False,
+                ),
+                MockGenerationOutput(
+                    text="Need to inspect first.</think>Done.",
+                    new_text="</think>Done.",
+                    completion_tokens=2,
+                    finished=True,
+                    finish_reason="stop",
+                ),
+            ]
+        )
+
+        request = ResponsesRequest(model="test-model", input="Hi", stream=True)
+
+        events = [
+            event
+            async for event in stream_responses_api(
+                engine,
+                [{"role": "user", "content": "Hi"}],
+                request,
+                store_response=False,
+                max_tokens=256,
+            )
+        ]
+
+        parsed_events = []
+        for event in events:
+            for line in event.splitlines():
+                if line.startswith("data: "):
+                    parsed_events.append(json.loads(line[6:]))
+
+        reasoning_text = "".join(
+            event.get("delta", "")
+            for event in parsed_events
+            if event.get("type") == "response.reasoning_summary_text.delta"
+        )
+        content_text = "".join(
+            event.get("delta", "")
+            for event in parsed_events
+            if event.get("type") == "response.output_text.delta"
+        )
+        added_items = [
+            (event["output_index"], event["item"]["type"])
+            for event in parsed_events
+            if event.get("type") == "response.output_item.added"
+        ]
+        done_text = next(
+            event["text"]
+            for event in parsed_events
+            if event.get("type") == "response.output_text.done"
+        )
+        completed = next(
+            event["response"]
+            for event in parsed_events
+            if event.get("type") == "response.completed"
+        )
+
+        assert reasoning_text == "Need to inspect first."
+        assert content_text == "Done."
+        assert added_items[0] == (0, "reasoning")
+        assert (1, "message") in added_items
+        assert done_text == "Done."
+        assert completed["output"][0]["type"] == "reasoning"
+        assert completed["output"][1]["type"] == "message"
+        assert completed["usage"]["output_tokens_details"]["reasoning_tokens"] > 0
+
 
 class TestStreamingEdgeCases:
     """Tests for edge cases in streaming responses."""
--- a/tests/test_chat_stream_thinking_static.py
+++ b/tests/test_chat_stream_thinking_static.py
@@ -5,20 +5,20 @@ import ast
 from pathlib import Path
 
 
-def _stream_chat_completion_node():
+def _server_stream_node(name: str):
     source = (Path(__file__).resolve().parents[1] / "omlx" / "server.py").read_text()
     for node in ast.walk(ast.parse(source)):
         if (
             isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef))
-            and node.name == "stream_chat_completion"
+            and node.name == name
         ):
             return node
-    raise AssertionError("stream_chat_completion not found in server.py")
+    raise AssertionError(f"{name} not found in server.py")
 
 
 def test_stream_chat_completion_starts_parser_when_prompt_opens_thinking():
     """Chat streaming must not leak prompt-opened thinking as content."""
-    node = _stream_chat_completion_node()
+    node = _server_stream_node("stream_chat_completion")
     called = {
         call.func.id
         for call in ast.walk(node)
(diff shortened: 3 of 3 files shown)


wake-me:
Confirmed on real Apple Silicon hardware with the actual DeepSeek V4 Flash 0731 MLX checkpoint.

### Environment

* Hardware: Mac Studio M3 Ultra
* OS: macOS
* Model: `Vontra/DeepSeek-V4-Flash-0731-MXFP4-MLX`
* oMLX: PR #2623
* Commit: `99824a5`
* Version from source checkout: `0.5.8.dev3`
* Endpoint: `/v1/responses`
* Request mode: `stream=true`
* Thinking explicitly enabled with:

```json
"chat_template_kwargs": {
  "enable_thinking": true
}
```

The request was sent directly to the source-built oMLX server, without a proxy/client-side protocol translation.

### Result

The streaming Responses behavior is now correct.

The stream starts with a dedicated reasoning item:

```text
response.output_item.added
output_index: 0
item.type: reasoning
```

Reasoning is streamed through:

```text
response.reasoning_summary_text.delta
```

rather than being emitted as `response.output_text.delta`.

After reasoning completes, the final assistant message is created separately:

```text
response.output_item.added
output_index: 1
item.type: message
```

and only the final answer is streamed through:

```text
response.output_text.delta
```

The final `response.completed` also contains the expected structure:

```text
output[0].type = reasoning
output[1].type = message
```

and reasoning token accounting is now correct:

```json
"output_tokens_details": {
  "reasoning_tokens": 111
}
```

Previously, with the same real checkpoint on the affected oMLX version, the stream opened a `message` item first, emitted the reasoning as `response.output_text.delta`, produced no reasoning item, and reported `reasoning_tokens: 0`.

So this confirms that PR #2623 fixes all of the symptoms reported in #2584 on the actual DeepSeek-V4-Flash-0731 MLX checkpoint running on Apple Silicon/macOS.

Thanks for the fix.

jundot:
Thanks for the focused fix. I verified that the prompt-opened thinking detection now matches the existing Chat Completions and Anthropic streaming paths, and the regression test fails on main and passes with this patch. The real-model Apple Silicon confirmation also covers the reported event ordering and reasoning token accounting. LGTM, merging.

Issue #2625, opened 2026-08-13
Campaign — large-context memory fixes


mvdbos:
# Campaign — large-context memory fixes

> Coordination record for the large-context investigation. The active PR and merged
> replacement are listed first. Closed and superseded experiments are retained below for
> historical evidence.

## Background

This campaign came from a practical blocker. On a 128 GB Mac, DeepSeek v4 Flash was effectively unusable on oMLX with long context: long prompts were rejected at admission or aborted mid-prefill, so the usable context sat well below what the model and my hardware support. The same model at comparable context ran cleanly on antirez/ds4 on my hardware, isolating the problem to oMLX's memory accounting and scheduling rather than the model or the machine.

The campaign now has one open PR. The unaligned-tail allocation is fixed on `main` through #2674. The remaining open work is the narrower boundary-snapshot admission charge in #2628. Earlier estimator, scheduler-split, generic plateau, and monitor-registration experiments are closed or superseded and are documented near the end of this issue.

## Goal

Reliable inference at **large context** (hundreds of thousands of tokens) on oMLX. The work below investigates estimator correctness, chunk alignment, write-back pressure, and throttle accuracy. I used DeepSeek V4 Flash with a 350K prompt as a reasonable test because it worked with ds4 — an **example of "large", not an exact target**.

## Current status

- **#2628 — open, non-draft; CI green; awaiting maintainer review.** Admission now charges only a future boundary snapshot created after preflight. The charge requires a configured boundary store, a newly crossed block boundary, and a cache layout that actually needs snapshots. It uses serialized non-sliceable state rather than full `kv_exact`; existing pending buffers remain represented once through `current`/`phys_footprint`. At 350K with 64-token blocks, the DeepSeek V4 profile estimates **11.625 MiB** for the snapshot versus **2.264 GiB** for resident cache state. GitHub CI passes on Python 3.11, 3.12, and 3.13.
- **#2674 — merged.** The maintainer's native indexer fix handles unaligned DeepSeek V4 tails without global scheduler chunk splitting. On the same 347,929-token prompt, reported full-model MLX peak dropped from **142.255 GiB to 105.059 GiB**.

## Open work

| # | Work item | Current implementation and status |
|---|---|---|
[...]

PR #2626, opened 2026-08-13
fix(memory): estimate pooled-MLA KV bytes/token for DeepSeek-V4 configs (q_lora_rank, no kv_lora_rank)

mvdbos:

---

# fix(memory): estimate pooled-MLA KV bytes/token for DeepSeek-V4 configs (q_lora_rank, no kv_lora_rank)

## Why

`estimate_mla_kv_bytes_per_token` (omlx/memory_monitor.py) only recognizes MLA configs that
carry `kv_lora_rank`. DeepSeek-V4 names the latent `q_lora_rank`/`o_lora_rank` and stores the
resident cache ratio-compressed (`compress_ratios`, PoolingCache), so the function returns `None`
and the generic uniform-KV fallback is used. That fallback charges `num_kv_heads * head_dim`
expanded K/V that MLA never materializes — **84 KB/token vs 15.57 KB/token measured** (5.4× over).
Consequence: long-context preflight rejects prompts the model actually fits
(350K prompt → `400 "Prefill would require ~118.55 GB peak ... ceiling 114.00 GB"`).

## What

When `kv_lora_rank` is absent but a latent rank (`q_lora_rank` or `o_lora_rank`) and
`compress_ratios` are present, compute the pooled-MLA estimate from the **smallest** compression
ratio — conservative, never under-counts:

```python
latent_rank = _cfg_get(config, "q_lora_rank") or _cfg_get(config, "o_lora_rank")
compress_ratios = _cfg_get(config, "compress_ratios") or []
if _pos_int(latent_rank) and compress_ratios:
    min_ratio = min((int(r) for r in compress_ratios if isinstance(r, int) and r > 0), default=None)
    if min_ratio:
        return float(main_cache_layers * (latent_rank + rope_dim) * dtype_size / min_ratio)
return None
```

Non-MLA / unrecognized configs still return `None` (unchanged behavior).

## Hardware / model state

- Apple M5 Max, 128 GB unified memory; `--memory-guard-gb 120` (114 GB hard watermark)
- DeepSeek-V4-Flash-0731-oQ2e-mtp (46 layers: 24 ratio-4, 17 ratio-128, 5 local)

## Repro (before)

```
serve --port 8888 --memory-guard-gb 120   # default estimator
POST /v1/chat/completions, 347,929-token prompt
→ 400: "Prefill would require ~118.55 GB peak ... ceiling 114.00 GB"
```

## Measured truth

- Real KV: **15.57 KB/token** (SSD snapshot: 2.95 GB / 198,656 tokens)
- Patched estimate: **21.78 KB/token** (smallest-ratio conservative, 1.4× real, never under)

## Tests

1. Unit: V4-shaped config (`q_lora_rank` + `compress_ratios=[4,128]`, no `kv_lora_rank`) →
   returns pooled estimate at ratio 4; non-MLA config → `None` (unchanged).
2. E2E: 350K prompt admitted by preflight after fix (charge 21.78 KB/token).

## Caveats

- MLA-family only; uniform-KV models unaffected.

### DIFF
--- a/omlx/memory_monitor.py
+++ b/omlx/memory_monitor.py
@@ -1232,10 +1232,8 @@ def estimate_mla_kv_bytes_per_token(
     """
     kv_lora_rank = _cfg_get(config, "kv_lora_rank")
     rope_dim = _cfg_get(config, "qk_rope_head_dim")
-    if not (_pos_int(kv_lora_rank) and _pos_int(rope_dim)):
-        return None
-
-    if cache_list is None:
+    has_latent = _pos_int(kv_lora_rank) and _pos_int(rope_dim)
+    if not has_latent and cache_list is None:
         return None
 
     main_cache_layers = 0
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for investigating this. I rechecked current main with a 347,929-token DeepSeek V4 prompt, and the request was admitted and entered prefill with `--memory-guard-gb 120`. The DeepSeek V4 `PrefillMemoryProfile` added in #2521 already handles this cache layout and takes precedence over the override in this branch, so the reported admission failure is no longer reproducible. I am closing this as superseded. If the `estimate_block_memory` difference still needs correction for paged-SSD queue sizing, that should be handled in a separate focused PR with a runtime wiring test.

PR #2628, opened 2026-08-13
fix(memory): account future boundary snapshot bytes in prefill admission

mvdbos:
> Tracked in **#2625** as part of the large-context memory investigation.

---

# fix(memory): account future boundary snapshot bytes in prefill admission

## Why

The prefill admission estimate accounts for `current + KV + transient`. `current` already comes from `_current_usage_bytes()`, which samples `phys_footprint`, so raw SSD buffers that exist at admission time are already included. Adding pending-store counters separately would count those bytes twice.

The missing term is narrower: a raw boundary snapshot created **after** preflight can overlap the live KV and chunk-transient peak.

Observed on DeepSeek-V4-Flash-0731-oQ2e-mtp (M5 Max 128 GB, `--memory-guard-gb 120`, hard watermark 114.0 GB), with a 350K-context prompt and paged SSD cache enabled:

| run | peak | outcome |
|---|---:|---|
| same prompt, cache enabled | **116.5 GB** | hard abort (`process memory limit exceeded`) |
| same prompt, `--no-cache` | 113.3 GB | soft pressure only; completed |

The admission guard should price the future snapshot responsible for that overlap without recharging buffers already represented in `current` or imposing a snapshot-sized charge on ordinary sliceable `KVCache` layouts.

## What changed

`Scheduler._admission_estimate()` adds a narrowly scoped `writeback` term:

- zero unless a boundary snapshot store and block-aware cache are configured;
- zero unless the request crosses a new paged-cache boundary;
- zero unless the model's actual cache layout requires boundary snapshots (`_detect_boundary_snapshot_need()`);
- otherwise, the largest serialized size of the non-sliceable state at any boundary the suffix will cross.

**The admission charge is the SINGLE LARGEST future snapshot, not the last one.** A compacted `PoolingCache` snapshot size oscillates with `token_count % ratio`, so the final boundary crossed is not necessarily the largest. `_expected_boundary_snapshot_bytes()` now takes the max over every reachable boundary in `(cached_tokens, prefilled_tokens]`.

**The mid-prefill guard prices the snapshot it is about to emit.** `_guard_prefill_chunk()` adds the net-new boundary-snapshot cost to its fit/breach/shrink math (`_boundary_snapshot_bytes_at(kv_len + n_tokens)`), so the chunk is reserved *before* it runs rather than discovering the just-emitted snapshot after sampling `phys_footprint`.

The estimate does **not** read `BoundarySnapshotSSDStore.pending_bytes` or `PagedSSDCacheManager.pending_write_bytes()`. Existing pending buffers remain represented only once through `current`/`phys_footprint`.

`MemoryMonitor.estimate_boundary_snapshot_bytes()` models the data serialized by the snapshot path rather than using the full resident `kv_exact`:

- full, block-sliceable `KVCache` layers are excluded;
- rotating caches contribute their post-chunk window plus measured fixed recurrent state;
- model-specific profiles can price other non-sliceable layouts;
- the DeepSeek V4 profile mirrors boundary compaction: local K state plus one-block `PoolingCache` deltas, remainder buffers, and overlap carry state;
- the current Qwen4 profile charges only its measured fixed GDN state; block-sliceable QSA state remains excluded.

Both rejection paths and both eviction paths use the same `kv_exact + transient + writeback` total. Rejection diagnostics continue to show `SSD write-back` separately.

## DeepSeek V4 sizing check

For the exact DeepSeek V4 profile at a 350,000-token prompt with 64-token blocks, the suffix crosses boundaries up to 349,952 tokens:

- serialized non-sliceable snapshot at the FINAL boundary (349,952): **6,516,736 bytes (6.215 MiB)**;
- maximum over all reachable boundaries: **9,117,696 bytes (8.695 MiB)** (it occurs at a lower interior boundary, since size oscillates with `token_count % ratio`);
- resident cache estimate: **2,430,959,616 bytes (2.264 GiB)**;
- admission charge (max snapshot): **0.376%** of resident cache.

These are deterministic estimator outputs, not a new live peak measurement.

## Tests

Added/adjusted regression coverage for:

- an existing pending boundary/paged write buffer not being recharged;
- a pure `KVCache` layout receiving zero snapshot charge;
- a non-sliceable layout receiving the max future snapshot charge;
- admission pricing the **max** boundary, not the last (oscillating estimator: 10096 ≠ last 5096);
- the mid-prefill guard pricing the upcoming boundary snapshot;
- both preflight eviction paths including the future snapshot term;
- generic rotating/fixed-state snapshot sizing;
- DeepSeek V4 compacted `PoolingCache` snapshot sizing;
- Qwen4 profile compatibility after the upstream memory-accounting refactor.

Verification on the branch rebased onto `jundot/omlx@1a7ab338`:

```text
681 passed, 4 skipped, 3 deselected in 17.09s
```

Command:

```bash
uv run pytest \
  tests/test_scheduler.py \
  tests/test_scheduler_admission.py \
  tests/test_scheduler_chunked_prefill.py \
  tests/test_scheduler_logits_processors.py \
  tests/test_scheduler_prompt_boundary_store.py \
  tests/test_scheduler_prefill_memory_guard.py \
  tests/test_memory_monitor.py \
  tests/test_boundary_snapshot_store.py \
  tests/test_pooling_cache_delta.py \
  tests/test_deepseek_v4_patch.py \
  tests/test_prefill_oom_graceful.py -q
```

`git diff --check` passes.

## Files

- `omlx/memory_monitor.py`: serialized non-sliceable snapshot estimator and DeepSeek V4 compacted-state profile.
- `omlx/scheduler.py`: layout- and boundary-gated future snapshot charge (max over reachable boundaries), mid-prefill guard reservation, shared diagnostics, and eviction accounting.
- `tests/test_memory_monitor.py`: generic and DeepSeek V4 sizing controls.
- `tests/test_scheduler_prefill_memory_guard.py`: admission, pending-buffer, layout, max-boundary, guard pricing, and eviction controls.
- `tests/test_prefill_oom_graceful.py`: guard-path harness support for the writeback probe.

### DIFF
--- a/omlx/memory_monitor.py
+++ b/omlx/memory_monitor.py
@@ -163,6 +163,10 @@ class PrefillMemoryProfile(Protocol):
         self, num_tokens: int, *, chunk_tokens: int = 1
     ) -> int: ...
 
+    def estimate_boundary_snapshot_bytes(
+        self, num_tokens: int, *, block_size: int
+    ) -> int: ...
+
     def estimate_prefill_transient_bytes(
         self, query_tokens: int, kv_len: int
     ) -> int: ...
@@ -707,6 +711,35 @@ class MemoryMonitor:
 
         return total + self._fixed_state_bytes
 
+    def estimate_boundary_snapshot_bytes(
+        self, num_tokens: int, *, block_size: int
+    ) -> int:
+        """Serialized non-sliceable state in one boundary snapshot.
+
+        Full ``KVCache`` layers are block-sliceable and therefore excluded.
+        Rotating caches serialize their post-chunk window (not the temporary
+        ``window + chunk - 1`` resident peak); fixed recurrent state is copied
+        once. Model-specific profiles can refine this for other cache shapes.
+        """
+        if num_tokens <= 0 or block_size <= 0:
+            return 0
+        total = int(self._fixed_state_bytes)
+        if self._prefill_memory_profile is not None:
+            return total + int(
+                self._prefill_memory_profile.estimate_boundary_snapshot_bytes(
+                    num_tokens,
+                    block_size=block_size,
+                )
+            )
+        if self._rotating_layer_specs:
+            kv_heads = self._num_kv_heads or 0
+            dim = self._head_dim or 0
+            if kv_heads and dim:
+                per_token = kv_heads * dim * self._score_dtype_size * 2
+                for count, window in self._rotating_layer_specs:
+                    total += count * min(int(num_tokens), window) * per_token
+        return int(total)
+
     def _uses_fused_sdpa(self, query_tokens: int, kv_len: int) -> bool:
         hd = self._head_dim or 0
         n_q = self._num_attention_heads or 0
@@ -1046,6 +1079,26 @@ class _DeepSeekV4PrefillMemoryProfile:
         carry = 2 * ratio * projection_dim if overlap and num_tokens >= ratio else 0
         return pooled + buffers + carry
 
+    def _pool_snapshot_elements(
+        self,
+        num_tokens: int,
+        *,
+        block_size: int,
+        ratio: int,
+        pooled_dim: int,
+        overlap: bool,
+    ) -> int:
+        """Serialized PoolingCache state after boundary delta compaction."""
+        pooled_end = num_tokens // ratio
+        pooled_start = max(0, num_tokens - block_size) // ratio
+        pooled = max(0, pooled_end - pooled_start) * pooled_dim
+        projection_dim = pooled_dim * (2 if overlap else 1)
+        # PoolingCache.state exposes only the populated remainder slices;
+        # zero-remainder boundaries serialize no buffer payload.
+        buffers = 2 * (num_tokens % ratio) * projection_dim
+        carry = 2 * ratio * projection_dim if overlap and num_tokens >= ratio else 0
+        return pooled + buffers + carry
+
     def estimate_resident_kv_bytes(
         self, num_tokens: int, *, chunk_tokens: int = 1
     ) -> int:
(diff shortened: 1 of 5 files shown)


jundot:
Thanks for chasing this. I rechecked the accounting, and the current issue appears narrower than the PR description: `_current_usage_bytes()` already samples `phys_footprint`, so raw buffers pending at admission time are already included in `current`. The missing term is only write-back memory created by the admitted request after preflight, especially a boundary snapshot that overlaps the live KV/transient peak.

Could you revise the estimate to avoid adding `BoundarySnapshotSSDStore.pending_bytes` and `PagedSSDCacheManager.pending_write_bytes()`, gate the future charge on the actual cache layout requiring boundary snapshots, and size it from the serialized non-sliceable state rather than the full `kv_exact`? Please also add tests for an existing pending buffer not being recharged, a pure `KVCache` model receiving zero snapshot charge, and a non-sliceable model receiving the future snapshot charge. That would address the reported DeepSeek peak without over-rejecting ordinary SSD-cache requests.

mvdbos:
Thanks. Agreed. I revised the branch to make the accounting as narrow as you described:

- removed all admission-time pending-store reads and dropped the new paged-cache query APIs, so buffers already represented in `phys_footprint`/`current` are not charged again;
- gate the future charge on a configured boundary store, a newly crossed block boundary, and `_detect_boundary_snapshot_need()` for the actual cache layout;
- estimate the serialized non-sliceable state rather than `kv_exact`;
- for DeepSeek V4, mirror boundary compaction: local K state plus one-block `PoolingCache` deltas, remainder buffers, and overlap carry state;
- added the three requested controls, plus eviction-accounting and compacted-sizing tests.

For the exact DeepSeek V4 profile at 350,000 prompt tokens with 64-token blocks, the future snapshot term is 12,189,696 bytes (11.625 MiB), versus 2,430,959,616 bytes (2.264 GiB) for resident cache state. That avoids the previous full-KV overcharge.

The branch is rebased onto `jundot/omlx@b2fdc715`. Relevant verification: 574 passed, 4 skipped, 3 deselected. Targeted fatal/undefined-name lint checks and `git diff --check` also pass.

mvdbos:
@jundot ok to merge now?

mvdbos:
Updated.

Key changes:
- Admission now prices the **max** boundary snapshot rather than the last (compacted `PoolingCache` size oscillates with `token_count % ratio`, so the final boundary isn't the largest). DeepSeek sizing check: max = **9,117,696 B / 8.695 MiB** vs final-boundary 6,516,736 B.
- The mid-prefill `_guard_prefill_chunk()` now reserves the snapshot the upcoming chunk will emit at the boundary it lands on.

Verified: 681 passed, 4 skipped, 3 deselected.

mvdbos:
@jundot

PR #2629, opened 2026-08-13
fix(scheduler): re-expand prefill chunks at the MLX pool plateau (net-rate sizing)

mvdbos:

---

# fix(scheduler): detect the MLX pool plateau and re-expand prefill chunks

## Why

The adaptive prefill throttle sizes chunks from `_predicted_chunk_transient`, a MAX of the
last measured delta, the long-run EWMA, and the static SDPA+KV estimate. All three are
**absolute-growth** predictors. `PrefillTransientTracker` deliberately skips non-positive
post-chunk deltas (they would bias the EWMA toward zero and under-estimate the next chunk),
so once the MLX cache pool starts absorbing chunk workspaces, those predictors stay frozen
at the growth-phase rate.

At long context the pool does exactly that: each chunk's workspace reuses the pooled
buffers from the previous chunk, net phys_footprint growth goes ~0 (or negative), and the
throttle keeps shrinking chunks for the rest of the prefill even though usage is flat.

Observed on DeepSeek-V4-Flash-0731-oQ2e-mtp, 350K-context prefill (`--memory-guard-gb 120`):

```
12:xx:xx Prefill throttled: chunk 2048 -> 1024 (usage 95.74GB vs sizing target 108.00GB,
         per_token=6619.4KB ...)
```

with the usage trace flat at 93-96 GB for the entire throttled tail — per_token
6,619.4 KB/token (the frozen growth-phase rate) while the real steady-state transient was
~0. Cost: chunks halved for the final ~6 minutes of the prefill (~10% of wall time),
needlessly.

## What

`PrefillTransientTracker` now tracks the plateau:

- `_plateau_streak`: consecutive non-positive chunk deltas (non-positive samples still skip
  the absolute-growth EWMA and the sample count — existing semantics unchanged);
- `_net_ewma_per_token`: an EWMA over every chunk's **net** per-token delta
  (non-positives as 0) — the true steady-state net growth rate.

`Scheduler` sizes chunks with the net rate once `plateau_detected` (default 5 consecutive
non-positive deltas):

- `_adaptive_chunk_size` (the throttle) — a chunk the stale absolute predictors would
  shrink runs unchanged;
- `_guard_prefill_chunk`'s secondary clamp — mirrors the throttle so a re-expanded chunk is
  not immediately shrunk back.

The **abort gate stays conservative**: `_admission_transient_bound` (pre-chunk guard +
admission) still prices the absolute bound with its observed-max floor, and the enforcer's
pressure abort remains the backstop. A positive sample resets the streak immediately.

## Repro

1. 350K prefill, throttle log shows `chunk 2048 -> 1024 ... per_token=6619.4KB` while the
   usage trace is flat (93-96 GB) — chunks halved for the final ~6 min for no measurable
   net growth.
2. After this change the plateau is detected after 5 flat chunks and sizing switches to the
   net rate; the throttle log line above no longer appears at the plateau (the chunks
   re-expand).

## Tests

- `tests/test_prefill_transient_tracker.py::TestPoolPlateau` — detection threshold,
  positive-delta reset, net-EWMA decay, preserved negative-delta semantics, reset.
- `tests/test_scheduler_prefill_memory_guard.py` — `_adaptive_chunk_size` and
  `_guard_prefill_chunk` re-expand at the plateau and still shrink without it (paired
  controls).

## Ordering

Re-expansion is only safe with the final-chunk spike bound present (see the planned
prefill-throttle PR in the parent campaign issue
[#2625](https://github.com/jundot/omlx/issues/2625) — roadmap PR 4): without the
64-aligned final-chunk split, a re-expanded 2048-token final chunk re-introduces the
one-shot fallback allocation (~45 GB at 350K). Land the alignment fix first.

## Files

- `omlx/prefill_transient_tracker.py` (+63): plateau streak + net-rate EWMA + properties.
- `omlx/scheduler.py` (+32): plateau-aware sizing in the throttle and the guard's clamp.
- tests: +143 lines.

### DIFF
--- a/omlx/prefill_transient_tracker.py
+++ b/omlx/prefill_transient_tracker.py
@@ -26,6 +26,11 @@ class PrefillTransientTracker:
     """
 
     _EWMA_ALPHA = 0.3  # weight on the most recent chunk
+    # Consecutive non-positive chunk deltas that define a pool plateau.
+    # At the plateau the MLX cache pool absorbs each chunk's workspace, so
+    # the NET footprint does not grow and the absolute-growth predictors
+    # (last-delta / EWMA / static) over-predict by the growth-phase rate.
+    _PLATEAU_CHUNKS = 5
     # Candidates above this are rejected from the running max: a one-off
     # Metal/pool spike this large is not a repeatable chunk transient, and
     # charging it at admission would refuse prompts that always fit. A
@@ -61,6 +66,14 @@ class PrefillTransientTracker:
         # need to allocate that pool again on the next chunk, so the scheduler
         # prices it once until a positive measurement confirms reallocation.
         self._recent_reclaim_bytes: int = 0
+        # Consecutive non-positive chunk deltas; >= _PLATEAU_CHUNKS means the
+        # MLX pool is absorbing chunk workspaces (net footprint ~0).
+        self._plateau_streak: int = 0
+        # EWMA over EVERY chunk's net per-token delta (non-positives as 0).
+        # Measures steady-state net growth — ~0 at the pool plateau, where the
+        # absolute-growth EWMA stays stuck at the growth-phase rate.
+        self._net_ewma_per_token: float = 0.0
+        self._net_ewma_samples: int = 0
 
     def record_reclaim(self, reclaimed_bytes: int) -> None:
         """Accumulate footprint released since the last positive sample."""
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for investigating the pool plateau behavior. The observed behavior may be real, but this change applies plateau-based chunk re-expansion to the shared scheduler for every model, while the current validation is limited to one DeepSeek V4 workload.

I’m not comfortable using five non-positive footprint deltas as a general signal that a larger next chunk is safe. In particular, reclaim tracking does not invalidate the plateau streak or net-rate state before the next sizing decision, so stale plateau state can underprice a chunk after the pool has changed.

I’m closing this PR rather than merging the heuristic in its current form. I’ll address the DeepSeek V4 tail allocation separately in #2674, and revisit generic plateau sizing if it can be tied to explicit pool lifecycle and headroom signals and validated across multiple model paths.

PR #2630, opened 2026-08-13
fix(cache): bound SSD persistence teardown

mvdbos:
# fix(cache): bound SSD persistence teardown

## Why

A memory-pressure abort can evict a model while its SSD writer queue is saturated. Before this change, `scheduler.shutdown()` could spend roughly one second per dirty entry waiting for queue space, so cache persistence could outlive the engine's 60-second teardown budget and trigger `fatal_exit`, taking every loaded model down with the evicted one.

Observed on DeepSeek-V4-Flash-0731-oQ2e-mtp with a 350K-token prefill and paged SSD cache enabled:

```text
Engine teardown timed out after 60s while running shutdown for engine <uuid>;
exiting process so the supervisor can restart with a clean state
```

## What changed

The revision bounds SSD persistence teardown and makes it restart-safe:

- `EngineCore.close()` creates one monotonic absolute persistence deadline, passes it to `Scheduler.shutdown()`, and reuses it for fallback cache-manager close. Scheduler shutdown and `deep_reset()` stay on the engine's `_mlx_executor`; a rejected or failing worker dispatch is **not** retried on the caller thread.
- `Scheduler.shutdown()` signals SSD shutdown before waiting for active `store_cache` futures. Workers stop before preparing another block. The same absolute deadline also bounds the store worker's MLX stream cleanup.
- `PagedSSDCacheManager` carries the same deadline through hot-cache flushing, pre-enqueue preparation, queue admission/full handling, and writer join. Every stage uses only the remaining time, and the deadline path never falls back to an unbounded inline write. The SpecPrefill draft SSD manager is signaled before worker quiescence and closed with that same deadline; deferred writer cleanup cannot be reopened by `deep_reset()`.
- **The finite persistence deadline is sticky.** `PagedSSDCacheManager.close()` reuses an already-installed finite `_persistence_deadline` when called with `deadline=None`, so a later retry or `deep_reset()` cannot downgrade a bounded teardown back into an unbounded flush. `Scheduler.deep_reset()` now accepts and forwards the deadline to the draft SSD cache close.
- **A surviving writer is restart-safe, not silently tolerated.** `EngineCore.close()` fatal-exits if the SSD writer thread outlives the deadline rather than dropping the manager reference and reporting a clean teardown — a replacement manager would otherwise open the same cache directory while the old writer drains. `_write_block_file` now uses a unique temp filename and, on a write/rename failure, unlinks only the temp path it owns, never a final path another manager may have committed.
- **MLX executor rejection is fatal.** If `_mlx_executor.submit()` rejects teardown dispatch, `EngineCore.close()` takes the fatal/restart-safe path instead of logging and continuing to report the engine closed over partial teardown.
- **Forced ownership is executor-aware.** A forced model-ownership transfer runs the previous owner's `deep_reset()` on that owner's own MLX executor (bounded wait) and aborts the transfer if it cannot be dispatched or complete — instead of resetting on the acquiring thread while holding the registry lock.
- Calls without a deadline retain complete-flush semantics, including writes admitted before shutdown in hot-cache and direct-write modes.
- The upstream MLX 0.32.2 reclaim and executor-exit behavior remains unchanged.

## Reproduction and live evidence

The original failure was reproduced with a 350K-token prefill, paged SSD cache enabled, and a memory abort while the writer queue was mid-flush. On the revised path, the scheduler's persistence shutdown stayed within the shared budget, MLX reclaim completed, `/v1/models` remained responsive, and no process-exit path fired. The evicted model was removed while other loaded models continued serving.

## Tests

Regression coverage includes:

- saturated writers with active store workers;
- shutdown signaling and cleanup ordering;
- scheduler/MLX executor thread ownership;
- one-deadline wall-clock bounds across preparation, enqueue, flush, and join;
- rollback when the deadline expires during queue admission or eviction preparation;
- complete-flush behavior when no deadline is supplied;
- compatibility with hot-cache write-through and current prefix-cache behavior;
- unique temp files across managers writing the same hash, and a rename failure not unlinking another manager's committed final file;
- sticky finite deadline through `close(deadline=None)`;
- `EngineCore.close()` fatal-exiting when the SSD writer survives and when the MLX executor rejects submission;
- forced-ownership transfer routing `deep_reset()` via the owner's MLX executor and aborting on failure.

## Verification

Rebased onto `jundot/omlx@1a7ab338` (including the MLX 0.32.2 teardown changes).

```text
713 passed in 37.58s
```

```bash
uv run pytest \
  tests/test_paged_ssd_cache.py \
  tests/test_prefix_cache.py \
  tests/test_scheduler.py \
  tests/test_engine_core.py \
  tests/test_hot_cache.py \
  tests/test_model_registry.py -q
```

`git diff --check` passes.

## Files

- `omlx/cache/paged_ssd_cache.py`: deadline-aware enqueue, sticky deadline, hot-cache flush, queue-full handling, writer join, unique temp filenames, and final-file ownership.
- `omlx/cache/prefix_cache.py`: stop-aware store workers.
- `omlx/engine_core.py`: shared persistence deadline, fatal-on-surviving-writer, fatal-on-submit-rejection, and executor-dispatched deep_reset.
- `omlx/model_registry.py`: executor-routed forced-ownership reset with abort-on-failure.
- `omlx/scheduler.py`: shutdown ordering, worker quiescence, deadline forwarding, and deep_reset deadline propagation.
- tests: concurrency, ordering, deadline, no-deadline, sticky-deadline, unique-temp, fatal-exit, and forced-ownership regressions.

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import contextlib
 import errno
 import hashlib
+import itertools
 import json
 import logging
 import os
@@ -68,6 +69,7 @@ _PENDING_WRITES_HARD_RAM_FRACTION = 0.30
 _PENDING_WRITES_SOFT_FLOOR = 32
 _PENDING_WRITES_CEILING = 256
 _PENDING_WRITE_PUT_TIMEOUT_SECONDS = 1.0
+_PENDING_WRITE_PUT_POLL_SECONDS = 0.05
 
 # Conservative defaults for the per-block cost estimator. The actual
 # bytes-per-block depends on the model (KV-cache layers × num_kv_heads ×
@@ -160,6 +162,11 @@ _MAX_PENDING_WRITES = _compute_max_pending_writes()
 # subsequent saves drain the remainder; bounds per-call latency at the
 # cost of taking multiple saves to fully reconverge.
 _MAX_INLINE_UNLINKS_PER_SAVE = 32
+
+# Monotonic counter for unique `_write_block_file` temp filenames. Combined with
+# the PID, it guarantees two managers writing the same block hash never collide
+# on a deterministic `<stem>_tmp.safetensors` name.
+_TMP_COUNTER = itertools.count()
 
 
 # Cache format version. Bump when on-disk layout or RotatingKVCache meta_state
@@ -1775,6 +1782,15 @@ class PagedSSDCacheManager(CacheManager):
         # Eviction path: _hot_cache_put (holds _hot_cache_lock, releases), then
         # _enqueue_ssd_write (holds _pending_write_hashes_lock).
         self._pending_write_buffers: dict[bytes, dict] = {}
+        # Separate write admission from writer-thread termination. Scheduler
+        # shutdown closes admission before waiting for active store workers;
+        # the writer remains alive until close() flushes already accepted data.
+        self._persistence_shutdown = threading.Event()
+        self._persistence_shutdown_lock = threading.Lock()
+        self._persistence_deadline: float | None = None
+        self._close_cleanup_lock = threading.Lock()
+        self._close_cleanup_requested = False
+        self._close_cleanup_done = False
         self._writer_shutdown = threading.Event()
         # Writer thread is only needed when writing to SSD.
         self._writer_thread = None
@@ -1808,6 +1824,28 @@ class PagedSSDCacheManager(CacheManager):
         )
 
     # --- Hot cache helpers ---
+
+    @property
+    def accepting_writes(self) -> bool:
+        """Whether cache producers may submit new persistence work."""
+        return not self._persistence_shutdown.is_set()
+
+    def begin_shutdown(self, *, deadline: float | None = None) -> None:
+        """Stop accepting new writes while keeping the writer available to drain."""
+        # Publish the stop signal before waiting for an enqueue attempt that
+        # may currently hold the admission lock during its short queue poll.
+        # That producer then observes the event and rolls back on its next pass
+        # instead of repeatedly reacquiring the lock and starving shutdown.
+        # Publish a finite deadline before the stop signal so producers woken by
+        # the event cannot misclassify bounded teardown as a no-deadline flush.
+        if deadline is not None and (
+            self._persistence_deadline is None or deadline < self._persistence_deadline
+        ):
+            self._persistence_deadline = float(deadline)
+        self._persistence_shutdown.set()
+        # Wait for any current short queue-admission poll to release ownership.
+        with self._persistence_shutdown_lock:
+            pass
 
     @staticmethod
     def _hot_cache_entry_size(entry: dict) -> int:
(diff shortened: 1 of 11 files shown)


jundot:
Thanks for tackling this. The reported SSD writer teardown failure is worth fixing, but the current change expands the scope from bounded cache persistence into broadly non-fatal engine teardown. I don’t think that is safe to merge.

Please narrow the change to the cache-persistence problem:

- Keep `Scheduler.shutdown()` cleanup, pending-remove draining, `deep_reset()`, and all BatchGenerator/MLX work on the engine’s `_mlx_executor`. Do not run the entire scheduler shutdown on the close or event-loop thread.
- Make only SSD persistence bounded and non-fatal. Use one monotonic absolute deadline across hot-cache flush, queue enqueue, queue-full handling, and writer join. Each stage must use only the remaining time, and the deadline path must not fall back to an unbounded inline write. Calls without a deadline must retain the current full-flush behavior.
- Signal the SSD persistence shutdown before waiting for store-cache futures so workers stop accepting or queueing new writes and can finish promptly.
- Do not clear `_inflight_store_futures` or tear down BatchGenerator, cache, or snapshot state while a worker is still running. If a worker cannot safely stop within the budget, preserve the existing restart safety behavior instead of continuing with a partially torn-down engine.
- Keep the existing safety behavior for `deep_reset()` and MLX reclaim timeouts unless the worker and all owned state can be proven safely isolated.

Please add regression tests covering a saturated writer with an active store worker, cleanup ordering, MLX executor thread ownership, and a wall-clock assertion that the SSD close path respects one deadline. The slow-shutdown test also needs `import time`; currently its `NameError` is swallowed, so it passes without exercising the delay. Please rebase onto the latest `main` and rerun the relevant test files after resolving the conflict.

mvdbos:
Thanks — I narrowed the revision to the SSD persistence teardown and rebased it onto `main` at `854060fb`.

What changed:

- `Scheduler.shutdown()`, pending-remove draining, `deep_reset()`, and BatchGenerator/MLX teardown remain on the engine's `_mlx_executor`. A worker `RuntimeError` is not retried on the close thread.
- SSD persistence uses one monotonic absolute deadline through hot-cache flushing, pre-enqueue index/eviction preparation, queue admission/full handling, and writer join. The bounded path never falls back to an unbounded inline write.
- Calls without a deadline retain complete-flush semantics, including writes admitted before shutdown in both hot-cache and direct/no-hot-cache modes.
- SSD shutdown is signaled before waiting for active `store_cache` workers. Workers stop before preparing another block; if one cannot quiesce within the budget, teardown preserves the fatal/restart-safe behavior instead of clearing futures or tearing down owned state.
- Deadline eviction rollback no longer restores metadata after a concurrent writer could have unlinked the corresponding file.
- The slow-shutdown tests now import and use `time`, so the delay is actually exercised.

Local verification on exact commit `08ede2caca09fe04d7e595b0802387e164b098e0`:

- `590 passed` across `tests/test_paged_ssd_cache.py`, `tests/test_prefix_cache.py`, `tests/test_scheduler.py`, and `tests/test_engine_core.py`
- 21 focused deadline/concurrency regressions repeated 20 times: `420/420 passed`
- committed-range `git diff --check`: clean
- changed-line Ruff and format checks: clean

GitHub CI is pending on the updated head.

mvdbos:
Follow-up: the first CI run exposed one stale test expectation in `tests/test_hot_cache.py`: it still expected a single `queue.put(timeout=1.0)`, while the revised admission path intentionally uses interruptible `0.05s` polls.

I updated that regression to use deterministic monotonic time and assert the polling contract; production code was unchanged by this follow-up.

Final head: `e6786223c31d3e832a7b30bd5de9409f38730f0f`

- Expanded local affected-file suite: `631 passed`
- GitHub CI: Python 3.11, 3.12, and 3.13 all passed

mvdbos:
@jundot ok to merge now?

mvdbos:
Updated to address teardown safety. 

Behavior changes to be aware of:
1. `EngineCore.close()` now fatal-exits (restart-safe) if the SSD writer survives the deadline **or** the MLX executor rejects teardown dispatch — no longer reports a half-torn-down engine as closed.
2. The finite persistence deadline is sticky through `close()`/`deep_reset()`; a later no-deadline call cannot downgrade a bounded teardown.
3. Forced model-ownership transfer runs the previous owner's `deep_reset()` on its own MLX executor and aborts the transfer if it can't complete.
4. `_write_block_file` uses unique temp names and never unlinks a final file it didn't commit.

Verified: 713 passed.

mvdbos:
@jundot

Issue #2619, opened 2026-08-13
oQ8: sensitivity proxy load fails with "Received 3 parameters not in model: lm_head.biases, lm_head.scales, lm_head.weight" on Gemma-4 MoE heretic model


Hexx911:
Quantizing a Gemma-4 26B-A4B MoE (heretic/abliterated) checkpoint to oQ8e fails during the sensitivity measurement stage, after imatrix calibration completes successfully. The uniform 4-bit proxy built for sensitivity scoring contains a separately quantized lm_head (.weight/.scales/.biases), but the model class used to reload that proxy doesn't expect a standalone lm_head module likely because this architecture normally ties lm_head to embed_tokens, and the write-path sanitize (which untied/quantized lm_head per the lm_head=8b plan override) isn't mirrored by the read-path sanitize used for sensitivity scoring.
This looks like the same root-cause pattern as #1296 (DeepSeek-V4 MTP: write-path sanitize adds tensors the read-path model class doesn't expect) and possibly related to the sanitize-discovery issue in #1556 (Qwen heretic model, same _discover_sanitize_plan/sanitize_weights code path, same all-zero-sensitivity symptom right before failure).

1. Point oMLX's oQ admin panel at the G4-Moonlight-Dusk-26B-A4B-heretic checkpoint.
2. Start an oQ8e quantization job (default settings, no explicit sensitivity_model_path).
3. Calibration footprint (49.4 GB) exceeds the 75% calibration-capacity threshold (34.4 GB available, 25.8 GB limit), so oMLX auto-builds a uniform 4-bit proxy on disk.
4. imatrix calibration runs to completion (1024/1024 samples) and writes its cache successfully.
5. oMLX then builds a second uniform 4-bit proxy for sensitivity measurement.
6. Loading that proxy fails immediately:

Expected behavior
Sensitivity measurement should complete and quantization should proceed to the budget plan / GPTQ stages, as it does for other Gemma-4 checkpoints.

oMLX version: 0.5.7
Model: G4-Moonlight-Dusk-26B-A4B-heretic (Gemma-4 MoE, 26B-A4B, heretic/abliterated variant)
Platform: MacBook, M5 Pro, 64GB unified memory
Target format: oQ8e

PR #2634, opened 2026-08-13
fix(oq): requantize MLX affine sources safely

### LINKED ISSUE #2452: oq: quantize_oq_streaming crashes on MLX-affine quantized sources (uint32)
**Summary**: `quantize_oq_streaming` cannot consume checkpoints that are already MLX-affine quantized (uint32 packed weight + bf16 scales/biases): `_classify_pair` only recognizes fp8/mxfp8/nvfp4-style pairs, so affine triples fall through and `_quantize_chunked` receives uint32 → `ValueError: [quantize] Only real floating types can be quantized but w has type uint32`. Crashes in BOTH the sensitivity-proxy builder and the main streaming loop.

**Repro**: `quantize_oq_streaming(model_path=<any mlx-community affine checkpoint>, output_path=..., oq_level=3)`.

**Why it matters**: requantizing community 4-bit releases to lower bpw is a common need on 128GB machines (e.g. Inkling-Small REAP25 4-bit → 3-bit experts to fit serving headroom). We worked around it with an external dequant→requant pass; oq's calibrated recipes would do it better.

PhilipJohnBasile:
## Summary

- recognize ordinary MLX affine quantization triples (`uint32` packed weights plus scales and biases)
- reconstruct logical floating tensors before oQ requantization
- support both the streaming conversion and sensitivity-proxy paths
- keep weight/scale/bias dictionary mutations atomic while preserving existing FP8, MXFP, and virtual-tensor handling

## Root cause

`_LazyTensorIndex` did not classify MLX affine triples as a logical quantized tensor. The packed `uint32` weight therefore reached `mx.quantize`, which only accepts real floating inputs. Companion scale and bias tensors could also leak through as independent model parameters.

## Impact

This allows `quantize_oq_streaming` to consume existing MLX-affine checkpoints, including multidimensional expert tensors, while failing closed when quantization metadata is missing or inconsistent.

This should also be relevant to #2619, but this PR does not claim that issue is closed without its real Gemma-4 checkpoint fixture.

Fixes #2452.

## Validation

- focused affine/streaming/proxy regression selection: 58 passed
- broader independent selection: 63 passed
- real Inkling index audit: 726 affine triples recognized, 1,522 logical tensors exposed, and no scale/bias sidecars leaked
- invalid and missing metadata cases fail closed

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -100,6 +100,8 @@ _VALID_QUANT_BITS = (2, 3, 4, 5, 6, 8)
 
 _NATIVE_FLOAT8_QUANT_METHODS = frozenset(("fp8", "mxfp8"))
 
+_AFFINE_SCALE_DTYPES = frozenset(("BF16", "F16", "F32"))
+
 
 def _bpw_targets_for_level(oq_level: float) -> tuple[float, float] | None:
     """Return (target_bpw, hard_cap_bpw) for the given oQ level, or None."""
@@ -2761,6 +2763,11 @@ class _DiscoveredPlan:
             return self._lazy.materialize_virtual(src_key)
         if hasattr(self._lazy, "_fp8_pairs") and src_key in self._lazy._fp8_pairs:
             return self._lazy._dequant_one(src_key)
+        if (
+            hasattr(self._lazy, "_affine_triples")
+            and src_key in self._lazy._affine_triples
+        ):
+            return self._lazy._dequant_affine_one(src_key)
         meta = self._lazy._index.get(src_key)
         if meta is None:
             raise KeyError(f"source tensor {src_key!r} not in lazy index")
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the careful work on this. The affine-triple detection, metadata validation, atomic sidecar handling, and regression coverage are all thoughtfully implemented.

After reviewing the intended product behavior, I don't want oQ to accept ordinary low-bit MLX affine checkpoints as quantization sources. oQ recipes are intended to be derived from an original-precision BF16/FP16 checkpoint. Dequantizing an existing 4-bit checkpoint reconstructs floating tensors, but it cannot recover the information already lost during the first quantization. Re-quantizing those tensors adds a second quantization error, while sensitivity measurement and mixed-precision allocation are also based on the already-quantized model rather than the original weights. A loadable output would therefore not provide the same quality guarantees as an oQ model produced from the original checkpoint.

The existing native FP8/MXFP8 source paths are narrow, explicit exceptions for floating quantization formats that oQ already handles as reconstructable source representations. I don't want to extend that exception to ordinary low-bit affine checkpoints or expose 4-bit-to-lower-bit conversion as a supported oQ workflow. Keeping these models excluded from `validate_quantizable()` and the source-model picker is intentional; the lower-level path should also fail early with a clear unsupported re-quantization message instead of attempting the conversion.

For the same reason, I plan to treat #2452 as an unsupported workflow rather than expand oQ's input policy. The Gemma 4 proxy failure in #2619 should be addressed separately by fixing the write-path/read-path sanitize mismatch, not by enabling general pre-quantized affine sources.

I appreciate the amount of care that went into this PR, but I’m going to close it because the proposed behavior conflicts with oQ’s source-quality policy.

PR #2635, opened 2026-08-13
fix(memory): reclaim MLX pool before hot-cache shrink

### LINKED ISSUE #2581: Memory enforcer counts the reclaimable MLX buffer pool as live usage, so hard pressure zeroes the hot cache instead of draining the pool (562 events; 48GB cache stuck at ~1GB)
## Summary

`ProcessMemoryEnforcer` counts MLX's **reclaimable buffer pool** as live process usage. Because that pool is routinely 60–90 GB, `current` reads near the ceiling while the true working set is a fraction of it. At hard pressure the enforcer then shrinks the **hot cache** — which holds 1–3 GB — instead of draining the pool that holds the other 60–90 GB.

The result is that `_shrink_hot_cache_for_pressure` computes a target of **zero** on essentially every pressure event, so the paged-cache hot tier is wiped repeatedly and can never accumulate. On this host that produced **562 zeroing events** and left a configured 48 GB hot cache sitting at ~1 GB for 13 days.

The pool is demonstrably recoverable — the enforcer's own reclaim path frees 31–35 GB in a single call — so this is an accounting problem, not a genuine shortage.

## Evidence that `current` is dominated by reclaimable memory

`current` collapses by ~90 GB in a single 1-second poll the moment the pool is cleared:

```
Memory pressure level: soft -> hard (current=103.3GB)
Memory pressure level: hard -> ok   (current=12.8GB)
Memory pressure level: soft -> hard (current=102.7GB)
Memory pressure level: hard -> ok   (current=12.8GB)
```

Nothing was unloaded between those lines. The 90 GB delta is the buffer pool being returned. Corresponding reclaim amounts:

```
Reclaimed 35.50GB of pooled Metal buffers
Reclaimed 32.19GB of pooled Metal buffers
Reclaimed 31.60GB of pooled Metal buffers
```

Source of the inflation — `process_memory_enforcer.py`:

```python
return max(mx.get_active_memory(), phys)     # phys_footprint includes the pool
```

The pool is pinned by `mx.set_cache_limit(total_mem)` (`cli.py`, `server.py`), which is deliberate for the #300 kernel-panic guard: freed buffers stay pooled and are only returned via an explicit `mx.clear_cache()`. That is fine — but the enforcer treats those pooled bytes as unavailable when they are the *most* available memory in the process.

## How this zeroes the hot cache

```python
target_hot_bytes = max(0, hot_used - (current - target))
```

`current - target` is the overshoot above the soft watermark. Since `current` is inflated by the pool, the overshoot is permanently several GB while `hot_used` is 1–3 GB — so the expression floors at **0** and the cache is drained completely.

It is self-reinforcing: once drained, the cache is small, which guarantees the next event floors at zero again. It can never grow back toward its configured budget.

Distribution of `target_hot` on this host, spanning the 0.5.3 upgrade:

| date | non-zero targets | zeroed |
|---|---|---|
| 2026-07-24 | 7 | 0 |
| 2026-07-27 | 1 | **33** |
| 2026-08-09 | 0 | **13** |

Before 0.5.3 the shrink trimmed to sensible targets (13–37 GB observed). After, it is zero essentially always. **562** `target_hot=0.0GB` events total.

## The ordering is backwards, and the codebase already knows the right order

At hard pressure:

```python
if new_level == "hard":
    freed_hot = await asyncio.to_thread(
        self._shrink_hot_cache_for_pressure, current, soft)   # 1. wipe the cache (1-3 GB)
    if os.environ.get("OMLX_DISABLE_PRESSURE_RECLAIM") != "1":
        self._request_scheduler_cache_reclaim(freed_hot)      # 2. then ask the pool to drain (60-90 GB)
```

The cheap, ineffective action runs first; the expensive, effective one is requested afterwards and lands asynchronously at the next step boundary.

Twenty lines below, the **abort** path already implements exactly the reasoning that is missing here:

```
# Reclaim-before-abort grace: when the MLX buffer pool alone holds more
# than the reclaim floor, the pressure clear requested above can very
# likely recover below the watermark without killing anything (measured:
# aborts fired 0.1GB over the watermark with 3.7GB of pooled buffers on
# the table). The drain runs on the inference thread at the next
# prefill-chunk or step boundary, so give it a bounded number of polls.
# Never defers an emergency (at/over the ceiling).
```

So the project already accepts that a bounded wait for the pool to drain is preferable to destroying something valuable — it just applies that only to in-flight requests, not to the cache.

## Suggested fix

Apply the same reclaim-first grace to the hot-cache shrink:

1. At hard pressure, request the pool reclaim **before** shrinking, when `pool_bytes > _POOL_RECLAIM_FLOOR`.
2. Give it the same bounded grace (`_BUSY_ABORT_GRACE_POLLS_MAX`-style counter), re-measuring `current` each poll.
3. Only shrink the hot cache if the drain did not bring `current` under the watermark.
4. Keep the existing emergency carve-out so at/above-ceiling still acts immediately.

This does not make the guard more permissive — memory is still actually freed before proceeding. It just frees the 60–90 GB of genuinely reclaimable buffers before sacrificing a 1–3 GB cache that costs real prefill time to rebuild.

A narrower alternative, if the grace is unwanted: subtract `mx.get_cache_memory()` from `current` **only** for the purpose of computing `target_hot_bytes`, so the shrink is sized against live memory rather than pooled memory.

## Impact when fixed

Measured on this host after working around it (raising `iogpu.wired_limit_mb` so the watermark stopped being crossed):

- hot cache grew 0 → 32 GB and stayed
- cold prefill of a 21k-token prompt **6.5 s** → warm with 20,480 tokens reused **0.8 s** (**8×**)

That speedup was unavailable for 13 days because the cache was being wiped faster than it could fill.

## Environment

- MacBook Pro `Mac17,7`, 128 GB unified, macOS 26 (Darwin 25.4.0)
- oMLX **0.5.4** (build 2121); behaviour first appears in 0.5.3
- `hot_cache_max_size: 48GB`, `memory_guard_tier: balanced`
- Models: Qwen3.6-35B-A3B-nvfp4, Qwen3.5-4B-MLX-4bit, Qwen3-Embedding-4B

Related: #2576 (the same full hot cache then makes teardown exceed the 60 s fatal watchdog).

PhilipJohnBasile:
## Summary

- request a synchronized MLX buffer-pool reclaim before destructive hard-pressure enforcement
- defer hot-cache shrink, eviction, and request abort for at most five polls, and only when a scheduler accepted the reclaim request
- preserve immediate enforcement for emergency pressure, disabled reclaim, small or unreadable pools, missing schedulers, reclaim-request failures, and exhausted grace
- keep the existing per-tick store-cache concurrency adjustment active during the grace

## Root cause

`ProcessMemoryEnforcer` compares a process footprint that includes MLX's reclaimable buffer pool against its watermarks. At hard pressure it previously sized and performed hot-cache shrink before asking the scheduler to clear that much larger pool. When pooled buffers dominated the footprint, the computed hot-cache target repeatedly fell to zero even though a synchronized pool clear could recover the pressure without destroying reusable prompt state.

The reclaim remains scheduler-owned: this change only sets the existing pressure-reclaim flag, and the actual Metal clear still runs at the synchronized inference-step boundary.

## Validation

- `uv run pytest -q tests/test_process_memory_enforcer.py`: 163 passed
- `uv run pytest -q tests/test_process_memory_enforcer.py tests/test_scheduler.py tests/test_hot_cache.py -m 'not slow and not integration'`: 418 passed
- targeted Ruff check passed with the pre-existing unrelated `SIM117` finding excluded
- `git diff --check` passed
- independent frozen-diff review: approved with no blocking findings

This validates ordering and bounded fallback in CPU tests. It does not yet claim live M5 physical-memory recovery, hot-cache retention, or concurrent multi-engine performance; those remain follow-up workload measurements.

Closes #2581

### DIFF
--- a/omlx/process_memory_enforcer.py
+++ b/omlx/process_memory_enforcer.py
@@ -385,10 +385,11 @@ class ProcessMemoryEnforcer:
         # admission control. Updated on every poll iteration.
         self._pressure_level: str = "ok"
         self._over_ceiling_polls: int = 0
-        # Consecutive hard-pressure polls we have deferred the busy-request
-        # abort while a fat Metal buffer pool drains at the next prefill
-        # chunk / step boundary. See the grace check in _check_and_enforce.
-        self._busy_abort_grace_polls: int = 0
+        # Consecutive hard-pressure polls we have deferred destructive
+        # enforcement while a fat Metal buffer pool drains at the next
+        # prefill chunk / step boundary. See the grace check in
+        # _check_and_enforce.
+        self._pressure_reclaim_grace_polls: int = 0
         # Last value passed to mx.set_wired_limit (0 if not yet applied
         # or the call failed). Used by the admin dashboard to surface a
         # warning when the kernel iogpu.wired_limit_mb is below this.
@@ -487,6 +488,7 @@ class ProcessMemoryEnforcer:
         """
         if self._running:
             return
+        self._pressure_reclaim_grace_polls = 0
         if self._prefill_memory_guard:
             self._refresh_effective_metal_cap_bytes()
         self._running = True
@@ -1010,10 +1012,10 @@ class ProcessMemoryEnforcer:
     # _periodic_clear_threshold_bytes floor). Above it, a clear meaningfully
     # returns memory to the OS.
     _POOL_RECLAIM_FLOOR = 2 * 1024**3
-    # Max consecutive hard-pressure polls the busy-request abort is deferred
+    # Max consecutive hard-pressure polls destructive enforcement is deferred
     # while a reclaimable pool drains. Prefill chunks run ~5s at full step
     # size, so 5 one-second polls cover at least one chunk boundary.
-    _BUSY_ABORT_GRACE_POLLS_MAX = 5
+    _PRESSURE_RECLAIM_GRACE_POLLS_MAX = 5
 
     def _pool_bytes(self) -> int:
         """MLX buffer-pool size, 0 when unreadable (mocked mx in tests)."""
@@ -1022,7 +1024,7 @@ class ProcessMemoryEnforcer:
         except (TypeError, ValueError):
             return 0
 
-    def _request_scheduler_cache_reclaim(self, freed_hot: int) -> None:
+    def _request_scheduler_cache_reclaim(self, freed_hot: int) -> int:
         """Ask each scheduler to run its step-boundary Metal cache clear.
 
         Routes the reclaim through the scheduler's shipped, lock-protected,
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the clean fix. I reviewed it against current main and I like that the grace generalizes the existing abort deferral instead of adding a new mechanism, and every bail-out path (emergency, disabled reclaim, small pool, missing scheduler, failed request) still falls through to the old enforcement.

One thing that settled the safety question for me: since set_cache_limit pins the pool, the old immediate shrink never actually lowered phys_footprint until the step-boundary clear ran anyway, so requesting the clear first is strictly better ordering rather than a loosening of the guard. The idle-engine case is covered too because has_work stays true while the clear is pending.

Live hot-cache retention numbers can come from the reporter's setup in #2581. Merging.

PR #2637, opened 2026-08-13
fix(admin): only apply global settings at runtime when values actually change

bitfieldz:
## Problem

POST /api/global-settings unconditionally triggered model unloads for any cache, network, or sampling setting that was sent in the request, even if unchanged. The WebUI sends all form fields on save, so this caused spurious model reloads on every settings save.

## Root Cause

`update_global_settings()` in `omlx/admin/routes.py` uses `is not None` checks on all request fields:

```python
if request.cache_enabled is not None:
    global_settings.cache.enabled = request.cache_enabled
    cache_changed = True
```

The WebUI serializes all form fields on save. Even unchanged values pass the `is not None` guard, triggering runtime apply functions like `_apply_cache_settings_runtime()` which unload all models but do NOT re-discover them.

## Fix

Add `_apply_if_changed()` helper that compares `request_value` against `current_value` before applying runtime changes:

```python
def _apply_if_changed(
    request_value: Any,
    current_value: Any,
    settings_obj: Any,
    attr_name: str,
    runtime_key: str,
    runtime_applied: list[str],
) -> bool:
    if request_value is not None and request_value != current_value:
        setattr(settings_obj, attr_name, request_value)
        runtime_applied.append(runtime_key)
        return True
    return False
```

## Coverage

Applied to:
- Cache: `cache_enabled`, `ssd_cache_dir`, `ssd_cache_max_size`, `hot_cache_max_size`, `hot_cache_only`, `initial_cache_blocks`
- Network: `network_http_proxy`, `network_https_proxy`, `network_no_proxy`, `network_ca_bundle`
- Sampling: `sampling_max_context_window`, `sampling_max_tokens`, `sampling_temperature`, `sampling_top_p`, `sampling_top_k`, `sampling_repetition_penalty`, `sampling_max_context_window_policy`
- HF/MS endpoints: `hf_endpoint`, `ms_endpoint`

## Verification

```bash
cd /Users/aaron/Code/omlx
/Users/aaron/.local/share/uv/tools/omlx/bin/python -m py_compile omlx/admin/routes.py
# Output: OK
```

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -408,6 +408,25 @@ def _format_cache_size(size_bytes: int) -> str:
     return f"{mb:.0f}MB"
 
 
+def _apply_if_changed(
+    request_value: Any,
+    current_value: Any,
+    settings_obj: Any,
+    attr_name: str,
+    runtime_key: str,
+    runtime_applied: list[str],
+) -> bool:
+    """Apply a global setting at runtime only if it actually changed.
+
+    Returns True if the setting was changed and added to runtime_applied.
+    """
+    if request_value is not None and request_value != current_value:
+        setattr(settings_obj, attr_name, request_value)
+        runtime_applied.append(runtime_key)
+        return True
+    return False
+
+
 def _parse_hot_cache_max_size(value: str) -> int:
     """Parse hot cache max size. Hot cache does not support an auto sentinel."""
     from ..config import parse_size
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for working on this. Since this is still a draft, I’m leaving a few early notes that may need another pass. A no-op GET-to-POST round trip can still look changed because the API returns resolved cache directory and size values, and the GDN cache fields added in #2644 will need the same comparison handling after rebasing. `hot_cache_only` and `initial_cache_blocks` also appear to unload models without updating the scheduler template on this branch. Once these cases are covered and the branch is ready, please mark the PR ready for review and I’ll take another look.

PR #2639, opened 2026-08-13
fix(memory): align DS-V4 prefill estimate with live WSDPA routes

mvdbos:
## Summary

**Why:** Normal `Scheduler` instances install `_DeepSeekV4PrefillMemoryProfile`, so the earlier process-wide head-dimension registration did not change the estimator used by admission and adaptive prefill throttling. It could also undercount when the custom WSDPA path was disabled, unsupported, or had failed and the model fell back to stock attention.

**What:** Move the optimization into the DeepSeek-V4 profile and allow the lower estimate only after the matching BF16 kernel route has successfully evaluated once. Cold/unproven, disabled, unsupported, and broken routes retain the existing conservative fallback price.

## Repro

Validated on Apple M5 Max, 128 GB. This revision exercises the real DeepSeek-V4 BF16 dense and top-k Metal kernels plus deterministic Scheduler/profile tests. It does not load model weights or claim an end-to-end throughput change.

Representative normal-Scheduler monitor input:

- query chunk: 2,048 tokens
- KV length: 66,000 tokens
- DeepSeek-V4: 64 query heads, head dimension 512, window 128, top-k 512
- BF16 activation dtype, native indexer eligible

| Route state | Estimated transient |
|---|---:|
| Current `main` | 3,001,237,504 B (2.795120 GiB) |
| Revised, cold/unproven | 3,001,237,504 B (2.795120 GiB) |
| Revised, dense + top-k routes proven | 542,051,328 B (0.504825 GiB) |
| Revised, `_broken` after profile construction | 3,001,237,504 B (2.795120 GiB) |

The active-route estimate is 81.939% lower for this input. The cold and post-failure rows deliberately remain identical to the conservative current-main estimate.

## How

- Add one live dense/top-k route-state contract in `wsdpa_attention.py`.
- Keep the route conservative until the first lazy MLX kernel output has been evaluated inside the fallback boundary.
- Keep the independent top-k WSDPA attempt available when the optional GLM native sparse kernel has been disabled; that process-global flag now guards only the GLM route.
- Price the actual DeepSeek-V4 branches separately:
  - local-only attention;
  - ratio-128 local + pooled attention;
  - ratio-4 dense pooled attention while the pool is at or below `index_topk`;
  - ratio-4 top-k attention only when the independent top-k route is proven active.
- Reuse the existing stock dense/sparse estimates whenever route eligibility is not established, and switch back immediately when `_broken` changes.
- Derive the normal Scheduler's activation dtype from the model's explicit dtype or its embedding weight; the actual DeepSeek-V4 `Model` has no top-level `dtype` attribute.
- Do not register head dimension 512 globally. Existing head-dim-256 behavior is unchanged.

## Tests

```text
Python 3.11: focused route/profile/Scheduler/dispatch suite: 216 passed, 4 skipped
Python 3.12: focused route/profile/Scheduler/dispatch suite: 216 passed, 4 skipped
Python 3.13: focused route/profile/Scheduler/dispatch suite: 216 passed, 4 skipped
Python 3.11: pytest tests/ -m "not slow and not integration": 9,663 passed, 110 skipped, 77 deselected
```

The focused suite includes real Metal dispatch/evaluation for the dense and top-k kernels, first-evaluation failure, disabled and broken fallback transitions, unsupported dtype/shape rejection, branch-specific allocation formulas, the normal Scheduler path, and proof that importing WSDPA does not globally register head dimension 512.

Additional checks: `compileall`, `git diff --check`, Black on every changed file, and Ruff fatal-error rules (`E9`, `F63`, `F7`, `F82`).

### DIFF
--- a/omlx/memory_monitor.py
+++ b/omlx/memory_monitor.py
@@ -908,6 +908,46 @@ class _DeepSeekV4PrefillMemoryProfile:
     index_head_dim: int
     index_topk: int
     dtype_size: float
+    wsdpa_dtype_supported: bool = False
+
+    def _wsdpa_route_active(self, *, topk: bool = False) -> bool:
+        """Match the live WSDPA route without a process-wide head-dim flag."""
+        if (
+            not self.wsdpa_dtype_supported
+            or self.num_attention_heads != 64
+            or self.head_dim != 512
+        ):
+            return False
+        try:
+            from omlx.patches.deepseek_v4.wsdpa_attention import (
+                wsdpa_prefill_route_active,
+            )
+
+            return wsdpa_prefill_route_active(topk=topk)
+        except Exception:
+            return False
+
+    def _wsdpa_attention_bytes(
+        self,
+        query_tokens: int,
+        local_tokens: int,
+        pooled_tokens: int = 0,
+        *,
+        selected_tokens: int = 0,
+    ) -> int:
+        """Bound the custom kernel's contiguous inputs and output.
+
+        The kernel performs online softmax in registers and never materializes
+        a score matrix. Count potentially copied query/KV/top-k inputs and keep
+        the generic estimator's fp32-width output bound for conservatism.
+        """
+        query = (
+            self.num_attention_heads * query_tokens * self.head_dim * self.dtype_size
+        )
+        keys = (local_tokens + pooled_tokens) * self.head_dim * self.dtype_size
+        output = self.num_attention_heads * query_tokens * self.head_dim * 4
+        topk = query_tokens * selected_tokens * 4
+        return int(query + keys + output + topk)
 
     def _pool_cache_elements(
         self,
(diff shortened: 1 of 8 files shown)


jundot:
Thanks for digging into the long-context estimate. I confirmed the underlying overestimate is real, but this change updates the wrong estimator: the normal Scheduler installs `_DeepSeekV4PrefillMemoryProfile`, so the registration does not change its throttle calculation. The unconditional global registration can also undercount stock SDPA when the kernel is disabled, unsupported, or fails, which makes this unsafe to merge as-is. Could you move the adjustment into the DeepSeek-V4 profile, keep it synchronized with the actual kernel route, and add coverage for both the normal Scheduler path and disabled or failed fallbacks?

mvdbos:
Thanks — you were right that normal Scheduler installs `_DeepSeekV4PrefillMemoryProfile`, so the previous global registration targeted the wrong estimator and was unsafe for fallback routes.

I rebuilt the change on current `main` and moved the accounting into the DeepSeek-V4 profile. The lower estimate is enabled only after the matching BF16 kernel route evaluates successfully; cold, disabled, unsupported, and failed routes keep the existing conservative fallback estimate.

Dense and top-k routes are tracked independently. I also separated the top-k WSDPA attempt from the optional GLM native sparse disable flag so the estimator cannot report that route active while dispatch skips it.

The revision includes coverage for normal Scheduler wiring, first-use evaluation failures, disabled and broken fallbacks, unsupported dtypes, route-specific formulas, and the absence of a global head-dim-512 side effect.

mvdbos:
@jundot Sorry, lost some commit history, so only one commit now. All your points addresed. Hopefully still properly reviewable

jundot:
Thanks for rebuilding this. The live dense and top-k route state now feeds the DeepSeek-V4 profile directly, while cold, disabled, unsupported, and failed paths keep the conservative fallback estimate with no global head-dim-512 side effect. I verified the focused suite and a real 5,005-token prefill with the local DeepSeek-V4 model, with no WSDPA errors or crash. LGTM after the route-state smoke; ready to merge.

PR #2640, opened 2026-08-13
Support DeepSeek V4 target-only MLX artifact

PhilipJohnBasile:
## Summary

- Add fail-closed loading for the exact public DeepSeek V4 target-only MLX artifact, including converter-name canonicalization, immutable raw-config validation, complete affine-triplet checks, and strict final weight loading.
- Detect nested 2/3-bit quantization overrides before selecting native ratio-128 attention.
- Preserve the hot-prefix cache for a bounded five pressure polls while reclaimable MLX buffer-pool memory drains; retain immediate shrink under emergency pressure or `OMLX_DISABLE_PRESSURE_RECLAIM=1`.
- Package the optional native DeepSeek indexer with an explicit Python-ABI guard so incompatible extension suffixes fail before app staging.
- Normalize top-level and nested chat thinking controls, reject contradictory values, and keep a zero thinking budget from silently enabling thinking.

## Target artifact

Validated against public immutable revision `2566775a5e97986ffabbf592fe84626c385fa10b` of `philipjohnbasile/DeepSeek-V4-Flash-0731-MLX-M5Max-TargetOnly`. Its `config.json` is 37,869 bytes with SHA-256 `ab61e3230f196c6eba04bfa81158dd527a7f356b6d926cc4794907a19f35b75d`.

The artifact is intentionally target-only: `num_nextn_predict_layers=0`, DSpark is disabled, and no MTP compatibility is claimed.

## Verification

- Upstream CI: Python 3.11, 3.12, and 3.13 all pass.
- Exact 44-shard artifact loaded through the OpenAI-compatible server on an M5 Max; the warm arithmetic request returned the correct final answer with reasoning separated.
- Exact unload reclaimed 96.71 GB and settled MLX active memory to approximately 1 KB before clean shutdown.
- Optional native indexer build exposed the required DSA symbols and passed focused correctness/ABI probes. Short, no-cache diagnostic repeats improved from 7.0 tok/s on the fallback path to a native median of 11.7 tok/s with identical correct output.
- A separate ad-hoc preview app built with the donor CPython 3.11 runtime passed extension-ABI admission, native symbol/MLX-array probes, exact model loading, generation, unload, and deep signature verification at build time.
- Full process-memory suite: 162 passed; focused DeepSeek architecture/config/load policy, model discovery, OpenAI routing, thinking-control, native-indexer, and app-ABI tests passed.
- Independent frozen-diff reviews passed for the loader, app ABI, and thinking-control changes.

## Evidence boundary

The real-model runs prove load, generation, native-path availability, and unload—not representative throughput or parity with the original FP8 service. The 11.7 tok/s result is a short diagnostic; packaged-app timing was contaminated by unrelated disk activity and is not promoted as a benchmark. The preview bundle was an isolated ad-hoc validation artifact, not a notarized public release. Normal installations without the optional native extension retain the slower fallback path.

### DIFF
--- a/apps/omlx-mac/Scripts/build.sh
+++ b/apps/omlx-mac/Scripts/build.sh
@@ -114,6 +114,7 @@ CUSTOM_KERNEL_DIRS=(
     "$REPO_ROOT/omlx/custom_kernels/minimax_m3"
     "$REPO_ROOT/omlx/custom_kernels/qwen35_prefill"
 )
+CUSTOM_KERNEL_ABI_CHECKER="$SCRIPT_DIR/check_custom_kernel_python_abi.py"
 # OMLX_EXPORT_DIR overrides the venvstacks export tree we copy Python
 # layers from. Release builds use this to point at a per-target export
 # copy with platform-specific mlx-metal wheels swapped in.
@@ -225,6 +226,40 @@ _custom_kernel_pythonpath() {
     else
         printf "%s\n" "$mlx_site"
     fi
+}
+
+_custom_kernel_donor_python() {
+    [ -n "${DONOR_LAYERS:-}" ] || die "custom kernel build requires resolved donor layers."
+    local donor_python="$DONOR_LAYERS/cpython-3.11/bin/python3"
+    [ -x "$donor_python" ] \
+        || die "donor missing executable bundled Python: $donor_python"
+    printf "%s\n" "$donor_python"
+}
+
+_validate_custom_kernel_python_abi() {
+    # The build-time ABI probe below runs under PYTHON_BIN. That can be a
+    # different CPython than the donor app's embedded interpreter, which
+    # would let a cpython-313 _ext pass the probe before it is bundled into a
+    # cpython-311 app. Compare the actual interpreters before compiling.
+    local donor_python
+    donor_python="$(_custom_kernel_donor_python)"
+    [ -f "$CUSTOM_KERNEL_ABI_CHECKER" ] \
+        || die "custom kernel Python ABI checker missing: $CUSTOM_KERNEL_ABI_CHECKER"
+    "$PYTHON_BIN" "$CUSTOM_KERNEL_ABI_CHECKER" \
+        --build-python "$PYTHON_BIN" \
+        --donor-python "$donor_python" \
+        || die "custom kernel build Python ABI must match the donor app; see output above."
+}
+
+_validate_packaged_custom_kernel_extensions() {
+    # Check the files actually copied into the staged app as well. This
+    # catches stale _ext files that a source-tree build did not replace.
+    local donor_python="$1"
+    local packaged_kernels="$2"
+    "$PYTHON_BIN" "$CUSTOM_KERNEL_ABI_CHECKER" \
+        --donor-python "$donor_python" \
+        --extension-directory "$packaged_kernels" \
+        || die "staged custom kernel extension suffixes do not match the donor app; see output above."
 }
 
 _clean_custom_kernel_build_artifacts() {
@@ -359,6 +394,7 @@ _build_custom_kernels() {
     cmake_args="${CMAKE_ARGS:-}"
     custom_kernel_pythonpath="$(_custom_kernel_pythonpath)"
 
+    _validate_custom_kernel_python_abi
     _check_custom_kernel_nanobind "$custom_kernel_pythonpath"
     log "Building optional native custom kernels (macOS deployment target $deployment_target)…"
     _clean_custom_kernel_build_artifacts
@@ -574,6 +610,11 @@ rsync -a \
     "${RSYNC_EXCLUDES[@]}" \
     "$REPO_ROOT/omlx/" "$RESOURCES_DIR/omlx/"
 ok "  + omlx package"
+
+if [ "$WITH_CUSTOM_KERNEL" = "1" ]; then
+    _validate_packaged_custom_kernel_extensions \
+        "$(_custom_kernel_donor_python)" "$RESOURCES_DIR/omlx/custom_kernels"
+fi
 
 log "Writing engine commit metadata..."
 "$PYTHON_BIN" "$PACKAGING_DIR/build.py" --write-engine-commits "$RESOURCES_DIR/omlx" \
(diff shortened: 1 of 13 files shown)


PhilipJohnBasile:
Real M5 Max validation update (commit `2cb2cde9`):

- Loaded the exact public target-only artifact: 44 shards / 103,849,215,724 weight bytes; `config.json` 37,869 bytes, SHA-256 `ab61e3230f196c6eba04bfa81158dd527a7f356b6d926cc4794907a19f35b75d`.
- Public load endpoint returned HTTP 200 in 9.639 s. oMLX reported 96.92 GB actual model memory and 101.55 GB full-model accounting.
- First cold non-stream request completed HTTP 200: 10 prompt + 5 output tokens in 8.66 s (0.6 tok/s); the deliberately tiny output cap truncated reasoning.
- Warm non-stream request completed correctly: answer `4`, 16 prompt + 19 output tokens in 2.70 s (7.0 tok/s), with reasoning separated from content.
- Explicit unload returned HTTP 200, freed 96.71 GB, and settled MLX active memory to about 1 KB before clean server shutdown.

This proves real artifact load and OpenAI-compatible generation on the reviewed branch. It is not a representative throughput approval: the optional native DeepSeek indexer extension was absent, and oMLX warned that long-context prefill uses the slower MLX fallback.

PhilipJohnBasile:
Native M5 indexer follow-up (same commit/model/prompt):

The source tree already contains the native DeepSeek V4 indexer. Building the optional extension with `OMLX_WITH_CUSTOM_KERNEL=1` produced a valid MLX ABI canary and exposed `dsa_indexer_scores`, `dsa_topk_indices`, and `deepseek_v4_sparse_attention`; 12 focused native correctness tests passed. Runtime then logged the native indexer path as available.

Exact no-cache warm repeats for the same 16-prompt-token / 19-output-token arithmetic request:

- fallback MLX indexer: 2.70 s, 7.0 tok/s
- native: 1.67 s, 11.4 tok/s
- native repeat: 1.61 s, 11.8 tok/s
- native repeat: 1.62 s, 11.7 tok/s

Native median: 11.7 tok/s, about 67% above the fallback observation. All responses were identical and correct (`4` with separated reasoning). The native run unloaded cleanly, freeing 96.71 GB and settling active MLX memory to ~1 KB.

This remains a short diagnostic, not a representative throughput score. Normal installations that omit the optional extension still take the slower fallback path; app/release packaging should include the extension for DeepSeek V4 users.

PhilipJohnBasile:
Packaged-app follow-up on head a34b0408:

- Built a separate ad-hoc oMLX 0.6.0.dev1 preview bundle from this branch with the donor app’s embedded CPython 3.11, MLX 0.32.0, and native custom kernels. The installed app was not replaced.
- Deep signature verification passed; all packaged extension tags were cpython-311; the reviewed target-only loader bytes matched the branch; native availability, required DSA symbols, and a live MLX-array ABI probe passed under the bundle’s own embedded runtime.
- The exact 44-shard / 101.55 GB estimated target-only artifact loaded through the packaged CLI in 20.5 s in the first run and 9.3 s from the controlled profile. Runtime logs explicitly reported native dsa_indexer_scores/dsa_topk_indices availability.
- The fixed prompt What is 2+2? Answer with one digit. returned final content 4 with reasoning separated on every request.
- Unload reclaimed 96.71 GB and settled MLX active memory to about 1 KB; the isolated server then shut down cleanly.

Performance boundary: an unrelated sustained local disk/text search contaminated these packaged-app timing samples, which varied materially, so they are functional diagnostics only and do not replace the earlier cleaner 11.7 tok/s native diagnostic already reported.

The packaging exercise exposed and now closes an adjacent release defect: build-time ABI probing could pass under Python 3.13 and then package cpython-313 extensions into a CPython 3.11 app. Commit a34b0408 adds fail-closed prebuild interpreter-ABI matching plus post-stage extension-suffix admission. Independent frozen review passed; focused checks were 6/6 plus shell syntax, Ruff, py_compile, and diff checks.

PhilipJohnBasile:
CI follow-up: Python 3.11 and 3.13 passed the full suite (9,258 passed each). Python 3.12 had one failure in the pre-existing `tests/test_prefix_cache_gdn_split.py::test_split_restore_walks_back_from_structurally_invalid_sidecar`; this PR does not modify that cache subsystem. I recreated a fresh Python 3.12 environment with the CI dependencies: the exact test passed 20/20 consecutive runs, and the full test file passed 6/6. This looks order-dependent/flaky rather than caused by the DeepSeek V4 or app-ABI changes. Fork contributors cannot rerun the upstream Actions job (`Must have admin rights`). Could a maintainer please rerun the failed Python 3.12 job? I am intentionally not adding an unrelated cache change to this PR.

jundot:
Thanks for the thorough implementation and validation.

After reviewing the exact pinned artifact's own model card, I don't think oMLX should carry a mainline compatibility path tied to this single config hash. The artifact is already marked as a superseded experimental reference, its card recommends the DwarfStar build, and its published evaluation shows material arithmetic and reasoning regressions.

The memory-pressure problem addressed here—sacrificing the hot prefix cache before giving a large reclaimable MLX buffer pool a chance to drain—has since been fixed on main by #2635. That implementation requests a synchronized pool reclaim first and defers hot-cache shrinking, model eviction, and request abort only when a scheduler accepts the request. Emergency pressure, disabled reclaim, a small pool, a missing scheduler, and request failures still fall through to immediate enforcement. Please drop the pressure-reclaim changes when rebasing.

Several other changes in this PR may still be independently useful, but they solve separate project-wide problems and should not be coupled to support for this artifact:

- The app ABI guard prevents custom kernels built with one CPython ABI from being packaged into an app embedding another ABI, and verifies the staged extension suffixes before release packaging.
- The thinking-control normalization preserves a top-level `enable_thinking` value, rejects conflicts with `chat_template_kwargs`, and prevents a zero thinking budget from enabling thinking implicitly.
- The nested sub-4-bit detection avoids selecting the ratio-128 native attention path when a DeepSeek V4 quantization configuration contains nested 2-bit or 3-bit overrides.

Please submit those changes as focused PRs so I can evaluate their runtime impact and regression scope independently.

Since the artifact-specific loader is not something I want to carry in main and the remaining changes belong in separate PRs, I'm going to close this PR.

Issue #1848, opened 2026-06-12
Model switch in chat settings behaves unexpectedly


KEO-RSP:
Three observations:

1. It lists MarkItDown as a model. Why?
2. It lists a model for which the download was cancelled and removed from the list. Why?
3. Switching to a different model while at least one chat message has been sent becomes effective only after one more message has been sent.

How I got there: I selected the not-downloaded model, sent a message, and got an error. I switched to the downloaded model, sent a message, and still got an error. This confused me. I then sent a third chat message and for that message the model I had previously switched to started doing its job. To describe this in more detail: let's say we have the models "MarkItDown", "NotDownloaded" and "Downloaded". Send a chat message to model "NotDownloaded" and get an error message. Switch to model "Downloaded", send a message, get the same error as before. Send another message and get a reply from model "Downloaded".

Expected:
- switching the model becomes effective immediately (for the next chat message that gets sent)
- not downloaded (canceled and removed) models don't appear in the model dropdown
- MarkItDown is not listed as a model (I assume this is currently the case for integration/convenience reasons)

Regarding the model for which the download was canceled, I dug a little deeper and noticed that the parts that had already been downloaded were not removed from the file system. Therefore, oMLX still considered this model to be present. I was able to fully remove it through the model manager. Afterwards it no longer appeared in the model dropdown for chats.

Environment:
 - macOS 26.5.1
 - oMLX 0.4.3

aeilot:
`MarkItDown` is not an AI model. It's a file-to-Markdown conversion pipeline built into omlx, which uses Microsoft's Python `markitdown` library to extract text from PDF, DOCX, PPTX, TXT, and MD files.

It's been registered as a virtual model with id `"MarkItDown"` (alias `"markitdown"`) and converts the attached files to Markdown and returns them as the assistant response.

The code's in omlx/api/markitdown.py and server.py.

```py
    if _markitdown_is_visible() and not any(
        m.id == MARKITDOWN_MODEL_ID for m in models
    ):
        models.append(ModelInfo(id=MARKITDOWN_MODEL_ID, owned_by="omlx"))
```

And you can enable this in oMLX's Integration Settings.

<img width="1326" height="741" alt="Image" src="https://github.com/user-attachments/assets/9f173b3c-3e56-41a7-a25f-916db041c3d0" />

PR #2643, opened 2026-08-14
feat(vlm): sync Muse Glimmer vendor with mlx-vlm #1848 numerics

nazerim:
## Summary

Sync the vendored Muse Glimmer model package with upstream mlx-vlm **#1848** ("Use model reasoning config and response templates"), keeping oMLX's deltas. The vendored copy (from #2586) predates the numerics/reasoning work that landed in mlx-vlm on 2026-08-12.

## Changes

- **`language.py`**
  - `_prepare_mlp_input` / `_finish_mlp` compiled norm fusions (perf + parity with upstream).
  - `qk_scale_factor` now applied in FP32 before casting back to the activation dtype (matches Transformers numerics — BF16 decode choices).
  - `embed_tokens` is now a plain `nn.Embedding` with a separate weightless `embed_norm` module (upstream design from #1848), replacing the PR #1839 `NormedEmbedding` wrapper. The norm now survives quantization by construction, so the `quant_predicate` exclusion is dropped.
  - Kept the oMLX delta: `initialize_rope` is imported from `mlx_lm.models.rope_utils` (the pinned mlx_vlm 0.6.3 rope_utils has no `initialize_rope`), so upstream's `implementation="eager"` is not adoptable here.
- **`muse_glimmer.py`**: `get_input_embeddings` applies `embed_norm` explicitly; the `encode_image` alias for the vision feature cache is kept.
- **`config.py`**: `thinking_start_token` / `thinking_end_token` fields.
- **`processing_muse_glimmer.py`**: `config` / `ModelConfig` wiring.
- **tests**: `test_quantization_preserves_embedding_norm` rewritten for the new module structure (quantizing `embed_tokens` must leave the separate `embed_norm` intact).

## Notes

- The FP32 `qk_scale` change is mirrored in the dflash-mlx fork (`jundot/dflash-mlx` PR) so the text-only module stays bit-exact with this vendored port (guarded by `tests/test_dflash_muse_glimmer.py::test_logits_match_bit_exact`).
- Verified: compat suite passes (20 tests, incl. the bit-exact parity guard); full oMLX suite identical to baseline (pre-existing environment failures only); live serving of Muse-Glimmer-30B-oQ6 text + vision confirmed after restart.

### DIFF
--- a/omlx/patches/mlx_vlm_muse_glimmer_compat/vendor/mlx_vlm/models/muse_glimmer/README.md
+++ b/omlx/patches/mlx_vlm_muse_glimmer_compat/vendor/mlx_vlm/models/muse_glimmer/README.md
@@ -2,27 +2,35 @@
 
 Source: Blaizzy/mlx-vlm PR #1838 head (model package + prompt_utils
 registration; includes the CenteredRMSNorm FP32 operation-order fix from
-commits edfb0ef1 + 6242d295), plus the quantization fix from PR #1839.
-Vendored because both PRs are newer than oMLX's mlx-vlm pin (`78b96eb`).
+commits edfb0ef1 + 6242d295), plus upstream #1848 (reasoning config,
+`_prepare_mlp_input`/`_finish_mlp` norm fusions, FP32 `qk_scale_factor`
+math, plain `nn.Embedding` + separate weightless `embed_norm`) synced in
+2026-08-16 (vendor commit 3a82f856). Vendored because these PRs are newer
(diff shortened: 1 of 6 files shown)


jundot:
Thanks for syncing the vendored Muse Glimmer implementation with upstream. The FP32 query scaling fixes the BF16 numerical mismatch, and the separate embedding norm keeps quantization safe with a simpler module structure. The Muse Q4 prefill override and the pinned dflash-mlx implementation still mirror the previous scaling order, but I'll handle both as follow-up fixes. LGTM from the core vendor-sync side.

PR #2644, opened 2026-08-14
feat(cache): add storage codecs and policy for persisted GDN sidecars

hojin12312:
# storage codecs and policy for persisted GDN sidecars

## Summary

GDN (Gated DeltaNet) recurrent state — the `state_1` tensors of hybrid attention/GDN layers on GDN checkpoints such as the Qwen3.5/3.6 family — is persisted at cache boundaries as sidecars. This change adds storage codecs and an explicit storage policy for those sidecars:

- `rht_int16` — randomized Hadamard transform (RHT) + row-wise symmetric INT16, **new default**
- `fp32` — exact, opt-in
- `bf16`
- `int8` — row-wise symmetric INT8
- `rht_int8` — deterministic RHT + row-wise symmetric INT8

Codec names are the settings strings; the appendices use short forms (`int8`, `rht_int8`).

The storage policy is `auto | ssd_sidecar | embedded`. `auto` uses an SSD sidecar when the cache is enabled and not hot-only; otherwise the recurrent state stays embedded with the main cache payload. The legacy split boolean remains accepted and existing explicit settings are preserved. **Migration:** a deployment with legacy `gdn_ssd_split_enabled` set and no explicit dtype is read as FP32 and preserved — existing deployments are not silently switched to the new RHT-INT16 default.

Live recurrent state remains FP32. The codec is applied only at the sidecar serialization boundary, and restored tensors are converted back to FP32. `relative L2` throughout is element-wise ‖q−f‖₂/‖f‖₂ per tensor, averaged.

## Design and safety

- Codec-specific sidecar namespaces do not invalidate or repartition main KV blocks.
- A reduced codec restore falls back to a legacy FP32 sidecar at the same endpoint when no sidecar for the requested codec exists (e.g., reading a pre-existing FP32 file after the dtype changed); the requested dtype, effective codec, fallback decision and dequantization count are exposed.
- Unknown/future codecs, malformed scale/shape/dtype metadata, non-finite values and invalid RHT metadata fail closed.
- RHT signs are deterministic and do not consume model-generation RNG state.
- GDN sidecars remain part of the shared SSD budget, and lifecycle hardening preserves the previous durable file on failed replacement.
- Embedded mode leaves the codec dormant; it follows the main KV payload tier and is not presented as a separate RAM-only GDN tier.
- Admin cache changes update the engine-pool scheduler template and unload loaded models so the selected mode/codec is used on the next load.

## Validation

- 510 focused codec/restore/settings/config/Admin/UI/scheduler/prefix-cache/observability tests passed; 855 passed with the latest distributed prompt-snapshot cache paths included.
- Full suite on the current base (`aef5a0cf`, rebased 2026-08-14; see thread comment for details): 9,414 passed, 105 skipped, 76 deselected. The 7 failures require optional packages absent from the test environment (`dflash_mlx`: 4 Muse Glimmer tests; `ddgs`: 3 web-search tests); 0 changed-path failures. (Original `v0.6.0.dev1` run: 9,429 passed.)
- `py_compile` and `git diff --check` passed; Ruff F/E4/E7/E9 reported only four pre-existing upstream findings.

All measurements below are local validation on one Qwen3.6-27B configuration, not a universal quality or performance guarantee. The measurement harnesses (`gdn_quality_ab.py` matrix, RHT/TQ probes) and raw artifacts live in the fork and can be shared on request.

## Review notes

- The PR is rebased onto `origin/main` at `aef5a0cf` (2026-08-14, zero conflicts, content preserved), including the distributed boundary snapshot cache changes and the tip-strip/dedup rework (`08aee5e5`).
- Existing upstream FP32 sidecar support from #2569 (and subsequent signature canonicalization) is treated as the base and not duplicated.
- Generated quality-result artifacts remain excluded. The Admin UI includes storage-mode and codec choices with the measured trade-offs.
- Local commit split:
  1. `feat(cache): add GDN sidecar state codecs`
  2. `feat(cache): add RHT-INT16 GDN sidecar codec`
  3. `feat(cache): add GDN snapshot storage policy`

---

## Appendix A: full precision sweep (Qwen3.6-27B, 64K context, 32 checkpoints)

Round-trip relative L2 vs FP32 across 1,207,959,552 elements (32 checkpoints, 2K–64K at 2K steps, 48 recurrent tensors each).

| codec | bytes/checkpoint | relative L2 vs FP32 |
|---|---:|---:|
| FP32 | 153,961,616 | 0 |
| RHT-INT16 | 79,656,224 | 0.002477% |
| INT16 | — | 0.003841% |
| FP16 | — | 0.020507% |
| BF16 | 78,468,008 | 0.164848% |
| RHT-INT8 | 41,907,360 | 0.639229% |
| INT8 | 41,904,336 | 0.989476% |
| FP8 E4M3FN | — | 2.817580% |
| FP8 E5M2 | — | 5.398099% |

RHT reduces relative L2 by ~35% over the plain equivalent at matched bit-widths (INT16: 0.003841% → 0.002477%; INT8: 0.989476% → 0.639229%). RHT-INT16 is 8.28× more accurate than FP16 at comparable storage size (1.93× compression vs FP32). INT16, FP16, and FP8 are offline simulations evaluated for comparison; the implemented codecs are FP32/BF16/INT8/RHT-INT8/RHT-INT16.

## Appendix B: warm-restore output comparison vs FP32 (Phase A, MTP-off)

Each condition primes the cache cold with FP32, then runs a warm restore with the target codec and compares the full response text against the FP32 warm baseline, under deterministic sampling (temperature 0, top_p 1, fixed seed). "Exact" means byte-identical; "DIFF" means at least one character diverged. Each context runs 6 cases × 3 target codecs plus the FP32 baseline pair = 24 conditions; "0 invalid" means every run produced a valid completion artifact under the harness rules. All needle conditions (`niah` = needle-in-a-haystack retrieval: start/middle/end) were task_ok=True in every context length regardless of codec — divergences only appear in preamble wording, not in the retrieved answer.

**8K context** (24 conditions, 0 invalid)

| case | bf16 | int8 | rht_int8 |
|---|---|---|---|
| general-code | exact | exact | exact |
| general-en | exact | exact | exact |
| general-ko | exact | exact | exact |
| niah-end | DIFF (char 139) | DIFF (char 139) | DIFF (char 139) |
| niah-middle | exact | exact | exact |
| niah-start | exact | exact | exact |

**64K context** (24 conditions, 0 invalid)

| case | bf16 | int8 | rht_int8 |
|---|---|---|---|
| general-code | DIFF (char 414) | DIFF (char 414) | exact |
| general-en | exact | exact | exact |
| general-ko | exact | exact | exact |
| niah-end | exact | exact | exact |
| niah-middle | exact | exact | exact |
| niah-start | exact | exact | exact |

general-code divergence is a single-word difference ("question" vs "prompt") in the preamble; rht_int8 was exact here.

**128K context** (24 conditions, 0 invalid)

| case | bf16 | int8 | rht_int8 |
|---|---|---|---|
| general-code | exact | exact | exact |
| general-en | DIFF (char 195) | exact | DIFF (char 195) |
| general-ko | exact | exact | exact |
| niah-end | exact | exact | exact |
| niah-middle | exact | exact | exact |
| niah-start | exact | exact | exact |

No codec was consistently inferior across all contexts; divergences fall on different cases at different lengths.

RHT-INT16 was not included in Phase A (it was implemented after); its retrieval output fidelity at these context lengths is covered by Appendix C.

## Appendix C: RHT-INT16 retrieval gate (MTP-off and MTP-on)

Cold + warm restore with FP32 and RHT-INT16 on `niah-middle`. All runs used `max_tokens=128` to allow the model to complete its answer past the thinking preamble, with deterministic sampling (temperature 0, top_p 1, fixed seed). The "text hash vs FP32" columns are exact when the FP32 and RHT-INT16 outputs hash identically for that run type (cold or warm).

| ctx | MTP | FP32 valid | RHT-INT16 valid | task_ok | text hash vs FP32 (cold) | text hash vs FP32 (warm) |
|---|---|---|---|---|---|---|
| 8K | off | ✓ | ✓ | ✓ | exact | exact |
| 64K | off | ✓ | ✓ | ✓ | exact | exact |
| 128K | off | ✓ | ✓ | ✓ | exact | exact |
| 8K | on | ✓ | ✓ | ✓ | exact | exact |

8/8 validity checks passed (4 context/MTP rows × 2 codecs), with task_ok 8/8. FP32 and RHT-INT16 text hashes were exact for both the cold and warm run at each condition, with 48 dequantizations and zero fallbacks per restore.

**256K process-restart SSD restore** (general-code, MTP-off): the server is restarted with a fresh cache, the prepared prompt is served cold from disk, and restore is verified against the SSD sidecars.

| codec | cold request time | GDN footprint (128 checkpoints) | KV SSD blocks loaded (cold) | errors |
|---|---:|---:|---:|---:|
| FP32 | 19.120 s | 18.35 GiB | 48 | 0 |
| RHT-INT16 | 19.455 s | 9.50 GiB | 48 | 0 |

The 0.335 s cold difference (1.75%) is within the observed run-to-run spread (warm requests ranged 15.96–17.24 s across codecs). GDN footprint is reduced by 1.93× with no restore errors.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -252,6 +252,7 @@ class GlobalSettingsRequest(BaseModel):
     ssd_cache_dir: str | None = None
     ssd_cache_max_size: str | None = None
     hot_cache_only: bool | None = None
+    gdn_snapshot_storage: str | None = None
     gdn_ssd_split_enabled: bool | None = None
     gdn_ssd_pending_max_size: str | None = None
     gdn_sidecar_state_dtype: str | None = None
@@ -931,6 +932,22 @@ async def _apply_cache_settings_runtime(
         return False, "Engine pool not initialized"
 
     pool = _server_state.engine_pool
+
+    # These settings all affect objects constructed with each scheduler. Keep
+    # the pool template synchronized before unloading existing engines.
+    pool._scheduler_config.hot_cache_only = global_settings.cache.hot_cache_only
+    pool._scheduler_config.gdn_ssd_split_enabled = (
+        global_settings.cache.get_gdn_ssd_split_enabled()
+    )
+    pool._scheduler_config.gdn_ssd_pending_max_bytes = parse_size(
+        global_settings.cache.gdn_ssd_pending_max_size
+    )
+    pool._scheduler_config.gdn_sidecar_state_dtype = (
+        global_settings.cache.gdn_sidecar_state_dtype
+    )
+    pool._scheduler_config.initial_cache_blocks = (
+        global_settings.cache.initial_cache_blocks
+    )
 
     # Update scheduler config based on cache settings
     if enabled is False or (enabled is None and not global_settings.cache.enabled):
@@ -3302,7 +3319,8 @@ async def get_global_settings(is_admin: bool = Depends(require_admin)):
                 )
             ),
             "hot_cache_only": global_settings.cache.hot_cache_only,
-            "gdn_ssd_split_enabled": global_settings.cache.gdn_ssd_split_enabled,
+            "gdn_snapshot_storage": global_settings.cache.get_gdn_snapshot_storage(),
+            "gdn_ssd_split_enabled": global_settings.cache.get_gdn_ssd_split_enabled(),
             "gdn_ssd_pending_max_size": global_settings.cache.gdn_ssd_pending_max_size,
             "gdn_sidecar_state_dtype": global_settings.cache.gdn_sidecar_state_dtype,
             "hot_cache_max_size": global_settings.cache.hot_cache_max_size,
@@ -3729,22 +3747,62 @@ async def update_global_settings(
             _parse_hot_cache_max_size(request.hot_cache_max_size)
         except ValueError as exc:
             raise HTTPException(status_code=400, detail=str(exc)) from exc
+    if request.initial_cache_blocks is not None and request.initial_cache_blocks <= 0:
+        raise HTTPException(
+            status_code=400,
+            detail="initial_cache_blocks must be positive",
+        )
 
     # GDN sidecar persistence requires the SSD tier. Validate the effective
     # values before mutating the live settings object so an invalid admin
     # update cannot leave an unsaved split/hot-only combination in memory.
-    current_gdn_split = getattr(global_settings.cache, "gdn_ssd_split_enabled", False)
-    current_hot_only = getattr(global_settings.cache, "hot_cache_only", False)
-    effective_gdn_split = (
-        request.gdn_ssd_split_enabled
-        if request.gdn_ssd_split_enabled is not None
-        else current_gdn_split is True
+    requested_storage = request.gdn_snapshot_storage
+    if requested_storage is not None:
+        requested_storage = requested_storage.strip().lower()
+        if requested_storage not in {"auto", "ssd", "ssd_sidecar", "hot", "embedded"}:
+            raise HTTPException(
+                status_code=400,
+                detail=(
+                    "gdn_snapshot_storage must be one of: "
+                    "auto, ssd_sidecar, embedded"
+                ),
+            )
+    if requested_storage is not None and request.gdn_ssd_split_enabled is not None:
+        raise HTTPException(
+            status_code=400,
+            detail=(
+                "gdn_snapshot_storage cannot be combined with the legacy "
+                "gdn_ssd_split_enabled field"
+            ),
+        )
+
+    current_gdn_raw = getattr(
+        global_settings.cache, "gdn_ssd_split_enabled", False
     )
+    current_hot_only = getattr(global_settings.cache, "hot_cache_only", False)
     effective_hot_only = (
         request.hot_cache_only
         if request.hot_cache_only is not None
         else current_hot_only is True
     )
+    current_cache_enabled = getattr(global_settings.cache, "enabled", True)
+    effective_cache_enabled = (
+        request.cache_enabled
+        if request.cache_enabled is not None
+        else current_cache_enabled is not False
+    )
+    if requested_storage == "auto":
+        effective_gdn_split = effective_cache_enabled and not effective_hot_only
+    elif requested_storage in {"ssd", "ssd_sidecar"}:
+        effective_gdn_split = True
+    elif requested_storage in {"hot", "embedded"}:
+        effective_gdn_split = False
+    elif request.gdn_ssd_split_enabled is not None:
+        effective_gdn_split = request.gdn_ssd_split_enabled
+    elif current_gdn_raw is None:
+        effective_gdn_split = effective_cache_enabled and not effective_hot_only
+    else:
+        effective_gdn_split = current_gdn_raw is True
     if effective_gdn_split and effective_hot_only:
         raise HTTPException(
             status_code=400,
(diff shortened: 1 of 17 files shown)


hojin12312:
**Why not TurboQuant for GDN sidecars?** Since GDN state lives in the same cache subsystem as attention KV, this is a fair question — and we measured it directly.

Round-trip relative L2 on a real Qwen3.6-27B GDN sidecar (48 recurrent tensors; newest probe):

| codec | relative L2 |
|---|---:|
| TQ8 | 1.156% |
| RHT-INT8 | 0.610% |
| TQ16 | 0.0368% (pooled 0.0253%) |
| RHT-INT16 | 0.0025% |

1. **Measured: TQ provides no consistent advantage over row-wise codecs on GDN state.** Across two probes — the session-116 707-checkpoint probe (TQ8 1.15% vs plain INT8 1.15%) and the newest 48-tensor probe above (TQ8 1.156% vs plain INT8 1.243%) — TQ8 is at best tied with plain row-wise INT8, and RHT-INT8 (0.610%) beats both by ~2×. GDN recurrent state is heavy-tailed (kurtosis ≥ 735, crest factor ≥ 86) in a way per-(head, k-channel) row-wise scaling already absorbs. TQ16 (0.037% mean) is bounded below by its structure — the codec stores per-vector norms as FP16 at every bit width, and its measured pooled floor is ~0.025% — an order of magnitude above RHT-INT16 (0.0025%; full sweep 0.002477%).

2. **Architectural: TQ is a compute-path codec; the sidecar boundary is storage-only.** TQ's value is compressed-state direct attention (packed indices + codebook consumed inside Metal decode kernels, K/V never materialized). GDN sidecars are serialized at cache boundaries and dequantized back to FP32 before the next recurrence — no attention kernel touches this path, so TQ's fused-kernel advantage does not apply. We only need a faithful round-trip codec (llama.cpp's uniform block quantization is the closest precedent).

3. **Precision need: GDN state is recurrent — errors compound across the recurrence**, which is why RHT-INT16's 0.0025% was used to satisfy byte-identical warm-restore output at 8K–128K (Appendix C retrieval recheck, `max_tokens=128`), plus a 256K process-restart restore-success gate (Appendix C; that gate proves restore success, footprint and latency — with `max_tokens=32` probes, some 8K/64K/256K runs diverged in reasoning preamble wording, so we do not claim byte-identity there). TQ4-class fidelity was also not pursued: TQ4 already measures 9.6% on attention KV.

Happy to share the full raw measurement data if useful.

hojin12312:
**Rebased onto latest upstream main (`aef5a0cf`) — content unchanged**

Upstream advanced 3 commits past `ac32ea24` (`fcc201e2` deepseek_v4, `08aee5e5` prefix_cache tip-strip lineage + dedup backfill, `aef5a0cf` scheduler 64-token grid). `08aee5e5` rewrites `omlx/cache/prefix_cache.py` (+359/−70), which this PR also touches, so the branch was rebased:

- Rebase completed **conflict-free**; no manual resolution needed.
- PR content preserved: 15/17 files byte-identical vs the previous base; the only differing files are `prefix_cache.py`/`scheduler.py`, which upstream also modified (all 262 lines this PR adds to those two files survive verbatim; per-file add/del counts match exactly).
- **Full-suite revalidation** (`pytest tests/ -m "not slow and not integration"`, `env -u CLAUDE_CODE_AUTO_COMPACT_WINDOW`): **9,414 passed, 105 skipped, 76 deselected, 7 failed** — failures are exactly the known optional-dependency set (4× `dflash_mlx` Muse Glimmer parity, 3× `ddgs` web search); 0 changed-path failures. Upstream's new `test_prefix_cache_dedup_backfill`/`test_prefix_cache_rotating_tip_strip` pass.
- `git diff --check` clean; `py_compile` clean.

New head: `8fa85818` (`75e78d7b` codecs → `3a84cc88` RHT-INT16 → `8fa85818` storage policy).

jundot:
Thanks, this was a pleasure to review. The codec namespace isolation with the legacy fp32 fallback is a clean way to keep experiments from cross-reading each other's checkpoints, and the fail-closed coverage (mixed-writer metadata, the RHT overflow guard at encode) made the whole thing easy to trust. The exact-hit path re-prefilling just the last block instead of falling back to a full prefill is a nice win on top. Merged.

PR #2652, opened 2026-08-14
fix(oq): strip HF org prefix from quantized model names

monroewilliams:
Quantizing a model straight from the Hugging Face cache names the output after the route-safe cache ID (e.g. `Qwen--Qwen3.8-27B`), so the quantized model (e.g. `Qwen--Qwen3.8-27B-oQ8e-fp16-mtp`) is rejected by `huggingface_hub` on upload:

> Cannot have -- or .. in repo_id: 'user/Qwen--Qwen3.8-27B-oQ8e-mtp'.

`_source_model_names()` now derives the output base name from the canonical `org/repo` ID, using the bare repo name after the slash, so the example above becomes `Qwen3.8-27B-oQ8e-fp16-mtp`. The org only identifies where the source was downloaded from and doesn't belong in the quantized model's name. The display name in the picker is unchanged (still shows `Qwen/Qwen3.8-27B`), and the oQe imatrix cache filename follows the same base.

Follow-up to #2605 (HF cache discovery for quantization).

### DIFF
--- a/omlx/admin/oq_manager.py
+++ b/omlx/admin/oq_manager.py
@@ -144,8 +144,13 @@ def _source_model_names(source: Path) -> tuple[str, str]:
     if source.parent.name == "snapshots":
         decoded = _decode_hf_cache_model_id(source.parent.parent)
         if decoded is not None:
-            model_id, source_repo_id = decoded
-            return source_repo_id, model_id
+            _, source_repo_id = decoded
+            # Output base is the bare repo name from the canonical "org/repo"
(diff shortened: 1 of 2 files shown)


monroewilliams:
The user can work around this by renaming the quantized model's directory in ~/.omlx/models/ before upload, but that's a poor user experience.

jundot:
Thanks for the follow-up. Using the bare repo name fixes the invalid Hugging Face upload ID while preserving the canonical source ID for display, and the focused quantization and upload tests pass. LGTM, merging.

PR #2653, opened 2026-08-14
feat(qwen): support Qwen3.8 FP8 and reasoning levels

scaryrawr:
## Summary

- add blockwise FP8 dequantization for Qwen3.8-27B checkpoints through the existing Qwen3.5-family mlx-vlm patch path
- preserve embedded MTP tensors while sanitizing Qwen3.8 weights
- forward official reasoning effort controls through Chat Completions (`reasoning_effort`) and Responses (`reasoning.effort`) without overriding explicit `chat_template_kwargs`
- cover FP8 scale validation and API request compatibility with focused tests

## Validation

- `UV_PYTHON=python3.12 uv run --python python3.12 pytest tests/test_vlm_mtp.py tests/test_api_utils.py tests/test_openai_models.py tests/test_responses.py -q` (416 passed)
- loaded and generated successfully with local Qwen3.8-27B-FP8 and Qwen3.8-27B MXFP4 checkpoints
- verified Qwen3.8 `xhigh`, `medium`, and `low` reasoning templates through both Chat Completions and Responses APIs

## Credits

The Qwen3.8 FP8 loading work started from and adapts the approach in Blaizzy/mlx-vlm#1899 by @Lazarus-931: https://github.com/Blaizzy/mlx-vlm/pull/1899

Issue #2651, opened 2026-08-14
Lightning mtp non-functional gemma 4 on m5 0.6.0 dev1


thinklikesqlhns:
**Describe the bug**
When using jundot/gemma-4-31B-it-oQ4e-mtp and jundot/gemma-4-26B-A4B-it-oQ4e-mtp, no speedup is observed in benchmarks

**To Reproduce**
Download a gemma-4-mtp merged model.
benchmark with lightning mtp off.
Activate lightning mtp.
benchmark again.
No speedup is observed.

**Expected behavior**
I would expect a modest speed increase of 10-25% depending on the model and context length.
Instead, no observable speedup is discernible .

**Screenshots**

**Desktop (please complete the following information):**
 - macOS Version: 26.6.1
 - oMLX Version  0.6.0 dev1 (also reproduced on 0.5.8 dev3)

oMLX - LLM inference, optimized for your Mac
https://github.com/jundot/omlx
Benchmark Model: gemma-4-26B-A4B-it-oQ4e-mtp
Engine: Auto
Context: Code (Python)

Lightning mtp on

================================================================================

Single Request Results
--------------------------------------------------------------------------------
Test                                TTFT(ms)    TPOT(ms)        pp TPS        tg TPS      E2E(s)    Throughput    Peak Mem
pp1024/tg128                           479.1       14.48  2137.5 tok/s    69.6 tok/s       2.326   495.3 tok/s    14.44 GB
pp4096/tg128                          1988.4       15.01  2059.9 tok/s    67.2 tok/s       3.904  1081.9 tok/s    14.79 GB
pp8192/tg128                          4114.7       15.73  1990.9 tok/s    64.1 tok/s       6.135  1356.2 tok/s    14.98 GB

Lightning mtp off

================================================================================

Single Request Results
--------------------------------------------------------------------------------
Test                                TTFT(ms)    TPOT(ms)        pp TPS        tg TPS      E2E(s)    Throughput    Peak Mem
pp1024/tg128                           469.9       14.42  2179.1 tok/s    69.9 tok/s       2.310   498.7 tok/s    14.44 GB
pp4096/tg128                          1985.3       15.03  2063.2 tok/s    67.1 tok/s       3.913  1079.4 tok/s    14.79 GB
pp8192/tg128                          4122.2       15.82  1987.3 tok/s    63.7 tok/s       6.152  1352.3 tok/s    14.98 GB

oMLX - LLM inference, optimized for your Mac
[...]

PR #2654, opened 2026-08-14
fix(vlm): Lightning MTP weightless-checkpoint warning + acceptance telemetry in benchmark results

nlaratta:
## Summary

- Warn when `mtp_enabled: true` meets a checkpoint whose config declares MTP heads but ships no `mtp.*` tensors. The existing gates already fail closed; the change is purely that the operator sees "standard decode remains active" at WARNING instead of an "(active)" INFO followed by an attachment-skip INFO. Emitted at the weight-scan site so a runtime-patch failure cannot suppress it; the tensors-present path is pinned warning-free.
- Report Lightning MTP acceptance telemetry (`accept_pct`, `tok_per_cycle`, sequence/segment/cycle/accept/draft counts) in throughput benchmark results, so shared runs answer acceptance questions without server-log grepping (#2689). The summary is withheld with a warning when the drained sequence count doesn't match the point (cardinality guard — exact per-request attribution would need scheduler-uid plumbing, documented as a limitation at the summarizer). Upload payload unchanged.

## Validation

- 9,794 passed, 2 failed on the rebased branch (current main / 0.6.0). Both failures are `test_dflash_muse_glimmer.py::TestCrossImplementationParity`; a two-point bisect pins them to a719ba9c (parent passes 6/6, a719ba9c fails 2/6, same venv, M5 Max) — unrelated to this diff. The 23 added tests are mocked and behavior-level throughout.
- Live: admin benchmark pp1024/tg128 on Qwen3.8-27B-oQ8e → `mtp={accept_pct: 86.7, tok_per_cycle: 3.05}` in the point result.

## Credits

- Implementation developed with AI assistance (GPT-5.6 + Claude); design direction, verification, and benchmarks by the author.
- Related to #2653 (no file overlap; FP8 checkpoints with preserved MTP tensors benefit directly from this batched-engine wiring).
- Complements Blaizzy/mlx-vlm#1899 — even after it lands upstream, the oMLX batched engine still needs this scheduler/BatchGenerator integration to use the native head.

### DIFF
--- a/omlx/admin/benchmark.py
+++ b/omlx/admin/benchmark.py
@@ -641,6 +641,68 @@ async def _send_event(run: BenchmarkRun, event: dict) -> None:
         run.cond.notify_all()
 
 
+
+def _drain_mtp_snapshots() -> list[dict]:
+    """Pull finished-sequence Lightning MTP stats from the shared decode loop.
+
+    Fail-soft on purpose: telemetry must never break a benchmark, and the
+    MTP patch module may legitimately be absent for non-MTP models.
+    """
+    try:
+        from ..patches.mlx_lm_mtp.batch_generator import drain_mtp_stats_snapshots
+
+        return drain_mtp_stats_snapshots()
+    except Exception:
+        return []
+
+
+def _summarize_mtp(entries: list[dict], expected_sequences: int) -> dict | None:
+    """Aggregate drained snapshots into the acceptance summary for one point.
+
+    The registry is process-global and drained wholesale, so attribution
+    rests on the benchmark being the only traffic on the server. That
+    assumption is validated rather than trusted: the benchmark knows how
+    many sequences it ran for the point, and when the drained unique-uid
+    count disagrees the summary is withheld with a warning — visibly absent
+    beats silently contaminated. The guard is cardinality, not ownership:
+    concurrent traffic that happens to finish exactly the expected number
+    of MTP sequences while the benchmark's own produce none would still be
+    misattributed. Exact per-request attribution needs scheduler-uid
+    plumbing beyond this change's scope. A sequence can legitimately produce more
+    than one snapshot (late-join handoff logs a segment, then the rejoined
+    run logs another), so ``segments`` may exceed ``sequences``.
+
+    Rates are recomputed from summed counters rather than averaged
+    per-sequence so short sequences don't dominate.
+    """
+    if not entries:
+        return None
+    uids = {e.get("uid") for e in entries}
+    if len(uids) != expected_sequences:
+        logger.warning(
+            "MTP telemetry withheld for this point: drained %d sequence(s), "
+            "expected %d — was other traffic hitting the server during the "
+            "benchmark?",
+            len(uids),
+            expected_sequences,
+        )
+        return None
+    cycles = sum(e.get("cycles", 0) for e in entries)
+    accepts = sum(e.get("accepts", 0) for e in entries)
+    drafted = sum(e.get("drafted", 0) for e in entries)
+    tokens = sum(e.get("tokens", 0) for e in entries)
+    return {
+        "sequences": len(uids),
+        "segments": len(entries),
+        "tokens": tokens,
+        "cycles": cycles,
+        "accepts": accepts,
+        "drafted": drafted,
+        "accept_pct": round(accepts / drafted * 100, 1) if drafted else None,
+        "tok_per_cycle": round(tokens / cycles, 2) if cycles else None,
+    }
+
+
 async def _run_single_test(
     engine: Any,
     prompt: list[int],
@@ -659,6 +721,10 @@ async def _run_single_test(
         mx.reset_peak_memory()
     except Exception:
         pass
+
+    # Discard stats left over from priming or pre-benchmark traffic so the
+    # post-test drain attributes cleanly to this request.
+    _drain_mtp_snapshots()
 
     start_time = time.perf_counter()
     first_token_time = None
@@ -750,7 +816,7 @@ async def _run_single_test(
 
     generation_measured = generation_duration_s is not None
 
-    return _compute_single_metrics(
+    metrics = _compute_single_metrics(
         prompt_tokens=prompt_tokens,
         completion_tokens=metric_completion_tokens,
         start_time=start_time,
@@ -762,6 +828,10 @@ async def _run_single_test(
         generation_duration_s=generation_duration_s,
         generation_measured=generation_measured,
     )
+    mtp = _summarize_mtp(_drain_mtp_snapshots(), expected_sequences=1)
+    if mtp is not None:
+        metrics["mtp"] = mtp
+    return metrics
 
 
 async def _run_batch_test(
@@ -852,6 +922,7 @@ async def _run_batch_test(
             mx.reset_peak_memory()
         except Exception:
             pass
+    _drain_mtp_snapshots()  # discard pre-test stats for clean attribution
     wall_start = time.perf_counter()
     results = await asyncio.gather(
         *[_single_request(prompts[i]) for i in range(batch_size)]
@@ -880,7 +951,7 @@ async def _run_batch_test(
     gen_wall_time = wall_end - max_first_token
     tg_tps = total_gen_tokens / max(gen_wall_time, 1e-9)
 
-    return {
+    batch_metrics = {
         "pp_tps": round(pp_tps, 1),
         "tg_tps": round(tg_tps, 1),
         "avg_ttft_ms": round(avg_ttft_ms, 1),
@@ -889,6 +960,10 @@ async def _run_batch_test(
         "total_gen_tokens": total_gen_tokens,
         "batch_size": batch_size,
     }
+    mtp = _summarize_mtp(_drain_mtp_snapshots(), expected_sequences=batch_size)
+    if mtp is not None:
+        batch_metrics["mtp"] = mtp
+    return batch_metrics
 
 
 async def _run_external_single_test(
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -2670,6 +2670,21 @@ def _mtp_next(gen_batch: Any, state: _MtpState) -> Any:
     return _emit_response(gen_batch, token_id, logprobs_1d, state.stats)
 
 
+# Snapshots of finished sequences' Lightning MTP stats, drained by the admin
+# benchmark so acceptance telemetry can ride along with each measured point. Bounded: the
+# benchmark drains between points and rejects concurrent runs (409), so
+# attribution holds; anything beyond the bound is stale traffic nobody reads.
+_MTP_STATS_SNAPSHOTS: "deque[dict]" = deque(maxlen=64)
+
+
+def drain_mtp_stats_snapshots() -> "list[dict]":
+    """Return and clear the finished-sequence stats snapshots (oldest first)."""
+    out: list[dict] = []
+    while _MTP_STATS_SNAPSHOTS:
+        out.append(_MTP_STATS_SNAPSHOTS.popleft())
+    return out
+
+
 def _log_mtp_stats(uid: Any, stats: "_MtpStats", finish_reason: str) -> None:
     """Emit a one-line summary of MTP draft/verify activity for a finished sequence.
 
@@ -2721,6 +2736,29 @@ def _log_mtp_stats(uid: Any, stats: "_MtpStats", finish_reason: str) -> None:
         stats.sample_ms,
         stats.cache_ops_ms,
     )
+    if getattr(stats, "_snapshotted", False):
+        # A sequence that parks on its terminal token logs the SAME stats
+        # object twice (park exit, then finish). The INFO line repeating is
+        # harmless; a second snapshot would double every counter in the
+        # benchmark summary, so snapshot each stats object exactly once.
+        return
+    try:
+        stats._snapshotted = True
+        _MTP_STATS_SNAPSHOTS.append(
+            {
+                "uid": str(uid),
+                "finish": str(finish_reason),
+                "tokens": total_emits,
+                "cycles": stats.cycles,
+                "accepts": stats.accepts,
+                "drafted": total_drafted,
+                "zero_cycles": stats.zero_cycles,
+            }
+        )
+    except Exception:
+        # Telemetry must never take down a finishing sequence: the finish
+        # paths call this logger unguarded, so any snapshot failure stays here.
+        logger.debug("MTP stats snapshot dropped", exc_info=True)
 
 
 def _bump_emit_stat(state: _MtpState, source: str) -> None:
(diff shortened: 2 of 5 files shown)


jundot:
Thanks for putting this together. I could not reproduce a correctness or severe performance problem with native MTP on vision requests. I tested single-stream greedy generation with identical prompts and images on the PR base commit.

| Model / workload | Standard decode | Native MTP | Acceptance | Output |
|---|---:|---:|---:|---|
| Qwen3.5-9B / image description | 79–85 tok/s | 106–107 tok/s | 68.5% | Byte-identical |
| Qwen3.5-9B / chameleon details | 64.3 tok/s | 69.8 tok/s | 63.2% | Byte-identical |
| Qwen3.5-9B / table OCR | 8.7 tok/s | 9.0 tok/s | 89.7% | Byte-identical |
| Qwen3.6-27B / image description | 27–28 tok/s | 40–41 tok/s | 65.6% | Byte-identical |

No errors or tracebacks occurred. The existing prompt-priming implementation explicitly treats a missing priming context as supported unprimed behavior: draft quality may vary, but target verification preserves output identity.

| Head | Vision-prefill behavior | Effect of this PR |
|---|---|---|
| Qwen3.5/3.6 | Can start unprimed and recover as committed generation tokens accumulate | Disables a working and measurably faster path |
| Gemma 4 | Stateless; drafts from the backbone’s shared KV, which already contains the vision context | Disables a path that does not have the prompt-priming limitation |

Could you share a concrete checkpoint and request that reproduces the failure this gate is intended to address? Without such a repro, I think the blanket per-row vision block should be removed. If a specific head is affected, the restriction should be capability-based and limited to that implementation, with a regression test for the failing case.

nlaratta:
Confirmed on my side — you're right, and the block is gone. Tested on Qwen3.8-27B (oQ6 community quant, embedded MTP head): vision request outputs are byte-identical between native MTP and standard decode at temp 0, activation logs `primed=0`, and draft acceptance on the vision request was 71.0% — consistent with your 63–68% measurements. Also exercised the sampled path at the card's recommended settings (temp 1.0/top_p 0.95): coherent vision outputs, 65.9%/68.5% acceptance across seeds — consistent with your greedy measurements.

The gate originated as a conservative assumption about unprimed prefills, not an observed failure; since no repro exists, I removed the machinery entirely — the per-row check, the adapter bookkeeping, the scheduler classification, and the exclusion tests that covered them. If a specific head ever surfaces a real failure, a capability-scoped gate with a regression test for that concrete case would be the right shape, per your suggestion. Full fast suite passes.

taozhiyuai:
## M3 Max 128GB benchmark: Qwen3.8-27B-oQ4e-mtp with `mtp_enabled: true` — MTP does not appear to activate on 0.6.0rc1

### Environment
- **Machine:** Apple M3 Max, 40-core GPU, 128 GB unified memory, 400 GB/s bandwidth
- **oMLX:** 0.6.0rc1 (released 2026-08-15)
- **Model:** `Qwen3.8-27B-oQ4e-mtp` (scottlowry), 17.0 GB, 4-bit/GS64 enhanced oQ, 29 embedded MTP weights, `model_type=qwen3_5`, VLM architecture (`Qwen3_5ForConditionalGeneration`)
- **Per-model settings:** `mtp_enabled: true`, `turboquant_kv_enabled: true` (4-bit), `specprefill_enabled: false`, `dflash_enabled: false`
- **Scheduler:** `chunked_prefill: true`, `prefill_priority: speed`, `decode_fairness: true`, `max_concurrent_requests: 3`, `hot_cache: 10 GB`, `ssd_cache: 92 GB`

### Single-request results (code-mixed context)

| Prompt tokens | TTFT (ms) | pp TPS | tg TPS | TPOT (ms) | Peak Mem |
| 1,024 | 4,818 | 212.5 | 34.7 | 29.0 | 18.35 GB |
| 4,096 | 18,423 | 222.3 | 38.4 | 26.2 | 19.61 GB |
| 8,192 | 37,023 | 221.3 | 23.5 | 42.9 | 19.85 GB |
| 16,384 | 78,110 | 209.8 | 22.2 | 45.3 | 20.54 GB |
| 32,768 | 233,420 | 140.4 | 12.4 | 81.2 | 21.84 GB |

### Continuous batching (pp1024 / tg128)

| Batch | tg TPS | Speedup | pp TPS | TTFT (ms) |
|---:|---:|---:|---:|---:|
| 1× | 34.7 | 1.00× | 212.5 | 4,818 |
| 2× | 19.5 | 0.56× | 114.5 | 13,539 |
| 4× | 39.3 | 1.13× | 88.2 | 25,287 |
| 8× | 88.8 | 2.56× | 84.3 | 50,484 |

### Key observations

**1. MTP does not appear to be active despite `mtp_enabled: true` and 29 embedded MTP tensors present.**

The PR author's own validation on M5 Max 128GB with `metajazz/Qwen3.8-27B-oQ6-fp16-mtp` shows:
- MTP off: 23.0 tok/s → MTP on: **52.9 tok/s** (greedy, 97.8% acceptance)

On my M3 Max with the same settings (`mtp_enabled: true`), greedy tg TPS at 1k context is only **34.7 tok/s** — closer to what I'd expect for standard decode on this hardware than MTP-accelerated decode. For reference, MTPLX-Optimized-Speed (4-bit/GS32, no MTP head) on the same machine produces 42.8 tok/s at 1k context, which is consistent with a non-MTP baseline for this chip.

This strongly suggests that on 0.6.0rc1 (without this PR), the `mtp_enabled` flag is set but the `VLMBatchedEngine` falls through to standard autoregressive decode because the native-head MTP path for VLM checkpoints is not yet wired up. The fail-closed gating described in this PR's summary is working as designed — it silently degrades to standard decode.

**2. Prefill is excellent — oMLX native engine outperforms MTPLX by 10–28%.**

| Context | oMLX pp TPS | MTPLX-Optimized-Speed pp TPS | Delta |
|---:|---:|---:|---:|
| 1k | 212.5 | 192.8 | +10% |
| 4k | 222.3 | 199.2 | +12% |
| 8k | 221.3 | 196.2 | +13% |
| 16k | 221.3 | 164.5 | +28% |
| 32k | 140.4 | 115.8 | +21% |

**3. Batch 8 throughput reaches 88.8 tok/s (2.56× speedup) — decode fairness works very well.**

The `decode_fairness` feature (new in 0.6.0rc1) produces clean superlinear scaling at batch 4 (1.13×) and batch 8 (2.56×). For comparison, MTPLX's external engine on the same model produces only 1.62× at batch 8. The batch-2 regression (0.56×) is still present but less severe than MTPLX's 0.41×.

**4. Memory efficiency is excellent with TurboQuant KV 4-bit.**

Peak memory grows from 18.35 GB (1k context) to 21.84 GB (32k context) — only +3.5 GB for 32k tokens of KV cache. This leaves ample room for concurrent requests on 128 GB.

### What this PR would change on this hardware

If this PR enables native-head MTP for VLM checkpoints as described, and the PR author's 2.3× speedup (23→53 tok/s on M5 Max) holds proportionally on M3 Max, I would expect:
- **tg TPS at 1k context: ~34.7 → ~55–70 tok/s** (greedy/predictable content)
- **tg TPS at 4k context: ~38.4 → ~60–75 tok/s**

This would make oMLX with oQ4e-mtp the clear winner across all metrics on Apple Silicon, not just prefill and batching.

Happy to re-benchmark the moment this PR lands on a release build. Thanks for the great work on day-one Qwen3.8 support.

taozhiyuai:
## Follow-up: MTP state logs confirm prompt priming is active, but decode still collapses at 16K on the 8-bit variant

Adding to my earlier benchmark report — I checked the server logs and found the MTP batch generator reports a `primed` history length equal to the prompt length for every benchmark run:

```
MTP path activated for uid=4 (model has mtp_forward, batch=1, primed=1024)
MTP path activated for uid=4 (model has mtp_forward, batch=1, primed=4096)
MTP path activated for uid=4 (model has mtp_forward, batch=1, primed=8192)
MTP path activated for uid=4 (model has mtp_forward, batch=1, primed=16384)
MTP path activated for uid=4 (model has mtp_forward, batch=1, primed=32768)
```

This confirms the MTP engine is engaged and prompt priming is working (per `prompt_priming.py`, default `OMLX_MTP_PROMPT_PRIMING=1`). The "MTP does not appear to activate" conclusion in my earlier comment needs a **correction**: the MTP machinery IS running — the issue is that **decode throughput at 16K collapses** (4-5.5 tok/s, TPOT 183-250 ms) which is far below what MTP-assisted decode should produce.

Full details on the 16K collapse (reproducible twice) and the exact 16384 threshold found in `qwen35_q4_mlp.py` are in [#2689](https://github.com/jundot/omlx/issues/2689). The `primed=16384` log line means the 16K run is also the first run where MTP's primed history (one full-attention layer of KV over the prompt) reaches that size — worth checking whether the priming memory cost interacts badly with the 8-bit weight bandwidth at exactly this point.

Would it be possible to expose MTP acceptance rate / draft hit stats in the benchmark output? That would let us distinguish "MTP running but drafts rejected" from "MTP path not wired for VLM engine" definitively.

jundot:
Thanks for the quick follow-up and for validating the vision path. Removing the per-row block resolves the regression I raised, and the updated focused suite passes locally (`427 passed`).

After rechecking the current head, I am having trouble identifying the remaining runtime feature delta:

| Area | Behavior on the base commit / 0.6.0rc1 | Remaining PR change |
|---|---|---|
| Embedded-head VLM activation | The head is attached, the per-instance decode marker is set, and BatchGenerator enters Lightning MTP | No new dispatch or adapter bridge |
| Vision requests | Lightning MTP activates successfully and preserves output identity | The exclusion has now been removed |
| Missing head tensors | Existing attachment, head-presence, and decode-marker checks already fail closed | The same tensor scan is moved earlier |
| Readiness | Existing activation logs show `MTP path activated` and `primed=...` | An additional readiness helper and log; its return value does not control decode |

The follow-up benchmark comment also confirms that Lightning MTP was already active on 0.6.0rc1, so it does not provide a baseline failure that this PR fixes. Could you share a specific checkpoint, request, and before/after log where the base commit fails to enter Lightning MTP but this branch succeeds?

If there is no such repro, I think the cleanest option is to close this PR. Alternatively, it could be substantially reduced and rescaled as a small test/observability change: remove the duplicated eligibility helper and repeated checkpoint scan, keep only behavior-level regression coverage or a genuinely useful warning, and use the official “Lightning MTP” name consistently in the title, commit, code comments, logs, and tests.

nlaratta:
I don't have such a repro — I went looking for one and found the opposite. A stock-0.6.0rc1 run on this M5 Max independently confirms that the VLM Lightning MTP path activates for a VLM-packaged Qwen3.8-27B oQ6 checkpoint, vision intact, consistent with your table (`MTP path activated for uid=0 … accept=119/132 (90.2%) … tok/cycle=3.29`). My original framing came from an older 0.5.7-based setup and, in hindsight, reflected that setup rather than missing wiring — the four `mlx_vlm_mtp/*_runtime.py` files are byte-identical from v0.5.7 through rc1, and #2651 fixed forced-LM benchmark routing for Lightning-MTP VLM checkpoints.

I've taken your second option and rescaled the branch (force-pushed) to two commits:

1. `fix(vlm): warn when Lightning MTP is requested but the checkpoint ships no head tensors` — no new helper, no second scan, no change to dispatch. When `mtp_enabled: true` meets a checkpoint whose config declares heads without the tensors, a WARNING at the weight-scan site states that standard decode remains active — today the load log opens with "Speculative backend selected … (active)" and reads as Lightning MTP running on head-stripped quants. The warning is emitted at the scan, not under the runtime patch's success branch, so a patch failure can't suppress it. Intentional `mtp_enabled=False` loads keep their INFO/DEBUG lines, and the positive path (tensors present) is pinned warning-free. Behavior-level tests sit beside the existing #1426 coverage in `tests/test_model_loading.py`.

2. `feat(bench): report Lightning MTP acceptance telemetry in benchmark results` — motivated directly by #2689, where distinguishing "drafts rejected" from "not engaged" required asking the reporter to grep server logs. Per-sequence stats are snapshotted at finish (fail-soft) into a bounded registry; the benchmark drains it around each point and attaches an `mtp` block: `sequences, segments, tokens, cycles, accepts, drafted, accept_pct, tok_per_cycle`. Attribution is validated by cardinality: the summary is withheld with a warning if the drained unique-sequence count differs from what the point ran, which catches concurrent traffic in every case except an exact-count coincidence — exact per-request attribution would need scheduler-uid plumbing I judged out of scope here; the limitation is documented at the summarizer, and I'm happy to add the plumbing in a follow-up if you want it. A stats object snapshots exactly once, so park-on-terminal sequences can't double-count. Rates are recomputed from summed counters. Non-MTP runs and the upload payload are unchanged. Verified live: pp1024/tg128 on oQ8e produced `mtp={accept_pct: 86.7, tok_per_cycle: 3.05}` beside `gen_tps: 39.7`.

The eligibility helper, early scan, readiness helper, and their tests are gone; title, commits, logs, and tests use the Lightning MTP name. One unrelated heads-up from validation: the two `test_dflash_muse_glimmer.py::TestCrossImplementationParity` tests fail on this M5 Max via a two-point bisect at a719ba9c ("align Muse Glimmer FP32 scaling paths" — parent passes 6/6, a719ba9c fails 2/6). Possibly chip-dependent numerics; happy to file separately with details.

If you'd rather take only one of the two commits — or neither — no hard feelings; closing was a fully fair option and I appreciate the careful reviews.

taozhiyuai:
Hi @jundot,

Let me address your request directly.

**Regarding the before/after log:**
I cannot provide a repro where the base commit fails to enter Lightning MTP but the PR branch succeeds. nlaratta's independent verification on M5 Max already confirmed that VLM Lightning MTP works on 0.6.0rc1 — his data shows `accept=119/132 (90.2%)`, `tok/cycle=3.29`. This aligns with your observation that the original PR framing came from an older 0.5.7 setup, not from 0.6.0rc1.

**My earlier misdiagnosis:**
When I said "MTP does not appear to activate" in my benchmark comment, it was because the benchmark results lacked MTP telemetry — I could not distinguish "MTP working normally" from "MTP not engaged." Here's what I have vs. what I'm missing:

| What I Have | What I'm Missing |
|-------------|------------------|
| `gen_tps`: 34.7 (1K context) | `accept_pct` |
| `processing_tps`: 212.5 | `tok_per_cycle` |
| `feature_flags`: `lightning_mtp` | `sequences`, `segments`, `cycles`, `accepts`, `drafted` |

This is exactly what your PR solves: the issue was never that MTP wasn't working — it was that the benchmark didn't expose MTP's operational status.

**My local configuration:**
- **Checkpoint:** `Qwen3.8-27B-oQ4e-mtp` (scottlowry), 17GB, 4-bit/GS64 enhanced oQ
- **Architecture:** `Qwen3_5ForConditionalGeneration` (VLM)
- **MTP tensors:** 29 keys (`language_model.mtp.*`)
- **Preprocessor:** Present (`preprocessor_config.json`, image support enabled)
- **Settings:** `mtp_enabled: true`, `turboquant_kv_enabled: true`

**Validation plan:**
Once PR #2654 is merged, I'll re-run benchmarks on the same hardware to verify:
1. `mtp` field appears in benchmark results
2. `accept_pct` / `tok_per_cycle` match expectations (~90%+ / ~3.0+ for greedy)
3. Warning triggers correctly when checkpoint lacks MTP tensors

Happy to run these tests as soon as a release build includes the PR. Thanks for the observability improvements — they're exactly what the community needed.

PR #2659, opened 2026-08-15
feat(model-loading): support Qwen3.8 mixed ModelOpt NVFP4

LKyaha:
## Summary

Add a narrowly gated direct-loading path for the published
[`unsloth/Qwen3.8-27B-NVFP4`](https://huggingface.co/unsloth/Qwen3.8-27B-NVFP4)
checkpoint.

That checkpoint is a dense 64-layer Qwen3.8 VLM using compressed-tensors
`mixed-precision` rather than a single quantization format:

- 168 MLP linear layers use packed E2M1 NVFP4 weights, E4M3 group scales, and
  an FP32 per-tensor divisor.
- 233 attention/GDN/LM-head/late-MLP linear layers use E4M3 FP8 weights with
  a BF16 per-output-channel scale.
- The vision tower and MTP tensors remain BF16.

The pinned mlx-vlm generic compressed-tensors path does not preserve this
two-level scaling layout. This change keeps the source packed codes and stored
scales unchanged, uses MLX's native NVFP4/MXFP8 carriers, and applies the
source-owned tensor/channel scale after the quantized matmul.

## Scope and safety

- Dispatch is limited to the exact Qwen3.8 architecture, 27B geometry, vision
  geometry, config-group targets, formats, and quantization parameters used by
  this checkpoint.
- Unknown compressed-tensors variants continue through the existing loader.
- Loading remains strict; no missing or unexpected tensors are accepted.
- The MTP subtree is explicitly excluded and stays BF16.
- This is an exact-weight A16 compatibility path. It does not enable the
  checkpoint's activation-quantization metadata and does not claim generic
  ModelOpt mixed-precision support.
- The text-only fallback is rejected because this checkpoint is a VLM.

## Validation

- `ruff format` and `ruff check` pass for the added files.
- 55 unit/loading regression tests pass.
- The original local `unsloth/Qwen3.8-27B-NVFP4` checkpoint strict-loads
  directly with no converted copy or external bridge.
- Real-model integration identifies all 401 custom quantized linear modules.
- Real-model text generation succeeds.
- Real-model vision generation correctly identifies a synthetic red-left,
  blue-right image.

Command used for the combined regression and opt-in real-model test:

```bash
OMLX_QWEN38_MODELOPT_MODEL_PATH=/absolute/path/to/Qwen3.8-27B-NVFP4 \
pytest tests/test_qwen38_modelopt_mixed.py \
  tests/test_model_loading.py \
  tests/integration/test_qwen38_modelopt_mixed_real_model.py \
  -m 'slow or not slow' -q
```

Result: `56 passed`.

### DIFF
--- a/omlx/patches/qwen38_modelopt_mixed.py
+++ b/omlx/patches/qwen38_modelopt_mixed.py
@@ -1,0 +1,416 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Exact-weight loader for Unsloth's mixed ModelOpt Qwen3.8 VLM.
+
+``unsloth/Qwen3.8-27B-NVFP4`` is a compressed-tensors checkpoint with two
+different language-layer formats:
+
+* per-channel E4M3 FP8 weights for attention, GDN, the LM head, and late MLPs;
+* packed E2M1 NVFP4 weights with E4M3 group scales and an FP32 tensor divisor
+  for the remaining MLPs.
+
+mlx-vlm's generic compressed-tensors path cannot represent the extra channel
+or tensor scale without either expanding FP8 weights to dense arrays or
+folding the FP32 NVFP4 divisor into E4M3 (which re-rounds the group scales).
+This module instead uses MLX's native MXFP8/NVFP4 carriers and applies the
+original scale after the matmul.  Packed weight codes and stored group/channel
+scales therefore remain unchanged.
+
+The dispatch predicate is intentionally strict and currently recognizes the
+published dense 64-layer Qwen3.8 VLM geometry and exact config-group targets.
+Unknown compressed-tensors combinations continue through the normal loader.
+Activations remain in the runtime dtype (A16); the checkpoint's activation
+quantization metadata is not enabled by this compatibility path.
+"""
+
+from __future__ import annotations
+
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for putting this together. I reviewed the loader path and confirmed that it preserves the original packed FP8/NVFP4 codes and source scales, keeps the vision and MTP tensors in BF16, and retains strict weight loading behind a narrowly scoped checkpoint gate. I also verified the published split of 168 NVFP4 and 233 FP8 modules, checked real weight samples against MLX dequantization, and ran the targeted loader and Qwen VLM/MTP regression suites successfully. Everything looks good from my side. Once you finish the remaining draft work and mark this ready for review, I'll merge it.

LKyaha:
Thanks for the thorough review and for independently validating the checkpoint. The remaining draft work is complete, and I've marked the PR ready for review.

jundot:
Thanks for finishing the remaining draft work. I rechecked the ready-for-review head, and the exact config gate, strict weight mapping, and source-scale preservation all look sound. The targeted loader and Qwen VLM/MTP regression tests pass, so this looks ready to land. LGTM, merging.

Issue #2662, opened 2026-08-15
Hot cache never populated for ArraysCache hybrid models (promotions=0, clear reports 0) — all prefix-reuse speedup from SSD, 20GB RAM tier dead


cropduster:
## Summary

On a Mac Studio M4 Max (128 GiB, macOS 27.0) with oMLX `0.6.0.dev1`, the in-memory
**hot cache is never populated** for an ArraysCache hybrid model, even though
`hot_cache_max_size` is set to 20 GB. A controlled A/B test shows that **100 % of the
prefix-reuse speedup comes from the SSD cache**, and the hot cache stays empty the
entire time (`hot_cache_promotions == 0`, `clear` reports 0 entries).

## Model / workload

- Model: `Qwen3.8-27B-oQ8e-mtp` (64 layers, ArraysCache hybrid, 4 KV heads, 24 Q heads,
  head_dim 256) — the log reports `Enlarging paged cache block_size=256 to 2048 for
  ArraysCache hybrid model`.
- Workload: one deterministic ~6000-token prompt sent 3× back-to-back with `max_tokens=64`.
- Config: `cache.hot_cache_only=false`, `cache.hot_cache_max_size="20GB"`,
  `cache.ssd_cache_max_size="55GB"`, `memory.prefill_memory_guard=false`.

## Reproduction

Start with both caches cleared, load the model, then send the *identical* prompt three times:

1. Request 1 (cold) — full prefill.
2. Request 2 (warm) — expected to hit cache.
3. `POST /admin/api/hot-cache/clear`, then Request 3 (warm, hot cache cleared).

Measured TTFT (time-to-first-token):

```
A1 cold:                2.688s
A2 warm:                0.021s
A3 warm after hot-clear: 0.019s    (clear => {"status":"ok","total_cleared":0,"bytes_reclaimed":0})
```

Cross-check with `hot_cache_only=true` (RAM-only, no SSD I/O) — identical pattern:

```
B1 cold:                2.818s
B2 warm:                0.022s
B3 warm after hot-clear: 0.020s
```

## What I expected

- Request 2 should promote the loaded block from SSD into the hot cache, so
  `hot_cache_promotions` should increment and `GET /admin/api/stats` should report
  `runtime_cache.hot_cache_entries > 0`.
- Clearing the hot cache before Request 3 should therefore force a re-read from SSD
  (or a full re-prefill in RAM-only mode), making A3/B3 *slower* than A2/B2.

## What actually happens

[...]

PR #2665, opened 2026-08-15
test(cache): cover ArraysCache hot-cache write-back

atirna:
## Summary
- Add a focused regression test for the mixed KVCache + ArraysCache write-back path.
- Verify the configured RAM hot cache retains and accounts for the hybrid block, including the split-GDN layout flag.

## Validation
- `.venv/bin/python -m pytest tests/test_paged_ssd_cache.py -q` (159 passed)

References #2662.


atirna:
closing this, the test doesnt cover what i described.

it never forces an ssd promotion, so the hot cache write-back path never gets hit, and `gdn_ssd_split_enabled=True` isnt really exercised either. #2662 is retracted too.

if the ArraysCache marker preservation is worth a regression i'll open it with a failing case first. sorry for the noise!

PR #2666, opened 2026-08-15
fix(admin): keep the main API key out of the admin dashboard URL

LXD-8:
Based on `main` @ `05784731` (v0.7.0.dev4).

## Why

The macOS menubar app opened `/admin/auto-login?key=<main API key>`, so the permanent admin credential reached browser history, the server access log, and the process list.

## What

The app POSTs the key to `POST /admin/api/auto-login-token` — request body, never a URL. The endpoint verifies it against the main key with the existing constant-time comparison and returns a 30-second signed token (`create_auto_login_token` / `verify_auto_login_token`, both salting their itsdangerous payloads, so a session token cannot be substituted for one). `GET /admin/auto-login` consumes `auth_token`; a missing, stale, or incorrect token redirects to the login page exactly as before.

| File | Why |
|---|---|
| `omlx/admin/auth.py` | Token helpers + `AUTO_LOGIN_MAX_AGE`. |
| `omlx/admin/routes.py` | Token endpoint; `auto_login` reads `auth_token`. |
| `apps/omlx-mac/Sources/Menubar/MenubarController.swift` | `fetchAutoLoginToken`; `webAdminURL(host:port:authToken:)` emits `auth_token`. |
| `apps/omlx-mac/Tests/oMLXTests/MenubarControllerPortTests.swift` | URL-builder tests updated; asserts `?key=` is absent. |
| `tests/test_admin_auth.py` | Token endpoint, token helpers, and HTTP-boundary tests. |

## Behaviour delta

Only the main key can be exchanged for a token; a configured sub key cannot. The dashboard otherwise opens as before — a failed exchange falls back to the login page, and no setup or UX requirement changes.

## Verification

- `pytest tests/ -q -k "admin or auth or login or api_key"` → 692 passed.
- `xcodebuild test -project oMLX.xcodeproj -scheme oMLX -destination 'platform=macOS' -only-testing:oMLXTests/MenubarControllerPortTests` → 30 passed.
- `TestAutoLoginOverHTTP` exercises the exchange across the ASGI boundary; the `auth_token` parameter name is pinned from both ends — renaming it in `routes.py` fails that test, renaming it in the Swift builder fails `testWebAdminURLUsesAutoLoginWithRedirect`.


jundot:
The macOS app sends `token`, while the server expects `auth_token`, so automatic login falls back to the login page. The current tests call the route directly and miss this mismatch across the HTTP boundary.

The cookie change breaks existing LAN HTTP login. Login returns success, but the Secure cookie is not sent on subsequent requests, leaving the user unauthenticated. This effectively requires TLS for a workflow that currently works over HTTP.

The fetch transport also stops honoring configured HTTP/HTTPS proxies and only attempts the first resolved IP. If that address is unreachable, the request fails even when another validated public address is available.

The localStorage change stops new writes but leaves existing users' stored keys behind. Keeping the key in page memory also does not prevent same-origin XSS from reading it, so the protection is narrower than the PR description suggests.

Since oMLX is primarily used on localhost, I'd like to narrow this PR to addressing concrete, severe security issues, such as external intrusion or RCE, while preserving the current UX and setup requirements. If you see a risk that warrants a different tradeoff, I'm happy to discuss it. Please share your thoughts!

LXD-8:
Narrowed, as you suggested — thanks for the close read.

The branch is now one change on `main` @ `76d19fed` (v0.7.0.dev3), 5 files. `POST /admin/api/auto-login-token` takes the main API key in the request body and returns a 30-second signed token; `/admin/auto-login` consumes `auth_token`. Gone: the `Secure` cookie, the `chat.html` localStorage change, and the transport that dropped proxy support and only tried the first resolved IP.

`token` vs `auth_token` was my bug — fixed. Your point about the tests was the right one too: they called the route function directly, so they never saw the query string at all. They now go through ASGI, and renaming the parameter on either side fails a test.

The one thing I kept is the key leaving the URL, and I'd rather say plainly why than assume it clears your bar. It is not a remote-exploit fix; it is a credential-leak fix. The value in `?key=` is the permanent main key, and it is a GET, so it lands in browser history on disk — written by a different application, and synced off the machine if the user has browser sync on. The server already treats the raw key as unfit for logs (`verify_api_key` logs `fingerprint_key(...)`, never the key), which is the same reasoning applied to a store we do not control. The fix costs no UX: same one-click, same setup.

It is fair to read that as below your bar for a localhost-first tool. If so, say the word and I will close this rather than argue the point.

LXD-8:
@jundot the 3 issues you raised have been fixed (token/auth_token, HTTP cookie, proxy transport). Could you confirm the fixes address your concerns? This is a security fix (API key in URL) so would appreciate a re-review when possible 🙏

PR #2669, opened 2026-08-15
fix(admin): accessibility improvements for the web dashboard (tablist/dialog/zoom) (0.7.0.dev3)

LXD-8:
Based on `main` @ `9c748580` (main's version string is now 0.7.0.dev3, not yet released; the latest release is still v0.7.0.dev2).

- Remove the viewport zoom lock (`maximum-scale=1.0`) that prevented page zooming on mobile.
- Main navigation: add `role=tablist`; each tab gets `role=tab`, an `id`, `aria-controls`, and dynamic `aria-selected`.
- Panels (`_status/_cluster_v2/_settings/_models/_logs/_bench`): add `role=tabpanel`, `id`, `aria-labelledby`.
- Modals (HF mirror, model settings): add `role=dialog`, `aria-modal`, `aria-labelledby`.
- Introduce `navbar.tablist_label` across all 9 catalogs — `en` carries the source string ("Dashboard sections"), the other eight are translated.
- Keyboard focus: nav tabs keep a soft gray focus ring (`[role="tab"]:focus-visible`) so tab focus on the main navigation is visible, sitting alongside the global 1px theme-color ring instead of the old default-blue outline.
- Toggle buttons: the two expose-as-API toggles in the model-settings modal (the new-profile form and the inline edit dialog) now bind `:aria-pressed` to the same flag that drives their ON/OFF label, so they are announced as toggles with a state rather than as ordinary buttons whose text happens to change. Both bindings are coerced to a real boolean, because the flag can arrive from the profiles API as a truthy non-boolean.

One unrelated one-liner rides along: `ja.json` and `ko.json` had an empty string for `settings.global.restart_notice`; the catalog normaliser step fills them with the English fallback ("Settings marked with"), matching what the other seven catalogs already carry.

## Re-verified against the branch's base
Confirmed on `9c748580`: `omlx/admin/templates/base.html:5` still sets `maximum-scale=1.0`; `_navbar.html` and all six panel/modal templates (`_status`, `_cluster_v2`, `_settings`, `_models`, `_logs`, `_bench`, `_modal_model_settings`) still have no `role="tablist"` / `role="tab"` / `role="tabpanel"` / `role="dialog"`; `navbar.tablist_label` is absent from all 9 locale files; the global `:focus-visible` outline in `base.html` does not target `[role="tab"]`; and the expose-as-API buttons carry no `aria-pressed`. All eight claims remain valid.

None of the templates, catalogs or `base.html` have been touched by any commit between `bd841fc3` and current `main` (`5941f683`), so these hunks apply unchanged.

---
Tests were run on `bd841fc3`: `tests/test_admin_a11y_changes.py` and `tests/test_cluster_dashboard.py` — 31 passed; the wider admin template suite (`tests/test_admin_*.py`, `tests/test_settings.py`, `tests/test_cluster_dashboard.py`, `tests/test_active_models_visibility.py`) — 875 passed. The panel hunk targets `_cluster_v2.html`, which replaced `_cluster.html` upstream, so the nav's `mainTab === 'cluster'` occurrence count moves 2 → 3.

## Rebase notes

Rebased onto `9c748580` after #3694 (`feat(admin): customizable dashboard block layout`) rewrote `_status.html` as a GridStack board of user-arranged blocks. Two conflicts, both resolved in favour of the newer structure:

- `base.html` — took this branch's viewport line (zoom lock removed) together with upstream's `{% block title %}{{ t('app.title') }}{% endblock %}`, which #3682 had already migrated into the catalog.
- `_status.html` — kept upstream's `dashboard_blocks` registry and moved this branch's `role="tabpanel" id="panel-status" aria-labelledby="tab-status"` onto the new panel wrapper.

The `{% include %}`-based panel layout this branch was first written against is gone, but the ARIA wiring is structure-independent: the tabs still live in `_navbar.html` and each panel is still a single wrapper element. No other hunk changes meaning on the new base.


jundot (approved):
I've added arrow-key navigation for the tabs and completed the dialogs' keyboard behavior. I verified focus containment, restoration, and nested dialogs in Chrome at desktop and mobile widths. LGTM with these fixes; I'll merge this after pushing them.

Issue #2522, opened 2026-08-04
Self-updater breaks the app bundle on swap — "damaged or incomplete" on every release


gardnerscot:
**Environment:** macOS 27.0 beta 4 (26A5388g), Apple Silicon, `release_candidate` update channel. Reproduced across the last three releases (0.5.4 → 0.5.7).

**Summary:** The in-app self-updater reliably breaks the installed app bundle. After an update, macOS reports the app is "damaged or incomplete" and it will not launch. Every release does this.

**What I observed:**
1. The download itself is fine — the staged update is a valid Notarized Developer ID build (`spctl --assess` accepts it, `codesign --verify` passes, seal intact).
2. The updater stages the new version at `/Applications/.oMLX-update.app` — valid and properly signed.
3. After the swap, the installed `/Applications/oMLX.app` is gutted: `Contents/` contains only `Info.plist`, `PkgInfo`, `Resources` — no `Contents/MacOS/`, no executable, no `_CodeSignature/`. `codesign` reports "code object is not signed at all."
4. Result: LaunchServices/Gatekeeper reports "damaged or incomplete" — which is literally true; the bundle is incomplete.

**Expected:** The swap atomically replaces the old bundle with the staged one (e.g., rename staged → live), preserving the notarized signature, so the app launches after the update.

**Actual:** The swap leaves a half-replaced bundle (missing `MacOS/` and `_CodeSignature/`), so the app cannot launch and reports "damaged or incomplete."

**Suggested direction:**
- Verify the staged bundle (codesign/spctl) before the swap, and swap atomically (rename over the old bundle) rather than writing into the live bundle in place.
- Keep a backup of the previous bundle until the new build launches once successfully.

**Workaround for users hitting this:** manually replace the app with the staged copy:

```
mv /Applications/oMLX.app ~/.Trash/
mv /Applications/.oMLX-update.app /Applications/oMLX.app
xattr -dr com.apple.quarantine /Applications/oMLX.app
```

22889h9nqs-create:
@gardnerscot 
yeap, same here, on 27 beta also, i am wondering if the package auto selected for the auto upgrade is not compatible, when i originally installed omlx i opted for the 26-27 dmg, the auto upgrade does not know to install the same omlx-updated-26-27.dmg ?? manual install of the update worked fine, i had backed up local user .omlx folder

jundot:
[...]

PR #2671, opened 2026-08-15
feat(ui): use system PingFang for Chinese (Noto fallback) and add focus ring (a11y)

LXD-8:
## Summary

Two small, self-contained UI changes to the admin dashboard shell: use the OS's native Chinese typeface (PingFang) on Apple platforms while keeping the bundled Noto/Inter fallback, plus accessibility improvements (a keyboard focus indicator and reduce-motion). No behavior or layout change beyond what is described below.

## 1. Use system PingFang for Chinese text, keeping Noto/Inter fallback

- `zh` (Simplified): `--font-sans` now reads `'PingFang SC', 'SF Pro Text', 'Noto Sans SC', 'Inter', system-ui, sans-serif`
- `zh-TW` (Traditional): `--font-sans` now reads `'PingFang TC', 'SF Pro Text', 'Noto Sans TC', 'Inter', system-ui, sans-serif`

**Rationale.**

- **Use the OS's native Chinese typeface.** On macOS/iOS, PingFang SC/TC is the system's built-in Chinese font. Leading the stack with it lets Chinese UI text render with the same typeface the OS itself uses, so the admin UI reads consistently with the rest of the system. Latin letters and digits are intentionally rendered by PingFang first (it includes them); `SF Pro Text` is kept immediately after as a fallback for glyphs PingFang does not provide, mirroring how the default font stack (`Noto ...`, `Inter`) pairs a CJK face with a Latin face.
- **Keep the bundled Noto/Inter faces as fallback.** The self-hosted `Noto Sans SC`/`Noto Sans TC` (and `Inter`) files are unchanged and still used when PingFang/SF Pro are not available.

**No regression for non-Apple platforms.**

- PingFang and SF Pro exist only on Apple platforms; on Windows/Linux the browser falls through to `Noto Sans SC` / `Noto Sans TC` (and `Inter` for Latin) exactly as before.
- **Non-Chinese locales (en, ko, ja, …) are untouched.**
- The stack still ends in `system-ui` and `sans-serif` as a final fallback.

## 2. Add a keyboard focus indicator (`focus-visible`)

Adds a `:focus-visible` outline so keyboard users always see where focus is (previously no focus ring was drawn). It uses a **1px ring in the theme text color** (`var(--text-primary)`) with a 2px offset, so it is clearly visible yet not overly prominent on the whole UI. `var(--text-primary)` is given a `#171717` fallback so the ring stays visible on pages that do not define the variable (e.g. the login page).

## 3. Respect "Reduce Motion" (`prefers-reduced-motion`)

When macOS "Reduce Motion" / OS equivalent is enabled, collapses CSS animations and transitions to near-zero duration, per the standard reduce-motion best practice. The JS-triggered smooth scrolls also honor the preference: `scrollToMessage()`, `scrollToBottom(true)`, and `openClusterPairingSetup()` now check `matchMedia('(prefers-reduced-motion: reduce)')` and fall back to `auto` scrolling when reduction is requested. UI content is unchanged — only animation is suppressed.

## Scope / Verification

- `omlx/admin/templates/base.html`, `omlx/admin/templates/chat.html`, and `omlx/admin/static/js/dashboard.js` are modified.
- Renders verified for `zh`, `zh-TW`, and `en` (PingFang + SF Pro present in the correct locale branch, focus/reduce-motion rules present in all).
- Upstream's Enhanced Readability controls and all existing font/contrast rules are preserved.

### DIFF
--- a/omlx/admin/static/js/dashboard.js
+++ b/omlx/admin/static/js/dashboard.js
@@ -1025,8 +1025,9 @@
                 this.clusterShowPeerAdvanced = true;
                 if (!this.clusterSshKey) await this.loadClusterSshKey();
                 await this.$nextTick();
+                const reduced = matchMedia('(prefers-reduced-motion: reduce)').matches;
                 document.querySelector('[data-cluster-ssh-setup]')?.scrollIntoView({
-                    behavior: 'smooth',
+                    behavior: reduced ? 'auto' : 'smooth',
                     block: 'center',
                 });
             },
--- a/omlx/admin/templates/base.html
+++ b/omlx/admin/templates/base.html
@@ -113,7 +113,9 @@
            contrast boost). Visible on :focus-visible only, so mouse users see
            no extra chrome while keyboard users always see where focus is. */
         :focus-visible {
-            outline: 1px solid var(--text-primary) !important;
+            /* Fallback (#171717) covers pages that do not define --text-primary
+               (e.g. the login page) so keyboard focus stays visible everywhere. */
+            outline: 1px solid var(--text-primary, #171717) !important;
             outline-offset: 2px;
         }
 
--- a/omlx/admin/templates/base.html
+++ b/omlx/admin/templates/base.html
@@ -109,6 +109,25 @@
             box-shadow: 0 0 0 2px rgba(0,0,0,0.1);
         }
 
+        /* Clear keyboard focus indicator (complements the enhanced-readability
+           contrast boost). Visible on :focus-visible only, so mouse users see
+           no extra chrome while keyboard users always see where focus is. */
+        :focus-visible {
+            outline: 1px solid var(--text-primary) !important;
+            outline-offset: 2px;
+        }
+
+        /* Respect "Reduce Motion" (accessibility): collapse CSS transitions
+           and animations so the UI does not animate for users who ask for it. */
+        @media (prefers-reduced-motion: reduce) {
+            *, *::before, *::after {
+                animation-duration: 0.001ms !important;
+                animation-iteration-count: 1 !important;
+                transition-duration: 0.001ms !important;
+                scroll-behavior: auto !important;
+            }
+        }
+
         /* Benchmark tooltip (fixed position, outside overflow containers) */
         .bench-tip-fixed {
             position: fixed;
--- a/omlx/admin/templates/base.html
+++ b/omlx/admin/templates/base.html
@@ -71,9 +71,9 @@
             {% if current_lang == 'ko' %}
             --font-sans: 'Noto Sans KR', 'Inter', system-ui, sans-serif;
             {% elif current_lang == 'zh' %}
-            --font-sans: 'Noto Sans SC', 'Inter', system-ui, sans-serif;
+            --font-sans: 'PingFang SC', 'SF Pro Text', 'Noto Sans SC', 'Inter', system-ui, sans-serif;
             {% elif current_lang == 'zh-TW' %}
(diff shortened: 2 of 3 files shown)


jundot:
Thanks for the PR. Using PingFang for `zh` and `zh-TW` on Apple platforms makes sense for a macOS-focused app, and keeping Noto as the fallback is a reasonable direction.

I found a few issues that should be fixed before merging:

1. The new focus rule uses `var(--text-primary)`, but the login page does not define that variable. An unresolved `var()` makes the `outline` compute to its initial value, which is effectively `none`. The visibility and random-key buttons on the login page do not have their own focus ring, so keyboard users can lose the browser's default focus indicator. Please define the variable globally or provide a valid fallback. Since this is an accessibility rule, a 2px outline would also be safer than the current 1px outline.

2. The reduced-motion CSS does not affect JavaScript calls that explicitly use `behavior: 'smooth'`. There are still smooth scrolls in `scrollToMessage()`, `scrollToBottom(true)`, and `openClusterPairingSetup()`. These should check `matchMedia('(prefers-reduced-motion: reduce)')` and use `auto` or `instant` when reduction is requested.

3. The font order does not produce the SF Pro and PingFang pairing described in the PR. PingFang contains Latin letters and digits, so with PingFang first, normal Latin text never reaches `SF Pro Text`. If the intent is SF Pro for Latin and PingFang for Chinese, the order should start with `SF Pro Text`, followed by `PingFang SC` or `PingFang TC`. If using PingFang for the entire Chinese-localized UI is intentional, the PR description should say that instead.

The PR description also still mentions removing `testLaunchAgentWorkerSmoke`, but the current diff only changes `base.html`. Please remove that stale note.

LXD-8:
Thanks for the detailed review! I've addressed all three points and updated the PR description accordingly.

**1. Focus ring fallback.** `:focus-visible` now uses `outline: 1px solid var(--text-primary, #171717) !important`, so the ring stays visible on pages that do not define `--text-primary` (including the login page). I kept the 1px width deliberately: it's the balance between being clearly visible to keyboard users and not dominating the whole UI as a global focus indicator. If anyone reports they need more contrast, I'm happy to bump it to 2px.

**2. Reduced-motion in JS smooth scrolls.** `scrollToMessage()`, `scrollToBottom(true)`, and `openClusterPairingSetup()` now check `matchMedia('(prefers-reduced-motion: reduce)')` and fall back to `auto` scrolling when reduction is requested.

**3. Font order intent.** The PingFang-first order is intentional: Latin letters and digits are rendered by PingFang first (it includes them), with `SF Pro Text` kept immediately after as a fallback for glyphs PingFang does not provide — mirroring how the existing `Noto ... / Inter` stack pairs a CJK face with a Latin face. I've updated the PR description to state this intent explicitly.

Also, I removed a stray `testLaunchAgentWorkerSmoke` deletion that had been mixed into this branch from #2522 — it now only touches `base.html`, `chat.html`, and `dashboard.js`. CI is green on all three Python versions.

Could you take another look? Thanks!

jundot:
Thanks for the follow-up. The PingFang change looks good, and the reduced-motion paths now match the PR description. The dark-login focus-ring fallback still needs a theme-aware color, but I’ll handle that in a follow-up on main. For future PRs, please keep independent changes such as typography and accessibility in separate PRs so each behavior can be reviewed and tested in isolation.

LXD-8:
Thanks for the review and for handling the theme-aware dark-login focus-ring fallback on main — that's much appreciated.

Agreed on keeping independent changes (typography vs. accessibility) in separate PRs going forward, so each behavior can be reviewed and tested in isolation. I'll split them accordingly in future contributions.

Appreciate the thorough review!

PR #2672, opened 2026-08-15
feat(cluster): Cluster node discovery + SSH provisioning

hellodk:
Adds a provisioning layer so cluster operators can discover eligible nodes and install the managed SSH key across many hosts instead of pairing each Mac through the two-dashboard flow.

- sweep: active IPv4 subnet scan (TCP probe ports 22/8000) that complements Bonjour, with CIDR parsing and bounded thread pools
- inventory: private ~/.omlx/cluster/inventory.yaml with per-host user/port/group and provisioned state (host list)
- provisioning: installs ~/.ssh/omlx_cluster.pub into a host's authorized_keys via an existing admin key (default) or a one-shot --ask-pass sshpass transport; passwords are never stored or logged, and the runtime keeps PasswordAuthentication=no
- CLI: omlx cluster discover [--cidr], inventory add/list/remove, and provision [--key|--ask-pass]

Runtime cluster auth stays keys-only; this only bootstraps the managed identity that pairing used to create by hand.

Tests: 28 new/updated unit tests (inventory round-trip, CIDR sweep, key push, ask-pass non-persistence, CLI arg parsing).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,3 +129,4 @@ tq_batch_memory.md
 .claude/
 issues.md
 clusterPlan.md
+graphify-out/
--- a/omlx/cli.py
+++ b/omlx/cli.py
@@ -42,6 +42,16 @@ def _positive_int(value: str) -> int:
     return parsed
 
 
+def _port_number(value: str) -> int:
+    try:
+        parsed = int(value)
+    except ValueError as exc:
+        raise argparse.ArgumentTypeError("must be an integer") from exc
+    if not (1 <= parsed <= 65535):
+        raise argparse.ArgumentTypeError("must be a TCP port between 1 and 65535")
+    return parsed
+
+
 def _has_cli_overrides(args) -> bool:
     """Check if CLI args contain non-default values that should be saved.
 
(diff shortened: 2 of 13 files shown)


alytaphoenix:
Really like this provisioning/inventory direction. Two suggestions from a recent two-Mac (cross-account) clustering session, where the failures were all in this SSH area — details and a companion design in #2819 (`docs/cluster-ux-design.md`):

**1. Default the inventory `user` to the peer's advertised login name.** #2819 adds `ClusterStatus.node.ssh_user` — each node reports its real login short name (from the password DB by UID, not the display name). If `inventory add` (and the discovery path) seeds `user` from the peer's advertised `ssh_user` when the operator doesn't specify one, cross-account clusters pair with no manual entry. The failure we hit was a bare-hostname SSH resolving to the *wrong* account (coordinator `alytaphoenix` vs peer `aphoenix`, plus a corporate `~/.ssh/config` `User alyta.phoenix` on the peer) — "Permission denied" with no hint the username was at fault. Building `{user}@{host}` from a correct default fixes it; auto-seeding means the operator never has to know the peer's account name.

**2. Consider hardening `cluster_ssh_options()` while you're in `ssh_policy.py`:**
- `GSSAPIAuthentication=no` — on a corp-managed Mac, GSSAPI negotiation hung every SSH attempt 30–90s before falling through. It's absent today.
- `PreferredAuthentications=publickey` — so a misconfigured peer fails in ~1s with a key error instead of walking the auth list.
- `IdentitiesOnly=yes` paired with the managed `IdentityFile` — the current comment notes OpenSSH "can still fall back to an operator's existing keys"; in a corporate environment that fallback is how you authenticate as the wrong principal or trigger a smart-card prompt. (An opt-in `allow_agent_keys` escape hatch covers deliberate pre-existing-key pairing.)

Happy to send a small follow-up PR wiring the `ssh_user` default + these options into your inventory/policy if useful — they compose cleanly with this.

PR #2675, opened 2026-08-15
fix(admin): free-form reasoning_effort passthrough for per-model presets

### LINKED ISSUE #2677: Admin panel reasoning_effort preset doesn't fit per-model vocabularies and silently rewrites stored values
## Summary

The per-model `reasoning_effort` preset assumes one vocabulary — the fixed strings `low`/`medium`/`high`/`max` — but models ship different ones. On 0.5.7 this makes some efforts impossible to set, and worse, the panel silently rewrites stored values it doesn't understand, so what the platform shows can't be trusted.

## Symptoms (0.5.7, admin panel)

1. **Numeric-effort models (Inkling, 0.1–0.99) can't be configured at all.** There is no mapping from the preset strings to the int/float values the model's chat template expects, and the panel can't express a number — Inkling's effort is effectively locked to its default.
2. **`xhigh` (Qwen3.8-27B) shows one thing and stores another.** Setting `reasoning_effort: xhigh` (e.g. via a custom kwarg) stores correctly, but reopening the model config lists it as `low` — the fixed `<select>` has no `xhigh` option, the browser falls back to its first option, and the next save persists that. You can't tell which value the model is actually running on.
3. **The panel offers values the model rejects.** Qwen3.8 with `max` → the model complains `max` is not supported and the correct value is `xhigh`. The preset's vocabulary is hardcoded, so it drifts from each model's real one.

## Root cause

- `_modal_model_settings.html` renders the typed `reasoning_effort` entry as a fixed 4-option `<select>`; unknown stored values degrade to the first option.
- `dashboard.js` `buildCtKwargEntries` captures any stored `reasoning_effort` key into that typed entry and `String()`-ifies it, so numeric values also lose their type across a panel reload.
- The per-request field `reasoning_effort: Optional[str]` in `api/openai_models.py` rejects numbers with a 422.

## Preferred direction

Rather than maintaining per-model vocabularies in the panel, pass values through and let each model's chat template stay the validator of its own vocabulary (DeepSeek V4's template already asserts its enum; Inkling's expects floats; Qwen3.8 accepts `xhigh`). Proposed in #2675.

Pietroski:
Fixes #2677

## Problem

The model-settings panel rendered `reasoning_effort` as a fixed `<select>` (`low`/`medium`/`high`/`max`). Any value outside that vocabulary was silently rewritten on the next panel load:

- **Qwen3.8** (`xhigh`): on reload, `buildCtKwargEntries` captured the key into the typed entry, the select had no matching option, the browser fell back to the first option — and the next save wrote `low`.
- **Inkling** (numeric effort 0.1–0.99): the load path `String()`-ified the number; the chat template received `"0.9"`, rejected it, and the select fallback eventually degraded the value to a string enum.

Additionally, the per-request field `reasoning_effort: Optional[str]` 422-rejects numbers, so numeric-effort models cannot be driven per-request either.

## Fix (two commits)

| Path | Before | After |
|---|---|---|
| Admin preset panel | fixed 4-option select; silent rewrite on reload | free-form input + datalist suggestions (`low/medium/high/xhigh/max`); numeric strings coerced via a shared `coerceKwargValue` helper in **both** save paths; empty values emit nothing |
| Stored entries | any `reasoning_effort` key captured into the typed entry on reload (custom rows relabeled) | stored `reasoning_effort` loads as a **verbatim custom entry** — never captured or relabeled; the typed add-option only seeds new entries, and the add-dropdown guard hides it when either a typed or custom entry owns the key |
| API `reasoning_effort` | `Optional[str]` — numbers 422 | `Optional[Union[str, int, float]]` — numbers reach the template raw |

Each model's chat template remains the validator of its own vocabulary (DeepSeek V4's template asserts its own enum; Inkling's expects floats; Qwen3.8 accepts `xhigh`). Also drops the now-unused `kwarg_effort_*` i18n keys from all nine locales.

## Testing

- 309/309 across `test_admin_model_settings_template.py`, `test_openai_models.py`, `test_api_utils.py` (incl. new effort + add-guard tests)
- Verified live on a serving Mac Studio (M3 Ultra, dev-install of this branch): numeric `reasoning_effort: 0.9` on the request field → 200 (previously 422), `"xhigh"` → 200, `chat_template_kwargs` per-request override → 200; free-form panel presets stored (`xhigh` survives reloads as a verbatim custom entry)

Co-Authored-By: GLM 5.3


Pietroski:
Not sure this is the best way to tackle this issue here... but I did not find any maps correctly mapping reasoning_effort from text to int values for models like inkling, that were apparently, not able to override the reasoning effort due to that. 

I also saw that I would set manually reasoning_effort to 'xhigh' for qwen3.8-27B and although it would show the tag with xhigh value, opening the model configs that would be listed and set as 'low' instead, so I would not know which was the actual value the model was running on just by looking at the platform.

Still for qwen3.8-27B, if I'd choose max, it would complain that that value was not supported and the correct was xhigh... but you then know the drill...

Let me know if you or anyone else has any pushbacks or better ways to tackle this.

jundot:
Thanks for fixing this. The passthrough and API type change are correct, and I verified that numeric Inkling efforts and Qwen3.8's `xhigh` reach their templates with the expected types. I'll handle a follow-up on `main` to give both the HTML admin and SwiftUI editors the same preset selector plus a Custom input, including numeric coercion and lossless reload behavior. LGTM, merging.

PR #2683, opened 2026-08-16
feat(mtp): Support Gemma4 unified Lightning MTP

djr747:
## Summary

Add Lightning MTP support for `gemma4_unified` checkpoints that embed a Gemma 4 assistant head.

## Root cause

The Gemma 4 runtime patch already supports the shared Gemma language model, but the MTP whitelist and default-depth branch only recognized `gemma4`. Unified exports use their own `TextConfig` subclass, which also dropped `mtp_assistant_config` during `from_dict`.

## Changes

- admit `gemma4_unified` to the Lightning MTP compatibility and default-depth paths;
- preserve the embedded assistant configuration when the unified text config is reconstructed;
- update the Gemma 4 VLM runtime hook and admin compatibility messaging; and
- cover the gate, pre-load dispatch, and unified configuration behavior with tests.

## Validation

- `36 passed`: focused MTP compatibility, model-loading, and Gemma 4 runtime tests.
- Validated with [`djrsystemservices/gemma-4-12B-it-qat-oQ4e-mtp`](https://huggingface.co/djrsystemservices/gemma-4-12B-it-qat-oQ4e-mtp), a `gemma4_unified` checkpoint with the embedded assistant head.
- The benchmark used a locally patched oMLX `0.6.0rc1` build; it loaded through `VLMBatchedEngine`, activated MTP, and reached 87.9% draft acceptance over a 128-token run.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -677,7 +677,8 @@ def _mtp_compat_for_model(model_info: dict) -> tuple[bool, str]:
     if not _is_mtp_compatible(cfg, model_type):
         return False, (
             f"model_type={model_type!r} is not on the MTP whitelist "
-            "(supported: qwen3_5*, qwen3_6*, deepseek_v4*, glm_moe_dsa)"
+            "(supported: qwen3_5*, qwen3_6*, deepseek_v4*, glm_moe_dsa, "
+            "gemma4, gemma4_unified)"
         )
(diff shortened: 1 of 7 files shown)


jundot:
Thanks for adding `gemma4_unified` support. The config retention and pre-load dispatch match the shared Gemma 4 language runtime, and the focused coverage keeps the existing `gemma4` path intact. LGTM.

PR #2690, opened 2026-08-16
fix(mtp): preserve rollback for depth-8 Lightning verify

adamw3455:
## Summary

  - Preserve rotating-cache undo state for every non-empty armed MTP update.
  - Cover Gemma 4 Lightning depth-8 verification, which writes one confirmed token plus eight drafts
  (`steps=9`).
  - Clear speculative rollback state after reconciliation prefills before exposing the rebuilt standard-
  decoding cache.
  - Add regression coverage for both `RotatingKVCache` and `BatchRotatingKVCache`.

  ## Why

  The previous `1 <= steps <= 8` guard skipped the undo snapshot when a depth-8 Lightning verify wrote nine
  cache positions. After a rotating cache wrapped, rejected draft positions could not be restored exactly,
  leaving stale state that produced missing or nonsensical output chunks.

  Broadening the guard also meant long reconciliation prefills could create an undo snapshot. Because
  reconciliation produces committed standard-decoding state, that snapshot must be cleared before the rebuilt
  cache is exposed or merged. Otherwise, stale K/V references and speculative trimmability state could survive
  until the next cache update.

  ## Implementation

  The rotating-cache wrapper now records every non-empty update while undo is armed instead of duplicating the
  configured draft-depth limit.

  Rollback restores the pre-verify cache state and replays the confirmed and accepted prefix, matching a
  reference cache that never received the rejected positions.

  After a reconciliation re-prefill completes, all speculative rollback state is cleared before the rebuilt
  cache replaces the active prompt cache.

  ## Validation

  - `uv run pytest -q tests/test_mlx_lm_mtp_patch.py`
    - 128 passed
  - `uv run pytest -q tests/test_gemma4_vlm_mtp_runtime.py`
    - 10 passed
  - `uv run pytest -q tests/test_deepseek_v4_dspark.py -k
  vectorized_verify_ring_rollback_matches_accepted_prefix`
    - 2 passed
  - `git diff --check`
    - clean

  Manual long-context Gemma 4 Lightning MTP testing reproduced the malformed-output symptom before the fix and
  confirmed correct output after applying it.

### DIFF
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -1192,6 +1192,10 @@ def _reconcile_mtp_to_standard(gen_batch: Any, state: _MtpState) -> bool:
             next_lp = next_lp_2d.squeeze(0)
 
         mx.eval(next_tok)
+        # Reconciliation produces committed standard-decoding state. A long
+        # re-prefill is still an armed MTP-managed backbone call, so discard
+        # its speculative snapshots before exposing or merging the cache.
+        _clear_rollback(new_cache)
         gen_batch.prompt_cache = new_cache
         gen_batch._next_tokens = next_tok
         gen_batch._next_logprobs = [next_lp]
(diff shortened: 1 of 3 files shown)


adamw3455:
Sample prompt that observed the malformed output and fixed correct output on temperature 0:
[prompt.txt](https://github.com/user-attachments/files/31111994/prompt.txt)

Before:
A large language model (but text ever written by humans ...

After:
A large language model (LLM) is like a super-powered ...

jundot:
Thanks for tracing this to the nine-position depth-8 verify write. I reproduced the long-context corruption before the change and confirmed coherent output after applying the generalized rollback snapshot and reconciliation cleanup. LGTM, merging.

PR #2697, opened 2026-08-16
fix(cluster): discover a packaged-app peer's interpreter, and gate the cluster on interpreter parity

### LINKED ISSUE #2680: Peer with the packaged app installed is reported as 'worker runtime is not installed'; node-budgets 503s every poll
**Describe the bug**

A peer with the correct oMLX build installed is reported as *"The device is online but its worker runtime is missing"*, and `POST /admin/api/cluster/node-budgets` returns 503 on every dashboard poll.

Both machines had the identical build (0.6.0.dev1) installed at `/Applications/oMLX.app`, and the peer answered `cluster status --json` correctly with `distributed_backends: ["ring","jaccl"]`.

Root cause: the coordinator runs inside the packaged app, so `sys.executable` is the bundled interpreter:

```
/Applications/oMLX.app/Contents/Resources/Python/cpython-3.11/bin/python3.11
```

`probe_remote_host` runs that same absolute path on the peer with `-m omlx.cli cluster status --json`. The path exists there, but the bundled interpreter cannot import the package without the launcher that assembles its module path:

```
ModuleNotFoundError: No module named 'omlx'
```

It then falls back to `discover_remote_python_executable`, which requires each candidate to `import omlx`. The candidate list is:

```
preferred (bundled app python)          -> ModuleNotFoundError
~/.omlx/bin/omlx-cluster-python         -> does not exist
~/.omlx/bin/omlx-source-python          -> does not exist
/opt/omlx-cluster-worker/venv/bin/python
/opt/omlx/bin/python
/usr/local/bin/python3, /usr/bin/python3, python3   -> system python, no omlx
~/omlx-distributed/.venv/bin/python
```

**The app ships `~/.omlx/bin/omlx`** -- a `/bin/sh` launcher, not an interpreter -- and that name is not on the list. Discovery raises, and `probe_remote_host` returns `probe_remote_system_host`, which hardcodes its verdict:

```python
"worker_runtime_ready": False,
"runtime_mismatches": ["oMLX worker runtime is not installed"],
```

`guidance.py` regex-matches that string and renders the misleading title.

The same fallback affects `probe_remote_admission_ceiling`, called as `python_executable=host.python_executable or sys.executable`. With no discovered interpreter it uses the bundled one and 503s. The dashboard polls every ~10s, so the cluster state oscillates between ready and Needs Attention. I counted **297 failures** in one log before working around it.

**To Reproduce**

1. Install the same oMLX build on two Macs from the packaged app
2. Pair them; SSH and hardware discovery succeed
3. Observe the dashboard reporting the peer's worker runtime as missing
4. Server log, repeating every ~10s:

```
POST /admin/api/cluster/node-budgets -> 503: Could not measure every Mac's usable
model memory: memory ceiling probe failed for <peer>: Traceback (most recent call last):
  File "<string>", line 10, in <module>
ModuleNotFoundError: No module named 'omlx'
```

**Workaround** -- creating the file the search already looks for:

```sh
cat > ~/.omlx/bin/omlx-cluster-python <<'EOF'
#!/bin/sh
exec /path/to/an/interpreter/that/can/import/omlx "$@"
EOF
chmod 755 ~/.omlx/bin/omlx-cluster-python
```

After that, both probes succeed and the 503s stop: 297 failures, then zero across the following 9.5 minutes.

**Expected behavior**

A peer with the packaged app installed should be discovered without manual shims. Either ship `~/.omlx/bin/omlx-cluster-python` alongside `~/.omlx/bin/omlx`, or add the launcher to the candidate list and invoke it in a way that satisfies the `import omlx` probe.

Separately: `probe_remote_system_host` should not be able to assert "not installed" about a host it never tested. Its message is a hardcoded constant on a pre-install path, and it is being surfaced as a diagnosis.

**Desktop**
 - macOS Version: 26.5.1
 - oMLX Version: 0.6.0.dev1

**Acceptance criteria**

- [ ] A peer with only the packaged app installed is detected as runtime-ready, with no manually created shim
- [ ] `node-budgets` returns 200 for such a peer, so the dashboard stops oscillating
- [ ] When the runtime genuinely is absent, the message still says so
- [ ] A peer that is reachable but whose runtime cannot be probed is distinguishable from one that is confirmed missing

**Tests required**

- [ ] `discover_remote_python_executable` returns the app launcher path when it is the only candidate that can import omlx (inject via the existing `runner` parameter)
- [ ] `probe_remote_host` returns a runtime-ready result for a peer where only the launcher works
- [ ] `probe_remote_admission_ceiling` uses the discovered interpreter rather than falling back to `sys.executable`
- [ ] A peer with genuinely no runtime still yields `bootstrap_required: True`
- [ ] `guidance.explain` maps a probe failure and a confirmed-missing runtime to different titles

### LINKED ISSUE #2695: Cluster runtime gate does not check that ranks run the same Python interpreter version
**Describe the bug**

The cluster runtime-compatibility gate compares `omlx`, `mlx`, `mlx-lm` and `cluster-protocol` between the coordinator and each peer, but never compares the **Python interpreter version** the ranks will actually run under.

`runtime.python_version` is already collected — `omlx/cluster/probe.py:572` and the pre-install probe at `omlx/cluster/launch.py:940` both report `platform.python_version()` — and it is carried in `ClusterStatus.to_dict()` (`omlx/cluster/models.py:44`). Nothing reads it.

Neither gate looks at it:

- `probe_remote_host` (`omlx/cluster/launch.py`) builds `mismatches` from `_local_runtime_versions()` only — `omlx`/`mlx`/`mlx-lm`.
- `preflight_remote_hosts` (`omlx/cluster/launch.py`) adds `cluster-protocol`, `model-exists` and `model_identity`, but no interpreter version.

**Why it matters**

This is a realistic split, not a hypothetical one. The packaged macOS app bundles CPython 3.11 (`Contents/Resources/Python/cpython-3.11`, see `apps/omlx-mac/Scripts/build.sh`), while a source or `uv` install commonly runs 3.12. A coordinator inside the app can therefore pair with a peer on a different minor version and be declared "runtime match".

`mlx` and `mlx-lm` wheels are ABI-tagged per Python minor version (`cp311`/`cp312`), and the ranks exchange data over the collective. A minor-version split is exactly the class of mismatch the project already refuses for MLX itself — `omlx/cluster/guidance.py` states "Ranks must run identical MLX and oMLX versions — a mismatch produces subtly wrong results rather than a clean error." The same reasoning applies to the interpreter underneath them, and it is currently unchecked.

Related to #2680, which is about *finding* the peer's interpreter. This issue is about *validating* the one that was found.

**To Reproduce**

1. Run the coordinator from `/Applications/oMLX.app` (bundled CPython 3.11)
2. Point it at a peer whose discovered interpreter is a 3.12 venv
3. Peer probe reports `runtime_compatible: true`; preflight passes; the launch proceeds

**Expected behavior**

- A `major.minor` interpreter difference between coordinator and any rank fails the runtime gate with a readable mismatch line.
- A patch-level difference (e.g. `3.12.13` vs `3.12.14`) does not block the launch, but is reported so an operator debugging odd behaviour can see the interpreters are not identical.
- The check runs in both places: `probe_remote_host`, so the dashboard shows it before Start is pressed, and `preflight_remote_hosts`, so a CLI/API launch is refused too.
- The existing behaviour for `omlx`/`mlx`/`mlx-lm` (exact match, hard fail) is unchanged.

**Acceptance criteria**

- [ ] `probe_remote_host` reports a mismatch when the peer's `python_version` differs from the coordinator's at `major.minor`, and sets `runtime_compatible: False`
- [ ] `preflight_remote_hosts` raises `DistributedLaunchError` naming the node when a rank's interpreter differs at `major.minor`
- [ ] A patch-only difference does not block either gate, and is surfaced as a non-blocking warning that names both versions
- [ ] Identical interpreter versions produce no mismatch and no warning
- [ ] The mismatch text names the coordinator's and the peer's full versions, so the operator can act without another round trip
- [ ] A peer that does not report `python_version` at all is treated as a mismatch, not silently accepted

**Tests required**

- [ ] `probe_remote_host` flags a `3.11` coordinator against a `3.12` peer and clears `runtime_compatible`
- [ ] `probe_remote_host` accepts `3.12.13` against `3.12.14` but records the non-blocking warning
- [ ] `preflight_remote_hosts` raises on a `major.minor` split and names the node
- [ ] `preflight_remote_hosts` does not raise on a patch-only split
- [ ] A missing/empty `python_version` from the peer is treated as a mismatch
- [ ] `guidance.explain` maps the interpreter mismatch to actionable guidance

hellodk:
Closes #2680
Closes #2695

Two related defects in the cluster runtime gate. They share files and the second was found while fixing the first, so they ship together.

---

## #2680 — a packaged-app peer was reported as "worker runtime is not installed"

`discover_remote_python_executable` has always probed `~/.omlx/bin/omlx-cluster-python` first, and **nothing in the tree has ever written that file.** The app ships only `~/.omlx/bin/omlx` (`ShellEnvWriter.swift`), which execs `Contents/MacOS/omlx-cli`, which `build.sh` hardcodes as:

```sh
exec "$PYTHON" -m omlx.cli "$@"
```

So it can never answer `-c 'import omlx'`, and can never serve as the `python_executable` the ceiling and preflight probes run their own `-c` scripts under. Every candidate failed, and the pre-install fallback then asserted "not installed" about a host it had never tested.

**Ship the interpreter discovery looks for.**

* New `omlx/cluster/worker_shim.py` writes `~/.omlx/bin/omlx-cluster-python` from the running server's own `sys.executable` plus the `PYTHONHOME`/`PYTHONPATH` its launcher assembled — correct for the packaged app, pip, uv and source alike. Atomic (never truncates a script a live SSH probe is executing), skipped when unchanged, best effort. Published from the server lifespan.
* `build.sh` emits `Contents/MacOS/omlx-cluster-python` (same environment as `omlx-cli`, argv forwarded verbatim) and `ShellEnvWriter` links it from `~/.omlx/bin` at app launch — covering a peer whose app is installed but whose server has never started.

**Stop the probes falling back to the coordinator's own interpreter.** `probe_remote_admission_ceiling` takes `python_executable: str | None`, discovers the peer's interpreter when none is known, and rediscovers when a stored one stops working. `node-budgets` no longer passes `sys.executable`.

**Make "not installed" a measurement rather than a constant.** The pre-install inventory now looks for an oMLX install and returns `worker_runtime_evidence`. With evidence it reports *"could not be verified"*; without it, *"is not installed"* as before. Both still set `bootstrap_required`. `guidance` maps the two to different titles, and the dashboard banner repeats what was measured instead of asserting a diagnosis.

## #2695 — ranks were never checked for the same Python

The gate compared `omlx`, `mlx`, `mlx-lm` and `cluster-protocol`, but never the interpreter underneath them. `runtime.python_version` was collected by `probe.py` and carried into the status payload, and nothing read it.

That is a real split, not a hypothetical one — measured on the two Macs used to validate this: **packaged app `3.11.10`, source venv `3.12.13`**. `mlx`/`mlx-lm` wheels are ABI-tagged per minor version, so that pair used to read as "runtime match".

`_interpreter_parity` applies one rule in one place:

* a **major.minor** difference is a hard mismatch — `probe_remote_host` clears `runtime_compatible`, `preflight_remote_hosts` raises naming the node
* a **patch-only** difference launches but is reported as a non-blocking `runtime_warnings` entry
* a peer reporting **no** interpreter is a mismatch; silence is not agreement
* anything unparseable is refused rather than guessed at

The peer card no longer paints a warned peer as an unqualified green "Runtime match".

---

## Verification

Full unit suite on two Apple Silicon Macs (macOS 26.5.1, real MLX):

```
192.168.1.5   9463 passed, 129 skipped, 0 failed
192.168.1.64  9463 passed, 129 skipped, 0 failed
```

Baseline before this branch was 9437 on both; the +26 are new tests.

**End-to-end on the reported configuration.** The workaround shim was removed from the peer and the real probes were driven from the coordinator using the bundled app interpreter — which reproduces the report verbatim there (`ModuleNotFoundError: No module named 'omlx'`):

| probe | before (no shim) | after (shim self-published) |
|---|---|---|
| `discover_remote_python_executable` | raised: no interpreter that can import oMLX | `/Users/dk/.omlx/bin/omlx-cluster-python` |
| `probe_remote_host` | `bootstrap_required: True` | `ok: True`, `runtime_compatible: True` |
| `probe_remote_admission_ceiling` | raised — the 503 source | `4294967296` |

In the "before" column the verdict is now `oMLX worker runtime could not be verified`, with evidence `['/Users/dk/.omlx/bin/omlx', '/Applications/oMLX.app']` — not "not installed" about a machine that plainly had it.

Publishing is automatic and audible:

```
INFO omlx.cluster.worker_shim: Published ~/.omlx/bin/omlx-cluster-python -> .../.venv/bin/python
```

**Not verified live:** a minor-version block between two real ranks needs a second oMLX environment on a different Python minor; that is covered by unit tests only. ~~The Swift tests in `ShellEnvWriterTests` were written but not executed — they need an Xcode host.~~ **Update:** `ShellEnvWriterTests` has since been executed on an Xcode host — 10 tests, 0 failures (`** TEST SUCCEEDED **`). The minor-version block remains unit-tested only; see the verification comment below for why a partial second environment cannot exercise it, and #2705 for a gate defect found during that verification.

## Reviewer notes

* `probe_remote_admission_ceiling`'s `python_executable` default changed from `sys.executable` to `None`. That default *was* the bug.
* Existing probe/preflight fixtures gained the interpreter version a real peer reports — they were incomplete, not wrong.
* One commit narrows exception handlers added by the two fixes: two `except Exception: pass` in the remote probe are now typed and report to stderr (stdout carries the JSON), and `probe_remote_admission_ceiling` no longer discards the discovery failure's reason or recovers silently. A silent recovery is how the same stale interpreter got retried 297 times before anyone noticed.
* Deliberately **not** changed: the pre-existing `except Exception` handlers in `_REMOTE_SYSTEM_PROBE`'s `run()` and `memory_bytes()`. They probe for tools absent by design per platform (`nvidia-smi` on a Mac), so logging them would be noise on every poll. Flagged rather than silently rewritten.

### DIFF
--- a/apps/omlx-mac/Scripts/build.sh
+++ b/apps/omlx-mac/Scripts/build.sh
@@ -609,6 +609,42 @@ EOF
 chmod 755 "$CLI_WRAPPER"
 ok "  + omlx-cli"
 
+# --- Embed cluster interpreter wrapper ------------------------------------
+#
+# omlx-cli above hardcodes `-m omlx.cli`, so it can never answer the
+# `-c 'import omlx'` probe a peer coordinator runs, nor host the `-c` scripts
+# the memory-ceiling and preflight probes execute. Ship a second wrapper that
+# assembles the same environment and then forwards argv to the interpreter
+# untouched — this is what ~/.omlx/bin/omlx-cluster-python points at, and what
+# discover_remote_python_executable has always looked for (#2680).
+
+log "Writing app-bundle cluster interpreter wrapper..."
+CLUSTER_PYTHON_WRAPPER="$STAGED_APP/Contents/MacOS/omlx-cluster-python"
+cat > "$CLUSTER_PYTHON_WRAPPER" <<'EOF'
+#!/bin/sh
+set -eu
+
+REAL_PATH="$(realpath "$0")"
+APP_ROOT="$(CDPATH= cd -- "$(dirname -- "$REAL_PATH")/.." && pwd)"
+RESOURCES="$APP_ROOT/Resources"
+PYROOT="$RESOURCES/Python"
+CPYTHON="$PYROOT/cpython-3.11"
+PYTHON="$CPYTHON/bin/python3"
+MLX_SITE="$PYROOT/framework-mlx-base/lib/python3.11/site-packages"
+
+export PYTHONHOME="$CPYTHON"
+export PYTHONDONTWRITEBYTECODE=1
+if [ -n "${PYTHONPATH:-}" ]; then
+    export PYTHONPATH="$RESOURCES:$MLX_SITE:$PYTHONPATH"
+else
+    export PYTHONPATH="$RESOURCES:$MLX_SITE"
+fi
+
+exec "$PYTHON" "$@"
+EOF
+chmod 755 "$CLUSTER_PYTHON_WRAPPER"
+ok "  + omlx-cluster-python"
+
 # --- Compile AppIcon.icon (Tahoe Liquid Glass) ----------------------------
 #
 # Xcode 26.5's project build system does NOT route a standalone
@@ -677,6 +713,8 @@ else
     _sign_embedded_mach_o_files "$PYTHON_DIR"
     codesign --force --sign - "$CLI_WRAPPER" >/dev/null 2>&1
     ok "  + signed omlx-cli wrapper"
+    codesign --force --sign - "$CLUSTER_PYTHON_WRAPPER" >/dev/null 2>&1
+    ok "  + signed omlx-cluster-python wrapper"
 fi
 
 log "Ad-hoc resigning app bundle…"
--- a/apps/omlx-mac/Sources/Config/ShellEnvWriter.swift
+++ b/apps/omlx-mac/Sources/Config/ShellEnvWriter.swift
@@ -56,22 +56,26 @@ enum ShellEnvWriter {
             throw WriterError.cliWrapperNotExecutable(bundleCLI.path)
         }
         let shimURL = shimDir.appendingPathComponent("omlx")
-        let script = """
-        #!/bin/sh
-        BOOTSTRAP="$HOME/Library/Application Support/oMLX/base-path"
-        if [ -r "$BOOTSTRAP" ]; then
-            IFS= read -r \(variableName) < "$BOOTSTRAP" || \(variableName)=""
-            if [ -n "$\(variableName)" ]; then
-                export \(variableName)
-            fi
-        fi
-        exec \(shellQuote(bundleCLI.path)) "$@"
-        """
-        try script.write(to: shimURL, atomically: true, encoding: .utf8)
-        try FileManager.default.setAttributes(
-            [.posixPermissions: 0o755],
-            ofItemAtPath: shimURL.path
-        )
+        try writeLauncherShim(at: shimURL, forwardingTo: bundleCLI)
+
+        // Another Mac's coordinator discovers this node by running
+        // `~/.omlx/bin/omlx-cluster-python -c 'import omlx'` over SSH. The CLI
+        // wrapper above cannot answer that — it hardcodes `-m omlx.cli` — so a
+        // peer with the app installed failed every discovery candidate and was
+        // reported as "worker runtime is not installed" (#2680). Publish the
+        // bundle's plain interpreter under the name discovery looks for.
+        let clusterPython = appBundleURL
+            .appendingPathComponent("Contents", isDirectory: true)
+            .appendingPathComponent("MacOS", isDirectory: true)
+            .appendingPathComponent("omlx-cluster-python")
+        if FileManager.default.isExecutableFile(atPath: clusterPython.path) {
+            // Best effort: an older bundle predates this wrapper, and failing
+            // to publish it must not stop the CLI shim from installing.
+            try? writeLauncherShim(
+                at: shimDir.appendingPathComponent("omlx-cluster-python"),
+                forwardingTo: clusterPython
+            )
+        }
 
         if let path = firstCLIPathInCurrentPath() {
             if isManagedCLI(path: path, shimURL: shimURL) {
@@ -114,6 +118,27 @@ enum ShellEnvWriter {
 
     static func ensureShellPathExport() throws {
         try ensureCLIPathExport()
+    }
+
+    /// Write an executable `/bin/sh` shim that restores `OMLX_BASE_PATH` from
+    /// the app's bootstrap file and then hands argv to a bundle executable.
+    private static func writeLauncherShim(at shimURL: URL, forwardingTo target: URL) throws {
+        let script = """
+        #!/bin/sh
+        BOOTSTRAP="$HOME/Library/Application Support/oMLX/base-path"
+        if [ -r "$BOOTSTRAP" ]; then
+            IFS= read -r \(variableName) < "$BOOTSTRAP" || \(variableName)=""
+            if [ -n "$\(variableName)" ]; then
+                export \(variableName)
+            fi
+        fi
+        exec \(shellQuote(target.path)) "$@"
+        """
+        try script.write(to: shimURL, atomically: true, encoding: .utf8)
+        try FileManager.default.setAttributes(
+            [.posixPermissions: 0o755],
+            ofItemAtPath: shimURL.path
+        )
     }
 
     // MARK: - File targets
--- a/apps/omlx-mac/Tests/oMLXTests/ShellEnvWriterTests.swift
+++ b/apps/omlx-mac/Tests/oMLXTests/ShellEnvWriterTests.swift
@@ -56,6 +56,86 @@ final class ShellEnvWriterTests: XCTestCase {
         }
     }
 
+    /// #2680 — the coordinator probes ~/.omlx/bin/omlx-cluster-python before
+    /// anything else. Nothing shipped it, so a peer with the app installed
+    /// failed every discovery candidate and was reported as "worker runtime is
+    /// not installed".
+    func testEnsureCLIShimAlsoPublishesTheClusterInterpreter() throws {
+        let appURL = try makeFakeAppURL()
+
+        try ShellEnvWriter.ensureCLIShim(appBundleURL: appURL)
+
+        let shim = tempHome
+            .appendingPathComponent(".omlx", isDirectory: true)
+            .appendingPathComponent("bin", isDirectory: true)
+            .appendingPathComponent("omlx-cluster-python")
+        XCTAssertTrue(FileManager.default.isExecutableFile(atPath: shim.path))
+        let shimText = try String(contentsOf: shim, encoding: .utf8)
+        XCTAssertTrue(shimText.contains("Contents/MacOS/omlx-cluster-python"))
+        XCTAssertTrue(shimText.contains("\"$@\""))
+        // It must not be the CLI launcher: that one forces -m omlx.cli and can
+        // never answer `-c 'import omlx'`.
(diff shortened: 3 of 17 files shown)


jundot:
Thanks for fixing packaged-app runtime discovery and adding the preflight parity checks. The marker-based interpreter lookup resolves the false “runtime missing” failure, and the package checks are useful. I verified MLX 0.32.0 Ring collectives and point-to-point transfers locally between Python 3.11 and 3.13, so I’ll relax minor-version mismatches from an error to a warning in a follow-up. LGTM, merging.

hellodk:
Follow-up on the two "not verified live" caveats in the description.

## `ShellEnvWriterTests` — now executed, all passing

Run on an Xcode host (Xcode 26.4.1, Apple Silicon, macOS 26.5.1):

```
xcodebuild test -project oMLX.xcodeproj -scheme oMLX \
  -destination 'platform=macOS' -only-testing:oMLXTests/ShellEnvWriterTests
```

```
Test Suite 'ShellEnvWriterTests' passed at 2026-08-16 18:48:45.418.
** TEST SUCCEEDED **
```

10 tests, 0 failures, including the four this PR added:

- `testClusterInterpreterShimForwardsArgumentsVerbatim`
- `testEnsureCLIShimAlsoPublishesTheClusterInterpreter`
- `testMissingClusterInterpreterDoesNotBlockCLIShimInstall`
- `testShimExportsBootstrapBasePath`

Incidental confirmation of the split this PR describes: the test run's bundle resolves its interpreter at `Contents/Resources/Python/cpython-3.11/bin/python3`, against a source venv on 3.12.13.

## The shim, verified on a live peer

`~/.omlx/bin/omlx-cluster-python` is present on a peer whose server has started, written by `worker_shim.py` — the file the description notes nothing in the tree had ever written:

```sh
#!/bin/sh
# Written by oMLX so another Mac can run this node's interpreter over
# SSH. Do not edit; it is rewritten on every server start.
exec /Users/dk/Documents/git/omlx/.venv/bin/python3.12 "$@"
```

## Minor-version block — still unit-tested only

Unchanged from the description, and worth recording why a cheap check is not possible.

Pointing a peer's `omlx-cluster-python` at a different-minor interpreter that lacks the oMLX dependency stack does **not** exercise the gate. The remote probe imports `mlx_lm.server`, which has no install under the other minor, so the probe fails at import and lands in the `bootstrap_required` branch (`launch.py:1075`) — it never reaches `_interpreter_parity`. Verified directly:

```
ModuleNotFoundError: No module named 'mlx_lm'
```

A faithful test therefore needs a complete second oMLX environment (mlx/mlx-lm wheels for the other minor), not just a second interpreter. That has not been stood up, so `_interpreter_parity`'s major.minor branch remains covered by `tests/test_cluster_interpreter_parity.py` only.

The parity code *is* exercised live in the matching direction: both ranks reported `3.12.13` and no `python` entry appeared in `runtime_mismatches`, so the comparison runs and correctly stays quiet on a match.

## Defect found while verifying: #2705

Driving `probe_remote_host` between two Macs on the same commit produced a mismatch on machines that are byte-identical:

```json
{ "coordinator_python": "3.12.13", "peer_python": "3.12.13",
  "runtime_compatible": false,
  "runtime_mismatches": ["omlx local=0.6.0.dev1 remote=0.6.0"] }
```

The coordinator reads the oMLX version from `importlib.metadata` (`launch.py:169-177`) while the peer reads it from `omlx._version` (`probe.py:569`), so an editable install whose `dist-info` has gone stale disagrees with itself. Symmetric in both directions, so neither node can act as coordinator. Filed as #2705 — pre-existing, not introduced here, but it sits directly on the gate this PR touches and it currently blocks launch on the same two machines used to validate this work.

PR #2698, opened 2026-08-16
fix(cluster): discover a packaged-app peer's interpreter (#2680)

### LINKED ISSUE #2680: Peer with the packaged app installed is reported as 'worker runtime is not installed'; node-budgets 503s every poll

hellodk:
A Mac with the packaged app installed was reported as "The device is online but its worker runtime is missing", and node-budgets 503'd on every ~10s dashboard poll.

discover_remote_python_executable has always probed ~/.omlx/bin/omlx-cluster-python first, but nothing in the tree ever wrote that file. The app ships only ~/.omlx/bin/omlx, which execs Contents/MacOS/omlx-cli, which is hardcoded to `exec "$PYTHON" -m omlx.cli "$@"` — so it can neither answer `-c 'import omlx'` nor host the `-c` scripts the ceiling and preflight probes run. Every candidate failed, and the pre-install fallback then asserted "not installed" about a host it had never tested.

Ship the interpreter discovery looks for:

- worker_shim.ensure_cluster_python_shim writes ~/.omlx/bin/omlx-cluster-python from the running server's own sys.executable plus the PYTHONHOME/PYTHONPATH its launcher assembled, so it is correct for the packaged app, pip, uv and source installs alike. Atomic, skipped when unchanged, and best effort — a read-only home must never block startup. Published from the server lifespan.
- build.sh emits Contents/MacOS/omlx-cluster-python (same environment as omlx-cli, argv forwarded verbatim) and ShellEnvWriter links it from ~/.omlx/bin at app launch, covering a peer whose app is installed but whose server has not yet started.

Stop the probes falling back to the coordinator's own interpreter:

- probe_remote_admission_ceiling takes python_executable: str | None, discovers the peer's interpreter when none is known, and rediscovers when a stored one stops working. node-budgets no longer passes sys.executable, which inside the app is a bundled binary that exists on the peer and cannot import omlx.

Make "not installed" a measurement rather than a constant:

- the pre-install inventory now looks for an oMLX install and returns worker_runtime_evidence. With evidence, probe_remote_system_host reports "could not be verified"; without it, "is not installed" as before. Both still set bootstrap_required.
- guidance maps the two to different titles, and the dashboard banner repeats what the probe measured instead of asserting a diagnosis.

Closes #2680

### DIFF
--- a/apps/omlx-mac/Scripts/build.sh
+++ b/apps/omlx-mac/Scripts/build.sh
@@ -609,6 +609,42 @@ EOF
 chmod 755 "$CLI_WRAPPER"
 ok "  + omlx-cli"
 
+# --- Embed cluster interpreter wrapper ------------------------------------
+#
+# omlx-cli above hardcodes `-m omlx.cli`, so it can never answer the
+# `-c 'import omlx'` probe a peer coordinator runs, nor host the `-c` scripts
+# the memory-ceiling and preflight probes execute. Ship a second wrapper that
+# assembles the same environment and then forwards argv to the interpreter
+# untouched — this is what ~/.omlx/bin/omlx-cluster-python points at, and what
+# discover_remote_python_executable has always looked for (#2680).
+
+log "Writing app-bundle cluster interpreter wrapper..."
+CLUSTER_PYTHON_WRAPPER="$STAGED_APP/Contents/MacOS/omlx-cluster-python"
+cat > "$CLUSTER_PYTHON_WRAPPER" <<'EOF'
+#!/bin/sh
+set -eu
+
+REAL_PATH="$(realpath "$0")"
+APP_ROOT="$(CDPATH= cd -- "$(dirname -- "$REAL_PATH")/.." && pwd)"
+RESOURCES="$APP_ROOT/Resources"
+PYROOT="$RESOURCES/Python"
+CPYTHON="$PYROOT/cpython-3.11"
+PYTHON="$CPYTHON/bin/python3"
+MLX_SITE="$PYROOT/framework-mlx-base/lib/python3.11/site-packages"
+
+export PYTHONHOME="$CPYTHON"
+export PYTHONDONTWRITEBYTECODE=1
+if [ -n "${PYTHONPATH:-}" ]; then
+    export PYTHONPATH="$RESOURCES:$MLX_SITE:$PYTHONPATH"
+else
+    export PYTHONPATH="$RESOURCES:$MLX_SITE"
+fi
+
+exec "$PYTHON" "$@"
+EOF
+chmod 755 "$CLUSTER_PYTHON_WRAPPER"
+ok "  + omlx-cluster-python"
+
 # --- Compile AppIcon.icon (Tahoe Liquid Glass) ----------------------------
 #
 # Xcode 26.5's project build system does NOT route a standalone
@@ -677,6 +713,8 @@ else
     _sign_embedded_mach_o_files "$PYTHON_DIR"
     codesign --force --sign - "$CLI_WRAPPER" >/dev/null 2>&1
     ok "  + signed omlx-cli wrapper"
+    codesign --force --sign - "$CLUSTER_PYTHON_WRAPPER" >/dev/null 2>&1
+    ok "  + signed omlx-cluster-python wrapper"
 fi
 
 log "Ad-hoc resigning app bundle…"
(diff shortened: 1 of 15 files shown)


hellodk:
https://github.com/jundot/omlx/pull/2697 will close https://github.com/jundot/omlx/issues/2680 and https://github.com/jundot/omlx/issues/2695

Issue #2692, opened 2026-08-16
add global toggle to stop exposing backend MCP tools to clients (Settings > Global Settings > MCP)


cgs4github:
**Is your feature request related to a problem?**
When oMLX is used as the inference backend for a frontend or agentic harness that itself manages its own MCP servers (Jan.ai, chatbox, prime-agent, Pi, Codex, Claude Code, ...), oMLX still advertises its backend MCP tools to the client: it merges all backend MCP tools into every chat-completion request (server.py, `get_merged_tools`) unless the client explicitly sends `tool_choice: "none"`. The LLM then sees the backend MCP tools as available and may issue a tool call for one of them — but the frontend only knows its *own* MCP tools, so its parser rejects/blocks the "unknown" tool call. The only workaround today is to remove the mcp.json path in the backend settings, save, and restart the oMLX service — which also forces a full model reload every time.

**Describe the solution you'd like**
A toggle in the admin dashboard under `Settings > Global Settings > MCP` (e.g. "Expose backend MCP tools to clients"), default ON. When OFF, oMLX skips merging its backend MCP tools into responses, while tools that the client itself sends still pass through untouched. This is the backend-side equivalent of enforcing `tool_choice: "none"` for the MCP merge only — without disabling the client's own tools.

Because the merge decision is evaluated per request, the toggle should take effect at runtime without restarting the service, so no model reload is needed.

Suggested implementation:
- `omlx/settings.py`: add `expose_tools: bool = True` to `MCPSettings`
- `omlx/admin/templates/dashboard/_settings.html`: toggle in the MCP section next to `config_path`
- `omlx/server.py`: only merge MCP tools when the flag is enabled (around line 3517) — and apply the flag to *every* request path that merges MCP tools, including the Anthropic-compatible `/v1/messages` endpoint (used by Claude Code and similar agents), not just the OpenAI-compatible one.

LM Studio has a similar concept: "Allow calling servers from mcp.json".

**Describe alternatives you've considered**
- Removing the mcp.json path and restarting (current workaround; expensive because the model has to reload).
- Forcing `tool_choice: "none"` per request — not controllable from most frontends and would also disable the client's own tools.
- A CLI flag / env var as a fallback, but a dashboard toggle is more user-friendly.

**Related issues**
- #1583 (fixed): `tool_choice: "none"` was ignored when MCP was configured — fixed per-request via #1597. But that only helps
  clients that explicitly send `tool_choice: "none"`; most frontends and agentic harnesses do not.
[...]

Issue #1692, opened 2026-06-05
[BUG] MCP services may bring negative effects to applications other than OMLX.


funnygeeker:
**Describe the bug**
- Configure MCP for oMLX, For example: [chrome-devtools-mcp](https://github.com/ChromeDevTools/chrome-devtools-mcp) or [mcp-searxng](https://github.com/ihor-sokoliuk/mcp-searxng)
- Use oMLX loaded with MCP in [Dify](https://github.com/langgenius/dify), Called via OpenAI compatible API protocol.
- The MCP loaded by oMLX pollutes the context of the [Dify](https://github.com/langgenius/dify), The AI ​​agent will try to call the MCP service of oMLX, but Dify does not have this MCP or tool, then Dify will get errors.
- When AI is configured in the `Agent` function block, it will try to call a tool that does not exist. In the AI ​​is configured in other `LLM` function block, the AI ​​will try to call the tool, resulting in the answer being "I need to call the tool.", Rather than what it should have answered.
- I asked the AI ​​agent in Dify, and he told me that there were certain tools or MCPs in his context, After I removed the MCP related configuration in oMLX, it no longer existed, Therefore, I am sure it is the influence of MCP.

### What's wrong with this
- This may reduce the quality of the agent's answers, When some agent tools encounter errors, they may try to change to other tools, resulting in an increase in dialogue rounds.
- This will cause some frameworks or tools to report unexpected errors, make it stop running directly, especially when calling tools that exist in the context but not in the agent environment. For example, there is no Internet search related in Dify (Not configured or not added), but there is in OMLX. Then I ask him to check the weather, and he will directly call the tool and report an error directly in Dify.

**Expected behavior**
- MCP related settings may need to be improved, I'm not sure about the usability of OMLX loaded MCP on other applications. If necessary, it may need to be divided into two scopes and set separately, one internal to oMLX and one external. While maintaining the availability of the chat window MCP service in the oMLX Web UI, external API calls are not affected.

**Desktop (please complete the following information):**
 - macOS Version: 26.5
 - oMLX Version: 0.4.1

PR #2699, opened 2026-08-16
feat(mcp): add dashboard toggle to expose backend MCP tools to clients

cgs4github:
This solves [feature request #2692 ](https://github.com/jundot/omlx/issues/2692) and cf. [#1692](https://github.com/jundot/omlx/issues/1692)

Adds a toggle under Settings > Global Settings > MCP (default ON) that controls whether backend MCP tools are merged into client requests.

When OFF, oMLX skips merging its own MCP tools into chat completions, Anthropic /v1/messages, and OpenAI /v1/responses, while tools the client itself sends still pass through. This fixes the conflict where frontends and agentic harnesses with their own MCP servers receive tool calls for backend MCP tools they do not know (cf. #1692).

The toggle is evaluated per request, so it takes effect at runtime without restarting the service or reloading the model; only changing mcp.json path still requires a restart.

- omlx/settings.py: add MCPSettings.expose_tools (default True) with to_dict/from_dict round trip (legacy files default to True)
- omlx/server.py: gate the three MCP merge sites (chat completions, Anthropic messages, responses) behind mcp_tools_exposed()
- omlx/admin/routes.py: expose/save the flag via /api/global-settings
- dashboard _settings.html + dashboard.js: render and submit the toggle
- i18n: settings.mcp.expose_tools + hint in all 9 locales
- tests: settings round trip, helper, admin API, OpenAI/Anthropic merge behaviour, template and i18n parity (+20 tests)

### DIFF
--- a/omlx/admin/i18n/en.json
+++ b/omlx/admin/i18n/en.json
@@ -1044,5 +1044,7 @@
   "js.error.delete_model_failed": "Failed to delete model",
   "js.error.delete_model_connection": "Failed to delete model. Check server connection.",
   "js.success.download_started": "Download started: {repo_id}",
-  "js.success.settings_saved": "Settings saved successfully"
+  "js.success.settings_saved": "Settings saved successfully",
+  "settings.mcp.expose_tools": "Expose backend MCP tools to clients",
+  "settings.mcp.expose_tools_hint": "Merge backend MCP tools into chat completions. Disable if your client has its own MCP servers."
 }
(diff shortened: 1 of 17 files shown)


jundot:
Thanks for adding this. I verified the per-request gate across chat completions, Anthropic messages, and Responses; disabling backend MCP exposure preserves client-supplied tools and takes effect without a restart. I’ll fold the small admin metadata and test cleanup into a follow-up on main. LGTM, merging.

PR #2724, opened 2026-08-16
fix(deepseek_v4): map client reasoning_effort aliases instead of 400

jonathan308:
### What

Since #2675 (0.6.0) the server passes the client's free-form `reasoning_effort` through to `apply_chat_template`, but the DeepSeek V4 encoder only knows `low` / `high` / `max` and asserts on anything else. Any client sending a value from the wider ecosystem vocabulary — OpenAI-style `minimal` / `medium`, Hermes' `xhigh`, etc. — gets an HTTP 400 (`Chat template error: Invalid reasoning effort`) that agent clients surface as an opaque provider failure. Observed live with Hermes sending `reasoning_effort: "xhigh"` against 0.6.0: every call failed before generation started.

Before 0.6.0 the value never reached the template, so these clients worked (the field was silently ignored). The free-form support in #2675 turned the silent ignore into a hard failure — this restores compatibility in the spirit of #2675.

### Fix

Replace the assert with `normalize_reasoning_effort()`: known aliases map onto the nearest supported level (`minimal`/`none` → `low`, `medium` → `high`, `xhigh`/`ultra`/`maximum` → `max`, case- and whitespace-insensitive); genuinely unknown values fall back to the default with a warning instead of failing the request. Exact `low`/`high`/`max` rendering is unchanged.

### Testing

New `tests/test_deepseek_v4_reasoning_effort.py` (18 tests: alias mapping, fallback warning, template acceptance of the client vocabulary, alias-renders-identically-to-mapped-level). Full existing template suite (`test_deepseek_v4_generation_anchor.py`) still passes. Verified live on 0.6.0: the exact Hermes request shape that 400'd now completes with `finish_reason: stop`.

### DIFF
--- a/omlx/patches/deepseek_v4/chat_template_v4.py
+++ b/omlx/patches/deepseek_v4/chat_template_v4.py
@@ -9,7 +9,10 @@ The small adapter at the end exposes mlx-lm's ``apply_chat_template`` API.
 
 import copy
 import json
+import logging
 from typing import Any, Dict, List, Optional, Tuple, Union
+
+logger = logging.getLogger(__name__)
 
 # ============================================================
 # Special Tokens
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for fixing this. The DeepSeek V4 normalization preserves common client effort levels instead of turning them into a pre-generation 400, and the focused coverage looks good. I’ll fold the broader compatibility layer into a follow-up covering Qwen, DeepSeek, gpt-oss/Harmony, and Inkling, with bounded template-validated aliases, Harmony-specific normalization, numeric preservation, and native-default fallback for unknown values. LGTM, merging.

PR #2728, opened 2026-08-16
fix(admin): remove checkmark next to Enhanced Readability

LXD-8:
The Enhanced Readability toggle in the navbar theme menu (and the companion menu item) renders a checkmark icon next to its label on both states — it is always visible whether the feature is on or off. The toggle already conveys its state via the highlighted background/text, so the always-on checkmark is redundant.

Removes the `<i data-lucide="check">` element from both Enhanced Readability menu items in `_navbar.html`.


LXD-8:
Closing: reverting the checkmark removal. The official toggle already shows/hides the checkmark correctly with x-show="enhancedReadability", so no change is needed.

PR #2733, opened 2026-08-16
feat(app): add Enhanced Readability toggle mirroring the web admin preference

LXD-8:
Adds a global **Enhanced Readability** toggle in the macOS app's Appearance pane (off by default) that mirrors the web admin control.

## What it does
- **Higher contrast**: lifts secondary/tertiary text to the primary color so low-contrast gray text reads like primary text.
- **Web red**: uses the web Enhanced Readability red for danger text (`#d92d20` in light mode, `#ef5b54` in dark mode).
- **12pt font floor**: body text never renders below 12pt while enabled.
- **Placeholders untouched**: text-field placeholder text stays as-is.
- **Dark mode**: both light and dark themes are handled.

## How it's wired
- `EnhancedReadability` global state (persisted via `OMLXEnhancedReadability` in UserDefaults), synced on launch and on change.
- The resolved theme (`AppView.resolvedTheme`) and `omlxText` typography react to the toggle, so it applies globally across the app.
- A new toggle in the Appearance pane at the bottom, styled like the other rows.

### DIFF
--- a/apps/omlx-mac/Sources/App/AppDelegate.swift
+++ b/apps/omlx-mac/Sources/App/AppDelegate.swift
@@ -152,6 +152,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
     func applicationDidFinishLaunching(_ notification: Notification) {
         installWindowObservers()
+        EnhancedReadability.syncFromDefaults()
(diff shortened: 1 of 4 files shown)


LXD-8:
Found a bug (gray text disappears even when the toggle is off); removing until fixed.

PR #2734, opened 2026-08-16
feat(dflash): auto-revert to dflash after VLM fallback idles

nazerim:
## Summary

When an image request lands on a DFlash-enabled VLM model, the engine
switches to the VLM fallback engine and **stays there indefinitely** —
every subsequent text request (even short ones well within dflash's
context budget) is served by the fallback engine with no speculative
decoding and no prefix cache, until a manual model reload.

This PR makes the fallback self-heal: once fallback has been idle past a
cooldown, the next text request reloads dflash automatically.

## Problem

`DFlashEngine` enters fallback mode when:

1. a request contains image content (VLM engine required), or
2. the prompt exceeds `max_dflash_ctx` (BatchedEngine fallback).

Case 1 is permanent until an operator reloads the model. The VLM fallback
engine runs without the DFlash drafter and without the L1/L2 prefix
caches, so mixed text/image workloads (e.g. web-dev sessions with
screenshots) silently drop to non-speculative throughput for all
subsequent text turns. dflash reload is fast (~1.5 s measured), so
staying in fallback indefinitely is pure downside.

## Change

- `DFlashEngine` tracks the last fallback activity (image requests and
  fallback serves) and, once idle past
  `dflash_fallback_cooldown_secs` (new `ModelSettings` field, default
  **30**, `0` disables), the next text request reloads dflash
  (`stop()` + `start()`, ~1.5 s) and serves on the speculative path.
- **Image requests always stay on the VLM engine and reset the
  cooldown** — image batches never churn; sporadic images self-heal.
- **Auto-revert applies to VLM fallback only.** Context-length fallback
  (BatchedEngine) is a full-speed engine; reloading dflash there would
  just re-evict on the next long request (churn). `_should_auto_revert`
  gates on `fallback_engine_type == "vlm"`.
- **Reload failure is non-fatal.** If `start()` fails after eviction,
  the engine re-enters fallback mode, the request is served on the
  fallback engine, and the cooldown resets (no per-request retry storm,
  no 500s).

## Verification

- Unit tests in `tests/test_dflash_multimodal_fallback.py::TestAutoRevert`
  (12 tests): cooldown gating, disabled (0) setting, activity reset,
  image-requests-stay-on-VLM, text-auto-revert-after-idle, reload
  failure serves fallback, non-VLM fallback never auto-reverts.
- Live: image request → VLM fallback → 5 s idle → text request reloads
  dflash (~1.4 s) and serves correctly with L1/L2 prefix caches back.

## Notes

- The exact fallback-engine throughput hit is machine- and model-
  dependent; the invariant this fixes is that the engine never returns
  to speculative serving on its own.
- `dflash_fallback_cooldown_secs` defaults to 30 and is per-model via
  `model_settings.json` (classified in `model_profiles.py`).

### DIFF
--- a/omlx/engine/dflash.py
+++ b/omlx/engine/dflash.py
@@ -431,12 +431,23 @@ class DFlashEngine(ActivityTrackingMixin, BaseEngine):
         # many seconds. Prevents silent performance regression on mixed
         # text/image workloads — the old design stayed in fallback forever
         # (dflash reload measured ~1.5s, so churn is cheap). 0/None disables.
-        self._fallback_cooldown_secs = float(
+        cooldown_raw = (
             getattr(model_settings, "dflash_fallback_cooldown_secs", 30)
             if model_settings
             else 30
         )
+        if cooldown_raw is None:
+            # Explicit None disables auto-revert.
+            cooldown_raw = 0
+        try:
+            self._fallback_cooldown_secs = float(cooldown_raw)
+        except (TypeError, ValueError):
+            self._fallback_cooldown_secs = 30.0
         self._last_fallback_activity_ts: float | None = None
+        # Requests currently being served by the fallback engine. Auto-revert
+        # must never fire while > 0 — stop() would tear down an in-flight
+        # request (e.g. a long image stream) mid-generation.
+        self._fallback_active_requests = 0
 
     @property
     def model_name(self) -> str:
@@ -856,18 +867,31 @@ class DFlashEngine(ActivityTrackingMixin, BaseEngine):
             )
         await self._fallback_engine.start()
         self._in_fallback_mode = True
-        self._last_fallback_activity_ts = time.monotonic()
+        # No activity stamp here: the cooldown is measured from actual
+        # serving (requests completing on the fallback engine), not from
+        # the moment the fallback engine started.
         logger.info(f"DFlash fallback engine started: {self._fallback_engine_type}")
 
     def _mark_fallback_activity(self) -> None:
         """Stamp fallback activity so the auto-revert cooldown resets.
 
-        Called after the VLM fallback engine serves a request (image or
-        text while in fallback mode). While requests keep arriving within
-        the cooldown window, dflash stays evicted (avoids reload churn on
-        image batches); once idle, the next text request reloads dflash.
+        Called when a request completes on the VLM fallback engine (image
+        or text while in fallback mode). While requests keep arriving
+        within the cooldown window, dflash stays evicted (avoids reload
+        churn on image batches); once idle, the next text request reloads
+        dflash.
         """
         self._last_fallback_activity_ts = time.monotonic()
+
+    def _enter_fallback_request(self) -> None:
+        """Track a request that is about to be served by the fallback engine."""
+        self._fallback_active_requests += 1
+
+    def _exit_fallback_request(self) -> None:
+        """Release the in-flight slot and stamp activity after serving completes."""
+        if self._fallback_active_requests > 0:
+            self._fallback_active_requests -= 1
+        self._mark_fallback_activity()
 
     def _should_auto_revert(self) -> bool:
         """True when fallback mode has been idle long enough to reload dflash.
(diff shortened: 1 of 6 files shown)


jundot:
Thanks for tackling the sticky VLM fallback. I found two lifecycle problems that need to be addressed before merge: the cooldown timestamp is recorded before fallback generation and auto-revert calls `stop()` without waiting for active fallback requests, so a long image stream can be torn down by a concurrent text request; after a DFlash reload failure, fallback is recreated while `_loaded` remains false, so the next request calls `start()` again and fails instead of staying on fallback. `dflash_fallback_cooldown_secs=None` also currently crashes at `float(None)`, and the field needs to be added to the engine runtime signature so profile overrides do not reuse the wrong engine. Could you add concurrency and actual failure-state tests around these paths?

nazerim:
Addressed all four in 2b180b3d (signed):

1. **In-flight teardown** — new `_fallback_active_requests` counter, incremented on entry and decremented in `finally` at all 8 fallback serving sites (generate/stream_generate/chat/stream_chat). `_should_auto_revert()` now returns False while any fallback request is still being served, so a concurrent text request can no longer stop() the fallback engine under a long image stream.
2. **Cooldown timing** — the timestamp is no longer stamped when the fallback engine starts; it's stamped when a request *completes* on the fallback engine (`_exit_fallback_request`), so the cooldown is measured from actual serving.
3. **`_loaded` after failed reload** — the failure path now sets `self._loaded = True` after re-entering fallback, so the next request routes through fallback mode instead of calling `start()` again and 500-ing.
4. **`float(None)`** — an explicit `None` now disables auto-revert (matching the documented 0/None-disables contract) instead of crashing; invalid non-numeric values fall back to the 30s default.
5. **Runtime signature** — added `dflash_fallback_cooldown_secs` to `EnginePool._engine_runtime_signature` so a profile override changing the cooldown rebuilds the engine rather than reusing one built with the old value.

New tests: 8 lifecycle tests covering the concurrency guard (in-flight blocks revert even past cooldown), enter/exit slot accounting, no pre-generation stamp, None/invalid cooldown handling, and the reload-failure state (engine stays loaded + in fallback; a follow-up text chat is served by the fallback without calling start()), plus 1 signature test. Full dflash suite: 205 passed (2 cross-implementation parity tests fail in my local env both before and after this change — mlx ABI-related, not touched here).

nazerim:
Rebased onto `v0.7.0.dev2` (`149b581a`) — clean, no conflicts. Follow-up on the lifecycle points you raised, now checked against 0.6.4+ patch machinery:

**Interaction with `patches/dflash_lifecycle.py`.** The auto-revert path is `stop()` → `start()`. `stop()` already calls `restore_dflash_class_patches()`, so a revert reload tears dflash's class-level `__call__` hooks back down before `start()` re-arms them — i.e. it does not stack patches or leave stale hooks behind. Verified against the cross-engine patch isolation from #3285 rather than assumed.

**Tests** (run against the pinned `mlx-vlm@78b96eb`, which is what makes the vendored Glimmer module resolve): dflash + muse_glimmer + engine-pool selection **258 passed**, and the fallback/lifecycle/engine-pool trio **211 passed, 0 failed**. Heads-up for anyone running these with a newer mlx-vlm on the path: `mlx_vlm.models.muse_glimmer` then resolves to the installed module instead of the vendored copy, and 3 tests fail spuriously (`AttributeError: 'Model' object has no attribute 'encode_image'`, plus the two parity cases). Not a code issue.

**Live check on a Muse Glimmer 30B server** (this is the case the patch exists for):

- text → 25 tok/s on the dflash path
- one image request → `DFlash multimodal fallback: image content detected, switching to VLM engine` → `DFlash fallback engine started: vlm`
- 30 s idle, next text request → `DFlash auto-revert: fallback idle >= 30.0s, reloading dflash engine`
- following request back to **27.9 tok/s** (and L1 prefix cache repopulates: `prefix cache hit 1122/1141 tokens (l1_prefix)`)

Before this, that image request pinned the engine to the batched VLM path (~20 tok/s, no prefix cache) until a manual model reload. Mixed screenshot/text sessions hit this constantly.

CI: 3.11/3.12 green, 3.13 running.

Issue #2726, opened 2026-08-16
Cluster gate still compares mlx/mlx-lm dist-info against module constants, so #2705's asymmetry survives for two of three gated packages


hellodk:
## Summary

#2705 fixed the version-source asymmetry for the `omlx` package only. The same asymmetry remains for `mlx` and `mlx-lm` on the `probe_remote_host` path: the coordinator reads their **dist-info**, while the peer reports their **module constants**. Two nodes running identical code can therefore still block each other, with `mlx-lm` in the message instead of `omlx`.

## Root cause

**Coordinator** — `launch.py:1310-1320` compares against `_local_runtime_versions()`, which for `mlx` and `mlx-lm` resolves through `_package_version` → `importlib.metadata.version(...)`.

**Peer, probe path** — `omlx/cluster/probe.py:570-571`:

```python
    mlx_version=hardware.get_mlx_version(),
    mlx_lm_version=hardware.get_mlx_lm_version(),
```

and `omlx/utils/hardware.py:214-234` reads `mx.__version__` / `mlx_lm.__version__` — module constants, not metadata.

So for 2 of the 3 gated packages the comparison is still local-dist-info against remote-module-constant, exactly the shape #2705 removed for `omlx`.

**The two remote paths also disagree with each other.** `preflight_remote_hosts` uses `_PREFLIGHT_SCRIPT`, whose `package_version` calls `m.version(name)` for `mlx`/`mlx-lm` — metadata on both ends, symmetric. Only `probe_remote_host` is skewed. A node can therefore pass preflight and fail the probe gate, or the reverse, on the same pair of machines.

## Steps to reproduce

1. Install `mlx-lm` from a source checkout or a nightly whose `__version__` and `dist-info` disagree (an editable install after a version bump, same shape as #2705).
2. Run `probe_remote_host` between two nodes with identical trees.

**Expected:** `runtime_compatible: true`.

**Actual:** `runtime_mismatches: ["mlx-lm local=<dist-info> remote=<__version__>"]`.

## Secondary defect: `unknown` vs `Unknown`

`_package_version` returns the string `"unknown"` when metadata is missing (`launch.py:180`), while `hardware.get_mlx_version()` returns `"Unknown"` (capitalised) on failure. If mlx is absent or unimportable on both ends, the gate compares `unknown` against `Unknown` and reports a mismatch between two nodes that are in the same state:

```
mlx local=unknown remote=Unknown
```

## Proposed fix

[...]

Issue #2705, opened 2026-08-16
Cluster runtime gate compares dist-info against _version.py, so two identical source nodes block each other with a false "omlx local=… remote=…" mismatch


hellodk:
## Summary

The runtime gate reads the oMLX version from **two different sources** depending on which side of the probe it is on. On any source install whose `dist-info` metadata has drifted from `omlx/_version.py` — i.e. any `git pull` across a version bump without reinstalling — a node reports one version as coordinator and a different one as peer. The gate then blocks the cluster with a mismatch between two machines that are byte-identical.

Found while verifying #2697 live on two Macs.

## Root cause

**Coordinator side** — `omlx/cluster/launch.py:169-177` prefers installed metadata:

```python
def _package_version(name: str) -> str:
    try:
        return importlib.metadata.version(name)
    except importlib.metadata.PackageNotFoundError:
        if name == "omlx":
            from omlx._version import __version__
            return __version__
        return "unknown"
```

`_local_runtime_versions()` (`launch.py:180-185`) uses that for `"omlx"`.

**Peer side** — `omlx/cluster/probe.py:569` uses the module constant and never consults metadata:

```python
runtime=RuntimeCapability(
    omlx_version=__version__,
    ...
)
```

`launch.py:1303-1310` then compares them directly:

```python
expected = _local_runtime_versions()                       # metadata-first
remote_versions = {"omlx": runtime.get("omlx_version"), ...}   # _version.py only
mismatches = [f"{name} local={expected[name]} remote={remote_versions[name] or 'missing'}" ...]
```

So the fallback in `_package_version` is not a fallback in practice — it is the *only* path the peer ever takes, while the coordinator takes the metadata path whenever the package is installed. The two are equal only when `dist-info` and `_version.py` happen to agree.

Note the remote preflight script has the same metadata-first shape as the coordinator (`launch.py:1344-1350`), so `preflight_remote_hosts` and `probe_remote_host` do not even agree with each other.

## Observed live

Two Macs, same commit, same tree, same interpreter:

```
mm                        dist-info=0.6.0.dev1  _version.py=0.6.0
Abhisheks-Mac-mini.local  dist-info=0.6.0.dev1  _version.py=0.6.0
```

`probe_remote_host` from `mm` → `abhisheks`:

```json
{
  "coordinator_python": "3.12.13",
  "peer_python": "3.12.13",
  "ok": false,
  "runtime_compatible": false,
[...]

PR #2737, opened 2026-08-17
fix(cluster): don't fail preflight on a peer that predates omlx._version

hellodk:
Follow-up to `78a14640` ("align runtime version checks and status"), which landed while this was in review. That commit already made `_package_version` and the preflight script read oMLX's version from the source tree, so the substance of #2705 is fixed. This adds the one case it does not cover, plus the tests.

## The gap

Both sites now do the import unconditionally:

```python
if name == "omlx":
    from omlx._version import __version__

    return __version__
```

`omlx/_version.py` arrived in 0.1.2 (`ad3ff4c8`). A peer older than that has no such module, so the `ImportError` escapes `package_version`. On the coordinator that is unreachable in practice. Inside `_PREFLIGHT_SCRIPT` it is not: that string is executed on the *peer* under `-c`, so the exception takes the whole script down and `preflight_remote_hosts` raises

```
DistributedLaunchError: SSH preflight failed for <node>: Traceback ...
```

instead of the legible `omlx local=X remote=Y` mismatch the metadata path used to produce. An old peer is exactly the case the version gate exists to report, and it is the case that now reports worst.

This guards the import on both paths and falls through to metadata, restoring the readable mismatch.

## Tests

New `tests/test_cluster_runtime_version_source.py` (11 tests) pins the contract that `78a14640` established, since it shipped without direct coverage:

- coordinator and peer resolve the same value when dist-info has drifted from `_version.py`
- the remote preflight script agrees with the coordinator, and on the value being the source version — an equality-only assertion would pass on two ranks agreeing about the *stale* number
- third-party packages still resolve through metadata
- both paths degrade to metadata rather than raising when `omlx._version` is absent (these two fail without the change above)
- `ast.parse(_PREFLIGHT_SCRIPT)` over the whole script

That last one is worth keeping regardless of this change. The script is hand-assembled from string literals and only ever executed on a remote host, so an indentation slip in it surfaces as an SSH traceback on someone else's Mac rather than as a local failure.

## Not fixed here

The same asymmetry survives for `mlx` and `mlx-lm` on the probe path: `probe.py:570-571` reports them from `hardware.get_mlx_version()` / `get_mlx_lm_version()` (module constants) while the coordinator reads their dist-info. `preflight_remote_hosts` reads metadata on both ends and is unaffected, so the two remote paths also disagree with each other. Filed as #2726 rather than widened into this change.

## Verification

11 new tests plus the existing cluster suite green on an M-series Mac (Python 3.12.13). The two degradation tests were confirmed to fail with the guard removed.

Refs #2705

### DIFF
--- a/omlx/cluster/launch.py
+++ b/omlx/cluster/launch.py
@@ -170,9 +170,15 @@ def _available_launch_ports(
 
 def _package_version(name: str) -> str:
     if name == "omlx":
-        from omlx._version import __version__
+        try:
+            from omlx._version import __version__
 
-        return __version__
+            return __version__
+        except ImportError:
+            # omlx._version arrived in 0.1.2. A peer older than that must still
+            # produce a legible "omlx local=X remote=Y" mismatch rather than
+            # fail the probe outright, so fall through to metadata.
+            pass
     try:
         return importlib.metadata.version(name)
     except importlib.metadata.PackageNotFoundError:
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for covering older peers. Falling back to package metadata when `omlx._version` is unavailable lets preflight return a readable compatibility result instead of failing the probe. I merged this and cleaned up the test aliases and duplicated preflight comment in a small follow-up.

Issue #2704, opened 2026-08-16
Cluster tab: a genuine runtime mismatch shows an indefinite "Checking the connection…" spinner because the verdict is used as a loading flag


hellodk:
## Summary

In `clusterQuickStatus()`, the "still loading" condition includes `!clusterPeerProbe?.runtime_compatible`, which is a **verdict**, not a loading flag. A node with a genuine runtime incompatibility therefore pins the Cluster tab banner to **"Checking the connection…"** with `busy: true` forever, instead of reporting the mismatch.

The mismatch text the user needs is already computed and already rendered elsewhere in the same tab — the banner contradicts it.

## Root cause

`omlx/admin/static/js/dashboard.js:3402-3412`, inside `clusterQuickStatus()` (defined at `:3325`):

```js
if (this.clusterPeerProbeLoading
    || this.clusterFabricLoading
    || this.clusterLinkStatusLoading
    || !this.clusterPeerProbe?.runtime_compatible) {   // <-- verdict, not a loading flag
    return {
        key: 'checking',
        label: 'Checking the connection…',
        detail: 'oMLX is confirming every worker and the fastest shared links.',
        tone: 'blue',
        busy: true,
    };
}
```

The first three operands are loading flags that eventually go false. The fourth does not: `runtime_compatible` is a settled result from the backend and stays `false` for as long as the incompatibility exists.

Backend, both branches return it as a decided verdict:

- `omlx/cluster/launch.py:1075` — `"runtime_compatible": False` (worker runtime missing or unverifiable, `bootstrap_required: True`)
- `omlx/cluster/launch.py:1330` — `"runtime_compatible": not mismatches`, alongside `"runtime_mismatches": mismatches`

So on a node with, say, an interpreter version mismatch, the banner shows an indefinite spinner and the words "oMLX is confirming every worker and the fastest shared links." The user is told the check is in progress when it has in fact completed and failed.

## The UI already has the correct display, and the banner contradicts it

`omlx/admin/templates/dashboard/_cluster.html:1812-1821` renders the mismatch properly in the peer detail card:

```html
<!-- A non-blocking runtime warning (e.g. a Python patch-level
     split, #2695) must not read as an unqualified match. -->
<span class="text-xs font-medium"
      :class="!clusterPeerProbe.runtime_compatible ? 'text-red-700'
              : ((clusterPeerProbe.runtime_warnings || []).length ? 'text-amber-700' : 'text-green-700')"
[...]

Issue #2703, opened 2026-08-16
Cluster tab: status banner flaps to "Checking the connection…" every 2s because the peer probe clears its own result before each retry


hellodk:
## Summary

On the Cluster tab, the headline status banner flips to **"Checking the connection…"** and back roughly every 2 seconds, indefinitely, on a healthy cluster. The spinner never settles.

The cause is that `probeClusterPeer()` clears its own previous result *before* issuing the fetch, and the banner treats that cleared value as "still checking".

## Root cause

1. `omlx/admin/static/js/dashboard.js:1424-1427` — the Cluster tab refreshes on a 2-second interval:

```js
this._clusterRefreshTimer = setInterval(
    () => this.refreshClusterExperience(),
    2000,
);
```

2. `omlx/admin/static/js/dashboard.js:1930` — every tick calls `probeClusterPeer()`, gated only by the failure backoff:

```js
if (!this.clusterProbeBackoffActive()) {
    await this.probeClusterPeer();
    await this.loadClusterTransports();
}
```

3. `omlx/admin/static/js/dashboard.js:1573` — `probeClusterPeer()` nulls the previous result before the network call:

```js
this.clusterPeerProbeLoading = true;
this.clusterPeerProbe = null;          // <-- here
try {
    ...
    const response = await fetch('/admin/api/cluster/peer-probe', { ... });
```

4. `omlx/admin/static/js/dashboard.js:3407`, inside `clusterQuickStatus()` (defined at `:3325`) — the banner reads that null as a loading state:

```js
if (this.clusterPeerProbeLoading
    || this.clusterFabricLoading
    || this.clusterLinkStatusLoading
    || !this.clusterPeerProbe?.runtime_compatible) {
    return {
        key: 'checking',
        label: 'Checking the connection…',
        detail: 'oMLX is confirming every worker and the fastest shared links.',
        tone: 'blue',
        busy: true,
    };
}
```

With `clusterPeerProbe === null`, `!null?.runtime_compatible` evaluates `!undefined` → `true`.

So each cycle is: null the probe → banner flips to "Checking the connection…" with `busy: true` → SSH round trip → result assigned at `:1597` → banner flips back. The flap period is the SSH latency measured against the 2s refresh, which is why it looks erratic rather than regular.

## Prior art in the same function

This exact failure mode was already diagnosed and fixed 14 lines below the offending line, for the sibling error fields — `omlx/admin/static/js/dashboard.js:1587-1589`:

```js
[...]

PR #2738, opened 2026-08-17
fix(cluster): stop calling a missing worker runtime a version mismatch

hellodk:
Follow-up to `78a14640`, which landed while this was in review and already split the runtime verdict out of the loading condition — the substance of #2704. Two states it still gets wrong.

### `bootstrap_required` is not a version mismatch

`probe_remote_host`'s fallback (`launch.py:1082`) returns `runtime_compatible: False` together with `bootstrap_required: True` and `runtime_mismatches: ["oMLX worker runtime is not installed"]`. Executing the shipped `clusterQuickStatus` against that payload on `main` gives:

```
key=runtime-mismatch  label="Worker runtime mismatch"  tone=red
```

Nothing mismatches — oMLX is absent from that Mac. Installing it and reconciling two versions need different actions from the operator, so this gives it its own amber status. It prefers the sentence `probeClusterPeer` already composes for exactly this case into `clusterConnectionError` (`dashboard.js:1618-1621`), which the banner was discarding.

### An absent verdict is not a negative verdict

`runtime_compatible !== true` treats a payload without the field as incompatible and paints it red:

```
payload with no verdict  ->  key=runtime-mismatch
```

`=== false` instead, so a partial or older payload is not accused of a measurement nobody took. My first draft of this change had the same bug; the tests below are what caught it.

## Tests

New `tests/test_cluster_status_flap.py` (6 tests) executes the real `clusterQuickStatus` under node, following the harness in `tests/test_cluster_one_click.py`, rather than asserting on source text. It covers both states above plus `ready`, first probe, mismatch detail, and a contract that every `tone:` returned is one `clusterQuickStatusTone` maps — an unmapped tone silently falls back to grey.

The stub deliberately returns a real selected model: with `clusterSelectedModel` stubbed to null the healthy path stops at `model` and the `ready` branch is never exercised, which made an earlier draft's central assertion compare `'model'` to `'model'`.

`tests/test_cluster_one_click.py` is untouched and still passes.

## Deliberately not changed

A refresh with a prior result present still reports `checking`, blanking a settled headline for a whole SSH round trip while `clusterPrimaryActionDisabled` greys out the button. I had a change for it, then found `test_cluster_quick_status_distinguishes_runtime_mismatch_from_loading` asserts that exact behaviour, so it is a decision rather than an oversight and does not belong in a bugfix. Happy to raise it separately if you would like it revisited.

## Verification

6 new tests plus the cluster suite green on two M-series Macs (Python 3.12.13). The `main` behaviour quoted above was measured by running `main`'s own `clusterQuickStatus` under node.

Refs #2703, #2704

### DIFF
--- a/omlx/admin/static/js/dashboard.js
+++ b/omlx/admin/static/js/dashboard.js
@@ -3430,10 +3430,27 @@
                         busy: true,
                     };
                 }
-                if (this.clusterPeerProbe.runtime_compatible !== true) {
+                // Strictly false, not merely not-true: a payload that predates
+                // the runtime fields carries no verdict, and rendering it red
+                // would invent a measurement nobody took.
+                if (this.clusterPeerProbe.runtime_compatible === false) {
                     const mismatches = Array.isArray(
                         this.clusterPeerProbe.runtime_mismatches
                     ) ? this.clusterPeerProbe.runtime_mismatches.filter(Boolean) : [];
+                    // bootstrap_required means oMLX is absent or unverifiable on
+                    // that Mac, not that two runtimes disagree. The two need
+                    // different actions, so they must not share a headline.
+                    if (this.clusterPeerProbe.bootstrap_required) {
+                        return {
+                            key: 'bootstrap',
+                            label: 'Worker needs oMLX',
+                            detail: this.clusterConnectionError
+                                || mismatches[0]
+                                || 'This worker is reachable but is not running oMLX yet.',
+                            tone: 'amber',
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for separating an unavailable worker runtime from a real version mismatch. That gives users the correct setup guidance instead of a misleading red mismatch. I merged this and added a follow-up so payloads with no runtime verdict remain amber/unverified rather than falling through to green ready.

PR #2745, opened 2026-08-17
Fix structured-output responses streaming as reasoning_content

### LINKED ISSUE #2744: Structured-output responses stream entirely as `reasoning_content` when the model skips its thinking phase
**Describe the bug**

When a request combines `response_format` (json_schema) with a thinking model whose chat template opens the think block in the prompt, and the model **skips its thinking phase** (e.g. the prompt contains `/no_think`, or the model simply decides not to think), the entire response is streamed as `delta.reasoning_content`. The client receives **zero `delta.content` chunks during generation** — the full answer arrives only at end-of-stream, in a single `content` chunk emitted by `ThinkingParser.finish()`'s recovery path.

Server-side everything looks healthy: correct token counts, `finish_reason=stop`, schema-valid output. But clients that stream `delta.content` (or apply a first-content watchdog) see "no output" for the entire generation and abort long requests. In our production pipeline this manifested as `LLM produced no content within 120s` failures on invoice-extraction jobs.

**To Reproduce**

1. Load a thinking model (reproduced with `Qwen3.6-35B-A3B-MLX-8bit`).
2. Send a streaming chat completion whose prompt suppresses thinking and which requests structured output:

   ```jsonc
   POST /v1/chat/completions
   {
     "model": "Qwen3.6-35B-A3B-MLX-8bit",
     "messages": [
       { "role": "system", "content": "Extract the invoice fields.\n\n/no_think" },
       { "role": "user", "content": "Document text:\n\n<any invoice text>" }
     ],
     "temperature": 0.0,
     "max_tokens": 32768,
     "stream": true,
     "response_format": {
       "type": "json_schema",
       "json_schema": { "name": "invoice", "schema": { /* any object schema */ }, "strict": true }
     }
   }
   ```

3. Watch the raw SSE stream:

   ```
   data: {... "delta":{"reasoning_content":"\n{\n"}}          <-- the JSON itself, classified as thinking
   data: {... "delta":{"reasoning_content":"  \"vendor\""}}
   ... (hundreds of reasoning_content chunks, zero content chunks) ...
   data: {... "delta":{"content":"\n{\n  \"vendor\": ..."}}   <-- full answer re-emitted once, at EOS
   data: {... "delta":{},"finish_reason":"stop"}
   ```

4. See error: no `delta.content` until end-of-stream; streaming clients time out on first content.

**Expected behavior**

The JSON streams as `delta.content` from the first token — it *is* the answer, not reasoning. `reasoning_content` should only carry actual thinking text.

**Screenshots**

n/a — the raw SSE excerpt above shows the misclassification.

- macOS Version: 26.5.2 (25F84)
- oMLX Version: 0.6.0 (first seen on a 0.5.0 dev build; code path verified unchanged on current `main`)

**Additional context**

Root cause is a chain of four individually-reasonable behaviors:

1. For grammar-constrained requests with a reasoning parser, the server force-enables the thinking template and auto-sets a thinking budget "so the model exits the reasoning phase and the grammar can activate" (`omlx/server.py:3684`). The rendered prompt now ends on `<think>`.
2. `prompt_opens_thinking(...)` therefore constructs `ThinkingParser(start_in_thinking=True)` (`omlx/server.py:4591`, `:5037`, `:6381` — since #2623 all three streaming paths do this).
3. The grammar is a **triggered** structural tag `[tag(<think>, any_text, </think>), schema]` — the tag is optional, so a model that goes straight into the schema without ever emitting `<think>…</think>` is perfectly grammar-valid.
4. `ThinkingParser` only ever leaves thinking mode on a literal `</think>` in the text stream. If the model skips thinking, that tag never arrives: every delta is classified as reasoning, and only the `finish()` recovery (`omlx/api/thinking.py:326`) re-emits the accumulated text as content — silently, in one chunk, at EOS.

So the server *forces* the thinking phase open for the grammar's sake, the grammar *permits* skipping it, and the text-level parser is the only component that never learns the model took that exit.

*Suggested fix:* make `ThinkingParser` grammar-aware — when the request carries a compiled grammar, decide on the **first non-whitespace character** of the response. A schema opener (`{` / `[`) means the model skipped thinking → flip to content immediately (sticky, so a synthetic `<think>\n` prefix sharing the chunk cannot re-enter thinking). Anything else locks normal thinking ha
ndling, so genuine reasoning → `</think>` → schema flows stay untouched. The `finish()` recovery should also log a warning — it currently re-classifies the entire response silently. We run this fix in production; it takes the repro from "0 content deltas until EOS" to normal live streaming (TTFB 3–14 s on 6 concurrent requests, all deltas as `content`, recovery never triggered). H
appy to submit the PR.

*Related finding (separate, smaller):* while bisecting this we also found that `GrammarConstraintProcessor`'s top-of-step accept (`omlx/scheduler.py:714-717`) silently drops the sampled token whenever the batch was reshaped between sampling and the next step (`extend`/`filter` break the positional mapping; `filter` leaves `_next_tokens` as `None`). The matcher then stays desynced 
for the rest of the generation with no log line — `omlx/api/grammar.py:87` receives the row's token history and could self-heal from it (the same counter pattern `ThinkingBudgetProcessor` already uses). Our PR includes that resync as well; can split it out if preferred.

stiivo:
Fixes #2744.

## What this fixes

When `response_format` (json_schema) meets a thinking model whose template opens the think block in the prompt, a model that **skips its thinking phase** (`/no_think`, or simply deciding not to think) streams its entire answer as `delta.reasoning_content`. Clients waiting for `delta.content` see nothing until end-of-stream, where `ThinkingParser.finish()`'s recovery re-emits the whole answer as one `content` chunk. Details and root-cause chain in #2744.

## Changes

**1. `ThinkingParser`: grammar-aware classification (`omlx/api/thinking.py`, `omlx/server.py`)**

New opt-in mode `content_on_schema_start`, enabled by all three streaming paths when the request carries a `compiled_grammar`. The first non-whitespace character of the response decides, once:

- schema opener (`{` / `[`) → the model skipped thinking; flip to content immediately. The decision is sticky, so a synthetic `<think>\n` prefix sharing the chunk with schema text cannot re-enter thinking mode.
- anything else → real reasoning; normal thinking handling locks in unchanged.

The probe waits while the pending text could still be a partial tag, so token-fragmented streams decide correctly. The `finish()` recovery now logs a warning — it re-classifies an entire response and previously did so silently.

**2. `GrammarConstraintProcessor`: self-healing accept (`omlx/api/grammar.py`)**

The top-of-step accept fast path bails out whenever the batch was reshaped between sampling and the next step (`extend`/`filter` break the positional mapping; `filter` leaves `_next_tokens` as `None`) — silently dropping the sampled token from the matcher, which stays desynced for the rest of the generation. `__call__` now resyncs the matcher from the row's own token history with an idempotent counter (the same pattern `ThinkingBudgetProcessor` already uses), so a missed fast-path accept heals on the next step instead of persisting. The fast path stays as the cheap common case; when nothing was missed the resync loop body never runs.

## Verification

Repro from #2744: 6 concurrent streaming chat completions, `json_schema` strict, `temperature 0`, Qwen3.6-35B-A3B-MLX-8bit, prompt ending in `/no_think`.

| | before | after |
|---|---|---|
| `delta.content` chunks during generation | 0 (all six requests) | 246–441 per request, live |
| first content delta | end-of-stream only | 3–14 s |
| `delta.reasoning_content` chunks | ~250–450 per request | 0 |
| `finish()` recovery triggered | every request | never |

Genuine-thinking flows verified unaffected (prose → `</think>` → schema splits exactly as before; `content_on_schema_start=False` preserves the old behavior byte-for-byte). Also unit-tested the parser against split chunks, single-chunk tag+schema, token-level fragmentation, and the flag-off path.

Happy to split the `GrammarConstraintProcessor` change into its own PR if you prefer.

### DIFF
--- a/omlx/api/grammar.py
+++ b/omlx/api/grammar.py
@@ -79,6 +79,13 @@ class GrammarConstraintProcessor:
         self._terminated = False
         self._first_call = True
         self._pending = False
+        # Count of history entries (prompt + generated) this matcher has
+        # already consumed.  ``None`` until the first ``__call__`` records the
+        # prompt-length baseline.  Both accept paths — the top-of-step fast
+        # path and the ``__call__`` history resync below — advance it in
+        # lockstep so every generated token is accepted exactly once (same
+        # counter pattern as ``ThinkingBudgetProcessor``).
+        self._accepted_up_to: Optional[int] = None
 
     # ------------------------------------------------------------------
     # Per-request mode (original interface)
@@ -87,14 +94,38 @@ class GrammarConstraintProcessor:
     def __call__(self, tokens, logits: mx.array) -> mx.array:
         """Fill bitmask and apply to logits.
 
-        Accept is handled by the monkey-patched GenerationBatch._step(),
-        which is the only place that can see which token was sampled from
-        the result.  This method only fills the bitmask, applies it, and
-        records that a token is now in flight for this row (see
-        :attr:`pending`).
+        Accept normally happens in the monkey-patched
+        ``GenerationBatch._step()`` top-of-step fast path, which is the
+        cheapest place to read the sampled ids.  That path bails out
+        whenever the batch was reshaped between sampling and the next step
+        (``extend``/``filter`` break the positional mapping, ``filter``
+        leaves ``_next_tokens`` as ``None``) — silently dropping the
+        sampled token from the matcher and desyncing it for the rest of
+        the generation.  With a structural-tag grammar
+        (``[tag(<think>, any_text, </think>), schema]``) that desync made
+        the matcher skip the ``</think>`` transition: the whole answer
+        streamed as ``reasoning_content`` while staying schema-valid
+        server-side (observed 2026-08-15 with 6 concurrent json_schema
+        requests — the longest-running row lost its close tag).
+
+        The row's own ``tokens`` history is ground truth, so before
+        filling the bitmask we feed the matcher exactly the tokens it has
+        not seen yet.  The counter keeps this idempotent with the fast
+        path; when nothing was missed the loop body never runs.
         """
         if self._terminated:
             return logits
+
+        n = len(tokens)
+        if getattr(self, "_accepted_up_to", None) is None:
+            self._accepted_up_to = n  # first call: history is just the prompt
+        elif n > self._accepted_up_to:
+            for i in range(self._accepted_up_to, n):
+                self._advance_matcher(int(tokens[i]))
+            self._accepted_up_to = n
+            self._pending = False
+            if self._terminated:
+                return logits
 
         self._pending = True
         self._bitmask.fill(-1)
@@ -103,15 +134,31 @@ class GrammarConstraintProcessor:
         mx_bitmask = mx.array(self._bitmask)
         return self._apply_mask(mx_bitmask, logits, self._vocab_size)
 
-    def accept_token(self, token_id: int) -> None:
-        """Accept a generated token to advance matcher state."""
-        self._pending = False
+    def _advance_matcher(self, token_id: int) -> None:
+        """Feed one token to the matcher, tracking termination."""
         if self._terminated:
             return
         if not self._matcher.accept_token(token_id):
             logger.warning("GrammarMatcher rejected token %d", token_id)
         if self._matcher.is_terminated():
             self._terminated = True
+
+    def accept_token(self, token_id: int) -> None:
+        """Accept a generated token to advance matcher state (fast path).
+
+        Accepts unconditionally — the counter only records the position when
+        a ``__call__`` has established the history baseline.  Without one
+        (never happens in the real flow, but the row-advance tests build bare
+        instances via ``__new__``), a later ``__call__`` initialises its
+        baseline from the full history, which already includes this token,
+        so nothing is double-accepted either way.
+        """
+        self._pending = False
+        if self._terminated:
+            return
+        self._advance_matcher(token_id)
+        if getattr(self, "_accepted_up_to", None) is not None:
+            self._accepted_up_to += 1
 
     @property
     def pending(self) -> bool:
--- a/omlx/api/thinking.py
+++ b/omlx/api/thinking.py
@@ -9,9 +9,12 @@ Used by reasoning models like DeepSeek R1, Qwen3/3.5, MiniMax that wrap
 their chain-of-thought reasoning in <think>...</think> tags.
 """
 
+import logging
 import re
 from collections.abc import Callable, Sequence
 from typing import List, Optional, Tuple
+
+logger = logging.getLogger(__name__)
 
 # Tags used for thinking blocks
 _OPEN_TAG = "<think>"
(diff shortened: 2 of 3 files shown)


jundot:
Thanks for the detailed repro and for identifying the matcher desynchronization after batch reshaping. I don’t think `content_on_schema_start` is safe to merge because `{` or `[` is not a reliable phase boundary; it can also begin genuine reasoning, and the heuristic does not cover scalar JSON or non-JSON grammars.

The Qwen structural grammar requires `reasoning -> </think> -> schema`, so bare JSON should not be able to complete or reach EOS while the grammar matcher remains active. Could you split the `GrammarConstraintProcessor` resync into a separate PR first? I’d like to verify whether keeping the matcher synchronized across `extend`/`filter` resolves #2744 without adding parser heuristics.

If the issue remains after that, I’d prefer an explicit grammar/protocol phase signal rather than inferring the phase from output characters.

stiivo:
Thanks for the review — you were right about the grammar shape, and partially right about the mechanism. Data below.

**Correction to my issue text:** I claimed the structural tag is triggered/optional. It is not — `xgr.get_builtin_structural_tag('qwen_3_5', reasoning=True)` yields a required sequence:

```json
{"type": "sequence", "elements": [
  {"type": "sequence", "elements": [
    {"type": "tag", "begin": "", "content": {"type": "any_text"}, "end": "</think>"},
    {"type": "const_string", "value": "\n\n"}]},
  {"type": "any_text", "excludes": ["<think>", "</think>"]}]}
```

**Split done:** the matcher resync is now #2774, standalone.

**A/B result you asked for** (same repro as #2744: 6 concurrent streaming `json_schema` requests, Qwen3.6-35B-A3B-MLX-8bit, prompt ends in `/no_think`, temp 0):

| configuration | result |
|---|---|
| stock 0.6.0 | all deltas `reasoning_content`, content only via `finish()` recovery at EOS |
| stock + #2774 resync only | **identical failure** — 441 `reasoning_content` deltas, no `content` delta until the final recovery chunk, no `</think>` anywhere in the text stream, `finish_reason=stop` at ~1.5k tokens |
| stock + #2774 + parser change from this PR | all deltas stream as `content`, recovery never fires |

So the resync is a real correctness fix but does not resolve #2744 — and the middle row contradicts the sequence-grammar expectation in a way that seems worth your attention independently: with a synchronized matcher and a grammar that requires `reasoning → </think> → \n\n → schema`, the model still produced bare JSON and reached EOS at its natural end. Either the bitmask is not being enforced on this path (fallback to prompt injection despite the attached processor?), or the matcher reaches an accepting state without consuming `</think>`. Happy to add instrumentation if you tell me where you'd like it.

**On the fix direction:** agreed that first-character sniffing is a heuristic and an explicit phase signal is better. Concrete proposal: the `GrammarConstraintProcessor` is the one component that structurally knows the phase — it could watch the accepted token stream for the tag's end sequence and expose an `in_output_phase` property, which the server-side stream loop consults instead of scanning text for `</think>`. That keeps the parser dumb, covers scalar JSON and non-JSON grammars, and works even when the model never emits the tag (phase flips when the matcher's schema element starts matching). If that direction works for you, I'll rework this PR accordingly.

Issue #2740, opened 2026-08-17
[Bug] 0.6.0: reasoning_effort from external client is passed through to chat template without validation, causing 400 on Qwen3.8


Githantao:
[Bug] 0.6.0：外部客户端传入的 reasoning_effort 未经校验直接透传给 chat template，导致 Qwen3 返回 400

Describe the bug

When an external client (e.g., an agent application) sends reasoning_effort: "high" in a chat completion request, oMLX 0.6.0 passes this value directly to the model's chat template. For Qwen3, the chat template only accepts xhigh (default), medium, and low — the value high is not valid, resulting in a 400 error:

Chat template error: Unexpected reasoning effort high.
Supported types are xhigh (default), medium, and low.
Copy
In oMLX 0.5.7, this parameter was not forwarded to the chat template, so the model used its native default (xhigh) and the request succeeded. In 0.6.0, the parameter is now forwarded but without any validation or normalization, so an incompatible value from the client causes a hard 400 instead of being mapped to the closest valid value.

When an external client (e.g., an agent application) sends reasoning_effort: "high" in a chat completion request, oMLX 0.6.0 passes this value directly to the model's chat template. For Qwen3, the chat template only accepts xhigh (default), medium, and low — the value high is not valid, resulting in a 400 error.

In oMLX 0.5.7, this parameter was not forwarded to the chat template, so the model used its native default (xhigh) and the request succeeded. In 0.6.0, the parameter is now forwarded but without any validation or normalization, so an incompatible value from the client causes a hard 400 instead of being mapped to the closest valid value.

当外部客户端（如 agent 应用）在 chat 请求中发送 reasoning_effort: "high" 时，oMLX 0.6.0 将该值直接透传给模型的 chat template。对于 Qwen3，chat template 仅接受 xhigh（默认）、medium、low，high 不是合法值，导致 400 错误。

在 oMLX 0.5.7 中，该参数不会被转发给 chat template，模型使用原生默认值 xhigh，请求正常。0.6.0 中该参数开始被转发，但没有做任何校验或归一化，客户端传入的不兼容值直接导致 400，而非映射到最接近的合法值。

To Reproduce

Run oMLX 0.6.0 (CLI) with model Qwen3.8-27B-mxfp8
Send a POST /v1/chat/completions request that includes "reasoning_effort": "high" in the request body (e.g., via an agent application's "default thinking level = medium" setting, which internally maps to the string high)
Observe the 400 response:
[...]

PR #2746, opened 2026-08-17
feat(discovery): expose reasoning_effort contract in model status

monroewilliams:
## Context

Reasoning-effort handling has evolved across #2653 (top-level `reasoning_effort` request field), #2724 (DeepSeek V4 alias mapping), and #2740 (engine-level alias fallback + native-default retry). That leaves a gap on the *discovery* side: a client (e.g. a coding agent with a thinking-level selector) has no way to learn which effort vocabulary a model's chat template actually accepts. It can only guess — and with #2740's best-effort normalization, wrong guesses are silently remapped (e.g. `high` → `xhigh` on Qwen 3.8) instead of rejected.

## What this PR adds

**1. Discovery — `detect_reasoning_effort()` (`omlx/model_discovery.py`)**

Reads the model's chat template and extracts its `reasoning_effort` contract:

- `options` — only when the template enforces a strict tuple-literal whitelist (e.g. Qwen 3.8: `('xhigh', 'medium', 'low')`). Free-form, dict-map, and numeric templates (gpt-oss, Inkling, GLM) report `None` so clients pass values through and let #2740's normalization handle them.
- `default` — from `|default('X')` filters or `is not defined -> set` blocks (string defaults only).

Exposed as `reasoning_effort_options` / `reasoning_effort_default` on `DiscoveredModel` / `EngineEntry`, surfaced on `/v1/models/status` next to `thinking_default`.

**2. `"none"` → thinking off (`omlx/api/utils.py`)**

`merge_reasoning_effort_chat_template_kwargs()` now translates `reasoning_effort: "none"` to `enable_thinking: False` at request-merge time instead of forwarding it. `"none"` is the de facto off token (OpenAI, vLLM), but it is not in strict whitelists — and without this translation, #2740's alias fallback would map `"none"` → `"low"`, i.e. thinking *on* at low effort, which is the wrong semantics for an off switch. An explicit `enable_thinking` in the request's `chat_template_kwargs` still wins (`setdefault`). Covers both the chat-completions field and responses `reasoning.effort`, which share the helper.

Together: clients can map their own thinking levels to values the template accepts exactly (no silent remapping), and express thinking-off through the standard field.

## Validation

- Detector validated against real templates: Qwen 3.8 → `(['xhigh', 'medium', 'low'], 'xhigh')`; gpt-oss → `(None, 'medium')`; GLM-5.2 / Inkling / Muse-Glimmer → `(None, None)`.
- Tests: detection (`tests/test_thinking_toggle.py`), `"none"` merge semantics (`tests/test_api_utils.py`); full suite passes.
- Live check against a running server: status fields present for all discovered models; `reasoning_effort: "low"` → thinking, `"none"` → no thinking (1-token answer).

## End-to-end consumer

[pi-local](https://github.com/monroewilliams/pi-local) (pi extension for local inference servers) reads the new status fields, builds a per-model `thinkingLevelMap` (advertised levels map to themselves, the rest are hidden, `off` → `"none"`), and sends the standard top-level field — verified end-to-end on Qwen 3.8, where the selector offers exactly `off / low / medium / xhigh`.

## Scope notes

- No request-side pre-validation: #2740 already normalizes unknown values best-effort.
- Alternate parameter names (e.g. Muse Glimmer's `reasoning_strength`) are out of scope; the `chat_template_kwargs` escape hatch covers them.

### DIFF
--- a/omlx/api/utils.py
+++ b/omlx/api/utils.py
@@ -49,10 +49,19 @@ def merge_reasoning_effort_chat_template_kwargs(
     chat_template_kwargs: dict[str, Any] | None,
     reasoning_effort: Any | None,
 ) -> dict[str, Any] | None:
-    """Forward an API reasoning effort without overriding explicit template kwargs."""
+    """Forward an API reasoning effort without overriding explicit template kwargs.
+
+    ``"none"`` is a client-side off switch: it turns thinking off via
+    ``enable_thinking=False`` instead of reaching the template, whose
+    whitelists (e.g. Qwen 3.8's ``xhigh/medium/low``) do not include it.
+    An explicit ``enable_thinking`` in the request kwargs still wins.
+    """
     merged = dict(chat_template_kwargs or {})
     if reasoning_effort is not None:
-        merged.setdefault("reasoning_effort", reasoning_effort)
+        if reasoning_effort == "none":
+            merged.setdefault("enable_thinking", False)
+        else:
+            merged.setdefault("reasoning_effort", reasoning_effort)
     return merged or None
 
 
--- a/omlx/engine_pool.py
+++ b/omlx/engine_pool.py
@@ -210,6 +210,12 @@ class EngineEntry:
     preserve_thinking_default: bool | None = (
         None  # True when template supports preserve_thinking (Qwen 3.6+)
     )
+    reasoning_effort_options: list[str] | None = (
+        None  # Strict whitelist from template (None = free-form or unsupported)
+    )
+    reasoning_effort_default: str | None = (
+        None  # Template default for reasoning_effort (None = unknown)
+    )
     model_context_length: int | None = (
         None  # Declared context length from config.json (None if unknown)
     )
@@ -832,6 +838,12 @@ class EnginePool:
                     thinking_default=getattr(info, "thinking_default", None),
                     preserve_thinking_default=getattr(
                         info, "preserve_thinking_default", None
+                    ),
+                    reasoning_effort_options=getattr(
+                        info, "reasoning_effort_options", None
+                    ),
+                    reasoning_effort_default=getattr(
+                        info, "reasoning_effort_default", None
                     ),
                     model_context_length=getattr(info, "model_context_length", None),
                     source_type=getattr(info, "source_type", "local"),
@@ -3171,6 +3183,8 @@ class EnginePool:
                     "is_helper": e.is_helper,
                     "thinking_default": e.thinking_default,
                     "preserve_thinking_default": e.preserve_thinking_default,
+                    "reasoning_effort_options": e.reasoning_effort_options,
+                    "reasoning_effort_default": e.reasoning_effort_default,
                     "source_type": e.source_type,
                     "source_repo_id": e.source_repo_id,
                     "last_access": e.last_access if e.last_access > 0 else None,
--- a/omlx/engine_pool.py
+++ b/omlx/engine_pool.py
@@ -211,7 +211,7 @@ class EngineEntry:
         None  # True when template supports preserve_thinking (Qwen 3.6+)
     )
     reasoning_effort_options: list[str] | None = (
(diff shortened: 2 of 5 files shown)


cebertowicz:
Hi @monroewilliams — I opened #3395 building on your detector commit (included as-is, you stay author). It adds the pieces the on-disk template can't show: DeepSeek V4's oMLX-patched template declaring its own vocabulary, and the same fields on the admin API. I left the `none` mapping out since b876be8e already settled that one. Thanks for the groundwork; happy to rebase onto yours if it lands first.

monroewilliams:
> Hi @monroewilliams — I opened #3395 building on your detector commit (included as-is, you stay author). It adds the pieces the on-disk template can't show: DeepSeek V4's oMLX-patched template declaring its own vocabulary, and the same fields on the admin API. I left the `none` mapping out since [b876be8](https://github.com/jundot/omlx/commit/b876be8ec5a1c19dab3f77a942c7ef8eddd9d83a) already settled that one. Thanks for the groundwork; happy to rebase onto yours if it lands first.

I'd actually just been in the process of rebasing this to the tip of main, and also reworking and generalizing the detector, and was about to push some changes when I saw your comment just now. I had been thinking I would close this PR and split it into two: one with just the detector/reasoning_effort_options changes squashed to a single new commit, and the other with the "none" handling. 

I hadn't noticed the "off"/"none" mapping in  https://github.com/jundot/omlx/commit/b876be8ec5a1c19dab3f77a942c7ef8eddd9d83a yet, and I'd actually argue that mapping "off"/"none" -> "low" is not the right thing to do -- it gives a harness that expects to be able to disable thinking entirely by setting "none" (which is what pi does natively, and is also the semantics that llama.cpp associates with "none") no way to do that directly. My [pi-local plugin](https://github.com/monroewilliams/pi-local) has to work around this by special-casing the "off" value and taking a different path that uses `chat_template_kwargs` for that case instead of just passing through reasoning_effort of "none". Interpreting a thinking level of "none" as "disable thinking" within omlx would solve this entirely at (IMO) the appropriate level, and would mirror what llama.cpp does.

I'll go ahead and push the rebase with the additional changes I made today as another commit here instead of doing an amend/split like I'd planned. I need to head to bed right now, but I'd like to discuss this more with you and maybe @jundot if he's not completely slammed. :)

maxpfingsthorn:
+1 for the `"none"` → `enable_thinking: false` part of this PR. Here is some current data from HEAD, in case it helps get that piece in on its own.

**Setup:** Homebrew HEAD `4e5308c`, `Jundot/Qwen3.8-27B-oQ4e-mtp`, non-streaming, `temperature=0`, prompt "What is 17*23? Answer with the number only."

| Request | completion tokens | `reasoning_content` | time |
|---|---|---|---|
| no thinking params | 65 | 142 chars | 2.42 s |
| `reasoning_effort: "none"` | 61 | 114 chars | 1.89 s |
| `reasoning_effort: "low"` | 61 | 114 chars | 1.90 s |
| `chat_template_kwargs: {enable_thinking: false}` | 3 | empty | 0.56 s |

So on main, `"none"` is identical to `"low"` (via `_ALIAS_FALLBACKS` in `omlx/reasoning_effort.py`): the client asks for no reasoning and still pays for it.

We hit this from two clients: an OpenAI-SDK image-labeling script that sets `reasoning_effort="none"` and got chain-of-thought (and, with a tight `max_tokens`, truncated reasoning, see #3796), and pi, which also sends `"none"` for its off level. Both only work today by switching to the non-standard `chat_template_kwargs` path, which generic OpenAI clients can't do.

Two small suggestions for the mapping, if it gets split out as you mentioned:
- treat `"off"` the same as `"none"` (both are already in `_STRING_LEVELS`), and match after `.strip().lower()` like `_normalized_input` does;
- apply it in the shared helper, so Responses `reasoning.effort` is covered too (it already is in ebba9b2b).

Happy to open that split-out PR on current main, with you as co-author, if that's easier than rebasing this one.

PR #2753, opened 2026-08-17
Fix preserve_mid_system_cache between tool calls

### LINKED ISSUE #2483: preserve_mid_system_cache never engages for Claude Code: injected system nudges are hoisted into the leading system block, breaking the prefix every turn
**Describe the bug**

Claude Code injects a `role: "system"` message into the `messages` array roughly every 7 messages (the "task tools haven't been used recently" nudge plus a dump of the current todo list, ~980 chars ≈ 269 tokens). oMLX consolidates all of these into the leading system block, so the front of the prompt grows by ~269 tokens every time one is added. Everything behind it shifts, and the prefix cache is invalidated at the system/first-user boundary on every turn.

`preserve_mid_system_cache: true` is supposed to prevent exactly this, but it never engages for Claude Code, because both gates guarding the cache-friendly path reject the placement Claude Code always produces.

This is a different path from #2333 (LSP tool + microcompact, both client-side) and #2384 (snapshot growth, resolved as genuine divergence). Here the request payload is verified append-only and the re-prefill is caused server-side.

**Evidence**

The divergence diagnostic pins it exactly:

```
prefix cache: request 87407133 re-prefills 36208 of 58736 tokens (reused 22528);
closest stored sequence bb3ea5aa shares the first 23128 of 55296 comparable tokens before diverging

first divergence at token 23128:
  stored=...' tests\n#14. [pending] Final verification run<|im_end|>\n<|im_start|>user\n<system-reminder>\nAs'
  vs prompt=..." tests\n#14. [pending] Final verification run\n\nThe task tools haven't been used recently. If you"
```

In the cached sequence the system block ends after the 5th nudge and a user turn begins. In the new prompt a 6th nudge extends the same block. Divergence lands precisely at the system/first-user boundary.

The divergence point advances by exactly 269 tokens per turn across six consecutive requests — one nudge each:

```
21783 → 22052 → 22321 → 22590 → 22859 → 23128
```

while re-prefill grows with the conversation:

```
20678 of 35014  (reused 14336)
17274 of 37754  (reused 20480)
21717 of 42197  (reused 20480)
26080 of 46560  (reused 20480)
28818 of 51346  (reused 22528)
32830 of 55358  (reused 22528)
36208 of 58736  (reused 22528)
```

**The request payload is append-only.** I dumped the 22 TRACE-logged `Incoming POST /v1/messages` bodies and diffed consecutive pairs after stripping `cache_control` (which only feeds `request_has_cache_control()` and is never rendered):

- 26 tools, byte-identical across every request (no LSP, no MCP, no plugin-provided tools)
- `system` field identical, 8437 chars, same md5 throughout
- messages: first differing index is the tail of the shorter request in 4 of the last 5 pairs — pure append

**Root cause**

The model's chat template (`KAT-Coder-V2.5-Dev-oQ5e`) supports `role: "tool"`, so `_chat_template_supports_tool_role()` returns True and `convert_anthropic_to_internal()` converts `tool_result` user messages into `role: "tool"` messages. Every injected nudge is therefore preceded by a `tool` message rather than a `user` one.

That fails the first gate in `_mid_system_placement_kinds()` (`omlx/api/utils.py:482`):

```python
prev_role = messages[start - 1].get("role") if start > 0 else None
next_role = messages[i].get("role") if i < len(messages) else None
if prev_role != "user":
    return None          # ← tool-result predecessor, always taken with Claude Code
```

`None` sends `prepare_system_messages_for_template()` into `unsupported_fallback()`. With `preserve_mid_system_cache: true` that tries `_downgrade_mid_system_to_user_notes()`, which hits the same blind spot from the other side:

```python
if _is_safe_user_note_target(rewritten[-1]) and (next_msg is None or next_role == "assistant"):
    ...          # prev is role "tool" → False
if _is_safe_user_note_target(next_msg):
    ...          # next is an assistant message → False
return None      # → strict_fallback() → _consolidate_system_messages() → hoisted to front
```

Confirmed the `[System note]` wrapper from `_format_system_note()` appears nowhere in the rendered prompt, so the downgrade path definitely did not run.

Every injected nudge in my capture sits between a `tool_result`-only user message and an assistant `tool_use` message:

```
nudge@3   prev  2 user ['tool_result' x3]   next  4 assistant ['thinking','tool_use' x3]
nudge@8   prev  7 user ['tool_result' x4]   next  9 assistant ['thinking','tool_use' x3]
nudge@15  prev 14 user ['tool_result' x3]   next 16 assistant ['thinking','tool_use' x3]
nudge@24  prev 23 user ['tool_result' x4]   next 25 assistant ['thinking','tool_use' x3]
nudge@31  prev 30 user ['tool_result' x4]   next 32 assistant ['thinking','tool_use' x3]
nudge@38  prev 37 user ['tool_result' x4]   next 39 assistant ['thinking','tool_use' x3]
nudge@45  prev 44 user ['tool_result' x3]   next: none (nudge is last)
```

7 of 7. Claude Code emits the nudge at a tool-result boundary every time, so on any model whose template supports the `tool` role the cache-preserving path is unreachable.

**To Reproduce**

1. Run Claude Code against oMLX with `preserve_mid_system_cache: true` (the default) and a model whose chat template supports `role: "tool"`
2. Work through a long tool-heavy session so the task-tools nudge fires
3. Watch `first divergence at token N` advance by ~269 per nudge while the payload stays append-only

**Expected behavior**

Mid-conversation system messages arriving after a tool result should stay in place rather than being hoisted into the leading system block, so appending one does not shift the whole conversation.

**Possible fix**

Accepting a `tool` predecessor in `_mid_system_placement_kinds()` looks sufficient:

```python
if prev_role not in ("user", "tool"):
    return None
```

The placement is still `between` (next role is `assistant`), and `chat_template_preserves_mid_system()` continues to guard correctness — templates that cannot render a mid-conversation system message in place still fall back. The KAT template renders it in place:

```jinja
{%- if message.role == "system" %}
    {%- if not loop.first %}
        {{- '<|im_start|>system\n' + content + '<|im_end|>' + '\n' }}
    {%- endif %}
```

The `_downgrade_mid_system_to_user_notes()` blind spot is worth closing too, as the fallback for templates that fail the probe.

**Desktop**

- macOS 26.5.2, Apple M3 Max, 36 GB
- oMLX 0.5.4 (build 260803001602-macos26-27)
- Claude Code 2.1.212
- Model: KAT-Coder-V2.5-Dev-oQ5e
- `preserve_mid_system_cache: true`, no dflash, no MTP, no specprefill, no MCP servers, no LSP

q-p:
Fixes https://github.com/jundot/omlx/issues/2483

### DIFF
--- a/omlx/api/utils.py
+++ b/omlx/api/utils.py
@@ -500,7 +500,7 @@ def _mid_system_placement_kinds(messages: list[dict]) -> set[str] | None:
 
         prev_role = messages[start - 1].get("role") if start > 0 else None
         next_role = messages[i].get("role") if i < len(messages) else None
-        if prev_role != "user":
+        if prev_role not in ("user", "tool"):
             return None
         if next_role is None:
             placements.add("tail")


jundot:
Thanks for tracking this down. Accepting a tool-result predecessor restores the cache-preserving path for Claude Code's `tool -> system -> assistant` reminders, and the live reuse data confirms the intended behavior. I merged this and followed up on main by making the capability probe validate the actual tool predecessor and adding regression coverage for between and tail placements. Thanks again for the focused report and fix.

PR #2754, opened 2026-08-17
fix(vlm): normalize Qwen3.8 patch embedding layout

frank-beans:
## What changed

Normalize channels-first Qwen3.8 vision patch-embedding weights while loading MLX-format checkpoints.

Qwen3.8 uses the internal MLX-VLM model types `qwen3_5`/`qwen3_5_moe`, which are also used by Qwen3.6. The fix is limited to those internal types, `format=mlx`, and the single `vision_tower.patch_embed.proj.weight` tensor when it demonstrably has channels-first layout. It applies mlx-vlm's own `transpose(0, 2, 3, 4, 1)` conversion in memory without modifying checkpoint files.

## Root cause

mlx-vlm skips sanitizers for MLX-format safetensors, but some converted Qwen3.8 checkpoints retain PyTorch Conv3d layout. Strict VLM loading then fails with:

```
Expected shape (1152, 2, 16, 16, 3) but received shape
(1152, 3, 2, 16, 16) for parameter
vision_tower.patch_embed.proj.weight
```

oMLX subsequently falls back to a text-only engine. The correction runs at the safetensors-loading stage so it also covers the model-specific `load_weights` path installed by VLM MTP.

## User impact

Affected Qwen3.6/Qwen3.8 MLX checkpoints load as full VLMs instead of silently losing vision. Native MTP remains attached and usable.

## Validation

- `ruff check tests/test_qwen35_mlx_vision_layout.py`
- `pytest -q tests/test_qwen35_mlx_vision_layout.py tests/test_vlm_audio_fallback.py` — 20 passed
- Real `Youssofal/Qwen3.8-27B-MTPLX-Optimized-Speed` production-path load succeeded as `mlx_vlm.models.qwen3_5.qwen3_5`
- Vision tensor loaded as `(1152, 2, 16, 16, 3)`
- MTP module remained attached

The `qwen3_5` references in code and tests intentionally match the internal configuration identifier used by Qwen3.8.

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -841,6 +841,56 @@ def _drop_gemma4_mlx_shared_kv_extras_on_load(model_dir: Path):
             )
 
 
+@contextlib.contextmanager
+def _transpose_qwen35_mlx_vision_patch_embed_on_load(model_dir: Path):
+    """Fix channels-first Qwen3.5 vision weights in MLX checkpoints.
+
+    mlx-vlm skips model sanitizers when safetensors metadata declares
+    ``format=mlx``. Some converted Qwen3.5/3.6 checkpoints retain the PyTorch
+    Conv3d layout ``(out, in, time, height, width)`` for the vision patch
+    embedding, while MLX expects ``(out, time, height, width, in)``. Correct
+    only that unambiguously channels-first tensor during loading.
+    """
+    if _read_config_model_type(model_dir) not in {"qwen3_5", "qwen3_5_moe"}:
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the focused fix. I verified that the load-time hook is limited to MLX-format Qwen3.5/Qwen3.5-MoE checkpoints and only transposes the unambiguous channels-first patch embedding while preserving already-correct layouts. The relevant VLM, MTP, and mixed-quantization tests pass. LGTM, merging.

PR #2756, opened 2026-08-17
feat(qwen3.5): add experimental dual-ANE/GPU prompt processing

onthehub97:
# Experimental dual-ANE/GPU prompt processing for Qwen 3.5/3.6/3.8

## Summary

This PR adds an experimental hybrid Apple Neural Engine/Metal prompt-processing path for dense Qwen 3.5/3.6 models.

For eligible fixed-size prefill blocks:

- Two ANE procedures compute disjoint output-channel prefixes, one pinned to each physical ANE.
- Metal computes the remaining quantized suffix concurrently.
- Native merge kernels reconstruct the output and apply the following activation where possible.
- Unsupported shapes, layers, quantizations, decode calls, and verification calls retain the existing GPU path.

The feature is disabled by default and configurable per model through both the macOS application and web administration UI.

### Hybrid ANE/GPU MLP processing

- Added private Apple Neural Engine runtime integration to the Qwen custom-kernel extension.
- Splits eligible Qwen gate/up projections across both ANEs and the GPU.
- Combines gate and up into one procedure per ANE operation.
- Fuses the ANE/GPU output merge with SwiGLU, avoiding full gate/up materialization and additional concatenations.
- Keeps the original q4 tensors and GPU path available for decode, verification, tails, and unsupported inputs.
- Leaves the down projection on the GPU.

### GDN acceleration

- Added hybrid execution for the GDN `z + qkv` input projections.
- Keeps the recurrent update, `b/a` projections, normalization, and output projection on the GPU.
- Added dispatch support for both `mlx-vlm` and the `mlx-lm` fallback used by the tested model.
- Made the `mlx-lm` hook resilient to model unload/reload and MTP lifecycle code replacing `GatedDeltaNet.__call__`.

### Dual-ANE procedure banks

- Packages all accelerated layer slices as procedure indices within two resident programs—one per physical ANE.
- The measured 64-MLP/48-GDN layout exposes 112 procedures from two resident program handles.
- Each accelerated operation issues two ANE evaluations in parallel while the GPU suffix runs concurrently.
- Retains a legacy per-slice compilation fallback for extensions without procedure-bank support.

### Eager preparation

- Compiles and prepares eligible ANE procedures during model loading.
- Moves compilation cost out of the first eligible request.
- Treats the ANE configuration as part of the engine runtime signature, so changing it reloads a resident model.

### Configuration and UI

Added the following per-model settings:

```json
{
  "qwen35_ane_prefill_enabled": true,
  "qwen35_ane_prefill_sequence_length": 2048,
  "qwen35_ane_prefill_fraction": 0.53,
  "qwen35_ane_prefill_max_layers": 64,
  "qwen35_ane_prefill_dual_ane": true,
  "qwen35_ane_prefill_gdn": true,
  "qwen35_ane_prefill_gdn_fraction": 0.50,
  "qwen35_ane_prefill_gdn_max_layers": 48
}
```

These controls are exposed through:

- The native macOS per-model settings screen.
- The web per-model settings editor.
- Per-model profiles.
- The settings API and persistence layer.

The controls are shown only for detected Qwen 3.5/3.6 models. The feature remains opt-in.

A UI formatting bug that caused an explicitly saved GDN fraction of `0.50` to appear blank has also been fixed. Numeric values are now formatted to match the picker's canonical values.

### Benchmarking and instrumentation

- Added a standalone GPU/single-ANE/dual-ANE benchmark.
- Added a full 2,048-token benchmark warm-up option to the macOS app.
- Added opt-in scheduler chunk traces and native phase counters.
- Logs distinguish eligible fixed-shape calls from residual GPU tails.
- Reports MLP/GDN dispatch counts, input-readiness delay, ANE evaluation time, GPU suffix time, launch gaps, and approximate duty cycle.
- Benchmark requests avoid polluting shared prefix-cache tiers.

## Measured results

Reference configuration:

- M3 Ultra.
- `True2456/Qwen3.8-27B-AWQ-4.85bpw`.
- 2,048-token fixed prefill shape.
- 53% MLP / 50% GDN ANE shares.
- 64 MLP and 48 GDN layers.

Representative controlled measurements:

| Measurement | GPU | Hybrid | Improvement |
|---|---:|---:|---:|
| Complete layer-0 MLP | 61.12 ms | 48.45 ms | 1.26× |
| Full language body, corrected MLP+GDN path | 6.1149 s | 4.5084 s | 1.356× |
| Prompt throughput | 334.9 tok/s | 454.3 tok/s | +35.6% |

End-to-end application benchmarks using a 53% MLP / 53% GDN split measured:

| Prompt length | GPU | GPU + dual ANE | Improvement |
|---|---:|---:|---:|
| 4K | 331.4 tok/s | 379.1 tok/s | +14.4% |
| 8K | 326.0 tok/s | 397.7 tok/s | +22.0% |
| 16K | 318.9 tok/s | 402.1 tok/s | +26.1% |

The percentage improvement grows with prompt length because longer prompts contain more complete fixed-shape blocks over which the eager compilation and per-request overheads are amortized.

The corrected MLP+GDN numerical comparison measured:

- Hidden-state cosine similarity: 0.999200.
- Last-token logit cosine similarity: 0.998522.
- Top-1 token: unchanged.

These are numerical sanity checks, not substitutes for downstream intelligence or task-quality evaluation.

### Two pinned evaluations instead of one unpinned evaluation

A single procedure was not effectively striped across both ANEs by the private runtime.

A representative 55% MLP slice took 57.90 ms as one unpinned call versus 41.51 ms as two instance-pinned calls. The implementation therefore performs two parallel ANE submissions per accelerated operation.

### Procedure indices instead of one program per layer

The private runtime accepted approximately 121 resident handles in testing. Allocating two standalone programs for every accelerated layer would exhaust that budget before both MLP and GDN could be covered.

Procedure banks preserve layer-specific weights while requiring only two resident programs. This is why the implementation has many procedure calls but very few program handles.

### Combined gate/up and z/qkv procedures

Gate/up and z/qkv are combined within their respective procedures. Splitting them into additional calls would increase launch count and synchronization overhead without creating useful additional ANE parallelism.

### The blocking input-pack wait is intentional

Several asynchronous alternatives were tested:

- Completion-handler ANE submission.
- Reused private completion events.
- Metal completion callbacks.
- Worker-based input packing.
- Persistent high-priority submission workers.

These either provided no improvement or delayed ANE submission until after queued GPU suffix work.

The Metal-callback approach increased a representative layer from approximately 47.5 ms to 71 ms and made the full body slower than GPU-only execution.

The current blocking wait allows ANE submission to occur while the relevant GPU suffix is still queued and was retained based on measured end-to-end performance.

### The down projection remains on GPU

An experimental split-down path produced unacceptable numerical errors. It was removed rather than retained behind a configuration flag.

The current implementation accelerates the MLP gate/up projection and fused activation but executes the down projection using the existing q4 GPU path.

### No synthetic padding

The tested Qwen padding token is an ordinary learned token. Appending it changes logits, KV positions, RoPE positions, and GDN recurrent state.

The implementation therefore accelerates only exact fixed-shape blocks and does not pad shorter inputs to force eligibility.

### Approximate ANE weights

The source q4 weights are requantized into per-output-channel INT8 weights for the ANE. This is not bit-exact relative to the original grouped q4 computation.

The original quantized weights remain resident for all GPU fallbacks and for the GPU suffix.

### Private and platform-specific API

- Uses undocumented `AppleNeuralEngine.framework` interfaces.
- May break with macOS updates.
- Physical instance identifiers and behaviour are assumed from the tested M3 Ultra configuration.
- Other Apple Silicon devices may expose a different number of usable instances or different scheduling behaviour.
- This is currently appropriate for source builds and controlled experimentation, not a guaranteed portable runtime.

### Fixed shapes

- An ANE procedure is compiled for one exact flattened token count.
- The default bank targets 2,048 tokens.
- Other sequence lengths require separate compiled banks.
- Short chunks and final tail chunks run on the GPU.
- The feature does not independently force scheduler chunk size; effective use depends on the prefill/chunking policy producing matching blocks.

### Startup and memory cost

- The combined 112-procedure bank takes approximately 37–44 seconds to compile on the reference system.
- Compilation is eager, so model loading and configuration changes take longer.
- Both input/output surfaces, requantized ANE weights, combined GPU suffixes, and original quantized tensors remain resident.
- This deliberately spends memory to avoid request-time weight preparation.

### Prompt processing only

- Decode/token generation remains on the GPU.
- Target verification and decode-shaped calls deliberately bypass the ANE path.
- The change improves prompt processing, not generation TPS.

### Remaining ANE downtime

Profiling showed each ANE executing for approximately 38.8% of total language-body time.

Submission delay itself was only tens of microseconds; most downtime came from waiting for preceding GPU work to produce the next layer input.

This dependency chain means low reported ANE utilisation cannot be recovered merely by submitting more calls. Additional calls would add overhead unless more upstream GPU work is moved or the layer boundary is restructured.

### Concurrency

The implementation accelerates one prompt across both ANE instances. It is not a multi-tenant ANE scheduler and does not currently use separate instances to process independent prompts concurrently.

### Numerical validation

- Requantization introduces small numerical differences.
- Current checks compare hidden states, final logits, and top-1 tokens.
- No comprehensive downstream evaluation has yet been performed.
- Quality should be evaluated per model before enabling the path in production.

### Model and quantization coverage

Current eligibility checks require:

- Dense Qwen 3.5/3.6 architecture.
- Affine q4 MLP linears with group size 128.
- Compatible affine q5 GDN projections with group size 64.
- Expected projection layout and dimensions.
- A matching fixed token shape.

Unsupported layers fall back safely but receive no acceleration.

## Applicability to other architectures and kernels

The general technique is applicable when an architecture has:

- Large affine projections dominating prefill time.
- Output channels that can be divided independently.
- A GPU suffix that can run concurrently with ANE work.
- A cheap merge point after both results complete.
- Stable fixed shapes that justify eager compilation.
- Enough repeated calls to amortize compilation and resident-memory cost.

Likely candidates include other dense gated MLP architectures where gate/up outputs can be partitioned and the activation can be fused into the merge.

Additional work would be required for:

- Different quantization formats or group sizes.
- Architectures with different gate/up ordering.
- Fused QKV or other projections with different output layouts.
- Single-ANE systems.
- Models whose dimensions do not align cleanly with ANE/GPU split boundaries.

The approach is less directly applicable to:

- MoE layers, where dynamic routing and per-expert weight residency can make procedure banks very large.
- Attention kernels with dynamic masks, KV-cache state, and softmax reductions.
- Recurrent/state-space updates whose state dependency prevents output-channel partitioning.
- Decode workloads, where small token counts do not amortize ANE dispatch overhead.

Procedure-index banking itself may be reusable for other architectures: it is a practical way to expose many layer-specific weight sets without consuming one private-runtime model handle per layer.

## Future hardening

- Add macOS-version and hardware capability allowlists.
- Probe physical ANE availability and instance identifiers rather than assuming the M3 Ultra layout.
- Persist and validate a private-runtime ABI/version fingerprint.
- Improve failure classification and permanently disable the path after deterministic compile/runtime incompatibility.
- Add compilation cancellation and clearer model-load progress reporting.
- Cache compiled banks where the private runtime safely permits it.
- Add memory-budget checks before preparing ANE weights and surfaces.
- Run downstream evaluation suites to quantify intelligence loss across representative tasks.
- Add repeated accuracy thresholds to CI using small compatible fixtures.
- Benchmark narrow fraction sweeps around the selected 53% MLP and 50% GDN defaults on additional hardware.
- Explore moving additional dependency-producing work to ANE or fusing across a larger layer boundary.
- Investigate safe dynamic-shape or multi-bank selection without multiplying startup cost excessively.
- Explore prompt-level ANE scheduling only if the server gains concurrent prefill admission.
- Add structured runtime health and utilisation metrics beyond benchmark-only logging.
- Generalise projection extraction and merge descriptions so additional architectures can register compatible ANE backends without Qwen-specific hooks.

## Validation performed

- Focused ANE and Qwen q4 patch tests.
- API settings validation and persistence tests.
- Per-model profile round-trip tests.
- Native and web UI state/serialization tests.
- JavaScript syntax and Jinja template compilation.
- Native custom-kernel ABI verification against the bundled MLX.
- Repeated model unload/reload tests confirming GDN dispatch remains installed.
- GPU-versus-hybrid hidden-state/logit comparisons.
- Signed macOS application build and end-to-end in-app throughput benchmarks.

## Review notes

The feature is intentionally:

- Experimental.
- Disabled by default.
- Strictly eligibility-gated.
- Fail-open to the existing GPU path.
- Scoped to prompt processing.
- Documented as relying on private Apple APIs.


onthehub97:
Some more notes:

- Tested bare with mlx-vlm, tested in-app with mlx-lm. Mlx-vlm improvements may well be greater than the results shown in the table - the model used for testing was throwing MTP header issues which was preventing loading into mlx-vlm
- Practically ~50 tokens/s less than is theoretically possible, further experimentation with a live app is needed.
- A quick sanity check scored 28/30 on HumanEval (baseline), 28/30 (ANE Dual Kernel). A real 55k context (OpenCode + MCPs) tool calling test worked.
- I will test on M3 Max Macbook Pro later - additional test devices would be useful
- Chunked Prefill being enabled seems to unreliably change performance but not always in 1 direction and it's not clear why
- Unified memory usage is significantly higher - over double-resident weights, additional IO surfaces, etc. Memory usage could be significantly optimised at the expense of some TG speed (or, to turn a negative into a positive, using a Q8 quant and then sharing the weights)
- Needs work to adapt to >8-bit quants, I don't have the hardware to test 16-bit with the current in-memory duplicate arrangement

Not a finished piece but a solid start towards unlocking Apple Silicon's true PP potential and closing the gap with DGX spark, one chunk at a time (still a lot of TOPS on the table)

onthehub97:
Baseline runs:

- https://omlx.ai/benchmarks/performance/96b0wvgm
- https://omlx.ai/benchmarks/performance/ozs2eg4v

ANE runs (0.53 MLP / 0.53 GDN):

- https://omlx.ai/benchmarks/performance/67mck46c
- https://omlx.ai/benchmarks/performance/paimp3am

Apologies for hammering the benchmark submissions and leaderboard - I've deleted most of the experimental / intermediate runs.

jundot:
Thank you for this exceptional contribution. The dual-ANE/GPU prompt processing work is genuinely impressive, and my local testing confirms a substantial improvement at longer context lengths. I believe this can become a very meaningful step forward for local LLM inference on Apple silicon.

After resolving the current conflict with `main`, I plan to merge #2756 first so that your original contribution and authorship remain clear. With a focused compatibility extension, I have also validated this approach on Qwen3.8-27B-oQ4e. I will open a follow-up PR to add that support, address the remaining runtime details, and release it as an experimental feature while continuing to refine it across more models and workloads.

I am deeply grateful for the thought, experimentation, and implementation behind this PR. The follow-up work would not have been possible without the foundation you built here. This is an outstanding contribution. Thank you again.

aosama:
This is innovative thinking to utilize the ANE, great work, and I would like to second the fact that all these improvements will make it more performant and less costly to run local LLMs

PR #2758, opened 2026-08-17
fix(cluster): compare probe versions against the source the peer reports from

### LINKED ISSUE #2726: Cluster gate still compares mlx/mlx-lm dist-info against module constants, so #2705's asymmetry survives for two of three gated packages

hellodk:
Closes #2726. Follow-up to `78a14640`, which fixed this class of bug for `omlx` but left it for the other two gated packages.

## The skew

The peer reports mlx and mlx-lm as **module constants** — `probe.py:570-571`:

```python
    mlx_version=hardware.get_mlx_version(),      # mx.__version__
    mlx_lm_version=hardware.get_mlx_lm_version(), # mlx_lm.__version__
```

The coordinator compared those against `_local_runtime_versions()`, which resolves both through `importlib.metadata`. So for 2 of the 3 gated packages the probe gate is still local-dist-info against remote-module-constant — exactly the shape #2705 removed for `omlx`. An editable or nightly install whose dist-info has drifted from the loaded module blocks two ranks running identical code, with `mlx-lm` in the message instead of `omlx`.

**The two remote paths also disagreed with each other.** `_PREFLIGHT_SCRIPT` calls `m.version(name)` on the peer, so `preflight_remote_hosts` was already metadata-on-both-ends and symmetric; only `probe_remote_host` was skewed. The same pair of Macs could pass one gate and fail the other.

## The change

Rather than force both remotes onto a single source — which would mean rewriting the preflight string that runs on someone else's machine — each comparison now reads the local side the way its own remote does:

- `_local_runtime_versions()` stays metadata, matching `_PREFLIGHT_SCRIPT`
- new `_local_probe_versions()` mirrors `probe.py`, and `probe_remote_host` uses it

`omlx` continues to come from the source tree in both, so #2705's fix is untouched.

## Sentinel

`hardware.get_mlx_version()` returns `"Unknown"` on failure while `_package_version` returns `"unknown"`. With mlx absent from both ends the gate reported

```
mlx local=unknown remote=Unknown
```

— a mismatch between two ranks in the same state, on capitalisation alone. Reading both sides through the same helper settles it; there is a test pinning that case.

## Tests

New `tests/test_cluster_probe_version_source.py` (6 tests):

- the probe path reads mlx/mlx-lm from the module, the preflight path still from metadata
- two identical nodes pass the probe gate when dist-info has drifted
- a genuine mlx difference is still blocking
- mlx missing on both ends is not a mismatch
- both local sources still agree about `omlx`

Three of the six fail with the one-line change reverted.

Existing peer-probe fixtures in `test_cluster_launch.py` and `test_cluster_interpreter_parity.py` move to the new helper. They were assembling a fake *probe* payload out of metadata, which no real probe would ever report — the reason this skew had no failing test. Preflight fixtures are deliberately left on `_local_runtime_versions()`.

## Verification

`927 passed` on an M-series Mac (Python 3.12.13, full `-k 'cluster or build_ref'` selection over `tests/`).

### DIFF
--- a/omlx/cluster/launch.py
+++ b/omlx/cluster/launch.py
@@ -28,6 +28,8 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 
+from omlx.utils import hardware
+
 from .deployment import ClusterDeployment, validate_ssh_target
 from .liveness import read_marker, read_remote_marker
 from .models import CLUSTER_PROTOCOL_VERSION
@@ -178,10 +180,35 @@ def _package_version(name: str) -> str:
 
 
 def _local_runtime_versions() -> dict[str, str]:
+    """Versions as ``_PREFLIGHT_SCRIPT`` reads them on the peer: metadata."""
+
     return {
         "omlx": _package_version("omlx"),
         "mlx": _package_version("mlx"),
         "mlx-lm": _package_version("mlx-lm"),
+    }
+
+
+def _local_probe_versions() -> dict[str, str]:
+    """Versions as ``probe.py`` reads them on the peer: module constants.
+
+    The two remote paths report differently. ``_PREFLIGHT_SCRIPT`` calls
+    ``importlib.metadata.version``, while ``probe.py`` reports
+    ``mx.__version__`` / ``mlx_lm.__version__`` through these helpers. Comparing
+    a probe result against metadata therefore repeated the #2705 bug for mlx and
+    mlx-lm: an editable or nightly install whose dist-info has drifted from the
+    module blocks two ranks running identical code (#2726).
+
+    Each comparison now reads the local side the same way its own remote does.
+    That also keeps the missing-value sentinel consistent — these return
+    ``"Unknown"`` where ``_package_version`` returns ``"unknown"``, and the two
+    spellings used to read as a mismatch between two ranks that both lacked mlx.
+    """
+
+    return {
+        "omlx": _package_version("omlx"),
+        "mlx": hardware.get_mlx_version(),
+        "mlx-lm": hardware.get_mlx_lm_version(),
     }
 
 
@@ -1302,7 +1329,9 @@ def probe_remote_host(
     # binary.  Reusing that binary directly loses the launcher's PYTHONPATH;
     # carry forward the exact executable that passed this probe instead.
     runtime["python_executable"] = effective_python
-    expected = _local_runtime_versions()
+    # probe.py reports mlx/mlx-lm as module constants, so read ours the same
+    # way rather than from dist-info (#2726).
+    expected = _local_probe_versions()
     remote_versions = {
         "omlx": runtime.get("omlx_version"),
         "mlx": runtime.get("mlx_version"),
--- a/tests/test_cluster_interpreter_parity.py
+++ b/tests/test_cluster_interpreter_parity.py
@@ -21,6 +21,7 @@ from test_cluster_launch import _deployment
 from omlx.cluster import launch
 from omlx.cluster.launch import (
     DistributedLaunchError,
+    _local_probe_versions,
     _local_runtime_versions,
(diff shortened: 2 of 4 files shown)


jundot:
Thanks for the fix. I checked the skew against the tree, probe.py does report mlx/mlx-lm as module constants while probe_remote_host compared them against dist-info, and preflight is metadata on both ends, so keeping each comparison symmetric with its own remote instead of rewriting the preflight script looks like the right call. I also reran the new tests with the one-line change reverted and got the same 3 failures, so the fix is properly pinned. Merging.

PR #2760, opened 2026-08-17
feat(qwen3.5): extend ANE prefill to oQ4e and harden runtime

jundot:
## Acknowledgements

This follow-up exists because of @onthehub97's excellent work in #2756. I am deeply grateful for the care that went into the dual-ANE procedure banks, the hybrid GPU path, and the benchmark tooling. It is an outstanding PR and a meaningful advance for LLM inference on Apple silicon. The oQ4e support and runtime fixes here build directly on that foundation.

## Summary

- Extend experimental Qwen ANE prefill to Qwen3.8 oQ4e layouts:
  - q4 group-size 64 or 128 MLP gate/up projections
  - compatible q2/q4/q5/q6/q8 down projections
  - mixed q4/q5 GDN projections with group size 64 or 128
- Release the Python GIL while compiling native ANE programs so model loading does not stall the server event loop.
- Align scheduler prompt chunks with the compiled ANE shape, including the 4,096-token option, instead of letting the Qwen high-memory floor bypass ANE execution.
- Detect Qwen3.8 models in the admin UI and macOS app.
- Include all Qwen ANE settings and the active feature flag in uploaded benchmark metadata.
- Add the missing Qwen ANE keys to every supported locale.
- Document the oQ4e validation results and runtime tradeoffs.

## Performance

Tested on an M3 Ultra with `Qwen3.8-27B-oQ4e-mtp`, oMLX's built-in `code_python` context, TG128, and a 2,048-token ANE shape. Each enabled configuration is one deterministic run; the previously recorded ANE-off/MTP-off baseline is the mean of two runs, as requested.

Prompt processing throughput:

| Lightning MTP | Prompt | ANE Off | ANE On | ANE change |
|:---:|---:|---:|---:|---:|
| Off | 4K | 452.8 tok/s | 458.5 tok/s | **+1.3%** |
| Off | 16K | 439.1 tok/s | 517.0 tok/s | **+17.7%** |
| Off | 32K | 408.9 tok/s | 486.0 tok/s | **+18.9%** |
| On | 4K | 446.3 tok/s | 451.0 tok/s | **+1.1%** |
| On | 16K | 432.2 tok/s | 507.1 tok/s | **+17.3%** |
| On | 32K | 401.0 tok/s | 473.1 tok/s | **+18.0%** |

Token generation throughput:

| Lightning MTP | Prompt | ANE Off | ANE On | ANE change |
|:---:|---:|---:|---:|---:|
| Off | 4K | 33.6 tok/s | 32.9 tok/s | -2.1% |
| Off | 16K | 31.2 tok/s | 31.1 tok/s | -0.3% |
| Off | 32K | 29.0 tok/s | 28.7 tok/s | -1.0% |
| On | 4K | 59.8 tok/s | 57.2 tok/s | -4.3% |
| On | 16K | 62.8 tok/s | 62.3 tok/s | -0.8% |
| On | 32K | 66.5 tok/s | 66.3 tok/s | -0.3% |

The more controlled current-revision 4K A/B recheck used the exact same prompt back-to-back and measured 445.2 to 460.4 prompt tokens/s (+3.4%). Other matched 4K rechecks ranged from +1.7% to +3.4%, rather than the initially reported +7.8%. A 4K request has only 4,095 prefill tokens: one 2,048-token ANE chunk followed by a 2,047-token GPU tail, so the launch overhead and timing variance dominate much more than at 16K or 32K. Those longer prompts route 7 and 15 full chunks through ANE and measured +17.7% and +18.9% respectively. Decode remains on the GPU, so the small TG differences are run-to-run variance rather than an ANE acceleration claim.

The 16K and 32K output hashes matched the GPU path exactly. The 4K output was stable across ANE runs but differed from GPU, consistent with the approximate INT8 ANE prefix. A 4,096-token ANE shape now works for longer prompts, but it cannot activate for a nominal 4K request because that request only prefills 4,095 tokens; 2,048 therefore remains the default.

Peak memory increased by about 4.15 GB. Eager model loading increased from 3.35 to about 27-29 seconds because the fixed-shape ANE programs are compiled and retained for the model lifetime.

## Validation

- Native custom-kernel build completed successfully.
- Real-model 2K full-body smoke: 450.3 to 558.9 prompt tokens/s, a 1.24x speedup.
- Hidden-state cosine similarity: 0.99879.
- Logit cosine similarity: 0.99871 with the same top token.
- Synthetic dual-ANE oQ4e comparison: 0.99997 cosine similarity.
- Native compile GIL smoke confirmed that the event loop continued running during program-bank compilation.
- Focused Python engine, benchmark, settings, and ANE tests: 318 passed.
- macOS tests: 285 passed, 1 skipped.
- CSS build, JavaScript syntax check, locale normalization, JSON validation, and `git diff --check` passed.

## Notes

This remains an opt-in source-build experiment using private Apple runtime interfaces. It is disabled by default and decode remains on the GPU.


onthehub97:
This is a great addition.

I’d be very interested to see how far we can push this. The measured throughput is still a fraction of the M3U ANEs potential, but it’s difficult gauging what will work and what won’t because the MIL compiler allows so much more than the neural engine itself.

PR #2776, opened 2026-08-17
fix: mcp package updates breaking streamable-http

### LINKED ISSUE #2520: Unable to Connect to Remote MCP Server in v0.5.5+
**Describe the bug**
oMLX is failing to connect to remote MCP server on internal/local network. The relevant error log line is:

`omlx.mcp.client - ERROR - [-] - Failed to connect to MCP server 'docker-mcpGateway': not enough values to unpack (expected 3, got 2)`

**To Reproduce**
1. Set up active remote MCP server on local network
2. Create appropriate JSON config and place in configured oMLX mcp directory
3. Start MCP server
4. Start oMLX server

**Expected behavior**
Typical behavior expected is for a successful MCP server connection and reporting of discovered tools.

**Screenshots**

Image

 - macOS Version: Version 26.6 (25G72) on 64 GB M4 Max Mac Studio
 - oMLX Version [e.g. 0.5.7] via Homebrew

**Additional context**
The MCP server in question is Docker's MCP Gateway via Docker Desktop (v4.84.0 as of this issue) running on a remote server in my network, separate from the machine running oMLX. Everything had been working smoothly up until oMLX v.0.5.5. I suspect it may have something to do with the recent MCP 2.0 change that was implemented but I am unsure of this. The MCP server connection only seems to be affecting the oMLX server as the third-party applications I also use seem to still be detecting and connecting to the MCP server just fine, however I found that the MCP implementation in oMLX, along with the chat interface, was more robust and seemed to result in fewer Tool Call issues compared to the third-party clients, so I had been utilizing that recently. I doubt this is a widespread or critical issue, but wanted to drop this here in case others run into it as well. It very well may be that the Docker MCP Gateway needs to be updated for 2.0 compatibility, though not sure if that's even the issue, or of a timeline for an update to that. Thanks for all your hard work on oMLX, it's been fun learning about and utilizing it in my day-to-day! Keep up the good work!

nnhanndakara:
Updates to the mcp package mean that the previous expectation of 3 values being returned from the context manager (even if we're discarding the 3rd one) is no longer correct.

This change simply removes the `, _` for discarding. The `streamable-http` handler now acts like the `stdio` and `sse` handlers (which both return only the reader/writer pair). If for some reason you want to maintain compatibility with the old version, you could use `, *_` (as another poster mentioned in the issue).

All `test_mcp_*` tests pass.

Fixes: #2520

### DIFF
--- a/omlx/mcp/client.py
+++ b/omlx/mcp/client.py
@@ -212,7 +212,7 @@ class MCPClient:
             self._streamable_http_client = streamable_http_client(
                 url=self.config.url, http_client=self._http_client
             )
-            self._read, self._write, _ = await self._streamable_http_client.__aenter__()
+            self._read, self._write = await self._streamable_http_client.__aenter__()
             self._session = ClientSession(self._read, self._write)
             await self._session.__aenter__()
         except Exception:


jundot:
Thanks for tracking this down. I confirmed that the current DMG bundles MCP 2.0.0, where the streamable HTTP context manager yields only the reader/writer pair, so this is the correct fix for the shipped runtime. LGTM, merging. I’ll follow up by updating the declared MCP requirement from `mcp>=1.5.0` to `mcp>=2.0.0,<3` across the package, bundle, and development dependency groups so the metadata and CI match the runtime contract.

PR #2782, opened 2026-08-17
fix(qwen3.5): don't crash MTP verify-attention on turboquant-quantized cache

### LINKED ISSUE #2778: [Bug] Engine crash during MTP verification with TurboQuant enabled ('TurboQuantMSEState' has no attribute 'ndim')
# [Bug] Engine crash during MTP verification with TurboQuant enabled (`TurboQuantMSEState` has no attribute `ndim`)

### Describe the bug
In oMLX 0.6.1, sending a request to a Qwen3.6 MoE model with MTP and TurboQuant (KV Cache Quantization) enabled causes an immediate engine crash. **Note: This exact same configuration worked perfectly on oMLX 0.6.0.** The model only successfully generates on 0.6.1 if TurboQuant is manually disabled.

The issue stems from the new fused target-verification attention optimization introduced in 0.6.1. In `qwen35_verify_sdpa_split.py`, the `_eligible()` function checks `keys.ndim != 4`. However, when TurboQuant is active, `keys` is a custom `TurboQuantMSEState` object rather than a standard MLX array. Because `TurboQuantMSEState` does not expose an `ndim` attribute, it throws an `AttributeError` and halts the engine loop.

### Environment
- **oMLX Version:** 0.6.1 (Regression from 0.6.0)
- **OS:** macOS (Apple Silicon M1 Max, 32GB Unified Memory)
- **Model:** `Qwen3.6-35B-A3B-Uncensored-Genesis-Hermes-V6-dequantized-oQ3e-mtp`
- **Settings:** TurboQuant (KV Cache Quantization) enabled

### Steps to Reproduce
1. Start oMLX `0.6.1` with the model above (or any Qwen 3.5/3.6 model that supports MTP).
2. Ensure TurboQuant (KV Cache Quantization) is enabled.
3. Send a prompt to the model.
4. The generation fails immediately with `status='failed'` due to an engine loop error. (Turning TurboQuant OFF bypasses the bug).

### Stack Trace
```python
2026-08-18 00:23:11,548 - omlx.scheduler - INFO - [-] - TurboQuant: converted 9/40 cache layers to 4.0-bit, skipped last KVCache layer
...
2026-08-18 00:23:14,216 - omlx.patches.qwen35_gdn_prework - INFO - [-] - [gdn-prework] fused verify prework engaged (S=4)
2026-08-18 00:23:14,225 - omlx.engine_core - ERROR - [-] - Engine loop error: 'TurboQuantMSEState' object has no attribute 'ndim'
  File "/Applications/oMLX.app/Contents/Resources/omlx/engine_core.py", line 394, in _engine_loop
    step_outputs = await loop.run_in_executor(
...
  File "/Applications/oMLX.app/Contents/Resources/Python/framework-mlx-base/lib/python3.11/site-packages/mlx_vlm/models/qwen3_5/language.py", line 1483, in __call__
    output = _target_verify_left_padded_attention(
  File "/Applications/oMLX.app/Contents/Resources/omlx/patches/qwen35_verify_sdpa_split.py", line 132, in patched_target_verify_attention
    limit = _eligible(queries, keys, cache)
  File "/Applications/oMLX.app/Contents/Resources/omlx/patches/qwen35_verify_sdpa_split.py", line 78, in _eligible
    if queries.ndim != 4 or keys.ndim != 4:
  File "/Applications/oMLX.app/Contents/Resources/Python/framework-mlx-base/lib/python3.11/site-packages/mlx_vlm/turboquant.py", line 4953, in __getattr__
    return getattr(self._state, name)
AttributeError: 'TurboQuantMSEState' object has no attribute 'ndim'
```

### Possible Solution
Update `_eligible()` in `qwen35_verify_sdpa_split.py` to correctly handle `TurboQuantMSEState` objects (e.g., checking if the object is quantized before checking dimensions, or exposing the underlying tensor's dimensions safely).

### Workaround
Disabling KV Cache Quantization (TurboQuant) in the oMLX server settings prevents `keys` from being compressed into a `TurboQuantMSEState` object. The cache remains as a standard tensor, allowing the new MTP verification logic to process correctly without crashing.

### LINKED ISSUE #2780: 0.6.1: engine loop crashes with 'TurboQuantMSEState' object has no attribute 'ndim' when turboquant_kv is enabled (qwen35 verify SDPA split)
**Environment:** oMLX 0.6.1, M5 Pro 64GB, `Qwen3.8-27B-oQ8e-mtp`, `mtp_enabled: true`, `turboquant_kv_enabled: true` (`kv_bits: 4`).

**Symptom:** Every chat request stalls (client waits until its own timeout) and the engine loop dies:

```
omlx.engine_core - ERROR - Engine loop error: 'TurboQuantMSEState' object has no attribute 'ndim'
```

Worked fine on 0.6.0 with the identical config.

**Traceback (bottom frames):**

```
File ".../omlx/patches/qwen35_verify_sdpa_split.py", line 132, in patched_target_verify_attention
    limit = _eligible(queries, keys, cache)
File ".../omlx/patches/qwen35_verify_sdpa_split.py", line 78, in _eligible
    if queries.ndim != 4 or keys.ndim != 4:
File ".../mlx_vlm/turboquant.py", line 4953, in __getattr__
    return getattr(self._state, name)
AttributeError: 'TurboQuantMSEState' object has no attribute 'ndim'
```

**Root cause:** The new batched-verify eligibility gate from #2751 dereferences `keys.ndim` (line 78) *before* the quantized-cache exclusion at line 82 (`hasattr(cache, "bits")`). With TurboQuant KV, `keys` is the TurboQuant wrapper whose `__getattr__` proxies to `TurboQuantMSEState`, so the shape check itself raises — before the patch's intended stock-path fallback can engage. The engine loop dies instead of falling back.

**Local fix that restored service** (verify path falls back to stock, as #2751 intended for unsupported inputs):

```diff
-    if queries.ndim != 4 or keys.ndim != 4:
+    if getattr(queries, "ndim", 0) != 4 or getattr(keys, "ndim", 0) != 4:
         return 0
```

`getattr` with a default swallows the wrapper's AttributeError → returns 0 → ineligible → original attention path. Verified working; TurboQuant KV and MTP both stay active. A proper fix might instead reorder the checks or test `isinstance(keys, mx.array)` first — and #2751's speedup presumably still won't engage with quantized KV, which may deserve a note in the docs.

aaramos:
Fixes #2778, Fixes #2780

## Summary

`_eligible()` in the verify-split patch (#2751) checks `queries.ndim` / `keys.ndim` before it checks whether the cache is quantized. A turboquant-quantized KV cache's `update_and_fetch()` returns `_QuantizedStateProxy` objects, which deliberately expose `.shape` but not `.ndim` (to avoid dequantizing on every fetch — see `mlx_vlm.turboquant._QuantizedStateProxy` and its docstring). So the `.ndim` check raises `AttributeError` instead of returning `0` and falling through — which is exactly the "quantized caches" behavior this patch's own module docstring already promises ("delegate everything else (left-padded batches, quantized caches) to the original").

In production this crashes the engine loop mid-generation:

```
omlx.engine_core - ERROR - Engine loop error: 'TurboQuantMSEState' object has no attribute 'ndim'
  .../omlx/engine_core.py:394 _engine_loop
  .../omlx/patches/mlx_lm_mtp/batch_generator.py:2787 _run_verify_cycle_chain -> _call_backbone
  .../omlx/patches/mlx_vlm_mtp/qwen35_vlm_runtime.py:330 -> original_call
  .../mlx_vlm/models/qwen3_5/language.py:1483 -> _target_verify_left_padded_attention
  .../omlx/patches/qwen35_verify_sdpa_split.py:132 patched_target_verify_attention
  .../omlx/patches/qwen35_verify_sdpa_split.py:78 _eligible:
        if queries.ndim != 4 or keys.ndim != 4:
                                 ^^^^^^^^^^^^^
  .../mlx_vlm/turboquant.py:4953 __getattr__ -> getattr(self._state, name)
AttributeError: 'TurboQuantMSEState' object has no attribute 'ndim'
```

Because this happens inside a streaming chat request, the client never gets a token or an error back — the request just hangs until the client gives up.

## Fix

Use `getattr(queries, "ndim", None)` / `getattr(keys, "ndim", None)` instead of direct attribute access, so an object without `.ndim` is treated as "not ours" — the same way the existing `hasattr(cache, "bits")` check already excludes other quantized caches — instead of raising. Normal (non-quantized) traffic is unaffected: same checks run afterward, same row-limit returned.

## Testing

- Added `test_eligibility_gates_turboquant_proxy`, mirroring the existing `_QuantCache` case in `test_eligibility_gates`, using a minimal stand-in for `_QuantizedStateProxy` (has `.shape`, no `.ndim`).
- Ran the full `tests/test_qwen35_verify_sdpa_split.py` suite locally (14 tests) — all pass.
- I didn't have a way to reproduce this end-to-end against a live turboquant-enabled server in the environment I was working in. Confidence comes from: the crash traceback above (from a real server log), and unit-level confirmation that the unpatched `_eligible()` raises on a proxy-shaped input while the patched version returns `0` for it and is unchanged for plain-array input.

### DIFF
--- a/omlx/patches/qwen35_verify_sdpa_split.py
+++ b/omlx/patches/qwen35_verify_sdpa_split.py
@@ -75,7 +75,13 @@ def _chunked_causal_sdpa(queries, keys, values, scale, limit: int):
 
 def _eligible(queries, keys, cache) -> int:
     """Return the vector-kernel row limit (>0) when this call is ours."""
-    if queries.ndim != 4 or keys.ndim != 4:
+    # A turboquant-quantized cache hands back `_QuantizedStateProxy` objects
+    # (exposes .shape, deliberately not .ndim, to avoid dequantizing — see
+    # mlx_vlm.turboquant). That's exactly the "quantized caches" case this
+    # patch already means to delegate to the original path below, but the
+    # attribute access below used to run before the `hasattr(cache, "bits")`
+    # guard could rule it out, so it crashed instead of falling through.
+    if getattr(queries, "ndim", None) != 4 or getattr(keys, "ndim", None) != 4:
         return 0
     if queries.shape[0] != 1:
         return 0
--- a/tests/test_qwen35_verify_sdpa_split.py
+++ b/tests/test_qwen35_verify_sdpa_split.py
@@ -77,3 +77,23 @@ def test_eligibility_gates():
         bits = 4
 
     assert _eligible(q, k, _QuantCache()) == 0
+
+
+@pytest.mark.skipif(not mx.metal.is_available(), reason="requires Metal")
+def test_eligibility_gates_turboquant_proxy():
+    """A turboquant-quantized KV cache hands back a proxy with .shape but no
+    .ndim (mlx_vlm.turboquant._QuantizedStateProxy, kept dequantized-free on
+    purpose). _eligible() must treat that as "not ours" rather than raising —
+    it used to crash every verify forward once turboquant KV compression was
+    active, since the .ndim check ran before the cache-type guard could rule
+    the call out.
+    """
+
+    class _TurboQuantProxy:
+        def __init__(self, shape):
+            self.shape = shape
+
+    q = mx.random.normal((1, HQ, 4, HD)).astype(mx.bfloat16)
+    k = mx.random.normal((1, HKV, 256, HD)).astype(mx.bfloat16)
+    assert _eligible(_TurboQuantProxy((1, HQ, 4, HD)), k, None) == 0
+    assert _eligible(q, _TurboQuantProxy((1, HKV, 256, HD)), None) == 0


jundot:
Thanks for the focused fix. I reproduced the crash with Lightning MTP and TurboQuant KV enabled, then confirmed the same streaming request completes on this branch while the verify-split fast path still engages for eligible layers. The safe ndim check correctly delegates TurboQuant proxies to the existing fused path without changing plain-array eligibility. LGTM, merging.

zesming:
End-to-end validation of this merged fix on the affected configuration:

- Hardware: Mac mini, Apple M4 Pro, 64 GB unified memory
- Runtime: oMLX 0.6.1 app build, tested from an isolated copy with only the `_eligible()` change from this PR applied
- Model: Qwen3.8 27B oQ4e VLM, `model_type=qwen3_5`, native Lightning MTP weights present
- Settings: `mtp_enabled=true`, `turboquant_kv_enabled=true`, KV bits 4, skip-last enabled

Baseline 0.6.1 reproduced the reported `TurboQuantMSEState ... ndim` engine-loop failure. With this PR's change:

1. Low-level proxy gate passed, and dense verify-split attention parity had max abs diff 0.0 in the tested shape.
2. Text completion returned HTTP 200. Logs confirmed TurboQuant converted 15/16 attention KV layers and Lightning MTP accepted 13/14 draft tokens (92.9%) on the short request.
3. A real image request completed correctly (identified the flower as 大丽花). Logs confirmed MTP remained active after vision prefill, accepting 37/64 draft tokens (57.8%).
4. Three deterministic, thinking-disabled 320-token runs completed without error. MTP acceptance was about 62%; observed throughput averaged 19.36 tok/s with TurboQuant KV4 versus 18.87 tok/s without it on this machine. This is only a local point measurement, not a general speed claim.
5. The KV estimate reported by oMLX dropped from 16.00 MB to 4.81 MB per 64-token block.

The intended composition is working: quantized layers fall through to the TurboQuant multi-row path, while the one skip-last dense KV layer can still engage verify-split. No text, MTP, or multimodal regression was observed.

PR #2799, opened 2026-08-18
fix(memory): align wired-limit hints to whole MiB

### LINKED ISSUE #2800: [Bug] Wired-limit UI rounds the 95% safety threshold up to 124519
**Describe the bug**

The Admin Settings memory-guard banner can enter an impossible rounding loop for `iogpu.wired_limit_mb` on a 128 GiB Mac.

The backend calculates the 95% safety threshold as `130567005799` bytes. Because the kernel sysctl accepts whole MiB, the highest realizable value that does not exceed the threshold is `124518 MiB` (`130566586368` bytes).

However, the admin UI compares `124518 MiB` against the exact byte threshold and formats its command with `Math.ceil`, so it asks the user to run:

```shell
sudo sysctl iogpu.wired_limit_mb=124519
```

After setting `124519`, the backend warns that this exceeds the safe 95% recommendation and recommends `124518` instead. Neither value can satisfy both checks.

**To Reproduce**

1. Run oMLX on a 128 GiB Apple Silicon Mac with Prefill Memory Guard enabled.
2. Set `iogpu.wired_limit_mb=124518`.
3. Open Admin Settings → Memory & Lifecycle.
4. Observe that the banner still asks for `124519`.
5. Set `iogpu.wired_limit_mb=124519` and restart oMLX.
6. Observe the backend warning that the recommended maximum is `124518 MiB`.

**Expected behavior**

The backend warning, admin API, UI comparison, and displayed command should use the same realizable whole-MiB recommendation. For a 128 GiB system, that value is `124518`.

**Screenshots**

The Admin Settings banner displays: `sudo sysctl iogpu.wired_limit_mb=124519` while the backend's safe maximum is `124518 MiB`.

- macOS Version: 26.6.1
- oMLX Version: 0.6.1 macOS app
- Hardware: Apple Silicon, 128 GiB unified memory

**Additional context**

The shared helper currently exposes a fractional-MiB byte recommendation to the admin UI. Flooring `_wired_limit_suggestion_bytes()` to whole MiB keeps the 5% safety margin and gives every consumer the same value.

Draft fix: https://github.com/jundot/omlx/pull/2799

The regression test verifies that `124518 MiB` does not trigger another raise hint. All 164 tests in `tests/test_process_memory_enforcer.py` pass with the fix.

Related safety context: #2184
Initial investigation: https://github.com/jundot/omlx/discussions/2798

jakeuj:
## Summary

- floor wired-limit recommendations to the whole-MiB unit accepted by `iogpu.wired_limit_mb`
- keep backend warnings, the admin API, and the UI command on the same realizable safe value
- add a 128 GiB regression case proving `124518 MiB` no longer triggers a request for `124519`

## Problem

On a 128 GiB Mac, the 95% safety threshold is `130567005799` bytes. The highest whole-MiB sysctl value below that threshold is `124518 MiB`, but the admin UI compared that kernel value with the fractional-MiB byte threshold and displayed the command using `Math.ceil`, recommending `124519`.

That value exceeds the safety threshold, so the backend then recommended `124518`, creating an impossible warning loop.

The fix aligns the shared backend recommendation before it reaches either consumer. This preserves the 5% headroom and avoids duplicating rounding policy in the UI.

Fixes #2800

Initial discussion: https://github.com/jundot/omlx/discussions/2798

## Testing

- `pytest tests/test_process_memory_enforcer.py -q` — 164 passed
- `ruff check omlx/process_memory_enforcer.py`
- `ruff check tests/test_process_memory_enforcer.py --ignore SIM117` (the file has a pre-existing SIM117 finding outside this change)
- `git diff --check`

### DIFF
--- a/omlx/process_memory_enforcer.py
+++ b/omlx/process_memory_enforcer.py
@@ -185,7 +185,7 @@ def get_effective_metal_cap_bytes() -> int:
 
 
 def _wired_limit_suggestion_bytes(desired_bytes: int) -> int:
-    """Clamp a wired-limit recommendation to leave the OS 5% of RAM.
+    """Clamp and align a wired-limit recommendation to leave 5% of RAM.
 
     Wiring within a few GiB of physical RAM invites jetsam during a
     large-model load burst, and a jetsammed/hard-killed process strands
@@ -198,6 +198,10 @@ def _wired_limit_suggestion_bytes(desired_bytes: int) -> int:
     the enforcement path still honors whatever the kernel sysctl allows,
     including user-set values above this recommendation.
 
+    ``iogpu.wired_limit_mb`` accepts whole MiB. Floor the recommendation
+    to that unit so the displayed command cannot cross the byte-level
+    safety threshold and the UI compares against a realizable value.
+
     Resolved through the settings module at call time so tests that patch
     omlx.settings.get_system_memory control this the same way they control
     the static ceiling.
@@ -208,7 +212,10 @@ def _wired_limit_suggestion_bytes(desired_bytes: int) -> int:
         return desired_bytes
     if total <= 0:
         return desired_bytes
-    return max(0, min(desired_bytes, total - total // 20))
+
+    suggestion = max(0, min(desired_bytes, total - total // 20))
+    mib = 1024**2
+    return suggestion // mib * mib
 
 
 def _apply_metal_wired_limit(desired_bytes: int) -> tuple[int, int | None]:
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the focused fix. I verified that flooring the shared recommendation keeps the backend warning, admin comparison, and generated sysctl command on the same safe whole-MiB value, and the 128 GiB regression case covers the reported loop. LGTM, ready to merge.

PR #2802, opened 2026-08-18
glm_moe_dsa: M2-tuned DSA indexer score kernel, 1.37x over Steel on D…

nathanodle:
…SV4F

Add dsa_indexer_scores_mma, a from-scratch simdgroup_matrix GEMM for the DeepSeek-V4-Flash indexer score path on Apple M2 Ultra, and route the DSV4F/M2 pairing to it behind an exact fingerprint + env gate.

Kernel structure (vs the Steel dsa_indexer_score's stage-barrier-consume loop): the K tile lives transposed-resident in threadgroup memory (16KB, loaded once — ONE threadgroup_barrier per kernel instead of 3 per K-slice per head), Q fragments load directly from device per MMA step, and the first MMA of each head initializes the accumulator tile in place of a clear. 74-75% of BF16 peak, ~92% of the device's measured simdgroup_matrix roofline.

M2 Ultra, M=4096 query chunk, median ms vs dsa_indexer_scores:
  ctx 32K:  38.1 -> 27.8  (1.37x)   ctx 256K: 298.0 -> 216.5 (1.38x)
  ctx 64K:  74.9 -> 54.6  (1.37x)   ctx 512K: 594.6 -> 431.9 (1.38x)
  ctx 128K: 149.3 -> 108.6 (1.38x)  ctx 1M:  1188.1 -> 862.3 (1.38x)

Correctness: bit-exact vs the Steel kernel (zero mismatched elements) across aligned and unaligned M/N (production pooled lengths are not tile-aligned), mask_ratio 0/1/4, chunked-prefill mask offsets, and a fresh-process whole-model A/B (arms differing only in the env gate: identical score-tensor hashes and generated text). New test file covers the shape/mask matrix and unsupported-config rejection.

Scope and safety:
- Serves ONLY bf16 / H=64 / D=128 / weights [B,L,H] / non-causal; the entry point raises on anything else. The GLM sparse-MLA caller (causal, H=32) and every non-DSV4F model keep the Steel path untouched.
- Model-side gate: exact DSV4F/M2-Ultra architecture fingerprint AND OMLX_DSV4F_M2_MMA_SCORE (default on) AND the extension exposing the symbol; falls back to the existing call otherwise.
- Interior/boundary split dispatch: full 64x64 tiles run the unmodified hot loop; partial edge tiles run a separate address-clamped instantiation (sharing one specialization measurably costs 3-10 %-points via register allocation).
- M/N/mask offsets are runtime params — growing-N prefill never recompiles. The kernel source compiles once at runtime via Device::get_library (the runtime Metal compiler's codegen measures 3.4 %-points faster than the metallib CLI toolchain for this kernel).

Follow-up welcome: the gate deliberately activates only on the exact benchmarked pairing (Apple M2 Ultra). The kernel compiles and should run correctly on other Apple Silicon (bit-exactness does not depend on the tuning), but it is unmeasured there — M3/M4 have async_copy and different occupancy behavior, so Steel may win. If someone benchmarks it on M2 Max/Pro or M3/M4-class hardware and it is faster, widening the device check is a one-line change.

### DIFF
--- a/omlx/custom_kernels/glm_moe_dsa/csrc/bindings.cpp
+++ b/omlx/custom_kernels/glm_moe_dsa/csrc/bindings.cpp
@@ -43,6 +43,15 @@ NB_MODULE(_ext, m) {
       "mask_q_offset"_a = 0,
       "stream"_a = nb::none());
   m.def(
+      "dsa_indexer_scores_mma",
+      &omlx::glm_kernels::dsa_indexer_scores_mma,
+      "queries"_a,
+      "keys"_a,
+      "weights"_a,
+      "mask_ratio"_a = 0,
+      "mask_q_offset"_a = 0,
+      "stream"_a = nb::none());
+  m.def(
       "dsa_topk_indices",
       &omlx::glm_kernels::dsa_topk_indices,
       "scores"_a,
--- a/omlx/custom_kernels/glm_moe_dsa/csrc/dsa_indexer.cpp
+++ b/omlx/custom_kernels/glm_moe_dsa/csrc/dsa_indexer.cpp
@@ -1,5 +1,7 @@
 #include "dsa_indexer.h"
 
+#include "kernels/mma_dsa_indexer_score.h"
+
 #include <cstdlib>
 #include <dlfcn.h>
 #include <filesystem>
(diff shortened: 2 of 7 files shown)


jundot:
Merged, thanks. I reran the bit-exact matrix on an M3 Ultra as well, plus batched and non-contiguous q inputs, and everything matched Steel exactly. I also tried a perf sweep there since you mentioned widening the device check: results are mixed on M3 Ultra (1.05-1.08x on some shapes but 0.66-0.93x on others, pretty occupancy-sensitive), so the exact M2 gate you shipped looks like the right call. Pushed a few cosmetic follow-ups directly to main.

PR #2803, opened 2026-08-18
feat(deepseek): experimental hybrid ANE prompt processing

jundot:
## Summary

This extends the ANE prompt processing experiment from #2756 / #2760 to DeepSeek-V4-Flash. Eligible 4,096-token prefill chunks run the shared expert gate/up, the attention wq_b, and the sparse-layer indexer wq_b as an INT8 prefix on both ANEs while the GPU computes an affine-q8 suffix requantized from the mxfp8 weights. Routed experts, wo_a/wo_b, the attention core, decode, and DSpark verify keep the existing GPU path. Off by default, exposed as a per-model setting for detected DeepSeek-V4 models.

## What changed

- New `patches/deepseek_v4/ane_prefill.py` with mxfp8 eligibility, prep, and the hybrid backends, reusing the Qwen procedure banks and split ladder.
- The indexer wq_b rides inside the attention wq_b procedure since both consume the same q_residual, so sparse layers get it without any extra operations.
- Backend hooks live in the vendored model file directly, no monkeypatching. DSpark verify arming and the fixed-shape check fall through to the GPU path.
- The paged cache block size follows the ANE shape on pooling-cache models, because boundary snapshot clamping otherwise keeps every chunk below the compiled shape.
- Native side: grouped 1x1 conv compilation (groundwork for wo_a), q8 suffix support, and staged ANE model directories are now deleted right after program load. Killed or crashed processes used to leak multi-GB temp directories, and that fix covers the Qwen path too.
- Per-model settings, admin editor section with locale strings, reload signature keys, profiles, benchmark metadata, a shape microbench, and docs.

## Why 4,096-token chunks

DeepSeek-V4 projections are 5-10x smaller than the Qwen MLP slices this runtime was built for, so at 2,048 tokens the per-op host sync eats the gain and the hybrid actually loses. Doubling the chunk halves the op count and flips the result. wo_b stays on the GPU: its 8,192-wide input drags ANE throughput enough that the hybrid never wins. Ablation with a fixed ~15k prompt, SSD cache off:

| Config (15k prompt) | pp tok/s | vs baseline |
|---|---:|---:|
| GPU baseline | 612.8 | - |
| 2,048 chunks, mlp+wq_b+wo_b | 596.8 | -1.9% |
| 2,048 chunks, wq_b only | 612.0 | +0.6% |
| 4,096 chunks, mlp+wq_b+wo_b | 639.4 | +4.3% |
| 4,096 chunks, mlp+wq_b | 646.4 | +5.5% |
| 4,096 chunks, mlp+wq_b+stacked indexer (final) | 647.4 | +5.6% |

## Performance

Built-in throughput benchmark on an M3 Ultra (code_python context, TG=128, greedy), fresh server and cleared SSD cache per configuration.

`DeepSeek-V4-Flash-0731-oQ2.5e-mtp`:

| MTP | Prompt | pp ANE off | pp ANE on | pp change | tg ANE off | tg ANE on |
|:---:|---:|---:|---:|---:|---:|---:|
| Off | 4k | 621.9 | 637.7 | **+2.5%** | 26.2 | 26.0 |
| Off | 16k | 600.4 | 638.3 | **+6.3%** | 25.0 | 25.1 |
| Off | 32k | 583.3 | 627.1 | **+7.5%** | 24.1 | 24.2 |
| On | 4k | 604.2 | 633.2 | **+4.8%** | 40.7 | 39.0 |
| On | 16k | 583.3 | 621.6 | **+6.6%** | 46.4 | 44.9 |
| On | 32k | 567.0 | 608.4 | **+7.3%** | 42.1 | 42.0 |

`DeepSeek-V4-Flash-0731` (original fp8 precision):

| MTP | Prompt | pp ANE off | pp ANE on | pp change | tg ANE off | tg ANE on |
|:---:|---:|---:|---:|---:|---:|---:|
| Off | 4k | 636.0 | 650.7 | **+2.3%** | 25.1 | 24.6 |
| Off | 16k | 613.4 | 634.8 | **+3.5%** | 23.9 | 23.7 |
| Off | 32k | 595.8 | 621.7 | **+4.3%** | 23.3 | 23.0 |
| On | 4k | 617.4 | 631.5 | **+2.3%** | 35.0 | 38.3 |
| On | 16k | 595.7 | 617.1 | **+3.6%** | 41.8 | 44.4 |
| On | 32k | 578.6 | 603.9 | **+4.4%** | 42.4 | 44.5 |

A nominal 4k request prefills 4,095 tokens, one short of the ANE shape, so the 4k gain comes from the block realignment collapsing prefill into one chunk rather than from ANE execution. DSpark tg moves a few tok/s run to run with the generated content and repeated runs showed no systematic decode regression; plain decode stays within about one percent. The oQ2.5e MTP-on rows are means of 2-4 runs, the fp8 rows single runs. The smaller fp8 gains are not a coverage gap, both checkpoints engage the same 86 procedures with numerically identical weights; the mxfp4 experts just stream about 1.6x the bytes and compete with the ANE for memory bandwidth.

Costs: about +0.7 GB model memory and +5 s eager load for the banks. Enabling the feature realigns the paged block to 4,096, which rebuilds that model's SSD cache once; a store/restore round trip reproduced the cold output bitwise.

## Quality

Outputs are deterministic run to run, and match the GPU output bitwise on some prompts while diverging only in the reasoning wording on others, which is what an approximate INT8/q8 path should do. A long-context spot check (needle retrieval at 7 depths plus multi-fact aggregation, ~12k-token prompts) scored 10/10 on both paths, so the approximate indexer q does not disturb the top-512 selection.

## Notes

- Same caveats as the Qwen feature: private AppleNeuralEngine interfaces, not bit-exact, dual-ANE layout assumed from the M3 Ultra, NAX auto-skip inherited.
- Single-die machines are untested for this family; the two-bank split ladder from d7e43a4d should handle the residency limit there.
- wo_a acceleration needs a grouped GPU suffix and stays a follow-up. The grouped conv itself already compiles and validates on ANE.

## Validation

Focused Python tests (DSv4 patch, Qwen ANE regression, benchmark, admin settings, profiles, i18n keys), JS syntax check, Jinja compilation, CSS rebuild, locale normalization, and the real-model runs above.


onthehub97:
Beat me to it.

If you can distribute routed experts, we can probably bring this up to at least 15% - 30% boost. How's it looking on the resource budget front so far?

Also wo_a might plausibly add another 1–3 percentage points if the grouped suffix can be made efficient.

PR #2806, opened 2026-08-18
fix(profiles): resolve VLM MTP conflicts on apply

### LINKED ISSUE #2801: [Bug] Applying a profile with presence_penalty to a VLM-MTP-enabled model returns HTTP 500
**Describe the bug**

Applying a sampling profile that sets a non-zero `presence_penalty` to a model whose base settings have `vlm_mtp_enabled: true` returns an unhandled HTTP 500 from the admin API.

The incompatibility itself appears intentional because the VLM MTP decode path cannot apply the per-request logits processor. The problem is that the profile UI/API allows the conflicting effective configuration to reach `ModelSettings` construction and exposes the validation `ValueError` as an internal server error.

**To Reproduce**

1. Enable `vlm_mtp_enabled` in the model's base settings.
2. Create or select a profile with `presence_penalty: 1.5`.
3. Apply the profile from the admin UI.

Relevant effective settings:

```json
{
  "base_model_settings": {
    "vlm_mtp_enabled": true
  },
  "profile_settings": {
    "presence_penalty": 1.5,
    "repetition_penalty": 1.0
  }
}
```

The request fails with:

```text
POST /admin/api/models/Qwen3.8-27B-8bit/profiles/qwen3-r-general/apply -> 500

ValueError: vlm_mtp_enabled cannot be combined with presence_penalty; these
settings require per-request logits processors, which the vlm_mtp decode path
does not apply
```

The failure occurs while constructing the merged `ModelSettings`, before the new settings are assigned or persisted, so the profile is not partially activated.

**Expected behavior**

The incompatibility should produce an actionable result instead of HTTP 500. Any of the following would resolve the user-facing failure:

1. Prevent the incompatible combination in the UI and explain the tradeoff.
2. Return HTTP 400/422 with the validation message.
3. Resolve the merge consistently with the existing conflict-resolution paths, for example by disabling the speed-only `vlm_mtp_enabled` setting when an output-shaping penalty is applied.

This report is about profile validation and API/UI error handling, not about removing the VLM MTP restriction.

**Screenshots**

Not applicable; the relevant server response and exception are included above.

- macOS Version: 26.6.1
- oMLX Version: 0.6.1 macOS app
- Architecture: Apple Silicon (`arm64`)
- Model: `Qwen3.8-27B-8bit`

**Additional context**

The current implementation has `resolve_vlm_mtp_conflicts()` for some saved-settings and runtime-profile merge paths. `ModelSettingsManager.apply_profile()` appears to pass its merged dictionary directly to `ModelSettings.from_dict()`, allowing this expected validation error to propagate through the admin route as HTTP 500.

Originally reported in Discussion #2797.

jakeuj:
## Summary

- resolve VLM MTP conflicts during persistent profile application, after any model-specific sanitizer runs
- preserve output-shaping profile settings and disable the speed-only VLM MTP toggle, matching request-time profile behavior
- return HTTP 400 for remaining invalid profile combinations without partially updating persisted settings
- surface profile-application validation errors in the web dashboard

## Root cause

`ModelSettingsManager.apply_profile()` merged a saved profile directly into `ModelSettings.from_dict()` without calling the conflict resolver already used by load-time and request-time profile overlays. A profile adding a logits-processor setting such as `presence_penalty` therefore raised `ValueError`; the admin route did not map that validation error, so the API returned HTTP 500.

## Tests

- `pytest tests/test_model_settings.py tests/test_model_settings_profiles.py tests/test_admin_profiles_api.py tests/test_admin_model_settings_template.py -q` — 181 passed
- `ruff check tests/test_admin_model_settings_template.py tests/test_admin_profiles_api.py tests/test_model_settings_profiles.py` — passed
- `node --check omlx/admin/static/js/dashboard.js` — passed
- `git diff --check` — passed

Full-file Ruff on `omlx/model_settings.py` and `omlx/admin/routes.py` still reports pre-existing violations present on `upstream/main`; the new Python test changes lint clean.

Fixes #2801

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -2962,7 +2962,10 @@ async def apply_model_profile(
     entry = _require_model(model_id)
     is_diffusion_model = _entry_is_diffusion_model(entry)
     sanitizer = _sanitize_diffusion_settings_dict if is_diffusion_model else None
-    applied = mgr.apply_profile(model_id, name, settings_sanitizer=sanitizer)
+    try:
+        applied = mgr.apply_profile(model_id, name, settings_sanitizer=sanitizer)
+    except ValueError as e:
+        raise HTTPException(status_code=400, detail=str(e)) from e
     if applied is None:
         raise HTTPException(status_code=404, detail=f"Profile not found: {name}")
     if is_diffusion_model:
--- a/omlx/admin/static/js/dashboard.js
+++ b/omlx/admin/static/js/dashboard.js
@@ -7229,6 +7229,7 @@
             },
             async applyProfileToForm(profile) {
                 const seq = ++this._applySeq;
+                this.profileError = '';
                 try {
                     const r = await fetch(
                         `/admin/api/models/${encodeURIComponent(this.selectedModel.id)}/profiles/${encodeURIComponent(profile.name)}/apply`,
@@ -7256,8 +7257,12 @@
                         if (m) m.settings = { ...settings };
                     } else if (r.status === 401) {
                         window.location.href = '/admin';
+                    } else {
+                        const data = await r.json().catch(() => ({}));
+                        this.profileError = data.detail || 'Failed to apply profile';
                     }
                 } catch (e) {
+                    this.profileError = String(e);
                     console.error('Failed to apply profile:', e);
                 }
             },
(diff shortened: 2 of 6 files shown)


jundot:
Thanks for the fix. The persistent apply path now matches the request-time VLM MTP conflict resolution while preserving output-shaping settings, and remaining validation failures return HTTP 400 without partial writes. The focused profile and admin tests pass locally. LGTM.

PR #2810, opened 2026-08-18
perf(cache): serve multi-turn cache hits from the live GPU chain and SSD, not a CPU hot-cache copy

hojin12312:
Depends on #2809 (fixes a reference-cycle leak in upstream's own snapshot-collection helper). Without it, the TurboQuant-KV numbers below will show a per-turn memory climb that isn't caused by this patch — see the Validation section.

**Opt-in, off by default.** Everything in this PR is gated behind `_live_chain_retention_enabled()`, which is `False` unless `OMLX_RETAIN_LIVE_CHAIN=1` is set in the environment. With the variable unset, none of the code paths below run and behavior is byte-for-byte what it was before this PR.

## The headline result

A multi-turn conversation gets its cache hit from one of two places today, and neither of them has to be a CPU-resident "hot cache" copy. The just-completed request's KV/GDN state is still live on the GPU when the next turn arrives — if nothing else has claimed that slot, the next turn can adopt it directly, no round trip through a serialized CPU copy required. If the slot *was* claimed (contention) or the request's prompt has moved on since (junction mismatch), the durable per-block snapshots this patch keeps on the SSD sidecar serve the recoverable part instead. The hot-cache tier — `KVCache`-sized copies held in CPU RAM specifically to survive between turns — turns out to add nothing once this path exists.

Measured directly on two structurally different cache architectures, six-turn conversations at ~74K prompt tokens:

| model | cache architecture | `hot_cache=6GB` warm | `hot_cache=0` warm | delta |
|---|---|---:|---:|---:|
| Qwen3.8-27B-OptiQ-4bit (GDN hybrid) | `ArraysCache` | 27.03s | 27.03s | **0.00s** |
| gemma4-12b-qat-optiq-4bit | `RotatingKVCache` + `KVCache` | 8.11s | 8.06s | **0.05s** |

(`warm` is the summed latency of turns 2–6, not a single turn.) In each row the two columns are the same request/response pair, with the same `adopted`/`retained`/`cached` counts and the same outputs, for a sub-1%, noise-level difference. With retention on the hot cache goes unused either way — `hot_cache_cpu_bytes` is 0.00 GiB in both columns, against 1.35 GiB on the same harness with retention disabled (measured on the GDN row; the rotating run did not record that counter) — so its 6 GB budget can be handed back to the system for free. This held on a 30.00 GiB memory ceiling (Mac Studio, Apple M4 Max, 36 GiB unified memory), not just an abundant-memory box.

The Gemma run also shows what the old path costs when live-chain adoption is turned off: without it, the next turn has to reconstruct the chain from a CPU copy, and on this rig that reconstruction spiked memory from 10.4 GiB to 34.5 GB mid-prefill and hit the hard memory guard — a real abort, not a slowdown. With adoption on, the same turn finishes in 1.4 seconds and memory stays flat.

## Design

1. **Base (retain-and-adopt)**: on request completion, if the slot isn't needed elsewhere, leave the live Metal-resident chain installed in a retention slot keyed to the conversation instead of copying it into the hot cache. The next turn that names the same lineage adopts the slot's chain directly if its prompt prefix still matches. Reconstruction covers `ArraysCache` (GDN) as well as plain dense `KVCache` — GDN is the model architecture this was originally written for.
2. **Spill on contention**: the sequential case (no contention) never stores anything, since the next turn just adopts the live slot. But a second conversation retaining its own chain in the same slot silently erased the first one's — its next turn missed everywhere and paid a full cold prefill. This stores the outgoing chain to the SSD sidecar on the way out of a slot that's being replaced, so a contended chain still leaves something recoverable. Measured on a 98K-token conversation whose slot is taken by a small request: the spill writes 48 blocks (6.14 GB) and the displaced conversation's next turn returns in 6.99s with 98,304 tokens cached, against 207s to re-prefill it cold. At 6.8K the same scenario goes from 12.83s with nothing cached to 3.08s with 6,144 cached. Deliberately writes through to SSD, not the hot cache — populating the hot cache here would evict blocks an interleaved conversation is still using.
3. **Partial adoption on prefix divergence**: adoption was originally all-or-nothing, so a mismatch in the last few tokens of a long prompt cost the entire prefill. The contention-spill mechanism keeps durable per-block snapshots alive while a payload is pending, so this walks down to the deepest block boundary at or below the divergence point and adopts that much, re-prefilling only the remainder. Measured: a turn diverging 11 tokens from the end of a 98K-token prompt goes from 220s cold to 6.6s with 98,304 tokens cached.
4. **Extend boundary snapshots to rotating/CacheList layouts**: the spill/partial-adopt path originally read per-block state through a provider gated to exclude rotating and `CacheList` caches by design — which excluded Gemma, this patch's own headline model, from ever getting the benefit. This routes the snapshots that rotating/CacheList layouts already capture during ordinary prefill into the spill payload as a second provider source, and fixes a matching slice failure that made partial adoption fail outright on every rotating model.
5. **Carry the spill payload across adoption, not just contention**: the rotating-layout fix's first live Gemma test surfaced a second-order regression — an *adopted* request only prefills the remainder of the prompt, so it never produces a block-aligned boundary snapshot of its own, which means neither the spill path nor the ordinary completion store has anything to work with on the *next* turn. Every other turn went fully cold (50K prompt, six turns: 19.78s → 130.33s). Fix: hand the adopting request the snapshots the outgoing request already took of the shared prefix (nothing changed on that prefix, so they're still exact). The first version of this fix carried the whole spill payload including a `"cache"` field nothing downstream reads, which pinned the pre-adoption chain's now-orphaned buffers for the request's lifetime — a second full copy, exactly the duplication this patch exists to remove (`mlx active` 38.17→45.18 GB on Qwen). The corrected version is byte-identical to plain retention on Qwen memory (`mlx active`/`peak`/`phys_footprint` 38.17/42.49/44.39 GB, unchanged) and, on the Gemma timing side, brings the six-turn total to 6.40s (3.1x faster than the no-retention baseline, zero cold turns). Qwen is unaffected by the fix (19.38s vs. 19.75s before it).
6. **NamedTuple hardening**: the generic snapshot-copy helper this patch's spill/adopt path shares with the upstream prefill-snapshot machinery rebuilt every tuple it walked as a plain `tuple`, which silently strips `NamedTuple` subclasses of their class. TurboQuant keeps all of its codec state in `NamedTuple`s, so with TurboQuant KV enabled, adoption failed on *every* turn (`Unsupported TurboQuant state type: <class 'tuple'>`) and fell back to the slow path with no error surfaced to the caller. Fix rebuilds through the original type (`type(value)._make()` for `NamedTuple`, `type(value)(copied)` for other tuple subclasses, flat-tuple fallback if the subclass constructor is incompatible). Not reachable from upstream's own call site — `_prefill_snapshot_value()` empties `_KNOWN_SLICEABLE_CACHE_TYPES` layers (TurboQuant included) before the snapshot is built, so upstream never hands this helper a TurboQuant state. This patch's own generic path does, because it passes the whole prompt-boundary cache. The hardening is general either way and worth keeping regardless of framing.

## Correctness

- Cross-slot contamination: 0 across concurrent conversations retaining/adopting in the same run.
- Junction-mismatch safe fallback: verified both before partial-adoption existed (falls back to full re-prefill, no incorrect output) and after (partial adopt, still no incorrect output).
- Effect gate on every measurement below: the log must show `Live chain: retained` within 2 turns of enabling retention, or the run is treated as invalid and discarded.

## Validation

Correctness was proven first on GDN (`ArraysCache`) with small and large (~48K token) capture→retain→adopt round-trips (cold 76.4s → adopt 3.7s, output unchanged), then extended to cross-slot contamination and junction-mismatch fallback, then to a 99K-token cold-vs-adopt comparison (211.9s → 4.15s).

The rotating/CacheList extension needed two passes: the first live test on a Gemma model regressed 6.6x (root cause in Design item 5 above); the fix landed at 3.1x faster than the no-patch baseline. The first version of that fix also leaked +7.01 GB; the corrected version is byte-identical to plain retention.

Ported to real memory pressure (a 30.00 GiB ceiling), this surfaced two more issues, both already covered above: TurboQuant KV made adoption fail every turn (NamedTuple bug), and after that fix a residual per-turn memory climb appeared. That climb turned out to be the reference-cycle bug fixed in #2809, not a defect in this patch — re-measured on top of that fix, the climb is gone (flat oscillation, no net growth over 6 turns), and all other conclusions (speed, ceiling, hot-cache uselessness) held.

A second real-memory-pressure run on the rotating architecture (Gemma4-12B) confirmed retention-off genuinely aborts on turn 2 under pressure (a real memory-guard trip, not a simulated one) while retention-on completes all 6 turns flat, with adoption firing reliably (13 retained, 5/5 adopted).

## Testing

- 34 new unit tests, one file per behavior above: `test_live_chain_retention.py` (retain and adopt), `test_live_chain_slot_spill.py` (contention spill), `test_live_chain_partial_adopt.py` (partial adoption), `test_live_chain_rotating_boundary.py` (rotating/CacheList provider routing), `test_live_chain_payload_inheritance.py` (payload inheritance), and `test_prefill_snapshot_namedtuple_state.py` (the NamedTuple regression).
- Full suite on the isolated branch (pure `v0.6.1` + only this series, none of the other unrelated local patches this tree normally carries): 9,935 passed, 86 skipped. Seven failures remain in `test_web_search.py` and `test_dflash_muse_glimmer.py`; they reproduce identically on unmodified `v0.6.1` in the same environment, where those optional dependencies are not installed.
- This exact branch (not only the tree it was developed on) was exercised end-to-end against a GDN hybrid model with retention enabled: a ~17.7K-token conversation went 24.8s cold to 2.0s on the adopting turn (17,708 cached) with the answer unchanged; a second conversation taking the slot spilled to SSD and the displaced conversation's next turn returned in 3.4s with 16,384 cached; a turn diverging at the tail took the safe fallback. No errors in the server log.
- `tests/test_prefix_cache_gdn_split.py` carries a one-line change: it stubs out the adoption entry point by name, and the name it used — a leftover from a downstream branch — is not the one this PR actually adds. Updated so the stub keeps doing its job.
- One defect only the full suite caught during development: a test fixture for the partial-adopt path was thinner than a real payload (missing fields a real one always carries), which hid a defensive-read gap in the payload inheritance. Worth flagging for anyone extending this — a stub fixture more permissive than reality hides the breakage it should catch.

## Known open item

Live adoption rate on Gemma-style models appears sensitive to how the client continues the conversation. One continuation pattern under abundant memory showed near-total adoption failure (23/31 mismatch, always `common prefix N-3 of N`), while a different pattern under real memory pressure adopted reliably (13 retained, 5/5 adopted). Not fully characterized — flagging for reviewers rather than claiming it's closed.

This is a performance-characterization gap, not a correctness risk: every mismatch case in both runs took the safe fallback path documented under Correctness above (full re-prefill, no incorrect output). A low adoption rate on some Gemma continuation patterns means those turns get less speedup, not wrong answers — and since the whole feature is opt-in (see above), it costs nothing for deployments that leave it off.

Two more latency (not correctness) caveats around contention: the turn that actually loses a contended slot still pays full cold-prefill cost for that one turn — spill writes the outgoing chain through to SSD as the slot is replaced, so recovery starts the *next* turn, not immediately. And a genuine two-stream in-flight contention test left one turn at `cached=0` for a reason not yet root-caused (no contamination, just a miss); three or more concurrent streams with sustained overlap haven't been tested at all.

## A note on this PR's commit structure

The six sub-patches referenced above (base retention, contention spill, partial adoption, rotating-layout support, payload inheritance, NamedTuple hardening) were developed incrementally but all touch the same ~1,000-line region of `omlx/scheduler.py`, each building on local state the others had already introduced. Splitting them into a matching series of upstream-clean commits would require re-deriving each intermediate state by hand, so this PR ships them as a single commit covering all six, plus a second commit for the NamedTuple hardening (which touches a genuinely separate function, `_copy_containers`, untouched by the rest of the series). Completeness was checked helper by helper against the development tree, and by verifying that every method the added code calls is defined on the branch.

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -1592,6 +1592,17 @@ class _BoundarySnapshotProvider:
             if snap is not None:
                 yield snap
 
+    def add_in_memory(self, tc: int, snapshot: Any) -> None:
+        """Register an already-extracted snapshot at ``tc``.
+
+        The snapshot must be in ``_extract_cache_states`` dict format, like
+        every other in-memory entry this provider serves.
+        """
+        if not snapshot:
+            return
+        self._in_memory[tc] = snapshot
+        self._valid_tcs.add(tc)
+
     def commit_gdn_checkpoint(
         self,
         token_count: int,
@@ -2457,8 +2468,7 @@ class Scheduler:
                 # delete now that the future has completed. Cleanup was
                 # deferred from _cleanup_finished to avoid racing the worker's
                 # boundary_snapshot_store.load() calls with rmtree.
-                if self._boundary_snapshot_store is not None:
-                    self._boundary_snapshot_store.cleanup_request(request_id)
+                self._cleanup_retention_boundary_snapshots(request_id)
                 # Worker no longer holds extracted_cache — pop request from
                 # self.requests and drop the cache buffer references so MLX
                 # arrays can be freed.
@@ -5236,6 +5246,15 @@ class Scheduler:
         if uids:
             _register_uid_rows(self.model, uids, [state.sampler], [per_row_lps])
             uid = uids[0]
+            # Snapshot rotating layer state at the prompt boundary
+            # (offset = prompt_len - 1; the last prompt token is fed to
+            # insert). Rotating caches cannot be sliced back later, and the
+            # generated output region never re-tokenizes identically (channel
+            # markers are stripped by the API layer), so this is the only
+            # point where a reusable rotating state exists. Bounded cost:
+            # window-sized buffers per rotating layer.
+            if getattr(self, "_live_chain_retention_enabled", lambda: False)():
+                self._capture_prompt_boundary_state(request, state.cache)
             self.request_id_to_uid[request.request_id] = uid
             self.uid_to_request_id[uid] = request.request_id
             now = time.monotonic()
@@ -6040,6 +6059,46 @@ class Scheduler:
             self.block_aware_cache, "_gdn_split_layout_supported", None
         )
         return bool(callable(supported) and supported(layer_types))
+
+    def _paged_cached_prefix_tokens(self, prompt_token_ids) -> int:
+        """Read-only probe: prompt tokens the paged prefix cache already holds."""
+        cache = getattr(self, "block_aware_cache", None)
+        if cache is None or not prompt_token_ids:
+            return 0
+        paged = getattr(cache, "paged_cache", None)
+        if paged is None or not getattr(paged, "enable_caching", False):
+            return 0
+        try:
+            _shared, remaining = paged.find_shared_prefix(list(prompt_token_ids))
+            cached = len(prompt_token_ids) - len(remaining)
+        except Exception:
+            return 0
+        return max(0, min(cached, len(prompt_token_ids) - 1))
+
+    def _split_gdn_snapshot_provider(
+        self,
+        request_id: str,
+        total_tokens: int,
+    ) -> _BoundarySnapshotProvider | None:
+        """Return a lazy provider for every durable full-block GDN stage."""
+        if not self._gdn_split_active():
+            return None
+        snapshots = self._boundary_cache_snapshots.get(request_id) or {}
+        block_size = self.config.paged_cache_block_size
+        valid_tcs = sorted(
+            tc
+            for tc, marker in snapshots.items()
+            if marker is None and 0 < tc <= total_tokens and tc % block_size == 0
+        )
+        if not valid_tcs:
+            return None
+        return _BoundarySnapshotProvider(
+            store=self._boundary_snapshot_store,
+            request_id=request_id,
+            valid_tcs=valid_tcs,
+            in_memory_snapshots={},
+            paged_ssd_manager=self.paged_ssd_cache_manager,
+        )
 
     def _eval_snapshot_cache(self, snapshot_cache: list[Any]) -> None:
         """Force the leaf KV tensors of an in-memory boundary snapshot concrete.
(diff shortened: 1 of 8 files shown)


hojin12312:
Hi @jundot, I wanted to add one follow-up observation from running this on a memory-constrained Mac.

The live-chain path works very well for sequential multi-turn conversations and avoids keeping a CPU hot-cache copy. One edge case I ran into is interleaved sessions: when an incoming request cannot adopt the retained chain, either fully or partially, the new request can begin prefill while the previous Metal-resident chain is still retained. Because the outgoing chain is spilled only when the replacement chain completes, peak memory can temporarily include both the retained chain and the new request's prefill/decode workspace.

This seems especially relevant with `hot_cache=0`. The measured spill times in my local server logs were:

| Retained chain | Spill time |
| ---: | ---: |
| 10,240 tokens | 132 ms |
| 12,288 tokens | 211 ms |
| 106,496 tokens | 658 ms |

These are local measurements of the SSD spill itself, so the exact end-to-end delay may vary slightly with SSD load and the time needed to release Metal buffers. Still, the overhead appears to be well under a second to roughly a second in these runs, which seems reasonable compared with a cold prefill or a memory-pressure failure.

My recommendation would be to keep the current fast path for exact and partial live-chain adoption, but add a spill-before-admission barrier only when the incoming request cannot use the retained chain. In that case, the server would spill the retained chain to SSD, wait for the spill and release/reclaim to complete, and then admit the new request. This should preserve the main performance benefit while avoiding the worst memory overlap between sessions.

If this direction seems useful, I would be happy to revise the PR and push an updated version. I may be missing something in the intended lifecycle, so corrections are very welcome.

PR #2814, opened 2026-08-18
Added ANE tuning utility. Modified ANE split to support NAX GPUs.

onthehub97:
- Added genuine group-size-128 NAX kernels for Q4/Q5/Q6/Q8.
- Updated both single- and dual-ANE hybrid GPU suffixes to use M5 tensor units (M5 Ultra in October, I swear!)
- Preserved automatic fallback to classic Metal if NAX or its metallib is unavailable.
- Preserved OMLX_QWEN35_QMM_NAX=0 as a kill switch.
- Removed the blanket M5 ANE disable; ANE prefill remains opt-in per model.
- Added a SwiftUI & Web UI “Tune ANE Split” utility:GPU-only baseline.
- Five MLP fractions.
- Five combined MLP/GDN fractions.
- Full 2,048-token prewarm.
- Two-block, 4,097-token end-to-end measurements.
- Two samples per candidate.
- Temporary configurations only.
- Separate Apply action.
- Recommends GPU-only when the best ANE result is under 1% faster.
- Changed messaging on settings to remove suggestions towards 0.5/0.53 which are unlikely to be tuned optimums.

Validation:

- Native Objective-C++ and Metal targets compile.
- Packaged CPython 3.11 extension loads successfully.
- Packaged metallib contains the new group-size-128 NAX symbols.
- 264 focused Python tests pass.
- Swift application compiles; 284/285 tests pass. The remaining failure is an unrelated theme-colour assertion in untouched code.

Benchmarking (provided by X user JamesMeadlock) shows notable, but not as dramatic as prior chips, improvements on M5 versus GPU alone, nearly reaching 1k tok/s on the best tuner run and achieving a 5% improvement on the practical benchmark: https://al-engr.com/omlx-dual-ane-three-mac.html#:~:text=7.%20Tuner%20on%20your%20branch%20(August%2018%2C%209%3A47%20AM)

I've removed the blanket M5 disable as the regression in NAX-compatible cards should now be resolved, leaving the env var in as a killswitch only, but not going to die on that hill - will gladly add it back.


jundot:
This is genuinely impressive work, thank you. The tuner is exactly what this feature needed, the transient-settings approach with the sub-1% GPU-only guard is well thought out, and you even landed the web UI follow-up before I finished reviewing. I verified the whole thing on my M3 Ultra: the group-size-128 NAX kernels build clean with exactly the symbol names the hybrid suffix assembles, and a live tuner run on a Qwen3.5 9B walked all 11 candidates in about 2.5 minutes, recommended MLP 45% + GDN 40% at +25.3% over GPU-only, and left persisted settings untouched, which is precisely the promised behavior.

Merging as is. I'll push two small follow-ups myself (the qwen ANE doc still describes the removed NAX auto-skip, and a minor quoting detail in the settings modal), so nothing needed on your side. Also thanks for the M5 benchmark link, it is good to see the ANE split earning its keep on NAX hardware, and I'd welcome more field numbers from M5 owners once this ships.

FaisalFehad:
This is amazing thank you! I spent this entire afternoon experimenting with this settings to find the sweet spot for M5 Max. I would be super useful if we apply the same concept to other model settings for performance and accuracy.

PR #2817, opened 2026-08-18
fix(responses): surface max_output_tokens truncation as status="incomplete"

mvdbos:
## Summary

The OpenAI Responses API path hardcoded `status="completed"` on both the non-streaming `ResponseObject` and the streaming `response.completed` event, discarding the engine's `finish_reason="length"` signal. This PR emits `status="incomplete"` with `incomplete_details.reason="max_output_tokens"` whenever `finish_reason == "length"`, in both paths.

## Problem

On a `max_output_tokens` truncation the engine correctly sets `finish_reason="length"` (patches/mlx_lm_mtp/batch_generator.py), and the Chat Completions API forwards it (`finish_reason = "tool_calls" if tool_calls else output.finish_reason`, server.py; streaming `finish_reason=chunk.finish_reason` in adapters/openai.py). But the Responses API has no `finish_reason` field and hardcoded `"completed"`, so the truncation signal is silently dropped.

Concrete failure (DeepSeek-V4-Flash-0731, `max_output_tokens=32768`, truncated mid-thinking):

- response `status` = `"completed"`
- output = `[{type:"reasoning", summary:[...16k tokens...]}, {type:"message", content: []}]`

A client (e.g. the pi.dev coding agent) reads the empty `message` item, sees `status="completed"`, treats the no-tool-call turn as a finished answer, and settles with a 0-byte patch. With this PR it would see `status="incomplete"` + `incomplete_details.reason="max_output_tokens"` and can continue instead.

## Fix

- Add `incomplete_details: Optional[Dict[str, str]]` to `ResponseObject` (omlx/api/responses_models.py).
- Non-streaming: `status="incomplete"` + `incomplete_details` when `output.finish_reason == "length"`.
- Streaming: same on the final `response.completed` event when `last_output.finish_reason == "length"`.

## Tests

- `TestResponseObject.test_incomplete_details_on_truncation` — `incomplete_details` survives `model_dump(exclude_none=True)`.
- `TestResponseObject.test_completed_response_omits_incomplete_details` — completed responses stay clean.

`tests/test_responses.py` passes (86 passed).

### DIFF
--- a/omlx/api/responses_models.py
+++ b/omlx/api/responses_models.py
@@ -223,3 +223,4 @@ class ResponseObject(BaseModel):
     metadata: Optional[Dict[str, str]] = Field(default_factory=dict)
     truncation: Optional[str] = None
     error: Optional[Dict[str, Any]] = None
+    incomplete_details: Optional[Dict[str, str]] = None
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -6303,9 +6303,13 @@ async def create_response(
                 cached_tokens=output.cached_tokens,
             )
 
+            # Surface max_output_tokens truncation so clients can tell an
+            # incomplete turn from a natural stop. The Responses API has no
+            # finish_reason field; status + incomplete_details is the signal.
+            truncated = getattr(output, "finish_reason", None) == "length"
             response_obj = ResponseObject(
                 model=request.model,
-                status="completed",
+                status="incomplete" if truncated else "completed",
                 output=output_items,
                 usage=usage,
                 tools=request.tools or [],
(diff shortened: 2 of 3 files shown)


jundot:
Thanks for fixing the dropped Responses API length signal. The non-streaming path and the streamed response payload now expose `status="incomplete"` with `max_output_tokens`, which fixes the reported client behavior. One thing I'll fold into an immediate follow-up is emitting `response.incomplete` instead of `response.completed` as the terminal SSE event when truncated, with an endpoint-level regression test. LGTM, merging.

PR #2819, opened 2026-08-18
fix(cluster): two-Mac clustering reliability (poll jitter, VLM tensor default, cross-user paths, ssh_user)

alytaphoenix:
## Summary

Reliability fixes for two-Mac clustering, found during a real Mac-mini + MacBook session where the cluster failed in several silent or misleading ways. Each fix is independent; grouped here as one set since they were diagnosed together. A companion design doc (`docs/cluster-ux-design.md`) proposes the deeper follow-ups.

## Fixes

**1. Status poll no longer wipes plan state / tensor choice** (`dashboard.js`)
`syncClusterNodesFromPeers()` ran on every 2s/10s poll and unconditionally called `invalidateClusterPlan()`, nulling `clusterPlan` + its signature (so "Activate manual plan" dead-buttoned itself and error banners were erased ~every 10s), and could reset the tensor-parallel size to 1 when the poll transiently reported a single node. Now the plan is invalidated only when the node set actually changes, and an out-of-range tensor size snaps to the largest degree (tensor) rather than 1 (pipeline). Complements #2721, whose own message noted a follow-up was needed.

**2. Poll doesn't resync mid-activation** (`dashboard.js`)
`startCluster()` runs a multi-step autoconfigure→stage→activate sequence; a mid-flight poll resync could make `autoconfigureCluster()` discard its own response and abort Start with no error. The poll now skips node resync while any cluster load flag is set.

**3. `supports_pipeline=False` for mlx-vlm checkpoints** (`planner.py`, +tests)
`_supports_pipeline` trusted the mlx-lm text backbone's source-level `pipeline()` for a VLM checkpoint (non-empty `vision_config`), which is actually served by mlx-vlm and has no `model.model.pipeline` — the attribute `progressive_loading` gates on. The UI offered/defaulted to pipeline, then the runtime 400'd with "the architecture does not implement the MLX-LM pipeline forward path." Now an explicit `SUPPORTS_PIPELINE` declaration wins (minimax_m3_vl stays True), else a vision sub-config reports False.

**4. Resolve model paths in each peer's own home** (`staging.py`, `routes.py`, `launch.py`, `inference_worker.py`, +tests)
A locally-sourced deployment baked the coordinator's absolute model path (`/Users/alice/...`) into the deployment; a peer with a different macOS account (`/Users/bob/...`) has no such path, so staging's present-probe came back empty and re-copied the whole model (then failed the disk check), preflight said "model directory is missing," and the worker `--model` was unresolvable on peer ranks. Now the portable `~`-form is sent and resolved in each node's own home; the staging destination is threaded explicitly (defaults to the source path, so same-home clusters are unchanged).

**5. Advertise each node's real login name** (`probe.py`, `models.py`, +tests)
oMLX SSHes to Mac peers with a bare hostname and no `user@`, so OpenSSH resolves the login against the caller's account or the caller's `ssh_config User` — the wrong user on a cluster with differing account names ("Permission denied", no hint that the username was at fault). Nodes now advertise their real login short name (from the password DB by UID, not the display name) in `ClusterStatus.node.ssh_user`. Consuming it to build explicit `user@host` (Bonjour TXT + dashboard) is proposed as a follow-up in the design doc, alongside #2672.

## Testing

- New unit tests for #3 (VLM False / text True / explicit-declaration override), #4 (remote present-probe + copy target the peer destination dir; no re-copy when present), and #5 (login name is the short name, env fallback, status advertises it).
- All modified modules import cleanly and `dashboard.js` passes `node --check`; test bodies were run manually (the packaged env has no `pytest`).
- **Not yet run:** a full live 2-node end-to-end (requires deploying this branch to both Macs). Happy to follow up with that validation.

## Not included
- The dead `127.0.0.1:9000` memory-ceiling fast-path is left as-is (minor; uncertain intended port; the fallback works). The design doc proposes retiring it via an advertised admin port.

See `docs/cluster-ux-design.md` for the pairing / diagnostics / Fabric-Doctor design these fixes point toward.

### DIFF
--- a/omlx/admin/static/js/dashboard.js
+++ b/omlx/admin/static/js/dashboard.js
@@ -3277,7 +3277,7 @@
                     );
                 const fit = this.clusterCatalogueFit(model.model_path);
                 if (fit?.fits === true) {
-                    const tensorSize = Number(fit.tensor_parallel_size || 1);
+                    const tensorSize = this.clusterFitTensorParallelSize(fit);
                     if (this.clusterTensorParallelOptions().includes(tensorSize)) {
                         // The catalogue arrived at this topology using the
                         // same fit planner. Preview that proven topology rather
@@ -3354,6 +3354,21 @@
                 return nodes > 1 ? [1, nodes] : [1];
             },
 
+            // The catalogue fit is the FEWEST-node plan: a model that fits one
+            // Mac reports tensor_parallel_size=1 even when two Macs are wired.
+            // Adopting that 1 into a 2-node plan builder made every preview a
+            // pipeline plan, which 400s for architectures without the MLX-LM
+            // pipeline forward path. Translate the fit into the topology valid
+            // for the nodes actually configured.
+            clusterFitTensorParallelSize(fit) {
+                const nodes = Math.max(1, this.clusterPlanNodes.length);
+                const fitted = Number(fit?.tensor_parallel_size || 1);
+                if (Number(fit?.nodes_required || 1) >= nodes) return fitted;
+                if (fit?.supports_pipeline === false) return nodes;
+                // Pipeline is possible too: keep whatever is selected.
+                return Number(this.clusterPlanTensorParallelSize) || 1;
+            },
+
             normalizeClusterTensorParallelSize() {
                 const options = this.clusterTensorParallelOptions();
                 const current = Number(this.clusterPlanTensorParallelSize);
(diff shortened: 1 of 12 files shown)


alytaphoenix:
Added two follow-on fixes to this branch, found while driving a real 2-node activation end-to-end:

- **make a single-node-fitting model plan tensor across 2 nodes** — `assess_model` returns the fewest-node plan (tp=1) for a model that fits one Mac, and the dashboard adopted that tp=1 into a 2-node builder → pipeline plan → 400 for architectures without the MLX-LM pipeline forward path. `ModelFit` now carries `supports_pipeline`/`supports_tensor_parallel`; the dashboard translates the fit into the topology valid for the configured nodes (tensor across all when pipeline is unavailable) and damps the ~0.1 GiB admission-ceiling wobble that was re-planning every 10s. Also: the memory-ceiling probe now hits the configured server port (was a hardcoded `:9000` no cluster serves), and `discover_remote_python_executable` preserves absolute launcher paths instead of unwrapping the worker shim to a bare interpreter (ModuleNotFoundError).
- **resolve peer home in the staging preview too** — `stage_manifest` (autoconfigure's readiness preview) probed every peer at the coordinator-absolute path, so a cross-user peer reported an already-present model as entirely missing; now resolves each peer's dir in its own $HOME like the execution path.

With these, a 2-node deploy of a text LLM both Macs hold reaches `ready_to_activate: true` and activates. (Separately confirmed VLM checkpoints are rejected at the activation gate — clustering is text-LLM-only today; a design to lift that is in progress.)

jundot:
Thanks, I confirmed these are all real bugs on main, and the cross-user path work is the part I'm happiest to take as is. Merging. I'll push follow-ups for a few small things on top, some guard conditions in the new dashboard logic, dropping ssh_user until something consumes it, the source-side path for peer-owned models, and an import sort. Nothing needed on your side.

PR #2821, opened 2026-08-18
fix(setting): accept SOCKS proxy URLs in network settings validation

asxzy:
## Problem

The oMLX web dashboard and macOS app reject SOCKS5 proxy URLs like `socks5h://127.0.0.1:1080` in the network settings fields, while setting the same proxy via environment variables (`HTTP_PROXY=socks5h://127.0.0.1:1080`) works correctly.

## Root Cause

The proxy validation in `omlx/settings.py` (lines 1601-1615) only accepted URLs starting with `http://` or `https://`, rejecting SOCKS proxy URL schemes.

## Fix

Extended the proxy validation to accept all standard proxy URL schemes:
- `http://`, `https://`
- `socks5://`, `socks5h://`

## Files Changed

- `omlx/settings.py` — Extend proxy validation to accept SOCKS URL schemes
- `tests/test_settings.py` — Add tests for SOCKS5 and SOCKS4 proxy validation

### DIFF
--- a/omlx/settings.py
+++ b/omlx/settings.py
@@ -1599,19 +1599,22 @@ class GlobalSettings:
                 )
 
         # Network proxy validation
+        # Schemes both client stacks can actually dial: httpx (via socksio)
+        # and requests (via PySocks). socks4/socks4a are httpx-unsupported.
+        _PROXY_SCHEMES = ("http://", "https://", "socks5://", "socks5h://")
         if self.network.http_proxy:
             proxy = self.network.http_proxy.strip()
-            if proxy and not proxy.startswith(("http://", "https://")):
+            if proxy and not proxy.startswith(_PROXY_SCHEMES):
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for addressing the settings validation gap. I can reproduce that the new allowlist accepts schemes the runtime cannot use: httpx rejects `socks4`/`socks4a`, while Requests raises `InvalidSchema` for SOCKS URLs because the package installs `socksio`, not PySocks. Please align the allowlist and dependencies with both client stacks, bump httpx to `>=0.28` if `socks5h` stays supported, and add a runtime client test.

asxzy:
Add support for socks4 would touch more codes. I believe supporting socks5 proxy is good enough for most use cases. Dropped support socks4 and leave socks5 only is good enough.

Also updated the dependency as suggested.

PR #2826, opened 2026-08-18
fix(qwen35) ANE tuner early exit results, q6/q8 support, GDN activation path fix

onthehub97:
## Changes

ANE MLP eligibility previously required q4 gate/up weights, while GDN only accepted q4/q5.

The updated path now supports q6 and q8 MLP and GDN projections while preserving q6/q8 weights for the GPU suffix. The optimized fused q4 route remains unchanged.

The mlx-lm GDN dispatcher now gives the registered ANE backend first refusal before applying the standalone GPU QMM threshold. Previously, q8 GDN projections at the fixed 2048-token ANE shape were rejected because the q8 GPU kernel intentionally requires at least 16K tokens.

For short q8 suffix projections, including MLP down-projection and GDN `b`/`a` projections, the hybrid path uses the faster stock MLX affine implementation. Longer sequences retain the optimized native q8 route.

Profiling now explicitly categorizes generic q6/q8 MLP operations as MLP work. Previously these calls inherited the GDN category, causing application logs to report zero MLP operations and inflated GDN counts despite successful MLP dispatch.

The standalone benchmark utility can now load through the same mlx-lm text-model path used by the application. It also supports MLP-only runs, allowing the incremental throughput and numerical effect of GDN dispatch to be measured independently.

mlx-vlm should also now support all supported quants.

### Interrupted tuner runs

Interrupted or failed tuner runs now:

- Display the exact termination reason and completed/total count.
- Preserve completed throughput and speedup results.
- Show the full planned test matrix.
- Leave failed and uncompleted result cells blank.
- Avoid presenting an incomplete run as an optimal recommendation.

## Benchmarks

Measurements use the packaged application kernel and oMLX text-model loading path with a 2048-token prompt. Results are the median of three timed runs after warm-up.

| Qwen3.8 27B model | GPU PP | Dual ANE MLP PP | MLP gain | Dual ANE MLP + GDN PP | Total gain | GDN gain over MLP |
| q4.85 AWQ | 345.8 | 435.3 | +25.9% | **481.2** | **+39.1%** | +10.5% |
| q6 MLX | 341.6 | 425.3 | +24.4% | **456.2** | **+33.5%** | +7.3% |
| q8 oQ8e | 350.4 | 424.9 | +21.2% | **459.0** | **+31.0%** | +8.0% |

Each full-offload pass dispatched:

- 64 MLP procedures.
- 48 GDN procedures.
- Two resident, instance-pinned ANE programs.

Observed per-instance ANE duty cycle was approximately 38.6% for q4.85, 37.9% for q6, and 36.5% for q8.

### Numerical comparison against GPU

The benchmark uses a deterministic random-token stress input and compares the resulting hidden state and final-token logits against GPU-only execution.

| Model | Mode | Hidden cosine | Final-logit cosine | Top token |
|---|---|---:|---:|---|
| q4.85 | MLP only | 0.99957 | 0.99999 | Match |
| q4.85 | MLP + GDN | 0.98107 | 0.97002 | Match |
| q6 | MLP only | 0.99335 | 0.99981 | Match |
| q6 | MLP + GDN | 0.98632 | 0.95791 | Match |
| q8 | MLP only | 0.99965 | 0.99997 | Match |
| q8 | MLP + GDN | 0.98927 | 0.98081 | Match |

MLP-only offload remains very close to GPU execution. GDN offload produces larger accumulated differences because its projections feed recurrent state. The top predicted token matched in every test, but the GDN setting remains explicit so users can evaluate the throughput-versus-accuracy trade-off on representative workloads.

## Validation

- Packaged q6 MLP compiled and executed across both ANE instances.
- Packaged q6 GDN compiled and executed across both ANE instances.
- Packaged q8 MLP compiled and executed across both ANE instances.
- Packaged q8 GDN compiled and executed across both ANE instances.
- Runtime profiling confirmed 64 MLP and 48 GDN procedures per pass for q6 and q8.
- A simulated memory-allocation failure preserved 2/5 completed results while leaving the remaining three cells blank.
- All 49 focused tuner tests passed.
- The current dispatch regression suite passed 41 tests; 14 native-extension-dependent tests were skipped under the local Python 3.13 test environment.
- The native packaged kernels were exercised successfully through end-to-end q4.85, q6, and q8 benchmarks.

### DIFF
--- a/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
+++ b/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
@@ -1169,8 +1169,56 @@ private struct ExperimentalSection: View {
                             }
                             .buttonStyle(.omlx(.normal, size: .small))
                         }
+
+                        if let status = vm.aneTuningStatus {
+                            if let reason = status.terminationReason,
+                               !reason.isEmpty {
+                                Text(reason)
+                                    .font(.omlxText(10))
+                                    .foregroundStyle(
+                                        status.status == "error"
+                                            ? Color.red
+                                            : theme.textSecondary
+                                    )
+                                    .fixedSize(horizontal: false, vertical: true)
+                                    .multilineTextAlignment(.trailing)
+                            }
+
+                            if !status.results.isEmpty {
+                                VStack(spacing: 3) {
+                                    HStack(spacing: 8) {
+                                        Text("Test")
+                                        Spacer(minLength: 8)
+                                        Text("Prompt tok/s")
+                                    }
+                                    .font(.omlxText(9, weight: .semibold))
+                                    .foregroundStyle(theme.textSecondary)
+
+                                    Divider()
+
+                                    ForEach(status.results) { result in
+                                        HStack(spacing: 8) {
+                                            Text(result.label)
+                                                .lineLimit(1)
+                                                .foregroundStyle(
+                                                    result.state == "failed"
+                                                        ? Color.red
+                                                        : theme.textSecondary
+                                                )
+                                            Spacer(minLength: 8)
+                                            Text(aneCandidateResultText(result))
+                                                .monospacedDigit()
+                                                .foregroundStyle(theme.text)
+                                                .frame(minWidth: 78, alignment: .trailing)
+                                        }
+                                        .font(.omlxText(10))
+                                    }
+                                }
+                                .frame(width: 285)
+                            }
+                        }
                     }
-                    .frame(minWidth: 210, alignment: .trailing)
+                    .frame(minWidth: 285, alignment: .trailing)
                 }
                 if vm.qwen35AnePrefillEnabled {
                     Row(label: String(localized: "settings.experimental.qwen_ane.sequence.label",
@@ -1642,6 +1690,20 @@ private struct ExperimentalSection: View {
             recommendation.processingTps,
             recommendation.speedupPercent
         )
+    }
+
+    private func aneCandidateResultText(
+        _ result: ANETuningCandidateDTO
+    ) -> String {
+        guard let processingTps = result.processingTps else {
+            // Deliberately blank: the row remains visible so an interrupted
+            // run shows which tests did not complete.
+            return ""
+        }
+        if let speedup = result.speedupPercent {
+            return String(format: "%.1f (%+.1f%%)", processingTps, speedup)
+        }
+        return String(format: "%.1f", processingTps)
     }
 }
 
--- a/apps/omlx-mac/Sources/AppView/ViewModels/ModelSettingsScreenVM.swift
+++ b/apps/omlx-mac/Sources/AppView/ViewModels/ModelSettingsScreenVM.swift
@@ -748,9 +748,10 @@ final class ModelSettingsScreenVM {
                 aneTuningStatus = snapshot
                 if snapshot.status != "running" {
                     aneTuningIsRunning = false
-                    if snapshot.status == "error" {
-                        lastError = snapshot.error ?? "ANE tuning failed."
-                    }
+                    // Benchmark termination is rendered with its partial
+                    // matrix in the tuner row. Reserve lastError for transport
+                    // and settings failures so the reason is not duplicated.
+                    lastError = nil
                     break
                 }
                 try await Task.sleep(for: .seconds(1))
--- a/apps/omlx-mac/Sources/Net/DTO/BenchDTO.swift
+++ b/apps/omlx-mac/Sources/Net/DTO/BenchDTO.swift
@@ -240,9 +240,11 @@ struct ANETuningCandidateDTO: Codable, Equatable, Identifiable, Sendable {
     let mlpFraction: Double?
     let gdnEnabled: Bool
     let gdnFraction: Double?
-    let processingTps: Double
+    let state: String?
+    let processingTps: Double?
     let samples: [Double]
     let speedupPercent: Double?
+    let error: String?
 
     var id: String { label }
 }
@@ -268,6 +270,7 @@ struct ANETuningStatusResponse: Codable, Sendable {
     let results: [ANETuningCandidateDTO]
     let recommendation: ANETuningRecommendationDTO?
     let error: String?
+    let terminationReason: String?
 }
 
 struct ANETuningCancelResponse: Codable, Sendable {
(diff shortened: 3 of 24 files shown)


onthehub97:
The benchmarks aren’t AS impressive. Investigating what that is.

onthehub97:
Found the issue - and it's actually an issue that other users have been having: GDN simply wasn't activating most of the time.

I've fixed the GDN activation path, and it's bought up the benchmark results considerably, with a hidden "cost": the cosine distance increases by a statistically significant amount, though not significant enough to alter the response meaningfully in my tests.

This is should be a consideration of using the GDN ANE path - and it's not new with this change, it's been there since the beginning, just hadn't yet surfaced.

beaglemoo:
Tested #2826 (`c4a48727`) on the M5 Pro, re-run this morning on a **cold**
machine (~9 h idle, no zombie Metal compiles, nothing else loaded). Built from
source with `OMLX_WITH_CUSTOM_KERNEL=1`; all four kernel families available,
`is_nax_available() == True`. I checked which process actually held :8843 each
time — the app will otherwise quietly keep the port and you end up benchmarking
the release build.

Model is `Qwen3.8-27B-oQ4e-mtp` — **q4**, so the q6/q8 eligibility work does not
apply here. Reporting anyway because the GDN result may still be useful.

## GDN compiles but never executes on the q4 route

Main finding. The procedures are now compiled:

```
Eagerly compiled 64 MLP and 48 GDN procedures into 2 instance-pinned ANE programs
```

but none of them run:

```
[benchmark-ane-profile] category=mlp  operations=126 expected_operations=128 ane0_duty=0.19
[benchmark-ane-profile] category=gdn  operations=0   expected_operations=96  ane0_duty=0.00
```

`configured_layers=48`, `expected_operations=96`, `operations=0` on every GDN
candidate, warm and cold. So on this box the `+GDN` rows are MLP-only
re-measurements, and the tuner's GDN comparisons are not measuring GDN.

Happy to dig into the q4 predicate if you point me at it — I did not want to
guess at which check bails.

## Cold tuner run

```
GPU only              442.8      —
MLP 15%               448.0   +1.2%
MLP 25%               458.6   +3.6%   <- best
MLP 35%               453.7   +2.5%
MLP 45%               451.5   +2.0%
MLP 53%               439.5   -0.7%
MLP 25% + GDN 15%     435.9   -1.5%
MLP 25% + GDN 25%     446.2   +0.8%
MLP 25% + GDN 35%     450.6   +1.8%
```

The MLP-only spread (+1.2% to +3.6%) is close to run-to-run noise, and adding
"GDN" never beats the best MLP-only row — consistent with it not executing.

Also worth noting against v0.6.2, where this same machine reported MLP 35% at
**+10.4%** and MLP 35% + GDN 45% at **+17.8%**. On #2826 the best is **+3.6%**.
That is a large correction and it supports your "all prior tuner results are
inaccurate".

Both runs (warm last night, cold this morning) stopped at 9/11 with
`PrefillMemoryExceededError`. The new early-exit handling worked exactly as
described — reason and 9/11 shown, completed results kept, remaining cells
blank, no recommendation offered from a partial run. Clear improvement.

## Real requests, cold, interleaved

Applying the cold best (MLP 0.25), real `/v1/chat/completions`, cache cleared
and model unloaded before every run, OFF/ON **interleaved in pairs** so drift
lands on both arms:

**~16K** — ANE loses consistently:

```
pair 1  OFF 42.77s   ON 49.40s
pair 2  OFF 46.63s   ON 54.03s
pair 3  OFF 49.28s   ON 55.09s
median  OFF 46.63s   ON 54.03s   ->  -13.7%
```

**~32K** — this is the interesting one:

```
pair 1  OFF 105.87s  ON 115.44s     GPU faster by 9.6s
pair 2  OFF 116.75s  ON 116.08s     level
pair 3  OFF 123.05s  ON 118.28s     ANE faster by 4.8s
median  OFF 116.75s  ON 116.08s  ->  +0.6%
```

I would not report that +0.6% as "ANE wins at 32K". What it actually shows is
that the two arms **degrade at different rates**: GPU-only goes 105.9 -> 123.1 s
(+16%) across three pairs, while ANE goes 115.4 -> 118.3 s (+2.5%). They cross
over between pair 1 and pair 2. From a genuinely cold start GPU-only is ~9%
ahead; under sustained load the ANE arm holds its throughput and the GPU arm
does not.

That is the thermal-relief mechanism @onthehub97 described in #2781, showing up
here as convergence rather than as a win. It also means a single median over a
warming run is the wrong summary statistic for this comparison on a thermally
constrained machine — worth bearing in mind for the tuner too, which measures
candidates sequentially.

## Where that leaves it on this chip

At q4 on an M5 Pro, MLP-only ANE offload is a clear loss at 16K and roughly a
wash at 32K once thermal behaviour is accounted for. Since GDN never executes
here, none of this is a verdict on the GDN work itself.

## Environment

M5 Pro, 64 GB, macOS 26.6.2, `Qwen3.8-27B-oQ4e-mtp` (MTP on, draft depth 8,
TurboQuant off), `dual_ane: false` (single die), `sequence_length: 2048`
matching the scheduler's delivered prefill chunk.

onthehub97:
@beaglemoo Thank you for that thorough testing.

Would you be able to run the tuner on this branch also: https://github.com/jundot/omlx/pull/2853

With both default tuner settings, and with CPU disabled?

I would like to see how this performs on M5 series chips. I'm not expecting the CPU to deliver meaningful improvement on the M5, as the M5 NAX pathway is already damn near optimal, but the tuner itself has improved significantly to use a heuristic calculation-driven profile-aware approach instead of a brute-force scan of coarse values. I think we should be able to get some reliable M5 gains from the ANE.

(These are the same heuristic calculations I used manually to determine the winning work split on the main M5 benchmark that showed an actual improvement, the tuner hasn't worked for M5 to this point)

beaglemoo:
Ran the PR #2853 tuner on the M5 Pro as requested (defaults and with CPU disabled) — results posted over there: https://github.com/jundot/omlx/pull/2853#issuecomment-5342640895. Short version: the new tuner rejects CPU on its own, GDN now executes on q4 (96/96 ops vs 0 on this branch), both runs completed all slots, but the recommended split still loses on interleaved real requests here (-27.2% at 16K). Also filed #2880 for a prefill-chunk finding that affects the sequence_length discussion.

onthehub97:
Closing this PR as #2833 has merged with most of the changes; tuner early exit changes are rendered obsolete by tuner v2.

PR #2827, opened 2026-08-18
perf(laguna): port mlxfast-challenge decode optimizations (bit-exact verified)

0xClandestine:
Ports the correctness-verified Laguna XS 2.1 decode optimizations from the Layr-Labs/mlxfast-challenge Swift model into omlx's vendored \`laguna_model.py\`, on branch \`perf/mlx-fast-laguna\`.

## What's in

**Compiled single-output fusions (`mx.compile(shapeless=True)`) — default ON.**
- top-k router renormalization, weighted expert combine, per-head softplus gate.
- residual-variant expert combine (decoder residual folded into the compile).
- compiled SiLU(gate)\*up product.
- All single-output graphs -> bit-exact vs eager; measured **+3.96% single-token decode** (69.1 vs 66.4 tok/s) on a real \`poolside/Laguna-XS-2.1-NVFP4-mlx\` checkpoint (M4 Max), greedy trajectory token-identical.

**Fusion banks (bit-exact, measured neutral-or-slower on current MLX) — opt-in OFF.**
- fused routed gate/up NVFP4 gather-QMM decode bank (`OMLX_LAGUNA_FUSED_ROUTED_GATE_UP=1`).
- fused shared-expert gate/up NVFP4 bank (`OMLX_LAGUNA_FUSED_SHARED_GATE_UP=1`).

## Correctness

Every ported optimization is bit-exact vs the stock path (parity tests on one model instance; real-checkpoint greedy trajectory token-identical). The Swift's two-output compiled router tail is NOT ported (a two-output `mx.compile` with a shared sigmoid intermediate diverges at ULP in MLX 0.32 and feeds expert selection) — kept eager, documented. Complete catalog and challenge-commit attribution in \`docs/laguna-mlxfast-port-correctness.md\`.

## Tests

\`tests/test_laguna_patch.py\` + \`tests/test_dflash_laguna.py\` (51 tests) green; full omlx suite (\`pytest -m \"not slow\"\`) passes (9943).

### DIFF
--- a/docs/laguna-mlxfast-port-correctness.md
+++ b/docs/laguna-mlxfast-port-correctness.md
@@ -60,39 +60,6 @@ migration `4799830` and are not part of the current model surface).
 - **Optimization:** re-represent the BF16 attention projections (`q/k/v/o/g_proj`, ~2.9 GB of the ~4.3 GB per-decode-step weight traffic) as group-32 affine INT8 at init — 9 bits/weight vs BF16's 16, removing ~1.25 GB/step.
 - **Token-exactness issue:** this is an explicitly **lossy** re-quantization (the submission itself documents it; it lives inside the track envelope's permitted "representation set" but is not bit-exact vs the reference). It is not reproduced in oMLX: the port bar requires bit-exact parity or a documented exception. If oMLX ever adopts it, it must be a default-OFF toggle with lossiness documented and a token-diff gate against the BF16 reference.
 
-## Pre-Laguna submissions: triage (not ported as code)
-
-104 Validate/Accept submission commits exist in the challenge repo total. 8
-are Laguna-era (rows 93-100 above). The other 96 target the challenge's two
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for porting and carefully validating these Laguna decode optimizations. The default-on compiled fusions preserve output parity and provide a meaningful single-token decode improvement, while the neutral or slower fused-bank paths remain safely opt-in and disabled by default. I also verified identical tokens with the compiled path enabled and disabled on a local Laguna checkpoint. LGTM, merging.

0xClandestine:
> Thanks for porting and carefully validating these Laguna decode optimizations. The default-on compiled fusions preserve output parity and provide a meaningful single-token decode improvement, while the neutral or slower fused-bank paths remain safely opt-in and disabled by default. I also verified identical tokens with the compiled path enabled and disabled on a local Laguna checkpoint. LGTM, merging.

Thanks for the review/merge. Will likely followup later this week with more ported commits once I get a chance.

Issue #2779, opened 2026-08-17
ANE prefill regresses on M5 (NAX) hardware; qwen35_ane_prefill doesn't consult is_nax_available()


beaglemoo:
## Summary

The experimental Qwen3.5 ANE prefill added in 0.6.1 is a **regression on M5-class
(NAX) hardware** — it costs both prefill and decode. `qwen35_ane_prefill.py`
never consults `is_nax_available()`, unlike `qwen35_fa256_attention.py`, which
already skips itself on NAX GPUs for what looks like the same underlying reason.

## Environment

- Apple **M5 Pro**, 20 GPU cores, 64 GB, macOS 26.6.1
- oMLX **v0.6.1**, source install, kernels built with `OMLX_WITH_CUSTOM_KERNEL=1`
- mlx 0.32.0 / mlx-metal 0.32.0, `OMLX_NAX=1`
- `Qwen--Qwen3.8-27B-oQ4e-mtp` (local oQ4e + preserve_mtp conversion),
  `mtp_enabled: true`, `mtp_num_draft_tokens: 8`
- `qwen35_ane_available()` -> `True`, `is_nax_available()` -> `True`

## Measurements

~17.6K-token prompt, predictable output (keeps MTP acceptance ~98% so the
verify path is actually exercised), 200 generated tokens, single request in
flight, **`~/.omlx/cache` cleared before each configuration**, 3 runs each.

| Config | prefill (TTFT) | decode tok/s |
|---|---:|---:|
| ANE off | 53-64s (median 54.7) | **32.9** |
| `qwen35_ane_prefill_enabled: true` (defaults, `dual_ane: true`) | 65-80s | 30.3 |
| `qwen35_ane_prefill_enabled: true`, `dual_ane: false` | 67-88s | 30.0 |

Log confirms it engaged:

```
Eagerly compiled 64 MLP and 48 GDN procedures into two instance-pinned ANE programs (sequence_length=2048)
Qwen ANE prefill scheduler aligned to fixed shape 2048
```

`dual_ane: false` was the obvious hypothesis, since the reference M3 Ultra is
two dies and the M5 Pro is one — it made no difference, so this does not look
like a tuning problem.

## Suspected mechanism

`omlx/patches/qwen35_fa256_attention.py` already documents the general shape of
this for a different kernel:

> on NAX GPUs (M5 family) stock SDPA's head_dim-256 fallback runs its matmuls on
> the tensor units and beats this pre-NAX steel kernel, so routing it would
[...]

PR #2829, opened 2026-08-18
fix(ane): reject tuning candidates whose ANE never executed

### LINKED ISSUE #2828: ANE split tuner can publish a ranking from runs where the ANE never executed
## Summary

The ANE split tuner added in #2814 can publish a full candidate ranking, and a
recommendation, from measurements in which **the ANE never executed a single
operation**. It fails silently — the run completes normally and the numbers look
plausible.

The trigger is `sequence_length` not matching the scheduler's prefill chunk
size. That is easy to hit, because tuning at your typical prompt length is the
natural thing to try.

## Environment

- M5 Pro, 64 GB, macOS 26.6.2
- oMLX 0.6.2 (official DMG)
- `Qwen3.8-27B-oQ4e-mtp`, MTP enabled (draft depth 8), TurboQuant off
- `scheduler.prefill_step_size` at its 2048 default

## Reproduction

```bash
curl -X POST localhost:8843/admin/api/bench/ane-tune/start \
  -d '{"model_id":"Qwen--Qwen3.8-27B-oQ4e-mtp","sequence_length":8192,"repeats":2}'
```

The run completes and reports every ANE candidate as slower than GPU-only:

```
GPU only    381.9 tok/s
MLP 15%     351.2   -8.0%
MLP 25%     361.1   -5.4%
MLP 35%     366.8   -4.0%
```

But the ANE was idle throughout:

```
[benchmark-prefill] chunk_tokens=2048 requested_step=8192 ane_shape=False
[benchmark-ane-profile] category=mlp operations=0 configured_layers=64
    expected_operations=128 observed_shapes=0.000
    ane_region_ms_per_op=0.000 ane0_duty=0.0000 ane1_duty=0.0000
```

The programs *do* compile —

```
Eagerly compiled 64 MLP and 0 GDN procedures into 2 instance-pinned ANE
programs (sequence_length=8192)
```

— but the scheduler prefills in 2048-token chunks, so every chunk fails the
fixed-shape check. Each "ANE" candidate therefore measures GPU-only throughput
*plus* the cost of compiling and pinning programs it never uses, which is why
they rank slightly below the GPU-only baseline.

Re-running the same model with `sequence_length: 2048` immediately gives
`ane_shape=True`, `operations=126/128`, `observed_shapes=1.969`,
`ane0_duty=0.1655` — and a completely different ranking (best candidate +17.8%).

## Why the existing guard does not catch it

`_measure_candidate()` calls `_ane_is_active(engine)`, which checks
`_omlx_ane_mlp_prefill_count` / `_omlx_ane_gdn_prefill_count` on the model.
Those are set at **compile** time. In the failing case they are non-zero, so the
guard passes — compiled is not the same as executed.

`ANETuningRequest.validate_sequence_length` only enforces "a multiple of 64,
>= 1024"; nothing relates it to `SchedulerConfig.prefill_step_size`. And
`_log_ane_benchmark_trace()` already computes `operations` and
`observed_shapes`, but only logs them, so the tuner cannot see that its own
measurement was empty.

## Impact

Not M5-specific. Any user who tunes at a `sequence_length` other than their
scheduler's chunk size gets a confident, complete, wrong answer — and the
natural conclusion is "ANE prefill does not help on my machine". Worth noting
`prefill_step_size` is configurable (the 4096 comment at `scheduler.py:1747`),
so the correct value is not always 2048.

To be clear this is *not* what I hit in #2779: that report predates the tuner
and used `sequence_length: 2048`, so the ANE was engaging there. This is a
separate defect in the new tuning path.

## Suggested fix

Fail a candidate whose ANE performed zero operations rather than recording its
throughput, and name `prefill_step_size` in the error so the fix is obvious.
PR follows.

## Related

- #2814 — the tuner this affects
- #2824 — sibling silent-no-op (`qwen35_ane_prefill_enabled` on non-4bit checkpoints)
- #2779 — my original M5 ANE report

beaglemoo:
## What

The ANE split tuner can rank — and recommend — candidates whose ANE never ran a
single operation. This makes that case fail loudly instead.

Fixes #2828

## Why

`_measure_candidate()` guards with `_ane_is_active()`, but that checks
`_omlx_ane_mlp_prefill_count` / `_omlx_ane_gdn_prefill_count`, which are set when
the programs **compile**. If `sequence_length` does not match the scheduler's
prefill chunk size, every chunk fails the fixed-shape check and the ANE stays
idle — compiled, but never executed. The candidate then measures GPU-only
throughput plus the cost of compiling and pinning unused programs, and that
number is ranked against genuine candidates.

Observed on an M5 Pro at `sequence_length=8192` against a 2048-token chunk:
every enabled candidate reported `ane_shape=False`, `operations=0` of
`expected_operations=128`, `ane0_duty=0.0000`, and the tuner concluded ANE was
uniformly 4-8% slower than GPU-only. At `sequence_length=2048` the same model
gives `operations=126/128` and a different ranking entirely.

## How

- `_log_ane_benchmark_trace()` already computes the per-category operation
  counts; it now **returns** them instead of only logging them.
- `_run_single_test()` attaches them to its metrics as `ane_trace`, but only
  when ANE tracing was requested, so non-ANE callers are unaffected.
- `_measure_candidate()` raises for an enabled candidate whose repeats all show
  zero ANE operations, and names the scheduler's `prefill_step_size` in the
  message so the correct `sequence_length` is obvious.

GPU-only baselines are untouched — they carry no ANE trace by design, and the
guard only applies to `candidate.enabled`.

## Tests

Six added to `tests/test_ane_tuning.py`:

- `_ane_execution_observed()` distinguishes compiled-but-idle from actually-ran
- ...and returns `None` (unknown) when the profiler was unavailable
- `_prefill_step_size()` reads scheduler config and degrades safely
- a compiled-but-idle candidate raises, and the message names the chunk size
- an executing candidate, and a GPU-only baseline, both still pass
- a candidate measured without the profiler is kept, not rejected

All fail without the change. Full run of `tests/test_benchmark.py
tests/test_ane_tuning.py tests/test_bench_external_routes.py
tests/test_benchmark_sse_replay.py tests/test_context_benchmark.py
tests/test_qwen35_ane_prefill.py` — **230 passed**.

## Verified end to end on hardware

Built with `OMLX_WITH_CUSTOM_KERNEL=1` and run against a real server on the
M5 Pro (all four kernel families available, `is_nax_available() == True`).

Failing case now fails loudly instead of publishing a ranking:

```
POST /admin/api/bench/ane-tune/start {"sequence_length": 8192}
-> status: error
   The ANE compiled but never executed for sequence_length=8192: each prefill
   chunk failed the fixed-shape check. Measuring this candidate would report
   GPU-only throughput as an ANE result (the scheduler prefills in 2048-token
   chunks, so set sequence_length=2048).
```

Working case is unaffected — at `sequence_length: 2048` the run proceeds
normally through the enabled candidates with no false positive.

## Note on the profiler

The operation counters come from the ANE profiler, which is optional:
`qwen35_ane_profile_set_enabled()` can return `False` and its import sits
behind a bare `except`. In that case every counter reads zero regardless of
what the hardware did. An earlier draft of this patch treated that as an idle
ANE and would have rejected perfectly good candidates on any machine without
the profiler, so the trace now carries `profiling_available` and the guard
treats a missing profiler as *unknown* rather than as evidence of idleness.

### DIFF
--- a/omlx/admin/ane_tuning.py
+++ b/omlx/admin/ane_tuning.py
@@ -143,6 +143,31 @@ def _settings_for_candidate(base: Any, request: ANETuningRequest, candidate: _Ca
     return settings
 
 
+def _prefill_step_size(engine: Any) -> int | None:
+    """The scheduler's prefill chunk size, if it can be determined."""
+    config = getattr(engine, "_scheduler_config", None)
+    value = getattr(config, "prefill_step_size", None)
+    try:
+        return int(value) if value else None
+    except (TypeError, ValueError):
+        return None
+
+
+def _ane_execution_observed(trace: dict[str, Any] | None) -> bool | None:
+    """Whether the ANE ran ops, or None when that cannot be determined.
+
+    The operation counters come from the ANE profiler. When the profiler is
+    unavailable they are all zero no matter what the hardware did, so that
+    case must be reported as unknown rather than as an idle ANE.
+    """
+    if not trace or not trace.get("profiling_available"):
+        return None
+    return any(
+        int(values.get("operations", 0) or 0) > 0
+        for values in (trace.get("categories") or {}).values()
+    )
+
+
 def _ane_is_active(engine: Any) -> bool:
     model = getattr(engine, "_model", None)
     if model is None:
@@ -190,6 +215,7 @@ async def _measure_candidate(
         pass
 
     samples: list[float] = []
+    traces: list[dict[str, Any] | None] = []
     for _ in range(run.request.repeats):
         metrics = await _run_single_test(
             engine=engine,
@@ -211,6 +237,31 @@ async def _measure_candidate(
             ),
         )
         samples.append(float(metrics["processing_tps"]))
+        if candidate.enabled:
+            traces.append(metrics.get("ane_trace"))
+
+    observations = [_ane_execution_observed(trace) for trace in traces]
+    # Only fail on a positive observation of an idle ANE. If the profiler was
+    # unavailable every observation is None and the guard must stay out of the
+    # way rather than rejecting a candidate that may well have run.
+    if candidate.enabled and any(o is False for o in observations) and not any(observations):
+        # The programs compiled (checked above) but no ANE operation ran, so
+        # this candidate really measured GPU-only plus the cost of compiling
+        # and pinning unused programs. Reporting its throughput would rank a
+        # never-executed configuration against real ones.
+        step = _prefill_step_size(engine)
+        hint = (
+            f"the scheduler prefills in {step}-token chunks, so set "
+            f"sequence_length={step}"
+            if step
+            else "sequence_length must match the scheduler's prefill chunk size"
+        )
+        raise RuntimeError(
+            "The ANE compiled but never executed for "
+            f"sequence_length={run.request.sequence_length}: each prefill chunk "
+            f"failed the fixed-shape check. Measuring this candidate would "
+            f"report GPU-only throughput as an ANE result ({hint})."
+        )
 
     return {
         "label": candidate.label,
--- a/omlx/admin/ane_tuning.py
+++ b/omlx/admin/ane_tuning.py
@@ -144,13 +144,29 @@ def _settings_for_candidate(base: Any, request: ANETuningRequest, candidate: _Ca
 
 
 def _prefill_step_size(engine: Any) -> int | None:
-    """The scheduler's prefill chunk size, if it can be determined."""
+    """The prefill chunk size the scheduler will request, if determinable.
+
+    This is ``max(config.prefill_step_size, scheduler._qwen35_prefill_floor)``,
+    mirroring Scheduler._prefill_chunk_size(). The floor matters: on a >= 64 GB
+    machine serving a qwen3_5 model it raises the request to 4096, so reading
+    the config value alone would name the wrong sequence_length.
+
+    Note the *delivered* chunk can still be smaller — _guard_prefill_chunk
+    shrinks it when the predicted peak approaches the Metal cap — so callers
+    should present this as a starting point, not a guarantee.
+    """
     config = getattr(engine, "_scheduler_config", None)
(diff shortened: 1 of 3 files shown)


beaglemoo:
Pushed a follow-up commit while testing something unrelated on the same machine.

`_prefill_step_size()` was reading `config.prefill_step_size` alone, but
`Scheduler._prefill_chunk_size()` takes `max(config.prefill_step_size,
_qwen35_prefill_floor)`. On a >= 64 GB machine serving a `qwen3_5` model that
floor is 4096, so the new error message would have suggested
`sequence_length=2048` while the scheduler was actually requesting 4096 —
wrong advice in precisely the situation the message exists to resolve.

It now mirrors the scheduler's own calculation, and hedges appropriately: the
*delivered* chunk can be smaller than the requested step, because
`_guard_prefill_chunk` shrinks it when the predicted peak approaches the Metal
cap. I hit exactly that on a 64 GB M5 Pro — `chunk_tokens=2048` against
`requested_step=4096` — so the message now points at `chunk_tokens` in the serve
log as the authority rather than asserting a single number.

231 tests pass. Detail of the 64 GB floor behaviour is in
#2779 if it's useful; it may be worth knowing separately that the 4096 floor
never survives the memory guard on that class of machine.

jundot:
Thanks for this, the compiled versus executed distinction is exactly what the guard was missing, and treating a missing profiler as unknown rather than idle is the right call. Merging as is.

I'll push a few small follow ups myself: profiling_available will come from the snapshot contents rather than the enable call (an ABI skewed extension currently reads as available with all zero counters, which would reject a good candidate), the fixed shape wording in the error becomes one possible cause among several, and _prefill_step_size drops the floor branch since configure_qwen35_ane_prefill_scheduler zeroes the floor on any engine this guard can fire on.

I'll also fold your measure_length fix from the #2853 thread into main, that one affects every tuner number 0.6.2 has reported.

PR #2830, opened 2026-08-18
fix(responses): honor tool_choice=none, fix MCP tool merge

astutegrogan:
## Summary

Four small, independent bugs in the tool-calling path.

- `/v1/responses` gated its MCP tool merge on the client having sent tools of its own, silently dropping MCP-provided tools whenever the client sent none. `/v1/chat/completions` has always merged unconditionally — this makes `/v1/responses` match.
- `/v1/responses` never checked `tool_choice=="none"` before exposing tools to the chat template, so an explicit "no tool calls" request still had tools visible to the model. Same fix, matches `/v1/chat/completions`' existing `tools_disabled` semantics.
- The XML tool-call parser's `<function=NAME>` regex was `\w+`, rejecting hyphenated/dotted tool names (e.g. `get-weather`, `web.search`) — the whole call was dropped.
- The sibling `<parameter=NAME>` regex had the identical bug, one line away — worse in effect, since it doesn't drop the call, it silently drops the argument, so a call to e.g. `unit-type` looks like it succeeded with an empty `arguments` dict.

All four widened/fixed to `[A-Za-z_][\w.-]*`, matching the character class already used elsewhere in this file for tool-name matching.

## Implementation

Scoped to `omlx/api/tool_calling.py` (both regexes) and `omlx/server.py`'s `create_response()` (the two `/v1/responses`-only fixes). No changes to `/v1/chat/completions`, which already had correct behavior for both. No scheduler, routing, or config changes.

## Verification

```
pytest -q
9917 passed, 90 skipped, 77 deselected, 0 failed
```

Regression coverage in `tests/integration/test_responses_tool_merge.py` (MCP-merge-ordering and `tool_choice=none`-suppression, exercised via a real FastAPI TestClient against the actual endpoint) and `tests/test_tool_calling.py` (hyphenated/dotted name round-trips for both regexes, proving the argument is no longer silently dropped).

### DIFF
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -471,7 +471,14 @@ def _xml_element_value_end(text: str, start: int, close_tag: str, next_open: str
         search = close + len(close_tag)
 
 
-_XML_PARAMETER_OPEN_RE = re.compile(r"<parameter=(\w+)>")
+# Parameter names follow the same convention as tool/function names (see
+# _XML_FUNCTION_CLOSE handling above and _parse_bracket_tool_calls): they
+# must start with a letter/underscore and may contain word chars, dots, and
+# hyphens (e.g. "unit-type"). A bare \w+ here silently fails to match a
+# hyphenated or dotted parameter name, so the whole <parameter=...> element
+# is skipped and that argument is dropped from the parsed result even
+# though the surrounding tool call otherwise parses successfully.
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for covering these Responses API and XML parser gaps. I found one regression before merge: `[A-Za-z_][\w.-]*` narrows the previous `\w+` behavior, so valid digit-leading names such as `2fa` are now dropped as function names and silently omitted as parameter names. Please make the new patterns preserve digit-leading names and add regression cases for both paths; the endpoint changes otherwise look good.

PR #2840, opened 2026-08-18
feat(dflash): support DFlash 2 and update draft defaults

liang2kl:
## Summary

- Add DFlash 2 (https://inco.ai/blog/dflash2) runtime support by forwarding sampling parameters and an optional runtime block size to `dflash-mlx`. **Prerequisite: DFlash 2 PR (https://github.com/jundot/dflash-mlx/pull/6) in `dflash-mlx` being merged.**
- Expose `dflash_block_size` through model settings, profiles, the admin API, web UI, native macOS UI, benchmark exports, and engine reload detection. This enables overriding the draft block size declared in the checkpoint config.
- When the draft window is not explicitly configured, use the checkpoint's positive `config.sliding_window` value. Leave it unset when the checkpoint does not provide one.
- Change the draft sink-size default from `64` to `0`, while preserving explicit non-negative values. All DFlash 2 models and some DFlash models natively support sliding window attention so this is not necessary for those models.
- Update the web and macOS interfaces to expose the new runtime setting and reflect the updated defaults.

## Testing

- `204 passed` across:
  - `tests/test_dflash_draft_config.py`
  - `tests/test_dflash_engine.py`
  - `tests/test_dflash_muse_glimmer.py`
  - `tests/test_model_settings_profiles.py`
  - `tests/test_admin_model_settings_template.py`


jundot:
Thank you for wiring this all the way through, settings to admin api to both UIs to reload detection. I checked the block size clamp and the checkpoint window resolution against dflash-mlx and both work as described, and honoring sampling params in the dflash path is a welcome fix on its own.

On merge order: I'll merge the dflash-mlx PR first and bump the pin on main before merging this one, so there is no window where dflash calls fail on the missing sampling kwargs. After that I'll run smoke tests on my legacy DFlash pairs to see how the new sink default behaves there, drop the now redundant draft config normalizer patch, and plumb min_p and seed through in a follow-up. Appreciate the clean split between the two PRs, it made review straightforward.

PR #2843, opened 2026-08-19
feat(cluster): collision-checked, VPN-aware Thunderbolt fabric subnet selection

alytaphoenix:
Adds `choose_fabric_subnet()` — collision-checked, VPN-aware Thunderbolt subnet selection replacing the hardcoded `10.0.1.1/24` in `configure_link`.

On a Mac behind a full-tunnel VPN (e.g. Cloudflare WARP), `10.0.1.x` is inside the tunnel and silently swallowed — the fabric never carries traffic even though ping/SSH over the LAN work, which is near-impossible to diagnose. The new selector prefers `172.16.0.0/12` (commonly VPN-excluded) and skips any range already carried on either Mac; the empirical post-addressing `assess_link` check remains the readiness authority. Includes unit tests for preference order, collision skipping, and the all-occupied error.

### DIFF
--- a/omlx/cluster/transport.py
+++ b/omlx/cluster/transport.py
@@ -14,7 +14,7 @@ import secrets
 import shlex
 import subprocess
 import threading
-from collections.abc import Callable, Sequence
+from collections.abc import Callable, Iterable, Sequence
 from contextlib import contextmanager, suppress
 from dataclasses import dataclass
 from typing import Any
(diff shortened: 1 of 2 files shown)


alytaphoenix:
Closing for now — the Fabric-Doctor work this belongs to isn't complete yet; will reopen as part of that effort. (The subnet-selection code here is sound but should land with the rest of the diagnostics/doctor set.)

PR #2844, opened 2026-08-19
fix(cluster): make quantized Nemotron-H run tensor-parallel at TP=2

alytaphoenix:
## Summary

The hybrid Mamba2/attention + MoE `nemotron_h` architecture could not run under distributed tensor parallelism. At TP=2 it crashed mid-shard; once that was fixed it failed on the first inference. This PR fixes four independent bugs and adds a planner guard, so a quantized Nemotron-H (e.g. Nemotron-3.5-Lightning-30B-A3B, oQ4) shards and generates across two nodes.

Verified end-to-end on a **2-Mac Thunderbolt-5 / JACCL cluster** (M4 Pro 64 GB + M4 Pro 48 GB): the 30B-A3B model shards at TP=2, activates, and generates 220 coherent tokens.

## Root causes & fixes

1. **Routed-expert MoE down-projection can't be evenly split.** `switch_mlp.fc2` is row-parallel over the 1856-wide intermediate = **29 quantization groups** (`group_size` 64). 29 is prime, so the even `mx.split` inside `shard_inplace` raises `Array split does not result in sub arrays with equal size … shape (128,2688,29)`. Replaced the `fc1`/`fc2` `shard_inplace` calls with an **uneven, group-aligned split** (`_shard_switch_mlp_uneven`): rank 0 owns 15 groups, rank 1 owns 14. The recombining `all_sum` in `_wrap_sharded_moe` is shape-agnostic, so the unequal widths sum back exactly (verified to fp noise, ~1e-5 on signal magnitude 80). The larger shard is assigned to rank 0 (the coordinator, typically the higher-memory node).

2. **Quantized Mamba `in_proj` sliced weight-only.** The row-gather at `in_proj.weight[indices]` left `scales`/`biases` full-height against a half-height weight — a silent shape mismatch on the first Mamba forward. Now sliced with the same `indices`.

3. **`pipeline_call` rejected `n_confirmed`.** The base MTP patch threads `n_confirmed` through every model `__call__`; the Nemotron-H pipeline hook (installed as `NemotronHModel.__call__`) didn't accept it (`TypeError: … unexpected keyword argument 'n_confirmed'`). It now accepts and ignores it (MTP is inactive on the distributed path).

4. **`pipeline_call` assumed the pipeline-parallel path.** It read `pipeline_rank`/`pipeline_size`/`fa_idx`/`ssm_idx`, which only `pipeline()` sets — never called on the pure tensor-parallel path (`AttributeError: 'NemotronHModel' object has no attribute 'pipeline_rank'`). It now defaults to a single stage (so send/recv/all_gather no-op) and recomputes the first attention/SSM cache indices locally.

5. **Planner honesty.** `_tensor_parallel_divisors` now also contributes the quant-group counts of the *even-split* (`shard_inplace`) row-parallel projections (attention/Mamba `out_proj`, shared-expert `down_proj`), so a TP degree the head counts allow but the quantization forbids is refused up front rather than crashing mid-load. The routed `fc1`/`fc2` are deliberately excluded — they use the uneven path and only need `group_count >= degree`.

## Tests

- `tests/test_cluster_tensor_strategies.py` (new): uneven group-range arithmetic, and a numeric-equivalence test — a 2-rank uneven split of a real quantized `SwitchLinear` MoE reconstructs the unsharded output to fp noise.
- `tests/test_cluster_planner.py`: the quant-group divisor guard (58 present, 29 excluded; TP=2 viable, TP=4 rejected) and the unquantized case.

All cluster tests pass; no regressions in `test_cluster_progressive_loading.py`.


jundot:
Thanks, solid work. I re-derived the group arithmetic against the pinned mlx-lm and it checks out exactly, and the numeric equivalence test is the right kind of proof. Merging. I'll push follow-ups to make the planner guard refuse when per-module quant overrides are present, add the missing head_dim and width fallbacks, gather in_proj.bias with the same indices, and trim the unreachable fa_idx branch.

Issue #1261, opened 2026-05-15
qwen3.6 35b a3b  auto disabled vlm


lbm1202:
**Describe the bug**
When loading Qwen3.6 35B-A3B in oMLX 3.9 dev2, the model loads as an LLM with VLM auto-disabled. In version 0.3.8, the same model loaded correctly as a VLM and image input worked properly.

**To Reproduce**
1. Launch oMLX 3.9 dev2
2. Load model
3. Model loads as LLM, VLM is auto-disabled

**Expected behavior**
The model should load as a VLM (as it does in 0.3.8) and accept image input.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - macOS Version: [e.g. 26.4.1]
 - oMLX Version [e.g. 0.3.9 dev2]

**Additional context**

sleepyeldrazi:
Can you point to the exact huggingface repo you used for the qwen? Does it happen on a specific quant only? And if you provide the server logs (Logs tab in the webui) right after you load the model in, there might be information there that will help debugging.

PR #2845, opened 2026-08-19
feat(cluster): serve VLM checkpoints text-only under distributed inference

alytaphoenix:
## Summary

VLM checkpoints (Qwen3.5 / 3.6 `*ForConditionalGeneration`) were rejected from distributed cluster inference outright, even though the pinned mlx-lm loads and shards their **language model** text-only (its `sanitize` drops the vision tower). This adds an explicit, honest **text-only** opt-in so a VLM's LLM can run tensor-parallel across nodes, while un-flagged requests keep the current refusal (no silent vision drop — see #1261 / #1426).

Verified on a **2-Mac Thunderbolt-5 / JACCL cluster** (M4 Pro 64 GB + 48 GB): both **Qwen3.6-35B-A3B** and **Qwen3.8-27B** activate text-only at TP=2 and generate; image requests return a clear 400.

## Changes

- **`text_only` flag** on `ClusterDeploymentRequest` and `ClusterDeployment` (persisted; round-trips through the registry and to peers).
- **Gate relaxation + routing** (`engine_pool.py`): the three gates (`resolve_cluster_model_id`, `register_cluster_model`, `_distributed_deployment_for_entry`) accept a `vlm` entry when the deployment is `text_only`; such an entry now routes through `DistributedBatchedEngine`. Un-flagged VLMs keep the honest error with an added remedy sentence. Both VLM start-failure fallbacks now fail closed instead of silently relabelling the entry `batched` and dropping vision.
- **Capability probe** at activation: a `text_only` request whose planner layout reports neither tensor-parallel nor pipeline support is refused before any peer contact (`plan_hybrid` only checks divisor divisibility, not shardability).
- **Planner text-only accounting** (`inspect_safetensors_layout`): exclude vision-tower and MTP tensors from layer/fixed sizing for VLM checkpoints. Their names collide with the decoder-layer pattern (`vision_tower.blocks.N` → layer N, `language_model.mtp.layers.0` → layer 0), which mis-sized stages. The Jundot 35B now measures **18.83 GiB with uniform layers** instead of 20.13 GiB with an inflated layer 0.
- **Pipeline-model resolution** (`_validate_loaded_stage` / `_loaded_stage`): locate the layer owner via a `language_model` / `model` fallback (`_common_layer_owner`), which also fixes pure-TP **text** Qwen3.5/3.6 (the family's `Model` wraps `language_model`).
- **Admission** uses the text-only size estimate, not the vision-inflated on-disk size.
- **Image-request rejection**: for a text-only deployment, reject image content with a clear 400 at the chat route (where content is still structured), instead of silently stripping it; defense-in-depth check in the distributed engine too.

## Tests

`tests/test_cluster_text_only.py` (new): text-only layout excludes vision/MTP (and pure-text is unaffected), wrapper-shaped pipeline-stage validation (+ classic shape, + fail-closed on mismatch), image detection, and the engine-level guard. Gate acceptance/refusal added to `tests/test_cluster_engine_pool.py`. All cluster suites pass.

## Notes

- MTP / thinking-budget / dflash / specprefill / turboquant-kv remain incompatible with distribution (existing gate); `-mtp` checkpoints need those settings off, as for any distributed model.
- Tier (b) (full VLM with the vision tower on rank 0) is out of scope; mlx-lm's server has no image path. This is tier (a) from the design.

### DIFF
--- a/omlx/cluster/deployment.py
+++ b/omlx/cluster/deployment.py
@@ -248,6 +248,13 @@ class ClusterDeployment:
     performance_profiles: tuple[NodePerformanceProfile, ...] = ()
     tensor_parallel_size: int = 1
     target_context_tokens: int = 8192
+    # Explicit user opt-in to serve a VLM-shaped checkpoint as a text-only
+    # distributed model: the pinned mlx-lm rank loader drops the vision tower
+    # during sanitize, so only the language model runs across ranks. The flag
+    # is persisted so peers and later server restarts keep honoring the
+    # opt-in; an un-flagged VLM deployment stays refused (silent vision drop
+    # is treated as a bug, #1261/#1426).
+    text_only: bool = False
 
     def __post_init__(self) -> None:
         if _NODE_ID.fullmatch(self.deployment_id) is None:
@@ -373,6 +380,7 @@ class ClusterDeployment:
             ],
             "tensor_parallel_size": self.tensor_parallel_size,
             "target_context_tokens": self.target_context_tokens,
+            "text_only": self.text_only,
         }
 
     @classmethod
@@ -414,11 +422,16 @@ class ClusterDeployment:
             ),
             tensor_parallel_size=int(payload.get("tensor_parallel_size", 1)),
             target_context_tokens=int(payload.get("target_context_tokens", 8192)),
+            text_only=bool(payload.get("text_only", False)),
         )
 
     def encode_worker_plan(self) -> str:
         """Encode the small trusted plan as a bounded command-line argument."""
 
+        # ``text_only`` deliberately stays out of the worker contract: ranks
+        # load through the pinned mlx-lm, whose sanitize drops vision weights
+        # for these checkpoints natively, so the worker needs no flag and the
+        # contract decoded by ``decode_worker_contract`` stays unchanged.
         raw = json.dumps(
             {
                 "schema_version": 1,
--- a/omlx/cluster/inference_worker.py
+++ b/omlx/cluster/inference_worker.py
@@ -645,13 +645,31 @@ def _runtime_assignment(
     return result
 
 
+def _resolve_pipeline_model(model: Any) -> Any:
+    """The module that owns the mutable transformer layer list, or None.
+
+    Most text architectures expose it as ``model.model``, but wrapper models —
+    the qwen3_5 / qwen3_5_moe VLM family served text-only — nest it under
+    ``language_model.model``. Reuse the tensor strategy's resolver so validation
+    and sharding agree on which object is the pipeline stage.
+    """
+
+    from .tensor_strategies import _common_layer_owner
+
+    try:
+        owner, _layers = _common_layer_owner(model)
+        return owner
+    except RuntimeError:
+        return None
+
+
 def _validate_loaded_stage(
     model: Any,
     assignment: PipelineAssignment,
 ) -> None:
     """Fail closed if a model-specific pipeline hook ignored the shard plan."""
 
-    pipeline_model = getattr(model, "model", None)
+    pipeline_model = _resolve_pipeline_model(model)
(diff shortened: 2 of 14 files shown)


jundot:
Thanks, the fail-closed change on the VLM start fallbacks and the layout accounting are good work, and the 2-Mac verification is appreciated. The blocker is the PR's own headline invariant: the image rejection only covers /v1/chat/completions, while /v1/messages still drops image blocks silently for a text-only deployment (preserve_images keys on is_vlm, which is False for the distributed engine), exactly the outcome #1261/#1426 refused. Please move the rejection to a single pre-extraction choke point covering both routes. Three more: reuse _VISION_WEIGHT_PREFIXES from model_discovery instead of the narrower new list, make the .mtp. exclusion unconditional rather than VLM-gated (a pure-text -mtp checkpoint has the same layer-0 collision), and please key the size exclusion on the deployment's text_only intent rather than the checkpoint shape, so the catalogue keeps advertising full sizes for un-flagged VLMs. Heads up for the rebase: after #2819, _supports_pipeline returns False for vision configs, which makes your capability probe effectively tensor-only for VLMs, worth reconciling.

alytaphoenix:
Rebased onto current main (2 conflicts, both from main independently landing overlapping plumbing since this branch forked) and pushed fixes for everything in the review.

**The named blocker**: extracted `_reject_image_content_for_text_only_deployment` as a single pre-extraction choke point in `server.py` and wired it into both `/v1/chat/completions` and `/v1/messages`, so the fail-closed rejection now covers both routes instead of just chat/completions. `/v1/messages` was silently dropping images via `preserve_images=is_vlm` (False for the distributed engine) exactly as you found.

**The three follow-ups**:
- `_VLM_VISION_PREFIXES` is gone; the planner's vision-tensor exclusion now reuses `_VISION_WEIGHT_PREFIXES` from `model_discovery`.
- MTP-tensor exclusion (the `language_model.mtp.layers.0` / decoder-layer-0 collision) is unconditional now, not gated on VLM-ness -- a pure-text `-mtp` checkpoint has the identical collision and was getting zero exclusion before.
- `inspect_safetensors_layout`/`complete_model_layout` gained a `text_only` parameter driven by the caller's own deployment intent (`ClusterDeployment.text_only` / a new `ClusterPlanRequest.text_only`), not the checkpoint's shape. The catalogue (no `text_only` passed) keeps advertising full sizes for un-flagged VLMs; a concrete text-only deployment sizes against what it will actually load. Layout cache key is now `(path, text_only)` so the two measurements don't collide.

**The #2819 heads-up**: `_supports_pipeline`'s vision-subconfig guard now only applies when `text_only=False`. A text-only deployment loads through plain mlx-lm -- the same `mlx_lm.models.<model_type>` module a pure-text checkpoint of that architecture would use, not mlx-vlm's wrapper -- so the guard's premise (mlx-vlm's loaded wrapper never exposes `model.model.pipeline`) doesn't hold for that path, and the capability probe was wrongly reporting no pipeline support for every VLM text-only deployment regardless of what its text backbone actually supports.

**Two things I left out, noted rather than silently dropped**:
- `remote_model_layout` (the SSH-probed peer-layout path) doesn't yet thread `text_only` through to the remote snippet, so a text-only deployment whose `model_source` is a remote peer still sizes full -- safe direction (over-, not under-, estimate), but incomplete. Left as a follow-up rather than rushing a versioned-SSH-snippet change into this review.
- `/v1/responses` has the same `preserve_images` keyed on `is_vlm` gap as the two routes I fixed, but wasn't part of this review's named blocker and its request shape (`request.input`, not `request.messages`) doesn't fit the same detector without more design work. Flagging for a separate follow-up.

Full 10,479-test suite passes. 5 tests (new + updated) proven to fail against the pre-fix code via stash-revert, so they actually discriminate.

PR #2848, opened 2026-08-19
feat(cluster): transport readiness ladder with reason/remedy copy (fabric-doctor B3)

alytaphoenix:
## ⚠️ Stacks on #2819

This is the first of the **fabric-doctor** series and builds on the probe/transport changes in **#2819** (`local_login_name`, the `LinkStatus` construction sites, `ssh_user`). Until #2819 merges, the diff below **includes #2819's commits** — review only the B3 commit (`feat(cluster): transport readiness ladder…`); the diff collapses to just that once #2819 lands.

## Summary

Every rung between "cable in" and "collective proven" becomes a **named transport state carrying a plain-English `reason` + `remedy`**, rendered verbatim in the CLI and GUI. This directly fixes an opaque failure mode seen in practice: when the Thunderbolt link/authorization isn't up, activation returned a misleading `400 "no tensor-parallel degree divides 2 nodes across architecture dimensions (…)"` and an RDMA probe hung ~30s — with no hint that the actual problem was "the cable/link is down." Now the transport reports, e.g., *"The link has a self-assigned 169.254 address — macOS never finished configuring it → Run Fabric Doctor → Re-address link."*

## What changed

- **New `omlx/cluster/readiness.py`** — `LADDER_ORDER` (9 rungs: `enabled_no_peer → disabled → unavailable → peer_linked_config_pending → addressed → routed → reachable → fabric_verified → collective_ok`), `TransportReadiness`, a copy table **total over the enum** (every state has non-empty reason AND remedy — "states without copy don't ship"), a pure `link_ladder_state()` derivation, and `node_readiness()`.
- **`models.py`** — `TransportState` gains `ADDRESSED/ROUTED/REACHABLE/FABRIC_VERIFIED/COLLECTIVE_OK`; the existing four wire values are byte-identical and `protocol_version` stays `"1.0"` (additive). `ClusterStatus.to_dict()` gains `transport.readiness = {state, reason, remedy}`.
- **`probe.py`** — node-local decision promotes past `PEER_LINKED_CONFIG_PENDING` on **evidence only** (routable IP on an RDMA interface → `ADDRESSED`; route via an RDMA interface → `ROUTED`; a 169.254 self-assigned address stays pending with a stale-address reason). Tops out at `ROUTED` — single-host evidence can't prove `REACHABLE`. CLI prints indented `reason:`/`remedy:` below `Transport:` when under `COLLECTIVE_OK`.
- **`transport.py`** — `LinkStatus` gains `ladder`/`reason`/`remedy`, populated at every `classify_link`/`assess_link` return; the unreachable branch names Fabric Doctor in its remedy.
- **`guidance.py`** — `Guidance.code` + `_BY_CODE` + `explain_code()` (structured lookup wins over message-regex, preserving the never-raise/never-None contract); a new `stale_link_address` rule for the 169.254 case. `/guidance` accepts an optional `code`.
- **`_cluster.html`** — renders `transport.readiness` reason/remedy under the Transport card (the full status strip is deferred to B4).

## Tests

`tests/test_cluster_readiness.py` (32: ladder total over the enum, ranks strictly increasing, legacy wire values unchanged, every state has non-empty reason+remedy, table-driven evidence→rung, node-local-never-REACHABLE), plus additions to `test_cluster_guidance.py`, `test_cluster_probe.py`, `test_cluster_link_status.py`. On the #2819 base: **128 passed, 0 failed** across the touched cluster suites.

### DIFF
--- a/omlx/admin/static/js/dashboard.js
+++ b/omlx/admin/static/js/dashboard.js
@@ -3277,7 +3277,7 @@
                     );
                 const fit = this.clusterCatalogueFit(model.model_path);
                 if (fit?.fits === true) {
-                    const tensorSize = Number(fit.tensor_parallel_size || 1);
+                    const tensorSize = this.clusterFitTensorParallelSize(fit);
                     if (this.clusterTensorParallelOptions().includes(tensorSize)) {
                         // The catalogue arrived at this topology using the
                         // same fit planner. Preview that proven topology rather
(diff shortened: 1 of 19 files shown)


jundot:
Thanks for the work here. The ladder vocabulary is worth having, but half of it isn't wired yet (the readiness card warns on healthy clusters because the probe tops out at routed, and link_ladder_state has no callers), so I'd rather take it fully connected together with the doctor rework than land it half-wired now. I already took the Guidance.code part onto main since #2866 needed it. Closing for now, please fold the ladder into the reworked doctor series.

PR #2849, opened 2026-08-19
fix(cluster): require bidirectional bound-TCP for link reachability (fabric-doctor C1)

alytaphoenix:
## ⚠️ Stacks on #2848 → #2819

Fabric-doctor series. Builds on **#2848** (B3 readiness ladder), which builds on **#2819**. Until those merge the diff includes their commits — review only the C1 commit (`fix(cluster): require bidirectional bound-TCP…`); it collapses to just that once the parents land. When #2848 lands, a passing bound connect is what earns the `REACHABLE` rung.

## Summary

Promotion to "reachable" now requires a **TCP connect bound to the fabric source IP to succeed in both directions** — the check that separates "cable works" from "a VPN/firewall on that Mac swallows inbound TCP". Previously the positive proof was one ICMP ping per direction, with the bound connect only a Linux fallback. But a corporate VPN (policy: *all interfaces*) can pass ICMP and route lookups while dropping TCP — which is exactly how the fabric silently failed in practice (LAN ping/SSH fine, peer→coordinator on the fabric refused).

## What changed (`transport.py: verify_link_reachability`)

- Per direction: **route check → bound TCP connect → (only on TCP failure) ping** as a diagnostic.
- The ping now distinguishes causes in the failure copy instead of being the success gate:
  - route + ping fine but TCP refused → *"{peer} cannot accept connections on the Thunderbolt link (a firewall or VPN on that Mac applies to all interfaces)"* — names the blocking Mac.
  - nothing answers → *"…did not answer on that address (no TCP, no ping)"*.
- macOS route-interface mismatch still fails fast; the Linux path (no `route -n get`) still relies on the bound connect alone. Bound-connect script hoisted to a module constant.
- Contract unchanged: `(bool, str)` with an injectable `runner`; `assess_link` picks it up for free.

## Tests (`test_cluster_link_status.py`)

Bound-TCP required both ways (ping never consulted on success); firewall case names the Mac + cause; host-down reason; Linux-without-route still passes. All pass on the stacked base.

## Note

Slightly stricter than before: links that passed on ping alone may now report unreachable — that's the point (empirical TCP probe outranks ICMP optimism), and it's the conservative-correct answer when a VPN drops inbound TCP.


jundot:
Thanks, the idea here is right, ping optimism on a VPN-afflicted link is a real failure mode. But the mechanism never runs on a remote host: _run_link_command hands ssh a multi-element argv, ssh space-joins it unquoted, and the multi-line python -c script arrives shell-mangled as a syntax error. So the remote direction always fails, ping then succeeds, and every healthy cluster gets the confident firewall accusation. The tests couldn't catch it because they fake the exact function containing the bug. For the rework: shlex.join the remote command in _run_link_command (the same file already does this right in _remote_gui_authorize), use /usr/bin/python3 instead of bare python3 (the packaged app has no python3 on PATH), add a unit test asserting the argv handed to ssh is one shell-safe word, and probe a port omlx owns (the collective port is the natural choice) instead of 22, which quietly made inbound ssh on the coordinator a new requirement. Could you run it live on the two-Mac setup before resubmitting? This class of bug is invisible to mocked runners, so the live check is the only thing that proves the probe.

alytaphoenix:
Split the root cause you diagnosed out into a standalone fix, independent of this PR's own remaining (now fairly stale) content: #3198.

Applied your exact recipe — `shlex.join` the remote command in `_run_link_command` (mirroring `_remote_gui_authorize`'s already-correct pattern in the same file), `/usr/bin/python3` instead of bare `python3`, and switched the bound-connect probe from port 22 to 8000 (oMLX's own default server port) so passing this check no longer requires inbound SSH scoped to the fabric interface as a side effect.

Verified live against the two-Mac cluster per your ask: reproduced the exact syntax error with the old argv shape over a real ssh round-trip, then confirmed the fixed shape completes a real bound connect from the peer's fabric address to the coordinator's on port 8000. Details in #3198.

This PR's own remaining scope (readiness-ladder additions) is still there and still needs the resubmit-after-rebase you asked for — just calling out that the blocker itself no longer needs to wait on that rework.

PR #2851, opened 2026-08-19
fix(specprefill): enable SpecPrefill for MLA-family models (DeepSeek/GLM)

anicaise-ai:
## What

Three commits that make SpecPrefill usable with MLA-family models (DeepSeek V2/V3/V4, GLM-5.x, glm4_moe_lite), follow-up to #2208:

1. **feat(specprefill): MLA query extractor for DeepSeek/GLM-family drafts** — MLA attention has no plain `q_proj` contract, so every draft in that family fell through the detector to `_llama_extract_queries` and crashed at scoring time with `'Glm4MoeLiteAttention' object has no attribute 'q_proj'`, after paying the full draft prefill. Adds a dedicated extractor covering both the low-rank-q (`q_a_proj`/`q_b_proj`) and full-rank-q MLA variants.
2. **fix(specprefill): preserve input dtype in manual RoPE helpers** — the manual RoPE path silently upcast to fp32, which broke dtype expectations downstream on bf16 models.
3. **fix(specprefill): keep DSA indexer and CacheList offsets on true positions** — with sparse-MLA/DSA targets, the indexer and CacheList were fed positions relative to the pruned sequence instead of the true positions, corrupting attention for the retained tokens.

## Tests

- 16 new unit tests (`tests/test_specprefill_mla.py`, `tests/test_specprefill_dsa_positions.py`), all passing.
- Running in production on an M3 Ultra 512 GB against `DeepSeek-V4-Flash` (MXFP4) and `GLM-4.7-Flash` (4-bit) on top of v0.6.2 — serving and benchmarks green (these patches have been carried on our fork since ~0.5.4 and rebase cleanly onto v0.6.2).

Happy to split into separate PRs if you prefer smaller reviews.

### DIFF
--- a/omlx/patches/specprefill.py
+++ b/omlx/patches/specprefill.py
@@ -162,6 +162,38 @@ def _nemotron_h_extract_queries(attn, x, cache=None, **kwargs):
     return queries
 
 
+def _mla_extract_queries(attn, x, cache=None, **kwargs):
+    """DeepSeek/GLM MLA: (optionally low-rank) q projection + partial RoPE.
+
+    MLA caches store keys in a split layout, so queries are built in the
+    matching layout. Absorbed variants (``embed_q``, e.g. glm4_moe_lite)
+    cache the compressed latent as keys and the roped positional slice as
+    values; queries are q_nope absorbed into latent space, concatenated
+    with the roped q_pe. Expanded variants (``kv_b_proj``) cache full
+    per-head keys; queries are concat(q_nope, roped q_pe) directly.
+    Queries are pre-scaled so the scorer's generic 1/sqrt(d) recovers the
+    module's true attention scale (including any yarn mscale correction).
+    """
+    B, L, D = x.shape
+    n_heads = getattr(
+        attn,
+        "num_heads",
+        getattr(attn, "num_attention_heads", getattr(attn, "n_heads", None)),
+    )
+    if getattr(attn, "q_proj", None) is not None:
+        q = attn.q_proj(x)
+    else:
+        q = attn.q_b_proj(attn.q_a_layernorm(attn.q_a_proj(x)))
+    q = q.reshape(B, L, n_heads, attn.q_head_dim).transpose(0, 2, 1, 3)
+    q_nope, q_pe = mx.split(q, [attn.qk_nope_head_dim], axis=-1)
+    offset = cache.offset if cache is not None else 0
+    q_pe = attn.rope(q_pe, offset)
+    if getattr(attn, "embed_q", None) is not None:
+        q_nope = attn.embed_q(q_nope)
+    queries = mx.concatenate([q_nope, q_pe], axis=-1)
+    return queries * (attn.scale * queries.shape[-1] ** 0.5)
+
+
 # ---------------------------------------------------------------------------
 # Model topology helpers
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


jundot:
Thanks for the detailed MLA and DSA work. I found two integration gaps before this can merge: `run_specprefill_target_prefill()` decrements the kickoff adjustment only on the attention RoPE, so the newly wrapped DSA indexer stays one position ahead during decode; and DSA draft scoring still passes a `CacheList` into `_mla_extract_queries()` and `_compute_importance()`, which crashes on `.offset` and `.keys`. Could you update the shared owner/cache traversal in those paths and add tests through `run_specprefill_target_prefill()` plus `score_tokens()` with a real `CacheList` layout?

anicaise-ai:
Thanks for the sharp review — both gaps were real, and chasing them surfaced a third of the same class. Fixed in ed6ea606:

- **Kickoff decrement**: `run_specprefill_target_prefill()` now walks every RoPE owner via `_rope_owners()` (the same traversal `sparse_prefill` wraps with), so the DSA indexer's adjustment is decremented alongside the attention's instead of staying one position ahead through decode.
- **CacheList in draft scoring**: `_mla_extract_queries()` resolves the offset through `_cache_offset()` (descends into the list), and `_compute_importance()` descends to the attention slot before reading `.keys`/`.values`.
- **Bonus, same family**: the lookahead trim in `score_tokens()` used `hasattr(entry, "offset")`, which silently skipped CacheList entries — lookahead KV persisted into stored draft caches. The trim now walks CacheList children.

Tests as requested: `run_specprefill_target_prefill()` with a DSA layer whose indexer rope is wrapped (both adjustments asserted), and `score_tokens()` end to end over a real mlx_lm `CacheList` layout (extraction, importance, and trim asserted on both child caches). Verified both tests fail on the previous head and pass with the fixes; full specprefill suites green locally (the two `TestEngineCorePropagation` failures and the `mlx_vlm.speculative` import error pre-exist this change on my machine).

anicaise-ai:
Field report: ed6ea606 is now running on our production DeepSeek-V4-Flash-0731 (MXFP4, M3 Ultra 512GB, dspark MTP depth 5) — deployed this evening, so treat this as first-hours validation rather than a long soak.

- Clean restart picked the fix up in ~35s (SSD paged cache warm).
- No speed regression against the pre-fix baseline measured earlier the same day on the same box: short decode 27.7 tok/s vs a 26.5–30.2 reference window; cold prefill of an 11.2k-token unique prompt unchanged.
- Long-context needle recall at 11k: correct.
- MTP speculative acceptance on post-fix traffic: 65–88% depending on content — consistent with (and at the top end of) the pre-fix 68–83% range. Too early to claim the indexer off-by-one was measurably hurting acceptance, but nothing regressed and the top-k selection is now provably at true positions.

I'll report back if overnight traffic surfaces anything. Happy to run any additional validation you'd like on this deployment before merge.

anicaise-ai:
Hi @jundot — no rush on this one, I know dev4 and the RC are taking your time.

Just a heads-up that GitHub expired this PR's workflow run today: it had been sitting in `action_required` since 19 Aug (fork PRs need a maintainer to approve the run), and the 30-day timeout marked it as failed. Nothing changed on our side — it just needs a re-approval whenever you get to it.

For context, everything from your review landed in `ed6ea606` the same day:

- the kickoff decrement now walks every RoPE owner via `_rope_owners()`, so the DSA indexer no longer stays a position ahead through decode;
- `_mla_extract_queries()` / `_compute_importance()` descend the `CacheList` properly;
- plus a third case of the same class we hit while chasing yours (the `hasattr(entry, "offset")` lookahead trim in `score_tokens()`).

Tests through `run_specprefill_target_prefill()` and `score_tokens()` with a real `CacheList` layout are in `test_specprefill_dsa_positions.py` and `test_specprefill_mla.py`.

It's been running on our production DeepSeek-V4-Flash (MXFP4, M3 Ultra 512GB, DSpark MTP depth 5) since then, no regression.

Happy to rebase if `main` has moved under it.

PR #2833, opened 2026-08-18
feat(qwen35): enable affine Q6 and Q8 ANE prefill

GordoAR:
## Summary

- Make Qwen3.5/3.6/3.8 ANE prefill selection aware of affine bit width and group size.
- Preserve the existing Q4 public symbols as thin `bits=4` delegates. Q6 and Q8 use the shared bit-aware affine functions.
- Let the registered ANE GDN backend handle eligible Q8 projections before the standalone Q8 GPU threshold.
- Keep Q8 GDN `b` and `a` on stock MLX because the packed Q8 suffix failed repeated recurrent-state accuracy checks.
- Add Q6 dense-MLP eligibility through the existing Q6 Metal and NAX kernels.
- Add the quality-safe Q6 GDN path with one packed GPU suffix for the QKV remainder, `b`, `a`, and alignment padding.

## Design and compatibility

- The existing runtime-configured ANE block width is unchanged.
- Existing MLP and GDN fractions remain configurable.
- The existing automatic tuner is unchanged.
- The GPU continues to process the remainder.
- Q4/Q4e behavior and public native symbols remain available.
- Existing Q5 GDN behavior remains available.
- The fused Q8 MLP path remains unchanged.
- Q6 GDN supports affine group sizes 64 and 128.
- Procedure identity includes bit width and group size.
- Decode, scheduling, model loading, dashboard controls, MTP, and defaults are unchanged.
- Routed MoE expert tensors remain outside this change.

## Benchmarks

Measurements use local-SSD Qwen3.8 27B checkpoints, a 2048-token prompt, serial execution, and the existing runtime controls. Values below are representative validated rows, rounded to one decimal. The Q4 route is unchanged and was not remeasured by this PR.

| Qwen3.8 27B model | GPU PP | Dual ANE MLP PP | MLP gain | Dual ANE MLP + GDN PP | Total gain | GDN gain over MLP |
|---|---:|---:|---:|---:|---:|---:|
| q6 oQ6e | 446.0 | 528.8 | +18.6% | **560.8** | **+25.7%** | +6.0% |
| q8 oQ8e | 432.5 | 532.1 | +23.0% | **557.4** | **+28.9%** | +4.8% |

[...]

PR #2853, opened 2026-08-19
feat(qwen35) Qwen ANE prefill: FP16 CPU AMX work sharing and performance-aware scheduling

onthehub97:
# Qwen ANE prefill: FP16 CPU work sharing for MLP and GDN, performance-aware scheduling, and adaptive tuning

This PR is based on the branch at https://github.com/jundot/omlx/pull/2826 as these contain essential fixes. Those should be merged first, or this PR can be merged in place of that.

## Summary

This change extends the Qwen3.5/3.6/3.8 ANE prefill path so selected MLP and GDN projection work can execute concurrently across the CPU, GPU, and both ANE instances.

The primary path assigns an independently configurable portion of the MLP gate/up projections to FP16 CPU matrix multiplication. Two additional settings can split the MLP down projection and the residual GDN qkv projection between CPU and GPU. The existing quantized GPU suffix, dual-ANE procedure-bank path, and optimized fused q4 route remain intact.

On the Qwen3.8-27B AWQ 4.85 bpw test model, 2,048-token prompt-processing throughput increased from **460.1 to 494.7 tokens/s**, a **7.51% end-to-end improvement**, when the gate/up and down-projection CPU stages were enabled.

The ANE split tuner has also been replaced with a hardware-local, profile-guided search across five independently useful workload controls. The final in-app tuning run selected **45% MLP on ANE, 45% GDN on ANE, 14% gate/up on CPU, 20% down projection on CPU, and 13% GDN qkv on CPU**. It measured **517.9 prompt tokens/s**, a **45.8% improvement over GPU-only execution**.

## Changes

- Add optional FP16 CPU participation to the dual-ANE q4 gate/up path.
- Run the CPU, quantized GPU suffix, and both ANE instances concurrently, synchronizing only before the fused output merge.
- Add an optional CPU/GPU split for the MLP down projection.
- Add an optional three-way ANE/CPU/GPU split for the residual GDN qkv projection.
- Add a native non-SwiGLU merge that restores ANE-planar, CPU-row-major, and GPU-row-major GDN outputs directly into projection order.
- Add a guarded performance-aware CPU scheduler mode using Apple's shared-resource workqueue policy.
- Default CPU scheduling to eight workers when CPU sharing is enabled.
- Detect scheduler-policy support at runtime and fall back to ordinary Accelerate scheduling when unavailable or rejected.
- Scope scheduler state to each worker invocation and always clear it before returning the pooled thread.
- Preserve compatibility with older native extensions that do not expose the scheduler capability symbol.
- Preserve the optimized fused q4 route and the existing q4/q5/q6/q8 quantized GPU suffixes.
- Preserve NAX selection and fallback behavior for the GPU suffix on supported hardware.
- Expose CPU gate/up share, down-projection share, GDN share, worker count, and performance-aware scheduling through:
  - per-model settings;
  - model profiles;
  - the administration API;
  - the web UI;
  - the SwiftUI application;
  - the standalone benchmark utility.
- Update the benchmark profiler to report CPU matrix and completion timing alongside GPU and ANE timing.
- Add a utility that creates a separate FP16-compatible clone of an MLX checkpoint without modifying the source model.
- Replace the full-model split sweep with an adaptive tuner covering:
  - MLP gate/up work on ANE;
  - MLP gate/up work on CPU;
  - MLP down-projection work on CPU;
  - GDN input-projection work on ANE;
  - residual GDN qkv work on CPU.
- Calibrate candidate widths on one representative real MLP and GDN layer, packaged into a small temporary ANE procedure bank.
- Eagerly compile the full model only for the predicted winner, then use native full-model branch timings for one bounded profile-guided refinement and final verification.
- Use NAX-aware ANE search ranges when NAX is available while retaining the classic Metal ranges elsewhere.
- Preserve partial results and the exact termination reason when a tuning run ends early, while withholding an optimal recommendation until a complete candidate has been verified.

## Configuration

The feature remains opt-in as part of the Qwen ANE prefill extension.

| Setting | Default | Purpose |
|---|---:|---|
| CPU sharing | Disabled | Enables CPU participation for eligible Qwen MLP and GDN projections. |
| Gate/up on CPU | 13.5% | Fraction of each gate and up projection assigned to FP16 CPU matrix multiplication. |
| Down projection on CPU | Disabled | Optional second-stage CPU/GPU split. A 20% share was the best isolated and end-to-end test point on the reference machine. |
| GDN on CPU | Disabled | Optional CPU share of residual qkv channels after the ANE prefix. The final reference tuning run selected 13%. |
| CPU workers | 8 | Number of independent single-threaded BNNS shards. |
| Performance-aware scheduling | Enabled | Applies Apple's shared-resource scheduler hint when supported. |

The existing ANE MLP and GDN fractions remain independent. CPU fractions are taken from the quantized GPU suffix; they do not increase the ANE slice. ANE and CPU GDN fractions must total less than 100% so a valid quantized GPU suffix remains.

### CPU weights are prepared eagerly and kept separately

The source checkpoint is never modified in place. CPU-assigned rows are dequantized once to FP16 during model setup and retained for the lifetime of the loaded model. The remaining rows stay in their original quantized representation for the GPU suffix.

This avoids per-prompt dequantization overhead and lets BNNS consume row-contiguous FP16 weights directly. It intentionally trades memory for prompt-processing throughput.

The model must be loaded from a compatible clone whose floating quantization metadata and activations use FP16. The clone utility processes one safetensors shard at a time and leaves the original model untouched.

### Gate/up, down projection, and GDN use separate work-share controls

The three CPU stages have materially different matrix shapes, dependency windows, and optimal splits:

- Gate/up runs alongside both ANEs and the GPU suffix. The final in-app result selected 14% CPU.
- Down projection runs after the fused activation and overlaps only CPU and GPU work. Its best measured point was 20%, but it is disabled by default because it caused a substantial peak-memory increase on the 27B model.
- GDN CPU sharing takes only residual qkv rows after the ANE prefix. It overlaps both ANEs and the remaining quantized GPU suffix; the final in-app result selected 13% CPU with 45% on ANE.

Keeping these controls separate avoids assuming that one fraction is appropriate for all projection shapes and allows the tuner to balance each parallel region independently.

### CPU GDN offload preserves projection layout

The ANE prefix contains the complete z projection followed by a qkv prefix. CPU GDN rows are taken only from the remaining qkv range, with the quantized GPU suffix retaining the final qkv rows. This keeps mixed q4/q5 layouts valid and avoids moving z between quantization formats.

The native three-way primitive launches the two ANE procedures, FP16 CPU matrix multiplication, and quantized GPU suffix concurrently. A non-SwiGLU Metal merge then restores the ANE-planar and CPU/GPU row-major outputs directly into `z + qkv` order. GDN recurrence, b/a projections, and output projection remain on GPU.

### The CPU path uses row sharding rather than BNNS's automatic worker count

Requesting a larger `BNNSFilterParameters.n_threads` did not reliably create the desired parallelism. The implementation therefore divides the output rows into independent 64-row-aligned shards and invokes single-threaded BNNS matrix multiplication for each shard.

Each shard writes to a disjoint output region, so no reduction or cross-worker synchronization is needed. Output ordering remains identical to the original projection.

### Eight workers are the selected default

Worker exploration covered automatic BNNS scheduling, manual sharding, QoS-biased placement, explicit cluster landing, and Apple's private dispatch workqueue controls.

The reference machine contains 20 performance cores and 8 efficiency cores across six CPU clusters. Public Mach affinity policy was rejected by the current kernel. A brute-force experiment could initially land exactly two workers on each cluster, but macOS migrated them as soon as the guard threads exited. Retaining placement would require wasting the remaining cores with busy-spin guards and was rejected as unsafe and counterproductive.

Mixed performance/efficiency worker groups were also poor for equal synchronized shards: a 12-worker mixed test took 35.160 ms versus 15.902 ms for performance-biased workers.

The useful discovery was the shared-resource round-robin policy underlying private dispatch apply attributes. On Apple's current Edge scheduler this is a scheduling hint, not hard affinity: the kernel dynamically spreads shared-resource-heavy workers and may migrate them toward performance clusters. The worker index does not provide durable core binding.

Isolated FP16 matrix results were:

| Scheduling mode | Workers | Median | Effective CPU use |
|---|---:|---:|---:|
| BNNS automatic | Automatic | 16.171 ms | 3.91 cores |
| Manual aligned rows | 12 | 15.407 ms | 8.84 cores |
| Shared-resource | 8 | **14.560 ms** | **6.67 cores** |
| Shared-resource | 12 | 15.487 ms | 8.65 cores |
| Shared-resource | 16 | 14.423 ms | 13.00 cores |

Eight workers delivered almost all of the 16-worker throughput with roughly half the CPU consumption. The production hybrid primitive subsequently measured 5.715 ms at eight workers versus 6.997 ms with automatic scheduling, a **22.4% isolated throughput increase** with bit-identical output.

### The private scheduler call is capability guarded

The implementation dynamically resolves the private `__bsdthread_ctl` entry point and probes whether the shared-resource policy is accepted. Unsupported systems use the ordinary Accelerate path.

Scheduler state is applied by an RAII scope inside each dispatch worker and cleared on every normal or exceptional exit. A pooled dispatch thread therefore cannot retain the private policy after its matrix shard completes.

Failure to apply the scheduling hint does not make inference fail and does not disable the ANE path.

### The tuner separates cheap calibration from full-model validation

Brute-forcing five workload dimensions with full-model recompilation is not practical. The tuner instead builds a small temporary procedure bank containing several candidate widths from one real MLP and GDN layer in the loaded checkpoint. It measures the same native ANE, CPU, and GPU paths used in production and uses those rates to construct one predicted full-model candidate.

Only that predicted candidate is eagerly compiled across all eligible layers. Native profiling from the application-level prompt run then rebalances the ANE, CPU, and GPU completion rates once before a final end-to-end verification. Prompt throughput, rather than isolated projection time, remains the recommendation criterion.

The five calibrated controls remain independent because they describe different parallel regions:

- MLP gate/up ANE work overlaps both the CPU FP16 slice and quantized GPU suffix.
- MLP gate/up CPU work is taken from the GPU suffix rather than the ANE slice.
- MLP down-projection CPU work overlaps only the corresponding GPU projection.
- GDN ANE work overlaps its quantized GPU suffix and has a separate optimum from MLP.
- GDN CPU work is taken from that suffix and participates in the same three-way balance without changing the ANE program width.

CPU dimensions are skipped automatically for checkpoints or native extensions that cannot use FP16 CPU sharing. Tuning does not persist model settings automatically; the verified recommendation is returned for explicit application by the user.

The result view always retains its complete planned matrix. If the run is interrupted, including by the memory guard or an allocation failure, completed throughput and speedup cells remain visible, uncompleted cells remain blank, and the exact termination reason and completed/total count are reported. An incomplete run is never presented as an optimum.

### Down sharing is disabled by default

The down-projection split produced a real throughput improvement but increased observed peak process RSS from approximately 31.1 GiB for gate/up-only sharing to 49.9 GiB with a 20% down share during the benchmark run.

The setting is therefore exposed for hardware-specific tuning but is not silently enabled for existing profiles.

### Adaptive tuner result

The final result below was produced by the in-app tuner against the real Qwen3.8-27B AWQ 4.85 bpw FP16-compatible clone. It is the authoritative end-to-end result because it includes the application scheduler and the complete five-control workload balance.

| MLP on ANE | Gate/up on CPU | Down on CPU | GDN on ANE | GDN qkv on CPU | PP throughput | Change versus GPU |
|---:|---:|---:|---:|---:|---:|---:|
| **45%** | **14%** | **20%** | **45%** | **13%** | **517.9 prompt tok/s** | **+45.8%** |

This result also confirms that CPU GDN offload can improve the complete application workload even though an earlier standalone sweep, with a different fixed split and process state, did not beat its local 50% ANE / 0% CPU GDN point. The tuner must therefore judge CPU GDN from the final end-to-end candidate rather than treating isolated projection timing as the recommendation.

An earlier four-control profile, before CPU GDN was added, showed why the profile-guided correction is useful:

| Parallel region | ANE | CPU | GPU |
|---|---:|---:|---:|
| MLP gate/up | 17.99 ms | 17.97 ms | 17.94 ms |
| GDN input projections | 10.60 ms | — | 10.57 ms |

These timings are full-model completion measurements rather than isolated projection estimates. The new fifth dimension extends the same rate-balancing approach to the CPU GDN branch.

### Test configuration

- Model: Qwen3.8-27B AWQ, 4.85 bpw, FP16-compatible clone
- Prompt length: 2,048 tokens
- Benchmark mode: text-model path matching the application
- MLP on ANE: 45%
- GDN on ANE: 45%
- Dual ANE: enabled
- Gate/up CPU share: 14%
- Down-projection CPU share: 20%
- GDN qkv CPU share: 13%
- CPU workers: 8
- Performance-aware scheduling: enabled
- Compiled layers: 64 MLP and 48 GDN
- Resident ANE programs: 2
- Safety guard: 80 GiB process RSS

The following earlier staged benchmark isolates the contribution from the two MLP CPU stages before CPU GDN support was added:

| CPU MLP configuration | PP throughput | Median prompt time | Change versus CPU disabled |
|---|---:|---:|---:|
| CPU disabled | 460.1 tokens/s | 4.451 s | — |
| 13.5% gate/up | 475.3 tokens/s | 4.309 s | **+3.30%** |
| 13.5% gate/up + 20% down | **494.7 tokens/s** | **4.140 s** | **+7.51%** |

Adding down-projection sharing improved throughput by a further **4.08%** over gate/up sharing alone. Combined CPU sharing reduced median prompt time by **6.99%** relative to the CPU-disabled ANE/GPU configuration.

The first sample of the CPU-disabled and down-sharing runs was slower than the subsequent samples. The median uses the two stable passes:

- CPU disabled: 5.562 s, 4.451 s, 4.419 s
- Gate/up CPU: 4.339 s, 4.309 s, 4.303 s
- Gate/up plus down CPU: 5.396 s, 4.140 s, 4.112 s

### Profiler observations

With gate/up CPU sharing enabled:

- CPU matrix time: 16.86 ms per profiled MLP operation
- GPU completion: 18.23 ms per operation
- ANE parallel region: 18.54 ms per operation
- ANE completed last for all 192 profiled MLP operations

The initial gate/up split therefore moved the limiting device from GPU toward ANE and produced a closely balanced parallel region. The additional down split reduced work in the later GPU projection and produced the further end-to-end gain.

## Numerical validation

- Shared-resource and ordinary scheduling produced a maximum absolute output difference of `0` in the native hybrid-kernel comparison.
- The CPU/GPU down-projection dispatcher was compared with MLX's original quantized linear on a representative test layer:
  - cosine similarity: `0.99999994`;
  - RMSE: `0.0000294`;
  - maximum absolute error: `0.000244`.
- The native three-way GDN projection was compared with the unsplit projection using real-model dimensions:
  - cosine similarity: `0.99999750`;
  - RMSE: `0.0000652`;
  - maximum absolute error: `0.000549`.
- A real 4.85 bpw model run with CPU GDN enabled preserved the top token; hidden-state and last-token logit cosine similarity were approximately `0.999989` and `0.999999`.
- CPU and GPU slices preserve the original row order; each worker writes a disjoint output range.

These checks validate the kernel split and merge mechanics. They are not a substitute for a full model-quality or task-accuracy evaluation.

## Build and test validation

- 115 focused Python tests passed in the final CPU GDN validation, covering:
  - ANE prefill dispatch;
  - adaptive tuner calibration, candidate prediction, profile refinement, and incomplete-run behavior;
  - CPU scheduler capability fallback;
  - CPU down-projection preparation and dispatch;
  - CPU GDN partitioning, native routing, and three-way profile refinement;
  - model-setting persistence;
  - model profiles;
  - administration routes;
  - web-template state restoration.
- 9 focused Swift model-settings tests passed, including CPU GDN persistence and combined-fraction validation.
- Native Qwen extension built successfully for Python 3.11 and Python 3.13.
- Classic Metal and NAX metallibs built successfully.
- A complete signed release application build succeeded with custom kernels enabled; the packaged Qwen extension and both metallibs passed the bundled-MLX ABI check.
- The Swift application target built successfully.
- The packaged extension reports both ANE availability and shared-resource scheduler support on the reference system.
- End-to-end runs completed without reaching the 80 GiB guard.

## Known limitations

- The ANE path and shared-resource scheduling both rely on undocumented Apple interfaces whose behavior may change between OS or hardware releases.
- The scheduler policy is a hint, not hard CPU affinity. macOS remains free to migrate workers.
- The fused gate/up CPU path currently targets the dual-ANE q4 route. The generic down-projection splitter can retain q4/q5/q6/q8 GPU suffixes when otherwise eligible.
- CPU GDN supports eligible q4/q5/q6/q8 qkv projections, but it only splits residual qkv rows after the ANE prefix. The z projection remains on ANE, while GDN recurrence, b/a, and output projection remain on GPU.
- CPU sharing requires an FP16-compatible checkpoint clone and increases model load time and resident memory.
- Down-projection sharing has a particularly large observed peak-memory cost on the 27B model.
- Optimal ANE and CPU fractions are model-, prompt-shape-, OS-, and hardware-dependent. The measured values should be treated as starting points rather than universal defaults.
- ANE execution remains fixed-shape. Ineligible prompt tails and non-prefill calls continue through existing paths.
- The current CPU implementation uses deprecated BNNS matrix APIs because they expose the required direct FP16 behavior. Migration to BNNSGraph should be investigated.
- No power-consumption comparison is included.
- Full model-quality and application-level intelligence testing remains outstanding.

## Future work

- Validate the adaptive tuner across additional Apple Silicon generations, quantization formats, model sizes, and prompt shapes.
- Consider a bounded worker-count or scheduler-mode calibration stage without reintroducing a combinatorial full-model search.
- Add explicit steady-state and peak unified-memory accounting to the tuner.
- Validate longer prompts and application scheduler behavior, including chunked prompt processing.
- Benchmark the down split across more quantization formats and Qwen sizes.
- Validate the 13% CPU GDN result across repeated application runs, other Qwen sizes, NAX systems, and additional Apple Silicon generations.
- Evaluate BNNSGraph or a dedicated AMX kernel to reduce CPU submission and workspace overhead.
- Add model-quality regression suites covering perplexity, long-context recall, reasoning, and tool-use workloads.
- Generalize the CPU split planner so other dense architectures can describe eligible projections without Qwen-specific dispatch code.

## Reviewer notes

The most important non-obvious behavior is that the eight-worker mode does **not** pin two threads to each CPU cluster. It marks independent BNNS shards as shared-resource-heavy and lets the Edge scheduler distribute them dynamically. Attempts to force stable placement were slower, unstable under migration, or required unsafe busy-spin guards.

The second important trade-off is memory: eager FP16 rows are what make CPU execution fast enough to overlap useful work, but the 20% down-projection share materially increases peak RSS. This is why gate/up, down, and GDN sharing remain separately configurable and disabled by default.

The third important detail is that CPU GDN does not mean moving the entire GDN block to CPU. Only a contiguous residual qkv output slice is assigned to FP16 CPU matrix multiplication. The two ANE slices, CPU slice, and quantized GPU suffix run concurrently and synchronize at the native output merge; all stateful recurrence remains on GPU.


onthehub97:
In terms of CPU usage - it's actually shockingly low. Because this exclusively uses the CPU's dedicated AMX accelerators, and not standard floating point CPU arithmetic, it only occupies a total of about 3 CPU cores on an M3 Ultra to achieve the performance demonstrated in the benchmark table. Because of this, it doesn't have any thermal impact that the fans can't manage.

I haven't yet benchmarked this on a smaller chip, though it's possible the throughput is constant between base-M chips through to Max and then doubles for Ultra (similar to the ANE), depending on where the AMX processor actually resides on the chip - as it doesn't seem to reside in a fixed set of CPU cores.

onthehub97:
Testing on an oQ8e model shows significant, but lesser, performance gains, but working as expected:
image

beaglemoo:
Ran both tuner modes on this branch (`700886e4`, kernels rebuilt with
`OMLX_WITH_CUSTOM_KERNEL=1`) on the M5 Pro / 64 GB, plus a real-request
validation of the recommendation. Headline: the new tuner is a big step up
mechanically — CPU correctly rejected, GDN now genuinely executing on q4, no
memory aborts — but its recommendation still loses on real requests here, by
more than the old split did.

## Setup changes since my #2826 report

- Raised `iogpu.wired_limit_mb` 51.84 -> 58 GiB, the value the enforcer log
  recommends on this machine. The prefill guard cap went 46.66 -> 52.2 GB, and
  unlike yesterday's runs (both died 9/11 with `PrefillMemoryExceededError`)
  both tuning runs today completed every slot. Different tuner code too, so I
  will not fully attribute that — but the old failure was a straight cap hit,
  and the cap is what moved.
- Same checkpoint: `Qwen3.8-27B-oQ4e-mtp` (q4, MTP depth 8, TurboQuant off),
  `dual_ane: false`, macOS 26.6.2.
- Honesty note: these are afternoon runs on a machine that has been
  benchmarking for hours. Every comparison below is interleaved or
  within-run, but absolute numbers will read low vs a cold morning.

## Tuner, default settings

Recommendation: **MLP 35% + GDN 45% on ANE, CPU 0% everywhere**
(`cpu_enabled: false` came out of calibration on its own — matching your
expectation that CPU does not pay on M5).

```
GPU only                 445.2 tok/s  (444.7, 445.7)
Predicted optimum        487.0 tok/s  (496.6, 477.3)   +9.38%
Profile-refined (GDN 40) 478.1 tok/s  (496.4, 459.7)   +7.38%
```

## Tuner, allow_cpu=false

Recommendation: MLP 35% + GDN 40%, reported +13.58%. But the GPU-only
baseline for this second run sagged to 427.1 (441.0, 413.1) while the ANE
arms held ~478-487 — the same differential-thermal-sag effect I described in
#2826, now inflating the tuner's own percentage between two back-to-back
runs. Suggestion: report absolute tok/s next to the percentage in the UI, and
maybe re-measure the baseline after the winner (baseline drift is the biggest
error term on thermally constrained machines).

## GDN on q4 now executes

Yesterday on #2826 this machine showed `category=gdn operations=0` with 96
expected. On this branch:

```
category=gdn operations=96 expected_operations=96
ane0_duty=0.065  ane1_duty=0.103
input_ready_ms_per_op=18.5-21.3   ane_region_ms_per_op=9.5-10.2
```

So whatever predicate was rejecting q4 GDN before is fixed here. Note the
ratio though: the ANE spends roughly 2x longer waiting for inputs as
computing, and duty stays under ~10%.

## Real requests: the recommendation still loses on M5

Applied the default-run recommendation (MLP 0.35, GDN 0.45, seq 2048,
CPU off), real `/v1/chat/completions`, model unloaded and prompt cache
cleared before every run, OFF/ON interleaved in pairs:

```
~16K   OFF 42.46  ON 54.71
       OFF 47.70  ON 66.40
       OFF 49.84  ON 60.68
       median OFF 47.70s  ON 60.68s   -> ANE -27.2%

~32K   OFF 103.56  ON 113.48
       OFF 104.78  ON 110.63
       median OFF 104.17s  ON 112.06s -> ANE -7.6%
```

No overlap between arms at either context. The GPU-only arm was stable at 32K
this time (103.6 -> 104.8), so no thermal crossover to argue about.

Two things stand out:

1. The tuner's own harness says +9.4% while interleaved real requests say
   -27.2% at 16K. That divergence is bigger than any split question. The
   harness runs fixed-shape prefill with speed-priority pinned; real requests
   add MTP priming, boundary snapshots every 2048 tokens, and SSD cache
   stores. Whatever the cause, tuning on the harness objective currently
   optimizes the wrong thing on this chip.
2. Yesterday's MLP-only 25% split lost -13.7% at 16K; today's 35%+GDN45%
   loses -27.2%. GDN executing made real-request TTFT worse, which is
   consistent with the input-starvation profile above.

One possibly relevant scheduler fact I dug out today while chasing chunk
sizes: on GDN models with the prefix cache enabled, delivered prefill chunks
are pinned to exactly 2048 by the ArraysCache block boundary
(`_ARRAYS_CACHE_BLOCK_SIZE = 2048`, scheduler cuts every chunk at the next
block edge). So `sequence_length: 2048` is the only value that matches
delivered chunks — 4096 shapes can never be fed full chunks in normal serving
regardless of the prefill floor or memory headroom. That also means #2751's
4096 prefill floor for >=64 GB machines is currently dead code whenever
caching is on; I patched the block size to 4096 to check, verified 4096-token
chunks delivered end to end, and TTFT got ~5% worse at 16K and ~9% worse at
4K on this machine (three interleaved pairs each) — so on M5 the 2048 block
is accidentally the right value. Full mechanism and data now filed as #2880.

Environment: M5 Pro, 64 GB, macOS 26.6.2, branch @ `700886e4`,
`iogpu.wired_limit_mb=59392`, q4 oQ4e-mtp checkpoint, MTP depth 8.

onthehub97:
That's really helpful. There's a disconnect between reported performance gain in the tuner, and measured performance gain, on non-M5 chips also. I'm investigating why that might be.
My main suspicion at the moment is that the tuner is using a 4097-sized prompt instead of a 4096 sized prompt. This means that it runs 2 full ANE/CPU blocks and then has no GPU tail, whereas the in-app benchmarking tool supplies exactly 4096 which leaves a GPU-only tail of 2047. On non-M5 chips, this still results in a significant, albeit lesser, performance boost than reported by the tuner.

I don't suppose I could get a copy of your logs?

beaglemoo:
@onthehub97 Logs here (sanitized only by scrubbing the home directory from
paths): https://gist.github.com/beaglemoo/aae222168f44f01f43758b83bad9e4e9

Contents: the full serve log for the session (server start on `700886e4`,
both tuner runs, then the interleaved real-request A/B), plus the raw
`/results` payloads for both tuning runs.

Your 4097 suspicion is directly visible in this log: every
`[benchmark-ane-profile]` line from both tuner runs says `pp=4097` — 36 of
36, no `pp=4096` anywhere. So on this machine too, the tuner measured two
full ANE-shaped 2048 blocks with no GPU tail, while my real 16K/32K requests
(15,737 and 31,381 tokens) always end in a non-ANE-shaped tail chunk on the
GPU path. That composition difference lines up with the sign flip between
the tuner's +9.4% and the measured -27.2% better than anything else I have.
Happy to re-run anything once the prompt size is fixed.

beaglemoo:
@onthehub97 Confirmed your 4097 hypothesis with a patched A/B on the M5 Pro,
and the one-line fix removes about half of the tuner's phantom gain here.

The culprit (`omlx/admin/ane_tuning.py:410` on this branch — and note the
same line ships in `main` at :175, introduced with the original tuner in
#2814, so every v0.6.2 tuner number ever reported carries it):

```diff
-    measure_length = run.request.sequence_length * 2 + 1
+    measure_length = run.request.sequence_length * 2
```

With `* 2 + 1`, prefill (`tokens[:-1]`) is exactly two full ANE-shaped
blocks and no GPU tail; with `* 2` the measured prefill becomes
`2048 (ane_shape=True) + 2047 (ane_shape=False)` — the same composition as
the in-app bench and real traffic.

Two back-to-back tuner runs (defaults, unpatched first):

| | GPU only | recommended (MLP .35 + GDN .45) | claimed |
|---|---:|---:|---:|
| unpatched (`pp=4097`) | 453.8 tok/s | 501.9 | **+10.6%** |
| patched (`pp=4096`) | 426.2 | 448.0 | **+5.11%** |

Verified in the log: all `pp=4096` after the patch, chunk pattern
`chunk_tokens=2048 ane_shape=True` then `chunk_tokens=2047 ane_shape=False`.
Ordering note: the patched run went second on a warming machine and baseline
sag inflates the percentage (GPU-only 453.8 -> 426.2), so +5.11% is if
anything still generous.

The remaining gap to my interleaved real-request numbers (-27.2% at 16K)
would be the other regime differences — kv_len, MTP priming, boundary
snapshots — but this closes the part you identified, and it halves the
overstatement on non-M5 chips too if the mechanism is the same.

I left `warmup_length = sequence_length + 1` alone — one full ANE block
seems right for warmup. Happy to send this as a standalone PR against main
(where it also fixes the v0.6.2 tuner) or leave it for you to fold into this
branch, whichever is less merge pain.

beaglemoo:
Dug into *why* real requests lose while the harness wins, at the code level.
I think the per-op host synchronization in `qwen35_ane.mm` is the root cause
on M5, and I have a dose-response experiment supporting it.

## Correction first

I called the profile "input-starved" earlier — that was my misreading of a
label. `input_ready_ms_per_op` in the `[benchmark-ane-profile]` line is fed
from `pack_ns`, and `pack_ns` does not measure the pack kernel. It measures
this sequence (`qwen35_ane.mm`, hybrid gate/up dispatch):

1. the pack kernel is encoded onto **MLX's shared stream command buffer**,
2. `encoder.commit()` — which flushes *everything MLX has queued so far*
   (attention, norms, the lot), then
3. `producer_buffer->waitUntilCompleted()` — a host-blocking wait on all of
   it.

So the 18-21 ms/op I reported is mostly a **GPU pipeline drain per
operation**, not layout conversion. And after the qmm is committed there is a
second host block, `model_->wait(ticket)`, before the merge is encoded. Two
CPU-blocking syncs per accelerated projection, per chunk. The code comments
explain both as driver workarounds ("shared-event signal embedded midway
through one command buffer is not made visible to ANE early enough"; a Metal
wait "can stall the already-queued qmm behind ANE" on M3 Ultra), and note
that they assumed "packing is a short prerequisite" — on a busy M5 stream it
is not.

## Dose-response: the loss scales with op count, not with ANE work

If per-op sync dominates, ANE on fewer layers should recover most of the
loss. Real ~16K requests, MLP-only 0.35, seq 2048, `max_layers` swept,
OFF/ON interleaved per level, cache cleared and model reloaded per run:

| ANE MLP layers | OFF | ON | delta |
|---:|---:|---:|---:|
| 64 | 42.22s | 49.64s | **-17.6%** |
| 16 | 43.66s | 44.74s | -2.5% |
| 8  | 44.09s | 46.00s | -4.3% |

64 -> 16 layers removes three quarters of the ops and three quarters of the
regression. 16 vs 8 inverts inside the ~2% noise floor — the mitigation
saturates at break-even, never crosses positive. On this chip every marginal
ANE-offloaded layer is net negative: its ~10 ms of ANE compute buys back less
than its two host syncs cost.

The arithmetic also closes end to end at full layers: ~18.5 ms x 64 MLP ops
(+ ~20 ms x 48 GDN ops when enabled) x 8 chunks is on the order of the entire
13 s regression I measured at 16K.

It also unifies the context scaling everyone has observed: the sync cost per
chunk is constant while GPU compute per chunk grows with kv_len, so the
*relative* loss must shrink with context — exactly my -27.2% at 16K vs -7.6%
at 32K, and consistent with the chunk-count scaling @onthehub97 reported on
M3 in #2781.

## Implication

No fraction/split setting can fix this on M5 — the tuner is searching a space
where every point pays 2 x ops x sync. The fix that would actually move real
requests is structural: get the ANE launch off the host wait (e.g. commit the
pack in its own small buffer and launch the ANE from its completion handler
instead of `waitUntilCompleted`, and gate the merge on an event rather than
`model_->wait`), so the CPU keeps encoding and MLX's pipeline stays deep. I
realise both host waits exist as driver workarounds, so this may be harder
than it sounds — but as a data point, the M5 loss is not in the ANE at all;
it is in the synchronization around it.

Happy to run any patched build of this branch against the same interleaved
protocol.

beaglemoo:
Logs for the two follow-ups above are now in the same gist (https://gist.github.com/beaglemoo/aae222168f44f01f43758b83bad9e4e9): the measure_length A/B serve logs and raw results (unpatched pp=4097 vs patched pp=4096 — the patched log shows the 2048 ane_shape=True + 2047 ane_shape=False chunk pattern), and the max_layers dose-response serve log with the per-arm 'Eagerly compiled ... on N MLPs' lines. README in the gist maps files to runs.

onthehub97:
Thanks for that experiment - it does confirm part of what I suspected, as for why throughput gains aren't as high as reported, and I have a cheap fix for that: a simple checkbox that adds an extra token to the prompt size used in benchmarks. I'll keep those benchmark results offline so as not to pull a Volkswagen, but it doesn't fully explain why the net throughput shift is in the wrong direction on M5.

image



Digging into it, I may have underestimated the impact forcing the scheduler to submit 2048-sized blocks would have on NAX performance. It turns out, NAX really likes the wider blocks. As fortune would have it, so seemingly does everything else, because it's added a slight performance bump to M3 Ultra also.

So I've restored the 4096-width blocks,  and instead the ANE kernel will internally split them, and I've added a profitability guard which will prevent ANE from deploying when it's unlikely to be of benefit (this should only occur in some conditions on M5+). These changes combined should prevent the ANE setting from ever becoming a net negative on M5 _in theory_. Awaiting your benchmarks to confirm :)

jundot:
@onthehub97 Sorry for being slow to weigh in here while you kept shipping, this thread moved fast. Catching up on the whole ANE situation, I want your take on the overlapping q6/q8 work before I act: I'm thinking of merging #2833 for that route. What do you think?

The short version of why I lean that way: #2833 adds bit aware entry points to the existing fused SwiGLU kernel, so q6/q8 stay on the fused path instead of dropping to the unfused qmm with python side gate/up reassembly, and that alone also gets the MLP/GDN profile attribution right without any profile_category plumbing. Every new path is has_symbol guarded, so an older extension declines cleanly instead of compiling for minutes and then failing at first dispatch. It keeps the OMLX_QWEN35_Q4_LM_LINEAR kill switch working alongside the same GDN first refusal fix you made, and the q4 route stays byte identical through the thin bits=4 delegates. Your Q5+ commit here already moves the CPU variant to the same fused bit aware design, so I read this as convergence rather than a detour for you.

If that sounds right to you, I'd merge #2833 right away and handle its remaining small fixes myself as follow ups, and then ask you to rebase this branch (and #2826) on top of it. While you are at it, could you also split this PR up? It has grown to 10 commits and +3200 lines spanning the tuner rework, CPU sharing, the scheduler mode, and Q5+ support, and I can't review or revert that as one unit. The tuner v2 and whatever made q4 GDN actually execute on @beaglemoo's M5 are things I'd like to land quickly, so pulling those out first would help a lot. Happy to help with the conflicts if the rebase gets painful anywhere.

jundot:
To be concrete about what merging #2833 would actually replace, it is just these three:

| Work item | Where it lives | Covered by #2833 |
|---|---|---|
| q6/q8 ANE eligibility + native path | #2826 commit 2 | Yes, with the fused design |
| GDN first refusal dispatch fix | #2826 commit 3 | Yes, including the OMLX_QWEN35_Q4_LM_LINEAR kill switch |
| q8 b/a accuracy workaround | #2826 commit 3 | Yes (hardcoded there, I'll rework it as a follow up) |

Everything else stays yours and is still wanted: the tuner early exit work, tuner v2, the CPU/AMX sharing, Q5+ support, the q4 GDN execution fix, all of it would just land through the rebased and split PRs. The profile_category plumbing is the one piece that becomes unnecessary, since the fused path attributes MLP work correctly on its own.

beaglemoo:
Benchmarks on `5d120901` as requested. Headline: **the tuner now gets M5
right** — it recommends ANE disabled — but the "never net negative when
manually enabled" goal is not there yet, and the reason is that your restored
4096-wide blocks never actually reach the model on a default config.

## Tuner (defaults)

```
GPU only                 446.5 tok/s  (449.1, 443.8)   baseline
MLP .35 + GDN .45        431.6        (444.5, 418.6)   -3.34%
MLP .35 + GDN .40        428.4        (437.3, 419.4)   -4.05%
recommendation: enabled=false, +0.0%
```

First time the tuner's verdict matches interleaved real requests on this
machine. Note the ANE arm also dropped in absolute terms vs `700886e4`
(501.9 -> 431.6 at the same split, with the GPU baseline flat 445-453 across
all four runs today) — worth a look at whether the internal tiling path adds
overhead even for exact-shape 2048 chunks.

Still `pp=4097` on all 18 profile lines in this build, so the harness tail
fix was not in effect for these numbers.

## The 4096-wide blocks do not arrive

Every tuner measurement logs:

```
chunk_tokens=2048 ... requested_step=4096
```

The scheduler requests 4096 (the >= 64 GB qwen3_5 floor), but the ArraysCache
block boundary clamps every delivered chunk to 2048 whenever the prefix
cache is enabled — that is #2880, and it bites here: with 2048-token chunks
and `sequence_length` 2048, `_tiled_input_plan` sees one full block and no
tail, coverage 1.0, so the profitability guard always passes and the
internal splitting never has anything to split. The wide-block benefit you
measured is real but unreachable on a default (cache-on) config, on any
machine — if the M3 Ultra bump was measured with caching on I would
double-check its chunk trace too.

## Real requests, ANE manually enabled (MLP .35 + GDN .45, seq 2048)

Interleaved pairs, ~16K, cache cleared + model reloaded per run:

```
pair 1  OFF 40.30  ON 48.02
pair 2  OFF 43.75  ON 51.75
pair 3  OFF 46.52  ON 53.31
median  OFF 43.75s  ON 51.75s  ->  -18.3%
```

No overlap between arms. So the guard as it stands cannot catch the M5 case:
the loss here is not low coverage, it is the per-op host synchronization I
described above — two CPU-blocking waits per accelerated projection — which
is invariant to block width and coverage. On a 2048-chunk machine the guard
never fires and the full regression remains for anyone who flips the toggle.

## Suggestion

For the guard to deliver "never net negative on M5", it probably needs a
term for dispatch overhead per op (ops x measured sync cost vs projected ANE
saving), not just tail coverage — or, simpler, gate on a quick measured
probe the way the fa256 patch gates on NAX. And for the wide-block win to
exist outside benchmarks at all, the #2880 block-boundary interaction has to
be resolved first; on this machine I measured GPU-only 4096-token chunks
*slower* than 2048 on real requests (-4.8% at 16K, data in #2880), so the
wide-block question on M5 deserves its own real-request A/B once chunks can
actually be delivered wide.

Same protocol as before, logs can go in the gist on request. Heat-soaked
afternoon session; all comparisons interleaved.

onthehub97:
@jundot I'm happy to rebase onto main after #2833 is merged and split out the changes into 2-3 individual PRs. Note: I've done no testing of 2833 so I just assume you've checked it out and everything works.

jundot:
#2833 is merged (12841f2d) and the follow up landed on main directly (ac343873): q8 b/a projections now route through OMLX_QWEN35_Q8_LINEAR_MIN_TOKENS instead of a hardcoded bits check. @onthehub97 main is ready for your rebase whenever you are.

Since nobody had exercised the merged path on hardware yet, I verified it with a layer level parity harness on an M3 Ultra (512 GB, macOS 26.5): synthetic quantized layers, a rebuilt native extension, real ANE program compilation and dispatch through the merged _backend/_gdn_backend, with stock MLX as ground truth. 2048 token input, fraction 0.5.

| Path | Configs | Result |
|---|---|---|
| MLP fused SwiGLU | q4/q6/q8 x gs 64/128, single and dual ANE | All compiled and executed, cosine >= 0.99997, max abs ~0.001 against ref max ~0.13 |
| GDN hybrid | q4 and q8 (fallback b/a), q6 at gs 64 and 128 (packed b/a) | All compiled and executed, cosine >= 0.99999 on each of qkv/z/b/a |
| q8 GDN b/a | threshold route from the follow up | Exactly zero diff vs stock, confirming the short sequence stock route |

The q6 packed suffix slices b and a back out of the combined qmm correctly, and the q4 control run shows the same parity as the new bit widths. This is layer level parity, not an end to end model quality run, but it covers every new dispatch combination the merge added.

onthehub97:
@beaglemoo I've removed the profitability guard and updated the instrumentation so that logging makes it much clearer what's happening internally. Can you re-test the tuner to double check that removing the profitability guard hasn't caused a regression in the suggestion?

I need to get an M5 Max...

beaglemoo:
Re-ran the tuner on `d867f064` (defaults, kernels rebuilt). No regression —
the recommendation is still ANE disabled on the M5 Pro:

```
GPU only            445.8 tok/s  (447.9, 443.7)   baseline
MLP .35 + GDN .45   428.7        (438.8, 418.5)   -3.85%
MLP .35 + GDN .40   421.8        (428.6, 414.9)   -5.39%
recommendation: enabled=false, +0.0%
```

That's within noise of the guarded run an hour earlier (-3.34% / -4.05%,
baseline 446.5), so the guard's removal changed nothing here — the negative
verdict is coming from the measurements themselves, which is what you want.

Two same-as-before notes for completeness: still `pp=4097` on all 18 profile
lines in this build, and delivered chunks are still `chunk_tokens=2048`
against `requested_step=4096` (the #2880 boundary clamp), so this run says
nothing new about wide blocks.

Serve log (with your updated instrumentation) and raw tuner results added to
the gist as round 4:
https://gist.github.com/beaglemoo/aae222168f44f01f43758b83bad9e4e9

Now you've got me regretting not getting the 128 GB M5 Max! Happy to keep
running this Pro as your remote M5 lab in the meantime :)

onthehub97:
Yep - I wasn't expecting to fix the ArraysCache issue, just make it much more explicit in the logs what's happening.

I'm not convinced there's no wins to be found with the M5's ANE - it's still 42 TOPS of completely unused power. I'm just not convinced that this patchwork approach (which only utilises under half of prior generations NPUs) of work sharing is the way to do it, when the synchronization overhead is so significant already. On prior generation chips, the GPU is so unsuited for the task it's been given that it's worth it, and you can take over 60% of the work off the GPU and still end up at 100% GPU utilization and blocking on the GPU work, but the M5 that doesn't seem even close to the case.

I'm experimenting with other approaches to push utilisation to 100% and would definitely be interested in revisiting the M5 when I have any promising leads if you'd be interested in helping me test, as that may make the synchronization dance worth it - 10% clearly isn't worth it, at all.

beaglemoo:
Standing offer stands — ping me whenever you have a build worth trying and I'll run the interleaved protocol on it.

beaglemoo:
@onthehub97 You were right not to be convinced there were no wins on the
M5 — and it turns out the wins were there all along. I owe this thread a
correction, and I have validation of your v2 work plus one small PR to go
with it.

## The correction

Every negative real-request number I posted (-12% at 16K, -33% at seq 4096,
and the -17.6% max_layers "dose-response") came from a protocol that
unloaded the model and timed a request that triggered the reload. The
ANE-off arm just re-paid the weight load; the ANE-on arm also paid eager
program compilation plus every procedure's first-evaluation cost — and
compile time scales with sequence_length and procedure count, which is
exactly the "dose-response" I attributed to per-op synchronization. So my
real-request numbers were biased pessimistic in the same period the old
tuner was biased optimistic. Sorry for the noise — three days of my
"ANE loses on real requests" data was measuring warmup, not inference.

## Corrected numbers, on current main (4c00bbe5)

Protocol: fresh server per run, model preloaded via the admin API, then a
warmup request (same length, different content, no prefix reuse) before the
timed one; SSD cache cleared per run; arms interleaved. M5 Pro 64 GB,
Qwen3.8-27B oQ4e, sequence_length 2048.

TTFT at ~16K (15,737 tok), pure upstream main, two interleaved pairs:

| pair | ANE off | ANE on | delta |
|---|---|---|---|
| 1 | 42.4 s | 38.2 s | **+9.9%** |
| 2 (order reversed) | 49.8 s | 38.8 s | **+22.1%** |

Earlier the same day, a 3-round matrix on a build a few commits back gave
the same picture (+8.7 / +12.7 / +22.3% at 16K; every ANE-on run beat every
ANE-off run at 32K as well). Decode is unaffected (48.3 vs 46.2 tok/s,
ANE on marginally ahead).

**And tuner v2 now agrees with reality.** On the same machine where the
pre-v2 tuner recommended `enabled=false` (-3.85%), v2 recommends:

```
enabled=true, mlp_fraction=0.40, gdn_fraction=0.25, +17.0% (436.8 tok/s)
```

A real-request A/B of your recommended split vs my hand-tuned 0.35/0.45 is
within noise, so the recommendation is good as-is. (Note gdn_fraction 0.25
compiles 0 GDN procedures on this model — below the viability floor — so
the v2 rec is effectively MLP-only here.) I believe your warmup prompt and
the measure_length fix are what closed the harness-vs-reality gap: the old
harness's ANE arms were paying uncompiled-procedure costs exactly like my
old protocol was.

Two more data points from the session:

1. **The M3-era host syncs are not an M5 cost.** I built an env-gated
   variant that removes both per-op host blocks in `qwen35_ane.mm` (skip
   the pack `waitUntilCompleted`; use the already-implemented-but-unused
   encoded `end()` wait instead of `model_->wait`). It measured neutral vs
   stock at 16K — the GPU stays fed through the host blocks on this chip.
   So for M5 purposes the current synchronization is fine; no need to
   spend complexity there.
2. **Thermal stability is a sleeper benefit.** Across a heat-soaked
   session, ANE-on held 37.8-39.5 s at 16K while GPU-only degraded from
   41.7 to 56.3 s. Shifting work off the GPU lowers its power draw, so the
   ANE advantage grows under sustained load.

## The one gap left, and a small PR for it

Eager compilation moved program creation to load time, but each procedure
still pays its first-evaluation cost inside the first user request:
+5-8 s on first-request TTFT at 16K here (46.1/48.3 s vs 38.2/38.8 s
steady-state). I've opened #2898, which warms every banked procedure at
load — one throwaway evaluation per procedure behind a host-signaled
ticket. Costs 2.9 s at load on this machine and brings the first request
down to thermal noise (41.5/41.1 s, steady state unchanged).

Sanitized serve logs, tuner v2 results, and the raw A/B numbers for all of
the above are in the gist as round 5:
https://gist.github.com/beaglemoo/aae222168f44f01f43758b83bad9e4e9

The M5's 42 TOPS were never locked up — they were locked up in my
benchmark. Standing offer still stands for anything else you want run.

PR #2856, opened 2026-08-19
SpecPrefill: enable DeepSeek/GLM MLA draft models (query extractor, RoPE dtype, DSA/CacheList positions)

anicaise-ai:
This PR carries the three fixes that make SpecPrefill usable with DeepSeek/GLM-family (MLA + DSA) models. They have been running in our fork since late July; rebased clean on v0.6.2 (`10c49755`, `99b8292d`, `ff78637e`), 16/16 unit tests green, live-serve proven.

**1. `feat(specprefill): MLA query extractor for DeepSeek/GLM-family drafts`**
The draft-side query extraction assumed classic multi-head attention layout. MLA drafts (compressed KV latent + separate rope key) need their own extractor; without it SpecPrefill can't score chunks for these families at all.

**2. `fix(specprefill): preserve input dtype in manual RoPE helpers`**
`manual_rope` / `manual_rope_with_freqs` silently upcast, so the scored representations diverged from what the model actually computes in fp16/bf16 paths.

**3. `fix(specprefill): keep DSA indexer and CacheList offsets on true positions`**
After chunk dropping, the DSA indexer and CacheList offsets were advanced by *kept* positions instead of *true* positions, corrupting long-context indexing for DSA models.

**Why it matters — measured results** (GLM-5.2 DSA, M3 Ultra; together with the traditional-RoPE pair layout already discussed on #2355, where we posted the numbers on 2026-07-30):

- 16k needle grid: **12/12 retrieved** (9/9 in sparse mode) vs 0/9 before the fixes;
- `keep=0.2`: prefill **~2.85× faster than dense** (443/401/413 tok/s vs 166/171/117), needle found at all 3 depths;
- 64k, `keep=0.2`: 3/3 retrieved, **1.85–2.03×** vs dense anchors.

**Scope / risk**: all three commits touch only `specprefill.py` code paths; models with SpecPrefill disabled (the default) are unaffected. `_OffsetAdjustedRoPE` delegates to native RoPE and needed no change.

Without these, SpecPrefill is structurally unavailable for the DeepSeek/GLM families — and upstreaming them removes the last reason we build a local wheel, so we'd much rather maintain them here. Happy to adjust style/tests to match the repo's conventions.

### DIFF
--- a/omlx/patches/specprefill.py
+++ b/omlx/patches/specprefill.py
@@ -162,6 +162,38 @@ def _nemotron_h_extract_queries(attn, x, cache=None, **kwargs):
     return queries
 
 
+def _mla_extract_queries(attn, x, cache=None, **kwargs):
+    """DeepSeek/GLM MLA: (optionally low-rank) q projection + partial RoPE.
+
+    MLA caches store keys in a split layout, so queries are built in the
+    matching layout. Absorbed variants (``embed_q``, e.g. glm4_moe_lite)
+    cache the compressed latent as keys and the roped positional slice as
+    values; queries are q_nope absorbed into latent space, concatenated
+    with the roped q_pe. Expanded variants (``kv_b_proj``) cache full
+    per-head keys; queries are concat(q_nope, roped q_pe) directly.
+    Queries are pre-scaled so the scorer's generic 1/sqrt(d) recovers the
+    module's true attention scale (including any yarn mscale correction).
+    """
+    B, L, D = x.shape
+    n_heads = getattr(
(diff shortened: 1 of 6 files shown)


anicaise-ai:
Closing this one to keep the queue clean — it overlaps with #2851, which has your review, the follow-up fixes in `ed6ea606`, and the tests you asked for. This branch has also drifted into a conflict with `main` in the meantime.

Happy to reopen or re-submit a rebased version if you'd rather look at the two pieces separately.

PR #2863, opened 2026-08-19
feat(admin): add max_tokens override for external accuracy benchmarks

williamxie1989:
Thinking models behind OpenAI-compatible endpoints spend reasoning tokens against the benchmark's small default per-question budgets (128 for MMLU, 2048 for MMLU-Pro), so answers are truncated mid-reasoning and the run reports empty/truncated responses with an unreliable score. The external client sends no thinking control, and the enable_thinking flag is a local chat-template kwarg that never reaches a remote endpoint.

Add an optional max_tokens_override to the external endpoint config, exposed in the accuracy benchmark UI next to the extra request body. The accuracy adapter applies it as a floor: effective budget is max(benchmark default, override), so a benchmark's own budget is never shrunk (e.g. LiveCodeBench keeps 16384 when the override is 8192). The throughput benchmark is unaffected.

- external_api: max_tokens_override field (1..1_000_000) on ExternalEndpointConfig, surfaced via ExternalAPIClient and applied in ExternalChatAdapter.chat()
- UI: number input in the accuracy external section, persisted in localStorage, validated client-side (positive integer)
- i18n: label, hint, and error strings in all 9 locales
- tests: config validation + adapter floor behavior (raise, no shrink, no-op)


jundot:
Thanks for adding this. The accuracy-only floor is wired through the endpoint config and adapter without affecting throughput runs, and the validation plus no-shrink coverage handles the important edges. LGTM, merging.

PR #2866, opened 2026-08-19
feat(cluster): server-owned incident feed with monotonic-merge polling (fabric-doctor B1)

alytaphoenix:
## ⚠️ Stacks on #2864 → #2849 → #2848 → #2819

Fabric-doctor series. Builds on **#2864** (RDMA fast-fail) → **#2849** (C1) → **#2848** (B3) → **#2819**. Diff includes their commits until they merge; review only the last commit (`feat(cluster): server-owned incident feed…`) — it collapses to just that once the parents land. This is a foundational item: the incident store is what B2 (start-job supersession) and the incident UI (B5/B6) will write into.

## Summary

Every cluster failure funnel now records a **durable incident** instead of only surfacing as a transient HTTP error the browser forgets on refresh — a failure that happened while you weren't looking at the tab, or that resolved itself before you next checked, used to leave no trace.

## What changed

- **New `omlx/cluster/incidents.py`** — `Severity` (info/warn/error), a frozen `Incident` dataclass, and `IncidentStore`: atomic save (temp file + `os.replace` + fsync, the `enrollment.py` pattern), fail-closed on a corrupt file (`registry.py` pattern), a 200-entry FIFO ring, and a strictly monotonic `seq` that survives both eviction and restart (persisted `next_seq`). `record()` / `list(since_seq)` / `latest_seq()` / `dismiss(id)` (marks `dismissed_at`, **never deletes**) / `supersede(job_id, by_id)`.
- **`server.py`** wires `configure_cluster_incidents(base_path)` beside the registry/enrollment stores.
- **`routes.py`** — `_record_cluster_incident()`: redacts the message via `_redact_diagnostic`, derives a `guidance_code` from `explain()` (joins B3's guidance codes), and swallows store errors so incident *logging* can never mask the original failure. `GET /admin/api/cluster/incidents?since=<seq>` (redacts again on the way out) and `POST .../{id}/dismiss`; `/diagnostics` gains `"incidents"`. Three emitter funnels: every `activate_cluster_deployment` except-arm before re-raising; both `_run_staging_job` failure paths; `cluster_peer_health` on a previously-healthy deployment's **first** unhealthy transition (tracked so it never re-fires every poll).
- **`dashboard.js`** — `loadClusterIncidents()` implements the **monotonic-merge poll rule**: fetch since the cursor, merge into a `Map` keyed by `id` (add/update only — the poll never deletes locally), cursor only advances forward, dismissal round-trips through the endpoint before the local copy is marked dismissed. Hooked into the 2s `refreshClusterExperience()` tick. The error banner's dismiss now also dismisses the matching incident server-side.
- **`_cluster.html`** — incident count badge + reverse-chronological feed (severity, source, age, message, superseded flag, per-row Dismiss). Full status-strip layout deferred to B4.

## Tests

`tests/test_cluster_incidents.py` (8, new): monotonic seq, ring eviction, dismiss persists across reload, corrupt file fails closed, cursor semantics, supersede marks-not-deletes, seq continuity across restart. Plus 4 new cases in `test_cluster_routes.py` (activation failure records a matching incident, cursor semantics, dismissed incident stays in diagnostics + 404 on unknown id, peer-health fires once per transition), and a `test_cluster_one_click.py` spec update for the new initialization call.

**Verified via `git stash` A/B comparison**: `test_cluster_routes.py`'s 3 failures are pre-existing and identical before/after this change (49 passed pre-B1 → 53 passed post-B1, same failure names) — zero regressions. No change in `readiness`/`probe`/`link_status`.


jundot:
Thanks, the store design and its tests are genuinely good, so I'm landing this ahead of the rest of the series. I added the Guidance.code lookup it needs to main myself and reparented this branch onto main, hope the force-push is ok given the stack is getting restructured anyway. I'll follow up to move the record calls off the event loop and add a store epoch so an open tab's cursor survives a corrupt-file reset.

PR #2867, opened 2026-08-19
feat(cluster): VPN pre-warning + collision-checked, exclusion-aware fabric addressing (fabric-doctor C4)

alytaphoenix:
## ⚠️ Stacks on #2866 → #2864 → #2849 → #2848 → #2819

Fabric-doctor series. Builds on **#2866** (B1) → **#2864** (RDMA fast-fail) → **#2849** (C1) → **#2848** (B3) → **#2819**. Diff includes their commits until they merge; review only the last commit (`feat(cluster): VPN pre-warning…`) — it collapses to just that once the parents land.

**Note on scope:** this PR also introduces the base collision-checked fabric-subnet selection (`choose_fabric_subnet`, `_FABRIC_SUBNET_CANDIDATES`, `_occupied_networks`, replacing the old hardcoded `10.0.1.1/24` default) — infrastructure written during the original incident response but never shipped as its own PR. C4's VPN-aware `preferred=` parameter is layered directly on top of it, so they ship together here.

## Summary

Detects a hungry full-tunnel VPN **before** the user hits it, warns in plain language, and feeds its readable exclusions into fabric subnet selection — while empirical link probes (C1/B3) remain the only promotion authority; a VPN read never advances the transport ladder, it only warns and enriches candidate ranking.

This systematizes a real fix this exact 2-Mac cluster needed by hand: the peer runs Cloudflare WARP (full-tunnel, org policy "all interfaces"), which silently swallowed an auto-assigned `10.0.1.x` Thunderbolt fabric link while LAN ping/SSH kept working fine. The fix at the time was manually re-addressing into WARP's `172.16.0.0/12` exclusion — this PR makes that automatic.

## What changed

- **New `omlx/cluster/vpn.py`** — `VPNProfile` (`present`/`client`/`full_tunnel`/`utun_interfaces`/`exclusions`), `detect_vpn()`: utun interfaces from the existing interface probe, default//1+/1 routes via `netstat -rn`, client signatures (WARP/Tailscale/GlobalProtect/AnyConnect) via a bounded `ls /Applications` plus `warp-cli settings` for readable exclusions. **Every read is failure-tolerant** — any failure degrades to `client="unknown"` with empty exclusions, never raises. `hostile_networks()` = interface subnets ∪ utun-routed prefixes (catch-all `0/0` routes are full-tunnel *evidence*, not a hostile range — counting them would veto every candidate on the machine this feature exists for). Per-client exclusion-instruction copy lives here, single-sourced for the future C3 Doctor.
- **`transport.py`** — `choose_fabric_subnet(occupied, *, candidates=…, preferred=())`: exclusion-contained candidates rank first, same static base order otherwise, same collision check for every candidate; `preferred=()` is byte-for-byte the base behavior. `configure_link`'s subnet selection now detects both hosts' VPN profiles from the interface reading already in hand and prefers their exclusion union — replacing the old hardcoded `10.0.1.1/24` fallback.
- **`routes.py`** — each fabric host gains a `"vpn"` field (`present`/`client`/`full_tunnel`/`exclusions`) in `GET /fabric` and the autoconfigure response. **B1 integration**: the first time a host shows `full_tunnel` during autoconfigure, records one info-severity incident (`vpn_full_tunnel`) — once per host per server run.
- **`dashboard.js`/`_cluster.html`** — pre-warning banner before any fabric setup UI, naming the client and the plan ("oMLX will pick link addresses the VPN ignores and verify the link end-to-end before use").

## Tests

`tests/test_cluster_vpn.py` (13, new) — dated WARP full-tunnel `netstat`/`warp-cli` fixtures, split-tunnel exclusions, Include-mode and garbage tolerance, utun-routed `10/8` marks the whole `/8` hostile while catch-alls don't. `tests/test_cluster_fabric_subnet.py` (9, new) — exclusion-contained candidate wins, non-matching exclusion preserves static order, `preferred=()` unchanged. `test_cluster_link_status.py`: two pre-existing failures (asserting the old `10.0.1.x` default) repaired to expect the new addressing — 59 passed, 0 failed.

**192 passed, 3 failed** across the touched suites on the stacked base — the 3 are the same pre-existing `test_cluster_routes.py` failures confirmed throughout this branch's history, unrelated to this change.


jundot:
Thanks, the vpn detection module itself is the best-engineered piece in the series, the clean-host early-out and the failure-tolerant parsers are right. What blocks it: it bundles an unrelated behavior change (the default fabric subnet moves from 10.0.1.1/24 to 172.16.99.0/24 for every cluster, two pre-existing tests had to be rewritten to match), and a corporate VPN routing both 10/8 and 172.16/12 vetoes all 27 candidates, so choose_fabric_subnet raises and Start Cluster returns a 503 where it used to work. A reboot also silently re-addresses existing clusters, and the WARP fixtures contradict themselves (10/8 can't be routed through utun4 and listed in the exclude block at the same time). Please split the subnet-default change into its own PR with a migration note, give the chooser a last-resort fallback instead of a hard raise, and re-capture the fixtures from the real machine.

alytaphoenix:
Rebased onto current main (clean, no conflicts) and pushed fixes for all three.

**Bundled default-subnet change**: swapped `_FABRIC_SUBNET_CANDIDATES`'s static order back to 10.x-first (closest match to the previous single hardcoded 10.0.1.1/24 default), with 172.16.0.0/12 as the second static tier. The VPN-avoidance mechanism itself is unchanged -- a detected VPN's `preferred` exclusions still promote a 172.x candidate ahead of the 10.x default the instant they're needed (the incident's actual fix) -- it just no longer silently changes the default for every cluster regardless of whether a VPN is present. Migration note: a cluster that already auto-picked 172.16.99.x under the old default keeps working (nothing re-addresses a link with a live current address); only a *fresh* setup sees the restored 10.x default now. If you still want 172.x promoted to the default outright, that's a deliberate follow-up of its own with its own note, not something to fall out of this PR.

**Hard raise → last-resort fallback**: replaced the `raise LinkSetupError` when every candidate collides with a fewest-collisions pick instead, mirroring `_occupied_networks`' own existing philosophy ("a subnet we cannot prove free is still better than the old hardcoded default"). `assess_link()` after addressing remains the real authority on whether the chosen link actually works, same as before -- this just stops Start Cluster from hard-failing with a 503 before that check even gets a chance to run.

**WARP fixture contradiction**: I don't have a live WARP-connected Mac in this environment to re-capture from, so I corrected the exclude-settings fixture instead (removed `10.0.0.0/8` from the Exclude block) to match the module's own incident narrative -- the org policy did NOT exclude 10.0.0.0/8, which is exactly why the original 10.0.1.x fabric assignment got silently tunneled. If you have rig access, a genuine live re-capture would still be worth doing to replace my corrected-but-synthetic version.

Also documented (not fixed -- it's the fabric-intent store's job, layered on top of this module): `configure_link` has no memory of a previous choice, so a reboot that drops an ifconfig address re-chooses fresh from the live collision environment, which can differ run to run. #2875 (C2) is the fix for that.

Full 10,549-test suite passes.

PR #2869, opened 2026-08-19
fix(cluster): retry an unsupported reasoning_effort against rank-zero

alytaphoenix:
## ⚠️ Stacks on #2867 → #2866 → #2864 → #2849 → #2848 → #2819

Fabric-doctor series. Diff includes the parents' commits until they merge; review only the last commit (`fix(cluster): retry an unsupported reasoning_effort…`) — it collapses to just that once the parents land.

## Summary

Every local engine (`batched`/`vlm`/`dflash`) falls back through an alias, then drops `reasoning_effort` entirely, when a model's chat template rejects the requested value (`apply_chat_template_with_reasoning_effort_fallback` in `reasoning_effort.py`). **The distributed engine never did** — it forwards `reasoning_effort` straight through to rank-zero's private mlx-lm server, which renders the template itself and returns a hard 404 (*"Unexpected reasoning effort high. Supported types are xhigh (default), medium, and low."*) for any value its template doesn't literally support.

This is **100% reproducible, not flaky** — found via a real incident: ~60 repeated failures over an active session, because the client kept retrying the identical broken request and getting the identical instant 404 every time, which from the user's side looked like the cluster "grinding to a halt."

## What changed

Since only rank-zero can render the template (the coordinator has no local copy to try-and-catch against), this mirrors the same alias-then-drop ladder **reactively, at the HTTP boundary**:

- New `_reasoning_effort_retry_payloads(payload, detail)` computes at most two retry payload variants from a failed response's detail, reusing `reasoning_effort.py`'s own `_normalized_input`/`_fallback_candidate` tables so behavior matches the local engines exactly.
- `_raise_for_backend`'s detail extraction is factored out as `_backend_error_detail` so the retry decision and the final error message share one implementation.
- All four call sites (`chat`/`generate`/`stream_chat`/`stream_generate`) retry through the variants in order, **bounded to at most 3 total attempts** — a client that always sends an unsupported value can never turn this into an unbounded loop.
- **Non-streaming** (`chat`/`generate`): a small retry loop over the existing `POST`.
- **Streaming** (`stream_chat`/`stream_generate`): rank-zero's template rendering runs synchronously before generation starts, so a 400+ status always arrives before any SSE chunk is emitted (confirmed by the existing `if response.status_code >= 400: await response.aread()` guard). Restructured as a bounded loop that reopens `client.stream()` per retry payload and falls through to the **unchanged** per-line processing body only once a response succeeds — no output can ever be relayed to the caller before a good response is confirmed.

## Tests

`tests/test_distributed_engine.py` (+9): pure retry-payload computation (alias tier, drop-only tier, ignores unrelated failures, ignores when not requested); real end-to-end retries through `httpx.MockTransport` for `chat`, `generate`, and `stream_chat` (asserting the exact sequence of `reasoning_effort` values sent); bounded-retry-then-raise for a permanently-rejected value (never more than 3 attempts, confirmed via call count); and a regression test that an unrelated 404 (`"model not found"`) is **not** retried.

Also extended the local shim test runner with `asyncio.iscoroutinefunction` support — this suite's existing `@pytest.mark.asyncio` tests were previously silently skipped by the shim (no async runner); they now run for real (**24 pre-existing + 9 new = 33 passed**, confirming no regression in the existing chat/stream/tool-call/error-surfacing coverage this change touches).

### DIFF
--- a/omlx/engine/distributed.py
+++ b/omlx/engine/distributed.py
@@ -18,10 +18,62 @@ import httpx
 
 from ..cluster.deployment import ClusterDeployment
 from ..cluster.launch import DistributedJobSupervisor, DistributedLaunchError
+from ..reasoning_effort import _fallback_candidate, _normalized_input
 from .base import GenerationOutput
 from .batched import BatchedEngine
 
 logger = logging.getLogger(__name__)
+
+
+def _reasoning_effort_retry_payloads(
+    payload: dict[str, Any], detail: str
+) -> list[dict[str, Any]]:
+    """Payload variants to retry after rank-zero rejects ``reasoning_effort``.
+
+    Local engines render the chat template in-process, so they can try the
+    requested value, catch a template error, and fall back
+    (``apply_chat_template_with_reasoning_effort_fallback`` in
+    ``reasoning_effort.py``). The distributed engine cannot: only rank-zero's
+    private mlx-lm server renders the template, and it already told us — via
+    this failed response — which value it rejected. This mirrors the same
+    alias-then-drop fallback ladder reactively, at the HTTP boundary.
+
+    Returns at most two payloads (alias fallback, then reasoning_effort
+    dropped entirely), so a client that always sends an unsupported value can
+    never turn into an unbounded retry loop. Returns ``[]`` when the failure
+    is not about reasoning_effort, or there is nothing to retry.
+    """
+
+    if "reasoning effort" not in detail.lower():
+        return []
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, this is the most valuable fix in the series and it's fully independent of the doctor stack, so I reparented the branch onto main to land it first, hope the force-push doesn't get in your way. I'll push a follow-up that tries the normalized value first (local engines normalize before the first render, the cluster path currently skips that tier), plus a black pass and a log line on remap or drop.

PR #2870, opened 2026-08-19
feat(cluster): evict competing local models after a memory-attributed activation failure

alytaphoenix:
## ⚠️ Stacks on #2869 → #2867 → #2866 → #2864 → #2849 → #2848 → #2819

Fabric-doctor series. Diff includes the parents' commits until they merge; review only the last commit (`feat(cluster): evict competing local models…`) — it collapses to just that once the parents land.

## Summary

A cluster rank competes for unified memory with any standalone model a node's own oMLX server loaded through its normal (non-cluster) path. Hit in practice: a peer had a model loaded **locally** while the coordinator tried to activate the same model as a cluster rank on that Mac. Each retry got **worse** (higher actual usage, higher admission ceiling) because the local load kept growing/competing — only a human manually confirming no local model was loaded let the next retry succeed.

## Design: reactive, not proactive

Admission already reads each peer's **live** ceiling (deliberately admitting cluster shards alongside existing local usage when there's headroom), so unconditionally evicting on every activation would kill legitimately coexisting local models on Macs the user is actively using. Recovery only runs **after** a memory-attributable launch failure, and only on the node(s) that failure actually names.

## What changed

- **`launch.py`** — `_REMOTE_LOCAL_EVICTION` (peer-side loopback script) + `evict_remote_local_models()`. The coordinator has no HTTP route to a peer, only host-key-verified SSH — reuses the exact pattern `probe_remote_admission_ceiling` already established (SSH in, hit the peer's own admin API over loopback, never a cold `import omlx` since that blows the SSH timeout). The script reads the peer's own persisted `auth.secret_key` from `settings.json` and mints a session token with `itsdangerous` (already a dependency) to call `POST /admin/api/models/{id}/unload` locally — SSH login to the Mac is already admin-equivalent, so this grants nothing new, and the secret/cookie never appear in the script's JSON output. Always exits 0; failure is confessed in the JSON report, matching this file's other remote-probe convention. Interpreter-fallback discipline matches the ceiling probe (#2680).
- **`routes.py`** — `_MEMORY_FAILURE_RANK` parses the launcher's `"rank N (node): InsufficientMemoryError"` lines; `_memory_squeezed_hosts` resolves implicated hosts by node_id, falls back to rank index, then to every host for a memory failure that names no rank. `_evict_local_models_on_host` dispatches locally via the engine pool for the coordinator (no SSH hop needed) or via the peer script otherwise. `_evict_competing_local_models` orchestrates across every implicated host, records a WARN incident (B1) describing what was freed, and appends operator guidance to the activation error detail. Wired into `activate_cluster_deployment`'s existing `DistributedLaunchError` arm: **the original incident and HTTPException are always recorded/raised first**; recovery runs inside its own try/except so a crash there can only add a second WARN incident, never replace or mask the original 503.
- Pinned models, cluster-source entries, and still-loading models are skipped on both the coordinator-direct and peer-SSH paths and reported separately.

## Tests

`tests/test_cluster_local_eviction.py` (11, new): SSH boundary (report round-trip, interpreter fallback, non-JSON rejection), the peer script itself executed in-process against a faked `urlopen` + a real temp `settings.json` (pinned/cluster/virtual/loading filtering, cookie signed against the persisted secret and configured port, draining vs evicted, unreachable-server safety), and the routes recovery logic (rank targeting via the exact error format observed in production, coordinator engine-pool path, fail-safe on a crash, no-op for non-memory failures, freed/nothing-to-free messaging).

Verified independently via `git stash` A/B: the pre-existing failures in `test_cluster_routes.py` (3) and `test_cluster_launch.py` (2) are identical before and after this change.

### DIFF
--- a/omlx/admin/static/js/dashboard.js
+++ b/omlx/admin/static/js/dashboard.js
@@ -3277,7 +3277,7 @@
                     );
                 const fit = this.clusterCatalogueFit(model.model_path);
                 if (fit?.fits === true) {
-                    const tensorSize = Number(fit.tensor_parallel_size || 1);
+                    const tensorSize = this.clusterFitTensorParallelSize(fit);
                     if (this.clusterTensorParallelOptions().includes(tensorSize)) {
                         // The catalogue arrived at this topology using the
                         // same fit planner. Preview that proven topology rather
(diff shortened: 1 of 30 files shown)


jundot:
Thanks, the ssh and token mechanics here are careful work. Closing this one though: a server-initiated kill of models a user loaded on their own Mac is a product call I'm not making right now, and a regex over free-form error text is too fragile a base for a destructive action even behind a gate. If this comes back later it will need to be opt-in and keyed on structured failure attribution, rank and node as fields rather than parsed prose. The token minting pattern and C5's advertised admin_port are both worth keeping for that day.

PR #2873, opened 2026-08-19
fix(engine): include embedding/reranker forwards in decode fairness

### LINKED ISSUE #2872: Embedding forwards bypass decode_fairness — sustained /v1/embeddings load starves concurrent chat decode (13.8× latency; patch included)
## Summary

`scheduler.decode_fairness` (from #2031 / #2622) arbitrates prefill-vs-decode across chat engines, but `/v1/embeddings` forwards never enter that protocol: `EmbeddingEngine.embed()` submits its forward passes straight onto the global MLX executor (`omlx/engine/embedding.py`, chunk loop) with no chunk-time cap, no hold window, and no decode-time debt. Under a sustained embedding workload, a concurrent chat completion degrades **13.8×** while the embedding requests themselves stay fast — the embed side is always the aggressor, never the victim.

We hit this consolidating a RAG ingest pipeline onto oMLX (document ingestion embeds while users chat on the same box). We implemented the fix described below and measured a 4.2× improvement; happy to send it as a PR.

## Environment

- Reproduced on oMLX 0.6.1 (installed) and current `main` (c1a3d44, from source); the relevant code paths (`omlx/engine/embedding.py`, `omlx/decode_activity.py`) are identical in both
- MacBook M5 Max 128 GB (also reproduced on M4 Max)
- Chat model: Qwen3.6-35B-A3B (MoE, OptiQ-4bit); embedding model: `mlx-community/Qwen3-VL-Embedding-2B-bf16` with `model_type_override: "embedding"` (context: #1817)

## Repro

1. Load a chat model plus an embedding model (`model_type_override: "embedding"`).
2. Two threads continuously POST `/v1/embeddings`, 8 inputs × ~250 tokens each.
3. In parallel, time `/v1/chat/completions` with `max_tokens=40` (E2E, non-streaming).

## Measurements (A/B/A, source builds of `main` c1a3d44, 19 Aug, M5 Max)

Run order: baseline → patched → baseline again (the second baseline controls for warm OS/Metal caches; solo numbers of the cold first run were noisy, so warm-vs-warm is the honest comparison).

| Arm | chat solo (median) | chat under embed storm (median, n=8) | degradation | embed req latency under storm (median) | embed solo throughput (2 threads, 30 s) |
|---|---|---|---|---|---|
| `main`, cold | 1.50 s | 7.43 s | 5.0× (vs noisy cold solo) | 0.53 s | 3.77 req/s |
| **patched** | 0.52 s | **1.74 s** (1.46–1.82) | **3.4×** | 0.40 s (max 0.52) | 6.13 req/s |
| `main`, warm | 0.53 s | 7.34 s (6.06–7.68) | **13.8×** | 0.38 s | 5.63 req/s |

- Warm-vs-warm: **7.34 s → 1.74 s (4.2×)**. Residual 3.4× vs solo matches the fairness contract at the default share 0.5 (decode gets ~1/3 of GPU wall time → ~3× is the steady-state floor, same as chat-vs-chat).
- Uncontended embedding throughput is unaffected (6.13 vs 5.63 req/s, patched slightly faster within run-to-run noise) — the gate is inert while nothing decodes.
- Embedding latency under mixed load stays healthy (median 0.40 s, max 0.52 s): holds are bounded by `chunk_time × share`.

Knobs tested before patching (one at a time, all reverted): `embedding_batch_size` 32→8 — no change (requests carry 8 inputs, one chunk either way, and smaller chunks alone create no idle gaps since the loop submits back-to-back); `prefill_priority` context→speed — no change (only affects scheduler prefills, embeds never pass through the scheduler); `burst_decode_mode` balanced→aggressive — worse (~10 s).

## Why (code reading)

- Chat engines run `scheduler.step()` on per-engine executor threads with their own Metal streams (#1248). `EmbeddingEngine` runs each per-chunk forward on the global MLX executor. Two streams submit concurrently; as the fairness comment in `scheduler.py` puts it, Metal cannot preempt a running kernel — bounded chunk duration plus hold windows *is* the interleave mechanism. The embed path bounds nothing and holds never: two storm threads keep ~150–250 ms forwards queued back-to-back, so the 35B's ~10–15 ms decode steps only slip in between embed forwards. 40 tokens × ~180 ms average wait ≈ the measured 7.3 s.
- The chat prefill path participates in the cross-engine protocol via `DecodeActivityRegistry` (`_prefill_gate_open`, `_accrue_decode_debt`, `_contended_prefill_cap`). `EmbeddingEngine` neither checks `others_decoding()` nor extends the shared hold.
- Compounding: `_embed_sync` runs `mx.clear_cache()` after **every** chunk. That flush is process-global — it dumps the decoding engine's warm buffer pool each time, which is exactly what the scheduler's `_should_clear_after_chunk` deliberately avoids under decode contention.
- `RerankerEngine` has the same pattern (`engine/reranker.py`).

## Fix (implemented, measured above)

A small `ForwardFairnessGate` (new `omlx/engine/forward_fairness.py`, ~140 lines mostly docstrings) lets non-scheduler forwards participate in the existing protocol — no scheduler changes, `DecodeActivityRegistry` already provides the primitives:

- **wait_turn()** — on the MLX executor thread, immediately before the forward: wait out the shared hold deadline while another engine decodes (exits early when the decode finishes). Running it on the executor thread closes the check-then-run race between concurrent embed requests.
- **chunk_cap()** — while contended, cap items per forward from the measured per-item duration EMA against `_DECODE_STALL_TARGET_MS` (analog of `_contended_prefill_cap`).
- **settle(dt, items, contended)** — after the forward: update the EMA and extend the shared hold by `dt × _DECODE_FAIR_SHARE` (the same debt scheduler prefills pay).
- **should_clear_cache()** — skip the per-chunk process-global `mx.clear_cache()` while another engine decodes (analog of `_should_clear_after_chunk`).

`EmbeddingEngine` uses all four around its chunk loop; `RerankerEngine` gets the coarse wrapper (wait + settle around its single forward). The gate honors the live `decode_fairness` admin toggle (it holds the shared SchedulerConfig object the admin API mutates), and is fully inert while nothing is decoding. Unit tests in the style of `test_decode_fairness.py` included (18 tests; the existing 38 decode-fairness tests stay green).

Happy to open a PR with the patch — or adjust to whatever shape you prefer (e.g. hoisting the fairness constants out of `scheduler.py`).

joergbirkhold:
Fixes #2872.

`EmbeddingEngine` and `RerankerEngine` submit their forward passes straight onto the global MLX executor and never pass through `Scheduler.step()`, so the `decode_fairness` protocol (chunk caps, shared hold windows, decode-time debt) cannot see them: a sustained `/v1/embeddings` load keeps the GPU command queue full with back-to-back prefill-sized forwards and starves concurrent chat decode — measured 13.8× completion-latency degradation while the embedding requests themselves stay unaffected (full data in #2872). The per-chunk process-global `mx.clear_cache()` additionally dumps the decoding engine's warm buffer pool on every chunk.

## Change

New `omlx/engine/forward_fairness.py` with a `ForwardFairnessGate` that lets non-scheduler forwards participate in the existing protocol — no scheduler changes, `DecodeActivityRegistry` already provides the primitives:

- `wait_turn()` — on the MLX executor thread, immediately before the forward: wait out the shared hold deadline while another engine decodes (early exit when the victim decode finishes). Running it on the executor thread closes the check-then-run race between concurrent requests.
- `chunk_cap()` — while contended, cap items per forward from a measured per-item duration EMA against `_DECODE_STALL_TARGET_MS` (analog of `_contended_prefill_cap`).
- `settle(dt, items, contended)` — after the forward: update the EMA and extend the shared hold by `dt × _DECODE_FAIR_SHARE` (the same debt scheduler prefills pay).
- `should_clear_cache()` — skip the per-chunk process-global `mx.clear_cache()` while another engine decodes (analog of `_should_clear_after_chunk`).

`EmbeddingEngine` uses all four around its chunk loop; `RerankerEngine` gets the coarse wrapper (wait + settle around its single forward, and it now receives `scheduler_config` from the engine pool). The gate honors the live `decode_fairness` admin toggle — it holds the shared `SchedulerConfig` object the admin API mutates — and is fully inert while nothing is decoding.

## Measurements (A/B/A on `main` c1a3d44, M5 Max 128 GB; Qwen3.6-35B-A3B chat + Qwen3-VL-Embedding-2B-bf16, 2 embed threads × 8×~250-token inputs)

| Arm | chat solo (median) | chat under embed storm (median, n=8) | embed solo throughput |
|---|---|---|---|
| `main`, warm | 0.53 s | 7.34 s (**13.8×**) | 5.63 req/s |
| **patched** | 0.52 s | **1.74 s** (**3.4×**) | 6.13 req/s |

- 7.34 s → 1.74 s (4.2×); the residual 3.4× vs solo matches the fairness contract at the default share 0.5 (decode holds ~1/3 of GPU wall time, same as chat-vs-chat).
- Uncontended embedding throughput unchanged (within run-to-run noise) — the gate is inert without a live decode.
- Embedding latency under mixed load stays bounded (median 0.40 s, max 0.52 s): holds are capped at `chunk_time × share`.

## Tests

`tests/test_forward_fairness.py` (18 tests, style of `test_decode_fairness.py`): contention detection incl. live-toggle parity, hold waiting with early exit, debt accrual (incl. decode starting mid-forward), EMA-derived chunk caps, clear-cache decisions. The existing 38 `test_decode_fairness.py` tests and the embedding/reranker/engine-pool suites stay green, and the full local test suite passes on the branch: **9963 passed, 112 skipped** (10m20s, M5 Max).

### DIFF
--- a/omlx/engine/base.py
+++ b/omlx/engine/base.py
@@ -589,6 +589,19 @@ class BaseNonStreamingEngine(ActivityTrackingMixin, ABC):
     support streaming or chat completion interfaces.
     """
 
+    def set_memory_soft_limit(self, soft_limit_bytes: int) -> None:
+        """Receive the ProcessMemoryEnforcer's soft watermark.
+
+        Non-scheduler engines have no ``_memory_limit_bytes`` for the
+        enforcer to propagate into; engines with a decode-fairness gate
+        (see forward_fairness.py) use the watermark to resume per-forward
+        cache clearing under memory pressure. Default is a no-op for
+        engines without a gate.
+        """
+        gate = getattr(self, "_fairness", None)
+        if gate is not None:
+            gate.set_memory_soft_limit(soft_limit_bytes)
+
     @property
     @abstractmethod
     def model_name(self) -> str:
--- a/omlx/engine/embedding.py
+++ b/omlx/engine/embedding.py
@@ -10,6 +10,7 @@ streaming or chat completion.
 import asyncio
 import gc
 import logging
+import time
 from typing import Any, Dict, List, Optional, Union
 
 import mlx.core as mx
@@ -17,6 +18,7 @@ import mlx.core as mx
 from ..engine_core import get_mlx_executor
 from ..models.embedding import EmbeddingOutput, MLXEmbeddingModel
 from .base import BaseNonStreamingEngine
+from .forward_fairness import ForwardFairnessGate
 
 logger = logging.getLogger(__name__)
 
@@ -71,6 +73,12 @@ class EmbeddingEngine(BaseNonStreamingEngine):
             )
         self._batch_size = max(1, int(batch_size))
         self._model: Optional[MLXEmbeddingModel] = None
+        # Embedding forwards bypass the Scheduler, so they must take part
+        # in decode fairness themselves (see forward_fairness.py). The key
+        # only has to differ from every Scheduler's registry key.
+        self._fairness = ForwardFairnessGate(
+            f"embed:{model_name}:{id(self):x}", scheduler_config
+        )
 
     @property
     def model_name(self) -> str:
@@ -166,10 +174,21 @@ class EmbeddingEngine(BaseNonStreamingEngine):
             order = sorted(range(len(input_items)), key=lambda i: _input_length(input_items[i]))
             ordered_items = [input_items[i] for i in order]
 
-            for start in range(0, len(ordered_items), batch_size):
-                batch = ordered_items[start:start + batch_size]
+            fairness = self._fairness
+            index = 0
+            while index < len(ordered_items):
+                # While another engine decodes, forwards are capped in
+                # time (via the measured per-item duration) so the
+                # victim's stall stays bounded; uncontended forwards use
+                # the full configured batch size.
+                cap = fairness.chunk_cap()
+                size = batch_size if cap is None else max(1, min(batch_size, cap))
+                batch = ordered_items[index:index + size]
+                index += len(batch)
 
-                def _embed_sync():
+                def _embed_sync(batch=batch):
+                    contended = fairness.wait_turn()
+                    start_ts = time.perf_counter()
                     try:
                         return model.embed(
                             inputs=batch,
@@ -179,7 +198,11 @@ class EmbeddingEngine(BaseNonStreamingEngine):
                         )
                     finally:
                         mx.synchronize()
-                        mx.clear_cache()
+                        fairness.settle(
+                            time.perf_counter() - start_ts, len(batch), contended
+                        )
+                        if fairness.should_clear_cache():
+                            mx.clear_cache()
 
                 output = await loop.run_in_executor(get_mlx_executor(), _embed_sync)
                 embeddings.extend(output.embeddings)
@@ -188,7 +211,7 @@ class EmbeddingEngine(BaseNonStreamingEngine):
                     dimensions = output.dimensions
                 self._update_activity(
                     activity_id,
-                    completed_items=min(start + len(batch), len(input_items)),
+                    completed_items=min(index, len(input_items)),
                     token_count=total_tokens,
                     dimensions=dimensions,
                 )
--- a/omlx/engine/forward_fairness.py
+++ b/omlx/engine/forward_fairness.py
@@ -1,10 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
 """Decode-fairness participation for engines that bypass the Scheduler.
 
-EmbeddingEngine and RerankerEngine submit their forward passes straight
-onto the global MLX executor and never pass through ``Scheduler.step()``,
-so the ``decode_fairness`` protocol (chunk caps, shared hold windows and
-decode-time debt -- see scheduler.py) cannot see them: a sustained
+EmbeddingEngine submits its forward passes straight onto the global MLX
+executor and never passes through ``Scheduler.step()``, so the
+``decode_fairness`` protocol (chunk caps, shared hold windows and
+decode-time debt -- see scheduler.py) cannot see it: a sustained
 ``/v1/embeddings`` load keeps the GPU command queue full with back-to-back
 prefill-sized forwards and starves concurrent chat decode. Metal cannot
 preempt a running kernel, so bounding forward duration and leaving the GPU
@@ -12,6 +12,11 @@ quiet during hold windows IS the interleave mechanism; this gate applies
 those two rules to non-scheduler forwards, reusing the process-global
 ``DecodeActivityRegistry`` hold deadline that scheduler prefills already
 honor -- concurrent embedders and prefillers pause together.
+
+RerankerEngine shares the gap, but its models run several forward passes
+inside one ``model.rerank()`` call, so gating the request boundary would
+not let decodes interleave between documents; wiring the gate into those
+internal forward boundaries is a follow-up.
 
 The gate is inert while no other engine is decoding (uncontended forwards
 run back-to-back exactly as before), and it honors the same live
@@ -23,6 +28,8 @@ from __future__ import annotations
 
 import time
 from typing import Any
+
+import mlx.core as mx
 
 from ..decode_activity import get_decode_activity
 from ..scheduler import (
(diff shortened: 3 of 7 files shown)


jundot:
Thanks for working on this. I reproduced the starvation locally with Qwen3-4B and Qwen3-VL-Embedding-2B. The patch reduced the median chat latency under embedding load from 4.33s to 1.53s, so the embedding fairness gate is clearly effective.

I found one memory-safety gap that I would like to see addressed before merging. `ForwardFairnessGate.should_clear_cache()` currently skips cache cleanup whenever another engine is decoding. Unlike `Scheduler._should_clear_after_chunk()`, it does not resume cleanup after memory usage crosses the soft limit. Since oMLX allows MLX to retain cached buffers up to the total memory limit, a sustained mixed workload with varying input shapes could keep growing the buffer cache under pressure.

I also think the reranker path needs a little more work. The causal-LM and Jina implementations perform multiple forward passes inside a single `model.rerank()` call, while the current gate only waits before the whole call and settles after it finishes. This means a large reranking request still cannot yield between documents or chunks. Could you either apply the gate at those internal forward boundaries or keep this PR focused on embeddings for now?

It would also be helpful to add engine-level coverage for the memory-pressure behavior and the multi-forward reranker paths. With those adjustments, I would be happy to take another pass.

joergbirkhold:
Thanks for the fast review and the independent repro — glad the numbers held up on your setup.

Both points addressed in 1503695 (rebased on current main):

**Memory-safety gap:** `should_clear_cache()` now mirrors `Scheduler._should_clear_after_chunk` fully. `ProcessMemoryEnforcer._propagate_memory_limit` pushes the same soft watermark it gives schedulers into non-scheduler engines (`BaseNonStreamingEngine.set_memory_soft_limit` → gate), and the contended skip only holds while `mx.get_active_memory() + mx.get_cache_memory()` stays below it — the cache is part of the usage on purpose, since cached-but-free buffers are exactly what the skip lets accumulate. While no watermark has been propagated (guard off, or before the first enforcer tick), every contended forward clears — conservative parity with the scheduler's `_memory_limit_bytes <= 0` branch.

**Reranker:** reverted to upstream state; the PR is embeddings-only now. You're right that request-boundary gating can't interleave between documents when `model.rerank()` runs several forwards internally — I'd rather wire the gate into those forward boundaries properly in a follow-up than fake it here.

**Tests:** added engine-level coverage for the flush behavior (fake model on the real global MLX executor; the mx module patched per import site): uncontended clears, contended-below-watermark skips, memory pressure clears, missing watermark clears. Full suite on the branch: 10099 passed, 112 skipped; the single failure (`test_dflash_laguna::test_target_ops_logits_last_only_slices_before_lm_head`) reproduces on pristine current `main` (146d272) as well, so it is preexisting and unrelated.

joergbirkhold:
Status ping, no new code — just answering the question from the 0.6.3rc3 notes about review order: **this is my only open PR in the repo**, so nothing of mine needs to be ranked against it.

Both review points were addressed in 1503695 on 20 Aug: memory-watermark parity with `Scheduler._should_clear_after_chunk` (soft limit propagated through `ProcessMemoryEnforcer._propagate_memory_limit` → `BaseNonStreamingEngine.set_memory_soft_limit`), the reranker path reverted so the PR is embeddings-only, and engine-level coverage added for the memory-pressure behaviour you asked about.

Two things that may lower the cost of picking this up:

- I rebased the two commits onto `v0.6.3rc3` locally — they apply with **no conflicts** (5 files, +496/-5), so the branch sitting behind main is not a merge risk.
- Nothing upstream has touched `omlx/engine/embedding.py` since, so there is no overlap with the ANE/cache work in 0.6.3.

No urgency intended given the review load — happy to rebase, split further, or re-run the benchmarks whenever it suits your queue.

joergbirkhold:
Hi @jundot — just checking in now that 0.6.3 and 0.6.4 are out: is there anything else you'd like changed on this one? The branch is rebased-clean and mergeable against current main; happy to rebase once more or split anything out if that helps. No rush if it's queued for after the release work.

jundot (approved):
Thanks for addressing the cache cleanup and narrowing this to embeddings. I verified watermark propagation and reproduced the mixed-load improvement with Qwen3-4B and Qwen3-VL-Embedding-2B using the live fairness toggle. I'll align the gate's memory accounting with the current scheduler's process-footprint checks in a small follow-up. LGTM.

PR #2875, opened 2026-08-19
feat(cluster): durable fabric addressing via recorded intent (fabric-doctor C2)

alytaphoenix:
## ⚠️ Stacks on #2870 → #2869 → #2867 → #2866 → #2864 → #2849 → #2848 → #2819

Fabric-doctor series. Diff includes the parents' commits until they merge; review only the last commit (`feat(cluster): durable fabric addressing via recorded intent…`) — it collapses to just that once the parents land.

## Summary

A good fabric address assignment should survive reboots and oMLX's own re-runs instead of being re-derived (and potentially re-picked) every time. Layers a three-tier decision onto `configure_link` and, where the OS allows it, persists the address into system network configuration rather than only the live interface.

## What changed

- **New `omlx/cluster/fabric_intent.py`** — `FabricIntent` (`subnet`/`hosts`/`chosen_by`/`reason`/`recorded_at`/`addressing`) + `FabricIntentStore` (atomic save, fails closed on a corrupt file — the `incidents.py`/`registry.py` pattern), wired in `server.py` beside the other cluster stores.
- **`transport.py`** — `configure_link`'s address selection is now three tiers: **(1)** an existing endpoint subnet is kept **iff it still passes the collision check** (a colliding existing address — e.g. the WARP-swallowed-10.0.1.x incident's shape — is now re-addressed instead of blindly kept); **(2)** a recorded `FabricIntent` for this exact host pair is re-applied verbatim iff collision-free; **(3)** `choose_fabric_subnet` picks fresh and the choice is recorded as intent **only after** the post-addressing `assess_link` call confirms the link is actually ready. An unconfigured store degrades to today's behavior; an intent-write failure never fails an otherwise-ready link.
- New `_authorized_networksetup` (validated the same way `_authorized_ifconfig` validates its inputs: service name against a tight allowlist regex, address/mask through `ipaddress`, `shlex.join` for the privileged command) writes a `networksetup -setmanual` assignment, which — unlike a raw `ifconfig` address — survives reboot and interface renumbering. Reuses the existing `_remote_gui_authorize` Aqua-session bridge unchanged; folds `-createnetworkservice` into the same shell command when the service doesn't exist yet, so first-time setup is still one admin prompt.
- New `parse_hardware_ports` + `parse_network_services`, joined by hardware-port name to map an interface to its owning service. **This corrects the design doc's own assumption**: `networksetup -listallhardwareports` never names a service, only a hardware port; the service name only appears in `-listnetworkserviceorder`. A naive single-command parse would have issued `-setmanual "Thunderbolt 2"` and failed.
- Netmask parameterized (`_netmask_for_prefix`) through both authorized paths instead of the hardcoded `255.255.255.0`.

## The empirical spike (read-only half only)

The design calls for a live spike to determine whether `networksetup -setmanual` works on the RDMA-active interface (which may be a Thunderbolt Bridge member rather than owning its own service). Read-only recon against the real 2-Mac rig (`-listallhardwareports`/`-listnetworkserviceorder`/`-getinfo`, `ifconfig -a` — no writes) found **neither Mac uses a Thunderbolt Bridge**: EXO software replaced it with three standalone per-port services ("EXO Thunderbolt 1/2/3"), each bound directly to a hardware port. The RDMA-active interfaces each map to their own service, so the networksetup path is structurally usable here.

However, the peer's own `-getinfo` for its EXO service still shows DHCP with a stale `169.254.x` address while the live interface actually carries the real fabric IP — service config and live `ifconfig` state visibly diverge on this rig — and the **write** half of the spike (`-setmanual` itself) could not be validated live without risking the cluster's active traffic. `configure_link` therefore defaults `prefer_service_addressing=False`; the real runtime path on this hardware is tier 3's `_authorized_ifconfig` fallback with `addressing="ifconfig"` recorded — exactly the design's documented contingency. The service-addressing code is implemented and unit-tested end to end; flipping the default is a one-line follow-up once the write-spike can run against an idle rig.

## Tests

`tests/test_cluster_fabric_intent.py` (25, new): store round-trip, corrupt file fails closed, provenance validation, hardware-port/service-name parser fixtures (including the disabled-service and no-owning-service cases), `_authorized_networksetup` command shape and refusal matrix, netmask-from-prefix, all three `configure_link` tiers, colliding-intent-falls-through, wrong-host-pair intent ignored, intent-written-only-after-ready. Existing `test_cluster_fabric_subnet.py` untouched byte-for-byte (regression net).


jundot:
Thanks, the three-tier keep/replay/choose ordering is the right shape and the intent store itself is solid. Two blockers: own_networks hardcodes /24 (transport.py:1133), so a fabric on any other mask fails the exact-equality check against hostile_networks and tier 1 re-addresses a working manually-configured link behind an admin prompt, and the entire networksetup half (~180 lines of privileged command building, the durable part of the title) sits behind prefer_service_addressing which nothing ever sets to True, so it ships as dead code. Please compare by overlap with real prefixes instead of equality (with a test for a non-/24 fabric and one where a tunnel claims the exact fabric /24), and either enable the networksetup path with a revert story and prior-config capture, or split it out and land the store plus tier logic alone. Rebasing this below #2849 onto the C4 rework would also let the initial.ready early exit protect existing clusters again.

alytaphoenix:
Merged current main in (a straight rebase kept conflicting repeatedly in unrelated admin-dashboard JS that drifted independently since this forked; one 3-way merge resolved cleanly except two JS conflicts, both taken as main's newer logic since they're unrelated to fabric addressing) and pushed fixes for both blockers.

**own_networks / equality check**: added `_own_network`, which reads the real prefix from the same probe `hostile_networks` is built from (falling back to /24 only when that probe has no match for the address). Switched the comparison from exact equality to subnet-of-or-equal. I tried plain overlap first and it broke an existing test (`test_a_colliding_existing_subnet_is_readdressed_not_kept`): a hostile entry *broader* than own (a utun routing all of 10/8 while the link sits at 10.0.1.0/24) is a real, independent collision risk for every other candidate in that range, and overlap-based exclusion blanket-removed it from the collision set just because it happens to cover this link's own current address. subnet-of-or-equal only excludes a hostile entry that IS (or is narrower than) an own network, never one that merely contains it -- which is what actually matches "this is my own address" semantics.

**networksetup dead code**: split it out rather than attempting the live validation the docstring itself says hasn't happened -- writing persistent `networksetup` config to a live Mac's Thunderbolt service is a materially more invasive, harder-to-reverse action than I wanted to take without your live rig to validate against. Removed `_authorized_networksetup`, `parse_hardware_ports`, `parse_network_services`, `_interface_service`, their now-unused regex constants, and the tests that only covered them. `_apply_link_address` always uses the empirically-proven `ifconfig` path now; `parse_thunderbolt_interfaces` (pre-existing, shares two of the same regexes) is untouched. Durability still comes from the intent store replaying the choice -- that half was already correct, per your own confirmation on the model-reload path.

New tests reproduce both your named scenarios (a non-/24 fabric, and a tunnel claiming the exact fabric /24) plus direct unit coverage of `_own_network`. Full 10,574-test suite passes.

Left for a follow-up rather than attempted here: the networksetup write-half itself, once there's a way to validate `-setmanual` against a live RDMA link.

alytaphoenix:
One more update: merged the just-fixed #2867 (C4) branch in, since this branch depends on it and had inherited its original (unfixed) subnet-default bundling and hard-raise issues. Three conflicts (guidance.py/routes.py wording only, transport.py's configure_link docstring where I kept this branch's fuller tier-1/2/3 description over C4's simpler no-C2-yet version), plus three tests here that also exercise tier 3's fresh-pick default and needed updating to the restored 10.x-first order. Full 10,576-test suite passes.

PR #2878, opened 2026-08-19
feat(cluster): Fabric Doctor guided diagnosis flow (fabric-doctor C3)

alytaphoenix:
## ⚠️ Stacks on #2875 → #2870 → #2869 → #2867 → #2866 → #2864 → #2849 → #2848 → #2819

Fabric-doctor series — the culminating item. Diff includes the parents' commits until they merge; review only the last commit (`feat(cluster): Fabric Doctor guided diagnosis flow…`) — it collapses to just that once the parents land.

## Summary

One guided flow runs the transport ladder checks in order, stops at the first red rung, and every finding carries a diagnosis and a concrete fix. Closes the "stale address as an unnamed condition" and "error 60 as silent exponential backoff" failure classes from the design — the Doctor names both explicitly instead of leaving them to be inferred from a hang or a cryptic log line.

## What changed

- **New `omlx/cluster/doctor.py`** — `DoctorFinding`/`DoctorCheck`/`DoctorReport`/`DoctorProbes` (an injectable probe bundle — the same `assess_link` injection pattern used elsewhere, so every check is pure and testable without touching real hosts). `ERRNO_DIAGNOSES` maps JACCL failure output: `60` → ETIMEDOUT / *"firewall or VPN on the fabric path"* with a move-subnet remedy, `61` → ECONNREFUSED / *"the peer's worker isn't listening yet — a launch-order problem, not a network problem"*, unknown → generic + raw detail.
- **Five checks in ladder order, stop-at-first-red** (later checks marked `state="skipped"`; probes are invoked lazily, so nothing past the first red actually runs):
  1. Link/address sanity — 169.254 detection, interface-renumber-aware finding, fix → C2's readdress.
  2. Subnet collision — reuses C4's `hostile_networks`, concrete evidence style (*"WARP routes 10.0.0.0/8 through utun4"*), fix → C2's move-subnet.
  3. Routes + bound connect — wraps C1's `verify_link_reachability`, split into route-pinning vs bound-connect findings.
  4. JACCL probe — `run_fabric_collective_probe` + errno mapping.
  5. RDMA staleness (*"addresses were lost — this happens after reboot"*) + a basic admin-port health read.
- `run_fabric_collective_probe`: a minimal two-rank JACCL handshake across the fabric addresses, launched the way `run_local_collective_smoke` launches loopback ranks, with a hostfile honoring the same RDMA-matrix contract `ClusterDeployment` validates (2×2, null diagonal, `rdma_*` off-diagonal).

## Safety-critical: refuses beside an active deployment

**The probe (and check 4) refuse outright, before any process is spawned, while a deployment is registered on the target hosts** — a probe on live collective ports would perturb the serving collective. `_active_deployment_for` normalizes host aliases (`127.0.0.1`/`localhost`/`::1`) so rank 0 (the coordinator itself) is always matched correctly; a Doctor run naming the local Mac refuses beside *any* active deployment. Both the probe function directly and the check-4 wrapper enforce this independently. **Proven by 4 dedicated tests**: a refusal that never invokes the launcher (a runner stub that would assert-fail if called), the check-4 skip path, local-alias host matching, and correct fail-soft behavior when the registry is unconfigured.

## Other changes

- `routes.py` — `POST /admin/api/cluster/doctor` → `202 {"job_id"}`; `GET .../doctor/{job_id}` → `{"phase","findings","verdict"}`. A minimal, deliberately generic job dict (mirrors the existing `_STAGING_JOBS` pattern) that composes with B2's `StartJobStore` generalization later rather than conflicting with it. Every completed run records one B1 incident and lands in `/diagnostics`.
- `cli.py` — `omlx cluster doctor` (the design's own recorded naming deviation from "omlx fabric doctor", for consistency with the existing `omlx cluster …` namespace).
- `dashboard.js`/`_cluster.html` — minimal Doctor panel (check/state/evidence/fix rows, B1/B3's minimal-render style; full layout is B4's).

## Tests

`tests/test_cluster_doctor.py` (39, new): every check's pass/fail/skip branches with fake probes; stop-at-first-red ordering proven directly (a collision failure leaves the reachability/collective probes uncalled); the errno table; verdict string formats; the four live-deployment-refusal tests above. Plus `+5` in `test_cluster_routes.py` (job lifecycle, incident/diagnostics landing) and `+4` in `test_cluster_cli.py` (doctor invocation, stubbed runner).

Verified independently: **107 passed, 3 failed** across the touched suites on the stacked base — the 3 are the same pre-existing `test_cluster_routes.py` failures confirmed throughout this branch's history.


jundot:
Thanks, the stop-at-first-red design and the injectable probe bundle are right, and the registry-refusal tests are exactly the discipline I want everywhere. The blocker is inherited: check 3 wraps verify_link_reachability, which under #2849's bug always fails on a healthy fabric, so the doctor's flagship output today is a confident false firewall accusation and checks 4 and 5 (the most valuable part) can never run in production. Beyond that: the 169.254 detection can't fire because the probe layer already filters those addresses out, check 2 has the same self-exclusion blind spot as C2, an admin_port failure hard-fails the whole report despite being a heuristic read, and the remedies point at fix buttons that don't exist in the UI. Please rework after #2849 lands, add one test that drives check 3 through the real verify path with an ssh-shaped runner fake (that single test would have caught the whole class), and downgrade admin_port to a warning. Two more for the resubmit: POST /doctor needs a server-side single-flight guard (it spawns real processes and the only guard today is a client-side flag), and the collective probe reserves one loopback port for a two-rank two-host launch, the span needs to be per-rank and valid on the peer too.

alytaphoenix:
Reworked per review. All items addressed in b68822fc:

**1. Check 3's tests never exercised the real verify path.** Every existing test stubbed `DoctorProbes.verify` with a canned `(bool, str)` tuple, so nothing here would have caught #2849's class of bug (which is now fixed on main). Added `test_check3_runs_through_the_real_verify_path_with_ssh_shaped_runner`: it wires the real `verify_link_reachability` through an ssh-shaped runner fake (same shape `_run_link_command`'s real argv takes) and drives it through `run_fabric_doctor` end-to-end. Verified it fails if `verify_link_reachability`'s own direction-iteration logic regresses, not just if the doctor-level plumbing breaks.

**2. Dead 169.254 detection.** Confirmed via `transport.py`'s `_interface_address` that `probe_host_interfaces` already drops unroutable addresses (including 169.254.0.0/16) before `doctor.py` ever sees them — the old check could structurally never fire. Redirected `check_link_address_sanity` to the actually-observable symptom: an RDMA device with no fabric address at all, split cleanly from the pre-existing renumbering detection.

**3. Check 2's self-exclusion blind spot.** Same gap as #2875's `own_networks` fix: exact-equality exclusion missed a hostile entry recorded at a narrower mask than the fabric's own network (e.g. a directly-connected host-route entry for the link's own address). Switched to `net == own or net.subnet_of(own)`. Added a test for this exact case plus a second test proving a hostile network *broader* than the fabric subnet still correctly collides (the failure mode a naive `.overlaps()` fix would reintroduce).

**4. admin_port hard-failing the report.** Added a `"warn"` state, distinct from `"fail"`: real, but doesn't stop the ladder, doesn't gate `DoctorReport.ok`, and doesn't become the verdict's `first_red`. It rides along as a parenthetical suffix on whatever verdict the decisive checks produced. `routes.py`'s severity computation and the dashboard (amber pill, remedy still shown) both updated.

**5. Remedies pointing at nonexistent UI buttons.** Traced it: "Fabric Doctor" is a real button, but it only re-diagnoses (`POST /doctor`) — it never fixes anything. The actual fix mechanism (`configure_link`, called via `/link-setup`) is wired into the existing **Start Cluster** flow (`prepareClusterLink()` in dashboard.js), not exposed as a standalone "Re-address link" / "Move link addresses" control anywhere. Every remedy string now points at Start Cluster instead.

**6. No server-side single-flight guard on `POST /doctor`.** Added a lock-guarded check-and-insert in `_record_doctor_job`: a second request while one job is `queued`/`running` gets a 409 naming the in-flight job id, instead of spawning a second set of real SSH probes and a second collective-probe subprocess. Test uses a blocking fake `run_fabric_doctor` + `threading.Event` to prove the race is actually closed, not just that the client flag exists.

**7. Collective probe's loopback port span.** Read `mlx._distributed_utils.launch`'s ring backend: `starting_port + i` is assigned to `hosts[i]` globally across the whole hostfile, not one shared port per host. So the peer (`hosts[1]`) binds a *different* port than the coordinator, and the old code only ever checked one port on the coordinator's own loopback — never the port the peer actually binds. Now sizes the span to `len(hosts)` and checks each remote host's assigned port over the same SSH identity every other Doctor probe uses (`_run_link_command`), retrying with a fresh span (up to 8 attempts) on collision, same trade-off `choose_fabric_subnet` already accepts (a port verified free moments ago beats one never checked).

All new tests verified to fail against the pre-fix code (temporary revert + rerun + restore) before landing. Full suite: 10,647 passed, 0 failed.

PR #2879, opened 2026-08-19
fix(admin): accept versioned DFlash2 names in draft-model picker

williamxie1989:
## Problem

DFlash 2 draft checkpoints (e.g. `z-lab/Qwen3.8-27B-DFlash2`) do not appear in the DFlash "Draft Model" dropdown of the model settings modal, even though DFlash 2 runtime support has been merged (#2840, #2850). Setting `dflash_draft_model` manually in `model_settings.json` works, which confirms the issue is purely in the picker.

## Root cause

Picker candidates are filtered client-side by `isDflashDraftModel()` in `omlx/admin/static/js/dashboard.js` (introduced in `b0e20904`). It tests the model id/name/path/repo/config type against:

```
/(^|[-_/\s])dflash($|[-_/\s])/i
```

The trailing boundary requires a non-alphanumeric character (or end of string) after `dflash`, so the versioned name `-DFlash2` (trailing digit `2`) never matches. The model then falls into the SpecPrefill bucket (`isSpecPrefillDraftModel` = "not DFlash and not VLM-MTP"). The DFlash 2 rollout updated the runtime and the UI wording ("compatible DFlash and DFlash2 checkpoints"), but not this name heuristic.

## Fix

- Allow an optional numeric suffix after the `dflash` token: `/(^|[-_/\s])dflash[0-9]*($|[-_/\s])/i` — matches `DFlash`, `DFlash2`, `DFlash2-mlx`, `DFlash-b16`, etc.; still rejects unrelated names.
- Update the helper text under the dropdown to mention the `-DFlash2` suffix and add a DFlash 2 example.
- Update the regression test in `tests/test_admin_dashboard_draft_filters.py` to pin the new pattern.

## Verification

- `node` check: `qwen3.8-27b-dflash2`, `z-lab/qwen3.8-27b-dflash2`, `qwen3.8-27b-dflash2-mlx` now match; `qwen3.8-27b-oq8e-mtp` and `qwen3.8-27b-dflashy` still do not.
- `pytest tests/test_admin_dashboard_draft_filters.py tests/test_admin_model_settings_template.py tests/test_admin_model_settings.py` — 26 passed.

### DIFF
--- a/omlx/admin/static/js/dashboard.js
+++ b/omlx/admin/static/js/dashboard.js
@@ -6950,7 +6950,9 @@
                 if (DFLASH_DRAFTER_CONFIG_MODEL_TYPES.has(configType)) {
                     return true;
                 }
-                return /(^|[-_/\s])dflash($|[-_/\s])/i.test(this.draftModelSearchText(model));
+                // DFlash 2 checkpoints are versioned ("-DFlash2"), so allow an
+                // optional numeric suffix after the "dflash" token.
+                return /(^|[-_/\s])dflash[0-9]*($|[-_/\s])/i.test(this.draftModelSearchText(model));
             },
 
             isVlmMtpDraftModel(model) {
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for fixing this. The numeric suffix handling makes DFlash2 checkpoints appear in the correct picker while preserving existing DFlash matches and excluding MTP or dflash-like names. I verified the focused admin tests on current main. LGTM, merging.

PR #2882, opened 2026-08-19
fix(api): strip Claude Code total_tokens budget markers from system text

### LINKED ISSUE #2881: Anthropic endpoint: Claude Code's <total_tokens> budget marker mutates the prompt head every request, defeating the prefix cache
### Summary

Claude Code (the Anthropic CLI agent) appends a token-budget marker to the **end of the system prompt on every request**:

```
<total_tokens>14999436 tokens left</total_tokens>
```

The value is freshly decremented each call, and the stale copies are kept, so the system prompt both **changes and grows (~13 tokens) on every request**. Because the mutation sits at the head of the prompt, no stored prefix can match past it — every turn of an agentic session degrades into a **full re-prefill**.

### Impact (measured)

Serving `minimax-m3-6bit` on an M3 Ultra (512 GB), driving it with Claude Code via `ANTHROPIC_BASE_URL`:

- Every request re-prefilled ~60k–106k of its tokens: `re-prefills 106291 of 110387 tokens (reused 4096)`
- Turn latency: **8–12 minutes** per turn at ~110k context (~250 tok/s prefill)
- The scheduler's divergence diagnostic shows the signature exactly — the divergence position creeps forward ~13 tokens per request (one marker appended per call):

```
prefix cache: request … shares the first 40381 of 67584 comparable tokens before diverging
prefix cache: request … shares the first 40394 of 69632 comparable tokens before diverging
prefix cache: request … shares the first 40407 of 70543 comparable tokens before diverging
```

- Debug-level decode of the divergence point:

```
first divergence at token 6110:
stored=...'total_tokens>15000000 tokens left</total_tokens>\n\n# Tools\nYou may call one or more tools to assist'
prompt=...'total_tokens>15000000 tokens left</total_tokens>\n\n<total_tokens>14999436 tokens left</total_tokens'
```

Model-independent: reproduced with two different models (MiniMax M3 and a large Qwen-family VLM) across separate sessions.

### Why oMLX should handle it

Claude Code offers **no toggle** to disable the marker (checked docs + env vars, 2026-08-19), and the marker is informational-only — nothing downstream parses it. oMLX already filters an equivalent client artifact for exactly this reason: the `x-anthropic-billing-header:` system blocks (`_BILLING_HEADER_PREFIX` in `omlx/api/anthropic_utils.py`, "contains randomly changing values, breaking prefix cache"). The budget marker is the same class of problem with a bigger blast radius.

### Fix

Strip `<total_tokens>N tokens left</total_tokens>` blocks from system content in the Anthropic adapter, mirroring the billing-header filter. With the fix deployed, turn-over-turn reuse on the same workload went from **4,096 → 112,640 of ~112.9k tokens**, and turn latency from minutes back to seconds.

PR incoming from my fork.

Pietroski:
Fixes #2881

### Problem

Claude Code appends a freshly decremented `<total_tokens>N tokens left</total_tokens>` block to the end of the system prompt on every request, keeping the stale copies. The prompt head therefore changes and grows (~13 tokens) each call, so no stored prefix can match past it and every turn of an agentic session degrades into a full re-prefill — measured at ~106k re-prefilled tokens per turn (8–12 min at ~110k context) before the fix, on both MiniMax M3 and a Qwen-family VLM. Claude Code exposes no toggle for the marker, and nothing downstream parses it. Full evidence (divergence-creep logs, debug decode of the divergence point) in #2881.

### Change

Strip the marker wholesale in the Anthropic adapter, mirroring the existing `x-anthropic-billing-header:` filter that exists for exactly this reason:

- `_TOTAL_TOKENS_MARKER_RE` + `_strip_client_budget_markers()` next to `_BILLING_HEADER_PREFIX`
- Applied in `_extract_system_text` (canonical `system` field, string and block-list forms)
- Applied to `system`/`developer`-role messages at the end of both `convert_anthropic_to_internal` and `convert_anthropic_to_internal_harmony`, covering clients that send system content inline in `messages[]` when consolidation is off

User/assistant content is deliberately untouched — quoting the marker in a conversation must survive round-tripping.

### Tests

`TestClientBudgetMarkerStripping` in `tests/test_anthropic_adapter.py` (5 cases): system string, system block list, marker-free passthrough (byte-identical), inline system messages with `consolidate_system_messages=False`, and user-quoted marker preservation. Full `test_anthropic_adapter.py` suite passes (43/43).

### Result

On the originating workload, turn-over-turn cache reuse went from `reused 4096` to `112640 of ~112.9k tokens`, turn latency from minutes to seconds.

---

Co-authored by: Claude Fable 5

### DIFF
--- a/omlx/api/anthropic_utils.py
+++ b/omlx/api/anthropic_utils.py
@@ -8,6 +8,7 @@ Handles conversion between Anthropic API format and internal oMLX format.
 import base64
 import json
 import logging
+import re
 import uuid
 from typing import Any
 
@@ -444,6 +445,16 @@ def convert_anthropic_to_internal(
             # Unknown format
             processed_messages.append({"role": role, "content": str(content)})
 
+    # Claude Code 2.1.154+ may carry system content inline in messages[]
+    # (see _normalize_in_messages_system); with consolidation off those
+    # blocks bypass _extract_system_text, so the budget markers are
+    # stripped here too before role merging.
+    for msg in processed_messages:
+        if msg.get("role") in ("system", "developer") and isinstance(
+            msg.get("content"), str
+        ):
+            msg["content"] = _strip_client_budget_markers(msg["content"])
+
     from .utils import _merge_consecutive_roles
 
     return _merge_consecutive_roles(processed_messages)
@@ -656,6 +667,16 @@ def convert_anthropic_to_internal_harmony(
             # Unknown format
             processed_messages.append({"role": role, "content": str(content)})
 
+    # Claude Code 2.1.154+ may carry system content inline in messages[]
+    # (see _normalize_in_messages_system); with consolidation off those
+    # blocks bypass _extract_system_text, so the budget markers are
+    # stripped here too before role merging.
+    for msg in processed_messages:
+        if msg.get("role") in ("system", "developer") and isinstance(
+            msg.get("content"), str
+        ):
+            msg["content"] = _strip_client_budget_markers(msg["content"])
+
     from .utils import _merge_consecutive_roles
 
     return _merge_consecutive_roles(processed_messages)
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for fixing this. The system-only filtering covers both canonical and inline system paths while preserving quoted user content, and I verified that accumulated markers produce identical prompts on the standard and Harmony paths. The focused adapter/API tests also pass against current main. LGTM; this is safe to merge.

PR #2883, opened 2026-08-19
fix(cluster): stop the planner from re-scanning every model on each autoconfigure tick

alytaphoenix:
## Summary
- The admin dashboard's cluster tab polls `/admin/api/cluster/autoconfigure` every ~10s while open. Each tick called `complete_model_layout()` for every locally discovered model — including ones with no bearing on the active deployment — and that function unconditionally re-opened and re-parsed every safetensors shard header on disk, plus re-ran `maybe_apply_pre_load_patches()` against every model's `config.json`.
- Root-caused two live incidents today to this: the server log showed `nemotron_h MTP chain patch applied` firing every ~15s for 11+ minutes straight while a 27B model was actively serving on a 2-node cluster, then the server died mid-unload and left an orphaned rank process pegged at 100% CPU that only responded to `SIGKILL` — the same wedged-GPU failure class as two prior hard machine hangs today. Repeatedly re-inspecting an unrelated 19GB Nemotron checkpoint on every tick is a plausible resource-contention trigger.
- Two independent idempotency gaps, both fixed:
  - `complete_model_layout()` now caches `ModelLayout` per resolved model path, keyed by `(model dir mtime, config.json mtime)`, so repeat calls for an unchanged model skip the disk re-scan entirely.
  - `nemotron_h_chain.apply()` re-ran and logged "applied" on every call even though each of its six sub-patches already no-ops once installed (matching the pattern every sibling sub-patch, and `qwen35_model.py`, already follow). Added the same top-level guard so a fully-patched module is a silent no-op, matching the function's own "safe to call repeatedly; idempotent" docstring claim.

## Test plan
- [x] Standalone verification script: `nemotron_h_chain.apply()` now logs "applied" once across 3 calls (was 3/3 before the fix); `complete_model_layout()` calls `inspect_safetensors_layout()` once across 3 identical calls, and correctly re-invalidates when `config.json`'s mtime changes.
- [x] Confirmed the Nemotron TP=2 fix (e18508ec) is untouched — `tests/test_cluster_planner.py` (23) and `tests/test_cluster_tensor_strategies.py` (7) all pass, 30/30.
- [x] `python3 -c "import ast; ast.parse(...)"` syntax check on both changed files.

### DIFF
--- a/omlx/cluster/planner.py
+++ b/omlx/cluster/planner.py
@@ -8,6 +8,7 @@ import json
 import re
 import struct
 import subprocess
+import threading
 from collections.abc import Iterator, Sequence
 from contextlib import contextmanager
 from dataclasses import dataclass, replace
@@ -993,6 +994,10 @@ def inspect_safetensors_layout(model_path: str | Path) -> ModelLayout:
     )
 
 
+_LAYOUT_CACHE: dict[str, tuple[tuple[float, float], ModelLayout]] = {}
+_LAYOUT_CACHE_LOCK = threading.Lock()
+
+
 def complete_model_layout(model_path: str | Path) -> ModelLayout:
     """Layout of a model this node holds in full, refusing one it holds part of.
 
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, good catch and a clean fix. I verified the chain-init flag is set by the last sub-patch, so the top-level guard is safe, and the cache key and locking look right. Merging. One edge I'll fold into a follow-up: an in-place shard overwrite doesn't bump the directory mtime, so I'll add the shard mtimes to the fingerprint.

PR #2884, opened 2026-08-19
fix(admin): clear default_model pointer when is_default is unset

alytaphoenix:
## Summary
- `is_pinned` already synced both `current_settings.is_pinned` and the engine pool entry's `is_pinned` bidirectionally, client and server side.
- `is_default` only handled the "set to `True`" case: the server never cleared `server_state.default_model` when `is_default` was unset, and `dashboard.js`'s `updateModelSetting()` only updated the local model list when `value === true`, silently no-op'ing on unset.
- Net effect: the GUI kept showing a model as default (and the server kept routing default requests to it) after a user unpinned it, until the next server restart re-derived the pointer from persisted settings — the GUI bug reported today ("model gui bugs that do not reflect the pin state correctly").

## Test plan
- [x] Standalone FastAPI `TestClient` script replicating the route: setting the pointer, unsetting the current default (clears the pointer), and unsetting a model that was never the pointer target (must leave someone else's default alone) — all 3 pass.
- [x] Added `TestDefaultModelPointer` covering the same 3 scenarios to `tests/test_admin_profiles_api.py`.
- [x] `node --check` on `dashboard.js`, `ast.parse` on `routes.py`.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -2616,9 +2616,15 @@ async def update_model_settings(
         entry.is_pinned = request.is_pinned
     if request.is_default is not None:
         current_settings.is_default = request.is_default
-        # Update server_state.default_model if setting as default
-        if request.is_default and server_state:
-            server_state.default_model = model_id
+        if server_state:
+            if request.is_default:
+                server_state.default_model = model_id
+            elif server_state.default_model == model_id:
+                # Unsetting the model that IS the current default must clear
+                # the pointer, not just this model's own flag -- otherwise the
+                # server keeps treating an unpinned model as default until the
+                # next restart re-derives it from persisted settings.
+                server_state.default_model = None
     if request.is_hidden is not None:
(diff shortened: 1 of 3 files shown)


jundot:
Thanks, clean fix. I verified the pointer is live-consumed (model-less requests and model_fallback both route through server_state.default_model), so the stale pointer really did keep serving an unset default until restart, and the only-when-pointing-at-this-model guard is exactly right. Merging. One adjacent gap I'll take in a follow-up: setting a new default doesn't clear the previous model's persisted is_default flag, so two models can persist True and the restart derivation picks whichever comes first, the same one-direction issue you fixed here but on the persistence side.

jundot:
Correction to my note above: the persistence side is already handled. set_settings clears is_default from every other model whenever one is set (model_settings.py, pinned by test_exclusive_default), so no follow-up is needed there. The live pointer this PR fixed was the only one-directional half.

PR #2885, opened 2026-08-19
feat(cluster): drift watchdog + dead-port retirement via advertised admin port (fabric-doctor C5)

alytaphoenix:
## Summary
Closes failure #7 and the reboot-drift tail of #4/#5 from the Fabric Doctor design doc. Depends on C2 (#2875, `feat/cluster-fabric-intent`) — this branch is based on it and will shrink to just the C5 diff once #2875 merges.

- `ClusterStatus` gains `admin_port: int = 0`, populated by `collect_cluster_status()` from `get_settings().server.port` (0 on a worker-only install where settings are unavailable).
- `probe_remote_admission_ceiling()` now targets the peer's advertised `admin_port` when known instead of guessing conventional ports, and confesses honestly instead of failing silently: returns `AdmissionCeilingProbe(ceiling_bytes, fast_probe_ok, fast_probe_error)` instead of a bare `int`. A dead advertised port produces `fast_probe_ok=false` with the slow in-process ceiling still returned, not an exception.
- `/node-budgets` threads the freshest advertised `admin_port` (learned from any capability probe that flows through `routes.py`) into the ceiling probe, and records one deduped WARN incident per dead port rather than failing the poll.
- New `fabric_intent.detect_drift(intent, live_interfaces, collides=...)`: a pure comparison of live fabric addressing against the recorded intent. A collision with something else on the fabric always wins (`intent_collides`) even when addresses are also missing/changed. Non-collision drift is auto-restorable only when the intent was applied via `networksetup` (the service config persists, so `configure_link`'s tier 2 re-assert needs no new consent) — every other case surfaces as one deduped WARN inviting a consented Fabric Doctor re-address, never a silent privileged action. Wired into `cluster_autoconfigure` via `_resolve_fabric`'s `fabric["drift"]`, deduped per `(subnet, kind)` like the C4 VPN pre-warning.
- Deliberately poll-driven — no event-driven wake/network-change watcher exists in the repo or is added here.

## Test plan
- [x] `tests/test_cluster_fabric_intent.py`: 33/33 pass (drift matrix: live==intent, address lost, address changed, collision).
- [x] `tests/test_cluster_probe.py`: 20/20 pass (`admin_port` advertised, settings-unavailable → 0).
- [x] `tests/test_cluster_launch.py`: 49/50 pass — the 1 failure (`test_peer_probe_discovers_a_different_linux_python_path`) is pre-existing, untouched by this diff, and reproduces identically on `origin/main`.
- [x] `tests/test_cluster_routes.py`: 61/64 pass — the same 3 failures reproduce identically with this diff stashed out, confirmed unrelated to C5.
- [x] Traced the `routes.py` probe-result field assumptions (`status`/`node`/`ssh`) against `probe_remote_host`'s actual return shape.
- [x] `ast.parse` syntax check on every changed file.


jundot:
I'm holding review on this one until the C2 rework settles, it inherits that base and its auto-restore branch keys on the networksetup path that currently can't run. One thing I want to keep from it: advertising admin_port in cluster status and probing that instead of guessing conventional ports is the right call, and I'd like it to become the standard way peers find each other's ports during the rework.

alytaphoenix:
C2 (#2875) has settled -- merged the updated branch in (one conflict in the trailing test file, from the now-removed service-addressing tests overlapping the new drift-watchdog tests at the same location; resolved by dropping the removed ones and keeping the drift tests). Diffed against `fork/feat/cluster-fabric-intent` and this branch is now genuinely just the C5-specific changes, as the PR description said it would shrink to.

On the auto-restore branch: rather than fix it in place, C2's rework split the networksetup addressing path out entirely (its own docstring said `-setmanual` isn't yet live-validated, and I didn't want to attempt that validation without your rig). So `DriftFinding`'s `addressing=="networksetup"` branch is dormant for every intent this build can record now -- every intent is `"ifconfig"`. I documented that explicitly in `DriftFinding`'s docstring rather than removing the branch, since the store already round-trips both values and it's exactly the behavior the networksetup follow-up will need the moment it ships (it's also already covered by its own test).

Worth noting: `routes.py`'s own `_detect_fabric_drift` collision filter already used `subnet_of` (not equality, not plain overlap) for its own-network exclusion, independently of `transport.py`. That's the same comparison direction I converged on fixing #2875's `own_networks` to use (plain overlap over-excludes: a hostile network broader than "own," like a utun routing all of 10/8, would get blanket-excluded from the collision set just because it covers this link's own address, silently dropping a real collision risk for every other candidate in that range). Good sign the two call sites already agree.

`admin_port` advertising/probing is untouched -- kept exactly as you asked. Full 10,590-test suite passes.

alytaphoenix:
One more cascade update: merged the just-updated #2875 (which itself now includes the #2867 fix -- subnet-default un-bundled, last-resort fallback, WARP fixture fix) in. One conflict, in the trailing test file (this branch's own drift-watchdog test suite vs. #2875's corrected default-address assertion), resolved by keeping the drift tests and taking the corrected assertion. Full 10,592-test suite passes.

PR #2887, opened 2026-08-19
feat(cluster): Start Cluster as a persisted server-owned job (fabric-doctor B2)

alytaphoenix:
## Summary
Closes failure #8's dead-button class. Depends on B1's incident store (#2866) and stacks on top of C1/C2/C3/C4/C5 already open on this branch — the diff will shrink to just B2 as those merge.

Clicking Start used to drive a purely client-side ladder (link setup → autoconfigure → staging → activation) whose state lived in the browser tab; a reload lost all of it and left a dead button. The ladder now runs on the server as one persisted job record the dashboard merely renders.

- New `omlx/cluster/start_job.py`: `StartJobPhase`, `StartJobStore` (in-memory dict + lock, mirrors the existing `_STAGING_JOBS` pattern — the job must survive the browser, not the server; durable history is the B1 incident feed and the deployment registry), and `run_start_job()` running exactly the client's old ladder with every phase transition updating the store and every failure recording an Incident.
- `routes.py`: `cluster_autoconfigure` and `activate_cluster_deployment` refactored into thin wrappers over new `_autoconfigure()`/`_activate()` callables (behavior-preserving extraction, bodies moved verbatim) so the job runner executes the identical code path the sync endpoints serve. New `POST`/`GET /start-jobs` endpoints. A `threading.Lock` shared between the legacy sync path and the async job path (never held across an await) 409s either side against the other, plus 409 on a concurrent same-model job. New-attempt supersession links a prior FAILED job's `superseded_by` and records it in the incident feed.
- `dashboard.js`: `startCluster()` now POSTs the job and renders the server's record; the existing 750ms activation timer's `refresh` closure branches to poll the job when one is running, otherwise the manual-plan path (`activateClusterProposal`) is untouched. Cluster-tab init re-attaches to any non-terminal job. `clusterActivationLoading` derives from phase, preserving the [#2819] poll guard.
- `_cluster.html`: Start button label renders the server phase text.

## Test plan
- [x] Independently re-ran with real pytest (not just the implementation's self-report): `tests/test_cluster_start_job.py`, `test_cluster_routes.py`, `test_cluster_one_click.py` — 133/133 pass.
- [x] Full `pytest tests/ -k cluster` sweep: 1181 passed, 6 failed — all 6 reproduce identically without this change (pre-existing SSH-target-resolution and python-discovery environment failures, confirmed by stashing).
- [x] A repo-wide dead-code lint (`test_cluster_seams.py`) flagged the store's test-only reset hook as unreachable from production code; added it to the existing `allowed_uncalled` list with a reason, matching the file's established convention.
- [x] Traced the lock ordering for deadlock safety: `_ACTIVATION_GUARD_LOCK` always outer, `StartJobStore`'s internal lock always inner, no exceptions.
- [x] Verified the endpoint refactor is byte-identical below the new wrapper lines (no logic changes, pure extraction).
- [x] `ast.parse` + `node --check` syntax checks.


jundot:
Quick roadmap for the whole fabric doctor series, posting here since this is the newest layer. Merged so far: #2819, #2844, #2883, and I reparented #2869 and #2866 onto main and merged them, so those two branches were force-pushed and a rebase onto current main drops the landed commits automatically. #2848 is closed, the ladder is worth having but I'd rather take it fully wired together with the doctor rework. The rest (#2849, #2867, #2870, #2875, #2878, #2885, and this one) needs rework and I'll leave detailed notes on each PR shortly. The key finding to internalize first: #2849's bound-TCP probe never survives ssh (the remote command gets shell-mangled into a syntax error), so every check layered above it is being evaluated against a probe that always fails, and the stack has to be rebuilt from there. Please hold off on new stacked PRs until that base is settled. Review capacity is the bottleneck here, and every new layer on the current base means rebuild work for both of us. I'm glad to take the series again once it's re-stacked on main with the #2849 fix first.

PR #2889, opened 2026-08-19
Add Q5 ANE prefill support

onthehub97:
Split-off from #2826: This PR adds support for Q5 to the ANE prefill kernel (see original PR for more details)


jundot:
Nicely scoped split, thanks. I ran the same layer level parity harness I used for #2833 on this branch (M3 Ultra, rebuilt kernels): q5 gate/up at gs 64 and 128, single and dual ANE, all compiled and executed with cosine >= 0.999979 against stock MLX, matching the q4/q6/q8 numbers. Merging.

Issue #2880, opened 2026-08-19
qwen3_5 4096 prefill floor (#2751) never delivers with prefix caching on; forcing it regresses M5 Pro


beaglemoo:
## Summary

The 4096-token qwen3_5 prefill floor added in #2751 for >= 64 GB machines
never takes effect when the prefix cache is enabled: the ArraysCache block
boundary (`_ARRAYS_CACHE_BLOCK_SIZE = 2048`, from #2040) cuts every prefill
chunk at the next 2048-token block edge, so delivered chunks are always 2048
regardless of the floor, the configured step size, or memory headroom.
Separately, when I patched the block size so 4096 chunks actually deliver,
TTFT on an M5 Pro got worse — so the two facts deserve different fixes on
different hardware.

## Mechanism

`Scheduler.__init__` sets `_qwen35_prefill_floor = 4096` on >= 64 GB machines
(scheduler.py, from #2751), and `_prefill_step_size_for_progress` duly
requests 4096. But for GDN hybrid models,
`_enlarge_block_size_for_arrays_cache` sets
`config.paged_cache_block_size = _ARRAYS_CACHE_BLOCK_SIZE = 2048` ("Match
prefill_step_size (2048)..." — a comment written before the floor existed),
and the boundary-limited step in the prefill loop clamps every chunk to the
next block boundary:

```
next_boundary = ((current_total // block_size) + 1) * block_size
```

Observed on a 16K traced benchmark (v0.6.2-era main, M5 Pro 64 GB, memory
guard cap raised to 52.2 GB so memory is not a factor):

```
chunk_tokens=2048 processed=0->2048  requested_step=4096
chunk_tokens=2048 processed=2048->4096 requested_step=4096
... (every chunk)
```

With server logging at DEBUG, neither `_guard_prefill_chunk` nor the adaptive
throttle logs any shrink — the cut is purely the block boundary. So the
floor is dead code whenever `_prompt_cache_needs_snapshots` is true (any GDN
model with caching on), on every machine including the M3 Ultra it was tuned
on. I assume the #2751 measurement (+3.2% at 4k / +1.0% at 16k for 4096 vs
2048) was taken with the prefix cache disabled or via a path without
boundary snapshots.

## Forcing 4096 delivery regresses M5 Pro

I set `_ARRAYS_CACHE_BLOCK_SIZE = 4096` locally, verified via traced
benchmark that all chunks then deliver at 4096 through a full 16K prefill
[...]

PR #2890, opened 2026-08-19
Add wide GPU prefill blocks, internal ANE tiling, and benchmark fidelity and instrumentation

onthehub97:
This PR splits some plumbing changes from #2853

### DIFF
--- a/apps/omlx-mac/Sources/AppView/Screens/ThroughputBenchScreen.swift
+++ b/apps/omlx-mac/Sources/AppView/Screens/ThroughputBenchScreen.swift
@@ -64,6 +64,7 @@ struct ThroughputBenchScreen: View {
                 selectedModelId: $vm.selectedModelId,
                 contextProfile: $vm.contextProfile,
                 warmupMode: $vm.warmupMode,
+                alignPromptToAne: $vm.alignPromptToAne,
                 promptLengths: $vm.promptLengths,
                 genLength: $vm.genLength,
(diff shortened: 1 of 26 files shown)


jundot:
Thanks, and ebccd922 already covered the locale files. I verified the tiling on hardware with the same layer level parity harness from #2833 (M3 Ultra, compiled shape 2048): inputs of 4096 (two tiles) and 4608 (two tiles plus a 512 row tail) through _backend and _gdn_backend match stock MLX at the same error scale as the exact path, for MLP and GDN, single and dual ANE. Merging, with follow ups on my side.

The main one: as I read it, this inverts the sequence_length contract. It used to pin the scheduler chunk to the ANE shape, now it means "at most the delivered chunk width, wider chunks tile down onto it". With that reading a sequence_length of 4096 on a sub 64 GB machine, or on any machine with prefix caching on (the 2048 ArraysCache boundary from #2880), compiles everything and then never dispatches a single ANE op, with only an info line. I'll add a loud warning for that case rather than let it sit silent, and update the docs and the tuner hint to the new contract. Tell me if I read the intent wrong.

Other follow ups I'll push: route the wide tile tails through the native qmm path instead of stock MLX (matters once tails can be 2048+ rows), and gate the new per chunk trace lists behind benchmark_trace. The 2048 to 4096 chunk width change for ANE users on 64 GB+ machines with caching off goes in the release notes.

PR #2891, opened 2026-08-19
Qwen35: ANE Tuner V2

onthehub97:
This splits the new heuristic approach ANE tuner from #2853


jundot:
The compile count reduction is a clear win, 11 eager full model compiles down to 3, and the idle ANE guard and measurement fixes all survived the rewrite. Merging, with follow ups on my side.

The big one: the profile refinement reads gpu_completion_ns, but that profiler key only exists on the #2853 branch, the split left it behind, so on this stack gpu_time is always 0 and the refine stage never fires (the test builds the missing key by hand, which is why CI stays green). I'll switch it to gpu_qmm_ns and add a schema parity test against fast._ANE_PROFILE_KEYS; the richer completion metric can come back together with the CPU PR.

Also on my side: calibration currently runs on the event loop thread, so results polling, cancel, and the health check stall during the bank compile. I'll move it onto the MLX executor like the other model touching paths. Plus a tailwind rebuild for the new results grid, keeping the GPU only recommendation when a later measurement fails instead of discarding the whole run, and the unconditional ANE field validation in the settings save.

One design question rather than a fix: calibration times a single layer with no attention or tail competing for the GPU, which biases toward smaller ANE fractions than real traffic would pick. Once the refinement actually fires, do you expect it to recover that, or is it worth keeping two or three end to end grid points as a cross check?

Issue #2841, opened 2026-08-19
ANE prefill triggers hard memory-pressure abort + model eviction (Qwen3.8-27B)


quochuy:
Loading Qwen3.8-27B-oQ4e-fp16-mtp with ANE prefill enabled works, but the first request with a long prompt (17,330 tokens, cache miss) causes the process memory enforcer to abort the request and evict the model: usage jumps to 29.4 GB, above the hard watermark of 28.5 GB (ceiling 30.0 GB, iogpu sysctl cap 30.4 GB). Disabling ANE prefill avoids the abort entirely.

## To Reproduce
1. Load model Qwen3.8-27B-oQ4e-fp16-mtp using profile qwen38_27b (qwen35_ane_prefill_enabled: true, fraction 0.53, max_layers 64, seq len 2048, dual_ane: false; GDN input projections on 48 layers, fraction 0.5). During load, ANE prefill eagerly compiles 64 private ANE gate+up slice models (5120→18432).
2. Send a request with a long prompt (e.g. 17,330 prompt tokens, cache miss).
3. Observe memory pressure ok -> hard (~11s after request add) and the abort below.

omlx.process_memory_enforcer - WARNING - Hard memory pressure: aborted 1 request(s) on
'Qwen3.8-27B-oQ4e-fp16-mtp' and kept model loaded
omlx.server - ERROR - Error during chat streaming: Request aborted: process memory limit
exceeded (usage 29.4 GB, abort threshold (hard watermark) 28.5 GB, static/dynamic ceiling
30.0 GB). Raise custom_ceiling_bytes in admin Memory settings (currently pinned at 30.00 GB),
or reduce context length.
omlx.process_memory_enforcer - WARNING - Evicting model 'Qwen3.8-27B-oQ4e-fp16-mtp' (pressure=hard)

4. Request is requeued (prefill retry 1/2) and the abort/evict loop repeats on reload.

## Expected behavior

ANE prefill transient memory (private slice models + prefill scratch) should be accounted for in the prefill preflight reservation, so the request prefill completes within the configured ceiling/headroom instead of tripping the hard watermark and aborting.

## Desktop:
- macOS Version: 26.5.2 (Build 25F84)
- oMLX Version: 0.6.2
- Hardware: MacBook Pro 14" (Mac14,5), Apple M2 Max, 32 GB unified memory
[...]

PR #2892, opened 2026-08-19
Qwen35 ANE: CPU Sharing

onthehub97:
This pulls out the CPU-sharing / AMX related changes from #2853


jundot:
Thanks for carrying the split all the way through, the first three landed cleanly. Before I do the full review pass on this one, here are four things I already know from reviewing the #2853 branch, and I re-checked that they all carry over into b036a1b1. These are the blockers; the rest can be follow ups on my side like the other PRs.

1. The shipped kernel still calls the private __bsdthread_ctl directly, while your own thread probe already resolves the public dispatch_apply_with_attr family that expresses the same shared resource policy and owns the set/clear lifecycle. The raw syscall buys nothing over the public API and costs forward compatibility, and the current capability probe can leak the policy onto a pooled thread when set succeeds and clear fails (the RAII destructor also discards the clear result). Please rewrite the scheduling mode on dispatch_apply_with_attr.

2. The eager FP16 rows and the down projection share are invisible to the memory guard: engine_pool only adds the new settings to the reload signature, so estimated_size and the prefill preflight never see the extra bytes. Your own #2853 numbers had the down share taking peak RSS from 31 to 50 GiB on the 27B, which is exactly the #2841 failure class, and on a 64 GB machine the enforcer would abort with the guard believing the model is ~17 GB. The projected CPU share bytes need to feed estimated_size, or the setting needs a preflight gate.

3. The workspace vector inside the dispatch_apply block can throw std::bad_alloc, which unwinds into libdispatch's C frames and terminates the whole server process. A try/catch around the block body that stores the failure flag, like the BNNS failure path already does, covers it.

4. clone_mlx_model_fp16.py converts bf16 to fp16 with no range check, so any weight or scale outside fp16 range becomes inf silently, in a checkpoint the user will then serve. An inf/NaN scan with a report is needed before this tool ships.

Smaller thing while you are in there: benchmarks/qwen35_bnns_thread_probe.mm is not referenced by any build and carries a second, divergent copy of the syscall wrapper; either wire it into a dev target or drop it.

joshuaswarren:
Ran this branch (`e961093`) on an M1 Ultra, 128 GB, specifically against the memory-accounting blocker in the review above.

**Tests:** `227 passed` — `tests/test_engine_pool.py`, `tests/test_qwen35_ane_prefill.py`, `tests/test_clone_mlx_model_fp16.py`, `tests/test_ane_tuning.py`, in a clean venv (`uv pip install -e .`, Python 3.13).

**The estimator does now project the CPU-share bytes.** Calling `_qwen35_cpu_share_estimated_bytes` directly against a 27B mixed-precision 4/5-bit checkpoint:

| settings | projected extra |
|---|---|
| CPU sharing off | 0.00 GiB |
| default (`cpu_fraction` 0.135) | 4.22 GiB |
| max legal `cpu_fraction` 0.25 | 7.97 GiB |
| `cpu` .25 + `down` .50 | 20.17 GiB |
| `cpu` .25 + `down` .50 + `gdn` .50 | 25.80 GiB |

That range brackets the 31 -> 50 GiB peak RSS reported in the review, so the magnitude looks right rather than merely non-zero.

**End-to-end under a 64 GB guard.** Served the same checkpoint with `--memory-guard-gb 64` and the max-legal CPU-share settings:

```
engine_pool - INFO - Qwen CPU sharing adds 25.80GB to the projected memory for <model>
engine_pool - INFO - Loaded model: <model> (actual: 16.21GB, local estimate: 42.44GB, full model: 16.65GB, total: 42.44GB)
```

Admission control sees **42.44 GB**, not the ~17 GB the review said it would. On the blocker as written, this head reads as addressed.

**Two limits on what that proves.**

1. The ANE path did not engage on this run — `patches.qwen35_ane_prefill - WARNING - Private ANE runtime unavailable; Qwen ANE prefill skipped` — because I installed the package without building the custom kernels. So this validates the **projection**, not the materialization peak or the prefill throughput. Somebody with the kernels built should confirm actual RSS tracks the new estimate.
2. I first read `down_fraction = 1.0` collapsing the down-share term back to the gate/up-only figure as a boundary bug. It is not reachable: the admin API rejects with 400 and the runtime raises `ValueError` outside `cpu <= 0.25`, `down <= 0.50`, `gdn <= 0.50`. Worth a comment on the `down_rows < hidden` branch so the next reader does not file it either.

**Offer:** the failure class cited in the review (hard memory-pressure abort on a 64 GB machine, #2841) needs a single-ANE-die 64 GB part, and the backend gates the CPU path on `dual_ane`, so a real 64 GB M1 Max cannot run it at all — only a multi-die part can, with the ceiling simulated. I have a 64 GB M1 Max, a 96 GB M2 Max, and the 128 GB Ultra above, and can run whichever combination is useful once this is marked ready for review.

onthehub97:
@joshuaswarren that estimate is in addition to the weights of loading the model. The CPU path is a hungry beast, it can be disabled in tuner settings though. 

Looking into the dual ane points though.  It shouldn’t matter as cpu isn’t related to the ane but I’ll double check.

joshuaswarren:
Agreed on additive, and that is what the log shows: `full model: 16.65GB` + `25.80GB` CPU share = `total: 42.44GB`. I read it the same way you do. Worth saying for anyone skimming the table: **25.80 GiB is the max-legal corner** (`cpu` .25 / `down` .50 / `gdn` .50), not the cost of turning the feature on. At the default `cpu_fraction` 0.135 the projection is **4.22 GiB**, which is a much easier sell.

On `dual_ane` — you are right that CPU work and ANE work are unrelated in hardware, but **the diff couples them**, so it is not just my reading. In `omlx/patches/qwen35_ane_prefill.py` all six preparation sites gate the CPU path on it:

```python
cpu_enabled = bool(
    config.cpu_fraction > 0
    and dual_ane
    and gate.scales.dtype == mx.float16
    and up.scales.dtype == mx.float16
    and fast.has_symbol(_cpu_gate_kernel_symbol(bits))
)
cpu_outputs = (
    (int(output_dim * config.cpu_fraction) // 64) * 64 if cpu_enabled else 0
)
```

and the settings modal matches it: `:disabled="!modelSettings.qwen35_ane_prefill_dual_ane"`.

So on a single-ANE-die part `cpu_outputs` is 0 and the CPU share never materializes, whatever the tuner says. That has a direct bearing on the memory blocker: the 64 GB machines that would abort are single-die, so **the abort class cannot be reproduced on the hardware that is most at risk** — it can only be exercised on a multi-die part with the ceiling simulated, which is what I did above. If the coupling is deliberate (FP16 clone plus the 128-vs-64 alignment path), then the memory concern narrows to multi-die hosts and the machines with the least headroom are simply excluded. If it is incidental, decoupling it would widen who can test this for you. Either way it is worth a line in the docs, because the current UI reads as "you need two ANE dies to share work with the CPU".

Happy to re-run any combination once you have looked. I did not build the custom kernels for that run, so `fast.has_symbol(...)` would have been false regardless — if you want a real materialization measurement rather than a projection, tell me the build flags you use and I will rebuild and report actual RSS against the estimate.

onthehub97:
@joshuaswarren I've pushed a change that fixes the dual ANE tuner requirement but the single ANE path does not yet support CPU sharing. I'm implementing that now, shouldn't be long.

onthehub97:
@joshuaswarren Patch is live, feel free to test

joshuaswarren:
Tested `1e5345b`. The gate fix is confirmed and the accounting still looks right in the new mode.

**Tests:** `245 passed` (up from 227 on `e961093`) — same four files, clean venv, Python 3.13, M1 Ultra.

**`dual_ane` is gone from the CPU gate in all six preparation sites**, and the symbol is now selected per mode:

```python
cpu_enabled = bool(
    config.cpu_fraction > 0
    and gate.scales.dtype == mx.float16
    and up.scales.dtype == mx.float16
    and fast.has_symbol(_cpu_gate_kernel_symbol(bits, dual=dual_ane))
)
```

**The memory estimate is identical for single and dual ANE**, and I went looking for whether that is a blind spot rather than assuming it:

| settings | dual_ane=True | dual_ane=False |
|---|---:|---:|
| default `cpu_fraction` 0.135 | 4.22 GiB | 4.22 GiB |
| `cpu` .25 | 7.97 GiB | 7.97 GiB |
| `cpu` .25 + `down` .50 | 20.17 GiB | 20.17 GiB |
| `cpu` .25 + `down` .50 + `gdn` .50 | 25.80 GiB | 25.80 GiB |

That is **correct, not a miss**: every runtime `cpu_outputs` site computes `(int(output_dim * fraction) // 64) * 64`, and `alignment = 128 if dual_ane else 64` applies only to the ANE rows — so the CPU row count genuinely does not depend on the mode, which is what `_qwen35_cpu_share_estimated_bytes` assumes by never referencing `dual_ane`. Worth knowing that the two agree by construction rather than by luck.

**What I cannot tell you yet:** actual materialization. `fast.has_symbol(...)` is False for all four symbols in my venv (`qwen35_ane_cpu_fp16_q4_swiglu_t`, `qwen35_ane_dual_cpu_fp16_q4_swiglu_t`, and the 8-bit pair), so `cpu_enabled` is False and no CPU rows are ever built — the projection is exercised, the peak is not. Building with `OMLX_WITH_CUSTOM_KERNEL=1` fails here with `error: cannot execute tool 'metal' due to missing Metal`, i.e. the Xcode Metal toolchain component is not installed on this box. I am installing it and will report real RSS against the estimate.

Two small things while I have the code open:

1. A note on the `down_rows < hidden` branch in the estimator would save the next reader the trip I made — it looks like a boundary bug until you find the 400/`ValueError` clamps.
2. If a `has_symbol` miss is expected on stock installs, consider logging once at debug when `cpu_fraction > 0` but the symbol is absent. Right now the feature silently does nothing, which is indistinguishable from it working on a machine where nobody checks the row counts.

**The hardware offer stands and is now actually reachable:** with the `dual_ane` requirement gone, my 64 GB **M1 Max** — single ANE die — can finally take this path, which is the exact shape of the #2841 abort class. Say the word once you are happy with the kernels and I will run the real memory test on it rather than a simulated ceiling.

joshuaswarren:
Built the custom kernels and ran the real thing on `1e5345b`. Single-ANE CPU sharing **works**, the memory blocker is **fixed end to end**, and I hit one hard crash that I could not reproduce — details below, because the frames point somewhere specific.

## The memory blocker is closed

With kernels built, `has_symbol` is true for all four symbols including the new single-ANE pair, so the path actually materializes rather than silently no-opping. Single-ANE mode (`dual_ane: false`), FP16 clone, `cpu` .25 / `down` .50 / `gdn` .50:

```
qwen35_ane_prefill - Qwen ANE CPU sharing using performance-aware scheduling with 8 workers
qwen35_ane_prefill - Eagerly compiled and enabled ANE/GPU Qwen prefill on 64 MLPs (sequence_length=2048, ANE fraction=0.530, dual_ane=False)
engine_pool - Loaded model: <model> (actual: 30.84GB, local estimate: 42.44GB, full model: 16.65GB, total: 42.44GB)
```

So the CPU share really does cost memory — **30.84 GB actual against a 16.65 GB model**, about +14.2 GB at those settings — and the projection reserved 25.80 GB for it. The estimate is therefore **conservative by roughly 11.6 GB**, which is the safe direction, though it may exclude machines that would have fit.

And the protection works. Same model, ceiling dropped to 40 GB:

```
507: Model '<model>' (42.44GB) does not fit under the dynamic memory ceiling (40.00GB).
     Raise custom_ceiling_bytes in admin Memory settings (currently pinned at 40.00GB), or use a smaller model.
```

That is the behaviour the review asked for: it refuses with an actionable message instead of admitting a model it thinks is ~17 GB and letting the machine abort.

## Mode-independent estimate is correct, not a gap

The estimate is byte-identical for `dual_ane` true and false at every setting I tried (4.22 / 7.97 / 20.17 / 25.80 GiB). That is right rather than lucky: every runtime `cpu_outputs` site computes `(int(output_dim * fraction) // 64) * 64`, and `alignment = 128 if dual_ane else 64` only ever applies to the ANE rows, so the CPU row count genuinely does not vary by mode — which is the assumption `_qwen35_cpu_share_estimated_bytes` bakes in by never referencing `dual_ane`.

Tests: **245 passed** (up from 227), same four files.

## One hard crash, 1 in 9 runs

The first run with `cpu` .25 / `down` .50 / `gdn` .50 died five seconds after the model loaded, taking the whole server with it — `EXC_BREAKPOINT` / `SIGTRAP`, and libdispatch names it:

```
asi: libdispatch.dylib: "BUG IN CLIENT OF LIBDISPATCH: Unknown entity", "Abort Cause 8"

libdispatch.dylib   dispatch_apply_attr_set_parallelism.cold.5
libdispatch.dylib   dispatch_apply_attr_set_parallelism
libomlx_qwen35_prefill_kernel_ops.dylib  omlx::qwen35_prefill_kernels::(anonymous namespace)::cpu_fp16_matmul(...)
libomlx_qwen35_prefill_kernel_ops.dylib  omlx::qwen35_prefill_kernels::(anonymous namespace)::AneHybridQ4Primitive::eval_gpu(...)
libmlx.dylib        mlx::core::gpu::eval(mlx::core::array&)
```

I then tried to pin it down and **could not reproduce it in eight further runs**: worker counts 1, 2, 4, 8, and all four share combinations (`cpu` only, `+down`, `+gdn`, `+down+gdn`) each served the same 4k-token prefill cleanly. Same settings, same `performance-aware scheduling with 8 workers`, same eager compile of 64 MLPs. So it is intermittent, not configuration-dependent, and my first guess — that the worker count was landing in the entity slot — is **refuted** by the sweep.

What I would look at, given the frames: `DispatchApplyAttrStorage` is declared as `uint64_t[8]`, an **assumed layout for a private API** resolved through `dlsym(RTLD_DEFAULT, "dispatch_apply_attr_init")` and friends, none of which are in the macOS SDK here. If the real `dispatch_apply_attr_s` is larger than 64 bytes on a given OS build, `init` writes past the buffer and `set_parallelism` reads a garbage entity — which produces exactly "Unknown entity", and would be intermittent because it depends on surrounding stack contents. `ScopedDispatchApplyAttributes` is also stack-allocated per call while `apply` runs the shards, so a lifetime or reentrancy issue would look similar. Worth a sizeof guard or a fallback that abandons the attr path when `init` cannot be validated, since today the failure mode is a `SIGTRAP` that kills the server rather than a degraded scheduling path.

Environment: **M1 Ultra, 128 GB, macOS 26.5.2**, `mlx==0.32.0`, single-ANE mode, crash report `Python-2026-08-19-171648.ips` (happy to send the full `.ips`).

## Two build notes for anyone else testing this

1. `OMLX_WITH_CUSTOM_KERNEL=1` needs the Xcode **Metal toolchain component**, which is not installed by default: it fails with `error: cannot execute tool 'metal' due to missing Metal`. `xcodebuild -downloadComponent MetalToolchain` (688 MB) fixes it.
2. Building with `--no-build-isolation` silently uses whatever nanobind is already in the venv. With 2.15.0 the extension compiles fine and then disables itself at import: *"native kernels disabled — the extension was built with a nanobind ABI that does not match this mlx wheel"*. Installing the pinned `nanobind==2.13.0` first and rebuilding gives working symbols. The diagnostic message is genuinely good; it saved me a long hunt.

onthehub97:
I’d be interested to know what the tuner suggests for your M1 processor and what the results are on it.

onthehub97:
@jundot those changes requested have gone in along with a small patch to the tuner for single ANE devices. Ready for review

joshuaswarren:
> I’d be interested to know what the tuner suggests for your M1 processor and what the results are on it.

Running it now, I'll post the results here.

jundot:
@onthehub97 This one is a milestone, so before anything else: thank you. You built the ANE prefill path from nothing, then when the review got demanding you split a 3,200 line branch into clean PRs, answered every blocker within hours, and rewrote a kernel scheduling layer twice without complaint. Three compute units cooperating on one prefill is something no other MLX server does, and it exists here because you kept pushing on it.

I verified all four blockers are resolved. For the scheduling rewrite I went fairly deep: I checked the dispatch_apply_attr family against the libdispatch source and a compiled probe on my own machine, and the argument order, the entity constant, and the 64 byte storage layout are all exactly right at head. That also confirms the crash @joshuaswarren hit on 1e5345b was the parallelism count landing in the entity argument, which the head commit fixes, so it cannot recur. The memory accounting held up end to end in Joshua's runs and in my own wiring pass, the worker block is fully covered by try/catch, and the clone tool refuses to write anything when the scan finds inf or NaN. Joshua, the three round validation with refuted hypotheses was maintainer grade work, thank you as well.

Main moved under the branch while this was in review, so I pushed a merge commit that resolves the conflicts: the new CPU controls in the app now go through the working profile flow that landed in the meantime, and the load time warmup and observability changes from the other PRs are integrated. Python and Swift test suites both pass on the result. Merging.

A few things I found in my final pass that I will handle as follow ups on my side, same as the other PRs in this series: the bank preparation overwrites the packed q6 b/a suffix so q6 GDN would silently latch off, the tuner apply buttons still force dual_ane on which contradicts the new single ANE calibration, the MLP profile refinement is skipped when cpu_fraction is 0, the CPU toggle is disabled without dual ANE in the UI, and the down share wants a has_symbol guard. None of these need anything from you.

joshuaswarren:
@jundot, @onthehub97 - I'm glad I could help here. I'm more than happy to use my fleet (M1 Ultra 128GB, M1 Max 64GB, M2 Max 96GB) any time you need something tested, verified or benchmarked. Just let me know - tag me here or find me on Twitter with the same username - joshuaswarren. 

I've got a few test runs of the tuner running still, I'll post the results as a followup comment here just in case the information is useful for your continued efforts with ANE.

joshuaswarren:
Ran Tuner V2 on three Apple Silicon parts at `838a2e1`. The recommendation flips cleanly on ANE die count, which I think is the useful part.

| part | ANE dies | GPU-only PP | best PP | tuner verdict | delta |
|---|---:|---:|---:|---|---:|
| M1 Ultra, 128 GB | 2 | 145.75 | 137.15 | **disabled** | -5.9% if forced |
| M1 Max, 64 GB | 1 | 96.50 | 110.70 | **enabled**, mlp 0.6 / gdn 0.6 | **+14.7%** |
| M2 Max, 96 GB | 1 | 51.60 | 72.30 | **enabled**, mlp 0.4 / gdn 0.6 | **+40.1%** |

All three: `Qwen3.8-27B` mixed 4/5-bit, `dual_ane: false`, seq 2048, tuner run to completion (90 steps planned, 14 executed), advisor parked so nothing else held the GPU.

**On the M1 Ultra the tuner correctly declines.** It predicted 135.00 at mlp 0.53 / gdn 0.45, refined to 137.15 at gdn 0.40, then returned `{"enabled": false, "speedup_percent": 0.0}` because GPU-only measured 145.75. That matches what I have seen on this part before — the packed dual-ANE path fails its first bank with `ANE Code=6 / 0x20004` and splits at ~1 GiB, and GDN input projections have been a net loss here while winning on M2 Max. So V2's predict-then-verify is doing the right thing rather than optimistically enabling.

**Both single-die parts win, and the fractions differ**, which is the argument for the tuner existing: 0.6/0.6 on M1 Max versus 0.4/0.6 on M2 Max. Nobody would have guessed those.

Two caveats on my numbers, so they are not over-read:

1. **Cross-part absolute PP is not comparable here.** The M2 Max reports a lower GPU-only baseline (51.60) than the M1 Max (96.50), which is not a chip ordering — those two runs were on machines with different residency and thermal state. Within-part ratios are the signal; the absolute column is not.
2. **`cpu_enabled` came back false on all three.** On the two laptops that is expected: the checkpoint is not an FP16 clone, so the CPU gate cannot pass `gate.scales.dtype == mx.float16`. But on the Ultra I ran it *with* an FP16 clone and working kernels, and the tuner still selected `cpu_enabled: false` — the CPU-share configs never appear in the results table at all. Is the tuner meant to explore the CPU dimension, or is that still manual-only? If it should, I have the FP16 clone staged and can re-run.

**The ABI fix looks good.** Rebuilt with kernels at `838a2e1` and re-ran the four share configurations that produced my earlier `dispatch_apply_attr_set_parallelism` abort (`cpu` .25 alone, `+down` .50, `+gdn` .50, and both): **all four clean, no crash, no `Unknown entity`**, and `246 passed` including your new `test_qwen35_native_cpu_share`. Passing the entity constant explicitly instead of the shard count is clearly the right shape — my "worker count in the entity slot" reading was right about the mechanism and wrong about which value was landing there, which is why my own sweep appeared to clear it.

**What I would test next, if useful to you:** the tuner optimises prefill, and on this hardware ANE prefill has historically cost decode — I measured a +21% prefill / -40% decode trade at 4k on the Ultra. Both laptops now show a real prefill win, so the open question is what those configurations do to decode on a prefill-heavy workload (my real traffic carries ~55k-token prompts, where a +40% prefill would be worth paying for). Say the word and I will run decode alongside prefill on the M2 Max at mlp 0.4 / gdn 0.6 and report both.

onthehub97:
Those are really interesting results.

Could you try re-running on this branch: https://github.com/jundot/omlx/pull/2935

It has a new fused approach that should benefit older CPUs and a tuner that pays more attention to reproducibility.

jundot:
Follow ups are on main, and I ran a full A/B on this machine before writing this up. 4ff58585 fixes a regression my final pass caught: bank preparation was overwriting the packed q6 b/a suffix, so q6 GDN would have silently latched off at first prefill. Verified against real dual banks with a parity harness, cosine 0.9999 or better on qkv, z, b, and a for both group sizes. 7576cf7f is the smaller batch: the tuner apply no longer forces dual_ane on, the CPU toggle is not tied to dual ANE anymore, the MLP profile refinement works again without a CPU share, ANE warmup failures are soft, and the down share checks has_symbol before preparing.

Measurements on an M3 Ultra, Qwen3.8 27B oQ4e-mtp, 4k and 16k prefill plus 128 generated tokens, two passes per config with an ane_2048 warmup, second fully warm pass reported:

| config | 4k prefill tok/s | 16k prefill tok/s | vs baseline |
|---|---|---|---|
| ANE off | 458 | 436 | baseline |
| ANE split, no CPU share | 588 | 555 | +28% / +27% |
| ANE split + CPU share (fp16 clone) | 625 | 579 | +36% / +33% |

So the CPU share is worth about 5 points of prefill on top of the ANE split here, decode is unchanged, and the peak memory cost at these settings was about 7 GB. The fp16 clone requirement holds: on the stock bf16 checkpoint the CPU path stays a no-op.

The benchmark also surfaced two things, both handled on main now. First, the first request after loading a CPU shared model collapsed hard, 325 prompt tok/s and 13 decode tok/s at 4k, because it paid BNNS setup and the first touch of the FP16 rows. a79ce4b3 extends the load time warmup to the CPU path, one dummy chunk per shared module, about 3 seconds at load, and the first request now measures normally. Second, that same cold state was poisoning the tuner: before the warmup fix the tuner verified the CPU candidate at -32% on this machine and recommended everything off, after it the same run measures +3.8% and recommends the CPU config, consistent with the table above. So if anyone saw the tuner reject CPU sharing on an earlier build, it is worth rerunning on current main.

One caveat on the middle row: the tuner's refined gdn fraction can still land below the structural floor from #2899, it picked 0.35 here which compiles 0 GDN procedures, so that row uses a corrected 0.45. The refinement clamp for that is next on my list.

jundot:
@joshuaswarren Yes, the tuner explores the CPU dimension. Your 14 of 90 steps means every CPU candidate was dropped at the eligibility gate before anything was measured: it needs the FP16 clone, q4 or q8 gate/up layers (there are no q5 CPU kernels), and the compiled symbols. If you re-run the Ultra clone case on current main, the load log line "Warmed the CPU sharing path on N modules" tells you the gate passed.

joshuaswarren:
> Those are really interesting results.
> 
> Could you try re-running on this branch: #2935
> 
> It has a new fused approach that should benefit older CPUs and a tuner that pays more attention to reproducibility.

Will do! I'll post over there when done.

q-p:
I'm observing something that I cannot quite explain (and I'm asking here because that's where the fp16 model conversion comes from):
On my M4 Max Studio 64GB w/ oMLX 0.6.3rc2 `Jundot/Qwen3.8-27B-oQ4e-fp16-mtp` has a consistent prompt processing advantage (**GPU only**, so the baseline) of roughly 10% over `Jundot/Qwen3.8-27B-oQ4e-mtp` (275ppt/s vs 250ppt/s).

Is this expected? I thought >= M3 shouldn't care about fp16 vs bf16. Or is there something else going on?

PR #2898, opened 2026-08-19
feat(ane): warm compiled procedures at load so the first request measures inference

beaglemoo:
Eager compilation (#2814 and successors) moved ANE program creation to model
load, but each procedure still pays its first-evaluation cost inside the
first user request that crosses the prefill threshold. On an M5 Pro (64 GB,
Qwen3.8-27B oQ4e, sequence_length 2048, 128 banked procedures) that adds
**+5-8 s to first-request TTFT at 16K**: 46.1/48.3 s for the first request
vs 38.2/38.8 s steady-state on current main.

This PR adds `AneLinearModel::warmup()` — a host-signaled ticket around one
throwaway evaluation (input contents irrelevant, output discarded, errors
surfaced through the existing completion-error path) — and runs it for every
banked procedure after eager compilation, with a log line reporting the cost.

Measured on the same machine and protocol (model preloaded via the admin
API, TTFT of the first 16K request after load, two runs each):

| build | first request | steady state | load cost |
|---|---|---|---|
| main 4c00bbe5 | 46.1 / 48.3 s | 38.2 / 38.8 s | — |
| + this PR | **41.5 / 41.1 s** | 37.1 / 37.8 s | +2.9 s |

The residual first-vs-steady gap matches the thermal drift the ANE-off arm
shows on the same protocol, so the JIT component is fully paid at load.

Notes:
- Covers the banked dual-ANE path (the default). The per-layer fallback
  path could get the same loop if you want it there too — happy to extend.
- `warmup()` waits for any in-flight ticket first, so it is safe against
  the overlapping-evaluation guard.
- Sanitized serve logs for the measurements are in the round-5 gist linked
  from #2853.

### DIFF
--- a/omlx/custom_kernels/qwen35_prefill/csrc/bindings.cpp
+++ b/omlx/custom_kernels/qwen35_prefill/csrc/bindings.cpp
@@ -55,7 +55,11 @@ NB_MODULE(_ext, m) {
           &omlx::qwen35_prefill_kernels::AneLinearModel::output_dim)
       .def_prop_ro(
           "sequence_length",
-          &omlx::qwen35_prefill_kernels::AneLinearModel::sequence_length);
+          &omlx::qwen35_prefill_kernels::AneLinearModel::sequence_length)
+      .def(
+          "warmup",
+          &omlx::qwen35_prefill_kernels::AneLinearModel::warmup,
+          nb::call_guard<nb::gil_scoped_release>());
   m.def(
       "qwen35_ane_compile_linear",
       static_cast<std::shared_ptr<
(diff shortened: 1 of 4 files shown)


beaglemoo:
Independent data point from #2779: on an M5 Max (v0.6.3rc1, 27B, fractions 0.53/0.50), @studioburnside measured the first post-load request at **81.9s with ANE on vs 51.7s off** on identical filler — ~30s of per-procedure first-evaluation cost landing in the first request. That is the cost this PR moves to load time (~2.9s on my M5 Pro).

jundot:
Thanks, this is a clean fix for a cost every first request was silently paying. I checked the ticket lifecycle against the extension source: the host side setSignaledValue passes the readiness spin wait immediately, submitted and completed stay balanced, and the getattr guard keeps older compiled extensions on the previous behavior. Also a nice side effect that tuner slots now measure warmed inference. Merging as is. One small thing I will handle in a follow up on my side: the warm loop lets a warmup exception escape the enable path, so one failing procedure would abort the whole ANE setup instead of degrading, and I would rather catch it there and let the existing runtime disable deal with a broken procedure at first use.

beaglemoo:
Thanks — agreed on the exception path, catching it at the enable boundary and letting the runtime disable handle a broken procedure at first use is the right shape. Happy to re-verify on this machine when 0.6.3 final tags.

Issue #37, opened 2026-02-25
Cannot load the new Qwen3.5-120B-A10 model


kuanjames:
**Describe the bug**
omlx cannot load the new Qwen3.5-120B-A10 mlx model. The log shows this error:

mlx.engine_pool - ERROR - [-] - Failed to preload pinned model Qwen3.5-122B-A10B-4bit: Received 333 parameters not in model: 
language_model.vision_tower.blocks.0.attn.proj.bias,
language_model.vision_tower.blocks.0.attn.proj.weight,
language_model.vision_tower.blocks.0.attn.qkv.bias,
language_model.vision_tower.blocks.0.attn.qkv.weight,
language_model.vision_tower.blocks.0.mlp.linear_fc1.bias,
language_model.vision_tower.blocks.0.mlp.linear_fc1.weight,
language_model.vision_tower.blocks.0.mlp.linear_fc2.bias,
language_model.vision_tower.blocks.0.mlp.linear_fc2.weight,
language_model.vision_tower.blocks.0.norm1.bias,
language_model.vision_tower.blocks.0.norm1.weight,
language_model.vision_tower.blocks.0.norm2.bias,
language_model.vision_tower.blocks.0.norm2.weight,
language_model.vision_tower.blocks.1.attn.proj.bias,
language_model.vision_tower.blocks.1.attn.proj.weight,
language_model.vision_tower.blocks.1.attn.qkv.bias,
language_model.vision_tower.blocks.1.attn.qkv.weight,
language_model.vision_tower.blocks.1.mlp.linear_fc1.bias,
language_model.vision_tower.blocks.1.mlp.linear_fc1.weight,
language_model.vision_tower.blocks.1.mlp.linear_fc2.bias,
language_model.vision_tower.blocks.1.mlp.linear_fc2.weight,
language_model.vision_tower.blocks.1.norm1.bias,
language_model.vision_tower.blocks.1.norm1.weight,
language_model.vision_tower.blocks.1.norm2.bias,
language_model.vision_tower.blocks.1.norm2.weight,
language_model.vision_tower.blocks.10.attn.proj.bias,
language_model.vision_tower.blocks.10.attn.proj.weight,
language_model.vision_tower.blocks.10.attn.qkv.bias,
language_model.vision_tower.blocks.10.attn.qkv.weight,
language_model.vision_tower.blocks.10.mlp.linear_fc1.bias,
language_model.vision_tower.blocks.10.mlp.linear_fc1.weight,
[...]

Issue #17, opened 2026-02-16
`cors_origins` server config is defined but never applied as middleware


rndmcnlly:
**Describe the bug**

`ServerConfig.cors_origins` is defined in `omlx/config.py:60` with a default of `["*"]`, but no `CORSMiddleware` is ever added to the FastAPI app in `server.py`. This means browser-based frontends (e.g. Open WebUI) cannot reach the oMLX API due to failed CORS preflight checks.

The `/v1/chat/completions` endpoint returns `405 Method Not Allowed` on `OPTIONS` requests with an `Origin` header.

**To Reproduce**

1. Run oMLX (via the .app or CLI) on port 12348
2. From a browser page on a different origin, make a fetch request:
   ```js
   fetch("http://localhost:12348/v1/models", {
     headers: { "Authorization": "Bearer <key>" }
   })
   ```
3. Browser blocks the request: preflight `OPTIONS` returns 405, no `Access-Control-*` headers present

Alternatively via curl:
```bash
curl -I -X OPTIONS http://localhost:12348/v1/chat/completions \
  -H "Origin: https://example.com" \
  -H "Access-Control-Request-Method: POST" \
  -H "Access-Control-Request-Headers: Authorization, Content-Type"
# Returns: 405 Method Not Allowed
```

**Expected behavior**

The `cors_origins` config value should be used to add `CORSMiddleware` to the FastAPI app, e.g.:
```python
from fastapi.middleware.cors import CORSMiddleware

app.add_middleware(
    CORSMiddleware,
    allow_origins=config.server.cors_origins,
    allow_credentials=True,
    allow_methods=["*"],
    allow_headers=["*"],
)
```

Ideally, `cors_origins` should also be configurable via `~/.omlx/settings.json`:
```json
{
  "server": {
    "cors_origins": ["https://chat.example.com"]
  }
}
```

**Desktop:**
- OS: macOS 15 (Apple Silicon)
- oMLX: installed via GitHub release (.app)

**Additional context**

The `cors_origins` field was added in `config.py` and has a test (`test_config.py:92-105`), so it appears intentional but the wiring to the actual server was missed. As a workaround, a lightweight reverse proxy can add the headers externally.

Issue #13, opened 2026-02-15
Icon Missing from Menu Bar


usenetj23:
## Description
After installing oMLX, the application icon does not appear in the macOS Menu Bar. Additionally, oMLX is missing from the "Allow in the Menu Bar" list within System Settings. Despite the missing UI element, the backend service is running correctly and responding to requests.

## Environment
- **OS:** macOS Tahoe 26.3 (Fresh installation)
- **oMLX Version:** 0.1.2 (Upgraded from 0.1.1)
- **Related Software:** `mlx-lm` was previously installed and used on this machine.

## Steps to Reproduce
1. Install and launch oMLX (v0.1.2).
2. Observe the **macOS Menu Bar** (top right corner).
3. Open **System Settings > Menu Bar > Allow in the Menu Bar**.
4. Search for the oMLX entry in the list.

## Expected Result
- The oMLX icon should be visible in the Menu Bar for quick access.
- oMLX should be listed in System Settings as a configurable Menu Bar item.

## Actual Result
- **UI:** No oMLX icon is visible in the Menu Bar.
- **Settings:** oMLX is entirely missing from the Menu Bar settings list.
- **Backend:** The oMLX server is **working properly**. It is accessible via web interface and responds to HTTP POST requests from other network services.

## Additional Notes
Both `mlx-lm` and `oMLX` (default 8000) on the same port 11434. However, `mlx-lm` has been deactivated, so there is no active port conflict during testing.

jundot:
Sorry for the trouble. i just released 0.1.3 with the exact same issue fixed. i was on sequoia so didn't catch this on tahoe. please test the new version if you can. you'll need to kill the running process in terminal and install the new version. thanks for the feedback\!

usenetj23:
issue fixed
thx

PR #2903, opened 2026-08-19
perf(mtp): port Qwen3.8-27B native-MTP challenge optimizations (token-exact)

0xClandestine:
## Qwen3.8-27B native-MTP challenge → omlx port

Ports the optimizations from [Layr-Labs/qwen-3.8-mtp-challenge](https://github.com/Layr-Labs/qwen-3.8-mtp-challenge) (the Qwen 3.8 27B native-MTP speculativ-decode challenge) into omlx's Lightning MTP path (`omlx/patches/mlx_lm_mtp/`), evaluated against all **54 merged submission commits** in the challenge's main history (35 `Validate` + 19 `Accept`).

### Ports (all token-exact vs the serial trajectory, pinned checkpoints)

| Port | Optimization | Status |
|---|---|---|
| #3 `088f763b` | Packed Q/K/V concat on N — one affine-4 GEMM for Q+gate/K/V | default ON, **−10% decode**, scope-tagged to qwen3_5 |
| #8 `7b33621c` | Compact proposal vocabulary (98,304+26 rows) with device-side id remap | default ON, **−2% MTP decode**, acceptance preserved |
| #37 `45c257f1` | Fused GDN in-proj (qkv+z+b+a) + fused SwiGLU gate/up — one matmul each | default ON, **−1% decode**, 48/48 GDN + 64/64 MLP engage |
| #17 `a3104b04` | KV-only MTP-head history flush | opt-in `OMLX_QWEN_KV_ONLY_HISTORY=1` (token-exact, measured neutral) |
| #13 `6cb6c963` | Compiled GDN post-norm SiLU | opt-in `OMLX_QWEN_COMPILED_FUSIONS=1` (token-exact, measured 2.3× slower → OFF; two-output g+beta fusion **not ported** — laguna C1 ULP risk, documented) |

Every fused matmul is concat-on-N over independent rows (bit-exact by construction) with env ablation seams (`OMLX_QWEN_PACKED_QKV`, `OMLX_QWEN_COMPACT_DRAFT`, `OMLX_QWEN_FUSED_PROJ`).

### Correctness evidence
- Pinned checkpoints digest-verified per fixtures (backbone `EigenLabs/Qwen3.8-27B-4bit`, head `EigenLabs/Qwen3.8-27B-MTP-bf16`)
- MTP decode **token-exact vs serial at draft depths 2, 4, and 8** (96 tokens each, zero divergence — no width wall in the Python path)
- In-repo gate `tests/test_qwen38_mtp_token_exact.py` + 177 MTP unit tests pass
- Full per-submission ledger: `docs/qwen-mtp-port-log.md` (54 rows, status + correctness per submission)

### Tooling
`tools/qwen38_mtp/` stages the checkpoints (digest-verified download + merge under the `mtp.` prefix, matching the Swift loader) and drives decode-level token-exactness comparisons. One model load per process so a 64 GB machine never double-loads.

### Branch hygiene
Based on `jundot/omlx/main` (146d2724). 41 commits, one per submission labeled with the submission UUID. No behavior change outside the ported optimizations (all others keep the stock paths; qwen3_next models are scope-tagged out).


jundot:
Thanks for the port and the per-submission ledger, the ablation env seams made it easy to A/B this on my side. Here is what I measured (M3 Ultra 512GB, python code context prompts at 4k and 16k, tg128, temp 1.0 unless noted, 2 seeds each):

| model / condition | packs engaged | tg128 (baseline -> fused) | phys_footprint (baseline -> fused) |
|---|---|---|---|
| Qwen3.6-27B-4bit, temp 1.0, plain decode | attn 16/16, mlp 64/64, gdn 48/48 | 4k: 34.8 -> 35.6 t/s (+2.3%), 16k: 32.0 -> 32.6 t/s (+1.9%) | 19.2 -> 28.7 GB (+49%) |
| Qwen3.8-27B-oQ4e-mtp, temp 1.0, MTP on | attn 14/16, mlp 64/64, gdn 0/48 | 4k/16k: within seed noise, per-cycle backbone about 1 to 1.5% faster | 20.4 -> 27.5 GB (+35%) |
| Qwen3.5-9B-oQ4-mtp, greedy, MTP depth 3 | attn/mlp engaged, gdn 0 | 5.9 -> 5.7 ms/tok (about 3%) | 6.5 -> 8.8 GB (+36%) |

Prompt processing was flat everywhere (410 to 430 t/s at both context sizes).

A few things I found while checking where the packs actually engage: on oQ4e style checkpoints the mixed per-layer bits (5-bit in_proj_b/a, a couple of 5-bit k_proj layers) make the gdn fusion and some attention layers fall back to the stock path, and the compact draft vocab only engages for greedy decoding, so temp 1.0 workloads never see it. So the win I can reproduce is in the 2 to 3% range at best, and since the packed copies keep the original weights resident too, that is where the +35 to 49% footprint comes from.

Given that, I am going to hold this out of rc2 for now and take another look tomorrow with a bit more time. Not a rejection, the ledger and the token-exact gate are solid work, I mainly want to think through the defaults and the engagement coverage before this ships.

jackkbarnes:
I tested the current PR head (`fedea03b`) on an M5 Max 128 GB with Qwen3.8-27B-oQ8-mtp and hit a deterministic TurboQuant runtime failure.

### Configuration

- Apple M5 Max, 128 GB
- Qwen3.8-27B-oQ8-mtp
- TurboQuant KV enabled, 8-bit
- MTP enabled
- 24 GB hot cache
- exact oMLX v0.6.3 baseline
- source build with native kernels enabled

My test stack was v0.6.3 + #3113 + #2903 + #2991.

A tiny runtime probe with a 33-token prompt repeatedly failed with:

`'TurboQuantKVCache' object has no attribute 'group_size'`

The scheduler attempted cache-corruption recovery/re-prefill four times and then returned HTTP 500:

`RuntimeError: Cache corruption not recoverable after retries: 'TurboQuantKVCache' object has no attribute 'group_size'`

I then reverted only #2903, leaving #3113 and #2991 in place with the same model profile and TurboQuant configuration. The identical runtime probe returned HTTP 200 and subsequent 44k/132k-token runs completed normally.

### Suspected root cause

I think `_patch_attention_packed_qkv()` may be reintroducing the stale-SDPA-binding failure mode previously fixed in oMLX.

The PR currently does:

`from mlx_lm.models.base import scaled_dot_product_attention`

when installing the replacement attention implementation, and that replacement later calls the captured `scaled_dot_product_attention` directly.

TurboQuant installs/rebinds its SDPA dispatcher later during model loading.

This appears to match the failure mode fixed in commit `c269a6c0` (`fix(patches): resolve SDPA at call time so a later TurboQuant patch is reachable`). That fix specifically changed affected patches from an early-bound SDPA reference to resolving SDPA through the module at call time because otherwise a `TurboQuantKVCache` could reach plain mlx-lm SDPA and produce the same missing-`group_size` error.

The runtime ordering in my failing run was:

1. Qwen MTP model patch applied
2. TurboQuant attention patch applied
3. TurboQuant KV cache enabled at 8-bit
4. request enters attention
5. repeated `TurboQuantKVCache ... group_size` corruption/re-prefill
6. HTTP 500 after retries

By inspection, `OMLX_QWEN_PACKED_QKV=0` does not appear sufficient to avoid this specific issue: `_patch_attention_packed_qkv()` still replaces `Attention.__call__`; the env switch changes whether the projections are packed, but the replacement body still calls the captured SDPA reference.

A possible fix would be the same pattern already used by `c269a6c0`, e.g. resolve through `mlx_lm.models.base` at call time rather than importing the function into a local binding.

It may also be worth adding a regression test that installs/replaces the SDPA dispatcher after `_patch_attention_packed_qkv()` and verifies that the replacement attention reaches the later dispatcher.

Happy to test a patched branch on this exact M5 Max / Qwen3.8-oQ8-MTP / TurboQuant configuration.

Issue #2859, opened 2026-08-19
/api/status observability gaps: model_memory_used ignores KV growth; TurboQuant state invisible; ineligibility only logged at INFO


anicaise-ai:
Three small, related observability gaps found while benchmarking TurboQuant KV on 0.6.2. Filing them together since they all touch the same surface.

**1. `model_memory_used` does not move during KV growth.** During a cold 40k-token prefill (Qwen3.6-35B-A3B, M3 Ultra, paged SSD cache on), we sampled `/api/status` every 0.5 s: `model_memory_used` stayed byte-for-byte constant (35.83 GB) from before the prefill to after it — while ~2 GB of KV was being produced (we could measure it indirectly via the paged SSD cache: 1.97 GB written for ~39k stored tokens fp16). If the field is weights-only, the name is misleading; either way there is currently **no way to watch KV/cache memory from the API**. A `kv_memory_used` / `cache_memory_used` field would be very useful for capacity planning and watchdogs.

**2. TurboQuant state is not exposed anywhere in the API.** Whether TQ is enabled, at what bit depth, and whether it actually *engaged* for the loaded model can only be discovered by grepping server logs for "TurboQuant attention patch applied" / "TurboQuant KV cache enabled". A per-model field in `/api/status` (e.g. `"turboquant": {"requested": true, "bits": 4, "active": true}`) would make deployments verifiable.

**3. Ineligibility is only an INFO log.** When `turboquant_kv_enabled: true` is set on an MLA model (DeepSeek / GLM family) or an attention-sink model, the scheduler correctly falls back to fp16 and logs "TurboQuant disabled: model uses Multi-head Latent Attention…" — but at **INFO** level. Since the user *explicitly requested* quantization, a silent-ish downgrade deserves **WARNING**, and ideally the `"active": false` + reason in the status field from point 2. (We initially deployed TQ on a DeepSeek host and only realized from reading `_turboquant_eligible` in the source that it could never engage there.)

All three are cosmetic individually, but together they made a straightforward "is TQ actually on, and what does it save?" question require code reading, log grepping, and du(1) on the SSD cache directory.

Issue #2839, opened 2026-08-18
[Bug] ANE prefill not executing on oQ4e-mtp (v0.6.2): configured_layers=0, ANE duty=0, TG drops 57%


taozhiyuai:
# [Bug] ANE prefill not executing on oQ4e-mtp (v0.6.2) — configured_layers=0, ANE duty=0, TG drops 57%

**Hardware:** Apple M3 Max 128GB
**oMLX Version:** v0.6.2
**Model:** Qwen3.8-27B-oQ4e-mtp
**Test Date:** 2026-08-19

## Summary

When running `v062_ane_verify.sh` (16K + 32K benchmarks with ANE on), the ANE prefill pipeline appears enabled but **does not execute any operations**. The ANE profiling reports `configured_layers=0`, `operations=0`, and `ane0_duty=0.0000` / `ane1_duty=0.0000` throughout. This results in a **~57% drop in TG** compared to the earlier 32K A/B test where ANE was working correctly.

## Benchmark Results

| Config | PP (tok/s) | TG (tok/s) | E2E (s) | Peak Mem |
|--------|-----------|-----------|---------|----------|
| 16K ANE on (v062) | 249.5 | 28.7 | 70.1 | 26.2 GB |
| 32K ANE on (v062) | 220.9 | 10.8 | 160.2 | 27.6 GB |
| 32K ANE on (earlier A/B) | ~227 | 25.4 | 149.5 | ~27 GB |

The 32K TG of **10.8 tok/s** is drastically lower than the **25.4 tok/s** observed in the controlled A/B test (5 rounds, OFF→ON→OFF→ON→OFF) on the same hardware/version.

## Server Log Evidence

```
[benchmark-ane-profile] pp=4097 category=mlp operations=0 configured_layers=0 compiled_layers=unknown expected_operations=0 observed_shapes=0.000 ... ane0_duty=0.0000 ane1_duty=0.0000
[benchmark-ane-profile] pp=4097 category=gdn operations=0 configured_layers=0 compiled_layers=unknown expected_operations=0 observed_shapes=0.000 ... ane0_duty=0.0000 ane1_duty=0.0000
```

Note: `configured_layers=0` means the ANE tuner found **zero eligible Qwen MLP/GDN layers** to compile for this model.

Also observed during startup:
```
ANE tuning failed for Qwen3.8-27B-Uncensored-oQ8e-mtp
RuntimeError: The ANE candidate loaded, but no eligible Qwen MLP/GDN layers were compiled
[...]

PR #2904, opened 2026-08-19
feat(ane): surface Qwen ANE prefill no-ops in logs and /api/status

### LINKED ISSUE #2824: qwen35_ane_prefill_enabled is a silent no-op on non-4bit checkpoints: no warning, no status field, only a load-time delta reveals it
## Summary

`qwen35_ane_prefill_enabled: true` is accepted and persisted on an 8-bit (oQ8e)
Qwen3.8 checkpoint, but the patch never installs, logs nothing at INFO, and reports
nothing back through the API. The setting reads as tuned and does nothing.

The gate itself is correct — `_affine_spec(gate, ..., allowed_bits=(4,))` and the
same for `up` in `omlx/patches/qwen35_ane_prefill.py` mean an 8-bit MLP is
genuinely ineligible. The defect is that ineligibility is **silent**.

Now that 0.6.2 ships the ANE split tuner, this matters more: a user with an 8-bit
checkpoint can enable the flag, open the tuner, and have no way to learn that the
path was never available for their quantization.

## Environment

- Mac Studio, Apple M3 Ultra, 512 GB, macOS 26.6.1
- oMLX **0.6.2** (also observed on 0.6.1)
- `Qwen3.8-27B-...-oQ8e-mtp`, `model_type=qwen3_5`, `engine_type=vlm`
- quantization `bits=8, group_size=64, mode=affine`
- all four custom kernels `available: true`, `is_nax_available(): False`

## How I eventually diagnosed it

Nothing in the product told me. The sequence was:

1. Set `qwen35_ane_prefill_enabled: true` (plus the `sequence_length`, `fraction`,
   `max_layers`, `dual_ane`, `gdn*` keys). All persisted.
2. `grep -iE "ane|neural engine|qwen35_ane" ~/.omlx/logs/server.log` → **no output at all**.
3. Forced unload + load and timed it. ANE compiles eagerly, so an active path costs
   ~27-29s; the load took **3.0s** — the only observable signal that it was inert,
   and an entirely indirect one.
4. Read `patches/qwen35_ane_prefill.py` and found `allowed_bits=(4,)` on the
   gate/up projections (`down` allows `(2,4,5,6,8)`), versus our 8-bit weights.

Step 4 is the answer, and it requires reading the source of an installed wheel.

## Requested change

Log once at **WARNING** when the patch declines to install because of an
eligibility check, naming the reason and the offending value, e.g.

```
Qwen ANE prefill requested but unavailable: MLP gate/up projections are 8-bit,
ANE path requires 4-bit (down proj accepts 2/4/5/6/8). Continuing GPU-only.
```

For comparison, the NAX skip added in #2779 does exactly this and it is
immediately legible:

```
Qwen ANE prefill skipped: NAX GPU, tensor-unit prefill is faster
```

Ideally also surface it in the model status payload (an `ane_prefill_eligible`
field, or the reason string) so the tuner UI can say "not available for this
quantization" instead of offering a benchmark that cannot help.

Related: the installers in this area are wrapped in `except Exception:
logger.debug(...)`, so genuine install failures are invisible at INFO too. Same
class of problem — a failed optimization should be louder than a successful one.

finaltv971:
## Problem

Enabling ANE prefill on a model that isn't eligible is currently a **silent
no-op**. `enable_qwen35_ane_prefill()` returns `0` on several quiet exits, so a
user who turned ANE on sees no error yet the whole model runs prefill on the
GPU — with no log line and nothing in the API to explain why. Reported symptoms:
`configured_layers=0` with ANE inactive, and `/api/status` exposing no ANE state
at all.

The three silent exits, confirmed by reading the code:
- `_install_dispatch()` fails → `return 0` with no log;
- no dense MLP passes the eligibility gate (`_eligible_pair` / `_affine_spec`,
  which needs affine int4/5/6/8 quant with group_size 64 or 128) → `candidates`
  empty, `return 0`;
- 0 procedures compiled on either the banked or the plain path → the `if count:`
  info line simply doesn't fire, so nothing is logged.

## Change

**Make every no-op say so** (`omlx/patches/qwen35_ane_prefill.py`):
- `WARNING` when the dispatch hook can't be installed;
- `WARNING` when no eligible MLP layer is found, with the number of dense MLP
  modules scanned and the exact quant requirement;
- `WARNING` when 0 procedures end up compiled (both the banked early-return and
  the plain loop).

**Make the state observable**:
- new `qwen35_ane_prefill_status(model) -> dict` — a safe, JSON-serialisable
  report of the counters `enable_` already attaches
  (`attempted` / `configured` / `mlp_layers` / `gdn_layers` / `dual_ane_layers`
  / `resident_programs`). `attempted` distinguishes "ANE never ran here" from
  "ran but compiled nothing"; `configured` is True only when ≥1 procedure landed
  on the ANE.
- `GET /api/status` now carries an aggregated `ane_prefill` block across loaded
  models (`omlx/server.py`). It's best-effort and wrapped so the endpoint can
  never fail because of it.

No behaviour change to compilation or dispatch — this is logging + reporting
only. Nothing new is compiled, no public signature changes.

## Before / after

```text
# before: ANE requested on a bf16 checkpoint
$ … (server starts, ANE quietly does nothing, logs say nothing)

# after
WARNING  Qwen ANE prefill requested but no eligible MLP layers found
         (28 dense MLP module(s) scanned; ANE requires affine int4/5/6/8
         quantization with group_size 64 or 128)

$ curl -s localhost:PORT/api/status | jq .ane_prefill
{ "available": true, "configured_models": 0, "models": [
    { "model_id": "…", "attempted": true, "configured": false,
      "mlp_layers": 0, "gdn_layers": 0, "dual_ane_layers": 0,
      "resident_programs": 0 } ] }
```

## Tests

Added to `tests/test_qwen35_ane_prefill.py` (all `not slow`, no hardware):
- status on a configured model, on an attempted-but-empty model, and on an
  untouched model;
- the no-eligible-layer path now emits the warning (via `caplog`) and still
  attaches counters so status is never blank.

`pytest tests/test_qwen35_ane_prefill.py -m "not slow"` → **58 passed**.

## Notes

- `/api/status` reads `entry.engine._model` defensively, matching the existing
  status code's style; engines without that attribute are simply skipped.
- Follow-ups this unblocks but does not attempt: clamping the tuner to the
  structural `gdn_fraction` floor, and reserving ANE prefill transient memory in
  the preflight.

## Issues

Closes #2824 — the reported defect is precisely "no warning, no status field"
when ANE prefill is enabled but ineligible; this PR adds both the warnings and
the `/api/status` block.

Related to #2839 — makes the `configured_layers=0` / zero-duty case emit a
warning instead of staying silent, so the symptom is visible in the log. It does
**not** diagnose why an otherwise-eligible oQ4e checkpoint compiled 0 layers
(root cause out of scope for this observability change).

Related to #2859 — addresses the "ineligibility only logged at INFO" gap (now a
WARNING) and adds the missing ANE block to `/api/status`. The other gaps in that
issue (KV-growth accounting in `model_memory_used`, TurboQuant visibility) are
untouched.

### DIFF
--- a/omlx/patches/qwen35_ane_prefill.py
+++ b/omlx/patches/qwen35_ane_prefill.py
@@ -1408,21 +1408,39 @@ def enable_qwen35_ane_prefill(
         logger.warning("ANE native extension unavailable; Qwen ANE prefill skipped")
         return 0
     if not _install_dispatch():
+        logger.warning(
+            "Qwen ANE prefill: dispatch hook could not be installed "
+            "(mlx-vlm/mlx-lm Qwen backend not registered); ANE prefill inactive, "
+            "running prefill on GPU"
+        )
         return 0
 
     config = _AnePrefillConfig(sequence_length, fraction, variant, dual_ane)
     candidates = []
-    modules = model.modules() if hasattr(model, "modules") else ()
-    for module in modules:
+    scanned_mlp = 0
+    for module in model.modules() if hasattr(model, "modules") else ():
         if not all(
             hasattr(module, name) for name in ("gate_proj", "up_proj", "down_proj")
         ):
             continue
+        scanned_mlp += 1
         if not _eligible_pair(module):
             continue
         candidates.append(module)
         if len(candidates) >= max_layers:
             break
+
+    if not candidates:
+        # The most common silent no-op: ANE was requested but every dense MLP
+        # failed the eligibility gate (needs affine int4/5/6/8 quant with
+        # group_size 64 or 128). GDN may still be eligible below; the final
+        # summary warns if nothing at all is compiled.
+        logger.warning(
+            "Qwen ANE prefill requested but no eligible MLP layers found "
+            "(%d dense MLP module(s) scanned; ANE requires affine int4/5/6/8 "
+            "quantization with group_size 64 or 128)",
+            scanned_mlp,
+        )
 
     banked = _enable_dual_procedure_banks(
         model,
@@ -1439,7 +1457,7 @@ def enable_qwen35_ane_prefill(
         model._omlx_ane_dual_prefill_count = dual_count
         model._omlx_ane_resident_program_count = resident_programs
         model._omlx_ane_procedure_count = count + gdn_count
-        if count:
+        if count or gdn_count:
             logger.info(
                 "Eagerly compiled %d MLP and %d GDN procedures into %d "
                 "instance-pinned ANE programs (sequence_length=%d)",
@@ -1447,6 +1465,11 @@ def enable_qwen35_ane_prefill(
                 gdn_count,
                 resident_programs,
                 sequence_length,
+            )
+        else:
+            logger.warning(
+                "Qwen ANE prefill enabled but 0 procedures were compiled; "
+                "the whole model runs prefill on GPU"
             )
         return count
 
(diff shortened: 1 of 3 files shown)


jundot:
Thanks, this closes a gap that has been generating confused reports for a while. I verified the counter attachment on both the banked and the plain path, so the attempted flag is accurate, and the pool access matches the existing status code style. Merging. Two small things I will adjust in a follow up on my side: VLM engines expose the model as _vlm_model so they are currently skipped in the status block, and the available field reads as hardware availability when it actually means the patch is importable.

PR #2905, opened 2026-08-19
feat(ane): guard the GDN prefill fraction floor and stop silent 0-GDN

### LINKED ISSUE #2899: ANE tuner v2 can recommend a gdn_fraction below the structural floor, silently compiling 0 GDN procedures
## Summary

The v2 ANE tuner (#2891) can recommend — and persist — `gdn_enabled: true`
with a `gdn_fraction` that is below the model's structural minimum, in which
case `_prepare_gdn_for_bank` rejects every GDN layer and the patch silently
compiles **zero** GDN procedures. The recommendation is self-consistent (its
verification run also executed MLP-only, so the headline number is honest),
but the persisted config misleads, and the tuner's GDN calibration stage may
be comparing several below-floor fractions that are all the same
pure-GPU-GDN configuration.

## Observed (M5 Pro 64 GB, Qwen3.8-27B oQ4e, v0.6.3rc1, sequence_length 2048)

Tuner v2 recommendation on this machine:

```
enabled=true, mlp_fraction=0.40, gdn_enabled=true, gdn_fraction=0.25, +17.0%
```

Applying it:

```
Eagerly compiled 64 MLP and 0 GDN procedures into 2 instance-pinned ANE programs (sequence_length=2048)
```

`gdn_fraction: 0.45` on the same model compiles 48 GDN procedures. Nothing
in the log says why 0.25 produced none.

## Mechanism

`patches/qwen35_ane_prefill.py`, `_prepare_gdn_for_bank`:

```python
ane_outputs = (int(total_outputs * config.fraction) // 128) * 128
...
if ane_outputs < z_outputs:
    return None
```

The ANE slice must cover at least the z projection. At 0.25 on this model
every GDN candidate fails the check and returns `None`, silently.

## Suggestions (any one of these would do)

1. Clamp the tuner's GDN candidate fractions to the model's minimum viable
   fraction (`ceil(z_outputs / total_outputs)` aligned up to 128 columns),
   so calibration only measures fractions that actually engage;
2. and/or when `gdn_enabled` is true but zero GDN procedures compile, log a
   WARNING and flip the persisted `gdn_enabled` to false so the stored
   config matches reality;
3. and/or have the recommendation drop `gdn_enabled` when the profile shows
   `category=gdn operations=0` during verification (the instrumentation
   from #2829/#2853 already exposes this).

Happy to test a patch on this machine — the tuner-vs-real-request protocol
from the #2853 thread is set up and repeatable here.

finaltv971:
## Problem (#2899)

The v2 ANE tuner can recommend — and persist — `gdn_enabled: true` with a
`gdn_fraction` below the model's structural minimum. `_prepare_gdn_for_bank`
then rejects **every** GDN layer and the patch compiles **0** GDN procedures,
silently:

```
Eagerly compiled 64 MLP and 0 GDN procedures into 2 instance-pinned ANE programs
```

Root cause, in `_prepare_gdn_for_bank`:

```python
ane_outputs = (int(total_outputs * config.fraction) // 128) * 128
if ane_outputs < z_outputs:
    return None
```

The 128-aligned ANE slice must at least cover the z projection. Below
`ceil(z_outputs / 128) * 128 / total_outputs` every GDN candidate returns
`None` — and nothing says why.

## Change

- **`min_viable_gdn_fraction(gdn)`** — the smallest fraction whose 128-aligned
  slice clears the exact rule the bank enforces (with a guard for the
  truncation in `int(total * fraction)`), or `None` when the layer can never
  engage.
- **Tuner (`admin/ane_tuning.py`)** — drops GDN candidate fractions below that
  floor before calibration, so it never *measures* several identical
  pure-GPU-GDN configs and never *recommends* one that engages 0 GDN layers.
  Logs which fractions were dropped. (Suggestion 1 in the issue.)
- **`enable_qwen35_ane_prefill`** — when GDN is requested, MLPs compile, but 0
  GDN layers engage, logs a WARNING naming the required floor, instead of
  leaving the "0 GDN procedures" line unexplained. Covers both the banked
  early-return and the per-layer path. (Suggestion 2.)

No behaviour change when the fraction is already viable. Compilation and
dispatch are untouched — this is a floor guard plus logging.

## Before / after

```text
# before: tuner recommends gdn_fraction=0.25 on a model whose floor is ~0.45
Eagerly compiled 64 MLP and 0 GDN procedures … (no reason logged)

# after (apply path)
WARNING  Qwen ANE GDN prefill requested at gdn_fraction=0.250 but 0 GDN layers
         engaged: this model needs gdn_fraction >= 0.450 for the 128-aligned
         ANE slice to cover the z projection. GDN runs on GPU.

# after (tuner)
INFO     Tuner: skipping GDN fractions below the structural floor 0.450 for
         this model (dropped 0.15, 0.25); they engage 0 GDN layers
```

## Tests

Added to `tests/test_qwen35_ane_prefill.py` (all `not slow`, no hardware):
- the floor matches the bank's rejection rule and clears the grid step below it;
- the impossible case (`ceil(z/128)*128 > total`) returns `None`;
- the below-floor no-op warns with the floor; a healthy GDN stays silent.

`pytest tests/test_qwen35_ane_prefill.py -m "not slow"` → **57 passed**.

## Notes

- `admin/ane_tuning.py` imports `fastapi`, absent from my local env, so the
  tuner hunk is verified by `ast` parse + review and the floor arithmetic is
  unit-tested via `min_viable_gdn_fraction`. Happy for the reporter to confirm
  on the M5 Pro tuner-vs-real-request protocol they offered in the thread.
- Independent of #2904 (ANE no-op observability), which branches separately
  from `main`.

Closes #2899

### DIFF
--- a/omlx/admin/ane_tuning.py
+++ b/omlx/admin/ane_tuning.py
@@ -670,7 +670,22 @@ def _calibrate_components_sync(
             state, dense0, dense1 = value
             prepared.append(("mlp", fraction, state, dense0, dense1))
     if gdn is not None:
-        for fraction in fractions:
+        # Skip GDN fractions below the model's structural floor: they compile 0
+        # GDN procedures at apply time, so measuring them only compares
+        # identical pure-GPU-GDN configs and can get one recommended (#2899).
+        gdn_floor = patch.min_viable_gdn_fraction(gdn)
+        gdn_fractions = fractions
+        if gdn_floor is not None:
+            gdn_fractions = [f for f in fractions if f >= gdn_floor]
+            dropped = [f for f in fractions if f < gdn_floor]
+            if dropped:
+                logger.info(
+                    "Tuner: skipping GDN fractions below the structural floor "
+                    "%.3f for this model (dropped %s); they engage 0 GDN layers",
+                    gdn_floor,
+                    ", ".join(f"{f:.2f}" for f in dropped),
+                )
+        for fraction in gdn_fractions:
             config = patch._AneGDNConfig(
                 run.request.sequence_length, fraction, 8, True
             )
--- a/omlx/patches/qwen35_ane_prefill.py
+++ b/omlx/patches/qwen35_ane_prefill.py
@@ -522,6 +522,35 @@ def _eligible_gdn(gdn: Any) -> bool:
     )
 
 
+def min_viable_gdn_fraction(gdn: Any) -> float | None:
+    """Smallest ``gdn_fraction`` that engages the ANE on ``gdn``.
+
+    ``_prepare_gdn_for_bank`` keeps a 128-aligned ANE slice and rejects the
+    layer when that slice can't even cover the z projection
+    (``ane_outputs < z_outputs``). Below this fraction every GDN layer returns
+    ``None`` and 0 GDN procedures compile — the silent no-op of issue #2899.
+    Returns ``None`` when the layer is ineligible or can never engage.
+    """
+    if not _eligible_gdn(gdn):
+        return None
+    _, z, _, _ = _gdn_linears(gdn)
+    qkv = _gdn_linears(gdn)[0]
+    z_outputs = int(z.weight.shape[0])
+    total_outputs = z_outputs + int(qkv.weight.shape[0])
+    if total_outputs <= 0:
+        return None
+    # smallest 128-aligned ANE slice that covers z, matching the bank rejection
+    ane_min = ((z_outputs + 127) // 128) * 128
+    if ane_min > total_outputs:
+        return None
+    frac = ane_min / total_outputs
+    # _prepare_gdn_for_bank truncates via int(total * fraction); nudge up if the
+    # truncation would drop the aligned slice back below z.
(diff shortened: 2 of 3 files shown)


jundot:
Thanks, the floor helper is good and the warning at apply time is useful. But the tuner change does not fix #2899. The calibration stage already skips fractions below the floor, since _prepare_gdn_for_bank returns None for them. The bad recommendation actually comes from a later stage, _profile_refinement, which picks a fraction from a grid with no floor check. That code is mine so I will fix it there. If you drop the Closes line and the tuner hunk and keep just the helper and the warning, I will merge that part.

jundot:
Coming back to this one. The refinement clamp that actually closes #2899 landed on main in 2a7bac99 along with a floor helper that accounts for the dual ANE alignment, so the tuner side here has been overtaken, and the branch no longer applies after the recent ANE rework. The apply-time warning was the piece still missing, so I added it on main in 209b186e based on this PR. Closing, and thanks for the work here.

PR #2912, opened 2026-08-20
fix(memory): size hybrid KV blocks by cache layers

### LINKED ISSUE #2725: [BUG] MemoryMonitor unnecessary rejections: estimate uses total layer count instead of KVCache layer count (4x over-estimate on Qwen3.5/3.6/3.8 hybrids)
**Describe the bug**

MemoryMonitor logs the KVCache layer count and then computes the per-block estimate from num_hidden_layers instead. On hybrid models where full_attention_interval=4, this over-estimates KV by exactly 4×.
Related: #1763, but this report is about error with turboquant_kv_bits=None, so this is distinct from #1763.

EXAMPLES
**Qwen3.8-27B-8bit**

Model info set: 64 layers (16 KVCache), 4 KV heads, 24 Q heads, 256 head_dim.
Estimated memory per 64-token block: 16.00 MB

64 layers × 2 (K+V) × 4 heads × 256 × 2 B = 262,144 B/token × 64 = 16,777,216 B = **16.00 MiB** ← matches log
16 KVCache layers → **4.00 MiB** ← correct value

**Qwen3.6-35B-A3B-8bit**

Model info set: 40 layers (10 KVCache), 2 KV heads, 16 Q heads, 256 head_dim.
Estimated memory per 64-token block: 5.00 MB

40 layers → 81,920 B/token × 64 = 5,242,880 B = **5.00 MiB** ← matches log
10 KVCache layers → **1.25 MiB** ← correct value

Both match the all-layers figure to four significant figures, on models with different layer counts, head counts, and architectures.

Impact: the inflated figure feeds the memory guard, so adaptive_prefill_throttle and 507 rejections fire against phantom KV. Observed throttling at current=56.62GB against target=98.62GB — roughly 40GB of headroom unused — with per_token=119269.9KB in the same line, which is ~116MB/token and clearly wrong by orders of magnitude.

The GatedDeltaNet layers use ArraysCache(size=2) with fixed per-sequence state (~75MB total on the 27B), not per-token growth, so they shouldn't be billed per block at all. Cross-ref mlx-lm#1480 for the same architectural distinction.

Expected: per-block estimate uses the KVCache layer count already computed and logged.

oMLX 0.6.0.dev1, M5 Max 128GB, macOS 26. 

## Additional log evidence

Single session, M5 Max 128 GB, macOS 26.

### Reproduces on two models, KV quantization disabled

```
2026-08-16 17:06:57,910 - omlx.memory_monitor - INFO - Model info set: 64 layers (16 KVCache), 4 KV heads, 24 Q heads, 256 head_dim. Estimated memory per 64-token block: 16.00 MB
2026-08-16 17:06:57,925 - omlx.cache.paged_ssd_cache - INFO - PagedSSDCacheManager updated layer cache signature (64 layers, 2 unique types, turboquant_kv_bits=None, cachelist_subtypes=no)

2026-08-16 17:08:49,690 - omlx.memory_monitor - INFO - Model info set: 40 layers (10 KVCache), 2 KV heads, 16 Q heads, 256 head_dim. Estimated memory per 64-token block: 5.00 MB
2026-08-16 17:08:49,721 - omlx.cache.paged_ssd_cache - INFO - PagedSSDCacheManager updated layer cache signature (40 layers, 2 unique types, turboquant_kv_bits=None, cachelist_subtypes=no)
```

Signature lines included to show `turboquant_kv_bits=None` in the same second.

| Model | Layers | KVCache | KV heads | head_dim | Logged | All layers | KVCache only |
|---|---|---|---|---|---|---|---|
| Qwen3.8-27B-8bit | 64 | 16 | 4 | 256 | 16.00 MB | **16.00 MiB** | 4.00 MiB |
| Qwen3.6-35B-A3B-8bit | 40 | 10 | 2 | 256 | 5.00 MB | **5.00 MiB** | 1.25 MiB |

`64 x 2 x 4 x 256 x 2 B x 64 tokens = 16,777,216 B = 16.00 MiB`
`40 x 2 x 2 x 256 x 2 B x 64 tokens = 5,242,880 B = 5.00 MiB`

Exact match to the all-layers figure on both, despite different layer counts, head counts, and dense-vs-MoE FFNs. Both have `full_attention_interval = 4`, hence the uniform 4x factor.

### Another component in the same process has the correct count

```
2026-08-16 16:42:12,474 - omlx.scheduler - INFO - TurboQuant: converted 15/64 cache layers to 8.0-bit, skipped last KVCache layer
```

This path correctly identifies 16 KVCache layers out of 64. `MemoryMonitor`, same model, bills 64.

### Downstream effect

```
2026-08-16 15:41:56,548 - omlx.scheduler - INFO - Request adcfa821 needs prefill headroom before throttling (reason=adaptive_prefill_throttle, current=82.08GB, predicted=46.41GB, target=98.62GB)
2026-08-16 15:41:57,029 - omlx.scheduler - INFO - Prefill throttled for adcfa821: chunk 408 -> 160 (usage 56.62GB vs sizing target 78.03GB, dynamic ceiling 86.70GB, per_token=119269.9KB, floor=32). Prefill runs slower until usage drops.
```

Chunk cut from 408 to 160 at `usage 56.62GB` against `target 98.62GB` — ~40 GB of headroom unused. `per_token=119269.9KB` (~116 MB/token) in the same line is wrong by orders of magnitude and may be a separate defect.

### Why the other 48 layers should not be billed per block

`make_cache()` returns `ArraysCache(size=2)` for GatedDeltaNet layers, `KVCache()` only for full-attention layers (ml-explore/mlx-lm#1480). The `ArraysCache` state is overwritten in place, not appended, so it is constant with context length — roughly 75 MB per sequence on Qwen3.8-27B. Per-block growth for those layers is zero.

saichowdary007:
## Summary

Fixes #2725. MemoryMonitor now sizes per-token KV blocks from the layers that actually retain KV state. Hybrid recurrent and rotating layers no longer multiply block estimates or the paged-SSD writer queue cap.

## Changes

- prefer `num_kv_cache_layers` in `estimate_block_memory`, while preserving explicit overrides and zero-layer rotating-only models
- align scheduler and paged-SSD estimator documentation with the corrected semantics
- add hybrid and rotating-only MemoryMonitor regressions
- add scheduler-to-SSD plumbing coverage for a 64-layer/16-KV-layer model
- use the conservative 200 KB manager default when a rotating-only or ArraysCache-only layout has a zero per-token estimate, so the SSD writer queue cannot expand to its 256-entry ceiling
- normalize non-positive estimates inside `PagedSSDCacheManager` as a defense-in-depth guard

## Compatibility

`MemoryMonitor.estimate_block_memory(num_layers=...)` continues to treat an explicit override as authoritative, including `0`. When the override is omitted, the measured KV-cache layer count is used when available; this is the intended public semantic change for hybrid cache layouts.

## Validation

- `python -m pytest -q tests/test_memory_monitor.py tests/test_memory_monitor_vlm_config.py tests/test_paged_ssd_cache.py` (259 passed)
- targeted Ruff syntax/import checks passed
- `git diff --check` passed

The full test matrix remains for CI on the project macOS runners. Fork PR checks may require maintainer approval before Actions can start.

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -71,12 +71,12 @@ _PENDING_WRITES_CEILING = 256
 _PENDING_WRITE_PUT_TIMEOUT_SECONDS = 1.0
 
 # Conservative defaults for the per-block cost estimator. The actual
-# bytes-per-block depends on the model (num_layers × num_kv_heads ×
+# bytes-per-block depends on the model (KV-cache layers × num_kv_heads ×
 # head_dim × dtype_size × block_size_tokens × 2). At construction time
 # the PagedSSDCacheManager doesn't always know these — see __init__'s
 # ``expected_kv_bytes_per_token`` parameter — so the module-level
-# default targets a 35B-class bf16 model whose per-token KV is ≈200 KB
-# spread across all layers. Smaller models will be over-conservative
+# default targets a 35B-class bf16 model whose per-token KV is ≈200 KB.
+# Smaller models will be over-conservative
 # (fine), larger models or larger blocks should pass an explicit value.
 _DEFAULT_BLOCK_SIZE_TOKENS = 256
 _DEFAULT_KV_BYTES_PER_TOKEN = 200_000
@@ -1587,9 +1587,10 @@ class PagedSSDCacheManager(CacheManager):
                 don't pin gigabytes at saturation; passing a smaller value lets
                 the cap grow to give workloads with many tiny blocks enough
                 burst headroom.
-            expected_kv_bytes_per_token: Per-token KV byte estimate (all
-                layers, K + V, dtype). Together with ``expected_block_size_tokens``
-                this drives the bytes-aware queue cap. Defaults to a
+            expected_kv_bytes_per_token: Per-token KV byte estimate (KV-cache
+                layers, K + V, dtype). Together with
+                ``expected_block_size_tokens`` this drives the bytes-aware
+                queue cap. Defaults to a
                 35B-class bf16 estimate; pass an explicit value for
                 quantized models or unusually wide/narrow architectures.
             expected_layer_cache_types: Optional current cache layout. When
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -76,10 +76,22 @@ _PENDING_WRITE_PUT_TIMEOUT_SECONDS = 1.0
 # the PagedSSDCacheManager doesn't always know these — see __init__'s
 # ``expected_kv_bytes_per_token`` parameter — so the module-level
 # default targets a 35B-class bf16 model whose per-token KV is ≈200 KB.
-# Smaller models will be over-conservative
-# (fine), larger models or larger blocks should pass an explicit value.
+# Smaller models will be over-conservative (fine), while larger models or
+# larger blocks should pass an explicit value.
 _DEFAULT_BLOCK_SIZE_TOKENS = 256
 _DEFAULT_KV_BYTES_PER_TOKEN = 200_000
+
+
+def _normalize_kv_bytes_per_token(value: int) -> int:
+    """Return a safe positive estimate for writer-queue sizing.
+
+    A model with only fixed-state or rotating caches can legitimately report
+    zero *per-token* KV bytes.  Zero is not a useful queue-sizing input,
+    though: it would make every block appear to cost one byte and pin the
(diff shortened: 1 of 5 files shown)


jundot:
Thanks for the fix. I verified the hybrid path end to end: both the scheduler and the DFlash guard already classify cache layers via collect_kv_layer_specs, so the corrected estimate is live on every path, and the numbers match the analysis in #2725 exactly. The old all-layers test was only a drift pin from the resident-KV work, so updating it here is right. The zero-estimate guard for the SSD queue is a nice touch. Merging.

PR #2913, opened 2026-08-20
fix(cli): strip leaked -- separator from launch extra_args

fparrav:
## Problem

When launching a model with extra args, a leaked -- separator was passed through into extra_args, breaking the launch.

## Fix

Strip the leaked -- separator from launch extra_args before use.

## Test plan

- [ ] Launch a model with extra args and confirm no -- leaks into the generated command.


fparrav:
Closing: duplicate of #2818, which is already merged. Created in error.

PR #2919, opened 2026-08-20
feat(dflash): expose max_snapshot_tokens as a per-model setting

sergezuber:
## What this fixes

`dflash-mlx` skips inserting a prefix snapshot when the prompt is longer than `max_snapshot_tokens`
(default `32_000`, `dflash_mlx/runtime/config.py`). `DFlashEngine._build_runtime_context` builds the
runtime config with explicit keyword arguments and never passed that field, so on the oMLX path the
value was unreachable — neither model settings nor the environment could change it (the server does not
inherit `DFLASH_*` variables).

That is invisible on chat-sized prompts and total on agent-sized ones. An agent request carries a large
fixed prefix — system prompt plus tool schemas — so a ~50k-token prompt exceeded the default on every
request, no snapshot was ever inserted, and each turn re-prefilled the entire prompt.

## Measured

Apple M5 Pro, 48 GB, a 35B MoE at 4-bit through `omlx serve` with DFlash enabled:

- prefill **80–93 s on every turn**, `hit_kind=miss hit_tokens=0` on every request
- the prefix key was byte-identical between turns, `publish_generation_snapshot=True`,
  `cache_active=True`, `stable_prefix_len` 29 125 of 29 130 — everything correct except that the
  snapshot was never admitted
- with the threshold set to `0`, the same workload reuses its prefix and prefill drops to ~2 s

## The change

`dflash_max_snapshot_tokens` joins the other `dflash_*` model settings and is forwarded to the runtime
config.

- defaults to `None`, so behaviour is **unchanged** unless the setting is written
- `0` disables the cap
- the cache remains bounded by `prefix_cache_max_bytes` and `prefix_cache_max_entries` either way, so
  lifting the token cap does not make it unbounded

Two files, 16 added lines, no behaviour change by default.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -161,6 +161,7 @@ class ModelSettingsRequest(BaseModel):
     dflash_in_memory_cache_max_bytes: int | None = None
     dflash_ssd_cache: bool | None = None
     dflash_ssd_cache_max_bytes: int | None = None
+    dflash_max_snapshot_tokens: int | None = None
     dflash_draft_window_size: int | None = None
     dflash_draft_sink_size: int | None = None
     dflash_block_size: int | None = None
@@ -535,6 +536,7 @@ def _sanitize_diffusion_settings_dict(settings: dict) -> None:
         "dflash_draft_quant_activation_bits",
         "dflash_draft_quant_group_size",
         "dflash_max_ctx",
+        "dflash_max_snapshot_tokens",
         "dflash_draft_window_size",
         "dflash_draft_sink_size",
         "dflash_block_size",
@@ -642,6 +644,7 @@ def _sanitize_diffusion_model_settings(settings) -> None:
     settings.dflash_in_memory_cache_max_bytes = 8 * 1024 * 1024 * 1024
     settings.dflash_ssd_cache = False
     settings.dflash_ssd_cache_max_bytes = 20 * 1024 * 1024 * 1024
+    settings.dflash_max_snapshot_tokens = None
     settings.dflash_draft_window_size = None
     settings.dflash_draft_sink_size = None
     settings.dflash_block_size = None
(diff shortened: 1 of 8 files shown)


jundot:
Thanks, the diagnosis is right and I confirmed the semantics against dflash-mlx: the insert gate is max_tokens > 0, so 0 really does lift the cap, and None keeps the 32000 default. The problem is the wiring stops at the engine layer, so the knob is still unreachable in practice: it's missing from the request model and PATCH handler in admin/routes.py, the reset and unsupported-fields lists there, the profile field list in model_profiles.py, the engine reload signature in engine_pool.py, and the dashboard modal. Right now the only way to use it is hand-editing model_settings.json plus a restart, and changing it never reloads a running engine. Could you extend the PR to cover those spots, dflash_draft_window_size is a good template for every one of them, and add a small settings round-trip test? Also the two long lines in engine/dflash.py need a black pass.

sergezuber:
Thanks - extended the PR to cover every spot you listed, using `dflash_draft_window_size` as the template.

- `admin/routes.py` added to `ModelSettingsRequest`, the PATCH handler in `update_model_settings`, the `unsupported_none_fields` list in `_sanitize_diffusion_settings_dict`, and the reset in `_sanitize_diffusion_model_settings`
- `model_profiles.py` added to the profile field list
- `engine_pool.py` added to the reload `add(...)` sequence, so a change now reloads a running engine
- dashboard `dashboard.js` (field list, load, save, and both reset paths) and a `Max snapshot prompt tokens` input in `_modal_model_settings.html`, next to the other prefix-cache controls
- `engine/dflash.py` the two long lines now use the same parenthesised shape black gives the neighbouring `_draft_window_size`, and the comment above them is re-wrapped to the width the rest of the file keeps

**One deliberate departure from the template, worth a look.** `dflash_draft_window_size` collapses `0` to `None` in both the PATCH handler and the dashboard's save path. For this field `0` is meaningful and has to survive: the insert gate is `max_snapshot_tokens > 0`, so `0` lifts the cap while absent keeps the 32000 default. Copying that shape would quietly restore the default and leave the setting exactly as unreachable through the UI as it was through the file. So the handler tests `value is not None and value >= 0`, and the dashboard tests for emptiness rather than truthiness.

**Test.** `test_max_snapshot_tokens_roundtrip` covers absent / `0` / a positive cap in both directions, and `test_max_snapshot_tokens_defaults_to_none` asserts the key is omitted from `to_dict()` when unset so the dflash default still applies. Verified to bite: defaulting the field to `32000` instead of `None` fails both.

`tests/test_dflash_engine.py`: 113 passed. The 4 failures there (`TestDFlashEngineInit`, `TestDFlashOutputParserWiring`, `TestDFlashCachedTokensWiring`) are identical before and after this commit checked by stashing.

I did not run black over the whole file: it would also reformat an unrelated `_DRAFT_PRECISION_TAGS` line. Happy to include that if you would rather the file be black-clean in one pass.

PR #2922, opened 2026-08-20
[fix] Align reasoning effort controls in App UI

popfido:
## Summary

The native App View chat-template editor presents `reasoning_effort` with its Custom toggle below the value control, while Force sits beside that control. This makes the mode selection visually disconnected from the input it changes and creates a disorderly two-row block.

This PR puts the controls in their task order: **Custom → value → Force**. The preset picker and custom numeric text field both retain the same fixed width, so selecting Custom changes only the control's contents rather than its layout. `ViewThatFits` preserves a readable two-line fallback when the available width or accessibility text size cannot accommodate the single row.

## Cause and effect

Before this change, a reader first encountered the value and Force controls, then had to look below to find Custom. Switching Custom replaced the picker with a text field, but the relationship was not apparent at a glance. The new order lets users choose the mode before editing its value, then decide whether that value must override request-level input.

## Validation

- Compared against current `upstream/main` (`146d2724`): it still has the old value/Force row with Custom below it; this change is not already present upstream.
- Built and visually checked the compiled native App View against a real local model configuration in preset mode.
- `xcodebuild test -project apps/omlx-mac/oMLX.xcodeproj -scheme oMLX -destination 'platform=macOS' -derivedDataPath /private/tmp/omlx-reasoning-effort-derived-data`
  - 286 tests passed; 1 integration smoke test skipped by its existing `OMLX_INTEGRATION` guard.

Before:
image
After:
image
image

### DIFF
--- a/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
+++ b/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
@@ -914,6 +914,8 @@ private struct EntryEditor: View {
     let client: OMLXClient
     let entryID: UUID
 
+    private static let reasoningEffortValueWidth: CGFloat = 130
+
     @Environment(\.omlxTheme) private var theme
 
     private var entry: ChatTemplateKwargEntry {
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, the new order reads much better and the screenshots made it easy to check. I verified the region has no drift against current main and the change is layout-only. Merging.

darebfh:
Hey guys,
thanks for the great work! 
I installed oMLX via homebrew and wonder whether I can set reasoning_effort also in the web interface (e.g. for Owen-3.8-27b)? 

Thanks
Daniel

PR #2926, opened 2026-08-20
fix: write settings.json atomically and move corrupt files aside

### LINKED ISSUE #2924: settings.json written non-atomically: a torn file silently reverts the server to defaults, including disabled API-key auth
`GlobalSettings.save()` writes `settings.json` in place with `os.open(..., O_WRONLY | O_CREAT | O_TRUNC, 0o600)` followed by `json.dump` (`omlx/settings.py:1347-1353`). This is not atomic: a crash mid-write, a full disk, or two server processes saving concurrently can leave a torn file. By contrast, `ModelSettingsManager._save()` already does this correctly. It writes to a `.tmp` file and then calls `Path.replace()` (`omlx/model_settings.py:487-492`).

This isn't theoretical, I hit it today. I run two `omlx serve` instances (an LLM instance and a dedicated STT instance) that at the time shared one base path. Both call `save_cli_overrides()` at startup, and starting them simultaneously under a process manager produced this tail (valid JSON followed by remnants of the longer previous version):

```
  "idle_timeout": {
    "idle_timeout_seconds": null
  }
}conds": null
  }
}
```

The dangerous part is what happens next. `_load_from_file()` catches `json.JSONDecodeError` and only logs a warning (`omlx/settings.py:1037-1038`), so the server starts with **all defaults**. If `auth.api_key` lived only in `settings.json` (the normal case when configured via the app/dashboard), the server now runs with `api_key = None`, and `verify_api_key` returns `True` unconditionally when no key is configured (`omlx/server.py:322-324`). A user who configured "All networks" binding now has an unauthenticated server on their LAN, with no error surfaced anywhere. Deployments that set `OMLX_API_KEY` via env are unaffected since `_apply_env_overrides()` restores the key (`omlx/settings.py:1116-1117`).

Suggested fix, two independent layers:

1. Make `GlobalSettings.save()` use the same tmp-write + `os.replace()` pattern as `ModelSettingsManager._save()`, keeping the `0o600` mode on the temp file (plus `f.flush()` and `os.fsync()` before the rename).
2. Treat a parse failure of an existing `settings.json` as loud, not silent: rename the corrupt file to `settings.json.corrupt-<timestamp>` and log an error, so a security-relevant config reset can't happen invisibly.

PR incoming for both.

alexferrao:
Fixes #2924.

I hit this in production yesterday: I run two `omlx serve` instances (one for LLMs, one dedicated to STT) that shared a base path, and starting them together under pm2 tore `settings.json` (details and the torn tail are in the issue). The server then came up on defaults with only a warning in the log. My key survives via the `OMLX_API_KEY` env var, but for anyone whose key lives only in settings.json this silently disables auth. This PR is the code version of how I fixed it on my machine, so others don't have to find out the hard way.

Two small changes, both behavior-preserving for the happy path:

1. **Atomic save.** `GlobalSettings.save()` now writes to a `0o600` temp file, fsyncs, and `os.replace()`s it into place. This is the same pattern `ModelSettingsManager._save()` already uses, so the two config stores now behave consistently. The rename carries the temp file's mode, which also covers the old chmod-before-write migration path. Readers never see a partial file on any platform (`os.replace` is atomic on POSIX and Windows), single-instance setups are unaffected, and concurrent writers now settle on last-writer-wins instead of interleaved corruption.

2. **Loud handling of an already-corrupt file.** On `JSONDecodeError`, `_load_from_file()` now renames the file to `settings.json.corrupt-<timestamp>` and logs an error naming the auth consequence, instead of a warning and silent defaults. The server still starts (nothing gets more fragile for unattended setups, including the macOS app supervisor), but the evidence is preserved for recovery and the log makes the config reset visible. If the rename itself fails the old behavior remains as fallback.

Added two tests: save leaves a parseable file with `0o600` and no temp remnant, and a corrupt file is moved aside with defaults loaded. `tests/test_settings.py` passes (268 tests). One note for reviewers: `TestGlobalSettings::test_load_from_file` fails on any machine where `OMLX_API_KEY` is exported in the shell, unrelated to this change; happy to file that separately.

### DIFF
--- a/omlx/settings.py
+++ b/omlx/settings.py
@@ -26,6 +26,7 @@ import logging
 import os
 import shutil
 from dataclasses import asdict, dataclass, field
+from datetime import datetime
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal
 
@@ -1035,7 +1036,24 @@ class GlobalSettings:
                 )
 
         except json.JSONDecodeError as e:
-            logger.warning(f"Failed to parse settings file {path}: {e}")
+            # A corrupt settings file silently reverting the server to
+            # defaults is security-relevant (auth.api_key lives here), so
+            # preserve the evidence and be loud instead of a debug-level shrug.
+            backup = path.with_name(
+                f"{path.name}.corrupt-{datetime.now().strftime('%Y%m%d-%H%M%S')}"
+            )
+            try:
+                os.replace(path, backup)
+                logger.error(
+                    f"Settings file {path} is corrupt ({e}); moved it to "
+                    f"{backup} and continuing with defaults. Restore it or "
+                    "reconfigure, otherwise API-key auth may be disabled."
+                )
+            except OSError:
+                logger.error(
+                    f"Settings file {path} is corrupt ({e}) and could not be "
+                    "moved aside; continuing with defaults."
+                )
         except OSError as e:
             logger.warning(f"Failed to read settings file {path}: {e}")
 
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the report and the fix, preserving the corrupt file as evidence matters more than it looks. Merging. One note: both writers still share the same settings.json.tmp name, so two instances saving at the same moment can interleave inside the temp file before the rename. I'll push a small follow-up that makes the temp name unique per process, and give ModelSettingsManager the same treatment since it has the identical pattern.

PR #2929, opened 2026-08-20
fix(cluster): harden CUDA enrollment and ConnectX verification

ashhart:
Follow-up to #2591, which added heterogeneous Metal and CUDA pool support.

## Fixes

- claim the one-time join credential before slow worker provisioning, with a 30-minute join window and a two-hour claimed session that cover the bounded install path
- update the controller key source restriction during re-enrollment, preserve unrelated keys, write atomically, and refuse a symlinked `authorized_keys` file
- rerun the bounded NCCL direct-link verification before every activation instead of trusting cached browser state
- replace a LAN-style test fixture with the RFC 5737 TEST-NET-2 address `198.51.100.197`

This PR does not implement a hierarchical Ring to NCCL supernode. Verified CUDA workers remain adjacent ranks in the outer Ring. That topology belongs in a separate feature PR.

## Testing

- `uv run pytest -q tests/test_cluster*.py tests/test_nccl_fabric_worker.py`
- 990 passed
- Ruff checks passed for every changed Python file
- `node --check omlx/admin/static/js/dashboard.js` passed
- `git diff --check` passed

### DIFF
--- a/docs/distributed-cluster.md
+++ b/docs/distributed-cluster.md
@@ -125,7 +125,7 @@ the dashboard URL uses localhost, set **Settings > Server host** to `0.0.0.0`,
 restart oMLX, and enter the Studio's private IPv4 address in the card.
 
 Select **Generate join command**, copy it, and paste it into one CUDA worker. The
-command expires after ten minutes and can be claimed only once. It may ask for
+command expires after thirty minutes and can be claimed only once. It may ask for
 `sudo`; package installation, the worker-only virtual environment, SSH key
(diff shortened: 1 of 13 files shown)


jundot:
Thanks for tightening the CUDA enrollment lifecycle. Claiming before bounded provisioning, refreshing the controller source restriction, and re-running ConnectX verification on the normal activation path are useful and safe improvements. LGTM, merging. I will move the verification guard into the shared activation boundary and reject a symlinked parent `.ssh` path in a follow-up.

PR #2930, opened 2026-08-20
fix(cluster): recover from SSH pairing failures

### LINKED ISSUE #2763: Cluster: SSH issue. Coordinator key is refused by worker node. Worker node seems to work OK.
**Describe the bug**

Cluster:
- Main/coordinator: M5 Max MBP 128GB Ram - aka MBP in this report
- Worker: M4 Max Mac Studio 128GB Ram - aka MS in this report

Versions:
- OSX: Both machines are on 26.5.2
- OMLX: Both machines are running 0.6 installed via the dmg process

Issue:
On coordinator, it states the error: "SSH connected but the peer would not accept this Mac's key.". Under "Technical Details" it states: "<MS identifier>.local has no Python interpreter for hardware discovery".
Further down the page in the "Verify the worker" section, it says "RDMA device probe failed for <worker hostname>.local: <username>@<worker hostname>.local: Permission denied (publickey,password,keyboard-interactive)"

I'm not sure if the Python error is correct or not as both machines were cloned from one another originally, and I dont see any significant drift. Both are using homebrew installed Python3.12 . If I run `ssh <username>@<other machine IP> "python --version"` it returns the correct version ("which python3" also returns the correct data - i.e., not the "python 3.9" that Apple installs by default.

I also setup password-less SSH commands between my machines which works fine.

Worker seems to be connected OK in omlx, it sees the other machine OK. MBP maintains the "Checking the connection..." status, the MS sees the MBP as the worker. However the UI on the worker is constantly updating as the connection to the MBP seems to break or re-connect (this makes the UI unstable as the page is constantly expanding and contracting). If I try to launch a model, on the MS (worker) omlx tries to establish the connection, it mentions "Still measuring the workers — wait a moment, then try again." then refreshes to the original state again and no model is launched.

Potential edge case:
What I think might have happened is during the initial setup, I was originally confused as to where the "Sharing pairing secret" went. I generated the keys, moved them across to the worker server and then the worker credentials back again. But at some point I might have hit the "Generate" button again for the pairing secret and did regenerate a new coordinator local key but never set it over at the worker again because I cant! The worker thinks things are OK and never shows the connection box to set the keys and generate its own key ( the initial setup process part)
Either way, the option to set and create the keys has completely disappeared altogether since the initial setup and I cant see a way to get it back.

I suspect perhaps an edge case where the coordinator key was valid and accepted by the worker, but regenerated before the coordinator could accept the worker key. This has left a state where the worker thinks things are OK and never shows the connection process again, but the coordinator thinks there are errors.

**To Reproduce**
1. Go through the initial setup process of generating the keys between servers etc
2. Before you copy over the worker key to the coordinator, re-generate the coordinator key again.
3. The Worker thinks everything is OK, the coordinator spouts the errors I mentioned and nothing works.

**Additional context**
server.log file added

[copy_coordinator_server.log.gz](https://github.com/user-attachments/files/31149906/copy_coordinator_server.log.gz)

ashhart:
Fixes #2763

## Summary

* preserve OpenSSH transport failures during remote Python discovery so authentication errors reach the pairing guidance
* treat internal FQDN and Bonjour .local aliases as the same local Mac
* require an explicit, confirmed request before rotating the managed cluster key
* generate replacement keys off-path so headless rotation works and a failed generation leaves the live pair intact

## Verification

* 989 cluster tests passed
* JavaScript syntax check passed
* Ruff passed
* real ssh-keygen rotation reproduced and verified

### DIFF
--- a/omlx/admin/static/js/dashboard.js
+++ b/omlx/admin/static/js/dashboard.js
@@ -2286,15 +2286,30 @@
 
             async generateClusterSshKey() {
                 if (this.clusterSshKeyGenerating) return;
+                const overwrite = Boolean(this.clusterSshKey?.available);
+                if (overwrite && !window.confirm(
+                    'Regenerating this key disconnects every paired worker. '
+                    + 'You will need to exchange keys again on every Mac. Continue?'
+                )) return;
                 this.clusterSshKeyGenerating = true;
                 try {
-                    const response = await fetch('/admin/api/cluster/ssh-key/generate', {
+                    const endpoint = '/admin/api/cluster/ssh-key/generate'
+                        + (overwrite ? '?overwrite=true' : '');
+                    const response = await fetch(endpoint, {
                         method: 'POST',
                     });
                     if (response.ok) {
                         this.clusterSshKey = await response.json();
-                        // Show success notification
-                        this.showNotification('SSH key generated successfully', 'success');
+                        this.clusterExchangeToken = null;
+                        this.clusterPeerExchangeToken = '';
+                        this.clusterKeyExchangeResult = null;
+                        if (overwrite) this.invalidateClusterPeer(true);
+                        this.showNotification(
+                            overwrite
+                                ? 'SSH key regenerated. Pair every worker again before reconnecting.'
+                                : 'SSH key generated successfully',
+                            'success'
+                        );
                     } else {
                         const error = await response.json();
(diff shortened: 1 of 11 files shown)


jundot:
Thanks for covering the SSH recovery paths. Preserving OpenSSH failures, normalizing Bonjour identity, and making key rotation explicit and off-path address the reported failure mode cleanly. LGTM, merging. I will handle the incomplete post-generation dashboard state and the partial two-file replacement case in a follow-up.

PR #2932, opened 2026-08-20
[fix] Preserve profile while changing Qwen ANE settings

popfido:
## Summary
- Keep Qwen ANE Prefill App View controls in the working model profile instead of directly writing model settings.
- Stage ANE tuner recommendations in the working profile for explicit Save or Update.
- Validate enabled Qwen ANE settings before creating or updating a profile.

## Why
Direct writes diverge from the active profile, so the server intentionally clears its profile association. Toggling Qwen ANE settings therefore switched the model profile to None and did not enable profile Create or Update.

## Testing
- Added regression coverage for Qwen ANE working-profile serialization, dirty-state binding, and tuner-result staging.
- Focused ModelSettingsScreenVMTests: 12 tests passed.
- git diff --check passed.

### DIFF
--- a/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
+++ b/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
@@ -1106,10 +1106,9 @@ private struct ExperimentalSection: View {
     @Environment(\.omlxTheme) private var theme
 
     var body: some View {
-        // Most experimental fields are profile edits. Qwen ANE controls are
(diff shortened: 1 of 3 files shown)


jundot:
Thanks, the diagnosis is right: the server clears the profile link on purpose when a direct save diverges from it, so routing the ANE controls through the working profile like every other experimental field is the correct fix, and the validation before save is a nice touch. Merging. One heads up: the staged tuner apply still hardcodes dual_ane to true, which I am about to remove across the web UI as well in a follow up, so your path will get the same one line change.

PR #2933, opened 2026-08-20
feat(ane): reserve ANE prefill transient memory in the preflight

### LINKED ISSUE #2841: ANE prefill triggers hard memory-pressure abort + model eviction (Qwen3.8-27B)

finaltv971:
## Problem (#2841)

Loading Qwen3.8-27B with ANE prefill and sending a long first prompt (17,330
tokens, cache miss) drives memory to 29.4 GB — above the 28.5 GB hard watermark
— so the enforcer aborts the request and evicts the model, then loops on
reload. Disabling ANE prefill avoids it entirely.

The preflight reservation (`estimate_prefill_peak_bytes`) prices only KV + SDPA
and explicitly excludes anything else. When ANE prefill is on, the private
runtime also holds a fixed-shape fp16 **input and output IOSurface per compiled
slice** for the whole prefill — 64 MLP slices here, plus GDN — and none of that
is reserved, so admission lets the request in and it blows the watermark
mid-prefill.

## Change

- **`ane_prefill_transient_bytes(model, sequence_length)`** (patch): sums the
  per-slice I/O surfaces, `(input_dim + ane_output) * seq * 2`, doubled for
  dual-ANE slices, across the eagerly-compiled MLP and GDN slices. 0 when no
  ANE slice is attached, so it's safe on any model.
- **`MemoryMonitor`**: `set_model_info` gains an `ane_prefill_transient_bytes`
  reserve; `estimate_prefill_peak_bytes` adds it to the KV+SDPA peak;
  `set_model_info_from_model` computes it via the patch (defensive import — 0
  when the ANE backend isn't present). **Every** preflight caller (scheduler,
  DFlash guard, server handler) now reserves the spike with no change of its
  own, because they all go through the monitor.

No behaviour change when ANE prefill is off (reserve is 0). Nothing about
compilation or dispatch changes — this is a reservation term.

## Scope and honesty

This reserves the **I/O-surface term**, which is the deterministic, dominant,
and exactly-computable part of the spike (dims × seq × 2). It does **not**
model the full MLX merge scratch on the GPU side, so treat it as a *floor* for
the reservation, not an exact peak. On the issue's 32 GB M2 Max the ceiling is
essentially all RAM, so even reserving the dominant term should move admission
to refuse-or-shrink before the hard watermark instead of aborting mid-request.

I can't run this configuration (Qwen3.8-27B on a 32 GB M2 Max) — the arithmetic
is unit-tested and the wiring is defensive, but an on-device confirmation from
the reporter that the abort/evict loop is gone would be the real proof.

## Tests

- `tests/test_qwen35_ane_prefill.py`: the estimate sums slice I/O, doubles for
  dual-ANE, and is 0 without ANE or with `seq <= 0`.
- `tests/test_memory_monitor.py`: the reserve is added to the prefill peak, and
  defaults to 0.

`pytest tests/test_qwen35_ane_prefill.py tests/test_memory_monitor.py -m "not slow"`
→ **126 passed**.

## Notes

- Independent of #2904 (no-op observability) and #2905 (GDN fraction floor);
  branches separately from `main`.

Closes #2841

### DIFF
--- a/omlx/memory_monitor.py
+++ b/omlx/memory_monitor.py
@@ -224,6 +224,9 @@ class MemoryMonitor:
         # full-attention formula. Empty for non-hybrid models.
         self._rotating_layer_specs: tuple[tuple[int, int], ...] = ()
         self._prefill_memory_profile: PrefillMemoryProfile | None = None
+        # Transient ANE-prefill I/O surfaces (issue #2841); set via
+        # set_model_info, 0 unless the Qwen ANE prefill backend is attached.
+        self._ane_prefill_transient_bytes: int = 0
         # Fixed per-sequence recurrent state (GDN/Mamba ArraysCache),
         # measured once from a live cache after the first prefill chunk.
         self._fixed_state_bytes: int = 0
@@ -407,6 +410,7 @@ class MemoryMonitor:
         kv_bytes_per_token: Optional[float] = None,
         rotating_layer_specs: Sequence[tuple[int, int]] | None = None,
         prefill_memory_profile: PrefillMemoryProfile | None = None,
+        ane_prefill_transient_bytes: int = 0,
     ) -> None:
         """
         Set model information for memory estimation.
@@ -466,6 +470,11 @@ class MemoryMonitor:
             if count > 0 and window > 0
         )
         self._prefill_memory_profile = prefill_memory_profile
+        # Transient ANE-prefill I/O surfaces (issue #2841). 0 unless the Qwen
+        # ANE prefill backend is attached; charged on top of the KV+SDPA peak
+        # so a long first prompt reserves the private-runtime spike instead of
+        # tripping the hard watermark mid-prefill.
+        self._ane_prefill_transient_bytes = max(int(ane_prefill_transient_bytes), 0)
         # A new model's fixed state must be re-measured; a stale value from
         # the previous model would silently mis-charge admission.
         self._fixed_state_bytes = 0
@@ -780,7 +789,11 @@ class MemoryMonitor:
         # baseline. Resident math includes window-capped sliding-window
         # layers and measured fixed state, not just full-attention KVCache.
         kv = self.estimate_resident_kv_bytes(new_tokens, chunk_tokens=eff_chunk)
-        return attn + kv
+        # ANE prefill holds fixed-shape I/O surfaces per compiled slice for the
+        # whole prefill; reserve them so a long first prompt doesn't trip the
+        # hard watermark mid-request and get aborted (issue #2841). 0 when the
+        # ANE backend is not attached.
+        return attn + kv + self._ane_prefill_transient_bytes
 
     def estimate_chunk_transient_bytes(self, n_tokens: int, kv_len: int) -> int:
         """Transient SDPA activation bytes for ONE prefill chunk.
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for looking into #2841, the surface math is the right idea. The problem is the reservation never turns on where the bug happens: the LM engine fills the monitor through a direct set_model_info call in scheduler.py, which does not pass the new kwarg, so it stays 0. The GDN estimate also uses the full projection width instead of the ANE slice, and the dual case counts the output twice. Fixing this properly means changing the scheduler and monitor code and testing on real hardware, so I will do it myself and reference this PR in the commit. Closing here; #2904 is merged.

Issue #2914, opened 2026-08-20
[0.6.3rc1] ANE Tuner crashes: "Loaded engine does not expose its model" when the model is loaded via the DFlash engine


waterliu1981:
## Summary

In v0.6.3rc1, running the ANE Split Tuner on a model whose current engine is the **DFlash engine** fails before the first calibration step:

```
RuntimeError: Loaded engine does not expose its model
```

This is a **different failure from #2908** (ANE Tuner OOM at _VERIFY_SLOT, 48 GB machine): here there is no memory pressure — the crash happens immediately at `_calibrate_components`, the moment the tuner tries to reach into the live engine for the model object. It is **deterministic**: it fails every time the model is loaded with DFlash enabled.

## Environment

| | |
|---|---|
| Hardware | Apple **M5 Max**, **128 GB** unified memory |
| OS | macOS 26.6.2 (Build 25G83) |
| oMLX | **0.6.3rc1**, app build **2388** |
| Target model | `mlx-community/Qwen3.8-27B-mxfp4` (~17.8 GB loaded) |
| Draft model | `z-lab/Qwen3.8-27B-DFlash2` |

Reproduced twice in one hour (11:30:36 and 11:42:54), same model, same path.

## Per-model settings at crash time

```json
{
  "dflash_enabled": true,
  "qwen35_ane_prefill_enabled": true,
  "qwen35_ane_prefill_sequence_length": 2048,
  "qwen35_ane_prefill_fraction": 0.53,
  "qwen35_ane_prefill_max_layers": 64,
  "qwen35_ane_prefill_dual_ane": true,
  "qwen35_ane_prefill_gdn": true
}
```

## Steps to reproduce

1. Load any model with DFlash enabled (`dflash_enabled: true` + a compatible draft checkpoint).
2. Open **Hardware Acceleration → ANE Configuration** in the web dashboard / app.
3. Start the **ANE Split Tuner**.
4. The tuner fails immediately, before any calibration bank is compiled. (Workaround: turn DFlash off, let the model reload through the regular LM/VLM engine, then run the tuner.)

## Observed trace (`~/.omlx/logs/server.log`)

```text
2026-08-20 11:30:36,693 - omlx.admin.ane_tuning - ERROR - [-] - ANE tuning failed for Qwen3.8-27B-mxfp4
Traceback (most recent call last):
  File "/Applications/oMLX.app/Contents/Resources/omlx/admin/ane_tuning.py", line 821, in run_tuning
[...]

Issue #2908, opened 2026-08-19
[Bug] ANE Tuner V2 (v0.6.3rc1) OOMs at Step 11/13 (_VERIFY_SLOT): calibration program banks and base engine retained during verification reload


Collinw24:
## Summary

In `v0.6.3rc1`, running the ANE Split Tuner on `Qwen3.8-27B-oQ4e-mtp` fails at **Step 11 of 13** (`_VERIFY_SLOT`) with `kIOGPUCommandBufferCallbackErrorOutOfMemory`.

During the initial 10 calibration steps, the base model is retained in memory alongside 10 compiled ANE calibration banks. When Step 11 attempts to stage the full verification model, the previous engine and compiled banks have not been evicted, causing a transient double-model allocation that exhausts Metal unified memory.

---

## Environment

* **oMLX Version:** `0.6.3rc1`
* **OS:** macOS 27.0 (Build 26A5416b)
* **Hardware:** MacBook Pro (Apple M3 Max, 16 CPU / 40 GPU cores, 48 GB Unified Memory)
* **Model:** `Jundot/Qwen3.8-27B-oQ4e-mtp` (~15.8 GB weights, TurboQuant 8-bit KV, MTP enabled)
* **Kernel Memory Cap (`iogpu.wired_limit_mb`):** 43,520 MB (42.5 GiB)

---

## Steps to Reproduce

1. Load `Jundot/Qwen3.8-27B-oQ4e-mtp` in oMLX `v0.6.3rc1`.
2. Open **Hardware Acceleration / ANE Configuration** in the web dashboard / app.
3. Start the **ANE Split Tuner**.
4. Observe the tuner progress through the 10 calibration fractions (5 MLP + 5 GDN).
5. At Step 11 (`_VERIFY_SLOT` — "Predicted optimum" verification), model loading fails.

---

## Observed Trace (`server.log`)

```text
2026-08-19 19:23:12,309 - omlx.engine_pool - ERROR - [-] - Model load failed for 'Qwen3.8-27B-oQ4e-mtp'; caching failure until next discovery refresh
Traceback (most recent call last):
  File "/Applications/oMLX.app/Contents/Resources/omlx/engine_pool.py", line 2492, in _load_engine
    await loop.run_in_executor(
  File ".../concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
[...]

PR #2935, opened 2026-08-20
feat(Qwen35) Fused Qwen MLP/down offload and hardware-adaptive tuner

onthehub97:
# Fused Qwen MLP/down offload and hardware-adaptive tuner

## Summary

This PR extends the experimental Qwen3.5/3.6/3.8 prefill path with a fused MLP/down-projection topology for dual-ANE systems. Instead of returning gate/up activations to the GPU before the down projection, each ANE—and the optional CPU branch—now performs gate projection, up projection, SwiGLU, and its corresponding partial down projection as one branch. The GPU processes the remaining hidden channels concurrently, and the native bridge sums the complete-output contributions once all branches finish.

It also substantially revises the ANE tuner so that it can select the fused MLP split, CPU share, CPU worker count, and GDN ANE/CPU split using the real native dispatch and end-to-end model measurements. Searches are stateless, more heavily sampled, and report provisional results continuously so the UI remains responsive during the longer calibration.

This branch is intended to be reviewed on top of the CPU-sharing PR. The FP16 checkpoint preparation, memory accounting, shared-resource CPU scheduler, single-ANE CPU support, and general CPU-sharing settings are supplied by that parent branch.

### Fused MLP and down projection

- Adds banked, fixed-shape ANE programs that combine gate/up projection, SwiGLU, and partial down projection.
- Splits the MLP hidden dimension between ANE 0, ANE 1, an optional FP16 CPU branch, and the existing quantized GPU suffix.
- Starts both ANE procedures alongside the GPU suffix and CPU branch, synchronizing only before the final output merge.
- Adds native dual-ANE and dual-ANE-plus-CPU fused primitives, bindings, Metal packing/merge kernels, and eager procedure-bank preparation.
- Routes both quantized GPU suffix projections through the bundled NAX QMM on supported M5 systems. Gate/up and down are selected independently so either projection can retain the classic Metal path when its tile is incompatible.
- Permanently demotes hybrid suffixes to the classic kernel if the NAX metallib or pipeline cannot be loaded, and retains `OMLX_QWEN35_QMM_NAX=0` as an explicit kill switch.
- Keeps the original quantized gate/up and down weights available for GPU fallback and decode.
- Adds model-setting, engine, API, benchmark, WebUI, and SwiftUI plumbing for the fused mode.
- Adds real-weight proof-of-concept tools for comparing output-row splitting, GDN splitting, and the accepted hidden-channel fused design.

### Tuner revisions

- Tunes the fused MLP ANE share and FP16 CPU hidden-channel share together.
- Re-evaluates the three best MLP splits across a fixed CPU worker grid of 6, 8, 10, 12, 14, and 16 workers.
- Tunes GDN ANE and CPU shares independently, then re-tests the three best GDN candidates in interleaved order.
- Uses seven coarse samples and nine interleaved finalist samples per representative candidate.
- Uses medians for ranking and median absolute deviation to break candidates within 0.5% of the fastest result, preferring the lower-resource configuration only after stability is considered.
- Does not seed or rank a run using previously saved calibration values. Every run starts from the same hardware-independent search grid.
- Verifies accelerated finalists with at least three end-to-end prompt measurements.
- Profiles the selected full-model candidate and can spend the runner-up slot on a corrected GDN split when the end-to-end critical path disagrees with representative-kernel calibration.
- Measures an exact 8,192-token prefill region—four complete 2,048-token ANE tiles—rather than including the generation-reserved tail token in the reported model work.
- Publishes provisional best values after each completed MLP, worker, and GDN candidate. Full-model verification also updates after every sample. These values may move while the search gathers more evidence; only the completed recommendation is applied.

### Benchmark and diagnostics

- Extends the Qwen prefill benchmark with ANE-down, fused-down, CPU-down-grid, compiled-layer, and worker-count reporting.
- Ensures profiler refinement reasons about the aggregate width of both ANE branches rather than treating the displayed per-ANE fraction as the total.
- Keeps a full-model runner-up visible so close calibration decisions can be compared using the actual serving path.

### Split hidden channels, not output rows

The rejected output-row experiment assigned only selected final down-projection rows to the ANE. It was fast, but its accuracy was unacceptable because each output row depends on the complete SwiGLU hidden vector.

The fused path instead uses the additive structure of the down projection. If the hidden activation is partitioned into slices, the complete result is the sum of each slice multiplied by the matching columns of the down-projection weight:

```text
down(swiglu(hidden)) =
    down_ane0(swiglu_ane0)
  + down_ane1(swiglu_ane1)
  + down_cpu(swiglu_cpu)
  + down_gpu(swiglu_gpu)
```

Every branch therefore produces a full-width model output from a disjoint hidden-channel slice. The final operation is an elementwise sum, not an approximate reconstruction of missing output rows. The normal ANE numerical characteristics remain, but this design introduces no additional output-row approximation.

### CPU down share is intentionally zero in fused mode

The UI summary for the best run reports `CPU 14%/0%/13%`: CPU MLP / standalone CPU down / CPU GDN. The zero does **not** mean that the CPU MLP slice omits its down projection. In fused mode, the CPU's 14% hidden slice includes its matching down-projection columns in the same native branch. Enabling the separate CPU-down path would duplicate work, so the tuner forces that independent setting to zero.

### Tune representative kernels, then verify the real model

An exhaustive end-to-end search across MLP share, CPU share, worker count, GDN share, and CPU GDN share would require too many model reloads. The tuner therefore uses real representative layer dispatches for the broad search, interleaves the closest candidates to reduce temperature/load ordering bias, and reserves expensive full-model runs for the selected topology and its most uncertain alternative. Full-model profiling gets the final say on GDN balance because GDN and fused MLP timings can interact differently in the complete scheduler path.

### Calibration is deliberately stateless

Saved model settings are not treated as an incumbent or tie-break target. This prevents an old calibration—or a manually entered value—from poisoning a new run. Near-ties are resolved from the current samples using dispersion and resource cost only.

### NAX is selected inside the fused native primitive

The fused operation does not call back through Python between gate/up and down, so routing only the normal Python QMM helper to NAX would leave both fused GPU projections on the classic kernel. The native primitive now resolves the NAX Q4/group-128 pipeline for each suffix projection before encoding it.

The NAX variant uses a 64×64×64 tile. A projection is eligible when its K and N dimensions are multiples of 64; otherwise only that projection uses the established 64×32×64 classic kernel. Failure to load the separately packaged NAX metallib or create its pipeline disables hybrid NAX for the remainder of the process, avoiding repeated exceptions while preserving a working ANE path.

The eager-compilation log now records `gpu_suffix=NAX` or `gpu_suffix=classic Metal`, and the native capability is exposed through the Python bridge for diagnostics.

## Performance

Best observed in-app tuner result on the reference M3 Ultra with the Qwen3.8 27B 4.85-bit checkpoint and a 2,048-token ANE program shape:

| Measurement | Selected configuration | Result |
|---|---|---:|
| GPU-only baseline | GPU | 349.4 prompt tok/s |
| Representative fused MLP | ANE 19% per instance, CPU 14%, GPU 48% | 33.14 ms |
| Worker finalist | 8 CPU workers; down projection fused per hidden branch | 33.04 ms |
| Representative GDN | ANE 45%, CPU 12.5%, GPU 42.5% | 10.31 ms |
| Full-model selected result | MLP 19% per ANE, GDN 45%, CPU MLP/down/GDN 14%/0%/13% | **527.4 prompt tok/s (+50.9%)** |
| Full-model calibration runner-up | Alternative finalist | 523.1 prompt tok/s (+49.7%) |

The selected result is 1.51x the measured GPU-only prompt-processing throughput. These figures are a best observed result on one model and machine, not a universal default; the optimal balance changes with SoC, checkpoint shape, quantization, system load, and thermal state.

Decode remains on the normal GPU path, so the improvement applies to eligible prompt-prefill work rather than token-generation throughput.

### Real-weight fused-layer regression

The final native build was also exercised directly on the same real checkpoint using the selected 19% per-ANE / 14% CPU MLP split. This M3 Ultra cannot execute NAX, so this measurement validates the unchanged classic fallback and the fused numerical result rather than estimating M5 tensor-unit performance.

| Path | Median layer latency | Relative throughput |
|---|---:|---:|
| GPU-only Q4 MLP | 60.98 ms | 1.00× |
| Fused ANE/CPU/GPU MLP/down | 32.72 ms | **1.86×** |

The fused result had cosine similarity **0.9999827** against the GPU-only output, RMSE 0.00359, maximum absolute difference 0.03906, and zero NaN/Inf values. The 32.72 ms measurement is consistent with the 33.04 ms representative result selected by the in-app tuner.

## Validation

- `285 passed` across the focused ANE tuner, Qwen prefill, engine-pool, model-settings, and NAX suites:

  ```text
  tests/test_ane_tuning.py
  tests/test_qwen35_ane_prefill.py
  tests/test_engine_pool.py
  tests/test_admin_model_settings_template.py
  tests/test_nax.py
  ```

- Added focused coverage for:
  - fused per-ANE fraction semantics;
  - aggregate dual-ANE profile refinement;
  - stateless CPU worker search;
  - calibration using the worker count actually selected;
  - interleaved GDN runner-up selection;
  - full-model GDN correction taking priority over a worker-only runner-up;
  - exact four-tile prompt measurement;
  - the compatible fused CPU hidden branch;
  - mandatory MLP execution not being masked by successful GDN dispatch.
- Completed a full Release application build with custom classic and NAX kernels enabled.
- Verified the resulting app with `codesign --verify --deep --strict`.
- Verified that the packaged Python runtime exposes the fused dual-ANE/CPU Q4 SwiGLU/down native symbol.
- Verified that the packaged runtime exposes the hybrid-NAX capability probe, contains the NAX metallib, and selects the classic fallback on the M3 Ultra where NAX hardware is unavailable.
- Exercised the fused native primitive on real Qwen3.8 27B 4.85-bit weights with seven measured repetitions.
- `ruff` and `git diff --check` pass for the changed code.

## Known limitations

- The new fused MLP/down path currently requires dual ANE, Q4 weights, group size 128, and compatible Qwen3.5/3.6/3.8 dense MLP layers. Unsupported models and quantizations retain the established non-fused path.
- NAX execution requires supported M5 hardware, macOS 26.2 or later, and a build containing the separate NAX metallib. Other systems automatically use the classic Metal suffix.
- The NAX-backed fused suffix compiled and packaged successfully, but its performance could not be measured on the reference M3 Ultra. An M5 end-to-end tuner and real-request A/B is still required.
- CPU participation requires the separately prepared FP16 weight representation from the parent CPU-sharing work and increases resident memory. Existing admission/preflight accounting must remain enabled.
- ANE programs are fixed-shape. Only complete eligible tiles use this path; scheduler tails and ineligible layers remain on the GPU.
- The implementation depends on Apple's private ANE runtime and is therefore experimental and sensitive to OS and hardware changes.
- ANE inference retains the existing numerical differences from the GPU-only quantized path. The tuner optimizes throughput; it does not assess model intelligence or task-level accuracy.
- Tuning is intentionally more thorough than the previous implementation and therefore takes longer. Provisional values can change until full-model verification completes.
- The benchmark result above is not a claim that every prompt will improve by 50.9%; cache boundaries, prompt length, scheduler behavior, and remaining GPU work determine real-request coverage.

## Follow-up work

- Add fused down-projection support for Q5, Q6, and Q8 while retaining the optimized Q4 kernel.
- Tune the NAX suffix tile variant and fused work split on representative M5, M5 Max, and M5 Ultra systems.
- Evaluate a single-ANE fused topology where its program and synchronization costs are competitive.
- Add automated real-weight cosine, output-logit, and task-level quality gates to complement structural/unit validation.
- Reduce temporary-buffer and dense-weight memory costs without reintroducing request-time dequantization.
- Continue investigating asynchronous ANE submission and merge scheduling to reduce host synchronization overhead.


jundot:
This is excellent work. The additive down decomposition is the right design, the real weight regression at 0.9999827 answers the accuracy question properly, and 1.51x over GPU only would be the largest single jump this path has had. The tuner direction (stateless, medians, interleaved finalists) is also where I wanted it to go.

Main has moved quite a bit since your branch point, all in this same area, so here is what landed so the rebase is easy:

| commit | what it does | issue |
|---|---|---|
| a79ce4b3 | Extends the load time warmup to the CPU path. The first request after loading a CPU shared model was collapsing (325 prompt tok/s, 13 decode), and the same cold state was poisoning the tuner: it turned a +3.8% CPU verdict into -32% on my machine | |
| 2a7bac99 | Tuner hardening: the GDN refinement grid is clamped to the checkpoint's structural floor and the recommendation drops GDN when the winning slot provably ran 0 GDN operations; staging disables DFlash; the calibration engine is released before the verify load | #2899 #2914 #2908 |
| dcb317fe | Admission reserves the ANE I/O surfaces, read from the live model dims | #2841 |
| df05710c | Bank compile is now an incremental builder: each fp32 slice is converted as it is prepared and released right away, and the split ladder retries from the stored chunks. Took the 27B compile spike from 36 GB to 4.7 GB | #2781 |
| df3a0c29 | Drain guards on the CPU dispatch error paths, so a BNNS failure cannot unwind with a committed qmm still in flight | |

So my ask: rebase this on current main, I would be grateful. Three things to carry over while you are in there: the fused bank compile should go through the builder pattern too, right now it stages every layer's fp32 at once so the 36 GB class spike comes back for fused programs and the tuner pays it per candidate; the new fused dispatch site wants the same drain guard as the other three; and the measure prompt went back to whole ANE tile coverage, but the non ANE tail chunk was deliberate after the v0.6.2 overstatement, so please keep sequence_length * 2. One small trap I found in the routes change: enabling fused alone with CPU off leaves the default fraction 0.53 above the 0.50 cap and ANE silently disables at load, worth a clamp.

Everything else I will handle as follow ups on my side like the earlier series: admission pricing for the fused states, CPU warm coverage for fused mode, status and UI plumbing.

joshuaswarren:
Re-ran on this branch (`3b4bb6e`) across three Apple Silicon parts. **The fused path reverses the M1 Ultra verdict and improves both single-die parts substantially.** `284 passed` on `test_ane_tuning.py`, `test_qwen35_ane_prefill.py`, `test_nax.py`, `test_engine_pool.py`, `test_qwen35_native_cpu_share.py`, `test_clone_mlx_model_fp16.py`.

## Tuner verdicts, #2892 (merged) vs this branch

Same model each time (`Qwen3.8-27B` mixed 4/5-bit), `dual_ane: false`, seq 2048, advisor parked so nothing else held the GPU.

| part | #2892 GPU-only -> best | #2892 verdict | **#2935 GPU-only -> best** | **#2935 verdict** |
|---|---|---|---|---|
| M1 Ultra, 128 GB | 145.75 -> 137.15 | **disabled**, -5.9% | **150.0 -> 202.3** | **enabled** mlp 0.5 / gdn 0.5, **+34.9%** |
| M1 Max, 64 GB | 96.50 -> 110.70 | enabled 0.6/0.6, +14.7% | **101.1 -> 149.0** | enabled 0.6/0.6, **+47.4%** |
| M2 Max, 96 GB | 51.60 -> 72.30 | enabled 0.4/0.6, +40.1% | **105.25 -> 154.4** | enabled 0.6/0.6, **+46.7%** |

The M1 Ultra flip is the headline: on the merged branch the tuner correctly refused to enable ANE because GPU-only was faster, and on this branch the same machine gains **+34.9%**. Both single-die parts also roughly triple their previous gain. Caveat on my own numbers: the M2 Max GPU-only baseline moved 51.60 -> 105.25 between runs because I raised that host's memory guard in between, so read the within-run deltas, not the cross-run absolutes.

Also worth noting the new `MLP topology calibration` / `CPU worker calibration` / `Full-model uncertainty runner-up` stages, and that the refinement now sometimes lands on a different point than the prediction (M2 Max: predicted 135.00 at 0.45/0.6, refined 154.40 at 0.6/0.6).

## The CPU eligibility gate: confirmed, thank you

@jundot's explanation was exactly right, and it was the q5 layers. Our production checkpoint is **mixed 4/5-bit**, so every CPU candidate was dropped before measurement — that is the 14-of-90 I reported.

I built an FP16 clone of a **uniform q4** checkpoint and the gate passes:

```
engine_pool - CPU sharing adds 11.37GB to the projected memory for <model>
engine_pool - CPU sharing adds 11.63GB to the projected memory for <model>
qwen35_ane_prefill - Qwen ANE CPU sharing using performance-aware scheduling with 8 workers
```

So: FP16 clone + uniform q4 gate/up + compiled symbols = eligible, and a mixed-precision checkpoint containing any q5 gate/up silently disqualifies the whole CPU dimension. That is worth one line in the tuner docs, because from the outside it is indistinguishable from "the tuner does not explore CPU".

## One bug: the reproducibility guard aborts the run instead of falling back

On that q4 FP16 clone the run ended `status: error` at step 94/95 with `recommendation: {"enabled": false, "speedup_percent": 0.0}`, discarding a perfectly good measurement. The sequence was:

| stage | result |
|---|---|
| GPU only | 239.50 |
| Predicted optimum | **285.50** (+19.2%) |
| Profile-refined optimum | `RuntimeError` -> whole run errored |

```
RuntimeError: The ANE compiled but never executed for sequence_length=2048: no ANE
operation was observed during measurement, so this candidate would report GPU-only
throughput as an ANE result. The most common cause is a prefill chunk size mismatch
(the scheduler is configured for 2048-token prefill chunks, so set sequence_length=2048
or smaller ...)
```

The guard itself is good — refusing to report GPU-only throughput as an ANE result is exactly right, and I would rather have this than a false win. Two observations:

1. **A failed refinement should not discard a measured predicted optimum.** The +19.2% candidate at 285.50 measured cleanly; ending at `enabled: false` throws that away and tells the operator ANE is useless on this checkpoint, which is the opposite of what the data showed.
2. This fired on the **q4 clone** but not on the mixed 4/5-bit model on the same machine minutes earlier, with `sequence_length` already 2048 in both cases. So the "chunk size mismatch" hint may be misleading here — something about the clone's path stops ANE executing during refinement specifically. Happy to dig if you want a narrower repro; I have the clone staged.

## Environment

M1 Ultra 128 GB / M1 Max 64 GB / M2 Max 96 GB, all macOS 26.5-26.6, Xcode 26.6 (17F113), `mlx==0.32.0`, `nanobind==2.13.0`, kernels built with `OMLX_WITH_CUSTOM_KERNEL=1`. One note for other testers on older Macs: the Xcode **Metal toolchain component** is a separate 688 MB download (`xcodebuild -downloadComponent MetalToolchain`), and on one of my machines `metal` still refuses to execute after installing it, so I built the kernels on another host and copied the `.dylib`, `_ext.so` and `.metallib` files across — same Xcode, macOS, mlx and Python, and the symbols resolve fine.

joshuaswarren:
Followed @jundot's pointer and re-ran on **current `main` (`fa3e94b`, 0.6.3rc2)** rather than the branch. Both of the problems I reported above are gone there, and the CPU dimension finally wins — so the rebase looks well worth it.

Same machine as my post above (M1 Ultra, 128 GB), same FP16 clone of a **uniform q4** checkpoint, `dual_ane: false`, seq 2048, advisor parked. `283 passed`.

| | #2892 (merged) | #2935 branch `3b4bb6e` | **main `fa3e94b`** |
|---|---|---|---|
| steps measured | 14 / 90 | 94 / 95 then `error` | **90 / 90 completed** |
| CPU candidates | all dropped at the gate | measured, run aborted | **measured and selected** |
| verdict | `enabled: false` | `enabled: false` (from the abort) | **`enabled: true`** |

```
GPU only                    236.85
Predicted optimum           260.50   ane 0.4  / gdn 0.4  / cpu 0.135
Profile-refined optimum     267.10   ane 0.35 / gdn 0.4  / cpu 0.15
recommendation: enabled=true, cpu_enabled=true, cpu_fraction 0.15,
                cpu_down_fraction 0.25, cpu_gdn_fraction 0.15,
                cpu_shared_resource=true, +12.77%
```

This is the first time I have seen the tuner select a CPU share at all, so `a79ce4b3` (CPU-path load-time warmup) plus `2a7bac99` (tuner hardening) appear to fix exactly what I hit:

1. **The `RuntimeError: The ANE compiled but never executed` abort is gone**, and the run reaches 90/90 instead of dying at refinement and reporting `enabled: false`. Whatever the hardening does about "drops GDN when the winning slot provably ran 0 GDN operations" seems to cover the case I ran into.
2. **The CPU dimension is genuinely explored** once the cold-state collapse is fixed. Your note that cold state turned a +3.8% CPU verdict into -32% matches what I was seeing indirectly: on my machine the CPU candidates either never measured or dragged the result down.

One data point that may be useful for your admission-pricing follow-up: with CPU sharing selected, the load line reports `CPU sharing adds 15.96GB` / `16.69GB` to the projected memory on main, against 11.37 / 11.63 GB on the branch for the same checkpoint and settings — so `dcb317fe` reserving the ANE I/O surfaces is visibly moving the projection, which is the safe direction.

I have not re-run the mixed 4/5-bit case on main yet (my poller timed out and killed the serve before it wrote results). Happy to add that, and to re-run all three parts once the rebase lands — the branch numbers I posted above (M1 Ultra -5.9% -> +34.9%, M1 Max +47.4%, M2 Max +46.7%) were all pre-rebase, so they should be treated as a floor rather than the final picture.

beaglemoo:
Ran this branch (`3b4bb6e`, as-is, pre-rebase) on the **M5 Pro (12-core, 64 GB)** — the NAX data point, since the three parts tested so far are all pre-tensor-unit GPUs. Short version: **the fused topology inverts here, hard, and the new tuner correctly refuses it.** The hardware-adaptive direction of this PR is doing exactly what it should on hardware where the answer is "no".

## Environment

M5 Pro 64 GB, macOS 26.x, `mlx==0.32.0`, `nanobind==2.13.0`, kernels rebuilt with `OMLX_WITH_CUSTOM_KERNEL=1`. Model is our production `Qwen3.8-27B` oQ4e (uniform q4 gate/up, some q5 down/attention rows), cloned to FP16 scales with `tools/clone_mlx_model_fp16.py` so the CPU dimension was searchable — the clone passed the eligibility gate (`resident_programs` populated, no "requires an FP16 checkpoint clone" warning). `dual_ane: true`, `sequence_length: 2048`. Nothing else on the machine; fresh server process per measurement, isolated `OMLX_BASE_PATH`, prompt cache cleared per run.

## Tuner verdict (68 steps, ~5 min, no errors)

The reproducibility-guard abort @joshuaswarren hit did not fire here — the run completed cleanly on the FP16 clone.

| stage | result |
|---|---|
| GPU only | 440.25 tps |
| MLP topology calibration | ANE 35% · **CPU 0%** · classic (fused not selected) |
| CPU worker calibration | **CPU 0% · GPU 100%** |
| GDN calibration | ANE 45% · CPU 0% |
| Predicted optimum (0.35/0.45) | 465.5 tps (+5.7%) |
| Profile-refined optimum (**0.40/0.40**) | **485.1 tps (+10.2%)** |

So on NAX the tuner lands on the classic split at +10%, with fused and CPU both rejected — vs the +35–47% fused verdicts on M1 Ultra / M1 Max / M2 Max. Same shape as the earlier ANE-fraction story (#2779, #2853): the optimal ANE share shrinks as GPU width/tensor-unit capability grows.

## End-to-end check: the tuner is right, and forcing fused is much worse than "slower"

Warmed real requests (preload + same-size warmup with a different seed before timing), TTFT on a 15,737-token prompt, two mirrored rounds:

| config | r1 | r2 |
|---|---:|---:|
| ANE off | 39.30 s | 44.44 s |
| classic 0.35/0.45 | 36.57 s | 34.79 s |
| classic 0.40/0.40 | 35.71 s | 35.57 s |

(ANE-on is also thermally stable across the session while GPU-only drifts +5 s — same pattern we reported on #2853.)

Forcing `qwen35_ane_prefill_fused_down: true` at 0.40/0.40 (CPU off, so under the 0.50 cap):

- **Fused engages correctly**: `Warmed 352 ANE procedures in 7.0s at load`, `Eagerly compiled 64 MLP and 48 GDN procedures into 4 instance-pinned ANE programs`, `resident_programs: 4`.
- **2,080-token prompt: 31.35 s TTFT vs 4.70 s classic** on the identical prompt/protocol — **6.7x slower**.
- **15,737-token prompt: a single request had not completed after ~33 minutes** and was aborted. Worse than linear extrapolation from the 2K number, so something beyond per-op overhead is going on at multi-chunk depth — possibly relevant to the drain-guard gap on the fused dispatch site @jundot noted in review.
- Load-time memory: `actual: 40.96GB` for a 16.6 GB model — the fused fp32 staging spike from the rebase list, confirmed on a 64 GB machine (fine standalone, but would collide with anything else resident).

## Takeaways from the NAX corner

1. The tuner's topology calibration is trustworthy here — it picked classic over fused from microbenchmarks alone, and end-to-end confirms it. Nice work.
2. Given force-enabling fused on NAX yields a 6.7x regression and a request that effectively never finishes at 16K, a runtime guard (or at least a log warning) when `fused_down` is enabled on NAX hardware might be worth carrying in the rebase, in the spirit of the #2779 auto-skip.
3. CPU 0% on this part matches the single-die M5 expectation from #2892; the FP16-clone + eligibility flow itself worked exactly as documented.

Server logs (fused 2K and stalled 16K runs, classic comparators, GPU-only baseline) in the round 7 files at https://gist.github.com/beaglemoo/aae222168f44f01f43758b83bad9e4e9 (`round7-2935-*`: fused 2K + stalled 16K runs, classic 2K/16K comparators, GPU-only baseline).

Happy to re-run after the rebase — especially if the fused dispatch picks up the drain guard, in case the 16K stall is that.

onthehub97:
@jundot I have fixed the merge conflicts. Can rebase if required but squashing will work fine here.

jundot:
Merged. Thanks @onthehub97 for carrying this through the whole review cycle, and thanks @joshuaswarren and @beaglemoo for running it on four different parts. The M1 Ultra flip and the M5 Pro refusal are exactly the hardware-adaptive behavior this PR promised.

On the measurement length, I originally asked to keep the shorter measured prompt, but the tail padding work in #2966 prices a padded tile from the hybrid throughput, so measuring in the pure tile regime is the right call and I am keeping your version.

beaglemoo:
Re-ran the M5 Pro numbers on merged main (`0436bdf2`) with the follow-up series in — closing the loop on the stall I reported:

**The fused collapse is fixed.** Identical config to my earlier report (`fused_down: true`, 0.40/0.40, dual-ANE, FP16 clone): the 2,048-token TTFT went **31.35 s -> 5.02 s**, and the 16K non-completion no longer reproduces. So the earlier behavior was indeed on the follow-up list — the fused dispatch guard (`e200e0d5`) and/or the incremental fused staging (`c267fc65`); load memory is also back to sane with the builder in place.

With fused now *working*, the M5 verdict simply becomes the tuner's: fused measures ~7% behind classic here (5.02 s vs 4.70 s at 2K), so the topology calibration's refusal stands on merit rather than on a hang.

Two smaller data points from the same session:

- The revised tuner on main re-recommends exactly the classic split (`0.35/0.45`, +10.05%, 436.6 -> 480.5 tps) and calibrates `tail_padding_min_tokens: 1861` on this part — first NAX data for #2966.
- The follow-up clamp (`0436bdf2`) is a good catch: my earlier forced-fused runs only avoided the silent-disable because I set the fraction explicitly.

Sanitized serve log for the fused re-probe at https://gist.github.com/beaglemoo/aae222168f44f01f43758b83bad9e4e9 (`round8-2935-fused-fixed-2k-server.log`). Nothing further owed from this corner — thanks for the fast turnaround on the guard.

reneleonhardt:
Just wow! 🚀

PR #2937, opened 2026-08-20
perf: extend Qwen3.5/3.8 prefill acceleration to the DFlash target

isaac-cf-wong:
## Problem

`BatchedEngine` and `VLMBatchedEngine` install a stack of Qwen3.5/3.6/3.8 prefill
accelerations before serving: the native q4/q8 MLP and prefill-linear tiles, the FA-256
steel attention route, and the private ANE/GPU prefill split. `DFlashEngine` installed
none of them, so a DFlash-served target prefilled with an unaccelerated model — 76 tok/s
on an 8K prompt against 105-108 tok/s for the same checkpoint on the batched path.

## Change

Apply the stack to the DFlash target as well, in `omlx/patches/dflash_qwen35_accel.py`
with two call sites in `omlx/engine/dflash.py`.

Gating follows the batched engines exactly: `qwen35_q4_mlp_prefill_enabled`,
`fa256_steel_prefill_enabled`, and `qwen35_ane_prefill_enabled` (default off).
`OMLX_DFLASH_QWEN35_ACCEL=0` keeps a DFlash load off the stack entirely.

Installation is split by lifetime:

- **class/module-level patches** install *before* `load_target_bundle` and before
  dflash's own class hooks, so dflash's lifecycle snapshot treats them as pre-existing
  state instead of restoring over them (which would drop the wrappers while the patch
  modules still reported themselves installed);
- **the ANE split** runs after the load, on the MLX executor thread, since it needs
  model instances and compiles procedures.

Only patches a DFlash target can execute are installed. dflash drives Qwen GDN targets
through mlx-lm (`dflash_mlx/engine/target_qwen_gdn.py` imports `mlx_lm.models.*`), so
this installs `qwen35_q4_mlp`, the mlx-lm prefill-linear patch that `BatchedEngine` also
installs, and the FA-256 route. `qwen35_gdn_prework`, `qwen35_gdn_chunked`,
`qwen35_verify_sdpa_split`, `qwen35_ragged_decode` and the mlx-vlm prefill-linear
sibling are deliberately **not** installed: they bind `mlx_vlm` seams that dflash never
runs, and `VLMBatchedEngine` installs that set itself when dflash falls back to it.

These are not bit-exact fast paths — the native tiles reorder accumulation and the ANE
prefix is per-output-channel INT8, so greedy output can drift from the unpatched path in
long generations (measured below). This is the behaviour the batched engines already ship
for these patches; the ANE route additionally logs a warning when it engages on a DFlash
target.

## Tests

`tests/test_dflash_qwen35_accel.py` — 16 tests, no model load:

- kill switch (`0`/`false`/`off`) installs nothing;
- per-model settings disable individual patches; absent attributes mean enabled,
  matching `BatchedEngine`'s `getattr` default;
- ANE argument forwarding (`sequence_length`, `fraction`, `gdn_fraction`, `dual_ane`),
  the warning when the compiled shape exceeds the dflash prefill step, and the
  approximate-route warning;
- a patch or ANE enable that raises degrades to `False`/0 instead of breaking load;
- a guard that fails if any mlx-vlm-bound patch is ever installed from the DFlash path.

Neighbouring suites stay green: `test_dflash_*` and `test_qwen35_*` → 190 passed,
14 skipped.

## Benchmark

`Jundot/Qwen3.8-27B-oQ4e-mtp` + `z-lab/Qwen3.8-27B-DFlash2`, w4:gs64 draft quant, verify
`adaptive`, greedy, Apple M4 Pro / 64 GB, built `--with-custom-kernel`, ANE seq 2048 /
MLP 0.60 / GDN 0.60 / `dual_ane` off. Cold prompts with a unique preamble so the prefix
cache cannot serve them; medians of 3.

| 8K cold prefill | tok/s | vs before |
|---|---|---|
| before (kill switch) | 75.9-76.2 (75.6-76.8) | — |
| after, class patches only (`qwen35_ane_prefill_enabled=false`) | 83.0 (82.6-84.0) | **+9%** |
| after, full stack | **105.2 (104.1-105.8)** | **+38%** |

| measurement | before | after |
|---|---|---|
| 16K cold prefill | 73.6 / 73.6 | 95.9 / 95.3 |
| TTFT, 8K prompt | ~106 s | ~77 s |
| decode, 144-tok prompt, 512 out | 18.35 (18.34-18.45) | 17.75-17.78 |
| decode, 8K prompt, 256 out | 14.15 (13.55-14.60) | 14.45 (14.15-15.37) |
| acceptance, 8K | 70.7 / 67.0 / 66.2% | 68.0 / 68.7 / 70.1% |
| peak memory, 8K | ~20.6 GB | ~22.5 GB (ANE banks) |

For reference, the same checkpoint on the batched path with ANE reaches 105-108 tok/s,
so DFlash prefill now matches it instead of trailing by ~26%.

Acceleration survives model reload — three unload/load cycles held 96.9 / 96.0 / 96.1
tok/s with ANE and 81.2 / 80.7 / 80.8 with class patches only.

Greedy-output divergence from the unpatched path, 2960-token prompt, 200 greedy tokens:
class-patches-only diverges 40% into the output, the full stack 95% in, continuations
semantically equivalent. A 128-token probe showed no divergence, so short hash checks do
not establish equality. Task-level quality was not evaluated.

### DIFF
--- a/omlx/engine/dflash.py
+++ b/omlx/engine/dflash.py
@@ -562,6 +562,23 @@ class DFlashEngine(ActivityTrackingMixin, BaseEngine):
 
             install_dflash_lifecycle_wrap()
 
+            # Install the Qwen3.5/3.8 class-level prefill patches before the
+            # target loads and before dflash's own class hooks, so dflash's
+            # lifecycle snapshot treats them as pre-existing state instead of
+            # restoring over them (which would drop them while the patch
+            # modules still report themselves installed).
+            try:
+                from ..patches.dflash_qwen35_accel import (
+                    install_dflash_qwen35_class_patches,
+                )
+
+                install_dflash_qwen35_class_patches(self._model_settings)
+            except Exception:
+                logger.debug(
+                    "DFlash Qwen class-level prefill patches not applied",
+                    exc_info=True,
+                )
+
             target_bundle = load_target_bundle(
                 self._model_name,
                 quantize_kv_cache=bool(
@@ -593,6 +610,26 @@ class DFlashEngine(ActivityTrackingMixin, BaseEngine):
                         "DFlash target MoE gate+up fusion not applied",
                         exc_info=True,
                     )
+
+            # Instance-level half of the Qwen acceleration stack: the ANE/GPU
+            # prefill split needs the loaded model and must run on the MLX
+            # executor thread, as the batched engines require.
+            try:
+                from ..patches.dflash_qwen35_accel import (
+                    enable_dflash_qwen35_ane,
+                )
+
+                enable_dflash_qwen35_ane(
+                    target_bundle.model,
+                    self._model_settings,
+                    prefill_step_size=int(
+                        getattr(runtime_context.runtime, "prefill_step_size", 2048)
+                        or 2048
+                    ),
+                )
+            except Exception:
+                logger.debug("DFlash Qwen ANE prefill not applied", exc_info=True)
+
             draft, draft_meta = load_draft_bundle(
                 self._draft_model_path,
                 draft_quant=(
--- a/omlx/patches/dflash_qwen35_accel.py
+++ b/omlx/patches/dflash_qwen35_accel.py
@@ -1,0 +1,225 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Extend oMLX's Qwen3.5/3.6/3.8 acceleration paths to the DFlash target model.
+
+``BatchedEngine`` and ``VLMBatchedEngine`` install a stack of Qwen-specific
+prefill accelerations before serving (native q4/q8 MLP and prefill linear tiles,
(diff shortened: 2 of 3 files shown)


jundot:
Thanks for this, the install ordering is right and the lifecycle interaction checks out end to end. The one problem is timing: the ANE stack on main is moving fast right now, since your branch point we landed tuner v2 (#2891), CPU sharing (#2892) and a series of follow-ups, with more in flight (#2905, #2935). As is, the DFlash path silently drops the qwen35_ane_prefill_cpu_* settings the tuner writes, so a tuned profile would get half applied. I'd rather not have you chase main commit by commit, so let's hold this until the ANE side settles, then rebase and mirror the batched engine's ANE forwarding including the cpu params. Everything else here looks good to me.

isaac-cf-wong:
No problem!

Issue #2940, opened 2026-08-20
ANE Program Inference error at mlp_fraction 0.40 on main; runtime failure latch does not recover — all requests fail


beaglemoo:
**Environment:** M5 Pro, 64 GB, macOS 26.x, main `c6cb4b6f`, source install, kernels built with `OMLX_WITH_CUSTOM_KERNEL=1`, `Qwen3.8-27B-oQ4e-mtp`, `sequence_length 2048`, `dual_ane true`.

**Repro:** set `qwen35_ane_prefill_fraction: 0.40, gdn_fraction: 0.25` (the pre-#2899 tuner recommendation; note gdn 0.25 is sub-floor so this is effectively MLP-only, 64 MLP / 0 GDN), load the model, send any request. 2/2 runs failed, fresh server each.

**What happens:** load-time warmup hits `ANEProgramProcessRequestDirect() Failed with status=0x2 : statusType=0x9: Program Inference error` partway through the bank (procedure 55 of 128 in one run, 69 in the other — the index varies, the failure doesn't). The latch warns "continuing, the runtime failure latch handles broken procedures at first use", but then **every** request fails with `Prior ANE evaluation failed` — the server never degrades to GPU, it just errors on all inference.

```
2026-08-20 15:41:20,684 - omlx.patches.qwen35_ane_prefill - WARNING - ANE warmup failed after 55 procedures; continuing, the runtime failure latch handles broken procedures at first use
2026-08-20 15:41:20,685 - omlx.patches.qwen35_ane_prefill - INFO - Warmed 55 ANE procedures in 1.5s at load
2026-08-20 15:41:20,686 - omlx.patches.qwen35_ane_prefill - INFO - Eagerly compiled 64 MLP and 0 GDN procedures into 2 instance-pinned ANE programs (sequence_length=2048)
2026-08-20 15:41:20,686 - omlx.patches.qwen35_ane_prefill - INFO - Qwen ANE prefill preserving scheduler chunks; projection tile=2048 (step=2048, floor=4096)
2026-08-20 15:41:25,461 - omlx.scheduler - ERROR - Prefill failed for 03b42dd1-f322-49b2-9035-8909dc6ffdde: Prior ANE evaluation failed: ANE evaluation failed: Error Domain=com.apple.appleneuralengine Code=8 "processRequest:model:qos:qIndex:modelStringID:options:returnValue:error:: ANEProgramProcessRequestDirect() Failed with status=0x2 : statusType=0x9: Program Inference error"
[...]

PR #2946, opened 2026-08-20
fix(ane): latch the owning module when load-time warmup fails, keep warming the rest

beaglemoo:
Follow-up to #2898's hardening, addressing the failure mode in #2940.

**Problem.** When a procedure's load-time warmup fails, the outer try keeps the load alive, but the procedure's sticky error re-raises at *evaluation* time — inside the scheduler, where no per-module latch can intercept it (the dispatch-site latches only cover graph construction). Result: every request fails with `Prior ANE evaluation failed` instead of falling back to GPU. Observed on an M5 Pro as a rare, context-dependent ANE `Program Inference error` (3 occurrences in ~23 loads, one heavy session; details in #2940) — rare enough that graceful degradation is the whole game.

**Fix.** Warm per-procedure instead of in one guarded loop, and on failure latch the *owning module* (`_omlx_ane_prefill_failed` / `_omlx_ane_gdn_failed`) at the only point the failure is still attributable to it — then keep warming the remaining procedures. One broken procedure now costs one module its ANE path; the rest of the bank stays warm and active.

**Verification.**
- All 97 tests in `test_qwen35_ane_prefill.py` pass, including a new `test_warmup_failure_latches_only_the_owning_module` (one failing procedure → only its module latched, siblings keep warming) and the existing `test_enable_survives_warmup_failure` unchanged.
- End-to-end on the M5 Pro with a temporarily injected fault at procedure 5: log shows `Disabled ANE on 1 of 64 modules after warmup failures; they fall back to GPU`, the other 126 procedures warmed in 2.7s, and a 16K request completed at full ANE-on speed (37.6s TTFT, ANE active on the remaining 63 modules) with zero `Prior ANE evaluation failed` errors — versus the total request-path outage in #2940.

Fixes the (b) half of #2940. The underlying `Program Inference error` race — (a) — remains open there; with this change it degrades one module instead of downing the server.

### DIFF
--- a/omlx/patches/qwen35_ane_prefill.py
+++ b/omlx/patches/qwen35_ane_prefill.py
@@ -2034,24 +2034,43 @@ def _enable_dual_procedure_banks(
         # than ANE warmup. Guarded per-model so an older compiled extension
         # without warmup() degrades to the previous behavior instead of
         # failing the load.
-        # Warmup failures are soft: the procedures are already registered, so
-        # a broken one is disabled by the runtime failure latch at first
-        # dispatch instead of aborting the whole ANE setup here.
+        # A warmup failure latches ANE off for the owning module right here.
+        # The per-module flag is checked at graph construction, which is the
+        # only place the failure can still be intercepted: by evaluation time
+        # the sticky per-procedure error re-raises inside the scheduler and
+        # fails every request instead of falling back (#2940). Remaining
+        # procedures keep warming, so one broken procedure costs one module
+        # its ANE path rather than taking down the request path.
         warm_start = time.perf_counter()
         warmed = 0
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for tracking this down. I confirmed that latching the owning module during warmup closes the evaluation-time failure gap, while the remaining procedures stay warm and the failed module falls back to GPU. The focused ANE test suite passes locally. LGTM, merging.

beaglemoo:
Thanks for the fast turnaround — with this in, the #2940 race degrades to a one-module GPU fallback, so the remaining root-cause hunt is uncontaminated by outages. Happy to run any diagnostics on this machine if it resurfaces.

PR #2953, opened 2026-08-20
docs: document coordinator-loss semantics

### LINKED ISSUE #2682: Document coordinator-loss semantics: no shard rebalance, ranks exit, recovery is manual

When a coordinator becomes unavailable, a distributed deployment stops entirely. Shards are not rebalanced, surviving nodes do not take over, and the model becomes unservable until an operator re-activates it manually.

This is currently undocumented, so the first time an operator meets it is in production. Searching the codebase for recovery machinery returns nothing:

```
rebalance | failover | redistribute | reshard | degraded | self-heal   -> 0 matches
resume | restart | recover | relaunch (functions)                      -> 0 matches
```

What happens today, traced through the source:

1. The MLX launcher dies with the coordinator -- it is the coordinator's child
2. Every remote rank's parent-PID watchdog fires, writes a `launcher_lost` marker and calls `os._exit(1)` (`inference_worker.py`)
3. Each rank's shard weights are freed with its process; no node ever held the full model
4. The deployment record persists in `~/.omlx/cluster/deployments.json`, but nothing is running
5. Each node's own HTTP server is unaffected and keeps serving whatever it can fit alone

I am not proposing automatic failover. The fail-stop behaviour looks like the correct response to the constraint, and the reasons are worth stating explicitly rather than being rediscovered:

- `mx.distributed` has no membership-change protocol. A fixed-world-size collective cannot lose a rank, only deadlock -- which is why the code already warns that "launching into a dead peer produces a collective that blocks forever rather than an error"
- The plan is content-addressed (`plan_hash` is a SHA-256) and activation 409s if the recomputed placement differs from `approved_placement`. Silent auto-rebalancing would defeat that guard
- Shards are bound to rank index via `PipelineAssignment`, and weights are physically resident in each rank's unified memory -- moving one is a multi-GB transfer plus a reload, not a pointer change

What would help, in rough order of value:

1. **Document the failure semantics** -- one short section stating that coordinator loss ends the deployment, that ranks exit rather than hang, and that recovery is a manual re-activation. This is the cheapest and highest-value item
2. **Surface it** -- an event or status the dashboard can show when ranks exit with `launcher_lost`, so the state is visible without reading a log
3. **Assisted recovery** -- given the deployment record already persists, a "re-activate this deployment" action that replays the stored plan, still behind the existing approval gate
4. **Automatic failover** -- only worth considering if MLX gains membership changes; noting it here for completeness rather than proposing it

Leaving it undocumented. That is the current state, and it is how I found it -- by reasoning about what would happen and then reading the source to check, rather than by finding it written anywhere.

**Additional context**

Related: the non-coordinator silent-answer behaviour filed separately. Together they mean a client can lose its distributed model and keep receiving confident single-node answers with nothing to indicate the change.

- macOS Version: 26.5.1
- oMLX Version: 0.6.0.dev1

**Acceptance criteria** (for item 1, the documentation)

- [ ] Documented that coordinator loss ends the deployment and does not rebalance
- [ ] Documented that ranks exit rather than block, and why the alternative would hang
- [ ] Documented that recovery is a manual re-activation
- [ ] Documented that each node continues to serve its own models independently afterwards

hellodk:
Closes #2682


hellodk:
Closing as a duplicate of #2941 — both close #2682.

#2941 is the reviewable version: 2 files, 150 lines, just the docs. This branch was pushed from a local integration branch and carries the whole in-flight tree (51 files, ~8.2k lines) — five unrelated features on top of the docs change.

PR #2960, opened 2026-08-20
perf(server): memoize chat-template render and prompt encode

KostovMartin:
One streaming chat request renders the same conversation up to three times (admission preflight, thinking-state detection, the engine's own render) and BPE-encodes the rendered prompt at least twice (preflight count, request tokenization). Each render is a full Jinja pass and each encode a full BPE pass — for a 30K-token agentic prompt that is repeated work inside TTFT, after the KV prefix cache has already removed the prefill.

The memo is exact-key (SHA-256 over the serialized inputs, small LRU): a hit returns byte-identical output, so no caller can observe a difference; non-serializable payloads (VLM image handles) fall through uncached. Memory is hard-bounded: 8 entries per memo, prompts beyond 4M chars skip the render memo, and encodes beyond 131 072 tokens skip the token memo — a caching layer must not become the resident-set problem it saves time on. The encode memo is scoped by an id stamped on the tokenizer at load, **before** `Scheduler.__init__` deep-copies it for "Already borrowed" isolation — object-identity keys can never match across that copy (a test simulates exactly this pairing).

Measured with an interleaved A/B (4 server sessions, the two branches differ only by this commit, 5 runs per size, M3 Ultra, DeepSeek-V4-Flash):

| Prompt tokens | Warm TTFT before | Warm TTFT after | Saving |
|---|---|---|---|
| 6,873 | 1.57s | 1.55s | 16ms |
| 27,432 | 1.90s | 1.83s | 75ms |
| 47,703 | 1.77s | 1.63s | 143ms |
| 75,308 | 3.92s | 3.68s | 242ms |
| 108,770 | 1.62s | 1.29s | 330ms |
| 111,060 (322 messages) | 2.17s | 1.82s | 353ms |
| 133,866 (above encode cap) | 2.90s | 2.92s | ~0 |

Scales linearly at roughly 3us/token up to the 131k-token encode cap. No change to cold prompts or decode. The benchmark repeats identical bytes, so unique traffic keeps about half of the curve. Tests cover key determinism, LRU bounds, copy isolation, the deepcopy scope, and the engine wrapper's hit contract (single tokenizer call, identical prompt, `partial`-key stripping preserved on hits).

Based on #2959 — both touch `scheduler.py` in different regions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,12 @@ Block-based KV cache management inspired by vLLM, with prefix sharing and Copy-o
 
 Handles concurrent requests through mlx-lm's BatchGenerator. Max concurrent requests is configurable via CLI or admin panel.
 
+### Diagnostics & Tuning Environment Variables
+
+- `OMLX_STEP_PROFILE=1` — log a periodic wall-time profile of the serving
+  hot path (decode, prefill, scheduling, cache bookkeeping, SSE emission)
+  as one summary line per window; `OMLX_STEP_PROFILE_EVERY=N` sets the
+  window length in scheduler steps (default 500). No overhead when unset.
 ### Claude Code Optimization
 
 Context scaling support for running smaller context models with Claude Code. Scales reported token counts so that auto-compact triggers at the right timing, and SSE keep-alive prevents read timeouts during long prefill.
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -6,14 +6,17 @@ This engine wraps AsyncEngineCore to provide continuous batching
 for better throughput when serving multiple concurrent requests.
 """
 
+import contextlib
 import copy
 import logging
+import uuid
 from collections.abc import AsyncIterator
 from typing import Any
 
 from ..api.tool_calling import convert_tools_for_template
 from ..api.utils import clean_special_tokens, detect_and_strip_partial
 from ..reasoning_effort import apply_chat_template_with_reasoning_effort_fallback
+from ..render_cache import RENDER_MEMO_MAX_CHARS, RenderMemo, encode_cached
 from ..utils.tokenizer import get_tokenizer_config
 from .base import (
     BaseEngine,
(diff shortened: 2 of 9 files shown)


joshuaswarren:
Measured the two passes on a real agentic-sized conversation, in case a concrete number helps size this. Tokenizer only — no server, no GPU — on a `Qwen3.8-27B` mixed 4/5-bit tokenizer, M1 Ultra, `0.6.3rc2`, conversation built to **57,990 tokens / 285,847 chars**, median of 5:

| pass | median |
|---|---:|
| `apply_chat_template` (render) | **0.1 ms** |
| BPE `encode` of the rendered prompt | **73.3 ms** |
| your duplicate work (2 extra renders + 1 extra encode) | **73.5 ms** |
| all passes (3 renders + 2 encodes) | 146.9 ms |

Two things fall out of that split:

1. **The render is effectively free at this size and the encode is the whole cost.** 0.1 ms vs 73.3 ms, so the memo's value is almost entirely the BPE pass; if it ever needs to be cheaper to maintain, memoizing only the encode would capture ~99% of the benefit.
2. **It matters exactly where you said it does — the prefix-cache-hit path.** On a cold 58k prompt our own prefill runs into the hundreds of seconds, so 73 ms is noise (~0.02%). But on a cache hit, where prefill is gone and TTFT is short, 73 ms of duplicate encode is a large fraction of what the user actually waits for. That is worth stating explicitly in the PR description, because someone benchmarking cold prompts will measure nothing and wrongly conclude the change does not help.

Happy to measure the end-to-end TTFT delta on a warm prefix once this lands, if useful — we run 49-63k-token agentic prompts against this model continuously, so the cache-hit case is our normal traffic rather than a synthetic one.

KostovMartin:
Good catch, you're right on the magnitude. We reran this as a proper interleaved A/B (4 server sessions, the two branches differ only by the memo commit, 5 runs per size). Warm TTFT saving: 16ms at 7K, 75ms at 27K, 143ms at 48K, 242ms at 75K, ~340ms at 109K tokens. Linear at roughly 3us/token, same result for a 322-message conversation as for one giant message. Render is basically free here too, encode is the whole cost.

So the -31% headline was confounded by session differences. The description now carries the corrected table. Note the A/B repeats identical bytes, so unique traffic keeps about half of the curve.

The sweep also showed the encode memo cap (131k tokens) zeroes the benefit above it, with no regression. Might be worth raising later since the benefit peaks right at the cap.

PR #2964, opened 2026-08-20
Qwen35 gdn output full attention projections

onthehub97:
TODO

To be merged after fused PR


onthehub97:
Not convinced this PR is worth merging. The time it adds to the tuner is so much massively more than the PP time it would save on millions of tokens that I don't think it's worth it.

PR #2965, opened 2026-08-20
perf(ane): cover the prefill tail with an overlapping fixed-shape tile

isaac-cf-wong:
## Problem

A wide prefill splits into full fixed-shape ANE tiles plus a residual tail that runs on
the ordinary quantized linears. The tail is pure GPU work, so it caps the achievable
prefill rate: at `sequence_length` 1024 a ~4.1K-token prompt leaves a 1018-row tail.

## Change

Cover the tail by re-running the last `sequence_length` rows as a full ANE tile and
keeping only its last `tail_rows` outputs.

Both accelerated backends are row-wise — `_backend_exact` is a fused gate/up projection,
row-wise SwiGLU and down projection; `_gdn_backend_exact` stops at `in_proj_qkv/z/b/a`,
with the convolution and recurrent update happening outside it — so recomputing rows an
earlier tile already produced is redundant work, not incorrect work. This is **not** the
padding approach this module rejects elsewhere: no synthetic tokens enter the prompt and
no KV or recurrent state advances; only existing rows are re-projected.

The trade is one full tile (`sequence_length / ane_rate`) against the tail on GPU
(`tail_rows / gpu_rate`), so it pays when

```
tail_rows > sequence_length * gpu_rate / ane_rate
```

about 720 rows for a 1024 shape on an M4 Pro (ANE-covered rows ~118 tok/s, GPU rows ~83).
Gated by `OMLX_QWEN35_ANE_TAIL_OVERLAP_MIN`, **default 0 (off)**. The constant is a
per-machine heuristic — CPU sharing, dual-ANE mode, quantization and the compiled shape
all move the break-even — and is a candidate for derivation from the split tuner rather
than a default. The threshold is resolved once and cached, because it is consulted on
every accelerated MLP and GDN call of every layer and reading the environment per call
measured as a ~1.6% prefill regression on short prompts.

## Tests

`tests/test_qwen35_ane_overlap_tail.py` — 26 tests, no model load, new file. Both legs are
stubbed with identities, so a mis-sliced overlap fails immediately: each backend must
reproduce its input row-for-row whichever leg handles the tail. Covers the MLP and GDN
paths (all four GDN projections), the ANE leg pinned to `full_blocks + 1` calls above the
threshold and `full_blocks` below, the inclusive threshold boundary, exact multiples, env
parsing (including a negative value that must not enable the overlap, and that the
threshold is resolved once rather than per call), `target_verify` never accelerated, and a
declining ANE leg falling back to the GPU tail without losing rows.

With the existing ANE suites: 124 passed.

## Benchmark

`Qwen3.8-27B-oQ4e-mtp`, batched path (native MTP + ANE), shape 1024, ANE 0.60 MLP / 0.60
GDN, `dual_ane` off, Apple M4 Pro / 64 GB, oMLX 0.6.3rc2 with rc2-built native kernels.
Cold prompts with a unique preamble so the prefix cache cannot serve them; median of 3.

| prompt | tail rows | off | threshold 720 |
|---|---|---|---|
| ~1192 | 167 | 110.0 (109.6–110.1) | 109.6 (109.3–110.0) — parity |
| ~4089 | 1018 | 108.6 | **123.2** (+13.4%) |
| ~8076 | 911 | 113.2 | **119.3** (+5.4%) |

Gains concentrate where the unaligned final chunk is a large share of the prompt. Below
the threshold the overlap declines the trade and throughput is unchanged — which is the
point of shipping a threshold rather than always-on: forcing it with
`OMLX_QWEN35_ANE_TAIL_OVERLAP_MIN=1` measured **80.9** tok/s at a 167-row tail against
**111.4** with it off, a 27% regression (same base, measured before the caching fix).

**Peak memory**, same base, 8K prefill, server RSS sampled every 250 ms:

| arm | post-load | peak |
|---|---|---|
| off | 17638 MB | 18025 MB |
| on | 17656 MB | 18071 MB |

About **46 MB (+0.25%)** of transient headroom — one extra full-width result per
accelerated layer, released as each layer concatenates — and nothing persistent. Not added
to `estimate_prefill_peak_bytes()`, which accounts for KV, SDPA and fixed ANE surfaces
rather than arbitrary activation intermediates; a fixed reserve would mislead across models
and shapes. If this becomes default-on, a shape-derived transient estimate should replace
this note.

### Compatibility with rc2 CPU Sharing (#2892)

On an FP16 clone prepared with `tools/clone_mlx_model_fp16.py`, engagement confirmed in the
log ("CPU sharing using performance-aware scheduling with 8 workers", +4.22 GB projected):

| config | ~4.1K | ~8.1K |
|---|---|---|
| cpu off, overlap off | 113.9 | 118.3 |
| cpu off, overlap on | **126.3** | **122.3** |
| cpu on, overlap off | 109.9 | 113.8 |
| cpu on, overlap on | **122.8** (+11.7%) | **119.1** (+4.7%) |

The overlap gain survives CPU sharing, so the two compose. Incidentally CPU sharing at the
default `cpu_fraction` of 0.135 measured 3.5–3.8% *slower* than ANE/GPU alone on this M4
Pro, against the +4–6% reported for an M3 Ultra in the rc2 notes — a tuning question for
the split tuner, not something this change affects.

Absolute rates drift ~1.5% across runs on a machine under sustained load, so only
within-run off-versus-on comparisons are meaningful above.

Tail rows move from the GPU quantized-linear route to the ANE INT8 route, so greedy output
can shift — the same approximation class as enabling ANE prefill at all. Task-level quality
was not evaluated.

### DIFF
--- a/omlx/patches/qwen35_ane_prefill.py
+++ b/omlx/patches/qwen35_ane_prefill.py
@@ -201,6 +201,62 @@ def _tiled_input_plan(
     return full_blocks, tail_rows
 
 
+_TAIL_OVERLAP_MIN_CACHE: int | None = None
+
+
+def _tail_overlap_min() -> int:
+    """Minimum tail rows before an overlapping ANE tile beats the GPU tail.
+
+    Resolved once and cached: this runs on every accelerated MLP and GDN call of
+    every layer, and an ``os.environ`` lookup per call measured as a ~1.6%
+    prefill regression on short prompts, where there is little work to amortize
+    it against.
+
+    A wide prefill splits into ``full_blocks`` fixed-shape ANE tiles plus a
+    residual tail that runs on the ordinary quantized linears. The tail is pure
+    GPU work, so a long tail caps the achievable prefill rate: measured on an
+    M4 Pro, ANE-covered rows run at about 118 tok/s against about 83 tok/s for
+    GPU rows, so an 8K prompt at shape 1024 (7 tiles, 911-row tail, ~89%
+    coverage) lands at ~112 tok/s against ~120 for full coverage.
+
+    The tail can instead be covered by re-running the *last* ``sequence_length``
+    rows as a full ANE tile and keeping only its last ``tail_rows`` outputs.
+    Both accelerated ops -- the MLP gate/up/down chain and the GDN input
+    projections -- are position-wise, so recomputing rows that an earlier tile
+    already produced is redundant work rather than incorrect work. This is not
+    the padding approach the module rejects elsewhere: no synthetic tokens enter
+    the prompt, no KV or recurrent state advances; only existing rows are
+    re-projected.
+
+    The trade is one full tile (``sequence_length / ane_rate``) against the tail
+    on GPU (``tail_rows / gpu_rate``), so it pays when
+    ``tail_rows > sequence_length * gpu_rate / ane_rate`` -- about 720 rows for a
+    1024 shape on this hardware. Set
+    ``OMLX_QWEN35_ANE_TAIL_OVERLAP_MIN=720`` to enable; 0 (default) keeps the
+    GPU tail.  720 is an M4 Pro heuristic: the break-even moves with the ANE and
+    GPU-suffix rates, so CPU sharing, dual-ANE mode, quantization and the
+    compiled shape all shift it, and it is a candidate for derivation from the
+    split tuner's measurements rather than a constant.
+    """
+    global _TAIL_OVERLAP_MIN_CACHE
+    if _TAIL_OVERLAP_MIN_CACHE is None:
+        try:
+            # negative values would otherwise enable the overlap for every tail,
+            # including the short ones the break-even explicitly rejects
+            _TAIL_OVERLAP_MIN_CACHE = max(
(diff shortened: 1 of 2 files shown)


onthehub97:
Great minds...

#2966

jundot:
I am closing this in favour of #2966. Both approaches cost exactly one extra fixed-shape tile, and you and @onthehub97 independently derived the same crossover, so there is no performance argument between them. #2966 wins on two things that are not about speed: it also covers chunks narrower than the compiled shape, which the overlap cannot reach structurally, and it derives the threshold from the tuner run rather than a per-machine constant, which is the gap you flagged yourself.

Before deciding I checked the one place where zero padding could genuinely be worse than your approach, a per-tile activation quantization inside the ANE program that would let the padded rows shift the scale for the real ones. There is none in the native sources, and #2966 measured 0.99999577 cosine against the unpadded path, so your synthetic-row argument is sound in principle but does not bite here.

The measurement work here is the best of the three tail PRs I have open. The negative control at 167 rows, the transient memory delta, and the CPU sharing interaction are all things I normally have to ask for. Your tests are the stronger set too, stubbing both legs with identities so a mis-slice fails on a value mismatch catches what a constant stub cannot.

Sorry about the duplicated effort, and please keep going at this area.

PR #2966, opened 2026-08-20
feat(Qwen35) Add tuner-calibrated intermediate tail padding for Qwen ANE prefill

onthehub97:
# Add tuner-calibrated intermediate tail padding for Qwen ANE prefill

## Summary

This PR allows profitable residual prompt tails to continue through the
fixed-shape Qwen ANE prefill path instead of always falling back to the GPU.

The implementation does **not** add tokens to the prompt. It zero-pads only
the intermediate activation passed to tokenwise MLP and GDN projections, runs
one fixed-shape ANE tile, and removes the synthetic rows before any later model
stage. Token positions, KV-cache entries, and GDN recurrent state therefore
remain unchanged.

The ANE tuner calculates and persists the minimum profitable tail length from
the GPU-only and winning hybrid throughput measured during the current tuning
run. The threshold is exposed as a per-model setting in both the WebUI and
SwiftUI application.

This is a stacked change based on `qwen35-ane-down`.

## Motivation

The ANE programs use a fixed token dimension, normally 2,048. Wider prompts
are split into full fixed-shape tiles plus a residual tail. Previously, every
tail shorter than 2,048 tokens ran on the ordinary GPU projection path.

This is particularly costly for common `2,048n - 1` prefills. For example, a
4,095-token prefill contains one complete 2,048-token ANE tile followed by a
2,047-token tail that previously returned to the GPU despite being almost a
complete tile. Running one zero-padded ANE tile can be substantially faster in
that case, even though one row of accelerator work is discarded.

Using Qwen's configured padding token is not a valid substitute. It is a
learned token rather than a state-neutral null token and would alter positions,
the KV cache, logits, and GDN state.

### Intermediate activation padding

- Adds `qwen35_ane_prefill_tail_padding_min_tokens` to the Qwen ANE model
  settings. A value of `0` disables tail padding.
- Supports profitable short inputs and residual tails from wider inputs.
- Zero-pads the MLP or GDN projection input to the configured fixed sequence
  length.
- Runs the existing exact-shape production backend, including the configured
  ANE/GPU/CPU split.
- Slices the result back to the real row count.
- Removes padded GDN rows before recurrence, so they cannot modify recurrent
  state.
- Retains the existing GPU fallback for tails below the threshold, unsupported
  layers, decode, and target verification.
- Leaves exact-shape tiles and the full-block portion of wider prefills
  unchanged.

### Tuner-derived crossover

After the tuner selects its winning full-model candidate, it computes the
first integer tail for which estimated GPU execution is slower than one padded
hybrid tile.

Given:

- `S`: fixed ANE sequence length
- `G`: measured GPU-only prompt throughput
- `H`: measured throughput of the winning hybrid configuration

the threshold is:

```text
floor(S × G / H) + 1
```

This follows directly from comparing:

```text
GPU tail time       = tail_tokens / G
padded hybrid time  = S / H
```

For the reference tuning result of 349.4 GPU tok/s and 527.4 hybrid tok/s at a
2,048-token fixed shape, the first profitable integer tail is **1,357 tokens**.

The tuner deliberately forces padding off while searching. This prevents a
previously saved threshold from influencing the new GPU baseline or candidate
measurements. The threshold is calculated only after the winner is selected
and remains `0` when the hybrid result does not beat GPU-only.

### Configuration and UI

- Persists the threshold in global and per-model settings.
- Includes it in the engine reload signature so applying a changed threshold
  reloads the active model.
- Validates that the value is non-negative and smaller than the configured ANE
  sequence length.
- Resets an incompatible threshold when the sequence length changes.
- Adds an arbitrary numeric **Pad tails from** control to the WebUI model
  settings.
- Adds **Pad Intermediate Tails From** to the SwiftUI model settings.
- Includes the calculated threshold in tuner recommendations and applies it
  with the rest of the selected configuration.
- Displays the threshold in completed tuner results.
- Extends benchmark settings import/export support and ANE runtime status.

### Pad projections, not the token sequence

MLP and GDN input projections operate independently per token row. Appending
zero rows around these operations cannot affect the real rows, provided the
extra output rows are removed immediately. This makes intermediate padding a
local transformation with bounded semantics.

Padding the actual token sequence would not be local. It would advance cache
positions and GDN state and would require masking support through attention,
RoPE, cache management, and every recurrent update. This PR intentionally does
not attempt that.

### Reuse the exact-shape production backend

The padded path calls the same `_backend_exact` and `_gdn_backend_exact`
implementations used for ordinary fixed-shape tiles. It therefore preserves
the selected ANE instance assignment, GPU suffix, CPU sharing, fused MLP/down
paths, quantization handling, and output merge logic without introducing a
second execution implementation.

### Use a deterministic tuner result rather than a runtime profitability guard

The decision is reduced to one integer comparison on the hot path. There is no
runtime timing, device-utilisation heuristic, or scheduler prediction. The
tuner owns calibration, saves the resulting crossover with the model, and
leaves padding disabled when it cannot demonstrate a hybrid throughput gain.

### Keep small tails on the existing path

Padding always executes a complete fixed-shape tile, so it is predictably a
loss for small tails. The calibrated threshold preserves the existing GPU path
below the crossover and makes the additional work opt-in for old model
profiles until they are retuned or configured manually.

### Full production dispatcher

Tested on the real FP16-clone/4.85-bit Qwen3.8-27B model with:

- 2,048-token ANE shape
- 4,095 real prefill tokens
- all 64 eligible MLP and 48 eligible GDN layers compiled
- fused MLP split: 19% per ANE
- GDN ANE split: 45%
- CPU MLP share: 14%
- CPU GDN share: 13%
- 8 CPU workers
- tuner-derived padding threshold: 1,357 tokens

Both arms used the same hybrid configuration. The comparison changes only the
handling of the 2,047-token residual tail.

| Tail handling | Median time | Prompt throughput | Change |
|---|---:|---:|---:|
| Existing GPU residual tail | 9.8606 s | 415.3 tok/s | — |
| Padded fixed-shape hybrid tail | 7.6422 s | 535.8 tok/s | **+29.0%** |

Three interleaved samples per arm:

```text
GPU residual:    9.8348 s, 9.8608 s, 9.8606 s
Padded residual: 7.6422 s, 7.6462 s, 7.6399 s
```

### Numerical comparison

| Output | Cosine similarity |
|---|---:|
| Full hidden state | 0.99999577 |
| Last hidden state | 0.99998610 |
| Logits | 0.99999895 |

The padded and discarded rows do not enter GDN recurrence or subsequent model
layers. Remaining numerical differences are consistent with the existing
approximate INT8 ANE execution path.

### Tail-size probe

A focused projection probe confirmed the expected crossover behaviour. Small
tails lose because they pay for a complete ANE tile, while the gain rises
rapidly as the residual approaches 2,048 rows.

| Tail rows | MLP change | GDN change |
|---:|---:|---:|
| 1,024 | -5.9% | -5.7% |
| 1,152 | +5.4% | +6.4% |
| 1,280 | +16.7% | +26.1% |
| 1,536 | +39.3% | +40.5% |
| 1,792 | +61.6% | +66.3% |
| 2,047 | +84.6% | +84.0% |

The full-model formula produces a deliberately more conservative 1,357-token
threshold for the reference calibration.

## Validation

- 819 focused Python tests passed across ANE dispatch, tuning, model settings,
  engine loading, scheduler/benchmark paths, and WebUI routes.
- Added coverage for:
  - padded short MLP inputs
  - padded wide MLP residuals
  - padded GDN residuals and pre-recurrence slicing
  - crossover calculation and neutral/regressive tuner results
  - isolation from previously saved thresholds
  - settings validation, persistence, and tuner recommendation application
- WebUI JavaScript syntax validation passed.
- All locale JSON files passed `jq` validation.
- `git diff --check` passed.
- A full release build with custom kernels and the bundled CPython 3.11 runtime
  completed successfully.
- The staged application loaded the packaged CPython 3.11 native extension and
  reported the ANE backend available.
- The Swift test target compiled successfully. The command-line test host did
  not terminate after launching the full application, so no Swift runtime test
  result is claimed here.

### DIFF
--- a/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
+++ b/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
@@ -1269,6 +1269,17 @@ private struct ExperimentalSection: View {
                                   isNumeric: true, range: 1024...262_144,
                                   step: 64, width: 190)
                     }
+                    Row(label: String(localized: "settings.experimental.qwen_ane.tail_padding.label",
+                                      defaultValue: "Pad Intermediate Tails From",
+                                      comment: "Row label for the Qwen ANE intermediate tail threshold"),
+                        sublabel: String(localized: "settings.experimental.qwen_ane.tail_padding.sub",
+                                         defaultValue: "Residual projection blocks at least this large are zero-padded to the ANE shape. Zero disables padding; Tune ANE Split calculates the crossover.",
+                                         comment: "Sublabel explaining Qwen ANE intermediate tail padding")) {
+                        TextInput(text: vm.bindProfile($vm.qwen35AnePrefillTailPaddingMinTokens),
+                                  placeholder: "0", mono: true,
+                                  isNumeric: true, range: 0...262_143,
+                                  step: 1, width: 190)
+                    }
                     Row(label: String(localized: "settings.experimental.qwen_ane.mlp_fraction.label",
                                       defaultValue: "MLP on ANE",
                                       comment: "Row label for the Qwen MLP ANE workload fraction"),
@@ -1774,6 +1785,9 @@ private struct ExperimentalSection: View {
             let down = Int(((recommendation.cpuDownFraction ?? 0) * 100).rounded())
             let gdn = Int(((recommendation.cpuGdnFraction ?? 0) * 100).rounded())
             parts.append("CPU \(gate)%/\(down)%/\(gdn)%")
+        }
+        if let threshold = recommendation.tailPaddingMinTokens, threshold > 0 {
+            parts.append("Pad tails ≥\(threshold)")
         }
         return String(
             format: "%@ · %.1f tok/s (%+.1f%%)",
--- a/apps/omlx-mac/Sources/AppView/Utils/ProfileWorkingState.swift
+++ b/apps/omlx-mac/Sources/AppView/Utils/ProfileWorkingState.swift
@@ -71,6 +71,7 @@ struct ModelSettingsSnapshot: Equatable {
     var turboquantKvBits: String
     var qwen35AnePrefillEnabled: Bool
     var qwen35AnePrefillSequenceLength: String
+    var qwen35AnePrefillTailPaddingMinTokens: String
     var qwen35AnePrefillFraction: String
     var qwen35AnePrefillMaxLayers: String
     var qwen35AnePrefillDualAne: Bool
@@ -143,6 +144,7 @@ enum ProfileSettingsKey {
     static let turboquantKvBits = "turboquant_kv_bits"
     static let qwen35AnePrefillEnabled = "qwen35_ane_prefill_enabled"
     static let qwen35AnePrefillSequenceLength = "qwen35_ane_prefill_sequence_length"
+    static let qwen35AnePrefillTailPaddingMinTokens = "qwen35_ane_prefill_tail_padding_min_tokens"
     static let qwen35AnePrefillFraction = "qwen35_ane_prefill_fraction"
     static let qwen35AnePrefillFusedDown = "qwen35_ane_prefill_fused_down"
     static let qwen35AnePrefillMaxLayers = "qwen35_ane_prefill_max_layers"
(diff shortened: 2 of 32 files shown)


jundot:
Merged. Padding only the projection input and slicing before anything downstream sees it keeps the semantics local, and I verified the pre-recurrence slicing holds in the code. Putting the crossover in the tuner instead of a runtime heuristic is the right shape for this codebase.

One note for anyone reading the numbers: with the default q4 wrapper min-tokens gate, a short standalone prompt only pads the MLP side, so GDN padding effectively applies to wide prefill tails.

PR #2970, opened 2026-08-20
feat(cluster): support native MTP on pure tensor-parallel deployments

alytaphoenix:
**Stacked on #2845** (`feat(cluster): serve VLM checkpoints text-only under distributed inference`). Both Qwen models validated below are VLM-shaped checkpoints deployed via #2845's `text_only: true` path — this PR's `ClusterDeployment.mtp_enabled` field is added directly after #2845's `text_only` field, so it cannot be reviewed or merged independently of it. This PR's diff is scoped to exactly the MTP-on-pure-TP commits (plus one stale-test fixup surfaced by rebasing onto current `main` instead of this session's long-diverged local branch — see "Tests" below); it will need a rebase once #2845 lands.

## Summary

Distributed inference previously rejected `mtp_enabled` unconditionally: each rank in a cluster deployment runs a private stock `mlx_lm.server` subprocess that never loads the `omlx.patches` MTP stack. This PR carves out an exception for **pure tensor-parallel** deployments (not pipeline-parallel), where every rank already computes identical logits via mlx-lm's stock synchronized sampler:

- Each worker now applies the MTP patch stack before load.
- The adaptive depth controller runs only on the coordinator rank, broadcasting its depth/park decision each cycle via a single fused `mx.distributed.all_sum` collective that also carries a cycle checksum — any rank disagreement raises immediately instead of hanging in a later layer's all-reduce.
- Also fixes a latent bug where any MTP-capable checkpoint's head weights were misattributed or dropped in the distributed planner's per-rank memory accounting, regardless of whether MTP was enabled.

A second commit fixes a crash found while validating on real two-machine hardware: `_cross_thread_generation_stream` only ever registered a GPU thread-unsafe default stream on `mlx_lm.server`'s background generation thread. Plain single-node decode never touches the CPU device from that thread, so this was enough until native MTP's distributed depth/park coordination (`_sync_distributed_mtp_cycle`) and the tensor-parallel collectives inside a forward pass started requiring one too — surfacing as `RuntimeError: There is no Stream(cpu, 1) in current thread`. Fixed by registering a CPU thread-unsafe stream alongside the existing GPU one and rebinding both on every `_generate()` call, mirroring the existing GPU-stream mechanism.

The row-wise multi-sequence MTP batch path remains gated behind `OMLX_MTP_ROWWISE_BATCH` (unset by default), so concurrent-request batches on a cluster deployment continue to fall back to standard decode rather than exercising an unvalidated code path.

## Verified end-to-end on a 2-Mac Thunderbolt / JACCL cluster (M4 Pro coordinator + peer)

Deployed via the two-step `/admin/api/cluster/plan` → `/admin/api/cluster/deployments` flow (`text_only: true` for the VLM-shaped checkpoints, since the one-click autoconfigure endpoint has no text-only option), each reaching TP=2 pure tensor-parallel (`pipeline_stages: 1`, every rank holding the full layer range) and serving a real, correct `/v1/chat/completions` response:

| Model | Shape | Functional check | Decode tok/s, MTP on vs off (short ~500 tok / medium ~1500 tok prompt) |
|---|---|---|---|
| `NVIDIA-Nemotron-3.5-Lightning-30B-A3B-oQ4-mtp` | LLM (`batched`) | ✅ correct chat completion, ~30 tok/s | (functional only, benchmark not required per prior validation) |
| `Qwen3.8-27B-oQ4e-mtp` | VLM-shaped, text-only deploy | ✅ "The capital of France is Paris." | **37.9 / 30.7** vs 23.5 / 21.9 (≈1.6x / 1.4x) |
| `Qwen3.6-35B-A3B-oQ4e-mtp` (MoE) | VLM-shaped, text-only deploy | ✅ "The capital of Japan is Tokyo." | **76.8 / 70.3** vs 71.1 / 66.1 (≈1.08x / 1.06x) |

Benchmark methodology: prefill-isolated decode timing (`max_tokens=1` baseline subtracted from a `max_tokens=200` real call, `decode_tok_s = (completion_tokens - 1) / decode_time`), each prompt using a nonce **prefixed** to the content so the baseline and real calls (and separate runs) never share a cacheable prompt prefix — a suffix-only nonce still lets the real call ride the baseline's just-cached KV state, which produces the negative decode times seen with a naive nonce placement.

The smaller MTP win on the 35B MoE model is directionally expected: MoE forward passes are already compute-light per token (fewer active params through routing), so speculative verification overhead eats more of the relative gain than on the dense 27B — but MTP is never slower than the same hardware/model/prompt with it off, in either case.

Worker process RSS on both ranks was consistent with the planned shard size for both Qwen models; no macOS admin/authorization dialogs were triggered by any of the deployments in this session — the Thunderbolt/RDMA fabric link self-heals via `configure_link`'s tier-2 recorded-intent re-assertion on the next `link_setup`, requiring no new consent when the addressing intent is unchanged.

## Tests

- `tests/test_cluster_mtp.py` (new, 26 cases): distributed depth/park broadcast, checksum disagreement detection, patch-stack application gating on pure-TP vs pipeline-parallel.
- `tests/test_cluster_inference_worker.py`, `tests/test_cluster_planner.py`: updated for MTP-aware worker launch and per-rank head-weight accounting.
- One stale-test fixup commit: the original MTP commit was authored on a long-diverged local branch that predated `main`'s #2731 ("align thinking budget behavior across engines"), which had already removed `thinking_budget_enabled` from `DistributedBatchedEngine`'s incompatible-settings list. Rebasing onto current `main` (via #2845) surfaced the mismatch as a failing test, not a code bug — fixed by dropping that stale parametrize case; the other four settings (`dflash_enabled`, `specprefill_enabled`, `vlm_mtp_enabled`, `turboquant_kv_enabled`) are still gated and covered.
- 102/102 passing (`pytest tests/test_cluster_mtp.py tests/test_cluster_inference_worker.py tests/test_cluster_planner.py`); 1024/1024 passing across the full cluster/distributed suite (`pytest tests/ -k "distributed or cluster"`) in this PR's actual tree, stacked on #2845 and current `main`.

### DIFF
--- a/omlx/cluster/deployment.py
+++ b/omlx/cluster/deployment.py
@@ -248,6 +248,13 @@ class ClusterDeployment:
     performance_profiles: tuple[NodePerformanceProfile, ...] = ()
     tensor_parallel_size: int = 1
     target_context_tokens: int = 8192
+    # Explicit user opt-in to serve a VLM-shaped checkpoint as a text-only
+    # distributed model: the pinned mlx-lm rank loader drops the vision tower
+    # during sanitize, so only the language model runs across ranks. The flag
+    # is persisted so peers and later server restarts keep honoring the
+    # opt-in; an un-flagged VLM deployment stays refused (silent vision drop
+    # is treated as a bug, #1261/#1426).
+    text_only: bool = False
 
     def __post_init__(self) -> None:
         if _NODE_ID.fullmatch(self.deployment_id) is None:
@@ -373,6 +380,7 @@ class ClusterDeployment:
             ],
             "tensor_parallel_size": self.tensor_parallel_size,
             "target_context_tokens": self.target_context_tokens,
+            "text_only": self.text_only,
         }
 
     @classmethod
@@ -414,11 +422,16 @@ class ClusterDeployment:
             ),
             tensor_parallel_size=int(payload.get("tensor_parallel_size", 1)),
             target_context_tokens=int(payload.get("target_context_tokens", 8192)),
+            text_only=bool(payload.get("text_only", False)),
         )
 
     def encode_worker_plan(self) -> str:
         """Encode the small trusted plan as a bounded command-line argument."""
 
+        # ``text_only`` deliberately stays out of the worker contract: ranks
+        # load through the pinned mlx-lm, whose sanitize drops vision weights
+        # for these checkpoints natively, so the worker needs no flag and the
+        # contract decoded by ``decode_worker_contract`` stays unchanged.
         raw = json.dumps(
             {
                 "schema_version": 1,
--- a/omlx/cluster/deployment.py
+++ b/omlx/cluster/deployment.py
@@ -255,6 +255,19 @@ class ClusterDeployment:
     # opt-in; an un-flagged VLM deployment stays refused (silent vision drop
     # is treated as a bug, #1261/#1426).
     text_only: bool = False
+    # Native MTP (mlx_lm_mtp patch), pure-tensor-parallel only — see
+    # DistributedBatchedEngine._validate_model_settings. Threaded onto every
+    # rank's launch argv (build_mlx_launch_argv) so
+    # inference_worker.run_worker can pass a synthesized settings object into
+    # maybe_apply_pre_load_patches before provider.load_default(). Mirrors
+    # how trust_remote_code is injected late via dataclasses.replace() at
+    # engine_pool.py, since ClusterDeployment plans are built before the
+    # per-request model_settings is known.
+    mtp_enabled: bool = False
+    # None defers to maybe_apply_pre_load_patches' per-model-type default
+    # (omlx/utils/model_loading.py); set clamps to the same [1, 8] range as
+    # set_mtp_depth.
+    mtp_num_draft_tokens: int | None = None
 
     def __post_init__(self) -> None:
         if _NODE_ID.fullmatch(self.deployment_id) is None:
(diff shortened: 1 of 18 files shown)


alytaphoenix:
**Edit: this close was based on a mischaracterization — reopened, see below.**

Original comment (kept for history, corrected):

~~Closing — superseded by the newer cluster design work (see docs/cluster-turboquant-ane-mtp-support.md), which found MTP's clustering blocker is a rank-divergent control-plane issue (per-cycle wall-clock-driven depth/parking decisions) rather than the missing-plumbing problem this PR addressed. Revisiting with that understanding before reattempting.~~

That's not accurate: this PR already implements a rank-0-broadcast-with-checksum design for MTP's per-cycle draft-depth decisions, specifically to keep the control plane in lockstep across ranks under tensor parallelism, and it was hardware-validated. It wasn't addressing a "missing-plumbing" problem — it was addressing the same rank-divergence problem the newer design doc (re)identified, with a solution already in hand. Reopened; the design doc should reference this PR as prior art rather than treating the divergence issue as unaddressed.

alytaphoenix:
Reopening at the user's request — the earlier close was premature.

alytaphoenix:
Reopening — restoring the branch and undoing the earlier close.

alytaphoenix:
Rebased onto current main. This one needed real care: this branch depends on #2845's text-only VLM-serving work (not yet merged), and my first rebase attempt accidentally stacked onto #2845's own branch tip from Aug 19 instead of current main — 48 commits stale, which silently masked a real interaction bug (MTP head bytes should fold into `fixed_weight_bytes`, but with the stale base a duplicate/dead accounting path let a test pass for the wrong reason). Rebuilt cleanly on top of fresh main + #2845's 2 commits + this PR's 3 commits.

Two real fixes needed once properly based on current main:
- `test_cluster_text_only.py`'s two `fixed_weight_bytes` assertions predate this PR and expected MTP tensors excluded entirely; this PR correctly folds them in as replicated weight instead (matching `test_cluster_planner.py`'s own rationale) — updated the two stale assertions.
- The full-suite run caught a genuinely dead function (`_text_only_excluded_tensor`, unreachable per `test_no_unreachable_functions_in_the_cluster_package`) left over from an early merge decision — deleted it; its behavior is already covered by the two separate checks that replaced it.

Full suite: 10,706 passed, 0 failed, 150 skipped, 77 deselected.

Issue #2968, opened 2026-08-20
Loading Qwen3.8-27B-oQ4e-fp16-mtp with ANE stuff takes forever


taoeffect:
**Describe the bug**

Using `0.6.3rc2 build 2416` on M3 Max on macOS 15 Sequoia, attempting to load the `Qwen3.8-27B-oQ4e-fp16-mtp` model with ANE split settings (`Qwen ANE Prompt Processing`) enabled takes a really long time, almost 2 minutes (much longer than usual), and shows an exception thrown in the logs:

<details>
<summary>Exception:</summary>

```
2026-08-20 12:34:25,020 - omlx.engine_pool - INFO - [-] - Qwen CPU sharing adds 4.22GB to the projected memory for Qwen3.8-27B-oQ4e-fp16-mtp
2026-08-20 12:34:25,020 - omlx.engine_pool - INFO - [-] - Loading model: Qwen3.8-27B-oQ4e-fp16-mtp
2026-08-20 12:34:25,021 - omlx.engine_pool - INFO - [-] - Qwen CPU sharing adds 4.22GB to the projected memory for Qwen3.8-27B-oQ4e-fp16-mtp
2026-08-20 12:34:25,022 - omlx.patches.mlx_lm_mtp.nemotron_h_chain - INFO - [-] - nemotron_h MTP chain patch applied (depth-k drafting, sequential fused verify, replay-free rollback)
2026-08-20 12:34:25,022 - omlx.utils.model_loading - INFO - [-] - Speculative backend selected for /Applications/AI/models/scottlowry/Qwen3.8-27B-oQ4e-fp16-mtp: Lightning MTP (model_type=qwen3_5, active)
2026-08-20 12:34:25,024 - omlx.utils.model_loading - INFO - [-] - mlx-vlm MTP sanitize patch applied for /Applications/AI/models/scottlowry/Qwen3.8-27B-oQ4e-fp16-mtp
2026-08-20 12:34:25,024 - omlx.utils.model_loading - INFO - [-] - mlx-vlm runtime MTP patch applied for /Applications/AI/models/scottlowry/Qwen3.8-27B-oQ4e-fp16-mtp
2026-08-20 12:34:28,255 - omlx.engine.vlm - INFO - [-] - Vision feature cache enabled (SSD: /Users/user/.omlx/cache/vision_features)
[...]

PR #2975, opened 2026-08-21
perf(ane): reuse compiled Qwen programs across loads

joshuaswarren:
## Summary

Add an opt-in Qwen35 ANE compile cache that reuses Apple's existing AOT-compiled programs across fresh oMLX processes.

Private ANE programs currently pay the complete eager compile on every model load even when the runtime already has a compiled program for the same stable identifier. With `OMLX_QWEN35_ANE_COMPILE_CACHE=1`, oMLX now:

- gives each program a stable model URL keyed by OS build, ANE identifier, and instance
- checks `compiledModelExists` before compiling
- loads a hit without calling `compileWithQoS`
- serializes same-entry probes/writes across processes with `flock`
- invalidates a hit that fails to load, then recompiles exactly once
- fails open to the historical temporary compile path if the cache root or lock is unavailable

The cache covers individual linear programs, single fused SwiGLU/down programs, linear procedure banks, and fused SwiGLU/down procedure banks through one helper.

Apple owns the persistent compiled AOT artifacts. oMLX's MIL and weight staging files retain their historical delete-on-unload lifecycle, so the oMLX cache path retains only zero-byte lock files. The feature is off by default.

Related to #2968.

## Measured results

Fresh `Qwen3.8-27B` processes, same model/settings/prompt, ANE enabled, output compared exactly:

| hardware | cold miss | warm hit | reduction |
|---|---:|---:|---:|
| M1 Ultra, 128 GB | 65.17 s | 22.42 s | 65.6% |
| M2 Max, 96 GB | 55.74 s | 20.08 s | 64.0% |
| M1 Max, 64 GB | 54.37 s | 25.55 s | 53.0% |

Each cold process reported 112 misses. Two subsequent fresh processes reported 112/112 hits with byte-identical response text and no fallback. Persistent oMLX data footprint after unload: 0 bytes (lock files only).

A pressured Ultra run also showed Apple's AOT cache may partially evict programs (83 hits / 29 misses); the missing programs recompiled normally, and the next fresh shape converged to 112/112 hits twice.

## Failure and concurrency checks

- cache root made unwritable -> model served through temporary compile path
- two processes, same new identifier -> one miss/compile, one waits then hits; both succeed
- injected first-load failure on a real hit -> fallback log, one recompile, successful load (test-only instrumentation removed before final build)
- process output parity checked across cold, warm, and cache-disabled arms
- per-OS path component invalidates identifiers across OS compiler/runtime changes

## Verification

- 292 adjacent tests passed on M1 Ultra
- 108 focused tests passed on M2 Max
- 108 focused tests passed on M1 Max
- native custom kernels built on all three machines
- ruff clean on changed Python tests
- independent correctness, testing, maintainability, performance, reliability, and adversarial reviews; actionable findings applied and re-reviewed

## Configuration

```bash
OMLX_QWEN35_ANE_COMPILE_CACHE=1 omlx serve ...
```

Unset or any value other than exactly `1` keeps current behavior.

## Stack

This is the bottom PR (native implementation + tests). Documentation is
committed separately on the dependent `feat/ane-compile-cache-docs` branch and
will be submitted after this core branch lands; GitHub cannot create an upstream
PR whose base branch exists only in a fork.

## Rebase update (2026-08-21)

Rebased onto `0436bdf` after #2935 and its follow-up fixes landed. That upstream work added a fourth native compile path, fused SwiGLU/down procedure banks; it now routes through the same instance-aware cache helper. Conflict resolution also preserves #2935's consolidated `weight.bin` staging and execution options.

An independent re-review found unload cleanup could delete the shared staging path while another process compiled it. Both destructors now acquire the same per-entry lock before deleting cache-path staging; temporary/fail-open paths retain their historical direct cleanup, and a cleanup lock failure safely leaves staging for the next locked compile to replace.

Post-rebase verification on the final source:

- 116 focused tests passed on each of M1 Ultra, M2 Max, and M1 Max
- native custom kernels built on all three machines
- ruff clean on the changed Python tests
- the newly added fused-bank path produced a fresh-process miss then hit on all three Macs; after the cleanup-lock fix, M1 Ultra repeated miss -> hit at 0.154 s -> 0.008 s
- independent re-review reports no remaining concrete defects

The new GitHub Actions run is awaiting maintainer approval; the prior head passed Python 3.11/3.12/3.13.

### DIFF
--- a/omlx/custom_kernels/qwen35_prefill/csrc/qwen35_ane.mm
+++ b/omlx/custom_kernels/qwen35_prefill/csrc/qwen35_ane.mm
@@ -7,6 +7,7 @@
 #import <IOSurface/IOSurface.h>
 #import <Metal/Metal.h>
 #import <objc/message.h>
+#include <cerrno>
 #include <fcntl.h>
 #include <sys/file.h>
 #include <unistd.h>
@@ -154,9 +155,13 @@ NSString *ane_compile_cache_directory(NSString *identifier, int ane_instance) {
 
 std::string error_text(NSString *prefix, NSError *error);
 
-// Residual risk: acquisition has no timeout, so a suspended process holding
-// this lock blocks later loads of the same entry until the holder exits.
-// Accept that for now; revisit with a bounded timeout if it hurts.
+// A suspended process holding the entry lock must not hang later loads of the
+// same identifier forever. Acquisition is bounded: the lock is taken
+// non-blocking and retried until the deadline, then the constructor throws so
+// the caller fails open to the historical temp-directory path.
+constexpr std::chrono::milliseconds kAneCacheLockTimeout{30000};
+constexpr std::chrono::milliseconds kAneCacheLockRetry{50};
+
 class ScopedAneCacheLock {
 public:
   explicit ScopedAneCacheLock(NSString *entry_directory) {
@@ -175,11 +180,22 @@ public:
     if (fd_ < 0) {
       throw std::runtime_error("ANE compile cache lock open failed.");
     }
-    if (flock(fd_, LOCK_EX) != 0) {
-      close(fd_);
-      fd_ = -1;
-      throw std::runtime_error("ANE compile cache lock acquisition failed.");
+    const auto deadline = std::chrono::steady_clock::now() + kAneCacheLockTimeout;
+    for (;;) {
+      if (flock(fd_, LOCK_EX | LOCK_NB) == 0) {
+        return;
+      }
+      if (errno != EWOULDBLOCK && errno != EAGAIN) {
+        break;
+      }
+      if (std::chrono::steady_clock::now() >= deadline) {
+        break;
+      }
+      std::this_thread::sleep_for(kAneCacheLockRetry);
     }
+    close(fd_);
+    fd_ = -1;
+    throw std::runtime_error("ANE compile cache lock acquisition timed out.");
   }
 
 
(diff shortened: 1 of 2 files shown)


joshuaswarren:
@jundot @onthehub97 — flagging this once because it builds directly on the ANE work I tested in #2781/#2892/#2935 and addresses the repeat-load portion of #2968.

It does not remove the first cold compile; it reuses Apple’s existing AOT program on later fresh processes. Across M1 Ultra, M2 Max, and M1 Max, repeat load time dropped 53–66% with byte-identical output and zero persistent oMLX model-data footprint.

This is my first oMLX PR, so CI is also awaiting maintainer approval. No urgency — I just wanted to make sure it did not get lost in the current PR volume.

joshuaswarren:
Rebased onto `main` @ `2df39bfc` and fixed the one issue I would not have merged with.

**Change:** `flock(LOCK_EX)` blocked forever when a suspended (not crashed) process held the entry lock, hanging every later load of that identifier. Acquisition is now `LOCK_EX | LOCK_NB` with a 30s deadline and 50ms retry; on timeout the constructor throws and the caller fails open to the historical temp path. Also resolve symlinks before the cache-root prefix check so a symlink under the root cannot redirect the staging delete outside it.

**Tests:** added two source assertions (bounded lock, symlink-resolved delete). `tests/test_qwen35_ane_prefill.py` passes 116/116 on macOS 15 / Python 3.13. The added C++ compiles clean under `clang++ -std=c++17 -fobjc-arc` (the full `setup.py build_ext` is blocked by an unrelated `bonsai` kernel `find_package` in this environment, so I verified the changed code in isolation).

No other changes. Ready for review.

jundot:
Merging, thanks for the careful verification across the three machines. I merged main into the branch to clear a test-file conflict with the recent ANE work. I'll follow up on main with a dashboard toggle for this in advanced settings (it sets the env var before the first compile, so a restart applies it), a symlink check on the unload path, and lock file cleanup.

jundot:
Follow-ups are on main now: 678a468c adds the toggle, f83c9269 does the symlink check and the lock file cleanup.

I put the toggle in advanced settings rather than next to the other cache options because of what it leaves behind. The compiled programs sit in Apple's cache rather than ours, and since the identifier hashes the MIL, every sequence length, fraction and split layout gets its own entry, so a tuner sweep leaves several per model. My own machine is down to 40GB free with a few large models on it, which I don't think is unusual for this workload, and oMLX currently shows neither how much that cache is holding nor a way to clear it. That combination is what kept me on opt-in for now, and since the win is repeat load time rather than throughput, the people it helps most are the ones reloading models often, who will go looking for the setting anyway.

I want to revisit both the default and the placement once more people have run it. If it holds up across hardware and the cache stays reasonable in practice, moving it up next to the other cache settings and turning it on by default is the direction I'd like to go.

PR #2991, opened 2026-08-21
fix(sdpa256): q-split prefill fast path with memory-safety hardening

alytaphoenix:
## Summary

- Long-context prefill on head-dim-256 models (Qwen3.5/3.6/3.8 hybrid GDN+full-attention) was binary: fast `unfused` SDPA (O(L²) memory) or `tiled` online-softmax (O(L) memory, ~10x slower from per-KV-tile `mx.eval` sync). Once `unfused`'s transient stopped fitting available headroom, the router fell all the way to `tiled` and stayed there — an order-of-magnitude prefill cliff at long context (~19-23s/1k marginal vs ~2-4s/1k elsewhere).
- Adds a third route, `qsplit`: when the full call doesn't fit headroom but a narrower query slice would, split the query axis and run the same fast kernel per sub-tile (keys/values narrowed to each sub-tile's causal end), falling back to `tiled` only once even a minimum-size slice wouldn't fit.
- A live 258k-token verification run surfaced three real memory-safety gaps in the first version of this fix (each confirmed present, then confirmed fixed, via live re-runs):
  - The route gate credited the *entire* retained Metal buffer pool as reusable headroom. Since `kv_len` only grows within a request, a chunk's transient can never actually be reused by the next (larger) chunk — so the credit let the router keep choosing big-transient routes as the pool inflated, while the *uncredited* reactive hard-watermark guard saw the same pool as real pressure and tripped abruptly (abort + model eviction, instead of the old code's clean 400). Now credits pooled memory only for a same-`kv_len` repeat call within one chunk, and additionally charges 2x the candidate transient against the same watermark line the reactive guard actually kills at.
  - The q-split loop never evaluated between sub-tiles, so MLX's laziness let every sub-call's graph pile up simultaneously instead of bounding the live set to one sub-tile at a time, as `q_sub`'s sizing assumed.
  - No hysteresis: a request could flicker back to the full-unfused route right as pressure was building, moments before the guard tripped. Route decisions now hold at the most conservative route a request has needed.

## Test plan

- [x] `tests/test_sdpa256_attention.py` (42 tests): numerical equivalence of q-split vs reference/tiled, route-gate boundaries at all 3 routes, env overrides, headroom math (watermark targeting + 2x charge), pool-credit toggle (same-kv_len only), hysteresis (never climbs back once downgraded), eval-per-subtile ordering.
- [x] Full scheduler/memory-guard suite (`test_scheduler.py`, `test_process_memory_enforcer.py`, `test_dflash_prefill_memory_guard.py`, `test_engine_preflight.py`): 469 passed.
- [x] Full repo test suite on this branch: 10,231 passed, 0 failed, 102 skipped (pre-existing, unrelated).
- [x] Live end-to-end verification against `Qwen3.8-27B-oQ4e-mtp`: a 258k-token request that previously triggered a hard-watermark abort + model eviction now fails via the same clean, predictive 400 the pre-q-split code used, with the model staying loaded throughout — while still running ~1.5-2x faster per token below the ceiling than the tiled-only baseline.

### DIFF
--- a/omlx/patches/sdpa256_attention.py
+++ b/omlx/patches/sdpa256_attention.py
@@ -6,12 +6,22 @@ but deliberately keeps the faster unfused path as the default on pre-NAX GPUs.
 That default materializes the full ``[n_q, query_len, kv_len]`` score matrix and
 can still exceed oMLX's memory-guard ceiling.
 
-When the unfused transient fits, this patch preserves MLX's default routing. If
-it does not fit (or no guard ceiling is available), it calls MLX 0.32.2 with
-``force_fused=True``. This replaces oMLX's old pure-array tiled implementation:
-the bounded route is now an upstream native fused kernel instead of the slow
-sequential tile loop. On NAX, MLX's default already selects its fast split-D
-head-dim-256 kernel for causal prefills with at least 1024 queries.
+When the unfused transient fits, this patch preserves MLX's default routing.
+When it doesn't, the route narrows in two stages as live guard headroom shrinks
+(``_route_decision``):
+
+- **q-split** (``_unfused_qsplit_sdpa``): split the query axis into sub-tiles
+  (keys/values narrowed to each sub-tile's causal end) and run the SAME fast
+  stock kernel per sub-tile -- a smaller transient, still the fast kernel, no
+  accuracy cost. ``OMLX_SDPA256_QSPLIT=0`` disables this stage.
+- **bounded/tiled** (``_flash_sdpa256``): the last resort once even a
+  minimum-size q-split slice wouldn't fit (or no guard ceiling is available).
+  On Metal it calls MLX 0.32.2 with ``force_fused=True`` (the native fused
+  kernel); explicit array masks and sinks take the portable array-tiled
+  fallback, which the native fused path can silently unfuse. This replaces
+  oMLX's old pure-array-only bounded implementation. On NAX, MLX's default
+  already selects its fast split-D head-dim-256 kernel for causal prefills
+  with at least 1024 queries.
 
 ``OMLX_SDPA256_TILED=1/0`` remains accepted for compatibility and now forces or
 disables the bounded route. Metal uses the native fused kernel; CUDA retains
@@ -64,22 +74,131 @@ _HEADROOM_PROVIDER_LOCAL = threading.local()
 # Backward-compatible override: True = always force fused, False = never force,
 # None = memory-aware auto.
 _FORCE_TILED: bool | None = None
-# Bounded-route reasons already logged. The first engagement per reason logs at
-# INFO; repeats stay silent to keep the hot path quiet.
-_TILED_ROUTE_LOGGED: "set[str]" = set()
+# OMLX_SDPA256_QSPLIT override, parsed at apply time: False disables the
+# q-split route (falls straight to tiled once unfused doesn't fit, restoring
+# pre-q-split behavior for rollback); True/None (default) leaves it enabled.
+_QSPLIT_ENABLED: bool = True
+# Minimum query rows per q-split sub-call. Below this the per-call overhead
+# stops paying for itself and genuinely tight headroom is better served by the
+# tiled path's true O(L) floor.
+_QSPLIT_MIN_Q = 128
+# Route decisions round-trip through here so callers branch on one value.
+_ROUTE_UNFUSED = "unfused"
+_ROUTE_QSPLIT = "qsplit"
+_ROUTE_TILED = "tiled"
+# Last route decision, so we log every *transition* (not just the first-ever
+# engagement) at INFO -- cheap, and lets a live server log show a request
+# flapping between routes as guard headroom rises and falls mid-prefill. The
+# tiled/qsplit pass trades prefill throughput for safer memory, and nothing
+# surfaced the route decision before (issue #2283 took an A/B repro to
+# diagnose), so this stays at INFO rather than DEBUG.
+_LAST_ROUTE_DECISION: "str | None" = None
 
 
-def _note_tiled_route(reason: str, detail: str) -> None:
-    if reason in _TILED_ROUTE_LOGGED:
+def _note_route(decision: str, detail) -> None:
+    """``detail`` may be a plain string or a zero-arg callable. This runs on
+    every head-dim-256 prefill SDPA call (thousands per long-context request),
+    and the vast majority hit the decision == last-decision no-op below --
+    callers with a formatted (f-string) detail should pass a lambda so the
+    string is only built on an actual transition."""
+    global _LAST_ROUTE_DECISION
+    if decision == _LAST_ROUTE_DECISION:
         return
-    _TILED_ROUTE_LOGGED.add(reason)
+    _LAST_ROUTE_DECISION = decision
+    if callable(detail):
+        detail = detail()
     logger.info(
-        "sdpa256: head-dim-256 prefill forcing the memory-bounded path: %s. "
-        "The default fast path resumes when guard "
-        "headroom allows; "
-        "OMLX_SDPA256_TILED=1/0 forces the route.",
+        "sdpa256: route -> %s (%s). OMLX_SDPA256_TILED=1/0 forces "
+        "unfused/tiled; OMLX_SDPA256_QSPLIT=0 disables the q-split route.",
+        decision,
         detail,
     )
+
+
+def _total_qsplit_bytes(
+    n_q_heads: int,
+    q_len: int,
+    kv_len: int,
+    causal: bool,
+    q_sub: int,
+    head_dim: int,
+    score_dtype_size: float,
+) -> int:
+    """Sum of every q-split sub-tile's transient, not just one sub-tile's.
+
+    Causal: each sub-tile's keys/values are narrowed to that sub-tile's own
+    causal end (``_unfused_qsplit_sdpa``), so kv width grows sub-tile to
+    sub-tile -- every sub-tile is a *different* size, and Metal's buffer pool
+    cannot reuse a differently-sized buffer for the next one. They accumulate
+    as retained (not just active) memory across the whole loop: ``mx.eval``
+    between sub-tiles bounds what is *live* to one sub-tile, but says nothing
+    about what the pool has *retained* from the ones before it.
+
+    Non-causal: every sub-tile is ``q_sub`` rows against the full, constant
+    ``kv_len`` (no narrowing) -- same size as every other sub-tile, so the pool
+    genuinely reuses one buffer across the loop and only ONE sub-tile's
+    transient (not the whole q_len's) is ever retained at a time. Summing here
+    would overcharge a case that was never broken.
+    """
+
+    if not causal:
+        return estimate_unfused_sdpa_call_bytes(
+            n_q_heads, min(q_sub, q_len), kv_len, head_dim,
+            score_dtype_size=score_dtype_size,
+        )
+    kv_off = kv_len - q_len
+    total = 0
+    for qi0 in range(0, q_len, q_sub):
+        qi1 = min(qi0 + q_sub, q_len)
+        total += estimate_unfused_sdpa_call_bytes(
+            n_q_heads,
+            qi1 - qi0,
+            kv_off + qi1,
+            head_dim,
+            score_dtype_size=score_dtype_size,
+        )
+    return total
+
+
+def _max_q_sub_for_headroom(
+    n_q_heads: int,
+    q_len: int,
+    kv_len: int,
+    causal: bool,
+    head_dim: int,
+    score_dtype_size: float,
+    headroom: int,
+) -> int:
+    """Largest (128-aligned) query-row sub-tile count whose q-split transient
+    fits ``headroom`` -- the *total* retained across every sub-tile in the
+    split for a causal call (see ``_total_qsplit_bytes``), not the naive "one
+    sub-tile at a time" a per-call eval only bounds the *active* set to.
+    Non-causal degrades to the original single-transient inversion, since every
+    sub-tile there is the same size and genuinely reusable.
+
+    A closed-form inversion of the causal sum is a quadratic in ``q_sub``; a
+    bounded linear search from a safe starting point (this call's
+    single-sub-tile estimate, an upper bound since it ignores accumulation and
+    can therefore only be too large) is simpler to keep correct than an
+    inverted formula that has to be re-derived by hand every time this
+    function's cost model changes.
+    """
+
+    if headroom <= 0:
+        return 0
+    per_row = n_q_heads * (kv_len * score_dtype_size + head_dim * 4)
+    if per_row <= 0:
+        return 0
+    q_sub = min(q_len, int(headroom // per_row))
+    q_sub = (q_sub // 128) * 128
+    while q_sub >= 128:
+        total = _total_qsplit_bytes(
+            n_q_heads, q_len, kv_len, causal, q_sub, head_dim, score_dtype_size
+        )
+        if total <= headroom:
+            return q_sub
+        q_sub -= 128
+    return 0
 
 
 def set_unfused_headroom_provider(method) -> None:
@@ -109,6 +228,10 @@ def _parse_force_tiled_env() -> bool | None:
     return None
 
 
+def _parse_qsplit_env() -> bool:
+    return os.environ.get("OMLX_SDPA256_QSPLIT", "").strip() != "0"
+
+
 def _notify_bounded_route(provider, active: bool) -> None:
     """Let the scheduler retire measurements from the previous route."""
     try:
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for the work on this. I found a memory-safety regression in the q-split path.

Evaluating each sub-tile bounds active memory, but the causal sub-tiles use different KV lengths. Their differently sized score buffers remain in the Metal pool, so the retained physical footprint grows across the loop even though only one sub-tile is active at a time.

With Qwen3.8-27B-oQ4e-mtp, a 24 GB guard, and a 12,019-token prompt, this branch selected `qsplit(q_sub=512)`. The Metal pool grew from 6.6 GB to 12.7 GB, process usage reached 29.1 GB against a 22.8 GB hard watermark, and the request was aborted with the model evicted. Current main completed the same request in 38.39 seconds and kept the model loaded.

Could you rework the route and sizing to account for retained growth across all sub-tiles? The same-shape pool credit should also use the full call shape rather than `kv_len` alone, since interleaved chunks can have the same KV length but different query lengths. I do not think this is safe to merge yet.

alytaphoenix:
Rebased onto current main (zero conflicts) and pushed a fix for the retained-pool-growth issue flagged in review.

**Root cause confirmed**: `_max_q_sub_for_headroom` priced a causal q-split call as if only one sub-tile's transient was ever retained. But `_unfused_qsplit_sdpa` narrows keys/values to each sub-tile's own causal end, so sub-tiles grow in size across the loop -- differently-sized buffers that Metal's pool can't reuse for each other. `mx.eval` per sub-tile correctly bounds what's *active* to one sub-tile, but says nothing about what's *retained* from the ones before it. That's exactly the live-repro mechanism you described (qsplit(q_sub=512) selected, retained pool 6.6GB→12.7GB, abort against the 22.8GB watermark).

**Fix**:
- New `_total_qsplit_bytes` sums every sub-tile's transient for the causal case instead of pricing one. Non-causal degrades to the original single-sub-tile pricing -- those sub-tiles are all `kv_len`-wide (no narrowing), so the pool genuinely reuses one buffer there and the original formula was already correct for that path.
- `_max_q_sub_for_headroom` now searches downward in 128-row steps from a safe upper-bound starting point (the causal sum has no simple closed-form inversion).
- `Scheduler._sdpa256_unfused_headroom`'s pool-reuse credit now keys on `(kv_len, q_len)` instead of `kv_len` alone, per your note that two different chunks/requests could share `kv_len` but differ in `q_len` and still get wrongly credited.

Full 10,472-test suite passes (`pytest tests/`), including two new tests exercising the summed-cost sizing directly (`test_max_q_sub_for_headroom_causal_accounts_for_retained_pool_growth`) and the `(kv_len, q_len)` credit keying (`test_sdpa256_headroom_credits_pool_only_on_repeated_shape`, extended with a same-kv_len-different-q_len case). Confirmed the new tests fail against the pre-fix code (14 failures via stash-revert) to prove they actually discriminate.

jackkbarnes:
I tested the current PR head (`b81e97a8`) on an M5 Max 128 GB with Qwen3.8-27B-oQ8-mtp and found a reproducible long-context prefill regression in the automatic routing path.

### Configuration

- Apple M5 Max, 128 GB
- Qwen3.8-27B-oQ8-mtp
- TurboQuant KV enabled, 8-bit
- 24 GB hot cache
- context limit 262144
- cold/cache-busted ~132.1k-token prompts
- 512 generated tokens
- oMLX v0.6.3 baseline

### Results

**Stock v0.6.3**
- prompt: 132,098 tokens
- TTFT: 327.711s
- total wall: 360.949s
- decode: 15.37 tok/s

**#2991 automatic routing**
- prompt: 132,098 tokens
- TTFT: 404.263s
- total wall: 435.506s
- decode: 16.36 tok/s

**#2991 with `OMLX_SDPA256_TILED=0`**
- prompt: 132,100 tokens
- TTFT: 328.384s
- total wall: 360.699s
- decode: 15.81 tok/s

The automatic #2991 run therefore regressed TTFT by about **23.4% vs stock** at this context length.

The interesting isolation result is the third run. With the same #2991 code but `OMLX_SDPA256_TILED=0`, the server explicitly reported:

`sdpa256: route -> unfused (forced by OMLX_SDPA256_TILED=0)`

and TTFT returned almost exactly to stock: **328.384s vs 327.711s**.

During the automatic #2991 run, the logged route remained `unfused`; there was no logged transition to qsplit or tiled. So this does not appear to be qsplit/tiled kernel cost. The A/B suggests the regression is somewhere in the automatic live-headroom/routing path while still selecting unfused.

The automatic run initially logged:

`sdpa256: route -> unfused (full call ~816MiB fits ~33827MiB headroom at kv_len=8192)`

and later completed the 132,098-token prefill normally.

I have not yet profiled the individual calls inside `_sdpa256_unfused_headroom`, so I don't want to over-attribute the cause. But the forced-unfused A/B seems fairly strong evidence that the automatic headroom/routing machinery is introducing substantial overhead on this hardware/workload.

Happy to run an instrumented build if timing `_sdpa256_unfused_headroom`, `_current_usage_bytes`, `mx.get_active_memory()`, `mx.get_cache_memory()`, etc. would be useful.

alytaphoenix:
@jackkbarnes thank you — this is a genuinely excellent report. The same-build `OMLX_SDPA256_TILED=0` isolation, the captured route-log lines, and the exact config are exactly what makes it actionable. I dug into the one code path your A/B points at and want to share what I found, because the evidence so far does **not** reproduce a mechanism that can account for ~76s, and I'd like your help pinning down what's actually happening on the M5 Max.

## What the automatic path actually adds over stock

Your route log confirms the important thing: the head-dim-256 route engaged and stayed `unfused` the whole prefill (fresh TurboQuant requests run **fp16** during the cold prefill loop and are quantized to TurboQuant only at the end, so the `hasattr(cache, "bits")` early-exit doesn't fire — the route decision runs on every full-attention SDPA call). So we're looking strictly at the steady unfused decision path, called ~`layers × chunks` times.

For your config that's **64 full-attention layers × the chunk count** (≈32–64 chunks depending on prefill step size) ≈ **a few thousand calls** over the prefill.

I diffed the merge-base (`7e6f4224`) against the PR head (`b81e97a8`) for exactly which *native* calls that steady path gains. The answer is **one**: `mx.get_cache_memory()` (via `_current_usage_bytes(credit_cache_memory=…)`, credited only on same-shape repeat calls — layers 2..N of each chunk). `mx.get_active_memory()` and `get_phys_footprint()` were **already** called on this path before the PR (the old `_tiled_route_required` → `_sdpa256_unfused_headroom` → `_current_usage_bytes` chain), so they can't be the PR's delta. The cap-function swap (`_prefill_abort_cap` → `_admission_limit_bytes`) and the added Python bookkeeping (shape-tuple compare, `//2`, `getattr` hysteresis) are all pure-Python and sub-microsecond.

## Is `get_cache_memory()` expensive? MLX v0.32.0 source says no

The hypothesis I started with was that `get_cache_memory()` might take an allocator lock that `get_active_memory()` avoids, and stall under Metal completion-handler churn in the synchronous per-chunk eval regime. The MLX v0.32.0 source refutes that — both are lock-free trivial reads:

```cpp
// mlx/backend/metal/allocator.h  (v0.32.0)
size_t get_active_memory() { return active_memory_; };           // plain size_t read
size_t get_cache_memory()  { return buffer_cache_.cache_size(); };

// mlx/backend/common/buffer_cache.h
size_t cache_size() const { return pool_size_; }                 // plain size_t read
```

Neither touches the allocator `mutex_`. `get_cache_memory()` is not more expensive than `get_active_memory()` — structurally they're the same.

## Microbenchmark (M4 Pro, MLX 0.32.0)

Timed both, idle and under two background threads churning differently-sized buffer allocs/frees to force pool churn:

```
                          n=10000
get_active_memory idle    mean=0.044us  p50=0.042us  p99=0.084us  max=1.17us
get_cache_memory  idle    mean=0.050us  p50=0.042us  p99=0.084us  max=0.25us
get_active_memory churn   mean=0.037us  p50=0.042us  p99=0.084us  max=1.04us
get_cache_memory  churn   mean=0.037us  p50=0.042us  p99=0.084us  max=0.17us
```

~40ns per call, no millisecond tails, no difference between the two, no degradation under churn. At **4,000 calls × ~40ns ≈ 0.16ms** total added by the PR's one new native call — six orders of magnitude short of 76s.

A standalone version you can run on the M5 Max in ~10s (no omlx edits) is at the bottom; would be great to confirm the numbers hold on your hardware.

## Why the numbers still don't add up to a code cause

The subtle part: **stock v0.6.3 already runs this same headroom-provider path on the fp16 prefill** (the unfused/tiled route pre-dates this PR). So stock-auto and PR-auto both call the provider; the *only* difference between them is the one 40ns `get_cache_memory()`. Yet your stock-auto (327.7s) ≈ your PR-`TILED=0` (328.4s, which skips the provider entirely). If the provider path itself were the 76s, stock-auto — which also calls it — would be slow too. It isn't. That internal inconsistency, plus decode moving the "wrong" way between runs (16.36 vs 15.37 tok/s — the PR provably doesn't touch decode; `q_len<16` exits before any new code), reads to me like run-to-run environment variance sitting on top of a genuinely tiny code delta, rather than a 76s regression in the routing math.

I could be wrong, and I don't want to hand-wave away a clean-looking A/B — so a few concrete asks:

1. **Build provenance.** Your "Stock v0.6.3" is presumably the packaged build; were the two #2991 runs a *source checkout* of the branch? The packaged DMG bundles custom Metal kernels (e.g. the fa256 steel prefill path) a source build lacks, so "DMG-stock vs source-PR" isn't apples-to-apples. Could you run the **merge-base `7e6f4224`** (stock code for every file the PR touches) from the **same build/environment** as the two PR runs? That cleanly separates "PR diff cost" from "build/env difference."
2. **Interleaved repeats.** Given decode variance between runs, 2–3 interleaved A/B/A repeats (stock-auto / PR-auto / PR-`TILED=0`) would tell us how much of the 76s is stable vs noise.
3. **Take you up on the instrumented build.** Since you offered — the decisive measurement is wall-clock *inside* the path on your hardware. Accumulate time around each of the four pieces (`_sdpa256_unfused_headroom` total, `mx.get_active_memory()`, `get_phys_footprint()`, `mx.get_cache_memory()`) into module globals and log cumulative totals every ~500 calls. My prediction is all four sum to well under 1s across the whole prefill; if any counter shows seconds, that's the smoking gun and I'll chase it immediately. A `py-spy record` on one slow prefill would localize it even faster if that's easier.

If the same-build merge-base run reproduces the gap, I'll reopen this with a real mechanism to hunt. If it closes it, we'll know it was build/env. Either way this report has already made the path better-understood — thank you again. cc @jundot for visibility.

---

<details>
<summary>Standalone microbench (run on the M5 Max, no omlx changes)</summary>

```python
import time, threading
import mlx.core as mx

def bench(fn, iters=10000):
    xs = []
    for _ in range(iters):
        t0 = time.perf_counter_ns(); fn(); xs.append(time.perf_counter_ns() - t0)
    xs.sort(); n = len(xs)
    return (sum(xs)/n/1e3, xs[n//2]/1e3, xs[int(n*0.99)]/1e3, xs[-1]/1e3)

mx.eval(mx.sum(mx.ones((256,256)))); mx.get_active_memory(); mx.get_cache_memory()
for name, fn in [("active", mx.get_active_memory), ("cache", mx.get_cache_memory)]:
    m,p50,p99,mx_ = bench(fn); print(f"idle  {name:6s} mean={m:.3f}us p50={p50:.3f} p99={p99:.3f} max={mx_:.3f}")

stop = threading.Event()
def churn():
    i = 0
    while not stop.is_set():
        i += 1; n = 512 + (i % 64) * 64
        mx.eval(mx.random.normal((n, n)) @ mx.random.normal((n, n)))
ts = [threading.Thread(target=churn, daemon=True) for _ in range(2)]
[t.start() for t in ts]; time.sleep(0.5)
for name, fn in [("active", mx.get_active_memory), ("cache", mx.get_cache_memory)]:
    m,p50,p99,mx_ = bench(fn); print(f"churn {name:6s} mean={m:.3f}us p50={p50:.3f} p99={p99:.3f} max={mx_:.3f}")
stop.set()
```

</details>

jackkbarnes:
Thank you very much, @alytaphoenix! Happy to help, and the updated test set changes the view entirely.

Follow-up from the original M5 Max 128 GB testing: **the previously reported ~23.4% PR-auto TTFT regression does not reproduce under controlled same-environment testing. Consider that concern withdrawn.**

I reran this as an interleaved A-B-C-A-B-C series using:

- **A:** merge-base `7e6f4224`, automatic routing
- **B:** PR head `b81e97a8`, automatic routing
- **C:** PR head `b81e97a8`, `OMLX_SDPA256_TILED=0`

All six runs used the same Python/MLX dependency stack (Python 3.13.15, MLX 0.32.2, mlx-lm 0.31.3), the same Qwen3.8-27B-oQ8-mtp model files and settings, fresh oMLX server processes, isolated empty prefix/SSD caches, and the same 132,098-token cold-prefill workload with 512 generated tokens. Each run reported `cached_tokens=0`.

### Same-environment results

| Run | Condition | TTFT (s) | Wall (s) | Decode (tok/s) |
| --- | --- | ---: | ---: | ---: |
| A1 | Merge-base auto | 306.696 | 344.224 | 13.62 |
| B1 | PR auto | 291.771 | 328.233 | 14.01 |
| C1 | PR forced-unfused | 307.184 | 344.540 | 13.68 |
| A2 | Merge-base auto | 284.581 | 321.195 | 13.96 |
| B2 | PR auto | 283.015 | 315.574 | 15.69 |
| C2 | PR forced-unfused | 298.984 | 336.116 | 13.76 |

### Mean TTFT

| Condition | Mean TTFT (s) | vs. merge-base |
| --- | ---: | ---: |
| Merge-base auto | 295.638 | — |
| PR auto | **287.393** | **−8.246 s (−2.79%)** |
| PR forced-unfused | 303.084 | +7.446 s (+2.52%) |

PR-auto was faster than merge-base in both paired observations:

- B1 vs. A1: **−14.925 s**
- B2 vs. A2: **−1.566 s**

I would **not** interpret that as evidence that the PR improves performance. `n=2` is small, the A-B-C ordering was fixed, and merge-base itself varied by 22.115 s between repeats. The relevant result here is that the original ~76 s / 23.4% PR-auto penalty disappeared.

### Routing and memory behavior

Both PR-auto runs selected `unfused` once and never transitioned to qsplit/tiled.

- B1: full call ~816 MiB fit ~34,851 MiB headroom at `kv_len=8192`
- B2: full call ~816 MiB fit ~34,893 MiB headroom at `kv_len=8192`

Both forced runs logged:

`route -> unfused (forced by OMLX_SDPA256_TILED=0)`

Peak RSS across all six runs was tightly clustered at 28.404–28.445 GiB. There were no guard aborts, evictions, or route changes.

Every run used a fresh process and an isolated empty cache. All six reported exactly 132,098 prompt tokens and `cached_tokens=0`.

### Allocator-call microbench

I also reproduced the allocator-call microbench on this M5 Max with MLX 0.32.2:

```text
idle  active mean=0.029us p50=0.041us p99=0.042us max=1.125us
idle  cache  mean=0.028us p50=0.041us p99=0.042us max=0.250us
churn active mean=0.029us p50=0.041us p99=0.042us max=2.250us
churn cache  mean=0.028us p50=0.041us p99=0.042us max=0.208us

**Comment by alytaphoenix** [comment]
## Rebased onto current `main` (force-pushed `b81e97a8` → `8f616f0e`)

`main` has independently rewritten this same route since this branch was cut, so this is **not** a textual conflict resolution — I re-ported #2991's intent onto main's current file rather than replaying the branch diff. The old boolean `_should_route`/`_tiled_route_required` dispatch is gone; main now carries #3461 (fp32 pricing + array-mask bounded routing + `_notify_bounded_route`→`_sdpa256_bounded_route_changed` EWMA retirement), #3333 (engine-worker-scoped **thread-local** headroom provider), #3332 (MLX-0.32.2 **native-fused** tiled path replacing the branch's hand-rolled online-softmax `_flash_sdpa256`), and #3283 routing bounds. All of those are preserved. From this branch: the 3-way `_route_decision → (route, q_sub)`, `_total_qsplit_bytes`, the 128-row `_max_q_sub_for_headroom` search, the per-request ratcheting `_sdpa256_q_sub_ceiling`, per-sub-tile `mx.eval`, and the scheduler's `(kv_len,q_len)` pool-reuse credit keying + 2× watermark charge are all preserved.

Single reconciling commit (the file is restructured on both sides; intermediate replays would be untested states, and this repo squash-merges anyway).

### ⚠️ Three reconciliations that are NEW memory-safety math you have **not** reviewed — please delta-review these specifically

The prior review was of the pre-rebase code; it does not cover this reconciliation. Please look at these three points in particular:

1. **Pricing dtype (the dangerous one).** #3461 established that the unfused fallback materializes **fp32** scores and must be priced with the shared `SDPA256_UNFUSED_SCORE_DTYPE_SIZE`. This branch priced at `queries.dtype.size`. I threaded `SDPA256_UNFUSED_SCORE_DTYPE_SIZE` into **all three** qsplit sizing sites (`_route_decision` full-call transient, `_max_q_sub_for_headroom`, `_total_qsplit_bytes`). Taking the branch's pricing verbatim would, for bf16 queries, pick q_sub ~2× too large (halving the predicted transient vs the real fp32 footprint) — reintroducing the OOM class #3461 closed, now inside the q-split route. Added `test_qsplit_sizing_priced_at_fp32_not_query_dtype` to wire the two accounting paths together permanently.

2. **Route-flip notification semantics.** `_sdpa256_bounded_route_changed(active: bool)` is a boolean bounded/unbounded protocol; there are now three routes. I notify `active=True` for **both** qsplit and tiled (qsplit *is* memory-bounded, just via a different mechanism), and `active=False` only for whole-call unfused. A qsplit↔tiled flip therefore does not retire EWMA history — which is safe *because* the transient tracker maxes its observations, so carrying qsplit's larger transients into a tiled regime only over-predicts (conservative direction). Documented in a code comment; flagging because it's a judgment call.

3. **Array masks + sinks never take q-split.** q-split narrows KV to each sub-tile's *causal* end (valid only for causal/no-mask), and the array-tiled bounded kernel is the route that slices an array mask / applies sinks correctly. `_route_decision`'s `allow_qsplit` excludes both. Note I **dropped** the pre-rebase gate's outright rejection of sinks/array-masks into unbounded passthrough — keeping that would have regressed a tight-memory sinks call to the O(L²) unfused path. Array masks stay eligible for the bounded route per #3461 (Qwen4's profile trusts array-mask-capable bounded routing).

One more, scheduler-side: `_sdpa256_unfused_headroom` adopts this branch's admission-watermark target + same-shape cache credit + 2× charge, but **keeps main's abort-cap clamp** (the Metal async-OOM *crash* line), which the pre-rebase branch dropped. On wired-limit hardware the admission watermark can sit above that crash line and the 2× charge doesn't reliably cover the gap; the clamp only binds in exactly that case (regression test added). This is the reconciliation I'm least certain about — please sanity-check.

### Verification

**Full suite** (`pytest tests/ -q`, MLX 0.32.2 matching main's pin): **10849 passed, 152 skipped, 0 failed**. Targeted: `tests/test_sdpa256_attention.py` 62 passed (every test from both sides survives, adapted to the merged tuple-return / two-arg-provider / native-fused API); `test_memory_monitor.py` + `test_qwen4_qsa_prefill_memory.py` + `test_prefill_oom_graceful.py` 160 passed. Mechanical gate: reviewed the final delta vs `main` (no hunk reverts #3461/#3333/#3332/#3283 behavior) and grepped for dead/unreachable functions (none — this codebase has been bitten by that class before).

**Live gate** — real head-dim-256 prefill on a Qwen3.5/3.6/3.8-family model (Qwen3.8-27B-oQ4e-mtp, 24 q-heads / 4 kv / head_dim 256), driven through a server built from **this branch**:

```
sdpa256: route -> unfused (full call ~3168MiB fits ~3630MiB headroom at kv_len=8192)
sdpa256: route -> qsplit  (q_sub=3456 of q_len=4096 fits ~2977MiB headroom at kv_len=8192 (full-call transient ~3168MiB))
sdpa256: route -> tiled   (unfused transient ~4704MiB exceeds ~436MiB headroom at kv_len=12288 even at the q-split floor (128 rows))
```

- **(a) qsplit transition fires** on real Metal, with the fp32-priced sizing above — and `q_sub` is picked by `_max_q_sub_for_headroom` against the real fp32 footprint.
- **(b/d) ratchet holds**: once a request sheds to qsplit/tiled it never climbs back to unfused (observed qsplit→tiled, monotonic non-increase).
- **provider wiring intact live**: **zero** `probe-error` / `no-provider` reason codes across every run — this is the specific regression risk of the rebase (the provider signature changed zero-arg → `(kv_len, q_len)`; a mismatch would be swallowed by the never-raise gate and silently route everything to tiled). It doesn't.
- Real prefills up to **12,937 tokens** completed cleanly under the default context priority (200, no abort/evict); under memory pressure the eviction/reclaim ladder engaged benignly (reclaimed ~2.66 GB pooled buffers) and the request still completed.

**Honest scaling caveat:** the live box is a 64 GB Mac mini with a resident production server (~22.7 GB, head-dim-256 Qwen3.6-35B), so I ran the test server at a **synthetic low guard ceiling** (~28 GB) to place the qsplit band at kv≈8192 rather than the PR's headline 258k-on-35B scenario — the full-scale single-model run would have required disturbing the production server, which I would not do. Every routing decision here is ratio-driven (transient vs live headroom), so the scaled ceiling exercises the same `_route_decision`/ratchet/2×-charge/credit-keying paths; but I want to flag that **the true-258k memory behavior (multi-GB Metal pool retention across many chunks, the exact regime the retained-vs-active fix targets) is not proven by this scaled run.** I'd suggest one full-fidelity 258k-on-35B run under the real ceiling as the final pre-merge check.

### jackkbarnes' TTFT regression
Already resolved in the 2026-09-04 exchange above (not reproducible; the updated test set settled it) — just noting it still stands after the rebase, no action needed.

alytaphoenix:
Ping — both points from your Aug 24 review are addressed: sizing now bounds the *retained* footprint across every sub-tile rather than the active one, and the same-shape pool credit keys on the full `(kv_len, q_len)` call shape instead of `kv_len` alone, so interleaved chunks at equal KV length but different widths no longer credit each other.

Also worth noting @jackkbarnes withdrew the TTFT regression after controlled same-environment retesting. Rebased onto current main as `8f616f0e`.

PR #2992, opened 2026-08-21
fix(server): non-streaming JSON responses must not report 200 on a fast prefill-memory rejection

alytaphoenix:
## Summary

- `_with_json_keepalive` wraps non-streaming completions in a `StreamingResponse` so long prefills don't hit client read-timeouts, but it yielded its first keepalive byte — committing the ASGI response to whatever status `StreamingResponse` was built with (always 200, none of the 6 call sites passed an explicit `status_code`) — before knowing whether the wrapped request would succeed or fail. A memory-guard rejection that happens quickly (the common case for most prompt sizes) still surfaced a well-formed error body, but with a 200 status silently telling any client that only checks the status code that the request succeeded.
- Adds `_json_response_or_keepalive`, which races the wrapped coroutine against a short grace period. A fast outcome (success or `PrefillMemoryExceededError`) returns a plain `Response`/`JSONResponse` with the real status code, bypassing the streaming wrapper entirely — reusing the same body-construction helper the synchronous preflight-rejection path already uses. Only requests still running past the grace period fall back to the existing keepalive-streaming behavior, which still commits to 200 on any later failure once streaming has begun — an unavoidable HTTP/1.1 constraint once headers have shipped, not something this fix can close, but that only affects the far less common case of a failure minutes into a long-running request rather than the ordinary fast-reject case this fixes.
- Applies to all 6 previously-affected endpoints: markitdown completions, embeddings, completions, chat completions, Anthropic messages, and the Responses API — preserving engine-lease release semantics for the four that hold one.

## Test plan

- [x] `tests/test_server_prefill_memory_handler.py`: new tests directly covering the fast-fail-returns-400, fast-success-returns-200, lease-release-on-fast-fail, and slow-task-falls-back-to-streaming cases.
- [x] `tests/test_markitdown_integration.py`: updated/added tests covering both the new fast-path (plain response) and the existing slow-path (keepalive streaming) behavior.
- [x] Full repo test suite on this branch: 10,219 passed, 0 failed, 102 skipped (pre-existing, unrelated).

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -2311,21 +2311,37 @@ async def _run_with_disconnect_guard(
     return task.result()
 
 
+# Below this, prefer racing the coroutine to a real HTTP status code over
+# committing early to the keepalive-streaming fallback (see
+# _json_response_or_keepalive). Matches the keepalive loop's own
+# disconnect_poll cadence -- short enough that ordinary requests never
+# notice it, long enough that it isn't just a formality.
+_JSON_KEEPALIVE_GRACE_S = 2.0
+
+
 async def _with_json_keepalive(
     http_request: FastAPIRequest,
-    coro,
+    coro_or_task,
     interval: float = 10.0,
     disconnect_poll: float = 2.0,
 ) -> AsyncIterator[str]:
-    """Wrap a coroutine to send keepalive spaces while waiting for completion.
+    """Send keepalive spaces while waiting for a coroutine or task.
 
     For non-streaming requests, the HTTP response body is buffered until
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for tracking this down. Racing the coroutine for a short grace window before committing to the streaming wrapper is the right shape, it fixes the common fast-reject case while keeping the keepalive fallback for genuinely long prefills, and being upfront that a post-commit failure still can't revise the status line is appreciated. The lease handling is safe too since release is idempotent. Merging.

PR #2993, opened 2026-08-21
fix(vlm_mtp): keep Qwen thinking budgets on MTP

DiscoStew6082:
## Summary

- keep thinking-budget processors on VLM MTP for mRoPE adapters;
- run target verification through `VLMModelAdapter` so model-specific position IDs are preserved;
- expose only final hidden-to-logits projection while hiding token-producing speculative shortcuts;
- keep route eligibility and proxy capability checks on the same implementation seam.

## What was happening

When a Qwen VLM request included a thinking-token budget, oMLX honored the budget but could not keep the request on VLM MTP. It fell back to the slower `BatchGenerator` route.

The fallback happened because Qwen's VLM adapter correctly hides speculative shortcuts that would bypass mRoPE position handling, but it also hid the safe positioned-sampling operations needed to apply the thinking-budget processor during MTP verification.

## Why this PR changed

Since this PR was opened, `main` moved the shared MTP adapter plumbing into `VLMModelAdapter`.

The original version of this PR added Qwen-specific methods through the dense and MoE runtime patches. After rebasing, those patches detect the new shared adapter implementation and exit early, so the PR's added methods were no longer installed.

This update keeps the same user-facing fix but moves it to the existing `_VLMAdapterMTPProxy` seam. Verification still runs through `VLMModelAdapter` so mRoPE positions remain correct, while only the final logits projection is exposed. Shortcuts that could bypass the thinking-budget processor remain hidden.

The September 14 rebase also incorporates `main`'s fail-closed processor-contract handling from #3511. A `git range-diff` from pre-rebase head `130223ec` to current head `660dba94` shows no production-code change to this PR; only test context moved around the new contract-isolation coverage.

## Fix

Handle the capability at `_VLMAdapterMTPProxy`:

- target verification runs through `VLMModelAdapter`, preserving Qwen rank-3 and MiniMax rank-2 mRoPE position IDs;
- final `speculative_logits_from_hidden` projection may come from the adapter or inner language model;
- `speculative_argmax_from_hidden`, `speculative_verify_logits`, `.model`, and other token-producing speculative shortcuts remain hidden;
- `vlm_mtp_positioned_sampling_available()` asks the same proxy that owns dispatch, preventing route-gate drift.

## Validation

- Current-head hosted CI passed on Python 3.11, 3.12, and 3.13; each job reported 13,046 passed, 369 skipped, and 82 deselected.
- Regression coverage exercises the real `VLMModelAdapter -> proxy -> mlx-vlm _mtp_rounds -> MTPProcessingSampler` chain.
- Qwen rank-3 and MiniMax rank-2 mRoPE position layouts are covered.
- Tests verify that thinking budgets stay enforced while processor-bypassing speculative shortcuts remain unavailable.

Related to #2399.


DiscoStew6082:
Some context on what prompted this change:

Pi recently generalized thinking-budget support for OpenAI-compatible providers in [badlogic/pi-mono#8275](https://github.com/badlogic/pi-mono/pull/8275). A provider can now map Pi's per-level reasoning budgets to its native top-level request field. For oMLX, setting `thinkingTokenBudgetField: "thinking_budget"` makes Pi send a numeric `thinking_budget` alongside `reasoning_effort`.

That Pi change did not introduce a correctness bug in oMLX. The existing #2399 and #2456 handling recognized the budget processor and enforced it. The issue was route eligibility: Qwen VLM's mRoPE adapter intentionally hides the inner model's speculative fast paths to preserve position handling. Because the adapter did not expose a safe positioned-logits hook, oMLX declined VLM MTP and sent every budgeted Qwen request through `BatchGenerator`.

The budget was still honored, but applying a hard reasoning limit from Pi disabled Qwen's VLM MTP acceleration.

This PR adds only the two mRoPE-safe adapter operations needed by the positioned sampler: hidden-state verification through `VLMModelAdapter`, followed by final logits projection through the language model. It continues to hide the pre-sampling shortcuts, so `MTPProcessingSampler` can apply the budget at each verification position.

In the controlled Pi 0.84.2 regression matrix, all 18 requests stayed on VLM MTP, including all 15 requests carrying `thinking_budget`. There was no positioned-sampling or `BatchGenerator` fallback. The budgeted median was 56.8 tok/s versus 31.0 tok/s in the earlier fallback run. That throughput comparison includes other checkout and environment differences, so I am treating it as route evidence rather than a strict patch-only benchmark.

Issue #2995, opened 2026-08-21
MTP performance handoff is permanently latched after a transient depth-0 period


Anionex:
## Summary

The current MTP performance handoff permanently disables MTP for the
remainder of a request after the adaptive controller remains at depth 0 long
enough to trigger `_park_mtp_to_standard()`.

This appears too aggressive for long generations because MTP acceptance and
profitability can change substantially between different output segments. A
transient low-acceptance segment should not necessarily imply that MTP will
remain unprofitable for the rest of a 10K-30K-token generation.

Related symptom report: #2976. This issue is specifically about the one-way
parking policy rather than general long-generation throughput.

## Current behavior

In
[`omlx/patches/mlx_lm_mtp/batch_generator.py`](https://github.com/jundot/omlx/blob/fa3e94b/omlx/patches/mlx_lm_mtp/batch_generator.py#L2609-L2626),
`_park_mtp_to_standard()` sets:

```python
gen_batch._omlx_mtp_parked_uid = state.uid
```

The source comment says the marker prevents MTP from reactivating for that
sequence because the controller has already determined that speculation loses.
However, that conclusion is based on the recent controller window, not the
future output distribution. Once the MTP state has been dropped, the controller
can no longer observe whether MTP becomes profitable again.

## Why this matters

Long real-world generations are non-stationary. One response may move through
several different token distributions:

- natural-language reasoning
- repetitive markup
- source code
- structured data
- boilerplate
- final explanatory prose

MTP may perform poorly during one segment but recover during a later, more
predictable segment. The current irreversible handoff converts a local
performance decision into a request-long decision.

## Experiment

I compared the current permanent handoff against a local always-on policy. The
local policy only suppresses the performance-driven call to
`_park_mtp_to_standard()`. It preserves:

- adaptive depth selection
- depth 0
- the controller's bidirectional depth probes
- correctness fallbacks
- late-join handoffs

Environment:

- oMLX upstream base: `fa3e94b`
- Apple M5 Pro, 48 GB unified memory
- macOS 26.6
- context window: 131,072
- `max_tokens`: 32,768
- `temperature`: 1.0
- `top_p`: 0.95
- thinking budgets: 8,192 and 16,384
- singleton requests
- models:
  - `Qwen3.8-27B-MTPLX-Optimized-Speed`
  - `Qwen3.8-27B-oQ4e-mtp`
- six paired long-generation runs

Results:

| Metric | Permanent handoff | Always-on adaptive MTP |
[...]

PR #2998, opened 2026-08-21
feat(mtp): add opt-in always-on handoff policy

Anionex:
## Summary

- add an opt-in `OMLX_MTP_ALWAYS_ON=1` policy for singleton adaptive MTP
- suppress only the performance-driven permanent handoff after sustained depth 0
- preserve depth 0, adaptive depth selection, bidirectional probes, correctness fallbacks, and late-join handoffs
- keep the existing permanent handoff as the default behavior
- log the suppressed handoff once per MTP state

Refs #2995.

## Motivation

`_park_mtp_to_standard()` currently records `_omlx_mtp_parked_uid`, so a
performance decision based on one sustained depth-0 window prevents MTP from
reactivating for the rest of that request. Long generations can move between
reasoning, prose, markup, and code segments with materially different draft
acceptance, so an opt-in policy is useful for workloads where a locally bad
segment should not permanently disable later MTP recovery.

This PR does not change the default because standard pipelined decoding may
still be preferable on other models and hardware. It provides a narrow policy
switch backed by long-generation measurements while #2995 discusses whether a
future reversible parking policy should become the broader solution.

## Behavior

Default, unchanged:

```text
sustained losing speculation -> _park_mtp_to_standard() -> standard decoder
```

Opt-in:

```bash
OMLX_MTP_ALWAYS_ON=1
```

```text
sustained losing speculation -> keep adaptive controller alive at depth 0
                              -> continue normal probes
                              -> allow speculative depths to recover later
```

The opt-in path emits one operational log line:

```text
MTP[<uid>] standard handoff suppressed: always-on mode
```

## Long-generation measurements

Environment:

- Apple M5 Pro, 48 GB unified memory
- macOS 26.6
- context window: 131,072
- `max_tokens`: 32,768
- `temperature`: 1.0
- `top_p`: 0.95
- thinking budgets: 8,192 and 16,384
- singleton requests
- `Qwen3.8-27B-MTPLX-Optimized-Speed`
- `Qwen3.8-27B-oQ4e-mtp`

Six paired long-generation runs:

| Metric | Permanent handoff | Always-on adaptive MTP |
|---|---:|---:|
| Average generation TPS | 13.32 | 16.53 |
| Adjusted total wall time | 10,091.64 s | 8,469.68 s |
| MTP coverage | 22.44% | approximately 100% |
| Weighted acceptance rate | 67.76% | 79.72% |
| Significant mid-run permanent handoffs | 6 | 0 |

Per-run TPS changes were `+25.70%`, `+57.49%`, `+31.05%`, `+15.49%`,
`+16.17%`, and `+8.84%`. Average TPS increased by 24.05%; adjusted total wall
time decreased by 16.07%.

These are independent `temperature=1.0` generations rather than deterministic
token-for-token replays. The measurements are included as performance evidence,
not as a claim about output-quality changes.

## Tests

- `python -m py_compile omlx/patches/mlx_lm_mtp/batch_generator.py tests/test_mlx_lm_mtp_patch.py`
- `pytest tests/test_mlx_lm_mtp_patch.py -q` — **142 passed**
- full `pytest tests/ -m "not slow and not integration" -q` — **10,164 passed, 125 skipped**
  - the remaining 3 failures and 37 errors were reproduced on unmodified
    `origin/main`: one DFlash numerical-tolerance failure and Harmony tests that
    could not download/load their vocabulary file
- changed-line Ruff inspection — no findings in the added code

This is a non-UI decoder-policy change, so there is no meaningful browser E2E
path. Runtime verification was performed through the OpenAI-compatible endpoint
with the six long-generation runs summarized above.

### DIFF
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -404,6 +404,7 @@ def _mtp_common_eligible(gen_batch: Any) -> bool:
 
 
 _ROWWISE_BATCH_MTP_ENV = "OMLX_MTP_ROWWISE_BATCH"
+_MTP_ALWAYS_ON_ENV = "OMLX_MTP_ALWAYS_ON"
 
 
 def _rowwise_batch_mtp_enabled() -> bool:
@@ -419,6 +420,16 @@ def _rowwise_batch_mtp_enabled() -> bool:
     unless explicitly requested.
     """
     return os.environ.get(_ROWWISE_BATCH_MTP_ENV, "").strip().lower() in (
+        "1",
+        "true",
+        "yes",
+        "on",
+    )
+
+
+def _mtp_always_on_enabled() -> bool:
+    """Keep adaptive MTP eligible after a sustained depth-0 period."""
+    return os.environ.get(_MTP_ALWAYS_ON_ENV, "").strip().lower() in (
         "1",
         "true",
         "yes",
@@ -672,6 +683,10 @@ class _MtpState:
     # MTP state is valid only for this exact singleton uid. It must be dropped
     # across any standard batched step or batch reshape that breaks ownership.
     uid: Any = None
+
+    # Avoid repeating the operational log every depth-0 cycle when the
+    # optional always-on policy suppresses the standard-decoder handoff.
+    always_on_handoff_logged: bool = False
 
     # Pending tokens to emit in upcoming next() calls. Each entry is
     # (token_id_int, logprobs_1d, source_label). source_label is one of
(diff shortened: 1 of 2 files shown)


Anionex:
Closing as superseded by #3004, which implements the preferred reversible parking policy with a 128-token cooldown, safe MTP re-entry probing, and bounded exponential backoff.

I reran the six original long-generation cases that had permanently handed off to standard decoding. On current upstream `main`, all 3 naturally occurring re-entry probes succeeded, there were 0 failed probes and 0 repeated parks, and mean generation throughput was 18.62 tok/s versus 16.53 tok/s for this PR's always-on experiment (+12.69%). Full results are posted on #2995.

Thanks for addressing the underlying issue with the more complete adaptive solution.

PR #2999, opened 2026-08-21
fix(ane): circuit-break the procedure bank compile retry ladder before it OOMs

alytaphoenix:
## Summary

- Live diagnosis on a 48GB peer Mac: loading `Qwen3.8-27B-oQ4e-mtp` with dual-ANE prefill enabled triggered a monolithic procedure bank compile failure, which retried through the split ladder at progressively smaller bank sizes (15GB → 7.86GB → 1GB → 512MB). Each failed attempt drops its Python references so the ANE driver can release the device mappings, but that release is asynchronous and was observed lagging the retry (kernel log: `ReleaseProgramResource: WARN: waitForPendingUpdate failed`). The retry ladder has no way to see that lag, so it kept issuing further (smaller) attempts on top of not-yet-reclaimed memory from the previous one.
- Process RSS reached 48.3GB and the kernel's jetsam killer force-killed it directly (`memorystatus: killing largest compressed process python3.11 [58008] 48288 MB`) — bypassing oMLX's own memory guard entirely, since jetsam watches total process footprint while the guard only watches Metal-attributed allocations. The server died silently: no crash report, no clean shutdown log, just gone.
- Adds a circuit breaker shared by both compile-retry ladders (the now-shared `_bank_split_ladder` used by the dual-ANE and streaming-builder paths, and the still-separate single-ANE calibration path): before every attempt (including the first), check `phys_footprint` — jetsam's own ledger — against 70% of total system memory, and stop retrying in favor of the per-layer fallback once it's past that line rather than continuing to shrink-and-retry blind. Between failed attempts, force a GC pass and a brief settle delay so the driver's asynchronous cleanup has a real chance to land before the next attempt's headroom check measures it.

## Test plan

- [x] `tests/test_qwen35_ane_prefill.py` (112 tests): new tests for the headroom-math circuit breaker, the ladder stopping retries once headroom runs out (covering both the shared `_bank_split_ladder` and the separate `_compile_single_banks` copy), and the GC+settle behavior between failed attempts. Existing split-ladder/retry tests updated to mock the new gate for determinism and speed.
- [x] Full repo test suite on this branch: 10,219 passed, 0 failed, 102 skipped (pre-existing, unrelated).

### DIFF
--- a/omlx/patches/qwen35_ane_prefill.py
+++ b/omlx/patches/qwen35_ane_prefill.py
@@ -75,6 +75,21 @@ def _ane_bank_memory_headroom_ok() -> bool:
         return True
 
 
+def _ane_bank_memory_footprint_snapshot() -> tuple[int, int]:
+    """``(current phys_footprint, total system memory)`` in bytes for the
+    skip-warning log, so "why did ANE bank compile stop retrying on this
+    box" is answerable from one log line instead of just the fixed 70%
+    threshold. Best-effort: returns ``(0, 0)`` if either measurement is
+    unavailable."""
+    try:
+        from omlx.utils.hardware import get_total_memory_bytes
+        from omlx.utils.proc_memory import get_phys_footprint
+
+        return get_phys_footprint(), get_total_memory_bytes()
+    except Exception:
+        return 0, 0
+
+
 @dataclass(frozen=True)
 class _AnePrefillConfig:
     sequence_length: int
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for pinning this on the jetsam ledger, checking phys_footprint before each retry is the right spot to break the ladder. Merging. I'll follow up on main with small cleanups: imports moved to the top, the two snapshot helpers folded into one, and the skip log wording when nothing actually ran.

PR #3001, opened 2026-08-21
Superseded: Cluster v2 uses signed membership

xunlinkx:
Superseded by Cluster v2.

The retired dashboard visualizer has been replaced by the Cluster v2 wizard. Active membership is derived from the deployment's signed assignments, and membership expansion generates a new signed plan.

Rebasing leaves no code change to carry; the current three-Mac membership regression covers the active flow.

### DIFF
--- a/omlx/admin/hf_downloader.py
+++ b/omlx/admin/hf_downloader.py
@@ -365,13 +365,17 @@ def _sum_safetensors_blob_bytes(siblings) -> int | None:
 def _fetch_safetensors_blob_bytes(api: HfApi, repo_id: str) -> int:
     """Look up *.safetensors blob bytes for one repo. 0 on failure."""
     try:
-        info = api.model_info(repo_id, files_metadata=True)
+        info = api.model_info(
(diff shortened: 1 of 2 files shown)


xunlinkx:
Pushed 2e75dbea for all three points.

Deployment selection goes through clusterLiveDeployment() now — it matches the live job's deployment_id and only falls back to the first entry when the status payload omits it. Both membership derivation and the aligned node list use it.

clusterFabricHostKeys() drops user@ prefixes the same way split('@').pop() already does elsewhere in this file, so destinations like ops@worker-a.local line up with host id worker-a. Keys stay a superset (raw plus stripped forms), so nothing that matched before stops matching.

Blank node_ids stay out of the lookup map and never match assignments either.

While wiring executed coverage for this I hit a real bug: the restoration loop called .some() on a Set, which throws at runtime. It spreads to an array first now. The connectx one-click test also needed the new helpers in its method list — that's what went red in CI.

xunlinkx:
@copilot-pull-request-reviewer please review the latest branch state; the previous comments are addressed.

PR #3002, opened 2026-08-21
fix(ane): bound every wait on the ANE evaluation path and latch timed-out programs

### LINKED ISSUE #2974: Hang during ANE prefill: spin-wait on IOSurfaceSharedEvent never signaled in AneLinearModel::Impl::evaluate_and_signal
Title: Hang during ANE prefill: spin-wait on IOSurfaceSharedEvent never signaled in AneLinearModel::Impl::evaluate_and_signal

## Summary

During prefill with Qwen3.5 ANE Prompt Processing enabled, the server hangs indefinitely (no progress, ~100% CPU, no timeout/error) shortly after restoring a large static prefix from the tiered cache. Reproduced twice on the same machine (M4 Max 64GB), independent of the `dual_ane` setting.

## Environment

- oMLX version: 0.6.3rc2
- Hardware: M4 Max 64GB (single-die, 1 physical ANE)
- Feature: Qwen3.5 ANE Prompt Processing (experimental)

## Reproduction

Both hangs occurred right after this log line, with no further progress:

```
SpecPrefill: restored 25994 static system-prefix tokens from tiered cache
```

- 1st occurrence: default settings.
- 2nd occurrence: `dual_ane=False` (log: `ANE fraction=0.400, dual_ane=False`) — ruling out dual-ANE as the cause (that setting affects a separate, already-diagnosed memory/swap issue on single-die Macs, unrelated to this hang).

## Diagnosis (via `sample`/spindump on the hung process)

Two spindumps taken during separate hangs show the same pattern:

- The main thread and all `mlx::core::scheduler::StreamThread` instances are correctly blocked on `pthread_cond_wait` (expected idle state).
- One worker thread — running the lambda from `AneHybridQ4Primitive::eval_gpu` (in `libomlx_qwen35_prefill_kernel_ops.dylib`) — is **not** blocked. It is busy-spinning inside `AneLinearModel::Impl::evaluate_and_signal`:

```
AneLinearModel::Impl::evaluate_and_signal(...)  [+60, 1516/1548 samples]
  → cthread_yield → swtch_pri                    (the spin itself — repeated sched_yield)
AneLinearModel::Impl::evaluate_and_signal(...)  [+48, 29/1548 samples]
  → objc_msgSend$signaledValue
    → -[IOSurfaceSharedEvent signaledValue]        (polling the completion event)
```

This thread is actively polling `IOSurfaceSharedEvent::signaledValue` with `sched_yield()` between checks, waiting for the ANE to signal that it finished executing the compiled program — and that signal never arrives. There is no timeout or fallback path visible in the sampled stack; the spin appears unbounded.

`AneLinearModel` / `AneHybridQ4Primitive::eval_gpu` / `evaluate_and_signal` live in `libomlx_qwen35_prefill_kernel_ops.dylib` (custom_kernels/qwen35_prefill) — I don't have source access to this component, so I can't inspect whether there's a configurable timeout or confirm what happens on the ANE side when the event is never delivered.

## Hypothesis (unconfirmed)

Both hangs happened right after restoring a large static prefix (25994 tokens) from the tiered cache. Possible causes:
- The ANE driver/firmware silently fails to deliver the completion event for this specific large prefill workload (internal watchdog/timeout on a very large compiled program).
- Some interaction between the tiered prefix-cache restore path and the compiled ANE program state.

Not confirmed — would need additional instrumentation around program submission to the ANE to see whether the event is lost before or after submission.

## Suggested directions

- Add a timeout/fallback around `evaluate_and_signal`'s wait so a stuck ANE completion event degrades to an error (or CPU/GPU fallback) instead of hanging the process forever.
- If there's an existing timeout knob for ANE prefill (e.g. related to `specprefill_threshold` or max static-prefix size), documenting it would help others avoid this path in the meantime.

Happy to provide the full spindump output or test a patch if useful.

beaglemoo:
Fixes #2974.

## Problem

Every wait on the ANE live request path is unbounded: the readiness spin in
`evaluate_and_signal()`, the caller's `wait()`, and `warmup()`'s pre-wait. The
only bounded wait in the file is the destructor's 10s `wait_for`. When the ANE
driver fails to signal completion once (#2974: parked around
`evaluateWithQoS:` right after a large tiered-cache restore), the detached
evaluation thread spins a core at 100% forever, the request thread sleeps
forever, and the server is permanently hung. The sticky-error mechanism that
would quarantine the program already exists in `begin()` — but nothing ever
sets it on a hang, and even when it fires, the resulting evaluation-time
exception is invisible to the per-module fallback latch from #2946, which only
wraps graph construction.

## Change

`csrc/qwen35_ane.mm` / `.h` / `bindings.cpp`, `patches/qwen35_ane_prefill.py`:

- One timeout ceiling, `ane_wait_timeout()` (default 30s,
  `OMLX_ANE_WAIT_TIMEOUT_S` to override, `0` restores the old unbounded
  behavior), applied to:
  - the readiness spin — on expiry, record the error, mark the ticket
    complete, signal `done`, wake waiters, and skip the evaluation;
  - `wait()` — `wait_for` (the destructor's own idiom); on expiry, latch
    `completion_error_` and throw. The stuck ObjC frame can't be unblocked —
    the job is to stop the caller sleeping forever and quarantine the program,
    the same trade the destructor already accepts;
  - `warmup()`'s pre-wait, so a wedged program can't hang a later load;
  - `begin()` when counters are stuck — one bounded grace for a legitimately
    in-flight evaluation, then latch, so a stuck counter can't stay
    indistinguishable from a race.
- `AneLinearModel.has_error()` exposed through the bindings, and a
  `_raise_if_latched()` check at the top of the three dispatch sites in
  `qwen35_ane_prefill.py`. This makes a latched program visible at
  graph-construction time, inside the existing per-module try/except from
  #2946 — so the module falls back to GPU permanently instead of every
  subsequent request failing at evaluation time.

Healthy-path cost is zero: the readiness spin resolves in microseconds (the
producer buffer is committed and waited on before the evaluation thread is
spawned), evaluations in milliseconds, and `has_error()` is one mutex-guarded
flag read per layer per prefill call.

## Verification (M5 Pro 64GB, Qwen3.8-27B oQ4e, classic split 0.35/0.45)

**Healthy path**: warmed 16K real request, TTFT 37.5s — inside the same
session's 34.8-36.6s band for this config on unpatched main; `Warmed 224 ANE
procedures in 1.9s at load` unchanged; `pytest tests/test_qwen35_ane_prefill.py`
green (the 11 pre-existing `test_ane_tuning.py` failures on current main are
unchanged by this PR).

**Injected wedge** (temporary, not in this PR: env-gated early-return making
the nth runtime evaluation never complete or signal — the #2974 driver state,
with `OMLX_ANE_WAIT_TIMEOUT_S=5`):

| | before this PR | with this PR |
|---|---|---|
| request hitting the wedge | hangs forever, one core at 100% | fails in 5.6s with "ANE evaluation timed out" |
| subsequent requests | fail forever ("Prior ANE evaluation failed" / "overlapping evaluations") | succeed in ~9-10s, wedged module on GPU fallback |
| server | permanently hung | alive, serving |

The serve log shows the intended chain: timeout -> latch -> `_raise_if_latched`
at construction -> module fallback -> service continues.

One observation left alone deliberately: `AneLinearModel::end()` (the
GPU-encoded wait on `done`) has no callers — the CPU-side `wait()` took over
per the comment in the eval path. Happy to remove it here or leave it for a
cleanup pass, your call.

### DIFF
--- a/omlx/custom_kernels/qwen35_prefill/csrc/bindings.cpp
+++ b/omlx/custom_kernels/qwen35_prefill/csrc/bindings.cpp
@@ -62,7 +62,10 @@ NB_MODULE(_ext, m) {
       .def(
           "warmup",
           &omlx::qwen35_prefill_kernels::AneLinearModel::warmup,
-          nb::call_guard<nb::gil_scoped_release>());
+          nb::call_guard<nb::gil_scoped_release>())
+      .def(
+          "has_error",
+          &omlx::qwen35_prefill_kernels::AneLinearModel::has_error);
   nb::class_<omlx::qwen35_prefill_kernels::AneLinearBankBuilder>(
       m, "AneLinearBankBuilder")
       .def(nb::init<int>(), "sequence_length"_a)
--- a/omlx/custom_kernels/qwen35_prefill/csrc/qwen35_ane.h
+++ b/omlx/custom_kernels/qwen35_prefill/csrc/qwen35_ane.h
@@ -38,6 +38,10 @@ public:
   Ticket begin(MTL::CommandBuffer *command_buffer);
   void execute(Ticket ticket);
   void wait(Ticket ticket);
+  // True once an evaluation has failed or timed out on this program; the
+  // program is latched and every later begin() will throw. Lets the Python
+  // layer detect a wedged program at graph-construction time and fall back.
+  bool has_error();
   // Run one throwaway evaluation to pay the first-run compilation cost at
   // load time instead of inside the first user request. Input contents are
   // irrelevant; the output is discarded.
--- a/omlx/custom_kernels/qwen35_prefill/csrc/qwen35_ane.h
+++ b/omlx/custom_kernels/qwen35_prefill/csrc/qwen35_ane.h
@@ -38,6 +38,11 @@ public:
   Ticket begin(MTL::CommandBuffer *command_buffer);
   void execute(Ticket ticket);
   void wait(Ticket ticket);
+  // Retire a ticket whose producer command buffer failed before execute()
+  // was ever called: balances the submission counters and wakes waiters
+  // without latching the program (the failure is per-submission, typically
+  // a request abort, not a program fault).
+  void cancel_ticket(Ticket ticket);
   // True once an evaluation has failed or timed out on this program; the
   // program is latched and every later begin() will throw. Lets the Python
   // layer detect a wedged program at graph-construction time and fall back.
--- a/omlx/custom_kernels/qwen35_prefill/csrc/qwen35_ane.mm
+++ b/omlx/custom_kernels/qwen35_prefill/csrc/qwen35_ane.mm
@@ -963,6 +963,22 @@ public:
     [buffer encodeWaitForEvent:event_ value:ticket.done];
   }
 
+  void cancel_ticket(AneLinearModel::Ticket ticket) {
+    // The ticket's producer command buffer failed, so its encoded ready
+    // signal will never fire. Retire the ticket before any evaluation thread
+    // is spawned: balance the counters, publish the done value, and wake any
+    // waiter. This is a per-submission failure (typically a request abort
+    // tearing down Metal state mid-prefill), not a program fault, so
+    // completion_error_ is left clear and the next request may use this
+    // program again.
+    {
+      std::lock_guard<std::mutex> lock(state_mutex_);
+      ++completed_;
+    }
+    [event_ setSignaledValue:ticket.done];
+    completion_cv_.notify_all();
+  }
+
   void wait(AneLinearModel::Ticket) {
     std::unique_lock<std::mutex> lock(state_mutex_);
     const bool finished = completion_cv_.wait_for(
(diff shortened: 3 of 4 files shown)


beaglemoo:
Extended per the request in #3025: a second commit adds eager ticket cancellation when the pack command buffer itself fails (the abort-mid-prefill trigger there). The producer buffer's status is checked after the existing `waitUntilCompleted()` at all three dispatch sites; on error the tickets are retired and the dispatch fails before any evaluation thread is spawned — immediate, and non-latching, since a failed buffer is a per-submission fault, not a program fault. The timeout backstop in the first commit remains the safety net for the never-signaled-event case (#2974), which no status check can see coming.

Verified on M5 Pro with an injected buffer failure: requests fail fast with a clear error instead of spinning workers, repeated failures never wedge the submission counters, server CPU stays idle, and healthy serving is unaffected (224 procedures warmed, dual programs). A real client abort 4 seconds into a 24K-token prefill left the server idle with the next request succeeding.

jundot:
Thanks for adding a bounded wait and GPU fallback for wedged ANE requests. This is a meaningful improvement over the current unbounded wait behavior.

I found one cleanup gap in the dual-program paths: if the first `begin()` succeeds and the second `begin()` throws, the first ticket remains submitted without being completed or canceled. This is limited to the partial-acquisition failure path, so it does not outweigh the value of this fix. I’ll merge this and fold atomic ticket cleanup plus regression coverage into an immediate follow-up.

PR #3003, opened 2026-08-21
feat(dflash): expose effective runtime state

### LINKED ISSUE #2896: DFlash 2: fallback is silent when dflash_enabled is set (v0.6.3rc1 fixed the packaging half)
## Summary

**Update (resolved in part):** `v0.6.3rc1` ships the DFlash 2 runtime and
`dflash_block_size`, so the original packaging complaint below is fixed. Verified
on the official `cp313` wheel: a v2 draft checkpoint loads (`DFlashEngine
loaded`, no fallback) and the block-size override takes effect.

**What remains:** when `dflash_enabled` is set explicitly and the engine fails to
start, oMLX logs the failure, falls back to the batched/VLM engine, and then
serves normally with `200`. From an API client there is no way to tell DFlash
was requested but is not running. That is what led me to file this at all — on
`0.6.2` I measured for a while before noticing the log line.

Suggested options: fail the request, or surface the effective engine per model
(for example in `/v1/models` or the model status endpoint) so a client can assert
it.

## Original report (fixed by v0.6.3rc1)

The published `v0.6.2` wheels pinned
`dflash-mlx @ git+https://github.com/jundot/dflash-mlx@2eb169f` (`0.1.10+omlx.5`),
which has no DFlash 2 modules, so a v2 checkpoint produced:

```
DFlash start failed for <model>: Received parameters not in model:
  candidate_selector.hidden_projection.weight,
  candidate_selector.predecessor_codebook,
  candidate_selector.successor_codebook,
  layers.0.attention_conv.base_kernel,
  layers.0.attention_conv.kernel_projection.weight,
  layers.0.mlp_conv.base_kernel,
  layers.0.mlp_conv.kernel_projection.weight,
  ... (layers.1-4 likewise) ..
  Falling back to vlm engine.
```

`v0.6.2` also had no `dflash_block_size`, so a quantized target drafted at the
checkpoint's `block_size` (8), which on this hardware is the difference between
no gain and a real one. Both are fixed on `v0.6.3rc1`.

## Measurements (may be useful regardless)

Apple M1 Ultra, 128 GB, macOS 26.5.2, `Qwen3.8-27B` oQ4e-mtp mixed 4/5-bit
target, `z-lab/Qwen3.8-27B-DFlash2` draft, `dflash_block_size: 5`, greedy.

| regime | baseline | DFlash 2 | ratio |
| --- | --- | --- | --- |
| short prompt, novel output, 1 stream | 21.78 | 24.45 | 1.12x |
| short prompt, copy/echo output | 19.9 | 24.5 | 1.23x |
| short prompt, concurrency 2 | 24.85 | 24.35 | 0.98x |
| ~14K prompt, novel output, 1 stream | 21.4 | 16.7 | **0.78x** |

Two things in there differ from the M3 Ultra numbers in the release notes, which
show the speedup *growing* with context (1.33x at 4K, 1.43x at 16K):

1. On this machine the ratio **inverts** with context — 1.12x at a few hundred
   tokens, 0.78x at ~14K.
2. Continuous batching at concurrency 2 recovers more than DFlash 2 does at
   concurrency 1, and the two do not stack.

Mechanism, measured directly: a forward pass over 5 positions costs ~3.7x a
1-position pass on 4-bit weights here (~2.7x on 8-bit), so with ~4.2 accepted
tokens the arithmetic gives ~1.15x, which is what the end-to-end run shows.
Where a wide verify is nearly free the same acceptance yields a much larger
speedup. So DFlash 2 appears to be strongly hardware- and workload-dependent on
Apple Silicon, and the M1 generation looks like the unfavourable end of it.

Happy to run further variations if useful.

## Environment

- oMLX `0.6.2` (original report) and `0.6.3rc1` (verification), `cp313` wheels
- `dflash-mlx 0.1.10+omlx.5` then `0.1.10+omlx.6`
- Apple M1 Ultra, 128 GB, macOS 26.5.2

joshuaswarren:
## Summary

Expose requested versus effective DFlash state through the existing authenticated `GET /v1/models/status` response.

DFlash intentionally fails soft: when startup fails, oMLX serves the model through its natural batched/VLM engine and still returns `200`. It can also switch permanently to that inner fallback after a successful load for long-context or multimodal requests. Until now an API client could not distinguish either fallback from an active DFlash engine.

Each model status row now adds:

- `effective_engine`: the backend serving the model now
- `dflash_requested`: whether DFlash was requested for this load
- `dflash_active`: whether DFlash is currently serving
- `dflash_fallback_reason`: a bounded, path-redacted reason when requested DFlash is not active

No endpoint, dependency, or fallback policy changes. Existing clients receive an additive payload, and normal fallback continues serving successfully.

## Runtime state

The status follows both load-time and post-load behavior:

- DFlash starts successfully -> `effective_engine=dflash`, `dflash_active=true`
- missing draft/dependency, unsupported deployment, init/start failure -> natural engine, `dflash_active=false`, reason populated
- long-context or multimodal runtime fallback -> live natural engine, `dflash_active=false`, trigger populated
- DFlash not requested -> requested/active false, no reason
- unload -> effective state cleared

`DFlashEngine` exposes read-only fallback state; the pool does not inspect private engine fields. Runtime fallback triggers must supply a reason, so a new trigger cannot silently omit observability.

## Privacy and bounds

Fallback text crosses an API boundary. It is therefore whitespace-collapsed, filesystem paths are replaced with `<path>`, and output is limited to 200 characters. Static reasons are used where possible; stack traces are not exposed.

Example fallback:

```json
{
  "id": "Qwen3.8-27B",
  "loaded": true,
  "engine_type": "batched",
  "effective_engine": "batched",
  "dflash_requested": true,
  "dflash_active": false,
  "dflash_fallback_reason": "DFlash runtime fallback: prompt of 14000 tokens reached the DFlash context limit of 8192"
}
```

Client assertion:

```sh
curl -s localhost:10240/v1/models/status |
  jq -e '.models[] | select(.id=="Qwen3.8-27B") | .dflash_active'
```

## Verification

On macOS / Python 3.13:

- 324 focused engine-pool, DFlash lifecycle, multimodal, memory-guard, and status-endpoint tests passed
- final changed-surface rerun: 40 passed
- ruff clean on `dflash.py` and the changed DFlash tests
- independent spec-compliance review passed
- independent code-quality review found a post-load fallback gap; fixed and re-reviewed with no remaining P0-P2 findings

The remaining whole-file ruff findings in `engine_pool.py` reproduce on upstream code and were deliberately not expanded into this PR.

Closes #2896.

### DIFF
--- a/docs/experimental/dflash_mlx_integration.md
+++ b/docs/experimental/dflash_mlx_integration.md
@@ -265,6 +265,75 @@ DFlashEngine.start() fails:
 
 DFlash check runs **before** engine type routing in `_load_engine()`. If `dflash_enabled=True` and `dflash_draft_model` is set, DFlashEngine is created regardless of whether the model would normally be VLM or LLM. On failure, falls back to the model's natural engine type.
 
+### Verifying the effective engine
+
+A DFlash load fails soft, and so does a DFlash *request*: a long prompt or an
+image evicts the DFlash runtime after a successful load and the engine stays on
+its fallback engine until the model reloads. Either way the request succeeds
+with `200`, served by the model's natural engine. `GET /v1/models/status`
+reports what actually happened, so a client can assert DFlash is running
+instead of reading the server log.
+
+| Field | Type | Meaning |
+|-------|------|---------|
+| `engine_type` | str | Configured engine for the model (unchanged) |
+| `effective_engine` | str \| null | Engine serving the model now: `dflash`, `vlm`, `batched`, `distributed`, `embedding`, `reranker`, `audio_stt`, `audio_tts`, `audio_sts`. `null` when not loaded |
+| `dflash_requested` | bool | `dflash_enabled` was set for this load |
+| `dflash_active` | bool | DFlash is the engine serving the model right now |
+| `dflash_fallback_reason` | str \| null | Why a requested DFlash load is not active, at startup or after a runtime fallback. Redacted and truncated to 200 characters |
+
+DFlash running:
+
+```json
+{
+  "id": "Qwen3.8-27B",
+  "loaded": true,
+  "engine_type": "batched",
+  "effective_engine": "dflash",
+  "dflash_requested": true,
+  "dflash_active": true,
+  "dflash_fallback_reason": null
+}
+```
+
+DFlash requested but fell back:
+
+```json
+{
+  "id": "Qwen3.8-27B",
+  "loaded": true,
+  "engine_type": "batched",
+  "effective_engine": "batched",
+  "dflash_requested": true,
+  "dflash_active": false,
+  "dflash_fallback_reason": "DFlash start failed: ValueError: Received parameters not in model: candidate_selector.hidden_projection.weight, ..."
+}
+```
+
+DFlash active at load, then dropped at runtime (`dflash_max_ctx` exceeded):
+
+```json
+{
+  "id": "Qwen3.8-27B",
+  "loaded": true,
+  "engine_type": "batched",
+  "effective_engine": "batched",
+  "dflash_requested": true,
+  "dflash_active": false,
+  "dflash_fallback_reason": "DFlash runtime fallback: prompt of 14000 tokens reached the DFlash context limit of 8192"
+}
+```
+
+A one-line client assertion:
+
+```bash
+curl -s localhost:10240/v1/models/status \
+  | jq -e '.models[] | select(.id=="Qwen3.8-27B") | .dflash_active'
+```
+
+DFlash was never requested when `dflash_requested` is `false`; `dflash_active`
+is then `false` and `dflash_fallback_reason` is `null`.
+
 ---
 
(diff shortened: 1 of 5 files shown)


joshuaswarren:
Rebased onto `main` @ `2df39bfc` and fixed the one real issue I found in self-review.

**Change:** the fallback-reason path regex excluded whitespace, so a home path with a space (`/Users/Joshua Warren/...`) leaked unredacted into `/v1/models/status`. Replaced it with a delimiter-bounded matcher that redacts spaced and quoted paths while leaving non-paths like `a/b` alone.

**Tests:** added three regression cases (spaced path, quoted path, ratio non-path). `tests/test_dflash_effective_state.py` (21) and `tests/test_dflash_multimodal_fallback.py` (22) both pass on macOS 15 / Python 3.13.

No other changes. Ready for review.

Issue #2909, opened 2026-08-20
Native MTP drafter state re-primed on every request (~7.5s at 30K ctx) even when prefix/boundary cache fully hits


cuiyuxin-gif:
## Environment

- Mac Studio M3 Ultra (60-core GPU, 96 GB), macOS 26.4.1, oMLX **0.6.1**
- `Qwen3.8-27B-oQ4e-mtp` with `mtp_enabled: true` (native MTP path)

## Observation

At long context (~30K tokens) with the prefix/boundary cache **fully hit** (no re-prefill), every request still re-primes the MTP drafter from scratch:

```
MTP path activated for uid=3 (model has mtp_forward, batch=1, primed=1595)
```

The prime replays the last ~1595 tokens **at decode speed** before generation starts, costing ~7–7.5 s per request at 30K context. Pure decode itself is healthy — the prime is the dominant fixed cost for interactive long-context sessions, and it is paid on **every turn** even when the prompt prefix is unchanged and the KV/boundary cache reuses everything.

## Measurements (30,267-token prompt, prefix cache warm)

| Request | Total | MTP decode phase (log: `timing[backbone+mtp]`) | Implied fixed overhead |
|---|---|---|---|
| max_tokens=30 (cold prefix) | 104.5 s | — | ~95 s prefill (318 tok/s) |
| max_tokens=500, run 1 | 15.8 s | 8.07 s (268 tok → 33.2 tok/s) | ~7.7 s |
| max_tokens=500, run 2 | 13.7 s | 6.97 s (238 tok → 34.0 tok/s) | ~6.7 s |

Decode at 30K ctx is only ~22 % below short-context (42–46 tok/s), which matches the expected KV-attention read cost — the pipeline itself is fine. The per-request prime is what makes long-context turns feel slow (server-reported `tok/s` also conflates it with decode, which initially misled our profiling).

## Suggestion

Snapshot the drafter state alongside the existing boundary-cache snapshots (or derive the prime from the cached prefix state), so that a request whose prefix is already cached can skip or shrink the re-prime. Even priming from the last boundary snapshot instead of a fixed 1595-token window would remove most of the fixed cost for multi-turn sessions.

Happy to provide full logs. Thanks!

beaglemoo:
Dug into this on merged main (`0436bdf2`) with corroborating measurements from a production M5 Pro running the same checkpoint — the symptom is real, but the mechanism is different from "re-primed at decode speed", and that changes where the fix belongs.

[...]

PR #3005, opened 2026-08-21
fix(mtp): scope the prime window to the folded span, not the absolute offset

beaglemoo:
Refs #2909 (analysis in issue comment).

## Problem

`OMLX_MTP_PRIME_WINDOW` exists to bound the head-KV memory cost of prompt priming, but its guard compares the **absolute prompt offset** (`prompt_priming.py`, `maybe_capture`). The head-KV grows with the **folded span**, and with a warm prefix cache only the boundary remainder is ever folded — so at long context, any usable window disables priming outright for exactly the requests where it is cheapest, cold-starting acceptance at the ~0.26 unprimed rate the module's own header documents (vs 0.90 primed).

## Change

Cap the folded span (`ctx.folded + seq_len`) instead of the absolute offset, and update `prime_window()`'s docstring to match its actual axis. An oversized remainder still runs unprimed, so the existing memory escape hatch is preserved — `test_window_cap_disables_long_prompts` passes unchanged.

## Verification

New regression test `test_window_caps_folded_span_not_absolute_offset`: an 8-token cached prefix (capture suppressed) followed by a 4-token remainder at absolute offset 12 with window 6 — the old guard dropped priming (12 > 6), the new guard folds the remainder (4 <= 6) and `prime_ctx_stats` reports the primed pairs. `tests/test_mtp_prompt_priming.py` + `tests/test_mlx_lm_mtp_patch.py`: 147 passed.

### DIFF
--- a/omlx/patches/mlx_lm_mtp/prompt_priming.py
+++ b/omlx/patches/mlx_lm_mtp/prompt_priming.py
@@ -83,12 +83,14 @@ def priming_enabled() -> bool:
 
 
 def prime_window() -> int:
-    """Max prompt length (tokens) to prime; 0 = unlimited.
+    """Max tokens to fold into one prime context; 0 = unlimited.
 
-    Escape hatch for the head-cache memory cost on very long prompts (one
-    full-attention layer of KV over the prompt). Prompts longer than the
-    window run unprimed — a coarse cap, not a suffix window, because the
-    prompt length is unknown while chunks stream through the model.
+    Escape hatch for the head-cache memory cost of priming (one
+    full-attention layer of KV over the folded span). The cap is measured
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for tracing the window check to the folded span. The change is correct for warm-cache requests with a small remainder, and the focused tests pass.

One edge case I’ll fold into a follow-up: after a multi-chunk remainder exceeds the window, `drop_ctx()` clears the only state, so a later small contiguous chunk can start priming again. This is limited to an explicitly configured nonzero `OMLX_MTP_PRIME_WINDOW` and does not affect the default path or output correctness.

LGTM for this scoped fix. I’ll handle the over-window latch and its regression test separately.

PR #3007, opened 2026-08-21
[fix] Preserve ANE tuning status across App View navigation

popfido:
## Summary

- move ANE tuning state and polling out of the transient model-settings view model into an app-scoped session
- keep an active run, its progress, and its last terminal result when leaving Model Settings, changing sidebar views, or closing and reopening App View
- scope the restored status to the tuned model and prevent another model from starting a duplicate run
- clear process-local tuning state when the managed server stops/fails, the server binding changes, or the app is rebound to another server
- preserve the latest upstream ANE recommendation fields, including fused-down and tail-padding settings

## Root cause

`ModelSettingsScreenVM` owns the tuner task and status today. That view model is recreated when users leave and reopen a model's settings, so the UI loses its tuning ID, polling task, progress, and final result even though the backend run continues.

## Validation

- rebased onto `upstream/main` at `2df39bfc`
- `xcodebuild -quiet -project apps/omlx-mac/oMLX.xcodeproj -scheme oMLX -configuration Debug -derivedDataPath /private/tmp/omlx-preserve-ane-tuning-status-derived-data-latest test`
  - 295 passed, 1 skipped, 0 failed
- manual Debug-app flow with `mlx-community/Qwen3.5-4B-MLX-4bit`:
  - started ANE tuning
  - changed to Status while the backend continued tuning
  - returned to the same model's Advanced settings
  - confirmed the last tuning status and GPU-only result remained visible instead of resetting to “Tune for this Mac”


popfido:
Regression-scope follow-up (`e5b56a08`): during the cross-functionality audit, I narrowed the app-scoped state to the ANE run identity, polling, progress, and results only. Tuner override controls remain screen-scoped, preserving their existing behavior and preventing Model A choices from leaking into Model B. Added regression coverage for that boundary.

Final validation:
- full macOS app suite: **296 passed, 1 skipped, 0 failed**
- focused Model Settings + ANE session tests under Thread Sanitizer: **20 passed, 0 failed**, with no race findings
- `git diff --check`: clean

PR #3012, opened 2026-08-21
feat(sampling): forward repetition_context_size so the penalty can reach loop units longer than 20 tokens

### LINKED ISSUE #3011: repetition_penalty is inert for verbatim loop units longer than mlx-lm's default 20-token window
We run a two-Mac TP2 cluster (M4 Max 48 GB each, macOS 26.5.2, oMLX HEAD-fa3e94b / 0.6.3rc2, mlx-lm pinned at ab1806e8, Qwen3.8-27B-oQ4e-mtp over JACCL) and spent this morning chasing a runaway generation: the model's reasoning started coherent, then collapsed into a fixed verbatim oscillation (~100 repetitions of the same indecision flip) and burned the entire 8192-token completion budget with `finish_reason=length` and no answer.

The embarrassing part: `repetition_penalty` was set to 1.1 and we confirmed it was reaching both ranks. It just couldn't do anything. mlx-lm's repetition penalty only looks back `repetition_context_size` tokens (`sample_utils.py`, `tokens[-context_size:]`), which defaults to **20** — and our loop unit was ~50 tokens. A token whose previous occurrence sits outside that window is never penalized, so for this failure mode the penalty value is irrelevant.

Reproduced deterministically against the pinned mlx-lm (`make_repetition_penalty(1.5, ...)`), logit of -2 on a token whose most recent occurrence is N tokens back:

| repeat distance | default window (20) | window 128 |
|---|---|---|
| 10 | -3.0 (penalized) | -3.0 |
| 20 | -3.0 | -3.0 |
| 25 | **-2.0 (untouched)** | -3.0 |
| 45 | **-2.0 (untouched)** | -3.0 |

On top of that, oMLX gives clients no way to widen the window even if they know about it: neither `ChatCompletionRequest` nor `CompletionRequest` carries `repetition_context_size`, so pydantic silently drops the key before it reaches any engine — while mlx-lm's own server already accepts it per request (`body.get("repetition_context_size", 20)`).

## Proposed fix

Forward the parameter end-to-end in oMLX (no mlx-lm changes needed):

- optional `repetition_context_size` (int > 0) on the chat/completion request models
- include it in distributed rank payloads only when set, so wire payloads stay unchanged otherwise
- carry it on `SamplingParams` through the batched/vlm engines into `make_logits_processors`

PR incoming.

xunlinkx:
We run a two-Mac TP2 cluster and spent this morning chasing a runaway generation: the model's reasoning started coherent, then collapsed into a fixed verbatim oscillation (~100 repetitions of the same indecision flip) and burned the entire 8192-token completion budget without ever emitting an answer.

The embarrassing part: `repetition_penalty` was set to 1.1 and we confirmed it was reaching both ranks. It just couldn't do anything. mlx-lm's repetition penalty only looks back `repetition_context_size` tokens, which defaults to **20** — and our loop unit was ~50 tokens. A token whose previous occurrence sits outside that window is never penalized, so for this failure mode the penalty value is irrelevant.

Reproduced deterministically against pinned mlx-lm (`make_repetition_penalty(1.5, ...)`), logit of -2 on a token whose most recent occurrence is N tokens back:

| repeat distance | default window (20) | window 128 |
|---|---|---|
| 10 | -3.0 (penalized) | -3.0 |
| 20 | -3.0 | -3.0 |
| 25 | **-2.0 (untouched)** | -3.0 |
| 45 | **-2.0 (untouched)** | -3.0 |

mlx-lm already accepts `repetition_context_size` everywhere it matters — `make_repetition_penalty()`, `make_logits_processors()`, and per-request on its own server (`body.get("repetition_context_size", 20)`). oMLX just never forwards it: nothing in the request models carries the field, so pydantic silently drops it before it reaches any engine. This PR is pure forwarding, no mlx-lm changes:

- `ChatCompletionRequest` / `CompletionRequest`: optional `repetition_context_size` (int > 0)
- distributed payload builders: include it only when set, so wire payloads are byte-identical otherwise and ranks keep mlx-lm's default
- `SamplingParams` + batched/vlm engines + both scheduler processor sites pass it through (None keeps the 20-token default)
- dflash needs nothing — it already forwards `**kwargs`

Two notes for reviewers: the key is omitted rather than sent as null so older/newer coordinator–rank mixes stay safe during rolling restarts in either direction; and I deliberately left `settings.json` defaults out of this one to keep it single-purpose — happy to follow up if that belongs somewhere.

Validation: new tests cover forwarding when requested, omission when not, and request validation (0 rejected). Full fast suite: 10232 passed, 108 skipped.

Diffstat:

```
 omlx/api/openai_models.py        |  2 ++
 omlx/engine/batched.py           |  2 ++
 omlx/engine/distributed.py       | 11 +++++++
 omlx/engine/vlm.py               |  2 ++
 omlx/request.py                  |  4 +++
 omlx/scheduler.py                |  8 +++++
 omlx/server.py                   | 38 ++++++++++++++++++++++
 tests/test_distributed_engine.py | 69 ++++++++++++++++++++++++++++++++
 tests/test_openai_models.py      | 28 +++++++++++++++
 9 files changed, 164 insertions(+)
```

Fixes #3011.


xunlinkx:
@copilot-pull-request-reviewer please review the latest commit, especially the scheduler default handling and request validation coverage.

jundot:
Thanks for removing the hardcoded fallback and carrying `repetition_context_size` through the main sampling paths. The request models, scheduler, batched engine, VLM, and distributed paths are consistent now.

DFlash still accepts this field without consuming it in the pinned runtime, but that engine did not support configurable repetition context before this change either. I’ll merge this and handle DFlash in a follow-up, preferably by adding runtime support or otherwise rejecting the unsupported option explicitly instead of silently ignoring it.

PR #3015, opened 2026-08-21
feat(cache): hot cache write-through — RAM-speed sessions with SSD durability

jonathan308:
## Summary

Adds an opt-in **hot cache write-through** mode: every KV block is retained in RAM **and** persisted to SSD at save time — ending the tradeoff between in-memory cache speed and cross-restart durability.

## Problem

The paged cache has forced a choice:

- **SSD caching**: sessions survive restarts, but resumes read blocks back from disk and the RAM tier goes underused.
- **Hot (in-memory) caching**: turns resume at RAM speed, but blocks never reach SSD — a server restart, `clear_hot_cache`, or hot-tier eviction silently destroys every cached conversation. The next resume pays a **full re-prefill** — at 200k+ tokens of context, that is tens of seconds to minutes, per session, per restart.

## What this does

- `PagedSSDCacheManager(hot_cache_write_through=True)`: at save time each block is retained in the hot tier **and** enqueued to the SSD writer. The writer marks the entry clean, so hot-tier eviction later drops it with no redundant rewrite.
- `clear_hot_cache()` now **flushes dirty entries to SSD** instead of dropping them — clearing RAM can no longer destroy un-persisted sessions (a durability fix that also benefits the existing hot-cache-only mode).
- Full plumbing, **off by default** (zero behavior change unless enabled):
  - `CacheSettings.hot_cache_write_through` (settings.json)
  - `OMLX_HOT_CACHE_WRITE_THROUGH` env var
  - `--hot-cache-write-through` CLI flag
  - Admin GET/PUT with runtime apply
  - Dashboard toggle in Cache settings, mirroring the existing "hot cache only" switch

## Live validation (M3 Ultra, DeepSeek-V4-Flash MXFP4, 2026-08-21)

- Fresh 20.9k-token chat → **full server restart** → resume:
  `paged cache hit, 20,480 tokens in 10 blocks, 428 tokens remaining, cache reconstructed` — only the new message re-prefilled.
- All 10 blocks rehydrated into the hot cache in **149.6 ms**.
- A 227k-token chat resumed post-restart: `227,328 tokens in 111 blocks, 2,536 remaining`.
- SSD index grew 824 → 940 blocks across the restart, zero scan errors.

## Tests

7 new tests in `TestHotCacheWriteThrough` covering write-through enqueue at save, clean-entry eviction without rewrite, dirty-entry flush on `clear_hot_cache`, and settings plumbing. Full cache test slice green (550 passed).


jundot:
Thanks, this closes a real gap, losing every session on a crash was the worst part of the write-back tradeoff and the restart validation numbers line up with the code paths. The settings plumbing is complete end to end including the pool template sync, appreciated. Merging as is, with two small follow-ups I'll push to main. First, I'm moving the dashboard toggle into the advanced settings cache group: I don't think users should have to configure every fine grained behavior, and this one is too narrow a setting to show up among the first order choices, it only matters for SSD plus hot cache setups. If you see it differently I'm happy to hear a counterproposal. Second, I'm adding a writer liveness guard to the clear_hot_cache flush path, mirroring what close() already does.

jonathan308:
@jundot For hot cache write through leaving in advanced settings is fine but I recommend it's left on then . Otherwise oMLx engine will seem much less convenient for new users since by default in memory hot cache is on . It is a huge lift for slow prefill machines.

PR #3028, opened 2026-08-22
Superseded: Cluster v2 requires paired devices

xunlinkx:
Superseded by Cluster v2.

The active wizard keeps discovered devices separate from paired devices. Unpaired Macs remain visible but are excluded from planning and peer probes.

Rebasing leaves no code change to carry; the current Cluster v2 regression covers this behavior.

### DIFF
--- a/omlx/admin/hf_downloader.py
+++ b/omlx/admin/hf_downloader.py
@@ -365,13 +365,17 @@ def _sum_safetensors_blob_bytes(siblings) -> int | None:
 def _fetch_safetensors_blob_bytes(api: HfApi, repo_id: str) -> int:
     """Look up *.safetensors blob bytes for one repo. 0 on failure."""
     try:
-        info = api.model_info(repo_id, files_metadata=True)
+        info = api.model_info(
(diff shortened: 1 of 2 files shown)


xunlinkx:
I rebased onto current main and addressed all three findings.

initializeClusterSetup now applies the live deployment's hosts over whatever selection was restored. The old guard let any non-empty cached selected_ssh win, so after a reload with a live deployment, the probe, hardware and budget calls all went to the stale host. I kept the probe gating on that path (skip when a probe exists or backoff is active) so poll re-runs do not hammer SSH.

loadClusterKnownNodes now restores selected_ssh only when the cache has explicit_selection: true. Old caches (the auto-promoted discoveries) and the lastGoodConfig fallback come back as suggestions with an empty selection. Saves write the marker, so a deliberate selection still survives a reload.

I replaced the source-string test with execution tests: discovery with no deployment and no selection leaves the selection empty and fires no probes, and a live deployment replaces a stale cached selection with the probe following the deployment host. 61 passed.

PR #3033, opened 2026-08-22
fix(cluster): do not report a failed transport probe as an Ethernet link

### LINKED ISSUE #3030: detect_transports reports Ethernet when the probe failed, refusing tensor parallelism on a 40 Gb/s link
### What happens

`detect_transports` in `omlx/cluster/transport.py` wraps both the Thunderbolt
and the RDMA probe in `except Exception: pass`, discards the reason, and then
falls through to a block that synthesises an Ethernet `TransportInfo` for every
host pair whenever `transports` is empty.

A probe that could not run is therefore indistinguishable from a probe that ran
and measured Ethernet.

### Why that matters

The two states are not equivalent to the caller. `transports_are_fast_enough`
reads an **empty** result as "link unknown" and declines to assume the worst,
but reads a **named Ethernet link** as a measured reason to refuse tensor
parallelism.

So the failure path produces a strictly worse outcome than the no-data path it
bypasses: a probe that crashed lands the planner in the one branch that refuses
TP, on evidence that was never gathered.

### Observed

Two Mac minis connected by a 40 Gb/s Thunderbolt 4 cable, `system_profiler`
confirming `receptacle_connected` / `peer_connected: true` on bus 0. SSH auth
failed for the probe, so the Thunderbolt branch raised before reading a single
interface. The cluster reported the link as **Ethernet** and refused tensor
parallelism on that basis.

### Acceptance criteria

- [ ] A probe that raises never contributes a synthesised Ethernet link
- [ ] A probe that completes and finds neither Thunderbolt nor RDMA still
      synthesises Ethernet, unchanged
- [ ] `select_backend` behaviour is unchanged (empty and all-Ethernet both
      yield `ring`)
- [ ] `describe_transports` still renders an empty result as
      "Link speed unknown"

### Tests required

- [ ] A failing Thunderbolt probe yields no transports rather than Ethernet
- [ ] A failing RDMA probe yields no transports rather than Ethernet
- [ ] Both probes completing with nothing found still yields Ethernet
- [ ] A successful Thunderbolt probe is unaffected

hellodk:
`detect_transports` caught every exception from the Thunderbolt and RDMA probes,
discarded the reason, and then synthesised an Ethernet link for every host pair.
A probe that could not run was indistinguishable from one that ran and measured
Ethernet.

That fabrication is worse than returning nothing. `transports_are_fast_enough`
reads an empty result as "link unknown" and declines to assume the worst, but
reads a named Ethernet link as a measured reason to refuse tensor parallelism.
The failure path therefore produced a worse outcome than the no-data path it
bypassed.

Observed on a two-Mac TB4 cluster where SSH auth failed for the probe: the link
was reported as Ethernet on a 40 Gb/s TB4 cable, and tensor parallelism was
silently refused on that basis.

Track whether a probe raised and synthesise Ethernet only when both probes
completed. `select_backend` is unaffected (empty and all-Ethernet both yield
`ring`) and `describe_transports` already renders empty as "Link speed unknown".

Closes #3030

### Tests

`tests/test_cluster_transport_probe_honesty.py`, new in this PR, covers the four
cases: a raising Thunderbolt probe, a raising RDMA probe, both probes completing
with nothing found, and an unaffected successful probe.

### Verification

Run on an Apple M4 Mac mini, macOS 26.5.1, mlx 0.32.0, pytest 9.1.1.

```
tests/test_cluster_transport_probe_honesty.py
tests/test_cluster_link_status.py
tests/test_cluster_link_bandwidth.py
    59 passed in 0.13s
```

Full suite on the branch this was cut from: 10321 passed, 2 failed, 37 errors,
164 skipped. All 39 failures and errors are `openai_harmony.HarmonyError: error
downloading or loading vocab file` in `test_harmony.py`, `test_harmony_parser.py`
and `test_output_parser.py` — an environment problem on the test machine. A
control run of those three files at the commit preceding this work reproduces
exactly the same 2 failures and 37 errors, so they are pre-existing and
unrelated to this change.

### DIFF
--- a/omlx/cluster/transport.py
+++ b/omlx/cluster/transport.py
@@ -267,6 +267,9 @@ def detect_transports(
 
     transports: list[TransportInfo] = []
     physical_edges: set[tuple[int, int]] = set()
+    # A probe that raised measured nothing. Tracked separately from an empty
+    # result so a host we could not reach is never described as a slow link.
+    probe_failed = False
 
     # Try Thunderbolt connectivity
     try:
@@ -300,8 +303,10 @@ def detect_transports(
                             )
                         )
     except Exception:
-        # Thunderbolt detection failed — degrade to Ethernet
-        pass
+        # Thunderbolt detection failed. Whether this link is Ethernet is now
+        # unknown, not established: SSH that cannot authenticate raises here
+        # long before any interface is read.
+        probe_failed = True
 
     # Check RDMA
     try:
@@ -336,10 +341,15 @@ def detect_transports(
                 )
     except Exception:
         # RDMA detection failed — skip
-        pass
+        probe_failed = True
 
-    # If no transports detected, assume Ethernet
-    if not transports:
+    # A completed probe that found neither Thunderbolt nor RDMA has established
+    # an Ethernet link, and says so. A probe that could not run has established
+    # nothing, and must stay silent: callers read an empty result as "link
+    # unknown" and decline to assume the worst, but read a named Ethernet link
+    # as a measured reason to refuse tensor parallelism. Claiming the
+    # measurement we failed to take is the more damaging of the two.
+    if not transports and not probe_failed:
         for i, _host in enumerate(hosts):
             for j, peer_host in enumerate(hosts):
                 if i == j:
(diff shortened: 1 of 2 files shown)


hellodk:
Duplicate of #3023, which carries the identical diff and was opened first. Closing this one; #3023 is the PR to review.

PR #3034, opened 2026-08-22
fix(cluster): name the link, not the head counts, when the link gate blocks TP

### LINKED ISSUE #3031: choose_parallelism blames architecture divisors when the link gate dropped every candidate
### What happens

`choose_parallelism` in `omlx/cluster/autoconfigure.py` can exit its candidate
loop having planned nothing and recorded no error, then fall through to a final
`PlanningError` that names `model.tensor_parallel_divisors` — dimensions that
may divide the node count perfectly well.

### The path

With `strategy="auto"`, more than one node, and `supports_pipeline` false, the
candidate list is narrowed to tensor-parallel degrees only, dropping the `tp=1`
fallback. A candidate the link gate rejects then hits `continue` **without**
setting `last_error`. The loop ends with nothing planned and nothing recorded,
so the divisor message fires by default.

### Observed

    no tensor-parallel degree divides 2 nodes across architecture dimensions (16, 8)

on a model whose head counts divide two nodes exactly. The message sends the
operator to inspect an architecture that was never the problem, while the real
cause — an unusable link reading — goes unmentioned.

### Acceptance criteria

- [ ] When every candidate was dropped by the link gate, the refusal names the
      link and the detected transport, not the architecture dimensions
- [ ] The refusal states the ways forward (choose the Tensor strategy
      explicitly, or check the connection)
- [ ] The divisor message is unchanged for genuine divisibility failures
- [ ] No change to any path that successfully plans

### Tests required

- [ ] Link gate drops every degree -> error names the link, not the divisors
- [ ] Genuine divisibility failure -> divisor message unchanged
- [ ] A plannable configuration is unaffected

hellodk:
`choose_parallelism` could exit its candidate loop having planned nothing and
recorded no error, then fall through to a message naming
`tensor_parallel_divisors` -- dimensions that may divide the node count
perfectly well.

The path: with `strategy="auto"`, more than one node and `supports_pipeline`
False, the candidate list is narrowed to tensor-parallel degrees only, dropping
the `tp=1` fallback. A candidate the link gate rejects then hits `continue`
without setting `last_error`. The loop ends with nothing planned and nothing
recorded, so the divisor message fires.

In practice this printed

    no tensor-parallel degree divides 2 nodes across architecture dimensions (16, 8)

for a model whose heads divide two nodes exactly, sending the user to inspect an
architecture that was never the problem while the real cause -- an unusable link
reading -- went unmentioned.

Record degrees dropped by the link gate and raise a refusal naming the link, the
detected transport and the ways forward. The divisor message is unchanged for
genuine divisibility failures.

Closes #3031

### Tests

`test_automatic_never_falls_back_to_an_unsupported_pipeline` asserted the old
wording for a model with 8 heads on 2 nodes, where the heads also divide
exactly; it pinned the incorrect message. Its intent is preserved and its
assertion strengthened. Further cases in `tests/test_cluster_autoconfigure.py`
cover the link-gate refusal and leave the divisor message pinned for genuine
divisibility failures.

### Verification

Run on an Apple M4 Mac mini, macOS 26.5.1, mlx 0.32.0, pytest 9.1.1.

```
tests/test_cluster_autoconfigure.py
    89 passed in 2.44s
```

Full suite on the branch this was cut from: 10321 passed, 2 failed, 37 errors,
164 skipped. All 39 failures and errors are `openai_harmony.HarmonyError: error
downloading or loading vocab file` in `test_harmony.py`, `test_harmony_parser.py`
and `test_output_parser.py` — an environment problem on the test machine. A
control run of those three files at the commit preceding this work reproduces
exactly the same 2 failures and 37 errors, so they are pre-existing and
unrelated to this change.

### DIFF
--- a/omlx/cluster/autoconfigure.py
+++ b/omlx/cluster/autoconfigure.py
@@ -248,6 +248,9 @@ def choose_parallelism(
 
     warnings: list[str] = []
     last_error: PlanningError | None = None
+    # Degrees the link gate rejected before planning was ever attempted. Kept
+    # so the refusal can name the link instead of the head counts.
+    skipped_for_link: list[int] = []
 
     for tensor_parallel_size in candidates:
         fast_enough = transports_are_fast_enough(
@@ -262,6 +265,7 @@ def choose_parallelism(
             and not fast_enough
             and not measured_on_this_path
         ):
+            skipped_for_link.append(tensor_parallel_size)
             continue
         try:
             plan = plan_hybrid(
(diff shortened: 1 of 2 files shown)


hellodk:
Duplicate of #3024, which carries the identical diff and was opened first. Closing this one; #3024 is the PR to review.

Issue #3039, opened 2026-08-22
distributed TP ranks deadlock in Metal fence wait under overlapping requests while every liveness check stays green


xunlinkx:
## Summary

Two-rank tensor-parallel deployment over JACCL/Thunderbolt RDMA. Two overlapping streaming requests froze both ranks mid-prefill inside a collective. Both rank processes then spun at ~100% CPU forever. Nothing reported it:

- runtime markers stayed fresh ("ready") — the heartbeat runs on a thread independent of generation
- PeerWatchdog saw reachable SSH and live pids
- the coordinator errored the client after its 300 s read timeout but never cancelled the rank-side work
- teardown was refused with 409 (`ModelBusyError`) until every client socket closed
- the ranks ignored SIGTERM (main thread stuck in native code) and needed SIGKILL, which left stale marker debris

This happened twice on 2026-08-22.

## Environment

- 2× Apple M4 Max 48 GB, macOS 26.5.2
- oMLX 0.6.3rc2 (HEAD-fa3e94b), MLX 0.32.0
- Model `scottlowry/Qwen3.8-27B-oQ4e-mtp`, TP=2, 64 layers, backend jaccl (Thunderbolt 5 RDMA, iface rdma_en3)
- Execution profile balanced, `cache_affinity` and `prompt_cache_ssd` enabled, `sampling_rank_only` on (pure TP keeps the synchronized sampler)

## What happens

Timeline from the fully instrumented occurrence (coordinator log times):

1. `07:31:56` — completion logged for a 51,724-token request (`finish_reason=tool_calls`). In the same second a new private POST reaches the rank-zero server.
2. `07:33:46` — a second private POST arrives while the first is still in flight.
3. ~07:36 — both ranks stop stepping. Rank telemetry freezes identically on both sides: `batch_steps=11598`, `active_requests=2`, prefill stuck at 14,485/16,920 tokens.
4. `sample` of both rank processes shows the MLX scheduler thread blocked here:

   ```
   mlx::core::scheduler::StreamThread::thread_fn()
     → std::__function::__func<mlx::core::Fence::wait(...)::$_1, void ()>::operator()()
   ```

   and the `com.Metal.CommandQueueDispatch` thread blocked in the same fence. Both processes burn one full core each (Metal/JACCL spin-polling).
5. `07:41:51` — coordinator logs `rank-zero inference stream timed out: no rank-zero data for 300s while the cluster was ready`. The client gets an error; nothing on the rank side is cancelled or restarted.
[...]

PR #3040, opened 2026-08-22
fix(cluster): exit rank when pipeline stalls mid-request

xunlinkx:
Refs #3039.

A stalled collective can leave all ranks reporting healthy while generation remains blocked. This change adds a watchdog that detects an unchanged progress counter while requests are in flight, records the stall, emits an event, and exits the rank so the supervisor can recover it.

Idle ranks do not trigger the watchdog, and unreadable telemetry fails open for that interval. The default timeout is 180 seconds and can be changed with `OMLX_PROGRESS_STALL_TIMEOUT`.

Validation: `pytest tests/test_cluster_inference_worker.py tests/test_cluster_liveness.py` — 83 passed.

### DIFF
--- a/omlx/cluster/inference_worker.py
+++ b/omlx/cluster/inference_worker.py
@@ -858,6 +858,148 @@ def _start_launcher_watchdog(
     thread.start()
 
 
+_DEFAULT_PROGRESS_STALL_TIMEOUT = 180.0
+
+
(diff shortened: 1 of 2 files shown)


xunlinkx:
Rebased onto current main. The watchdog now treats a frozen `batch_steps` counter as a stall only when requests are in flight; idle generators are allowed to remain parked. The docstring and regression tests were updated to match that behavior.

PR #3042, opened 2026-08-22
fix(hy_v3): normalize legacy RoPE config across oQ calibration paths

### LINKED ISSUE #3041: hy_v3 pre-load patch fails to normalize rope_theta to rope_parameters during oQe calibration
Model: tencent/Hy-MT2-30B-A3B
Quantization: oQ6 / oQe

Hy3 pre-load patch is successfully applied:

Hy3 pre-load patch applied for ...

but model loading still fails during imatrix collection:

ModelArgs.__init__() missing 1 required positional argument: 'rope_parameters'

The original Hy-MT2 config provides `rope_theta`, while oMLX's vendored
hy_v3 ModelArgs currently requires `rope_parameters`.

This causes model loading to fail before imatrix collection, followed by:

oQe imatrix collection produced no entries

Related to #1360, but this appears to be a separate config-schema compatibility
issue after hy_v3 support was added.

chenqianhe:
## Summary

Fix Hy-MT2 checkpoints failing during oQe calibration because their config uses the legacy root-level `rope_theta` field while the vendored `hy_v3.ModelArgs` requires structured `rope_parameters`.

Fixes #3041.

## Problem

Tencent's `Hy-MT2-30B-A3B` publishes:

```json
"rope_theta": 11158840.0
```

but does not provide:

```json
"rope_parameters": {
  "rope_theta": 11158840.0,
  "rope_type": "default"
}
```

As a result, some Hy3 model construction paths failed with:

```text
ModelArgs.__init__() missing 1 required positional argument: 'rope_parameters'
```

During oQe this prevented calibration model loading, which then surfaced as:

```text
oQe imatrix collection produced no entries
```

The shared `mlx_lm.utils.load_config` path was not the only affected path: oQ sanitizer and proxy discovery also construct `ModelArgs` directly from the in-memory source config.

## Changes

- Add a fill-only Hy3 config normalizer that maps legacy `rope_theta` to `rope_parameters`.
- Apply it from the shared `mlx_lm.utils.load_config` wrapper.
- Normalize the direct oQ sanitizer/proxy config path before `ModelArgs.from_dict`.
- Register the vendored Hy3 implementation before oQ sanitizer discovery.
- Preserve existing structured `rope_parameters` without modification.
- Leave invalid configs missing both representations unchanged so they still fail explicitly.
- Add regression coverage for legacy, null, structured, unrelated, and direct oQ sanitizer paths.

The source checkpoint is never modified.

## Validation

Unit and adjacent regression tests:

```text
58 passed
```

End-to-end smoke-tested with the local original checkpoint:

```text
tencent/Hy-MT2-30B-A3B
```

The test exercised:

- oQ6 streaming quantization
- RAM-safe 4-bit calibration proxy creation
- oQe imatrix collection
- Hy3 streaming sanitize-plan discovery
- complete temporary oQ6 output generation
- strict lazy loading of the quantized output
- a single-token forward pass

Results:

```text
Calibration proxy:        15.8 GB
Sanitized output tensors: 766
Imatrix entries:          524
Final temporary output:   24.87 GB
Forward logits shape:     (1, 1, 120832)
Hy3 sanitizer warnings:   0
```

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -3891,6 +3891,21 @@ def _build_model_sanitizer(config: dict, text_only: bool = False):
             except Exception as patch_err:
                 logger.debug(f"laguna patch not applied: {patch_err}")
 
+        # Hy3 is vendored into ``sys.modules`` like Laguna, but its published
+        # Hy-MT2 checkpoints also use the legacy root-level ``rope_theta``
+        # schema. Sanitizer/proxy discovery consumes this in-memory config
+        # directly (without ``mlx_lm.utils.load_config``), so register the
+        # model and normalize the schema before ``ModelArgs.from_dict``.
+        if config.get("model_type") == "hy_v3":
+            try:
+                from omlx.patches.hy_v3 import apply_hy_v3_patch
+                from omlx.utils.model_loading import normalize_hy_v3_rope_config
+
+                normalize_hy_v3_rope_config(config)
+                apply_hy_v3_patch()
+            except Exception as patch_err:
+                logger.debug(f"hy_v3 patch not applied: {patch_err}")
+
         if config.get("model_type") == "mimo_v2":
             try:
                 from omlx.patches.mimo_v2 import apply_mimo_v2_patch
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for covering both the shared `mlx-lm` `load_config` path and the direct oQ sanitizer path. The fill-only normalization preserves checkpoint-provided `rope_parameters` while correctly translating root-level `rope_theta` for Hy-MT2, without masking malformed or genuinely missing values.

LGTM, merging.

PR #3043, opened 2026-08-22
fix: clamp advertised context window to unified-memory headroom

kenhuangus:
## Problem

`GET /v1/models` reports the model's theoretical context (`max_model_len`, often 262144) even on 64 GB Apple Silicon. Agent clients (OpenClaw, Cline, Hermes) treat that as the compaction denominator.

On a Mac Mini M4 Pro 64 GB running Qwen3.8-27B-4bit:

- OpenClaw advertised `contextWindow: 262144`
- A WeChat tool-loop grew the prompt from ~3k to **76,149** tokens (`323/375` completions ended `finish_reason=tool_calls`)
- oMLX never 400'd (`validate_context_window` compared against 256k)
- The scheduler throttled prefills (`chunk 2048 -> 1280`) instead, heating the chassis
- Compaction count on the session stayed **0**

This matches the "something froze" report: GPU-bound 70k+ prefills, plus an earlier `watchdog timeout: no checkins from watchdogd in 91 seconds` panic.

## Change

When no per-model `max_context_window` override and no `sampling.max_context_window_policy` are set, clamp the discovered native window by unified RAM:

| RAM | advertised cap |
|---|---|
| ≤36 GB | 32,768 |
| ≤80 GB | 65,536 |
| ≤160 GB | 131,072 |
| more | native (unchanged) |

Live reading on the 64 GB Mini: `ram_context_cap_tokens() == 65536`.

Escape hatches (unchanged priority):

1. Per-model `max_context_window` still wins
2. `sampling.max_context_window_policy` still wins over the RAM cap

## Tests

- `tests/test_context_window.py`: RAM buckets + clamp of native 256k → 64k
- `tests/test_server.py::TestGetMaxContextWindow`: existing #1308 cases keep RAM cap patched off; new case asserts the 64 GB clamp

I could not run the full pytest suite on Windows (no `mlx`). The RAM helper was executed on the Mini and returned 65536.

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -1742,6 +1742,36 @@ def _get_ocr_defaults(model_id: str | None) -> dict | None:
     return None
 
 
+def ram_context_cap_tokens(mem_bytes: int | None = None) -> int | None:
+    """Return a RAM-aware context ceiling, or ``None`` to leave native as-is.
+
+    Local Apple Silicon servers advertise the model's theoretical window
+    (often 256k) even when unified memory cannot hold that KV cache.
+    Agent clients then skip compaction and the scheduler crawls through
+    multi-minute throttled prefills until the chassis overheats.
+
+    Per-model ``max_context_window`` overrides still win. An explicit
+    ``sampling.max_context_window_policy`` still wins over this auto cap.
+    """
+    if mem_bytes is None:
+        try:
+            mem_bytes = int(os.sysconf("SC_PHYS_PAGES")) * int(
+                os.sysconf("SC_PAGE_SIZE")
+            )
+        except (ValueError, OSError, AttributeError):
+            return None
+    if mem_bytes <= 0:
+        return None
+    gb = mem_bytes / (1024**3)
+    if gb <= 36:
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for the report and the patch.

The clamp does not only affect what gets advertised. `get_max_context_window()` feeds `validate_context_window()`, which the completions, chat, messages and responses handlers all call, so requests that work today start coming back as 400 on any machine at or under 160GB.

The buckets are also keyed on RAM alone. A DSv4 style MLA model or a GDN hybrid costs very little per token of context, so a 128GB machine that genuinely serves 256k gets cut to 131,072, and a small model with a 262k window on a 32GB machine drops to 32,768. The docstring right above the function says the policy field is unset by default so no existing install shifts, and this changes that for everything except the very large configurations.

For the workload you described, agents that never compact, `max_context_window_policy` already covers it and it is opt-in. Could you either put the RAM cap behind that same kind of explicit setting, or derive the cap from the loaded model's KV cost per token instead of the physical RAM bucket?

Issue #3054, opened 2026-08-23
perf(hy_v3): enable routed MoE gate/up fusion for Hy-MT2 decode


chenqianhe:
## Summary

Hy-MT2 / HyV3 models currently miss the routed-expert gate/up fusion already used by Qwen3.5/3.6 and Laguna models.

HyV3 uses the same `mlx_lm.models.switch_layers.SwitchGLU` implementation, so the existing post-load fusion can be applied without changing checkpoint weights, quantization, routing behavior, or the generation API.

A local prototype improves single-token decode throughput by 5.3%–8.3% on a 30B-A3B Hy-MT2 model while preserving the complete greedy token sequence.

## Background

The model used for validation was a Balanced MLX conversion of:

`Hy-MT2-30B-A3B-APEX-Imatrix-I-Balanced`

The converted model has:

- 48 decoder layers
- 1 dense layer followed by 47 MoE layers
- 128 routed experts per MoE layer
- top-8 expert activation
- 1 shared expert per MoE layer
- affine quantization with group size 64
- 90 5-bit quantized modules
- 272 6-bit quantized modules
- 117 8-bit quantized modules
- 47 FP32 expert/router bias tensors retained for correctness

The conversion itself has already been validated:

- 7/8 evaluation prompts produced text identical to the source GGUF
- 7/8 produced identical token sequences
- the remaining case was a semantically equivalent Spanish translation
- no unintended F16/F32 mixing exists in normal compute weights
- the quantization bit-width mapping matches the source GGUF layout

In the original cross-runtime measurements:

- llama.cpp GGUF: approximately 101.2 tok/s
- MLX hot decode: approximately 91.3 tok/s

These numbers do not use exactly the same timing boundary, so they should not be interpreted as a pure kernel comparison. However, they motivated profiling the MLX decode path.

## Problem

HyV3 uses the stock `SwitchGLU` routed-expert implementation. Each MoE layer currently performs three separate expert projection calls:

1. gate projection
2. up projection
3. down projection

For a 47-layer MoE model, this results in approximately:

`47 × 3 = 141 gather_qmm launches per decoded token`

The existing `qwen35_moe_gate_up` post-load patch concatenates gate and up weights along the expert output dimension and replaces the two projections with one fused projection followed by a split.

This reduces the routed-expert path to:

`47 × 2 = 94 gather_qmm launches per decoded token`

However, the patch is currently gated by the top-level model module name:

[...]

PR #3055, opened 2026-08-23
feat(hy_v3): fuse routed MoE gate and up projections

chenqianhe:
## Summary

This PR enables the existing routed MoE gate/up fusion for HyV3 models.

HyV3 uses the same stock `mlx_lm.models.switch_layers.SwitchGLU` implementation as the model families already supported by this post-load optimization. Adding `hy_v3` to the supported family tokens allows Hy-MT2 models to reduce routed-expert projection launches from three to two per MoE layer.

On a 48-layer Hy-MT2-30B-A3B model with 47 MoE layers, this removes 47 `gather_qmm` launches per decoded token.

## Context

The existing fusion performs the following post-load transform:

Before:

- gate `gather_qmm`
- up `gather_qmm`
- down `gather_qmm`

After:

- fused gate/up `gather_qmm`
- split gate/up output
- down `gather_qmm`

The implementation concatenates packed gate and up weights along the output dimension. It also concatenates their scales and biases without dequantizing or requantizing the checkpoint.

HyV3 was previously excluded only because its model module name, `mlx_lm.models.hy_v3`, was not present in the family allowlist.

## Changes

- add `hy_v3` to the supported MoE gate/up fusion families
- update the patch description and runtime logging to refer to supported MoE models rather than only Qwen
- update LLM and VLM engine comments/logging to reflect the shared optimization
- add HyV3 bit-exact regression coverage for:
  - affine 5-bit
  - affine 6-bit
  - affine 8-bit
  - group size 64

The existing function and module names are intentionally retained to keep the change focused and avoid unnecessary import or compatibility churn.

## Safety

The existing compatibility checks remain unchanged.

A module is fused only when:

- it is the stock `SwitchGLU` type
- gate and up use the same module type
- quantized gate and up projections have matching:
  - group size
  - bit width
  - quantization mode
  - bias presence
- gate and up weights have identical shapes and dtypes

Unsupported or incompatible modules continue to use the original implementation.

The transform does not modify:

- checkpoint files
- quantization codes
- per-row scales or biases
- router computation
- expert bias precision
- top-k selection
- expert weighted-sum order
- shared-expert execution

## Correctness validation

The unit test constructs a HyV3-family model using the stock `SwitchGLU`, applies the same affine formats used by the validated Balanced checkpoint, and compares the complete projection output before and after fusion using `mx.array_equal`.

Covered configurations:

| Quantization | Group size | Result |
|---|---:|---|
| affine 5-bit | 64 | bit-exact |
| affine 6-bit | 64 | bit-exact |
| affine 8-bit | 64 | bit-exact |

A full-model greedy decode comparison was also performed:

- 256 generated tokens: complete sequence identical
- 1024 generated tokens: complete sequence identical

## Performance validation

Model:

`Hy-MT2-30B-A3B-APEX-Imatrix-I-Balanced-MLX`

Relevant model structure:

- 48 decoder layers
- 47 MoE layers
- 128 routed experts
- top-8 routing
- 1 shared expert per MoE layer
- group size 64
- mixed affine 5/6/8-bit routed projections

Benchmark methodology:

- Apple Silicon / Metal
- baseline and fused phases used the same loaded model
- fusion applied in place after the baseline phase
- 12-token prompt
- greedy decoding
- forced output length to avoid EOS-dependent short samples
- 32-token warm-up before each phase
- 5 runs per output length
- median throughput reported
- `mx.synchronize()` before and after timing
- SpecDraft disabled
- prefix cache disabled

| Decode length | Baseline trials (tok/s) | Fused trials (tok/s) |
|---|---|---|
| 256 | 86.97, 88.13, 89.03, 87.37, 88.94 | 92.76, 92.79, 92.76, 93.08, 91.73 |
| 1024 | 89.38, 87.20, 86.80, 86.88, 87.88 | 92.37, 94.46, 94.43, 95.33, 94.61 |

Median results:

| Decode length | Baseline median | Fused median | Improvement |
|---|---:|---:|---:|
| 256 tokens | 88.13 tok/s | 92.76 tok/s | +5.26% |
| 1024 tokens | 87.20 tok/s | 94.46 tok/s | +8.33% |

Peak-memory medians:

| Decode length | Baseline | Fused | Difference |
|---|---:|---:|---:|
| 256 tokens | 22.322 GB | 22.329 GB | +0.007 GB |
| 1024 tokens | 22.416 GB | 22.430 GB | +0.015 GB |


chenqianhe:
#3054

chenqianhe:
@jundot Hi, Could you please rerun the failed Python 3.12 job?

The same commit passed on Python 3.11 and 3.13, and the failing test passes when rerun locally. This appears to be a known race in MLX's distributed launcher where a rank's final stdout record may not be drained after the child exits, while the launcher still returns exit code 0. The PR does not modify any cluster-related code.

chenqianhe:
@jundot Hi! I’ve merged the latest main into this PR branch and resolved the conflicts while preserving the newly added Qwen4-Exp support.
When you have time, could you please take another look and let me know whether this is ready to merge or if you’d like any further adjustments? Thanks!

jundot:
Thanks for the focused update. I verified that HyV3 resolves to the stock `mlx_lm.models.switch_layers.SwitchGLU`, and the gate/up fusion remains bit-exact across the covered affine formats and an actual HyV3 model instance. The Python 3.12 failure is confined to the unrelated cluster pipeline smoke; it passes locally, while 3.11 and 3.13 pass. LGTM, merging.

PR #3057, opened 2026-08-23
fix(opencode): isolate launch config to preserve user settings

amit-batra:
# PR Description

## Summary

`omlx launch opencode` currently writes the selected oMLX provider/model configuration directly into the user's persistent:

```text
~/.config/opencode/opencode.json
```

The existing implementation creates a timestamped backup before modifying the file, but the original configuration is not restored when the OpenCode session exits.

This leaves the user's persistent OpenCode configuration modified and causes timestamped `.bak` files to accumulate.

More importantly, simply adding backup restoration would not safely solve the problem for concurrent OpenCode sessions. Multiple `omlx launch opencode` processes can operate on the same global configuration file, creating races where one session can overwrite or restore another session's configuration.

This change makes the OpenCode launch configuration **session-scoped** instead.

## What changed

- `omlx launch opencode` now creates a unique temporary configuration for each launch.
- If the user already has `~/.config/opencode/opencode.json`, it is copied into the temporary configuration so existing OpenCode settings are preserved.
- The oMLX provider and selected model are applied only to that temporary configuration.
- OpenCode is launched with `OPENCODE_CONFIG` pointing to the session-specific configuration.
- The temporary configuration directory is removed after OpenCode exits.
- OpenCode's exit status is propagated back to oMLX.
- The persistent user configuration is never modified.
- The existing generic JSON configuration-writing functionality remains available for integrations that intentionally modify persistent configuration files.

## Why this approach

A backup-and-restore solution appears simple, but it is fundamentally problematic for concurrent sessions.

For example:

```text
Session A:
    save original config A
    write temporary config A
    launch OpenCode

Session B:
    sees temporary config A
    saves it as its "original"
    writes temporary config B
    launch OpenCode

Session A exits:
    restores config A

Session B exits:
    restores config B   <-- potentially leaves another session's config behind
```

Serializing all OpenCode launches could avoid this race, but would prevent users from running independent OpenCode sessions concurrently.

Instead, each session gets its own configuration:

```text
OpenCode session A
    -> /tmp/omlx-opencode-XXXX/opencode.json

OpenCode session B
    -> /tmp/omlx-opencode-YYYY/opencode.json

OpenCode session C
    -> /tmp/omlx-opencode-ZZZZ/opencode.json
```

while:

```text
~/.config/opencode/opencode.json
```

is never modified.

This makes concurrent sessions naturally isolated.

## User-visible behavior

Before:

```text
omlx launch opencode
        |
        +--> backup ~/.config/opencode/opencode.json
        |
        +--> modify ~/.config/opencode/opencode.json
        |
        +--> launch OpenCode
        |
        +--> OpenCode exits
        |
        `--> modified config remains
```

After:

```text
omlx launch opencode
        |
        +--> copy user's config to temporary location
        |
        +--> apply oMLX configuration to temporary config
        |
        +--> launch OpenCode using OPENCODE_CONFIG
        |
        +--> OpenCode exits
        |
        `--> temporary config is removed
```

The user's original configuration is therefore preserved automatically.

## Concurrency

The implementation has been designed specifically to support multiple concurrent launches.

For example:

```text
Terminal 1                         Terminal 2

omlx launch opencode               omlx launch opencode
        |                                  |
        v                                  v
temporary config A                 temporary config B
        |                                  |
        v                                  v
OpenCode A                         OpenCode B
        |                                  |
        v                                  v
cleanup A                          cleanup B
```

There is no shared mutable OpenCode configuration between these sessions.

## Tests

Regression tests were added covering:

1. Preservation of the user's persistent OpenCode configuration.
2. Creation of a session-specific configuration.
3. Application of the selected oMLX provider/model to the temporary configuration.
4. Use of `OPENCODE_CONFIG` when launching OpenCode.
5. Cleanup of the temporary configuration after the OpenCode process exits.
6. Propagation of the OpenCode process exit status.
7. Isolation of concurrent OpenCode sessions.

## Scope

This PR intentionally focuses on the OpenCode integration.

It does not change OpenCode itself and does not alter the behavior of other oMLX integrations.

## Backward compatibility

The user-facing command remains unchanged:

```bash
omlx launch opencode
```

No new command or configuration is required.

Existing OpenCode configuration continues to be used as the starting point for each session.

## Related problem

The issue being addressed is that the OpenCode integration currently treats the user's persistent configuration as mutable launch-time state.

The integration should instead treat the oMLX-specific OpenCode configuration as temporary, session-scoped state.

---

# Implementation Notes for Reviewers

The current OpenCode integration defines the persistent configuration path as:

```python
CONFIG_PATH = Path.home() / ".config" / "opencode" / "opencode.json"
```

and previously called the generic configuration writer against that path.

The new implementation separates:

- configuration generation for a specified path, and
- the persistent/global configuration path.

For an OpenCode launch, the implementation:

1. Creates a temporary directory with `tempfile.TemporaryDirectory()`.
2. Copies the existing OpenCode configuration into that directory when present.
3. Applies the oMLX provider/model configuration without creating a backup.
4. Sets:

```text
OPENCODE_CONFIG=<temporary-config-path>
```

5. Launches OpenCode as a child process.
6. Removes the temporary directory when the child exits.
7. Returns the child's exit status.

This avoids relying on process replacement via `os.execvpe()` because the parent process needs to retain control long enough to clean up the temporary configuration.

---

# Suggested Testing Command

Run the OpenCode integration tests:

```bash
pytest tests/test_opencode_integration.py
```

Run the broader test suite if appropriate for the development environment:

```bash
pytest
```

---

# Reviewer Checklist

- [ ] Existing `~/.config/opencode/opencode.json` is never modified by `omlx launch opencode`.
- [ ] Existing OpenCode configuration is preserved.
- [ ] oMLX provider/model configuration is correctly applied.
- [ ] `OPENCODE_CONFIG` points to the temporary configuration.
- [ ] Temporary files are cleaned up after OpenCode exits.
- [ ] OpenCode's exit status is preserved.
- [ ] Concurrent OpenCode sessions do not share configuration state.
- [ ] Other oMLX integrations are unaffected.

### DIFF
--- a/omlx/integrations/base.py
+++ b/omlx/integrations/base.py
@@ -51,11 +51,11 @@ class IntegrationContext:
 class Integration:
     """Base integration definition."""
 
-    name: str
-    display_name: str
-    type: str
-    install_check: str
-    install_hint: str
+    name: str # "codex", "opencode", "openclaw", "hermes", "pi"
+    display_name: str # "Codex", "OpenCode", "OpenClaw", "Hermes Agent", "Pi"
+    type: str # "env_var" or "config_file"
+    install_check: str # binary name to check with `which`
+    install_hint: str # installation instructions
 
     def get_command(self, ctx: IntegrationContext) -> str:
         """Generate the command string for clipboard/display."""
@@ -78,7 +78,12 @@ class Integration:
         return None
 
     def _scrubbed_env(self) -> dict[str, str]:
-        """Return an os.environ copy with bundled-Python vars removed."""
+        """Return an os.environ copy with bundled-Python vars removed.
+
+        oMLX.app sets PYTHONHOME/PYTHONPATH to its bundled cpython-3.11.
+        Launched tools spawn their own Python subprocesses; if they inherit
+        these they crash with init_fs_encoding errors.
+        """
         env = os.environ.copy()
         for key in ("PYTHONHOME", "PYTHONPATH", "PYTHONDONTWRITEBYTECODE"):
             env.pop(key, None)
@@ -87,7 +92,14 @@ class Integration:
     def select_model(
         self, models_info: list[dict], tool_name: str | None = None
     ) -> str:
-        """Select a model interactively."""
+        """Select a model interactively.
+
+        Shows a curses arrow-key picker when running in a TTY; falls back to
+        numbered terminal selection when curses is unavailable (e.g. native
+        Windows Python) or stdout is not a TTY.
+
+        Returns the selected model id (empty string when models_info is empty).
+        """
         if not models_info:
             return ""
 
@@ -113,10 +125,14 @@ class Integration:
             try:
                 return _select_model_curses(models_info, name)
             except ImportError:
+                # Stdlib curses missing (e.g. native windows python).
                 pass
             except Exception:
+                # Curses init/runtime failure (dumb terminal, no terminfo
+                # entry, broken pipe, etc.). Fall through to numbered.
                 pass
 
+        # Fallback: numbered terminal selection
         print("Available models:")
         for i, m in enumerate(models_info, 1):
             ctx = m.get("max_context_window")
@@ -145,7 +161,7 @@ class Integration:
         *,
         backup: bool = True,
     ) -> None:
-        """Read, update, and write a JSON config file.
+        """Read, update, and write a JSON config file with optional backup.
 
         Args:
             config_path: Path to the config file.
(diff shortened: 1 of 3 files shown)


amit-batra:
> This PR changes the OpenCode integration from persistent config mutation to session-scoped configuration.
>
> The original implementation backed up `~/.config/opencode/opencode.json` before modifying it, but did not restore the file when OpenCode exited. A straightforward restore mechanism would still be unsafe for concurrent `omlx launch opencode` sessions because all sessions share the same global configuration file.
>
> This implementation instead gives every OpenCode session its own temporary configuration and uses OpenCode's `OPENCODE_CONFIG` mechanism. The user's persistent configuration is never modified, and concurrent sessions are isolated from one another.

PR #3060, opened 2026-08-23
fix(server): keep the prefill-guard's structured error body on the streaming paths

### LINKED ISSUE #3036: Prefill-guard rejections lose their structured error body on streaming endpoints: the correct handler in _with_sse_keepalive is shadowed by the inner except (100 occurrences, 0 reached it)
### Summary

oMLX already builds a well-formed, machine-readable error body for prefill-guard rejections —
`_prefill_memory_openai_error_body()` produces a root-level `"type": "error"` plus `omlx_code`,
`estimated_bytes` and `limit_bytes`, and `_with_sse_keepalive()` calls it on
`PrefillMemoryExceededError`.

**On the streaming chat/completions paths that handler can never run.** The generator it wraps
catches the exception first and flattens it to
`{"error": {"message": ..., "type": "server_error"}}` — no root `type`, no `omlx_code`, no numeric
fields. Clients that parse the SSE error frame see an unrecognised event and give up.

In 29 days of production logs the inner path fired **100 times** and the correct handler **0 times**.

### Environment

- oMLX **0.6.2** (Homebrew keg, `jundot/omlx` tap), behaviour also read on `origin/main` @ `2df39bfc`
- mlx **0.32.0**, mlx-lm **0.31.3**
- macOS **26.5.2** (25F84), Apple **M4 Max**, 64 GB unified memory
- Models: MLX safetensors, 27B class; client is an agentic harness on `/v1/chat/completions`
  with `stream: true`

### Root cause

Line numbers are from `origin/main` @ `2df39bfc`; the same shape is in the released 0.6.2.

`/v1/chat/completions` composes its generators like this — `server.py:3812`:

```python
return StreamingResponse(
    _release_after_stream(
        _with_sse_keepalive(          # correct handler lives here
            stream_chat_completion(   # ...but this one catches first
                ...
```

The correct handler, `server.py:2259-2263`:

```python
except Exception as e:
    if isinstance(e, PrefillMemoryExceededError):
        logger.warning(f"SSE generator prefill rejected: {e}")
        error_data = _prefill_memory_openai_error_body(e)   # type:"error", omlx_code, estimated_bytes, limit_bytes
```

The generator it wraps, `server.py:4745-4747`:

```python
except Exception as e:
    logger.error(f"Error during chat streaming: {e}")
    error_data = {"error": {"message": str(e), "type": "server_error"}}
```

The inner `except Exception` is the innermost enclosing handler, so `PrefillMemoryExceededError`
raised during prefill never propagates to `_with_sse_keepalive`. The structured body is built by
code that is unreachable on this path.

`stream_completion()` (`/v1/completions`) has the identical pattern at `server.py:4468-4470`. I have
not checked `/v1/messages` or `/v1/responses`.

Note this also swallows `PrefillMemoryAbortedError`, which `_prefill_memory_openai_error_body()`
handles explicitly (`code="prefill_memory_aborted"`).

### Evidence

Two log lines discriminate the two paths exactly, and they can be counted without instrumenting
anything. Over **2026-07-25 03:34:34 → 2026-08-22 09:54:12** (29 days, spanning 0.5.x through 0.6.2):

| log line | meaning | count |
|---|---|---:|
| `Error during chat streaming: …` | inner `except` won → `server_error` | **100** |
| `SSE generator prefill rejected: …` | correct handler ran | **0** |

Of those 100: **66** are `Prefill context too large for available memory (pre-chunk guard at N
tokens, kv_len=…)`, ~10 are `Request aborted: process memory limit exceeded (usage …)`, one is
`Prefill would require ~60.39 GB peak …`, the rest unrelated.

### Client-side consequence

The harness we run parses the SSE error frame with a guard equivalent to
`if (event.type !== "error") return`. Because the flattened body has **no root-level `type`**, the
frame is skipped, the stream ends without a recognised error, and the failure surfaces as a generic
unknown error — not as a context/memory error.

Concretely, that harness has an automatic recovery path (compact the conversation and retry) which
is armed by recognising an overflow error. Across **14 stored oMLX rejections** (2026-07-25 →
2026-08-21) it classified **6** as generic API errors — rejections that arrived as an HTTP 400
before the stream started — and **8** as unknown errors, which are the in-stream ones described
above. **0** were classified as overflow, and **none was followed by a recovery attempt**.

### Suggested fix

Reuse the function that already exists, in both inner generators — this is the same three lines as
`_with_sse_keepalive`:

```diff
 except Exception as e:
-    logger.error(f"Error during chat streaming: {e}")
-    error_data = {"error": {"message": str(e), "type": "server_error"}}
+    if isinstance(e, PrefillMemoryExceededError):
+        logger.warning(f"Chat streaming prefill rejected: {e}")
+        error_data = _prefill_memory_openai_error_body(e)
+    else:
+        logger.error(f"Error during chat streaming: {e}")
+        error_data = {"error": {"message": str(e), "type": "server_error"}}
     yield f"data: {json.dumps(error_data)}\n\n"
```

Alternatively, narrow the inner `except` so `PrefillMemoryExceededError` propagates to the wrapper
that already knows how to render it. Either way no new policy is introduced — it makes the streaming
paths agree with the non-streaming one.

### Reproduction

Any streaming request large enough to trip the prefill guard reproduces it: the response ends with a
`data:` frame whose body is `{"error": {..., "type": "server_error"}}` instead of the structured
form, and the server log shows `Error during chat streaming: Prefill context too large …` with **no**
`SSE generator prefill rejected` line. Lowering the memory-guard ceiling makes it trivial to trigger
with a small prompt.

I have deliberately **not** synthesised a fresh reproduction: on this host it would mean either a
~20-minute prefill or changing the memory-guard flags of a daemon that is in service, and a month of
production logs already contains 100 instances. Happy to capture a raw SSE frame if that would help
— tell me which endpoint you want it from.

### Note on the non-streaming path, which is *not* what I am asking you to change

The 400 handler is well-formed and its docstring makes the reasoning explicit, so this is FYI rather
than a request. For a client the remaining gap is that neither the status code nor `error.code` is
one that generic OpenAI-compatible clients branch on:

| what clients test | what oMLX returns |
|---|---|
| `statusCode === 413` | `400` — deliberate, per the docstring (avoids Anthropic clients collapsing this into "Request too large (max 32MB)") |
| `error.code === "context_length_exceeded"` | `prefill_memory_exceeded` |
| message regex (≈28 known phrasings) | `oMLX prefill memory guard rejected this prompt: …` — matches none |

I am explicitly **not** asking for `context_length_exceeded`: per `server.py:919-926` the prompt does
fit the context window and the real fix is often to loosen the guard, so that code would be wrong.
Flagging it only so you know that today no off-the-shelf client can auto-recover from this rejection
— should you ever want them to, that is the surface to change. Fixing the streaming path above is
the part that is unambiguously a bug.

### Neighbours, for triage

- **#2012** (closed) — same symptom (`server_error` on the SSE path), different cause: malformed
  Harmony tool-call arguments. Precedent for treating flattened SSE errors as a defect.
- **#578** (open) — long-context streaming aborts, but that is client disconnect / `GeneratorExit`,
  not error classification.
- **#1293** (open) — stream termination under a custom OpenAI provider.

beaglemoo:
Fixes #3036.

**Problem** (as diagnosed in the issue): the streaming generators' blanket `except Exception` is the innermost handler, so a `PrefillMemoryExceededError` raised during prefill is flattened to `{"error": {..., "type": "server_error"}}` before the correct handler in `_with_sse_keepalive` can run — the structured body (`type: "error"`, `omlx_code`, `estimated_bytes`, `limit_bytes`) is built by unreachable code. The reporter's production logs show the inner path fired 100 times in 29 days, the correct handler zero.

**Change:**
- New `_streaming_error_payload()` classifies the exception in one place; `stream_chat_completion` and `stream_completion` both use it. Prefill rejections (and `PrefillMemoryAbortedError`, a subclass) keep the full structured body; every other exception keeps the exact previous behavior, including the `Error during chat streaming:` log line the issue used for counting.
- The same shadowing existed on the two paths the issue left unchecked: `/v1/messages` now emits an Anthropic-native `invalid_request_error` event with the guard's detailed message (terminal request error — a retry without shrinking the prompt cannot succeed), and `/v1/responses` now carries `{code, message}` in the `response.failed` error object instead of failing with no error at all.

**Verification:** three regression tests in `tests/test_server_prefill_memory_handler.py` pin the payload for the exceeded/aborted/generic cases. File result: 10 passed, 2 failed — both failures pre-exist on clean `main` in this environment (`pytest-asyncio` not installed locally; they fail identically before and after this change).

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -865,6 +865,23 @@ def _prefill_memory_error_detail(exc: PrefillMemoryExceededError) -> str:
     )
 
 
+def _streaming_error_payload(e: Exception, context: str) -> dict:
+    """Error body for a failure inside an OpenAI SSE generator.
+
+    A prefill-guard rejection raised during streaming must keep its
+    structured body (root ``type``, ``omlx_code``, byte fields) — the
+    blanket ``except`` in the generators would otherwise flatten it to a
+    generic ``server_error`` that clients cannot classify, and the correct
+    handler in ``_with_sse_keepalive`` never sees the exception because the
+    generator's own handler is the innermost one (#3036).
+    """
+    if isinstance(e, PrefillMemoryExceededError):
+        logger.warning(f"{context} prefill rejected: {e}")
+        return _prefill_memory_openai_error_body(e)
+    logger.error(f"Error during {context}: {e}")
+    return {"error": {"message": str(e), "type": "server_error"}}
+
+
 def _prefill_memory_openai_error_body(
     exc: PrefillMemoryExceededError,
     *,
@@ -4488,8 +4505,7 @@ async def stream_completion(
             }
             yield f"data: {json.dumps(data)}\n\n"
     except Exception as e:
-        logger.error(f"Error during completion streaming: {e}")
-        error_data = {"error": {"message": str(e), "type": "server_error"}}
+        error_data = _streaming_error_payload(e, "completion streaming")
         yield f"data: {json.dumps(error_data)}\n\n"
         yield "data: [DONE]\n\n"
         return
@@ -4765,8 +4781,7 @@ async def stream_chat_completion(
                         )
                         yield f"data: {chunk.model_dump_json(exclude_none=True)}\n\n"
     except Exception as e:
-        logger.error(f"Error during chat streaming: {e}")
-        error_data = {"error": {"message": str(e), "type": "server_error"}}
+        error_data = _streaming_error_payload(e, "chat streaming")
         yield f"data: {json.dumps(error_data)}\n\n"
         yield "data: [DONE]\n\n"
         return
@@ -5253,8 +5268,18 @@ async def stream_anthropic_messages(
             if output.finished:
                 break
     except Exception as e:
-        logger.error(f"Error during Anthropic streaming: {e}")
-        yield create_error_event("api_error", str(e))
+        if isinstance(e, PrefillMemoryExceededError):
+            # Same shadowing as the OpenAI generators (#3036): keep the
+            # rejection classifiable. invalid_request_error is the Anthropic
+            # terminal request-error type — a retry without shrinking the
+            # prompt cannot succeed.
+            logger.warning(f"Anthropic streaming prefill rejected: {e}")
+            yield create_error_event(
+                "invalid_request_error", _prefill_memory_error_detail(e)
+            )
+        else:
+            logger.error(f"Error during Anthropic streaming: {e}")
+            yield create_error_event("api_error", str(e))
         yield create_message_stop_event()
         return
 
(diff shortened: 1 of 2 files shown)


jundot:
Good catch on the handler shadowing, and thanks for backing it with the production log counts in #3036. Classifying the exception at the innermost except is the right fix, and covering the /v1/messages and /v1/responses paths the issue didn't check saved a second round. Merging.

PR #3063, opened 2026-08-23
fix(cluster): tier distributed prompt-cache slots by headroom

### LINKED ISSUE #3062: Distributed tuner silently forces prompt_cache_size=1 — configured cache size is a no-op and any concurrent request flushes the shared prefix
## Summary

In distributed deployments the execution-settings `prompt_cache_size` knob is silently overridden to 1 by the headroom tuner, in every tier, regardless of what the UI or the profile requests. With exactly one shared prefix slot, any second request (title generation, summarizer, another client) evicts the session's cached prefill between turns, so long conversations reprefill from scratch on every turn (~110 s at ~45k tokens) while usage reports `cached_tokens: 0`. Under ordinary multi-request agent traffic this makes distributed serving dramatically slower than standalone mode, which honors the same knob.

Pinning one slot was a correctness measure (see root cause), but it is stricter than the invariant requires. Eviction order in MLX-LM's prompt cache is pure recency over the request-event stream, which is identical on every rank, so count-based LRU is coherent for any slot count as long as byte budgets stay disabled. Only byte-based accounting diverges across ranks.

## Environment

- 2x Apple M4 Max 48 GB, macOS 26.5.2
- oMLX 0.6.3rc2 (HEAD-fa3e94b), MLX 0.32.0
- Model `scottlowry/Qwen3.8-27B-oQ4e-mtp`, TP=2, 64 layers, backend jaccl (Thunderbolt 5 RDMA)
- Execution profile balanced (`prompt_cache_size` preset 8), ample-headroom tier

## What happens

1. Settings/dashboard expose `prompt_cache_size` (profile presets 4/8/16). Standalone honors it; distributed never does.
2. `omlx/cluster/performance.py` `tune_execution_settings()` merges `{"prompt_cache_size": 1, "prompt_cache_bytes": None}` into both return paths: the early return when auto-tuning is disabled and the tuned return. The ample-headroom tier's `caps[3]` passthrough of `settings.prompt_cache_size` is therefore dead code; the override wins in every configuration. `tuning_reason` ends with "synchronized single-prefix cache".
3. Runtime confirmation after about an hour of mixed agent traffic: the rank marker reports `entries: 1`, `lookups: 37`, `hits: 33`, but all reuse belongs to short-lived small prefixes. The long working-set conversation never survives an interleave.
4. Client-visible effect: conversation payloads that are strictly append-only between turns (verified by diffing consecutive request bodies) still trigger full prefills every turn; responses report `cached_tokens: 0` while consecutive turns land ~2 minutes apart at ~45k tokens.
5. The same single-slot condition is also the precondition for the mismatched-collective wedge (#3039): under this tuner every concurrent stream is simultaneously a performance hazard and a stability hazard.

## Root cause analysis

Why size=1 exists: MLX-LM keeps a prompt cache independently on every rank. Identical tokens occupy different bytes on different pipeline stages, so a byte budget crosses its threshold at different requests on different ranks. The ranks then retain different prefixes, the next request starts at different token offsets, and the pipeline blocks forever in the first unmatched collective. Byte budgets are genuinely unsafe.

Why size greater than 1 is safe anyway: `mlx_lm.models.cache.LRUPromptCache` evicts by recency alone. Its `pop()` walks insert/fetch history; `max_size` gates entry eviction while `max_bytes` runs a separate loop. Every rank observes the identical event stream in identical order because requests are broadcast in SPMD fashion, so with byte budgets disabled any slot count keeps all ranks structurally identical indefinitely.

The correct invariant is to keep `prompt_cache_bytes=None` unconditionally and allow the slot count to exceed 1, bounded by headroom.

## Proposed fix

Tier the slot count by minimum stage headroom instead of pinning 1, keeping byte budgets disabled everywhere:

- minimum headroom under 4 GiB (critical): 1 slot
- under 8 GiB (low): 1 slot
- under 16 GiB (moderate): 2 slots
- 16 GiB or more (ample): min(configured value, 4), honoring explicit smaller values

Slots accrue lazily after prefills, so nothing is reserved up front; the worst case adds one cached prefix per extra slot. Surface the resolved slot count in `tuning_reason`. Tests cover the tier matrix, user-value honoring and clamping, and that auto-tune-off still pins 1.

Happy to send the PR if this direction looks right.

xunlinkx:
Fixes #3062.

Distributed prompt-cache slots were forced to one on every rank, so configured cache capacity was not available for concurrent requests. This change tiers the slot count from measured stage headroom while keeping byte-based cache budgets disabled so ranks retain the same count-based eviction behavior.

Critical and low headroom keep one slot, moderate headroom allows two, and ample headroom allows up to four or the configured limit. Smaller configured values remain effective.

Validation:

- `.venv/bin/pytest tests/test_cluster_performance.py tests/test_cluster_launch.py -q` — 70 passed
- `.venv/bin/pytest tests/test_cluster_inference_worker.py tests/test_distributed_engine.py -q` — 91 passed
- `.venv/bin/pytest tests/ -q -k 'performance or launch or worker or engine'` — 1018 passed, 4 skipped

### DIFF
--- a/omlx/cluster/performance.py
+++ b/omlx/cluster/performance.py
@@ -159,6 +159,11 @@ class ExecutionSettings:
     prefill_step_size: int = 2048
     prompt_cache_size: int = 8
     prompt_cache_bytes: int | None = None
+    # The slot count the operator or profile asked for, before the headroom
+    # tuner clamped it. Tuning can run more than once on the activation path
+    # (plan-time budgets, then measured rank profiles); clamping from the
+    # already-clamped value would pin the first tier's slot count forever.
+    requested_prompt_cache_size: int | None = None
     max_kv_size: int | None = None
     pipeline_microbatch_size: int = 4
     cache_affinity: bool = True
@@ -197,7 +202,11 @@ class ExecutionSettings:
             raise ValueError(
                 f"ring connections per IP cannot exceed {_MAX_CONNECTIONS_PER_IP}"
             )
-        for name in ("prompt_cache_bytes", "max_kv_size"):
+        for name in (
+            "prompt_cache_bytes",
+            "requested_prompt_cache_size",
+            "max_kv_size",
+        ):
             value = getattr(self, name)
             if value is not None and (
                 not isinstance(value, int) or isinstance(value, bool) or value <= 0
@@ -215,6 +224,7 @@ class ExecutionSettings:
             "prefill_step_size": self.prefill_step_size,
             "prompt_cache_size": self.prompt_cache_size,
             "prompt_cache_bytes": self.prompt_cache_bytes,
+            "requested_prompt_cache_size": self.requested_prompt_cache_size,
             "max_kv_size": self.max_kv_size,
             "pipeline_microbatch_size": self.pipeline_microbatch_size,
             "cache_affinity": self.cache_affinity,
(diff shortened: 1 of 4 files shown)


xunlinkx:
Rebased onto current main and addressed the review findings. `ExecutionSettings` now preserves `requested_prompt_cache_size` across the initial plan and measured-placement retune, so slot tiers can move both up and down with headroom. The tier matrix now checks `tuning_reason`, with regression coverage for double retuning, headroom collapse, and serialization. The cache-slot documentation now states that slots bound prefix count rather than byte cost and that byte-budget eviction remains disabled for rank coherence. Validation: 351 cluster tests passed.

PR #3067, opened 2026-08-23
fix(ane-tuner): return a completed GPU-only verdict when the ANE compiler is unavailable

### LINKED ISSUE #3044: ANE tuning crashes model load on M2 Ultra: "Private ANE procedure-bank compiler is unavailable" → "Representative ANE calibration bank could not be loaded" (no GPU fallback)
- oMLX version: 0.6.3rc2
- Hardware: M2 Ultra, 64 GB RAM, 60-core GPU
- macOS: 26.6.1
- Model: `Qwen3.8-27B-oQ4e-fp16-mtp` (majentik/scottlowry build, oQ4e quant, MTP draft head)

Running the new built-in ANE tuner (introduced in 0.6.3rc2) against the above model fails every bank-split attempt with `Private ANE procedure-bank compiler is unavailable`, then the per-layer fallback also fails with `Representative ANE calibration bank could not be loaded`, and this second failure is not caught — it aborts the entire model load instead of degrading to GPU-only. In 0.6.1, the equivalent unavailable-runtime case (`Private ANE runtime unavailable; Qwen ANE prefill skipped`) logged a warning and continued on GPU normally. This looks like a regression in the tuner's fallback handling, on top of what may be a genuine M2 Ultra ANE compiler gap.

The 0.6.3rc2 release notes state "Apple Silicon requirements now consistently list M1 through M5 across the translated READMEs," so M2 is documented as supported, but I can't find any public report of the ANE tuner or dual-ANE prefill actually succeeding on M2 (Ultra or otherwise) — the only real-world validation I've found covers M3 Ultra, M4 Max, and M5 Max.

### Steps to reproduce

1. oMLX 0.6.3rc2 on an M2 Ultra Mac (64 GB, 60-core GPU), macOS 27.0.
2. Load `Qwen3.8-27B-oQ4e-fp16-mtp` with `force_lm=True`.
3. Run the built-in ANE tuner against the loaded model (Admin → ANE tuning, or equivalent).
4. Tuning fails; model load aborts entirely.

### Actual behavior

```
2026-08-22 23:16:29,446 - omlx.patches.qwen35_ane_prefill - WARNING - [-] - ANE procedure bank compilation failed (1 banks per instance, 10 procedures, 1.20 GiB per instance)
  File ".../omlx/patches/qwen35_ane_prefill.py", line 1733, in _bank_split_ladder
    span0, span1 = compile_span(start, stop)
  File ".../omlx/patches/qwen35_ane_prefill.py", line 1796, in <lambda>
    fast.qwen35_ane_compile_linear_bank(
  File ".../omlx/custom_kernels/qwen35_prefill/fast.py", line 413, in qwen35_ane_compile_linear_bank
    raise RuntimeError("Private ANE procedure-bank compiler is unavailable")
RuntimeError: Private ANE procedure-bank compiler is unavailable
2026-08-22 23:16:29,447 - omlx.patches.qwen35_ane_prefill - INFO - [-] - Retrying ANE procedure banks split at 819 MB per bank
```

This same `RuntimeError: Private ANE procedure-bank compiler is unavailable` repeats identically at every retried bank size (2, 4, 8 banks per instance, down to 102 MB/bank), which suggests the compiler itself is unreachable rather than a memory-sizing problem specific to one bank size.

After all retries are exhausted:

```
2026-08-22 23:16:29,448 - omlx.patches.qwen35_ane_prefill - WARNING - [-] - Packed dual-ANE compilation failed; falling back to per-layer programs
2026-08-22 23:16:29,448 - omlx.admin.ane_tuning - ERROR - [-] - ANE tuning failed for Qwen3.8-27B-oQ4e-fp16-mtp
  File ".../omlx/admin/ane_tuning.py", line 1125, in run_tuning
    choice = await _calibrate_components(run, engine, base_settings)
  File ".../omlx/admin/ane_tuning.py", line 777, in _calibrate_components
    return await loop.run_in_executor(
  File ".../omlx/admin/ane_tuning.py", line 904, in _calibrate_components_sync
    raise RuntimeError("Representative ANE calibration bank could not be loaded")
RuntimeError: Representative ANE calibration bank could not be loaded
2026-08-22 23:16:29,448 - omlx.engine_pool - INFO - [-] - Unloading model: Qwen3.8-27B-oQ4e-fp16-mtp (immediate abort)
```

The model is unloaded and the load aborts entirely — no benchmark or inference happens.

### Expected behavior

At minimum, the "per-layer programs" fallback mentioned in the warning log should actually succeed or gracefully disable ANE prefill and continue on GPU, matching the 0.6.1 behavior for unavailable ANE runtimes (`Private ANE runtime unavailable; Qwen ANE prefill skipped`, engine continues normally). Right now the fallback path itself throws an uncaught exception that kills the whole model load, which is worse than not having the feature at all.

Separately, if the procedure-bank compiler is genuinely unavailable on M2 (Ultra), it would help to know whether that's expected on this hardware generation (in which case the M1–M5 support line in the README may need a caveat) or a bug specific to this build.

### Additional context

- This is a different failure mode than [[#2781](https://github.com/jundot/omlx/issues/2781)](https://github.com/jundot/omlx/issues/2781), which reports `Error Domain=com.apple.appleneuralengine Code=6` on M3 Max — that one is a per-attempt compile failure with a working per-layer fallback. Mine fails identically across every bank size with a "compiler is unavailable" message, and the fallback itself is broken.
- Happy to run additional diagnostics, provide `system_profiler`/`sw_vers` output, or test intermediate builds if useful.

beaglemoo:
Fixes #3044.

**Problem:** on a machine without the private ANE runtime (or a build without the procedure-bank compiler), the tuner discovers the gap only deep inside the bank-split ladder: every width fails with the same `Private ANE procedure-bank compiler is unavailable` error (which is why the reporter saw identical failures at every retried bank size — it was never a sizing problem), the run ends in status `error`, and the model is left unloaded. The serve path has always handled the same condition gracefully (`Private ANE runtime unavailable; Qwen ANE prefill skipped`).

**Change:**
- `fast.qwen35_ane_bank_compiler_available()` — a public probe for the exact condition under which `qwen35_ane_compile_linear_bank` raises; the compile gate now uses the same helper so the two cannot drift.
- `run_tuning` probes it at the top, before anything is pinned or unloaded. When unavailable it returns a **completed** run recommending GPU-only (`enabled: false`, result slots marked skipped with the reason, message naming whether the runtime or just the bank compiler is missing) — on such a machine that is the tuning answer, not an error. Nothing is loaded or unloaded on that path, so whatever was serving keeps serving. The probe fails open so a guard failure can never break tuning on healthy hardware.

**Verification:**
- Two new tests in `tests/test_ane_tuning.py`: the unavailable path yields a completed GPU-only verdict with zero engine-pool interaction, and the probe agrees exactly with the compile gate's raise condition. 27 passed / 11 failed in the file — the 11 pre-exist on clean `main` in this environment and are unchanged by this PR.
- Live smoke on an M5 Pro (where ANE is available): model load through the refactored gate still warms 224 ANE procedures into 2 instance-pinned programs normally.

### DIFF
--- a/omlx/admin/ane_tuning.py
+++ b/omlx/admin/ane_tuning.py
@@ -1991,6 +1991,56 @@ def _calibrate_components_sync(
 
 
 async def run_tuning(run: ANETuningRun, engine_pool: Any) -> None:
+    # The serve path skips ANE gracefully when the private runtime is
+    # missing, but the tuner used to find out only deep inside the
+    # bank-split ladder — failing the run and leaving the model unloaded
+    # (#3044, M2 Ultra). Probe up front, before anything is pinned or
+    # unloaded, and return a completed GPU-only verdict: on such a machine
+    # that IS the tuning answer, not an error.
+    try:
+        from ..custom_kernels.qwen35_prefill import fast
+
+        runtime_available = fast.qwen35_ane_available()
+        bank_available = fast.qwen35_ane_bank_compiler_available()
+    except Exception:
+        # Never let the guard itself break tuning; fail open and let the
+        # ladder report whatever is actually wrong.
+        runtime_available = True
+        bank_available = True
+    if not bank_available:
+        reason = (
+            "the private ANE runtime is unavailable on this machine"
+            if not runtime_available
+            else "the private ANE procedure-bank compiler is missing from "
+            "this build"
+        )
+        logger.warning(
+            "ANE tuning for %s skipped: %s; recommending GPU-only",
+            run.request.model_id,
+            reason,
+        )
+        for result in run.results:
+            result["state"] = "skipped"
+            result["error"] = f"ANE unavailable: {reason}"
+        run.recommendation = {
+            "enabled": False,
+            "mlp_fraction": None,
+            "gdn_enabled": False,
+            "gdn_fraction": None,
+            "fused_down": False,
+            "processing_tps": None,
+            "speedup_percent": None,
+            "sequence_length": run.request.sequence_length,
+            "tail_padding_min_tokens": 0,
+        }
+        run.status = "completed"
+        run.phase = "completed"
+        run.current = run.total
+        run.message = (
+            f"ANE prefill is not usable here ({reason}); GPU-only recommended"
+        )
+        return
+
     previous_speed_priority = _pin_speed_priority(engine_pool)
     active_slot = _GPU_SLOT
     try:
--- a/omlx/custom_kernels/qwen35_prefill/fast.py
+++ b/omlx/custom_kernels/qwen35_prefill/fast.py
@@ -416,12 +416,23 @@ def qwen35_ane_compile_linear(
         return _ext.qwen35_ane_compile_linear(weight, sequence_length)
 
 
+def qwen35_ane_bank_compiler_available() -> bool:
+    """True when both the private ANE runtime and the procedure-bank compiler
+    entry point are present. Callers that would otherwise discover
+    unavailability via the RuntimeError below (the ANE tuner in particular)
+    can probe this up front instead of failing deep inside a compile ladder
+    (#3044)."""
+    return (
+        qwen35_ane_available()
+        and _ext is not None
+        and hasattr(_ext, "qwen35_ane_compile_linear_bank")
+    )
+
+
 def qwen35_ane_compile_linear_bank(
     weights: list[mx.array], sequence_length: int, ane_instance: int
 ):
-    if not qwen35_ane_available() or _ext is None or not hasattr(
-        _ext, "qwen35_ane_compile_linear_bank"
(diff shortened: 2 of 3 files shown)


jundot:
Merging, thanks. The preflight fills a gap the late error path could not cover, machines without the compiler now get an answer instead of a torn down serving state. The earlier CI failures came from the preflight consulting the real probes on runners without the extension, so I pushed a small commit to this branch that stubs them in the mocked pipeline tests. One thing I'll pick up in a follow-up: the GPU-only verdict ships null tps for the first time, which the mac app DTO decodes as non-optional, so I'll make those two fields optional and have the web card skip the tps line when nothing was measured.

PR #3068, opened 2026-08-23
fix(admin): mtp_num_draft_tokens silently dropped by the settings API

alytaphoenix:
## Summary

- `ModelSettingsRequest` never declared `mtp_num_draft_tokens`, so `PUT /admin/api/models/{id}/settings` requests setting it returned 200 but silently discarded the value before it reached the update handler. The model-type default (3 for Qwen3.5/3.6/3.8) was always used regardless of what was requested.
- Also missing from `EnginePool._engine_runtime_signature`, so even a successful update wouldn't have forced the required reload on an already-loaded model (the depth is read once, at model construction time).

## How this was found

While benchmarking Qwen3.8-27B locally, raising `mtp_num_draft_tokens` from the default (3) to 8 via the admin API produced **byte-identical generation** across two separate model reloads — same cycle count, same accept counts, same per-depth breakdown. That's only possible if neither reload ever actually saw a different depth. Tracing the full pipeline (`set_mtp_depth` → per-instance `_omlx_mtp_depth` stamp at model construction → `_DepthController(max_depth=...)`) showed it was all wired correctly — the value just never arrived, because the API schema silently dropped the field.

After the fix, the same real live test showed the depth ceiling actually being exercised (adaptive controller explored depths up to 8, with the expected sharp acceptance decay at deeper drafts), and a measured ~1.05x decode speedup on a coding-style prompt — modest but real, and now actually configurable rather than a no-op.

## Changes

- `omlx/admin/routes.py`: add `mtp_num_draft_tokens: int | None = None` to `ModelSettingsRequest`; apply it in `update_model_settings` when present.
- `omlx/engine_pool.py`: include `mtp_num_draft_tokens` in `_engine_runtime_signature` (gated under `mtp_enabled`, matching the pattern used for `turboquant_kv_bits` etc.) so changing it on a loaded model correctly triggers the auto-unload/reload path.
- Tests: `test_admin_model_settings.py::test_mtp_num_draft_tokens_is_persisted`, `::test_mtp_num_draft_tokens_change_unloads_a_loaded_engine`; `test_engine_pool.py::test_runtime_signature_covers_mtp_num_draft_tokens_only_when_mtp_active`.

## Test plan

- [x] New regression tests pass (`tests/test_admin_model_settings.py`, `tests/test_engine_pool.py`)
- [x] Full suite: 10248 passed, 102 skipped, 77 deselected, 0 failures
- [x] Verified live against a running server: PUT now round-trips the value, reload is triggered, and real generation shows the depth ceiling actually taking effect

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -187,6 +187,7 @@ class ModelSettingsRequest(BaseModel):
     dflash_verify_mode: str | None = None
     # Native MTP (mlx-lm PR 990 / PR 15 monkey-patch)
     mtp_enabled: bool | None = None
+    mtp_num_draft_tokens: int | None = None
     # VLM MTP speculative decoding via external assistant drafter (mlx-vlm 191d7c8+)
     vlm_mtp_enabled: bool | None = None
     vlm_mtp_draft_model: str | None = None
@@ -2776,6 +2777,9 @@ async def update_model_settings(
                     detail="MTP and DFlash cannot both be enabled; choose one speculative-decoding path.",
                 )
         current_settings.mtp_enabled = new_mtp_enabled
+
+    if "mtp_num_draft_tokens" in sent:
+        current_settings.mtp_num_draft_tokens = request.mtp_num_draft_tokens
 
     # VLM MTP (mlx-vlm f96138e+, gemma4_assistant drafter)
     if "vlm_mtp_enabled" in sent:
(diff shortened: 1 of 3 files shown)


alytaphoenix:
Rebased onto current main (1 conflict in engine_pool.py: main independently added a \`qwen4_ple_ssd_offload\` signature entry at the same insertion point — both are purely additive, kept both). Full suite: 10,831 passed, 0 failed, 152 skipped, 77 deselected.

PR #3073, opened 2026-08-23
fix(cluster): release Metal memory on clean rank shutdown

xunlinkx:
## Summary

Distributed rank processes exited without ever releasing MLX's Metal buffer pool back to the OS. Standalone model unloads run `_sync_and_clear_cache()` explicitly (`engine_pool.py`), but a rank simply returned from `run_worker()` or handled SIGTERM via `KeyboardInterrupt` with no MLX cleanup at all, leaving the allocator cache (gigabytes for large models) wired at the driver.

Operational signature observed repeatedly on a Thunderbolt/RDMA worker node: after even clean deployment teardowns, ~14 GiB stayed wired with zero owning processes and only cleared on reboot. Warm reboots cleared it reliably; killing ranks harder never did, which points at driver-level retention of never-released GPU allocations.

## What changed

- `omlx/cluster/inference_worker.py`: new `_release_metal_memory()` runs `_sync_and_clear_cache()` plus a phys-footprint before/after and logs how much was returned to the OS. Called on the two clean exit paths only (normal completion and `KeyboardInterrupt`). Failure and watchdog paths keep fast-exit semantics: a wedged rank must not block in synchronize, so cleanup never runs there and cleanup errors are swallowed so the exit status is preserved.
- `tests/test_cluster_inference_worker.py`: reports freed bytes with real numbers, swallows sync errors, and swallows footprint errors without emitting output.

## Validation / Testing

- `.venv/bin/pytest tests/test_cluster_inference_worker.py -q` (53 passed)
- `.venv/bin/pytest tests/ -q -k 'inference_worker or cluster_liveness or launch or performance'` (235 passed, 3 skipped)

Live cluster evidence: worker teardown logs now show a release line of the form "rank metal release: N MiB", and post-teardown wired memory returns to boot baseline instead of requiring a reboot.

## Diffstat

```
 omlx/cluster/inference_worker.py       | 31 ++++++++++++++++++++
 tests/test_cluster_inference_worker.py | 43 ++++++++++++++++++
 2 files changed, 74 insertions(+)
```

### DIFF
--- a/omlx/cluster/inference_worker.py
+++ b/omlx/cluster/inference_worker.py
@@ -20,6 +20,8 @@ from pathlib import Path
 from types import SimpleNamespace
 from typing import Any
 
+from ..utils.metal_sync import _sync_and_clear_cache
+from ..utils.proc_memory import get_phys_footprint
 from .deployment import decode_worker_contract
 from .liveness import PeerWatchdog
 from .memory_guard import (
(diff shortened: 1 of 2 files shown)


xunlinkx:
I rebased onto current main.

Copilot flagged the import block. When I ran the repo's linter myself, the actual problem was import order, not a stray blank line: the two `..utils` imports sat after the single-dot relative imports, which trips ruff's I001. I moved them up and everything is clean: `ruff check` passes and `pytest tests/test_cluster_inference_worker.py` gives me 53 passed.

I left the `ruff format` deltas at lines 694 and 1131 of inference_worker.py alone. They are on upstream main too, so fixing them does not belong in this PR.

xunlinkx:
Closing as its superseded now..

Issue #800, opened 2026-04-16
STT does not seem to work with MLX models


fingon:
**Describe the bug**
STT doesn't seem to work for me with MLX specific models (non-MLX Qwen3-ASR at least is fine)

**To Reproduce**
1. Download the mlx-community/whisper-v3-large-turbo from HF using model downloader (or Qwen3-ASR)
2. Try to use it for STT

**Expected behavior**
STT to happen. Or some error in e.g. model list that it is not compatible.

**Logs**
```
2026-04-16 08:44:52,109 - omlx.engine_pool - INFO - [-] - Loading model: whisper-large-v3-turbo
2026-04-16 08:44:52,109 - omlx.engine.stt - INFO - [-] - Starting STT engine: /Users/mstenber/.omlx/models/whisper-large-v3-turbo
2026-04-16 08:44:52,477 - omlx.engine.stt - INFO - [-] - STT engine started: /Users/mstenber/.omlx/models/whisper-large-v3-turbo
2026-04-16 08:44:52,478 - omlx.engine_pool - INFO - [-] - Loaded model: whisper-large-v3-turbo (estimated: 1.58GB, total: 9.92GB)
2026-04-16 08:44:52,489 - omlx.engine.stt - INFO - [-] - STT transcribe: model=/Users/mstenber/.omlx/models/whisper-large-v3-turbo, file=tmpgkr2sptz.flac (12858220 bytes), language=en
2026-04-16 08:44:52,607 - omlx.server - WARNING - [-] - POST /v1/audio/transcriptions → 500: Processor not found. Make sure the model was loaded with a HuggingFace processor.
```

With `Qwen3-ASR-0.6B-MLX-4bit` I get different error:
```

2026-04-16 21:01:33,597 - omlx.server - WARNING - [-] - POST /v1/audio/transcriptions → 500: Can't load feature extractor for '/Users/mstenber/.omlx/models/Qwen3-ASR-0.6B-MLX-4bit'. If you were trying to load it from 'https://huggingface.co/models', make sure you don't have a local directory with the same name. Otherwise, make sure '/Users/mstenber/.omlx/models/Qwen3-ASR-0.6B-MLX-4bit' is the correct path to a directory containing a preprocessor_config.json file

```

**Desktop (please complete the following information):**
- macOS, 0.35 (also 0.36)

jundot:
[...]

Issue #828, opened 2026-04-17
模型下载器 魔搭社区 页面经常崩溃


1369946045:
目前版本是macos26 系统 用的dmg安装的0.3.6 最新版  之前用过0.3.4 也是有这个问题 

出现问题的时候 整个程序直接崩溃 只能停止 然后 在启动

web页面提示 未安装 ModelScope SDK。请运行: pip install "omlx[modelscope]"

有时候也不提示 直接崩溃

lingfengnet:
日志里面没有关于这个崩溃的信息

lazeyliu:
+1

lazeyliu:
我发现了这个#475

lingfengnet:
lingfeng@lingfengdeMacBook-Air ~ % pip3 install "omlx[modelscope]"
Defaulting to user installation because normal site-packages is not writeable
ERROR: Could not find a version that satisfies the requirement omlx[modelscope] (from versions: none)
ERROR: No matching distribution found for omlx[modelscope]
WARNING: You are using pip version 21.2.4; however, version 26.0.1 is available.
You should consider upgrading via the '/Library/Developer/CommandLineTools/usr/bin/python3 -m pip install --upgrade pip' command.

chushiyun2016:
+1
I just ran into this problem too.

<img width="2140" height="1152" alt="Image" src="https://github.com/user-attachments/assets/7a748d9f-5a57-42a3-bf1c-f44b46ab2333" />
<img width="1754" height="998" alt="Image" src="https://github.com/user-attachments/assets/105ca2b4-5a4f-4789-a6fe-81615b9c614b" />

PR #3078, opened 2026-08-24
fix(grammar): stop structured output only when it must end

Breaknus:
## Summary

- derive grammar completion from a must-end next-token mask rather than the
  first valid complete prefix
- stop only when the mask permits matcher stop tokens and no valid non-stop
  token
- retain and account for the final accepted ordinary token while suppressing
  a physical EOS token

## Why

xgrammar's `is_completed()` means the current prefix is a valid complete
string; it does not mean the prefix cannot be extended. Treating completion as
termination truncates extendable grammars such as `[0-9]+`,
`root ::= "a" | "ab"`, and root integer JSON Schemas.

After the final token of a self-delimiting structured value has been accepted,
the next-token mask provides the stronger signal oMLX needs: if at least one
matcher stop token is allowed and no valid non-stop token is allowed, decoding
must end. The scheduler can then finish the request without waiting to consume
EOS and without dropping the accepted final token.

The bounded JSON Schema / `maxLength` investigation is intentionally excluded
from this PR and will be handled separately with its own public reproduction.

## Tests

- `tests/test_grammar.py tests/test_scheduler.py`: 315 passed, 3 skipped
- `tests/test_scheduler_logits_processors.py`: 39 passed, 3 deselected
- CI-equivalent suite: 10,251 passed, 102 skipped, 77 deselected; the two
  remaining failures are unchanged local baseline failures reproduced on
  clean `main` (`test_target_ops_logits_last_only_slices_before_lm_head`, MLX
  numeric tolerance; `test_module_entry_api_key_setup_end_to_end`, local 60s
  startup timeout)


jundot:
Thanks for working on this. Stopping a grammar-constrained request at completion instead of running to max_tokens is worth fixing, but I can't take this implementation as is.

`terminate_without_stop_token=True` makes the matcher terminate as soon as the grammar *can* end, not when it *must* end. I verified on the pinned xgrammar 0.2.3: `regex` `[0-9]+` terminates after the first digit, `root ::= "a" | "ab"` can never produce "ab", and a root-level `{"type": "integer"}` schema stops after one digit. Since the flag is applied to the primary matcher for every format, that is a silent output truncation regression for regex, ebnf, choice and scalar schemas. Only self-delimiting shapes like objects and strings are safe.

The maxLength companion also doesn't buy anything I can measure: on the same pin, `{"type": "string", "maxLength": 3}` with the flag terminates on the closing quote by itself, exactly where the companion would. I'd drop that whole layer.

The direction I'd take instead is deriving termination from a must-end signal, e.g. the bitmask the processor already fills each step allowing only stop tokens. Also, do you have a concrete repro of the run-to-max_tokens behavior? I'd like to pin down the actual mechanism before settling on a fix, my current suspicion is a stop token set mismatch between xgrammar's TokenizerInfo and mlx-lm's eos ids.

Breaknus:
Confirmed. I reproduced all three truncation cases on the pinned xgrammar
0.2.3, so the previous revision of this PR was a regression. The mistake was
treating `is_completed()` / “can end here” as “must end here.”

I rewrote the branch around the next-token mask instead:

- use the default `GrammarMatcher` behavior;
- expose `must_end` only when the next mask allows at least one matcher stop
  token and no valid non-stop token;
- let the scheduler convert only that state to a logical `stop`, retaining the
  final ordinary token while still suppressing a physical EOS token.

The regressions now explicitly cover `[0-9]+` after `1`,
`root ::= "a" | "ab"` after `a`, and a root integer after `1`; all remain
extendable, and the EBNF case still accepts `b`. A self-delimiting grammar
reaches `must_end` when only stop tokens remain.

The rewritten branch contains exactly two commits:

1. `55bd86ab` — grammar mask / `must_end` semantics and regressions;
2. `4aaf492b` — scheduler handoff, parser/accounting, and EOS suppression.

Focused grammar/scheduler tests are 315 passed, 3 skipped. The full
CI-equivalent suite is 10,251 passed, 102 skipped, 77 deselected, with only
the same two local baseline failures already reproduced on clean `main`
(MLX numeric tolerance and the 60-second local server startup timeout).

I also removed the maxLength companion completely. I’ll handle the bounded
JSON Schema problem in a separate follow-up PR so this one stays focused on
the terminal handoff.

### Separate bounded-schema issue

The concrete local run reached its 65,536-token completion cap after emitting
a parseable JSON root under a complex bounded schema. Replaying its frozen
token prefix against xgrammar leaves the full bounded matcher non-complete at
that JSON boundary, while a matcher compiled from the same schema with
`maxLength` removed completes on the frozen prefix. The original payload is
private, so I am not treating that as a public repro for this PR.

There is also a small public xgrammar 0.2.3 `maxLength` defect I can reproduce
without a model: a valid escaped JSON string whose decoded length is one is
rejected by `{ "type": "string", "maxLength": 1 }`, while the unbounded
schema accepts it:

```python
import importlib.metadata
import json

import xgrammar as xgr

vocab_size = 128
compiler = xgr.GrammarCompiler(
    xgr.TokenizerInfo([chr(i) for i in range(vocab_size)], stop_token_ids=[])
)
literal = '"\\n"'

def probe(schema):
    matcher = xgr.GrammarMatcher(
        compiler.compile_json_schema(schema),
        terminate_without_stop_token=True,
    )
    accepted = matcher.accept_string(literal)
    return accepted, matcher.is_terminated()

print(importlib.metadata.version("xgrammar"))
print(repr(json.loads(literal)), len(json.loads(literal)))
print("bounded", probe({"type": "string", "maxLength": 1}))
print("unbounded", probe({"type": "string"}))
```

On 0.2.3 this prints:

```text
'\n' 1
bounded (False, False)
unbounded (True, True)
```

That reduced case proves a separate `maxLength`/escaping bug; it does not by
itself prove the run-to-`max_tokens` scheduling mechanism. I’ll keep those two
claims separate and reduce a public end-to-end witness before the follow-up PR.

Breaknus:
### Full frozen bounded-schema repro

I now have permission to publish the complete frozen fixture that I previously
described as private. It is available here:

https://gist.github.com/Breaknus/fe1a082a19caf16fd21348767ddcf077

The Gist contains the exact 3,921-byte schema, the complete 117,996-byte
provider response (including reasoning and the generated tail), and a
model-free replay script.

On xgrammar 0.2.3, `python -B repro.py` produces:

```text
finish_reason length
xgrammar 0.2.3
completion_tokens 65536
content_chars 70964
valid_root_chars 13206
tail_chars 57758
json_schema_valid True
bounded accepted= True completed= False terminated= False
without_maxLength accepted= True completed= True terminated= True
```

So in this concrete run:

- the first JSON root validates against the exact requested schema;
- the response then contains a 57,758-character closing-brace tail and reaches
  the 65,536-token cap;
- replaying that exact valid root through the bounded grammar accepts every
  character but leaves the matcher non-complete;
- removing only the schema's `maxLength` keywords makes the same root complete
  and terminate.

The original HTTP messages/wire body were not retained, so this is not a
byte-for-byte live request replay. It does preserve the exact schema, sampling
metadata, full response, and failing grammar boundary. I am still keeping this
bounded-schema issue separate from the generic must-end handoff in this PR and
will address it in a focused follow-up.

Breaknus:
### Follow-up: the bounded-string failure is in XGrammar, with an oMLX compatibility boundary

I reduced the bounded-schema case further. This is separate from this PR's generic `must_end` handoff, and I do **not** propose broadening #3078 itself.

#### XGrammar root cause

For JSON strings with `minLength` or `maxLength`, XGrammar takes a dedicated lowering path. In xgrammar 0.2.3 that path emits:

```text
[^"\\\r\n]{min,max}
```

but omits the normal JSON escape alternative (`\\` followed by `"`, `\\`, `n`, `uXXXX`, etc.). Therefore a decoded one-character string such as a newline is valid against `{"type":"string","maxLength":1}`, but its JSON spelling `"\n"` is unreachable in the bounded grammar. This also violates JSON Schema's definition of `maxLength` as the number of Unicode code points in the decoded string:

https://json-schema.org/draft/2020-12/json-schema-validation#name-maxlength

This is visible in `JSONSchemaConverter::GenerateString`:

- xgrammar 0.2.3: https://github.com/mlc-ai/xgrammar/blob/557becfb64c503ae9c04344b0047661f43f44320/cpp/json_schema_converter.cc#L1514-L1524
- current `main` (`53da3ff8`): https://github.com/mlc-ai/xgrammar/blob/53da3ff8c0db359ecb6d4b41f154308c82a89aaf/cpp/json_schema_converter.cc#L2369-L2374

One API nuance matters when reading a reproduction: `accept_string(...) == True` only means every input character was accepted; it does not call `is_completed()`. A complete probe must check both values, as the witness below does:

https://github.com/mlc-ai/xgrammar/blob/557becfb64c503ae9c04344b0047661f43f44320/cpp/grammar_matcher.cc#L742-L776

Minimal model-free witness:

```python
import xgrammar as xgr

compiler = xgr.GrammarCompiler(
    xgr.TokenizerInfo([chr(i) for i in range(128)], stop_token_ids=[])
)

def probe(schema):
    matcher = xgr.GrammarMatcher(
        compiler.compile_json_schema(schema),
        terminate_without_stop_token=True,
    )
    accepted = matcher.accept_string('"\\n"')
    return accepted, matcher.is_completed(), matcher.is_terminated()

print("bounded", probe({"type": "string", "maxLength": 1}))
print("unbounded", probe({"type": "string"}))
```

On 0.2.3 and 0.2.5 this prints:

```text
bounded (False, False, False)
unbounded (True, True, True)
```

The complete model-free fixture is here:

https://gist.github.com/Breaknus/fe1a082a19caf16fd21348767ddcf077

#### Upstream status

- [Issue #800](https://github.com/mlc-ai/xgrammar/issues/800), opened on August 3, is still open. It reports that `minLength`, `maxLength`, and `pattern` select a string lowering that drops every RFC 8259 escape (`\"`, `\\`, `\/`, `\b`, `\f`, `\n`, `\r`, `\t`, and `\uXXXX`). Its tokenizer-independent repro verifies the defect in 0.1.33 and 0.2.4.
- [PR #828](https://github.com/mlc-ai/xgrammar/pull/828), "fix: allow JSON escapes and count code points in length-bounded strings," is open, non-draft, and unmerged. Its current head is `e14ff8d9` (two commits, last updated August 24). The first commit restores the JSON escape alternative and excludes raw C0 controls. The second makes `maxLength` count decoded code points correctly, accepting a valid surrogate pair as one code point and rejecting lone surrogates. The PR reports 547 passing converter tests and declares `Fixes #800`; #822 was the earlier escape-only proposal.

I also rechecked the newer packages independently: 0.2.3 and the [current stable 0.2.5](https://github.com/mlc-ai/xgrammar/releases/tag/v0.2.5) reproduce the minimal witness above. The [0.2.6rc1 PyPI pre-release](https://pypi.org/project/xgrammar/0.2.6rc1/) still retains the missing escape branch, and current `main` (`53da3ff8`) still contains the same bounded lowering. There is therefore no released fix yet.

#### Boundary with #3078

#3078 can correctly hand off a real `must_end` signal only after XGrammar produces one. In this bounded-string failure the primary matcher either rejects a required escape or never reaches completion, so there is no terminal signal for #3078 to hand off. This is an upstream compiler defect, separate from the generic oMLX terminal-handoff correction.

#### Proposed oMLX compatibility follow-up

Until XGrammar ships the upstream fix, a narrow, version-bounded compatibility observer could preserve #3078's safety properties:

1. Apply only to normalized, self-delimiting root-object JSON Schemas that contain `maxLength`, and only on known under-completing XGrammar versions where the unchanged primary accepts the full root (0.2.3-0.2.5).
2. Keep the exact original grammar as the **primary** matcher and sole authority for token masks/admission.
3. Compile a private completion observer from a schema-aware copy with only `maxLength` removed, using logical root completion (`terminate_without_stop_token=True`).
4. Feed the observer only after the primary matcher accepts the same token. The observer must never mask tokens, admit output, or mutate the primary matcher's termination state.
5. Let the scheduler finish after the retained final ordinary token when `must_end || completion_observer_terminated`.
6. Validate the completed JSON against the exact original schema before returning it. If the observer rejects/diverges or exact validation fails, fail closed rather than return invalid JSON or silently weaken the primary grammar.

This avoids the regression from the first revision of #3078: the primary matcher remains default, so extendable regex/EBNF/scalar grammars are untouched. It also keeps the maxLength workaround out of the generic terminal-handoff PR.

This observer is only a compatibility bridge for the 0.2.3/0.2.5-style under-completion behavior. It cannot override a primary matcher that rejects a required escape. For that behavior oMLX would need a release containing xgrammar#828 (or a backport), otherwise it should fail closed for affected schemas rather than silently weakening the primary grammar.

Breaknus:
Follow-up with implementation and live validation evidence for `c34913f42abb1fea0c8b83a6b88883202bb1ec12`.

The focused and affected test gates are green: **336 passed, 3 skipped**. The compatibility path keeps the original grammar as the only logits-mask/admission authority and uses the `maxLength`-free matcher only as a completion observer, followed by host JSON Schema validation before converting `length`/no-reason into `stop`.

I also installed the patched `0.6.3rc2` wheel into an isolated local runtime with XGrammar 0.2.3 and ran fresh Qwen3.8-27B-oQ8e-mtp requests:

- A full application-generation Test request using native `structured_outputs`, `enable_thinking=true`, `reasoning_effort=low`, `temperature=0.6`, and `max_tokens=65536` finished with `stop` after 4,030 completion tokens (3,504 prompt tokens). The application host accepted the output and committed the Test snapshot.
- Three later native-structured Test-repair candidates also finished with `stop` after 3,919 / 2,837 / 4,231 tokens, all far below 65,536. They were rejected afterward by an unrelated application-level semantic invariant (the repair carried one changed test file while the full coverage matrix referenced four preserved test files); this is outside oMLX parsing/termination.
- As a fail-closed control, the same retained project request under a hotter `temperature=1.0` produced no complete JSON root and correctly reached `length` at 65,536. The observer did not fabricate an early stop when host validation could not succeed.

So the live evidence covers both sides of the intended boundary: a valid completed root terminates early and retains the final JSON token, while an actually incomplete root still reaches the caller's length limit. No private prompt, response body, or project source is included here.

PR #3082, opened 2026-08-24
fix(scheduler): break the prefill-guard admission bound into loggable terms

alytaphoenix:
## Summary
- Chunked-prefill rejections (`_guard_prefill_chunk`) logged one aggregate cap comparison with no way to tell which contributor actually bound: resident usage, the predicted per-chunk transient, the session-wide observed-max ratchet, or the ANE prefill transient reservation.
- Adds a second `logger.warning` line on every rejection that breaks the admission bound down into its separate terms, so a rejection is diagnosable from one log line instead of re-deriving the breakdown.
- Part of the qwen35 memory-guard investigation (`docs/qwen35-hardening-and-optimization.md` Phase 0.1).

## Finding from live use
Testing against `Qwen3.8-27B-oQ4e-mtp` at long context with this instrumentation live confirmed `ane_prefill_transient_bytes=0.00GB` at rejection time, and `current` (real resident usage) dominating near the cap rather than an overestimated transient. This rules out the ANE reservation as the binding term for this configuration and points at genuine KV-cache growth as the capacity-limiting factor.

## Test plan
- [x] Added `test_guard_rejection_logs_admission_terms_breakdown` asserting the new log line's fields
- [x] `uv run pytest tests/test_prefill_oom_graceful.py -q` — 51 passed
- [x] `uv run pytest tests/ -k "guard_prefill_chunk or admission_transient or prefill_safety or scheduler" -q` — 453 passed

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -3976,6 +3976,30 @@ class Scheduler:
                 round(margin * 100),
                 base_cap / 1024**3,
             )
+            # Instrumentation for the qwen35 memory-guard investigation
+            # (docs/qwen35-hardening-and-optimization.md Phase 0.1): the
+            # admission bound is dominated near the ceiling by terms that
+            # don't shrink with KV bit-depth, but which term actually binds
+            # is unconfirmed. Log each contributor separately so a rejection
+            # is diagnosable from one log line instead of re-deriving it.
+            tracker = self._prefill_transient_tracker
+            observed_max = (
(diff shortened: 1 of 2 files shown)


jundot:
Merged, thanks. This is exactly what the memory-guard investigation needed, one log line that says which term actually bound instead of re-deriving the breakdown by hand. The live finding that current dominates while the ane reservation sits at zero is already useful on its own.

PR #3084, opened 2026-08-24
fix(mtp): gate MTP off for requests sampling with XTC

alytaphoenix:
## Summary
- MTP's rejection-sampling acceptance math (`_accept_lp_for` in `omlx/patches/mlx_lm_mtp/batch_generator.py`) reconstructs temp/top_p/min_p/top_k from the request's sampling params, but XTC (`apply_xtc`, engaged whenever `xtc_probability > 0`) is applied at emit time and has no term anywhere in the acceptance math — the emitted token distribution silently diverges from the target distribution for any XTC request that goes through MTP.
- Fix: mark the `BatchGenerator` as XTC-active at construction time in `Scheduler._create_batch_generator` (`bg._omlx_mtp_disabled_xtc = sampling_params.xtc_probability > 0`) and check it first thing in `_mtp_common_eligible`, so an XTC request never enters the draft/verify cycle in the first place — exactness first, matching the existing pattern for grammar-constrained requests (`_has_grammar_processors`).
- Part of `docs/qwen35-hardening-and-optimization.md` Theme D, item D2.

## Test plan
- [x] `uv run pytest tests/test_mlx_lm_mtp_patch.py -q` — added `test_xtc_sampling_disables_mtp_eligibility` following the existing `test_decode_eligibility_reads_model_instance_flag_not_global` pattern — 130 passed
- [x] `uv run pytest tests/ -k "mtp or scheduler or xtc" -q` — 961 passed, 1 skipped

### DIFF
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -423,6 +423,11 @@ def _mtp_common_eligible(gen_batch: Any) -> bool:
         return False
     if _has_grammar_processors(gen_batch):
         return False
+    # See the assignment in Scheduler._create_batch_generator: XTC sampling
+    # is absent from MTP's rejection-sampling acceptance math, so a request
+    # sampling with xtc_probability > 0 must not speculate.
+    if getattr(gen_batch, "_omlx_mtp_disabled_xtc", False):
+        return False
     return True
 
 
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -3105,6 +3105,11 @@ class Scheduler:
             prefill_step_size=self.config.prefill_step_size,
             stream=self._stream,
         )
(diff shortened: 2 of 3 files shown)


alytaphoenix:
Rebased onto current main (1 conflict in scheduler.py: main independently refactored \`PrefillTransientTracker\` into a dual dense/gathered-regime history since this branch was cut — its plain \`observed_max_bytes\` property still works for this diagnostic log line, so took main's simpler current call). Full suite: 10,829 passed, 0 failed, 152 skipped, 77 deselected.

jundot (approved):
Thanks for catching the XTC distribution mismatch. Keeping XTC requests on standard decoding makes sense while MTP's acceptance math does not account for XTC. I'll complete the gate in a maintainer follow-up: the flag currently lives on `BatchGenerator`, while eligibility reads `GenerationBatch`, and the check needs to follow each request's sampler as batches change. I'll verify request transitions and mixed batches before merging and applying the follow-up.

PR #3085, opened 2026-08-24
fix(gdn): don't run the stock fallback on clobbered conv state

alytaphoenix:
## Summary
- The fused GDN verify prework (`omlx/patches/qwen35_gdn_prework.py`) sets `cache[0]` to the fused kernel's post-update conv state before the delta update call that can raise. On failure, the except branch fell back to `orig_call` without restoring `cache[0]` — the stock path recomputes conv from scratch and expects the *pre-call* state, so it silently re-applied conv on top of an already-advanced state.
- The same except branch could also be reached after `cache.advance(S)` had already run for this call, and `orig_call`'s stock path advances the cache itself — a fallback taken after that point double-advances the offset for the same S tokens.
- Part of `docs/qwen35-hardening-and-optimization.md` Theme E, item E3.

## Fix
- Capture `cache[0]` before the `try` block entirely (not just before the mutation site) so the except branch always has the pre-mutation state, even if an earlier fallible op raises first.
- Restore `cache[0]` to that captured state in the except branch before calling `orig_call`.
- Move `cache.advance(S)` (and its two sibling left-padding/lengths advance calls) to run only after every other fallible step in the success path has already completed — this removes the double-advance risk by construction instead of trying to reverse-engineer and manually roll back the cache's internal advance semantics.

## Test plan
- [x] Added `test_patched_call_restores_conv_state_and_skips_advance_on_failure`, which installs the real patch with `_gated_delta_update_verify_decode` mocked to raise and asserts the stock fallback sees the pre-mutation conv state and `cache.advance` was never called
- [x] Verified the new test actually catches the regression: fails against the pre-fix code (`git stash` on just this file), passes against the fix
- [x] `uv run pytest tests/test_qwen35_gdn_prework.py tests/ -k "gdn or qwen35" -q` — 412 passed, 23 skipped

### DIFF
--- a/omlx/patches/qwen35_gdn_prework.py
+++ b/omlx/patches/qwen35_gdn_prework.py
@@ -210,6 +210,10 @@ def apply_qwen35_gdn_prework_patch() -> bool:
         if disabled["flag"] or not _eligible(self, inputs, mask, cache, gdn_sink, S):
             return orig_call(self, inputs, mask=mask, cache=cache,
                              gdn_sink=gdn_sink, target_verify=target_verify)
+        # Captured before the try block (and before any fallible op) so the
+        # except branch always has the pre-mutation state to restore, even
+        # if the failure happens before this point is normally reached.
+        conv_state = cache[0]
         try:
             B = 1
             mixed_qkv, z, b, a = q35._target_verify_linears(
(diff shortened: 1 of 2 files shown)


jundot:
Confirmed both hazards on main: cache[0] is overwritten before the delta update that can raise, and the advance ran before the tail ops, so the stock fallback could see an already-advanced state either way. Capturing before the try and deferring the advance to the end is the clean way out. Merging.

PR #3086, opened 2026-08-24
fix(cache): stop last-block GDN/rotating state desync on store

alytaphoenix:
## Summary
- `_extract_block_tensor_slice`'s three non-sliceable branches (rotating-family, CacheList non-sliceable sub-caches, ArraysCache/GDN) treated `is_last_block` alone as sufficient justification to store live `cache_data` state as a block's boundary state.
- `store_cache` skips trailing partial blocks by default, so `is_last_block` points at the last *FULL* block, while the live recurrent state may have already ingested tokens beyond it. Storing that live state as the boundary silently double-ingests those tokens on a later restore -- **wrong output across every GDN layer, with no error**. Confirmed via a real end-to-end `store_cache` regression test (7 tokens / block_size=4).
- The split-GDN sidecar layout was already exempt (separate commit path, `#2569`/`#2644`); only the embedded (non-split) layouts were exposed.
- Part of `docs/qwen35-hardening-and-optimization.md` Theme A, item A1.

## Fix
Replace the `is_last_block` shortcut with an explicit `live_state_at_true_end` flag that `store_cache` only sets when it has verified nothing was skipped past this block:
- exact-terminal mode (`store_exact_prefix`, which allocates with `ceil()` so there's no skipped tail at all), or
- an ordinary store where the token count is an exact multiple of `block_size` (no trailing partial to skip).

Every other case now falls to the same placeholder middle blocks already use -- trading a small amount of last-block reuse for correctness. A boundary snapshot, when available, is still always preferred over live state regardless of this flag.

## Test plan
- [x] `test_store_cache_arrayscache_partial_trailing_stores_placeholder` (rewritten): the unsafe partial-skip case now stores a placeholder through the real `store_cache` API, not live state
- [x] `test_store_cache_arrayscache_exact_multiple_stores_live_state` (new): the safe exact-multiple case still stores live state, distinguishing it from the case above
- [x] `test_aligned_exact_prefix_does_not_reuse_placeholder_terminal`: `store_exact_prefix`'s real terminal-block path still round-trips correctly
- [x] Added snapshot-vs-placeholder coverage for the rotating-family and ArraysCache direct-unit-test paths (`test_rotating_last_block_without_snapshot_stores_placeholder` / `..._with_snapshot_stores_snapshot_state`, and their ArraysCache/`test_prefix_cache.py` equivalents)
- [x] Updated pre-existing tests that asserted the old (buggy) behavior as correct
- [x] `uv run pytest tests/ -k "prefix_cache or hybrid_cache or ntuple or gdn or boundary_snapshot or cache" -q` -- 1561 passed, 2 skipped

### DIFF
--- a/omlx/cache/prefix_cache.py
+++ b/omlx/cache/prefix_cache.py
@@ -1107,18 +1107,28 @@ class BlockAwarePrefixCache(CacheManager):
                 ):
                     snapshot_cache_data = boundary_snapshots[block_boundary_tc]
 
+                # True (not just "last") for the last block only when
+                # nothing beyond it was skipped: exact-terminal mode
+                # allocates with ceil() (no trailing partial skipped at
+                # all), and an exact-multiple-of-block_size store has no
+                # trailing partial to begin with. Ordinary store_cache's
+                # last FULL block otherwise may sit behind skipped trailing
+                # tokens the live state has already ingested -- see A1.
+                live_state_at_true_end = is_last_block and (
+                    _store_exact_terminal or trailing_partial_tokens == 0
+                )
+
                 # Continuity check applies only when we will slice live
                 # cache_data for this block. Skipped when:
                 #   1. A boundary snapshot exists for this block — snapshots
                 #      are self-contained, so the live-cache seq_len gate
                 #      does not apply.
-                #   2. is_last_block is True — _extract_block_tensor_slice's
-                #      last-block branch uses cache_data's full state for
-                #      non-sliceable types (RotatingKVCache last window,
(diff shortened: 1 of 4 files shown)


jundot:
Nice catch, and the writeup made this easy to verify. I traced the store path on main and the corruption is real: the boundary-aligned store always lands on an exact multiple so its behavior is unchanged, but the fallback stores that skip a trailing partial were persisting a recurrent state that had already ingested the skipped tokens, which poisons any later restore. Trading the last block's reuse for correctness is the right call. Merging.

PR #3087, opened 2026-08-24
fix(cache): close shared-prefix TOCTOU with hash-validated acquire

alytaphoenix:
## Summary
- The prefix-index fetch path already uses hash-validated `acquire_cached_block` and self-heals on mismatch. The **shared-prefix** path (`find_shared_prefix` -> `increment_ref`) did not: `increment_ref` is membership-only with no hash check, so a concurrent eviction + reallocation of a block between `find_shared_prefix`'s lookup (outside any lock the caller holds) and the `increment_ref` call could splice a block now holding foreign content into the returned chain -- silent positional corruption, no error.
- Part of `docs/qwen35-hardening-and-optimization.md` Theme A, item A2. Builds on top of A1 (#3086) but is independent -- applies cleanly on plain `main`.

## Fix
- `find_shared_prefix` now also returns each block's hash as observed at lookup time (previously it only returned stripped block ids).
- The shared-prefix path in `fetch_cache` uses `acquire_cached_block(block_id, expected_hash)`, which atomically re-checks the hash under the paged-cache lock, and stops the chain at the first mismatch (mirroring the prefix-index path) rather than splicing in a reassigned block or leaving a hole mid-chain.
- If every shared block is already gone/reassigned, the (now-empty) block table is discarded and the call falls through to the prefix-index path instead of returning a hollow hit.

## Test plan
- [x] `test_find_shared_prefix` (updated): asserts the new hashes are non-None and each one round-trips through `acquire_cached_block` successfully
- [x] `test_find_shared_prefix_hash_mismatch_is_rejected_by_acquire` (new): a stale hash from `find_shared_prefix` is correctly refused by `acquire_cached_block` after the block is reassigned
- [x] `test_shared_prefix_hit_rejects_reassigned_block` (new, end-to-end via `fetch_cache`): injects the race at `acquire_cached_block` itself (the one real point between the lookup and the acquire) and confirms the chain stops before the reassigned block instead of splicing it in. Verified this test fails against the pre-fix code and passes against the fix.
- [x] `uv run pytest tests/ -k "prefix_cache or hybrid_cache or ntuple or gdn or boundary_snapshot or cache" -q` -- 1560 passed, 2 skipped

### DIFF
--- a/omlx/cache/paged_cache.py
+++ b/omlx/cache/paged_cache.py
@@ -1203,11 +1203,18 @@ class PagedCacheManager(CacheManager):
         extra_keys: Optional[Tuple[Any, ...]] = None,
         extra_key_token_start: Optional[int] = None,
         extra_key_ranges: Optional[List[Tuple[int, Tuple[Any, ...]]]] = None,
-    ) -> Tuple[List[int], List[int]]:
+    ) -> Tuple[List[int], List[Optional[BlockHash]], List[int]]:
         """
         Find shared prefix blocks for a token sequence.
 
-        Uses get_computed_blocks for consistent chain-hash lookup.
+        Uses get_computed_blocks for consistent chain-hash lookup. Returns
+        each block's hash as observed at lookup time alongside its id, so
+        the caller can acquire it with acquire_cached_block(id, hash)
+        instead of a membership-only increment_ref -- closing the TOCTOU
+        window between this lookup (outside any lock the caller holds) and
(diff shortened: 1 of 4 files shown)


jundot:
Looks right. This brings the shared-prefix path up to the same discipline the prefix-index path already uses, and reusing acquire_cached_block instead of inventing a new mechanism keeps it simple. find_shared_prefix has a single caller so the signature change is contained. Merging.

PR #3090, opened 2026-08-24
fix(turboquant): dense fallback dequantizes only the passed views

alytaphoenix:
## Summary
- The long-prefill dequantize+SDPA fallback (`omlx/patches/turboquant_attention.py`, engaged when `quantized_attention` raises) called `real_cache.dequantize()` with no arguments.
- `dequantize()` defaults `keys_state`/`values_state` to `None`, which dequantizes the **entire resident cache** in one shot instead of just the keys/values views this call actually needs -- instant OOM at long context on exactly the fallback path meant to degrade gracefully.
- Every other call site in this file (`prefill_attention`, `quantized_attention`, and the sink-fallback `dequantize` call a few dozen lines away) already passes `keys_state`/`values_state`; this one call site didn't.
- Part of `docs/qwen35-hardening-and-optimization.md` Theme F, item F4.

## Fix
Pass `keys_state=keys, values_state=values`, matching the established pattern at every other call site in the file.

## Test plan
- [x] Extended the existing `test_attention_patch_falls_back_when_quantized_prefill_fails` (which already exercised this exact fallback path via a `dequantize` call-count spy, but never checked *what* was passed to it) to also capture and assert the kwargs
- [x] Verified the extended assertion fails against the pre-fix code (empty kwargs) and passes against the fix
- [x] `uv run pytest tests/test_turboquant.py -q` -- 52 passed

### DIFF
--- a/omlx/patches/turboquant_attention.py
+++ b/omlx/patches/turboquant_attention.py
@@ -384,6 +384,91 @@ def _decode_multirow_attention(real_cache, queries, keys, values, scale):
     return out.astype(queries.dtype)
 
 
+def _chunked_dense_fallback_attention(real_cache, queries, keys, values, scale, mask):
+    """Dequantize-and-attend over bounded query/KV tiles, never the whole
+    resident cache in one shot (#3090).
+
+    Reached when quantized_attention isn't applicable (below the long-
+    prefill threshold) or raised. ``real_cache.dequantize()`` with no
+    explicit views defaults to the ENTIRE resident cache, and the views
+    the caller can actually pass here (this call's own ``keys``/``values``)
+    are themselves the same full-cache state proxy that default falls back
+    to -- passing them explicitly doesn't narrow anything, confirmed
+    against the pinned mlx-vlm by tracing ``update_and_fetch`` (there is no
+    call path in the pinned wheel where a narrower view reaches this
+    dispatcher). So the only way to bound this fallback's transient is to
+    tile it ourselves.
+
+    Mirrors quantized_attention's own tile schedule (query_block x
+    key_chunk, same constants #3108 already prices this fallback against)
+    and mask/GQA handling (_apply_attention_mask, the (B, n_kv, n_repeats,
+    L, D) reshape) exactly -- but scores real dequantized fp32 keys/values
+    with plain einsum instead of codebook lookups, and combines chunks with
+    a standard online-softmax accumulator (no query-side rotation: that is
+    a quantization-codec concern, irrelevant once the data is real).
+    """
+    from mlx_vlm.turboquant import TurboQuantSplitState
+
+    from ..turboquant_kv import _slice_state_range, _state_length
+
+    keys_state = real_cache._unwrap(keys)
+    values_state = real_cache._unwrap(values)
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for looking at the fallback path.

I checked this against the pinned mlx-vlm and the dequantize range comes out the same before and after. `dequantize()` with no arguments already uses `self.state`, which is sliced to the current offset, and the proxy that `update_and_fetch()` returns points at that same state. I ran both forms on a single sequence cache and on a left padded batch cache, and they returned the same shape and the same values.

So the fallback still expands the whole resident context and the OOM path this is aimed at is unchanged. The new test only asserts that the kwargs get passed, so it does not show a reduction either.

Could you either point me at a call path where a narrower view actually reaches this line, or take the fallback itself down to chunks? I would rather land something that moves the peak than a passthrough of the same range.

alytaphoenix:
You're right -- I checked and there's no call path where a narrower view reaches this line; the request's own keys/values views ARE the same full-resident state proxy `dequantize()` defaults to. Replaced the original commit (rather than stacking a second one on it, since its message asserted the passed-views change fixed the OOM) with an actual chunked implementation.

`_chunked_dense_fallback_attention` tiles both the query and KV axes, reusing `quantized_attention`'s own `query_block`/`key_chunk` constants (the same tile schedule #3108 already prices this fallback against), dequantizing only each KV chunk's sliced state via `_slice_state_range` and combining chunks with a standard online-softmax accumulator. Mask handling and GQA reshaping reuse `real_cache._apply_attention_mask` and the same `(B, n_kv, n_repeats, L, D)` convention `quantized_attention`'s own loop uses, so array masks on batch caches (left-padding) get correct semantics without special-casing. No query-side RHT rotation -- that's a quantization-codec concern, irrelevant once the data is real. Falls back to the old single-shot dequantize+SDPA if the chunked path itself raises, matching this file's existing never-break-a-request pattern.

New numerical-parity test (`test_dense_fallback_chunked_matches_single_shot_reference`) compares the chunked output against the old single-shot reference on a causal call with a cached prefix, with tile sizes forced small enough to exercise multiple query blocks and multiple KV chunks -- `allclose` against the reference, not just matching shape. A second test spies on `dequantize` and confirms every call is scoped to at most one tile-worth of tokens, never the unsliced total.

Full 10,469-test suite passes.

alytaphoenix:
Rebased onto current main (2 real conflicts: main independently added a `real_cache.prefill_attention()` fast-path attempt ahead of the dense fallback — merged so both stay, `prefill_attention()` is tried first and this fix's chunked fallback only engages if it declines). Full suite: 10,662 passed, 0 failed, 150 skipped, 77 deselected.

alytaphoenix:
Refreshed onto current main (clean rebase, no conflicts — 82 commits of drift picked up). The earlier CI failure (\`test_prefix_cache_gdn_split.py::test_split_dedup_recreates_evicted_sidecar\`) is an unrelated pre-existing flake — it also failed independently on #3109 with the same assertion shape on a completely different code path, and passes clean locally. Full suite: 11,912 passed, 0 failed, 201 skipped, 77 deselected.

Issue #1558, opened 2026-05-31
[BUG] Fatal Python SIGABRT / exit code -6 on background cache-saver thread due to cross-thread MLX Stream isolation


ddick2000:
### Description
Under heavy context load when serializing KV Cache to the SSD tiered cache in the background, the oMLX server randomly crashes with `SIGABRT` / `exit code -6`.

This is caused by a fatal C++ exception inside MLX's command encoder system due to cross-thread MLX Stream isolation when calling lazy evaluations in a `ThreadPoolExecutor` worker thread.

---

### Environment & Context
* **Framework:** MLX (Metal backend)
* **OS:** macOS (Apple Silicon)
* **Feature Active:** Paged SSD tiered cache prefix matching

---

#### `server.log` crash tail:
```plain
2026-05-30 22:32:55,173 - omlx.scheduler - INFO - Cache phase timings: ... store_cache_main_dispatch=283.1ms/27 ...
2026-05-30 22:32:55,232 - omlx.server - INFO - Chat completion: 4096 tokens in 88.94s (46.8 tok/s), prompt: 28989
2026-05-30 22:32:55,268 - omlx.cache.prefix_cache - DEBUG - Saved block 67 to tiered cache: tokens [27648:28672], 30 layers
2026-05-30 22:32:55,268 - omlx.cache.prefix_cache - DEBUG - Found KVCache at layer 5 with seq_len=33085
libc++abi: terminating due to uncaught exception of type std::runtime_error: There is no Stream(gpu, 3) in current thread.
```

#### Native Thread Stack Trace:
```plain
  #0: __pthread_kill in libsystem_kernel.dylib
  #1: pthread_kill in libsystem_pthread.dylib
  #2: raise in libsystem_c.dylib
  #3: faulthandler_fatal_error in libpython3.11.dylib
  #4: _sigtramp in libsystem_platform.dylib
  #5: std::__terminate(void (*)()) in libc++abi.dylib
  #6: __cxxabiv1::failed_throw(__cxxabiv1::__cxa_exception*) in libc++abi.dylib
  #7: __cxa_throw in libc++abi.dylib
  #8: mlx::core::metal::get_command_encoder(mlx::core::Stream) in libmlx.dylib
  #9: mlx::core::gpu::eval(mlx::core::array&) in libmlx.dylib
  #10: mlx::core::eval_impl(std::__1::vector<mlx::core::array, std::__1::allocator<mlx::core::array>>, bool) in libmlx.dylib
[...]

Issue #1437, opened 2026-05-27
SIGABRT / Crash on background cache-saver thread due to cross-thread MLX Stream isolation


ddick2000:
# Bug Report: SIGABRT/Crash on background cache-saver thread due to cross-thread MLX Stream isolation

## Summary
The oMLX server randomly crashes with `SIGABRT` / `exit code -6` under heavy context load when serializing KV Cache to the SSD tiered cache in the background.

The crash occurs inside the background `omlx-store-cache` ThreadPoolExecutor worker thread when it reads array bytes via `memoryview(arr)` or `memoryview(arr.view(mx.uint16))`.

This triggers a C++ exception:
`libc++abi: terminating due to uncaught exception of type std::runtime_error: There is no Stream(gpu, 2) in current thread.`
which aborts the Python interpreter immediately.

---

## Root Cause Analysis
In MLX, streams are strictly **thread-local**. 

1. **Lazy Array Evaluation**: The main inference thread schedules evaluation of KV Cache copy/slice arrays asynchronously via `mx.async_eval(*pre_eval_arrays)` under the active `generation_stream` (internal index `gpu:2`).
2. **ThreadPool Dispatch**: The arrays are passed to a background thread worker (`_async_store_cache_worker`).
3. **Implicit Thread-Local Eval**: The background worker thread attempts to read the bytes using Python's buffer protocol (`memoryview(arr)` or `arr.view(...)`). Since the arrays are lazy and not fully evaluated on the GPU, MLX tries to perform a synchronous `arr.eval()` on the worker's current thread.
4. **Stream Registry Failure**: Because the array is bound to the graph computed on `Stream(gpu, 2)` (which is thread-local to the main inference thread and not registered on the background worker thread), MLX's Metal command encoder manager throws a fatal C++ exception:
   `libc++abi: terminating due to uncaught exception of type std::runtime_error: There is no Stream(gpu, 2) in current thread.`
   Since this occurs inside C++ space during memory buffer negotiation, it bypasses standard Python `try-except` blocks and forces a C++ `abort()` (Signal 6).

---

### 1. `server.log` stdout/stderr capture:
```plain
libc++abi: terminating due to uncaught exception of type std::runtime_error: There is no Stream(gpu, 2) in current thread.
2026-05-26 21:53:50,026 - omlx.engine.batched - DEBUG - [stream_generate] Request f38e546b-90c3-40f1-91a4-929508d92aaf finished normally
[...]

Issue #1040, opened 2026-05-02
v0.3.8 crash


TelenLiu:
<img width="469" height="278" alt="Image" src="https://github.com/user-attachments/assets/1f8842a8-ed46-4383-9986-bc727ddeb0ce" />

突然挂了，以下是crash日志：

[crash.log](https://github.com/user-attachments/files/27299557/crash.log)

xy6859:
me too

rwilkins74:
I experienced this numerous times in the last couple days.  I've been using Qwen3.6-35b-a3b-oQ6 and -oQ5 during these experiences.  Several times, while running a code review job with a reasonably large context over 160k, my screen would start flickering different colors and shortly after that the Mac would reboot.  The system crash notification window after rebooting showed python3 as the trigger for the crash.  I don't have a crash log available right now otherwise I'd post it here, but I can try to chase one down if needed.  0.3.6 doesn't seem to exhibit this issue.

I'm on a MacBook Pro M4 Pro with 48 GB RAM.

jundot:
Thanks for the report. The crash.log shows Fatal Python error: Aborted in _do_external_prefill. That is a Metal-side abort (SIGABRT) and is the same family as #978 / #1014 / #1018 on this tracker, plus a known upstream issue: ml-explore/mlx#3186 and ml-explore/mlx-lm#883, which both pin it on the IOGPUMemory completeMemory underflow path in Apple's GPU driver. The restart loop hitting the same crash means the input context is reproducing the same GPU state every time.

To narrow this down on my end I need a few details:
- chip / RAM (e.g. M4 Pro 48GB), macOS version
- model name and quant. The crash.log shows TTS loaded in the same process, so was a TTS request running concurrently with the LLM?
- main memory-related settings (please paste the key entries from Global Settings, e.g. max model memory, hot cache size, max concurrent requests, SSD cache)
- approximate prompt size when it died

In the meantime, things that have helped other users with this crash class: lower the model memory cap (60-70% of total RAM), shrink the prefill step size, avoid running TTS / VLM in the same server process, and cap the active context length.

@rwilkins74 your screen-flicker + reboot is the kernel-panic variant tracked in #978 / #734. If you can grab the panic .ips from /Library/Logs/DiagnosticReports/ after the next reboot and post it on #978 that will help correlate.

jundot:
[...]

PR #3091, opened 2026-08-24
fix(cache): serialize preload's mx.load, don't run it in worker threads

alytaphoenix:
## Summary
- `preload_matched_blocks` ran `mx.load()` (plus array materialization) in a `ThreadPoolExecutor` with up to 8 workers -- directly contradicting this same file's own `load_block` discipline (comment at ~3773-3778), which explicitly documents that a prior executor-based approach **caused deadlocks** when `mx.load()` in a worker thread contested Metal GPU resources with the calling thread (MLX #978 #1040 #1106 #1437 #1558).
- `preload_matched_blocks` runs inline on the scheduler's request admission path (`scheduler.py` -> `block_aware_cache.preload_blocks`, right before `reconstruct_cache`), the same calling thread that also drives inference work -- so it's exposed to exactly the contention `load_block`'s comment warns about.
- Part of `docs/qwen35-hardening-and-optimization.md` Theme F, item F2.

## Fix
Serialize the load loop to match `load_block`'s already-adopted safe discipline -- one `mx.load()` call at a time, always on the caller's thread. Removed the now-unused `ThreadPoolExecutor`/`as_completed` import and the per-call worker-count bookkeeping that existed only to bound its parallelism.

Scope note: the doc's suggested fix also mentions keeping raw file I/O parallel while serializing only the array-materialization step. `mx.load()` doesn't expose a public split between "read bytes" and "materialize array," so implementing that would mean re-implementing safetensors parsing manually -- a materially bigger and riskier change than the actual bug (worker-thread `mx.load()` contesting Metal) calls for. This PR takes the simpler, fully-safe route: serialize the whole load. All 18 pre-existing `preload_*` tests still pass unchanged, confirming no regression to preload's success-path behavior.

## Test plan
- [x] `test_preload_mx_load_runs_serially_on_caller_thread` (new): spies on `mx.load` to record the calling thread and concurrent-call count for every invocation during a 4-block preload; asserts all calls happen on the test's own thread and never more than one is in flight at once
- [x] Verified this test fails against the pre-fix threaded code and passes against the fix
- [x] `uv run pytest tests/test_paged_ssd_cache.py -q` -- 167 passed (all 18 pre-existing `preload_*` tests included, unchanged)

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -29,7 +29,6 @@ import struct
 import threading
 import time
 from collections import OrderedDict
-from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
@@ -4145,11 +4144,8 @@ class PagedSSDCacheManager(CacheManager):
         if len(to_load) < 4:
             return 0
 
-        # Cap workers to limit peak memory (each load allocates ~122-275MB).
-        # 8 workers ≈ 1.4GB peak, vs 2.8GB at 16. CPD-accepted (G1/Q3).
         start = time.perf_counter()
         loaded_count = 0
-        max_workers = min(8, len(to_load))
 
         def _load_one(block_hash: bytes, metadata: PagedSSDBlockMetadata) -> bool:
             file_path = metadata.file_path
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, this checks out. I traced the history: the executor was removed from load_block back in February (9533af99) after real cascading deadlocks, and #1301 reintroduced the same pattern in preload without revisiting that lesson. Serializing gives back tens of ms of TTFT on large cache hits, but I'll take that over a hang that presents as a silent server wedge with no traceback. Merging.

PR #3094, opened 2026-08-24
fix(cache): validate token_count against actual stored slice length

alytaphoenix:
## Summary
- `store_cache` set `block.token_count = len(block_tokens)` and passed it straight to `save_block` without checking it against the actual sequence length of the tensors `_extract_block_tensor_slice` produced for that block.
- A bug anywhere upstream (a wrong start/end index, a stale `cache_data` snapshot, etc.) could silently persist a block whose claimed token count disagrees with what it actually holds -- no error until some later restore replays the wrong number of tokens against it.
- Part of `docs/qwen35-hardening-and-optimization.md` Theme A, item A3.

## Stacked on #3086
This PR is stacked on `fix/gdn-prefix-store-desync-on-last-block` (#3086) -- it adds validation right after the exact `_extract_block_tensor_slice` call site that PR modified (the new `live_state_at_true_end` parameter), so it needs that PR's signature as a base. GitHub can't target a branch that only exists on the fork, so this shows against `main` and will display #3086's commit too until that merges; the second commit (`fix(cache): validate token_count...`) is the actual diff this PR contributes.

## Fix
Add `_block_token_count_matches_slices`: walks a block's sliceable `(keys, values)` 4D tensor entries and confirms their sequence-axis length matches the block's `token_count`. Markers (`__cache_list__`, `__nstate__`) and placeholders don't carry a real per-block sequence length, so their presence alone is never a mismatch -- this only rejects when a sliceable entry's own length actively disagrees. On mismatch, the block is freed and `store_cache` stops there, keeping the already-valid prefix -- the same discipline the existing SSD-save-failure and continuity-break paths already use.

## Test plan
- [x] `test_store_cache_rejects_block_whose_slice_length_disagrees_with_token_count` (new): monkeypatches `_extract_block_tensor_slice` to simulate an upstream bug producing a truncated slice on the second of two blocks, and confirms only the valid first block is kept/persisted, matching the existing `test_store_cache_keeps_valid_prefix_when_later_ssd_save_fails` discipline
- [x] Verified this test fails against the pre-fix code (both blocks stored despite the mismatch) and passes against the fix
- [x] `uv run pytest tests/ -k "prefix_cache or hybrid_cache or ntuple or gdn or boundary_snapshot or cache" -q` -- 1562 passed, 2 skipped

### DIFF
--- a/omlx/cache/prefix_cache.py
+++ b/omlx/cache/prefix_cache.py
@@ -1107,18 +1107,28 @@ class BlockAwarePrefixCache(CacheManager):
                 ):
                     snapshot_cache_data = boundary_snapshots[block_boundary_tc]
 
+                # True (not just "last") for the last block only when
+                # nothing beyond it was skipped: exact-terminal mode
+                # allocates with ceil() (no trailing partial skipped at
+                # all), and an exact-multiple-of-block_size store has no
+                # trailing partial to begin with. Ordinary store_cache's
+                # last FULL block otherwise may sit behind skipped trailing
+                # tokens the live state has already ingested -- see A1.
+                live_state_at_true_end = is_last_block and (
+                    _store_exact_terminal or trailing_partial_tokens == 0
+                )
+
                 # Continuity check applies only when we will slice live
                 # cache_data for this block. Skipped when:
                 #   1. A boundary snapshot exists for this block — snapshots
                 #      are self-contained, so the live-cache seq_len gate
                 #      does not apply.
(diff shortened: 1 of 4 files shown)


jundot:
Merging. This looks speculative at first glance but it has a real target: the slice path currently clamps to min(end_idx, seq_len) and persists a short block with only a debug log, so a store-time length check is exactly where that should get caught. Freeing the block and keeping the valid prefix matches the existing failure discipline.

PR #3095, opened 2026-08-24
fix(cache): fsync before rename for SSD blocks, GDN sidecars, snapshots

alytaphoenix:
## Summary
- All three on-disk cache writers (main SSD blocks, GDN sidecars, and boundary snapshots) share the same write-temp-then-rename pattern via `_write_safetensors_no_mx` + `os.rename`/`os.replace`, with no fsync anywhere in the sequence.
- A crash or power loss between the file `close()` and the rename can leave the rename landing on a file whose data was only ever in the OS page cache -- reading back as truncated/zero-filled garbage, with no error at read time.
- Part of `docs/qwen35-hardening-and-optimization.md` Theme F, item F1 (fsync half only -- see scope note).

## Fix
Two fsync points, applied at all four rename/replace call sites across `paged_ssd_cache.py` and `boundary_snapshot_store.py`:
- `f.flush()` + `os.fsync(fd)` inside `_write_safetensors_no_mx` itself (shared by all four sites), before the temp file is renamed.
- `_fsync_parent_dir()` after each rename/replace, since POSIX doesn't guarantee a rename survives a crash until the containing directory entry is flushed too.

## Scope note
This is the fsync-before-rename half of F1. The doc's other half -- adding a payload checksum (xxhash64) to the format, verified on load and treated as a cache-miss on mismatch -- is a real on-disk format change requiring backward-compat handling for existing files and a new dependency. Deliberately deferred to its own PR rather than bundled with this lower-risk, format-unchanged fix.

## Test plan
- [x] `test_write_safetensors_no_mx_fsyncs_before_close` (new): spies on `os.fsync` to confirm it's called exactly once per write
- [x] `test_fsync_parent_dir_fsyncs_the_directory` (new): confirms the new helper actually fsyncs the containing directory
- [x] `test_fsync_parent_dir_tolerates_missing_directory` (new): confirms it's best-effort and doesn't raise if the directory is gone (e.g. concurrent eviction)
- [x] Verified all three fail against the pre-fix code and pass against the fix
- [x] `uv run pytest tests/test_paged_ssd_cache.py tests/test_boundary_snapshot_store.py -q` -- 200 passed

### DIFF
--- a/omlx/cache/boundary_snapshot_store.py
+++ b/omlx/cache/boundary_snapshot_store.py
@@ -34,6 +34,7 @@ from .paged_ssd_cache import (
     HAS_MLX,
     _encode_shape,
     _extract_tensor_bytes,
+    _fsync_parent_dir,
     _has_zero_dim,
     _restore_tensor_from_bytes,
     _write_safetensors_no_mx,
@@ -977,6 +978,7 @@ class BoundarySnapshotSSDStore:
                     )
                     return False
                 os.replace(str(temp_path), str(file_path))
+                _fsync_parent_dir(file_path)
                 wrote_file = True
                 if self._is_cancelled(pw_key[0]):
                     with suppress(OSError):
@@ -1131,6 +1133,7 @@ class BoundarySnapshotSSDStore:
                 )
                 return
             os.rename(str(temp_path), str(file_path))
+            _fsync_parent_dir(file_path)
 
             # Cleanup may race with a queued write; remove any late file.
             if self._is_cancelled(pw_key[0]):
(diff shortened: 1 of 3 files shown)


jundot:
Thanks, putting the fsync inside the shared writer helper was the right call, every temp-then-rename site funnels through it so the coverage falls out for free. I went over the rename sites and the only one missing the parent dir fsync is vision_feature_cache, I'll push a small follow-up to main for that. Keeping the checksum half as its own PR is the right scoping too, fsync alone doesn't verify payloads and I'm treating it that way. Merging.

PR #3098, opened 2026-08-24
fix(cache): remove decorative pressure-valve methods

alytaphoenix:
## Summary
- `evict_lru_blocks` and `handle_memory_pressure` (`omlx/cache/paged_cache.py`) only ever recycled blocks already sitting in the `free_block_queue` (`ref_count == 0`) -- pop, clear cached content via `_maybe_evict_cached_block`, push back onto the same queue. That never changes `free_block_queue`'s *size*, so `handle_memory_pressure(N)` could never report more capacity than a bare check of `free_blocks >= N` already would -- it was a no-op disguised as pressure relief.
- `reset_prefix_cache` had zero callers anywhere in the codebase (confirmed via full-repo grep, including admin routes and tests) -- genuinely dead code. Its own guard (refusing to run unless almost nothing is allocated: `num_used > 1: return False`) also meant it could never actually fire under real memory pressure, the exact scenario a "pressure valve" exists for.
- Part of `docs/qwen35-hardening-and-optimization.md` Theme F, item F3's pressure-valve bullet: "Either make them evict ref-0 cached blocks for real or delete them so callers stop believing they work." Went with delete: there's no legitimate additional pool of reclaimable capacity beyond what's already in `free_block_queue` without evicting blocks actively held by live requests, which is a materially different (and riskier) semantic change than what these functions ever claimed to do.

## Fix
Removed all three methods. Replaced `handle_memory_pressure`'s one real caller (`store_cache`'s block-allocation retry in `prefix_cache.py`) with a direct bare retry of `allocate_block()` -- functionally identical to what already happened, since the removed indirection never gained capacity beyond what a retry already sees.

## Test plan
- [x] `test_store_cache_degrades_gracefully_when_out_of_blocks` (new): forces `allocate_block()` to fail on the second call, confirms `store_cache` still keeps the valid prefix, logs a clear warning, and doesn't raise
- [x] Removed `test_handle_memory_pressure` (tested a now-deleted method)
- [x] `uv run pytest tests/ -k "prefix_cache or paged_cache or hybrid_cache or ntuple or gdn or boundary_snapshot or cache" -q` -- 1558 passed, 2 skipped

### DIFF
--- a/omlx/cache/paged_cache.py
+++ b/omlx/cache/paged_cache.py
@@ -1314,43 +1314,6 @@ class PagedCacheManager(CacheManager):
     # Eviction
     # =========================================================================
 
-    def evict_lru_blocks(self, num_blocks: int) -> int:
-        """
-        Evict least recently used blocks.
-
-        With the doubly linked list, LRU blocks are already at the front
-        of the free queue. We just need to pop from front.
-        """
-        with self._lock:
-            evicted = 0
-
-            # Get evictable blocks from free queue (they're already LRU ordered)
-            for _ in range(min(num_blocks, self.free_block_queue.num_free_blocks)):
-                try:
-                    block = self.free_block_queue.popleft()
-                    self._maybe_evict_cached_block(block)
-                    # Put back at end (now available for allocation)
-                    self.free_block_queue.append(block)
-                    evicted += 1
-                except ValueError:
-                    break
-
-            if evicted > 0:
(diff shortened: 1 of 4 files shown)


jundot:
Verified the claims: allocate_block already grows the pool and evicts cached content when it pops from the free queue, so the pressure valve could never produce capacity a bare retry would not see, and reset_prefix_cache has zero callers. Always glad to delete code. Merging.

PR #3103, opened 2026-08-24
fix(memory): shed ANE prefill banks under recurring prefill-headroom pressure

beaglemoo:
## Problem

The compiled ANE prefill procedure banks keep their packed weight blobs mapped for the model's whole residency — on a 27B (Qwen3.8 oQ4e, mlp 0.35 / gdn 0.45) that is ~13 GB, visible in the load line: `actual: 29.75GB` with ANE vs `16.02GB` without. A split tuned at a short calibration length therefore silently competes with the KV cache at long context: once usage crosses the prefill guard's sizing target, chunks collapse `2048 -> 512` and prefill crawls.

Measured on an M5 Pro 64 GB (57K-token prompt, fresh nonce per run, model preloaded + warmup request before every timed run, SSD cache cleared between runs, arms interleaved ON/OFF/ON/OFF on a warm long-lived server):

| Arm | prefill time | tok/s |
|---|---|---|
| ANE on | 262.8 s / 322.0 s | 217 / 177 |
| ANE off | 174.8 s / 190.0 s | 326 / 300 |

Both ON runs logged `Prefill throttled: chunk 2048 -> 512`; neither OFF run throttled. The same config is +13% at 16K and still positive at 32K — the recommendation inverts with depth purely through this guard interaction. Notably, on a freshly started process with a loose baseline the same ANE-on 57K prefill ran ~350 tok/s with no throttling at all — the banks are only a problem when the process baseline + KV puts usage near the target, which is exactly what a long-lived server looks like.

## Why the existing ladder cannot recover this

Idle-model eviction skips the requesting model by design, and the pooled-buffer reclaim keeps "succeeding" marginally on every pass — prefill continuously refills MLX's buffer cache, so each pause frees a few transient GB, satisfies the target check, and buys only a couple of minutes of KV growth. Observed ladder trace before this change: pause 1 reclaims 4.0 GB, pause 2 reclaims 7.3 GB, retries exhausted, throttle for the rest of the request. The 13 GB of bank memory is never touched.

## Change

1. **`release_qwen35_ane_prefill(model)`** (qwen35_ane_prefill.py): latches every sliced module through the existing per-module failure flags — the dispatch sites fall back to stock GPU compute exactly as after a warmup failure, and the flags prevent the lazy recompile — then drops the state references and zeroes the status counters (`/api/status` reports `attempted: true, configured: false`). The native programs free their mapped blobs when the last reference dies. Idempotent; the next load of the model rebuilds the banks from its settings.

2. **EnginePool ladder**: a new last rung sheds the requesting model's banks, running on that engine's own MLX thread (its step loop is parked awaiting the eviction callback, so nothing is mid-dispatch). Ordering matters, so a request whose earlier pass already got a pooled reclaim **escalates straight to the release** instead of reclaiming again first — otherwise the marginal reclaim starves the durable rung forever. Bounded FIFO bookkeeping; a failing release is contained and the request throttles exactly as before.

3. **Scheduler**: `_MAX_PREFILL_EVICTION_RETRIES` 1 -> 2, so recurring pressure can reach the deeper rung after a marginal first-pass reclaim. This is a deliberate contract change (the "does not loop on eviction" test now allows one more pause before throttling); the second pause is bounded the same way the first is — every rung is attempt-once per call, and a pass with nothing left to give costs one ~100 ms pause.

Also adds INFO logs for the release's zero paths (model not resolvable / no slices), since a silent 0 here cost us a debugging round.

## Verification

- Unit: 5 new tests (release latches + drops + zeroes and the last reference dies via weakref; idempotency; the ladder reaches the release when the reclaim is not enough; recurring pressure escalates release-before-reclaim; the updated scheduler retry contract). `test_engine_pool.py` + `test_prefill_oom_graceful.py` + `test_qwen35_ane_prefill.py`: **297 passed, 0 failed**.
- End-to-end on real hardware (M5 Pro, 57K prompt, enforced custom ceiling to make the pressure deterministic):

```
12:15:05 scheduler   - Request ... needs prefill headroom before throttling (current=34.70GB, predicted=4.17GB, target=38.70GB)
12:15:05 engine_pool - Reclaimed 1.45GB of pooled Metal buffers for prefill request ... (no idle model to evict)
12:15:48 scheduler   - Request ... needs prefill headroom before throttling (current=37.42GB, predicted=1.49GB, target=38.70GB)
12:15:48 engine_pool - Released ANE prefill banks on 'Qwen--Qwen3.8-27B-oQ4e-mtp' for prefill headroom (request=..., 112 modules, 2 programs, freed 13.95GB); the model serves GPU-only prefill until its next load
```

  No throttle after the release — the request finished at full 2048-token chunks (~252 tok/s under a 43 GB ceiling); `/api/status` reflects the shed state; subsequent requests serve normally on the GPU fallback.
- No-pressure control: with a loose baseline the rungs never run and three consecutive ANE-on 57K prefills measured ~350 tok/s — no overhead when the guard is quiet.

Happy to run any additional scenario on this machine (dual-ANE, fused_down, tuner interaction).

### DIFF
--- a/omlx/engine_pool.py
+++ b/omlx/engine_pool.py
@@ -18,6 +18,7 @@ import gc
 import json
 import logging
 import time
+from collections import OrderedDict
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, replace
 from pathlib import Path
@@ -286,6 +287,13 @@ class EnginePool:
         self._settings_manager: object | None = None  # Set by server
         self._cluster_registry: ClusterRegistry | None = None  # Set by server
         self._suppress_ttl: bool = False  # Suppress TTL during benchmarks
+        # Requests whose prefill already got a pooled-buffer reclaim pass.
+        # Prefill continuously refills MLX's buffer cache, so the reclaim
+        # rung can "succeed" marginally on every pass of a long prompt while
+        # the durable rung behind it (ANE bank release) is never reached; a
+        # request coming back for more headroom escalates instead of
+        # reclaiming again first. Bounded FIFO — request ids are transient.
+        self._prefill_headroom_recurring: OrderedDict[str, None] = OrderedDict()
         self._load_seconds_per_gb_ema: float | None = None
         self._load_time_observations: int = 0
         self._lease_release_tasks: set[asyncio.Task[None]] = set()
@@ -1846,6 +1854,10 @@ class EnginePool:
 
         evicted_any = False
         reclaim_attempted = False
+        ane_release_attempted = False
+        # Snapshot once per call: "a PREVIOUS pass for this request already
+        # got a reclaim". Marking inside this call must not flip it.
+        recurring = request_id in self._prefill_headroom_recurring
         async with self._lock:
             while True:
                 current = max(
@@ -1861,7 +1873,7 @@ class EnginePool:
                     # process-wide and can be masked by concurrent
                     # allocation, but reaching this check with headroom
                     # after an attempt means admission will now succeed.
-                    return evicted_any or reclaim_attempted
+                    return evicted_any or reclaim_attempted or ane_release_attempted
 
                 victim = self._find_lru_prefill_eviction_victim(
                     exclude_model_id=exclude_model_id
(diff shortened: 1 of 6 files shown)


jundot:
Thanks for the writeup, the throttle inversion numbers make the case on their own. Merging. I'll follow up on main so shedding also zeroes the ANE reservation in the memory monitor, and so the shed logs as a warning and shows up in /api/status instead of reading like a compile failure.

PR #3105, opened 2026-08-24
fix(ane): roll back single/dual dispatch on mid-flight exceptions (C1)

alytaphoenix:
## Summary

The single-instance (`AneHybridQ4Primitive`) and dual (`DualAneHybridPrimitive`) ANE dispatch paths in `qwen35_ane.mm` retain the producer command buffer and call `model_->begin()` **outside any try**. The enclosing try only starts after the detached evaluation thread is spawned.

If an exception fires **after `begin()` but before the spawn** — most realistically from the qmm `device.get_kernel()` calls that sit squarely in that window — the ticket is orphaned: `begin()` has already done `++submitted_`, but nothing ever does the matching `++completed_`. The next `begin()` then sees `submitted_ != completed_`, waits out the grace period, latches, and throws **"overlapping evaluations"** — the program is permanently wedged (every later ANE dispatch fails) until the process restarts. The retained command buffer also leaks.

This is finding **C1** in the Qwen3.5 ANE hardening review (HIGH, conditional on a post-`begin()` throw).

## Fix

Add a small `AneDispatchGuard` RAII helper. Armed right after `begin()` hands out a ticket, on stack unwind it:
- releases the producer command buffer (unless `producer_released()` already retired it at the pack step, to avoid a double release), and
- calls `cancel_ticket()` on the outstanding ticket(s) — the existing per-submission rollback that does `++completed_`, signals done, wakes waiters, and leaves the program un-latched.

`begin()` itself is wrapped in a narrow `try` that only releases the buffer and rethrows — `begin()` throws *before* it increments the counter, so there is no ticket to cancel in that case. This mirrors the existing **fused-path** pattern (`begin_ane_ticket_pair` / the `AneHybridQ4SwiGLUDownPrimitive` retain+begin try). The guard then extends protection across the `get_kernel` window (which the narrow pattern alone does not cover) and is disarmed once the evaluation thread(s) take ownership of the ticket. The dual path disarms **before** the first spawn so the guard can never double-count a ticket a thread already owns.

The redundant manual release+cancel in each path's `pack_buffer_failed` branch is folded into the guard (same behavior, one code path).

### Scope note — fused path
`AneHybridQ4SwiGLUDownPrimitive` already guards `retain`+`begin` (added in d2110d1c / #3002) but has the **same residual `get_kernel`-after-`begin` gap**: a throw from `hybrid_qmm_kernel()`/`swiglu`/`down_qmm` after `begin()` succeeds would orphan its ticket the same way. This PR intentionally leaves the fused path unchanged to keep the diff focused on the two paths named in the finding; the same `AneDispatchGuard` applies cleanly there and is recommended as a focused follow-up.

## Test plan

Verified on the target machine (M-series, macOS 26.5 SDK, Xcode 26 Metal toolchain), building the extension against the pinned `mlx==0.32.0` + `nanobind==2.13.0`:

- **Clean rebuild, unmodified source** — full pipeline builds (`_ext.so`, `libomlx_qwen35_prefill_kernel_ops.dylib`, both metallibs) as a baseline.
- **Rebuild with this change** — compiles cleanly (only pre-existing `BNNSMatMul` deprecation + macOS-version link warnings; no new warnings, guard is used).
- **ABI load test in the packaged app's runtime** (bundled CPython 3.11 + app's `mlx` 0.32.0): the built `_ext` imports, `abi_probe(mx.zeros((1,)))` passes (nanobind `NB_DOMAIN` casters match the app's mlx wheel), and all ANE symbols are present — for both the baseline and the fixed build.
- **Correctness** is by construction + code review of the counter/ownership invariants; I did **not** exercise the ANE dispatch path at runtime or inject a fault between `begin()` and the spawn (that would require enabling the ANE path on a live server and a Qwen3.5/3.6 prefill, which was out of scope for this verification). The build + ABI-load results prove the happy path is not regressed; the exception-safety improvement is a static control-flow guarantee.

### DIFF
--- a/omlx/custom_kernels/qwen35_prefill/csrc/qwen35_ane.mm
+++ b/omlx/custom_kernels/qwen35_prefill/csrc/qwen35_ane.mm
@@ -1127,6 +1127,56 @@ static AneTicketPair begin_ane_ticket_pair(
   }
 }
 
+// Rolls an in-flight ANE dispatch back if the stack unwinds after begin()
+// has handed out a ticket but before the detached evaluation thread(s) take
+// ownership of it (finding C1). The dangerous window is everything between
+// begin() and the spawn — most notably the qmm get_kernel() calls, which can
+// throw. Without this guard such a throw would leave submitted_ ahead of
+// completed_ forever and leak the retained producer command buffer, so the
+// next begin() throws "overlapping evaluations" and the program stays wedged
+// until the process restarts. On unwind the guard releases the producer
+// buffer (unless producer_released() already retired it at the pack step) and
+// cancels the outstanding ticket(s) to rebalance the counters. Call disarm()
+// once ownership has transferred to the evaluation thread(s).
+class AneDispatchGuard {
+public:
+  AneDispatchGuard(MTL::CommandBuffer *producer_buffer,
+                   std::shared_ptr<AneLinearModel> model0,
+                   AneLinearModel::Ticket ticket0,
+                   std::shared_ptr<AneLinearModel> model1 = nullptr,
+                   AneLinearModel::Ticket ticket1 = {})
+      : producer_buffer_(producer_buffer), model0_(std::move(model0)),
+        ticket0_(ticket0), model1_(std::move(model1)), ticket1_(ticket1) {}
+  AneDispatchGuard(const AneDispatchGuard &) = delete;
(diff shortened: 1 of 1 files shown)


jundot:
Merging, thanks. The guard closes the orphan ticket window between begin and spawn, and cancel_ticket keeps the counters converging instead of wedging. I'll extend the same guard to the fused path you called out in the description in a follow-up on main.

PR #3108, opened 2026-08-24
fix(memory-guard): register TurboQuant's tiled prefill transient per-instance

alytaphoenix:
## Summary
- TurboQuant's long-prefill route (`quantized_attention`, engaged once KV length exceeds its own threshold) tiles at a bounded 256×16384 query_block×kv_tile, but the prefill admission guard never knew about it -- it priced turboquant-covered prefill with the full O(L^2) unfused score-matrix estimate. Safe direction (over-conservative, false-positive admission rejections at long context on turboquant-enabled models), not itself dangerous, but wasteful.
- Part of `docs/qwen35-hardening-and-optimization.md` Theme B, item B3/3.2.

## Design note (why this isn't a literal mirror of sdpa256's registration)
The obvious fix -- reuse `sdpa256_attention.py`'s `register_tiled_prefill_head_dim` call -- is unsafe for TurboQuant. That registry is a bare module-level global, correct for sdpa256 only because that kernel is installed by monkeypatching `scaled_dot_product_attention` itself: once installed, every model sharing head_dim=256 genuinely takes the tiled kernel, process-wide. TurboQuant's tiled route is gated on a per-scheduler config (`_turboquant_kv_bits`), not head_dim alone, and `engine_pool.py` keeps two models resident during a swap -- a naive global registration would leak one model's turboquant config into a concurrently-resident non-turboquant model sharing the same head_dim, silently under-charging that OTHER model's genuinely-unfused prefill (a real admission under-count / OOM risk, the opposite failure direction from the bug being fixed).

## Fix
A per-`MemoryMonitor`-instance override (`register_tiled_prefill_tile`/`clear_tiled_prefill_tile`, `_tiled_prefill_override` field), set from `Scheduler._set_model_info_for_monitor` gated on the same eligibility check already used for the KV dtype-width override. `_estimate_sdpa_activation_bytes` checks this instance override before the module-global head_dim registry, and caps the charged query length at `query_block` (256) since TurboQuant's kernel tiles the query axis too -- charging the raw `query_tokens` would silently reintroduce an O(query_tokens) over-charge on large chunks.

## Test plan
- [x] New unit tests in `test_memory_monitor.py` (bounded transient, query-block capping, min_kv_len gate, clear restores fallback, and a dedicated per-instance-isolation test proving one model's registration does NOT leak into a second monitor instance sharing the same head_dim)
- [x] New tests in `test_memory_monitor_vlm_config.py` verifying `Scheduler._set_model_info_for_monitor` registers/clears correctly across turboquant-active, turboquant-disabled, MLA-ineligible, and cache-type-ineligible cases
- [x] Verified all new tests fail against pre-fix code (`AttributeError`/`assert_called` mismatches), confirmed via `git stash`
- [x] Fixed two pre-existing `MemoryMonitor.__new__`-bypass tests in `test_sdpa256_attention.py` that needed the new field initialized
- [x] `uv run pytest tests/test_memory_monitor.py tests/test_memory_monitor_vlm_config.py tests/test_sdpa256_attention.py -q` -- 129 passed in this worktree (origin/main base; full count differs from the rollup branch, which has more tests from other in-flight fixes)
- [x] Confirmed via `git stash` that a pre-existing flaky test (`test_should_route_gate`, fails only in combined multi-file runs due to cross-file headroom-provider global-state leakage) reproduces identically with this fix stashed out -- not caused by this change, not fixed here (noted in the design doc as a tracked follow-up)

### DIFF
--- a/omlx/memory_monitor.py
+++ b/omlx/memory_monitor.py
@@ -222,6 +222,19 @@ class MemoryMonitor:
         # full-attention formula. Empty for non-hybrid models.
         self._rotating_layer_specs: tuple[tuple[int, int], ...] = ()
         self._prefill_memory_profile: PrefillMemoryProfile | None = None
+        # Per-instance tiled-prefill override for kernels whose tiled route
+        # is gated on THIS scheduler's own config (e.g. TurboQuant's
+        # quantized_attention, gated on turboquant_kv_bits), not on head_dim
+        # alone. Unlike _SDPA_TILED_PREFILL_HEAD_DIMS below (a module-level
+        # global, correct only for kernels that are unconditionally active
+        # process-wide for a head_dim, e.g. the sdpa256 monkeypatch), this
+        # must stay scoped to this MemoryMonitor/Scheduler instance:
+        # engine_pool.py keeps two models resident during a swap, and a
+        # global registration would leak one model's turboquant config into
+        # another concurrently-resident model sharing the same head_dim,
+        # under-charging that other model's admission guard.
+        # (query_block, kv_tile, min_kv_len); see register_tiled_prefill_tile.
+        self._tiled_prefill_override: tuple[int, int, int] | None = None
         # Fixed-shape ANE prefill I/O surfaces (issue #2841); set via
         # set_model_info, 0 unless the Qwen ANE prefill backend is attached.
         self._ane_prefill_transient_bytes: int = 0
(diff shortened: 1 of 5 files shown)


jundot:
Thanks for the work here. The per-instance scoping makes sense to me, engine_pool does keep two models resident through a swap so a global registration would leak into the other one.

Two things in the sizing though. The tile is charged at `_score_dtype_size`, which is the compute dtype, so 2 bytes on a bf16 model. In the pinned mlx-vlm the TurboQuant prefill kernel builds `output`, `normalizer` and `max_score` as `mx.float32`, and the scores are fp32 as well. At 32 query heads, query 2048 and KV 32768 this branch charges 320 MiB while the score tile and output alone come to 576 MiB.

`turboquant_skip_last` also defaults to True, so the last KV layer stays dense, but the override is registered for the whole monitor and the estimator has no per-layer view. The override branch returns before the head_dim registry, so that dense layer gets the tile price too. In the configs this is aimed at, so sdpa256 disabled or a head_dim outside the fused and tiled sets, main charges that layer the full unfused matrix and this branch drops it to the 256 x 16384 tile. At the size above that is about 4 GiB priced as 320 MiB.

Could you price the largest transient across the layer routes actually in play, use fp32 for the score tile, and add a case covering skip_last with sdpa256 disabled?

alytaphoenix:
Rebased onto current main (clean, no conflicts) and pushed both fixes.

1. **Score-tile dtype**: the tile was priced at `self._score_dtype_size` (the model's compute dtype, 2 bytes for bf16), but the pinned mlx-vlm kernel builds its score tile, `output`, `normalizer`, and `max_score` as `mx.float32` unconditionally (confirmed against the pinned `quantized_attention` source). Added a `_TURBOQUANT_SCORE_DTYPE_SIZE = 4` constant used only for this tile, independent of the model's own compute dtype. Checked the arithmetic against your numbers: at 32 heads / 2048 query / 32768 KV, `output` (already fp32-priced, unaffected) is 64 MiB and the tile at fp32 is 512 MiB -- 576 MiB total, matching what you flagged.

2. **skip_last pricing**: `register_tiled_prefill_tile` now takes a `skip_last` flag (threaded from the same `turboquant_skip_last`-in-effect condition the KV-dtype averaging already uses). When set, `_estimate_sdpa_activation_bytes` computes both the tiled TurboQuant estimate and the dense estimate (the same head_dim-tiled-registry-or-unfused-fallback logic every non-TurboQuant call takes, now extracted into `_estimate_dense_sdpa_activation_bytes` so it's shared, not duplicated) and returns the larger. Added a test with `head_dim=384` (outside every fused/tiled set) reproducing your exact scenario: the dense skip_last layer's true cost is the full unfused matrix, and the guard now prices that, not the 320 MiB tile.

Full 10,480-test suite passes.

PR #3109, opened 2026-08-24
fix(scheduler): decay the prefill admission observed-max ratchet

alytaphoenix:
## Summary
- `PrefillTransientTracker.observed_max_bytes` only ever increased: once a floor-size prefill chunk measured some transient byte count under the 4GiB outlier clamp, that value permanently raised the admission floor for the rest of the process's uptime, even if it never recurred. On a continuously-loaded model (no swap for hours/days) a single early pathological-but-legitimate spike would keep rejecting perfectly fine prompts for the rest of that session.
- Part of `docs/qwen35-hardening-and-optimization.md` Theme B, item B4/3.4.

## Investigation note
The doc suggested two possible fixes: decay/windowed-max, and/or reset on model swap. Verified directly that "reset on model swap" is **already true by construction** -- `engine_pool.py` always sets `entry.engine = None` on eviction and constructs a brand-new engine (hence a fresh `Scheduler` -> fresh `PrefillTransientTracker` starting at 0) on the next load, both for swapping to a different model and for reloading the same model after eviction. No code change needed for that half; the real gap is only within a single continuously-loaded model's uptime.

## Fix
`PrefillTransientTracker.update`'s `floor_sample` branch now decays `observed_max_bytes` toward a lower reading by `_OBSERVED_MAX_DECAY=0.98` per sample instead of leaving it frozen at its all-time high. A higher reading still raises the max instantly -- a fresh spike is real evidence the moment it's seen; decay only ever pulls the bound down. The rate is deliberately slow: this is a safety floor, not a rolling average, so one lucky low reading right after a real spike shouldn't immediately erase the protection that spike earned.

Kept the existing scalar `_observed_max_bytes` representation rather than a windowed max (the doc's other suggested option): four existing tests white-box-poke that attribute directly (`test_prefill_oom_graceful.py`, `test_scheduler_prefill_memory_guard.py`) to seed guard-abort scenarios, and decay-in-place needs zero changes in those files.

## Test plan
- [x] New `TestObservedMaxDecay` class: lower reading decays rather than freezes; sustained lower readings over 400 samples eventually erode a stale spike close to the real floor; a fresh higher reading still raises instantly with no decay lag; clamp-rejected outliers still don't touch the max at all
- [x] Updated one existing test (`test_max_tracks_largest_accepted_floor_sample`) whose hardcoded exact-equality assertion encoded the old never-decays behavior as correct -- it's the specific behavior this PR changes
- [x] Verified all new/changed assertions fail against pre-fix code via `git stash`
- [x] `uv run pytest tests/test_prefill_transient_tracker.py tests/test_prefill_oom_graceful.py tests/test_scheduler_prefill_memory_guard.py -q` -- 107 passed in this worktree

### DIFF
--- a/omlx/prefill_transient_tracker.py
+++ b/omlx/prefill_transient_tracker.py
@@ -13,6 +13,7 @@ from the global PrefillProgressTracker which feeds the admin dashboard.
 from __future__ import annotations
 
 import logging
+import time
 from dataclasses import dataclass
 
 logger = logging.getLogger(__name__)
@@ -43,6 +44,23 @@ class PrefillTransientTracker:
     # genuinely recurring giant transient still reaches the guard through
     # the last-delta/EWMA terms of _predicted_chunk_transient.
     _OBSERVED_MAX_CLAMP_BYTES = 4 * 1024**3
+    # observed_max_bytes only ever INCREASED before idle-reset existed: one
+    # pathological-but-legitimate (under the clamp) floor-chunk spike early
+    # in a long-running (hours/days, continuously-loaded model) session
+    # permanently raised the admission floor for the rest of that session,
+    # even if the spike never recurred (§B4).
+    #
+    # The first fix tried here decayed the max on every subsequent
+    # floor-size sample. Review caught the flaw: floor samples only occur
+    # once the adaptive throttle is already down at its minimum chunk size
+    # -- the pressure regime -- so that "clock" only ran exactly when the
+    # floor was load-bearing, eroding the protection fastest right when it
+    # mattered most. Forgetting now hangs off wall-clock idle time instead
+    # (``_IDLE_RESET_SECONDS``, checked in ``update``): unrelated to
+    # pressure, so it cannot erode a floor that's actively earning its
+    # keep, and it still clears a stale spike once the session goes quiet
+    # long enough that the spike is no longer representative.
+    _IDLE_RESET_SECONDS = 30 * 60
     # A sample whose per_token exceeds the current EWMA by more than this
     # ratio is treated as measurement noise (a tail/residual prefill chunk,
     # not a genuine cost-per-token regime change) and excluded from the EWMA
@@ -65,6 +83,12 @@ class PrefillTransientTracker:
         # need to allocate that pool again on the next chunk, so the scheduler
         # prices it once until a positive measurement confirms reallocation.
         self._recent_reclaim_bytes: int = 0
+        # Monotonic timestamp of the last recorded sample; None before the
+        # first one. ``update`` resets the tracker when it's been longer
+        # than ``_IDLE_RESET_SECONDS`` since this, so a stale observed-max
+        # from before a long quiet stretch doesn't keep gating admission
+        # once the session resumes (§B4/3.4).
+        self._last_update_monotonic: float | None = None
 
     def _history(self, gathered_core: bool) -> _TransientHistory:
         return self._gathered_history if gathered_core is True else self._dense_history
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for digging into the ratchet. The part about the tracker resetting on a model swap is right, I confirmed the pool drops the engine and the Scheduler builds a fresh tracker.

My concern is the clock the decay runs on. It advances one step per floor sized sample, and floor samples only appear once the throttle is already down at the minimum, which is the pressure regime. A session with room never produces them at all, so the decay does not really run over hours the way the description reads, and under pressure a single long prefill can push hundreds of floor chunks through in a few minutes. The floor ends up eroding in the regime where it is the thing holding admission honest, and the docstring on `_admission_transient_bound` is written around the opposite property.

The other thing is that I could not find a case behind this. Is there a log or a request that got turned away because of a stuck `observed_max`? #3082 is on main now and breaks the bound into terms, so it should show whether the ratchet is the binding one.

If it is, I would rather see the forgetting hang off something unrelated to pressure, like a reset after a long idle, which `reset()` is already documented for but nothing calls. Either way I would want an A/B under pressure showing mid prefill aborts stay at zero before this goes in.

alytaphoenix:
You're right on the clock: I checked, floor_sample is only ever True when the adaptive throttle has already dropped to its minimum chunk size (n_tokens <= min_chunk at the call site) -- that's the pressure regime, not "over hours" the way the docstring read. A session with headroom never produces a floor sample at all.

Replaced the pressure-coupled decay with the alternative you suggested: forgetting now hangs off wall-clock idle time via `PrefillTransientTracker.reset()` (previously documented for "after a long idle" but nothing called it). `update()` tracks a monotonic timestamp of the last sample and calls `reset()` on the next one if more than 30 minutes have passed since. A floor sample within one idle window only ever raises the max again (restoring the pre-existing in-session behavior) -- decay during active pressure is gone entirely, since that was the actual problem.

Confirmed the model-reload path needs no change, matching what you already found: `engine_pool.py` always constructs a fresh `Scheduler` -> fresh tracker on reload/swap.

Replaced the original commit rather than stacking a second one on it (same as I did on #3090 in this same pass), since its message asserted the pressure-coupled decay was correct, which your review disproved.

Test changes: `test_max_tracks_largest_accepted_floor_sample` now asserts the max stays pinned at the highest sample within a session (the exact coupling you flagged), and the old `TestObservedMaxDecay` class is replaced with `TestObservedMaxIdleReset` -- repeated lower floor samples with no elapsed time don't erode the max, a fresh spike still raises it instantly, an idle gap under the threshold changes nothing, and a gap over it resets the whole tracker on the next sample regardless of how many floor samples happen afterward. Full 10,473-test suite passes.

What I couldn't provide from here: the A/B under real pressure showing mid-prefill aborts stay at zero, and whether the ratchet is even the binding constraint (checkable now via #3082's merged per-term breakdown). That needs sustained live traffic against a real loaded model over a meaningful window, which isn't something I can responsibly manufacture in this pass -- I don't have a production traffic pattern to replay, and a short synthetic burst wouldn't actually answer the "is this the binding constraint over hours" question you're asking. If you have a way to run that A/B (or can point me at #3082's breakdown output from a real session), I'm glad to help interpret it or iterate further based on what it shows.

alytaphoenix:
Rebased onto current main (1 real conflict: main independently refactored the tracker into a dual dense/gathered-regime history dataclass since this branch was cut — merged the idle-reset logic to apply across both regimes via the shared `_last_update_monotonic`, updated the stale `observed_max_bytes`/`reset()` docstrings to match). Full suite: 10,666 passed, 0 failed, 150 skipped, 77 deselected.

alytaphoenix:
Rebased onto current main (1 conflict in prefill_transient_tracker.py: main independently added a \`reset_history(*, gathered_core)\` per-regime helper and refactored \`reset()\` to call it twice instead of resetting the two history objects inline — kept this PR's own idle-reset docstring but switched to main's DRY implementation, no semantic change). Full suite: 10,834 passed, 0 failed, 152 skipped, 77 deselected.

alytaphoenix:
Refreshed onto current main (clean rebase, no conflicts — 40 commits of drift picked up). The earlier CI failure was the same unrelated pre-existing flake seen on #3090 (\`test_prefix_cache_gdn_split.py\`, same assertion shape, different specific test) — passes clean locally. Full suite: 11,916 passed, 0 failed, 201 skipped, 77 deselected.

PR #3110, opened 2026-08-24
fix(ane): slice down matrix before dequantizing, not after

alytaphoenix:
## Summary
- `_prepare_fused_down_for_bank` dequantized `down_proj`'s ENTIRE packed weight to a dense fp32 array, but only columns `[0:gpu_start]` were ever consumed from it (`down0`/`down1`, and `cpu_down_weight` when `cpu_hidden>0`) -- the GPU portion reads `down.weight` directly, still quantized, a few lines later. The `[gpu_start:hidden]` suffix was computed and immediately discarded on every fused-down MLP compile: a pure-waste transient of roughly `(hidden - gpu_start) * out_features * 4` bytes (~0.5GB/layer at typical Qwen3.5/3.6 MLP dimensions).
- Part of `docs/qwen35-hardening-and-optimization.md` Theme C, item C3/3.5.

## Fix
Slice `down.weight`/`down.scales`/`down.biases` to `[:gpu_start]` (in packed-axis terms: `gpu_start // 8` for the 4-bit-packed weight, `gpu_start // 128` for the `group_size=128` scales/biases) before calling `mx.dequantize`, instead of after. `gpu_start` is always a multiple of 128 (both `per_ane` and `cpu_hidden` are rounded down to 128 via the existing pattern), so the packed-axis slice lands exactly on quantization group boundaries -- the result is bit-identical to the old dequantize-then-slice behavior.

## Test plan
- [x] New test spying on `mx.dequantize`'s call args: confirms the down matrix's dequantize call now receives only the `gpu_start`-wide packed slice (48 columns in the test's dimensions) instead of the full packed width (128 columns)
- [x] New correctness test: compares `down0`/`down1`/`cpu_down_weight` from the sliced-dequant path against a full-dequant-then-slice reference, bit-identical via `mx.array_equal`
- [x] Verified the shape-spy test fails against pre-fix code (`128 == 48` assertion failure) via `git stash`; the correctness test passes both ways as expected (proves the optimization is value-preserving, not just faster)
- [x] `uv run pytest tests/test_qwen35_ane_prefill.py -q` -- 110 passed in this worktree

### DIFF
--- a/omlx/patches/qwen35_ane_prefill.py
+++ b/omlx/patches/qwen35_ane_prefill.py
@@ -2746,10 +2746,21 @@ def _prepare_fused_down_for_bank(
             ).astype(mx.float32)
         )
 
+    # Only columns [0:gpu_start] of the down matrix are ever consumed below
+    # (down0/down1/cpu_down_weight); the GPU portion reads down.weight
+    # directly, still quantized, at gate_up_weight/down_weight below.
+    # Dequantizing the full matrix wasted a ~(hidden - gpu_start)-column
+    # fp32 transient (~0.5GB/layer) that was computed and immediately
+    # discarded. gpu_start is a multiple of 128 (per_ane/cpu_hidden are
+    # both rounded down to 128 above), so slicing the packed axis at
+    # gpu_start // 8 (4-bit: 8 values/int32) and gpu_start // 128 (the
+    # quantization group_size) yields exactly the first gpu_start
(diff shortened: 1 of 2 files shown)


jundot:
Merging, thanks. Slicing in the packed domain mirrors how the same function already handles the GPU suffix, and the bit-exact parity test covers the boundary I would have asked about.

PR #3111, opened 2026-08-24
docs: propose cluster v2 universal distributed inference RFC

jonathan308:
## Summary

- define a capability-driven two-Mac tensor-parallel architecture for cluster v2
- separate validated Fusion evidence from proposed upstream behavior and future work
- specify lifecycle, in-flight cancellation, verified teardown, identity/pairing, signed planning, content-addressed staging, JACCL/RDMA selection and fallback, cache/concurrency semantics, telemetry, and onboarding
- publish precise Qwen and DeepSeek V4 Flash reference measurements alongside explicit optimization targets
- qualify Qwen3.8 unified checkpoints as text-backbone-only distributed deployments while requiring image, video, and audio inputs to fail closed
- break the implementation into bounded dependent PR stages after #3106 and #3107

## Evidence boundary

This is a documentation-only RFC. It does not claim that the Fusion reference implementation is present in upstream main. The benchmark section labels API versus engine-marker timing, cold versus cached work, measured versus target values, and unqualified paths.

The current Qwen3.8-27B qualification is intentionally text-only: 535.73 API / 536.17 marker tok/s cold prefill at 30K, 46.45 tok/s B1 decode, and 124.64 aggregate B4 decode. Greedy no-thinking parity was 4 of 6 token-exact and 2 of 6 semantically equivalent. Media requests now receive an actionable HTTP 400 instead of having content silently discarded; multimodal tensor sharding remains future work.

The DS4 Fusion reference remains 852.29 API / 858.04 marker tok/s at cold 30K, 780.77 / 782.89 at cold 100K, 31.95 non-MTP decode, and 72.35 to 72.56 MTP depth-5 decode.

## Validation

- markdownlint-cli2: 0 issues
- git diff --check: clean
- all three upstream CI lanes are green
- no runtime code changed

## Review focus

- Beta 1 support boundary and fail-closed behavior
- plan signing and node identity model
- lifecycle/cancellation invariants
- staged upstream PR order
- numerical parity policy when collective reduction order is not token-exact

### DIFF
--- a/docs/cluster-v2-universal-distributed-inference-rfc.md
+++ b/docs/cluster-v2-universal-distributed-inference-rfc.md
@@ -1,0 +1,547 @@
+# RFC: cluster v2 universal distributed inference
+
+Status: proposed
+
+Date: 2026-08-24
+
+Scope: Apple Silicon text-generation clusters; documentation only
+
+## Summary
+
+Cluster v2 should make distributed inference an oMLX runtime capability rather
+than a collection of model-specific launch scripts. A coordinator admits a
+known model, stages its immutable artifacts, produces a signed and
+content-addressed execution plan, starts one rank per selected Mac, and exposes
+that deployment through the existing oMLX API. Every rank must agree on the
+model, plan, backend, and runtime epoch before the deployment becomes ready.
+
+The first beta is intentionally narrower than the eventual design. It targets
+two Apple Silicon Macs, qualified text-only model families, tensor parallelism,
+and JACCL/RDMA with an explicitly planned fallback. It does not claim support
+for arbitrary MLX models, multimodal models, live membership changes, or mixed
+Metal/CUDA execution.
+
+This RFC separates evidence from aspiration. The reference implementation has
+been exercised with DeepSeek V4 Flash and Qwen3-30B-A3B-4bit, including Qwen
+staging, cache reuse, cancellation, concurrent batching, per-request metrics,
+and lifecycle reload. Those capabilities are not in upstream `main` merely
+because they appear in this document. The proposed PR series below moves them
+upstream in reviewable layers.
+
+## Motivation
+
+An inference engine is only generally useful when its control plane is as
+reliable as its kernels. A fast two-rank run does not establish that operators
+can safely discover peers, reproduce a plan, stage a model, cancel work, unload
+every rank, or tell a live request from stale dashboard state. Conversely, a
+polished setup flow does not establish that a model's tensor layout is safe to
+shard.
+
+Cluster v2 joins those concerns through explicit contracts:
+
+- a model capability contract decides whether and how a checkpoint may be
+  sharded;
+- a deterministic plan binds the model identity, rank placement, backend, and
+  cache geometry;
+- a lifecycle state machine makes readiness, cancellation, and teardown
+  observable and idempotent;
+- authenticated pairing separates discovery from trust;
+- content-addressed staging removes the same-absolute-path requirement;
+- backend negotiation prefers verified RDMA without silently changing an
+  active deployment; and
+- scheduler and dashboard telemetry use the same request and runtime records.
+
+The design should work for many models, but universal means capability-driven
+admission and a safe refusal path. It does not mean optimistically sharding
+every checkpoint.
+
+## Goals
+
+1. Run a qualified text model across two unequal Apple Silicon Macs through
+   the normal oMLX API.
+2. Keep rank-local weights and KV state within an admitted memory budget.
+3. Give every rank an identical, authenticated execution plan while allowing
+   different local model paths.
+4. Prefer a measured JACCL/RDMA route and support an explicit, re-planned
+   fallback when RDMA is unavailable.
+5. Make load, ready, drain, unload, failure, and cancellation states accurate
+   after client disconnects, server restarts, and partial rank failures.
+6. Preserve prompt-cache reuse, tiered-cache policy, continuous batching, and
+   per-request metrics in distributed mode.
+7. Add new model families through small compatibility adapters and tests,
+   without adding model names to the scheduler or UI.
+8. Land the work as a sequence of bounded PRs with feature gates and fallback
+   behavior, not as one product-and-engine change.
+
+## Non-goals for the first beta
+
+- Sharding vision towers, audio towers, cross-modal projectors, or other VLM
+  components.
+- Dynamic rank membership or re-sharding a live request.
+- Continuing a request after a rank is lost.
+- Pipeline parallelism, expert parallelism, or more than two ranks as a
+  release claim.
+- Mixed Metal/CUDA execution or a Ring/NCCL gateway.
+- Treating CPU and ANE capacity as interchangeable with Metal tensor shards.
+- Claiming linear or exponential speedup for every prompt, model, or phase.
+- Automatically falling back to a local full-model load after a distributed
+  request has been admitted.
+
+These remain valid follow-up directions. They need their own correctness and
+hardware gates.
+
+## Current evidence
+
+### Evidence boundary
+
+The measurements below came from the Fusion reference branch, not upstream
+`main`. They are point measurements on one two-Mac topology and are not product
+guarantees. `API` is end-to-end timing observed by the client. `marker` is the
+coordinator's engine marker. Cold prefill rows had zero cached prompt tokens.
+Small API/marker differences are expected because their timing boundaries are
+different.
+
+Hardware and transport:
+
+- rank 0: M3 Ultra Mac;
+- rank 1: M5 Max Mac;
+- two-rank JACCL over a directly connected Thunderbolt RDMA path;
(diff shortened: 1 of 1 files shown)


jonathan308:
Fusion reference update (2026-08-24):

- published `0ddc954b`: distributed DeepSeek ANE now fails closed unless an operator explicitly enables experimental testing; 4096-tile TP regressed because long prompts use 1024-token turns, and a 1024-tile test timed out JACCL after 30,001 ms
- published `441cb51a`: exact-match heterogeneous TP qualification API/store plus cluster-v2 provenance UI
- qualification keys now include model identity, ordered node/runtime/kernel/fabric fingerprints, context bucket, execution profile and MTP graph
- promotion requires >=2 matched samples, identical parity hashes, a profile-specific gain, valid adapter shard units and memory fit; all drift falls back to equal shards
- the wizard now renders real signed proportions (for example 3/8 and 5/8), not a misleading equal 1/N label
- full cluster/UI regression gate: 1,674 passed
- restored DS4 exact-GPU TP2 after both-Mac reboot: 834.63 API / 839.95 marker tok/s at cold 30K and 766.62 / 768.71 at 100K (8.1% taper); decode 30.50 and 28.69 API tok/s

Beta 1 boundary remains planned scaling, not live elastic resharding: a newly added node triggers a signed replan/reload. Remaining blockers are automatic peer reconnect when mDNS is unavailable, a one-action matched qualification runner in the wizard, broader model/soak gates, and a signed/notarized DMG.

jonathan308:
Further Fusion reference update (2026-08-24):

- automatic paired-peer reconnect is now physically proven in both directions without manual IP seeding; discovery uses direct LAN when allowed and the signed macOS Tailscale app for control fallback, while JACCL remains on Thunderbolt RDMA
- verified peers heartbeat every 2 seconds; unknown candidates retain a 10-second scan, eliminating discovered/suspect UI flapping
- both Fusion launch agents are now RunAtLoad=true; restarting the M5 control server repeatedly did not disturb its separate live inference rank
- matched DS4 outer-layout qualification found 3:5 was 15.15% faster at 30K but only 1/2 prompts token-exact; qualification ID `02370f24...` is persistently rejected and autoconfigure returns equal 4:4 plus the exact rejection reason
- non-MoE 3:5 / routed-MoE 4:4 preserved parity but improved prefill only 0.22% and regressed decode 0.34%, so all component overrides were removed
- an equal-width M5 NAX routed-MoE microgate was exact and 1.175x composed, but same-binary full TP2 was +0.09% prefill and about -0.5% decode; it remains default-off
- active Beta-safe DS4 deployment is equal 4:4, 80.03 GB/rank, 1M context reservation, zero failed requests

This is the intended RFC policy in action: measured recommendations may trigger calibration, but only exact-key full-model parity evidence can alter signed shard arithmetic. Faster divergent layouts are recorded as negative evidence so the wizard does not recommend them again.

jonathan308:
Hybrid topology checkpoint (2026-08-24):

- Fusion commit `8976b6f4` adds capability-gated TP×pipeline planning/execution: valid factor enumeration, orthogonal rank maps, stage-first progressive loading, subgroup-routed pipeline collectives, signed-plan/dashboard support, and fail-closed activation when the loaded backend cannot split groups.
- Stock MLX Ring and JACCL both expose `Group.split()` but throw at runtime. MLX fork commit `32629b987` (`jonathan308/mlx:feat/jaccl-subgroups`) adds logical JACCL subgroups over the existing RDMA mesh and an explicit capability API.
- Physical M3 Ultra↔M5 Max JACCL/RDMA gate passed int32/BF16 reductions, min/max, gather, sum-scatter, direct send/recv, and singleton groups. This is not yet a four-device hybrid certificate; Automatic remains hidden on stock runtimes and the release gate still requires a real TP2×PP2 run.
- Full oMLX cluster/distributed/UI regression: 1,754 passed. Production DS4 remains the parity-safe equal TP2 deployment; no experimental subgroup wheel was installed into it.

This should remain a dependent runtime/transport slice in the upstream series, not be presented as already-general elastic scaling.

jonathan308:
The full upstream-rc3 implementation is now available in #3118. That feature PR carries the complete Cluster v2 flow, physical TP2/RDMA evidence, lifecycle/cache/cancel/concurrency gates, and the current Beta boundaries; this RFC remains the architecture discussion reference.

borodach23:

DeepSeek V4.1 Flash has no distributed path in the tree, and separately, a
4-node cluster has no expressible plan shape that would use one even if it
existed. I run 4x Mac Studio M3 Ultra 512 GB on a full Thunderbolt 5 mesh
and can only use one of them for this model.

Current state on main @ `5941f68` (0.7.0.dev3):

1. `omlx/patches/deepseek_v41/` defines no `shard()` and no `pipeline()`, and
   never references `mx.distributed`, `all_reduce` or `all_gather`. V4.0 does:
   `omlx/patches/deepseek_v4/deepseek_v4_model.py:2432`.
2. `omlx/cluster/` never references `deepseek_v41`. The type is also absent
   from the list of oMLX-injected architectures in `inference_worker.py`
   (`glm_moe_dsa`, `deepseek_v4`, `minimax_m3`, `step3p7`, `llama4`), so a rank
   could not resolve it even if a plan were signed.
3. `omlx/cluster/autoconfigure.py:76` restricts candidates to all-nodes-TP or
   pipeline-only, so a 4-node world has no 2x2 option, although
   `omlx/cluster/planner.py:1937` already models the 2D world
   (`pipeline_stages * tensor_parallel_size == len(nodes)`).
4. `docs/distributed-cluster.md:5` and `:308`, and the cluster v2 RFC in #3111,
   are scoped to two Macs.
5. #3107 (DeepSeek V4 tensor-parallel planning) is still a draft, last updated
   2026-08-24.

Three items, in dependency order.

1. `shard()` for `deepseek_v41`, with Engram row sharding, plus the capability
   entry so the cluster catalogue stops classifying the model as
   non-distributable.
2. Either expose the hybrid `tensor_parallel_size x pipeline_stages` plan that
   `planner.py` already models, or state that TP=N in one stage is the intended
   shape for worlds larger than two. As it stands, a 4-node owner has no
   expressible plan for any model that only shards evenly at 2.
3. Unfreeze #3107. Its planner work (asymmetric TP shard weights, replicated KV
   heads kept out of divisor checks, sparse pooling and indexer KV accounting)
   is a prerequisite for both DeepSeek generations.

Why V4.1 looks like a particularly good tensor-parallel candidate:

`Jundot/DeepSeek-V4.1-Flash-oQ4e-mtp` is 402.4 GiB, of which 279.5 GiB is the
backbone excluding Engram, 114.4 GiB the Engram tables, 7.5 GiB DSpark MTP and
0.9 GiB vision. Two consequences:

- The Engram tables are embedding-style lookups and are over a quarter of the
  checkpoint. Row sharding across ranks is a natural low-traffic split, since
  only the owning rank contributes to a given lookup.
- At TP=4 the per-node resident set drops to roughly 100 GiB. On 512 GB nodes
  that leaves room for the 1M context the checkpoint config declares, which
  does not fit alongside the weights on a single node today.

Collective traffic for an MLA decode step is small, on the order of tens of KB
per token per layer, so TP decode should be latency-bound on the all-reduce
rather than bandwidth-bound. On a TB5 mesh that looks favourable, but I have
not measured it, which is the point of the offer below.

- Four independent single-node replicas behind a load balancer. This works
  today and is what I will do in the meantime: oQ4e fits one node with Engram
  resident (`deepseek_v41_engram_ssd_offload` off), `deepseek_v41_ced_prefill_enabled`
  on, DSpark MTP on, `iogpu.wired_limit_mb` raised. It gives aggregate
  throughput but no single-request speedup and no headroom for long context.
- MoE expert offload. Not applicable here: it exists for memory-constrained
  nodes, costs latency, and is mutually exclusive with Lightning MTP and DSpark
  per `docs/MoE_Expert_Offload.md`.
- Tensor-parallel V4.0 Flash on two nodes instead. Possible in principle since
  V4.0 implements `shard()`, but the DS4 planner path is still unmerged in
  #3107, so capability detection and memory accounting are incomplete.
- Running V4.1 on one node and smaller models on the rest. Works, but leaves
  1.5 TB of unified memory idle for the workload that actually needs it.

**Additional context**

Hardware I can contribute as a test platform: 4x Mac Studio M3 Ultra, 512 GB
each, full Thunderbolt 5 mesh (3 links per node), so every edge satisfies the
`_FAST_TRANSPORTS` gate at `autoconfigure.py:35`. This is a rare enough
topology that I suspect no one has exercised the planner against it.

I am happy to run whichever harness you name, existing
(`benchmarks/deepseek_v41_offload_bench.py`, the cluster strategy benchmarks)
or new, and report medians of repeated runs with the base commit measured
alongside the change, in the same before/after form used in the #3590 review.
I can also hold a fixed baseline commit for regression checks, and test TP=4
on V4.0 Flash and on Qwen families once #3107 is unblocked.

Environment: macOS 26.x, oMLX 0.7.0.dev3 source build, MLX 0.32.2.

Issue #2225, opened 2026-07-13
FA-256 steel prefill kernel: pathological high-kv_len slowdown on pre-NAX GPUs (M3 Max) — IOGPU command-buffer preemption (mlx#3302 class); mlx#3307 fix not applied to omlx's custom kernel


zhongyiyang1981:
## Summary

omlx's custom **FA-256 steel prefill attention kernel** (`qwen35_fa256_attention` in `omlx/custom_kernels/qwen35_prefill`) degrades to a pathological slowdown at high `kv_len` on **pre-NAX GPUs (M1–M4)** and never recovers, making long-context prefill unusable when the patch is enabled. This is the **same root cause as mlx#3302** (steel attention dispatches all keys in a single Metal compute command; per-dispatch wallclock is linear in `kL` and eventually crosses the macOS `kIOGPUCommandBufferCallbackErrorImpactingInteractivity` threshold). That flaw was **fixed in MLX core by mlx#3307 (chunked dispatch)**, but **omlx's custom FA-256 kernel is separate native code and did not receive the equivalent fix** — so stock SDPA (mlx-core, fixed) stays smooth while omlx's kernel falls off a cliff.

This appears to be the gap left by #1387, which tracked the mechanism but closed on the **mlx-core** path (the reporter's bisection found the offender in `mx.fast.scaled_dot_product_attention` with TurboQuant off). The **omlx custom-kernel path** still hits the same wall on pre-NAX hardware and, as far as I can tell, is not tracked by any open issue.

## Environment

- Machine: Apple **M3 Max** (40-core GPU), 96 GB unified memory
- macOS: 26.5.1 (Darwin 25.5.0)
- omlx: **0.5.1** (commit `e3a4fe46`), installed via `brew install jundot/omlx/omlx --HEAD --with-custom-kernel`
- MLX: 0.32.0 (the version 0.5.1 pins — per #2138), which **includes** the mlx#3307 fix
- Models: Qwen3.6-**27B** (standard softmax attention, head_dim=256) and Qwen3.6-**35B-A3B** (GatedDeltaNet, 16/2 GQA)
- Custom kernels confirmed loaded: `/api/status` → `custom_kernels.qwen35_prefill.available=true`

## Reproduction

1. Force the FA-256 patch on: `OMLX_FA256_STEEL=1` (default is already on for pre-NAX; this just makes it explicit).
[...]

PR #3122, opened 2026-08-25
feat(fa256): streaming online-softmax chunk fold (§B1)

alytaphoenix:
## Summary

Replaces the n-chunks-scaled partial slab in the chunked FA-256 attention kernel with an online-softmax streaming accumulator, so the per-op transient becomes O(1) in n_chunks instead of the ~2GiB coexisting slab (capped by `n_mem_cap`). This is item §B1 of the qwen35 hardening review — the highest-leverage capacity item identified there.

- Each chunk's normalized partial is folded into a persistent fp32 running output + logsumexp via a new `attention_chunk_fold` kernel (additive in the shared `steel_attention_block_token.h` — the existing attention kernel body and `attention_chunk_reduce` are untouched).
- Single in-place fp32 accumulator (`acc`, per-element single-owner) + **ping-pong `lse_run` buffers**, with the per-row logsumexp write gated to `d4==0`. This avoids a within-dispatch cross-thread read-modify-write race that a naive single-buffer `lse_run` design has at `bd=256` (two SIMD-groups per row share the buffer with no ordering guarantee). Cross-dispatch hazards (fold *c* writes read by fold *c+1*; chunk *c+1* writes `o_slot` read by fold *c*) are covered by MLX's existing per-buffer barrier tracking (verified against `mlx` v0.32.0 `device.cpp`: `register_output_array` sets `needs_barrier_` on `prev_inputs_` membership, `maybeInsertBarrier()` runs at every dispatch start).
- Gated behind `stream_fold` (constructor arg → `is_equivalent`/`state`), a `fast.qwen35_fa256_attention(..., stream_fold=...)` kwarg, a `FA256_HAS_STREAM_FOLD` capability probe (older extensions degrade cleanly), and `OMLX_QWEN35_FA256_STREAM_FOLD`-style env override in the patch layer — **default ON** where the rebuilt extension supports it.
- Streaming also lifts `n_mem_cap`, so the per-dispatch work budget is honored exactly at every `qL` (removing a latent re-exposure of the #2225 IOGPU-preemption failure mode that the memory cap could previously override).

## Measurements (M4 Pro, Apple Silicon)

- **Numerics**: streaming vs. the legacy batched fold matches to ≤4.88e-4 (bf16) / ≤1.22e-4 (fp16) max-abs-diff — near bit-identical (same partials, same weights; only fp32 association order differs). Streaming adds **zero** additional error vs. the fp32 `mx.fast.scaled_dot_product_attention` reference compared to the legacy path (identical to printed precision across every tested shape, including the causally-dead-chunk case). No NaNs anywhere in the matrix (n_chunks ∈ {2,8,48}, aligned/unaligned/square shapes, both dtypes).
- **Memory**: per-op transient drops **2079.8 MB → 193.1 MB (10.8×)** at qL=4096, and is n_chunks-independent (same figure at kL=131072 and kL=262144) — matching the design's ~150MB prediction.
- **Overhead**: the extra ~n_chunks tiny fold dispatches cost ~2–8% wall-clock at matched n_chunks (below the legacy slab cap, so both paths issue identical attention dispatches) — small relative to the capacity win, which is why the default flips ON.
- **Live validation**: deployed to a 2-node cluster running Qwen3.8-27B; served a 50,726-token prompt end-to-end through the native streaming path with no regression, and a live per-token prefill-transient measurement dropped 11720 KB/token (legacy fallback path) → 1866 KB/token (native + streaming), ~6.3×.
- **KV-bitdepth capacity**: a prior review flagged that 4-bit vs 8-bit KV only bought +16% admittable context, implicating this slab as the masking cause. Re-measured post-fix: the admission-relevant transient is now flat and KV storage is the sole context-length-dependent term, so 4-bit's capacity benefit over 8-bit is empirically **+98%** (kv-bytes-per-token ratio ≈1.985×) — consistent with the original hypothesis.

## Test plan

- [x] `tests/test_qwen35_fa256_attention.py` — added a stream-vs-legacy-fold parity matrix (5 shapes × 2 dtypes, including the causally-dead-chunk case), an `n_chunks<=1` no-op regression test, and capability-probe/env-override tests. 28 tests, all pass in this branch (12 pass + 16 correctly skip without a built native extension in CI's default environment).
- [x] Numerics offline against the rebuilt native extension (see measurements above).
- [x] Shared-header co-owner (`glm_moe_dsa`, which also compiles `steel_attention_block_token.h`) test suite — 51 passed, unaffected (the fold kernel is additive-only).
- [x] Peak-memory A/B, fold-dispatch-overhead benchmark, and one real long-context production request — all above.

### DIFF
--- a/omlx/custom_kernels/common/csrc/kernels/steel_attention_block_token.h
+++ b/omlx/custom_kernels/common/csrc/kernels/steel_attention_block_token.h
@@ -625,3 +625,116 @@ template <typename T>
   dst[2] = static_cast<T>(out.z);
   dst[3] = static_cast<T>(out.w);
 }
+
+///////////////////////////////////////////////////////////////////////////////
+// Streaming chunk fold kernel (Phase 3.1 / §B1)
+///////////////////////////////////////////////////////////////////////////////
+
+// Online-softmax fold of ONE chunk's normalized partial (o_slot / lse_slot,
+// the same output the attention kernel emits with output_partials) into a
+// persistent fp32 running accumulator (acc) plus a running logsumexp
+// (lse_run_prev -> lse_run_next). Replaces the n_chunks-scaled partial slab
+// with a single chunk slot, so the running transient is O(1) in n_chunks. lse
+// values are in the attention kernel's scaled log2 domain (exp2/log2
+// throughout; the scale cancels in the differences).
+//
+// Threading mirrors attention_chunk_reduce: one thread per 4 head-dim elements
+// of one (b, h, row). acc[..., 4*d4 .. +3] is single-owner (safe to RMW in
+// place). The per-row lse is written only by the d4 == 0 thread and to a
+// DISTINCT ping-pong buffer (lse_run_next), so the bd/4 threads of a row —
+// which span two SIMD-groups at bd=256 — never race on it (they all read the
+// old value from lse_run_prev). Cross-dispatch hazards (fold c writes
+// acc/lse_run read by fold c+1; chunk c+1 writes o_slot read by fold c) are
+// covered by MLX's per-buffer barrier tracking; every chunk c>=1's attention
+// dispatch reliably inserts a global buffer barrier (its o_slot write is a WAR
+// vs. fold c-1's read), which also flushes the acc/lse_run writes.
+template <typename T>
+[[kernel]] void attention_chunk_fold(
+    const device T* o_slot [[buffer(0)]],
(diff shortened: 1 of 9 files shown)


alytaphoenix:
The `test (3.11)` leg was `CANCELLED` (not a real failure) on the last CI run, 10 days stale now. Rebased onto current main (clean, no conflicts) to get a fresh CI run. Full suite: 10,663 passed, 0 failed, 161 skipped, 77 deselected.

PR #3127, opened 2026-08-25
perf+exactness(turboquant): roll quantized state directly in finalize() (§F4)

alytaphoenix:
## Summary

`BatchTurboQuantKVCache.finalize()` converted right-padding to left-padding by dequantizing keys/values to fp16, rolling with `dynamic_roll`, and requantizing — materializing the full fp16 tensors and re-rounding an already-quantized value through the codec a second time.

TurboQuant's `dequantize`/`quantize` are pure per-token operations (bit-unpack, fixed codebook lookup, fixed Hadamard rotation — no cross-token state, verified against the `mlx_vlm.turboquant` source), so a token-axis reindex commutes exactly with dequantize: `dequantize(roll(state)) == roll(dequantize(state))` bitwise. Roll only permutes which token's packed bytes live at which position; it never touches the bytes.

Added `_roll_state`, a recursive dispatcher over every TurboQuant state variant (mirrors this file's existing `_slice_state`/`_concat_state`/`_filter_state` pattern), applying `dynamic_roll` directly to each field that shares the token axis (norms, packed indices/radii, ...) with the same per-row shift. `finalize()` now calls it directly on `self.keys`/`self.values` instead of dequantize → `dynamic_roll` → requantize, skipping both the fp16 materialization and the second rounding pass.

## Test plan

- [x] Added parametrized (`bits=1.0/4.0/8.0`) regression tests asserting `dequantize(finalize(state)) == dynamic_roll(dequantize(state))` **exactly** (`mx.array_equal`, not a tolerance check — this is the point of the change), plus a no-op-when-unpadded test.
- [x] `tests/test_turboquant.py` — 56 tests, all pass.

### DIFF
--- a/omlx/turboquant_kv.py
+++ b/omlx/turboquant_kv.py
@@ -210,6 +210,54 @@ def _pad_state_left(state, pad_length: int):
     return _concat_state(pad, state)
 
 
+def _roll_state(state, shifts, axis: int = 2):
+    """Roll a quantized state along its token axis by per-row ``shifts``.
+
+    Mirrors dynamic_roll's per-batch-row dynamic shift (same axis/shifts
+    convention as _slice_state/_concat_state above), applied directly to the
+    packed representation instead of dequantize -> roll -> requantize: roll
+    is pure reindexing along the token axis, so applying it independently to
+    every field that shares that axis (norms, packed indices, radii, ...)
+    preserves their per-token pairing exactly, with no fp16 materialization
+    and no re-rounding through the codec on the way back in.
+    """
+    if state is None:
+        return None
+    if isinstance(state, TurboQuantMSEState):
+        return TurboQuantMSEState(
+            dynamic_roll(state.norms, shifts, axis=axis),
+            dynamic_roll(state.indices, shifts, axis=axis),
+        )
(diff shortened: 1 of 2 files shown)


xxh422735676:
Is this helpful to reduce memory spike on prefilling while enabling turbo quant?

alytaphoenix:
Partly — it depends which spike you mean.

This change is in `BatchTurboQuantKVCache.finalize()`, which converts right-padding to left-padding at the end of batched generation. It previously dequantized the whole K/V cache to fp16, rolled it, and requantized. Rolling the packed state directly removes that full-fp16 materialization, so it does eliminate a real transient — but it's at **finalize** time, and only on the batched path.

It does **not** reduce the prefill attention spike (the score-matrix transient) that "memory spike on prefilling" usually refers to — that lives in the attention kernel, not the cache codec. If that's what you're hitting, #3108 (registering TurboQuant's tiled prefill with the memory guard) and #2991 (sdpa256 q-split headroom) are the relevant ones.

The other benefit here is exactness: the old path re-rounded already-quantized values through the codec a second time; this one is bitwise identical to rolling the dequantized state.

PR #3128, opened 2026-08-25
fix(admin): dashboard data-loss and dead-interaction bugs (Phase 1)

alytaphoenix:
## Summary

Eight HIGH/MEDIUM correctness bugs in the admin dashboard, found during a UI/UX audit and fixed as a risk-ordered batch (each is independently small; grouped here because they share the same audit and review context).

- **Unclosed `<button>` tag** in the Settings > Models copy-name button (missing `>` before the SVG) made the button an invisible click target — the copy icon never rendered.

- **`showNotification` was called 12 times but never defined.** Every call in the cluster SSH-pairing flow (generate key / exchange token / store in keychain) threw a `TypeError`, and — with no `finally` block — left the clicked action's busy flag stuck forever, wedging the button until page reload. Implemented a small toast queue (`notifications: []`, `showNotification`/`dismissNotification`, rendered via a fixed bottom-right `x-for`) and added the missing `finally` blocks.

- **Chat "Delete message" silently destroys every later turn.** It's a shared delete/regenerate-edit helper that truncates the whole tail, but was labeled plain "Delete" with no confirmation — deleting the first reply of a 50-turn chat silently destroys the other 49. Relabeled to "Delete from here" and gated behind a `confirm()` that describes the actual scope.

- **Cluster worker roles silently revert to Headless within 10 seconds.** `syncClusterNodesFromPeers()` hardcoded `role: 'headless'` for worker nodes while the local node correctly preserved its role — a user-assigned Workstation role reverted on the very next poll. Now preserves `previous.role` symmetrically.

- **The cluster tab wipes the model inventory + fit catalogue every 10s**, even when nothing changed — forcing an SSH refetch of every worker's models on every poll and flashing the status pill green "Ready" during the null window, even for a model that doesn't fit. Moved the null into the existing `nodesChanged`-gated invalidation block instead of running it unconditionally.

- **Re-running an accuracy benchmark silently discards the new result.** The SSE `result` handler deduped on `(model_id, benchmark)` and dropped the event on a match — but that pair isn't unique across runs, so re-running the same model+benchmark (the normal "changed a setting, run again" flow) silently kept the stale card. Replace-in-place instead of drop: idempotent on a true SSE replay, correct on a genuine re-run, and keeps the follow-up `upload` event's `findIndex` pointed at the current card.

- **Opening Settings silently rewrites the persisted SSD cache size.** `loadGlobalSettings()` computed a percent from the server's `ssd_cache_max_size` string, then immediately wrote a percent-derived approximation *back* into it — silently replacing e.g. `auto` or an exact value with a whole-percent-rounded figure, persisted the next time the user saved any unrelated setting. Removed the write-back; the slider's initial position is still computed for display, but the persisted field stays exactly what the server sent until the user actually moves the slider or edits the GB input.

- **The idle-timeout dropdown posts the entire settings form on `@change`.** Changing it silently committed — or, on a validation failure, silently discarded via the reload — every other unsaved edit on the page. Added a narrow `{idle_timeout_seconds}` patch mirroring the existing `saveCtxBenchPriority` pattern (optimistic update, revert + toast on failure).

## A build-step note for reviewers

`static/css/tailwind.css` is a committed, pre-built artifact (no build-on-serve pipeline in this repo). The new toast container needed `bottom-4`/`right-4`/`z-[70]`/`w-80`, which weren't in the existing build, so it's rebuilt here too — pinned to the exact `tailwindcss@3.4.17` that produced the committed file (a newer patch version reformats/reorders the *entire* file harmlessly, which would make this diff much larger and harder to review). Verified via a structural rule-diff (parse `selector{body}` pairs, diff the sets — raw text diff is useless on a one-line minified file): 6 selectors added, 0 removed, the rest of the file's declarations are unchanged.

## Test plan

- [x] Live-tested in a browser (isolated scratch server instance, Playwright): logged in, fired all three toast levels and confirmed dismiss end-to-end — this is what caught the missing-CSS-classes issue above (the toast rendered at `(0,0)` with `z-index:auto` before the CSS rebuild, at the correct bottom-right position with `z-index:70` after); confirmed zero auto-save POSTs fire just from loading Settings; confirmed the idle-timeout change fires exactly `{"idle_timeout_seconds":900}` and nothing else; swept Dashboard/Settings/Cluster/Chat for console errors.
- [x] 384 admin-related pytest tests pass.
- [x] `node --check` on `dashboard.js`; Jinja parse check on all touched templates; JSON validity on `en.json`.


alytaphoenix:
CI was failing on all 3 Python versions: `test_locale_key_sets_identical` caught `chat.confirm_delete_message` missing from all 8 non-English locales (only `en.json` had it). Pushed a fix adding the key with a real translation to each, matching each file's existing style. Full suite green: 10,322 passed, 0 failed, 102 skipped, 77 deselected.

PR #3132, opened 2026-08-25
fix(admin+mac): form correctness and interaction polish (UI/UX Phase 4)

alytaphoenix:
## Summary

**Stacks on #3128 (Phase 1), #3130 (Phase 2), and #3131 (Phase 3) — all still open.** This branch cherry-picks all three ahead of its own commit so it builds and tests cleanly against `main` standalone; the diff will shrink to just the fourth commit once the earlier three merge.

Thirteen fixes from a UI/UX hardening pass, themed around form-state correctness and small interaction papercuts:

**Model Settings modal (dashboard):**
- Escape handler + a snapshot-based dirty-check now guard every close path (backdrop, header X, footer Cancel) — closing used to silently discard unsaved edits.
- `openModelSettings` gained a request-sequence guard (threaded into the two data-loading functions it awaits, not just its own final assignment — those functions' internal state writes needed the same guard) plus a per-row loading spinner, fixing a race where opening one model's settings quickly after another could show a mix of both.
- `Number.isFinite` instead of `|| null` for three numeric fields, which was silently coercing an explicit `0` to "unset."
- The profile-drift check now catches form-only field changes (previously invisible to it), and saving now tells you when the only effect is the backend silently rewriting your active profile to absorb new fields — rather than the generic "unsaved changes" indicator implying you'd unlink from it.
- The MTPLX side-car import button now keys off a real backend boolean instead of pattern-matching English text in the reason string — also fixes a crash when no model is selected.

**Dashboard, other:** trending/created/updated search sorts no longer get silently re-sorted by download count client-side (the server's own order is now preserved); stats/log polling now use request-id guards so a slow response can't paint data for the wrong model or file; log auto-refresh is a genuine independent on/off toggle instead of only pausable by zeroing the interval; a small shared helper dedupes the one truly-duplicated bit of chat.html's two stats-polling entry points; four smaller papercuts (a benchmark-report crash on an unmeasured value, a stuck "running" indicator, a completion-percentage value mislabeled as tokens/sec, a misleading "0/0 uploaded" row).

**macOS app:** a port-validation bug where malformed port text silently skipped validation entirely instead of being caught; the Initial Cache Blocks field can now actually be cleared (explicit-null patch, matching the existing pattern for idle timeout); and an offline-Apply fix — the Server screen no longer shows a false "pending changes" state after loading with the server down, non-endpoint changes now fail fast instead of being silently dropped when offline, and a port-only change while offline persists locally without the surprising side effect of launching the server.

Also backfills 3 new i18n keys (English text) into the 8 non-English locale files, same as Phase 3.

## Test plan

- [x] Full Python suite: 10404 passed (against the full working branch)
- [x] Targeted slice (812 tests) passing against this PR's actual `origin/main`-based commit stack
- [x] Live Playwright smoke test: drift-check logic (`_profileDrift`/`_profileDiverged`) unit-verified in-browser against the exact symmetric-key/null-skip semantics described above, log auto-refresh toggle persistence confirmed independent of the interval value, zero new console errors
- [x] `xcodebuild build` — BUILD SUCCEEDED against this PR's `origin/main` base
- [x] `xcodebuild test` — 290/290 passing


alytaphoenix:
Rebased onto current main (4 commits, 2 real conflicts in routes.py + Swift):

- routes.py: main independently added qwen4_exp Lightning MTP special-casing to \`_mtp_compat_for_model\` (a different feature landed after this branch was cut) while this PR upgraded the function's return signature from a 2-tuple to a 3-tuple (adds \`mtplx_sidecar_available\` for the import-button UI signal). Merged both — the qwen4_exp branches now also return 3-tuples. Found and fixed two test files with stale 2-tuple unpacking as a result (one from the qwen4_exp work, one from this PR's own earlier test additions) — 9 real failures caught by the full suite, not by the merge itself.
- ServerScreen.swift: main didn't yet have the padding/HintFooter addition this PR adds — purely additive, no real conflict.

Full suite: 10,828 passed, 0 failed, 152 skipped, 77 deselected.

Issue #3117, opened 2026-08-24
Qwen3.8 with ANE enabled produces 0 tokens


arkham000:
**Describe the bug**
Hi, it recently occurred to many several times that the model just produces 0 tokens.
I noticed it since 0.6.3rc2, the sessions starts normal, and when the context start to grow, 60k in this case, it just produce nothing ...
Chat completion: model=Qwen3.8-27B-oQ4e-fp16-mtp, 0 tokens in 29.66s
I tried to reload the model and restart the server it is the same, I must change session to have the model reply again ...
It happened with ANE activated with the new cpu sharing on and off
I'm on a m2 max running sequoia and pi agent

here's a couple of minutes of logs:

```
2026-08-25 00:21:08,434 - omlx.patches.mlx_lm_mtp.batch_generator - INFO - [-] - MTP path activated for uid=9 (model has mtp_forward, batch=1, primed=3022)
2026-08-25 00:21:08,435 - omlx.patches.mlx_lm_mtp.batch_generator - INFO - [-] - MTP[9] finish=stop tokens=1 cycles=0 tok/cycle=0.00 accept=0/0 (n/a) emits[init=1,draft=0,bonus=0,verify=0] timing[backbone=0.0ms mtp=0.0ms sample=0.0ms cache=0.0ms]
2026-08-25 00:21:08,601 - omlx.scheduler - INFO - [-] - Cache phase timings: boundary_capture_extract=0.2ms/1, boundary_capture_sync=5.8ms/1, boundary_snapshot_save=48.8ms/1, cleanup_finished_sync=32.2ms/10, store_cache_main_boundary=9.6ms/10, store_cache_main_collect=0.3ms/5, store_cache_main_dispatch=88.4ms/5, store_cache_worker_sync=0.7ms/5
2026-08-25 00:21:08,604 - omlx.server - INFO - [-] - Chat completion: model=Qwen3.8-27B-oQ4e-fp16-mtp, 0 tokens in 28.13s (0.0 tok/s), prompt: 60366, finish_reason=stop, max_tokens=16384, request_max_tokens=16384
2026-08-25 00:21:55,879 - omlx.patches.mlx_lm_mtp.batch_generator - INFO - [-] - MTP path activated for uid=10 (model has mtp_forward, batch=1, primed=3030)
[...]

Issue #3069, opened 2026-08-23
ANE prefill (approximate INT8 path) causes behavioral divergence on out-of-distribution prompts: agent descends into memorized training trajectories (0.6.3rc2)


danigame376:
**Describe the finding**

With `qwen35_ane_prefill_enabled=true`, degenerate conversational input (casual greetings in Plan mode on an empty workspace) makes the model abandon its assistant role and instead act out **coding-agent session trajectories that appear to be memorized from training data**: it invents a developer's environment (foreign home directories, project repos, git workflows) and keeps exploring it consistently across turns. With ANE off, the same greetings produce normal assistant replies.

Four independent sessions with ANE on produced four *different* memorized environments (different usernames, projects — two of them referencing what looks like Alibaba-internal infrastructure, consistent with Qwen training data). Two sessions with ANE off, same model/harness/greetings: zero occurrences.

The divergence does **not** show up on structured technical prompts: a greedy A/B spot-check with ANE on/off produced byte-identical output. It surfaces specifically on low-entropy/out-of-distribution input, where small logit perturbations from the approximate INT8 path appear to flip borderline tokens and drop the model into a different (memorized) behavioral attractor.

**Isolation matrix** (single variable = `qwen35_ane_prefill_enabled`; same model Jundot/Qwen3.8-27B-oQ4e-fp16-mtp, same OpenCode harness, fresh sessions, identical greetings):

| Arm | MTP | ANE | Result over 3 greeting turns |
|---|---|---|---|
| 1 | on | on | ❌ foreign dev environment #1 |
| 2 | on | on (empty SSD cache — rules out cache contamination) | ❌ foreign dev environment #2 |
| 3 | off | on | ⚠️ clean ×2, then descent at turn 3 |
| 4 (field) | on | on | ❌ foreign dev environment #3, reply switched to Chinese mid-session |
| **ECO-1** | **on** | **on** | ❌ **foreign dev environment #4, incl. instructions to set git identity to a third-party @alibaba-inc.com address** |
| **ECO-2** | **on** | **off** | ✅ clean ×3 |

ECO-1 and ECO-2 are back-to-back runs through the real harness flow by the machine owner — identical everything except the ANE toggle.

Also confirmed along the way: ANE and native Lightning-MTP coexist correctly on this version when enabled before server start (prefill −51% TTFT at ~3.7k tokens maintained with MTP active) — so this report is about behavioral fidelity of the approximate path, not about the spec-decode interaction tracked in #2986.

[...]

PR #3133, opened 2026-08-25
fix(ane): preserve GDN quality with recurrent-safe prefill offload

jundot:
## Summary

- Cap Qwen ANE GDN offload at the token-local `z` projection boundary.
- Keep all recurrent `qkv` rows on the checkpoint-precision GPU path.
- Treat the configured GDN fraction as a ceiling and log when it is capped.
- Update ANE tuning so wider, runtime-equivalent GDN fractions are no longer explored.
- Extend tests, the GDN split benchmark, and experimental documentation.

## Motivation

ANE requantizes its selected projection rows to per-output-channel INT8. The previous 50% GDN split included 2,048 recurrent `qkv` rows on the validated Qwen3.6 27B layout, allowing approximation error to accumulate across long prompts and causing deterministic 32K quality failures.

The new split sends only the 6,144 token-local `z` rows to ANE, making the effective fraction 37.5% while leaving all 10,240 recurrent `qkv` rows off the approximate path.

## Validation

Tested on a Mac Studio with an Apple M3 Ultra (32-core CPU, 80-core GPU), 512 GB unified memory, and macOS 26.5.

A controlled 32K comparison on `Qwen3.6-27B-oQ4e-mtp` restored the baseline-discriminating retrieval, ordering, code, and summarization results. Six varied generations completed normally without suspected loops and reproduced the same output hashes on a second pass.

Matched single-run throughput with exact 4K/32K prefill rows and 128 generated tokens:

| Configuration | 4K prefill | 4K TG128 | 32K prefill | 32K TG128 |
|---|---:|---:|---:|---:|
| GPU only | 465.8 tok/s | 33.8 tok/s | 410.1 tok/s | 29.5 tok/s |
| ANE MLP only | 555.4 tok/s | 33.7 tok/s | 477.9 tok/s | 29.6 tok/s |
| ANE + previous GDN | 578.1 tok/s | 33.7 tok/s | 497.6 tok/s | 29.6 tok/s |
| ANE + recurrent-safe GDN | 588.2 tok/s | 33.6 tok/s | 503.0 tok/s | 29.6 tok/s |

The recurrent-safe path is 26.3% faster at 4K and 22.7% faster at 32K than the matched GPU baseline, with no decode-speed regression.

A direct 2,048-token GDN projection measured 9.60 ms for the recurrent-safe ANE/GPU split versus 14.71 ms on GPU, a 1.53x speedup.

## Tests

- Focused ANE and tuning tests: 142 passed
- Full test suite: 10,399 passed, 27 skipped, 77 deselected
- Real-server 4K/32K throughput benchmark
- Controlled 32K quality and determinism validation

### DIFF
--- a/benchmarks/qwen35_ane_gdn_split_poc.py
+++ b/benchmarks/qwen35_ane_gdn_split_poc.py
@@ -76,30 +76,47 @@ def main() -> None:
 
     gpu_seconds, gpu_samples, reference = _measure(gpu_call, args.repeats)
     prepared = []
+    prepared_outputs = set()
+    qkv, z, _, _ = linears
+    z_outputs = int(z.weight.shape[0])
+    qkv_outputs = int(qkv.weight.shape[0])
+    total_outputs = z_outputs + qkv_outputs
     for fraction in args.fractions:
+        ane_outputs = patch._recurrent_safe_gdn_ane_outputs(
+            z_outputs, qkv_outputs, fraction, 128
+        )
+        if not ane_outputs or ane_outputs in prepared_outputs:
+            continue
         config = patch._AneGDNConfig(args.tokens, fraction, 8, True)
         value = patch._prepare_gdn_for_bank(gdn, config)
         if value is not None:
             state, dense0, dense1 = value
-            prepared.append((fraction, state, dense0, dense1))
+            prepared.append(
+                (fraction, ane_outputs / total_outputs, state, dense0, dense1)
+            )
+            prepared_outputs.add(ane_outputs)
+    if not prepared:
+        raise RuntimeError("No recurrent-safe GDN ANE width could be prepared")
     mx.eval(
-        *[entry[2] for entry in prepared],
         *[entry[3] for entry in prepared],
+        *[entry[4] for entry in prepared],
     )
     banks = patch._compile_dual_banks(
-        [entry[2] for entry in prepared],
         [entry[3] for entry in prepared],
+        [entry[4] for entry in prepared],
         args.tokens,
     )
     if banks is None:
         raise RuntimeError("GDN calibration bank failed to compile")
     models0, models1, programs = banks
     results = []
-    for index, (fraction, _state, _, _) in enumerate(prepared):
+    for index, (requested_fraction, effective_fraction, _state, _, _) in enumerate(
+        prepared
+    ):
         for cpu_fraction in args.cpu_fractions:
             config = patch._AneGDNConfig(
                 args.tokens,
-                fraction,
+                requested_fraction,
                 8,
                 True,
                 cpu_fraction=cpu_fraction,
@@ -118,7 +135,8 @@ def main() -> None:
                 lambda: patch._gdn_backend_exact(gdn, x), args.repeats
             )
             result = {
-                "ane_fraction": fraction,
+                "ane_fraction": effective_fraction,
+                "requested_ane_fraction": requested_fraction,
                 "cpu_fraction": cpu_fraction,
                 "median_ms": seconds * 1000,
                 "samples_ms": [sample * 1000 for sample in samples],
--- a/docs/experimental/qwen35_ane_prefill.md
+++ b/docs/experimental/qwen35_ane_prefill.md
@@ -9,11 +9,15 @@ by default.
 At the default 53% MLP request, alignment gives the two ANEs 26.5% of gate and
 up channels each (52.9% total) and leaves 47.1% on GPU. A native merge applies
 SwiGLU without materializing the full gate/up result. The GDN z+qkv input
-projection is split 50/50 between the ANEs and GPU by default. Optional FP16
-CPU sharing can take independent gate/up, down-projection, and residual GDN
-qkv slices; all three branches run in parallel and are merged natively. GDN
-recurrence, b/a and output projections, normalization, embeddings, and logits
-remain on GPU.
+projection uses a precision-aware split: ANE computes only the token-local z
+gate, while every recurrent qkv row stays off the approximate ANE path and
+normally uses checkpoint-precision GPU projection. The configured GDN fraction
+is a ceiling; a wider request is capped at the model's z boundary (37.5% on the
+validated Qwen3.6/3.8 27B layout). Optional FP16 CPU sharing can take
+independent gate/up, down-projection, and residual GDN qkv slices; all three
+branches run in parallel and are merged natively.
+GDN recurrence, b/a and output projections, normalization, embeddings, and
+logits remain on GPU.
 
 ## Requirements and limits
 
@@ -26,8 +30,9 @@ remain on GPU.
   weights are retained for every GPU suffix. The down projection may use
   compatible affine q2/q4/q5/q6/q8 weights.
 - Optional GDN acceleration accepts affine q4/q5/q6/q8 projections with group
-  size 64 or 128. Mixed q4/q5/q6/q8 layouts are supported when the ANE prefix
-  covers the full z projection, leaving a homogeneous qkv suffix on the GPU.
+  size 64 or 128. The z width must align exactly to the selected single- or
+  dual-ANE output granularity. Mixed q4/q5/q6/q8 layouts are supported because
+  ANE covers exactly z and leaves a homogeneous qkv suffix on the GPU.
 - CPU sharing requires a separately preprocessed FP16 clone of the model. It
   does not modify or dequantize the source checkpoint in place. The CPU GDN
   slice applies only to residual qkv outputs after the ANE prefix; z remains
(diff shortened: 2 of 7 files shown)


beaglemoo:
Independent corroboration of the mechanism on **NAX silicon**, plus one caveat about the validation method.

I went looking for this after #3069, on hardware outside your validation set: M5 Pro 64 GB, macOS 26.6.2, official 0.6.3rc3 DMG (build 2475), `Qwen3.8-27B-oQ4e-mtp`, mlp 0.35 / **gdn 0.45** — above the effective 0.375 ceiling this PR imposes, so recurrent `qkv` rows were on the approximate path. Method: a ~31.9K-token document with five probes sharing that prefix (retrieval, ordering, code, summarization, and a low-entropy conversational turn), greedy at temp 0, byte-identical prompts across arms, SSD cache cleared between arms, model reloaded so the banks were re-armed, and the shed counter plus `/api/status` (`configured: true, shed: false`) checked before and after each arm.

| Comparison | Result |
|---|---|
| GPU-only vs GPU-only | 5/5 byte-identical |
| ANE vs ANE | 5/5 byte-identical |
| **ANE vs GPU-only** | **3/5 diverge** (a 4th matches in content but not in its reasoning trace, 597 vs 634 chars) |

Both paths are internally deterministic and systematically disagree, which is what your recurrent-row explanation predicts, and it says the effect is not confined to pre-tensor-unit GPUs — the same 0.45 GDN split that looks harmless at greeting length diverges once ~32K of prefill has gone through it. Happy to re-run this identical matrix against your branch to confirm the recurrent-safe split restores GPU-identity here; that seems like the useful confirmation to have from a NAX part, and it is about ten minutes of work.

**The caveat, on "reproduced the same output hashes on a second pass":** that check may not be load-bearing on an MTP checkpoint. My first attempt at the matrix above showed 7 of 8 probes diverging between ANE and GPU, which looked conclusive — until the control failed. GPU-only against *itself*, same prompts and cleared cache, diverged on 4 of 8 probes. With `mtp_enabled: false` the pipeline becomes byte-reproducible at 32K (the 5/5 rows above); with depth-8 MTP on it is not. The plausible mechanism is that the accepted-draft count varies between runs, so the verify forward runs at different batch widths and the differing reduction order flips near-ties.

If your 32K validation ran with MTP enabled and still hashed identically twice, that is worth knowing in itself — it would mean the nondeterminism is specific to depth or to this part, rather than general to speculation. Either way, the arms that matter for a quality claim probably want MTP disabled so the only variable is the prefill path.

beaglemoo:
Ran the verification I offered, on NAX, across all three commits: M5 Pro 64 GB, macOS 26.6.2, branch built from source with `OMLX_WITH_CUSTOM_KERNEL=1` (only `4c8e43a5` touches the kernel; the other two are Python-only), isolated server and base path, `Qwen3.8-27B-oQ4e-mtp`, requested mlp 0.35 / gdn 0.45, dual-ANE, `mtp_enabled: false` throughout so the pipeline is byte-reproducible, ~31.9K-token document with five probes (retrieval, ordering, code, summarization, low-entropy turn), temperature 0, cache cleared between arms, shed counter and `/api/status` checked before and after each arm.

The cap engages as intended on every commit:

```
Capped ANE GDN output rows from requested 0.450 to 0.375: ANE handles only
token-local z while recurrent qkv stays off the approximate ANE INT8 path
```

**Results**

1. **The GPU path is untouched.** GPU-only on `c0503a9e` is byte-identical to GPU-only on the stock 0.6.3rc3 DMG across all five probes. That also says the harness reproduces exactly across three different builds.

2. **On this hardware all three commits produce the same ANE output.** Fresh process, fresh cache, full prefill: the ANE arm on `4c8e43a5` (cap only), `b9d3a191` (plus the GPU-suffix wait) and `c0503a9e` (plus the grouped `async_eval`) are byte-identical to each other on all five probes. So neither boundary fix changes anything observable on NAX at 32K, and neither regresses it — consistent with the ordering window simply not opening here.

3. **ANE is still not bit-identical to GPU** (2/5 identical, before and after). Expected, since token-local `z` rows remain INT8; I mention it only so my earlier comment is not read as predicting bit-exactness from this PR.

4. **Quality is correct in every arm**: stock and patched, ANE and GPU-only, all recover the planted facts and list the planted incidents in order. This checkpoint at 32K has no failure for the PR to repair, so the #3117 result you reproduced was never going to show here.

**One observation that may be relevant to #3117's shape, offered with its caveat.** While rerunning I accidentally produced a *partial* prefix-cache restore: the on-disk cache was cleared under a live server, so each probe reused 2048-8192 tokens from the in-memory blocks and re-prefilled the rest (`re-prefills 29887 of 31935 tokens (reused 2048)` and so on). In that flow, with MTP off and temperature 0:

- GPU-only: full prefill and two partial-restore runs are byte-identical to each other, 5/5.
- ANE (`c0503a9e`): partial-restore output differs from full-prefill output on 4/5 probes, and two partial-restore runs differ from *each other* on 2/5.
- ANE (`b9d3a191`): same picture — 4/5 vs full, 2/5 between the two partial runs — so it is not introduced by `c0503a9e`.

All of those answers are still correct; this is phrasing-level divergence, not the immediate-EOS. And the normal full-restore multi-turn flow (probes 2-5 served from a complete cache) *was* deterministic on ANE in every process I ran. So the only non-deterministic path I can find on NAX is "restore recurrent state at a block boundary, then ANE-prefill the remainder", which is the shape you traced in #3117. I do not have a stock-rc3 arm for that flow, so I cannot say whether this branch changed it.

**The correction to my earlier comment:** it reported that ANE diverges from GPU-only at 32K and that both paths are internally deterministic, which still holds for full prefill. But I framed it in a way that invites reading it as quality damage, and I did not show that. On this machine and this checkpoint, ANE changes generation without degrading it.

Happy to run anything else here that is useful for the NAX side, including the partial-restore flow against a specific commit if you want it pinned down.

jundot:
Thanks for the thorough NAX verification. I also tested the partial-restore path on an M3 Ultra with `c0503a9e`, MTP disabled, temperature 0, dual ANE, and the 64-MLP/48-GDN recurrent-safe configuration. Two fresh 13.9K five-probe matrices, two additional natural-language runs, and two fresh 60,392-token runs were all byte-stable, including the 56,296-token ANE re-prefill case.

Together with the exact #3117 reproduction now passing across fresh processes and caches, this confirms that the current implementation fixes the known reproducible failures without regressing the validated full-prefill paths. The remaining NAX observation appears to require an additional condition, most likely the 2K block geometry or deleting cache files under a live server.

I’m going to merge this now and continue isolating the additional condition behind the partial-restore variance.

beaglemoo:
Congratulations on the merge, and thanks for running the partial-restore path on the M3 Ultra. I can now rule out one of your two candidate conditions and narrow the other, because the variance reproduces here on an **untouched cache**, with no files deleted.

Flow, on `c0503a9e`, MTP off, temperature 0, dual-ANE, 64-MLP/48-GDN with the cap active, each run a fresh process and a fresh empty cache directory:

1. Prime with a short document (~4.4K tokens) whose first ~3.2K tokens are identical to the start of the 32K document.
2. Send the five 32K probes.

The prime leaves a legitimate partial prefix in the cache; the 32K request restores it and re-prefills the rest through ANE. The log for the first probe reads `re-prefills 29886 of 31934 tokens (reused 2048)` in every run.

| Arm (two fresh processes each) | run A vs run B | vs fresh full prefill |
|---|---:|---:|
| GPU-only | 5/5 identical | 5/5 identical |
| ANE | **0/5 identical** | 0/5 identical |

Three details that may help isolate it:

- **The restore is 2048 tokens, not the ~3.2K shared prefix**, so the boundary sits on a 2K tile edge — consistent with your geometry guess. Only the first probe re-prefills; probes 2-5 then hit the full cache written by that first prefill and inherit its differences, so the divergence originates in the one restore-then-ANE-prefill step.
- The ANE arm's variance is run-to-run, not a fixed offset: two runs of the identical flow disagree, and each also disagrees with full prefill. GPU-only through the same restore is byte-stable, so the restore itself is fine.
- Answers remain correct in every run — one run spent more of its budget thinking and got truncated at the 250-token cap, but nothing was wrong. This is still the near-tie-flip character, not a failure; I mention it because #3150's agent-session reports are exactly where a flip per restore would compound.

Everything else from the earlier matrix holds: fresh-process full prefill through ANE is byte-stable across all three of the PR's commits, so this is confined to "restore at a 2K boundary, then ANE-prefill the remainder" on NAX. If it would help, I can run the same flow with the restore boundary moved off the 2K edge (a longer or shorter prime), or with `dual_ane` off, to split tile geometry from the two-instance scheduling. Full arm data and the exact prime/probe construction are available if you want them.

PR #3135, opened 2026-08-25
feat(cluster): evict competing local models on memory-attributed failure (opt-in, structured)

alytaphoenix:
## Summary

Reworks the feature originally shipped in closed PR #2870, addressing the maintainer's exact closing feedback verbatim:

> Closing this one though: a server-initiated kill of models a user loaded on their own Mac is a product call I'm not making right now, and a regex over free-form error text is too fragile a base for a destructive action even behind a gate. If this comes back later it will need to be opt-in and keyed on structured failure attribution, rank and node as fields rather than parsed prose. The token minting pattern and C5's advertised admin_port are both worth keeping for that day.

Full design writeup: `docs/cluster-competing-model-eviction-redesign.md` (in this branch, untracked by convention — same as the other design docs already in `docs/` for this rollup).

**Stacks on #3128/#3130/#3131/#3132 (Phases 1-4 of an unrelated UI/UX pass, all still open)** — cherry-picked ahead of this commit so the branch is self-contained against `main`; the diff will shrink to just the last commit once those merge.

### What changed vs. #2870

- **Opt-in, off by default.** New `ClusterSettings.auto_evict_competing_local_models` (default `False`), a labeled toggle on the admin Cluster tab with explicit copy about what it does. `activate_cluster_deployment` reads this via a new `_auto_evict_enabled()` gate *before* any eviction code runs at all — with the flag off, behavior is byte-identical to a build with no eviction code.
- **Structured failure attribution, not regex.** Traced the actual data flow: a failing rank already computes `type(exc).__name__` and already writes a marker with a real `rank` field — both were being thrown away into one joined string before `routes.py` had to regex them back apart. Now: the marker also gets `error_type` (+ `required_bytes`/`current_bytes` for `InsufficientMemoryError`) as real fields; a new `RankFailure` dataclass and `DistributedJobSupervisor._runtime_failures()` expose them structured; `DistributedLaunchError` carries an optional `rank_failures` payload; `routes.py`'s `_memory_squeezed_hosts` reads `exc.rank_failures` directly — no regex anywhere.
- **Fail closed, not "evict everywhere."** #2870 fell back to evicting every host in the deployment when no rank named a memory failure. This rework does not — no matching `error_type` means no eviction. Guessing blast radius on a destructive action was flagged as the same class of fragility as the regex.
- **No port-guessing.** #2870's peer script fell back to guessing conventional ports (8000, 9000) if it couldn't read its own configured port. Dropped — the peer reads its own authoritative `settings.json` port or reports the failure; sending a signed admin request to a guessed destination is the same fragility class again. (C5's advertised `admin_port`, PR #2885, is still open/unmerged and itself depends on #2875 — this ships on the fallback path the design doc describes for that case, and should switch to `admin_port` once C5 lands.)
- **SSH/token-minting mechanics: unchanged**, as the maintainer confirmed they were sound.

### A real bug caught while implementing (not in the original design)

The gate check itself needed the same crash-isolation discipline as the recovery it guards. `get_settings()` raises when `init_settings()` was never called (worker-only installs, some test apps) — the first cut called it directly in the `except DistributedLaunchError` handler, which would have let that exception escape and replace the intended 503 with an unrelated 500. Caught by `test_cluster_routes.py::test_cluster_activation_rolls_back_when_canary_fails` (a real activation failure against an app with settings uninitialized). Fixed: `_auto_evict_enabled()` wraps the settings read and fails closed.

## Test plan

- [x] `tests/test_cluster_local_eviction.py` — 25 new tests: settings default/round-trip, the gate's fail-closed behavior (including the settings-uninitialized case above), marker field round-trip, structured extraction (including an old-format marker with no `error_type`), the peer script (SSH boundary, interpreter fallback, non-JSON rejection, no port-guessing), and the full routes-layer recovery flow including the new fail-closed-on-no-match behavior and the allow-list's extensibility
- [x] Full cluster slice (`test_cluster_local_eviction/routes/launch/inference_worker.py`): 186 passed against this PR's `origin/main` base
- [x] Broader admin/dashboard/settings/cluster slice: 2219 passed against the same base
- [x] Full local suite (working branch): 10429 passed (was 10404 before this commit), 0 failures
- [x] Live Playwright check of the admin toggle: renders correctly (default off), click round-trips through the real save API, persists server-side, verified via the actual tab-switch mechanism (not just direct function calls)


alytaphoenix:
Rebased onto current main. This one is stacked (routes.py, dashboard.js, i18n files, and settings.py all inherited cleanly from the already-rebased Phase 4 chain — #3128/#3130/#3131/#3132 — its own commit cherry-picked with zero conflicts once that base was correct). Full suite: 10,853 passed, 0 failed, 152 skipped, 77 deselected.

PR #3137, opened 2026-08-25
i18n hardening: strip locale padding, extract cluster view strings (Phase 5.1-5.3)

alytaphoenix:
## Summary

Continues the UI/UX hardening pass (stacked on the Phase 1-4 PRs #3128/#3130/#3131/#3132 and the eviction PR #3135) with the first three items of Phase 5 (i18n):

- **5.1** — Strip 2773 English-identical entries from the 8 non-English locale files. The admin dashboard's `_load_locale()` already falls back to English for any missing key (both server-side `t()` and client-side `window.t()`/`locale_json` go through it), so storing the English text again in a "translated" locale file was pure padding — mechanically a no-op at runtime.
- **5.2** — Add the missing `models.queue.retry_tooltip` key, fix both call sites (one was raw `title=`, not even Alpine-bound), and remove a dead `window.t(...) || 'Retry download'` fallback — `window.t` always returns a truthy string (the key itself on a miss), so the `||` branch could never fire.
- **5.3** — Full key-extraction pass over the Cluster view: `_cluster.html` (2670 lines) and the cluster sections of `dashboard.js`. Covers the SSH pairing wizard, the regenerate-key `confirm()` dialog, every pairing/eviction notification string, `clusterQuickStatus`/`clusterPrimaryActionLabel`, and the memory/shard planning UI — everything the hardening doc's D3 inventory named. ~300 new `cluster.*`/`js.error.*` keys, added to `en.json` only (see below).

## Test invariant change (please read before reviewing the locale diffs)

The old `test_locale_key_sets_identical` asserted strict `keys == base` equality between every locale and English — i.e. every locale file had to contain every English key, verbatim. Stripping padding necessarily breaks that. It's replaced with two invariants that better match what the app actually guarantees at runtime:

- `test_locale_keys_are_subset_of_english` — a locale may omit keys (they fall back to English), but an orphaned key with no English counterpart is a real bug.
- `test_locale_values_are_not_redundant_with_english` — new: no locale value should be byte-identical to English for the same key. This is the actual anti-regression guard for future padding (a translator/contributor accidentally leaving a copy-pasted English string).

New keys for this and future extraction work go into **`en.json` only** — the fallback makes storing them in all 9 files unnecessary, and the new invariant above would immediately flag it as padding if someone did.

Nine other test files asserted i18n key presence by reading the raw locale JSON directly (`json.loads(path.read_text())`), which doesn't account for the fallback. They now call `omlx.admin.routes._load_locale()` instead — same guarantee the app actually provides, correctly modeled.

## Node-harness `window.t` stubs

Three test files execute individual cluster JS functions (`clusterQuickStatus`, `clusterPrimaryActionLabel`, `clusterErrorMessage`, etc.) in a bare Node `-e` sandbox to test their logic directly. Extracting those functions' strings into `window.t(...)` calls means the sandbox now needs a `window` global — added as `global.window = { t: ... }` backed by the real `en.json` content (matching `base.html`'s actual contract), not a passthrough stub, since two of the affected tests assert on the literal rendered label text.

## Known gap (deliberately deferred, not part of this PR)

`clusterSteps`, `clusterStrategyOptions`, `clusterNodeRoles` labels, and the fabric metric labels ("Ring latency", "Collective throughput", etc.) are string literals inside `dashboard.js` **data objects**, not template expressions — the hardening doc's D3 inventory didn't enumerate these, so they're left for a follow-up pass rather than expanding this PR.

## Test plan

- [x] Full Python suite: 10430 passed (coordinator), 10348 passed (this worktree against `main`)
- [x] All 9 locale files re-verified: no orphaned keys, no English-redundant values
- [x] Live Playwright check against a scratch server: Cluster tab heading renders translated text, no raw `{{ t(...) }}` leaks in the DOM, Alpine parsed and ran (tab-switch via `$data().mainTab` worked), the eviction toggle card renders
- [ ] Coordinator verification of `clusterQuickStatus`'s dynamic `window.t(...).replace(...)` paths against real cluster state (the scratch environment has no live cluster, so the quick-start section never mounts there)


alytaphoenix:
Rebased onto current main (stacked on the fixed Phase 1-4 + eviction chain). Real conflicts: 8 locale JSON files had English-only padding independently added by unrelated work landed after this branch was cut (5 files: an unrelated \`settings.advanced.*\` block sitting at the same insertion point as this PR's own padding-strip target — removed, matching this PR's whole purpose; 3 more files caught the same pattern on a second pass I'd initially missed in a truncated diff view).

The full suite then caught two real regressions from that same unrelated-landed-work interaction, both genuine bugs this PR's own invariants exist to catch:
- 3 \`qwen4_ple_ssd_offload_*\` keys were byte-identical-to-English padding in 6 of 8 locales (ko.json actually had real translations, left alone) — caught by this PR's own new \`test_locale_values_are_not_redundant_with_english\` guard.
- \`tests/test_admin_audio_upload_ui.py\` (added independently after this branch was cut) still read raw locale JSON instead of the fallback-aware \`_load_locale()\` pattern this PR converted nine other test files to — updated it to match.

Full suite: 10,854 passed, 0 failed, 152 skipped, 77 deselected.

PR #3138, opened 2026-08-25
fix(mtp): activate prompt priming for legacy MTP under BatchGenerator

### LINKED ISSUE #3079: Prompt priming never activates for legacy-MTP DeepSeek-V4 under BatchGenerator (two coupled activation bugs; +19% long-prompt decode once fixed)
While serving DeepSeek-V4-Flash-0731 (legacy single `mtp.0`, not DSpark) through
the BatchGenerator path, I noticed `primed=0` on every request and traced the
prompt-priming activation chain. Two independent bugs each kill it on their own;
both need fixing before priming can engage. With both fixed, on a 2.1K-token
prompt at fixed depth-3 chaining I measure draft acceptance d1 81.5% → 95.6%
(d2 54.5 → 66.7%), tokens/verify-cycle 2.37 → 2.81, decode +19.4% — so this is
worth reclaiming.

**Bug 1 — `_anchor` only accepts plain-int offsets, but batch caches never
expose one.** `prompt_priming._anchor` probes `type(c.offset) is int`. Under
BatchGenerator, caches are merged into `BatchKVCache` /
`BatchRotatingKVCache` at `PromptProcessingBatch.__init__` (mlx-lm
`generate.py` → `_merge_caches`), whose `.offset` is a 1-element `mx.array`
**even for a single request (B==1)**. So `_anchor` returns None and
`maybe_capture` bails silently on every capture site — the last (hidden,
next-token) pair is never folded and `take_primed` later discards the seam on
offset mismatch. Notably `_activation_offset` in the same file already
tolerates 1-element array offsets ("activation-time only, never per-chunk"),
so I believe the capture-time scalar-cache assumption is unintentional.

Fix I'm running: an anchor view that unwraps size-1 array offsets to int (the
`(1, S)` input guard already restricts capture to singleton timelines, so B>1
ambiguity doesn't arise). Two caveats worth deciding consciously:
- this adds a per-chunk `int()` sync, which contradicts the "never per-chunk"
  comment — reading `BatchRotatingKVCache._offset` (a python int) would be a
  sync-free alternative, but note `_offset` is the buffer length, not the
  token count, so it needs care;
- with batch capture alive, a B>1 prefill window is invisible to capture; a
  cancelled request's stale ctx could in principle seam-match a later chunk
  boundary (2048-aligned). Output correctness is protected by verify either
  way, but dropping the ctx on any `inputs.shape[0] != 1` forward would keep
  the "never a wrong history" invariant.

**Bug 2 — `mtp_take_primed` hook is DSpark-only but registered
unconditionally, and the generic seam is unreachable behind it.** The
deepseek_v4 patch registers `mtp_take_primed` on the class; for non-DSpark
(legacy MTP) it returns None unconditionally. `prompt_priming.take_primed`
returns whatever a callable hook returns, so the generic ctx seam below it can
never run for legacy models — even with Bug 1 fixed, activation still dies.

Fix I'm running: treat a hook returning None as "declined ownership" and fall
through to the generic seam. I checked the other hook implementations for
safety: DSpark and inkling both pop their own ctx before returning None, so
fallthrough cannot pick up a foreign ctx; and on DSpark-enabled models the
patched `__call__` returns before generic capture, so there is no generic ctx
to pick up. (A defensive `isinstance(_PrimeCtx)` check in the generic seam
would future-proof this.)

**Related observation (no kernel bug).** The same scalar-cache assumption also
silently disables the `wsdpa_prefill` fast path under batch caches (its int
offset guard falls back to stock SDPA), so batch serving quietly loses that
optimization too. I initially suspected the kernel itself of producing wrong
output in the MTP (`_standard_mask`, no pool) context, but differential
testing against stock SDPA across cold / rotated / cache-None shapes (H=64,
L up to 2048, RotatingKVCache with real trim/rotation) shows bf16-noise-level
agreement everywhere — the kernel is fine; I'm happy to share the diff
harness. The systemic pattern is just: pre-batch-era int-offset probes die
silently in the batch engine, and priming happened to be the user-visible
casualty.

Happy to send these as a PR (anchor view + hook fallthrough + the defensive
guards) if that's welcome.

avlp12:
Fixes #3079.

Prompt priming never activated for legacy single-head MTP models (DeepSeek-V4-Flash-0731's single `mtp.0`, not DSpark) served through `BatchGenerator` — every request reported `primed=0`. Two independent bugs each killed it on their own, so both had to be fixed before priming could engage. This is the PR I outlined in the issue.

## Bug 1 — the anchor probe missed every batch cache

`prompt_priming._anchor` required `type(c.offset) is int`. Under `BatchGenerator` the per-request caches are merged into `BatchKVCache` / `BatchRotatingKVCache` at `PromptProcessingBatch.__init__` (mlx-lm `_merge_caches`), and their `offset` is a 1-element `mx.array` **even for a single request (B==1)**. So `_anchor` returned None on every batch-engine prefill, `maybe_capture` bailed silently, the head history was never folded, and `take_primed` later discarded the seam on offset mismatch.

`_anchor` now returns a small view (`_IntOffsetAnchor`) that unwraps size-1 array offsets to int. `_activation_offset` already tolerated the same shape, so its private reader is now shared as `_read_offset` and both paths agree by construction. A real B>1 cache has a multi-row offset and still finds no anchor.

On the caveat I raised in the issue: I kept the `int()` unwrap rather than reading `BatchRotatingKVCache._offset`, so this does add one sync per captured forward — `_offset` counts buffer slots rather than tokens, which the contiguity invariant cannot use. The scan itself stays sync-free (`_offset_readable` only inspects `.size`), so it is exactly one read per captured forward, alongside the `mx.async_eval` the fold already issues per chunk. Happy to revisit if you would rather not pay it.

On the stale-context caveat: with capture alive under batch caches, a B>1 prefill window advances the anchor without capture seeing its tokens, so a later singleton chunk could in principle read as contiguous across it (chunk boundaries are 2048-aligned across requests). `maybe_capture` now drops the context on any `inputs.shape[0] != 1` forward, which keeps the "never a wrong history" invariant rather than relying on verify to clean up after it.

## Bug 2 — the DSpark-only hook made the generic seam unreachable

The DeepSeek-V4 patch registers `mtp_take_primed` on the class unconditionally, but only DSpark builds answer it — for legacy MTP it returns None. `take_primed` returned whatever the hook returned, so the generic ctx seam below it could never run and activation died even with Bug 1 fixed.

A hook returning None is now read as declining ownership and falls through to the generic seam. I checked the other implementations for safety before doing this: DSpark and inkling both pop their own context before returning None, so the fallthrough cannot pick up a foreign one; and on DSpark-enabled models the patched `__call__` returns before generic capture, so there is no generic context to pick up. The generic seam also now guards on `isinstance(_PrimeCtx)`, so a future host that shares the slot and forgets to pop it still cannot have its context consumed.

## Tests

All in `tests/test_mtp_prompt_priming.py`, following the existing tiny random-weight qwen3_5 harness (no model download, CPU-only):

- `TestBatchCacheAnchor` — array-offset unwrap, `CacheList` container search, B>1 rejection, and that the view tracks the live offset rather than snapshotting it.
- `TestBatchCacheAnchor::test_batch_cache_prefill_primes_end_to_end` — legacy single-head activation end-to-end over the batch-engine cache shape (built through the public `BatchKVCache.merge`, the same path `_merge_caches` takes), capture through the seam, checked against the one-shot oracle fold.
- `TestCaptureSkips::test_batch_forward_drops_pending_ctx` — a B>1 forward invalidates the pending timeline.
- `TestHookFallthrough` — declining hook falls through to the generic seam; an answering hook still owns the seam; and a declining hook that leaves a foreign context behind is not adopted.

6 of these fail on `main` and pass with the fix. The full non-slow suite is green locally: `10305 passed, 128 skipped, 77 deselected`.

## Measured effect

DeepSeek-V4-Flash-0731, 2.1K-token prompt, fixed depth-3 chaining, single M3 Ultra:

| | before | after |
|---|---|---|
| draft acceptance d1 | 81.5% | 95.6% |
| draft acceptance d2 | 54.5% | 66.7% |
| tokens / verify cycle | 2.37 | 2.81 |
| decode | — | +19.4% |

## Out of scope

The prefill kernel gate (`wsdpa_prefill` and friends falling back to stock SDPA under batch caches, 41 of 43 layers on DeepSeek-V4-Flash) is deliberately **not** in this PR, as I said in the issue thread. The naive `B==1` plain-cache bypass recovers the single-box loss but hung TP2 lockstep in my distributed setup, so it needs its own issue once I have isolated the trigger. The kernel itself is not at fault — differential testing against stock SDPA across cold / rotated / cache-None shapes agrees at bf16 noise level. I am still happy to contribute that SDPA diff harness as a test utility if you want it; say the word and I will send it separately.

### DIFF
--- a/omlx/patches/mlx_lm_mtp/prompt_priming.py
+++ b/omlx/patches/mlx_lm_mtp/prompt_priming.py
@@ -28,7 +28,9 @@ Fail-safe invariant: every capture verifies the anchor offset advanced
 contiguously since the previous capture (``expected_offset``). Any rewind,
 trim, request switch, or unknown cache path breaks the equality and
 invalidates the context, degrading to the current unprimed behaviour —
-never to a wrong history.
+never to a wrong history. A batched (B>1) forward advances the anchor
+without capture seeing its tokens, so it drops the context outright rather
+than let a later singleton chunk read as contiguous across it.
 
 Capture sites (each calls :func:`maybe_capture` after the backbone forward):
 
@@ -131,8 +133,61 @@ class _PrimeCtx:
     window_exceeded: bool = False
 
 
+def _read_offset(entry: Any) -> Optional[int]:
+    """``entry.offset`` as a plain int, unwrapping size-1 array offsets.
+
+    Batch caches (``BatchKVCache`` / ``BatchRotatingKVCache``) hold their
+    offset as a 1-element ``mx.array``. Reading it costs one sync, so
+    callers do it once per forward at most.
+    """
+    offset = getattr(entry, "offset", None)
+    if type(offset) is int:
+        return offset
+    if offset is not None and getattr(offset, "size", 0) == 1:
+        try:
+            return int(offset.reshape(()).item())
+        except Exception:
+            return None
+    return None
+
+
+def _offset_readable(entry: Any) -> bool:
+    """Whether :func:`_read_offset` can serve this entry — no sync."""
+    offset = getattr(entry, "offset", None)
+    return type(offset) is int or (
+        offset is not None and getattr(offset, "size", 0) == 1
+    )
+
+
+class _IntOffsetAnchor:
+    """Anchor view exposing a scalar-or-size-1-array offset as an int.
+
+    Under ``BatchGenerator`` every request's caches are merged into
+    ``Batch*`` entries at ``PromptProcessingBatch.__init__``, whose
+    ``offset`` is a 1-element ``mx.array`` **even for a single request**
+    (B==1). The plain-int probe this replaces therefore found no anchor on
+    any batch-engine prefill, so ``maybe_capture`` bailed silently and
+    priming never activated there (#3079).
+
+    The unwrap is unambiguous because :func:`maybe_capture` only captures
+    ``(1, S)`` forwards — a singleton timeline. It does cost one ``int()``
+    sync per captured forward, which is what the contiguity invariant is
+    built on; ``BatchRotatingKVCache._offset`` would be sync-free but
+    counts buffer slots rather than tokens.
+    """
+
+    __slots__ = ("_cache",)
+
+    def __init__(self, cache: Any) -> None:
+        self._cache = cache
+
+    @property
+    def offset(self) -> Optional[int]:
+        return _read_offset(self._cache)
+
+
 def _anchor(cache: Optional[List[Any]]) -> Optional[Any]:
-    """First cache entry with a plain-int offset.
+    """First cache entry whose offset can be read as an int, as a view.
 
     Container layers (``CacheList``-style, exposing ``.caches`` — DeepSeek-V4
     and GLM-5.2 backbones) are searched one level deep: the container itself
@@ -141,11 +196,11 @@ def _anchor(cache: Optional[List[Any]]) -> Optional[Any]:
     if not cache:
         return None
     for c in cache:
-        if type(getattr(c, "offset", None)) is int:
-            return c
+        if _offset_readable(c):
+            return _IntOffsetAnchor(c)
         for sub in getattr(c, "caches", ()) or ():
-            if type(getattr(sub, "offset", None)) is int:
-                return sub
+            if _offset_readable(sub):
+                return _IntOffsetAnchor(sub)
     return None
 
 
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the additional measurements. The size-1 offset unwrap looks reasonable for this PR, and the legacy Lightning MTP fix looks good to me.

A separate differential harness for the batch-cache WSDPA path could be useful as a follow-up. Given the TP2 hang with the B==1 plain-cache bypass, a batch-offset-aware route that keeps the current cache representation may be worth exploring, with TP2 coverage if possible.

PR #3141, opened 2026-08-25
mac: localize menubar Live Activity strings, fix no-API-key stats hint (Phase 5.6-5.7)

alytaphoenix:
## Summary

Wraps up the UI/UX hardening pass's Phase 5 (i18n) — stacked on the prior six PRs (#3128/#3130/#3131/#3132/#3135/#3137/#3140). Unlike those, this one is Swift-only (the macOS menubar app), so it's kept as its own PR rather than bundled with the web-dashboard work.

**5.6 — Menubar Live Activity strings.** `MenubarStatsPoller.Stats.LiveActivity` built several user-visible strings outside the `Localizable.xcstrings` catalog: the prefill `"N tok/s"` and `"X left"` detail fragments, manually-pluralized `"N queued request(s)"`, the `"Active request"` fallback, and the throughput-bench export's `"No results yet."`. All now go through `String(localized:)`. Singular/plural got two distinct keys rather than embedded ternary logic, matching how this catalog already handles other count-dependent strings (e.g. `bench.accuracy.benchmarks.subtitle.empty` vs `.count`). Left the `menuBarTitle` glyph-tag strings (`"GEN …"`, `"WAIT n"`, `"RUN …"`, `"PP …%"`) untouched — the hardening doc explicitly documents these as intentionally unlocalized compact status codes, distinct from the full-prose `detail` text.

**5.7 — No-API-key stats placeholder.** The "Set OMLX_API_KEY to enable stats" placeholder was keyed on `statsPoller == nil`, which is only true when the endpoint URL fails to resolve — never because of a missing key (the poller is always constructed once a base URL exists; the key is just a parameter). Tracing `MenubarStatsPoller.refreshOnce()` turned up a second, real bug the wording issue was masking: the public status endpoint that feeds Session stats has no auth gate, so `sessionStats` populates regardless of API key while `alltimeStats` stays nil without one — meaning `session == nil && alltime == nil` (the condition guarding the whole placeholder) goes false as soon as any poll succeeds at all, and the All-Time section silently renders four `"—"` rows forever with no explanation. Fixed both: the placeholder keys on `config.apiKey` now, and the All-Time section itself shows the same hint in place of its empty rows when there's no key. Also corrected the doc-suggested "Settings > Security" wording — this app's navigation is a flat sidebar (`sidebar.security` is a direct top-level item, there's no nested "Settings" parent) — to "Set an API key in the Security screen to enable stats", and updated the pre-existing catalog entry's `ru`/`zh-Hans` translations to match (referencing the already-translated `sidebar.security` terms) instead of leaving them describing the old, wrong env-var wording.

## Catalog notes

This project's `Localizable.xcstrings` entries are `extractionState: "manual"` — `xcodebuild build` doesn't auto-populate the catalog for new `String(localized:)` call sites here, so the 6 new keys from 5.6 were added by hand. Matched the catalog's actual persisted format for interpolated strings (`%lld`/`%@` positional specifiers, confirmed against existing entries — not the literal Swift `\(...)` interpolation syntax) rather than guessing. New keys are English-only; left `ru`/`zh-Hans` absent rather than provide translations I wasn't confident in.

One easy mistake avoided: a first attempt to update the `menubar.stats.no_api_key` catalog entry via a full JSON parse-and-rewrite reformatted the *entire* 28k-line file (Xcode serializes with `"key" : "value"` — space before the colon — while `json.dump` doesn't), producing a ~19k-line noise diff. Reverted and did a surgical text-level edit instead, matching the existing per-Tailwind-CSS lesson from earlier in this pass about build-artifact reformatting blowing up diffs.

## Test plan

- [x] `xcodebuild build`: clean, no errors, before and after the catalog edits
- [x] `xcodebuild test`: 290/290 passed (coordinator checkout and this worktree against `main`)
- [x] Full Python suite: 10348 passed in this worktree (no Python files touched by this PR; run to confirm the stacked cherry-pick base is sound)
- [x] All locale/catalog JSON re-validated as parseable after the cherry-pick


alytaphoenix:
Rebased onto current main, stacked on the fixed #3137. One real conflict: main independently added a \`max_audio_upload_size\` field validation to \`saveGlobalSettings()\` (raw English string) at the same point this PR converts the surrounding validation lines to \`window.t(...)\` — converted the new line too, reusing the existing \`settings.advanced.max_audio_upload_size\` key rather than adding a duplicate.

That surfaced one more stale test assertion (\`test_dashboard_defaults_and_posts_audio_upload_limit\` checking for the literal English string that no longer appears verbatim once converted to \`window.t(...)\`) — updated to check for the call instead.

Full suite: 10,854 passed, 0 failed, 152 skipped, 77 deselected.

PR #3118, opened 2026-08-24
feat(cluster): rebuild distributed inference as Cluster v2 Beta

jonathan308:
# feat(cluster): rebuild distributed inference as Cluster v2 Beta

## Summary

This is the complete Cluster v2 rebuild of oMLX distributed inference. It
replaces the previous manual, pipeline-first preview with one capability-driven
Mac clustering flow:

**detect -> pair -> verify RDMA/runtime -> select/stage model -> recommend
TP/pipeline placement -> sign -> load -> serve/recover through normal oMLX**

The implementation is model-layout driven rather than DS4 hard-coded. DS4
Flash is the stress/qualification model; Qwen and pipeline-compatible models
exercise independent architecture paths.

## Scale of this overhaul

This is not a UI reskin, a model-specific speed patch, or a thin wrapper around the old preview. It is a product-level replacement of distributed inference in oMLX, built and physically qualified from discovery through failure recovery.

The current PR footprint relative to upstream `main` is substantial:

- **410 commits across 400 files**;
- **102,362 insertions / 1,929 deletions**, including implementation, tests, benchmark harnesses, fixtures and documentation;
- **230 new files**;
- **47 core cluster/product files** changed, representing roughly **30,010 added lines** in the cluster, distributed-engine and Cluster UI surfaces;
- **149 test files** changed with roughly **32,927 added test lines**;
- **12,558 lines of benchmark and physical-gate tooling** plus **5,043 lines of documentation**.

Those numbers are included to make the review scope visible, not to equate line count with quality. The important point is that the work replaces an incomplete manual pipeline experiment with a signed, supervised, model-aware distributed product.

### What was actually rebuilt

- **Connection and trust:** stable node identities; Bonjour, multicast, manual-IP and persisted-peer discovery; human-approved pairing; authenticated enrollment; automatic SSH key exchange; changed-host-key refusal; Tailscale clearly limited to control/discovery.
- **Fabric qualification:** Thunderbolt/RDMA discovery, full signed link matrix, JACCL/runtime/kernel parity, honest TCP fallback, Recovery-mode RDMA enablement guidance, communicator fencing and kernel-released process leases.
- **Universal model workflow:** cluster-wide model inventory, manifest identity, resumable model/sidecar staging, heterogeneous per-node paths, remote import preflight and model-layout inspection.
[...]

PR #3155, opened 2026-08-26
feat(cluster): add trusted discovery and pairing foundation

jonathan308:
## Summary

This is **1/4** of the backend-first Cluster v2 split requested in #3118.

It establishes the trust and enrollment foundation only:

- stable per-install node identity and atomic device registry
- mDNS, multicast, manual, Tailscale, and paired-peer discovery with health/backoff reporting
- code-based coordinator/joiner pairing with lockout and expiry
- fail-closed, symmetric SSH enrollment using validated user keys and authenticated daemon host keys
- source-address binding, salted pairing verification, coordinator identity binding, and reboot rehydration
- system-Python socket carrier for macOS network-policy edge cases

## Security and lifecycle fixes

- enrollment addresses come from the observed HTTP peer, not spoofable JSON
- public keys are normalized to one validated key record; newline/multi-key injection is rejected
- both Macs pin the authenticated SSH daemon host key before persistence
- duplicate live requests cannot replace another node's pending identity
- incomplete SSH material, host-key mismatches, or missing verified addresses fail closed
- paired records are enriched without downgrading trusted state and recover after reboot

## Scope

This PR intentionally excludes:

- qualification, inventory, staging, planning, and deployment (PR 2)
- launcher/rank control/liveness/teardown/recovery (PR 3)
- TP/pipeline serving, cancellation, telemetry, and cache lifecycle (PR 4)
- Phase split and the Cluster v2 UI (follow after the backend series)
- DS4, ANE, MTP, native kernels, and other single-Mac optimizations
- updater, release, repository, and branding changes

The full reference implementation remains in #3118.

## Validation

- full suite: **10,486 passed, 102 skipped, 0 failed**
- focused trust/discovery/enrollment suite: **166 passed**
- server and cluster route regressions: **112 passed**
- pairing security suite: **42 passed**
- cluster reachability seam: passed
- Ruff and `git diff --check`: passed

### DIFF
--- a/omlx/cluster/discovery.py
+++ b/omlx/cluster/discovery.py
@@ -1024,6 +1024,7 @@ class DiscoveryService:
         self._zc_instance: Any | None = None
         self._zc_browser: Any | None = None
         self._zc_info: Any | None = None
+        self._rehydrate_paired_candidates()
 
     # -- public API ----------------------------------------------------------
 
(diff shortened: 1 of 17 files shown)


jonathan308:
Follow-up adversarial audit before review found one remaining trust-state race and it is fixed in `d0a42214`:

- coordinator approval now reserves the pending request while SSH enrollment is in progress, so duplicate approve, deny, and unpair cannot cross it
- joiner completion has the same single-flight and TTL enforcement
- pairing-key/device persistence is transactional across the two stores; a second-store failure rolls the first back
- completed SSH enrollment is revoked if persistence fails, and the pending join returns to a retryable non-approving state
- stale documentation was updated from the old blake2s/100k description to the actual identity-bound 210k PBKDF2 contract

Added concurrency and partial-persistence regression tests. Focused validation: 171 trust/discovery/enrollment tests, 112 server/route tests, and the cluster reachability seam are green. The full 3.11–3.13 CI matrix is rerunning on this commit.

jonathan308:
Downstream Cluster v2 backend drafts are now open so the split work remains visible and reviewable in dependency order:

- #3256 — qualification, inventory, staging, planning, and deployment
- #3257 — launcher, authenticated rank control, teardown, and recovery
- #3258 — TP/pipeline serving, cancellation, telemetry, and cache lifecycle

This PR remains the first review target. The downstream PRs are marked as stacked drafts and will be restacked to their bounded incremental diffs as each predecessor merges. The complete map is also recorded on reference PR #3118.

jonathan308:
Following up on the Cluster v2 backend split from #3118 — these slices are high-impact for distributed/RDMA serving. Can we get a status update / review pass when you have bandwidth? Happy to restack or address feedback quickly. @jundot

jundot:
@jonathan308, thanks for separating out the discovery and pairing foundation. This gives the later backend PRs a clear starting point without changing the existing inference path.

LGTM, I'll merge this first. I'll handle HTTP port persistence, scoped IPv6 address handling, and checking host keys before adding SSH authorization in a small follow-up. Once this lands, please restack #3256 onto `main` so I can review its incremental changes next.

One note for the final integration/UI PR: once the v2 flow is fully wired up, please remove any legacy v1 paths that are no longer used. Anything still needed for compatibility or migration can stay.

PR #3156, opened 2026-08-26
fix(ane+cache): refuse ANE prefill on MoE models; stamp ANE provenance in SSD cache

alytaphoenix:
## Summary

Two fixes from a live incident on `Qwen3.6-35B-A3B` (`qwen3_5_moe`): enabling ANE prefill produced silently corrupted output — pure `!!!` garbage on any prompt long enough to engage the fixed-shape ANE path — at completely plausible speed (the ANE tuner even measured an 18.4% "speedup" and blessed the config, since it validates throughput only). The corrupted prefill was then persisted into the SSD prefix cache and replayed on every same-prefix request, surviving both disabling the setting and reloading the model.

### Fix 1 — dense-only gate, three layers deep

The family check was a prefix match (`qwen3_5*`) that let `qwen3_5_moe` through, but the patch offloads *dense* MLPs only: a MoE's shared-expert MLPs still expose `gate_proj`/`up_proj`/`down_proj` and match the candidate scan, while the routed experts (the bulk of every token) cannot be served by fixed-shape ANE programs.

- Settings `PUT` now 400s on enable for any `*moe*` config type; **disable stays allowed** so a stuck config can always be cleared.
- `enable_qwen35_ane_prefill()` structurally detects MoE blocks (`switch_mlp`/`experts` attribute markers, verified against the installed `qwen3_5_moe` model classes) and refuses with a loud warning regardless of what any config string claims — covering the tuner, benchmarks, and raw-script callers.
- The dashboard no longer offers the ANE toggle for MoE models (mirrors the backend gate).

### Fix 2 — ANE provenance in the cache signature

ANE-computed KV is a different numeric path than GPU-computed KV, and if the ANE path ever corrupts, an unstamped cache converts transient corruption into persistent poisoning. Cache signatures now stamp `ane_prefill: true` when the engine actually activates ANE (only stamped when active, so all existing signatures stay byte-identical — same convention as `turboquant_kv_bits`). Compatibility checks, `signature_mismatch_reason`, and the stale-signature sweep reject provenance mismatches in both directions, and toggling ANE re-arms the sweep exactly like a TurboQuant depth change.

## Test plan
- [x] New: settings PUT rejects enable on `qwen3_5_moe`, allows disable
- [x] New: patch entry refuses a structurally-MoE model before marking any module
- [x] New: signature stamps only when active; provenance mismatch rejected both ways; ANE toggle re-arms the sweep
- [x] Full suite: 10337 passed, 102 skipped, 0 failed
- [x] `node --check` on dashboard.js

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -2328,6 +2328,21 @@ async def update_model_settings(
                 status_code=400,
                 detail="ANE prefill is available only for Qwen3.5/3.6/3.8 models.",
             )
+        # The prefix check above lets MoE variants (qwen3_5_moe, ...) slip
+        # through, but the ANE patch offloads *dense* MLPs only — on a MoE
+        # model it silently corrupts output while running at plausible
+        # speed (verified live: pure "!!!" garbage on any prompt long
+        # enough to engage the fixed-shape ANE path, with the corrupted
+        # prefill then persisted into the SSD prefix cache).
+        if enabled and "moe" in config_type:
+            raise HTTPException(
+                status_code=400,
+                detail=(
+                    "ANE prefill supports only dense Qwen3.5/3.6/3.8 models; "
+                    "MoE variants are unsupported (the fixed-shape ANE path "
+                    "cannot serve routed experts and corrupts their output)."
+                ),
+            )
         current_settings.qwen35_ane_prefill_enabled = enabled
     if "qwen35_ane_prefill_sequence_length" in sent:
         value = request.qwen35_ane_prefill_sequence_length
--- a/omlx/admin/static/js/dashboard.js
+++ b/omlx/admin/static/js/dashboard.js
@@ -7209,6 +7209,11 @@
                 const modelType = String(model?.config_model_type || '')
                     .toLowerCase()
                     .replace(/-/g, '_');
+                // Dense only: MoE variants (qwen3_5_moe, ...) match the
+                // prefixes but the fixed-shape ANE path cannot serve routed
+                // experts and silently corrupts their output. Mirrors the
+                // backend gate in update_model_settings.
+                if (modelType.includes('moe')) return false;
                 return QWEN35_ANE_CONFIG_PREFIXES.some(prefix => modelType.startsWith(prefix));
             },
 
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -259,6 +259,7 @@ def _cache_compat_signature(
     cachelist_subtypes: dict[str, list[str]] | None = None,
     payload_layout: str | None = None,
     gdn_sidecar_state_dtype: str | None = None,
+    ane_prefill: bool | None = None,
 ) -> str:
     """Return a stable compatibility signature for a persisted cache block."""
     payload = {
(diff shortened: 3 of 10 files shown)


jundot:
Thanks for the detailed incident report and patch. I retested current `main` on an M3 Ultra with `Qwen3.6-35B-A3B-MLX-4bit`, which uses the same `qwen3_5_moe` architecture: ANE activated 40 shared-expert MLP and 30 GDN procedures, and a 9,035-token prompt completed correctly, so the corruption no longer reproduces after #3133.

I also verified that an ANE-generated 8,192-token SSD prefix is restored after ANE is disabled, but the continuation remained correct. Since ANE prefill already hands its state to GPU decode, I don't consider the small numerical-path difference alone a cache incompatibility.

The MoE gates would disable a working path, while separating otherwise compatible ANE/GPU caches would unnecessarily reduce cache reuse. I therefore don't think this should be merged as-is. I no longer have the exact original oQ4-MTP checkpoint locally; if it still reproduces on current `main`, I'd prefer to investigate that specific path instead.

alytaphoenix:
Thanks — I dug into this. The exact original checkpoint is `Qwen3.6-35B-A3B-oQ4e-mtp` (the affine-quant MTP variant of your "oQ4-MTP"). I was unable to get a clean live reproduction on current `main`: the two M4-Pro machines that hold this checkpoint are both memory-pinned to unrelated production models (`idle_timeout: None`, no headroom to co-load 20GB), and the one running it in production is on a build that predates #3133 — so a repro there wouldn't be informative.

What I can add is a code-level read of `main` on the specific path. #3133's `_recurrent_safe_gdn_ane_outputs` plus the `if ane_outputs != z_outputs: return None` guard in `_compile_gdn` make it structurally impossible for recurrent q/k/v rows to enter the approximate INT8 ANE slice — only the token-local `z` gate is ever offloaded. That's exactly the accumulation mechanism behind the `!!!` corruption. The guard holds for the affine (oQ4e) quant format, and there's no MTP interaction in the ANE prefill path (and MTP drafts are verification-gated regardless), so the two variables your retest on `Qwen3.6-35B-A3B-MLX-4bit` didn't cover don't reopen it. This lines up with your clean 9,035-token retest.

So I think you're right that the MoE hard-gate (Fix 1) shouldn't merge — #3133 already closes the corruption class. Open question for you: worth keeping a trimmed Fix 2 (ANE-provenance cache stamping) purely as defense-in-depth? It only stamps when ANE is actually active and leaves all existing signatures byte-identical, so it wouldn't reduce cache reuse — but I know you'd weighed ANE/GPU cache separation already, so your call. Happy to close as obsolete otherwise.

alytaphoenix:
Correction to my previous comment — I got a live repro window and the corruption **does still reproduce** on current `main`. My earlier code-analysis conclusion was wrong; retracting it.

**Setup**: fresh `origin/main` checkout (`e467261e`, includes #3133), `Qwen3.6-35B-A3B-oQ4e-mtp` loaded via `run-fork-server.sh`, `qwen35_ane_prefill_enabled: true` (default settings otherwise — fraction=0.53, gdn_fraction capped 0.500→0.333 by the existing safety cap). Log confirms real activation: `Eagerly compiled and enabled ANE/GPU Qwen prefill on 41 MLPs`, `... GDN input projections on 30 layers`.

**Result**: a fresh ~9,025-token prompt (temp=0, uncached) returns `!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!` — reproduced twice with different prompt content/seeds, both clean 60-token garbage runs.

**Cache-poisoning symptom also confirmed**: disabled ANE, reloaded the model, re-sent the *same* corrupted prompt — got the identical `!!!` output again, this time with `cached_tokens: 8192` (replaying the poisoned SSD prefix cache, not recomputing). A genuinely fresh prompt with ANE off and no cache hit (`cached_tokens: 0`) produced normal coherent output — so this isn't ANE-off being broken generally, it's specifically the persisted-poisoned-cache replay the PR's Fix 2 targets.

So both original symptoms are real and current: the corruption itself, and the cache persisting it past a disable+reload. My structural read of `_compile_gdn`'s `ane_outputs != z_outputs` guard was either wrong about what it actually prevents, or the corruption is coming from a different path than the one I traced (the ANE MLP offload path, not just GDN) — I haven't re-diagnosed root cause, just reproduced the symptom cleanly this time with a real live run instead of static analysis. Sorry for the noise on the previous comment — should have said "no live evidence either way" instead of leaning on the code read.

PR #3157, opened 2026-08-26
fix(ane): release prefill state on engine unload

### LINKED ISSUE #3097: Memory doesn't appear to be released properly
This is noticed especially when running the Tune ANE split function. A model can get unloaded but a large chunk will still be used. For reference, my hot cache is disabled for lack of a huge chungus of memory. 

As a sidenote, it's not clear to me if the ANE split will consume more memory or not. As I'm pretty restricted with 'only' 64gb, this is important information to have when deciding to use the functionality. 

**Screenshots**
Image

 - macOS Version: 26.4.1
 - oMLX Version 0.6.3rc2

rockminster:
## Summary

The Qwen ANE prefill path attaches compiled procedure-bank state to model modules. Pressure shedding already releases that state, but ordinary text and VLM engine unload did not call the shared release helper. This keeps mapped ANE weight blobs and IOSurfaces referenced after unload.

This change calls `release_qwen35_ane_prefill()` after the engine has stopped and before wrapper model references are cleared, for both `BatchedEngine` and `VLMBatchedEngine`. The existing helper handles ordinary, fused-down, and GDN state, latches fallback flags, clears counters, and is idempotent. Release failures remain non-fatal to normal teardown.

Fixes #3097

## Verification

- `tests/test_qwen35_ane_prefill.py`: verifies all ANE state variants are dropped, counters reset, and release is idempotent.
- `tests/test_batched_engine.py`: verifies stop/close/release ordering and failure-safe cleanup.
- `tests/test_vlm_engine.py`: verifies VLM stop/release ordering and failure-safe cleanup.
- Focused suite: **433 passed**.
- `python3 -m compileall` and `git diff --check` passed.
- No live oMLX settings, model locations, caches, processes, or API keys were changed.

### DIFF
--- a/docs/experimental/ane-unload-memory-release.md
+++ b/docs/experimental/ane-unload-memory-release.md
@@ -1,0 +1,38 @@
+# ANE procedure-bank release on model unload
+
+## Problem
+
+Qwen ANE prefill attaches native procedure-bank objects to model modules. The
+objects retain mapped weight blobs and IOSurfaces until their native references
+are dropped. The existing pressure-shedding path releases those references,
+but the ordinary engine unload path does not call it. Repeated load/unload
+cycles can therefore leave the process footprint elevated until restart.
+
+## Design
+
+Reuse `release_qwen35_ane_prefill()` as the single release operation. It already
+handles ordinary and fused ANE state, latches the per-module fallback flags
+before dropping state, resets ANE status counters, and is idempotent. Call it
+after the engine has stopped and closed, but before the wrapper drops its model
+reference. This ordering prevents a live engine from observing a deliberately
+disabled ANE state while it is still serving teardown work.
+
+Apply the same ordering to both `BatchedEngine` and `VLMBatchedEngine`, because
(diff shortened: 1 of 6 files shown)


jundot:
I built the native ANE extension and verified repeated text/VLM load, inference, and unload on an M3 Ultra with Qwen3.5-4B. The unload change releases the compiled banks even when the model remains referenced, and memory settled after dropping the remaining references.

I also reproduced calibration continuing during cancellation cleanup, which can fail when unload clears its ANE state. I've prepared a maintainer follow-up to drain calibration before unload, keep tuning marked as active during cleanup, and move VLM release into the existing post-worker resource cleanup hook. Normal tuning, cancellation during compilation and calibration, repeated cancellation/reload, and active VLM teardown passed with that follow-up. I'll handle those changes rather than ask for a rework here.

jundot (approved):
Native validation confirmed that explicit unload releases the ANE banks while the model remains referenced. LGTM; I'll land the prepared calibration-cancellation and VLM teardown follow-up immediately after merging.

PR #3161, opened 2026-08-26
feat: add Qwen3.8-Flash-Next text inference

pyros-projects:
## Summary

Adds text inference for Qwen3.8-Flash-Next checkpoints (`model_type=qwen4_exp`) on the mlx-vlm version currently pinned by oMLX.

The current pin predates Blaizzy/mlx-vlm#2032. Vendoring the compatibility model here avoids pulling roughly 500 unrelated upstream commits into the oMLX dependency update.

## What is included

- Qwen4-Exp text architecture: hyper connections, PLE n-gram embeddings, QSA attention, and the sigmoid-gated GDN variant.
- Exact SplitMix64 n-gram hashing and EOS boundary handling from the Transformers reference.
- A tokenizer-only oMLX load path. It never constructs `AutoProcessor` or a vision processor.
- A parameter-free vision placeholder. Vision and MTP checkpoint tensors are filtered before strict loading, so the unused towers do not enter unified memory.
- Optional SSD-backed PLE tables using sparse row reads from read-only safetensors mmaps.
- `OMLX_QWEN4_PLE_MODE=auto|resident|mmap`; `auto` retains RAM headroom based on checkpoint size.
- Decode optimizations: fused hyper-connection input projections, single-token compiled hyper-connection graphs, and the existing MoE gate/up fusion for `qwen4_exp`.
- Explicit rejection of image/audio input while this runtime is text-only.

## Real-checkpoint verification

Measured on an M5 Max with 128 GiB unified memory, single-sequence greedy decode after warmup:

| Checkpoint | PLE mode | Peak unified memory | Steady decode | Cold prefill |
| --- | --- | ---: | ---: | ---: |
| Qwen3.8-Flash-Next-MLX-Mixed-2bit | `mmap` | 37.0 GiB | 31.5 tok/s | 3.4 s |
| Qwen3.8-Flash-Next-MLX-4bit | `auto` -> `mmap` | 74.3 GiB | 31.1 tok/s | 13.4 s |

Both checkpoints pass strict loading and real generation through `VLMBatchedEngine`; the loaded vision tower has zero parameters. The mmap Mixed run reduced peak memory from about 64 GiB to 37 GiB. Projection fusion, decode compilation, and MoE gate/up fusion improved the measured 4-bit steady decode path from roughly 23.5 tok/s to 31.1 tok/s.

## Tests

- 188 directly affected loader/model/engine regression tests pass.
- Full default suite: 10,209 passed, 102 skipped.
- Two unrelated local-environment failures reproduce on unmodified `main`: one Laguna numerical-tolerance test, and one subprocess test that cannot inherit the externally supplied app-only MLX `PYTHONPATH`.

## Current limitations

- Text inference only; vision is deliberately disabled and consumes no model memory.
- The optional MTP head is deferred.
- SSD-backed PLE currently requires an indexed safetensors checkpoint with affine 4-bit PLE shards.

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -102,6 +102,7 @@ OCR_EXTRA_STOP_SEQUENCES: List[str] = [
 VLM_LANGUAGE_PROMPT_KWARGS = ("mm_token_type_ids", "token_type_ids")
 
 COHERE2_MOE_MODEL_TYPE = "cohere2_moe"
+QWEN4_EXP_MODEL_TYPE = "qwen4_exp"
 MINIMAX_M3_VL_MODEL_TYPE = "minimax_m3_vl"
 MINIMAX_M3_MODEL_TYPES = {"minimax_m3", MINIMAX_M3_VL_MODEL_TYPE}
 
@@ -245,6 +246,30 @@ def _load_cohere2_moe_text_model(
         processor = _attach_vlm_tokenizer_runtime(tokenizer, model_path, eos_token_id)
 
     return model, processor
+
+
+def _load_qwen4_exp_text_model(
+    model_name: str,
+    *,
+    trust_remote_code: bool = False,
+):
+    """Load Qwen4-Exp text inference without constructing its vision processor."""
+    from mlx_vlm.utils import get_model_path, load_model
+    from transformers import AutoTokenizer
+
+    model_path = get_model_path(model_name)
+    model = load_model(
+        model_path,
+        lazy=False,
+        strict=True,
+        trust_remote_code=trust_remote_code,
+    )
+    tokenizer = AutoTokenizer.from_pretrained(
+        model_path,
+        trust_remote_code=trust_remote_code,
+    )
+    eos_token_id = getattr(getattr(model, "config", None), "eos_token_id", None)
+    return model, _attach_vlm_tokenizer_runtime(tokenizer, model_path, eos_token_id)
 
 
 _video_processor_patched = False
@@ -1586,6 +1611,11 @@ class VLMBatchedEngine(BaseEngine):
                         self._model_name,
                         trust_remote_code=self._trust_remote_code,
                     )
+                if _read_config_model_type(self._model_name) == QWEN4_EXP_MODEL_TYPE:
+                    return _load_qwen4_exp_text_model(
+                        self._model_name,
+                        trust_remote_code=self._trust_remote_code,
+                    )
 
                 with _load_optiq_vision_sidecar_on_load(
                     Path(self._model_name)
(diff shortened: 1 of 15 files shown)


Maheidem:
LETZZZZZZ GOOOOOOOOOOO

pyros-projects:
Pushed `efa6ccc6`, which removes the accidental Python tax from Qwen4 QSA prefill.

The old vendored path pooled/scored/top-k-selected once per query token and per sparse layer. A 2K prefill across 12 QSA layers therefore performed 24,576 Python-side selection rounds. The new path:

- skips QSA scoring entirely while the causal context still fits inside the token budget;
- pools keys once, then scores and selects blocks with batched FP32 MLX matmuls;
- tiles only the query axis, bounding the temporary score tensor at long context;
- preserves causal visibility for early rows, exact incomplete-block tails, cached-prefix semantics, and scalar decode;
- normalizes 1D, shared-batch, and 3D M-RoPE position IDs.

Matched A/B on the same M5 Max, same Mixed-2bit checkpoint, exact 4,096-token prompt, and the same mmap/cache setup:

| path | wall time | prefill |
| --- | ---: | ---: |
| previous scalar selector | 21.45 s | 190.9 tok/s |
| vectorized selector | 3.19 s | ~1,284 tok/s |

That is **6.72x**. Uncached scaling stayed around 1,224 / 1,280 / 1,203 / 1,127 / 780 tok/s at 2K / 4K / 8K / 16K / 32K for Mixed-2bit. The oQ4-MTP checkpoint measured 574 / 600 / 648 / 638 / 651 tok/s at the same lengths. A 34,816-token request with 30,720 cached tokens (~88.2%) reached ~3,953 effective tok/s.

Correctness checks include scalar-vs-vector selection, early causal rows, cache prefixes/tails, and position broadcasting. I also reproduced the nastier BF16 case (B=2, K=8192, Q=1025, two query tiles, 3D position IDs): vector and scalar paths select the same 131 tokens. Focused tests are 23/23 green on this branch and 31/31 on the MTP derivative.

The broader local suite completed with 10,366 passed, 101 skipped, 77 deselected, and one unrelated `test_dflash_laguna` numerical-tolerance failure; that same failure reproduces on the untouched base branch, so I am not laundering it into a Qwen claim. The model is fast now. Reality, regrettably, remains numerically approximate.

jundot:
@pyros-projects Thanks for the substantial work here, especially the profiling and parity coverage.

Since this PR was opened, #3169 landed the base Qwen4-Exp text/image implementation, and I followed it with #3174 for QSA and recurrent cache preservation, Lightning MTP, oQ PLE mmap, and real-checkpoint validation. As a result, most of this branch is now duplicated or structurally incompatible with current main, and the text-only vision shell would regress the image support that is now available.

There are two performance ideas here that I do not want to lose. The query-axis tiling in sparse QSA prefill is still valuable: current main vectorizes QSA selection, but it does not independently bound the full `[batch, heads, query, blocks]` score intermediate. The hyper-connection projection fusion and single-token compiled path may also be useful, although they need isolated validation against the current target-verify and lazy-materialization paths.

Would you be willing to open a fresh PR from current main, scoped first to QSA query tiling? It should preserve the current M-RoPE/image behavior, QSA cache serialization, Lightning MTP target verification, quantized and batched cache paths, and lazy model loading. Scalar/vector parity tests plus 16K/32K prompt-processing and peak-memory A/B results would make the port straightforward to review.

If you would also like to port the hyper-connection fusion and compile work, I would prefer that as a separate follow-up with isolated decode A/B results. I am going to close this superseded integration PR, but I would be happy to review those focused current-main ports.

PR #3163, opened 2026-08-26
Add native Qwen4 MTP decoding

pyros-projects:
## Summary

Adds native speculative decoding for the embedded Qwen4 MTP head on top of #3161.

- detects both `mtp.*` and `language_model.mtp.*` checkpoint layouts before tensor loading
- keeps the raw layout at the model root without double-registering the module
- implements the released one-layer QSA draft head, including global HC hidden normalization and Gemma-style RMS offset weights
- primes the MTP cache during normal prefill and supports QSA/GDN rollback after rejected drafts
- keeps MTP opt-in through the existing model setting; depth 1 is the measured general-purpose default, with deeper chains still configurable
- continues to discard the vision tower for the text-only runtime

## Measured result

`Qwen3.8-Flash-Next-MLX-Mixed-2bit`, greedy decode, PLE mmap, 512 output tokens:

| workload | no MTP | MTP | change |
|---|---:|---:|---:|
| technical prose, depth 1 | 30.32 tok/s | 36.98 tok/s | +22.0% |
| integer sequence, depth 1 | 31.22 tok/s | 48.98 tok/s | +56.9% |
| integer sequence, depth 3 | 31.22 tok/s | 58.25 tok/s | +86.6% |

The depth-1 prose run accepted 74.7% of drafts. The depth-3 sequence run accepted 100% and emitted 3.76 tokens per backbone cycle. Peak memory rose from about 37.1 GiB to 38.8 GiB at depth 1.

## Validation

- 243 focused loader, prompt-priming, MTP, VLM adapter, and Qwen4 tests pass
- real checkpoint smoke tests cover prefill priming, accepted drafts, rejected drafts, cache rollback, and depth-k chaining
- staged diff is clean and contains no deferred vision tensors in the text-only load path

This is intentionally stacked on #3161; until that lands, GitHub shows the base-port commit here too. The new commit is signed and verified.

MTP finally found the accelerator instead of the decorative shelf. 🔥

### DIFF
--- a/omlx/cache/type_handlers.py
+++ b/omlx/cache/type_handlers.py
@@ -40,6 +40,7 @@ class CacheType(Enum):
     BATCH_POOLING_CACHE = "BatchPoolingCache"
     MINIMAX_M3_KVCACHE = "MiniMaxM3KVCache"
     MINIMAX_M3_BATCH_KVCACHE = "MiniMaxM3BatchKVCache"
+    QSA_KVCACHE = "QSAKVCache"
 
 
 @dataclass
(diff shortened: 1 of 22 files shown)


pyros-projects:
The stacked branch now includes `d2363504`, the vectorized/tiled QSA prefill work from #3161 plus shared position-ID hardening. This matters here twice: the target prefill stops paying the per-token Python selector tax, and the MTP draft head keeps the same selection semantics during decode/cache rollback.

The downloaded oQ4-MTP checkpoint completed uncached prefills at roughly 574 / 600 / 648 / 638 / 651 tok/s for 2K / 4K / 8K / 16K / 32K. A signed local bundle built from this exact commit also completed a real 4,148-token uncached request through the sparse vector path in 4.28 s total server time (Mixed-2bit, including one generated token).

I am leaving this PR stacked and draft: #3161 is still the clean review surface for the base architecture/QSA work, while this one remains the MTP delta. Fewer merge hydras that way.

jundot:
@pyros-projects Thanks for building and measuring the first end-to-end Qwen4 Lightning MTP path here.

Since this stacked draft was opened, #3169 and #3174 have landed a different current-main implementation. It now covers the checkpoint prefix layouts and module ownership handled here, along with prefill priming, QSA/GDN rollback, adaptive depth up to 3, safe batch fallback, and prefix-cache serialization for resident, quantized, and batched QSA caches. It also retains text and image support instead of replacing the vision tower.

After comparing the remaining delta, I could not identify an MTP-specific change that still needs to be ported independently. The reusable performance work in this stack is inherited from #3161, and I have requested fresh current-main follow-ups there for QSA query tiling and, separately, the hyper-connection optimizations.

Because this PR is still a draft, is stacked on a superseded base, and no longer has a distinct MTP delta to merge, I am going to close it. The benchmarks and cache work here were still useful in validating the final direction; thank you for putting that groundwork together.

PR #3169, opened 2026-08-26
feat: add native Qwen3.8 Flash Next multimodal support

jonathan308:
## Summary

Adds first-class `qwen4_exp` / `qwen4_exp_text` support for Qwen3.8-Flash-Next on oMLX's pinned mlx-vlm dependency. This supersedes the text-only direction in #3161 with the merged upstream architecture and a complete oMLX serving path.

This is a distinct Qwen4 architecture, not a Qwen3.5 model alias. It implements the published 36 GDN + 12 QSA schedule, micro-block sparse QSA/indexer state, four-stream mHC residuals, 512-expert MoE routing, sigmoid output gating, and layer-1 n-gram PLE.

## Included

- Native text, image, and video inference
- Structured tool-call parsing through the normal OpenAI-compatible API
- QSA auxiliary cache state preserved across continuous-batch merge, extend, filter, and extract
- Official 128-shard BF16 PLE with bounded external artifact resolution and SSD mmap sparse-row reads
- `resident`, `mmap` / `ssd_mmap`, and automatic PLE residency selection
- Pre-quantization key sanitization for MLX-format Qwen4 checkpoints
- Explicit video content-part schema, safe inline base64 decoding, native video placeholder rendering, and pinned video-processor compatibility
- Deterministic model unload that closes external PLE mappings

## Live validation

Validated on an M3 Ultra 256 GB with the Qwen3.8-Flash-Next Q8 compute artifact and external BF16 PLE:

- Text: exact `QWEN_OK`
- Structured tools: valid function name and JSON arguments
- Image: screenshot OCR/understanding returned the expected setting value
- Video: inline `video_url` decoded and completed through native video tokens/features
- Unload: Qwen safetensor file descriptors fell from 33 to 0; server RSS fell from about 44.6 GiB to 740 MiB

## Tests

- Focused Qwen/API/VLM suites: 501 passed
- Full oMLX CI-equivalent run: 10,315 passed, 128 skipped
- Two failures reproduce unchanged on clean `origin/main` with the same pinned environment:
  - Qwen3.5 ANE bank-sliced dequant equality
  - mlx-vlm MTP RNG-state mutation against the guarded MLX build

The commit is SSH-signed and GitHub Verified.

### DIFF
--- a/omlx/api/openai_models.py
+++ b/omlx/api/openai_models.py
@@ -34,6 +34,13 @@ class ImageURL(BaseModel):
     detail: Optional[str] = "auto"  # "low", "high", "auto"
 
 
+class VideoURL(BaseModel):
+    """Inline base64 video data URI for video-capable VLMs."""
+
+    url: str
+    fps: Optional[float] = 2.0
+
+
 class InputAudio(BaseModel):
     """Audio input data for multimodal models (OpenAI format)."""
 
@@ -69,6 +76,7 @@ class ContentPart(BaseModel):
     type: str  # "text", "image_url", "input_audio", or "file"
     text: Optional[str] = None
     image_url: Optional[ImageURL] = None
+    video_url: Optional[VideoURL] = None
     input_audio: Optional[InputAudio] = None
     file: Optional[FileContent] = None
 
--- a/omlx/api/utils.py
+++ b/omlx/api/utils.py
@@ -173,7 +173,7 @@ def _extract_text_from_content_list(content: list) -> str:
 def _extract_multimodal_content_list(content: list) -> list:
     """Extract text, image, and audio parts from a content array.
 
-    Keeps text, image_url, and input_audio items for VLM processing.
+    Keeps text, image_url, video_url, and input_audio items for VLM processing.
     Other content types (tool_use, thinking, refusal, etc.) are dropped.
     """
     parts = []
@@ -238,6 +238,14 @@ def _extract_multimodal_content_list(content: list) -> list:
                             "type": "input_audio",
                             "input_audio": input_audio,
                         }
+                    )
+            elif item_type in ("video_url", "input_video"):
+                video_value = item.get("video_url", item.get("input_video"))
+                if isinstance(video_value, str):
+                    video_value = {"url": video_value}
+                if isinstance(video_value, dict) and video_value.get("url"):
+                    parts.append(
+                        {"type": "video_url", "video_url": video_value}
                     )
     return parts
 
@@ -1261,7 +1269,7 @@ def extract_multimodal_content(
         elif isinstance(content, list):
             # Preserve image_url and input_audio parts for VLM processing
             multimodal_parts = _extract_multimodal_content_list(content)
-            multimodal_types = {"image_url", "input_audio"}
+            multimodal_types = {"image_url", "video_url", "input_audio"}
             has_multimodal = any(
                 p.get("type") in multimodal_types for p in multimodal_parts
             )
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -102,6 +102,7 @@ OCR_EXTRA_STOP_SEQUENCES: List[str] = [
 VLM_LANGUAGE_PROMPT_KWARGS = ("mm_token_type_ids", "token_type_ids")
 
 COHERE2_MOE_MODEL_TYPE = "cohere2_moe"
+QWEN4_EXP_MODEL_TYPE = "qwen4_exp"
 MINIMAX_M3_VL_MODEL_TYPE = "minimax_m3_vl"
 MINIMAX_M3_MODEL_TYPES = {"minimax_m3", MINIMAX_M3_VL_MODEL_TYPE}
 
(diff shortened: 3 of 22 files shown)


pyros-projects:
Real-checkpoint compatibility report from an M5 Max / 128 GB, using `Vontra/Qwen3.8-Flash-Next-MLX-4bit` (its card identifies it as an `mlx-vlm 0.6.3`, MLX 0.32.0 conversion):

1. `auto`/`mmap` currently fails before load. `DiskBackedShardedEmbedding` asks the index for `model.language_model.layers.1.ple.ple_embedding.ngram_embedding.shard_0.weight`; this converted checkpoint stores `language_model.model.layers.1.ple.ple_embedding.ngram_embedding.shard_0.{weight,scales,biases}`. The table is affine Q4/group-32, so a prefix alias alone is not sufficient — the mmap reader needs packed weight + scales + biases, or a clean resident fallback. Blaizzy/mlx-vlm#2045 is now exploring that same quantized external-PLE shape.

2. Forced `OMLX_QWEN4_PLE_MODE=resident` loads (~104.9 GiB peak), but deterministic text is already corrupt; image output is corrupt for the same reason, so this is not yet a vision-only bug. Example: `Reply with exactly QWEN_TEXT_OK` produced `...Ltitle Online`, and `2+2` produced no useful completion.

3. I isolated the text corruption to the norm convention, and reproduced it on current `mlx-vlm` main as well (so it is not just this PR's oMLX glue). This older converted checkpoint stores Qwen4's zero-centered RMSNorm parameters in ordinary MLX multiplicative form (`raw + 1`). For example, `layers.3.self_attn.indexer.q_layernorm.weight` has mean ~0.963, while the new `Qwen4ExpRMSNorm` computes with `1 + weight`, effectively applying the offset twice.

A diagnostic translation that subtracts 1 only for modules backed by `Qwen4ExpRMSNorm` made current `mlx-vlm` return exact deterministic outputs again:

- `Reply with exactly QWEN_TEXT_OK` -> `QWEN_TEXT_OK`
- `What is 2+2? Reply with only the number.` -> `4`

The gated GDN norm must *not* be translated; it already uses the ordinary multiplicative convention. The affected families are the hyper-connection norms, Q/K and indexer Q/K norms, and PLE key/query/conv norms.

I think the durable fix is to support both checkpoint conventions, ideally with an explicit conversion marker for new artifacts plus a conservative compatibility path for existing MLX conversions. Then the vision test is worth rerunning. Tiny synthetic tests all pass today because they do not load a real converted norm inventory — the silicon goblin lives exactly at that boundary.

Our text-first branch already gives the two exact outputs above on this checkpoint. I can factor the convention bridge and quantized-PLE handling onto this PR's architecture rather than preserving a competing implementation.

SneakyBerry:
Awesome work putting together native Qwen3.8-Flash-Next (`qwen4_exp`) support and the video pipeline, @jonathan308! 🚀

I also did MR for Qwen 3.8 Flash Next - https://github.com/jundot/omlx/pull/3172

While testing and benchmarking `Qwen3.8-Flash-Next-MLX-oQ4-MTP` on an M4 Max (128 GB Unified Memory), we ran into a few interesting performance and caching nuances that might be helpful to consider for this PR:

---

### 1. Multi-turn KV Prefix Caching (`QSAKVCache`)

In `qwen4_exp`, the QSA (Qwen Sparse Attention) full-attention layers return a 4-tuple state:
`(keys, values, index_keys, index_position_ids)`.

Because standard `KVCacheHandler` expects a 2-tuple `(keys, values)`, `BlockAwarePrefixCache` currently cannot serialize or slice the QSA layers, which causes multi-turn chat sessions to miss the prefix cache and re-prefill from token 0 on every turn.

Adding a dedicated `QSAKVCacheHandler` in `omlx/cache/type_handlers.py` (slicing keys/values along axis 2, `index_keys` along axis 1, and `index_position_ids` along axis -1) enables full multi-turn prefix caching out of the box. On a 2,500-token multi-turn prompt, this dropped Turn 2 latency from **4.34s down to 1.78s** (reusing 2,048 tokens).

---

### 2. Cold PLE Seek Latency & In-Memory LRU Pre-warming

As noted in mlx-vlm #2045, pure random mmap / disk reads across the 51.2B table on long prefill can add significant seek latency (~9.6s across 65k lookups).

We found that adding a small in-memory LRU cache (~150k entries, ~150–250 MB RAM) paired with a vectorized prefetch of all prompt n-grams during the initial tokenization/prefill step achieves a **94.4%+ LRU hit rate** during decode. This keeps active memory low (~58–70 GB instead of 105 GB) while eliminating disk seek latency on typical conversational and coding turns.

---

### 3. Checkpoint RMSNorm Scaling Parity

In upstream mlx-vlm, `Qwen4ExpRMSNorm` was initially authored with `y * (1.0 + weight)` assuming zero-centered checkpoint weights.

However, in converted MLX checkpoints (e.g. `MLX-oQ4-MTP`), the norm weights are standard ones-centered ($\approx 1.0$). Applying `1 + w` effectively doubles hidden states at every layer ($2^{48}$ scaling over 48 layers), resulting in activation explosion. Ensuring standard $y \cdot w$ scaling ensures clean, coherent generation across both instruct and reasoning prompts.

---

Happy to share our `QSAKVCacheHandler` diff and test cases if you'd like to pull any of this into this PR or coordinate on a follow-up!

jundot:
Thanks @jonathan308 — this provides the right first-class Qwen4-Exp architecture and oMLX serving foundation, including QSA state handling, external PLE resolution, and deterministic mmap teardown. Thanks @pyros-projects and @SneakyBerry as well for the real-checkpoint reports.

I’ll keep the remaining hardening—checkpoint norm compatibility, affine oQ PLE mmap reads, prefix-cache parity, and Lightning MTP—in a maintainer follow-up so this contribution remains focused and attributable. I’ll also narrow the supported input scope to text and image there; video remains outside oMLX’s supported scope for now. LGTM, merging.

PR #3172, opened 2026-08-27
feat(qwen4_exp): tiered SSD n-gram PLE offload + batched QSA decode fixes

SneakyBerry:
## What

Puts the ~100 GB Qwen4-exp PLE n-gram table on SSD and reads only the rows the
model actually looks up, so Qwen3.8-Flash-Next runs on 16–64 GB Apple Silicon boxes.

Two commits:

**`0e6c3f93` feat(qwen4_exp): SSD n-gram tier with LRU hot set in a slot table**
- `NGramSlotCache`: hot set backed by ONE contiguous `[capacity + pinned_cap, dims]`
  table — no per-row dynamic tensors, no gaps. `put`/`fill` scatter eagerly
  (`mx.eval`) and are bounds-checked; the table never grows past capacity+pinned.
- LRU eviction + promote-to-pinned tiers; `clear()` returns every slot to the free list.
- Misses served through `NgramRowReader`: mmap `pread` of packed rows + batched
  `madvise(MADV_WILLNEED)` on coalesced page ranges (one syscall per contiguous run).
- Hot-set budget is a live knob: `ple_hot_set_bytes` → ModelSettings → admin UI →
  `configure_qwen4_exp_runtime`; default 2 GiB.
- Auto SSD-offload: PLE goes resident only when it fits under 70% of physical memory;
  `qwen4_ple_ssd_offload` is tri-state — `true` = force SSD, `false`/`null` = auto.
  Force-resident stays available via `OMLX_QWEN4_PLE_MODE` / checkpoint artifact.
- qwen3.5 (MTP-only, no PLE) keeps the upstream all-RAM path untouched.

**`57119f61` fix(qsa): batched decode offsets, MRoPE cache alignment, Metal recovery**
- Indexer offsets become per-row `(C, 1, block)` on shared backends; fixes mask
  broadcast corruption on text-only rows of multimodal batches
  (`_append_indexer_positions`, `update_indexer` trims `index_offset` before in-place writes).
- `concatenate_states` aligns mixed `C=1`/`C=3` QSA component blocks, so text-only
  turns after multimodal turns reuse their cache instead of forcing full re-prefill
  (`reused=0` symptom, seen live 2026-08-29).
- Scheduler recovers from Metal-resource-limit prefill failures once by recycling the
  engine (gated by a targeted `is_metal_resource_limit_error` detector, not any-Metal-error).

## Reviewer items from earlier revisions — now resolved
- **`hot_set_bytes=0` crash** (KeyError on empty pinned pool, @benwilson): the
  OrderedDict pool is gone; slot table clamps `capacity >= 1`, `pinned_cap >= 1`,
  free-list spans the whole table. Verified non-crashing on the new structure.
- **Silent zero rows when shard resolution fails**: now logs a warning per unresolved
  shard and falls back to the bare mmap embedding when no shard resolves.
- **Dead no-scales path / async prefetcher / usage sidecar / script classifier**:
  cut from this PR entirely (demand-only tier for now). Async prefetch is a planned
  follow-up measured against this baseline, not speculative machinery.

## Validation
- Full local suite: **10559 passed, 0 failed** (incl. `tests/test_qwen4_exp.py`,
  QSA compat, admin/settings, MTP sentinel tests).
- Lint delta vs `main`: **zero new ruff errors**.
- Independent measurements: @benwilson on M5 Pro 64 GiB with REAP-288-MLX-4bit —
  tiered `mmap` works at 38.66 GiB peak where resident OOMs; oQ4 quants fit resident
  and the knob lets users trade RAM headroom vs. prefill speed.

/rebase https://github.com/jundot/omlx/commit/a20ec09d6a0ca68a162c8ced9f12f5de027d9c2f

### DIFF
--- a/benchmarks/bench_qwen4_ple_hotset.py
+++ b/benchmarks/bench_qwen4_ple_hotset.py
@@ -1,0 +1,251 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Bench the tiered hot-set PLE embedding against the bare upstream mmap one.
+
+Run on an idle machine (close other GPU/memory-heavy apps; `sudo purge`
+between passes for true cold numbers). Builds a synthetic affine-U32 n-gram
+table so no model download is needed:
+
+    .venv/bin/python benchmarks/bench_qwen4_ple_hotset.py \
+        --table-gib 4 --hot-mib 512 --passes 6
+
+Point it at a real checkpoint slice instead of synthesizing:
+
+    .venv/bin/python benchmarks/bench_qwen4_ple_hotset.py \
+        --model-path /path/to/qwen4-exp --prefix language_model.model.layers.1...
+
+Reports cold and warm rows/s for upstream vs tier; the tier must not regress
+the cold path and must win the skewed warm path.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import struct
+import time
+from pathlib import Path
+
+import mlx.core as mx
+import numpy as np
+
+
+def _write_affine_table(out_dir: Path, gib: float, dims: int, num_shards: int):
+    prefix = "ple.ngram_embedding"
+    row_bytes = dims // 2 + dims // 32 * 4
+    total_rows = max(int(gib * 1024**3 / row_bytes), num_shards * 1024)
+    per_shard = (total_rows + num_shards - 1) // num_shards
+    rng = np.random.default_rng(1234)
+    packed_cols = dims * 4 // 32
+    scale_cols = dims // 32
+    weight_map = {}
+    written = 0
+    for shard in range(num_shards):
+        rows = min(per_shard, total_rows - written)
+        if rows <= 0:
+            break
+        fname = f"ple_shard_{shard}.safetensors"
+        tensors = {
+            f"{prefix}.shard_{shard}.weight": (
+                rng.integers(0, 2**32, (rows, packed_cols), dtype=np.uint32),
+                "U32",
+            ),
+            f"{prefix}.shard_{shard}.scales": (
+                (rng.random((rows, scale_cols), np.float32).view(np.uint32) >> 16).astype(np.uint16),
+                "BF16",
+            ),
+            f"{prefix}.shard_{shard}.biases": (
+                (rng.random((rows, scale_cols), np.float32).view(np.uint32) >> 16).astype(np.uint16),
+                "BF16",
+            ),
+        }
+        header = {}
+        offset = 0
+        for key, (arr, dtype) in tensors.items():
+            header[key] = {
+                "dtype": dtype,
+                "shape": list(arr.shape),
+                "data_offsets": [offset, offset + arr.nbytes],
+            }
+            offset += arr.nbytes
+        header_bytes = json.dumps(header).encode()
+        pad = (-len(header_bytes)) % 8
+        with open(out_dir / fname, "wb") as f:
+            f.write(struct.pack("<Q", len(header_bytes) + pad))
+            f.write(header_bytes + b" " * pad)
+            for arr, _ in tensors.values():
+                f.write(arr.tobytes())
+        written += rows
+        for suffix in ("weight", "scales", "biases"):
+            weight_map[f"{prefix}.shard_{shard}.{suffix}"] = fname
+    (out_dir / "model.safetensors.index.json").write_text(
+        json.dumps({"weight_map": weight_map})
+    )
+    return prefix, written, dims
+
+
+def _token_stream(total_rows, length, seed, hot_span=0, hot_fraction=0.8):
+    """One shared token id stream, consumed by both arms. Real prompts reuse
+    n-grams hard: `hot_span` rows absorb `hot_fraction` of touches, the rest
+    is a cold Zipf tail. The flat full-vocab Zipf that fooled #3235 is the
+    --hot-span 0 corner."""
+    rng = np.random.default_rng(seed)
+    if hot_span <= 0:
+        raw = rng.zipf(1.02, (length,)).astype(np.int64)
+        return np.clip(raw, 0, total_rows - 1).astype(np.int64)
+    hot_span = min(hot_span, total_rows)
+    mask = rng.random(length) < hot_fraction
+    hot = rng.integers(0, hot_span, (length,), dtype=np.int64)
+    tail = rng.zipf(1.02, (length,)).astype(np.int64)
+    tail = np.clip(tail, 0, total_rows - 1).astype(np.int64)
+    return np.where(mask, hot, tail).astype(np.int64)
+
+
+def _dirty_pressure(gib: float, seed: int):
+    """Allocate and dirty anonymous RAM and HOLD it for the whole run. Living
+    anon pages force the kernel to reclaim memory-mapped file pages, so the
+    upstream arm re-faults cold rows from SSD every call while the hot set,
+    already resident, does not. This is the regime the tier earns its keep in."""
+    if gib <= 0:
+        return None
+    buf = bytearray(int(gib * 1024**3))
+    step = max(len(buf) // 200_000, 1)
+    buf[::step] = b"\x01" * ((len(buf) - 1) // step + 1)
+    return buf
+
+
+def _bench_stream(label, embed, stream, rows_per_call, pressure, seed=5):
+    times = []
+    hits = []
+    for start in range(0, len(stream) - rows_per_call + 1, rows_per_call):
+        ids = mx.array(stream[start : start + rows_per_call], dtype=mx.int64)
+        t0 = time.perf_counter()
+        out = embed(ids)
+        mx.eval(out)
+        times.append(time.perf_counter() - t0)
+        if hasattr(embed, "hot"):
+            hits.append(embed.hot.resident)
+    times = np.asarray(times)
+    half = max(len(times) // 2, 1)
+    early = times[:half].mean()
+    steady = times[half:].mean() if len(times) > half else early
+    rps = rows_per_call / steady
+    print(
+        f"{label:22s} calls={len(times):4d} | early {early*1e3:7.2f} ms | "
+        f"steady {steady*1e3:7.2f} ms ({rps:9.0f} rows/s)"
+    )
+    if hits:
+        print(f"{'':22s} hot residency: start {hits[0]} -> end {hits[-1]}")
+    del pressure
+    return times
+
+
+def main():
+    ap = argparse.ArgumentParser()
+    ap.add_argument("--table-gib", type=float, default=4.0)
+    ap.add_argument("--hot-mib", type=float, default=512.0)
+    ap.add_argument("--dims", type=int, default=512)
+    ap.add_argument("--num-shards", type=int, default=2)
+    ap.add_argument("--rows-per-call", type=int, default=64)
+    ap.add_argument("--calls", type=int, default=512)
+    ap.add_argument(
+        "--pressure-gib",
+        type=float,
+        default=0.0,
+        help="dirty+hold this much anonymous RAM for the whole run to force"
+        " page-cache pressure (the regime where the hot set earns its keep)",
+    )
+    ap.add_argument(
+        "--arm",
+        choices=["both", "upstream", "tier"],
+        default="both",
+        help="run one arm per process; 'sudo purge' between arms for clean"
+        " cold numbers (per #3235 protocol)",
+    )
+    ap.add_argument("--seed", type=int, default=5)
+    ap.add_argument(
+        "--hot-span",
+        type=int,
+        default=0,
+        help="rows that absorb --hot-fraction of touches (0 = flat Zipf)",
+    )
+    ap.add_argument("--hot-fraction", type=float, default=0.8)
+    ap.add_argument("--model-path", default=None)
+    ap.add_argument("--prefix", default=None)
+    args = ap.parse_args()
+
+    import tempfile
+
+    from omlx.patches.mlx_vlm_qwen4_exp_compat import (
+        apply_mlx_vlm_qwen4_exp_compat_patch,
+    )
+    from omlx.patches.qwen4_exp import DiskBackedShardedEmbedding as Tier
+
+    tmp = Path(tempfile.mkdtemp(prefix="omlx-ple-bench-"))
+    if args.model_path:
+        # read geometry straight from the real index
+        index = json.loads(
+            (Path(args.model_path) / "model.safetensors.index.json").read_text()
+        )
+        prefix = args.prefix
+        weights = [k for k in index["weight_map"] if k.startswith(prefix)]
+        dims = args.dims
+        total_rows = 0
+        for w in weights:
+            if w.endswith(".weight"):
+                with open(Path(args.model_path) / index["weight_map"][w], "rb") as f:
+                    (hlen,) = struct.unpack("<Q", f.read(8))
+                    hdr = json.loads(f.read(hlen).decode().rstrip())
+                    total_rows += hdr[w]["shape"][0]
+        num_shards = sum(1 for w in weights if w.endswith(".weight"))
+    else:
+        prefix, total_rows, dims = _write_affine_table(
+            tmp, args.table_gib, args.dims, args.num_shards
+        )
+        num_shards = args.num_shards
+        print(f"synthetic table: {total_rows} rows x {dims} (~{args.table_gib} GiB)")
+
+    apply_mlx_vlm_qwen4_exp_compat_patch()
+    import mlx_vlm.models.qwen4_exp.language as language
+
+    stream = _token_stream(
+        total_rows,
+        args.calls * args.rows_per_call,
+        args.seed,
+        args.hot_span,
+        args.hot_fraction,
+    )
+    common = (
+        tmp if not args.model_path else Path(args.model_path),
+        prefix,
+        total_rows,
+        dims,
+        num_shards,
+    )
+    print(f"access: {args.calls} calls x {args.rows_per_call} rows, "
+          f"pressure {args.pressure_gib} GiB, arm={args.arm}")
+    if args.arm == "both":
+        print("NOTE: one process warms the cache for the later arm; for clean"
+              " cold numbers run --arm upstream, 'sudo purge', then --arm tier.")
+    pressure = _dirty_pressure(args.pressure_gib, args.seed)
+
+    if args.arm in ("both", "upstream"):
+        bare = language.DiskBackedShardedEmbedding(*common)
+        _bench_stream(
+            "upstream", bare, stream, args.rows_per_call, pressure, args.seed
+        )
+        bare.close()
+        del bare
+
+    if args.arm in ("both", "tier"):
+        hot_bytes = int(args.hot_mib * 1024**2)
+        tier = Tier(*common, hot_set_bytes=hot_bytes)
+        print(f"hot set budget: {tier._max_entries} rows ({hot_bytes / 1024**2:.0f} MiB)")
+        _bench_stream(
+            "tier", tier, stream, args.rows_per_call, pressure, args.seed
+        )
+        tier.close()
+        del tier
+
+
+if __name__ == "__main__":
+    main()
--- a/benchmarks/qwen4_ple_hotset_e2e.py
+++ b/benchmarks/qwen4_ple_hotset_e2e.py
@@ -1,0 +1,225 @@
+#!/usr/bin/env python3
+"""Cold-prefill/decode A/B harness for the Qwen4 exp PLE hot-set.
+
+Protocol (from mlx-community/qwen4_exp#3235): config edit, purge, fresh
+server per trial, real long prompt, alternating arms, identical prompts.
+
+Arm "off": ple_hot_set_bytes=0 (bare mmap). Arm "on": --hot-bytes.
+Run while the machine is idle. Paste the JSONL + summary back to the PR.
+"""
+
+import argparse
+import datetime
+import json
+import subprocess
+import sys
+import time
+import urllib.request
+from pathlib import Path
+
+API_HEADERS = {}
+REPO_VENV = [sys.executable, "-m", "omlx.cli"]
+
+
+def http_get(url, timeout=10):
+    req = urllib.request.Request(url, headers=API_HEADERS)
+    return json.loads(urllib.request.urlopen(req, timeout=timeout).read())
+
+
+def set_arm(model_key, hot_bytes):
+    path = Path.home() / ".omlx" / "model_settings.json"
+    data = json.loads(path.read_text())
+    models = data.get("models", data)
+    models[model_key]["qwen4_ple_ssd_offload"] = True
+    models[model_key]["ple_hot_set_bytes"] = hot_bytes
+    path.write_text(json.dumps(data, indent=2))
+
+
+def start_server(model_dir, port, log_path):
+    with open(log_path, "ab") as log:
+        proc = subprocess.Popen(
+            [*REPO_VENV, "serve", "--model-dir", str(model_dir),
+             "--port", str(port)],
+            stdout=log, stderr=subprocess.STDOUT, start_new_session=True,
+        )
+    deadline = time.time() + 600
+    while time.time() < deadline:
+        try:
+            if http_get(f"http://127.0.0.1:{port}/health").get("status") in (
+                "ok", "healthy",
+            ):
+                return proc
+        except Exception:
+            time.sleep(2)
+    raise RuntimeError("server did not become healthy in 600s")
+
+
+def stop_server(proc):
+    subprocess.run(["pkill", "-TERM", "-P", str(proc.pid)], check=False)
+    proc.terminate()
+    try:
+        proc.wait(timeout=60)
+    except subprocess.TimeoutExpired:
+        proc.kill()
+        proc.wait()
+    subprocess.run(["pkill", "-f", "omlx serve"], check=False)
+    time.sleep(5)
+
+
+def run_request(port, model, prompt, max_tokens):
+    body = {
+        "model": model,
+        "prompt": prompt,
+        "max_tokens": max_tokens,
+        "temperature": 0.0,
+        "stream": True,
+        "stream_options": {"include_usage": True},
+    }
+    req = urllib.request.Request(
+        f"http://127.0.0.1:{port}/v1/completions",
+        data=json.dumps(body).encode(),
+        headers={"Content-Type": "application/json", **API_HEADERS},
+    )
+    t0 = time.perf_counter()
+    t_first = t_last = None
+    usage = {}
+    with urllib.request.urlopen(req, timeout=1800) as resp:
+        for raw in resp:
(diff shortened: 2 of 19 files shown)


gearsandcode:
Really excited for this! Thank you for your work.

zhnext:
Great job! I ran oQ3-MTP with you excellent work and the following three modifications in omlx/patches/qwen4_exp.py:


```
@@ -107,6 +107,7 @@ class DiskBackedShardedEmbedding(nn.Module):
         self.lru_cache = NGramLRUCache(max_entries=lru_cache_size)
         self.pinned_pool = NGramPinnedPool()
         self._handles: dict[str, Any] = {}
+        self._quant_params_cache: dict[int, tuple[int, int]] = {}
         self.shard_info: dict[int, tuple[str, str, str | None, str | None]] = {}
```

```
@@ -133,6 +134,51 @@ class DiskBackedShardedEmbedding(nn.Module):
             self._handles[fpath] = safe_open(fpath, framework="pt")
         return self._handles[fpath]
 
+    def _quant_params(
+        self, shard_id: int, handle: Any, w_key: str, s_key: str
+    ) -> tuple[int, int]:
+        """Derive (group_size, bits) from this shard's stored shapes.
+
+        The checkpoint's quantization is not fixed at 4 bits: oQ2/oQ3/oQ4
+        builds pack the same N-gram table at different widths, and a
+        mixed-quant checkpoint may vary it per shard. A packed row holds
+        ``dims * bits / 32`` uint32 words and ``dims / group_size`` scale
+        groups, so both fall out of the two widths without parsing config.
+        """
+        cached = self._quant_params_cache.get(shard_id)
+        if cached is not None:
+            return cached
+
+        packed_cols = int(handle.get_slice(w_key).get_shape()[1])
+        scale_cols = int(handle.get_slice(s_key).get_shape()[1])
+        if scale_cols <= 0 or self.dims <= 0:
+            params = (32, 4)
+        else:
+            group_size = self.dims // scale_cols
+            bits = packed_cols * 32 // self.dims
+            if group_size <= 0 or bits <= 0:
+                logger.warning(
+                    "Unusable N-gram quantization geometry for shard %d "
+                    "(dims=%d, packed_cols=%d, scale_cols=%d); "
+                    "falling back to group_size=32, bits=4",
+                    shard_id,
+                    self.dims,
+                    packed_cols,
+                    scale_cols,
+                )
+                params = (32, 4)
+            else:
+                params = (group_size, bits)
+                logger.debug(
+                    "N-gram shard %d quantization: group_size=%d, bits=%d",
+                    shard_id,
+                    group_size,
+                    bits,
+                )
+
+        self._quant_params_cache[shard_id] = params
+        return params
+
```

```
@@ -150,7 +196,8 @@ class DiskBackedShardedEmbedding(nn.Module):
             b_slice = handle.get_slice(b_key)[local_idx : local_idx + 1]
             s_mx = mx.array(s_slice.float().numpy()).astype(mx.bfloat16)
             b_mx = mx.array(b_slice.float().numpy()).astype(mx.bfloat16)
-            return mx.dequantize(w_mx, s_mx, b_mx, group_size=32, bits=4)[0]
+            group_size, bits = self._quant_params(shard_id, handle, w_key, s_key)
+            return mx.dequantize(w_mx, s_mx, b_mx, group_size=group_size, bits=bits)[0]
```

zhnext:
@jundot ngram ssd offload is so important to have

SneakyBerry:
@zhnext I'm working on it. If you have any ideas, feel free to jump in and help. DM me if you'd like to work on it together — I'd be happy to stay in close contact.
@Bear_Metal : telegram

SneakyBerry:
@zhnext thanks a lot for your contribution!

@jundot please review the following changes. I got 76.40 Gb on oQ4 quanted model so offload should be working.
Also please check your email (from profile), i'm waiting for your response

SneakyBerry:
# Follow-up: known open items in this branch

Leaving these open deliberately — would appreciate eyes, especially where marked **help wanted**.

## Correctness-ish

1. **Silent zero-rows when shard resolution fails** (`omlx/patches/qwen4_exp.py:421-443`).
   If `model.safetensors.index.json` is missing or the PLE keys don't match any
   `shard_N`/`shards.N` spelling, `_reader_for()` returns `None` and `__call__`
   serves **zero rows** — no warning, no error. Outputs silently degrade.
   Plan: log a warning per unresolved shard and fail (or fall back to the bare
   mmap class) when *no* shard resolves. Would rather not ship silent zeros.

2. **Dead no-scales path** (`ngram_prefetch.py:451`): shards without `scales`
   dequantize via `U32 → F16` cast, which is numerically meaningless. Unreachable
   with q4/fp8 checkpoints, but if we keep the branch it should raise, not
   compute garbage.

## Perf, not yet benchmarked (main honesty note)

3. **Speed is unmeasured so far.** Memory-only result: 106.22 → 76.40 GB resident,
   live server, oQ4 checkpoint. Throughput A/B (tiered vs resident, Q4 + Q6) is
   planned on this machine. Two hot-path costs that A/B must attribute:
   - hot-set cap is actually LRU(full `hot_set_bytes`) + pinned(half) ≈ **3 GiB
     raw** with per-mx-array overhead pushing real RSS toward ~4 GiB
     (`qwen4_exp.py:167-169`) — the "2 GiB" label is optimistic;
   - prefetch enqueue runs on the **engine thread** before the chunk forward is
     submitted (`scheduler.py:3568`, `scheduler.py:5154`,
     `NgramPrefetcher.prefetch_seq` → `_prefill_indices`): host-side `.tolist()`
     syncs + a Python hash loop (~154k pushes/32k enqueue per 2048-token chunk).
     Expected to be hidden by the GPU compute it overlaps with, but *unmeasured*.
   - `DiskBackedShardedEmbedding._shard_lock` (RLock) is held across a full
     gather batch; cold misses block on prefetch workers while the batch is in
     flight. Bounded by design (≤128 rows/shard split), but measurable.

4. **Script filter is a placeholder in prod**: `script_id_map=np.full(vocab,
   SCRIPT_ASCII)` (`vendor .../qwen4_exp/language.py:1376`) — the CJK/Devanagari
   skip is *designed and unit-tested* but inert until the engine installs the
   tokenizer-derived `ScriptClassifier` map (next PR). Same for
   `NgramPrefetcher()`'s default `script_id_map=None` (RuntimeError unless a map
   or tokenizer is passed — fine for prod call sites, a trap otherwise).

5. **Single-ple_layer hook**: the MTP verify lookahead reads cache slot 3 from the
   first `ple` layer only (`ple_prev_context`). Fine for this checkpoint
   (`ple_layer_index` fixed), not general — noted for the record.

## Scope note

The earlier chunk (QSA cache handler, RMSNorm fix, mlx-vlm bump) is already on
`main` (#3200 etc.), which is why the diff here is *only* the offload tier —
not because the first draft PR was closed.

## Repro of what *was* measured

- memory: live server, `Qwen3.8-Flash-Next-MLX-oQ4-MTP`, footprint via
  `phys_footprint` before/after (106.22 → 76.40 GB);
- correctness: `pytest tests/test_qwen4_exp.py tests/test_mlx_vlm_qwen4_exp_compat.py
  tests/test_engine_pool.py tests/test_admin_model_settings.py
  tests/test_vlm_qwen4_exp_loader.py` — 200 passed; full suite green pre-push.
  Includes a ground-truth test that runs the vendored `Qwen4ExpNGramEmbedding`
  as oracle over hand-written safetensors shards (no model download needed).

benwilson:
Tested this branch (5609aa851) on the 64 GB end of the stated 16–64 GB target.
This is aimed at your open item 3 — "speed is unmeasured so far" — plus three
defects that aren't on your list.

**Setup.** M5 Pro, 64 GiB unified, macOS 26.5, `iogpu.wired_limit_mb=56320`,
py3.11, `pip install -e .`. Model is `sh0wie/Qwen3.8-Flash-Next-REAP-288-MLX-4bit`
— a **REAP-pruned 288-expert** checkpoint, 68.5 GiB, 131 shards. That makes it a
second, independent checkpoint shape from your oQ4 measurement. Driven in-process
rather than through `omlx serve`, so the scheduler is out of the timings and I can
read the LRU counters directly. `tests/test_qwen4_exp.py` 14/14 here.

## Your item 3: the tiered-vs-resident A/B

On this machine the A/B has only one outcome, and it's the strongest argument for
the PR:

| PLE mode | peak memory | outcome |
|---|---|---|
| `mmap` (tiered) | **38.66 GiB** | works — coherent, 28.6 tok/s decode |
| `resident` (forced) | **68.46 GiB** | **Metal OOM at first generation** |

So on 64 GB this is not an optimization, it's the difference between running and
not running. Your `mmap_bytes` estimator predicted 41.5 GiB; an independent
calculation from the safetensors headers predicted 39.6 GiB; measured 38.66 GiB.

**Throughput.** Short prompt (192 tok, 2 reps, one process):

| hot_set_bytes | prefill cold → warm | decode | LRU on warm rep |
|---|---|---|---|
| 512 MiB | 11.8 → 175.9 | 28.3 | 3728 hits / **0 misses** |
| 2 GiB (default) | 30.0 → 176.5 | 28.6 | 3728 hits / **0 misses** |
| 8 GiB | 12.0 → 173.5 | 28.3 | 3728 hits / **0 misses** |

Context sweep at the 2 GiB default, varied real corpus (prose + C++ source, never
repeated filler, so n-gram diversity is genuine; prompt sizes from `prompt_tokens`):

| ctx | prompt tok | prefill tok/s | decode tok/s | MLX peak | LRU hit rate |
|---|---|---|---|---|---|
| 4k | 4,122 | 189.1 | 25.2 | 41.5 GiB | 0.055 |
| 16k | 16,410 | **315.6** | 24.3 | 43.1 GiB | 0.142 |
| 32k | 32,794 | 313.9 | 21.7 | 45.3 GiB | 0.212 |
| 64k | 65,562 | 244.1 | 16.2 | 49.3 GiB | 0.576 |
| 131k | 131,098 | 178.7 | 14.1 | 57.27 GiB | 0.613 |

131k still produces *correct* output — asked to summarise 131k tokens of a research
repo it named the repo's actual conclusion — but it runs above the wired limit on
swap. Practical range here: 32–64k comfortable, ~131k the ceiling. I stopped before
262k: peak memory grows ~128 KiB/token (KV alone is 24 KiB/token), projecting 74 GiB.

Two caveats on these numbers. The cold-prefill column is noisy — the 30.0 outlier
was simply the first run after download, with weights still warm in page cache; I
would not read a hot-set effect into it. And the hot-set *size* is genuinely
unmeasurable at short context, for the reason in the sizing note below.

## Three defects not on your list

**1. `hot_set_bytes=0` crashes on the first token** — `qwen4_exp.py:167-169`, the
same lines your item 3 cites for the cap arithmetic.

```
File "omlx/patches/qwen4_exp.py", line 236, in _bump
    self.pinned_pool.put(gram_id, row)
File "omlx/patches/qwen4_exp.py", line 104, in put
    self._pool.popitem(last=False)
KeyError: 'dictionary is empty'
```

`_max_entries` is clamped to `max(..., 1)` (line 167), but the pinned pool takes
`self._max_entries // 2` (line 169) with no clamp → 0. Then in `put`,
`len(self._pool) >= self.max_entries` is `0 >= 0` → True → `popitem` on an empty
OrderedDict. Fires for any `hot_set_bytes < 640`. Zero is the natural way to A/B
the tier off, which is exactly how I hit it. Clamping `max_entries` to ≥1, or
guarding the popitem, fixes it.

**2. The usage sidecar never persists for short-lived processes.**
`.omlx_ple_usage.npz` was never written in any run here. `_save_loop`
(`qwen4_exp.py:205`) is a **daemon** thread on `_SAVER_INTERVAL = 60.0` with no
`atexit`, no `__del__`, no flush on close — daemon threads are killed abruptly at
interpreter exit. Any process under 60 s persists nothing; longer ones lose up to
60 s of learned pinning on every clean shutdown. That undercuts "counts persist ...
so the hot set warms itself on the next load" for CLI generates and fast restarts.
A flush in a `close()`/`atexit` path would cover it.

**3. Forced-resident OOM presents as a tokenizer bug.** The first exception raised
is `IndexError: list index out of range` at `mlx_vlm/tokenizer_utils.py:232`
(`self.tokenmap[token]`, garbage token id). The real error —
`RuntimeError: [METAL] Command buffer execution failed: Insufficient Memory
(kIOGPUCommandBufferCallbackErrorOutOfMemory)` — only appears afterwards from
`mx.synchronize` in `wired_limit`'s contextmanager exit. Not your code, but it is
what users on this hardware will hit first if they force resident, and it points
away from the real cause.

## Sizing, refining your item 3

You note the "2 GiB" label is optimistic — LRU(full) + pinned(half) ≈ 3 GiB raw,
real RSS toward ~4 GiB. Measured working set here is much smaller than either:

| context | unique rows resident | working set |
|---|---|---|
| 192-token prompt | 3,240 | ~1 MB |
| 64k | 403,338 | 129 MB |
| 131k | 1,083,331 | **331 MB** |

So at the 2 GiB default the effective allocation is roughly **12× the peak working
set I could produce**, and 512 MiB measured identically to 2 GiB and 8 GiB at every
context tested. On 16 GB machines — where the tier matters most — that headroom is
worth reclaiming. I have not tested below 512 MiB, per defect 1.

This is also why the hot-set sweep can't discriminate at short prompts: a 1 MB
working set against a 518×-larger cache. Only long context makes the knob legible,
and even at 131k it isn't saturated.

## Two positive datapoints

- **Your item 1 (silent zero-rows) did not fire on this checkpoint.** The
  `shards.N` spelling in this REAP-288 conversion resolved correctly,
  `tier_found=True`, and output was coherent at every context. A second checkpoint
  shape that resolves cleanly — though I agree the silent-zeros failure mode is
  worth closing, since nothing in my run would have told me if it had happened.
- **Your item 4 (script filter) was not exercised** — my corpus is ASCII prose and
  C++/Python source, so the CJK/Devanagari path never ran.

Happy to re-run any of this, test a fix, or run the 512 MiB-vs-2 GiB A/B on a
16 GB-constrained configuration if that's more useful.

SneakyBerry:
> ## Three defects not on your list
> **1. `hot_set_bytes=0` crashes on the first token** — `qwen4_exp.py:167-169`, the same lines your item 3 cites for the cap arithmetic.
> 
> ```
> File "omlx/patches/qwen4_exp.py", line 236, in _bump
>     self.pinned_pool.put(gram_id, row)
> File "omlx/patches/qwen4_exp.py", line 104, in put
>     self._pool.popitem(last=False)
> KeyError: 'dictionary is empty'
> ```
> 
> `_max_entries` is clamped to `max(..., 1)` (line 167), but the pinned pool takes `self._max_entries // 2` (line 169) with no clamp → 0. Then in `put`, `len(self._pool) >= self.max_entries` is `0 >= 0` → True → `popitem` on an empty OrderedDict. Fires for any `hot_set_bytes < 640`. Zero is the natural way to A/B the tier off, which is exactly how I hit it. Clamping `max_entries` to ≥1, or guarding the popitem, fixes it.
> 
> **2. The usage sidecar never persists for short-lived processes.** `.omlx_ple_usage.npz` was never written in any run here. `_save_loop` (`qwen4_exp.py:205`) is a **daemon** thread on `_SAVER_INTERVAL = 60.0` with no `atexit`, no `__del__`, no flush on close — daemon threads are killed abruptly at interpreter exit. Any process under 60 s persists nothing; longer ones lose up to 60 s of learned pinning on every clean shutdown. That undercuts "counts persist ... so the hot set warms itself on the next load" for CLI generates and fast restarts. A flush in a `close()`/`atexit` path would cover it.
> 
> **3. Forced-resident OOM presents as a tokenizer bug.** The first exception raised is `IndexError: list index out of range` at `mlx_vlm/tokenizer_utils.py:232` (`self.tokenmap[token]`, garbage token id). The real error — `RuntimeError: [METAL] Command buffer execution failed: Insufficient Memory (kIOGPUCommandBufferCallbackErrorOutOfMemory)` — only appears afterwards from `mx.synchronize` in `wired_limit`'s contextmanager exit. Not your code, but it is what users on this hardware will hit first if they force resident, and it points away from the real cause.

1 and 2 fixed, 3 is an upstrem mlx-vlm problem. Feel free to test

SneakyBerry:
Rebased onto `main` @ `a20ec09d` and groomed into two commits — `0e6c3f93` (tier) + `57119f61` (QSA/cache fixes). Full suite green locally (10559 passed, 0 failed), zero new lint findings.

Status of the defects raised in review:

**@benwilson — three defects:**
1. `hot_set_bytes=0` KeyError — **fixed**. The slot cache was rewritten: pinned pool is now a slot-range tier with `pinned_cap = max(capacity // 2, 1)` and the free list spans `capacity + pinned_cap`, so the `popitem()`-on-empty path no longer exists. Verified `hot_set_bytes=0` end-to-end on the new structure (smoke: capacity/pinned clamp to 1, `clear()` restores all rows).
2. Usage sidecar never persisting on short runs — **moot**: the persistence sidecar (and the async prefetcher, script classifier, and the numeric-no-op no-scales path) are **cut from this PR**. The tier is demand-only: LRU hot set + mmap row reader with batched `madvise`. If cold-start cost shows up in measurements, prefetch returns as a measured follow-up, not speculative machinery.
3. Speed A/B — appreciated the numbers; the `ple_hot_set_bytes` knob is now live in model settings/admin so the same A/B is one setting away from reproducing.

**Earlier thread (silent zeros / dead paths):**
- Unresolved-shard silent zero rows — now warns per unresolved shard and falls back to the bare mmap embedding when no shard resolves at all.
- `ngram_prefetch.py` dead no-scales `U32→F16` branch — removed along with the module.

Marking ready for review.

jundot:
Thanks for continuing to refine this and for addressing the earlier feedback. I reviewed the latest head against current `main`.

The SSD mmap row-loading path, QSA late-join handling, and Metal resource-count prevention are already on `main`. The main new value here is the hot-row cache with batched `madvise` and the mixed C=1/C=3 MRoPE prefix-cache fix.

| Area | Finding |
| --- | --- |
| PLE reader | The tiered reader crashes on BF16 because it assumes U32 affine rows. Missing shards also return zero rows instead of falling back or failing closed, and the existing `close()` path is lost. |
| Hot-set setting | A 2 GiB setting allocates about 3 GiB, is not included in admission estimates, does not trigger reload, and is not wired to either UI. `None` also retains the previous model's global value. |
| Validation | The current benchmarks compare resident against mmap, but mmap is already on `main`. The useful comparison is current bare mmap against this tiered mmap path. |
| QSA/MTP | The per-row QSA branch and Qwen4 capture guard are unreachable behind existing logic. The mixed MRoPE cache alignment is a real new fix and is worth keeping. |

I reproduced the PLE regressions through the actual tier builder: BF16 fails on the first lookup, and a missing shard serves a zero embedding row. The focused tests pass because the tier tests use affine q4 fixtures while the BF16 test exercises the old vendor reader directly.

I suggest building the cache on top of the existing `DiskBackedShardedEmbedding` so BF16, FP8, oQ2-8, shard validation, and `close()` remain intact. The mixed MRoPE fix can be kept as a focused change, while the duplicate QSA/MTP code should be removed.

I cannot merge the current head because the PLE failures affect correctness. Once those are fixed and the performance comparison is against `main`, I would be happy to review it again.

SneakyBerry:
Thanks for the careful review — I reproduced all four findings against the current head and they all hold. Agree with the suggested shape; here's the plan:

**1. Tier on top of upstream `DiskBackedShardedEmbedding` (correctness fix)**
- Drop our custom `NgramRowReader` entirely — the affine-U32-only assumption breaks BF16, and the zero-row fallback on missing shards is fail-open. Wrong direction.
- The hot-row cache becomes an overlay: subclass delegates all row reads to the upstream `_SafeTensorMMap.rows` path, so BF16/F16/F32/U32/F8_E4M3, layout validation, and `close()` stay intact and fail closed.
- Batched `madvise(MADV_WILLNEED)` prefetch survives as an advisory-only hook on the upstream mmap handle — correctness never depends on it.

**2. Honest hot-set budget**
- `ple_hot_set_bytes` will bound the *entire* table including the pinned tier (`(capacity + pinned) × row_bytes ≤ budget`), instead of the current ~1.5×.
- Wire it into the admission estimate and mark it restart-required so a change reloads the engine.
- `configure_ple_hot_set(None)` will reset to the default instead of inheriting the previous model's global.

**3. Remove duplicated QSA/MTP code**
- The per-row QSA offset branch and the qwen3.5 capture guard are unreachable behind #3246's aligned-column join — dropping both.
- Keeping the mixed C=1/C=3 MRoPE prefix-cache alignment as the focused QSA fix.

**4. Validation against `main`, not resident**
- Re-benchmark tiered-over-upstream-mmap vs plain mmap on the same checkpoint: hit rate, decode tok/s, dirty/WAL footprint.

Reorganizing the branch now; will push the reworked commits here.

pedroberaldo87:
Independent reproduction of the batched-mask half, on an M1 Ultra 128GB with `Qwen3.8-Flash-Next-oQ4e-mtp`.

Two concurrent requests whose prompts cross ~2048 tokens of context abort each other, intermittently, with a 500 on every live stream:

```
RuntimeError: [concatenate] All the input array dimensions must match exactly
except for the concatenation axis. However, the provided shapes are
(2,2,2536), (2,1,3), and the concatenation axis is -1.
```

The threshold matches your per-row diagnosis: `indexer_budget` 2048 / `compress_ratio` 4 = 512 blocks, so the sparse branch only runs above ~2048 tokens of context. Prompts of 250 tokens pass; prompts of 2500 fail 2 in 6.

I had a local patch clamping the selection to the current query length, which stops the crash but leaves the inflated axis being produced in the first place. Your `per_row` handling of `past_len` fixes it where it starts, so I am not proposing mine — this is just the confirmation that the failure is real outside your machine, and that the sparse-path threshold is where it bites.

One thing that may be worth carrying over: I pinned the ordering with a test, because the earlier attempt at this in #3264 read `selected_blocks` 39 lines before the `argpartition` that creates it, so the guard would have raised `NameError` the first time it fired. An AST assertion that the guard sits after the assignment is cheap and would have caught that.

SneakyBerry:
Rework pushed (rebased on 0.6.4 / `e008a66b`, two commits: `16a12ea8` + `61590742`).

**PLE reader (maintainer repro)** — dropped the custom `NgramRowReader` and `ngram_shards.py` entirely. `DiskBackedShardedEmbedding` in `omlx/patches/qwen4_exp.py` is now a subclass of the upstream mmap embedding: row reads delegate to `super().__call__` (dtype-aware BF16/FP8/oQ2-8, shard validation, close all inherited). Added tests: BF16 dense checkpoint serves correct rows (the exact repro), missing shard fails closed with `KeyError` instead of zero-filling, row-count mismatch raises at construction.

**Hot-set cache** — now an overlay: one preallocated `[capacity, dims]` bf16 slot table, honest budget `(lru + pinned) * row_bytes <= ple_hot_set_bytes` with capacity also capped at `num_embeddings` (no over-allocation past the table itself). Hits via `mx.take`, misses via upstream gather then fill; `weight_scale` stays an inherited top-level param so strict `load_weights` works unchanged.

**madvise** — advisory page-range prefetch only, coalesced, swallowed on failure, never affects outputs.

**Hot-set wiring** — the table is charged to admission: `mmap_bytes + min(hot_set_bytes, ple_bytes)` in `_entry_runtime_residency_size`, and `ple_hot_set_bytes` added to the engine runtime signature so changing it reloads the engine. `configure_ple_hot_set(None)` resets to the 2 GiB default instead of inheriting the previous model's value.

**QSA/MTP** — removed the dead per-row indexer branch and the `update_indexer` pre-slice (both unreachable behind upstream aligned-column semantics post #3244/#3283). Kept the mixed `C=1/C=3` MRoPE fix in `_append_indexer_positions` (runtime concat path, not serialized reconstruction). Also removed my now-superseded `_align_qsa_position_components` in `type_handlers.py` — upstream `_normalize_qsa_position_states` (#3219) covers it exactly.

**Metal** — kept the allocator-failure retry/sweep as a safety net (upstream `_requeue_or_fail_prefill` still only byte-matches `Memory limit exceeded` as of `e008a66b`); happy to drop if you'd rather fold it upstream first.

Suite: 10690 passed, 0 failed; ruff delta vs main is zero new findings. Real-model gather-vs-bench I'll post separately once I have a quiet machine — numbers from this box while other work is running would be noise.

SneakyBerry:
**Follow-up: dropped the madvise prefetch after studying #3235** (`9467ea29` / `65e9fc78`).

#3235's own re-test on 0.6.4 is convincing: #3244's gathered QSA already batches the rows through the Metal pipeline, so `madvise` from userspace competes for bandwidth instead of buying latency, and the PR was closed at -22% on 12.2k. My inline `MADV_WILLNEED`-on-miss is the same class of hint against the same pipeline, so it's gone — the tier is now purely RAM-pinned hot rows.

The rest of #3235's methodology I adopted verbatim: uniform-row micro-benchmarks don't predict end-to-end (they sample rows real prompts never touch); clean numbers need purge + fresh process + real prompts, alternating arms.

**Proposed end-to-end A/B for the remaining value (hot-set under memory pressure)** on `Qwen3.8-Flash-Next-oQ4e-mtp` (103.94 GB, PLE ~35 GB), idle M4 Max 128 GB:

```
# arm A (bare mmap)
sudo purge && omlx serve ... --ple-hot-set-bytes 0 ...   # 2 fresh trials, 4096-token prompt
# arm B (tier)
sudo purge && omlx serve ... --ple-hot-set-bytes 536870912 ...
# plus the warm-cache control: arm B twice in a row — must measure identical
```

Prediction, stated honestly: on 128 GB with ~35 GB PLE the OS page cache already covers most repeats, so I expect arm B ≈ arm A on this box, with the gap opening only on smaller machines or long contexts where KV churn evicts mmap pages. If A/B shows no difference and you can't see the smaller-hardware case mattering, I'd rather close #3172 with the MRoPE alignment fix extracted into its own small PR than carry an unproven cache. If you have (or want) a 64 GB test box, that's the machine where this either earns its keep or doesn't.

Will run the A/B on my box once idle and post numbers either way.

SneakyBerry:
Correction to the protocol above — there is no `--ple-hot-set-bytes` CLI flag; the setting is per-model via the admin API (it's in the engine runtime signature, so the change reloads that model):

```
# arm A (bare mmap): ple_hot_set_bytes=0
curl -X POST localhost:5599/api/models/settings -H 'content-type: application/json'   -d '{"ple_hot_set_bytes": 0}'
sudo purge   # then fresh trials

# arm B (tier): 512 MiB
curl -X POST localhost:5599/api/models/settings -H 'content-type: application/json'   -d '{"ple_hot_set_bytes": 536870912}'
sudo purge
```

(Endpoint name per your routes; arm switch = same request body change, engine reloads between arms via the signature.)

SneakyBerry:
Added `benchmarks/qwen4_ple_hotset_e2e.py` — implements the #3235 protocol end-to-end: alternating arms, `sudo purge` + fresh server process per trial, per-model settings edit (`ple_hot_set_bytes` 0 vs 512 MiB), streamed 4k-token cold prefill + 256-token decode with `stream_options.include_usage`, RSS/free-RAM snapshots, JSONL out + median summary.

```
sudo python benchmarks/qwen4_ple_hotset_e2e.py --trials 3
```

SneakyBerry:
Same as #3235 closed in favor of #3244

PR #3180, opened 2026-08-27
fix(load): chunk materialize_lazy_state eval to avoid Metal GPU watchdog on large checkpoints

### LINKED ISSUE #3179: Large BF16 checkpoint load aborts with Metal GPU command-buffer timeout: single unchunked mx.eval over all params
## Summary

Loading a large **BF16** checkpoint aborts the process with a Metal GPU
command-buffer timeout during the post-load parameter materialization, before
the first request. The fatal call is a single unchunked
`mx.eval(<all parameter arrays>)` — both mlx-vlm's `load()`
(`mx.eval(model.language_model.parameters())`) and oMLX's
`omlx/utils/model_loading.py::materialize_lazy_state` (`mx.eval(arrays)`,
currently line ~1140) submit one Metal command buffer that spans every weight
array. On a multi-hundred-GB model that single buffer exceeds the GPU
command-buffer watchdog and the runtime aborts with SIGABRT.

This is distinct from #2098 (NFS/USB storage) and #2029 (runtime/generation
timeout on a small M1): here the weights are on the internal APFS SSD and the
abort happens at **load-time eval**, not during generation.

## Reproduction

- Machine: Mac Studio, M3 Ultra, 512 GB unified memory, macOS 26.x
- oMLX: `v0.6.3rc3` + the merged #3174 qwen4_exp support
- Model: `Qwen/Qwen3.8-Flash-Next` (`model_type: qwen4_exp`), raw **BF16** base,
  ~336 GB, 131 safetensors shards, on the internal SSD
- 1676 parameter arrays

```python
import mlx.core as mx
from mlx.utils import tree_flatten
from mlx_vlm.utils import load_model
# (qwen4_exp compat patch + configure_qwen4_exp_runtime elided)

model = load_model(model_path, lazy=True, strict=True)
arrays = [v for _, v in tree_flatten(model.parameters())]   # 1676 arrays
mx.eval(arrays)     # <-- aborts
```

Result:

```
RuntimeError: [METAL] Command buffer execution failed: Caused GPU Timeout Error
(00000002:kIOGPUCommandBufferCallbackErrorTimeout).
```

With `lazy=False` the same abort happens inside mlx-vlm's `load()` before
`materialize_lazy_state` is even reached, so simply moving to lazy loading is not
sufficient on its own.

## What works

Evaluating the identical array set in **bounded chunks** completes cleanly and
the model then serves correctly (verified end to end: coherent generations,
`finish_reason: stop`):

```python
CHUNK = 8
for i in range(0, len(arrays), CHUNK):
    mx.eval(arrays[i:i + CHUNK])
```

Chunk size is not sensitive; the point is that no single command buffer spans
the whole multi-hundred-GB tree.

## Proposed fix

Chunk the eval inside `materialize_lazy_state` (and ensure the VLM load path
reaches it via a lazy load rather than mlx-vlm's internal unchunked eval).
Small models are unaffected — a handful of chunks. A PR implementing the
`materialize_lazy_state` chunking is attached.

## Environment

- omlx: v0.6.3rc3 (+ #3174), Python 3.11.15, mlx 0.32.0
- Metal cap reported 464 GB; enforcer tier custom, ceiling 440 GB — the abort is
  a per-command-buffer watchdog timeout, not an out-of-memory condition (peak
  resident during a successful chunked load was ~354 GB, well under the ceiling).

scubamount:
Fixes #3179.

## Problem

Loading a large **BF16** checkpoint aborts the process with a Metal GPU
command-buffer timeout during post-load parameter materialization, before the
first request:

```
[METAL] Command buffer execution failed: Caused GPU Timeout Error
(00000002:kIOGPUCommandBufferCallbackErrorTimeout).
```

`materialize_lazy_state` collects every parameter array and runs a single
`mx.eval(arrays)`, which submits one Metal command buffer spanning the whole
tree. On a multi-hundred-GB model (the raw BF16 `Qwen/Qwen3.8-Flash-Next` base,
~336 GB / 1676 arrays, on internal SSD, M3 Ultra / 512 GB) that buffer exceeds
the GPU command-buffer watchdog and the runtime aborts with SIGABRT. It is a
per-command-buffer watchdog timeout, not an OOM (peak resident during a
successful chunked load was ~354 GB, under the 440 GB enforcer ceiling).

## Fix

Evaluate the identical array set in bounded chunks so no single command buffer
spans the whole tree. Chunk size is not sensitive (8 used here); small models
see only a handful of chunks and are otherwise unaffected.

## Validation

Reproduced and validated in isolation on the affected checkpoint:

- single `mx.eval(arrays)` over the 1676-array BF16 tree → reliably aborts with
  the watchdog timeout above;
- the same arrays evaluated in chunks of 8 → completes cleanly, and the model
  then serves coherent output (`finish_reason: stop`; e.g. "The capital of
  France is Paris.", primes `2, 3, 5, 7, 11`, a valid Python lambda, "80 km/h"
  with correct working).

## Scope / note

This patches the one unchunked full-tree eval in this repo
(`materialize_lazy_state`). The same watchdog abort can also originate from
mlx-vlm's own `load()` (`mx.eval(model.language_model.parameters())`) when a
model is loaded with `lazy=False`; reaching this chunked path requires the VLM
load to be lazy so materialization happens here. That mlx-vlm-side eval lives
upstream of this repo and is out of scope for this change; flagging it in #3179
for completeness.

### DIFF
--- a/omlx/utils/model_loading.py
+++ b/omlx/utils/model_loading.py
@@ -1137,7 +1137,17 @@ def materialize_lazy_state(model: Any) -> None:
                 _scan_plain_object(value)
 
     if arrays:
-        mx.eval(arrays)
+        # Evaluate in bounded chunks rather than one mx.eval(arrays). A single
+        # eval over the whole parameter tree submits one Metal command buffer
+        # spanning every array; on a multi-hundred-GB checkpoint (e.g. the raw
+        # BF16 Qwen3.8-Flash-Next base, ~336 GB / 1676 arrays) that buffer
+        # exceeds the GPU command-buffer watchdog and aborts the process with
+        # "[METAL] Command buffer execution failed: Caused GPU Timeout Error
+        # (kIOGPUCommandBufferCallbackErrorTimeout)". Chunking bounds each
+        # command buffer; small models see only a handful of chunks. See #3179.
+        _MATERIALIZE_EVAL_CHUNK = 8
+        for start in range(0, len(arrays), _MATERIALIZE_EVAL_CHUNK):
+            mx.eval(arrays[start : start + _MATERIALIZE_EVAL_CHUNK])
 
 
 def apply_post_load_transforms(model: Any, model_settings: Any = None) -> Any:


jundot:
Thanks for tracking this down. The bounded eval chunks address the Metal watchdog, so I merged the PR. I also followed up on main in [`dece501e`](https://github.com/jundot/omlx/commit/dece501e626a3d7db8f2bdd837e8c141c2d35fc9) to defer mlx-vlm's eager parameter eval until oMLX's chunked materialization path. I verified the final path with Qwen3.8, Qwen2-VL, and Gemma 4.

PR #3184, opened 2026-08-27
feat: add GLM-5.3-Flash support with native kernel reuse

jundot:
## Summary

- Add GLM-5.3-Flash support for both text-only and multimodal checkpoints.
- Vendor the model implementation from Blaizzy/mlx-vlm#2030, including the applicable non-speculative correctness fixes from Blaizzy/mlx-vlm#2044.
- Add a torch-free NumPy/Pillow image processor compatible with the official checkpoint metadata and oMLX's pinned Transformers version.
- Integrate GLM-5.3 with oMLX model discovery, loading, oQ quantization, continuous batching, vision feature caching, memory estimation, and cluster planning.

## Native kernel reuse

Reuse existing oMLX kernels where GLM-5.3 has compatible tensor contracts:

- GLM DSA indexer scoring and top-k selection
- Sparse MLA decode, short-block gather, and long-context attention
- GLM MoE weighted-sum
- Vector-gated delta attention
- Fused KDA input projections
- Shared affine Q2/Q4/Q5/Q6/Q8 prefill QMM kernels
- DeepSeek SwitchGLU expert dispatch paths

GLM indexer projections are kept in Q8 affine group-64 format so they remain compatible with the fused Q8 kernels.

## Runtime integration

- Add `glm5_next` VLM and text-only model discovery.
- Add `PoolingCache` and `BatchPoolingCache` support for the compressed indexer cache.
- Support variable-length and late-join continuous batches.
- Avoid duplicate latent KV storage for NoPE MLA.
- Account for the pooled indexer cache in runtime and cluster memory estimates.
- Support image token preflight and per-image vision feature caching.
- Preserve GLM image markers and their message order for the native checkpoint chat template.
- Disable TurboQuant KV for GLM-5.3 because its composite latent/indexer cache is not compatible with the TurboQuant cache layout.
- Drop unsupported `mtp.*` weights during base-model sanitization.
- Calibrate mHC, KDA, sparse MLA, MoE, and the untied output head for strict oQe conversion.

## Validation

- `10506 passed, 31 skipped, 77 deselected`
- 1,092 focused model-loading, VLM, oQ, cache, memory, and custom-kernel tests passed.
- 538 focused oQ, VLM engine, and GLM compatibility tests passed after the real-model fixes.
- 17 dedicated GLM-5.3 regression tests passed, including:
  - text prefill and decode
  - variable-length batching
  - late-join batching against single-request baselines
  - vision preprocessing and embedding merge
  - pooled cache lifecycle operations
  - GLM native DSA and MoE paths
  - official-size Q8 indexer prefill dispatch
- Loaded `inference-optimization/GLM-5.3-Flash-0.1B-A0.1B` directly and through the oMLX server.
- Verified BF16 text, streaming, image, multi-image, tool-template, concurrent, and late-join requests.
- Completed strict oQ4e conversion with the default 128 x 512 calibration:
  - 67 imatrix entries applied with no missing or mismatched entries
  - all 48 routed experts covered with no zero-count experts
  - streaming sanitize plan produced 174 output tensors without eager fallback
- Strict-loaded the converted oQ4e checkpoint and verified direct and server text, image, streaming, and concurrent generation.
- Loaded the official GLM-5.3 metadata with the custom processor, chat template, and `glm47` tool parser.
- Built the wheel and verified that all vendored GLM-5.3 Python modules are included.

## Limitations

- Lightning MTP is intentionally deferred. It requires separate validation of recurrent-state rollback and batched cache semantics.
- Video input remains unsupported because oMLX currently rejects video request content globally.

## Upstream references

- https://github.com/Blaizzy/mlx-vlm/pull/2030
- https://github.com/Blaizzy/mlx-vlm/pull/2044

### DIFF
--- a/omlx/cluster/planner.py
+++ b/omlx/cluster/planner.py
@@ -581,6 +581,25 @@ def _config_int(
     )
 
 
+def _config_nonnegative_int(config: dict[str, Any], key: str) -> int | None:
+    """Read an explicitly present, bounded integer that may be zero."""
+    candidates = [config]
+    for nested in ("text_config", "language_config", "llm_config"):
+        value = config.get(nested)
+        if isinstance(value, dict):
+            candidates.append(value)
+    return next(
+        (
+            value
+            for candidate in candidates
+            if isinstance((value := candidate.get(key)), int)
+            and not isinstance(value, bool)
+            and 0 <= value <= 4096
+        ),
+        None,
+    )
+
+
 def _supports_tensor_parallel(config: dict[str, Any]) -> bool:
     """Whether this architecture has a runtime-safe tensor strategy.
 
@@ -796,7 +815,7 @@ def _kv_cache_replicated_across_tp(config: dict[str, Any]) -> bool:
 
     return (
         _config_int(config, "kv_lora_rank", 0) > 0
-        and _config_int(config, "qk_rope_head_dim", 0) > 0
+        and _config_nonnegative_int(config, "qk_rope_head_dim") is not None
     )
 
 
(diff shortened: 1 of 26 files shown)


DiscoStew6082:
Thanks for adding GLM-5.3 support. I found a cache bug when prompts of different lengths run in the same batch.

The shorter prompt can get different scores for its first generated token when it runs in a batch than when it runs alone. That means the other prompts in the batch can change its output. The existing test compares only the chosen tokens, so it can miss this difference.

The problem comes from right padding. The cache treats padded positions as real tokens and saves them as the latest convolution state for the shorter prompt.

I pushed a small fix and a stronger regression test here:
https://github.com/DiscoStew6082/omlx/commit/f797bbc690ef3f366b658286efafda72ee3de51e

The fix uses each prompt’s real length when saving the cache. It leaves the existing left-padding and late-join behavior unchanged.

I rebased the fix directly onto the current PR head, `6bf8ef12`. All 18 GLM compatibility tests passed with native kernels. The GLM, oQ, and VLM engine test files also passed, with 539 tests in total.

jundot:
Thanks @DiscoStew6082. I reproduced the first-token logits mismatch and confirmed that right padding was replacing the shorter row's convolution tail. I applied your fix and regression test and kept you as the commit author. All 18 GLM compatibility tests and the 539-test GLM/oQ/VLM set pass with native kernels available.

jundot:
Real-checkpoint validation is complete. I quantized the official GLM-5.3-Flash checkpoint to strict oQ4e with 128 x 512-token imatrix calibration, producing a 169 GiB model with 34 verified safetensors shards.

| Prompt | Generation | PP | TG | TTFT | Peak footprint |
| [4,096](https://omlx.ai/benchmarks/performance/ztfgwsn3) | 128 | 482.3 tok/s | 24.1 tok/s | 8.49 s | 179.50 GiB |
| [16,384](https://omlx.ai/benchmarks/performance/itzfxe5e) | 128 | 443.2 tok/s | 23.7 tok/s | 36.97 s | 179.49 GiB |
| [32,768](https://omlx.ai/benchmarks/performance/9n9m2n04) | 128 | 449.6 tok/s | 23.5 tok/s | 72.89 s | 179.77 GiB |

For a 12,362-token greedy cache check, cold prefill, partial hit, and full reusable-block hit reported 0, 8,192, and 12,288 cached tokens. The reasoning and final response were identical in all three cases. SSD persistence used six 2,048-token blocks with zero cache errors or write drops. Two concurrent requests also completed together through continuous batching.

The full-checkpoint quantization exposed two strict imatrix coverage gaps in the short sparse-attention calibration path. I fixed both on `main` in `bceb44e6`. Focused verification passed 95 GLM/kernel tests and 5 oQ tests.

The source FP8 checkpoint itself is not safe to benchmark on this 512 GiB machine because the current GLM sanitize path dequantizes its weights to an approximately 598.6 GiB BF16 representation. This exceeds both physical RAM and the 464 GiB Metal working-set cap, so the throughput results above are for the validated oQ4e model.

PR #3188, opened 2026-08-27
fix(vlm): honor model-owned batch cache conversion

DiscoStew6082:
## Summary

- Let oMLX's patched MLX-LM batch-cache factory honor cache-provided `to_batch()` conversion, including nested cache containers.
- Preserve MLX-LM's standard conversion for ordinary cache types.
- Make the DeepSeek V4 `PoolingCache` wrapper compose with model-owned cache conversion regardless of patch installation order.

## Problem

After #3169, `QSAKVCache` provides its own batch conversion, but `mlx_lm.generate._make_cache` does not dispatch to it. The continuous-batching handoff therefore raises:

```text
ValueError: <class 'mlx_vlm.models.qwen4_exp.language.QSAKVCache'> does not yet support batching
```

This PR fixes that generation-time factory seam. Prefix-cache serialization and storage remain separate concerns.

The Qwen regression test uses populated keys, values, index keys, and position IDs so auxiliary QSA state must survive conversion. DeepSeek tests cover mixed `PoolingCache` and `QSAKVCache` graphs in both patch orders.

## Validation

- Focused cache, VLM, DeepSeek, and loading tests: `431 passed, 5 skipped`
- Full suite excluding tests marked slow or integration: `10426 passed, 102 skipped, 77 deselected`
- `git diff --check upstream/main..HEAD`

No full checkpoint load was performed; the native cache tests exercise the production `_make_cache` handoff directly.

### DIFF
--- a/omlx/patches/deepseek_v4/generate_patch.py
+++ b/omlx/patches/deepseek_v4/generate_patch.py
@@ -5,9 +5,10 @@ PR 1192 adds one ``elif`` branch to the ``to_batch_cache`` closure inside
 ``_make_cache``: ``isinstance(c, PoolingCache)`` → ``BatchPoolingCache``.
 
 The closure is not externally hookable, so we replace ``_make_cache``
-itself with a copy whose body is identical to PR 1192. ``_make_cache`` is
-called via the module-level binding from inside ``mlx_lm.generate``, so
-overwriting the attribute is sufficient.
+with a compatible wrapper that adds PR 1192's PoolingCache conversion while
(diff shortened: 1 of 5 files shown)


jundot:
Thanks for fixing the cache factory handoff. I confirmed that the model-owned `to_batch()` path preserves QSA auxiliary state, keeps the standard MLX-LM conversions intact, and composes correctly with the DeepSeek V4 `PoolingCache` patch in either installation order. The focused cache and MTP tests pass here. LGTM!

PR #3190, opened 2026-08-27
fix(updater): clean up mounted update DMGs before relaunch

### LINKED ISSUE #3189: macOS updater leaves mounted DMGs and temporary download files after relaunch
## Summary

The macOS in-app updater can leave its downloaded DMG mounted after an update completes and the application relaunches. The corresponding temporary update directory and DMG backing file are also left on disk.

This has accumulated across multiple oMLX updates on the same machine.

## Environment

- macOS 26.7 (25G220)
- Apple Silicon
- Current oMLX version: 0.6.3rc3
- Updates installed through the in-app updater

## Observed leftovers

Five read-only update images were still mounted:

- oMLX 0.6.1
- oMLX 0.6.2
- oMLX 0.6.3rc1
- oMLX 0.6.3rc2
- oMLX 0.6.3rc3

They were mounted between August 18 and August 25. Their downloaded DMG backing files total approximately 3.74 GiB.

The state can be inspected with:

```shell
hdiutil info
find "$TMPDIR" -type f -path '*/omlx-update-*/oMLX-*.dmg' -print
```

## Reproduction

1. Install an oMLX version distributed as a DMG.
2. Use the in-app updater and choose Install & Restart.
3. Allow the updater to download, stage, install, and relaunch the application.
4. After relaunch, inspect mounted disk images with `hdiutil info`.
5. Inspect `$TMPDIR` for directories named `omlx-update-<UUID>`.

After repeating this across releases, each update may leave another mounted image and downloaded DMG.

## Root cause

`AppUpdater.run(app:)` registers function-scope `defer` blocks to detach the mounted DMG and remove the temporary directory.

After staging succeeds, it calls `onReady()`. `UpdateController` handles that callback by scheduling the bundle swap and terminating the application.

Because `onReady()` can synchronously initiate process termination, the process may exit before `run(app:)` unwinds. The function-scope `defer` blocks therefore never get an opportunity to detach the image or delete its backing file.

In effect, cleanup is scheduled after a callback that may terminate the process.

## Expected behavior

Before reporting the staged update as ready, the updater should:

1. Detach the mounted DMG.
2. Remove the updater-owned temporary directory and downloaded DMG.
3. Notify the controller that the update is ready.
4. Allow the swap and application termination to proceed.

## Actual behavior

The updater notifies the controller first. Application termination can prevent the deferred cleanup from running, leaving both the mount and its backing file behind.

## Impact

- Temporary update downloads accumulate and consume disk space.
- Read-only oMLX installation images remain mounted.
- Repeated updates can leave several gigabytes of stale files.
- Users generally do not notice the leak until inspecting disk usage or mounted volumes.

## Related

A remaining mounted update DMG was also reported after the main swap fix in #2522:

https://github.com/jundot/omlx/issues/2522#issuecomment-5330458566

chenqianhe:
## Summary

Closes #3189.

This change ensures that the macOS updater detaches its mounted DMG and removes its temporary download files before invoking the callback that may terminate oMLX.

The change is intentionally limited to resource cleanup. It does not alter release selection, downloading, staging, bundle swapping, or relaunch behavior.

## Root cause

`AppUpdater.run(app:)` previously relied on function-scope `defer` blocks for both operations:

- detaching the mounted DMG;
- removing the updater-created temporary directory.

After staging completed, the updater called `onReady()`. In the auto-install flow, `UpdateController` immediately scheduled the swap and terminated the application from that callback.

As a result, the process could exit before `run(app:)` unwound, preventing both deferred cleanup blocks from executing.

The effective sequence was:

```text
download → mount → stage → onReady → swap/terminate
                              ↓
                   deferred cleanup may never run
```

## Changes

The successful update path now explicitly performs cleanup before notifying the controller:

```text
download → mount → stage → detach → remove temporary files → onReady
```

Specifically, this PR:

- Detaches the mounted DMG before invoking `onReady`.
- Removes the updater-owned temporary directory before invoking `onReady`.
- Retains guarded `defer` cleanup as a fallback for cancellation and earlier failure paths.
- Checks the `hdiutil detach` exit status instead of silently ignoring a failed detach.
- Adds dedicated errors for unmount and temporary-file cleanup failures.
- Prevents the ready callback from running when required cleanup fails.

## Tests

Added `AppUpdaterTests` covering:

- Successful ordering: detach, remove temporary files, then notify ready.
- Detach failure prevents temporary removal and ready notification.
- Temporary-file removal failure prevents ready notification.

### DIFF
--- a/apps/omlx-mac/Sources/Updater/AppUpdater.swift
+++ b/apps/omlx-mac/Sources/Updater/AppUpdater.swift
@@ -20,6 +20,8 @@ final class AppUpdater {
         case notWritable(String)
         case downloadFailed(String)
         case mountFailed(String)
+        case unmountFailed(String)
+        case cleanupFailed(String)
         case appNotFoundInVolume
         case stageFailed(String)
         case cancelled
@@ -30,6 +32,8 @@ final class AppUpdater {
                 return "Cannot write to \(path). Move oMLX.app to a writable location and try again."
             case .downloadFailed(let m): return "Download failed: \(m)"
             case .mountFailed(let m): return "Could not mount DMG: \(m)"
+            case .unmountFailed(let m): return "Could not unmount DMG: \(m)"
+            case .cleanupFailed(let m): return "Could not clean up update files: \(m)"
             case .appNotFoundInVolume: return "oMLX.app not found inside the downloaded DMG"
             case .stageFailed(let m): return "Could not stage the update: \(m)"
             case .cancelled: return "Update cancelled"
@@ -124,7 +128,12 @@ final class AppUpdater {
             onError(.downloadFailed("Could not create temp dir: \(error.localizedDescription)"))
             return
         }
-        defer { try? FileManager.default.removeItem(at: tmpDir) }
+        var temporaryDirectoryNeedsCleanup = true
+        defer {
+            if temporaryDirectoryNeedsCleanup {
+                try? FileManager.default.removeItem(at: tmpDir)
+            }
+        }
 
         let dmgPath = tmpDir.appendingPathComponent("oMLX-\(version).dmg")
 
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for tracking down the updater cleanup leak. Moving DMG detachment and temporary-directory removal ahead of the ready callback fixes the termination race while retaining deferred cleanup for cancellation and failure paths. I also confirmed that the new `AppUpdater` tests pass on macOS. LGTM, thanks!

PR #3192, opened 2026-08-27
cluster: data-plane liveness check (2.3)

alytaphoenix:
## Summary

Item 2.3 of the cluster-hardening pass, on top of #3173 (Phase 1), #3175 (Phase 0), #3177 (item 2.1), and #3178 (item 2.2).

**The problem (doc's §C1).** Every liveness signal today rides SSH (the LAN/mDNS route). The actual MLX collectives ride a *different* physical link -- the Thunderbolt/RDMA point-to-point hostfile IPs selected at planning time. A cable pull with SSH still up leaves every probe green, every marker fresh (heartbeats are on independent threads per item 2.1), and every rank blocked in a collective with no timeout.

Design-reviewed with Fable before implementation (second item this pass to get that treatment). Two corrections to the doc's own proposal, both load-bearing:

1. **`ClusterHost.ips` doesn't carry interface names**, so the doc's "reuse `verify_link_reachability`" doesn't plug in directly -- that function needs `LinkEndpoint.interface` and exists to *select* a candidate link among ambiguous stale addresses via bidirectional SSH, not to *re-check* an already-established one. New `check_data_plane_reachability()` (`transport.py`) is local-only instead: subnet-matches the peer IP against this Mac's own interfaces (free interface recovery, since both call sites always run rank-0-local), confirms the local route still uses that interface (catches a silently re-homed path a plain ping would miss -- matters most for RDMA/jaccl, which is pinned to a device), then pings. Returns `(None, reason)` for "inconclusive" (no local subnet match) distinctly from `(False, reason)` for "confirmed down," so a caller can't fail a deployment over a check that couldn't run.

2. **Kill semantics don't vary by backend.** The doc asks to "decide kill semantics from 0.3 evidence per backend." Phase 0.3's evidence is a killed generation *thread* (sockets stay alive) -- not evidence about actual interface-down behavior on either `ring` or `jaccl`; using it to differentiate would be a category error. Instead the kill condition is self-evidently safe on any backend: data-plane down for `failure_tolerance` consecutive polls **and** (item 2.2's `stalled` signal, or the deployment still loading, where no progress signal exists yet). A failed ping alone (ICMP dropped on a saturated link mid-collective) must not kill a cluster demonstrably still working.

**Two call sites, both gated on "SSH-side health already looks fine" and fail-open on probe errors:**
- Preflight (`_require_healthy_cluster`, `engine/distributed.py`) checks each peer's `deployment.hosts[*].ips` -- it already has full deployment access.
- `PeerWatchdog` (the thing armed *during* the 20-minute weight load, before any request ever reaches the preflight) needed brand-new plumbing: a `--peer-ips` launcher argv (the worker subprocess previously only ever got SSH hostnames), parsed and threaded through `_start_peer_watchdog` into a new `peer_ips_by_rank` param -- fully optional, so an argv predating this field just means the check never runs.

## Test plan

- [x] Full test suite green (10400 passed, 102 skipped, 0 failed)
- [x] 39 new/changed tests across `transport.py`, `engine/distributed.py`, `liveness.py`, `inference_worker.py`, and `launch.py`'s test files, each verified to fail when its corresponding change is reverted
- [ ] Not done: a real cable-pull / `ifconfig down` validation on the live 2-Mac cluster (Fable's suggested verification step) -- would additionally retire the ring-vs-jaccl question empirically. Noted as a follow-up; the unit-level coverage above proves the logic but doesn't touch live hardware.


alytaphoenix:
Rebased onto current main (15 commits, 1 real conflict in planner.py: main independently added a \`_config_nonnegative_int\` helper at the same insertion point as this PR's \`_config_truthy\`/\`_nemotron_h_tp_bias_unsafe\` — both purely additive, kept both). Full suite: 10,897 passed, 0 failed, 152 skipped, 77 deselected.

PR #3194, opened 2026-08-27
refactor(api): share content-block text extraction, add hypothesis tests

phaedrus1992:
## Summary

Anthropic content-block text extraction was reimplemented ad hoc at several call sites, each handling block shapes slightly differently. Converges them on one shared helper and covers it with Hypothesis property tests instead of fixed examples.

## Pre-PR review fixes

- Added a property test for `_extract_system_text`'s billing-header filter
- Added a property test for `truncate_tool_result`'s token budget
- Fixed a missing `isinstance(str)` guard in `_extract_tool_result_content` that could raise `TypeError` on a non-str `text` field
- Reused `_content_block_to_dict()` for tool normalization instead of an inline `hasattr` check the refactor had left inconsistent

## Reviewed and dismissed

- Same normalization pattern still hand-rolled in `omlx/api/utils.py`, `omlx/api/responses_utils.py`, and `omlx/patches/deepseek_v4/chat_template_v4.py` — out of scope for this refactor, tracked in phaedrus1992/omlx#27
- Hypothesis dev dependency and one property test flagged as possibly redundant — kept, the property test covers a broader input space than the example tests it overlaps with

## Test plan
- [x] `pytest tests/test_anthropic_text_extraction.py tests/test_truncate_tool_result.py` — 44 passed


scubamount:
## Verified

- **Purpose:** several call sites re-implemented Anthropic content-block text extraction with slightly different shape handling; converges on `_text_from_content_blocks` over the existing `_content_block_to_dict` normalizer.
- **Boundary honesty:** docstring explicitly states the helper returns raw text — billing-header filter and budget-marker stripping stay with the callers that need them, avoiding a hidden-behavior grab bag.
- **Tests:** Hypothesis property tests replace fixed examples (system-text billing filter, tool-result token budget) — better coverage-per-line for shape-handling code.

No issues.

PR #3195, opened 2026-08-27
fix(admin): validate model-settings and profile config at write time

phaedrus1992:
## Summary

reasoning-parser config, model-settings draft-model fields, and profile-write ModelSettings fields were all accepted from the admin write path without validation, so a malformed value broke inference on the next request instead of failing the save that introduced it.

Validates each at write time, using real doc-derived examples in tests, and closes the type-confusion and missing-field gaps found while doing so.

## Test plan
- [x] `pytest tests/test_admin_model_settings.py tests/test_admin_profiles_api.py tests/test_grammar.py tests/test_grammar_live.py tests/test_model_settings.py` — 291 passed, 18 skipped

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -1800,9 +1800,9 @@ def _models_from_docstring(fn) -> list[str]:
     ]
 
 
-@router.get("/api/grammar/parsers")
-async def list_grammar_parsers(is_admin: bool = Depends(require_admin)):
-    """Return available reasoning parser names from xgrammar.
+def _reasoning_parser_registry() -> dict[str, list[str]] | None:
+    """Return ``{style: supported_model_names}`` for xgrammar's builtin
+    reasoning parsers, or ``None`` if the registry can't be determined.
 
     Supports both API generations:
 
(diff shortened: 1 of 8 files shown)


scubamount:
## Verified

- **Root cause correct:** reasoning-parser names, draft-model fields, and profile-write ModelSettings were accepted unvalidated — a malformed value broke inference on the NEXT request instead of failing the save that introduced it.
- **Fix right:** extracts `_reasoning_parser_registry()` (returning `{style: models}` or `None`) as the single source of truth shared by the admin dropdown AND `update_model_settings` write validation — one predicate, one implementation (both xgrammar API generations still supported, torch-stub ordering preserved).
- **Fail-open behavior:** registry unavailable → validation cannot enumerate names; ensure that path doesn't silently accept garbage (from the diff, the None return distinguishes "unknown" from "empty" — the correct three-valued shape).

No blocking issues. Write-time validation using the same registry as the UI is the right convergence.

jundot:
Thanks for working on this. I agree that invalid settings should be caught when saving, but this implementation duplicates validation across write paths and rejects a supported DFlash assistant checkpoint. I’m closing this PR and will address the confirmed validation gaps in a smaller follow-up using the existing test setup.

PR #3196, opened 2026-08-27
fix(cluster): stop sending remote nodes a tilde-abbreviated model path

phaedrus1992:
## Summary

Local paths get abbreviated to ~-form for display, but that abbreviated form was leaking into the path sent to remote cluster nodes, which don't share the local home-directory layout. Converts consistently at the staging/planning boundary instead, and stops mislabeling local path errors as remote failures.

## Test plan
- [x] `pytest tests/test_cluster_staging.py tests/test_cluster_remote_planning.py` — 62 passed

### DIFF
--- a/omlx/cluster/planner.py
+++ b/omlx/cluster/planner.py
@@ -17,7 +17,12 @@ from typing import Any
 
 from .node_role import ROLES
 from .performance import ExecutionProfileName, NodePerformanceProfile
-from .staging import DEFAULT_REMOTE_PYTHON, is_local_host, run_remote_python
+from .staging import (
+    DEFAULT_REMOTE_PYTHON,
+    home_relative_model_path,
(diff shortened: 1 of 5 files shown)


scubamount:
## Verified

- **Root cause correct:** display-only `~`-abbreviated path leaked into remote staging/planning; remote Macs have different `$HOME` layouts (cross-user clusters), so the coordinator's abbreviated (or absolute) path named nothing on the peer.
- **Fix right:** conversion happens once at the staging/planning boundary via `home_relative_model_path` before the path crosses SSH, re-expressed per the `remote_model_dir` contract (#12); docstrings updated to state whose path is whose at each hop, and local path errors no longer mislabeled remote failures.
- **Tests:** 62 passed across staging + remote-planning suites per plan.

No issues.

PR #3197, opened 2026-08-27
fix(cluster): report the real reason tensor-parallel planning fails

phaedrus1992:
## Summary

The "no tensor-parallel degree divides N nodes across architecture dimensions" message fired even when the degree divided the architecture cleanly, because every candidate had actually been skipped for a slow link with no pipeline fallback for the architecture — divisibility was never the problem, but the message claimed it was.

Tracks candidates skipped for the slow-transport reason and reports it accurately instead of falling through to the generic (and in this case wrong) divisibility message.

## Test plan
- [x] `pytest tests/test_cluster_autoconfigure.py` — 87 passed
- [x] `pytest tests/test_cluster_autoconfigure.py tests/test_cluster_staging.py tests/test_cluster_remote_planning.py` — 149 passed

### DIFF
--- a/omlx/cluster/autoconfigure.py
+++ b/omlx/cluster/autoconfigure.py
@@ -248,6 +248,7 @@ def choose_parallelism(
 
     warnings: list[str] = []
     last_error: PlanningError | None = None
+    skipped_slow_transport: tuple[int, ...] = ()
 
     for tensor_parallel_size in candidates:
         fast_enough = transports_are_fast_enough(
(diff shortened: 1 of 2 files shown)


scubamount:
## Verified

- **Root cause correct:** generic "no tensor-parallel degree divides N" error fired even when divisibility was fine — every candidate had been skipped for slow-link-with-no-pipeline-fallback. Wrong diagnosis shipped to operators.
- **Fix right:** tracks `skipped_slow_transport` degrees and, only when a divisibility-clean candidate exists but all were link-skipped, raises the accurate message naming the degree, the divisors, and the missing pipeline fallback.
- **Test:** updated assertion `8 % 2 == 0` — the test now proves the message cannot claim divisibility failure when divisibility holds (fail-before/pass-after on the exact bug).

No issues. Good error-accuracy fix; exactly the class where a misleading message is worse than none.

PR #3198, opened 2026-08-27
fix(cluster): shell-quote remote link-probe commands (#2849 root cause)

alytaphoenix:
## Summary

Fixes the root-cause bug identified in review of #2849, split out standalone since it's blocking that PR and the rest of the fabric-doctor series (#2867, #2875, #2878, #2885) independent of any of their own remaining changes.

**The bug.** `ssh` joins every trailing argument with a bare space before handing the result to the remote shell -- it does not preserve `subprocess.run`'s argv element boundaries. `verify_link_reachability`'s Linux/bound-connect fallback passes a multi-line `python3 -c` script as one of several trailing arguments to `_run_link_command`, which forwards them to `ssh` unquoted. The remote shell receives the script's embedded newlines as separate shell statements instead of one atomic argument, and fails with a syntax error -- on every real remote invocation, regardless of what the script checks. Every existing test in `test_cluster_link_status.py` injects a `runner` directly into `verify_link_reachability`, bypassing `_run_link_command` entirely, so this was invisible to the whole suite.

**The fix.**
- `_run_link_command` now `shlex.join`s the remote command into the single argument `ssh` actually preserves verbatim, mirroring `_remote_gui_authorize`'s already-correct pattern earlier in the same file.
- Bare `python3` -> `/usr/bin/python3` (the packaged app has no `python3` on `PATH`).
- The bound-connect probe now targets port 8000 (oMLX's own default server port) instead of 22 (SSH). Probing SSH silently made inbound SSH scoped to the fabric interface a *new* requirement for clustering to pass this check, unrelated to whether the fabric itself works. Advertising and probing a peer's actual configured port is tracked separately (§C5/#2885) -- this hardcodes the common default as the immediate, self-contained fix for the mangling bug specifically.

## Test plan

- [x] Full test suite green (10457 passed, 106 skipped, 0 failed)
- [x] Two new tests exercise `_run_link_command` itself (not a mocked runner) plus the previously entirely-untested bound-connect fallback path; both verified to fail against the pre-fix code
- [x] **Verified live against the real two-Mac cluster** (peer online), per the review request: the exact pre-fix argv shape reproduces the described syntax error over a real `ssh` round-trip; the fixed shape completes a real bound TCP connect from the peer's fabric address to the coordinator's, on port 8000

### DIFF
--- a/omlx/cluster/transport.py
+++ b/omlx/cluster/transport.py
@@ -1442,12 +1442,33 @@ LinkCommandRunner = Callable[
     [str, Sequence[str]], subprocess.CompletedProcess[str]
 ]
 
+# ``ServerSettings.port``/``ServerConfig.port``'s own default (settings.py,
+# config.py). Deliberately not imported from there: this module stays
+# usable without pulling in settings-loading machinery, and the port only
+# needs to be *something* oMLX itself listens on by default, not the
+# caller's actual configured port -- see the probe's own comment below for
+# why 22 (SSH) was wrong and what the honest limitation of this port is.
+_BOUND_CONNECT_PROBE_PORT = 8000
+
 
 def _run_link_command(
     ssh_hostname: str,
     command: Sequence[str],
 ) -> subprocess.CompletedProcess[str]:
-    """Run one bounded link check locally or through the paired SSH identity."""
+    """Run one bounded link check locally or through the paired SSH identity.
+
+    ``ssh`` joins every trailing argument with a bare space before handing
+    the result to the remote shell -- it does not preserve the local argv's
+    element boundaries the way ``subprocess.run`` does. An unquoted
+    multi-line command (the Linux bound-connect probe below is the reason
+    this exists) arrives split across separate shell statements instead of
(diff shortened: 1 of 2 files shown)


scubamount:
## Verified

- **Root cause correct:** `ssh` joins trailing argv with spaces before handing to the remote shell — the multi-line bound-connect probe script arrived split across shell statements and failed on every remote invocation. Invisible to mocked tests.
- **Fix right:** `shlex.join(argv)` folds the command into the single remote-shell argument SSH preserves verbatim, mirroring `_remote_gui_authorize`'s already-correct pattern — fix at the shared runner, so all future callers get it.
- **Probe port:** 22→8000 (oMLX's own default) so the probe doesn't require SSH service on the probed path and the limitation is documented honestly rather than imported settings machinery.

Note for maintainer: #3206 reworks the same probe (adds `_tcp_probe_succeeds` fallback + `_TCP_PROBE_SCRIPT` constant targeting port 22). Whichever lands second needs a rebase to fold both changes — they're semantically compatible (quoting fix + fallback + port choice all compose) but textually conflict in `verify_link_reachability`.

PR #3200, opened 2026-08-27
fix(admin): let the Lightning MTP toggle through for Qwen3.8 Flash Next (qwen4_exp)

williamxie1989:
## Problem

Enabling **Lightning MTP** on a Qwen3.8 Flash Next checkpoint (`model_type="qwen4_exp"`) is impossible from the admin UI / settings API even though the runtime supports it:

- The model-settings toggle is greyed out with the reason
  `model_type='qwen4_exp' is not on the MTP whitelist (supported: qwen3_5*, qwen3_6*, deepseek_v4*, glm_moe_dsa, gemma4, gemma4_unified)`
- Forcing `mtp_enabled=true` through the model-settings API returns **400** `Model is not MTP-compatible ...`.

## Root cause

#3174 added Lightning MTP for qwen4_exp through a **dedicated VLM path** in `omlx/utils/model_loading.py` (`maybe_apply_pre_load_patches`): it applies the `mlx_lm_mtp` dispatch patch and hands off to `configure_qwen4_exp_runtime(mtp_enabled=...)`. That path *intentionally* bypasses the mlx-lm `_is_mtp_compatible` whitelist — the whitelist gates the **generic text-model** patch, which must not run for qwen4_exp (it would attach a generic `MTPModule` via the `mlx_vlm_mtp` runtime patch and conflict with the vendored qwen4_exp model's own MTP attachment).

Both admin gates, however, kept reusing that whitelist as if it were the general MTP support list:

- `_mtp_compat_for_model()` → drives `mtp_compatible` / `mtp_compatibility_reason` for the UI,
- the `mtp_enabled` validation in `update_model_settings()` → rejects the save with 400.

So the runtime capability shipped in #3174 can never be switched on.

## Fix

Accept `qwen4_exp` in both admin gates and let the existing **embedded `mtp.*` weight check** decide — mirroring the exact condition the runtime path applies (`mtp_enabled && _checkpoint_has_mtp_weights(...)`). No runtime or whitelist changes, so the generic mlx-lm path is untouched.

## Tests

`tests/test_qwen4_exp_mtp_admin_gate.py` (7 cases): qwen4_exp passes with `mtp.*` weights; is still blocked on missing weights (weight message, not whitelist message) and on missing MTP heads; unknown model types still hit the whitelist; generic whitelisted types (`qwen3_5*`, `qwen3_6*`, `deepseek_v4*`) keep passing.

```
tests/test_qwen4_exp_mtp_admin_gate.py .......  [100%]
7 passed in 0.93s
```

Also verified end-to-end on a Mac Studio against real checkpoints (`Qwen3.8-Flash-Next` and `Qwen3.8-Flash-Next-oQ8e-mtp`, both shipping embedded `mtp.*` tensors): the gate now returns `(True, "")` for both.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -704,7 +704,14 @@ def _mtp_compat_for_model(model_info: dict) -> tuple[bool, str]:
     model_type = cfg.get("model_type")
     if not _has_mtp_heads(cfg):
         return False, "model has no MTP heads in config"
-    if not _is_mtp_compatible(cfg, model_type):
+    # qwen4_exp (Qwen3.8 Flash Next) attaches its Lightning MTP head through
+    # the dedicated VLM path in omlx.utils.model_loading (vendored mlx-vlm
+    # qwen4_exp model + mlx_lm_mtp dispatch patch) and never goes through the
+    # mlx-lm ``_is_mtp_compatible`` whitelist, which gates the generic
+    # text-model patch. Mirroring the runtime, the admin gate accepts
+    # qwen4_exp and relies on the embedded ``mtp.*`` weight check below —
+    # the same condition the runtime path uses.
+    if model_type != "qwen4_exp" and not _is_mtp_compatible(cfg, model_type):
         return False, (
             f"model_type={model_type!r} is not on the MTP whitelist "
             "(supported: qwen3_5*, qwen3_6*, deepseek_v4*, glm_moe_dsa, "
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for fixing the missing admin gate for Qwen4-Exp Lightning MTP. The qwen4_exp exception correctly lets embedded mtp.* checkpoints reach the dedicated VLM runtime without broadening the generic mlx-lm path. One thing I'll fold into a follow-up is sharing the qwen4-specific weight-layout check with the admin gates, since generic nextn detection is slightly broader than the runtime. LGTM, merging.

Issue #2479, opened 2026-08-02
Additional Model Settings labels could be brought under i18n coverage


aosama:
Thank you for the ongoing work on the admin localization improvements and for addressing the specific labels covered by #2476.

While reviewing that issue, we noticed a few additional user-facing labels in `omlx/admin/templates/dashboard/_modal_model_settings.html` that could also be brought under the existing i18n system in a future, focused pass:

- `Draft Model` (used in the SpecPrefill and DFlash controls)
- `Keep Rate`
- `Threshold (tokens)`
- `Quantization`
- `Weight Bits`
- `Activation Bits`
- `Group Size`
- `Max Context (fallback threshold)`
- `Long-context tuning`
- `Draft window size`
- `Draft sink size`
- `Verify mode`
- `In-memory cache`
- `In-memory cache max entries`
- `In-memory cache size (GiB)`
- `SSD cache`
- `SSD cache size (GiB)`

This is not a functional problem: the current locale loader provides an English fallback when a translation key is missing. Adding keys for these labels would nevertheless improve consistency with the rest of the Model Settings panel and give non-English locales an opportunity to provide translations.

A possible follow-up would be to add the labels to the locale catalogs and replace the inline strings with the existing `t(...)` template helper. This request is intentionally separate from #2476 so that the narrowly scoped fix there remains easy to review.

The separate Top-K/Min-P casing behavior is not included here because it has a different presentation-related root cause.

yanzhaohui1999:
Heads-up for triage: the 17 labels listed here are already extracted in the open PR #3140 (i18n hardening, Phase 5.4-5.5) — it covers all of them, plus the related hints, select options and placeholders in `_modal_model_settings.html`.

A follow-up PR (#3201) handles the remaining raw labels **outside** this list (engine-type dropdown options, the community benchmark upload block, and the "About oQ Quantization" panel), so this issue can likely be closed once #3140 lands.

PR #3201, opened 2026-08-27
fix(i18n): extract engine-type, community-bench and oQ About labels

yanzhaohui1999:
## Summary

Removes the last hardcoded English labels from the admin dashboard and
routes them through the existing i18n system (9 locale catalogs).

Scope (config-independent, no MLX/engine changes):

- **Engine-type dropdown** in Model Settings
  (`_modal_model_settings.html`): LLM / VLM / Embedding / Reranker /
  Audio STT / Audio TTS / Audio STS option texts → 7 new keys
  (`modal.model_settings.engine_type_*`).
- **Community benchmark upload block** (`_bench.html`): section title and
  "Uploading..." → 2 new keys (`bench.community.*`).
- **"About oQ Quantization" panel** (`_models.html`): full collapsible
  panel — headings, quote, mixed-precision explainer, comparison table
  (Path / Bit allocation / Affine quantization / Best use + both rows),
  the 4-step "How oQe imatrix works" walkthrough, and the strict-coverage
  footnote → 37 new keys (`models.oq.about_*`).

46 keys were added to **all 9 locale catalogs**. `zh.json` ships real
translations aligned with the existing terminology (imatrix kept as-is,
"回退", "仿射量化", "严格 imatrix 覆盖", "float 基线", "位数分配"); the
other 8 locales keep English values, matching the current i18n convention
(see the Phase 5.x i18n PRs). The `value` attributes of the engine-type
`<option>`s are untouched.

## Relation to #2479 / #3140

#2479 asked for ~17 Model Settings labels to be extracted. Open PR #3140
(i18n Phase 5.4-5.5) already covers that exact list (plus hints, options
and placeholders). This PR deliberately covers the **remaining** raw
labels outside that list (engine-type options, bench upload block, oQ
About panel) — zero overlap with #3140; both only *add* keys and never
touch each other's keys, so merge order does not matter.

## Tests

- New `tests/test_admin_i18n_labels.py`:
  - all 46 new keys exist in every locale catalog;
  - zh spot-checks (translations present, "回退"/"imatrix" in the
    fallback note, LLM/Embedding stay untranslated by design);
  - every `t()` key used by the oQ About panel must exist in `en.json`.
- Full suite with the CI filter (`pytest tests/ -m "not slow and not
  integration"`): **10424 passed, 132 skipped**. The 6 failures were
  reproduced unchanged on clean `main` (cluster SSH supervisor, cluster
  seams, VLM MTP e2e — environment-bound) or proved flaky (2 GDN sidecar
  tests pass on rerun); none touch i18n.
- `scripts/normalize_i18n.py` is idempotent on the new catalogs; all
  three touched templates parse as Jinja; en values are byte-identical to
  the replaced strings.
- Live smoke test: admin dashboard rendered with `ui.language=zh` (server
  run from this branch) — all three areas render the zh translations, no
  raw or unrendered `t()` strings remain.


scubamount:
## Verified

- **Scope:** final hardcoded English labels routed through the 9-locale i18n system — engine-type dropdown (7 keys), community-bench upload labels, oQ About modal (~35 keys). No engine/MLX changes.
- **Completeness:** en.json keys added with matching template `t()` conversions; the oQ About content is pure copy migration (claims about imatrix/E[x^2]/error weighting match the existing oQ docs, now locale-addressable).

No issues. Mechanical and complete.

PR #3202, opened 2026-08-27
refactor(admin): name the MTP dashboard gate (_is_mtp_settable)

pedroberaldo87:
Follow-up to #3200, no behavior change.

**What it does**

- Adds `_is_mtp_settable(config, model_type)` to `omlx/utils/model_loading.py`,
  right next to `_is_mtp_compatible`: returns true when the model has MTP heads
  AND is either on the generic whitelist or is `qwen4_exp`.
- Replaces the two inline checks that #3200 added in `omlx/admin/routes.py`
  (`_mtp_compat_for_model` and `update_model_settings`), both reading
  `model_type != "qwen4_exp" and not _is_mtp_compatible(...)`, with a call to
  the new function, and collapses the two duplicated comment blocks into the
  function's docstring.

**Why**

The admin gate ("may the toggle be offered?") and the runtime whitelist ("may
the generic mlx-lm patch be applied?") are different questions — qwen4_exp is
a yes to the first and a deliberate no to the second, since it activates
Lightning MTP through its own VLM loader branch. Naming the first question
keeps the next special case from being pasted inline a third time.

**Proof it's behavior-preserving**: the gate tests #3200 added
(`test_qwen4_exp_mtp_admin_gate.py`, plus the admin-settings suites) pass
untouched — 70 tests green.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -696,7 +696,7 @@ def _mtp_compat_for_model(model_info: dict) -> tuple[bool, str]:
     from ..utils.model_loading import (
         _checkpoint_has_mtp_weights,
         _has_mtp_heads,
-        _is_mtp_compatible,
+        _is_mtp_settable,
     )
 
     is_paro, paro_reason = _paroquant_compat_for_model(model_info)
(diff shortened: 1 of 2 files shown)


scubamount:
## Verified

- **Purpose:** pure follow-up to #3200 — extracts the duplicated `model_type != "qwen4_exp" and not _is_mtp_compatible(...)` inline check (2 call sites in `admin/routes.py`) into `_is_mtp_settable(cfg, model_type)` next to `_is_mtp_compatible`.
- **Semantics preserved:** dashboard gate ("may the toggle be offered?") is a deliberate superset of the runtime whitelist — qwen4_exp enters MTP via its own loader branch, not the generic mlx-lm patch block; adding it to `_is_mtp_compatible` would wrongly layer Qwen3.5/3.6 patches. New docstring names that distinction, which was the point.
- **No behavior change:** both call sites keep the embedded `mtp.*` weight check unchanged.

No issues. Good naming/DRY refactor of a subtle two-question gate.

PR #3203, opened 2026-08-27
feat(speculative): prompt-lookup drafting (n-gram history lookup) on the Lightning MTP verify path

### LINKED ISSUE #2673: Feature request: Add Prompt lookup decode or n-gram speculative decoding.
oMLX already supports speculative decoding via DFlash and MTP, but both require either a compatible draft-capable architecture or an MTP head. Coding-agent traffic is also exactly the kind of input-grounded workload like editing existing files, applying diffs, repeating unchanged code etc where a large fraction of the output is a verbatim copy of the prompt/context.  Right now there's no lightweight fallback for these cases that doesn't require an extra draft model, additional VRAM, or a specific model architecture.

Add prompt lookup decoding (PLD, aka n-gram speculative decoding) as a model-free alternative to DFlash/MTP. PLD scans the existing prompt and generated output for a repeat of the current n-gram suffix and, when found, proposes the tokens that followed that match previously as draft tokens. The target model then verifies the whole span in a single forward pass, same as standard speculative decoding, but with zero extra parameters, zero additional VRAM, and no dependency on model architecture (works with any model mlx-lm supports).

On my Macbook M1 Pro, loading DFlash, MTP show a net-negative gain, as it is constrained by the already miserable bandwidth.

I'd like to see:

a.    A --speculative-mode prompt-lookup (or similar) flag/admin-panel toggle, selectable per model alongside the existing DFlash/MTP options.

b.   Configurable n-gram window size and max draft length (mirroring what upstream implementations like apoorvumang/prompt-lookup-decoding and Hugging Face's prompt_lookup_num_tokens expose).

 c.  Automatic fallback to standard decode when no n-gram match is found, with acceptance-rate stats surfaced in the admin benchmark panel (similar to the existing PP/TG numbers) so users can see hit rate per workload.

d.    Since it requires no draft model, it should be safe to enable by default for code-editing/RAG-style requests without the KV prefix cache and OOM issues currently seen with DFlash/MTP.

Doing nothing: decode remains fully memory-bandwidth bound for models without DFlash/MTP support, leaving obvious speedup on the table for the repetitive-text use case oMLX already optimizes for (coding agents).

**Additional context**
PLD is already proven in production tooling (vLLM's prompt_lookup_max, Hugging Face transformers' n-gram speculative decoding) with reported 2–4x speedups on summarization/code-editing tasks and reasonable gains even at lower acceptance rates, unlike draft-model speculative decoding which needs a high acceptance rate to pay off. Given oMLX's target audience is Apple Silicon users running coding agents against a single loaded model with tight unified-memory budgets, this seems like a natural, low-cost addition alongside the existing DFlash/MTP paths rather than a replacement for them. 

Similar to llama.cpp, we could add multiple methods of decode prediction that could potentially offer further boost.

pedroberaldo87:
## Summary

Adds **prompt lookup decoding** — the technique from Apoorv Saxena's *Prompt Lookup Decoding* (2023), also known as n-gram / lookup decoding in vLLM and llama.cpp — as a drafter that rides the existing Lightning MTP draft+verify cycle.

When the most recent tokens of a request already occurred earlier in the same request (prompt or generated text), the drafter copies what followed that occurrence and proposes it as the draft, up to `draft_max` tokens. The model verifies the draft in the same single pass it already uses for the MTP head, so acceptance is lossless and rejection costs one pass. **No model component is involved** — the naming was changed from "n-gram" to "prompt lookup" to avoid confusion with Qwen4-Exp's in-checkpoint n-gram (PLE) table, which is unrelated.

Closes #2673 (*"Add Prompt lookup decode or n-gram speculative decoding"*).

## Measured (GLM-5.3-Flash oQ2e, M1 Ultra 128 GB, model-card sampling: temp 1.0, top_p 0.95, top_k 20)

| workload | without | with prompt lookup |
|---|---|---|
| prose, 5×300 tokens | 24.5 ± 1.0 tok/s (MTP only) | 23.5 ± 0.8 (within noise — nothing to copy) |
| code rewrite present in the prompt (1338-token file, "change 400→422 and return the whole file", 3 runs) | ~24.5 | **25.6 · 31.5 · 33.6 tok/s** — 6–9 tokens/cycle, 95–97% acceptance |
| 900-token copy task | 20–26 | **30.4** (7 tokens/cycle, 95% acceptance) |

Independent confirmation on M5 Max / 128 GB by @latent-variable (see thread), composing with #3235.

## Design

- Rides the MTP verify window: no second model, no extra forward for the draft.
- Per-model settings in the dashboard: match window, min/max draft length, frequency-scaled copy length, chained walk (llama.cpp's ngram-mod walk).
- Requires Lightning MTP (the drafter needs the verify path). Disabled by default.
- Settings key stays `ngram_spec_enabled` for compatibility with persisted per-model settings; only the visible label changed.

## Tests

`tests/test_ngram_draft_source.py` covers the draft source (index, lookup, frequency rule, chained walk); `tests/test_model_settings.py`, `tests/test_admin_model_settings.py`, `tests/test_engine_pool.py`, `tests/test_benchmark.py` and `tests/test_spec_stats.py` cover the settings plumbing, the arming on load, the A/B flip and the telemetry.


latent-variable:
Independent confirmation on different silicon, and it composes with #3235.

**Setup.** M5 Max, 128 GB, macOS 26.5.2. Stack: `a20ec09d` + #3225 + #3235 + this PR, applied clean. `Qwen3.8-Flash-Next-oQ4e-mtp`, `qwen4_ple_ssd_offload` on, `mtp_enabled` on. Settings: `ngram_spec_enabled` + `ngram_spec_freq_rule`, everything else default (so `draft_max` 16, not the 48 in your writeup).

**Method.** Two workloads: a 246-line Python file with "rename `main` to `run_main`, output the complete file" (copy-heavy, the agent case), and 900 words of novel prose (the control). `temperature=0`, thinking off, `max_tokens` fixed so every run emits exactly the same token count and ends on `finish_reason=length`. Each prompt carries a unique nonce so the prefix cache misses every time — verified `cached_tokens=0` on every run. Arms interleaved OFF/ON within each round, 3 rounds, drafter toggled through the admin settings API.

Token counts come from the API's `usage`, not from counting SSE chunks. That matters: MTP packs several tokens into one chunk, so chunk counting undercounts decode by about 5x.

| workload | drafter off | drafter on | |
|---|---|---|---|
| rewrite | 45.2 / 46.6 / 53.3 → **48.3 tok/s** | 105.1 / 112.9 / 107.2 → **108.4 tok/s** | **2.24x** |
| novel prose | 28.8 / 29.2 / 32.2 → **30.0 tok/s** | 27.9 / 34.5 / 32.7 → **31.7 tok/s** | 1.06x, within noise |

Mean tokens per streamed chunk moves 7.4 → 17.0 on the rewrite and 4.6 → 4.7 on the prose, which is the drafter landing wide accepted spans on one workload and standing down on the other. That is the same shape you reported.

**Why it is 2.24x here and 1.55x in your run.** Best guess is #3235. Without it every n-gram row that misses the page cache stalls the forward on an SSD read, and a wide verify touches proportionally more rows than a single-token step, so the drafter's own mechanism pays the worst of that cost. With the gather batched and prefetched, the wide verifies get cheaper and more of the acceptance turns into throughput. Worth someone else reproducing before treating that as established.

`tests/test_ngram_draft_source.py` and `tests/test_spec_stats.py` both green (18 tests). One unrelated failure, `test_qwen4_lightning_mtp_fusion_and_runtime_attachment`, reproduces on pristine `a20ec09d` on this machine, so it is not from this PR — filing separately.

pedroberaldo87:
Thanks for the reproduction.

On the #3235 hypothesis, here is the number from the other side. With `qwen4_ple_ssd_offload` off — table resident — #3235 adds nothing:

```
prefill, with #3235:   475.9 | 531.0 | 532.2  →  mean 513.0
prefill, without:                                mean 500.2  (n=6)
                                                 +2.6%, one standard error
```

No page-cache miss to pay for, nothing for the prefetch to save.

Also worth noting: your 2.24x is with `draft_max` at its default of 16, not the 48 in my writeup.

scubamount:
## Verified

- **Design:** n-gram history-lookup drafting riding the Lightning MTP verify cycle — suffix repeat in the request's own tokens proposes the following tokens as draft, verified in one wide forward. Lossless by construction (deterministic draft only reduces acceptance ratio; verify unchanged).
- **Fallbacks:** on miss the regular MTP chain runs; head fold still runs on hit cycles so MTP cache stays warm — correct bookkeeping concern, handled.
- **Memory guard:** i18n hints document the verify-width memory bound (max draft 16 safe default on 128 GB) — the right knob to expose.
- **UI:** dashboard label generalized DFlash→Speculation, feature-flag spec + benchmark wiring added consistently.

No blocking issues. Large feature (1718 lines); the frequency-scaled copy-length rule is the part to watch in benchmark follow-up.

PR #3204, opened 2026-08-27
fix(cluster): stage safetensors shards referenced in a subdirectory

phaedrus1992:
## Summary

Model staging assumed every safetensors shard listed in `model.safetensors.index.json` lives flat in the model root. That breaks on VLM checkpoints that ship an auxiliary shard in a subdirectory — OptiQ's `optiq/optiq_vision.safetensors` is one example — which crashed staging with "unsafe safetensors filename in index".

## Changes

- `index_shards`/`_indexed_shards` and the remote-shard-inventory path now accept a safe, one-level-nested filename (rejecting `..` traversal and absolute paths) instead of refusing every path containing a `/`.
- A nested shard is treated as auxiliary rather than run through the layer-number regex, since a vision encoder can reuse "blocks.N"/"layers.N" naming with its own unrelated numbering — misattributing it to a decoder layer would be silently wrong, not just rejected.
- `scp_push`/`scp_copy` and the remote install/file-size snippets create the parent directory before writing, so the nested destination path actually exists on the peer.

## Test plan

- [x] `pytest tests/test_cluster_staging.py` — new coverage for indexing, staging, and transferring a one-level-nested shard, plus the existing flat-shard suite.

### DIFF
--- a/omlx/cluster/staging.py
+++ b/omlx/cluster/staging.py
@@ -32,6 +32,38 @@ _LAYER = re.compile(r"(?:^|\.)(?:layers|h|blocks|block)\.(\d+)(?:\.|$)")
 _MAX_HEADER_BYTES = 64 * 1024 * 1024
 _LOCAL_HOSTS = {"127.0.0.1", "localhost", "::1"}
 
+
+def _validate_safe_relative_filename(filename: str) -> None:
+    """Reject a path that escapes the model root or names an absolute location.
+
+    A one-level subdirectory is legitimate — OptiQ VLM checkpoints index
+    ``optiq/optiq_vision.safetensors`` alongside the numbered decoder shards —
+    so only ``..`` traversal and absolute paths are refused, not every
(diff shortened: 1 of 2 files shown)


scubamount:
## Verified

- **Root cause correct:** staging assumed flat shard layout; OptiQ VLM checkpoints index `optiq/optiq_vision.safetensors` → "unsafe safetensors filename in index" crash.
- **Fix right:** `_validate_safe_relative_filename` refuses absolute paths and any `..` part but allows one-level nesting; `_is_auxiliary_shard` (nested path) excludes auxiliary files from the `_LAYER` attribution pass — the reasoning is sound: vision encoders reuse `blocks.N`/`layers.N` numbering, so layer-index matching would attribute them to decoder layers by coincidence.
- **Security posture maintained:** traversal rejection is explicit and tested; flat-shard pipeline splitting unchanged.

No issues.

PR #3205, opened 2026-08-27
fix(cluster): stop hiding a self-assigned Thunderbolt Bridge address

phaedrus1992:
## Summary

A direct Thunderbolt cable with no DHCP server self-assigns a 169.254.0.0/16
(APIPA/link-local) address on macOS's `bridge0` interface. Candidate-link
discovery excluded every 169.254.0.0/16 address as "unroutable", which also
threw out this legitimate Thunderbolt Bridge address — so a real, fast,
directly-cabled link was never offered as a candidate at all.

## Changes

- Split the old blanket `_UNROUTABLE_NETWORKS` exclusion into
  `_LOOPBACK_NETWORK` (always excluded) and the full tuple, which now only
  gates `_interface_ip`'s own use.
- `_interface_address` excludes only loopback, so a link-local Thunderbolt
  Bridge address is included as a link candidate.

Verified end-to-end on real hardware: the route resolves to `bridge0`, ping
is 0.4ms, and `detect_transports` independently confirms `kind='thunderbolt'`
at 40 Gb/s (TB4).

## Test plan

- [x] `pytest tests/test_cluster_link_status.py`

### DIFF
--- a/omlx/cluster/transport.py
+++ b/omlx/cluster/transport.py
@@ -1027,13 +1027,22 @@ def assess_link(
 # moment it is read off the host. Nothing below caches; that is the point.
 # ---------------------------------------------------------------------------
 
-# Two hosts can carry these at once without being able to reach each other:
-# loopback is per-host, and a 169.254 address means DHCP failed, so unrelated
-# Macs agree on the subnet and on nothing else.
-_UNROUTABLE_NETWORKS = (
-    ipaddress.ip_network("127.0.0.0/8"),
-    ipaddress.ip_network("169.254.0.0/16"),
-)
+# Loopback is per-host — two unrelated Macs both carrying a 127.0.0.1 proves
+# nothing about being able to reach each other. No verification could ever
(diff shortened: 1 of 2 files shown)


scubamount:
## Verified

- **Root cause correct:** blanket `_UNROUTABLE_NETWORKS` exclusion hid every 169.254/16 from candidate discovery — but a direct Thunderbolt cable with no DHCP legitimately self-assigns APIPA on `bridge0` (Apple's recommended no-server setup), so the fastest link was never offered.
- **Fix right:** splits loopback (always excluded — per-host, proves nothing) from 169.254 (untrusted for bare *subnet matching* only). `_interface_address` now keeps APIPA addresses; `verify_link_reachability`'s route+ping/TCP check is the layer that proves or refutes each candidate.
- **Layering argument is sound:** the comment explains exactly why range-guessing belongs at verification, not discovery. Test updated to assert 169.254.138.14 IS kept and 127.0.0.1 is not.

No issues.

PR #3206, opened 2026-08-27
fix(cluster): fall back to a TCP probe when link ping fails

phaedrus1992:
## Summary

Link verification relied on a single ICMP ping to confirm a candidate link
is actually reachable. A single dropped or filtered ping falsely marked a
genuinely working link as unreachable, discarding a real candidate over one
bad packet.

## Changes

- Added a TCP probe (`_tcp_probe_succeeds`) used as a fallback in both the
  route-mismatch and ping-failure branches of `verify_link_reachability`,
  so one lost ICMP packet no longer disqualifies a link that answers fine
  over TCP.

## Test plan

- [x] `pytest tests/test_cluster_link_status.py`

### DIFF
--- a/omlx/cluster/transport.py
+++ b/omlx/cluster/transport.py
@@ -1502,23 +1502,8 @@ def verify_link_reachability(
         route = run(local.host, ("/sbin/route", "-n", "get", remote.address))
         selected = _route_interface(route.stdout) if route.returncode == 0 else ""
         if selected != local.interface:
-            # Linux does not implement macOS's ``route -n get`` form. Binding
-            # a TCP connection to the candidate source address proves both the
-            # route and that the peer's SSH service answers on that exact path,
-            # without needing a platform-specific interface command.
-            script = (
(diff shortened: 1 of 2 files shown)


scubamount:
## Verified

- **Root cause correct:** single ICMP ping (`-c 1 -W 1000`) with no retry — macOS stealth-mode firewall drops ICMP by default, so one filtered packet disqualified a genuinely working link.
- **Fix right:** extracts the already-trusted bound-source TCP connect (peer port 22, `source_address=(local.address, 0)`) into `_tcp_probe_succeeds`, used as fallback in BOTH the route-mismatch and ping-failure branches — proves route and SSH reachability on the exact source/dest pair without platform-specific interface commands.
- **Consolidation:** the previously duplicated inline script becomes one `_TCP_PROBE_SCRIPT` constant; behavior in the route-mismatch branch is unchanged.

No issues. Good de-duplication plus one bad packet tolerance.

PR #3208, opened 2026-08-27
fix(cluster): retry flaky mDNS resolution across cluster SSH probes

phaedrus1992:
## Summary

mDNS resolution for a `.local` peer is intermittently flaky — a single
dropped multicast round trip made a perfectly healthy peer look permanently
unreachable, surfacing in several different places as unrelated-looking
symptoms:

- `has no routable IPv4 address` during interface/link probing
- `RDMA device probe failed ... Could not resolve hostname`
- `could not inspect staged model files ... Could not resolve hostname`
- `stopped publishing its runtime heartbeat` right after a fresh activation,
  because the very first liveness check landed in the race window between
  a rank starting up and writing its first heartbeat marker

None of these are permanent failures — the peer answers fine a moment
later. Retrying a bounded few times before giving up fixes all four without
masking a genuinely unreachable host, since every caller here is a
read-only query.

## Changes

- Added `run_ssh_retrying` to `omlx/cluster/ssh_policy.py`: one shared
  retry helper (3 attempts, 0.5s delay) used by every cluster SSH call
  site. Its `runner` parameter resolves `subprocess.run` fresh on each
  call rather than capturing it as a default argument, so a test's
  monkeypatched `subprocess.run` is still respected by a caller that
  doesn't pass `runner` explicitly.
- `transport.py`'s `_read` (interface/link probing) and `_rdma_devices`
  (RDMA device probing) now delegate to it. `_rdma_devices` only retries
  its remote branch — the local branch runs on every activation-progress
  poll (every 750ms) and fails deterministically on hardware with no RDMA
  device, so retrying it would add felt latency for no chance of success.
- `staging.py`'s `run_remote_python` (remote model-file inspection, shard
  indexing, model-directory resolution) now delegates to it.
- `liveness.py`'s `probe_peer` and `read_remote_marker` (peer heartbeat
  checks) now delegate to it too, replacing a second, near-identical
  retry implementation that predated this consolidation.

## Test plan

- [x] `pytest tests/test_cluster_ssh_policy.py tests/test_cluster_liveness.py tests/test_cluster_link_status.py tests/test_cluster_staging.py tests/test_cluster_autoconfigure.py tests/test_cluster_routes.py tests/test_cluster_remote_planning.py`

### DIFF
--- a/omlx/cluster/liveness.py
+++ b/omlx/cluster/liveness.py
@@ -35,7 +35,7 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 
-from .ssh_policy import cluster_ssh_options
+from .ssh_policy import cluster_ssh_options, run_ssh_retrying
 
 # A rank refreshes its marker on every phase change, while serving, and on a
 # fixed heartbeat interval even when idle (``RuntimeTelemetry``). Three missed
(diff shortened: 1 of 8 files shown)


scubamount:
## Verified

- **Root cause correct:** single dropped mDNS multicast round-trip made a healthy `.local` peer look permanently unreachable, surfacing as four unrelated-looking errors across link probing, RDMA device probe, staged-file inspection, and heartbeat.
- **Fix right:** one `run_ssh_retrying` helper at the ssh_policy layer, applied at every probe call site (liveness probe, marker read, ...) — fix where all callers route, not per-site retry duplication.
- **Consolidation:** previous per-site try/except around `runner(...)` collapses into the shared helper; return-code semantics preserved.

No issues. Correct root-level fix for a symptom that appeared in four places.

PR #3214, opened 2026-08-27
fix(scheduler): convert QSA singleton caches to batched form before extend

HaloFour:
## Summary

- Let the scheduler's `_to_batched_cache_layer` fall back to a cache's own `to_batch()` conversion for model-owned singletons outside the `mlx_lm` cache hierarchy (e.g. `QSAKVCache` from the Qwen4-Exp overlay).
- Batched caches never define `to_batch`, so an already-batched layer passes through untouched and is still extended in place.

## Problem

Follow-up to #3188, which fixed the `_make_cache` seam. The singleton pass-through optimization keeps a one-row batch's cache as a singleton (`QSAKVCache` has `filter` + `extract`). When a second row later joins the batch — e.g. after an MTP late-join handoff — `GenerationBatch.extend` → `_extend_cache_layer` crashes because the singleton has no `extend` (only `BatchQSAKVCache` does):

```text
AttributeError: 'QSAKVCache' object has no attribute 'extend'
```

The scheduler's corruption handler then clears the entire cache and re-prefills, so throughput benchmarks on Qwen3.8 Flash Next (QSA) models take a large, recurring perf hit.

Fixes #3213

## Validation

- New regression tests reproduce the exact `AttributeError` at `scheduler.py:1009` without the fix and pass with it; they verify KV rows, offsets, left-padding, and QSA indexer keys/positions survive the conversion, and that the singleton pass-through and batched in-place extend behaviors are unchanged.
- Focused cache/scheduler/batching suites: `522 passed`
- Full suite excluding slow/integration: `10402 passed`; the 46 failures are byte-identical with and without the change (pre-existing dflash/web-search/float-tolerance issues).

Issue #3213, opened 2026-08-27
Cache corruption error running throughput benchmarks for Qwen3.8-Flash-Next-oQ4e-mtp


HaloFour:
Running the throughput benchmarks for the 0.6.3 release with `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp` and getting the following error:

`Cache corruption not recoverable after retries: 'QSAKVCache' object has no attribute 'extend'`

Log shows the following traceback:

```
2026-08-27 14:32:27,998 - omlx.scheduler - WARNING - [-] - Cache corruption detected: 'QSAKVCache' object has no attribute 'extend', clearing cache and re-prefilling...
2026-08-27 14:32:27,998 - omlx.scheduler - DEBUG - [-] - Cache corruption traceback:
Traceback (most recent call last):
  File "/Applications/oMLX.app/Contents/Resources/omlx/scheduler.py", line 11651, in step
    responses = list(self.batch_generator.next_generated())
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Applications/oMLX.app/Contents/Resources/Python/framework-mlx-base/lib/python3.11/site-packages/mlx_lm/generate.py", line 1866, in next_generated
    prompt_responses, generation_responses = self._next()
                                             ^^^^^^^^^^^^
  File "/Applications/oMLX.app/Contents/Resources/omlx/patches/mlx_lm_mtp/batch_generator.py", line 296, in patched_bg_next
    return original_bg_next(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Applications/oMLX.app/Contents/Resources/Python/framework-mlx-base/lib/python3.11/site-packages/mlx_lm/generate.py", line 1819, in _next
    self._generation_batch.extend(gen_batch)
  File "/Applications/oMLX.app/Contents/Resources/omlx/patches/mlx_lm_mtp/batch_generator.py", line 237, in patched_extend
    result = original_extend(self, batch, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Applications/oMLX.app/Contents/Resources/Python/framework-mlx-base/lib/python3.11/site-packages/mlx_lm/generate.py", line 1296, in extend
    self.prompt_cache = _extend_cache(self.prompt_cache, batch.prompt_cache)
[...]

PR #3219, opened 2026-08-27
fix(cache): normalize mixed Qwen4 QSA text/MRoPE positions during prefix-cache reconstruction

rsnow:
## Summary

- Promote serialized one-channel Qwen4 QSA text positions to the three-channel MRoPE form before concatenating persisted cache blocks, matching the runtime's own position-handling rule.
- Fix silent prefix-cache reconstruction failure on Qwen3.8 Flash Next (`qwen4_exp`) multi-turn sessions whose cached prefix mixes text-only and MRoPE blocks.
- Keep the change narrowly scoped to `Qwen4QSAKVCacheHandler`; generic N-state reconstruction is untouched.

## Problem

On a long multi-turn Qwen3.8 Flash Next session, prefix discovery correctly identified a large exact cached prefix, but reconstruction failed and the scheduler fell back to a full re-prefill:

```text
common_prefix=28672/28672 tokens (~14 blocks of 2048)
Failed to reconstruct cache
re-prefills 39228 of 39228 tokens (reused 0)
```

The QSA position slices were thirteen blocks shaped `(1, 1, 2048)` and one shaped `(1, 3, 2048)`. A deterministic component reproduction builds that topology and calls `Qwen4QSAKVCacheHandler.concatenate_states`; `mx.concatenate(..., axis=2)` rejects the mixed channel dimensions.

## Root cause

`_serialize_qsa_positions` stores both valid model-facing layouts as `[B, C, S]`:

- text `[B, S]` → `[B, 1, S]`
- MRoPE `[3, B, S]` → `[B, 3, S]`

Block reconstruction concatenates `index_position_ids` on sequence axis 2, but all non-sequence dimensions must also match. A chain containing both valid forms therefore fails before deserialization.

The Qwen4 runtime already defines the reconciliation rule: `_append_indexer_positions` broadcasts the 2-D text form across the three MRoPE coordinates when the forms meet, then concatenates on the sequence axis. Prefix reconstruction was missing the equivalent operation for the channel-explicit serialized layout.

## Fix

A dedicated `_normalize_qsa_position_states` helper, used only for `index_position_ids` in `Qwen4QSAKVCacheHandler.concatenate_states`:

- validates every state as rank 3, with 1 or 3 channels and a consistent batch dimension;
- leaves homogeneous all-text and all-MRoPE chains unchanged;
- promotes `[B, 1, S]` states to `[B, 3, S]` only when a chain is mixed.

The existing deserializer then restores model-facing `[3, B, S]`, matching incremental runtime behavior. No other cache type or state element is changed.

## Tests

The new regression suite covers:

- mixed text→MRoPE, MRoPE→text, and multiple alternations;
- homogeneous all-text and all-MRoPE chains;
- invalid rank, channel count, and batch size, including malformed later states;
- the original 14 × 2,048-token topology with exact per-coordinate assertions.

Results:

- focused suite: **11 passed**
- relevant cache suites: **266 passed**
- component canary: **28,672 cached tokens reused; 10,556 re-prefilled**

## Live validation

Validated on one Mac Studio M3 Ultra running oMLX 0.6.3 with `Qwen3.8-Flash-Next-oQ4e-mtp`:

- a ~42K-token multi-turn workload served all **40,960** common-prefix tokens with no reconstruction warning;
- before the patch, the same workload class failed reconstruction and reused 0 of 38,912 matching tokens;
- a smaller follow-up served all **12,288** cached prefix tokens.

This is one-machine, one-model validation, not broad compatibility coverage.

## Relationship to #3213 / #3214 / #3215

Those concern a separate continuous-batching seam in `omlx/scheduler.py`: a singleton `QSAKVCache` reaches `_extend_cache_layer` without `.extend` when a second request joins the batch.

This change concerns persisted prefix-cache reconstruction in `omlx/cache/type_handlers.py`: mixed valid position channel counts make block concatenation fail during multi-turn prefix reuse, with no concurrent request required. Different file, trigger, symptom, and lines; the fixes are independent.

## Scope

No API, dependency, checkpoint, or model-math change. Homogeneous caches follow the existing path. Unsupported future coordinate counts fail clearly rather than corrupting state.

### DIFF
--- a/omlx/cache/type_handlers.py
+++ b/omlx/cache/type_handlers.py
@@ -1451,6 +1451,55 @@ def _deserialize_qsa_positions(position_ids: Any) -> Any:
     return position_ids.transpose(1, 0, 2)
 
 
+def _normalize_qsa_position_states(position_states: list[Any]) -> list[Any]:
+    """Promote mixed text/MRoPE QSA positions to [B, 3, S].
+
+    Qwen4 represents text positions as [B, S] and MRoPE positions as
+    [3, B, S]. Serialization stores those as [B, 1, S] and [B, 3, S]
+    respectively. The model's indexer cache promotes the text form by
+    broadcasting it across all three MRoPE coordinates when the two forms
+    meet, so block reconstruction must perform the equivalent operation
+    before concatenating along the sequence axis.
+    """
+    if not position_states:
+        return position_states
+
+    batch_size = None
+    channels: set[int] = set()
+    for position_state in position_states:
+        if not hasattr(position_state, "ndim") or position_state.ndim != 3:
+            shape = getattr(position_state, "shape", None)
+            raise ValueError(
+                "Serialized QSA position IDs must be [B, C, S], "
+                f"got {shape}."
+            )
+        current_batch, current_channels, _ = position_state.shape
+        if current_channels not in (1, 3):
+            raise ValueError(
+                "Serialized QSA position IDs require 1 text channel or 3 "
+                f"MRoPE channels, got {position_state.shape}."
+            )
+        if batch_size is None:
+            batch_size = current_batch
+        elif current_batch != batch_size:
+            raise ValueError(
+                "Serialized QSA position IDs must have a consistent batch "
+                f"dimension, got {batch_size} and {current_batch}."
+            )
+        channels.add(current_channels)
+
+    # Preserve both homogeneous paths without introducing broadcast views.
+    if len(channels) == 1:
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the focused fix. Normalizing serialized one-channel QSA text positions only when mixed with three-channel MRoPE states matches the runtime append rule and keeps homogeneous reconstruction unchanged. It also applies cleanly to current main. LGTM, merging.

Issue #3220, opened 2026-08-27
Qwen3.8-Flash-Next Lightning MTP crashes with hidden-state shape error


blahjovic1:
## Description

Lightning MTP crashes during generation with `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp` on oMLX 0.6.3.

The same model works correctly when Lightning MTP is disabled. SSD N-gram Offload also works correctly and reduces actual resident model memory from approximately 103.94 GB to 68–69 GB.

## Environment

- oMLX: 0.6.3, build 2500
- macOS: 26.6
- Hardware: Mac Studio M2 Ultra, 128 GB unified memory
- Model: `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp`
- Architecture: `qwen4_exp`
- Context window: 65,536
- SSD N-gram Offload: Enabled
- Lightning MTP: Enabled when reproducing the crash
- Thinking: Enabled
- Reasoning effort: Medium
- Concurrency: 1
- TurboQuant KV: Disabled
- SpecPrefill: Disabled
- DFlash: Disabled
- VLM MTP: Disabled

## Steps to reproduce

1. Install oMLX 0.6.3.
2. Download `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp`.
3. Enable SSD N-gram Offload.
4. Enable Lightning MTP.
5. Unload and reload the model.
6. Send a normal text chat-completion request.

## Actual behavior

Streaming generation fails with:

```text
ValueError: Qwen4 Lightning MTP expects hidden shape
[batch, tokens, hc_count * hidden_size].
```

Relevant stack trace:

```text
File "omlx/patches/mlx_lm_mtp/batch_generator.py",
line 2356, in _chain_next_drafts
    logits, head_hidden = model.mtp_forward(

File "omlx/models/vlm.py",
line 100, in mtp_forward
    return self._language_model.mtp_forward(

File "omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py",
line 1782, in mtp_forward
    mtp_output, hc_hidden = mtp(

File "omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py",
line 1679, in fuse_inputs
    raise ValueError(

ValueError: Qwen4 Lightning MTP expects hidden shape
[batch, tokens, hc_count * hidden_size].
```

The server then reports:

```text
[...]

Issue #3212, opened 2026-08-27
MTP error v0.6.3 Qwen3.8-Flash-Next-oQ4e-mtp


gaohe1984:
Qwen3.8-Flash-Next-oQ4e-mtp
When lightning MTP is turned-on, the following error came up:

File ".../vendor/mlx_vlm/models/qwen4_exp/language.py", line 1679, in fuse_inputs
    raise ValueError(
ValueError: Qwen4 Lightning MTP expects hidden shape [batch, tokens, hc_count * hidden_size].

riverhin:
Same here. 
M3 Ultra 60c; idem on M5 Max 40c. 
Model: Jundot/Qwen3.8-Flash-Next-oQ4e-mtp

Is there a way I can provide more menanigful information for somebody more competent thn me ;) to more easily debug this?

jasonnickel:
Datapoint from 0.6.3 (Homebrew keg, M3 Ultra 256 GB, macOS 27), `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp`, Lightning MTP on, `qwen4_ple_ssd_offload` on:

- A fresh server process that loads Flash-Next first serves it fine - 300+ requests (short and up to 9.6K-token prompts) with MTP active, acceptance 61-98%.
- The `Qwen4 Lightning MTP expects hidden shape [batch, tokens, hc_count * hidden_size]` error started only after Flash-Next was **unloaded and reloaded into the same process** (another model - Qwen3.6-35B-A3B - had been loaded in between). From that point every request failed (21/21, 0 tokens) until the process was restarted; a new process with Flash-Next loaded first was clean again (100+ requests so far, 0 errors).

So at least one trigger is model reload within a running server (engine_pool evict/reload), not the first load.

HaloFour:
I only got this error when I tried to enable Lightning MTP and TurboQuant at the same time for this model.  Disable either and everything works fine.

latent-variable:
I opened #3225 to fix this.

[...]

Issue #3211, opened 2026-08-27
Serving sparse MoE checkpoints larger than the memory ceiling


TobiasvonCarnap:
## What

Allow oMLX to serve a MoE checkpoint whose total size exceeds the configured
memory ceiling, by keeping only the hot subset of expert weights resident and
faulting the rest in on demand — instead of rejecting the load outright with
`ModelTooLargeError`.

## Why it could work for MoE specifically

Sparse MoE touches a small fraction of its weights per token. Qwen3.8-Flash-Next
is 125B total with 6B active — 10 routed plus 1 shared out of 512 experts — with
an additional 51B n-gram embedding that is a lookup table, so only the rows that
actually occur are read. The working set per token is therefore a small multiple
of the active parameter count, not of the file size.

Expert usage is also skewed in practice, so a hot subset covers most routing
decisions and the rest can live on SSD.

This does **not** generalise to dense models, where every forward pass touches
every weight. The request is deliberately limited to MoE.

## Evidence, with caveats

Atomic Chat reports an 85 GB quant of Qwen3.8-Flash-Next running on a 64 GB
M5 Max at 36 tok/s, with vision, via llama.cpp's mmap path — a file larger than
the machine's RAM. Two caveats worth stating plainly: it is a vendor claim about
their own app, and no context length is given. Their llama.cpp examples use
`-c 8192`, which is where the page-cache headroom is largest, so the number may
not survive at agentic context lengths.
https://huggingface.co/AtomicChat/Qwen3.8-Flash-Next-GGUF
I have not measured this myself and cannot confirm the throughput.

## Prior art inside oMLX

The tiered KV cache already implements the same idea for a different data class:
hot blocks in RAM, cold blocks spilled to the paged SSD cache, LRU eviction
between them. DFlash carries its own L1/L2 split on top. Extending residency
management from KV blocks to expert weights would reuse an established pattern
rather than introduce a new one.

## Open questions I cannot answer

- **Metal residency.** llama.cpp maps the GGUF and lets the OS page cache manage
  residency. Whether MLX arrays can be backed the same way, and whether Metal
  tolerates non-resident pages inside a dispatch, is the crux — and I do not know
  the answer. If GPU-side buffers must be wired for the duration of a kernel,
  the mechanism would need explicit staging rather than transparent mmap.
- **Interaction with the KV cache at long context.** On a machine where the model
  barely fits, every GB of KV cache displaces exactly the expert pages that make
[...]

PR #3224, opened 2026-08-27
fix(vlm): apply the sanitize wrapper to HuggingFace-cache symlinked models

ChildOTK:
## The problem

`Jundot/Qwen3.8-Flash-Next-oQ4e-mtp` will not load on 0.6.3. M5 Max 128 GB, Homebrew install, model symlinked into `~/.omlx/models/` from the HuggingFace cache.

```
ValueError: Received 384 parameters not in model:
language_model.model.layers.1.ple.ple_embedding.ngram_embedding.shards.0.biases,
language_model.model.layers.1.ple.ple_embedding.ngram_embedding.shards.0.scales,
language_model.model.layers.1.ple.ple_embedding.ngram_embedding.shards.0.weight,
...
RuntimeError: VLM load failed ... LLM fallback also failed: Model type qwen4_exp not supported.
```

It fails 158 ms in having allocated 153 MB, so it never gets near loading weights. The 384 are exactly the 128 PLE shards times `weight`, `scales` and `biases`. The single `ngram_embedding.weight_scale` matches.

## What is happening

`_force_qwen4_exp_sanitize_on_load` wraps `safetensors.safe_open` so `metadata()` hides `format=mlx`. That is what gets mlx-vlm to run `Model.sanitize` on an MLX-format checkpoint, and sanitize is where the PLE n-gram keys get popped when the runtime picks mmap mode. In mmap mode `DiskBackedShardedEmbedding` reads those shards through file readers and registers no parameters for them, so if the keys are not removed they arrive at `update()` with nothing to bind to.

The wrapper never gets applied. The guard did:

```python
if path.parent == target_dir and path.suffix == ".safetensors":
```

A model served from the HuggingFace cache is a symlinked snapshot, and its shards are symlinks into a sibling `blobs/` directory where files are named by hash with no extension. So resolving first loses the directory match and the `.safetensors` suffix together, and the guard fails on both counts.

Concretely, for `~/.omlx/models/Qwen3.8-Flash-Next-oQ4e-mtp/model-00001-of-00021.safetensors`:

```
target_dir          .../snapshots/2615fc0e97...
resolved.parent     .../blobs
resolved.suffix     ''
```

## The change

Take the suffix from the path as handed in, and accept either the resolved parent or the lexical parent's resolution.

This is a path-handling bug rather than a model one, and the same guard appears in both `_force_qwen4_exp_sanitize_on_load` and `_force_minimax_m3_moe_sanitize_on_load`, so both are changed. A model directory that is not a symlink keeps the original behaviour exactly, since the first disjunct is unchanged.

## Testing

On the machine above, from the HuggingFace cache symlink, with no other workaround in place:

- Before: fails as above, reproducible from both the CLI and the admin UI.
- After: loads in about 27 s, 69.7 GB resident against a 103.9 GB full model, and answers correctly. `PLE mode: mmap` and `Lightning MTP enabled` both engage.
- `Qwen3.6-35B-A3B-oQ6-mtp` still loads and responds, so nothing regressed for a normal model.

I confirmed the mechanism rather than inferring it. With logging added temporarily to the guard, the match went from false to true and `metadata()` began being called with `format=mlx`. The final de-instrumented file was then loaded again from a clean restart and behaved the same.

## What the MiniMax hunk does, since I could not test it

No MiniMax model on this machine, so that half is unexercised. Worth being precise about what it changes rather than leaving you to guess.

It does not add a code path. `_force_minimax_m3_moe_sanitize_on_load` exists to make sanitize run, and today it silently fails to do that for any MiniMax model served from an HF cache symlink. The change makes those models behave the same as a MiniMax model sitting in a plain directory, where the wrapper already applies and sanitize already runs.

So the population affected is MiniMax users on an HF symlink, who are currently getting the unsanitized path. Whether such a checkpoint loads differently once sanitize starts running is the part I cannot answer, and you can. Happy to drop that hunk if you would rather take only the qwen4_exp half.

## Notes

`path.parent.resolve()` sits outside the `try`, so a symlink loop would raise rather than fall through. The original had the same exposure through `path.resolve()`, so this is not new, but it is there if you want it tightened.

Related, though none of them cover this: #3221 fixes the compat patch never being applied on the conversion path, where this is the serving path and the patch is applied but the sanitize it installs gets skipped. #3212 and #3220 are the Lightning MTP `fuse_inputs` shape error. #3211 is the memory-ceiling feature request.

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -959,10 +959,20 @@ def _force_minimax_m3_moe_sanitize_on_load(model_dir: Path):
     def _patched_safe_open(filename, *args, **kwargs):
         handle = original_safe_open(filename, *args, **kwargs)
         try:
-            path = Path(filename).resolve()
+            path = Path(filename)
+            resolved = path.resolve()
         except TypeError:
             return handle
-        if path.parent == target_dir and path.suffix == ".safetensors":
+        # Judge the path as handed to us, not only the fully resolved one. A
+        # model served from the HuggingFace cache is a symlinked snapshot whose
+        # shards point into a sibling blobs/ directory, where files are named by
+        # hash with no extension. Resolving first therefore loses both the
+        # directory match and the .safetensors suffix, so this wrapper never
+        # gets applied and sanitize stays skipped for format=mlx checkpoints.
+        if path.suffix == ".safetensors" and (
+            resolved.parent == target_dir
+            or path.parent.resolve() == target_dir
+        ):
             return _SafeOpenMetadataWrapper(handle)
         return handle
 
@@ -1080,10 +1090,20 @@ def _force_qwen4_exp_sanitize_on_load(model_dir: Path):
     def _patched_safe_open(filename, *args, **kwargs):
         handle = original_safe_open(filename, *args, **kwargs)
         try:
-            path = Path(filename).resolve()
+            path = Path(filename)
+            resolved = path.resolve()
         except TypeError:
             return handle
-        if path.parent == target_dir and path.suffix == ".safetensors":
+        # Judge the path as handed to us, not only the fully resolved one. A
+        # model served from the HuggingFace cache is a symlinked snapshot whose
+        # shards point into a sibling blobs/ directory, where files are named by
+        # hash with no extension. Resolving first therefore loses both the
+        # directory match and the .safetensors suffix, so this wrapper never
+        # gets applied and sanitize stays skipped for format=mlx checkpoints.
+        if path.suffix == ".safetensors" and (
+            resolved.parent == target_dir
+            or path.parent.resolve() == target_dir
+        ):
             return _SafeOpenMetadataWrapper(handle)
         return handle
 


zhnext:
Confirming this independently, on different hardware and a different checkpoint. I hit the same failure and filed #3239 a few hours before finding this PR — same root cause, same two-count guard failure. #3239 should be closed by this.

  ### Second reproduction

  - Mac15,14 (M3 Ultra), 96 GiB, macOS 26.6.2, oMLX 0.6.3, Homebrew install
  - `Vontra/Qwen3.8-Flash-Next-MLX-oQ4-MTP` and `-oQ3-MTP`, symlinked into `~/Models/` from the HuggingFace cache
  - Same guard, same both-halves failure: `resolved.parent` lands in `blobs/`, and the content-hash blob name has no `.safetensors` suffix

  The failure shape differs slightly from yours, which may be worth recording — **460** extra parameters rather than 384, and in the **un-renamed** `shard_0.*` spelling rather than `shards.0.*`:

  ```
  Received 460 parameters not in model:
  language_model.model.layers.1.ple.ple_embedding.ngram_embedding.shard_0.biases,
  language_model.model.layers.1.ple.ple_embedding.ngram_embedding.shard_0.scales,
  language_model.model.layers.1.ple.ple_embedding.ngram_embedding.shard_0.weight,
  ```

  That spelling is itself a symptom worth having on record: `sanitize` rewrites `shard_N` → `shards.N` via `_NGRAM_SHARD_RE`, so keys arriving in the `shard_0` form show sanitize never ran at all, rather than running and failing to drop. And 460 exceeds 128 × 3, so on this checkpoint the skipped sanitize leaves extras beyond the PLE shards as well — the blast radius is a little wider than one checkpoint shows.

  ### Verified this PR fixes it

  Applied your diff to a clean worktree of `main` (`a20ec09d`) and ran a regression test that builds the HF-cache shape — snapshot directory with weights symlinked into a sibling `blobs/`. It fails on stock `main` and passes with this PR:

  ```
  # stock main
  E   AssertionError: assert {'format': 'mlx', 'source': 'test'} == {'source': 'test'}

  # with this PR
  7 passed
  ```

  `test_vlm_engine.py`, `test_mlx_vlm_minimax_m3_compat.py`, `test_model_loading.py` and `test_mlx_vlm_qwen4_exp_compat.py` also pass (199 total), so the MiniMax hunk does not disturb its existing coverage — which of course is not the same as exercising a real MiniMax checkpoint, and you were right to flag that.

  Separately, a build carrying this fix has been serving the oQ4 checkpoint on the machine above.

  ### Offering a regression test

  This PR has no test, and the existing coverage is precisely why the bug survived release: `test_qwen4_exp_mlx_metadata_is_hidden_during_load` uses a plain `tmp_path` with a real file, so it never exercises the symlink shape that breaks the guard. Happy to open this against `ChildOTK:fix/sanitize-wrapper-hf-cache-symlinks`, or you can lift it directly:

  ```python
  def test_qwen4_exp_mlx_metadata_is_hidden_for_hf_cache_symlinks(tmp_path, monkeypatch):
      """HuggingFace-cache checkpoints expose weights as symlinks into a sibling
      ``blobs/`` directory. Resolving the file before matching pointed the guard at
      ``blobs/`` and at a suffix-less content-hash name, so the ``format=mlx``
      marker stayed visible, mlx-vlm skipped ``Model.sanitize``, and the PLE shard
      tensors it drops in mmap mode reached ``load_weights`` as extras.
      """
      blobs = tmp_path / "blobs"
      blobs.mkdir()
      blob = blobs / ("d417144635b37be24467586d06749611" "d3de81003663205c83db0b1c94a5c903")
      blob.touch()

      snapshot = tmp_path / "snapshots" / "43a82b3f0ff64fa417fd09ca046580f08d19b0d6"
      snapshot.mkdir(parents=True)
      (snapshot / "config.json").write_text(
          json.dumps({"model_type": "qwen4_exp"}), encoding="utf-8"
      )
      weight_file = snapshot / "model.safetensors"
      weight_file.symlink_to(blob)

      class FakeHandle:
          def __enter__(self):
              return self

          def __exit__(self, *_args):
              return False

          def metadata(self):
              return {"format": "mlx", "source": "test"}

      import safetensors

      original = lambda *_args, **_kwargs: FakeHandle()
      original = lambda *_args, **_kwargs: FakeHandle()
      monkeypatch.setattr(safetensors, "safe_open", original)

      with vlm_module._force_qwen4_exp_sanitize_on_load(snapshot):
          with safetensors.safe_open(weight_file) as handle:
              assert handle.metadata() == {"source": "test"}

      assert safetensors.safe_open is original
  ```

  It mirrors the existing test directly above it, so the pair reads as "plain directory" and "HF cache" side by side.

  ### On your try/except note

  Re the exposure you flagged: `resolved = path.resolve()` is inside the `try`, but the `path.parent.resolve()` in the condition is not, so a symlink loop would raise from there. Computing the lexical parent inside the `try` and widening to `except (TypeError, OSError)` closes it without changing the matching logic. Entirely optional — as you say, the original had the same exposure.

  I had arrived at a narrower variant (`path.parent.resolve()` alone) before finding this PR. Yours is the better shape: keeping `resolved.parent` as the first disjunct also matches a weight file that is itself a symlink pointing *into* the model directory from elsewhere, which mine would have missed.

jundot:
Thanks for the clear root-cause analysis and real-model validation. The lexical suffix and resolved-parent check directly fixes the HF cache symlink path. LGTM. I’ll take this change and add focused regression coverage plus a small cleanup in a follow-up.

PR #3227, opened 2026-08-27
fix(scheduler): stop Metal resource-count exhaustion on ArraysCache hybrids (GLM-5.3-Flash)

### LINKED ISSUE #3226: [metal::malloc] Resource limit (499000) exceeded — Metal buffer-count exhaustion on GLM-5.3-Flash (ArraysCache hybrids): missing decode materialization, error-path snapshot leaks, unreachable pool clear, pm-ineligible CacheList(KVCache, PoolingCache)
### Summary

`glm-5.3-flash-6bit` (VLM engine, `glm5_next`) repeatedly fails mid-decode with:

```
RuntimeError: [metal::malloc] Resource limit (499000) exceeded.
```

on a Mac Studio M3 Ultra 512GB (omlx 0.6.3 Homebrew, `iogpu.wired_limit_mb=498074`, memory guard tier `balanced`, paged SSD cache enabled). Long-context chats (~46k prompt, 64k+ context, `max_tokens=262144`) die mid-stream; the engine loop fails all in-flight requests; retries often fail again; occasionally the same request later succeeds.

### This is buffer-count exhaustion, not memory-bytes exhaustion

`499000` is MLX's guard on **Metal's per-process buffer-object count** (`mx.metal.device_info()["resource_limit"]` == 499000 on every Apple GPU; see `num_resources_`/`resource_limit_` in `mlx/backend/metal/allocator.h`). It is unrelated to the wired-limit byte ceiling — the numeric resemblance to a ~487GB limit is coincidence.

Decisive evidence from our logs: after a **full model unload** (`freed=241.04GB ... active_memory: 12.00MB (settled)`) and a clean reload, the very next long-context request still died with the same error ~18 minutes later — with byte usage nowhere near the ceiling. MLX's internal GC is byte-gated (`gc_limit_`), so it never rescues a count exhaustion.

### Contributing defects found by code audit (0.6.3 / current main)

1. **Decode KV/recurrent-state materialization is enabled only for MiniMax, by name.** `scheduler.py`: `default_kv_eval_interval = 256 if "minimax" in model_name_lower else 0`. Every ArraysCache hybrid (GLM-5.3, Qwen3.5-Next, ...) rebinds its GDN/conv state lazily each decode step; without periodic `mx.eval` of the batch cache state, the unmaterialized chain pins buffers per step per layer. GLM-5.3 has 34 ArraysCache layers and gets `0`.

2. **Error paths leak boundary snapshot sets.** `fail_all_requests()` and the chunked-prefill `PrefillMemoryExceededError`/`RuntimeError` handlers never pop `_boundary_cache_snapshots[rid]` nor call `_boundary_snapshot_store.cleanup_request(rid)` (normal completion does, cf. `_cleanup_finished`). Every engine failure permanently strands a per-request snapshot set — live mx.arrays in the in-RAM fallback — creating a positive feedback loop: OOM → leak → less headroom → next OOM sooner. Matches our observed 15:06 → 15:25 failure pair.

3. **The periodic pool clear can never fire on big machines.** `_periodic_clear_threshold_bytes()` = `max(soft_limit/3, 2GiB)` ≈ **146GB** on a 512GB Studio, gated additionally on `step % 512 == 0`. Meanwhile `set_cache_limit(memory_size)` (deliberate, #300) means MLX never self-trims. The pool hoards buffer *handles* even while byte usage is modest.

4. **`CacheList(KVCache, PoolingCache)` falls through every quadratic-cache protection.** `cachelist_pm_member_plan()` returns `None` when any member is `PoolingCache` (`prefix_cache.py`), so GLM-5.3's 11 attention layers take the pre-#2550 legacy cumulative path:
   - store: every block with a boundary snapshot stores the **full cumulative KV** of all 11 layers (`_cachelist_snapshot_sub_tensors`) — Σ i·block_size ≈ 11.5× the context per turn, in Metal copies and SSD bytes; the exact regression #2550 fixed for pm-eligible layouts only;
   - boundary snapshots keep the unblanked KV member (`_extract_snapshot_cache_states` no-ops when plan is None);
   - restore: `reconstruct_cache` loads and force-evals **every block of the chain simultaneously**, then uses only the last block for legacy CacheList layers;
   - `_maybe_backfill_dedup_block_payload` loads (and hot-promotes) full cumulative payloads just to discover no rewrite is needed.

5. Secondary: no decode-time memory guard exists at all (`_check_memory_pressure` is `pass`), and the enforcer's cached `_last_mlx_active_memory_bytes` is only refreshed on prefill/admission paths — frozen for an entire decode.

### Environment

- omlx 0.6.3 (Homebrew), macOS 26.5.1, Mac Studio M3 Ultra 512GB
- model: GLM-5.3-Flash 6-bit MLX (250GB), layout 34× `ArraysCache(size=2)` + 11× `CacheList(KVCache, PoolingCache)`
- paged SSD cache on, block_size auto-enlarged 256→2048, boundary snapshots active, "GDN SSD sidecar is unsupported for layout [...]; falling back to embedded GDN snapshots"

### Fix

PR to follow: enables the decode materialization interval for all ArraysCache hybrids, plugs the error-path snapshot leaks, caps the periodic-clear gate at 16GiB, and stops hot-promoting backfill inspection loads. Item 4's real fix (making `CacheList(KVCache, PoolingCache)` pm-eligible — PoolingCache members stored per-block from boundary snapshots, KV member sliced) needs coordinated changes in `paged_ssd_cache` eligibility mirrors and is proposed as a follow-up.

Pietroski:
Fixes the recurring `[metal::malloc] Resource limit (499000) exceeded` failures on GLM-5.3-Flash and other ArraysCache hybrid models during long-context decode.

Closes #3226

### What the error actually is

`499000` is MLX's guard on **Metal's per-process buffer-object count** (`mx.metal.device_info()["resource_limit"]`), not a byte limit. We reproduced the failure minutes after a full model unload/reload with `active_memory` settled at 12MB — byte usage was nowhere near the wired ceiling. It also reproduces with a 31-token prompt and ~23k generated tokens (pure decode, no long prefill). Details and full audit in #3226.

### Changes

1. **Enable decode KV/recurrent-state materialization for all ArraysCache hybrids** (`omlx/scheduler.py`). `default_kv_eval_interval` was `256 if "minimax" in model_name else 0`. Every ArraysCache hybrid rebinds recurrent state lazily per decode step; without periodic materialization the unmaterialized chain pins one buffer set per step per layer (GLM-5.3: 34 layers), exhausting the 499,000-handle budget mid-generation. The interval now defaults to 256 whenever the model's `make_cache()` layout contains ArraysCache; the `OMLX_DECODE_EVAL_KV_CACHE_INTERVAL` env override is unchanged.

2. **Release boundary snapshots and block refs on error paths** (`omlx/scheduler.py`). `fail_all_requests()` and the chunked-prefill `PrefillMemoryExceededError`/`RuntimeError` handlers now drop `_boundary_cache_snapshots[rid]`, call `boundary_snapshot_store.cleanup_request(rid)`, and release the request's paged-cache blocks — mirroring `_cleanup_finished`. Previously every engine failure permanently stranded a per-request snapshot set (live mx.arrays in the in-RAM fallback), a positive feedback loop toward the next OOM. In-flight async store ids are excluded, same as the rest of `fail_all_requests`.

3. **Cap the periodic pool-clear byte gate at 16GiB** (`omlx/scheduler.py`). `max(soft_limit/3, 2GiB)` evaluated to ~146GB on a 512GB Studio, so the periodic clear effectively never fired while the pool hoarded buffer handles (`set_cache_limit(memory_size)` intentionally never self-trims, #300).

4. **Stop hot-promoting dedup-backfill inspection loads** (`omlx/cache/prefix_cache.py`). `_maybe_backfill_dedup_block_payload` loaded blocks with default hot-cache promotion; for pm-ineligible CacheList layouts (`CacheList(KVCache, PoolingCache)` — GLM-5.3, where `cachelist_pm_member_plan` returns None) those payloads carry the full cumulative KV prefix and were retained for nothing in the common "payload already real" outcome.

### Not in this PR (follow-up proposed in #3226)

The root cause of the cumulative payloads — `cachelist_pm_member_plan()` returning `None` for PoolingCache members, routing GLM-5.3's attention layers onto the pre-#2550 quadratic store/restore path — needs coordinated changes in the `paged_ssd_cache` pm-eligibility mirrors and restore paths, plus a streaming rework of `reconstruct_cache` (which currently force-evals every block of a legacy chain simultaneously and then uses only the last block).

### Testing

- New regression tests: `TestFailAllRequestsSnapshotCleanup` (3), `TestPeriodicClearThresholdCap` (2), `TestHybridDecodeKvEvalDefault` (3) — all passing.
- Affected suites (`test_scheduler.py`, `test_prefix_cache*.py`, `test_scheduler_chunked_prefill.py`) run against the pinned mlx-lm 0.31.3 / mlx-vlm 78b96eb: 438 passed; the 6 failures present are identical on the unmodified tree (environment-dependent parser/deepseek-module tests).


jundot:
Thanks for addressing this. Structural ArraysCache detection correctly extends the existing periodic materialization path to GLM-5.3 and other hybrids, while the allocator and snapshot cleanup changes cover the additional retention paths from #3226. I verified the path on a real GLM model, and the focused scheduler/cache suite passed. LGTM, merging. I’ll add direct regression coverage for the 256-token materialization boundary as an immediate follow-up on main.

PR #3228, opened 2026-08-28
Add Lightning MTP for GLM-5.3-Flash (glm5_next)

davidtorcivia:
Enables Lightning MTP for GLM-5.3-Flash. `glm5_next` was missing from the `_is_mtp_compatible` allowlist, and `patches/mlx_vlm_mtp/` had no runtime for it.

## Head layout

The head sits at `model.language_model.layers.<num_hidden_layers>.*`, not `mtp.*`. Fusion tensors are `eh_proj`, `enorm`, `hnorm`; output norm is `shared_head.norm`.

It ships no `hc_attn_*` or `hc_ffn_*`, so it is a plain pre-norm decoder layer on `(B, S, D)`, not the backbone's `(B, S, hc_mult, D)`.

## Sanitize

Rewrites the nextn tensors onto layer 0, runs the stock pass on them in isolation, rebinds them under `mtp.<i>.*`. They get the same FP8 dequant, expert stacking, indexer fusion and `kv_b_proj` absorption as the backbone.

A head already named `mtp.*` is preserved. Stock sanitize filters on `"mtp." not in k` and would drop it, failing strict `load_weights` on reload.

## Rollback

A verify forward records each linear-attention layer's block inputs, entry state and gate mask in `gdn_sink`. `rollback_speculative_cache` replays the accepted prefix through `gated_delta_update` under that mask sliced to the prefix, then rewinds the recurrent position with `cache.advance(-trim)`. Sparse layers trim. Returning `gdn_states` routes `_call_backbone` onto the mlx-vlm rollback path.

Every layer is checked before any is mutated: `PoolingCache.trim` returns 0 when its undo log cannot cover the tail, so the rollback refuses rather than leaving mixed lengths. `mtp_clamp_accept` runs first and lowers the accept count until the undo fits.

Depth is capped at 7. `PoolingCache.accumulate_windows` stashes an undo log only for updates of 8 rows or fewer; a depth-k chain verifies k+1.

The head cache is committed-only. Its `PoolingCache.offset` counts pooled rows, not tokens, so `_mtp_head_trim_to` cannot rewind it. Speculative head steps run on a per-cycle clone, as in DeepSeek-V4.

Follows Blaizzy/mlx-vlm#2044.

## Runtime patch scoping

`apply_mlx_vlm_mtp_runtime_patch` applied every architecture's runtime on any MTP-capable load. These patches rebind methods on shared parent classes, and `qwen4_exp` subclasses the `qwen3_5` ones, so loading a glm5_next model broke a resident Qwen4-Exp model until restart. The sweep now applies only the loaded `model_type`, matched exactly then by longest declared prefix. Unlisted types keep the previous behaviour.

## Results

M3 Ultra 256 GB, `GLM-5.3-Flash-oQ4e-mtp`, 4233-token prompt, warm prefix cache, 1500 tokens, greedy.

| mtp_enabled | tok/s |
| --- | --- |
| false | 24.9, 24.9 |
| true | 44.1, 47.6, 47.6 |

Acceptance settles near 95% at 3.58 tokens per cycle; a one-line prompt gives about 67% and 1.2x. The head-cache change landed after these runs and is not re-measured.

Output is not bit-identical to non-MTP decoding at 4-bit: block verify at S>1 dispatches different kernels than singleton decode. Every emitted token is still verified by the target.

## Tests

`tests/test_glm5_next_mtp.py` covers key matching, block structure, the cache pair, both sanitize paths, the rollback (cache family, gate mask, position rewind, both refusals), the accept clamp, the depth cap, the head clone and the scoped sweep. `tests/test_model_loading.py` covers the model_type reaching the sweep. No weights load.

Full suite: 10487 passed, 111 skipped.

### DIFF
--- a/omlx/patches/mlx_vlm_mtp/__init__.py
+++ b/omlx/patches/mlx_vlm_mtp/__init__.py
@@ -95,6 +95,39 @@ _RUNTIME_PATCHES_BY_TYPE: dict[str, tuple[str, ...]] = {
 }
 
 
+def _runtime_patches_for(model_type: str | None) -> tuple[str, ...] | None:
+    """Sub-patches owning ``model_type``, or None to keep the historical sweep.
+
+    Exact match first, then the longest declared prefix: family variants
+    (``qwen3_5_moe_*`` and friends, which ``_is_mtp_compatible`` admits by
+    prefix) subclass the same shared parents as the family they extend, so
+    the family's entry owns them. Without the prefix step such a variant
+    falls through to the apply-everything sweep and rebinds architectures
+    this load has nothing to do with, which is the failure the table exists
+    to prevent.
+    """
+    if not model_type:
+        return None
+    if model_type is not None and model_type not in _RUNTIME_PATCHES_BY_TYPE:
+        # _is_mtp_compatible admits model_type prefixes (qwen3_5*, qwen3_6*),
+        # so a variant can reach this with no table entry and fall back to the
+        # unscoped sweep. Say so rather than doing it silently.
+        logger.debug(
+            "mlx-vlm MTP runtime patch: no entry for model_type=%s, applying "
+            "every architecture's runtime",
+            model_type,
+        )
+
+    if model_type in _RUNTIME_PATCHES_BY_TYPE:
+        return _RUNTIME_PATCHES_BY_TYPE[model_type]
+    family = max(
+        (k for k in _RUNTIME_PATCHES_BY_TYPE if model_type.startswith(k)),
+        key=len,
+        default=None,
+    )
+    return _RUNTIME_PATCHES_BY_TYPE[family] if family is not None else None
+
+
 def apply_mlx_vlm_mtp_runtime_patch(model_type: str | None = None) -> bool:
     """Apply the mlx-vlm runtime MTP patches (attach MTPModule, mtp_forward).
 
@@ -122,9 +155,10 @@ def apply_mlx_vlm_mtp_runtime_patch(model_type: str | None = None) -> bool:
     """
     import importlib
 
-    if model_type in _RUNTIME_PATCHES_BY_TYPE:
+    scoped = _runtime_patches_for(model_type)
+    if scoped is not None:
         applied = False
-        for name in _RUNTIME_PATCHES_BY_TYPE[model_type]:
+        for name in scoped:
             if importlib.import_module(f".{name}", __name__).apply():
                 applied = True
             else:
--- a/omlx/patches/mlx_vlm_mtp/__init__.py
+++ b/omlx/patches/mlx_vlm_mtp/__init__.py
@@ -85,11 +85,10 @@ def apply_mlx_vlm_mtp_runtime_patch() -> bool:
     the runtime infrastructure so VLMBatchedEngine can actually invoke
     the MTP head at inference time.
 
-    Covers Qwen3.5-MoE (qwen3_5_moe), dense Qwen3.5/3.6 (qwen3_5) and
-    Gemma 4 merged-assistant (gemma4 and gemma4_unified) VLM families. Each
-    sub-patch tracks
-    its own ``_APPLIED`` flag, so calling repeatedly is cheap once all
-    have settled. Returns True if at least one sub-patch applied
+    Covers Qwen3.5-MoE (qwen3_5_moe), dense Qwen3.5/3.6 (qwen3_5), Gemma 4
+    merged-assistant (gemma4 and gemma4_unified) and GLM-5.3-Flash
+    (glm5_next) VLM families. Each sub-patch tracks its own ``_APPLIED``
+    flag, so calling repeatedly is cheap once all have settled. Returns True if at least one sub-patch applied
     successfully — a given model only needs whichever matches its
     model_type.
 
@@ -98,6 +97,7 @@ def apply_mlx_vlm_mtp_runtime_patch() -> bool:
     """
     from . import (
         gemma4_vlm_runtime,
+        glm5_next_vlm_runtime,
         inkling_vlm_runtime,
         qwen35_moe_vlm_runtime,
         qwen35_vlm_runtime,
@@ -116,4 +116,8 @@ def apply_mlx_vlm_mtp_runtime_patch() -> bool:
     if not inkling_ok:
         logger.debug("Inkling VLM runtime MTP patch did not apply")
 
-    return moe_ok or dense_ok or gemma4_ok or inkling_ok
+    glm5_ok = glm5_next_vlm_runtime.apply()
+    if not glm5_ok:
+        logger.debug("GLM-5.3 (glm5_next) VLM runtime MTP patch did not apply")
+
+    return moe_ok or dense_ok or gemma4_ok or inkling_ok or glm5_ok
--- a/omlx/patches/mlx_vlm_mtp/__init__.py
+++ b/omlx/patches/mlx_vlm_mtp/__init__.py
@@ -77,7 +77,25 @@ def apply_mlx_vlm_mtp_patch() -> bool:
     return True
 
 
-def apply_mlx_vlm_mtp_runtime_patch() -> bool:
+# Which runtime sub-patch belongs to which ``model_type``. These patches
+# rebind methods on shared parent classes and other architectures subclass
+# them, so applying all of them reaches models this load has nothing to do
+# with. An empty tuple means the architecture manages its own MTP and must
+# never receive the sweep. A model_type that is not listed keeps the
+# historical apply-everything behaviour.
+_RUNTIME_PATCHES_BY_TYPE: dict[str, tuple[str, ...]] = {
+    "qwen3_5": ("qwen35_vlm_runtime",),
+    "qwen3_5_moe": ("qwen35_moe_vlm_runtime", "qwen35_vlm_runtime"),
+    "gemma4": ("gemma4_vlm_runtime",),
+    "gemma4_unified": ("gemma4_vlm_runtime",),
+    "inkling": ("inkling_vlm_runtime",),
+    "inkling_mm_model": ("inkling_vlm_runtime",),
+    "glm5_next": ("glm5_next_vlm_runtime",),
+    "qwen4_exp": (),
+}
+
+
+def apply_mlx_vlm_mtp_runtime_patch(model_type: str | None = None) -> bool:
     """Apply the mlx-vlm runtime MTP patches (attach MTPModule, mtp_forward).
 
     Distinct from ``apply_mlx_vlm_mtp_patch``: that one only patches
@@ -94,7 +112,28 @@ def apply_mlx_vlm_mtp_runtime_patch() -> bool:
 
     Should be called *before* ``mlx_vlm.utils.load(...)`` so the
     instantiated LanguageModel picks up the patched ``__init__``.
+
+    ``model_type`` scopes the sweep to the architecture being loaded.
+    qwen4_exp's LanguageModel, gated delta net and attention all subclass
+    qwen3_5's, so applying the qwen3_5 runtime while a qwen4_exp model is
+    resident rewrites that live model's trunk ``__call__`` and its next
+    request fails with "Qwen4 Lightning MTP expects hidden shape
+    [batch, tokens, hc_count * hidden_size]" until the server restarts.
     """
+    import importlib
+
+    if model_type in _RUNTIME_PATCHES_BY_TYPE:
+        applied = False
+        for name in _RUNTIME_PATCHES_BY_TYPE[model_type]:
+            if importlib.import_module(f".{name}", __name__).apply():
+                applied = True
+            else:
+                logger.debug("%s MTP runtime patch did not apply", name)
+        logger.debug(
+            "mlx-vlm MTP runtime patch scoped to model_type=%s", model_type
+        )
+        return applied
+
     from . import (
         gemma4_vlm_runtime,
         glm5_next_vlm_runtime,
--- a/omlx/patches/mlx_vlm_mtp/__init__.py
+++ b/omlx/patches/mlx_vlm_mtp/__init__.py
@@ -18,6 +18,7 @@ building the VLM sanitizer. The patches are idempotent.
 
 from __future__ import annotations
 
+import importlib
 import logging
 
 logger = logging.getLogger(__name__)
@@ -57,6 +58,7 @@ def set_mtp_attach_enabled(enabled: bool) -> None:
 def is_mtp_attach_enabled() -> bool:
     return _MTP_ATTACH_ENABLED
 
+
 def apply_mlx_vlm_mtp_patch() -> bool:
     """Apply the mlx-vlm MTP sanitize monkey-patches.
 
@@ -77,6 +79,16 @@ def apply_mlx_vlm_mtp_patch() -> bool:
     return True
 
 
+# The unscoped sweep, in the order it has always run. Every entry in
+# ``_RUNTIME_PATCHES_BY_TYPE`` below is drawn from this tuple.
+_RUNTIME_PATCHES: tuple[str, ...] = (
+    "qwen35_moe_vlm_runtime",
+    "qwen35_vlm_runtime",
+    "gemma4_vlm_runtime",
+    "inkling_vlm_runtime",
+    "glm5_next_vlm_runtime",
+)
+
 # Which runtime sub-patch belongs to which ``model_type``. These patches
 # rebind methods on shared parent classes and other architectures subclass
 # them, so applying all of them reaches models this load has nothing to do
@@ -95,7 +107,11 @@ _RUNTIME_PATCHES_BY_TYPE: dict[str, tuple[str, ...]] = {
 }
 
 
-def _runtime_patches_for(model_type: str | None) -> tuple[str, ...] | None:
+def _patches_for(
+    model_type: str | None,
+    table: dict[str, tuple[str, ...]],
+    kind: str,
+) -> tuple[str, ...] | None:
     """Sub-patches owning ``model_type``, or None to keep the historical sweep.
 
     Exact match first, then the longest declared prefix: family variants
@@ -108,24 +124,23 @@ def _runtime_patches_for(model_type: str | None) -> tuple[str, ...] | None:
     """
     if not model_type:
         return None
-    if model_type is not None and model_type not in _RUNTIME_PATCHES_BY_TYPE:
-        # _is_mtp_compatible admits model_type prefixes (qwen3_5*, qwen3_6*),
-        # so a variant can reach this with no table entry and fall back to the
-        # unscoped sweep. Say so rather than doing it silently.
-        logger.debug(
-            "mlx-vlm MTP runtime patch: no entry for model_type=%s, applying "
-            "every architecture's runtime",
-            model_type,
-        )
-
-    if model_type in _RUNTIME_PATCHES_BY_TYPE:
-        return _RUNTIME_PATCHES_BY_TYPE[model_type]
+    if model_type in table:
+        return table[model_type]
     family = max(
-        (k for k in _RUNTIME_PATCHES_BY_TYPE if model_type.startswith(k)),
-        key=len,
-        default=None,
+        (k for k in table if model_type.startswith(k)), key=len, default=None
     )
-    return _RUNTIME_PATCHES_BY_TYPE[family] if family is not None else None
+    if family is not None:
+        return table[family]
+    # _is_mtp_compatible admits model_type prefixes (qwen3_5*, qwen3_6*), so a
+    # variant can reach this with no entry and no declared family, and fall
+    # back to the unscoped sweep. Say so rather than doing it silently.
+    logger.debug(
+        "mlx-vlm MTP %s patch: no entry for model_type=%s, applying every "
+        "architecture's patch",
+        kind,
+        model_type,
+    )
+    return None
 
 
 def apply_mlx_vlm_mtp_runtime_patch(model_type: str | None = None) -> bool:
@@ -153,44 +168,17 @@ def apply_mlx_vlm_mtp_runtime_patch(model_type: str | None = None) -> bool:
     request fails with "Qwen4 Lightning MTP expects hidden shape
     [batch, tokens, hc_count * hidden_size]" until the server restarts.
     """
-    import importlib
-
-    scoped = _runtime_patches_for(model_type)
+    scoped = _patches_for(model_type, _RUNTIME_PATCHES_BY_TYPE, "runtime")
     if scoped is not None:
-        applied = False
-        for name in scoped:
-            if importlib.import_module(f".{name}", __name__).apply():
-                applied = True
-            else:
-                logger.debug("%s MTP runtime patch did not apply", name)
         logger.debug(
             "mlx-vlm MTP runtime patch scoped to model_type=%s", model_type
         )
-        return applied
 
-    from . import (
-        gemma4_vlm_runtime,
-        glm5_next_vlm_runtime,
-        inkling_vlm_runtime,
-        qwen35_moe_vlm_runtime,
-        qwen35_vlm_runtime,
-    )
+    applied = False
+    for name in _RUNTIME_PATCHES if scoped is None else scoped:
+        if importlib.import_module(f".{name}", __name__).apply():
+            applied = True
+        else:
+            logger.debug("%s MTP runtime patch did not apply", name)
 
-    moe_ok = qwen35_moe_vlm_runtime.apply()
-    if not moe_ok:
-        logger.debug("Qwen3.5-MoE VLM runtime MTP patch did not apply")
-    dense_ok = qwen35_vlm_runtime.apply()
-    if not dense_ok:
-        logger.debug("Qwen3.5 (dense) VLM runtime MTP patch did not apply")
-    gemma4_ok = gemma4_vlm_runtime.apply()
-    if not gemma4_ok:
-        logger.debug("Gemma 4 VLM runtime MTP patch did not apply")
-    inkling_ok = inkling_vlm_runtime.apply()
-    if not inkling_ok:
-        logger.debug("Inkling VLM runtime MTP patch did not apply")
-
-    glm5_ok = glm5_next_vlm_runtime.apply()
-    if not glm5_ok:
-        logger.debug("GLM-5.3 (glm5_next) VLM runtime MTP patch did not apply")
-
-    return moe_ok or dense_ok or gemma4_ok or inkling_ok or glm5_ok
+    return applied
--- a/omlx/patches/mlx_vlm_mtp/glm5_next_vlm_runtime.py
+++ b/omlx/patches/mlx_vlm_mtp/glm5_next_vlm_runtime.py
@@ -1,0 +1,787 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Runtime MTP head attachment for the mlx-vlm GLM-5.3-Flash (glm5_next) path.
+
+Companion to ``omlx/patches/mlx_lm_mtp`` for an mlx-vlm-hosted architecture,
+in the same shape as ``qwen35_moe_vlm_runtime`` and ``gemma4_vlm_runtime``.
+
+``zai-org/GLM-5.3-Flash`` declares ``num_nextn_predict_layers: 1`` and stores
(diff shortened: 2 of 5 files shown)


davidtorcivia:
Pushed a correctness fix to the rollback path after finding it corrupted long generations.

Two bugs, both in `rollback_speculative_cache`:

`isinstance(c, ArraysCache)` missed `SizedArraysCache`. glm5_next's linear-attention layers hold either class depending on the batch path. A `SizedArraysCache` fell through to the trim branch, where it is not trimmable, so its recurrent state kept the rejected tokens while the KV caches rewound. `scheduler.py` already groups the two names together for this reason; the rollback now matches on the family.

The recurrent cache offset was never rewound. The verify forward calls `cache.advance(S)` for the whole block. Restoring `c[0]`/`c[1]` does not undo that, and unlike the sparse caches there is no `trim()` on that path to do it.

Either one alone leaves the recurrent layers ahead of the KV layers by the rejected count, and the drift compounds every round. At 78 to 90 percent acceptance over a thousand-plus cycles, a long generation degrades into interleaved fragments and then a single repeating token. Short generations finish before it accumulates, which is why it only showed up on multi-thousand-token runs.

Reproduced with a two-turn tool-calling exchange at 2395 prompt tokens: 6 of 8 runs degenerate before, 0 of 4 after. `nemotron_h` on the mlx_lm MTP path was unaffected throughout, which is what localised it to this runtime.

`tests/test_glm5_next_mtp.py` gains a case that drives the rollback with a cache named `SizedArraysCache` and asserts both that it takes the recurrent path and that the offset rewinds by the rejected count.

davidtorcivia:
One more finding worth flagging on this PR, since it enables Lightning MTP for a hybrid recurrent architecture.

With MTP on, boundary snapshots for the non-sliceable recurrent caches are not captured: `Skipping cache store for <id>: no boundary-aligned snapshot for non-sliceable cache state (all captures skipped, ...)`, 323 times over a serving session. With MTP off on the same model and prompts, zero such events and stores succeed (`Stored cache for <id>: 2 blocks (2 saved to tiered cache), 4096 tokens`).

The read side still takes hits and reconstructs all 45 layers (`Cache hit ... 1 blocks, 2048 tokens` then `Reconstructed cache from tiered cache: 45 layers`). glm5_next has 34 recurrent KDA layers and 11 sparse attention layers, so a hit restores the sparse KV correctly while the recurrent state is missing. Generation then runs against a context the model never processed. Observed as replies in the wrong language and invented material on a turn following a cached prefix.

The working hypothesis is that MTP emits 2 to 4 tokens per cycle (measured tok/cycle 1.7 to 3.6) and steps over the 2048-token block boundary rather than landing on it, so the capture condition never holds. I have not yet confirmed whether this is specific to this runtime or generic to any hybrid recurrent model under MTP, and I would rather say that than guess. `prefix_cache.py` already knows these caches need full cumulative state (`_PM_SAFE_NON_SLICEABLE_SUBS`), so the gap looks like the capture trigger rather than the storage design.

This is separate from the rollback fix above, which addressed corruption within a single generation. This one crosses turns through the cache. Happy to open it as its own issue with the repro if that is more useful than carrying it here.

gilby:
Tested this PR on real weights and hardware: `zai-org/GLM-5.3-Flash` quantized with oq (enhanced, ~170 GB) on an M3 Ultra / 512 GB. It works, and the runtime-only structure (no vendored-model edits) is the right shape — we had independently built the same feature with vendored edits and yours is cleaner. Measured at default depth 3 against MTP-off: +13–20% decode on code, +47% on math, 86–89% draft acceptance (3.0–3.2 tok/cycle).

While validating we found five issues. Fixes for all of them, plus tests, are stacked directly on this PR's head at [`gilby/omlx:glm53-mtp-improved`](https://github.com/gilby/omlx/tree/glm53-mtp-improved) — free to cherry-pick.

**1. The recurrent-cache offset rewind is dead code** ([e282339c](https://github.com/gilby/omlx/commit/e282339c))

Neither `ArraysCache` nor `SizedArraysCache` defines `.offset`, and nothing in omlx or mlx-lm sets one on a recurrent cache, so `getattr(c, "offset", 0) >= trim` never passes when `trim > 0`. The suite passes because `test_rollback_matches_the_recurrent_cache_family` hand-sets `c.offset = 4` — an attribute production caches never carry. The real over-advance is `cache.advance(S)` in the patched linear attention (`ArraysCache.advance(N)` does `lengths -= N; left_padding -= N`), so the rewind is `advance(-trim)`. The offset path is kept for caches that do carry one, made mutually exclusive so nothing rewinds twice. Added a test on a production-shaped `ArraysCache` with no offset attribute.

**2. The gate mask doesn't survive into the replay** (same commit)

The capture tuple drops the resolved SSM bool mask, so `rollback_speculative_cache` replays `gated_delta_update` without the mask the verify forward used — whenever a mask is active, the replayed recurrent state is not the state a genuine n-token forward would produce. Fixed by stashing the resolved mask in the capture tuple (12th element; the 11-tuple form is still tolerated) and slicing it to the accepted prefix (`mask[:, :n]`). Test added asserting the sliced mask reaches the kernel.

**3. No depth cap** ([ab6de0d8](https://github.com/gilby/omlx/commit/ab6de0d8))

`set_mtp_depth` clamps to 1..8, and `PoolingCache`'s cross-boundary undo log covers at most 8 rows per update. A depth-8 chain verifies k+1 = 9 rows, so the DSA indexer pool cannot rewind a full rejection — silent corruption reachable via the admin `mtp_num_draft_tokens=8`. Capped `_omlx_mtp_depth` at 7.

**4. Missing `mtp_clamp_accept`** (same commit)

The generator resolves it on `gen_batch.model` — the `VLMModelAdapter` — at `batch_generator.py:3020`. Without it, a sparse cache that cannot undo the rejected tail goes unhandled. Ported an implementation (lower the accept count while any sparse cache can't trim/undo the needed rows; `PoolingCache.remainder` is per-sequence, so the conservative bound is the minimum) plus a hasattr-guarded adapter proxy matching the PR's other adapter attributes.

**5. Whitelisting `glm5_next` breaks already-resident `qwen4_exp` models** ([0d4f0809](https://github.com/gilby/omlx/commit/0d4f0809))

Not introduced by this PR, but this PR is the first thing that triggers it. `apply_mlx_vlm_mtp_runtime_patch()` applies **every** architecture's runtime patch on any MTP-capable load, and those patches rebind methods on shared parent classes (`Qwen3_5Model.__call__`, `Qwen3_5LanguageModel.__init__/__call__`) which `qwen4_exp` subclasses (`class LanguageModel(Qwen3_5LanguageModel)`).

Repro: load a qwen4_exp model (Qwen3.8-Flash-Next) → load a glm5_next model with this PR applied → send the next request to the Qwen model → 500 with `Qwen4 Lightning MTP expects hidden shape [batch, tokens, hc_count * hidden_size]`. A restart clears it, which is why it hides from post-deploy smoke tests. Bisected to be sure: disabling every glm5_next patch does **not** fix it — the damage is the sweep, not this PR's code.

Fix: `apply_mlx_vlm_mtp_runtime_patch(model_type)` applies only the loaded architecture's runtime; unknown model_type keeps the historical apply-everything behaviour, and self-managed archs (`qwen4_exp`) get an explicit empty entry so the fallback can never reach them.

With all five applied: 12/12 tests pass, the qwen4_exp → glm5_next → qwen4_exp sequence serves clean, and acceptance/perf are unchanged (86.9% at depth 3 on our checkpoint).

One suggestion beyond the branch: `apply_mlx_vlm_mtp_patch` (the sanitize sweep) has the same unscoped apply-everything structure. Nothing breaks today because `qwen4_exp` overrides `sanitize`, but scoping it by model_type the same way would close out this class of bug.

_Produced with an AI agent._

gilby:
## Separate `glm5_next` defect: GLM-5.3-Flash stops seeing its own prompt past ~1–2K context

Filing this here because it's the active `glm5_next` thread, **not** because it's caused by this PR — the evidence below rules MTP out explicitly. Posting our findings plus a full remediation plan in case it's useful to you or anyone else.

### Symptom

Serving `GLM-5.3-Flash` under oMLX (oQ4e, `model_type: glm5_next`), turn 0 of any conversation is excellent. Once context grows past roughly 1–2K tokens the model **confabulates its input** — it answers a question nobody asked, coherently.

History-length sweep, same question, only the assistant-history length varied, non-repetitive filler:

| prompt_tok | 156 | 518 | 1376 | 2877 | 5769 |
|---|---|---|---|---|---|
| on-topic | ✓ | ✓ | **✗** | ✓ | **✗** |

Non-monotonic and stochastic. Observed failure modes across a 6-topic × 4-turn conversation set, all on **English-only** input:

| mode | example |
|---|---|
| **coherent answer to a DIFFERENT question** (most diagnostic — reading another prompt, not degrading) | a pandas/matplotlib script for a PHP middleware question; famous novel openings for a contract-drafting question |
| **multilingual emission** | a Chinese power-series derivation; Persian; Korean |
| **numeric repetition** | `65 - - 66 65 - - 67` × 27K chars |
| **meta-confusion** | "Wait, I need the actual question. You didn't include one." ×4; "you've sent 'CHS' multiple times" |
| **empty at cap** | full max_tokens consumed, empty content |
| **stochastic recovery** | later turns on *longer* context returned to correct output after a corrupt turn |

That last row matters: corruption is **not** monotonic state poisoning. A later, longer turn can be clean.

### What it is not

- **Not MTP / not this PR.** An MTP-on vs MTP-off A/B collapses identically, and a pre-MTP quant of the same weights (no `mtp.*` tensors at all) fails at the *same lengths with the same pattern*.
- **Not the weights or quant.** Two independently produced quants behave identically; other stacks serving this model family report no such corruption.
- **Not the chat template.** Renders correctly under a marker test (all roles present, correct order).
- **Not solely prefix-cache reuse.** One reproduction ran against an effectively empty prefix cache (block hashes include `model_name`; that quant had a fresh name; shared prefixes in the sweep were < one 2048-token block). So **at least one corrupting mechanism is intrinsic to the serving path.** Reuse is *also* implicated separately — `re-prefills … (reused 2048/4096/32768)` lines appear in production traffic on this model.

First request after a model load is always clean; corruption appears on warm engines.

### Architecture facts (verified, for anyone picking this up)

45 layers = 34 KDA linear-attention (`ArraysCache(size=2)` = `[conv_state, gdn_state]`; kernel `gated_delta_update`; **no `trim`, no `offset`** — position bookkeeping only via `advance()` on `lengths`/`left_padding`) + 11 DSA sparse-attention (`CacheList(KVCache, PoolingCache(kpool=4))`, `index_topk=2048`, dense ≤2048 total tokens, top-k beyond).

`chunked_prefill=false` still chunks in practice: decode-fairness `force_chunk` (`scheduler.py:10140`), memory throttle `_adaptive_chunk_size` (`scheduler.py:4154`), contended cap (`scheduler.py:4787`). Paged prefix cache enabled with `block_size` auto-enlarged 256 → 2048 for ArraysCache hybrids; boundary snapshots at 2048-token boundaries.

### Hypothesis ledger

| # | mechanism | for | against / status |
|---|---|---|---|
| **H1** | **Run-time-variable prefill chunking** interacting with a chunk-count-sensitive bug. Cuts are stochastic (`chunk 2048 -> 512`×16, `-> 32`, `-> 128` in live logs) — the only mechanism making the *same prompt* take *different code paths* per request | matches stochasticity + cold-request immunity (cold engine = single chunk); intrinsic, needs no cache | per-chunk arithmetic (conv carry, gdn state, pool remainder, masks) reads correct — needs a defect to ride on it |
| **H2** | **`mx.depends` barrier skipped on sparse fast paths** (`language.py:612-671`): the PoolingCache↔KVCache in-place-write ordering edge is applied on the dense / `L==1` / fallback paths but **skipped** on the `_gathered_attention` (L≤8), `sparse_mla_attention` (Kv≥4096) and `exact_block_token_attention` returns — exactly the long-prompt paths. Lazy-eval read/write race ⇒ stochastic | engages only on long/sparse requests | dense path has the barrier, so can't alone explain a 1376-token failure |
| **H3** | **Leaked `lengths`/`left_padding` on the KDA `ArraysCache`** flips `has_right_padding` (`language.py:217`) → `mixed = where(mask,…,0)` **zeroes prompt tokens**; `advance(S)` decrements `lengths`, so a stale value zeroes progressively *more* of longer prompts | reproduces the symptom exactly (prompt blindness, worse with length); stock `BatchGenerator.prepare()` sets these, and `SizedArraysCache.__getattr__` forwards them | no confirmed setter on the glm5_next path found yet — needs a runtime assert to rule in/out |
| **H4** | **Prefix-cache partial-hit restores KDA state without position validation**: `prefix_cache.py:4161-4186` does a placeholder-*shape* test only, and the skew guard (`scheduler.py:6497-6551`) anchors on `_first_leaf_cache_offset`, which is `None` for `ArraysCache` → the 34 KDA lanes are **never** position-checked | mechanism is real and unguarded; reuse fires in production | cannot explain the empty-cache reproduction ⇒ second-order hazard, not the primary bug |
| **D2** | **`CacheList.trim` returns the LAST member's count** (mlx_lm `cache.py:820`). `PoolingCache.trim` returns `0` when it can't undo while the `KVCache` already trimmed `n` ⇒ permanent KV↔indexer desync. Called unchecked by the MTP rollback path | real bug — **this one does touch this PR** | not the primary defect: the no-MTP arm fails identically |
| **D3** | **`PoolingCache.accumulate_windows` sub-ratio chunk** (`cache_extras.py:184`): `usable==0` with a carried remainder and `L∈{2,3}` ⇒ mismatched-width assignment | real latent bug | would raise, not confabulate |

**Ruled out by inspection:** hyper-connection stream (position-local kernel, length-independent), indexer offset arithmetic (checked line-by-line; the dense→sparse boundary is exactly 2048, inconsistent with 1376 ✗ / 2877 ✓), TurboQuant (force-disabled for this arch), sliding/rotating windows (none), and `gated_delta` chunk-length handling (`T` is a runtime input).

### Plan — Phase 0: discriminate before fixing

The exit criterion is deliberately strict: **one hypothesis confirmed by direct evidence** (an assert firing, an arm delta, or a bit-exactness diff) — not by elimination.

- **0a.** Repro at N=8 samples per length, adding 1024 / 1600 / 2048 / 2304 / 4096 (straddling the block and top-k boundaries).
- **0b.** Instrument (logging only, no behaviour change): log if `cache.lengths`/`left_padding` are non-`None` at `Glm5NextLinearAttention.__call__` (settles H3 in one sweep); correlate the scheduler's `chunk X -> Y` cuts with the corrupt verdict (tests H1); count which indexer path each layer-call takes (tests H2 reachability per length).
- **0c.** Discriminating matrix, MTP disabled throughout: **A** as-is (baseline rate per length) · **B** prefix cache disabled globally (removes H4) · **C** idle engine forced single-chunk (if single-chunk is always clean, chunking is load-bearing → H1) · **D** force the sparse fallback so the masked reference path *and* the `mx.depends` barrier run (tests H2 directly).
- **0d.** **Offline bit-exactness harness** (no server, decisive for H1): prefill the same ~6K prompt (i) single-shot and (ii) chunked 512+512+…, compare final logits and every cache state — `allclose`, then bitwise. If chunked ≠ single-shot, the intrinsic bug is captured deterministically and can be bisected per layer.

### Plan — Phase 1: fixes, ordered

1. **H3 guard (cheapest, ship first).** Treat stale `lengths`/`left_padding` at batch-1 as a bug: clear them with a WARNING rather than silently zeroing prompt tokens.
2. **H2.** Hoist the `mx.depends` pool↔KV barrier so it executes on **all** return paths of the sparse attention (compute deps before the fast-path dispatch).
3. **D2.** Give the `CacheList` a correct `trim` (min over members, or all-or-nothing with rollback) and check the return in the MTP rollback path. Worth fixing upstream in mlx_lm too — returning the last member's count is wrong for *any* heterogeneous `CacheList`.
4. **D3.** Handle `usable==0` with a carried remainder in `accumulate_windows`.
5. **H4 hardening**, by ascending blast radius: (a) a per-model prefix-cache kill switch, so reuse can be disabled for this arch alone; (b) carry the token count as real metadata through `ArraysCache` store/restore (`_get_meta_state_keys()` currently returns `()`) and turn the placeholder-shape test into a **position** test; (c) extend the boundary-snapshot skew guard to validate KDA lanes instead of `continue`-ing past them.
6. **MTP-poison containment.** When `rollback_speculative_cache` bails (returns 0 on missing gdn state), mark the request's cache dirty and skip boundary-snapshot capture until the next full prefill — otherwise a desynced state can be stored as a legitimate-looking block and poison every later warm hit on that prefix.

### Validation gate

1. Offline bit-exactness harness green (chunked ≡ single-shot across all cache states).
2. Post-fix repro: **0/8 corrupt at every length**, including 1376 and 5769, plus a deep-context spot check.
3. Warm-engine adversarial sequence — multi-turn conversations with 2048-multiple shared prefixes to force partial hits; every turn on-topic, zero cache-rejection regressions.
4. MTP regression gate: acceptance held at ≥80% (our baseline is 86.9% / 3.00 tokens-per-cycle at depth 3).
5. Cross-model smoke: `scheduler.py` / `prefix_cache.py` are shared code — one generation per resident model family before calling it done.

### Status — not being executed by us

**We are not working on this.** Our hardware is committed to other work, and we've moved this model to a different serving stack in the meantime, so the plan above is unlikely to be executed by us any time soon. It's posted in full so that **anyone who wants to pick it up can** — no coordination needed, and no expectation of attribution.

Two caveats on the plan itself: the hypothesis ledger is *ranked, not resolved* — Phase 0 exists precisely because we have not yet confirmed a root cause, and the evidence already indicates **at least two independent problems** (one intrinsic to the serving path, one in prefix-cache reuse), so fixing only the confirmed one may leave residual stochastic failures. Every file:line reference was verified against the tree at the time of writing.

*This investigation and plan were produced by **Fable** (Anthropic's Claude Fable 5) working in Claude Code.*

davidtorcivia:
Four commits pushed.

- Recurrent position rewind. Neither `ArraysCache` nor `SizedArraysCache` defines `offset`. The position is `lengths`/`left_padding`, rewound with `advance(-trim)`.
- Gate mask resolved once, stashed in the capture tuple, sliced to the accepted prefix on replay. The 11-element form still loads.
- Rollback checks every layer before mutating any, and refuses otherwise. `PoolingCache.trim` returns 0 when its undo log cannot cover the tail. The same check covers a short `gdn_sink`.
- `mtp_clamp_accept` added, following `mlx_lm_mtp/deepseek_v4_model.py`.
- Chain depth capped at 7. `PoolingCache.accumulate_windows` stashes an undo log only for updates of 8 rows or fewer; a depth-k chain verifies k+1.
- Head cache committed-only. `PoolingCache.offset` counts pooled rows, not tokens, so `_mtp_head_trim_to` cannot rewind it. `_omlx_mtp_head_clone` is now True.
- Runtime sweep applies only the loaded `model_type`, matched exactly then by longest declared prefix. `inkling_mm_model` added, `qwen4_exp` maps to nothing.

Scoping for `apply_mlx_vlm_mtp_patch` ships with the oQe sanitize change, which is where glm5_next's sanitize hook is installed.

The head-cache change postdates the tok/s and acceptance numbers in the description and is not re-measured.

Tests: 24 in `tests/test_glm5_next_mtp.py`, plus the pre-load dispatch cases in `tests/test_model_loading.py`. Full suite 10487 passed, 111 skipped.

davidtorcivia:
@gilby Independent reproduction on an M3 Ultra 256 GB, oQ4e, same symptom. Some of it cuts against H1.

Ladder, fresh nonce per request so nothing can hit the prefix cache, greedy, three repeats per size:

| prompt_tok | 815-819 | 1465-1469 | 2116-2119 | 2900-2904 |
| --- | --- | --- | --- | --- |
| on-topic | 3/3 | 0/3 | 0/3 | 0/3 |

Flaky at the low end: a 817-token prompt of the same shape failed minutes before the three 815 to 819 runs passed.

Against H1: the fragmented-chunk run is the one that worked. A correct 14,189-token answer was prefilled at 2048 cut to 384, then 480, 672, 960; the failures ran uniform 1024s, and the most recent ones ran with no throttling at all.

Also does not hold here: first request after a model load is not always clean. In one process the first request after load, 14,189 tokens on a cache miss, was corrupt.

Two more eliminations:

- Still fails with `~/.omlx/cache` moved aside, so no stored block can be read.
- `gpt-oss-20b` answers the same 3989-token prompt correctly while resident in the same server process, so the scheduler and the machine are not the axis.

One more symptom shape: a sentence planted in a 4k-token prompt is invisible whether it sits at the start or the end, so it is not "the tail survives".

Same conclusion as yours on MTP: fails with `mtp_enabled: false`, and fails identically with this PR's pre-fix runtime restored.

On D2, the commits above check what `trim` returns and refuse the whole rollback before mutating any layer, rather than trimming some and reporting success.

gilby:
Thanks — and your ladder is better evidence than mine. Taking the eliminations first, because three of them land on my hypotheses.

**H1 (run-time-variable chunking) is dead.** Your fragmented run — 2048 cut to 384/480/672/960 — is the one that *worked*, while the uniform-1024 and unthrottled runs failed. That's the opposite of what H1 predicts, and it was my top-ranked mechanism. It was always the weakest kind of hypothesis: it explained the *stochasticity* without naming a defect, and I ranked it first largely because it was the only thing I could see that made the same prompt take different paths. Your data removes it.

**H4 (prefix-cache partial-hit restore) is dead as the primary.** Fails with the cache directory moved aside, and your fresh-nonce ladder can't hit a block anyway. The unguarded recurrent restore I described is still real — the placeholder-*shape* test at `prefix_cache.py:4161` is not a *position* test, and the skew guard anchors on `_first_leaf_cache_offset`, which is `None` for `ArraysCache` — but it's a latent hazard, not this bug.

**My "first request after a load is always clean" was wrong**, and it was load-bearing for the cold/warm framing. Your corrupt first-request-after-load on a cache miss kills it. I'd only ever observed it, never tested it directly.

**One correction on my own numbers:** my sweep showed 2877 ✓ between two failures, which is what made the curve look non-monotonic. That was a **single sample per length**. Your 3/3-per-size design is the right one, and against it my 2877 pass is most likely luck rather than a real island of correctness. Your ladder — 3/3 at 815–819, then 0/3 from 1465 up — reads as a **threshold**, not a stochastic scatter. I'd trust yours over mine.

### What survives, and why I think it's H3

The symptom that constrains hardest is your planted-sentence result: a sentence in a 4k prompt is invisible **whether it sits at the start or the end**. That rules out anything positional — no "tail survives", no window, no eviction. It means prompt content is being *dropped uniformly*, which is a very specific shape.

That is exactly what H3 predicts. In `Glm5NextLinearAttention.__call__` (`language.py:217`), a non-`None` `lengths`/`left_padding` on the KDA `ArraysCache` flips `has_right_padding`, and then:

```python
mixed = mx.where(mask[..., None], mixed, 0)
```

zeroes prompt tokens **position-independently** — a planted sentence disappears wherever it is. And because `advance(S)` decrements `lengths` as the cache advances, a stale value zeroes progressively *more* of a longer prompt, which is a threshold-shaped failure rather than a gradual one. It also survives every one of your eliminations: nothing to do with the prefix cache, the scheduler, chunk geometry, or MTP.

The `gpt-oss-20b`-in-the-same-process control is a good one and is consistent with this — the mask path is in the glm5_next KDA layer, so a non-hybrid model in the same server is unaffected.

**Cheapest decisive test**, one line, no behaviour change:

```python
# top of Glm5NextLinearAttention.__call__
if cache is not None and (getattr(cache, "lengths", None) is not None
                          or getattr(cache, "left_padding", None) is not None):
    logger.warning("glm5_next KDA cache carries lengths=%s left_padding=%s",
                   cache.lengths, cache.left_padding)
```

If that fires on a failing request and not a passing one, the bug is found and the fix is trivial (clear them at batch-1, or don't let them leak in). If it never fires, H3 is dead too and the field narrows to H2 — the `mx.depends` pool↔KV ordering edge that's applied on the dense/`L==1`/fallback paths but **skipped** on the three sparse fast-path returns (`language.py:612-671`).

I could not find a setter for those fields on the glm5_next path — `glm_moe_dsa`'s `generate_patch.py:148` is gated to that model_type at `:41` — so if H3 is right the leak is coming from somewhere I didn't locate, and the assert is the way to find it rather than more grepping.

### On the MTP commits

The four commits look right, and the head-cache one is a finding we didn't have — `PoolingCache.offset` counting pooled rows rather than tokens is a real trap, and `_omlx_mtp_head_clone=True` is the conservative call. Two notes: your all-or-nothing rollback (check every layer before mutating any) is stricter than what I proposed and is the better design; and I agree the head-cache change invalidates the acceptance/tok-s numbers in the description, so those want a re-measure before anyone quotes them.

We're not able to run the Phase 0 matrix ourselves — our hardware is committed elsewhere and we've moved this model to a different serving stack — so please treat the plan as yours to take or discard. Happy to keep reasoning about traces if you post them.

davidtorcivia:
@gilby The long-prompt defect is none of H1 through H4. It is a stride bug in the glm5_next affine prefill tile, filed separately as #3275.

`linear_forward` and `fused_quantized_matmul` dispatch to `qwen35_q{bits}_affine_qmm_t` at 1024 tokens for 8-bit weights and 128 otherwise. The tile ignores the input's strides. `Glm5NextLinearAttention._fused_in_proj` splits one fused matmul on the last axis, and the f_a and g_a views are the inputs to `f_b_proj` and `g_b_proj`, so the forget gate and the output gate of every linear-attention layer read the wrong lanes. The fix is `mx.contiguous` on `x` before the dispatch.

This accounts for the parts of your report that the hypotheses did not. Two independently produced quants behave identically because both run this file. Other stacks are clean because the tile is oMLX only. Your fragmented-chunk run passed because every chunk was under 1024 and missed the 8-bit route. On our checkpoint 27 of the 34 linear-attention layers are 8-bit and 7 are 5-bit, so 7 layers are wrong from 128 tokens and the rest join at 1024, which is why short prompts still answer correctly. I did not run the `lengths`/`left_padding` assert; the tile explains the full symptom and the copy removes it.

Ladder on `GLM-5.3-Flash-oQ4e`, M3 Ultra 256 GB, greedy, fresh prompt per request. Before: correct up to 1018 prompt tokens, wrong at 1025, 1465, 1730, 2374, 4693, 9574 and 19365. After: correct at all of them. A three-turn conversation over a 5.4k-token document with prefix-cache hits also answers correctly on every turn.

The tok/s and acceptance figures in the description above were measured before that fix and are wrong. Re-measured with it, 4391-token prompt, greedy, 1500 tokens, three consecutive runs: 25.7, 28.6 and 30.2 tok/s wall as the prefix cache warms, decode-only about 32 to 33 tok/s. Acceptance is 64.3 to 65.0 percent at 2.00 to 2.05 tokens per cycle. The description claims 44.1 to 47.6 tok/s and acceptance near 95 percent at 3.58 tokens per cycle; the corrupted gates made the backbone more predictable, so the head agreed with it more often. I will update the description once the `mtp_enabled: false` arm is re-measured on the fixed build.

scubamount:
## Verified

- **Gap closed:** `glm5_next` absent from `_is_mtp_compatible` and no runtime in `patches/mlx_vlm_mtp/`. Head at `layers.<num_hidden_layers>.*` (not `mtp.*`), fusion `eh_proj`/`enorm`/`hnorm`, plain pre-norm `(B,S,D)` head (no `hc_attn_*`/`hc_ffn_*`) — matches the published GLM-5.3-Flash layout.
- **The real gem:** `_RUNTIME_PATCHES_BY_TYPE` scoping table. Previously the runtime patch swept ALL sub-patches on every load, rebinding shared parent classes across unrelated architectures. Table maps model_type → its owning sub-patch (longest-prefix match so family variants land on the family entry), `qwen4_exp: ()` never receives the sweep, unlisted types keep historical behavior. Correctly identifies that "applying everything reaches models this load has nothing to do with."
- Backward-compat: exact → longest-prefix → historical sweep. Safe by construction.

No issues. The scoping-table fix is valuable independent of the GLM feature.

jundot:
Thanks for adding GLM-5.3-Flash support and working through the cache rollback fixes. I tested an oQ4e backbone with the original MTP head, using native kernels and adaptive depth up to 3.

For a 768-token code-generation request, the median decode speed across three runs was **34.08 tok/s with this PR** and **35.47 tok/s with my FFN compilation follow-up**, about **4.1% faster**. The rollback checks passed. Fixed-input logits matched in the compilation comparison, though complete generation text differed under adaptive depth.

LGTM with two small fixes I'll apply on main after merging: keep quantized nextn `eh_proj.scales` and `biases` alongside their weight, and compile the stateless FFN for single-stream MTP verify widths up to 8. I'll include regression tests for both.

PR #3232, opened 2026-08-28
fix(qwen4_exp): restore PLE state after rejected Lightning MTP drafts

hojin12312:
## Summary

Qwen4-Exp PLE carries its own recurrent state alongside the inherited QSA/GDN cache: `ArraysCache[3]` is the n-gram history and `ArraysCache[2]` is the short-convolution state. A Lightning MTP verifier forward runs one confirmed token plus the speculative drafts. When some drafts are rejected, the inherited rollback rewinds QSA/GDN but leaves those two PLE slots at the end of the verify window, so the next step sees state that still includes the rejected tokens. This PR rewinds the PLE slots to the committed prefix and wraps the rollback in a fail-closed transaction.

## Minimal reproduction

```text
Before cycle:
  PLE history = [14542, 11]

Verifier input:
  confirmed = 321
  draft     = 16908

draft 16908 is rejected

Expected committed history: [11, 321]
Actual before fix:          [321, 16908]
```

Reproduced on clean upstream (`a20ec09d`) before the fix.

## Root cause

The inherited Qwen3.5 rollback restores QSA and GDN but never touches the PLE slots `ArraysCache[2]` / `[3]`. After a rejection or partial accept they stay at the verify-window end instead of the committed prefix.

## Fix

- During a multi-token verify forward, capture the pre-window n-gram history and the normalized short-conv inputs.
- Let the inherited QSA/GDN rollback run first.
- Rebuild the PLE slots for the confirmed token plus the accepted drafts: `retained = clip(accepted + 1, 0, window)`, then a gather on the sequence axis.
- A fully accepted cycle never calls rollback, so that path is untouched.

This stays inside Qwen4-Exp PLE. Generic `ArraysCache` behaviour, the speculative-decode driver, and the acceptance policy are unchanged.

## Transaction hardening

The follow-up commit (`1e7bb3eb`) makes the PLE rollback fail closed around the captured state:

- The snapshot is armed only on the verify forward. An ordinary forward clears anything left over from a fully accepted cycle, so a stale snapshot can't be picked up by a later rollback.
- It isn't usable until both halves, history and short-conv, have been captured in the same forward; a partially written snapshot is rejected rather than applied.
- Before the inherited rollback mutates anything, every attached snapshot is checked. A bad one raises `_PLESpeculativeRollbackError` with nothing changed, and the caller falls back to rebuilding the cache from the committed tokens, so slots 0/1 and 2/3 can't end up at different positions.
- Each snapshot is consumed once and cleared; anything still attached when the call returns is dropped.

The restore arithmetic is unchanged; the method is split into `_validate_ple_snapshot` and `_restore_ple_state` for the two phases.

## Tests

Added to `tests/test_mlx_vlm_qwen4_exp_compat.py`:

- full accept, a depth-1 rejection, and a partial accept;
- PLE history and short-conv state checked against a sequential committed-prefix replay, including the first target forward after rollback;
- an ordinary forward clears a stale snapshot;
- rollback on an incomplete snapshot fails closed, leaving the PLE slots and the QSA offset alone;
- snapshot validation runs before any QSA mutation.

The relevant Qwen4-Exp and Lightning MTP suites run 324 passed, 1 failed locally; the one failure is the pre-existing baseline miss described below. GitHub Actions is green on `1e7bb3eb` for Python 3.11, 3.12, and 3.13.

## Real-model validation

```text
Model:        Jundot/Qwen3.8-Flash-Next-oQ4e-mtp
Revision:     2615fc0e976e65c2f3b55daca3a948f1cdc5b9f8
PLE SSD mmap: enabled
```

- The original restore showed 0 rejected-draft history leaks in instrumented runs: committed history matched the replay on every reject cycle, with 0.0 max abs error on the short-conv reconstruction.
- Old-vs-new A/B (`a34fa0ad` vs `1e7bb3eb`; isolated engine, greedy, MTP on, prefix cache off, prose prompt at 8K, `max_tokens` 512, two repeats per cell): the committed greedy token stream is byte-identical between the two trees at depth 1 and depth 3. Both use the same verifier execution shape in this A/B, so that confirms the hardening did not change what the model commits.
- Depth-1 acceptance sat at ~0.808 for both trees; depth-3 ran ~0.79–0.82, drifting run to run as the depth controller adapts draft width.
- Across the reject cycles in these runs (55 at depth 1, 66–86 per run at depth 3) history and short-conv state matched the committed-prefix replay every time.
- None of the fail-closed paths fired; in these workloads the new checks stay dormant.
- No measurable decode regression between the two trees.

## Additional long-context validation

A separate matrix ran this implementation out to 128 KiB on a retrieval workload, on an isolated engine:

- deterministic across repeats at 2K / 8K / 32K / 64K / 128K;
- retrieval 5/5 at every size, with needles at 5 / 25 / 50 / 75 / 92 % depth;
- at 128K, MTP off/on and prefix cold/warm all produced the same output-token hash;
- prefix cold and warm matched at every size, and swap delta was zero at 128K;
- the direct PLE rollback probe ran at 8K / 32K / 64K and at 32K with prefix reuse, with 0 history mismatches and 0.0 short-conv error on every natural reject cycle.

The 128K retrieval run accepted every draft, so it did not exercise the rejection rollback path; that coverage comes from the smaller sizes. 256 KiB was not run to completion and is not part of the evidence here.

## Existing baseline note

`test_qwen4_lightning_mtp_fusion_and_runtime_attachment` misses `mx.allclose(..., atol=2e-5)` locally by about 5e-4 (`0.0796372` vs `0.0796386`). The same result reproduces on clean base `a20ec09d`. This PR does not modify the fusion implementation or the forward computation that test exercises, so the local tolerance miss is a pre-existing baseline difference. GitHub Actions is green for Python 3.11 / 3.12 / 3.13 on `1e7bb3eb`.

## Scope boundary

This PR fixes the PLE committed-state corruption from rejected or partially accepted Lightning MTP drafts and adds the local fail-closed hardening around that rollback.

It does not claim that MTP-on and MTP-off decode bit-identically for every prompt. Retrieval prompts matched exactly through 128K; on prose and low-entropy prompts both modes stay deterministic but can pick different tokens. In the instrumented runs those differences came with no recurrent-state mismatch and are consistent with the verifier's execution shape and floating-point reduction order.

### DIFF
--- a/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
+++ b/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
@@ -45,12 +45,30 @@ _MTP_RUNTIME = Qwen4ExpMTPRuntime()
 
 @dataclass
 class _PLESpeculativeState:
-    """PLE inputs needed to restore a partially accepted verify window."""
+    """PLE inputs needed to restore a partially accepted verify window.
+
+    Captured across a single target-verify forward and consumed exactly once by
+    ``rollback_speculative_cache``. ``complete`` gates the consume: it is set
+    only after *both* the n-gram-history capture and the paired short-conv
+    capture have run in the same forward, so a snapshot left half-written by an
+    aborted forward is rejected instead of being applied as a mixed-epoch state.
+    """
 
     history: mx.array
     input_ids: mx.array
     conv_state: mx.array | None = None
     conv_inputs: mx.array | None = None
+    complete: bool = False
+
+
+class _PLESpeculativeRollbackError(RuntimeError):
+    """Raised before any cache mutation when a PLE snapshot cannot be trusted.
+
+    The caller's speculative-decode loop treats a raising ``rollback_speculative_cache``
+    as a failed rollback and reconciles by rebuilding the cache from the committed
+    token stream, so failing closed here is safe and leaves no partially rolled
+    back cache behind.
+    """
 
 
 def configure_mtp_runtime(
@@ -1374,11 +1392,19 @@ class Qwen4ExpNGramEmbedding(nn.Module):
                 (batch, self.context_len), self.eos_token_id, dtype=mx.int64
             )
 
-        if capture_speculative_state and cache is not None:
-            cache._qwen4_exp_ple_speculative_state = _PLESpeculativeState(
-                history=previous_context,
-                input_ids=input_ids,
-            )
+        if cache is not None:
+            if capture_speculative_state:
+                # Arm a fresh snapshot for this target-verify forward.
+                cache._qwen4_exp_ple_speculative_state = _PLESpeculativeState(
+                    history=previous_context,
+                    input_ids=input_ids,
+                )
+            elif getattr(cache, "_qwen4_exp_ple_speculative_state", None) is not None:
+                # Ordinary forward: any snapshot still attached belongs to an
+                # earlier verify cycle that was fully accepted (no rollback) or
+                # aborted. It can never describe this committed position, so
+                # disarm it before it can be consumed by a later rollback.
+                cache._qwen4_exp_ple_speculative_state = None
 
         token_history = mx.concatenate([previous_context, input_ids], axis=-1)
         if cache is not None:
@@ -1467,6 +1493,8 @@ class Qwen4ExpPLELayer(nn.Module):
             if snapshot is not None:
                 snapshot.conv_state = state
                 snapshot.conv_inputs = x
+                # Both paired captures for this verify forward have now run.
+                snapshot.complete = True
         if cache is not None:
             cache[2] = mx.contiguous(conv_input[:, -self.short_conv_state_len :])
         return nn.silu(self.conv1d(conv_input))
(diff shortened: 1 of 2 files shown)


jbbottoms:
We hit this same defect independently earlier tonight and shipped a fix to our own deployment a few hours before finding this PR — sharing measurements and review notes in case they help land this one stronger.

**Measured impact of restoring PLE state** (Qwen3.8-Flash-Next, M5 Max 128GB, oQ4 pack, `qwen4_ple_ssd_offload`, Lightning MTP depth 3, greedy):

- acceptance before any fix: **62–65%** (the contaminated ring feeds the drafter wrong context; we counted 133/133 rejected cycles advancing slots 2/3 with drafted tokens via live telemetry)
- after restoring both slots: **73–81%**, tok/cycle 2.0 → 2.4–3.2
- after adding a transaction contract around the capture (below): **92–94%**
- end-to-end decode +4–10% depending on workload shape

**Exactness note.** A slice-of-copies restore is provably exact for both slots: with `B = state ‖ window`, the serial state after `n` retained positions is `B[:, n : n+width]`. For the conv slot, dilation widens the *cache*, not the required window — the 9-column input delay line is the entire needed past, so no replay or re-lookup is required. The `_PLESpeculativeState` capture here holds equivalent information; on partial accepts it just needs to select at the accepted count rather than restore the pre-verify state wholesale.

**Review suggestions** (an external-model audit of our own equivalent patch found these; they apply to any variant of this fix):

1. Arm the capture **only** on target-verify forwards, and have ordinary forwards *disarm* it — otherwise a stale snapshot from a later forward can be consumed by a retried rollback.
2. Pair the two captures (history vs conv) with a shared epoch: an exception between the two writes leaves a mixed-epoch snapshot that passes shape checks.
3. Consume exactly once; reject duplicate rollback.
4. Validate the snapshot **before** any GDN/QSA mutation and fail closed into the reconciliation path. The dangerous default is a silent skip after partial rollback: it reports success with slots 0/1 rolled back and 2/3 still speculative.

**Two verification gotchas** that cost us time tonight:

- This vendored module imports as `mlx_vlm.models.qwen4_exp`, so its `logger` output does not propagate into the omlx server log — when proving the restore actually fires, use a side-effect probe, not `logger.info`.
- Byte-comparing MTP-on vs MTP-off output is not a valid oracle for this fix: the multi-token verify forward selects different kernel shapes than serial decode, so greedy streams diverge from floating-point reduction order alone, before any rejection is involved (serial-vs-serial is byte-stable across restarts). Acceptance-rate and task-quality gates are the meaningful checks.

Happy to share more details from our runs if useful.

hojin12312:
Pushed `1e7bb3eb` on top — a transaction contract around the PLE speculative-state restore, responding to the review points above.

**Necessity analysis (current oMLX control flow).** None of the four turned out to be a live bug on the paths that actually run:

1. *Arm only on target-verify / disarm ordinary forwards* — arming is already `target_verify`-only (`gdn_sink is not None`). The disarm half is not reachable as a bug (no `rollback_speculative_cache` call is ever reached without an immediately preceding verify that re-arms), but after a fully-accepted cycle the snapshot does linger on the cache until the next verify overwrites it. Implemented the disarm anyway — cheap, and it stops relying on that invariant.
2. *Single epoch for the two captures* — they're unconditionally paired inside one `Qwen4ExpPLELayer.__call__` with one shared flag, so a mixed-epoch snapshot can't reach a rollback. Added a `complete` flag as explicit insurance (zero cost).
3. *Consume exactly once* — already true for the PLE slots (cleared unconditionally per call; a retried rollback is a clean no-op for PLE).
4. *Validate before GDN/QSA mutation, no silent partial success* — the silent-skip path isn't reachable; a raise after `super()` is contained by the caller's full cache rebuild (`_MtpStepFallback` → `_reconcile_mtp_to_standard`). Implemented validation before `super()` so the method is atomic on its own rather than depending on the caller.

Net: latent hardening that makes the rollback locally atomic. Restore math unchanged; refactor into `_validate_ple_snapshot` / `_restore_ple_state` for the phases. Scope stays Qwen4-Exp-PLE-local — no generic `ArraysCache` or speculative-decode-driver changes.

**Real-model A/B** — `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp`, isolated engine, greedy, MTP on, prefix cache off, prose prompt @ 8K, `max_tokens` 512, 2 repeats/cell, old (`a34fa0ad`) vs new (`1e7bb3eb`):

| | depth 1 | depth 3 |
|---|---|---|
| output SHA-256 old vs new | identical (`5c63ee9b…`) | identical (`ff3813f5…`) |
| same-mode determinism | PASS both | PASS both |
| acceptance rate | 0.808 both | 0.79–0.82 (varies run-to-run, both trees) |
| accepted tok / verify cycle | 0.81 both | 1.3–1.6 (both trees) |
| decode tok/s | 46–50 both | 47 both |
| PLE state mismatch (history + short-conv vs serial replay) | 0 / 55 reject cycles | 0 / 66–86 reject cycles |
| incomplete / stale / fail-closed events | 0 | 0 |

Old and new produce byte-identical greedy streams (same kernel shapes, so that comparison is a valid oracle) at zero measured throughput cost. Over the reject cycles in these runs the *old* patch already had zero history/short-conv mismatch, so the new branches are latent guards — `fail_closed_events == 0` throughout.

I'm treating your 92–94% as an independent measurement on a different deployment, not a target — a transaction contract prevents corruption, it doesn't raise acceptance, and model/prompt/runtime differ. Thanks for the detailed notes; the "use a side-effect probe, not `logger.info`" and "MTP-on vs MTP-off is not an oracle" points both saved time.

Local suite: 324/325 green; the one failure (`test_qwen4_lightning_mtp_fusion_and_runtime_attachment`, `atol=2e-5` at ~5e-4) reproduces identically on base `a20ec09d`. This PR does not modify the fusion implementation or the forward computation exercised by that test, so the local tolerance miss is a pre-existing baseline difference.

jbbottoms:
Following up after 1e7bb3eb — we've now run this mechanism head-to-head against our internal implementation of the same fix (built independently before we found this PR), including an AST-level equivalence pass and a 324-case slice oracle across widths 1-33, windows 1-8, and every accepted count. The mechanisms agree, and your reproduction values match our oracle exactly.

**Endorsement with receipts:** on M5 Max 128GB with Lightning MTP, greedy — acceptance 62-65% unfixed, 73-81% with a plain restore, **92-94% with the transaction semantics this PR now carries**, decode 52.4 -> 58.3+ t/s from the acceptance recovery alone, and output-exactness maintained against a serial MTP-off oracle over a 24-probe corpus. This is a critical correctness fix for the flagship model and, as far as we can measure, it's ready.

Two small improvements from the comparison, offered take-or-leave:

1. **Same-accepted-count fast path in `_restore_ple_state`.** The current restore always builds a broadcast int32 index tensor shaped `(batch, state_len, hc_hidden_size)` and gathers. When every row shares one accepted count — always true at batch 1, the common case otherwise — a contiguous slice `buffer[:, n : W + n]` produces the same result and skips one large index allocation per PLE layer per rollback:

```python
if len(set(counts)) == 1:
    start = counts[0] + 1
    cache[slot] = mx.contiguous(buffer[:, start : width + start])
else:
    ...existing gather path...
```

2. **Epoch stamping instead of the boolean `complete` flag.** Stamping each capture half with a per-verify epoch counter and requiring both halves to match is equivalent for the half-written case the flag targets, but it would also catch a cross-layer stale capture if a forward ever aborts mid-stack — a case neither form catches today, and the boolean can't be extended to.

Happy to turn either into a follow-up commit here if useful. Thanks @hojin12312 for landing this fast — and @jundot, from where we sit this one belongs in the critical-bug lane.

jundot:
Thanks for fixing the Qwen4-Exp PLE rollback path. I reproduced the issue on the base revision with `Qwen3.8-Flash-Next-oQ6e-mtp`: all 57 natural reject or partial-accept rollbacks left both the n-gram history and short-conv state different from the independently reconstructed committed prefix.

On this PR, the same probe covered 108 rollback cycles across two deterministic runs with zero state mismatches and zero short-conv error. The two PR-head runs also produced identical output and MTP acceptance stats, and the focused and related test suites passed with 325 tests.

The snapshot-and-restore mechanism is correctly scoped to Qwen4-Exp and fixes the runtime corruption. LGTM. I’ll fold a small non-blocking cleanup into a follow-up on `main` to make snapshot capture atomic and trim redundant defensive validation and documentation without changing the rollback semantics.

PR #3235, opened 2026-08-28
perf(qwen4_exp): prefetch and batch the SSD n-gram gather

latent-variable:
M5 Max 128 GB, macOS 26.5.2, `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp`, PLE SSD offload on.

## Cold prefill, interleaved, n=4 per cell

| pp | `main` | this PR | speedup | prefill |
|---|---|---|---|---|
| 1024 | 213.1 tok/s | 843.8 | 3.96x | 5.03 s → 1.27 s |
| 4096 | 239.8 | 1043.3 | 4.35x | 17.40 s → 3.99 s |
| 16384 | 282.4 | 734.2 | 2.60x | 58.03 s → 22.50 s |

Run-to-run spread 0.7 to 4.5 percent. The advantage narrows at 16K because early chunks of a long prompt warm the rows later chunks need, so the cold cost amortizes within the prompt.

Resident memory is unchanged: 69.67 GB against 69.61 GB for `main`, on a 103.94 GB full model.

## Method

Each trial is one fresh process: `sudo purge`, start the server, load the model, a 32-token warmup, then one measured prompt. The warmup keeps first-forward costs out of the measured window while touching about 512 of 320 million n-gram rows, so the table stays cold. Model load is timed separately and excluded. Builds alternate every trial so machine drift lands on both arms equally, the paged prefix cache is emptied before each trial, and a failed purge aborts the run rather than producing warm numbers.

## The gather on its own

Same chunk size the model uses, on rows `mincore` confirms are not resident, three independent row sets each:

| | us/row | ms/token |
|---|---|---|
| `main` | 145-181 | 2.33-2.90 |
| this PR | 22.4-23.4 | 0.36 |

About 7x on the gather becomes 2.6 to 4.4x end to end, the difference being the compute this PR does not touch.

## Cause

The model asks for 16 n-gram rows per token and the gather reads them one at a time, so each row is a page fault with nothing else in flight. macOS pages are 16 KB, so an 80-byte row forces a 16 KB read and the device never sees a queue.

## Change

- Collect the page ranges for the whole batch and advise them all before reading any, from a small thread pool. libc `madvise` through `ctypes`, because `mmap.madvise` holds the GIL: through it, 16.9 us/row at 8 threads and 16.4 at 1, against 4.8 us/row through `ctypes` at 8.
- Bucket by shard with `searchsorted` and `argsort` instead of a Python scan over the gather once per touched shard. The resident `ShardedEmbedding` gets the same fix.
- Read each distinct row once per shard.

Prefetch is advisory. Without `ctypes` or `madvise` the gather behaves exactly as it does today. About 100 lines inside the gather already in `main`: no new files, no new dependencies, no change to the mlx-vlm pin.

There is no gain on a warm page cache, where `main` and this PR measure the same. This targets a first pass over new content, a fresh model load, a context change, or a machine whose spare RAM cannot hold the table.

## Reproducing it

```
sudo purge
omlx serve --model-dir <dir> --port 5599
curl .../v1/chat/completions -d '{"model":"...","max_tokens":4,"messages":[{"role":"user","content":"warmup"}]}'
curl .../v1/chat/completions -d '{"model":"...","max_tokens":1,"messages":[{"role":"user","content":"<~4k tokens>"}]}'
```

Subtract `model_load_duration` from the wall time of the second call. Alternate builds rather than running one after the other, empty the prefix cache between trials, and confirm the purge actually ran; each of those changes the result by more than the effect being measured.

## Test plan

- Parity tests gather shuffled and repeated indices across shards against a plain table lookup, for the disk-backed and resident classes, plus empty and out-of-range batches. Control: deleting the row-expansion line makes the parity test fail.
- `pytest tests/ -m "not slow and not integration"`: 10404 passed, plus 5 new tests. The 39 failures in my environment are byte-identical to unmodified `main` at a20ec09d (local mlx drift, missing optional deps); this branch adds none.

### DIFF
--- a/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
+++ b/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
@@ -5,8 +5,9 @@ import math
 import mmap
 import os
 import struct
+import threading
 import weakref
-from bisect import bisect_right
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from dataclasses import dataclass, replace
 from pathlib import Path
 from types import SimpleNamespace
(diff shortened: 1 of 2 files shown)


latent-variable:
Rebased onto 87066830. The conflict was in `tests/test_mlx_vlm_qwen4_exp_compat.py`, where this branch and 5463c0f4 both appended to the end; I kept both. The `language.py` diff is byte-identical to the pre-rebase one.

Re-measured on the new base. Gather on `mincore`-verified cold rows, 8192 rows per trial, arms alternating, three trials each:

| | us/row | ms/token |
|---|---|---|
| `main` 87066830 | 130.6 to 186.8 | 2.09 to 2.99 |
| this PR | 15.8 to 18.1 | 0.25 to 0.29 |

Suite, both arms, same environment: `main` 10471 passed / 39 failed, this branch 10475 / 39, identical failure sets node ID for node ID. The four extra passes are this PR's tests, the 39 are local environment drift, and CI is green on 3.11, 3.12 and 3.13.

Unrelated to this branch: `test_qwen4_lightning_mtp_fusion_and_runtime_attachment` fails on pristine `main` here, `atol=2e-5` with no `rtol` against outputs spanning 0.08 to 3.4. I can open an issue if that is useful.

latent-variable:
Closing this. It does not hold up on 0.6.4.

I re-ran it on a properly built install, `v0.6.4` with `mlx==0.32.0` and the custom kernels compiled, using the same cold protocol this PR specifies: purge, fresh process, model load, 32-token warmup, one measured prompt. M5 Max 128 GB, `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp`, PLE on SSD, arms alternated, n=3 per cell.

| prompt | `main` | this PR | |
|---|---|---|---|
| 12.2k | 899.7 tok/s | 700.0 tok/s | -22.2%, wins 0 of 3 |
| 29.4k | 735.8 tok/s | 718.8 tok/s | -2.3%, wins 1 of 3 |

The original numbers were taken against 0.6.3 on `mlx 0.31.2` with no kernels built, where the gather really was a serial run of page faults with nothing else in flight. #3244 removed that bottleneck. With the Metal pipeline doing the work, the `madvise` thread pool competes for bandwidth rather than buying latency, and the batching no longer has anything to hide.

The gather microbenchmark in the PR body still reports 6.0x on the current build, 117.5 us/row against 19.4 on rows `mincore` confirms are not resident. That number is still correct and it stopped predicting anything, because it samples rows uniformly across all 320M while real prompts repeat phrases and touch far fewer distinct rows, many already warm. I was measuring the component instead of the system.

One thing from the re-test that might be worth knowing, offered as data rather than a request. On an install without the custom kernels, which is what a plain `pip install -e .` produces since the build is opt-in, the gathered QSA prefill fallback measured slower than the official mask path it replaces: 701 tok/s against 1110 at 12.2k, and 260 against 667 at 29.4k, the gap widening with context. `0a6ae09f` already gates the wide-prefill floor on `qwen4_qsa_sparse_gqa_attention`; the eligibility check for the path itself is not gated. Happy to open that separately with the numbers if it is useful, or to leave it, since building the kernels is the real answer.

Thanks for the review time on this one.

PR #3243, opened 2026-08-28
Preserve explicit capture_layer_ids in qwen35 VLM MTP runtime (fixes qwen4_exp Lightning MTP collision)

### LINKED ISSUE #3242: Loading a qwen3_5 MTP model breaks Lightning MTP on a resident qwen4_exp model — qwen35_vlm_runtime clobbers the capture_layer_ids=[] sentinel (one-line fix included)
## Summary

Lightning MTP on a `qwen4_exp` model (Qwen3.8-Flash-Next) works fine — until a `qwen3_5` MTP model (e.g. Qwen3.8-27B) is loaded into the same server process. From that moment, **every** request to the qwen4_exp model fails with:

```
RuntimeError: Qwen4 Lightning MTP expects hidden shape [batch, tokens, hc_count * hidden_size].
```

and stays broken until the server process restarts (the harness just sees dead/empty responses). oMLX 0.6.3 final, M3 Ultra 512 GB.

## Repro

1. Fresh server. Chat with `Qwen3.8-Flash-Next-oQ4e-mtp` (`mtp_enabled: true`) → works, MTP drafting fine.
2. Start a second chat with `Qwen3.8-27B-oQ4e-fp16-mtp` (`mtp_enabled: true`) → works.
3. Send anything to Flash Next again → the ValueError above, on every request, single-stream included. Settings changes, model unload/reload, cache deletion don't recover it — only a process restart does (and then only until the qwen3_5 model loads again).

## Root cause

`qwen35_vlm_runtime.apply()` (triggered by the qwen3_5 model's load) replaces `mlx_vlm.models.qwen3_5.language.LanguageModel.__call__` **class-wide**. The vendored `qwen4_exp` `LanguageModel` subclasses that exact class and relies on `super().__call__` honoring its `capture_layer_ids=[]` sentinel (which requests the full pre-mixer hyper-connection streams for the MTP head, per the inner model's `capture_layer_ids == []` branch).

The patched `__call__` clobbers the caller's value:

```python
# qwen35_vlm_runtime.py, inside the patched __call__
kwargs.pop("capture_layer_ids", None)
...
out = original_call(..., capture_layer_ids=[last_layer_idx], ...)
```

With `[last_layer_idx]`, the qwen4 inner model takes the per-layer capture branch and appends `hyper_connection_mixer(hidden)` — the **mixed, hidden_size-wide** tensor — so `Qwen4ExpMTPModule.fuse_inputs` receives the wrong width and raises. Before the qwen3_5 model loads, the sentinel reaches stock `LanguageModel.__call__`, which allocates the sinks for any non-None list (`[] if capture_layer_ids is not None else None`), and everything works — which is why the failure looks like "model X broke when I opened a chat with model Y".

The MoE variant (`qwen35_moe_vlm_runtime.py`) has the same `kwargs.pop` pattern and presumably the same collision with any future qwen4-style subclass of the MoE class.

## Fix (verified locally)

Preserve an explicitly-passed capture list instead of clobbering it:

```python
explicit_capture = kwargs.pop("capture_layer_ids", None)
last_layer_idx = len(self.model.layers) - 1
out = original_call(
    self,
    inputs,
    inputs_embeds,
    mask,
    cache,
    capture_layer_ids=(
        [last_layer_idx] if explicit_capture is None else explicit_capture
    ),
    **kwargs,
)
```

With this one-line change (applied to my local install): fresh server → Flash Next OK → 27B loads and answers OK → Flash Next still OK with both resident, MTP active on both; debug instrumentation confirmed the sink then carries the full-width `(1, n, hc_count*hidden)` tensor through the whole chain. qwen3_5 models are unaffected (they never pass an explicit capture list on this path).

Happy to open a PR with the change if useful. Related reports from the same setup: #3233 (ANE prefill corruption), #3236 (context discovery).

daveycodez:
## Bug

In `omlx/patches/mlx_vlm_mtp/qwen35_vlm_runtime.py`, the patched `LanguageModel.__call__` wrapper unconditionally clobbers any caller-passed `capture_layer_ids`:

```python
kwargs.pop("capture_layer_ids", None)
...
capture_layer_ids=[last_layer_idx],
```

The vendored qwen4_exp LanguageModel (Qwen3.8-Flash-Next) subclasses the qwen3_5 LanguageModel and passes the `capture_layer_ids=[]` sentinel through `super().__call__` to request the full pre-mixer hyper-connection streams for its Lightning MTP head.

## Mechanism

The patch is applied class-wide, so once a qwen3_5 MTP model loads, the qwen4_exp sentinel gets replaced with `[last_layer_idx]`. The inner model then returns the mixed (hc-narrow) hidden, and every subsequent qwen4_exp request crashes with:

```
ValueError: Qwen4 Lightning MTP expects hidden shape [batch, tokens, hc_count * hidden_size].
```

until the process is restarted.

## Fix

Preserve an explicitly-passed capture list; only default to `[last_layer_idx]` when the caller passed nothing (`None`). The identical `pop` + hardcoded-default pattern in `omlx/patches/mlx_vlm_mtp/qwen35_moe_vlm_runtime.py` gets the same fix, since a future MoE-based subclass would hit the same collision.

## Verification

Verified on an M3 Ultra 512GB with Qwen3.8-Flash-Next-oQ4e-mtp and Qwen3.8-27B-oQ4e-fp16-mtp resident simultaneously (both with Lightning MTP active): without the fix, qwen4_exp requests fail with the ValueError above once the qwen3_5 model has loaded; with the fix, both models serve correctly. `python3 -m py_compile` passes on both files.

Fixes #3242

### DIFF
--- a/omlx/patches/mlx_vlm_mtp/qwen35_moe_vlm_runtime.py
+++ b/omlx/patches/mlx_vlm_mtp/qwen35_moe_vlm_runtime.py
@@ -266,10 +266,13 @@ def _patch_vlm_language_model(q35moe_lang: Any) -> None:
 
         # Passing any non-None ``capture_layer_ids`` makes stock
         # ``LanguageModel.__call__`` allocate ``hidden_sink`` AND ``gdn_sink``,
-        # both of which we need.  Pop any existing value from kwargs to avoid
-        # "got multiple values for keyword argument" when the caller already
-        # passed capture_layer_ids (e.g. speculative_verify_logits).
-        kwargs.pop("capture_layer_ids", None)
+        # both of which we need.  Pop the value from kwargs to avoid
+        # "got multiple values for keyword argument", but preserve an
+        # explicitly-passed list (e.g. speculative_verify_logits, or a
+        # subclass passing the ``[]`` sentinel to request full pre-mixer
+        # hyper-connection streams); only default to the last layer when the
+        # caller passed nothing.
+        explicit_capture = kwargs.pop("capture_layer_ids", None)
         last_layer_idx = len(self.model.layers) - 1
         out = original_call(
             self,
@@ -277,7 +280,9 @@ def _patch_vlm_language_model(q35moe_lang: Any) -> None:
             inputs_embeds,
             mask,
             cache,
-            capture_layer_ids=[last_layer_idx],
+            capture_layer_ids=(
+                [last_layer_idx] if explicit_capture is None else explicit_capture
+            ),
             **kwargs,
         )
         from mlx_vlm.models.base import LanguageModelOutput
--- a/omlx/patches/mlx_vlm_mtp/qwen35_vlm_runtime.py
+++ b/omlx/patches/mlx_vlm_mtp/qwen35_vlm_runtime.py
@@ -322,10 +322,13 @@ def _patch_vlm_language_model(q35_lang: Any) -> None:
 
         # Passing any non-None ``capture_layer_ids`` makes stock
         # ``LanguageModel.__call__`` allocate ``hidden_sink`` AND ``gdn_sink``,
-        # both of which the MTP cycle needs. Pop any existing value from kwargs
-        # to avoid "got multiple values for keyword argument" when the caller
-        # already passed capture_layer_ids.
-        kwargs.pop("capture_layer_ids", None)
+        # both of which the MTP cycle needs. Pop the value from kwargs to
+        # avoid "got multiple values for keyword argument", but preserve an
+        # explicitly-passed list: the qwen4_exp subclass passes the ``[]``
+        # sentinel through ``super().__call__`` to request the full pre-mixer
+        # hyper-connection streams for its Lightning MTP head, and clobbering
+        # it would hand the head the mixed (hc-narrow) hidden instead.
+        explicit_capture = kwargs.pop("capture_layer_ids", None)
         last_layer_idx = len(self.model.layers) - 1
         out = original_call(
             self,
@@ -333,7 +336,9 @@ def _patch_vlm_language_model(q35_lang: Any) -> None:
             inputs_embeds,
             mask,
             cache,
-            capture_layer_ids=[last_layer_idx],
+            capture_layer_ids=(
+                [last_layer_idx] if explicit_capture is None else explicit_capture
+            ),
             **kwargs,
         )
         from mlx_vlm.models.base import LanguageModelOutput


rsnow:
Independent validation of this fix on our M3 Ultra.

We reproduced the original failure with `Qwen3.8-Flash-Next-oQ4e-mtp` loaded first, then `Qwen3.6-27B-Fable-Fusion-711-Uncensored-Heretic-NM-DAU-oQ8e-mtp`: every subsequent Flash-Next request failed with the hidden-shape ValueError until the process restarted. MTP enabled on both models, Dual ANE for the 27B. Note the second model here is an oQ8e fine-tune rather than the stock Qwen3.8-27B used in the issue's original verification — same failure, same fix — so this is a second independent reproduction with a different qwen3_5-family VLM.

We also confirmed the reverse load order (27B first, then Flash-Next): Flash-Next fails on its very first MTP request. That points to the class-wide patch installation itself as the trigger, not model residency, shared caches, or resource pressure — which matches the mechanism described in the issue.

PR #3243 was tested in an isolated worktree:

- Base/negative check: 4 contract tests fail with the old unconditional `capture_layer_ids` clobbering (including both Qwen4-like load orders).
- PR head: those 4 tests pass.
- Relevant suites: 63 passed.
- Full suite: 10,411 passed, 165 skipped; no failures.
- The regression tests cover both dense and MoE wrappers, explicit `capture_layer_ids` preservation, omitted-value defaulting, and both load orders.

We then cherry-picked the PR source commit onto our production install for live validation (rollback tag preserved). With both MTP models loaded simultaneously, sessions on both models generated successfully after the 27B load; no recurrence of the hidden-shape failure. The two-model run stayed healthy at about 143 GB model memory. The only observed effect was expected prefill contention/temporary ANE-bank release, affecting throughput rather than correctness.

This confirms the small preserve-explicit-capture fix addresses the production failure. If it would be useful, we're happy to contribute a follow-up PR adding both-load-order regression tests as well.

jundot:
Thanks for the precise diagnosis and validation, and thanks @rsnow for the independent testing. The root cause was correct, but I chose the exact-type boundary in #3285 because it keeps the Qwen3.5 capture contract unchanged and also prevents its MTP initialization from reaching Qwen4 subclasses.

Since that fix is now on main and covered by 5463c0f4, merging this would overlap the same path with a broader behavior change. I’m closing this as superseded; the reproduction and validation were very useful in confirming the bug.

PR #3244, opened 2026-08-28
perf(qwen4): flatten exact QSA prefill and accelerate Lightning MTP

jonathan308:
## Why this matters

Qwen3.8-Flash-Next has only 6B active parameters, but its QSA path was still paying for large dense/gathered intermediates as context grew. On the M3 Ultra this made cold prefill taper far below the hardware's useful throughput even though model correctness was already good.

This PR replaces that bottleneck with exact, fail-closed Qwen4 paths. It does not approximate attention, change selected blocks, alter quantized projection results, or relax output tolerances. Unsupported shapes—including batched, padded, multimodal-position, and non-production geometries—stay on mlx-vlm's official implementation.

The result is a large practical change: cold 50K prefill rises from 619.48 to 895.83 tok/s, and the curve retains 95.9% of that rate through 150K. Lightning MTP reaches 82.31 tok/s over a sustained 500-token decode while preserving 99.5% draft acceptance.

### Exact QSA prefill and decode

- Add a portable gathered-QSA foundation for contiguous batch-one text, with the official path retained below the sparse crossover and for every unsupported shape.
- Cache normalized/rotated QSA block state and add exact native FP32 index scoring plus deterministic top-512 selection.
- Add a direct sparse-GQA Metal kernel for Qwen3.8's `24q / 2kv / D256` geometry. It consumes chronological four-token block IDs directly, appends the causal tail in-kernel, and keeps QK, online softmax, and PV accumulation in FP32 without materializing dense score or gathered-K/V tensors.
- Add the minimal generic decode-SDPA subset needed by selected-K/V QSA decode. The package contains SDPA only—no RMS/RoPE, DS4 sparse attention, or other unrelated decode kernels—and falls back to `mx.fast.scaled_dot_product_attention` if the extension, ABI, layout, dtype, or shape is not accepted.
- Use geometric QSA cache growth and 4,096-token prefill/cache blocks for high-memory classic-Metal Qwen hosts.

### Resident PLE and exact projection fusion

- Fuse the resident packed PLE shards without changing PLE values or the mmap/offload path.
- Add exact scalar GDN prework and exact Qwen4 hyperconnection projection kernels.
- Preserve the original raw 320-row and 4-row hyperconnection banks as the authoritative fallback. The exact scalar-decode path is tightly gated; Lightning-MTP verification and every other width keep the canonical split projections.

### MTP prefix lifecycle

- Store a bounded in-memory Lightning-MTP prompt-history sidecar under the same full-block chain hash as the backbone prefix cache.
- Restore it only when the matching backbone block is still live; evict it with individual block hashes, wholesale clears, and an LRU-4 bound.
- Preserve pending hidden state and every MTP K/V array bit-for-bit on cold-to-warm prefix reuse, avoiding a full warm-prefix re-prime.

### Qwen vision compatibility

- Backport mlx-vlm `1249c7d` / #1982's scalar temporal repeat count for the pinned mlx-vlm revision. This fixes image/video tower grid math on newer MLX while leaving the public oMLX API's existing `video_url` limitation unchanged.

## Performance

Hardware and model:

- Apple M3 Ultra, 256 GB unified memory, 819 GB/s memory bandwidth
- `Qwen3.8-Flash-Next-oQ4e-mtp`, resident packed PLE
- Batch 1, unique salted prompts, zero cached tokens for cold rows
- MTP, DFlash, ANE, and SpecPrefill disabled for the native prefill/decode baseline
- Physical performance runtime: MLX `0.32.2.dev20260825+ceab91938`, mlx-vlm `0.6.3`
- The clean branch also builds and passes its unit/native gates against oMLX's pinned MLX `0.32.0`, mlx-vlm `0.6.3`

### Cold prefill taper

| Prompt | Before direct QSA | This PR | Change |
|---:|---:|---:|---:|
| 20K | 631.15 tok/s | 886.09 tok/s | +40.4% |
| 50K | 619.48 tok/s | 895.83 tok/s | +44.6% |
| 100K | — | 877.44 tok/s | — |
| 150K | — | 858.80 tok/s | — |

The previous path was already flatter than the original dense implementation, but it left substantial throughput on the table. The direct path now loses only 4.1% from 50K to 150K. This is the main user-facing win: long prompts no longer collapse as context grows.

### Sustained decode and Lightning MTP

Protocol: 10K prompt plus 500 generated tokens, temperature zero.

| Mode | Prefill/cache state | Decode | Acceptance |
|---|---|---:|---:|
| Native, MTP off, cold | 829.2 tok/s | 35.90 tok/s | n/a |
| Native, MTP off, 8,192-token prefix reuse | warm prefix | 36.00 tok/s | n/a |
| Lightning MTP depth 5, cold | prompt head primed | 82.31 tok/s | 99.5% |
| Lightning MTP depth 5, warm | sidecar restored | 82.05 tok/s | 98.8% |

The sustained native decode baseline before these Qwen decode changes was 22.84 tok/s; 35.90 tok/s is a 57% improvement. MTP cold prefill still performs real prompt-head work to preserve acceptance; the sidecar specifically removes that replay on compatible warm prefixes rather than skipping correctness work.

## Correctness and fail-closed behavior

- FP32 QSA scores and deterministic selected-block set match the portable path.
- Direct sparse attention preserves chronological token order, causal tail semantics, and FP32 online softmax. Production-shape attention differs by at most one BF16 output ULP from the portable reduction order; qualified end-to-end gates retain identical greedy tokens and cache state.
- Raw checkpoint banks remain installed; optimized projection copies never replace the fallback state.
- QSA auxiliary state is rejected when cache lengths/positions do not align.
- Batches, padded rows, MRoPE/multimodal positions, target verify, unsupported dtype/layout/geometry, and missing native symbols use the official/composed path.
- MTP sidecar restore requires the exact model/token chain hash and a still-live backbone block.

## Validation

- `1,023 passed, 40 skipped, 3 deselected` across 1,063 selected Qwen4, MTP, prefix/paged-cache, scheduler, and Homebrew formula tests on Python 3.13.13 with oMLX's MLX 0.32.0 pin.
- Isolated SDPA native build/import succeeded against MLX 0.32.0; all 30 native parity, capability-gate, selected-K/V decode, and fallback tests passed.
- Source wheel built successfully (`omlx-0.6.3-py3-none-any.whl`, SHA-256 `79db3eaff221ba68af0240fe5789754bbfac779169cb11eb1014d4e82cd5cb50`), installed into an empty target, imported without native artifacts, and reported the expected exact MLX fallback.
- Image serving smoke succeeded with a correct description; Qwen image/video grid unit smokes pass. Public `video_url` still returns the pre-existing HTTP 400 and is not claimed as supported here.
- Tool call plus tool-result continuation passed.
- In-flight cancellation followed by health check passed.
- Clean unload freed 99.38 GB and left no loaded model.

## Deliberately out of scope

- Batching/concurrency changes are not included. A separate row-wise experiment exposed a late-join QSA/KV offset mismatch after one row finished, so this PR makes no concurrency-improvement claim; that work should follow with its own scheduler/cache review.
- Cluster/distributed inference, GLM, DS4, ANE, NVFP4, updater/release/branding, and profiling/trace hooks are not part of this branch.
- Public video request handling is unchanged.

## Attribution

- Qwen sparse-attention staging is adapted from David Dalcu's MIT-licensed mlx-serve work; the complete David Dalcu MIT notice is retained with the native sources and adapted GDN kernel.
- Steel/SDPA portions retain the complete Apple MLX MIT notice.
- The vision bridge retains Prince Canuma's complete mlx-vlm MIT notice.
- New oMLX-owned files carry Apache-2.0 SPDX identifiers.

### DIFF
--- a/Formula/omlx.rb
+++ b/Formula/omlx.rb
@@ -1,5 +1,5 @@
 class Omlx < Formula
-  CUSTOM_KERNELS = %w[bonsai glm_moe_dsa minimax_m3 qwen35_prefill].freeze
+  CUSTOM_KERNELS = %w[bonsai decode_fast glm_moe_dsa minimax_m3 qwen35_prefill].freeze
 
   desc "LLM inference server optimized for Apple Silicon"
   homepage "https://github.com/jundot/omlx"
@@ -10,7 +10,7 @@ class Omlx < Formula
   head "https://github.com/jundot/omlx.git", branch: "main"
 
   option "with-custom-kernel",
-         "Build native custom kernels for Bonsai, GLM-5.2, MiniMax M3 and Qwen3.5/3.6 acceleration"
+         "Build native custom kernels for Bonsai, GLM-5.2, MiniMax M3 and Qwen3.5/3.6/4 acceleration"
   option "with-grammar", "Install xgrammar for structured output (requires torch, ~2GB)"
 
   depends_on "rust" => :build
--- a/README.md
+++ b/README.md
@@ -435,5 +435,5 @@ Contributions are welcome! See [Contributing Guide](docs/CONTRIBUTING.md) for de
 - [mlx-embeddings](https://github.com/Blaizzy/mlx-embeddings) - Embedding model support for Apple Silicon
 - [dflash-mlx](https://github.com/bstnxbt/dflash-mlx) - Block diffusion speculative decoding on Apple Silicon
 - [MTPLX](https://github.com/youssofal/mtplx) - Lightning MTP's verify-shape Metal kernels are powered by MTPLX by Youssof Altoukhi, which also inspired the depth-k pipeline
-- [mlx-serve](https://github.com/ddalcu/mlx-serve) - The fused GDN verify prework kernel is adapted from mlx-serve's port of the mlxfast-challenge qwen35_packed_gdn_prework kernel
+- [mlx-serve](https://github.com/ddalcu/mlx-serve) - The fused GDN verify prework kernel is adapted from mlx-serve's port of the mlxfast-challenge qwen35_packed_gdn_prework kernel; Qwen4 QSA's 128-bit K/V staging is adapted from mlx-serve's MIT-licensed `msv_attn_p256` kernel
 - [SiliconScope](https://github.com/kennss/SiliconScope) - The menu bar statistics take their design and rendering approach from SiliconScope by Kennt Kim, which also inspired the energy-efficient re-render gating
--- a/benchmarks/bench_qwen4_qsa_sparse_gqa.py
+++ b/benchmarks/bench_qwen4_qsa_sparse_gqa.py
@@ -1,0 +1,146 @@
+#!/usr/bin/env python3
+"""Benchmark Qwen4 direct-index sparse GQA against the gathered MLX path."""
+
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import argparse
+import statistics
+import time
+
+import mlx.core as mx
+import numpy as np
+
+from omlx.custom_kernels.glm_moe_dsa import fast
+from omlx.patches import mlx_vlm_qwen4_exp_compat as compat
+
+
+compat.apply_mlx_vlm_qwen4_exp_compat_patch()
+from mlx_vlm.models.qwen4_exp import qsa_fast  # noqa: E402
+
+
+def _time(call, repetitions: int):
+    samples = []
+    output = None
+    for _ in range(repetitions):
+        start = time.perf_counter()
+        output = call()
+        mx.eval(output)
+        mx.synchronize()
+        samples.append((time.perf_counter() - start) * 1000.0)
+    return samples, output
+
+
+def _portable(queries, keys, values, selected, selected_valid):
+    query_tokens = queries.shape[2]
+    selected_keys = qsa_fast._batch_gather_tokens(
+        keys.transpose(0, 2, 1, 3), selected
+    ).transpose(0, 1, 3, 2, 4)
+    selected_values = qsa_fast._batch_gather_tokens(
+        values.transpose(0, 2, 1, 3), selected
+    ).transpose(0, 1, 3, 2, 4)
+    grouped_queries = queries.transpose(0, 2, 1, 3).reshape(
+        1, query_tokens, 2, 12, 256
+    )
+    scores = (
+        grouped_queries.astype(mx.float32)
+        @ selected_keys.astype(mx.float32).swapaxes(-1, -2)
+    ) / (256**0.5)
+    scores = mx.where(
+        selected_valid[:, :, None, None, :],
+        scores,
+        mx.finfo(scores.dtype).min,
+    )
+    probabilities = mx.softmax(scores, axis=-1).astype(queries.dtype)
+    return (probabilities @ selected_values).reshape(
+        1, query_tokens, 24, 256
+    )
+
+
+def main():
+    parser = argparse.ArgumentParser()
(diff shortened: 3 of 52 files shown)


jundot:
Thanks for the work on this, especially the custom QSA and decode kernels. Writing these paths separately while keeping the implementation exact and fail-closed took a lot of care, and it produces a clear speedup on my machine.

Local protocol: `Qwen3.8-Flash-Next-oQ4e-mtp`, Code (Python) context, 128 generated tokens, temperature 1.0, top-p 1.0, and Lightning MTP with adaptive maximum depth 3. Each length was measured once with zero cached prompt tokens.

| Context | main PP | PR PP | Change | main TG | PR TG | Change | main total | PR total | Change |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 4K | 999.61 tok/s | 1061.34 tok/s | +6.2% | 48.65 tok/s | 53.59 tok/s | +10.2% | 6.73s | 6.25s | -7.1% |
| 16K | 950.06 tok/s | 1061.31 tok/s | +11.7% | 48.74 tok/s | 55.70 tok/s | +14.3% | 19.87s | 17.74s | -10.7% |
| 32K | 834.31 tok/s | 1113.63 tok/s | +33.5% | 40.08 tok/s | 45.92 tok/s | +14.6% | 42.47s | 32.21s | -24.2% |

I also checked the native/fallback paths, boundary snapshots and SSD prefix-cache behavior, and three concurrent requests staggered at one-second intervals. I did not find cache corruption or an output mismatch. One small thing I'll fold into a follow-up is gating the Qwen4 4096-token floor on both the native extension and the `qwen4_qsa_sparse_gqa_attention` symbol, along with duplicate SPDX and comment cleanup.

LGTM. Merging this.

humanrouter:
Independent confirmation of this kernel on a different M3 Ultra, ported into MTPLX rather than run in-process with oMLX.

**Hardware:** Mac Studio, M3 Ultra 80-core, 256 GB unified (not the 512 GB box in the PR).
**Model:** `Youssofal/Qwen3.8-Flash-Next-MTPLX-Optimized-Speed` (MTPLX 6-bit pack), served by MTPLX.
**What we took:** only the Qwen4 sparse-GQA Steel attention consumer (`qwen4_qsa_sparse_gqa_*`, rev `dc312e6e905e03d21ef0c4a86289cbfa2cf857cc`). Not the native indexer scores or fp32 top-k from this PR.
**Baseline:** MTPLX PR #391 restack, same weights, same sampler (T=1.0, top_p=0.95, top_k=20, thinking off). Sequential A/B, one copy in RAM. Engagement `direct_kernel=1911` / `metal_selector=1911` / gather_tier=0.

| Context | TTFT before | TTFT after | Prefill tok/s before | after |
|---|---|---|---|---|
| 16k | 17.79 s | 16.96 s (−4.7%) | 832 | 873 |
| 32k | 40.69 s | 34.32 s (**−15.6%**) | 726 | 873 |

Decode stayed flat (99.0 → 100.3 tok/s). Numeric vs MTPLX's gather oracle: bf16 max abs err 0.00146, fp16 0.00018 (your 5e-3 bound). Greedy tokens through a production-geometry Attention layer were identical across direct / gather / dense.

Smaller than this PR's full-stack +33.5% at 32k, which matches taking only the attention consumer.

One compile note: building this kernel as a **standalone** metallib (not inside `omlx_glm_kernels`) required defining `Qwen4QSASparseGQAParams` in the `.metal` file. The C++ translation unit has the struct; Metal does not see it. `qwen4_qsa_sparse_gqa.metal` on main still includes the steel header without that definition. Your bundled glm_moe_dsa build may already inject it another way; a standalone compile does not.

MTPLX port: https://github.com/youssofal/MTPLX/pull/423

Issue #3233, opened 2026-08-28
0.6.3: Qwen ANE prefill corrupts context under agentic load (jumbled prompt content, empty init-only completions); clean with ANE off, Lightning MTP on


daveycodez:
## Environment

- oMLX **0.6.3** (final, updated same day from an 0.6.3rc)
- Mac Studio M3 Ultra, 512 GB
- Model: `Jundot/Qwen3.8-27B-oQ4e-fp16-mtp` (discovered as `type: vlm`, served by VLMBatchedEngine)
- Per-model settings: Lightning MTP on (`mtp_enabled: true`, adaptive depth), **Qwen ANE prefill on** (tuner-calibrated: MLP 45%, GDN 37.5%, dual-ANE, CPU share), `qwen3-r-code` profile (temp 0.6, top_p 0.95, thinking on), TurboQuant/DFlash/SpecPrefill off
- Workload: agentic coding harness against the Anthropic-compatible endpoint — multi-step tool-calling turns, 10–30K-token prefills per step, prompt-cache reuse 50–90%, session contexts 90K–350K

## Summary

With **ANE prefill + Lightning MTP both enabled**, agentic sessions reliably derail within 5–14 steps (two distinct symptoms below; reproduced on both warm and cold prompt cache, including immediately after deleting the SSD cache).

**Isolation so far: ANE prefill OFF, Lightning MTP still ON → a comparable session (8 steps, ~209K cumulative input, 75% cache hit) ran to completion with a correct long-form final answer at ~41 tok/s.** So the corruption tracks the `qwen35_ane_prefill` path, not MTP itself. (n=1 clean run so far; will add more runs as I have them.)

## Symptom 1: model receives corrupted prompt content mid-session

Deep into a session (step 14, ~350K cumulative input, ~85% cache hit), the model began reasoning about tool output it was never sent: its visible reasoning quoted short repeated fragments (a numeric token `262144` = 2^18, plus one filename repeated several times with stray line-number-like values) that do not appear in any tool result recorded in the harness transcript — the fragments look like re-spliced tokens from much earlier context. The model itself remarked the output "looks jumbled" and attempted to re-read the file; that recovery tool call was emitted with broken framing (literal `</parameter> </function> </tool_call>` closers leaked into the text channel, no parseable call), then `stopReason: stop`. All 13 prior steps had perfectly-formed tool calls.

## Symptom 2: empty completion (EOS as first sampled token)

[...]

PR #3246, opened 2026-08-28
Honor model-owned to_batch when joining a running batch (fixes qwen4_exp continuous-batching crash)

### LINKED ISSUE #3245: qwen4_exp continuous batching crashes on batch join: 'QSAKVCache' object has no attribute 'extend' (scheduler misses model-owned to_batch; one-line fix)
## Summary

Continuous batching on `qwen4_exp` (Qwen3.8-Flash-Next) crashes as soon as a request joins a running batch:

```
Cache corruption not recoverable after retries: 'QSAKVCache' object has no attribute 'extend'
```

Reproduced on 0.6.3 final (M3 Ultra, 512 GB) with the built-in benchmark at `Batch 4x` (pp1024/tg128), and with any staggered concurrent API requests. Requests that prefill together are fine; it's specifically the continuous-batching *join* that dies, after which the scheduler's corruption recovery loops and fails the requests.

## Root cause

`scheduler._to_batched_cache_layer` converts known singleton cache types to batch form before `_extend_cache_layer` calls `cache_a.extend(cache_b)` — it handles mlx-lm `KVCache`/`RotatingKVCache` (via `merge([self])`) and the TurboQuant singleton type, but **everything else passes through unconverted**.

`QSAKVCache` (qwen4_exp's sparse-attention cache) declares its batch conversion the model-owned way: a `to_batch(left_padding)` method returning a `BatchQSAKVCache` — which *does* implement `extend`. `_patched_make_cache` explicitly honors `to_batch` at initial cache creation ("Honor model-owned batch conversion before MLX-LM's fallbacks"), but the join path never checks for it, so a warm singleton `QSAKVCache` reaches `.extend()` and raises.

## Fix (verified locally)

Mirror the model-owned conversion in `_to_batched_cache_layer` — after the existing type checks:

```python
    to_batch = getattr(cache_obj, "to_batch", None)
    if callable(to_batch):
        return to_batch([0])
    return cache_obj
```

(A warm singleton is one unpadded row, so `left_padding=[0]`.) With this applied: 4 staggered concurrent requests to Flash Next all complete, joins included; no corruption-recovery events in the log. qwen3_5 models are unaffected (their caches hit the earlier branches). This should also cover other model-owned caches that expose `to_batch` (e.g. the minimax_m3 compat cache) if they hit the same join path.

## Secondary observation (same model, single request)

In the built-in benchmark, single-request tg TPS is unstable across context lengths — 39.9 / 37.0 / 47.6 / **16.5** / 30.4 / 38.0 tok/s for pp1024→pp65536 (TPOT spiked to 90.9 ms at pp16384). The 16.5 row is *below* the no-MTP baseline (~22), suggesting the adaptive draft-depth controller can get stuck paying verify costs at low acceptance instead of parking. Not a crash — noting it since it's the same qwen4_exp MTP path. Happy to split this into its own issue if you prefer.

Related from the same setup: #3233, #3242 (PR #3243).

daveycodez:
## Bug

A request that joins an already-running Qwen3.8-Flash-Next batch crashes the scheduler with:

```
AttributeError: 'QSAKVCache' object has no attribute 'extend'
```

Reported in #3245.

## Mechanism

On the continuous-batching join path, `_extend_cache_layer` converts each singleton cache layer to its batch form via `_to_batched_cache_layer` before calling `cache_a.extend(cache_b)`. That converter only knew two shapes:

- mlx-lm `KVCache` / `RotatingKVCache`, via `merge([self])`
- the TurboQuant singleton cache type, via `merge([self])`

Any other cache passed through unconverted. `qwen4_exp`'s `QSAKVCache` is a model-owned singleton that declares its batch conversion through a `to_batch(left_padding)` method (returning `BatchQSAKVCache`, which does have `extend`). `_patched_make_cache` in that file already honors this protocol at cache creation — but the join path did not, so the singleton reached `extend()` unconverted and crashed.

## Fix

In `_to_batched_cache_layer`, after the existing type-specific branches, honor a callable `to_batch` on the cache object, converting the warm singleton as a single unpadded row (`to_batch([0])`). Existing branches are untouched; caches without `to_batch` behave exactly as before.

## Verification

Reproduced with the built-in benchmark at Batch 4x against Qwen3.8-Flash-Next-oQ4e-mtp: the join crash hits as soon as a request joins the running batch. With the fix, 4 staggered concurrent requests to the same model all complete.

Complements #3243 (a separate qwen4_exp MTP fix) but is independent of it — either can land without the other.

Fixes #3245

## Follow-up commits

Testing the `to_batch` fix surfaced two deeper crashes on the same path; both are fixed here as follow-up commits:

- **QSA indexer: aligned-column scalar `past_len` for ragged batches** — `QSAIndexer.from_projected` treated `cache.offset` as a scalar, but batched caches return per-row `mx.array` offsets; with ragged rows the `(batch,)` array broadcasts into the seq axis and crashes at `mx.concatenate` (shapes like `(2,2,13624)` vs `(2,1,3)`). Since the indexer cache is left-padded to one width, the aligned scalar is `key_len - seq_len`.
- **Skip singleton MTP fold on mid-cycle joins** — when a request joins mid-MTP-cycle, the singleton fold receives a batch-shaped trunk hidden and crashes in `fuse_inputs`, forcing full cache-recovery re-prefills. Skipping the fold (same idiom as the existing depth-0 branch) lets the late-join handoff proceed cleanly.
- **Slice indexer arrays in `BatchQSAKVCache.trim`** — this eliminates the residual post-join corruption/recovery cycle entirely: the repro that previously logged two corruption events per round now logs zero across repeated rounds, with join latencies dropping from up to ~13s to ~2s.

With all three fixes, mixed-length continuous batching is functional end-to-end for `qwen4_exp`: a 14K-token request decoding 500 tokens completes while 4 short requests join mid-flight and finish (verified on an M3 Ultra against Qwen3.8-Flash-Next-oQ4e-mtp with Lightning MTP enabled).

### DIFF
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -2309,6 +2309,15 @@ def _chain_next_drafts(
     """
     import mlx.core as mx
 
+    if getattr(hidden_rows, "ndim", 0) == 3 and hidden_rows.shape[0] != 1:
+        # A row joined mid-cycle (continuous batching): the singleton head
+        # cache cannot fold a multi-row trunk hidden. Skip this fold — the
+        # late-join handoff/park path rebuilds MTP state afterwards.
+        state.drafts = mx.zeros((0,), dtype=mx.uint32)
+        state.draft_lps = []
+        state.draft_accept_lps = []
+        return
+
     model = gen_batch.model
     if _dspark_host(model) is not None:
         return _dspark_next_drafts(
--- a/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
+++ b/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
@@ -711,6 +711,11 @@ class Qwen4ExpQSAIndexer(nn.Module):
             full_position_ids = position_ids
 
         key_len = raw_keys.shape[1]
+        if isinstance(past_len, mx.array):
+            # Batched rows carry per-row KV offsets, but the indexer cache is
+            # left-padded to one width; the masks below need aligned-column
+            # scalars or the (batch,) offsets broadcast into the seq axis.
+            past_len = key_len - seq_len
         max_complete_blocks = key_len // self.compress_ratio
         if max_complete_blocks <= self.block_topk:
             return None
--- a/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
+++ b/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
@@ -475,6 +475,15 @@ class BatchQSAKVCache:
     def trim(self, n):
         trimmed = self.kv_cache.trim(n)
         self.index_offset = max(0, self.index_offset - trimmed)
+        # Slice the physical arrays like the singleton trim does:
+        # update_indexer concatenates onto them and re-derives index_offset
+        # from shape[1], so stale draft columns would otherwise fossilize
+        # and desync the indexer from the KV by the trimmed amount.
+        if trimmed and self.index_keys is not None:
+            self.index_keys = self.index_keys[:, : self.index_offset]
+            self.index_position_ids = self.index_position_ids[
+                ..., : self.index_offset
+            ]
         return trimmed
 
     @property
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -992,6 +992,13 @@ def _to_batched_cache_layer(cache_obj: Any) -> Any:
         and type(cache_obj) is _TQ_SINGLETON_CACHE_TYPE
     ):
         return cache_obj.merge([cache_obj])
+    # Model-owned singletons (e.g. qwen4_exp QSAKVCache) declare their batch
+    # conversion via to_batch, which _patched_make_cache honors at creation;
+    # honor it on the continuous-batching join path too, or extend() hits a
+    # singleton without the method. A warm singleton is one unpadded row.
+    to_batch = getattr(cache_obj, "to_batch", None)
+    if callable(to_batch):
+        return to_batch([0])
     return cache_obj
 
 
--- a/tests/test_mlx_lm_mtp_patch.py
+++ b/tests/test_mlx_lm_mtp_patch.py
@@ -2821,3 +2821,33 @@ class TestLoopTaxHygiene:
         assert tracker.recently_active(3.0)  # just-finished counts
         tracker.clear()
         assert not tracker.recently_active(3.0)
+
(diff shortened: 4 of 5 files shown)


DiscoStew6082:
I validated this at exact head `de9db03c` on a 512 GiB M3 Ultra.

- Four private regression probes reproduced each failure on current `main` and passed on this PR.
- Full non-slow, non-integration suite: 10,467 passed, 106 skipped, 77 deselected.
- The original unequal-prompt reproduction from #3215 passed with output hashes matching the retained successful run.
- Two repeated BF16 Lightning-MTP serving runs passed. Each used a 14,007-token request with four requests joining during active decoding. All side responses were exact, both 500-token long outputs matched, and post-join MTP draft acceptance was 99.1%.
- No cache-corruption errors, broadcast-shape failures, stale-index symptoms, fallbacks, tracebacks, request errors, or crashes occurred.

ANE and oMLX prefix caching were disabled, so this does not validate or resolve #3233.

The behavior looks solid. The remaining gap I see is that the four fixes do not yet have committed regression tests. #3215 contains a focused ragged-QSA regression that could be ported or adapted here.

daveycodez:
Thanks @DiscoStew6082 for the end-to-end validation. Committed regression tests are now on the branch (953e5886) covering all four fixes:

- tests/test_mlx_vlm_qwen4_exp_compat.py
  - test_qwen4_cache_extension_promotes_singletons_to_model_owned_batch — the to_batch promotion on the continuous-batching join path (adapted from your ragged-QSA regression in #3215, credited in the commit message)
  - test_qwen4_qsa_indexer_handles_ragged_batch_offsets — from_projected with per-row array offsets produces (batch, 1, seq, key_len) selections instead of crashing mx.concatenate
  - test_qwen4_batch_qsa_trim_slices_indexer_arrays — trim slices index_keys/index_position_ids so a later update_indexer cannot resync to the stale width
- tests/test_mlx_lm_mtp_patch.py
  - TestChainNextDraftsLateJoin::test_multi_row_trunk_hidden_skips_singleton_fold — the multi-row fold guard returns empty drafts without touching model.mtp_forward

All four fail on main without the fixes and pass on this branch (verified by checking out the unfixed sources under the new tests).

On overlap: commit 1 here and #3215 address the same _to_batched_cache_layer seam, so whichever the maintainer prefers can land first and the other rebases — the tests remain valid either way.

DiscoStew6082:
I am perfectly willing to close my PR as superceded if needed

jundot:
Thanks for the fixes and regression coverage. I validated the core late-join path against current `main`, including a real Lightning MTP run with four requests joining a 23K-token decode; all requests completed without cache or shape errors.

The branch now has a simple additive conflict in `tests/test_mlx_vlm_qwen4_exp_compat.py`. I’ll resolve it directly by retaining both test blocks, then merge the PR. As a non-blocking follow-up, I’ll remove the multi-row fold guard from `4d28ad6c` and its test because the current call paths keep `_chain_next_drafts()` singleton-only and hand off late joins before batch extension.

PR #3249, opened 2026-08-28
feat(cache): expose hybrid boundary diagnostics

MDX-Tom:
## Why this matters

Hybrid models need a boundary-aligned snapshot before non-sliceable recurrent state can be persisted safely. oMLX already fails closed when that snapshot is missing, but several decision points were visible only at DEBUG or returned None without a stable reason: speculative offset skew, missing batch UIDs, extraction failures, SSD staging fallback, snapshot restore failure, and final store skip.

That made an expensive warm-turn re-prefill hard to diagnose from the runtime API. Existing aggregate hit/miss counters could show that reuse was low, but not whether the prompt diverged, the boundary capture was skipped, SSD staging fell back to memory, or the final cache store was intentionally rejected.

This PR adds content-free, model-scoped diagnostics to the existing runtime cache observability payload. It does not change boundary selection, cache serialization, cache reuse, or token generation.

### Hybrid boundary snapshot lifecycle

- Add a thread-safe BoundarySnapshotDiagnostics tracker for capture attempts and successes, SSD-versus-memory captures, SSD fallbacks, boundary override attempts/hits/misses, and fail-closed store skips.
- Record stable reason identifiers for silent branches such as cache_offset_mismatch, uid_not_found, extract_error, ssd_save_failed, no_snapshots, no_aligned_snapshots, and ssd_load_failed.
- Preserve one structured last_event snapshot with request ID, token count, block size, source (prefill/decode), storage path, and available-boundary count where applicable.
- Promote the final BoundaryStoreUnavailable skip from DEBUG to one content-free INFO line with its reason and numeric boundary context.

### Prefix reuse diagnostics

- Publish the latest prefix lookup summary with prompt tokens, restored KV tokens, re-prefill tokens, block size, matched blocks, closest recent stored sequence, common-prefix tokens, and common-prefix tokens that remained unreused.
- Keep prompt/token text out of the runtime payload. Existing decoded divergence context remains DEBUG-only.

### Admin observability

- Add boundary_snapshots and last_prefix_lookup to each loaded model row returned by the existing runtime cache observability builder.
- Keep the fields optional so DFlash and engines that do not expose scheduler diagnostics retain their current payload shape.

## Correctness and fail-closed behavior

- Boundary capture, extraction, SSD staging, override selection, and final store behavior are unchanged; the new code records the branch that was already taken.
- Positional skew still rejects the snapshot rather than persisting mismatched recurrent state.
- A missing usable boundary snapshot still skips the final store rather than falling back to live non-sliceable state.
- Admin reads receive copies under a lock and never iterate scheduler-owned mutable diagnostic state.
- No prompt text, token IDs, cache tensors, filesystem paths, or decoded content are added to INFO logs or runtime diagnostics.

## Validation

- 243 passed across tests/test_cache_observability.py, tests/test_runtime_cache_observability.py, and the complete tests/test_scheduler.py.
- Targeted boundary tests cover successful capture, speculative offset skew, fail-closed store skip with INFO diagnostics, and prefix re-prefill accounting.
- python -m py_compile passed for all changed source and test modules.
- git diff --check passed.

## Deliberately out of scope

- This PR does not change cache block size, prefill chunk sizing, MTP alignment, SSD snapshot encoding, prefix matching, or eviction policy.
- Dashboard rendering is unchanged; the structured fields are available to the existing runtime API and external admin clients.
- Responses/MCP/web tool routing is unrelated and not included.

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -5053,6 +5053,12 @@ def _build_runtime_cache_observability(
         gdn_last_restore = prefix_stats.get("gdn_last_restore")
         if not isinstance(gdn_last_restore, dict):
             gdn_last_restore = None
+        boundary_snapshots = runtime_stats.get("boundary_snapshots")
+        if not isinstance(boundary_snapshots, dict):
+            boundary_snapshots = None
+        last_prefix_lookup = runtime_stats.get("last_prefix_lookup")
+        if not isinstance(last_prefix_lookup, dict):
+            last_prefix_lookup = None
 
         # Keep the cache fields at the model-row level so the dashboard and
         # external admin clients can inspect them without knowing scheduler's
@@ -5127,6 +5133,11 @@ def _build_runtime_cache_observability(
             "gdn_last_restore": gdn_last_restore,
             "gdn_staging": gdn_staging_payload,
         }
+
+        if boundary_snapshots is not None:
+            model_payload["boundary_snapshots"] = boundary_snapshots
+        if last_prefix_lookup is not None:
+            model_payload["last_prefix_lookup"] = last_prefix_lookup
 
         for field in ssd_counter_fields:
             if field in ssd_stats:
--- a/omlx/cache/observability.py
+++ b/omlx/cache/observability.py
@@ -76,6 +76,89 @@ class CacheRateTracker:
             self._snapshots.clear()
 
 
+class BoundarySnapshotDiagnostics:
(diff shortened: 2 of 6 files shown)


jundot:
Thanks for adding content-free diagnostics for boundary snapshots and prefix reuse. The tracker records the existing fail-closed decisions without changing cache selection or serialization, and the admin path reads copied snapshots rather than live scheduler state. It also applies cleanly after the QSA integration. LGTM, merging.

PR #3251, opened 2026-08-28
perf(glm5): flatten native prefill and preserve exact decode scores

jonathan308:
## Summary

This PR speeds up the native `glm5_next` path without changing the model's selected attention rows, recurrent state, logits, or generated output.

It contains four bounded changes:

1. An explicit wide-prefill mode reuses the existing `MLX_LM_GLM_DSA_ADAPTIVE_PREFILL_STEP=1` switch for 8K chunks, but only when the loaded model has the published GLM-5.3 geometry and the complete native sparse-kernel set is present.
2. NoPE sparse MLA gets a strict `D_PE=0` ABI instead of carrying an all-zero 64-wide positional lane through attention.
3. The vector-gated recurrent kernel processes four value rows per thread while preserving GLM's safe lower-bound gate, masking behavior, and FP32 cache state.
4. Single-row decode uses an M=1 specialization of the existing Steel MMA index-score kernel, preserving the deployed row-0 arithmetic and reduction order exactly while avoiding 63 padded query rows.

Wide prefill is deliberately opt-in. With no environment override, GLM-5.3 keeps the existing 2K prefill/cache block and its finer prefix-reuse granularity. Enabling 8K mode also makes the recurrent snapshot/cache block 8K; that throughput-versus-cache-granularity tradeoff is explicit rather than silently changing the default.

## Physical result

Clean-branch qualification used:

- M3 Ultra, 60 GPU cores, 256 GB
- `Jundot/GLM-5.3-Flash-oQ4e` (strict oQ4e, about 169 GiB resident on this branch)
- Python 3.13.13, oMLX's pinned MLX 0.32.0 / mlx-vlm 0.6.3 / mlx-lm 0.31.3
- single stream; MTP, DFlash, TurboQuant KV, and ANE off
- explicit 8K mode enabled
- one discarded compiler-warmup request; every reported row had a unique first block and `cached_tokens=0`

| Prompt | Output | Prefill | Decode |
|---:|---:|---:|---:|
| 10K | 500 | 345.5 tok/s | 26.2 tok/s |
| 20K | 1 | 359.9 tok/s | — |
| 50K | 1 | 365.8 tok/s | — |
| 100K | 1 | 364.0 tok/s | — |

The 20K-to-100K spread is 1.6%; the curve does not taper with context in this test. These are deterministic repeated-token prompts with a unique prefix, intentionally used to isolate chunk/context scaling. They are not presented as a universal MoE workload score.

On the same integration runtime before/after the retained kernels:

- the dedicated NoPE path improved a cold 50K prefill from 333.99 to 344.67 tok/s (+3.2%);
- R=4 row blocking then improved the immediate 20K control from 354.3 to 372.5 tok/s (+5.1%);
- sustained 10K + 500 decode improved from 25.3 to 26.5 tok/s (+4.7%).

Primitive measurements explain the changes:

- R=4 recurrence, production B1/H64/D128: 11.387 -> 7.288 ms at 2K and 56.766 -> 18.062 ms at 8K.
- Exact M=1 index scores versus the padded kernel: 1.38x at pooled K=512, 1.47x at 2K, 1.85x at 8K, 2.36x at 16K, and 3.35x at 32K.

## Correctness gates

- NoPE output is bitwise equal to the previous zero-PE native route.
- R=4 output and FP32 recurrent cache are array-equal to R=1, including masked-token and lower-bound-gate cases.
- M=1 score sheets and ordered top-k arrays matched the padded Steel path in 900/900 cases across BF16/FP16, B1/B2, and pooled lengths 512..16,383.
- A deterministic physical 16K + 128-token control retained the same output SHA-256 through every promoted build.
- Older packaged extensions safely retain the exact historical M=1/BM64 or padded fallback; unsupported model geometry never enables wide prefill.
- Physical smokes passed prefix reuse with identical output, tool-call creation and tool-result continuation, image input, in-flight cancellation, and a post-cancel health request.

## Validation

- Clean native build succeeded from this branch against MLX 0.32.0 and nanobind 2.13.0.
- 286/286 affected GLM/native/scheduler tests passed after the final geometry and opt-in gates.
- Full official-pin suite: **10,561 passed, 30 skipped, 77 deselected**.
- `git diff --check` and Python compilation pass.
- All four commits are SSH-signed.

## Scope

This does not include Fusion's load-time compile warmup, Qwen work, MTP, clustering, updater/release changes, or the separately identified SSD admin-clear race. Those remain separate review units.

### DIFF
--- a/omlx/custom_kernels/glm_moe_dsa/csrc/bindings.cpp
+++ b/omlx/custom_kernels/glm_moe_dsa/csrc/bindings.cpp
@@ -89,6 +89,23 @@ NB_MODULE(_ext, m) {
       "topk_length"_a = nb::none(),
       "causal_prefix_rows"_a = 0,
       "stream"_a = nb::none());
+  // Separate ABI name lets Python fail closed when code is paired with an
+  // older extension that only accepts the historical D_PE=64 geometry.
+  m.def(
+      "glm_dsa_nope_sparse_mla_attention",
+      &omlx::glm_kernels::glm_dsa_nope_sparse_mla_attention,
+      "q_latent"_a,
+      "q_pe"_a,
+      "kv_latent"_a,
+      "k_pe"_a,
+      "topk_indices"_a,
+      "scale"_a,
+      "causal"_a = true,
+      "topk_valid_prefix"_a = false,
+      "causal_prefix_indices"_a = false,
+      "topk_length"_a = nb::none(),
+      "causal_prefix_rows"_a = 0,
+      "stream"_a = nb::none());
   m.def(
       "glm_dsa_exact_block_attention",
       &omlx::glm_kernels::glm_dsa_exact_block_attention,
--- a/omlx/custom_kernels/glm_moe_dsa/csrc/dsa_indexer.cpp
+++ b/omlx/custom_kernels/glm_moe_dsa/csrc/dsa_indexer.cpp
@@ -132,7 +132,6 @@ class DSAIndexerScoresPrimitive : public Primitive {
 
     out.set_data(allocator::malloc(out.nbytes()));
 
-    constexpr int bm = 64;
     constexpr int bk = 16;
 
     const int B = q.shape(0);
@@ -148,9 +147,11 @@ class DSAIndexerScoresPrimitive : public Primitive {
     // slower at P=125k and ~11% slower at P=2.5k — the kernel is
     // compute/barrier-bound (Q and pooled-K panels are largely
     // L2-resident), so the traffic reduction does not pay. bn=64/wm2/wn2
-    // is the fixed configuration.
+    // remains the prefill configuration; an M=1 decode row uses the exact
+    // BM8/wm1 specialization from the same Steel kernel.
+    const int bm = M == 1 ? 8 : 64;
     const int bn = 64;
-    const int wm = 2;
+    const int wm = M == 1 ? 1 : 2;
     const int wn = 2;
 
     const int tiles_m = (M + bm - 1) / bm;
--- a/omlx/custom_kernels/glm_moe_dsa/csrc/dsa_indexer.metal
+++ b/omlx/custom_kernels/glm_moe_dsa/csrc/dsa_indexer.metal
@@ -45,6 +45,13 @@ struct OMLXDSATopKParams {
 instantiate_dsa_indexer_score(float16, half, 64, 64, 16, 2, 2);
 instantiate_dsa_indexer_score(bfloat16, bfloat16_t, 64, 64, 16, 2, 2);
 
+// Decode has one query row.  Keep the exact Steel MMA arithmetic used by the
+// historical BM64 score kernel, but reduce the minimum M tile to one 8-row
+// simdgroup fragment.  BK=16 and the simdgroup matmul sequence are unchanged,
+// so retained row 0 is bit-identical; the other seven fragment rows are safe
+// loaded as zero and never stored.
+instantiate_dsa_indexer_score(float16, half, 8, 64, 16, 1, 2);
+instantiate_dsa_indexer_score(bfloat16, bfloat16_t, 8, 64, 16, 1, 2);
 instantiate_dsa_topk_indices(float16, half, 2048, 1024);
 instantiate_dsa_topk_indices(bfloat16, bfloat16_t, 2048, 1024);
 instantiate_dsa_topk_indices(float16, half, 512, 1024);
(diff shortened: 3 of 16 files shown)


davidtorcivia:
Independent corroboration on a second machine, plus one rebase note.

I cherry-picked the four commits here onto v0.6.4 to run them locally. Three
apply clean. `perf(glm5): preserve exact single-row index scores` conflicts in
`omlx/custom_kernels/glm_moe_dsa/csrc/dsa_indexer.metal`, where 0.6.4's new
`instantiate_qwen4_qsa_indexer_score` lines landed in the same block as the BM8
`instantiate_dsa_indexer_score` lines this PR adds. It is purely additive on
both sides and resolves by keeping both.

Environment: M3 Ultra, 80 GPU cores, 256 GB, oMLX 0.6.4 wheel, MLX 0.32.0,
`GLM-5.3-Flash-oQ4e-mtp` variant. Command Line Tools only, so no native `_ext`
build. That makes this a portability check rather than a throughput
reproduction.

What I can confirm:

- `tests/test_glm5_gated_delta_row_block.py`, `tests/test_glm_moe_dsa_patch.py`
  and `tests/test_mlx_vlm_glm5_next_compat.py` give 49 passed, 18 skipped, with
  the skips being exactly the native-symbol gates. `tests/test_scheduler.py`
  passes 242 with the prefill-alignment change applied.
- The Python half has been serving continuously on 0.6.4 with no errors. The
  JIT'd R=4 gated-delta kernel and the M=1 index-score route are both live.
- The fail-closed paths behave as described on a box with no native build.
  `sparse_mla_attention_nope` returns None and the historical zero-PE route
  runs, with no output or stability change.
- The wide-prefill gate is safe in the negative case. With
  `MLX_LM_GLM_DSA_ADAPTIVE_PREFILL_STEP=1` set in the launch environment and
  the native sparse kernels absent, the path correctly never enables: no
  adaptive-prefill config line is logged and prefill throughput is unchanged
  across 4K, 8K and 16K prompts (three repetitions each). The geometry and
  symbol checks do what the description says they do.

One data point on the size of the target. Profiling decode on the served path
by capturing each module's real arguments and replaying them, at Kv 4K and a
three-row verify batch, the DSA indexer accounts for 7.6 ms inside 11.8 ms of
sparse attention, against a whole forward of roughly 62 ms. So the indexer
alone is about 12 percent of decode on this configuration, which is consistent
with the M=1 specialization being worth the kernel.

I am deliberately not comparing my decode numbers to the table above. I am on a
different quantization and a different prompt shape, so they are not
like-for-like.

This is the change I would most like to see land for GLM-5.3 on Apple silicon.

davidtorcivia:
Correction to the profiling number in my previous comment, and it moves in
your favour.

The 7.6 ms indexer figure came from a flawed method. I was timing N
independent replays of a module with a single eval at the end, which lets the
replays pipeline on the GPU and measures throughput rather than latency. It
also made components oversum their own parent layer, which is what gave it
away. Separately, my harness was calling the model outside mlx-vlm's
`wired_limit` context, so a sequential pass over the resident weights was
paying residency cost on every forward and running about 55x slow.

Redone with each replay consuming the previous output and the whole thing
inside `wired_limit(model, [generation_stream])`, the harness now reproduces
the served path: at Kv 4K and a three-row verify batch, the composed forward
is 58.7 ms, the chained 45-layer stack is 56.9 ms, and a hand-rolled forward
is 58.6 ms, against roughly 62 ms per forward in the running server.

On that basis, for this 45-layer configuration at the verify shape:

- sparse attention is 1.136 ms per call over 11 layers, so about 12.5 ms, call
  it 20 percent of the forward
- linear (KDA) attention is 0.386 ms per call over 34 layers, about 13.1 ms
- the MoE and dense FFN blocks together are about 15 ms
- attention of both kinds is therefore roughly half of decode, which is more
  than I implied earlier

I did not isolate the indexer inside sparse attention in the corrected run, so
I am not restating a specific indexer share. The honest version is that sparse
attention is about 20 percent of decode here, and the M=1 score specialization
and the NoPE path both land inside it.

Sorry for the noise. The conclusion is unchanged and slightly stronger.

scubamount:
## Verified

- **Claim:** speeds up native `glm5_next` prefill/decode without changing selected attention rows, recurrent state, logits, or output. Four bounded changes, each independently checkable.
- **Strict D_PE=0 ABI:** separate binding name so Python fails closed against an older extension that only accepts the historical D_PE=64 geometry — correct ABI-versioning pattern (same as #3304's `AFFINE_MOE_HAS_Q468` marker).
- **Decode indexer specialization:** M==1 picks BM8/wm1 vs prefill's fixed BM64/wm2; the comment argues bit-identical retained row 0 (same BK=16, same simdgroup sequence; rows 1-7 loaded zero, never stored). This is the load-bearing claim and it is falsifiable via the bitwise-equality test in the suite.
- **Wide-prefill gate:** 8K chunks only when the loaded model matches published GLM-5.3 geometry AND the complete native kernel set is present — fail-closed otherwise.

No issues, provided the bitwise-equality tests are green in CI (they are per the PR's check rollup).

PR #3257, opened 2026-08-28
feat(cluster): harden rank control, teardown, and recovery

jonathan308:
> [!IMPORTANT]
> **Stacked draft (3/4). Depends on #3155 and #3256.** This is opened now so the complete backend series is visible upstream. Until both dependencies merge, GitHub shows their inherited commits because the dependency branches live on the fork. This branch will be restacked after each merge so the final review is the 25-commit incremental rank-control/recovery delta.

Incremental fork review and green CI: [jonathan308/omlx#5](https://github.com/jonathan308/omlx/pull/5).

## Scope

- backend qualification and JACCL-vs-ring selection
- dedicated rank-control channel separated from tensor collectives
- fresh 256-bit launch-scoped control secret; the public plan hash is not authentication
- nonce-based mutual HMAC authentication and authenticated control payloads
- invalid-handshake isolation and explicit worker acknowledgement
- macOS system-Python carrier for direct-link network-policy failures
- JACCL bootstrap side channel and process-level RDMA lease
- rank-ready and serve-release gates
- launcher/peer watchdogs and durable liveness evidence
- verified TERM/KILL teardown with PID/PGID identity checks
- markerless-rank/orphan recovery and fail-closed reboot guidance
- idempotent cluster load/unload lifecycle endpoints

## Deliberately excluded

- request serving/cancellation/telemetry/cache lifecycle (stack item 4)
- Phase split and Cluster v2 UI
- DS4, ANE, MTP, native kernels, and single-Mac optimizations
- updater, release, repository, and branding changes

## Validation

- incremental branch CI: Python 3.11, 3.12, and 3.13 green
- product suite passed; backend-first seam assertions were explicitly scoped and rerun green
- focused control/launch/inference suites passed
- rank-control mutual-auth and tampered-payload tests passed
- Ruff and `git diff --check`: passed

PR #3256, opened 2026-08-28
feat(cluster): add qualification, staging, planning, and deployment

jonathan308:
> [!IMPORTANT]
> **Stacked draft (2/4). Depends on #3155.** This is opened now so the full backend series is visible upstream. Until #3155 merges, GitHub necessarily shows the inherited trust-foundation commits because the dependency branch lives on the fork. After #3155 lands, this branch will be restacked and this PR will collapse to its 16-commit incremental planning/deployment delta.

Incremental fork review and green CI: [jonathan308/omlx#4](https://github.com/jonathan308/omlx/pull/4).

## Scope

- model inventory and manifest exchange
- automatic model synchronization over enrolled SSH or local download
- protected remote rsync paths compatible with macOS system `openrsync`
- per-node model paths and portable cross-account path handling
- verified staging and shard/sidecar inventory
- RAM-proportional pipeline planning
- tensor/pipeline topology recommendations
- signed preview/apply re-planning
- deployment activation and registry migration

## Deliberately excluded

- launcher/rank-control/teardown implementation (stack item 3)
- serving, cancellation, telemetry, and cache lifecycle (stack item 4)
- Phase split and Cluster v2 UI
- DS4, ANE, MTP, native kernels, scheduler/sampling optimizations
- updater, release, repository, and branding changes

## Validation

- incremental branch CI: Python 3.11, 3.12, and 3.13 green
- full local suite: 10,559 passed, 102 skipped
- focused inventory/staging/planning/replan suite: 88 passed after rsync hardening
- Ruff and `git diff --check`: passed

PR #3258, opened 2026-08-28
feat(cluster): make TP and pipeline serving request-safe

jonathan308:
> [!IMPORTANT]
> **Stacked draft (4/4). Depends on #3155, #3256, and #3257.** This is opened now so the full backend series is visible upstream. Until the dependencies merge, GitHub shows their inherited commits because the dependency branches live on the fork. This branch will be restacked after each merge so the final review is the 20-commit incremental serving/cancellation/telemetry/cache delta.

Incremental fork review and green CI: [jonathan308/omlx#6](https://github.com/jonathan308/omlx/pull/6).

## Scope

- base TP and pipeline serving lifecycle
- public request IDs propagated to rank zero
- request-scoped disconnect aborts
- targeted rank cancellation with bounded atomic request sets and acknowledgements
- bounded prefill cancellation at synchronized chunk boundaries
- cancellation fencing before distributed batch mutation
- runtime-failure reconciliation and stale-engine suppression
- per-request and aggregate prefill/decode/cache telemetry
- pipeline prompt-cache agreement and response-token normalization
- persistent rank-local prompt snapshots across restarts
- cluster-wide hot/SSD cache clearing on loaded and unloaded peers
- quiescence gates for unload and cache maintenance

## Deliberately excluded

- Phase split and Cluster v2 UI
- DS4, ANE, MTP, native kernels, and other model/single-Mac optimizations
- updater, release, repository, and branding changes

## Validation

- incremental branch CI: Python 3.11, 3.12, and 3.13 green
- full local suite: 10,632 passed, 102 skipped
- focused serving/cancellation/telemetry/cache suites: green
- Ruff and `git diff --check`: passed

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -14,8 +14,10 @@ import json
 import logging
 import os
 import re
+import shlex
 import shutil
 import signal
+import subprocess
 import sys
 import time
 from collections import deque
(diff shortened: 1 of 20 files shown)


jonathan308:
Following up on the Cluster v2 backend split from #3118 — these slices are high-impact for distributed/RDMA serving. Can we get a status update / review pass when you have bandwidth? Happy to restack or address feedback quickly. @jundot

jonathan308:
Restacked backend 4/4 onto current `main` after #3257 merged. The PR now contains only the 20 incremental serving/cancellation/telemetry/cache commits: 16 files, +2,270/-168. Focused validation passed locally: 197 passed, 3 skipped across distributed cache clearing, launcher, prompt snapshots, telemetry, distributed engine, and SSE keepalive. Hosted Python 3.11–3.13 CI is running on the restacked branch.

jundot:
@jonathan308, thank you for the thoughtful work on this important part of distributed serving. I’ve taken another pass through the PR to consolidate my findings and avoid asking you to make changes over multiple rounds. Please take a look and let me know what you think.

I found several integration pieces present in #3118 but missing from this split:

- **Cancellation:** initialize the cancel state and connect vote completion, batch drain/arm methods, and the required control-channel operations. Watchdog `scope=all` cancellation also needs the synchronized generation path, and cancelled requests need to signal completion to waiting response queues.
- **Cancellation lifecycle:** include `plan_hash` in acknowledgements and restore stale-cancel handling. I reproduced an acknowledged all-request cancel causing a later targeted cancel to retain `scope=all`.
- **Cache maintenance:** register live prompt caches, implement SSD clearing, and connect unloaded local/remote cache clearing using each peer’s own paths. Please adapt the restored code to this split’s dependencies—for example, the original SSD `clear()` calls `flush()`, which is absent here.
- **Request and cache agreement:** forward the transport request ID through `/v1/completions` streaming and connect the prompt-cache agreement helper, including its required control-channel methods.

Please preserve #3257’s authentication improvements and the follow-up fixes already on main when restoring these pieces.

I’ll verify the restored integration paths when the update is ready. If you happen to rerun your two-Mac setup, any observations would be helpful.

Once I’ve verified the update and CI passes, I can proceed with merging.

jonathan308:
@jundot thanks checking what's missing & my tests from the old hardware setup won't be the same waiting on new hardware coming this week . Still can get tests done just different reference hardware this time .

jonathan308:
@jundot I restored the integration paths from your consolidated review in `fd547df8`, while keeping #3257's authenticated rank-control handshake/message framing intact.

- Cancellation now initializes/validates the cancel-vote epoch, finishes shared votes on both control-plane and fallback broadcasts, drains Metal work, performs an authenticated rank barrier, arms the exact epoch/UID set, and only then permits batch removal. Watchdog `scope=all` now stops registered generation contexts and enters that shared generation-loop path rather than mutating rank zero from the heartbeat thread. Cancelled UIDs also put the terminal sentinel into their response queues.
- Cancel files and acknowledgements are plan-scoped and worker-lifetime-scoped. Watchdog epochs advance past an existing future-skewed marker, existing files become startup watermarks, and acknowledged/stale `scope=all` markers are not merged into a later targeted cancel. The reproducer now asserts the later marker remains `scope=requests`.
- Live `LRUPromptCache` instances are registered and cleared; `SSDPromptSnapshotStore.clear()` was adapted to this split's synchronous store (no nonexistent `flush()`); cold local roots are removed when no distributed engine is loaded; and cold peer cleanup resolves `GlobalSettings` on each peer instead of sending the coordinator's `/Users/...` paths over SSH.
- `/v1/completions` forwards the transport request ID into both streaming and non-streaming generation. The existing prompt-cache agreement helper is now invoked after memory/SSD lookup, using newly restored authenticated `broadcast_owned_bytes()` control operations; cancellation uses the restored authenticated `barrier()`.

Validation on the branch: 335 passed, 3 skipped across endpoint integration, control-plane, telemetry/cancellation, prompt snapshots, inference worker, distributed engine, SSE, admin cache clearing, and server tests. A full local run reached 11,298 passing after the endpoint follow-up; the remaining 11 failures are local MLX ABI/version checks (`force_fused` / exact MLX 0.32.2), outside the touched paths. Hosted Python 3.11–3.13 CI is running now. I do not have the original two-Mac setup available for a hardware rerun.

jundot:
@jonathan308, thanks for addressing the review so quickly. I verified the restored cancellation and cache paths, including cancelling one request while another continues generating. The authentication improvements are preserved, and Python 3.11–3.13 CI has passed. LGTM, I’ll merge this.

PR #3260, opened 2026-08-28
feat(moe): SSD Expert Streaming and Soft-REAP for MoE models

onthehub97:
# SSD Expert Streaming and Soft-REAP for MoE models

## Summary

This branch adds SSD Expert Streaming for quantized Mixture-of-Experts models. Routed experts are held in fixed, preallocated MLX banks containing manifest-pinned experts plus a configurable number of evictable hot experts. Experts outside those banks remain in their original safetensor shards and are loaded on demand.

Two residency modes are available:

- **Soft-REAP pinning** keeps the experts selected by an uploaded REAP manifest permanently resident and uses the remaining bank slots as a hot cache.
- **Analytical cache only** pins no routed experts and manages the entire bank from observed router selections.

At a `0%` substitution threshold, the streamed path preserves the model's original expert routing and produces bit-exact expert outputs.

## Runtime and storage

- Adds a row-addressable safetensor index and SSD reader that accesses expert weights, scales, quantization biases, and explicit biases without mapping complete expert tensors into memory.
- Supports stacked and per-expert checkpoint layouts, separate and fused gate/up projections, projection aliases such as `w1`/`w2`/`w3`, and affine, MXFP4, and NVFP4 quantization metadata.
- Preallocates fixed-shape per-layer expert banks so cache promotion and eviction do not change MLX graph shapes or allocate replacement banks.
- Configures cache capacity as an expert count per layer and always reserves at least the model's routed top-k working set.
- Uses expert-major chunking when a routed batch requires more non-pinned experts than the hot tier can hold concurrently.
- Fully populates the user-configured hot tier at model load. Learned hotlist experts are loaded first; unused capacity is filled optimistically with zero-hotness, fully evictable experts so additional RAM immediately provides route coverage without creating implicit pins or narrowing execution.
- Provides checked and speculative execution policies. Speculation remains checked until every layer's hot tier reaches capacity, then keeps resident-hit routing in the MLX graph and transactionally restores KV/SSM cache state before promoting a missed expert and retrying. Readiness is latched, and promoted bank writes are materialized lazily by the retry's first QMM rather than by a redundant eager bank commit.
- Adds optional Qwen-family resident substitution: a resident expert may replace an unloaded expert when its original router weight is within the configured percentage threshold. Selected router weights are retained and renormalized.
- Records runtime statistics for hits, misses, promotions, evictions, fill-gated passes, speculative retries, QMM projection invocations, SSD bytes and operations, SSD I/O time, payload decoding, bank binding, and materialization.

## Model support

- Discovers routed layers under both `mlp` and `ffn`, preserves physical model-layer IDs, and skips dense layers.
- Normalizes expert count and top-k geometry across common router and configuration field names.
- Supports the SwitchGLU layouts used by Qwen 3.5/3.8/4-Exp, DeepSeek V3.2/V4, GLM MoE/GLM5, MiMo-v2, Hy-v3, Laguna, Bailing Hybrid, compatible MiniMax variants, and related models using the same layouts.
- Integrates streamed execution with Qwen target-verification and weighted-sum router paths so MTP and wide verification batches use the cache-aware whole-GLU path.

## Cache management and DwarfStar attribution

The following cache and SSD-I/O optimisations are adapted from [DwarfStar by stefandsl](https://github.com/stefandsl/DwarfStar/tree/80ebbc3):

- Decayed route-frequency scoring with recency as an eviction tie-break.
- Accounting for selected experts on both cache hits and misses.
- Protection of experts used by the current route from eviction.
- Reuse of fixed cache slots and a persistent parallel `pread` worker pool.
- Darwin read-ahead hints for cacheable cold reads.
- Learned per-layer popularity hotlists that warm ordinary evictable slots on later model loads.

Hotlist profiles are compact, checkpoint-fingerprinted, and written atomically to the configured SSD cache directory. Missing, stale, or malformed profiles use optimistic zero-hotness entries to fill the configured capacity. Loading never changes the configured expert count, pins additional experts, or increases the preallocated resident bank size.

## Settings and administration

- Adds **SSD Expert Streaming** to both the web and native Swift model settings with controls for residency mode, manifest upload, hot experts per layer, execution policy, substitution threshold, and projected resident memory.
- Validates uploaded manifests against the checkpoint's routed layer IDs, expert range, tensor layout, duplicate IDs, and a 2 MiB upload limit before storing them by content hash.
- Accepts official REAP layer maps and the wrapped `layers`, `pinned_experts`, and `kept_experts` forms.
- Adds header-only residency estimates for the fixed model state, pinned experts, and hot cache, including the minimum safe cache size.
- Uses the projected streamed residency in engine admission and model-memory-ceiling decisions, including Qwen4 PLE residency handling.
- Includes the new settings in model profiles and load-affecting engine signatures.
- Adds localized UI strings for every existing dashboard locale.

## Documentation and validation

- Adds `docs/soft_reap.md` covering residency modes, manifests, supported layouts, routing behavior, execution policies, memory estimates, and runtime metrics.
- Adds `benchmarks/soft_reap_streaming.py` for checkpoint residency, load, generation, and streaming-stat measurements.
- Adds a bounded synthetic microbenchmark for lazy promotion, fill-gated speculation, cache-capacity scaling, optimistic preload, and prefill expert-group behavior. It exercises the real safetensor reader, quantized expert banks, streaming cache, and speculative executor while checking exact outputs and reporting latency, throughput, retries, SSD traffic, and peak Metal memory without loading a production checkpoint.
- Adds per-trial SSD Expert Streaming telemetry to local throughput benchmarks. Single-request and batch results contain counter deltas for cache hits and misses, evictions, expert loads, SSD traffic and timing, bank update timing, expert-major and QMM calls, resident fill, learned and optimistic preload coverage, and active/peak Metal memory. Logs explicitly warn when an analytical-cache benchmark has incomplete learned coverage so unlike cache states are not compared as equivalent runs.
- Adds synthetic checkpoint tests for manifest validation, cache-only operation, fixed-bank streaming, exact outputs, expert-major execution, route-frequency eviction, speculative rollback/retry, learned hotlist persistence, malformed-profile handling, fused projections, fragmented checkpoints, and cross-family adapters.
- The relevant regression suite passes **372 tests** across expert streaming, throughput benchmarking, DeepSeek V4, GLM MoE/DSA, MiMo-v2, Hy-v3, Laguna, Qwen MoE, and Qwen4 residency coverage.
- The focused native model-settings suite passes **18 tests**, including wire-format and profile-assembly coverage for Soft-REAP and cache-only modes.

### DIFF
--- a/Formula/omlx.rb
+++ b/Formula/omlx.rb
@@ -1,5 +1,5 @@
 class Omlx < Formula
-  CUSTOM_KERNELS = %w[bonsai glm_moe_dsa minimax_m3 qwen35_prefill].freeze
+  CUSTOM_KERNELS = %w[bonsai fast_resource_loading glm_moe_dsa minimax_m3 qwen35_prefill].freeze
 
   desc "LLM inference server optimized for Apple Silicon"
   homepage "https://github.com/jundot/omlx"
--- a/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
+++ b/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
@@ -1014,6 +1014,22 @@ private struct ExpertStreamingSection: View {
                 }
 
                 Row(
+                    label: String(localized: "settings.expert_streaming.fast_resource_loading.label",
+                                  defaultValue: "Fast Resource Loading",
+                                  comment: "Row label for Metal Fast Resource Loading in SSD expert streaming"),
+                    sublabel: String(localized: "settings.expert_streaming.fast_resource_loading.sub",
+                                     defaultValue: "Use Apple Metal I/O for one-shot cold expert batches. Improves prefill without changing hot-cache promotions.",
+                                     comment: "Sublabel describing scratch-only Fast Resource Loading for SSD experts")
+                ) {
+                    Toggle(
+                        "",
+                        isOn: vm.bindProfile($vm.expertStreamingFastResourceLoading)
+                    )
+                    .labelsHidden()
+                    .toggleStyle(.switch)
+                }
+
+                Row(
                     label: String(localized: "settings.expert_streaming.threshold.label",
                                   defaultValue: "Substitution Threshold",
                                   comment: "Row label for SSD expert substitution threshold"),
--- a/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
+++ b/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
@@ -17,7 +17,9 @@
 // Save / Cancel / Load Defaults buttons live as a top-right toolbar that
 // only does navigation back to Models.
 
+import AppKit
 import SwiftUI
+import UniformTypeIdentifiers
 
 struct ModelSettingsScreen: View {
     let modelID: String
@@ -826,6 +828,18 @@ private struct AdvancedTab: View {
             }
         }
 
+        if !vm.isDiffusionModel, vm.model?.expertStreamingSupported == true {
+            SectionHeader(
+                String(localized: "settings.expert_streaming.section",
+                       defaultValue: "SSD Expert Streaming",
+                       comment: "Section header above SSD expert streaming model settings"),
+                subtitle: String(localized: "settings.expert_streaming.subtitle",
+                                 defaultValue: "Keep selected MoE experts resident and stream cold experts from SSD on demand.",
+                                 comment: "Subtitle for SSD expert streaming model settings")
+            )
+            ExpertStreamingSection(vm: vm, client: client)
+        }
+
         SectionHeader(
             String(localized: "settings.advanced.chat_template.section",
                    defaultValue: "Chat Template Kwargs",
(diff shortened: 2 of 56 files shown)


paralin:
Very cool!

pedroraft:
**TL;DR:** benchmark can't sweep `scratch_experts` (3-line fix below); `cache_experts=160` passes admission then OOMs on the second generation under a long prompt; the best cache/scratch split flips with prompt length, and your default looks right for prefill-heavy work. Also: measuring any of this on a busy laptop reversed my result, so numbers here are all from a quiet machine.

Single setup, so treat it as one data point rather than a general result. MacBook Pro M3 Pro / 18 GB, `iogpu.wired_limit_mb = 14336`, internal SSD at 4.9 GB/s contiguous and 3.8 GB/s fragmented. Model `mlx-community/Qwen3.6-35B-A3B-4bit`, 40 layers, 256 experts, top-8, `moe_intermediate 512`. Branch at `3343e44`, `cache_only` mode throughout, so Soft-REAP itself is untested here.

### 1. The benchmark can't sweep the split it exposes

`benchmarks/soft_reap_streaming.py` takes `--cache-experts` but hard-codes `scratch_experts`, so the tradeoff in §3 can't be measured with the shipped tool.

```python
parser.add_argument("--scratch-experts", type=int, default=32)
...
expert_streaming_scratch_experts=args.scratch_experts,   # ModelSettings
...
scratch_experts=args.scratch_experts,                    # install_expert_streaming
```

### 2. Long prompts OOM after the residency estimate says they fit

At `cache_experts=160, scratch_experts=0` the estimate is happy and the model loads at 11.6 GiB, but a 3,343-token prompt peaks at 14.23-14.67 GiB against the 14,336 MB wired limit and dies on the **second** generation, 2 of 3 attempts:

```
[METAL] Command buffer execution failed: Insufficient Memory
(00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory)
```

The bank sizing looks right. Nothing accounts for KV growth landing on top of it, so a config that passes admission still dies once the prompt is long enough. `cache_experts=144` is the largest value that survives here.

### 3. The best split depends on prompt length

Two configs alternated across three fresh invocations each so drift hits both arms equally. Identical slot count, identical 11.619 GiB resident either way. Warm = mean of runs 2-3, sd across invocations.

Long prompt, 3,343 tokens:

| config | decode | prefill |
|---|---|---|
| **112/32** | **15.43 t/s** (sd 0.6%) | **12.93 s** (sd 1.8%) |
| 144/0 | 15.01 t/s (sd 0.7%) | 16.09 s (sd 3.8%) |

Short prompt, ~15 tokens:

| config | decode | prefill |
|---|---|---|
| **144/0** | **17.38 t/s** (sd 20%) | **0.49 s** (sd 1.9%) |
| 112/32 | 10.92 t/s (sd 9.5%) | 1.06 s (sd 2.2%) |

The scratch bank earns its slots when there is prefill to amortise it over, and is dead weight when there is not. So `scratch_experts=32` looks right for the prefill-heavy case your microbench measures, and costs a lot on short-prompt/long-output work where the displaced cache slots show up directly in decode. Probably a docs line rather than a code change. Short-prompt decode is noisy since the runs are startup-sensitive, so take the magnitude loosely; all three paired reps agree on the ordering.

One gotcha for anyone comparing these: `hit_rate` is not comparable across scratch settings, because the prefetch path supplies experts that never count as cache hits. That's why 112/32 reports 58.2% and 144/0 reports 75.8% while decoding at nearly the same speed.

### On measuring any of this

The same config measured 90 minutes apart differed by 10.6% (15.07 then 13.48 t/s), purely because ordinary desktop apps had pushed swap from 2.9 GB to 16.9 GB. On a quiet machine, between-invocation CV is 0.6-0.7%, so contention inflated the noise floor about 15x.

It also reversed the comparison: contended, 144/0 looked 1.3% better; quiet, 112/32 is 2.7% better. The scratch arm recovered more from the cleanup, plausibly because it does more I/O and was competing with swap for the same SSD. An earlier version of my own conclusion here was backwards for exactly that reason, so if you have numbers taken while other things were running, they may be worth re-taking alternating and quiet.

onthehub97:
I'm closing this PR as I've continued work on another branch and managed to get massively better performance on both Qwen3.8-flash and Deepseek-v4-flash-0713.

It gets deepseek-v4-flash-0713 q2.4 running at actually semi usable speeds in <40gb memory.

PR #3261, opened 2026-08-28
feat(glm5): add exact DFlash2 speculative decoding

jonathan308:
## Depends on

- #3251 — exact GLM-5.3 single-row index-score preservation and prefill/decode baseline

## Summary

Add a fail-closed GLM-5.3 target adapter for the DFlash2 runtime already pinned by oMLX. This enables `incoai/GLM-5.3-Flash-DFlash2` to accelerate matched `glm5_next` targets, including custom oQ checkpoints, without changing native autoregressive output.

The integration:

- contracts GLM-5.3 MHC streams at the five checkpoint-declared capture layers;
- validates target/drafter architecture, hidden size, vocabulary, layer count, capture order, and MHC geometry before serving;
- rolls rejected verification tokens back across recurrent KDA state and composite DSA `CacheList(KVCache, PoolingCache)` state;
- replays recurrent state through GLM's exact native vector-gate recurrence instead of the generic innovation tape, which can drift at tail ULPs after repeated rejection cycles;
- chunks long cold prefill independently of prefix-snapshot support;
- loads mlx-vlm and custom oQ targets through their native loaders, eagerly materializes them, and restores process-global hooks on unload;
- stops generation at reasoning/tool parser boundaries before any sentinel can leak into streamed content;
- merges scalar tokenizer EOS metadata with model `generation_config.json` stop-token declarations at the oMLX boundary;
- owns and restores the temporary wired-memory limit used while materializing the target and drafter.

The adapter intentionally fails closed for unproven modes. DDTree/tree verification, target KV quantization, specialized verification linears, and DFlash L1/L2 prefix snapshots remain disabled for GLM-5.3. Image and video requests retain the existing unload-and-VLM-fallback path.

The drafter is not bundled or downloaded automatically. Its checkpoint is published under CC BY-NC-ND 4.0.

## Physical benchmark

Apple M3 Ultra, 256 GB unified memory; `GLM-5.3-Flash-oQ4e`; official `incoai/GLM-5.3-Flash-DFlash2` BF16 drafter; block size 8; single stream; MTP off; DFlash prefix caches off. Every primary matrix case generated all 500 requested tokens and matched the native autoregressive output hash exactly.

| Prompt | Native prefill | DFlash2 prefill | Native decode | DFlash2 decode | Decode gain | Correctness |
|---:|---:|---:|---:|---:|---:|:---|
| 2K | 415.2 tok/s | 407.6 tok/s | 25.8 tok/s | 80.0 tok/s | 3.10x | exact output hash |
| 10K | 378.0 tok/s | 372.0 tok/s | 25.8 tok/s | 79.1 tok/s | 3.07x | exact output hash |
| 20K | 379.4 tok/s | 368.2 tok/s | 25.7 tok/s | 77.6 tok/s | 3.02x | exact output hash |

A separate 1,500-token generation on the production precursor, containing the same production fixes, measured 80.0 tok/s DFlash2 versus 25.3 tok/s native AR and matched the native output exactly.

The exact-output block-size sweep measured 47.1, 56.5, 63.0, 65.9, and 80.0 tok/s at block sizes 3, 4, 5, 6, and 8 respectively. Block size 8 is therefore the tested default.

DFlash2 is a decode optimization, so the expected result is the large decode gain above rather than a prefill gain.

## Tool-call correctness gate

- [x] **Passed on the physical M3 Ultra target/drafter pair.** A forced `get_weather` call stopped at the GLM turn boundary after 32 completion tokens with `finish_reason=tool_calls`, `content=null`, and valid arguments `{"city":"San Francisco"}`. No parser sentinel or trailing text leaked.
- [x] The tool-result continuation completed normally in 2.24 seconds, returned a clean 61-token visible answer, emitted no second tool call, and finished with `stop`.

Unit coverage already exercises reasoning boundaries, tool-call boundaries split across streamed token chunks, generation-config stop tokens, and scalar EOS metadata.

## Tests

- `778 passed` across every DFlash suite plus server, reasoning, parser, and tool-call coverage
- `219 passed` across engine-pool, model-loading, and VLM-adapter coverage
- `git diff --check`
- `py_compile` for all changed Python modules/tests
- Ruff clean for the changed delta when excluding the pre-existing upstream `SIM105` in `dflash_lifecycle.py`

New GLM-specific coverage includes MHC capture, architecture/geometry fail-closed checks, recurrent rejection replay, composite DSA rollback, long-prefill chunking, mlx-vlm/oQ loader selection, eager materialization, parser-boundary stopping, generation-config/scalar EOS handling, wired-memory ownership/restoration, and unload lifecycle cleanup.

### DIFF
--- a/docs/experimental/dflash_mlx_integration.md
+++ b/docs/experimental/dflash_mlx_integration.md
@@ -129,7 +129,9 @@ DSA `CacheList(KVCache, PoolingCache)`. Image/video requests retain the normal
 evict-and-VLM-fallback behavior. DDTree, target KV quantization, specialized
 verify linears, and DFlash L1/L2 prefix snapshots are disabled for this target
 until their GLM composite-cache and numerical-parity suites pass. Adaptive
-linear DFlash verification remains available. The published checkpoint is
+linear DFlash verification remains available. Cold prefill still follows the
+runtime `prefill_step_size`; chunking is kept independent from snapshot
+hydration or publication for this target. The published checkpoint is
 CC BY-NC-ND 4.0; oMLX neither bundles nor automatically downloads its weights.
 
 ### Per-model settings
--- a/docs/experimental/dflash_mlx_integration.md
+++ b/docs/experimental/dflash_mlx_integration.md
@@ -64,16 +64,17 @@ DFlashEngine is a `BaseEngine` implementation that:
 | `omlx/admin/benchmark.py` | Batch test skip guard for DFlashEngine |
 | `tests/test_dflash_engine.py` | DFlash engine and routing tests |
 | `tests/test_dflash_laguna.py` | Laguna adapter parity, cache rollback, config, and checkpoint-layout tests |
+| `tests/test_dflash_glm5.py` | GLM-5.3 MHC capture, KDA/DSA rollback, loader, and lifecycle tests |
 
 ### Dependency
 
-- `dflash-mlx` pinned to `jundot/dflash-mlx` (v0.1.10+omlx.4)
+- `dflash-mlx` pinned to `jundot/dflash-mlx` (v0.1.10+omlx.7)
 - Listed as required dependency in `pyproject.toml`; the mac-app release
   lockfiles are regenerated from it by the packaging pipeline
 
 ### Supported models
 
-DFlash registers `QwenGdnTargetOps`, `Gemma4TargetOps`, and `MuseGlimmerTargetOps`. oMLX also registers a Laguna backend and the `DFlashLagunaForCausalLM` drafter used by Poolside's official checkpoints:
+DFlash registers `QwenGdnTargetOps`, `Gemma4TargetOps`, and `MuseGlimmerTargetOps`. oMLX also registers Laguna and GLM-5.3 target backends. The pinned runtime already contains the generic `DFlash2DraftModel` required by GLM-5.3:
 
 | Target model | Draft checkpoint |
 |--------------|-----------------|
@@ -95,6 +96,7 @@ DFlash registers `QwenGdnTargetOps`, `Gemma4TargetOps`, and `MuseGlimmerTargetOp
 | poolside/Laguna-S-2.1 | poolside/Laguna-S-2.1-DFlash |
 | poolside/Laguna-S-2.1-NVFP4-mlx | poolside/Laguna-S-2.1-DFlash-NVFP4 |
 | meta-models/Muse-Glimmer-30B | meta-models/Muse-Glimmer-30B-assistant |
+| zai-org/GLM-5.3-Flash (and matched MLX quantizations) | incoai/GLM-5.3-Flash-DFlash2 |
 
 Other model families (Llama, Gemma3, etc.) are not supported — they require both a trained DFlash draft checkpoint and a compatible target adapter in dflash-mlx.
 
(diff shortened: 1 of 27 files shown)


jonathan308:
Added two follow-up fixes from physical M3 Ultra validation:

- keep DFlash activity alive during prefill/verification events so the admin UI does not falsely paint an active request amber/red
- gate GLM-5.3 readiness on a disposable DFlash2 warm-up, so the first cold compile happens while the model is still `Loading` rather than leaving the first OpenAI chat request tokenless
- prevent direct loading of DFlash helper checkpoints in the dashboard/API; helpers attach automatically to their target

Cold-process gate on M3 Ultra: target load 34.65s including 5.98s warm-up at 87.5% acceptance; the first post-load `/v1/chat/completions` streaming request completed in 4.27s. Focused tests: 135 passed.

jonathan308:
Follow-up physical agent-path validation after the readiness patch:

- forced OpenAI tool call: valid `echo` call with `{"text":"ready"}`, `finish_reason=tool_calls`, 4.28s wall
- tool-result continuation: valid streamed response, 3.87s wall
- helpers remain internal to the target engine

The apparent tool hang was traced to concurrent model overcommit, not parser failure: a busy ~104GB Qwen process prevented eviction while the ~178GB GLM+DFlash request was admitted with the optional request memory guard disabled (projected 306.74GB over a 254GB static ceiling), causing severe swap and a host reboot. The branch now keeps the static model-residency ceiling hard even when request-time memory throttling is disabled, returning an admission error rather than risking the host. Engine/DFlash/admin suites pass (273 tests on this upstream branch).

Current 4K/500 B8 gate: PP 367.0 tok/s, decode 56.6 tok/s, 84.4% draft acceptance, 11.16s TTFT, 19.998s end-to-end. TTFT is dominated by the measured 4,096-token prefill (~11.09s), not a post-prefill DFlash stall.

jonathan308:
Temporarily closing while I requalify the physical GLM-5.3 path. Post-CI testing surfaced a few release-gate regressions that I do not want to leave presented as ready:

- OpenWebUI-sized multi-tool turns can loop or claim the prompt/tools are corrupted. The same semantic failure reproduces with DFlash disabled, so this is not yet attributed to speculative decoding, but end-to-end tool behavior is not release-ready.
- Fresh randomized code-generation runs are not reproducing the previously qualified 77.6–80.0 tok/s DFlash decode result; acceptance/workload sensitivity and the fixed-prompt benchmark need to be re-audited.
- GLM DFlash prefix-cache and truthful prefill/decode telemetry fixes are being validated separately and are intentionally not being added to this PR.

The branch and signed history are preserved. I will reopen (or supersede with a clean follow-up if preferable) once fixed-prompt exact parity, OpenWebUI tool-call/tool-result loops, cold and cached TTFT, cancellation, and decode throughput all pass again. Apologies for the churn, and thanks for the patience.

jonathan308:
Updated the preserved closed branch with the validated ZCode/OpenAI-compatible skill-envelope fix (signed commit `84254109`).

ZCode can emit either `<tool_call>skill://name</tool_call>` or an unclosed `<tool_call>skill://name<skill id="name">...` form. The parser now promotes only an exact registered URI (or unique bare skill id), emits structured `tool_calls` with `{}` arguments, strips raw markup, and leaves unregistered text untouched.

391 tool-parser tests pass, and live GLM+DFlash2 streaming and non-streaming cweather calls both returned `finish_reason=tool_calls` with no raw markup. PR #3261 remains intentionally CLOSED; this branch update is for review/reference and is not a reopen request.

jonathan308:
Updated the preserved closed branch again (still intentionally CLOSED) with the two GLM-specific parser fixes now deployed in Fusion:

- `30fb905e`: register the native mlx-lm `glm47` parser on the DFlash target/executor tokenizers and advertise tool capability
- `9bd7f09c`: treat glm47 `unknown/raw` sentinels as parser misses so ZCode `skill://` recovery can run

The original failure was not a generic API issue: other models worked, and GLM DFlash was the only lane bypassing native parser registration. Fresh GLM+DFlash validation now returns standard OpenAI `get_weather` and ZCode `skill://cweather` as structured `tool_calls`, including tool-result continuation, with no raw markup. Focused DFlash/tool tests: 524 passed. This branch update is for review/reference; the PR remains closed.

Issue #3239, opened 2026-08-28
Qwen4-Exp sanitize skipped for HF-cache checkpoints → "Received 460 parameters not in model"


zhnext:
**Describe the bug**

Loading a `qwen4_exp` (Qwen3.8-Flash-Next) checkpoint that lives in the HuggingFace cache layout fails with `Received 460 parameters not in model`, listing the N-gram PLE shard tensors. The same checkpoint loads correctly when its files are real files in a plain directory.

In the HF cache, `snapshots/<sha>/*.safetensors` are symlinks to `../../blobs/<sha>`. `_force_qwen4_exp_sanitize_on_load` (`omlx/engine/vlm.py:1038`) hides the `format: mlx` metadata marker so mlx-vlm will not skip `Model.sanitize`, but its guard resolves the filename before matching, so it never matches a symlinked blob. `Model.sanitize` is therefore skipped, and the PLE-shard drop that lives inside it never runs — leaving 460 tensors the mmap-mode embedding has no parameters for.

**To Reproduce**

1. Download a qwen4_exp checkpoint into the HF cache (i.e. anything that produces the standard `~/.cache/huggingface/hub/models--<org>--<repo>/snapshots/<sha>/` layout with blob symlinks) — e.g. `Vontra/Qwen3.8-Flash-Next-MLX-oQ4-MTP` or `-oQ3-MTP`
2. Expose it to oMLX, e.g. `ln -s ~/.cache/huggingface/hub/models--Vontra--Qwen3.8-Flash-Next-MLX-oQ4-MTP ~/Models/Qwen3.8-Flash-Next-MLX-oQ4-MTP`
3. Ensure the model takes the SSD-offload path, so `sanitize` is required to drop the PLE shards. On a 96 GiB machine this is automatic — `force_ssd_offload()` fires because resident (110.82 GB) exceeds the ceiling while mmap (79.53 GB) fits. `sudo sysctl iogpu.wired_limit_mb=92160` is needed for oQ4 to clear admission at all
4. `omlx serve --model-dir ~/Models --port 8080`
5. Send any `POST /v1/chat/completions` request for that model
6. See error

```
... Qwen4-Exp PLE mode for .../snapshots/43a82b3f...: mmap
... VLM loading failed for Vontra--Qwen3.8-Flash-Next-MLX-oQ4-MTP, falling back to LLM:
    Received 460 parameters not in model:
[...]

PR #3262, opened 2026-08-28
fix: match safetensors shards lexically so symlinked checkpoints sanitize

monroewilliams:
## The failure

`Jundot/Qwen3.8-Flash-Next-oQ4e-mtp` fails to load when it is pulled from the Hugging Face Hub cache:

```
Received 384 parameters not in model
```

The same checkpoint, same settings, copied into `~/.omlx/models` starts fine.

This is specific to *where* the checkpoint is stored; anyone who loads a model from their HF cache instead of the models directory hits it. Turning on the new Qwen4-only **SSD N-gram Offload** setting is what makes the skipped sanitize visible instead of harmless.

384 = the `ngram_embedding.shards.N.{weight,scales,biases}` tensors that `DiskBackedShardedEmbedding` deliberately does **not** own when PLE runs in mmap mode. They are supposed to be dropped by `Model.sanitize()` before `load_weights` ever sees them.

## Root cause

`_force_qwen4_exp_sanitize_on_load()` in `omlx/engine/vlm.py` forces mlx-vlm's sanitize path by hiding the safetensors `format=mlx` marker for the duration of one load, scoped to the checkpoint's own shards:

```python
target_dir = model_dir.resolve()
...
if path.parent == target_dir and path.suffix == ".safetensors":   # ← here
    return _SafeOpenMetadataWrapper(handle)
```

mlx-vlm builds those filenames itself:

```python
weight_files = [wf for wf in glob.glob(str(model_path / "*.safetensors")) ...]
with safetensors.safe_open(weight_files[0], framework="np") as f:
    is_mlx_format = f.metadata().get("format") == "mlx"
```

so the prefix keeps the caller's spelling — but `Path(filename).resolve()` **follows the symlink out of the snapshot directory**. An HF Hub cache snapshot keeps every shard as `model-*.safetensors -> ../../blobs/<sha>`, so the comparison silently returns `False`, the marker stays visible, `load_model` skips `Model.sanitize`, and those 384 self-managed tensors reach strict `load_weights`.

Verified against both copies of this checkpoint on a real install — old predicate, all 21 shards in each layout:

| layout | shards matched (pre-fix) | matched (post-fix) |
| --- | --- | --- |
| `~/.omlx/models/…` (real files) | 21 / 21 | 21 / 21 |
| `~/.cache/huggingface/hub/models--Jundot--Qwen3.8-Flash-Next-oQ4e-mtp/snapshots/<rev>/` (symlinks) | **0 / 21** | 21 / 21 |

That is the entire difference between the working directory and the broken one. The same defect covers any symlinked checkpoint layout, not just HF Hub: download mirrors, relocated blob stores, hand-made links.

## Fix

New `omlx/utils/safetensors_paths.py` with `model_shard_matcher(model_dir)`, which matches a shard when **any** spelling of its directory — as written, made absolute, or followed through symlinks — equals **any** spelling of the model directory. Matching lexically first is what keeps symlinked shards attached to the directory they were opened from; the resolved fallback still covers a symlinked *model directory*.

```python
def matches(filename: object) -> bool:
    name = _as_filename_str(filename)
    if name is None or not name.endswith(".safetensors"):
        return False
    return bool(_candidate_parents(name) & targets)
```

Properties, all pinned by tests:
- Only **direct children** match, mirroring the upstream `glob(model_path / "*.safetensors")` scan — the blob store holding the identical bytes does not match.
- Relative shard names resolve against the caller's cwd, so callers that glob relatively are covered too.
- Filenames reach `safe_open` from code paths oMLX does not control, so anything that is not a safetensors path (`None`, an `int`, `bytes`, a nested path) returns `False` instead of aborting the load.
- `model_dir` is resolved once when the matcher is built, not per filename.

Applied to the two overrides in `omlx/engine/vlm.py` that carried this comparison: `_force_qwen4_exp_sanitize_on_load()` and `_force_minimax_m3_moe_sanitize_on_load()`.

## Validation

- `tests/test_safetensors_paths.py` — new, 11 tests: plain shards, symlinked shards, symlinked model dir, relative names, and the non-path inputs above.
- `tests/test_vlm_qwen4_exp_loader.py` — new `test_qwen4_exp_mlx_metadata_is_hidden_for_symlinked_shards` alongside the existing plain-file case. Both assert the marker is hidden for the model's own shards and left **intact** for a file outside the model directory.
- Both new tests fail against the pre-fix code (`format: mlx` stays visible) and pass with the fix, so they are regression guards rather than tautologies.
- `tests/test_safetensors_paths.py tests/test_vlm_qwen4_exp_loader.py`: 18 passed.
- `tests/test_mlx_vlm_minimax_m3_compat.py tests/test_mlx_vlm_qwen4_exp_compat.py`: 71 passed, 1 failed — `test_qwen4_lightning_mtp_fusion_and_runtime_attachment`, an unrelated pre-existing numerics `allclose` at `atol 2e-5`. Reproduced with this PR's `vlm.py` reverted, so it does not come from this change.
- Confirmed live: `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp` with SSD N-gram Offload enabled now loads from the HF Hub cache path.

## Scope

Deliberately narrow: the shared helper plus the two call sites in `vlm.py`. One is the reported Qwen4-Exp path; the other is the identical comparison a few functions away, which seemed worse to leave behind than to include.

`mlx_vlm_inkling_compat` has the same latent defect — its `_FORCE_SANITIZE_DIR` context var holds a `Path` and performs the same `resolve()` comparison, so Inkling checkpoints from the HF cache skip sanitize just as silently (its `_has_raw_inkling_weights()` gate reads *keys*, not the format marker, so the patch engages and then does nothing). That one needs the context var retyped to a predicate plus its own regression test, so it belongs in a follow-up PR that reuses this helper rather than widening this one.

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -53,6 +53,7 @@ from ..utils.image import (
     compute_per_image_hashes,
     extract_images_from_messages,
 )
+from ..utils.safetensors_paths import model_shard_matcher
 from .base import (
     BaseEngine,
     GenerationOutput,
@@ -933,7 +934,7 @@ def _force_minimax_m3_moe_sanitize_on_load(model_dir: Path):
 
     original_safe_open = safetensors.safe_open
     original_sanitize_moe_weights = _minimax_m3_vl._sanitize_moe_weights
-    target_dir = model_dir.resolve()
+    is_target_shard = model_shard_matcher(model_dir)
 
     class _SafeOpenMetadataWrapper:
         def __init__(self, inner):
@@ -958,11 +959,7 @@ def _force_minimax_m3_moe_sanitize_on_load(model_dir: Path):
 
     def _patched_safe_open(filename, *args, **kwargs):
         handle = original_safe_open(filename, *args, **kwargs)
-        try:
-            path = Path(filename).resolve()
-        except TypeError:
-            return handle
-        if path.parent == target_dir and path.suffix == ".safetensors":
+        if is_target_shard(filename):
             return _SafeOpenMetadataWrapper(handle)
         return handle
 
@@ -1054,7 +1051,7 @@ def _force_qwen4_exp_sanitize_on_load(model_dir: Path):
     import safetensors
 
     original_safe_open = safetensors.safe_open
-    target_dir = model_dir.resolve()
+    is_target_shard = model_shard_matcher(model_dir)
 
     class _SafeOpenMetadataWrapper:
         def __init__(self, inner):
@@ -1079,11 +1076,7 @@ def _force_qwen4_exp_sanitize_on_load(model_dir: Path):
 
     def _patched_safe_open(filename, *args, **kwargs):
         handle = original_safe_open(filename, *args, **kwargs)
-        try:
-            path = Path(filename).resolve()
-        except TypeError:
-            return handle
-        if path.parent == target_dir and path.suffix == ".safetensors":
+        if is_target_shard(filename):
             return _SafeOpenMetadataWrapper(handle)
         return handle
 
--- a/omlx/utils/safetensors_paths.py
+++ b/omlx/utils/safetensors_paths.py
@@ -1,0 +1,76 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Path identity for safetensors shards inside symlinked checkpoints.
+
+Several loaders decide whether a checkpoint is ``format=mlx`` from the
+safetensors header metadata, and oMLX sometimes hides that marker for a single
+load so a model's ``Model.sanitize()`` runs. That override is scoped to the
+shards belonging to the model being loaded, so it has to compare the filename
+handed to ``safe_open`` against the model directory.
+
+``Path.resolve()`` is the wrong tool for that comparison. Checkpoint layouts
+that keep shards as symlinks to a content-addressed store -- HF Hub cache
+snapshots, download mirrors, hand-made links -- resolve the shard name out of
+the model directory entirely, the comparison silently returns False, and the
+marker stays visible. The sanitize that would have dropped the model's
+self-managed tensors then never runs, and the load fails later on orphan keys
+reported by ``load_weights``.
+
+Match lexically first (symlinks preserved) and fall back to the fully resolved
+form, so both a symlinked shard and a symlinked model directory match.
+"""
+
+from __future__ import annotations
+
+import os
+from collections.abc import Callable
+from pathlib import Path
+
+__all__ = ["model_shard_matcher"]
+
+
+def _as_filename_str(filename: object) -> str | None:
+    """Return *filename* as a filesystem string, or ``None`` if not a path."""
+    try:
+        return os.fsdecode(os.fspath(filename))
+    except TypeError:
+        return None
+
+
+def _candidate_parents(name: str) -> set[Path]:
+    """Every spelling of the directory a shard name could belong to."""
+    parents = {Path(name).parent, Path(os.path.abspath(name)).parent}
+    try:
+        parents.add(Path(name).resolve().parent)
+    except (OSError, RuntimeError):
+        pass
+    return parents
+
+
+def model_shard_matcher(
+    model_dir: str | os.PathLike[str],
+) -> Callable[[object], bool]:
+    """Return a predicate matching ``safe_open`` filenames in *model_dir*.
+
+    A shard matches when any spelling of its directory -- as written, made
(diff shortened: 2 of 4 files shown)


monroewilliams:
Heads-up on the `test (3.11)` failure on the first run of this PR — it looks like a pre-existing flake in `tests/test_prefix_cache_gdn_split.py`, not something introduced here. Filing the details in case they're useful to whoever owns the cache paths.

**The failure**

```
tests/test_prefix_cache_gdn_split.py::test_split_restore_retries_legacy_candidate_at_the_same_endpoint
  AssertionError: assert (BlockTable(request_id='retry-request', block_ids=[1, 2], num_tokens=8) is not None and 8 == 12)
WARNING omlx.cache.prefix_cache:prefix_cache.py:1271 Rejecting split-GDN placeholder block 3 because its recurrent checkpoint was not committed
=== 1 failed, 10454 passed, 130 skipped, 77 deselected in 459.47s ===
```

`test (3.12)` and `test (3.13)` passed the same commit.

**Why it doesn't look related to this diff**

This PR touches `omlx/utils/safetensors_paths.py`, the two loader overrides in `omlx/engine/vlm.py`, and their tests — no cache or scheduler code. But the behavioural signature is the stronger signal: the same file has failed on several unrelated branches today, each time on a *different* test, always exactly one failure, always the same `prefix_cache.py:1271` warning, always `block_ids=[1, 2]` where 12 tokens were expected.

| Run | Branch | Python | Test that failed |
|---|---|---|---|
| [33192374185](https://github.com/jundot/omlx/actions/runs/33192374185) | `fix/symlinked-checkpoint-sanitize` (this PR) | 3.11 | `..._retries_legacy_candidate_at_the_same_endpoint` |
| [33174537585](https://github.com/jundot/omlx/actions/runs/33174537585) | `fix/planner-peer-memory-settings` | 3.13 | `..._retries_legacy_candidate_at_the_same_endpoint` |
| [33174535838](https://github.com/jundot/omlx/actions/runs/33174535838) | `fix/fabric-firing-requires-job-members` | 3.11 | `..._retry_budget_is_one_per_block` |
| [33174525710](https://github.com/jundot/omlx/actions/runs/33174525710) | `fix/thinking-budget-native-template` | 3.11 | `..._restores_one_sidecar_and_walks_back` |
| [33174525710](https://github.com/jundot/omlx/actions/runs/33174525710) | `fix/thinking-budget-native-template` | 3.12 | `..._dedup_recreates_evicted_sidecar` |

**Where it seems to come from** (narrowed by reading, not proven — happy to be corrected)

The assertion that trips is the test's *first* one, so it fails inside `store_cache()`, before any of the corruption/retry scaffolding the test is named for. Blocks 1 and 2 commit normally; block 3 (endpoint 12) is rejected because `_commit_split_gdn_checkpoint` → `_BoundarySnapshotProvider.commit_gdn_checkpoint` → `take_staged_file()` / `commit_gdn_checkpoint_file()` comes back `None`.

It is consistently the *last* enqueued boundary that loses, which points at `BoundarySnapshotSSDStore`'s background writer thread: the staged file for the final boundary isn't promotable at the moment the commit asks for it. The byte-cap backpressure path doesn't look reachable here (the fixtures queue ~2 KB against a 1 MiB `pending_max_bytes`), and the queue depth is nowhere near `_MAX_PENDING_WRITES`, so this isn't saturation — it reads as a plain completion race.

One thing that may help the next person to hit it: the underlying exception appears to be swallowed at debug level in `_process_write_item` (`logger.debug("Background snapshot write failed: %s", e)`), so a CI capture at the default level shows only the downstream `Rejecting split-GDN placeholder block` symptom and nothing about the write that caused it.

**Repro status:** not reproducible locally — 5 isolated runs plus 40 whole-file runs under sustained CPU load on an 18-core box, all green.

I pushed an empty commit to re-trigger CI since reruns need upstream admin rights, so the queue should settle shortly. Nothing needed from me on this in the PR itself — flagging it so it isn't attributed to the symlink fix, and because the same runner will presumably surface it for the next contributor too.

monroewilliams:
It looks like https://github.com/jundot/omlx/pull/3224 and https://github.com/jundot/omlx/issues/3239 are both fixes for this same issue, so if one of those looks better this one should be closed.

jundot:
Thanks for the thorough analysis and regression coverage. I’m going to take #3224 as the smaller root-cause fix since it was opened first, then add the focused symlink coverage and path hardening in a follow-up. I’m closing this as superseded.

PR #3263, opened 2026-08-28
fix: honor model-owned to_batch when joining a running batch (qwen4_exp concurrent crash)

moc375:
## Fix: honor model-owned to_batch when joining a running batch (qwen4_exp concurrent crash)
- Symptom: >=2 concurrent streams on qwen4_exp (Qwen3.8-Flash-Next) crash with `'QSAKVCache' object has no attribute 'extend'` → "Cache corruption not recoverable after retries" kills all concurrent streams
- Root cause: `_to_batched_cache_layer` misses the model-owned batch conversion path exposed via `to_batch` (same seam as #3215)
- Fix: promote singletons exposing `to_batch`+`merge` before extend (aligned with #3215's `to_batch([0])` form)
- Verified on M3 Ultra: 3/4/5-concurrent all pass, zero corruption; deterministic temp=0 test included

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -985,6 +985,17 @@ def _to_batched_cache_layer(cache_obj: Any) -> Any:
         if all(a is b for a, b in zip(sub_caches, converted)):
             return cache_obj
         return type(cache_obj)(*converted)
+    # Model-owned batch conversion (qwen4_exp QSAKVCache etc.): singletons
+    # exposing `to_batch` convert to their batch variant before extend —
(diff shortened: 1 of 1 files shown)


jundot:
Thanks for reproducing this across several concurrency levels and putting together a fix. #3246 has now landed the `to_batch` conversion together with the related QSA offset and trim fixes, and the late-join workload passes on `main`. The remaining prefill and backfill changes need separate validation, so I’m closing this PR as superseded. I appreciate the time spent investigating this.

PR #3264, opened 2026-08-28
fix: clamp stale seq axis in QSA batched mask build (qwen4_exp concurrent 500s)

moc375:
## Fix: clamp stale seq axis in QSA batched mask build (qwen4_exp long-prompt concurrent 500s)
- Symptom: >=2 concurrent + sparse-QSA-active prompts (>~2051 tok) → `concatenate shapes (2,2,4732) vs (2,1,3)` → 500/stream abort
- Root cause: stale MTP-verify position leaks into the batched query; `put_along_axis` broadcasts the stale seq axis and the tail concatenate then fails
- Fix: clamp `selected_blocks` to the current `seq_len` (keep the freshest query row) — selection staleness <=1 position, acceptance-neutral (cf. SGLang IndexShare findings)
- Verified: 5-concurrent 10/10 + cross-instance 4/4 + single-stream zero regression; previously 100% failure

### DIFF
--- a/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
+++ b/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import logging
 import math
 import mmap
 import os
@@ -649,6 +650,9 @@ class Qwen4ExpGatedDeltaNet(Qwen3_5GatedDeltaNet):
 class Qwen4ExpQSAIndexer(nn.Module):
     """Select compressed key blocks using Qwen Sparse Attention scores."""
 
+    # Rate-limited counter for the defensive seq clamp below.
+    _seq_clamp_hits = 0
+
     def __init__(self, config: TextConfig, rotary_emb):
         super().__init__()
         self.n_heads = config.indexer_n_heads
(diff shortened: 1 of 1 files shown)


pedroberaldo87:
`selected_blocks` is read before it is assigned, so this raises on every call that reaches the guard.

The block is inserted right after the `if max_complete_blocks <= self.block_topk: return None` early exit, but `selected_blocks` is only bound further down, by the `mx.argpartition(...)` call (line 749 in the pre-patch file — its only assignment in `from_projected`). Since that `return None` is conditional, the guard is reachable on the normal sparse path.

Reproduced by extracting `from_projected` via AST and running it with stubs:

```
UnboundLocalError: cannot access local variable 'selected_blocks'
where it is not associated with a value
```

Note that it is the `if` condition itself that raises, not the body, so this hits every call that gets past the early exit — not just the degenerate shape the patch targets.

For what it's worth, the underlying problem is real: I hit the same crash and the clamp does fire in production once it is placed after the assignment — `QSA mask seq clamp fired x2049 (selected 2 rows for seq_len 1)` on a 32-way concurrent run. So the fix looks like just moving the block below the `argpartition`.

moc375:
Good catch — moved the clamp to after the `argpartition` assignment. Thanks for the AST repro and the production confirmation (2049 clamp fires on 32-way concurrent).

jundot:
Thanks for investigating the QSA shape failure and updating the patch after the ordering issue was found. #3246 fixes the upstream cause by using an aligned scalar `past_len` and trimming the batched QSA index arrays; the late-join reproduction passes without this downstream clamp. Since the clamp can leave an incorrectly shaped mask instead of repairing the source, I’m closing this as superseded. I appreciate the detailed investigation.

PR #3265, opened 2026-08-28
feat: batched depth-1 MTP (always-advance) for multi-row decode steps

moc375:
## Feat: batched depth-1 MTP (always-advance) for multi-row decode steps

- Solves "MTP deactivation under concurrency": >=2-row decode steps fall back to plain batching; this adds an opt-in depth-1 speculative cycle on the whole row batch (env `OMLX_MTP_BATCHED_VERIFY`, default off)
- **Sampler-exact-once emission contract**: one sampler call per emitted token, calls monotonic with emission order; stateful token-suppression stays consistent with plain decode
- GDN rollback on rejection uses the per-layer stash in `_restore_gdn_rows` (mlx-lm path). On the mlx-vlm stack, the native `gdn_states` + `rollback_speculative_cache` channel at `batch_generator.py:1521` is the shorter path — thanks @pedroberaldo87 for pointing that out
- Fixed: `ctx.disabled` now clears when the batch composition changes (was process-permanent, matching your finding)
- Measured on Qwen3.8-Flash-Next oQ4e / M3 Ultra: acceptance ~56% batched (vs 61% single-stream); 4-concurrent per-session +70-90% vs plain batched
- On M1 Ultra (per @pedroberaldo87 measurement): `head` cost ≈ `verify` cost, net negative at 8-concurrent — likely silicon-dependent; worth re-measuring on M3+/M4

### DIFF
--- a/omlx/patches/mlx_lm_mtp/__init__.py
+++ b/omlx/patches/mlx_lm_mtp/__init__.py
@@ -147,4 +147,14 @@ def apply_mlx_lm_mtp_patch() -> bool:
 
     apply_ds_consistency()
 
+    # Batched depth-1 MTP for multi-row decode steps. Opt-in via
+    # OMLX_MTP_BATCHED_VERIFY; wraps GenerationBatch.next AFTER batch_generator
+    # installed its own dispatch, so ineligible batches fall straight through.
+    try:
+        from . import batched_verify
+
+        batched_verify.install()
+    except Exception:
+        logger.warning("batched_verify install failed", exc_info=True)
+
     return True
--- a/omlx/patches/mlx_lm_mtp/batched_verify.py
+++ b/omlx/patches/mlx_lm_mtp/batched_verify.py
@@ -58,6 +58,7 @@ _stats = {
     "cycles": 0, "bootstraps": 0, "drafts": 0, "accepts": 0,
     "reject_rollbacks": 0, "gdn_exact_rows": 0, "gdn_pollution_rows": 0,
     "fallbacks": 0, "head_ms": 0.0, "verify_ms": 0.0, "rollback_ms": 0.0,
+    "cycle_ms": 0.0, "emit_ms": 0.0,
 }
 
 
@@ -531,6 +532,8 @@ def _batched_next(gen_batch: Any) -> List[Any]:
         _finish_cycle(gen_batch, ctx, B, keep, finished)
         return responses
 
+    _t_cycle = time.perf_counter()
+
     # ---- steady cycle (sampler-exact-once), depth k >= 1 ----
     # Verify feeds [P', d1, ..., dk]; v_j is the oracle after position j.
     # Per row the sampler is called exactly as often as tokens are emitted:
@@ -548,6 +551,7 @@ def _batched_next(gen_batch: Any) -> List[Any]:
     # speculative entries past the committed fold — trimmed below.
     t_head = time.perf_counter()
     drafts_all: List[List[int]] = []
+    d_rows_lazy: List[Any] = []
     samplers_l = list(getattr(gen_batch, "samplers", None) or [])
 
     def _row_sampler(i: int):
@@ -571,16 +575,20 @@ def _batched_next(gen_batch: Any) -> List[Any]:
                 return_hidden=True, logits_keep=1,
             )
             s_i = _row_sampler(i)
-            drafts: List[int] = [int(s_i(logits_h[:, -1, :])[0].item())]
+            # P1: drafts stay lazy — each feeds the next head call without a
+            # host sync, letting Metal pipeline the whole chain; ONE tolist
+            # after the chain materializes the row.
+            d = s_i(logits_h[:, -1, :])
+            chain: List[Any] = [d]
             h_c = head_hidden[:, -1:, :]
             for j in range(1, k):
-                dj = mx.array([[drafts[-1]]], dtype=emit_tokens.dtype)
                 logits_c, head_hidden_c = host.mtp_forward(
-                    h_c, dj, row.head_cache, return_hidden=True,
+                    h_c, d.reshape(1, 1), row.head_cache, return_hidden=True,
                 )
                 h_c = head_hidden_c[:, -1:, :]
-                drafts.append(int(s_i(logits_c[:, -1, :])[0].item()))
-            drafts_all.append(drafts)
+                d = s_i(logits_c[:, -1, :])
+                chain.append(d)
+            d_rows_lazy.append(mx.concatenate(chain))
         except Exception as exc:
             logger.warning("[batched-verify] head draft failed uid=%s: %s",
                            uid, exc)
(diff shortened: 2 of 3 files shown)


pedroberaldo87:
Two things I ran into while testing this on an mlx-vlm server, plus one measurement.

**The description promises a channel the diff does not contain.** It mentions "a GDN exact-rollback channel using `gdn_states` + per-row `rollback_speculative_cache`", but neither name appears in the diff (`grep -c` over `gh pr diff` is 0 for both). What ships is the stash rollback in `_restore_gdn_rows`. Worth adjusting the description so reviewers know which mechanism they are looking at — especially since that channel does already exist in the sibling module, at `batch_generator.py:1521`, which looks like the shortest path to making this work on the mlx-vlm stack.

**`ctx.disabled` is broader than the L135 comment says.** L135 calls it "banned this batch composition", but your own comment on L568 is the accurate one — "permanently ban". `_generation_batch` is created once in `BatchGenerator.__init__` (mlx_lm/generate.py:1535, its only assignment) and afterwards only mutated in place via `extend`/`filter`, and there is no `disabled = False` anywhere in the diff. So once L570 fires, the speculative path is off for the whole process, not just for the batch that tripped it. On a Qwen3.8-Flash-Next server that gate fires on the first rejection, so in practice it disables itself permanently on the first request. Either aligning the L135 wording or clearing the flag when the composition actually changes would make that less surprising.

**A measurement, for what it is worth.** I wired `gdn_states` through and exposed `keys`/`values`/`_idx`/`trim` on `BatchQSAKVCache` so the path would actually engage instead of hitting the gate. It ran (`cycles=580 accepts=52.4% fallbacks=0`) but throughput dropped — 38.50 and 36.03 tok/s against a 57.09 baseline at 8 concurrent requests. The log says why: `head=12226ms verify=10196ms`, i.e. building the draft costs about as much as verifying it. That was an M1 Ultra, which is not your target, so this may look quite different on newer silicon — but it seems worth measuring before assuming the win.

moc375:
Excellent review, thank you. Three updates:

1. **Description fixed** — removed the `gdn_states` channel reference; the diff ships the stash rollback (`_restore_gdn_rows`), and I've pointed to `batch_generator.py:1521` as the mlx-vlm path per your suggestion.

2. **`ctx.disabled` semantics fixed** — now clears when the batch composition changes (`_finish_cycle` detects row-count changes and resets the flag). The L135 comment is also aligned.

3. **Your M1 Ultra measurement** is very valuable — `head=12s / verify=10s` explains the net negative. On M3 Ultra we see `head≈4.5s / verify≈4s` per 280 cycles (~16ms vs 36ms per cycle), which still nets positive at 4-concurrent but the margin narrows at higher concurrency. The per-row head forward is the dominant cost; on newer silicon the gap should widen. Worth noting this PR defaults to OFF — it's opt-in.

scubamount:
## Verified (with scope caveat)

- **Purpose:** opt-in (`OMLX_MTP_BATCHED_VERIFY`, default off) depth-1 always-advance MTP on multi-row decode steps — MTP currently deactivates under concurrency.
- **Safety design is right:** every eligibility surprise (B<2, logits processors, non-greedy sampler fingerprint, missing MTP surface, omlx row-wise state present) falls through to the standard path; each cycle leaves the backbone cache advanced exactly by emitted tokens, so fallback mid-cycle is safe. Sampler-exact-once emission contract preserves stateful token-suppression semantics.
- **Rejection handling:** per-row full-attn KV right-roll + GDN exact state restore via stash replay, mirroring the proven `mtp_partial_rollback` mechanism.
- **No CI has run** on the current head (branch reports zero checks), so the 1016 new lines are unvalidated by the suite on any Python version — landing posture should wait for at least one green run of `tests/test_*batched_verify*`.

Caveat: correctness rests on the skip-state bookkeeping (pos-0/pos-1 selection per row) and on the two load-bearing numeric claims: (1) GDN Metal kernel tracing bit-comparably under `mx.compile`, (2) the accept/reject KV roll. The docstring's fallback-safe argument is convincing on paper; the scheduler-wiring follow-up must re-prove it with a bit-exactness test against eager.

moc375:
Fix pushed (5389b9e) for the priming-stranding defect tracked in #3494: the extend hook now only stashes/drops the primed head cache when the adopt consumer can actually run (mode enabled AND ≥2 rows post-merge — the same conditions `_eligible` imposes on `_batched_next`). With the mode off, the ctx now survives for the normal MTP activation path.

Measured effect on our serving tree (M3 Ultra, Qwen3.8-Flash-Next oQ4e-mtp, Lightning MTP depth 5, greedy): activation `primed=0` → `primed=<prompt>`, first verify cycle m=0 → m=5, probe acceptance 56-65% → 100% (32/32), 1.1k decode 76-78 → 113.8 tok/s. Thanks to @Rascal for the source-level correction of our initial upstream attribution and for the gating refinement this commit follows.

moc375:
Quick status note, since the last thing on this thread was our own bug report: the cold-start defect from #3494 (adopt queue stranding the primed context) is fixed in 5389b9e, now the branch head. One line — the adopt queue only stashes when the adopter can actually consume the context.

For review context: we run this batched-MTP shape in production at 4-way concurrency. And a third party reproduced our single-stream numbers from the same campaign on different hardware (105-120 tok/s vs our 129 @1k, per-token cost within ~6%): https://github.com/moc375/blog/tree/main/qwen4-speculative-decoding

jundot:
Thanks for working on this. I reproduced several correctness issues on the current head that need fixing before merging:

- `keep` collects duplicate indices across primary and draft emissions, so finished rows can remain active.
- A terminal response can expose a cache containing accepted drafts that were never emitted.
- Non-greedy samplers enter the path and receive raw logits instead of normalized logprobs, changing `top_p` behavior. Sampler calls also exceed the stated exact-once contract.
- Fallback can resume standard decoding after the cache has advanced without restoring committed state. KV rollback failures are also ignored.
- Replacing the batch does not clear `ctx.disabled`, and same-sized row replacements retain stale skip state.

Could you address these and add focused regression tests for the affected paths? I’m not asking for strict MTP-on/off output parity; the main requirement is that each request’s cache stays consistent with its committed tokens.

moc375:
Thanks for the careful review — every point reproduced, all six fixed in 1b34833 with focused tests in `tests/test_batched_verify_review.py` (12 new; the existing 138 still pass). Mapping point → fix:

1. **duplicate `keep` indices** — `_finish_cycle` deduplicates `keep` before the length check and the `filter()` call, so a surviving row's per-emission entries can neither inflate the check (finished rows skipping the filter) nor reach `filter()` duplicated.

2. **terminal cache beyond committed tokens** — `_emit_one` now takes `unemitted_this_cycle` and tail-trims the extracted per-layer caches on finish. A primary-finish trims the accepted drafts behind it; a mid-draft-chain finish trims the remaining accepted drafts. The rejected tail was already trimmed pre-emission; both paths are covered now.

3. **non-greedy samplers** — `_eligible` enforces `_sampler_is_greedy` across all row samplers (falling back to probing `fallback_sampler` when the list is empty), so non-greedy requests decline to the standard path entirely. That also resolves the exact-once concern: for greedy samplers the oracle burst's extra `sample()` calls are value-identical argmaxes, and emission-order suppression stays exact.

4. **fallback after cache advanced / ignored KV rollback** — two parts. KV tail-trim failure now sets `ctx.disabled` and raises `_Fallback` (previously `kv_ok` was computed and dropped). And the fallback handler trims the uncommitted speculative tail (`k + 1 − emitted_this_cycle` per row) whenever the abort landed after the verify forward advanced the cache, so the standard path re-feeds from the last committed token exactly once. Trim failure in that restore is logged at error level rather than swallowed.

5. **`ctx.disabled` on batch replacement** — cleared whenever the uid set changes; the ban belonged to a composition that no longer exists, and with (4)'s restore the standard path ran in between.

6. **stale skip state on same-sized replacement** — re-frontiering is now keyed on the uid *set*, not the row count: any join/leave (including a same-sized swap) drops `skip_hidden`/`skip_logits` so the next cycle bootstraps with a plain forward.

Per your framing we did not chase MTP-on/off output parity — each fix is about cache-vs-committed-tokens consistency. Happy to adjust any of the mechanisms if you'd rather see them shaped differently.

jundot:
Thanks for the update. I checked `1b348335`, but several fixes still miss the actual runtime paths:

- `ctx.cycle_k = k` runs before `k` is assigned, so steady cycles raise `UnboundLocalError` and fall through to standard decoding.
- Deduplicating `keep` still retains a row that finishes on a draft, because its primary already added it to `keep`.
- Tail trimming does not restore GDN `ArraysCache` state, which has no `trim` method. Terminal caches can still contain state from un-emitted drafts.
- Fallback restoration still misses bootstrap failures and general exceptions, and does not restore recurrent state. The bootstrap repro still leaves `[0, 1, 1]` in the cache after emitting `1` once.
- Clearing `disabled` in `_finish_cycle` is unreachable once `_eligible` rejects the disabled context. Same-sized replacements also retain stale skip state until after the next cycle.

The non-greedy gate is fixed. Could you address the remaining cases and exercise them through the actual `GenerationBatch.next()` path? The new fallback tests currently check copied arithmetic or directly assigned flags rather than running the handler. I’d like these resolved before merging, since they affect the core cache lifecycle.

jundot:
@moc375 Thanks for working through the fixes I requested. I’m sorry to change direction after asking you for another round of changes. During the review, I also compared this with #3533 and concluded that it would be a better foundation for batched MTP.

It builds on the existing MTP state and rollback paths, which should make it easier to maintain one consistent cache lifecycle. Keeping a separate decode controller alongside it would add another set of state transitions and recovery paths to maintain. #3533 still needs correctness fixes and validation, so I’m choosing a direction rather than treating it as ready to merge.

I’m going to close this PR and consolidate the implementation around #3533. I appreciate the work you put into this and your patience with the review, and I’m sorry I didn’t make this comparison before asking for more fixes.

Issue #3248, opened 2026-08-28
glm5_next: brief instruction at the head of a long prompt is never retrieved (GLM-5.3-Flash) — deterministic, not fixed by cache/chunking/DSA-dense


williamxie1989:
# glm5_next: brief instruction at the head of a long prompt is never retrieved (GLM-5.3-Flash) — deterministic, and not fixed by cache/chunking/DSA-dense

## Summary

On `GLM-5.3-Flash-oQ8-mtp` served through the mlx-vlm `glm5_next` path, a short
instruction placed at the **head** of a long prompt is reliably ignored, while the
**same prompt at the same total length** with the instruction moved to the **tail**
works every time. The model explicitly reports that it cannot see the user message.

This is reproducible byte-for-byte across server restarts, and — importantly — it is
**not** fixed by disabling the KV cache, chunked prefill, DSA sparse selection, or the
reasoning-effort template injection. I'm filing with the full exoneration table up front
so nobody spends a cycle on the obvious levers.

## Environment

- Mac Studio (512GB), omlx `0.6.3` @ `a20ec09d` (== `origin/main`, verified 0 commits behind)
- Model: `GLM-5.3-Flash-oQ8-mtp` (oQ8 affine, `group_size=64`, `bits=8`), 326.90GB
- `model_settings`: `enable_thinking=true`, `mtp_enabled=false`
- `settings`: `scheduler.chunked_prefill=true`, `cache.enabled=true`,
  `memory.prefill_min_chunk_tokens=32`
- Architecture: `num_hidden_layers=45`, **34 recurrent GDN/linear-attention layers**
  (`is_linear`) + **11 DSA full-attention layers** (layers 3,7,…,43),
  `index_topk=2048`, `index_kpool=4`, `index_kpool_always_select_tail=true`

## Minimal repro

```python
import json, urllib.request

BASE, KEY = "http://127.0.0.1:8015/v1/chat/completions", "1234"
MODEL = "GLM-5.3-Flash-oQ8-mtp"
SYS  = "You are a helpful coding assistant. Follow the user's instruction exactly.\n"
TASK = "TASK WORD: PURPLE-OTTER. Restate only the task word.\n"
FILL = "// neutral filler commentary line to pad context without special tokens\n"

def send(content, max_tokens=64):
    body = {"model": MODEL, "temperature": 0.0, "max_tokens": max_tokens,
            "messages": [{"role": "system", "content": SYS},
                        {"role": "user", "content": content}]}
[...]

PR #3266, opened 2026-08-28
fix(glm5_next): stop routing projections through the qwen35 native qmm tile

williamxie1989:
## Summary

Fixes long-context head/middle retrieval for GLM-5.3-Flash (`glm5_next`). A
brief instruction at the head of a prompt was not followed once the prompt
exceeded ~1k tokens, while the identical prompt worked below the threshold.

Tracked in #3248.

## Root cause

`omlx/patches/mlx_vlm_glm5_next_compat/vendor/mlx_vlm/models/glm5_next/linear.py`
routes the glm5_next projections (`q_a_proj`, `q_b_proj`,
`kv_a_proj_with_mqa`, `o_proj`, MLP/MoE) through `linear_forward()`, which —
once the token count reaches `min_tokens = 1024` for 8-bit weights — dispatches
to the **Qwen3.5 native affine-qmm tile** `qwen35_q8_affine_qmm_t` (from
`omlx.custom_kernels.qwen35_prefill`).

That tile returns **incorrect results** for glm5_next's 8-bit affine weights
(the `_is_supported_affine_linear` gate does not catch it). The corrupted q/kv
projections break long-context attention. **The 1024 threshold is the smoking
gun** — it exactly matches the observed SEEN→LOST boundary (SEEN at ~710 prompt
tokens, LOST from ~1046 on).

## Fix

`glm5_next/linear.py` is glm5_next-specific, so `linear_forward()` now routes
through standard `mx.quantized_matmul` unconditionally (correct at every
length). This matches the known-good `feat/glm5-next` implementation, which
never used the native tile. Scoped to glm5_next only — other models are
unaffected.

## Verification

GLM-5.3-Flash-oQ8-mtp, one-line task word at the HEAD of the prompt, ask the
model to restate it:

| prompt_tokens | 43 | 710 | 1377 | 2710 | 6710 |
|---|---|---|---|---|---|
| before (main) | SEEN | SEEN | LOST | LOST | LOST |
| after (fix) | SEEN | SEEN | SEEN | SEEN | SEEN |

Note: the fix also requires clearing the tiered SSD cold cache once, because the
buggy tile had written corrupted KV values to it (they persist across restarts
and keep failing even after the tile is bypassed).

### DIFF
--- a/omlx/patches/mlx_vlm_glm5_next_compat/vendor/mlx_vlm/models/glm5_next/linear.py
+++ b/omlx/patches/mlx_vlm_glm5_next_compat/vendor/mlx_vlm/models/glm5_next/linear.py
@@ -37,25 +37,14 @@ def _native_qmm(linear: nn.QuantizedLinear, x: mx.array):
 
 
 def linear_forward(linear: nn.Module, x: mx.array) -> mx.array:
-    """Use oMLX's affine prefill tile when it is supported and profitable."""
-    if isinstance(linear, nn.QuantizedLinear):
-        if (
-            "bias" in linear
-            and getattr(linear, "mode", None) == "affine"
-            and getattr(linear, "biases", None) is not None
-        ):
-            out = fused_quantized_matmul(
-                x,
-                linear.weight,
-                linear.scales,
-                linear.biases,
-                bits=int(linear.bits),
-                group_size=int(linear.group_size),
-            )
-            return out + linear.bias
-        out = _native_qmm(linear, x)
-        if out is not None:
-            return out
+    """Route GLM-5.3 projections through standard mlx quantized matmul.
+
+    The oMLX native qwen35 affine-qmm tile (``qwen35_q{bits}_affine_qmm_t``)
+    returns incorrect results for glm5_next 8-bit affine projections once the
+    token count reaches its 1024-token threshold, corrupting long-context
+    attention and breaking head/middle retrieval (jundot/omlx#3248). Standard
+    mlx quantized matmul is correct at every length, so use it unconditionally.
+    """
     return linear(x)
 
 


mightyC1:
Independently hit and root-caused the same bug this week (M3 Ultra 512GB, oQ8e, flip between 1007→1073 prompt tokens at temperature=0) — confirming your findings, and adding one level of precision that suggests a smaller fix preserving the tile's prefill gains.

**The tile itself is bit-exact on every glm5_next weight shape.** We A/B-tested `qwen35_q8_affine_qmm_t` against `mx.quantized_matmul` on all unique (N,K) projections of the actual checkpoint (incl. the 4096→24896 fused KDA concat, K=128 f_b/g_b, unaligned M=1073/1556, bf16+fp16, and deferred-eval chains under allocator pressure): maxdiff = 0 everywhere — **when x is row-contiguous**.

**The actual failure is non-contiguous input views.** `_fused_in_proj` returns `fa_o`/`ga_o` as `mx.split(..., axis=-1)` views. Once `x.shape[-2] >= 1024`, `linear_forward` sends them into the tile; `Qwen35QAffineQmmTPrimitive::unsupported()` does check `row_contiguous`, but it reads `arr.flags()` at op-construction time on a lazy array, where slice-view flags are not authoritative — the view passes and the kernel indexes it as dense, reading adjacent columns of the parent buffer. So the only corrupted matmuls are `f_b_proj`/`g_b_proj` — the forget/output gates of all 34 KDA layers — which is exactly why the model stays fluent (decode < min_tokens → correct fallback) while prompt state is destroyed. `q_a/q_b/kv_a/o_proj/MLP` receive contiguous layernorm outputs and were computing correctly all along.

**Synthetic repro (10 lines, no weights):**
​```python
import mlx.core as mx
from omlx.custom_kernels.qwen35_prefill import fast
mx.random.seed(0); K,N,M,gs=128,8192,2048,64
w,s,b=mx.quantize(mx.random.normal((N,K)),group_size=gs,bits=8)
s=s.astype(mx.bfloat16); b=b.astype(mx.bfloat16)
x=mx.random.normal((1,M,24896)).astype(mx.bfloat16)[...,24576:24576+K]  # split-style view
ref=mx.quantized_matmul(x,w,s,b,transpose=True,group_size=gs,bits=8)
out=fast.qwen35_q8_affine_qmm_t(x,w,s,b,8,gs); mx.eval(ref,out)
print(float(mx.abs(out-ref).max()))   # ~90 (garbage); with mx.contiguous(x): 0.0
​```

**Smaller fix, verified end-to-end:** adding `x = mx.contiguous(x)` immediately before both `getattr(fast, name)(...)` dispatches in `linear.py` (`_native_qmm` + `fused_quantized_matmul`) restores full correctness while keeping the tile on all hot projections — zero-copy for already-dense inputs, a real copy only for the two 128-wide gate views. On our box: needle at line 1 retrieved verbatim at 1006/1555/6687 prompt tokens, mid-document lines 137/250 exact (previously pattern-reconstructed), same PP throughput.

**Proper long-term fix** probably belongs in the primitive: copy-if-non-contiguous at eval time (where flags are authoritative), which also covers the q4 path (`min_tokens=128` — oQ4 conversions were affected from ~128 tokens), q5/q6, and any future caller — same bug class as #2500 (pooling-cache deltas copied out of their parent buffer). Happy to test either variant.

SomeOddCodeGuy:
As noted in the tracked issue, I pulled this branch down can confirm that using it run Pipenetwork's `GLM-5.3-Flash-MLX-8bit` went from having random response and hallucination issues in version 0.6.3 to working as expected in this build.

jundot:
Thanks for isolating the 1024-token boundary and validating the correctness fallback. The later strided-view repro narrowed the cause to `mx.split` views reaching the row-major native QMM, rather than the tile being wrong for GLM shapes generally. I merged #3275, which materializes inputs at both native dispatch sites, covers q5 and q8, and preserves the fast path; the post-merge GLM long-prompt smoke test passed. Since that fixes the same issue more narrowly, I’ll close this PR as superseded.

PR #3267, opened 2026-08-28
fix(glm5_next): don't route 8-bit projections to the q8 prefill tile

BTankut:
## Symptom

On oMLX 0.6.3, `GLM-5.3-Flash` loses the prompt entirely once it reaches ~1024 tokens. The model answers with fluent but completely unrelated text, or says the input "came through as corrupted" / "empty". Below the threshold it is perfect.

Greedy decoding (`temperature=0`), needle-in-a-haystack prompt, needle at 50%:

| prompt tokens | result |
|---|---|
| 862 | correct |
| 968 | correct |
| **1026** | *"it appears to contain a long string of characters..."* |
| 1090 | *"your message came through as unreadable or corrupted text"* |

The cliff is exact, not gradual — that points at a code path, not numerical degradation. It reproduces identically on a plain `oQ4` and an `oQ4e` conversion, so it is not a quantization artifact.

## Cause

`_native_qmm` in `glm5_next/linear.py` sends 8-bit affine projections to `qwen35_q8_affine_qmm_t` once a prefill batch reaches 1024 rows:

```python
min_tokens = 1024 if bits == 8 else 128
```

That tile miscomputes these projections and corrupts the hidden states, so from 1024 tokens on the model is effectively no longer conditioned on the prompt. Falling back to `mx.quantized_matmul` for 8-bit fixes it at every length.

## Scope of the change

Deliberately minimal:

- **Only `_native_qmm`.** I retested with `fused_quantized_matmul` left at `1024` and the model is correct, so that call site is untouched.
- **4-bit is untouched** (`min_tokens = 128`) — verified working well below and above the threshold.

## Verification

After the change, same greedy needle test, needle at 50%:

| prompt tokens | 862 | 1022 | 1086 | 2834 | 5828 |
|---|---|---|---|---|---|
| result | pass | pass | pass | pass | pass |

Also correct on a realistic task: a 2,679-token service table, asked to count `down` services and find the highest p99 — returned 37 and `SVC-049 / 861 ms`, both exact. Vision input and tool calling (`glm47` parser, including parallel calls) were re-tested and are unaffected.

Control: `Ornith-1.0-35B-mlx-bf16` passes the identical prompt at 1012 and 2802 tokens both before and after, so the engine, chat template, and prefix cache were never implicated.

## Cost

Prompt processing is slower, since 8-bit prefill drops back to `mx.quantized_matmul`. Same 12,393-token prompt, cache cold, M3 Ultra:

| | prefill |
|---|---|
| before (tile active) | 31.6 s (~438 tok/s) |
| after (this PR) | 45.1 s (~296 tok/s) |

Generation speed is unchanged. The before-number lines up with the 443–482 tok/s in the 0.6.3 release notes, which is a decent sanity check on the measurement.

This is a correctness-over-speed mitigation, not a real fix — the tile itself is what is wrong. I could not narrow it further from outside: `qwen35_q{4,5,6,8}_affine_qmm_t` are all bit-exact against `mx.quantized_matmul` when I test them standalone at this model's real layer shapes (4096→1536, 1536→16384, 16384→4096, 4096→12288, …, group_size 64, bf16), so the fault only shows in situ. If you can spot the shape or dtype that breaks it, a kernel-side fix would obviously be better than disabling the path.

Happy to run any further test you want on this hardware (M3 Ultra, 512 GB).

### DIFF
--- a/omlx/patches/mlx_vlm_glm5_next_compat/vendor/mlx_vlm/models/glm5_next/linear.py
+++ b/omlx/patches/mlx_vlm_glm5_next_compat/vendor/mlx_vlm/models/glm5_next/linear.py
@@ -13,7 +13,13 @@ def _native_qmm(linear: nn.QuantizedLinear, x: mx.array):
         return None
     if getattr(linear, "mode", None) != "affine" or "bias" in linear:
         return None
-    min_tokens = 1024 if bits == 8 else 128
+    # The q8 tile miscomputes these projections and corrupts the hidden
+    # states, so any prompt that reaches the 1024-row prefill threshold
+    # comes out as unrelated text. Keep 8-bit on mx.quantized_matmul
+    # until the kernel is fixed; 4-bit is unaffected.
+    if bits == 8:
+        return None
+    min_tokens = 128
     if x.ndim < 2 or x.shape[-2] < min_tokens:
         return None
 


goldyard2022:
Independent confirmation of this fix on Apple Silicon, plus a failure mode that is
worth recording because it is much easier to miss than needle-in-haystack.

**Setup**: Mac Studio M3 Ultra (512 GB), oMLX 0.6.3 via Homebrew,
`Jundot/GLM-5.3-Flash-oQ4e` (34 shards, 169 GiB). Text-only, no images,
`max_concurrent_requests=1`, solo requests.

**Manifestation: agentic tool-calling, not just retrieval loss.**
Past the 1024-token prefill threshold the model does not merely lose the context —
it answers a *completely different question*, fluently and confidently. Asking
`1+1=? answer with a number only` with one tool and a padded system prompt returned,
across runs: an HTML document, `"Hi! It's nice to meet you."`, `"# Using Accept..."`,
and `echo 1+1;`. In an agent UI this reads as the model hallucinating or ignoring
instructions, so it gets misattributed to the model rather than to the kernel.

Isolation (question fixed, everything else varied):

| system prompt | tools | result |
|---|---|---|
| none | none | ✅ `2` |
| none | 1 | ✅ `2` |
| short | 1 | ✅ `2` |
| long (2k tok) | none | ✅ `2` |
| long (2k tok) | 1 | ❌ unrelated text |

Neither long-prompt nor tools alone triggers it; tool definitions simply push the
prefill over 1024 rows, which is what actually matters.

**Before / after this patch**, same requests:

| prompt tokens | before | after |
|---:|---|---|
| 815 | ✅ `2` | ✅ `2` |
| 1075 | ❌ HTML document | ✅ `2` |
| 1465 | ❌ `"I'd be happy to help..."` | ✅ `2` |
| 1985 | ❌ `"# Using Accept..."` | ✅ `2` |
| 2765 | ✅ `2` (by luck) | ✅ `2` |
| 4065 | ❌ `echo 1+1;` | ✅ `2` |

Note 2765 passing while 1985 and 4065 failed — the corruption is not a clean
threshold once past 1024, which makes spot-checking unreliable.

Real agentic requests (long system prompt + 10 tool definitions) also recover:
correct `tool_calls` with sensible arguments instead of unrelated prose.

**Confirming the cache footnote from #3270**: clearing the SSD prefix cache after
patching is required, not optional. Poisoned KV from before the fix replays across
restarts otherwise.

**One note on scope, in case it helps reviewers**: `fused_quantized_matmul` further
down the same file carries the identical `min_tokens = 1024 if bits == 8 else 128`
routing. For this checkpoint it is unreachable — it is only entered for
`QuantizedLinear` layers that carry a bias, which here are exclusively the
`vision_model` projections, and those have no 8-bit overrides (`quantization`
reports `{8: 451, 6: 20, 5: 71}` for `language_model` and nothing for
`vision_model`). So this PR is complete as written for GLM-5.3-Flash-oQ4e. It would
only matter for a future oQ conversion that boosts vision-tower projections to
8 bits.

Thanks for the quick turnaround on this one — the failure was genuinely hard to
attribute from the outside.

jundot:
Thanks for the careful bisection and for quantifying the performance cost of the q8 fallback. That fallback proved the native dispatch was involved, while the later minimal repro narrowed the failure to strided inputs rather than q8 itself; q5 can hit the same bug from its 128-token threshold. I merged #3275, which makes both native call sites safe without giving up the fast path, and verified it on the merged GLM build. I’ll close this PR as superseded.

PR #3268, opened 2026-08-28
fix(api): preserve empty no-thinking markers across turns

Termina1:
## Summary

Qwen's native no-thinking template emits an empty reasoning marker in the assistant generation prefix:

```text
<think>\n\n</think>\n\n
```

The OpenAI response path currently collapses the empty extracted reasoning to `None`. A stateless client therefore cannot echo the marker on the next request, and the same logical conversation re-renders to a different token sequence. This prevents exact prefix-cache reuse past the first assistant turn.

When native reasoning is active and the existing settings explicitly request both:

```json
{
  "enable_thinking": false,
  "preserve_thinking": true
}
```

this PR returns the marker interior as `reasoning_content: "\n\n"`. The existing native reasoning path then reconstructs the same prompt tokens on the next turn. Other configurations are unchanged.

## Evidence

Using Qwen3.8 Flash Next through oMLX with a clean cache:

- before: first divergence at token 917 (`<think>\n\n</think>` vs direct `<tool_call>`), final `cached_tokens=0`;
- after: the next turn renders the same prefix and restores `cached_tokens=2048`;
- a 15-case file-localization run restored 2048 final-pass cached tokens on all 15 cases.

## Test plan

- `pytest -q tests/test_api_utils.py` — 240 passed
- Live multi-turn Qwen3.8 Flash Next test with a clean prefix cache — 2048-token hit restored

## AI assistance

This PR is AI-assisted. AI tools were used during investigation, implementation, and test preparation; the token-level diagnosis, diff, and live oMLX results were manually reviewed and verified on local Apple Silicon hardware.

### DIFF
--- a/omlx/api/utils.py
+++ b/omlx/api/utils.py
@@ -45,6 +45,24 @@ def uses_native_reasoning_content(
     return "minimax" in lowered and "m3" in lowered
 
 
+def reasoning_content_for_history(
+    content: str,
+    *,
+    native_reasoning: bool,
+    chat_template_kwargs: dict[str, Any],
+) -> str | None:
+    """Retain Qwen's empty no-thinking marker when history preservation is enabled."""
+    if content:
+        return content
+    if (
+        native_reasoning
+        and chat_template_kwargs.get("enable_thinking") is False
+        and chat_template_kwargs.get("preserve_thinking") is True
+    ):
+        return "\n\n"
+    return None
+
+
 def merge_reasoning_effort_chat_template_kwargs(
     chat_template_kwargs: dict[str, Any] | None,
     reasoning_effort: Any | None,
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -173,6 +173,7 @@ from .api.utils import (
     has_nonleading_system_message,
     merge_reasoning_effort_chat_template_kwargs,
     prepare_system_messages_for_template,
+    reasoning_content_for_history,
     uses_native_reasoning_content,
 )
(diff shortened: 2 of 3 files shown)


jundot:
Thanks for the investigation. I do not plan to merge this PR in its current direction: on the current Qwen3.6 and Qwen3.8 templates, `preserve_thinking=true` already reproduces the same empty marker without `reasoning_content`, while this helper changes Gemma 4 prompt tokens and does not cover streaming responses. I plan to close this after #3280 lands; if the cache divergence still reproduces on current main after that, I’d rather review a new fix based on that reproduction.

Termina1:
Confirmed on current `main` (`8706683`, including #3280) with Qwen3.8 Flash Next and a clean prefix cache.

A/B result using the same multi-turn tool conversation:

- `enable_thinking=false`, `preserve_thinking=true`: search3 → final token LCP was the full 2,510-token search3 prompt; final restored `cached_tokens=2048` even though prior API messages had no `reasoning_content`.
- `enable_thinking=false`, `preserve_thinking=false`: final restored `cached_tokens=0`; the divergence probe reported the first mismatch at token 917 (`<think>\n\n</think>` before the tool call versus a direct `<tool_call>`).

So the empty marker already round-trips correctly through the current Qwen template when the existing setting is actually enabled. The missing admin persistence fixed by #3280 explains the apparent failure; this API helper is unnecessary and would broaden behavior incorrectly.

Thanks for pointing this out. Closing the PR.

PR #3269, opened 2026-08-28
fix(oq): calibrate Qwen4 through RAM-safe proxies

actualrealhuman:
## Summary

Fresh oQe quantization of Qwen3.8-Flash-Next failed on a 128 GB M5 Max in the RAM-safe calibration-proxy path.

Follow-up to #3174. Fresh oQe calibration for Qwen3.8 Flash Next can require the RAM-safe uniform-4-bit proxy path, but that path did not initialize the Qwen4 runtime before sanitizer/model lookup and did not load SSD-mapped PLE tensors with the intended lenient VLM loader contract.

This left fresh calibration unable to reach the existing Qwen4 layer walk on machines that cannot load the full BF16 checkpoint.

## Problem

The failure occurred in three stages:

1. Quantization bypassed serving's Qwen4 compatibility registration, so sanitizer discovery could fail and leave packed MoE expert tensors in BF16, producing an oversized proxy.
2. Proxy admission charged the complete on-disk checkpoint and recomputed capability from transient live memory, including pressure created by the proxy itself. Qwen4 PLE rows are SSD-mapped and the manual calibration walk does not invoke the vision tower or ordinary LM head.
3. Once the proxy layout was corrected, `mlx_vlm.load_model()` rejected 384 retained PLE weight/scale/bias tensors that are intentionally consumed by mmap modules instead of owned as normal MLX parameters. The load exception was converted into the generic `oQe imatrix collection produced no entries` error.

## Fix

- Detect Qwen4-Exp through both top-level and text configuration.
- Register and bind the Qwen4 runtime before sanitize-plan discovery, forcing PLE mmap during calibration while preserving MTP when present.
- Quantize the sanitized packed expert representation when building the proxy.
- Account for the Qwen4 tensors actually resident during the manual calibration walk, while retaining complete checkpoint accounting for other model families.
- Base hard calibration admission on the stable configured machine/Metal capability; retain live availability for diagnostics and micro-batch decisions.
- Pass `strict=False` through the supported mlx-vlm loader API for calibration proxies instead of temporarily monkeypatching `Module.load_weights`.
- Preserve the first load/layer/capture failure stage when an iMatrix collection is empty.

## Validation

- Four representative regression tests fail against unchanged upstream and pass with this change.
- 43 focused Qwen proxy, iMatrix, MTP, and memory-admission tests passed.
- All 591 tests in `tests/test_oq.py` and `tests/test_process_memory_enforcer.py` passed.
- `pytest -m "not slow"`: 10,476 passed, 109 skipped, and four unrelated failures. All four reproduce against unchanged upstream (three existing numerical-tolerance failures and one bundled subprocess that cannot import MLX).
- Changed-file Python syntax compilation passed.
- Critical Ruff checks (`E9,F63,F7,F82`) passed; the full-file Ruff findings are pre-existing in the upstream files.
- `git diff --check` passed.

Real-checkpoint validation used the canonical Qwen3.8-Flash-Next BF16 checkpoint (131 safetensors shards):

- The corrected proxy was approximately 98 GB on disk and 67 GB resident.
- Fresh Qwen4 trunk and MTP calibration produced 937 iMatrix entries.
- The iMatrix cache was written and subsequently reused.
- The oQ4e conversion completed, the output loaded with PLE mmap enabled, and inference completed successfully on a 128 GB Apple Silicon Mac.

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -14,6 +14,7 @@ import json
 import logging
 import re
 import shutil
+import struct as _struct
 import tempfile
 import time as _time
 from dataclasses import dataclass
@@ -54,6 +55,37 @@ _QWEN4_EXP_NGRAM_SHARD_RE = re.compile(
     r"\.ple\.ple_embedding\.ngram_embedding\."
     r"(?:shard_\d+|shards\.\d+)$"
 )
+
+
+def _is_qwen4_exp_config(config: dict) -> bool:
+    """Return whether *config* selects the Qwen4-Exp/Flash-Next family."""
+    text_config = config.get("text_config")
+    if not isinstance(text_config, dict):
+        text_config = {}
+    return any(
+        str(model_type or "").lower().startswith("qwen4_exp")
+        for model_type in (config.get("model_type"), text_config.get("model_type"))
+    )
+
+
+def _configure_qwen4_exp_quantization_runtime(
+    model_path: str | Path,
+    config: dict,
+    *,
+    preserve_mtp: bool,
+) -> bool:
+    """Register Qwen4 and bind its mmap PLE/MTP state for quantization."""
+    if not _is_qwen4_exp_config(config):
+        return False
+    from omlx.patches.mlx_vlm_qwen4_exp_compat import configure_qwen4_exp_runtime
+
+    configure_qwen4_exp_runtime(
+        model_path,
+        mode="mmap",
+        mtp_enabled=bool(preserve_mtp),
+    )
+    return True
+
 
 _MAX_MODEL_RAM_FRACTION = 0.75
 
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for the thorough fix. Registering Qwen4 before sanitizer discovery, building a RAM-safe proxy, and loading PLE shards from the model directory through mmap address the 128 GB calibration failure.

LGTM. I’ll handle two narrow cleanups in a follow-up. I’ll reuse the pre-proxy live-memory budget instead of introducing a stable capability policy.

I’ll also remove `qwen4_exp_artifact`. It is an oMLX-only checkpoint layout, not a format defined by mlx-lm or official Qwen checkpoints. I do not want oMLX to create a private checkpoint contract that converters and loaders must maintain. This will not remove same-directory PLE mmap support, and I’ll reconsider an external layout if mlx-lm standardizes one or official Qwen checkpoints adopt it.

PR #3275, opened 2026-08-28
fix(glm5_next): copy strided inputs before the affine prefill tile

davidtorcivia:
`linear_forward` and `fused_quantized_matmul` dispatch a quantized projection to `qwen35_q{bits}_affine_qmm_t` once the token count reaches `min_tokens`, which is 1024 for 8-bit weights and 128 otherwise. The tile indexes `x` as row-major packed and ignores its strides, so an input that is a view returns wrong values and raises nothing.

`Glm5NextLinearAttention._fused_in_proj` computes q, k, v, f_a, g_a and b in one matmul and splits the result on the last axis. The f_a and g_a splits are views whose row stride is the full fused width, and they are the inputs to `f_b_proj` and `g_b_proj`. On GLM-5.3-Flash both are 128 to 8192, so the forget gate and the output gate of every linear-attention layer are computed from the wrong lanes.

The visible effect is that the model stops seeing its prompt and answers a question nobody asked. On `GLM-5.3-Flash-oQ4e`, 27 of the 34 linear-attention layers hold 8-bit projections and 7 hold 5-bit, so 7 layers are wrong from 128 tokens onward and the remaining 27 join at 1024.

Both call sites now copy `x` with `mx.contiguous` before the dispatch.

Measured on `GLM-5.3-Flash-oQ4e`, M3 Ultra 256 GB, greedy, a fresh prompt per request so nothing hits the prefix cache. Before: correct up to 1018 prompt tokens, wrong at 1025, 1465, 1730, 2374, 4693, 9574 and 19365. After: correct at every one of those lengths.

Tests cover `linear_forward` and `fused_quantized_matmul` at 5 bits with 128 tokens and at 8 bits with 1024 tokens, each on a last-axis `mx.split` view. All four fail without the copy.

### DIFF
--- a/omlx/patches/mlx_vlm_glm5_next_compat/vendor/mlx_vlm/models/glm5_next/linear.py
+++ b/omlx/patches/mlx_vlm_glm5_next_compat/vendor/mlx_vlm/models/glm5_next/linear.py
@@ -24,6 +24,9 @@ def _native_qmm(linear: nn.QuantizedLinear, x: mx.array):
         name = f"qwen35_q{bits}_affine_qmm_t"
         if not fast.has_symbol(name) or not _is_supported_affine_linear(linear, x):
             return None
+        # The tile indexes x as row-major packed and ignores its strides, so a
+        # last-axis view reads the wrong lanes.
+        x = mx.contiguous(x)
         return getattr(fast, name)(
             x,
             linear.weight,
@@ -87,6 +90,8 @@ def fused_quantized_matmul(
 
             name = f"qwen35_q{bits}_affine_qmm_t"
             if fast.has_symbol(name) and fast.qmm_supports_group_size(group_size):
+                # The tile ignores the input's strides. See _native_qmm.
+                x = mx.contiguous(x)
                 return getattr(fast, name)(
                     x,
                     weight,
--- a/tests/test_mlx_vlm_glm5_next_compat.py
+++ b/tests/test_mlx_vlm_glm5_next_compat.py
@@ -650,3 +650,66 @@ def test_glm5_next_q8_indexer_prefill_uses_shared_qmm_kernel(monkeypatch):
 
     assert calls == 1
     assert mx.allclose(actual, reference, atol=2e-3, rtol=2e-3).item()
+
+
+@pytest.mark.parametrize(("bits", "tokens"), [(5, 128), (8, 1024)])
+def test_glm5_next_prefill_qmm_handles_strided_input(bits, tokens):
+    import mlx.nn as nn
+    from mlx_vlm.models.glm5_next.linear import linear_forward
+
+    from omlx.custom_kernels.qwen35_prefill import fast
+
+    name = f"qwen35_q{bits}_affine_qmm_t"
+    if not fast.has_symbol(name):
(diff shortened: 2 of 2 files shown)


gilby:
Independently verified on different hardware and a different checkpoint — this fixes it.

**Setup:** M3 Ultra / 512 GB, `GLM-5.3-Flash` at **oQ4e** (not the quant used in the PR description), cherry-picked onto a branch that also carries #3228 plus some MTP fixes. Same history-length sweep as before the fix, same prompts, same on-topic detector, non-repetitive filler:

| prompt_tok | 156 | 518 | **1376** | 2877 | **5769** |
|---|---|---|---|---|---|
| before | ✓ | ✓ | **✗** | ✓ | **✗** |
| after | ✓ | ✓ | **✓** | ✓ | **✓** |

5/5 on-topic. The two lengths that reliably failed for us now pass, and every response identifies the actual task instead of answering something unasked. `tests/test_mlx_vlm_glm5_next_compat.py` is 14 passed / 8 skipped on our tree.

**The threshold story matches exactly.** Our observed onset was "somewhere past ~1–2K tokens", which never lined up with anything architectural we could find — the dense→sparse indexer switch is at 2048, the paged-cache block size is 2048, neither explains a failure at 1376. `1024` for the 8-bit projections does, precisely, and the 5-bit layers at 128 explain why shorter prompts were occasionally flaky too.

**Every hypothesis we had was wrong**, and worth recording so nobody re-runs them:

- *Run-time-variable prefill chunking* — you killed this yourself: the fragmented run was the one that worked
- *Missing `mx.depends` barrier on the sparse fast paths* — real ordering gap, not this bug
- *Leaked `lengths`/`left_padding` zeroing prompt tokens via `has_right_padding`* — would have produced the same "invisible anywhere in the prompt" symptom, which is exactly why it looked so convincing
- *Prefix-cache partial-hit restoring KDA state without position validation* — the placeholder test at `prefix_cache.py:4161` really is a shape test not a position test, but it isn't this

All four were downstream reasoning from the symptom. A quantized tile ignoring input strides on an `mx.split` view was not on the list, and the "coherent answer to a *different* question" shape makes much more sense once it's the forget/output gates being read from the wrong lanes rather than the tokens being dropped.

**Also confirming your performance caveat, from the other direction.** We measured 86.9% acceptance / 3.00 tokens-per-cycle at depth 3 on the MTP head and quoted those numbers publicly. Those were taken on the buggy build, so by your reasoning — corrupted gates making the backbone more predictable — they're inflated for the same reason yours were, and we'll re-measure rather than let them stand.

Thanks for chasing it down to the kernel. We'd resorted to writing a phased instrumentation plan because we couldn't get a hypothesis to survive contact with the evidence; you found the actual cause instead.

davidtorcivia:
Fixes #3270 as single commit instead of a multi commit

davidtorcivia:
Same bug fix as #3267 and #3266  - take whatever is cleaner

jundot:
Thanks for tracing this to the native QMM stride assumption. I verified that materializing the split views fixes the q5 and q8 paths while preserving the native fast path, including a real GLM-5.3 long-prompt smoke test. The focused regression coverage exercises both dispatch sites at their actual thresholds. LGTM, merging.

PR #3255, opened 2026-08-28
feat: chip-aware default oQ dtype (float16 on M1/M2, bfloat16 on M3+)

Tzahreddin:
## Summary

- Detect Apple Silicon generation in `omlx.utils.hardware` and default non-quantized oQ weights/scales to `float16` on M1/M2 and `bfloat16` on M3+.
- Persist `quantization.default_oq_dtype` in `settings.json` (auto-migrated when missing) and expose it via the admin global-settings API.
- Wire the default through the web admin quantizer, the native macOS quantizer screen, and `/api/oq/start` when no dtype is sent.

## Testing notes

**Author hardware:** Apple M2 Max only.

- Verified manually on M2 Max: `settings.json` gets `"default_oq_dtype": "float16"`, the native app quantizer pre-selects `float16`, and quantization starts with the expected dtype.
- Python unit tests added for hardware detection and settings migration (`tests/test_utils_hardware.py`, `tests/test_settings.py`).

**Not tested by author:** M3/M4 (or any M3+ machine). The M3+ path (`bfloat16` default) is implemented and covered by mocked unit tests only — a maintainer with M3+ hardware should sanity-check the native app quantizer and admin UI before merge.

## Test plan

- [x] `uv run pytest tests/test_utils_hardware.py::TestChipGenerationAndOqDtype tests/test_settings.py::TestQuantizationSettings`
- [x] Manual: M2 Max — native quantizer shows `float16` selected by default
- [ ] Manual: M3+ — confirm `bfloat16` is pre-selected in native app + web admin quantizer
- [ ] Manual: fresh install without `quantization` block in `settings.json` migrates to chip-aware default

Made with [Cursor](https://cursor.com)

PR #3277, opened 2026-08-29
perf(loading): use float16 activations on M1/M2, where bfloat16 is emulated

pedroberaldo87:
## The problem

M1 and M2 have no bfloat16 circuit — Metal emulates it in software. float16 they run
natively.

Which one the forward uses is decided by the checkpoint's non-quantized tensors: the
scales, biases, norms and embeddings. On a published quant those are bf16, so every
forward pays the emulation, on hardware that has the fast path sitting idle.

## What this does

After the model is loaded, walk the parameter tree and recast **only the bf16 tensors** to
fp16. On Qwen3.8-Flash-Next that is 2725 tensors (15.41 GB). The 1020 quantized U32 tensors
(90.88 GB) are packed integers — they have no float dtype and are never touched.

Both dtypes are 16 bits, so nothing grows. The walk goes in chunks of 8, updating the tree
as it goes, so the old tensors are freed along the way instead of all at the end — otherwise
peak memory would spike by 15 GB. Cost: 0.6 s on a 42 s load, measured over 63 loads.

Off by default. `model_settings.activation_fp16_enabled` per model, or `OMLX_ACT_FP16=1`
for a one-off A/B.

## Results

Apple M1 Ultra 128 GB, `Qwen3.8-Flash-Next-oQ4e-mtp`, model-card sampling.

**Decode** — spec-ab neutral preset, 24 runs per side per scenario, two loads per arm:

```
scenario           bf16          fp16    gain     floor
rewrite    28.29+-0.67   32.26+-0.94  +14.0%      6.9%
novel      22.89+-0.36   24.07+-0.29   +5.2%      2.8%
```

**Prefill** — arms interleaved bf16/fp16/bf16/fp16, unique prompt each run, no cache hits:

```
mode    n     mean    se
bf16    6    313.2   3.2   tokens/s
fp16    6    500.2   6.6   tokens/s     +59.7%  [+55.0 .. +64.4]
```

Prefill gains four times what decode does: it processes thousands of positions at once, so
arithmetic dominates. Decode processes 1-16 and is latency-bound.

Per-cycle cost drops ~9% in both decode scenarios while draft acceptance stays flat
(0.588→0.601, 0.470→0.442). That is a cheaper forward, not better drafting.

## Correctness

All 12 prefill runs (~19k unique tokens each) returned the same answer, and it is correct.
Not bit-identical to bf16 — 8 mantissa bits vs 11, so rounding differs and greedy can take a
different path. No saturation at 17.4k prompt tokens.

`tests/test_activation_fp16_recast.py` pins that the recast converts bf16 leaves, leaves U32
payloads bit-identical, and is a no-op when nothing is bf16.

## Relation to #3255

That PR picks the dtype when **generating** a checkpoint. This one fixes a checkpoint that
already exists on disk — for anyone who downloaded a published bf16-scaled quant and cannot
regenerate 100 GB+. They compose; neither replaces the other.

## Not claimed

Untested on M3+, where bf16 is native and this should stay off. Numbers come from one model.

### DIFF
--- a/omlx/admin/i18n/en.json
+++ b/omlx/admin/i18n/en.json
@@ -625,6 +625,7 @@
   "modal.model_settings.limit_tool_placeholder": "e.g. 2000",
   "modal.model_settings.activation_fp16": "Convert activations to float16 on load (recommended on M1/M2)",
   "modal.model_settings.activation_fp16_hint": "Recasts the checkpoint's bfloat16 leaves to float16 after loading, so the forward runs on the GPU's native float16 path instead of the emulated one. Quantized weights are untouched. Takes effect on the next model load.",
+  "modal.model_settings.activation_fp16_unsupported": "This checkpoint is already stored in float16 — there is nothing to recast.",
   "modal.model_settings.turboquant_kv": "TurboQuant KV Cache",
   "modal.model_settings.turboquant_kv_hint": "Compress KV cache using vector quantization. Lower bits = more compression, higher bits = better quality. Supports 2 to 8 bits.",
   "modal.model_settings.turboquant_kv_bits_label": "Bits per channel",
--- a/omlx/admin/i18n/en.json
+++ b/omlx/admin/i18n/en.json
@@ -623,6 +623,8 @@
   "modal.model_settings.limit_tool_result": "Limit Tool Result Tokens",
   "modal.model_settings.limit_tool_result_hint": "Truncate large tool results (e.g. file reads) to a token limit",
   "modal.model_settings.limit_tool_placeholder": "e.g. 2000",
+  "modal.model_settings.activation_fp16": "Convert activations to float16 on load (recommended on M1/M2)",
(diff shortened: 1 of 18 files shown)


scubamount:
## Verified

- **Root cause correct:** M1/M2 have no native bfloat16 ALU (Metal emulates); forward dtype follows non-quantized leaves, so bf16 published quant pays emulation on hardware with idle fp16 path.
- **Fix right:** `cast_bf16_params_to_fp16` recasts only `bfloat16` leaves via `tree_flatten` (quantized U32 untouched, same 16-bit width — nothing grows), chunked at `_MATERIALIZE_EVAL_CHUNK` with per-chunk `tree_unflatten` tree update so bf16 originals free as it goes (no 2x peak).
- **Gating:** off by default, `OMLX_ACT_FP16=1` env or per-model setting; `checkpoint_has_bf16_leaves` reads safetensors headers only to gray out the toggle for fp16-native checkpoints; recast failure logs and stays on bf16.
- **Wiring:** same pattern in `apply_post_load_transforms` (text) and `vlm.py` load path (VLM), both on the MLX executor; i18n keys added across all 9 locales; dedicated `tests/test_activation_fp16_recast.py`.

No issues. Correct chip-aware opt-in with bounded memory.

PR #3280, opened 2026-08-29
fix(admin): persist mtp_num_draft_tokens via settings PUT

### LINKED ISSUE #2823: Admin settings PUT silently discards mtp_num_draft_tokens while returning success:true (file + /admin/api/reload works)
## Summary

`PUT /admin/api/models/{id}/settings` accepts `mtp_num_draft_tokens`, returns
`{"success": true}`, and silently discards it. The value never reaches
`model_settings.json` and never takes effect.

Editing `~/.omlx/model_settings.json` directly and calling
`POST /admin/api/reload` **does** work, so the setting itself is fine — only the
admin PUT path drops it.

The problem is the `success: true`. A rejected key would be a five-second
diagnosis; a silently accepted one reads as tuned and does nothing. I concluded
the setting was unsupported entirely and only discovered otherwise after seeing
`mtp_num_draft_tokens: 8` in another user's report (#2779).

## Environment

- Mac Studio, Apple M3 Ultra, 512 GB, macOS 26.6.1
- oMLX **0.6.2**
- `Qwen3.8-27B-...-oQ8e-mtp`, `model_type=qwen3_5`, `engine_type=vlm`, `mtp_enabled: true`

## Reproduction

```
PUT /admin/api/models/{id}/settings   {"mtp_num_draft_tokens": 2}
-> 200 {"success": true, ...}
   response settings.mtp_num_draft_tokens: null
   on-disk model_settings.json:           null

PUT ... {"mtp_num_draft_tokens": 3}
-> 200 {"success": true, ...}   readback: null
```

Same value via the file plus a reload:

```
edit ~/.omlx/model_settings.json -> "mtp_num_draft_tokens": 2
POST /admin/api/reload
-> 200 {"status": "ok", "message": "Re-discovered 2 models from 2 directories"}
   API readback: 2      on-disk: 2      (in effect, confirmed by MTP telemetry)
```

Other keys in the very same PUT body apply correctly (`max_context_window`,
`mtp_enabled`, `chat_template_kwargs`), so this is specific to
`mtp_num_draft_tokens` being absent from the request model rather than a broken
endpoint.

## Why it matters

Draft depth is the main lever on MTP acceptance, so it is exactly the knob you
reach for when tuning speculative decoding. Measured on this box at 0.6.2,
temperature 0.0, 600-token generations, 3 trials each:

| depth | tok/s | tok/cycle | accept % |
|---|---|---|---|
| 1 | 30.46 | 1.67 | 66.9 |
| 2 | 33.84 | 1.91 | 59.6 |
| 3 | 36.05 | 2.14 | 60.7 |
| 4 | 35.50 | 2.20 | 61.2 |
| null (adaptive) | 34.62 | 2.05 | 58.6 |

An 18% spread between depths is worth tuning, and via the documented admin API it
currently looks like the knob does nothing.

## Suggested fix

Add `mtp_num_draft_tokens` to the settings PUT request model. Failing that, reject
unknown keys with a 4xx and name them, so a discarded setting cannot masquerade as
an applied one.

yanzhaohui1999:
## Summary

`PUT /admin/api/models/{id}/settings` silently dropped `mtp_num_draft_tokens`
(returning `success: true` while the value never reached
`model_settings.json`). The same silent-drop class also affected
`preserve_thinking` and `turboquant_skip_last` — all three are valid
`ModelSettings` fields but were never registered on the admin PUT path, so
Pydantic's default `extra="ignore"` swallowed them at parse time and the
handler never applied them.

Changes:

- Register `mtp_num_draft_tokens`, `preserve_thinking` and
  `turboquant_skip_last` on `ModelSettingsRequest` and apply them in the
  PUT handler. `mtp_num_draft_tokens` is validated as a positive integer
  or null, and — being an engine-construction control — a change also
  triggers the existing auto-reload path.
- Switch `ModelSettingsRequest` to `extra="forbid"`: unknown fields now
  fail with a 422 instead of silently returning `success: true`. The admin
  frontend sends exactly the declared fields (verified: all 65 payload
  keys are within the schema), so this cannot break the UI — it turns
  future "added to the engine but never registered on the API" mistakes
  into loud errors.

Verification:

- New unit tests: persistence for all three fields, positive-int
  validation, loud rejection of unknown keys. `tests/test_admin_model_settings.py`
  passes 19/19.
- Adjacent admin suites: 92 passed.
- Full suite with the CI filter (`pytest tests/ -m "not slow and not
  integration"`): 10439 passed. The 6 failures reproduce unchanged on
  clean `main` (cluster SSH supervisor, cluster seams, VLM MTP e2e —
  environment-bound) or are flaky (2 GDN sidecar tests pass on rerun);
  none touch this change.
- Live smoke against a server run from this branch: PUT
  `mtp_num_draft_tokens=8` → returned in the response **and** persisted in
  `model_settings.json`; unknown key → 422; `mtp=0` → 400; omitted fields
  are left untouched.

Closes #2823

### DIFF
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -2395,7 +2395,12 @@ async def update_model_settings(
     if "turboquant_kv_bits" in sent:
         current_settings.turboquant_kv_bits = request.turboquant_kv_bits or 4
     if "turboquant_skip_last" in sent:
-        current_settings.turboquant_skip_last = bool(request.turboquant_skip_last)
+        # null = clear to the model default (True); bool(None) would flip it
+        # to False and silently disable the skip-last corruption guard.
+        current_settings.turboquant_skip_last = (
+            True if request.turboquant_skip_last is None
+            else bool(request.turboquant_skip_last)
+        )
     # Private Qwen3.5/3.6/3.8 ANE/GPU fixed-shape prefill. These are all load-time
     # controls; the runtime signature below causes a loaded model to be
     # re-created when the user applies a changed profile.
@@ -2930,8 +2935,6 @@ async def update_model_settings(
         or "dflash_in_memory_cache_max_bytes" in sent
         or "dflash_ssd_cache" in sent
         or "dflash_ssd_cache_max_bytes" in sent
-        # MTP draft token count is an engine-construction control.
-        or "mtp_num_draft_tokens" in sent
         # trust_remote_code is plumbed at model load time; toggling it on
         # an already-loaded engine has no effect until reload.
         or "trust_remote_code" in sent
--- a/omlx/admin/routes.py
+++ b/omlx/admin/routes.py
@@ -28,7 +28,7 @@ import requests
 from fastapi import APIRouter, Depends, HTTPException, Request, Response
 from fastapi.responses import FileResponse, HTMLResponse, JSONResponse, RedirectResponse
 from fastapi.templating import Jinja2Templates
-from pydantic import BaseModel, Field, field_validator
+from pydantic import BaseModel, ConfigDict, Field, field_validator
 
 from ..api.markitdown import MARKITDOWN_MODEL_ID, markitdown_model_visible
 from ..api.openai_models import _coerce_tool_call_arguments
@@ -112,6 +112,8 @@ class CacheProbeRequest(BaseModel):
 class ModelSettingsRequest(BaseModel):
     """Request model for updating per-model settings."""
 
+    model_config = ConfigDict(extra="forbid")
+
     model_alias: str | None = None
     model_type_override: str | None = None
     max_context_window: int | None = None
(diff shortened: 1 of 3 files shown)


jundot:
Thanks, the missing request fields and `extra="forbid"` fix the silent-drop bug. I need two changes before merge: include `mtp_num_draft_tokens` in `_engine_runtime_signature` only when Lightning MTP is active, and preserve the `turboquant_skip_last=True` default when null is sent. With those corrections and focused tests covering both cases, I’m comfortable merging this.

yanzhaohui1999:
Thanks for the fast review — both requested changes are in:

- `mtp_num_draft_tokens` is now part of `_engine_runtime_signature`, gated on `mtp_enabled` (Lightning MTP active), so a depth change reloads a loaded engine but a stale value never forces one when MTP is off; the manual `requires_reload` entry was removed in favor of the signature.
- `turboquant_skip_last=null` now restores the model default (`True`) instead of `bool(None)` disabling the skip-last guard.

Focused tests cover both cases (null-default preservation + signature gating with active/inactive MTP), and the full suite passes (10441 passed; the 6 failures reproduce on clean `main` — environment-bound).

Note: the core `mtp_num_draft_tokens` fix overlaps with #3068; this PR additionally covers `preserve_thinking` + `turboquant_skip_last` and `extra="forbid"`. Happy to trim/rebase if you prefer #3068 for the core.

jundot:
Thanks, both requested follow-ups look correct. The runtime signature now reloads the engine only when Lightning MTP is active, and `turboquant_skip_last=null` correctly preserves the safe default.

One thing I’ll fold into a follow-up is constraining `mtp_num_draft_tokens` to the runtime-supported range of 1–8 so the persisted value always matches the effective depth.

This overlaps with #3068, so I’ll preserve that contributor’s attribution in the squash commit. LGTM, merging.

PR #3281, opened 2026-08-29
fix(chat): promote "Regenerate with <model>" to the active model

### LINKED ISSUE #2729: Chat UI keeps failed model name after switch; bottom status shows correct active model
**Describe the bug**  
When a model fails to load and you switch/regenerate with another model, the chat header still shows the failed model name on the next message. The bottom-left status correctly shows the newly active model.

**To Reproduce**  
1. Start a chat and select a model that fails to load.  
2. Switch to a different working model and regenerate / send a new message.  
3. Observe chat header still displays the failed model name.  
4. Bottom-left status shows the correct active model.

**Expected behavior**  
Chat header model name should update immediately to the newly selected/active model after switch.

**Screenshots**  

Image

**Environment**  
- oMLX version:  0.6.0
- macOS version: 26.6.1

yanzhaohui1999:
## Summary

When a model failed to load and the user regenerated with a different
model via the "Regenerate with" dropdown, the floating header/picker
kept showing the failed model. The one-off `_modelOverride` streamed
with the right model (so the reply's model label was correct), but it
never updated `currentModel` or `session.model`, and the call reused the
failed model's stale generation settings.

Changes (in `regenerateMessage`):

- An explicit `opts.model` is now promoted to the chat's active model:
  `currentModel` and `session.model` are both updated, so the header
  follows the model that actually responds, and later sends / reopening
  the chat stay on it (the per-model settings map is keyed by session
  model, so settings for the regenerated-with model are loaded before
  streaming instead of reusing the previous model's).
- The "one-off override never mutates saved settings" invariant is
  preserved: only the active-model pointer and its settings view change,
  never the stored per-model settings.

Verified:

- New regression test in `tests/test_chat_ui_overhaul.py` asserts the
  promotion and settings-load logic in `regenerateMessage`.
- Chat-related suites: 53 passed; chat.html inline JS passes `node
  --check` (Jinja stripped).
- Full suite with the CI filter: 10440 passed; the 6 failures reproduce
  unchanged on clean `main` (cluster SSH supervisor, cluster seams, VLM
  MTP e2e — environment-bound) or are flaky (2 GDN sidecar tests pass on
  rerun); none relate to this change.

Closes #2729

### DIFF
--- a/omlx/admin/templates/chat.html
+++ b/omlx/admin/templates/chat.html
@@ -6691,6 +6691,21 @@
             generationOverride = { temperature: temp };
         }
 
+        // "Regenerate with <model>" promotes the override to this chat's active
+        // model (fixes #2729): the header/picker must follow the model that
+        // actually generates the reply, and the per-model settings for that
+        // model must feed the call instead of the stale settings of the model
+        // that previously failed to load.
+        if (opts.model) {
+            const gatewayId = this.resolveGatewayModelId(opts.model);
+            this.currentModel = gatewayId;
+            session.model = gatewayId;
+            if (!this.loadModelSettingsForModel(session, gatewayId)) {
+                await this.ensureSessionModelSettings(session, gatewayId);
+                this.saveModelSettingsForModel(session, gatewayId);
+            }
+        }
+
         await this.streamResponse({
             chatId: this.currentChatId,
             model: this.currentModel,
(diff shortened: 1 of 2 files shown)


scubamount:
## Verified

- **Root cause correct:** "Regenerate with <model>" streamed with the override but never updated `currentModel`/`session.model`, so the header kept the failed model and the call reused its stale per-model settings.
- **Fix right:** explicit `opts.model` promoted via `resolveGatewayModelId`, both `currentModel` and `session.model` updated, settings loaded (with ensure+save fallback when absent for that model) before `streamResponse`.
- **Test:** `test_regenerate_with_model_promotes_override_to_active_model` asserts all five calls appear in the extracted section.

No issues.

PR #3283, opened 2026-08-29
fix(qwen4-exp): resolve prefill memory blowup for QSA long-context

alytaphoenix:
## Summary

Two independent fixes for Qwen4-Exp (Qwen3.8-Flash-Next) long-context prefill, which previously hard-aborted around ~43k tokens on a 64GB Mac despite the architecture's 262144-token native context window. After both fixes, verified ceiling on the same hardware is ~96,000-100,000 tokens clean (~2x).

- **`sdpa256_attention.py`**: teach the tiled/streaming SDPA routes to accept boolean array masks (QSA's sparse-attention indexer produces one past ~2048 tokens). Previously any non-causal array mask was routed to stock MLX's unfused O(L·T) fallback, materializing full `[heads, L, T]` score slabs that scale linearly with context and dominated the memory budget. Also disables q-split for array masks (it can't narrow keys the way causal masks allow; combined with the admission hysteresis ratchet this thrashed the Metal buffer pool).
- **`turboquant_attention.py`**: try the tiled `quantized_attention` route before the slab-building `prefill_attention` above the long-prefill threshold, matching the ordering the scheduler's cost estimator already assumes.
- **`scheduler.py` / `boundary_snapshot_store.py`**: `QSAKVCache` and `QSAQuantizedKVCache` were missing from `_KNOWN_SLICEABLE_CACHE_TYPES`, so every boundary/prompt-cache snapshot captured the model's full K/V+index prefix instead of an incremental slice — quadratic cost in context length (~1.1GB per snapshot at 40k tokens). This was the dominant remaining cost after the sdpa256 fix alone, and explains why the measured ceiling after that fix undershot the initial prediction. Also adds a diagnostic warning for any future sliceable-cache-type omission of the same shape, plus a lockstep test against the handler registry.

The first commit on this branch squash-ports pre-existing work (q-split prefill fast path, hysteresis ratchet, TurboQuant dense-fallback fix, ANE-prefill-aware paged SSD cache signatures) that had accumulated on our internal fork but was never itself upstreamed — it's a prerequisite for the actual fix commit to apply cleanly and is a pure port, no new behavior.

## Test plan

- [x] Full existing test suite: 10,524 passed, 0 failed, 106 skipped (pre-existing skips), 77 deselected (slow/integration, as configured)
- [x] 8 new bool-mask numerics tests for the sdpa256 tiled-mask routing, including the QSA-specific edge case of a row whose first tile is fully masked (compared against dense fp32 reference, ≤5e-3 tolerance)
- [x] New regression-guard test (`test_boundary_snapshot_qsa.py`) locking `_KNOWN_SLICEABLE_CACHE_TYPES` to the handler registry so the next sliceable cache class can't silently regress the same way
- [x] Live validation on real hardware (M4 Pro, 64GB): empirical stepped-context ladder with unique/non-cached prompt content at each step (no prefix-cache reuse), before and after each fix, confirming the measured ceiling improvement end-to-end — clean prefill at 43k/58k/77k/96k tokens, clean guard rejection (no crash) at ~125k
- [x] Coherence check: needle-in-prose retrieval correct at 6k, 10k, and 41,529 tokens through the new tiled routing path

## Notes for reviewers

- `Fix 2` (real KV quantization for QSA caches, to push toward the full 262k native ceiling) was scoped and investigated but not implemented — the naive approach (call the vendored `to_quantized()`) doesn't work: the quantized cache class has no `merge()` (breaks batch generator insert) and MTP verify slices tuple keys (breaks, and MTP is on for this model family). Left as follow-up.
- The prerequisite-port commit is large but mechanical (a direct diff replay of already-reviewed internal-fork history) — the actual new logic is entirely in the second commit.

https://claude.ai/code/session_5f72bb11-5f96-454f-8071-e7412e896f13

### DIFF
--- a/omlx/cache/boundary_snapshot_store.py
+++ b/omlx/cache/boundary_snapshot_store.py
@@ -97,8 +97,37 @@ def _gdn_rht_sign_values(dim: int, seed: int) -> tuple[float, ...]:
     )
 
 
+def _session_owner_pid_is_live(pid: int) -> bool:
+    """Is the process that owns a boundary-snapshot session still running?
+
+    Mirrors ``cluster/liveness.py``'s ``_pid_is_live``, kept as a local
+    copy rather than an import so this module carries no dependency on
+    the optional cluster package — the same reasoning that module states
+    for keeping its own copy local. An unreadable pid is treated as live:
+    refusing to reap is the safe answer when we cannot tell.
+    """
+    try:
+        os.kill(pid, 0)
+    except ProcessLookupError:
+        return False
+    except (PermissionError, OverflowError, TypeError, ValueError):
+        return True
+    return True
+
+
 def reset_boundary_snapshot_root(base_dir: Path) -> None:
-    """Remove all boundary snapshot sessions for a server lifecycle boundary."""
+    """Remove dead-owner boundary snapshot sessions for a lifecycle boundary.
+
+    Per-session-aware (design doc §B2): a blanket rmtree strips every
+    session under this root, including one belonging to a *second* oMLX
+    process on this Mac (a server on another port, or a start racing a
+    slow shutdown) whose request is still in flight — the reset ignored
+    that sessions are per-process (``<pid>-<uuid>``) precisely so multiple
+    processes can coexist. Delete only sessions whose owning pid is dead;
+    in the common single-process case every existing session's owner is
+    necessarily dead (a fresh start has no other live line), so this
+    reduces to exactly the prior blanket-rmtree behavior.
+    """
     snapshot_root = base_dir / "_boundary_snapshots"
     if snapshot_root.is_symlink():
         logger.warning(
@@ -106,11 +135,29 @@ def reset_boundary_snapshot_root(base_dir: Path) -> None:
             snapshot_root,
         )
         return
-    if snapshot_root.exists():
-        try:
-            shutil.rmtree(snapshot_root)
-        except Exception as e:
-            logger.warning("Failed to reset boundary snapshots: %s", e)
+    if snapshot_root.is_dir():
+        for session_dir in list(snapshot_root.iterdir()):
+            if session_dir.is_symlink() or not session_dir.is_dir():
+                continue
+            pid_str, _, _rest = session_dir.name.partition("-")
+            try:
+                pid = int(pid_str)
+            except ValueError:
+                # This module only ever names a session "<pid>-<uuid>hex>";
+                # anything else was never a live session under this
+                # convention and is foreign debris, same treatment as a
+                # malformed name anywhere else in the cache tree (§A1).
+                pid = None
+            if pid is not None and _session_owner_pid_is_live(pid):
+                continue
+            try:
+                shutil.rmtree(session_dir)
+            except Exception as e:
+                logger.warning(
+                    "Failed to reap dead boundary snapshot session %s: %s",
+                    session_dir,
+                    e,
+                )
     snapshot_root.mkdir(parents=True, exist_ok=True)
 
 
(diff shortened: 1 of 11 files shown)


jundot:
Thanks for tracing the QSA long-prefill memory growth and contributing the core fixes. I incorporated the focused changes into main in `9934b288`: boolean QSA masks can use tiled SDPA, long TurboQuant prefills take the tiled path first, and sliceable QSA caches are excluded from boundary snapshots. The integration commit explicitly references #3283 so the provenance of your contribution remains attached to this PR.

I left out the broader deploy-branch SSD, ANE, and scheduler port because it includes unrelated behavior and several paths that are not fully wired. Rebasing this PR onto current main would therefore leave mostly those unrelated changes rather than the original fix.

If there are remaining changes you believe should still land, please split them into focused follow-up PRs based on current main. I’ll leave this open for now so you can respond.

alytaphoenix:
Pushed one additional commit (`fae1d3d3`) while preparing a follow-up PR that builds on this branch: found that `scheduler.py`'s `memory_monitor.register_tiled_prefill_tile()` / `clear_tiled_prefill_tile()` calls (already present on this branch) had no corresponding method definitions in `memory_monitor.py` — a gap from this PR's original prerequisite-port scope missing that file. Squash-ported the missing `_tiled_prefill_override` machinery from the internal fork (pure port, no behavior change vs. the fork's current state). Full suite still green after the fix.

scubamount:
## Review

- **Fix 1 (sdpa256):** tiled/streaming SDPA routes now accept boolean array masks (QSA's indexer produces one past ~2048 tokens); previously any non-causal array mask fell to stock MLX's unfused O(L·T) path, materializing full [heads, L, T] score slabs that dominated the memory budget — this was the blowup. Array masks also disable q-split (it cannot narrow keys the way causal masks allow), which is correct reasoning: q-split under an admission hysteresis would undersample key coverage.
- **Fix 2:** second, independent memory path (scheduler/monitor).

Caveats: 2889-line change touching 7 core files (cache, memory_monitor, scheduler, two attention patches) with no CI run recorded yet — this is the kind of wide, memory-critical change where each file's edits should be reviewed somewhat separately despite being one PR. The measured claim (~43k→~96-100k token ceiling on the same 64 GB machine, verified clean) is strong evidence the combined effect is real and lossless; confirming losslessness (identical outputs pre/post on the same prompts) in the test suite is the thing I'd want green before merge.

Not blocked — flagging the surface area and the missing CI run.

alytaphoenix:
Closing this out — the work here has been resolved and split up as requested.

- **Core fix landed:** the QSA long-prefill memory bound (Fix 1/3/4, the
  context-ceiling work) merged directly to `main` via 9934b288.
- **memory_monitor tiled-prefill guard-registration gap:** already covered by
  the separate, independently-open PR #3108
  (`fix(memory-guard): register TurboQuant's tiled prefill transient
  per-instance`), which is a superset of what this branch's addendum did
  (skip-last handling, fp32 pricing, scheduler callers, tests). No new PR is
  needed for it — nothing further required here.
- **Remainder of this branch** (paged_ssd_cache extras, sdpa256 extras,
  boundary_snapshot, turboquant_kv additions, and the deploy-branch
  engine_pool/vlm/enforcer/deployment port) was deploy-branch-only material that
  was already explicitly declined, and is not being re-proposed at this time.

The two remaining wanted changes are now up as focused, single-commit PRs
targeting current `main` directly:

- **Fix 2 — real KV quantization for QSA caches:** #3436 (full review history on
  alytaphoenix/omlx#1)
- **MoE routed-expert weight streaming:** #3437 (full review history on
  alytaphoenix/omlx#2)

Both were rebuilt onto current `main` and re-verified against the full suite
(zero regressions vs. the `main` baseline). The old deploy-branch prereq commit
was dropped as unnecessary — `main` already carries the symbols it depended on.
One semantic-drift fixup was needed on the Fix 2 side (a later accounting
refactor had inlined a condition the rebased block still referenced by name);
it's called out in #3436's description. Closing this in favor of those two.

Issue #2972, opened 2026-08-21
Process-global Qwen patches contaminate a resident DFlash engine after loading a Lightning MTP model


LKyaha:
## Summary

Extended multi-model testing for #2962 uncovered process-local state contamination between a resident Qwen3.8 DFlash engine and a second Qwen3.8 engine using Lightning MTP.

The DFlash target is stable after a clean oMLX start. If another Qwen3.8 model with Lightning MTP is loaded in the same process while the DFlash engine remains resident, subsequent DFlash generations can enter hard repetitive loops. Unloading the second model does not restore the original behavior; only a full oMLX process restart does.

This was found while validating #2962, but the current evidence does **not** show that the loader change in #2962 causes the contamination. The single-target DFlash path remains stable.

## Environment

- oMLX `0.6.3rc1` (build `2388`) with the #2962 changes applied
- DFlash target: `unsloth/Qwen3.8-27B-NVFP4`
- Drafter: `z-lab/Qwen3.8-27B-DFlash2`
- Second model: Qwen3.8 27B AWQ with Lightning MTP enabled
- Apple Silicon macOS

## Reproduction

1. Fully quit and restart oMLX.
2. Load `unsloth/Qwen3.8-27B-NVFP4` with `z-lab/Qwen3.8-27B-DFlash2`.
3. Run the same long-form writing prompt several times. No hard repetition loop is observed.
4. Keep that DFlash model resident and load a second Qwen3.8 27B AWQ model with Lightning MTP enabled.
5. Generate again with the original DFlash target.
6. Unload the AWQ model and repeat generation with the DFlash target.

## Actual behavior

- Before loading the second model: 1 x 1024-token and 3 x 2048-token DFlash runs completed without a long-cycle repetition loop.
- With the second model loaded: 2 of 3 short runs entered hard loops, repeating short substrings 52 and 71 times.
- After unloading the second model: all 3 follow-up runs still looped, repeating different short substrings 44, 90, and 50 times.
- A full oMLX process restart restored the clean behavior.

The original prompt and output substrings are omitted for privacy. Exact raw samples can be provided privately if required.

## Expected behavior

Loading or unloading a second Qwen/Lightning-MTP engine should not mutate the behavior of an already resident DFlash engine. Engine-specific patches and lifecycle state should be isolated or fully restored.

## Initial diagnosis

[...]

Issue #2252, opened 2026-07-15
Broken-load DFlash helper poisons concurrent requests to an unrelated model (Qwen3-Next, batched engine) — #1510-class patch leak on 0.5.1


theuseless-ai:
## Summary

A DFlash draft model that **fails to load** still poisons the engine of an **unrelated model**: with a broken-load DFlash registered as helper for `Qwen3.6-35B-A3B-bf16`, any **concurrent** (≥2) requests to `Qwen3-Coder-Next-MLX-8bit` (batched engine) fail — solo requests work. Detaching/deleting DFlash fixes it immediately, with no other change.

This looks like the class-patch leak family of #1510 (which was fixed for the loaded-DFlash case) — but triggered by a DFlash that *never successfully loads*, so its hooks apparently still patch shared Qwen model classes. Filing as a new issue since #1510 is closed and this is a different trigger path.

## Environment

- oMLX **0.5.1** (menu-bar app), Apple Silicon M3 Ultra (512GB; `final_ceiling` ≈ 498GB)
- mlx-lm **0.31.3** (includes ml-explore/mlx-lm#1169; verified by tag — that fix did not help here, different layer)
- Target model: `lmstudio-community/Qwen3-Coder-Next-MLX-8bit` (engine `batched`, loaded, pinned)
- Bystander model: `Qwen3.6-35B-A3B-bf16` (engine `vlm`, loaded, pinned) — **never affected**
- Draft: `z-lab/Qwen3.6-35B-A3B-DFlash`, registered as helper for Qwen3.6, but permanently in load-failure state (#873):
  ```
  Model 'Qwen3.6-35B-A3B-DFlash' is unavailable after a previous load failure:
  Received 91 parameters not in model: fc.weight, hidden_norm.weight, layers.0. ...
  ```

## Repro (while broken-load DFlash is registered as helper)

Two identical, non-streaming requests in parallel to the *Coder* model:

```sh
req() { curl -s -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' \
  -d '{"model":"Qwen3-Coder-Next-MLX-8bit","messages":[{"role":"user","content":"Count 1 to 10"}],"max_tokens":60}' \
  http://HOST:8080/v1/chat/completions; }
req            # solo: 200 OK, correct output
(req & req &)  # concurrent: both fail
```

Observed failure modes (same day, same box):

1. API error surfaced to clients:
   ```
[...]

PR #3285, opened 2026-08-29
fix: isolate cross-engine monkey-patches between DFlash and Lightning MTP (#2972)

e-Evolution:
Fixes the cross-engine contamination reported in #2972, in **both directions**. Full investigation, deterministic reproduction recipes, and validation evidence are in the issue thread ([repro](https://github.com/jundot/omlx/issues/2972#issuecomment-5460717869), [leg 1 root cause](https://github.com/jundot/omlx/issues/2972#issuecomment-5460815228), [leg 2 root cause](https://github.com/jundot/omlx/issues/2972#issuecomment-5461101649)).

## Leg 1 — Lightning MTP load degenerates a resident DFlash engine

The `GatedDeltaNet` self-heal in `mlx_lm_mtp/qwen35_model.py` didn't recognize the dflash batch-cache guard (#2252) on `cls.__call__` and clobbered it, silently stripping the resident DFlash engine's speculative hook → hard repetition loops. The stale dflash idempotency flag then made **new** DFlash engines skip hook installation, so they were broken from birth.

**Fix (compose, don't clobber):**
- `guarded_call` resolves its fallback base dynamically from `_DFLASH_BACKUP` and routes `n_confirmed` calls to the base
- the MTP self-heal installs its body as the guard's fallback base (`swap_backup_base`) when dflash owns the class
- `_wrap_installer` repairs the stale idempotency flag

## Leg 2 — Qwen3.5 VLM load breaks qwen4_exp Lightning MTP

The `qwen4_exp` vendor `LanguageModel` inherits from the q35 `LanguageModel` and reaches stock behavior via `super().__call__`. Loading any q35 VLM model (a load alone suffices) runs `_patch_vlm_language_model`, whose replacement `__call__` pops the caller's `capture_layer_ids=[]` and forces `[last_layer_idx]` — collapsing the raw hyper-stream hidden `(B, T, hc_count·H)` the Qwen4 MTP head requires down to `(B, T, H)` → `Qwen4 Lightning MTP expects hidden shape [...]`. The patched `__init__` would also bolt a q35 `MTPModule` onto the foreign architecture.

**Fix:** exact-type guards (`type(self) is not cls`) in the patched `__init__`/`__call__` — the patch serves plain q35 instances; subclasses keep stock behavior.

## Validation (M5 Max 128 GB, 0.6.3 stable + these patches)

- Both previously deterministic reproductions pass.
- Interleaving a DFlash target (`Qwen3.8-27B-8bit` + DFlash2 draft), a q35 Lightning-MTP model (`Qwen3.8-27B-oQ4e-mtp`), and a qwen4_exp model (`Qwen3.8-Flash-Next-oQ4e-mtp`) in **one process, no restarts** stays clean in all directions: zero degeneration across 12+ DFlash generations, MTP acceptance 75–92%, speculative speeds preserved.
- Single-engine setups are behaviorally unchanged (guards only alter the multi-engine paths).

### DIFF
--- a/omlx/patches/dflash_lifecycle.py
+++ b/omlx/patches/dflash_lifecycle.py
@@ -65,6 +65,20 @@ def _wrap_installer(mod: Any, fn_name: str, flag_name: str) -> bool:
 
     def wrapped(module_target: Any) -> Any:
         cls = type(module_target)
+        current = cls.__dict__.get("__call__")
+        if getattr(cls, flag_name, False) and getattr(
+            current, "_omlx_mtp_call_marker", False
+        ):
+            # A Native-MTP self-heal replaced dflash's hook while the
+            # idempotency flag stayed set (issue #2972): dflash's installer
+            # would skip re-installing and the engine would run without its
+            # speculative hook. Clear the stale flag and re-snapshot the
+            # current (MTP) __call__ as the base to restore/fall back to.
+            try:
+                delattr(cls, flag_name)
+            except AttributeError:
+                pass
+            _DFLASH_BACKUP[cls] = {"call": current, "flag": flag_name}
         if not getattr(cls, flag_name, False):
             # First time dflash installs on this class — snapshot the
             # current __call__ so restore can put it back unchanged.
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for tracing and fixing both directions of the process-global patch collision. I merged this because it isolates Qwen4 from the Qwen3.5 runtime patch while preserving the DFlash guard during Lightning MTP repatching.

I followed up in 5463c0f4 with simpler helpers, stricter state checks, and regression coverage. Qwen3.5, Qwen4, and DFlash now interleave cleanly in one server process with Lightning MTP active.

PR #3287, opened 2026-08-29
fix: use concurrent pread for Qwen4-Exp PLE row lookups (~10x SSD throughput)

scott-kim-7:
## Summary

`_SafeTensorMMap.rows()` (used by the SSD-backed PLE / n-gram bank path for Qwen4-Exp, e.g. Qwen3.8-Flash-Next) gathers rows with numpy fancy-indexing over an `mmap`'d buffer opened with `MADV_RANDOM`. Each row that isn't resident blocks synchronously on a page fault, and the gather walks the index array in a single thread — so N row lookups serialize into N blocking page faults with zero overlap, even though the underlying SSD has plenty of parallelism to hide that latency.

This PR keeps the mmap gather and puts a **cold-page prefetch** in front of it. A per-reader seen-page bitmap records which pages this process has already pulled in. The first time a gather needs pages that have never been touched, those pages are read concurrently through a thread pool of `os.pread()` calls (threads rather than asyncio/io_uring, because `os.pread` releases the GIL for the syscall duration), and the gather then runs through the original numpy fancy-index. A fully-warm call pays only a bitmap probe, so a resident page keeps `main`'s memcpy speed.

**Credit:** the root-cause analysis and the concurrent-`pread` thread pool are mine; the cold-page gate is @williamxie1989's work, contributed as scott-kim-7/omlx#1 and folded in here. The first iteration of this PR used `pread` as the data path for *every* row, which cost ~21 % of warm prefill end to end — the gate removes that while keeping the cold-path parallelism.

## Measurements

**End to end, warm path.** Native oMLX path (`OMLX_USE_BUNDLE=0`), oMLX 0.6.4 / mlx 0.32.0 with custom kernels built, M4 Max 128 GB, `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp`, 12.2k-token `code_python` prompts, 200 tokens generated, prefix-cache miss on every request. Decode timing is read from the server's own `timing[backbone=…]` MTP log rather than by subtraction. PLE table 20–40 % resident.

| arm | n | prefill s | prefill tok/s | decode tok/s |
|---|---:|---:|---:|---:|
| `main` | 3 | 6.95 | 1761 | 40.6 |
| **this PR** | 2 | **6.84** | **1790** | 41.7 |
| per-row `pread` (first iteration) | 3 | 8.42 | 1455 | 40.5 |

**Gather microbenchmark**, 4096 random rows of a 316 MB shard, all pages resident:

| gather path | µs/row |
|---|---:|
| `main` (mmap fancy-index) | 0.018 |
| per-row `pread` (first iteration) | 7.99 |
| **this PR** (warm hit) | **0.085** |

**Isolated random-read throughput** (M4 Max, internal NVMe) — this is the parallelism the cold path exploits, and the basis for the 48-worker default:

```
workers=  1    8,827 IOPS   (113.3us avg latency)
workers=  8   61,510 IOPS
workers= 16   89,028 IOPS
workers= 24  109,701 IOPS
workers= 32  110,664 IOPS
workers= 48  115,185 IOPS   <- peak; plateau spans 24-64
workers= 64  109,417 IOPS
workers= 96  102,239 IOPS   <- throughput regresses past here
workers=256   72,091 IOPS
```

**Cold path: re-measurement in progress.** The cold TTFT figures in the first version of this description were taken through a model-bundled `omlx_support` compat shim rather than oMLX's own reader. I am re-running them on the native path under the protocol @latent-variable used in #3235 (per-arm `purge`, fresh process, model load, warmup, one measured prompt, arms alternating) and will post the results in the thread whatever they show.

## Correctness

Read the same random row indices through the original and patched `_SafeTensorMMap` against real PLE shard files and compared output with `np.array_equal` — byte-identical. The gate does not change what is read: `pread` is used only to fault pages in and its result is discarded; every value still comes from the original mmap fancy-index. A regression test covers a multi-page gather with rows straddling page boundaries.

## Scope

- Only affects `OMLX_QWEN4_PLE_MODE=mmap` (or equivalent SSD-backed PLE offload) — `resident` mode doesn't go through this code path.
- Only touches `_SafeTensorMMap`: no change to the on-disk format, header parsing, BF16/F8_E4M3 unpack, or the mmap setup used elsewhere in the class.
- Prefetch is skipped for gathers of 8 rows or fewer, so decode-sized calls take the unmodified path.
- Memory cost of the bitmap is 1 byte per page, ~320 KB per 5 GB shard.
- Related work in the same area: #3372 (batched cross-shard gather) and #3235 (batched `madvise`, closed by its author).

## Test plan

- [x] Byte-identity check against real PLE shards, original vs patched reader
- [x] `test_disk_backed_ple_page_prefetch_returns_identical_rows` — multi-page gather with straddling rows
- [x] End-to-end warm A/B on the native path (numbers above)
- [ ] Cold-path A/B on the native path — in progress
- [ ] Re-run of `tests/test_mlx_vlm_qwen4_exp_compat.py`, `test_vlm_qwen4_exp_loader.py`, `test_qwen4_exp_residency.py`, `test_qwen4_exp_mtp_admin_gate.py` after the merge with scott-kim-7/omlx#1


scott-kim-7:
Follow-up on the thread-pool size: I picked 48 originally from just two data points (32 and 64 threads, which looked similar), not a proper sweep. Ran a fuller sweep since to check that wasn't a lucky/unlucky guess:

```
workers=   1    8,827 IOPS   (113.3us avg latency)
workers=   4   33,352 IOPS
workers=   8   61,510 IOPS
workers=  16   89,028 IOPS
workers=  24  109,701 IOPS
workers=  32  110,664 IOPS
workers=  48  115,185 IOPS   <- peak in this run
workers=  64  109,417 IOPS
workers=  96  102,239 IOPS   <- throughput starts regressing past here
workers= 128   98,213 IOPS
workers= 192   78,873 IOPS
workers= 256   72,091 IOPS
```

(Same M4 Max / internal NVMe, same shard file, 3000 random 4KB reads per point.)

Shape: throughput rises ~linearly up to ~24 workers, plateaus 24-64 (within
~5% of each other — likely the SSD controller's realistic internal queue
depth/channel count), then *regresses* past ~64 as thread/kernel scheduling
overhead outweighs the added parallelism. 48 lands inside the plateau with
comfortable margin from the regression zone on both sides, so I'm keeping
it, but wanted to post the actual curve rather than leave it as "seemed
fine in two spot checks." Happy to make it configurable (env var or
constant near the top of the file) if that's preferred over a hardcoded
default — the plateau is wide enough (24-64) that I'd expect most Apple
Silicon internal SSDs to tolerate 48 fine, but I've only measured this on
one machine.

RobinBially:
**Independent 96 GB data point, plus a second-order effect this fix should also remove**

Setup: M3 Ultra 96 GB (60-core GPU), macOS 27.0, oMLX 0.6.3 → 0.6.4, `Vontra/Qwen3.8-Flash-Next-MLX-oQ3-MTP` (62–63 GiB resident) with `qwen4_ple_ssd_offload=true`, `iogpu.wired_limit_mb=93388`. Batch 1, temperature 0, 400 generated tokens, ~4k-token prompts with a changed prefix per run (prefix-cache miss, PLE rows warm).

**1. Reproduces the serial page-fault cost.** Isolated `DiskBackedShardedEmbedding.__call__` against the checkpoint, 1024 tokens × 16 heads, rows verified cold: **4.2 ms/token**; same rows warm: 0.04 ms/token. Single token cold: 6.6–7.4 ms. In-server this is the whole cold prefill: 3965 tokens → TTFT 17.1 s (231 tok/s); same document with a different prefix afterwards → 5.4 s (731 tok/s). After sequentially reading the 23.8 GiB of PLE tensors into the page cache (5 s at ~4.6 GiB/s) a fresh-prefix 4k prompt prefills at 620–780 tok/s.

**2. The decode-side cost is larger than 16 reads/token would suggest, because of how macOS handles the idle GPU.** Sampling `vm_stat` + `ioreg` GPU utilization every 0.25 s during decode:

| run | wired during decode | dips <60 GiB | decode |
|---|---|---|---|
| typical | oscillates 31–74 GiB | 5–7 per 400 tok | 21–27 tok/s |
| occasional | flat ~67–70 GiB | 0 | 34–35 tok/s (MTP off) |

Every dip coincides with GPU utilization dropping to 0–5 %. The picture: the host-side row gather stalls the pipeline, the GPU goes idle, the driver un-wires 20–35 GiB of Metal buffers, and the next command buffer pays for re-wiring them. The re-wire, not the 16 reads, is what makes a single stall cost hundreds of ms. Three things I ruled out as the trigger: free memory (`sudo purge` → 18 GiB free, dips unchanged), `iogpu.disable_wired_collector=1` (unchanged), and MTP (dips with MTP on and off). The only condition that produced flat wired + full speed was a warm PLE cache. `sample` of the server process during decode shows the main thread in `mlx::core::array::wait` and `IOGPUMetalResidencySet`/`IOGPUResourceListAddResource` frames — re-wiring, not compute.

For reference, `llama.cpp` b10679 on the same machine with `unsloth/Qwen3.8-Flash-Next-UD-IQ1_S` (table resident as `CPU_Mapped`, `ggml_get_rows` in C): wired flat at 74.7 GiB, zero dips, 33.5–33.9 tok/s across three runs within 0.4 tok/s; cold == warm prefill. The backbone step time is comparable to oMLX's when oMLX is not stalling (oMLX 0.6.4 native decode ~34.5 tok/s in dip-free runs), so the gap on 96 GB boxes is essentially this gather.

**3. Why this matters for the PR.** Overlapping the faults doesn't just cut gather time; it shortens the idle window enough that the un-wire/re-wire cascade should stop firing, which is where the run-to-run variance (18–37 tok/s for an identical configuration) comes from. That would also explain why your decode gain (+37 %) is much larger than 16 × 0.1 ms per token can account for. #3235 attacks the same window from the other side (batched `madvise` before the reads) and should have the same secondary effect.

A small prototype of the `pread` approach here plateaued at ~2.7× (5.4 → 2.0 ms/token cold) with 16–128 Python threads because I read via mmap slices; your `os.pread` numbers (10–12×) are consistent with the GIL being released for the whole syscall, so no objection to the approach — just noting the mmap-slice variant is not a substitute.

Happy to re-run the exact same series (`vm_stat` trace + decode/prefill, warm and cold) on this branch if useful.

scubamount:
## Verified

- **Root cause correct:** mmap fancy-index gather in `_SafeTensorMMap.rows()` serializes N row lookups into N synchronous page faults (~100 us each on Fabric SSD, `MADV_RANDOM`) with zero overlap.
- **Fix right:** 48-worker `ThreadPoolExecutor` of `os.pread()` calls on the same fd — pread is positional/thread-safe and releases the GIL for the syscall duration, letting the SSD controller overlap the waits. Single-row and empty cases bypass the pool.
- **Correctness:** BF16 widen path unchanged; affine/dense layout validation still applies before reads.

Nits (non-blocking):
- `os.pread` can return short reads at EOF; a truncated row would silently produce a wrong-length `frombuffer`. Practically unreachable for in-range rows, but a length check would close the hole.
- Module-level pool is created at import (48 threads per process) and never shut down; fine for a server, worth knowing.

williamxie1989:
@scott-kim-7 Great catch on the serialized page-fault root cause — the concurrent pread direction is exactly right. I forked your branch and kept your 48-worker pool and the GIL rationale, but hit a second-order effect on **warm** pages: per-row `pread` + futures + `np.stack` replaces what used to be a near-free mmap memcpy once the page cache is hot.

Microbenchmark, 4096 random rows of a 316 MB shard, page cache warm (M5 Max 128 GB, `Qwen3.8-Flash-Next-oQ4e-mtp`, PLE on SSD):

| gather path | µs/row |
|---|---|
| main (mmap fancy-index) | 0.04 |
| this PR (per-row pread) | 22.1 |
| cold-only gate (below) | 0.14 |

End-to-end this matches what I saw after your fix: the first prefill batch is fast (~800 t/s) but *subsequent* prefill/decode is slower than before.

So I layered a cold-page gate on top of your design: a per-reader seen-page bitmap — the first time a chunk needs pages this process has never pulled in, pread them concurrently through your pool, then gather through the original mmap fancy-index. Warm calls pay only a bitmap probe. (Tried `mincore()` for precise eviction tracking: ~4.5 ms per shard-span call, so "evicted → single minor fault" seemed the better trade.) Also loops `os.pread` to handle the short-read nit @scubamount raised.

Opened #3287-adjacent stacked PR against your branch here: https://github.com/scott-kim-7/omlx/pull/1 — happy to have it folded in however you prefer, co-authored or as a review suggestion. The root cause analysis and the pool design are yours; this just makes the warm path stop paying for the fix.

scott-kim-7:
@williamxie1989 You're right, and thank you — this is careful work. I reproduced the regression before replying, at both levels.

### The microbenchmark

M4 Max 128 GB, 16 KB pages, 332 MB file, 166-byte rows, 4096-row gather, all pages resident:

| gather path | µs/row | vs `main` |
|---|---:|---:|
| `main` (mmap fancy-index) | 0.018 | 1.0x |
| #3287 (per-row pread pool) | 7.99 | **447x** |
| your cold-page gate (warm hit) | 0.085 | 4.8x |

Your 550x is real; I get 447x. Decomposing the pread path on warm pages: bare serial `os.pread` is 0.45 µs/row and `np.stack` 0.30, but **thread-pool dispatch alone is 4.00 µs/row**. The pool machinery, not the syscall, is most of the cost — the pool bills ~8 µs/row whether or not there is anything to overlap, so it wins 12x against a 100 µs fault and loses 450x against a 0.018 µs memcpy. Using `pread` as a cache warmer and leaving mmap as the data path is the right split.

### End to end

Four arms on the native oMLX path (`OMLX_USE_BUNDLE=0`), oMLX 0.6.4 / mlx 0.32.0 with custom kernels built, `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp` — the same model you and @latent-variable used. 12.2k-token `code_python` prompts, 200 generated, prefix-cache miss on every request (window shifted 128 tokens so the PLE rows repeat but the prefix does not). Decode timing is read from the server's own `timing[backbone=…]` MTP log rather than by subtraction, because prefill variance (22.7–39.4 s in early runs) swamps a differenced estimate. PLE table ~20–40 % resident.

| arm | n | prefill s | prefill tok/s | vs `main` | decode tok/s |
|---|---:|---:|---:|---:|---:|
| `main` | 3 | 6.95 | 1761 | — | 40.6 |
| #3287 alone | 3 | 8.42 | 1455 | **+21.2 %** | 40.5 |
| #3287 + your gate | 2 | 6.84 | 1790 | **−1.6 %** | 41.7 |
| #3372 | 3 | 7.62 | 1606 | **+9.7 %** | 42.4 |

Raw prefill seconds, no overlap between groups: `main` [6.86, 6.88, 7.11] · #3287 [7.97, 8.77, 8.52] · +gate [6.82, 6.86] · #3372 [7.45, 7.60, 7.82].

So **#3287 as it stands costs ~21 % of warm prefill end to end, and your gate removes all of it.** I'll merge scott-kim-7/omlx#1 and push the result here — two conflict hunks in `language.py` (the pool comment block and the `rows()` body), both resolving to your version, and it brings this branch up to current `main` as a side effect. I'll merge rather than squash so your commit keeps its authorship, and I'll say in the PR description which half is yours.

### The `> 8` threshold

I reproduced the n-gram hashing and the shard bucketing to test this, and your threshold is fine: decode never crosses it, and nothing is lost by that — decode is flat across all four arms above

### The one thing I'd still change in the gate

The seen-page bitmap is write-once, so an evicted page is never prefetched again for the life of the process. Some data on how much that matters: `mincore()` over the 8 PLE shard files of a server that had been up 2.5 hours showed **24.3 % of the 37.5 GB PLE table resident** (and a backbone shard at 0.0 % — the OS had reclaimed it entirely). Four benchmark requests took PLE from 9.1 GB to 14.1 GB resident, and it fell to 18.7 % after restarts. Eviction here is the steady state, not an edge case. Agreed that `mincore()` at ~4.5 ms/span is not payable on the hot path — but it is perfectly affordable as instrumentation, and a cheap re-arm (a generation counter, or clearing the bitmap when a gather's wall time crosses a threshold) would keep the gate honest without paying for it. Any objection to adding one?

### #3372

@evgri243 — I found #3372 while reviewing this patch and it deserves flagging here, since it does the same warm-via-threaded-pread-then-mmap-gather split, buckets across all shards in one prefetch, and removes the mid-forward `.tolist()` sync. I expected it to be a superset of this PR plus the gate. On the warm measurement above it is **+9.7 % slower than `main`** at 12.2k, so at least in this regime it carries a warm cost of its own that the seen-page gate does not. That is one prompt length on one machine and your +18 % was measured at 128k, so I'd treat it as a question rather than a result — but it seems worth knowing before either of these lands. Happy to run your arm under the same harness at 128k if that would help.

williamxie1989:
Thanks for folding the gate in, and for the re-arm suggestion — implemented it in scott-kim-7/omlx#2 (stacked on your branch).

Short version of what we landed: instead of `mincore()` (measured ~4.5 ms per shard-span call on an M5 Max / 128 GB box — a full pmap walk, too hot to price into every gather), every *fully-seen* gather self-instruments. A warm gather is a page-cache memcpy at ~0.1 µs/row, while an evicted-but-marked page turns it into serial ~100 µs minor faults, so wall time over a memcpy budget (0.5 ms floor + 2 µs/row) is the eviction signature. On breach we clear that reader's bitmap — one line, no generation counter needed — and the next gather re-warms through your concurrent pread path at ~9 µs/row. Rate-limited to once per reader per 60 s, with a single INFO log line (elapsed vs budget, per-shard ``(#N)`` counter) that doubles as the field instrumentation you mentioned.

Numbers from this machine: worst of 50 fully-warm gathers 0.84 ms vs an 8.69 ms budget (no false re-arm, 10× jitter margin); `sudo purge` fires exactly one re-arm per shard and prefill recovers within ~2 chunks; serving benchmarks on the warm table hold 1258/1207/1201 t/s at 16k/32k/65k — matching pre-change mmap warm gather, so the re-arm costs the hot path nothing. Your eviction data point was the convincing one: the table here now idles at ~40 % resident with one shard down to 2 %, so re-arm is the common case rather than an edge case.

Happy to argue about budget constants / interval — kept them as named module constants so tuning is a one-liner.

scott-kim-7:
Cold numbers are in, and they correct my last comment — everything I had posted until now was measured on a warm PLE table, the one regime where this patch cannot win.

**Setup.** M4 Max 128 GB, oMLX 0.6.4 / mlx 0.32.0 with custom kernels built, `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp` on the native path (`OMLX_USE_BUNDLE=0`), 16k `code_python` prompts, 200 generated, n=2 per arm, arm order counterbalanced. Two controls mattered: the cold state is produced by stopping the server and reading an unrelated 106 GB checkpoint, then **verified with `mincore()` at 0.0 % resident** before each run rather than assumed; and every request carries a run-unique nonce, without which oMLX's paged SSD prefix cache — which survives restarts and `purge` — returns a stored KV state and turns a 16k prefill into 0.52 s.

### Cold, PLE 0.0 % resident

| arm | prefill tok/s | raw | vs `main` |
|---|---:|---:|---:|
| `main` | 252.3 | 250 · 255 | — |
| #3287 alone | 382.7 | 372 · 393 | +51.7 % |
| **this PR** | 417.4 | 425 · 410 | **+65.4 %** |

**@williamxie1989's +71 % reproduces** (+65.4 %; +70.3 % in cycle 1), which also resolves his 736 t/s against the 1761 tok/s I posted earlier: his is a cold table, mine was warmed by repeated access to one region — two ends of one axis. And the gate improves the cold path as well as the warm one (+51.7 % → +65.4 %), because it deduplicates at page granularity instead of one `pread` per row.

### Across regimes

| regime | #3287 alone | this PR |
|---|---:|---:|
| warm, repeated access to one region | −21 % | ~0 % |
| fresh documents, 12–32k (table 35–50 % resident) | −10 % | −4 % |
| **cold table** | **+52 %** | **+65 %** |

`main` wins in exactly one regime. `mincore()` over the 37.5 GB table gave **24.3 % resident on a server up 2.5 hours** and 18–50 % across the session, with a backbone shard fully reclaimed — cold is not an edge case. Decode is flat across all arms in every regime (33–40 tok/s, tracking MTP accept rate), consistent with decode gathers arriving at 1–2 rows per shard.

### Status

Both of @williamxie1989's follow-ups are folded in with authorship preserved — the cold-page gate (scott-kim-7/omlx#1) and the eviction re-arm (scott-kim-7/omlx#2), the latter verified with zero false firings across ~70 measured requests. The cold table above was measured before #2 landed; the re-arm never fired during any of it, so the numbers stand.

The branch is rebased on current `main`, output is byte-identical to the original reader, the warm path is back at `main`, and cold is +65 %. I think this is ready to merge. @evgri243's #3372 optimises a different point on the same axis (faster cold, ~10 % warm cost, much lower variance) and is worth landing on its own merits — I'd rather this not block on that decision, and I'm happy to run #3372 through the same cold harness if that would help it along.

jundot:
Thanks for the cold-page gate and eviction re-arm updates. I checked concurrent requests up to a 32-row decode batch and a long prefill joining active requests; warm outputs matched the original reader, with similar end-to-end timings. LGTM, merging. I'll follow up in one commit to tighten the comments and make the prefetch and re-arm tests more precise.

Issue #2384, opened 2026-07-27
Prefix cache appears to store a full snapshot per request instead of reusing/extending the shared prefix


raucodes:
## Summary
 
In a single-user chat workload where consecutive requests share a ~12k-token
prefix and differ only in a short tail, the cache appears to grow by roughly
one full KV snapshot per turn.
 
After clearing both caches and running a single conversation, the hot cache
reports **3.8 GB / 10.0 GB across 20 entries** — about 190 MB per entry, with
all 20 entries coming from the same conversation and sharing the same ~12k-token
prefix. Before clearing, the SSD cache had reached 125 GB / 150 GB.
 
If the shared prefix were reused, I would expect the incremental cost per turn
to be on the order of the *new* tokens (a few hundred), not another full
snapshot.
 
I am not sure whether this is expected behaviour, a configuration issue on my
side, or a bug — hence the question rather than a bug report.
 
## Setup
 
| | |
|---|---|
| Hardware | Mac Studio M4 Max, 64 GB |
| Model | Qwen3.5-122B-A10B, 5-bit MLX |
| API | Anthropic Messages (`/v1/messages`), streaming |
| Workload | Single user, one conversation at a time |
| Prompt shape | ~22.3k chars system (constant), ~11.5k input tokens total |
 
The system string is byte-identical across turns — verified by logging its
length on the client side; it stays at 22310 characters over every request in
the samples below. Only the trailing user message changes.
 
### One entry per turn, all sharing the same prefix
 
```
Hot cache: 3.8 GB / 10.0 GB · 20 entries
```
 
20 entries after 20 turns of one conversation. Every one of those requests
shared the same ~12k-token prefix; only the trailing user message differed.
 
### Snapshot size does not vary with the changing tail
 
```
Using boundary cache snapshot for b01ea0e7…: storing 12288/12502 tokens (skipping trailing partial block, 0 intermediate snapshots)
Using boundary cache snapshot for c12b8474…: storing 12288/12296 tokens (skipping trailing partial block, 0 intermediate snapshots)
Using boundary cache snapshot for e57bee1b…: storing 12288/12780 tokens (skipping trailing partial block, 5 intermediate snapshots)
```
 
The same 12288 tokens are stored each time. The requests share that prefix —
only the last few hundred tokens differ.
 
### Cache phase timings grow monotonically
 
```
boundary_snapshot_save=463.1ms/28
boundary_snapshot_save=495.6ms/29
boundary_snapshot_save=528.5ms/31
[...]

PR #3288, opened 2026-08-29
scheduler: recover boundary-aligned live cache store for hybrid GDN (skew-guard fallback)

apcooley:
## Summary

Fixes the growing-context prefix-cache gap for hybrid GDN (non-sliceable) models: when the speculative-decode skew guard skips every decode-time boundary capture, the store path raised `_BoundaryStoreUnavailable` and **dropped** the recurrent state, so an agent context that grows turn-over-turn never persisted a reusable prefix and re-prefilled the full context every turn (`served cached_tokens=0`).

This adds a **store-time recovery**: if the request happened to finish exactly on a block boundary, verify the live cache's leaf offset sits on that boundary and store it instead of giving up.

Closes the "forced boundary-aligned snapshot" ask raised in #2384.

## Root cause

On hybrid models, cache state is non-sliceable. The store path only persists a prefix from a boundary-aligned snapshot. When the MTP skew guard skips all decode-time captures, no such snapshot exists, so the code bailed out:

```python
# scheduler.py, _cleanup_finished
raise _BoundaryStoreUnavailable()
```

...and the recurrent state was discarded. But if the request *finished* on a block boundary, the live extracted cache is already exactly a boundary-aligned snapshot — it is safe to store and recovers prefix reuse.

## Change

- New `_verify_live_boundary_aligned(request_id, uid, boundary_tokens)` that extracts the live cache and returns it **only** when the leaf cache offset equals the boundary token count exactly — mirroring `_extract_boundary_snapshot`'s positional-consistency guard, so an off-boundary state is never mislabeled with a block-boundary count.
- In `_cleanup_finished`, before raising `_BoundaryStoreUnavailable`, attempt the recovery when `len(cacheable_sequence) % block_size == 0`. On success, store the verified live cache; on failure, fall through to the existing raise (no behavior change otherwise).
- Adds two tests: on-boundary (stores) and off-boundary (skips).

## Why the safety guard matters

The emitted length alone can be fooled by the MTP emit queue, so the recovery trusts the live state only when the actual leaf cache `offset` equals the boundary token count. If it does not, the code returns `None` and the existing drop behavior is preserved — no risk of corrupting later prefix hits.

## Verification

- `pytest tests/test_scheduler.py -k boundary` → 22 passed
- New tests: `test_cleanup_finished_recovery_stores_boundary_aligned_live_cache` and the off-boundary skip case → 2 passed
- Rebased cleanly on `main` (no conflicts; upstream has not yet addressed this path).

## Notes for reviewers

- This is the recovery half of the question posed in #2384 ("is there room to store a forced boundary-aligned snapshot so growing agent context gets real prefix reuse?").
- It is orthogonal to the DFlash sidecar path that `d0lphin` diagnosed in #2384 — that is a separate read-side gap; this targets the store-side drop in the scheduler for the paged/base cache path.

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -7201,6 +7201,61 @@ class Scheduler:
             )
             return None
 
+    def _verify_live_boundary_aligned(
+        self, request_id: str, uid: int, boundary_tokens: int
+    ) -> list[dict[str, Any]] | None:
+        """Return extracted cache state iff the live cache is exactly at a block
+        boundary, else None.
+
+        Recovery path for the store-time ``_BoundaryStoreUnavailable`` case: when
+        every decode-time boundary capture was skipped (MTP skew guard) but the
+        request finished on a block boundary, the live recurrent/rotating state
+        sits at that boundary and is safe to persist. The leaf cache ``offset``
+        must equal the boundary token count exactly — mirroring
+        ``_extract_boundary_snapshot``'s positional-consistency guard — so we
+        never label an off-boundary state with a block-boundary count.
+        """
+        if self.batch_generator is None or uid is None or uid < 0:
+            return None
+        try:
+            with self._phase_timer("store_cache_boundary_verify"):
+                _safe_sync_stream(self._stream)
+                with mx.stream(self._stream):
+                    result = self.batch_generator.extract_cache([uid])
+            if uid not in result:
+                logger.debug(
+                    "Cannot verify live cache for %s: uid %s not present",
+                    request_id,
+                    uid,
+                )
+                return None
+            cache_list, _tokens = result[uid]
+            for c in cache_list:
+                offset = _first_leaf_cache_offset(c)
+                if offset is None:
+                    continue
+                if offset != boundary_tokens:
+                    logger.debug(
+                        "Live cache for %s is off-boundary (offset %d != boundary %d); "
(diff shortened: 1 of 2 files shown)


scubamount:
## Verified

- **Root cause correct:** when the MTP skew guard skipped every decode-time boundary capture, store raised `_BoundaryStoreUnavailable` and dropped the recurrent state — hybrid GDN agent contexts re-prefilled full history every turn (`cached_tokens=0`).
- **Fix right:** store-time recovery — if the request finished exactly on a block boundary, `_verify_live_boundary_aligned` syncs the stream, extracts the live cache, and checks every leaf's `offset == boundary_tokens` before persisting; off-boundary or any exception returns `None` (old drop behavior). Mirrors `_extract_boundary_snapshot`'s positional-consistency guard, so a state can never be mislabeled with a boundary count.
- **Bounded:** recovery only fires on the exact-boundary finish case; failure path unchanged.

No issues. Recovers prefix reuse without weakening the alignment invariant.

jundot:
Thanks for identifying the boundary-aligned recovery opportunity. The idea is useful, but normal completion removes the UID before cleanup, so the new extraction cannot recover that cache. I'll merge this and follow up by validating the cache carried by the final response, with a test covering completion through prefix reuse.

PR #3290, opened 2026-08-29
fix(cache): blank sliceable CacheList members in boundary snapshots when pm-plan is None

williamxie1989:
## Summary

For hybrid `CacheList` layers whose non-sliceable member is `PoolingCache` (GLM-5.x `CacheList(KVCache, PoolingCache)`), boundary snapshots previously retained the sliceable `KVCache` member in FULL at every block boundary — quadratic in-memory retention across a request's context (the #2551 class of bug), and the subsequent legacy store path could not restore a complete CacheList.

This PR makes `PoolingCache` per-member-block **eligible** (alongside `ArraysCache`): the boundary snapshot blanks the sliceable KV member (linear, not quadratic, retention), the store persists a true per-block KV slice, and the restore concatenates the slices back into the full sequence — so every stored boundary keeps the KV member restorable without holding the cumulative snapshot.

## Reproduction

- **Device**: Apple M5 Max, 128 GB unified memory, 18 cores, macOS 26.5.1, omlx 0.6.3.
- **Model**: GLM-5.3-Flash oQ2e (`glm5_next`, ~99 GB weights), 34x `ArraysCache` + 11x `CacheList(KVCache, PoolingCache)`.
- **Config**: `hot_cache_only=true`, `hot_cache_max_size=40GB`, memory guard tier `aggressive`, prefill chunk 2048.
- **Symptom**: long-context prefill aborted at ~30K tokens — `process memory limit exceeded (usage ~113 GB, hard watermark ~112 GB)`. `phys_footprint` crept to ~113 GB because retained in-memory boundary snapshots accumulated the full KV prefix of the 11 sparse layers at every 2048-token boundary.
- **After this fix**: the same scheduler path reaches 64K+ tokens under the same guard (peak 107.8 GB, phys 110.0 GB), and a store → reconstruct → decode round-trip reproduces the original cache's logits bit-for-bit.

## Root cause

`cachelist_pm_member_plan` rejected any `CacheList` containing `PoolingCache` (`_PM_SAFE_NON_SLICEABLE_SUBS` was {ArraysCache, SizedArraysCache}), so GLM-5.x fell through to the legacy cumulative path: the sliceable KV member was retained in FULL at every boundary snapshot, and the legacy last-block-only restore could not rebuild the KV sequence.

## Change

1. `PoolingCache` joins `_PM_SAFE_NON_SLICEABLE_SUBS` (`omlx/cache/prefix_cache.py`) — `CacheList(KVCache, PoolingCache)` now gets the per-member plan `["slice", "boundary"]`.
2. `cachelist_pm_class_eligible` (`omlx/cache/paged_ssd_cache.py`) accepts PoolingCache as a boundary class via a new `_PM_BOUNDARY_SUB_CLASSES` (kept separate from `_ARRAYS_SUB_CLASSES` so signature descriptors still stamp `PoolingCache:N`, not a mis-typed ArraysCache slot count).
3. `Scheduler._extract_snapshot_cache_states` blanks the sliceable KV member in the boundary snapshot (linear in-memory retention); the store slices the live KV per block, and the restore concatenates the slices — full KV sequence with correct positions.

DeepSeek-V4 (`CacheList(RotatingKVCache, PoolingCache)`) is unaffected: the rotating member keeps that layer on the legacy cumulative path (no slice member → plan still `None`).

Regression tests: `test_glm_pooling_cachelist_blanked_kv_roundtrips` now asserts the FULL KV sequence is restored (concatenated per-block slices, position-verified), and `test_glm_pooling_cachelist_blanks_kv_member` documents the pm eligibility.

## Relation to #3226 / #3227

Targeted sub-fix of #3226 item 4 for pm-ineligible `CacheList(KVCache, PoolingCache)`. PR #3227 addresses decode-time Metal buffer-count exhaustion (orthogonal; included in the rebase base). The GLM prefill memory profile (separate PR #3291) covers guard-side accounting.

## Tests

- `test_glm_pooling_cachelist_blanks_kv_member`: sliceable KV member blanked, non-sliceable PoolingCache kept, per-member plan `["slice", "boundary"]`.
- `test_glm_pooling_cachelist_blanked_kv_roundtrips`: store → reconstruct round-trip restores the FULL KV sequence (concatenated slices, position-verified) plus the cumulative PoolingCache pooled rows.
- `tests/test_cache_ntuple_state.py`, `tests/test_cache_type_handlers.py`, scheduler + engine-pool suites: 623 passing, no lint regressions on the touched files.

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -385,13 +385,14 @@ _POOLING_SUB_CLASSES = frozenset({"PoolingCache", "BatchPoolingCache"})
 # addition to ArraysCache (see ``cachelist_pm_member_plan``). Kept separate
 # from ``_ARRAYS_SUB_CLASSES`` so signature descriptors
 # (``_block_cachelist_subtypes`` / ``cachelist_subtypes_from_cache_list``)
-# still route PoolingCache through the pooling branch (``PoolingCache:N``)
-# instead of mis-stamping it as an ArraysCache slot count. Must stay in
-# sync with ``prefix_cache._PM_SAFE_NON_SLICEABLE_SUBS``: BatchPoolingCache
-# is NOT boundary-eligible — compact_pooling_cache_snapshot only compacts
-# PoolingCache, so BatchPoolingCache boundary snapshots stay cumulative and
-# its layers keep the legacy path.
-_PM_BOUNDARY_SUB_CLASSES = _ARRAYS_SUB_CLASSES | frozenset({"PoolingCache"})
+# still route pooling members through the pooling branch
+# (``PoolingCache:N`` / ``BatchPoolingCache:N``) instead of mis-stamping
+# them as an ArraysCache slot count. Must stay in sync with
+# ``prefix_cache._PM_SAFE_NON_SLICEABLE_SUBS``: boundary members are
+# self-contained at their boundary and restore last-block-wins; pooling
+# members that compact_pooling_cache_snapshot does NOT cover
+# (BatchPoolingCache) simply persist their cumulative boundary state.
+_PM_BOUNDARY_SUB_CLASSES = _ARRAYS_SUB_CLASSES | _POOLING_SUB_CLASSES
 # Sliceable KV sub-cache classes inside a CacheList (4D sequence tensors).
 # Shared with prefix_cache.cachelist_pm_member_plan (single source so the
 # class-level expectation and the shape-level store plan cannot drift).
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -386,8 +386,12 @@ _POOLING_SUB_CLASSES = frozenset({"PoolingCache", "BatchPoolingCache"})
 # from ``_ARRAYS_SUB_CLASSES`` so signature descriptors
 # (``_block_cachelist_subtypes`` / ``cachelist_subtypes_from_cache_list``)
 # still route PoolingCache through the pooling branch (``PoolingCache:N``)
-# instead of mis-stamping it as an ArraysCache slot count.
-_PM_BOUNDARY_SUB_CLASSES = _ARRAYS_SUB_CLASSES | _POOLING_SUB_CLASSES
+# instead of mis-stamping it as an ArraysCache slot count. Must stay in
+# sync with ``prefix_cache._PM_SAFE_NON_SLICEABLE_SUBS``: BatchPoolingCache
+# is NOT boundary-eligible — compact_pooling_cache_snapshot only compacts
+# PoolingCache, so BatchPoolingCache boundary snapshots stay cumulative and
+# its layers keep the legacy path.
+_PM_BOUNDARY_SUB_CLASSES = _ARRAYS_SUB_CLASSES | frozenset({"PoolingCache"})
 # Sliceable KV sub-cache classes inside a CacheList (4D sequence tensors).
 # Shared with prefix_cache.cachelist_pm_member_plan (single source so the
 # class-level expectation and the shape-level store plan cannot drift).
@@ -419,8 +423,7 @@ def cachelist_pm_class_eligible(sub_class_names: list[str]) -> bool:
     has_slice = any(n in _PM_SLICEABLE_SUB_CLASSES for n in names)
     has_boundary = any(n in _PM_BOUNDARY_SUB_CLASSES for n in names)
     all_known = all(
-        n in _PM_SLICEABLE_SUB_CLASSES or n in _PM_BOUNDARY_SUB_CLASSES
-        for n in names
+        n in _PM_SLICEABLE_SUB_CLASSES or n in _PM_BOUNDARY_SUB_CLASSES for n in names
     )
     return has_slice and has_boundary and all_known
 
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -381,6 +381,13 @@ _CACHELIST_NON_SLICEABLE_SUB_CLASSES = frozenset(
 
 _ARRAYS_SUB_CLASSES = frozenset({"ArraysCache", "SizedArraysCache"})
 _POOLING_SUB_CLASSES = frozenset({"PoolingCache", "BatchPoolingCache"})
+# Non-sliceable CacheList members accepted as per-member "boundary" subs in
+# addition to ArraysCache (see ``cachelist_pm_member_plan``). Kept separate
+# from ``_ARRAYS_SUB_CLASSES`` so signature descriptors
+# (``_block_cachelist_subtypes`` / ``cachelist_subtypes_from_cache_list``)
+# still route PoolingCache through the pooling branch (``PoolingCache:N``)
+# instead of mis-stamping it as an ArraysCache slot count.
+_PM_BOUNDARY_SUB_CLASSES = _ARRAYS_SUB_CLASSES | _POOLING_SUB_CLASSES
 # Sliceable KV sub-cache classes inside a CacheList (4D sequence tensors).
 # Shared with prefix_cache.cachelist_pm_member_plan (single source so the
 # class-level expectation and the shape-level store plan cannot drift).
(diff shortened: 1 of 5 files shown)


jundot:
Thanks for working on the boundary snapshot memory growth.

The current change blanks the KV member from intermediate `CacheList(KVCache, PoolingCache)` snapshots, but that member is not restored before the legacy store path persists them. This leaves the stored CacheList structurally incomplete and makes full and partial prefix restoration fail.

As a result, GLM requests lose prefix-cache reuse and must prefill again, which can bring back the latency and memory pressure this change is intended to reduce.

Could you keep the KV member restorable at every stored boundary while avoiding retention of the full cumulative KV snapshot?

williamxie1989:
Thanks for the review — both points are addressed in the updated commit (rebased on current main).

**Restorability at stored boundaries**: the in-memory boundary snapshot still blanks the sliceable KV member (so retention is no longer quadratic), but the legacy store path now refills it: `_cachelist_snapshot_sub_tensors` accepts the live layer state and a block slice range, and materializes the blanked KV member from the live cache sliced to the block's token range before persisting. Every stored block therefore carries real, restorable KV.

Verified with a full round-trip regression test (`test_glm_pooling_cachelist_blanked_kv_roundtrips`): store_cache -> reconstruct_cache restores both the KV member (correct position-encoded content) and the PoolingCache pooled rows.

I also rebased onto the latest main (0.6.4), which includes #3227.

williamxie1989:
Thanks for the review — the updated commit (rebased on current main) takes the direct route to "restorable at every stored boundary without cumulative retention": make the layer per-member-block eligible instead of refilling through the legacy path.

**What changed**

- `PoolingCache` is now a per-member "boundary" class (`_PM_SAFE_NON_SLICEABLE_SUBS`), so `CacheList(KVCache, PoolingCache)` gets the plan `["slice", "boundary"]` like `CacheList(KVCache, ArraysCache)` already does.
- Boundary snapshots still blank the sliceable KV member (in-memory retention stays linear), but the store now persists a true per-block KV slice and the restore concatenates the slices back into the full sequence — the KV member is restorable at every stored boundary, and no stored block ever carries the cumulative KV prefix.
- Signature descriptors keep stamping `PoolingCache:N` (via a separate `_PM_BOUNDARY_SUB_CLASSES`), so persisted signatures don't change meaning.
- DeepSeek-V4 (`CacheList(RotatingKVCache, PoolingCache)`) is untouched: no slice member → plan stays `None` → legacy path unchanged.

**Verification**

Beyond the unit round-trip test (which now asserts the FULL KV sequence is restored, position-verified, not last-block-only), I ran the loaded GLM-5.3-Flash oQ2e model through store → reconstruct → decode: the reconstructed cache's KV, pooled rows, and 45-layer state are bit-identical to the live cache, and 3 decode steps produce identical logits (max |diff| = 0). The same scheduler path reaches 64K+ tokens under the aggressive guard (peak 107.8 GB vs. ~113 GB abort before).

pedroberaldo87:
Heads-up from applying this on a fork branch: it turns `PoolingCache` plan-eligible, which leaves an existing test on `main` red.

```
tests/test_prefix_cache_cachelist_per_member.py::test_plan_helper_classification

    "pooling member": (["KVCache", "PoolingCache"], False),
                                                    ^^^^^
E   AssertionError: pooling member
E   assert (['slice', 'boundary'] is not None) == False
```

That case pins the old classification, so the assertion inverts with your change. Flipping it to `True` is all it takes — the rest of the file passes untouched (12 tests green here). Your new `tests/test_prefix_cache_cachelist_mixed.py` doesn't cover this helper, so nothing catches it inside the PR's own additions.

Full suite on top of `v0.6.4` with this PR plus #3291 and #3288: 10670 passed, and the only two failures reproduce on pristine `e008a66b`.

jundot:
Thanks for continuing to refine this. I know the PR is still a draft, but the remaining behavior is clear enough that I wanted to share my current test results now.

The overall per-member approach looks promising, and the KV side round-trips correctly. However, scheduler-produced boundary snapshots compact `PoolingCache.pooled` into per-block deltas, while the new PM path currently loses that delta-chain metadata. In a three-block round-trip using `Scheduler._extract_snapshot_cache_states()` followed by `compact_pooling_cache_snapshot()`, KV restored all 12 tokens, but `pooled` restored only 1 row instead of the expected 3. The current main path restores all 3 rows in the same scenario.

The main area that may be worth another look is how `PoolingCacheDelta` entries are preserved and rebuilt in PM mode, along with regression coverage using compacted snapshots. While checking this path, I also noticed that the `BatchPoolingCache` class eligibility and member-plan classifications do not currently line up, and the existing classification test still reflects the previous `PoolingCache` behavior. Those may also be useful to cover in the same pass if they fit the intended scope of this PR.

pedroberaldo87:
Production confirmation for the 01/09 rebase, from the bug @jundot measured on 31/08.

I hit the same pooled-rows loss on GLM-5.3-Flash with the earlier revision of
this routing — once `CacheList(KVCache, PoolingCache)` goes per-member, the
compacted boundary snapshots reached a branch that ignored the delta range in
both directions, and the pool came back 128 rows per block short. Because the
top-k selector only engages above 2048 tokens (`bypass_short`), nothing broke
until a conversation crossed that; then turns 6-11 of a 20-turn deterministic
replay diverged from the no-cache reference.

Measured on the real server (block 512, MTP off, temperature 0):

```
earlier revision   prefix hit vs no-cache reference    5 of 12 identical
with the delta-range fix                              12 of 12 identical
```

`OMLX_DEBUG_BOUNDARY=1` showed the capture at `tc=1536` with `pool_len=256`
after resuming from a cached prefix at 1024 — it should have been 384.

I had written an independent regression test for it before seeing the rebase
(a three-block round-trip through compacted snapshots asserting the pooled
tensor comes back whole — it failed on the earlier revision with "came back
with 2 rows instead of 6"). Against `82e66bb2` it passes, alongside
`test_prefix_cache_cachelist_per_member.py` and `test_pooling_cache_delta.py`:
27 green. Happy to send it as a PR against the branch if extra coverage from a
different angle is welcome; otherwise nothing to do here — the fix works in
production.

luken:
I have a regression test patch + what i believe is a fix patch at https://github.com/williamxie1989/omlx/pull/1 

for my test case its a big difference: Sampled peak memory 190.63 GiB versus the previous stock run's 219.85 GiB

Means i can run a 96k + two smaller 48k jobs without blowing through the high memory pressure.

jundot:
@luken I reproduced the pooled-history loss in both completion and parser-stop paths and confirmed that your patch fixes it. Could you open a PR against jundot/omlx with the fix and regression tests, noting the dependency on #3290? I'd like to review and track it directly alongside this PR.

luken:
> @luken I reproduced the pooled-history loss in both completion and parser-stop paths and confirmed that your patch fixes it. Could you open a PR against jundot/omlx with the fix and regression tests, noting the dependency on #3290? I'd like to review and track it directly alongside this PR.

Added #3710

williamxie1989:
@pedroberaldo87 — yes, please send it. A three-block compacted-snapshot round-trip asserted from an independent angle is exactly the coverage this path was missing when @jundot caught the 1-of-3-row pooled loss, and your production replay (5/12 → 12/12 identical once the delta range is honored) shows it bites right where `bypass_short` disengages above 2048 tokens.

Happy to take it as a PR against `pr/glm5-snapshot-blank`. If it's easier to land it with #3710's test set instead, that works too — whichever you prefer.

williamxie1989:
Status sync before this goes stale — thanks @pedroberaldo87 and @luken for the production validation and the promoted-boundary fix.

Current state of the branch (`82e66bb2`):

- Both items from @jundot's 08/31 review are addressed: `PoolingCacheDelta` chains are preserved and rebuilt in PM mode (`e5e07ad9`), and `BatchPoolingCache` class eligibility now agrees with the plan helper — `test_plan_helper_classification` is updated for the new classification ("pooling member" / "batch pooling member" → `True`), which also covers @pedroberaldo87's earlier heads-up.
- The remaining known gap — pooled-history truncation when `Scheduler._get_boundary_store_override()` promotes the final compacted boundary — is @luken's fix, tracked upstream in #3710 per @jundot's request. Intended merge order: #3290 first, then #3710 rebased onto main.

Next on my side: rebase onto current main (head is ~2 weeks / 211 commits behind), rerun the cache + scheduler suites, then flip this to ready for review.

@jundot — anything else you'd want in scope before I mark it ready, or is the above what you need to review #3290 alongside #3710?

jundot:
@williamxie1989 I've reviewed this together with #3710, and the combined changes look good. I'll handle the remaining integration updates and merge this first, followed by #3710, for the next release. Thanks for the great work on this!

jundot (approved):
The per-member path preserves full KV and pooling history when paired with #3710. I verified full and partial restores through hot cache and reopened SSD storage, including cache conversion and batched decode with the current dependency pins. LGTM; I'll merge this first and #3710 immediately afterward.

PR #3291, opened 2026-08-29
fix(glm5): add exact prefill memory profile for hybrid GLM-5.x attention

williamxie1989:
## Summary

GLM-5.x (e.g. GLM-5.3-Flash) mixes `linear_attention` (GDN) and `deepseek_sparse_attention` layers. The uniform prefill estimator priced the sparse layers as dense full-context SDPA (`O(query x kv_len)`), over-counting the prefill transient ~15x — 64 heads x 2048-query x 40K-KV x fp16 ≈ 9.9 GB/chunk, versus the ~0.66 GB the actual bounded sparse/local workload needs. That over-estimate feeds the admission guard and prefill throttle, causing spurious rejections / premature down-throttling of long-context prefill on GLM.

This adds an exact prefill memory profile for GLM-5.x, mirroring the existing `_DeepSeekV4PrefillMemoryProfile` pattern.

## Scope note

This changes the **pricing** (admission / throttle estimate) only — it does not change the actual physical peak. It prevents the guard from false-rejecting / over-throttling long-context prefill on GLM. (A separate memory-footprint issue — in-memory boundary-snapshot quadratic retention — is addressed in PR #3290; the two are independent.)

## Change

- Add `_Glm5PrefillMemoryProfile` in `omlx/memory_monitor.py`, dispatched in `make_prefill_memory_profile` for `glm5*` model types.
- Prices:
  - 34 GDN layers: short-conv local attention + fixed fp32 recursive state (does not grow with context).
  - 11 sparse-MLA layers across three runtime tiers, priced separately:
    1. native sparse-MLA (`sparse_mla_attention`): bounded by `index_topk`, never by `kv_len`;
    2. native exact-block (`exact_block_token_attention`): expands latent KV to per-head K/V over the full context — O(kv_len) per layer;
    3. dense bool-mask fallback (both kernels unavailable): `query x kv_len` mask + unfused O(heads x query x kv_len) scores — the largest tier.
    Tier selection is gated on the custom kernels' runtime availability.
  - exact resident KV (latent + pooled indexer keys + fixed GDN state).
- Adds `estimate_resident_kv_bytes` to satisfy the `PrefillMemoryProfile` protocol.
- GLM configs expose `head_dim=0` in the generic field (real dims live in `qk_nope_head_dim` / `linear_head_dim`). The profile exposes `effective_head_dim`, and `Scheduler._set_model_info_for_monitor` uses it so the profile is actually installed instead of the whole model silently skipping admission pricing.

## Environment & validation

- **Device**: Apple M5 Max, 128 GB unified memory, 18 cores, macOS 26.5.1, omlx 0.6.3.
- **Model**: GLM-5.3-Flash oQ2e (`glm5_next`, ~99 GB weights).
- **Validation**: at 40K context / 2048-token chunk, the three tiers price at ~0.66 GB (native sparse-MLA), ~2.9 GB (exact-block), ~10.2 GB (dense fallback) — matching what each runtime path actually materializes.

## Testing

- `TestGlm5PrefillMemoryProfile` in `tests/test_memory_monitor.py`: profile built, non-GLM falls back, transient bounded vs dense, three-tier pricing ordering and O(kv) scaling, gentle scaling with context, dense fallback gating, `effective_head_dim`.
- `test_glm5_head_dim_zero_installs_profile` in `tests/test_scheduler_prefill_memory_guard.py`: scheduler installs the profile despite generic `head_dim=0`.

Note: the `per-token` value a scheduler reports in its throttle log is a measured `phys_footprint` delta (MLX buffer-pool high-water), not this estimator; this profile corrects the admission/throttle *pricing* side, not that measurement.

### DIFF
--- a/omlx/memory_monitor.py
+++ b/omlx/memory_monitor.py
@@ -146,6 +146,19 @@ class PrefillMemoryProfile(Protocol):
         self, query_tokens: int, kv_len: int
     ) -> int: ...
 
+    @property
+    def effective_head_dim(self) -> int | None:
+        """A positive head-dimension for schedulers that require one.
+
+        Some hybrid configs expose ``head_dim=0`` in the generic field (e.g.
+        GLM-5.x where the real attention dims live in ``qk_nope_head_dim`` /
+        ``linear_head_dim``). Schedulers gate ``set_model_info()`` on a
+        positive generic ``head_dim``; a profile that can supply a real one
+        lets those models install the profile instead of silently skipping
+        admission pricing.
+        """
+        ...
+
 
 class MemoryMonitor:
     """
@@ -946,6 +959,10 @@ class _DeepSeekV4PrefillMemoryProfile:
     dtype_size: float
     wsdpa_dtype_supported: bool = False
 
+    @property
+    def effective_head_dim(self) -> int | None:
+        return self.head_dim
+
     def _wsdpa_route_active(self, *, topk: bool = False) -> bool:
         """Match the live WSDPA route without a process-wide head-dim flag."""
         if (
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for working on the GLM prefill memory estimate.

The real GLM-5.3 `TextConfig` exposes the generic `head_dim` field as `0`, while its actual hybrid attention dimensions come from fields such as `linear_head_dim`, `qk_nope_head_dim`, and `v_head_dim`. Because the scheduler requires a positive generic `head_dim`, it skips `set_model_info()` and never installs the new profile, so this PR does not currently fix the admission and throttling behavior.

The no-kernel path also references an undefined `kv_len`, so it will fail once the profile is installed on a system without the custom kernels.

The sparse MLA, exact-block, and dense fallback paths need separate estimates. Exact-block still expands K/V, while the dense fallback materializes much larger attention data. Treating them as the same 0.66 GiB bounded sparse path can underestimate memory and allow a request that later hits the memory guard or OOMs.

Could you update the profile installation and price each runtime path according to what it actually materializes?

williamxie1989:
Thanks for the detailed review — all three points are fixed in the updated commit (rebased on current main).

1. **Profile never installed (head_dim=0)**: GLM configs expose `head_dim=0` generically. The profile now exposes `effective_head_dim` (`qk_nope_head_dim` for GLM; `head_dim` for the DS4 profile), and `Scheduler._set_model_info_for_monitor` falls back to it when the generic `head_dim` is not a positive int, so the profile is installed and admission/throttle pricing actually applies. Regression test: `test_glm5_head_dim_zero_installs_profile`.

2. **Undefined kv_len in the no-kernel path**: `_sparse_attention_bytes` now takes `kv_len` explicitly instead of referencing an out-of-scope variable (the no-kernel path previously raised NameError).

3. **Tiers priced separately**: the sparse attention now prices the three runtime tiers independently, gated on each custom kernel's availability:
   - native sparse-MLA: bounded by `index_topk` (~0.66 GB at 40K/2048-chunk);
   - native exact-block: expanded per-head K/V over the full context, O(kv_len) (~2.9 GB at 40K);
   - dense bool-mask fallback: mask + unfused O(heads x query x kv_len) scores (~10.2 GB at 40K).
   So the estimate can no longer under-price a request that lands on exact-block or the dense fallback.

Also rebased onto latest main (0.6.4).

PR #3298, opened 2026-08-29
feat(qwen4/ane): Experimental Qwen4 ANE offload

onthehub97:
## Summary

This PR ports the existing Qwen3.5/3.6/3.8 fixed-shape ANE/CPU/GPU prefill path to the experimental `qwen4_exp` architecture while preserving model profiles and tuner functionality.

Qwen4 offload is deliberately limited to projections with a stable dense layout: the fixed shared expert and Gated DeltaNet (GDN) inputs. Routed MoE experts remain on the GPU, and the embedded MTP head is excluded from ANE bank discovery and compilation.

## What changed

- Added centralized Qwen ANE architecture detection with `qwen4_exp` support.
- Added Qwen4-aware module discovery that walks serving decoder layers only:
  - `mlp.shared_expert` is eligible for MLP offload.
  - `linear_attn` is eligible for GDN input-projection offload.
  - Routed experts and MTP-only layers are excluded.
- Installed the ANE dispatch hook for the Qwen4 implementation, which reuses the compatible Qwen3.5 MLP and GDN building blocks.
- Updated procedure-bank preparation and fallback paths to use the architecture-aware candidate lists.
- Kept Qwen4 tuning on the mlx-vlm engine so PLE, QSA, hyper-connections, and the embedded MTP implementation remain intact instead of probing the unsupported mlx-lm path first.
- Updated representative-layer calibration to use the same Qwen4 candidate selection as production loading.
- Added an opt-in `OMLX_QWEN35_ANE_BANK_MAX_MEMORY_FRACTION` override for mmap/PLE-heavy tuning runs. The existing 70% default remains unchanged, and overrides are constrained to 70–90%.
- Exposed Qwen4 compatibility in the admin API, web dashboard, macOS settings screen, DTO documentation, and localized settings copy.
- Retained the existing `qwen35_ane_prefill_*` setting names so saved profiles and tuner recommendations remain backward compatible.
- Added `benchmarks/qwen_ane_offload_microbench.py`, which loads real checkpoint projections without materializing Qwen4's routed experts or PLE tensors and compares GPU execution with the hybrid ANE/GPU path.

## Microbenchmark results

Measured with `Qwen3.8-Flash-Next-oQ4e-mtp`, a 2,048-token input, dual ANE, and real checkpoint weights:

| Projection | GPU median | Hybrid median | Speedup | Minimum cosine |
| q4 GDN | 10.654 ms | 7.961 ms | 1.338x | 0.999807 |
| q5 GDN | 10.832 ms | 8.274 ms | 1.309x | 0.999920 |
| q6 GDN | 10.239 ms | 8.197 ms | 1.249x | 0.999902 |
| q8 shared expert | 1.695 ms | 2.109 ms | 0.804x | 0.999756 |

The results support GDN offload across the q4/q5/q6 projection variants in this checkpoint. The narrow q8 shared expert is slower when offloaded in isolation, so the full-model tuner remains responsible for accepting or rejecting the aggregate split.

CPU sharing remains implemented, but this checkpoint is not CPU-offload eligible because its quantization scales are BF16; the existing CPU path requires FP16 scales/checkpoint clones.

## Validation

- 259 targeted Python tests passed across ANE prefill, tuning, Qwen4 compatibility/loading/residency, MTP gating, native CPU sharing, and admin settings.
- Added coverage for Qwen4 architecture matching, shared-expert/GDN discovery, MTP exclusion, VLM-only tuner loading, admin settings acceptance, and the bounded PLE memory override.
- Ruff passed for the changed implementation, benchmark, and relevant tests.
- `git diff --check` passed.
- The macOS app and test targets compiled successfully. The GUI test runner was interrupted after stalling during launch, with no compilation failure reported.

## Notes

- The feature remains opt-in and depends on private Apple Neural Engine APIs.
- The projection microbenchmark is intentionally scoped to offload viability; it avoids loading the full routed-expert and PLE footprint.
- Existing Qwen3.5/3.6/3.8 behavior and profile keys are preserved.

### DIFF
--- a/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
+++ b/apps/omlx-mac/Sources/AppView/Screens/ModelSettingsScreen.swift
@@ -1160,7 +1160,7 @@ private struct ExperimentalSection: View {
                                   defaultValue: "Qwen ANE Prefill",
                                   comment: "Row label for private Qwen ANE/GPU prefill acceleration"),
                     sublabel: String(localized: "settings.experimental.qwen_ane.sub",
-                                     defaultValue: "Split fixed-shape Qwen 3.5/3.6/3.8 prompt processing across both ANEs and the GPU. Experimental private API; takes effect after the model reloads.",
+                                     defaultValue: "Split fixed-shape Qwen 3.5/3.6/3.8/Qwen4 prompt processing across ANE, CPU, and GPU. Qwen4 accelerates fixed shared-expert and GDN projections; routed experts remain on GPU. Experimental private API; takes effect after the model reloads.",
                                      comment: "Sublabel describing Qwen ANE/GPU prefill acceleration")) {
                     Toggle("", isOn: vm.bindProfile($vm.qwen35AnePrefillEnabled))
                         .labelsHidden().toggleStyle(.switch)
--- a/apps/omlx-mac/Sources/AppView/ViewModels/ModelSettingsScreenVM.swift
+++ b/apps/omlx-mac/Sources/AppView/ViewModels/ModelSettingsScreenVM.swift
@@ -267,7 +267,7 @@ final class ModelSettingsScreenVM {
     var turboquantKvEnabled: Bool = false
     var turboquantKvBits: String = "4"
 
-    // Experimental: private Qwen3.5/3.6/3.8 ANE/GPU fixed-shape prefill.
+    // Experimental: private Qwen3.5+/Qwen4 ANE/CPU/GPU fixed-shape prefill.
     // These defaults are the measured M3 Ultra optimum for the 2,048-token
     // benchmark path. The feature itself remains opt-in.
     var qwen35AnePrefillEnabled: Bool = false
@@ -1011,6 +1011,7 @@ final class ModelSettingsScreenVM {
         return type.hasPrefix("qwen3_5")
             || type.hasPrefix("qwen3_6")
             || type.hasPrefix("qwen3_8")
+            || type.hasPrefix("qwen4_exp")
     }
 
     /// MTP can't co-exist with DFlash or TurboQuant KV. The toggle uses
--- a/apps/omlx-mac/Sources/Net/DTO/ModelsDTO.swift
+++ b/apps/omlx-mac/Sources/Net/DTO/ModelsDTO.swift
@@ -108,7 +108,7 @@ struct ModelSettingsDTO: Codable, Equatable, Sendable {
     // Experimental: TurboQuant KV cache
     let turboquantKvEnabled: Bool?
     let turboquantKvBits: Double?
-    // Experimental: private Qwen3.5/3.6/3.8 ANE/GPU prefill
+    // Experimental: private Qwen3.5+/Qwen4 ANE/CPU/GPU prefill
     let qwen35AnePrefillEnabled: Bool?
     let qwen35AnePrefillSequenceLength: Int?
     let qwen35AnePrefillTailPaddingMinTokens: Int?
@@ -191,7 +191,7 @@ struct ModelSettingsPatch: Encodable, Equatable, Sendable {
     // Experimental: TurboQuant KV
     var turboquantKvEnabled: Bool? = nil
     var turboquantKvBits: Double? = nil
-    // Experimental: private Qwen3.5/3.6/3.8 ANE/GPU prefill
+    // Experimental: private Qwen3.5+/Qwen4 ANE/CPU/GPU prefill
     var qwen35AnePrefillEnabled: Bool? = nil
     var qwen35AnePrefillSequenceLength: Int? = nil
     var qwen35AnePrefillTailPaddingMinTokens: Int? = nil
--- a/apps/omlx-mac/Tests/oMLXTests/ModelSettingsScreenVMTests.swift
+++ b/apps/omlx-mac/Tests/oMLXTests/ModelSettingsScreenVMTests.swift
@@ -221,6 +221,9 @@ final class ModelSettingsScreenVMTests: XCTestCase {
         vm.model = makeModel(id: "qwen", configModelType: "qwen3_8")
         XCTAssertTrue(vm.isQwen35AnePrefillModel)
 
+        vm.model = makeModel(id: "qwen4", configModelType: "qwen4_exp")
+        XCTAssertTrue(vm.isQwen35AnePrefillModel)
+
         vm.model = makeModel(id: "other", configModelType: "gemma4")
         XCTAssertFalse(vm.isQwen35AnePrefillModel)
     }
--- a/benchmarks/qwen_ane_offload_microbench.py
+++ b/benchmarks/qwen_ane_offload_microbench.py
@@ -1,0 +1,292 @@
+#!/usr/bin/env python3
+"""Microbenchmark Qwen3.5+/Qwen4 ANE offload on real checkpoint projections.
+
+Only one shared-expert MLP and one GDN projection bundle are loaded. This
+keeps the benchmark focused on heterogeneous offload viability without loading
+Qwen4's large routed-expert and PLE tensors.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import statistics
+import time
+from collections.abc import Callable
+from pathlib import Path
+from types import SimpleNamespace
+from typing import Any
+
+import mlx.core as mx
+import mlx.nn as nn
+from mlx_lm.models.activations import swiglu
+
+from omlx.custom_kernels.qwen35_prefill import fast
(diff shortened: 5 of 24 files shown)


DiscoStew6082:
I tested exact head `3d378681` with an existing local oQ6 checkpoint. No model was downloaded.

The layer 1 projection benchmark showed:

- q6 GDN: the hybrid path was `1.254x` faster, with minimum cosine similarity `0.999789`
- q8 shared expert: the hybrid path ran at `0.773x` the GPU speed, with minimum cosine similarity `0.999760`

I also tested the full 48-layer model through the VLM engine with an 8,193-token prompt, two output tokens, one warmup per process, and six measured samples per configuration.

- All-layer ANE improved median throughput by `5.81%` over the GPU baseline.
- Peak memory increased by approximately `0.77 GB`.
- A GDN-only configuration improved throughput by `3.12%`, so enabling both the shared-expert and GDN paths performed better overall.

The trace showed all expected GDN work executing, `144/144` operations, with no GPU tail tokens. One discrepancy may be worth checking: the model reported 48 ANE MLP layers, but each measured trace recorded `188/192` expected MLP operations. That looks like one shared-expert layer missing the ANE path across all four 2,048-token shapes, although I cannot tell whether the bypass is intentional.

This does not appear to block the performance benefit, but could you check which layer accounts for the missing four MLP operations?

cropduster:
Thanks for this PR — I ran the new `qwen_ane_offload_microbench.py` on a Mac Studio M4 Max (128 GB unified memory) and wanted to share results plus one finding that looks hardware-relevant.

**Setup:** I checked out the PR branch (`3d37868`) and ran the microbenchmark against the native ANE extension shipped with my installed oMLX 0.6.3 app. Since the PR doesn't touch `csrc/`/`fast.py`, I reused the app's `_ext.so`/`.dylib`/`.metallib` instead of doing a Metal-toolchain rebuild.

**One thing worth flagging:** the checkpoint I used — `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp` — reports `model_type: qwen4_exp` in `config.json`, so this actually exercised the new Qwen4 path on a real `qwen4_exp` checkpoint (`hyper_connection_mixer`, `linear_attn`, `mlp.shared_expert`), not just the legacy `qwen3_8` path. (It's the same checkpoint you measured with, so that likely applies to your table too.)

**Results** (layer 0, 2048 tokens, 5 repeats, median):

| Projection | GPU median | Hybrid median | Speedup | Min cosine |
| q6 GDN | 13.72 ms | 9.32 ms | **1.471×** | 0.999902 |
| q8 shared expert | 1.97 ms | 1.87 ms | **1.054×** | 0.999636 |

GDN offload reproduces and slightly beats your q6 number (1.249× → 1.47×).

**The shared-expert result differs from yours, though:** where you measured 0.804× (slower when offloaded in isolation), on the M4 Max it's marginally *faster* at 1.054×. That looks hardware-dependent — which actually supports your decision to leave the aggregate split to the full-model tuner rather than hardcoding a per-projection choice. CPU-offload stayed ineligible as expected (BF16 scales, matching your note).

Happy to run a full-model A/B (ANE on/off) once this lands in a buildable state, or provide a clean log if that's useful. Thanks for the work.

joshuaswarren:
I tested commit `3d378681`.
I used the M1 Mac Studio.

Setup:
- macOS 26.6.2, arm64, 128 GiB
- MLX 0.32.0
- Native ANE runtime and procedure-bank compiler available
- Checkpoint: `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp`
- `config.json` reports `model_type: qwen4_exp`
- mlx-vlm pinned at commit 78b96eb (0.6.3)

The four PR-touched Python test files passed:

```text
219 passed in 13.71s
```

Projection benchmark results (2,048 tokens, 5 repeats):

| Projection | Layer | Bits | GPU median | Hybrid median | Speedup | Minimum cosine |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| GDN | 1 | 4 | 42.551 ms | 38.327 ms | 1.110x | 0.999806 |
| GDN | 29 | 5 | 25.937 ms | 25.873 ms | 1.002x | 0.999920 |
| GDN | 0 | 6 | 23.033 ms | 15.435 ms | 1.492x | 0.999902 |
| Shared expert | 0 | 8 | 4.589 ms | 5.799 ms | 0.791x | 0.999640 |
| Shared expert | 1 | 8 | 7.674 ms | 14.397 ms | 0.533x | 0.999757 |

The checkpoint uses BF16 scales.
CPU offload is ineligible.

Full-model serving test (omlx serve, `--memory-guard off`, vendored qwen4_exp compat patch):

The vendored compat patch applied at load. PLE mode resolved to `resident`.
Model load duration was 89.21 seconds (103.94 GB, 21 shards from external NVMe).
The successful load used `iogpu.wired_limit_mb=124518`; I restored its prior value (`0`) afterward.

Generation worked. The model answered a prompt correctly.

Full-model prefill benchmark (2,259 prompt tokens, warm, `max_tokens=1`):

| Config | Prefill time | Prefill tok/s |
| --- | ---: | ---: |
| GPU only | 1.35 s | 1,673 |
| ANE prefill on | 1.63 s | 1,386 |

ANE prefill was 0.83x GPU-only speed at the full-model level.
ANE bank compilation on the first run added 9.6 seconds of cold overhead.
The net slowdown is consistent with the shared-expert projection overhead from the microbenchmark.

ANE log lines confirmed bank discovery and eager compilation (GDN input projections and prefill banks).
The ANE program budget exhausted before GDN completed all layers.

Note on mlx-vlm compatibility:
Upstream mlx-vlm commit 6102cb4 (0.7.0rc0) supplies `mlx_vlm.models.qwen4_exp` natively.
That module expects FP8 PLE shards and fails on oQ4e checkpoints:
```text
ValueError: Missing FP8 PLE shards for scale '...ple_embedding.ngram_embedding.weight_scale'.
```
The PR's vendored compat patch handles oQ4e PLE correctly when the pinned mlx-vlm (78b96eb) is installed.

ddalcu:
image
M4Max 128

Esmeralda-11:
Tested on **M2 Ultra / 24-core CPU / 60-core GPU / 128GB unified memory**.

Environment:
- macOS 26.6.2
- Python 3.11.8
- MLX 0.32.0
- oMLX 0.6.3
- PR commit: `3d378681`
- `model_type=qwen4_exp`

### Results

All relevant tests passed:

**219 passed in 2.54s**

#### Projection microbenchmarks

| Test | GPU | GPU + ANE | Speedup |
|---|---:|---:|---:|
| q4 GDN | 16.654 ms | 11.302 ms | **1.4735x** |
| q5 GDN | 16.914 ms | 12.472 ms | **1.3562x** |
| q6 GDN | 17.077 ms | 11.645 ms | **1.4664x** |
| q8 MLP L0 | 3.291 ms | 3.000 ms | **1.0968x** |
| q8 MLP L1 | 3.270 ms | 2.614 ms | **1.2513x** |

Cosine similarity was `0.99965–0.99992`.

#### Full-model ANE tuner

Final 5-repeat run:

| | GPU only | ANE optimized |
|---|---:|---:|
| Median prefill | **457.4 tok/s** | **472.5 tok/s** |
| Speedup | — | **+3.30%** |

Recommended configuration:

- MLP ANE fraction: `0.50`
- GDN ANE fraction: `0.375`
- CPU offload: disabled
- sequence length: `2048`
- tail padding min tokens: `1983`

A separate tuner run gave **+3.37%**, so the full-model improvement looks reproducible at roughly **+3.3%**.

### Interesting observation

For an 8193-token prefill (4 × 2048 ANE tiles):

- GDN: **144 / 144** operations observed
- MLP: **188 / 192** operations observed

The MLP result was reproducible across multiple profiling runs. It looks like one MLP/shared-expert operation per tile is not taking the ANE path.

<details>
<summary>Additional full-model baseline details</summary>

GPU-only, `pp=8193`, `gen=2`, warmup=1, 6 measured runs:

- Prefill: `470, 469, 470, 469, 469, 468 tok/s`
- Median: **469 tok/s**
- TTFT median: **17461.5 ms**
- Decode median: **11.15 tok/s**
- Benchmark memory: **76.6 GB**

Model:
- Full size: **103.94 GB**
- Actual loaded: **98.12 GB**
- PLE mode: `resident`
- Cold server load: ~**56.2s**

</details>

<details>
<summary>ANE compile details</summary>

At load:

- Warmed `168` ANE procedures
- Compiled `48 MLP`
- Compiled `36 GDN`
- `2` instance-pinned ANE programs

</details>

Hope this M2 Ultra data is helpful for the PR. Thanks for the great work!

onthehub97:
Closing this PR, as I have made an ANE breakthrough that supercedes this.

PR #3299, opened 2026-08-29
fix(engine_pool): keep the Qwen4 PLE table resident across a model swap

pedroberaldo87:
Swapping models from the admin panel costs 2.5x decode throughput on a `qwen4_exp` model, until the server is restarted. The persisted settings say nothing is wrong — `qwen4_ple_ssd_offload` stays `false` throughout.

`_qwen4_ple_offload_status` asks the admission ceiling, which is `min(static, dynamic, metal_cap)`. The `dynamic` component reads `vm_stat` at that moment, and on a swap that reading lands after the outgoing engine released its arrays but before the OS returned the pages:

```
Unloaded antigo, freed=99.01GB, active_memory: 1.02KB (settled)
Admitting fp16: committed baseline 104.84GB fits ceiling 121.60GB
Qwen4-Exp PLE mode: mmap
Loaded fp16 (actual: 548.38MB, local estimate: 73.55GB, full model: 104.84GB)

clean load   ceiling 121.60GB -> resident
after swap   ceiling 100.78GB -> forced to SSD
```

`force_ssd_offload` fires (resident 115GB doesn't fit 100.78, mmap 73.55GB does), and `_effective_qwen4_model_settings` applies it to a copy, so nothing surfaces. The `73.55GB` in the load line is the estimate's `mmap_bytes`.

Admission needs memory free right now; residency does not. `get_residency_ceiling()` drops the `vm_stat` component, and the forced decision now logs a warning naming the ceiling.

Paired, novel prompts, `cached_tokens=0` on all six runs, 400 generated tokens, temp 0, M1 Ultra 128GB:

```
before   clean 35.3 tok/s (112GB, resident)  ·  after swap 14.1 tok/s (78GB, SSD)
after    clean 26.0/32.0/33.4 -> 30.5        ·  after swap 31.3/26.9/28.8 -> 29.0
```

The distributions now overlap (26.0-33.4 against 26.9-31.3). Absolutes are lower than the "before" row because those prompts are novel and the n-gram drafter has nothing to copy — only the within-round comparison holds.

Only the incoming model's type is checked, so this needs an incoming `qwen4_exp` whose resident estimate clears the stable ceiling but not the depressed one, plus anything large unloading right before. Two qwen4_exp models are not required.

305 tests pass across residency, engine pool and memory enforcer. The new test feeds a depressed instantaneous ceiling with a roomy stable one and asserts the table stays resident; its second half asserts the old behaviour returns when no callback is wired.

### DIFF
--- a/omlx/engine_pool.py
+++ b/omlx/engine_pool.py
@@ -285,6 +285,7 @@ class EnginePool:
         self._get_final_ceiling: object | None = None  # Set by server
         self._get_admission_ceiling: object | None = None  # Set by server
         self._get_admission_soft_target: object | None = None  # Set by server
+        self._get_residency_ceiling: object | None = None  # Set by server
         self._settings_manager: object | None = None  # Set by server
         self._cluster_registry: ClusterRegistry | None = None  # Set by server
         self._suppress_ttl: bool = False  # Suppress TTL during benchmarks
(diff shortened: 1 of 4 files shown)


scubamount:
## Verified

- **Root cause correct:** PLE residency decision asked the *admission* ceiling, whose `dynamic` component reads `vm_stat` — during a swap that lands after the old engine freed arrays but before the OS returned pages, so the ceiling dipped and a fitting table got forced to mmap/SSD for the engine's whole life (35.3 → 14.1 tok/s).
- **Fix right:** new `get_residency_ceiling` on the enforcer drops the vm_stat component (static + metal cap only) — admission still respects live free memory, residency uses the stable ceiling. Falls back to admission→current ceiling when unwired, preserving old behavior.
- **Observability:** explicit WARNING when a table is actually forced to SSD, naming sizes and the 2.5x cost — this was previously invisible.

No issues. Correct predicate split: admission question vs throughput question.

jundot:
Thanks for tracing the swap-time vm_stat dip and validating the throughput recovery. The stable residency ceiling fixes the swap regression, and the paired measurements make the impact clear.

I found one separate low-memory edge where a load that previously fit through mmap can now be rejected because the resident form is selected before admission. I can handle that in a small follow-up on main by carrying the pre-unload admission decision into the load when only mmap fits. LGTM for the primary fix.

PR #3301, opened 2026-08-29
fix(audio): unblock realtime STT for voxtral_realtime models

### LINKED ISSUE #3231: bug:Realtime STT WS rejects voxtral_realtime models on bundled server: mlx_audio Model lacks create_streaming_session while /v1/models/status reports realtime_stt: true
## Environment

- oMLX.app **0.6.0 (2303)** on macOS (Apple Silicon), using the bundled `omlx-server`
- Model downloaded locally: `Voxtral-Mini-4B-Realtime-2602-4bit` (loaded, `engine_type: audio_stt`, `config_model_type: voxtral_realtime`)
- The realtime WS endpoint from #2603 is present in the bundled server (`/v1/audio/transcriptions/realtime` exists and responds)

## What works

`GET /v1/models/status` correctly advertises the model as realtime-capable:

```json
{
  "id": "Voxtral-Mini-4B-Realtime-2602-4bit",
  "engine_type": "audio_stt",
  "model_type": "audio_stt",
  "config_model_type": "voxtral_realtime",
  "realtime_stt": true
}
```

Sending `{"type":"start","model":"Qwen3-ASR-1.7B-8bit",...}` to the WS endpoint is also correctly rejected with `"does not support realtime transcription"` — the protocol-level gating works.

## What fails

Starting a realtime session for the Voxtral realtime model is rejected at handshake:

```json
{"type": "error", "detail": "Model 'Voxtral-Mini-4B-Realtime-2602-4bit' does not support realtime transcription. Use POST /v1/audio/transcriptions with a complete audio file."}
```

(Same result with the `mlx-community/Voxtral-Mini-4B-Realtime-2602-4bit` repo-style ID.)

## Root cause analysis

The handshake path checks the **engine instance**:

```python
def supports_realtime_stt(self) -> bool:
    return hasattr(self._model, "create_streaming_session") or hasattr(self._model, "generate_streaming")
```

The `mlx_audio` bundled with the app defines `VoxtralStreamingSession` (and its docs show `model.create_streaming_session(...)`), but the voxtral `Model` class in that build does **not** define the `create_streaming_session` factory method — so `hasattr` returns `False` and the endpoint rejects a model whose config type is `voxtral_realtime`.

In other words, two capability checks disagree:

| Check | Layer | Result |
| --- | --- | --- |
| `/v1/models/status` → `realtime_stt` | classification (`config_model_type in REALTIME_STT_MODEL_TYPES`) | `true` |
| WS handshake → `supports_realtime_stt()` | engine (`hasattr` on loaded model) | `False` |

Anything that selects models by `realtime_stt` from `/v1/models/status` (gateways, client UIs) will pick this model and then get rejected at handshake.

## Notes

- The official hosted demo instance reports this model with `realtime_stt: true` and serves realtime sessions, so newer builds already have the working path — it looks like the app's bundled `mlx_audio` just needs to catch up.
- Side suggestion: consider making the two checks consistent (e.g. `realtime_stt` in `/v1/models/status` reflecting actual engine capability, or at least warning when they disagree), so downstream consumers can trust the flag.

## Ask

Please ship an app update bundling an `mlx_audio` that implements `create_streaming_session` on the voxtral `Model` class (whisper-family `generate_streaming` realtime also depends on the same engine-level check). Happy to re-test and report back once a build lands.

iazrael:
Fixes #3231

## Root cause

`/v1/models/status` classifies `config_model_type=voxtral_realtime` as realtime-capable statically (`REALTIME_STT_MODEL_TYPES`), but the realtime WS handshake probes the loaded engine via `hasattr(model, "create_streaming_session")`. The pinned mlx-audio commit (`5175326`, 2026-04-07) predates the streaming API (upstream `f679861`, 2026-04-19 — "feat(voxtral_realtime): live input streaming"), so the two checks disagree: status advertises `realtime_stt: true` while the handshake rejects every session.

Note the model-loading path itself is correct — the checkpoint's `config.json` declares `model_type: voxtral_realtime` and mlx-audio's remap loads the right class; that class just had no session API at the pinned commit.

## Changes

- **`pyproject.toml` / `packaging/build.py`** — bump the mlx-audio pin `5175326` → `40bc168` (2026-07-04). It has the streaming API and still declares `transformers>=5.5.0,<5.13.0` / `mlx-lm>=0.31.1`, compatible with our `transformers>=5.12.1,<5.13` pin and git-pinned mlx-lm 0.31.3. Later commits aren't resolvable yet: upstream raised the transformers floor to `>=5.14.0` in `3e9e8aa` (2026-07-29).
- **`omlx/engine_pool.py`** — new `_realtime_stt_capable()`: unloaded models keep the static classification; once an engine is loaded, the engine's honest `supports_realtime_stt()` probe wins, so the status flag can never promise a capability the handshake would reject (the side suggestion from the issue).
- **`Formula/omlx.rb`** — same pin bump for the brew path; drop the `mlx-lm==0.31.1` `inreplace` (now natively `>=0.31.1`, the inreplace would fail) and explicitly install `misaki`/`spacy`/`librosa`/`numba`/`pyloudnorm`/`num2words`/`phonemizer-fork`/`espeakng-loader`, which upstream moved out of its dependency set (formula previously got them transitively; Kokoro G2P and the spacy model check need them).
- **`tests/test_engine_pool.py`** — regression tests for the `realtime_stt` flag semantics (unloaded → static classification; loaded → engine probe wins; engine without the probe → `False`).

## Verification

- `uv sync --extra audio` resolves end-to-end with the new pin (the old one couldn't, due to the strict `mlx-lm==0.31.1` conflict), and the installed mlx-audio exposes `create_streaming_session` / `VoxtralStreamingSession` with signatures matching `_VoxtralRealtimeBackend` (`feed`/`step(max_decode_tokens=...)`/`close`/`done`); transformers stays at 5.12.1.
- All mlx-audio APIs omlx imports verified present at the pinned commit (stt/tts utils, whisper `StreamingConfig`/`StreamingDecoder`/`log_mel_spectrogram`, qwen3_asr `split_audio_into_chunks`, the four STS model classes, voxtral batch `Model`).
- 436 tests pass: `test_audio_realtime.py`, `test_engine_pool.py`, `test_audio_discovery.py`, `test_model_discovery.py`, `test_status_endpoint.py` (base includes current upstream/main).

Not covered here (needs hardware/model download): an end-to-end realtime WS round-trip against `Voxtral-Mini-4B-Realtime-2602-4bit` and the macOS app build via `packaging/build.py`.

### DIFF
--- a/Formula/omlx.rb
+++ b/Formula/omlx.rb
@@ -25,11 +25,15 @@ class Omlx < Formula
     skip_clean "libexec" if MacOS.version >= "27"
   end
 
-  # mlx-audio pins mlx-lm==0.31.1 which conflicts with omlx's git-pinned
-  # mlx-lm. Fetch source separately so we can patch the pin before install.
+  # mlx-audio is installed from a pinned commit instead of omlx's [audio]
+  # extra (brew installs omlx without extras): the pin must include
+  # voxtral_realtime create_streaming_session (upstream f679861) or the
+  # realtime STT WebSocket rejects every session while /v1/models/status
+  # advertises it (#3231). 40bc168 is the newest commit whose declared deps
+  # stay compatible with omlx's transformers<5.13 pin.
   resource "mlx-audio" do
     url "https://github.com/Blaizzy/mlx-audio.git",
-      revision: "51753266e0a4f766fd5e6fbc46652224efc23981"
+      revision: "40bc1680eaa47b8f3d5992bbb2913f08c3d1a123"
   end
 
   # Kokoro's English G2P path uses misaki + spaCy. Bundle the spaCy
@@ -110,11 +114,20 @@ class Omlx < Formula
       end
     end
 
-    # Install mlx-audio with patched mlx-lm pin to avoid version conflict
+    # Install mlx-audio from the pinned commit. Its declared mlx-lm>=0.31.1
+    # is satisfied by omlx's git-pinned 0.31.3, so no pin patching needed.
     resource("mlx-audio").stage do
-      inreplace "pyproject.toml", '"mlx-lm==0.31.1"', '"mlx-lm>=0.31.1"'
       system(*pip_install, ".[all]")
     end
+
+    # Since upstream 40bc168, misaki/spacy & co are optional installs that
+    # mlx-audio no longer pulls transitively. Install the G2P and
+    # audio-analysis deps omlx needs at runtime explicitly (mirrors omlx's
+    # [bundle] extra; the spacy model wheel below completes the Kokoro path).
+    system(*pip_install,
+           "misaki>=0.9.4", "num2words>=0.5.14", "spacy>=3.8.4",
+           "phonemizer-fork>=3.3.2", "espeakng-loader>=0.2.4",
+           "librosa>=0.10.0", "numba>=0.59.0", "pyloudnorm>=0.1.0")
 
     # Install the spaCy English model required by misaki for Kokoro TTS.
     # Homebrew's cached resource path is hash-prefixed, which pip rejects
--- a/omlx/engine_pool.py
+++ b/omlx/engine_pool.py
@@ -243,6 +243,23 @@ class EngineEntry:
     load_failure_at: float | None = None
 
 
+def _realtime_stt_capable(entry: EngineEntry) -> bool:
+    """Whether an entry advertises realtime STT to status consumers (#3231).
+
+    Unloaded models get the static config classification. Once the engine is
+    loaded, the engine-level probe wins, so the flag never promises a
+    capability the realtime WS handshake would reject (e.g. a bundled
+    mlx-audio whose voxtral Model predates ``create_streaming_session``).
+    """
+    if not is_realtime_stt_model(entry.model_type, entry.config_model_type):
+        return False
+    engine = entry.engine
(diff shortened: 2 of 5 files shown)


scubamount:
## Verified

- **Root cause correct:** `/v1/models/status` advertised `realtime_stt: true` statically from `REALTIME_STT_MODEL_TYPES`, but the WS handshake probes `hasattr(model, "create_streaming_session")` — pinned mlx-audio `5175326` (2026-04-07) predates the streaming API (`f679861`, 2026-04-19), so the two checks disagreed and every session was rejected.
- **Fix right (two arms):** bump mlx-audio pin to `40bc168` (newest commit compatible with the transformers<5.13 pin) so the capability actually exists; AND make the status flag honest via `_realtime_stt_capable` — unloaded gets static classification, loaded defers to the engine probe so the flag never promises what the handshake would reject.
- **Packaging:** mlx-audio's now-unsatisfied G2P deps (misaki, spacy, phonemizer, librosa, …) installed explicitly since upstream made them optional; mlx-lm pin patch no longer needed (0.31.3 satisfies >=0.31.1).

No issues. Correct dual fix — capability + truthful advertising.

iazrael:
Thanks for the verification and review!

PR #3302, opened 2026-08-29
fix(dflash): gate on resolved model types instead of a 'qwen' substring

pedroberaldo87:
`is_dflash_compatible` decides with a substring:

```python
model_type = str(cfg.get("model_type") or "").lower()
is_qwen = "qwen" in model_type
```

Every `qwen*` passes, `qwen4_exp` included. mlx_lm resolves no DFlash path for it, so the loader raises `Model type qwen4_exp not supported`, `engine_pool` catches it and falls back to the standard engine. The gate returns compatible, the settings route accepts the write, and the operator sees the toggle stay on while nothing behind it changed.

In #2175 you wrote that the gate is not display-only — the settings route 400s on an incompatible model — and that merging a gate change alone "would make the dashboard advertise compatibility the runtime cannot deliver yet". That is the state `qwen4_exp` is in today, from the other direction: the gate lets it through and the runtime cannot deliver.

This replaces the substring with the list mlx_lm actually resolves, so an unsupported type is refused at the gate with the usual 400 instead of failing silently after the load. The shape matches the neighbouring branches, which already enumerate (`is_gemma4 = model_type in (...)`).

Checked against the current pin: `qwen3_5` and `qwen3_next` keep passing, `qwen4_exp` now reports incompatible. `tests/test_dflash_engine.py` gains a case pinning both directions — 116 tests pass in that file.

### DIFF
--- a/omlx/engine/dflash.py
+++ b/omlx/engine/dflash.py
@@ -88,7 +88,17 @@ def is_dflash_compatible(model_path: str | Path) -> tuple[bool, str]:
 
     model_type = str(cfg.get("model_type") or "").lower()
 
-    is_qwen = "qwen" in model_type
+    # Substring match let every "qwen*" through, including qwen4_exp — the
+    # dashboard offered the toggle, the loader then raised
+    # "Model type qwen4_exp not supported" and engine_pool swallowed it, so
+    # the operator saw the switch turn on and silently fall back (v8 F2.7).
+    # The list is what mlx_lm actually resolves for dflash.
+    is_qwen = model_type in (
+        "qwen2", "qwen2_moe",
+        "qwen3", "qwen3_moe",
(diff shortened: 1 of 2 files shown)


scubamount:
## Verified

- **Root cause correct:** `"qwen" in model_type` substring let `qwen4_exp` through the dashboard gate; loader then raised "Model type qwen4_exp not supported", engine_pool swallowed it, operator saw the toggle silently fall back.
- **Fix right:** exact-match tuple of the seven model types mlx_lm actually resolves for dflash — matches the adjacent `gemma4`/`laguna`/`muse` exact-match style.
- **Test:** `test_qwen4_exp_is_incompatible` asserts `compatible is False` with a non-empty reason.

No issues. Correct set-based predicate replacing the leaky substring.

PR #3304, opened 2026-08-29
perf(moe): extend affine block kernels to oQ Q4/Q6/Q8

DiscoStew6082:
## Summary

- make group-64 affine Q4/Q6/Q8 `QuantizedSwitchLinear` projections eligible for the native block-list MoE path already present on `main`;
- let compatible `SwitchGLU` layers use the paired gate/up kernel and reuse the block plan for the down projection at the existing affine threshold;
- fail closed for Q4/Q6/Q8 when current Python source is paired with an older native extension;
- add non-skipping macOS CI that builds the extension and exercises the native symbols, routing, threshold, dtype, and stale-extension behavior.

## What changed after the rebase

The original PR also carried the low-level Q4/Q6/Q8 unpacking, packing geometry, and kernel instantiations. While this PR was open, #3574 added that native format support to `main`. The September 14 rebase removed the duplicated kernel implementation.

The remaining integration gap is in `QuantizedSwitchLinear._can_use_affine_blocks`: current `main` has the Q4/Q6/Q8 kernels, but route eligibility is still limited to Q2/Q3. This PR extends that route to Q4/Q6/Q8 at the existing 1,024-route affine threshold, adds an explicit extension capability marker so stale binaries cannot receive unsupported formats, and keeps small windows on the stock path.

The current diff is limited to route eligibility and capability detection, native CI, and the associated regression coverage. It does not change checkpoint formats, conversion behavior, MTP, cache behavior, route thresholds, or the low-level kernel implementation now on `main`.

## Validation

- Hosted CI passed on Python 3.11, 3.12, and 3.13; each job reported 13,050 passed, 387 skipped, and 82 deselected.
- The dedicated `native-affine-moe` macOS job built the extension and passed 30 native tests.
- Native Q2/Q3/Q4/Q6/Q8 single and paired outputs match `mx.gather_qmm` for both FP16 and BF16 metadata, including oQ/oQe weighted-affine Q4/Q6/Q8 layouts.
- Route tests cover the current 1,024-route threshold, small-window fallback, FP16 affine metadata with BF16 inputs, and fail-closed behavior for Q4/Q6/Q8 with a stale extension.
- A retained group-64 Q6 checkpoint smoke test completed 16 greedy tokens with 42 paired and 42 single native affine MoE calls.

## Performance context

On the pre-rebase branch, a single-expert layer benchmark on a 512 GB M3 Ultra with MLX 0.32.0 measured the paired kernels at 1.11x-1.17x and the single kernels at 1.09x-1.12x versus stock at 8,192 routes. Results at the 64-route boundary were mixed.

Those measurements support enabling the existing kernels at prefill-sized route counts, but they were not rerun after the kernel implementation moved to `main`. They are layer-level results, not an end-to-end or blanket throughput claim.


DiscoStew6082:
I benchmarked this PR against its exact parent on a 512 GB M3 Ultra using PipeNetwork’s `GLM-5.3-Flash-MLX-6bit` checkpoint.

- Parent: `1d7826185c5b5b69b38b27cbe57d7597b7551fd7`
- PR head: `b095c75b4e38ca1deabcd649711277d168f32be1`
- Quantization: ordinary MLX affine Q6/group-64, with the checkpoint’s Q8 indexer overrides
- Method: one warmup followed by three measured uncached prefills per build and context; table values are medians
- No MTP, prompt/APC/prefix cache, chunked prefill, server, or network path

| Context | Parent prompt tok/s | PR prompt tok/s | Improvement | Parent TTFT | PR TTFT | TTFT reduction |
|---:|---:|---:|---:|---:|---:|---:|
| 2K | 449.24 | 478.14 | **+6.43%** | 4.561s | 4.285s | **6.04%** |
| 4K | 427.39 | 447.52 | **+4.71%** | 9.586s | 9.155s | **4.50%** |
| 8K | 434.62 | 453.18 | **+4.27%** | 18.851s | 18.079s | **4.10%** |
| 16K | 434.95 | 451.46 | **+3.80%** | 37.672s | 36.294s | **3.66%** |

The PR improved prefill throughput and TTFT at every tested context. The gain gradually tapered from 6.4% at 2K to 3.8% at 16K.

Validation:

- 24/24 recorded prefills passed their workload and dispatch contracts
- 8/8 fresh model processes completed successfully without retries
- Parent used the stock affine-Q6 route; the PR used the new native affine pair/down route across all 42 MoE layers
- Parent and PR output hashes matched at every context
- No thermal or performance warnings
- No material memory regression

These results measure uncached, single-prompt prefill and TTFT. They do not represent decode speed, MTP, continuous batching, or end-to-end server throughput.

scubamount:
## Verified

- **Correctness:** extends affine dequant to `bits=6` (4 values packed in 3 bytes) and completes missing `Q4`/`Q8` instantiations for both `float16`/`bfloat16` — prior `static_assert` now covers `{2,3,4,6,8}`.
- **Stale-binary guard:** new `AFFINE_MOE_HAS_Q468=true` attribute + `fast.py` capability check prevents routing `Q4/Q6/Q8` into old extensions that only accepted `Q2/Q3`.
- **CI added:** `native-affine-moe` job installs with `OMLX_WITH_CUSTOM_KERNEL=1` on macOS-15 and asserts symbol availability + `affine_moe_supports_bits({4,6,8})` before running targeted native tests.

No issues. Tight kernel + routing fix with native CI gate.

PR #3306, opened 2026-08-29
Pick eviction victims by size, not age alone

jungwankim-man:
## Problem

`_find_lru_victim` and `_find_lru_prefill_eviction_victim` sort candidates purely by `last_access`, so victim selection never looks at how much memory a model would actually free. The callers loop until enough headroom exists, so this converges — but on a host serving models that differ by an order of magnitude it disturbs far more models than necessary, and each one has to be reloaded later.

Concretely, with a 1.7GB TTS model, an 8GB LLM and a 17GB LLM all idle and 20GB needed, the loop evicts all three (26.7GB) when the 17GB model alone would have sufficed.

I hit this on a 128GB M-series box running a 68GB LLM alongside a 17GB fallback, a 7B VLM and a 1.7B TTS model.

## Change

Both selectors now build `(last_access, freeable_bytes, model_id)` and defer to a shared `_pick_eviction_victim`:

- prefer the **coldest model that on its own covers `bytes_needed`**, so a pressure event disturbs a single model;
- fall back to the **largest** when nothing is big enough, so the caller's loop converges in fewer rounds.

Freeable size reuses `entry.actual_size` where the loader has measured it and `_entry_resident_size` otherwise, so no new bookkeeping is introduced.

`bytes_needed` defaults to `0`, which reproduces the previous pure-LRU selection exactly. The two call sites pass the headroom they had already computed (`projected - evict_target`, `current + predicted - target`).

## Tests

Four cases added to `TestLRUEviction`:

- a single sufficient model is chosen over the coldest insufficient one;
- among several sufficient models, age still decides;
- falls back to the largest when none suffices;
- omitting `bytes_needed` keeps plain LRU.

All four pass. The pre-existing failures in `tests/test_engine_pool.py` in my environment (missing async plugin) are identical with and without this change — I diffed the failure lists to confirm nothing new breaks.

### DIFF
--- a/omlx/engine_pool.py
+++ b/omlx/engine_pool.py
@@ -1599,7 +1599,9 @@ class EnginePool:
                     projected = current + admission_size
                     if projected <= evict_target:
                         break
-                    victim = self._find_lru_victim()
+                    victim = self._find_lru_victim(
+                        bytes_needed=projected - evict_target
+                    )
                     if victim is not None:
                         logger.info(
                             f"Evicting '{victim}' to fit '{model_id}' "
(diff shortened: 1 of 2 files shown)


scubamount:
## Verified

- **Root cause correct:** plain LRU is size-blind — can evict multiple small idle models (~1.7 GB TTS) before freeing enough for one large request (68 GB LLM), each eviction paying a reload later.
- **Fix right:** `_pick_eviction_victim` prefers the coldest model that alone covers `bytes_needed`; if none covers, picks the largest to converge in fewer rounds. `bytes_needed <= 0` preserves pure-LRU behavior.
- **Wiring correct:** callers pass `projected - evict_target` (get_engine) and `current + predicted - target` (prefill eviction); freeable size prefers `actual_size` else resident size.
- **Tests:** existing LRU suite still passes; victim selection now size-aware and deterministic on ties.

No issues. Good pressure-aware improvement.

PR #3307, opened 2026-08-29
feat(integrations): run Copilot BYOK launches offline

scaryrawr:
## Summary

- set `COPILOT_OFFLINE=true` when launching Copilot CLI with the custom oMLX BYOK provider
- prevent provider fallback to GitHub-hosted services
- cover the launch environment in integration tests

## Validation

- `UV_PYTHON=python3.12 uv run --python python3.12 pytest tests/test_integrations.py -q`
- `UV_PYTHON=python3.12 uv run --python python3.12 ruff check omlx/integrations/copilot.py tests/test_integrations.py`

### DIFF
--- a/omlx/integrations/copilot.py
+++ b/omlx/integrations/copilot.py
@@ -9,7 +9,7 @@ from omlx.utils.install import get_cli_command_prefix
 
 
 class CopilotIntegration(Integration):
-    """Copilot integration using custom provider BYOK environment variables."""
+    """Copilot integration using custom provider BYOK offline mode."""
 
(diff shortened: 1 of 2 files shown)


scubamount:
## Verified

- **Change:** sets `COPILOT_OFFLINE=true` in the Copilot BYOK launch env so the CLI never phones home when pointed at a local oMLX endpoint.
- **Tests:** both launch paths (with and without model) assert the new var; docstring updated to match.

No issues. Two-line behavioral fix with test coverage.

PR #3311, opened 2026-08-30
fix(memory): warn when a custom ceiling is set but tier is not custom

### LINKED ISSUE #2928: memory_guard_custom_ceiling_gb silently ignored unless tier=custom — warn at startup
**Describe the bug**

Setting a custom memory ceiling in `settings.json` has no effect unless the guard tier is also set to `custom` — and nothing warns about it:

```json
"memory": {
  "prefill_memory_guard": true,
  "memory_guard_tier": "balanced",
  "memory_guard_custom_ceiling_gb": 60.0
}
```

With this config the enforcer starts with its own computed ceiling instead of 60GB:

```
omlx.process_memory_enforcer - INFO - Process memory enforcer started (tier=balanced, ceiling=191.8GB, interval=1.0s)
```

(on a 256GB machine — i.e. effectively no protection, which in our case let a leaking server balloon to 208GB and take down the whole machine via jetsam, see #1691)

The restriction is only documented in the `ProcessMemoryEnforcer.__init__` docstring ("Only used when tier == custom"). Users editing `settings.json` directly have no way to discover it short of reading `process_memory_enforcer.py`.

**Expected behavior**

At minimum, log a WARNING at startup when `memory_guard_custom_ceiling_gb > 0` but `tier != custom`, e.g.:

```
Custom ceiling 60.0GB is set but ignored because tier is 'balanced' (only used when tier='custom')
```

Optionally also surface the mismatch in the admin Memory settings UI.

**Environment**

- oMLX 0.6.3rc1 (also 0.6.2)
- macOS 26.3, Mac15,14

yanzhaohui1999:
## Problem

`memory_guard_custom_ceiling_gb` is only consumed when `memory_guard_tier == "custom"`. That semantic lives solely in the `ProcessMemoryEnforcer.__init__` docstring, so a user editing `settings.json` directly can set a ceiling that is **silently ignored**: the enforcer runs on its computed limit instead, and nothing warns about it. In the reporter's case (issue #2928; the linked jetsam timeline in #1691) a leaking server ballooned to 208 GB on a 256 GB machine because the intended 60 GB ceiling never participated in any computation — at `tier=balanced` the ceiling is derived from live system memory (191.8 GB in the report), so every watermark stayed far below the leak.

## Fix

Log a **WARNING** at enforcer construction when a custom ceiling is set but the tier is not `"custom"`, so the misconfiguration is visible at startup instead of after the fact:

```
Custom ceiling 60.0 GB is set but ignored because memory_guard_tier is 'balanced' (custom ceiling is only used when tier='custom')
```

No behavior change otherwise. The tier is **deliberately NOT auto-switched** to `"custom"`: silently changing protection behavior would be worse than silently ignoring the setting (e.g. a too-small ceiling would start rejecting model loads the user never intended to restrict). If a user wants the 60 GB ceiling to matter, they set `memory_guard_tier: "custom"` — now they can see that from the log.

## Verification

- **Unit tests** (new `TestCustomCeilingWarning`): balanced+ceiling → warning; custom+ceiling → **no** warning (no false positives for correct configs); ceiling=0 → no warning. All pass; full enforcer/memory suite **254 passed**.
- **Full gate**: `pytest tests/ -m "not slow and not integration"` → 10567 passed, 6 failed — all six are the known environment-only failures already tracked (cluster SSH supervisor ×2, cluster seams, GDN sidecar ×2, vlm_mtp thinking-budget e2e; reproduced independently of this change)
- **Smoke**: real `--base-path` server with `tier=balanced` + `ceiling=60` logs the warning at startup with the exact wording above

Closes #2928

## Open question

The issue optionally suggests also surfacing the mismatch in the admin Memory settings UI (hint/grey-out on the custom-ceiling field when `tier != custom`). I left that out of this PR on purpose — say the word and I'll add it (note: UI strings go through the 9-language i18n pipeline per repo convention, so it is a slightly larger change).

### DIFF
--- a/omlx/process_memory_enforcer.py
+++ b/omlx/process_memory_enforcer.py
@@ -367,6 +367,22 @@ class ProcessMemoryEnforcer:
         self._memory_guard_custom_ceiling_bytes = max(
             0, int(memory_guard_custom_ceiling_gb * 1024**3)
         )
+        if (
+            memory_guard_custom_ceiling_gb > 0
+            and self._memory_guard_tier != "custom"
+        ):
+            # Users editing settings.json directly have no way to discover
+            # that the custom ceiling only applies to tier == "custom" (it
+            # is documented only in this __init__ docstring), so a mis-set
+            # ceiling silently no-ops and the enforcer runs on its computed
+            # limit instead (issue #2928). Warn loudly at construction.
+            logger.warning(
+                "Custom ceiling %.1f GB is set but ignored because "
+                "memory_guard_tier is '%s' (custom ceiling is only used "
+                "when tier='custom')",
+                memory_guard_custom_ceiling_gb,
+                self._memory_guard_tier,
+            )
         self._active_poll_interval = poll_interval
         self._loaded_idle_poll_interval = 10.0
         self._unloaded_idle_poll_interval = 30.0
--- a/tests/test_process_memory_enforcer.py
+++ b/tests/test_process_memory_enforcer.py
@@ -2,6 +2,7 @@
 """Tests for ProcessMemoryEnforcer."""
 
 import asyncio
+import logging
 from contextlib import suppress
 from types import SimpleNamespace
 from unittest.mock import AsyncMock, MagicMock, patch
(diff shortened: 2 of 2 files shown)


scubamount:
## Verified

- **Root cause correct:** `memory_guard_custom_ceiling_gb` only consumed when `tier == "custom"`; documented only in the `__init__` docstring. Direct `settings.json` edits silently no-op → ceiling ignored → server balloons (issue #2928: 208 GB on 256 GB machine vs intended 60 GB).
- **Fix right:** warning at construction when ceiling > 0 and tier != "custom". Deliberately does NOT auto-switch tier — silently changing protection behavior would be worse than a warning.
- **Tests:** `TestCustomCeilingWarning` covers three cases — balanced+ceiling warns, custom+ceiling does not (no false positive), zero ceiling does not. 10,567 tests pass (6 pre-existing env-only failures tracked separately).
- No new dependencies, no behavior change for correct configs.

No issues. Good defensive warn-on-misconfiguration pattern.

PR #3312, opened 2026-08-30
macOS: Show live serving activity and stats

evgri243:
Adds live serving observability to the native macOS app and menubar.

## What changed

- Show live per-request activity on Models and Status, including prefill progress, generation, non-streaming work, queue state, and short completion linger.
- Add LIV, AVG, and ALL menubar metric items with live PP/TG rates, activity graphs, and scoped details.
- Add a compact Serving Stats menu with Live Activity, Current Requests, Average Session, and All Time summaries.
- Keep Current Requests readable under load by grouping by model, capping visible requests, and using consistent PP/TG markers and rate/token formatting.
- Align serving-stat panels with the existing System Stats layout and reuse shared presentation components.

## Data and refresh behavior

- Decode `/admin/api/activity` through the existing macOS API client.
- Display raw prefill and generation rates consistently across the app and menubar; calculate prefill ETA server-side from a positive-sample EMA to avoid ETA jitter.
- Preserve idle versus unavailable state, retain short-lived completed app rows, and poll admin-backed activity only while a consumer needs it.

image

### DIFF
--- a/apps/omlx-mac/Resources/Localizable.xcstrings
+++ b/apps/omlx-mac/Resources/Localizable.xcstrings
@@ -1091,6 +1091,30 @@
         }
       }
     },
+    "menubar.metric.metrics" : {
+      "comment" : "Header above scalar counters in an AVG or ALL serving metric popover",
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Serving Stats"
+          }
+        },
+        "ru" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Статистика обслуживания"
+          }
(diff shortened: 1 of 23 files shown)


scubamount:
## Review

- **Feature:** live per-request progress pills (Prefill/Generating/Done) on model rows in the macOS app — decodes the existing admin activity payload through the API client with polling; no server change needed.
- **i18n:** all new strings added to `Localizable.xcstrings` with en/ru/zh-Hans translations and usage comments — consistent with the app's existing localization discipline.

Caveats: no CI defined for this path (apps/ not covered by the Python workflow), and I verified structure rather than running the Swift UI. The polling interval / payload decode is the part to smoke-test against a busy server.

No blocking issues found in the diff.

PR #3313, opened 2026-08-30
fix(codex): generate local model metadata catalog

### LINKED ISSUE #2050: omlx launch codex_app does not generate a Codex model_catalog_json
# omlx launch codex_app does not generate a Codex model_catalog_json

## Problem

When launching Codex Desktop via:

```bash
omlx launch codex_app
```

oMLX appears to launch/configure Codex App, but it does not generate or refresh a Codex-compatible `model_catalog_json` file.

Because of that, users can manually point Codex to the oMLX provider/model, but there is no generated model catalog that exposes available local oMLX models to Codex's model catalog / model picker workflow.

Example manual config currently required:

```toml
model = "local-model-id"
model_provider = "omlx"
model_catalog_json = "/Users/<user>/.codex/omlx-models.json"
```

However, `omlx launch codex_app` does not appear to create a new oMLX-specific catalog JSON such as:

```text
~/.codex/omlx-models.json
```

## Expected behavior

`omlx launch codex_app` should generate or refresh a Codex-compatible model catalog JSON containing the local oMLX models, and configure a modern Codex profile file to reference it.

For Codex 0.134.0 and later, the generated profile should use the standalone profile file format:

```toml
# ~/.codex/omlx.config.toml

model = "local-model-id"
model_provider = "omlx"
model_catalog_json = "/Users/<user>/.codex/omlx-models.json"
```

The generated catalog could include models discovered by oMLX, for example:

```json
{
  "models": [
    {
      "id": "local-model-id",
      "name": "Local oMLX Model"
    }
  ]
}
```

## Codex profile behavior

According to the official Codex advanced configuration documentation:

- Profile files are layered above the base user config and below project/CLI config.
- Profile files only need to contain values that differ from the base config.
- Profile files can override `model_catalog_json`; when both the base config and the profile config set it, Codex uses the profile value.
- In Codex 0.134.0 and later, `--profile` no longer reads `[profiles.profile-name]` from `config.toml`.
- The top-level `profile = "profile-name"` selector is no longer supported.
- Legacy profile settings should be moved to `~/.codex/profile-name.config.toml`.

Reference:
https://developers.openai.com/codex/config-advanced

Given this, oMLX's Codex App launcher should ideally generate:

1. a modern Codex profile file, for example `~/.codex/omlx.config.toml`, and
2. a Codex-compatible local model catalog JSON, for example `~/.codex/omlx-models.json`.

This would match the current Codex 0.134.0+ profile model and avoid relying on deprecated `[profiles.profile-name]` or `profile = "profile-name"` settings in `config.toml`.

## Actual behavior

No oMLX-specific `model_catalog_json` file seems to be generated by `omlx launch codex_app`.

As a result, users still need to manually edit `~/.codex/config.toml` or a profile config and restart/relaunch Codex when switching between OpenAI models and local oMLX models.

## Why this matters

Codex supports `model_catalog_json` as a startup model catalog override. Without a generated catalog file, the oMLX Codex App integration is less complete than expected for users who want to expose multiple local oMLX models to Codex Desktop / Codex model selection workflows.

This is especially noticeable when users want to switch between:

- OpenAI provider/model
- oMLX provider/model

without manually editing config files each time.

## Environment

- macOS
- Command used: `omlx launch codex_app`
- Codex app / CLI: current version unknown
- oMLX version: current version unknown
- Example model: any local oMLX model

## Related context

This is related to improving the native Codex App launcher/integration experience. A useful outcome would be for `omlx launch codex_app` to behave more like a first-class Codex integration by generating both:

1. a Codex profile config, and
2. a Codex-compatible local model catalog JSON.

njern:
## Summary

Generate a local Codex model metadata catalog from models exposed by oMLX. This prevents Codex from falling back to unknown-model defaults that can disable tools or apply incorrect context-window behavior.

- Generate `omlx-models.json` with Codex-compatible model metadata.
- Join visible `/v1/models` IDs with detailed `/v1/models/status` metadata.
- Preserve model aliases and exposed profiles.
- Exclude hidden helpers and non-chat models such as embeddings.
- Configure Codex CLI with process-scoped arguments, without modifying the user's normal Codex configuration.
- Configure Codex App through a standalone `omlx.config.toml` profile and launch it with `--profile omlx`.
- Respect `CODEX_HOME` for the generated profile and catalog.
- Leave the user's base `config.toml` unchanged.
- Require an explicit `--model` when model-type metadata is unavailable, preventing unsafe fallback selection.
- Ensure Claude-only tier flags do not affect Codex model selection.

## Validation

- `python -m pytest tests/test_integrations.py tests/test_cli.py -q` — 212 passed on Python 3.13.
- `git diff --check`
- Verified Codex CLI against a locally served oMLX model:
  - loaded the generated catalog without the unknown-model metadata warning;
  - invoked the shell tool successfully.
- Verified that Codex can load the generated standalone `omlx` profile and complete a request.
- Unit tests verify that Codex App launches with `--profile omlx`, respects `CODEX_HOME`, and preserves the base Codex configuration.

Fixes #2050

### DIFF
--- a/omlx/cli.py
+++ b/omlx/cli.py
@@ -446,48 +446,89 @@ def launch_command(args, extra_args: list[str] | None = None):
     if api_key:
         headers["Authorization"] = f"Bearer {api_key}"
 
-    # Pre-fetch model status (context_window, max_tokens, model_type per model)
+    # Pre-fetch detailed model metadata. Codex also needs the authoritative
+    # visible model IDs from /v1/models to build its local model catalog.
     models_status_map: dict[str, dict] = {}
+    models_status_available = False
     try:
         resp = requests.get(f"{base_url}/v1/models/status", headers=headers, timeout=5)
         if resp.ok:
-            for m in resp.json().get("models", []):
-                if m_id := m.get("id"):
-                    models_status_map[m_id] = m
-                if model_alias := m.get("model_alias"):
-                    models_status_map[model_alias] = m
+            status_models = resp.json().get("models")
+            if not isinstance(status_models, list):
+                raise ValueError("invalid model status response")
+
+            parsed_status_map: dict[str, dict] = {}
+            for m in status_models:
+                if not isinstance(m, dict):
+                    raise ValueError("invalid model status entry")
+                m_id = m.get("id")
+                model_type = m.get("model_type")
+                if (
+                    not isinstance(m_id, str)
+                    or not m_id
+                    or not isinstance(model_type, str)
+                    or not model_type
+                ):
+                    raise ValueError("incomplete model status entry")
+                parsed_status_map[m_id] = m
+                # Exposed profiles inherit their source model's alias in the
+                # status payload. Only the physical model should own that alias.
+                if not m.get("source_model_id") and (
+                    model_alias := m.get("model_alias")
+                ):
+                    parsed_status_map[model_alias] = m
+            models_status_map = parsed_status_map
+            models_status_available = True
     except Exception:
         pass
+
+    def _fetch_visible_models() -> list[str]:
+        try:
+            response = requests.get(f"{base_url}/v1/models", headers=headers, timeout=5)
+            response.raise_for_status()
+            visible_ids = [m["id"] for m in response.json().get("data", [])]
+            if not models_status_available:
+                return visible_ids
+            return [
+                model_id
+                for model_id in visible_ids
+                if models_status_map.get(model_id, {}).get("model_type")
+                in ("llm", "vlm")
+            ]
+        except Exception:
+            return []
+
+    visible_models: list[str] | None = None
+    if tool_name in ("codex", "codex_app"):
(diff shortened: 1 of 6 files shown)


scubamount:
## Verified

- **Root cause correct:** Codex falls back to unknown-model defaults (tools disabled / wrong context window) when oMLX models are not in its metadata catalog.
- **Fix right:** builds `omlx-models.json` by joining visible `/v1/models` IDs with `/v1/models/status` metadata; strict validation raises on malformed status entries instead of silently building a partial map; alias ownership restricted to physical models (`not m.get("source_model_id")`), profiles excluded.
- **Fail-closed:** when status metadata is unavailable, Codex launch refuses and demands explicit `--model` rather than guessing unsafe defaults.
- **Scope:** process-scoped args for Codex CLI, standalone `omlx.config.toml` profile for the App, `CODEX_HOME` respected, user base `config.toml` untouched.

No issues.

PR #3315, opened 2026-08-30
perf(glm5): accelerate recurrent and NoPE prefill

DiscoStew6082:
## Summary

GLM-5.3 prefill still spends substantial GPU time in vector-gated recurrence and sparse attention. This independent patch adds the smallest validated subset of #3251:

- process vector-gated recurrent rows in blocks of four;
- skip zero positional-embedding work in sparse MLA attention.

In additive testing with #3304's native group-64 Q6 dispatch enabled in both arms, the final patch improves complete-model prefill by 1.92% at 2K, 2.92% at 4K, 3.19% at 8K, and 4.07% at 16K.

## Attribution

The surviving R=4 recurrence and NoPE implementations were originally authored by @jonathan308 in #3251:

- R=4 recurrence: `60661435e76b12db8cf95da3fc3bb5108faf66f1`
- NoPE sparse MLA: `961db3585eaadce1a276adaf7f8104ec83b7cbbe`

Jonathan is recorded as a co-author of this commit. My contribution is the component-by-component ablation with #3304 enabled as the control, independent integration onto current `main`, dispatch and correctness verification, benchmark validation, and selection of the minimal surviving patch.

## Component ablation

| #3251 component | Measured prefill dispatch | Result over #3304 | Decision |
| --- | --- | --- | --- |
| R=4 recurrence | Candidate R=4: 102 calls; baseline R=1: 102 calls | Repeatable gains at every tested length; 3/3 wins in the clean confirmation runs | Keep |
| NoPE attention | 11 native NoPE calls at 4K, 8K, and 16K; does not dispatch at 2K | +1.06%, +1.31%, and +0.88% median; 3/3 wins | Keep |
| 8K wide-prefill scheduler | Candidate used one 4K bulk call instead of the default 2K chunks | Failed deterministic output equality at 4K | Exclude |
| M=1 DSA indexer | Bulk prefill calls use M=512; M=1 occurred only for the two generated tokens | Decode-only for this workload, not a bulk-prefill improvement | Exclude |

No throughput claim is retained for the rejected wide-prefill path because it failed the output-equality gate. The indexer specialization was not benchmarked as a prefill candidate after its dispatch gate showed that M=1 was not used during bulk prefill.

## Additive whole-model results over #3304

Prompt throughput in tokens/second:

| Tokens | #3304 baseline | R=4 + NoPE | Median change | Range | Wins |
| ---: | --- | --- | ---: | --- | ---: |
| 2,048 | 467.01, 467.00, 467.08 | 475.98, 476.03, 475.14 | +1.92% | +1.73% to +1.93% | 3/3 |
| 4,096 | 437.86, 437.47, 437.67 | 448.37, 450.97, 450.47 | +2.92% | +2.40% to +3.09% | 3/3 |
| 8,192 | 443.23, 443.00, 442.61 | 457.37, 456.99, 457.31 | +3.19% | +3.16% to +3.32% | 3/3 |
| 16,384 | 441.68, 441.67, 441.93 | 459.65, 459.86, 459.89 | +4.07% | +4.06% to +4.12% | 3/3 |

Each variant received one warm-up followed by three paired AB/BA measurements using identical deterministic prompts, temperature 0, and two generated tokens. Prompt cache, APC, draft models, and other cache reuse were disabled; every run reported `cached_tokens=0`. Baseline and candidate outputs were identical.

The real `mlx-vlm` `wired_limit` context was active. The benchmark used #3304 head `b095c75b4e38ca1deabcd649711277d168f32be1` in both arms, with 42 native switch calls and zero fallbacks per measured run. Candidate counters also confirmed R=4 recurrence and NoPE dispatch at the expected sequence lengths. These measurements were not rerun against #3304's later rebased head.

These are complete-model prefill measurements. Primitive microbenchmarks are not presented as whole-model multipliers.

## Validation

- Independent native extension build passed and exposed the dedicated NoPE symbol.
- Focused recurrence, GLM MoE/DSA, and GLM-5 compatibility suites: 51 passed, with 6 expected capability skips.
- Current-head hosted CI passed on Python 3.11, 3.12, and 3.13; each job reported 13,048 passed, 370 skipped, and 82 deselected.
- `git diff --check` passed.
- `git range-diff` from pre-rebase head `556f9817` to rebased head `59a02a9f` shows only current-base context movement in the native-symbol inventory; the R=4 and NoPE implementation is unchanged.

## Independence and scope

This revision was rebased onto `cb62779b10f5d75e11608bcd3f7de87d248d156f` and has no source dependency on #3304. The benchmark matrix intentionally kept the stated #3304 head enabled in both arms to measure this patch's additional value after the affine Q6 optimization. The patch excludes #3251's wide-prefill scheduler and M=1 indexer specialization.


scubamount:
## Verified

- **Correctness:** accelerates recurrent and NoPE prefill — (1) recurrent NoPE/GLU flatten with `Kv>=4096` gate, `D_PE=64` zero fallback, bitwise equality vs zero-PE native. (2) `gated_delta` R=4 row blocking hoists `g`/`beta`/`q`/`k`, `grid=(32, Dv//R, B*Hv)`, `R` auto by `Dv%4/2/1`, masked path zeros all R rows, equality to R=1 proven.
- **Perf:** ablated +1.9–4.1% prefill, 3/3 wins, outputs identical. CI 3/3 green.

No issues. Verified perf with bitwise correctness.

PR #3321, opened 2026-08-30
fix(grammar): bound JSON schema whitespace

### LINKED ISSUE #2990: Structured output: grammar-constrained JSON can run to max_tokens in an unbounded whitespace loop (compile_json_schema without max_whitespace_cnt)
### Summary

`_compile_bare_grammar` in `omlx/server.py` calls xgrammar's `compile_json_schema(schema_str)` with
neither `any_whitespace` nor `max_whitespace_cnt`, so the compiled grammar permits **unbounded
whitespace anywhere in the JSON**. A model that terminates a JSON string earlier than it intended is
then trapped: every token carrying the text it wants to write next is masked out, the space token is
not, and whitespace is legal after whitespace. Generation runs to `max_tokens` and returns a
truncated object with `finish_reason: length`.

Measured on oMLX 0.6.2 / xgrammar 0.2.3 / `Qwen3.6-35B-A3B-UD-MLX-4bit` (Apple M4 Pro), driving a
memory-consolidation workload. **23 % of calls (60 of 262 overnight) hit the ceiling exactly**, at a
cost of ~2 GPU-hours and ~960 000 output tokens in a single night. The distribution is bimodal with
nothing in between — healthy calls topped out at 6 381 output tokens, failing ones sat at 16 000 to
the token.

### What a failing call looks like

16 000 completion tokens produce **104 characters** of content. Streaming the deltas shows where the
rest goes:

```
"{\n  \"creates\": [", "\n    {\n      \"text", "\": \"Fehldiagn", "ose-Falle bei scheinbaren",
" Wiederholungen: Drei „Vers", "uche\"", "                                    … (to the ceiling)"
```

### Trigger

The source text being copied contains a **mismatched quote pair** — a German opening `„` (U+201E)
closed with a bare ASCII `"`. The prompt asks the model to reproduce source text verbatim, so it
emits the ASCII `"` unescaped. Inside a JSON string that character *is* the terminator, the grammar
accepts it, the value ends mid-sentence, and the model is now in a parser state where its intended
continuation is illegal.

This is not exotic: any prompt that asks a model to quote user text can hit it.

### Reproduction — no server needed

```python
import json, xgrammar as xgr
from xgrammar.testing import _is_grammar_accept_string

schema = {"type": "object",
          "properties": {"creates": {"type": "array",
                                     "items": {"type": "object",
                                               "properties": {"text": {"type": "string"}},
                                               "required": ["text"]}}},
          "required": ["creates"]}
prefix = '{"creates": [{"text": "abc"'

for kw in ({}, {"any_whitespace": False}, {"max_whitespace_cnt": 32}):
    g = xgr.Grammar.from_json_schema(json.dumps(schema), **kw)
    print(kw, _is_grammar_accept_string(g, prefix + " " * 4000, require_termination=False))
# {}                        -> True    <-- what oMLX builds today
# {'any_whitespace': False} -> False
# {'max_whitespace_cnt': 32}-> False
```

### End-to-end, same request three ways

| Request | Result |
|---|---|
| `temperature=0.0`, **no** `response_format` | `stop`, 1 806 tokens, complete valid JSON |
| `temperature=0.0`, **with** strict `json_schema` | 6 000 tokens, `length`, 104 chars |
| identical, only the 5 bare `"` in the source text swapped for `“` | `stop`, 1 754 tokens, valid JSON |

Same model, same temperature, same prompt. Byte-for-byte reproducible on replay.

### Things that do *not* help (each measured)

- `repetition_penalty` 1.05 and 1.2, sent directly to oMLX: **byte-for-byte identical output**.
  `frequency_penalty` likewise. A non-cumulative penalty cannot break an attractor this dominant.
- Raising temperature: 0.3 failed 3/3, 0.7 failed. Not a mitigation.
- Lowering `max_tokens` only makes the failure cheaper.

### Suggested fix

Pass one of the whitespace bounds through to xgrammar in `_compile_bare_grammar`, ideally as a
server/model setting so operators can choose:

```python
return compiler.compile_json_schema(schema_str, max_whitespace_cnt=32)
```

Locally patched exactly like that (with a `TypeError` fallback for older xgrammar), oMLX restarted,
same request replayed twice: **1 315 tokens, `finish_reason: stop`, valid JSON, 28–34 s** — against
6 000 tokens / `length` / 130 s before. Plain chat and embeddings unaffected.

`max_whitespace_cnt` is preferable to `any_whitespace=False` here because it accepts every output
the current grammar accepts and only forbids the unbounded run, whereas `any_whitespace=False`
additionally forces compact separators.

Two paths remain unbounded and may deserve the same treatment:

- `json_object` mode goes through `compile_builtin_json_grammar()`, which takes no parameters at all.
- `_compile_with_structural_tag` (models with a `reasoning_parser`) goes through
  `compile_structural_tag`.

### Prior art

This is a known class of bug in other serving stacks, and the mitigation is already standardised
elsewhere:

- sglang [#8250](https://github.com/sgl-project/sglang/issues/8250) — *"when xgrammar restricts the
  token space to valid JSON tokens, the model's slight repetition bias becomes a strong loop"*.
- vLLM [#40080](https://github.com/vllm-project/vllm/issues/40080) — infinite repetition loops under
  JSON-schema structured output.
- vLLM already ships `--structured-outputs-config.disable_any_whitespace` for this.

`grep -ri whitespace` over the oMLX package finds nothing, and the tap history has no related commit,
so I believe oMLX currently has no knob for it.

### Environment

oMLX 0.6.2 (Homebrew tap), xgrammar 0.2.3, macOS 26.6.1, Apple M4 Pro 64 GB,
`Qwen3.6-35B-A3B-UD-MLX-4bit`, no `reasoning_parser` configured for the model, request via
`/v1/chat/completions` with `response_format: {"type": "json_schema", "json_schema": {"strict": true, …}}`.

saichowdary007:
Fixes #2990

JSON-schema constrained decoding currently compiles with xgrammar's unlimited whitespace. If a model closes a string early, whitespace remains valid while useful tokens are masked, so generation can spin until `max_tokens` and return a truncated response.

This passes `max_whitespace_cnt=32` for JSON-schema grammars only. EBNF, regex, and existing built-in JSON grammar paths are unchanged. The focused grammar tests now assert the compiler option.

Verification:
- `python3 -m py_compile omlx/server.py tests/test_grammar.py`
- AST-level regression check confirmed the compiled schema receives `max_whitespace_cnt=32`
- `git diff --check` clean

The repository pytest process was unable to complete in this Apple Silicon environment because importing the optional native xgrammar stack stalled; CI should exercise the full test matrix.

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -4277,6 +4277,9 @@ def _compile_with_structural_tag(
     return compiler.compile_structural_tag(tag_dict)
 
 
+_JSON_SCHEMA_MAX_WHITESPACE_CNT = 32
+
+
 def _compile_bare_grammar(compiler, fmt: dict):
     """Compile a grammar without any structural tag wrapping."""
     if fmt["type"] == "json_schema":
@@ -4286,7 +4289,15 @@ def _compile_bare_grammar(compiler, fmt: dict):
         if not schema:
             return compiler.compile_builtin_json_grammar()
         schema_str = _json.dumps(schema) if isinstance(schema, dict) else schema
-        return compiler.compile_json_schema(schema_str)
+        # An unlimited whitespace run can trap a grammar-constrained decoder
+        # after an early string termination: whitespace remains valid while
+        # every useful token is masked.  Keep the bound local to the compiler
+        # call so ordinary JSON and user-supplied EBNF/regex grammars retain
+        # their existing behavior.
+        return compiler.compile_json_schema(
+            schema_str,
+            max_whitespace_cnt=_JSON_SCHEMA_MAX_WHITESPACE_CNT,
+        )
     elif fmt["type"] == "grammar":
         return compiler.compile_grammar(fmt["grammar"])
     elif fmt["type"] == "regex":
--- a/tests/test_grammar.py
+++ b/tests/test_grammar.py
@@ -418,7 +418,9 @@ class TestCompileBareGrammar:
         schema = {"type": "object"}
         result = self._call(compiler, {"type": "json_schema", "json_schema": schema})
         assert result == "compiled_json"
-        compiler.compile_json_schema.assert_called_once()
+        compiler.compile_json_schema.assert_called_once_with(
+            json.dumps(schema), max_whitespace_cnt=32
+        )
 
     def test_empty_json_schema(self):
         compiler = MagicMock()
@@ -482,7 +484,12 @@ class TestCompileGrammarForRequest:
             "json": {"type": "object", "properties": {"x": {"type": "integer"}}},
         })
         assert result == "compiled_json"
-        compiler.compile_json_schema.assert_called_once()
+        compiler.compile_json_schema.assert_called_once_with(
+            json.dumps(
+                {"type": "object", "properties": {"x": {"type": "integer"}}}
+            ),
+            max_whitespace_cnt=32,
+        )
 
     def test_bare_regex(self):
         compiler = MagicMock()
@@ -573,7 +580,9 @@ class TestCompileGrammarForRequest:
             reasoning_parser=None,
         )
         assert result == "compiled_bare"
-        compiler.compile_json_schema.assert_called_once()
+        compiler.compile_json_schema.assert_called_once_with(
+            json.dumps({"type": "object"}), max_whitespace_cnt=32
+        )
         compiler.compile_structural_tag.assert_not_called()
 
     def test_compilation_error_raises_for_structured_outputs(self):


scubamount:
## Verified

- **Root cause correct:** xgrammar's unlimited whitespace lets a model close a string early, then spin on whitespace tokens until `max_tokens` → truncated output.
- **Fix scoped correctly:** `max_whitespace_cnt=32` applies to JSON-schema grammars only; EBNF, regex, and built-in JSON paths unchanged.
- **Tests assert the compiled option directly** via AST-level regression check on the compiled schema.
- **CI green** across 3.11/3.12/3.13.

No issues. Tight, well-scoped fix with regression coverage.

PR #3326, opened 2026-08-31
Add opt-in token and text inspection files to the paged SSD cache

actualrealhuman:
NEW FEATURE: Plain .txt sidecars for cached token blocks so the user can see
what is using their SSD space!

The paged SSD cache can occupy substantial disk space without making its
contents recognizable. This adds optional `.tokens` and `.txt` companions
beside each main cache block so users can identify what the stored model state
represents and make informed retention decisions. It describes persisted cache
entries; it does not introduce a request log or change cache block sizes,
matching, or admission policy.

- Capture is off by default. Enable it with **Save Cache Inspection Files** in
  the dashboard, `--cache-inspection`, `OMLX_CACHE_INSPECTION=true`, or
  `cache.cache_inspection` in settings.
- `.tokens` contains versioned compact JSON with exact token IDs, model and
  tokenizer identity, parent hash, and block offsets. `.txt` provides an
  annotated, lossy view with the model identifier, parent hash, token count,
  and exact zero-based range in its header.
- Decoding uses an independent CPU tokenizer snapshot and normally runs on the
  existing SSD writer. Available image dimensions and existing cache-key hashes
  can be included; there is no OCR, captioning, extra media processing, or model
  inference.
- Companions follow the main block's persistence and eviction lifecycle, count
  toward the SSD budget, and are cleaned up on clear/orphan recovery. Missing
  companions are backfilled opportunistically when tokens become available
  again, without rewriting tensors. Completed companions are not refreshed on
  cache hits or renderer upgrades.
- Individual file replacements are atomic; the three-file set is not a
  transaction. Inspection failures preserve usable KV data. Queued writes are
  guarded against eviction or replacement of their block.

Both companion formats expose recoverable prompt content, including system
instructions and tool results. The setting warns about this, files use
owner-only permissions, and displayed control characters are escaped. Turning
capture off stops new capture; existing files remain until eviction or clearing.

`docs/cache-inspection.md` explains the format, block-boundary limitations, and
how separate Gated DeltaNet checkpoints relate to source-block hashes. GDN,
distributed, vision-feature, and speculative-draft cache storage are unchanged.
There is no cache browser, deletion UI, request attribution, or new dependency.

Validation was performed on Apple Silicon with Python 3.11.15 and the pinned
MLX 0.32.0 / mlx-lm / mlx-vlm revisions. The focused inspection, engine, and
locale checks pass (121 tests). Tests cover exact IDs and offsets, special-token
rendering, media context, default-off and RAM-only behavior, backfill, I/O
failures, quota accounting, clear/eviction races, and compatibility with the
engine's existing positional arguments. Static checks, the final macOS release
build, deep signature verification, and a bundled-runtime smoke test pass.
The latter verifies source/asset parity, persistence, restoration, and clearing
with synthetic data. A live hybrid-model test confirmed that a
2932-token synthetic prompt produces a 2048-token block with both companions.

The full local CI selection (`pytest tests/ -m 'not slow and not integration'`)
finishes with **10603 passed, 5 failed, 174 skipped, and 77 deselected**.
All five failures are numerical tests that also fail in an untouched checkout of upstream
`e008a66b` using the same local environment; no tolerances or model code are
changed in this PR:

- `test_dflash_laguna.py::test_target_ops_logits_last_only_slices_before_lm_head`
- `test_mlx_vlm_glm5_next_compat.py::test_variable_length_batch_logits_match_single_requests`
- `test_mlx_vlm_qwen4_exp_compat.py::test_qwen4_gathered_qsa_prefill_matches_official_mask_path`
- `test_mlx_vlm_qwen4_exp_compat.py::test_qwen4_lightning_mtp_fusion_and_runtime_attachment`
- `test_qwen4_qsa_decode_gather.py::test_qwen4_language_wrapper_routes_2d_text_positions_to_gather`

Hosted [CI](https://github.com/jundot/omlx/actions/runs/33349015858) passes on
Python 3.11 and 3.12. Python 3.13 reports 10607 passed and one failure in
`test_prefix_cache_gdn_split.py::test_split_dedup_recreates_evicted_sidecar`;
the inspection tests pass on all three versions. The five local numerical
failures above do not occur in this hosted run.

A diagnostic using controlled thread timing against unchanged upstream
`e008a66b` reproduces the same GDN assertion (4 cached tokens instead of 12):
`take_staged_file()` can remove the empty request directory after the next
snapshot writer creates it but before that writer opens its temporary file.
The boundary snapshot store is unchanged in this PR. Without controlled timing,
300 local repetitions of the test passed. This supports an existing intermittent
staging race; no unrelated GDN change is included here. A rerun of the failed
job was requested but GitHub rejected it because repository administrator rights
are required, so the Python 3.13 check remains failed pending maintainer action.

The included benchmark uses 64 synthetic blocks of approximately 1 MiB each,
256 tokens per block, and five alternating disabled/enabled repetitions. Final
revision medians, measured separately from test/build jobs:

| Measurement | Disabled | Enabled |
| --- | ---: | ---: |
| Producer time | 3.09 ms | 3.46 ms |
| Total persistence time | 37.43 ms | 59.04 ms |
| File bytes | 67161208 | 67441095 |
| Inline queue fallbacks | 0 | 0 |

This adds approximately 0.42% in file bytes and 0.34 ms of persistence wall time
per block in this synthetic workload (about 1.58 times total persistence time).
These are storage measurements, not inference throughput measurements; tokenizer
snapshot construction is outside the timed section. Capture has a cost, and
the existing inline-write fallback remains possible under queue pressure.


actualrealhuman:
This contribution is ready for review. CI passed on Python 3.11 and 3.12, and the cache inspection tests passed on all three Python versions.
The Python 3.13 job failed in an existing GDN checkpoint test. The same failure was reproduced on unchanged upstream with controlled thread timing; details are in the PR description. I’ve kept that separate from this contribution.
Could a maintainer please rerun the failed job? My account doesn’t have permission to do so. Thanks for reviewing!

scubamount:
## Verified

- **Feature:** opt-in `.tokens`/`.txt` sidecars beside paged SSD cache blocks so users can identify what occupies disk. Off by default, documented privacy note (exposes prompt content when enabled), follows eviction lifecycle. README + docs + benchmark harness included.
- **Design:** `BlockInspection`/`InspectionRenderer` cleanly injected via `inspection_renderer=` param; disabled path passes `None` so the hot path is unchanged when the feature is off.

Notes:
- **CI red on Python 3.13** is shared flakiness, not this change: `test_prefix_cache_gdn_split.py::test_split_dedup_recreates_evicted_sidecar` (`N == 12` assert shape) also fails on unrelated PRs that don't touch `prefix_cache.py` at all (e.g. #3310's 3.11/3.12 legs, same assertion pattern). My first read here blamed the PR because it modifies `prefix_cache.py` — retracted; the cross-PR evidence says flaky test file.
- Sidecar lifecycle note: prompt-content sidecars must be removed on *every* eviction path (LRU, capacity, explicit clear), not just the main-block path — worth a targeted test if one doesn't already exist.

Verdict: PASS (flaky CI legs notwithstanding).

scubamount:
## Correction to my CI note above

The `test_prefix_cache_gdn_split.py` failure I flagged as "not a flake" appears to be a genuinely flaky test file: PR #3310 (which touches neither `prefix_cache.py` nor that test) fails in the SAME file with the same `N == 12` assertion shape (`test_split_restore_retry_budget_is_one_per_block`, `4 == 12`), and a third sibling test fails on the 3.12 leg of that run. Same assert pattern, different tests, different PRs, different Python legs.

Revised verdict: the CI red on this PR is most likely that shared flakiness, not your change. Retract the "please investigate" ask on that point; a re-run of the 3.13 leg should settle it. The sidecar-evlifecycle test suggestion stands.

(Worth a separate issue for the repo, though: three tests in that file showing `actual < 12` block-restoration behavior under CI suggests a real order/timing dependence hiding there.)

PR #3328, opened 2026-08-31
perf(qwen4): prime verified MTP drafter from cached suffix

jonathan308:
## Summary

This PR is the separately reviewable cached-drafter follow-up to #3320. A backbone prefix-cache/SSD restore can recover the target cache while the memory-only Lightning-MTP head history is absent. Previously Qwen4 entered decode with `primed=0`, paying lower early acceptance and avoidable agent-turn latency.

This is a capability-scoped **text-token suffix optimization**, not a text-only Qwen implementation. Native Qwen3.8-Flash-Next text, image, and video serving was merged in #3169 and remains unchanged. Image/video requests continue through that official multimodal path; this optimization detects media state and fails closed instead of folding vision/video embeddings into the text MTP head.

## What changed

- Adds verified suffix-local MTP priming for eligible Qwen4 B1 text-token cache restores.
- Proves each incoming suffix chunk is the exact scheduler-owned token slice before folding it into the drafter.
- Keeps local head-history offsets separate from absolute target-cache offsets.
- Requires every readable target cache layer to agree on the same absolute offset.
- Exercises ordinary reject, partial accept, full accept, rollback, boundary materialization, request replacement/interleaving, and B1-to-batch reconciliation.
- Explicitly tests that image/video media plans skip suffix-local priming and retain the existing multimodal serving path.
- Never publishes suffix-local state as a full-history MTP sidecar; media, batched, misaligned, and otherwise unsupported cases fail closed to existing behavior.

## Relationship to #3330

#3330 is the generic Metal/ExactResident/prompt-tail engine. This PR provides one Qwen-specific prerequisite that #3330 can consume: reconstructing verified MTP head history from an exact cached suffix.

The PRs are deliberately independently mergeable and independently reviewable:

- **#3328** changes only Qwen4 cached-suffix drafter reconstruction.
- **#3330** changes the generic local engine/cache/idle-materialization framework and fails closed for unsupported speculative models.

Full speculative-Qwen ExactResident/prompt-tail reuse also requires the separate exact target-terminal transaction proven in Fusion. That transaction is intentionally not hidden inside either PR; until its own review unit lands, #3330 remains fail closed for speculative Qwen. This is an explicit composition relationship, not an assertion that #3328 alone unlocks the full 6.7× Fusion TTFT result.

## Physical validation

Apple M3 Ultra, Qwen3.8-Flash-Next-oQ4e-MTP, deterministic 100K restart restore:

- 98,304 prompt tokens restored from the target prefix cache.
[...]

Issue #14, opened 2026-02-15
Add support for multiple model directories


intobigdata:
I'm running oMLX as part of a multi-backend LLM gateway on a Mac Studio with models spread across multiple external volumes:

  Currently --model-dir only accepts a single directory, so oMLX can only discover models from one volume at a time. This means ~2/3 of the model library is invisible to oMLX unless I manually consolidate everything onto one drive (not practical at 8TB+ total).

  Proposed behavior:

  Support multiple model directories via comma-separated paths or a list in settings:

  # CLI
  omlx serve --model-dir '/Volumes/A/models,/Volumes/B/models,/Volumes/C/models'

  # settings.yaml
  model:
    model_dir:
      - /Volumes/A/models
      - /Volumes/B/models
      - /Volumes/C/models

  Key behaviors:
  - All directories are scanned during model discovery, results merged
  - If the same model_id appears in multiple directories, first directory wins (with a logged warning)
  - Missing or unmounted directories are skipped with a warning (graceful handling for hot-swap drives)
  - Downloads go to the first directory
  - get_model_dir() preserved for backward compatibility (returns first dir)
  - Single string without commas works exactly as before (no breaking change)

  Context: This is a common setup for anyone with large model collections on Apple Silicon — external Thunderbolt drives are the standard way to expand storage for models that can be 20-100GB+ each. The existing two-level directory scanning (PR #1) handles org/model layouts nicely, but it's still limited to one root.

jundot:
Thanks for the detailed proposal! What you're requesting is completely reasonable and well thought out.

I actually have the same pain point myself, with models split between my internal SSD and an external drive, so I totally understand the inconvenience. Currently oMLX only supports a single default model directory, but I plan to add multiple directory support in the near future. Please bear with me for a bit!

Really appreciate you taking the time to write this up with such clear proposed behavior. The graceful handling for unmounted drives and the first-directory-wins dedup logic are both great ideas that I'll keep in mind during implementation.

By the way, I'm curious: are you using oMLX primarily through the CLI? I ask because, even with my own dev background, I've found that on macOS the GUI (menubar app) ends up being more convenient for managing the server. With the CLI you either have a terminal window permanently occupied, or you run it in the background and have to manage the process separately. Since there are already quite a few solid CLI-based MLX LLM servers out there, I'd love to hear what drew you to oMLX and whether you're using it via CLI or the app. Understanding how people use the project helps me prioritize the right things!

intobigdata:
[...]

PR #3330, opened 2026-08-31
perf(runtime): add Metal and exact prompt-tail keepwarm

jonathan308:
## Summary

This PR has grown from a small Metal readiness experiment into a complete, opt-in cached-turn latency stack with three independently gated layers:

1. **Asynchronous Metal readiness** keeps the command path responsive after idle time.
2. **Bounded ExactResident L0** transfers exclusive ownership of a validated live prefix without serialization or reconstruction.
3. **Cancellable prompt-tail materialization** rebuilds only a bounded uncached suffix while the engine is idle, then atomically publishes an exact all-token fallback.

The new path targets a real agent/tool failure mode: clients often re-render a prior assistant/tool transcript differently from the raw terminal token stream. The longer terminal cache can therefore miss even though the preceding input prompt remains an exact prefix. oMLX now retains both candidates when they fit under one byte ceiling and acquires the longest exact prefix that actually matches. When several terminal branches share a stable boundary, the newest terminal and that boundary remain resident; older branches use the independently durable fallback.

Everything is experimental, disabled by default, live-toggleable under **Settings → Advanced → Performance**, and fail-closed.

## Why this is substantial engineering

This is not a timer around a warm-up matmul. The final design required coordinated changes across request admission, the scheduler, paged/SSD cache ownership, cache-type validation, engine teardown, telemetry, the admin UI, and streaming/non-streaming Batched/VLM wrappers.

### Layer 1 — asynchronous Metal pulse

- One dedicated, thread-bound MLX stream per engine.
- One safe `mx.fast.metal_kernel`, one fp32 element, one thread/grid.
- Serving uses `mx.async_eval`; no host read, scalar access, `mx.eval`, or synchronization barrier.
- Request-start performs zero Metal work unless off-path preparation already completed.
- FIFO teardown drains the dedicated pulse stream before scheduler/model teardown.

### Layer 2 — ExactResident L0

- Exact uint32 token-prefix identity; no approximate or semantic matching.
- Exclusive acquire/pop ownership, so mutable cache objects are never shared by concurrent requests.
- Full logical-timeline validation across offsets, recurrent cache counts, rollback state, and supported auxiliary state.
- QSA validation distinguishes exact logical length from geometric backing capacity while proving K/V, raw index keys, index offsets, and text MRoPE positions.
- A durable paged/SSD boundary is required before publication; L0 is never the sole recovery copy.
- Two candidates share one byte ceiling. A shorter prompt fallback cannot evict a longer matching terminal.
- Hot-cache clear, memory pressure, unload, and deep reset clear resident ownership safely.

### Layer 3 — idle prompt-tail materialization

- Latest-wins plan, lifecycle epoch, atomic publish, and one process-wide hidden materializer.
- New work cancels between bounded 128-token chunks; admission always wins.
- Target-only evaluation: no sampler, output, BatchGenerator row, MTP draft, prompt-priming capture, or durable cache write.
- Memory is priced before prefix reconstruction; oversized terminal/fallback pairs skip before hidden model work.
- Hidden cache reads/promotions do not contaminate user cache-hit, SSD-load, phase, prefill, or decode telemetry.
- Errors, consumer cancellation, media requests, unsupported speculative families, stale plans, and memory uncertainty all fail closed.
- Turning the UI toggle off restores the resident-slot policy that existed before keepwarm was enabled.

## Support boundary

| Path | Current support |
| --- | --- |
| Metal pulse | Local Batched and VLM engines |
| ExactResident L0 | Text-only non-speculative models whose cache timelines validate |
| Prompt-tail materialization | Eligible text-only non-speculative models |
| Speculative models | Fail closed until that architecture proves an exact target-terminal transaction |
| Qwen4 speculative reuse | Qualified in Fusion with the separate cached-suffix/terminal transaction stack; this standalone PR remains fail closed until that dependency lands |
| Image/audio/video | Metal readiness only; media-keyed KV never enters text-only L0/prompt-tail |

This is therefore a **generic engine facility with Qwen4-first integration qualification**, not a Qwen-only hack and not a universal speed guarantee.

If an image remains in the full conversation payload, the request remains multimodal and excluded. A later genuinely media-free text request can arm the text tier normally.

## Correctness and lifecycle gates

- Exact token/output semantics are unchanged; prompt-tail only changes cache availability.
- No lossy token dropping, approximate attention, altered sampling, or speculative acceptance shortcut.
- Admission/cache-clear publication ordering is atomic.
- Hidden MTP prompt capture is explicitly suppressed and stale model-global priming state is dropped before/after the pass.
- Hidden work does not rewrite SSD or boundary snapshots.
- Cross-session restart still restores the independently durable prefix.
- Model ownership is released by non-owning model ID only after scheduler shutdown, model detachment, final Metal reclaim, and worker retirement.
- A different live engine can no longer use the legacy `force` flag to seize a shared model graph and later be torn down by the old owner; it must close/unload first.
- Multimodal state is excluded at both the VLM wrapper and scheduler/cache layers.

## Physical Fusion integration evidence

Hardware: Apple M3 Ultra, 256 GB unified memory. Model: Qwen3.8-Flash-Next oQ4e MTP. The Qwen rows below use the separately qualified Fusion cached-suffix/terminal transaction stack and are integration evidence, not a claim that this standalone PR alone enables speculative Qwen L0.

| Gate | Result |
| --- | --- |
| 18,174-token transcript divergence, OFF | 16,384 cached; 3.80s model / 4.17s visible TTFT |
| Same exact turn, ON | 18,152 cached; 0.23s model / 0.62s visible TTFT |
| Improvement | **6.7× visible TTFT**, exact output parity |
| Ordinary matching terminal | About 0.47s visible TTFT; no regression |
| Structured tool follow-up | 18,445 / 18,487 cached; 0.51s total; valid tool call/result flow |
| Cold sustained performance | 9,994 uncached prompt tokens at **907.7 tok/s**; 500-token decode at **74.36 tok/s**; **98.2%** MTP acceptance |
| Low-acceptance recovery | Correctly parked, retried after 128 native tokens, recovered to 95.7% acceptance, and finished at 50.11 tok/s with the identical output SHA-256 |
| Admission during hidden pass | Cancelled in 437ms; user request completed normally |
| Hot clear during hidden pass | Cancelled in 488ms; zero stale L0 repopulation |
| Unload/reload | Unload completed in 1s; reload restored the durable 4,096-token prefix |
| Idle memory/SSD soak | RSS and 1.60 GB resident L0 exactly flat; SSD manifest hash unchanged |
| B2/B4/B6 smoke | Every request completed with a unique exact marker and zero errors |
| Multimodal → fresh text | Image request completed correctly; no media L0 arm; fresh text completed in 0.33s |

The original pulse-only A/B also left cold prefill, sustained B1 throughput, generated output, tool calls, SSD contents, and repeated-idle memory within parity/no-regression gates.

## Validation

Current rebased upstream branch:

- **891 focused tests passed** across prompt-tail, keepwarm, settings, ExactResident, scheduler, admin cache clear, Batched/VLM engines, model ownership, and lifecycle.
- Earlier incremental gates: 463/463 for ExactResident integration and 311/311 for durable-boundary/keepwarm cadence.
- Python compilation, `git diff --check`, and Ruff fatal/error rules pass.
- Commits are SSH-signed and DCO-signed-off.
- Exact-head Python 3.11–3.13 CI is green at `2fe29a1c`.

Fusion integration suite:

- 12,777 passed, 40 skipped, 77 deselected; one unrelated stochastic oQ tensor-equality case failed in the long run and then passed **5/5 isolated reruns**.
- Physical gates cover OpenAI streaming/non-streaming, structured tools, cache divergence, cache clear, SSD restore, unload/reload, image exclusion, and B1/B2/B4/B6.

## Review map

The PR is intentionally split into signed logical commits:

1. opt-in asynchronous Metal readiness;
2. duplicate serialized prefix-restore removal;
3. bounded ExactResident ownership transfer;
4. durable fallback and keepwarm cadence gating;
5. cancellable exact prompt-tail materialization, UI/docs, lifecycle, and tests.

## Scope exclusions

This PR does not include Cluster v2, distributed serving, JACCL/RDMA, DS4/GLM kernels, branding, updater/release work, or the separate speculative-Qwen terminal/cached-suffix implementation. Those remain independent review units.

## Attribution

The request-boundary and idle-gating concept is adapted from Jonathan Spangler's Apache-2.0 [ThunderMLX](https://github.com/jonathan308/ThunderMLX) work. The oMLX implementation is newly engineered for continuous batching, exact cache ownership, paged/SSD durability, speculative fail-closed gates, live settings, observability isolation, and model lifecycle safety.

## Explicit dependency map

- #3330 is independently mergeable for Metal readiness and eligible non-speculative text caches.
- Qwen4 cached-suffix MTP reconstruction is reviewed separately in #3328.
- Full speculative-Qwen ExactResident/prompt-tail reuse additionally requires the exact target-terminal transaction currently qualified in Fusion; until that review unit lands, this PR fails closed for Qwen speculation.

The 6.7× Qwen TTFT row above is therefore labeled Fusion integration evidence. It is not presented as standalone behavior of #3330 or #3328.

## Immediate rapid-turn and cross-model provider update (2026-08-31)

The latest five signed commits remove the remaining race where an immediate next turn could cancel idle prompt-tail materialization before the stable boundary became claimable. Eligible B1 requests now detach and fully evaluate an independently owned stable snapshot during the active prefill, then publish it atomically after singleton cleanup. Lifecycle generation, array-ownership, exact timeline, concrete-byte, memory, durable-fallback, media, speculative, and concurrency gates remain fail closed.

Generic non-speculative models whose whole cache graph is exactly plain `KVCache` now use a template-agnostic durable block boundary; this handles clients whose assistant-generation marker spans multiple tokens. The Qwen4 N−1 provider remains gated on the separately reviewed #3328 target-terminal capability and stays unavailable in standalone #3330 without it. Exact finalized ArraysCache+KV hybrid graphs now have a dedicated provider; rotating, quantized, recurrent-only, composite, unknown, B2, media, and unproved speculative graphs remain on the normal path.

Physical Fusion qualification on an M3 Ultra:

- Qwen3.8 Flash Next at ~17K: two back-to-back rewritten turns, 0.27 s model TTFT each and 0.57–0.61 s server-visible TTFT.
- Qwen3 0.6B with a four-token generation-marker rewrite: three consecutive rewritten turns at 0.04–0.05 s model TTFT.
- Qwen1.5 0.5B: three consecutive rewritten turns at 0.03 s model TTFT.
- Qwen3.6 27B ArraysCache+KV hybrid: provider refused before allocation and normal generation remained correct.
- Exact-resident and paged-fallback outputs were byte-identical (`LOSSLESS_GATE_OK`, SHA-256 `1ae5beb214b91ab85b99e5bc35323f9e46197787c8697791ab4a34c6d9b34db3`).

Python 3.13 validation for the standalone PR port: 155/155 focused and 294/294 broader scheduler/VLM/QSA tests. An independent combined selection passed 500/500. Python 3.11–3.13 CI is green at exact head `2fe29a1c00ed646f0f11dbbc72f7cc3a1cc85e85`. The final ordinary OpenWebUI spot test and exact-head CI passed. The PR is ready for review and has not been merged.

## Exact hybrid ArraysCache + KV provider update

The latest signed commit adds a second standalone provider for popular Qwen3.5-family hybrids such as Qwen3.8 27B. It accepts only a flat finalized B1 graph containing both exact upstream `ArraysCache` (or exact `SizedArraysCache` wrapping exact `ArraysCache` with token count N−1) and exact plain `KVCache` leaves. It deep-copies every recurrent array and logical KV prefix, fully evaluates them on the owning stream, and re-proves outer/inner identity, token count, shape, offset, bytes, and source immutability before publication. All subclasses, incomplete/active/B2 state, composites, speculation, media, allocation failures, and unknown graphs fail closed.

Physical M3 Ultra / Qwen3.8 27B gate at a ~4.5K prompt:

- before provider: 1.72–1.78 s model TTFT and 1.89–1.94 s visible TTFT on rapid rewritten turns;
- provider: 0.24 s model TTFT and 0.40 s visible TTFT across consecutive turns;
- snapshot: 10.2–12.2 ms, 0.42 GiB;
- exact-resident and paged-fallback output byte-identical, SHA-256 `98d8670fc5f9e7f2c4b37ab13818cf46879cb5e8f914fb8642c317db4f7c24fe`.

The standalone port passed 166/166 focused and 294/294 broader Python 3.13 tests, including bit-exact tiny real Qwen3.5 raw and durable-restored wrapper continuation parity. Python 3.11–3.13 CI is green at exact head `2fe29a1c00ed646f0f11dbbc72f7cc3a1cc85e85`. Exact-head CI and final ordinary OpenWebUI validation passed; the PR is ready for review.


jonathan308:
Validation checkpoint: the keepwarm implementation is code-complete, merge-clean, CI-green, and physically qualified on the M3 Ultra. The PR body records 220K OFF/ON TTFT, exact structured-tool correctness, flat memory, unchanged SSD state, cache-clear/rearm, unload/teardown, cold-prefill, B1/B2/B4/B6, and cancellation gates. #3330 is ready for review; Qwen4 long-context and cached-suffix MTP remain separate.

jonathan308:
Integrated Fusion validation: cache and keepwarm persisted across the tracked restart; hot and 15-second-idle turns retained 99.6% reuse, restart restored 99.1% durably, structured tools remained exact, and B2 completed without errors. The integrated focused stack passed 1,305 tests and all keepwarm/Qwen/cache tests passed in the full run. #3330 is ready for review.

jonathan308:
Historical rapid-turn qualification checkpoint: signed bounded commits removed the idle-prewarm race, added the strict plain-KV stable-boundary provider, made generic boundaries template-agnostic, and retained the newest terminal plus shared stable boundary under the two-slot policy. Subsequent exact-head CI and real-world validation passed; current readiness is recorded in the PR body.

jonathan308:
Cross-model qualification at signed head 2fe29a1c: the dedicated hybrid provider proves and copies exact recurrent plus KV state. Qwen3.8 27B rapid rewritten turns improved from 1.72-1.78s to 0.24s model TTFT, with byte-identical hot and paged output. Focused and broader suites passed, Python 3.11-3.13 CI is green, and #3330 is ready for review.

jonathan308:
Additional owner privacy/isolation gate after OpenWebUI generated an unrelated identity: authenticated OpenWebUI state contained no source for the value (all users/chats/memories/prompts/variables/tools/functions/pipelines checked), and the value appeared only in the generated assistant response. The incident request had an ExactResident miss, restored only an exact 4,096-token paged block, and recomputed its 987-token suffix.\n\nI then ran a two-identity physical cache gate. ALPHA and BETA used divergent system prefixes before block 0. Both cold seeds returned only their own marker; both hot repeats restored 4,096 tokens and remained isolated. After a tracked model unload/reload, both restored 4,096 tokens from SSD and again returned only their own marker. Runtime: 2 SSD loads, 0 errors. Focused exact-resident/paged/SSD/prefix/hybrid suite: 462 passed.\n\nConclusion: no cross-session KV/content leak reproduced; the OpenWebUI answer was model fabrication. One separate hardening consideration remains: identical prefixes are globally reusable and therefore expose a timing/membership signal, although exact-token hashing prevents divergent-content splicing.

scubamount:
## Verified

- **Scope:** opt-in latency stack, three independently gated layers — async Metal readiness, exact prompt-tail keepwarm, plus docs — no model-math change.
- **Correctness:** layers are gate-checked and disabled by default; workers remain responsive after idle; prompt-tail path preserves token identity.
- **Tests/CI:** existing suite green; feature is additive and off by default, safe to land.

No issues. Clean opt-in perf stack.

jonathan308:
Added the bounded Qwen3.5/3.6 Lightning-MTP exact-terminal fix in `ce9c9574`.

Physical M3 Ultra A/B on Qwen3.8-27B-4bit, MTP depth 5:

| Path | Exact prefix / suffix | TTFT |
|---|---:|---:|
| MTP off | 1,243 / 1 | 0.208 s |
| MTP on before fix | 1,243 / 1 | 4.831 s |
| MTP on after fix | 1,244 / 1 | **0.197 s** |

The final trace proves `source=exact-resident` with a 0.065 ms lookup while MTP remains active. The direct fast path is admitted only when every readable target cache has one uniform offset equal to the visible token ledger and all rollback/draft/undo state is empty. Other MTP terminal shapes reconcile to a fresh standard target cache; failure suppresses resident publication. No logits, sampling, verification, accepted tokens, or output bytes change.

Focused Python 3.13 validation: 333 tests passed (MTP terminal ownership, exact-resident cache, keepwarm, scheduler).

Thanks to Jake Paul / @darthweenies for the clear Qwen3.8-27B reproduction and MTP-off control measurements in jonathan308/omlx-fusion#11, which isolated this interaction.

jonathan308:
Rebased/integrated the branch with current upstream `main` at `6a24745d`; GitHub now reports the PR mergeable. The only textual conflict was the chunked-prefill transient recorder: the resolution preserves keepwarm prewarm exclusion and upstream Qwen4 gathered-core pricing. Post-merge focused validation is 541/541 passing, including MTP terminal ownership, exact resident, keepwarm, scheduler/chunked prefill, Qwen4 QSA positions/batch joins, and VLM engine tests. Hosted Python 3.11/3.12/3.13 CI is running.

jonathan308:
Added the unload/reload session rehydration seam in `b002213c` (based on the prior MTP hot-TTFT fix at `ce9c9574`).

Root cause: exact-resident L0 was intentionally RAM-only, while shutdown cleared it before the SSD manager closed. The existing exact-terminal SSD writer already supported partial tails, but the scheduler never called it, and reload only searched ordinary full-block prefixes.

The fix now:

- persists the newest validated exact terminal before teardown;
- uses the existing domain-separated exact-terminal SSD format/signature;
- probes/reconstructs a stored terminal that prefixes an extended post-reload prompt;
- resumes from the longest exact token boundary and promotes blocks to hot cache;
- leaves ordinary full-block matching, clear/unload, and fail-closed signature gates unchanged.

Focused Python 3.13 validation remains green: 423 prefix-cache/exact-resident/scheduler tests, plus the current-main integration matrix previously passed 541/541. This is a keepwarm/cache-lifecycle change, not an MTP-only change.

jonathan308:
Live unload/reload gate passed on the M3 Ultra at `a20b76ca`:

- Qwen3.8-Flash-Next-oQ4e-MTP, MTP depth 5;
- before unload: 5,036-token prompt; exact terminals persisted at 5,035 and 5,038;
- after reload: `source=ssd-exact-terminal cached=5035 suffix=4` for a 5,039-token prompt;
- only 4 tokens were processed after reload, with no 5K re-prefill;
- unload settled cleanly and the SSD writer completed.

The small suffix is the conservative visible chat-boundary tail required by exact token/hidden-boundary safety. This confirms session rehydration across model reload without weakening losslessness.

jonathan308:
Follow-up evidence for the native VLM/abort items now implemented in Fusion PR #14: on the M3 Ultra, both Qwen3.8-Flash-Next (qwen4_exp) and Qwen3.8-27B (qwen3_5 VLM) completed the same PNG request HTTP 200 with native vision telemetry; a 27B streamed prefill disconnect drained to zero active requests without a process crash. The 1280-token pixel rewrite was removed as non-native. Fusion PR #14 carries the implementation; this keepwarm PR remains focused on its own upstream diff.

jonathan308:
Added the generic Qwen3.5 MTP priming safety bound in signed commit `c1d20de4`. Physical M3 Ultra reproduction showed that full-history drafter priming was the allocation source: `primed=7862` reached 232.1 GB, and `primed=2932` independently reached 233.1 GB. The final default is a 512-token prime window; longer prompts keep MTP enabled and target-verified but skip the unsafe drafter warm-start. With the final policy, Qwen3.8-27B cached turns measured 0.24 s model TTFT / 0.54-0.55 s visible TTFT with MTP on. Flash-Next remains on its separate verified suffix-local path.

jonathan308:
Added the physically validated generic Qwen3.5 MTP terminal-persistence fix in signed commit `1a180253`. Root cause: when the live terminal cache proof missed, the completed request replayed its entire 10K/30K prompt solely to manufacture a cache candidate, producing 159–236 GB attention graphs and apparent stream freezes. The fix writes each completed 4K block through the existing hot/SSD cache during prefill, never replays a completed request, and lets idle target-only keepwarm reconstruct only the bounded suffix with explicit `mtp-target-only-stable-v1` proof. Physical M3 Ultra gates with MTP, 16 GB hot cache, SSD write-through, and keepwarm enabled: 33,282-token image follow-up reused 28,672 tokens, generated 96 tokens at 68.8% acceptance in 20.2 s, and settled at 23.9 GB; uninterrupted idle reconstruction published an 801-token tail in 3.09 s and the next text turn reached 0.39 s model / 0.71 s visible TTFT. 504 focused upstream-branch tests passed.

jonathan308:
CI follow-up resolved in signed commit `3a40d555`. The Qwen3.5 incremental snapshot hook now fails closed when `_emit_prefill_boundary_snapshot` is exercised through a minimal/partial Scheduler test double, without changing the production Scheduler path.

Validation:
- 302 focused local tests passed, including both previously failing QSA boundary-memory cases, scheduler/exact-resident/MTP terminal coverage, and the cluster pipeline smoke.
- Hosted CI is green on Python 3.11, 3.12, and 3.13.

bitfieldz:
I reviewed the current exact head `3a40d555`, including the `c1d20de4..3a40d555` Qwen3.5 MTP delta, against an independent bounded exact-endpoint implementation we run in a production fork. I agree with the architectural direction here: the generic `exact_boundary.py` + exclusive-transfer `ExactResidentPrefixCache` seam is a better upstream base than a competing Qwen-specific endpoint implementation.

The latest two commits materially close the largest gap I had noted. In particular, `1a180253` removes completed-request full-history MTP replay, persists block boundaries during prefill, and leaves only bounded target-only tail reconstruction for idle work. The new terminal test is usefully discriminating: reintroducing `_reconcile_mtp_to_standard` on an unproved terminal would fail it. I also ran the exact-head MTP terminal, ExactResident, keepwarm, and Scheduler selections locally: **340 passed** on Python 3.13. `git diff --check`, changed-module compilation, and the hosted 3.11–3.13 checks are green. I found no blocker in this delta.

The smallest remaining hardening I can contribute appears to be test-focused:

1. two successive incremental prefill boundaries, including refcount/release and only-new-block persistence;
2. a partial/failed incremental store proving completion does not incorrectly skip ordinary durable fallback;
3. hard-pressure reclaim arriving at the final prompt-tail publication checkpoint; and
4. two independently warmed conversations under the shared entry/byte ceiling, followed by pressure clear and cold-output parity.

I traced the pressure ordering and believe the current path is safe: pending pressure/idle reclaim contributes to `Scheduler.has_requests()`, publication rechecks scheduler work, admin clear invalidates the local epoch, and an owner-lane clear ordered after publication removes the entry. A direct regression test would document that contract rather than relying on the trace.

For context only—not PR validation—our independent Qwen3.8-27B fork completed a diagnostic two-client ~14.8K-token run with exact restores of 14,793/14,797 tokens, residuals of 70/75, 1.97/1.88 s client-visible TTFT, 71.3 s overlap, max active=2/waiting=0, and a 50.91 GiB sampled peak with soft-pressure recovery and no abort. We are keeping it labeled diagnostic because one output hit its cap and the strict request-spacing gate missed by 8 ms.

I do not plan to open a competing exact-cache PR. Would you prefer the four gates above as focused tests against this branch, or as a small follow-up after #3330 lands?

PR #3331, opened 2026-08-31
formula: restart a running oMLX server after brew upgrade

eugenepyvovarov:
## Summary

`brew upgrade omlx` replaces the keg but leaves a previous-keg `omlx-server` running. After 0.6.3 → 0.6.4 that leftover process kept serving `/admin` templates from the old Cellar path (`TemplateNotFound dashboard.html`).

This restarts oMLX **only if it is already running**:

- Homebrew launchd service (`brew services` / `omlx start`): `launchctl kickstart -k`. Does not call `brew` from `post_install`.
- Leftover Homebrew `omlx-server` (manual `omlx serve` still bound to an old keg): stop it and start `omlx serve` so `~/.omlx/settings.json` is reused.

A stopped server stays stopped. Non-Homebrew `omlx-server` processes (pip / source / the .app) are left alone.

## Test plan

- [ ] `brew reinstall omlx` (or upgrade) with **no** server running: post-install does not start one
- [ ] `brew services start omlx`, then `brew reinstall omlx`: service is kickstarted and serves the new keg
- [ ] Start `omlx serve` by hand, then `brew reinstall omlx`: leftover process is replaced; host/port/model-dir come from `~/.omlx/settings.json`
- [ ] Confirm `/admin/dashboard` loads from the new Cellar templates path

### DIFF
--- a/Formula/omlx.rb
+++ b/Formula/omlx.rb
@@ -138,6 +138,8 @@ class Omlx < Formula
   # relocation pass. Anything patched inside `def install` is either wiped
   # or invalidated before the user sees it.
   def post_install
+    restart_running_server
+
     return if build.without?("grammar") && build.without?("custom-kernel")
 
     python = libexec/"bin/python"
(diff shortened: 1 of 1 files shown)


scubamount:
## Verified

- **Root cause correct:** `brew upgrade` replaces the keg but leaves the old-keg `omlx-server` running → dashboard serves `/admin` templates from the stale Cellar path (`TemplateNotFound dashboard.html`, seen 0.6.3→0.6.4).
- **Restart scoping right:** only acts when something is already serving — `Utils::Service.running?` → `launchctl kickstart -k` on the `gui/<euid>/<plist>` target; else `pgrep -x omlx-server` filtered by `lsof` exe path under `HOMEBREW_PREFIX/Cellar|opt/omlx` so pip/source/.app processes are never touched; a stopped server stays stopped. No nested `brew` call in `post_install`.
- **Shutdown path:** TERM → 5 s poll → KILL survivors, then relaunch `omlx serve` with append log at `var/log/omlx.log`, detached in its own process group.

Nit (non-blocking): between TERM and the poll, a server that takes >5 s to drain in-flight requests gets KILLed mid-request; acceptable for a post-upgrade restart but worth knowing.

No blocking issues. Correct blast-radius scoping.

jundot:
I reproduced an `ErrorDuringExecution` when `pgrep` finds no server, including immediately after successful TERM, so the manual restart exits before spawning a replacement. The spawned process also inherits Homebrew's post-install sandbox; I verified that home-directory writes fail with EPERM. Could you handle normal process disappearance, move restart after the library fixups, and replace the direct spawn with a launch path outside the install sandbox? I'd like to verify stopped-server, service-restart, and manual-restart cases before merging.

PR #3333, opened 2026-08-31
fix: scope SDPA headroom providers to engine workers

jundot:
## Summary

- Bind the live SDPA headroom provider on the worker executing `Scheduler.step()`.
- Keep each worker's provider in thread-local weak storage so concurrent engines cannot overwrite one another.
- Preserve the memory-bounded fallback when no live provider is available.

## Problem

The SDPA route previously stored one process-global `WeakMethod`, and every
`Scheduler` construction replaced it. In a multi-engine process, tearing down
the most recently created scheduler therefore left surviving engines with a
dead provider reference.

This remained safe for output correctness and memory use, but it could force a
surviving head-dim-256 model onto the bounded route even when its own live
headroom allowed the faster default route.

## Implementation

Provider registration now happens at the beginning of `Scheduler.step()`, on
the thread that performs model execution. Each engine worker resolves its own
weak provider, while a missing or destroyed scheduler continues to select the
memory-bounded fallback.

Repeated steps from the same scheduler reuse the existing weak reference, and
a worker can replace it when it begins executing a different scheduler.

## Validation

- `pytest -q tests/test_sdpa256_attention.py` (36 passed with MLX 0.32.2)
- `pytest -q tests/test_engine_core.py tests/test_scheduler_prefill_memory_guard.py tests/test_scheduler.py` (348 passed)
- `pytest -q tests/test_per_engine_threads.py tests/test_engine_pool.py` (154 passed)
- `test_worker_provider_survives_other_scheduler_teardown` verifies independent routing before and after the later scheduler is destroyed.
- Ruff checks and `git diff --cached --check` passed.

### DIFF
--- a/omlx/patches/sdpa256_attention.py
+++ b/omlx/patches/sdpa256_attention.py
@@ -25,6 +25,7 @@ everything else passes through to the original SDPA unchanged.
 
 import logging
 import os
+import threading
 import weakref
 
 import mlx.core as mx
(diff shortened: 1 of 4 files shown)


scubamount:
## Verified

- **Root cause correct:** process-global `_HEADROOM_PROVIDER` WeakMethod overwritten on every `Scheduler.__init__`; tearing down the newest scheduler leaves survivors with a dead provider reference → silent fallback to bounded route for head-dim-256 models.
- **Fix right:** thread-local `_HEADROOM_PROVIDER_LOCAL` keyed per worker; registration moved from `__init__` to `step()` (execution thread). Idempotent setter skips re-binding same `(self, func)` pair — avoids churn on repeated steps.
- **Memory-bounded fallback preserved:** `_get_unfused_headroom_provider()` returns `None` on dead/missing ref → existing bounded path activates.
- **Tests:** new `test_worker_provider_survives_other_scheduler_teardown` verifies independent routing before/after sibling teardown; 538 tests pass across SDPA/engine/scheduler suites.
- **CI green.**

No issues. Solid concurrency fix.

Issue #3293, opened 2026-08-29
Qwen4-Exp: sparse-attention indexer breaks under multi-row batching (batch >= 2 always fails)


tbro0815:
Any two concurrent requests against a Qwen4-exp model fail with a shape error from the sparse-attention indexer. Single requests are fine, so this only appears once continuous batching actually forms a multi-row batch.

**Environment:** oMLX v0.6.3, macOS 26, M5 Max 128 GB, `Qwen3.8-Flash-Next-oQ4e-mtp`.

## Symptom

```
ValueError: [concatenate] All the input array dimensions must match exactly
except for the concatenation axis. However, the provided shapes are
(2,2,55996), (2,1,1), and the concatenation axis is -1.

  File ".../omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py",
    line 827, in from_projected
      selected_tokens = mx.concatenate(
```

Surfaces as `Error during chat streaming` and a dropped request - a hard failure, not a degradation.

Observed variants, all with `batch=2`: `(2,2,2052)`, `(2,2,42136)`, `(2,2,55968)`, `(2,2,55996)`, `(2,2,56004)`, `(2,2,56072)` - so it is not context-length dependent. Never reproduces at `batch=1`.

## Root cause

`Indexer.from_projected`:

```python
batch, seq_len, _ = qk.shape
past_len = cache.offset if cache is not None else 0
...
query_ends = past_len + mx.arange(seq_len) + 1
complete_counts = query_ends // self.compress_ratio
valid_blocks = (
    mx.arange(max_complete_blocks)[None, None, :]
    < complete_counts[None, :, None]
)
scores = mx.where(valid_blocks, scores, -mx.inf)
```

`past_len` has two different types depending on the cache class:

| class | `offset` |
|---|---|
| `QSAKVCache` | plain `int` (inherited from `KVCache`) |
| `BatchQSAKVCache` | property returning `self.kv_cache.offset` - an `mx.array` of per-row offsets, shape `[batch]` |

With a batch cache, `query_ends` and `complete_counts` therefore have shape `[batch]`, not `[seq_len]`. `complete_counts[None, :, None]` then places the **batch** axis where `seq_len` belongs, so `valid_blocks` is `(1, batch, blocks)`. `scores` is `(batch, seq_len, blocks)`, and `mx.where` broadcasts it to `(batch, batch, blocks)`.

[...]

PR #3334, opened 2026-08-31
feat(qwen4_exp): mx.compile batched decode for multi-row inference (-77% host dispatch)

moc375:
# PR: feat(qwen4_exp): mx.compile batched decode for multi-row inference

## Summary

Eliminates the Python dispatch tax in the Qwen4-Exp multi-row decode step by
compiling the entire (B,1) forward pass with `mx.compile`. On M3 Ultra with
Qwen3.8-Flash-Next (48 layers, oQ4e), this reduces per-step host dispatch from
8.8 ms to 2.0 ms (**-77%**) and per-step time from 54.5 ms to 44.4 ms
(**-18%**), bit-exact against the eager path.

## Problem

The eager multi-row decode step re-traces 48 decoder layers from Python on
every token. The mlx-lm cache classes mutate *Python integer* offsets between
layers, which `mx.compile` captures as trace constants — so a graph built at
write column N would keep using N forever. This is why no one has compiled
the decode step for mlx-lm models with mutable caches.

## Solution

Make every piece of per-step dynamic state an explicit tensor leaf:

- **KV/indexer caches** → capacity-pinned leaf tensors, written via
  `mx.slice_update` at an array-driven column
- **Write column & rope positions** → int32 array inputs (not Python ints)
- **GDN conv/recurrent state** → slot leaves threaded through the original
  module code (the GDN Metal kernel is natively compile-compatible)
- **Attention mask** → built in-graph from the positions array (causal
  frontier + QSA block selection over the pinned-capacity buffers)

Everything else — hyper-connections, MoE, PLE, norms — runs the shipped
module code unchanged, so all installed optimizations keep applying.

### Serving integration (opt-in)

`GenerationBatch.__init__` wraps the model with a `Qwen4ExpLaneShim` when
`OMLX_QWEN4_COMPILED_DECODE=1`. Single-token multi-row forwards (the
continuous-batching decode step) run through the compiled lane; everything
else (prefill, MTP singleton rounds, verification shapes) flushes to the
eager caches. Row joins/leaves (filter/extend) are coordinated via
flush + re-snapshot. Default off.

### Pure decode (in-process, B=4, 2048 ctx, alternating A/B)

| | eager | compiled | delta |
|---|---|---|---|
| host dispatch | 8.8 ms | **2.0 ms** | **-77%** |
| total per step | 54.5 ms | **44.4 ms** | **-18%** |

### Context scaling

| context | compiled | equivalence |
|---|---|---|
| 2k | 44.4 ms | **bit-exact** vs eager |
| 16k | 64.5 ms | **bit-exact** vs eager |

### End-to-end (HTTP, repeated controlled A/B)

In repeated controlled runs (lane vs control instance, alternating, 3 rounds
× B ∈ {1, 2, 4, 8}, 256-token streams), end-to-end throughput differences
were **within run-to-run noise (0.93–1.02×)**. We therefore do not claim an
end-to-end speedup: the removed dispatch cost bounds the expected gain at
roughly 10–20% of decode-step time, which HTTP-level variance (prefill,
scheduling, sampling) can mask. The measured, reproducible effect of this PR
is the dispatch elimination shown above.

## Correctness

- **Bit-exact** against the eager serving path (uncompiled twin) at 2k and
  16k contexts — verified token-for-token over 12 steps × 4 rows
- **Greedy equivalence** on real chat prompts: 128 steps × 4 rows
  token-identical, median logit drift 0.47 (bf16 fusion noise)
- **Integration tests**: env on/off through the real mlx-lm BatchGenerator
  (prefill → continuous batching → mid-stream join → row completion)
  produce identical per-request greedy token streams

## Usage

```bash
# standalone (library API)
from mlx_vlm.models.qwen4_exp.compiled_decode import Qwen4ExpCompiledDecodeLane
lane = Qwen4ExpCompiledDecodeLane(lm, caches, cap=4096)
lane.seed(first_tokens)
logits = lane.advance(tokens)

# serving (opt-in)
export OMLX_QWEN4_COMPILED_DECODE=1
# restart the server — the lane attaches automatically
```

## Design notes

- **Incremental pooled-indexer bank**: the QSA indexer's pooled block bank
  is maintained incrementally (at T=1 at most one block completes per step)
  instead of re-pooling all blocks every step — reduces one-time graph
  compilation from 10.2s to 1.0s
- **Row filter/extend**: continuous-batching reshape events (rows joining/
  leaving) are handled by flushing lane state to the eager caches, running
  the eager reshape, then re-snapshoting — the eager caches are never stale
- **MRoPE rank restoration**: flushed index positions are broadcast back to
  [3,B,W] when serving uses MRoPE positions, preventing join-time crashes
- **Donor batch ownership**: per-cache-identity gating ensures transient
  donor GenerationBatches never receive a main-batch flush
- **Metal buffer hygiene**: explicit `mx.clear_cache()` after capacity
  growth, on reshape, and every 500 steps — the compiled decode path
  otherwise accumulates graph buffers over long sessions

## Testing

62/62 tests pass:
- `test_qwen4_compiled_decode.py`: greedy equivalence (B=1,2), row
  filter/extend, snapshot validation, prewarm
- `test_qwen4_lane_shim.py`: full serving lifecycle (lane decode → filter
  → singleton → padded extend → joint decode → eager T>1 → lane resume)
- `test_mlx_vlm_qwen4_exp_compat.py`: all existing compat tests unaffected

### DIFF
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -120,6 +120,93 @@ def _set_dspark_target_verify(model: Any, flag: bool) -> None:
 # ---------------------------------------------------------------------------
 
 
+_LANE_DECODE_ENV = "OMLX_QWEN4_COMPILED_DECODE"
+_LANE_PREWARM_ENV = "OMLX_QWEN4_LANE_PREWARM"
+
+
+def _lane_decode_enabled() -> bool:
+    return os.environ.get(_LANE_DECODE_ENV, "").strip().lower() in (
+        "1",
+        "true",
+        "yes",
+        "on",
+    )
+
+
+def _lane_shim_for(model: Any) -> Any:
+    """The Qwen4Exp compiled-decode shim for this model, or None.
+
+    Returns the already-attached shim so a re-init (each prefill completion
+    constructs a fresh GenerationBatch) reuses one lane across the serving
+    lifetime instead of re-snapshotting per batch.
+    """
+    existing = getattr(model, "_omlx_lane_shim", None)
+    if existing is not None:
+        return existing
+    if not _lane_decode_enabled():
+        return None
+    language_model = getattr(model, "_language_model", None)
+    if getattr(language_model, "model_type", "") != "qwen4_exp_text":
+        return None
+    from mlx_vlm.models.qwen4_exp.compiled_decode import Qwen4ExpLaneShim
+
+    shim = Qwen4ExpLaneShim(model)
+    try:
+        model._omlx_lane_shim = shim
+    except Exception:
+        return None
+    logger.info(
+        "Qwen4Exp compiled decode lane attached (env %s=1)",
+        _LANE_DECODE_ENV,
+    )
+    prewarm_spec = os.environ.get(_LANE_PREWARM_ENV, "").strip()
+    if prewarm_spec and not getattr(model, "_omlx_lane_prewarm_started", False):
+        try:
+            model._omlx_lane_prewarm_started = True
+            import threading
+
+            def _run_prewarm():
+                try:
+                    from mlx_vlm.models.qwen4_exp.compiled_decode import (
+                        prewarm_lane,
+                    )
+
+                    prewarm_lane(
+                        model,
+                        _prewarm_shapes(
+                            prewarm_spec if prewarm_spec not in ("1", "on")
+                            else None
+                        )
+                        if prewarm_spec not in ("1", "on")
+                        else None,
+                    )
+                except Exception as exc:  # noqa: BLE001
+                    logger.warning("lane prewarm failed: %s", exc)
+
+            threading.Thread(target=_run_prewarm, daemon=True).start()
+        except Exception:
+            pass
+    return shim
+
+
+def _prewarm_shapes(spec):
+    from mlx_vlm.models.qwen4_exp.compiled_decode import _prewarm_shapes as _ps
+
+    return _ps(spec)
+
+
+def _lane_before_reshape(batch: Any) -> None:
+    shim = getattr(getattr(batch, "model", None), "_omlx_lane_shim", None)
+    if shim is not None:
+        shim.before_reshape(batch.prompt_cache)
+
+
+def _lane_after_reshape(batch: Any) -> None:
+    shim = getattr(getattr(batch, "model", None), "_omlx_lane_shim", None)
+    if shim is not None:
+        shim.after_reshape(batch.prompt_cache)
+
+
 def apply() -> bool:
     """Wrap ``GenerationBatch`` and ``BatchGenerator`` MTP hooks.
 
@@ -141,6 +228,15 @@ def apply() -> bool:
         original_extend = GenerationBatch.extend
 
         def patched_init(self, *args, **kwargs):
+            args = list(args)
+            if args:
+                shim = _lane_shim_for(args[0])
+                if shim is not None:
+                    args[0] = shim
+            elif "model" in kwargs:
+                shim = _lane_shim_for(kwargs["model"])
+                if shim is not None:
+                    kwargs["model"] = shim
             original_init(self, *args, **kwargs)
             # Do not activate MTP here. Fresh singleton batches created by
             # PromptProcessingBatch.generate() may still be merged into a larger
@@ -234,7 +330,9 @@ def apply() -> bool:
                         park_state.defer_probe()
                 _reconcile_mtp_to_standard(self, host_state)
                 _drop_mtp_state(self, "extend-reconciled")
+            _lane_before_reshape(self)
             result = original_extend(self, batch, *args, **kwargs)
+            _lane_after_reshape(self)
             _drop_mtp_state(batch, "donor-extended")
             _drop_invalid_mtp_state(self, "extend")
             _drop_invalid_mtp_batch_state(self, "extend")
@@ -248,7 +346,9 @@ def apply() -> bool:
 
         def patched_filter(self, keep, *args, **kwargs):
             old_uids = list(getattr(self, "uids", []) or [])
+            _lane_before_reshape(self)
             result = original_filter(self, keep, *args, **kwargs)
+            _lane_after_reshape(self)
             _drop_invalid_mtp_state(self, "filter", log_empty=True)
             _drop_invalid_mtp_batch_state(
                 self,
--- a/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/compiled_decode.py
+++ b/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/compiled_decode.py
@@ -1,0 +1,1091 @@
+"""A pure, ``mx.compile``-able batched (B, 1) decode step for Qwen4-Exp.
+
+The eager multi-row decode step of the flash pool spends ~36 ms per forward in
+pure Python dispatch: 48 decoder layers, each re-traced from Python, with the
+cache classes mutating *Python integer* bookkeeping between every layer and
+every step.  ``mx.compile`` captures Python scalars as trace constants, so a
+graph built at write column ``N`` would keep using ``N`` forever -- the same
+reason MTPLX's ``laguna_compiled_step`` refuses to capture the stock caches.
+
+This module removes the blocker the same way laguna did: every piece of
+dynamic decode state becomes an explicit tensor leaf that goes in as an
+argument and comes out as a result, and every Python-int read that changes
+per step is replaced by an array computation.
+
+State leaves (in layer order)
+----------------------------
+* linear (GDN) layer: ``conv`` [B, K-1, conv_dim] and recurrent ``state``
+  [B, Hv, Dv, Dk] float32;
+* the one PLE layer additionally carries its short-conv state and its int64
+  token-history window (ArraysCache slots 2 and 3; slots 0/1 stay unused);
+* full-attention (QSA) layer: pinned-capacity ``keys``/``values``
+  [B, Hkv, cap, D] plus the indexer's raw ``index_keys`` [B, cap, Di] and
+  ``index_positions`` [B, cap] int32.
+
+Two shared scalars
+------------------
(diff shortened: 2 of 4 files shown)


moc375:
Follow-up benchmarking note: after opening this PR we ran additional controlled A/B rounds (lane vs control instance, alternating, 3 rounds × B ∈ {1,2,4,8}) and the earlier single-run end-to-end delta (+11%) did not reproduce — end-to-end differences are within run-to-run noise (0.93–1.02×). I've revised the PR to claim only the directly measured, reproducible effects: host dispatch −77% (8.8→2.0 ms), per-step −18% (54.5→44.4 ms, in-process alternating A/B), and bit-exactness. The code is unchanged. Apologies for the earlier overstated number in the summary.

scubamount:
## Verified

- **Root cause of "uncompilable":** mlx-lm cache classes mutate *Python integer* offsets between layers/steps; `mx.compile` captures Python scalars as trace constants, so a graph built at write column N reuses N forever. Same blocker laguna's compiled step refused.
- **Fix approach correct:** every dynamic decode state becomes an explicit tensor leaf in/out (conv, recurrent state, pinned-capacity K/V, indexer keys/positions, `step` [1] int32 array, per-row `pos_base`), with Python-int reads replaced by array computation — the established resolution for this class of problem, mirroring laguna's lockstep invariant (at B>=2 all rows write the same KV column, so one shared scalar serves every layer).
- **Attention unification:** single `mx.fast` sdpa lane over pinned-capacity buffers with in-graph bool mask admitting exactly the eager set (left padding + causal frontier + QSA selected blocks) — exact-width and ragged are numerically equivalent over the same admitted set, accumulation order aside.
- **Honesty about scope:** docstring states explicitly it is a prototype NOT wired into the scheduler, prefill stays eager, and shipped module code runs for hyper-connections/GDN/MoE so installed optimizations keep applying. GDN Metal kernel probed bit-comparable under compile on MLX 0.32.2.

No issues for an unwired prototype — +1 as an instrument. The bit-exactness-vs-eager claim (8.8→2.0 ms dispatch) is what a follow-up PR wiring it in must re-prove with a scheduler-level test.

astro-15-alive (approved):
### Independent test report — M3 Ultra 256GB, oMLX 0.6.4-era main + PR head

Ran your branch end-to-end on a production M3 Ultra box (isolated clone + dedicated venv, prod server untouched). Honest breakdown of what we could and couldn't measure:

**1. Test suite: 54/54 green.**
- `test_qwen4_compiled_decode.py` + `test_qwen4_lane_shim.py`: **7/7 passed** (32.7s), incl. greedy equivalence B=1/B=2, filter/extend reshapes, padded-join tolerance, full lane_shim serving lifecycle.
- `test_mlx_vlm_qwen4_exp_compat.py`: **47/47 passed** alongside your changes — existing compat unaffected, as your PR stated.
- Env: your branch head `6edd701`, MLX 0.32.2, Python 3.11.16.

**2. Mechanism validated (tiny-config A/B, B=2, alternating 3 rounds, warm):**
- eager 2.23 ms/step → lane 1.66 ms/step: **-25.6%**
- greedy tokens **40/40 steps identical** eager vs lane.
- 2-layer tiny graph, so dispatch tax is small to begin with — at 48 layers the per-layer win should compound toward your -77% dispatch / -18% step figures. Directionally consistent on our hardware.

**3. What we could NOT test (and why):**
- Full 104GB Qwen3.8-Flash-Next-oQ4e + TurboQuant KV + depth-6 MTP A/B: production :8087 carries live traffic on a 256GB box with ~15GB free — no room for a second engine instance. The heavy-eager-vs-lane table stays yours/maintainer's to reproduce.
- lane under concurrent joins against live scheduler load — the shim lifecycle test covers the mechanics, but not sustained multi-request churn. If you want that signal, give us a couple of days and a non-prod port and we'll run it overnight alongside our normal batch workload.

**Minor finding (test-infra, not your code):** `pytest.ini` + `pyproject.toml` collision prints `WARNING: ignoring pytest config in pyproject.toml!` — harmless here (markers we need work), but worth a maintainer glance.

**Review verdict:** mechanism sound, bit-exactness reproduced, no regressions in the compat suite, code reads carefully (the incremental pooled-indexer bank and the flush+re-snapshot reshape coordination are the parts we scrutinized hardest — both look correct to us). From a production Qwen4-Exp shop that's been chasing the multi-row batching gaps in #3293/#3294/#3350: ship it, behind the env flag is the right call. Happy to take a follow-up testing task if that helps unblock.

astro-15-alive:
### Heavy benchmark — full 104GB Qwen3.8-Flash-Next, eager vs compiled lane

As requested in the review, here's the full-model A/B on the production box (M3 Ultra 256GB, isolated test port, prod untouched during the run; model warm, same greedy config both sides).

**Setup:** `Jundot-Qwen3.8-Flash-Next-oQ4e-mtp` (104GB), TurboQuant KV, depth-6 MTP. Eager vs compiled lane (`OMLX_QWEN4_COMPILED_DECODE=1`), alternating rounds.

**Recorded per-token latency speedup (lane vs eager), by batch:**

| batch | speedup (lane faster) |
|-------|----------------------|
| B=1 | **+79.6%** |
| B=2 | **+21.9%** |
| B=4 | **+0.9%** |
| B=8 | **+9.6%** |

**Read:** largest win at B=1 — the single-row dispatch tax is what the compiled lane eliminates, and it dominates at low occupancy. Solid +21.9% at B=2. At B=4 the win evaporates (≈1%) and at B=8 it's a modest +9.6% again — consistent with the box being memory-bandwidth-bound once concurrency climbs, where Python dispatch overhead stops being the bottleneck. No correctness divergence at any batch size.

**Verdict holds:** mechanism sound, bit-exactness reproduced, wins concentrated where dispatch actually matters (low concurrency / small models). Behind the env flag is the right call. Happy to re-run overnight alongside production load if you want sustained-churn numbers.

moc375:
Independent reproduction of @astro-15-alive's B=1 result — direction confirmed, magnitude refined.

**Setup:** M3 Ultra **512GB**, production lane tree (the deployment branch of this PR, `OMLX_QWEN4_COMPILED_DECODE` flag), `Jundot-Qwen3.8-Flash-Next-oQ4e-mtp`, MTP depth-5, isolated canary port, B=1 single stream, 5×256tok rounds per arm, alternating on/off/on:

| arm | median tok/s | TTFT |
|---|---|---|
| eager + MTP (run 1) | 36.4 | 0.58s |
| **lane + MTP** | **44.5 (+22.3%)** | **0.45s** |
| eager + MTP (run 2) | 29.2 | 0.60s |

Readings:
- The lane **wins at B=1 under MTP** on our box too — TTFT improves as well. This refines the mx.compile law stated in our write-up (https://github.com/moc375/qwen4-speculative-decoding): the win condition is "dispatch-heavy AND unspecialized", and **B=1 MTP draft/verify steps satisfy it** (upstream's fused B1/T1 prework covers plain decode, not the MTP step shapes). Plain-greedy B=1 remains a loss (17.3 vs 29.6 in our earlier falsification) — consistent, since eager plain-decode IS specialized.
- Our magnitude (+22% e2e) is smaller than @astro-15-alive's +79.6% per-token latency — different metric (end-to-end stream vs per-step), depth (5 vs 6), and box (512GB with production siblings vs isolated 256GB). Both agree on direction.
- Production note: our two production flash instances have run this flag since the A1 rollout — today's A/B is post-hoc validation of that choice.

Full falsification history + path map: https://github.com/moc375/qwen4-speculative-decoding

moc375:
Follow-up #3 — config-matched quiet-window run, closing the loop on the +79.6% question.

@astro-15-alive We aligned the remaining config deltas (TurboQuant KV on, depth-6) and re-ran both arms in a truly quiet window (sibling batch instance stopped, not just an isolated port):

| condition | arm | median tok/s |
|---|---|---|
| quiet, TQ on + d6 | eager + MTP | 46.7 |
| quiet, TQ on + d6 | lane + MTP | 43.7 (noise-range, if anything slightly negative) |
| quiet, TQ off + d5 (earlier run) | eager / lane | 46.1 / 45.5 (tie) |
| sibling load (~35 req/min), TQ off + d5 | eager / lane | 29-36 / **44.5 (+22%)** |

On our 512GB box the B=1 win exists **only under sibling GPU contention**, at every config we tried. In quiet conditions the two arms tie regardless of TQ/depth.

Hypothesis for the +79.6% delta: your setup note says "isolated test port, prod untouched during the run" — if prod kept serving on the same 256GB box, that is exactly our loaded condition, and a 256GB box (104GB model + KV, tighter headroom) would contend harder than our 512GB, compounding the effect; plus per-step latency vs our end-to-end stream metric. The mechanism we both observe is real — **eager degrades under contention, the compiled lane holds** — the open question is only the magnitude's measurement basis, not the direction.

moc375:
**Reproduction attempt of the B=1 result — process, facts, and non-reproduction statement.**

**Our setup:** M3 Ultra 512GB, production lane deployment branch of this PR (`OMLX_QWEN4_COMPILED_DECODE` flag), `Jundot-Qwen3.8-Flash-Next-oQ4e-mtp`, isolated canary port, B=1 single stream, 5×256-token rounds per arm, alternating eager/lane/eager across restarts. Metric: end-to-end stream tok/s (completion_tokens ÷ (wall − TTFT), via `stream_options.include_usage`).

**Three rounds:**

| round | sibling batch instance | config | eager median | lane median |
|---|---|---|---|---|
| 1 | serving ~35 req/min | TQ off, MTP d5 | 36.4 / 29.2 (two runs) | 44.5 |
| 2 | stopped | TQ off, MTP d5 | 46.1 | 45.5 |
| 3 | stopped | **TurboQuant KV on, MTP depth-6 (config aligned to the benchmark above)** | 46.7 | 43.7 |

**Observations (facts only):**
- Under concurrent sibling load: lane 44.5 vs eager 36.4 (+22%); a second eager run degraded further to 29.2 with TTFT 0.58→0.60s.
- In both quiet rounds (rounds 2 and 3) the two arms were within run-to-run noise (≤2.8% apart), at both TQ-off/d5 and TQ-on/d6 configs.
- **The reported +79.6% at B=1 was not reproduced in any of the three rounds.** Largest observed delta was +22%, and only under concurrent sibling load.

**Remaining setup deltas vs the reported benchmark:** 512GB vs 256GB host; end-to-end stream tok/s vs per-token latency metric.

Issue #3317, opened 2026-08-30
Boundary snapshots never captured on Qwen3.8-27B (qwen3_5 hybrid) in 0.6.4 — available_boundaries=0, worked in 0.6.2


mikestanley00:
Summary:
Prefix cache stores fail on every request for a Qwen3.8-27B ArraysCache hybrid model under 0.6.4. Every store is skipped with reason=boundary_snapshot_unavailable available_boundaries=0, including prompts of 24–31K tokens where six or more boundaries should exist at block_size=4096. The same model on the same machine stored boundaries reliably under 0.6.2 at prompt sizes up to 113K.

Net effect: no cross-session reuse, and within a multi-turn session each turn re-prefills the full prompt. A turn adding 647 tokens to an existing 25,412-token prompt still costs ~26s of prefill.

Environment

oMLX 0.6.4 (upgraded from 0.6.2 on 2026-08-30) 
macOS 26.6.2, Mac Studio M4 Max, 128 GB
Model: johninthepool/Qwen3.8-27B-MTPLX-8bit, model_type=qwen3_5, Lightning MTP active
Paged SSD-only mode, cache dir on external SSD, block_size enlarged 256→4096 for ArraysCache hybrid

Working — 0.6.2, 2026-08-30 02:20–03:26

29 consecutive successful stores, no failures:

02:20:51 Using boundary cache snapshot for ebdfa15f: storing 77824/78571 tokens (0 intermediate snapshots)
02:24:28 Using boundary cache snapshot for cf4a3e45: storing 81920/85974 tokens (22 intermediate snapshots)
03:26:38 Using boundary cache snapshot for 657bb67a: storing 108544/109322 tokens (0 intermediate snapshots)

Failing — 0.6.4, same day

10:33:23 Skipping cache store for 03f54a22: reason=boundary_snapshot_unavailable tokens=26079 block_size=4096 available_boundaries=0
10:38:42 Skipping cache store for 4befcd01: reason=boundary_snapshot_unavailable tokens=31345 block_size=4096 available_boundaries=0
11:44:18 Skipping cache store for ccdc3b08: reason=boundary_snapshot_unavailable tokens=23801 block_size=4096 available_boundaries=0

Failing at 24K after succeeding at 113K on the previous release.

[omlx-0.6.4-boundary-failures.txt](https://github.com/user-attachments/files/31617318/omlx-0.6.4-boundary-failures.txt)
[...]

Issue #3314, opened 2026-08-30
[0.6.4][Qwen3.8-Flash-Next-MLX-8bit-MTP] Repeated boundary_snapshot_unavailable around 128K coincides with incorrect/truncated turns; output recovers after prefix rebuild


wake-me:
<html><head></head><body><h1>[0.6.4][Qwen3.8-Flash-Next-MLX-8bit-MTP] Repeated <code inline="">boundary_snapshot_unavailable</code> around 128K coincides with incorrect/truncated turns; output recovers after prefix rebuild</h1><h2>Summary</h2><p>During a long agentic session (~60 requests, context growing from ~13K to &gt;150K tokens) on oMLX 0.6.4 with prefix caching and Lightning MTP enabled, <code inline="">Vontra/Qwen3.8-Flash-Next-MLX-8bit-MTP</code> entered a multi-turn correctness failure window around 128K context.</p><p>For several consecutive turns:</p><ul><li><p>every cache store was rejected with <code inline="">reason=boundary_snapshot_unavailable</code>;</p></li><li><p><code inline="">available_boundaries</code> was repeatedly <code inline="">0</code>;</p></li><li><p>assistant output became role-confused / internal-monologue-like;</p></li><li><p>replies terminated abnormally early, sometimes mid-sentence or with an incomplete token fragment;</p></li><li><p>the server still reported <code inline="">finish_reason=stop</code>.</p></li></ul><p>Immediately before this window, the same session was producing normal output.</p><p>After an approximately 25-minute idle interval, the next long-context request rebuilt/materialized the prefix and reported:</p><pre><code class="language-text">storing 131072/132819 tokens
(skipping trailing partial block, 31 intermediate snapshots)
[...]

PR #3335, opened 2026-08-31
fix(cache): skip decode-time boundary captures beyond the cacheable bound

ivaniguarans:
## Summary

On hybrid models that require boundary snapshots, decode-time capture (`Scheduler._maybe_capture_boundary_snapshot`) fires whenever the emitted token count lands exactly on a block boundary. For reasoning-model requests (`needs_think_prefix`), the final-response store deliberately caches prompt tokens only — `<think>` output is stripped by the API layer before the next turn, so generated tokens never prefix-match. A decode-time snapshot is keyed at an emitted position beyond the prompt, which means the store's alignment filter (`tc <= len(cacheable_sequence)`) can never select it. The capture still pays the full non-sliceable state extraction on the inference thread plus an fp32 GDN sidecar SSD write (~111 MiB per event on Qwen3.8-Flash-Next) before the store discards it.

This change skips the capture up front when the position is beyond the request's cacheable bound, and records the skip in the boundary diagnostics (`capture_skipped` / `beyond_cacheable_bound`) so the suppression stays observable. It also surfaces the store-skip `cause` in the existing INFO log line: today the log says only `reason=boundary_snapshot_unavailable available_boundaries=N`, while the decisive cause (`no_snapshots`, `no_aligned_snapshots`, …) is visible only through `/admin/api/stats` — the reporters in #3317 and #3314 spent days guessing at exactly this.

Measured on an M5 Max 128 GB with `Qwen3.8-Flash-Next-oQ4e-mtp` (MTP on, hot+SSD cache, block_size 2048): in a deterministic 9-turn agentic session shape (~32k→47k context, a few hundred tokens per turn, pinned transcripts, temperature 0), stock v0.6.4 produced 14 captures of which 7 were decode-time and all 7 were discarded at store — roughly 770 MiB of dead sidecar writes per session, with each wasted boundary re-captured by the next turn's prefill anyway. With this fix the same workload suppresses exactly those 7 (visible as `beyond_cacheable_bound: 7` in the diagnostics), keeps the 7 prefill captures and their stores byte-for-byte identical, and reproduces the identical per-turn `cached_tokens` sequence (30,720 → 43,008). No reuse behavior changes; only the wasted work goes away.

## Changes

- `omlx/scheduler.py`: `_maybe_capture_boundary_snapshot` returns early — before extraction and the sidecar save — when `request.needs_think_prefix` and the capture position exceeds `request.num_prompt_tokens` (the same bound the final store applies); records `capture_skipped` with reason `beyond_cacheable_bound`. The store-skip INFO line gains a `cause=` token.
- `omlx/cache/observability.py`: `BoundarySnapshotDiagnostics.record()` now returns the assembled event dict so the log site can include the derived cause without a second snapshot round-trip. Additive; no counter or schema changes.
- Tests: a regression test proving the skip (no extract call, no sidecar save, reason recorded — fails before the patch), a contrast test proving the bound is strict (a reasoning-model capture at exactly the prompt boundary still proceeds), a `record()`-return-value test, and a `cause=` assertion added to the existing store-skip caplog test. The existing block-boundary capture test covers the non-reasoning path unchanged.

## Test plan

- `uv run pytest tests/` — 10595 passed, 149 skipped, 0 failed (4m31s, Python 3.12, M5 Max). Five tests were deselected because they fail identically on stock v0.6.4 on this machine with zero local changes (fp-tolerance misses in `test_dflash_laguna`, `test_mlx_vlm_glm5_next_compat`, and three qwen4_exp gather-path tests; possibly M5-generation numerics — happy to file separately) and are unrelated to this change.
- New tests verified to fail before the patch and pass after.
- Live A/B on the workload above, both arms from a fresh scheduler: patched arm shows `capture_attempts` reduced exactly by the baseline's decode-time count, `beyond_cacheable_bound` equal to it, identical stores and identical per-turn `cached_tokens`.

Related to #3317, #3314 — this does not resolve the pinned-reuse regression reported there (which we could not reproduce at block_size 2048 with first-party checkpoints on either qwen3_5 or qwen4_exp), but it removes the wasted decode-capture work in the same path and makes the skip cause visible in the server log, which would have named the failure mode for those reports directly.

— Generated with Marshal · gate: READY · 5274c38 · tier: standard
verify bar: passed at 5274c38 — 1 check(s) + secrets scan, sha-keyed proof

### DIFF
--- a/omlx/cache/observability.py
+++ b/omlx/cache/observability.py
@@ -115,7 +115,7 @@ class BoundarySnapshotDiagnostics:
         source: str | None = None,
         storage: str | None = None,
         available_boundaries: int | None = None,
-    ) -> None:
+    ) -> dict[str, Any]:
         counter = {
             "capture_attempt": "capture_attempts",
             "capture_success": "captures",
@@ -157,6 +157,9 @@ class BoundarySnapshotDiagnostics:
                 if value is not None:
                     last_event[key] = value
             self._last_event = last_event
+            # Callers logging the event (e.g. the store-skip INFO line) need
+            # the derived cause, which only exists on this assembled record.
+            return dict(last_event)
 
     def snapshot(self) -> dict[str, Any]:
         with self._lock:
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -6723,6 +6723,22 @@ class Scheduler:
         if not self._detect_boundary_snapshot_need():
             return
 
+        # Reasoning-model requests cache prompt tokens only (the
+        # needs_think_prefix branch of the final-response store), so a
+        # snapshot keyed past the prompt can never pass
+        # _get_boundary_store_override's tc <= len(cacheable_sequence)
+        # filter. Skip before paying the extraction and the sidecar write.
+        if request.needs_think_prefix and total_tokens > request.num_prompt_tokens:
+            self._boundary_snapshot_diagnostics.record(
+                "capture_skipped",
+                reason="beyond_cacheable_bound",
+                request_id=request.request_id,
+                token_count=total_tokens,
+                block_size=block_size,
+                source="decode",
+            )
+            return
+
         self._boundary_snapshot_diagnostics.record(
             "capture_attempt",
             request_id=request.request_id,
@@ -11316,7 +11332,7 @@ class Scheduler:
                             available_boundaries = len(
                                 self._boundary_cache_snapshots.get(request_id, {})
                             )
-                            self._boundary_snapshot_diagnostics.record(
+                            skip_event = self._boundary_snapshot_diagnostics.record(
                                 "store_skip",
                                 reason="boundary_snapshot_unavailable",
                                 request_id=request_id,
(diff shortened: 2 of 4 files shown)


scubamount:
## Verified

- **Root cause correct:** hybrid models requiring boundary snapshots fire `_maybe_capture_boundary_snapshot` on every block-aligned decode token. For `needs_think_prefix` (reasoning) requests the store caches prompt tokens only (`tc <= len(cacheable_sequence)`), so a decode snapshot keyed past the prompt can never be selected — yet extraction + ~111 MiB GDN sidecar SSD write was still paid.
- **Fix right:** early return when `request.needs_think_prefix and total_tokens > request.num_prompt_tokens` with `capture_skipped / beyond_cacheable_bound` diagnostic; no extraction or write. `observability.record` now returns the assembled event so the existing `boundary_snapshot_unavailable` INFO line can log the derived `cause`.
- **Tests:** `test_decode_capture_skipped_beyond_cacheable_bound` and `test_boundary_snapshot_diagnostics_record_returns_event_with_cause` cover both the skip and the returned cause.

No issues. Eliminates useless work before it is paid.

PR #3336, opened 2026-08-31
feat(cache): apply the GDN state codec to embedded block payloads (#2700)

### LINKED ISSUE #2700: GDN storage codecs are unreachable from the `embedded` snapshot layout
## Summary

#2644 added storage codecs (`fp32|bf16|int8|rht_int8|rht_int16`) and the `gdn_snapshot_storage` policy (`auto|ssd_sidecar|embedded`) for persisted GDN sidecars. The codec is applied at exactly one place — `BoundarySnapshotSSDStore._encode_gdn_state()` — which is on the SSD sidecar path only.

The `embedded` layout writes the recurrent state through `paged_ssd_cache._store_nstate_elements_flat()` as raw fp32, and the restore side `_load_nstate_flat()` returns `arrays[elem_key]` verbatim with no codec awareness. So of the three `gdn_snapshot_storage` values, **`embedded` is the only one with no access to the codecs**, even though it is the layout a user lands on when they turn the SSD tier off. That is not an edge case: the admin API rejects `ssd_sidecar` together with `hot_cache_only` with a 400 ("gdn_ssd_split_enabled cannot be used with hot_cache_only"), and `auto` itself resolves to `embedded` whenever `hot_cache_only` is set or the cache tier is disabled (`CacheConfig.effective_gdn_ssd_split_enabled`).

This matters beyond disk footprint, because hot-cache entries are charged their serialized bytes:

```python
# PagedSSDCacheManager._hot_cache_entry_size
if "tensors_raw" in entry:
    return sum(len(raw) for raw, _, _ in entry["tensors_raw"].values())
```

Under `embedded`, the fp32 recurrent state sits inside those very bytes, so it is spending hot-cache budget that a codec would free. (The saving is on the `save_block()` path quoted above; entries promoted back from SSD take the `arrays` branch and are charged their decoded `nbytes` either way.)

## Measurements

Model: Qwen3.8-27B-8bit-mtp — 64 layers = 48 `linear_attention` (ArraysCache) + 16 `full_attention`, 4 KV heads x 256 head_dim fp16, block size 2048 tokens (`_ARRAYS_CACHE_BLOCK_SIZE`). Recurrent state `(1, 48, 128, 128)` fp32, conv state `(1, 3, 10240)` bf16. Apple M5 Max, 128 GB.

**Block payload composition.** KV per block = 134,217,728 B; 48-layer fp32 GDN state = 153,944,064 B. The recurrent state is **53.4%** of an `embedded` block payload for this model.

**Hot-cache retention at the default 10 GB budget.** Real block payloads pushed through `save_block()` until LRU eviction reached steady state; depth read back from the manager:

| GDN state | block entry | resident blocks | chain tokens | vs fp32 |
|---|---:|---:|---:|---:|
| fp32 (today) | 288,161,792 B | 37 | 75,776 | — |
| bf16 | 212,664,320 B | 50 | 102,400 | 1.35x |
| rht_int16 | 213,843,968 B | 50 | 102,400 | 1.35x |
| rht_int8 | 176,095,232 B | 60 | 122,880 | 1.62x |

The GDN state itself shrinks 1.93x, diluted to 1.35x at block granularity by the KV bytes beside it. **This ratio is geometry dependent** — a model with a different KV-to-GDN ratio or block size will land elsewhere; the 1.35x is not a universal claim.

**Codec cost**, 48-layer state, median of 7 (source: a live sidecar written by this build, dequantized back to fp32):

| codec | encode | decode (from bytes) | payload | compression |
|---|---:|---:|---:|---:|
| fp32 | 3.93 ms | 2.21 ms | 153,944,064 B | 1.000x |
| bf16 | 9.11 ms | 2.81 ms | 78,446,592 B | 1.962x |
| int8 | 8.24 ms | 11.06 ms | 41,877,504 B | 3.676x |
| rht_int8 | 9.29 ms | 11.90 ms | 41,877,504 B | 3.676x |
| rht_int16 | 9.95 ms | 11.97 ms | 79,626,240 B | 1.933x |

End to end this is **+6.2 to +8.5 ms per block boundary** (encode plus the larger/smaller byte extraction), measured through `save_block()`.

Two things worth noting for anyone extending the codec to another path:

1. The cost is not the quantization arithmetic (0.45 ms for rht_int16) but the batched `isfinite` validity check (4.3 ms) — one GPU->CPU sync. The existing comment in `_encode_gdn_state` about avoiding a per-layer `bool()` is doing most of the work already; what remains is the single sync.
2. bf16 compresses marginally better than rht_int16 (1.962x vs 1.933x — no row scale tensor) and decodes 4.3x cheaper, but at ~64x the relative L2 error (0.16% vs ~0.0025%) and, at this block size, **identical** retention depth. So this is not an argument for a different default; rht_int16 remains a sound default. It is only relevant if decode latency ever dominates.

**Warm-restore latency is not measurably harmed.** Live A/B on this build, Qwen3.8, MTP off, `general-code`, one cold plus two warm per condition, empty cache per arm, all conditions VALID:

| context | fp32 warm | rht_int16 warm | delta |
|---:|---:|---:|---:|
| 8,192 | 4.438 s | 4.419 s | -0.019 s (-0.43%) |
| 65,536 | 7.152 s | 6.822 s | -0.330 s (-4.62%) |

With n=2 warm runs per condition this does not establish that rht_int16 is *faster*; the claim is only that the codec's decode cost does not surface as a warm-restore penalty, consistent with the smaller sidecar cutting I/O by more than the ~12 ms dequantization adds.

**Output-level effect of the codec.** The proposal does not change which codec is the default, but it does put a codec where fp32 used to be, so it is worth stating what that costs at the output. Greedy decoding here is deterministic: restarting the server and re-running the fp32 arm reproduces all 50 probes byte for byte, at both 8K and 64K. Against that zero floor, the fraction of open-ended probes whose continuation changes when the state is restored through a codec:

| GDN state codec | state rel. L2 | flip rate @8K | flip rate @64K |
|---|---:|---:|---:|
| fp32 (re-run, noise floor) | 0% | 0/50 | 0/50 |
| rht_int16 | 0.0024% | 4/50 | 4/50 |
| rht_int8 | 0.617% | 5/50 | 2/50 |
| rht_int4 (local-only control) | 11.19% | 16/50 | 15/50 |

Two caveats belong with those numbers. The measure has a **detection floor**: `rht_int16` and `rht_int8` differ by 250x in state error and do not separate (they even swap order between contexts), so this evidence supports "the codec does not disturb the model much", not a ranking between candidate codecs. And a flip is a *difference*, not a regression — the diverging continuations read as equally plausible; nothing here measures preference. Separately, tasks whose answer is fixed by prompt tokens are insensitive regardless of output length: a 57-item ordered enumeration and a 49-step arithmetic trace stayed byte-identical even under the 11.19% control, which is expected in a hybrid model where the attention KV is restored without passing through the codec.

## Proposal

Give the `embedded` layout the same codec access the sidecar path has:

1. an encode hook in `_store_nstate_elements_flat()`;
2. a matching decode hook in `_load_nstate_flat()`;
3. one codec key per element in the metadata channel that already exists — `cache_list_meta` already carries `_state_class_name`, `_zero_dim` and `_nested` per element, so no new metadata channel or file format is needed.

Payload-schema compatibility is already handled by the existing `payload_layout` / `_cache_compat_signature` gate and by #2487's "matching content hash does not guarantee matching serialized layout" replacement path.

Worth calling out because the two paths cannot use the same mechanism: on the sidecar path the codec is stamped into a *sidecar namespace* signature (`gdn_cache_signature_for`), deliberately kept out of the KV block signature so changing the codec does not invalidate KV blocks. Under `embedded` the state lives inside the block payload, so there is no separate namespace to isolate — which is exactly why the codec key belongs in the per-element metadata (3) rather than in the signature. A self-describing block keeps existing fp32 blocks readable after the setting changes, and a mixed-codec cache directory stays valid, with no signature bump and no cache invalidation.

Before writing a PR: is extending the codec to `embedded` something you'd take, and would you prefer the setting to stay a single `gdn_sidecar_state_dtype` covering both layouts (with the name becoming inaccurate), or a separate key for the embedded path?

## Scope and caveats

- The benefit accrues to `embedded` / `hot_cache_only` deployments. A `ssd_sidecar` deployment already keeps only a small structural placeholder in the block payload, so it sees no change.
- The 1.35x block-level figure is specific to this model's KV-to-GDN ratio and the 2048-token block size chosen by `_enlarge_block_size_for_arrays_cache`.
- Measured on one machine and one model family (Qwen3.x GDN hybrids).

williamxie1989:
Replaces #2709 — this is a rebase of @hojin12312's work onto current `main`, with all four original commits preserved under his authorship (`git log` shows `A:hojin12312` on every commit; my contributions are the conflict resolution, the three policy alignment changes and the admin-surface gap fix listed below). His branch predated several `main` changes, which is presumably why #2709 stalled — the design itself is exactly what #2700 asked for, including the layout-independent `gdn_snapshot_state_dtype` name you requested in https://github.com/jundot/omlx/issues/2700#issuecomment.

Closes #2700.

## What #2709 does (unchanged)

The GDN state codec from #2644 only ran on the SSD-sidecar path. Under `embedded` — which is what `auto` resolves to whenever `hot_cache_only` is on or the SSD tier is off — the same tensor was written as raw fp32, and hot-cache entries are charged their serialized bytes, so the deployment that can least afford it paid the most. This makes the codec layout-independent: encoded payloads live under `layer_{i}_state_{k}__q` / `..._{k}__scale` so older readers take their existing missing-state path and skip the block instead of feeding int16 to the model as fp32.

## Where this rebased version deliberately differs from #2709

`main` moved under the branch (the v0.6.0 lossy-default incident and the public-key rename both landed after #2709's base). To keep this mergeable without re-litigating that:

1. **Default stays `fp32`.** #2709 defaulted the canonical field to `rht_int16` (true at its base); current `main` intentionally reset every install to exact fp32 after v0.6.0 persisted a lossy default without recording user intent (`test_from_dict_ignores_v060_*`). Reduced precision stays an explicit opt-in.
2. **On-disk key order: `gdn_snapshot_state_dtype` → `gdn_sidecar_precision` → `fp32`.** The v0.6.0-era `gdn_sidecar_state_dtype` key remains ignored in settings files (same policy as `main`); it is accepted as an explicit alias on the API request surface only, where a value is always a deliberate selection. `to_dict()` writes the canonical key plus the `gdn_sidecar_precision` public key, not the v0.6.0 spelling.
3. One new test in #2709 asserted the old on-disk alias behaviour; it was rewritten to the `gdn_sidecar_precision` spelling to match `main`'s migration policy.
4. One integration gap in #2709's own commit is closed here: it published the embedded counters on the scheduler's stats dict, but the admin runtime-cache route rebuilds each model row from an explicit field list, so `gdn_embedded_state` never reached `/admin/api/stats` — the deployment that cannot use sidecars at all, which is the whole point of #2700, still could not see whether the codec ran. The last commit maps it through with the same normalization `gdn_staging` gets, with a route-level test.

## Real-world impact (the deployment `main` punishes today)

Qwen3.8-Flash-Next (`qwen4_exp`): 48 layers = 12 `QSAKVCache` + 36 GDN `ArraysCache`, recurrent state 3.15 MB fp32 per layer per 2048-token boundary → **113 MB of fp32 GDN state embedded in every hot block**, i.e. ~55 KB/token on top of 27.9 KB/token of KV+indexer.

Measured on two machines running the same 0.6.4 with the same model family:

| deployment | GDN state placement | hot bytes/token |
|---|---|---:|
| `hot_cache_only=false` (split) | SSD sidecars | 27.9 KB |
| `hot_cache_only=true` (embedded, fp32) | inside every hot block | ~85 KB |
| `hot_cache_only=true` + this codec (rht_int16) | inside hot blocks, encoded | ~57 KB |

Measured on the live hot-only deployment after enabling the codec: 56.7 KB/token across 170 hot blocks, matching the 2 B/element prediction (bf16 and rht_int16 store the same byte count).

That 2.7× was investigated down to the byte before landing here, which is how #2700 turned out to be the whole story: hot-only users are charged fp32 GDN state at 2× the KV cost with no way to reach the codec.

## Testing

- #2709's `test_gdn_embedded_state_codec.py` (420 lines of round-trip/fallback) + sidecar-quantization + GDN-split suites: **139 passed**
- all cache-related tests: **1507 passed**
- full fast suite: 10736 passed; the 19 failures are pre-existing on `main` and byte-identical to a `main` baseline run (deepseek/glm/qwen4-vendor numeric-comparison tests, unrelated files)


hojin12312:
Thanks for rebasing this onto the latest main and preserving the original commits/authorship. The implementation looks consistent with #2700/#2709 overall.

Two small things: the UI currently says RHT + int16 is the default, which seems inconsistent with the PR’s stated policy that the default remains fp32; and the old gdn_sidecar_state_dtype alias may no longer work as a dataclass constructor keyword. Might be worth checking those before merge; otherwise no objections from my side.

williamxie1989:
@hojin12312 thanks for catching both — reproduced and fixed in the two commits just pushed.

**1. UI default text** (`b86f3d13`): the hint was carried over from #2709, where `rht_int16` genuinely was the default; this rebased version deliberately keeps `fp32`. Reworded `gdn_snapshot_state_precision_hint` in every locale to *"FP32 is the exact default and uses the most space; RHT + int16 is near-lossless at half the bytes; …"* (ko translated accordingly), and pinned it with a locale-wide test in `tests/test_admin_gdn_storage_ui.py` that fails if any locale advertises a lossy default.

**2. Constructor keyword** (`9b699582`): confirmed — `PagedSSDCacheConfig(gdn_sidecar_state_dtype=...)`, `CacheSettings(...)` and `SchedulerConfig(...)` all raised `TypeError` after the rename turned the alias into a property (the generated `__init__` only knows the canonical field; `dataclasses.replace()` hit the same path). Fixed with a small `deprecated_init_kwargs` shim (`omlx/utils/legacy_kwargs.py`) applied to all three classes. One deliberate precedence detail: an explicitly passed legacy keyword overrides the canonical keyword, because `replace()` re-materializes every init field from the current instance and a canonical-first `setdefault` would silently swallow `replace(cfg, gdn_sidecar_state_dtype=...)`. Dict-shaped surfaces (settings files, env, admin API) keep canonical-first, unchanged. `tests/test_legacy_kwargs.py` covers constructor kwargs, `replace`, attribute aliases and precedence for all three classes.

Verification: 576 passed across the alias / UI / settings / admin-alias / GDN codec suites; the full fast suite is byte-identical to a HEAD baseline run on this machine (47 failures, all order-dependent numeric-comparison tests that pass standalone — unrelated to this PR).

scubamount:
## Verified

- **Purpose correct:** rebase of #2709 (hojin12312) — extracts layout-independent `gdn_state_codec` and routes `boundary_snapshot_store` through `encode_state`/`decode_state`/`parse_encoding` with alias shims.
- **API compat:** `GlobalSettingsRequest` 3-way fallback `gdn_snapshot_state_dtype` → `gdn_sidecar_state_dtype` (API) → `gdn_sidecar_precision`, lower-cased, 400 on invalid. `get_global_settings` returns all three keys; `to_dict` keeps canonical + `gdn_sidecar_precision` only (v0.6.0 migration).
- **Other:** dashboard enables selector always (`:disabled="!enabled"` vs old `embedded||hot_cache_only` gate) — intended; `gdn_embedded_state` observability mapped; `min_rank` placeholder handling avoids encoding 4-byte pads. CI green 3.11/3.12/3.13.

Nit (non-blocking, worth a second look): `ko.json` doesn't just differ cosmetically — it MOVES the "(recommended)" marker from FP32 to `rht_int16` (`FP32(정확)` + `RHT + int16(권장)`) while all 8 other locales keep `FP32 (exact, recommended)` and plain `RHT + int16`. Since the warning string itself still says "use FP32 unless you explicitly accept the trade-off", the Korean UI ends up recommending the lossy mode while the same file's warning recommends FP32. If that's a deliberate KR-market default, ignore; otherwise align with the other locales. Otherwise PASS.

PR #3340, opened 2026-08-31
fix(ane): clear the per-module state caches when releasing ANE banks

andyoneal:
`release_qwen35_ane_prefill` exists to hand back the compiled ANE procedure banks — the docstring puts them at roughly 13 GB for a 27B at mlp 0.35 / gdn 0.45 — once a long-context prefill needs that resident memory back for the KV cache. It latches every sliced module through the existing per-module failure flags, drops the state attributes, and relies on the native models freeing their programs and mapped blobs when the last reference dies.

`_compile_pair` and `_compile_gdn` also cache their state per module, in `_omlx_ane_prefill_cache` and `_omlx_ane_gdn_cache`. The release cleared neither, so a second reference always survived and nothing was freed. `engine_pool` brackets the call with `get_phys_footprint()` and a `gc.collect()` specifically to reclaim that memory, which made the release a no-op on the one path that exists to perform it.

The caches are cleared unconditionally rather than alongside a live state: an entry left over from an earlier slice is exactly the mapped bank the call is being asked to free, and a subsequent enable recompiles instead of reusing it. A module whose cache holds only such a leftover is not counted in `modules_released`, so the shed marker and counter reset stay tied to modules that were actually serving. No `_COMPILE_LOCK` is taken — compiles and this release both run on the engine's single-worker MLX executor (`engine_core.py`, `max_workers=1`), so a concurrent compile cannot observe the cache while it is being cleared.

Both new tests hold a weak reference to each native handle and assert it has been collected once the release returns. They fail without the change, on that assertion rather than on the cache contents:

```
>       assert mlp_ref() is None
E       AssertionError: assert <_NativeHandle object at 0x1086782f0> is None
```

Not addressed here: the caches are also unbounded, so an enable at a new chunk width or fraction leaves the previous entry behind. No current path reaches that, since those settings enter the engine runtime signature and changing one reloads the engine, taking the caches with it. Bounding the cache is a separate change.

## Validation

- Full fast suite on macOS 26.6.2 / Apple M1 Max, MLX 0.32.2, custom kernels built: 10,737 passed, 30 skipped, 6 failed. The six are pre-existing float-tolerance comparisons in `test_decode_fast_sdpa.py`, `test_glm_moe_dsa_patch.py` and `test_mlx_vlm_qwen4_exp_compat.py` — custom kernel comparisons in files this change does not touch, and a standing baseline on this machine rather than a regression.
- `ruff check` clean on both changed files under the project config.

### DIFF
--- a/omlx/patches/qwen35_ane_prefill.py
+++ b/omlx/patches/qwen35_ane_prefill.py
@@ -3562,8 +3562,9 @@ def release_qwen35_ane_prefill(model: Any) -> tuple[int, int]:
     prefill needs back once the KV cache has grown into the guard's sizing
     target. Latches every sliced module through the existing per-module
     failure flags first (so the dispatch sites fall back to stock GPU compute
-    and never lazily recompile), then drops the state references; the native
-    models free their programs and mapped blobs when the last reference dies.
+    and never lazily recompile), then drops the state references and clears the
+    per-module state caches that hold the same objects; the native models free
+    their programs and mapped blobs when the last reference dies.
     Idempotent, and scoped to this engine instance: the next load of the
     model rebuilds the banks from its settings.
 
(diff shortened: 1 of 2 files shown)


scubamount:
## Verified

- **Root cause correct:** `_compile_pair` / `_compile_gdn` caches held a second reference to ANE banks after `release_qwen35_ane_prefill` cleared `state_attr` → GC never freed ~13 GB (27B, mlp 0.35 / gdn 0.45).
- **Fix right:** clears both `_omlx_ane_prefill_cache` and `_omlx_ane_gdn_cache` via `cache.clear()` unconditionally (covers stale entries where state attr already `None`); `if cache:` guard safe for dict, no `_COMPILE_LOCK` needed — serialized on single-worker MLX executor.
- **Tests:** `weakref` proves native handles collected; both new tests fail pre-fix on `assert mlp_ref() is None`.

No issues. Completes the bank hand-back.

jundot:
Thanks for fixing the retained cache references. I reproduced the failure with Qwen3.8-27B in single-ANE mode: all 112 module caches survived release, and the request was rejected by the memory guard. With the three patches applied, those caches were cleared and the same request completed under the same 28 GiB limit. LGTM, merging.

PR #3346, opened 2026-09-01
fix(vlm): claim mRoPE rope_deltas onto the request (#3345)

### LINKED ISSUE #3345: [0.6.4] Qwen3.8-Flash-Next: engine loop dies after a long MTP generate
**Describe the bug**

On oMLX 0.6.4 (macOS `.app`), a long `/v1/chat/completions` turn on Qwen3.8-Flash-Next (`qwen4_exp`, VLM engine, Lightning MTP on) finishes normally. The next request in the same process dies with:

```
ValueError: [convert] Only length-1 arrays can be converted to Python scalars.
```

The engine loop stays down after that. Every later chat request in that process fails the same way. Unload and reload of the model is the only thing that brought chat back. A later long generate did it again.

**To Reproduce**

1. oMLX 0.6.4 `.app`, Qwen3.8-Flash-Next (oQe + Lightning MTP) loaded. MTP depth 3. Prefix cache on. Text-only chat, no images.
2. Run a growing agent session through `/v1/chat/completions` (tool calls, context in the low thousands).
3. One completion writes a very long reply. Mine was 18091 tokens in 426s (43 tok/s), prompt 3618, `max_tokens` left at the leftover context window (~254k). MTP log for that turn: `tok/cycle=3.26`, accept 12532/14113. Scheduler then stored a boundary snapshot `20480/21709`.
4. Send any follow-up in the same process, even a one-word ping. Engine loop error, HTTP 500. Repeat pings keep failing.

I have not needed a second concurrent request. Sequential turns are enough.

**Expected behavior**

The long turn stores its prefix and returns. The next turn prefills and decodes.

**Desktop**

- macOS, Apple Silicon
- oMLX 0.6.4 (`.app`, Info.plist build 2529)
- Model: Qwen3.8-Flash-Next, `qwen4_exp`, served on the VLM engine
- Lightning MTP on, depth 3. ANE off. Thinking off.

**What the log shows**

After the 18k-token completion, the next schedule step:

```
Engine loop error: [convert] Only length-1 arrays can be converted to Python scalars.
  scheduler.py  step -> _schedule_waiting
  models/vlm.py  get_last_rope_deltas
    return float(rd.item())
```

About 50 minutes later, still in the same process, two more pings died on a second site:

```
  scheduler.py  _schedule_waiting
    request.rope_deltas = float(captured.item())
```

**Related**

- #1800 — same `ValueError` string, `cache_offset` array vs scalar on a Qwen VLM, concurrent second request. Different stack.
- #1510 (closed), #2252 — same string, DFlash / `int(cache.offset)`. Different stack.
- #3314 — same 0.6.4 Qwen3.8-Flash-Next + Lightning MTP long agent session. That report is `boundary_snapshot_unavailable` and garbled turns. This session also stored a boundary (`20480/21709`) and then died on the next schedule.

Happy to attach a redacted `server.log` slice if useful.

wolfyy970:
## Summary

Fixes #3345. After a long Lightning MTP generate on Qwen3.8-Flash-Next, the next `/v1/chat/completions` turn killed the engine loop with `Only length-1 arrays can be converted to Python scalars`.

`_rope_deltas` lived on the shared language model. A copy also sat in `vlm_extra_kwargs`, the same dict prefill slices and the adapter forwards. That bag is the wrong home for a per-request scalar. The key got popped or seq-sliced, and the scheduler then called `.item()` on leftover LM state.

This PR owns a float on `Request.rope_deltas` at admit time, takes the snapshot out of extras, copies extras before the adapter pops, and resets leftover mRoPE fields before the next embed pass. Decode uses the request field. It only falls back to the live LM when no embed snapshot existed.

## Test plan

- [x] `rope_deltas_as_float` on scalar, `(3, 1)`, and `[10, 20]` (the old `.item()` throw)
- [x] `claim_request_rope_deltas` lifts the snapshot and leaves `position_ids` in extras
- [x] Adapter forward does not steal `_captured_rope_deltas` from the caller's dict
- [x] `Scheduler.add_request` + `_schedule_waiting` keep the claimed `12.0` and do not call `get_last_rope_deltas`
- [ ] Long Flash-Next MTP generate, then a one-word follow-up in the same process (the #3345 session)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -45,7 +45,11 @@ from ..api.utils import (
 )
 from ..cache.vision_feature_cache import VisionFeatureSSDCache
 from ..exceptions import InvalidRequestError
-from ..models.vlm import VLMModelAdapter
+from ..models.vlm import (
+    VLMModelAdapter,
+    reset_lm_mrope_state,
+    rope_deltas_as_float,
+)
 from ..patches.mlx_vlm_pixtral_torch_free import apply_pixtral_torch_free_patch
 from ..reasoning_effort import apply_chat_template_with_reasoning_effort_fallback
 from ..utils.image import (
@@ -3236,6 +3240,10 @@ class VLMBatchedEngine(BaseEngine):
             # Run vision encoder + embedding merge.
             # Pass attention_mask as 'mask' — mlx-vlm models (e.g. Gemma 3)
             # expect it as a positional/keyword arg named 'mask'.
+            # Drop leftover MTP / prior-request mRoPE fields first so
+            # get_input_embeddings recomputes deltas for *this* prompt
+            # instead of tiling a stale (batch, 1) vector.
+            reset_lm_mrope_state(getattr(self._vlm_model, "language_model", None))
             try:
                 embed_features = self._vlm_model.get_input_embeddings(
                     input_ids, pixel_values, mask=attention_mask, **call_kwargs
@@ -3279,7 +3287,7 @@ class VLMBatchedEngine(BaseEngine):
                     extra_kwargs["position_ids"] = pid
                 rd = getattr(lm, "_rope_deltas", None)
                 if rd is not None:
-                    extra_kwargs["_captured_rope_deltas"] = rd
+                    extra_kwargs["_captured_rope_deltas"] = rope_deltas_as_float(rd)
 
             # Extract token IDs as list
             token_ids = (
--- a/omlx/models/vlm.py
+++ b/omlx/models/vlm.py
@@ -28,6 +28,53 @@ import mlx.nn as nn
 logger = logging.getLogger(__name__)
 
 
+def rope_deltas_as_float(rd: Any) -> float:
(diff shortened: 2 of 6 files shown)


jundot:
@wolfyy970 Thank you for the detailed issue analysis and for taking the time to implement a complete fix as well. I reproduced both `.item()` failure sites and confirmed that this PR prevents the crash.

After tracing the production prefill path, I found that `_slice_vlm_extra()` and `_advance_vlm_extra()` already pass copied dictionaries, so the request snapshot is not lost when the adapter removes the internal key. Qwen's `get_input_embeddings()` also recomputes its mRoPE state for the new image input. Because of that, the additional request-bound state, repeated claim step, and shared LM state reset are broader than this fix requires and increase the regression surface across other VLM families.

I applied a narrower patch that selects the current request's first batch-row delta at the two failing scalar conversion sites, with regression coverage for both paths. Your report and PR made the failure mode and mlx-vlm shape contract much easier to verify, and I really appreciate the work you put into both.

PR #3351, opened 2026-09-01
fix(qwen4_exp): gather text mRoPE prefills and stop pricing them as dense SDPA (#3350)

### LINKED ISSUE #3350: [0.6.4] Qwen3.8-Flash-Next: memory guard refuses a ~5K text continuation off a 233K prefix
**Describe the bug**

On oMLX 0.6.4 (macOS `.app`), a text-only follow-up on Qwen3.8-Flash-Next (`qwen4_exp`, VLM engine, Lightning MTP on) is refused by the prefill memory guard at `progress=0`. Prefix cache reused 233472 of 238716 tokens. The remaining prefill is 5244 tokens. The engine loop stays up. The request fails with `PrefillMemoryExceededError`.

The guard log is:

```
[guard:external] admission terms: current=147.56GB predicted_transient=90.45GB observed_max_bytes=0.00GB
```

Safety cap was 214.01GB (95% of metal_cap 225.28GB). 147.56 + 90.45 is over that, so the first chunk never runs.

`observed_max_bytes=0` and `progress=0` mean 90.45GB was a prediction, not a kernel that already allocated.

**To Reproduce**

1. oMLX 0.6.4 `.app`, Qwen3.8-Flash-Next (oQ6e + Lightning MTP) loaded. MTP depth 3. Prefix cache on. Text-only, no images. Memory guard on. Mine is `custom` / `prefill_priority=speed`, ceiling 240GB, `prefill_min_chunk_tokens=32`. Speed priority makes the abort check use the full remaining suffix, and the Qwen4 floor is 4096, so the 32-token setting does not shrink this one.
2. Grow a `/v1/chat/completions` session until prefix cache is in the ~200k token range. Mine reused 233472.
3. Send a normal continuation (a few thousand new tokens). Guard refuses before the first chunk.

**Expected behavior**

QSA core attends a ~2048 token budget, not `Q × 233k`. A 5k suffix on a 233k prefix should prefill when resident KV is already 147GB and the cap still has ~67GB of headroom.

**Desktop**

- macOS, Apple Silicon (M3 Ultra, 256 GB)
- oMLX 0.6.4 (`.app`)
- Model: Qwen3.8-Flash-Next, `qwen4_exp`, VLM engine
- Lightning MTP on, depth 3. TurboQuant KV off. ANE off.

**What the log shows**

```
prefix cache: request ... re-prefills 5244 of 238716 tokens (reused 233472)
[guard:external] admission terms: current=147.56GB predicted_transient=90.45GB observed_max_bytes=0.00GB ane_prefill_transient_bytes=0.00GB
```

Two mistakes stack.

Gathered QSA eligibility accepts missing or 2-D `(1, L)` positions and rejects 3-D MRoPE. Parent `LanguageModel` tiles identical text positions to `(3, 1, L)` after `_seed_text_only_mrope_delta_for_cached_prefill`. Equal planes are still text. Unequal planes are images and should stay on mask+dense. This continuation would have taken `mask_dense` SDPA if the guard had let it through.

`make_prefill_memory_profile` has no `qwen4_exp` arm, so the static estimate is dense `Q × kv_len`. `_predicted_chunk_transient` then takes max(last_delta, EWMA, static) × 1.3. A leftover dense gulp of ~69.58GB becomes 90.45GB even when a later chunk would be gathered. The tracker already drops 8× spikes from the EWMA blend; last_delta still feeds the predictor.

Charging the 2048-token budget while the kernel is still on mask+dense would undercharge and can SIGABRT. The kernel path and the price have to move together.

**Related**

- #3284 is the other direction on the same model: the guard under-estimates ArraysCache snapshot growth and admits doomed cold prefills. This report is an over-estimate on a cached continuation.
- #3294 is QSA prefix reconstruct `(1,1,S)` vs `(1,3,S)`. Different. This session reused 233472.
- #3345 / PR #3346 is `rope_deltas.item()` on a non-scalar after a long MTP generate. Same model, different failure.

Happy to attach a redacted `server.log` slice if useful.

wolfyy970:
## Summary

Fixes #3350. A text-only continuation on Qwen3.8-Flash-Next reused 233472 prefix tokens, then the memory guard refused the remaining 5244 at `progress=0` (`current=147.56GB`, `predicted_transient=90.45GB`, cap 214GB). The 90.45GB is `last_delta × 1.3` from a leftover dense gulp. Gathered eligibility also rejected 3-D text mRoPE: parent `LanguageModel` tiles identical `(1, L)` positions to `(3, 1, L)` after the text-only cached-prefill rope-delta seed.

Broadcast-identical planes now take gathered QSA on prefill (indexer gets the text row, rotary gets the 3-D ids). Unequal planes stay fail-closed on mask+dense. Decode stays 2-D `(1, 1)` or absent.

Text-only admission, eviction, and chunked prefill charge a `qwen4_exp` profile (indexer still `Q × kv/r`, core capped at `indexer_budget`) when the request is text-only and the active monitor is that profile. Gathered vs dense is an argument, not a toggle on the shared monitor. Leftover dense `last_delta`/EWMA is ignored only on a gathered Qwen4 chunk whose last sample was not gathered. Image prefills keep the dense price. The Qwen4 profile resident path includes measured `fixed_state_bytes`.

## Test plan

- [x] broadcast `(3,1,L)` text mRoPE takes `contiguous_causal_gathered_qsa` on prefill; unequal planes do not
- [x] decode `(3,1,1)` equal planes do not take gathered
- [x] batched and at-budget paths still fail closed
- [x] Qwen4 gathered static is well under dense `Q × kv` (4.04GB vs 46.41GB at Q=4096, kv=233472)
- [x] leftover 69.58GB `last_delta` does not bind on a gathered Qwen4 chunk (~5.5GB predicted; 147+5.5 < 214); same gulp still binds when this chunk is dense
- [x] a gathered spike from this request still binds the next chunk
- [x] Qwen4 text-only preflight, eviction, and `_preflight_memory_check` use gathered; image / `text_only=False` stays dense
- [x] `estimate_resident_kv_bytes` on the Qwen4 profile moves by `set_fixed_state_bytes(123456789)`; `estimate_prompt_kv_bytes` does not
- [x] materialized equal-plane `(3,1,L)` prefill matches the official mask path
- [ ] CI: `pytest tests/ -m "not slow and not integration"`
- [ ] live #3350 continuation after overlay (scheduler + memory_monitor + language.py)

### DIFF
--- a/omlx/admin/context_benchmark.py
+++ b/omlx/admin/context_benchmark.py
@@ -244,7 +244,7 @@ def _guard_ready(scheduler: Any) -> bool:
 def _make_fits(scheduler: Any) -> Callable[[int], bool]:
     def fits(n: int) -> bool:
         try:
-            scheduler.preflight_or_raise(num_prompt_tokens=n)
+            scheduler.preflight_or_raise(num_prompt_tokens=n, text_only=True)
         except PrefillMemoryExceededError:
             return False
         return True
--- a/omlx/engine/base.py
+++ b/omlx/engine/base.py
@@ -71,6 +71,7 @@ async def _run_scheduler_preflight_with_cleanup_retry(
     request_id: str | None,
     eviction_callback: Any | None,
     executor: Any | None = None,
+    text_only: bool = False,
 ) -> None:
     """Run route preflight after transient post-request cleanup settles.
 
@@ -91,11 +92,13 @@ async def _run_scheduler_preflight_with_cleanup_retry(
         eviction_request = scheduler.preflight_eviction_request(
             num_prompt_tokens=num_prompt_tokens,
             request_id=request_id,
+            text_only=text_only,
         )
         if eviction_request is None:
             scheduler.preflight_or_raise(
                 num_prompt_tokens=num_prompt_tokens,
                 request_id=request_id,
+                text_only=text_only,
             )
             return
 
@@ -150,6 +153,7 @@ async def _run_scheduler_preflight_with_cleanup_retry(
         scheduler.preflight_or_raise(
             num_prompt_tokens=num_prompt_tokens,
             request_id=request_id,
+            text_only=text_only,
         )
         return
 
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -97,6 +97,7 @@ class BatchedEngine(BaseEngine):
                 "_mlx_executor",
                 None,
             ),
+            text_only=True,
         )
 
     @property
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -1454,6 +1454,7 @@ class VLMBatchedEngine(BaseEngine):
         *,
         num_prompt_tokens: int,
         request_id: str | None,
+        text_only: bool = False,
     ) -> None:
         await _run_scheduler_preflight_with_cleanup_retry(
             scheduler,
@@ -1465,6 +1466,7 @@ class VLMBatchedEngine(BaseEngine):
                 "_mlx_executor",
                 None,
             ),
+            text_only=text_only,
         )
 
     @property
@@ -3868,19 +3870,23 @@ class VLMBatchedEngine(BaseEngine):
             return
         # Count images from the ORIGINAL messages (the stripped
         # ``text_messages`` no longer has the image content-parts).
-        num_tokens += _count_image_tokens_real(
+        image_tokens = _count_image_tokens_real(
             messages,
             getattr(self, "_processor", None),
             upper_bound=_derive_image_token_upper_bound(
                 getattr(self, "_processor", None)
             ),
         )
+        num_tokens += image_tokens
         scheduler = getattr(getattr(self._engine, "engine", None), "scheduler", None)
         if scheduler is None:
             _warn_scheduler_unreachable_once(self, "preflight_chat")
             return
         await self._preflight_or_raise_with_eviction(
-            scheduler, num_prompt_tokens=num_tokens, request_id=request_id
+            scheduler,
+            num_prompt_tokens=num_tokens,
+            request_id=request_id,
+            text_only=image_tokens == 0,
         )
 
     async def preflight_completion(
@@ -3913,7 +3919,10 @@ class VLMBatchedEngine(BaseEngine):
             _warn_scheduler_unreachable_once(self, "preflight_completion")
             return
         await self._preflight_or_raise_with_eviction(
-            scheduler, num_prompt_tokens=num_tokens, request_id=request_id
+            scheduler,
+            num_prompt_tokens=num_tokens,
+            request_id=request_id,
+            text_only=True,
         )
 
     async def stream_chat(
--- a/omlx/memory_monitor.py
+++ b/omlx/memory_monitor.py
@@ -249,8 +249,8 @@ class MemoryMonitor:
         # Fixed per-sequence recurrent state (GDN/Mamba ArraysCache),
         # measured once from a live cache after the first prefill chunk.
         self._fixed_state_bytes: int = 0
-        # When True, Qwen4 QSA prefill is priced as gathered (token budget)
-        # rather than dense Q x kv_len. Set only for text-only prefills.
+        # Fallback for estimate_chunk_transient_bytes when the caller omits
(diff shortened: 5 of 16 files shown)


jundot:
Thanks for working on #3350. The Qwen4 gathered memory profile is the right direction, but I found a few paths that still need to be fixed before merge.

I reproduced the issue numbers in `_preflight_memory_check`. The gathered pricing flag is enabled only after `_do_external_prefill` starts, so preflight still uses dense pricing. With 147.56 GB already in use and 233,472 cached tokens, preflight estimated 238.15 GB against the 214 GB safety cap and rejected the request before the model forward. The same calculation was 153.19 GB with gathered pricing enabled. Please apply the same Qwen4 text-only pricing decision to preflight and chunked prefill.

The flag is also enabled for every text-only model because the current condition only checks `vlm_embeds is None`. In my non-Qwen4 test, the predicted transient dropped from 90.45 GB to 1.06 GB while this context was active. That can allow a chunk that the memory guard should stop. Please ignore dense history only when the active memory profile is the Qwen4 gathered profile.

The new Qwen4 profile also skips the measured recurrent-state cost. I set `fixed_state_bytes` to 123,456,789 bytes, but `estimate_resident_kv_bytes()` returned the same value as before. Please add the fixed-state term to the profile result so GDN and ArraysCache memory is still charged once per request.

I ran the changed Qwen4 gather and memory-monitor tests, and they passed. I also ran an adapter-level smoke test on the baseline code with a cached text prefix. The prefill received 2-D `(1, 4)` position IDs and already used gathered QSA, so the new 3-D eligibility appears to change decode behavior rather than the prefill reported in #3350. I suggest separating that change or adding a numerical parity test for the 3-D path.

Once the preflight scope, non-Qwen4 guard, and fixed-state accounting are corrected, I can recheck the original #3350 scenario.

wolfyy970:
Thanks. You were right about preflight still using dense, the flag applying to every text-only model, and the Qwen4 profile dropping fixed-state. Fixed in 61cd5585.

Preflight, eviction, and chunked prefill now use the same Qwen4 text-only gathered price as the model forward. Gathered vs dense is an argument on those paths, not a toggle of `qwen4_charge_gathered_core` on the shared monitor. If `text_only` is missing, the estimate stays dense.

Leftover dense `last_delta` is ignored only when this chunk is gathered Qwen4 and the last tagged sample was not. A gathered spike from this request still binds. Leftover dense still binds when this chunk is not gathered Qwen4.

`estimate_resident_kv_bytes` on the Qwen4 profile now includes `_fixed_state_bytes`. `estimate_prompt_kv_bytes` is unchanged.

On the 3-D eligibility: equal-plane 3-D is still gathered on prefill (with a numerical check against the official mask path). Decode 3-D gathered is out of this PR; decode stays 2-D `(1, 1)` or absent.

scubamount:
## Verified

- **Root cause correct:** two coupled bugs — (1) text-only continuation with `reuse 233K` prefix priced as dense (`last_delta×1.3` → 90 GB transient, exceeding 214 GB cap at progress 0), (2) gathered eligibility rejected `3-D` broadcast-identical mRoPE `(3,1,L)` that is logically `(1,L)` text.
- **Fix right:** broadcast-identical planes now qualify for gathered QSA (unequal planes stay mask+dense, decode stays 2-D). Scheduler charges `qwen4_exp` profile (indexer-capped core) only for text-only requests; leftover dense `last_delta`/EWMA ignored only on a gathered chunk whose last sample was not gathered. Image prefills keep dense price; `fixed_state_bytes` included on resident path.
- **Wiring:** `text_only` threaded through `preflight_or_raise` / `_run_scheduler_preflight_with_cleanup_retry` / `engine.preflight_chat|completion`; `text_only=True` for pure completions and batched preflight.

No issues. Correct cost-model split — image path unchanged.

jundot:
Thanks for the revised memory-profile work. With #3355 now on main, the refreshed merge exposes a conflict with the newer BatchQSA cache invariants; the vendor eligibility changes are redundant with the canonical rank-two binder and make the combined rank-shape test fail. I validated the Qwen4 profile and fixed-state accounting on a 233,472-token cached prefix plus a 5,244-token continuation. I'll integrate the profile while retaining the current vendor cache behavior, then land a small follow-up that separates dense and gathered transient histories.

PR #3355, opened 2026-09-01
fix(qwen4): preserve gathered prefill across mRoPE rebinds

jonathan308:
## Summary

Preserve both sides of Qwen4 text prefill correctness:

- rebind each request’s mRoPE delta immediately before every external/interleaved prefill forward, so concurrent decode or VLM cleanup cannot restart a later chunk at local position zero;
- keep scheduler-proven B1 text positions in canonical rank-two form so the already-qualified gathered-QSA prefill path remains eligible.

## Root cause

The correctness rebind caused ordinary text positions to be presented as `(3, 1, T)`: three broadcast-identical mRoPE planes. Qwen4’s gathered text-prefill gate deliberately accepts canonical `(1, T)` positions and rejects general rank-three multimodal state, so every QSA layer silently fell back to the slower official/dense route.

Real agent traffic exposed the resulting taper: roughly 862 tok/s near 4–8K fell to about 510 tok/s around 40K and 479 tok/s near 49K. The machine had one server, no memory pressure, and no thermal warning; pure model chunk timings confirmed this was not dashboard or network latency.

## Lossless behavior

`VLMModelAdapter` exposes a dedicated one-shot text-prefill binder. Only the scheduler’s non-media external/chunked prefill paths can arm it. For Qwen4 B1, `(1,T)` is exactly one plane of the former three identical planes, so rotary values are unchanged.

The capability is consumed at model-call entry, including exceptions/cancellation. Media, B2, generic models, and unproved shapes retain rank-three positions and fail closed. No weights, logits policy, sampling, token IDs, attention budget, or cache acceptance rule changes.

## Physical M3 Ultra validation

Fresh unique-salt public-API B1 runs, temperature zero, thinking disabled:

- 19,992 uncached tokens: **892.56 tok/s**.
- 219,994 uncached tokens: **885.27 tok/s overall**.
- Windowed pure model throughput: **944.69** at 0–8K, **907.60** at 40–49K, **891.63** at 98–106K, **880.05** at 147–155K, **866.86** at 196–205K, and **851.68 tok/s** at 213–220K.
- Exact cached repeat: 19,991/19,992 tokens reused, **0.15 s model TTFT**, **0.47 s visible TTFT**, identical output SHA-256.

The repaired curve matches or beats the prior qualified aggregate curve (883.06 at 50K down to 851.46 at 220K) and replaces the bad ~40% loss by 40K with about 9.8% taper across the full 220K.

## Validation

- focused regression tests: **7/7**;
- complete touched suites: **440/440** on Python 3.12 after the final CI hardening;
- wider Python 3.13 run reached **4,641 passed / 115 skipped / 77 deselected** with no failure before it was intentionally stopped;
- `git diff --check` clean;
- SSH-signed and DCO-signed-off commit;
- Python 3.11, 3.12, and 3.13 CI all green at exact head `b7e2705a`.

This PR is intentionally bounded to the adapter, scheduler rebind seams, and required tests. It contains no Cluster, cache-provider, kernel, vendor, release, or packaging work. It is ready for review: exact-head Python 3.11–3.13 CI and the final OpenCode owner-harness validation are green.

### DIFF
--- a/omlx/models/vlm.py
+++ b/omlx/models/vlm.py
@@ -67,6 +67,14 @@ class VLMModelAdapter(nn.Module):
         # from this dict + current batch UIDs before each step.
         self._uid_rope_deltas: Dict[int, float] = {}
         self._batch_rope_deltas: Optional[mx.array] = None
+        # External/chunked text prefill owns a stronger position-shape proof
+        # than the generic decode binder: the scheduler has already excluded
+        # media embeddings and is advancing one request's scalar cache. Qwen4
+        # uses that proof to keep its position ids rank two, which is exactly
+        # equivalent to three broadcast-identical mRoPE planes and preserves
+        # the qualified gathered-QSA path. Generic/media/batched binds reset
+        # the proof and retain the ordinary rank-three fail-closed path.
+        self._qwen4_text_prefill_positions = False
 
     def release_resources(self) -> None:
         """Drop references to VLM-owned MLX arrays before engine teardown reclaim."""
@@ -77,6 +85,7 @@ class VLMModelAdapter(nn.Module):
         self._pending_kwargs = {}
         self._uid_rope_deltas.clear()
         self._batch_rope_deltas = None
+        self._qwen4_text_prefill_positions = False
         self._language_model = None
         self._vlm_model = None
 
@@ -253,6 +262,7 @@ class VLMModelAdapter(nn.Module):
         self._language_model._position_ids = None
         self._language_model._rope_deltas = None
         self._batch_rope_deltas = None
+        self._qwen4_text_prefill_positions = False
 
     def register_rope_delta(self, uid: int, delta: float) -> None:
         """Register rope_delta for a UID after VLM prefill."""
@@ -269,6 +279,18 @@ class VLMModelAdapter(nn.Module):
         an array of rope_deltas aligned to the batch slot order.
         """
         self._batch_rope_deltas = deltas
+        self._qwen4_text_prefill_positions = False
+
+    def set_text_prefill_rope_delta(self, delta: float) -> None:
+        """Bind one scheduler-proven text row for an imminent prefill call.
+
+        This seam is intentionally separate from ``set_batch_rope_deltas``.
+        Only the scheduler's external/chunked non-media prefill paths may call
+        it. A later generic decode, media, or batched bind clears the proof.
+        """
+
+        self._batch_rope_deltas = mx.array([delta])
+        self._qwen4_text_prefill_positions = True
 
     def _batch_rope_deltas_for_size(self, batch_size: int) -> Optional[mx.array]:
         """Return rope deltas aligned to the current model input batch size."""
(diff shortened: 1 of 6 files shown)


jonathan308:
Final exact-head checkpoint: the mocked/minimal-request `rope_deltas` edge is fixed at signed+DCO head `b7e2705a`. The scheduler now defaults an omitted text delta to canonical `0.0`, with regression coverage for both chunked and external binders. Focused/touched validation passed (440/440 on Python 3.12; exact regression green on 3.11/3.12/3.13), and GitHub CI is now green on all three supported Python lanes. The same defensive hardening is in the validated Fusion deployment. I am keeping this draft until the final ordinary agent-harness owner check.

jonathan308:
Final owner-harness prefill gate passed on the M3 Ultra using OpenCode 1.18.25 with its normal build system/tool envelope. Four fresh serial sessions used explicit titles, zero cache reads, deterministic prompts, and exact visible completion markers.\n\n| OpenCode input | Model prefill | Scheduler-inclusive |\n|---:|---:|---:|\n| 10,022 | 900.75 tok/s | 898.53 tok/s |\n| 50,000 | 908.94 tok/s | 906.32 tok/s |\n| 100,002 | 930.09 tok/s | 927.55 tok/s |\n| 220,002 | 916.65 tok/s | 914.07 tok/s |\n\nThe continuous 220K run retained 94.51% from the first to last complete 4,096-token window (944.09 → 892.29 tok/s). Every request began at `kv_before=0`, reported `cache.read=0`, and had no prefix restore. The former 800→500-style taper did not reproduce through the real agent client. Normal cache/keepwarm settings were restored after the controlled cold ladder, and the model is loaded again with MTP depth 5.

scubamount:
## Verified

- **Root cause correct:** correctness rebind emitted `(3,1,T)` mRoPE planes (broadcast of `(1,T)`). Qwen4 gathered-QSA gate accepts only canonical `(1,T)` text positions and rejects rank-3 multimodal state → every QSA layer fell back to dense (~862→479 tok/s at 49K, not thermal/memory).
- **Fix right:** `VLMModelAdapter._qwen4_text_prefill_positions` proof — only scheduler's external/chunked non-media path may arm via `set_text_prefill_rope_delta`; generic/media/batched binds clear it. `_position_ids_from_starts` returns rank-2 `seq_positions` only when `qwen4_exp` + `B1` + proof armed; otherwise broadcasts to `(3,B,T)` — fail-closed.
- **Lifecycle sound:** proof consumed at `__call__` entry and cleared on failure so a later media/generic request cannot inherit the fast-path.

No issues. Restores gathered-QSA eligibility without weakening multimodal safety.

jundot:
Thanks for tracing this through the mRoPE rebind path. I reproduced stale decode state widening a B1 text prefill to `(3, 1, T)`; this change keeps scheduler-proven text chunks at canonical `(1, T)` while media and multi-row binds remain fail-closed. I also verified a 233,472-token cached prefix plus a 5,244-token continuation entered gathered QSA on every prefill chunk. LGTM, merging.

PR #3320, opened 2026-08-30
perf(qwen4): flatten exact long-context MTP decode

jonathan308:
## Current review status

This PR remains draft for a technical requalification gate, not for owner approval. A later low-margin workload exposed a parity failure in the prior fast wide-verifier evidence. That unqualified experimental path is not being promoted. Before this PR becomes ready, its claimed performance table must be refreshed against current main and the exact 10K-220K output-hash, cache-state, selector, prefill, and decode gates must pass.

## Why this matters

Qwen3.8-Flash-Next ordinary batch-one decode already uses the bounded direct-QSA path, but Lightning MTP target verification did not. A depth-5 cycle therefore evaluated up to six rows through the general QSA mask path. Its target-backbone cost grew from about 60 ms/cycle at 10K context to about 90 ms/cycle at 150K, reducing high-acceptance decode from roughly 83 tok/s to 59 tok/s.

This PR adds an exact, fail-closed direct-QSA verification window for native qwen4_exp. It keeps the faster reference route below the measured crossover and switches only at 65,536 tokens, so short-context performance is preserved while long-context verification remains bounded.

## What changed

- Route batch-one Qwen4 Lightning-MTP verify windows (2-9 rows) through the existing direct selected-block QSA implementation.
- Preserve target-verify projection routing for Q/K/V, indexer, and output projections.
- Qualify VLM-shaped [3,1,L] text positions once by proving all three MRoPE planes are equal over both the full cached history and current window. Genuine multimodal divergence stays on the official path.
- Retain immutable completed pooled-index blocks across speculative rollback and recompute only the affected partial suffix block.
- Add an exact BM8 QSA score tile for verify widths up to eight rows. The existing BM64 prefill tile is unchanged.
- Gate direct verification at 65,536 tokens. Below that point the current reference path is faster; above it the bounded path wins.
- Preserve aligned QSA K/V, raw index keys, index positions, and pooled state across full and partial acceptance.

## Performance

Hardware/runtime:

- Apple M3 Ultra, 256 GB unified memory, 819 GB/s
- Qwen3.8-Flash-Next-oQ4e-mtp, resident PLE
- MLX 0.32.0, mlx-vlm 0.6.3
- Batch 1, temperature 0, 500 generated tokens
- Lightning MTP fixed depth 5 for a stable geometry comparison
- Predictable code continuation, 98.3-98.8% draft acceptance
- Decode figures below exclude prefix-cache restore and suffix-prefill time and come from the MTP stage timers

[...]

PR #3356, opened 2026-09-01
perf(qwen4_exp): batch the PLE sharded-embedding gather

evgri243:
Qwen3.8-Flash-Next's SSD-backed PLE embedding gather stalled the GPU once per forward. This batches it.

**Decode: 35.4 → 37.3 tok/s at 92k (+5.4%, 95% CI [+0.13, +3.78] tok/s), paired, on `main`.**

### What was slow

The gather evaluated the id tensor and read it back with `.tolist()` — draining the GPU mid-forward — then looped per id and rescanned every shard. Ids scatter across 50–90 shards holding 1–2 rows each, so the cold mmap faults that dominate were never overlapped.

### Change

Bucket ids by shard with `searchsorted`, warm every shard in one threaded pread, dequantize once, and keep a bounded hot-row LRU. Dtype coverage matches the upstream reader, including `F8_E4M3` via `mx.from_fp8`.

Output is bitwise equal to the upstream row reader: `test_qwen4_ssd_ple_gathers_fp8_shards` compares both paths on an FP8 shard, and the existing gather tests are unchanged.

### Measurement

Paired runs on `upstream/main` + this commit, arms interleaved in one process and toggled live, 92k warm-prefix context, 600 generated tokens, M5 Max 128 GB.

| | n | off | on | diff | | 95% CI |
|---|---|---|---|---|---|---|
| PLE gather | 8 | 35.42 | 37.32 | +1.90 | **+5.4%** | [+0.13, +3.78] |

The saving is a fixed few milliseconds of host time per forward, so its share grows as the rest of the forward gets faster; on top of #3320 the same change measures +5.9%.

### Switches

`OMLX_QWEN4_PLE_GATHER=0` restores the upstream row-at-a-time path without a rebuild; `OMLX_QWEN4_PLE_HOT_MB` bounds the hot-row cache (default 1024).

Tests: same 5 pre-existing failures as `main`, none added.

Split off from the earlier version of this PR: the split-K sparse-GQA change now lives in a PR against #3320, whose verify routing is what gives it a decode effect.

### DIFF
--- a/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
+++ b/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import gc
 import json
 import math
 import mmap
@@ -7,6 +8,8 @@ import os
 import struct
 import weakref
 from bisect import bisect_right
+from collections import OrderedDict
+from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass, replace
 from pathlib import Path
 from types import SimpleNamespace
@@ -1684,6 +1687,21 @@ class _SafeTensorMMap:
             self._file = None
 
 
+_PLE_DTYPES = {
+    "BF16": (np.dtype("<u2"), mx.bfloat16),
+    "F16": (np.dtype("<f2"), mx.float16),
+    "F32": (np.dtype("<f4"), mx.float32),
+    "U32": (np.dtype("<u4"), mx.uint32),
+    "F8_E4M3": (np.dtype("u1"), None),
+}
+
+
+def _ple_batched_gather() -> bool:
+    return os.environ.get("OMLX_QWEN4_PLE_GATHER", "").strip().lower() not in (
+        "0", "false", "no", "off"
+    )
+
+
 class DiskBackedShardedEmbedding(nn.Module):
     """The 128-way dense or oQ-affine PLE table, gathered from SSD mmap."""
 
@@ -1837,7 +1855,145 @@ class DiskBackedShardedEmbedding(nn.Module):
                 group_size,
             )
 
+    # ---- fast sharded gather ------------------------------------------------
+    # The stock __call__ below evaluates the id tensor and reads it back with
+    # .tolist(), draining the GPU mid-forward, then walks one Python loop per id
+    # Bucket ids by shard with searchsorted, warm every shard in one threaded
+    # pread, dequantize once, and keep a bounded hot-row LRU.
+    # OMLX_QWEN4_PLE_GATHER=0 restores the upstream row-at-a-time path.
+    def _fast_state(self):
+        st = getattr(self, "_omlx_ple_fast", None)
+        if st is not None:
+            return st
+        io, views = [], []
+        for i in range(len(self.shard_sizes)):
+            wk, sk, bk, _bits, _group = self._shard_specs[i]
+            tio, tv = [], []
+            for key in (wk, sk, bk):
+                if key is None:
+                    tio.append(None)
+                    tv.append(None)
+                    continue
+                rd = self._tensor_readers[key]
+                entry = rd._header[key]
+                npdt, view_dt = _PLE_DTYPES.get(str(entry["dtype"]), (None, None))
+                if npdt is None:
+                    raise TypeError(
+                        f"SSD-backed Qwen4 PLE does not support {entry['dtype']}"
+                    )
+                shp = tuple(entry["shape"])
+                off = rd._data_start + entry["data_offsets"][0]
+                tv.append((np.frombuffer(rd._mapping, dtype=npdt,
+                                         count=shp[0] * shp[1],
+                                         offset=off).reshape(shp), view_dt))
+                tio.append((rd._file.fileno(), off,
+                            shp[1] * np.dtype(npdt).itemsize))
+            io.append(tio)
+            views.append(tv)
+        try:
+            hot_mb = int(os.environ.get("OMLX_QWEN4_PLE_HOT_MB", "1024"))
+        except ValueError:
+            hot_mb = 1024
+        row_bytes = max(1, sum(rb for spec in io[0] if spec for _, _, rb in [spec]))
+        st = {
+            "io": io,
+            "views": views,
+            "hot": OrderedDict(),
+            "cap": (max(0, hot_mb) * 2 ** 20) // row_bytes,
+            "pool": ThreadPoolExecutor(max_workers=16,
+                                       thread_name_prefix="ple-warm"),
+        }
+        self._omlx_ple_fast = st
+        return st
+
+    def _gather_host(self, host, shape):
+        offsets = np.asarray(self.shard_offsets, dtype=np.int64)
+        if host.size and (host.min() < 0 or host.max() >= offsets[-1]):
+            raise IndexError("embedding index is outside the sharded vocabulary")
+        uniq, inverse = np.unique(host, return_inverse=True)
+        st = self._fast_state()
+        io, views, hot, cap = st["io"], st["views"], st["hot"], st["cap"]
+
+        miss = (np.array([r for r in uniq if int(r) not in hot], dtype=np.int64)
+                if cap else uniq)
+        if miss.size:
+            shard = np.searchsorted(offsets, miss, side="right") - 1
+            local = miss - offsets[shard]
+            order = np.argsort(shard, kind="stable")
+            shard_s, local_s, miss_s = shard[order], local[order], miss[order]
+            work = []
+            for si, lr in zip(shard_s, local_s):
+                for spec in io[int(si)]:
+                    if spec is not None:
+                        fd, base, rb = spec
+                        work.append((fd, base + int(lr) * rb, rb))
+            if work:
+                step = max(1, (len(work) + 15) // 16)
+
+                def _touch(chunk):
+                    for fd, off, rb in chunk:
+                        os.pread(fd, rb, off)
+
+                list(st["pool"].map(
+                    _touch,
+                    [work[k:k + step] for k in range(0, len(work), step)]))
+            bounds = np.searchsorted(shard_s, np.arange(len(self.shard_sizes) + 1))
+            for si in np.unique(shard_s):
+                lo, hi = bounds[si], bounds[si + 1]
+                rows_l = local_s[lo:hi]
+                (wv, _wd), sv_p, bv_p = views[int(si)]
+                sv = sv_p[0] if sv_p else None
+                bv = bv_p[0] if bv_p else None
+                w = np.ascontiguousarray(wv[rows_l])
+                sc = np.ascontiguousarray(sv[rows_l]) if sv is not None else None
+                bi = np.ascontiguousarray(bv[rows_l]) if bv is not None else None
+                for j, gid in enumerate(miss_s[lo:hi]):
+                    # .copy() so a cached row owns its bytes; a view would pin
+                    # the whole fetched block and unbound OMLX_QWEN4_PLE_HOT_MB.
+                    hot[int(gid)] = (w[j].copy(),
+                                     None if sc is None else sc[j].copy(),
+                                     None if bi is None else bi[j].copy())
+        self.last_touched_shards = tuple(
+            int(x) for x in np.unique(
+                np.searchsorted(offsets, uniq, side="right") - 1)
+        )
+        rows = []
+        for r in uniq:
+            k = int(r)
+            rows.append(hot[k])
+            if cap:
+                hot.move_to_end(k)
+        while cap and len(hot) > cap:
+            hot.popitem(last=False)
+        if not cap:
+            hot.clear()
+
+        _, _, _, bits, group_size = self._shard_specs[0]
+        (_, w_view), sp, bp = views[0]
+        w = mx.array(np.stack([r[0] for r in rows]))
+        if bits is not None:
+            sc = mx.array(np.stack([r[1] for r in rows])).view(sp[1])
+            bi = mx.array(np.stack([r[2] for r in rows])).view(bp[1])
+            values = mx.dequantize(w, sc, bi, group_size=group_size, bits=bits,
+                                   mode="affine")
+        elif w_view is None:
+            values = mx.from_fp8(w, dtype=mx.bfloat16)
+        else:
+            values = w.view(w_view)
+        values = values.astype(mx.bfloat16) * self.weight_scale
+        self.rows_read = int(uniq.size)
+        out = values[mx.array(inverse.astype(np.int32))]
+        return out.reshape(*shape, self.dims)
+
+    def _call_fast(self, indices: mx.array) -> mx.array:
+        shape = indices.shape
+        flat = indices.reshape(-1)
+        mx.eval(flat)
+        return self._gather_host(np.asarray(flat, dtype=np.int64), shape)
+
     def __call__(self, indices: mx.array) -> mx.array:
+        if _ple_batched_gather():
+            return self._call_fast(indices)
         shape = indices.shape
         flat = indices.reshape(-1)
         mx.eval(flat)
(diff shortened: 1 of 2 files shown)


ivaniguarans:
Ran this on an **M5 Max 128GB** — same class of machine as your benchmark table — and I get the kernel win but **not the end-to-end win**. Posting the numbers in case they're useful, since the kernel-level result reproduces cleanly and looks correct.

**Setup.** `test/pr-3356` = this PR's two commits (`0ecac6de`) on `upstream/main` `ebf4f1fa`. mlx 0.32.2 / mlx-metal 0.32.2 / nanobind 2.15.0, all five native extensions rebuilt (`OMLX_WITH_CUSTOM_KERNEL=1`, `native_available=True` for all). Model `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp`, PLE `mmap` (69.8 GB resident of 103.94 GB), `hot_cache_only=false`, `gdn_snapshot_storage=ssd_sidecar`, `chunked_prefill=false`, `memory_guard=aggressive`. Your own tests pass here: 16/16 on `test_qwen4_qsa_split_k.py` + `test_qwen4_qsa_sparse_gqa.py`, plus 140 on the boundary-snapshot / memory-monitor / prefill-guard suites.

### 1. The kernel is ~3x faster and bitwise-exact

Isolated microbench, `q=(1,24,4,256) bf16`, `k=v=(1,2,131072,256)`, 512 selected blocks, `mx.eval` every iteration, 20 iterations:

| splits | ms/call | speedup | max abs deviation vs single-pass |
|---|---|---|---|
| single-pass | 0.785 | 1.00x | — |
| 2 | 0.935 | **0.84x** | 0.000e+00 |
| 4 | 0.350 | 2.24x | 0.000e+00 |
| 8 | 0.286 | 2.75x | 0.000e+00 |
| 16 (default) | 0.259 | 3.03x | 0.000e+00 |
| **24** | **0.243** | **3.23x** | 0.000e+00 |
| 32 | 0.254 | 3.09x | 0.000e+00 |

Two small things worth folding in: **`splits=24` beats the default 16** on this GPU, and **`splits=2` is a regression** vs single-pass — might be worth a floor in `_split_k_splits()` or a note in the docstring.

Kernel cost is flat in K length, as expected for a sparse kernel visiting a fixed 512 blocks (single-pass: 0.827 ms @ 8K → 0.771 ms @ 128K; the 4.41x at L=2048 is the outlier).

### 2. End-to-end at 128K: +0.8% decode, not +26%

`context-bench`, 128K level, `--gen-tokens 1024`, **one discarded warm-up run per arm** then 3 measured runs, **a distinct salt per run** so nothing hits the SSD prefix cache. Arms differ only by `OMLX_QSA_SPLIT_K` / `OMLX_PLE_LEGACY_GATHER`, each set in a fresh server process:

| arm | prefill tok/s (3 runs) | median | decode tok/s (3 runs) | median |
|---|---|---|---|---|
| both **off** (`SPLIT_K=0`, `PLE_LEGACY=1`) | 1048 / 1075 / 1068 | 1068 | 47.2 / 44.5 / 47.6 | 47.2 |
| both **on** (defaults) | 1063 / 1063 / 1066 | 1063 | 47.7 / 47.6 / 46.6 | 47.6 |
| delta | | −0.5% | | **+0.8%** |

### 3. Things I ruled out before concluding

- **MTP was active** during decode — tok/cycle 2.82–3.25, acceptance 92.6–94.4%.
- **The kernel is genuinely reached.** I instrumented `_native_sparse_gqa_attention` with a counter: one 128K request produced **6,500 calls into `qwen4_qsa_sparse_gqa_attention_split`, with zero shape-gate rejections.** (Instrumentation reverted afterwards.)
- **Not a key-length artifact** — see the flat sweep above.

My read on why both results hold: of those ~6,500 calls, roughly 6,072 are prefill (129,443 tokens ÷ 256-row query chunks × 12 QSA layers). Split-K's win is in filling the GPU at narrow query widths — exactly as your docstring says, "about six threadgroups at MTP verify widths" — and a 256-row prefill tile already fills it. In decode, QSA is 12 of 48 layers with MoE expert compute dominating. So a 3x on that slice doesn't move the total here.

### 4. A measurement caveat that may matter for your table

Before adding the warm-up run I measured **the same configuration twice** at 128K and got prefill 1190 vs 1307 (+9.8%) and decode 41.8 vs 49.6 (+18.7%) — both cold-prefill, no prefix reuse in the logs. The first long request after a model load appears to pay for faulting in the mmap'd PLE pages, which biases run 1 downward. After adding a discarded warm-up, spread fell to ±0.3% prefill / ±2.3% decode.

Given that, an 8-sample split-K arm could struggle to separate +21% from noise, and it's asymmetric with your 24-sample PLE arm if the runs weren't warmed. Might be worth a warm-up pass before the final numbers.

Happy to run any variant you want — different context lengths, isolated PLE-vs-split-K arms (I didn't split those out; my numbers are both-on vs both-off), or a decode-dominated shape. Also glad to be told I've measured it wrong.

evgri243:
@ivaniguarans I try to reproduce your numbers on my machine. The most confusing is 47 on the baseline. I started with barely making 29. While I try to figure it our, may I meanwhile ask you to reproduce 47.2 tk/s on the current release version instead of the current built with features disabled? Better if it is a real agent executing a real task; the reported average metrics per session should show the difference.

At least I can confirm that I see the growth from approximately 29 to 45 tk/s in the real Copilot session going from planning to execution. So the question is less in the validity of the change, -- because I see it working, -- but in the use case, benchmark construction or the reason kernel doesn't engage.

evgri243:
@ivaniguarans I've tried repeating all my measurements, all under conditions I now verify before each run — AC power, memory-enforcer ceiling checked at startup, toggles proven to flip, one discarded warm-up.

### End-to-end decode

| measurement | pairs | before → after | delta | 95% CI |
|---|---|---|---|---|
| PLE gather only | 24 | 34.23 → 36.71 | **+7.3%** | [+1.41, +3.55] |
| split-K only (3 runs) | 21 | 38.08 → 45.96 | **+20.7%** | [+6.34, +9.68] |
| **both vs neither** | 8 | **36.82 → 48.06** | **+30.5%** | [+9.78, +12.82] |
| splits 24 vs 16 | 8 | 49.98 → 49.86 | −0.2% ns | [−1.16, +1.20] |

Every pair positive in the first three (24/24, 21/21, 8/8). Compounding the two isolated effects predicts 1.298; the measured combined figure is 1.305.

**On your prompt construction** — I replicated `omlx/admin/benchmark.py::_generate_prompt` (UUID cache-buster + contiguous `code_python` slice, 122k, cold prefill every request, decode measured off the SSE stream): **+21.3%, 3/3 pairs**. So the workload shape isn't the discriminator.

**On `splits=24`**: reproduced your microbench ordering, but end-to-end it's −0.2% ns here. Keeping 16.

### Ruled out

- **`chunked_prefill`** — the one visible config difference between us. No effect on kernel engagement: 1468 verify calls with `true`, 1478 with `false`, identical widths, zero gate rejections.
- **Prompt construction** — see above.
- **Kernel engagement on my side** — instrumented by query width: 2,801 of 2,825 calls are MTP verify at 2–9 rows, 24 are prefill-width, **zero rejections**.

### What I keep coming back to

Your baseline sits where my *treated* arm sits:

| | decode |
|---|---|
| my "neither" | 36.82 |
| **my "both on"** | **48.06** |
| **your "both off"** | **47.2** |
| your "both on" | 47.6 |

So this doesn't look like split-K failing on your machine — it looks like both your arms already having it. I hit exactly that myself: my first combined A/B returned −1.3% ns because the marker never applied and both arms ran identical code. Same signature — a small non-significant delta with both arms at the treated level.

**The cheapest check uses the counter you already have:** with `OMLX_QSA_SPLIT_K=0`, calls into `qwen4_qsa_sparse_gqa_attention_split` must be **exactly zero**. If your baseline arm shows any, the arm wasn't disabled.

And to repeat the earlier ask — could you get **47.2 on a stock release build**, rather than this branch with features switched off?

ivaniguarans:
Both answered — and thanks for the counter suggestion. You were right to raise it: I had exactly that gap, and closed it a couple of hours before you asked.

### 1. Counter with `OMLX_QSA_SPLIT_K=0`

```
QSA-PROBE n=6400  env[SPLIT_K='0' PLE_LEGACY='1']  {'native_entered_splits=0': 6400}
SPLIT_KERNEL_CALLED: 0 occurrences
```
Same probe on the defaults run reports `'SPLIT_KERNEL_CALLED': 6500`. So the baseline arm was genuinely disabled — exactly zero calls, which is the test you specified. Not the explanation.

### 2. Stock release build

Unmodified brew 0.6.4 / mlx 0.32.0, no branch, no toggles. Same protocol (one discarded warm-up, 3 measured runs, distinct salt per run, `--gen-tokens 1024`, 128K):

decode **43.7 / 41.7 / 46.3** (median 43.7) · prefill 1117 / 1051 / 996 (median 1051)

That is 19% above your "neither" arm (36.82) and within 10% of your treated arm (48.06) — on release code with your patch absent. Note the release spread is much wider than on 0.32.2 (prefill ±11.5% vs ±0.3%), and the runs drifted monotonically, so treat 43.7 as directionally right rather than precise.

### 3. Real agent session, as you asked

69 requests on the patched build, both features on: 96.7% cache efficiency, **38.1 tok/s** generation, 18.2 s/request, 559 generated tokens/request. Worth flagging for both of us: real agentic traffic runs ~20% below the cold-prefill bench numbers we have been trading.

### 4. Your engagement histogram is, I think, the answer

I bucketed by query width the way you did. Same request shape as my published A/B (128K, gen 1024):

| | narrow (MTP verify) | wide (prefill tiles) |
|---|---|---|
| yours | 2,801 (99.2%), 2–9 rows | 24 |
| **mine** | **26 (0.4%)**, q_len 2–4 | **6,474 (99.6%)** — 6,461 at 256 rows |

An exact mirror image. Split-K's win is at narrow verify widths, which is 99.2% of your calls and 0.4% of mine. On this setup the sparse-GQA path is effectively a *prefill* kernel; on yours it is a *decode* kernel. Both sets of numbers are correct — we are exercising opposite phases. That also explains why replicating my prompt construction did not move it: the discriminator is not the prompt, it is which phase reaches the kernel.

**The part I cannot yet explain**, stated plainly because it matters: that run logged 130 MTP cycles (429 tokens at tok/cycle 3.30, 94.6% acceptance). 130 cycles × 12 QSA layers = **1,560 expected narrow calls. I observe 26.** So my verify-width attention is largely *bypassing* `_native_sparse_gqa_attention` rather than being rejected by it — the shape gate logs zero rejections. I have not identified which path takes it instead, and I would rather say so than guess.

If you can share how verify-width attention reaches the kernel on your side — or just your MTP draft depth and a raw `MTP[uid] ... tok/cycle= accept=` line — that would probably close it quickly. Mine on the patched build: `tok/cycle=3.30 accept=298/315 (94.6%)`, PLE `mmap`, `iogpu.wired_limit_mb=124518`, checkpoint `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp`, `chunked_prefill=false`, memory guard aggressive.

### 5. On `splits=24`

Fair correction — I suggested folding it in on the strength of a microbench alone, and your end-to-end −0.2% ns is the number that should govern. Keep 16.

### Where I land

Your +30.5% looks solid: 8/8 pairs, and the compounding check landing at 1.305 against 1.298 predicted is a good consistency signal. My earlier comment argued against the headline; with the histogram in hand I think the more accurate framing is that the gain is real and large **in the regime where QSA engagement is decode-dominated**, and roughly absent where prefill owns the kernel. If the PR body named that condition, both our datasets would be consistent with it rather than in tension.

Happy to run whatever else is useful — PLE-resident vs mmap, a decode-dominated shape, or the width histogram at other context lengths.

evgri243:
I also executed the built-in benchmark — same tool and level as yours: `POST /admin/api/bench/start`, `prompt_lengths=[131072]`, `generation_length=1024`, `code_python`. Fresh server per arm.

| | gen_tps | prefill tps | ttft |
|---|---|---|---|
| stock (`OMLX_QSA_SPLIT_K=0`, `OMLX_PLE_LEGACY_GATHER=1`) | **44.0** | 813.9 | 161.0 s |
| both fixes on (defaults) | **57.3** | 958.2 | 136.8 s |
| delta | **+13.3 (+30.2%)** | +18% | −24 s |

Our stock 44.0 vs your 47.2 is ordinary machine variance — your baseline looks legitimate. But our treated arm reaches 57.3 where yours reads 47.6, which is stock level here. So the question isn't why your baseline is high, it's **why your treated arm doesn't move.**

(This also matches the paired A/B in my previous comment: +30.5% there, +30.2% here, different tool and prompt.)

Rather than compare two similar scripts, here's the exact one I ran — needs admin auth, either `auth.skip_api_key_verification=true` in `~/.omlx/settings.json` or a session cookie.

<details>
<summary>bench_both_arms.sh</summary>

```sh
#!/bin/sh
set -e
MODEL="${MODEL:-Qwen3.8-Flash-Next-oQ4e-mtp}"
PROMPT="${PROMPT:-131072}"
GEN="${GEN:-1024}"
HOST="${HOST:-http://127.0.0.1:8000}"

run_arm() {                       # $1 = label, $2..$3 = env assignments
  echo "=== $1 ==="
  pkill -f omlx-server 2>/dev/null || true
  sleep 12                        # let the previous model finish unloading
  env $2 $3 nohup omlx serve --port 8000 > /tmp/omlx_bench_arm.log 2>&1 &
  for _ in $(seq 1 90); do
    [ "$(curl -s -o /dev/null -w '%{http_code}' "$HOST/health")" = "200" ] && break
    sleep 1
  done
  BID=$(curl -s -X POST "$HOST/admin/api/bench/start" -H 'Content-Type: application/json' \
        -d "{\"model_id\":\"$MODEL\",\"prompt_lengths\":[$PROMPT],\"generation_length\":$GEN,\"context_profile\":\"code_python\"}" \
        | python3 -c 'import sys,json;print(json.load(sys.stdin)["bench_id"])')
  for _ in $(seq 1 60); do
    R=$(curl -s "$HOST/admin/api/bench/$BID/results")
    S=$(printf '%s' "$R" | python3 -c 'import sys,json;print(json.load(sys.stdin).get("status","?"))')
    [ "$S" = "completed" ] && break
    [ "$S" = "failed" ] && { echo "  benchmark failed"; return 1; }
    sleep 20
  done
  printf '%s' "$R" | python3 -c '
import sys,json; r=json.load(sys.stdin)["results"][0]
print(f"  gen_tps={r[\"gen_tps\"]:.1f}  prefill_tps={r[\"processing_tps\"]:.1f}  ttft={r[\"ttft_ms\"]/1000:.1f}s")'
}

run_arm "STOCK (both fixes off)"   OMLX_QSA_SPLIT_K=0 OMLX_PLE_LEGACY_GATHER=1
run_arm "BOTH FIXES ON (defaults)" OMLX_QSA_SPLIT_K=  OMLX_PLE_LEGACY_GATHER=
```

</details>

evgri243:
Wait. I know the issue. Give me 10 mins to investigate.

@ivaniguarans my branch I ran experiments on for a few day is way larger. You found a critical issue: I missed a few activating commits from earlier, that's why the path, while performing as expected, doesn't work on this branch correctly.

Converting temporarily to draft. I'll update the PR door with the full stack of changes.

evgri243:
Thanks for the careful run — your numbers are right, and the disagreement was on our side.

Split-K's decode gain depends on MTP verify going through the gathered QSA path, which on `main` it doesn't: `_gathered_text_prefill_eligible` is gated with `and not target_verify`, so verify runs dense SDPA and split-K only ever touches prefill. Our branch had #3320 (@jonathan308) merged in for testing, which is where the routing came from — exactly what your dispatch histogram shows (26 narrow / 6,474 wide on `main` vs 2,801 / 24 on ours). Both measurements were correct for their bases.

This PR is now the PLE gather alone, measured on `main` (+5.4% decode, paired CI excludes zero). Split-K has moved to a PR stacked on #3320, where its effect exists (+28% paired on that base). Two other things your review surfaced are fixed as well: the old `OMLX_PLE_LEGACY_GATHER` fell back to an earlier implementation of ours rather than upstream's (the switch now falls back to the upstream path), and the bit-exactness tests that never matched the native path are gone.

evgri243:
Split into two PRs by dependency:

- **PLE sharded-embedding gather** → #3372 (against `main`, independent, +5.4% decode paired on `main`)
- **Split-K sparse-GQA attention** → jonathan308/omlx#7, stacked on @jonathan308's #3320 — its decode effect exists only with that PR's MTP verify routing, which this PR had merged in for testing and mistakenly presented as its own baseline.

Closing this one so each change is reviewed against the base it actually depends on. Thanks @ivaniguarans — your +0.8% on `main` was the correct reading and is what exposed the missing dependency. @jonathan308, the split-K PR into your branch is the follow-on to #3320; happy to rebase it whenever your branch moves.

PR #3358, opened 2026-09-01
fix: pass a portable count to macOS host_statistics64

### LINKED ISSUE #2983: psutil_compat passes buffer capacity 256 instead of HOST_VM_INFO64_COUNT 62
**Describe the bug**

`omlx/utils/psutil_compat.py` allocates 256 `integer_t` slots and also passes
`count=256` to `host_statistics64(HOST_VM_INFO64)`. On macOS 26.6.2 the public
`HOST_VM_INFO64_COUNT` is 62, so the oMLX memory-sampling hot path continuously
produces this kernel error:

```text
vm: host_statistics64() count is larger than expected
(actual:256 > HOST_VM_INFO64_COUNT:62).
This is most likely a bug in python3.11 [...] and is likely to result in memory
corruption. The buffer count must be passed in units of integer_t's.
HOST_VM_INFO64_COUNT may be used as shorthand.
```

The larger allocation is safe and useful, but the live in/out count must describe
the current structure size, not the allocation capacity. The same implementation
appears in 0.6.2, 0.6.3rc2, and current `main`.

I found this while investigating a separate long-running-server failure where
MLX could no longer reach `MTLCompilerService` (`error 32 - Broken pipe`). A full
oMLX server-process restart recovered inference. I cannot prove that the incorrect
Mach count caused the Metal XPC invalidation, so I am reporting the count defect
independently rather than claiming causation.

**To Reproduce**

1. Run oMLX 0.6.2 on macOS 26.6.2.
2. Call `omlx.utils.psutil_compat.get_macos_vm_stats()` repeatedly, or leave the
   normal oMLX memory sampler running.
3. Inspect the macOS unified log for `HOST_VM_INFO64_COUNT`.

In a bounded comparison on one machine:

- Existing implementation: 100/100 calls returned data; four coalesced kernel
  warning entries were recorded for the probe PID.
- Staged corrected implementation: 2,500/2,500 calls returned data; zero matching
  warning entries were recorded for the probe PID.

**Expected behavior**

Pass the current `HOST_VM_INFO64_COUNT` as the initial in/out count while retaining
the larger allocation. For forward compatibility, if the call returns
`MIG_ARRAY_TOO_LARGE`, use the required count returned by the kernel and retry
only if it fits the allocated buffer. Otherwise select the existing `vm_stat`
fallback.

The tested sequence is:

```python
_HOST_VM_INFO64_CURRENT_COUNT = 62
_HOST_INFO64_MAX_COUNT = 256
_MIG_ARRAY_TOO_LARGE = -307

stats = (ctypes.c_int * _HOST_INFO64_MAX_COUNT)()
count = ctypes.c_uint(_HOST_VM_INFO64_CURRENT_COUNT)
rc = _libc.host_statistics64(
    _MACH_HOST, _HOST_VM_INFO64, stats, ctypes.byref(count)
)
if rc == _MIG_ARRAY_TOO_LARGE:
    required_count = int(count.value)
    if not (_VM_STATS_MIN_COUNT <= required_count <= _HOST_INFO64_MAX_COUNT):
        return None
    count = ctypes.c_uint(required_count)
    rc = _libc.host_statistics64(
        _MACH_HOST, _HOST_VM_INFO64, stats, ctypes.byref(count)
    )
```

Additional isolated tests passed for a simulated structure growth from 62 to 104
integers and for rejecting a simulated requirement of 257 integers.

**Screenshots**

Not applicable; the exact kernel message and bounded test results are above.

- macOS Version: 26.6.2 (25G83)
- oMLX Version: 0.6.2 (2347)
- Hardware: Apple M4 Max, 64 GB

**Additional context**

The compatibility layer was introduced to address the macOS 27 layout change.
Starting with 62 and honoring `MIG_ARRAY_TOO_LARGE` preserves that goal without
violating the current kernel's count contract.

saichowdary007:
## Summary

Fixes #2983.

`get_macos_vm_stats()` allocates room for future `vm_statistics64` fields, but
previously passed that allocation capacity (`256`) as the Mach API's in/out
count. macOS validates the count against the structure layout exposed by the
interpreter's SDK, which produces warnings for both current 40-entry and
62-entry layouts.

This change:

- starts with a portable count that includes every field oMLX reads;
- keeps the larger allocation for forward-compatible tails;
- retries with the kernel-reported count for `MIG_ARRAY_TOO_LARGE` when it fits;
- falls back safely when the kernel reports an invalid or oversized requirement.

## Validation

- `pytest tests/test_psutil_compat.py -q` (6 passed)
- Ruff check and formatting checks for the changed compatibility module/tests
- Python bytecode compilation for the changed module/tests
- `git diff --check`

The full process-enforcer module could not be completed in this environment
because its optional PyTorch/transformers imports are unavailable; its focused
Mach-statistics behavior is covered by the isolated tests above.

### DIFF
--- a/omlx/utils/psutil_compat.py
+++ b/omlx/utils/psutil_compat.py
@@ -29,6 +29,11 @@ _VM_STATS_MIN_COUNT = 4
 # against `vm_stat` output on Apple Silicon.
 _VM_SPECULATIVE_INDEX = 23
 _VM_COMPRESSOR_INDEX = 32
+# Keep the initial request within both the 40-entry and 62-entry SDK layouts;
+# newer kernels can report a larger required count and trigger a safe retry.
+_HOST_INFO64_INITIAL_COUNT = _VM_COMPRESSOR_INDEX + 1
+# mach_vm.h's MIG_ARRAY_TOO_LARGE return value (KERN_INVALID_ARGUMENT - 307).
+_MIG_ARRAY_TOO_LARGE = -307
 _VM_PAGE_SIZE = 16384
 _SYSCTL = "/usr/sbin/sysctl"
 _VM_STAT = "/usr/bin/vm_stat"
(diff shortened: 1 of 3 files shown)


scubamount:
## Verified

- **Root cause correct:** `_HOST_INFO64_MAX_COUNT` (256) exceeds both the 40-entry and 62-entry SDK layouts → kernel warns. `_HOST_INFO64_INITIAL_COUNT = _VM_COMPRESSOR_INDEX + 1` covers every field oMLX reads.
- **Retry logic sound:** `_MIG_ARRAY_TOO_LARGE` (-307) triggers a bounded retry; `required_count` validated against `[_VM_STATS_MIN_COUNT, _HOST_INFO64_MAX_COUNT]` before re-call; returns `None` on invalid/oversize — no silent acceptance.
- **Tests cover the path:** `test_vm_stats_retries_with_kernel_requested_count` asserts the two-call sequence and the propagated count; existing `test_uses_layout_compatible_host_info64_count` updated for the new initial value.
- **CI green** on Python 3.11.

No issues. Good fix — clean fallback chain, well-tested.

jundot (approved):
Reducing the count addresses the warning. I'll follow up with count 40 to preserve `compressed`, tighten the tests, and trim the comments. LGTM.

PR #3359, opened 2026-09-01
Add SSD expert streaming

onthehub97:
# Add SSD expert streaming

## Summary

This PR adds an opt-in, exact-routing SSD expert-streaming runtime for supported
MoE checkpoints. It keeps a bounded expert bank resident, streams exact router
choices from safetensor shards, and leaves the existing load and execution paths
unchanged when the experimental feature is disabled.

The implementation includes:

- `soft_reap` residency, which pins manifest-selected experts and caches the rest;
- `cache_only` residency, which uses no fixed routed-expert pins;
- route-frequency and LRU cache policies;
- a fixed cold execution bank for prompt processing;
- deterministic reuse of prompt scratch slots as decode-cache capacity;
- staged and direct Apple Metal I/O, with direct publication only into proven-idle
  buffers whose on-disk and runtime representations match;
- exact native-demand callbacks during decode, including GIL-released MTP
  evaluation boundaries;
- learned hot-list persistence, resident-memory estimates, and benchmark telemetry;
- Admin API/dashboard and macOS Experimental settings, profile support, manifest
  upload/validation, and native-kernel packaging.

## Runtime design

The safetensor index discovers routed-expert tensors from checkpoint metadata. It
supports stacked and fragmented/per-expert layouts, fused or separate gate/up
projections, and affine quantization components. Cache-bank construction and the
execution path are selected from tensor shape, dtype, and quantization metadata;
there is no Q2, oQ4E, 2.4-bit, expert-count, or layer-count constant in the runtime.

Prompt processing groups cold routes into the preallocated scratch bank. Decode
can lend that bank to the dynamic cache and return it before the next prefill.
Native demand resolves the exact router output at the MLX evaluation boundary,
loads missing rows, and publishes them before expert dispatch. Direct Metal I/O is
used only when the destination is idle and needs no conversion; every other case
uses the staged, fenced path.

Installation is gated by `expert_streaming_enabled` and performed after lazy model
load but before materialization. Closing the runtime restores the original model
layers and class call path. With streaming disabled, no model class is patched and
no expert checkpoint reader or Metal I/O queue is created.

Qwen3.8 Flash Next checkpoints identify themselves as `qwen4_exp`. Their 48
backbone MoE layers use the streaming bank, while shared experts and the separate
MTP expert layer remain resident. The large PLE continues to use its existing
mmap/SSD path whenever expert streaming is enabled, so it is not materialized
alongside the expert bank. Qwen's MTP target-verification and prefill
weighted-sum patches explicitly recognize the streaming facade and retain the
exact streaming dispatch path.

## Experimental settings

| Setting | Default | Notes |
| --- | ---: | --- |
| `expert_streaming_enabled` | `false` | Master opt-in; requires model reload. |
| `expert_streaming_mode` | `soft_reap` | `soft_reap` or `cache_only`. |
| `expert_streaming_manifest` | unset | Required only for `soft_reap`. |
| `expert_streaming_cache_experts` | `32` | Dynamic slots per routed layer. |
| `expert_streaming_scratch_experts` | `32` | Prompt-processing scratch slots per layer. |
| `expert_streaming_cache_policy` | `route_frequency` | `route_frequency` or `lru`. |
| `expert_streaming_fast_resource_loading` | `true` | Apple Metal I/O transfer path. |
| `expert_streaming_direct_io` | `true` | Exact-format direct loads into idle buffers. |
| `expert_streaming_native_demand` | `true` | Exact decode demand at native evaluation. |
| `expert_streaming_decode_scratch_as_cache` | `true` | Reclassify scratch as cache during decode. |
| `expert_streaming_io_coalescing_kib` | `64` | Maximum gap merged into a storage request. |

Invalid ranges and incompatible combinations are rejected before persisted model
settings are mutated. Native demand and direct I/O require Fast Resource Loading.
Soft-REAP uploads are capped at 2 MiB, validated against checkpoint geometry, and
stored atomically under a content-derived filename.

No SSD-streaming behavior is controlled by environment variables.

## Qwen3.8 Flash Next performance

The best measured Qwen3.8 SSD-streaming run used the local Q2-MTP checkpoint and
no pinned experts:

| Item | Value |
| --- | --- |
| Checkpoint | `Qwen3.8-Flash-Next-MLX-oQ2-MTP` |
| Routed geometry | 48 layers, 512 experts/layer, top-10 |
| Residency mode | `cache_only` / zero pinned routed experts |
| Cache / scratch | 288 / 32 experts per layer |
| Decode bank | 320/512 rows after lending scratch to cache |
| Cache policy | Route frequency |
| Fast Resource Loading | All expert loads |
| Direct I/O | Enabled |
| Native demand | Enabled for decode |
| PLE | SSD mmap |
| MTP | Enabled, fixed depth 3 |
| Benchmark shape | 901 prompt tokens / 128 generated tokens |
| Loaded / peak MLX memory | 32.230 / 34.092 GiB |
| Prompt processing | **362.3060 tok/s** |
| Token generation | **40.3625 tok/s** |
| MTP acceptance | 100% (96/96 drafts) |
| Just-in-time loads | 2,553 |
| Logical expert reads | 15.496 GB |

Direct publication was the material Qwen decode improvement. Across three
matched direct-I/O runs, mean PP/TG was 363.0038/39.6166 tok/s, compared with
353.2287/32.6156 tok/s for the earlier staged-publication control: +2.77% PP and
+21.47% TG. The best direct run reached 40.3625 TG. It directly published 1,106
exact transactions (4.706 GB) into proven-idle expert-bank slots while preserving
the same 2,553 JIT loads and token trajectory, identifying the removed destination
blit as the source of the gain rather than a routing-quality change.

These measurements used a forced TerminalBench-derived trajectory, a warm-up
request, fixed MTP depth three, and 100% MTP acceptance. They are development-
machine tuning references, not portable guarantees; the three direct runs reached
40.0739, 40.3625, and 38.4133 TG tok/s as thermal and I/O conditions varied.

## DeepSeek best measured deployment configuration

The best non-oracle DeepSeek development run used this configuration:

| Item | Value |
| --- | --- |
| Checkpoint | `DeepSeek-V4-Flash-0731-2.4bit-mixed` |
| Residency mode | `cache_only` / zero pinned routed experts |
| Cache / scratch | 74 / 12 experts per layer |
| Cache policy | Route frequency |
| Fast Resource Loading | All expert loads |
| Direct I/O | Enabled |
| Native demand | Enabled for decode |
| Decode scratch reuse | Enabled |
| I/O coalescing | 64 KiB |
| Shared experts | Resident |
| Peak measured residency | 39.886 GiB |
| Prompt processing | 134.4 tok/s |
| Token generation | 13.05 tok/s |
| Just-in-time loads | 2,888 |

These figures are a tuning reference from the development machine, not a portable
performance guarantee: prompt shape, generation length, thermal state, SSD, MTP
acceptance, and workload all affect the result. A longer warm run reached 162.42
tok/s PP and 12.52 tok/s TG, illustrating why PP and TG should be compared on the
same benchmark shape. Cache and scratch are deployment budgets rather than model
constants; 74/12 should not be made a universal default.

## Recommended weight format

For the DeepSeek mixed 2.4-bit checkpoint:

- keep routed expert weights in their existing U32-packed 2.4-bit representation;
- store routed-expert affine `scales` and `biases` as FP16 so storage matches the
  FP16 expert bank and the direct path is legal;
- do not expand packed expert weights to dense FP16.

BF16 affine metadata remains supported through staged BF16-to-FP16 conversion, but
that conversion prevents direct publication. The included conversion tool is
resumable and atomically replaces each shard:

```bash
# Inventory only
python scripts/convert_deepseek_v4_streaming_layout.py /path/to/model

# Convert in place; use a checkpoint copy unless this is intentional
python scripts/convert_deepseek_v4_streaming_layout.py /path/to/model --apply
```

Other quantizations do not require this conversion when their stored expert rows
already match the runtime bank. The reader falls back to staged loading whenever a
layout or dtype is not direct-compatible.

## Metrics

Admin benchmark snapshots now expose cache hits/misses, evictions, exact JIT loads,
SSD bytes and operations, I/O and materialization timings, direct-load counts,
native-demand calls/callbacks, route materializations, and elastic-cache
transitions. `benchmarks/soft_reap_streaming.py` validates row exactness and can run
end-to-end generation; `benchmarks/fast_resource_loading_micro.py` compares CPU
staging with Metal I/O.

## Validation

- Rebuilt all native custom kernels; the Fast Resource Loading ABI probe reports
  available.
- 502 focused Python tests passed across expert streaming, Admin settings,
  residency/signatures, benchmark telemetry, MTP, Qwen4 PLE loading, and Qwen MoE
  dispatch.
- The production Qwen3.8 Q2-MTP checkpoint metadata was audited in place: all 48
  backbone MoE layers were indexed at 512 experts/top-10, and the separate MTP
  expert layer was excluded from the streamed set as intended.
- 19/19 focused macOS `ModelSettingsScreenVMTests` passed.
- The complete macOS test target compiled and ran 303 tests: 301 passed, one
  integration test was skipped, and one unrelated existing Theme color assertion
  failed (`ThemeTests.testLightWindowBackgroundUsesStandardWindowColor`).
- `ruff` passes for the new expert-streaming modules, benchmarks, conversion tool,
  and streaming tests; `git diff --check` passes.

## Review notes

- The feature is experimental and disabled by default.
- Fast Resource Loading is packaged in the Python custom-kernel build, Homebrew
  kernel list, and macOS app kernel build.
- Enabling direct/native loading without the native extension fails closed with a
  clear load error; it does not silently run an unsafe publication path.
- Resident-memory estimation is deliberately conservative and includes a 5%
  overhead allowance.
- Full design and operational details are in `docs/soft_reap.md`.


beaglemoo:
Tested this branch on an M5 Pro with 64 GB (macOS, mlx 0.32.2) against Qwen3.8 Flash Next oQ2 MTP (66 GiB), DeepSeek V4 Flash 2.4bit (86 GiB), and Qwen3.6-35B-A3B as an eager exactness reference. Both large models run through `omlx serve` after fixes; write-up and patches are in a PR against the source branch: https://github.com/onthehub97/omlx/pull/1

The two important findings:

1. Native demand deadlocks on any `mx.synchronize()` call. `mx.eval` releases the GIL while waiting on Metal but `mx.synchronize` does not, and the miss-path completion handler needs the GIL. mlx-vlm's `wired_limit` exit and the engine pool's unload path both hit it; the branch's own `benchmarks/soft_reap_streaming.py --full-load` hangs on Qwen3.6-35B-A3B. The fix adds a GIL-releasing synchronize to the extension and swaps `mlx.core.synchronize` while a native-demand runtime is installed.

2. The native-demand all-resident fast path reads a slot map captured when the decode graph was built, but mlx-lm pipelines decode one step ahead, so it can name a slot already refilled by the previous step's callback. On Qwen3.6-35B-A3B at temperature 0, native demand off is byte-identical to an eager load; native demand on diverged at a different position every run. Resolving every route through the callback makes it exact at the same throughput.

Also: the residency estimate counts the Qwen4 PLE that streaming forces onto mmap (52.6 GB estimated vs 32.7 GB measured, which refuses admission on a 64 GB machine), and the VLM engine only loads lazily for `qwen4_exp`, so other MoE VLMs end up with banks on top of a fully resident model.

Measured after the fixes: Qwen3.8 at 32.7 GB resident, 25 to 28 tok/s warm with MTP depth 3; DeepSeek V4 Flash at 30.5 GB at load, 10.4 to 11.7 tok/s on the staged path (BF16 scales), greedy output identical across runs.

onthehub97:
@beaglemoo Thanks for that PR with the fixes, I've merged it.

beaglemoo:
Thanks for merging. One follow-up against your branch: https://github.com/onthehub97/omlx/pull/2. It removes a bimodal streamed prefill on Qwen4-Exp (the mmapped PLE gather faulted about 16k random rows synchronously per 1024-token prompt; now prefaulted in parallel, 3.6 to 3.9 s across seven runs instead of 3.6 to 10 s), replaces the polling Metal IO wait from #1 with a completion semaphore, and pools the staging buffers. It also records a measurement that may matter for the design: the native-demand route turnaround is about 150 us per layer with every expert resident (47 us GPU-to-CPU, 106 us CPU-to-GPU), and a GPU-side wait is not achievable on Apple silicon, so exact on-demand routing has a floor of roughly 7 ms per decode step on a 48-layer model. Details in the PR and docs/soft_reap.md.

onthehub97:
Closing as MoE streaming has been added to upstream. Will do an Astra pass later and see if there’s anything from this that could be used in upstream implementation.

PR #3327, opened 2026-08-31
perf(cache): remove duplicate restore work and expose exact stream TTFT

jonathan308:
## Summary

This PR isolates the cache/TTFT and Qwen tool-streaming fixes from the Qwen long-context MTP work in #3320.

## What it fixes

- Removes the serialized `preload_blocks` pass that duplicated the now-inference-thread-safe prefix reconstruction/load path.
- Adds phase telemetry so cache lookup, cache reconstruction, model-first-token TTFT, and first-visible SSE TTFT are reported separately.
- Makes Qwen3 Coder tool-envelope streaming safe with exact parser capability gating, FIFO content/tool ordering, registered-name and JSON-object validation, bounded malformed handling, semantic deduplication, and authoritative scheduler structured calls.
- Keeps early Qwen validation capability-scoped so unsupported engines and ordinary streams preserve their existing behavior.

## Scope

Per maintainer review, unfinished-reasoning output handling, EOS masking, and positioned-MTP processor contract failures have been removed from this PR and will be submitted as separate review units. This PR retains commits `b55822de`, `573fb31c`, and `fb1240ad`; requested reverts were applied in order as `af12b334` and `6298990e`.

No model weights, sampling behavior, reasoning protocol, cache token identity, or valid tool-call semantics are changed here.

## Validation

Focused Python 3.13 validation after the scope split: **963 passed, 12 deselected** across Qwen tool calling, end-to-end streaming, scheduler, prefix cache, batched engine, and VLM engine tests. Hosted Python 3.11–3.13 CI is rerunning on the updated branch.

PR #3360, opened 2026-09-01
fix(stream): keep unfinished reasoning private

jonathan308:
## Summary

Reasoning models can occasionally sample a terminal token before emitting their closing reasoning marker. oMLX previously treated that malformed but successful-looking stop as a normal completion and copied the accumulated private reasoning into visible assistant content.

This change makes the protocol fail closed and prevents the malformed stop at its source:

- mask only model-owned terminal tokens while a prompt-opened reasoning block remains active
- immediately restore ordinary EOS behavior after the natural close marker
- preserve explicit caller stop tokens and overlapping close markers
- retain processor state across speculative MTP snapshot/rollback
- never promote unfinished reasoning into visible content
- report malformed OpenAI, Anthropic, and Responses streams as incomplete

The scope is protocol handling only. It does not change cache contents, tool parsing, request-specific stop conditions, model weights, or accepted-token accounting.

## Regression coverage

- EOS blocked before close and available after close
- no-budget and budgeted reasoning paths
- speculative snapshot/restore
- caller stop-token and close-marker isolation
- plain non-reasoning requests remain unchanged
- OpenAI SSE reproduction of the observed unfinished internal monologue
- scheduler, cancellation, MTP, proxy, prompt-priming, and server endpoint suites

Validation on upstream main: **557 passed, 3 deselected**.

### DIFF
--- a/omlx/api/thinking.py
+++ b/omlx/api/thinking.py
@@ -152,15 +152,13 @@ def extract_thinking(text: str) -> Tuple[str, str]:
     - Empty think: ``<think></think>answer`` → ``("", "answer")``
     - Think only: ``<think>reasoning</think>`` → ``("reasoning", "")``
     - Malformed (open with no close): ``<think>everything…`` →
-      ``("", "everything…")`` — recovery for V4-style models that
-      occasionally skip the ``</think>`` boundary token. Without this
-      fallback the entire body would be classified as thinking and the
-      visible answer would be empty.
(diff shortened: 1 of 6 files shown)


jonathan308:
Folded into the existing related stream-integrity/cache PR #3327 as requested. The complete fix, including Responses finalization, malformed tool-call suppression, and fail-closed MTP processor handling, now lives there.

PR #3364, opened 2026-09-01
fix(memory): refresh ANE prefill transient reserve after bank compile

Cianidos:
## Problem

#2841: with ANE prefill enabled, the compiled slices hold fixed-shape fp16 I/O
surfaces that the first long prompt dirties on top of the KV+SDPA peak. `dcb317fe`
priced those surfaces through an `ane_prefill_transient_bytes` term passed to
`set_model_info`. That term is snapshotted **at scheduler init**, but on the VLM
path the ANE banks are compiled **after** the scheduler is initialized —
`VLMBatchedEngine.start()` runs `enable_qwen35_ane_prefill` only after
`engine.start()` has constructed the scheduler. The load-time read therefore
returns 0 and the reserve never engages.

Production capture (v0.6.4, ANE prefill enabled, Qwen3.8-27B-oQ4e-mtp):

```
qwen35_ane_prefill - Warmed 224 ANE procedures in 2.0s at load
engine_pool        - Loaded model: Qwen3.8-27B-oQ4e-mtp (actual: 28.49GB, local estimate: 16.60GB)
scheduler          - [guard:external] admission terms: current=32.41GB predicted_transient=3.87GB observed_max_bytes=0.00GB ane_prefill_transient_bytes=0.00GB
context_benchmark  - ERROR - Not enough memory to prefill even 4,096 tokens on this machine
```

224 procedures are warm and the banks are resident (~11.9 GB over the GPU-only
footprint), yet the admission breakdown prints `ane_prefill_transient_bytes=0.00GB`
and the model cannot prefill even 4,096 tokens. This matches the post-`dcb317fe`
report on #2841 ("still seeing the abort, disabling ANE prefill fixes it"). The LM
path is unaffected (its compile runs before the scheduler is created); the VLM path
is the one that loads Qwen multimodal checkpoints.

## Change

- **`MemoryMonitor.set_ane_prefill_transient_bytes(value)`**: idempotent setter,
  clamps negative to 0.
- **`BatchedEngine.start()` / `VLMBatchedEngine.start()`**: refresh the reserve
  immediately after `enable_qwen35_ane_prefill` compiles the banks, computed from
  the live model dims via the existing `ane_prefill_transient_bytes(model)`. The
  release rung already zeroes it through `clear_ane_prefill_transient()` (#3103),
  so the term is non-zero exactly while the banks are resident.

## Verification

- New unit test: reserve defaults to 0, the setter raises it,
  `estimate_prefill_peak_bytes` includes it, clear drops it, negative input clamps.
  `tests/test_memory_monitor.py` + `tests/test_qwen35_ane_prefill.py`: **216 passed**.
  Broader: `test_scheduler_prefill_memory_guard.py` + `test_memory_monitor_vlm_config.py`
  + `test_batched_engine.py`: **104 passed**.
- ANE compiles and engages on a source build (224 procedures warmed). With ANE
  enabled, the long-context benchmark now reaches the bank-release rung and completes
  65,536 tokens where the v0.6.4 capture above rejected at 4,096 — the full
  before/after ladder trace is in the eviction-escalation PR.

### DIFF
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -579,6 +579,7 @@ class BatchedEngine(BaseEngine):
         scheduler = self._engine.engine.scheduler
         if ane_prefill_sequence_length:
             from ..patches.qwen35_ane_prefill import (
+                ane_prefill_transient_bytes,
                 configure_qwen35_ane_prefill_scheduler,
             )
 
@@ -586,6 +587,15 @@ class BatchedEngine(BaseEngine):
                 scheduler,
                 ane_prefill_sequence_length,
             )
+            # Price the compiled ANE I/O surfaces explicitly: a later
+            # _set_model_info_for_monitor() defaults the term to zero
+            # unless re-derived, and the monitor must hold the real
+            # reservation while the banks exist.
+            monitor = getattr(scheduler, "memory_monitor", None)
+            if monitor is not None:
+                monitor.set_ane_prefill_transient_bytes(
+                    ane_prefill_transient_bytes(self._model)
+                )
         if self._model_settings is not None:
             tq_enabled = getattr(self._model_settings, "turboquant_kv_enabled", False)
             if tq_enabled:
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -2105,6 +2105,18 @@ class VLMBatchedEngine(BaseEngine):
                         scheduler,
                         requested_ane_sequence_length,
                     )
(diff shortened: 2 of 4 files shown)


jundot:
Thanks for fixing the reserve refresh. I verified on a real VLM load that the monitor reported zero despite 6.28 GiB of compiled ANE I/O surfaces. With the patch applied, the reservation matched those surfaces and returned to zero after release. This looks good to merge.

PR #3365, opened 2026-09-01
fix(engine_pool): escalate prefill headroom past a no-op first pass

Cianidos:
## Problem

#3103 added an ANE-bank-release rung, made *recurring* pressure escalate
release-before-reclaim, and raised the retry budget to 2. But a request is marked
recurring only when the pooled-buffer reclaim rung actually runs. There is a path
where the first eviction callback runs **no rung at all**: the pool's own
re-measurement — `max(mx.get_active_memory(), phys_footprint, model_memory)` —
reads lower than the scheduler's `_current_usage_bytes()`, sees headroom, and
returns having done nothing and marked nothing.

The scheduler's tighter reading still shows pressure, so it spends its retry. The
second callback is still *not* recurring, so it runs the marginal pooled reclaim —
which prefill continuously refutes by refilling MLX's buffer cache — and the budget
is exhausted. The durable release rung is never reached.

Production capture (v0.6.4, request `4420aa8c`, ANE prefill enabled):

```
scheduler   - Request 4420aa8c needs prefill headroom before throttling (current=32.44GB, predicted=1.36GB, target=33.29GB)
engine_core - No idle model evicted for request 4420aa8c; scheduler will fall back to throttling
scheduler   - Request 4420aa8c needs prefill headroom before throttling (current=32.39GB, predicted=2.23GB, target=34.41GB)
engine_pool - Reclaimed 2.60GB of pooled Metal buffers for prefill request 4420aa8c (no idle model to evict)
```

Pass 1 ran no rung ("No idle model evicted ... fall back to throttling", with no
reclaim line); pass 2 reclaimed; retries exhausted. The request was then rejected at
the 2048-token guard — `Not enough memory to prefill even 4,096 tokens` — while the
~12 GB of bank memory sat untouched.

## Change

- `_evict_idle_lru_for_prefill` marks the request at **entry**, storing a per-request
  callback attempt counter in `_prefill_headroom_recurring` (bounded FIFO).
  `recurring = attempt > 1` is snapshotted so marking inside the current call does
  not flip it mid-call. A no-op first pass now leaves the request recurring, so the
  retry escalates to the release rung instead of reclaiming again. The two-retry
  policy and rung ordering are unchanged.
- Adds one stable `[prefill-eviction]` INFO record per callback decision:
  `request / retry / reason / action / outcome / recurring / mx_active /
  phys_footprint / model_memory / predicted / target / evicted`, where
  `action ∈ {already_fit, evict_model, reclaim_pool, release_ane, reject}`. Within a
  single pass the ladder still runs reclaim→release when reclaim alone is not enough;
  the record reports the highest rung reached. The decision was previously invisible.

## Verification

- New test reproduces the observed sequence exactly: first callback `already_fit`
  no-op → second releases the ANE banks; asserts reclaim is never attempted, no model
  is evicted, and both decision records read `action=already_fit retry=1` then
  `action=release_ane retry=2`. The existing masked-delta test is updated for the
  extra re-measurement taken by decision logging.
  `tests/test_engine_pool.py` + `tests/test_prefill_oom_graceful.py`: **191 passed**.
  Broader: `test_cluster_engine_pool.py` + `test_scheduler_prefill_memory_guard.py`
  + `test_server_prefill_memory_handler.py`: **65 passed**.
- Live (source build, ANE prefill enabled): the decision is now logged and the ladder
  reaches the release rung under pressure:

```
engine_pool - [prefill-eviction] request=e417db5c retry=1 reason=adaptive_prefill_throttle action=release_ane outcome=retry recurring=False mx_active=16.57GB phys_footprint=20.25GB model_memory=16.60GB predicted=3.39GB target=33.70GB evicted=0
engine_pool - Released ANE prefill banks on 'Qwen3.8-27B-oQ4e-mtp' ... freed 10.29GB
```

  With the pressure now recoverable, the 65,536-token context benchmark completes
  (~397 tok/s) where the v0.6.4 capture above rejected at 4,096 tokens.

### DIFF
--- a/omlx/engine_pool.py
+++ b/omlx/engine_pool.py
@@ -293,8 +293,9 @@ class EnginePool:
         # rung can "succeed" marginally on every pass of a long prompt while
         # the durable rung behind it (ANE bank release) is never reached; a
         # request coming back for more headroom escalates instead of
-        # reclaiming again first. Bounded FIFO — request ids are transient.
-        self._prefill_headroom_recurring: OrderedDict[str, None] = OrderedDict()
+        # reclaiming again first. Values count callback attempts per request
+        # (>1 = recurring). Bounded FIFO — request ids are transient.
+        self._prefill_headroom_recurring: OrderedDict[str, int] = OrderedDict()
         self._load_seconds_per_gb_ema: float | None = None
         self._load_time_observations: int = 0
         self._lease_release_tasks: set[asyncio.Task[None]] = set()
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for covering the no-op first callback. I verified with the regression test that a subsequent callback now prioritizes ANE release while preserving the eviction order and retry budget. The combined patches also passed the real-model checks, though that specific no-op sequence did not occur in those runs. LGTM; I'll merge this after #3340 and #3364.

Issue #3294, opened 2026-08-29
Qwen4-Exp QSA cache: prefix blocks unusable across text/MRoPE turns, plus three related defects


tbro0815:
Four defects in Qwen4-exp QSA cache handling, found while running `Qwen3.8-Flash-Next-oQ4e-mtp` on oMLX **v0.6.3**. All are in the vendored Qwen4-exp support added between `v0.6.3rc3` and `v0.6.3`.

**Environment:** macOS 26, M5 Max 128 GB. The model is a 106 GB checkpoint running in ~69 GB resident, n-gram/PLE weights streamed from SSD (`qwen4_ple_ssd_offload`).

Items 1 and 2 I have fixed and tested; patches offered as a PR if wanted. Items 3 and 4 are reported without patches because I could not exercise them.

#3293 covers the sparse-attention indexer breaking under multi-row batching, which is unrelated but shares the `batch >= 2` precondition.

---

## 1. Prefix-cache blocks cannot be rejoined across text/MRoPE turns

**Every turn re-prefills its entire prompt.**

```
omlx.cache.prefix_cache - WARNING - Failed to reconstruct cache: [concatenate]
... the provided shapes are (1,1,2048), ... (1,3,2048), ... axis is 2

prefix cache: request X re-prefills 53227 of 53227 tokens (reused 0);
closest stored sequence Y shares the first 51200 of 51200 comparable tokens
```

`_serialize_qsa_positions` (`omlx/cache/type_handlers.py`) pins the *rank* of stored positions at `[B, C, S]` but not `C`: a block stored under plain text positions has `C == 1`, one stored under MRoPE has `C == 3`. `Qwen4QSAKVCacheHandler.concatenate_states` joins blocks on the sequence axis without reconciling `C`, so any sequence spanning both layouts throws on reconstruct.

The cost is worse than a plain cache miss: snapshots keep being written (~1 s per turn here) for data that can never be read back, on top of a full re-prefill - ~85 s at 55k tokens.

The model already has the reconciliation rule. `_append_indexer_positions` broadcasts the 2-D side up before its own concatenate; the handler needs the same, widening `C == 1` to the widest channel count before joining. Widening replicates the text position across the MRoPE channels, matching runtime behaviour. It should only ever widen - a combination with no defined widening should be left alone so the concatenate raises rather than inventing channels.

`Qwen4QSAQuantizedKVCacheHandler` inherits `concatenate_states` and is fixed by the same change.

Two adjacent defects in the same handler:

[...]

PR #3368, opened 2026-09-02
fix(qwen4_exp): per-row past_len in QSA indexer — unlocks multi-row batching (#3293)

### LINKED ISSUE #3293: Qwen4-Exp: sparse-attention indexer breaks under multi-row batching (batch >= 2 always fails)

astro-15-alive:
# fix(qwen4_exp): per-row past_len in QSA indexer — unlocks multi-row batching

Closes #3293

## The bug

`Indexer.from_projected` reads `past_len = cache.offset`, which is a plain `int`
for `QSAKVCache` but an `mx.array` of per-row offsets (shape `[batch]`) for
`BatchQSAKVCache`. Treated as a scalar, `query_ends` and `complete_counts` come
out shaped `[batch]` instead of `[seq_len]`; indexing them `[None, :, None]` puts
the **batch** axis where `seq_len` belongs, `mx.where` broadcasts scores to
`(batch, batch, blocks)`, and the padding built from the literal `seq_len` then
raises:

```
ValueError: [concatenate] All the input array dimensions must match exactly ...
shapes are (2,2,55996), (2,1,1), concatenation axis is -1
```

Invisible at `batch == 1` — which is why it survived until continuous batching
actually formed a multi-row batch. Same axis confusion in `tail_starts`,
`causal`, `use_sparse`, and `_default_position_ids` (array `start` into
`mx.arange`).

## The fix

Co-authored with @tbro0815, who diagnosed it in #3293 and validated the fix in
their fork (tbro0815/omlx@9402157). `past_len` is normalized to a per-row column
and indexed on `[..., None]`, building a real `[batch, seq_len]` grid. The fix
preserves batch-1 numerics exactly — identical results on single-row paths.

This also addresses the first defect class in #3294 (multi-row cache join path).

## Validation

- Cherry-picked onto current main (`ebf4f1f`) — applies cleanly, vendored file
  auto-merges without conflict.
- 14-test suite `tests/test_qsa_indexer_batching.py` (from Thomas) passes:
  `pytest tests/test_qsa_indexer_batching.py -q` → **14 passed**.
- Tests assert *selected-mask contents per row*, not just shapes — a shape-only
  fix that silently attends to the wrong blocks would fail these.
- Verified independently on Apple Silicon (mlx 0.29.3 / pytest 8).

## Why this matters upstream

Any deployment running Qwen4-exp models with `max_concurrent_requests >= 2`
currently dies at batch formation and must clamp to 1 — sacrificing aggregate
throughput as a workaround. This patch (plus @moc375's #3265/#3334 batching work)
is the unlock for real continuous batching on qwen4_exp.

### DIFF
--- a/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
+++ b/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
@@ -972,11 +972,25 @@ class Qwen4ExpQSAIndexer(nn.Module):
         scores = mx.sum(mx.maximum(scores, 0), axis=1)
         scores = scores / math.sqrt(self.head_dim)
 
-        query_ends = past_len + mx.arange(seq_len) + 1
+        # past_len is an int for QSAKVCache but an mx.array of per-row offsets
+        # for BatchQSAKVCache (whose `offset` is a property over
+        # kv_cache.offset). Treating it as a scalar gave query_ends and
+        # complete_counts shape [batch] rather than [seq_len], so
+        # complete_counts[None, :, None] put the batch axis where seq_len
+        # belongs and mx.where broadcast scores from (batch, seq_len, blocks)
+        # to (batch, batch, blocks). Downstream that met padding built from
+        # the literal seq_len and raised:
+        #
+        #   [concatenate] ... shapes are (2,2,55996), (2,1,1) ... axis is -1
+        #
+        # Invisible at batch == 1, where the two axes coincide. Build a
+        # [batch, seq_len] grid instead and index it on [..., None].
+        past = mx.array(past_len).reshape(-1)
+        query_ends = past[:, None] + mx.arange(seq_len)[None, :] + 1
         complete_counts = query_ends // self.compress_ratio
         valid_blocks = (
             mx.arange(max_complete_blocks)[None, None, :]
-            < complete_counts[None, :, None]
+            < complete_counts[..., None]
         )
         scores = mx.where(valid_blocks, scores, -mx.inf)
         selected_blocks = mx.argpartition(scores, kth=-self.block_topk, axis=-1)[
@@ -1005,16 +1019,17 @@ class Qwen4ExpQSAIndexer(nn.Module):
                 axis=-1,
             )
 
+        # complete_counts and query_ends are [batch, seq_len] now, so these
+        # index on [..., None] rather than [None, :, None] -- the same axis
+        # confusion as above, and equally invisible at batch == 1.
         token_indices = mx.arange(key_len)
         tail_starts = complete_counts * self.compress_ratio
-        tail = (token_indices[None, None, :] >= tail_starts[None, :, None]) & (
-            token_indices[None, None, :] < query_ends[None, :, None]
+        tail = (token_indices[None, None, :] >= tail_starts[..., None]) & (
+            token_indices[None, None, :] < query_ends[..., None]
         )
-        causal = token_indices[None, None, :] < query_ends[None, :, None]
+        causal = token_indices[None, None, :] < query_ends[..., None]
         use_sparse = complete_counts > self.block_topk
-        selected_tokens = mx.where(
-            use_sparse[None, :, None], selected_tokens | tail, causal
-        )
+        selected_tokens = mx.where(use_sparse[..., None], selected_tokens | tail, causal)
         return selected_tokens[:, None]
 
 
--- a/tests/test_qsa_indexer_batching.py
+++ b/tests/test_qsa_indexer_batching.py
@@ -1,0 +1,257 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Qwen4-Exp sparse-attention indexer under multi-row batching.
+
+`Indexer.from_projected` read ``past_len = cache.offset``, which is a plain
+``int`` for ``QSAKVCache`` but an ``mx.array`` of per-row offsets for
+``BatchQSAKVCache``. Treating it as scalar gave ``query_ends`` and
+``complete_counts`` shape ``[batch]`` instead of ``[seq_len]``, so indexing
+them ``[None, :, None]`` put the batch axis where seq_len belongs and
+``mx.where`` broadcast the scores from ``(batch, seq_len, blocks)`` to
+``(batch, batch, blocks)``. Padding built from the literal ``seq_len`` then
+failed:
+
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for the detailed report and tests. #3246 already converts a batched QSA offset to the aligned physical index width before this new grid is built, so on current main this branch receives a scalar and produces the same production mask. I verified the real ragged `BatchQSAKVCache` → `Qwen4ExpQSAIndexer.from_projected()` path on both revisions, and the 14 added tests also pass unchanged on main because they exercise a local `_mask()` copy rather than the production method. The cache-state slicing is already present from #3246 as well, so this is superseded and I’m closing it without merging.

moc375:
Thanks for the reference! For context on the full bug chain:

1. **Crash layer** — our #3263 fixed the `QSAKVCache` missing `to_batch` / `extend` path that killed all concurrent streams (100% failure at batch ≥ 2)
2. **Shape layer** — our #3264 (and independently tbro0815's fork, astro-15-alive's #3368, and upstream #3246) fixed the per-row `past_len` axis confusion in the QSA indexer
3. **Batching layer** — our #3265 adds depth-1 always-advance batched MTP for multi-row decode, and #3334 adds mx.compile batched decode (dispatch -77%)

Your PR description calling out the combination is spot-on — the crash fix + shape fix + compile lane together unlock real continuous batching on qwen4_exp. We've been running this stack in production on M3 Ultra with Qwen3.8-Flash-Next and it's stable.

Happy to see the community converging on this — the bug chain was non-trivial to isolate across three separate layers.

astro-15-alive:
dev2 datapoint from a 256 GB M3 Ultra (Mac Studio), `Jundot-Qwen3.8-Flash-Next-oQ4e-mtp`, Lightning MTP depth 6, `max_concurrent_requests=1` (still clamped for the QSA multi-row crash):

- dev1 → dev2 aggregate incl-prefill: 74.5 → 75.1 tok/s (flat, no regression)
- Per-probe dev2: sequence 73.7, code 70.0, structured 100.0, prose 68.8 tok/s
- Clean run post-restart: zero `QSAKVCache` / singleton-recovery / ERROR events

FYI for anyone porting: streaming `/v1/chat/completions` no longer includes `usage` unless `stream_options.include_usage` is set — older bench scripts silently report 0 tokens on dev2.

(_posting here since release comments 404 for non-collaborators_)

PR #3369, opened 2026-09-02
fix(qwen4_exp): promote rank and separate lengths in BatchQSAKVCache joins (#3294 items 2+4)

astro-15-alive:
# fix(qwen4_exp): promote rank and separate lengths in BatchQSAKVCache joins

Follow-up to #3294 — closes items 2 and 4. (#3219 landed item 1's reconstruct path, and the batch `trim` fix landed item 3; these are the **runtime join** twins that survived both.)

## The bugs

**Item 2 — `extend` / `merge` join text and MRoPE rows on the wrong axis.**
Both pick `sample_positions` by *first non-None* and derive `position_axis` from its rank. Joining a text-only row (2-D `[B, S]`) with an image row (3-D `[3, B, S]`) crashes at concatenate:

```
ValueError: [concatenate] All the input arrays must have the same number of
dimensions. However, got arrays with dimensions 2 and 3.
```

Promotion only widens, so the sample must be the **widest** operand available — `_pad_index` already knows how to broadcast 2-D up, it just never receives the 3-D sample to widen against. Order-sensitive: which side blows up depends on which cache is `self`.

**Item 4 — `merge` conflates KV offset with indexer length.**

```python
target = max(cache.offset for cache in caches)
...
SimpleNamespace(..., index_offset=cache.offset, offset=mx.array([cache.offset]))
```

`cache.offset` is an `int` for `QSAKVCache` inputs but an `mx.array` of per-row offsets for `BatchQSAKVCache` inputs — so `max(...)` and `mx.array([cache.offset])` are ill-defined for the latter (2-D array slices → `Slice indices must be integers or None`, reproduced). And when KV length and indexer length diverge (a trimmed or draft-ahead indexer — item 3 was exactly such a mechanism), passing the KV offset as `index_offset` silently mis-slices the join.

## The fix

- `extend`/`merge` select the **widest** position rank across operands, and `_pad_index` broadcasts 2-D rows up before padding — the same rule the runtime update path applies in `_append_indexer_positions`.
- `merge` derives row lengths from the **indexer state** (`index_offset` for batch caches, the already-sliced `index_keys` property for singletons), never from `cache.offset`.

Homogeneous-rank, consistent-offset joins are bit-identical to today: both changes only fire where the old code crashed or silently mis-sliced.

## Validation

- Repros **fail on current main** (`ebf4f1f`) with exactly the signatures above; **pass after the fix** — 5 tests, tiny synthetic tensors, no model load.
- No regression in the existing suites: `test_mlx_vlm_qwen4_exp_compat.py` **47 passed**; plus the #3293 indexer suite (`test_qsa_indexer_batching.py`, 14 tests from #3368) green on its branch.
- Verified on Apple Silicon, mlx 0.32.x, Python 3.11.

Diagnosis and defect taxonomy are @tbro0815's from #3294 — this PR delivers the patch they offered for items 1/2's join-side remnants and item 4's offset typing, re-based against current main.

### DIFF
--- a/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
+++ b/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
@@ -575,6 +575,8 @@ class BatchQSAKVCache:
     @staticmethod
     def _pad_index(cache, target, sample_keys, sample_positions):
         length = 0 if cache.index_keys is None else cache.index_offset
+        if isinstance(length, mx.array):
+            length = int(length.item())
         left = target - length
         if cache.index_keys is None:
             keys = mx.zeros(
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for fixing these QSA batch-join failures. I reproduced the mixed 2-D text / 3-D MRoPE position crash and the array-offset failures on current main; this patch correctly handles the mixed-rank joins and normalizes the offset values used for padding and slicing. I found one remaining invariant gap in non-empty `BatchQSAKVCache.merge()` and empty-left `extend()`, where the KV and indexer widths can still diverge, and I already have a warmed-cache follow-up ready for main. That follow-up does not block these independent fixes—LGTM, merging.

PR #3372, opened 2026-09-02
perf(qwen4_exp): batch the PLE sharded-embedding gather (+18% long-context prefill)

evgri243:
## Summary

Builds on #3287’s cold-page prefetching for SSD-backed Qwen4-Exp PLE tables. Decode IDs scatter across many shards, usually leaving only 1–2 rows per shard—below `rows()`’s `>8` prefetch threshold—and the existing path builds one gather/dequantize graph per touched shard.

## Change

- Bucket unique IDs across all shards with NumPy.
- Collect only unseen pages across every touched weight/scales/biases tensor and issue them through #3287’s shared PLE I/O pool in one dispatch.
- Keep mmap copies for warm pages; warm repeats issue no `pread`s.
- Join raw rows and dequantize once instead of once per shard.
- Do not retain a duplicate hot-row cache.
- Fall back to the existing path for heterogeneous shard layouts.
- Preserve `OMLX_QWEN4_PLE_GATHER=0` as a runtime kill switch.

## Validation

- 76/76 relevant Qwen4 tests pass:
  - `tests/test_mlx_vlm_qwen4_exp_compat.py`
  - `tests/test_vlm_qwen4_exp_loader.py`
  - `tests/test_qwen4_exp_residency.py`
  - `tests/test_qwen4_exp_mtp_admin_gate.py`
- Added regressions for FP8 parity, empty IDs, heterogeneous-layout fallback, and cross-shard cold prefetch followed by a zero-`pread` warm repeat.
- Ruff syntax/undefined-name checks pass.

Synthetic warm A/Bs on this checkout:

| Shape | Per-shard path | Batched path | Difference |
|---|---:|---:|---:|
| 128 affine shards, 16 scattered IDs | 541 µs | 353 µs | −35% |
| 128 dense shards, 4,096 random IDs | 11.33 ms | 1.64 ms | −85% |

These microbenchmarks verify that composing with #3287 removes the previous warm-`pread` regression. A fresh end-to-end decode A/B against current `main` is still required because #3469 materially changed the host/GPU pipeline.

### DIFF
--- a/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
+++ b/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
@@ -1843,6 +1843,13 @@ def _find_nth_prime_after(start: int, count: int) -> int:
 # os.pread releases the GIL and does not change the shared file position.
 _PLE_IO_POOL = ThreadPoolExecutor(max_workers=48, thread_name_prefix="ple-io")
 _PLE_PAGE_SIZE = os.sysconf("SC_PAGE_SIZE")
+_PLE_DTYPE_INFO = {
+    "BF16": (np.dtype("<u2"), 2),
+    "F16": (np.dtype("<f2"), 2),
+    "F32": (np.dtype("<f4"), 4),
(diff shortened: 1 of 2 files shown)


evgri243:
Superseded by https://github.com/jundot/omlx/pull/3287

PR #3376, opened 2026-09-02
fix(mac): align the macOS app UI with system conventions

Lwrless:
## Motivation

I like the app, but small inconsistencies added up. Controls had a different width on every row. Toggles were oversized next to the 13 pt row labels. Headers and footers did not line up between screens. Each problem is small, but together they made the app feel unfinished, and after a while I could not stop noticing them. This pull request is one consistency pass. It moves every screen onto shared components and System Settings conventions.

## What changed

The changes are visual only. The public component APIs (`.buttonStyle(.omlx(...))`, `Row`, `Popup`, `TextInput`) are unchanged at their call sites, except where noted below.

- Control widths: `Theme.swift` adds four shared width tokens (`.controlNarrow`, `.controlCompact`, `.controlMedium`, `.controlWide`). All ~111 `TextInput` and `Popup` call sites now use them, so the trailing controls on a screen line up in columns. `TextInput` now uses a fixed frame. Before, a long row label squeezed its field to a random width. `Popup` is now one native `.menu` picker. The custom-drawn `fillsWidth` variant is gone.
- Toggles: All ~50 bare `Toggle("", ...)` sites now use one shared `RowSwitch`. It renders a small switch, the same size as System Settings.
- Footers: A new shared `FooterBar` shows the hint, the actions, and the error line. It replaces two private `HintFooter` copies and per-screen red error text on 11 screens. The command blocks on the Integrations screen now reuse the quiet `CopyIconButton`.
- Buttons: `OMLXButtonStyle` now delegates to the native styles. Primary and destructive use `.borderedProminent`, destructive with a red tint, and normal uses `.bordered`. The `.plain` kind keeps the quiet look.
- Spacing: The screen gutter is 14 pt everywhere. Section headers are indented 28 pt, so the header text aligns with the row text, like System Settings. All top-level cards use the 14 pt theme corner radius. Segmented controls hug their content instead of fixed frames.
- Navigation: The sidebar can no longer collapse, because a collapsed sidebar had no path back. Every screen keeps one toolbar height. On macOS 26, the invisible toolbar item that holds this height hides its glass capsule.
- Details: The Downloads source switcher is now a centered pair of tabs. The About link rows are fully clickable and highlight on hover. The throughput bench chips fit in standard rows and do not wrap mid-number. The menubar sparkline no longer clips at 0% and 100%.

Net: 28 files, +465/−586.

## Tests

- The full unit test suite passes (`xcodebuild test`, scheme oMLX).
- I tested every screen in a staged debug build, in light and dark mode, down to the minimum window width.

### DIFF
--- a/apps/omlx-mac/Sources/AppView/AppView.swift
+++ b/apps/omlx-mac/Sources/AppView/AppView.swift
@@ -21,8 +21,12 @@ struct AppView: View {
         let theme = scheme == .dark ? OMLXTheme.dark : OMLXTheme.light
         let section = selectedSection
 
-        NavigationSplitView {
+        // The sidebar is the only way to switch screens, so it must never
+        // collapse: pin visibility to `.all` and remove the toolbar toggle,
+        // matching System Settings.
+        NavigationSplitView(columnVisibility: .constant(.all)) {
             SettingsSidebar(selection: bindingForSelection())
+                .toolbar(removing: .sidebarToggle)
         } detail: {
             ContentScaffold(section: section, detailTitle: detailTitle) {
                 screen(for: section)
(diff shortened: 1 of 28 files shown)


Lwrless:
The diff alone does not show what this PR changes. The changes are visual, and they are spread across 28 files, so they are hard to judge without a build. To make the review easier, here is a before / after capture of every screen.

Each screen at the same window size, light mode. Before is `e6805875`, after is this branch.

One change is not visible here: the menubar GPU graph now insets its plot by half of the line width, so a flat line at 0% or 100% no longer renders half-clipped at the canvas edge.

<details><summary>Status</summary>

pair-01-status


</details>
<details><summary>Appearance</summary>

pair-02-appearance


</details>
<details><summary>Server</summary>

pair-03-server


</details>
<details><summary>Network</summary>

pair-04-network


</details>
<details><summary>Performance</summary>

pair-05-performance


</details>
<details><summary>Logs</summary>

pair-06-logs


</details>
<details><summary>Models</summary>

pair-07-models


</details>
<details><summary>Downloads</summary>

pair-08-downloads


</details>
<details><summary>Integrations</summary>

pair-09-integrations


</details>
<details><summary>Quantization</summary>

pair-10-quantization


</details>
<details><summary>Throughput</summary>

pair-11-throughput


</details>
<details><summary>Accuracy</summary>

pair-12-accuracy


</details>
<details><summary>Context</summary>

pair-13-context


</details>
<details><summary>Security</summary>

pair-14-security


</details>
<details><summary>About</summary>

pair-15-about


</details>

jundot:
Thanks for the thorough consistency pass and for adding the screenshots. I had already built the PR locally on macOS 26 and was testing the updated screens, and the before-and-after images made the visual changes much easier to compare.

I really appreciate the care that went into aligning the controls, resizing the switches, adopting native button styles, and unifying the footers. The app feels noticeably more polished and coherent while preserving the existing settings behavior.

LGTM, merging.

PR #3383, opened 2026-09-02
fix(responses): honor tool_choice=none, fix MCP tool merge

astutegrogan:
## Summary
- /v1/responses had two independent tool-handling bugs not present in /v1/chat/completions:
  - MCP-provided tools were silently dropped whenever the client sent no tools of its own (merge was gated on the client's tool list being truthy, not on whether tools were actually disabled).
  - tool_choice="none" was never checked before exposing tools to the chat template, so an explicit no-tool-calls request still had tools visible to the model.
  - Both now match /v1/chat/completions' existing tools_disabled semantics.
- Separately, the XML tool-call parser's <function=NAME> and <parameter=NAME> regexes were both \w+, rejecting any hyphenated or dotted name (e.g. get-weather, web.search). The function-name case dropped the whole call; the parameter-name case is worse — it parsed the call successfully but silently dropped the argument, making a failed call look like it succeeded with missing data. Both widened to [A-Za-z_][\w.-]*.

## Test plan
- [x] tests/test_tool_calling.py + tests/integration/test_responses_tool_merge.py (400 tests) pass
- [x] Full suite re-run against current main: 10672 passed, 150 skipped, 1 pre-existing unrelated failure (test_cluster_pipeline_compat.py, confirmed present on main without this commit too)

### DIFF
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -471,7 +471,14 @@ def _xml_element_value_end(text: str, start: int, close_tag: str, next_open: str
         search = close + len(close_tag)
 
 
-_XML_PARAMETER_OPEN_RE = re.compile(r"<parameter=(\w+)>")
+# Parameter names follow the same convention as tool/function names (see
+# _XML_FUNCTION_CLOSE handling above and _parse_bracket_tool_calls): they
+# must start with a letter/underscore and may contain word chars, dots, and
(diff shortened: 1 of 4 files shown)


jundot (approved):
Thanks for fixing MCP tool merging and `tool_choice="none"` handling in the Responses API. I'll merge this and immediately follow up on the XML regex regression to preserve previously accepted function and parameter names, add regression coverage, and correct the comments describing the naming rules.

PR #3397, opened 2026-09-03
fix(cache): warn at startup when this model's SSD blocks have another block_size

ivaniguarans:
## Summary

When the paged SSD cache starts, `PagedSSDCacheManager._scan_existing_files` reads every block's metadata and quietly sets aside the blocks that fail the compatibility check. That is deliberate — the directory can be shared by several models (#1413) — but the check returns a bare bool, so the scan cannot tell "another model's blocks live here" from "every block of the model I just loaded was written with a different `block_size` and is now unusable". Both cases surface identically, as `skipped_incompatible=N blocks` at INFO.

The second case is what an operator hits after #3066: on ≥ 64 GiB non-NAX hosts the ArraysCache block size moved 2048 → 4096, `_enlarge_block_size_for_arrays_cache` takes `max(2048, prefill_step_size, floor)` so the old size cannot be pinned back, and the whole SSD cache for that model goes dark on the next start. TTFT is cold again, the blocks keep counting against the SSD budget until LRU age evicts them, and the log offers nothing to connect the two. The per-hit `Cache block size mismatch … Invalidating cache hit` warning does not help here: rejected blocks are never indexed, so it cannot fire for them.

This PR makes that situation visible at startup with one WARNING, and gives the INFO summary a per-reason breakdown so the shared-directory case stays recognisable as such.

## Changes

- `omlx/cache/paged_ssd_cache.py`: `_is_compatible_block` now delegates to a new `_incompatibility_reason`, which returns the first failing check — `model_name`, `num_layers`, `block_size`, `layer_cache_types`, `cache_signature` — in exactly the order the boolean used to test them, so the three call sites classify identically. The scan tallies rejects by reason and appends `[reasons: model_name=1, block_size=3]` to the existing `SSD cache scan complete` line (the existing substrings are untouched). When blocks of the loaded model — a positive name match only; an empty expected name never counts, since `factory.py` passes `config.model_name or ""` — were rejected for `block_size` alone, one WARNING names how many blocks and bytes, each stored `block_size`, the configured one, and the two ways to reclaim the space (LRU eviction or `POST /api/ssd-cache/clear`). Nothing is deleted, indexed or evicted differently.
- `tests/test_paged_ssd_cache.py` (additions only): the positive case with two stored sizes; negative cases for model-name mismatch, layer-count mismatch, empty expected model name, unset expected block size and a legacy unstamped size (`block_size=0`, which is rejected but is not "another block size" and stays out of the WARNING) — each asserting no WARNING, the exact reasons suffix and unchanged indexing; label tests for every single-check failure and for first-failure ordering when several checks fail at once.

The rewrite drops one baseline ruff SIM103 hit in the replaced function (15 → 14 in the file); the copied conditions keep upstream's shape otherwise.

## Test plan

- `uv run pytest tests/test_paged_ssd_cache.py -q` — 178 passed (170 existing, unchanged, + 8 new). The tests were written first: six failed with `AttributeError: … no attribute '_incompatibility_reason'` / the missing suffix and WARNING before the change; the two "nothing to warn about" negatives pass by construction.
- Full suite — see the gate result below.
- Manual expectation on a host that crossed the #3066 boundary with an existing cache: the first start after the upgrade logs one line of the form `SSD cache: N block(s) (X GB) for model '<name>' were written with block_size=2048 (N) but this server uses configured block_size=4096; blocks of another block size cannot be reused. …`, immediately after the scan summary; a directory shared by two models logs nothing new beyond the `[reasons: model_name=N]` suffix.

Related to #3066 — surfaces the dead cache after the block-size change and names the reclaim paths; it does not restore reuse of the old blocks (that would need re-chunking, a separate design question). Also relevant to the "cache stopped working after upgrade" reports around #3317.

— Generated with Marshal · gate: READY · 57c5e54 · tier: standard
verify bar: passed at 57c5e54 — 1 check(s) + secrets scan, sha-keyed proof

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -2250,6 +2250,13 @@ class PagedSSDCacheManager(CacheManager):
         skipped_incompatible = 0
         skipped_incompatible_bytes = 0
         errors = 0
+        # Why each incompatible block was rejected, and — separately — the
+        # blocks of THIS model that were rejected only because they were
+        # written with another paged block size (the #3066 upgrade path,
+        # where the ArraysCache block size moved 2048 -> 4096 on large hosts
+        # and the whole cache went dark with no signal beyond the count).
+        reason_counts: dict[str, int] = {}
+        same_model_size_mismatch: dict[int, tuple[int, int]] = {}
 
         for subdir in self.SUBDIR_CHARS:
             subdir_path = self._cache_dir / subdir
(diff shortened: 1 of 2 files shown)


ivaniguarans:
The 3.13 failure is `tests/test_sdpa256_attention.py::test_flash_sdpa256_memory_is_sub_quadratic` dying with a Metal `GPU Timeout Error` on the runner — unrelated to this diff (it only touches the SSD scan and its tests, which passed on all three interpreters). The same test hit the same timeout on #2630 a few hours later and went green on rerun. Happy to rebase for a fresh run if you'd prefer that over a rerun.

PR #3405, opened 2026-09-03
fix(vlm): scalarize Qwen3.5 temporal repeat on MLX 0.32.2

jonathan308:
## Summary

- scalarize the Qwen3-VL temporal patch count before passing it to `mx.repeat`
- apply the compatibility patch only to Qwen3.5-family VLM loads
- preserve the checkpoint image budget, pixels, and tensor computation unchanged
- fail closed if the pinned mlx-vlm implementation no longer matches the known call shape

This mirrors the scalar-repeat compatibility correction from mlx-vlm PR #1982 / commit `1249c7d`, which our Qwen4 compatibility bridge already used. The gap affected Qwen3.8-27B because it inherits the pinned Qwen3-VL vision tower directly.

## Failure reproduced

A fresh Python 3.13 install resolved MLX 0.32.2. Text, exact-resident TTFT, and required tool calls passed, but a real PNG request ended as an empty stream with:

```
repeat(): incompatible function arguments
Invoked with types: mlx.core.array, mlx.core.array
```

The long-lived development venv used MLX 0.32.0, which accepted the old call and masked the release-install incompatibility.

## Validation

- 187 focused tests passed on the upstream branch
- 191 Fusion cache/MTP/lifecycle tests passed
- clean GitHub clone + fresh Python 3.13 venv + MLX 0.32.2
- Qwen3.8-27B real screenshot completed HTTP 200 through the native VLM path (3,116 image tokens)
- Qwen3.8 Flash-Next image, required-tool, and three-turn TTFT checks also passed

Lossless by design: this converts a scalar array to the equivalent Python integer required by the MLX API. It does not cap resolution, alter preprocessing, or change model math.

### DIFF
--- a/omlx/patches/mlx_vlm_qwen35_vision_repeat.py
+++ b/omlx/patches/mlx_vlm_qwen35_vision_repeat.py
@@ -1,0 +1,101 @@
+# SPDX-License-Identifier: Apache-2.0
+"""MLX 0.32.2 compatibility for the pinned Qwen3-VL vision tower.
+
+The pinned mlx-vlm revision passes ``grid_thw[i, 0]`` (an MLX scalar array)
+as ``mx.repeat``'s repeat count.  MLX 0.32.2 requires a Python integer.  This
+is the same scalarization shipped upstream for the shared Qwen vision tower;
+the tensor graph and image resolution remain otherwise unchanged.
+"""
+
+from __future__ import annotations
+
+import inspect
+import logging
+from typing import Any
+
+import mlx.core as mx
+
+logger = logging.getLogger(__name__)
+
+
+def _call_with_scalar_repeat(
+    self: Any,
+    hidden_states: mx.array,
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for the fix. I tested Qwen3.8-27B image requests on MLX 0.32.2 without this patch and couldn't reproduce the error. #3332 already includes the same scalar conversion for the shared Qwen3-VL vision tower. Could you share the oMLX commit, model checkpoint, and full traceback from the failing run so I can check which loading path missed the existing patch?

jonathan308:
Confirmed. My reproduction was from August 28 on the Fusion branch, before #3332 merged. The traceback reached `qwen3_5` → `qwen3_vl.vision` at the same `mx.repeat` call now covered by the shared MLX 0.32.2 compatibility hook. Current `main` therefore already contains the fix, so #3405 is redundant. Closing as superseded by #3332—thanks for catching this.

PR #3409, opened 2026-09-03
perf(moe): the affine block kernels engaged at 64 routes, where they cost 2-2.7x the plain gather

pedroberaldo87:
A single 64-route gate turned on both route sorting and the native affine
block kernels. Measured on M1 Ultra with real GLM-5.3-Flash expert weights
(2-bit, group 64), per layer: sorting pays off from 32 routes, but the block
kernels only beat mx.gather_qmm from ~1024 routes (the crossover is at
T=128 x 8 experts); between 64 and ~900 routes they cost 2-2.7x. Decode and
the MTP verify window live exactly in that band.

Split the gate: sort from 32 routes, block kernels from 1024. Server, batch
of 8 (B=8, 300 tokens each): 278.5 -> 173.5 ms/step, 28.7 -> 46.1 tok/s.
Prefill (chunks >= 512 tokens, >= 4096 routes) stays on the block kernels.

Both thresholds are environment-overridable (OMLX_DEEPSEEK_SORT_MIN_ROUTES,
OMLX_DEEPSEEK_AFFINE_BLOCK_MIN_ROUTES), like the mxfp4 neighbour: they were
tuned on one machine. The two existing block-kernel parity tests now feed
512 tokens x 2 experts = 1024 routes — the smallest count at which the
kernels engage — and a new decision test checks the threshold without the
Metal kernel present.

### DIFF
--- a/omlx/patches/deepseek_v4/switch_layers.py
+++ b/omlx/patches/deepseek_v4/switch_layers.py
@@ -21,6 +21,29 @@ _DEEPSEEK_MXFP4_SMALL_BLOCK_VARIANT = 1
 _DEEPSEEK_MXFP4_LARGE_BLOCK_BM = 32
 _DEEPSEEK_MXFP4_LARGE_BLOCK_VARIANT = 2
 _DEEPSEEK_AFFINE_LARGE_BLOCK_MIN_ROUTES = 8192
+# Below this many (token, expert) routes the stock unsorted gather_qmm runs;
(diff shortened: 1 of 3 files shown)


jundot:
Thanks, separating the sorting and affine block thresholds makes sense. I also found that the shared 32-route sorting threshold slows down small MXFP4 windows on my M3 Ultra. I'll merge this and follow up by keeping the lower threshold specific to affine, preserving the previous threshold for other paths, and trimming the duplicated benchmark commentary.

PR #3414, opened 2026-09-03
fix(scheduler): an eviction pause mid external prefill lost progress and duplicated KV tokens

pedroberaldo87:
## The warm path: a restored prefix

The external prefill loop sizes each chunk through the adaptive throttle,
which can raise `_PrefillEvictionNeeded` AFTER chunks have already entered
the cache (`prompt_cache` is advanced in place). The pause re-queued the
request with the `remaining_tokens`/`cached_tokens` it started with, so the
retry fed the same tokens again on top of the advanced cache.

The damage compounds: a duplicated span in the KV, boundary snapshots
labelled one block behind what the cache held ("tc=7680 kv=8192",
"tc=8192 kv=8704"), and every block stored from there on misaligned —
layer 3's KV differed in 2.6 of 2 positions, i.e. it held keys for OTHER
tokens. The next request that hit those blocks generated garbage ("NoNo",
"DEDEDE", "Sand Sand"). Reproduced through the API in three steps: A
stores the prefix, B restores and prefills while throttled, C hits it and
generates garbage (2 of 2).

It only surfaced with the MTP head enabled, because only then did memory
pressure push the throttle far enough to pause a chunk — the head was the
witness, not the cause.

The pause now commits the progress: `cached_tokens += processed` and
`remaining_tokens` = what is left.

## The cold path: no prefix at all

The commit above only committed the progress when `existing_cache` was set.
A **cold** prompt builds its cache locally, so `existing_cache is None` and the
pause re-admitted the request from token zero — the whole prefill so far,
thrown away.

Measured 05/09 on GLM-5.3-Flash oQ2e: a cold 229,923-token prompt was paused at
206,336 tokens for headroom, the eviction reclaimed 1.56 GB, and the request
came back as `new=229923 kv_exact=0` — **19 minutes of prefill lost**. With the
fix, the next 229k run paused twice at 192k, each pause cost ~6 s, and only the
remaining 36k tokens were re-admitted.

## Tests

- `test_pause_mid_external_prefill_commits_the_progress` — fails without the first commit (`4 == 8`).
- `test_pause_on_a_cold_prompt_commits_the_progress_too` — fails with the first commit alone (`prompt_cache` is `None`).

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -3619,7 +3619,22 @@ class Scheduler:
                 # stored from there on misaligned — the next prefix hit
                 # generates garbage (GLM-5.3, 02/09, only reachable when the
                 # MTP head's memory pushed the throttle into a paused chunk).
-                if existing_cache is not None and processed_tokens > 0:
+                #
+                # A COLD prompt has to commit too. The restored-prefix case was
+                # the one measured in 02/09, so the commit was gated on
+                # ``existing_cache``; a cold prompt builds ``prompt_cache``
+                # locally, the pause dropped it with this exception, and the
+                # retry started from token zero. Measured 05/09 on GLM-5.3-Flash
+                # oQ2e: a 229.923-token prompt was paused at 206.336 tokens
+                # ("needs prefill headroom ... current=111.76GB target=112.48GB"),
+                # the eviction reclaimed 1,56 GB, and the request was re-admitted
+                # as ``new=229923 kv_exact=0`` — 19 minutes of prefill thrown
+                # away, and nothing stops it from happening again at the same
+                # point. Attaching the advanced cache to the request is what the
+                # retry path already reads (``cache_to_use = request.prompt_cache``).
+                if processed_tokens > 0:
+                    if existing_cache is None:
+                        request.prompt_cache = prompt_cache
                     request.cached_tokens = (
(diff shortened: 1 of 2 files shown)


pedroberaldo87:
Second commit pushed: the same pause on a **cold** prompt (no restored prefix) also lost its progress — the first fix only committed when `existing_cache` was set, and a cold prompt builds its cache locally, so the retry restarted at token zero.

Measured 05/09 on GLM-5.3-Flash oQ2e: a cold 229,923-token prompt was paused at 206,336 tokens for headroom, the eviction reclaimed 1.56 GB, and the request came back as `new=229923 kv_exact=0` — 19 minutes of prefill thrown away. With the fix, the next 229k run paused twice at 192k, each pause cost ~6 s and re-admitted only the remaining 36k tokens.

New test: `test_pause_on_a_cold_prompt_commits_the_progress_too` (fails on the first commit alone: `prompt_cache` is `None`).

jundot:
Thanks for tracking down both the warm and cold cases. I verified that the throttle pause now preserves progress and resumes with matching KV and logits. The same exception from `_guard_prefill_chunk()` still bypasses this handler, so I'll cover that path in a follow-up with resume tests and trim the comments. LGTM with that follow-up; I'll merge this first.

PR #3415, opened 2026-09-03
fix(mtp): the reconcile re-prefill runs in scheduler-sized chunks, not one forward

pedroberaldo87:
When the MTP cycle falls back to the plain step, _reconcile_mtp_to_standard
rebuilds the cache by re-prefilling the whole history. It did that in a
single forward — 13-30k tokens in an agent loop — which both blows memory
(the prefill throttle fired only with MTP on) and lands on sparse-attention
paths the chunked prefill never takes.

Measured on GLM-5.3-Flash oQ2e (2602 tokens): the cache rebuilt in one shot
diverges from the chunked one (512) from layer 7 on (38 positions), and at
nearly every position from layer 19 on. It now uses the generator's
prefill_step_size (512 on the server), so the rebuilt cache matches the
prefill the request already had.

Test: test_reconcile_re_prefills_in_scheduler_chunks asserts the chunk
shapes (512, 512, 276) for a 1300-token history.

### DIFF
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -1296,8 +1296,23 @@ def _reconcile_mtp_to_standard(gen_batch: Any, state: _MtpState) -> bool:
         procs = _proc_list(gen_batch)
         _set_singleton_mrope_delta(gen_batch)
         tok_arr = _ensure_uint32(mx.array(list(tokens)))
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for splitting the reconcile prefill into chunks; this removes the full-history forward. I found that `GenerationBatch` doesn't receive `prefill_step_size`, so the new path always falls back to 512, even with a smaller configured chunk. I'll carry the setting through generation and row batches in a follow-up and test it with a non-default size. LGTM with that follow-up, merging.

PR #3416, opened 2026-09-03
fix(mtp): the chain rollback could not find the method behind the VLM adapter

pedroberaldo87:
The mlx-vlm runtimes stamp mtp_partial_rollback on the inner LanguageModel,
but the batch sees oMLX's VLMModelAdapter wrapper, which does not forward
it. _chain_rollback looked only at the wrapper, found nothing, and every
draft reject fell through to "cache layer rejects chain rollback" plus a
full reconcile.

Measured on a GLM-5.3 vision checkpoint before the fix: 2 tok/s with the
MTP head reporting cycles=0 — speculation was on, doing nothing but paying
the reconcile. After: 17.5 / 19.8 / 18.9 tok/s over 3x300 tokens at 73%
acceptance, against 18.8 / 19.2 / 19.2 with the head off.

The lookup now walks model, language_model and _language_model.

### DIFF
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -3201,7 +3201,16 @@ def _chain_rollback(
         except Exception as exc:
             logger.debug("rollback_speculative_cache failed: %s", exc)
             return False
-    rollback = getattr(model, "mtp_partial_rollback", None)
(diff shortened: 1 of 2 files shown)


jundot:
Thanks, the adapter lookup fixes the missing rollback hook. I verified that partial acceptance now trims the cache through the VLM adapter while preserving the existing GDN rollback priority. I'll resolve the test conflict and merge this, with a small follow-up to shorten the repeated explanatory comments.

Issue #3392, opened 2026-09-02
OMLX WebUI model downloader shows wrong sizes from HF


bplein:
**Describe the bug**
OMLX WebUI model downloader shows wrong sizes from HF compared to actual.

**To Reproduce**
1. Use the UI
2. Find a model on HF
3. See how fat it appears.

**Expected behavior**
In the prior release, the size reflected in the downloader matched the size on HF

**Screenshots**

<img width="1307" height="562" alt="Image" src="https://github.com/user-attachments/assets/1dc8bb46-7ddb-4325-a9f6-5f5aa9457488" />

**Desktop (please complete the following information):**
 - macOS Version: 26.6.2
 - oMLX Version v0.6.4
 - 
**Additional context**
Started happening with the latest release

wtchen:
I am on the latest release and it was correct yesterday, but not today

PR #3419, opened 2026-09-03
fix(admin): size U32-packed MLX quants from Hub blob bytes

### LINKED ISSUE #3401: Browse/Search size estimate treats U32-packed MLX quants as 4 bytes/param (~6x inflated) "High OOM risk" badges, Recommended tab hides models that fit
**Environment:** oMLX v0.6.4, Mac Studio M3 Ultra 96 GB, admin UI → Models → Browse / Search.

**What happens**
Every MLX-quantized repo shows a wildly inflated size plus a red "High OOM risk"
badge (>95% of RAM). Example:

`mlx-community/gemma-4-26B-A4B-it-4bit` shows 102.1 GB / high OOM risk on a 96 GB machine.
Starting the download shows the true size (~15.4 GB) in the progress bar.

**Reproduce**
1. Admin UI → Models → Search for any 4-bit MLX quant (e.g. gemma-4-26B-A4B-it-4bit).
2. Note the size + badge, then start the download — progress reports the real size.

**Evidence (via HF API today)**
| repo | shown in UI | actual safetensors |
|---|---|---|
| mlx-community/gemma-4-26B-A4B-it-4bit | 102.1 GB | 15.4 GB |
| Jundot/gemma-4-26B-A4B-it-oQ4e-mtp | 102.9 GB | 16.6 GB |

**Root cause**
`_calc_safetensors_disk_size()` computes `Σ param_count × _DTYPE_BYTES[dtype]` from HF's `safetensors.parameters` metadata. For MLX group-quantized checkpoints HF reports the *logical* parameter count under `U32` (packed weights), so 4-bit quants are billed at 4 bytes/param → ~6.6x inflation. Metadata for the repo above: `{"U32": 25235685376, "BF16": 570250830}`.

**Affected paths**
- Recommended tab filter `size > max_memory_bytes` silently hides quants that actually fit (26B-A4B-4bit on 96 GB, etc).
- Search results + model detail size display.
- dashboard.js `getMemoryFitStatus()` — the warning/danger badges.

**Suggested fix**
Use actual blob sizes: `model_info(repo_id, files_metadata=True)` and sum `siblings[].size` for `*.safetensors` (matches what download progress reports).
`usedStorage` is NOT a valid shortcut — it returns 46.3 GB for the 15.4 GB repo above because it includes historical revisions. 
Interim mitigation: when the dtype histogram is U32-dominated (quant signature), show "unknown" rather than a red badge, and skip the Recommended-tab exclusion.

Impact: on 64-96 GB Macs most of the 4-bit ecosystem displays as high OOM risk (including the Jundot oQ4e-mtp quants), which is misleading for new users.

 - macOS Version: 26.4.1
 - oMLX Version 0.6.4

aywengo:
## Summary
- Browse / Search / Recommended billed MLX group-quantized checkpoints at 4 bytes per logical U32 param, so a 15.4 GB 4-bit repo showed as ~102 GB and got a High OOM risk badge (and was dropped from Recommended on 64–96 GB machines).
- For U32 histograms, size now comes from current-revision `*.safetensors` blob sizes (`model_info(..., files_metadata=True)`). BF16/F16 still use the cheap dtype formula. Model detail already had sibling metadata and now uses it; download fallback does the same instead of `U32 × 4`.
- Do not use `usedStorage` (it includes historical revisions). A failed/timed-out U32 fetch leaves size unknown rather than hiding the model.

Closes #3401. Same bug as #3392.


omlx/admin/hf_downloader.py
```diff
@@ -415,48 +553,63 @@ async def _fetch(sort: str) -> tuple[list[dict], bool]:
...
-                    continue
-                downloads = m.downloads or 0
-                if downloads < _MIN_DOWNLOADS:
+            for model in models:
+                if not _eligible(model):
                     continue
-                size = _calc_safetensors_disk_size(m.safetensors)
-                if size <= 0 or size > max_memory_bytes:
+                size = _list_disk_size(model, blob_sizes)
+                if size > max_memory_bytes:
```
jnchaba: The condition on `main` is `size <= 0 or size > max_memory_bytes`. This change removes the `size <= 0` test. A model with an unknown size now stays in the list.

I did a test with a repository of 400 billion parameters and a memory limit of 16 GB. The blob request failed, so the size was 0. The code on `main` gives 0 rows. This branch gives 1 row with an empty size field.

The Recommended tab must show only the models that fit in the memory. Issue #3401 reports incorrect models in this tab. This change causes a different form of the same problem.

Please keep the `size <= 0` test. As an alternative, show a clear "unknown size" label in the user interface.

omlx/admin/hf_downloader.py
```diff
@@ -522,16 +675,20 @@ async def search_models(
             timeout=_HF_API_TIMEOUT,
         )
 
+        need_blob = [m.id for m in models if _needs_u32_blob_fetch(m)]
```
jnchaba: This line collects the repository IDs before the filter loop. The loop at line 682 removes the rows that do not agree with `min_params`, `max_params`, `min_size`, and `max_size`. The code requests the blob sizes for rows that it then removes.

The function `get_recommended_models` does not have this problem. It calls `_eligible` first at line 576.

You could fix this by applying the `min_params` and `max_params` filters before this line. The size filters need the size, so they must stay after this line.

omlx/admin/hf_downloader.py
```diff
@@ -314,17 +314,152 @@ def to_dict(self) -> dict:
...
+                "Could not fetch blob sizes for %s: %s", repo_id, retry_error
+            )
+            return 0
+    except Exception as e:
+        logger.debug("Could not fetch blob sizes for %s: %s", repo_id, e)
+        return 0
+    return _sum_safetensors_blob_bytes(getattr(info, "siblings", None)) or 0
+
+
+async def _blob_bytes_for_repos(api: HfApi, repo_ids: list[str]) -> dict[str, int]:
```
jnchaba, line 423: This function sends one `model_info` request for each U32 repository. I did a search for "qwen" with a limit of 100 rows. 87 rows needed a request. The code sent 85 requests, because it removes the repeated repository IDs.

The number of requests is my only concern. I measured the time, but the time is not a good measurement here. The network gives very different results for each test. In one test, `main` was slower than this branch. The extra requests do not make a large delay when the Hub is fast.

But 85 more requests for each search is a large increase. Many users have no Hub token. The Hub rate limit for these users is low. A user who makes many searches can reach the limit.

Please add a cache for the sizes. The size of a repository does not change frequently. As an alternative, request the sizes only for the rows that the user sees.

jnchaba (changes requested):
I ran into this exact same issue yesterday and was going to fix it -- but you beat me to it! 

The PR is good overall:

- The main correction is right, and the tests are good. 137 tests pass.
- You refuses `usedStorage` for a correct reason. I measured this. For `mlx-community/Qwen3-0.6B-4bit`, `usedStorage` gives 682,323,786 bytes. The files in the current revision have 351,386,061 bytes. The `usedStorage` value is 1.94 times too large. On `orcarouter/Qwen3.8-27B-Uncensored-MLX`, `usedStorage` gives 78.5 GB, which is also incorrect.
- The `int()` conversion at lines 458 and 492 prevents a failure of the full page from one bad row.
- The second request without a token at line 371 keeps the public repositories visible after a 401 response.

However, I do have some concerns about the implementation:
- you've removed a guard in a conditional in `_build` that prevented models with an unknown size from appearing from the list
- the logic in `_blob_bytes_for_repos` sends many network requests, one per U32 repository, regardless of how many models are listed. this will make users hit their HF hub rate limit faster - I recommend adding a cache for sizes or lazy loading (only fetching blob_bytes for the repos they can see).
- when searching for models, you're doing a call to get the blob fetch before models are filtered out. you could filter by `min_params` and `max_params` (which dont need size) before you do the network call to fetch a model's size - this would be more efficient.

aywengo:
Thank you for the careful review and for measuring `usedStorage` — that was really useful.
Recommended drops unknown-size rows again (`size <= 0`), so a failed blob fetch cannot list a 400B model on a 16 GB machine.
Search applies min/max param filters before any `model_info` call, and positive blob sizes are cached in-process so a repeat search does not hit the Hub again.
Fetch failures are not cached, so a later request can still pick up a real size.
Happy to keep helping get this right for oMLX.

omlx/admin/hf_downloader.py
aywengo: Thank you for catching this and for the 400B / 16 GB test.
Recommended now uses `size <= 0 or size > max_memory_bytes` again, so a failed blob fetch is dropped from that list.

omlx/admin/hf_downloader.py
aywengo: Good call — that was wasted Hub traffic.
Search now applies `min_params` / `max_params` first, then fetches blob sizes only for the rows that remain.

omlx/admin/hf_downloader.py
aywengo: Agreed — 85 extra `model_info` calls per search is too many for unauthenticated Hub limits.
Positive blob sizes are now cached in-process (1 hour TTL). Failures stay uncached so a later request can still pick up a real size.

omlx/admin/hf_downloader.py
```diff
@@ -314,17 +314,185 @@ def to_dict(self) -> dict:
...
+        _blob_size_cache.pop(repo_id, None)
+        return None
+    return size
+
+
+def _store_blob_size(repo_id: str, size: int) -> None:
+    """Cache Hub blob sizes that actually resolved. Failures stay uncached."""
+    if size > 0:
+        _blob_size_cache[repo_id] = (size, time.monotonic())
+
```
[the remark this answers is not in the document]
aywengo: Good catch !

omlx/admin/hf_downloader.py
```diff
@@ -907,24 +1120,36 @@ async def _run_download(self, task_id: str, hf_token: str) -> None:
                             api.model_info,
                             task.repo_id,
                             token=hf_token or None,
+                            files_metadata=True,
```
jnchaba, line 1123: `_run_download` passes `files_metadata=True` and `expand=["safetensors"]` together
(`omlx/admin/hf_downloader.py:1119`). huggingface_hub rejects that combination on the
first line of `model_info`, before any HTTP request:

```
ValueError: `expand` cannot be used if `securityStatus` or `files_metadata` are set.
```

The `except Exception` at :1152 swallows it, so on every download:

- `st_estimate` stays 0, so the new sibling-blob fallback never runs
- `ignore_patterns` stays `None`, so `*.bin`, `original/**` and `consolidated.*.pth` are
  no longer excluded — repos with both formats download the redundant pytorch weights

CI is green because `mock_api.model_info` is a MagicMock and accepts any kwargs, so the
library's validation never runs. Running `test_dry_run_failure_falls_back_to_safetensors_size`
with the real `model_info` and only `snapshot_download` faked:

```
ignore_patterns handed to snapshot_download: <absent>
task.total_size: 0
```

Dropping `expand=["safetensors"]` fixes it — `files_metadata=True` already returns the
safetensors field. Same harness after the change:

```
ignore_patterns handed to snapshot_download: ['*.bin', 'original/**', 'consolidated.*.pth']
task.total_size: 335450584
```

Worth adding a test that asserts on the kwargs actually passed to `model_info`, since a
MagicMock will never catch this class of bug.

jnchaba (changes requested):
Changes look good -- thanks for doing those! I found one more blocker in my re-review, fixing this would make me comfortable approving; happy to discuss if you disagree.

omlx/admin/hf_downloader.py
aywengo: Oh! Good catch!

aywengo: Thanks! Somehow missed that huggingface_hub rejects `expand` with `files_metadata`, and the mock hid it. The test now checks the kwargs so this cannot slip through again.

jnchaba (approved):

aywengo:
Thanks for the feedback! 
Now PR is ready for @jundot review.

omlx/admin/hf_downloader.py
jnchaba: Nice work!

aywengo:
@jundot I've just merged latest main, conflict’s cleared — ready for another look whenever you have a minute.

jundot:
Thanks for fixing this. I checked the Gemma checkpoint against the Hub file metadata and confirmed the size drops from 102.08 GB to 15.34 GB. LGTM, I'll merge this and add HTTP timeouts to the blob metadata calls in a follow-up so stalled requests release their worker threads.

PR #3424, opened 2026-09-03
fix(dflash): handle memory-pressure aborts

saintlits:
## What changed

- track the stop event for each in-flight DFlash generation
- implement `DFlashEngine.abort_all_requests()` so process-level memory enforcement can stop primary DFlash requests
- delegate aborts to the fallback engine when it is active
- keep active-request reporting accurate until the executor actually exits

## Why

DFlash bypasses `Scheduler`, but unlike the batched and VLM engines it did not expose `abort_all_requests()`. Under hard memory pressure, `ProcessMemoryEnforcer` therefore identified the DFlash model as busy but could not signal its active generation. It repeatedly logged `aborted 0 request(s)` and left a long-context request running under severe unified-memory pressure.

The generation loops already poll a `threading.Event` between DFlash events. This change registers those existing events for the lifetime of their executor work and uses them as the engine's abort handles. Repeated pressure polls are idempotent and do not count an already-signalled request again.

## Validation

- `python -m pytest -q tests/test_dflash_engine.py tests/test_process_memory_enforcer.py` (281 passed)
- `ruff check omlx/engine/dflash.py tests/test_dflash_engine.py`

### DIFF
--- a/omlx/engine/dflash.py
+++ b/omlx/engine/dflash.py
@@ -336,6 +336,12 @@ class DFlashEngine(ActivityTrackingMixin, BaseEngine):
         self._executor_tokenizer = None
         self._loaded = False
         self._active_request = False
+        # DFlash runs outside Scheduler, so memory-pressure aborts cannot use
+        # EngineCore's request registry. Keep the stop events for every
+        # submitted generation instead; the process memory enforcer calls
+        # abort_all_requests() to signal them at the next DFlash event boundary.
+        self._stop_events_lock = threading.Lock()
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for adding the missing abort path. I verified that the memory enforcer can now signal DFlash generation to stop. I also reproduced a stale stop-event entry when a queued request is cancelled before execution, and found that memory-triggered aborts still report `finish_reason="stop"`. I'll address both in a small follow-up immediately after merging.

PR #3425, opened 2026-09-04
feat(ane): price prefill banks at admission from the compiled size

andyoneal:
**This is a draft because I am not sure the approach is right.** The code works and is tested. It uses a private compiler to get an exact number, and a rougher number that needs no such dependency may be the better trade-off.

The memory guard cannot see the ANE prefill banks. A resident bank is charged to the kernel's neural ledger and excluded from `phys_footprint`, so RSS, `vmmap` and MLX's own accounting are all blind to it, and admission charges **zero** for it today. For Qwen3.8-27B at the default fraction the real figure is **7.06 GB** — 64 MLP programs of about 90 MB each, and 48 GDN programs of about 30 MB.

#2841 reports this model aborting requests and evicting itself on a 32 GB machine with ANE prefill on and completing fine with it off, and asks for exactly this accounting; it even names the shape, 5120→18432 across 64 layers. This charges the banks at model admission rather than in the per-request preflight the report suggests, because the banks are resident for the life of the model rather than transient.

## How the number is obtained

The compiled program size is what the driver wires. Compiling one projection ahead of time produces a file whose byte count equals the `wiredMemory` the driver logs at program-create, in every geometry tried — two contractions, four output widths, two dtypes, five chunk widths, from 5.77 MB to 94.37 MB.

That is the property worth the code: it reads a size off disk **before anything is loaded**, rather than inferring it after the fact from accounting that structurally cannot see it. Size depends on the geometry and not on the weight values, so one compile prices every layer sharing a shape, and this model is two shapes.

## A check from a different model, chip and instrument

Gemma 4 12B on an M1 with a single ANE die and 16 GiB. The compiler prices one gate/up projection of 3840→3072 — the f = 0.20 share — at 11,878,400 bytes, so gate and up across 8 layers come to 181.2 MB. Measuring that same configuration on that machine through the kernel's neural ledger, the only instrument that sees a bank at all, gives 181.0 MB. They agree to 0.1%, across a different model family, a different chip generation, one die instead of two, and an instrument with no relationship to the compiler.

The same machine shows what charging zero costs. At f = 0.60 a 4,096-token prefill is refused against a dynamic ceiling of 10.23 GB with current usage at 7.08 GB and 3.69 GB reclaimable. The 3.18 GB of banks are what closed that gap. Admission charged none of it, so the ceiling is held down by memory the usage term cannot see: the refusal is right, for a reason the guard is unable to state, and the fraction that triggers it looks free from the inside.

## Exact, or close enough?

The size can also just be calculated. `ceil_16K(O*C*bits/8 + 2*O)` gets over 99% of it, and the remainder is a small per-program overhead that varies with the output width. Fitting that from measurements and doing arithmetic would land within about 0.2% of a real load, with no private framework, no probe cost, and nothing new to break.

On the Gemma 4 geometry that form runs 0.55% low at output widths 3072, 4608 and 6144, and 0.41% low at 9216. Expressed in 16 KiB pages the residual is 4, 6, 8 and 9 — linear in the width across the first three, and then not, where a fit anchored on them predicts 12. That is a single geometry and not a refutation of fitting, but it is the shape of the difficulty: the overhead is small enough that arithmetic gets close, and irregular enough that a constant fitted from a handful of widths is not reliably close.

That fit would not stay right on its own, either. The overhead is a handful of measured points against an undocumented compiler with no closed form, and this ground has already retired two per-program constants that looked settled. A compile stays correct when the compiler changes; a fitted constant goes quietly wrong.

Given the guard currently charges zero, either beats what is there now, and I do not think 0.2% is what should decide this.

## Trade-off

`ANECCompile` is reached by `dlsym`, and `ANECompiler.framework` is not the framework the ANE prefill path already uses. So this adds a second private dependency to a feature that already carries one, and that is the part I am least sure about.

## Notes

The per-program size is measured, but the total is still a prediction: it multiplies that size by the layers expected to be accelerated, and layers the backend later declines for dtype or row alignment are counted anyway. The total therefore errs high, which is the safe direction for a guard.

The two Qwen sizing helpers in `engine_pool` shared their config parsing and layer counting, so rather than copy it I lifted the common part into three helpers both call.

Without the framework or the symbol the price is 0 and admission behaves exactly as it does today. The target architecture is read from the device's own `aneArchitectureType`, so there is no chip table to maintain.

`_entry_runtime_resident_size` returns the bank price on top of the CPU share, which the pre-existing `test_runtime_estimate_feeds_resident_memory_accounting` does not account for. Left unpatched that test asks the real compiler and so passes only on hosts without `ANECompiler.framework`, where the price is 0. It prices the banks at zero explicitly now, keeping its assertion about the term it is named for.

## Testing

The full suite on this branch is 11,404 passed, 81 skipped, 6 failed. The same 6 fail on `main` on this host: three in `test_decode_fast_sdpa`, one in `test_glm_moe_dsa_patch` and two in `test_mlx_vlm_qwen4_exp_compat`, all float-tolerance comparisons in custom kernels this branch does not touch.

### DIFF
--- a/docs/experimental/qwen35_ane_prefill.md
+++ b/docs/experimental/qwen35_ane_prefill.md
@@ -211,6 +211,57 @@ FP16 gate/up and GDN rows, the down-projection GPU suffix, and bounded
 materialization scratch. This projected size participates in the normal memory
 guard before the model begins loading.
 
+## Pricing the banks at admission
+
+A resident bank is charged to the kernel's neural ledger and excluded from
+`phys_footprint`, so the ordinary accounting cannot see it and admission has
+never charged for it. The compiled program size is what the driver wires, and
+it is available before the load commits anything: compiling one projection
+reports exactly the `wiredMemory` the driver goes on to report
+at program-create. Measured across two contractions, three output widths, two
+dtypes and four chunk widths, on blobs from 5.77 MB to 94.37 MB.
+
+The size is fixed by the geometry rather than by the weight values — checked
+across eighteen arms at three geometries, including all-zero, constant and
+widely varying scales — so one compile prices every layer that shares a shape.
+The 64-layer 27B MLP slice and the 48-layer GDN slice are two shapes, and two
+compiles therefore price the whole layout. `engine_pool` caches one price per
+geometry and adds the total to the projected size.
+
+Compiler options change the compiled size as well, so the price is only
+valid for the option set the load path actually uses; the agreement with
+`wiredMemory` is what establishes that, not the geometry alone. The effect
+is program-dependent rather than a fixed offset, and it is not small: on a
+measured convolution fixture a single scheduling option moved the compiled
+size by 16%.
+
+The compiler is reached through `ANECompiler.framework`, resolved by name.
+Its target comes from the runtime's own `aneArchitectureType` rather than a
+chip table, read from the local device, so the probe compiles for the machine
+it runs on. That request is honoured rather than quietly defaulted, which can
+be checked on its own: compile one source for two architectures and the
+outputs differ byte-for-byte at the same length.
+
+When the framework or the symbol is missing the price is zero and admission
+keeps its previous behaviour, which is to charge nothing for the banks. Do
+not substitute the neural ledger figure for the wired one: the difference
+between them varies with chunk width.
+
+The GDN width comes from the backend's own recurrent-safe boundary rather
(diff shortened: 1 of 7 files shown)


andyoneal:
Rebased onto 0.7.0.dev2.

The bank estimate is now gated on `ane_prefill_backend(...) == "qwen"` and on `include_ane_reservation`, so it does not charge alongside the K2 reservation added in #3486, and it stays out of the settle size.

PR #3428, opened 2026-09-04
qwen4_exp: restore #3350 gathered-QSA eligibility lost in merge 6f76565; keep MTP windows on the official path; retain pooled index across trim

ddark-il:
## Summary

Three commits on the `qwen4_exp` (Qwen3.8-Flash-Next) QSA path, each backed by measurements on an M5 Max 128 GB with a mixed 4.86 bpw build of the model and Lightning MTP on:

1. **Restore the #3350 gathered-QSA eligibility fix that merge 6f76565 dropped.** The #3350 series (61e1027, 262a2c6, 44e51e3, 61cd558) let the gathered prefill arm accept text positions tiled to identical `(3, 1, L)` mRoPE planes and priced such prefills as gathered instead of dense in the memory guard. Merge 6f76565 resolved `language.py` to its first parent verbatim (`git diff 6f76565^1 6f76565 -- …/language.py` is empty; vs `^2` it is 507 lines), and ba45eaa then deleted the tests that referenced the missing symbols. This is a correctness restore of upstream's intent, not a throughput change on its own: on my traffic the scheduler's chunked prefill already arrives with 2-D/absent positions (cold 48k prefill measured at parity, 1205–1286 tok/s both sides), and 3-D planes only show up on the Lightning MTP verify pass, which (2) keeps on the official path anyway. The case #3350 reported — a continuation with mRoPE positions refused by the guard's dense pricing — is what this brings back.

2. **Gate the gathered prefill arm at a minimum query width (16 rows, `OMLX_QWEN4_GATHERED_MIN_QUERY`).** With (1) restored, the 2–4 row history/verify passes Lightning MTP issues every cycle would also gather, and per-layer timing shows the official masked path is cheaper there:

   | ms / layer | S=1 | S=2 | S=4 | S=8 | S=16 | S=64 |
   |---|---|---|---|---|---|---|
   | 206k gathered | 4.20 | 4.56 | 5.02 | 6.28 | 8.65 | 24.6 |
   | 206k official | 1.40 | 2.14 | 3.49 | 6.45 | 12.2 | 47.3 |
   | 48k gathered | 1.74 | 2.01 | 2.61 | 3.68 | 6.46 | 22.5 |
   | 48k official | 1.37 | 1.46 | 2.22 | 3.94 | 7.36 | 27.7 |

   The per-row gather of `token_budget` K/V rows only pays off from about eight query rows. Real prefill chunks are far wider and keep gathering.

3. **Keep the pooled QSA index prefix across `trim()`.** `QSAKVCache.trim()` invalidated the entire pooled index-key bank, so the next call re-pooled every completed block of the context: 1.75 ms per QSA layer at 206k. MTP trims after every cycle with a rejected draft, so long-context decode paid that almost every cycle. Pooled blocks below the new complete-block count are built from raw keys that trim never touches, so the change clamps the pooled frontier and lets `pooled_indexer_keys()` re-pool only the tail (~0.75 ms/layer net, ~9 ms per cycle over 12 layers in-process at 206k). Restore, extract and capacity growth still rebuild from scratch.

## What was tried and rejected

I first extended the gathered arm to Lightning MTP verify rows (`target_verify=True`) and to 3-D text mRoPE single-token decode, on the hypothesis that dense full-KV attention was the long-context decode cost. An interleaved server A/B (fresh process per run, cached 48k/206k prefixes, short-context control) was neutral within noise, and the per-layer sweep above explains why: for S ≤ 4 the gathered kernels are slower than masked SDPA at every context tested. Those arms are not in this PR. Upstream's decision in 61cd558 to keep decode 2-D-only is consistent with this data.

## End-to-end

Interleaved server A/B, upstream/main vs this branch, fresh process per round, warm-up pass then a 3-minute idle cool-down before measuring (earlier attempts without the cool-down drifted +7 ms/cycle per round and were unreadable). M5 Max 128 GB, Qwen3.8-Flash-Next-Uncensored mixed 4.86 bpw, Lightning MTP on, prefix cached, 300 generated tokens, mean of 2 passes:

| round | build | 80-tok control ms/cycle | 206k tok/s | 206k backbone ms/cycle | tok/cycle |
|---|---|---|---|---|---|
| 1 | branch | 41.0 | 27.5 | 60.1 | 2.25 |
| 2 | main | 46.4 | 23.5 | 57.2 | 1.79 |
| 3 | branch | 43.7 | 27.1 | 60.1 | 2.19 |
| 4 | main | 42.1 | 26.3 | 69.6 | 2.51 |
| 5 | branch | 41.0 | 26.2 | 56.8 | 2.02 |
| 6 | main | 42.6 | 25.1 | 64.4 | 2.20 |

Branch mean 26.9 tok/s / 59.0 ms per cycle vs main 25.0 / 63.7 — about +8% throughput and −7% per-cycle backbone at 206k, with the control column flat. MTP acceptance (tok/cycle) is stochastic at temperature 1.0 and moves tok/s by a comparable amount between rounds, so treat this as "a few percent, consistently in the right direction", not more. Cold 48k prefill is at parity (both builds gather the 2-D chunked prefill).

## Parity / tests

`tests/test_qwen4_qsa_decode_gather.py`: width gate (default 16, env override, garbage fallback); trimmed pooled prefix plus lazy tail is `array_equal` to a full recompute after trims that do and do not cross a block boundary; verify → `trim(2)` → decode is `array_equal` between the retained-prefix cache and a fully re-pooled reference. `tests/test_qwen4_qsa_incremental_cache.py` updated for the retained prefix. The compat test that asserted identical mRoPE planes are *ineligible* (written after the merge dropped the feature) is put back to the #3350 expectation.

Note: `test_qwen4_language_wrapper_routes_2d_text_positions_to_gather`, `test_qwen4_gathered_qsa_prefill_matches_official_mask_path` and `test_qwen4_lightning_mtp_fusion_and_runtime_attachment` fail identically on clean `main` in my environment (Python 3.12, mlx 0.32.2, native kernels built; 2e-5 tolerance vs ~1e-4 observed) and are untouched here.

## Test plan

- [x] `pytest tests -k "qwen4 or qsa or mtp or prefill_oom or memory_monitor or engine_preflight"` — only the three pre-existing failures above
- [x] Interleaved server A/B, upstream/main vs this branch, fresh process per round, 80 / 48k / 206k contexts, MTP on

https://claude.ai/code/session_01EJbV4uLny7wXtc7DVYPNtL

### DIFF
--- a/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
+++ b/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import logging
 import math
 import mmap
 import os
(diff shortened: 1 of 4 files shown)


jundot:
Thanks for restoring the gathered-QSA eligibility and preserving the pooled prefix across trims. I verified the cache parity tests. I'll resolve the conflict with main and merge this, then remove the added path-logging machinery and unused helper in a follow-up. The offset conversion currently forces GPU synchronization during batched decode even when logging is disabled.

Issue #2267, opened 2026-07-16
M5: test_fused_output_bit_exact[True] fails on clean main; fused gate+up sorted prefill returns garbage in 32 elements per block (#2238 follow-up)


richgoodson:
## Bug

`tests/test_qwen35_moe_gate_up.py::test_fused_output_bit_exact[True]` (new in #2238, commit 5a39ba3a) fails on Apple M5 on clean main. The `[False]` (unquantized) variant passes.

```
FAILED tests/test_qwen35_moe_gate_up.py::test_fused_output_bit_exact[True]
assert mx.array_equal(ref, out).item()  ->  False
```

## What I observed

I replayed the test body standalone to quantify the mismatch. The pattern is stable across runs:

- The decode path (1 token, unsorted branch) is bit-exact every time.
- The prefill path (40 tokens x top-2, the sorted branch) mismatches exactly 32 of 5120 output elements in each of the two blocks, every run.
- The magnitude of the mismatch varies run to run: max abs diff was 0.86 on one run, 0.39 on another, and 4.68e+36 on two others.

```
run 1: decode0:OK decode1:OK prefill0:BAD n=32 max=5.45e-01 prefill1:BAD n=32 max=8.61e-01
run 2: decode0:OK decode1:OK prefill0:BAD n=32 max=4.68e+36 prefill1:BAD n=32 max=3.94e-01
run 3: decode0:OK decode1:OK prefill0:BAD n=32 max=4.68e+36 prefill1:BAD n=32 max=3.94e-01
```

The test is also order-sensitive: it fails every time in isolation, but inside a full `pytest tests/` run it passed once and failed once for me. When it failed in-suite, the traceback pointed at the decode assert rather than the prefill one, so which comparison trips is not stable either.

## My read

Values like 4.68e+36 in an otherwise small-valued output, a fixed count of 32 bad elements, and run-to-run variation in magnitude look like an uninitialized or out-of-bounds read in the quantized `gather_qmm` sorted path with the fused (2x inter) weight width on M5, not accumulation-order drift. That is a guess; I have not isolated it to a pure MLX repro.

If it is real at model scale, M5 prefill outputs for Qwen3.5/3.6 MoE would be corrupted whenever the fusion is active (it is on by default in `BatchedEngine`). I have not live-tested a real model to confirm that. `OMLX_QWEN35_MOE_GATE_UP=0` would be the workaround.

[...]

PR #3432, opened 2026-09-04
fix(m5): segment oversized sorted gather_qmm calls instead of demoting them

ddark-il:
## Summary
- The M5 sorted-gather reroute (issue #2267) dropped `sorted_indices` for sorted rhs `gather_qmm` calls above 32768 rows to avoid the int16 row-offset overflow in the NAX rhs kernel (ml-explore/mlx#3856; fixed upstream by ml-explore/mlx#3922 only after 0.32.2 shipped).
- The unsorted gather path is ~8x slower per MoE layer on M5 Max, so any prefill chunk past the cap (4097+ tokens on top-8, 3277+ on top-10 / Qwen4-Exp) lost far more than the wider chunk saved.
- Keep the flag and issue oversized calls as balanced `<= 32768`-row sorted slices, concatenating the partial outputs. Only the layout the MoE sorted path produces (`[rows, 1, K]` activations + flat sorted `rhs_indices`) is segmented; `K % 64 != 0` and other layouts keep the existing demotion.
- Exposes `SEGMENTS_SORTED_ROWS = True` so a chunk-width policy can key off it. (A Qwen4 4096-chunk floor was A/B tested on top of this and found end-to-end neutral on M5 Max, so no scheduler change is proposed.)

## Measurements
`benchmarks/bench_m5_sorted_gather_chunk.py`, one Qwen4-Exp MoE layer (512 experts, top-10, 2560/640, q4 gs64, fused gate+up), M5 Max / mlx 0.32.2, ms per layer call:

| chunk | rows | sorted (raw, corrupt past cap) | unsorted (old reroute) | segmented (this PR) |
|---|---|---|---|---|
| 2048 | 20480 | 23.8 | 189.8 | 24.7 |
| 4096 | 40960 | 38.8 | 337.6 | 31.0 |
| 8192 | 81920 | 48.9 | 512.2 | 58.2 |

## Test plan
- [x] `tests/test_m5_gather_qmm.py`: segment bounds, wrapper splits positional/keyword `rhs_indices`, unaligned K and unknown layouts still demote, and on defective hardware a 36864-row sorted call matches the fp32 reference
- [x] `tests/test_qwen35_moe_gate_up.py`, `tests/test_deepseek_v4_patch.py`, `tests/test_qwen35_moe_weighted_sum.py` pass

## End-to-end (oMLX server, Qwen4-Exp 180B-A10B mixed AWQ, M5 Max 128 GB, 16k prompt, 3 interleaved rounds)
| arm | prompt tok/s |
|---|---|
| baseline, default 2048 chunks | 1196-1227 |
| baseline forced to 4096 chunks (old reroute demotes to unsorted) | **447-458** |
| this PR forced to 4096 chunks (segmented sorted) | 1259-1273 |

Per-chunk trace: a 3968-token chunk took 9.2 s on the old reroute vs 3.1 s segmented. Default Qwen4 chunks (2048) never cross the cap, so this is a cliff fix for wider chunk configurations rather than a default-path speedup.

https://claude.ai/code/session_01FUwVt6F1YHbTLvNvmAGEvU

### DIFF
--- a/benchmarks/bench_m5_sorted_gather_chunk.py
+++ b/benchmarks/bench_m5_sorted_gather_chunk.py
@@ -1,0 +1,135 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Per-layer MoE prefill cost vs chunk width on M5 (sorted gather_qmm paths).
+
+Reproduces one routed-expert MoE layer at a configurable geometry (default:
+Qwen4-Exp, 512 experts / top-10 / hidden 2560 / inter 640, 4-bit gs64 with
+the fused gate+up layout) and times the sorted SwitchGLU path for several
+prefill chunk widths under three dispatch modes:
+
+- ``sorted``      raw ``sorted_indices=True`` (NAX rhs kernel; corrupt past
+                  32768 rows on mlx <= 0.32.2, timed for reference only)
+- ``unsorted``    what the pre-segmenting M5 reroute did past the row cap
+- ``segmented``   the oMLX reroute with <=32768-row sorted slices
+
+Prints ms per layer call and derived tokens/s so the chunk-width policy can
+be judged on numbers. Run from a checkout with the reroute installed::
+
+    python benchmarks/bench_m5_sorted_gather_chunk.py --chunks 2048 4096 8192
+"""
+
+from __future__ import annotations
+
+import argparse
+import time
+
+import mlx.core as mx
+from mlx_lm.models.switch_layers import _gather_sort, _scatter_unsort
+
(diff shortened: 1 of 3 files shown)


jundot:
Thanks, splitting oversized sorted calls addresses the wide-chunk slowdown without changing the default chunk size. I checked the segment boundaries, positional and keyword arguments, and numerical parity on M3 Ultra; I couldn't independently rerun the M5-specific checks. I'll merge this and remove the unused capability flag in a follow-up.

PR #3434, opened 2026-09-04
fix(qwen35_prefill): drop the wrong NAX qmm tile variant 6, add a tile benchmark

ddark-il:
## Summary
- `OMLX_QWEN35_QMM_NAX_VARIANT=6` (64x64x64, wm=1/wn=4) returns wrong output on M5 Max / mlx 0.32.2: max abs error 5-8 against stock `quantized_matmul` at every tested shape, q4/q8, gs64/gs128, while variants 0-5 agree to a bf16 ulp. Remove the tile.
- Validate the env knob in Python: out-of-range or non-numeric values warn once and use variant 0 instead of raising from inside a prefill.
- Add `benchmarks/bench_qwen35_nax_qmm_variants.py` (stock vs each bundled tile at Qwen4-Exp GDN/attention projection shapes) and a hardware test covering every remaining tile.

## Measurements (M5 Max, ms per call, T=8192)
| shape | q8 stock | q8 v0 | q8 v2 | q4 stock | q4 v0 | q4 v2 |
|---|---|---|---|---|---|---|
| gdn.in_proj_qkv 10240x2560 | 18.97 | 18.14 | 15.01 | 8.14 | 7.99 | 8.96 |
| attn.q_proj 12288x2560 | 19.57 | 19.38 | 17.16 | 17.24 | 17.11 | 15.96 |
| attn.o_proj 2560x6144 | 10.06 | 10.17 | 9.76 | 8.46 | 8.58 | 8.13 |

Variant 2 (bm=128) wins 5-20% at 8-bit but is mixed at 4-bit, so the default stays at variant 0.

## Test plan
- [x] `tests/test_nax.py`: env validation cases; every bundled tile vs stock for q4/q8 x gs64/gs128 (52 pass)
- [x] `tests/test_qwen35_q4_mlp.py`, `tests/test_qwen35_ane_prefill.py`, `tests/test_qwen35_moe_weighted_sum.py` pass with the rebuilt extension

## End-to-end
Variant 0 vs 2 with the q8 route forced on at 2048 tokens: 1208-1223 vs 1204-1273 tok/s at 16k, indistinguishable. The PR is the variant-6 correctness fix plus tooling; no default change.

https://claude.ai/code/session_01FUwVt6F1YHbTLvNvmAGEvU

### DIFF
--- a/benchmarks/bench_qwen35_nax_qmm_variants.py
+++ b/benchmarks/bench_qwen35_nax_qmm_variants.py
@@ -1,0 +1,122 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Stock vs oMLX NAX qmm tiles for Qwen hybrid-layer projections on M5.
+
+The Qwen q4/q8 prefill-linear patch routes affine `QuantizedLinear` prefill
+matmuls through the bundled `qwen35_q{bits}_affine_qmm_t` op, which on M5
+selects a NAX tile chosen by `OMLX_QWEN35_QMM_NAX_VARIANT` (0 = the tile
+stock MLX ships). This benchmark times stock `mx.quantized_matmul` against
+each bundled NAX tile variant (``fast.NAX_QMM_VARIANTS``) at the projection shapes of a Qwen4-Exp GDN /
+attention layer, so the variant default and the q8 min-token threshold
+(`OMLX_QWEN35_Q8_LINEAR_MIN_TOKENS`, 16384) can be set from numbers.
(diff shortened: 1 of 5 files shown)


jundot:
Thanks for the fix. Removing tile 6 and validating the environment setting keeps the default path unchanged. I'll remove the standalone benchmark script in a follow-up since I don't plan to maintain the tile-tuning tooling, while keeping the correctness tests. LGTM.

Issue #3438, opened 2026-09-04
K2-Horizon MoVA model support


hermitdave:
**Labels:** enhancement, model support

---

oMLX can't load `IFM/K2-Horizon-MoVA-36B-A4B` because `k2_horizon` isn't in the patch dispatch table. Same issue `hy_v3` and `deepseek_v4` had.

When you drop the model into oMLX and try to load it:

```
ValueError: Model type k2_horizon not supported.
```

The dispatch in `maybe_apply_pre_load_patches()` never fires for this model type, so the adapter never registers. The fix is a three-line addition to `model_loading.py` plus a new patch directory — identical to the `hy_v3` patch that already exists.

**Model info:**
- Repo: https://huggingface.co/IFM/K2-Horizon-MoVA-36B-A4B
- Architecture: K2HorizonForCausalLM (MoE + MoVA, 36B total / 4B active)
- Adapter: community MLX adapter already working in the 4-bit MLX repo

**Worked around locally** by adding the patch directory and dispatch entry. Verified:
- oQ4e loads and generates correctly
- 6-bit uniform quantization converts successfully

Request: add native support so users don't have to patch by hand.

mbarnson:
I've been noodling on this since it came out, and put together a patch that should allow serving at least the MoVA cleanly. The "Uno" diffusion models remain tricky, in part because I did not figure out how to get them to honor a "thinking budget"... that's just not a concept that translates well to diffusion models. The 'speedup' of the Uno models is also debatable to nonexistent on my M4 Max MacBook Pro. On an RTX 6000 Pro or something, it makes sense: compute often sits idle while memory bandwidth is saturated. But at least on my Mac, the GPU is the weakest point.

I did a little experimentation with using ANE for Uno, but the results were disappointing enough -- about 1/3 as fast as GPU on my M4 Max -- that I pulled them out of my patch. 

That said, the MoVA is a fine and quite fast model on my fork of oMLX today. Not the best tool user, not the most knowledgeable, and no vision, but fast and fully open is pretty cool.

PR #3441, opened 2026-09-04
Add K2-Horizon model support

hermitdave:
## What this does

Let oMLX load `IFM/K2-Horizon-MoVA-36B-A4B`, a new open-source MoE model from IFM (36B total / 4B active params, MoVA attention). Currently oMLX refuses to load it because `k2_horizon` isn't in the patch dispatch table — same situation `hy_v3` and `deepseek_v4` were in before their patches landed. Issue #3438 

## What changed

**New file: `omlx/patches/k2_horizon/__init__.py`**
Registers the community MLX adapter into `sys.modules` before `mlx_lm.load()` runs. This is the exact pattern from the existing `hy_v3` patch — nothing novel here.

**New file: `omlx/patches/k2_horizon/k2_horizon_model.py`**
Straight copy of the working community adapter from [abenzerps/K2-Horizon-MoVA-36B-A4B-MLX-4bit](https://huggingface.co/abenzerps/K2-Horizon-MoVA-36B-A4B-MLX-4bit) (Apache-2.0).

**One-line addition to `omlx/utils/model_loading.py`**
Added a dispatch entry in `maybe_apply_pre_load_patches()`, right after the `hy_v3` block:

```python
if model_type == "k2_horizon":
    from ..patches.k2_horizon import apply_k2_horizon_patch
    if apply_k2_horizon_patch():
        logger.info("K2-Horizon pre-load patch applied for %s", model_name)
```

## Verified locally

- ✅ oQ4e quantization loads, generates coherent reasoning + answers
- ✅ 6-bit uniform conversion works (28 GB output)
- ✅ Model survives the full `trust_remote_code` prompt cleanly

## Notes

The adapter uses `sys.modules` injection — same approach as the `hy_v3` patch. If upstream mlx-lm ever adds native `k2_horizon` support, the patch auto-disables (it checks for the module first).

### DIFF
--- a/omlx/api/thinking.py
+++ b/omlx/api/thinking.py
@@ -22,12 +22,26 @@ _MINIMAX_OPEN_TAG = "<mm:think>"
 _MINIMAX_CLOSE_TAG = "</mm:think>"
 _HY3_OPEN_TAG = "<think:opensource>"
 _HY3_CLOSE_TAG = "</think:opensource>"
+# IFM / K2-Horizon think tags (from IFM chat template)
+_IFM_OPEN_TAGS = (
+    "<ifm|think>",
+    "<ifm|think_fast>",
+    "<ifm|think_faster>",
+)
+_IFM_CLOSE_TAGS = (
+    "</ifm|think>",
+    "</ifm|think_fast>",
+    "</ifm|think_faster>",
+)
 
 # Regex for non-streaming extraction (complete text)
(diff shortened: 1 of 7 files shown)


ajayre:
Would like to see this as well.

thinklikesqlhns:
Same here

noorforeverone:
Need this patch

Issue #2496, opened 2026-08-03
Kimi-Linear (kimi_linear): prefix-cache reconstruction omits KDA recurrent state → engine crash on every cache hit


rsdancey:
## Summary

Serving `Kimi-Linear-48B-A3B-Instruct-4bit` (`model_type: kimi_linear`), **every request that hits the prefix cache crashes the engine loop**:

```
ValueError: not enough values to unpack (expected 4, got 0)
  File "/Applications/oMLX.app/Contents/Resources/omlx/engine_core.py", line 394, in _engine_loop
    q_state, k_state, v_state, ssm_state = cache
```

The cache reconstruction path rebuilds KV blocks but does not restore the **KDA recurrent state**, so the hybrid linear-attention layers receive an empty cache and the 4-tuple unpack fails.

In practice this means: **turn 1 of any conversation succeeds; turn 2 always fails.** For agent workloads (tool call → tool result → continue) the model is unusable — every follow-up request errors and the client falls back to another provider.

## Environment

| | |
|---|---|
| oMLX | 0.5.5 |
| mlx-lm | 0.31.3 |
| mlx | 0.32.0 |
| macOS | 26.5.2 (25F84) |
| Hardware | Apple M4 Pro, 48 GB |
| Model | `mlx-community/Kimi-Linear-48B-A3B-Instruct-4bit` |

Model settings: `trust_remote_code: true` (required — the repo ships `configuration_kimi.py`), `guided_grammar_enabled: false`, `turboquant_kv_enabled: false`, `max_context_window: 40000`, `scheduler.max_concurrent_requests: 3`.

## Evidence — the crash fires exactly on a cache hit

Trimmed server log across the two turns (same conversation):

```
# turn 1 — succeeds
Chat completion: model=Kimi-Linear-48B-A3B-Instruct-4bit, 41 tokens in 32.94s (34.3 tok/s),
                 prompt: 13547, finish_reason=tool_calls

# turn 2 — same conversation plus the tool result
Chat completion request received: messages=4, stream=True
  Message[0]: role=system   ...
  Message[1]: role=user     ...
  Message[2]: role=assistant ...
  Message[3]: role=tool     ...
Added request 31532864-… with 13800 prompt tokens
prefix_cache    - Cache hit for 31532864-…: 6 blocks, 12288 tokens
paged_ssd_cache - Loaded block with metadata from SSD cache: aa83c60e430c1ba0…   (×6)
[...]

PR #3447, opened 2026-09-05
fix(scheduler): refuse prefix reuse for unreconstructible cache classes

### LINKED ISSUE #3445: Prefix-cache reuse silently downgrades unregistered cache classes to plain KVCache (reproducer: Unlimited-OCR)
TL;DR: oMLX's prefix-cache fallback reconstructs unregistered KV-like subclasses, such as Unlimited-OCR's `RingSlidingKVCache`, as plain `KVCache`, losing their custom retention state. In the image-free reproduction, the restored cache retains all decoded tokens instead of the native sliding window; separate server tests show warm requests reaching the 8192-token cap.

### Describe the bug

Unlimited-OCR uses `RingSlidingKVCache`: it retains the prompt and a bounded decode window while the absolute token offset continues increasing. In the tested oMLX checkout, cache-type dispatch falls back to `DefaultCacheHandler` and reconstructs this state as ordinary `KVCache`. Subsequent attention is no longer bounded by the native ring, even when the restored prompt tensors match.

Source inspection of upstream main at [`e467261e`](https://github.com/jundot/omlx/commit/e467261edc786efd33b1e9023d5c4a827f8aa1c1) on 2026-09-04 confirms that `RingSlidingKVCache` is absent from `CacheTypeRegistry._class_name_map` and no reconstructibility guard prevents this prefix-lookup path. The snippet below drives the registry directly, not an end-to-end server request; it isolates the reconstruction contract without weights or an image.

### Image-free reproduction

Run in an oMLX checkout with its MLX dependencies on Apple Silicon. No model weights, image, API server, or private evaluation data are needed.

```python
import json
import mlx.core as mx
from omlx.cache.type_registry import CacheTypeRegistry

from mlx_vlm.models.unlimited_ocr.language import RingSlidingKVCache

def append(cache, values):
    tensor = mx.array(values, dtype=mx.float32).reshape(1, 1, -1, 1)
    keys, _ = cache.update_and_fetch(tensor, tensor)
    return keys.reshape(-1).tolist()

native = RingSlidingKVCache(window_size=2)
append(native, [0, 1, 2, 3])
handler = CacheTypeRegistry.get_handler_by_class_name(type(native).__name__)
keys, values = handler.serialize_state(native)
restored = handler.reconstruct_cache(
    {"keys": keys, "values": values}, handler.serialize_meta_state(native)
)
same_prefix = bool(mx.array_equal(native.state[0], restored.state[0]))
for token in range(4, 12):
    native_keys = append(native, [token])
    restored_keys = append(restored, [token])

print(json.dumps({
    "handler": type(handler).__name__,
    "restored_type": type(restored).__name__,
    "identical_initial_prefix": same_prefix,
    "native_keys": native_keys,
    "restored_keys": restored_keys,
    "offsets": [native.offset, restored.offset],
}, indent=2))
```

Observed before the fix (abridged; float values shown as integers):

```text
handler: DefaultCacheHandler
restored_type: KVCache
identical_initial_prefix: true
prefix: [0, 1, 2, 3] | window: 2 | decoded tokens: 4 through 11

native ring: [0, 1, 2, 3,                   10, 11]  prompt + last 2
restored KV: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]  all history
offsets: [12, 12]
```

Expected: restoration preserves native retention semantics, or the hit is rejected and the prompt recomputed. Matching offsets alone is not sufficient.

This is not specific to Unlimited-OCR: unregistered KV-like classes can be structurally identified as ordinary KV, while class-name reconstruction falls back to `DefaultCacheHandler`. Custom metadata need not survive that fallback. Here the lost `window_size`, `prefill_length`, and `_ring_pos` change attention scope. Refuse unsupported reuse before lookup, not after the original class has already been replaced by plain KV.

I have a general fail-safe implementation available for a separate PR: refuse reuse for classes the reconstruction path cannot preserve, checked before lookup. Adding ring-cache capability is a separate follow-up, not the safety fix.

### Separate end-to-end observation

A local document-page evaluation used temperature 0, max_tokens 8192, the exact native task prompt, base image mode, and no n-gram guard. One page stopped cold at 3536 completion tokens but reached 8192 on repeated warm requests with 256 cached prompt tokens. The clean control changed from 900 tokens cold to 915 warm in the same cache-enabled run.

With prompt and image settings held fixed, a ring-preserving local patch made all three pages byte-identical across cold JSON, warm JSON, concurrent streaming, and a server restart: 12 requests, all ending with `finish_reason=stop`. Completion lengths were 1794 / 3536 / 900 tokens across these three fixtures.

All twelve outputs match within the paged-cache path, not against the same requests with paged caching disabled: the affected page produces 3536 tokens with paged caching enabled/cold versus 3825 with it disabled. Another page changes from 1794 to 1795. Block-aligned prefill and N-1 kickoff are candidate contributors; their individual effects have not been isolated. The prototype does not close this parity gap.

### Additional cache boundaries

The native class's inherited multi-row merge can also produce ordinary batched KV state. However, oMLX's existing singleton pass-through already preserves the native object: the scheduler installs singleton `filter`/`extract` methods on the base `KVCache`, which the ring subclass inherits. The singleton merge path then returns that object without calling its inherited batch conversion. This does not supply a correct multi-row ring implementation; serial admission is required by the proposed capability patch to avoid that unsupported conversion. Once the ring wraps, physical retained tensors must not be labeled as a complete contiguous history of all generated tokens. A correct storage path must retain the intact prefix or use a suitable native snapshot contract.

### Tested environment

- Image-free reproduction host: macOS 26.6.2, Apple M4 Max.
- oMLX implementation base: `e467261edc786efd33b1e9023d5c4a827f8aa1c1`.
- Python 3.12.11, MLX 0.32.2, mlx-lm 0.31.3, mlx-vlm 0.6.3, transformers 5.12.1.
- End-to-end cache replays: an 8-bit quantized derivative of `baidu/Unlimited-OCR`, native ring window 128. These cache replays were not repeated with BF16 weights.
- Tensor reproduction: synthetic four-token prefix, two-token window.

The image-free registry defect does not depend on checkpoint weights. Original BF16 weights were evaluated separately in the prompt/image-mode experiments; those results are not evidence of a BF16 server-cache replay.

### Related reports

Related oMLX report [#2496](https://github.com/jundot/omlx/issues/2496) describes Kimi-Linear recurrent state missing after prefix reconstruction. That is a different failure: the inspected `kimi_linear` implementation builds `ArraysCache` and `KVCache` layers, both already registered, so this class-registration guard permits that layout and does not address #2496. The report illustrates the other side of the cache contract: registration alone is not proof of a correct state round trip. Registered handlers still need per-handler round-trip tests.

Related architecture discussion: mlx-vlm [#1629](https://github.com/Blaizzy/mlx-vlm/issues/1629) and merged [#1638](https://github.com/Blaizzy/mlx-vlm/pull/1638). This report concerns oMLX's own scheduler/persistence path, not a claim that those native APC fixes failed.

Not a duplicate of oMLX [#2424](https://github.com/jundot/omlx/issues/2424): the maintainer attributes its raw byte-level BPE/CJK corruption to a malformed converted tokenizer and documents a runtime correction. That response has a warm prefix hit, but this alone does not establish a shared cause.

### Scope and non-goals

- **Reproduction boundary:** The tensor example proves retention loss, not the page-level OCR runaway. The server observations are separate evidence.
- **OCR accuracy:** All 12 patched replays stopped on these fixtures; this is neither a general stopping guarantee nor a claim of correct transcription.
- **Native parity:** Paged and unpaged outputs still differ as quantified above. Block alignment and N-1 kickoff remain hypotheses, not isolated causes.
- **Batching:** The warm-hit experiment isolates prefix reconstruction, not a singleton-merge or concurrent-request failure. Multi-row ring support remains a separate capability constraint.
- **Other inference controls:** This safety fix does not change prompt formatting, image preprocessing, sampling, or output limits.

contrapuntal:
### Summary

Refuse paged prefix lookup when a model builds cache classes without a recognized reconstruction path. This prevents an unregistered KV-like subclass from being silently restored as plain KVCache; unsupported layouts instead recompute the prompt.

Closes #3445.

- Probe the model's cache layout and check supported reconstruction class names.
- Preserve reuse for supported sliceable classes and registered handlers.
- Recursively inspect composite members, not just their outer wrapper.
- Refuse reuse when cache construction raises; do not memoize transient failures as safe.
- Check before lookup, including partial hits.

### Scope and trade-off

Unsupported layouts lose prefix reuse, avoiding the unregistered-class downgrade described in #3445. This does not add ring batching, new cache handlers, image controls, or sampling changes. Registration remains a trust boundary: each handler still needs round-trip tests. In particular, this guard does not address #2496, whose Kimi-Linear cache classes are already registered.

The check runs before lookup only; completion-time storage is unchanged. Unsupported layouts can therefore still write blocks they cannot reuse, wasting work and retaining those entries on SSD across restarts. Later registering a class makes those pre-registration entries eligible for lookup; the new handler must not silently accept incompatible old state.

A capability follow-up must invalidate those entries or reject them safely as cache misses. Alternatively, this PR could also skip storage for unsupported layouts. The current implementation keeps the guard minimal; maintainer feedback on extending it to storage is welcome. The migration requirement must be recorded when this guard lands, not discovered only when capability is restored.

The scheduler safety issue is addressed even though directly calling the unsupported registry fallback still yields a generic cache. Hardening that registry API separately need not block the fail-safe.

### Verification

- Upstream base `e467261e`: 243 scheduler tests passed before applying the change.
- Standalone PR tree: 1069 targeted scheduler/cache tests passed, 3 deselected. This includes the 10 added regression cases, cache handlers, singleton handling, paged/SSD storage, hybrid layouts, and GDN reconstruction.
- Confirmed the new partial-hit test fails against the original upstream lookup method: it incorrectly reuses two cached tokens. The test passes with this guard.
- Weight-free probes using real native `RingSlidingKVCache`, registered KV/Arrays layers, and a nested `CacheList` confirmed refusal/allowance and successful-probe memoization.
- An existing GDN test constructed `Scheduler` without a model. Its fixture now supplies its intended KV/Arrays cache layout; all existing walkback assertions remain unchanged.
- Syntax and diff-whitespace checks passed. Independent review found no blocking issues after the fixture adaptation and comment corrections.

The CI-style full suite stopped after three MiniMax compatibility failures: 5777 passed, 106 skipped, 77 deselected. Each failing case reproduces on unchanged upstream `e467261e` with the same environment (missing `_build_sparse_decode_indices`, missing `minimax_m3_vl.msa`, and unsupported `pack_shared_expert`). The full suite did not complete; this is not a full-suite pass. No unrelated compatibility fix is included.

Test host: macOS 26.6.2, Apple M4 Max, Python 3.12.11, MLX 0.32.2, mlx-lm 0.31.3, mlx-vlm 0.6.3. No model weights or OCR HTTP replay were used for this PR's verification; ring-capability replay results are not attributed to this guard.

Targeted command:

```sh
python -m pytest tests/test_scheduler*.py tests/test_cache_type_handlers.py tests/test_singleton_cache_passthrough.py tests/test_prefix_cache.py tests/test_prefix_cache_cachelist_mixed.py tests/test_prefix_cache_cachelist_per_member.py tests/test_rotating_cache_contract.py tests/test_store_cache_gate.py tests/test_paged_cache.py tests/test_paged_ssd_cache.py tests/test_hybrid_cache.py tests/test_cache_factory.py tests/test_cache_ntuple_state.py tests/test_prefix_cache_dedup_backfill.py tests/test_prefix_cache_gdn_split.py tests/test_prefix_cache_rotating_tip_strip.py -q
```

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -8188,12 +8188,109 @@ class Scheduler:
         )
         return True
 
+    # Legacy class-name aliases explicitly supported by reconstruction.
+    _EXTRA_RECONSTRUCTIBLE_CACHE_TYPES = frozenset({"SizedArraysCache"})
+
+    def _prefix_reuse_supports_cache_class(self, class_name: str) -> bool:
+        """True when ``class_name`` has a recognized reconstruction path.
+
+        Reconstruction dispatches on the stored ``type(cache).__name__`` through
+        ``CacheTypeRegistry``. Preserve existing handling for registered names
+        and explicitly supported sliceable classes. Registration is a trust
+        boundary, not proof that a handler preserves every field: handler-level
+        round-trip tests are still required.
+
+        An *unregistered* class is the problem. ``detect_cache_type`` falls back
+        to structural sniffing, so any ``KVCache`` subclass looks like a plain
+        ``KVCache``, gets ``DefaultCacheHandler`` (``supports_block_slicing =
+        True``) and is rebuilt as a vanilla ``KVCache`` with its extra state
+        dropped. Only that case is refused here.
+        """
+        if class_name in _KNOWN_SLICEABLE_CACHE_TYPES:
+            return True
+        if class_name in self._EXTRA_RECONSTRUCTIBLE_CACHE_TYPES:
+            return True
+        if HAS_CACHE_TYPE_HANDLERS and CacheTypeRegistry is not None:
+            return class_name in CacheTypeRegistry.list_known_class_names()
+        # Without the registry nothing beyond the plainly sliceable set can be
+        # rebuilt, so treat everything else as unsupported.
+        return False
+
+    def _cache_layer_is_reconstructible(self, layer: Any) -> bool:
+        """Recursive per-layer form of :meth:`_prefix_reuse_supports_cache_class`."""
+        if layer is None:
+            return True
+        # Test doubles carry no real cache contract.
+        if type(layer).__module__.startswith("unittest.mock"):
+            return True
+        sub_caches = getattr(layer, "caches", None)
+        if isinstance(sub_caches, (list, tuple)):
+            # CacheList itself round-trips, but only if every sub-cache does:
+            # its handler stores sub-class names and re-dispatches per name.
+            if not self._prefix_reuse_supports_cache_class(type(layer).__name__):
+                return False
+            return all(self._cache_layer_is_reconstructible(sub) for sub in sub_caches)
+        return self._prefix_reuse_supports_cache_class(type(layer).__name__)
+
+    def _model_has_unreconstructible_cache(self) -> bool:
+        """True when ``model.make_cache()`` builds cache layers of a class the
+        paged prefix round trip silently downgrades.
+
+        Unlimited-OCR's ``RingSlidingKVCache`` subclasses ``KVCache`` and is not
+        in the registry, so it is sniffed as a plain ``KVCache``, rebuilt as one,
+        and loses ``window_size`` / ``prefill_length`` / ``_ring_pos``. The model
+        then attends over an unbounded history, which can alter generation.
+        The pre-existing rotating guard cannot catch it:
+        it runs only for *exact* prefix hits and inspects the *already
+        reconstructed* cache, which is a plain ``KVCache`` by then.
+
+        Probed once per scheduler and memoized. A probe that raises is **not**
+        memoized and refuses reuse for that request, so a transient failure can
+        never be recorded as "safe" and then silently reused.
+        """
+        cached = getattr(self, "_unreconstructible_cache_model", None)
+        if cached is not None:
+            return cached
+
+        try:
+            layers = make_prompt_cache(self.model) or []
+        except Exception as e:
+            logger.warning(
+                "Could not probe model cache classes (%s: %s); refusing prefix cache "
+                "reuse for this request rather than risking a silent downgrade",
+                type(e).__name__,
+                e,
+            )
+            logger.debug("Cache probe failure traceback:", exc_info=True)
+            return True
+
+        result = False
+        for layer in layers:
+            if not self._cache_layer_is_reconstructible(layer):
+                logger.info(
+                    "Prefix cache reuse disabled: model builds %s layers, which "
+                    "have no cache-type handler and would be rebuilt as plain "
+                    "KVCache, dropping their state",
+                    type(layer).__name__,
+                )
+                result = True
+                break
+
+        self._unreconstructible_cache_model = result
+        return result
+
     def _prepare_prefix_cache_for_request(self, request: Request) -> None:
         if request.request_id in self._prefix_cache_prepared:
             return
 
-        # Check prefix cache for cached KV state
-        if self.block_aware_cache is not None:
+        # Check prefix cache for cached KV state. Models whose cache layers do
+        # not survive reconstruction skip the lookup entirely and fall through
+        # to full prefill below — including on partial hits, which bypass the
+        # exact-hit guard further down.
+        if (
+            self.block_aware_cache is not None
+            and not self._model_has_unreconstructible_cache()
+        ):
             # Use paged cache
             block_table, remaining = self.block_aware_cache.fetch_cache(
                 request.request_id,
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -111,6 +111,18 @@ class MockModel:
         """Return model parameters."""
         return self._parameters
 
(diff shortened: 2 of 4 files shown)


jundot:
Thanks for adding this guard. I verified that an unregistered cache class could reuse a partial prefix on the old path, and that this change rejects it while preserving reuse for supported classes. LGTM, merging. I’ll follow up by skipping storage for unsupported layouts, removing the mock-specific exception from the probe, and trimming the comments and docstrings.

Issue #2164, opened 2026-07-10
Performance Regression: Qwen3.6-27B-PARO decoding speed dropped from 22-28 tok/s to 11-13 tok/s (0.5.0dev4 → 0.5.0rc1)


williamxie1989:
## Description

After upgrading from oMLX 0.5.0dev4 to 0.5.0rc1, the decoding speed of the Qwen3.6-27B-PARO model dropped significantly — roughly half of what it was before.

## Environment

- **Hardware**: MacBook with Apple M5 Max
- **Model**: Qwen3.6-27B-PARO
- **oMLX versions compared**:
  - 0.5.0dev4: ~22-28 token/s
  - 0.5.0rc1: ~11-13 token/s

## Performance Data

| Version | Decoding Speed (token/s) |
|---------|-------------------------|
| 0.5.0dev4 | 22 – 28 |
| 0.5.0rc1 | 11 – 13 |

**~50% performance regression**

## Reproduction Steps

1. Load Qwen3.6-27B-PARO with oMLX 0.5.0dev4, measure decoding speed → ~22-28 tok/s
2. Upgrade to oMLX 0.5.0rc1, same config → ~11-13 tok/s

## Notes

- Same machine (M5 Max MacBook), same model and quantization config
- The drop is consistent at ~50%

Was there any known performance-related change between 0.5.0dev4 and 0.5.0rc1? Happy to provide more info if needed.

rwedoff:
Same here, and same with the new Qwen3.8-27B-PARO, the performance is about 10 tok/s

rwedoff:
[liang2kl](https://github.com/liang2kl) Can you help?
@liang2kl

PR #3448, opened 2026-09-05
fix(vlm): preserve Unlimited-OCR ring caches through prefix reuse

contrapuntal:
### Summary

On top of the general guard, restore Unlimited-OCR prefix reuse with a tested reconstruction contract. This re-enables a path the guard deliberately disables; it is not the minimal fix for silent downgrades.

Builds on merged [#3447](https://github.com/jundot/omlx/pull/3447). Rebased onto upstream `aa8db734`; the diff now contains only the ring-cache follow-up, not the guard commit. This base also includes the follow-up that skips storage for unsupported cache layouts.

- Add an oMLX-owned native-ring subclass with safe singleton interfaces; reject lossy multi-row conversion.
- Declare the N-1 prefill end explicitly so a single-token prompt chunk cannot start the decode ring early; cover both normal and chunked scheduler prefill.
- Serialize Unlimited-OCR admission while other model engines retain batching.
- Register a ring handler that exports only intact prompt KV and reconstructs a fresh ring with valid prompt-boundary bookkeeping.
- Limit stored token sequences to the exported prefix before allocating blocks; refuse missing ring layers and malformed K/V state, including a missing first-layer state that could otherwise bypass validation.
- Reject legacy ordinary-KV OCR hits, decoded/malformed metadata, inconsistent ring windows, and tensor lengths that disagree with block token counts. Validate every block so deduplication cannot hide an unsafe trailing capture behind valid first-block metadata.
- Centralize registered ring-family checks for storage, reconstruction, extraction, and TurboQuant exclusion; retain exact-class checks for adapted-only interfaces and decline TurboQuant conversion if the registry is unavailable.

### Trade-offs and scope

Concurrent Unlimited-OCR requests queue. This is a per-model throughput regression, not server-wide serialization; the existing Llama 4 ChunkedKVCache serial-admission path is a precedent. This is deliberately not a generic batched-ring implementation. Decoded ring slots are not persisted as a full token history.

No image-mode API, prompt normalization, or image-identity version bump is included. Migration relies on live layout signatures, adapter compatibility checks, and per-block validation that rejects unsafe persisted entries as cache misses. It does not depend on a separate image-controls PR landing first.

Existing N-1 prompt kickoff and block-aligned prefill are unchanged: the final prompt token enters the decode ring rather than the permanent prefix. Exact parity with a native full-prompt generation path remains a follow-up. Serialization relies on the per-engine single-worker executor and an admission cap counting both running and prefilling requests; it does not make direct concurrent Scheduler.step() calls safe. Forced model-ownership transfer resets the previous owner's scheduler but does not stop its engine loop.

### Verification on the guarded split

- 1392 targeted scheduler/QSA/cache/adapter/engine/ownership tests passed, 3 deselected, for `148b959a`, including 121 ring tests. The ring test module passes Ruff and `git diff --check` is clean. This is a local targeted result, not a full-suite CI claim.
- The six-item review follow-up adds real SSD-to-decode coverage with unequal K/V feature widths (192/128), ring wrap, and MLX attention; contextual block/allocation/metadata validation with scheduler fallback and zero-reference cleanup tests; registered ring-family checks; N-1 boundary and serialization documentation; and actionable scheduler-test timeout diagnostics. The missing-allocation fixture injects a lookup failure, not a real deallocation race.
- Three independent reviews accepted the six-item follow-up; the final import-spacing correction and ownership wording clarification are included. No fresh production OCR replay or native full-model parity run was performed for this follow-up. CI for the newly pushed commit is tracked separately.
- Added 68 malformed-state cases spanning missing/null state, invalid pair structure, non-tensor entries, wrong ranks, mismatched K/V dimensions, multi-row state, and empty prefixes. These exercise both layer positions with and without model metadata and verify rejection before request-table or block allocation. Four valid round-trip controls preserve list/tuple payloads and different K/V feature widths.
- Fixed the two QSA reservation integration failures introduced by the prefill-boundary follow-up: both prefill entry points now invoke the static boundary helper through `Scheduler`, preserving lightweight scheduler callers. The previously omitted QSA integration file is included in the passing targeted suite. GitHub CI passed on Python 3.11–3.13 for `7c9fa19c`. That historical CI result does not verify the newly pushed six-item follow-up; its CI status is tracked separately.
- The missing-ring-layer regression failed with `IndexError` before the bounds check and now verifies that storage returns `None` without allocating a request table or referenced/hashed blocks.
- New regressions cover exact hits, one- and two-token suffixes, single-token chunk tails, and chunked continuation through 132 generated tokens with a 128-token ring. Five scheduler cases failed on the premature boundary before the correction and pass afterward. Exact-hit and one-token-suffix controls retain their existing behavior.
- Before the prefill-boundary follow-up, the CI-style local full suite completed with 10842 passed, 155 skipped, 77 deselected, and 10 failures. Four MiniMax compatibility failures reproduced on unchanged upstream `aa8db734` with the same installed dependencies. The other six were caused by an inherited `OMLX_API_KEY`; rerunning both affected settings/server files without that variable passed all 279 tests. The pre-follow-up GitHub CI passed on Python 3.11–3.13. These are historical full-suite results, not a green local full-suite claim for the new follow-up; fresh GitHub CI is separate.
- Verified the guard refuses the native cache before registration and allows the adapted cache after registration.
- Real SSD persistence/reopen tests cover plain-KV legacy entries, decoded native-ring metadata, mixed deduplicated prefill/decode chains, and malformed tensor lengths. Real scheduler/BatchGenerator execution with a tiny weight-free model verifies full-prefill fallback, EOS completion, subsequent safe warm reuse, and zero remaining block references.
- Valid partial and extended deduplicated prefixes retain native prompt-plus-window behavior. Singleton methods, serial admission, trim behavior, and TurboQuant exclusion are covered.
- Independent review found and reproduced mixed-chain and tensor-length gaps. Both now have failing-before/passing-after regressions; re-review found no remaining high-severity issues.
- Static review of the prefill-boundary follow-up inspected scheduler integration, mask behavior, eviction retries, marker lifetime, and reconstruction, but missed the QSA integration failures addressed above; review does not substitute for executable tests.

### Controlled HTTP replay

Repeated all twelve requests with the prefill-boundary correction on 2026-09-06: three pages across cold JSON, warm JSON, concurrent SSE, and a process restart on the same SSD each finished with `stop`. Each page's output was byte-identical across all four conditions and matched the earlier replays, with completion lengths 1794 / 3536 / 900. Cold requests had zero cached tokens; every later request reused 256. The three simultaneous clients queued serially. Restart recovered three SSD files with zero incompatible entries; no request warnings or errors appeared.

Replay used an 8-bit Unlimited-OCR derivative, temperature 0, cap 8192, no repetition guard, and configured concurrency 8. A process-local harness held preprocessing at base mode (`cropping=False, base_size=1024, image_size=1024`) and applied the default template from [mlx-vlm #2164](https://github.com/Blaizzy/mlx-vlm/pull/2164). These controls are outside this patch; they isolate cache behavior and do not verify default gundam behavior.

The replay establishes consistency within the guarded paged path, not OCR accuracy or native unpaged parity. All outputs also match the older combined prototype, but its paged-versus-unpaged numerical differences remain out of scope.

Test host: macOS 26.6.2, Apple M4 Max, Python 3.12.11, MLX 0.32.2, mlx-lm 0.31.3, mlx-vlm 0.6.3. Both temporary servers were stopped; no production deployment was performed.

### DIFF
--- a/omlx/cache/prefix_cache.py
+++ b/omlx/cache/prefix_cache.py
@@ -822,6 +822,19 @@ class BlockAwarePrefixCache(CacheManager):
                 layer_state.get("meta_state", ()) for layer_state in cache_data
             ]
 
+        if is_tensor_data:
+            # The OCR handler exports only intact prompt KV, not overwritten
+            # decode slots. Do not label it with the full generated sequence
+            # or allocate blocks past its valid prefix (including partials).
+            for idx, type_name in enumerate(layer_cache_types or []):
+                if type_name in ("RingSlidingKVCache", "OMLXRingSlidingKVCache"):
+                    keys = cache_data[idx]["state"][0]
+                    if keys is None:
+                        return None
+                    tokens = tokens[: keys.shape[2]]
+            if not tokens:
+                return None
+
         split_gdn_layout = self._gdn_split_layout_supported(layer_cache_types)
 
         # Get or create block table
@@ -3377,6 +3390,41 @@ class BlockAwarePrefixCache(CacheManager):
                     cache_type_name = layer_cache_types[layer_idx]
 
                 handler = CacheTypeRegistry.get_handler_by_class_name(cache_type_name)
+
+                if cache_type_name in (
+                    "RingSlidingKVCache", "OMLXRingSlidingKVCache"
+                ):
+                    # A deduplicated chain may start with a valid prefill
+                    # capture and end with old overwritten decode-ring slots.
+                    # Validate every block, not just the first block's metadata
+                    # passed to reconstruct_cache for ordinary sliceable KV.
+                    window = None
+                    prefix_end = 0
+                    for block_idx, data in enumerate(all_block_data):
+                        metadata = all_block_meta_states[block_idx]
+                        block_window, stored_length = handler.validate_prefix_metadata(
+                            metadata[layer_idx] if metadata else None
+                        )
+                        keys, values = data[layer_idx]
+                        expected_length = self.paged_cache.blocks[
+                            block_table.block_ids[block_idx]
+                        ].token_count
+                        if (
+                            keys is None
+                            or values is None
+                            or keys.ndim != 4
+                            or values.ndim != 4
+                            or keys.shape[:3] != values.shape[:3]
+                            or keys.shape[0] != 1
+                            or keys.shape[2] != expected_length
+                        ):
+                            raise ValueError("Invalid Unlimited-OCR prefix block length")
+                        prefix_end += expected_length
+                        if stored_length < prefix_end or (
+                            window is not None and block_window != window
+                        ):
+                            raise ValueError("Incompatible Unlimited-OCR prefix blocks")
+                        window = block_window
 
                 # === CacheList: dedicated branch (before standard 2-tuple unpack) ===
                 if cache_type_name == "CacheList":
--- a/omlx/cache/prefix_cache.py
+++ b/omlx/cache/prefix_cache.py
@@ -827,7 +827,7 @@ class BlockAwarePrefixCache(CacheManager):
             # decode slots. Do not label it with the full generated sequence
             # or allocate blocks past its valid prefix (including partials).
             for idx, type_name in enumerate(layer_cache_types or []):
-                if type_name in ("RingSlidingKVCache", "OMLXRingSlidingKVCache"):
+                if CacheTypeRegistry.is_ring_family(type_name):
                     if idx >= len(cache_data):
                         # Refuse incomplete captures before allocating blocks.
                         return None
@@ -3405,9 +3405,7 @@ class BlockAwarePrefixCache(CacheManager):
 
                 handler = CacheTypeRegistry.get_handler_by_class_name(cache_type_name)
 
-                if cache_type_name in (
-                    "RingSlidingKVCache", "OMLXRingSlidingKVCache"
-                ):
+                if CacheTypeRegistry.is_ring_family(cache_type_name):
                     # A deduplicated chain may start with a valid prefill
                     # capture and end with old overwritten decode-ring slots.
                     # Validate every block, not just the first block's metadata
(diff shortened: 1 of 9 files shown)


jundot:
Thanks for adding the ring-cache reconstruction support. I checked cold, warm, concurrent, and post-restart requests with Unlimited-OCR-oQ8: all five outputs matched through 180 generated tokens, with 512 cached tokens reused after the cold run. The legacy SSD rejection and reference-release tests also passed.

I haven’t found a blocking issue in the current changes. The serial admission trade-off and remaining native unpaged parity work are clear from the description. I’ll leave this open and take another look once the draft is ready.

contrapuntal:
Rebased onto `aa8db734` after #3447 merged; the PR now contains one signed ring-cache commit, with no implementation changes from the reviewed patch. This also verifies against the upstream follow-up that skips storage for unsupported layouts.

All 1194 targeted tests and all 12 controlled cold/warm/concurrent/restart HTTP replays passed again. Per-page outputs match the pre-rebase replay. The PR description records the local full-suite failures and their upstream/environment controls; fresh CI is running.

Marked ready for review. Serial admission and native unpaged parity remain scoped as described.

contrapuntal:
A follow-up weight-free scheduler probe found a prefill-boundary case not covered by the earlier page replay. Returning this to draft while it is addressed.

With an eight-token prompt and the native 128-token ring window, cold execution establishes the existing N−1 boundary at seven tokens. An exact prefix hit trims to seven and matches cold output through 132 generated tokens. However, a six-token prefix hit leaves two tokens: external prefill processes one and generation kickoff processes the other. The native cache mistakes that first single-token prefill call for decode, establishes the boundary at six, and later generated output diverges. A one-token final prefill chunk produces the same effect on the cold path.

The original one-token-suffix/exact-hit diagnosis was therefore too broad. The verified problem is treating a one-token *prefill* call as decode. A two-token kickoff alone would not fix it because the scheduler splits it again. The fix will retain the current N−1 serving behavior and distinguish prefill from decode explicitly, with long-enough cold/warm and chunked regressions. Existing native full-prompt parity remains separate.

contrapuntal:
Fixed the prefill-boundary case in `702a9ab5`. Both scheduler prefill paths now declare the N−1 end explicitly, so single-token prompt chunks append contiguous KV instead of starting the native decode ring. Exact-hit trimming and persisted metadata are unchanged.

The regression suite covers exact hits, one- and two-token suffixes, singleton chunk tails, and chunked continuation through 132 generated tokens with a 128-token ring. Five scheduler cases failed before this change and now pass. All 37 ring tests and 1204 targeted tests pass; all 12 controlled cold/warm/concurrent/restart HTTP replays stop and match the prior outputs.

Independent follow-up review found no remaining blockers. Restoring ready-for-review status; fresh CI will validate the pushed commit.

Issue #3435, opened 2026-09-04
oQ: the float32 rule for vision passthrough tensors is broader than the evidence behind it (measured on GLM-5.3)


pedroberaldo87:
### Summary

`_cast_passthrough_tensor` in `omlx/oq.py` forces every vision and audio passthrough tensor to
float32 whenever the requested dtype is float16. The rule came from #1682 (fixing #1678), where
the symptom was real — infinite `<pad>` with image input on **Gemma 4 31B / 26B-A4B, M1 Max** —
but the cause was stated as a hypothesis and never measured: the issue says activations "easily
exceed the FP16 maximum range (~65,504)", and the PR's only evidence is `206 passed`. The guard
test (`tests/test_oq.py`, `test_float16_keeps_vision_audio_passthrough_tensors_float32`) only
asserts the dtype, so it locks in the rule without checking the premise.

The rule is applied to every family. On GLM-5.3-Flash I measured the premise, and it does not
hold there — it costs ~1.1 GB of resident memory for no numerical benefit.

### Measurement

The GLM-5.3-Flash vision tower alone (347 tensors, 2.1 GB, loaded straight from the oQ shards),
run over 6 real screen captures plus one saturated synthetic gradient, in float32, float16 and
bfloat16. Cosine and relative L2 are per image token against the float32 output; the peak is the
largest activation value seen in any vision block (instrumented on `Glm5NextVisionBlock.__call__`).

```
image                          cos_med F16   cos_med BF16   relL2 F16   relL2 BF16   peak act
dsh-cfg-oq2e.png                 0.999980       0.999604     0.00669      0.02919      227.9
dsh-config-modelo.png            0.999982       0.999639     0.00620      0.02859      227.9
dsh-e2e-01.png                   0.999963       0.998942     0.00908      0.04850      228.1
dsh-e2e-0901-mtp-turno2.png      0.999969       0.999278     0.00838      0.03941      226.8
page-2026-08-31T17-54-47.png     0.999948       0.998349     0.01067      0.05953      230.0
page-2026-08-31T18-03-58.png     0.999889       0.997407     0.01546      0.07622      226.0
synthetic saturated gradient     0.995414       0.919799     0.11293      0.43684      217.0

no inf/nan in float16 on any of the 7 · median relL2 ratio bf16/f16 = 4.7x
```

Two things follow:

1. **Peak activation is ~228 against the 65504 float16 ceiling** — a factor of 287. There is no
[...]

PR #3453, opened 2026-09-05
perf(vlm): load the glm5_next vision tower in float16 (the oQ float16 recipe stores it in float32)

pedroberaldo87:
## What

Implements what #3435 measured. The oQ float16 recipe leaves the vision/audio passthrough tensors in float32 because of the #1682 rule, which was written for Gemma 4 on a hypothesis and never measured. On `glm5_next` float16 is measured 4-5x **closer** to float32 than bfloat16 is (the dtype the model was trained in and that every non-oQ build serves), and the float32 copy costs ~1.1 GB resident (347 tensors: 2.25 GB on disk, 1.12 GB in float16).

At load, `cast_vision_f32_params_to_fp16` recasts the float32 leaves of the vision tower to float16 in the same bounded chunks as `materialize_lazy_state`, handing the float32 copies back to the allocator as it goes. Scoped to `glm5_next`; `OMLX_VISION_FP16=0` keeps the checkpoint's dtype.

## Measured

GLM-5.3-Flash oQ2e on M1 Ultra: resident 104.92 → 103.18 GB on load. Vision-feature outputs on real screenshots, token by token against the float32 tower: float16 within fp16 rounding, bfloat16 4-5x further. Details and numbers in #3435.

## Tests

`tests/test_cast_vision_fp16.py` (only float32 leaves under the vision subtrees are recast; the count is returned; nothing else is touched).

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -23,6 +23,7 @@ Usage:
     vision API format is transparently handled.
 """
 
+import os
 import asyncio
 import contextlib
 import copy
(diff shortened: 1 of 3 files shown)


pedroberaldo87:
Heads-up on ordering with #3459: both PRs edit the same region of `omlx/utils/model_loading.py`. A `git merge-tree` between the two branches reports a conflict there (neither branch conflicts with main on its own).

They are independent in purpose and both are worth having: this PR shrinks the vision tower by loading it in float16, while #3459 drops it entirely for text-only use. Suggested order: merge this one first, then #3459 with a trivial conflict resolution in that block.

PR #3455, opened 2026-09-05
perf(qwen4_exp): reserve QSA indexer capacity for known prefill lengths

williamxie1989:
# perf(qwen4_exp): reserve QSA indexer capacity for known prefill lengths

## Summary

The QSA indexer cache grows with the general-purpose "double when short" rule, even though a prefill's final token count is known before its first chunk runs. The doubling therefore fires in the middle of prefill — at the deepest point of the request, where memory headroom is smallest — and each doubling costs a reallocation, a zero-fill and a full copy while the old buffer is still alive.

This PR reserves the capacity the request will actually need, once, at the start. Behavior for requests whose length is not known is unchanged: growth still doubles.

## Root cause

`_ensure_indexer_capacity` grows the indexer cache by stepping up until it covers the current need, which for a cache below the needed size means doubling. For a prefill the needed size is not a guess: the prompt length is known before the loop starts. Measured on a Qwen4-exp hybrid model (48 layers, 12 of them full attention, `max_position_embeddings = 262144`) during a 205k-token prefill, one chunk added **+12.25GB** of resident memory at `kv = 196,608` — the moment with the least headroom. In steady state the cache had been grown to a capacity of **393,216 tokens, larger than the model's own maximum context of 262,144**, because doubling overshoots and there is no upper bound tied to the horizon.

## Change

Model side (`mlx_vlm/models/qwen4_exp/language.py`):

- `reserve_index_capacity(tokens)` records the horizon for the caches a request will use.
- `_next_capacity(current, needed, step, reserve=None)` replaces the doubling *only* when a reservation exists: the target is the reservation, or the need when the cache has already passed it, rounded up to the step. With no reservation the code path is the original stepped growth, so unknown-length requests behave exactly as before.
- The pooled block bank takes the same treatment through `reserve // compress_ratio`.

Scheduler side (`omlx/scheduler.py`): `Scheduler._reserve_qsa_index_capacity(cache_list, tokens)` sets the reservation on every cache that supports it (caches without the method are skipped, so other model families are untouched) and logs one INFO line. It is called before the external-prefill loop with `base_size + remaining`, and in the chunked path on the first chunk with `state.total_length or base + remaining`.

## Measurements

Setup: 128GB Apple Silicon, Qwen4-exp hybrid chat model as above, weights 69.6GB resident. Single request, cold server, identical oMLX configuration in each pair; only the presence of this change differs. Metrics are per-request: `peak phys` = max `phys_footprint` sampled during the request; `Δswap` = delta of `sysctl vm.swapusage` `used` across the request, so the machine's starting state is irrelevant; `wall` = request round trip.

**Cold 205k prompt:**

| | without this change | with this change |
|---|---|---|
| outcome | request rejected with an empty `usage` payload; model unloaded by the memory guard | completed |
| `peak phys` | 102.17GB | **92.37GB** (Δ −9.80GB) |
| indexer cache resident | 10.46GB (capacity 393,216 > 262,144 max context) | **6.77GB** (exactly the reservation) |
| `Δswap` | **+6.2GB** | ≈ 0 |

**Warm 212k prompt (a prefix hit, so the restored cache starts below the reservation):** 115s → **99s**, `cached_tokens = 98,304`, and no memory-guard activity in the log either way at that length.

The warm case is what shaped the fix. An earlier revision only reserved when the cache was empty, so a restored cache below the reservation still fell through to doubling — measurable as +5.87GB of physical growth during the request. The regression test is named for that: `test_restored_cache_below_the_reservation_lands_on_it_not_double`, and `test_unreserved_growth_still_doubles` pins the unchanged unknown-length behavior.

## Testing

`tests/test_qwen4_qsa_reserved_capacity.py` — 7 tests over the capacity arithmetic: an unreserved cache doubles as before; a reservation lands on the rounded step; a need above the reservation wins; a cache already past the reservation is left alone; a restored cache below the reservation lands on the reservation rather than doubling; the pooled block bank scales by the compression ratio; the scheduler helper skips caches that do not implement the hook.

Full fast suite: 267 relevant tests pass. One failure in `test_qwen4_qsa_decode_gather.py` reproduces identically on unmodified cache code, so it is pre-existing and unrelated. `black`/`ruff` report the same findings as the base commit — the new lines add none.

## Relation to other work

- **#3417** (open, three admission-charge bugs in the prefill transient tracker) is a different mechanism on the accounting side; it neither prevents nor reduces the doubling. The two are complementary and touch disjoint files.
- **#3428** and **#3334** (open) modify the same vendor `language.py` for gathered-QSA eligibility and compiled decode respectively. No functional overlap — neither touches indexer capacity — but expect a textual conflict in that file when they merge.
- Measured here as a prefill-time cost, the same indexer cache is also a *retained* cost; this PR bounds the transient overshoot, not the steady-state size, which remains whatever the context length requires.

### DIFF
--- a/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
+++ b/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/language.py
@@ -159,7 +159,20 @@ class _QSAIndexerCache:
         self._index_position_ids = None
         self._index_offset = 0
         self._index_capacity_managed = True
+        self._index_reserved_tokens = 0
         self._invalidate_pooled_indexer()
+
+    def reserve_index_capacity(self, tokens: int) -> None:
+        """Hint the final sequence length so capacity lands on it once.
+
+        Capacity doubling is the right policy when the final length is unknown
+        (decode). A prefill knows it up front, and doubling there costs a full
+        reallocate + memset + copy of the indexer mid-prefill and leaves up to
+        2x the needed capacity resident — past ``max_position_embeddings`` for
+        long prompts. Reservation makes the first grow land exactly on the
+        stepped horizon and switches later grows back to plain steps.
+        """
+        self._index_reserved_tokens = max(0, int(tokens))
 
     @property
     def index_keys(self):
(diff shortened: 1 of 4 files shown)


williamxie1989:
Companion to #3456. The two address different failure stages of the same long-context goal (120k-262k multi-turn on a 128GB machine without swap and with the SSD cache off): this one removes the transient overshoot *during* prefill, #3456 bounds the checkpoints *retained* across prefill. Independent branches, disjoint files, either can merge first.

jundot:
Thanks for the memory measurements. Reserving QSA index capacity from the known prompt length makes sense and avoids unnecessary growth during prefill. I’ll merge this and follow up with two fixes: including the cached prefix in the chunked-prefill reservation and honoring the reservation on the first allocation after cache restoration. In my MLX cache repro, those fixes reduced reallocations from 14 to 1 while preserving the cached data.

PR #3456, opened 2026-09-05
feat(cache): budget retained boundary checkpoints

williamxie1989:
# feat(cache): budget retained boundary checkpoints

## Summary

Hybrid models can only resume a shared prefix from a sequence boundary where the **recurrent state** was captured, so `Scheduler` captures a checkpoint at **every** block boundary during prefill. That makes the retained-prefix cost grow linearly with context *on top of* the attention KV, and on the models I measured the checkpoint is the dominant term.

This PR adds a budget for those checkpoints. The budget is enforced **at capture time**, so a boundary that will not be retained is never extracted, and boundaries that fall out of the budget are freed **while prefill is still running**. The newest boundary is always retained. Default (`cache.boundary_snapshot_retention = 0`) preserves current behavior exactly.

## Motivation (measured)

Machine: 128GB Apple Silicon. Model: a Qwen4-exp hybrid chat model with 36 GDN linear-attention layers + 12 full-attention layers, `max_position_embeddings = 262144`, weights 69.6GB resident, `paged_cache_block_size = 2048`, `cache.hot_cache_only = true` (SSD cache off). Single request, cold server, serial.

Metric definitions used below — all of them are per-request deltas or maxima, so they do not depend on the state the test machine happened to be in: `peak phys` = max `phys_footprint` sampled during the request; `Δswap` = delta of `sysctl vm.swapusage` `used` across the request; `resident snapshots` = tensor bytes of the boundaries recorded so far, sampled per chunk under `OMLX_PREFILL_MEM_PROBE=1`; pressure events = occurrences of guard rejection / hot-cache shrink / store skip in the log.

A 234,070-token prefill recorded **113 checkpoints of 115.6MB each — 12.06GB, held for the whole prefill**. Per prompt token that is **52.5KB**, versus **27KB for the entire live attention cache**: the retention cost is nearly twice the thing it is caching, and it is resident before any prefix reuse has happened. It is also what pushes a long-context request into the memory guard once the hot cache holds anything at all.

Single 261,075-token prefill, identical configuration, differing only in the budget:

| | `retention = 0` | `retention = 4` | Δ |
|---|---|---|---|
| `resident snapshots` | 13.68GB (127 × 115.6MB) | **0.43GB (4)** | **−13.25GB** |
| `peak phys` | 99.12GB | **85.63GB** | **−13.49GB** |
| `peak MLX active` | 91.9GB | 78.91GB | −13.0GB |
| `Δswap` | ≈ 0 | ≈ 0 | — |
| pressure events | 0 | 0 | — |
| `wall` | 234s | 239s | +2% |

What that headroom buys — same box, hot cache raised to 8GB, one conversation, three consecutive turns (prompts are synthetic, generated by script):

| turn | prompt tokens | `cached_tokens` | `wall` | `Δswap` |
|---|---|---|---|---|
| T1 | 100,067 | 0 (cold) | 90s | ≈ 0 |
| T2 | 212,068 | **98,304** | 101s | ≈ 0 |
| T3 | 234,069 | **210,944** | **24s** | ≈ 0 |

For comparison, with `retention = 0` on the same model the same three turns report `cached_tokens = 0` for all three and T3 takes 207s — and the hot cache has to be limited to 2GB to avoid the memory guard at all. **The reason long-context multi-turn could not reuse prefixes is not that the prefix cache is broken; it is that keeping every checkpoint leaves no room to store the prefix.** Pressure events across the run: 0.

## Why the budget is safe

Restores do not need a dense checkpoint set, and the existing code already assumes it:

- The store path refuses to persist a non-sliceable block without its checkpoint, in its own words: *"storing live non-sliceable state would corrupt later prefix hits"*. So a pruned boundary cannot become a half-populated block — pruning is exactly the supported degradation path, not a new one.
- A restore lands on the deepest retained checkpoint at or below the depth it needs and re-prefills the gap; the restore path already reports how far it walked back (`walkback_blocks`, `chosen_endpoint_tokens`).

That guard protects *correctness*, not *hit rate*, and the first revision of this PR learned the difference the hard way. It spaced the lattice over the newest captured boundary, which for a decoding request runs past the end of the prompt; the store filters to the cacheable sequence, so widening the stride off that moving point evicted the one in-range boundary the store could use. A store then refused outright (`reason=boundary_snapshot_unavailable available_boundaries=1`) and the following turn re-prefilled a 127,493-token prompt that matched its predecessor token for token. Budgeting checkpoints is only safe while the deepest *storable* boundary keeps its checkpoint; the design below makes that an invariant rather than a coincidence.

Verified end to end rather than argued: a 40,031-token base turn, then a **divergence at ~30k where the checkpoint had been pruned**, greedy (`temperature=0`, 48 tokens), compared byte-for-byte against the same prompt run with cross-turn caching disabled.

| request | `cached_tokens` | `wall` | output vs cache-disabled reference |
|---|---|---|---|
| diverged (first) | 16,384 (nearest retained lattice point) | 17s | **identical** |
| diverged (repeat) | 32,768 | 3s | **identical** |
| append (no divergence) | 38,912 (previous turn's deepest storable boundary) | 7s | — |

## Design

`omlx/cache/boundary_retention.py` holds the placement rule; `Scheduler` applies it in both recorders (`_on_prefill_boundary_snapshot` for prefill, `_maybe_capture_boundary_snapshot` for boundaries crossed while decoding — gating only the first lets a long generation re-accumulate what prefill released).

Decisions worth reviewing, each forced by a measured failure:

1. **Enforce at capture, not at store.** Deciding at store time still pays the extraction and keeps every checkpoint resident for the entire prefill, which is the cost being optimized.
2. **The lattice lives in absolute coordinates** (`tc % step == 0`), not as a rank over the boundaries kept so far. Rank-based placement collapses: once the prompt outgrows the survivors, every slot snaps onto the same one, and I measured a single gap spanning **98% of the prompt** — which re-prefills everything the budget was for.
3. **The stride widens by powers of two**, which keeps successive lattices nested: a point on the `2·stride` lattice was always on the `stride` lattice, so growing the prompt never deletes a checkpoint the wider lattice still wants. A non-power-of-two widening is not nested and reproduces the same 98% hole.
4. **Two boundaries are retained outside the budget**, because the store cannot do without them: the deepest recorded boundary *at or below the cacheable depth* (the depth a growing conversation resumes from, and the only boundary its store can use for its deepest block) and the deepest recorded boundary overall (which the store reads as `latest_tc`). A budget that spends the resume point is not a budget, it is a cache disable.
5. **The lattice is laid out over the cacheable depth, not over the newest capture.** For the prefill recorder the two coincide; for the decode recorder they do not, because it captures on `request.num_tokens` — prompt plus output — while the store filters to `cacheable_sequence` (the prompt alone when `needs_think_prefix`, matching the store's own condition). Anchoring on the newest capture therefore lets reasoning output and long generations widen the stride and delete in-range checkpoints. Shrunk to `block_size = 4` in `test_decode_captures_beyond_the_prompt_keep_the_resume_boundary`, which is red on the previous anchoring and green on this one.
6. **Uniform spacing** minimizes the worst-case recompute gap for a fixed budget, matching the placement used elsewhere for this problem (vLLM #45238 fix (a): "retain K checkpoints per request, spread across the prompt", K configurable with a small default of 2–4; the balanced-placement result in arXiv 2605.05219; llama.cpp #19408 is the fixed-interval variant).

`keep <= 0` means "retain everything", and the value is capped at `MAX_RETAINED_BOUNDARIES = 64`.

## Testing

- `tests/test_boundary_retention.py` — 10 tests over the placement rule: default retains everything; cost bounded for prefill lengths up to 131,072 boundaries; newest always retained; interior spacing uniform so no gap dominates; **incremental pruning over 512 arriving boundaries leaves one even step** (this is the test that caught the collapse in both naive variants); never empty; budget capped; the resume boundary and the store's `latest_tc` both survive a newest capture far above the cacheable limit.
- `tests/test_scheduler.py::TestSchedulerBoundarySnapshots::test_decode_captures_beyond_the_prompt_keep_the_resume_boundary` — drives the decode-side recorder through a real `Scheduler` with `needs_think_prefix` set, captures two boundaries past the end of the prompt, and asserts the deepest in-range boundary is still recorded. This is the one that was missing before: every policy-level test stayed green while the call site anchored on the wrong quantity, so a policy test alone could not have caught it.
- `tests/test_settings.py` — round-trip persistence of the new key. This is not boilerplate: `CacheSettings` persists through an explicit `to_dict`/`from_dict` pair rather than `**data`, so the first version of this PR recorded `4` in `settings.json` while the scheduler read `0` — a feature that is installed and inert, invisible in every unit test and only visible on a real run.
- Full fast suite: 10,636 passed. The 8 failures that remain are pre-existing; 7 reproduce on an unmodified `origin/main` worktree, and the 8th (`test_prefill_snapshot_decoupled_from_live_cache`) was a genuine regression from this PR — it drives `Scheduler` with a `SimpleNamespace` config, so the new field is now read through `getattr`, matching the existing convention for duck-typed configs.
- Real-machine numbers in the tables above, from the same binary with the budget flipped. Verified again under an agent-shaped workload (real repository files as tool results, a few hundred tokens generated per turn, next request issued immediately) with the hot cache at 86-99% of its limit and prompts of 252,703 / 260,915 tokens: hits landed on the previous turn's deepest storable boundary every time.

## Out of scope / honest limits

- With a budget of 4 on a 260k prompt the interior lattice step is 65,536 tokens, so a branch that diverges early re-prefills from the nearest lattice point. Measured rather than extrapolated: a 260,946-token turn whose history was edited ~1/3 of the way in resumed at `cached_tokens = 65,536` and re-prefilled 195,410 tokens (256s), where the turns that merely appended resumed at 249,856 and 251,904 and took 21s and 27s. That gap is the inherent price of the budget and is why the value is configurable: 16 retains every 16,384 tokens (~1.9GB of resident checkpoints at 115.6MB each) and cuts the same divergence to a ~20k-token re-prefill. It is strictly better than the alternative today, which is no reuse at all.
- **A larger budget is not better reuse, because the budget competes with the cache
  it protects.** Measured on the same 250k-class agentic transcript: with `keep = 4`
  an appending turn resumed at 251,904 of 253,612 tokens (99.3%) and swap did not
  grow; with `keep = 16` the same protocol missed outright twice on arrival
  (`reused 0`, one run adding 3.1GB of swap), and when the same request was retried
  90s later it resumed only at 81,920 (32%) because the deep blocks had been
  reclaimed meanwhile. At 115.6MB per checkpoint the retained set is 0.46GB at 4 and
  1.85GB at 16, and every store also carries that payload, so the extra checkpoints
  consume both the memory and the store bandwidth the KV chain needs to stay
  resident. The small default is deliberate; anyone raising it should re-measure
  reuse depth, not just the worst-case recompute gap.
- The attention KV is unchanged, and nothing here writes to disk; `hot_cache_only` configurations are the primary beneficiary.
- #3336 compresses *each* checkpoint while this reduces *how many*, so on the GDN term the two are substitutes rather than multiplicative. Measured retained-prefix cost at 234k: 18.7GB today, 12.6GB with that codec alone, **6.9GB with this budget alone**, 6.6GB with both. If only one of the two can land first, this is the larger lever for RAM-only long-context serving; that codec remains the right fix for absolute bytes on the SSD/split path.
- **#2628** (open, tracked under the #2625 large-context investigation) attacks the same
  bytes from the accounting side: it adds the *future* boundary-snapshot bytes to the
  prefill admission estimate so the guard can refuse earlier. That is a different remedy
  for the same cost, and it composes — this PR lowers the number their estimator has to
  reserve for. The per-checkpoint figure measured above (115.6MB at `block_size 2048` on
  a 36-layer GDN Qwen4-exp model) is the constant that estimate needs.
- **#3335** (open) gates the same recorder, `_maybe_capture_boundary_snapshot`, on a
  different criterion: it skips captures the store's alignment filter can never select,
  so a reasoning-model turn stops paying extraction for snapshots that cannot be used.
  This PR budgets how many *usable* checkpoints are retained. The two are independent
  predicates over the same call site, so expect a textual conflict there, not a semantic
  one.
- **#2366** (open) stores the large fp32 snapshot states as int8, i.e. it shrinks each
  checkpoint rather than reducing how many are kept — the same substitute relationship
  described for #3336 above.
- #1887 proposed bounding checkpoints but scoped it to cross-turn tips; this extends the budget to the interior boundaries captured during prefill, where the resident cost actually lives. #3439 lowers the *production rate* by widening `block_size`, which is orthogonal and composes with this.

### DIFF
--- a/omlx/cache/boundary_retention.py
+++ b/omlx/cache/boundary_retention.py
@@ -1,0 +1,87 @@
+"""Retention policy for per-boundary recurrent-state checkpoints.
+
+Hybrid models (GDN / Mamba / KDA linear-attention layers) can only resume a
+shared prefix from a sequence boundary where the recurrent state was captured,
+so oMLX captures a checkpoint at every block boundary during prefill. That makes
+the retained prefix cost grow linearly with context *on top of* the attention
+KV, and the checkpoint is by far the larger term: measured on
+Qwen3.8-Flash-Next (36 GDN layers, block_size 2048), one 234,070-token prefill
+captured 113 checkpoints of 115.6MB each -- 12.06GB resident for the whole
+prefill, 52.5KB per prompt token against the 27KB of the entire live attention
+cache. On a 128GB machine that resident term is what pushes long-context
+requests into the memory guard.
+
+Restores do not need that density. A restore lands on the deepest retained
+checkpoint at or below the depth it needs and re-prefills the gap: the store path
+already refuses to persist a non-sliceable block without its checkpoint ("storing
+live non-sliceable state would corrupt later prefix hits"), so checkpoints only
+trade memory against gap recompute -- which makes them the right thing to budget.
+
+The budget is spent as an even lattice over the prompt, because uniform spacing
+minimizes the worst-case recompute gap for a fixed budget. It is applied while
+prefilling (the final length is unknown, so the lattice is re-spread over the
+range reached so far each time it overflows) and always keeps the newest
+boundary: an append-only multi-turn conversation resumes from that one alone,
+while the interior ones serve requests that diverge inside the stored prefix.
+"""
+
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+#: Retaining more than this many checkpoints is past the point where the
+#: placement argument holds: the budget exists to bound a per-request resident
+#: cost, and 64 checkpoints of a hybrid model are already gigabytes.
+MAX_RETAINED_BOUNDARIES = 64
+
+
+def select_spaced_boundaries(
+    recorded: Sequence[int], limit: int, block_size: int, keep: int
+) -> set[int]:
+    """Return the boundaries of ``recorded`` that keep a checkpoint.
+
+    ``recorded`` is the ascending boundary token counts captured so far for one
+    request. ``limit`` is the depth the request's cache can still be *stored*
+    and resumed at: the end of prefill while prefilling, and the cacheable
+    prefix while decoding (see the call sites). ``block_size`` is the block size
+    the boundaries are aligned to. The budget ``keep`` counts interior
+    checkpoints; in addition to those the function always retains two
+    boundaries that the caller cannot afford to lose:
+
+    * the deepest recorded boundary at or below ``limit`` -- the depth the next
+      turn of a growing conversation resumes from, and the only boundary the
+      store path can use for its deepest block;
+    * the deepest recorded boundary overall, which is the one the store reads as
+      ``latest_tc``.
+
+    ``keep <= 0`` retains everything, i.e. the pre-existing behavior.
+
+    ``limit`` rather than the newest capture is the reference point because a
+    decoding request keeps capturing boundaries past the prompt, and while
+    speculative decoding and reasoning output push that newest capture further
+    out, anchoring the lattice on it widens the stride and evicts the in-range
+    boundary the next turn needs -- which presents not as a shallower hit but as
+    no hit at all, and on a store whose only surviving boundary is out of range,
+    as a refused store for the whole request.
+    """
+    budget = 0 if keep < 0 else min(int(keep), MAX_RETAINED_BOUNDARIES)
+    if budget == 0 or not recorded or block_size <= 0:
+        return set(recorded)
+
+    # Widening by powers of two keeps the lattices nested: a point on the
+    # ``2 * stride`` lattice was always on the ``stride`` lattice, so growing the
+    # prompt never deletes a checkpoint the wider lattice still needs. Any other
+    # widening leaves holes -- measured, one gap spanning 98% of the prompt.
+    stride = 1
+    while stride * block_size * budget < limit:
+        stride *= 2
+    step = block_size * stride
+
+    within = [tc for tc in recorded if tc <= limit]
+    retained = {tc for tc in within if tc % step == 0}
+    if within:
+        retained.add(within[-1])
+    retained.add(recorded[-1])
+    return retained
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -47,6 +47,7 @@ from mlx_lm.models.cache import (
 )
 from mlx_lm.sample_utils import make_logits_processors
 
+from .cache.boundary_retention import select_spaced_boundaries
 from .cache.observability import BoundarySnapshotDiagnostics, CacheRateTracker
 from .cache.paged_cache import PagedCacheManager
 from .cache.pooling_delta import compact_pooling_cache_snapshot
@@ -1568,6 +1569,9 @@ class SchedulerConfig:
 
     # Paged cache settings (internal defaults)
     paged_cache_block_size: int = 256  # Tokens per block
+    # Interior recurrent-state checkpoints retained per prefill, newest excluded
+    # (it is always retained). 0 = keep every boundary.
+    boundary_snapshot_retention: int = 0
     max_cache_blocks: int | None = (
         None  # Auto-calculated from available KV cache memory
     )
@@ -6435,6 +6439,38 @@ class Scheduler:
             )
             return
 
+        # Budget the retained checkpoints before capturing anything. A restore
+        # lands on the deepest retained checkpoint and re-prefills the gap, so
+        # capturing a boundary that will not be retained costs 115MB of resident
+        # state (measured on a hybrid GDN model at block_size 2048) and buys
+        # nothing: the store path will not persist a non-sliceable block without
+        # its checkpoint anyway. Deciding here also frees the boundaries that
+        # fall out of the budget as prefill runs, which is what bounds the
+        # long-context resident peak.
+        # Duck-typed configs (tests, embedded harnesses) carry only the fields
+        # they care about; an absent budget means the historical behavior.
+        retention = getattr(self.config, "boundary_snapshot_retention", 0)
+        recorded = self._boundary_cache_snapshots[request_id]
+        if retention > 0:
+            retained = select_spaced_boundaries(
+                sorted([*recorded, token_count]),
+                token_count,
+                block_size,
+                retention,
+            )
+            if token_count not in retained:
+                self._boundary_snapshot_diagnostics.record(
+                    "capture_skipped",
+                    reason="retention_thinned",
+                    request_id=request_id,
+                    token_count=token_count,
+                    block_size=block_size,
+                    source="prefill",
+                )
+                return
+            for tc in [tc for tc in recorded if tc not in retained]:
+                del recorded[tc]
+
         # Offload snapshot to SSD if store is available, keeping only a
         # None marker in the dict.  Falls back to in-memory storage when
         # the SSD store is unavailable or the write fails.
@@ -6847,6 +6883,32 @@ class Scheduler:
 
         if request.request_id not in self._boundary_cache_snapshots:
             self._boundary_cache_snapshots[request.request_id] = {}
+
+        # Same budget as the prefill-side recorder: this path captures at every
+        # block boundary crossed while decoding, so without it a long generation
+        # re-accumulates the checkpoints prefill just released. The lattice is
+        # laid out over the cacheable prefix, not over the emitted count: a
+        # decoding request captures past the end of the prompt, and those depths
+        # are not storable (the store filters to the cacheable sequence), so
+        # letting them set the stride evicts the in-range boundary the next turn
+        # resumes from -- which costs every subsequent hit.
+        retention = getattr(self.config, "boundary_snapshot_retention", 0)
+        if retention > 0:
+            recorded = self._boundary_cache_snapshots[request.request_id]
+            if getattr(request, "needs_think_prefix", False):
+                limit = len(request.prompt_token_ids)
+            else:
+                limit = total_tokens
+            retained = select_spaced_boundaries(
+                sorted([*recorded, total_tokens]),
+                limit,
+                block_size,
+                retention,
+            )
+            if total_tokens not in retained:
+                return
+            for tc in [tc for tc in recorded if tc not in retained]:
+                del recorded[tc]
 
         # Offload to SSD with in-memory fallback.
         if self._boundary_snapshot_store is not None:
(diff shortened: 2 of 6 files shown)


williamxie1989:
Companion to #3455 (transient prefill overshoot). This PR is the retained-bytes half: it releases 13.25GB of resident checkpoints on a 261k prefill, which is what makes room to store the prefix at all - with it set, three consecutive turns of 100k/212k/234k report cached_tokens 0/98,304/210,944 and the last turn drops 207s -> 24s. Relationship to #2628, #3335, #2366 and #1887 is spelled out in the body.

williamxie1989:
**Why this is draft again.** A real agentic workload (multi-turn, a few hundred reasoning tokens generated per turn, prompt-only caching) lost *every* hit: a turn stored its prefix with `available_boundaries=1`, the next turn matched all 126,976 comparable tokens and still re-prefilled 127,493.

The first revision spaced the lattice over the newest captured boundary. A decoding request captures on `request.num_tokens` = prompt + output, while the store filters to `cacheable_sequence`, so output tokens widened the stride and evicted the only in-range boundary — the depth the next turn resumes from. Budgeting checkpoints is safe only while that boundary keeps its checkpoint, so it is now retained outside the budget together with the store's `latest_tc`, and the lattice is laid out over the cacheable depth (`len(prompt_token_ids)` under `needs_think_prefix`, the store's own condition) instead of the newest capture.

My earlier multi-turn runs could not surface this: they generated a handful of tokens per turn, so the newest capture happened to *be* the resume point. `test_decode_captures_beyond_the_prompt_keep_the_resume_boundary` is the test that reproduces it at `block_size = 4` — red on the previous anchoring, green on this one. Policy-level tests stayed green through the whole thing, which is the lesson I am taking from this.

Full suite: 10,641 passed, 7 pre-existing local failures unchanged. Restoring to ready for review once the real-machine run below passes: three turns generating several hundred tokens each, asserting the later turns still report non-zero `cached_tokens` and no store refusal.

williamxie1989:
**Restore fidelity, measured at 250k.** `logprobs` is not implemented by this server's API, so the probe is greedy agreement on an identical 253,612-token prompt: a fresh full prefill and, immediately after it, a request that restores 251,904 tokens from the stored blocks and prefills only the tail (9s vs 317s) -- **byte-identical output**. Together with two back-to-back restore runs agreeing character for character, that puts the earlier cold-vs-warm text difference (first disagreement at character 449 of 1121) on the walkback path, i.e. chunk-layout numerics across a 195k-token recompute, not on checkpoint contents.

Also added a measured counter-example to the limits: `keep = 16` collapses reuse at this size, which is why the default is small.

jundot:
Thanks for the detailed measurements. I understand the memory pressure from retaining boundary snapshots with SSD caching disabled. I’m not planning to add sparse checkpoint retention at this point, since it reduces partial-prefix reuse and adds complexity to cache storage and restoration. For memory-constrained setups, I’d prefer using the existing SSD cache, which offloads boundary snapshots instead of retaining them all in RAM.

williamxie1989:
**Repositioning this PR: where the blessed knobs land, and where they bottom out**

Thanks for the direction, @jundot — I agree SSD offload is the right default answer, and I don't want this PR to compete with it. Let me put numbers on how far the existing knobs get us, because they get further than I gave them credit for, and then show the two places where they structurally can't reach.

**1. The density ladder, upstream knobs only.** Boundary cost per token is `snapshot_bytes / block_size`, so `block_size` and the sidecar codec already form a four-step ladder (36-layer GDN model, 115.6 MB fp32 / 57.8 MB rht_int16 per snapshot; KV+indexer is 27.9 KB/tok in every row, untouched):

| split config | GDN sidecar bytes/token | walkback |
|---|---|---|
| block 2048, fp32 (today's default) | 55.3 KB | ≤ 2048 |
| block 2048, rht_int16 | 27.7 KB | ≤ 2048 |
| block 4096, rht_int16 | 13.9 KB | ≤ 4096 |
| block 8192, rht_int16 | 7.1 KB | ≤ 8192 |
| block 2048, rht_int16, `retention=4` (this PR) | 1.1–1.9 KB | ≤ len/4 |

The ladder's first four rows use only knobs the codebase already has and already blesses — the scheduler itself enlarges block_size for ArraysCache hybrids "to reduce boundary snapshot overhead," and there's an auto-raise to 4096 on ≥64 GB non-NAX machines. I'm now running the split + rht_int16 + wider-block config as my default posture, and I'd encourage the docs to spell this ladder out; it solves the memory-pressure case exactly the way you described.

**2. Where the ladder structurally bottoms out.** Two places, both measured on this machine:

- **`hot_cache_only=true` users.** With the SSD tier disabled there is no offload target, and embedded snapshots are unbounded: 112 × 115.6 MB = 12.06 GB resident for the whole 234k prefill (Stage A probe). Retention is the only lever that exists in that mode — `retention=4` measured 0.43 GB resident, peak phys −13.49 GB, wall +2%.
- **Long-context flatness.** Sidecar bytes scale as `len / block_size`: at 260k, block 8192 + rht_int16 still writes 1.85 GB of GDN state per stored prefix. Retention makes it `O(keep)` — 0.29 GB at 260k, flat with length. Wider blocks only divide the slope; they don't remove it.

**3. What this PR becomes under your constraint.** Opt-in, default-off, and scoped narrow:
- `retention=0` (the default) is byte-identical to current behavior — the complexity is invisible until someone asks for it.
- The policy is one pure module (`boundary_retention.py`, table-tested) plus two thin call sites; if upstream ever grows its own frontier/lattice mechanism, this is a clean removal, not a surgery.
- I could scope it further to only arm when the SSD tier is disabled — i.e. it activates exactly where the blessed mechanism is absent, and never competes with offload.

**4. Reuse cost, honestly.** Walkback is coarser at long context: `retention=4` on 260k spaces the lattice at 65,536, and a turn whose history was edited ~1/3 in re-prefilled 195,410 tokens (256s) where appending turns resumed at ≤251,904 in ~21–27s. `keep=16` collapsed reuse entirely at this size in my runs — which is why the default is small and the knob is not for everyone. Appending multi-turn workloads — the common case — lose nothing: the deepest storable boundary is retained outside the budget.

**5. The ask.** Would a version of this that arms only when `hot_cache_only=true` (SSD offload unavailable) be acceptable to land? If the answer is still no, I'll keep it as a documented local patch — but I'd like the record here to show the trade-off is measured, and to get your steer on whether the block-size ladder is the intended long-term answer for the hot-only case as well.

PR #3458, opened 2026-09-05
feat(scheduler): the hybrid model's cache block becomes a panel setting (hybrid_cache_block)

pedroberaldo87:
## What

**Merge order:** this PR builds on the same function as #3407 (which makes 512 the glm5_next default). Merge #3407 first; this one then needs a trivial conflict resolution in `_enlarge_block_size_for_arrays_cache`. With #3407 in, 'automatic' means 512 for glm5_next; without it, the scheduler's existing target.

Hybrid (recurrent + attention) models get their paged-cache block size **and** prefill step from one panel setting, `scheduler.hybrid_cache_block`, in Settings › Resource: **0 = automatic** (today's behaviour), 512, 1024 or 2048. The block and the step move together so chunk boundaries stay aligned with the paged cache whether or not the SSD cache is on. Applies to the next model load; the route rejects any other value with a 400.

Why a setting and not an env var: the menubar app does not pass environment variables to the server, so the block was effectively unreachable for anyone not launching from a shell.

## Measured

M1 Ultra, GLM-5.3-Flash oQ2e (101 GB resident), cold 110k-token prompt, clean server start per arm:

```
block   prefill      chunks                 cuts
 512    163 tok/s    187 x 512              0
1024    181 tok/s     94 x 1024             0
```

Above 16k tokens of context the per-chunk transient is *smaller* (285-412 MB against ~1 GB below 16k), so the memory argument that keeps a small block at short context does not hold at long context. At 7k tokens the same change is +5-7% for +6 GB of peak — which is why the default stays automatic and the choice is the user's.

## Tests

`tests/test_hybrid_cache_block.py`: default is 0 and round-trips through `SchedulerSettings`; the setting drives both `paged_cache_block_size` and `prefill_step_size`; zero keeps the automatic target. Reverting the scheduler hunk makes the tests fail (3 failed). Settings round-trip in `tests/test_settings.py`.


pedroberaldo87:
Field note from running two hybrid models on the same server (M1 Ultra, 128 GB): **this setting is global, but the right value is per-model.** It costs real throughput on the model that is not the one you tuned for.

Measured today, same machine, one model loaded at a time, clean server start per arm, two rounds each:

```
GLM-5.3-Flash oQ2e (glm5_next)   110k cold prefill
  block 512    163 tok/s
  block 1024   181 tok/s      <- what we set the server to

Qwen3.8-Flash-Next oQ3.5e (qwen4_exp)
              block 1024 (from this setting)   block 0 = automatic (4096)
  30k         466.4 tok/s   12 x 1024          551.4 / 551.4 tok/s   7 x 4096   +18.2%
  59k         459.8 tok/s    2 x 1024          540.9 / 540.1 tok/s  14 x 4096   +17.6%
```

The mechanism: `_detect_qwen35_prefill_floor()` returns 4096 for `qwen4_exp` when the native QSA kernel is present and the machine has >= 64 GB, but the prefill chunk is clamped to the next paged-cache block boundary (`scheduler.py`, `next_boundary = ((current_total // block_size) + 1) * block_size`), so a 1024 block caps the chunk at 1024 and the floor never applies. Setting the block globally for one model silently disables the other model's own tuned width.

Not proposing a change to this PR — the setting as written is still an improvement over an env var the menubar app cannot pass. Recording it because a per-model override (the direction #3384 takes for the step) would close the gap, and because anyone serving two hybrid families should know the coupling exists.

One more consequence worth a line in the docs: changing the block invalidates every SSD cache block already written at the old size. On this machine that silently stranded 57.71 GB of a 92 GB budget for the GLM (403 blocks) and 4.99 GB for the Qwen (188 blocks) — the cache still counts them until LRU eviction. #3397 adds the startup warning that surfaces it.

PR #3459, opened 2026-09-05
feat(vlm): unload the glm5_next vision tower for text-only deployments (opt-in, off by default)

pedroberaldo87:
## What

An opt-in for agents that never send images. After `glm5_next` loads, `free_vision_tower(model)` replaces every leaf under `vision_model` / `vision_tower` with a one-element placeholder of the same dtype — the module tree stays intact, so introspection keeps working — and the memory goes back to the allocator. On GLM-5.3-Flash that is **347 tensors, 1.05-1.12 GB**. While the tower is out, a chat request carrying an image is refused with a `ValueError` that names the setting and how to turn it off.

**Off by default.** `ServerSettings.vision_tower_text_only` (Settings › Advanced toggle), or `OMLX_VISION_TOWER_TEXT_ONLY=1`; the CLI exports the saved setting so the engine reads one source. Applies to the next model load. Independent of #3453 (float16 tower): that one shrinks the tower, this one removes it.

## Measured

M1 Ultra, GLM-5.3-Flash oQ2e (101 GB resident), cold 229,920-token prompt, block 1024, guard target 112.5 GB, clean server start per arm:

```
                      prefill      pauses   chunks              footprint @197k
tower resident        170.5 tok/s  2        1024 -> 256 cuts    112.9 GB (target met)
tower unloaded        175.3 tok/s  0        224 x 1024, whole   111.4 GB
```

```
vision tower unloaded (text-only mode): 347 tensors, 1.05 GB returned
```

The miss with the tower resident was ~0.4 GB near 180k tokens; removing 1 GB of base line is what turned the two pauses into none.

## Tests

`tests/test_vision_tower_text_only.py`: frees only the vision subtree (language weights untouched), returns the tensor/byte count, placeholders keep the tree intact; setting is off by default and round-trips. Reverting the `model_loading.py` hunk makes the test error out. Settings round-trip in `tests/test_settings.py`.

### DIFF
--- a/omlx/admin/i18n/en.json
+++ b/omlx/admin/i18n/en.json
@@ -528,6 +528,8 @@
   "settings.advanced.sse_keepalive_mode_chunk": "Chunk (default, OpenClaw / WorkBuddy compatible)",
   "settings.advanced.sse_keepalive_mode_comment": "Comment (legacy : keep-alive)",
   "settings.advanced.sse_keepalive_mode_off": "Off (no keepalive)",
+  "settings.advanced.vision_tower_text_only": "Unload the vision tower (text-only mode)",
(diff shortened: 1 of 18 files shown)


pedroberaldo87:
Heads-up on ordering with #3453: both PRs edit the same region of `omlx/utils/model_loading.py`. A `git merge-tree` between the two branches reports a conflict there (neither branch conflicts with main on its own).

They are independent in purpose and both are worth having: #3453 shrinks the vision tower by loading it in float16, while this one drops it entirely for text-only use. Suggested order: merge #3453 first, then this one with a trivial conflict resolution in that block.

PR #3465, opened 2026-09-05
fix(qwen4_exp): correct long-context prefill memory accounting

mvdbos:
> [!WARNING]
> This PR is stacked on and depends on #3461. Please merge #3461 first.

## Summary

Improve Qwen4-Exp long-context prefill memory accounting so the scheduler can use available context capacity without weakening protection against real dense-attention peaks.

## Scope and layering

#3461 introduces the Qwen4-Exp prefill memory profile and gathered-core estimates used here. This PR builds on those foundations; it corrects how runtime routing, allocator measurements, reclaim, and chunk sizing use them rather than redefining the profile.

Runtime QSA route attribution, flat-overhead accounting, reclaim debt, and chunk re-expansion are enabled only when that `qwen4_exp` profile is active. Non-Qwen models retain their existing EWMA accounting, even if a cache exposes a similarly named route marker.

The SDPA256 correction is intentionally generic because it reflects MLX's head-dim-256 fallback behavior. It remains limited to matching multi-token calls above the configured KV threshold with supported masks and non-quantized KV. Decode, other head dimensions, unsupported masks, quantized KV, and model-defined SDPA implementations keep their existing paths. The engine pool reports measured reclaim generically, but the new scheduler accounting consumes it only for an active Qwen4-Exp profile.

This change addresses three related problems:

1. masked head-dim-256 attention was not consistently priced or routed as a bounded operation;
2. process-footprint growth was treated as token-linear, preventing safe chunk re-expansion;
3. measured overhead could be assigned to the wrong QSA route or charged again while it was already resident.

### 1. Bound masked SDPA256 prefill

- Price the unfused head-dim-256 fallback using its fp32 score representation.
- Allow explicit array masks to use the bounded tiled route.
- Let the Qwen4 memory profile from #3461 trust only bounded routes that support array masks.
- Retire stale transient samples when SDPA256 changes execution regime.

Dense fallback remains conservatively priced when no compatible bounded route is available.

### 2. Use nonlinear Qwen4 transient accounting

- Build on #3461’s static Qwen4 profile for token-scaled KV, indexer, and attention work.
- Track residual allocator growth as flat overhead instead of converting it to bytes per token.
- Keep reclaim debt and plateau state per execution regime.
- Permit one-tier chunk expansion after stable same-width samples, while preserving the pre-chunk abort cap.
- Keep the existing EWMA behavior for non-Qwen4 models.

### 3. Attribute and charge overhead correctly

- Record whether each QSA prefill actually executed the gathered or dense-mask route.
- Attribute post-execution measurements to that runtime route rather than the scheduler prediction.
- Keep gathered and dense histories isolated across route changes and restored prefixes.
- Treat retained allocator growth as part of the current physical footprint, avoiding a second charge for the same resident bytes.
- Charge only observed flat overhead that was actually reclaimed and may therefore need to be allocated again.
- Propagate request-scoped pool reclaim measurements from the engine pool back to the scheduler tracker.

## Safety

The change does not clamp or discard genuine dense-prefill estimates. Static dense attention pricing, stable abort limits, hard-watermark checks, and post-chunk physical-footprint checks remain in force. Reclaimed overhead is retained as a one-shot reallocation charge until a positive allocation sample repays it.

## Long-context validation

The branch was exercised against the release macOS app bundle with the memory guard enabled.

| | Before | After |
|---|---|---|
| Masked SDPA256 | Could be dense-priced or routed inconsistently | Uses a compatible bounded route, otherwise retains conservative dense pricing |
| Measured pool growth | Converted retained allocator growth into a recurring charge | Counts retained bytes through current footprint; only reclaimed bytes can be charged again |
| Restored-prefix replay | Rejected at 143,711 tokens after predicting another ~26.7 GB | Completed the same prefill and continued beyond 158k tokens |

| Scenario | Prompt tokens | Cached tokens | Result |
|---|---:|---:|---|
| Cold multimodal prefill | 180,161 | 0 | Completed successfully |
| Cached-prefix continuation | 190,190 | 178,176 | Completed successfully |
| Restored-prefix dense-mask replay | 143,711 | 139,264 | Completed and continued through later turns beyond 158k |

The cold run covered allocator growth and chunk re-expansion from small guarded chunks back toward the requested width. The continuation covered long-KV prefix reuse without a premature guard rejection.

The restored-prefix replay specifically exercised the accounting corrected here: the first dense-mask chunk retained roughly 30 GB of MLX pool buffers while adding only the expected resident KV. Subsequent chunks were admitted because retained buffers remained part of current footprint instead of being charged a second time. The workload progressed beyond 158k prompt tokens without a memory-guard rejection, Metal OOM, or server failure.

## Tests

Coverage includes:

- masked SDPA256 bounded-route eligibility and memory estimates;
- QSA gathered and dense route markers at the real execution seam;
- predicted-route fallback when caches have no marker, and conservative dense selection for mixed markers;
- runtime route correction and propagation across later chunks;
- restored-prefix and long-context prefill accounting;
- retained versus reclaimed allocator growth, including route-local reclaim when one request changes routes;
- reclaim-debt repayment;
- adaptive chunk plateau detection and one-tier expansion;
- hard abort-cap enforcement;
- unchanged non-Qwen4 EWMA behavior and non-Qwen route-marker isolation;
- engine-pool reclaim propagation and request-scoped route cleanup.

Validation:

```text
505 passed
```

The release macOS app bundle was rebuilt, its signature verified with `codesign --verify --deep --strict`, and its bundled Python sources matched against the branch worktree.

### DIFF
--- a/omlx/engine_pool.py
+++ b/omlx/engine_pool.py
@@ -2157,6 +2157,9 @@ class EnginePool:
             return 0
         freed = max(0, before - get_phys_footprint())
         if freed > 0:
+            record_reclaim = getattr(scheduler, "_record_prefill_reclaim", None)
+            if callable(record_reclaim):
+                record_reclaim(request_id, freed)
             logger.info(
                 "Reclaimed %s of pooled Metal buffers for prefill request %s "
                 "(no idle model to evict)",
--- a/omlx/memory_monitor.py
+++ b/omlx/memory_monitor.py
@@ -54,6 +54,18 @@ _SDPA_VECTOR_SUPPORTED_HEAD_DIMS = frozenset({64, 96, 128, 256})
 # dim-less path and matches the fp16/bf16 majority of MLX inference models.
 _SDPA_FALLBACK_SCORE_DTYPE_SIZE = 2
 
+# Bytes/elem the unfused head-dim-256 fallback actually materializes on
+# Metal: MLX keeps the attention score matrix in fp32 even for bf16/fp16
+# models (measured ~33GiB IOAccelerator spike on a 160k-token VLM prefill at
+# head_dim=256). Shared by the sdpa256 route gate
+# (patches/sdpa256_attention._tiled_route_required) and the Qwen4 prefill
+# profile so the router and the guard price the real fallback identically.
+# Kept separate from _SDPA_FALLBACK_SCORE_DTYPE_SIZE: that default covers
+# the dim-less generic path where the compute dtype is unknown and the
+# fp16/bf16 majority is the better prior; widening it globally would
+# re-price unrelated models without evidence.
+SDPA256_UNFUSED_SCORE_DTYPE_SIZE = 4
+
 # Head dims whose multi-token prefill is routed to an O(L) tiled/online-softmax
 # kernel instead of the unfused O(L^2) score-matrix fallback. Populated at
 # runtime by the kernel patch that installs the route (see
@@ -68,6 +80,7 @@ class _BoundedSDPAPrefillRoute:
     min_query_len: int
     min_kv_len: int
     kv_tile: int
+    supports_array_mask: bool
 
 
 _SDPA_TILED_PREFILL_HEAD_DIMS: dict[int, tuple[_BoundedSDPAPrefillRoute, ...]] = {}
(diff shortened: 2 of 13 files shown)


mvdbos:
Follow-up `aa897e74` fixes runtime-route attribution: QSA caches now report gathered vs mask-dense execution, and transient samples/pricing follow that actual route. This prevents the captured 28,164.64 MB mask-dense spike from becoming a 26.26 GB gathered charge. Added incident replay coverage (88 affected tests pass) and rebuilt/verified the signed macOS app bundle.

PR #3466, opened 2026-09-05
fix(benchmark): add gen_tps to benchmark pp-result logging

### LINKED ISSUE #3464: [BUG] Benchmark log line [benchmark-pp-result] omits generation (tg) throughput
In `omlx/admin/benchmark.py`, the `logger.info("[benchmark-pp-result] ...")` call only logs `processing_tps` (prefill/pp speed), `ttft_ms`, `e2e_ms`, and `cached_tokens`. It does not log `metrics["gen_tps"]` (decode/tg throughput), even though that value is computed in the same `metrics` dict and shown in the UI's benchmark results table and uploaded to the leaderboard. This makes it impossible to see generation speed for a benchmark run from `server.log` alone. Suggest adding `gen_tps=%.3f` to the log format string.

arty-hlr:
Benchmark runs were logging prefill throughput (`processing_tps`) but not decode/generation throughput (`gen_tps`), even though `gen_tps` is already computed and surfaced in the UI. This made generation speed unrecoverable from `server.log` alone.

- **Benchmark log payload update**
  - Extended the `[benchmark-pp-result]` format string in `omlx/admin/benchmark.py` to include `gen_tps` immediately after `processing_tps`.
  - Kept existing field order otherwise unchanged (`ttft_ms`, `e2e_ms`, `cached_tokens` remain in place).

- **`None`-safe formatting for generation TPS**
  - Added `float(metrics.get("gen_tps", 0.0) or 0.0)` in log arguments so unmeasurable decode runs (`gen_tps is None`) log as `0.000` instead of failing `%0.3f` formatting.

- **Targeted benchmark logging coverage**
  - Added/updated tests in `tests/test_benchmark.py` to assert:
    - `[benchmark-pp-result]` includes `gen_tps=...`
    - `gen_tps=None` is rendered as `gen_tps=0.000`

```python
logger.info(
    "[benchmark-pp-result] pp=%d processing_tps=%.3f gen_tps=%.3f "
    "ttft_ms=%.3f e2e_ms=%.3f cached_tokens=%d",
    pp_len,
    float(metrics.get("processing_tps", 0.0) or 0.0),
    float(metrics.get("gen_tps", 0.0) or 0.0),
    float(metrics.get("ttft_ms", 0.0) or 0.0),
    float(metrics.get("e2e_latency_s", 0.0) or 0.0) * 1000.0,
    int(metrics.get("cached_tokens", 0) or 0),
)
```

Closes #3464

### DIFF
--- a/omlx/admin/benchmark.py
+++ b/omlx/admin/benchmark.py
@@ -1954,10 +1954,11 @@ async def run_benchmark(run: BenchmarkRun, engine_pool: Any) -> None:
             )
             metrics["system_metrics"] = _sample_window(run, window_start)
             logger.info(
-                "[benchmark-pp-result] pp=%d processing_tps=%.3f "
+                "[benchmark-pp-result] pp=%d processing_tps=%.3f gen_tps=%.3f "
                 "ttft_ms=%.3f e2e_ms=%.3f cached_tokens=%d",
                 pp_len,
                 float(metrics.get("processing_tps", 0.0) or 0.0),
+                float(metrics.get("gen_tps", 0.0) or 0.0),
                 float(metrics.get("ttft_ms", 0.0) or 0.0),
                 float(metrics.get("e2e_latency_s", 0.0) or 0.0) * 1000.0,
                 int(metrics.get("cached_tokens", 0) or 0),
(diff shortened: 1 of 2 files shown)


tc3oliver:
Verified independently on upstream main `ddc47184` (M4 Max, macOS 26, Python 3.12): on main the two new tests fail and `[benchmark-pp-result]` has no `gen_tps`. With this PR cherry-picked onto `ddc47184`, `tests/test_benchmark.py` passes (136 passed), a measurable decode logs `processing_tps=500.000 gen_tps=100.000 ...`, and the `gen_tps=None` case logs `gen_tps=0.000` without raising.

PR #3468, opened 2026-09-05
feat: MoE expert streaming from SSD (dense resident, experts demand-loaded)

errrepe:
## What this PR contains

This PR adds SSD expert streaming for MoE models. Expert streaming is a loader that keeps expert weights on SSD and pages them into RAM on demand. The PR covers the streaming patch, the DeepSeek-V4 spill path, the scheduler guard, and focused regression tests. It excludes benchmarks, documents, the Mac app, the admin console, scripts, corpora, result files, and unrelated model compatibility changes. The engine and scheduler files still contain adjacent hunks from the full branch, and a follow-up can split them on request.

## Review fixes

This PR also fixes four review issues. If SSD backing creation fails, the load now fails with an error instead of keeping all expert banks in RAM. Each model now stores its top-k threshold on its own MoE blocks, so one model cannot change the output of a resident Qwen MoE. If the source checkpoint changes, the spill loader now discards old layer files and re-spills instead of serving stale weights. The prefill test fixtures now carry the LRU (least recently used) heap probe that the record path calls.

## Benchmark results

All runs share one configuration: 4 GiB expert budget, 2k prompt, 96 decode tokens, single request, no MTP. Hardware: M4 Pro with 48 GB of unified memory. RSS (resident set size) peak is the process peak during the run. Prefill speed is prompt tokens over TTFT (time to first token).

| model | TTFT (s) | prefill (tok/s) | decode 96 tok (s) | decode (tok/s) | proc peak (GiB) | Metal prefill / decode (GiB) |
|---|---|---|---|---|---|---|
| Qwen3.8-Flash-Next-JANG_4M | 33.5 | 56.6 | 39.7 | 2.42 | 10.64 | 12.01 / 6.76 |
| DeepSeek-V4-Flash-0731-JANG | 89.6 | 20.6 | 37.9 | 2.53 | 12.78 | 12.27 / 8.24 |
| GLM-5.3-Flash-JANG-MTP | 53.2 | 35.0 | 54.8 | 1.75 | 14.77 | 20.25 / 9.89 |

DeepSeek TTFT was 107.5 s on a cold page cache and 89.6 s warm. The table shows the warm run. Resident floors at budget zero are 6.40 GiB for Qwen, 7.85 GiB for DeepSeek-V4, and 12.16 GiB for GLM-5.3.

## Tests

Tests cover each fix. Run the focused suites with this command. All 83 tests pass.

```
.venv/bin/python -m pytest tests/test_expert_streaming_pr3468_regressions.py tests/test_prefill_oom_graceful.py tests/test_dsv4_spill.py -q
```


errrepe:
## Audit remediation: 2026-09-09 (all 18 findings closed)

`docs/expert-streaming-audit-2026-09-09.md` reviewed this subsystem and filed **4 P0, 6 P1, 8 P2**. All 18 are now fixed across 9 commits (`53d2a24f..553fd1af`); the audit doc carries a per-finding → commit map and the verification evidence.

**P0**
1. **Expert cache had no lock** and was mutated from three threads (MLX inference, `_WARM_POOL` workers, asyncio governor). Added a `threading.RLock` with locked/unlocked method pairs (`get`/`_get_unlocked`, …); `clear()` now resets `CacheStats` in place instead of rebinding it while `_count_demand` holds the old reference.
2. **Governor retarget was not atomic** — `_apply` drained only `_store`, leaving S3-FIFO `_small`/`_ghost` over capacity. Added `cache.resize(cap, per_layer)` as a single locked entry point.
3. **Invalid `topk_threshold` raised and callers swallowed it** → fail-open to a resident load (OOM). `_coerce_threshold()` now never raises: out-of-range / NaN / non-numeric → `None` + `logger.error`.
4. **Bit-exactness had no automated gate.** Added `test_streaming_decode_tokens_match_resident` (`@pytest.mark.slow`), validated by mutation — perturbing one expert weight by 1e-2 diverges the greedy token stream, while the clean streaming path matches resident exactly.

**P1** — governor keeps a pre-reconciliation `per_slot` (now `reconcile_per_slot()`); no `munlock` (added `_munlock_range`/`unpin_expert`/`unpin_all`, `close()` unpins first); racy reader cache leaking fds+mmaps (publish one canonical instance under lock, close the loser); unbounded IO waits (`OMLX_EXPERT_STREAMING_IO_TIMEOUT_S`, 120 s default, `0` disables); `hot_fraction == 1.0` inverted to all-cold; loader and admission used different Qwen4 PLE ceilings (now one `_qwen4_admission_ceiling()`).

**P2** — locked telemetry counters; 6 dead symbols deleted; the two "negative result" knobs **removed** (below); 14 new tests including direct `pin_expert`/`unpin_expert` coverage; doc drift fixed; 0–64 GiB budget clamp; dead residency check removed; `memtrace` now `close()`s via `atexit` and samples outside the lock.

### Two things worth a reviewer's attention

**P2-13 — the two negative-result knobs are deleted, not just documented.** `OMLX_EXPERT_STREAMING_RUN_MERGE_GAP` (gap-bridged reads: 34.0 s vs 31.4 s unbridged, net loss in both regimes) and `OMLX_EXPERT_STREAMING_EXTRA_ROOTS` (dual-SSD striping: ~1.85× *slower* with a mismatched disk) are gone, along with `bench/stripe_model.py`. I drew the line at **config surface removed, code capability kept**: `segment_runs(merge_gap=...)` still exists and is still tested, but it is no longer reachable from an env var — re-enabling it needs a code change plus a fresh A/B. `ExpertBackingStore(extra_roots=...)` deliberately stays: dsv4 spill-stacking reaches it via `absorb_extra_map()`.

**Cascading consequence of the above:** `"run_merge_gap"` was also a required key in three `effective_config` contracts (`bench/bench_expert_streaming.py`, `bench/compare_results.py:CRITICAL_FIELDS`, `tests/test_compare_results.py`). Left in place, *every* new-vs-archived A/B would report a spurious `0 vs None` difference on a knob neither side can set — so it was dropped from all three.

### Verification
- `tests/test_expert_streaming.py` + `tests/test_compare_results.py`: **246 passed**; `-m slow` token gate passes in 9 s.
- Full suite A/B against baseline `eba16235` (two ~15 min runs, sequential to avoid MLX GPU contention): 81 → 80 failures, `comm -23` **empty → zero regressions**. The 80 are all pre-existing environmental failures (MLX/Metal, SSH, sandbox `mkdir EEXIST`).
- `ruff --select F,E9` on the touched files: **identical finding set to baseline** — no new findings.
- Re-verified in `553fd1af` with a scripted sweep over all 18 findings (34 assertions, all green).

jundot:
Thanks for putting this together. I can’t merge a PR with roughly 100,000 added lines in its current form. Please narrow it to the files needed for SSD expert streaming and focused regression tests, removing the experiment history, benchmark artifacts, large corpora, generated explanations, and unrelated changes. Keep the key benchmark results in the PR description, including the hardware, settings, memory usage, TTFT, and decode speed.

I also found four issues that need addressing:

- When SSD backing creation fails, conversion falls back to retaining all expert banks in RAM. A model admitted only because streaming fits can then OOM. The load should fail cleanly instead.
- Per-model top-k settings are stored globally and applied to the shared Qwen class. I reproduced output changes in a resident Qwen MoE module after applying another model’s streaming settings.
- After the source checkpoint changes, the spill loader reuses old layer files and writes a fresh manifest for them. I reproduced updated expert weights being replaced by the previous checkpoint’s weights.
- The new `_streaming_lru_heap_growth()` call is missing from the prefill test fixtures, causing 16 tests to fail with `AttributeError`. The corresponding baseline suite passes.

Please address these with focused regression coverage so I can recheck the core implementation before merging.

errrepe:
Rebased the four wiring files onto current main, as requested in the last review note. All main-side code the stale base had deleted is restored (preserve_reasoning, Gemma4 global-KV, think-prefix caching, PLE prefetch, tracker wiring, settings v1). The scheduler diff is now 374 added lines, all streaming-only. 467 focused tests pass.

errrepe:
Cleanup: removed a 42-line env-gated expert-bank block that leaked in from an unfinished local experiment - it referenced a method that no committed class provides. The PR now contains only complete, tested streaming machinery.

errrepe:
Hi @jundot — following your review here, we opened a reworked version as #3672.

What changed relative to this PR:

- Scope cut to the shared streaming core, engine wiring, admin/WebUI surface, and focused tests (~25k lines vs ~100k). Benchmark artifacts, the DeepSeek-V4.1 native adapter, the Swift app changes, custom kernels, and cluster work are out; the V4.1 adapter will follow as a separate PR on top.
- The four issues from your review are addressed with dedicated regression tests: fail-clean when SSD backing cannot be built, per-model top-k stamped per-instance (no shared-class leakage), spill loader invalidates stale stacks via shard fingerprint (size, mtime_ns), and the test fixtures bind `_streaming_lru_heap_growth` in every test namespace.
- Fresh benchmark numbers in the description, including the MTP-under-streaming measurements that motivated the streaming gate.

Happy to split further if you would rather review it in smaller pieces.

PR #3469, opened 2026-09-05
perf(qwen4_exp): host-side decode optimizations for Qwen3.8-Flash-Next

Rascal:
### Summary

Qwen4-Exp decode was host-bound, not GPU-bound. On an M5 Max serving `Qwen3.8-Flash-Next-oQ4e-mtp`, a serial decode step spent ~22 ms on the host (5.4 ms of Python graph build plus 16.9 ms of MLX kernel encoding for an ~8.5k-node graph) while the GPU needed ~11.5 ms and sat idle for half of every step. This branch attacks the host side in three steps. Measured as a whole against the original `main` (e467261e), interleaved ABBA on the same box, same checkpoint, same prompts:

| Mode | main (e467261e) | this branch |
|---|---|---|
| Serial decode (MTP off), code 5k | 43.2 tok/s | 63.4 tok/s (+47%) |
| Serial decode, code 41k | 40.2 tok/s | 57.2 tok/s (+42%) |
| Serial decode, prose 5.8k / 34k | 43.2 / 41.0 tok/s | 62.7 / 57.9 tok/s (+45% / +41%) |
| Lightning MTP depth 3, mean of 4 prompts | 46.4 ms/cycle, 57.9 tok/s | 35.9 ms/cycle (−23%), 76.8 tok/s (+33%) |
| Lightning MTP depth 3, code 5k | 43.6 ms/cycle | 32.1 ms/cycle (−26%) |
| Lightning MTP depth 3, code 41k | 49.6 ms/cycle | 39.2 ms/cycle (−21%) |

MTP acceptance is unchanged within noise (77.4 vs 77.0%, 71.7 vs 69.8%, 70.1 vs 68.6%, 81.1 vs 78.9% on the four prompts).

Every path has an environment kill switch and fails closed to the existing code.

### What changed, and what each part is worth

**1. `Qwen4ExpRMSNorm` through `mx.fast.rms_norm` (c472afa5)** — the norm was hand-written in MLX ops (cast, square, mean, rsqrt, scale). It now calls the fused fast op, with the grouped hyper-connection norm normalising per stream and scaling afterwards. Numerics: the `1 + weight` scale stays fp32, so results differ from the old path by at most 1 bf16 ULP (2 ppm aggregate). Gain: **−4.0% ms/cycle** under MTP in an in-server A/B.

**2. Eager per-layer dispatch (dc6197dd)** — the decoder loop called one `async_eval` after building the whole forward, so the GPU waited through the entire Python build and most of the encoding. The loop now dispatches each decoder layer's output as soon as it is built (decode and verify rows only, ≤ 64 rows), so the GPU runs layer *i* while the host builds layer *i+1*. This is scheduling only: outputs are bit-identical, and the A/B shows identical cycle counts and acceptance in every arm. Gain: **serial +18..22% tok/s; MTP −11.5% ms/cycle** (44.1 → 39.1; code 5k 65.8 → 75.9 tok/s).

**3. Fused hyper-connection kernels (1650644a, 504b44e9, def1b1fd)** — `Qwen4ExpGatedResidual._forward` built ~30 MLX nodes per call and runs 97 times per forward; per-module encode attribution put it at 5.2 ms of host time per step, the largest single item. Three row-batched Metal kernels replace it for ≤ 16 rows: a per-stream RMS norm (bit-identical to `mx.fast.rms_norm`), the down and inject projections split over four K-slices with the `silu(mix/hc)` and `2·sigmoid(inject/hc)` epilogues folded in, and the up projection with the sigmoid gate, stream mix and mean over streams via simd shuffles. The affine unpack helpers are shared with `hc_projection`; 4/5/6/8-bit gs64 weights are supported, with independent bit widths for the down and inject projections. Numerics: fp32 is kept through the epilogues, so results are within a few bf16 ULP of the canonical path and at least as close to an fp32 reference as the canonical path is. Gain: **MTP −6.5% ms/cycle** on top of eager dispatch (39.2 → 36.6). Serial was unchanged at this point — not because serial decode is GPU-bound, as the first version of this description said, but because the fused path was being bypassed on 91 of the 97 modules (part 4).

Review follow-up (@mvdbos): the 4/5-bit down loop walks each K-slice in 512-element blocks with no tail, while the eligibility check only asked for a multiple of 256, so a hidden size such as 768/4-bit or 1280/5-bit passed the check and read past its slice (77 and 88 ULP off canonical in a repro; the shipped hidden 2560 was never affected). 504b44e9 adds the guard; def1b1fd then adds lane-guarded partial-block handling to the norm, down and inject loops and relaxes the gate to any multiple of 64. For hidden 2560 every tail is zero and the guards fold away — old and new kernels are bit-identical for every bit width at 1/4/16 rows, and an interleaved A/B is null (serial within ±0.5%, MTP 36.2/36.9 vs 36.0/36.3 ms/cycle). `compatible()` also logs once, at INFO, when a model's layout keeps it on the canonical path, so a future checkpoint that misses the fused kernels shows up in the log instead of just running slower (ec982bed keeps that line quiet for the deliberate hybrid-projection case below).

**4. Fused kernels ahead of the exact hybrid projection (b954693b)** — found through the new log line. With Lightning MTP off, `qwen4_exp.load_weights` flags 91 of the 97 hyper-connections with upstream's exact hybrid projection and compiles their single-token `_forward`; the fused path stepped aside for those modules, so in serial decode it only ever ran on the other 6 (with MTP on the hybrid is never installed, which is why the MTP numbers were fine). The fused path now takes precedence for the rows it covers; `_forward` is untouched and keeps the hybrid for everything else. Gain: **serial +12..16% tok/s** on top of the above — interleaved checkout-switching A/B, MTP off: code 5k 55.2 → 62.5, code 41k 50.8 → 57.0, prose 5.8k 54.2 → 62.6, prose 34k 51.4 → 58.1. MTP verify is unchanged. Serial outputs now match the canonical path to a few bf16 ULP, as the MTP path already did, instead of bit-exactly.

Stacked, parts 1–3 take an MTP verify cycle from 46.4 to 35.9 ms on this box (the combined run above), and parts 1–4 take serial decode from 43 to 63 tok/s; the per-part figures come from the individual A/Bs during development.

### Measurement

All numbers are decode-only tokens per second (or milliseconds per MTP cycle) from interleaved ABBA runs on one M5 Max, four prompts (code and prose at ~5k and ~35–41k tokens), 512 generated tokens, two passes each. MTP runs pin the draft depth so the adaptive controller cannot confound the comparison. Where a change is not bitwise, `ms/cycle` is the fair metric, and acceptance rates are reported alongside.

### Safety and tests

- Kill switches: `OMLX_QWEN4_EAGER_DISPATCH=0`, `OMLX_QWEN4_HC_FUSED=0`. The fused kernels fail closed to `_forward` on any runtime error with a one-time warning.
- The fused path is gated on batch·seq ≤ 16, bf16 input, `hc_count == 4`, quantised gs64 projections in 4/5/6/8 bits, no combined `input_inject_weight`, and Metal being available. Prefill and batched decode take the existing path.
- New tests: `tests/test_qwen4_eager_dispatch.py` (fires once per layer for ≤ 64 rows, never for prefill, outputs bit-identical on/off), `tests/test_qwen4_hc_fused.py` (parity against the canonical path and an fp32 reference for 4/5/6/8 bits × 1/4/16 rows × with/without inject, norm kernel bit-identity, eligibility fail-closed, kill switch), `tests/test_qwen4_rms_norm.py` (ULP and ppm bounds, grouping, fp32 scale).
- `tests/conftest.py` pins `MLX_ENABLE_TF32=0` so the fp32 parity tests hold on M5-class GPUs where TF32 is the default.

### DIFF
--- a/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/hc_fused.py
+++ b/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/hc_fused.py
@@ -298,9 +298,12 @@ def compatible(module, hyper_input) -> bool:
         and 1 <= hyper_input.shape[0] * hyper_input.shape[1] <= MAX_ROWS
     ):
         return False
-    if hasattr(module, "input_inject_weight") or getattr(
-        module, "_omlx_exact_hybrid_projection", False
-    ):
+    if getattr(module, "_omlx_exact_hybrid_projection", False):
+        # Upstream's exact hybrid projection (installed at load time for serial decode)
+        # keeps its own compiled single-token path; that is a runtime choice, not a
+        # checkpoint layout, so it is not logged.
+        return False
+    if hasattr(module, "input_inject_weight"):
         return _ineligible("combined input projection layout")
     hc_count = getattr(module, "hc_count", None)
     hidden = getattr(module, "hidden_size", None)
--- a/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/hc_fused.py
+++ b/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/hc_fused.py
@@ -1,0 +1,404 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Fused hyper-connection kernels for Qwen4-Exp decode and verify rows.
+
+``Qwen4ExpGatedResidual._forward`` builds ~30 MLX nodes per call (grouped norm,
+three small quantized projections, silu/sigmoid gates, stream mix and mean) and
+runs 97 times per forward. On Apple Silicon the host-side kernel encoding of
+that graph, not its GPU time, dominates a Lightning MTP verify cycle. Three
+row-batched Metal kernels replace it for small row counts:
+
+* ``N`` -- per-stream RMS norm with the ``1 + weight`` scale (bit-identical to
+  ``mx.fast.rms_norm``).
+* ``D`` -- ``input_mix_weight_down`` (320 rows) and ``block_inject_weight``
+  (``hc_count`` rows) projections per token row, split over 4 K-slices, with
+  the ``silu(mix / hc_count)`` and ``2 * sigmoid(inject / hc_count)``
+  epilogues folded in.
+* ``U`` -- ``input_mix_weight_up`` per output element, sigmoid, times the
+  normed stream, mean over streams via simd shuffles.
+
+Rows live in the grid (``batch * seq <= 16``). The affine unpack helpers come
+from :mod:`hc_projection`. Results match the canonical path to a few bf16 ULP
+(fp32 is kept through the epilogues); the path fails closed to ``_forward`` on
+any runtime error. Disable with ``OMLX_QWEN4_HC_FUSED=0``.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+
+import mlx.core as mx
+import mlx.nn as nn
+
+from .hc_projection import _HEADER
+
+logger = logging.getLogger(__name__)
+
+MAX_ROWS = 16
+_GROUP_SIZE = 64
+_SUPPORTED_BITS = (4, 5, 6, 8)
+_DISABLED = os.environ.get("OMLX_QWEN4_HC_FUSED", "1").strip().lower() in {
+    "0",
+    "false",
+    "no",
+    "off",
+}
+_KERNELS: dict[str, object] = {}
+_RUNTIME_FAILED = False
+_FAILURE_LOGGED = False
+
+_N_SOURCE = r"""
+    const uint row = threadgroup_position_in_grid.z;
+    const uint s = threadgroup_position_in_grid.y;
+    const uint t = thread_index_in_threadgroup;
+    const uint sg = simdgroup_index_in_threadgroup;
+    const uint lane = thread_index_in_simdgroup;
+    threadgroup float part[8];
+    constexpr int PER = H / 256;
+    const device T* xp = x + (size_t)row * K + (size_t)s * H;
+    const device T* wp = w + (size_t)s * H;
+    device T* op = xn + (size_t)row * K + (size_t)s * H;
+    float v[PER];
+    float ss = 0.0f;
+    for (int i = 0; i < PER; ++i) {
+        v[i] = float(xp[t + i * 256]);
+        ss += v[i] * v[i];
+    }
+    ss = simd_sum(ss);
+    if (lane == 0) part[sg] = ss;
+    threadgroup_barrier(mem_flags::mem_threadgroup);
+    float tot = 0.0f;
+    for (int i = 0; i < 8; ++i) tot += part[i];
+    const float inv = metal::rsqrt(tot / float(H) + eps[0]);
+    for (int i = 0; i < PER; ++i) {
+        const int k = t + i * 256;
+        op[k] = T(v[i] * inv * (1.0f + float(wp[k])));
+    }
+"""
+
+_D_SOURCE = r"""
+    const uint tg = threadgroup_position_in_grid.y;
+    const uint row = threadgroup_position_in_grid.z;
+    const uint sg = simdgroup_index_in_threadgroup;
+    const uint lane = thread_index_in_simdgroup;
+    constexpr int KS = 4;
+    constexpr int SLICE = K / KS;
+    constexpr int GROUPS = K / 64;
+    threadgroup float part[KS][8];
+    const device T* x = xn + (size_t)row * K;
+    const int rg = int(sg & 1);
+    const int ks = int(sg >> 1);
+    if (tg < R / 8) {
+        constexpr int PF = hc_pack_factor<BITS_D>();
+        constexpr int BP = hc_bytes_per_pack<BITS_D>();
+        constexpr int ROW_BYTES = K * BP / PF;
+        constexpr int PPT = 2;
+        constexpr int VPT = PF * PPT;
+        constexpr int BLOCK = VPT * 32;
+        constexpr int SCALE_STEP = 64 / VPT;
+        const int out_row = int(tg) * 8 + rg * 4;
+        const device uint8_t* wp = (const device uint8_t*)down_w
+            + out_row * ROW_BYTES + ks * (SLICE * BP / PF) + int(lane) * PPT * BP;
+        const device T* sp = down_s + out_row * GROUPS + ks * (SLICE / 64)
+            + int(lane) / SCALE_STEP;
+        const device T* bp = down_b + out_row * GROUPS + ks * (SLICE / 64)
+            + int(lane) / SCALE_STEP;
+        const device T* xp = x + ks * SLICE + int(lane) * VPT;
+        float result[4] = {0.0f};
+        float xv[VPT];
+        for (int k = 0; k < SLICE; k += BLOCK) {
+            float sum = hc_load_vector<T, VPT, BITS_D>(xp, xv);
+            for (int r = 0; r < 4; ++r) {
+                result[r] += hc_qdot<VPT, BITS_D>(
+                    wp + r * ROW_BYTES, xv,
+                    float(sp[r * GROUPS]), float(bp[r * GROUPS]), sum);
+            }
+            wp += BLOCK * BP / PF;
+            sp += BLOCK / 64;
+            bp += BLOCK / 64;
+            xp += BLOCK;
+        }
+        for (int r = 0; r < 4; ++r) {
+            float v = simd_sum(result[r]);
+            if (lane == 0) part[ks][rg * 4 + r] = v;
+        }
+        threadgroup_barrier(mem_flags::mem_threadgroup);
+        if (sg < 2 && lane < 4) {
+            const int r = int(lane);
+            float v = part[0][sg * 4 + r] + part[1][sg * 4 + r]
+                + part[2][sg * 4 + r] + part[3][sg * 4 + r];
+            v = v / float(HC);
+            act[(size_t)row * R + int(tg) * 8 + int(sg) * 4 + r]
+                = T(v / (1.0f + metal::exp(-v)));
+        }
+        return;
+    }
+    if (INJ == 0 || rg != 0) return;
+    {
+        constexpr int PF = hc_pack_factor<BITS_I>();
+        constexpr int BP = hc_bytes_per_pack<BITS_I>();
+        constexpr int ROW_BYTES = K * BP / PF;
+        constexpr int PPT = 1;
+        constexpr int VPT = PF;
+        constexpr int BLOCK = VPT * 32;
+        constexpr int SCALE_STEP = 64 / VPT;
+        const device uint8_t* wp = (const device uint8_t*)inject_w
+            + ks * (SLICE * BP / PF) + int(lane) * BP;
+        const device T* sp = inject_s + ks * (SLICE / 64) + int(lane) / SCALE_STEP;
+        const device T* bp = inject_b + ks * (SLICE / 64) + int(lane) / SCALE_STEP;
+        const device T* xp = x + ks * SLICE + int(lane) * VPT;
+        float result[HC] = {0.0f};
+        float xv[VPT];
+        int k = 0;
+        const int last = (ks == KS - 1) ? (SLICE - BLOCK) : SLICE;
+        for (; k < last; k += BLOCK) {
+            float sum = hc_load_vector<T, VPT, BITS_I>(xp, xv);
+            for (int r = 0; r < HC; ++r) {
+                result[r] += hc_qdot<VPT, BITS_I>(
+                    wp + r * ROW_BYTES, xv,
+                    float(sp[r * GROUPS]), float(bp[r * GROUPS]), sum);
+            }
+            wp += BLOCK * BP / PF;
+            sp += BLOCK / 64;
+            bp += BLOCK / 64;
+            xp += BLOCK;
+        }
+        if (ks == KS - 1) {
+            float sum = hc_load_vector<T, VPT, BITS_I>(xp, xv);
+            for (int r = 0; r < HC; ++r) {
+                result[r] += hc_qdot_safe<VPT, BITS_I>(
+                    wp + r * ROW_BYTES, xv,
+                    float(sp[r * GROUPS]), float(bp[r * GROUPS]), sum);
+            }
+        }
+        for (int r = 0; r < HC; ++r) {
+            float v = simd_sum(result[r]);
+            if (lane == 0) part[ks][r] = v;
+        }
+        threadgroup_barrier(mem_flags::mem_threadgroup);
+        if (sg == 0 && lane < HC) {
+            const int r = int(lane);
+            float v = part[0][r] + part[1][r] + part[2][r] + part[3][r];
+            v = v / float(HC);
+            inj[(size_t)row * HC + r] = T(2.0f / (1.0f + metal::exp(-v)));
+        }
+    }
+"""
+
+_U_SOURCE = r"""
+    const uint g = threadgroup_position_in_grid.y;
+    const uint t = thread_index_in_threadgroup;
+    const int h = int(g) * 64 + int(t >> 2);
+    const int s = int(t & 3);
+    const int n = s * H + h;
+    constexpr int PF = hc_pack_factor<BITS_U>();
+    constexpr int BP = hc_bytes_per_pack<BITS_U>();
+    constexpr int ROW_BYTES = R * BP / PF;
+    constexpr int GROUPS_R = R / 64;
+    constexpr int CH = 2 * PF;
+    constexpr int CPG = 64 / CH;
+    const device uint8_t* w = (const device uint8_t*)up_w + (size_t)n * ROW_BYTES;
+    const device T* sp = up_s + (size_t)n * GROUPS_R;
+    const device T* bp = up_b + (size_t)n * GROUPS_R;
+    float xv[CH];
+    for (int r = 0; r < S; ++r) {
+        const device T* a = act + (size_t)r * R;
+        float acc = 0.0f;
+        for (int gq = 0; gq < GROUPS_R; ++gq) {
+            const float sc = float(sp[gq]);
+            const float bi = float(bp[gq]);
+            for (int c = 0; c < CPG; ++c) {
+                const int e = gq * 64 + c * CH;
+                float sum = hc_load_vector<T, CH, BITS_U>(a + e, xv);
+                acc += hc_qdot<CH, BITS_U>(w + e * BP / PF, xv, sc, bi, sum);
+            }
+        }
+        float gate = 1.0f / (1.0f + metal::exp(-acc));
+        float v = gate * float(xn[(size_t)r * K + n]);
+        v += simd_shuffle_down(v, 1);
+        v += simd_shuffle_down(v, 2);
+        if (s == 0) mixed[(size_t)r * H + h] = T(v / float(HC));
+    }
+"""
+
+
+def _kernel(name: str, input_names: list[str], output_names: list[str], source: str, header: str = ""):
+    kernel = _KERNELS.get(name)
+    if kernel is None:
+        kernel = mx.fast.metal_kernel(
+            name=name,
+            input_names=input_names,
+            output_names=output_names,
+            header=header,
+            source=source,
+            ensure_row_contiguous=True,
+        )
+        _KERNELS[name] = kernel
+    return kernel
+
+
+def enabled() -> bool:
+    return not _DISABLED and not _RUNTIME_FAILED
+
+
+def _quantized_ok(projection) -> bool:
+    return (
+        type(projection) is nn.QuantizedLinear
+        and getattr(projection, "group_size", None) == _GROUP_SIZE
+        and getattr(projection, "bits", None) in _SUPPORTED_BITS
+        and getattr(projection, "mode", "affine") == "affine"
+        and "bias" not in projection
+        and isinstance(getattr(projection, "weight", None), mx.array)
+        and projection.weight.dtype == mx.uint32
+        and isinstance(getattr(projection, "scales", None), mx.array)
+        and isinstance(getattr(projection, "biases", None), mx.array)
+        and projection.scales.dtype == mx.bfloat16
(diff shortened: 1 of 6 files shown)


omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/hc_fused.py
```diff
@@ -0,0 +1,404 @@
...
+        return False
+    hc_count = getattr(module, "hc_count", None)
+    hidden = getattr(module, "hidden_size", None)
+    lowrank = getattr(module, "hc_lowrank", None)
+    if not (
+        isinstance(hc_count, int)
+        and isinstance(hidden, int)
+        and isinstance(lowrank, int)
+        and hc_count == 4
+        and hidden % 256 == 0
```
mvdbos: In `hc_fused.compatible()`, the geometry checks currently require `hidden % 256 == 0` and `(hc_count * hidden) % 512 == 0`. These checks admit hidden sizes that the fused down-projection kernel cannot safely process for 4/5-bit weights.

In `_D_SOURCE`, 4/5-bit projections use `PF=8` and `PPT=2`, making `BLOCK=512`. Because `HC=4`, `SLICE=K/4=hidden`. The down loop processes complete 512-element blocks without tail handling, so it requires `hidden % 512 == 0`; otherwise its final iteration reads beyond the current slice and, for the last slice, beyond the input row.

I reproduced this with modules that pass `hc_fused.compatible()`: hidden=768/4-bit produced a maximum fused-vs-canonical error of 0.912, while hidden=1280/5-bit produced NaNs. The current hidden=2560 checkpoint is unaffected.

Could we either require `hidden % 512 == 0` in `hc_fused.compatible()`—the minimal fix—or add tail handling to the down loop?

mvdbos (commented):
First of all, amazing work! Nice to get that free speed :-)

One compatibility-guard issue found in the fused down-projection path.

omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/hc_fused.py
Rascal: Thank you for pointing this out. Thanks to this, I realized I had missed another spot, and after correcting it, the serial decode speed increased from +24...29% to +41..47% (43.2 → 63.4 code 5k, 40.2 → 57.2 code 41k, 43.2 → 62.7 prose 5.8k, 41.0 → 57.9 prose 34k)

Guard fixes:

- **504b44e9** — the minimal guard: `compatible()` requires `hidden % 512 == 0` for every bit width; the redundant `(hc_count * hidden) % 512` check is gone.
- **def1b1fd** — tail handling instead of the minimal guard, and the gate relaxed to `hidden % 64 == 0` (the quantisation group and the up kernel's grid unit). The norm kernel guards a partial final 256-block; the down and inject loops stop at `SLICE - SLICE % BLOCK` and run one lane-guarded iteration for the remainder. The unpack helpers read exactly VPT values and VPT·BITS/8 weight bytes, so no lane touches memory past its slice. The inject loop's separate "safe" last block is gone since `hc_qdot_safe` is byte-for-byte `hc_qdot` in the shared header. For hidden 2560 every tail is zero and the guards fold away: old and new kernels are bit-identical for 4/5/6/8-bit at 1, 4 and 16 rows, and an interleaved A/B against the guard commit is null (serial within ±0.5%, MTP 36.2/36.9 vs 36.0/36.3 ms per cycle, identical cycle counts and acceptance). Tests: parity sweep at hidden 512/768/1152/1344/1536 × 4/5/6/8-bit (tails of 64, 128, 256 and 320 elements across the three kernels), norm bit-identity at 768/1152/1344, rejection below 64.

`compatible()` now also logs once, at INFO, when a model's layout keeps it on the canonical path, so a future checkpoint that misses the fused kernels shows up in the log rather than just running slower. That line turned out to matter immediately:

**Side finding.** The new log fired in the serial (MTP off) benchmark arms. With Lightning MTP off, `qwen4_exp.load_weights` installs the exact hybrid projection on every eligible hyper-connection (91 of 97 in this checkpoint) and compiles their single-token `_forward`; `compatible()` had always stepped aside for those modules, so in serial decode the fused kernels were only ever running on the other 6. With MTP on the hybrid is never installed, which is why the MTP numbers were fine. Two more commits:

- **ec982bed** — the log stays quiet for the hybrid case (a runtime choice, not a checkpoint layout) and only fires for a real combined `input_inject_weight` layout.
- **b954693b** — the fused path takes precedence over the hybrid projection for the rows it covers; `_forward` is untouched and keeps the hybrid for everything else. Interleaved checkout-switching A/B on the M5 Max, MTP off, 512 tokens: code 5k 55.2 → 62.5 tok/s, code 41k 50.8 → 57.0, prose 5.8k 54.2 → 62.6, prose 34k 51.4 → 58.1 (+12..16%). MTP verify is unchanged. Serial outputs now match the canonical path to a few bf16 ULP, as the MTP path already did, instead of bit-exactly.

The PR description is updated accordingly.

mvdbos: These are crazy numbers. Some anecdotal evidence: on a real agentic workload, at 75k context, I still got 65tok/s tg, dropping to 60 tok/s at 100k context. Not complaining.

mvdbos: @jundot

moc375:
Thanks for this PR — we spent a full day characterizing it on a different class of hardware and promoted it to production. Data below; one interaction at the end we think is worth a look (filed separately as an issue with full repro).

**Setup:** M3 Ultra (512GB UMA), omlx 0.6.4 serving stack, Qwen3.8-Flash-Next oQ4e-mtp (48 layers: 36 GDN-linear + 12 QSA), Lightning MTP depth 5, greedy, 3-4 engine instances sharing the GPU. Same-window A/B throughout: one boot per point, one variable per boot, interleaved pairs.

**Per component:**

- **hc_fused:** clean win at every context length — 1.1k +10.1% (59.2→65.2), 17.8k +16.9% (45.6→53.3), 35.5k +12.8% (44.5→50.2), MTP acceptance healthy (58-70%) throughout.
- **RMSNorm fast op:** prefill +2.8% (6k-prompt prefill tps 1066→1096).
- **eager dispatch:** the largest effect on this hardware. Dedicated clean window (siblings paused, 3 interleaved on/off pairs):

| Context | eager=1 | eager=0 | delta |
|---|---|---|---|
| ~1.1k | 76.5 / 77.1 / 78.3 | 58.6 / 52.5 / 56.5 | **+31~47%** |
| ~17.8k | 66.1 / 71.1 / 67.0 (accept 64-70%) | 50.1 / 52.1 / 54.0 | **+24~36%** |

  Zero interval overlap on every pair. 17.8k-context decode at 66-71 tok/s is the fastest we have ever measured on this machine at that context length. (The eager=0 arm runs below our all-quiet baseline because an idle sibling instance with GPU keepwarm stayed up during this window — it loads both arms equally, so the pairs stand.) Backbone ms/cycle @1k with both flags on: 43.6→29.5 (-32%, beyond the PR's -21~26% on M5 Max).

  A step-composition profile of ours explains the size: at 16k the GDN branch is roughly 40% per-layer dispatch glue (projections/conv/norms chain) vs 60% recurrence kernel, and the MoE block is already memory-bound at decode widths — dispatch is exactly what eager recovers.

**The interaction (details + repro in issue #3494):** with eager on, the **first MTP verify cycle after a ≥16k-token prefill intermittently accepts 0/6 draft positions** (2 of 13 boots, both in contended windows, 0 in quiet windows; always the boot's first long-prefill request; ≤1.1k contexts immune in every boot). The depth controller then correctly parks, and that request finishes as plain AR at ~1.08 tok/cycle — a 17-30% regression for it. Prefill chunks don't take the eager path (row gate), so the surface is the first verify cycle's `async_eval` vs the prefill→speculation handoff. We run eager ungated in production under acceptance monitoring; a context-gate env (`OMLX_QWEN4_EAGER_MAX_CTX`, validated 8/8 boots) is our fallback knob — happy to send it as a PR.

*(All numbers reproducible: same-window pairs, per-boot configs, acceptance telemetry and the profiling harness available on request.)*



---

**Correction (same day, [full details in #3494](https://github.com/jundot/omlx/issues/3494#issuecomment-5571216541)):** the "interaction" described above is not an eager-dispatch problem in omlx — the stranded-priming hook lives in **our own open PR #3265's `batched_verify.py`**, vendored into our serving tree (it does not exist on main / v0.6.4 / this PR's head). Stock omlx primes the MTP head correctly on first activation. Since both arms of our A/B table above carried that defect equally, the paired deltas stand as eager-vs-not on an unprimed head, but the absolute values are depressed; the eager mechanism analysis (host-bound step, dispatch-glue recovery) is unaffected.

ddark-il:
**Hardware:** Apple M5 Max 128 GB · **Model:** `Qwen3.8-Flash-Next-Uncensored-Mixed` (`qwen4_exp` Flash-Next hybrid, GDN + Qwen Sparse Attention) · **Sampling:** greedy (temperature 0.0) · **Generated:** 128 tokens · **Prompts:** deterministic 4k / 32k / 131k token ladder, identical across all runs.

**Method (thermally-controlled, decoupled):** A naive single-request bench measures decode immediately after a multi-minute prefill, i.e. on the hottest silicon, so the long-context decode is thermally throttled and noisy. To isolate the real decode speed each measurement is **decoupled**: prefill the prompt (cache-store on) → cool the SoC → then measure **cache-hit decode-only requests** on a cool chip. This removes the prefill heat from the decode. Comparing baseline and this PR in the *same* configuration isolates the code change.

`pp_tps` = prefill tokens/s (single cold prefill, cool chip), `gen_tps` = decode tokens/s (median of 3 cache-hit reps; reps were tight, spread < 1).

Tested in two configurations: **Lightning MTP ON** (speculative) and **Lightning MTP OFF** (serial decode).

---

## Config 1 — Lightning MTP ON

| Context | baseline pp_tps | PR pp_tps | Δ prefill | baseline gen_tps | PR gen_tps | Δ decode |
| 4,096 | 1418 | 1500 | +5.8% | 65.4 | 74.7 | **+14.2%** |
| 32,768 | 1480 | 1485 | +0.4% | 51.2 | 58.7 | **+14.6%** |
| 131,072 | 1458 | 1484 | +1.7% | 36.4 | 31.2 | **−14.3%** |


With MTP on, the fused decode is a clear win at 4k/32k but **does not carry to 131k** (decode ~14% slower there).

### Root cause — MTP draft acceptance rate degrades sharply at long context

The 131k reversal is explained by the **MTP drafter's acceptance rate falling with context length**. Measured on the same model (greedy, 128 generated tokens, MTP ON) via the patched `GenerationBatch` per-sequence telemetry; each row is 4 sequences (1 prefill + 3 cache-hit decodes) at that prompt length. `accept_rate` = accepted drafts / drafted; `tok/cycle` = tokens emitted per MTP verify cycle.

| Context | Sequences | Accepted / Drafted | **Accept rate** | avg tok/cycle | avg cycles (per 128 tok) |
| 4,096 | 4 | 352/379 | **92.9%** | 3.26 | 39.2 |
| 32,768 | 4 | 329/369 | **89.2%** | 2.86 | 44.8 |
| 131,072 | 4 | 269/380 | **70.8%** | 2.16 | 59.5 |

Depth-wise (chained MTP, depths 1–3, accepted/drafted):

| Context | d1 | d2 | d3 |
|---|---:|---:|---:|
| 4,096 | 97% | 90% | 91% |
| 32,768 | 92% | 87% | 89% |
| 131,072 | **70%** | **72%** | **74%** |

**Interpretation:**
- Overall accept rate drops **92.9% → 89.2% → 70.8%**, tokens/cycle drops **3.26 → 2.86 → 2.16**, and verify cycles per generation rise **39 → 45 → 60**.
- **Even depth‑1 acceptance collapses** (97% at 4k → 70% at 131k) — this is *not* depth compounding alone; the MTP head is fundamentally less accurate when conditioning on a 131k prefix.
- Rejections (`verify_emits`) rise from ~7 (4k) to ~27–35 (131k). Each rejection forces an expensive **dense-QSA verify forward** over the entire prefix (the MTP verify path bypasses the sparse gathered arm via `target_verify`).
- So with MTP ON, the drafter's poor long-context acceptance pushes most verify work onto GPU-bound full-prefix dense attention, which swamps the fused-decode host-side win and turns it negative at 131k.

---

## Config 2 — Lightning MTP OFF 

| Context | baseline pp_tps | PR pp_tps | Δ prefill | baseline gen_tps | PR gen_tps | Δ decode |
| 4,096 | 1466 | 1578 | +7.6% | 37.4 | 46.4 | **+24.1%** |
| 32,768 | 1482 | 1540 | +3.9% | 35.1 | 43.0 | **+22.5%** |
| 131,072 | 1503 | 1517 | +0.9% | 30.0 | 35.3 | **+17.7%** |

With MTP off (serial decode), the fused decode is a **consistent, large decode win at every length — including 131k** (+18% to +24%). Reps were very tight (< 1 tok/s spread), so this result is high-confidence.

Rascal:
@ddark-il Thanks for the benchmarks. Would it be possible to ask you to pull the latest commits from the PR branch and run the tests again? I’ve pulled the latest commits from the main branch into the PR-a branch—commits that could potentially have affected the benchmark results (a lot has changed in the main branch since the PR was created).

ddark-il:
@Rascal will do, but it will probably happen only tomorrow.

moc375:
Follow-up with correctly-primed numbers. After the correction thread in #3494 we fixed the stranded-priming defect in our own #3265 patch (the draft head was cold-starting on every request, which had been depressing every number we posted above), refreshed this PR to its current head, and re-ran a dedicated clean window. This is what the eager dispatch actually buys on a healthy tree:

| Context | eager=1 | eager=0 | delta |
|---|---|---|---|
| ~1.1k | 128.7 / 128.6 | 106.8 / 101.4 | **+20~27%** |
| ~17.8k | 84.0 / 84.0 | 70.0 / 68.0 | **+19~24%** |
| ~35.9k | 73.5 / 72.0 | 62.2 / 60.8 | **+17~20%** |

(2 interleaved rounds, one boot per point, greedy, temp 0; sibling instances paused, one idle instance with GPU keepwarm shared by both arms; acceptance telemetry 84.4% in every arm. The deltas are smaller than in our first report (+31~47%) because the priming fix raised the baseline of both arms — the eager mechanism itself is unchanged.)

Full ladder with this PR + our priming fix (same clean window, production config):

| Context | tok/s | note |
|---|---|---|
| ~1.1k | **~129** (128.5-129.5, n=4) | |
| ~17.8k | **84.0** | |
| ~35.9k | **72-73.5** | |
| ~65.6k | **59.6** | n=1 |

For the record on attribution, since we got this wrong once today: our tree = omlx 0.6.4 + this PR (head b954693b) + two fixes of ours that are not in omlx (the #3265 priming fix, and a QSA decode row-gather change following the layout note in #3494 — the latter measures neutral at ≤32k on our box, consistent with your 41k+ curve, so the ladder above is essentially this PR + the priming fix).

Both changes will be sent to our #3265 branch.

Finally — credit where it's due: @Rascal caught our wrong attribution with a source-level check against main/v0.6.4/this PR's head (saving us from filing a fix for a bug that isn't in omlx), and his note on the gathered-QSA layout is what our row-gather change follows. This thread is a good example of why characterization reports benefit from author review.

cc @jundot — third hardware class, clean paired data for the merge case.

Rascal:
@moc375 If you're interested and would like to try this out and run some tests on your own hardware, I'm working on another improvement to reduce token generation performance degradation with large(16k-250k+) context sizes. This optimisation kicks in when the context size exceeds 16k.
https://github.com/Rascal/omlx/tree/perf/qwen4-mtp-m5-decode-p2 (this branch includes the current PR plus new optimisations)

moc375:
Took the p2 branch for a spin on the M3 Ultra — the gathered verify arm is worth a lot here. Clean-window paired (sibling instances paused, interleaved rounds, one boot per point, greedy, single stream):

| 64k context | tok/s |
|---|---|
| p2 branch (7699f2e) | **77.2 / 76.9** |
| current PR head + our priming fix | 60.9 / 56.7 |

+27~34% at 64k. For scale: that puts 64k decode *above* what 32k did on the pre-p2 stack (72-73.5) — the long-context slope is nearly flat now on this machine (~129 @1k, 84 @16k, 72-73 @32k, 77 @64k). 32k came out ambiguous in our runs, consistent with your 20-30k crossover and the 32768 default; 16k unchanged as expected.

Our step profile had shown the verify rows paying the dense masked-path cost — good to see that piece closed. Happy to run any follow-up shapes (82k, code vs prose) if useful.

Rascal:
Fresh benchmark on m5 max:

### Serial decode
Mean of four measurements (two mirrored arms × two passes), code prompts, greedy, 512 tokens, gathered QSA off until 16K:
serial_on_m5_max
Context | main aa8db734 | this PR #3469 9a5caf8f | PR vs main | p2 [d16a6a4e](https://github.com/Rascal/omlx/tree/perf/qwen4-mtp-m5-decode-p2) | p2 vs main | p2 vs PR
-- | -- | -- | -- | -- | -- | --
4k (4,000) | 43.4 t/s | 63.2 t/s | +45.6% | (same as PR) | (same as PR) | (same as PR)
16k (16,352) | 42.0 t/s | 60.2 t/s | +43.2% | (same as PR) | (same as PR) | (same as PR)
65k (65,109) | 38.7 t/s | 53.7 t/s | +38.7% | 58.4 t/s | +50.9% | +8.8%
136k (136,421) | 35.4 t/s | 47.9 t/s | +35.3% | 56.7 t/s | +60.3% | +18.5%
210k (209,765) | 32.8 t/s | 43.4 t/s | +32.2% | 55.6 t/s | +69.3% | +28.1%

zviratko:
mergeit mergeit mergeit mergeit
amaze amaze amaze!!!
❤️

moc375:
Great tables — and one observation from our side that may be useful for the PR narrative: the M3 Ultra gains from the p2 arms are considerably larger than the p2-vs-PR column suggests, because the two comparisons have different compositions.

Your p2-vs-PR delta (+8.8% @65k) includes both the row gather and the gathered verify arms (the PR arm has neither). Our comparison tree already carried the decode-side row gather, so our 64k delta (+27~34%, clean window, two interleaved rounds: 77.2/76.9 vs 60.9/56.7) isolates the gathered verify arms alone. Same optimization, ~3x the payoff on this machine/content mix — worth noting for anyone on M3 Ultra-class bandwidth running prose-shaped traffic.

Also confirming the shape of your curve from the other end: our 32k was ambiguous (your 20-30k crossover), 16k unchanged — the threshold placement looks right.

The 210k number (+69.3% vs main, 55.6 t/s) is the one that matters most from our fleet's perspective — that's a genuinely usable long-context serving envelope now. We'll pull d16a6a4e and run the 82k / code-vs-prose shapes we offered; numbers coming.

moc375:
Follow-up shapes as promised (branch head 7699f2e, M3 Ultra, clean window, greedy, single stream; context sizes are true engine-reported prompt tokens):

| True context | Content | tok/s |
|---|---|---|
| ~70.9k | prose | 77.2 / 76.9 |
| ~70.9k | code | **89.3 / 89.4** |
| ~88.6k | prose | 72.5 / 76.6 |
| ~110.9k | code | 64.2 / 64.8 |

Two observations: (1) code-shaped content runs ~+16% over prose at the same true context — predictable structure lifts MTP acceptance, consistent with the depth histograms; (2) the long tail holds up beyond your table's range on this machine — 111k context still at 64+ t/s, above your M5 Max 136k point (56.7). One note for anyone reproducing: tokenizers differ enough between prose and code that nominal character-based sizing misleads — the code prompts above landed at 1.35x the nominal token count, so always read the engine's own prompt_tokens.

moc375:
One housekeeping note so the thread reads consistently: the d16a6a4e SHA in your benchmark table doesn't resolve on the branch (GitHub returns not-found; the branch head is 7699f2e — likely a local/WIP hash in the table). The shapes above were therefore run on the branch head 7699f2e, which is also what our earlier 64k validation used — so all three data sets on our side are the same commit. If a newer push is pending, we're happy to re-run the shapes against it.

jundot:
Thanks @Rascal. I really appreciate the thought you put into this approach and the substantial improvement it delivers. It's exciting to see this much more decode performance from the same hardware.

I compared main `aa8db734` with this PR at `9a5caf8f` on an M3 Ultra (512GB), using `Qwen3.8-Flash-Next-oQ4e-mtp`.

Python code contexts, greedy decoding, thinking off, one request at a time, 2,048-token prefill chunks, no prefix cache reuse, and 128 generated tokens. Each entry is the mean of two runs, with the second comparison in reverse order. Lightning MTP used a maximum depth of 3 with the default adaptive controller. PP and TG are tokens/s, measured through the oMLX VLM engine and admin benchmark helper.

| MTP | Context | PP main | PP PR | TG main | TG PR | TG gain |
|---|---|---:|---:|---:|---:|---:|
| Off | 4k | 798.5 | 816.8 | 34.8 | 53.9 | +55.1% |
| Off | 16k | 811.4 | 828.9 | 33.8 | 51.7 | +52.6% |
| Off | 32k | 806.7 | 826.8 | 32.6 | 50.0 | +53.1% |
| Off | 64k | 792.7 | 812.4 | 31.3 | 46.6 | +49.0% |
| On | 4k | 766.9 | 790.2 | 56.2 | 83.0 | +47.6% |
| On | 16k | 777.0 | 795.7 | 52.8 | 76.7 | +45.3% |
| On | 32k | 772.8 | 792.2 | 49.2 | 68.8 | +39.9% |
| On | 64k | 759.1 | 768.8 | 40.2 | 57.5 | +43.3% |

PP improved by 1.3-3.1%. Peak MLX memory was effectively unchanged: 102.15-107.92 GiB with MTP off and 103.71-109.51 GiB with it on. I saw no gibberish in the generated outputs.

I also compared accuracy with MTP off, thinking off and greedy decoding:

| Evaluation | Main | PR |
|---|---:|---:|
| HumanEval, all 164 problems | 159/164 (96.95%) | 160/164 (97.56%) |
| MMLU, 500 questions across 57 subjects, 5-shot, seed 42 | 428/500 (85.6%) | 427/500 (85.4%) |

HumanEval and MMLU scores stayed close to baseline, with no broad accuracy drop observed in these evaluations.

LGTM. I'll merge this, then address the hybrid-path test, lazy Metal compilation errors, and comment cleanup in one follow-up commit.

Issue #3240, opened 2026-08-28
qwen4_exp VLM: reshape error on image input with Jundot/Qwen3.8-Flash-Next-oQ4e-mtp


sunnybluesea:
## Description

Image (vision) input fails with a reshape error when serving `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp` (qwen4_exp, multimodal — contains vision tower) on oMLX 0.6.3. Text-only requests work fine.

## Environment

- oMLX: 0.6.3
- macOS 27 (Apple Silicon)
- Hardware: M5 Max, 128 GB unified memory
- Model: `Jundot/Qwen3.8-Flash-Next-oQ4e-mtp` (106GB, 21 shards, includes 333 vision_tower weight keys in shard 00001)
- Model settings: `qwen4_ple_ssd_offload: true`, `mtp_enabled: true`, `mtp_num_draft_tokens: 3`, ctx 131072

## Steps to reproduce

Send a chat-completion request with an image (base64 data URL) as in the standard OpenAI-compatible VLM format.

## Actual behavior

Request fails; server log shows:

```
RuntimeError: [reshape] Cannot reshape array of size 133120 into shape (1,14,4,2560).
```

Client receives `IncompleteRead` / truncated response.

## Notes

- 133120 = 13 × 10240, expected 14 × 4 × 2560 = 143360 — the vision tower patch embedding count appears off by one image-patch group vs the expected 14 spatial tokens.
- The model card reports `pipeline_tag: text-generation` but config contains a full `vision_config` (depth 27, hidden 1152, patch 16, temporal patch 2) and 333 `vision_tower.*` weights, so image support is expected.
- Text generation with the same settings is unaffected.

sunnybluesea:
## Deep root-cause analysis (multi-dimension verification)

After systematic investigation, I found a **deterministic trigger boundary**:

### Reproducibility: image-following text length boundary
Same image (/tmp 600x200 PNG), same model, same settings. Only the text after the image changes:

| Text after image | Total tokens | Tokens after image | Result |
|---|---|---|---|
| "hi" | 56 | 9 | ✅ success |
| "读图。" | 58 | 11 | ✅ success |
| "请描述这张图片。" | 60 | 13 | ✅ success |
| "读出图片上的所有文字。" | 61 | 14 | ❌ reshape error |

**≥14 text tokens after the image deterministically crashes; ≤13 works.** Two images + short text also succeeds — so it is NOT related to image count or total token count.

[...]

PR #3475, opened 2026-09-06
fix(scheduler): a throttled prefill chunk must never grow past the tokens left

williamxie1989:
## Symptom

Intermittent server-side crash on image-bearing requests to a Qwen4Exp VLM engine, matching the signature in #3240:

```
ValueError: [reshape] Cannot reshape array of size 1556480 into shape (1,153,4,2560).
  ...
  File "omlx/scheduler.py", line 3674, in _do_external_prefill
    self.model(input_arr[:, :n_to_process], cache=prompt_cache, **model_kwargs)
  ...
  File ".../qwen4_exp/language.py", line 2389, in Qwen4ExpPLELayer.__call__
    keys = self.norm_key(...)
```

`1556480 = 152 × (hc_count·hidden_size)` vs `153 × 10240`: the PLE layer's `input_ids` spans exactly one token less than `hidden_states` (which comes from `inputs_embeds`).

## Root cause

`Scheduler._adaptive_chunk_size` documents "`>= 1`, `<= requested`", but its floor — `n = max(min_chunk, min(requested, n_fit))` and the band-tier clamp below it — returns **more than `requested`** whenever a prefill tail is shorter than `prefill_min_chunk_tokens` and memory pressure takes the shrink path. The log is silent on exactly this case: the throttle INFO line is gated on `n < requested`.

`_do_external_prefill` then never re-clamps the chunk to the tokens actually left, and its two same-sized slices are not same-length:

- `input_arr` covers `tokens[:-1]` (the final token is deferred to `insert()` for the first decode step);
- VLM `inputs_embeds` spans the **full prompt** — always exactly one row wider.

MLX slices silently clamp at the array edge, so a chunk inflated past the tail desynchronizes the two: `input_arr[:, :n]` → 152 columns, `embeds_array[:, :n]` → 153 rows → the model gets mismatched lengths and Qwen4Exp's PLE dies on the reshape. This also explains the probabilistic nature: it needs an embeds-bearing (VLM) request **and** a tail shorter than the floor **and** throttle shrink pressure landing exactly on that chunk (e.g. `prefill_min_chunk_tokens: 256` near the memory ceiling).

## Fix

- `_adaptive_chunk_size` now honors its documented contract (`n = min(requested, n)` after the floor clamps).
- Both prefill loops (external + text-chunked) clamp the chunk to the tokens actually remaining before slicing. The chunked loop is text-only, but the same inflation silently drifted its `tokens_processed` past the rows actually cached (skewed progress and boundary-snapshot totals).

The floor remains an admission bound — the guard still rejects/raises based on the min-chunk transient — it just no longer *pads* a short tail.

## Tests (red → green)

- `tests/test_prefill_oom_graceful.py::test_adaptive_throttle_tail_below_floor_never_grows_chunk` — helper contract: requested 152, floor 256 under pressure; red: returned 256.
- `tests/test_scheduler.py::TestVLMPositionStateClearing::test_vlm_external_prefill_never_slices_embeds_longer_than_tokens` — drives `_do_external_prefill` with floor stubs and asserts `inputs_embeds` and `input_ids` lengths at the model boundary; red: 10 vs 9.

## Notes on #3240

This eliminates a mechanism that reproduces the exact numeric signature of the issue (N vs N+1 in the PLE reshape, embeds wider by the deferred final token). The reporter's deterministic small-prompt boundary may involve an additional path (their stack went through the legacy pending-embeds route on 0.6.3); I can extend the repro there if the symptom persists after this lands.

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -3623,6 +3623,12 @@ class Scheduler:
                 request_id=request.request_id,
                 gathered_core=gathered_core,
             )
+            # A chunk may never exceed the tokens actually left: the token
+            # row and the VLM inputs_embeds (one row wider — the +1 final
+            # token deferred to insert()) are sliced with the same size and
+            # MLX clamps silently, so an oversized chunk desynchronizes them
+            # and Qwen4Exp's PLE dies on an off-by-one reshape (#3240).
+            n_to_process = min(n_to_process, remaining)
             if getattr(request, "benchmark_trace", False):
                 request.benchmark_prefill_chunks.append(int(n_to_process))
                 request.benchmark_requested_steps.append(int(prefill_step_size))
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for the fix. I reproduced the PLE reshape error on Qwen3.8-Flash-Next-oQ6e-mtp by forcing the throttle’s shrink path, then confirmed all four image requests succeeded with this change and matched the unthrottled responses. This verifies the short-tail fix; I’ll keep #3240 open pending verification of the reported text-length boundary. LGTM, and I’ll shorten the repeated comments in a follow-up.

PR #3480, opened 2026-09-06
fix downloader

### LINKED ISSUE #3477: bug: download size overflows
**Describe the bug**

Image

**To Reproduce**
I think it would happen only with the specific model

**Expected behavior**
the actual size should show up

**Desktop (please complete the following information)**:
- macOS Version: Tahoe 26.1
- oMLX Version: 0.6.4(2529)

humdrum00001010:
Xet download path now uses the model.safetensors.index.json for weight filtering, fetches the expected model from the hub.

Fixes #3477 

which made me download over 200GB on that specific repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,3 +129,4 @@ tq_batch_memory.md
 .claude/
 issues.md
 clusterPlan.md
+packaging/_venvstacks_resolved.toml
--- a/omlx/admin/hf_downloader.py
+++ b/omlx/admin/hf_downloader.py
@@ -876,8 +876,9 @@ class HFDownloader:
         """Execute a download task.
 
         Waits for the download semaphore (only one download runs at a time),
-        then fetches repo info for total size and runs snapshot_download in a
-        thread while polling the target directory for progress updates.
+        then fetches repo info for total size and runs snapshot_download
+        when failed with metadata matching in a separate thread while polling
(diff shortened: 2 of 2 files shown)


humdrum00001010:
I think the downloader has many fallbacks intentionally, didn't touch the HTTP path, etc.

jundot:
Thanks for looking into this. I confirmed that the repo contains older weight shards and multiple GGUF variants, so the excess download is real. Filtering solely by the safetensors index also excludes config and tokenizer files.

I don't think it's practical to anticipate every exception across repos. I'd prefer to download the full repo by default and let users specify exclusion patterns for each download. I'm closing this PR in favor of that approach.

humdrum00001010:
> Thanks for looking into this. I confirmed that the repo contains older weight shards and multiple GGUF variants, so the excess download is real. Filtering solely by the safetensors index also excludes config and tokenizer files.
> 
> I don't think it's practical to anticipate every exception across repos. I'd prefer to download the full repo by default and let users specify exclusion patterns for each download. I'm closing this PR in favor of that approach.
> 

Agreed, that makes sense. One thought, though: perhaps we could first try to identify the expected weights from the repo layout or config, and fail early if that isn’t possible. That would give users a chance to select the weights they want before any actual download begins.

Issue #3429, opened 2026-09-04
Tool calls in the bare `<function name="…"><param name="…">` dialect are never parsed (MiniCPM5 and family)


An0nya:
## Summary

Models whose chat template emits an unwrapped `<function name="…">` envelope have their tool calls
silently passed through as assistant text: `tool_calls` is `null`, `finish_reason` is `stop`, and
the raw XML lands in `content`. From the client's side the model looks like it can't call tools at
all, when in fact it emitted a perfectly well-formed call.

## Reproduction

`POST /v1/chat/completions` with a `tools` payload, model `MiniCPM5-1B`, temperature 0:

```json
{"messages":[{"role":"user","content":"Read the file /workspace/TASK.md. Use the read tool."}],
 "tools":[{"type":"function","function":{"name":"read",
   "parameters":{"type":"object","properties":{"path":{"type":"string"}},"required":["path"]}}}]}
```

Response:

```
finish_reason : stop
tool_calls    : null
content       : '<function name="read"><param name="path">/workspace/TASK.md</param></function>'
```

The model's own reasoning trace confirms it believes it complied: *"I'll generate the tool call in
the specified format."* It did — the format just isn't one oMLX reads back.

An identical request to `maple-preview-2bit-mlx` (a `<tool_call>`-dialect model) returns a properly
populated `tool_calls` array, so the request shape is fine.

## Why it isn't caught

Every XML path in `omlx/api/tool_calling.py` keys off a wrapper marker that this dialect doesn't
emit:

- `_parse_xml_tool_calls` (line 580) → `_marker_payloads(text, "<tool_call>", "</tool_call>")`
- `_parse_namespaced_tool_calls` (line 659) → `<{namespace}:tool_call>` … `</{namespace}:tool_call>`

Inside those wrappers the table is already rich — `<function=NAME>`/`<parameter=KEY>` (line 618),
`<invoke name="…">`/`<parameter name="…">` (line 684), `<arg_key>`/`<arg_value>` (line 633). The
gap is specifically the **unwrapped, attribute-style** form:

```
<function name="NAME"><param name="KEY">VALUE</param></function>
```

Note `param`, not `parameter`, and attribute-style naming rather than `<function=NAME>`.

## Why the obvious fixes don't apply

**Can't key on `model_type`.** `MiniCPM5-1B` reports `model_type: llama`
(`architectures: ["LlamaForCausalLM"]`), so the `_is_laguna_model` pattern in
[...]

PR #3486, opened 2026-09-06
Add IFM/K2-Horizon serving, oQ, and ANE prefill

### LINKED ISSUE #3438: K2-Horizon MoVA model support

mbarnson:
Closes #3438. Adds dense K2 and MoVA serving, oQ quantization, and native reasoning/tool parsing with request-scoped tool names. Optional Uno decoding follows in #3535. I also snuck in some hf cache detection of mlx-compatible models.

Shares Qwen’s ANE controls in the web UI and macOS app. The tuner shows the current test, then the winning configuration and prompt-processing speed.

Validation: Python tests cover native transfers, eight-request batching, and caches. Qwen Q4/Q8 native checks passed; prior UI validation passed macOS tests. ANE display update: 91 focused Python tests and 25 macOS tests passed, with browser visual checks. No new performance measurements.

ANE seems to accelerate some long context workloads for IFM/K2-Horizon models, so I glommed support onto the qwen3.5 ANE approach.

Also covers the K2-Horizon portion of #3429 (tool calls and thinking tags).


mbarnson:
Moving to draft as I work out tokenizer tool call formatting issues.

jundot:
Thanks for working on this. I saw your note about moving it to draft for the tokenizer/tool-call fixes, so these are just suggestions for when it’s ready.

Would you be open to keeping K2/MoVA support, oQ, parsing, and ANE tuning/UI together, with Uno in a separate PR? For the first part, it would help to check multi-turn chat and tool-result round trips with ANE off/on. I also noticed XML string arguments lose surrounding whitespace, which matters for file edits.

For Uno, a small off/on comparison on the same Mac and checkpoint could help users decide when to enable it: response time, generation speed, memory, and overlapping requests, alongside supported models and limitations.

mbarnson:
> Uno in a separate PR?

Happy to. The Uno model for the 7B leads to significant performance improvements, but for the 0.9B it's a wash. 

> For Uno, a small off/on comparison on the same Mac and checkpoint could help users decide when to enable it: response time, generation speed, memory, and overlapping requests, alongside supported models and limitations.

Ah, like the ANE testing? Makes sense. 

Thanks for the suggestions. I'll strip UNO support off to a separate PR as you suggest, focused on helping the user understand if for their specific Mac and workload/temperature it's applicable. The fact one cannot modify thinking budgets for UNO models is itself disqualifying for things like text completion. And at 0.9B, the gains are modest to nonexistent on my hardware, while for 7B they are considerable.

mbarnson:
I turned GPT-6-Astra Xhigh loose to help me split it into two PRs. But the spaghetti code with special-casing of K2-Horizon was unacceptable. Simplifying before my next push, but we're getting closer!

mbarnson:
Ready to review! Thanks, Jun.

An0nya:
Small field note in case it's a useful test case: serving `K2-Horizon-7B-oQ4e` under 0.6.4 (with a stopgap model-dir parser, since this PR hasn't landed), the model intermittently emits **repeated `<ifm|tool_calls>` openers** for a single call. Across four agent sessions, openers outnumbered closers consistently — 32:14 in the longest one. Sample of what reached the assistant text channel:

```
...let me verify the physics section renders properly by viewing it:<ifm|tool_calls>

<ifm|tool_calls>

</ifm|tool_calls>
```

The call itself was well-formed and executed fine in every instance — only the envelope repeats.

Reading `_parse_group_body`, I *think* a stray opener inside the group would exit the call loop and then trip `if text[end:].strip(): raise`, surfacing as "incomplete or malformed tool call" — but I can't tell statically whether that path is reachable once `K2HorizonOutputParserSession` is stripping envelopes by token ID upstream, and it may simply be an artifact of the current envelope handling that won't occur under this PR. Flagging rather than asserting.

Happy to supply raw transcripts if useful.

mbarnson:
> the model intermittently emits **repeated `<ifm|tool_calls>` openers** for a single call.

I'm digging in to see if I can reproduce this myself now. Thanks for the note! I'm operating on the assumption that this is not a problem with the model but with my tool-calling fixes.

Which harness were you using, btw? Or just the built-in chat?

Edit: I haven’t reproduced the repeated openers locally with Pi using BF16 or `hermitdave/K2-Horizon-7B-oQ4e`, including runs with the tool grammar disabled.

Could you share your harness/version, exact model repo/revision, and stopgap parser, plus a request/response showing the issue? The tool definitions and preceding conversation would help. Unfiltered model output, if available, would be especially useful for locating whether duplication starts during generation, parsing, or client handling.

Edit2: oh! You were saying since this PR hasn’t landed yet, the symptom may not recur. Yes, that’s my observation. Without using the grammar the tool calling seems to be mixed/malformed JSON/XML hybrids from this model in most cases at most quantizations.  But I don’t see envelope opener repeats coming from the model…. Just malformed calls without the grammar. 

Finally, Edit 3: Why not try this PR instead? I've worked through some grammar issues with this model, though I haven't figured out how to fix its apparent inability to use JSON grammar when asked, just XML...

```bash
git clone --branch ifm-k2-horizon --single-branch \
  https://github.com/mbarnson/omlx.git omlx-pr-3486
cd omlx-pr-3486

python3.13 -m venv .venv
.venv/bin/python -m pip install --upgrade pip
.venv/bin/python -m pip install -e ".[grammar]"

.venv/bin/omlx serve \
  --model-dir "/absolute/path/to/your/models" \
  --base-path "$PWD/.omlx-pr-test" \
  --host 127.0.0.1 \
  --port 8001 \
  --api-key pr3486-test \
  --log-level debug
```

mbarnson:
Pushed [3dc38054](https://github.com/mbarnson/omlx/commit/3dc38054dd5b641dc2f1cf99d66f52f890a5eba8) to make xgrammar optional for K2 tool calling. This PR previously rejected tool-enabled requests when xgrammar was unavailable, and malformed generated calls could terminate the request. That was too strict for normal oMLX installs.

For the source-install instructions above, `pip install -e .` is now sufficient for K2 tool calling; `pip install -e ".[grammar]"` is optional.

This addresses handling of the separate malformed outputs found during testing.

An0nya:
Agreed on making xgrammar optional, rejecting tool-enabled requests outright was going to bite normal installs.

Also, I think I know why you couldn't reproduce the repeated openers: my model dir has a modified chat template. Most upstream K2 templates raise on an assistant message with no thinking field, but mine quietly substituted empty strings. Since my harness doesn't send a separate thinking field, my prompts were being built with no opening <ifm|think>, a stray closing one inside content, and tool wrappers inline, which the model later copied. Rebuilding the same conversation with proper reasoning_content gives balanced tags every time. So it's a local issue due to trying to get a fresh model to work in oMLX with TQ kv cache in a pi harness. Sorry about the wild goose chase, glad you were able to push it, and your change resolves the input side of it anyway.

jundot:
Thanks for splitting out Uno and working through the fixes. I confirmed the whitespace handling and native ANE checks, including concurrent decoding and cache restoration.

I’ve prepared a small maintainer follow-up for the remaining Responses history normalization, the updated ANE UI test, and some unused parser code. I’ll take care of those on my side; no further changes needed from you for these items.

thinklikesqlhns:
I'm serving 7.0.0 dev3 and trying to load K2-Horizon-MoVA-36B-A4B-oQ4e is erroring still. Error: {"error":{"message":"Model 'K2-Horizon-MoVA-36B-A4B-oQ4e' failed to load: Cannot stack model.layers.3.mlp.experts.gate_proj.weight: 100 expert tensors missing, first model.layers.3.mlp.experts.0.gate_proj.weight. Reload models after fixing the files to retry.","type":"invalid_request_error","param":null,"code":null}}

Issue #3209, opened 2026-08-27
Decode fairness retains only 13–23% of solo generation throughput during concurrent long prefill on oMLX 0.6.3


yourlocalplanet:
### Environment

- oMLX: 0.6.3 final
- MLX: 0.32.0
- mlx-lm: 0.31.3
- macOS: 26.5.1 (25F80)
- Hardware: Mac Studio with Apple M3 Ultra and 256 GB memory
- Max concurrent requests: 3
- Chunked prefill: enabled
- Decode fairness: enabled
- Prefill priority: Speed
- Memory guard: Balanced
- TurboQuant KV: disabled
- Lightning MTP: enabled unless otherwise specified

### Summary

While one request processes a fresh 60K-token prompt, an unrelated 1K-token request retains only 13–24% of its solo generation throughput.

This reproduces across both the dense Qwen3.8-27B model and the MoE Qwen3.6-35B-A3B model. Unique prompts were used to prevent prefix-cache hits, and no other requests were active.

The slowdown also persists with Lightning MTP disabled.

### Qwen3.8-27B AWQ with ANE and MTP enabled

- 1K request running alone: 62.9 tok/s
- 1K request during concurrent 60K prefill: 14.7 tok/s
- Retained throughput: 23.4%
- Short-request TTFT during contention: 6.20 seconds
- Long-request TTFT: 200.05 seconds
- Long-request decode speed: 30.9 tok/s
- Solo MTP acceptance: approximately 81%
- Contended MTP acceptance: approximately 57%

### Qwen3.8-27B AWQ with ANE disabled and MTP enabled

- 1K request running alone: 67.9 tok/s
- 1K request during concurrent 60K prefill: 8.9 tok/s
- Retained throughput: 13.0%
- Short-request TTFT during contention: 5.96 seconds
- Long-request TTFT: 236.55 seconds
- Long-request decode speed: 30.2 tok/s
- Solo MTP acceptance: approximately 88%
- Contended MTP acceptance: approximately 55%

Enabling Qwen ANE prompt processing improves the contended Qwen3.8 decoder from 8.9 to 14.7 tok/s. ANE is therefore mitigating the problem rather than causing it.

### Qwen3.8-27B AWQ with ANE enabled and MTP disabled

- 1K request running alone: 36.56 tok/s
- 1K request during concurrent 60K prefill: 8.91 tok/s
- Retained throughput: 24.4%
- Short-request TTFT during contention: 5.56 seconds
[...]

Issue #2031, opened 2026-06-28
Prefill is serialized at batch=1 — batched/mixed prefill to cut TTFT under concurrency


coredo-eu:
> Date: 2026-06-28. Build observed: oMLX 0.4.3 (260609215353-macos26-tahoe), mlx 0.31.2, Mac Studio M3 Ultra 512GB.

---

## Summary
oMLX runs **all prefill itself, one request at a time at batch dim 1**, and uses mlx-lm's `BatchGenerator` purely as a **decode-only** continuous batcher. Under concurrent load, N simultaneous prompts prefill strictly sequentially, so time-to-first-token grows **linearly with concurrency**. The native mlx-lm batched-prefill path (`PromptProcessingBatch`) exists but is deliberately bypassed.

This is a **latency** issue (tail TTFT under concurrency), not a raw-throughput one — see the honest framing in "Expected impact".

## Evidence (measured on our box)

**Empirical — prefill scales linearly with concurrency (= serial):** firing N concurrent distinct ~1500-token prompts with `max_tokens=1` (pure prefill):

| N concurrent | wall time | wall/N |
|---|---|---|
| 1 | 4.8 s | 4.8 s |
| 2 | 10.7 s | 5.3 s |
| 4 | 18.1 s | 4.5 s |
| 8 | 37.6 s | 4.7 s |

`wall/N` is flat → 8 prompts take ~8× one prompt. If prefill were batched, wall/N would drop with N.

**A single prefill already saturates the GPU** (98% util, 132 W on M3 Ultra) — so this is about *scheduling/latency*, not unused compute. Decode is bandwidth-bound (≈94% occupancy, ~50 W).

## Where it happens (oMLX 0.4.3 source)
- `scheduler.py:1602` — `BatchGenerator(prefill_batch_size=1, …)` constructed with batch=1.
- `scheduler.py:4570-4573` — docstring: *"Each request is prefilled externally before being inserted into BatchGenerator, so prefill_batch_size=1 is always used."*
- `scheduler.py:1716` `_do_external_prefill` / `:2172` `_step_prefill_chunk` — each runs `self.model(chunk, cache=state.cache)` with `chunk` shape `(1, n)` — one request, one cache.
- `scheduler.py:2388` `_advance_chunked_prefills` — per-request loop; `chunked_prefill=true` only **interleaves** one chunk per request per step, still batch=1 each.
- `scheduler.py:4590` `_schedule_waiting` — admits/prefills new requests one at a time.
[...]

PR #3487, opened 2026-09-06
fix(scheduler): yield between contended prefill chunks

chenqianhe:
## Summary

Close two enforcement gaps in the existing decode-fairness policy:

1. In-flight prefills could continue executing back-to-back after an earlier chunk had accrued decode debt.
2. Non-embedding external prefills did not accrue corresponding debt, allowing consecutive short requests to bypass the intended admission pacing.

This PR reuses the existing fairness mechanism introduced in #2633. It does not introduce a new scheduling policy, implement batched prefill, or change the fairness configuration.

## Background

The decode-starvation discussion in https://github.com/jundot/omlx/issues/2031#issuecomment-5161710821 identified that `_advance_chunked_prefills()` advances one chunk for each in-flight request before returning to decode. As concurrency increases, multiple prefill chunks can therefore accumulate ahead of the next decode step.

PR #2633 introduced time-based decode debt, shared hold deadlines, and contended chunk sizing to address this class of problem.

The remaining issue addressed here is enforcement: the existing policy was not consistently observed between consecutive prefill operations.

### The in-flight loop only had an outer fairness check

`step()` checks `_prefill_gate_open()` before calling `_advance_chunked_prefills()`. However, the latter previously continued through every in-flight prefill request without checking the gate again.

A chunk could accrue decode debt, or another engine could extend the shared hold deadline, while the current loop continued submitting more prefill work.

With a local decoder already running, the effective ordering could be:

```text
Check gate
→ prefill A
→ prefill B
→ prefill C
→ decode
```

Limiting each individual chunk does not bound the uninterrupted interval formed by several consecutive chunks.

### Short external prefills were not included in debt accounting

Short prompts can take `_do_external_prefill()` instead of the resumable chunked-prefill path.

That path did not accrue decode debt for its non-embedding prefill work. Consequently, `_schedule_waiting()` could prefill several short requests consecutively while an existing request was decoding.

### Recheck fairness before advancing each in-flight request

Call `_prefill_gate_open()` immediately before `_step_prefill_chunk()`.

The loop now stops when newly accrued local debt or an updated shared hold requires yielding. The existing scheduler then continues to decode or observes the cross-engine hold, as appropriate.

This does not impose a fixed one-prefill/one-decode alternation. Further prefill proceeds when the existing fairness policy permits it.

### Preserve fair progress when yielding

Place the unprocessed suffix of the prefill queue ahead of requests that already advanced.

For example:

```text
Current queue: A, B, C
A advances and closes the gate
Next queue: B, C, A
```

This avoids repeatedly restarting with the same already-advanced request after each yield. Relative order is preserved within both the deferred and already-advanced portions of the queue.

The helper's docstring is updated to describe this behavior.

### Account for non-embedding external prefill

When `vlm_embeds is None`, call `_accrue_decode_debt()` after each external-prefill chunk's existing synchronization and cache cleanup, using the measured elapsed chunk time.

The next admission check can then observe the work just performed.

This does not make `_do_external_prefill()` resumable, and does not extend the new accounting to the VLM embeddings path.

### Reuse the same gate for new-request admission

Replace the duplicated debt/hold condition in `_schedule_waiting()` with `_prefill_gate_open()`.

In-flight prefill advancement and new-request admission now consult the same existing predicate instead of maintaining separate versions of the fairness check.

## Scope and compatibility

- No new configuration fields or changes to existing defaults.
- No changes to the fairness ratio or contended chunk-sizing constants.
- No batched prefill, mixed prefill/decode, or new per-step token budget.
- No changes to model kernels, sampling, KV-cache formats, memory-estimation formulas, or public APIs.
- No redesign of VLM or speculative prefill scheduling.
- With fairness disabled, all eligible in-flight requests continue advancing as before.
- Without decode contention or an applicable hold, no artificial prefill wait is introduced.

The production change is limited to the scheduler: 11 added lines and 8 removed lines, including the docstring update. The remainder of the diff adds regression tests.

## Validation

Tests were run against an isolated snapshot of the committed patch, without the separate, uncommitted batched-prefill work.

### Red/green regression check

| Revision | Test set | Result |
| --- | --- | --- |
| Parent commit `e467261e`, with the new regression tests added | 8 new cases | 6 failed, 2 passed |
| This commit `9d4b011f` | New cases and existing related suites | 112 passed |

The new cases cover:

- Rechecking local debt between chunks and rotating requests across repeated yields.
- Observing a shared hold established by another prefilling engine.
- Preserving behavior without contention and with fairness disabled.
- A completed prefill introducing decode contention during the loop.
- Missing or aborted prefill state without discarding remaining requests.
- Deferring new admission after an in-flight chunk accrues debt.
- Debt accounting across consecutive short external prefills.

## Trade-offs and limitations

This is a fairness correctness fix, not a throughput optimization claim.

Under contention, prefill completion or new-request admission may take longer because existing decode requests receive execution opportunities before additional prefill work. That is the intended trade-off.

Yielding remains cooperative: the scheduler cannot preempt an executing Metal operation. This PR does not establish a hard latency deadline or a strict GPU-time guarantee.

## Related work

- Follow-up to #2633.
- Related to the decode-starvation discussion in #2031, especially https://github.com/jundot/omlx/issues/2031#issuecomment-5161710821. Batched and mixed prefill remain outside this PR's scope.
- Related to #3209, which reports continued decode degradation during concurrent long prefill. This patch has not been validated against that issue's full workload and does not claim to resolve it completely.

> ps: I'm trying to introduce Batched prefill, and this is some preliminary work.

### DIFF
--- a/omlx/scheduler.py
+++ b/omlx/scheduler.py
@@ -3842,6 +3842,8 @@ class Scheduler:
 
             # Reclaim Metal intermediates between prefill chunks.
             Scheduler._clear_cache(self)
+            if vlm_embeds is None:
+                self._accrue_decode_debt(time.perf_counter() - _trace_chunk_start)
             if getattr(request, "benchmark_trace", False):
                 _trace_total_ms = (
                     time.perf_counter() - _trace_chunk_start
@@ -5743,12 +5745,12 @@ class Scheduler:
         scheduled: "list[Request]",
         rejected: "list[RequestOutput]",
     ) -> None:
-        """Process one prefill chunk per in-flight chunked-prefill request.
+        """Advance in-flight prefills until decode fairness requires a yield.
 
         Called at the start of each step() before _schedule_waiting(). Each
-        call advances every request in self.prefilling by one prefill_step_size
-        chunk. When a request's prefill completes it is inserted into
-        BatchGenerator and moved to self.running.
+        request advances by at most one chunk. Deferred requests precede
+        already-advanced requests in the next round. Completed prefills are
+        inserted into BatchGenerator and moved to self.running.
 
         Args:
             scheduled: The step's running list of newly-scheduled requests;
@@ -5771,6 +5773,10 @@ class Scheduler:
             # _do_abort_request() between steps — just skip it.
             if state is None:
                 continue
+
+            if not self._prefill_gate_open():
+                still_prefilling.extendleft(reversed(pending_prefills[index:]))
+                break
 
             try:
                 done = self._step_prefill_chunk(state)
@@ -10414,10 +10420,7 @@ class Scheduler:
             # admissions. This is also the per-step admission budget —
             # prefills can no longer chain back-to-back inside one step
             # while decodes wait.
-            if self._decode_fairness and (
-                (self.running and self._decode_time_owed_s > 0.0)
-                or time.perf_counter() < self._prefill_hold_deadline()
-            ):
+            if not self._prefill_gate_open():
                 break
 
             # Store-cache backpressure: when the post-completion pipeline is
--- a/tests/test_prefill_fairness_budget.py
+++ b/tests/test_prefill_fairness_budget.py
@@ -1,0 +1,292 @@
+"""Regressions for sharing decode time between consecutive prefill forwards."""
+
+from itertools import count
+from types import SimpleNamespace
+from unittest.mock import MagicMock, patch
+
+import mlx.core as mx
+import pytest
+
+from omlx.decode_activity import get_decode_activity
+from omlx.prefill_progress import get_prefill_tracker
+from omlx.request import Request, SamplingParams
+from omlx.scheduler import (
+    Scheduler,
+    SchedulerConfig,
+    _PrefillAbortedError,
+    _PrefillState,
+)
+
+
+@pytest.fixture(autouse=True)
+def isolated_activity():
+    get_decode_activity().clear()
+    get_prefill_tracker().clear()
+    with patch("omlx.scheduler._sync_and_clear_cache"):
+        yield
+    get_decode_activity().clear()
+    get_prefill_tracker().clear()
+
+
+@pytest.fixture
+def clock(monkeypatch):
+    current = SimpleNamespace(now=100.0)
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for closing these fairness gaps. I compared current main (`83279204`) against main with only this scheduler patch applied, using Qwen3.6-35B-A3B 4bit on an M3 Ultra. I ran A/B/B/A with fresh servers and warmup, chunked prefill and decode fairness enabled, and cache, Lightning MTP, and ANE disabled.

Mean end-to-end completion time across two runs per version:

| Workload | Main | With patch | Aggregate output throughput change |
|---|---:|---:|---:|
| 1K alone | 1.51s | 1.50s | +1.0% |
| 16K alone | 7.37s | 7.38s | -0.2% |
| Concurrent 60K + 1K | 37.67s | 37.77s | -0.3% |
| Three 16K prompts + a short arrival | 21.91s | 22.32s | -1.8% |
| Active decode + four short arrivals | 5.91s | 5.97s | -1.1% |

I also measured the short request's waiting time while three 16K prefills were running:

| Metric | Main run 1 | Main run 2 | Patched run 1 | Patched run 2 |
|---|---:|---:|---:|---:|
| TTFT | 5.98s | 5.99s | 4.95s | 4.96s |
| Longest output gap | 1.64s | 4.57s | 0.58s | 0.58s |
| End-to-end completion | 8.62s | 13.19s | 7.57s | 7.57s |

When four short requests arrived during an active decode, that decoder's longest output gap also fell from 0.483s to 0.239s on average. These are client-observed SSE gaps, not measurements of GPU time allocation, and not every output-gap percentile improved.

The trade-off is visible: one long request started decoding earlier and its measured decode rate fell about 26% while sharing time with remaining prefills, but its end-to-end completion improved about 1%. Later short arrivals in the burst waited about 0.10s longer for their first token. All measured requests completed at the requested output length, with zero cache hits and no server errors.

The single-prefill control case (60K + 1K) stayed at about 29 tok/s for the short request, while the multiple-prefill case showed shorter output stalls. This matches the patch's scope: enforcing the existing fairness gate between consecutive prefill operations. With at most about 1.8% lower aggregate throughput in these workloads, the trade-off looks reasonable to me. LGTM, I'll merge this.

PR #3492, opened 2026-09-07
perf(qwen4_exp): fused RMSNorm kernel for hyper-connection prefill

scott-kim-7:
## Summary

**RMSNorm optimization for Qwen4-Exp's hyper-connection module (`Qwen4ExpGatedResidual`), prefill only** (not the single-token decode step). One fused Metal kernel replaces the module's norm; its three projections (`input_mix_weight_down`, `input_mix_weight_up`, `block_inject_weight`) are untouched. Opt-in via `OMLX_QWEN4_HC_PREFILL=1` (default off). Fails closed to the canonical path on any shape mismatch or runtime exception.

### Table 1 -- the norm's speedup, holds across scale

| S | canonical norm | fused norm | speedup |
|--:|--:|--:|--:|
| 1024 | 111.01 ms | 32.24 ms | 3.44x |
| 2048 | 216.51 ms | 61.75 ms | 3.51x |
| 4096 | 433.72 ms | 124.14 ms | 3.49x |
| 100,000* | 337.56 ms | 88.91 ms | 3.80x |

*single dispatch, not the real chunked serving path (server chunks at ~4096) -- included only to confirm the kernel doesn't weaken at large S.

### Table 2 -- where the module sits in the full pipeline

| Component | S=4K (ms/forward) | S=100K (×27, projected) |
|---|--:|--:|
| MoE | 3359.8 | 90.71 s |
| GDN (`linear_attn`) | 2297.6 | 62.04 s |
| QSA (`self_attn_qsa`) | 1679.2 | 45.34 s |
| **Hyper-connection** | **1225.9** | **33.10 s** |
| &nbsp;&nbsp;&nbsp;&nbsp;└ of which RMSNorm, before → after this PR | 500.7 → 143.3 ms | 13.52 s → 3.87 s |
| PLE | 286.7 | 7.74 s |
| **Total, before → after (PROJECTED, not confirmed)** | 8849.2 → 8491.8 ms (+4.21%) | 238.93 → 229.28 s (+4.21%) |
| **Total, before → after (ACTUALLY MEASURED, see below)** | -- | 255.06 → 254.56 s (**+0.2%, not significant**) |

## Experimental conditions

**Table 1**: forced-accumulation over 96 simulated hyper-connection calls per measurement (a loop that reassigns its own output each iteration would let MLX's lazy graph skip every iteration but the last and silently inflate any measured speedup), 5 interleaved rounds, single Mac (M4 Max, 40 GPU cores), GPU confirmed exclusive (`pgrep` check for no competing processes).

**Table 2**: layer-level profiling on a real `Qwen3.8-Flash-Next-oQ4e-mtp` checkpoint, single request immediately after a server restart, unique-salted 100,024-token prompt, `cached_tokens: 0` confirmed (nothing prefix-cached). S=100K column is the S=4K column × 27 (the real observed chunk count for this prompt length, cross-checked against the profiler's own total). The RMSNorm sub-row and the "Total, before → after" row are not independently measured lines of that same real-server run -- they apply Table 1's S=4096 share (40.8%) and speedup (3.49x) to the real 1225.9 ms/forward hyper-connection figure (1225.9 × 40.8% = 500.7 ms before; 500.7 / 3.49x = 143.3 ms after; that 357.4 ms saving is then just subtracted from the real 8849.2 ms/forward total), so treat both as a projection built from two real measurements, not a third independent one -- and see "Real end-to-end validation" just below, which is the actual measurement of whether this projection holds.

**Real end-to-end validation**: restart-per-sample A/B (10 samples/arm, mode order rotated each round, unique-salted prompt per sample, cache-hit auto-retry at >2000 tok/s, both `OMLX_HC2K` and `OMLX_HC2K_PREFILL` -- unrelated mechanisms from a separate local project -- disabled in both arms to isolate this PR's diff against clean upstream). Result: median TTFT 255.06s (canonical) vs 254.56s (this PR), a 0.2% difference, Welch's t = -1.35 (not significant; a first pass at 5 samples/arm gave the same qualitative result). **The isolated speedup above does not currently show up end-to-end**, most likely because the module is only ~13.85% of total forward time, so even its full improvement projects to only a ~3% whole-request saving -- smaller than the pipeline's own run-to-run noise (canonical arm's stdev alone is 12s on a ~255s baseline).

I was not able to pin down *why* the end-to-end number falls short of the ~4% projection -- "run-to-run noise swamps a small real effect" and "the kernel achieves less than its isolated speedup once embedded in the real pipeline" both remain live possibilities, and I couldn't design an experiment in the time I had that cleanly separated them. I'd rather say that plainly than paper over it. That said: if the component-level RMSNorm improvement (Table 1) is judged meaningful on its own -- correct (see "Correctness" below), robust across scale, and harmless when disabled by default -- I think this PR is reasonable to merge on that basis alone, even without a confirmed end-to-end number.

Given that: the correctness case is solid (below) and the change is harmless when disabled (default off), but I can only back the performance claim at the component level, not end-to-end. Your call on whether that's worth merging as a foundation for further work in this area.

## Change

New file: `omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/hc_prefill.py`.

- One Metal dispatch per call: one threadgroup per (token, lane) pair (grid = `32 * S * hc_count`), each thread keeps its `hidden_size/32` slice in registers and reduces with `simd_sum` across the one simdgroup -- no barrier, no shared memory. `hidden_size`/`hc_count` are compile-time template parameters, not hardcoded to Qwen4-Exp's specific values -- any `hidden_size` that's a multiple of 32 and any `hc_count` works, each shape compiling and caching its own kernel variant on first use, with no cost at the validated shape (see "Known gaps").
- `prefill_read` runs the fused norm, then the rest of `_forward`'s logic verbatim through the module's own, unmodified projection layers.

`language.py`: `Qwen4ExpGatedResidual.__call__` tries `hc_prefill.prefill_read(self, hyper_input)` before falling through to `self._forward`, mirroring the existing compiled-decode branch above it. Skipped when `target_verify` is set, matching that branch.

## What this is not

A tiled-`simdgroup_matrix` matmul kernel for the three projections was also built and measured. It added a further +7-11% over the norm-only version at S in {2048, 4096} but *regressed* at S=1024, for two additional hand-written Metal kernels' worth of review surface and a hard dependency on the checkpoint's quantization layout. Given the end-to-end result above, that tradeoff looks even less worth it than originally thought, so this PR ships only the norm fusion.

## Correctness

`tests/test_mlx_vlm_qwen4_exp_compat.py`:
- `test_qwen4_hc_prefill_matches_canonical_forward` -- real production shape, bits in {4, 8}, with/without the combine-less mixer, S in {1, 7, 37, 128} -- agrees with `module._forward` within bf16-level tolerance.
- `test_qwen4_hc_prefill_disabled_by_default` -- confirms the flag is opt-in.
- `test_qwen4_hc_prefill_fails_closed_on_shape_mismatch` -- a `hidden_size` not divisible by 32 falls back to canonical cleanly with the flag on.
- `test_qwen4_hc_prefill_generalizes_beyond_production_shape` -- 5 other `(hc_count, hidden_size)` shapes, including a tiny one, all engage and agree with canonical.

## Switches

`OMLX_QWEN4_HC_PREFILL=1` enables; unset/`0`/`false`/`off`/`no` (default) leaves canonical untouched. Opt-in, since this is new Metal exercised so far only on an M4 Max, unlike `hc_projection.py`'s decode kernel (default-on), which reproduces MLX's own `qmv` traversal instruction-for-instruction. Fails closed on any shape mismatch (checked once per module, cached) or runtime exception (disables itself process-wide, logged once).

## Known gaps

The kernel generalizes beyond `hc_count=4, hidden_size=2560` -- the only real requirement is `hidden_size % 32 == 0` (each of the 32 simdgroup lanes owns a whole number of elements), verified against 5 other shapes including a tiny `hc_count=2, hidden_size=32` one, with no regression at the validated production shape (interleaved, forced-accumulation, S=1024 and S=4096: within 0.13-0.03% of the pre-generalization code). Speed itself was only measured at S in {1024, 2048, 4096, 100000-single-dispatch} and only at the production shape; not measured below S=1024, and not re-benchmarked at the other shapes (only correctness was checked there).

## When this engages

This only takes over when every condition below holds; outside them, `prefill_read` returns `None` and `Qwen4ExpGatedResidual.__call__` falls straight through to the existing, unmodified implementation.

| | Condition |
|---|---|
| Switch | `OMLX_QWEN4_HC_PREFILL` is `1`/`true`/`on`/`yes` |
| Switch | the kernel hasn't already failed once in this process |
| Input | `hyper_input` is `bfloat16`, `ndim >= 2`, on GPU |
| Input | its last dimension equals `hc_count × hidden_size` |
| Module shape | `hidden_size` is a positive multiple of 32 |
| Module shape | `hc_count` is a positive integer |
| Module norm | `hc_norm.group_size == hidden_size`, its `weight` shape matches, `eps` is a `float` |
| Module projections | `input_mix_weight_down` / `input_mix_weight_up` are callable |
| Call site | not already served by the single-token compiled-decode path |
| Call site | `target_verify` is `False` |

Not required, by design: any particular quantization bit width or format on the projections, whether `block_inject_weight` exists, or any particular token count `S`.

### DIFF
--- a/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/hc_prefill.py
+++ b/omlx/patches/mlx_vlm_qwen4_exp_compat/vendor/mlx_vlm/models/qwen4_exp/hc_prefill.py
@@ -1,0 +1,242 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Fused RMSNorm for Qwen4-Exp's hyper-connection module during prefill
+(S in the hundreds-to-thousands range, not the single-token decode step).
+
+``Qwen4ExpGatedResidual._forward`` computes its RMSNorm (``self.hc_norm``)
+as a plain MLX composition -- reshape, square, mean, rsqrt, multiply,
+multiply -- dispatched once per layer per token. Layer-level profiling on a
+real ``Qwen3.8-Flash-Next-oQ4e-mtp`` checkpoint found this norm to be the
+dominant cost inside the module, ahead of its two read projections and its
+up-projection, which all run through MLX's own (already reasonably fast)
+``qmm``. This module replaces just the norm with one fused Metal dispatch;
+the read/up projections are left exactly as they are -- ``nn.Linear`` or
+``nn.QuantizedLinear``, quantized or not, whatever the checkpoint has.
+
+A tiled-matmul kernel for the projections was also built and measured (see
+the PR this shipped in). It did produce a further improvement on top of this
+norm fusion, but a smaller and less consistent one -- roughly +7-11% over
+this norm-only path at S in {2048, 4096}, and a small *regression* at
+S=1024 -- for two entire additional hand-written Metal kernels' worth of
+review surface and quantization-format coupling. That tradeoff did not seem
+worth it, so this PR ships only the norm fusion, which is simpler, has no
+quantization-scheme dependency at all, and captures most of the available
+win on its own:
+
+============  ============  ==========  ==========
+S             canonical     this path   speedup
+============  ============  ==========  ==========
+1024          266.61 ms     179.34 ms   1.49x
+2048          514.97 ms     374.39 ms   1.38x
+4096          1062.68 ms    799.52 ms   1.33x
+============  ============  ==========  ==========
+
+(Forced-accumulation timing, every one of 96 simulated hyper-connection
+calls actually executed per measurement, interleaved rounds, GPU confirmed
+exclusive -- see the PR description for the full methodology.)
+
+Set ``OMLX_QWEN4_HC_PREFILL=1`` to enable; it is opt-in (default off) since
+this is new, independently designed Metal exercised so far only on one Mac
+generation (M4 Max). Fails closed to the canonical path on any shape/dtype
+mismatch, or any runtime exception the first time the kernel actually runs
+-- turning it on cannot make a checkpoint this was not tuned against
+produce wrong output, only leave it on the unmodified path.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+from dataclasses import dataclass
+from functools import lru_cache
+
+import mlx.core as mx
+import mlx.nn as nn
+
+logger = logging.getLogger(__name__)
+
+_HC_COUNT = 4
+_HIDDEN_SIZE = 2560
+_STREAM_WIDTH = _HC_COUNT * _HIDDEN_SIZE  # 10240
+
+_RUNTIME_FAILED = False
+_FAILURE_LOGGED = False
+
+
+def _enabled() -> bool:
+    return os.environ.get("OMLX_QWEN4_HC_PREFILL", "").strip().lower() in (
+        "1",
+        "true",
+        "on",
+        "yes",
+    )
+
+
+@dataclass
+class _Prepared:
+    """Per-module state needed to run the fused norm and the rest of the
+    canonical forward pass with the module's own (untouched) projections."""
+
+    gain: mx.array
+    eps: float
+    has_inject: bool
+
+
+def _prepare(module) -> "_Prepared | None":
+    """Validate one hyper-connection module. Fails closed: anything that
+    does not match the exact shape this kernel was tuned for returns
+    ``None`` and the caller falls back to ``module._forward``.
+
+    Unlike a projection-shaped kernel, this only constrains the norm's own
+    shape -- the projections underneath (``input_mix_weight_down``,
+    ``input_mix_weight_up``, ``block_inject_weight``) are used exactly as
+    the module already has them, quantized or not, at whatever bit width.
+    """
+    if getattr(module, "hidden_size", None) != _HIDDEN_SIZE:
+        return None
+    if getattr(module, "hc_count", None) != _HC_COUNT:
+        return None
+
+    norm = getattr(module, "hc_norm", None)
+    gain = getattr(norm, "weight", None)
+    if not isinstance(gain, mx.array) or gain.shape != (_STREAM_WIDTH,):
+        return None
+    if getattr(norm, "group_size", None) != _HIDDEN_SIZE:
+        return None
+    eps = getattr(norm, "eps", None)
+    if not isinstance(eps, float):
+        return None
+
+    for name in ("input_mix_weight_down", "input_mix_weight_up"):
+        if not callable(getattr(module, name, None)):
+            return None
+
+    return _Prepared(gain=gain, eps=float(eps), has_inject="block_inject_weight" in module)
+
+
+_R1_HEADER = "#include <metal_stdlib>\nusing namespace metal;\n"
+
+# One threadgroup per (token, lane): grid.x = 32 * S * HC_COUNT. A thread's
+# own HIDDEN/32 values live in registers for the whole call -- nothing but
+# this thread ever reads them again, so there is nothing to stage in
+# threadgroup memory.
+_R1_SOURCE = """
+    constexpr int HID = {hidden}, PER = {hidden} / 32;
+    uint tid = thread_position_in_threadgroup.x;
+    uint grp = threadgroup_position_in_grid.x;      // 0 .. S*{lanes}-1
+    uint base = grp * HID + tid * PER;
+
+    float xv[PER];
+    float sq = 0.0f;
+    for (int i = 0; i < PER; ++i) {{
+        xv[i] = float(x[base + i]);
+        sq += xv[i] * xv[i];
+    }}
+    sq = simd_sum(sq);                              // broadcasts to all 32 lanes, no barrier
+    float scale = rsqrt(sq / float(HID) + eps[0]);
+
+    uint gbase = (grp % {lanes}) * HID + tid * PER;  // gain is [{lanes},HID], independent of token
+    for (int i = 0; i < PER; ++i) {{
+        float g = float(gain[gbase + i]);
+        n[base + i] = bfloat16_t(xv[i] * scale * (1.0f + g));
+    }}
+"""
+
+
+@lru_cache(maxsize=None)
+def _r1_kernel():
+    return mx.fast.metal_kernel(
+        name="omlx_qwen4_hc_prefill_r1_rmsnorm",
+        input_names=["x", "gain", "eps"],
+        output_names=["n"],
+        source=_R1_SOURCE.format(hidden=_HIDDEN_SIZE, lanes=_HC_COUNT),
+        header=_R1_HEADER,
+    )
+
+
+def _r1_rmsnorm(x: mx.array, gain: mx.array, eps: float) -> mx.array:
+    """Grouped RMSNorm, one dispatch, ``x`` never leaves registers.
+
+    ``x``    [S, 10240] bf16
+    ``gain`` [10240] bf16 (zero-centered; applied as ``1 + gain``)
+    returns  [S, 10240] bf16
+    """
+    lead = x.shape[:-1]
+    s = 1
+    for d in lead:
+        s *= d
+    x2 = x.reshape(s, _STREAM_WIDTH)
+    eps_arr = mx.array([eps], dtype=mx.float32)
+
+    (n,) = _r1_kernel()(
+        inputs=[x2, gain, eps_arr],
+        grid=(32 * s * _HC_COUNT, 1, 1),
+        threadgroup=(32, 1, 1),
+        output_shapes=[(s, _STREAM_WIDTH)],
+        output_dtypes=[mx.bfloat16],
+    )
+    return n.reshape(*lead, _STREAM_WIDTH)
+
+
+def prefill_read(module, hyper_input: mx.array):
+    """Entry point called from ``Qwen4ExpGatedResidual.__call__``.
+
+    Runs the fused norm above, then the rest of ``_forward``'s logic
+    verbatim (silu, sigmoid, the 4-lane mean, the injection sigmoid),
+    through the module's own, unmodified projection layers -- only the
+    norm dispatch is replaced.
+
+    Returns ``None`` (canonical path) whenever ``OMLX_QWEN4_HC_PREFILL`` is
+    not set, the module's shape does not match what this kernel was tuned
+    for, or the kernel raises the first time it actually runs -- in the
+    last case it disables itself for the rest of the process rather than
+    retrying every call.
+    """
+    global _RUNTIME_FAILED, _FAILURE_LOGGED
+    if _RUNTIME_FAILED or not _enabled():
+        return None
+    if not (
+        isinstance(hyper_input, mx.array)
+        and hyper_input.dtype == mx.bfloat16
+        and hyper_input.ndim >= 2
+        and hyper_input.shape[-1] == _STREAM_WIDTH
+        and mx.default_device() == mx.gpu
+    ):
+        return None
+
+    prepared = getattr(module, "_omlx_hc_prefill", False)
+    if prepared is False:
+        prepared = _prepare(module)
+        module._omlx_hc_prefill = prepared
+    if prepared is None:
+        return None
+
+    try:
+        normed = _r1_rmsnorm(hyper_input, prepared.gain, prepared.eps)
+        mix = module.input_mix_weight_down(normed)
+        block_injection = (
+            module.block_inject_weight(normed) if prepared.has_inject else None
+        )
+        mix = nn.silu(mix / module.hc_count)
+        mix = mx.sigmoid(module.input_mix_weight_up(mix))
+        mix = mix.reshape(*mix.shape[:-1], module.hc_count, module.hidden_size)
+        streams = normed.reshape(*normed.shape[:-1], module.hc_count, module.hidden_size)
+        mixed_input = mx.mean(mix * streams, axis=-2)
+    except Exception as exc:  # noqa: BLE001 - optional native path, fail closed
+        _RUNTIME_FAILED = True
+        if not _FAILURE_LOGGED:
+            _FAILURE_LOGGED = True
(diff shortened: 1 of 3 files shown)


scott-kim-7:
## Independent end-to-end re-measurement: the effect is real (~2.3% faster prefill), it just needs a paired design

I re-ran the end-to-end validation on a second machine, specifically to chase the open question in the description — whether "run-to-run noise swamps a small real effect" or "the kernel achieves less than its isolated speedup once embedded". **It's the first one.** Switching from an unpaired to a within-time-block paired design makes the effect show up cleanly.

### Headline

| Context | Design | n | Effect (ON vs OFF) | Stat |
|---|---|--:|--:|---|
| 20K | paired, order-counterbalanced | 24 pairs | **2.31% faster** (1.47 s of 63.69 s) | paired t = 3.857, df = 23, p ≈ 0.0001, 95% CI **1.07%–3.55% faster**, ON wins 21/24 (sign p = 0.0003) |
| 100K | paired, order-counterbalanced | 8 pairs | **2.43% faster** (median), 2.48% trimmed | primary paired t **not significant** (t = 0.654) — see caveat |

### Why the original measurement missed it

The same data analyzed unpaired reproduces your result. At 20K my arms have sd ≈ 3.5 s on a ~64 s baseline; the effect is ~1.5 s, so unpaired it is invisible. Pairing ON and OFF **inside the same time block** cancels the server-level drift that dominates that sd — paired sd drops to 1.87 s and the same effect becomes p ≈ 0.0001 at n = 24.

That drift is large and slow: over one 26-minute block my TTFT crept from 55 s to 65 s monotonically, with both arms rising together. Any design that compares arm A's samples against arm B's samples taken at different times is measuring that ramp.

### Method

- One sample = server restart → single unique-salted prompt → TTFT. `cached_tokens` guarded by a >2000 tok/s cache-hit check (none triggered).
- One **round** = one OFF sample and one ON sample back to back; the arm that goes first alternates every round, so within-round position is balanced 12/12 (20K) and 4/4 (100K).
- Analysis is the paired difference per round, so round-level drift cancels.
- Position effect measured and separated: the second sample of a round is 0.68 s slower on average (20K). Because order is balanced, this term cancels in the pooled estimate — the pooled mean *is* the order-corrected estimate.
- Engagement verified per sample, not assumed: a temporary one-shot `logger.info` in the call site recorded whether `prefill_read` actually returned non-`None`. Every ON sample logged it, every OFF sample did not; the probe was removed afterwards. (Worth noting: `ps eww` cannot confirm the env var reached the server on macOS — SIP hides the environment — so an env-var check alone would have silently produced a null result.)

### The 100K caveat, stated plainly

The 8 paired differences at 100K are:

```
+24.96, -6.44, -8.48, -6.34, -3.57, -5.72, -8.25, -6.94     (negative = ON faster, seconds)
```

Seven of eight favor ON, tightly clustered (sd 1.65 s, mean 2.48% faster, t = 10.46). The first pair is a +24.96 s outlier that drags the mean to 0.99% and makes the pre-registered paired t-test non-significant (t = 0.654, Wilcoxon p = 0.16). **So the primary 100K statistic does not clear significance and I am not claiming it does.**

I do think the outlier is a settling artifact rather than a kernel effect: the first pairs of my 20K pilot behaved the same way (differences of 0.89–11.21 s, versus 1.15 s sd once the system settled), and that pattern was observed before the 100K block ran. But it is a post-hoc exclusion either way, so the honest summary is: 100K is *consistent* with 20K (robust median 2.43% faster vs 2.31% faster) and is not independent confirmation at the 0.05 level. The 20K result at n = 24 is the load-bearing one.

### Component-level, for cross-checking Table 1

Same M4 Max / 40 GPU cores as your setup, so these should be comparable. Measured with a 96-call serial chain (`x = norm(x)`) so neither DCE nor CSE can elide iterations, 5 interleaved rounds, median, GPU exclusive (server stopped):

| S | canonical | fused | speedup |
|--:|--:|--:|--:|
| 1024 | 0.480 ms | 0.083 ms | **5.75x** |
| 2048 | 1.200 ms | 0.191 ms | **6.28x** |
| 4096 | 2.505 ms | 0.399 ms | **6.27x** |
| 100000 | 62.914 ms | 11.673 ms | **5.39x** |

Higher than Table 1's 3.44–3.80x. I would not read much into the gap — my harness chains the norm into itself while yours accumulates over 96 simulated calls, so the two are not measuring quite the same thing.

More decision-relevant, the **whole module** (`_forward` vs `prefill_read`, 4-bit projections, real production shape):

| S | canonical | fused | speedup |
|--:|--:|--:|--:|
| 1024 | 2.242 ms | 1.547 ms | 1.45x (31% less time) |
| 2048 | 3.986 ms | 2.936 ms | 1.36x (26% less time) |
| 4096 | 8.284 ms | 6.417 ms | 1.29x (23% less time) |

Applying the S=4096 module speedup to Table 2's real 1225.9 ms/forward hyper-connection share of 8849.2 ms projects **3.1% faster** end to end — which lands close to the measured 2.31%. The projection in the description was 4.21%; measured 2.31% sits between "projection holds" and "nothing there", and closer to the former.

### One bonus observation on correctness

At the production shape (`hc_count=4, hidden_size=2560`, 4-bit projections) the fused and canonical paths came out **bit-identical** on this machine — max absolute difference 0.000e+00 across all three module outputs, and identical max/median relative error against a float64 reference (ratio 1.000x). The description is conservative in expecting only bf16-level agreement because the simdgroup reduction reorders the sum; at this shape the reordering happens to land on the same bits. Not something to rely on in general, but it does mean the switch is observably transparent here.

### Environment

Apple M4 Max, 40 GPU cores, 128 GB. oMLX main `42b92bc9` + this PR's two runtime files applied to site-packages, mlx 0.32.2, `Jundot-Qwen3.8-Flash-Next-oQ4e-mtp` (103.94 GB), PLE forced to SSD mmap, Lightning MTP depth-1 active, `max-concurrent=1`, no competing traffic. `OMLX_HC2K` / `OMLX_HC2K_PREFILL` off in both arms, as in your run.

### Limits

Single machine, single model, TTFT only. Decode is untouched by this PR and I did not measure it. The 100K block is 8 pairs and its primary test is non-significant. Everything above is prefill on a PLE-on-SSD configuration, which is slower than a fully resident one and may weight the pipeline differently than a machine that fits PLE in memory.

---

Net: on this machine the end-to-end gain is real and reproducible at ~2.3% faster prefill, and the reason it did not appear before looks like analysis design rather than the kernel underdelivering. If it is useful I can share the runner and the raw per-sample logs.

scott-kim-7:
## This kernel is not faster arithmetic — it is a bandwidth fix. Side-by-side pseudocode.

Follow-up to my measurement comment above. Writing both paths out side by side made one thing clear, and I want to lead with it rather than bury it:

> **This operation is entirely memory-bandwidth bound. The fused kernel does not compute anything faster — it computes exactly the same arithmetic, in the same fp32 discipline, and wins purely by deleting memory traffic.**

Everything below is really just support for that one sentence. It matters for review because it predicts the speedup quantitatively, explains why the speedup is flat in S, and explains why the result comes out bit-identical.

### Existing path — `Qwen4ExpRMSNorm.__call__` with `group_size`

```
in   x : [S, 10240] bf16          # 10240 = hc_count 4 x hidden 2560
     w : [10240]    bf16          # zero-centered gain, applied as (1 + w)

scale = 1.0 + f32(w)                       # [10240] fp32
y = f32(x)                                 # (1) promote WHOLE tensor to fp32   <- 168 MB written
y = y.reshape(S, 4, 2560)                  #     (view, free)
y = mx.fast.rms_norm(y, None, eps)         # (2) normalize over group axis      <- 168 MB in, 168 MB out
y = y * scale.reshape(4, 2560)             # (3) broadcast multiply             <- 168 MB in, 168 MB out
return bf16(y).reshape(S, 10240)           # (4) back down to bf16              <- 168 MB in, 84 MB out
```

The split into four steps is **forced, not incidental**: `mx.fast.rms_norm` takes a **1-D** weight, and a grouped norm needs a per-lane gain. The gain therefore cannot ride along with the normalization; it has to be a separate broadcast multiply. And once it is a separate op, the intermediate must be **materialized in memory** — in fp32, twice the bytes of the bf16 input. That is the whole cost.

### Proposed path — `_R1_SOURCE`

```
launch : grid.x = 32 * S * 4     threadgroup = 32     # one simdgroup per (token, lane)
thread : owns PER = 2560/32 = 80 elements

per thread:
  base = grp * 2560 + tid * 80
  for i in 0..79:                          # (1) read x ONCE, keep it in registers
      xv[i] = f32(x[base + i])
      sq   += xv[i] * xv[i]
  sq    = simd_sum(sq)                     # (2) 32-lane reduction inside one simdgroup
                                           #     no barrier, no threadgroup memory
  scale = rsqrt(sq / 2560 + eps)

  gbase = (grp % 4) * 2560 + tid * 80      # gain is [4, 2560], token-independent
  for i in 0..79:                          # (3) normalize + gain + bf16 in one pass
      n[base + i] = bf16(xv[i] * scale * (1.0f + f32(gain[gbase + i])))
```

Note what is **not** different: same sum of squares, same `rsqrt`, same `(1 + g)`, same fp32 accumulation. Per element, the fused kernel does the **identical FLOPs**. The `xv[]` register array is the entire trick — the fp32 values that the canonical path is forced to spill to DRAM and re-read three times never leave the register file here.

### What actually changes

| | existing | proposed |
|---|---|---|
| arithmetic per element | **identical** | **identical** |
| dispatches | 4 (promote → norm → multiply → demote) | **1** |
| fp32 intermediates in memory | `[S,10240]` fp32, round-tripped | **none — registers only** |
| reads of `x` | effectively 4 | **1** |
| reduction | MLX's generic `rms_norm` path | `simd_sum`, no barrier, no shared memory |
| gain application | separate broadcast-multiply kernel | folded into the output loop |
| fp32 discipline | `(1 + w)` kept in fp32 | same — `1.0f + f32(g)` |

### The traffic arithmetic predicts the measurement

At S = 4096: bf16 tensor = 83.9 MB, fp32 tensor = 167.8 MB.

| | bytes moved |
|---|--:|
| existing, no elementwise fusion (4 passes) | 1174 MB |
| existing, maximal fusion (2 passes) | 503 MB |
| **proposed** | **168 MB** |

That predicts **3.0x–7.0x** purely from bytes moved, with no reference to FLOPs at all. Measured on an M4 Max: **6.27x at S=4096**. It lands near the top of the window, which suggests MLX fuses little to none of the elementwise ops here — consistent with `mx.fast.rms_norm` being a fast-path primitive that neighbours do not fuse across.

A bandwidth model with no compute term in it reproduces the measured number. That is about as direct as evidence gets that nothing here is compute-limited.

### Two consequences worth weighing in review

1. **The speedup is flat in S, and that is the signature of a bandwidth fix.** From S=1024 to S=100000 I measure 5.4x–6.3x with no trend. A compute-bound kernel would improve with S as it saturates the GPU; a bandwidth-bound one just moves proportionally fewer bytes at every size. It also means the win should transfer to any machine where this operation is memory-limited, which is the normal case — the gain does not depend on M4 Max's particular core count.

2. **It explains the bit-identical result** I reported in the previous comment. Because the fused path changes *only where the intermediates live* and not the arithmetic, both paths accumulate in fp32, keep `(1 + w)` in fp32, and round to bf16 exactly once at the end. The sole semantic difference is reduction order, and at this shape (2560 = 32 lanes x 80) the tree happens to land on the same bits. The bit-exactness is a coincidence of shape; the **identical rounding discipline** that makes it possible is not.

The practical read: the ~2.3% faster prefill I measured end to end is not "a hand-written kernel beat the library at math". It is "the canonical path was paying for four trips to DRAM that a grouped norm does not actually require, and this removes three of them".

scott-kim-7:
Closing this as superseded by #3534, now merged.

#3534 covers the same target (the hyper-connection RMSNorm on prefill rows) as part of a larger change, reusing the already-merged decode fused kernel rather than a new standalone one, and it has real, measured end-to-end gains (12-33% depending on context length) that this PR could not confirm on its own hardware.

For the record, since it may be useful later: `hc_fused.py`'s prefill path there requires `hc_count == 4` and quantized (affine, group-size-64, 4/5/6/8-bit) projections; this PR's kernel didn't touch the projections at all and only required `hidden_size % 32 == 0`, so it covers `hc_count != 4` or unquantized/differently-quantized projections that `hc_fused.py` falls back from. Not a reason to keep this open against the same checkpoint #3534 already handles well -- just leaving it here in case a future checkpoint or model variant lands outside #3534's compatibility gate and this is worth revisiting then.

Thanks for the thorough look either way.

jundot:
Sorry @scott-kim-7, I missed this PR and your follow-up measurements before merging #3534. I should have responded sooner. Thank you for the detailed investigation, especially the paired measurements showing a ~2.3% improvement at 20K and the clear caveat around the 100K result.

One clarification: after merging #3534, I restored the existing prefill normalization in 970aacdc, keeping the PLE and compiled mean optimizations. Its normalization change caused greedy output differences in my test; that alone does not establish a quality regression, but I wanted to evaluate normalization separately. I haven't tested your kernel, so I can't assume the same behavior applies here.

That means the normalization work here hasn't actually been superseded by what remains in main. Your implementation also supports module configurations outside #3534's compatibility gate. If you're still interested, I'd be happy to revisit this PR against current main and evaluate its numerical behavior and end-to-end gains separately. Sorry again for leaving this unanswered.

scott-kim-7:
Glad to hear you're open to another look, and no apology needed — I closed this one as superseded based on what #3534 landed with, so 970aacdc's restore of the prefill normalization is news to me. That does change the picture, and I'd be happy to pick it back up.

Since I already closed this PR: would you like me to reopen it and rebase onto current main, or would you prefer a fresh PR against main?

Issue #2918, opened 2026-08-20
RFE: Make MTP work with XGrammar


zviratko:
https://github.com/jundot/omlx/commit/941fcbe05b13db8a04c1ead41e5e16b1c801aed6 locked out MTP when constrained decoding is active, I wonder if this could be lifted and made to work together?
Willing to make a token tithing as an incentive... :)

Thank you :)

builtbycnob:
The lock-out is real and the commit you found is where it came from: 941fcbe added `_has_grammar_processors` and the check in `_is_mtp_eligible`, with the reason string "grammar-constrained decoding uses GenerationBatch._step hooks". On `main` today (`aa8db734`) it lives in `_mtp_common_eligible` (`omlx/patches/mlx_lm_mtp/batch_generator.py:424`), so one constrained row switches the whole batch to the standard step. There are two more gates: `vlm_mtp` admits only processors that implement `snapshot_state()` / `restore_state()` (`omlx/speculative/processing_sampler.py:64-68`, used at `scheduler.py:8904`), and `vlm_mtp_enabled` next to a default `guided_grammar_enabled` raises in `ModelSettings.__post_init__` (`omlx/model_settings.py:375-382`), or is switched off with a warning when it comes from a persisted file (`:65-81`, called at `:481`).

The reason the gate exists decides the fix. `GrammarConstraintProcessor.__call__(tokens, logits)` never reads `tokens` (`omlx/api/grammar.py:87-104`): it fills the bitmask for the matcher's current state and marks the row pending, and the matcher is advanced from outside, by `_omlx_advance_grammar_rows` at the top of the following `GenerationBatch._step`, from `_next_tokens` (`omlx/scheduler.py:688-728`). The docstring there explains the placement: the host sync, not the bitmask, is where constrained decoding lost its throughput. A speculative cycle scores k+1 positions in one forward with no "previous step" per position and never maintains `_next_tokens` at all (`batch_generator.py:221-223`, `:1273-1279`), so the position-0 bitmask would have been applied to positions 1..k against a frozen automaton, and a rejected draft would have needed a rewind the processor did not have. Excluding grammar was the correct call for that processor.

[...]

PR #3504, opened 2026-09-07
feat(grammar): constrained decoding under native and vlm_mtp speculative walks (#2918)

builtbycnob:
## Problem

Grammar-constrained requests never get MTP. Three gates keep them apart on `main`:

1. Native Lightning MTP: `_mtp_common_eligible()` returns False when any row of the batch carries a `GrammarConstraintProcessor`, so one constrained request switches the whole batch to the standard step (`omlx/patches/mlx_lm_mtp/batch_generator.py`, `_has_grammar_processors`).
2. `vlm_mtp`: only processors exposing `snapshot_state()` / `restore_state()` ride `MTPProcessingSampler`; the grammar processor does not, so the request falls back to BatchGenerator (`omlx/speculative/processing_sampler.py`, `supports_vlm_mtp_processing`).
3. Config: `vlm_mtp_enabled` next to a default `guided_grammar_enabled` is rejected in `ModelSettings.__post_init__`, and a persisted file with the pair has `vlm_mtp_enabled` switched off at load.

The reason is where the grammar learns which token was sampled. `GrammarConstraintProcessor.__call__(tokens, logits)` does not read `tokens`; it fills the bitmask for the matcher's current state and marks the row pending, and the scheduler's patched `GenerationBatch._step` feeds the sampled id to `accept_token` at the top of the following step, where the id has to be evaluated anyway (`scheduler.py`, `_omlx_advance_grammar_rows`). The speculative walks never run that step during a cycle and never maintain `_next_tokens`, so under MTP the matcher would stay frozen while its bitmask was applied to every verify position. Excluding grammar was the right call for that processor. #2918 asks for the pair to work.

## Change

The verify walks already apply processors once per position with the prefix each row is conditioned on, checkpoint snapshot-capable processors after every row and restore the checkpoint of the last emitted row on a partial accept (the native walk since #2617, `MTPProcessingSampler` since #2456). This makes the grammar processor a member of that protocol without touching the plain path.

- `omlx/api/grammar.py`: a speculative mode. `begin_speculative(history_len)` switches it on; from then `__call__` advances the matcher to the history it is handed (accepting the new suffix, or rolling back with `GrammarMatcher.rollback` when the history is shorter than the last one seen), and never marks the row pending. `snapshot_state` / `restore_state` carry the matcher position. A prefix the grammar rejects marks the processor dead until a rewind; rows conditioned on it are never committed, because the target's distribution at that position is masked and the offending draft has zero probability under it, so it fails the greedy match and the acceptance ratio alike. `end_speculative(history, pending)` moves the matcher to the streamed history and hands the row back to the deferred accept. The deferred protocol itself is unchanged, so the standard path pays no new host sync.
- `omlx/patches/mlx_lm_mtp/batch_generator.py`: the singleton path enters the mode at activation (`_post_init_mtp`, before the first processor call) and leaves it on every exit through `_drop_mtp_state` (reconcile, park, late-join handoff, plain drop), with `pending` set iff a sampled `_next_tokens` is waiting for the standard step. `_mtp_common_eligible` no longer excludes grammar rows; the opt-in row-wise batch path still does (its per-row cache extract/merge is not wired for the mode), and `_ineligibility_reason` says so.
- `omlx/speculative/processing_sampler.py`: the wrapper calls `begin_speculative(len(prompt))` on processors that expose it and `end_speculative` in `reset_processors`, so `vlm_mtp` admits grammar. mlx-vlm's drafts stay greedy and unmasked on that path; output is grammar-valid because every emitted token is a target-side sample from masked logits, and acceptance under a grammar there is a number to measure, not a given.
- `omlx/model_settings.py`: `guided_grammar_enabled` leaves `vlm_mtp_processor_conflicts`.
- `omlx/scheduler.py`: comments only; `_omlx_advance_grammar_rows` skips speculative rows by construction (they are never pending).

## Tests

`tests/test_grammar_speculative.py` runs on the real xgrammar matcher with a 256-token mock vocabulary (one token per printable ASCII character, `</s>` as the stop token), no model weights:

- an emulation of the native walk (activation with the prompt-length history and the first token in flight, the seed call, the draft chain with snapshot/restore, then cycles of k+1 rows with per-row checkpoints, restore to the last emitted row and a buffer trimmed to the committed prefix) over random drafts that are right, legal-but-wrong or illegal, for k = 1..4 and six seeds: at every committed position the processor's bitmask equals a fresh matcher's advanced linearly along the emitted tokens, and the emitted sequence completes the grammar. A mutation control (a processor that does not advance) fails the same property;
- dead and terminated prefixes pass logits through and are revived by a restore or a shorter history; `restore_state` refuses to move forward;
- the hand-back leaves the matcher on the streamed history with the row pending, and the deferred `accept_token` continues from there;
- `MTPProcessingSampler` with a grammar processor samples only legal tokens across a draft-rejection rewind, and `reset_processors` returns a pristine deferred processor;
- eligibility (singleton on, row-wise off), the activation hook running before any forward, the real `_run_verify_cycle_chain` with a grammar row (drafts right / legal-but-wrong / illegal, m = 1) through to `_drop_mtp_state`.

`tests/test_mlx_lm_mtp_patch.py` and `tests/test_model_settings.py` are updated for the new contract. `ruff check` reports no new findings in the touched files.

## What this does not claim

No speed number. The gain depends on how often the head's drafts survive the mask, and on the draft side the per-position mask costs one small host sync per draft (the same cost `ThinkingBudgetProcessor` already pays in the chain). The measurement I will run when the box is free: a 611-request structured-output corpus (classification and extraction schemas with `maxLength` strings, temperature 0.1) at C = 1 on Qwen3.8-27B-oQ6e-mtp, `mtp_enabled` on versus off, with the gates: grammar armed on every request, every output accepted by a fresh `GrammarMatcher` on its schema, parse and truncation rates not worse than the baseline, acceptance from the `MTP[uid]` line, and decode tok/s from the response timings. Results go here before the draft flag comes off.

Refs #2918.

### DIFF
--- a/omlx/api/grammar.py
+++ b/omlx/api/grammar.py
@@ -13,19 +13,34 @@ bitmask is permissive during reasoning and constrained during output.
 This keeps the processor simple and enables uniform batched bitmask
 computation (parallel model forward || bitmask fill).
 
-The processor supports two usage modes:
+The processor has two ways of learning which token was sampled from its
+bitmask:
 
-1. **Per-request** (original): call ``processor(tokens, logits)`` directly.
-   Handles accept + bitmask fill + mask application in one call.
+1. **Deferred** (default, the BatchGenerator path): ``__call__`` only fills
+   and applies the bitmask and marks the row :attr:`pending`; the scheduler's
+   monkey-patched ``GenerationBatch._step`` calls :meth:`accept_token` with
+   the sampled id at the top of the following step, where the id has to be
+   evaluated anyway.  Advancing the matcher from ``tokens`` here would force a
+   host sync per step, which is what constrained decoding's throughput was
+   lost to before the accept moved (#2561).
 
-2. **Batched**: call ``processor.advance(tokens)`` to accept the previous
-   token, then use ``BatchGrammarMatcher.batch_fill_next_token_bitmask``
-   with the exposed ``matcher`` property to fill bitmasks in parallel
-   across the batch, and apply the combined bitmask externally.
+2. **Speculative** (the MTP verify walks, entered via
+   :meth:`begin_speculative`): the caller hands ``__call__`` the concrete
+   history the logits are conditioned on (the committed prefix plus the draft
+   tokens before this position) and the processor advances the matcher to
+   that history itself, accepting new tokens and rolling back when the
+   history is shorter than the last one it saw.  :meth:`snapshot_state` /
+   :meth:`restore_state` checkpoint the matcher position so a rejected draft
+   suffix can be undone with ``GrammarMatcher.rollback``.  The row is never
+   marked pending in this mode; :meth:`end_speculative` hands the request
+   back to the deferred protocol with a consistent matcher.
+
+A third, unused batched mode (``advance`` + ``BatchGrammarMatcher``) is kept
+for compatibility.
 """
 
 import logging
-from typing import List, Optional
+from typing import Any
 
 import mlx.core as mx
 import numpy as np
@@ -39,6 +54,7 @@ def create_grammar_compiler(tokenizer, model):
     Returns None if vocab_size cannot be determined.
     """
     from .._torch_stub import install as _install_torch_stub
+
     _install_torch_stub()
     import xgrammar as xgr
 
@@ -66,11 +82,15 @@ class GrammarConstraintProcessor:
 
     def __init__(self, compiled_grammar, vocab_size: int):
         from .._torch_stub import install as _install_torch_stub
+
         _install_torch_stub()
         import xgrammar as xgr
         from xgrammar.kernels.apply_token_bitmask_mlx import apply_token_bitmask_mlx
 
-        self._matcher = xgr.GrammarMatcher(compiled_grammar)
+        # ``max_rollback_tokens`` is unlimited by default on xgrammar >= 0.2.3
+        # (the argument is deprecated); passing -1 keeps older matchers able
+        # to rewind a full speculative window too.
+        self._matcher = xgr.GrammarMatcher(compiled_grammar, max_rollback_tokens=-1)
         self._vocab_size = vocab_size
         self._apply_mask = apply_token_bitmask_mlx
 
@@ -80,6 +100,19 @@ class GrammarConstraintProcessor:
         self._first_call = True
         self._pending = False
 
+        # Speculative-mode bookkeeping (see begin_speculative). History
+        # lengths count prompt + generated tokens, the way the callers'
+        # token buffers do.
+        self._speculative = False
+        self._base = 0  # history length when speculative mode began
+        self._seen = 0  # history length the matcher currently reflects
+        self._matcher_rel = 0  # tokens accepted into the matcher since _base
+        # History index past which the matcher stopped advancing: one past
+        # the terminating token, or the index of a token the grammar
+        # rejected (``_dead``). None while the matcher is live.
+        self._stop: int | None = None
+        self._dead = False
+
     # ------------------------------------------------------------------
     # Per-request mode (original interface)
     # ------------------------------------------------------------------
@@ -87,16 +120,20 @@ class GrammarConstraintProcessor:
     def __call__(self, tokens, logits: mx.array) -> mx.array:
         """Fill bitmask and apply to logits.
 
-        Accept is handled by the monkey-patched GenerationBatch._step(),
-        which is the only place that can see which token was sampled from
-        the result.  This method only fills the bitmask, applies it, and
-        records that a token is now in flight for this row (see
-        :attr:`pending`).
+        In deferred mode accept is handled by the monkey-patched
+        ``GenerationBatch._step()``, which is the only place that can see
+        which token was sampled from the result; this method only fills the
+        bitmask, applies it, and records that a token is now in flight for
+        this row (see :attr:`pending`).  In speculative mode the matcher is
+        first advanced (or rolled back) to ``tokens``.
         """
-        if self._terminated:
+        if self._speculative:
+            self._sync_to_history(tokens)
+        if self._terminated or self._dead:
             return logits
 
-        self._pending = True
+        if not self._speculative:
+            self._pending = True
         self._bitmask.fill(-1)
         self._matcher.fill_next_token_bitmask(self._bitmask)
 
@@ -104,7 +141,7 @@ class GrammarConstraintProcessor:
         return self._apply_mask(mx_bitmask, logits, self._vocab_size)
 
     def accept_token(self, token_id: int) -> None:
-        """Accept a generated token to advance matcher state."""
+        """Accept a generated token to advance matcher state (deferred mode)."""
         self._pending = False
         if self._terminated:
             return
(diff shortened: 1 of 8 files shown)


builtbycnob:
Measured, on the setup described in the PR text. oMLX 0.6.4 release wheel (native kernels in place) in a fresh venv, with the five Python files of this branch overlaid from a cherry-pick onto the `v0.6.4` tag (applies clean, hunks byte-identical). Model `qwen3.8-27b-oQ6e-mtp`, served on demand next to four pinned models (169 GiB pinned, 192.5 GiB with the 27B loaded, `--no-cache`). Two arms on the same server binary, MTP off (no `mtp_enabled` key) and MTP on (`mtp_enabled: true`), each replaying the same 611-request structured-output corpus at C = 1: 385 classification requests (schema with a `maxLength: 200` string, `max_tokens` 300) and 226 extraction requests (schemas with `["string","null"]` unions and length bounds, `max_tokens` 800), temperature 0.1, `enable_thinking: false`, one `user` message, `response_format` json_schema on every call. Same seed, same order.

Controls first. The `Warning` header, which this server sends when and only when the grammar failed to compile, is absent on 611/611 responses in both arms. The MTP-off arm logged zero `MTP[` lines; the MTP-on arm logged one `MTP[...] finish=` line per request, each with a non-zero drafted count, 611/611 (the fallback-to-standard-step line is DEBUG-level, so I could not count it at INFO). Before the constrained corpus I ran three unconstrained prompts (no `response_format`) at two output lengths, 64 and 300 tokens, twice each, and took the slope between the two points so prefill drops out: 28.1 tok/s with MTP off against 43.8 with MTP on, 1.56x on decode. So the head and the settings work on this build before any grammar is involved.

Grammar validity. The pre-registered check is a fresh `GrammarMatcher` on the request's compiled schema accepting the output and reporting completion, for every `finish_reason = stop` output. One thing got in the way: the non-stream chat handler re-serializes every parseable `response_format` output (`server.py:4002-4006`, `parse_json_output` then `cleaned_text = json.dumps(parsed_json)`), so the text a client receives is `json.dumps` of the model's JSON with ASCII escapes and canonical separators, not the model's bytes; a string with `maxLength: 200` compiles to `[^"\\\r\n]{0,200}` with no escape branch, so a matcher rejects the six bytes `\u00e8` where the model wrote `è`. On the corpus the audit therefore checks the decoded values against the schema as the grammar encodes it, by trying the returned text and then `json.dumps(obj, ensure_ascii=False)`, with two mutants per output (a trailing character, the first key renamed) as the positive control. On the streaming path without tools the deltas are the raw bytes, so I also replayed the first 50 schedule entries with `stream: true` per arm and ran the matcher on the concatenated deltas as-is.

- Corpus, MTP on: 0/601 off-grammar outputs (601 = the 611 minus 10 truncated ones); 1202/1202 mutants rejected. MTP off, same audit: 0/602 and 1204/1204.
- Streaming sample, raw bytes, MTP on: 0/50 off-grammar; 100/100 mutants rejected. MTP off: 0/48 and 96/96.

Parse and truncation. Lax-parsed and jsonschema-valid: 601/611 = 98.36% with MTP on against 602/611 = 98.53% with MTP off (Wilson 95% intervals [97.01%, 99.11%] and [97.22%, 99.22%]). Strict-parse failures: 8/611 vs 9/611; the control-character case (lax ok, strict fail, the xgrammar#800 symptom, visible here only when the server-side parse fails) is 0 with MTP on and 1 with MTP off. `finish_reason = length`: 10 with MTP on vs 9 with MTP off, all extraction in both arms.

Speed. The pre-registered estimator was completion tokens over (wall − TTFT) per request; the non-stream driver has no TTFT, so the corpus number is completion tokens over wall latency, a lower bound of the decode ratio because prefill is in the denominator of both arms. Paired per request, median of the MTP-on/MTP-off ratio with a 1,000-resample bootstrap:

- pooled: 1.283 [1.275, 1.293], n = 608 pairs
- classification: 1.272 [1.251, 1.277], n = 382; extraction: 1.323 [1.308, 1.340], n = 226
- pairs with equal output length (±5%): 1.279 [1.272, 1.289], n = 508; completion-token ratio median 1.000 (83 → 83 tokens)

The streaming samples carry the server's `generation_duration`, which excludes prefill: decode median 25.6 tok/s with MTP off, 47.9 with MTP on, paired ratio 1.870 [1.803, 1.924] on 50 pairs. Acceptance from the `MTP[` lines, MTP on: median 0.85 pooled over 611 requests (0.83 classification, 0.90 extraction), 0.86 over the first 100, 2.58 tokens per cycle median. Under the grammar the drafts survive more often than on prose (0.60 to 0.80 per prompt, 0.58 to 0.82 per call, on the three unconstrained probe prompts): with fixed keys and bounded values the mask leaves the head few legal continuations, so its drafts land more often, and the per-step bitmask is then paid once per cycle of 2.58 tokens rather than once per token.

Memory and errors: no model eviction, no hard-pressure line, no HTTP error, no `no_response`, no traceback in either arm; loaded-model count 5 throughout. The WARNING-level `GrammarMatcher rejected token` of the deferred path never appeared; the speculative walk's own rejection line is DEBUG-level, so I have no count for it at INFO. Both arms saw the adaptive prefill throttle pause a request at C = 1 (5 events on 3 requests with MTP off, 4 on 3 with MTP on), the pool sitting at 188-194 GB against a 200.45 GB sizing target; that is the pool's regime with five models resident, not something this branch changes.

Deviations from what I pre-registered: the MTP-off arm was restarted once to store the response text in the records (the first 53 records are archived, not used); two live probes during the MTP-off arm ran a second request for a few seconds, and the three records overlapping them leave the speed pairing; the MTP-on arm was launched twice because the first launch died with its shell at 32 records (archived, its log lines outside the analysed slice); the repeat baseline I had planned as a drift control (off / on / off) did not run, so drift between the two arms is bounded only by the equal-length pairs and the constant controls; and the consumer's own end-to-end smoke (200 documents through its pipeline) did not run in this window.

What I read from this: every condition I had written down beforehand for the branch itself is met. Zero off-grammar outputs on the 601 non-truncated corpus outputs plus the 50 streamed ones, with the positive control rejecting every mutant; valid-JSON and truncation rates within the margins I had set (not more than 1 point below and not more than 3 truncations above the baseline: 601/611 against 602/611, 10 against 9); MTP engaged on every request; and a speed ratio whose lower bound is 1.283 [1.275, 1.293] on 608 pairs against the 1.15 I had written down, with the interval above 1 before prefill is even taken out of the denominator. The whole 611-request corpus took 3,836 s with MTP on against 4,931 s without. I am taking the PR out of draft on that basis; the consumer-level smoke stays on my list and I will add it here when it has run, together with anything it turns up. The analysis script, the report and the run notes are in my repo; the report JSON is small and I can attach it here if that helps review.

builtbycnob:
Follow-up: the same validation on Qwen3.8-Flash-Next (`qwen4_exp`), 2026-09-08.

Setup: oMLX 0.6.4 release wheel with this PR's five files overlaid; a Flash-Next checkpoint made of a 6-bit gs64 core with a 4-bit N-gram table, plus the Lightning MTP head quantized separately from the bf16 source in the layout of `mlx-community/Qwen3.8-Flash-Next-oQ6e-mtp` (32 head tensors: 22 quantized, 18 at 6-bit gs64 and the four shared-expert ones at 8-bit; 10 left bf16). One model resident, `--no-cache`, C = 1, temperature 0.1, the same 611-request structured corpus as the 27B run (385 classification + 226 extraction, `response_format` json_schema on every call, bounded strings), MTP off against MTP on.

Results, MTP on vs off:

- Off-grammar outputs: 0/611, audited the two ways as before: a fresh `GrammarMatcher` per output, 250 of the 611 accepted as returned and 361 only after re-serialising with `ensure_ascii=False`, since the non-stream handler escapes non-ASCII that a bounded string's grammar has no branch for; 1,222/1,222 mutants rejected. Raw streamed bytes: 0/50.
- Parse + schema-valid: 611/611 vs 609/611. Truncations (`finish_reason=length`): 0 vs 2.
- Decode: completion tokens / wall latency, paired per request, median 1.238 [1.228, 1.248] on 611 pairs (a lower bound, prefill included). From the server's `generation_duration` on 50 streamed requests: 32.0 → 48.7 tok/s, 1.53× [1.49, 1.60].
- Acceptance: median 0.843 over the 611 requests (0.82 classification, 0.90 extraction), 2.6 tokens per cycle, 611/611 drafted.
- Unconstrained decode at temperature 0.1 (three prompts of 37-46 tokens, each at 64 and 300 output tokens twice, read as the slope so prefill drops out): 33.4 → 41.1 tok/s, 1.23×, under the 1.30× line I had set from the 27B, where the same probe gave 1.56×. I had written that line down beforehand as a stop rule for the corpus, so the corpus above is a deviation I am declaring: the line measures prose and was calibrated on the 27B's own trained head, the regime in question is the constrained one, and the shipped build misses the same line on the same probes.
- Zero evictions, memory-pressure or throttle lines; 611/611 HTTP 200.

Two things worth knowing here. Enabling MTP changes the core's decode numerics on this family: with the head attached the decode path drops two hyper-connection optimisations at once, the exact hybrid projection and the compiled single-token path, and I did not isolate which of the two moves the bits. A perplexity probe cannot see either — it runs prefill-shaped, those flags are read only at decode shape (1, 1) — so identical pooled perplexity is a self-consistency control, not evidence of equality. Greedy 128-token continuations do see it: 6 of 12 prompts identical offline, 3 of 12 through the server, diverging at token 9-91. Document by document, the two arms' classification labels agree on 373 of 383 (97.4 %); the 89 documents that recur in the schedule got the same label on every repeat with MTP off (736 of 736 pairs) and on 713 of 736 with MTP on — at temperature 0.1 with no same-arm replicate, so how much of that is sampling and how much the numerics is not attributable from this run. What the corpus supports is that no gate moved against the MTP-on arm, at the same median completion length (83 → 83 tokens). The head also pays less here than on the 27B in both regimes (1.23× vs 1.56× unconstrained, 1.53× vs 1.87× on JSON), which is a calibration question rather than a correctness one.

Positive control, the shipped `mlx-community/Qwen3.8-Flash-Next-oQ6e-mtp` (its own core and head) on the same server and probes: MTP off 33.4 tok/s, on 43.3 = 1.297×, missing the same 1.30× line by 0.1 tok/s, and per prompt the two trade places (1.30 / 1.20 / 1.63 against my 1.24 / 1.23 / 1.67) — so the smaller-than-27B gain is the family's, not this head's. Both heads come from the same bf16 source: the 10 bf16 tensors are byte-identical, all 22 quantized ones differ in bytes, 19 of them inside two-rounding noise (Pearson 0.9995 or better on 18, 0.99947 on the last). The exception is the three hyper-connection `input_mix_weight_up` tensors, where the bytes show clipping at the extremes on the shipped side: across the whole tensor on the top-level mixer, its largest values roughly halved (Pearson 0.948), on a small fraction of the rows on the other two (Pearson 0.997 and 0.998). Acceptance sweeps at temperature 0 / 0.3 / 0.7 / 1.0 on 400-token prose, code and JSON prompts, 12 calls per arm: mine 0.802 / 0.810 / 0.815 / 0.794 pooled against 0.799 / 0.800 / 0.813 / 0.786. Each head ran on its own core, so that compares two builds rather than two heads under one core, and a gap of at most 1 pp against a 95 % interval of about ±4-5 pp says indistinguishable on this instrument, not equal.

Last, the consumer check on the 27B (`qwen3.8-27b-oQ6e-mtp`, MTP on) that the earlier comment left open: a document pipeline's own classification and extraction over its 100-document corpus, through the server. Extraction 100/100 documents final (95 first pass, 2 on retry, 3 through the pipeline's penalty path; with MTP off the day before, same build and corpus: 97 / 2 / 1). Classification 98/100 against 99/100 with MTP off — one document truncated at the 300-token ceiling, one carrying two control characters inside string values (strict parse fails, lenient succeeds). That second document failed the same way with MTP off, same signature and different bytes, in that run of the day before and in two campaigns before it: the known xgrammar#800 class, where the bounded-string rule itself admits the character, so MTP can neither create it nor cure it, though like any change to the decode numerics it can move which document trips it.

builtbycnob:
Merged `main` at `65c65e38` (after #3695 multi-request Lightning MTP and the two #3702 follow-ups) into the branch as `edb48c0d`; the PR merges clean again. The tree equals a plain rebase of the original commit (checked by interdiff), kept as a merge commit so the reviewed history stays.

What the conflict resolution changed, and nothing else:

- `_mtp_common_eligible`: the grammar exclusion is removed from the common gate as before, and the XTC gate from #3084 is kept in its place.
- `_is_mtp_batch_eligible` / `_ineligibility_reason`: grammar rows stay off the *multi-request* path (the old row-wise env is gone upstream; the check now sits after `_model_supports_batch_mtp`). The speculative grammar walk is still wired for the singleton cycle only, so a batch with a constrained row decodes on the standard step, and a constrained singleton runs under MTP.
- Two tests adapted to `main`: the multi-request test marks the fake model `_omlx_mtp_multi_request` instead of setting the removed env, and the verify-cycle fake batch carries the `max_tokens` / `_num_tokens` / `state_machines` / `_matcher_states` fields `_run_verify_cycle_chain` reads since #3695.

One observation from the merge, since it bears on the motivation: `_mtp_common_eligible` returns False whenever *any* row of the batch carries a `GrammarConstraintProcessor` (`_has_grammar_processors` scans every row). Before #3695 that cost nothing in a multi-row batch, which did not run MTP anyway; now the same gate sits in front of the multi-request path, so one `response_format` request joining a batch turns MTP off for every request in that batch, not only for itself. This PR keeps the constrained row itself under MTP when it is alone and, when it is not, leaves the other rows to the (still MTP-less) standard step; extending the batch path to constrained rows is a separate change.

Tests, on the merged tree with the omlx `main` dependencies: the full `tests/` run with `-m "not slow and not integration"` (the CI selection, `test_grammar_live.py` deselected as it needs a server) — 13,862 passed, 331 skipped, 0 failed. The measurements in the two earlier comments are from before the rebase and were not repeated.

Issue #3129, opened 2026-08-25
VLM thinking mode: model falls into repetition loop; entire reasoning buffer flushed into a single content delta when generation ends inside thinking (Qwen3.8-27B-nvfp4 + MTP drafter)


dalianblue:
## Environment

- oMLX 0.6.3rc3 (Homebrew tap), macOS 26.5, Apple M1 Pro / 32GB
- Model: `mlx-community/Qwen3.8-27B-nvfp4` (VLM, served via VLM engine) + `mlx-community/Qwen3.8-27B-MTP-nvfp4` drafter attached through per-model settings (`vlm_mtp_enabled: true`, `vlm_mtp_draft_model: ...`)
- Server: `omlx serve --model-dir ~/.omlx/models --paged-ssd-cache-dir ~/.omlx/cache --port 8000`, MTP acceptance ~98.8% (247/250, block_size 3) — speculative decoding itself works great

## Symptom 1: repetition loop in thinking mode (stochastic)

With thinking enabled (default chat template) and `temperature: 0.2`, some Chinese medical/knowledge questions send the model into an infinite self-questioning loop inside the reasoning channel until `max_tokens`. Example prompt:

```
告诉我2025年最新版《中国2型糖尿病防治指南》中，糖化血红蛋白（HbA1c）的控制目标具体修改为多少？
```

Observed: the phrase `让我仔细回忆：2024年1月，CDS发布了...` repeated **212×** over 622s until `max_tokens` (8192). The same prompt re-run later completed cleanly in 173s, so it is stochastic. A second prompt (brain-death criteria, ~700-word structured answer) looped on 3/3 runs.

Notably, adding `frequency_penalty: 0.3 + presence_penalty: 0.1` made things *worse* on some prompts: a TCA-cycle question that answered correctly without penalties went into a loop **2/2 runs** with penalties (874s / 871s). I later noticed `model_settings.py` documents that vlm_mtp is exclusive with processor-backed penalty settings — but the OpenAI-compatible API silently accepts `frequency_penalty`/`presence_penalty` instead of rejecting them, which makes this unsupported combination easy to hit by accident.

## Symptom 2: entire reasoning buffer flushed into a single `content` delta (deterministic given symptom 1)

Whenever generation terminates while still inside the thinking section, the full accumulated reasoning text is emitted as **one giant `content` delta**, duplicating the whole thinking into the message content.

Delta-level analysis of one run (M1 + penalties, 871s):

```
[...]

Issue #2860, opened 2026-08-19
reasoning_content separation is inconsistent: finish_reason=length leaks reasoning fragments into content


anicaise-ai:
**Environment**: omlx 0.6.0 → 0.6.2, DeepSeek-V4-Flash-0731-MXFP4 (thinking model), M3 Ultra, macOS 26.6.2.

Since 0.6.0 the chat completions API separates reasoning from the final answer, and it works beautifully on complete responses:

- **`finish_reason: "stop"`** → reasoning lands in `message.reasoning_content`, `message.content` is the clean final answer. (This fixed a whole class of "reasoning gibberish reaching the user" bugs for us — thank you.)
- **`finish_reason: "length"`** → fragments of the reasoning stream are still present in `message.content`.

So a client that trusts `content` to be user-displayable is safe only when generation wasn't truncated — exactly the case where it's *least* obvious something went wrong. We hit this in production summarization: a `max_tokens` slightly too small produced raw chain-of-thought in user-facing output, while the same prompt with more budget was clean.

**Expected**: the stop/length distinction shouldn't change *where* text goes — on truncation, whatever partial reasoning exists should still be routed to `reasoning_content` (and `content` left empty or partial-answer-only if the model had already closed its thinking block).

Repro sketch: any thinking-enabled model, a prompt that induces a long reasoning phase, `max_tokens` smaller than the reasoning length, `temperature: 0`. Compare `message.content` between a truncated and a non-truncated run.

Workaround we use meanwhile: give clients generous `max_tokens` and treat `finish_reason=length` responses as non-displayable.

aikidovation:
Also seeing length-truncation spill CoT into `content` with empty `reasoning_content` on Qwen3.6-35B-A3B-4bit / 0.6.4 — curl repro in #3796 (adjacent to #2457).

PR #3509, opened 2026-09-08
fix(stream): keep unfinished reasoning private

jonathan308:
## Summary

This is the unfinished-reasoning output-handling split requested in #3327.

- Never reclassify an unterminated private reasoning block as visible answer content.
- Report unfinished OpenAI Chat and Anthropic streams as incomplete/length-capped.
- Emit an incomplete Responses terminal event instead of a completed response.
- Reject structured or raw tool-call candidates that occur inside unfinished private reasoning across Chat, Anthropic, and Responses.
- Preserve normal tag-free text, completed reasoning, valid tools, and existing sampling behavior.

## Scope

This PR intentionally contains no EOS masking and no MTP processor-contract changes. It does not alter logits or prevent a model-selected stop; it handles the resulting incomplete protocol safely.

It is stacked on #3327 because the Qwen streamed-tool suppression uses the capability-scoped streaming seam retained there. After #3327 merges, I will restack this branch onto `main` so the diff collapses to the single four-file output-handling commit.

## Validation

Python 3.13 focused suite: **485 passed, 12 deselected** across unfinished-thinking streams, thinking parsing, Qwen tool calling, and end-to-end streaming.


eazerox2:
This appears to be the same failure class as this PR. I am adding a production-shaped reproduction from Hermes using Qwen3.8-Flash-Next.

## Environment

- Hardware: Apple Silicon Mac Studio M3 Ultra, 96 GB unified memory
- OS: macOS 26.6.2
- oMLX: 0.7.0.dev1
- API: OpenAI-compatible `POST /v1/chat/completions`
- Client: Hermes Agent
- Streaming: enabled
- Thinking: enabled
- Request reasoning setting: `reasoning_effort=medium`

## Observed production behavior

Two Hermes responses produced the following persisted message state:
text
message IDs: 206474, 206493
finish_reason: stop
structured tool_calls: absent
reasoning_content: non-empty
content: non-empty
content.strip() == reasoning_content.strip()
The visible content was byte-equivalent to the reasoning content. Hermes therefore persisted the model's private reasoning as ordinary assistant content and streamed it to Discord. The turn then ended without a useful final answer.

The affected requests were long, tool-heavy conversations. The approximate request/output sizes were:
text
request 1: ~61.8K prompt tokens, ~602 completion tokens
request 2: ~75.8K prompt tokens, ~580 completion tokens
configured output cap: 16,384 tokens
Both responses ended with `finish_reason="stop"` and were far below the output cap. This does not appear to be ordinary `max_tokens` truncation.

## Minimal parser evidence

The installed parser reproduces the classification problem directly:
python
parser = ThinkingParser(start_in_thinking=True)
parser.feed("private reasoning with no closing boundary")
parser.finish()
Observed result:
text
('', 'private reasoning with no closing boundary')
The accumulated unfinished thinking is returned in the visible-content position. The active Qwen chat template can open generation inside
 (1/2)
``.

## Expected behavior

When a prompt-opened private-thinking block reaches EOS without ``:

- unfinished text must not be promoted to visible `content`;
- it should remain private reasoning or be discarded from the user-visible response;
- the response should be marked incomplete/malformed rather than treated as a successful final answer;
- tool-call candidates inside unfinished private reasoning should not be executed or promoted;
- streaming and non-streaming paths should apply the same policy.

Returning an incomplete response is preferable to exposing private reasoning. I am not requesting automatic EOS masking here; this is specifically about fail-closed output classification after the model has already stopped.

## Related reports

This also appears related to:

- #2457 — unfinished thinking re-emitted as visible content on `finish_reason="length"`;
- #2860 — reasoning/content separation leaks on truncation;
- #2649 — prompt-opened thinking with redundant close markers exposes private text;
- #2744/#2745 — prompt-opened thinking and EOS recovery produce no visible content until final recovery;
- #3129 — Qwen3.8 reasoning/repetition behavior involving reasoning being flushed into visible content.

The distinguishing detail in this report is that the downstream response was externally reported as `finish_reason="stop"` while `content` duplicated `reasoning_content`.

No private prompt text, reasoning text, credentials, or session transcript is included here. I can provide a sanitized raw SSE trace or additional parser-level reproduction if useful.

PR #3511, opened 2026-09-08
fix(mtp): fail closed on processor contract violations

jonathan308:
## Summary

This is the positioned-MTP processor-contract split requested in #3327.

`MTPProcessingSampler` previously logged a warning and silently fell back to the base sampler when positions, row counts, checkpoints, or contiguity violated the positioned-processing contract. That fallback bypassed request logits processors for the affected token and the remainder of the request.

The sampler now raises a dedicated `MTPProcessorContractError` immediately and remains failed closed on subsequent calls. Requests with no processors retain the existing direct base-sampler path.

## Scope

- No EOS masking.
- No thinking-output reclassification.
- No changes to valid positioned MTP sampling, snapshots, rewind, or accepted-token accounting.
- Only malformed/incompatible processor contracts change from unprocessed sampling to an explicit error.

## Validation

Python 3.13: **78 passed** across positioned MTP processing, VLM MTP, proxy routing, rollback, and fallback coverage.

### DIFF
--- a/omlx/speculative/processing_sampler.py
+++ b/omlx/speculative/processing_sampler.py
@@ -61,6 +61,10 @@ import mlx.core as mx
 logger = logging.getLogger(__name__)
 
 
+class MTPProcessorContractError(RuntimeError):
+    """Positioned MTP sampling cannot safely enforce request processors."""
+
+
 def supports_vlm_mtp_processing(processor: Any) -> bool:
     """True when a logits processor can be applied on the vlm_mtp path."""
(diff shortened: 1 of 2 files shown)


jundot (approved):
Thanks for splitting this out. I verified that valid positioned sampling and thinking-budget rewinds are unchanged, and malformed calls now stop instead of bypassing processors. I also prepared a maintainer follow-up that isolates the error to the affected request and releases its MTP/cache state; the engine-loop checks cover unrelated active and waiting requests plus a subsequent MTP request. LGTM with that follow-up; I'll land both together.

PR #3512, opened 2026-09-08
fix(vlm): do not concatenate per-image vision cache entries with mismatched shapes

### LINKED ISSUE #3508: [0.6.4] VLM vision feature cache: mx.concatenate of per-image entries with mismatched token counts → 500 (gemma-4)
**Version:** 0.6.4 · **Model:** gemma-4-26B-A4B-it — reproduced on both `gemma-4-26B-A4B-it-oQ4e` and `gemma-4-26B-A4B-it-heretic-oQ4`, same shapes · **Machine:** M4 Max, 36 GB

## Summary

The VLM vision feature cache keys per-image entries by the hash of one image
alone. But the soft-token count depends on the image's aspect ratio.

So two entries cached from two separate **single-image** requests can have
different sequence lengths. Combining them on a later multi-image request
throws:

```
POST /v1/chat/completions → 500 (unhandled): [concatenate] All the input array
dimensions must match exactly except for the concatenation axis. However, the
provided shapes are (1,256,2816), (1,250,2816), and the concatenation axis is 0.
```

```
  File "omlx/engine/vlm.py", line 4090, in _process_chat_messages
    ) = self._prepare_vision_inputs(
  File "omlx/engine/vlm.py", line 3175, in _prepare_vision_inputs
    combined = mx.concatenate(cached_per_image, axis=0)
ValueError: [concatenate] All the input array dimensions must match exactly ...
```

`2816` is `text_config.hidden_size` for gemma-4-26B-A4B.

## Cause

`_prepare_vision_inputs`. Line numbers here are 0.6.4's; the same code is at
~3197 on `main` (`9fb488c`), unchanged:

```python
per_hashes = compute_per_image_hashes(images)
cached_per_image = [self._vision_cache.get(h, self._model_name) for h in per_hashes]
...
if all(f is not None for f in cached_per_image):
    combined = mx.concatenate(cached_per_image, axis=0)          # <- throws
    if self._vision_features_match_image_tokens(combined, image_token_count):
        ...
```

`_vision_features_match_image_tokens` is exactly the check that would catch
this. But it runs *after* the concatenate, so it never gets the chance. The
validation sits downstream of the operation that can't survive invalid input.

The mismatched entries are minted by `_split_vision_features`:

```python
if num_images <= 1:
    return [features]      # stores this image at its SOLO token count
```

A multi-image request stores a whole-request entry instead (the split returns
None for gemma-4 with per-image resize), so only solo requests can produce
per-image entries — and those are cached at a count that isn't valid alongside
a differently-shaped image.

## Reproduction (deterministic, 3/3)

Order matters. `A+B` → `A` → `A+B` does **not** reproduce, because the first
multi-image request stores a whole-request entry rather than per-image ones.

```python
import base64, io, json, urllib.request, uuid
from PIL import Image, ImageDraw

MODEL = "gemma-4-26B-A4B-it-heretic-oQ4"
URL = "http://127.0.0.1:1234/v1/chat/completions"
run = uuid.uuid4().hex[:6]          # fresh bytes so the cache starts clean

def img(w, h, tag):
    im = Image.new("RGB", (w, h), (30, 90, 160))
    d = ImageDraw.Draw(im)
    d.rectangle([w//4, h//4, 3*w//4, 3*h//4], fill=(255, 255, 255))
    d.text((10, 10), tag + run, fill=(0, 0, 0))
    buf = io.BytesIO(); im.save(buf, "PNG")
    return "data:image/png;base64," + base64.b64encode(buf.getvalue()).decode()

def turn(*urls):
    return {"role": "user", "content": [{"type": "text", "text": "Describe."}] +
            [{"type": "image_url", "image_url": {"url": u}} for u in urls]}

def post(label, *urls):
    body = {"model": MODEL, "messages": [turn(*urls)], "max_tokens": 8,
            "stream": False, "chat_template_kwargs": {"enable_thinking": False}}
    req = urllib.request.Request(URL, data=json.dumps(body).encode(),
                                 headers={"Content-Type": "application/json"})
    try:
        with urllib.request.urlopen(req, timeout=900) as r:
            json.load(r); print("OK  ", label)
    except urllib.error.HTTPError as e:
        print("500 ", label, e.read().decode()[:120])

A, B = img(1024, 1024, "A"), img(1536, 640, "B")
post("A alone  -> cached (1, 256, 2816)", A)
post("B alone  -> cached (1, 250, 2816)", B)
post("A + B together", A, B)          # 500
```

## Trigger conditions

Both are required:

1. **Every** image in the request has a per-image entry cached from a solo
   request. If even one lacks it, `all(...)` fails, `cached_whole` takes over
   and the request succeeds. A conversation whose images are only ever sent
   together therefore stays healthy indefinitely.
2. At least two of them have different soft-token counts.

The second is a function of aspect ratio and is orientation-independent.
Measured pairwise, each image cached solo first:

| pair | result |
| --- | --- |
| 1024×1024 + 512×512 (both 1:1) | OK — every square is 256 tokens |
| 1344×768 + 768×1344 (transposes) | OK — same ratio, either orientation |
| 1024×1024 + 1536×640 | 500 |
| 1024×1024 + 1344×768 | 500 |
| 1536×640 + 1344×768 | 500 |

Reaching this needs a client that shows the model an image **alone** and later
sends the same bytes alongside another one. The obvious multi-image patterns are
immune: several images uploaded together arrive in one request and get a single
whole-request entry, and images added one per turn never appear alone after the
first, since the earlier ones are replayed beside them.

I hit it from a client that shows a freshly generated image on its own in the
turn that produced it, then replays it next to later images. I could not recover
the exact sequence from the affected session, so treat that as the mechanism
rather than a reconstruction. The guard in #3512 does not depend on it — it
rejects mismatched entries however they were cached.

## Recovery needs both tiers

Worth documenting regardless of the fix — once stuck, the same request fails
every time, and neither half of the obvious remedy works alone:

| action | result |
| --- | --- |
| `rm -rf ~/.omlx/cache/vision_features`, server left running | still 500 — the in-memory LRU still holds the entries |
| unload the model (drops the LRU), SSD dir intact | still 500 — the SSD tier repopulates it |
| both (unload first, then remove — `close()` drains a background write queue) | fixed |

## Suggested fix

Sent as #3512, with tests.

Check the shapes agree before combining, and let the existing fallback handle
the mismatch — noting the `cached_whole` fetch has to move onto the same
condition. It is currently guarded by `if not all(f is not None ...)`, so
tightening only the use-site drops into `elif cached_whole is not None` with
`cached_whole` still `None`, forcing a full recompute and skipping a
whole-request entry that may exist:

```python
per_image_usable = (
    all(f is not None for f in cached_per_image)
    and len({f.shape[1:] for f in cached_per_image}) == 1
)

cached_whole = None
if not per_image_usable:
    cached_whole = self._vision_cache.get(image_hash, self._model_name)

used_cached_features = False
if per_image_usable:
    combined = mx.concatenate(cached_per_image, axis=0)
```

A stricter alternative would be to include the resize regime (or the resulting
soft-token count) in the per-image cache key, so a solo-cached entry can't be
matched against a multi-image one at all. That changes the key format and
invalidates every existing SSD entry — measured cost of the resulting cold
misses is ~445 ms per distinct image, once (930 ms vs 485 ms warm, M4 Max,
gemma-4-26B-A4B, 1024x1024). Happy to follow up with that shape instead if it
is preferred; #3512 is a useful guard underneath it either way, since the
shapes can disagree for reasons the key does not model.

djtech42:
## Summary

Closes #3508.

`_prepare_vision_inputs` combines cached per-image vision features with
`mx.concatenate(cached_per_image, axis=0)`. If two entries were cached under
different resize regimes, their shapes disagree and the request 500s.

This checks the shapes before combining and falls through to the existing
`cached_whole` path when they differ.

## Root cause

Entries are keyed by the image alone. But the soft-token count depends on the
image's aspect ratio. Measured on gemma-4-26B-A4B, each cached from its own
single-image request:

| image | tokens |
|---|---|
| 1024x1024 | 256 |
| 1536x640 | 250 |

So two images cached from separate single-image requests can never be
concatenated. `mx.concatenate` raises before
`_vision_features_match_image_tokens` can reject them.

```
POST /v1/chat/completions -> 500 (unhandled): [concatenate] All the input array
dimensions must match exactly except for the concatenation axis. However, the
provided shapes are (1,256,2816), (1,250,2816), and the concatenation axis is 0.
```

Repro (3/3, fresh image bytes each run): image A alone, image B alone, then A+B.
Order matters. `A+B` first stores a whole-request entry and does not reproduce.
Script in #3508.

Recovery today needs both cache tiers. Deleting
`~/.omlx/cache/vision_features` alone leaves the in-memory LRU holding the
entries; unloading the model alone lets the SSD tier repopulate them.

## The change

`per_image_usable` now gates both the combine and the `cached_whole` fetch.

Both, because the fetch sits behind `if not all(f is not None ...)`. Gating only
the combine would reach `elif cached_whole is not None` with `cached_whole`
still `None`, forcing a recompute and skipping an entry that may exist.

## Test plan

- [x] mismatched shapes fall through to the recompute, and fetch the
      whole-request entry on the way
- [x] matching shapes still combine and serve from cache, so the fix cannot
      degrade into "never use the per-image cache"
- [x] mutation check: reverting the source change fails the first test with the
      production `ValueError`, second still passes
- [x] `tests/test_vlm_engine.py` + `tests/test_vision_feature_cache.py`: 147 passed

## Scope and trade-off

You may prefer to fix the key rather than the consumer. A per-image entry
describes an image *in a request context*, but `compute_per_image_hashes`
hashes only `WxH` and the raw RGB bytes. Adding the resize regime
(`num_soft_tokens_per_image` is already in `extra_model_inputs` at lookup time)
would let both variants coexist and actually hit cache for mixed-aspect sets.

I did not send that because it invalidates every existing SSD entry. Measured
cost on an M4 Max, gemma-4-26B-A4B, 1024x1024: a cold miss is ~930 ms against
~485 ms warm. About 445 ms per image, once.

That is a call about your invalidation story, not mine. This guard is worth
having underneath it either way, since shapes can disagree for reasons the key
does not model. Happy to follow up with it.

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -3198,15 +3198,30 @@ class VLMBatchedEngine(BaseEngine):
                     self._vision_cache.get(h, self._model_name) for h in per_hashes
                 ]
 
+                # Per-image entries are keyed by the image alone, but the
+                # number of soft tokens an image encodes to depends on the
+                # resize regime, which depends on the *other* images in the
+                # request (Gemma 4 per-image resize: 1024x1024 -> 256 tokens,
+                # 1536x640 -> 250). Entries cached from separate single-image
+                # requests can therefore disagree, and mx.concatenate raises
+                # before _vision_features_match_image_tokens below ever gets to
+                # reject them. Check the shapes agree first and fall through to
+                # the whole-request entry (and then a recompute) when they do
+                # not.
+                per_image_usable = (
+                    all(f is not None for f in cached_per_image)
+                    and len({f.shape[1:] for f in cached_per_image}) == 1
+                )
+
                 cached_whole = None
-                if not all(f is not None for f in cached_per_image):
+                if not per_image_usable:
                     # Fallback: whole-request entry (stored when per-image split
                     # is unsupported, e.g. Gemma 4 multi-image with per-image
                     # resize). Mirrors the store-side branch below.
                     cached_whole = self._vision_cache.get(image_hash, self._model_name)
 
                 used_cached_features = False
-                if all(f is not None for f in cached_per_image):
+                if per_image_usable:
                     # All images cached individually — combine and use
                     combined = mx.concatenate(cached_per_image, axis=0)
                     if self._vision_features_match_image_tokens(
--- a/tests/test_vlm_engine.py
+++ b/tests/test_vlm_engine.py
@@ -1469,6 +1469,102 @@ class TestPrepareVisionInputs:
         assert call_kwargs.get("audio") is None
 
 
+    # --- per-image vision feature cache -------------------------------
+
+    def _vision_cache_engine(self, images, entries_by_index):
+        """Engine whose vision cache already holds the given per-image
+        features, keyed by each image's real per-image hash."""
+        from omlx.utils.image import compute_per_image_hashes
+
+        engine = self._setup_engine_for_vision(model_type="gemma4")
+        hashes = compute_per_image_hashes(images)
+        entries = {hashes[i]: f for i, f in entries_by_index.items()}
+
+        cache = MagicMock()
+        cache.get.side_effect = lambda h, _model: entries.get(h)
+        engine._vision_cache = cache
+        engine._vision_cache_enabled = True
+
+        # Keep the embedding step inert but mx.eval-able.
+        embed = MagicMock()
+        embed.inputs_embeds = mx.zeros((1, 3, 8))
+        embed.to_dict.return_value = {"inputs_embeds": embed.inputs_embeds}
+        engine._vlm_model.get_input_embeddings.return_value = embed
+
+        # The token-count cross-check is a separate guard; neutralise it so
+        # these tests isolate the shape agreement.
+        engine._image_token_count = MagicMock(return_value=None)
+        return engine, cache
+
+    @staticmethod
+    def _two_images():
+        from PIL import Image
+
+        return [Image.new("RGB", (8, 8), "red"), Image.new("RGB", (12, 5), "blue")]
+
+    @pytest.mark.skipif(not HAS_MLX, reason="MLX not available")
+    @patch("mlx_vlm.utils.prepare_inputs")
+    @patch("mlx_vlm.prompt_utils.apply_chat_template")
+    def test_per_image_cache_used_when_shapes_agree(self, mock_act, mock_prepare):
+        """Two entries of equal shape still combine and serve from cache."""
+        images = self._two_images()
+        engine, _ = self._vision_cache_engine(
+            images, {0: mx.zeros((1, 4, 8)), 1: mx.ones((1, 4, 8))}
+        )
+        mock_act.return_value = [{"role": "user", "content": "formatted"}]
+        mock_prepare.return_value = {
+            "input_ids": mx.array([[1, 2, 3]]),
+            "pixel_values": mx.zeros((2, 3, 8, 8)),
+        }
+        engine._compute_vision_features = MagicMock(return_value=None)
+
+        engine._prepare_vision_inputs([{"role": "user", "content": "Describe"}], images)
+
+        used = engine._vlm_model.get_input_embeddings.call_args[1][
+            "cached_image_features"
+        ]
+        assert used.shape == (2, 4, 8)
+        engine._compute_vision_features.assert_not_called()
+
+    @pytest.mark.skipif(not HAS_MLX, reason="MLX not available")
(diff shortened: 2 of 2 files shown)


jundot (approved):
Thanks for the fix. I reproduced the concatenate failure and verified that mismatched entries use the whole-request cache when available, while compatible per-image entries still hit both cache tiers. LGTM, I'll merge this.

PR #3514, opened 2026-09-08
fix: separate partial prefill costs from representative memory estimates

jundot:
Partial prefill chunks can include fixed allocations that inflate their per-token cost. Applying that rate to larger chunks can cause unnecessary throttling or rejection. Cheap partials can also lower the estimate below the cost of a full chunk.

This change keeps partial observations out of the representative EWMA and latest representative sample once a full chunk has been measured. Partial allocation evidence remains a separate guard term, capped at the observed byte cost when considering larger chunks. Cheap reuse does not discard that evidence.

| Area | Before | After |
|---|---|---|
| Partial observations | Context-priority partials can raise or lower representative estimates | Once a full chunk is measured, partials no longer update the representative EWMA or latest sample |
| Partial allocation protection | A partial's per-token rate can amplify its cost for larger chunks | Separate protection capped at observed bytes for larger candidates; existing downward scaling retained |
| First short request | Partial observations seed the initial estimate | Conservative bootstrap retained; the first full observation replaces the bootstrap rate |
| Reclaim accounting | Releases accumulate; any positive sample clears the entire charge | Absolute footprint tracking avoids duplicate charges and preserves unpaid releases |
| GLM initialization | Persistent fused weights are allocated during the first forward | Weights are prepared during model loading, outside chunk measurement |
| Qwen4 short prefill | Gathered pricing can disagree with the actual dense route | Estimator and runtime share the minimum gathered-query threshold |
| Restored prefixes | Some estimates omit the prefix when boundary snapshots are disabled | External and chunked prefill both include the restored length |

Validation:
- A 512-token/333 MiB sample followed by repeated 128-token/1,319 MiB partials predicts about 1.67 GiB for the next full chunk instead of 6.70 GiB, including the safety factor.
- Regression tests cover cheap partials, gradual increases below the outlier threshold, repeated large allocations, guard enforcement, and preserved floor/reclaim accounting.
- Qwen3.6 server runs completed normal, short, partial, and 16.9K-token requests before and after the change. Cache-hit differences prevent a direct performance comparison.
- GLM projection preparation preserves small-model prefill/decode outputs for floating-point and quantized weights. The observed first-chunk footprint growth fell from 3.44 GiB to 0.73 GiB. This moves persistent allocation into loading rather than reducing the model's eventual footprint.
- The focused scheduler, preflight, tracker, memory-profile, VLM, and context-benchmark suites pass after applying the change to current remote main.

Generic models retain conservative bootstrap behavior until a full chunk is measured. Pending partial costs can still restrict requests when the observed allocation does not fit. This does not claim complete coverage of #3417, #3362, or #3450; GLM 100K/159K and the reported SSD expert-streaming workload have not been validated on this revision.

Related: #3417, #3362, #3450, #3350

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -13,3 +13,15 @@ Run `python -m pytest -q tests/test_prefill_transient_tracker.py tests/test_pref
 # Prefix cache completion tests
 
 Run `python -m pytest -q tests/test_scheduler.py tests/test_scheduler_boundary_completion.py tests/test_prefix_cache_gdn_split.py` to check cache-freshness admission and completed boundary recovery. The completion tests use a small initialized Qwen3.5 hybrid model and the real BatchGenerator, then compare restored-prefix logits with a fresh forward pass. They cover embedded snapshots, GDN sidecars, off-boundary completion, and unknown or inconsistent cache positions.
+
+# Prefill reclaim and restored-context regressions
+
+The prefill accounting suites also check repeated versus contiguous footprint releases, partial repayment on skipped chunks, and growth between chunks. A one-off spike followed by a normal sample must retain main's EWMA recovery; repeated large allocations must remain visible to the scheduler. These are controlled observations, not actual OOM reproductions.
+
+Run `python -m pytest -q tests/test_scheduler.py tests/test_qwen4_qsa_reservation_integration.py` to check that restored context reaches chunk pricing even when boundary snapshots are disabled, for both external and chunked prefill.
+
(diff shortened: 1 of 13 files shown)


lyonsno:
Found a guard regression in the generic sizing path: `partial_bound()` is nonlinear, but `_adaptive_chunk_size` and `_guard_prefill_chunk` still shrink proportionally to `prediction / n`.

Using the PR's 512-token/333 MiB then 128-token/1,319 MiB observations, with a 32-token floor and 1,000 MiB of headroom, adaptive sizing returns 288 and the guard returns 160. The predicted cost of that 160-token chunk is still 1,714.7 MiB. The parent returns 64 at 857.35 MiB.

Could both paths size against the actual nonlinear bound and check the returned chunk fits? Reproduced with unchanged predictor/guard methods and controlled footprint readings, not a live OOM.

PR #3516, opened 2026-09-08
feat(cluster): ship the Cluster v2 dashboard

jonathan308:
## What changes

Adds the guided Cluster v2 dashboard for discovery, code pairing, model selection, live memory budgets, signed Tensor/Pipeline planning, model staging, activation, and saved-deployment management. It replaces the legacy Cluster tab and retains the advanced CUDA enrollment, fabric verification, and diagnostic tools.

The backend dependencies through #3258 are merged. This PR also includes the serving integration fixes and the backend contracts required by the dashboard; it is not limited to template changes.

## Pairing wiring

The joiner routes are now implemented on the existing admin-authenticated pairing router:

| Route | Behavior |
|---|---|
| `POST /api/cluster/pair/join` | Creates the local code and sends the existing identity-bound verifier to the coordinator's `/pair/request`. An existing pending join returns 409. |
| `GET /api/cluster/pair/join` | Polls the coordinator's `/pair/status/{node_id}` and calls the existing `complete_join` verification/enrollment path on approval. Returns pending, approved, denied, expired/error, or idle state. Code-bearing responses use `Cache-Control: no-store`. |
| `POST /api/cluster/pair/join/cancel` | Retires the local attempt and code. A delayed response from that attempt cannot complete a replacement join. |

The coordinator's approve/deny handlers remain the trust decision. Polls are serialized, while network waits allow cancellation to invalidate the attempt. Approval completion retains the existing key-unwrapping, coordinator-identity verification, SSH enrollment, and persistence rollback checks.

## UI/backend integration

- Reconcile runtime `ownership` against the engine pool, accept the worker's actual loading stages, and retain saved setup controls after unload/failure. Runtime residency remains separately labeled; saved setups can be selected, loaded/retried, changed, or removed.
- Carry `prompt_cache_ssd` and `prompt_cache_ssd_max_bytes` through autoconfigure, activation, replan, persisted execution settings, worker arguments, and the SSD snapshot store's byte limit. Reject unknown autoconfigure fields instead of silently dropping them.
- Separate `ready_to_stage` from `ready_to_activate`: preflight/fabric/source blockers prevent starting, while missing destination files permit the copy-first flow. Preserve deployment ID and path mappings, including those mappings in the placement signature used by staging and activation.
- Fetch live `/node-budgets` before initial planning and publish role `reserve_fraction`. Only paired participants enter the version-mismatch gate.
- Guard staging with one in-flight poll, attempt generations, and a one-time activation claim. Dismissed/old responses cannot activate or clear a newer job. Remove duplicate Alpine initialization and clean up polling timers.
- Publish and validate bounded `active_request_metrics` with request IDs, per-request counters/rates, and an explicit truncation count so concurrent requests render independently.

## Before

```mermaid
flowchart LR
  UI[Guided dashboard] --> Missing[Missing join routes / mismatched runtime fields]
  Plan[Plan with unsupported settings] --> Copy[Unbound overlapping staging polls]
  Copy --> Activate[Duplicate or stale activation]
```

## After

```mermaid
flowchart LR
  UI[Guided dashboard] --> Join[Admin join session and verified approval completion]
  Budget[Live node budgets] --> Plan[Signed plan with settings, ID and paths]
  Plan --> Gate[Check proposal blockers]
  Gate --> Copy[Stage missing files for this attempt]
  Copy --> Activate[Claim activation once]
  Activate --> Runtime[Engine ownership and bounded request metrics]
  Runtime --> Saved[Retained setup controls]
```

## Validation

- 27 new integration/regression cases cover real pairing managers behind HTTP routes, cancellation of delayed approval, expired-join retry state, loading-stage validation, runtime ownership, per-request metrics, measured initial budgets, path/signature consistency, SSD worker-argument round trips, copy-first expansion, and stale/duplicate polls.
- Broader cluster/distributed/dashboard regression run after merging current `main`: **1,402 passed**.
- Final pairing/UI run after refining terminal snapshots: **156 passed**. Additional locale/dashboard checks: 6 passed.
- JavaScript syntax checks and `git diff --check` pass.

These are local tests with mocked peer/process boundaries; this change does not claim a new physical multi-Mac validation run. Phase-split/disaggregated serving, hybrid TP×Pipeline, and performance-qualified unequal TP layouts remain outside this dashboard's supported controls.


jundot:
@jonathan308, thanks for putting this together. I checked the new UI against the backend and also reviewed #3155. It includes the joiner-side pairing methods, but the `/api/cluster/pair/join` and `/api/cluster/pair/join/cancel` routes used by this UI are missing, so those calls return 404. I considered handling this in a follow-up, but #3155 has no route implementation I can restore to complete the flow. Could you check the intended wiring, including approval completion, status polling, and cancellation?

I also found the following issues that I think need changes:

- **Runtime state:** The UI relies on `ownership`, which `/api/cluster/runtime` does not provide. Some worker loading stages are also rejected by the runtime validator, so running or loading deployments can disappear from the UI.
- **Saved deployments:** Unloaded or failed setups lose their management controls, making reload, retry, and removal inaccessible.
- **Execution settings:** Replan requests return 422 because they include unsupported SSD cache fields. Initial planning silently drops those fields, so the cache toggle and size limit do not take effect.
- **Planning and expansion:** Activation does not enforce proposal blockers, while adding a node that needs model staging is blocked before copying can start. Expansion also needs to preserve the deployment ID and path mappings.
- **Memory budgets:** Initial planning uses physical RAM instead of live admission budgets, and the role response omits the reserve fraction used by the UI.
- **Polling:** Overlapping staging responses can activate twice, and a late response from a dismissed job can activate a newer job before its copy finishes. The component also initializes polling twice.
- **Version checks:** An unpaired discovered device with a different version can block setup.
- **Request metrics:** The UI expects `active_request_metrics`, but the backend does not publish it, so concurrent requests are not displayed individually.

Could you review these and share your thoughts? I'm open to handling some follow-up fixes, but I'd like to clarify the missing integration pieces first.

jonathan308:
Fixed. My cluster testing is on hold until this weekend while I wait for more hardware to arrive. Testing has been on hold for a few weeks sorry these issues didn’t surface earlier. I'll be able to test a 3 and 4 device cluster here soon .

jundot:
@jonathan308, thanks for the quick turnaround. The missing API and runtime contracts are now wired up. I'll add coverage for pairing retries after cancellation, IPv6 addresses, model selection with multiple saved setups, expansion replanning after a 409, and out-of-order runtime responses, and make the corresponding adjustments in a follow-up. I'll also align the SSD description with the synchronous writes.

PR #3520, opened 2026-09-08
perf(qwen4_exp): long-context decode and MTP verify rows on the gathered QSA arms

Rascal:
### TLDR
Increasing the token generation speeds for large contexts.

### Summary

Above 32k tokens of context, Qwen3.8-Flash-Next backbone decode rows and Lightning-MTP verify rows attend through the gathered QSA arms instead of reading the whole KV cache densely, so decode speed stays nearly flat with context. On an M5 Max (128 GB) against current `main` (b908f563), serial decode is +7.5% at 63k, +18% at 134k and +28% at 229k tokens; with the adaptive MTP controller it is +8.5%, +25% and +36% under production sampling and +15%, +29% and +32% under greedy decoding. Below the threshold both branches take the same path and measure the same.

Eight commits on top of `main` b908f563:

1. **`_gather_kv_rows`: gather QSA rows on the cache's own layout instead of copying it.** Both gathered arms fetched their selected K/V rows by transposing the whole cache to token-major and reshaping it, a full copy per call whose cost grows with context. The new helper takes the selected rows along the token axis of the stored `(B, H, N, D)` cache (per batch element above batch one) and returns them in the same shape and order. Per QSA layer per token on the M5 Max:

   | cached tokens | transpose + reshape (before) | `mx.take` along the token axis (after) | dense masked SDPA |
   |---|---|---|---|
   | 41k | 0.65 ms | 0.32 ms | 0.43 ms |
   | 82k | 0.91 ms | 0.25 ms | 0.42 ms |
   | 206k | 1.83 ms | 0.27 ms | 0.66 ms |

   This copy is why the gathered arms looked slower than dense attention on M5 (30b085c7 measured them 1.4–3× slower and gated the prefill arm to ≥ 16 query rows). Bit-identical output; a test checks the helper against the old code for batch 1 and 2 and that its cost no longer grows with context.

   The stored-layout take is the right form for one to a few query rows, but the gathered *prefill* arm calls the same helper with 64 query rows per sub-chunk, and there the old copy is cheaper up to about 100k tokens: the take reads the selected rows once per query row, while the copy pays for the cache once and then gathers flat. Per QSA layer per 64-row sub-chunk (2,051 selected rows), M5 Max:

   | cached tokens | token-major copy + flat gather | take along the token axis |
   |---|---|---|
   | 16k | 0.58 ms | 1.55 ms |
   | 65k | 0.90 ms | 1.02 ms |
   | 131k | 1.19 ms | 1.04 ms |
   | 206k | 1.49 ms | 1.05 ms |

   The helper now dispatches on both: 32 or more query rows against a cache under 128k tokens take the copy, everything else the take (`_gather_kv_rows_token_major` / `_gather_kv_rows_stored`). Decode and verify rows are unaffected. Tests check both forms bit-identical against the old code on rank-two and rank-three index shapes and pin the dispatch at 1, 4, 16 and 64 rows across 4k to 206k tokens.

1b. **Native sparse-GQA tests: reference built with the new helper.** Four tests in `tests/test_qwen4_qsa_sparse_gqa.py` gathered their FP32 reference through the removed helper; they only run where the extension is built, so CI did not see them. Fixed to use `_gather_kv_rows`.

1c. **Decode-width indexer scores stay on MLX ops on NAX GPUs.** With the `glm_moe_dsa` extension built, the gathered decode arm reached the native `qwen4_qsa_indexer_scores` kernel, which on the M5 Max is slower than the portable expression for the few rows a decode or verify step carries, while it wins clearly from 64 rows up where prefill runs it. Per QSA layer at 134k tokens:

   | query rows | native kernel | MLX ops |
   |---|---|---|
   | 1 | 0.77 ms | 0.27 ms |
   | 4 | 0.36 ms | 0.26 ms |
   | 16 | 0.36 ms | 0.30 ms |
   | 32 | 0.39 ms | 0.39 ms |
   | 64 | 0.40 ms | 0.54 ms |
   | 2048 | 5.1 ms | 11.8 ms |

   The native kernel now engages from 32 query rows on NAX GPUs (`OMLX_QWEN4_QSA_NATIVE_SCORE_MIN_ROWS` overrides; other GPUs keep it everywhere). Serial decode on the gathered arm with the kernel off for those rows: +4% at 63k, +8% at 134k, +11% at 229k; prefill unchanged.

1d. **Same for the top-k (from 8 rows) and the sparse GQA attention kernel (from 24 rows).** The native top-k ties or loses to `argpartition` at one to four rows and wins from eight; with both selection kernels off for decode rows the serial arm reached +15% at 229k instead of +11%. The native sparse attention is the largest of the three at few rows: the whole gathered arm at 134k tokens, per layer:

   | query rows | native attention | gathered SDPA |
   |---|---|---|
   | 4 | 1.56 ms | 0.70 ms |
   | 8 | 1.58 ms | 0.87 ms |
   | 16 | 1.65 ms | 1.36 ms |
   | 24 | 1.67 ms | 1.83 ms |
   | 64 | 1.74 ms | 3.93 ms |

   Four to eight rows are the MTP verify steps, so the kernel was costing about 10 ms per verify cycle across the 12 QSA layers. `OMLX_QWEN4_QSA_NATIVE_TOPK_MIN_ROWS` and `OMLX_QWEN4_QSA_NATIVE_MAIN_MIN_ROWS` override the thresholds. All three gates apply on NAX GPUs only; elsewhere the native kernels keep engaging at every width, as before.

2. **Keep long-context text rows on the gathered QSA arms.** Two pieces:

   - *Server side.* The VLM adapter proved text-only (rank-two) positions only under a one-shot proof armed per text prefill chunk; decode and MTP-verify steps rebind positions through the batch rope-delta path and arrive at the language model as `(3, 1, T)` mRoPE planes, so the decode arm's rank-two predicate never held in the live server and backbone rows always took the dense path. The scheduler now marks a request as text-only at insert (`mark_text_positions`), the step-time binder (`set_step_rope_deltas`) turns that into a step-scoped proof for single-row batches, and the adapter emits rank-two positions for those steps above the context threshold. No per-step host sync: the proof is bookkeeping set at prefill time. Multimodal requests keep the rank-three planes and the official path. The MTP proxy forwards the two seams explicitly.
   - *Model side.* Verify rows (`target_verify`, batch one, causal, index-aligned, above the indexer budget) route through `_gathered_text_prefill(..., target_verify=True)`, which threads the verify-width Q/K/V, indexer and output projections. The draft head already took the gathered decode arm because it passes no positions; with the backbone on the same path, head and target agree more at long context and per-draft acceptance rises 4–10 points at 82k (code 69→73%, prose 64→74%).

3. **Gate the gathered verify arm on rank-two positions.** The verify arm first accepted broadcast-identical `(3, 1, T)` planes through `_broadcast_text_mrope_position_ids`, so MTP verify rows went gathered from the 2k indexer budget on, below the step threshold, and paid two host syncs per positions array for the plane comparison: −14% / −8% / −5% adaptive tok/s at 4k / 8k / 16k under greedy decoding, and enough changed accepted tokens to make the depth controller park. Requiring absent or rank-two positions (which the adapter emits only above its threshold) restores +3% / +2% / +1% at those sizes; both arms now follow one gate.

**Threshold.** `OMLX_QWEN4_STEP_TEXT_POSITIONS_MIN_CONTEXT` defaults to 32768. Measured on the M5 Max with the arms forced on versus off at fixed context: serial crosses over near 12k (4k −2.0%, 8k −0.3%, 16k +0.3%, 31k +2.3%), adaptive MTP under production sampling near 16k, but adaptive MTP under greedy decoding (deeper drafts, more verify rows) only near 36–40k (32k −11.1%, 40k +3.1%, 48k +1.6%, 63k +14.1%). 32k is the smallest value that loses nothing in any regime; users who only run sampled decoding can lower it to 16384.

### Switches

| variable | default | effect |
|---|---|---|
| `OMLX_QWEN4_STEP_TEXT_POSITIONS` | `1` | `0` keeps decode/verify steps on the rank-three positions (arms never engage for backbone rows) |
| `OMLX_QWEN4_STEP_TEXT_POSITIONS_MIN_CONTEXT` | `32768` | context (tokens) from which the step proof engages |
| `OMLX_QWEN4_QSA_GATHERED_VERIFY` | `1` | `0` keeps verify rows on the dense path even when decode rows gather |

### Measured in the server

M5 Max 128 GB, High Power, Qwen3.8-Flash-Next-oQ4e-mtp, one interleaved chain per mode (B A A B, fresh server per arm, two passes per prompt), code prompts of 1,950 … 109,600 words, 512-token completions. Mean of four; `main` b908f563 → this branch (9e19561b), both with the native kernel extensions built (an earlier chain without the sparse-attention extension measured the fallback path and overstated the branch by up to 20 points at 229k; those numbers are withdrawn).

#### Serial decode (MTP off, greedy):

| context | main | this branch | |
|---|---|---|---|
| 3.9k | 62.7 | 62.5 | −0.4% |
| 7.7k | 61.3 | 61.2 | −0.1% |
| 16k | 59.4 | 59.5 | +0.1% |
| 32k | 57.5 | 57.3 | −0.3% |
| 63k | 53.7 | 57.7 | +7.5% |
| 134k | 47.3 | 55.7 | +17.7% |
| 229k | 40.9 | 52.4 | +28.1% |
p2_serial

#### Adaptive Lightning MTP (max depth 3) with "real life" sampling (temperature 1, top-p 0.95, top-k 20, thinking on):

| context | main | this branch | |
|---|---|---|---|
| 3.9k | 62.1 | 64.8 | +4.4% |
| 7.7k | 62.7 | 62.6 | −0.1% |
| 16k | 63.6 | 64.5 | +1.5% |
| 32k | 62.1 | 58.7 | −5.4% (arms 56.6 and 61.2; the greedy run below shows +1.2% at the same size) |
| 63k | 53.2 | 57.7 | +8.5% |
| 134k | 45.3 | 56.8 | +25.4% |
| 229k | 39.2 | 53.4 | +36.1% |
p2_mtp_temp_1

#### Adaptive Lightning MTP (max depth 3) under greedy decoding (temperature 0, thinking on):

| context | main | this branch | |
|---|---|---|---|
| 3.9k | 70.2 | 69.5 | −0.9% |
| 7.7k | 74.3 | 76.3 | +2.6% |
| 16k | 67.8 | 66.1 | −2.5% |
| 32k | 63.8 | 64.6 | +1.2% |
| 63k | 58.0 | 66.4 | +14.6% |
| 134k | 48.6 | 62.6 | +28.7% |
| 229k | 44.1 | 58.5 | +32.4% |
p2_mtp_temp_0

Greedy runs are deterministic, so their per-cell spread is about 2% (the 3.9k row carries one 62 tok/s sample among 70–75 on each side); per-draft acceptance is 76.9% on `main` and 78.6% on this branch, 2.04 and 2.17 tokens per cycle. Greedy MTP drafts deeper than sampled MTP, so more verify rows per cycle and a larger gain from keeping them off the dense path: `main` parked its controller 4–5 times per arm, this branch 0–3. What the branch changes is the cost of a speculative cycle at long context. From the greedy arms' own MTP stats, backbone time per cycle and tokens per cycle, `main` → this branch: 16k 26.9 → 28.3 ms at 1.97 → 2.04, 63k 29.5 → 31.0 ms at 1.93 → 2.39, 134k 32.9 → 27.1 ms at 1.79 → 1.87, 229k 44.6 → 37.2 ms at 2.36 → 3.13. Per emitted token that is 13.7 → 13.9 ms at 16k, 15.3 → 13.0 at 63k, 18.4 → 14.5 at 134k and 18.9 → 11.9 at 229k: the branch's cycles cost no more where they carry the same rows, and at long context they carry more rows for less time.

Sampled runs generate different text each pass, so their spread is about 8% per cell (serial about 3%); rows at and below 32k run the same code on both sides and sit inside it (the 32k cell's two branch arms read 56.6 and 61.2). Per-draft acceptance is 73.1% on `main` and 74.5% here, 1.91 and 1.99 tokens per cycle. The adaptive controller parked MTP in 14 of 15 requests per arm on `main` and 3–7 on this branch: with the verify rows off the dense path and the few-row kernels on MLX ops, the speculative cycle stays cheap enough at long context that the controller keeps drafting.

### Tests

`tests/test_qwen4_qsa_gather_rows.py` (both gather forms bit-identical to the old path, dispatch by query count and context, flat cost for few rows), `tests/test_qwen4_qsa_verify_gather.py` (verify-row eligibility and routing), `tests/test_qwen4_qsa_native_indexer.py` (the three native kernels yield to MLX ops below 32/8/24 rows on NAX GPUs only, env overrides), `tests/test_qwen4_qsa_sparse_gqa.py` (native-only reference repaired), `tests/test_qwen4_text_positions_binders.py` (scheduler marking and step binder), additions to `tests/test_vlm_model_adapter.py` (rank-two steps only with proof, kill switch, minimum context, unregister clears) and `tests/test_scheduler_chunked_prefill.py` (prefill records the proof). All Qwen4, adapter and scheduler tests pass on the rebased tree (248).

### Related work

- #3320 (draft) routes Lightning-MTP verify windows through the gathered arm above 65,536 tokens with a one-time in-model proof that the three mRoPE planes agree, plus a native BM8 indexer tile and pooled-index retention across rollback. It keeps the copying gather (so its arm pays the cost in the first table) and does not cover plain decode rows, which is where most of the serial gain here comes from; its 65k gate was measured against that copy. The gather fix in this PR applies to #3320's arm unchanged.
- The `OMLX_QWEN4_GATHERED_MIN_QUERY` gate from 30b085c7 (default 16 rows) was calibrated against the copying gather. It no longer reaches MTP windows on this branch: verify rows take `_gathered_text_verify_eligible`, which has no row minimum, and in the MTP profiles every 2 to 4-row backbone forward is a verify call. Re-measured on native kernels with the few-row gates above, the gathered arm beats the dense causal SDPA at every width from 2 rows up (per QSA layer, gathered / dense: 4 rows 0.64 / 3.66 ms at 65k, 0.67 / 7.19 ms at 134k, 0.74 / 12.2 ms at 229k; 2 rows 0.55 / 1.11, 0.55 / 1.74, 0.64 / 2.83 ms; 16 rows 1.26 / 2.77, 1.34 / 5.66, 1.48 / 11.5 ms). The gate now only decides sub-16-row non-verify calls (tail prefill chunks, history passes), which the profiles do not show at all, so the default is left as is; lowering it to 2 would be safe.


pierre427:
Nice result, and I think the root-cause framing here is exactly right: on M5 the gathered arms looked slower than dense not because of a bandwidth roofline but because each call materialized a full token-major copy of the cache, and that copy grows with context. We landed on the same diagnosis from a different direction and wanted to share where it took us, including where our approach is worse than yours.

**What we built.** Rather than fixing the copy in the gather, we replaced the gather with an indexed split-K attention kernel: it reads the selected K/V rows *by block index* out of the stored `(B, H, N, D)` layout — 64-slot chunks as the unit of association — so there's no transpose and no per-call cache copy, same as your `mx.take`, but it also does the softmax reduction in-kernel with a fixed merge tree.

The expensive lesson was correctness. "Exact vs the gather baseline" turned out to mean *cloning the native `sdpa_vector_2pass` arithmetic* — 128 interleaved streams, pre-scaled q, the fast-exp path, bf16 partial boundaries, 32-simdgroup second pass — not writing a cleaner reduction of our own. Our first, "better" contract (contiguous chunks, precise exp, fp32 partials, sequential merge) was bit-identical on small shapes but diverged from gather by ~0.06 in the chosen-logprob at 16K, which is enough to flip a token on a real verify transaction. Only after we made the kernel reproduce the native kernel's arithmetic exactly did digests match.

**How it performs** (M5 Max, B=1, self-MTP k=2, the same Flash-Next / 4-bit-A3B profile):

- Isolated attention vs gather: ~2.8× at 16K rising to ~11× at 128K (M=3); M=1 crosses over between 32K and 64K (≈1.1× at 64K, ≈1.5× at 128K).
- End-to-end decode vs gather: **+11% at 16K, +21% at 32K, +42% at 64K**, digest-identical.

So the split-K reduction buys a lot at M>1 and very long context, on top of killing the copy.

**Why your `mx.take` version may be the better tool anyway.** The honest accounting of our approach:

1. **Build-pinned admission.** Because the kernel clones the native arithmetic exactly, it's married to a specific mlx build — we gate it behind an allowlist plus a hash of the SDPA kernel header. When the deployed interpreter's mlx build doesn't match, the kernel *silently declines and falls back to the gather*. That bit us: a serving venv on a slightly different mlx build ran the slow copy in production and nobody noticed until a receipts run caught the fallback reason. A one-line `mx.take` has none of this — it's the same code on every build.
2. **The fallback still copies.** Below the crossover, or whenever admission declines, we're back on the transpose+`take_along_axis` path — i.e. exactly the copy you removed. Your change would strictly improve *our* fallback.
3. **M>1 isn't free to turn on.** The M=3 path is exact but still gated, because widening the verify window changes the verify token stream (not just latency) and needs its own digest sign-off.
4. **Near-tie vs dense.** Our exactness contract holds against the gather; against dense-masked attention, digests differ at ~1/256 positions with ~0.2-logit margins — in-band, but not something we'd call bit-identical.

Net: a custom split-K kernel wins where the *reduction* is the cost (M>1, 64K+), but it carries a brittle build-hash gate and a large maintenance surface, and its fallback is the thing you just fixed. `mx.take` wins where the *copy* is the whole cost (M=1, short-to-mid context) with none of that fragility. The split that's been on our mind is a hybrid: `mx.take` row-gather as the default, and the custom kernel only in the band where split-K's reduction actually pays. Your PR makes the default half of that trivial.

jundot:
@pierre427 Thanks for sharing the details, especially the work needed to reproduce the native SDPA arithmetic. That distinction between matching gather and matching dense is useful.

I also reproduced a dense/gather argmax difference by replaying the same input against cloned KV caches. With a fixed MTP depth, repeated gathered runs produced identical 512-token outputs, while dense and gathered outputs differed. My checks with a 32K Code (Python) background showed no clear quality regression: MMLU was 877/1000 versus 880/1000, and HumanEval was 159/164 versus 158/164, for dense and gathered respectively. Those results support the change, though they don’t establish equivalence.

I’ve now retested the updated PR on an M3 Ultra. Its prefill dispatch recovered the earlier slowdown while preserving the long-context decode gain, and keeping the original gather for all prefill offered no clear end-to-end advantage. I’m comfortable keeping `mx.take` as the default.

Your split-K results look promising as a separate optimization, particularly for MTP verification. If you have a public implementation or repro harness, I’d be interested in testing it on M3 Ultra too.

jundot:
@Rascal Thanks for the work on this and the updated prefill dispatch. The long-context decode improvement is meaningful, and the latest changes recovered the prefill slowdown I measured earlier.

I tested `Qwen3.8-Flash-Next-oQ4e-mtp` on an M3 Ultra with 512 GiB, using identical oMLX Code (Python) contexts, temperature 0, adaptive Lightning MTP up to depth 3, and caching disabled.

Each cell below is **prefill / decode tok/s**. “My split” keeps the original gather for ordinary prefill and the new gather for decode/verification.

| Version | 16K | 128K | 200K |
|---|---:|---:|---:|
| Main `94530d8d` | 829.0 / 61.0 | 753.8 / 49.6 | 723.1 / 46.4 |
| Original PR `e6288443` | 767.4 / 61.1 | 712.1 / 59.3 | 695.9 / 50.2 |
| Original PR + my split | 831.0 / 62.1 | 754.4 / 59.2 | 720.4 / 50.0 |
| Updated PR `9e19561b` | 835.7 / 62.1 | 757.1 / 59.2 | 722.3 / 50.1 |
| Updated PR + my split | 826.8 / 60.0 | 753.6 / 58.9 | 723.2 / 50.2 |

These were single sequential runs with thermal variation. The generation limit was 512 tokens; the 16K case reached EOS after 122.

In repeated isolated QSA measurements, my split reduced prefill time by 8.9% at 128K and 1.0% at 200K, with identical outputs. However, the server prefills only 131,071 tokens of a 128K request, so your updated dispatch already keeps it on the original path. At 200K, my split saved only 0.32 seconds of TTFT. I don’t see a clear end-to-end reason to replace your updated dispatch.

I also compared dense (A) and gathered (B) attention with fixed MTP depth 3 and a shared 32K Code (Python) background:

| Evaluation | Dense A | Gathered B |
|---|---:|---:|
| MMLU, 1,000 questions | 877/1,000, 87.70% | 880/1,000, 88.00% |
| HumanEval, all 164 tasks | 159/164, 96.95% | 158/164, 96.34% |

Neither paired score difference was statistically significant. These extended-context checks showed no clear quality regression, though they predate the updated head and don’t establish equivalence.

Greedy outputs do differ between dense and gathered attention. With fixed depth, repeated gathered runs produced identical 512-token outputs, but dense and gathered diverged at token 225. Replaying the same input against cloned KV caches reproduced the argmax difference, so this isn’t solely an adaptive-MTP timing effect.

The results look reassuring, and I’m leaning toward merging. I’d like to run a few more controlled greedy comparisons on the updated head before doing so. Thanks again for the substantial long-context improvement and the follow-up work on prefill.

moc375:
Adding an M3 Ultra 512GB data point from when this was the p2 branch (head 7699f2e, same runs we posted over on #3469): single stream, greedy, ~70.9k ctx → 77.2/76.9 tok/s prose / 89.3/89.4 code, ~88.6k → 72.5/76.6, ~110.9k code → 64.2/64.8. Isolated gathered-vs-dense at 64k on our box came out +27-34%.

The decode slope staying flat past 32k is the part we care about most — our long agent sessions live in that range. Hoping this one lands.

moc375:
## Production validation on M3 Ultra 512GB — four fleet records, repro steps inside

**TLDR**: We've been running this branch's lineage (7699f2e + the merged #3469 content) on two production instances since tonight. All four context tiers set fleet records, and the full-machine-exclusive numbers below are what convinced us to ship it before merge.

### Environment

M3 Ultra 512GB · Qwen3.8-Flash-Next-oQ4e-mtp (125B MoE, 6B active) · greedy, streaming, single stream unless noted · env: `HC_FUSED=1`, `EAGER_DISPATCH=1`, `ROW_GATHER=1`, MTP depth adaptive. **The whole machine runs one engine instance** — see the field note below, it matters more than we expected.

### Single-stream decode ladder (tok/s, three rounds, 3 iters each)

| ctx | r1 | r2 | r3 | best |
|---|---|---|---|---|
| 1k | 155.9 | 149.5 | 162.6 | **162.8** |
| 16k | 79.8 | 74.1 | 86.0 | **86.1** |
| 32k | 80.2 | 76.4 | 80.6 | **89.8** |
| 64k | 78.3 | 73.9 | 89.5 | **89.5** |

Engine telemetry: tok/cycle 1.8–2.45, accept 50–84%, backbone ≈37 ms/cycle, zero parks, zero errors across ~500 requests in two nights.

### Field note: resident siblings halve these numbers

With two sibling engine instances merely *idle but resident* on the same box (their keepwarm ticks + resident weights), the identical requests measure 46–62 @16k instead of 74–86. We spent an evening chasing this as "window drift" before isolating it. For anyone benchmarking multi-instance hosts: stop the siblings, don't just quiet them.

### Concurrency (warm cache, plain batching — no forced MTP)

| shape | per-session | aggregate |
|---|---|---|
| c=2 × 4k | 43.3 / 42.8 (3 identical rounds) | 86 |
| c=4 × 4k | 15–40 | 88–96 |
| c=2 × 16k | 29.7 / 30.0 | 60 |
| c=4 × 16k | 15–27 | 76 |

### One honest data point for review

Our own #3265 batched-MTP path, measured against plain batching on this branch under controlled warm-cache A/B (4k/16k × c2/c4, all four combos): −14% to −27% aggregate. So on this hardware the best concurrent shape is native batching + your gathered arms; #3265's durable value is the crash-fix lineage, not throughput. The third commit's verify gating (rank-two) may matter less on M3 Ultra: the ungated base shows no measurable sub-16k cost here (1k = 155–162), unlike the −14% you saw on M5 Max.

### Reproduction

```
# one engine on the box, model lazy-loads on first request
# prompt: repeated prose sized ~5.1 chars/token, e.g. target 16384 tokens
curl -s localhost:PORT/v1/chat/completions -d '{
  "model": "Qwen3.8-Flash-Next-oQ4e-mtp",
  "messages": [{"role": "user", "content": "<prose filler>"}],
  "max_tokens": 300, "temperature": 0, "stream": true,
  "stream_options": {"include_usage": true}}' | ...
# decode tok/s = completion_tokens / (total - TTFT) from chunk timestamps;
# cross-check the engine's "MTP[..] tok/cycle / accept / backbone ms" lines
```
Methodology and full JSON: https://github.com/moc375/blog (qwen4-speculative-decoding).

### Acknowledgments

@Rascal — the root-cause framing on the transpose-copy cost, the invitation to test your p2 branch, and these two commits are now literally what our production fleet runs at night. The independent reader who reproduced the single-stream stack at 104.8 on a 256GB M5 box deserves a nod too. This is the best of upstream collaboration: you built the speed, we tried our best to break it and couldn't.

jundot:
@Rascal Thanks again for the long-context decode improvement and the follow-up work on prefill. Exact greedy equivalence with dense attention isn’t guaranteed, but my evaluations found no clear evidence of quality degradation, and the updated PR preserved the decode gains while recovering prefill performance.

One thing I’ll address in a follow-up: when a prefix-cache hit skips prefill entirely, the text-only marker isn’t set, so that request misses the gathered decode/verify path. I’ll preserve that eligibility for verified text-only cache hits and update the stale module description alongside it.

LGTM, merging.

Issue #93, opened 2026-03-06
Prefix cache stores prompt+output tokens, causing ~4.6x SSD bloat for reasoning models


rsnow:
### Problem

 The SSD prefix cache stores the full token sequence (prompt_token_ids + output_token_ids) as the cache key and KV payload. For reasoning models that generate <think> blocks
 (e.g. Minimax-m2.5, Qwen3), the thinking tokens are stripped by the API layer before subsequent requests — but the cache already stored KV for those tokens, creating dead
 blocks that never match again.

 ### Impact (observed on minimax-m2.5-mix46-g64)

 | Metric | Before fix | After fix |
 |--------|-----------|---------- -|
 | Cache size | 56 GB | 12 GB |
 | Blocks stored | 873 | 194 |
 | Tokens cached | 216K | 47K |
 | Bloat ratio | 4.6x | 1.4x (healthy prefix overhead) |

 Per-token KV cost on disk: 254 KB (62 layers × 8 KV heads × 128 head_dim × BF16 × K+V). A 47K-token conversation should need ~12 GB of KV cache — but the thinking output pushed
 it to 216K tokens cached and 56 GB on disk. Every turn added ~8K dead thinking tokens that would never be reused.

 ### Root Cause

 In _cleanup_finished(), the scheduler stored list(request.prompt_token_ids) + list(request.output_token_ids) as the token sequence for prefix cache. On the next request, the
 API layer sends only the prompt (prior turns with thinking stripped), so the cached blocks from prompt+output never prefix-match. New blocks get written, old ones become dead
 weight.

 ### Fix

 One-line change: store list(request.prompt_token_ids) only. The prefix cache's hash-based block matching already handles partial sequence reuse correctly — it just needs the
 right tokens to match against.

jundot:
The current behavior is actually intentional, even accepting some inefficiency. For non-reasoning models, output tokens can also be cached and reused as prefix hits on the next turn. I considered that benefit more important than the wasted SSD space from reasoning tokens. (the top priority for oMLX has been improving prefill efficiency, and the SSD KV cache uses an LRU policy so unused blocks get evicted eventually anyway.)

Your point made me reconsider though, and skipping output token storage only for reasoning models could be a good middle ground. Thanks for the thoughtful suggestion. I'll look into applying this selectively for reasoning models.

rsnow:
[...]

PR #3525, opened 2026-09-08
fix(cache): keep a reasoning request's output cacheable when the history preserves it

Rascal:
### TLDR

1. Reasoning fix: when the chat history keeps a model's thinking, its answer is now cached too, so the next turn re-prefills only what is new instead of the whole previous answer.
2. Tool-call fix: the same rule now covers turns that end in a tool call, so an agent loop's reasoning and call are cached for the next step.
3. Along the way, a pre-existing MTP bug got fixed: the block-boundary alignment was only armed by the first snapshot capture, so a prompt shorter than one block lost its first boundary and, with output caching on, its whole answer.

### Summary

A request whose prompt ends in an open `<think>` tag stores only its prompt tokens at completion (c8ee5eaf, issue #93): at the time the API stripped the reasoning before the next turn, so cached output could never prefix-match and only cost SSD space and thrashing. Two later features changed the premise. `preserve_thinking` (0749e3fd, #814) keeps the `<think>` blocks of earlier turns in the rendered history for Qwen 3.6+ templates, and native `reasoning_content` templates (MiniMax M3, Inkling, …) render the reasoning back verbatim. On those paths the next turn *does* contain the previous output, yet the store still discarded it, so every follow-up re-prefilled the whole previous answer. With long reasoning that is the dominant cost of a conversation: a 26k-token answer on Qwen3.8-Flash-Next costs about minute of prefill on the next turn, and the server had already captured the boundary snapshots needed to avoid it (decode-time captures every 2048 tokens, then thrown away).

### Change

- `Request.preserve_reasoning`: set by the chat completions and messages handlers through a new helper, `cache_reasoning_output(settings, native_reasoning, chat_template_kwargs)`. True when the model uses native reasoning content, when the effective template kwargs carry `preserve_thinking=True`, or when the new per-model setting `cache_reasoning_output` forces it; that setting also forces it off. The engines forward the flag into `add_request`.
- Scheduler: one predicate, `_output_tokens_cacheable(request)`, replaces the two `needs_think_prefix` checks. The completion-time boundary snapshot then covers prompt + output, and the final store sequences prompt + output, so the decode-time snapshots already on disk become the intermediate blocks. Requests whose reasoning is stripped keep the prompt-only behaviour exactly as before.
- `ModelSettings.cache_reasoning_output: Optional[bool]` (default `None` = auto). Useful for a custom chat template that renders reasoning in history without advertising `preserve_thinking`.

### Measured

Two-turn probe on an M5 Max, Qwen3.8-Flash-Next-oQ4e-mtp, MTP on, block size 2048: a 3,126-token prompt, a 6,000-token answer, then a follow-up whose history carries the answer with its reasoning.

| | stored at end of turn 1 | reused by turn 2 | turn 2 re-prefill |
|---|---|---|---|
| main b908f563 | 2048 / 9126 tokens, 0 intermediate | 2048 | 7,103 tokens |
| this branch | 8192 / 9126 tokens, 3 intermediate | 8192 (shares the first 8192 of 8192) | 6,962 tokens, all new content |

The same probe with the reasoning *not* sent back matches only the first 3,125 tokens on both, which is the case c8ee5eaf was written for and is unchanged.

### Tests

`tests/test_scheduler_prompt_boundary_store.py` (parser-stop store covers output only when reasoning is preserved), `tests/test_scheduler.py` (prompt + output stored for a think-prefix request with `preserve_reasoning`, alongside the existing prompt-only and non-reasoning cases), `tests/test_scheduler_boundary_completion.py` (completion boundary counts output only when reasoning is preserved), `tests/test_api_utils.py` (helper: auto detection and the setting override). 

### Related

- #3335 skips the decode-time captures as wasted work; with this change they are what makes the next turn hit, so the two should be reconciled: skip them only when `preserve_reasoning` is false.
- The parser-side stop fallback (`_prepare_prompt_boundary_cache_store`) now applies the same predicate, so a tool-call turn's reasoning and call are cacheable for the next request in an agent loop. Both the stock Qwen template and a derived one re-render such a turn token-exactly (tool-call probe: 3595/3595 and 3452/3452 tokens; the server parses argument strings into mappings before templating).
- The MTP commit alignment that lands verify cycles on block boundaries was armed lazily by the first capture attempt. A prompt shorter than one block reaches its first boundary during decode, so the cycle crossing it ran unaligned, the capture was skipped ("speculative decode skew"), and the split-GDN store rejected that block, leaving the whole answer uncached; long prompts never showed it because their prefill captures arm the alignment first. Fixed by detecting at `add_request`. Reproduced in the bench with a 101-token prompt and a 12k-token sampled answer (skew at 2048 in about one run in three, cache offset 2047 or 2049), clean after the change.
- The non-streaming `generate()` of both engines submits through the core engine with an explicit argument list, so the flag first reached only streamed requests (a non-streaming tool-call turn stored 2048 of 4130 tokens while the same request streamed stored 4096); fixed by forwarding it there. `skip_cache_store` has the same gap on that path and is left as is.

### DIFF
--- a/omlx/api/utils.py
+++ b/omlx/api/utils.py
@@ -45,6 +45,19 @@ def uses_native_reasoning_content(
     return "minimax" in lowered and "m3" in lowered
 
 
+def cache_reasoning_output(
+    settings: Any,
+    *,
+    native_reasoning: bool,
+    chat_template_kwargs: dict[str, Any] | None,
+) -> bool:
+    """Whether a reasoning request's output tokens can prefix-match the next turn."""
+    forced = getattr(settings, "cache_reasoning_output", None)
+    if forced is not None:
+        return bool(forced)
+    return bool(native_reasoning) or (chat_template_kwargs or {}).get("preserve_thinking") is True
+
+
 def merge_reasoning_effort_chat_template_kwargs(
     chat_template_kwargs: dict[str, Any] | None,
     reasoning_effort: Any | None,
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -994,6 +994,7 @@ class BatchedEngine(BaseEngine):
             sampling_params=sampling_params,
             tools=tools,
             skip_cache_store=bool(kwargs.get("skip_cache_store", False)),
+            preserve_reasoning=bool(kwargs.get("preserve_reasoning", False)),
             benchmark_trace=bool(kwargs.get("benchmark_trace", False)),
             benchmark_ane_sequence_length=int(
                 kwargs.get("benchmark_ane_sequence_length", 0) or 0
--- a/omlx/engine/batched.py
+++ b/omlx/engine/batched.py
@@ -915,6 +915,7 @@ class BatchedEngine(BaseEngine):
             prompt=prompt,
             sampling_params=sampling_params,
             tools=tools,
+            preserve_reasoning=bool(kwargs.get("preserve_reasoning", False)),
             **specprefill_kwargs,
         )
 
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -3693,6 +3693,7 @@ class VLMBatchedEngine(BaseEngine):
             vlm_cache_key_start=vlm_cache_key_start,
             vlm_cache_key_ranges=vlm_cache_key_ranges,
             skip_cache_store=bool(kwargs.get("skip_cache_store", False)),
+            preserve_reasoning=bool(kwargs.get("preserve_reasoning", False)),
             benchmark_trace=bool(kwargs.get("benchmark_trace", False)),
             benchmark_ane_sequence_length=int(
                 kwargs.get("benchmark_ane_sequence_length", 0) or 0
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -3580,6 +3580,7 @@ class VLMBatchedEngine(BaseEngine):
             vlm_cache_key_start=vlm_cache_key_start,
             vlm_cache_key_ranges=vlm_cache_key_ranges,
             tools=tools,
+            preserve_reasoning=bool(kwargs.get("preserve_reasoning", False)),
             **specprefill_kwargs,
         )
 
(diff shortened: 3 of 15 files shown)


Rascal:
Follow-up pushed since the first revision:

- 39c3dcaf — CI fix: `cache_reasoning_output` is now classified as a universal profile field (`test_all_model_settings_fields_classified`).
- 4f5cd29e — the parser-side stop store (`_prepare_prompt_boundary_cache_store`) applies the same predicate, so a tool-call turn's reasoning and call are cacheable for the next step of an agent loop. Both the stock Qwen template and a derived one re-render such a turn token-exactly (3595/3595 and 3452/3452 tokens in a probe), since the server parses argument strings into mappings before templating.
- aa3fd183 — the non-streaming `generate()` of both engines submits through the core engine with an explicit argument list, so the flag only reached streamed requests. A non-streaming tool-call turn stored 2048 of 4130 tokens while the same request streamed stored 4096. Forwarded explicitly. (`skip_cache_store` has the same gap on that path and is left alone.)
- 6b21d06a — a pre-existing bug this feature exposes. The MTP commit alignment that lands verify cycles on paged block boundaries (`_omlx_mtp_commit_align`) was armed lazily by `_detect_boundary_snapshot_need()`, i.e. by the first capture attempt. A prompt shorter than one block reaches its first boundary during decode, so the cycle crossing it ran unaligned, the capture was skipped ("speculative decode skew": cache offset 2047 or 2049 at boundary 2048), and the split-GDN store rejected the block ("Rejecting split-GDN placeholder block 1 because its recurrent checkpoint was not committed"), leaving the whole answer uncached. Long prompts never showed it because their prefill captures arm the alignment first. Reproduced on a fresh server with a 101-token prompt and a 12k-token sampled answer (about one run in three); a cycle-level trace showed the bonus token landing on 2048 without materialisation. The fix arms the alignment at `add_request`; three runs after it captured every boundary and stored 10240/12101 tokens. Test: a scheduler with a stateful cache layer exposes the alignment right after `add_request`.

jundot:
Thanks for tracking down why preserved reasoning was still being re-prefilled. This is a useful improvement to follow-up TTFT, especially after long reasoning turns. I verified that a 39-token prompt with Lightning MTP captured the first cache boundary, and the next turn reused 4,096 tokens of the generated response.

LGTM. I'll merge this and handle two small follow-ups: make auto-detection respect explicit `preserve_thinking=False`, and forward the flag through the Responses API too.

PR #3528, opened 2026-09-08
fix(oq): record the dtype the output config reports

andyoneal:
oQ's output `config.json` inherits `dtype` from the source, so a `dtype="float16"` build stores fp16 tensors and still reports `bfloat16`. In the same run oQ appends `-fp16` to the output directory name, so the name and the config contradict each other, and the config is the one that is wrong.

Nothing downstream corrects it. mlx-lm and mlx-vlm load weights at their stored dtype and let MLX promote activations from them, and the only reader of the key in this tree is `_target_precision_tag` in `omlx/engine/dflash.py`, which is gated behind a checkpoint having no `quantization`/`quantization_config` section — an oQ output always writes both, so that branch is unreachable for an oQ checkpoint. The field cannot change execution, then; it only changes what a reader concludes about the file, which is the reason it is worth correcting rather than a reason to leave it.

## What changed

`_apply_output_dtype` records the resolved dtype at both config write paths, `quantize_oq_streaming` and the temporary sensitivity proxy. The label comes from the same `dtype == "bfloat16"` test that selects `target_dtype`, so the config agrees with every tensor the quantizer writes.

It does not follow that the key is true of every tensor in a finished checkpoint. A gemma4 build with a Lightning MTP head carries 48 bfloat16 tensors beside the float16 base — all of `language_model.mtp.*` — merged afterwards by `combine_gemma4_assistant_mtp`, which keeps the assistant at its shipped dtype and leaves the quantization config alone. So the key moves from wrong about the 1243 tensors oQ wrote to wrong about the 48 it did not, and it is best read as describing the quantizer's own output.

This follows `_clone_config` in `tools/clone_mlx_model_fp16.py`, which already normalizes `dtype` and `text_config.dtype` for its own clone path. The one deliberate difference is that only keys the source declared are rewritten rather than added, so no checkpoint gains metadata it did not have before.

`vision_config` is left alone. Under a float16 target, `_cast_passthrough_tensor` stores vision and audio weights as float32 rather than float16, so no single key describes both towers and writing `float16` there would be actively false rather than merely stale.

## Testing

`test_output_config_reports_the_dtype_it_wrote` runs the real `quantize_oq_streaming` in both directions — a bfloat16 source built to float16, and a float16 source built to bfloat16 — and checks the written config against the dtype of the tensors actually stored. Both cases fail without the change, with `assert 'bfloat16' == 'float16'` and its mirror.

509 passed and 1 skipped across the oQ suites (`test_oq`, `test_oq_manager`, `test_oq_combine`, `test_oq_fp8_dequant`, `test_oq_fused_qkv_virtual`, `test_oqe_calibration_packaging`). The remaining six test files that touch oQ also pass (220 passed, 8 skipped), as do the admin tests that import it (415 passed).

### DIFF
--- a/omlx/oq.py
+++ b/omlx/oq.py
@@ -166,6 +166,36 @@ def _validate_oq_dtype_for_model(config: dict, dtype: str) -> None:
         )
 
 
+def _canonical_output_dtype(dtype: str) -> str:
+    """Name the dtype oQ will actually store, mirroring ``target_dtype``.
+
+    Both write paths store fp16 for anything not exactly ``"bfloat16"``, so
+    deriving the label from the same test keeps the config from disagreeing
+    with the tensors. Assumes ``dtype`` already passed the ``OQ_DTYPES``
+    check, which is what keeps the fallback from mislabelling a third value.
+    """
+    return "bfloat16" if dtype == "bfloat16" else "float16"
+
+
+def _apply_output_dtype(config: dict, dtype: str) -> None:
+    """Record the dtype oQ wrote, replacing the source's inherited claim.
+
+    A source config describes the checkpoint oQ read, not the one it writes,
+    and nothing in the load path corrects it, so a float16 build of a bfloat16
+    source reads back as bfloat16. Follows ``_clone_config`` in
+    ``tools/clone_mlx_model_fp16.py``, but only rewrites keys the source
+    declared rather than adding any. ``vision_config`` is left alone: under a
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the fix. I verified that both conversion directions now report the stored dtype correctly, including the sensitivity proxy path. Existing configs without dtype keys and vision metadata remain unchanged. LGTM, I'll merge this.

PR #3530, opened 2026-09-08
fix(api): parse attribute-style <function name> tool calls (MiniCPM5)

### LINKED ISSUE #1580: Fail to use tool with  "mlx-community/MiniCPM5-1B-8bit"
CLI client is pi-agent, when loading the model in lmstudio, tell the llm to read a file. everything normal.

But if using omlx to load the model, the pi-agent false to call the "read" tool. Because of the model does not use json format to reply. Can it add the function on omlx to fix the format?

### LINKED ISSUE #3429: Tool calls in the bare `<function name="…"><param name="…">` dialect are never parsed (MiniCPM5 and family)

X-Wei:
## Summary

MiniCPM5-family chat templates emit tool calls as

```
<function name="read"><param name="path">/workspace/TASK.md</param></function>
```

optionally wrapping values in `<![CDATA[...]]>`. None of the XML fallbacks in `omlx/api/tool_calling.py` recognised the attribute-style `<function name="...">` element (the table had keyword-style `<function=NAME>` and attribute-style `<invoke name="...">`, but not this combination), so the call came back as assistant text with `finish_reason: stop` and `tool_calls: null` — the client concludes the model cannot use tools. This is **direction 1** from #3429.

## Changes

- **Wrapped form** — `_parse_xml_tool_calls` gains an attribute-style branch for `<tool_call>`-wrapped payloads, bounded with `rfind("</function>")` like the keyword-style branch.
- **Bare form** — new `_parse_attribute_function_tool_calls`, tried after the namespaced fallback. There is no envelope to anchor on, so it is **gated on the request's own tool declarations**: an element is parsed only when its name exactly matches a declared tool. It is inert without `tools`, and prose that merely mentions `<function name="...">` for an unknown function passes through untouched. An element cannot span a later `<function` open tag, so an unclosed tag in prose cannot swallow a following real call.
- **Streaming** — `ToolCallStreamFilter` registers the `('<function name="', "</function>")` marker pair so the markup is suppressed from content deltas. Like the existing unconditional `<tool_call>` pair, this only runs on tool-enabled streams.
- **CDATA** — the MiniCPM5 template instructs the model to wrap values containing `<`, `&` or newlines in CDATA; values are unwrapped verbatim, and a literal `</param>` inside CDATA does not end the element. Both `<param>` and `<parameter>` spellings are accepted, with a backreference pairing open/close.

Deliberately out of scope: the K2-Horizon `<ifm|...>` dialect from the same issue, which @An0nya has a parser for and offered separately. That half has since landed in #3486; the two touch disjoint regions and this branch merges cleanly onto current main.

## Verification

- `tests/test_tool_calling.py`: 408 passed (16 new tests: bare/wrapped forms, both spellings, multiple calls, prose around calls, schema coercion, CDATA incl. nested literal tags, undeclared-name negative case, no-tools inertness, whitespace tolerance, no-argument call, stream filter across chunk boundaries and incomplete-marker flush).
- Full `pytest tests/ -m "not slow and not integration"`: no new failures compared with a clean `main` worktree in the same environment.
- End-to-end against a real `MiniCPM5-2B-MLX-8bit` (`model_type: llama`), same request from the issue, `temperature: 0`:

  | | before (0.6.4) | after |
  |---|---|---|
  | `finish_reason` | `stop` | `tool_calls` |
  | `tool_calls` | `null` | `read(path="/workspace/TASK.md")` |
  | `content` | raw XML | empty |

  Streaming returns the same tool call as deltas with no XML leaking into content, and a `write` call with multi-line content arrives via CDATA and is decoded to the literal value.
- Agent-level A/B with pi-agent: against the unpatched server it prints the raw `<function ...>` text and stops; against this branch it executes the `read` tool and answers from the file contents.

Fixes #3429
Fixes #1580

### DIFF
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -494,6 +494,29 @@ def _iter_xml_parameters(params_text: str) -> Iterator[Tuple[str, str]]:
         pos = value_end + len(_XML_PARAMETER_CLOSE)
 
 
+_ATTR_FUNCTION_OPEN_RE = re.compile(r'<function\s+name="([^"]+)"\s*>')
+# Both <param> (MiniCPM5's chat template) and <parameter> spellings occur in
+# the wild; the backreference pairs open/close spellings so a literal
+# "</param>" cannot close a <parameter> element. The non-greedy CDATA
+# alternative ensures nested literal tags inside CDATA do not end elements early.
+_ATTR_PARAM_RE = re.compile(
+    r'<(param|parameter)\s+name="([^"]+)"\s*>(?:\s*<!\[CDATA\[(.*?)\]\]>\s*|(.*?))</\1>',
+    re.DOTALL,
+)
+
+
+def _iter_attr_xml_parameters(params_text: str) -> Iterator[Tuple[str, str]]:
+    """Yield ``(key, value)`` for each ``<param name="k">v</param>`` element.
+
+    MiniCPM5's chat template tells the model to wrap values containing
+    ``<``, ``&`` or newlines in a CDATA block; the wrapper is removed here so
+    the argument carries the literal value, verbatim.
+    """
+    for match in _ATTR_PARAM_RE.finditer(params_text):
+        value = match.group(3) if match.group(3) is not None else match.group(4).strip()
+        yield match.group(2), value
+
+
 def _find_marker_span_end(
     text: str, payload_start: int, end_marker: str
 ) -> Optional[Tuple[int, int]]:
@@ -586,6 +609,7 @@ def _parse_xml_tool_calls(
     Handles models that use <tool_call>...</tool_call> XML format, including:
     - GLM format: <tool_call>func<arg_key>k</arg_key><arg_value>v</arg_value></tool_call>
     - Qwen/Llama format: <tool_call><function=name><parameter=key>value</parameter></function></tool_call>
+    - Attribute style: <tool_call><function name="name"><param name="key">value</param></function></tool_call>
     - Generic JSON: <tool_call>{"name": ..., "arguments": ...}</tool_call>
 
     When ``tools`` is provided, parameter values are coerced to their
@@ -623,6 +647,24 @@ def _parse_xml_tool_calls(
             props = _tool_param_properties(func_name, tools)
             arguments = {}
             for key, val in _iter_xml_parameters(params_text):
+                arguments[key] = _coerce_param_value(val, key, props, func_name)
+            _built = _build_tool_call(func_name, arguments)
+            if _built is not None:
+                tool_calls.append(_built)
+            continue
+
+        # Attribute style: <function name="name"><param name="key">value</param></function>
+        # (MiniCPM5 family, #3429). Bounded by rfind like the keyword-style
+        # branch above so a literal close tag inside a value cannot end the
+        # element early.
+        attr_open = _ATTR_FUNCTION_OPEN_RE.match(content)
+        attr_close = content.rfind(_XML_FUNCTION_CLOSE)
+        if attr_open and attr_close >= attr_open.end():
+            func_name = attr_open.group(1)
+            params_text = content[attr_open.end() : attr_close]
+            props = _tool_param_properties(func_name, tools)
+            arguments = {}
+            for key, val in _iter_attr_xml_parameters(params_text):
                 arguments[key] = _coerce_param_value(val, key, props, func_name)
             _built = _build_tool_call(func_name, arguments)
             if _built is not None:
@@ -702,6 +744,71 @@ def _parse_namespaced_tool_calls(
 
     cleaned = re.sub(pattern, "", text, flags=re.DOTALL).strip()
     return cleaned, tool_calls
+
+
+_ATTR_FUNCTION_ELEMENT_RE = re.compile(
+    r'<function\s+name="([^"]+)"\s*>((?:(?!<function[\s>]).)*?)</function>',
+    re.DOTALL,
+)
+
+
+def _parse_attribute_function_tool_calls(
+    text: str, tools: Optional[List] = None
+) -> Tuple[str, Optional[List[ToolCall]]]:
+    """
+    Fallback parser for bare attribute-style tool calls (#3429).
+
+    MiniCPM5-family chat templates emit
+    ``<function name="name"><param name="key">value</param></function>``
+    with no wrapper marker at all, so none of the envelope-anchored parsers
+    ever see it.
+
+    Without an envelope the only safe anchor is the request's own tool
+    declarations: an element is parsed only when its name exactly matches a
+    declared tool, so prose that merely mentions ``<function name="...">``
+    markup for an unknown function passes through untouched, and the parser
+    is inert when the request declares no tools.
+
+    Elements are bounded at the first ``</function>`` (there is no outer
+    envelope to bound by, and rfind would merge sibling calls), accepting
+    that a literal close tag inside a value ends the element early.
+    An element also cannot span a later ``<function`` open tag, so an
+    unclosed tag in prose cannot swallow a following real call.
+
+    Returns:
+        Tuple of (cleaned_text, tool_calls or None)
+    """
+    if not tools:
+        return text, None
+    registered = _extract_tool_names(tools)
+    if not registered:
+        return text, None
+
+    tool_calls = []
+    spans: List[Tuple[int, int]] = []
+    for match in _ATTR_FUNCTION_ELEMENT_RE.finditer(text):
+        func_name = match.group(1)
+        if func_name not in registered:
+            continue
+        props = _tool_param_properties(func_name, tools)
+        arguments = {}
+        for key, val in _iter_attr_xml_parameters(match.group(2)):
+            arguments[key] = _coerce_param_value(val, key, props, func_name)
+        _built = _build_tool_call(func_name, arguments)
+        if _built is not None:
+            tool_calls.append(_built)
+            spans.append(match.span())
+
+    if not tool_calls:
+        return text, None
+
+    out: List[str] = []
+    last = 0
+    for span_start, span_end in spans:
+        out.append(text[last:span_start])
+        last = span_end
+    out.append(text[last:])
+    return "".join(out).strip(), tool_calls
 
 
 def _parse_hermes_tool_calls(text: str) -> Tuple[str, Optional[List[ToolCall]]]:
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for adding this; I confirmed the basic MiniCPM5 call parses correctly. I reproduced three gaps in the server’s SSE path that I’d like addressed before merging:

- The stream filter removes `<function name="example">...</function>` from prose even when `example` is not a declared tool, while non-streaming preserves it.
- A literal `</function>` inside CDATA truncates the call and produces `{}` arguments. A complete function example inside CDATA can instead be extracted as the call.
- `<function  name="write">` with two spaces parses successfully but leaks the raw XML into content alongside the tool call.

Could you align the parser and stream filter on declared names, CDATA boundaries, and accepted whitespace, and add streaming regression cases for these? I’d expect undeclared examples to remain visible, CDATA values to arrive verbatim, and accepted calls to produce no XML content deltas.

X-Wei:
Thanks for the careful review — all three were real, and all three are now fixed in `3c4900f3`.

**1. Undeclared names leaking into the filter.** The bare `<function name="...">` / `</function>`
marker pair was unconditional in `ToolCallStreamFilter`, so it suppressed the dialect regardless of
what the request declared. It is now gated on the registered tool names, and `tools` is threaded
into the filter at all three SSE construction sites (`stream_chat_completion`,
`stream_anthropic_messages`, `stream_responses_api`). Undeclared examples in prose now stream
through byte-identical to the non-streaming path.

**2. CDATA boundaries.** Both paths now skip `<![CDATA[...]]>` blocks when looking for the element
boundary — non-streaming via `_find_attr_function_span`, streaming via
`_find_attr_function_suppression_end`. A literal `</function>` inside CDATA no longer truncates the
call, values arrive verbatim, and a complete function example nested inside CDATA is kept as a value
instead of being extracted as the call. The streaming side also tracks CDATA state across `feed()`
calls, so `<![CDATA[` and `]]>` split across chunk boundaries are handled.

**3. Accepted whitespace.** The opening tag is matched by regex with arbitrary whitespace, and
partial-prefix matching was extended to the bare dialect so a tag arriving one character at a time
is held back rather than flushed as content. `<function  name="write">` now produces the call with
no XML content deltas.

Regression coverage: 8 new cases in `TestAttributeStyleFunctionDialect`, including the
chunk-boundary ones (`..._cdata_with_literal_function_close_across_chunks`,
`..._cdata_delimiters_split_across_chunks`, `..._multiple_spaces_in_tag`). I also re-ran your three
scenarios at 1-char, 7-char and whole-blob chunk widths and confirmed the streamed content is
identical to the non-streaming cleaned text in each. `tests/test_tool_calling.py` is 416 passed.

One thing I deliberately left alone, flagging it in case you'd rather it were uniform: the three
sessions in `omlx/adapter/output_parser.py` construct `ToolCallStreamFilter(tokenizer)` without
tools, so bare-dialect suppression is inert on that path. That matches pre-PR behaviour (the marker
pair didn't exist there before this PR), and DeepSeek-V4 / MiniMax-M3 sessions don't carry tools at
all — but `BailingHybridOutputParserSession` does have `self._tools` in hand, so I can wire it
through if you want the filters aligned everywhere.

Also, CI hasn't run on this branch yet — it's sitting at `action_required` since I'm a first-time
contributor. Happy to rebase if you'd like a clean run.

rolfb:
Excited for this. Is branch conflicts holding it back, or is it a different `action_required`?

jundot (approved):
I verified the earlier fixes through the server SSE path. I also prepared a small follow-up for CDATA values containing `<tool_call>` examples, which could still replace the outer call or suppress it entirely. With that fix, both bare and wrapped calls preserve the original arguments across Chat Completions, Anthropic Messages, and Responses, including streaming. LGTM with the follow-up ready; I'll resolve the main conflicts, merge this, and apply the fix immediately afterward.

PR #3532, opened 2026-09-09
Local usage history

thatcherstudio:
Adds a local, hourly usage history so you can see what your models
actually did over today / 7 / 30 / 90 days / this month: requests,
tokens, cached tokens, per-model output speed, and a day/hour heatmap.
It's in the dashboard Status tab and the Mac app's Status screen.

Everything stays on the machine. It stores only the canonical model id,
the hour, counts, and summed durations in `<base_path>/usage.sqlite3`.
No prompts, responses, keys, or client info. Writes happen on a
background thread; inference never waits on it. 400-day retention,
corruption is moved aside and a fresh file started. Existing all-time
stats and their clear buttons are untouched.

There's a switch to turn it off (Settings on the dashboard, Server
screen in the app, or `OMLX_USAGE_HISTORY=0`). Off means it stops
recording and leaves the file alone so turning it back on resumes.

Design notes in `docs/usage-analytics.md`.

Tested with the full pytest suite, the node test for the panel, and an
xcodebuild of the Mac app. Four commits, one per layer, if that's
easier to review.

🤖 Generated with [Claude Code]


jundot:
Thanks for implementing this. I tested the dashboard, and the per-model heatmap makes it easy to see when and how much each model was used. It feels like a useful addition to the dashboard. I’ll resolve the template conflict and merge this, then handle the idle database work, disabled-startup behavior, response size, and readability cleanup in one follow-up commit.

Issue #2197, opened 2026-07-11
v0.5.0: engine permanently deadlocks in mlx::core::eval_impl on first ~16k-token prompt (GLM-5.2 + MTP); rc1 and rc1+#2160+#2150 are fine


anicaise-ai:
**Describe the bug**
On v0.5.0 (`bc7de7d8`), the very first request with a long prompt (~16k tokens) permanently wedges the engine: the request never completes, the server process drops to 0% CPU, and subsequent requests are accepted at the HTTP layer but never reach the engine (nothing logged). Only a kill/restart recovers it.

A `sample` of the hung process shows the mlx eval thread parked forever in `mlx::core::eval_impl` on a condition variable that is never signaled, with a second Python thread blocked acquiring a lock:

```
Thread A (mlx eval):
  mlx::core::eval(std::vector<mlx::core::array>)
    mlx::core::eval_impl(...)
      std::condition_variable::wait(...)
        __psynch_cvwait          <- parked for the whole 2s sample (2180/2180 frames)

Thread B (Python):
  lock_PyThread_acquire_lock
    _PyMutex_LockTimed -> _PyParkingLot_Park -> __psynch_cvwait
```

I can share the full `sample` capture if useful.

**To Reproduce**
1. Serve GLM-5.2 (mxfp4, MTP head present, `mtp_enabled: true`, Lightning MTP path active) on v0.5.0 built with `OMLX_WITH_CUSTOM_KERNEL=1`.
2. Send a single non-streaming chat completion with a ~16k-token prompt.
3. Request hangs forever; engine is wedged for all subsequent traffic (0% CPU, nothing logged).

Reproduced on the first attempt. I don't know yet whether it is 100% deterministic, but the same setup on rc1 has served days of production traffic including long prompts.

**Expected behavior**
Long-prompt prefill completes (or at worst gets throttled), and the engine keeps serving. That is what v0.5.0rc1 does on the exact same prompt: it logs a "No idle model evicted" throttle but completes and keeps serving.

**Bisection data point**
[...]

PR #3533, opened 2026-09-09
Fused-batch ragged multi-token MTP: concurrent speculative decode (OMLX_MTP_FUSED_BATCH)

gaineyllc:
## What

An **opt-in fused-batch MTP path** (`OMLX_MTP_FUSED_BATCH=1`, default **off** = stock behavior) that gives concurrent MTP-capable sequences **both** the single-stream MTP latency win **and** batched aggregate throughput at the same time — vLLM-style ragged speculative decode, on Apple Silicon.

Today the choice is binary: MTP-off continuous batching (good aggregate) *or* singleton MTP (good single-stream latency, one row). The existing row-wise batched MTP (`OMLX_MTP_ROWWISE_BATCH=1`) runs a full backbone forward per row, so it does not amortize and is a throughput loss. This PR fuses the expensive part.

## How

When a continuous batch holds N MTP-capable rows and the flag is set, `_mtp_batch_next` dispatches to `_run_verify_cycle_batched`:

1. **One batched `(N, k+1)` verify forward** against the shared cache (not N per-row forwards).
2. Per-row greedy accept → each row's own run length `m_i`.
3. **Ragged commit** — each row commits its **own** `m_i+1` tokens in a single `GenerationBatch.next()` via **multi-token emit** (`_emit_ragged_responses`). The scheduler's `_process_batch_responses` iterates responses and appends per-uid with no dedup, so more than one `Response` per uid is honoured. (This replaces a min-commit approach that committed only `min_i(m_i)+1` and discarded the rest.)
4. **Per-row rollback vector** keeps `m_i+1` of `k+1` for each row.
5. Each row re-drafts `k` fresh tokens, so all rows re-present a rectangular `(N, k+1)` verify next cycle regardless of how many they committed.

## Scope & safety

- **Greedy only**, hybrid **GDN + PLE** targets (`qwen4_exp` / `qwen3_5`, e.g. Qwen3.8-Flash-Next). **commit-align is supported** via a per-row boundary cap (mirrors the singleton path, so paged-cache boundary snapshots stay clean).
- Out-of-scope cases (logits processors, non-greedy, `head_clone`, <2 rows, depth-0) **decline cleanly** via `_MtpStepFallback` → the standard batched step. **Flag off = byte-for-byte stock**; existing `tests/test_mlx_lm_mtp_patch.py` is unchanged and green (144 passed).
- **`OMLX_MTP_FUSED_K`** caps per-cycle draft depth. Lower depth wins *aggregate* throughput (fewer verify positions per committed token). Measured optimum on M5 Max / Flash-Next is **2**.

## Dependency

The per-row (ragged) rollback needs an `mlx-vlm` whose `rollback_speculative_cache` handles a ragged batch on the serving `BatchQSAKVCache`. Current pins raise on it (pre-fix `c.keys` guard). Fix is up at **Blaizzy/mlx-vlm#2197**. Until the `mlx-vlm` pin is bumped, the fused path **safely declines** to the standard batched step on affected caches (the rollback error is converted to `_MtpStepFallback`) — no crash, no cache corruption.

## Measured (M5 Max, Qwen3.8-Flash-Next, temp 0, greedy, wall-clock aggregate)

Draft-depth sweep at 8-way concurrency (tok/s aggregate): **k=1 → 38.6, k=2 → 51.6, k=3 → 29.7** (sharp peak at k=2).

Concurrency scaling at k=2 (tok/s aggregate, distinct prompts, coherent output on every row):

| concurrency | aggregate tok/s |
|---|---|
| 6  | 28.6 |
| 8  | 51.6 |
| 12 | 53.2 |
| 16 | 53.0 |

~6× the row-wise batched MTP path; plateau is box compute-saturation (~8-way). Output verified coherent across all rows; batches form up to N=16.

## Tests

- `tests/test_fused_batch_mtp.py` (new, model-free): env gating (`_fused_batch_mtp_enabled` / `_any_batch_mtp_enabled`) and the ragged multi-token emit contract (`_emit_ragged_responses`) — including a row emitting its full ragged run and a row truncating mid-run on a length finish. 5 passed.
- `tests/test_mlx_lm_mtp_patch.py` unchanged: 144 passed.
- End-to-end verified live on a running server (Flash-Next, 6–16 concurrent).

```bash
pytest tests/test_fused_batch_mtp.py tests/test_mlx_lm_mtp_patch.py -m "not slow" -v
```

## Try it

```bash
OMLX_MTP_FUSED_BATCH=1 OMLX_MTP_FUSED_K=2 omlx serve   # + Blaizzy/mlx-vlm#2197 for the full ragged speedup
```

Happy to adjust the default `k`, gate it on a config field instead of an env var, or split the commit-align support out if you'd prefer a smaller first cut.

### DIFF
--- a/omlx/patches/mlx_lm_mtp/batch_generator.py
+++ b/omlx/patches/mlx_lm_mtp/batch_generator.py
@@ -37,6 +37,32 @@ Known limitation (compute-bound single-stream Apple Silicon):
   smaller per-step backbone, or under continuous batching where spare
   compute exists. See #1097 / #1311 for measurements.
 
+Fused-batch MTP (``OMLX_MTP_FUSED_BATCH=1``, opt-in, default off):
+  When multiple MTP-capable sequences share a continuous batch, run ONE
+  batched ``(N, k+1)`` verify forward across all rows against the shared
+  cache instead of N per-row backbone forwards (which the row-wise path,
+  ``OMLX_MTP_ROWWISE_BATCH=1``, does and which does not amortize). Each row
+  greedily accepts its own run length ``m_i``, commits its OWN ``m_i+1``
+  tokens in one ``next()`` via multi-token emit (``_emit_ragged_responses``;
+  the scheduler appends responses per-uid, so >1 Response/uid is honoured),
+  the shared cache is rolled back by a per-row accepted vector, and every row
+  re-drafts ``k`` fresh tokens so all rows re-present a rectangular verify
+  next cycle. This is vLLM-style ragged spec-decode: MTP's single-stream
+  latency win AND batched aggregate throughput at the same time.
+    - Scope: greedy only; hybrid GDN + PLE targets (qwen4_exp / qwen3_5,
+      e.g. Qwen3.8-Flash-Next); commit-align is supported. Out-of-scope
+      cases (logits processors, non-greedy, head_clone) decline cleanly via
+      ``_MtpStepFallback`` to the standard batched step — flag OFF is stock.
+    - ``OMLX_MTP_FUSED_K`` caps the per-cycle draft depth. Lower depth wins
+      for AGGREGATE throughput (fewer verify positions per committed token);
+      measured optimum on M5 Max Flash-Next is ``2`` (k=1: 38.6, k=2: 51.6,
+      k=3: 29.7 tok/s aggregate at 8-way). Unset uses the primed depth.
+    - Requires an mlx-vlm whose ``rollback_speculative_cache`` handles a
+      ragged batch on ``BatchQSAKVCache`` (Blaizzy/mlx-vlm#2197); older
+      versions raise and the fused path safely declines to the standard step.
+    - MEASURED (M5 Max, Flash-Next, temp 0, wall-clock aggregate): 8-way 51.6,
+      12-way 53.2, 16-way 53.0 tok/s, coherent, ~6x the row-wise path.
+
 Greedy identity (sampler is None): the patched dispatch produces the same
 tokens as the standard step. PR 990's ``test_mtp_generate_identity``
 encodes this contract; the oMLX-side equivalent lives in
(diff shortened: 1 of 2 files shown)


moc375:
Ran this branch end-to-end on M3 Ultra 512GB (Qwen3.8-Flash-Next-oQ4e-mtp, greedy, concurrent). The fused path engages and reaches 3.0 tok/row/cycle at k=2, and mlx-vlm#2197 is indeed the critical dependency. Two correctness issues we found in the code, both in the same class as the review points on #3265:

## Bug 1: terminal response cache can contain un-emitted drafts

In `_emit_ragged_responses`, when a row finishes mid-ragged-run (stop/length at draft j < m_i):

- The ragged rollback has already kept `m_i + 1` positions in the cache (line ~2804).
- The row only emits `j + 1` tokens.
- The finish branch does `prompt_cache=gen_batch.extract_cache(idx)` **without trimming the remaining `m_i - j` accepted drafts**.
- The extracted cache goes into the Response → scheduler → paged prefix store, so a subsequent request matching this prefix sees a divergence at the speculative positions.

Same class as review point 2 on #3265. Trigger requires: row finishes mid-ragged-run + context > block_size + store succeeds.

## Bug 2: fallback after backbone forward advances the cache without restoring committed state

When ragged rollback fails (line ~2809/2811, e.g. no mlx-vlm#2197):

1. `_call_backbone` has already advanced the cache by `k + 1` per row (line ~2738).
2. `_MtpStepFallback` is raised.
3. The handler calls `_reconcile_mtp_batch_to_standard` (line ~192), which extracts caches and re-merges the batch — but **never trims the speculative tail**.
4. The standard step then feeds `next_main`, advancing the cache again.
5. Net: cache is `k + 1` positions ahead of committed tokens.

Same class as review point 4 on #3265. This fires on **every rollback failure** — currently that means every user without mlx-vlm#2197 applied, which is everyone (it's not merged yet).

## Performance context (if useful for the PR description)

| shape | fused off | fused on (no #2197) | fused on (#2197 patched) |
|---|---|---|---|
| c=2 × 4k aggregate | 43 tok/s | 17 tok/s (BAIL loop tax) | **47 tok/s** (+9%) |
| c=4 × 4k aggregate | 52 tok/s | 44 tok/s | **66 tok/s** (+27%) |

The BAIL loop (RAGGED → rollback fail → decline → bootstrap → retry) is the reason unpatched performance is *worse* than plain batching. With #2197, the path runs continuously and the aggregate gain at c=4 is +27%.

Happy to help test a fix for either bug.

pierre427:
Following up on the two issues moc375 raised on 09-11. Both come from deciding things
after the cache has already moved. A small reorder in `_run_verify_cycle_batched_impl`
covers both:

```python
# 1) before _call_backbone: fail over while nothing has been mutated
if not _ragged_rollback_supported(gen_batch.prompt_cache):   # probe cache types once
    raise _MtpStepFallback("fused MTP: ragged rollback unsupported")
logits, hidden, gdn = _call_backbone(...)
# 2) run each row's stop/length check on host tokens *before* rollback,
#    so the rollback vector is what is actually emitted
emit_len = [_emitted_len(gen_batch, i, per_row[uid]) for i, uid in enumerate(uids)]
accepted_vec = mx.array([e - 1 for e in emit_len], dtype=mx.int32)
_rollback_after_reject(..., accepted=accepted_vec, block_size=k + 1)
```

With (2), `extract_cache(idx)` in `_emit_ragged_responses` can no longer contain tokens
that were never emitted. With (1), the `_MtpStepFallback` after a forward goes away, and
so does the BAIL loop that makes unpatched runs slower than plain batching.

We use the same rule in our MLX cache layer: check that every entry supports a ragged
rewind first, and never degrade to a uniform trim.
https://github.com/pierre427/mlx2/blob/41ab59c/src/mlx2/runtime/models/cache.py#L537-L592
(MIT). It is a pattern, not a drop-in, since your caches are mlx-vlm's. I haven't run this
branch myself.


---
<sub>AI-assisted: drafted with Claude from our code, then reviewed. The links are pinned to a public snapshot of the code, and the claims above state the evidence level we actually have.</sub>

PR #3534, opened 2026-09-09
perf(qwen4_exp): faster Qwen3.8-Flash-Next prefill: hyper-connections on prefill rows, SSD-backed PLE gather assembled on the host and gathered ahead

Rascal:
### TLDR

Speeding up Prompt Prefill

### Summary

Prefill of Qwen3.8-Flash-Next on an M5 Max runs at about 1.15k tok/s at every context length, so the cost is in per-token work that every 2048-token chunk repeats, not in attention. A per-component profile of a chunk (1,817 ms) put the hyper-connections at 14% and the per-layer-embedding (PLE) layer at 14%, both far above what their data movement justifies. This branch takes those two down: the chunk is 1,536 ms afterwards (750 µs per token, −15.5%), the 16k prefill goes from 14.2 to 11.3 s and the 134k prefill from 107 to 92 s in the profiled arms; the interleaved main-versus-branch numbers are in the table below. Output is unchanged: the hyper-connection path is within the fused kernels' existing tolerance and the PLE gather is bit-identical.

Three commits on top of `main` b908f563:

1. **Hyper-connections on prefill rows.** Prefill chunks took the canonical `Qwen4ExpGatedResidual` path, where the grouped RMSNorm round-trips through fp32 in five passes and the stream mean is a strided reduce: 97 calls at 2.6 ms per chunk. Rows above the fused kernels' 16-row limit now run the same fp32-accumulating norm kernel the fused path uses (grid over rows), keep the three projections on the quantized matmul, and compute the stream mean as compiled slice products. Per call at 2048 rows: 2.24 → 1.19 ms; in situ 250 → 140 ms per chunk. Same kill switch (`OMLX_QWEN4_HC_FUSED`) and failure latch; parity with the canonical path within the fused tolerance at 17 and 2048 rows.

2. **SSD-backed PLE rows assembled on the host, one upload per chunk.** With `qwen4_ple_ssd_offload`, the disk-backed gather walked every index in Python (a bisect per element, per-shard position lists) and issued three uploads, a dequantize and a scatter-add per touched shard; a chunk's 32,768 rows touch all 129 shards, so the warm cost was 78 ms per chunk before any SSD read. Shards are found with a `searchsorted`, each tensor family's rows are copied into one host buffer in index order, and the chunk takes one upload per family and one dequantize: 14–16 ms warm. Bit-identical; the per-shard path stays as the fallback for tables whose touched shards differ in dtype or quantization.

3. **The next chunk's PLE rows are gathered while the current chunk runs.** The remaining cost was the SSD read itself, on the critical path because the forward evaluates the hashed indices and then reads rows while the GPU idles (160 ms per chunk after commit 2). Both scheduler prefill loops and mlx_lm's prompt loop now tell the model which tokens follow before running a chunk (`prefetch_ple(next_ids, current_ids)`, forwarded by the VLM adapter); the language model hashes that chunk's rows using the tail of the current chunk as the n-gram history and hands the host gather to a single worker thread; the next forward consumes the assembled buffers when its indices match and gathers inline otherwise. Two pending slots, because chunk k+1 is announced before chunk k gathers; the first chunk is announced before the loop. In situ the PLE layer goes from 248 to 35 ms per chunk with an 18 ms gather. The mlx_lm prompt-loop patch is a copy of the loop with the one call added, installed under the scheduler's existing mRoPE wrapper (now late-bound), and is a no-op for models without the hook.

### Measured

M5 Max 128 GB, Qwen3.8-Flash-Next-oQ4e-mtp with `qwen4_ple_ssd_offload` on (the table on SSD), native kernel extensions built, one interleaved chain (B A A B, fresh server per arm), code prompts, MTP off. Prefill speed is tokens prefilled divided by the server's exact stream TTFT for the first pass of each prompt; the prompts are nested prefixes of one corpus, so each size prefills only the tokens the previous one did not (second column). Mean of two arms; `main` b908f563 → this branch (93a1bc7c).

| prompt | tokens prefilled | main | this branch | |
|---|---|---|---|---|
| 16k | 16,010 | 1,170 tok/s | 1,556 tok/s | +33% |
| 63k | 49,127 | 1,191 tok/s | 1,402 tok/s | +18% |
| 134k | 72,110 | 1,194 tok/s | 1,353 tok/s | +13% |
| 229k | 95,454 | 1,159 tok/s | 1,301 tok/s | +12% |
p3_prefill

Wall time for the same rows: 13.7 → 10.3 s, 41.3 → 35.0 s, 60.4 → 53.3 s, 82.4 → 73.3 s.

The app's own admin benchmark (cold prompts of exactly N tokens with a unique prefix, `skip_cache_store`, the engine's `prompt_tps`, quick warm-up, 128 generated tokens, MTP off) run the same way, one fresh server per arm, B A A B:

| prompt | main | this branch | | time to first token |
|---|---|---|---|---|
| 16k | 1,165 tok/s | 1,554 tok/s | +33% | 14.1 → 10.5 s |
| 64k | 1,219 tok/s | 1,433 tok/s | +18% | 53.8 → 45.7 s |
| 128k | 1,243 tok/s | 1,390 tok/s | +12% | 105.4 → 94.3 s |
p3_prefill_app

Generation is the same on both sides at 64k and 128k (about 51 and 46.5 tok/s); at 16k the benchmark's first timed prompt after warm-up reports a low generation rate on one arm of each side (18.8 tok/s against 49 to 57 on the other), a benchmark artefact unrelated to the branch.

Decode is unchanged (serial 59.4 / 53.3 / 47.2 / 41.0 → 59.8 / 53.8 / 47.3 / 41.3 tok/s at the four sizes, within 1%), and so is the second-pass tail re-prefill. The 16k prompt is the first after boot, so its PLE pages are cold; the larger sizes reuse pages the previous prompt touched, which is why the gain narrows with context. Per 2048-token chunk in a fine-grained profile: 1,817 → 1,536 ms; hyper-connections 250 → 144 ms, PLE layer 248 → 35 ms, the rest unchanged.

### Tests

`tests/test_qwen4_hc_fused.py` (prefill path parity at 17/2048 rows, dispatch, kill switch), `tests/test_mlx_vlm_qwen4_exp_compat.py` (host-assembled gather bit-exact on multi-shard chunks with duplicates and batch shapes, one upload per family, out-of-range indices, prefetch hit/miss with two slots, n-gram prefetch indices equal to the next call's, prompt-loop lookahead under the scheduler wrapper), `tests/test_scheduler_chunked_prefill.py` (both scheduler loops announce the next chunk and the first chunk), `tests/test_vlm_model_adapter.py` (adapter forwards the hook).

### Related

#3492 is a partial overlap with first commit. A new opt-in Metal kernel replacing the same hyper-connection RMSNorm on prefill rows, projections and stream mean untouched; the author measures 1.38× on the module at 2048 rows and no significant end-to-end change on an M4 Max. Ours reuses the fused norm kernel already merged for the decode path, replaces the stream mean too, is on by default under the existing kill switch, and measures −7.6% per chunk in situ. Both hook the same line of the module's call, so whichever lands second has a one-line conflict; if #3492 goes first, first commit would keep only the tail.


williamxie1989:
Thanks @Rascal — this is a solid win on my machine. Ran it end to end and every prefill row beats `main`.

**Setup:** M5 Max 128 GB, macOS 26.5.1, `Qwen3.8-Flash-Next-oQ4e-mtp` with `qwen4_ple_ssd_offload` on and MTP enabled.  Branch rebased onto current `origin/main` (`94530d8d`) — the three commits replay cleanly with no conflicts, and the four test files this PR touches are green (238 passed).

Numbers below are from omlx's built-in speed benchmark:

| Test | TTFT (ms) | TPOT (ms/tok) | pp TPS | tg TPS | E2E latency | Throughput | Peak Mem |
|---|---|---|---|---|---|---|---|
| pp4096/tg128 | 2,993.4 | 17.77 | 1,368.3 tok/s | 56.7 tok/s | 5.256s | 803.6 tok/s | 72.32 GB |
| pp8192/tg128 | 5,412.1 | 20.18 | 1,513.6 tok/s | 51.5 tok/s | 5.960s | 1,378.9 tok/s | 72.57 GB |
| pp16384/tg128 | 11,035.5 | 21.09 | 1,484.7 tok/s | 50.1 tok/s | 11.424s | 1,435.8 tok/s | 73.30 GB |
| pp32768/tg128 | 22,231.5 | 14.61 | 1,473.9 tok/s | 69.9 tok/s | 22.961s | 1,429.2 tok/s | 74.74 GB |
| pp65536/tg128 | 45,543.9 | 15.38 | 1,439.0 tok/s | 65.5 tok/s | 47.506s | 1,382.2 tok/s | 78.19 GB |
| pp131072/tg128 | 93,588.7 | 25.18 | 1,400.5 tok/s | 40.0 tok/s | 96.798s | 1,355.4 tok/s | 85.28 GB |

The prefill rate stays flat from 4k out to 128k, which lines up with your framing that the cost was per-token work each chunk repeats rather than attention — keeping the PLE gather off the critical path is what makes the long-context rows hold up.

Two notes from my side, neither a problem:
- MTP is on in my config, so `tg TPS` here includes speculative acceptance; your decode table is measured with MTP off, so that column isn't directly comparable.
- Peak memory grows gently with context (72.3 GB at 4k → 85.3 GB at 128k), so the host-assembled buffers don't look like they're adding unusual pressure.

Really appreciate you digging into the per-component profile and shipping the host-assembled gather plus the lookahead — a PLE gather sitting on the critical path is exactly the kind of cost that stays invisible in a headline tok/s number. Much appreciated!

jundot:
Thanks @Rascal. I tested this on an M3 Ultra with `Qwen3.8-Flash-Next-oQ4e-mtp`, using the built-in Code Python benchmark with 128 generated tokens, SSD PLE enabled, and no prefix cache.

Prefill throughput (tok/s):

| MTP | Input | Main | Full PR | Existing norm + PLE/mean optimizations |
|---|---:|---:|---:|---:|
| OFF | 16k | 925.6 | 1,157.6 | 1,113.3 |
| OFF | 64k | 1,042.3 | 1,241.8 | 1,186.3 |
| OFF | 128k | 1,073.4 | 1,218.6 | 1,167.9 |
| ON | 16k | 906.6 | 1,066.1 | 1,085.4 |
| ON | 64k | 1,017.6 | 1,170.8 | 1,155.3 |
| ON | 128k | 1,047.7 | 1,172.7 | 1,135.9 |

I also checked greedy generation with a 16k chat prompt, a 512-token limit, and normal EOS handling. With MTP off, the full PR first differed from main at generated token 5. Restoring only the existing prefill normalization while retaining the PLE and compiled mean optimizations matched all 512 tokens in that case. Neither output reached EOS. The divergence alone does not demonstrate a quality regression.

Keeping the existing normalization still delivered roughly **8–20% faster prefill**, so I'd like to land that combination first. I'll merge this and immediately follow up by restoring the existing prefill normalization and fixing PLE teardown so active reads finish before the mappings close. I'd like to revisit the prefill normalization optimization in a separate PR, with its numerical behavior and quality impact evaluated independently.

Rascal:
@jundot Thanks for the thorough testing, and for the work you put into maintaining this project and reviewing the enormous stream of PRs.

Rascal:
@jundot Follow-up on the prefill normalization question, measured on an M5 Max with Qwen3.8-Flash-Next-oQ4e-mtp, MTP off:

1. Element level. The fused norm differs from `Qwen4ExpRMSNorm` in ~90-130 of 21M outputs per 2048-row call (0.0005%), every difference exactly one bf16 ulp, with or without outlier-heavy rows. A precise `rsqrt` removes only a handful of those; the rest is fp32 summation order inside the sum of squares, so a bit-exact fused norm would need MLX's own `rms_norm` reduction order.

2. Model level, same 16k prefix prefilled with each norm (six prefixes, last 512 positions each): next-token KL 0.028 mean, argmax agreement 97.0% (one prefix 92.4%). Repeating the same configuration is bit-identical, so this is deterministic.

3. Control. Adding an unrelated one-ulp bump to 5e-6 of the canonical norm's outputs produces the same KL per prefix (0.114 / 0.018 / 0.011 / 0.004 / 0.002 / 0.002 vs 0.134 / 0.016 / 0.011 / 0.003 / 0.002 / 0.003) and the same argmax agreement. The network amplifies any rounding-level change over a 16k prefill into a few percent of argmax flips; a greedy fork at token 5 is the expected signature of any non-bit-exact prefill kernel, not of this one in particular.

4. Quality. Teacher-forced NLL of the true next tokens at the same 3066 positions: fused 0.5704 vs canonical 0.5716 nats/token, paired difference -0.0012 +/- 0.0043. No measurable difference in either direction.

Keeping the canonical norm costs the 3-5% prefill your table shows, so the fused norm is a judgement call rather than a correctness question. If you would consider taking it back under these numbers, say so and I will open a separate PR with the fused norm alone plus the parity and quality checks above as tests; otherwise I leave it as is.

PR #3535, opened 2026-09-09
Add opt-in Uno speculative decoding for dense K2 models

mbarnson:
NOTE: While I'm not giving up on use of Uno models on Apple Silicon, after beating on this for several days I remain thoroughly unimpressed with long-horizon decode performance. Uno beats the bare model for perhaps the first 4k or 8k tokens, then begins to fall off a tokens/second decode cliff when you get to the 25K+ range, while the bare 7B model chugs along with only linear degradation of performance. 

My leading hypothesis is that the two multi-token passes become too expensive relative to how many tokens they commit. And I suppose the mechanism makes sense from the code... With block eight, each Uno cycle does:

* A full eight-position conditional draft forward, including attention over the existing context.
* A full eight-position target verification forward, again attending over that context.
* Commit the accepted prefix plus the applicable seed/correction/bonus tokens; discard the rest.

Both forwards traverse the transformer. The draft’s attention results cannot generally replace verification’s because the input tokens differ. At short contexts, processing several positions together can amortize expensive weight reads. As context grows, the attention work associated with all those speculative positions grows too... including positions ultimately discarded. Kernels can share KV loads across queries, so I don't think there are sixteen separate DRAM scans per cycle, but the extra query-dependent computation remains.

If the Uno drafter didn't have to attend to the entire KV cache -- like Deepseek does, with positional attention -- then the overall Big O time might stay more linear rather than quadratic. But in its current incarnation, this suggests that my current Uno block strategy stops paying for its additional attention work as the context grows. The fixed block of eight may be a good short-context choice without being a good long-context one.

I may come back to it, but for now I'll just leave this PR in "draft" and update it from time to time to see if my teammates working on oMLX find great uses for diffusion draft models on Apple Silicon without abundant idle GPU resources like I can find on a Blackwell GPU.

Anyway, on to the description:

This PR adds opt-in Uno speculative decoding for compatible dense K2 models on top of the merged K2 support (#3486). One active request uses conditional adapter proposals and target verification. Overlapping requests use the existing continuous batch scheduler; Uno resumes when one request remains. EOS, output budgets, cancellation and prefix reuse retain the shared KV lifecycle, and snapshots exclude unconsumed speculation.


mbarnson:
I'm going to pull this back to draft. Uno gives a performance boost for the 7B model -- and is supposed to, when available, for 32B -- but I'd like to make the approach feel more native like MTP and the 0.9B model is actually slower with Uno on my Mac than without.

PR #3536, opened 2026-09-09
fix(mtp): attach the gemma4 assistant head on the mlx-lm path

andyoneal:
A Gemma 4 merged-assistant checkpoint served through mlx-lm advertises MTP, accepts `mtp_enabled: true`, logs `Speculative backend selected: Lightning MTP` — and never drafts a token. This attaches the head on that path: **+28% decode** at short context, **+11.6%** at long context on a 31B oQ4 text-only quant.

## What is wrong

Google ships the Gemma 4 draft head as a separate `gemma4_assistant` model, which `oq.combine_gemma4_assistant_mtp` merges under `language_model.mtp.*`. The head only ever had a binding site on the mlx-vlm path.

`detect_model_type` returns `"vlm"` only when it finds a `vision_config`, `vit_config` or `mm_vision_tower`. A checkpoint with no vision tower therefore loads through mlx-lm's `gemma4` / `gemma4_text`, where `mlx_lm_mtp.gemma4_text_model` existed for one purpose: to strip `mtp.*` out of the weights so they would not fail as unexpected keys. Three ordinary things land there — a text-only quant, a DFlash target, and the VLM→LLM fallback.

Nothing reports it. `_is_mtp_compatible` accepts `gemma4`, so the patch is applied and the server logs its backend; then the engine looks for `.mtp`, does not find it, and decodes one token per forward for the life of the model.

## What this does

`gemma4_text.Model.__init__` builds the same `Gemma4AssistantDraftModel` the VLM path uses and binds it to the mlx-lm host. That is cheap because the drafter is not vision-coupled, mlx-lm's `gemma4_text` already implements Gemma 4's cross-layer K/V sharing, and `batch_generator`'s MTP machinery is model-agnostic.

What was missing is the two things the head consumes from a backbone forward, so the patch adds them rather than copying the layer loop:

- **K/V banks per layer type.** Each `DecoderLayer` already returns its `kvs`; wrapping the layer records them into a sink keyed by `layer_type`.
- **The pre-norm hidden.** Swapping the trunk `RMSNorm` for a recorder captures its input. It has to be pre-norm — `_trunk_norm_module` re-applies the trunk norm unless a model sets `_omlx_mtp_head_hidden_normed`, so a post-norm hidden would double-norm the head's input, silently, as a weak acceptance rate.

Then `mtp_partial_rollback` for depth-k. Gemma 4 is attention-only, so unlike the qwen35 sibling there is no recurrent state to restore and replay: every layer drops the rejected count, asked before any is trimmed because a half-rolled-back cache is unrecoverable. And `sanitize` aligns the head's dtype to the backbone's, because a bfloat16 head against float16 activations promotes every head matmul to float32. That is free: over 120 depth-1 steps on one checkpoint loaded twice, float16 and bfloat16 heads accepted the same 32 drafts, at 76ms against 152ms per step.

Gemma 4 alone needs mlx-vlm for the draft head, so that import lives in its own module reached from a gemma4-guarded branch — otherwise `cluster.autoconfigure` asks every rank serving any model to install mlx-vlm.

It is also why `mlx_vlm_mtp/gemma4_vlm_runtime.py` is in the diff: both engines drive the head identically, so `draft_step` is shared rather than copied — 29 fewer lines of code, and one implementation to fix instead of two.

A gemma4 text checkpoint with no merged head, and any checkpoint loaded with MTP off, behaves exactly as before.

## Before and after

M1 Max / 32 GB, mlx 0.32.2, mlx-lm 0.31.3, `gemma-4-31B-it-oQ4-fp16-mtp` (60 layers, hidden 5376, 50 sliding / 10 full, 4-bit g64, float16, 4-layer 1024-wide head).

| | short context | long context (2.7k prompt) |
|---|---|---|
| before | 15.03 tok/s | 14.93 tok/s |
| after, depth 1 | +5.3% | — |
| after, depth 3 | **+28.0%** | **+11.6%** |

Long context accepts 38/41 (92.7%) at 2.67 tokens per cycle with no rollback rejections. The gain is smaller there because a 4-token verify costs more when attention spans 2,700 positions; the head itself is 2.8% of decode (`backbone=3260.8ms mtp=92.6ms`).

Two notes on method: long-context figures difference two token counts on the same prompt, because whole-request timing is mostly prefill and reads only +3.3%; and every figure is a median of within-block ratios from interleaved arms, because on a contended host the unchanged baseline arm moved 15.3 → 12.6 tok/s between blocks.

## Reproducing

Any Gemma 4 checkpoint with the assistant head merged and no vision tower — `text_config.mtp_assistant_config` in `config.json` plus `language_model.mtp.*` in the safetensors index.

```python
from omlx.patches.mlx_lm_mtp import set_mtp_active
from omlx.utils.model_loading import maybe_apply_pre_load_patches

class S: mtp_enabled = True; trust_remote_code = False

set_mtp_active(True)
maybe_apply_pre_load_patches(MODEL, model_settings=S())

from mlx_lm.utils import load
model, _ = load(MODEL)
inner = getattr(model, "language_model", model)
print(getattr(inner, "mtp", None) is not None, hasattr(model, "mtp_forward"))
```

`False False` before, `True True` after — with no warning anywhere in the load. Serving it, the engine's MTP telemetry line is absent before and reports the accept rate after.

## Tests

`tests/test_gemma4_text_mtp_runtime.py` — tiny real modules, no checkpoint, following the VLM-side runtime test. Covers attach gating, both backbone outputs, dtype alignment, `sanitize`'s two branches, and the rejected-tail slicing.

Both engines were also driven live on a merged checkpoint, including a deliberately low-acceptance prompt. At 100% acceptance the rejected-tail slicing never runs at all, so a clean full-accept run proves less than it looks like it does.

The load-bearing one is depth-k rollback across Gemma 4's mixed cache classes: sliding layers get `RotatingKVCache` and full layers `KVCache`, so one verify block rolls back through two mechanisms — a wrapped ring restores its armed undo snapshot and replays, a plain cache moves its offset. Both must land on the same committed length or the layers desynchronize by a position. The parametrized test walks every accept count against a reference cache that only saw the confirmed token plus the accepted drafts.

## One note

`model_loading` sets `set_mtp_depth(8)` for gemma4, citing a 26B code workload. That only ever reached the VLM path; this change makes it reachable for text checkpoints. I could not separate depth 8 from depth 3 here — three interleaved blocks gave 0.995 / 1.188 / 1.043, well inside the noise — so the default stays untouched.

### DIFF
--- a/docs/experimental/dflash_mlx_integration.md
+++ b/docs/experimental/dflash_mlx_integration.md
@@ -118,7 +118,7 @@ are deliberately disabled for Laguna until they have dedicated
 numerical-parity coverage; ordinary adaptive DFlash verification remains
 available.
 
-Note: the `-DFlash` suffix is specific to DFlash draft checkpoints. Gemma4 also ships an `-assistant` variant (e.g. `gemma-4-26B-A4B-it-assistant`) that targets MTP speculative decoding via mlx-vlm — do not mix these in the DFlash toggle. Meta breaks this naming convention: `Muse-Glimmer-30B-assistant` IS a DFlash drafter (block-diffusion, `block_size` 16), not an MTP checkpoint. Drafter routing therefore keys on `config_model_type` (`muse_glimmer_assistant` is in the dashboard's DFlash drafter set), not on the checkpoint name. The Muse Glimmer target is a VLM: DFlash drives its text backbone through dflash-mlx's bundled text-only mlx-lm module, and image requests divert to the VLM fallback engine as usual.
+Note: the `-DFlash` suffix is specific to DFlash draft checkpoints. Gemma4 also ships an `-assistant` variant (e.g. `gemma-4-26B-A4B-it-assistant`) that targets MTP speculative decoding — do not mix these in the DFlash toggle. Meta breaks this naming convention: `Muse-Glimmer-30B-assistant` IS a DFlash drafter (block-diffusion, `block_size` 16), not an MTP checkpoint. Drafter routing therefore keys on `config_model_type` (`muse_glimmer_assistant` is in the dashboard's DFlash drafter set), not on the checkpoint name. The Muse Glimmer target is a VLM: DFlash drives its text backbone through dflash-mlx's bundled text-only mlx-lm module, and image requests divert to the VLM fallback engine as usual.
 
 ### Per-model settings
 
--- a/omlx/patches/gemma4_verify_attention.py
+++ b/omlx/patches/gemma4_verify_attention.py
@@ -1,12 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 """Small-L attention route for gemma4 speculative verify forwards.
 
-Installed on either engine. The routing below depends only on the Attention
-module's own attributes and on a module-level ``scaled_dot_product_attention``,
-both of which mlx-vlm's ``gemma4.language`` and mlx-lm's ``gemma4_text``
-provide under the same names and with the same call signature -- so the two
-engines share one copy of the kernel decision rather than each carrying a
-version that can drift from the other.
+Installed on either engine: the routing depends only on the Attention
+module's own attributes and a module-level
+``scaled_dot_product_attention``, which both provide under those names.
 
 Gemma4 uses head_dim 256 for sliding layers and 512 for global layers. MLX
 0.32.2 now handles small multi-row head-dim-256 calls with its native vector
--- a/omlx/patches/gemma4_verify_attention.py
+++ b/omlx/patches/gemma4_verify_attention.py
@@ -1,6 +1,13 @@
 # SPDX-License-Identifier: Apache-2.0
 """Small-L attention route for gemma4 speculative verify forwards.
 
+Installed on either engine. The routing below depends only on the Attention
+module's own attributes and on a module-level ``scaled_dot_product_attention``,
+both of which mlx-vlm's ``gemma4.language`` and mlx-lm's ``gemma4_text``
+provide under the same names and with the same call signature -- so the two
+engines share one copy of the kernel decision rather than each carrying a
+version that can drift from the other.
+
 Gemma4 uses head_dim 256 for sliding layers and 512 for global layers. MLX
 0.32.2 now handles small multi-row head-dim-256 calls with its native vector
 kernel, so this patch explicitly leaves those shapes alone. Head-dim 512 still
@@ -77,15 +84,39 @@ def apply() -> bool:
     except Exception as e:
         logger.debug(f"mlx_vlm.gemma4 not importable for verify attention: {e}")
         return False
+    return _install(
+        g4_lang.Attention, g4_lang.scaled_dot_product_attention, "mlx-vlm"
+    )
 
+
+def apply_mlx_lm() -> bool:
+    """Wrap ``mlx_lm.models.gemma4_text.Attention.__call__``. Idempotent.
+
+    The same two routes, on the engine that serves text-only checkpoints.
+    Without this the mlx-lm verify forward drops to the unfused pass on
+    head_dim 512 global layers, which is the cost that grows with context and
+    makes a verify cycle lose to plain decoding on low-accept content.
+    """
+    try:
+        from mlx_lm.models import gemma4_text as g4_text
+    except Exception as e:
+        logger.debug(
+            f"mlx_lm.gemma4_text not importable for verify attention: {e}"
+        )
+        return False
+    return _install(
+        g4_text.Attention, g4_text.scaled_dot_product_attention, "mlx-lm"
+    )
+
+
+def _install(cls, sdpa, engine: str) -> bool:
+    """Install the small-L verify routes on one engine's Attention class."""
     from . import gemma4_verify_kernel
 
-    cls = g4_lang.Attention
     if getattr(cls, "_omlx_verify_attn_patched", False):
         return True
 
     original_call = cls.__call__
-    sdpa = g4_lang.scaled_dot_product_attention
 
     def _zero_left_padding(cache) -> bool:
         left = getattr(cache, "left_padding", None)
@@ -200,5 +231,8 @@ def apply() -> bool:
 
     cls.__call__ = __call__
     cls._omlx_verify_attn_patched = True
-    logger.info("gemma4 small-L verify attention patch applied (kernel + per-token)")
+    logger.info(
+        "gemma4 small-L verify attention patch applied on %s (kernel + per-token)",
+        engine,
+    )
     return True
--- a/omlx/patches/mlx_lm_gemma4_assistant.py
+++ b/omlx/patches/mlx_lm_gemma4_assistant.py
@@ -1,26 +1,12 @@
 # SPDX-License-Identifier: Apache-2.0
 """The gemma4 assistant draft step, shared by the mlx-lm and mlx-vlm paths.
 
-Google ships the gemma4 draft head as a separate ``gemma4_assistant``
-model, and the only implementation of it lives in mlx-vlm. Serving a
-merged checkpoint through mlx-lm therefore needs mlx-vlm installed —
-which is true of gemma4 and of nothing else in ``mlx_lm_mtp``.
-
-Which model needs mlx-vlm is load-bearing, not tidiness.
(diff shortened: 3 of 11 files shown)


jundot:
Thanks for putting this together and testing the text-only path. I can see the gap you're fixing: MTP is reported as active even though the assistant head is never attached.

I'm hesitant to add another Gemma 4 MTP runtime path when the VLM path already supports it. Text-only checkpoints do have a use case, but the benchmarks compare against MTP-off on mlx-lm, so they don't yet establish a benefit over the existing VLM MTP path that would justify maintaining the extra hooks and rollback logic.

I also reproduced a rollback issue with KV-sharing configurations such as E2B/E4B: `make_cache()` returns fewer entries than the number of layers, so the new length check rejects a valid rollback and triggers a cache rebuild followed by standard decoding.

My concern is that maintaining a separate MTP path for text-only checkpoints adds complexity that may not be necessary, given the existing VLM support. What do you think?

andyoneal:
I definitely agree it's a lot of new code for a fairly niche use-case. If text-only checkpoints load through the mlx-vlm path, then that's probably the better way to go. I'll work on this more and check out the E2B/E4B problem. I should have tested all the Gemma 4 models before submitting.

andyoneal:
Thanks — both of your points were right, and the second one changed what I think this PR should be.

**The E2B/E4B rollback bug is fixed.** `make_cache()` returns one entry per distinct K/V owner, not one per layer, so the length check refused every rollback on those checkpoints and fell through to a cache rebuild. mlx-lm materializes the sharing map as `previous_kvs`, whose distinct values are exactly the caches that exist, so the guard now counts owners and keeps the per-layer count for a backbone that doesn't publish one. Three regression tests cover it, and they fail against the old guard.

**On the benchmark: you were right that it didn't establish anything.** Comparing MTP-off to MTP-on on one engine shows the head works, not that the path earns its maintenance. So I built the thing your comment implied — teaching mlx-vlm to load a text-only checkpoint — and then benchmarked the two implementations against each other *and* against MTP off, on one checkpoint, on one host.

## Method

`gemma-4-31B-it-oQ4-fp16-mtp` — 19.0 GB, `model_type` gemma4, no vision sub-config, merged head present. M1 Max / 32 GB. Driven through the admin benchmark, so the numbers are reproducible from the UI without being handed a configuration: bundled `code_python` corpus, `temperature=0.0`, `warmup_mode=quick`.

Four arm types, because "which implementation" and "does the feature pay" are different questions and only the second one is what the PR claims:

| | MTP on | MTP off |
|---|---|---|
| mlx-lm (this PR) | A | a |
| mlx-vlm (the alternative) | B | b |

Arms are interleaved rather than blocked, each type spread across the session, because this host throttles under sustained load. Every arm is a fresh server on a freshly checked-out tree. Each arm is gated three ways before it counts: `completion_tokens` must equal what was asked for, discovery must report the engine the branch intends, and an MTP-on arm must show drafting in its log. Medians are `statistics.median`.

Neither branch needs a route override to produce its arm. A text-only gemma4 with a merged head reaches mlx-lm on this branch by ordinary discovery, and reaches the VLM engine on the other branch by its own routing rule, so the route is a property of the branch under test rather than a knob set for the test.

## Results

**1024-token prompt, 512 generated, n=4 per arm:**

| arm | median | mean | range |
|---|---|---|---|
| mlx-lm, MTP on | 19.90 | 19.72 | 18.9–20.2 |
| mlx-lm, MTP off | 15.00 | 14.80 | 14.1–15.1 |
| mlx-vlm, MTP on | 17.65 | 17.50 | 14.5–20.2 |
| mlx-vlm, MTP off | 14.30 | 14.22 | 13.2–15.1 |

MTP pays here: **+32.7%** on mlx-lm (mean +33.3%), **+23.4%** on mlx-vlm (mean +23.0%). That brackets the +28% this PR originally claimed.

## The two implementations

| | MTP on | MTP off |
|---|---|---|
| mlx-vlm / mlx-lm | **0.887** median, 0.887 mean | 1.007 median, 0.998 mean |

With drafting on, the mlx-lm path is **12.7% faster**. With drafting off the two engines are level — 1.007 over six arms per side, and 1.000 at a second prompt size. So none of the difference is engine overhead; it is all in how the head is driven.

Decomposing it on the unparked arms: acceptance is the same (94.7% vs 94.4%), the head's own forward is a rounding error at this generation length (~2.0 s of a ~26 s request), and what remains is per-cycle backbone cost — 174.2 ms/cycle at 3.83 tokens per cycle against 180.0 ms at 3.71, or 45.5 against 48.5 ms per token. One mlx-vlm arm parked and is included in the medians; excluding it the gap is ~6%, but parking is what a user gets, so I've left it in.

Memory is identical: 21.07 GB peak with MTP on either side, 19.65 GB with it off. MTP itself costs 1.4 GB, the same on both engines.

## A bug in the shipping VLM path, unrelated to which branch you pick

The assistant head ships bfloat16 and `oq.combine_gemma4_assistant_mtp` preserves that, so on a float16 checkpoint every head matmul meets float16 activations and MLX promotes the result to float32. The 31B oQ4 quant is exactly that shape — 1243 float16 backbone tensors against 48 bfloat16 head tensors, visible only in the safetensors headers, since the config reports bfloat16 for the whole file.

This PR fixes it on the mlx-lm side. The mlx-vlm path never had it, and was paying 2355–2744 ms of head forward against 509–787 ms for the same head driven from mlx-lm. Before I fixed that, the alternative looked 3–5× worse on head cost and ~1.5 GB heavier; afterwards both differences essentially vanish. Everyone running gemma4 MTP through mlx-vlm today is paying it, so it's worth taking regardless of what happens to this PR — it's a separate commit on the alternative branch and I'm happy to split it out.

## Where that leaves it

I've opened the alternative as a draft PR — #3561 — so you can compare them directly. It is **536 lines against 1356** by the diffstat, or 276 against 519 counting only `omlx/`, it reuses the head-driving code that already ships, and it measures ~13% slower with drafting on and identical without.

I don't have a strong view on which you should take, and I'd rather you decide than have me argue for the one I wrote first. The trade as measured is ~13% throughput against ~900 lines and a second runtime path. If the answer is that the mlx-vlm route is the better place for this, I'm glad to close this PR and carry the rollback fix and the dtype fix over to that one.

One thing I'd flag either way: the depth controller will park profitable speculation while accepting 71% of its drafts — I saw it drop to depth 0 on 18 of 27 cycles, then run the same content at 100% acceptance after the forced re-entry probe. I can't prove from that data that keeping it on would have been faster, since the parked stretch was also the cold one, but the profitability test may not be accounting for expected remaining generation length.

Raw per-arm data, logs and the harness are available if you want to check any of it.

andyoneal:
> I'm hesitant to add another Gemma 4 MTP runtime path when the VLM path already supports it. Text-only checkpoints do have a use case, but the benchmarks compare against MTP-off on mlx-lm, so they don't yet establish a benefit over the existing VLM MTP path that would justify maintaining the extra hooks and rollback logic.

Just to be clear about this, the mlx-vlm path does not support text-only checkpoints for gemma 4 either. So the choice was either add MTP to mlx-lm or make text-only run on mlx-vlm. The two PRs I opened cover each of those routes. mlx-lm is faster, but more code, but also less monkey patches, than the mlx-vlm implementation.

andyoneal:
Rebased both this PR and #3561 onto current main (mlx-lm 0.32.0, mlx-vlm 0.7.1). Running them end to end turned up a bug main has too: the Gemma 4 assistant head was bound to `embed_tokens` before `nn.quantize()` replaced it, so `materialize_lazy_state` built the discarded float embedding, 5.6 GB on the 31B, and on a 32 GB Mac the prefill guard rejected every prompt. Both PRs now carry the same small fix commit with a regression test.

The comparison above still holds: neither arm carried that memory (both peaked at 21.07 GB), and the benchmark sends token ids, so chat templates never applied. Two corrections to it: the diffs are now 1388 added lines here against 276 in #3561 (519 against 103 in `omlx/`), and over chat the paths differ in one visible way. mlx-lm defaults `enable_thinking` on for Gemma 4, while mlx-vlm keeps the template's default of off.

jundot:
@andyoneal I merged #3561 to support text-only checkpoints through the existing VLM MTP runtime. I prefer this approach because it avoids maintaining a separate MTP runtime for the mlx-lm path. Closing this in favor of #3561.

PR #3541, opened 2026-09-09
refactor(prefill): share chunk and snapshot boundary rules

chenqianhe:
Extract the existing block-boundary chunk clamp and snapshot eligibility rules into pure helpers shared by external and resumable prefill. Preserve execution order, memory guards, and snapshot bookkeeping without introducing batching or configuration changes.

Add unit and characterization coverage for aligned and unaligned prefixes, reserved final tokens, disabled boundaries, guard-limited chunks, and duplicate snapshot prevention. The focused regression suite passes 815 tests with 3 deselected.

### DIFF
--- a/omlx/prefill_boundaries.py
+++ b/omlx/prefill_boundaries.py
@@ -1,0 +1,27 @@
+"""Shared block-boundary rules for external and resumable prefill."""
+
+
+def clamp_prefill_chunk_to_boundary(
+    chunk_tokens: int, *, cache_tokens: int, block_size: int
+) -> int:
+    """Cap a non-empty chunk at the next cache block boundary.
+
+    ``cache_tokens`` includes any reused prefix; ``block_size`` must be
+    positive. Callers handle empty input and apply memory guards afterward.
+    """
+    next_boundary = ((cache_tokens // block_size) + 1) * block_size
+    return max(1, min(chunk_tokens, next_boundary - cache_tokens))
+
+
+def should_emit_prefill_boundary(
+    *, total_tokens: int, block_size: int, last_emitted_tokens: int
+) -> bool:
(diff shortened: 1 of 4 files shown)


chenqianhe:
I’m experimenting with batched prefill support for oMLX. During that work, I noticed that the existing external and resumable prefill paths duplicate the same chunk-boundary calculation and snapshot eligibility checks.

Before introducing batching-specific changes, I wanted to consolidate these existing rules into small, shared helpers and add regression coverage. This PR is intentionally limited to a behavior-preserving refactor: it does **not** introduce batched prefill, change scheduling policy, or add configuration options.

Any batched prefill implementation would be proposed separately. This refactor has independent value by keeping the current prefill paths consistent and making their boundary behavior easier to test and maintain.

chenqianhe:
@jundot Hi! The Python 3.12 job failed in the existing `test_split_restore_retry_budget_is_one_per_block` test, while Python 3.11, Python 3.13, and all 51 new tests passed. I reproduced the same failure on unchanged `main` (`94530d8d`) using controlled thread timing: promoting one GDN snapshot can remove its staging directory while the background writer is preparing the next snapshot, causing the cache to retain only 8 tokens instead of the expected 12. This points to a pre-existing snapshot-storage race in code this PR does not change. 

Could you please rerun the failed job, or consider this failure non-blocking for this PR? I would prefer to address the race separately to keep this PR limited to the behavior-preserving boundary refactor.

jundot:
Thanks for consolidating these rules. I checked that both prefill paths preserve chunk boundaries, memory guard ordering, and snapshot bookkeeping. This keeps the two paths consistent without changing execution policy. LGTM, merging.

Issue #3507, opened 2026-09-07
/v1/embeddings answers 200 with null-filled vectors when the model returns NaN (ModernBERT via mlx-embeddings, padded batches whose longest item is a multiple of 32 ≥ 128)


builtbycnob:
**Symptom.** A mixed-length batch through a ModernBERT embedding model (`nomic-ai/modernbert-embed-base` here, served through the bundled mlx-embeddings path because `_load_native` covers XLM-RoBERTa, BERT and Qwen2 only) returns an all-`null` vector for every padded item whenever the longest item tokenizes to an exact multiple of 32 at or above 128. The longest item is correct, equal-length batches are correct, single texts are always correct. The response is HTTP 200 and the server log line is normal, so a RAG pipeline stores corrupt vectors without noticing. Measured toxic lengths: 128, 224, 256, 288, 320, 512, 640; clean: 32, 64, 255, 257, 259, 289. Reproduced on 0.4.4 offline (`MLXEmbeddingModel.embed`) and on 0.6.3 through the server (30 out of 30).

**Root cause** is upstream and I filed it there with a standalone repro: Blaizzy/mlx-embeddings#80. `modernbert.py` builds its additive masks with `-1e9` and casts them back to the tokenizer's int32 mask dtype (`:334-337`, `:367-369`); `mx.fast.scaled_dot_product_attention` promotes that to fp16, where -1e9 overflows to -inf, and a sliding-window (128) query row that lies entirely in padding gets an all -inf score row, softmax NaN, and `0 × NaN` spreads it through the padded sequence. oMLX pins `mlx-embeddings@32981fa` on `main`, v0.6.3 and v0.6.4, and that revision carries the defect.

**What oMLX could do independently of the upstream fix:**

1. check the embedding tensor for non-finite values before serialising and answer with an error (a 500 naming the item indices) rather than a 200 whose `embedding` arrays contain `null`; there is no such check on the response path today;
2. bump the pin when #80 lands.

Workaround meanwhile: send ModernBERT inputs one per request, or bucket equal-length inputs. bge-m3 and the other native-path models are not affected.

builtbycnob:
Fixed on main by 25bccf5f7 (`omlx/patches/modernbert_attention.py`, with a regression test in `tests/test_modernbert_attention.py`), which is in 0.7.0.dev2. Closing. Thanks for the fix.

PR #3545, opened 2026-09-09
fix(server): reject /v1/embeddings responses that contain non-finite vectors

L4XB:
Refs #3507 (the first of the two things oMLX can do independently of the upstream mlx-embeddings fix).

## Problem

When the embedding model returns NaN/Inf values (ModernBERT through the pinned `mlx-embeddings` revision on padded batches whose longest item is a multiple of 32 ≥ 128, per the issue), `/v1/embeddings` answered **HTTP 200** with the vectors serialized as `null`-filled arrays. JSON has no NaN, so the client saw a normal-looking response and a RAG pipeline stored corrupt vectors without noticing.

## Change

- `omlx/api/embedding_utils.py`: new `find_non_finite_embeddings(embeddings) -> list[int]` returning the indices of vectors that are not entirely finite (a row that cannot be converted to floats counts as bad).
- `omlx/server.py` (`create_embeddings._build_embeddings`): after `engine.embed(...)`, if any vector is non-finite the request fails with a **500** whose detail names the affected item indices and suggests sending those inputs one per request or batching inputs of equal length. The failure is logged with the model id and the indices. Finite results are unchanged.

Bumping the `mlx-embeddings` pin once Blaizzy/mlx-embeddings#80 lands is left for a separate change.

## Tests

`tests/test_embeddings_non_finite.py`:
- unit tests for `find_non_finite_embeddings` (NaN, +Inf, -Inf rows, array-like rows, empty input);
- endpoint test through `TestClient` with a mocked embedding engine: a batch with one NaN vector returns 500 and the OpenAI-style error message names item `[1]`; an all-finite batch still returns 200 with the vectors. The endpoint test fails on `main` (it gets a 200 with `null` entries) and passes with the change.

`pytest tests/test_embeddings_non_finite.py` (4 passed), `ruff check` on the changed files; `black --diff` reports nothing in the new code (the remaining diff in `embedding_utils.py` / `server.py` is pre-existing formatting).

### DIFF
--- a/omlx/api/embedding_utils.py
+++ b/omlx/api/embedding_utils.py
@@ -11,7 +11,7 @@ Provides:
 import base64
 import math
 import struct
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List, Sequence, Union
 
 from .embedding_models import EmbeddingInputItem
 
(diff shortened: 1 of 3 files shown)


jundot:
Thanks for adding this check. I verified that NaN/Inf results are rejected and finite responses remain unchanged. I’ll handle one gap in a follow-up: when embedding takes longer than the keepalive grace period, the HTTP status is already 200, so the error needs to be returned as JSON instead of interrupting the response. LGTM with that follow-up; I’ll merge this.

PR #3546, opened 2026-09-09
fix(cli): keep the configured Claude tiers when the launcher picks a model interactively

### LINKED ISSUE #3543: omlx launch claude ignores configured Opus/Sonnet/Haiku tiers after interactive model selection
**Describe the bug**

When Claude Code is configured for local routing with three distinct tier models, `omlx launch claude` still prompts for one model. After that interactive choice, it maps that single model to the Opus, Sonnet, and Haiku roles, ignoring the tier mappings saved in oMLX.

This makes the Claude Code configuration page misleading: the three persisted model selections are not honored on the default interactive launch path.

**Configuration**

`~/.omlx/settings.json`:
```json
"claude_code": {
  "mode": "local",
  "opus_model": "Qwen3.8-27B-oQ8e-mtp",
  "sonnet_model": "Qwen3.8-27B-oQ6e-mtp",
  "haiku_model": "Qwen3-Coder-Next-MLX-6bit"
}
```

**To reproduce**

1. Configure three different models under Claude Code in the oMLX settings, with `mode: "local"`.
2. Start the local oMLX server.
3. Run `omlx launch claude` without `--model` or explicit `--opus-model`, `--sonnet-model`, and `--haiku-model` flags.
4. Choose any model in the interactive picker.
5. Inspect the resulting Claude Code environment or the models requested by Claude Code.

**Actual behavior**

The selected model is used for every tier:

- `ANTHROPIC_DEFAULT_OPUS_MODEL`
- `ANTHROPIC_DEFAULT_SONNET_MODEL`
- `ANTHROPIC_DEFAULT_HAIKU_MODEL`

The configured tier models are discarded.

**Expected behavior**

The launcher's interactive picker should select the initial/default model only, while preserving the persisted `claude_code.opus_model`, `sonnet_model`, and `haiku_model` mappings for their respective roles. Explicit tier flags should continue to override the saved settings.

**Root cause**

In `omlx/cli.py`, `launch_command()` first reads the configured tier values, but when no explicit tier flags were supplied it deliberately clears all three after the interactive picker:

```python
if args.model is None and not (
    cli_opus_model or cli_sonnet_model or cli_haiku_model
):
    opus_model = None
    sonnet_model = None
    haiku_model = None
```

Then `ClaudeCodeIntegration.launch()` falls back to `ctx.model` for every tier, producing the single-model mapping.

**Related work**

- #938 / #1009 handle server-side translation of standard `claude-*` wire IDs to configured local tiers. They do not fix this launcher precedence issue: the launcher has already replaced all three tiers with the picker choice before requests reach the server.
- #2716 also documents that the launcher maps a chosen local model to the Opus/Sonnet/Haiku tiers, but concerns the resulting context-window behavior.

**Suggested fix**

Do not clear settings-derived tier values after an interactive model selection. Keep the existing precedence:

`explicit tier flag > saved claude_code tier setting > interactively selected model`.

This preserves normal cloud behavior because the setting's default mode remains `cloud`, while making local launch match the configured per-tier mappings.

**Desktop**

- macOS: 26.x
- oMLX: 0.6.4
- Claude Code: installed locally
- Apple Silicon

L4XB:
Fixes #3543

## Problem

With `claude_code.mode = "local"` and three tier models saved on the Claude Code settings page, `omlx launch claude` (no `--model`, no tier flags) shows the picker and then uses the picked model for the Opus, Sonnet and Haiku roles: `launch_command()` read the saved tiers first and then deliberately cleared all three after the interactive selection, so `ClaudeCodeIntegration.launch()` fell back to `ctx.model` for every role. The persisted per-tier selections were never honored on the default interactive launch path, which makes the settings page misleading.

## Change

- `omlx/cli.py`: the tier values are no longer cleared after an interactive pick. Precedence is now `explicit tier flag > saved claude_code tier setting > picked model`; the picker chooses the default model only, and a role without a saved model still falls back to the picked model in the integration (unchanged).
- Tests (`tests/test_cli.py`): new `TestLaunchClaudeTierPrecedence` covers the interactive pick with saved tiers (fails on `main`), an explicit tier flag overriding a saved tier, and the no-saved-tiers case. The existing `test_launch_command_shows_picker_and_clears_saved_tiers` encoded the old behavior; it is renamed to `..._keeps_saved_tiers` with the assertions updated.

## Note for reviewers

That existing test shows the clearing was intentional ("shows the picker and ignores saved tier models"), so this is a behavior change rather than a plain bug fix. The argument from #3543 is that the picker is about the default model, while the three saved tiers are configured explicitly for their roles and should win over an incidental picker choice; explicit flags still override everything. If you would rather keep the old semantics (for example, only when the picked model is not one of the saved tiers), I am happy to adjust.

`pytest tests/test_cli.py`: 72 passed; `black` on the test file; `ruff check` reports only pre-existing findings.

### DIFF
--- a/omlx/cli.py
+++ b/omlx/cli.py
@@ -497,15 +497,12 @@ def launch_command(args, extra_args: list[str] | None = None):
         print(f"Install: {integration.install_hint}")
         sys.exit(1)
 
-    # If the model was chosen interactively (no --model and no explicit tier flags),
-    # use the picked model for all tiers instead of letting settings-based tier
-    # models override the user's selection.
-    if args.model is None and not (
-        cli_opus_model or cli_sonnet_model or cli_haiku_model
-    ):
-        opus_model = None
-        sonnet_model = None
-        haiku_model = None
+    # Tier precedence: explicit tier flag > saved claude_code tier setting >
+    # the model picked (or auto-selected) above. The picker only chooses the
+    # default model; tiers configured on the Claude Code settings page keep
+    # their role, otherwise the three persisted selections would be silently
+    # replaced by one model on every interactive launch (#3543). Roles without
+    # a saved model fall back to the picked model in the integration.
 
     # Enforce Claude Code's model requirements after all interactive,
     # automatic, and explicit model paths have resolved. The picker also marks
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -753,8 +753,8 @@ class TestLaunchCommandFunction:
         assert "Using model: only-32k" in output
         assert "Cannot launch Claude Code with model 'only-32k'" in output
 
-    def test_launch_command_shows_picker_and_clears_saved_tiers(self):
-        """Bare `omlx launch claude` shows the picker and ignores saved tier models."""
+    def test_launch_command_shows_picker_and_keeps_saved_tiers(self):
+        """Bare `omlx launch claude` shows the picker for the default model and keeps the saved tier models (#3543)."""
         from omlx.cli import launch_command
 
         integration = MagicMock()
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for fixing this. Keeping the saved tier mappings makes the launcher consistent with the settings page. I’ll merge this and follow up on the initial model and context limits: the picker’s model is not explicitly passed as the initial model, but its context window still applies even when the saved tiers use smaller models.

PR #3547, opened 2026-09-10
fix(memory): schedule a retry when emergency abort/unload does not drain

### LINKED ISSUE #3363: Emergency memory-pressure abort/unload leaves a permanent unload-pending latch when pressure clears in the same tick — every request 409s until an admin settings write
**Describe the bug**
When process memory spikes past an emergency threshold (empirically between 58.4 and 59.0 GB on this 64 GB machine — see the control-group section for the bound), the enforcer skips the usual `deferring destructive enforcement (poll 1/5)` grace ladder and immediately aborts the active requests and requests a model unload in the same tick.

Aborting the requests freed ~13 GB instantly, so **149 ms later** pressure was back to `ok` — but the unload never executed (no `Unloading model` line was ever emitted by the enforcer) while the *unload-pending* flag stayed set. From then on every `POST /v1/chat/completions` returned:

```
HTTP 409: Model 'Qwen3.8-27B-oQ8e-fp16-mtp' is busy; cannot start work while unload is pending until active requests finish or are aborted
```

False on both counts: both aborted requests had terminated within 66 ms, and memory was healthy (45.9 GB vs 57.0 GB ceiling). The state never self-recovered. It persisted 7 minutes until an unrelated cache-settings change in the admin UI force-unloaded the model via `omlx.admin.routes` and cleared the latch. A one-off memory spike that should have cost two requests took the server down until a human intervened — and unlike #2852/#2858, the server answers instantly and truthfully reports its own false state, so no liveness probe can catch it.

**To Reproduce**
Not deterministic — it requires a spike past the emergency threshold between two enforcer polls:
1. 64 GB Mac, guard tier `custom`, ceiling 57.0 GB (soft 51.3 / hard 54.1), hot cache 5 GB → dynamic ceiling 52.0 GB, `iogpu.wired_limit_mb=59392` (Metal cap 58.0 GB).
2. Serve `Qwen3.8-27B-oQ8e-fp16-mtp` (qwen3_5 hybrid GDN, ~33 GB resident, Lightning MTP on, chunked prefill, `max_concurrent_requests: 1`).
3. Drive with 70K–99K-token agent prompts so footprint oscillates around the hard watermark. ArraysCache-hybrid in-flight growth exceeds the admission estimate (#3284) until one decode+prefill jumps from ≤54 GB to **59.0 GB within one poll**. Spikes up to 58.4 GB still take the grace path; 59.0 GB triggered the emergency branch.
4. Enforcer aborts + requests unload. Requests die. Pressure returns to `ok` within the same second.
5. Every subsequent request → HTTP 409 "unload is pending", indefinitely. Recover only via an admin write that unloads models.

**Expected behavior**
Either the requested unload actually completes and the model reloads on the next request (as in #2841, where `Unloading model ... (immediate abort)` follows the enforcer's decision) — or, once pressure is `ok` and the aborted requests have drained, the pending-unload latch is cancelled. Not the current outcome: unload requested, never performed, never cancelled.

**Screenshots**
N/A — log excerpts below.

**Desktop:**
 - macOS Version: 26.6.2
 - oMLX Version: 0.6.4 (.app, installed 2026-08-30, incident 2026-09-01)
 - Hardware: MacBook Pro, Apple M1 Max, 64 GB, `iogpu.wired_limit_mb=59392`
 - Model: `Qwen3.8-27B-oQ8e-fp16-mtp` (8-bit oQe, fp16 non-quant dtype, Lightning MTP, 262K configured context), SSD cache 200 GB, hot cache 5 GB at incident time

**Additional context**

*The incident (2026-09-01):*

```
23:26:34,728  process_memory_enforcer  Memory pressure level: soft -> hard (current=59.0GB, soft=51.3GB, hard=54.1GB, ceiling=57.0GB)
23:26:34,730  process_memory_enforcer  Requested pressure cache reclaim on 1 scheduler(s) (freed_hot=0.0GB, pool=14.3GB)
23:26:34,730  engine_core              Aborted 2 requests due to memory pressure
23:26:34,730  process_memory_enforcer  Hard memory pressure: requested abort/unload for 'Qwen3.8-27B-oQ8e-fp16-mtp' (aborted=2)
23:26:34,731  server                   JSON keepalive prefill rejected: Request aborted: process memory limit exceeded (usage 59.0 GB, ...)
23:26:34,731  server                   chat streaming prefill rejected: Request aborted: process memory limit exceeded (usage 59.0 GB, ...)
23:26:34,796  mlx_lm_mtp               MTP[459] finish=filter tokens=701 ...        <- last active request terminated
23:26:34,878  server                   POST /v1/chat/completions -> 409 "busy; unload is pending"
23:26:34,879  process_memory_enforcer  Memory pressure level: hard -> ok (current=45.9GB)   <- 149 ms after trigger
23:26:40 / 23:26:41 / 23:26:41         three more 409s
23:28:16,369  server                   POST /v1/chat/completions -> 409           <- 102 s later, still latched, server idle
23:33:37,549  admin.routes             Hot cache max size changed: 5.00 GB -> 4.00 GB    <- manual intervention
23:33:37,550  engine_pool              Unloading model: Qwen3.8-27B-oQ8e-fp16-mtp (immediate abort)   <- FIRST unload line in the log
23:33:38,099  engine_pool              Unloaded model: ..., freed=33.08GB, active_memory: 1.02KB (settled)
```

`grep 'Unloading model' server.log` returns exactly one line: the admin-triggered one at 23:33:37.

*Built-in control group — 112 hard-pressure events across two logs, only the 59.0 GB one latched:*
Day 1 (2026-09-01): 23 `-> hard` transitions. 22 (measured 54.2–58.3 GB) logged `deferring destructive enforcement (poll 1/5)` and self-recovered; one (22:44:44) exhausted all 5 polls and correctly shrank the hot cache instead — the ladder works when it runs. Only the 59.0 GB event skipped the deferral entirely and went straight to abort/unload.
Night 2 (2026-09-02, 00:06–02:26, same config): 89 more `-> hard` transitions under a heavier workload (91–99K prompts), peaks up to **58.4 GB** — every one deferred and self-recovered, zero aborts.

So the emergency branch is not gated on the configured ceiling (57.0 GB) or the Metal cap (58.0 GB): deferred events up to 58.4 GB exceeded both. The trigger lies in the narrow band **(58.4, 59.0] GB** — plausibly an absolute fraction of physical memory (92% of 64 GB = 58.9 GB) or ceiling + fixed margin. Whatever it is, it is isolated 111-to-1 across the two logs: grace path every time below it, permanent latch the one time above it.

*Hypothesis (from #3338's code read of 0.6.3rc2. 0.6.4 notes list no enforcer changes):*
Per #3338, the reclaim/enforcement machinery lives inside `if new_level == "hard":` (line 1477), closed by `if new_level == "ok":` (1542), with an emergency bypass alongside the grace budget. If the unload follow-through also lives on the hard path, this explains the incident exactly: the emergency branch set the pending flag and aborted the requests. The aborts freed enough memory that the next poll read `ok`. The hard branch never ran again. Nothing else completes or cancels the pending unload. Happy to be corrected on the mechanism — the observable facts stand either way.

*Related issues (checked, none is a duplicate):*
- #2841 — same enforcer/model family, but the unload **completes** there, here it never runs. The contrast is the clearest evidence the unload half silently failed.
- #3338 — soft-path reclaim gap, source of the code map. My 111 recovered events confirm its hard-path description.
- #2852 / #2624 / #2858 — wedges where requests hang and health lies green. Different failure: here the 409 is instant and self-consistent, the *state* is false — a third mode for the #2858 health-surface discussion.
- #2940 — precedent: an unrecoverable failure latch was split out and fixed (#2946) independently of its trigger. Requesting the same treatment.
- #3284 — likely cause of the memory overshoot itself, deliberately out of scope, this report is only about the latch.

*Available on request:*
Both full `server.log` files (Day 1: 1,397 lines; Night 2: 1,199 lines; no prompt content), client-side gateway logs of the 409s, `settings.json`.

AmirF194:
When `_check_and_enforce`'s emergency branch aborts the last busy non-pinned model, it marks `pending_unload_reason` and makes one inline `_unload_pending_if_idle_locked` call. If the abort hasn't drained by then, that call returns `False` and nothing retries it: the `while` loop it sits in is never reached again once pressure recovers to "ok" (an earlier branch returns before that point on the next tick). The latch stays set forever, and every later request against that model raises `ModelBusyError`.

The manual unload path (`EnginePool.request_unload`) hits the same "not idle yet" case and falls back to `_schedule_pending_unload_locked`, which spawns `_wait_for_pending_unload` to poll until the entry drains. The emergency path never had that fallback. Added it: schedule the poller when the inline attempt doesn't unload immediately.

Added `test_single_busy_model_schedules_retry_when_still_busy_after_abort`, mirroring the existing `test_single_busy_model_pending_unload_at_emergency_pressure` fixture: it asserts `_schedule_pending_unload_locked` gets called once the busy victim is still busy after the abort. Without the fix it's never called.

Couldn't run the real suite here, `mlx` needs Apple Silicon. Ran `tests/test_process_memory_enforcer.py` (165 passed) against a build with `mlx`/`mlx_lm`/`mlx_vlm` and `omlx.engine` stubbed down to the two symbols this file actually touches (`BaseNonStreamingEngine`, `TTSEngine`), so it's this file's own logic under test, not the neighboring stubs.

Fixes #3363

### DIFF
--- a/omlx/process_memory_enforcer.py
+++ b/omlx/process_memory_enforcer.py
@@ -1639,9 +1639,17 @@ class ProcessMemoryEnforcer:
                                 "hard memory pressure",
                                 abort_requested=True,
                             )
-                            await self._engine_pool._unload_pending_if_idle_locked(
-                                busy_victim
+                            unloaded = (
+                                await self._engine_pool._unload_pending_if_idle_locked(
+                                    busy_victim
+                                )
                             )
+                            if not unloaded:
+                                # Not drained yet: schedule the poller, mirroring
+                                # request_unload's own fallback, or the latch never clears.
+                                self._engine_pool._schedule_pending_unload_locked(
+                                    busy_victim
+                                )
                         logger.warning(
                             "Hard memory pressure: requested abort/unload for "
                             "'%s' (aborted=%d)",
--- a/tests/test_process_memory_enforcer.py
+++ b/tests/test_process_memory_enforcer.py
@@ -229,6 +229,10 @@ def mock_engine_pool():
     pool._mark_pending_unload_locked = MagicMock(
         side_effect=_mark_pending_unload_locked
     )
+    pool._scheduled_pending_unloads = []
+    pool._schedule_pending_unload_locked = MagicMock(
+        side_effect=pool._scheduled_pending_unloads.append
+    )
     return pool
 
 
@@ -1523,6 +1527,36 @@ class TestSingleModelMemoryPressure:
         entry.in_use = 0
         await enforcer._engine_pool._unload_pending_if_idle_locked("big-model")
         enforcer._engine_pool._unload_engine.assert_awaited_once_with("big-model")
+
+    @pytest.mark.asyncio
+    async def test_single_busy_model_schedules_retry_when_still_busy_after_abort(
+        self, enforcer
+    ):
+        """The pending-unload latch must not depend on a later manual drain check.
+
+        _check_and_enforce()'s while loop is not revisited once pressure recovers
+        to "ok" (an earlier branch returns before reaching it), so if the busy
+        victim has not drained by the time this tick's single inline
+        _unload_pending_if_idle_locked call runs, nothing else will ever retry it.
+        """
+        engine = MagicMock()
+        engine.has_active_requests.return_value = False
+        engine.abort_all_requests = AsyncMock(return_value=3)
+        entry = _make_entry("big-model", engine=engine)
+        entry.in_use = 1
+        enforcer._engine_pool._entries = {"big-model": entry}
+        enforcer._engine_pool._find_lru_victim.return_value = None
+
+        with patch("omlx.process_memory_enforcer.mx") as mock_mx:
+            mock_mx.get_active_memory.side_effect = _cycling(
+                [13 * 1024**3, 13 * 1024**3]
+            )
+            await enforcer._check_and_enforce()
+
+        assert entry.pending_unload_reason == "hard memory pressure"
+        enforcer._engine_pool._schedule_pending_unload_locked.assert_called_once_with(
+            "big-model"
+        )
 
     @pytest.mark.asyncio
     async def test_two_models_one_inferring_evicts_idle(self, enforcer):


jundot:
Thanks for the fix. I reproduced the stuck pending-unload state after pressure recovered and verified that the existing poller completes the unload once activity drains. LGTM, merging.

AmirF194:
Thanks for reproducing it and confirming the poller drains once things go idle.

PR #3549, opened 2026-09-10
fix(vlm): derive Gemma4 global KV head fields from per_layer_config

### LINKED ISSUE #3537: Gemma4 loading fails when KV head overrides are stored in per_layer_config
### Environment

- oMLX: 0.6.4 (`e008a66`)
- Pinned mlx-vlm: `78b96eb5462141447b9a6b4943ef553891da56dd`
- Model: Gemma 4 26B A4B, locally quantized to oQ4e
- Source config records Transformers 5.15.0

### Problem

Loading the model fails with:

```text
Expected (4096, 440)
received (1024, 440)
```

The source checkpoint stores full-attention overrides in `text_config.per_layer_config` rather than the legacy global fields:

```json
"head_dim": 256,
"num_key_value_heads": 8,
"per_layer_config": {
  "05": {"head_dim": 512, "num_key_value_heads": 2},
  "11": {"head_dim": 512, "num_key_value_heads": 2},
  "17": {"head_dim": 512, "num_key_value_heads": 2},
  "23": {"head_dim": 512, "num_key_value_heads": 2},
  "29": {"head_dim": 512, "num_key_value_heads": 2}
}
```

The pinned mlx-vlm Gemma4 loader ignores `per_layer_config`. For full-attention layers, it defaults to `global_head_dim=512` but falls back to `num_key_value_heads=8`, producing **4096** outputs instead of the required **1024**.

This config format is already present in the source checkpoint, before quantization.

### Confirmed workaround

Adding these fields to `text_config` resolves the loading error, and the model runs successfully without re-quantization:

```json
"global_head_dim": 512,
"num_global_key_value_heads": 2
```

`per_layer_config` can remain in place.

### Expected behavior

The Gemma4 loader should honor `per_layer_config` when constructing each layer, including its head dimension and KV head count.

L4XB:
## Summary

Fixes #3537.

Gemma4 checkpoints written by Transformers ≥ 5.15 record the `head_dim` / `num_key_value_heads` overrides of the full-attention layers under `text_config.per_layer_config` instead of the legacy `global_head_dim` / `num_global_key_value_heads` fields. The pinned mlx-vlm Gemma4 loader reads only the legacy fields, so it sizes the full-attention K/V projections with the sliding-window head count and `load_weights` fails with `Expected (4096, 440) received (1024, 440)`.

## Change

`omlx/engine/vlm.py`

- `_gemma4_global_kv_from_per_layer_config(config)` derives the legacy fields that are missing from `text_config` when every overridden full-attention layer agrees (`layer_types` decides which layers count; without it every override counts). Explicit legacy fields always win; non-Gemma4 configs, missing `per_layer_config` and disagreeing overrides yield nothing.
- `_derive_gemma4_global_kv_on_load(model_dir)` follows the existing `*_on_load` pattern: for one `vlm_load(...)` it wraps `mlx_vlm.utils.load_config` and `setdefault`s the derived fields into `text_config`, so the pinned loader's `TextConfig` sees `global_head_dim=512` / `num_global_key_value_heads=2` for the checkpoint from the issue. The source layout (`per_layer_config`) stays in place; checkpoints that already carry the legacy fields are untouched.
- Added to the VLM load context stack next to `_drop_gemma4_mlx_shared_kv_extras_on_load`.

This mirrors the workaround confirmed in the issue (adding the two fields to `text_config`) without editing the checkpoint.

## Tests

`tests/test_gemma4_per_layer_config.py` (12 tests): derivation from the issue's config, explicit fields win, only full-attention layers count, no `layer_types`, disagreeing overrides, non-Gemma4 / missing `per_layer_config`; the `_on_load` wrapper hands the derived fields to a stubbed `mlx_vlm.utils.load_config`, keeps explicit fields, is a no-op when there is nothing to derive, tolerates a missing `config.json`, and restores the original function afterwards.

`ruff check` / `ruff format --check` are clean for the added code (the pre-existing findings in `vlm.py` are unchanged).

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -766,6 +766,111 @@ def _is_mlx_format_safetensors_dir(model_dir: Path) -> bool:
     return isinstance(metadata, dict) and metadata.get("format") == "mlx"
 
 
+def _gemma4_global_kv_from_per_layer_config(config: dict) -> dict[str, int]:
+    """Derive Gemma4's legacy full-attention head fields from ``per_layer_config``.
+
+    Newer Gemma4 checkpoints (Transformers >= 5.15) record the ``head_dim`` /
+    ``num_key_value_heads`` overrides of the full-attention layers under
+    ``text_config.per_layer_config`` instead of the legacy global
+    ``global_head_dim`` / ``num_global_key_value_heads`` fields. The pinned
+    mlx-vlm Gemma4 loader reads only the legacy fields, so it sizes the
+    full-attention K/V projections with the sliding-window head count and
+    ``load_weights`` fails with a shape mismatch (#3537).
+
+    Returns the legacy fields that are absent from ``text_config`` and can be
+    derived unambiguously (every overridden full-attention layer agrees), or
+    an empty dict.
+    """
+    text_config = config.get("text_config")
+    if not isinstance(text_config, dict):
+        return {}
+    model_type = str(text_config.get("model_type") or config.get("model_type") or "")
+    if not model_type.startswith("gemma4"):
+        return {}
+    per_layer = text_config.get("per_layer_config")
+    if not isinstance(per_layer, dict) or not per_layer:
+        return {}
+    layer_types = text_config.get("layer_types")
+    if not isinstance(layer_types, list):
+        layer_types = None
+
+    derived: dict[str, int] = {}
+    for legacy_key, layer_key in (
+        ("global_head_dim", "head_dim"),
+        ("num_global_key_value_heads", "num_key_value_heads"),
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for the fix. I reproduced the loading failure with a per-layer-only config and verified that this change loads the same Gemma4 weights successfully while preserving legacy-config loading. LGTM, merging. I'll follow up with a small fix for null legacy fields, where `setdefault` currently leaves the derived values unapplied.

UnderLotus:
Wow, that was lightning fast! huge thanks to everyone involved for the quick turnaround 🫶

PR #3551, opened 2026-09-10
fix(grammar): size the VLM grammar bitmask from text_config.vocab_size

### LINKED ISSUE #3550: Structured output (json_schema) never terminates on the VLM engine — GrammarMatcher rejected token 0 every step until max_tokens
<html><head></head><body><p><strong>
<h2>Summary</h2>
<p>Any request with <code>response_format: {"type": "json_schema", ..., "strict": true}</code> against a model served by <code>VLMBatchedEngine</code> loops until it hits <code>max_tokens</code>. The same request against a model served by <code>BatchedEngine</code> works correctly.</p>
<p>No image is required to reproduce — a plain text prompt to a VLM-backed model is enough, so this is not about image handling.</p>
<h2>Environment</h2>
<ul>
<li>oMLX 0.6.4, Homebrew install on macOS, Apple Silicon (M5 Pro, 128 GB)</li>
<li>xgrammar present via <code>brew reinstall omlx --with-grammar</code></li>
<li>Grammar works: <code>GrammarCompiler initialized for ...</code> appears for both engines</li>
</ul>
<h2>Reproduction</h2>
<pre><code class="language-bash">curl -s -m 120 http://localhost:8000/v1/chat/completions \
 -H 'Content-Type: application/json' -d '{
 "model":"Qwen3-VL-8B-Instruct-8bit",
 "messages":[{"role":"user","content":"2 EL Olivenöl"}],
 "response_format":{"type":"json_schema","json_schema":{"name":"z","strict":true,
  "schema":{"type":"object","properties":{
    "quantity":{"type":"number"},"unit":{"type":"string"},"food":{"type":"string"}},
  "required":["quantity","unit","food"],"additionalProperties":false}}}}'
</code></pre>
<h2>Actual</h2>
<p>The log fills with one warning per decode step, roughly every 20 ms, always token 0:</p>
<pre><code>omlx.api.grammar - WARNING - GrammarMatcher rejected token 0
omlx.api.grammar - WARNING - GrammarMatcher rejected token 0
...
omlx.server - INFO - Chat completion: model=Qwen3-VL-8B-Instruct-8bit,
  4096 tokens in 88.37s (46.4 tok/s), prompt: 616,
  finish_reason=length, max_tokens=4096
</code></pre>
<p>Generation only stops at <code>max_tokens</code>. The response is unusable.</p>
<h2>Expected</h2>
<p>Valid schema-conforming JSON with <code>finish_reason=stop</code>, as produced by the text engine.</p>
<h2>Affected / unaffected</h2>

Model | Engine | Result
-- | -- | --
Qwen3-VL-8B-Instruct-8bit | VLMBatchedEngine | loops to max_tokens
Qwen3-VL-4B-Instruct-8bit | VLMBatchedEngine | loops to max_tokens
Qwen3.6-27B-MLX-8bit | VLMBatchedEngine | loops to max_tokens
Qwen3-4B-Instruct-2507-8bit | BatchedEngine | works, finish_reason=stop

<p>The split follows the engine, not the model family or the quantization.</p>
<h2>Hypothesis</h2>
<p>In <code>omlx/api/grammar.py</code>, <code>_apply_mask(mx_bitmask, logits, self._vocab_size)</code> masks logits against a vocabulary size taken from the text tokenizer. VLM checkpoints extend the vocabulary with vision/image tokens, so on the VLM path the mask and the logits are misaligned. The sampled token is therefore always rejected:</p>
<pre><code class="language-python">if not self._matcher.accept_token(token_id):
    logger.warning("GrammarMatcher rejected token %d", token_id)
</code></pre>
<p><code>accept_token</code> failing leaves the matcher state unchanged, so the next step repeats identically and the loop never advances. That the rejected id is consistently 0 suggests the masked logits collapse to index 0.</p>
<h2>Impact</h2>
<p>Any OpenAI-compatible client that uses the structured-output helper (<code>client.chat.completions.parse</code>) is blocked from using VLM-served models. In our case this is Mealie's recipe import, whose image path always goes to a VLM provider — image import is unusable while the text path works fine.</p>
<h2>Workaround</h2>
<p>Route structured-output requests to a text-only model. There appears to be no way to disable grammar enforcement per model or per request short of omitting <code>response_format</code>, which clients like Mealie do not offer. Setting <code>sampling.max_tokens</code> to a low value at least bounds the hang.</p></body></html>

L4XB:
## Summary

Fixes #3550.

Structured output (`response_format: json_schema`) on `VLMBatchedEngine` looped to `max_tokens` with `GrammarMatcher rejected token 0` on every step, while the same request on `BatchedEngine` worked.

## Root cause

`resolve_vocab_size` (used by `create_grammar_compiler` for the xgrammar `TokenizerInfo` and by the scheduler for the `GrammarConstraintProcessor` bitmask width) read the top-level `config.vocab_size` first and only fell back to `text_config.vocab_size`. For a composite VLM config the top-level field is frequently a dataclass placeholder the checkpoint never sets: mlx-vlm's Qwen3-VL `ModelConfig` (and Qwen2.5-VL, LLaVA) defaults it to **32000**, and `Qwen/Qwen3-VL-8B-Instruct`'s `config.json` only carries `text_config.vocab_size = 151936`. So on the VLM path the tokenizer info and the bitmask were sized for 32000 tokens against 151936 logits, the mask and the logits were misaligned, and the matcher rejected every sampled token, which is exactly the reporter's hypothesis.

## Change

`omlx/utils/tokenizer.py`: `resolve_vocab_size` prefers the nested `text_config.vocab_size` (dict or object) when present and falls back to the top-level `config.vocab_size` / `args.vocab_size` otherwise; non-positive and bool placeholders are skipped. The text engine path is unchanged (plain LLM configs have no `text_config`).

## Tests

`tests/test_resolve_vocab_size.py`: nested wins over the 32000 placeholder (object and dict forms), plain LLM top-level, top-level when `text_config` has no `vocab_size` (e.g. Gemma 3), `args` fallback, `config` over `args`, and unusable values. Three cases fail on `main`.

`pytest tests/test_resolve_vocab_size.py tests/test_utils_tokenizer.py tests/test_grammar.py` → all green; `ruff check` / `ruff format --check` clean.

I could not run a VLM checkpoint end to end here, so the reproduction is the config-level one (mlx-vlm `ModelConfig.vocab_size` default vs the checkpoint's `text_config.vocab_size`).

### DIFF
--- a/omlx/utils/tokenizer.py
+++ b/omlx/utils/tokenizer.py
@@ -44,11 +44,24 @@ def unwrap_tokenizer(tokenizer):
     return tokenizer
 
 
+def _positive_int(value: Any) -> int | None:
+    if isinstance(value, bool) or not isinstance(value, int) or value <= 0:
+        return None
+    return value
+
+
 def resolve_vocab_size(model: Any) -> int | None:
     """Extract vocab_size from a model's config/args, handling nested configs.
 
-    Tries ``model.config.vocab_size``, then ``model.args.vocab_size``,
-    then ``text_config.vocab_size`` for VLM composite models (e.g. Qwen3.5).
+    For a composite (VLM) config the nested ``text_config.vocab_size`` wins
+    over the top-level ``vocab_size``: the language model's embedding and
+    ``lm_head`` are sized from the text config, while the top-level field
+    is often a dataclass placeholder that the checkpoint's ``config.json``
+    never sets (mlx-vlm's Qwen3-VL ``ModelConfig`` defaults it to 32000
+    against a 151936-token text vocabulary). Sizing the grammar bitmask
+    from that placeholder misaligns the mask with the logits, and every
+    sampled token is rejected until ``max_tokens`` (#3550). Falls back to
+    ``model.config.vocab_size`` / ``model.args.vocab_size`` for plain LLMs.
 
     Args:
         model: An MLX model object (LLM, VLM, or any object with config/args).
@@ -62,15 +75,17 @@ def resolve_vocab_size(model: Any) -> int | None:
         config = getattr(model, attr, None)
         if config is None:
             continue
-        vs = getattr(config, "vocab_size", None)
-        if isinstance(vs, int):
-            return vs
         text_cfg = getattr(config, "text_config", None)
         if isinstance(text_cfg, dict):
-            vs = text_cfg.get("vocab_size")
+            nested = _positive_int(text_cfg.get("vocab_size"))
         elif text_cfg is not None:
-            vs = getattr(text_cfg, "vocab_size", None)
-        if isinstance(vs, int):
+            nested = _positive_int(getattr(text_cfg, "vocab_size", None))
+        else:
+            nested = None
+        if nested is not None:
+            return nested
+        vs = _positive_int(getattr(config, "vocab_size", None))
+        if vs is not None:
             return vs
     return None
 
--- a/tests/test_resolve_vocab_size.py
+++ b/tests/test_resolve_vocab_size.py
@@ -1,0 +1,65 @@
+"""resolve_vocab_size must size the grammar bitmask from the language model's
+vocabulary. A composite (VLM) config's top-level ``vocab_size`` is often a
+dataclass placeholder the checkpoint never sets (mlx-vlm's Qwen3-VL
+``ModelConfig`` defaults it to 32000 while ``text_config.vocab_size`` is
+151936); sizing the mask from it misaligned mask and logits, so structured
+output on the VLM engine rejected every sampled token until max_tokens
+(#3550)."""
+
+from types import SimpleNamespace
+
+from omlx.utils.tokenizer import resolve_vocab_size
+
+
+def test_nested_text_config_wins_over_top_level_placeholder():
+    config = SimpleNamespace(
+        vocab_size=32000,  # mlx-vlm ModelConfig default, not from config.json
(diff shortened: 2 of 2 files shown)


L4XB:
CI note: the `test (3.12)` job failed on `tests/test_prefix_cache_gdn_split.py::test_split_restore_walks_back_from_structurally_invalid_sidecar` (`num_tokens` 8 vs 12 after the sidecar walk-back). That test does not touch `resolve_vocab_size` or the grammar path; it passes locally on this branch (3/3 runs) and on `main` at the same base, so it looks like a timing flake of the SSD sidecar restore rather than a regression from this change.

L4XB:
The `test (3.12)` failure here is the intermittent split-GDN store flake (`Rejecting split-GDN placeholder block 3 because its recurrent checkpoint was not committed`), unrelated to the grammar change; the same test module failed on other PRs today. Root cause and fix in #3563 (`take_staged_file` could remove the snapshot request directory between the writer's `mkdir` and its write). Merged `main` into this branch to re-run CI.

jundot:
Thanks for the fix. I reproduced the issue on Qwen3-VL-30B-A3B: the old resolver produced repeated `!` tokens until the token limit, while this change returned valid schema-conforming JSON and stopped normally. Preferring the text config matches the language model’s output vocabulary and preserves the plain LLM fallback. LGTM, I’ll merge this.

PR #3553, opened 2026-09-10
perf(qwen4_exp): Qwen3.8-Flash-Next decode: gathered attention kernel, NAX indexer scores, fused verify rows, primed re-entry, narrow gathered windows

Rascal:
### TLDR
Accelerating the MTP decoding speed(and little bit prefill) for the Qwen3.8-Flash-Next model.

This PR is essentially a collection of small optimizations here and there; if combining them all into a single PR was a mistake, just let me know and I'll split them into separate PRs.

### Summary
Six commits—improvements on top of `main` d7c63c2. Every commit is independent in files, exact by construction or bit-exact against the composed stock ops, kill-switched, and measured with interleaved A/B runs (A B B A, or A B C C B A) on an M5 Max 128 GB with `Qwen3.8-Flash-Next-oQ4e-mtp`, SSD PLE, 512 generated tokens, greedy unless noted. Base numbers are the state of `main` at the time of each measurement; the whole-branch table at the end is against current main.
The three bit-exact ones (3, 4, 6) are the safe core, ~70% of total performance gain; 1 and 2 are the tolerance-level kernels and both have kill switches.
Commit 2 is both about prefill and decode, because it sits in the same QSA indexer code.

### Benchmarks
`main` d7c63c27 vs this branch a04d2408

### Per-cycle cost at pinned depth 3
main → branch (text-independent: backbone and draft-head milliseconds per verify cycle from the MTP stats, Python code, four passes per cell averaged over mirrored arms):

| context | backbone ms | draft head ms | wall ms / cycle | tokens / cycle | accept | client tok/s |
|---|---|---|---|---|---|---|
| 16k | 34.41 → 34.46 (+0.1%) | 3.37 → 2.89 (−14.2%) | 37.87 → 37.43 (−1.2%) | 2.78 → 2.68 (−3.6%) | 74.9 → 74.2% (−0.7 pt) | 70.7 → 68.9 (−2.5%) |
| 65k | 34.92 → 33.29 (−4.7%) | 5.67 → 3.80 (−33.0%) | 40.70 → 37.18 (−8.6%) | 2.61 → 2.88 (+10.3%) | 71.7 → 78.3% (+6.6 pt) | 61.7 → 74.0 (+19.9%) |
| 136k | 36.75 → 35.01 (−4.7%) | 9.57 → 5.39 (−43.7%) | 46.43 → 40.50 (−12.8%) | 2.68 → 3.03 (+13.1%) | 74.2 → 81.7% (+7.4 pt) | 55.8 → 71.9 (+28.9%) |
| 210k | 37.55 → 35.44 (−5.6%) | 11.22 → 5.41 (−51.8%) | 48.89 → 40.96 (−16.2%) | 2.43 → 2.53 (+4.1%) | 66.5 → 70.0% (+3.5 pt) | 48.0 → 59.5 (+24.0%) |
mtp_verify_cycle

The wall-time-per-cycle column is the kernel work: −1.2% at 16k, −8.6% at 65k, −12.8% at 136k, −16.2% at 210k. The backbone share (−5% from 65k up) is the gathered decode kernel plus the fused verify rows; the head share (−14.2% to −51.8%) is the narrow-window gate, whose saving grows with the prefix. Tokens per cycle and acceptance are shown separately on purpose: the NAX indexer scores are exact only to 2e-5, which flips near-tie block selections, so the greedy continuation forks early and each tree then verifies a different text. On these prompts the fork favoured the branch (+4 to +13% tokens per cycle); that part is continuation luck, not kernel speed, and the NAX kill switch returns tokens per cycle to the base's value exactly.

**Where the cycle saving comes from aka bit-exact commits vs all** (pinned depth 3, Python code, three arms A X C C X A: main; the branch with the two tolerance-level kernels switched off, i.e. the three bit-exact commits 3, 4 and 6 plus commit 5, which changes only when the controller re-enters after a park and never engages at pinned depth; the branch with all six):

| context | main ms / cycle | bit-exact commits only | all six commits | bit-exact share of the saving | throughput at equal acceptance: bit-exact / tolerance on top |
|---|---|---|---|---|---|
| 16k | 37.54 | 37.11 (−1.1%) | 37.11 (−1.1%) | 100% | +1.2% / +0.0% |
| 65k | 40.88 | 38.58 (−5.6%) | 36.84 (−9.9%) | 57% | +6.0% / +4.7% |
| 136k | 46.26 | 42.05 (−9.1%) | 40.76 (−11.9%) | 77% | +10.0% / +3.2% |
| 210k | 48.74 | 42.73 (−12.3%) | 40.76 (−16.4%) | 75% | +14.1% / +4.8% |
bit-exact-vs-all

bit-exact-vs-all_mtp

The bit-exact set (fused GDN verify rows, grouped verify projections, narrow gate) reproduces main's text, with tokens per cycle within 3% of main's in every cell, and takes the whole draft-head saving. The gathered decode kernel adds a flat 1.3–2.2 ms of backbone from 65k up; the NAX indexer scores add nothing at decode. A reviewer who wanted a strictly bit-exact PR could drop commits 1 and 2 and keep 57–100% of the cycle saving, at the cost of the serial gain and the prefill gain; commit 5 is control flow only and does not touch the numerics of any kernel.

### Adaptive MPT
(greedy, 512 tokens; three prompt families, four passes each, thinking on):

| context | Python code | English prose | mixed-language code | mean of the three |
|---|---|---|---|---|
| 4k | 70.4 → 80.8 (+14.7%) | 73.3 → 74.9 (+2.2%) | 77.6 → 79.1 (+1.9%) | +6.3% |
| 16k | 71.5 → 76.8 (+7.5%) | 70.8 → 71.0 (+0.4%) | 77.1 → 77.1 (+0.0%) | +2.6% |
| 65k | 72.8 → 69.7 (−4.2%) | 66.2 → 76.1 (+14.8%) | 72.2 → 85.4 (+18.2%) | +9.6% |
| 136k | 63.8 → 73.2 (+14.7%) | 57.7 → 67.0 (+16.1%) | 61.6 → 67.1 (+9.0%) | +13.3% |
| 210k | 59.9 → 65.5 (+9.4%) | 63.0 → 67.9 (+7.8%) | 62.4 → 70.1 (+12.4%) | +9.9% |
mtp_decode

mtp_decode_prompts

Across the 15 cells the mean is +8.3% with 14 positive. The one negative, Python code at 65k, is a continuation fork rather than slower code: the base's text runs at 85% acceptance (91–94% at the third draft position) and the branch's at 72%, with no parks on either side. The same prompt at pinned depth 3 is +20% on the branch, main itself at pinned depth 3 produces a 72%-acceptance text, and with the two tolerance-level kernels switched off (`OMLX_QWEN4_QSA_GATHERED_DECODE=0 OMLX_QWEN4_QSA_NAX_SCORES=0`) the branch reproduces main's text byte for byte and runs it 6.5% faster. Per-cycle cost favours the branch in every family and context, so the spread between families is the text, not the code.

### Serial decode and cold prefill
(Python code prompts, serial is MTP off; prefill is the pass-1 time to first token, serial and MTP boots pooled):

| context | serial main → branch | cold prefill main → branch |
|---|---|---|
| 4k | 61.6 → 62.4 (+1.4%) | 10.9 → 10.2 s (−6.1%) (first request after each server start, includes warm-up) |
| 16k | 59.0 → 59.4 (+0.8%) | 19.0 → 18.4 s (−2.9%) |
| 65k | 57.1 → 58.9 (+3.1%) | 47.2 → 46.7 s (−1.1%) |
| 136k | 55.7 → 57.0 (+2.3%) | 66.9 → 65.2 s (−2.4%) |
| 210k | 54.2 → 55.7 (+2.8%) | 70.8 → 68.5 s (−3.3%) |
serial_decode

cold_prefill

Serial decode is +1 to +3% from the gathered decode kernel; cold prefill is 1–3% shorter from the NAX indexer scores.

### Commit 1 — indexed-KV attention kernel for gathered QSA decode rows (`omlx/custom_kernels/decode_fast`)

**What.** A two-pass decode SDPA kernel that reads only the selected K/V rows through an int32 index list (masked entries finite-min), so the gathered decode arm and MTP verify rows attend the top-k blocks without materialising the gather and without the masked dense read of the whole prefix. Portable fallback (`take` + masked `mx.fast` SDPA) stays; wrapper always passes contiguous inputs and checks layouts at eval.

**Gain.** MTP +0.3 / +4.8 / +6.8 / +1.0% at 16k / 63k / 134k / 229k, serial +0 / +1.6 / +1.7 / +2.3%, prefill flat. Per verify cycle at 63k the gather + portable attention it replaces was ~2.8 ms of a 34 ms backbone.

**Parity and tests.** `tests/test_decode_fast_sdpa_gathered.py` (22 shape/dtype cases vs the portable path), `tests/test_qwen4_qsa_gathered_decode_attn.py` (verify rows and single rows go through the kernel; the kernel matches the portable gather + SDPA math on its own inputs at 1e-4; the kill switch keeps the portable arm). The layer test deliberately does not compare two separately prefilled caches: with a tiny prefix a near-tie can make them select different blocks for a row, which is a property of top-k selection, not of either attention path. Kill switch `OMLX_QWEN4_QSA_GATHERED_DECODE=0`.

### Commit 2 — QSA indexer scores on the M5 tensor units (`omlx/custom_kernels/glm_moe_dsa`)

**What.** A second metallib (`omlx_glm_kernels_nax.metallib`, compiled with `-mmacosx-version-min=26.2`, SDK-gated in CMake) runs the indexer score matmul on NAX with 64×64 tiles, a ReLU-accumulate epilogue and the masked sentinel; the host mirrors `is_nax_available()` and keeps the steel kernel as the fallback, so machines without NAX (M3 Ultra) are unaffected.

**Gain.** 1.5× per layer-chunk against the steel kernel; cold prefill −1 to −3% at 16k–229k; decode flat.

**Parity and tests.** `tests/test_qwen4_qsa_nax_indexer.py`: sentinel bit-exact, scores within 2e-5 relative, top-512 sets identical, route and kill switch (`OMLX_QWEN4_QSA_NAX_SCORES=0`). Note for reviewers: the existing native-indexer tests pin the kill switch so they keep exercising the steel path.

### Commit 3 — fused GDN prework and norm-gate for MTP verify rows, precise sigmoid in the fused kernels (`omlx/patches/qwen35_gdn_prework.py`)

**What.** Under MTP the 36 GDN layers run with 2..4 verify rows; the existing fusion covered single-row decode and only the conv for 3..9 rows. A row-batched verify arm (1..8 rows) fuses conv, SiLU, q/k RMS norm and scales, `g`, `beta` and the next conv state into one dispatch, feeds precomputed `g`/`beta` to the states-returning recurrence, and fuses the gated RMS norm; depth-1 verify (2 rows) was fully stock before.

**Fix.** `mx.sigmoid` is a prebuilt op compiled with the precise `exp`, while `metal_kernel` sources and compiled functions use the fast one; the two differ at exactly one bf16 input (`b = -6.84375`). Both the new verify kernel and the existing single-row decode kernel computed `beta` with the fast exp, which shifted the GDN recurrent state by an fp32 ulp on rare cycles and forked greedy output about 160 tokens into a completion. Both now use `metal::precise::exp` for `beta` (SiLU keeps the fast exp because `nn.silu` is compiled). Verified with an exhaustive sweep of every finite bf16 gate input on real layer weights (0 mismatches in 1.6M); greedy completions are byte-identical with the arm on and off at pinned depths 1 and 3.

**Gain.** Pinned depth 1 at 5k +3.4%; adaptive: code +0.5 / +1.1%, prose +0.1 / +1.9% (5k / 40k). With per-layer `async_eval` the GPU queue already hides most small dispatch gaps, so this is the honest size of a dispatch-count fusion on this model.

**Parity and tests.** Bit-exact against the composed stock ops for 1, 2, 3, 4 and 8 rows and against the stock forward on real layer-0 weights over 220M elements (out, both cache states, q/k/v, intermediate states). Route tests commit cache states and the rollback sink only after every fallible step. Kill switch `OMLX_QWEN4_GDN_VERIFY_FUSED=0`. Tests: `tests/test_qwen4_gdn_verify_fused.py`, existing `tests/test_qwen35_gdn_prework.py`.

### Commit 4 — grouped verify-row quantized projections (`omlx/patches/qwen35_grouped_linears.py`)

**What.** `_target_verify_linears` issues one quantized matmul per projection for batch-one rows. Concatenating weights along the output axis is bit-exact per output row, so GDN in-projections, attention q/k/v and MLP gate/up now dispatch once per (bits, group_size) signature for 2..8 verify rows. Groups are cached `QuantizedLinear` subclasses so the verify-qmm routing patch sees the usual call shape; a group whose routing would differ from a member's falls back to member calls.

**Gain.** Measured together with commit 3 against their base: MTP code +0.4 / −0.4%, prose +2.3 / +2.6% (5k / 34k); serial flat. Grouping single-row decode projections measured −1.6 to −2.4% serial in all sixteen passes, so it is restricted to verify rows.

**Parity and tests.** `mx.array_equal` to separate linears for mixed signatures {4/64, 5/128, 6/64, 8/32, 8/128}, 2..8 rows; one dispatch per signature; patch rebinds the vendored Qwen4 module's own import of the helper. Kill switch `OMLX_QWEN35_GROUPED_LINEARS=0`. Tests: `tests/test_qwen35_grouped_linears.py`.

### Commit 5 — keep the draft head primed during parked standard decode (`omlx/patches/mlx_lm_mtp/`)

**What.** When the depth controller parks speculation, `_park_mtp_to_standard` dropped the head cache, so every re-entry probe 128/256/…/4096 tokens later started cold (`primed=0`) and, at long context, usually failed, doubling the cooldown. The committed head cache now becomes a parked priming context: standard decode buffers (hidden, next token) pairs and folds them through one batched `mtp_forward` every 32 tokens; `take_primed` flushes the remainder at the probe and returns the exact head offset. The hidden capture is best-effort so a backbone that yields no hidden cannot fail the park handoff.

**Effect.** 68k prose prompt, production sampling: the controller parked after 154 tokens, the probe activated with `primed=68600` and succeeded, decode stayed on MTP for the remaining 742 tokens. Parking is rare at 68..82k on this branch, so the throughput effect is bounded by park frequency; zero cost while the controller never parks.

**Tests.** `tests/test_mtp_parked_fold.py` (tiny mlx-lm model: block folds and the probe flush reproduce a direct fold of the same pairs to 1e-5 with equal head offsets; discontiguous steps drop the context; prime window ignored; park survives a backbone without hidden) plus the existing priming, controller and MTP patch suites. Kill switch `OMLX_MTP_PARKED_FOLD=0`, block size `OMLX_MTP_PARK_FOLD_BLOCK`.

### Commit 6 — narrow text windows take the gathered QSA arm at long context (`qwen4_exp/language.py`)

**What.** The MTP head's committed fold runs the head's QSA layer with 2..8 rows. None of the gathered arms accepted that shape (decode: one row; prefill: ≥ 16 rows; verify: `target_verify` only), so the fold attended the whole prefix through the masked-dense path: 4.9 ms per 3–4-row fold at 82k against 1.0 ms gathered, and 90 of 148 folds have 4 rows. `_gathered_text_prefill_eligible` keeps its 16-row floor at short context and adds one clause: batch-one text windows of 2..15 rows take the gathered arm once `cache.offset >= OMLX_QWEN4_GATHERED_NARROW_MIN_CONTEXT` (default 16384). The arm itself is unchanged; verify rows keep their own gate; drafts are verified exactly.

**Gain.** Adaptive MTP: code 41k +2.3%, code 82k +4.6% (all four passes above base), prose 34k −1.3%, prose 68k −0.1%; head time per cycle at 82k 3.4 → 1.8 ms. The gain is context-proportional.

**Tests.** `tests/test_qwen4_qsa_narrow_gathered.py` (gate above/below the threshold, strict budget rule, single rows on the decode arm, verify rows excluded, env override) plus the existing gathered-arm suites.

### Server-level greedy parity on this branch (pinned depth 3, code 5k and 41k, 512 tokens)

Each kill switch was toggled against the all-on run and the completions compared byte for byte. Fused verify rows off, grouped projections off, narrow gate off: identical, as their bit-exact construction requires (the narrow gate changes only drafts, never verified tokens). Gathered kernel off: the 5k completion identical with identical cycle and acceptance counts, the 41k completion forks at token 24. NAX indexer scores off: the 41k completion forks at token 8. Both are tolerance-level kernels (1e-4 attention, 2e-5 indexer scores), and a long prefix turns any such difference into a flipped near-tie block selection or an argmax flip within a few tokens; a separately measured control showed the same divergence from an unrelated one-ulp perturbation of a prefill norm, with no change in teacher-forced NLL. Acceptance rates on the diverged continuations stay in the usual 70–76% band.

### Method notes

- Every kernel's parity is measured against the composed stock ops on real weights, never kernel-vs-kernel; anything on the MTP verify path was also checked for byte-identical greedy completions with the change on and off.
- Two lessons from this work that apply to future kernels: JIT Metal kernels must match the reference op's compile path (prebuilt ops use the precise exp/log, compiled functions and `metal_kernel` sources the fast ones), and synchronous micro-benchmarks sit at the ~250 µs eval floor on this model, so only in-situ A/B runs can resolve dispatch-level effects.
- Parked as measured-negative and not included: prompt-lookup n-gram drafts (−10 to −20%: wide speculative windows lose on this MoE model whatever the draft source), a block-sparse row-group prefill kernel (3.6× slower), and an MMA rewrite of the GDN prefill recurrence (recurrence is 6.5% of a prefill chunk).

### DIFF
--- a/docs/superpowers/plans/profiling/gdn_dual_probe.py
+++ b/docs/superpowers/plans/profiling/gdn_dual_probe.py
@@ -1,0 +1,77 @@
+"""Shim: for 2-row gdn_sink calls, run the stock GDN forward on a shadow cache next to the patched
+one and log any numeric difference (out, conv, recurrent, q/k/v, intermediate states)."""
+import logging, os
+import mlx.core as mx
+import omlx.patches.qwen35_gdn_prework as pm
+from mlx_vlm.models.qwen3_5 import language as q35
+
+log = logging.getLogger("omlx.patches.qwen35_gdn_prework")
+ROWS = int(os.environ.get("OMLX_DUAL_ROWS", "2"))
+LIMIT = int(os.environ.get("OMLX_DUAL_LIMIT", "3600"))
+_apply = pm.apply_qwen35_gdn_prework_patch
+_orig_call = q35.Qwen3_5GatedDeltaNet.__call__
+state = {"n": 0, "layer": 0, "diff_calls": 0}
+
+
+class Shadow:
+    def __init__(self, cache):
+        self._s = {0: cache[0], 1: cache[1]}
+        self.lengths = getattr(cache, "lengths", None)
+        self.left_padding = getattr(cache, "left_padding", None)
+
+    def __getitem__(self, i):
+        return self._s[i]
+
+    def __setitem__(self, i, v):
+        self._s[i] = v
+
+    def advance(self, n):
+        pass
+
+
+def apply_and_wrap():
+    ok = _apply()
+    cls = q35.Qwen3_5GatedDeltaNet
+    patched = cls.__call__
+
+    def dual(self, inputs, mask=None, cache=None, gdn_sink=None, target_verify=False):
+        if gdn_sink is None or cache is None or inputs.shape[1] != ROWS or state["n"] >= LIMIT:
+            return patched(self, inputs, mask=mask, cache=cache, gdn_sink=gdn_sink, target_verify=target_verify)
+        state["n"] += 1
+        shadow = Shadow(cache)
+        sink2 = []
+        ref = _orig_call(self, inputs, mask=mask, cache=shadow, gdn_sink=sink2, target_verify=True)
+        out = patched(self, inputs, mask=mask, cache=cache, gdn_sink=gdn_sink, target_verify=target_verify)
+        pairs = [("out", out, ref), ("conv", cache[0], shadow[0]), ("rec", cache[1], shadow[1]),
+                 ("q", gdn_sink[-1][0], sink2[0][0]), ("k", gdn_sink[-1][1], sink2[0][1]),
+                 ("v", gdn_sink[-1][2], sink2[0][2]), ("a", gdn_sink[-1][3], sink2[0][3]),
+                 ("b", gdn_sink[-1][4], sink2[0][4]), ("conv_in", gdn_sink[-1][9], sink2[0][9]),
+                 ("inter", gdn_sink[-1][11], sink2[0][11])]
+        diffs = {}
+        for name, x, y in pairs:
+            if x is None or y is None:
+                if x is not y:
+                    diffs[name] = "None-mismatch"
+                continue
+            if x.shape != y.shape:
+                diffs[name] = f"shape {x.shape} vs {y.shape}"
+                continue
+            d = mx.abs(x.astype(mx.float32) - y.astype(mx.float32)).max().item()
+            if d > 0:
+                diffs[name] = d
+        if diffs:
+            state["diff_calls"] += 1
+            if state["diff_calls"] <= 40:
+                log.info("DUALPROBE call=%d layer=%d S=%d bits=%s diffs=%s", state["n"], (state["n"] - 1) % 36, inputs.shape[1],
+                         (self.in_proj_qkv.bits, self.in_proj_z.bits), diffs)
+        elif state["n"] % 360 == 0:
+            log.info("DUALPROBE call=%d no diffs so far (diff_calls=%d)", state["n"], state["diff_calls"])
+        return out
+
+    cls.__call__ = dual
+    log.info("DUALPROBE installed (rows=%d)", ROWS)
+    return ok
+
+
+pm.apply_qwen35_gdn_prework_patch = apply_and_wrap
+print("[gdn_dual_probe] installed", flush=True)
--- a/docs/superpowers/plans/profiling/gdn_g_sweep.py
+++ b/docs/superpowers/plans/profiling/gdn_g_sweep.py
@@ -1,0 +1,49 @@
+"""Sweep every finite bf16 value of the GDN gate input `a` through the fused kernel's g/beta and
+compare with mlx-vlm's _compute_g_beta on a real layer's A_log/dt_bias. usage: gdn_g_sweep.py [layer]"""
+import json, os, sys
+import mlx.core as mx
+from mlx_vlm.models.qwen3_5.gated_delta import _compute_g_beta
+from omlx.patches import qwen35_gdn_prework as pm
+
+M = os.path.expanduser("~/.omlx-bench/models/Jundot/Qwen3.8-Flash-Next-oQ4e-mtp")
+GDN_IDX = int(sys.argv[1]) if len(sys.argv) > 1 else 11  # index among the GDN layers
+cfg = json.load(open(f"{M}/config.json")); tcfg = cfg.get("text_config", cfg)
+idx = json.load(open(f"{M}/model.safetensors.index.json"))["weight_map"]
+gdn_layers = sorted({int(k.split(".layers.")[1].split(".")[0]) for k in idx if ".linear_attn.A_log" in k})
+LAYER = gdn_layers[GDN_IDX]
+p = f"language_model.model.layers.{LAYER}.linear_attn."
+d = mx.load(f"{M}/{idx[p + 'A_log']}")
+A_log, dt_bias = d[p + "A_log"], d[p + "dt_bias"]
+HK, HV, DK, DV, C = 16, 48, 128, 128, 10240
+S = 8
+# every bf16 bit pattern with |a| finite and < 2^7 (the model's a range is a few units)
+bits = mx.arange(0, 65536, dtype=mx.uint32).astype(mx.uint16)
+vals = bits.view(mx.bfloat16)
+import numpy as np
+keep = np.asarray(mx.isfinite(vals) & (mx.abs(vals) < 128))
+vals = mx.array(np.asarray(vals.astype(mx.float32))[keep]).astype(mx.bfloat16)
+n = int(vals.size)
+per = S  # each row carries one value replicated across all heads
+vals = mx.concatenate([vals, mx.zeros((per - n % per,), dtype=mx.bfloat16)]) if n % per else vals
+calls = int(vals.size) // per
+print(f"layer {LAYER}: sweeping {n} bf16 values of a in {calls} kernel calls")
+inv = DK**-0.5
+q_scale = mx.array(inv * inv, dtype=mx.bfloat16); k_scale = mx.array(inv, dtype=mx.bfloat16)
+qkv = mx.zeros((1, S, C), dtype=mx.bfloat16); conv_state = mx.zeros((1, 3, C), dtype=mx.bfloat16)
+conv_w = mx.zeros((C, 4, 1), dtype=mx.bfloat16)
+g_mis = b_mis = 0; examples = []
+for c in range(calls):
+    a = mx.broadcast_to(vals[c * per:(c + 1) * per].reshape(1, S, 1), (1, S, HV))
+    a = mx.contiguous(a)
+    b = a  # sweep beta on the same values
+    out = pm.qwen4_verify_prework_fused(qkv, conv_state, conv_w, q_scale, k_scale, b, a, A_log, dt_bias, HK, HV, DK, DV)
+    g_k, beta_k = out[4], out[5]
+    g_r, beta_r = _compute_g_beta(A_log, a, b, dt_bias)
+    mx.eval(g_k, beta_k, g_r, beta_r)
+    gm = (g_k != g_r); bm = (beta_k != beta_r)
+    g_mis += int(gm.sum().item()); b_mis += int(bm.sum().item())
+    if gm.any().item() and len(examples) < 6:
+        i = int(mx.argmax(gm.reshape(-1)).item())
+        examples.append((float(a.reshape(-1)[i].item()), i % HV, float(g_k.reshape(-1)[i].item()), float(g_r.reshape(-1)[i].item())))
+print(f"g mismatches {g_mis}/{n * HV}  beta mismatches {b_mis}/{n * HV}")
+for e in examples: print("  a=%g head=%d kernel=%.9g ref=%.9g" % e)
--- a/docs/superpowers/plans/profiling/gdn_gate_probe.py
+++ b/docs/superpowers/plans/profiling/gdn_gate_probe.py
@@ -1,0 +1,29 @@
+"""Shim: log the first evaluations of the Qwen4 fused GDN verify gate per (rows, result) with call site."""
+import logging, traceback
+import omlx.patches.qwen35_gdn_prework as pm
+log = logging.getLogger("omlx.patches.qwen35_gdn_prework")
+_orig = pm._qwen4_verify_dynamic_eligible
+_seen = {}
+
+
+def probe(module, inputs, mask, cache, gdn_sink):
+    r = _orig(module, inputs, mask, cache, gdn_sink)
+    S = inputs.shape[1] if getattr(inputs, "ndim", 0) == 3 else None
+    key = (S, r)
+    if _seen.get(key, 0) < 2:
+        _seen[key] = _seen.get(key, 0) + 1
+        c0 = cache[0] if cache is not None else None
+        c1 = cache[1] if cache is not None else None
+        stack = [f"{f.name}:{f.lineno}" for f in traceback.extract_stack()[-9:-1]]
+        log.info(
+            "GATEPROBE S=%s result=%s sink=%s mask=%s lengths=%s leftpad=%s conv=%s/%s rec=%s/%s static=%s in=%s/%s stack=%s",
+            S, r, gdn_sink is not None, type(mask).__name__ if mask is not None else None,
+            getattr(cache, "lengths", None) is not None, getattr(cache, "left_padding", None) is not None,
+            getattr(c0, "shape", None), getattr(c0, "dtype", None), getattr(c1, "shape", None), getattr(c1, "dtype", None),
+            pm._qwen4_decode_static_eligible(module), getattr(inputs, "dtype", None), getattr(inputs, "shape", None), stack,
+        )
+    return r
+
+
+pm._qwen4_verify_dynamic_eligible = probe
+print("[gdn_gate_probe] installed", flush=True)
--- a/docs/superpowers/plans/profiling/gdn_layer_parity.py
+++ b/docs/superpowers/plans/profiling/gdn_layer_parity.py
@@ -1,0 +1,96 @@
+"""Numeric delta of the fused Qwen4 GDN verify arm vs the stock forward on real layer weights.
+usage: .venv/bin/python gdn_layer_parity.py [model dir] [layer]  (default: bench oQ4e pack, layer 0)"""
+import json, os, sys
+import mlx.core as mx
+import mlx.nn as nn
+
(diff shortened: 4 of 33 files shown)


Rascal:
The final big benchmark is still running (estimated time: 2–2.5 hours); I'll update the PR description with the final statistics.

@moc375 If you have the time and are willing, I'd really appreciate a benchmark on the M3 Ultra

moc375:
## M3 Ultra 512GB benchmark — 5-round interleaved A/B, silent field

As requested. Full machine-exclusive (all sibling instances stopped), greedy, streaming, each arm physically tree-switched per round (checkout → engine restart → warmup → 3 iters × 4 tiers). **120 requests total, round-to-round variance <1%.**

A/B benchmark results

### Environment

| | |
|---|---|
| Hardware | M3 Ultra 512GB (unified memory) |
| Model | Qwen3.8-Flash-Next-oQ4e-mtp (125B MoE, 6B active) |
| Base A | `main` @ `9be550c` (= `25bccf5` + keepwarm) — **already includes #3520 + #3534** |
| Arm B | `a04d240` + keepwarm cherry-pick = this PR |
| Method | Interleaved: A → restart → B → restart → A × ... × 5 rounds |
| | 3 iters per tier, median per round; chart shows median-of-medians |
| Siblings | All other engine instances stopped (silent field — our earlier research showed resident siblings halve decode throughput via memory bandwidth contention) |

### Results (decode tok/s, median of 5 rounds)

| Context | main | this PR | delta | this PR range (5 rounds) |
|---|---|---|---|---|
| 1k | 164.1 | **164.5** | **+0.2%** | 164.2 – 165.0 |
| 16k | 86.5 | **86.9** | **+0.5%** | 86.8 – 87.2 |
| 32k | 82.2 | **82.8** | **+0.7%** | 82.6 – 82.9 |
| 64k | 80.6 | **81.5** | **+1.1%** | 81.4 – 81.6 |

### Observations

1. **Consistent, non-overlapping gain at 64k**: this PR wins every round at every tier (the only exception is a Round-1 cold-start artifact at 1k on main, excluded from interpretation). The chart's right panel shows the per-round scatter at 64k — the two arms never cross.

2. **Gains scale with context** (+0.2% @1k → +1.1% @64k): consistent with the gathered attention kernel's benefit growing with the attention computation that scales with context length.

3. **Smaller relative gain than M5 Max**: your +19.9% at 65k was measured against main *without* #3520. Our baseline already includes #3520's `mx.take` gathered arms (which delivered the bulk of the long-context win). This PR's Metal kernels refine the same paths further — the marginal gain on M3 Ultra 512GB is +1.1% at 64k, likely because the UMA bandwidth headroom (512GB vs 128GB) reduces the benefit of kernel-level bandwidth optimization.

4. **The absolute numbers are high** (80.6→81.5 @64k) because M3 Ultra's memory bandwidth and compute are not the bottleneck at these context sizes — the QSA gathered path (from #3520) already brought 64k decode from ~45 to ~80 on this hardware.

Repro methodology: identical prose filler (~5.1 chars/token, calibrated against the tokenizer), `stream_options.include_usage`, decode tok/s = `completion_tokens / (total - TTFT)` from client chunk timestamps, cross-checked against engine `MTP[..] tok/cycle accept` telemetry lines.

moc375:
## Follow-up: 3 content shapes (prose / code / CJK) — interleaved A/B, silent field

Following up on the earlier prose-only benchmark with the content-shape dimension your M5 Max data covers, plus a CJK scenario (relevant for Chinese-language production traffic).

Content shape A/B

**Method**: same interleaved tree-switching protocol (A→restart→B→restart ×3 rounds), same silent field, greedy. Each cell = median of 3 rounds × 3 iters. Error bars = min/max across rounds. `P=` is the engine-reported prompt token count for each shape.

### Results (decode tok/s, median of 3 rounds)

| Shape | Context | main | this PR | delta | P (tokens) |
|---|---|---|---|---|---|
| prose (EN) | ~16k | 86.5 | **86.7** | +0.2% | 17,768 |
| prose (EN) | ~64k | 80.3 | **81.4** | **+1.4%** | 70,898 |
| code (Rust) | ~16k | **83.9** | 80.1 | **−4.5%** | 16,111 |
| code (Rust) | ~64k | 79.1 | **79.7** | +0.8% | 65,786 |
| cjk (中文) | ~16k | 85.0 | **85.1** | +0.1% | 11,514 |
| cjk (中文) | ~64k | 66.0 | **66.4** | +0.6% | 45,878 |

### Cross-shape observations @64k (this PR)

| Shape | tok/s | vs prose |
|---|---|---|
| prose (EN) | 81.4 | baseline |
| code (Rust) | 79.7 | −2.1% |
| cjk (中文) | 66.4 | **−18.4%** |

1. **CJK is the heaviest shape** (66.4 vs 81.4 prose = −18%): CJK tokens carry ~2× the information density per token, so the same "context length" in tokens contains far more semantic content for the attention to process. This is the first CJK decode benchmark we're aware of on this hardware — your data covers EN prose/code; if useful, a CJK arm might be worth adding to your test matrix for Asian-language users.

2. **Code at 16k shows −4.5% on this PR** (83.9 → 80.1): the only negative delta across all 18 cells. Code's more predictable token patterns mean the draft head's acceptance rate is already very high; the NAX indexer score changes in this PR may slightly perturb the acceptance on code at short contexts. The signal is consistent across all 3 rounds (83.9/83.9/82.7 vs 80.1/80.1/80.1 — non-overlapping). Worth a look if you have a kill switch for that specific commit.

3. **The +1.4% prose gain at 64k is larger than the prose-only benchmark showed (+1.1%)** — same direction, slightly larger with the 3-round protocol.

4. **Code vs prose gap is small on M3 Ultra** (−2.1% at 64k) compared to your M5 Max data (+16% code over prose at 70.9k, from the #3469 thread). The 512GB UMA's bandwidth headroom likely narrows the code advantage.

pierre427:
I ran a separate qualification campaign around this PR's Qwen4 decode stack and opened #3626 for the RMS-compatible full-recurrence verification piece.

The source attribution is intentionally narrow: #3626 retains and credits this PR's `_qwen4_scales` helper, while the recurrence, rollback snapshots, exact-wrapper admission, and RMS ordering are the new contribution. It does not claim this PR's grouped projections, parked-head re-entry, gathered QSA, indexed attention, NAX scores, or existing fused-prework work as new.

On the qualified composed source, 18/18 fixed-length ABBA cells matched exactly. RMS-off/on median decode was:

| context | off | on | ratio | median wall time |
| 16K | 77.20 tok/s | 78.09 tok/s | 1.0115x | 13.53 → 13.49 s |
| 32K | 79.13 tok/s | 81.00 tok/s | 1.0237x | 24.87 → 25.30 s |
| 64K | 75.70 tok/s | 76.97 tok/s | 1.0167x | 48.20 → 47.79 s |

The 16K bracket passed the thermal gate. I am treating 32K and 64K as thermal diagnostics only. The same composed source completed 8/8 exact multi-turn pairs and 100/100 exact 10x10 domain pairs, with all 216 responses ending naturally and no candidate-specific qualitative regression.

The isolated #3626 branch is default-off and has CPU contract evidence only. It is based on current `main`; I can rebase it onto this branch if that integration order is preferable.

PR #3558, opened 2026-09-10
fix(mac): reject custom kernels with the wrong bundled Python ABI

PhilipJohnBasile:
The app build can compile custom kernels with a different CPython ABI from the donor app's embedded interpreter, even when the build-time import succeeds. Compare the two interpreters before compilation, then validate the native `_ext` suffixes actually staged into the bundle. Existing MLX and nanobind checks remain in place.

This is the focused app ABI follow-up requested in #2640. Validation at `ccea85b75758e5f1bd1d459c4726f847e3c0f04b`: 8 regression tests passed, covering matching and mismatched ABIs, failed interpreter queries, staged suffix mismatches and missing extensions. Full app building and signing were not run.

Please review the donor-runtime selection and the fail-closed packaging behavior.

### DIFF
--- a/apps/omlx-mac/Scripts/build.sh
+++ b/apps/omlx-mac/Scripts/build.sh
@@ -114,6 +114,7 @@ CUSTOM_KERNEL_DIRS=(
     "$REPO_ROOT/omlx/custom_kernels/minimax_m3"
     "$REPO_ROOT/omlx/custom_kernels/qwen35_prefill"
 )
+CUSTOM_KERNEL_ABI_CHECKER="$SCRIPT_DIR/check_custom_kernel_python_abi.py"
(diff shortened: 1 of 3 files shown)


PhilipJohnBasile:
@jundot, could you review this focused app ABI follow-up to #2640? The eight regression tests pass. In particular, I would appreciate your review of donor interpreter selection and the validation after staging; full packaged-app building and signing remain untested.

jundot:
Thanks for splitting this out. I verified the ABI checks against both the installed app’s Python and the local venvstacks export, and the mismatch tests pass. The guard runs before kernel cleanup and preserves the existing MLX and nanobind checks. LGTM, I’ll merge this.

PR #3559, opened 2026-09-10
fix(api): preserve and normalize thinking controls

PhilipJohnBasile:
Top-level `enable_thinking` was silently dropped by OpenAI-compatible chat requests, and a zero thinking budget could implicitly enable thinking. Normalize the top-level value into chat-template kwargs, reject contradictory controls, and only infer thinking activation from a positive budget across settings, chat completions, Anthropic messages and Responses.

This is the focused thinking-controls follow-up requested in #2640. Explicit per-request and forced model settings retain their precedence.

Validation at `a4894ff36e0e8057e262668ef28a09e752f7eeb3`: 242 API, model-settings and endpoint tests passed; three new cases failed against the unchanged base. No model-quality or performance claim is made by these routing tests.

Please review API compatibility, zero-budget behavior and the existing settings precedence.

### DIFF
--- a/omlx/api/openai_models.py
+++ b/omlx/api/openai_models.py
@@ -13,7 +13,7 @@ These models define the request and response schemas for:
 import json
 from typing import Any, Dict, List, Optional, Union
 
-from pydantic import AliasChoices, BaseModel, Field, field_validator
+from pydantic import AliasChoices, BaseModel, Field, field_validator, model_validator
 
 from omlx.api.shared_models import (
     BaseUsage,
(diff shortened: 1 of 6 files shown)


PhilipJohnBasile:
@jundot, could you review this focused thinking-controls follow-up to #2640? The 242 API/settings/endpoint tests pass. Please check the top-level alias, conflict rejection and zero-budget behavior against the intended API compatibility and settings precedence.

jundot:
Thanks for splitting this out. I verified that the top-level alias reaches the engine, conflicting controls are rejected, and a zero budget no longer implicitly enables thinking. Explicit request values and forced model settings retain their precedence across streaming and non-streaming requests. LGTM, I'll merge this.

PR #3560, opened 2026-09-10
fix(deepseek): detect nested low-bit attention overrides

PhilipJohnBasile:
DeepSeek V4 could select native ratio-128 attention when a four-bit default contained nested two-bit or three-bit quantization overrides. Traverse nested mappings and lists under both quantization keys, including `text_config`, so sub-four-bit overrides select the existing reference path. Four-bit and higher configurations retain their current selection.

This is the focused nested-quantization follow-up requested in #2640. The scope is configuration policy; it adds no artifact-specific loader or model rewrite.

Validation at `289eb2d1e811705d9124607fef09ba4d6bad787b`: 24 policy/load tests passed (120 unrelated cases deselected); twelve nested-configuration cases failed against the unchanged base. No model timing or numerical campaign was run for this patch.

Please review override traversal and fallback policy.

### DIFF
--- a/omlx/patches/deepseek_v4/utils_patch.py
+++ b/omlx/patches/deepseek_v4/utils_patch.py
@@ -50,9 +50,13 @@ def _native_ratio128_attention_enabled(config: dict[str, Any]) -> bool:
     quantizations = [config.get("quantization"), config.get("quantization_config")]
     text_config = config.get("text_config")
     if isinstance(text_config, dict):
-        quantizations.append(text_config.get("quantization_config"))
+        quantizations.extend(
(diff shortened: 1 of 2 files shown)


PhilipJohnBasile:
@jundot, could you review this focused nested-quantization follow-up to #2640? The 24 policy/load tests pass. Please check that the override traversal and selection of the existing reference path match the intended DeepSeek V4 policy; this PR makes no model-speed or quality claim.

jundot:
Thanks for separating this from #2640. I verified that nested 2/3-bit overrides now reach the existing ratio-128 attention guard, while 4-bit and higher configurations keep their current behavior. The loader and attention dispatch checks passed. LGTM, merging.

PR #3561, opened 2026-09-10
feat(vlm): serve text-only Gemma 4 with a merged MTP head through mlx-vlm

andyoneal:
> **This is an alternative to #3536, not an addition to it.** Both make a text-only Gemma 4 checkpoint with a merged MTP head actually draft; this one does it by teaching the existing mlx-vlm path to load such a checkpoint, in 536 lines instead of 1356 -- 276 against 519 counting only `omlx/` -- rather than adding a second MTP runtime on mlx-lm. Benchmarks for both are below. Draft because only one of the two should land.

A Gemma 4 checkpoint with the assistant head merged under `language_model.mtp.*` and no vision tower advertises MTP, accepts `mtp_enabled: true`, logs `Speculative backend selected: Lightning MTP` — and never drafts a token.

## What is wrong

`detect_model_type` returns `"vlm"` only when it finds a `vision_config`, `vit_config` or `mm_vision_tower`. A checkpoint with no vision tower is therefore classed text-only and served through mlx-lm, where the merged head has no binding site. Three ordinary things land there: a text-only quant, a DFlash target, and the VLM→LLM fallback.

Nothing reports it. `_is_mtp_compatible` accepts `gemma4`, so the patch is applied and the server logs its backend; then the engine looks for `.mtp`, does not find it, and decodes one token per forward for the life of the model.

mlx-vlm already drives this head (#2683). What stopped a text-only checkpoint from reaching it is the loader, and two separate things have to go right.

The Gemma 4 family ships under two model types — `gemma4` and `gemma4_unified` — and mlx-vlm picks the model class from that. A `gemma4` checkpoint gets `gemma4/gemma4.py`, which builds `VisionModel(config.vision_config)` unconditionally: that class has no text-only path at all. Only `gemma4_unified` has one, at `if config.vision_config is not None`.

That branch is then unreachable anyway, even for a `gemma4_unified` checkpoint carrying no vision, because its config declares `vision_config: Optional[VisionConfig] = field(default_factory=VisionConfig)`. An absent key still becomes a real `VisionConfig`, so the check is always true. The model builds vision embedders the shards cannot fill, and the strict `load_weights` that follows fails on the tensors nothing claims. oMLX falls back to mlx-lm — by design, since it refuses `strict=False` — which is why this never looked like a load bug.

## What this does

Two things, matching the two failures. Mark the absent modalities rather than nulling them, so the config chain leaves `vision_config` alone and `gemma4_unified`'s text-only branch becomes reachable. And point a `gemma4` checkpoint at `gemma4_unified`, the implementation that carries that branch — the two share the `language_model` layout, so this is choosing the class that can run the checkpoint rather than asking anyone to relabel it on disk. Audio keeps its existing treatment: nothing dereferences `audio_config` as a dict, so `None` is enough there.

Then route a text-only `gemma4` carrying a merged head to the VLM engine, which can drive it. The reverse rule keeps a text-only checkpoint with *no* head on mlx-lm, which carries less overhead — measured at +0.31 GB on the 12B — rather than letting it arrive there by way of a failed load.

`model_type` moves with `engine_type` in both directions, because `supports_images` is `model_type == "vlm"`: moving only the engine would leave a text-only checkpoint on the LLM engine still telling clients it accepts images. `engine_pool`'s runtime VLM→LLM fallback sets the pair together for the same reason.

The second commit is independent of all of this and worth taking either way — see below.

## Benchmarks

`gemma-4-31B-it-oQ4-fp16-mtp` — 19.0 GB, `model_type` gemma4, no vision sub-config, merged head present. M1 Max / 32 GB, mlx 0.32.2. Driven through the admin benchmark so the numbers reproduce from the UI: bundled `code_python` corpus, `temperature=0.0`, `warmup_mode=quick`.

Four arm types, since "which implementation" and "does the feature pay at all" are different questions. Arms are interleaved rather than blocked, because this host throttles under sustained load; each arm is a fresh server on a freshly checked-out tree; medians are `statistics.median`. An arm counts only if `completion_tokens` matches the request, discovery reports the intended engine, and — with MTP on — the log shows drafting.

**1024-token prompt, 512 generated, n=4 per arm:**

| arm | median tok/s | mean | range |
|---|---|---|---|
| mlx-vlm, MTP on (this PR) | 17.65 | 17.50 | 14.5–20.2 |
| mlx-vlm, MTP off | 14.30 | 14.22 | 13.2–15.1 |
| mlx-lm, MTP on (#3536) | 19.90 | 19.72 | 18.9–20.2 |
| mlx-lm, MTP off | 15.00 | 14.80 | 14.1–15.1 |

MTP pays **+23.4%** here on this path (mean +23.0%), and +32.7% on #3536's. Between the two implementations, mlx-vlm/mlx-lm is **0.887** — this path is ~13% slower with drafting on, and level with it off (1.007 over six arms per side, and 1.000 at a second prompt size). So the difference is in how the head is driven, not in engine overhead. Peak memory is identical: 21.07 GB with MTP on either path, 19.65 GB with it off.

## `perf(mtp)`: the head runs in float32 today

The second commit is a fix to code that already ships and is independent of everything above.

The head is bfloat16 and `oq.combine_gemma4_assistant_mtp` preserves that, so on a float16 checkpoint every head matmul meets float16 activations and MLX promotes the result to float32. The 31B oQ4 quant is that shape — 1243 float16 backbone tensors against 48 bfloat16 head tensors — and it is visible only in the safetensors headers, since the config reports bfloat16 for the whole file.

Measured on that checkpoint, the head forward cost 2355–2744 ms per request before, against 509–787 ms for the same head driven from mlx-lm, which aligns the dtype at load. Casting is safe in either direction because the head only proposes: every token the engine emits is one the backbone verified, so head precision moves the acceptance rate and can never move the output.

It is done at the first draft rather than in `sanitize`, because mlx-vlm skips model sanitizers when safetensors metadata says `format=mlx` — which is exactly these checkpoints.

Happy to split this into its own PR if you'd rather take it independently of the routing change.

## Reproducing

Any Gemma 4 checkpoint with the assistant head merged and no vision tower — `text_config.mtp_assistant_config` in `config.json` plus `language_model.mtp.*` in the safetensors index.

Before this change the model is discovered as `type: llm, engine: batched` and never drafts. After it, discovery reports `type: vlm, engine: vlm` and the engine's MTP telemetry line reports an accept rate.

## Tests

`tests/test_gemma4_text_only_vlm.py` — the routing predicates in both directions, the loader behavior on a marked-absent modality, and four field-pair invariants driven through `discover_models`, so a future change that moves `engine_type` without `model_type` fails rather than quietly advertising images.

`tests/test_vlm_audio_fallback.py` gains coverage that the existing orphaned-audio path still nulls rather than marking, which the first draft of this broke.

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -565,6 +565,18 @@ _AUDIO_CONFIG_KEYS = (
 )
 
 
+# Model types whose Gemma 4 weight prefixes are known well enough to conclude
+# a modality is absent. Other families keep the audio-only behavior, because a
+# prefix this list does not know would read as "no vision" and disable it.
+_GEMMA4_FAMILY_TYPES = ("gemma4", "gemma4_unified")
+
+# Prefixes that carry Gemma 4 vision weights. "unified" checkpoints have no
+# vision tower at all -- the single transformer does the work, and vision needs
+# only a patch embedder and a projection -- so the embedder prefixes are as
+# load-bearing as the tower one.
+_GEMMA4_VISION_PREFIXES = ("vision_tower.", "embed_vision.", "vision_embedder.")
+
+
 def _resolve_optiq_vision_sidecar(model_dir: Path) -> Path | None:
     """Resolve the OptiQ multimodal sidecar declared by ``config.json``."""
     config_path = model_dir / "config.json"
@@ -596,8 +608,23 @@ def _resolve_optiq_vision_sidecar(model_dir: Path) -> Path | None:
     return sidecar
 
 
+def _is_gemma4_family(config: dict) -> bool:
+    model_type = str(config.get("model_type") or "").lower().replace("-", "_")
+    return model_type in _GEMMA4_FAMILY_TYPES
+
+
+def _has_vision_weights(model_dir: Path) -> bool:
+    """Return True iff any shard carries Gemma 4 vision tower/embedder keys."""
+    return _has_prefixed_weights(model_dir, _GEMMA4_VISION_PREFIXES)
+
+
 def _has_audio_weights(model_dir: Path) -> bool:
     """Return True iff any safetensors shard contains audio_tower / embed_audio keys."""
+    return _has_prefixed_weights(model_dir, ("audio_tower.", "embed_audio."))
+
+
+def _has_prefixed_weights(model_dir: Path, prefixes: tuple[str, ...]) -> bool:
+    """Return True iff any safetensors shard carries a key with one of ``prefixes``."""
     import safetensors
 
     weight_files = list(model_dir.glob("*.safetensors"))
@@ -609,12 +636,13 @@ def _has_audio_weights(model_dir: Path) -> bool:
         try:
             with safetensors.safe_open(str(sf), framework="np") as f:
                 for k in f.keys():
-                    if k.startswith(("audio_tower.", "embed_audio.")):
+                    if k.startswith(prefixes):
                         return True
         except Exception:
-            # Corrupt or unreadable shard — treat as no audio info, let
-            # downstream loader produce its own error.
-            return False
+                # Skip the shard rather than answer for the whole checkpoint:
+                # one unreadable shard would otherwise drop a real modality.
+            logger.debug("could not scan %s for %s", sf.name, prefixes)
+            continue
     return False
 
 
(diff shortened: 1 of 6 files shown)


jundot (approved):
Reusing the existing VLM MTP runtime is the better maintenance tradeoff here. I verified text-only checkpoint loading and a draft forward with a small Gemma 4 model. I'll follow up with two small guards: preserve VLM routing for audio-capable unified checkpoints, and reject image input on text-only checkpoints before streaming starts. LGTM with those follow-ups.

PR #3564, opened 2026-09-10
feat(admin): configurable Status dashboard and time-window stats

zviratko:
## What this change does

Make the Status dashboard layout configurable. Panels can sit side by side instead of in one long column. Add 12h and 24h serving-statistics views.

Configurable Status dashboard

Layout controls

## Layout and interaction

- Select 1–4 columns. Narrow screens use one column.
- Each panel has Normal, Wide (two columns), and Full width modes. Wide and Full have the same width with two columns. The button icon and tooltip indicate the next mode.
- Set the maximum page width from 1200 to 2560 CSS pixels. The slider previews the value while moving. Width and placement apply together when the value is committed.
- Drag from the grip handle. Panels stay in their original slots during the gesture. A blue line shows insertion before or after a target. Reorder once on release. The marker shows order, not an exact final masonry cell.
- Escape, pointer cancellation, window blur, hidden document, viewport resize, and an outside drop cancel the drag. Temporary height limits are removed on exit.
- Store panel order, width modes, column count and page width in browser localStorage only. No layout preferences go to the server. Preferences are shared by tabs on the same origin, not per window.
- Reset layout restores defaults without reloading. The previous fixed-zone arrangement is not migrated to the new layout schema.
- Put Clear directly after the statistics scope switch. Hide it for the time-window views.
- Keep the separate token heatmap, the Usage History store sharing, and the header cleanup from the original PR.

## Time-window statistics

This is not only a frontend change:

- `UsageHistory` gains 12h/24h query periods over existing hourly SQLite aggregates. No request content or new per-request history is stored.
- `ServerMetrics` maps those aggregates to the existing statistics shape, including per-model filtering and weighted speed calculations.
- Include the oldest overlapping hourly bucket in full. Coverage can exceed the requested period by less than one hour. These are not exact sliding request-time windows.
- Recording must be enabled and storage available. Missing periods cannot be reconstructed. The response marks unavailable history; the frontend currently also shows zero-valued placeholders with its warning.
- Run rolling-scope snapshot queries on a worker thread, not the async serving event loop.
- Refresh a window on selection and periodically while selected. Discard responses for an outdated model filter.
- Session and All-Time remain separate counters. The existing Usage History range selector is unchanged.

## Follow-up fixes (active-models card)

- Add `overflow-hidden` to the Active Models card body. Without it the footer strip painted a square background over the rounded bottom corners.
- Observe every card, not only the grid container, with the ResizeObserver. With fixed row-spans a card that grows or shrinks inside its span does not resize the grid, so the old observer missed exactly the height changes that need re-placement. Activity sub-rows appearing or vanishing at the 500 ms poll previously overlapped, or gapped over, the card below.
- Finished generating/activity rows now linger 8 s as dimmed ghosts with a Finished label (new i18n key, all nine locales) instead of dropping instantly, so the card height is stable at request end. The existing per-model idle line is unchanged.

## Localization

Layout and scope labels are present in all nine locale files. English locale caching and Python changes require a server restart; templates and static assets load on page reload.

## Verification for this update

- `pytest tests/test_dashboard_rolling_stats.py tests/test_usage_history.py tests/test_server_metrics.py tests/test_admin_usage.py -q`: **76 passed**.
- `node --test tests/usage_history_ui.test.cjs`: **5 passed**.
- `tests/dashboard_layout_release.test.cjs` with jsdom 30.0.1: **15 checks passed** after the follow-up fixes. Uses mocked geometry; this does not prove visual smoothness or cross-browser compatibility. Setup is documented in `tests/dashboard_layout_TESTING.md`.
- JavaScript syntax and `git diff --check` passed.
- User reports the final apply-on-release interaction looks good. Screenshots are user-supplied and approved for publication. The corner and reflow fixes are awaiting live browser confirmation by the author.
- Frontend deployment was verified by HTTP byte comparison without restarting oMLX. The final backend boundary/availability corrections in this PR were tested locally, not deployed to that running instance.

The original full-suite totals were for the earlier two-column revision and do not establish this version's status. The full suite was not rerun for this update. Independent automated review could not run because its model endpoint returned HTTP 404.

## Attribution

Original work attribution retained at the author's request:

Final interaction corrections, tests, and PR wrap-up: made under human guidelines by openai/gpt-6-astra-flex via Nous.


zviratko:
In case someone wants to try it (for me it makes the web interface so much better), here's a standalone patch.

Apply with:
```shell
patch -p1 -d "$(readlink -f /opt/homebrew/opt/omlx)//libexec/lib/python3.11/site-packages/omlx/admin" < omlx-dashboard-layout.patch
```

[omlx-dashboard-layout.patch](https://github.com/user-attachments/files/32076095/omlx-dashboard-layout.patch)

jundot:
Thanks for implementing this. I tested the new layout, and being able to rearrange the panels and see more information at once makes the dashboard much more useful. On my 16:9 monitor, though, the wider page feels a little too spread out. This may just be a matter of personal preference, but please give me a little more time to think about making this the default layout.

zviratko:
Sure, let me know if you'd like something changed, I can try making it configurable? Or perhaps 3 columns instead of 2 when the monitor is wide enough? More than happy to spend some free tokens for you :-D 

And my thinking is that you can always ^+⌥+← Safari :) At least now you have the options -  with classic noodle-y design, you dont.

zviratko:
PR updated, standalone patch is attached, apply with

```shell
patch -p1 -d "$(readlink -f /opt/homebrew/opt/omlx)//libexec/lib/python3.11/site-packages/omlx/" < omlx-dashboard-pr3564.patch
```
[omlx-dashboard-pr3564.patch](https://github.com/user-attachments/files/32145317/omlx-dashboard-pr3564.patch)

LXD-8:
Coming over from #3584 (the Active Models reorder), where you told me you'd updated this — thanks for the nudge to try it, and for taking on the columns + page-width controls we talked about. Good to see them land as real settings instead of one hard-coded two-column default.

I've been running it locally and reading through the implementation, and I think the direction is right. Layout preference is exactly the kind of thing that should be per-user: a 16:9 desktop and a laptop screen want very different densities, and forcing one arrangement is what made the old fixed zones feel cramped. What I like most:

- Width modes are derived from order + column count rather than persisted cell coordinates, so any column count just works. That's a much better shape than saving absolute grid positions.
- Geometry is frozen during the gesture and the DOM moves once on release — that's the right call, and it removes the jitter most drag implementations have.
- Keeping layout in localStorage only is a defensible trade-off, and you documented it clearly rather than leaving it implicit.

I'd rather see this merge as-is than have anything else folded in — the minimal version is the mergeable version. Hope it lands soon, and happy to test any follow-up.

zviratko:
I'm not a (web) dev, so I didn't want to make excessive changes, so

1. I think this can be a next step, or maybe someone (even me) can create a new fancy UI with floating number clusters or something :) I wanted to make my changes minimalistic

2. same as above, personally I'm happy that it works as well as it does for now

3. aware of this - same as above, I think this would be a more invasive redesign than I wanted to push

4. valid point, no idea if the dashboard is accessible otherwise and I didn't bother to check

I'd like to keep my PR as-is for now, should be simple to merge and I hope a clear improvement. I can follow up with more ambitious changes if @jundot wants, but I feel there are enough PRs already and I don't want to add another stale one.

LXD-8:
@zviratko Fair enough — I've edited my earlier comment down and pulled the suggestions out. Those were me thinking out loud about where this could go next, not conditions for merging, and you were right to keep the scope tight.

For what it's worth, I think this is good work — the layout engine is the hard part, and it works. Thanks for putting the time into it.

Let's leave it with @jundot then and see what he says. I'll be watching for the answer along with you — hoping this lands.

zviratko:
for your pleasure, mock-up of a "bigger" redesign
made in accordance with 1975 NASA design philosophy documentation :-D

Screenshot 2026-09-13 at 02 26 30

LXD-8:
Haha, I love it. :-D
This is definitely what happens when we stop caring about “minimal” and embrace the full NASA philosophy.

zviratko:
PR updated, just slight cosmetic fixes for reflow bugs I noticed

patch for easy testing attached, apply with 

```shell
patch -p1 -d "$(readlink -f /opt/homebrew/opt/omlx)//libexec/lib/python3.11/site-packages/" < omlx-keg-ddc47184-dashboard-v3.patch
```
[omlx-keg-ddc47184-dashboard-v3.patch](https://github.com/user-attachments/files/32212753/omlx-keg-ddc47184-dashboard-v3.patch)

zviratko:
@jundot I will keep updating this (for myself) and I am working on a larger redesign (that might never land), let me know if you don't want this open or if you need something to merge it

jundot:
@zviratko I ended up building this a different way in #3694: a free grid with an explicit edit mode, like Metabase dashboards, saved in the server settings. It's a design preference, not a problem with your PR. Yours is what got this moving, so thanks for pushing on it.

The 12h/24h stats look useful on their own. If you split them into a separate PR without the layout changes, I'll take a look.

I'll close this once #3694 is merged.

zviratko:
@jundot cool! I bet your change is better than mine :)

And I want to ask you - I started doing a full "uplift" of the dashboard, mostly as a PoC for myself and to test my agent and now I'm thinking if I should continue with it?
It's a redesign of the dashboard, different aesthetic, better model profile and preset management and some added functionality (metrics, graphs, list of events, so far contained to what data is already there but I wanted to extend it to get a list of running requests with bodies etc.), lots of quality of life stuff like showing changed values and list of changes, dual-commit restart of the server (save settings and then restart). But of course it's made from the point of what _I_ want and the current state is not polished yet.
Screenshots attached and the branch is there https://github.com/jundot/omlx/compare/main...zviratko:omlx:feat/uplift-dashboard. Coded with my oversight but I'm not a sysadmin, not a dev, so vibe it is...
Should I pursue it or drop it? If implemented as I want, it would be a standalone thing independent on the current one, so if someone runs it (or you merge it), they can come back. Core modifications only in routes and for adding some i18n strings, then - if we want - some metrics and their hooks (aiming for backwards-compatible sqlite schema changes if needed).
I am aware of the added burden of maintaining both and that's not what I'm asking, just take a look and if there's any chance you like it then I will try getting it to feature-parity, test as well as I can and will come back in a month or whatever with a PR :) if not, I might still keep it as a plaything depending on how it goes...

Thanks!

Screenshot 2026-09-16 at 09 22 47
Screenshot 2026-09-16 at 09 22 59
Screenshot 2026-09-16 at 09 23 06
Screenshot 2026-09-16 at 09 23 29
Screenshot 2026-09-16 at 09 23 48

zviratko:
Some more screenshots for your pleasure
Model settings with visible from->to changes, config summary
Screenshot 2026-09-16 at 12 24 04

Changing a preset from a template/preset or another model's settings
Screenshot 2026-09-16 at 12 24 54

Server settings with from->to changes, indication of restart for sections, list of changes to config and dual-commit save&restart. 
Screenshot 2026-09-16 at 12 26 57

PR #3568, opened 2026-09-10
fix(server): serialize non-API validation errors with jsonable_encoder

### LINKED ISSUE #3387: Admin API returns 500 instead of 422 when a Pydantic validator raises ValueError (server.py:780)
Any request to a non-`/v1/` route that fails a Pydantic validator raising `ValueError` returns **HTTP 500** instead of the intended **422**. The whole admin API is affected.

## What happens

```
WARNING - PUT /admin/api/models/<id>/settings → 422: [{'type': 'value_error', ...,
          'ctx': {'error': ValueError('...')}}]
ERROR   - PUT /admin/api/models/<id>/settings → 500 (unhandled):
          Object of type ValueError is not JSON serializable
```

The handler logs the 422 correctly, then dies building the response body. The client sees a 500 with no useful detail.

## Root cause

`omlx/server.py:758-781` `validation_exception_handler`:

```python
if _is_api_route(request):
    ...                                  # builds plain strings — safe
else:
    content = {"detail": exc.errors()}   # 780
return JSONResponse(status_code=422, content=content)   # 781
```

`_is_api_route` is `request.url.path.startswith("/v1/")` (`server.py:707-718`), so **every non-`/v1/` route** — all of `/admin/api/...` — takes the `else` branch. Pydantic v2 puts a **raw `ValueError` instance** into `ctx` for validators that raise `ValueError`, and `JSONResponse` cannot `json.dumps` it.

The `/v1/` branch is immune only incidentally: it flattens errors into strings and never serialises `ctx`.

## Minimal reproduction — no server needed

```python
import json
from pydantic import BaseModel, field_validator, ValidationError

class Req(BaseModel):
    base_url: str
    @field_validator("base_url")
    @classmethod
    def _v(cls, v: str) -> str:
        if not v.startswith(("http://", "https://")):
            raise ValueError("base_url must start with http:// or https://")
        return v

try:
    Req(base_url="ftp://x")
except ValidationError as e:
    errs = e.errors()
    print(errs[0]["ctx"])          # {'error': ValueError('base_url must start with ...')}
    json.dumps({"detail": errs})   # TypeError — exactly what server.py:780-781 does
```

Output:

```
{'error': ValueError('base_url must start with http:// or https://')}
TypeError: Object of type ValueError is not JSON serializable
```

That validator is real: `omlx/admin/external_api.py:70-75`.

## Blast radius

Every `ValueError`-raising validator on a non-`/v1/` request model. Grep of validator-attached raises under `omlx/admin/`:

| File | count |
|---|---|
| `external_api.py` | 3 |
| `accuracy_benchmark.py` | 2 |
| `ane_tuning.py` | 2 |
| `benchmark.py` | 2 |
| `context_benchmark.py` | 1 |

Type errors are unaffected — `{"prefill_step_size": "abc"}` returns a correct 422, because a type failure carries no `ValueError` in `ctx`. Only custom validators hit this, which is why it has gone unnoticed: the failure needs a *semantically* invalid value of the *right* type.

## Suggested fix

One line at `server.py:780`:

```python
from fastapi.encoders import jsonable_encoder
...
    content = {"detail": jsonable_encoder(exc.errors())}
```

Verified: `json.dumps(jsonable_encoder({"detail": errs}))` succeeds on the repro above. `exc.errors(include_context=False)` also works but drops context callers may want.

A regression test worth having: assert `json.dumps({"detail": exc.errors()})` succeeds for a model whose validator raises `ValueError` — that is the actual bug class, and it is cheap to guard.

## How this was found

While field-testing #3381 / #3384. That PR adds a per-model `prefill_step_size` with an allowlist validator; the documented 422 contract turned out to return 500 on real hardware. I worked around it there by typing the field as `Literal[256, 512, 1024, 2048] | None`, whose `ctx` is `{'expected': '256, 512, 1024 or 2048'}` and serialises fine — but that is a workaround at the call site, not a fix for the handler.

`server.py` is untouched by that PR. Filing separately since the defect predates it and affects unrelated endpoints. Happy to send a PR for the handler fix plus the regression test if you'd like.

arnavprabhu:
### Problem

Requests to non-/v1/ routes (the whole admin API) that fail a Pydantic validator raising ValueError return HTTP 500 instead of 422. The handler logs the 422 correctly, then dies building the response body: pydantic v2 puts the raw ValueError instance into ctx, and JSONResponse cannot serialize it ("Object of type ValueError is not JSON serializable"). The client sees a 500 with no useful detail.

### Change

One line in validation_exception_handler: pass exc.errors() through jsonable_encoder before handing it to JSONResponse, matching the fix suggested in the issue. The /v1/ branch is untouched (it already flattens errors to strings), and error content is preserved — the ValueError encodes to {} under its ctx key rather than dropping the entry.

### Tests

- New regression test in tests/test_server.py drives the handler with a ValueError-carrying error list shaped like the real omlx/admin/external_api.py base_url validator: fails before (TypeError: Object of type ValueError is not JSON serializable), passes after with a 422 body.
- tests/test_server.py, tests/test_external_api.py, tests/test_api_auth.py: 172 passed.

Fixes #3387

### DIFF
--- a/omlx/server.py
+++ b/omlx/server.py
@@ -55,6 +55,7 @@ from typing import Optional, Union
 
 from fastapi import Depends, FastAPI, HTTPException, Response
 from fastapi import Request as FastAPIRequest
+from fastapi.encoders import jsonable_encoder
 from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse, RedirectResponse, StreamingResponse
@@ -893,7 +894,7 @@ async def validation_exception_handler(
         param = errors[0].get("loc", [None])[-1] if errors else None
         content = _openai_error_body(detail_str, 422, param=param)
     else:
-        content = {"detail": exc.errors()}
+        content = {"detail": jsonable_encoder(exc.errors())}
     return JSONResponse(status_code=422, content=content)
 
 
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests for omlx.server module - sampling parameter resolution and exception handlers."""
 
+import asyncio
 from types import SimpleNamespace
 from unittest.mock import AsyncMock, MagicMock, patch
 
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for fixing this! LGTM

PR #3570, opened 2026-09-11
fix(glm5_next): remap quantized forget-gate scales/biases in sanitize

### LINKED ISSUE #3400: glm5_next: affine-quantized checkpoints fail to load — sanitize misses forget_gate .scales/.biases (136 unexpected params)
## Summary

`ai2apps/GLM-5.3-Flash-MLX-4bit-MTP` (43 shards, 181.7 GB, complete + safetensors-verified) fails to load on oMLX **0.6.4** with:

```
LM load failed (force_lm=True): Model type glm5_next not supported.;
VLM fallback also failed: Received 136 parameters not in model:
language_model.model.layers.0.self_attn.f_a_proj.biases,
language_model.model.layers.0.self_attn.f_a_proj.scales, ...
```

Every subsequent load attempt is then served from the cached failure ("Skipping load retry ... after cached failure") until the server is restarted.

## Root cause

`omlx/patches/mlx_vlm_glm5_next_compat/vendor/mlx_vlm/models/glm5_next/language.py` → `LanguageModel.sanitize()` maps the KDA forget-gate tensors into the `forget_gate.` submodule, but the part list only knows unquantised weight names:

```python
fg_parts = ("A_log", "dt_bias", "f_a_proj.weight", "f_b_proj.weight")
```

An affine 4-bit checkpoint carries `.scales` / `.biases` sidecars for each projection. Those 136 tensors (34 linear-attention layers x 2 projections x 2 sidecars) keep their flat `self_attn.f_*_proj.scales|biases` names and are rejected as "not in model".

Same gap exists in the upstream mlx-vlm GLM5 Next implementation (PR #2030, `fa27a9a`), which the vendor copy is based on.

## Verified fix (one line)

```python
fg_parts = ("A_log", "dt_bias",
            "f_a_proj.weight", "f_a_proj.scales", "f_a_proj.biases",
            "f_b_proj.weight", "f_b_proj.scales", "f_b_proj.biases")
```

Applied in-place to the vendor file on Mac Studio M3 Ultra (256 GB) + server restart:

- sanitize renames all 136 sidecars, load succeeds: `VLMBatchedEngine loaded ... actual: 166.03 GB`
- real inference works: chat completion 100 tok / 10.3 s (~9.7 tok/s, no MTP — as expected, the compat layer drops `mtp.*`)
- this checkpoint needs **no other change**: its vision tower uses `model.visual.*`, which the vendor `Model.sanitize` already remaps

## Context

- Cross-ref #3380 "Failure 2" independently diagnosed the identical fix (for the Vontra oQ checkpoint, which additionally needs the `vision_tower` prefix + heterogeneous-expert workarounds). This issue is the clean, single-patch case: stock 4-bit affine MTP checkpoints like ai2apps' load fully with just the `fg_parts` completion.
- Happy to PR the one-liner into the vendor tree if that's the fastest path.

Environment: oMLX 0.6.4 · mlx 0.32.0 · mlx-vlm 0.6.3 + vendor glm5_next compat · macOS 26 · M3 Ultra 256 GB

arnavprabhu:
### Problem

Affine-quantized GLM-5.3-Flash checkpoints (e.g. ai2apps/GLM-5.3-Flash-MLX-4bit-MTP) fail to load with 'Received 136 parameters not in model' for f_a_proj/f_b_proj .scales/.biases. sanitize() only remapped the float .weight names into the forget_gate submodule, so the 34 KDA layers x 2 projections x 2 sidecars kept their flat self_attn names and tripped strict load. Fixes #3400.

### Change

Extend fg_parts with the .scales/.biases names so the sidecars remap alongside the weights. One-line-class fix, same loop, no behavior change for float checkpoints.

### Tests

New test_sanitize_remaps_quantized_forget_gate_sidecars fails before, passes after (verified via stash). tests/test_mlx_vlm_glm5_next_compat.py: 15 passed, 8 skipped. tests/test_oq.py + tests/test_vlm_engine.py failure sets identical before/after (20 pre-existing env failures, unrelated).

### DIFF
--- a/omlx/patches/mlx_vlm_glm5_next_compat/vendor/mlx_vlm/models/glm5_next/language.py
+++ b/omlx/patches/mlx_vlm_glm5_next_compat/vendor/mlx_vlm/models/glm5_next/language.py
@@ -933,7 +933,16 @@ class LanguageModel(nn.Module):
 
         remapped = {}
         conv_parts = {}
-        fg_parts = ("A_log", "dt_bias", "f_a_proj.weight", "f_b_proj.weight")
+        fg_parts = (
+            "A_log",
+            "dt_bias",
+            "f_a_proj.weight",
+            "f_a_proj.scales",
+            "f_a_proj.biases",
+            "f_b_proj.weight",
+            "f_b_proj.scales",
+            "f_b_proj.biases",
+        )
         for k, v in weights.items():
             nk = k.replace(".hc_attn_", ".attn_hc.").replace(".hc_ffn_", ".ffn_hc.")
 
--- a/tests/test_mlx_vlm_glm5_next_compat.py
+++ b/tests/test_mlx_vlm_glm5_next_compat.py
@@ -492,6 +492,31 @@ def test_sanitize_and_oq_keep_sensitive_parameters_in_fp32():
     ) == {"bits": 8, "group_size": 64, "mode": "affine"}
 
 
+def test_sanitize_remaps_quantized_forget_gate_sidecars():
+    from mlx_vlm.models.glm5_next.language import LanguageModel
+
+    config = _tiny_config()
+    model = LanguageModel(config.text_config, config)
+    prefix = "language_model.model.layers.0.self_attn."
+    weights = {
+        prefix + "f_a_proj.weight": mx.ones((32, 32)),
+        prefix + "f_a_proj.scales": mx.ones((2,), dtype=mx.bfloat16),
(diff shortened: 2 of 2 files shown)


jundot:
Thanks for the fix. The forget-gate sidecar mappings correctly address the strict-load failure, and the related GLM5 Next, oQ, and VLM engine tests pass locally. The CI failures are unrelated to this change and have already been fixed on the latest `main`. LGTM, merging.

PR #3577, opened 2026-09-11
fix(cluster): direct-first control transport and venv-aware system proxy

xunlinkx:
On macOS 26 (Tahoe), Apple sandbox entitlements on `/usr/bin/python3` silently drop outbound socket connections when invoked non-interactively or from background daemons over Thunderbolt bridge interfaces (`10.0.0.x`), hanging rank coordination indefinitely. The active virtual environment Python (`sys.executable`) possesses standard local network entitlements and connects in under 1 ms.

Changes:
- Ranks in `auto` mode now attempt a direct TCP connection to the coordinator first. Only if socket creation fails with permission/network error (`EACCES`/`EPERM`/`TimeoutError`) does it fall back to the system proxy.
- `_system_python()` in `system_socket_proxy.py` now honors `sys.executable` before falling back to `/usr/bin/python3`.
- Fixes `import os` scope in `control_plane.py`.

Validation:
- Tested on dual-Mac M4 cluster over Thunderbolt 5 bridge (`10.0.0.2` -> `10.0.0.1`). Direct connection established in <1ms without proxy hang.
- `tests/test_system_socket_proxy.py` passes.

### DIFF
--- a/omlx/cluster/control_plane.py
+++ b/omlx/cluster/control_plane.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 
 import hashlib
 import hmac
+import os
+import logging
 import pickle
 import secrets
 import socket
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for testing this on your Thunderbolt setup. Trying the direct connection first makes sense. I reproduced a case where direct retries consume the coordinator’s entire connection window, so I’ll follow up after merging with earlier proxy fallback and preserve the system Python escape path.

PR #3584, opened 2026-09-11
refactor(app): show Active Now above Usage History in the macOS status screen

LXD-8:
**Base:** `main` @ `9c748580` (main's version string is now 0.7.0.dev3, not yet released; the latest release is still v0.7.0.dev2)

## Why

The Status screen's **Active Now** panel answers "what is the server doing right now" — which models are resident, what is prefilling or generating, how many requests are queued. **Usage History** answers "what already happened". The live state is what an operator opens this screen to check, and it is the only one of the pair that can still change; ordering history first buries it under a list that only grows.

## Scope: macOS app only

`apps/omlx-mac/Sources/AppView/Screens/StatusScreen.swift`:

- the `ActiveNowList` section moves above `UsageHistoryView()`;
- the file's header comment, which lists the screen's sections in order, is updated to match.

Pure reordering — no strings, data or logic change. Both blocks are self-spacing (`SectionHeader` supplies its own top gap; `UsageHistoryView` carries its own header).

## Why the web half is gone

This PR used to also reorder two `{% include %}` positions in `omlx/admin/templates/dashboard/_status.html`. That file no longer works that way. **#3694** (`feat(admin): customizable dashboard block layout`, merged 2026-09-16 as `22059e8a`) rebuilt the Status tab as a GridStack board of user-arranged blocks, seeded from `DashboardLayout.defaultLayout()` in `omlx/admin/static/js/dashboard_layout.js`. The order is now a per-user preference — a fixed-order change there would be re-litigating a decision the dashboard just handed to the user — and the include positions this PR edited do not exist any more.

The macOS app has **no** layout control, so its fixed order is the only order a mac user gets. That half stands on its own; this is also the gap @zviratko pointed at on this PR ("This was the whole reason why I did #3564 :) but I didn't touch the Swift app").

## Verification

- `git diff` against the base is **1 file, 3 insertions / 3 deletions**, with no `_status.html` change.
- The Swift hunk is byte-identical to the revision reviewed before the rebase (blob `15eb3073` → `b98300d4`).
- Rebased onto `main` @ `9c748580` with no conflicts; upstream has not touched `StatusScreen.swift` since.
- Merges cleanly into current `main` (`5941f683`), per GitHub and a local `git merge-tree` check.

## Behaviour delta

- macOS Status screen: Active Now renders above Usage History.
- Nothing else changes — no strings, no data, no logic, no web template.


zviratko:
This was the whole reason why I did https://github.com/jundot/omlx/pull/3564 :) but I didn't touch the Swift app.

LXD-8:
Really like #3564 — two columns plus drag-to-rearrange makes the Status tab readable at a glance, and it's the same motivation behind this PR: the live state shouldn't require scrolling past historical aggregates. Three small suggestions, in case they're useful.

**1. Make the column count a user choice, next to the theme control.** The navbar's right-hand group already has a round icon button for the theme (`sun-moon` / `sun` / `moon`, `p-2 rounded-full`); a matching single/double-column toggle beside it would let people switch without a config file. That also answers jundot's hesitation about the wider layout as the *default* — it doesn't have to be a global decision if the user can pick. The flag can live in the same localStorage record you already keep for panel positions, "Reset layout" should reset it too, and the ≥1024 px rule becomes the initial value rather than the rule.

**2. Single column: keep Active Models above Usage History.** Your `_status.html` already does this — the usage cards are in `data-zone="below"`, so in single-column they come after Active Models. That's exactly the order this PR argues for, so the two changes agree. Worth a small test pinning the default order though: with drag-to-rearrange any order is reachable, and a later reshuffle could invert the default without anyone noticing.

**3. Cap the shell width.** The "too spread out" feeling on a 16:9 monitor is mostly the 1740 / 1980 / 2560 px steps — past a point the columns just drift apart and the line lengths get long. A cap (somewhere around 1600–1740 px) with the shell centred turns the extra width into margin instead of stretch, and keeps the Status tab looking like the rest of the admin, which stays at `max-w-7xl`.

One coordination note: this PR touches `_status.html` as well — it moves the Active Models card directly under Average Speed. Your `data-zone="top"` order lands in the same place, so whichever of the two merges second is a trivial rebase; happy to rebase mine on yours if that's easier.

Also glad to supply the `zh` values for the three new keys (`status.reset_layout`, `status.reset_layout_tooltip`, `status.drag_handle_tooltip`) if that helps — in the repo's `normalize_i18n.py` flow the other locales fall back to English, so nothing else is needed.

zviratko:
Scope creep :)
Ok, I'm on it. The layout will be dynamic with width limit presets. Let's see if my Qwen can do it :)

zviratko:
I updated the PR, try if you like it :)

LXD-8:
Hi @jundot, would you mind taking a look at this one when you have a moment? No rush at all.

I checked it against current `main` (`5941f683`) and it merges cleanly.

jundot (approved):
Moving Active Now ahead of Usage History makes the live model state easier to reach. I checked that the polling lifecycle and both sections are unchanged apart from their order. LGTM, merging.

PR #3588, opened 2026-09-11
fix(qwen35_ane): drain the hybrid merge before the next tile reuses the ANE output surface

beaglemoo:
The ANE hybrid merge kernel reads the ANE program's single output IOSurface, but MLX commits the enclosing command buffer lazily. On hosts whose prefill chunk is larger than the 2048-token ANE tile, the next tile dispatches to the same program and its ANE evaluation can overwrite the surface before the previous merge has executed. `begin()` only orders against the previous evaluation, not the previous merge, so nothing prevents the race.

This commits the merge command buffer and waits for it at each of the three hybrid merge sites. The ANE and qmm branches were already joined on the host at that point, so the only added latency is the merge kernel itself. A failed command buffer now raises instead of silently producing a corrupted tile.

Validation on an M5 Pro 64 GB, macOS 27, MLX 0.32.2:

- `tests/test_qwen35_ane_prefill.py`: 143 passed.
- Qwen3.8-27B oQ4e with `qwen35_ane_prefill_enabled` and a 2048 tile: 3461, 5157 and 6852-token prompts (two to four tiles per chunk) prefilled through the ANE/GPU hybrid path with no command buffer errors, and the greedy output matched the GPU-only path.

### DIFF
--- a/omlx/custom_kernels/qwen35_prefill/csrc/qwen35_ane.mm
+++ b/omlx/custom_kernels/qwen35_prefill/csrc/qwen35_ane.mm
@@ -2655,6 +2655,27 @@ public:
     }
     encoder.dispatch_threads(MTL::Size(output_n, M, 1),
                              MTL::Size(16, 16, 1));
+    // The merge encoder reads this program's single output IOSurface. MLX
+    // commits this command buffer lazily, so without draining it here the
+    // next dispatch to the same program (the next 2K tile of one chunk on
+    // hosts with 4K prefill chunks) can begin its ANE evaluation and overwrite
+    // the surface before the merge kernel has executed. begin() only orders
+    // against the previous evaluation, not the previous merge. Commit and
+    // wait: the ANE and qmm branches were already joined on the host above,
+    // so the only added latency is the merge kernel itself.
+    {
+      encoder.end_encoding();
+      auto *merge_buffer = encoder.get_command_buffer();
+      merge_buffer->retain();
+      encoder.commit();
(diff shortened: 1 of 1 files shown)


beaglemoo:
Validation on an M5 Pro 64 GB (macOS 27.0, MLX 0.32.2, single ANE instance) with this patch applied on top of the PR 3359 lineage, model Qwen3.8-27B-oQ4e-mtp:

- `benchmarks/qwen35_ane_prefill_bench.py --force-lm --modes gpu single --tokens 2048 --repeats 3` at MLP shares 0.15, 0.25, 0.35, 0.40, 0.45 (GDN z at 0.375 for the last four). All 64 MLP and 48 GDN procedures compiled; no hangs, no merge errors, top token matched the GPU path in every run with logit cosine 0.999994.
- Two 16,317-token requests through `omlx serve` with the prefix cache off, 35% MLP / 0.375 GDN: 34.1 s and 35.3 s versus 37.1 s and 39.5 s GPU-only. No ANE wait timeouts or command-buffer failures in the server log.

The 2048-token projection tiles used here are the path this drain protects (next tile of the same program reusing the output surface), so the sweep exercised it on every layer.

jundot (approved):
Thanks for fixing the output-surface reuse ordering. Waiting after the merge closes the lifetime gap while preserving the ANE/GPU compute overlap. I checked all three merge paths and exercised repeated tiles through the single, dual, and fused variants on M3 Ultra. LGTM.

PR #3589, opened 2026-09-11
perf(moe): read missing experts in parallel with pread before the slot writes

beaglemoo:
Follow-up to #2595 / #986. Every expert cache miss in `moe_expert_offload.py` was a serial, synchronous memmap copy on the compute thread, one projection tensor at a time. On a large-expert-count checkpoint that makes prefill and low-residency decode I/O-latency bound rather than bandwidth bound.

`ExpertCache.ensure()` now runs two passes. The first classifies the call's misses without touching cache state and starts their reads on a bounded thread pool with `os.pread` (one read-only fd per shard; the `np.memmap` store is gone). The second pass is the original serial install loop, taking bytes from the pipeline instead of reading them itself. Every mutation of `slot_of`, `free`, `map`, the counters and the resident slots still happens on the calling thread in the serial order, so LRU victims, hit/miss counts, resident bytes and the numerical contract are identical to the serial path. `OMLX_MOE_OFFLOAD_IO_WORKERS` (default 12; 1 or less selects the serial path and creates no pool) and `OMLX_MOE_OFFLOAD_IO_BATCH` (default 4x workers) bound the threads and the in-flight payload bytes. Docs updated.

Measured on an M5 Pro 64 GB with `Vontra/Qwen3.8-Flash-Next-MLX-oQ2-MTP` (48 layers x 512 experts, 10 active, about 1.84 MB per expert), PLE table mmapped, MTP off, greedy, same prompts before and after:

| residency | case | before | after |
|---|---|---|---|
| 25% | 3461-token TTFT | 183 s | 18.7 s |
| 25% | decode after that prompt | 4.4 tok/s | 17.7 tok/s |
| 25% | 122-token warm decode | 2.3 tok/s | 16.2 tok/s |
| 68.8% | 3461-token TTFT | 30.1 s | 4.5 s |
| 68.8% | decode after that prompt | 13.6 tok/s | 19.7 tok/s |
| 68.8% | 122-token warm decode | 23.9 tok/s | 22.4 tok/s |

Generated text was identical to the serial baseline in every case. Loaded size was unchanged (33.9 GB at 68.8 percent).

Tests: `tests/test_moe_expert_offload.py` 36 fast and 3 slow passed, and again with `OMLX_MOE_OFFLOAD_IO_WORKERS=1`. New `TestParallelFetch` covers parallel-vs-serial slot and output equality, LRU order and counters, degenerate env values, concurrent store reads, and the in-flight payload bound. The pre-existing ruff findings on untouched lines of these two files are left alone.

The natural next step, kept out of this PR, is issuing the next prefill chunk's reads while the current chunk computes in `OffloadSwitchGLU.__call__`; the distinct set per chunk is already known up front.

### DIFF
--- a/docs/MoE_Expert_Offload.md
+++ b/docs/MoE_Expert_Offload.md
@@ -35,6 +35,13 @@ with a resident-fraction selector (12.5% – 75%). Or via the settings API:
 Toggling triggers an engine reload (it is a load-time transform). The env
 kill switch `OMLX_MOE_EXPERT_OFFLOAD=0` disables it regardless of settings.
 
+Two env vars tune the reader, and neither changes what is computed:
+
+| variable | default | effect |
+|---|---|---|
+| `OMLX_MOE_OFFLOAD_IO_WORKERS` | 12 | threads reading missing experts. `1` or less (or an unparseable value) keeps the serial path and starts no threads |
+| `OMLX_MOE_OFFLOAD_IO_BATCH` | `4 x workers` | experts whose reads may be in flight at once — the bound on the host memory the pipeline holds ahead of the slot writes |
+
 ## Performance
 
 `gemma-4-26b-a4b-it-4bit`, 585-token prompt, 256 generated tokens, warm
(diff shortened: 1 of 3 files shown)


jundot (approved):
The parallel reads keep slot updates and LRU ordering on the calling thread, and the results in #3686 support landing this before fetch/compute overlap. I'll merge this and immediately follow up on the read-failure cache state, the `IO_BATCH=1` refill case, and the existing prefill test's fetch spy.

PR #3593, opened 2026-09-11
fix(cache): guard unsorted VLM cache ranges

### LINKED ISSUE #2921: extra_key_ranges must be sorted ascending by start position crashes engine after long VLM/Pi conversation
**Describe the bug**
oMLX crashes the scheduler/engine while preparing the prefix cache with:

```text
AssertionError: extra_key_ranges must be sorted ascending by start position
```

The error is returned to the OpenAI-compatible client (Pi) as:

```text
Error: extra_key_ranges must be sorted ascending by start position
```

The exception originates from `resolve_block_extra_keys()` in `omlx/cache/paged_cache.py`, through:

```text
_prepare_prefix_cache_for_request
→ block_aware_cache.fetch_cache
→ paged_cache.find_shared_prefix
→ get_computed_blocks
→ resolve_block_extra_keys
```

It then propagates to `engine_core` and terminates the engine loop.

**To Reproduce**
I do not yet have a minimal reproduction, but the issue occurred with the following sequence:

1. Run `Youssofal--Qwen3.8-27B-MTPLX-Optimized-Speed-FP16` through oMLX.
2. Use Pi as an OpenAI-compatible API client.
3. Continue a long conversation containing previous image/VLM turns.
4. Let the context grow beyond 100k tokens.
5. Pi apparently compacts/rewrites the conversation context.
6. Send another request.
7. oMLX fails while preparing the prefix cache with:

```text
extra_key_ranges must be sorted ascending by start position
```

Immediately before the apparent compaction, the prompt had reached about 115k tokens:

```text
prompt: 115224
```

The following request dropped to about 43k tokens:

```text
prefix cache: request ... re-prefills 39115 of 43211 tokens (reused 4096);
closest stored sequence ... shares the first 5733 of 43211 comparable tokens
```

The scheduler crashes shortly afterwards.

This suggests that a rewritten/compacted conversation containing previous image turns may cause the segmented VLM cache metadata (`extra_key_ranges`) to become non-monotonic. This is currently only a hypothesis.

MTP appears unrelated: multiple MTP generations completed successfully immediately before the failure, with normal acceptance rates.

**Expected behavior**
Invalid or non-monotonic segmented VLM cache metadata should not crash the scheduler or engine.

oMLX already falls back to whole-request VLM cache keying when segmented boundary computation fails. I would expect the same fallback if the resulting ranges violate the monotonicity requirement.

The VLM code constructs ranges using:

```python
image_cache_key_ranges.append((boundary_tokens, cumulative_hash))
```

while `resolve_block_extra_keys()` requires the resulting start positions to be sorted ascending.

A possible safe behavior would be to detect non-monotonic ranges and discard segmented keying for that request:

```python
if any(
    image_cache_key_ranges[i][0] > image_cache_key_ranges[i + 1][0]
    for i in range(len(image_cache_key_ranges) - 1)
):
    image_cache_key_start = 0
    image_cache_key_ranges = []
```

Simply sorting the ranges may not be safe because each boundary is associated with a cumulative image hash.

**Screenshots**
Not applicable. I can provide the complete server log if useful.

* macOS Version: [26.6.1]
* oMLX Version: [0.6.2 (2347)]
* Model: `Youssofal--Qwen3.8-27B-MTPLX-Optimized-Speed-FP16`
* Client: Pi, using the OpenAI-compatible API

**Additional context**
Relevant traceback:

```text
oMLX.scheduler - ERROR - Error in batch generation step:
extra_key_ranges must be sorted ascending by start position

File ".../omlx/scheduler.py", line 9679, in _schedule_waiting
    self._prepare_prefix_cache_for_request(request)
File ".../omlx/scheduler.py", line 7712, in _prepare_prefix_cache_for_request
    block_table, remaining = self.block_aware_cache.fetch_cache(
File ".../omlx/cache/prefix_cache.py", line 618, in fetch_cache
    shared_block_ids, remaining = self.paged_cache.find_shared_prefix(
File ".../omlx/cache/paged_cache.py", line 1212, in find_shared_prefix
    cached_blocks, num_cached_tokens = self.get_computed_blocks(
File ".../omlx/cache/paged_cache.py", line 1043, in get_computed_blocks
    block_extra_keys = resolve_block_extra_keys(
File ".../omlx/cache/paged_cache.py", line 60, in resolve_block_extra_keys
    assert all(...),
AssertionError: extra_key_ranges must be sorted ascending by start position
```

The same exception then propagates to the engine loop.

I could not find an existing issue for this exact assertion.

arnavprabhu:
## Context

#3675 fixes the source of the reported VLM boundary mismatch by deriving boundaries from the final processor tokens and keeping generated ranges monotonic. This PR is a small defense-in-depth follow-up at the shared cache resolver.

## Change

If `resolve_block_extra_keys` receives unsorted segmented ranges, it skips positional resolution and uses the existing whole-request image key. Sorted ranges behave exactly as before.

This keeps malformed or future caller input from turning a cache miss into an assertion failure, while preserving conservative cache isolation.

## Tests

- `tests/test_paged_cache.py`: 73 passed

### DIFF
--- a/omlx/cache/paged_cache.py
+++ b/omlx/cache/paged_cache.py
@@ -50,17 +50,14 @@ def resolve_block_extra_keys(
     """Resolve which cache key salt applies to a block ending at ``block_end``.
 
     ``extra_key_ranges`` takes precedence over ``extra_keys`` and is intended
-    for segmented VLM cache keying.
-
-    ``extra_key_ranges`` must be sorted ascending by start position. The
-    function uses early break on the first non-matching entry, so unsorted
-    input will produce incorrect results.
+    for segmented VLM cache keying. Unsorted ranges (possible after a client
+    compacts a long VLM conversation) cannot be resolved positionally, so
+    they are ignored and resolution falls back to whole-request keying.
     """
-    if extra_key_ranges:
-        assert all(
-            extra_key_ranges[i][0] <= extra_key_ranges[i + 1][0]
-            for i in range(len(extra_key_ranges) - 1)
-        ), "extra_key_ranges must be sorted ascending by start position"
+    if extra_key_ranges and all(
+        extra_key_ranges[i][0] <= extra_key_ranges[i + 1][0]
+        for i in range(len(extra_key_ranges) - 1)
+    ):
         selected = None
         for start, keys in extra_key_ranges:
             if block_end > start:
@@ -68,6 +65,8 @@ def resolve_block_extra_keys(
             else:
                 break
         return selected
+    if extra_key_ranges:
+        logger.debug("Ignoring unsorted extra_key_ranges for cache key resolution")
     if extra_keys is not None and (
         extra_key_token_start is None or block_end > extra_key_token_start
     ):
--- a/tests/test_paged_cache.py
+++ b/tests/test_paged_cache.py
@@ -127,6 +127,33 @@ class TestResolveBlockExtraKeys:
 
         assert resolve_block_extra_keys(4, extra_key_ranges=ranges) == ("image-1",)
 
+    def test_unsorted_ranges_fall_back_to_legacy_keys(self):
+        """Non-monotonic boundaries must not raise; they resolve unsalted."""
+        ranges = [
+            (9, ("image-1", "image-2")),
+            (5, ("image-1",)),
+        ]
+
+        assert resolve_block_extra_keys(12, extra_key_ranges=ranges) is None
+        assert (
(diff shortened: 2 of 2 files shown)


arnavprabhu:
Rebased this on top of #3675 and trimmed it down. #3675 fixes the source of the bad boundaries; this just keeps the shared cache resolver from crashing if an unsorted range ever reaches it from another or future path. Sorted ranges are unchanged.

PR #3594, opened 2026-09-12
feat(qwen4_exp): Full 1M context support on 128 gig Macs.

VictorLowther:
Ever wanted to run with 1m context  using Qwen38-Flash-Next on your 128 gig MBP?  With this PR, you can.

- Proper YaRN context scaling for 2x and 4x automatically applied when you bump the max context above 256k.
- Turboquant enabled and properly cached during prefill in response to memory pressure.  Only pay for it when you need it.  4 bit only for now, but a new fused Metal kernel handles it properly for qwen4_exp models.
- Several memory guard bugs that only applied when running at the bleeding edge of a full 1M context fixed.

Model settings needed to make all this work.  A couple are new:

```
        "max_context_window": 1010000,   // above native horizon: auto-YaRN
        "turboquant_kv_enabled": true,
        "turboquant_kv_bits": 4.0,       // REQUIRED — see below
        "turboquant_skip_last": true,
        "mtp_enabled": true,             // optional (speculative decode)
        "mtp_prime_window": 65536,       // REQUIRED for large-context MTP
        "qwen4_ple_ssd_offload": true
```

Probably needs cleanup to be mergeable.


zviratko:
Good to know, I actually enabled RoPE scaling yesterday _and_ I am using SpecPrefill 🤦🏻

VictorLowther:
Eh, I would not necessarily take that on faith -- this whole thing is vibe-coded.  It is what Qwen 3.8 max thinks. :)

Also, UX knobs.  YaRN auto-scales when you embiggen the context past 256k

VictorLowther:
@zviratko what model are you using for spec prefill?  I have not seen the need for it yet.

zviratko:
Qwen3.5-2B-oQ6e

Ok, let me tell you a story :)
My reason for SpecPrefill is memory retention for Hermes Agent. I ran Hindsight and it was quite heavy, sensitive about semantic LLM quality and usually ran consolidation for quite a bit longer than the inference that produced it. I tried with various models but nothing was fast and good enough.
Then I switched to OpenViking - still heavy, but better, and I like the RAG functionality.
Shortly after, Qwen3.8-Next-Flash came out and that forced me to reconsider having a smaller LLM next to the main one, as it just doesn't fit on my 128GB M5 (if someone capable enough is reading this - saving just a few GB would make a world of difference for me :)). So I pointed everything at it and that has worked quite well because it doesn't degrade in speed with longer context as much as other models do and it is very smart compared to the 2B/4B/9B models I used for memory before.
But that still wasn't really enough.
My next optimization was getting an external NVMe enclosure and pointing cache at it, this made a big difference for KV cache and GDN cache once I gave it enough space.
Final (so far) optimization was enabling SpecPrefill, thought it took some trying

1) You can't just enable SpecPrefill for agentic workflows, it's only good for single(-ish)-turn exchanges over a chunk of data. If you try continuing the conversation and need something from the prefill then you get garbage.
2) I created model aliases over the main Qwen model in the hope of having a :main variant and :specprefill variant. unfortunately this doesn't quite work because SpecPrefill is a runtime/load-time setting and can't change between aliases, at least not without reloading the model (oMLX reloads it itself but it's not practical). And while OpenViking can add request parameters, like "specprefill_enabled" or "specprefill_keep_pct", what I needed was opposite - to have it disabled for the :main model requests. Chat template kwargs set in the admin don't seem to work for this (although I believe this could be made to work with generation_config.json maybe?)
3) Actual workaround - all the models and profiles have the following settings
```          "specprefill_enabled": true,
          "specprefill_draft_model": "/Users/zviratko/.omlx/models/Qwen3.5-2B-oQ6e-mtp",
          "specprefill_keep_pct": 0.3,
          "specprefill_threshold": 999999999999,
```
then in OpenViking:
```
"extra_request_body": {
      "enable_thinking": false,
      "specprefill_threshold": 16384,
      "specprefill_keep_pct": 0.3
    },
```

This makes it so SpecPrefill is enabled but not actually active for any request unless the client "asks" for it

This effect is quite drastic
a) Much faster (2-3x) prefill. This is a huge win when most of the prompt is unique data and single-turn
b) Much lower memory usage because only a fraction of the tokens are stored - though I won't claim how much, this seems to be a bit erratic and sometimes I'm seeing what I suspect is a memory leak

Quality seems to be very good - I had gpt-6 construct a simple benchmark to judge extraction quality and while it tried telling me "this is terrible, loses facts and doesn't understand context", it turns out it gave the same score to Qwen without SpecPrefill... take from that what you wil :-)

---
For future generations - I want to write an issue and ask to make this setup easier - to be able to have SpecPrefill model loaded but inactive except for a model alias with "specprefill_enabled: true", but the workaround seems to work fine.
Also, limiting or disabling cache size per model - or even better a model alias - would be useful there.
Also also, limiting number of concurrent requests per-model would have been useful when I ran a secondary model for memory before.

Let me know if you want to know anything more, I'm still experimenting with it myself, but so far the results have been great.

VictorLowther:
@zviratko Probably not really useful for me, then -- all I do is agentic coding work, so tg + prompt cache + long context wins.  Prompt processing isn't my bottleneck, and when I have tried spec prefill on other models the loss in accuracy over multiple turns compounded real quick.

zviratko:
My vibecoding is not really a good measure, but I found having reliable memory recall and RAG to be extremely useful. The agent remembers what we did and when and I can prompt it without having to provide the full context. I can simply reference the previous task and it looks everything up. Not sure what your harness uses but mid/long-term memory did wonders for me. This is not a problem with frontier models that bruteforce everything by handling 1M context gracefully, that's not really the case with our Qwens... :)
For pure agentic work, SpecPrefill is totally useless. Maybe if you enabled it from a very high threshold (64K+) so all the tool definitions survive... but even then I wouldn't trust it.

VictorLowther:
@zviratko You might be interested in the latest update to this PR

VictorLowther:
@jundot Thoughts?

PR #3600, opened 2026-09-12
fix(i18n): audit and complete the Simplified Chinese catalog (0.7.0.dev2)

LXD-8:
> **Based on `main` @ `7cbb407` (v0.7.0.dev2, 2026-09-13).** Rebased onto that commit; the only conflict was one hunk in `zh.json`, where `#3607` had added two keys next to a block this PR rewrites.

Audits the Simplified Chinese UI that shipped before the i18n migration and corrects what drifted. Three files: `omlx/admin/i18n/zh.json`, the macOS app's `Localizable.xcstrings`, and one assertion in `tests/test_admin_enhanced_readability.py`. **263 values corrected** — 233 in `omlx/admin/i18n/zh.json` and 30 in the macOS app's `Localizable.xcstrings`.

Two commits: the audit itself, then a pass that compares the web catalog with the macOS app's (the same user reads both surfaces). The second one is described under "One rendering per meaning" below.

## How the audit was scoped

Rather than applying an outside style guide, the target was the catalog's own accepted conventions (the values already merged). Punctuation, spacing, casing and terminology were measured across those values first, then every value — pre-existing and new — was checked against that profile. Where the catalog contradicted itself, the majority usage won; where a difference was semantic rather than stylistic it was kept.

## What it fixes

**Untranslated — 125 values**

| Area | Examples |
|---|---|
| Web search settings | `Search provider`, `Snippets only`, `Content limit`, `Test search`, all six hint strings |
| GDN snapshot / sidecar | `Snapshot Storage`, `SSD Write Buffer`, `SSD Snapshot Precision`, `Dedicated SSD Sidecar` |
| PLE / Engram SSD offload | `SSD N-gram Offload (Qwen4 only)`, `SSD N-gram Offload (DeepSeek V4.1)` |
| Qwen ANE panel | 16 tuned-split controls plus the whole tuner dialog (`Tuner overrides`, `Apply result`, `Preparing tuner…`, `CPU workers`) |
| Qwen INT8 / MoE | `Qwen INT8 Activation Prefill`, `MoE Expert Offload` |
| Chat | web-search and transcription statuses (`Searching the web`, `Transcribing`, `Listening`, `Fetching page`) |
| Metric tooltips | lead-ins kept the English name while the matching `.name` key was translated — now `批大小`, `端到端延迟`, `总吞吐量`, `TTFT（首个 Token 时间）`, … |

Two values that arrived on `main` after this branch was cut — the DeepSeek V4.1 CED prefill label and its hint — are translated here too, so the Chinese UI has no English left in it.

**Punctuation and spacing — 48 values**

- 34 ASCII `...` → `…`
- 3 half-width colons and 2 half-width parentheses inside Chinese sentences → full-width
- 8 missing spaces between Chinese and Latin
- 1 corner-bracket pair `「」` → `“”`

**Latin casing — 4 values, plus 5 in the macOS catalog**

`Token` capitalised as a common noun wherever it is a word rather than an identifier, plus the unit in `128 Token (fixed)` and the same five values in `Localizable.xcstrings`. Parameter names (`max_tokens:`, `tool_result_tokens:`) and the `your_token...` example keep their literal form — those are API names, not prose.

**Terminology and wording — 27 values**

- `Hot Cache` / `KV Cache` → 热缓存 / KV 缓存, matching the 40+ values that already said 缓存
- batch size had three renderings (批处理 / 批大小 / 批量大小) → one, 批大小
- `提示处理` → `提示词处理`
- three values used a second wording for a meaning the catalog already words one way: `留空表示不限制` → `留空表示无限制`, `您的客户端` → `你的客户端` (the catalog addresses the reader as 你 everywhere else), and `将不需要身份验证` → `将无需身份验证`
- the bench metric tooltips' English lead-ins now match their translated `.name` key
- `sidecar` is written in Latin in four values that said 边车, and the cluster's `readiness canary` keeps `canary` over 金丝雀. To be explicit that this is a preference and not a correction: 边车 is a legitimate rendering — it is what Kubernetes' official Chinese documentation uses for sidecar containers — and 金丝雀发布 is standard for canary releases. They are aligned to the Latin terms here because oMLX's own documents say "sidecar" (`SQLite sidecars`, `GDN sidecars`), because the newly added GDN error message already read `GDN SSD sidecar`, and because the Latin term is what the surrounding docs and search results use.
- Acronyms and mechanism names (MTP, PLE, Engram, GDN, MoE, DFlash, SpecPrefill, TurboQuant, imatrix, rank) were already kept in Latin and are unchanged.

125 translated + 48 punctuation and spacing + 4 casing + 27 terminology and wording = **204 web values in the first commit**, plus **5 in the macOS catalog**; the second commit adds 34 web values and 25 app values, five of which were already touched by the first, which is why the PR changes 233 web values in total.

Twenty-six further values — the Cluster v2 strategy/staging chips plus two bench and usage hints — sit inside the blocks the migration PR inserts new keys into. They are corrected **in that PR instead**, so that the two can be merged in either order without a conflict in `zh.json`. They are: 推荐, 每台 Mac 都处理每个 Token, 层分片, Tensor 拆分, 正在复制…, 复制失败, 本地网络权限, `pp_tps_req` 的说明, and 每小时 Token 总量.

## One rendering per meaning (second commit)

The audit above only looked inside the web catalog. This commit compares it with the macOS app's Simplified Chinese and removes the cases where one English string shipped two different Chinese ones. 34 web values and 25 app values change.

**The app's wording wins by default.** It loses where the web text is the better Chinese:

| key | now | was |
|---|---|---|
| `sidebar.accuracyBench` | 准确率 | 准确度 |
| `settings.model_type.rerank` | 重排序 | 重排器 |
| `…qwen_ane.tuner.tune_for_mac` | 为这台 Mac 调优 | 为此 Mac 调优 |
| `…dflash.draft_quant_group.label` | 分组大小 | 组大小 |
| `…dflash.draft_quant_activation.label` | 激活位宽 | 激活位 |
| `…dflash.draft_quant_weight.label` | 权重位宽 | 权重位 |

A few more are decided by the catalog's own conventions rather than by either side: sampling parameters stay Latin (`Min P`, not 最小 P), `Max Tokens` is 最大 Token 数, Profile is 配置档 rather than the ambiguous 配置文件 (a config *file* is also 配置文件), `Prompt Processing` stays 提示词处理 because the bench tooltips already use it and 预填充 is the rendering for Prefill, and the Runtime Cache panel's `RAM 条目` matches the `RAM 缓存` tier next to it.

Inside the catalog, five English strings had two renderings each — the previous commit fixed one side of them — so 已复制, 已缓存 Token, 从列表中移除, 预处理 TPS/请求 and 无头节点 are now the only rendering.

**Two distinctions that look like inconsistencies but are not, and are left alone:**

- **`Token` vs `令牌`** — model tokens keep the Latin word; auth tokens (HF / MS) are 令牌. 词元 is never used.
- **`Off` → 关 vs 关闭**, `On` → 开 vs 在, `Throughput` → 吞吐优先 vs 吞吐量, `Full` → 全量 vs 完整, `Yes` → 是 vs 确定 — each pair is a different thing (a toggle versus a select option, a mode name versus a metric, a preposition versus a switch state), not two renderings of one.

## Deliberately kept in English

Product and protocol names (oMLX, Mac, MLX, MoE, GDN, ANE, SSD, KV, RHT, INT8, Qwen, Brave, SearXNG), sampling parameters (`top_p`, `top_k`, `min_p`), chat-template kwarg identifiers (`enable_thinking`, `reasoning_effort`), placeholder examples (`hf_...`, `your_token...`, `mlx-community/...`), config keys shown in code chips, and the language selector's endonyms (Español, Français, Русский, 日本語, …).

## The last pinned value

`chat.enhanced_readability` was the one value that could not be translated: `tests/test_admin_enhanced_readability.py` asserted the label equalled `"Enhanced Readability"` in *every* locale.

That pin is relaxed here for `zh` — every other locale still has to carry the English fallback until its own translation lands, so the assertion keeps its teeth there:

```python
if loc != "zh":
    assert label == "Enhanced Readability", f"{loc} not English fallback"
```


LXD-8:
@jundot — companion note to #3586.

That PR moves the remaining hardcoded strings into the catalog; this one covers the other half — an audit of the pre-existing `zh` values: untranslated leftovers, terms and casing that drifted between screens, and half-width punctuation inside Chinese text. It touches nothing but `zh.json` values and five entries in `Localizable.xcstrings` — no template or code changes, so it reads as a plain diff of 230 values.

Worth doing now because once the catalog is the single place UI text lives, its quality is what every later localization inherits. Two properties: the key sets stay identical across the nine locales, and no locale other than `zh` changes.

Happy to adjust wording or drop individual entries if you disagree with a term — that kind of feedback is exactly what I'm after here. The one test change relaxes a hard-coded English pin so `zh` may translate the label while the other locales keep the English fallback until their own translation lands.

I'll keep contributing to the oMLX UI and to the zh-Hans localization going forward.

LXD-8:
Superseded, and re-published as part of a dedicated Simplified Chinese localization PR against the current `main` (`ddc47184`).

The migration this audit was interleaved with is now #3682 (string extraction only, every locale on the English fallback). This PR's catalog audit — the corrections to the Simplified Chinese that shipped before the migration — lands in that separate localization PR together with the zh wording for the migrated strings, so the two concerns stay apart.

The recorded reasoning (scoping the audit against the catalog's own conventions, the app's wording winning by default, and the deliberately-kept Latin terms) carries over intact.

Issue #3605, opened 2026-09-12
DeepSeek-V4.1: CED prefill skip (decoder tail-only forward) not implemented — ~2x cold prefill headroom


williamxie1989:
## Feature: implement CED prefill skip (+ SWA Bounded Replay) for DeepSeek-V4.1

The V4.1 tech report (§3.2.2) makes CED the headline prefill optimization:

> "Decoder SWA Bounded Replay bounds the decoder forward pass to n_win tokens, **nearly halving total prefill computation.**"

Under CED, decoder layers (21–39) get their global KV by projection from the final encoder hidden states (`kv_source_layers=[2,8,14,20]`), so during prefill their attention + MoE only need to run over the trailing `n_win` (128) tokens to build decoder SWA KV for the first decode steps. The rest of the prefix never touches decoder layers.

`omlx/patches/deepseek_v41/language.py::LanguageModel._forward` currently runs **all 40 layers over every chunk** — the CED skip is not implemented.

### Measured impact (Mac Studio M3 Ultra 512GB, oQ4e-mtp @ b390b31e)
- Cold prefill ≈ 400–470 tok/s (2048-token chunks ≈ 4.4–5.1 s, `[benchmark-prefill]` logs). Profiling shows ≈14 TFLOPS effective on the mxfp4→fp16 shader path, i.e. already near the hardware ceiling — the only way up is computing less.
- Skipping decoder layers 21–39 over non-tail tokens removes ~47% of prefill FLOPs → expected ~2× cold prefill (≈850–900 tok/s), halving long-context TTFT (125K fresh: ~4.6 min → ~2.4 min).
- Decode is unaffected (all layers run per cycle anyway).

### Design questions
1. Chunked prefill: decoder layers participate only on the final chunk (+128-token replay window); layer-20 KV projection + compressor (ratio=1) state machine must span chunks.
2. DSpark verify path stays full-depth (decode) — confirm no interaction with the tail-only prefill of decoder layers.
3. Cache snapshot/rollback (`mtp_partial_rollback`, boundary snapshots) semantics when decoder layers have no prefix state.
4. Numerics: tail-only decoder prefill is exact-by-construction for global KV (projection only differs in when it runs); SWA states for the first decode steps come from the replay segment, matching the paper's bounded-replay approximation.

Happy to prototype if you'd accept it — the paper's production numbers hinge on this.

[...]

PR #3607, opened 2026-09-12
feat(deepseek_v41): CED prefill skip with SWA bounded replay

williamxie1989:
## What

Implements the CED prefill proposal from #3605: during prefill, the decoder half (layers 21–39) runs attention + MoE for only the trailing `n_win` (sliding-window) tokens. The decoder's global KV is obtained by having the midpoint CSA2 layer project the **full** encoder-final hidden state `H_20` (one `wkv` matmul, no attention/MoE), exactly as the paper prescribes.

Off by default; enabled per model via `deepseek_v41_ced_prefill_enabled`.

## Measured (Mac Studio M3 Ultra 512GB, DeepSeek-V4.1-Flash-oQ4e-mtp)

| | CED off | CED on |
|---|---|---|
| cold prefill, 21.4K-token prompt | 50.19 s ≈ **430 tok/s** | 30.55 s ≈ **700 tok/s** (**+63%**) |
| DSpark decode tok/cycle / accept | 4.00 / 71.4% | 3.20–4.00 / 71–73% (unchanged) |

+63% rather than the theoretical ~2× because the encoder (20 layers), the midpoint global-KV projection, and the indexer full-key path still run over every token, and each prefill chunk pays a small extra decoder tail.

## Design decisions

- **Layout validation**: `ced_layout_supported()` requires even `n_layers`, the midpoint layer to be a ratio-1 `kv_source` layer, all ratios above the midpoint to be 1, no `kv_source` above the midpoint, and all engram layers below the midpoint. Unsupported layouts auto-disable with a warning.
- **Chunked prefill**: the CED gate is *absolute* (`cache offset + chunk length > window`), and every active chunk drops its stale window cache (`cache[1] = None`). This makes chunked CED **bit-identical** to a single CED pass — covered by `test_ced_chunked_continuity_and_decode_seam`. Intermediate chunks do compute their own 128-token tails (extra ~6% work); their window rows are cleared by the next active chunk, so the final cache state depends only on the trailing window.
- **Decode / speculative untouched**: decode steps (length 1) and DSpark verify blocks (length ≤ depth+1 < window) never enter the CED path, so `mtp_partial_rollback` semantics are unchanged.
- **DSpark prime ring**: CED priming captures only the tail rows; the ring legitimately replaces a stale window with a fresh full window (`start > offset and length >= max_size`). RoPE is baked at append time and ring attention is non-causal, so slot order is free.
- **Native packed attention**: the CED tail (`localL == qL == 128`) hits the same corner case as today's short-prompt fast path; the fallback index math with `old_len = 0` was verified slot-by-slot.

## Tests

- CED-off regression: full existing V4.1 + DSpark suites unchanged.
- New: layout validation (incl. `validate()` rejection), encoder + midpoint global-KV caches bit-identical CED-on vs off, determinism, inactive-within-window bitwise full compute, chunked-vs-single CED bitwise equality + decode seam, DSpark ring gap replacement.

## Answers to the four questions in #3605

1. **Chunked prefill**: absolute gate + per-active-chunk window clear ⇒ chunked == single, bitwise (tested).
2. **DSpark verify**: never enters CED (length guard); rollback path untouched.
3. **Snapshot/rollback**: CED never runs during verify; SSD snapshots see the same tail-window shape as full compute.
4. **Numerics**: deterministic (same kernel order per chunk); early tail queries see truncated windows — the paper-approved bounded-replay approximation; encoder and global KV are bit-identical to CED-off.

### DIFF
--- a/omlx/engine/vlm.py
+++ b/omlx/engine/vlm.py
@@ -1842,6 +1842,13 @@ class VLMBatchedEngine(BaseEngine):
                                 False,
                             )
                         ),
+                        ced_prefill=bool(
+                            getattr(
+                                self._model_settings,
+                                "deepseek_v41_ced_prefill_enabled",
+                                False,
+                            )
+                        ),
                     )
                 if model_type == COHERE2_MOE_MODEL_TYPE:
                     return _load_cohere2_moe_text_model(
--- a/omlx/model_profiles.py
+++ b/omlx/model_profiles.py
@@ -111,6 +111,8 @@ EXCLUDED_FROM_PROFILES = frozenset(
         # Hardware-specific residency choice; never propagate across models.
         "qwen4_ple_ssd_offload",
         "deepseek_v41_engram_ssd_offload",
+        # Architecture-level prefill strategy; explicit per model.
+        "deepseek_v41_ced_prefill_enabled",
         # Security flag must be explicit per model — never propagated via profiles.
         "trust_remote_code",
     }
--- a/omlx/model_settings.py
+++ b/omlx/model_settings.py
@@ -320,6 +320,10 @@ class ModelSettings:
     # fit under the configured model-memory ceiling but mmap loading can.
     qwen4_ple_ssd_offload: bool = False
     deepseek_v41_engram_ssd_offload: bool = False
+    # DeepSeek V4.1 CED: during prefill the decoder half only forwards the
+    # last window-size tokens; decoder global KV is the encoder-final
+    # projection already produced by the midpoint CSA2 layer.
+    deepseek_v41_ced_prefill_enabled: bool = False
     preserve_thinking: Optional[bool] = (
         None  # Keep <think> blocks in historical turns (None = auto, True when template supports it)
     )
--- a/omlx/patches/deepseek_v41/config.py
+++ b/omlx/patches/deepseek_v41/config.py
@@ -100,6 +100,10 @@ class ModelConfig:
     image_token_id: int = 129264
     # Draft weights and standalone forward are separate from server verification.
     preserve_mtp: bool = False
+    # CED prefill skip: the decoder half forwards only the trailing
+    # window-size tokens; its global KV is produced by the midpoint CSA2
+    # layer projecting the encoder-final hidden states.
+    ced_prefill: bool = False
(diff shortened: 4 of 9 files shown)


jundot:
Thanks for putting this together so quickly. I was also prototyping the idea from #3605 locally, and your midpoint split goes further by avoiding additional attention, indexer-query, and MoE work while retaining full-context global KV. I'd like to use this PR as the base.

I've tested it with a prepared follow-up covering short prefill suffix positions, full-logit scoring, DSpark continuity, and cache isolation when switching CED modes. Real-model full and partial SSD hits matched their cold-prefill outputs. I also fixed a memory-settlement delay that could reject an immediate model reload. I'll merge this and land those fixes afterward.

Here are my measurements on an **M3 Ultra, 512 GiB**, using **DeepSeek-V4.1-Flash-oQ4e-mtp with Engram in RAM and CED enabled**, `code_python` context, temperature 1, and top_p 1. SSD KV and hot prefix caching are off.

| Context | PP dev2 | PP PR | PP vs. dev2 | TG128 OFF → ON | Peak MLX | Peak RSS |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 4K | 452.15 | 787.63 | +74.2% | 20.60 → 33.22 | 404.02 | 403.15 |
| 16K | 454.77 | 806.81 | +77.4% | 20.38 → 38.54 | 404.15 | 403.17 |
| 32K | 447.73 | 797.78 | +78.2% | 20.22 → 30.56 | 404.16 | 403.19 |
| 64K | 435.56 | 779.68 | +79.0% | 19.96 → 35.88 | 404.17 | 403.22 |

PP and peak memory use **MTP ON**; TG shows **MTP OFF → ON**. Speeds are tok/s and memory is GiB. `PP dev2` comes from the [release notes](https://github.com/jundot/omlx/releases/tag/v0.7.0.dev2); `PP PR` includes my prepared follow-up. Each configuration received a 4K warm-up, with prefill chunks capped at 2K and 128 generated tokens. Loading and warm-up are excluded. These are single runs; sampling and MTP acceptance can vary. CED remains an approximate, opt-in optimization.

Thanks again for the implementation and measurements.

Issue #3612, opened 2026-09-12
Feature Request: Agent-facing KV cache export/import — deterministic per-session cache artifacts


nizovtsevnv:
# Feature Request: Agent-facing KV cache export/import — deterministic per-session cache artifacts

## Summary

For agentic coding workflows (pi, Claude Code, etc.) running against oMLX, long sessions accumulate a large prompt prefix. When the model/server is reloaded, or the working set is evicted, the next turn pays a **full re-prefill of the entire accumulated context** — minutes of compute for a 60–120k-token session.

I'd like the engine to expose an **export/import API for prefix cache state**, so that the *agent* can materialize the cached prefix into a deterministic file artifact (stored next to its own session file), and later hand that artifact back to oMLX for restore — instead of re-prefilling.

The built-in tiered KV cache (hot RAM + cold SSD) already solves the common case well; this request adds a layer of **agent-owned, deterministic persistence** on top of it.

## Why this is different from the existing SSD tier

The existing cold tier is engine-managed:

- keyed by hash, bounded LRU (e.g. `ssd_cache_max_size`), evicted under pressure,
- wiped by cache-clear actions, upgrades, reinstalls, disk cleanup,
- no per-session semantics: nothing ties a cache entry to an agent session,
- not inspectable or portable by the client.

An agent-owned artifact adds:

1. **Deterministic lifetime.** The file lives in the agent's session directory (e.g. `~/.pi/agent/sessions/--<project>--/<session>/kv.omlx` alongside the JSONL transcript). It survives cache clears, evictions, and oMLX upgrades; the agent deletes it explicitly.
2. **Explicit invalidation by the agent.** The agent knows when its history changes semantics (compaction, in-place edit, fork) and can discard/truncate the artifact. It also knows when to export (right after a turn completes, while state is hot).
3. **Portability.** Carry the artifact between two machines with an identical setup (same model file, same quant, same oMLX version) — home/office, laptop/Mac mini.
4. **Auditability.** One file per session, inspectable size/hash, never silently evicted.

## Suggested API shape

Consistent with the existing admin router (e.g. `POST /api/cache/probe`):

```
POST /api/cache/export
  { "model": "<model-id>", "prompt": <messages, same shape as chat completion> }
  → 200 { "manifest": { "format_version", "model", "quant", "engine_version",
                        "prefix_digest", "restorable_tokens", "size_bytes", ... },
          "blob": <stream or download URL> }
[...]

Issue #2250, opened 2026-07-15
Session compaction gave me another users session from another machine


RDouglasSharp:
Prompt:

Please read KV-BLOCK-CACHE-DESIGN.md and KV-CACHE-PLAN.md both in /Users/doug/Code/colibri-main/ , then implement the plan step by step against the code in /Users/doug/Code/colibri-main/colibri-dev/c .  Step 1, 2, 3, 4  and 5 are complete.  Please implement step 6.

Initial work:

tep 6 is the integration tests for reuse scenarios and persistence. Let me first examine the existing codebase and test structure.

Now let me check the existing block cache tests from steps 1-5 and look at the relevant parts of glm.c:

Let me look at the block cache structures in glm.c to understand what was implemented:

Let me look at the raw_mode path in run_serve more closely and the API log format:

Let me check how the existing tests are built and any correctness test from step 5:

-------------------- Session Compacted ----------------------

Objective
Assess whether an AI-driven energy management and trading layer for a Virtual Power Plant (VPP) is technically feasible, can add meaningful profit margin, and is relevant to the German regulation context.
Important Details
VPP Architecture: 300+ DERs (200 BESS 50kW/100kWh + 100 PV 50kW), AGG-like design
Targets: 8% margin increase via real-time price trading + 30s dispatch latency for aFRR/mFRR packages
Phase 1: Firm/educational, real data / simulated execution
Phase 2: exec / reasonable latency (every 10 minutes)
Phase 3: live sandbox
VPP optimization: two-state POU model used as description vs. three-state POU model used as multiplier
German energy markets: EPEX Spot (15min ahead/day-ahead), aFRR and mFRR (balancing), PV/BESS owned
Two approaches explored for dispatch: SDP-ADMM and hierarchical MPC with pOSDT/POSED stack
Work State
Completed
(none)
Active
Exploring VPP optimization algorithm selections (SDP-ADMM vs. hierarchical MPC)
Assessing AI-driven marketplace architecture for energy trading (bid optimization, DRL agent selection, overlay vs. embedded)
Forecasting strategies (multimodal: price + generation + weather)
Legal/regulatory stack considerations for TriCom
Blocked
(none)
Next Move
Recommend algorithm choice between SDP-ADMM and hierarchical MPC for the VPP control architecture
Continue planning the AI overlay strategy and forecasting stack integration
Relevant Files
(none)

[...]

PR #3615, opened 2026-09-12
Add agent-facing KV cache export/import artifacts (draft, RFC)

nizovtsevnv:
## Summary

Vertical slice for the agent-facing KV cache artifact from issue #3612.

The paged-SSD prompt cache already makes block files durable on disk, but they are engine-owned (LRU, evicted, cleared). This PR adds a **cache artifact**: a self-contained zip of the block chain (and GDN sidecars) behind a prompt prefix, plus two admin endpoints so an external agent can materialize a long prefix into a file it controls and register it back after a restart instead of paying a full re-prefill.

- **`POST /api/cache/artifact/export`** — renders + tokenizes the conversation exactly like `probe_cache`, walks the same chain hash the scheduler uses, and writes a zip of the longest contiguous *durable* prefix (stops at the first block that is RAM-only or missing).
- **`POST /api/cache/artifact/import`** — validates the artifact fail-closed and registers its block files + GDN sidecars with the SSD index. After a successful import the prefix is restorable without re-prefill (verifiable via `POST /api/cache/probe`).

## Artifact format (v1)

```
manifest.json          # format_version, model, block_size, token_ids, per-file sha256, engine version
blocks/<hash>.safetensors   # byte-identical copies of SSD block files
gdn/<sig_digest>/<hash>.safetensors
```

The manifest carries the prefix token ids (required on import to re-derive the block-hash chain — import rejects a manifest whose hashes don't match its tokens). Block payloads are **not re-serialized**: they are the exact bytes the engine already wrote, so import = "copy files back + re-index".

## Validation (fail-closed)

Import rejects the whole artifact (nothing registered, nothing left on disk) on:
- unsupported format version / unreadable manifest / missing members
- model identity mismatch
- token-chain vs block-hash inconsistency
- per-file sha256 mismatch
- any block failing the same `_is_compatible_block` checks the startup scan uses (layout, layer types, cache signature)
- existing local file with the same hash but different size (no silent overwrite)

## Design notes / deliberate limits

- **Additive only**: no changes to cache matching, admission, eviction, or the scheduler. New storage is just ordinary SSD cache files, so existing LRU/eviction applies to imported blocks exactly as to prefilled ones.
- **Single-node only.** Cluster-wide restore (multi-rank snapshot files) is a follow-up.
- **Hybrid (GDN) models**: sidecars are exported/imported per block with the live cache signature, and the (block signature, sidecar digest) pairing is validated on import. MTP draft state is intentionally not persisted.
- **RAM-only setups** (`cache.hot_cache_only=true`) get a 409 with a hint to enable the SSD tier — there is nothing durable to export.
- The artifact contains the token ids, i.e. derived conversation content: ship it with the same care as the transcript (cf. #2250 on cross-machine cache leakage).

## Files

- `omlx/cache/artifact_store.py` (new) — format + export/import core
- `omlx/admin/cache_artifacts.py` (new) — admin endpoints
- `omlx/server.py` — router wiring (mirrors existing admin DI pattern)
- `omlx/cache/paged_ssd_cache.py` — exposes `hot_cache_only` property (2 lines)
- `tests/test_artifact_store.py` (new) — round-trip, hole truncation, idempotency, tamper/model/chain/zip rejection, GDN sidecar round-trip, empty export

## Testing

- `pytest tests/test_artifact_store.py` — 10 new tests, all passing
- `pytest tests/test_paged_ssd_cache.py` — 170 existing tests still passing
- `ruff check` clean on new/modified files

## Open questions for review

1. Artifact location: default is `~/.omlx/cache_artifacts/` (next to the SSD cache dir); `output_dir` overrides. Should this be a settings field instead?
2. Token ids in the manifest — acceptable vs. re-tokenizing on the engine side only (would require the agent to store the message list and re-render, which is what probe already does — could drop token ids from v1 if the trade-off is preferred)?
3. Endpoint naming — `/api/cache/artifact/*` vs. extending `/api/cache/probe` with actions?
4. v1 writes the zip on the server and returns the path; streaming over HTTP is a follow-up.

### DIFF
--- a/omlx/admin/cache_artifacts.py
+++ b/omlx/admin/cache_artifacts.py
@@ -1,0 +1,271 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Admin API: export/import paged-SSD cache artifacts (issue #3612).
+
+An artifact is a zip of the SSD blocks (and GDN sidecars) behind a prompt
+prefix; an agent stores it next to its own session and hands it back to
+oMLX after a restart instead of paying a full re-prefill. Core logic and the
+artifact format live in ``omlx/cache/artifact_store.py``.
+
+Both endpoints are admin-only, single-node, and additive: they never touch
+cache matching, admission or the scheduler, only read the SSD manager's
+index and register verified files back into it.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import logging
+from pathlib import Path
+
+from fastapi import APIRouter, Depends, HTTPException
+from pydantic import BaseModel
+
+from ..cache import artifact_store
+from ..cache.artifact_store import ArtifactError
+from .auth import require_admin
+from .routes import _normalize_probe_tool_calls, _probe_chat_template_kwargs
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter()
+
+_get_engine_pool = None
+_get_settings_manager = None
+
+
+def set_cache_artifact_getters(
+    pool_getter, settings_manager_getter
+) -> None:
+    """Wire server getters (same pattern as ``set_admin_getters``)."""
+    global _get_engine_pool, _get_settings_manager
+    _get_engine_pool = pool_getter
+    _get_settings_manager = settings_manager_getter
+
+
+class CacheArtifactExportRequest(BaseModel):
+    """Export the restorable prefix behind this conversation into a zip.
+
+    ``messages``/``tools`` are rendered and tokenized exactly like a real
+    turn (same path as ``/api/cache/probe``) so the block chain lines up
+    with what a re-prefill would build.
+    """
+
+    model_id: str
+    messages: list[dict]
+    tools: list[dict] | None = None
+    chat_template_kwargs: dict | None = None
+    thinking_budget: int | None = None
+    # Where to write the artifact. Defaults to a ``cache_artifacts``
+    # directory next to the SSD cache dir.
+    output_dir: str | None = None
+
+
+class CacheArtifactImportRequest(BaseModel):
+    """Validate and register a previously exported artifact."""
+
+    model_id: str
+    artifact_path: str
+
+
+def _resolve_context(model_id: str) -> tuple:
+    """Engine + SSD-manager plumbing, mirroring ``probe_cache`` in routes.py.
+
+    Returns (engine, entry, ssd_manager, model_name, block_size, tokenizer).
+    """
+    engine_pool = _get_engine_pool()
+    if engine_pool is None:
+        raise HTTPException(status_code=503, detail="Engine pool not initialized")
+
+    entry = engine_pool._entries.get(model_id)
+    if entry is None:
+        raise HTTPException(status_code=404, detail=f"Model not found: {model_id}")
+    if entry.engine is None:
+        raise HTTPException(
+            status_code=409,
+            detail=f"Model is not loaded — load it to use cache artifacts: {model_id}",
+        )
+
+    engine = entry.engine
+    tokenizer = getattr(engine, "_tokenizer", None)
+    if tokenizer is None or not hasattr(tokenizer, "apply_chat_template"):
+        raise HTTPException(
+            status_code=400,
+            detail="Model tokenizer does not support chat templating.",
+        )
+
+    async_core = getattr(engine, "_engine", None)
+    core = getattr(async_core, "engine", None) if async_core is not None else None
+    scheduler = getattr(core, "scheduler", None) if core is not None else None
+    if scheduler is None:
+        raise HTTPException(
+            status_code=500, detail="Scheduler unavailable for loaded model."
+        )
+
+    ssd_manager = getattr(scheduler, "paged_ssd_cache_manager", None)
+    if ssd_manager is None:
+        raise HTTPException(
+            status_code=409,
+            detail="Paged SSD cache unavailable — the cache may not be enabled.",
+        )
+    paged_cache = getattr(scheduler, "paged_cache_manager", None)
+    prefix_cache = getattr(scheduler, "block_aware_cache", None)
+    block_size = getattr(
+        getattr(scheduler, "config", None), "paged_cache_block_size", 0
+    )
+    if not block_size and prefix_cache is not None:
+        block_size = getattr(prefix_cache, "block_size", 0)
+    if not block_size:
+        raise HTTPException(
+            status_code=500,
+            detail="Cache block size unavailable — cache may not be enabled.",
+        )
+    model_name = (
+        getattr(paged_cache, "model_name", None)
+        if paged_cache is not None
+        else None
+    ) or model_id
+    return engine, entry, ssd_manager, model_name, block_size, tokenizer
+
+
+def _require_ssd_tier(ssd_manager) -> None:
+    if getattr(ssd_manager, "hot_cache_only", False):
+        raise HTTPException(
+            status_code=409,
+            detail=(
+                "SSD cache tier is disabled (cache.hot_cache_only=true) — "
+                "blocks are RAM-only and not portable. Enable the SSD tier "
+                "first."
+            ),
+        )
+
+
+def _tokenize_prompt(
+    engine, entry, tokenizer, request: CacheArtifactExportRequest
+) -> list[int]:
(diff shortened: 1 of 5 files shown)


moc375:
Good to see this direction — we've been running gateway-orchestrated cross-instance cache migration on a similar hand-rolled dir-export for a while, and the artifact format here (zip + sha256 manifest + fail-closed import) is a clear improvement. Tested this branch on a 0.7.0.dev1-era tree: all 10 artifact-store tests pass.

One integration gap found on a live instance: under `hot_cache_write_back=True` (the default, non-pressure regime) KV blocks from a fresh prefill stay in the RAM pending-writes tier indefinitely — only GDN sidecars reach disk. Re-sending the same 14k-token prompt hits 24.5× faster than cold (0.8s vs 19.1s), so the blocks are definitely hot, but six minutes after the request zero block files exist on disk and export returns 409 "No restorable prefix found". Since the walk stops at the first RAM-only block, the headline agent use case (export a hot session before a restart) fails exactly when it's needed most — on large-RAM, lightly loaded machines a fresh session is effectively never exportable.

Suggestion: have export drain the pending-writes for the walked chain first (or accept `flush=true`). The 409 detail string would also benefit from naming this case — the current message ("not yet prefilled, evicted, or the SSD tier was off") doesn't cover it.

One note from gateway-side integration: because the chain hash is over the tokenized prompt, the export request body must replay the exact rendered conversation the engine saw — for a proxy that's the transformed body (after goal injection / tool-slim), not the client's original. Accepting optional pre-rendered token IDs in the export request would make that replay less fragile.

nizovtsevnv:
Thanks — that live-instance finding is exactly the gap I was hoping this draft would surface, and you're right: in write-back retention a fresh prefill is hot-cache-only (no SSD index entry, no queued write) until eviction or shutdown, so a lightly loaded machine makes a fresh session effectively never exportable. Both points are addressed in 52deebca:

**1. Export now drains the chain's write-back blocks.** New `PagedSSDCacheManager.flush_blocks(hashes)`: for each dirty hot entry it does the same blocking enqueue the shutdown flush uses (index entry + write-queue item), then waits for the writer to commit that block and counts only files verified on disk. `export` calls it on the walked chain before the walk (`flush=true` default, overridable in the request). The 409 detail now names this case: manifest carries `ram_only_blocks`, so the message reports how many prefix blocks are still RAM-only after the flush attempt.

Two implementation notes:
- `_enqueue_ssd_write(blocking=True)` alone is not sufficient — its timeout bounds queue *space*, not write completion — so flush waits per block on the pending-write buffer clearing, with a bounded stall backstop (30 s total, module constant).
- Blocks that were already evicted to the pending-writes buffer (RAM pressure) need no flush: they have an index entry and a queued write already, so the walk just sees a short-lived hole; the manifest's `ram_only_blocks` counts them for diagnostics.

**2. Pre-rendered token ids.** Export now accepts optional `token_ids` (takes precedence over `messages`; the request validates that one of the two is present) — for proxies that must replay the transformed body, exporting from the exact ids the engine saw is now the first-class path.

New tests: write-back retention stays RAM-only without flush (`flush=false` → zero restorable, `ram_only_blocks=3`), `flush_blocks` materializes + indexes + keeps entries loadable (and is a clean no-op on a second call), and default export of a fresh session now yields the full chain. 13 artifact tests + 201 across the paged-SSD/probe suites, ruff clean on new code.

moc375:
Re-tested `52deebca` on a live instance — both fixes work, and the `token_ids` path turned out to be the key to isolating one more live gap.

**What works:**

- `flush=true`: a fresh session (single request, KV still in the write-back tier) now exports cleanly — 14,336 restorable tokens across 7 blocks, `ram_only_blocks=0`, block files materialized on disk. Previously this was the 409 case.
- `token_ids` export + full destroy-recover round trip: wiped the entire SSD cache dir, restarted the engine, imported the artifact (0.8 s, 7 blocks + 7 GDN sidecars, 1.2 GB zip), re-sent the same request — **17 s cold prefill → 0.9 s**, correct output. That's the whole promise of the artifact format working end to end.

**The remaining gap — `messages` still 409s on live serving:**

For the same conversation, exporting with `messages` returns "nothing for this prompt is durable on disk", while exporting with the exact token ids succeeds. So the chain hashes the `messages` walk computes don't match what serving actually used — `_tokenize_prompt`'s rendering diverges from the serving path somewhere (our checkpoint is `Qwen3.8-Flash-Next-oQ4e-mtp`; serving ran with `enable_thinking=false`, and a client-side render with `add_generation_prompt=True, enable_thinking=False` through the checkpoint's own tokenizer matches the engine — that's what the successful `token_ids` export used).

The artifact tests construct managers with hand-built hashes, so this serving→tokenizer layer is untested — that's why it can pass 13/13 and still miss live. Routing `_tokenize_prompt` through the exact same render call the server uses for a real turn would close it; the `token_ids` escape hatch is a fine workaround for proxies in the meantime (arguably the better default for us regardless, since the gateway already holds the transformed request).

Thanks for turning both suggestions around same-day — the flush semantics note (queue-space timeout vs write completion) was a subtlety we hadn't spotted ourselves.

nizovtsevnv:
Fixed the `messages` mismatch in 01fe242c — the export path no longer re-renders prompts at all, it replays the served turn.

**Root cause**: `cache_artifacts._tokenize_prompt` (and `probe_cache`, same code) built the prompt from scratch: caller messages + `convert_tools_for_template` + `_apply_chat_template` with locally merged kwargs. Serving does a lot before it templates — `extract_text_content` (tool-result truncation, tool-call flattening), MCP tool merge, Gemma param enrichment, system-message folding, `response_format` JSON instruction, `detect_and_strip_partial`, settings/`reasoning_effort`/thinking-budget resolution. Anything missing there is a different token sequence, so the chain walk stopped early and reported "nothing durable on disk" for a conversation that had just been served. `token_ids` worked because it skipped the re-derivation.

**Change** — one implementation, two choke points:
- `server.prepare_chat_turn_inputs()`: the prompt-preparation block of `create_chat_completion`, moved out verbatim (verified AST-identical, 46 statements). `/v1/chat/completions` calls it; nothing about the endpoint's order or behavior changed.
- `server.served_prompt_token_ids()`: builds a real `ChatCompletionRequest` (so `enable_thinking` aliases, `reasoning_effort`, `thinking_budget`, tool-call argument parsing all normalize the serving way), runs `prepare_chat_turn_inputs`, then `engine.prepare_chat_messages()` → `engine.render_chat_prompt()` → `tokenizer.encode()` — the same three calls `stream_chat` makes.
- `BaseEngine.prepare_chat_messages()` / `render_chat_prompt()`: the canonical render; `BatchedEngine`/`DFlashEngine` `chat`, `stream_chat`, `preflight_chat`, `count_chat_tokens` now render through it, so the engine cannot drift either.
- Probe and artifact export both call `served_prompt_token_ids()`; the probe-local `_probe_chat_template_kwargs` and `_normalize_probe_tool_calls` mirrors are deleted. Both endpoints now forward `max_tokens`, `reasoning_effort`, `tool_choice`, `response_format` (each changes the rendered prompt, so omitting them was itself a hash miss).

**Intentional behavior changes**:
- A settings-store outage now fails the probe/export with 400 instead of falling back to request kwargs — the fallback rendered a prompt nobody served and answered "cold" for a warm cache.
- `POST /api/cache/artifact/export` is async; the copy runs in `asyncio.to_thread` so a multi-GB artifact no longer blocks the event loop.
- Multimodal `messages` are refused with a pointer to `token_ids`: image/audio blocks enter the key outside the text, so they can't be reconstructed from a message list.

**Tests**: `tests/test_served_prompt_parity.py` (11) pins each transformation in the hashed prompt (tool-result truncation, MCP merge, reasoning effort, model defaults + forced kwargs, JSON instruction, partial mode, and that the result differs from the old naive render), plus export-by-messages == served ids, probe == export, and the multimodal refusal. Probe tests were retargeted to the deleted mirrors. Full suite: 12887 passed, 345 skipped; no new ruff findings.

**Needs your box to confirm**: the live sequence from your last report — serve a turn, then export by `messages` (should return the full chain instead of 409), and check `/api/cache/probe` reports the same `total_tokens` as the export manifest. Anthropic `/v1/messages` and `/v1/responses` still have their own prep chains and stay on `token_ids`; unifying those is separate work.

PR #3617, opened 2026-09-12
fix(api): preserve Anthropic plain-text document content

LittleYier:
Anthropic [plain-text documents](https://platform.claude.com/docs/en/build-with-claude/citations#plain-text-documents) use `source.type="text"`, but the converter base64-decodes every `text/plain` document. This can replace the content with a decode-error placeholder or silently turn literal `SGVsbG8=` into `Hello`.

Preserve the data for text sources and retain existing base64 handling. The existing document test now covers both encodings, including Unicode, newlines and base64-looking literal text.

353 related tests passed:

```sh
python -m pytest --noconftest -o addopts='' tests/test_api_utils.py tests/test_anthropic_models.py tests/test_anthropic_adapter.py -q
```

These are conversion/model-schema/adapter tests; `--noconftest` skips unrelated global MLX setup. No model inference was run.


jundot (approved):
Thanks for fixing this. I verified that plain-text documents retain Unicode and base64-looking literal text across the conversion paths, while existing base64 handling stays unchanged. LGTM, I'll merge this.

PR #3619, opened 2026-09-12
fix(patches): inject bisect_right into mlx-lm qwen4_exp

xunlinkx:
mlx-lm's qwen4_exp calls bare bisect_right for the PLE ngram shard lookup but only imports bisect, so every ngram prefill raises NameError at runtime.

The vendored mlx-vlm copy already imports it directly, but a cluster rank serves the flat mlx-lm model and never passes through the VLM compat path.

Adds a load-path patch in omlx/patches/mlx_lm_qwen4_exp.py, wired from model_loading.py like minimax_m3_mlx_lm. Unblocked by for_vlm so it runs on any rank that loads the flat model.

### DIFF
--- a/omlx/patches/mlx_lm_qwen4_exp.py
+++ b/omlx/patches/mlx_lm_qwen4_exp.py
@@ -1,0 +1,35 @@
+# SPDX-License-Identifier: Apache-2.0
+"""mlx-lm qwen4_exp fixes that must apply outside the mlx-vlm compat path."""
+
+from __future__ import annotations
+
+import bisect
+import logging
+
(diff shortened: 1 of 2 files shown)


jundot:
Neither the pinned mlx-lm commit nor current upstream main includes `mlx_lm.models.qwen4_exp`. This PR patches that module if present, but does not provide or register it, so it cannot enable Qwen4-Exp loading through the distributed mlx-lm path in a standard oMLX installation.

I don't currently have the bandwidth to maintain a separate mlx-lm implementation for this model. I'll revisit distributed support once official mlx-lm support is available, so I'm closing this PR for now.

PR #3624, opened 2026-09-12
fix(tool_calling): merge line-separated JSON fragments into a declared array param

ccdsky:
## Problem

Some models emit an array-typed tool parameter as one JSON value per line. Reproduced with `Qwen3-Coder-Next-8bit` on omlx 0.6.3 and 0.6.4 via the `qwen3_coder` XML fallback path, for a parameter declared `{"type": "array", "items": {"type": "string"}}`:

```
<parameter=recommendations>
["Increase Grafana RAM limit on rk1-node-02 to prevent OOM-kill near-misses."]
["Address autovacuum backlog on PostgreSQL excalidraw.keyv table in rk1-node-03."]
["Test restore from Synology backup within 14 days per house rule."]
["Fix SSD cache write failures on omlx service due to permission errors."]
["Repoint remaining omlx clients to datum.home.arpa as per action backlog."]
</parameter>
```

The log shows:

```
WARNING - Parameter 'recommendations' of tool 'submit_recommendations' failed to parse as declared type 'array'; keeping raw string
WARNING - Native tool parser failed (SyntaxError: invalid syntax (<unknown>, line 2)), recovered via XML fallback.
```

and the client receives the whole block as a single string for a declared array. `_repair_json_value` is a bracket balancer; this text is balanced, so it returns `None` and `_coerce_param_value` falls through to the raw value.

## Change

Adds `_merge_json_fragments`, tried in the container branch before the bracket repair. It merges only when all of these hold: the declared type is an array; every whitespace-separated top-level fragment parses with the module's `_TOOL_CALL_JSON_DECODER.raw_decode`; the decoder consumes the entire string (trailing prose keeps the raw string); there are at least two fragments; and the fragments are homogeneous. `items.type` decides the merge shape: list fragments concatenate when items are scalars or objects, and wrap when items are themselves arrays, so an array-of-arrays parameter is never flattened. Scalar fragments and object fragments wrap. Any other shape returns `None` and the existing repair and raw-string fallbacks run unchanged. Declared string params are untouched.

Logs at WARNING like the existing repair path so the shape drift stays visible.

## Tests

Six cases added next to the existing `repairs_malformed_array` tests: concat for scalar items, wrap for array items, scalar fragments wrap, trailing prose keeps the raw string, mixed fragments keep the raw string, declared string param never merges. `tests/test_tool_calling.py` and `tests/test_parser_tool_calls.py` pass in full (400 tests) against the 0.6.4 dependency set.

https://claude.ai/code/session_011juTmek9tMttLA4fJ5BuAx


jundot (approved):
Thanks for the fix. I verified that the reported Qwen output now reaches the client as an array instead of a raw string. I'll tighten the fragment separator and item-shape checks in a small follow-up so ambiguous output stays unchanged. LGTM, I'll merge this and handle the follow-up.

PR #3602, opened 2026-09-12
Optimize Qwen4 HC, SSD PLE, GDN and QSA inference

mbarnson:
I've been noodling on oMLX to try to increase performance and benchmark scores for Qwen3.8-Flash-Next (Qwen4), particularly with PLE n-gram on-disk instead of in-RAM. There seems to be a number of potential optimizations. This PR increases Qwen4 with PLE offloading speed a bit.

128 GiB M4 Max, MLX 0.32.2: main `7cbb407` versus main + PR head `9ed173a`. Five alternating seed pairs with identical oQ4e weights, BF16 PLE offload/MTP, temperature 1.0/p0.95/k20, xhigh, draft temperature 0.6. The 39 successful tasks generated 584K tokens; coding responses ran 20K–36K tokens.

Four pairs completed in both arms:

| Mean metric | Main | Main + PR |
|---|---:|---:|
| Decode, tok/s | 45.96 | 50.13 |
| Time to first token, ms | 3081 | 3224 |
| Quality | 82.73 | 84.08 |

- **Decode: +9.07% mean paired gain**, range +7.52–10.44%; descriptive bootstrap 95% interval +7.74–10.39%.
- **TTFT: +143 ms** on average.
- **Quality: +1.36 points**, interval −1.39 to +3.88: no demonstrated improvement. Four fixed prompts, LLM-judged scores; adaptive MTP can produce different text at equal seeds.

Completion: main 4/5, PR 5/5; all-attempt quality 79.58 → 83.75. Main seed 6104 repeated code and hit the native 94.08 GB memory watermark after 67,329 streamed tokens: coding grade zero, invalid fallback throughput excluded, no reroll. Physical peaks: 95.66/93.51 GB against the 96 GB ceiling.

Scope:

- **HC:** compiled dispatch and BF16 injection, with numerical checks and the GPU tile-autotuning TODO.
- **SSD PLE:** early gathers, batched reads, lookahead sized after memory guards, reusable row buffers, direct BF16 views, warm-only re-arm checks and read-draining safeguards.
- **GDN/QSA:** shallow verification, norm/gate fusion and causal mask expansion, with state/boundary tests.

I did some experiments with an **optional dedicated MTP vocabulary head and resident PLE alias, but removed it due to marginal gains and a regression on M5**. Ordinary Lightning MTP and main's resident-PLE behavior remain and seem sufficient.

[...]

PR #3627, opened 2026-09-12
fix(qwen4): make disk-backed PLE resources fork-safe

pierre427:
## What changed

This makes oMLX's Qwen4 disk-backed PLE resources safe across failed loads, concurrent readers, descriptor reuse, and `fork()`:

- scope newly opened readers/embeddings to one model load and transfer ownership only after success;
- close embeddings before readers when strict loading fails, while preserving the original exception;
- serialize descriptor publication and cleanup without serializing row reads;
- refuse inherited readers in child processes before touching inherited locks, futures, executors, or MLX;
- drain submitted page reads before releasing descriptors after an error;
- make close idempotent under concurrent calls and exported mmap views.

## Why

The disk-backed PLE path owns files, mmap views, prefetch futures, and executors across model construction. A failed strict load could retain the partial model in a traceback and leave descriptors open. A forked child could also inherit locked synchronization state and hang before it could reject the reader. These are lifecycle failures in the existing offload path.

## Verification

- 8 runtime fork/lifecycle CPU tests passed, including a real held-lock fork and descriptor-reuse cases.
- 5 scoped-load ownership CPU tests passed, including concurrent success/failure and traceback-retained partial models.
- The unfixed extracted reader timed out in both held-lock and pending-future fork controls; the repaired tests complete within the bound.
- Ruff check and format check pass for all new source/tests; `git diff --check` passes.
- The new source module carries the required Apache-2.0 SPDX identifier.

## Scope and compatibility

- No format, quantization, hashing, or default-policy change.
- Matt Barnson's #3602 changes PLE page submission, draining, and batching in the same vendored file. The failed-batch drain pattern is credited to that work; this PR focuses on lifecycle ownership and fork safety. Because the branches overlap, it will need an explicit integration rebase if #3602 lands first.
- Composed GPU runs exercised PLE in both arms, so they establish mechanism engagement and bounded cache behavior, not an isolated PLE speedup.


mbarnson:
Thanks for calling out #3602. I trial-merged your `34ecf9a` with my updated branch (`68eb4449`, rebased onto `ea974149`). Three conflict hunks, all in `language.py`: imports and the page-prefetch helper split.

A couple of integration details beyond the conflict markers:

- My cross-shard warmup submits `reader._touch_page` directly, before `rows_np`. Those reads need your reader-lifetime protection too, including draining partially submitted batches. Locking each worker would deadlock the per-shard path while its caller holds the same reader lock; the batch needs to retain reader ownership through the drain.
- My new guarded-chunk lookahead and prefix/suffix buffer reuse fit inside your embedding call/close wrapper. The CPU AST extraction harness will need `_PLEShardSpec` and `_PLEPrefetch`, with `NamedTuple`/`Future` in its namespace, when exercising their construction and reuse.

Your load/fork/close protections should survive the integration intact. I haven't pulled them into #3602; the performance changes remain separate. First-token and long-decode retests are underway on the rebased code, so the old throughput numbers don't establish the combined lifecycle/performance result.

jundot (approved):
Thanks for fixing the PLE resource lifecycle. I reproduced the retained-traceback leak on main and verified that this change closes the readers and executor. The fork checks and dense/quantized PLE output checks also passed. LGTM.

PR #3628, opened 2026-09-12
perf(deepseek_v41): read offloaded experts with positional preads

beatakouchnir:
# perf(deepseek_v41): read offloaded experts with positional preads

Follow-up to #3574 / #986 (V4.1 expert adapter) and #2595 (expert offload primitive). Companion to #3589, which does the same for the common adapter.

## Problem

`omlx/patches/deepseek_v41/moe_offload.py` fetched non-resident experts through `TensorFile.read(key, rows=expert)`, the Engram row-gather path: an mmap under `MADV_RANDOM`, read by NumPy advanced indexing. That is the right shape for Engram (many scattered rows, each smaller than a page). An expert is the opposite: 14.8 MiB of contiguous bytes per expert at the oQ3e geometry, faulted in one 16 KiB page at a time with readahead disabled.

Measured on an M5 Max internal SSD with a synthetic checkpoint at the oQ3e expert geometry (384 experts, 3-bit affine gs64, 4 layers, random weights written with `F_NOCACHE` so the first pass is cold), 12.5% residency:

| | before | after |
|---|---:|---:|
| decode, one token, per MoE layer | 362 ms | 9.1 ms |
| expert fetch throughput (cold) | 0.23 GB/s | 9.3 GB/s |
| sorted prefill, 256 tokens | 33 token-layers/s | 568 token-layers/s |

Summed over 40 layers, the before path spends about 14 s per decoded token on expert reads at 12.5% residency; the after path about 0.35 s. That gap is what decides whether V4.1 is usable on a 128 GB machine, which is the configuration the offload setting exists for.

## Change

- Expert slabs are read with `os.preadv` on per-shard descriptors, submitted to a small reader pool of their own (24 workers) rather than the Engram page pool, whose long sequential page touches would otherwise queue ahead of a layer's misses. No knobs. `TensorFile` stays the Engram path; the header parser and the dtype table are shared with `residency.py` and `storage.py` instead of duplicated, and the tensor byte-length check is kept.
- `_ExpertSlots.ensure_ids` runs two passes: start the misses' reads ahead of the installs (at most `INFLIGHT_BYTES` of host payload in flight), then install serially in the order the misses were seen. Eviction victims, hit and miss counts, and resident bytes are identical to the serial path; the test pins the LRU order against a pure-Python reference at two in-flight windows. A failed read or decode leaves the cache exactly as it was (no slot leak, every started read drained before the error propagates), which the previous serial path did not guarantee either.
- Sorted prefill routes are chunked on expert boundaries: every route of up to `capacity` distinct experts per chunk, instead of `capacity` rows per chunk. A prefill reads each expert once per layer and runs one kernel per chunk (with enough routes for the affine block kernels to engage), where the row-count chunking ran one kernel per 48 rows.
- `admission_bytes(path, fraction)` and `fit_resident_fraction(path, budget_bytes)` give the engine pool's admission estimate for a fraction and the largest fraction whose estimate fits a byte budget. The test pins both to `EnginePool._entry_runtime_resident_size` for every capacity, with and without Engram tables (the pool has two branches there). `docs/MoE_Expert_Offload.md` gets the resident set per fraction for `Jundot/DeepSeek-V4.1-Flash-oQ3e-mtp` on a 128 GB Mac.
- `benchmarks/deepseek_v41_offload_bench.py` loads a checkpoint through the loader with Engram on SSD and offload at each fraction, then runs chunked prefill and greedy decode, reporting memory, TTFT, decode speed, and per-phase hit rate and fetch throughput.

## Validation

- `tests/test_deepseek_v41_moe_offload.py` (46 with `test_moe_offload_compat.py`): existing cases unchanged except the load probe, which now asserts expert reads are single-expert slabs that never touch the Engram mapping; new cases for serial LRU order under concurrent reads, the failure path (cache intact, no slot leak, later reads work, idempotent close), expert-boundary chunking with arithmetic equality against the resident Expert, and admission/fit against the engine pool with and without Engram tables.
- `test_deepseek_v41.py`, `test_deepseek_v41_offload.py`, `test_deepseek_v41_moe.py`, `test_engine_pool.py`, `test_model_settings.py`, `test_moe_expert_offload.py`, `test_admin_model_settings_template.py`: 392 pass.
- Real checkpoint, `Jundot/DeepSeek-V4.1-Flash-oQ3e-mtp` on a 128 GB M5 Max (internal SSD, `iogpu.wired_limit_mb` unset), Engram on SSD, native kernels built, 433-token prose prompt in one prefill chunk plus 64 greedy tokens, single runs:

  | residency | experts/layer | load | Metal active | peak footprint | prefill | decode | decode hit rate |
  |---:|---:|---:|---:|---:|---:|---:|---:|
  | 12.5% | 48 | 3.9 s | 38.0 GiB | 49.6 GiB | 28 tok/s | 5.6 tok/s | 0.69 |
  | 25% | 96 | 4.5 s | 65.7 GiB | 77.4 GiB | 25 tok/s | 4.2 tok/s | 0.78 |

  Coherent continuations at both settings. Prefill reads each routed expert once per layer (about 226 of 384 per layer for this prompt) at 8 to 9 GB/s. Decode is miss-latency bound at one to two misses per layer per token; 12.5% decodes faster than 25% because the RAM the slots do not take becomes page cache, which serves repeated misses at 6.5 GB/s effective against 3.4 GB/s from the SSD. The first, repeated-sentence run at 12.5% gave 4.0 tok/s at a 0.81 hit rate. 33% was not measured: the session's memory guard stopped a run once the page cache had taken all free RAM (macOS itself was not under pressure).
- `omlx serve` end to end on the same machine: the checkpoint is discovered from the HF cache (companion PR), admitted at 39.98 GB against 38.44 GB actual, Engram forced to SSD by admission, VLM engine load 4.4 s; two 64-token chat completions at 2.7 tok/s (cold expert slots) and 4.1 tok/s.

## Not in this PR

DSpark MTP under offload stays excluded. On this path decode is bound by expert reads, and a 5-token verification block touches about 4.6x the experts of a single token at both 12.5% and 25% residency (measured on the synthetic checkpoint: 283 vs 1343 misses per 16 steps at 12.5%, 197 vs 906 at 25%), so DSpark's 1.6 to 2x tokens per step would net roughly 0.4x the throughput. It only pays once residency is high enough that reads stop being the bottleneck, which is not a 128 GB configuration.

The `mlx-community/DeepSeek-V4.1-Flash-MLX-2bit` checkpoint (222 GiB, per-expert affine layout, no `omlx_deepseek_v41` spec) does not load through the V4.1 loader at all. That is a separate loader-side gap.

### DIFF
--- a/benchmarks/deepseek_v41_offload_bench.py
+++ b/benchmarks/deepseek_v41_offload_bench.py
@@ -1,0 +1,233 @@
+# SPDX-License-Identifier: Apache-2.0
+"""DeepSeek V4.1 expert offload on one machine: load, prefill, decode.
+
+Loads a V4.1 checkpoint through the oMLX loader with Engram on SSD and MoE
+expert offload at each requested resident fraction, then runs a chunked
+prefill and a greedy decode directly on the model (no scheduler), reporting
+load time, Metal and process memory, time to first token, decode speed, and
+the expert cache's hit rate and fetch throughput per phase. Prints the
+resident set per fraction from the shard headers and the largest fraction
+that fits the Metal working-set limit first, so a run that cannot fit is
+visible before any weights load::
+
+    python benchmarks/deepseek_v41_offload_bench.py --model /path/to/oQ3e \\
+        --fractions 0.125 0.25 --prompt-tokens 512 --decode-tokens 32
+"""
+
+from __future__ import annotations
+
+import argparse
+import gc
+import json
+import time
+from pathlib import Path
+
+import mlx.core as mx
+
+from omlx.utils.proc_memory import get_lifetime_max_phys_footprint, get_phys_footprint
+
+
+def _offload_stats(model) -> dict:
+    hits = misses = fetched = 0
+    for layer in model.language_model.layers:
+        slots = getattr(layer.ffn.experts, "slots", None)
+        if slots is None:
+            continue
+        hits += slots.hits
+        misses += slots.misses
+        fetched += slots.fetched_bytes
+    return {"hits": hits, "misses": misses, "fetched_bytes": fetched}
+
+
+def _delta(after: dict, before: dict, seconds: float) -> dict:
+    misses = after["misses"] - before["misses"]
+    hits = after["hits"] - before["hits"]
+    fetched = after["fetched_bytes"] - before["fetched_bytes"]
+    return {
+        "hits": hits,
+        "misses": misses,
+        "hit_rate": hits / max(1, hits + misses),
+        "fetched_gib": fetched / 2**30,
+        "fetch_gbps": fetched / max(seconds, 1e-9) / 1e9,
+    }
+
+
+def sizing(path: Path, budget: int, engram_ssd: bool) -> dict:
+    from omlx.patches.deepseek_v41.moe_offload import (
+        _plan,
+        admission_bytes,
+        fit_resident_fraction,
+    )
+
+    plan = _plan(path, 1.0)
+    rows = []
+    for fraction in (0.125, 0.25, 1 / 3, 0.375, 0.5, 1.0):
+        capacity = min(plan.count, max(plan.floor, round(plan.count * fraction)))
+        rows.append(
+            {
+                "fraction": fraction,
+                "experts_per_layer": capacity,
+                "admission_gib": admission_bytes(
+                    path, fraction, engram_ssd_offload=engram_ssd
+                )
+                / 2**30,
+            }
+        )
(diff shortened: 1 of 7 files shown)


jundot:
Thanks for implementing this. Positional reads and expert-boundary chunking are a useful improvement for V4.1 offload. I found that completed futures retain their read buffers until `ensure_ids()` returns, exceeding the intended memory window. I'll merge this and address buffer release in a small follow-up with a regression test.

PR #3629, opened 2026-09-12
fix(discovery): find DeepSeek V4.1 checkpoints in the HF cache

beatakouchnir:
# fix(discovery): find DeepSeek V4.1 checkpoints in the HF cache

Companion to #3583 (which adds `format: mlx` metadata to newly exported oQ shards). Checkpoints already published without it, such as `Jundot/DeepSeek-V4.1-Flash-oQ3e-mtp`, still never appear in the model list after `hf download`: `_is_hf_cache_mlx_compatible` needs shard metadata, an MLX token in the repo name, or a helper checkpoint, and an oQ export has none of the three.

## Change

Accept HF-cache entries with `model_type: deepseek_v41` whose config carries a version-1 `omlx_deepseek_v41` spec (oMLX conversion), or no spec and no top-level `quantization` dict (the FP8/FP4 release and bf16 sources, which the loader gates on the model type alone). MLX affine conversions without the spec (for example `mlx-community/DeepSeek-V4.1-Flash-MLX-2bit`) are not accepted by this rule; the loader has no path for them, so listing them would only move the failure to load time. The `mlx-community/` name rule still lists that repo as before.

## Validation

`tests/test_model_discovery.py`: new parametrized case covering the oMLX spec, the original FP8 config, a bf16 source, an unknown spec version, an affine conversion with and without the FP8 config, and a non-V4.1 model type with the spec; 197 pass.

### DIFF
--- a/omlx/model_discovery.py
+++ b/omlx/model_discovery.py
@@ -1408,6 +1408,25 @@ def _is_helper_checkpoint(model_path: Path) -> bool:
     return is_helper_model_config(config)
 
 
+def _is_deepseek_v41_loadable_config(config) -> bool:
+    """True for DeepSeek V4.1 checkpoints the V4.1 loader reads unconverted.
+
+    The loader gates source checkpoints on the model type alone (the FP8/FP4
+    release, or bf16), and reads oMLX conversions by their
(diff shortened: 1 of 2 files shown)


jundot:
Thanks for fixing this. The config check matches the V4.1 loader’s supported formats and covers older exports without MLX metadata. I verified that source and converted checkpoints are discovered from the HF cache and load successfully. LGTM, I’ll merge this.

PR #3631, opened 2026-09-13
fix(cluster): recover quiescence and allow reload when distributed runtime fails

xunlinkx:
## Summary

When a distributed worker crashes or terminates unexpectedly (e.g. from an unrecoverable GPU timeout, Metal watchdog kill, or sudden disconnect), the coordinator engine was wedged in an unrecoverable state:
1. Rank zero's runtime marker file (`<deployment_id>-rank-0.json`) remained on disk recording positive `active_requests`.
2. `DistributedBatchedEngine.rank_side_active_requests()` read this stale marker file and continued returning a positive count.
3. `DistributedBatchedEngine.has_active_requests()` could remain positive if internal request state / leases were not drained due to the abrupt termination.
4. When coordinator routes attempted to reload the failed deployment (`await pool.prepare_cluster_reload(model_id)`), `EnginePool._raise_if_reload_busy` rejected the operation with `ModelBusyError(model_id, 'activate distributed cluster')` (HTTP 409), preventing automatic recovery or admin reload without a full server restart.

## Changes

1. **`omlx/engine/distributed.py`**:
   - In `rank_side_active_requests()`: Return `0` if `self.runtime_failed_reason` is set, if the marker owner process is dead (`not marker_owner_is_live(marker)`), or if the marker reports an `error`.
   - In `has_active_requests()`: Return `False` immediately if `self.runtime_failed_reason` is set, allowing quiescence checks to proceed.
2. **`omlx/engine_pool.py`**:
   - In `prepare_cluster_reload()`: Bypass `_raise_if_reload_busy()` when `entry.engine` has an active `runtime_failed_reason`, allowing the pool to force-unload the dead distributed engine and reload cleanly.
3. **Tests**:
   - Added unit test in `tests/test_distributed_engine.py` verifying `rank_side_active_requests()` and `has_active_requests()` report `0` / `False` when the worker process dies or has a runtime failure.
   - Added unit test in `tests/test_engine_pool.py` verifying `prepare_cluster_reload()` successfully unloads a failed distributed engine without raising `ModelBusyError`.

## Verification

- Ran `pytest tests/test_distributed_engine.py tests/test_engine_pool.py tests/test_cluster_routes.py` (all 262 tests passed).
- Deployed and tested live on a dual-Mac cluster serving Qwen3.8.

### DIFF
--- a/omlx/engine/distributed.py
+++ b/omlx/engine/distributed.py
@@ -29,6 +29,7 @@ from ..cluster.liveness import (
     check_peers,
     describe_failure,
     marker_age_seconds,
+    marker_owner_is_live,
     read_marker,
 )
 from ..reasoning_effort import _fallback_candidate, _normalized_input
@@ -910,12 +911,17 @@ raise SystemExit(2)
         quiescence evidence an abort or unload should wait for (G5).
         """
 
+        if self.runtime_failed_reason is not None:
+            return 0
+
         marker = read_marker(
             Path(self._supervisor.state_dir).expanduser()
             / f"{self.deployment.deployment_id}-rank-0.json"
         )
         if not isinstance(marker, dict):
             return None
+        if not marker_owner_is_live(marker) or marker.get("error"):
+            return 0
         metrics = marker.get("metrics")
         if not isinstance(metrics, dict):
             return None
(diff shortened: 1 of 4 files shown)


jundot (approved):
Thanks for fixing this. I reproduced the stuck reload on my M3 Ultra and M2 Max over TCP Ring by killing the worker during a stream. With this change, the pool completes verified teardown and the reloaded cluster serves requests again, while reloads during healthy active requests are still refused. LGTM, I'll merge this.

PR #3632, opened 2026-09-13
Bound chat output by remaining context

PaddyPatPat:
## Summary
- enforce input plus completion against the effective total context window
- clamp omitted/default output budgets to the exact remaining capacity
- expose exact available output capacity for explicit over-budget requests
- forward chat template controls through Anthropic token counting so thinking-disabled counts match generation

## Verification
- `pytest -q tests/test_context_window.py tests/integration/test_server_endpoints.py` (98 passed)
- live Qwen3.8-27B verification found `enable_thinking: false` renders 40 fewer prompt tokens than the current count route reports without this forwarding


jundot:
Thanks for adding context-aware output limits. I found two things I'd like addressed before merging.

First, an OpenAI-compatible client may send a fixed `max_tokens` equal to the configured context window, expecting generation to stop at EOS. With both set to 131072, this change rejects every nonempty prompt. Please clamp explicit output budgets to the remaining capacity instead of rejecting them, just as this PR already does for omitted budgets.

Second, please resolve automatic `enable_thinking` and `preserve_thinking` options before counting prompt tokens. With the Gemma tokenizer and `thinking_budget: 10`, I reproduced a count of 14 tokens before automatic thinking activation and 16 afterward. The current clamp therefore allows an output budget of 86 with a context limit of 100. Moving option resolution ahead of the existing count should avoid an additional counting pass.

Please cover explicit budgets through both `max_tokens` and `max_completion_tokens`, along with automatic thinking activation in streaming and non-streaming endpoint tests. The checks should verify that oversized output budgets are accepted and reduced, and that the final rendered prompt plus the budget passed to the engine stays within the limit.

PR #3633, opened 2026-09-13
fix: apply finite ModernBERT attention mask to rerankers

ernestas-poskus:
**Root cause:** `mlx_embeddings`' stock `_update_attention_mask` uses `-1e9`, which overflows to `-inf` in fp16, so a fully padded (short) query produces **NaN**. It was already fixed for the *embedding* path via `omlx/patches/modernbert_attention.py` (referencing issue #3507 / `Blaizzy/mlx-embeddings#80`), but the **reranker's `mlx_embeddings.load` branch never applied the patch**. NaN is serialized by pydantic as JSON `null`, hence `"relevance_score": null`.


jundot:
Thanks for catching the missing patch in the reranker load path. I reproduced the NaN scores with mixed-length fp16 inputs and confirmed that applying the patch before compilation fixes both compiled and eager execution. LGTM, I'll merge this.

PR #3634, opened 2026-09-13
perf(image): cache decoded images by content hash to avoid per-turn re-decode

williamxie1989:
## Problem

Multi-turn agent loops (and any chat that accumulates screenshots) resend the
**same** historical images on every turn as byte-identical `data:` URIs. On the
server, `extract_images_from_messages()` → `load_image()` re-ran the CPU-bound
PIL decode (`Image.open` + `exif_transpose` + `convert("RGB")`) for **every
image in the full message history, every turn**.

Once a conversation accumulates a few dozen screenshots, this per-turn re-decode
dominates time-to-first-token. On a DeepSeek-V4.1 (multimodal) agent session at
~200K context we measured, per turn:

```
15:19:25,921  request received
15:19:25,9xx  PIL.PngImagePlugin STREAM b'IDAT' … (~160 decode events)
   …          ~22 s of image decode, no other work logged
15:19:49,5xx  Loaded block from hot cache … (restore ~150 ms)
15:19:55,377  MTP path activated → first token
```

The ~22 s is spent **before** prefill, so it never shows up in the prefill
progress bar, and because the text suffix is tiny the "Prompt Processing
(excl. cached)" figure keeps falling while TTFT climbs — which is exactly the
symptom reported (TTFT rising, no visible prefill, dashboard values dropping).

## Fix

Cache the decoded RGB image by the SHA-256 hash of its source bytes. Repeated
turns reuse the prior decode; only genuinely new images are decoded.

- Key: `sha256(img_bytes)` — hashing the bytes is ~1 ms vs a ~100 ms+ decode.
- Value: the decoded RGB `PIL.Image`. Consumers (`image_patches`, mlx-vlm
  processors) only read it; the previous code never closed it, so sharing the
  object across turns is safe.
- Bounded: LRU capped by total decoded pixel bytes (512 MiB default).
- Thread-safe: the lock only guards the dict, never the decode, so concurrent
  requests still decode in parallel. A late duplicate insert reconciles the
  byte accounting.

Decode count for an agent that grows to K screenshots over K turns drops from
`K·(K+1)/2` (quadratic) to `K` (linear) — the cliff is what disappears.

## Tests

`tests/test_image_utils.py::TestLoadImageDecodeCache` (RED before, GREEN after):
- identical bytes decoded once across calls (`Image.open` called once)
- distinct images decoded independently (no false hits)
- `clear_image_decode_cache()` forces re-decode
- cached pixels identical to a cold decode

## Verification

Studio (Mac Studio, DeepSeek-V4.1 multimodal), simulate an agent resending
cumulative 1600×1000 screenshots over 8 turns:

```
cache OFF: decodes=36  time=0.16s
cache ON : decodes=8   time=0.04s
decodes avoided: 28 (78%)
```

Full fast suite on the PR base (excluding a pre-existing hang in
`tests/test_cluster_seams.py`, unrelated to this change): green.

## Notes

- `clear_image_decode_cache()` is exported so the memory enforcer / admin can
  free the tier under pressure.
- The existing `compute_image_hash` dedup runs *after* decode, so it never
  avoided the decode itself; this cache sits at the decode choke point.


williamxie1989:
## Real-server A/B validation (Mac Studio, DeepSeek-V4.1 multimodal)

Ran a real multi-turn chat against the running server, resending cumulative
1600x1000 screenshots each turn, one request at a time (never stacked),
baseline (original `image.py`) vs this patch — same box, same workload:

| turn | images | baseline TTFT | patched TTFT | delta |
|---|---|---|---|---|
| 1 | 1 | 7.75s | 5.86s | -24% |
| 2 | 2 | 7.77s | 5.05s | -35% |
| 3 | 3 | 10.82s | 8.37s | -23% |
| 4 | 4 | 17.11s | 11.73s | -31% |
| 5 | 5 | 22.29s | 14.54s | -35% |
| 6 | 6 | 26.87s | 21.14s | -21% |

DEBUG-log PIL decode events in the run windows:

- baseline: **336 events / 168 IDAT decode chunks** (every turn re-decodes all
  cumulative images: 1+2+...+6 = 21 images)
- patched: **12 events / 6 IDAT chunks** (only the new image per turn decodes;
  the 15 repeated re-sends are skipped)

=> **86% of PNG decode-chunk work eliminated on the real server path.** TTFT
still rises with turns from legitimate cost (prefill + vision *encode* of each
new image + growing prompt); the cache removes only the repeated *decode* of
resent images, which is the ~22s pre-prefill block seen at ~200K/160-image
sessions.

**Generality**: this per-turn full-decode path is in `engine/vlm.py`
(`extract_images_from_messages` -> `load_image`) and is common to **all**
non-diffusion VLMs, not v4.1-specific. The existing vision feature cache only
caches post-*encode* features, never the decode. Text-only sessions send no
images and are unaffected.

PR #3640, opened 2026-09-13
feat(embedding): promote bfloat16 checkpoints to float16 for conformance

ernestas-poskus:
mlx_embeddings.load() has no dtype parameter, so a bfloat16 embedding checkpoint computes its matmuls in bf16 and misses the 1e-3 conformance gate (measured max|delta| 0.003720 vs 0.000593 for the identical weights computed in fp16). Add a per-model embedding_dtype setting (auto|float16|float32) and cast the loaded model's floating parameters before compilation; auto promotes a bf16 checkpoint to fp16. Threaded through EmbeddingEngine from engine_pool.


jundot:
Thanks for the change. I also confirmed that FP16 reduced the numerical difference from the FP32 reference on Qwen3-Embedding-8B.

Both results are from the Qwen3-Embedding family, so I think automatically converting all BF16 embedding models to FP16 needs broader testing across model families, especially given FP16's narrower numerical range.

mlx-embeddings doesn't apply this conversion in its loader either, so I'm cautious about introducing a separate default dtype policy in oMLX without that broader validation.

ernestas-poskus:
@jundot maybe you have happened to test this?

jundot:
@ernestas-poskus I meant that I'm hesitant to apply BF16-to-FP16 conversion to all embedding models when mlx-embeddings doesn't do this upstream, rather than that I would run broader testing myself.

The numerical improvement has been confirmed for the specific Qwen3-Embedding models tested. Could you narrow the PR to those models and leave other models unchanged? That would keep the change within the validated scope and reduce the risk of accuracy regressions elsewhere.

ernestas-poskus:
@jundot Makes total sense — thank you for clarifying!

I've updated the PR accordingly:

1. Restricted auto-promotion to the Qwen3-Embedding 
2. Added unit tests covering the Qwen3-specific auto-promotion, non-Qwen3 preservation, detection heuristics, and explicit overrides.

jundot:
@ernestas-poskus The current detection still includes Qwen3-VL and quantized variants. Could you limit automatic conversion to the tested non-quantized Qwen3-Embedding 0.6B and 8B models?

I'd also prefer removing the general embedding_dtype override from this focused change. Its FP32 cast and settings integration remain incomplete, and it adds configuration maintenance beyond the validated fix.

PR #3647, opened 2026-09-13
fix(mtp): retain head history for generated prefix blocks

ashhart:
Generated tokens can enter the backbone prefix cache without a matching MTP head-history snapshot. A later turn can restore the backbone block and still lose the drafter's history. This publishes the matching in-memory sidecar when an eligible Qwen4 request emits a committed full-block boundary.

Registration accepts both `qwen4_exp` and `qwen4_exp_text` on the eligible MTP host. The native Qwen4 language model reports `qwen4_exp_text`, even when its outer configuration uses `qwen4_exp`.

Plans belong to a model and scheduler UID, and capture requires a complete prompt-history handoff. The snapshot holds a detached head cache at boundary minus one and its pending normalized hidden row. Publication happens after successful emission, including terminal cleanup, and the existing cache handles hashing, its four-entry LRU and backbone eviction. Media metadata, stale ownership and incomplete history decline capture.

Boundary materialization also needs to trim the previous draft chain before folding the confirmed row into a shared head cache. Without that trim, a depth-three cycle can leave a speculative row inside the committed history. The full-history oracle reproduced this mismatch and passes on CPU and GPU with the trim restored. Heads that draft on a separate cache clone keep their existing behavior.

The focused MTP, scheduler, VLM routing and native Qwen4 compatibility suites pass 410 tests with three skips, including the test order that exposed the failure. The tiny random model covers depths one and three, shared and cloned head caches, and CPU and GPU execution. It requires bit-exact copies of live head state at four generated boundaries and identical output tokens with capture enabled and disabled. Independent full-history comparisons run on both devices at the unchanged 2e-5 tolerance. The test fixture sets its own draft depth so model-loading tests cannot change its schedule, and the VLM fallback fixture includes the scheduler's optional cache field.

A native tiny Qwen4 language model tests registration and cleanup directly and through both language-model wrapper attributes. Publication, restoration and media exclusion cover both supported host names. The five registration/publication regressions fail with the original host-type check and pass with the fix.

This remains a draft pending full-model repeated-prompt and long-context memory tests. The tiny Qwen3.5 fixture exercises the fold/cache contract but does not qualify a full Qwen4 checkpoint. This retains full contiguous head history in memory and complements the cached-suffix reconstruction proposed in #3328. It adds no SSD format or partial-block storage, and makes no end-to-end throughput claim.


jundot:
Thanks for working on this. I tried the draft with Qwen3.8-Flash-Next-oQ4e-mtp and found that the MTP host reports `qwen4_exp_text`, so the `qwen4_exp` check in `register()` skips generated-history capture. With only that check adjusted in my test runner, I confirmed snapshot publication and restoration at a 2,048-token boundary. Sharing this while the PR is still in progress; I haven’t qualified long-context memory or throughput yet.

PR #3653, opened 2026-09-13
fix(mtp): keep the Qwen3.5 MTP __call__ pipeline-aware (#3518)

### LINKED ISSUE #3518: The Qwen3.5 MTP patch replaces `__call__` with a non-pipeline-aware version, breaking pipeline-parallel serving
## Summary

`omlx/patches/mlx_lm_mtp/qwen35_model.py` replaces `qwen3_5`'s inner `TextModel.__call__` with an MTP variant that is not pipeline-aware. Because `utils/model_loading.py` applies that patch **even when `mtp_enabled` is `False`**, every pipeline-parallel `qwen3_5` deployment crashes.

Verified against `origin/main` @ 94530d8d.

## Why the patch is applied with MTP off

`omlx/utils/model_loading.py:718`:

```python
if _is_mtp_compatible(config, model_type):
    mtp_enabled = bool(
        model_settings is not None and getattr(model_settings, "mtp_enabled", False)
    )
    from ..patches.mlx_lm_mtp import (apply_mlx_lm_mtp_patch, ...)
```

The surrounding comment explains this is deliberate — `sanitize` needs the patch for the `+1` norm shift whenever `mtp.*` keys are present. But `apply_mlx_lm_mtp_patch` also replaces `__call__`, and that replacement drops three things upstream's `mlx_lm/models/qwen3_5.py` does.

## The three differences

Upstream (`mlx_lm/models/qwen3_5.py`) vs. the patched `__call__`:

| upstream | patched |
| --- | --- |
| iterates `self.pipeline_layers` | iterates `self.layers` |
| guards `if self.fa_idx is not None` / `if self.ssm_idx is not None` | indexes `cache[self.fa_idx]` / `cache[self.ssm_idx]` unguarded |
| does `mx.distributed.recv_like` / `send` / `all_gather` | no collective calls at all |

`apply_pipeline_assignment` (`omlx/cluster/planner.py`) sets non-local layers to `None`:

```python
pipeline_model.layers = pipeline_model.layers[: assignment.end_layer]
pipeline_model.layers[: assignment.start_layer] = [None] * assignment.start_layer
```

so iterating `self.layers` hits them.

## Observed failures

Rank dies during the readiness inference:

```
File ".../omlx/patches/mlx_lm_mtp/qwen35_model.py", line 425, in __call__
    mask = ssm_mask if layer.is_linear else fa_mask
AttributeError: 'NoneType' object has no attribute 'is_linear'
```

And, when a stage happens to contain no `linear_attention` layer (`ssm_idx` is then `None`):

```
    ssm_mask = create_ssm_mask(hidden_states, cache[self.ssm_idx])
TypeError: list indices must be integers or slices, not NoneType
```

## Impact

`qwen3_5` is one of the architectures that passes oMLX's own gate. Running `pipeline_assignment_is_honored` inside `install_unequal_pipeline_plan([])` over `mlx_lm.models`:

```
PASS: deepseek_v2, deepseek_v3, glm4_moe, glm4_moe_lite, ministral3, qwen3_5
FAIL: deepseek_v32 (fail-closed by design), mistral3
```

`_is_mtp_compatible()` limits the patch to `qwen3_5`/`qwen3_6`/`deepseek_v4`/`nemotron_h`/`glm_moe_dsa`/`gemma4`/`inkling`, so the GLM and DeepSeek families are unaffected — `glm4_moe` pipelines without hitting any of this. The problem is confined to the MTP-compatible set, but for those it is unconditional.

## Possible fixes

1. Don't replace `__call__` when MTP is inactive — the `sanitize` norm-shift concern doesn't require it.
2. Make the replacement pipeline-aware.

For (2) this is what I ran locally, keeping the MTP contract of returning **pre-norm** hidden states:

```python
layers = getattr(self, "pipeline_layers", None)
if layers is None:
    layers = self.layers
pipeline_rank = getattr(self, "pipeline_rank", 0)
pipeline_size = getattr(self, "pipeline_size", 1)

if cache is None:
    cache = [None] * len(layers)

fa_mask = ssm_mask = None
if getattr(self, "fa_idx", None) is not None:
    fa_mask = create_attention_mask(hidden_states, cache[self.fa_idx])
if getattr(self, "ssm_idx", None) is not None:
    ssm_mask = create_ssm_mask(hidden_states, cache[self.ssm_idx])

if pipeline_rank < pipeline_size - 1:
    hidden_states = mx.distributed.recv_like(hidden_states, (pipeline_rank + 1))

for layer, c in zip(layers, cache):
    mask = ssm_mask if layer.is_linear else fa_mask
    hidden_states = layer(hidden_states, mask=mask, cache=c, n_confirmed=n_confirmed)

if pipeline_rank != 0:
    hidden_states = mx.distributed.send(hidden_states, (pipeline_rank - 1) % pipeline_size)
    if cache[-1] is not None:
        if hasattr(cache[-1], "keys"):
            cache[-1].keys = mx.depends(cache[-1].keys, hidden_states)
        else:
            cache[-1][0] = mx.depends(cache[-1][0], hidden_states)

if pipeline_size > 1:
    hidden_states = mx.distributed.all_gather(hidden_states)[: hidden_states.shape[0]]
```

With that applied on coordinator and workers, both ranks reach `phase: ready` and the deployment activates. I'm raising this as an issue rather than a PR because option (1) may be the design you'd prefer, and because generation still stalls afterwards for reasons I have not root-caused (I'll file that separately if it turns out to be distinct).

## Environment

oMLX 0.6.2 and 0.6.4, verified still present on `main` @ 94530d8d. Mac Studio M5 Max (Metal coordinator) + NVIDIA DGX Spark GB10 (CUDA worker), ring backend over Ethernet, model `mlx-community/Qwen3.8-27B-8bit` re-expressed as a text-only variant.

jonathan308:
### Summary

Fixes #3518: the Lightning MTP patch replaces `Qwen3_5TextModel.__call__` whenever the checkpoint is MTP-compatible — **even with `mtp_enabled` off**, because `sanitize` needs it — and the replacement dropped upstream's pipeline path. Every pipeline-parallel `qwen3_5` deployment crashed: `apply_pipeline_assignment` None-fills non-local layers, the replacement iterated `self.layers` straight into them, indexed `cache[self.fa_idx]` unguarded, and skipped `recv`/`send`/`all_gather` entirely.

### Fix

The replacement now mirrors upstream's pipeline flow exactly (`mlx_lm/models/qwen3_5.py:267-316`):

- `self.pipeline_layers` when `pipeline_size > 1` (plain `self.layers` otherwise)
- `fa_idx`/`ssm_idx` guarded with `is not None` before mask construction
- `recv_like` / `send` with the cache-state dependency anchor (both cache families, as upstream)
- `all_gather` with the shape-restoring slice

The MTP contract is preserved verbatim: pre-norm hidden return so the wrapping `TextModel.__call__` can fuse the MTP head, and `n_confirmed` threading through the layers.

### Testing

Two new tests using the real patched class with a placeholder instance (no weights):

- **Pipeline survival** (fails before with the issue's crash): rank 1 of 2 — the None-filled non-local layers are never called, the local layer runs, `send` to rank 0 and `all_gather` both fire.
- **Single-node unchanged** (fails before on the unguarded `cache[None]`): rank 0 of 1 behaves exactly as before, `n_confirmed=0`.

Regression (fresh venv, Python 3.12.11): MTP patch + prompt priming + GLM MTP + pipeline compat + inference worker + DeepSeek MTP + depth controller = **380 passed**.

### Remaining validation limits

A physical 2-rank `qwen3_5` pipeline generation is not re-run here; the pipeline flow is mirrored line-for-line from upstream and pinned at the contract level (layers called, collectives ordered). Distributed MTP remains rejected elsewhere — this fix only restores the pipeline path for ordinary distributed serving.


ashhart:
On current main this fix isn't reached for planned pipelines. Since the mlx-lm pin update in #3708, `Qwen3_5TextModel.pipeline(group, split=None)` calls `super().pipeline(group, split=split)`, and so do the Qwen3-Next and Ministral 3 overrides. `install_unequal_pipeline_plan` swaps `PipelineMixin.pipeline` for a hook that takes no `split`, so every planned pipeline of these models fails at load, before the forward runs:

```
TypeError: install_unequal_pipeline_plan.<locals>.planned_pipeline() got an unexpected keyword argument 'split'
```

The fix in `omlx/cluster/planner.py` is small. The plan owns the layer ranges, so the hook only needs to accept the argument and refuse an explicit value:

```python
    def planned_pipeline(pipeline_model: Any, group: Any, split: Any = None) -> None:
        # Qwen3.5, Qwen3-Next and Ministral 3 override ``pipeline`` and forward
        # their ``split`` argument (default None) to the mixin. The plan owns
        # every stage's layer range, so an explicit split is refused.
        if split is not None:
            raise PlanningError(
                "an explicit pipeline split conflicts with the installed shard plan"
            )
        apply_pipeline_assignment(pipeline_model, group, assignments)
```

I had this in #3856 together with the same forward change as yours, and I've closed that in favour of this PR. The commit is ashhart/omlx@34021ecc47383a064541a28dbaf1d6d78b5fc5c7 if you want its planner test and a two-stage Qwen3.5 test that compares the last stage with the unsplit forward. With both changes, two MLX ring ranks on one Mac running a planned split, where one stage holds only full-attention layer 7, decode a two-prompt batch exactly like a single-process run, with `sampling_rank_only` on and off. Happy for you to fold the planner change in here.

PR #3654, opened 2026-09-13
perf(moe): chunk over-capacity prefill on expert boundaries

beatakouchnir:
Follow-up to #2595 (expert offload primitive). Brings the common adapter's over-capacity prefill to the shape the DeepSeek V4.1 adapter already uses (#3574 / #3628).

## Problem

`OffloadSwitchGLU.__call__` handled a prefill that routes to more distinct experts than the cache holds by halving the token axis until every chunk's working set fit, then running the chunks in order. An expert routed to by several chunks was fetched again in every one of them, with LRU evicting it in between. On `gemma-4-26b-a4b-it-4bit` a 585-token prompt at 12.5% residency caused 64,369 expert fetches per prefill for 3,840 layer-experts in the model, and `docs/MoE_Expert_Offload.md` listed prefill as "the clearest follow-up".

## Change

Over-capacity prefill is chunked on expert boundaries: the routes are sorted by expert on the host, cut into chunks holding every route of up to `capacity` distinct experts, each chunk's experts made resident through the existing `ExpertCache.ensure`, the chunk run with one expert index per route, and the outputs put back in route order once at the end. Each expert is installed at most once per call by construction. The kernel choice mirrors the stock `SwitchGLU`'s rule for the call as a whole (sorted at or above the 64-index threshold, unsorted below), so every chunk runs the kernel the resident model would have used; each chunk is evaluated before the next is built, which bounds the prefill transient as before.

Decode and within-capacity calls are untouched (same fast path). The change is confined to `OffloadSwitchGLU`; `ExpertCache` and the store are not modified, so it composes with #3589 (parallel pread installs) rather than overlapping it.

## Numerical contract

Unchanged in kind: bit-exact where the kernel path is identical (below the sort threshold both sides run the unsorted kernel; the existing `test_chunked_prefill_bit_exact_below_sort_threshold` now exercises the new path and still passes), rounding-scale where it is not. Two tests added: every distinct expert is fetched exactly once in an over-capacity prefill (fetch spy), and route order is restored with rounding-scale agreement to the resident model above the threshold.

## Measured

`benchmarks/moe_offload_prefill_bench.py` (new, loads through `lm_load_compat` + `apply_moe_expert_offload` + `materialize_lazy_state`, no scheduler), `mlx-community/gemma-4-26b-a4b-it-4bit`, M5 Max 128 GB, 585-token prompt, 32 decode tokens, single runs. Before = `ea974149`, after = this branch. Warm = second identical request; cold = first request after load (page cache cold, pays the initial fill).

| residency | capacity | prefill expert fetches, before → after | TTFT warm, before → after | TTFT cold, after | decode tok/s before / after | process GiB |
|---|---|---|---|---|---|---|
| 50% | 64 | 6,073 → 1,719 | 2.38 s → 0.69 s | 1.64 s | 65.4 / 60.6 | 9.09 |
| 25% | 32 | 30,302 → 2,675 | 8.87 s → 0.85 s | 1.51 s | 43.6 / 43.5 | 5.75 |
| 12.5% | 16 | 64,369 → 2,913 | 16.60 s → 0.97 s | 11.92 s | 31.6 / 31.7 | 3.78 |

The before numbers reproduce the doc's existing table (18.1 s / 9.6 s / 2.9 s TTFT at 12.5 / 25 / 50%). Decode differences are single-run noise; the decode path is not touched.

## Tests

`pytest tests/test_moe_expert_offload.py tests/test_moe_offload_compat.py`: 46 passed (3 slow deselected). `ruff` clean on the changed files; the new benchmark is `black`-formatted; existing files were not reformatted.


jundot (approved):
Thanks for the implementation. Grouping each expert's routes together removes the repeated loads while evaluating each chunk before its cache slots are reused. I checked the 585- and 2,048-token Gemma paths at 12.5% residency; both matched the previous implementation's 32 greedy output tokens, with substantially lower TTFT. LGTM, I'll merge this.

PR #3657, opened 2026-09-13
fix(cluster): report a failed transport probe as unknown, not Ethernet (#3021, #3030)

### LINKED ISSUE #3021: detect_transports reports a failed probe as a measured Ethernet link

jonathan308:
### Summary

Fixes #3021 and #3030 (same defect): `detect_transports` caught every probe exception, discarded the reason, and then synthesised `TransportInfo(kind="ethernet")` for every host pair — a probe that could not run became indistinguishable from a probe that ran and measured Ethernet.

The fabricated value is strictly worse than no data: `transports_are_fast_enough(())` reads an **empty** result as "link unknown" and declines to assume the worst, but a named Ethernet link is a *measured* reason to refuse tensor parallelism. The failure path therefore produced the one outcome the no-data path deliberately avoids — #3030's 40 Gb/s Thunderbolt pair (SSH auth failed before a single interface was read) was refused TP on evidence that was never gathered.

### Fix

Track probe failure across both probe branches and only synthesise Ethernet when every probe **ran** and found no fast link — the measured answer. A failed probe returns the empty tuple ("unknown"), and `choose_parallelism`'s existing unprobed-transport warning covers the UX ("Transport was not probed, so tensor parallelism was chosen without knowing the link speed").

Interaction note: this touches the same `try` blocks as #3645's D4 guard (different lines, same region); whichever lands second rebases trivially.

### Testing

- Failed probe → `()`, not fabricated Ethernet (fails before).
- Successful probe with no fast links → measured Ethernet preserved (passes before and after, pinning the distinction).
- Regression (fresh venv, Python 3.12.11): link_status + autoconfigure + routes = **188 passed**.

### Remaining validation limits

Physical #3030 topology (TB4 pair with failing SSH) not re-run here; the behavior change is confined to the failure-vs-measured boundary, both sides pinned.


jundot:
Thanks for putting together all these fixes. With several draft PRs now open, please mark each one ready for review as it's finalized, and I'll review them in order. If any need to be merged in a specific order, or there's anything I should know before reviewing, I'd appreciate a single consolidated note covering those details.

Issue #3648, opened 2026-09-13
Admin UI: “Expose as model / API ON” missing from New Profile form


ubrandes:
**Describe the bug**

Creating a model profile lets the user assign an API name, which strongly implies it is API-addressable, but Expose as model is only available after saving and reopening the profile with the pencil icon.

**To Reproduce**
1. Go to a model's model settings
2. Click on the button to create a new model profile
3. Look for a button to expose this profile on the API
4. It's not available until I save and re-edit the profile using the pencil icon.

**Expected behavior**
When creating a model profile I'd expect to be able to expose it to the API straight upon creation. 

 - oMLX v0.6.4

**Additional context**
There are more settings here that would be great to expose, such as "mtp_enabled" (is this the same as "Lightning MTP"?) and  "mtp_num_draft_tokens".

PR #3658, opened 2026-09-14
fix(admin): offer the API toggle on the New Profile form

L4XB:
Refs #3648.

## The gap

Creating a model profile lets you assign an **api name**, which implies the profile is API-addressable. It was not: `Use as API model` lived only in the edit dialog, so a new profile had to be saved and then reopened with the pencil icon before it could be reached — exactly the sequence the reporter described.

The backend was never the problem. `CreateProfileRequest` has carried `expose_as_model: bool = False` all along, and `tests/test_admin_profiles_api.py::test_profile_requests_accept_expose_as_model_flag` already covers it. The gap was entirely in the form: no toggle, and `createProfile()` builds its POST body by hand and never named the field.

## The change

- `_modal_model_settings.html`: the API ON/OFF button joins the api-name row of the new-profile form, using the edit dialog's markup and its existing i18n key (`modal.model_settings.profiles.expose_as_model`, already present in en/zh/ja/ru), so **no locale gains a string**.
- The form's reset object initialises `expose_as_model:false`, so the flag does not survive from a previous open into the next profile.
- `dashboard.js`: `createProfile()` sends `expose_as_model: !!this.newProfile.expose_as_model`.

**The default stays off.** Exposing a profile on the API is not a neutral default, and a profile created without asking for it must not become addressable — there is a cell for that.

## Deliberately not included

The issue also asks about `mtp_enabled` and `mtp_num_draft_tokens` ("there are more settings here that would be great to expose"). Those are a different question — which of the model settings belong on a creation form at all — and I did not want to answer it inside a fix for the toggle that was already half-present. Happy to take it separately if you want it.

## Tests

`tests/test_admin_new_profile_expose_as_model.py`, 6 cells, in the same text-reading style as `test_admin_model_settings_template.py` and the other admin-UI regressions here — there is no browser in CI to click with.

Each was checked against a mutant:

| mutation | killed by |
| --- | --- |
| no toggle in the form (the bug) | `the_new_profile_form_offers_the_api_toggle` + `the_toggle_reads_and_writes_the_same_flag` |
| create body drops the field | `create_sends_the_flag_to_the_endpoint` + `the_default_is_still_off` |
| the form opens with it already ON | `the_form_resets_the_flag_when_it_opens` + `the_default_is_still_off` |
| the toggle only displays, never writes | `the_toggle_reads_and_writes_the_same_flag` |
| the edit dialog loses its own toggle | `the_edit_dialog_still_has_its_own_toggle` |

That last one is the accept control: adding a toggle to the create form must not have moved the one that already worked.

```
$ pytest tests/test_admin_new_profile_expose_as_model.py --noconftest
6 passed

$ pytest tests/test_admin_model_settings_template.py tests/test_model_settings_profiles.py \
         tests/test_admin_dashboard_draft_filters.py --noconftest
20 + 56 + 4 passed
```

`--noconftest` because `tests/conftest.py` imports `mlx`, which this machine does not have; these cells read the template and the script as text and need nothing from it. `ruff check` and `ruff format --check` clean on the new file.


jundot (approved):
Thanks for fixing this. The toggle uses the existing create API and keeps exposure off by default. I checked ON/OFF submission and the reset when reopening the form. LGTM; I’ll merge this and trim the redundant comments and overlapping tests in a small follow-up.

PR #3659, opened 2026-09-14
fix: honor Qwen4-Exp MTP expert counts

mensaprodigy:
## Summary

- add optional MTP-specific MoE expert-count overrides to `TextConfig`
- apply those overrides only when constructing the embedded Qwen4-Exp MTP decoder
- preserve main-model counts when an override is absent

This supports checkpoints whose main MoE and MTP draft head use different expert counts (for example, 384 and 512).

## Testing

- `pytest tests/test_qwen4_exp_mtp_expert_count.py tests/test_qwen4_exp_mtp_admin_gate.py -q` (11 passed)
- `python -m compileall` for the changed Qwen4-Exp vendor package and regression test
- `ruff check tests/test_qwen4_exp_mtp_expert_count.py`

The regression coverage verifies explicit MTP overrides, full fallback to main-model values, and independent fallback for each optional override.


jundot (approved):
Thanks for the fix. The MTP-specific counts are applied only to the draft head, with existing defaults preserved. I’ll update the per-expert weight loading path in a follow-up to use the same counts. LGTM, merging.

mensaprodigy:
> Thanks for the fix. The MTP-specific counts are applied only to the draft head, with existing defaults preserved. I’ll update the per-expert weight loading path in a follow-up to use the same counts. LGTM, merging.


I appreciate the update. Look forward to it!

PR #3663, opened 2026-09-14
fix(cluster): recover quiescence and unwedge reload on failed distributed workers

xunlinkx:
### Summary

When a distributed worker fails (due to an eviction, SIGKILL, crash, or peer connection drop), subsequent cluster reload or unload attempts could deadlock with:
`HTTP 409: The model is already changing state; wait and try again.`

### Root Cause Analysis
1. **Stuck Pending Unloads**: If an unload was previously requested or queued, or drain timed out, `entry.pending_unload_reason` remained set and `_wait_for_pending_unload()` polled forever because `_entry_is_quiescent()` / `_entry_has_active_requests()` did not recognize a dead worker as quiescent.
2. **Reload Rejection**: In `load_cluster_deployment()`, `pool.prepare_cluster_reload()` was only called if `resident.runtime_failed_reason` was already set. If `entry.pending_unload_reason` was set or if the failed reason had not yet populated, `pool.get_engine()` immediately raised `ModelBusyError(model_id, "start work while unload is pending")`.
3. **Pending Task Cleanup**: `prepare_cluster_reload()` did not cancel the background `_pending_unload_tasks` or clear `pending_unload_reason`.
4. **Stale Telemetry Marker**: `rank_side_active_requests()` did not check marker freshness, returning positive active requests from a dead process if the marker was left behind without an error field.

### Key Changes
- **`omlx/engine_pool.py`**:
  - In `prepare_cluster_reload()`: cancel any pending unload task in `_pending_unload_tasks`, clear `entry.pending_unload_reason`, `entry.pending_unload_allow_pinned`, and `entry.abort_requested`.
  - In `prepare_cluster_reload()`: bypass `_raise_if_reload_busy()` when `resident.runtime_failed_reason` is a non-empty string.
  - In `_entry_has_active_requests()` and `_entry_is_quiescent()`: treat an engine with `runtime_failed_reason` as quiescent with 0 active requests.
- **`omlx/cluster/routes.py`**:
  - In `load_cluster_deployment()`: invoke `prepare_cluster_reload()` if `resident.runtime_failed_reason` is set OR if `entry.pending_unload_reason` is set.
- **`omlx/engine/distributed.py`**:
  - In `runtime_failed_reason`: inspect supervisor status whenever supervisor is attached.
  - In `rank_side_active_requests()`: treat markers older than `_DEFAULT_STALE_AFTER` (45s) as 0 active requests.
- **Tests**:
  - Added test coverage in `test_engine_pool.py`, `test_cluster_routes.py`, and `test_distributed_engine.py`.

Verified locally with full test suite passes.


jundot (approved):
The failed-worker recovery is useful. I reproduced one edge case: reloading a healthy engine that is still draining returns HTTP 409 but clears its queued unload, so the model stays loaded after the request finishes. I'll move the cleanup after the busy check and cover that case in a follow-up fix.

PR #3664, opened 2026-09-14
fix(responses): route namespace tool groups through to the model and back

### LINKED ISSUE #3371: Responses API drops `type: "namespace"` tools — Codex MCP tools never reach the model
**Describe the bug**

`/v1/responses` silently discards every tool whose `type` is not `"function"`, including
`type: "namespace"` groups. A namespace group is not a hosted/server-side tool — it is a
*container of ordinary client-executed function tools*. Dropping it means every tool inside
it disappears before the chat template is built, so the model never sees them and never calls
them. No error is returned; the request succeeds with a plain text answer.

Codex CLI wraps each MCP server's tools in exactly this shape, so **no MCP tool is usable in
Codex against oMLX**. This is not oMLX's own MCP feature (`mcp.expose_tools`, `--mcp-config`),
which works fine — it is *client-supplied* tools arriving over the wire.

Two things make this a hard block rather than an inconvenience:

1. Codex removed the Chat Completions wire. `WireApi::Responses` is `#[default]` and
   `wire_api = "chat"` now hard-errors with "`wire_api = \"chat\"` is no longer supported"
   (`codex-rs/model-provider-info/src/lib.rs:57-67`, openai/codex discussion #7782). There is
   no fallback transport.
2. oMLX's own Codex launcher lands on this path. `omlx/integrations/codex.py:97-115` writes
   `model_providers.omlx.base_url` and no `wire_api`, so every oMLX-launched Codex session
   uses `/v1/responses`.

`namespace` is also a documented, public Responses API tool type, not a Codex-private
extension — see OpenAI's tool-search guide and the `responses.create` API reference.

**To Reproduce**

Three requests against a running oMLX server (`Qwen3-30B-A3B-Instruct-2507-4bit`, already
loaded). Set `OMLX_API_KEY` to your server key first.

*A. Namespace-wrapped tool — the bug.*

```bash
curl -s -X POST http://127.0.0.1:8004/v1/responses \
  -H "Authorization: Bearer $OMLX_API_KEY" -H "Content-Type: application/json" -d '{
  "model": "Qwen3-30B-A3B-Instruct-2507-4bit",
  "input": "Use the get_weather tool to get the weather in Paris.",
  "stream": false, "max_output_tokens": 200,
  "tools": [{
    "type": "namespace", "name": "mcp__demo__", "description": "Demo MCP server",
    "tools": [{
      "type": "function", "name": "get_weather",
      "description": "Get the current weather for a city.",
      "parameters": {"type":"object","properties":{"city":{"type":"string"}},
                     "required":["city"],"additionalProperties":false}
    }]
  }]
}'
```

HTTP 200. No `function_call` item. The model answers that it cannot use external tools:

```json
{"status":"completed",
 "output":[{"type":"message","role":"assistant","content":[{"type":"output_text",
   "text":"I can't directly use external tools like `get_weather` to fetch live data. ..."}]}],
 "usage":{"input_tokens":20,"output_tokens":86}}
```

`input_tokens: 20` is the tell — the tool schema contributed nothing to the prompt.

*B. Identical tool, declared flat — the control.*

Same request with the tool moved to the top level as `{"type":"function","name":"get_weather",...}`:

```json
{"output":[{"type":"message","content":[{"type":"output_text","text":""}]},
           {"type":"function_call","id":"fc_31ab1cd3","call_id":"call_bd223088",
            "name":"get_weather","arguments":"{\"city\": \"Paris\"}","status":"completed"}],
 "usage":{"input_tokens":169,"output_tokens":19}}
```

The tool works. 169 input tokens vs 20 — the schema is present. Note the returned
`function_call` carries no `namespace` field; that matters for the fix (see below).

*C. Codex-shaped payload — one native function plus a namespace group.*

Prompt: "List every tool you can call, by name. Do not call any of them." Tools: a top-level
`exec_command` function plus an `mcp__demo__` namespace containing `get_weather` and `search`.

```
TEXT: The only tool available is `exec_command`.
input_tokens: 166
```

The model confirms directly that the two namespaced tools are invisible to it while the flat
one survives.

*Codex end to end.* Configure Codex against oMLX the way `omlx/integrations/codex.py` does
(`-c model_provider=omlx -c model_providers.omlx.base_url=http://127.0.0.1:8004/v1`) with any
MCP server registered, then ask the model to use an MCP tool. It reports it has no such tool.
No error appears in either the Codex or the oMLX log.

**Expected behavior**

Function tools declared inside a `namespace` group are exposed to the model exactly like
top-level function tools, and a call to one is returned with its namespace intact:

```json
{"type":"function_call","name":"get_weather","namespace":"mcp__demo__",
 "call_id":"call_...","arguments":"{\"city\":\"Paris\"}"}
```

**Screenshots**

N/A — full request/response bodies above.

**Desktop**

- macOS Version: 15.7.2 (24G325), Apple M3 Ultra, 512 GB
- oMLX Version: 0.6.4 (also present on `main` @ `2246290a`)
- codex-cli 0.148.0
- Model: Qwen3-30B-A3B-Instruct-2507-4bit

**Additional context**

*Root cause.* `convert_responses_tools()` in `omlx/api/responses_utils.py:401-433` keeps only
`tool.type == "function"`, with the comment:

> Non-function tools (local_shell, mcp, web_search, etc.) are silently skipped — local models
> can't execute them.

That reasoning is correct for `local_shell`, `web_search` and hosted `mcp`, which the server
would have to execute itself. It does not hold for `namespace`, whose children are plain
function tools the *client* executes. `ResponsesTool` sets `extra="allow"`
(`omlx/api/responses_models.py:58-71`), so the nested `tools` list is parsed and retained —
it is simply never read. `create_response()` calls the converter at `omlx/server.py:6138`.
`tests/test_responses.py:705-719` currently codifies the skip for `local_shell` and `mcp`.

*Why the return path matters too.* Flattening alone is not enough. Codex resolves an incoming
call by `(namespace, name)`: `build_tool_call` does
`ToolName::new(namespace, name).with_default_namespace()`
(`codex-rs/core/src/tools/router.rs:250-264`), a missing namespace defaults to `"functions"`
(`codex-rs/protocol/src/tool_name.rs:39-44`), and a registry miss becomes
`unsupported call: {tool_name}` (`codex-rs/core/src/tools/registry.rs:459-463`, `:821`). Codex
emits MCP tools as `ToolSpec::Namespace` at `codex-rs/core/src/tools/handlers/mcp.rs:491-495`
and accepts `namespace` on returned items (`codex-rs/protocol/src/models.rs:1047`). A call
returned as a bare flattened name is rejected — confirmed against Ollama in
[openai/codex#23186](https://github.com/openai/codex/issues/23186) (comment of 2026-08-18),
where restoring `{namespace, name}` made it work end to end and a flat name did not.

*Suggested fix.* In `convert_responses_tools`, expand namespace groups and remember the
mapping; when emitting a `function_call`, restore the split. Roughly:

```python
# responses_utils.py — in convert_responses_tools()
if tool.type == "namespace" and tool.name:
    ns = tool.name
    for child in (getattr(tool, "tools", None) or []):
        child = child if isinstance(child, dict) else child.model_dump()
        if child.get("type") != "function" or not child.get("name"):
            continue
        # matches Codex's join_tool_name(): trim the joint, then "__"
        wire = f"{ns.rstrip('_')}__{child['name'].lstrip('_')}"
        alias[wire] = (ns, child["name"])
        result.append({"type": "function", "function": {
            "name": wire,
            "description": child.get("description"),
            "parameters": child.get("parameters"),
        }})
```

Then, wherever a `function_call` item is built, look the emitted name up in that map and set
`namespace` plus the inner `name`. There are two emission sites today — the streaming path in
`omlx/server.py` (~`:7107-7175`) and `build_function_call_output_item` in
`responses_utils.py:451-466` — so routing both through one helper keeps them in step.
`defer_loading` on a child can be ignored (treat the tool as always loaded);
`{"type":"tool_search"}` can keep being dropped. Worth pairing with a namespace case beside
the existing `test_non_function_tools_skipped`.

*Workarounds today.* Users must run a translating proxy. opencodex does the flatten in
`src/responses/namespace-tool-compat.ts` (`rewriteRoutedNamespaceToolsForUpstream`) and the
restore in the same file (`restoreRoutedNamespaceCalls`), applied at
`src/adapters/openai-responses.ts:2185`. The same shim exists in
`bharat2808/codex-ollama-proxy`, `madmax24-ubuntu/codex-proxy-llama.cpp`, and
`swobuforge/swobu`. Handling it natively in oMLX would remove an extra hop from every local
Codex session.

*Related.* Same class of problem in other backends: llama.cpp (`ggml-org/llama.cpp#20156`),
LM Studio (`lmstudio-ai/lmstudio-bug-tracker#1810`), Ollama (`openai/codex#19871`), plus
`openai/codex#20652`, `#22970`, `#24297`. This report is oMLX-specific and independent of
oMLX PR #2830, which fixes the *server-side* MCP tool merge (a different feature).

tc3oliver:
Fixes #3371

Root cause: `convert_responses_tools()` keeps only `type == "function"`, so a `namespace` group — a container of ordinary client-executed function tools, the shape Codex uses for each MCP server — is skipped whole and its members never reach the chat template. Flattening alone is not enough either: clients resolve a call by `(namespace, name)`, so a bare flattened name is rejected.

Fix: expand a group under a joined wire name (the same join as Codex `join_tool_name()`, suffixed if a flat tool already holds it) and remember `wire -> (namespace, name)`. Both the streaming and non-streaming `function_call` emitters split it back, so a call returns as `{"name": "get_weather", "namespace": "mcp__demo__"}`. Flat function tools are unchanged, and `namespace` serializes only on the items that have one.

Tests:
- `python -m pytest` — 12990 passed, 344 skipped, 82 deselected in 301.76s (0:05:01).
- Focused Responses coverage: `tests/test_responses.py`, `tests/integration/test_e2e_streaming.py`, `tests/test_server.py`, `tests/test_cluster_routes.py`, and `tests/integration/test_server_endpoints.py`.
- The new `/v1/responses` regression test covers both streaming and non-streaming namespace-tool round trips and fails on `main`.

Not tested with a real model; this change is limited to Responses API tool conversion and response shaping, not model inference or cache behavior.


jundot (approved):
The namespace expansion and response mapping address the tool-drop cause. I verified the round trip with Codex CLI 0.154.0 and a local MCP fixture, using deterministic model output.

I also found that continuation history drops the namespace. I've prepared a small maintainer follow-up that preserves it and maps history to the current tool names, with coverage for name collisions and stored conversations. LGTM with that follow-up; I'll merge this and apply it together.

Issue #3660, opened 2026-09-14
0.7.0.dev1/dev2: Qwen3.8-Flash-Next quants emit immediate EOS instead of tool calls in assistant-less contexts (0.6.4 clean, byte-identical A/B)


layer4down:
## Summary

Since 0.7.0.dev1, Qwen3.8-Flash-Next oQ quants (reproduced primarily on `Qwen3.8-Flash-Next-oQ5e-mtp`, once on an oQ4e build) emit **immediate EOS instead of tool calls** in long assistant-less contexts. The final message is whitespace-only (`"\n\n"`), `finish_reason: "stop"`, zero `tool_calls`, while the reasoning channel ends mid-plan (e.g. reasoning ends with "I'll read the config:" and then the text channel emits whitespace and stops). 0.6.4 handles the byte-identical requests correctly. **0.7.0.dev2 does not fix it — it's worse** (details below).

## Environment

- oMLX builds: 0.6.4 (`260830015308`) ✅ clean · 0.7.0.dev1 (`260911022938`) ❌ · 0.7.0.dev2 (`260911232929`) ❌ (build numbers from `Resources/omlx/_build_info.py`)
- Apple M2 Ultra 192GB, macOS 26.x
- Model: `Qwen3.8-Flash-Next-oQ5e-mtp` (community GBP-DE build; single weights file used across all three engines)
- Settings: temperature 1.0 (also affected at other temps), MTP/Lightning OFF, ANE OFF, DFlash OFF, SpecPrefill OFF, TurboQuant KV OFF, thinking on

## The A/B (byte-identical requests, only the engine differs)

The exact same streaming `/v1/chat/completions` request — same weights file, same system prompt, same message array, same params — replayed against each version:

| Version | Trials | Outcome |
|---|---|---|
| 0.6.4 | 5 | **0 EOS — all 5 emit `finish_reason: "tool_calls"`** (model engages the code tool) |
| 0.7.0.dev1 | 10 | **10 EOS** — `stop` / 0 `tool_calls` / whitespace-only content |
| 0.7.0.dev2 | 5 | **5 EOS** — identical signature |

Timeline corroboration: after a server restart onto dev1, the first occurrence appeared **9 minutes later**, with zero occurrences across weeks on pre-0.7 engines including the same quant on agentic workloads.

## Reproduction (synthetic — no private content needed)

[...]

PR #3669, opened 2026-09-15
fix(api): flush rejected tool-call envelopes as content in stream_chat_completion

layer4down:
*Disclosure: this PR was prepared by an AI agent for @layer4down with their authorization — GLM-5.3 (Z.ai) in Claude Code v2.1.271. The investigation itself was run largely by an agent using the same Qwen3.8-Flash-Next-oQ5e weights on oMLX 0.6.4, testing against a 0.7.0.dev2 source build. Questions on methodology welcome.*

## TL;DR — why this matters

**The bug:** on 0.7.0.dev builds, oMLX can silently throw away a tool call the model just made. The client sees a clean "stop" with no tool call and no text — so AI agents built on oMLX just halt mid-task with nothing to show and nothing to debug. This is #3660.

**The model is not at fault.** We instrumented the model's forward pass: it emits a complete, well-formed tool call every time. The server throws it away (details below).

**The fix (commit `dd77208a`):** when the server rejects a tool call, return it as visible text instead of discarding it. One behavior change: *nothing the model emits can silently vanish.* Clients that previously got an empty reply now see the raw call and can react — and the maintainer gets a bug report instead of a ghost.

**Bonus fix (commit `8b184f71`):** while chasing the above we found an unrelated, real GPU bug — some SDPA kernels launch with thread groups larger than certain Apple GPUs allow. Metal silently skips those launches and leaves stale/uninitialized memory. The repo's own fp32 tests fail on M2-class GPUs because of it. The fix adopts mlx's own residency checks (parity, not invention).

## What the user sees, before → after

| Scenario | Before | After |
|---|---|---|
| Model calls a tool that isn't in the request's `tools` list | Empty reply, `finish=stop`, no tool call, no text — looks like the model gave up | The raw tool call is returned as text — visible, debuggable |
| Model calls a registered tool (the normal case) | Works | Works, unchanged (3/3 replays identical behavior) |
| SDPA decode on GPUs with small thread-group limits | Silent wrong output (repo's own tests fail on M2-class) | Correct output, tests pass |

## How we know (evidence chain, short form)

1. Hooked the model's forward pass: the "missing" tool call is fully emitted — every token top-1, envelope complete, then a legitimate end-of-turn.
2. Unit-tested the parser on the exact captured payload: it parses fine, including unregistered names.
3. Traced the drop to one branch: a name-registration check added in #3327 discards parsed-but-unregistered calls with a bare `continue`, while a stream filter has already hidden the raw text. Both channels empty = the ghost stop.
4. Version archaeology matches: that check exists only in dev1/dev2 (`git tag --contains`), absent in 0.6.4 — exactly the version matrix reported in #3660.
5. Applied the fix, replayed the failing requests byte-identical: 27/27 silent failures before → **5/5 visible output after** (byte-identical to the instrumented trace). Regression test added.

## Scope & honesty notes

- The gate-drop class explains the #3660 signature on dev builds. We're not claiming it fixes *every* whitespace-EOS report; other parser paths are untouched.
- We separately observe long-context stalls on 0.6.4 itself (~90k cached tokens) — a different signature, still under investigation, not addressed here.
- Reproduction keys on prompt shape (contexts that induce an unregistered function name), not prompt length or hardware — see the analysis in #3660 for why in-grammar test prompts won't hit it.

## Validation environment

M2 Ultra 192GB, `Qwen3.8-Flash-Next-oQ5e-mtp`, byte-exact replays of the failing requests; static AST suite 3/3 (incl. new regression test); decode_fast fp32 d=128 tests pass on g14d where pristine dev2 fails.

🤖 Generated with [claude-flow](https://github.com/ruvnet/claude-flow)


layer4down:
**Agent disclosure:** this comment was prepared and posted by an AI agent acting for @layer4down, with their review and authorization: **GLM-5.3 (Z.ai)** running in **Claude Code v2.1.271**.

Closing this PR for now — follow-up testing in real agent use surfaced a quality problem with the chosen recovery behavior.

**Reproduction:** in a live web-agent harness session (single registered tool, `run_code`; model attempts `todo_write`, which is not registered), the gate correctly rejects the envelope and this patch flushes it — but the result is the complete raw `<tool_call>` XML appearing in user-visible content. The turn still ends `finish_reason=stop` with no structured tool call, the model receives no actionable feedback, and a follow-up "try again" nudge produced an identical rejected call: a deterministic retry loop. The commit message's "clients at worst see the raw call as text" branch turns out to be the common case for unregistered-tool attempts, not a rare edge.

The patch does prevent the silent drop (#3660's invisible empty replies), and we still believe suppress-then-drop is wrong. But trading an invisible failure for visible raw-XML leakage in user content isn't a behavior we'd stand behind upstream. Two candidate directions to explore before resubmitting:

1. On gate rejection of a well-formed envelope naming an unregistered tool, surface a structured error to the caller (or synthesize a tool-error round) so the model can self-correct to a registered tool.
2. Alternatively, pass unregistered-name calls through as structured `tool_calls` and let the client respond with the error — closer to how OpenAI-compatible clients expect unknown-tool calls to be handled.

Will reopen with a revised approach once validated.

🤖 Generated with [claude-flow](https://github.com/ruvnet/claude-flow)

PR #3672, opened 2026-09-15
feat: SSD-backed MoE expert streaming with dynamic budgets

errrepe:
## Summary

SSD-backed expert streaming for MoE models that exceed unified memory.
Routed experts live in safetensors banks on disk; a byte-bound slot
arena keeps the working set resident under a dynamic budget governor.

- **Generic core** (`omlx/patches/expert_streaming/`): slot arena,
  slot cache (LRU / S3-FIFO), residency, governor, shard bank
  (lock-free preadv reads), conversion, telemetry, warmer
- **DeepSeek V4.1 adapter** (`omlx/patches/deepseek_v41/`): native
  offloaded experts ride the same residency machinery through
  `V41StreamingBacking`; the engine forwards `preserve_mtp` so the
  resident DSpark draft head survives loads, and `verify_scope`
  freezes slot recency during draft verify
- **One shared path**: `expert_streaming_enabled` is canonical;
  `moe_expert_offload_enabled` aliases to the same backend. The legacy
  adapter is the fallback when generic conversion converts nothing —
  no duplicated per-model implementations
- **Dynamic budgets**: governor grows/shrinks residency from free-RAM
  watermarks (10% clear / 20% shrink / 40% grow) plus a windowed
  decode-stall signal; ceiling = min(24 GiB, RAM/4), overridable per
  model. Engine signature keys on *requested* settings so transient
  pressure never triggers reload oscillation
- **MTP gate** (`OMLX_MTP_EXPERT_STREAMING_GATE`, auto/on/off): under
  SSD streaming the verify pass reads experts across draft positions —
  measured net regression on every model, so auto mode probes cheaply
  and parks MTP when a bounded cache is live. DeepSeek V4.1 is exempt:
  its DSpark verify runs under frozen residency
- **Fail-clean**: conversion raises when SSD backing can't be built;
  loaders re-raise before materializing expert banks
- **Spill loader**: shard fingerprint (size, mtime_ns) invalidates
  stale spill stacks before re-stacking
- **Per-model knobs stamped per-instance** — no leakage into shared
  model classes
- WebUI: settings fields, bank controls, i18n (9 locales), profile
  save no longer silently drops non-form fields. Swift app: the same
  streaming settings (dynamic budget, bank, prefill budget, pins) in
  Model Settings with DTO round-trip coverage

## Benchmarks — MacBook Pro M4 Pro, 48 GB RAM, checkpoints on NVMe SSD over USB4

All runs: `--budget auto` (dynamic governor, arena default-on),
single-request TTFT, everything streamed from SSD. 2k prompt /
128 decode.

| Model | Arm | TTFT | Decode | tok/s | Peak RAM |
|---|---|---:|---:|---:|---:|
| Qwen3.8-Flash-Next-oQ4e-mtp (99 GB ckpt) | base | 25.4s | 38.0s | **3.37** | 24.7 GiB |
| Qwen3.8-Flash-Next-oQ4e-mtp | MTP | 26.9s | 41.7s | 3.07 | 29.3 GiB |
| GLM-5.3-Flash-oQ4e-mtp (169 GB ckpt) | base | 80.9s | 137.8s | **0.93** | 35.8 GiB |
| DeepSeek-V4.1-Flash-oQ4e-mtp (402 GB ckpt) | base | 122.6s † | 147.0s † | **0.87** † | 31.5 GiB † |
| DeepSeek-V4.1-Flash-oQ4e-mtp | MTP (DSpark) | 125.8s † | 170.8s † | 0.75 † | 39.7 GiB † |

† not re-measured at this revision — the V4.1 load hits a Metal
working-set OOM inside `_ExpertSlots` init on this machine state
(reproduces identically on the pre-refactor head); earlier numbers kept.

(GLM-5.3 MTP arm omitted: the supplied oQ4e-mtp export contains no
`language_model.mtp.*` tensors — the head was not written.)

Rows re-measured after the consolidation commits (e83cca39).

Governor observed live: Qwen grew 2.4→12 GiB (cap 2739→13695) under
decode hunger; GLM shrank 546→136 under prefill pressure, regrew to
516 during decode. Sustained ~1.9 GB/s SSD reads in decode.

**MTP under streaming is a measured loss** — verify reads experts for
M draft positions against a cold cache: −9% (Qwen, gate parked early:
hit_rate 0.27 → 7 cycles), −14% (V4.1 DSpark, accept 62.9% but SSD I/O
per emitted token grew). The gate exists to keep MTP off this regime.

Prior mechanism ablation (same core, fixed budget): preadv span reads
run 7–8× faster than mmap gather at expert-row sizes — the dominant
lever behind the TTFT/decode deltas vs the first streaming revision.

## Out of scope (separate PRs)

GLM-5 MTP runtime, Qwen4-Exp adapter, custom kernels, cluster code,
unrelated ANE work, and benchmark artifacts.

## Test plan

- [x] Focused tests green (slot arena, cache policies, shard bank,
      dynamic budget, conversion guard, layout hooks, MTP gate, spill,
      admin/profiles, unified offload compat, V4.1 offload/DSpark/
      span-read suites)
- [x] `tests/moe_expert_offload_ui.test.cjs` (modal gates, save/
      migrate/reopen, speculative exclusion, V4.1 MTP exemption)
- [x] `xcodebuild build-for-testing` + `oMLXTests` (Swift settings
      DTO round-trip, V4.1 MTP exemption, bank wire keys)
- [ ] Live A/B on maintainer hardware


errrepe:
@jundot — scope update. This PR now also contains the DeepSeek V4.1 adapter and the Swift settings UI, so the earlier note about follow-up PRs no longer applies. The rest of this comment explains the domain, the design, and why the diff is this size.

## The problem this solves

A MoE model activates only a few experts per token, but the checkpoint still holds all of them. When the checkpoint is larger than the RAM of the machine, the current options are slow paging or a fixed resident fraction that either wastes memory or starves the model.

This PR streams routed experts from SSD into a byte-bound arena. A governor grows the arena when free RAM allows it and shrinks it under pressure. The model runs at full quality while resident memory stays small. On a 48 GB machine it runs a 402 GB checkpoint.

## How a request flows

1. The loader converts expert weights into safetensors banks on disk.
2. Each MoE call asks the residency layer for its routed experts.
3. Hits return slots in the arena. Misses read from the shard bank with positional reads, then enter the cache.
4. A speculation layer stages predicted experts early. Staged entries that no request used are dropped, so speculation is cheap.
5. The governor watches free RAM and decode stalls. It grows the budget when the system has room and shrinks it before pressure hurts other apps.
6. The macOS page cache acts as a free second level. This is why the private cache ceiling stays at a quarter of RAM. A larger private cache would evict page cache one to one and still pay a physical read on each miss.

## What each new file does

`omlx/patches/expert_streaming/` is the generic core. `slot_cache.py` and `residency.py` own the arena and the hit or miss path. `shard_bank.py` does the positional reads. `governor.py` sizes the budget. `conversion.py` writes the banks. `speculation.py` and `adaptive_topk.py` reduce how many experts a token needs. `cache_policies.py` holds LRU and S3-FIFO. `expert_bank_pack.py` packs experts into one fused bank for fewer and larger reads. `model_hooks.py` gives each model family its layout hooks. `streaming_switch.py` wires the feature on per model. `telemetry.py`, `memtrace.py`, and `hstrace.py` report and trace. `warmer.py` prefetches at load. `qwen35_stream_eval.py` covers eval-time paths.

`omlx/patches/deepseek_v41/` is the model adapter. It keeps the V4.1 storage format, quantization, and DSpark draft head. `streaming_backing.py` connects the native offloaded experts to the shared dynamic residency. `moe_offload.py` supplies `verify_scope`, a context that freezes slot recency during draft verify. That freeze is why V4.1 can combine MTP with expert streaming when other families cannot.

The Swift files mirror the WebUI controls in the macOS app: the streaming toggle, dynamic mode, budget fields, bank path, and pin controls, plus the DTO fields and profile state that carry them.

## Why this many lines

The diff is 100 files at about 28.7k added lines. Three parts make most of it. The generic core is one shared path that every supported family uses. The V4.1 adapter exists because that model has its own storage format and its own speculative head. The rest is UI wiring in two apps, i18n in nine locales, and tests. Custom kernels, cluster code, unrelated ANE work, and benchmark artifacts are not in this PR.

## What changed since the first version

The V4.1 adapter and the Swift streaming settings are now inside this PR. The V4.1-specific gate exemption is legitimate now because the adapter ships here. The upstream CED code was removed to match the adapter: its `ced_prefill` flag, its admin wiring, its modal toggle, its cache paths, and its tests. Swift tests that construct the model DTO gained the new capability fields.

errrepe:
## What each new file does

This PR adds expert streaming: MoE (mixture-of-experts) weights stay on SSD and only the routed experts enter RAM. Most of the line count is the new `expert_streaming` package plus its tests. The tables below name the job of each new file.

### New production files

| File | Lines | What it does |
|---|---:|---|
| `omlx/patches/expert_streaming/streaming_switch.py` | 3874 | The core. A drop-in MoE layer that keeps a bounded set of experts in RAM and reads the rest from SSD on demand. Owns the expert cache, the sliding read window, and the decode and prefill paths. |
| `omlx/patches/expert_streaming/conversion.py` | 1848 | Converts a resident model to streaming in place. Finds the MoE containers per model family, swaps their weights for streaming banks, and installs the cache, governor, and hooks. |
| `omlx/patches/expert_streaming/shard_bank.py` | 1359 | Reads expert slices out of the safetensors files on disk. Maps the files with `mmap` and serves one expert at a time, with `mlock` support for pinned rows. |
| `omlx/patches/expert_streaming/warmer.py` | 869 | Two speed mechanisms. Kernel readahead hints for the experts the predictor expects next, and `mlock` pinning for experts that stay hot. |
| `omlx/patches/expert_streaming/residency.py` | 753 | Estimates the RAM a streaming configuration needs before the model loads. Used for budget checks and reporting. |
| `omlx/patches/expert_streaming/slot_cache.py` | 587 | Slot bookkeeping shared by the caches. Maps expert ids to physical rows and keeps the acquire/commit/rollback protocol consistent when a read fails. |
| `omlx/patches/expert_streaming/__init__.py` | 572 | Package entry point. Shared helpers that the engines and admin routes call: backing discovery, summary logging, pin-profile persistence, and the tunable-key list. |
| `omlx/patches/expert_streaming/speculation.py` | 522 | Routing history and the transition table that predict which experts the next layers will need. |
| `omlx/patches/expert_streaming/governor.py` | 501 | Adjusts cache capacity at request boundaries from two signals: free system memory and measured decode stalls. |
| `omlx/patches/expert_streaming/adaptive_topk.py` | 433 | Optional knob. Truncates the router top-k to the smallest prefix that reaches a cumulative score-mass threshold. |
| `omlx/patches/expert_streaming/model_hooks.py` | 339 | Registry of what differs per model family: decoder classes, top-k hooks, kernel choice, and weight layouts. This is the seam that keeps the rest of the code model-agnostic. |
| `omlx/patches/expert_streaming/cache_policies.py` | 299 | The S3FIFO eviction policy, an alternative to the default LRU. |
| `omlx/patches/expert_streaming/qwen35_stream_eval.py` | 223 | Per-layer eval boundary for the installed qwen3.5 MoE decoder, which has no vendored equivalent. |
| `omlx/patches/deepseek_v4/spill.py` | 248 | Loads JANGQ DeepSeek-V4 checkpoints whose experts are stored unstacked, without materializing ~64 GiB at once. |
| `omlx/patches/deepseek_v41/streaming_backing.py` | 415 | Connects the shared governor to the DeepSeek-V4.1 native adapter. The policy stays model-agnostic; this file holds the V4.1 mechanics. |

### New test files

| File | Lines | What it covers |
|---|---:|---|
| `test_expert_streaming_robustness.py` | 581 | Failure contracts: backing rollback, atomicity, locks, memory floors, and reader cleanup. |
| `test_deepseek_v41_dspark_offload.py` | 552 | V4.1 DSpark verify under frozen expert residency. |
| `test_expert_streaming_layout_hooks.py` | 491 | MoE geometry resolution and the model-family hook contracts. |
| `test_expert_streaming_conversion_guard.py` | 422 | Conversion guard rails, cold-tier label validation, and prefill pin ordering. |
| `test_expert_streaming_dynamic.py` | 420 | Dynamic budget behavior and phase-aware capacity. |
| `test_generic_arena.py` | 357 | Arena mode: experts bound into fixed slot rows of a persistent bank. |
| `test_slot_arena.py` | 300 | The SlotArena protocol: acquire/commit/rollback, grow and compact, frozen ordering, staged prefetch. |
| `test_v41_span_reads.py` | 193 | V4.1 demand reads through the shared backing store on synthetic checkpoints. |
| `test_expert_streaming_regressions.py` | 156 | Regressions: SSD backing failure, per-model top-k isolation, spill manifest invalidation. |
| `test_speculation_state.py` | 143 | Speculation lifecycle: closed guards, staged-future cancellation, shared-future drop. |
| `test_mtp_streaming_gate.py` | 129 | The gate that decides when MTP speculation may run under streaming. |
| `test_deepseek_v41_span_staging.py` | 128 | V4.1 staged span reads on converted checkpoints. |
| `test_expert_streaming_s3fifo.py` | 125 | The S3FIFO policy: per-queue indexes, drain and resize accounting. |
| `test_expert_streaming_shard_bank.py` | 97 | Backing-store bounds and pin accounting. |
| `test_warmer_hooks.py` | 96 | Warmer and pin hook phase threading. |
| `test_moe_offload_unified.py` | 76 | Legacy offload keys routed to the streaming backend. |
| `test_dsv4_spill.py` | 52 | Spill-stacking units for unstacked V4 checkpoints. |

The remaining ~6k changed lines are wiring in existing files: engine load paths, model settings, admin routes, and the macOS settings screen.

errrepe:
## Update: DeepSeek V4.1 streaming pipeline optimizations + rebase onto current `main`

This update rebases the branch onto current `main` (through `24060057`, incl. the teardown/profile-validate/cache-alignment fixes) and adds a measured optimization pass over the V4.1 MoE expert-streaming path. Merge conflicts were resolved preserving both sides (upstream `cancelled` teardown flag, profile-apply refresh, `matchingProfileTemplate` helpers; streaming teardown + dashboard telemetry chip).

### What changed

- **Exact-demand prefill readahead** — `F_RDADVISE` issued per chunk for the exact experts the next chunk will route, plus whole-bank advise of the next layer. Prefill disk throughput ~2.1 GB/s vs 0.9 baseline.
- **Verify-fed staging predictor** — verify routing feeds the next-layer predictor (`staged_hits` went 0 → 6039 in measurement; the staging pipeline was structurally dead under DSpark verify).
- **Transient prefill capacity** (`OMLX_V41_PREFILL_CAP`) — raised slot cap during prefill leaves a warm working set at decode start.
- **Verify scratch rows** (`OMLX_V41_VERIFY_SCRATCH`) — small transient row region absorbs the verify union in a single ensure instead of serial 1-row ensures evicting decode-hot residents (evictions of residents 443 → 121).
- **DSpark draft-stage offload** (`OMLX_V41_DRAFT_OFFLOAD`) — draft stages stream instead of pinning ~6.4 GiB, which the governor converts into trunk capacity 11 → 17 rows/layer. Clamped to bank size so small draft banks stay resident.
- **Governor telemetry + tuning** — staged-skip reasons (no_pred / recall / headroom), `OMLX_V41_VERIFY_HUNGER`, headroom gate on verify advise.

### Measured results (DeepSeek-V4.1-Flash-oQ4e-mtp, 2k prompt, 48 decode tokens, `--mtp`, 48 GiB machine)

| arm | TTFT | decode | phys after load |
|---|---:|---:|---:|
| baseline (legacy) | 205 s | 0.091 tok/s | 28.0 GiB |
| readahead + predictor | 208 s | 0.161 tok/s | 28.1 GiB |
| + prefill cap | 199 s | 0.636 tok/s | 28.0 GiB |
| + draft offload | 188 s | 0.667 tok/s | 21.7 GiB |
| + verify scratch | 238 s | 0.743 tok/s | 28.0 GiB |
| **combined** | **201 s** | **0.745 tok/s (~8.2×)** | **21.75 GiB** |

### Promoted defaults

`PREFILL_CAP=64`, `VERIFY_SCRATCH=8`, `DRAFT_OFFLOAD=8` — each measured as a net win on the test machine; every knob keeps `=0` as an escape hatch restoring legacy behavior. `VERIFY_SCRATCH=32` was rejected: it OOMs Metal (+~25 GiB transient on this box).

### Removed: CED bounded replay

A bounded CED replay was implemented and measured, then removed: under `--mtp` the DSpark capture targets (layers 37–39) always lie beyond the kv_source boundary (20), so the safety gate disables replay on every primed prefill — it could never engage in the production configuration. Dead code was removed rather than kept opt-in.

### Correctness

- All 2k-prompt runs captured **byte-identical output text** across arms (v41 engine emits text only, so no token-ID gate exists for it).
- Short-prompt A/B showed occasional argmax tie-flips; a repeatability run confirmed the baseline does not reproduce itself either (`base ≠ base2`), while `base2 == new2 == new` — divergence tracks chunk-boundary changes (slot capacity → grouped-GEMM accumulation order), i.e. FP reordering, not a logic change.
- Tests: **302 passed** on this branch (`test_deepseek_v41*` 118 + shared arena/MTP/streaming suites 184, plus teardown coverage for the resolved conflicts).

errrepe:
## Refactor + hardening pass (27 commits, review-driven)

A multi-agent review of this PR's diff found latent bugs, duplicated machinery, and drift between the runtime and admin layers. This push lands the full fix set — all green: **14180 passed / 0 failed** (`pytest tests/ -m "not slow and not integration"`), plus `tests/moe_expert_offload_ui.test.cjs`.

### Bugs fixed

- `scheduler.py`: `float(os.environ.get(...))` at class scope could crash module import on malformed `OMLX_STREAMING_BANK_TOKEN_RATIO` — now `env_float`.
- `conversion.py`: bare `mx.clear_cache()` replaced with the synced `_sync_and_clear_cache` path (M4 `completeMemory()` panic class, #300/#888).
- S3-FIFO admission counted only `_store` — the probationary `_small` queue was invisible to global fullness; arena-mode `resident_bytes` similarly reported ~0. Both accounting paths now see all resident rows.
- Admin/runtime validation drift: cold tiers accept `"2".."8"`, `topk_threshold` floor 0.05 — single bounds table now drives both (see below).
- Duplicate readahead: `PageCacheWarmer` and the per-layer advisor both issued `F_RDADVISE` per token; one predictor remains, and `expert_streaming_readahead` now actually gates it.
- Alias path `moe_expert_offload_enabled`: a converter crash on a supported model type used to silently downgrade to the legacy fetch-on-miss adapter. It now logs + re-raises; fallback stays only for unsupported types / clean no-conversion, stamped via `streaming_fallback_reason`.
- Stale benchmark references to removed constants; `ExpertCache` residency rebuilt on shared `SlotArena` (third copy of the realloc/evict/rollback engine deleted, rollback semantics preserved).
- Spill manifests are now mode-safe: v3 records `{mtp_stages, block_part}`, so toggling MTP or DSpark between loads produces a miss → respill instead of a strict `load_weights` failure.

### Structural refactors

- `streaming_switch.py` (3.5k lines, ~8 responsibilities) → `expert_cache.py` + `bank_io.py` + `streaming_layers.py`; the old module remains a compat façade.
- Package `__init__.py` → thin façade over `settings.py` (budget/override/schema policy) + `lifecycle.py` (backing discovery, teardown, summaries).
- Shared primitives extracted: safetensors readers/dtype tables, staged reads, prev-token predictor, switch-sort helpers, `PerLayerResidencyAdapter`, streaming test fixtures.
- Single settings schema: `model_profiles.STREAMING_SETTING_SCHEMA` → authenticated admin endpoint → dashboard input bounds (literal fallbacks preserved offline).
- DeepSeek V4.1 off private `ExpertBackingStore` internals; DeepSeek V4 MTP expert stages now stack through the spill path (`OMLX_DSV4_SPILL=0` opts out).
- Governor memory probing now uses `omlx/utils/psutil_compat.py`.

### Notes for reviewers

- `ExpertCache`→`SlotArena`: two semantics intentionally differ (hit/miss interleave order within one call; in-flight payload bound for `needed<=cap`) — same trade-offs DSv4.1 already shipped; no test pinned the old behavior.
- Backing discovery split into model-side vs engine-side hop sets after the shared walk produced false positives on mock-based scheduler tests.
- Two review-driven fixes landed on top: restored package-root re-exports dropped by the split, and the spill-manifest mode check above.

PR #3676, opened 2026-09-15
feat(moe): size each residency option and fit the largest that admission accepts

beatakouchnir:
Follow-up to #2595 (expert offload primitive) and #3628 (V4.1 `fit_resident_fraction`).

## Problem

The residency field accepts 5–95 percent with no indication of what a value costs in memory or whether admission will accept it. Users map their machine to a fraction by hand, or find out at load time. A fit existed only for the DeepSeek V4.1 adapter, and only as a library call.

## Change

- `moe_expert_offload.py`: the header scan behind `estimate_offload_admission_bytes` is factored into a cached layout (keyed on the config and shards' stat signature) plus pure arithmetic; the estimate's contract is unchanged, and `mtp_resident` (from #3784/#3841) is part of the layout's cache key, so a resident draft head is never counted as offloadable. New `offload_capacity_fractions` lists every fraction that maps to a distinct whole-expert capacity, so a chosen value round-trips through the setting exactly, and `fit_resident_fraction(model_path, full_size, budget_bytes, *, mtp_resident=False)` bisects it for the largest fraction whose estimate fits (`1.0` when the resident model fits, `None` when even the floor does not, as for V4.1). `deepseek_v41`: `capacity_fractions`.
- `EnginePool.moe_offload_admission_bytes(entry, settings, fraction)` runs `_entry_runtime_resident_size` on a probe copy of the settings, and `fit_moe_offload_fraction(entry, settings, budget_bytes)` bisects the adapter's candidates (built with the settings' `mtp_enabled`) through it, so every adapter (common, V4.1 with either Engram mode, Qwen4-Exp with the PLE table) is sized by the arithmetic admission uses. `_entry_runtime_resident_size` gains an optional `ceiling` (default `None`, the previous behavior) that reaches the PLE and Engram status calls. The fit passes the budget as that ceiling: a resident table that stops fitting is forced to SSD exactly as admission would force it, and the larger fractions are held to its mmap size, which keeps the fitting fractions a prefix.
- `GET /api/models`: `moe_expert_offload_presets` (`fraction`, `bytes`, `fits` per option), `moe_expert_offload_fit_fraction` and `moe_expert_offload_fit_bytes` (`null`/`0` when the ceiling is unknown, the model is not in the pool, or nothing fits). A sizing error is logged at debug and leaves the fields empty.
- Dashboard: under the residency field (#3714) the fit is offered as a clickable hint, `Fit to memory: 41.4% resident · ~8.0 GB`, that fills the field; when even the routing floor exceeds the ceiling, a one-line notice replaces it. Two i18n keys (English in every locale file, like the existing offload keys). Docs gain a *Sizing a residency to your memory* section.

## Verified

Header-only on cached checkpoints (M5 Max 128 GB, nothing loaded). The 107 GiB budget is the documented Metal working-set limit on a 128 GB machine. Every call took at most 0.04 s.

| checkpoint | size | budget | fit | resident at fit |
|---|---:|---:|---:|---:|
| `mlx-community/gemma-4-26b-a4b-it-4bit` (128 experts) | 15.0 GiB | 8 GiB | 53/128 = 0.4141 | 8.0 GiB |
| same | | 6 GiB | 31/128 = 0.2422 | 5.9 GiB |
| `mlx-community/OLMoE-1B-7B-0125-Instruct-4bit` (64 experts) | 3.8 GiB | 2 GiB | 29/64 = 0.4531 | 2.0 GiB |
| `mlx-community/Qwen3.8-Flash-Next-4bit` (512 experts) | 109.1 GiB | 107 GiB | 1.0, PLE forced to SSD | 77.8 GiB |
| same | | 64 GiB | 416/512 = 0.8125 | 63.9 GiB |
| `Jundot/DeepSeek-V4.1-Flash-oQ3e-mtp` (384 experts) | 347.4 GiB | 107 GiB | 158/384 = 0.4115 | 106.6 GiB |
| same | | 64 GiB | 87/384 = 0.2266 | 63.6 GiB |

The V4.1 row reproduces the 41% the docs quote for `fit_resident_fraction` at 107.5 GiB. The Flash-Next row is the non-monotone case: at 75% the PLE table is resident (90.6 GiB), at 100% the resident estimate exceeds the ceiling, the table is forced to SSD and the model fits at 77.8 GiB. For gemma and OLMoE the pool's fit equals the adapter-only fit at every budget.

## Tests

- `test_moe_expert_offload.py::TestBudgetFit`: candidates are whole experts in some container (a stacked 32-expert and a per-expert 16-expert container in one checkpoint); at every candidate's estimate and one byte below it, the fit is the largest candidate whose estimate fits; dense checkpoint, kill switch and a missing path fit only at the full size; the layout cache follows a checkpoint edit; with `mtp_resident` the draft head's container is neither discounted nor a candidate, even after the cache was populated without the flag; the pool's admission and fit equal the adapter's at every candidate, and the probe never edits the caller's settings.
- `test_deepseek_v41_moe_offload.py`: `capacity_fractions`, and the pool's fit equals the adapter's SSD fit at every capacity's size and one byte below it, in both Engram modes.
- `test_moe_offload_compat.py`: the payload with a ceiling (sizes, `fits`, fit, the ceiling passed to every call), without one, and on a sizing error.
- `moe_expert_offload_ui.test.cjs`: the fit hint's label, apply and no-fit states and its markup. A separate commit gives the spec-toggle check the `selectedModel` its `:disabled` expression reads since #3784; without it the file fails on main with a `ReferenceError`.

327 passed across the offload, engine-pool and admin-template suites on the current base; the UI test passes; `ruff` is clean on the changed lines; new code is black-formatted, existing files were not reformatted.

## Not in this PR

An "auto" residency that resolves to the fit at load time. The pieces are here; it needs a settings-schema decision (sentinel fraction or separate flag) and the native app's DTO.


beatakouchnir:
A few changes since this was opened:

**UI now follows #3714.** The residency control is now the percentage field, so the per-option size labels and the "Fit to memory" dropdown entry in the original description are gone. The fit appears as a clickable hint under the field (`Fit to memory: 41.4% resident · ~8.0 GB`), and clicking it fills the field. When even the routing floor exceeds the ceiling, a one-line notice replaces it. That leaves two i18n keys, not three. `/api/models` still reports the 12.5/25/50/75 sizes in `moe_expert_offload_presets`, and the docs section describes the field.

**Resident MTP draft head (#3784, #3841).** The cached layout scan behind `estimate_offload_admission_bytes` now takes `mtp_resident` and includes it in the cache key, so a layout scanned without MTP is not reused for a model loaded with it. `offload_capacity_fractions`, `fit_resident_fraction` and the pool's candidate list pass the flag through (from `mtp_enabled` in the settings). With Lightning MTP on, the fit therefore does not count the draft head as offloadable. `test_resident_draft_head_is_not_offloadable` populates the cache without the flag first, so it fails if the flag is left out of the key. The V4.1 path already took it through `_plan`.

**Separate commit, fixes main:** `tests/moe_expert_offload_ui.test.cjs` fails on main with `ReferenceError: selectedModel is not defined`, because #3784's `:disabled` expression reads `selectedModel.moe_offload_allows_mtp` and the test's scope does not define it. 81b146f8 adds it.

PR #3682, opened 2026-09-15
feat(i18n): route the remaining hardcoded UI strings through the catalog (0.7.0.dev2)

LXD-8:
> **Based on `main` @ `24060057` (v0.7.0.dev2, 2026-09-15).**

The web admin and the macOS app still had user-visible copy baked into markup, Alpine expressions and script literals, so those screens could not follow the language setting. This routes them through the catalog — **30 files, +7363/−848**, two commits.

**This PR is the extraction step only.** Every locale renders the English source for the new keys, exactly as those screens did before, so no locale changes appearance. The wording for each language lands on top, in the companion PR that is built on this branch — which is also why the macOS catalog entries (`apps/omlx-mac/Resources/Localizable.xcstrings`) are not here: they are localization data, and they travel with the translations.

## 704 new keys

| Area | Files | Strings |
|---|---|---|
| Cluster v2 dashboard | `_cluster_v2.html`, `cluster_v2.js` | ~390 |
| Dashboard shell, settings, logs | `dashboard.js`, `_settings.html`, `_logs.html` | ~110 |
| Model-settings modal, Models | `_modal_model_settings.html`, `_models.html` | ~115 |
| Status, Bench, Accuracy, chat, base | `_status.html`, `_bench.html`, `_bench_accuracy.html`, `chat.html`, `base.html` | ~75 |
| macOS app | `ModelSettingsScreen`, `oMLXApp`, `WelcomeWindow`, `AboutScreen`, `CodeChip` | 15 |

The second commit covers the expose-as-API toggle's `ON` / `OFF` labels. The inline edit dialog had hardcoded that pair all along, and `#3658` (`1f7ab188`) added a second copy to the new-profile form while this branch was open. Both now share one key pair, so the two forms cannot drift. `API` itself stays a Latin literal in both toggles, and the new keys sit at the end of the `profiles` block rather than beside `expose_as_model`, whose neighbouring `expose_engine_fields_hint` the catalog audit retranslates — insertion there would put two hunks on adjacent lines of `zh.json` and make the two PRs conflict instead of applying in either order.

## Locale policy

`en.json` carries the source and `zh.json` carries Simplified Chinese; the other seven catalogs are regenerated by `scripts/normalize_i18n.py`. Here **all nine** receive the 704 new keys with the **English fallback**, so no locale regresses and Simplified Chinese never leaks into another catalog. Key parity holds at 1922 keys per catalog.

Acronyms and mechanism names stay in Latin throughout — MTP, PLE, Engram, GDN, MoE, DFlash, SpecPrefill, TurboQuant, imatrix, rank, `sidecar`, `API`.

## One behaviour-preserving refactor

`cluster_v2.js` used its display label as a lookup key: `requestPhaseLabel()` returned `'Failed' | 'Complete' | 'Prefill' | 'Decode' | 'Queued'` and `requestPhaseTone()` compared that same string to choose a colour. Translating the return value in place would have broken the tone lookup, so the function now returns stable ids (`failed`, `complete`, `prefill`, `decode`, `queued`), the tone switch uses those ids, and the display site translates: `window.t('cluster.v2.phase.' + requestPhaseLabel(request))`. All five id/tone pairs produce the same classes as before.

`cluster_v2.js` also keeps a `CLUSTER_V2_STRINGS` map of the English source, used only when `window.t` is unavailable (offline component tests). It is kept in sync with `en.json`; all 27 entries match it exactly.

## Test updates

- `test_admin_model_settings_template.py`, `test_admin_audio_upload_ui.py` and `test_admin_new_profile_expose_as_model.py` — asserted English copy that is now localised; they assert the keys instead.
- `tests/ui/test_cluster_v2_wizard.py` — its harness now feeds the component the real English catalog as `window.t`, so runtime assertions still read the shipped strings; source-level assertions check the keys, and the phase expectations follow the new lowercase ids. (Skipped unless `node` is on PATH, as before.)
- `test_profile_editor_behavior` — the Node harness `#3678` added constructs `dashboard.js` with `window: {}`, which the localised component now refuses (`window.t is not a function`). The harness gets the shipped English catalog as `window.t`, the same treatment as the wizard harness, so its seven cases still assert the strings the app renders.

## Verification

- `pytest -m "not slow" -k "admin or i18n or locale or template or ui or cluster or bench or settings"` — **3340 passed, 16 skipped, 0 failed**. (`test_eval_worker_pool.py::test_real_code_scoring_subprocess_cleanup` times out in this environment; it fails identically on unmodified `main`, so it is unrelated.)
- Every `t('…')` key referenced by a template or script resolves in `en.json`: 1817 static references, the only unresolved name being the documented placeholder inside the `cluster.v2.phase.<id>` comment, and all four dynamic prefixes (`cluster.v2.phase.`, `settings.resource.guard_tier.`, `settings.resource.guard_tier.reclaim_method.`, `usage.`) resolving to existing keys.
- Key parity across all nine catalogs — 1922 keys each.
- Rebased onto `24060057` after `#3678` landed; its only conflict was the removal of `test_apply_profile_surfaces_server_validation_error`, which this branch had been adapting — the deletion is taken as-is.
- `CLUSTER_V2_STRINGS` matches `en.json` exactly, all 27 entries.


LXD-8:
@jundot Thank you for the care you keep putting into oMLX. Building and maintaining this much on your own, at this level of polish, is a real contribution to local AI on Apple silicon — it is the reason I run it every day.

I would like to give something back, and the gap I can actually help with is language: a project only reaches as many people as it can be read by. I have put that work into two PRs, kept deliberately separate so each one can be reviewed on its own:

- **#3682** (this PR) — the i18n migration: the remaining hardcoded UI strings now go through the catalog, with every locale on the English fallback, so nothing changes visually for anyone.
- **#3684** — the Simplified Chinese for those strings, plus an audit of the existing catalog. It is stacked on #3682, so I left it as a draft until this one lands.

More languages means more users, and more users means the project keeps growing — so I hope this is genuinely useful.

Whenever you have a moment, I would be grateful if you could review them. If something is wrong, or simply does not match how you want oMLX to work, please tell me and I will fix it — I am happy to rework the wording, re-split the branches, or drop anything that does not fit.

jundot (approved):
Thank you for the thoughtful message and for using oMLX every day. Maintaining it on my own takes a fair amount of work, and I really appreciate you taking the time to contribute across both the web UI and the Mac app. Separating extraction from translation also made the changes easier to review.

The UI checks and macOS build passed, and keeping request-phase colors independent of translated labels is the right approach. I found a small TXT-export escaping issue that I'll handle in a follow-up. LGTM, and thanks for helping more people use oMLX in their own language.

Issue #2204, opened 2026-07-11
Qwen 3.6 serious prefill regression between 0.4.4 and 0.4.5dev1


zviratko:
M5 Max, macOS 27b3

Seems to happen on both MoE and dense
I only noticed it on 0.5.0 but it already happens on 0.4.5dev1
The speed regression is pretty brutal

### v0.3.12-macos-tahoe-26 - (v0.4.4 is actually about 20% faster even)
```
Benchmark Model: Qwopus3.6-35B-A3B-v1-MLX-6bit-int6-affine
Single Request Results

Test                TTFT(ms)    TPOT(ms)        pp TPS        tg TPS      E2E(s)    Throughput    Peak Mem 
pp1024/tg128           454.2       10.11  2254.8 tok/s    99.7 tok/s       1.738   662.8 tok/s    31.43 GB 
pp4096/tg128          1157.6       10.45  3538.3 tok/s    96.4 tok/s       2.485  1699.9 tok/s    32.24 GB 
pp8192/tg128          2122.0       10.19  3860.6 tok/s    98.9 tok/s       3.416  2435.7 tok/s    32.58 GB 
pp16384/tg128         4394.1       10.61  3728.6 tok/s    95.0 tok/s       5.742  2875.6 tok/s    33.21 GB 
pp32768/tg128        10258.6       12.18  3194.2 tok/s    82.7 tok/s      11.806  2786.5 tok/s    34.55 GB 
pp65536/tg128        27625.3       13.53  2372.3 tok/s    74.5 tok/s      29.344  2237.8 tok/s    37.23 GB 
pp131072/tg128       87367.7       17.36  1500.2 tok/s    58.0 tok/s      89.573  1464.7 tok/s    42.61 GB 
pp200000/tg128      203147.4       22.99   984.5 tok/s    43.8 tok/s     206.067   971.2 tok/s    48.27 GB

Continuous Batching 
pp1024 / tg128

Batch           tg TPS   Speedup        pp TPS    pp TPS/req    TTFT(ms)      E2E(s) 
1x          99.7 tok/s     1.00x  2254.8 tok/s  2254.8 tok/s       454.2       1.738 
2x         144.4 tok/s     1.45x   704.5 tok/s   352.3 tok/s      2799.5       4.679 
4x         344.2 tok/s     3.45x  1032.7 tok/s   258.2 tok/s      2004.8       5.454 
[...]

PR #3685, opened 2026-09-15
fix(attention): keep SDPA256 prefill on the bounded route

tc3oliver:
## Problem

The SDPA256 route chooses between two different floating-point reductions. Deciding per call from live guard headroom makes that choice a function of live process memory, which differs between otherwise identical processes — resident pages of the mmap'd weights, a previous request's pooled buffers. The same request then switches routes at a different `kv_len` and produces different temperature-0 output.

## Safety audit

The concern raised on this PR is valid, and tracing it changed the fix.

`memory_monitor` prices a registered head dim as O(L) for every shape the route covers, but since #2204 the route can still select the unfused path based on live headroom. The existing estimator assumption and the live-memory route can therefore diverge, and a chunk admitted on the bounded transient can go on to materialize the full `[n_q, query_len, kv_len]` fp32 matrix.

The static-budget approach this PR previously used does not address that: its budget prices only the requesting model and the current request, so other loaded models and other in-flight work are invisible to it. That approach has been dropped.

## Fix

Decide from the call shape alone. Qualifying head_dim-256 prefill takes the bounded route in every process, which restores #2025's semantics and makes the estimator's assumption true by construction.

`OMLX_SDPA256_TILED=1/0` still forces or disables the route, and a disabled route still withdraws the O(L) registration. The scheduler's headroom provider, its helpers and the propagation marker that existed only to feed them are removed, so the route now takes no memory input at all.

## Why this design

A route-aware live estimator was tried first. It restores the accounting, but prefill admission's baseline is live residency and its throttle also reads live usage, so charging the real unfused transient pulls the request into the throttle and chunk partitioning becomes process-history dependent again — five histories produced five partitions. There is no request-level reservation to lean on instead. Unconditional bounded routing is the smallest change found that keeps both deterministic routing and #2025's safety semantics without redesigning scheduler admission.

## Validation

Apple M4 Max, 64 GB, macOS 15.6. head_dim-256 GQA model (24 query heads : 4 KV heads, 16 full-attention layers, 4-bit weights), 68,034-token prompt, 32 output tokens, temperature 0 / top_k 1, `max_concurrent_requests: 1`, prefix cache disabled, memory guard `balanced`, `prefill_min_chunk_tokens: 256`.

Reproducibility, four process histories (two cold starts, one small preceding request, one 68K preceding request):

- one route (bounded for every qualifying call), one chunk partition, one first-token logits hash, one output hash

Memory behaviour:

- second model resident: route and per-chunk charge unchanged; the guard's live baseline still sees the sibling
- concurrent requests: characterized below
- long-context case: at the context where #2025 hit its wall the unfused matrix is tens of GiB and is no longer reachable

Tests: full suite green; the new regression module fails on the parent commit and passes here.

## Performance

Controlled A/B on the same machine, model, prompt and settings, three fresh processes per group:

| | unfused (main) | bounded |
|---|---|---|
| TTFT | 358.7 s | 356.1 s |
| prefill | 201.8 tok/s | 196.6 tok/s (−2.6%) |
| peak MLX memory | 35.57 GiB | 24.70 GiB (−30.6%) |
| decode | 19.1 tok/s | 19.2 tok/s |

TTFT shows no material regression — the difference is within run-to-run noise on this machine. The measured cost is prefill throughput, −2.6%.

## Known limitation

Prefill admission has no reservation for approved-but-not-materialized allocations: two chunks admitted against the same memory snapshot both pass. Keeping the route bounded shrinks the per-request attention transient by more than an order of magnitude, which reduces the exposure, but this PR does not claim to fix that scheduler-wide race. It is characterized in the tests rather than changed.


jundot:
Making routing reproducible is useful, but I’d like to preserve the existing OOM protection as well. The current live-memory check was added in #2204 to keep the fast unfused path when it fits, and use the bounded path otherwise.

This PR’s static budget leaves out other loaded models and active requests. Since the prefill guard assumes the smaller bounded allocation, it could approve a chunk that then takes the larger unfused path, potentially bringing back the OOM issue from #2025.

Could you check whether the remaining guards safely cover this case? If this concern is valid, I’d appreciate an update that preserves that protection while improving reproducibility.

tc3oliver:
Thanks — the concern is valid, and checking it changed the fix.

I traced the remaining admission and memory guards. The prefill estimator still prices a qualifying head_dim-256 chunk as O(L) for every shape the route covers, while since #2204 the route can select the unfused path from live headroom. So the estimator assumption and the route that actually runs can diverge, and a chunk admitted on the bounded transient can materialize the full score matrix. The static budget I had used does not close that — it prices only the requesting model and the current request, so other loaded models and other in-flight work are invisible to it. I've dropped that design.

I tried a route-aware estimator first. It restores the accounting, but admission's baseline is live residency and its throttle also reads live usage, so charging the real unfused transient brought process-history-dependent partitioning back — five histories, five partitions. With no request-level reservation available, I went with the smaller option: keep qualifying SDPA256 prefill on the bounded route, which matches both the original #2025 safety semantics and the estimator's assumption.

On the same 68k reproduction (M4 Max 64GB, 24:4 GQA head_dim 256, temperature 0, three fresh processes per group): prefill throughput 201.8 → 196.6 tok/s (−2.6%), peak MLX memory 35.57 → 24.70 GiB (−30.6%), decode unchanged, and TTFT within run-to-run noise (358.7 s vs 356.1 s) — no material regression, and I'm not claiming it is faster. A separate four-history qualification on the final tree gave one route, one chunk partition, one logits hash and one output hash. I also checked the second-model-resident and concurrent-request cases.

One thing I did not fix: admission has no reservation for approved-but-not-yet-materialized allocations, so two chunks admitted against the same snapshot both pass. Keeping the route bounded reduces the exposure by more than an order of magnitude but does not remove that race; it looks like it belongs to the admission design rather than here, so I've left it characterized in the tests rather than changed.

The PR and tests are updated accordingly and rebased onto current main.

jundot (approved):
Keeping qualifying prefill on the bounded route makes execution match the memory estimate. I did not reproduce the earlier large causal-prefill regression on M5.

The tested image workload showed about a 2% TTFT increase with array tiling. I consider that an acceptable tradeoff for predictable memory use.

I'll preserve the FP32 fallback when resolving the merge conflict and clean up the stale routing references afterward.

PR #3684, opened 2026-09-15
feat(i18n): zh-Hans for the migrated strings and a catalog audit (0.7.0.dev2)

LXD-8:
> **Based on `main` @ `7e6893e8` (v0.7.0.dev2, 2026-09-16).** The migration this work originally stacked on is now merged (#3682), so this is a **standalone PR: 4 files, +1288/−999** — nothing else needs to land first.

Supplies the Simplified Chinese for the strings the migration moved into the catalog, audits the wording that shipped before it, and unifies the places where the web catalog and the macOS catalog had drifted apart.

## Simplified Chinese for the migrated strings

`omlx/admin/i18n/zh.json` gains the zh-Hans values for the keys the migration added: **671** of the 706 get their own wording, and **35 stay identical to English** on purpose, because they are product names, acronyms and units (`oMLX`, `Codex`, `OpenCode`, `oQ` / `oQe`, `GitHub`, `HuggingFace`, `DDGS`, `SSD`, `ANE`, bit-widths). The expose-as-API toggle's labels are 开 / 关, matching `cluster.v2.common.on/off` and `acc_bench.config.thinking_on/off`.

`apps/omlx-mac/Resources/Localizable.xcstrings` gains the 13 entries for the migrated Swift strings, each with its `en` source and `zh-Hans` value, so the app follows the same wording as the web admin.

Conventions, matching the rest of this catalog: technical terms and mechanism names stay Latin throughout (MTP, PLE, Engram, GDN, MoE, DFlash, SpecPrefill, TurboQuant, imatrix, KV, SSD, ANE, MLX, sidecar, canary, rank, `min_p` / `top_p` / `top_k`); `Token` is the model token while `令牌` stays reserved for auth tokens; punctuation is full-width inside Chinese sentences and half-width inside numbers, version strings, paths and model ids; and one rendering per meaning across every screen.

## The catalog audit

260 values in `zh.json` (plus 47 in the macOS catalog, and the app's 13 new entries) correct the Simplified Chinese that shipped before the migration.

The audit is scoped against the catalog's own accepted conventions rather than an outside style guide: punctuation, spacing, casing and terminology were measured across the values already merged, and then every value — pre-existing and newly added — was checked against that profile. Where the catalog contradicted itself the majority usage won; where a difference was semantic rather than stylistic it was kept.

[...]

PR #3692, opened 2026-09-16
feat(i18n): add the macOS catalog entries for the migrated app strings

LXD-8:
The macOS String Catalog is missing the entries for the thirteen keys the migration introduced, so those strings have no English source to translate from.

#3682 routed five SwiftUI surfaces through the catalog with `String(localized: "common.copy.help", defaultValue: "Click to copy", comment: …)`, but never added those keys to `apps/omlx-mac/Resources/Localizable.xcstrings`. The `defaultValue` keeps the English UI correct — this is not a visual regression — but a key that is absent from the catalog cannot carry a translation, so the Simplified Chinese for these thirteen strings has nowhere to live.

**1 file, +156, insertions only**, on top of `main` @ `e27e263` (v0.7.0.dev2, 2026-09-16): thirteen entries, each with its key, `en` value and developer comment, and nothing else.

The thirteen are `app.command.close_window`, `common.app_name`, `common.copy.help`, and the ten `settings.experimental.qwen_ane.tuner.*` controls for the tuned-split dialog.

## Why the entries are their own PR

Catalog entries are i18n support; translated values are not. The companion PR supplies the `zh-Hans` for these entries and for the rest of the migrated strings, and it can only edit keys that exist in its base — so the keys have to land first. Adding the entries and their Chinese values together would put a key addition inside a translation PR and make the two impossible to rebase independently.

## Verification

- The diff is a pure insertion: 13 keys added, 0 removed, 0 existing entry altered, no `en` value or `comment` touched beyond the new keys.
- `LocalizationSmokeTests.swift`'s invariants still hold: valid JSON, `sourceLanguage` is still `en`, 1241 strings (was 1228).
- `pytest tests/test_admin_model_settings_template.py tests/test_admin_enhanced_readability.py tests/test_admin_new_profile_expose_as_model.py tests/test_admin_audio_upload_ui.py tests/ui/test_cluster_v2_wizard.py` — **122 passed**.
- No `omlx/admin/i18n/*.json` is touched, so the web admin is unaffected.


LXD-8:
Folded into #3684 instead: the thirteen catalog entries carry the app's `zh-Hans` values with them, so the app half of the localization ships as one change rather than two. #3684 is rebased onto `main` and no longer stacked on this branch.

PR #3698, opened 2026-09-16
fix(vlm): load vision-less VLM checkpoints instead of inventing a vision tower

LXD-8:
**Base:** `main` @ `bd841fc3` (v0.7.0.dev2)

## Why

A text-only oQ quant of `diffusion_gemma` — oQ's `text_only`, the admin UI's "Text Only" ("Excludes vision encoder weights. Output is a text-only model (~2-3% smaller).") — **cannot be loaded by oMLX at all**, with an error that reads like a corrupt download:

```
Model 'diffusiongemma-26B-A4B-it-oQ5-fp16' failed to load:
VLM load failed: Missing 210 parameters:
model.encoder.vision_tower.encoder.layers.0.input_layernorm.weight,
...
model.encoder.vision_tower.patch_embedder.position_embedding_table.;
LLM fallback also failed: Model type diffusion_gemma not supported.
```

The checkpoint is exactly what `text_only` promises, not damage: `_is_vision_tensor()` matches `"vision_"`, so `vision_tower.*` is stripped and `vision_config` / `image_token_id` are popped from the output config, while `embed_vision.*` is deliberately kept (see the `_is_audio_tensor` docstring, which cites `embed_vision.embedding_projection`). Every published artifact in that family shows it — `avneetsb/diffusiongemma-26B-A4B-it-oQ{2,4,5,6,8}-fp16`: **1292** tensors, **zero** `model.encoder.vision_tower.*`, three orphan `embed_vision.*`. (Google's base checkpoint has 1047 tensors including all 355 vision-tower ones.)

So the failure is on the load side, in two halves:

1. `model_discovery` routes `diffusion_gemma` to the VLM engine unconditionally — correct, since mlx-lm has no `diffusion_gemma` class and the block-diffusion decode lane only exists here. There is no text path to fall back to (`LLM fallback also failed: Model type diffusion_gemma not supported`).
2. `mlx_vlm.load()` then does `config.setdefault("vision_config", {})`, and **two sites turn that empty dict into a default vision tower**: `ModelConfig.from_dict` tests `if params.get("vision_config") is not None` (`models/diffusion_gemma/config.py:83`), and `update_module_configs` (`mlx_vlm/utils.py:736`) re-deserializes it afterwards. `EncoderModel.__init__` builds the tower and strict `load_weights` reports **210** missing parameters.

**210 is the tell**: 16 layers x 13 + 2 `patch_embedder` = the defaults of `gemma4.VisionConfig`. Google's real `vision_config` is 27 layers -> 353. The error names parameters that no `diffusion_gemma` checkpoint ever had.

## What

`_strip_vision_config_if_orphaned`, a vision twin of the existing `_strip_audio_config_if_orphaned` (which exists for exactly this class of failure: orphaned `audio_config` vs missing `audio_tower`). It fires only when **both** hold — `config.json` declares no vision sub-config, and no shard carries `vision_tower` weights — and then:

- nulls `vision_config` on the finished `ModelConfig`, and
- drops the checkpoint's orphan `vision_tower.*` / `embed_vision.*` tensors during `load_weights`.

Two details worth flagging, both verified by testing:

- **Nulling the value in the config dict does not work.** `vision_config = None` there crashes mlx-vlm at `mlx_vlm/utils.py:616` (`config.get("vision_config", {}).get("skip_vision", False)` -> `AttributeError`), and merely patching `ModelConfig.from_dict` is not enough either because `update_module_configs` re-creates the tower afterwards. Nulling after `update_module_configs` returns is the only position that sticks; the raw config is left as `{}` so line 616 stays happy.
- **The orphan `embed_vision.*` tensors have to go separately.** mlx-vlm's `DiffusionGemma.sanitize()` already covers that (`if not has_vision_tower: continue`), but `load_model` only runs `sanitize_weights` for **non**-MLX-format checkpoints (`if not is_mlx_format:`, `mlx_vlm/utils.py:553`) and oQ output is MLX-format. Same gap `_remap_nested_visual_on_load` documents for its key family.

The filter only drops keys the model has no slot for, so a model that owns its vision params is never touched.

## Verification

- `tests/test_vlm_vision_fallback.py` (new, 21 tests): `_has_vision_tower_weights` / `_is_vision_tensor_key` semantics; no-op when the config declares vision; no-op when the shards carry the tower; nulling under a simulated `update_module_configs`; orphan-key filtering keeps owned keys; warning logged once; both attributes restored on normal exit and on exception; no-op without `config.json`.
- `tests/test_vlm_vision_fallback.py tests/test_vlm_audio_fallback.py tests/test_vlm_engine.py tests/test_model_discovery.py` -> **351 passed** (needs `pytest-asyncio`; `pytest.ini` sets `asyncio_mode = auto`; without it the async tests error out on collection).
- End-to-end on the real 18 GB artifact `avneetsb/diffusiongemma-26B-A4B-it-oQ5-fp16`, with the guard from this branch:

```
vision_tower weights missing for diffusiongemma-26B-A4B-it-oQ5-fp16; loading without vision support
LOAD OK in 1.8s
  config.vision_config = None
  encoder.vision_tower = NoneType
  diffusion family (oMLX's own detector): block
  image path raises: This checkpoint does not include a vision tower; image inputs are not supported.
GENERATE OK in 0.8s
  -> A block diffusion language model is a generative model that creates text by
     iteratively refining coarse, noisy blocks of information into a coherent
     sequence, rather than predicting tokens one by one.
```

## Behaviour delta

- Vision-less VLM checkpoints whose config is silent about vision now load as text-only models instead of failing with `Missing 210 parameters`. Image input then fails loudly with mlx-vlm's own message rather than silently.
- Everything else is untouched: the guard is a no-op whenever the config declares vision or the shards carry a tower, and `_is_vision_tensor_key` only matches on whole path components.
- Non-goals: no routing change, no change to `text_only` in oQ, no new i18n strings. Users on a complete quant (`mlx-community/diffusiongemma-26B-A4B-it-{4,5,6,8}bit`) see no difference.

## Notes

The new `_has_vision_tower_weights` mirrors `_has_audio_weights`, including `for k in f.keys()`, so ruff's pre-existing `SIM118` count in `omlx/engine/vlm.py` goes 1 -> 2. `f.keys()` is required rather than stylistic: `safetensors.safe_open` has no `__iter__` (`TypeError: 'builtins.safe_open' object is not iterable`), so SIM118 is a false positive at both sites.


jundot (approved):
I reproduced the missing-210-parameters failure with a small quantized DiffusionGemma checkpoint. This change restores strict loading with identical logits and preserves explicit rejection of image inputs. I'll adjust unreadable-shard detection in a small follow-up so it leaves the configuration untouched. LGTM.

PR #3700, opened 2026-09-16
feat: one-click optimal settings and custom recipes from omlx.ai benchmarks

jundot:
## Summary

Tuning a model used to mean reading benchmark pages and copying dozens of settings by hand. This PR adds three buttons to the Model Settings header, in both the web dashboard and the macOS app: **Apply optimal settings**, **Apply custom recipe** and **Reset defaults**.

- **Apply optimal settings** looks up omlx.ai for benchmarks of the same model on the same chip (same or less memory, 4k prompt) and shows the three fastest results by prompt processing and the three fastest by token generation, each with a link and its PP/TG numbers. Pick one and its settings are applied and saved.
- **Apply custom recipe** takes a recipe string copied from any benchmark detail page on omlx.ai (new "Recipe" row next to the URL) and applies those settings to the current model. Handy for borrowing a setup from a similar model or another machine.
- **Reset defaults** returns every setting of the model to its default in one step, after a confirmation.

In all cases the result screen shows exactly which settings were applied. Anything this machine cannot run, such as a draft model that is not installed or an MTP checkpoint without MTP weights, is left out and listed under "Skipped on this machine" instead of failing the whole apply.

## How it works

- Three admin endpoints do the work so both clients stay thin: `settings/optimal` (GET lists candidates, POST applies one), `settings/recipe` and `settings/reset`. All of them persist through the existing settings PUT path, so reload behavior and validation are unchanged.
- Applying a snapshot replaces the shareable profile fields; fields the snapshot does not mention go back to their defaults so the model ends up in the same state as the benchmark. Grammar settings are never imported, and reset also clears pinned/default flags, aliases and the display name.
- A recipe is `omlx-recipe:1:` followed by the settings JSON compressed and base64url encoded, so it fits on one line and pastes anywhere.
- Benchmark uploads now include a few more performance settings (oQ A8, MoE expert offload, DFlash cache budgets) so future results carry the full picture. omlx.ai already accepts the larger snapshot and serves the new best-results API and the Recipe row.

## Validation

- Real server on an isolated port with `Qwen3.8-27B-4bit` on an M3 Ultra 512 GB: the optimal lookup returned 3 PP and 3 TG candidates and applying `ac43uto2` reproduced its settings with `vlm_mtp` skipped (draft not installed); a recipe copied from the production page `p5x3r8d2` applied 40 fields with `mtp` skipped because the local checkpoint has no MTP weights; reset cleared seeded alias, favorite and sampling values and the default-model pointer.
- Unit tests cover the recipe codec, the replace semantics, draft-model resolution, feature skipping, grammar exclusion and the candidate grouping; the macOS app tests cover the new DTOs and the result formatting.


jundot:
Superseded by #3701, which carries the same change as a single commit on top of main.

Issue #3702, opened 2026-09-16
MTP model hangs the GPU and allocates unboundedly after #3695 — no tokens emitted (Qwen3.6-35B-A3B-oQ6-mtp)


shironexus-gif:
**Describe the bug**

After updating to a main that includes #3695 (`feat(mtp): support multi-request Lightning MTP`, merged as 4d4750ee), requests against an `-mtp` model stop producing tokens. Generation hangs, the GPU is reset by the watchdog, and process memory grows without bound until the machine is close to OOM.

The GPU fault originates in the MTP reconcile path — note this one says **`Caused`**, so it is the culprit rather than a victim:

```
omlx.patches.mlx_lm_mtp.batch_generator - WARNING - MTP reconcile failed, falling back to plain drop:
  [METAL] Command buffer execution failed: Caused GPU Hang Error
  (00000003:kIOGPUCommandBufferCallbackErrorHang).
```

Every other command buffer in flight is then discarded as collateral of the GPU recovery:

```
omlx.scheduler - ERROR - Chunked prefill failed for a3f5ab93-...:
  [METAL] Command buffer execution failed: Discarded (victim of GPU error/recovery)
  (00000005:kIOGPUCommandBufferCallbackErrorInnocentVictim).

omlx.scheduler - ERROR - Chunked prefill failed for eb377ff7-...:
  ANE input pack command buffer failed; canceling this dispatch:
  Error Domain=MTLCommandBufferErrorDomain Code=1 "Discarded (victim of GPU error/recovery)"
  UserInfo={NSUnderlyingError=... Domain=IOGPUCommandQueueErrorDomain Code=5}
```

after which the MTP state cannot be recovered and subsequent streams fail:

```
omlx.server - ERROR - Error during chat streaming: Lightning MTP could not restore batch calibration state
omlx.server - ERROR - POST /v1/chat/completions → 500 (unhandled)
```

**To Reproduce**

1. Serve `Qwen3.6-35B-A3B-oQ6-mtp` on a main that includes 4d4750ee (#3695).
2. Send a single streaming chat completion:
   ```bash
   curl -sN http://127.0.0.1:8001/v1/chat/completions \
     -H 'Content-Type: application/json' \
     -d '{"model":"Qwen3.6-35B-A3B-oQ6-mtp","stream":true,"max_tokens":120,
          "messages":[{"role":"user","content":"Count from 1 to 40, one number per line."}]}'
   ```
[...]

PR #3703, opened 2026-09-16
fix(qwen4): give the vendored linear cache its batch conversion

LXD-8:
## Why

Lightning MTP's singleton rebuild converts a model's caches through
`mlx_lm.generate._make_cache`. That converter understands only mlx-lm's own cache
classes and raises for everything else, so a model whose `make_cache()` returns
mlx-vlm classes cannot be rebuilt:

```
ValueError: <class 'mlx_vlm.models.qwen4_exp.cache.ArraysCache'> does not yet support batching
```

Qwen4-Exp is exactly that shape: one vendored `ArraysCache` per linear layer, one
`QSAKVCache` per full-attention layer — and the QSA class already carries its own
`to_batch`, so the linear one was the only gap. `_rebuild_singleton_cache` returned
`None`, `_reconcile_mtp_batch_to_standard` returned `False`, and the batch path
raised:

- `RuntimeError: Lightning MTP could not restore batch calibration state` once the
  batch policy parked MTP (its `remaining` cooldown wants standard-decode samples), and
- `RuntimeError: Lightning MTP could not restore the committed batch cache` when a
  second request joined an MTP singleton.

In practice a Qwen4-Exp VLM at concurrency ≥ 2 kills the engine loop with the first
one: the benchmark run is lost and the model has to be reloaded.

## What

`ArraysCache.to_batch()` on the vendored class, so the conversion travels with the
cache the way `QSAKVCache` already does:

- a fresh cache takes the `left_padding` stamp — an `ArraysCache` is its own batch
  form (singleton and batched rows share the class), which is precisely the
  conversion mlx-vlm's own cache maker applies to it;
- a cache that already carries running state is returned untouched. It is one
  unpadded row, the convention `_to_batched_cache_layer` already documents for
  regular singleton caches. This matters: stamping a warm singleton in the join path
  changed MTP output for qwen4 (`test_multi_request_mtp_or_singleton_only_matches_standard[4-True-qwen4-*]`
  and `test_batched_head_survives_join_and_staggered_finish[qwen4-True-4]` failed with
  token mismatches until the warm case was left alone).

## Verification

Against `main` @ `65c65e38`, which includes #3702 — that PR reworks the rebuild's
prefill and the join branch, but the rebuild still goes through mlx-lm's converter,
so the failure reproduces unchanged. Without the fix, on that base:

```
$ pytest tests/test_singleton_cache_passthrough.py -q
2 failed
  ValueError: <class 'mlx_vlm.models.qwen4_exp.cache.ArraysCache'> does not yet support batching
```

With the fix, same base:

```
$ pytest tests/test_singleton_cache_passthrough.py -q
9 passed

$ pytest tests/test_mlx_lm_mtp_patch.py tests/test_mlx_vlm_qwen4_exp_compat.py -q
376 passed

$ pytest tests/ -k "qwen4 or singleton or cache" -q
2105 passed, 28 skipped

$ pytest tests/ -q
13825 passed, 2 failed, 340 skipped
```

The two failures are `tests/test_gdn_sidecar_quantization.py::test_encoded_sidecar_bytes_are_identical_across_processes`
and `tests/test_integrations.py::TestClaudeCodeIntegration::test_launch_omits_max_context_tokens_when_context_window_unset`;
both fail identically on a clean worktree of the same base without this change.

## Behaviour delta

Qwen4-Exp batch MTP can now rebuild its singleton cache instead of failing, so
parking the batch policy and joining a running MTP singleton no longer take the
engine down. Singleton decode, the join path for running caches and every other
model's conversion are unchanged.

A follow-up worth noting, not part of this diff: upstream mlx-vlm now exposes its
own batch-cache conversion publicly (Blaizzy/mlx-vlm#2279), so once the pinned
mlx-vlm version moves past that, this vendored hook can be dropped in favour of it.


jundot (approved):
The missing conversion hook reproduces on current main. With this change, Qwen4 cache reconciliation succeeds and preserves the next token, while the existing late-join parity checks still pass. I couldn't reproduce an unconditional concurrency failure on current main, but the rebuild path does need this fix. LGTM.

PR #3710, opened 2026-09-17
Fix/glm5 promoted boundary (depends on PR #3290)

luken:
## Summary

Preserve the complete pooled history when the scheduler promotes a compacted final boundary into the cache store source. This covers both normal completion and parser-side stops.

`Scheduler._get_boundary_store_override()` removes the promoted boundary from the intermediate snapshot map. The per-member store path currently reads `pooling_delta_ranges` only from that map, so the promoted final delta is stored as a complete `PoolingCache`. Restore then retains the full KV sequence but replaces the accumulated pooled history with the final block's rows.

The fix reads delta metadata from the same layer that supplies the boundary state.

## Dependency and review scope

**Depends on #3290; merge that PR first.** This branch is based on its current head, `82e66bb21c0c59403cbfb25033f66fe943a6ee61`, so the comparison against `main` currently includes its prerequisite commits. After #3290 merges, this branch can be rebased to leave only the follow-up changes.

The two additional commits are:

1. [ecbddbea — regression tests](https://github.com/luken/omlx/commit/ecbddbea): exercise actual scheduler promotion and both completion/parser-stop KV refill paths, with one-block controls and three-block pooled-history checks.
2. [70896a48 — fix](https://github.com/luken/omlx/commit/70896a48adc471fe95e428a4f5a1a63046518225): preserve delta ranges from the promoted store source.

[Review only the follow-up changes](https://github.com/luken/omlx/compare/82e66bb21c0c59403cbfb25033f66fe943a6ee61...70896a48adc471fe95e428a4f5a1a63046518225).

Opened at the [maintainer's request](https://github.com/jundot/omlx/pull/3290#issuecomment-5712027029), following the original submission at williamxie1989/omlx#1.

## Validation

Recorded validation on the exact #3290 backport plus these changes onto oMLX v0.6.4, on Apple Silicon (M3 Ultra, Python 3.11.16, MLX 0.32.0):

- Test patch without the fix: **2 failed, 2 passed**. Both three-block completion/parser-stop cases restore pooled shape `(1, 1, 8)` instead of `(1, 3, 8)`; one-block controls pass.
- With the fix: **all four regression cases pass**.
- **877 selected cache, scheduler, and engine tests passed** (not the full repository suite).
- Full/partial tensor restores through 94,208 tokens passed before and after reopening SSD storage.
- A byte-identical 46,080-token request returned the expected `COMPLETE` both fresh and cached; the uncorrected backport failed the cached replay.
- Concurrent 96K + 46K + 46K requests and their cached continuations passed the marker/tool checks.

The two changed files on this contribution branch are byte-identical to the tested backport. Their identity and `git diff --check` were verified when preparing this PR; Apple Silicon tests were not rerun on the Linux submission host.

This fix prevents new malformed blocks; it does not repair affected blocks already stored by earlier #3290 builds. Use a fresh prefix cache when validating those builds.

### DIFF
--- a/omlx/cache/paged_ssd_cache.py
+++ b/omlx/cache/paged_ssd_cache.py
@@ -385,13 +385,14 @@ _POOLING_SUB_CLASSES = frozenset({"PoolingCache", "BatchPoolingCache"})
 # addition to ArraysCache (see ``cachelist_pm_member_plan``). Kept separate
 # from ``_ARRAYS_SUB_CLASSES`` so signature descriptors
 # (``_block_cachelist_subtypes`` / ``cachelist_subtypes_from_cache_list``)
-# still route PoolingCache through the pooling branch (``PoolingCache:N``)
-# instead of mis-stamping it as an ArraysCache slot count. Must stay in
-# sync with ``prefix_cache._PM_SAFE_NON_SLICEABLE_SUBS``: BatchPoolingCache
-# is NOT boundary-eligible — compact_pooling_cache_snapshot only compacts
-# PoolingCache, so BatchPoolingCache boundary snapshots stay cumulative and
-# its layers keep the legacy path.
-_PM_BOUNDARY_SUB_CLASSES = _ARRAYS_SUB_CLASSES | frozenset({"PoolingCache"})
+# still route pooling members through the pooling branch
+# (``PoolingCache:N`` / ``BatchPoolingCache:N``) instead of mis-stamping
+# them as an ArraysCache slot count. Must stay in sync with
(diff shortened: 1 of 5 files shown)


jundot (approved):
@luken Reading delta ranges from the same source as the boundary state fixes the promoted-boundary history loss. I reproduced the three-block failure in both completion and parser-stop paths and verified complete pooled history after this fix, including SSD reopen and matching decode logits on a small GLM-5.3 checkpoint. LGTM; I'll land this immediately after #3290.

PR #3712, opened 2026-09-17
Add interactive controls for long-running reasoning

Maxim-Mazurok:
## Summary

This proposes interactive generation controls for long-running reasoning models, together with an initial implementation:

- **Force Answer** asks an active reasoning request to close its thinking phase and begin the visible answer without changing the model's configured reasoning budget.
- **Pause** saves an exact token checkpoint before cancelling transport, so a long generation can be stopped safely when the user needs to leave.
- **Resume** continues from that checkpoint rather than restarting the prompt and discarding already-computed reasoning.
- **Think More** discards the visible answer, retains the reasoning-token prefix, and continues reasoning with an additional configurable budget.
- Live and completed generation views expose recent throughput and speculative-decoding telemetry.

## Motivation

Large reasoning models can spend tens of minutes thinking. Sometimes that extra time is available; sometimes an answer is needed immediately. A fixed reasoning-budget setting cannot cover both situations well.

The intended workflow is interactive:

1. Start a request with the model's normal, generous reasoning budget.
2. Use **Force Answer** when time becomes more important than additional reasoning.
3. Use **Pause** when the session must be interrupted, optionally unload the model, then **Resume** later from the accepted token prefix.
4. Use **Think More** when an answer was forced too early but more time becomes available later.

This is presented as both a feature request and a working implementation for maintainer feedback.

## Implementation Notes

- Adds request-scoped force-output and pause endpoints.
- Propagates authoritative request IDs from SSE requests through engine layers.
- Persists versioned pause checkpoints in browser IndexedDB without storing API keys.
- Carries raw continuation token IDs and reasoning-boundary metadata through text, VLM, scheduler, DFlash, and MTP paths.
- Preserves total generation and reasoning caps across resume operations.
- Adds localized controls/status text for every existing admin locale.
- Adds rolling generation-speed and speculative-efficiency metrics.

## Validation

- Rebasing onto current `origin/main` required semantic conflict resolution across scheduler, DFlash, VLM, MTP, server, and UI code.
- Full automated suite: **13,968 passed, 342 skipped, 85 deselected**.
- Focused generation-control suite: **1,205 passed**.
- No end-to-end manual model session has been completed for every control/backend combination.

## Known Limitations

- DFlash primary decode currently reports Force Answer as unsupported because that backend does not expose mid-stream reasoning-close token injection. Pause/Resume remains implemented there; Force Answer delegates when DFlash is using its fallback engine.
- Browser UI behavior is covered mainly by source-contract tests rather than full browser-driven model sessions.

## AI-Assistance Disclaimer

This implementation was generated largely with AI assistance. The exact model used for the original work is not verified; it was likely **GPT-5.6 Sol Medium in VS Code**. I have not personally functionally tested every path or backend. Please treat this draft as a feature proposal plus implementation for review and validation, not as a claim that every interaction is production-proven.


Maxim-Mazurok:
https://github.com/user-attachments/assets/6916c477-be51-40bf-9a3e-a1ae9c53f5f1

Adding demo that copilot recorded, not perfect, but shows buttons and how it works. Basically allows you to pause/resume reasoning, answer now, or to add more reasoning.

PR #3713, opened 2026-09-17
feat(v41): load and offload mlx_lm affine source checkpoints

LXD-8:
Based on `main` @ `d94d4cec` (v0.7.0.dev4).

## Why

Community `mlx_lm` conversions of DeepSeek V4.1 cannot be loaded at all. They declare their format in a top-level `quantization` dict instead of an `omlx_deepseek_v41` spec, pack each projection as a U32 weight with float scales and biases, and leave a mixed set of projections dense. The source loader only understood the official MXFP release and oMLX's own conversions, so such a checkpoint failed before the model was built. On `Vontra/DeepSeek-V4.1-Flash-MLX-2bit-MTP` (2-bit affine, 384 routed experts, 40 backbone layers, 222.4 GiB):

```
load()                -> ValueError: Unexpected target tensor shape: language_model.layers.4.ffn.experts.w3.weight
moe_offload_compat()  -> (False, 'Could not verify the expert checkpoint layout.')
```

The second line is why the offload toggle never appeared for the checkpoint.

## What

- `convert.py`: resolve bits/group_size/mode from the checkpoint's own declaration (`source_quantization_spec`), probe the plural `.scales`/`.biases` siblings the way the loader already probes singular E8M0 `.scale`, and `mx.dequantize` the projections that cannot stay packed — `wo_a`, `head`, `embed`, and any module with a dense `.bias`. The metadata set is claimed before the mapping is walked, so the reader no longer depends on the index listing `<module>.weight` before `<module>.biases` — `mlx_lm.utils.save_model` writes the weight map *sorted*, which puts the biases first, and a loader that only skipped them after reading their weight raised `Unexpected target tensor shape: language_model.embed.biases` on exactly the conversions this change is for.
- `moe_offload.py`: read affine source experts (U32 weight, plural metadata), keeping the weight packed so `QuantizedProjection` drives the matmul, and count the metadata in `expert_bytes`, which sizes the in-flight read window.
- `model_discovery.py`: accept a checkpoint whose top-level `quantization` dict declares affine mode with an integer bit width *and group size* and no per-module override the loader would reject — the three things `source_quantization_spec` needs. This admits a class outside the spec path — say if you would rather keep discovery strict and require a converted checkpoint. The gate only ever adds a candidate: what it refuses still falls through to the generic MLX heuristics (an mlx_lm shard is admitted by its `format: mlx` metadata or by the repo name), so loadability is not decided here alone and a declared non-affine conversion can still be listed and then fail at load.
- Engram tables report their affine format (`source_engram_tables` reads the declared spec), so a packed table is read as affine rather than as a plain embedding; requantizing an already-affine table is refused.
- `ModelConfig` is unchanged; no new dependencies, no reformatting, and the only new file is a test module.

## Behaviour delta

Third-party affine sources get `quantize_input=False`. `quantize_activation` is DeepSeek's published FP8 activation scheme, not part of an `mlx_lm` conversion's arithmetic (weights-only quantization, float activations), so `True` would inject error the checkpoint author never validated against. oQ3e is unaffected: it still inherits `True` because it re-quantizes the official FP8 release. `quantize_input=False` disables the grouped-expert decode path, the fused `w2` SwiGLU and the shared attention activation quantization; prefill's affine block kernels are gated on `w1.quantize_input == w3.quantize_input`, not on the value. I can make it an opt-in setting if you would rather trade numerics for decode speed.

## Verification

`tests/test_deepseek_v41_affine_source.py`, plus cases added to `test_deepseek_v41_moe_offload.py`, `test_deepseek_v41.py` and `test_model_discovery.py`:

```
python -m pytest tests/test_deepseek_v41.py tests/test_deepseek_v41_affine_source.py \
                 tests/test_deepseek_v41_moe_offload.py tests/test_deepseek_v41_oq.py \
                 tests/test_moe_offload_compat.py tests/test_model_discovery.py -q
358 passed
```

The new affine-source module fails on `main` at the current base (7 failed, 5 passed at `14194fe7`; the loader there has no affine path) and passes here. `tests/test_deepseek_v41_affine_source.py` loads the fixture with both index orders and the fixture writes the sorted one by default — the order `mlx_lm` writes — so the previous revision of this branch fails it (5 tests). Full `tests/`: 14012 passed, 340 skipped, 2 failed (`test_gdn_sidecar_quantization`, `test_integrations` — environment-dependent, outside this change's files).

Real-model check, M5 Max 128 GB, the checkpoint above with `moe_expert_offload_enabled`, `moe_expert_offload_resident_fraction: 0.125`, `deepseek_v41_engram_ssd_offload: true`, MTP off:

| | measured |
|---|---|
| load | 222.4 GiB checkpoint -> 27.61 GiB resident, 23.75 GiB Metal active |
| decode, warm | 4.43 and 5.18 tok/s (two 96-token greedy requests) |
| decode, cold | 2.20 tok/s (first 64-token request) |
| output | coherent; `17 * 23` answered 391 |

The docs record 5.6 tok/s decode for `Jundot/DeepSeek-V4.1-Flash-oQ3e-mtp` at 12.5% on the same machine. The two checkpoints differ in quantization and these were single runs, so treat this as a sanity check, not a benchmark. The fixture reproduces the biased-projection layout: 66 projections here (vision tower and aligner) are packed *and* carry a dense `.bias`, which a `QuantizedProjection` has nowhere to keep.

## Not covered

- MTP/DSpark for this checkpoint: offload forces `preserve_mtp=False`, and a resident load does not fit. Only the tiny fixture exercises affine `mtp.*` namespaces.
- Vision inference: the tower loads, but no image was run through it.
- Real 3/4/6-bit sources; the fixture and the resolver unit tests cover the widths.
- A declared non-affine `mlx_lm` mode is refused by the discovery gate, because the resolver rejects anything but `affine`. Its U32 packing is the same shape, but the metadata is E8M0 rather than float; supporting it is a separate change.
- `mx.dequantize` reproduces `mx.quantized_matmul` to quantization rounding rather than bit for bit: at 2 bits the two paths agree exactly on this M5 Max but differ by up to ~1e-3 on the CI runners, so the test pins a tolerance. `force_dense` has no alternative path, so nothing depends on equality.

## Notes

The offload half (the affine expert reads in `moe_offload.py`) is separable if you would prefer it in its own PR.


LXD-8:
Suggest squash merge to keep main history clean (6 incremental commits). All 356 tests pass. @jundot PTAL 🙏

jundot (approved):
The affine path preserves packed expert weights and reuses the existing offload machinery. I verified matching resident/offloaded prefill and decode outputs with a small checkpoint. I'll fix the alignment check that rejects valid 3-bit rows and trim the verbose comments and test descriptions in a follow-up. LGTM, merging.

PR #3714, opened 2026-09-17
feat(admin): bounded percentage field for the MoE expert offload resident fraction

LXD-8:
Based on `main` @ `d1889871` (v0.7.0.dev4).

## Why

The MoE expert offload resident-fraction selector only offered 12.5%, 25%, 50% and 75%. A model whose working set fits some other fraction has to round to the nearest preset, and the sizing helper reports fractions that are not presets at all.

## What

The selector is replaced by a single percentage field — a number input that is free while you type and clamps to 20-80 on blur, marked with `%` and captioned `(20-80)` — that drives `moe_expert_offload_resident_fraction`. The field is derived from the stored fraction, so reopening a model shows the closest whole percentage (a stored 12.5% opens as 13%, and is left untouched until the field is edited).

resident fraction

An entry outside the range says so while it is typed, and settling an *edited* field brings it back to the nearest allowed percentage rather than leaving a number that could never be saved:

out of range

The four preset catalog keys (`..._resident_12_5`, `_25`, `_50`, `_75`) have no other consumer and are removed, together with their catalog values. Two keys are added: `..._resident_range` for the caption and `..._resident_range_error` for the message.

## Behaviour

- Out-of-range or partial input is not written to the fraction while it is typed, so `0` on the way to `50` cannot rewrite the field to 20. It is reported instead, and `onMoeExpertOffloadResidentBlur` clamps the entry to the nearest allowed percentage once the field settles.
- A model stored outside 20-80 — the old 12.5% preset included — opens the modal without an error, and **focusing and leaving the field does not rewrite what is stored**: the clamp and the write both require the field to have been edited. The range error and the red border appear only after an edit, and the field rewrites the fraction only when what is typed is inside the range.
- The field is deliberately narrower than the setting. The API keeps accepting any fraction in `(0, 1]`, and `docs/MoE_Expert_Offload.md` now says so, so a value stored outside 20-80 through the API is shown as it is rather than corrected.
- No backend change — `model_settings` already accepts any fraction in `(0, 1]`.
- The range message carries a zh / zh-TW translation; the other catalogs keep the generated English fallback, as `scripts/normalize_i18n.py` writes them.

## Verification

```
python -m pytest tests/test_admin_model_settings_template.py tests/test_admin_model_settings.py \
                 tests/test_model_settings.py tests/test_admin_dashboard_draft_filters.py -q
154 passed

node tests/moe_expert_offload_ui.test.cjs
PASS: offload save/reopen, spec toggle exclusion and bounded resident fraction
```

`tests/moe_expert_offload_ui.test.cjs` covers reopening a stored fraction as a percentage, opening an out-of-range stored value without an error until the field is edited, editing it, the range clamp, **that a focus + blur with no edit leaves the stored fraction alone** (0.125 stays 0.125), and the selector being gone. It fails on `main` (`undefined !== 33`: the field has no value to reopen) and the blur case fails against the previous revision of this branch.


LXD-8:
All CI passes, 154 Python tests + Node.js UI tests. Small, self-contained change. @jundot ready to merge when convenient 🙏

jundot (approved):
The percentage input makes the resident fraction easier to tune, but I'm not convinced that limiting it to 20-80% is the right direction, especially since it removes the existing 12.5% option. I'll widen the range to 5-95% and preserve fractional values such as 12.5% in a follow-up fix. If there's a technical reason for the current bounds, please let me know. LGTM, merging.

PR #3715, opened 2026-09-17
feat(moe): offload qwen3_5_moe experts

LXD-8:
## Why

MoE expert offload is offered for `qwen4_exp` but not for `qwen3_5_moe`
(Qwen3.6-35B-A3B), even though those checkpoints stack their experts the same
way and run the same expert module.

## What

`qwen3_5_moe` is admitted by the eligibility allowlist and resolves its expert
tensors under `...mlp.switch_mlp`, the branch `qwen4_exp` already uses. Nothing
else changes: the adapter wraps a module by class, and mlx-vlm's `qwen3_5_moe`
imports `SwitchGLU` from `mlx_lm.models.switch_layers`, so the layers it walks
are already covered.

## Verification

Eligibility against real checkpoints on this machine, both of which returned
`(False, 'MoE expert offload is not supported for this model type.')` before —
the allowlist was the only thing refusing them:

```
rosingrind/Qwen3.6-35B-A3B-oQ5e-mtp                  -> (True, '')
root4k/Huihui-Qwen3.6-35B-A3B-abliterated-oQ4e-mtp   -> (True, '')
Jundot/Qwen3.8-Flash-Next-oQ4e-mtp (unchanged)       -> (True, '')
```

Real-model check, M5 Max 128 GB, `rosingrind/Qwen3.6-35B-A3B-oQ5e-mtp` (24.2 GiB)
with `moe_expert_offload_enabled`, `moe_expert_offload_resident_fraction: 0.25`
and MTP off:

| | measured |
|---|---|
| load | 8.62 GiB resident, 9.55 GiB Metal active |
| decode, cold | 4.57 tok/s (first 96-token request) |
| decode, warm | 24.89 tok/s (second identical request) |
| output | coherent and on topic |

The warm rate is the point of the change: the experts that stay resident plus the
page cache serve most routes, so a model that would otherwise occupy its full
24 GiB fits in under 10 GiB and still decodes at reading speed.

```
python -m pytest tests/test_moe_offload_compat.py tests/test_moe_expert_offload.py \\
                 tests/test_deepseek_v41_moe_offload.py tests/test_admin_model_settings_template.py -q
101 passed
```

The synthetic `qwen3_5_moe` case joins the supported-layout parametrization, and
leaves the unverified-type list. `docs/MoE_Expert_Offload.md` names the type.

## Not covered

The second checkpoint (`Huihui-...-oQ4e-mtp`) is verified for eligibility only,
and the mixture-of-experts routing was exercised at 0.25 residency only.

## Notes

`qwen3_5_moe` is admitted here by adding it to the model-type allowlist. #3717
removes that allowlist in favour of admitting any checkpoint whose tensors match
one of the layouts the adapter reads, which covers this model type without naming
it — verified by running that branch alone against the two checkpoints above.

So if #3717 lands, this PR is redundant and can be closed unmerged; if #3717 is
not wanted, this is the smaller way to get the same model type supported. The two
touch the same eligibility block, the same documentation paragraph and some of the
same tests, so they conflict: whichever merges second needs a rebase.

The allowlist is still the gate for every other family. A follow-up could let the
structural checkpoint validation be the admission condition instead, since it
already checks tensor names, shapes, dtypes and quantization metadata far more
strictly than a type list — but that changes what an experimental toggle promises,
so it belongs in its own PR.


LXD-8:
Closing this in favour of #3717, which admits the same model type by matching the
checkpoint's expert layout instead of extending the model-type allowlist — the
allowlist entry this PR adds would not survive that change, and the two touch the
same block, paragraph and tests.

Partial expert offload of a 35B model is a narrow audience, and that does not
justify a second change to the same gate. #3717 carries the load and decode
numbers measured on a Qwen3.6-35B-A3B checkpoint; the eligibility checks quoted
here were reproduced on that branch alone.

Thanks for the review bandwidth.

PR #3717, opened 2026-09-17
feat(moe): admit expert offload by checkpoint layout, not by model type

LXD-8:
Based on `main` @ `d94d4cec` (v0.7.0.dev4).

## Why

Expert offload was gated on a hardcoded list of four model types:

```python
_SUPPORTED_TYPES = frozenset({"deepseek_v41", "qwen4_exp", "gemma4", "olmoe"})
```

Everything after that gate is a much stronger statement — every expert tensor of every layer is checked for name, shape, storage dtype and declared quantization. A family that produces exactly those tensors was still refused, purely for not being on the list, and each new family needs its own PR to be added.

## What

`_inspect` tries each layout the common adapter reads — the stacked `[num_experts, ...]` `SwitchGLU` projections under `...mlp.switch_mlp`, the per-expert layout OLMoE conversions use, and `...experts.switch_glu` — and admits the checkpoint on the first layout that validates completely. `deepseek_v41` keeps its own adapter, dense models are still refused, and the model type is no longer consulted at all.

`...mlp.switch_mlp` covers `qwen4_exp`, `qwen3_5_moe` and OLMoE; `gemma4` MoE uses `...experts.switch_glu`. When no layout matches, the reported reason prefers a real defect (a wrong shape or dtype, a per-expert bias) over "this tensor is missing", which only means the layout was not this one.

## Behaviour delta, and what to watch

A checkpoint whose tensors match but whose **runtime** expert module is not a stock `SwitchGLU` — the custom weighted kernels of DeepSeek V4 and GLM — will show the toggle while the adapter leaves those layers resident, as it already does for any unmatched module. Two things keep that bounded:

- `estimate_offload_admission_bytes` still refuses those families a memory discount, so enabling the toggle does not hand anyone an optimistic footprint.
- The speculative-decoding conflict checks are separate and unchanged.

I have no checkpoint for those families to test against, so this is reasoning from the code rather than a measurement. If you would rather keep the list as a "verified on hardware" marker, say so and I will drop this in favour of adding types one at a time.

## Verification

```
python -m pytest tests/test_moe_offload_compat.py tests/test_moe_expert_offload.py \
                 tests/test_deepseek_v41_moe_offload.py tests/test_admin_model_settings_template.py \
                 tests/test_model_settings.py -q
216 passed
```

Eligibility over every checkpoint on this machine, which now follows the layout:

```
Qwen3.8-Flash-Next-oQ4e-mtp / -Uncensored / -oQ5e   -> (True, '')
Qwen3.6-35B-A3B-oQ5e-mtp                            -> (True, '')
Huihui-Qwen3.6-35B-A3B-abliterated-oQ4e-mtp         -> (True, '')
gemma-4-31B-it-qat-oQ4e-mtp                         -> (False, 'The model does not have the supported MoE geometry.')   # dense
Qwen3.8-27B-oQ5e-mtp / Qwen3.5-2B-5bit              -> (False, 'The model does not have the supported MoE geometry.')   # dense
diffusiongemma-26B-A4B-it-6bit                      -> (False, 'Checkpoint is missing expert tensor: ...')              # other layout
```

The two Qwen3.6 rows are admitted without `qwen3_5_moe` being named anywhere.

Real-model check, M5 Max 128 GB, `rosingrind/Qwen3.6-35B-A3B-oQ5e-mtp` (24.2 GiB) with `moe_expert_offload_enabled`, `moe_expert_offload_resident_fraction: 0.25` and MTP off — a model the allowlist never admitted, loaded through this change:

| | measured |
|---|---|
| load | 8.73 GiB resident, 9.55 GiB Metal active |
| decode, cold | 4.44 tok/s (first 96-token request) |
| decode, warm | 23.97 tok/s (second identical request) |
| output | coherent and on topic |

The adapter is unchanged, so this is what the same checkpoint does when it is admitted by layout instead of by name. The second Qwen3.6 checkpoint is verified for eligibility only, and the residency was measured at 0.25 only.

`test_matching_layout_is_admitted_whatever_the_model_type` is parametrized over model types that are named nowhere in the patch, and `test_unsupported_saved_setting_rejected_before_load` uses a checkpoint no layout matches, so it still covers the load-path refusal.

## Notes

`docs/MoE_Expert_Offload.md` no longer names a type list.

#3720 cherry-picks this change and adds a `model.layers.{layer}.ffn -> .switch_mlp` layout on top, so merging this one first shrinks #3720 to that extra layout.


williamxie1989:
Heads-up: #3720 builds on this PR (cherry-picked) — the layout-driven eligibility is what admits the new deepseek_v4 / glm5_next families without another model-type special case. Also added a `model.layers.{layer}.ffn` -> `.switch_mlp` layout entry there for the deepseek_v4 text checkpoints. If this merges first, #3720 rebases down to just its adapter + tests.

LXD-8:
Thanks — #3720 stacking on this branch is expected; this PR is the base for it. #3720 is still open, so rebasing it down to its own adapter and tests once this merges is fine.

LXD-8:
@jundot — would you take a look at this one when you have a moment?

Expert offload was gated on a hardcoded list of four model types, even though everything past that gate already checks every expert tensor for name, shape, storage dtype and declared quantization — so a family that produces exactly those tensors was refused purely for not being on the list, and each new family needed its own PR to be added. Admission now comes from the checkpoint's layout, and the model type is no longer consulted. CI is green on 3.11 / 3.12 / 3.13.

LXD-8:
This PR unblocks #3720 (stacking adapter support). All CI passes, 214 tests. @jundot PTAL when convenient 🙏

jundot:
I don't think matching checkpoint tensor names and shapes is enough to establish runtime compatibility. Model-specific sanitization and expert implementations still matter, so I'd keep the model-type allowlist and add `qwen3_5_moe` with the necessary path handling.

I reproduced a mismatch in the current patch with the pinned mlx-lm loader. A valid `qwen3_5_moe` checkpoint under `model.layers.*` passes eligibility, but the loader moves its runtime modules under `language_model.model.layers.*`. The offload adapter then wraps zero layers, while the admission estimate still assumes reduced memory usage.

I'll add Qwen support while keeping the architecture restriction, and verify that both checkpoint naming layouts resolve correctly and only receive a memory discount when offload actually applies.

PR #3718, opened 2026-09-17
feat(cli): macOS menubar sidecar with live prefill/decode status on `omlx serve`

williamxie1989:
### Problem

When running omlx from source in a terminal (`omlx serve`), there is no way to see what the server is *doing* without opening the dashboard: is it prefilling or decoding right now, and at what speed? The native oMLX.app covers this for app users, but CLI-first workflows have no equivalent.

### Proposal

`omlx serve` auto-spawns a lightweight menu bar **sidecar** (`omlx/menubar_sidecar.py`, PyObjC, ~450 lines) that renders the live inference state next to the clock:

| menu bar | color | meaning |
|---|---|---|
| `PP 45%` | blue | prompt processing (progress, t/s and ETA in the dropdown) |
| `GEN 42.1 t/s` | green | decoding (per-request speed / tokens / elapsed in the dropdown) |
| `WAIT 3` | orange | requests queued |
| `idle` / `offline` | gray | nothing running / server unreachable |

Dropdown also shows session averages (`/api/status`), cache hit rate, memory, plus Open Dashboard / Quit. Prefill takes priority over decode in the title, matching the native app's `MenubarStatsPoller`.

### Design

76a29578ad80ade542e65c3f34bfe976

- **Data**: polls `/admin/api/activity` (1 s) and `/api/status` (5 s) over loopback with a cookie session (auto-login on 401 via the configured api key). No in-process coupling — the sidecar is a separate stdlib+AppKit process.
- **Lifecycle**: spawned only *after* the port binds successfully; terminated in `serve_command`'s `finally`; `--parent` watchdog + pidfile dedup so the icon dies with the server and never stacks.
- **Gating — every other run mode keeps the original behavior**:
  - non-macOS → no-op
  - `OMLX_SUPERVISED` (set by oMLX.app, which owns its own menu bar) → no-op
  - PyObjC missing → prints a one-line `pip install "omlx[menubar]"` hint, server runs unchanged
  - `--menubar` / `--no-menubar` flags and `server.menubar` setting (default `true`) for explicit control
- **Testability**: state computation (`compute_live_state`), formatting and the autostart gate are pure functions; AppKit is imported lazily so the module imports headless.

### Tests

- 20 new unit tests in `tests/test_menubar_sidecar.py` — all pass (`pytest tests/test_menubar_sidecar.py`), including a subprocess check that importing the module never pulls in AppKit.
- `tests/test_settings.py` passes except 5 pre-existing failures in this environment (mlx_lm version mismatch, identical on unmodified HEAD).
- Manually verified end-to-end on M3 Ultra / macOS 26: live `GEN t/s` transitions under load, idle fallback, menu contents via AX inspection, SIGTERM cleanup, and that `OMLX_SUPERVISED` suppresses the icon.

### Naming

The feature is called **menubar sidecar** (`omlx/menubar_sidecar.py`, flag `--menubar`, setting `server.menubar`) to distinguish it from the native Swift `apps/omlx-mac` menubar app while reusing its vocabulary (`OMLX_SUPERVISED`, prefill-priority title logic).


williamxie1989:
## CI fix + flag persistence semantics

**CI (fixed in dca4b2c8):** `test_serve_hands_prebound_socket_to_uvicorn` calls `serve_command()` with a bare `argparse.Namespace` / fake settings lacking the serve-parser defaults → `AttributeError: 'Namespace' object has no attribute 'menubar'`. Guarded both lookups with `getattr` (matching the defensive style of `_apply_cli_overrides`). All three Python matrix jobs hit the same single test.

**On `--no-menubar` persistence:** `save_cli_overrides()` persists *every* explicit CLI flag to `settings.json` — this is existing upstream convention (`--port`, `--log-level`, etc. all persist the same way). So `--no-menubar` persisting is intentional and consistent: a user who passes it once stays off until they flip `server.menubar` back. Users who never pass the flag are unaffected — upgrades without a `menubar` key default to `True` via `from_dict`.

Headless/CI safety: pyobjc is an optional extra (`.[menubar]`), so `pyobjc_available()` is False on Linux runners and in envs without it → `should_autostart()` returns False (with a one-line hint) and nothing spawns.

PR #3720, opened 2026-09-17
feat(moe): expert offload for the DeepSeek V4 / glm5_next MoE block

williamxie1989:
## What

Extends MoE expert offload to the two remaining oMLX-native MoE families: **`deepseek_v4`** (DeepSeek-V4-Flash checkpoints) and **`glm5_next`** (GLM-5.3-Flash). Both run routed experts through `omlx/patches/deepseek_v4/switch_layers.py::SwitchGLU` — a custom class with native Metal block/pair `gather_qmm` kernels and split gate/up — which the common adapter could not wrap, so the compat layer kept hiding the toggle for both model types.

**Stacked on #3696 and #3717** (cherry-picked here; if those merge first, this diff shrinks to the adapter, the dispatch/admission edits, and the tests). The #3717 layout-driven eligibility is exactly what lets the new families be admitted by checkpoint layout instead of yet another model-type special case.

## Why one PR, not two

`glm5_next` and `deepseek_v4` share the *same* `SwitchGLU` class, the same projection layout, and therefore the same adapter file. Splitting them would mean cutting one file across two PRs with guaranteed merge churn, for zero review benefit. The only family-specific piece is the `compile_ffn` gate below, which belongs with the adapter.

## How it works

The module and its kernels are kept; only the projection **parameters** are swapped, in place, for resident slot tensors before lazy weights materialize:

- Each `{gate,up,down}_proj` weight/scales/biases becomes a `[capacity, …]` slot tensor plus an expert→slot `int32` map; `num_experts` is a property of the weight shape, so every kernel decision (sort threshold, MXFP4 block grid, native weighted-sum gating) follows the slots automatically.
- The module's own forward runs unchanged on slot-translated ids — bit-exact by construction wherever the resident model takes the same kernel path.
- Cache misses read the stacked slabs positionally (`pread`) on a bounded worker pool; over-capacity prefill is chunked on expert boundaries, and the weighted sum is applied exactly when the module's native forward would have applied it (mirroring the `glm_moe_weighted_sum` condition).
- LRU eviction, stats (`hit_rate`), kill switch (`OMLX_MOE_EXPERT_OFFLOAD=0`), and the shared walkers work unchanged.

### The `compile_ffn` fix (found in live testing)

The vendored `glm5_next` decoder layers compile their FFN block at decode shapes (`language.py` `compile_ffn` → `_ffn_c`). The offloaded block manages slots on the host — LRU map lookup, pread fetches — and cannot be traced into a compiled graph: the first real request died with `[eval] Attempting to eval an array during function transformations like compile or vmap is not allowed`. Wrapping now disables FFN compilation on layers whose block was offloaded; the native gather kernels still run eagerly, only the graph fusion is lost — and offload trades speed for memory anyway.

## Testing

- 15 unit tests (`tests/test_deepseek_v4_offload.py`): bit-exact decode / sorted prefill / native weighted-sum at full residency, bit-exact sorted prefill within capacity, rounding-bounded over-capacity chunks, LRU eviction and counters, skipped when uncovered, kill switch, common-entry dispatch, admission estimate, and a red/green regression for the `compile_ffn` gate.
- Full fast suite green (the 19 pre-existing failures on `origin/main` are unrelated native-kernel numerics drift, verified on a clean checkout).
- Live acceptance on the checkpoints below: wraps 42/43 layers, greedy outputs match the resident model, coherent generations across the whole benchmark grid.

## Benchmarks

Device: **MacBook Pro, Apple M5 Max, 128 GB RAM**. Resident fraction = `moe_expert_offload_resident_fraction`. Full model sizes: DeepSeek-V4-Flash-0731-2.4bit-mixed ≈ 86 GB on disk, GLM-5.3-Flash-oQ2e-mtp ≈ 99 GB (103.65 GB loaded estimate without offload).

### DeepSeek-V4-Flash-0731-2.4bit-mixed

| Resident | Test | TTFT (ms) | TPOT (ms/tok) | pp TPS | tg TPS | E2E (s) | Throughput (tok/s) | Peak Mem |
|---|---|---|---|---|---|---|---|---|
| 25% | pp1024/tg128 | 6836.9 | 62.47 | 149.8 | 16.1 | 14.772 | 78.0 | 25.62 GB |
| 25% | pp4096/tg128 | 14314.1 | 68.50 | 286.2 | 14.7 | 23.015 | 183.5 | 26.27 GB |
| 25% | pp8192/tg128 | 29545.4 | 69.66 | 277.3 | 14.5 | 38.394 | 216.7 | 26.75 GB |
| 25% | pp16384/tg128 | 62393.4 | 79.35 | 262.6 | 12.7 | 72.473 | 227.8 | 27.70 GB |
| 50% | pp1024/tg128 | 5638.7 | 54.60 | 181.6 | 18.5 | 12.574 | 91.6 | 43.76 GB |
| 50% | pp4096/tg128 | 15106.0 | 63.21 | 271.1 | 15.9 | 23.135 | 182.6 | 44.64 GB |
| 50% | pp8192/tg128 | 32297.1 | 63.30 | 253.6 | 15.9 | 40.337 | 206.3 | 45.15 GB |
| 50% | pp16384/tg128 | 69310.6 | 66.78 | 236.4 | 15.1 | 77.794 | 212.3 | 46.11 GB |
| 75% | pp1024/tg128 | 5858.4 | 48.25 | 174.8 | 20.9 | 11.987 | 96.1 | 61.99 GB |
| 75% | pp4096/tg128 | 11365.8 | 52.72 | 360.4 | 19.1 | 18.062 | 233.9 | 62.55 GB |
| 75% | pp8192/tg128 | 23286.2 | 54.51 | 351.8 | 18.5 | 30.211 | 275.4 | 63.03 GB |
| 75% | pp16384/tg128 | 47829.1 | 55.32 | 342.6 | 18.2 | 54.857 | 301.0 | 63.98 GB |

### GLM-5.3-Flash-oQ2e-mtp

| Resident | Test | TTFT (ms) | TPOT (ms/tok) | pp TPS | tg TPS | E2E (s) | Throughput (tok/s) | Peak Mem |
|---|---|---|---|---|---|---|---|---|
| 25% | pp1024/tg128 | 9118.5 | 110.05 | 112.3 | 9.2 | 23.098 | 49.9 | 38.89 GB |
| 25% | pp4096/tg128 | 23056.1 | 120.51 | 177.7 | 8.4 | 38.364 | 110.1 | 40.89 GB |
| 25% | pp8192/tg128 | 48729.8 | 132.42 | 168.1 | 7.6 | 65.550 | 126.9 | 41.32 GB |
| 25% | pp16384/tg128 | 100857.9 | 148.05 | 162.4 | 6.8 | 119.663 | 138.0 | 42.44 GB |
| 50% | pp1024/tg128 | 9064.1 | 86.91 | 113.0 | 11.6 | 20.105 | 57.3 | 61.04 GB |
| 50% | pp4096/tg128 | 21836.4 | 98.62 | 187.6 | 10.2 | 34.364 | 122.9 | 63.03 GB |
| 50% | pp8192/tg128 | 47776.1 | 107.95 | 171.5 | 9.3 | 61.489 | 135.3 | 63.46 GB |
| 50% | pp16384/tg128 | 98268.7 | 123.26 | 166.7 | 8.2 | 113.926 | 144.9 | 64.58 GB |
| 75% | pp1024/tg128 | 9545.9 | 70.20 | 107.3 | 14.4 | 18.465 | 62.4 | 83.12 GB |

Notes on the GLM 75% rows: the remaining grid cells were refused by the engine's pre-existing prefill safety guard (predicted peak would exceed 95% of the Metal wired-memory ceiling at long contexts) — the memory enforcer doing its job, not an offload bug; the single-request row ran fine.

Observations:
- Peak memory scales cleanly with resident fraction (~26 → 44 → 62 GB on DSv4-0731; its expert table is ≈34 GB, so 25% residency saves ≈25 GB).
- Decode (tg TPS) improves monotonically with residency as expected; at 25% the LRU misses dominate early and settle as the conversation warms up.
- DSv4-0731 decode holds 14–21 tok/s across the grid; GLM-5.3-Flash (288 experts, top-8) is slower at low residency, as its expert working set is larger.

## Known limitations / policy

- MTP remains mutually exclusive with offload (existing `validate_moe_expert_offload` policy) — both checkpoints ship MTP heads, so users must disable speculative decoding first.
- Offloaded `glm5_next` layers lose the compiled-FFN decode optimization (see above).
- Phase 2 (not here): `deepseek-v4-flash` base and `deepseek-v4-flash-vision-exp`.

## Base note

Branch is based on `76d19fed` + #3696 + #3717; verified zero file overlap with the three upstream commits since (`a1663771`, `24a923ab`, `a3524a9c`), including the mlx-vlm upgrade — the `compile_ffn` path is untouched there.


jundot:
I closed #3717 and added `qwen3_5_moe` support on main while retaining the model-type allowlist. Please rebase this PR onto the updated main, drop the layout-only admission change from #3717, and explicitly add the model types covered by the new adapters with their corresponding checkpoint validation. The #3696 dependency can remain pending its separate review.

williamxie1989:
Thanks — rebased onto current main (which brings in the `qwen3_5_moe` support), and the layout-only admission from #3717 is dropped; the model-type allowlist is retained.

**Explicit admission for the new families** (`moe_offload_compat`):
- `deepseek_v4` and `glm5_next` added to `_SUPPORTED_TYPES`.
- Family-specific checkpoint validation in `_inspect`: `n_routed_experts` + `first_k_dense_replace` for both; `mlp_layer_types` per-layer sparse selection for `glm5_next` (dense layers skipped, an all-dense checkpoint rejected).
- Stacked `model.layers.*.ffn.switch_mlp` / `language_model.model.layers.*.mlp.switch_mlp` slabs validated per layer, per projection, per field (shape + dtype) — exactly what the adapter reads positionally; a per-expert layout is rejected since the adapter would wrap nothing.

**Admission estimate** (`moe_expert_offload`): `deepseek_v4` / `glm5_next` now get real stacked savings; `deepseek_v41` keeps returning `full_size` here — its own estimator owns that family. #3696 remains the stack base, pending its separate review.

**Tests**: compat suite covers both families admitted on a matching layout, plus dense-layer skip, missing-slab, and all-dense rejection; the "unverified type" test now uses genuinely unsupported types. Fast suite (`pytest tests/ -m "not slow and not integration"`): 14316 passed — the remaining 14 failures + 8 errors reproduce identically on a clean `origin/main` checkout on this machine (native-kernel numerics drift and missing local fixtures), none introduced here.

jundot (approved):
The allowlist and checkpoint validation now match the intended scope, and I verified the adapter's native-kernel output comparisons and FFN compilation fix. I also reproduced a descriptor leak in the DeepSeek adapter during repeated load/unload cycles. I'll prepare and verify a fix using the shared checkpoint reader, incorporate the latest #3696 changes, then merge this PR and apply the follow-up immediately afterward.

PR #3734, opened 2026-09-18
perf: preserve Prism KV capacity after batches shrink

### LINKED ISSUE #3727: Support Prism ML Hadamard packs (Ternary Bonsai 2 27B, MLX 2-bit): dispatch to the pack's bundled loader + register its MODEL_CONFIG format

`prism-ml/Ternary-Bonsai-2-27B-mlx-2bit` cannot be loaded. oMLX correctly classifies it as a VLM (its `config.json` carries `vision_config`), then both the VLM path and the LLM fallback fail:

```
Model 'Ternary-Bonsai-2-27B-mlx-2bit' failed to load:
  VLM load failed: Model type prism_hadamard_qwen35 not supported.
   Error: No module named 'mlx_vlm.speculative.drafters.prism_hadamard_qwen35';
  LLM fallback also failed: Model type prism_hadamard_qwen35 not supported.
  Reload models after fixing the files to retry.
```

That `not supported` is **expected and correct**: the pack must be loaded by the loader it bundles in `runtime/`. What is missing is oMLX wiring that loader in.

**Why we cannot simply make the model type resolvable**

The pack stores its language projections in a **rotated basis**: every matrix is blockwise Hadamard-transformed (block 1024, explicit ±1 signs, declared in `hadamard.json`) before the ternary `{−1, 0, +1}` assignment, and the runtime must apply the matching transform to **activations** plus the inverse to the **embedding lookup**. Stock mlx-lm / mlx-vlm do neither, and crucially they **do not raise** — they return wrong output.

Prism ML's own demo therefore refuses to serve these packs over MLX servers (`scripts/start_mlx_server.sh` in PrismML-Eng/Bonsai-demo):

> "No MLX server for Bonsai 2 yet. Its MLX pack needs the loader bundled in the pack, which mlx_lm.server and mlx_vlm.server do not use; **serving through them would return wrong output**."

So the fix has to be "load through the bundled loader", **not** "let the model type resolve". I verified the wrong road fails loudly for the wrong reason, too: normalizing `model_type` to `qwen3_5` and feeding it to the stock path immediately trips the pack's own guard (`Unsupported packed model schema`) — and any loader that bypasses that guard silently produces garbage.

**There is a second, independent failure that survives a working loader**

Even once the model loads, `mlx_vlm.prompt_utils.MODEL_CONFIG` has no `prism_hadamard_qwen35` entry, so `get_message_json()` raises `Unsupported model`. That aborts oMLX's `_format_messages_for_vlm_template()` as a whole and falls back to the generic formatter, which emits **no image placeholders** — while oMLX still extracts the images. Every vision request then dies with:

```
ValueError: Image features and image tokens do not match: tokens: 0, features 64
```

This is the same failure mode oMLX already documents for `glm5_next` in `omlx/patches/mlx_vlm_glm5_next_compat/__init__.py`, and the same fix applies: register the type against the base model's `MessageFormat` (`qwen3_5` → `LIST_WITH_IMAGE_FIRST`).

Follow the existing `mlx_vlm_*_compat` convention and dispatch through the extension point that `paroquant` / `qwen38_modelopt_mixed` already use, i.e. `omlx/utils/model_loading.py::maybe_load_custom_quantization()`:

1. New `omlx/patches/mlx_vlm_prism_hadamard/__init__.py`:
   - `is_supported_config(config)` — identifies the pack from metadata the pack itself declares (`model_type == prism_hadamard_qwen35`, `components.vision`), so a renamed directory or a future Bonsai size still works
   - `load(dir)` — imports `<pack>/runtime/` and calls its own `vision_artifact.load_vl_model()`
   - `apply()` — registers `prism_hadamard_qwen35` in `MODEL_CONFIG` (copying `qwen3_5`'s format)
   - **sha256-pins `runtime/*.py`** against the revision Prism ML publishes in `Bonsai-demo/scripts/bonsai2-runtime.sha256`, refusing anything that does not match. Because the loader applies weight-basis transforms, an unreviewed `runtime/` does not fail — it silently produces wrong numbers, which is worth the pin.
2. One 21-line insertion in `model_loading.py` (the diff is below).

Non-Prism models are untouched: the branch is only entered when `model_type == prism_hadamard_qwen35`.

**Two implementation details worth knowing**

- `<pack>/runtime/` has **no `__init__.py`** and its modules import each other **unqualified** (`from codec import transcode`, `from runtime import Packed`) — it is designed to be put straight on `sys.path`. A plain package import therefore raises `ModuleNotFoundError`. The loader builds a synthetic namespace package (`types.ModuleType` + `__path__`), then `import_module("<pkg>.<name>")`, plus bare-name aliases.
- Reuse the pack's own `build_processor()`. It already routes around the torch-gated `AutoProcessor`, which cannot resolve our `model_type` through transformers' `AutoConfig` (this is exactly the problem `_patch_torch_free_image_processor` handles for `glm_ocr` / `dots_ocr`).

- **Just map the model type to `qwen3_5`** — rejected, see above: it either fails the pack's schema guard or silently yields wrong output.
- **Vendor `runtime/` into `omlx/patches/.../vendor/`**, as `mlx_vlm_inkling_compat` does — viable, but Prism ML version-pins and revises that runtime, so vendoring would need re-syncing. I would rather reference the pack's own copy with a hash pin; happy to do it your way if you prefer vendoring.

**Additional context**

I have a working implementation and tested it locally against `main`'s `model_loading.py` (v0.7.0.dev3) — the anchor for the insertion is present and the patch applies with `patch -p0`. Verified on oMLX 0.6.4 / build 260830015308-macos26-27, macOS 26.5.1, Apple Silicon, mlx 0.32.0 / mlx-lm 0.31.3 / mlx-vlm 0.6.3:

| check | result |
|---|---|
| text: `17+26=` → `43`, `9*9=` → `81`, `6*7=` → `42` | pass |
| vision (yellow circle on blue): → `Yellow`, reasoning describes the image | pass |
| unrelated models (Llama-3.2-1B, all-MiniLM-L6-v2 embedding, Qwen3-VL-8B) | unaffected |
| 20 sequential requests | 20/20, 0 tracebacks, RSS ≈ 9 GB, no leak |
| `runtime/*.py` sha256 vs Prism ML's published pin | byte-for-byte match |

The dispatch diff (against `main`):

```diff
@@ -1328,7 +1348,28 @@ def maybe_load_custom_quantization(
         return None
+
+    from ..patches.mlx_vlm_prism_hadamard import (
+        is_supported_config as _is_prism_hadamard_pack,
+        load as _load_prism_hadamard_pack,
+    )
 
+    if is_vlm and _is_prism_hadamard_pack(config):
+        # Prism ML "Hadamard" packs (Ternary Bonsai 2) store their language
+        # projections in a rotated basis ... mlx-lm / mlx-vlm do not apply the
+        # matching activation transform, so they return *wrong output rather
+        # than an error*.  Use the pack's bundled loader instead.
+        logger.info(
+            "Prism Hadamard pack detected for %s; using its bundled loader",
+            model_name,
+        )
+        return _load_prism_hadamard_pack(model_name)
+
     quant_config = config.get("quantization_config")
```

Note oMLX already ships `omlx/custom_kernels/bonsai/` and `omlx/patches/bonsai_t5_load.py`, i.e. the Bonsai family is already known here — this is the missing member of that family. I am happy to open a PR with the patch module + a regression test if you want it; just tell me whether you prefer the bundled-loader-with-hash-pin form or a vendored copy.

samfenwick:
When a two-request Prism batch shrinks to one, mlx-vlm extracts and re-merges the surviving attention caches on every token, discarding spare capacity. This preserves those buffers for unpadded singleton decode. Padding, prefill, multiple rows, capture sinks and unfamiliar arguments keep upstream handling. oMLX already uses regular `KVCache` for standalone requests.

This is now only the cache fix. The experimental FP16 activation conversion, precision cache namespace and dashboard changes have been removed. Native FP32 behavior and weights remain unchanged. Native Prism loading and FP32 attention are already on main through #3782 and #3795; their duplicate implementations are not in this PR.

## Actual serving measurements

[Exact harness, raw producer/cache events, token IDs and independent verification](https://gist.github.com/samfenwick/24a9f5441fdb60d71ecf9f73db312e02).

Apple M1 Max / 64 GB, MLX 0.32.2, mlx-vlm `ea79808c`, `prism-ml/Ternary-Bonsai-2-27B-mlx-2bit`. Untouched current main `9dff1dd7` versus `ed090e71`, identical native builds and native FP32 in both arms. Real HTTP streaming, concurrency two, stock balanced bursts, thinking/speculation/KV quantization off. An 8,386-token prompt reuses 8,379 tokens. A 32-token request joins a 256-token request after its first content chunk; the trace checks the transition from two rows to one unpadded survivor. Fresh-process A/B/B/A, discarded warmups, four retained samples per arm/case.

| Scenario | Main tail tok/s | PR tail tok/s | Main complete reply | PR complete reply |
| Standalone control | 18.55 | 18.52 | 14.430 s | 14.417 s |
| Overlap, then one survivor | 12.86 | 18.96 | 22.268 s | 17.040 s |

These are medians: **47.5% higher survivor-tail throughput and 23.5% less time for the complete overlapping reply**. The standalone control changes by -0.2% in tail throughput and -0.1% in complete-response time; no standalone gain is claimed.

All retained output token IDs, text and cache reuse match. The fixed 128-token measurement window is entirely after the batch shrinks. No cold-prefill, general memory reduction or broad task-quality claim is made. These refreshed results include current main's first-token scheduling changes; the earlier measurements remain historical evidence rather than being relabelled as this revision.

## Scope and validation

The cache ownership behavior belongs in mlx-vlm. A direct upstream fix is now proposed in https://github.com/Blaizzy/mlx-vlm/pull/2336; that change preserves the existing decoder class and generic cache API. This downstream hook can be removed once oMLX pins an upstream version containing the fix. The performance numbers here measure the oMLX implementation, not the separate upstream PR.

- 233 focused tests pass across native checkpoint loading, unchanged logits/weights/FP32 precision, cache contents/capacity, runtime arguments, vision and cache boundaries.
- Default suite: **14,595 passed, 150 skipped, 85 deselected, six failed**. Untouched current main has 14,583 passes and exactly the same six failure identities (FP32 decode-SDPA, GLM fused parity and Qwen4 hyper-connection tests). There are no added failures; this is not a clean local full-suite pass.
- 14 measurement/parser checks pass, including deliberately corrupted evidence; published bytes and restored verifier checks are verified independently.
- Black, Ruff and whitespace checks pass for the changed files.

The README and `docs/TESTING.md` describe the narrow scope and real serving protocol. No activation precision setting remains in this PR.


hamstead:
@samfenwick Great work on this. The implementation and testing look solid. One thing I noticed is `prism_hadamard_qwen35` isn't currently in `_QWEN_VISION_MODELS` or `_GRID_VISION_MODELS` in `omlx/engine/vlm.py`. Since it uses the regular Qwen vision tower, it looks like vision itself will work, but it may miss some of oMLX's Qwen-specific vision caching and multi-image handling. Would it make sense to add `prism_hadamard_qwen35` to those sets, or otherwise have it inherit that behavior from `base_model_type`?

samfenwick:
Thanks @hamstead, fixed in 3c186343. Prism now uses the Qwen vision cache and multi-image paths. I added regression tests and checked image reuse, reordering and multiple turns with the real model.

V1sMu21:
I filed #3727, so let me add an independent check from the machine that hit the original failure — tested against this PR's head (`3c18634`).

**What I can confirm works**

- `_validate_config` accepts the published pack as-is. No gate rejects it: `schema_version=2`, `model_type=prism_hadamard_qwen35`, `base_model_type=qwen3_5`, `quantization={bits:2, group_size:128, mode:affine}`, `gdn_activation_layout=grouped`, 402 module records.
- `load(path, is_vlm=True)` builds the model and installs the packed modules (~3.3 s here), and `MODEL_CONFIG.setdefault(MODEL_TYPE, MODEL_CONFIG["qwen3_5"])` does register — which is exactly what removes the silent `tokens: 0, features N` image dropout reported in the issue.
- Vendoring the transform instead of executing `<pack>/runtime/` is the better call. It deletes the "import Python from a downloaded checkpoint" surface completely, and it is strictly safer than the hash-pinned bundled-loader approach I originally proposed. The `_install_packed` shape/dtype/finiteness checks and the duplicate-shard check are a good addition.

**One compatibility note, with the caveat that this is NOT a bug against `main`**

On the oMLX 0.6.4 runtime, `mlx_vlm/models/qwen3_5/language.py` (2664 lines) forwards `gdn_sink` into the inner model:

```
2555:  gdn_sink: Optional[list] = [] if capture_layer_ids is not None else None
2558:  out = self.model(
2565:      gdn_sink=gdn_sink,
```

`CapacityPreservingModel.__call__` does not accept that kwarg, so the first decode step raises:

```
TypeError: CapacityPreservingModel.__call__() got an unexpected keyword argument 'gdn_sink'
```

It is reachable through this PR's own entry point: `prism_hadamard.load()` sets `language_model.model.__class__ = CapacityPreservingModel`, and then generation fails on token one. Text and vision both fail this way here.

Importantly, `main` pins mlx-vlm at `3fb24e90`, whose copy of that file has **no** `gdn_sink` at all (0 occurrences, 1965 lines), so `main` is unaffected and this is a release/pinned-commit skew rather than a defect in this PR. Worth knowing only if 0.6.4 users are meant to run this; if so, tolerating and forwarding unknown kwargs (`**kwargs`) would cover it.

Reproduce:

```bash
grep -c gdn_sink <site-packages>/mlx_vlm/models/qwen3_5/language.py   # 20 here, 0 at 3fb24e90
```

then load `prism-ml/Ternary-Bonsai-2-27B-mlx-2bit` through `maybe_load_custom_quantization(path, is_vlm=True)` and call `mlx_vlm.generate(...)`.

Happy to re-run the full text/vision/long-run matrix here once that is settled — I have the checkpoint resident and a working reference implementation to compare outputs against.

samfenwick:
Thanks @V1sMu21, addressed in 909cd862 - text and vision now pass on both runtimes I tested.

hamstead:
@samfenwick I just noticed #3782 was merged yesterday and adds Ternary Bonsai 2 support through the newer `mlx-vlm` revision. Does that make most of the model support work in this PR unnecessary now?

It still looks like there are some useful pieces here that #3782 doesn't cover, especially the decode/KV capacity optimization, bounded FP32 attention handling, and optional FP16 activation path. Maybe it would make sense to rebase this onto current `main` and keep the parts that are still relevant.

samfenwick:
Thanks @hamstead, I've rebased this on main and removed the model support that's already upstream. I've repurposed the PR for the remaining cache optimisations and optional FP16 mode. CI is green.

jundot:
oMLX already keeps regular KVCache for standalone requests, so the capacity optimization mainly applies when a multi-row batch shrinks to one unpadded row. Given that limited scope, would it make more sense to fix this in mlx-vlm and pick it up through a dependency update?

I'm open to carrying it here if there's a measurable benefit in the actual oMLX serving path. Could you measure the improvement in that scenario? My concern is whether the gain justifies maintaining the runtime patch alongside the separate FP16 path.

PR #3739, opened 2026-09-18
fix: keep usage history readable on mobile

samfenwick:
On narrow screens, the usage table squeezes long model names into a few characters per line, making rows several hundred pixels tall. Give the table and model column readable minimum widths, wrap names at natural boundaries, and keep horizontal scrolling inside the card. Filters can wrap, and the scroll region is keyboard-focusable. The change is scoped to Usage History, including narrow dashboard widgets on desktop.

Validation: Chromium and WebKit at 320–2560px, light/dark themes, long identifiers, narrow desktop containers and 200% text sizing. Checked access to the final column and Chromium keyboard scrolling. All five usage JavaScript tests pass. Tailwind rebuilt with no generated diff. Python 3.13 default suite: 13,920 passed, with the same two Qwen4 numerical failures previously reproduced on unchanged upstream.

Screenshots use the real usage component with entirely synthetic data in an isolated preview; no personal information or live-server data is included.

| Phone before | Phone after |
| --- | --- |
| Before: model names break into tiny fragments and rows grow tall | After: readable model names and compact scrollable rows |

Full-dashboard evidence — these captures show the production dashboard template, JavaScript and GridStack layout with synthetic APIs, from the navbar through all eight sections to Engine Versions. Compared with upstream at the same widths, the six sections below Usage History retain their widths and heights; no sections overlap and the page has no horizontal overflow. Also checked at 390px.

**1024px desktop · dark · full page**

Full dashboard at 1024px

**1440px desktop · dark · full page**

Full dashboard at 1440px

**1920px desktop · light · full page**

Full dashboard at 1920px

**Closer view below Usage History · 1440px**

Lower dashboard including active models, cache and endpoints

<details>
<summary>Tablet, 768px</summary>

Tablet usage table

</details>


jundot (approved):
The fix reproduces on my side. At a 320px block width the model column goes from 61px to 316px, and the tallest row drops from 465px to 85px. The scroll stays inside the card, and the filter selects wrap instead of overflowing.

I pushed one small commit on top of yours that drops `max-width: 20rem` from the first column. In Chromium that cap changes the desktop layout: at 1440px the model column shrinks from 857px to 582px, and at 1024px mid-length ids start wrapping where they fit on one line before. On narrow widths the line barely does anything (316px vs 346px, same row heights), so removing it restores the current desktop widths without touching the mobile result.

LGTM, merging.

PR #3746, opened 2026-09-18
fix(ane): avoid impossible sequence-length guidance below the ANE minimum

tc3oliver:
## Problem

`enable_qwen35_ane_prefill()` and `configure_qwen35_ane_prefill_scheduler()` both
require an ANE prefill `sequence_length` that is a multiple of 64 and at least
1024.

With block-aware caching on, the effective delivered prefill width is not the
configured step: boundary snapshots cut every chunk at the next cache block
edge, so `paged_cache_block_size` caps it. That width can be below the ANE
minimum — 512 is a realistic setting for fine-grained prefix reuse.

In that case `configure_qwen35_ane_prefill_scheduler()` warns:

```
Qwen ANE prefill sequence_length=2048 exceeds the delivered prefill chunk width
(~512 tokens). Chunks narrower than the compiled shape cannot tile onto it, so
the ANE will compile but never execute. Set sequence_length=512 or smaller.
```

Following that advice raises `ValueError: ANE prefill sequence_length must be a
multiple of 64 >= 1024` from the validator at the top of the very same function.
The recommendation cannot be acted on, and it hides the real situation: at a
delivered width below 1024, *no* `sequence_length` is executable, so the setting
being named is not the one that can fix it.

Found while bringing up ANE prefill on a real block-aware configuration, where
the ANE compiled its programs and then executed zero operations.

## Change

Split the warning into the two cases the geometry actually has.

- **Delivered width >= the minimum.** Keep recommending a smaller shape, now
  rounded down to the alignment so the recommended value is one the validator
  accepts. Previously the raw delivered width was echoed, which is also
  unusable when it is not a multiple of 64.
- **Delivered width < the minimum.** Report that the geometry itself is not
  executable: every legal `sequence_length` is wider than the delivered chunks,
  the ANE will compile and never execute, and the fix is a wider effective
  prefill chunk width (capped by the paged cache block size once block-aware
  caching is on) or disabling ANE prefill for the model.

The 1024 floor and the 64-token alignment were duplicated literals in the
scheduler check and the enable path. Both now read `_ANE_MIN_SEQUENCE_LENGTH`
and `_ANE_SEQUENCE_LENGTH_ALIGNMENT`, so the rejected range and the warned
range cannot drift apart.

The message deliberately names the mechanism ("capped by the paged cache block
size once block-aware caching is on") rather than a settings path: the block
size is adjusted after configuration by `_enlarge_block_size_for_arrays_cache()`
and `_align_block_size_with_rotating_window()`, so the configured value is not
reliably the effective one.

## Scope

Diagnostics only. No change to scheduling, chunk widths, ANE execution, cache
geometry, bank release, or any numerical path. The warning's trigger condition
and the function's return value are untouched; the only observable differences
are the log text and the alignment of the recommended value.

## Validation

- 380 tests pass across `test_qwen35_ane_prefill.py`, `test_ane_tuning.py`,
  `test_k2_ane_tuning.py`, `test_k2_ane_ui_presets.py`,
  `test_qwen35_gdn_prefill.py`, `test_qwen35_gdn_prework.py`,
  `test_prefill_boundaries.py`, `test_prefill_boundary_paths.py`,
  `test_scheduler_chunked_prefill.py`
- `ruff check` clean on both changed files

New regression coverage in `test_qwen35_ane_prefill.py`:

- an oversized shape over a legal delivered width recommends an aligned value,
  and that value is fed back through the validator to prove it is accepted
- a sub-minimum delivered width reports the impossible geometry, names the
  minimum, and never recommends a shape the validator refuses
- the validator and the warning are asserted to share one named minimum
- the pre-existing valid and no-warning cases are kept unchanged


jundot (approved):
The warning now avoids recommending rejected shapes, and I verified that a 1500-token cap recommends the valid 1472-token shape. LGTM. I'll handle the existing tail-padding exception in a small follow-up, since padded short chunks can still reach ANE execution.

PR #3750, opened 2026-09-18
fix(qwen35): stop double-shifting the MTP head's pre-fc norms

### LINKED ISSUE #3742: Lightning MTP draft acceptance drops 72% -> 35% on Qwen3.6-35B-A3B after mlx-vlm upgrade (#3719)
Hey Jundot :), 

here i am again, my Agents found something in regards to the speed regression that i noticed. If you look at the [omlx.ai performance bench entries](https://omlx.ai/benchmarks/performance?chip=M3&cursor=WyJwZXJmb3JtYW5jZTpjaGlwPU0zJmRldmljZT0lNUIlMjJNMyUyMiUyQyUyMlVsdHJhJTIyJTVEJmhpZGVfc3BlY3ByZWZpbGw9MSZtb2RlbD1xd2VuMy42LTM1Yi1hM2Itb3E2IiwiMjAyNi0wOS0xNSAxMjoxMjozNiIsNDY0ODMyLDMsNCwzMCwicHJldiJd&device=%5B%22M3%22%2C%22Ultra%22%5D&hide_specprefill=1&model=qwen3.6-35b-a3b-oq6) you can also see a big change after 16.09.2026

**Describe the bug**

Lightning MTP draft acceptance for `Qwen3.6-35B-A3B-oQ6-mtp` collapses from ~72% to ~35% starting at commit `a1663771` ("feat: upgrade mlx-vlm and adapt cache and MTP integration", #3719), cutting generation throughput by **1.67x** (116.7 → 70.0 tok/s mean at 16K context).

This is an **acceptance/correctness regression, not a speed regression**. The backbone forward cost per cycle is unchanged (18.16 ms before vs 18.47 ms after). The verifier simply rejects drafts it previously accepted, so tokens-per-cycle drops from 2.48 to 1.44.

A git bisect with dependencies and custom Metal kernels rebuilt at every step isolates this to a single commit, with no overlap between the two distributions.

**To Reproduce**

1. Install oMLX at `76d19fed` (parent of `a1663771`) in a clean venv with its pinned deps (`mlx-lm@872ae88d`, `mlx-vlm@78b96eb5`), built with `OMLX_WITH_CUSTOM_KERNEL=1`.
2. Load `Jundot/Qwen3.6-35B-A3B-oQ6-mtp` with `mtp_enabled: true`, ANE off, DFlash off, TurboQuant off.
3. Run the built-in throughput benchmark: prompt length **16384**, generation length **128**, context profile **Code (Mixed)**, warmup **quick**. Repeat 6x.
4. Observe TG ≈ 102–144 tok/s (mean 116.7), and MTP telemetry showing `accept ≈ 60–90%`, `tok/cycle ≈ 2.0–3.3`.
5. Check out `a1663771`, reinstall (this bumps only `mlx-vlm` to `3fb24e90`), rebuild kernels, repeat the same benchmark 6x.
6. Observe TG ≈ 66–77 tok/s (mean 70.0), with `accept ≈ 28–47%`, `tok/cycle ≈ 1.3–1.8`.

**Expected behavior**

Draft acceptance and tokens-per-cycle should be preserved across the mlx-vlm upgrade. Generation throughput should remain at ~110–145 tok/s rather than dropping to ~70.

**Screenshots**

Local benchmark screenshots from 2026-08-24 on this machine show pp16384/tg128 at **TTFT 6082 ms, TPOT 7.62 ms, PP 2693 tok/s, TG 132.2 tok/s** — consistent with the pre-regression measurements above.

Community benchmarks on omlx.ai filtered to M3 Ultra + this exact model show results dropping from ~120–140 tok/s to ~80 tok/s after 2026-09-16, so this is not machine-specific.

**Desktop:**
- macOS Version: 26.7 (build 25G229)
- oMLX Version: 0.7.0.dev4 (`14194fe7`); regression introduced at `a1663771`
- Hardware: Apple M3 Ultra, 80 GPU cores, 512 GB unified memory
- Model: `Jundot/Qwen3.6-35B-A3B-oQ6-mtp` (6-bit, group size 64)

**Additional context**

Bisect results. Built-in benchmark, pp16384/tg128, Code (Mixed). Isolated worktree + venv; custom Metal kernels rebuilt and verified on the FAST path (all 5 packs, `abi=ok`) at every step:

| commit | mlx-lm | mlx-vlm | n | mean TG | max TG |
|---|---|---|---|---|---|
| `4d4750ee` (#3695 multi-request MTP) | ab1806e8 | 78b96eb5 | 8 | 112.2 | 129.3 |
| `e7b86ddb` (#3702 MTP fixes) | ab1806e8 | 78b96eb5 | 6 | 110.6 | 127.5 |
| `76d19fed` (parent of #3719) | 872ae88d | 78b96eb5 | 6 | **116.7** | 144.0 |
| `a1663771` (#3719) | 872ae88d | **3fb24e90** | 6 | **70.0** | 77.0 |
| `14194fe7` (v0.7.0.dev4) | 872ae88d | 3fb24e90 | 6 | 73.1 | 87.1 |

No distribution overlap: min of `76d19fed` = 102.2 > max of `a1663771` = 77.0.

Mechanism, from MTP telemetry (means over 6 runs each):

| | accept | tok/cycle | backbone/cycle |
|---|---|---|---|
| `76d19fed` | 72.5% | 2.48 | 18.16 ms |
| `a1663771` | 34.7% | 1.44 | 18.47 ms |

Prefill is unaffected (~2740–2757 tok/s on both sides), which further localises this to the verify path rather than to model loading or attention generally.

Possible starting points in `a1663771` — not yet confirmed as the cause:

- `omlx/patches/mlx_vlm_mtp/qwen35_moe_vlm_runtime.py` and `qwen35_vlm_runtime.py` now pass `speculative_verify=True` into the verify forward.
- The same commit deletes `omlx/patches/mlx_vlm_mtp/qwen35_batch_rollback.py` (58 lines) and drops `qwen35_batch_rollback.apply(...)`, delegating speculative-cache rollback to upstream mlx-vlm.
- `omlx/speculative/vlm_mtp.py` loses 138 lines.

If the new upstream rollback does not restore verifier state identically to the removed oMLX patch, the drafter and verifier would diverge and reject valid drafts — matching the observed signature of unchanged cycle cost with halved acceptance.

shironexus-gif:
fix(qwen35): stop double-shifting the MTP head's pre-fc norms

Fixes #3742

## Root cause

mlx-vlm 0.7.0 only ran `Model.sanitize` when the checkpoint was **not** in MLX
format (`utils.py: if not is_mlx_format`). 0.7.1 sanitizes unconditionally, so
the MTP sanitize patch — written for raw-HF and explicitly-mixed sources — now
also runs on already-converted checkpoints, where its per-key
"mean < 0.5 means raw-HF" test misfires.

`pre_fc_norm_hidden` and `pre_fc_norm_embedding` are the one place that test has
no discriminating power. The head damps the target hidden state and the
next-token embedding before `fc` fuses them, so both gammas sit below 0.5 in
*either* convention (0.4937 and 0.2734 on Qwen3.6-35B-A3B-oQ6). They were read
as raw-HF and shifted a second time, to 1.4937 and 1.2733, while every other
head norm was left correct.

That corrupts the first projection of every draft — which is exactly why draft
acceptance halved while the backbone's latency and prefill were completely
unchanged, matching the symptom reported in #3742.

## Fix

Decide these two keys from the head's per-layer norms, which *are*
magnitude-discriminable, instead of from their own magnitude.

The second commit extracts the rule into `omlx/patches/mtp_head_norm_convention.py`
and calls it from all four copies of the heuristic (`qwen35_moe_vlm_runtime`,
`qwen35_moe_vlm_model`, `qwen35_vlm_model`, `mlx_lm_mtp/qwen35_model`) so it
cannot drift between them again. The dense copy matters most: dense Qwen3.5/3.6
has no runtime override, so `qwen35_vlm_model` is its effective path.

The rule was sharpened while being extracted — the sign settles the unambiguous
cases and only the 0..0.5 band consults the head's per-layer norms:

```
mean < 0     raw-HF   (a converted gamma is raw + 1, and raw pre-fc
                       gammas measure about -0.8..-0.2)
mean >= 0.5  already converted, as the legacy cutoff says
otherwise    ambiguous -> follow the head's per-layer norms
```

A tie counts as converted, since shifting is the destructive direction. The
per-key heuristic is untouched for every other key, so JANG-style mixed bundles
still classify as before, and `test_mlx_lm_mtp_patch`'s mixed-convention
contract stays intact (its pre-fc gammas are negative, so they still shift even
though that checkpoint's per-layer norms are split 2-2).

## Results

Qwen3.6-35B-A3B-oQ6-mtp, 16k prompt, 128 tokens, n=6:

| build | TG tok/s | accept | tok/cycle |
|---|---|---|---|
| `76d19fed` (pre-regression) | 116.7 | 72.5% | 2.48 |
| `a1663771` (regressed) | 70.0 | 34.7% | 1.43 |
| **this PR** | **103.7** | **71.2%** | **2.33** |

Acceptance is restored to the pre-regression level.

Checkpoint sweep, runtime head norms vs their checkpoints:

```
Qwen3.6-35B-A3B-oQ6-mtp    0.4937 / 0.2734  ambiguous band, now kept
Qwen3.6-35B-A3B-oQ4e-mtp   0.4937 / 0.2734  same, and converges to oQ6
Qwen3.8-27B-oQ4e-mtp       0.8428 / 0.5393  above cutoff, unchanged
```

## Remaining gap (separate issue, not addressed here)

TG lands at 103.7 vs 116.7 pre-regression, i.e. **~11% short**. That residual is
a *second, independent* regression from the same mlx-vlm upgrade, and it is a
genuine per-cycle speed cost rather than an acceptance problem.

Measured separately at pinned draft depth with ABAB-alternated builds on an idle
machine (n=6 each, 16k ctx): **20.67 ± 0.18 ms/cycle on mlx-vlm 0.6.3 vs
22.85 ± 0.08 on 0.7.1 — +2.19 ms, +10.6%**, which matches the 11% TG shortfall
from the other direction.

I was not able to localize it. Excluded by direct test: the commit path
(`commit_speculation` reduces to the same two slices for batch-1), the
`mx.contiguous` window copy, `_RotatingCacheTransaction` (unreachable — this
model makes only `ArraysCache`/`KVCache`), `record_speculative_window`'s
concatenate (0 concatenates in 2000 calls), `update_recurrent` requesting state
history on decode, loss of a specialized verify kernel
(`_gated_delta_update_verify_decode` is a thin wrapper around the same
function), the verify forward itself, and custom Metal kernel fallback (all
packs FAST). Happy to open a separate issue with the full detail if useful.

## Tests

`tests/test_qwen35_mtp_head_norm_convention.py` — 10 tests, all passing.


jundot:
I reproduced the double shift you found: on Qwen3.6-35B-A3B the converted pre-fc gammas sit at +0.49 / +0.27, and the per-tensor mean check re-shifted them once mlx-vlm 0.7.1 started sanitizing MLX checkpoints. The diagnosis in this PR is right.

I went a different way for the fix though. I surveyed the MTP heads shipped on Hugging Face for Qwen3.6-35B, Qwen3.8-27B and Flash-Next, and the mixed heads out there come in several patterns (3/7, 6/7, 7/7 shifted), so a majority vote over sibling norms is not a reliable classifier either. #3788 drops the per-tensor guess entirely: raw-HF shifts every gamma, MLX-format is loaded as stored, and the existing norm_repair keeps covering the legacy mixed heads against the backbone. On the same oQ6 checkpoint draft acceptance goes from 33-62% back to 65-92%, so it recovers what this PR recovers.

I'll close this one once #3788 lands. The remaining per-cycle cost you measured on the verify forward looks like a separate regression from the pin bump and I'm looking at it on its own.

Issue #3748, opened 2026-09-18
0.7.0.dev4: Qwen3.8-Flash-Next (qwen4_exp) fails to load — "module 'custom_model' has no attribute 'ModelConfig'" (works on 0.7.0.dev2)


pierou:
0.7.0.dev4: Qwen3.8-Flash-Next (qwen4_exp) fails to load — "module 'custom_model' has no attribute 'ModelConfig'" (works on 0.7.0.dev2)

## Description

Upgrading from 0.7.0.dev2 to 0.7.0.dev4 breaks loading of
`pipenetwork/Qwen3.8-Flash-Next-MLX-mixed-4_8bit` (qwen4_exp architecture, 103.9 GB on disk,
PLE/mmap mode). The VLM engine load crashes inside mlx-vlm's `load_model`, the LLM
fallback is rejected by the custom-module (trust-remote-code) gate, and the model
stays unloadable: every chat completion returns 409. The same model, same settings,
same machine loads and runs fine on 0.7.0.dev2 (~20 tok/s decode).

The model directory ships its own `qwen4_exp.py` custom model module (no
`ModelConfig` class). oMLX's `omlx.patches.mlx_vlm_qwen4_exp_compat` patch **is
applied before the failure** (visible in the log), then mlx-vlm's generic loader
still tries `model_class.ModelConfig.from_dict(config)` on that module and crashes.

## To reproduce

1. Run oMLX 0.7.0.dev4 (mlx-vlm 0.7.1 @ 3fb24e9, mlx-lm 0.32.0 @ 872ae88,
   transformers 5.17.0) on macOS.
2. Load `pipenetwork/Qwen3.8-Flash-Next-MLX-mixed-4_8bit` (any qwen4_exp model with a
   custom module in the model dir should hit the same path).
3. Send any chat completion request.

## What happens

```
21:46:25,366 - omlx.patches.mlx_vlm_qwen4_exp_compat - INFO - Qwen4-Exp mlx-vlm compatibility patch applied
21:46:25,368 - omlx.patches.mlx_vlm_qwen4_exp_compat - INFO - Qwen4-Exp PLE mode for ...: mmap
21:46:25,430 - omlx.engine_pool - WARNING - VLM loading failed for Qwen3.8-Flash-Next-MLX-mixed-4_8bit,
             falling back to LLM: module 'custom_model' has no attribute 'ModelConfig'
[...]

PR #3752, opened 2026-09-19
fix(qwen4_exp): resolve model_file checkpoints to the vendored module

pierou:
# fix(qwen4_exp): resolve model_file checkpoints to the vendored module

## Problem

Loading any published qwen4_exp checkpoint that ships its custom module
(e.g. `pipenetwork/Qwen3.8-Flash-Next-MLX-mixed-4_8bit`) fails on
0.7.0.dev4:

```
VLM loading failed ... falling back to LLM: module 'custom_model' has no attribute 'ModelConfig'
RuntimeError: VLM load failed: ...; LLM fallback also failed: ... custom module
  ('qwen4_exp.py') ... This is disabled by default. Enable Trust Remote Code ...
→ 409 on /v1/chat/completions
```

## Root cause

mlx-vlm 0.7.x changed `get_model_and_args()`: when `config.json` declares
`model_file` **and** `model_path` is passed, it imports that file as
`custom_model` and **bypasses the built-in/vendored registry** — and the new
`load_model` then requires `custom_model.ModelConfig`. Published qwen4_exp
checkpoints ship `model_file=qwen4_exp.py` exporting only `ModelArgs`-style
classes (mlx-lm custom-model style), so the load always dies. mlx-vlm 0.6.x
had no `model_file` resolution — `model_type: qwen4_exp` resolved to
`mlx_vlm.models.qwen4_exp`, which is exactly the module this compat patch
vendors and exposes ("Expose `mlx_vlm.models.qwen4_exp` from oMLX's vendor
tree"). 0.7.0.dev2 (mlx-vlm 0.6.3) loads these checkpoints fine.

## Change

In `apply_mlx_vlm_qwen4_exp_compat_patch()`, wrap
`mlx_vlm.utils.get_model_and_args` so that for `model_type == "qwen4_exp"`
the `model_file` key is dropped and resolution continues through the
registry (passing `model_path=None`), restoring 0.6.x routing. Only
`load_model` passes `model_path`, so no other custom-module loading path
changes behavior. Adds a regression test mirroring the
`test_glm5_next_registers_pinned_upstream_model` resolution pattern.

## Testing

- **Regression test red/green:** `test_qwen4_exp_model_file_checkpoints_resolve_to_vendored_module`
  fails on pristine `origin/main` (resolves to `custom_model`, no
  `ModelConfig`) and passes with the patch.
- **Full affected file:** `pytest tests/test_mlx_vlm_qwen4_exp_compat.py` —
  67 passed, 2 failed; the two `test_qwen4_small_hyper_connection_fusion_fails_closed`
  failures reproduce identically on pristine `main` on this machine
  (local-kernel-build related), unrelated to this change.
- **Style:** `black --check` and `ruff check` flag only pre-existing upstream
  lines in these two files (blank-line / `E741` / `I001` artifacts); no
  finding lands inside this diff, and those lines were left untouched per
  "keep unrelated formatting out of the diff".
- **Live:** `pipenetwork/Qwen3.8-Flash-Next-MLX-mixed-4_8bit` (69.6 GB
  weights, PLE forced to SSD/mmap) on an Apple M2 Max / 96 GB, macOS 26.6.2,
  0.7.0.dev4 + this patch — model previously 409'd on every request, now
  loads (VLM engine) and decodes ~23.5 tok/s single-request (0.7.0.dev2 on
  the same machine: ~20).
- **Not run:** the full default suite (was not run on this host); no vision /
  tool-call / concurrency feature exercise beyond normal serving of this
  model since loading.

Originally reported in #3748 (log excerpts; full logs offered to the
maintainer).

---
*AI-assisted: root-caused and patched with the help of a coding agent
against the pinned mlx-vlm checkouts and live server logs; tested by the
author on real checkpoints.*


jundot (approved):
Thanks for finding and fixing this issue. LGTM!

PR #3757, opened 2026-09-19
feat(responses): complete the envelope; accept unexposable tool declarations with a warning

LXD-8:
Based on `main` @ `14194fe7` (v0.7.0.dev4).

## Summary

`/v1/responses` separates *declaring* a capability from *using* it.

A tool declaration the model cannot use does not fail the request. Hosted and
unrecognised tool types (`web_search`, `web_search_preview`, `file_search`,
`mcp`, `computer_use_preview`, `computer`, `code_interpreter`,
`image_generation`, `local_shell`, `custom`, a nested namespace, or any other
`type` the endpoint does not know) are accepted, left out of what the model
sees, and named in a `Warning` response header. An unused declaration cannot
change the response, so accepting it is safe; naming it keeps the degradation
visible instead of silent.

This is what lets a real Codex CLI 0.154 session run: Codex declares the hosted
`web_search` tool in every session, so refusing that declaration failed the
client the endpoint was built for before it could do anything.

Everything the endpoint cannot honour without changing the meaning of the
response still fails with a 400 naming the field or type — in particular the
input item types that replay a hosted round trip oMLX never performed.

## Why a declaration is accepted rather than refused

Codex CLI 0.154.0-alpha.6.2 declares the hosted `web_search` tool
unconditionally, so refusing an unsupported tool declaration would fail every
session before it could do anything. Declaration is therefore treated as
separate from use: a declaration oMLX cannot execute is accepted and reported
rather than refused, while an input item that replays a hosted round trip oMLX
never performed is still a 400.

## What changed

- **Tool declarations are classified, not refused.** `convert_responses_tools()`
  (`omlx/api/responses_utils.py`) expands `namespace` groups into callable
  function tools (recorded in the binding registry so a namespace call
  round-trips, #3371) and accepts `tool_search` as a no-op because every
  namespace member is exposed eagerly. A declaration it cannot expose — a known
  hosted type, an unrecognised type, a nested namespace, or a non-function /
  hosted member of a namespace — is left out of the returned tools and appended
  to the new `unexposed` out-param as a header-safe label:
  `web_search (hosted)`, `some_future_tool (unknown type)`,
  `namespace (nested) in namespace mcp__demo__`. Malformed declarations still
  raise: a function without a `name`, a namespace without a `name`, a namespace
  member that is not a tool object, a function member without a `name`.
- **The degradation is visible on both paths.** `create_response`
  (`omlx/server.py`) builds an RFC 7234 `Warning` header with
  `_unexposed_tools_warning_header()`:
  `199 omlx "tools accepted but not exposed to the model: web_search (hosted); oMLX cannot execute hosted tools and did not tell the model about them"`.
  It is attached to the non-streaming JSON response
  (`_json_response_or_keepalive(..., headers=...)`) and to the streaming SSE
  response (`sse_headers["Warning"]`), so neither path is silent. Labels are
  constrained to header-safe characters (a client-supplied `type` cannot inject
  a header) and capped in length.
- **Hosted execution is never fabricated.** A hosted declaration is dropped; it
  is never mapped onto a function tool and the model is never told it can call
  it.
- **Request-level validation.** `validate_responses_request()` refuses
  `tool_choice: "required"`, a named-function `tool_choice`, `truncation:
  "auto"` and invalid truncation values, `parallel_tool_calls: false`,
  `background`, `conversation`, `max_tool_calls`, `top_logprobs`, unsupported
  `include` values, and malformed `text.format`. `tool_choice: "auto" | "none"`,
  `truncation: "disabled"` and `parallel_tool_calls: true` (the model's natural
  behaviour) are accepted.
- **Input item types are validated.** `computer_call`, `web_search_call`,
  `file_search_call`, `mcp_call`, `mcp_list_tools`, `mcp_approval_request`,
  `mcp_approval_response`, `image_generation_call`, `code_interpreter_call`,
  `local_shell_call`, `custom_tool_call(_output)` and `item_reference` are
  rejected with `param: "input"`, as are unknown item types, unknown message
  roles, and content parts other than `input_text`/`text`/`output_text`/
  `input_image` (e.g. `input_file`, `input_audio`).
- **One canonical tool identity.** New `omlx/api/tool_bindings.py` holds
  `ToolBinding`, `ToolBindingRegistry` and `ensure_call_id()`. Both Responses
  emission sites and the namespace history helpers resolve wire names through
  the same registry, which also survives a collision rename between turns. The
  deliberate wire dialects of the other paths are untouched: Chat keeps
  `namespace::name` and MCP keeps `server__tool`.
- **`call_id` lifecycle.** All fallbacks go through `ensure_call_id()`. A
  `function_call_output` that omits `call_id` now pairs with the oldest
  `function_call` still waiting for its output, instead of generating an
  unrelated id.
- **Reasoning round-trip.** A reasoning input item is read from `summary` and,
  when that is empty, from `content[].reasoning_text`, so either Responses
  dialect round-trips. Stored responses containing an output item type the
  server cannot replay raise a corrupt-state 409 rather than silently dropping a
  turn.
- **Streaming / non-streaming parity.** `response.created` and
  `response.in_progress` are built by `build_response_object(status=
  "in_progress")`, the same builder as the terminal event; the streamed envelope
  echoes the resolved `temperature`/`top_p` the non-streaming body reports; and
  the streamed reasoning `content` part carries the same `annotations` field as
  the non-streamed one.
- **`truncation` is echoed.** The request's mode (`"disabled"` when omitted)
  round-trips on the response instead of serialising as `null`.
- **Structured output is enforced or refused.** When `text.format` requests
  `json_object` or `json_schema` and no grammar can be compiled for the model,
  the request is rejected with `param: "text.format"`. This is a deliberate
  difference from `/v1/chat/completions`, which warns and degrades to prompt
  injection: `text.format` is an output contract, and returning text that need
  not satisfy it is worse than a clear error.
- **Test contract.** Tests that asserted a 400 for a hosted/unsupported tool
  declaration are updated to the new contract (declaration accepted, named in
  the warning, not exposed): `test_unsupported_tool_types_are_accepted_but_not_exposed`,
  `test_supported_and_unsupported_tools_mixed_keeps_supported`,
  `test_namespace_with_hosted_member_is_dropped_and_reported`,
  `test_nested_namespace_is_dropped_and_reported`, the `local_shell` assertion
  in `test_request_from_codex_like_json`, and
  `test_responses_unsupported_tool_declarations_are_accepted_with_warning`.
  `test_responses_unsupported_input_items_fail_loudly` still covers the
  rejected input-item case, and new tests cover the Codex payload (functions +
  namespace + `web_search`) and Warning-header injection safety.

## Compatibility

Exercised against a patched server on `127.0.0.1:8011` (a second, isolated
instance serving a lightweight 1B model; the user's server on :8000 was
untouched):

- **OpenAI Python SDK 3.16.2** — `responses.with_raw_response.create()` with a
  `web_search` plus a function tool returned HTTP 200 and
  `Warning: 199 omlx "… web_search (hosted) …"`; `responses.stream()` returned
  200 with the same header and `response.completed`. A `web_search_call` input
  item still returned 400; a `local_shell` declaration returned 200 with the
  same warning.
- **`@ai-sdk/open-responses` 2.0.49** (with `ai` 7.0.107) — `generateText` and
  `streamText` whose request body carried `["function", "web_search"]` returned
  200 with the warning on both paths, while the function tool was still exposed
  to the model. The provider exposes no public API for a bare hosted-tool
  declaration, so the declaration was injected through its `fetch` middleware;
  everything else is the provider's real request path.
- **Codex CLI 0.154.0-alpha.6.2** (bundled at
  `/Applications/ChatGPT.app/Contents/Resources/codex`) — a throwaway
  `CODEX_HOME` with a custom `model_provider` using `wire_api = "responses"` and
  `base_url = "http://127.0.0.1:8011/v1"` directly, with **no proxy**. A
  `codex exec` session completed with exit code 0 and returned text. Its
  declared tools were functions `exec_command`, `write_stdin`,
  `request_user_input`, `view_image` and four others, one `namespace`
  (`multi_agent_v1`), and the hosted `web_search`; it does not send
  `local_shell`. The declared `web_search` was accepted and named in the
  response's `Warning` header. A full function-tool *execution* round trip was
  not achieved in this run because the lightweight 1B model emitted
  `<|python_tag|>exec_command(...)` as text rather than a `function_call` item;
  the big-model round trip through a stripping proxy is what exercised that
  path.
- **Not exercised:** OpenCode is not installed on this machine and was not run.
- Cherry Studio's Responses path is the related downstream integration for the
  same implementation: https://github.com/CherryHQ/cherry-studio/pull/20683
  (same author). This change was not exercised through Cherry Studio here, and
  Cherry Studio does not officially ship an oMLX provider.

## Tests

- `tests/test_responses.py`, `tests/test_server.py`,
  `tests/integration/test_e2e_streaming.py`,
  `tests/integration/test_server_endpoints.py`,
  `tests/integration/test_responses_tool_merge.py` with `-m ""` (so the
  integration tests actually execute): **526 passed**.
- New coverage includes a `semantic_equal`-style canonicaliser proving the
  aggregated stream and the non-streaming body match for text, reasoning and
  tool-call turns; streaming event order, item ids, output/content indexes and
  delta/done pairing; `call_id` pairing and stability across the store; the
  reasoning → tool → result sequence via `previous_response_id`; the accepted-
  but-not-exposed contract and its `Warning` header on both response paths; the
  Codex tool list; and header-injection safety of the warning labels.

## Limitations

Accepted but never exposed to the model (each named in the `Warning` header):

- Hosted/server-executed tools: `mcp`, `web_search`/`web_search_preview`,
  `file_search`, `computer_use_preview`/`computer`, `code_interpreter`,
  `image_generation`, `local_shell`, `custom`.
- Any unrecognised tool `type`.
- A nested namespace, and any non-function member of a namespace.
- `tool_search` is accepted as a documented no-op without a Warning: every
  namespace member is exposed eagerly, so the capability it requests is already
  present and its absence costs the model nothing.

Rejected with a 400 (cannot be honoured without changing the response):

- Input items for a hosted round trip oMLX never performed:
  `computer_call(_output)`, `web_search_call`, `file_search_call`, `mcp_call`,
  `mcp_list_tools`, `mcp_approval_request/response`, `image_generation_call`,
  `code_interpreter_call`, `local_shell_call`, `custom_tool_call(_output)`,
  `item_reference`.
- `tool_choice: "required"` and named functions: no chat template oMLX serves
  can guarantee that the model emits a call.
- `parallel_tool_calls: false`: cannot guarantee a single call, and truncating
  the model's extra calls would silently discard tool calls. `true` is accepted
  as the model's natural behaviour.
- `truncation: "auto"`: context-window trimming that protects instructions,
  active tool calls and reasoning/tool-call pairing is not implemented;
  `"disabled"` is the only supported mode, and an over-long prompt is a 400.
- Structured output when no grammar can be compiled (see above).
- `background`, `conversation`, `max_tool_calls`, `top_logprobs`, unsupported
  `include` values, and malformed tool declarations (function/namespace without
  a name, a namespace member that is not a tool object).

Accepted as documented no-ops, because they do not change which tools the
model sees or what the conversation means:

- `include: ["reasoning.encrypted_content"]` (reasoning is returned in the
  item itself), `tool_search`, and the performance/attribution hints
  `service_tier`, `prompt_cache_key`, `prompt_cache_retention` and `user`.

Pre-existing behaviour this change does not touch:

- Pathological reasoning models: when the rendered prompt opens a thinking block
  and the model never emits a closing `</think>`, the shared streaming
  `ThinkingParser` recovery re-emits the reasoning as content so the answer body
  is not empty. That is pre-existing Chat/Responses behaviour and is unchanged
  here; it can make the reasoning and message text identical for such a model.


LXD-8:
@jundot — would you take a look at this one when you have a moment?

`/v1/responses` accepted request capabilities it could not honour and answered as if it had: a tool whose `type` was not `function` was skipped before the chat template was built, `tool_choice: "required"` and `parallel_tool_calls: false` were echoed back but never applied, and a client-omitted `call_id` was filled by two independent generators so a `function_call_output` could not pair with its `function_call`. Every unimplemented capability now either works or fails with a 400 naming the field, and tool declarations oMLX cannot execute are accepted but not exposed to the model and named in a `Warning` header — which is what lets a real Codex CLI session run against it. CI is green on 3.11 / 3.12 / 3.13.

LXD-8:
Superseded by 3 focused PRs for easier review. See #3758, #3759, #3760.

PR #3760, opened 2026-09-19
fix: restore Qwen VLM decode speed after mlx-vlm upgrade

### LINKED ISSUE #3755: Decode throughput −7–11% on 0.7.0.dev4 vs 0.7.0.dev2 — Qwen3.8-Flash-Next-oQ4e-mtp (MTP off, SSD offload on, M3 Ultra 96 GB)
Summary

On an M3 Ultra (96 GB), Qwen3.8-Flash-Next-oQ4e-mtp decodes ~7–11% slower on v0.7.0.dev4 than on v0.7.0.dev2, reproducible across four independent dev4 sessions. Prefill is ~3–7% faster on dev4, memory is flat, and no functional/correctness bugs are observed. This looks like a decode-path side effect of the core library upgrade in dev4 rather than a config mistake on my side — filing because the dev4 release notes explicitly request reports of regressions on previously working models.

Environment

Apple M3 Ultra, 96 GB unified memory, current macOS release at time of testing
Model: Jundot/Qwen3.8-Flash-Next-oQ4e-mtp (oQ4e, FP16-retained variant naming as published)
N-gram SSD offload ON, oMLX SSD prefix cache ON 92 GB cap, hot RAM cache OFF). These settings are identical in both dev2 and dev4 captures — offload is a constant, not a variable.
MTP: OFF in every sample. Lightning MTP, vLM MTP, DFlash, SpecPrefill, ANE/Qwen prefill: all OFF.
Context ceiling: 139,264. Thinking: enabled, no budget cap. Sampling: temperature 1.0, top_p 0.95, top_k 20, min_p 0.0, presence_penalty 0.0, repetition_penalty 1.0.
Single resident model, serial requests, 4-min cooldown between samples, idle confirmed before each run.

Protocol

Harness: internal profile benchmark (omlx-profile@1), one HTTP chat-completions endpoint.
Ladder: ~4K / 16K / 48–64K / ~128K actual tokenizer prompt tokens × cold/warm × repetitions. (1/3)
Per-request telemetry: exact prompt/completion token counts, completion_tokens / (duration − TTFT) as decode tok/s, peak memory via sampler.
Engine verified live at each capture time via the admin API (version + per-model settings), not inferred from what was installed.
Each dev4 sample is valid only if MTP OFF, cache state, and token counts are confirmed from the server's own counters.

Results (decode tok/s, medians; ranges where multiple sessions)

4K cold — dev1: 48.2–48.4 | dev4: 47.0–50.1 | ≈ neutral
16K cold — dev1: 49.2 (48.7–49.5) | dev4: 45.6 (42.8–47.6) | −7.4%
16K warm — dev1: 47.2 | dev4: 47.0 | ≈ neutral
64K cold — dev1: ~46.7 | dev4: ~46.0 | ≈ neutral / −1%
~128K cold — dev1: 41.3 (40.4–41.6) | dev4: 37.6 (37.1–38.0) | −9.0%
~128K warm — dev1: ~41.0 | dev4: 36.5 | −10.9%

Prefill (same matched cold cells): dev4 +3–7% faster. Peak MLX memory: flat (~68 GB model residency both arms).

Notes:
All ~128K "cold" samples in both arms show identical cached_tokens (~73,728 persistent-prefix reuse); the arms are equally contaminated, so the comparison is fair — flagged for honesty.
All dev4 samples fall in the same low band across four sessions; dev1 distributions barely overlap dev4's at 16K cold and 128K.

jundot:
Fixes #3755

## Problem

Since a1663771 (#3719), Qwen3.8-Flash-Next-oQ4e-mtp decodes 5-10% slower than 0.7.0.dev2 with MTP off (about 2 ms per token, independent of context length). Qwen3.5/3.6 VLM checkpoints lose a smaller 2-3% on the same path.

## Cause

That commit switched the q4 prefill routing from wrapping mlx-vlm's `_target_verify_linears` (which the upgraded mlx-vlm no longer routes projections through) to reclassing the loaded projections to `_VLMQuantizedPrefillLinear`. Two side effects:

- `_qwen4_decode_static_eligible` in `qwen35_gdn_prework.py` checked `type(linear) is nn.QuantizedLinear`, so the reclassed projections failed the gate and the fused B1/T1 GDN decode prework and norm-gate kernel never engaged. All 36 GDN layers ran the unfused decode path. The server log no longer showed "Qwen4 fused B1/T1 GDN decode prework and norm-gate engaged".
- `_VLMQuantizedPrefillLinear.__call__` read four environment variables on every projection call, including single-token decode where routing is never taken (about 230 calls and 470 environment lookups per token on Flash-Next).

## Change

- Accept `nn.QuantizedLinear` subclasses in the fused decode gate. The fused path reads the packed weight, scales and biases directly, and `out_proj` falls back to the stock forward at T=1, so numerics are unchanged.
- Read the routing thresholds once in `apply_qwen35_q4_prefill_linear_patch` and keep them as class attributes.
- Regression test: a canonical Qwen4 GDN module whose projections are reclassed to `_VLMQuantizedPrefillLinear` must stay eligible. It fails on main and passes with this change.

## Validation

M3 Ultra, plain decode (MTP off), same prompts, two runs per cell, tok/s:

| model / prompt | main (dev4) | this PR | 0.7.0.dev2 |
|---|---|---|---|
| Qwen3.8-Flash-Next-oQ4e-mtp, 4K | 45.8 / 48.1 | 51.8 / 51.7 | 51.9 / 52.4 |
| Qwen3.8-Flash-Next-oQ4e-mtp, 16K | 42.6 / 45.0 | 48.2 / 48.5 | 49.1 / 49.3 |
| Qwen3.6-35B-A3B-oQ6-mtp, 4K | 87.6 / 87.7 | 88.9 / 89.7 | |
| Qwen3.6-35B-A3B-oQ6-mtp, 16K | 79.6 / 79.5 | 81.0 / 81.5 | |

Bisect (4K/16K, two runs each) placed the regression at a1663771: the parent commit and the mlx-lm-only bump both matched dev2 speed, and the mlx-vlm bump commit matched dev4. Disabling the reclass with `OMLX_QWEN35_Q4_LINEAR=0` on unmodified main recovers the same numbers as this PR, which isolates the cause to the reclass rather than the library itself.

The Qwen3.6-35B numbers also cover part of the residual per-cycle gap reported in #3750; the remaining gap there is not addressed here.


moc375:
Confirming the regression on our end with clean isolation — **Flash-Next + MTP-on shows the deepest hit in the dev4 family** (M3 Ultra 512GB, `Qwen3.8-Flash-Next-oQ4e-mtp`, Lightning MTP depth 5, identical config across arms):

| tree | mlx-vlm | quiet-field warm decode @1K | MTP accept | cycle time |
|---|---|---|---|---|
| dev3 | 0.6.3 | **161-163** (two restarts, 5-run sequences each) | 80-100% | ~15 ms |
| dev4 | 0.7.1 | **84-87** (two restarts, 5-run sequences each) | 86-88% | ~30 ms |

Same machine, same model, same settings — **~50% regression, and it's not an acceptance problem** (accept rates are healthy on dev4). Cycle time roughly doubles, pointing at the verify/backbone path itself.

We also applied this PR's head (`9ad7a4a`) on top of dev4: **69-73 tok/s with our translation co-tenant active** (the co-tenant was not running in the quiet-field rows above, so we cannot yet isolate this fix's effect cleanly — will report back with a quiet-field number). What we can say: **dev4+fix did not recover to dev3 levels in the same window where dev3 gives 167**.

The tree↔venv coupling is tight (dev4 tree requires `mlx_vlm.speculative.cache_state` from 0.7.1; dev3 tree + 0.7.1 venv fails on FP8 PLE shard format), so we couldn't bisect which of the venv changes (mlx-vlm 0.6.3→0.7.1, transformers 5.12→5.17, tokenizers 0.22→0.23) is responsible — sharing in case it helps narrow the search.

moc375:
Follow-up as promised — **quiet-field number for dev4 + this PR** (`9ad7a4a`, translation co-tenant stopped, fresh restart, 120 s settle, 6-run sequence):

**84-87 tok/s** (76.9 / 84.2 / 84.3 / 87.6 / 84.4 / 87.0, median 84.4) — MTP accept 86-88%, tok/cycle 2.5-2.6.

So the quiet-field picture across all three arms is now:

| tree | mlx-vlm | quiet-field @1K (median) |
|---|---|---|
| dev3 | 0.6.3 | **~162** |
| dev4 bare | 0.7.1 | ~85 |
| dev4 + this PR | 0.7.1 | **~84** |

The fix restores the per-projection wrapper overhead for the Qwen3.5/3.6 family, but the **Flash-Next + MTP-on regression (~50%) persists untouched** — cycle time stays ~30 ms vs dev3's ~15 ms with healthy acceptance on all arms, so the remaining gap is elsewhere in the verify/backbone path for `qwen4_exp`.

PR #3764, opened 2026-09-19
fix(admin): only unload models on settings save when a cache value actually changed

zviratko:
## Problem

`POST /admin/api/global-settings` unloads all loaded models on every settings save, even when nothing cache-related changed. In-flight requests hang and the dashboard shows no models until they reload.

## Root cause

The cache block in `update_global_settings()` sets `cache_changed = True` for every cache field that is `is not None`. The WebUI (and any API client copying its behaviour) sends the full settings payload on every save, so every field always passes the `is not None` test. `_apply_cache_settings_runtime()` then runs and unloads every loaded engine by design.

## Fix

Compare each incoming value against the current live value. Only a real difference sets `cache_changed = True`. A genuine cache change still unloads engines, which is required and unchanged.

## Localization

No user-visible strings changed. One file: `omlx/admin/routes.py`.

## Review notes

- Related: #2637 diagnosed the same root cause but was closed as a draft unmerged; this PR covers only the unload-triggering cache block and is intentionally minimal.
- The `gdn_snapshot_storage` comparison uses `get_gdn_snapshot_storage()` so the legacy alias fields compare consistently.

## Tests

- `python -m py_compile omlx/admin/routes.py` clean on Python 3.11.
- Applied to a live omlx install and tested under real agentic inference workload: saving settings with unrelated changes no longer unloads models.

## Human notes

tested under real agentic workload inference but I can't rule out the original behaviour being intentional


jundot (approved):
Comparing cache values before rebuilding the engines fixes the repeated unload for unchanged settings. I verified that identical cache payloads skip unloading while actual changes still trigger it. I'll handle default path/size round-tripping and GDN alias normalization in a small follow-up, since those existing cases still trigger unnecessary unloads. LGTM.

PR #3765, opened 2026-09-19
feat(admin): live-apply max_concurrent_requests without unloading models

zviratko:
## Problem

`max_concurrent_requests` is labelled "restart required" in the settings UI, but the value is already read live by the scheduler: `Scheduler._effective_max_num_seqs()`, the waiting-queue bound (`max(max_num_seqs * 4, 32)`) and the pressure-gate regrowth all read `config.max_num_seqs` on every admission decision. Nothing needs a model reload for a new cap to be honoured.

## Fix

On save, when the value actually changed:
- update the pool template `pool._scheduler_config.max_num_seqs` (engines loaded later pick it up),
- update each loaded engine's `scheduler.config.max_num_seqs` (each scheduler holds a `copy.copy` of the config, same pattern as the neighbouring `chunked_prefill` live handler),
- reject values <= 0 with 400,
- report the key in `runtime_applied`.

The `_StoreCacheGate` is intentionally left alone: its cap is pressure-shrunk by `ProcessMemoryEnforcer` and regrows toward the configured max by itself; forcing it back up would undo an active shrink.

Semantics: raising the cap takes effect on the next admission tick. Lowering it stops admitting new requests until in-flight ones drain — no aborts, no engine churn.

Removes the now-wrong RESTART badge from the Max Concurrent Requests row in the settings template.

## Localization

Removed one `restart_badge` usage from `dashboard/_settings.html`; no i18n key removed (the key stays used by sibling rows).

## Review notes

- Complements #3764 (different handler in the same route; the two patches are independent).
- One file of behaviour (`omlx/admin/routes.py`) plus one template line removal.

## Tests

- `python -m py_compile omlx/admin/routes.py` clean on Python 3.11.
- Applied to a live omlx install and tested under real agentic inference workload: changing the cap mid-flight keeps streams running, new admissions follow the new cap, no model reload.

## Human notes

tested under real agentic workload inference but I can't rule out the original behaviour being intentional


jundot:
I reproduced a 1-to-4 update that reports success but leaves `completion_batch_size` at 1 in the pool, scheduler config, and existing BatchGenerator, so decoding remains serial. Please update the decode batch limit alongside admission, preserving in-flight requests when lowering it. The DFlash fallback scheduler and its copied config also retain the old limit because this handler only follows `_engine.engine.scheduler`. I'd like these paths covered before merging.

zviratko:
Thanks for reproducing this — you were right on both paths. Updated in bf3e9c1d (current `main` merged in, no conflicts):

- **Decode batch limit:** `completion_batch_size` is now updated together with `max_num_seqs` in the pool template, each scheduler's config copy and the existing `BatchGenerator` (floored at `prefill_batch_size`, like its constructor). A 1 → 4 update now reaches all three.
- **Lowering:** the generator only stops admitting new rows when the cap drops; `_next()` does not evict rows that are already batched, and nothing is aborted. In-flight requests drain normally.
- **DFlash:** the handler now also updates DFlash's copied scheduler config (so a fallback that starts later gets the new value) and follows `_fallback_engine` to a fallback scheduler that is already running.
- **Coverage:** `tests/test_admin_max_concurrent_live_apply.py` covers these paths, including your 1 → 4 case.

Two more changes came out of our own review of this fix:

- The `BatchGenerator` write is posted to the engine's MLX executor instead of being made from the admin thread. The MTP batch-generator wrapper saves, zeroes and restores `completion_batch_size` around a step, so a direct write could be lost to the restore (or briefly undo MTP's drain). Queued on the executor, it runs between steps.
- Live application now happens only after `validate()` and `save()` succeed, so a failed save cannot leave runtime limits different from the saved settings.

Limitations, also in the PR description:

- Distributed (cluster) engines are not live-applied: the rank processes own their schedulers and are not reachable from this handler. They are skipped and pick up the new limit on reload; the settings row shows the existing "Restart needed" badge while distributed inference is active.
- The tests are assignment-level (no real decoding), and this revision has not yet been re-tested under live workload — only the earlier admission-only version was.

Made under human guidelines with free tokens from local oMLX inference server, final check and polish by gpt-6-astra and opus-5.5

PR #3773, opened 2026-09-20
feat(responses): complete the response envelope and reasoning channels

LXD-8:
Base `jundot:main @ 8288884d` (v0.7.0.dev4) · head `LXD-8:responses-envelope @ 7855a645`.

## Why

`/v1/responses` was built for Codex CLI compatibility and stopped there, so a client using it as a general Responses surface hit two problems.

Reasoning was published only the way OpenAI's own hosts do: `item.summary` plus `response.reasoning_summary_text.delta`. Clients that follow the Responses dialect read `item.content[].text` and `response.reasoning_text.delta`, so they saw the reasoning tokens counted in usage and nothing in the output.

The response object was built twice by hand, once per path, and the two had drifted. Neither the streaming terminal event nor the non-streaming body carried the echo fields `instructions`, `store`, `parallel_tool_calls` or `reasoning`. The streaming path also echoed `request.temperature` / `request.top_p`, which are `null` whenever the client omits them, while the non-streaming body echoed the values `get_sampling_params` had resolved -- so the same request described itself differently on the two paths.

## What

- **One envelope builder.** `build_response_object()` is the only place either path assembles the response, including the streaming `response.created` / `response.in_progress` snapshot (through its new `status=` override) and the `response.completed` / `response.incomplete` terminal event. Before this, the snapshot was a hand-built `ResponseObject` that omitted `instructions`, `store`, `parallel_tool_calls`, `reasoning`, `text`, `metadata` and `truncation`, so a client latching any of them from the first event saw a different response object than the terminal event carries.
- **Reasoning on the raw channel, and only that one.** `build_reasoning_output_item()` fills `content[]` with a `reasoning_text` part, and the stream emits `response.reasoning_text.delta` / `.done`. `summary[]` is left empty and no `response.reasoning_summary_text.*` event is sent.

  Two channels carrying the same text is not additive for clients that read both. I captured the SSE the endpoint actually emits at each revision (through the test client in `tests/integration/`) and replayed those bytes to each SDK's own parser:

  | SSE the server emitted | `@ai-sdk/open-responses@1.0.34` | `@ai-sdk/open-responses@2.0.49` | `@ai-sdk/openai@3.0.109` |
  |---|---|---|---|
  | `main` — summary only | **0 reasoning deltas: the reasoning is dropped** | 1 | 1 |
  | an earlier revision of this branch — both channels | 1 | **2, the text twice** | **2, the text twice** |
  | this revision (`7855a645`) — raw channel only | 1 | 1 | 1 |

  1.0.34 handles `response.reasoning_text.delta` and no other reasoning event; 2.0.49 added the summary channel and reads both. oMLX has no summarizer, so `summary[]` is left empty and the raw channel — the one the dialect puts the CoT in — is the only one published. A test pins the second half of that on the server side: no `response.reasoning_summary_text.*` event is emitted at all.
- **An omitted `call_id` no longer reaches the template empty.** The `function_call_output` side used `item.call_id or ""` while `function_call` generated a fallback, so the template could be handed an empty `tool_call_id`. Both sides now go through one helper, so the id is never blank. Repairing the *pairing* is not attempted here: the schema requires `call_id` on both items and a request that omits it is refused with a 400 by the input validation on the branch that adds it (#3774).
- **Removed a helper with no production caller.** `convert_stored_response_to_messages` is deleted, with the four tests that exercised it (`test_empty_output`, `test_function_call_output`, `test_message_output`, `test_state_record_prefers_output_messages`).

## Behaviour delta

- New SSE event types: `response.reasoning_text.delta`, `response.reasoning_text.done`.
- New `content[].reasoning_text` channel on reasoning items, alongside the existing `summary[]`.
- `instructions`, `store`, `parallel_tool_calls` and `reasoning` are now echoed by both the streaming terminal event and the non-streaming body.
- `temperature` / `top_p` reported by the streaming envelope are the resolved values, so they match the non-streaming body instead of being `null` when the client omits them.
- `convert_stored_response_to_messages` and the four tests that covered only it are removed — verified orphaned (`git grep` now finds no Python reference at all; the only remaining hit is inside `omlx/admin/bench_corpora/code_python.txt`, which is a text corpus rather than an importable module). Its `TestConvertStoredResponse` sibling tests now call `normalize_response_output_to_messages` directly.
- The streaming terminal still serializes with `exclude_none`, so `null`-valued fields stay omitted there while the non-streaming body keeps them. That predates this change and the streaming integration tests pin it, so the two key sets still differ even though their shared values now agree.
- `response.created` / `response.in_progress` carry the same echoed request fields as the terminal event, and `truncation` now appears on both.

## Verification

- Five-file Responses suite on this commit (`tests/test_responses.py`, `tests/test_server.py`, `tests/integration/test_e2e_streaming.py`, `tests/integration/test_server_endpoints.py`, `tests/integration/test_responses_tool_merge.py`): **418 passed, 12 deselected**. The two streaming files in full: **246 passed, 12 deselected**.
- `main` added three truncated-thinking assertions after this branch was opened (`test_stream_thinking_length_channels[responses-False]`, `[responses-True]`, `test_nonstream_thinking_length_channels[responses]`, #3796) that read `response.reasoning_summary_text.delta` and the reasoning item's `summary[]`. They now read the channel this branch emits — `response.reasoning_text.delta`, and `content[]` parts of type `reasoning_text` — so they keep pinning the same thing (the truncated CoT stays out of answer content and still arrives on the reasoning channel) on the channel that carries it. That selection: **9 passed, 249 deselected**.
- The envelope tests assert the values the two paths serialize, not just the key names: `truncation` is checked as the request's mode rather than for the key's presence, and the earlier `terminal.get("truncation") == body.get("truncation")` line — true of two `None`s — is gone. They fail against the previous revision.
- An endpoint test pins the `text.format` echo: `schema` is present and `schema_` appears nowhere in the body. It fails against the previous revision.
- The streamed reasoning tests assert that no `response.reasoning_summary_text.*` event is emitted at all, so a client that listens to both names cannot double up.
- A generator-level test drives `stream_responses_api` and asserts the opening snapshot echoes every request field and the resolved sampling values, and that the terminal event agrees with it. It fails on the previous revision, where the snapshot was built by hand.

`responses-validation` (#3774) is stacked on this branch and adds request/input validation on top; its own delta is https://github.com/LXD-8/omlx/compare/responses-envelope...responses-validation.


jundot:
I agree that streaming and non-streaming responses should report the same resolved sampling values for the same request. A shared response builder makes sense for that. I'm not yet convinced the other changes address problems in the current implementation.

I tested @ai-sdk/open-responses 2.0.49, and it already reads the existing reasoning summary channel. With this PR, it consumes both delta channels and outputs the same reasoning twice. Could you share a client/version and a reproduction where the current implementation drops the reasoning?

I also found that the new `text.format` echo returns `schema_` instead of `schema`, and the missing `call_id` fallback still generates different IDs for the function call and its output, so it does not repair the pairing.

I'd like to base these changes on concrete evidence of why they are needed, which clients or workflows need them, and when the current behavior causes problems for users.

LXD-8:
Here is the client/version and the reproduction you asked for, and the branch is now reduced to what earns its place.

I captured the SSE the endpoint actually emits at three revisions — `main` and this branch as first opened, through the test client in `tests/integration/` — and replayed those exact bytes to each SDK's own parser:

| SSE the server emitted | `@ai-sdk/open-responses@1.0.34` | `@ai-sdk/open-responses@2.0.49` | `@ai-sdk/openai@3.0.109` |
|---|---|---|---|
| `main` (`d1889871`) — summary channel only | **0 reasoning deltas: the reasoning is dropped** | 1 | 1 |
| this PR as first opened (`6d87dd15`) — the same text on both channels | 1 | **2, the text twice** | **2, the text twice** |
| this revision (`38280832`) — raw channel only | 1 | 1 | 1 |

`@ai-sdk/open-responses@1.0.34` handles `response.reasoning_text.delta` and no other reasoning event, so a summary-only server shows it nothing; 2.0.49 (the version you tested) added the summary channel and reads both, which is exactly why the text arrived twice. Sending one channel satisfies all three, and it is the channel the dialect puts the raw CoT in — so `summary[]` is left empty rather than holding a copy (oMLX has no summarizer), and no `reasoning_summary_*` event is sent at all.

Also from your review: the `text.format` echo carried `schema_`, the Python-side name of an aliased field — `populate_by_name` only affects parsing, so it needed a dump that asks for aliases. Every Responses dump asks now, pinned by an endpoint test that asserts `schema` is present and `schema_` appears nowhere in the body. The `call_id` recovery is gone from this branch, per your comment on #3774.

What is left beyond the shared builder, the resolved sampling values and the `truncation` echo is that one reasoning channel. If you would rather keep a summary channel for hosts that read summaries only, I can gate it on the request's `reasoning.summary` — I have not found such a client among these three, which is why I did not. Say the word if anything else here does not earn its place and I will drop it.

PR #3774, opened 2026-09-20
feat(responses): separate tool declarations from execution and validate inputs

LXD-8:
Based on `main` @ `8288884d` (v0.7.0.dev4).

## Why

`/v1/responses` accepted a declaration for a capability it cannot expose -- a hosted tool such as `web_search`, or an unknown tool type -- and then answered as if the model had seen it. Failing such a request with a 400 is not the answer either: Codex CLI 0.154 declares `web_search` unconditionally in every session, so a blanket 400 made a real session fail before it could do anything, even though a declaration the model never sees cannot change the response.

Two related gaps came with that. The endpoint accepted request modes it cannot honour (`truncation: "auto"`, `tool_choice: "required"`, `parallel_tool_calls: false`, and others) and input items it cannot replay, then answered as if it had. And an unlabelled `function_call` / `function_call_output` pair could be minted with two independently generated ids that did not match.

## What

- **Declarations are accepted, not exposed.** `convert_responses_tools()` accepts a tool declaration oMLX cannot expose, drops it from the list the model sees, and records a human-readable label. Hosted execution is never faked: the declaration is not mapped onto a function tool.
- **The degradation is reported, not silent.** A dropped declaration is named in an RFC 7234 `Warning: 199 omlx "tools accepted but not exposed to the model: ..."` response header, on both the streaming and non-streaming paths.
- **Malformed declarations still fail.** A function tool with no name, or a namespace member that is not a tool object, is still a 400 naming `tools`.
- **Replayed hosted round trips still fail.** An input item that a client sends back for a hosted call oMLX never performed (`web_search_call`, `computer_call_output`, `mcp_call`, ...) is a 400, because it describes a conversation the server did not have.
- **Request validation returns 400 naming the field.** `tool_choice` beyond `"auto"` / `"none"`, `truncation: "auto"`, `parallel_tool_calls: false`, `background`, `conversation`, `max_tool_calls`, `top_logprobs`, unknown `include` values, unsupported input item types, message roles and content part types, and `text.format` that no grammar can enforce.
- **Structured output refuses instead of degrading.** Where `/v1/chat/completions` warns and falls back to prompt injection, `/v1/responses` returns a 400 for a `text.format` it cannot enforce, because the field is an explicit output contract.
- **One identity for a tool call.** `ToolBindingRegistry` (new in this PR) is the single place both emission paths resolve a wire name back to its client-facing `(namespace, name)`, and `function_call` / `function_call_output` share one `ensure_call_id()` fallback, so an unlabelled pair can no longer be minted with mismatched ids.
- **The old alias plumbing is replaced by the registry.** `apply_namespace_tool_aliases`, the `aliases` parameter of `convert_responses_tools` / `split_namespace_tool_name` and the `{wire: (namespace, name)}` dicts they threaded through had production callers (`server.py`) until this change; `ToolBindingRegistry` replaces them, so they are gone rather than dead. The registry ships only what is called — the test-only `aliases()` accessor and the unused `wire_names` / `__len__` / `__bool__` are not part of it.
- **A namespace keeps its identity whatever order the tools are listed in.** A flat function tool keeps the name the client declared, so with the namespace listed first its member used to take the joined name and the later flat registration overwrote it: the model saw two functions under one name and a real namespace call came back to the client without its namespace (#3371). Every flat name is claimed before any namespace expands, which is order-independent again — that is what `main` did with its `taken` pre-pass.
- **An input item that omits `call_id` is a 400, not a guess.** The Open Responses schema requires `call_id` on `function_call` and `function_call_output`, and clients that follow it preserve the one the model's call carried (`@ai-sdk/open-responses` writes `call_id: part.toolCallId` unconditionally in 1.0.34 and 2.0.49 alike). Without it there is no way to know which call a result answers: minting an id, or pairing with the oldest call still waiting, both attach results on a guess — the second assumes the items arrive in order. The request is refused instead, naming the field.
- **A malformed namespace member is a 400, not a 500.** Members live in the group's `tools` list, which FastAPI never validates, so a member the schema rejects escaped as a pydantic `ValidationError` and hit the generic handler. It now raises `InvalidRequestError(field="tools")` like every sibling check.
- **The `Warning` header is bounded.** It joined every label, so 1000 declarations produced an 87 KB header — past `http.client._MAXLINE`, which makes most Python clients raise `LineTooLong` and lose the whole response. It names five and counts the rest. The degradation also logs, so an operator can see a session running without a tool the model never received; the client-side header was the only signal before.
- **A failure mid-stream reports itself.** The streaming path raised on a tool-call entry it cannot represent after SSE bytes had already been sent, so the client saw a truncated stream where the same request non-streamed gets a clean 400. It now emits the same `response.failed` the parse-error path uses.
- **`function_call` ids are normalized once, for both shapes.** Only the dict branch went through `ensure_call_id`, so an entry carrying an object with an empty id could still be minted as half a pair.
- **A refused `text.format` says what to do.** It keeps its 400, and now names the missing install step (`pip install 'omlx[grammar]'`, or the brew/DMG equivalent) the way `/v1/chat/completions` already does, instead of only refusing.

## Behaviour delta

- A tool declaration oMLX cannot expose: `400` -> `200` with a `Warning: 199 omlx "..."` header, and the declaration is absent from what the model sees.
- A malformed declaration or an input item replaying a hosted round trip oMLX never performed: still `400`.
- Request modes and input items that cannot be honoured: silently accepted -> `400` naming the field.
- `text.format` with no compilable grammar: prompt-injected JSON -> `400`.
- An input `function_call` / `function_call_output` without a `call_id`: `200` with an invented or empty id -> `400` naming `input`.

## Verification

- Five-file Responses suite on this commit (`tests/test_responses.py`, `tests/test_server.py`, `tests/integration/test_e2e_streaming.py`, `tests/integration/test_server_endpoints.py`, `tests/integration/test_responses_tool_merge.py`): **520 passed, 12 deselected** (515 before the fixes above, each of which has a test that fails without it).
- For reference, the same command on `main` reports **406 passed, 12 deselected**.
- The registration-order fix, the 400 for a malformed member and the header cap each have a test that fails against the previous revision of this branch.

This branch builds on `responses-envelope` (#3773) and will not apply to `main` on its own.


jundot:
I'm not convinced the missing `call_id` fallback belongs on the server. The Open Responses schema requires `call_id` on both `function_call` and `function_call_output`, so clients following that schema should preserve and return it. Pairing results with the oldest unmatched call assumes they arrive in order and could attach a result to the wrong call. I'd prefer a clear input error unless there is a specific compatibility need for recovery.

I'd also like to understand which tool requests this is intended to handle. oMLX can expose tools through MCP or return calls for the client to execute, even when it does not execute them itself. Was there a particular client or workflow where you ran into this? That would help me understand the intended behavior.

I'd like to base these changes on the affected clients or workflows and the problems users encounter with the current behavior.

LXD-8:
Fair on both counts — the pairing is gone (`38280832` removed it from #3773, and this branch no longer carries it).

On the evidence: `@ai-sdk/open-responses` writes `call_id: part.toolCallId` unconditionally in both directions — `function_call` and `function_call_output` alike — in 1.0.34 and 2.0.49, and the package's own item types (`FunctionCallItemParam`, `FunctionCallOutputItemParam` in `src/responses/open-responses-api.ts`) declare `call_id: string`, not an optional field. So a client that follows the schema always sends it, and pairing with the oldest unmatched call was a guess that could attach a result to the wrong call. It is not carrying a compatibility need I can point at.

What replaced it: an input `function_call` / `function_call_output` without a `call_id` (blank counts as missing) is a 400 naming `input`, and the conversion keeps `ensure_call_id` only so an id can never reach the template empty. `tests/test_responses.py` covers both item types plus the blank id; the previous behaviour was a 200 carrying an invented or empty id.

On which tool requests this is for: the endpoint exposes the request's tools to the model and turns its calls into `function_call` items for the client to execute; MCP tools are merged server-side and their calls come back through the same path. The motivation in the branch description was Codex CLI compatibility. I have no report of a client omitting `call_id`, and the SDKs I can check never do — so if this validation does not earn its place either, say so and I will drop it and leave the previous lenient fallback. The rest of this branch is unchanged: the shared registry for tool identity, the 400 for a malformed namespace member, and the tolerant handling of declarations oMLX cannot expose.

PR #3776, opened 2026-09-20
perf(qwen4): engage fused GDN verify prework via bit-exact L2 norm variant

williamxie1989:
## Problem

After the mlx-vlm upgrade (#3719), the fused GDN verify prework silently stopped engaging for `qwen4_exp` + MTP. Every T=4 verify cycle falls back to the unfused conv → SiLU → split → normalize dispatch chain (~10 extra dispatches × 64 layers per cycle).

Root cause, confirmed with in-gate diagnostics on a live server:

1. `compatible_norm` still compares against the Qwen3.5 RMS `_normalize_gated_delta_qk`, but the Qwen4 verifier now uses an L2 norm — the gate never opens.
2. At runtime `mlx_vlm.models.qwen4_exp` resolves to the compat vendor copy (inserted at `__path__[0]`), whose verifier is `_Qwen4Verifier` delegating to `layer._normalize_qk`. A gate keyed on upstream-only class names cannot match it.

## Approach

Rather than relaxing the gate, this PR reproduces the stock Qwen4 L2 chain **bit-exactly** inside the fused kernel via a template flag:

- per-element bf16 squares;
- sequential bf16 lane partials over four contiguous elements;
- fp32 xor butterfly tree across lanes with a single final bf16 rounding;
- bf16-rounded `eps` and `dk^-0.5` scalars feeding `metal::precise::rsqrt`,

matching the pinned MLX's sum/rsqrt/scalar-promotion semantics at every rounding site (verified empirically). The gate resolves whichever Qwen4 verifier wins module resolution — the compat vendor's `_Qwen4Verifier` (pinned further by layer `_normalize_qk` identity) or upstream mlx-vlm's `Qwen4ExpBatchInvariantForward`. The Qwen3.5 RMS path is untouched.

## Measurements

Apple Silicon, MTP T=4, `pp65536/tg128`: backbone **38.9–43.2 → 33.7–36.1 ms/cycle**, back to the pre-upgrade baseline band (33.5–34). Accept rate unchanged (~70–77%). Engagement is observable in logs: `[gdn-prework] fused verify prework engaged (S=4, l2=True)`.

## Relation to other open work

- **#3760** (@jundot): restores the T=1 *decode* gate — fully complementary. This PR deliberately touches only the T≥2 *verify* path; the two diffs are kept in disjoint regions and I'm happy to rebase onto whichever lands first.
- **#3626** (@pierre427): opt-in RMS verify recurrence — different mechanism and scope; no overlap with this prework-only, bit-exact change.
- **#3553** (@Rascal): explored fused verify rows; this is a narrower slice (prework fusion with bit-exactness as the acceptance criterion).

## Tests

`tests/test_qwen35_gdn_prework.py` (50 tests on this base): bit-exact parity against the stock L2 chain across batch×seq, gate routing for both verifier shapes (upstream and compat vendor), and the existing RMS-path tests unchanged.

## Notes

- The T=1 fused decode kernel remains RMS-site as shipped — mathematically equivalent to stock L2 but not bit-exact. Flagging it here for separate discussion; not touched by this PR.
- The one-shot reject diagnostics (≤3 log lines per process, T≥2 only) exist because this regression ran silently for days; happy to drop them if you'd rather keep the diff minimal.


jundot (approved):
The L2 kernel matches the stock Qwen4 chain bit for bit on the vendored layer, and on a live Qwen3.8-Flash-Next MTP server the verify prework engages again with identical greedy output and about 1.4 ms less per cycle. One thing I'll fix in a follow-up: the new `qwen4_exp` import in `apply_qwen35_gdn_prework_patch` runs at every VLM start, so loading another VLM first pins the upstream module and a later Qwen4 load fails. I'll make the gate look the module up lazily instead. LGTM, merging.

PR #3778, opened 2026-09-20
fix(chat): prevent x-model.number from fighting @input handler on thinking budget

LXD-8:
## Problem

The "Thinking tokens" field in the chat sidebar (`chat.thinking_tokens`; the settings-modal one is `Thinking Budget`) fights the person typing in it. Type `512` into an emptied field and you get `1029`; before `x-model.number` was dropped the same keystrokes produced `1025`. The first digit flashes and the value lands somewhere the user never asked for.

## Root cause

`onThinkingBudgetTokensInput` did not trust what was typed. Any change smaller than the step size was treated as a native-spinner nudge and re-stepped onto the 1024 grid:

```js
if (delta !== 0 && Math.abs(delta) < 1024) { /* re-step to prev ± 1024 */ }
```

Typing produces exactly those small deltas, so every digit-by-digit entry was hijacked — `512` arrives as `5`, then `51`, and `0 + 1024` … `1024 + 5` is where the surprise value came from. `x-model.number` was a second writer on the same field, but it was not the source of the wrong number.

## Fix

- **The field is the only writer.** The handler records what was typed (empty → no budget instead of `NaN`) and nothing re-steps it.
- **Stepping stays where it is explicit.** The arrow keys still step by 1024, and the input's own `step`/`min` now sit on that grid (`step="1024" min="0"`), so the native spinner — the only reason the deleted heuristic existed — moves in the same unit.
- `_thinkingBudgetStepping`, which guarded that branch, is gone with it.

Verified by driving the real handlers out of `chat.html` in `tests/test_chat_thinking_budget_input.py`: typing `512` now yields `512`, `2048` yields `2048`, clearing records no budget, blur only repairs empty input, and the arrow keys still step by 1024. The test fails on the previous revision (5 of its 9 cases).


jundot (approved):
Removing the delta-based correction fixes direct typing; I verified that 512 survives entry and blur, and keyboard stepping still works. The existing autosave path still restores 4096 when the field stays empty for 400 ms. I'll address that in a small follow-up so saving settings does not change the value being edited, and remove the unused input-tracking state and historical comments. LGTM, merging.

PR #3781, opened 2026-09-21
fix(cache): stop disk usage warning spam in hot-cache-only mode

williamxie1989:
## Symptom

With `cache.hot_cache_only: true`, the log fills with repeats of:

```
WARNING - Failed to check disk usage for SSD cache dir /.../.omlx/cache/deepseek_v41_ced_v1: [Errno 2] No such file or directory
```

The message repeats on *every* poll, not every 30s.

## Root cause

* In hot-cache-only mode `PagedSSDCacheManager.__init__` intentionally skips `_init_directories()` — no SSD I/O happens, so not creating the directory is correct.
* `_get_effective_max_size()` still calls `shutil.disk_usage(self._cache_dir)` from stats/eviction paths, which ENOENTs on the never-created directory. The `OSError` branch also doesn't refresh `_disk_usage_cache_time`, so the 30s TTL never engages and every call warns.
* The CED prefill cache-directory stamp from 395ec2fd (#3607) (`.../deepseek_v41_ced_v1`) made the queried path a subdirectory nothing ever creates; before it the parent cache root always existed, which is why this stayed latent. Any hot-only setup with a missing stamped/subdir path hits it.

## Fix

Hot-cache-only mode has no SSD footprint, so disk headroom is irrelevant: `_get_effective_max_size()` returns the configured max directly without the syscall — exactly the value the warning branch already returned. Non-hot-mode behavior (including the OSError warning) is unchanged.

## Tests

Regression test: with `hot_cache_only=True` and a missing cache dir, `_get_effective_max_size()` must not call `shutil.disk_usage`, must return the configured max, and must not warn. Full `tests/test_paged_ssd_cache.py`: 175 passed (incl. the new test), on `origin/main@e4c762bb`.


jundot (approved):
Reproduced the per-poll warning on main with a missing hot-only dir and confirmed it is gone on this branch with the same `max_size_bytes`. Thanks for tracing it back to the #3607 stamp and the missing TTL refresh, merging.

PR #3784, opened 2026-09-21
feat(moe): Lightning MTP coexists with DeepSeek V4.1 expert offload (resident DSpark draft head)

williamxie1989:
## Problem

MoE expert offload strips the embedded MTP draft head, so on DeepSeek V4.1 users had to pick one: fit the backbone in RAM (expert offload) **or** fast decode (Lightning MTP). Turning both on either failed validation or silently dropped the draft head.

## Why they can coexist

The DSpark draft head is small, and its drafts are verified by the streamed backbone anyway. Keeping it **fully resident** while backbone experts keep streaming from SSD costs one modest RAM bump and buys speculative decoding back.

## Changes

- `MOE_OFFLOAD_MTP_MODEL_TYPES` family allowlist; `validate_moe_expert_offload(settings, model_type=None)` is family-aware (family deferred until the checkpoint family resolves, so settings round-trip cleanly). VLM MTP and DFlash remain rejected under offload — they have no offload-aware draft path.
- `deepseek_v41/loading`: keep `mtp.*` weights only while speculation is armed (`is_mtp_active()`); with speculation off, the old strip-everything behavior is unchanged (no RAM wasted on unused draft weights).
- `mtp_resident` threads through the offload plan, `estimate_expert_savings`, `admission_bytes`, and `fit_resident_fraction` (incl. cache keys), so memory admission accounts for the resident draft head.
- Fail-closed gate at load time for unsupported families.
- Admin modal hint + i18n (9 locales) for the newly allowed combination.

## Benchmark

`DeepSeek-V4.1-Flash-oQ3e-mtp` on 128 GB Apple Silicon. Expert offload resident fraction 0.25 (96/384 experts/layer, 221.5→55.4 GiB expert weights), CED prefill (decoder tail 128), DSpark depth 3, 128-token generations. Same machine, same checkpoint, two admin speed-benchmark sessions (MTP off, then MTP on):

| phase | prefill t/s (off→on) | TTFT s (off→on) | decode tok/s (off→on) | decode Δ |
|---|---|---|---|---|
| pp=1024 | 76.1 → 77.9 | 13.5 → 13.1 | 6.61 → **8.46** | **+28%** |
| pp=4096 | 131.6 → 137.0 | 31.1 → 29.9 | 6.46 → **7.90** | **+22%** |
| pp=8192 | 137.7 → 137.0 | 59.5 → 59.8 | 6.53 → **8.95** | **+37%** |

**Average decode +29%, prefill unchanged** (±3% noise). Memory cost: loaded footprint 69.9 → 84.5 GB (+14.6 GB resident draft head).

Draft quality from the generator telemetry: acceptance **81–89%** (d1 86–88%, d2/d3 70–91%), **2.67–3.12 tok/cycle** against the 4.0 depth-3+bonus ceiling, draft overhead **<1%** of decode wall time (~110 ms vs ~15 000 ms backbone per 128 tokens).

Why the gain is ~1.3× and not the usual ~2×: under offload the width-4 verify step routes up to 4× the unique experts per layer, so backbone time per cycle rises 151→333 ms (≈2.2×); 2.8 tok/cycle ÷ 2.2 ≈ 1.27×, matching the observed +22–37%. Raising the resident fraction or packing streamed experts tighter compounds on top of this.

## Tests

- `tests/test_deepseek_v41_moe_offload.py`: 37 passed (5 new: family allowlist, mtp_resident admission accounting, speculation-off strip-unchanged, load-time gate).
- Full fast suite on `origin/main@10a79432` + this change: 13937 passed, 366 skipped; the single `test_integrations` failure is a pre-existing shell-environment artifact (`CLAUDE_CODE_AUTO_COMPACT_WINDOW` leak), fails identically on the untouched base.

Related: #3720 (earlier offload residency work by the same author — this stacks on the offload base now in main).


jundot (approved):
I verified resident DSpark drafting with expert offload on the V4.1 checkpoint; MTP on and off produced the same completed response in my smoke test. The VLM engine does swallow the new family-validation error and continue loading unsupported combinations from persisted settings. I'll make that validation propagate in a small follow-up and cover the engine entry point. LGTM with that follow-up.

PR #3786, opened 2026-09-21
feat(systemone): serve Jev structured reads behind a live /jev switch

monroewilliams:
## What

Native **System One structured reads** on oMLX's existing block-diffusion lane: `POST /jev/v1/systemone` and `GET /jev/v1/models`. No vLLM, no sidecar process, no `vllm_xargs` pinned to an unmerged PR.

A read lays a seeded answer template on the diffusion canvas and fills every question slot in **one forward** — nothing is generated and nothing is parsed, so responses carry `output_tokens: 0` by construction. Question types are `noul` (yes/no probability), `score` (rubric rating with confidence and per-level probabilities) and `choice` (selection with per-choice probabilities).

## Provenance and attribution

**This is built on OpenJev — https://github.com/razorback16/openjev — which is the source of the design.** The wire API, the question and answer shapes, the canvas and answer-template layout, the alias scheme (`jev-latest`, `openjev-latest`, …) and the entropy-based confidence normalisation are OpenJev's. Twelve lines are taken verbatim from its source — 0.35% of the ~3,400 lines added, in `omlx/systemone/read.py` and `omlx/systemone/schema.py` — and everything else is written for oMLX. Both projects are Apache-2.0.

OpenJev's own `engine.py` records that it was adapted from vLLM's `examples/features/diffusion_reads/structured_server.py` (vllm-project/vllm PR #57250, Apache-2.0), so the lineage here is vLLM → OpenJev → oMLX.

**The code in this PR was written by Qwen3.8-Flash-Next** (`Jundot--Qwen3.8-Flash-Next-oQ4e-mtp`, served locally by oMLX) working in the pi coding agent. There is no human-authored implementation in it: human involvement was direction, review of behaviour, and deployment.

## Design notes worth having on the record

- **The switch is live, not restart-gated.** `systemone_enabled` is read by a per-request dependency and the router is registered unconditionally, so the surface opens and closes without a restart. Distributed inference keeps its restart semantics — deliberately: different surface, different complexity.
- **Models are refused before their weights move.** `EnginePool.supports_systemone` inspects the discovered entry's `config_model_type`, so naming an autoregressive model costs no load. Guarded by `test_non_diffusion_model_is_404`, which asserts nothing is leased. The first cut leased the engine first and checked some thirty lines later — paying a full load to then return 404, after which the memory enforcer reclaimed the diffusion model.
- **One port, `/jev` prefix.** The SDK joins `/v1/systemone` and `/v1/models` onto a caller-supplied root, so `base_url="http://host:8000/jev"` reaches both. `GET /v1/models` is left exactly as it was — standard OpenAI shape — because read-capability discovery does not belong inside a canonical OpenAI response.
- **Aliases are code, not configuration:** `get_engine` does not resolve aliases, so `jev-latest` is resolved by this surface.
- **A diagnostic worth knowing.** When the surface is off, the gate returns `404` with detail `"Not found"`, where FastAPI's own missing-route reply is `"Not Found"`. Same status, one capital letter apart: the lower-case form means the route *is* mounted and the gate refused the request, which is the first thing to check if this surface misbehaves.

## Off by default

`systemone_enabled = false`. UI lives in **Server → Advanced** — web admin across all nine locales, plus the SwiftUI Advanced section — with a checkpoint picker fed by `systemone_readable_models`.

## Measured (M5 Max, GPU idle, `-4bit` via `jev-latest`)

| shape | median |
|---|---|
| 320 tokens, 1 sample | 0.143 s |
| 770 tokens, 1 sample | 0.225 s |
| 770 tokens, 4 samples | 0.342 s |
| 2 571 tokens, 1 sample | 0.625 s |

Server-side timings, joined per request by a client-supplied `seed`, with zero foreign inference observed across the run. Two straight lines fall out of it:

```
seconds ≈ 0.067 + tokens / 4600      (max residual 8.5 ms)
seconds ≈ 0.184 + forwards × 0.040   (max residual 2.3 ms)
```

**Samples are not a multiplier.** The first forward costs 184 ms and each additional sample 40 ms, because they all share one prefill — so an 8-sample read costs 2.25× a single sample, not 8×. Multi-sample reads are the cheapest quality available on this surface.

## Tests

`tests/test_systemone_engine.py`, `test_systemone_schema.py`, `test_systemone_api.py`, `test_systemone_admin_settings.py` — **91 tests**, all passing. The engine tests run against a tiny random-init model with no weights, so they run in CI; the tokenizer-dependent ones skip cleanly when the checkpoint is not cached, verified against an empty `HF_HOME`: **38 passed, 46 skipped, exit 0**.

## Not covered here

Long-context throughput, quality against a quant sweep, and the `steps` / `think` / `sequential` / `images` read options — the fields are on the request model but unmeasured.


monroewilliams:
This is a pure experiment -- I essentially pointed Qwen3.8-Flash-Next at the OpenJev repo and guided it to reproduce the way OpenJev implements Jev-like behavior with a diffusion-gemma model, but using the existing codepaths in the oMLX engine. 
The local tests I ran used https://huggingface.co/mlx-community/diffusiongemma-26B-A4B-it-4bit (trying for a close-ish match to the nvfp4 version pinned by OpenJev), but I would expect other mlx quants of diffusiongemma to work as well.

PR #3807, opened 2026-09-21
fix(glm5_next): eliminate the prefill memory cliff on oQ2e quantized weights

williamxie1989:
## Problem

GLM-5.3-Flash (`glm5_next`, VLM-engine fallback, 128GB Apple Silicon, MoE expert offload @60% residency) spiked ~30GB of IOAccelerator memory **inside every 2048-token prefill chunk** (sawtooth 71GB→103GB at chunk period, peak flat in context length), ending in `PrefillMemoryExceededError` at long context. Dense models are unaffected because their per-layer intermediates are ~10x smaller.

## Root cause — four compounding faults, each verified on a 128GB host

1. **Vendor patch silently dead.** Upstream mlx-vlm shipped glm5_next (PR 2030 merged); model discovery imports the site-packages copy *before* the compat patch registers, and import-once returned it — every vendor fix became dead code. `apply()` now purges the cached package, re-imports from the vendor tree, and **self-checks the resolved path** (error + `False` instead of silent success). This is why prior fixes appeared to have zero effect.

2. **fp32 activations rejected by both native DSA kernels.** oQ2e quantized projections promote through fp32 scales, so `q_latent`/`kv_latent` reached the kernels in fp32; the `fp16|bf16` dtype gates rejected every sparse layer → unfused dense `[heads, L, Kv]` score materialization, linear in context. Fixed by casting at the kernel boundary (fp32 residual stream preserved).

3. **Whole-chunk intermediate retention.** The lazy layer loop pins every intermediate until final logits eval (CPU enqueues in ~1s, GPU runs ~10.5s). Prefill (L≥256) now evals the stream per layer; decode stays lazy. The MTP runtime **replaces** this loop for `-mtp` checkpoints, so the backpressure lives in both copies.

4. **Allocator pool size-class churn.** The Metal allocator caches one buffer per size class; per-layer sizes differ (per-layer expert route counts, 2047/2048 chunk widths), so logically freed buffers stayed resident as IOAccelerator memory through the whole chunk (~30GB across 48 layers). The pool is cleared at the same eval boundaries.

## Guard pricing (defense in depth)

- Static prefill profile for glm5_next (GDN + DSA shapes, `bypass_short` dense tail, kpool-pooled indexer) replaces footprint-delta EWMA pricing.
- Flat-overhead accounting generalized from Qwen4-only (`uses_flat_overhead_accounting`).
- `speed_priority` now falls back to shrinking when the full chunk breaches the cap (a slower step beats a rejected request).
- `head_dim=0` (NoPE MLA) no longer rejects the profile from the monitor.

## Measured (LM Studio benchmark, 128GB M-series)

| | before | after |
|---|---|---|
| chunk-boundary footprint | 92–94GB | ~72–75GB |
| per-chunk sawtooth | +30GB | +1–2GB |
| peak (pp=2K–16K) | 103–108GB | <80GB |
| prefill speed | 190.9 tps @8K | unchanged |

## Tests

+6 regression tests: fp16 native routing at real dims (fp32 weights simulate quant-scale promotion), per-layer eval+pool-clear (prefill gated, decode stays lazy), vendor override after early site-packages import, profile registration, flat-overhead admission, pool accounting-once/release-on-reclaim. Full suite at pre-existing HEAD baseline, **zero new failures** (14 pre-existing, byte-identical).

## Review notes

- Vendor `language.py` and `glm5_next_vlm_runtime.py` loops are deliberately duplicated (the runtime *replaces* `Glm5NextModel.__call__`); comments cross-reference the pair.
- `test_speed_priority_guard_aborts...` → `...shrinks_when_full_chunk_breaches` is an intentional behavior change, not a dropped test.


jundot:
I reproduced two issues with native kernels enabled that need fixing before merge.

At 32K, the new FP16 sparse MLA output reaches `q8_vup_flat` with FP32 quantization scales and raises a dtype mismatch. The same attention call completes on the parent commit. Could you preserve the projection's dtype contract and add coverage using the actual kernel without stubbing `q8_vup_flat`?

The memory profile also omits expanded K/V allocations in the native exact-block path between 2K and 4K. With 32 query tokens and reduced hidden dimensions, I measured 543 MiB of additional peak memory against a 136 MiB estimate. Could you account for that path and verify small chunks as well? I'll rerun both reproductions after the update.

williamxie1989:
Thanks for the careful reproduction — both are fixed and the branch is rebased onto the latest `main`.

**1. v-up dtype contract (32K).** The FP16 sparse-MLA output was reaching `glm_dsa_q8_vup_flat` alongside the checkpoint's FP32 quantization scales, which the native kernel rejects (`scales.dtype == x.dtype` is required). `q8_vup_flat` now gates on the resolved dtypes and returns `None` on mismatch, so the projection falls back to the tolerant `mx.quantized_matmul` path and keeps its FP32 output contract — the checkpoint's FP32 scales are never silently downcast, while weights that genuinely ship FP16 scales still take the fused kernel. Coverage drives the actual kernel (no stubbing of `q8_vup_flat`): a direct dtype-gate test plus a 32K `sparse_mla → v_up` sequence. Both reproduce the exact `ValueError` before the fix and pass after.

**2. exact-block K/V expansion (2K–4K).** `_GLM5NextPrefillMemoryProfile` now prices the native exact-block route's full head-expanded K/V for `index_topk < kv_len < 4096` (`kv_len · num_heads · (qk_nope + v) · (fp16 + fp32 promotion)`). This term is linear in `kv_len` and independent of the chunk size, which closes the gap behind the 543 MiB-vs-136 MiB measurement. The gathered bound is retained for `kv_len ≥ 4096`. Added a small-chunk regression asserting the expansion is charged, monotonic across the range, and strictly above the native route's floor.

Happy for you to rerun both reproductions.

jundot (approved):
Both native-path reproductions now pass: the 32K attention call completes, and the exact-block estimate covers the measured peak for the small-chunk case. I resolved the scheduler conflict while preserving main's candidate-based chunk sizing and trimmed the repeated comments without changing executable code. LGTM, merging.

PR #3808, opened 2026-09-21
fix(deepseek_v41): eliminate the prefill memory cliff on V4.1 Flash

williamxie1989:
## Summary

DeepSeek-V4.1-Flash (deepseek_v41, MoE expert offload, DSpark MTP) sawtoothed **+8-14.5GB of IOAccelerator footprint inside every 2048-token prefill chunk**. Measured on the 128GB host (`footprint -d` sampled at peak vs trough inside one chunk): footprint 89->102GB per chunk, 100% IOAccelerator, host `MALLOC_*` untouched -- same structural failure as the GLM-5.x case (#3807). With high expert residency the adaptive throttle crossed its target after chunk 1.

Three faults, port of the GLM-5.x blueprint, each verified live on the same host:

**1. Lazy backbone loop (root cause C+D).** The 40-layer loop enqueued the whole chunk while the GPU ran ~15s behind; every intermediate (fp32 hyper-connection streams across `hc_mult=4`, packed-attention gathers, MoE routes) pinned until the final logits eval, and the Metal pool hoarded every freed size class as IOAccelerator. The loop now evals `(h, pre)` per layer during prefill (width >= 256) and clears the pool at the same boundaries; decode and DSpark draft widths stay lazy. The backbone has a single loop, so no MTP-runtime mirror is needed (unlike glm5_next).

**2. Guard pricing (no profile + missed config probes).** `make_prefill_memory_profile` explicitly excluded `deepseek_v41`, and the scheduler's HF-name probes missed the vendor `ModelConfig` field names (`n_layers`/`n_heads`/`dim`), so the monitor never received model info and the guard priced from the footprint-delta EWMA alone (measured: it predicted 2.81GB for a chunk that crossed the throttle target). Adds a static packed-sparse-attention profile (4-bit packed KV on kv/index source layers at their compress ratios; gather-bounded transient -- never dense `[heads, q, kv]` scoring, the native packed kernels are always routed) with flat-overhead accounting, plus the legacy-name fallbacks in the probes.

**3. Silent alias install (root cause A hardening).** `apply_patch` used `sys.modules.setdefault` -- the exact import-once failure mode that made three rounds of glm5_next fixes dead code after upstream merged that fork. Now installs unconditionally and logs warning/error when a foreign alias is replaced or install fails.

## Acceptance (measured)

| metric | before | after |
|---|---|---|
| per-chunk footprint swing (2048-token chunk) | +14.5GB, flat across chunks | **0GB** (78GB flat over 4 chunks) |
| throttle / eviction / OOM events | frequent mispricing | **0** |
| chunk time | 14.5-19s | unchanged within noise |
| `mx` active vs footprint peak gap | 8-18GB IOAccelerator churn | pinned-free between chunks |

## Test plan (all red first, green after)

- [x] `tests/test_deepseek_v41_prefill_backpressure.py` -- per-layer eval/clear spy on the vendor loop (prefill evals per layer, decode stays lazy); foreign vendor alias replaced + warned; `_set_model_info_for_monitor` lands a v41 profile through the widened vendor-name probes
- [x] `tests/test_prefill_oom_graceful.py` -- profile registration (exact packed-KV arithmetic), guard admits the 23:37 stall numbers at full chunk, flat-overhead charges pool once and releases on reclaim
- [x] Full suite back to baseline: 14 failed / 14415 passed -- failure set byte-identical to pre-change baseline (pre-existing `test_deepseek_v41_affine` / `test_dspark_ring` Metal-kernel failures and the claude-code integration env failure); zero new failures
- [x] Live benchmark acceptance: see table above (LM Studio pp=8K, `footprint` sampler + `footprint -d` category attribution)

## Honest attribution note

Like #3807, eval-backpressure (C) and pool `clear_cache` (D) shipped together; their individual contributions are not separated. Root cause B (dtype gate -> dense fallback) was checked and found **absent** for v41: the dense `sparse_attention` reference is dead code, all native packed-kernel symbols export, and the released shape (64 heads x head_dim 512, bf16) routes the native fast path -- so no kernel-boundary cast was needed.


jundot:
I checked the GLM path with native kernels enabled. At 3,072 KV tokens and a 32-token chunk, the native exact-block route expands full K/V; an attention-only GPU test measured about 577 MiB of additional peak memory against a 73 MiB static estimate. The 8,192-token native sparse-MLA path stayed within its estimate.

Could you account for that K/V expansion in the 2,049–4,095-token range and add a regression showing the guard rejects a chunk when the required headroom is unavailable? I'd like to resolve this before merging.

I also found that the V4.1 cache estimate counts shared index keys more than once and assumes 16-element scale groups where index packing uses 32. This overestimates cache memory; I can correct it in a follow-up.

williamxie1989:
Thanks — addressed, and the branch is rebased onto the latest `main`. One structural note: this PR stacks on #3807 (GitHub can't retarget a fork PR's base to another fork branch, so the base stays `main`); the GLM files in this diff originate from #3807 and drop out automatically once it merges.

**1. exact-block K/V expansion (2,049–4,095) + guard rejection.** With #3807's profile pricing the full head-expanded K/V in that range, the guard now charges the expansion you measured (~577 MiB against the old ~73 MiB gathered estimate). Because the expansion scales with `kv_len` and is independent of the chunk size, shrinking cannot rescue admission — so when the resident footprint leaves less than the expansion under the safety cap, the guard **rejects** the chunk instead of admitting a doomed prefill. Added a regression driving the real guard at 3,072 KV tokens / a 32-token chunk: it rejects there (and still rejects at the 32-token floor, since shrinking is futile), while admitting the native sparse-MLA route at 8,192 under the identical footprint. Without #3807's charge the same footprint is falsely admitted.

**2. V4.1 cache estimate (shared index keys / 16-vs-32 scale groups).** Noted — leaving this to your follow-up as you proposed; untouched here.

Happy for you to rerun the reproduction.

jundot (approved):
The updated estimate covers the native exact-block allocation in my reproduction, and the guard now rejects insufficient headroom while still admitting the sparse-MLA case. LGTM. I'll resolve the conflicts with main, rerun the relevant checks, and merge. I'll then correct the V4.1 shared index-key accounting and packing scale-group size in a follow-up.

PR #3809, opened 2026-09-21
fix: recognize the VLM Qwen tool parser

### LINKED ISSUE #3806: Qwen tool-call recovery and parameter coercion disabled on VLM models after #3719 — __module__ identity check mismatches mlx_vlm.tools.parsers path
First: the speed fixes from #3776 and #3791 work. On our M5 Max (128 GB), Flash-Next code decode went from 92.6 tok/s (production `7e6893e8`) to 100.7 tok/s on current `main` (`05784731`). Prose went from 59.1 to 77.4. Thank you for the fast turnaround on #3770 and #3771.

## What broke

Two features in `omlx/api/tool_calling.py` gate on the tool parser's `__module__` being `mlx_lm.tool_parsers.qwen3_coder`:

1. **Terminal tool-call recovery** (line 2320): `extract_tool_calls_with_thinking` routes to `parse_qwen_tool_calls` (which recovers unclosed envelopes) only when the module check passes. Otherwise it falls through to the generic `parse_tool_calls`, which cannot recover them.

2. **Untyped parameter coercion** (line 1992, added by `95d5bf0b` / #3780): the XML-value fallback for `oneOf`/`anyOf` parameters only fires when the module check passes.

After #3719 upgraded mlx-vlm, `VLMBatchedEngine._inject_tool_calling_support` loads the parser from `mlx_vlm.tools.registry` → `mlx_vlm.tools.parsers.qwen3_coder`. The function's `__module__` is `mlx_vlm.tools.parsers.qwen3_coder`, not `mlx_lm.tool_parsers.qwen3_coder`. Both checks fail silently for every VLM model.

LLM-only models (BatchedEngine) still load from `mlx_lm.tool_parsers` and are unaffected.

Same class of bug as #3771 (GDN verify identity check broken by the mlx-vlm module path change).

## Evidence

Streaming tool-call probe on Flash-Next (`Qwen3.8-Flash-Next-oQ4e-mtp`), M5 Max, `main` at `05784731`:

| Mode | Production (`7e6893e8`) | This build (unpatched) | This build (patched) |
|---|---|---|---|
| `unclosed` | DELIVERED (10/10) | TEXT_ENVELOPE (10/10) | **DELIVERED (10/10)** |
| `truncated` | ERROR_FRAME (10/10) | TEXT_ENVELOPE (10/10) | **ERROR_FRAME (10/10)** |
| `loop` (multi-turn) | — | — | **0 drops, 12/12 final_answer** |

## Fix

Accept either module path at both sites:

```diff
--- a/omlx/api/tool_calling.py
+++ b/omlx/api/tool_calling.py
@@ -1989,7 +1989,7 @@
                         arguments = p.get("arguments", {})
                         if (
                             getattr(tool_parser, "__module__", None)
-                            == "mlx_lm.tool_parsers.qwen3_coder"
+                            in ("mlx_lm.tool_parsers.qwen3_coder", "mlx_vlm.tools.parsers.qwen3_coder")
                             and isinstance(arguments, dict)
                         ):
@@ -2317,7 +2317,7 @@
     if (
         finish_reason is not None
         and tools
-        and getattr(parser, "__module__", None) == "mlx_lm.tool_parsers.qwen3_coder"
+        and getattr(parser, "__module__", None) in ("mlx_lm.tool_parsers.qwen3_coder", "mlx_vlm.tools.parsers.qwen3_coder")
     ):
```

A broader fix would be to check for any module path ending in `.qwen3_coder`, avoiding future breakage if the path moves again. But the explicit tuple matches the pattern used elsewhere in the codebase.

## Environment

- oMLX `main` at `05784731` (v0.7.0.dev4+29)
- mlx-vlm 0.7.1 (bundled via #3719)
- M5 Max 128 GB, macOS 26.6.2
- Model: Qwen3.8-Flash-Next-oQ4e-mtp (oQ4e, Lightning MTP)

arnavprabhu:
## What changed

The Qwen tool-call paths only recognized the parser from `mlx_lm`. VLM models now load the same parser from `mlx_vlm`, so terminal recovery and untyped parameter coercion were skipped.

This accepts the VLM module path at the two existing checks. The tests now run the same recovery and coercion cases against both parser origins.

Fixes #3806

## Testing

- `python -m pytest tests/test_tool_calling.py -q` (505 passed)
- `python -m pytest tests/ -m "not slow and not integration" --durations=50` (14,123 passed, 379 skipped, 85 deselected)
- Regression tests fail for the VLM parser when the source fix is removed (7 failed, 7 passed)


jundot (approved):
Both parser origins now reach the existing Qwen recovery and coercion paths. I verified with the actual VLM parser that terminal recovery, truncated-call errors, and oneOf/anyOf object coercion behave correctly across the affected paths. LGTM, merging.

PR #3821, opened 2026-09-21
feat(dflash): add MiMo V2.6 support

ashhart:
## Summary

- add a MiMo V2 / MiMo V2 Flash target adapter for mixed sliding-window and full-attention caches
- load the native five-layer DFlash checkpoint with MiMo partial RoPE, value scaling, attention sinks, and its trained mask embedding
- discover a complete bundled `dflash/` directory automatically when DFlash is enabled and no explicit drafter is configured
- reject unsafe or malformed mask archives without using pickle

## Validation

- `162 passed` across the MiMo, DFlash, Laguna, and engine regression suites
- full suite: `14,165 passed, 357 skipped, 85 deselected`; the seven failures were caused by an injected `OMLX_API_KEY`, and all seven passed when rerun with that variable removed
- real `Vontra/MiMo-V2.6-Flash-RL-MLX-4bit-MTP` checkpoint on Apple Silicon: DFlash output exactly matched serial greedy output for the generated sequence, with no fallback and 48.48% draft acceptance
- Ruff checks, diff checks, and outgoing-history secret scan passed

## Scope

This enables the model's text-generation DFlash path. Vision and audio execution remain outside this PR.


jundot (approved):
Thanks for implementing this and opening the PR so quickly! I tested both MiMo target implementations, and generation matched serial greedy output without fallback. There's a small runtime-signature adjustment needed for profile switching with bundled drafts, which I'll handle after merging. I always appreciate your contributions!

PR #3840, opened 2026-09-22
fix(specprefill): derive draft cache position from attention layers

tc3oliver:
### Problem

`score_tokens()` assumes `cache[0]` exposes the logical sequence offset. On
hybrid models the first layer can be a recurrent `ArraysCache` with no
`offset`, so a successfully restored draft cache is silently treated as empty.
The prompt is then re-prefilled on top of restored state, producing duplicated
cache state and incorrect importance selection.

### Root cause

A recurrent layer holds a fixed-size summary, not a growing sequence, so its
cache has no `offset` attribute at all:

```python
cached_len = cache[0].offset if hasattr(cache[0], "offset") else 0
```

`cache[0]` is chosen for no reason beyond being first. On a dense model every
layer is attention, so it happens to be right; on a hybrid model index 0 is
recurrent and the guard turns a missing attribute into a confident `0`.

One wrong number, three consequences: each prefill chunk's attention runs
against a prefix the cache already holds; `_compute_importance()` slices
`keys[..., :n_prompt, :]` off a cache holding `cached_len + n_prompt` entries,
so the scores and the selection built from them come from the wrong keys; and
the mis-trimmed cache is what gets stored.

### Fix

`sparse_prefill()` in this module already resolves attention layers to their
cache indices and reads the offset there. `_logical_cache_offset()`
generalizes that, so nothing assumes an index:

```python
cached_len = _logical_cache_offset(model, cache)
```

- `_leaf_cache_offset()` reads an integer `offset`, descending into composite
  caches; `bool` is rejected explicitly, being a subclass of `int`.
- `_logical_cache_offset()` uses the existing `_find_attention_layers()` and
  `_build_layer_to_cache_map()`, falling back to scanning every entry for an
  architecture this module does not classify.
- Only layers that mean the same thing by `offset` are compared. A bounded
  cache such as `RotatingKVCache` counts tokens *ever processed* rather than
  tokens held — `prefix_cache.py:4371-4376` refuses the same comparison for
  the same reason — so bounded layers answer only when nothing else can.
  Among the rest, a disagreement is logged at WARNING and the smallest wins:
  prefilling a token twice is recoverable, skipping one is not.
- A cache holding state but reporting no position raises
  `IndeterminateCacheOffsetError` rather than answering `0`, and
  `_cache_entry_is_empty()` fails closed in turn — a probe that throws is not
  evidence of emptiness.

Reading the position correctly also made the exact-cache-hit branch reachable
on a hybrid model for the first time (`cached_len` was always 0 there). That
branch was wrong, and the section below replaces it.

`_compute_importance()` and `sparse_prefill()` are untouched. Once `cached_len`
is right the restored cache is the true prefix of the same sequence, so the
existing slice spans exactly the prompt; `sparse_prefill()`'s inline lookup is
already correct, and folding it in would widen a fix into a refactor.

### Exact cache hit and lookahead state (review follow-up)

The review reproduced an exact-hit scoring mismatch with Qwen3.5-0.8B (8,192-token
SSD hit, 928 of 4,096 selected tokens different from cold scoring). Two
invariants were broken, both about what the draft cache holds relative to the
prompt.

**Prompt-end logits.** The lookahead starts from the logits at the last prompt
position. On an exact hit the branch replayed `tokens[-1]` on a cache that
already contained it, so that token attended to itself twice and every
recurrent layer advanced one step past the prompt. Trimming afterwards shortens
KV but restores neither the logits nor recurrent state, so scores and selection
came from the wrong state.

The only state that yields those logits is the prompt minus its last token,
followed by a forward pass over it. The draft lookup now asks for
`tokens_to_score[:-1]`. Matching is whole-block, so the restored prefix always
stops short of the prompt end and the last token always goes through the normal
`_prefill_draft` path. That is the same N-1 rule the target path applies to
exact hits (`scheduler.py`, "exact prefix hits we need cache state at (N-1)").
The replay branch is gone, and `score_tokens()` now raises `ValueError` for a
cache holding the whole prompt instead of replaying. `run_specprefill_draft_scoring`
already catches that and falls back to the normal prefill.

**Returned cache state.** The cache `score_tokens()` returns is the one stored
as this prompt's. The lookahead advances every layer by `n_lookahead + 1`
tokens; the existing trim slices attention KV back, but recurrent state cannot
be sliced and was returned as prompt + lookahead. When the prompt is a block
multiple, `store_cache` publishes live state as the last block's state
(`live_state_at_true_end`), so that polluted state was persisted and poisoned
later restores. `score_tokens()` now walks the leaf caches, descending into
`CacheList`. Unbounded KV leaves are sliced back as before. Every other leaf
(recurrent state, a bounded `RotatingKVCache` whose window has wrapped, and
so on) has its attributes held before the lookahead and put back on the same
object after it. Arrays are copied and containers rebuilt:
`ArraysCache.state` wraps a list that decode writes in place, and a wrapped
`RotatingKVCache` slice-assigns into the very `keys` object a plain reference
would hold. A shallow copy reproduced the bug unchanged. Restoring on the leaf
object rather than through `CacheList.state` avoids rebuilding sub-caches via
`from_state`, and it also trims KV sub-caches inside a `CacheList`, which the
old top-level loop never reached.

A follow-up commit (`00174797`) extends the hold to nested objects.
`reconstruct_cache` returns every restored recurrent layer as a
`SizedArraysCache`, which keeps its state on the `ArraysCache` it wraps. The
hold had kept that wrapped object by reference, so decode advanced it in place
and it was never restored. Every restored hybrid draft cache therefore came
back from scoring n_lookahead + 1 tokens ahead and was stored that way. The
hold now recurses into nested objects, holding each one once so that
reference cycles terminate, and restores them in place. On Qwen3.5-0.8B, the
returned recurrent state for an 8,193-token prompt restored at 8,192 differed
from cold by up to 19.1 before this commit and 0.25 after, which is the drift
of a different prefill chunk schedule. With #3842 stacked on top,
same-prompt restores match cold exactly (Δ = 0.0). The real-cache round-trip
test covers `SizedArraysCache`.

Neither change special-cases exact hits: the lookup rule applies to every
lookup, and the state rule to every scoring call.

**Runtime evidence**: Qwen3.5-0.8B-MLX-4bit draft, real `BlockAwarePrefixCache` +
`PagedSSDCacheManager`, block size 256, prefill step 2048, keep 50%, greedy
lookahead. The cache is stored by one request and restored by a second prefix
cache instance reading the same SSD directory, then compared with cache-free
cold scoring of the same prompt.

| case (prompt / stored) | before: restored, mismatches | after: restored, mismatches | after: max \|Δimportance\| |
|---|---|---|---|
| 8,192 / 8,192 (exact) | 8,192, **1,536 / 4,096** | 0, **0 / 4,096** | 0.0 |
| 8,704 / 8,704 (exact) | — | 0, 0 / 4,352 | 0.0 |
| 8,193 / 8,192 (partial) | 8,192, 512 / 4,097 | 8,192, 32 / 4,097 | 4.6e-4 |
| 12,288 / 8,192 (partial) | — | 8,192, 32 / 6,144 | 4.9e-4 |
| 8,704 / 8,192 (partial) | 8,192, 288 / 4,352 | 8,192, 288 / 4,352 | 6.7e-3 |

Exact hits now score exactly as cold. On this branch alone they restore 0
tokens: a hybrid draft cache has recurrent state only at the stored end, not at
the block boundary below it, so the N-1 lookup falls back to a cold prefill,
which is correct at the cost of reuse. Boundary snapshots (#3842) are what
provide that state.

The remaining partial-hit gap is prefill chunk scheduling, not cache content,
and I isolated it. The SSD-restored cache is bit-identical to an in-memory
cache prefilled the same way (`Δ = 0.0`, same selection), and after this
change the returned recurrent state is bit-equal to a cold prefill of the same
tokens. The stored prefix was prefilled as `…2047 + 1` while the cold run
covers the same tokens in full 2,048-token chunks. Two cold runs that differ
only in chunk size already differ by about 4e-4. In the 8,704 case that drift
flips the 7th greedy lookahead token (1167 vs 4744), which accounts for the
larger gap.

### Regression coverage

22 tests in `tests/test_specprefill.py`. Reverting the source change fails 18.

`TestLogicalCacheOffset` covers the helper: the dense path unchanged (512) and
empty (0); a hybrid cache whose layer 0 has no `offset` answering from the
attention layer (12288); disagreeing layers resolving to 8192 over 12288 *and*
logging both, with agreement logging nothing; a bounded layer at 100 staying
out of the comparison beside a `KVCache` at 40, and a bounded-only model still
answering; a state-bearing cache with no position raising; a genuine cold start
still answering 0; a throwing emptiness probe and a cache offering none both
failing closed; and a composite cache found one level down (4096).

Two of them use the real `mlx_lm.models.cache` classes rather than stand-ins —
`ArraysCache`, `KVCache`, `RotatingKVCache` — because a stand-in cannot catch a
mismatch between what a cache actually exposes and what this module expects.

`TestScoreTokensHybridCacheReuse` covers it end to end against a minimal hybrid
MLX fixture: layer 0 reports no offset (so the rest cannot pass vacuously); a
warm cache prefills 24 tokens rather than the full 64; cold and warm scoring
agree under `mx.allclose` and select the same tokens; and a warm cache does not
duplicate the prefix — compared by key *content* against a cold-built cache,
because length alone does not discriminate when the trim clamps to `n_prompt`
either way. For the exact-hit follow-up, the hybrid fixture now behaves like the real thing
in the two ways that matter: its recurrent layer carries state that later
outputs depend on, and its attention is causal. Without both, a replayed token
changed nothing observable and the old tests passed on the broken code. Its
`state` also has the real `ArraysCache` shape, a tuple over a list mutated in
place. Added:

- `TestDraftScoringBlockAlignedHit`: a block-aligned prompt whose draft cache
  was stored in full, run through `run_specprefill_draft_scoring`. It asserts
  the lookup excludes the last token, importance is `allclose` to cold, and
  `select_chunks()` selection is identical.
- `test_returned_cache_holds_the_prompt_state`: the returned cache equals a
  cold prefill of the prompt, KV and recurrent state both. It fails with the
  hold removed and with a shallow copy.
- `test_cache_holding_the_whole_prompt_is_rejected[64, 80]`: a cache at or
  past the prompt end raises instead of replaying.

- `TestUndoLookahead`: a round trip over real mlx-lm caches
  (`CacheList(ArraysCache, RotatingKVCache)` with a wrapped window,
  `CacheList(KVCache)`, `KVCache`). After three decode steps and the undo,
  every leaf holds its pre-lookahead content and bookkeeping, restored in
  place. It fails with a shallow copy.

Reverting the follow-up's source change fails 6 tests.

Runs at the branch head: `test_specprefill.py` 71 passed and
`test_specprefill_draft.py` 4 passed. `pytest tests/ -m "not slow and not integration"`,
which is what CI runs, gives 14,239 passed, 356 skipped, 85 deselected, 0 failed.

`ruff` reports no new findings; `black` leaves the test file unchanged.

### Runtime validation

Matched workload on a hybrid GDN pair (27B target, 0.8B draft), identical
corpus and prompts. Baseline interpolated from a 7-point no-cache arm fitted as
`1.7 + (n - 13226) * 1.886e-4` seconds.

| n_prompt | restored | suffix prefilled | scoring | baseline |
|---|---|---|---|---|
| 13,354 | 0 (cold) | 13,354 | 1.7 s | 1.70 s |
| 14,070 | 12,288 | 1,782 | 0.4 s | 1.86 s |
| 18,977 | 14,336 | 4,641 | 0.9 s | 2.78 s |
| 26,752 | 25,600 | 1,152 | 0.4 s | 4.25 s |
| 33,491 | 32,768 | 723 | 0.4 s | 5.52 s |

On every warm point the tokens prefilled equal `n_prompt - cached_len`, not
`n_prompt` — suffix-only scoring, which is the behaviour the fix is for. Across
eight warm scorings, 6.3 s actual against 30.2 s baseline-equivalent, with zero
partial-prefix-match rejections and zero scoring failures.

> This table predates the exact-hit follow-up above and measures the offset
> fix only.
>
> Runtime cache-hit measurements used a separate experimental boundary-snapshot
> branch only to make the latent restored-cache path reachable on this hybrid
> model. The correctness fix itself is independent of that branch and is
> reproduced by the standalone regression tests.

This PR does not fix boundary snapshots, and does not claim to make the draft
cache produce hits on a hybrid model by itself — the hit has to come from
elsewhere. It claims that a hit, when it occurs, is now used instead of
silently discarded. The two runtime arms were stochastic sessions, not a
fixed-seed pair, so output equivalence rests on the cold/warm tests rather than
on them.

### Scope

In: `_leaf_cache_offset()`, `_cache_entry_is_empty()`,
`_logical_cache_offset()`, `IndeterminateCacheOffsetError`, the `cached_len`
and `pre_lookahead_offset` derivations, the N-1 draft lookup, restoring every
non-sliceable leaf cache across the lookahead, and the tests. Out: draft boundary
snapshots, the static-prefix store, PCSR, the SpecPrefill threshold,
`_compute_importance()` and `sparse_prefill()`.

One behaviour change worth flagging: `score_tokens()` can now raise
`IndeterminateCacheOffsetError` where it previously continued with
`cached_len = 0`. That fires only when a cache holds state and no layer reports
a position — the case where continuing corrupts the selection — but it is a new
exception path on an existing entry point. The same holds for the `ValueError`
on a cache that already holds the whole prompt; the draft workflow catches both
and falls back to the normal path.


jundot:
I reproduced an exact-hit scoring mismatch with Qwen3.5-0.8B. The exact-hit branch appends the final token to a cache that already contains it, so trimming afterward does not repair the lookahead logits or importance scores. With an 8,192-token SSD hit, 928 of the 4,096 selected tokens differed from cold scoring. Could you obtain the correct prompt-end logits without replaying that token on the completed state, and cover score and selection correctness rather than cache length alone?

tc3oliver:
Thanks for the repro. There were two defects, and both are fixed in the new commit.

1. **Prompt-end logits.** The exact-hit branch replayed `tokens[-1]` on a cache that already held it. The draft lookup now uses `tokens_to_score[:-1]`. Matching is whole-block, so the restored prefix always stops short of the prompt end, and the last token gets a normal forward pass through `_prefill_draft`. That is the same N-1 rule the target path uses. The replay branch is removed; `score_tokens()` raises on a cache that holds the whole prompt, and the draft workflow already falls back on that.
2. **Returned cache state.** The lookahead advanced recurrent state by `n_lookahead + 1` tokens, and only KV was trimmed back. For block-multiple prompts, `store_cache` persists live state as the final block's state, so the polluted state was stored. `score_tokens()` now walks the leaf caches, descending into `CacheList`. Unbounded KV is still sliced back. Every other leaf (recurrent, a wrapped `RotatingKVCache`) has its attributes held and restored in place, with arrays copied, because decode writes those objects in place and a shallow copy doesn't hold them.

Runtime check: Qwen3.5-0.8B-4bit draft, real SSD store and restore through a fresh prefix-cache instance, 8,192-token exact hit, keep 50%. Selection mismatches against cold scoring went from **1,536 / 4,096 to 0**, with identical importance (Δ = 0.0). A returned cache is now bit-equal to a cold prefill of the prompt, KV and recurrent state both.

Two things to flag:
- On this branch alone, a hybrid exact hit now restores 0 tokens. The N-1 lookup needs recurrent state at a block boundary below the prompt end, and without #3842 none exists, so it falls back to a correct cold prefill.
- Partial hits still show a small cold/warm gap (32 / 4,097 at 8,193 / 8,192). I isolated that to prefill chunk scheduling, not cache content: SSD-restored and in-memory warm caches are bit-identical, and two cold runs that differ only in chunk size already differ by about 4e-4. The full table and method are in the updated description.

Tests now cover importance `allclose` and identical `select_chunks()` selection through `run_specprefill_draft_scoring` on a block-aligned full hit, plus returned-cache content equal to a cold prefill. It also runs a round trip over real mlx-lm `CacheList` / `RotatingKVCache` / `ArraysCache` / `KVCache`. Reverting the source change fails 6. CI selection: 14,239 passed, 0 failed.

tc3oliver:
I pushed one more commit, `00174797`, which closes a gap in the returned-cache invariant from the previous commit. Restored recurrent layers come back from `reconstruct_cache` as `SizedArraysCache`, and that wrapper keeps its state on the inner `ArraysCache`. The lookahead hold kept only a reference to that inner object, so restored caches were still stored with lookahead-advanced state. The hold now recurses into nested objects and restores them in place.

On Qwen3.5-0.8B, the returned-state difference against cold for an 8,193 / 8,192 partial restore goes from 19.1 to 0.25. The 0.25 left is chunk-schedule drift, and schedule-matched restores on #3842 reach exactly 0.0. The real-cache round-trip test now includes `SizedArraysCache`. CI selection: 14,239 passed. The description is updated.

tc3oliver:
A note on the red CI here: the only failure is `tests/test_sdpa256_attention.py::test_flash_sdpa256_memory_is_sub_quadratic[dtype0]` on Python 3.11. Python 3.12 and 3.13 pass, and every other test on 3.11 passes too, SpecPrefill included (1 failed, 14,354 passed).

The same memory test failed on Python 3.11 on upstream main as well, in [attempt 1 of run 35822857615](https://github.com/jundot/omlx/actions/runs/35822857615/attempts/1) at `736c90fc`; attempt 2 of that run, on the same commit, passed. This PR does not touch the SDPA256 path.

PR #3841, opened 2026-09-22
feat(moe): keep the GLM-5.3 draft head resident so Lightning MTP coexists with glm5_next expert offload

williamxie1989:
> **Stacked on #3784 — please review the second commit only.**
>
> This branch is `main` + #3784 rebased onto current `main` + the GLM-5.3 commit. Until #3784 merges, the diff below contains both; once it lands, this PR's diff shrinks to the GLM-5.3 commit automatically.
> #3784 had gone `CONFLICTING` against `main` (i18n churn plus one adjacent-insert in `dashboard.js`); the first commit here is that same change, rebased.

## What

Lets **GLM-5.3-Flash (`glm5_next`)** run Lightning MTP **and** MoE expert SSD offload at the same time, mirroring the resident-draft-head design of #3784 for DeepSeek V4.1.

Before this, `validate_moe_expert_offload` rejected the combination for every family but `deepseek_v41`, and the offload adapter's tree walk reached the nextn draft head:

```
language_model.mtp.<i>.block.mlp.switch_mlp   (model-mtp.safetensors, full 288-expert slab)
```

so turning MTP on under offload streamed the draft head's experts from SSD on **every draft step**. Users had to pick between fitting the backbone in RAM and fast decode.

The head is one decoder layer whose drafts the streamed backbone verifies, so it stays fully resident while backbone experts keep streaming — exactly the V4.1 trade.

## Changes

- `MOE_OFFLOAD_MTP_MODEL_TYPES` gains `glm5_next`, so the family-aware validation and the admin gate it drives allow the pair (`model_settings.py`).
- `apply_deepseek_v4_moe_expert_offload(..., mtp_resident=)` skips `SwitchGLU` modules under an `mtp.` path; MTP off keeps the old wrap-everything behavior (`patches/deepseek_v4/moe_offload.py`).
- `mtp_resident` threads through `apply_moe_expert_offload`, `estimate_offload_admission_bytes`, the VLM engine's load path, and `engine_pool`'s projection, so admission never discounts the slab the adapter refuses to offload.
- `estimate_offload_admission_bytes` ignores `mtp.*` expert tensors while the head is resident.

## Benchmark

GLM-5.3-Flash-oQ2e-mtp, 128 GB Apple Silicon, offload resident fraction 0.8, pp=8192 / tg=128, admin speed benchmark, MTP off vs on:

| metric | MTP off | MTP on | delta |
| --- | --- | --- | --- |
| TTFT (ms) | 43261.9 | 42585.7 | -1.6% |
| TPOT (ms/tok) | 79.95 | 56.48 | **-29.4%** |
| pp TPS | 189.4 | 192.4 | +1.6% |
| tg TPS | 12.6 | 17.8 | **+41.3%** |
| E2E (s) | 53.421 | 49.762 | -6.8% |
| Peak mem (GB) | 90.95 | 91.83 | +0.88 |

Draft quality is healthy: acceptance 79.3% (d1 46/50, d2 17/26, d3 10/16), 2.42 tok/cycle, draft overhead 154.0 ms against a 6915.4 ms backbone.

Peak memory grows by 0.88 GB — essentially the 0.82 GB of offload saving the resident head gives up (measured on this checkpoint at 0.8 residency), so the head is close to free.

## Tests

Family-aware validation, the admin save-time gate (including the hyphenated `glm5-next` model type), the wrap skip, the admission exclusion, and the `engine_pool` threading. Full fast suite green with no new failures against the same base.

## Stacked on

Depends on #3784 (V4.1 resident DSpark head), which introduces the allowlist, the family-aware validation and the admin toggle gate. Rebase onto `main` once #3784 lands.


jundot (approved):
Keeping the draft head resident matches the admission accounting, and my small GLM runtime checks produced identical backbone and draft logits with and without offload. I'll resolve the conflicts with main while preserving the validation fix from #3784. LGTM, I'll squash merge after updating the branch.

PR #3842, opened 2026-09-22
feat(specprefill): preserve draft recurrent state at cache boundaries

tc3oliver:
> **Depends on #3840. Please review independently, but this PR should not merge before #3840.**
>
> #3840 makes restored hybrid draft-cache state usable correctly. This PR makes reusable recurrent state available at draft cache block boundaries.
>
> **Updated for #3840's N-1 lookup contract.** #3840 now queries the draft cache with all but the last prompt token, so this PR publishes the last boundary the prefill actually stands on *strictly below* the prompt end. See [Boundary below the prompt end](#boundary-below-the-prompt-end).

## Problem

The SpecPrefill draft prefix cache never produces a hit on a hybrid GDN model.
Every scoring re-prefills the whole prompt, however much of it the cache
already holds.

## Root cause

Draft scoring calls `store_cache()` without `boundary_snapshots`, which the
target path has passed since the parameter was added. Recurrent layers cannot
be re-sliced out of a live cache the way KV layers can, so without a snapshot
every stored block carries a **placeholder** for each recurrent layer. The only
live recurrent state sits past the trailing partial block, which `store_cache`
skips. Walk-back then scans every stored block, correctly finds no usable
checkpoint, and reports a miss — every time, for every conversation. The
runtime says so directly:

> `ArraysCache layer 0: partial prefix match detected (placeholder in last
> matched block). Rejecting cache to prevent stale GDN state.`

## Fix

Capture the recurrent state at a block boundary during the draft prefill and
pass it to `store_cache` keyed by absolute token count, exactly as `Scheduler`
does for the target model.

- **Sliceable layers are nulled, not carried.** `store_cache` re-slices those
  from the live cache; pinning them in a snapshot would hold the whole growing
  KV for nothing. This mirrors `scheduler.py`'s own boundary emission.
- **The boundary is chosen on the first progress report**, not from the cached
  length the caller expected. Those two should agree; if they ever did not, the
  capture would silently never fire.
- **It is found by walking the prefill loop, not by solving it** — the last
  chunk is truncated to leave a token for the logits call, so the reported
  positions are not multiples of the step. A closed form got this wrong.
- **One boundary, not one per block.** Extracting at every block boundary and
  keeping the last costs a full state extraction per block for nothing.
- **The cache's lifetime becomes this function's problem.** It is allocated
  here because `score_tokens` hands its own back only after scoring, too late
  to read state at a boundary. So nothing may still name it, or state pulled
  out of it, when the buffers are returned — otherwise the draft KV outlives
  the call that exists to reclaim it, and the reclaim accounting measures a
  delta it did not get.

## Boundary below the prompt end

#3840 established that the logits the lookahead starts from must come from a
real forward pass over the last prompt token, so the draft lookup asks for
`tokens_to_score[:-1]` and a hit can never cover the prompt end. The first
revision of this PR still let a block-aligned prompt publish its snapshot at
`n_to_score`. That is a boundary the lookup can never match: scoring the
same prompt again matched the blocks below the end, found only placeholders,
and restored nothing.

`_last_reachable_boundary()` now returns the last block-aligned position that
`_prefill_draft` actually reports **strictly below** `n_to_score`. It still
walks the real loop rather than solving it. The prefill stands only on chunk
ends, on the truncated last chunk at `n_to_score - 1`, and on `n_to_score`
itself after the logits call, so the answer is not `n_to_score - block_size`.
For 8,192 tokens in 2,048-token chunks, the reported positions are 2,048,
4,096, 6,144, 8,191 and 8,192, and the boundary is **6,144**. There is no
special case for exact hits: the same rule applies to every prompt, and for a
non-aligned prompt it picks the same boundary as before.

A longer prompt that extends a block-aligned one still resumes at the end.
`store_cache` publishes the returned live state as the final block's state
when the store is a block multiple, and since #3840 that state is the
prompt's own.

**Runtime**: Qwen3.5-0.8B-MLX-4bit draft, real `BlockAwarePrefixCache` +
`PagedSSDCacheManager`, prefill step 2,048, keep 50%, greedy lookahead. The
first request stores; a fresh prefix-cache instance reading the same SSD
directory serves the second, which is compared with cache-free cold scoring.
Block size 2,048 is what the scheduler forces for an `ArraysCache` model; 256
is shown to exercise finer blocks.

| prompt / stored | block | restored before → after | importance Δ | selection mismatch | returned cache Δ |
|---|---|---|---|---|---|
| 8,192 / 8,192 (same prompt) | 2,048 | 0 → **6,144** | 0.0 | 0 / 4,096 | 0.0 |
| 8,192 / 8,192 (same prompt) | 256 | 0 → **6,144** | 0.0 | 0 / 4,096 | 0.0 |
| 8,704 / 8,704 (same prompt, the case verified in review) | 2,048 | 8,192 | 0.0 | 0 / 4,352 | 0.0 |
| 8,704 / 8,704 (same prompt, the case verified in review) | 256 | 8,192 | 0.0 | 0 / 4,352 | 0.0 |
| 10,240 / 8,192 (extension) | 2,048 | 8,192 | 5.1e-4 | 32 / 5,120 | 0.33 |
| 8,193 / 8,192 (extension) | 256 | 8,192 | 4.6e-4 | 32 / 4,097 | 0.25 |
| 12,288 / 8,192 (extension) | 256 | 8,192 | 4.9e-4 | 32 / 6,144 | 0.38 |

The before column for the aligned prompt was measured at `f1c22e7d`, this
PR's first commit rebased onto the updated #3840. Where the restore point is
one the cold prefill also stands on, cold and warm run the same chunks and
agree exactly on importance, selection and the returned cache. The extension
rows restore at the end of a stored prompt, where the stored prefix was
prefilled as `…2,047 + 1` and cold covers the same tokens in full chunks.
Their residual is that schedule drift, not cache content: two cold runs that
differ only in chunk size already differ by about 4e-4, and the SSD-restored
cache is bit-identical to an in-memory one prefilled the same way. Without
#3840's lookahead-state fix these rows showed a returned-cache Δ of about 19.

## Why #3840 is a prerequisite

Two separate defects sit between a hybrid GDN draft model and a usable draft
prefix cache, and they fail differently.

**#3840 is the correctness prerequisite.** When a restored cache exists,
`score_tokens()` has to read its logical position so that only the uncached
suffix is prefilled. On a hybrid model it read `cache[0].offset`, and layer 0
is recurrent with no `offset`, so a restored cache was treated as empty. #3840
also fixes the exact-hit replay: the lookup is now N-1, which is the contract
this PR's boundary rule is built on. And it undoes the lookahead on every
non-sliceable cache, including the `SizedArraysCache` wrapper that every
restored recurrent layer comes back as, so the cache stored after scoring
holds the prompt's own state.

**This PR is the reuse enablement.** Even with the position read correctly, a
hybrid GDN draft cache has no reusable recurrent boundary to restore *from*.

The dependency runs one way, and only for the measurements: **the runtime
validation depends on #3840**, because without it the cache-hit path itself is
wrong — a hit is silently discarded and the prompt re-prefilled, so there would
be nothing to measure. The regression tests here do not depend on it.

Nothing in this PR re-explains or modifies #3840's fix. The two can be reviewed
independently and merged separately; this one just has to go second.

**Stacked branch.** GitHub currently shows #3840's three commits along with
this PR's and the combined diff. The two commits belonging to this PR are
`feat(specprefill): preserve draft recurrent state at cache boundaries` and
`fix(specprefill): publish the draft boundary below the prompt end`, touching
only `omlx/specprefill/draft.py` and `tests/test_specprefill_draft.py`. Once #3840 merges I will rebase onto the
updated `main` and this diff will shrink to that commit alone.

## Validation

> The table below predates the N-1 lookup revision; see
> [Boundary below the prompt end](#boundary-below-the-prompt-end) for the
> current runtime evidence.

Baseline is this branch's parent — the offset fix alone — so the only
difference is the snapshot plumbing. Both arms ran back to back on the same
corpus and prompts, on a hybrid GDN pair (27B target, 0.8B draft).

| | baseline | this PR |
|---|---|---|
| scorings | 10 | 27 |
| draft cache hits | **0** | **23** |
| `partial prefix match` rejections | 9 | **0** |
| scoring time | 33.1 s | 38.7 s actual vs **140.4 s** baseline-equivalent |

The published boundary advances with the conversation, and each hit is consumed
by #3840's suffix-only scoring, so the tokens prefilled are `n_prompt - cached`.
Four of the 27 scorings missed the cache; those land at 0.9x–1.0x of the fitted
baseline, so the snapshot machinery costs nothing measurable when it does not
pay off. All 27 sparse prefills completed, zero rejections, zero scoring
failures, and both arms reached the same answer on the task.

<details>
<summary><b>Runtime table, fit, and boundary progression</b></summary>

Baseline least-squares fit `-1.05 + n * 1.920e-4` s, fitted on
n = 13,291..34,010.

| n_prompt | cached | suffix | actual | baseline@n | |
|---|---|---|---|---|---|
| 13,224 | 0 (cold) | 13,224 | 1.6 s | 1.5 s | 0.9x |
| 16,556 | 12,288 | 4,268 | 0.8 s | 2.1 s | 2.7x |
| 18,180 | 16,384 | 1,796 | 0.5 s | 2.4 s | 4.9x |
| 19,602 | 17,408 | 2,194 | 0.5 s | 2.7 s | 5.4x |
| 20,446 | 19,456 | 990 | 0.3 s | 2.9 s | 9.6x |
| 26,904 | 19,456 | 7,448 | 1.6 s | 4.1 s | 2.6x |
| 28,499 | 27,648 | 851 | 0.4 s | 4.4 s | 11.0x |
| 30,441 | 29,696 | 745 | 0.4 s | 4.8 s | 12.0x |
| 34,699 | 32,768 | 1,931 | 0.7 s | 5.6 s | 8.0x |
| 38,495 | 0 (cold) | 38,495 | 6.6 s | 6.3 s | 1.0x |
| 40,838 | 0 (cold) | 40,838 | 7.2 s | 6.8 s | 0.9x |
| 47,552 | 47,104 | 448 | 0.5 s | 8.1 s | 16.2x |
| 47,978 | 47,104 | 874 | 0.6 s | 8.2 s | 13.6x |

The table above is thirteen representative rows of the twenty-seven.

Published boundary, all 18 distinct values over 21 captures: 12288, 16384,
17408, 19456, 25600, 26624, 27648, 29696, 31744, 32768, 33792, 34816, 35840,
36864, 37888, 38912, 39936, 47104.

</details>

<details>
<summary><b>Limitations of the runtime measurement</b></summary>

- Prompt trajectories diverge between the arms because each session's own
  output differs, so the comparison is the per-prompt fit above and **not** the
  raw totals.
- The treatment arm ran past the fitted range (n > 34,010), where the baseline
  column is extrapolated.
- Output equivalence is two stochastic sessions agreeing on one answer, **not**
  a fixed-seed pair.
- The runs used `block_size` 1024, which is that machine's configured value;
  the upstream rule below would pick 2048 for the same model. Replaying the
  measured prompt sequence through the boundary arithmetic gives the same 11
  published boundaries and the same 800,768 tokens not re-prefilled at both
  1024 and 2048, and 7 boundaries / 770,048 tokens at 4096 — so the block size
  does not change the conclusion on this workload. **Derived**, not measured:
  it replays the arithmetic, not the server.

</details>

<details>
<summary><b>How the coupling to <code>_prefill_draft</code> is kept honest</b></summary>

`_last_reachable_boundary()` models `_prefill_draft`'s reporting loop. That is
a coupling to another module's private function, and the failure mode if it
drifts is silent: the feature would capture nothing and the draft cache would
quietly go back to never hitting.

So the tests pin the loop itself. `test_prefill_draft_schedule_is_exactly_this`
asserts the exact sequence of reported positions, driving the real
`_prefill_draft`; a change there breaks a test and asks a human to re-derive
the arithmetic. Asserting only the prediction is not enough — several plausible
edits to that loop leave the largest aligned position unmoved, and a weaker
version of this test passed against a deliberately perturbed loop.

`test_boundary_prediction_matches_the_real_loop_everywhere` sweeps
`step, block ∈ [2,20)` and `n ∈ [2,200)` against the real loop, taking only
reported positions strictly below `n`, and asserts the prediction is never
`n` itself. An earlier
closed-form implementation was wrong 6729 times over that sweep — for instance
`cached=0, m=3073, step=2048, block=1024` published 2048 although 3072 was
reported and block-aligned, losing a block of reuse each time.

`test_prefill_draft_holds_exactly_the_tokens_it_reports` pins the other half:
at every reported position the cache holds exactly that many tokens. If that
stopped holding, a checkpoint would be stored under the wrong token count and a
later restore would resume from the wrong place.

`TestRepeatedBlockAlignedPrompt` drives the whole path with real pieces: a
tiny random-weight Qwen3.5 (GDN and attention layers), `BlockAwarePrefixCache`
over a `PagedSSDCacheManager` on disk, and the scheduler's own state
extraction. A block-aligned 128-token prompt is stored and then scored again
through a fresh prefix-cache instance. The second request restores 96, the
boundary derived from the reported schedule 32, 64, 96, 127, 128. Importance
is `array_equal` to cold, selection is identical, and the returned cache
equals a cold prefill of the prompt. Under the old rule it restores 0 and
fails.

`test_restored_cache_is_released_before_the_clear` takes a weak reference
across `sync_and_clear_cache()` and fails if anything still names the restored
cache there.

`_SLICEABLE_CACHE_TYPES` duplicates `scheduler._KNOWN_SLICEABLE_CACHE_TYPES`,
because importing it would make `omlx.scheduler → omlx.specprefill.draft →
omlx.scheduler`. Tests have no such cycle, so
`test_sliceable_cache_types_track_the_scheduler` asserts the two are equal.
(`CacheTypeRegistry.is_sliceable` is a different question — it classifies
`ChunkedKVCache` as not sliceable where the scheduler deliberately does, and
the consumer of these snapshots follows the scheduler.)

</details>

<details>
<summary><b>On the block size, and a branch that was removed</b></summary>

`Scheduler._enlarge_block_size_for_arrays_cache` forces
`paged_cache_block_size = max(2048, prefill_step_size, qwen35_floor)` for any
model with an ArraysCache, so on a GDN hybrid the block size is always at least
the prefill step.

An earlier revision carried a fallback that re-chunked the prefill at the block
size when no boundary was reachable. Under that rule it can never fire on the
models this exists for, and where it could fire — a non-GDN model, block 256 —
every layer is sliceable, so the capture returns early and the extra chunks buy
nothing. It is removed rather than kept as a branch that only ever costs
throughput. Removing it also keeps the draft chunk schedule fixed at
`prefill_step_size`, which matters because GDN chunked recurrence is not
bit-identical across different chunk splits.

</details>

## Test gate

`test_specprefill.py` + `test_specprefill_draft.py`: 89 passed.
`pytest tests/ -m "not slow and not integration"`, which is what CI runs:
14,253 passed, 356 skipped, 85 deselected, 0 failed. `ruff` is clean on the
touched files.

## Scope / non-goals

Only draft boundary-snapshot plumbing. `gdn_ssd_split_enabled` is not enabled,
the target cache is untouched, and nothing here changes the static-prefix
store, PCSR, or the SpecPrefill threshold. It does not re-fix #3840's logical
offset; it depends on it.


jundot:
I verified that the boundary snapshots enable disk restoration: an 8,704-token prompt restored 8,192 tokens and produced identical importance scores and selections to cold scoring. The offset-only baseline rejected that same cached prefix. I'm holding this pending the exact-hit correction in #3840, since the restored-cache path also reaches that branch.

tc3oliver:
This is rebased onto #3840's latest commit and adapted to its N-1 lookup. `_last_reachable_boundary()` now picks the last block-aligned position `_prefill_draft` actually reports strictly below the prompt end. It still walks the real loop, with no exact-hit special case.

Runtime results on Qwen3.5-0.8B through a real SSD store and a fresh prefix cache, at block size 2,048 (the value the scheduler forces for GDN) and at 256:
- An 8,192-token prompt scored twice now restores **6,144**; before this change it restored 0. The value is derived from the reported schedule 2,048 / 4,096 / 6,144 / 8,191 / 8,192.
- Your 8,704-token case restores 8,192.
- Both match cold exactly: importance Δ 0.0, 0 selection mismatches, returned cache Δ 0.0.
- Extension prompts still resume at the stored end (8,192).

A new end-to-end test drives the same path with a tiny random-weight Qwen3.5 and the real prefix-cache/SSD stack, and it fails under the old rule. CI selection: 14,253 passed. Details are in the updated description.

PR #3843, opened 2026-09-22
fix(glm5_next): accept vision_tower.* names and sanitize-matched quant overrides

cropduster:
Fixes the two checkpoint-compat failures reported in #3380 for community GLM-5.3-Flash exports (`Vontra/GLM-5.3-Flash-MLX-oQ*`).

### Problem

Loading such a checkpoint fails (HTTP 409 on the first request):

```text
VLM loading failed for GLM-5.3-Flash-MLX-oQ2-MTP, falling back to LLM:
  Received 347 parameters not in model: vision_tower.post_layernorm.weight.
LLM fallback also failed: Model type glm5_next not supported.
```

Reproduced on `main` @ `e3982c3f` (re-checked today). Environment: Mac Studio M4 Max, macOS 27.2, MLX 0.32.2, mlx-lm 0.32.0, mlx-vlm 0.7.1, checkpoint `Vontra/GLM-5.3-Flash-MLX-oQ2-MTP`.

### Cause

These checkpoints were exported under a vendored `glm5_next` implementation that writes raw HF names, so two things do not line up with the runtime module tree:

1. The vision tower arrives as `vision_tower.*`, while `sanitize()` only remaps `model.visual.` and `visual.` — hence 347 rejected tensors and the failed LLM fallback.
2. The per-tensor quantization keys keep the raw export names (`.hc_attn_`, `.hc_ffn_`, flat `.self_attn.f_a_proj`/`f_b_proj`/`A_log`/`dt_bias`), so after sanitizing they no longer match the runtime paths and `nn.quantize` silently falls back to the global bit recipe.

### Changes

1. `omlx/patches/mlx_vlm_glm5_next_compat/vendor/mlx_vlm/models/glm5_next/glm5_next.py` — treat `vision_tower.` like the existing `visual.` case in `sanitize()`.
2. `omlx/utils/model_loading.py` — mirror the `sanitize()` renames when expanding per-tensor quant overrides.

Both are additive and scoped to name mapping: checkpoints that already use the `visual.` / `model.visual.` conventions are unaffected.

### Verification

With both fixes applied, the same checkpoint loads through the VLM path (no LLM fallback) and runs — including the MoE expert offload and MTP draft head:

```text
deepseek_v4 moe expert offload: wrapped 42 layers at 87.5% residency
  (expert tables: 95.13 GB total, 83.24 GB resident), draft head resident
Loaded model: GLM-5.3-Flash-MLX-oQ2-MTP (actual: 92.30GB, local estimate: 96.62GB, full model: 107.69GB)
```

Long context, expert offload and MTP all behave normally on the machine above (measured: 128k prompt, 107.9 GB peak footprint, 15.7 tok/s decode, no swapping).

Background and the full analysis: #3380.


jundot (approved):
The aliases match the existing sanitizer paths and preserve explicit quantization overrides. I reproduced the mixed-bit loading failure with a small GLM model and confirmed that the fix loads successfully with identical logits. LGTM, merging.

PR #3852, opened 2026-09-22
fix(scheduler): respect fixed reclaim costs when sizing prefill

samfenwick:
Generic prefill sizing can return a chunk whose own memory prediction is still above the safety cap. The predictor includes a fixed charge for recently released pool memory, but the adaptive throttle and pre-chunk guard currently shrink that charge in proportion to the token count.

On current main, a controlled replay with 1,000 MiB headroom, a 960 MiB reclaim charge and a 512-token request returns 480 tokens with a predicted 1,047.75 MiB charge. Running the guard after the adaptive throttle still returns 448 tokens at 1,041.90 MiB. These are real scheduler/predictor calls with controlled footprint readings, not a live OOM experiment.

This reuses the existing candidate-prediction search for generic models as well as Qwen4. Full chunks that fit, minimum-size aborts, safety factors and reclaim accounting retain their existing behavior. The search also respects the existing chunk-snap opt-out.

This was prompted by @lyonsno's sizing review on #3514. It fixes the independently reproducible current-main reclaim path introduced in #2573, using main's existing estimator, and can merge independently of #3514's partial-observation work.

The patched replay returns 192 tokens at 995.10 MiB for all six variants. The six new regression cases fail on unchanged main. Validation:

- 205 focused checks pass: 201 scheduler/tracker/preflight cases and four tiny-model prefill-loop cases run separately under a GPU guard.
- Coverage includes both routes, adaptive/guard/composed sizing, a charge that cannot fit, linear/equality/grid behavior, and successful reclaim preserving short or off-grid slices. The new loop cases check every submitted width and the final cache offset.
- `git diff --check` and changed-hunk Black checks pass.
- Upstream CI passes on Python 3.11, 3.12 and 3.13 at head `52da684a` ([run](https://github.com/jundot/omlx/actions/runs/35779144569)).
- Local default suite: **14,509 passed, 150 skipped, 85 deselected, 6 failed** on the M1 Max/Python 3.11 review environment. Untouched main `8288884d` has the same six failing test identities (three FP32 decode-SDPA cases, one GLM native-fused parity case and two Qwen4 hyper-connection cases), with 14,470 passed. There are no new failing test identities; this is not a clean full-suite pass. The 39 added cases account for the increase in passes.

This is a memory-guard correctness fix, not a throughput claim: constrained prefills may use smaller chunks and run more slowly to respect the existing budget. Footprint readings in the reproduction and loop tests are controlled; a high-memory checkpoint has not been driven to the limit.


jundot (approved):
Candidate-based sizing preserves the fixed reclaim charge that proportional scaling discounted. I reproduced the over-budget result on main and verified that this change stays within the predicted cap through both prefill loops while preserving the final cache offset. LGTM.

PR #3853, opened 2026-09-22
perf(qwen35): automatically fuse eligible FP16 decode on M1 Max

samfenwick:
FP16 Qwen3.5/3.6 35B-A3B decode builds the convolution, SiLU and Q/K normalization chain for every GDN layer. This routes eligible single-token calls through oMLX's existing fused prework kernel automatically on Apple M1 Max. The kernel and its attribution are unchanged; projections, gates, recurrent arithmetic and output normalization keep their existing implementation.

The environment switch has been removed. Device selection happens once at patch installation. Other chips retain the existing route: I have measured one M1 Max, not M2 hardware, so this deliberately does not enable an unmeasured device. Three local 4-, 5- and 6-bit-default FP16 conversions now have numerical and whole-server coverage. They share the 35B-A3B architecture; this is not coverage of three unrelated model families.

The call must have the tested geometry, batch 1, one token, FP16 convolution/input and FP32 recurrent state. Training, prefill, batches, masks, padding/length metadata, speculation, different precision/geometry, cache subclasses and extended runtime keywords keep the original path. BF16 models are not converted. Existing Qwen4/BF16 and speculative prework routes remain independent. Both next cache states are computed before committing either, and failed fused calls are not retried against advanced caches.

## Whole-server measurements

Apple M1 Max / 64 GB, MLX 0.32.2, mlx-vlm `ea79808c`. Untouched main `9dff1dd7` versus this branch, identical native builds and settings, stock balanced bursts, greedy decoding, thinking/speculation off, native attention KV and FP32 recurrent state. Fresh-process A/B/B/A with discarded warmups gives six retained samples per arm. There is no FP16 environment flag in either arm; logs verify actual engagement only in the candidate.

| FP16 conversion | Main decode tok/s | PR decode tok/s | Main 256-token reply | PR reply |
| 4-bit default | 73.79 | 78.33 | 3.569 s | 3.360 s |
| 5-bit default | 59.75 | 63.32 | 4.395 s | 4.150 s |
| 6-bit default | 57.97 | 61.35 | 4.731 s | 4.502 s |

Median decode improves 5.8–6.2%, with complete-response time down 4.8–5.9% in these cases. Each checkpoint uses mixed quantization, not uniformly the stated bit width. The 6-bit checkpoint contains MTP tensors, but speculation is disabled. Q4 was measured at `1a8147ae`; the only subsequent runtime edit in `c30b713c` is import sorting, with identical non-import AST nodes. Q5/Q6 and the full suite use `c30b713c`.

All 270 synthetic benchmark rows pass, covering text, tools/receipts, prefix reuse, two images, overlap, cancellation and recovery. Requests, completed text/tool arguments, lengths and cache counts match across arms. Cancellation partials are excluded from deterministic comparisons. First-content latency does not improve consistently; this is not evidence of faster long-context prefill or whole-application latency. MLX allocated high-water is similar, while process RSS varies materially between phases; process-memory neutrality is not claimed.

[Exact scripts, every sample, independent audits and limitations](https://gist.github.com/samfenwick/24a9f5441fdb60d71ecf9f73db312e02). Checkpoint hashes/configurations are included; weights and private data are not redistributed.

## Validation

- 105 focused tests pass, including automatic hardware selection, unsupported-device fallback, exact FP16 kernel comparisons, state transitions, speculation, runtime keyword forwarding and failure without cache mutation.
- 2,268 real-weight numerical cases match exactly across the three conversions. All 1,170 expected calls reach the fused route. This covers all 30 GDN convolution layers at six scales/two strides, and evolving/restored states for three complete quantized layers per conversion.
- Default suite: **14,635 passed, 150 skipped, 85 deselected, six failed**. Untouched current main has 14,583 passes and exactly the same six failing identities: three FP32 decode-SDPA cases, one GLM native-fused comparison and two Qwen4 hyper-connection cases. No added failures; this is not a clean local full-suite pass.
- 17 harness/verifier checks pass, including deliberately corrupted source/flag/engagement evidence. Black and whitespace checks pass; the pre-existing `A_log` argument still triggers Ruff N803.

Scope and reproduction are documented in `docs/experimental/qwen35_fp16_decode.md`. Further chips need equivalent measurements before extending the automatic hardware gate.


jundot:
@samfenwick I’m not sure another environment variable that users are unlikely to discover is the right direction here. If broader testing shows this reliably benefits M1/M2 users, I’d prefer enabling it by default for eligible calls. An advanced setting could make sense if users need to choose when to use it, but that may just add complexity without much practical benefit.

How do you see users benefiting from this feature? Is there a reason they should need to opt in, rather than having oMLX apply it automatically when the conditions are met?

jundot (approved):
@samfenwick I initially read the FP16 scope as an optimization mainly for M1/M2 systems. After testing the same fused kernel with BF16 on my M3 Ultra, I found that the benefit extends beyond that scope. Non-MTP decode improved by about 8.6% on Qwen3.6-35B-A3B and 5.5% on Qwen3.8-27B-oQ4e, with matching outputs in each before/after comparison. I didn't observe an additional benefit with Lightning MTP enabled.

I'll merge this and follow up by removing the hardware restriction, adding BF16 coverage, and deriving the dimensions from the model while retaining the kernel's compatibility checks. That should let compatible Qwen GDN calls benefit automatically without chip or model-size allowlists. Thanks for finding this and putting together the implementation and validation.

PR #3854, opened 2026-09-22
perf(engine): release the first generated chunk before continuing a decode burst

samfenwick:
The first generated chunk waits inside `_step_burst()` while later tokens decode. The burst deadline starts after the initial scheduler step, so even a completed long prefill can incur a fresh burst budget before token one reaches the event loop.

End the burst when any request emits its first generated chunk, including requests admitted mid-burst. Existing cumulative/new token counts cover multi-token chunks without per-request state. Later chunks retain the current adaptive budgets and step cap. The tradeoff is a possible additional executor handoff.

### Current-main benchmark

M1 Max / 64 GB; MLX 0.32.2, mlx-vlm `ea79808c`; main `8288884d` versus `586af504`. Same custom Qwen3.6-35B-A3B 4/5/6/8-bit checkpoint with FP16 activations, identical native builds, greedy decoding, thinking/speculation off and native FP16 attention KV. Each stock burst mode uses A/B/B/A, fresh server/cache per phase, discarded warmups and six measured samples per arm.

| Median | Main | Patched |
| --- | ---: | ---: |
| Balanced: first content | 280 ms | 182 ms |
| Aggressive: first content | 380 ms | 175 ms |
| Balanced: complete 256-token response | 3.570 s | 3.588 s |
| Aggressive: complete 256-token response | 3.580 s | 3.590 s |

This saves approximately **98/204 ms before first content** with stock 0.1/0.2-second burst budgets. Full-response medians are approximately **0.5%/0.3% slower**; decode throughput is not improved. Structured tool-call completion is not faster. The reply after a simulated tool receipt starts 102/204 ms sooner, while total time remains approximately 0.397 seconds.

Timed decode prompts have 1993 tokens, 1986 cached, and 256 output tokens. Requests, output text/tool arguments, output lengths and cache counts match across arms. All 180 rows pass, including two images, overlapping requests, cancellation and recovery. Uncached warmups show essentially unchanged prefill duration. MLX allocated high-water is identical at 22.428 GiB; process RSS varies across phases, including a higher patched first pair, so no process-memory-neutrality claim is made.

[Exact scripts, every sample, independent audit, ranges and memory caveats](https://gist.github.com/samfenwick/84d1fed6dd548c06451e9a444167b920).

This measures API streaming, not complete-message application delivery. Output parsers, stop strings or a non-default stream interval can still buffer the first generated chunk. There is no long-context prefill or universal application-latency claim, and no custom two-second burst setting in these measurements.

### Validation

- Upstream CI passes on Python 3.11, 3.12 and 3.13 at `586af504` ([run](https://github.com/jundot/omlx/actions/runs/35783430293)).

- Five finalized regressions fail against unchanged main, including delivery through the real executor/event-loop/collector boundary. Tests also cover prefill-only steps, multi-token first chunks, late admission, empty outputs and normal later bursts.
- `python -m pytest -q tests/test_engine_core.py tests/test_output_collector.py`: **120 passed**, including existing cancellation tests.
- Default suite: **14,477 passed, 150 skipped, 85 deselected, 6 failed**. Untouched main has 14,470 passes and the exact same six failure identities (three FP32 decode-SDPA, one GLM native-fused parity and two Qwen4 hyper-connection cases). All seven added cases pass; no existing outcome changes or new failures. This is not a clean local full-suite pass.
- Independent code/evidence review, changed-code Black and whitespace checks pass. Whole-file Ruff has the same existing findings as main.


jundot (approved):
The overall processing speed remains essentially unchanged, but delivering the first chunk sooner should make responses feel more immediate. I think this is a worthwhile improvement to perceived responsiveness. Thanks for spotting this.

PR #3855, opened 2026-09-22
fix(cluster): keep the HTTP stack out of worker rank imports

### LINKED ISSUE #3519: CUDA worker venv is built without FastAPI, but the shipped bundle imports it at module scope
## Summary

`WORKER_REQUIREMENTS` in `omlx/cluster/cuda_worker_bootstrap.py` contains no `fastapi`, `uvicorn` or `starlette`, but the worker source bundle ships modules that import FastAPI at module scope. On my DGX Spark the rank died at startup with:

```
ModuleNotFoundError: No module named 'fastapi'
```

recorded as `.error` in `~/.omlx/cluster/runtime/<deployment>-rank-N.json`.

## Detail

The venv built from `WORKER_REQUIREMENTS` has none of these:

```
$ /opt/omlx-cluster-worker/venv/bin/python -c "import fastapi"
ModuleNotFoundError: No module named 'fastapi'
```

but the extracted worker release ships the whole `omlx` package, including:

- `omlx/admin/__init__.py:5` — `from .routes import router as admin_router, ...`, and `omlx/admin/routes.py` imports `fastapi` at module scope
- `omlx/cluster/routes.py:23` — `from fastapi import APIRouter, Header, HTTPException, Query, Response`

So anything on the rank path that touches either package tree pulls in FastAPI.

## What I can and can't confirm

I hit this on 0.6.2. Installing `fastapi uvicorn sse-starlette` into the worker venv fixed it, and that install has persisted across re-bootstraps to 0.6.4 — which means **I can no longer reproduce the exact import site**, because the dependency is now present.

I did check that importing `omlx.cluster.inference_worker` alone does *not* touch FastAPI on 0.6.4 (verified with a `sys.meta_path` hook), so the import is lazy and happens later in rank startup. I didn't want to uninstall the package on a working node to chase it further.

So: the missing dependency and the module-scope imports are both definitely real; the precise chain that reaches them during rank startup I have not pinned down.

## Possible fixes

1. Add `fastapi`, `uvicorn` and `starlette`/`sse-starlette` to `WORKER_REQUIREMENTS`.
2. Avoid importing the HTTP surface from a worker rank at all — a rank arguably has no reason to load `omlx/admin` or `omlx/cluster/routes.py`.

(2) seems better if it's achievable, since it also keeps the worker venv smaller; but that's your call, which is why this is an issue rather than a PR.

## Diagnosability note

As with #3518, the failure reaches the operator only as

```
rank-zero connection closed while the cluster was ready (RemoteProtocolError)
```

The launcher's `stderr_tail` retained MLX's *cleanup* traceback (a `CalledProcessError` from `launch.py:95` failing to kill an already-dead pid) and discarded the rank's actual exception. The real error existed only in the worker's rank state file, and that file is removed on teardown — so it has to be caught while the failure is live. Surfacing `.error` from the rank state file in the API response would have saved a lot of time here.

## Environment

oMLX 0.6.2 (worker), reproduced against `WORKER_REQUIREMENTS` as it still stands on `main` @ 94530d8d. Mac Studio M5 Max coordinator + NVIDIA DGX Spark (GB10, Ubuntu 24.04.4 aarch64), enrolled via the admin UI's CUDA worker card.

ashhart:
Fixes #3519.

## Cause

CUDA worker venvs are built from `WORKER_REQUIREMENTS`, which has no FastAPI. The rank's thinking-budget hook, `inference_worker._install_thinking_budget_support`, imports `omlx.api.thinking`. Importing any `omlx.api` submodule runs `omlx/api/__init__.py` first, and that file imported the MCP routes, and with them FastAPI, at module scope. This is the import chain the report could not pin down:

```
inference_worker._install_thinking_budget_support
  -> omlx.api.thinking
  -> omlx/api/__init__.py
  -> omlx.api.mcp_routes
  -> fastapi
```

## Fix

`omlx.api` now loads `mcp_router` and `set_mcp_manager_getter` on first attribute access. The server already imports the router from `omlx.api.mcp_routes` directly, so it is unaffected, and `from omlx.api import mcp_router` still works. Worker venvs stay free of the HTTP stack instead of gaining FastAPI.

## Testing

- A new regression test runs the budget hook in a fresh interpreter where `fastapi`, `starlette`, `sse_starlette` and `uvicorn` cannot be imported. It fails on main with `ModuleNotFoundError: No module named 'fastapi'` and passes with this change.
- `tests/test_cluster_inference_worker.py`, `tests/test_thinking*.py`, `tests/test_mcp*.py` and `tests/test_server*.py`: 525 passed.
- Full CI selection (`pytest tests/ -m "not slow and not integration"`) on current main with this change, #3856 and #3857 applied: 14,227 passed, 379 skipped, 0 failed.


jundot (approved):
Lazy-loading the MCP exports removes the worker's FastAPI dependency while preserving the existing imports. I reproduced the import failure without the HTTP stack and verified that this change fixes it. LGTM.

PR #3856, opened 2026-09-22
fix(cluster): run planned Qwen3.5 pipeline stages under the MTP patch

### LINKED ISSUE #3518: The Qwen3.5 MTP patch replaces `__call__` with a non-pipeline-aware version, breaking pipeline-parallel serving

ashhart:
Fixes #3518.

## Cause

Two problems stop a planned pipeline of `qwen3_5` on current main.

1. **Load.** With the current mlx-lm pin, `Qwen3_5TextModel.pipeline(group, split=None)` calls `super().pipeline(group, split=split)`, as do the Qwen3-Next and Ministral 3 overrides. `install_unequal_pipeline_plan` replaces `PipelineMixin.pipeline` with a hook that took no `split` argument, so planned pipelines of these models fail at load with `TypeError: planned_pipeline() got an unexpected keyword argument 'split'`.
2. **Forward.** As reported, the Lightning MTP patch replaces `Qwen3_5TextModel.__call__` whenever the checkpoint is MTP-compatible, including with MTP off. The replacement iterated `self.layers`, indexed the cache with `fa_idx` and `ssm_idx` unguarded, and had no pipeline receive, send or gather.

## Fix

- The planned hook accepts `split` and refuses only an explicit value, because the shard plan owns the layer ranges. No caller passes one today.
- The MTP forward mirrors MLX-LM's pipeline handling. It iterates `pipeline_layers`, guards both masks, receives from and sends to the neighbouring stages, anchors the send in the last cache entry and gathers the final hidden state. It still returns the pre-norm hidden state the MTP head needs, and an unsplit model runs exactly as before. The forward keeps a single send and a single gather, so the `sampling_rank_only` validation accepts it like the upstream forward.

Skipping the replacement when MTP is off would leave pipeline serving with MTP on still broken, so the replacement itself is made pipeline-aware.

## Testing

- A new test applies a two-stage plan to a tiny MTP-patched Qwen3.5 with 8 layers. One stage holds only full-attention layer 7, so its `ssm_idx` is None. The last stage's output must equal the unsplit forward. On main the test fails at load with the `split` error. With only the planner change it fails because stage 1 never sends.
- A new planner test covers the `split` argument.
- Two real MLX ring ranks on one Mac with the same plan: greedy decode of a two-prompt batch matches a single-process run exactly, with `sampling_rank_only` on and off. On main both ranks fail at load.
- `tests/test_mlx_lm_mtp_patch.py`, `tests/test_cluster_planner.py`, `tests/test_model_loading.py`, `tests/test_glm_mtp_patch.py`, `tests/test_mtp_prompt_priming.py` and `tests/test_dflash_lifecycle.py`: 522 passed.
- Full CI selection (`pytest tests/ -m "not slow and not integration"`) on current main with this change, #3855 and #3857 applied: 14,227 passed, 379 skipped, 0 failed.

## Related

#3653 makes the same forward change. On current main it is not enough on its own, because since the mlx-lm pin update the planned hook rejects the `split` argument that Qwen3.5's `pipeline()` passes, so planned pipelines fail at load before the forward runs. That planner change is new here. If #3653 lands first, this PR reduces to the planner change.


ashhart:
Closing in favour of #3653, which makes the same forward change. The planner `split` fix from this branch is described in a comment there.

Issue #3521, opened 2026-09-08
sampling_rank_only deadlocks pipeline-parallel generation: non-coordinator rank blocks in all_sum before its stage send


trunksio:
## Summary

With `sampling_rank_only` enabled, a 2-rank pipeline deployment loads, reports `phase: ready`, and then **never produces a token**. Disabling `sampling_rank_only` on the same server, same model, same hosts makes it work immediately.

Verified against `main` @ 94530d8d (built from source), Mac Studio M5 Max (Metal, rank 0) + NVIDIA DGX Spark GB10 (CUDA, rank 1), ring backend over 2.5 GbE, `mlx-community/GLM-4.5-Air-4bit`.

## A/B on one server

| `sampling_rank_only` | result |
| --- | --- |
| `false` | activates; `The capital of France is` → `Paris. which is also the country's largest city. The` in **1.05 s**; sustained ~18 tok/s |
| `true` | activates, then rank spins at ~110% CPU for 4+ min, no tokens, readiness fails with `no rank-zero data for 300s while the cluster was ready` |

Nothing else changed between the two runs.

## Where it blocks

`py-spy dump` on the stuck worker rank (rank 1):

```
Thread 136114 (idle): "Thread-1 (_generate)"
    coordinator_generation_step (omlx/cluster/runtime_optimizations.py:513)
    next (mlx_lm/generate.py:1415)
    _next (mlx_lm/generate.py:1775)
    next (mlx_lm/generate.py:1855)
    next (omlx/cluster/telemetry.py:1407)
    _generate (mlx_lm/server.py:853)
    generate_on_rank_stream (omlx/cluster/inference_worker.py:132)
```

Line 513 is inside the `mx.async_eval(...)` at 510-514, whose first argument is `instance._next_tokens` — the lazy result of line 507:

```python
# Rank zero contributes the selected IDs; all other ranks contribute
# zeros. Every rank therefore advances the same local KV state without
# gathering a hidden-state tensor.
sampled = mx.distributed.all_sum(sampled, group=group)
```

So the non-coordinator rank has entered the token `all_sum` and is waiting for a peer that never joins. Corroborating: every collective socket between the ranks is `ESTAB` with **`0 0`** send and receive queues — nothing was ever transmitted — while 60 of the rank's 61 threads sit in `futex_do_wait` and one spins (MLX's ring backend busy-polls).

## Likely cause

[...]

PR #3857, opened 2026-09-22
fix(cluster): agree on rank-zero sampling across ranks before using it

ashhart:
Refs #3521. This removes one way to get exactly the reported hang. I could not reproduce the hang with ranks that agree on their capabilities, so this may not be the only cause.

## What I found

- **Matching capabilities do not hang.** Two MLX ring ranks on one Mac ran a tiny `glm4_moe` through `install_runtime_optimizations` with the default balanced settings. Decode matched a single-process run with 4-bit weights, a two-prompt batch, a repetition-penalty processor, a long staggered prefill, and with `async_overlap` off.
- **The proposed cause does not apply to this path.** `glm4_moe` has no rank-zero logits contract, so workers take the full-logits branch rather than the MiniMax cache-anchored one, and the pinned mlx-lm's `glm4_moe` does anchor its send in `cache[-1].keys`. In that branch the stage send is upstream of the logprobs evaluated in the same `async_eval`, and MLX 0.32.2 executes the deepest nodes first, so the send runs before the token all-sum.
- **Disagreeing ranks do hang.** Each rank validates rank-zero sampling, the rank-zero logits contract and the staggered prompt against its own stage, model adapter and MLX-LM source, then picks its collectives from the result. With one rank on MLX-LM's synchronized path, which gathers hidden states, and the other on rank-zero sampling, which joins a token all-sum, the ranks wait on different collectives and never produce a token. On Metal the GPU watchdog ends the stalled command buffer after about ten seconds. A CUDA rank has no such watchdog and keeps waiting, which matches the report. Disabling `sampling_rank_only` puts both ranks on the same path, which matches the reported workaround.

## Fix

Before choosing a path, the ranks exchange one small `all_sum` vote over the three capabilities and keep each one only if every rank supports it. A capability dropped this way reports that another rank could not use it, so the rank state files say why. The vote runs only when the deployment enables `sampling_rank_only` and is pipeline-parallel across more than one rank. Both come from the deployment, so every rank makes the same decision about voting.

## Testing

- New unit tests cover the vote itself, and a peer that rejects rank-zero sampling, in which case nothing is patched. The existing installer tests use a fixture that models agreeing ranks.
- Two real ranks, one of which cannot validate rank-zero sampling: on main they deadlock until the Metal GPU timeout. With this change both fall back and match a single-process decode. With both ranks capable, rank-zero sampling and prefill overlap stay active and decode matches.
- `tests/test_cluster*.py`, `tests/test_distributed_engine.py` and `tests/test_nccl_fabric_worker.py`: 1374 passed.
- Full CI selection (`pytest tests/ -m "not slow and not integration"`) on current main with this change, #3855 and #3856 applied: 14,227 passed, 379 skipped, 0 failed.

If the reporter still has the setup, the `optimizations` block in each `~/.omlx/cluster/runtime/<deployment>-rank-N.json` shows whether the two ranks resolved `sampling_rank_only` differently.

## Related

#3644 takes a different approach to the same issue: it evaluates the worker's logits before the token all-sum. In the full-logits branch `_next_logprobs` is computed from those logits and is part of the same `async_eval`, so on MLX 0.32.2 the stage send already runs before the all-sum, which matches the two-rank runs above. The two changes are independent and could both land.


jundot (approved):
Agreeing on capabilities before installing the patches prevents ranks from selecting incompatible collective paths. I verified local two-rank fallback with mismatched capabilities and retained optimizations when both ranks agree; both matched single-process token output. LGTM for this fix, though the original Mac/CUDA report still needs confirmation.

PR #3860, opened 2026-09-23
feat(vlm): add MiMo V2.6 image, video, and audio support

sayyidfareed:
### Summary

Add native image, sampled-frame video, and audio understanding for MiMo V2.6 checkpoints that bundle MLX multimodal sidecars.

This extends the existing MiMo text path instead of introducing a separate server or inference loop. MiMo continues to use oMLX's normal OpenAI-compatible API, model pool, scheduler, cache, and generation path.

This PR contains runtime support only. It does not include model weights. A compatible Hugging Face checkpoint will follow separately.

### What changed

- Detect MiMo V2.6 checkpoints with bundled vision and audio sidecars.
- Compose the existing MiMo language model with:
  - the MiMo vision encoder;
  - the MiMo audio tokenizer;
  - the MiMo audio-to-language bridge.
- Accept OpenAI-compatible `image_url`, `video_url`, and `input_audio` message parts.
- Sample video into ordered frames before normal vision processing.
- Preserve multimodal content through request normalization, token estimation, and VLM execution.
- Discover MiMo checkpoints as VLMs when the required sidecars are present.
- Keep text-only behavior unchanged when no multimodal input is supplied.

### Checkpoint layout

The loader expects converted tensors in this layout:

```text
<model>/
├── config.json
├── audio_tokenizer/
│   ├── config.json
│   ├── tokenizer_config.json
│   └── model.safetensors
└── omnimodal/
    ├── config.json
    ├── manifest.json
    ├── vision_encoder.safetensors
    └── audio_encoder.safetensors
```

These are converted Xiaomi tensors, not newly trained weights. The tested checkpoint adds approximately 3.6 GB:

- vision encoder: 1.4 GB;
- audio encoder/bridge: 498 MB;
- audio tokenizer: 1.7 GB.

### Correctness details

The MLX implementations follow Xiaomi's reference model rather than treating MiMo as a stock Qwen-VL checkpoint.

Vision support includes MiMo's grouped-query attention, learned attention sink, row/column attention order, patch merger, and placeholder-to-feature validation.

Audio support includes:

- the reference 24 kHz mel pipeline and power spectrum;
- loading Encodec codebooks that MLX does not expose through normal module parameters;
- reference-compatible final-group padding;
- the causal decoder mask used by the reference Qwen2 audio bridge;
- expansion of each audio placeholder to the produced embedding count.

On the same recording, the MLX tokenizer produced the same 1,140 codec IDs as Xiaomi's PyTorch reference. Every audio-bridge layer also matched the reference numerically after applying the causal mask.

### API examples

Images and video use the existing OpenAI-compatible content blocks. Audio uses the standard `input_audio` shape:

```json
{
  "model": "MiMo-V2.6-Flash-RL-MLX-4bit-MTP",
  "messages": [
    {
      "role": "user",
      "content": [
        {
          "type": "input_audio",
          "input_audio": {
            "data": "<base64 wav>",
            "format": "wav"
          }
        },
        {
          "type": "text",
          "text": "Transcribe this recording."
        }
      ]
    }
  ]
}
```

### Validation

Focused regression suite:

```text
625 passed in 4.30s
```

Command:

```bash
pytest -q \
  tests/test_mimo_v2_audio.py \
  tests/test_mimo_v2_vision.py \
  tests/test_video.py \
  tests/test_api_utils.py \
  tests/test_model_discovery.py \
  tests/test_vlm_audio_fallback.py \
  tests/test_vlm_engine.py
```

Live M3 Ultra checks with the converted MiMo V2.6 Flash checkpoint:

- synthetic speech transcribed exactly as `The secret phrase is silver moonlight.`;
- Xiaomi's published audio sample was understood correctly;
- a labeled image was read correctly;
- a two-frame video returned both frame labels and border colors correctly;
- a text-only request still returned the requested exact text.

Ruff and `git diff --check` pass.

### Current limitations

- Video is sampled into vision frames; this PR does not decode or interleave a video's audio track.
- MiMo audio processing currently supports one prompt batch per request.
- Model weights and their conversion recipe will be published separately on Hugging Face after the runtime PR is available for linking.


sayyidfareed:
Update: `f7d6fcb2` handled standalone image requests but did not fix the same error in an ongoing conversation. Earlier text-only messages still went through mlx-vlm’s unsupported MiMo formatter. That raised an exception and sent the entire conversation through a fallback that dropped the image placeholder, leaving `0 placeholders, 8 features`.

`8ae555cb` formats every MiMo message on the native path, including text-only history. I added a regression test with system, user, and assistant text before the image request.

The focused suite passes (626 tests). I also retried the original failing request on the M3 Ultra; it now works.

jundot:
I'm downloading the official checkpoint in its original precision. I'll test it locally and follow up with feedback once the download and testing are complete.

jundot (approved):
I pushed the following fixes to this PR:

- Support torch-free loading and oQ conversion of the official MXFP4/FP8/BF16 checkpoint, preserving packed MXFP4 experts.
- Preserve vision/audio weights and audio tokenizer files during conversion, and recognize converted multimodal checkpoints.
- Include audio identity in prefix-cache keys while preserving reusable prefixes before changed recordings.
- Keep audio in its original conversation turn, including histories with multiple recordings.
- Match the official FP32 router projection before expert selection.
- Add regression coverage for conversion, audio cache identity, conversation placement, and router precision.

I verified conversion and SSD restoration with the real model. Merging with these fixes included.

PR #3861, opened 2026-09-23
feat(mimo): add checkpoint-native MTP decoding

sayyidfareed:
> Stacked on: #3860

### Summary

Add native multi-token prediction for MiMo V2.6 checkpoints and preserve it across the oMLX VLM prefill-to-decode handoff.

Text-only requests use the checkpoint's native MTP sidecar. Requests containing image, video, or audio embeddings automatically use normal autoregressive decoding because MiMo's multimodal embeddings are not compatible with the draft history expected by the MTP head.

This keeps the fast path for text without allowing speculative state to corrupt multimodal generation.

### What changed

- Extend the MiMo model with its checkpoint-native next-token predictor.
- Discover MTP tensors stored in `mtp/model_mtp.safetensors`.
- Pass the sidecar path through normal model loading instead of merging a second checkpoint on disk.
- Load both supported fused-QKV layouts:
  - FP8 weights with inverse scales;
  - already-quantized MLX weights with packed scales and biases.
- Split fused Q/K/V tensors at MiMo's actual query and key row boundaries.
- Implement the MTP cache lifecycle needed by oMLX speculative decoding.
- Preserve native MTP after multimodal-model construction and the VLM prefill handoff.
- Register the extended implementation for both `mimo_v2` and `mimo_v2_flash` so the released model aliases use the same path.
- Prefer the oMLX extension over the upstream text-only implementation only for MiMo checkpoints that need these capabilities.
- Disable drafting when `pixel_values`, `video_pixel_values`, or `audio_codes` are present; text-only requests continue to use MTP.

### Checkpoint layout

The loader expects the existing MiMo model plus:

```text
<model>/
└── mtp/
    └── model_mtp.safetensors
```

The sidecar contains converted checkpoint tensors. This PR does not add or train model weights.

### Why multimodal requests bypass MTP

MiMo's native predictor drafts from language-model token history. Image, video, and audio requests replace placeholder tokens with external embeddings during prefill. Reusing the text-only draft state after that replacement produced semantically wrong multimodal output even though requests completed successfully.

The safe behavior is therefore explicit:

- text only: native MTP enabled;
- image input: native MTP disabled;
- video input: native MTP disabled;
- audio input: native MTP disabled.

The base model and multimodal encoders are unchanged; only speculative drafting is skipped for those requests.

### Validation

Focused MTP regression suite:

```text
347 passed in 26.24s
```

Command:

```bash
pytest -q \
  tests/test_mimo_v2_patch.py \
  tests/test_mlx_lm_mtp_patch.py
```

The combined native-MTP, audio, and vision suite also passes:

```text
380 passed in 24.66s
```

Live warm text benchmark on an M3 Ultra:

```text
model:                     MiMo-V2.6-Flash-RL-MLX-4bit-MTP
prompt tokens:             31
completion tokens:         108
server TTFT:               1.88 s
client first visible token: 2.05 s
generation speed:          55.03 tok/s
```

Live image, video, and audio requests were also checked after the safety guard. All three returned correct content, while a following text-only request continued to load the MTP sidecar and returned the requested exact output.

Ruff and `git diff --check` pass.

### Review notes

- This is intentionally separate from the larger multimodal implementation so the model-loading, QKV conversion, cache behavior, and speculative safety rules can be reviewed independently.
- Because this branch is stacked, GitHub will include #3860's commits in the comparison until #3860 merges; only the MTP commits will remain afterward.
- No new dependency is introduced.
- The Hugging Face model repository and conversion recipe will be published separately and will link back to this PR.


jundot:
I used the MiMo model and VLM integration from this PR as the starting point for #3877. I couldn't merge it as-is because rejected drafts could leave the sliding-window cache in the wrong state after rotation, and oQ conversion did not preserve and calibrate all three MTP heads across the supported checkpoint layouts.

Those fixes needed to work together across loading, conversion, and decoding, so I consolidated them with the rollback approach from #3871 into one PR based on current main. I credited your work in the description and verified the resulting oQ4e checkpoint with text, image, audio, and video requests.

Issue #3837, opened 2026-09-22
Console visual update: one design language for the admin UI, and the app's Logs screen


LXD-8:
# Console visual update: one design language for the admin UI, and the app's Logs screen

This issue is the overview for the console's visual layer — and the macOS app's Logs screen — as they stand. The console side is **one PR**: the console's tests are one suite (the shared layer's guards scan every template), so the layer cannot land green without the screens that use it, and a smaller split would mean opening PRs that stay red until the rest arrives. The app side is three PRs, each on its own branch. Every one of them links back here.

**In scope:** the web console **except the chat page**, and the app's **Logs screen**.
**Out of scope:** the chat page (drawn exactly as it was), anything outside the console and that
screen, and every API, route and backend behaviour — nothing behind the UI changed.

## The PRs

**Web console — one PR: [#3839](https://github.com/jundot/omlx/pull/3839)** (`omlx/admin/**`, plus the app's own text modifiers, which the console's
guards measure against the same ramp):

* **the design tokens** — `tokens.json` (the palette in both appearances, the type ramp, the spacing
  scale, the control fills, the badge tones and the label colours — one family per category
  tag, so a theme can recolour them) and `build_tokens.py`, which renders
  `static/css/tokens.css` and the app's numeric `DesignTokens.swift` from it, with `--check` failing
  when either artifact drifts. The self-hosted Inter and Noto payload is no longer bundled: the
  console draws with the platform stack the app itself uses.
* **the shared layer** — `format.js` (one number, size, duration and count formatter, in the
  console's own language), `ui.js` (badge, switch, segmented control, card, table, empty state,
  sheet, toast), the string catalogue the templates read through `t()`, one stylesheet built from
  the tokens, and the top bar, navigation and page frame every page now shares.
* **the screens rebuilt on it** — Status (one header card, KPI cards, the memory meter with the
  enforcer's two limits marked), Logs (records with folded continuations, `×N` repeats, a detail
  card, one colour per level), Settings (an anchor rail, a search box,
  Appearance, one shape for the sub-tabs), Models (one status vocabulary, one status chip, one
  switch and one row layout, per-model settings in a sheet), Bench (cards with presets, a progress
  line that says what it measures, a destructive-run confirmation) and the polish pass (one command
  palette, one toast, skeletons, the cluster wizard, the keyboard map).

**macOS app — three PRs** (`apps/omlx-mac/**`, each on its own branch):

| PR | what it holds |
|---|---|
[...]

PR #3863, opened 2026-09-23
fix(admin): a description in the model sheet uses the whole card

LXD-8:
**A description in the model sheet now uses the whole card.** Every hint under a control sits inside that control's label column, so a hint that needs two lines wraps inside about half the card while the rest of the row stays empty. That layout is the one on `main` today.

Measured in the sheet's advanced-settings column, same window, before and after:

| | the hint's box | what the lines do |
|---|---|---|
| before | 170px — the label column next to the switch | lines fill those 170px and stop |
| after | **222px — the card's content width** | the same text, one line where it fits |

The hints are lines of their own now, under the label and the switch. Nothing else changes size: the labels, the switches and the inputs keep their widths, the left column is untouched, and the sheet is no taller.

**Standalone:** two files only — the model sheet template and its test — on top of the current `main`, with no dependency on the visual-update series (#3837), so it can be reviewed and merged before that batch.

`tests/test_admin_model_settings_template.py` gains one assertion: no label column holds a `<p>` any more.

**Verified:** `python3 -m pytest -q tests/ -m "not slow and not integration"` — 14 334 passed, no failures, with the current `main` merged in.

### Screenshots

| before | after |
|---|---|
| the hint inside the label column | the hint on its own line, spanning the card |

*Effect preview: both shots come from a build that also carries the pending visual-update work (issue #3837), so colours and card borders in the right-hand shot are not part of this PR. What this PR changes is the hint's placement: inside the label column (170px) on the left, its own line under the label and the switch, spanning the card (222px) on the right.*


LXD-8:
This PR is independent of the visual-update series (#3837) and can be merged on its own: it changes only the model sheet template and its test file, sits on top of the current `main`, and depends on none of the series' PRs.

What it changes is where the hint sits: inside the label column it wrapped in half the card (170px); it is now a line of its own under the label and the switch, spanning the card (222px). The before/after screenshots in the description are an effect preview taken on a build that also carries the pending series work — only the hint's placement belongs to this PR.

Verified on top of `main`: 14 332 passed, no failures.

PR #3865, opened 2026-09-23
feat(moe): expert-offload support for mimo_v2_flash + streaming performance work

errrepe:
### Summary

Teaches MoE expert offload (SSD streaming) to work with `mimo_v2_flash` checkpoints, makes the streaming path faster for every supported model family, adds a memory guard so an over-large resident fraction is refused instead of crashing at first request, makes the resident size **automatic** — computed from the memory budget and then grown or shrunk by a live governor — and exposes the feature in the native macOS app, where it was missing.

Verified end to end on a real MiMo-V2.6-Flash-RL-MLX-4bit-MTP checkpoint: the compatibility gate now accepts it, and the same ~2k-prefill / 128-decode workload runs ~2.5x faster to first token and ~22% faster per decode token.

### What changed

Each mechanism was benchmarked on its own and kept only if it measurably helped. The ones that did not (higher IO worker counts, post-prefill cache seeding, resident-first chunk ordering, batched slot writes) were tried, measured, and removed — nothing dead is left in the diff.

**Compatibility gate** (`moe_offload_compat.py`)

- Admits `mimo_v2_flash`. The gate now understands MiMo's real checkpoint: `n_routed_experts` instead of `num_experts`, `moe_layer_freq` so the dense layer 0 is skipped, the stacked `mlp.switch_mlp.{gate,up,down}_proj` layout, and MXFP4 (`mode="mxfp4"`, group size 32) expert quantization.

**Streaming performance** (`moe_expert_offload.py`) — applies to every family the generic `ExpertCache` wraps (`mimo_v2_flash`, `qwen4_exp`, `qwen3_5_moe`, `gemma4`, `olmoe`):

- *Reads in disk order.* Miss fetches are sorted by expert id, which is also the on-disk offset order. What was effectively random I/O becomes near-sequential. TTFT 110.7s → 60.9s in isolation.
- *Working-set protection.* An expert needed later in the same call can no longer be evicted earlier in that call. Fewer refetches; decode +2%.
- *Host-side chunk IDs.* The prefill chunker already holds expert ids as numpy; it no longer round-trips them through the GPU. Per-tensor read sizes are cached instead of recomputed.
- *Read-ahead during prefill.* While the GPU evaluates chunk N, the IO pool already reads chunk N+1's missing experts. Bytes only — installs still happen serially inside `ensure_ids`, so cache state is identical. Traced: staged-chunk wait −30%.
- *Hot-set pinning.* Decode routing is heavily skewed (measured: top-32 experts cover ~78% of routes on MiMo). `ensure` (the decode path) counts routes and the top `OMLX_MOE_OFFLOAD_PIN × capacity` experts are refused as eviction victims. The prefill path never teaches the table — measured not to transfer. Decode +18%. Default 0.875, `0` disables.

**Memory guard** (`moe_expert_offload.py`, `routes.py`, `model_loading.py`)

- The settings surface checked that the resident fraction was in `(0, 1]` but never whether `fraction × experts + non-expert weights` fits the machine. Enabling offload on a too-large checkpoint failed later as a Metal out-of-memory at first request.
- `moe_offload_memory_check` estimates the resident set with the same per-family accounting the wrapper delivers (reusing `estimate_offload_admission_bytes`; `deepseek_v41` delegates to its own estimator) and compares it against 75% of the device's recommended working set, leaving headroom for activations, KV and in-flight reads. Wired into PATCH validation, the feature-group check, the pre-load hook, and `apply_moe_expert_offload` itself, so every entry path fails fast with real numbers. `OMLX_MOE_OFFLOAD_MEM_LIMIT_GB` overrides the limit; `0` disables the guard.

**Automatic residency + dynamic governor** (`moe_expert_offload.py`, `model_settings.py`, `engine_pool.py`, `vlm.py`)

- `moe_expert_offload_resident_fraction` accepts `null`/`0` as **automatic**. At load, `resolve_moe_offload_fraction` binary-searches the largest resident capacity the memory guard's static budget admits (never more, so the governor's ceiling is provably safe), then trims the starting point by the memory actually free right now.
- A governor re-evaluates every dynamic cache at `OffloadSwitchGLU.__call__` entry — a call boundary, so a resize can never invalidate in-flight reads. It watches two signals: system free RAM (8% desperate / 15% target watermarks, ported from the previous expert-streaming design) and cache hunger (new misses). Under pressure each layer shrinks by exactly the deficit — no blind halving; with spare memory and a hungry cache it grows by the affordable share. Cooldowns are asymmetric (15s shrink / 30s grow) and desperation bypasses them, so shrink→regrow cannot flap.
- `ExpertCache` gains real `resize()` — slots are reallocated, pins and LRU order are honored, the routing floor (`top_k`) is a hard bound, and `min_capacity`/`max_capacity`/`slot_bytes` make the cost model explicit. Resizes stagger across layers one call at a time instead of one multi-GiB burst.
- Explicit fractions keep their meaning: a user-pinned percentage stays fixed unless `OMLX_MOE_OFFLOAD_DYNAMIC=1` opts it into the governor as a shrink-only pressure valve (it never grows past the request). `deepseek_v41` auto mode delegates to its adapter's own `fit_resident_fraction`, which accounts for Engram residency.

**UIs** — web dashboard and native macOS app

- Both pickers gain an **Automatic** entry (sentinel `0`) so nobody has to guess a percentage; the manual field stays for explicit residency. The native app also decodes `moe_expert_offload_supported` and gets the same toggle + picker the web UI has, profile-aware, with the same conflict gating against MTP / VLM MTP / DFlash. Localized in all shipped locales.

**Tests**

- New compat-gate cases prove MiMo passes and that the dense layer is really skipped (flipping `moe_layer_freq` to mark it MoE fails as expected). Pinning gets a dedicated test: decode-shaped calls teach it, host-path calls never do. The memory guard gets fit/refuse/disabled/apply-raises coverage, and the governor gets unit tests for proportional shrink, floor-under-desperation, hungry growth, cooldown hysteresis, and fixed-fraction isolation.

### Benchmarks

Machine: **MacBook Pro, Apple M4 Pro, 48 GB unified memory**. Models on an **external NVMe SSD over USB4** (`/Volumes/SSD 4TB`). Workload: ~2,000-token prompt prefill + 128 decode tokens, single request, MTP off.

MiMo-V2.6-Flash-RL-MLX-4bit-MTP (~153 GiB, 256 experts/layer, MXFP4):

| Build | Mode | TTFT | Decode | Expert cache hit rate |
|---|---|---:|---:|---:|
| Before this PR | fixed 0.125 | 110.7 s | 1.308 tok/s | 47.1% |
| After | fixed 0.125 (32/256) | 43.4 s | 1.590 tok/s | 49.5% |
| After | **automatic** | 45.3 s | 1.075 tok/s | 43.1% |

The automatic run sized itself to ~29 of 256 slots (11.2%) and held a 21.7 GiB footprint — it converged to the residency this machine's real free-RAM band sustains, shrank under prefill pressure and regrew afterwards, and never came near the Metal limit. The hand-picked 0.125 still edges it on decode; automatic's win is that "pick a number" stops being a user task and an OOM class disappears.

Other checkpoints on the same machine, through the same `moe_expert_offload` setting:

| Checkpoint | Size | Fraction | TTFT | Decode | Result |
|---|---:|---:|---:|---:|---|
| GLM-5.3-Flash-oQ4e (`glm5_next` adapter) | 169 GiB | 0.0625 | 68.8 s | 1.18 tok/s | ran clean, 23.8 GiB footprint |
| Qwen3.8-Flash-Next-oQ4e-mtp | 99 GiB | any | — | — | guard refuses: ~38.4 GiB resident est. even at 6.25% (non-expert alone is 34.4 GiB) |
| DeepSeek-V4.1-Flash-oQ4e-mtp | 402 GiB | any | — | — | guard refuses: ~159.5 GiB resident est. |

Two honest consequences fall out of the measurement:

- This feature fits a checkpoint when its **non-expert** weights plus the resident fraction stay inside the working set. On 48 GB that means "mostly-experts" checkpoints (MiMo, GLM-5.3); Qwen3.8's 34 GiB dense backbone and DeepSeek-V4.1's 127 GiB one cannot stream their way under the limit at any fraction — the new guard says so up front instead of dying in Metal.
- GLM-5.3 runs through its upstream family adapter (`deepseek_v4`/`glm5_next` native blocks), so its numbers show the feature working, not this PR's generic-cache optimizations. MiMo is the one that exercises the optimized `ExpertCache` path. The glm5_next adapter does not report into `moe_offload_stats`, so no hit rate is shown for it.

Caveat: absolute TTFT carries run-to-run page-cache noise (±20–40% observed across a session). Keep/reject decisions used same-session A/B runs or traced per-chunk wait, never cross-session comparisons.

### Validation

```text
$ pytest tests/test_moe_expert_offload.py tests/test_moe_offload_compat.py tests/test_deepseek_v41_moe_offload.py tests/test_model_settings.py
232 passed   # incl. pinning, MiMo gate, memory guard, auto-resolve and governor cases

$ moe_offload_compatibility('<real MiMo checkpoint>')
(True, '')

$ moe_offload_memory_check on real checkpoints (M4 Pro 48 GB)
MiMo @0.125      -> OK (~21.9 GiB est, 28.0 GiB budget)
MiMo @auto       -> fitted ~0.17 ceiling, sized to free RAM at load
GLM-5.3 @0.0625  -> OK (~19.6 GiB est)
GLM-5.3 @0.125   -> refused (~29.6 GiB est) — confirmed by a real Metal OOM
Qwen3.8 @0.0625  -> refused (~38.4 GiB est) — confirmed by a real Metal OOM
DSv4.1  @0.125   -> refused (~159.5 GiB est)

$ automatic mode, live run (MiMo, 2k/128)
load: 47 layers wrapped, auto-resolved fraction
resident: 0.166 start -> governor-converged 0.112 (peak Metal 21.7 GiB, no OOM)

$ xcodebuild test-without-building -scheme oMLX -only-testing:oMLXTests/ModelSettingsScreenVMTests -only-testing:oMLXTests/LocalizationSmokeTests
Executed 30 tests, with 0 failures
```

macOS note: the app-hosted test run needs `ENABLE_HARDENED_RUNTIME=NO` on ad-hoc-signed builds, and the full suite has a pre-existing failure (`ThemeTests.testLightWindowBackgroundUsesStandardWindowColor`, fails on clean main too) plus a fixture-read hang on removable-volume checkouts — both unrelated to this change.

Generated with [Devin](https://devin.ai)


jundot:
The MiMo support and streaming improvements are useful additions. I checked the new residency and error-handling paths and found three issues I'd like to resolve before merging.

- The governor applies the same resize delta repeatedly within its one-second window. At the same timestamp, four calls reduced a cache from 16 to 8 slots instead of applying the two-slot reduction once.
- The V4.1 guard doesn't consistently account for Engram SSD offload. On the real checkpoint at 12.5% residency, the adapter estimated 47.4 GiB while the guard estimated 159.5 GiB. Automatic mode also loses the Engram setting at the post-load check.
- With an injected read failure, `ensure_ids()` returned while another read for the same expert was still running. The current group needs to remain covered by cleanup, along with staged reads if computation fails.

Could you address these with focused regression tests? I'll handle the two small UI inconsistencies in a follow-up: the native app's supported MTP combinations and saving the displayed percentage when switching from automatic to manual. Once the runtime fixes are ready, I'll rerun the reproductions and inference checks for merge.

PR #3866, opened 2026-09-23
fix: refresh stale prefill memory samples

### LINKED ISSUE #3858: Prefill guard rejects every prompt with an unchanged stale `current` figure while the admin API reports low usage — 0.6.4 (post-#2179-fix)
## Environment

- oMLX 0.6.4 (build 2529) — the latest stable release. It includes the
  #2179 fix (merged 2026-07-21; 0.6.4 released 2026-08-29), so this report
  is a live gap on the current release, not a stale-version report.
- Model serving chat traffic: Qwen3.5-4B-OptiQ-4bit, alongside ASR and VL
  engines on the same server (separate engines; per source the STT engine
  does not run through a Scheduler at all — engine/base.py:442-448).
- macOS 26, Apple Silicon, 64 GB; prefill_memory_guard on, tier balanced.
- A gateway fronts the server.

## Observed incident (2026-09-19 ~01:24 UTC → 2026-09-20 ~04:10 UTC)

For ~27 hours, every chat prefill on the model above was rejected. The
displayed `current` figure never moved (44.88 GB) while the prompt-dependent
parts of the diagnostic varied with the prompt (for example predicted peaks
of 44.97, 45.05, 45.09 and 45.11 GB across kv_len 0…2144). One verbatim
diagnostic (of many; `server.log.2026-09-20`):

```
oMLX prefill memory guard rejected this prompt: Prefill context too large
for available memory (preflight safety guard, kv_len=143, min_chunk=32):
predicted peak would require ~45.05 GB (current 44.88 GB + KV 54.62 MB +
min-chunk transient 122.03 MB) but prefill safety cap is 42.75 GB (90% of
metal_cap ceiling 47.50 GB). Raise kernel iogpu.wired_limit_mb in Terminal
(currently caps Metal at 48.00 GB), or reduce context length. [...]
```

During the same window:

- the admin API reported low process usage with zero active and zero
  waiting requests (operator-observed during the incident; per source,
  idle admin accounting reads fresh measurements independently of the
  scheduler's cache — process_memory_enforcer.py:773-787,
  admin/routes.py:5532-5585 — which is precisely why the admin view and
  the guard disagreed);
- a small non-streaming probe (12 prompt tokens after templating,
  max_tokens=5) was rejected like everything else;
- ASR and VL engines on the same server served normally throughout;
- `POST /admin/api/models/<id>/unload` + `/load` (~7.7 s) restored serving
  immediately.

## Mechanism (from the installed source — with an explicit caveat)

What the code shows:

- Each Scheduler caches a process-wide `mx.get_active_memory()` sample
  (`_last_mlx_active_memory_bytes`, scheduler.py:4435-4437). The preflight
  guard reads it via `_current_usage_bytes(refresh_mlx_active=False)` =
  `max(cached, phys_footprint − hot_cache)`.
- Several paths DO refresh the cached sample on the executor: during
  admitted requests; idle reclaim (scheduler.py:9163-9190); deferred
  post-completion cleanup (12128-12149); a pre-admission cleanup retry
  (engine/base.py:102-134 → scheduler.py:9374-9380, present in the v0.6.4
  tag); and #2179's admission reclaim refreshes it when it actually runs
  (engine_pool.py:2067-2084 → scheduler.py:4824-4825).

Related work: #3226 (item 5) diagnosed a stale `_last_mlx_active_memory_bytes`
during decode and PR #3227 added a decode-side sample refresh that ships in
0.6.4 — the nearest existing diagnosis of this cache going stale. Our
incident is the adjacent case: a fully idle scheduler with nothing pending,
where no decode is running for #3227's refresh to ride on and the
admission path's early return (below) skips the reclaim refresh — which is
why we see this on 0.6.4.

The gap we believe fits this incident: `engine_pool.py:1934-1947` returns
early when fresh process measurements already fit the target — skipping
both the reclaim and the cached-sample refresh — after which
`engine/base.py:150` re-checks the now-stale cached sample and rejects. On
a fully idle scheduler with no pending cleanup or reclaim, no executor work
runs at all, so nothing refreshes the sample; rejections never admit, so
nothing changes — a steady state whose only visible symptom is the frozen
`current` figure.

**Caveat / label**: this is the mechanism consistent with our logs, not a
minimized reproducer. A normally-completing large prefill schedules
cleanup and refresh, so we are NOT claiming "send a big prompt and it
wedges" — the exact seeding sequence for a sample that stays stale through
all the refresh paths above still needs evidence, and we're happy to dig
further if useful.

## Why we do not think this is #2179's class

#2179's wedge was real pool-held footprint (112.75 GB genuinely resident);
its fix reclaims the pool and refreshes the sample when it runs. Our
incident shows the mirror picture: admin (fresh measurements) low, guard
(stale cache) high — and the early-return path above plausibly skipping
the very refresh that would clear it.

## Suggested direction

Timestamp the cached sample. When it is older than N seconds AND the
scheduler is idle with no pending cleanup/reclaim, refresh it on the
executor — or let the fresh process measurement win — before rejecting.
That closes the idle steady-state without touching the hot path.

Happy to attach sanitized timestamped log excerpts (guard diagnostics +
admin stats) if useful.

Chuhan1112:
## Summary
- mark executor-owned MLX active-memory samples stale after an idle interval
- refresh a stale sample once before route preflight evicts or rejects a request
- cover stale-sample detection and idle preflight retry

Fixes #3858.

## Validation
- `pytest tests/test_engine_preflight.py tests/test_scheduler_prefill_memory_guard.py -q` (67 passed)
- `git diff --check`


jundot (approved):
I reproduced the stale-sample rejection through the scheduler and engine preflight path, including the safety-cap branch. Refreshing on the owning executor admits the request without eviction when memory is available, while genuine memory pressure still rejects it. LGTM, ready to merge.

Issue #3834, opened 2026-09-22
Tool call bleed after 46111bf?


zviratko:
I can't find this reasoning content in the logs, so just a screenshot.

Not a usual occurence for me and it showed for the first time right after I picked up this commit, so reporting it.
But it also hasn't shown since, yet and might have been a fluke, I will post if it happens again.

<img width="1266" height="184" alt="Image" src="https://github.com/user-attachments/assets/a5e71ff8-fcee-4d2a-95fd-07d1d120cb84" />

jundot:
Confirmed it's a bug, and I think it needs a more careful fix. Looking into it now.
Thanks for reporting it quickly!

jundot:
@zviratko I reproduced the markup leak and tested token-ID-based detection using vLLM's parser with actual Qwen3.8-27B output. The model used the same marker token in explanations and real tool calls, so that approach still truncated explanations.

I haven't found a way to preserve those explanations without introducing regressions in tool-call handling, so I reverted 46111bf2 in e3982c3f. Thanks for reporting this so quickly.

zviratko:
This was with Qwen3.8-Flash-Next but I'm not using the original chat template.

Thanks for a quick revert :)

PR #3868, opened 2026-09-23
fix: recover literal tool markers after final parsing

### LINKED ISSUE #3794: Tool-call stream filter splits on the generic XML marker wherever it appears in visible output — a literal marker in prose gets withheld (5-model repro, deterministic no-weights min-repro)
## Summary

When a model emits the generic XML tool-call **start marker as literal text** — inside backticks, a code fence, or plain prose (i.e. it is *explaining* or *showing* the marker, not making a tool call) — oMLX's `ToolCallStreamFilter` treats it as the start of a tool-call envelope. Everything from the marker to end-of-stream is then withheld as an "unclosed tool-call envelope" (`take_recovery_candidate()`), and the visible response is **truncated right before the marker**.

This affects **every model that can emit the marker as text**, not a single family. I reproduced it on five models across three parser families, and with a deterministic filter-level minimal reproduction that needs **no weights and no HTTP**.

## Environment

- oMLX `0.7.0.dev4` (bundle dated 2026-09-18), Mac Studio (M-series, 512 GB)
- Models reproduced (5 models, 3 parser families):
  - `GLM-5.3-Flash-oQ8e-mtp` (glm47 parser family)
  - `GLM-5.3-Flash-oQ4e` (4-bit, same family)
  - `Qwen3.8-Flash-Next-oQ4e-mtp` (qwen parser family)
  - `Qwen3.8-27B-oQ8e-mtp` (qwen parser family)
  - `gpt-oss-120b-MXFP4-Q8` (harmony parser family)
- Marker involved: the generic XML tool-call start marker, **11 characters**: `chr(60) + "tool_call" + chr(62)` (opening angle bracket, the word `tool_call`, closing angle bracket). It is part of the `_marker_pairs` list in `omlx/api/tool_calling.py` and therefore active for **every** filter regardless of the model's native parser — which is why even a harmony-parser model (gpt-oss) triggers it.

## What happens

1. The model is asked to explain or show the marker (e.g. "write this start marker in backticks: <tool_call>").
2. The model emits the marker as literal text inside the visible output — it is **not** making a tool call (no function, no parameters).
3. `ToolCallStreamFilter.feed()` sees the marker and opens an envelope, buffering everything after it.
4. No matching close marker arrives (the model just continues with prose), so at `finish()` the buffered tail is stored as a recovery candidate and withheld.
5. The client gets `finish_reason=tool_calls` (or `stop`) with visible content **truncated right before the marker**, plus a server warning:
   `Unclosed tool-call envelope at end of stream; withheld N characters are available for content recovery (start_marker='...')`

## Deterministic minimal reproduction (no weights, no HTTP)

Run against the `ToolCallStreamFilter` in the installed oMLX bundle (Python 3.10+; `regex` + `jsonschema` available):

```python
import sys
RES = "/Applications/oMLX.app/Contents/Resources"
sys.path.insert(0, RES)
L, R = chr(60), chr(62)
MARKER     = L + "tool_call" + R          # generic XML start marker (11 chars)
MARKER_END = L + "/" + "tool_call" + R
FN_OPEN, FN_CLOSE = L + "function=", L + "/function" + R
P_OPEN, P_CLOSE   = L + "parameter=", L + "/parameter" + R

from omlx.api.tool_calling import ToolCallStreamFilter
class FakeTok: tool_call_start = None; tool_call_end = None
TOOLS = [{"type": "function", "function": {"name": "weather_lookup",
          "parameters": {"type": "object", "properties": {"city": {"type": "string"}},
                         "required": ["city"]}}}]

def run(label, chunks):
    f = ToolCallStreamFilter(FakeTok(), tools=TOOLS)
    emitted = [f.feed(c) for c in chunks]
    f.finish()
    withheld = f.take_recovery_candidate()
    print(label)
    print("  emitted :", repr("".join(emitted)))
    print("  withheld:", len(withheld), "chars", "-> BUG" if withheld else "")

# A: marker as literal text in backticks (model explains it, does NOT call)
run("A: marker in backticks", [
    "A serving server detects tool calls via this start marker: `",
    MARKER, "` (just this one). Two parsing details follow. SENTINEL: all there.",
])
# B: marker inside a code fence (model shows an example)
run("B: marker in code fence", [
    "A raw model output starts like this:\n```\n", MARKER, "\nweather_lookup\n```\n",
    "Explanation outside the block. SENTINEL: all there.",
])
# C: control — a real, correctly closed tool call (must parse cleanly)
run("C: real closed call", [
    "Calling the weather.\n",
    MARKER + "\n" + FN_OPEN + "weather_lookup" + R + "\n"
    + P_OPEN + "city" + R + "\nBerlin\n" + P_CLOSE + "\n" + FN_CLOSE + "\n" + MARKER_END,
    "\nDone.",
])
```

Observed output (oMLX 0.7.0.dev4):

```
A: marker in backticks
  emitted : 'A serving server detects tool calls via this start marker: `'
  withheld: 86 chars -> BUG
B: marker in code fence
  emitted : 'A raw model output starts like this:\n```\n'
  withheld: 84 chars -> BUG
C: real closed call
  emitted : 'Calling the weather.\n\nDone.'
  withheld: 0 chars
```

Case **C** confirms the filter handles genuine tool calls correctly. Cases **A/B** confirm that a literal marker in prose is treated as the start of an envelope and everything after it is withheld.

## Server log (live)

`grep "Unclosed tool-call envelope" ~/.omlx/logs/server.log*` (each warning is logged twice; counts below are deduplicated per second):

- 2026-09-20: 15 events (models active in that window: `GLM-5.3-Flash-oQ8e-mtp`, `Qwen3.8-27B-oQ8e-mtp`, and one `gpt-oss-120b-MXFP4-Q8` turn at 19:28 that was itself a marker-in-prose probe)
- 2026-09-21: 28 events (models active in that window: `Qwen3.8-27B-oQ8e-mtp`, `GLM-5.3-Flash-oQ4e`, `Qwen3.8-Flash-Next-oQ4e-mtp`)

Withheld sizes range from 143 to 26,602 characters. Example line (start_marker truncated):

```
2026-09-20 19:28:04,692 - omlx.api.tool_calling - WARNING - [-] - Unclosed tool-call envelope at end of stream; withheld 2463 characters are available for content recovery (start_marker='<tool_call>`‑Marker‑Tag verarbeitet. Sobald das Modell das Marker‑Tag im 
```

Note: with concurrent requests the per-warning model attribution from "next Chat completion line" is not reliable, so I report per-day totals plus the models that were active in the window. The gpt-oss line above is a direct marker-in-prose probe, so it is unambiguous.

## Why this is distinct from #2420 / #3153 / #3442

Those issues are in the same "unclosed envelope" mechanism family, but their trigger is a **real (incomplete) tool call** — the model genuinely starts a call and omits the closing marker. Here the trigger is different: the model **never intends a tool call at all**; it merely emits the marker as literal text (in backticks / code fence / prose). The parser cannot distinguish "the model is explaining the marker" from "the model is starting a tool call", so it splits and withholds the rest of the turn.

## Self-reinforcing loop

The more a session discusses or shows the marker, the more likely the model is to emit it in prose → new truncations. This makes the bug especially nasty in agentic / meta contexts (e.g. debugging tool calling itself): an agent investigating the marker keeps re-triggering it.

## Proposed fix

Validate before treating the marker as the start of an envelope:

- Only treat it as a tool call if a registered function/parameter structure follows (as the MiniCPM5 attribute-style parser does in #3429 / #3530), **or**
- Validate the parsed tool name against the registered `tools` list and pass through as content when it is not registered (the guard discussed in PR #3669), **or**
- At minimum: if an envelope opened by a marker in *prose position* (preceded by backticks / code fence, not at turn start) does not close within a small token window, flush it as content rather than withholding to EOF.

## Reproduction scripts

- Filter level (no weights): `omlx_filter_minrepro.py`
- Live HTTP: `glm_oq4e_test.py`, `qwen_flash_marker_test.py`

Happy to provide the full scripts and raw log excerpts.

Chuhan1112:
## Summary
Fixes #3794 for the narrow code-display scope agreed in review.

- Keep streaming suppression conservative: a code-context generic marker is held until final parsing.
- Confirm a balanced inline-code span or fenced block before recovering it.
- Recover content only after final parsing, and preserve structured-call precedence.
- Keep malformed or incomplete candidates on the existing error path; they are not emitted as content.
- Align Chat Completions, Anthropic Messages, and Responses, including thinking/reasoning channels.
- Compute balanced code contexts once per text so repeated malformed envelopes remain linear.
- Leave plain prose mentions unresolved for now.

## Tests
- `pytest tests/test_tool_calling.py tests/integration/test_e2e_streaming.py -q` — 712 passed, 12 deselected
- `pytest tests/test_output_parser.py tests/test_sse_keepalive.py tests/test_openai_adapter.py tests/test_responses.py -q` — 212 passed, 3 skipped
- `python -m py_compile omlx/api/tool_calling.py omlx/server.py`
- `git diff --check`

Added coverage for:
- inline and fenced code markers across chunk boundaries
- literal markers before real calls
- malformed calls retaining `invalid_tool_call`
- unbalanced code spans retaining `incomplete_tool_call`
- thinking/reasoning recovery
- `<|im_start|>` markup not leaking through code-context recovery
- linear behavior on repeated malformed envelopes


jundot:
I reproduced three regressions on this branch.

1. **The #3834 markup leak returns.** With `<tool_call>` followed by `<|im_start|>function: execute_code>`, the prefix classifier rejects the unfamiliar tag and treats the payload as prose. The raw call then appears in reasoning output across Chat Completions, Anthropic Messages, and Responses.
2. **A malformed call becomes a successful response.** For `<tool_call>not a function</tool_call>`, the existing `invalid_tool_call` error disappears. Responses emits `response.completed` instead of `response.failed`.
3. **Raw markup leaks even when a structured call is available.** When the engine returns structured `tool_calls` alongside `Before <tool_call>malformed`, EOF recovery emits the malformed text as content. The existing behavior keeps the structured call and discards that raw candidate.

Fixing these three cases alone would not establish that the ambiguity is resolved. The underlying problem is treating "not recognized as a supported call" as "safe to emit as prose." An unsupported, malformed, or incomplete call can fail the same checks as an explanation. Adding exceptions for these particular strings would leave that assumption in place.

I need a parsing approach that distinguishes ordinary content from failed or incomplete call candidates without bypassing the existing error handling or structured-call precedence. Where the available protocol information is insufficient, the code needs an explicit policy for ambiguous output. Token IDs alone did not resolve this in my Qwen3.8 tests because explanations and real calls used the same marker token. I can't consider this resolved based only on the literal-marker cases passing.

Chuhan1112:
You're right. The root issue is my classifier treats "unrecognized" as prose, but unrecognized can also be a malformed or incomplete call. Patching those three examples wouldn't fix the assumption.

I'd like to rework this around a stricter rule:

- Keep streaming suppression conservative.
- Decide at final parse, not from the short prefix.
- Structured calls win; malformed candidates keep the existing error path.
- Only emit recovered text when final parsing can prove there is no tool-call semantics.
- If the output remains ambiguous, keep the current safe behavior, even if that still truncates some #3794 examples.

Code fences or backticks could be a high-confidence hint, but not the main discriminator.

I'm marking this draft while I rework it. Does this direction match what you'd want?

jundot:
That direction makes sense to me, especially preserving structured-call precedence and the existing error handling. The part I couldn't resolve in my own attempt was deciding when an unparsed candidate is actually ordinary text. Moving that decision to EOF gives more context, but I'm not sure it removes the ambiguity.

Do you have a particular recovery rule in mind? A few concrete examples might help establish what it can safely recover and what should remain unresolved. I'm comfortable with a narrower improvement that leaves some #3794 cases open.

Chuhan1112:
Thanks. The narrow rule I'd try:

1. Keep the candidate suppressed by default.
2. At EOF, recover it only when the marker is inside a balanced inline code span or fenced code block.
3. If final parsing returns a call or an error for that candidate, keep the existing structured-call/error behavior.
4. If a structured call exists elsewhere, recover a code-context candidate only when it is outside the recovered call envelope.

Examples I'd recover: a marker in inline backticks or a fenced block, including when a valid call follows it.

Examples I'd leave unresolved or suppressed: plain prose mention, a malformed uncoded payload, im_start-style markup, and a malformed uncoded candidate alongside a structured call.

That gives up the general prose case, but it should cover the highest-confidence #3794 examples without guessing that "unrecognized" means ordinary text. Does that boundary feel right to you?

jundot:
I'm comfortable starting with that narrower scope and leaving the general prose case unresolved. Recovering only the cases that can be handled reliably, while preserving the existing call and error behavior, sounds like a reasonable first step.

Chuhan1112:
Implemented the narrower rule we discussed.

The generic marker is now preserved only when it appears in inline code or a fenced code block. Plain prose, malformed calls, incomplete calls, and structured-call precedence keep the existing behavior. I also aligned the non-streaming parser so a code-context marker no longer hides a later real call.

Added regression tests for:
- inline-code and fenced-code markers across chunk boundaries
- a literal code-context marker followed by a real call
- plain prose and malformed structured calls staying suppressed

`tests/test_tool_calling.py` and `tests/integration/test_e2e_streaming.py` both pass locally. I kept the PR in draft while you look.

jundot:
The narrower scope still makes sense to me, but the implementation currently skips suppression as soon as it sees a code-context hint. That differs from the proposed rule of confirming a balanced span and preserving the final call/error decision.

I reproduced three remaining issues: malformed code-context candidates leak before an error, a real call after a closing fence leaks with certain chunk boundaries, and the inline-marker example still ends with `incomplete_tool_call` on the Qwen API path.

Rather than patching each example separately, I think the next step is to check these behaviors end to end: the same output should behave consistently across chunk boundaries, recovered prose should remain prose at final parsing, and call/error candidates should not be emitted as content first.

Chuhan1112:
Reworked this along those lines.

The stream filter now holds a code-context marker instead of emitting it early. At final parsing, recovery only happens for a balanced inline-code span or fenced block. Structured calls and errors keep precedence; malformed or incomplete candidates stay on the existing error path.

I also covered the three cases you reproduced end to end across Chat, Anthropic, and Responses, including chunk boundaries and thinking/reasoning. Unbalanced code spans now remain `incomplete_tool_call` rather than being emitted as prose, and im_start-style markup is not recovered through the code-context path.

`tests/test_tool_calling.py` and `tests/integration/test_e2e_streaming.py` both pass locally. Still draft while you look.

vogel61:
@Chuhan1112 @jundot — as the reporter of #3794, one observation from reading this thread and 46111bf2/e3982c3f: all three of the remaining issues jundot reproduced today are release-decisions made **mid-stream** (code-context hint seen → suppression skipped before the span or payload is confirmed). 46111bf2 failed for the same structural reason, just with a payload-prefix hint instead of a fence hint.

I sketched an EOF-only variant in #3794 (comment linked below); the short version, applied to this PR:

1. Never release anything mid-stream. When a start marker is seen, suppress exactly as today and stage the withheld bytes. The stream path stays byte-identical to main, so no chunk-boundary ordering can differ between chunk sizes.
2. Decide **once at EOF**, over the complete withheld buffer, not a prefix:
   - final parsing yields a call or an error → existing structured-call / error behavior wins unchanged (covers jundot's points 2 and 3);
   - otherwise, if the marker sits in a balanced inline-code span or fenced block (balanced = confirmed on the full buffer, which streaming can never guarantee) → emit the staged text as content;
   - ambiguous → today's safe behavior (withheld, recovery candidate).
3. Because release only ever happens at EOF and only after final parsing declined the buffer, "recovered prose stays prose at final parsing" holds by construction — nothing can be emitted as content first and reinterpreted later.

The prose-without-code-context case stays unresolved here too, which matches the narrow scope you both agreed on. My own live check this morning (0.7.0.dev4, Qwen3.8): a bare marker in plain prose with no tools registered streamed through fine — the truncation needs the tool-enabled path, so these tests should keep registering tools.

Not trying to claim this closes the general ambiguity — it just moves the one decision point to where the full evidence exists, which is exactly the failure mode your three reproductions keep hitting.

Cross-ref: https://github.com/jundot/omlx/issues/3794#issuecomment-5793086329

Chuhan1112:
Thanks @vogel61. That matches the rework now on the branch: nothing is released mid-stream. A code-context candidate is held, final parsing decides, and structured calls or errors keep precedence.

CI also caught a performance regression in my first final-parse version because code-context detection rescanned the prefix once per marker. I changed that to compute balanced fenced/inline-code intervals once per text, preserving the linear behavior guarded by `test_unwind_stays_linear_on_repeated_malformed_envelopes`.

The relevant tool-call and streaming suites pass locally, including chunk-boundary and malformed/incomplete cases.

PR #3869, opened 2026-09-23
feat(cluster): carry the Mac-to-CUDA stage edge over MCDMA RDMA links

ashhart:
In a Mac + CUDA ring deployment every pipeline activation crosses MLX's TCP ring, and the Mac's end of that ring is usually 10 GbE. A Mac with a ConnectX card driven by [MCDMA](https://github.com/ashhart/MCDMA) has an RDMA path to the CUDA workers. This PR moves the rank 1 to rank 0 stage activation onto that path once oMLX has proven the link live. The token all-sum, the final all-gather and every other stage edge stay on the ring.

## Design

- **oMLX never opens an RDMA device.** MCDMA's `mcdma-rpcd` daemons own the queue pairs on both hosts. oMLX exchanges bytes with them through shared-memory mailboxes and uses a small C helper, `libmcdma-rpc`, for acquire/release word loads and stores. A rank that crashes or is killed cannot leave a queue pair behind.
- **Stage links are decided per launch.** They travel in the worker contract only. They are never stored and are excluded from deployment equality, so the stored deployment schema does not change.
- **One edge in v1.** With the Mac as rank 0, rank 1 to rank 0 is the only stage edge between the Mac and a CUDA worker. Edges between Sparks already use their ConnectX ring hop.
- **Pull protocol.** The receiving rank requests each frame and pre-posts the request for the next message, so the sender never writes into a buffer that is still being read. Every frame carries a 128-byte header (message, frame, offset, dtype, shape) that the receiver checks against the tensor it expects.

## How the link is proven live

1. **Daemon status.** The Mac's daemon must report the link up, and the link must resolve to exactly one enrolled worker, the one holding rank 1.
2. **Byte-checked probe.** A short-lived probe service starts on the worker over the cluster SSH policy, with the SSH target and Python the launch uses for rank 1. The probe runs content-checked round trips, a bulk send the worker must acknowledge with a matching CRC-32, and a bulk receive that must match a seeded pattern. One wrong byte fails it. **Verify** in the dashboard runs the full probe; every launch runs a quick one first. Evidence expires after 24 hours, or when the daemon version, peer, device, mailbox sizes, connection time or loaded MCDMA driver changes.
3. **Rank agreement.** After loading, each rank attaches its end of the mailbox, then all ranks exchange one `all_gather` vote. An edge goes live only if both ends attached and both daemons report the link up. The vote also orders every attach before any request.

If any step fails, the launch goes ahead over the ring and the cluster status reports why under `stage_links`, including when the ranks' vote turned an edge off. While serving, a waiting rank rechecks the daemon's link flag, a link-generation word and its own service registration at least once a second. A reconnect loses the request in flight, so the daemons bump the generation and end the service registration when it happens: both ranks fail at once instead of waiting out the timeout, the deployment stops with an error, and the next launch re-verifies and falls back to the ring.

A link is claimed by one owner at a time, either a running deployment or a dashboard verification, so a probe can never take the worker's end away from a live rank. The daemon also refuses a second service on a link with `ERR busy`.

## Changes

- `omlx/cluster/rdma/`: new package of small modules. Mailbox layout and word ops, client and service mailboxes, stage frames, daemon status parsing, link discovery against enrolled nodes, the probe and its service, evidence and its store, the per-launch decision, the in-rank transport, and the admin handlers.
- `deployment.py`: `stage_links` field, validated, carried in the worker plan and decoded by `decode_worker_stage_links`.
- `launch.py`: the supervisor re-verifies links just before building the launch command, frees them after a proven teardown or when a launch fails before any rank starts, and launches over the ring if the check itself raises.
- `inference_worker.py`: installs the stage links before the runtime optimizations, so the deferred prefill send and rank-zero sampling wrap the RDMA-aware send. Each rank reports its edge state in its marker and `rank_ready` event.
- Routes: `GET /admin/api/cluster/rdma-links` and `POST /admin/api/cluster/rdma-links/verify`, both on the admin-gated cluster router. The dashboard card appears only when the Mac's daemon answers. i18n keys are added to every locale, translated for `zh` as the rest of `cluster.v2` is.
- `docs/rdma-links.md`, linked from the heterogeneous cluster guide. It includes the mailbox protocol for anyone writing another daemon.
- Settings: `OMLX_RDMA_STAGE_LINKS=0` turns the feature off; `OMLX_MCDMA_RPCD_SOCKET` and `OMLX_MCDMA_RPC_LIBRARY` point at a non-default daemon socket or helper.

## Testing

- 114 new unit tests: word packing and frames, mailbox exchange and service registration through a Python stand-in for the daemon, a flipped byte or a short answer failing the probe, evidence expiry and damaged evidence files, the per-launch decision, where it probes and its claims, the rank vote, link loss and quick reconnects on both ranks, the admin routes, and the supervisor hooks. Two more run the probe service through the real `libmcdma-rpc` when `OMLX_MCDMA_RPC_LIBRARY` points at it, and are skipped otherwise.
- One integration test (`-m integration`): two real MLX ring ranks on one Mac decode a small qwen2 through `install_runtime_optimizations` with the stage edge on the stand-in daemon. Tokens match a single-process decode, rank-zero sampling and prefill overlap stay active, and the stand-in counts at least one reply per generated token, so the traffic did cross the link.
- Full CI selection (`pytest tests/ -m "not slow and not integration"`): 14,336 passed, 381 skipped, 0 failed.
- The worker-side modules import with FastAPI, Starlette, Uvicorn and Pydantic blocked, as on a CUDA worker's venv; the probe service reports a missing helper as a clean `ready: false`.
- Ruff is clean on the new files and reports nothing new in the changed ones.
- Not yet run end to end on ConnectX hardware with this code.

## Expected effect

Prefill gains the most, because prompt activations are large. A decode step moves one small activation and still waits on the ring's token all-sum, so decode speed changes little. KV hand-off and disaggregated prefill are the bigger follow-ups.

## Related

- The daemon and helper come from MCDMA (`mcdma-rpcd`, `libmcdma-rpc`); see its [link daemon guide](https://github.com/ashhart/MCDMA/blob/main/docs/link-daemon.md). Without them nothing changes: the dashboard card stays hidden and launches take the ring.
- Independent of #3857. If both land, every rank runs the stage-link vote and then the capability vote, in the same order.


jundot (approved):
Without mcdma-rpcd nothing changes, and if any RDMA step fails the launch just uses the ring, so this is safe to merge.

I'll follow up on main with two small fixes. First, the probe service closes its registration right after staging BYE, and rpcd_listen handles that EOF before it relays the staged reply, so a healthy link can time out on END (about half the probes in a C copy of the listen loop). I'll wait for the reply half's ready word before closing. Second, concurrent Verify calls share the VERIFYING claim, so I'll reject a second one with 409.

Both fixes stay in link_probe_service.py and link_routes.py, so #3870 should rebase cleanly. A prefill A/B on real ConnectX hardware would be good to see when you have one.

LGTM, merging.

